PR #4106, opened 2019-08-19
[nrf52xxx] use Rx SFD time in Rx timestamp for promiscuous mode

brianljt:


examples/platforms/nrf52811/radio.c
```diff
@@ -733,17 +729,15 @@ void nrf_802154_received_raw(uint8_t *p_data, int8_t power, uint8_t lqi)
         receivedFrame->mInfo.mRxInfo.mAckedWithFramePending = false;
     }
 
-#if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
-    // Get the timestamp when the SFD was received.
-    uint32_t offset =
-        (int32_t)otPlatAlarmMicroGetNow() - (int32_t)nrf_802154_first_symbol_timestamp_get(time, p_data[0]);
-    receivedFrame->mInfo.mRxInfo.mTimestamp = otPlatTimeGet() - offset;
-#else
+#if !OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
```
bukepo: I don't think we need this any more. We should always fill the timestamp field.

brianljt: Sure.

examples/platforms/nrf52811/radio.c
```diff
@@ -733,17 +729,10 @@ void nrf_802154_received_raw(uint8_t *p_data, int8_t power, uint8_t lqi)
...
     uint32_t offset =
         (int32_t)otPlatAlarmMicroGetNow() - (int32_t)nrf_802154_first_symbol_timestamp_get(time, p_data[0]);
-    receivedFrame->mInfo.mRxInfo.mTimestamp = otPlatTimeGet() - offset;
-#else
-    if (otPlatRadioGetPromiscuous(sInstance))
-    {
-        receivedFrame->mInfo.mRxInfo.mTimestamp = nrf5AlarmGetCurrentTime();
-    }
-#endif
+    receivedFrame->mInfo.mRxInfo.mTimestamp = nrf5AlarmGetCurrentTime() - offset;
```
bukepo: Since `NRF_802154_TX_STARTED_NOTIFY_ENABLED` can be overridden, it would be better to verify it is enabled here.
```
#if !NRF_802154_TX_STARTED_NOTIFY_ENABLED
#error "NRF_802154_TX_STARTED_NOTIFY_ENABLED is required!"
#endif
```

PR #4110, opened 2019-08-19
[posix-app] remove node id from logging

bukepo:
* Simplified posix-app logging
* typedef otLogLevel to signed int so that when dynamic logging and
  reference_device is enabled, there will not be always true error.

### DIFF
--- a/include/openthread/platform/logging.h
+++ b/include/openthread/platform/logging.h
@@ -110,7 +110,7 @@ extern "C" {
  * This type represents the log level.
  *
  */
-typedef uint8_t otLogLevel;
+typedef int otLogLevel;
 
 /**
  * This enumeration represents log regions.
--- a/src/posix/platform/logging.c
+++ b/src/posix/platform/logging.c
@@ -30,65 +30,46 @@
 #include "platform-posix.h"
 
 #include <assert.h>
-#include <inttypes.h>
 #include <stdarg.h>
-#include <stdio.h>
 #include <syslog.h>
 
 #include <openthread/platform/logging.h>
-
-#include "code_utils.h"
-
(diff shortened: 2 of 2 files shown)


src/posix/platform/logging.c
```diff
@@ -30,65 +30,46 @@
...
-
-    otEXPECT_ACTION(charsWritten >= 0, logString[offset] = 0);
+    va_list args;
 
-exit:
     switch (aLogLevel)
     {
     case OT_LOG_LEVEL_NONE:
-        logLevel = LOG_ALERT;
+        aLogLevel = LOG_ALERT;
```
abtink: Suggestion: how about using an `int logLevel` here instead of re-using `aLogLevel` argument? Mainly since we may in future change `aLogLevel` to be an enumeration for OT log levels and we want to convert it to `systlog` log level.

include/openthread/platform/logging.h
```diff
@@ -110,7 +110,7 @@ extern "C" {
  * This type represents the log level.
  *
  */
-typedef uint8_t otLogLevel;
+typedef int otLogLevel;
```
abtink: See the other comments. I think with that we can keep this to be `uint8_t`.

bukepo: Changing the type is not for reusing `aLogLevel`. When dynamic logging level is enabled, otDumpCertMeshCoP() is using `OT_LOG_LEVEL_NONE`(i.e. 0), https://github.com/openthread/openthread/blob/master/src/core/common/logging.hpp#L1903 will always be true, and some compiler will complain and fail.

abtink: I see. Sounds good!

PR #4111, opened 2019-08-19
[code] remove unnecessary extern

bukepo:


abtink:
I fully agree with @LuDuda. Use of `extern` to indicate if an API is callback or not is indeed very useful.

jwhui:
Agree with @LuDuda. The real purpose of `extern` in this case is for the developer.

bukepo:
Thank you for the explanation! I was considering to add `-fvisibility=hidden` and only export symbols in `include/*`, thoughts?

PR #4112, opened 2019-08-19
[cmake] initial implementation

jwhui:
WORK IN PROGRESS - intended as an initial prototype for discussion.

Currently only supports posix example platform.

POSIX example:
```
$ cmake -GNinja -DOT_PLATFORM=posix .
$ ninja
```

CC2538 example:
```
$ cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=examples/platforms/cc2538/arm-none-eabi.cmake -DOT_PLATFORM=cc2538
$ ninja
```

No platform example:
```
$ cmake -GNinja -DOT_PLATFORM=none .
$ ninja
```

TODOs:

- [x] add cross-compile support
- [x] add cmake checks for library functions
- [x] add option to build without example platform
- [x] add option to build without executables
- [x] add option for builtin mbedtls
- [x] add Travis test


jwhui:
@LuDuda, the plan is to continue supporting autotools for the the time being (in addition to CMake). We can reevaluate at a later date if/when users/projects make the switch to CMake.

CiaranWoodward:
This sounds great for openthread library useability! Currently every cmake project that uses openthread as a library seems to have it's own unique cmake-autotools binding, and I've found autotools to be very slow.

I think to enable this use case, it's important that the openthread project doesn't override anything set by the parent cmake project, and also properly exports the transitive dependencies with the openthread targets (eg. using ``target_include_directories`` instead of ``include_directories`)

It would also be a nice-to-have in future if this allowed building from a windows environment.

Are you ready for Reviews/Contributions on this PR?

jwhui:
@CiaranWoodward, thanks for the initial feedback. Yes, please feel free to provide reviews/comments on this PR.

rlubos:
It looks very promising! With some tweaks to the top-tree CMakeLists.txt file, I was able to build the OpenThread with Zephyr samples. Not sure if those could be avoided (changes like passing compiler options from Zephyr to OT), I would need to consult some Zephyr CMake expert.

Most of the TODOs from the bullet point list would be nice to have for Zephyr integration, especially the ability to build w/o target platform or builtin mbedtls. 

I'm not sure about the `add option to build executables` bullet point - it seems to already be in there? Or did you mean build **without** executables (libraries only)? I guess that could be interesting use-case for external projects using OpenThread.

jwhui:
@rlubos, thank you for your comment. Yes, I did mean "add option to build **without** executables.

rlubos:
@SebastianBoe FYI

.gitignore
```diff
@@ -1,5 +1,6 @@
 *.a
 *.bak
+*.cmake
```
SebastianBoe: It is convention for CMake sources to either be named CMakeLists.txt or *.cmake.

So we would normally not want to gitignore *.cmake.

jwhui: Thanks for the comment. I wanted to target files autogenerated by CMake. I have pushed updates to be more specific. Thoughts?

examples/platforms/posix/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+#  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+#  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+#  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+#  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
```
CiaranWoodward: On Linux, this platform fails to link, as librt is not linked. To resolve this:

```
find_library(LIBRT rt)
if(NOT LIBRT)
    unset(LIBRT)
endif()
```

then below:
```
target_link_libraries(openthread-posix PRIVATE ${LIBRT})
```

CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+get_filename_component(__root_dir "${CMAKE_CURRENT_LIST_DIR}" ABSOLUTE)
+set(OT_ROOT_DIR ${__root_dir} CACHE INTERNAL "OpenThread source root.")
+
+include("${OT_ROOT_DIR}/etc/cmake/checks.cmake")
+
+execute_process(
+    COMMAND "${OT_ROOT_DIR}/third_party/nlbuild-autotools/repo/scripts/mkversion" "-b" "`cat .default-version`" "."
+    OUTPUT_VARIABLE __version OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+set(OT_VERSION "${__version}" CACHE INTERNAL "")
```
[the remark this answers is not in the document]
jwhui: No good reason. Updated.

etc/cmake/arm-none-eabi.cmake
```diff
@@ -0,0 +1,52 @@
...
+#  2. Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in the
+#     documentation and/or other materials provided with the distribution.
+#  3. Neither the name of the copyright holder nor the
+#     names of its contributors may be used to endorse or promote products
+#     derived from this software without specific prior written permission.
+#
+#  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+#  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+#  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
```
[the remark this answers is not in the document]
jwhui: Makes sense to me. Are you suggesting to have a `toolchain.cmake` file in each of the `examples/platforms/*` directories?

examples/platforms/posix/CMakeLists.txt
jwhui: Updated. FWIW, the `if(NOT LIBRT)` trick did not work for me. Instead, I've wrapped `target_link_libraries` for now. Let me know if there's a better solution.

examples/apps/cli/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+)
+
+target_include_directories(ot-cli-ftd PRIVATE ${COMMON_INCLUDES})
+target_include_directories(ot-cli-mtd PRIVATE ${COMMON_INCLUDES})
+
+target_link_libraries(ot-cli-ftd
+    openthread-cli-ftd
+    openthread-ftd
+    openthread-platform-utils
+    ${OT_PLATFORM_LIB}
```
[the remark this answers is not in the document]
jwhui: Do you have a preferred path forward? :)

.gitignore
SebastianBoe: Great, those files are safe to ignore.

SebastianBoe:
Zephyr's CMake sources could integrate with openthread's CMake sources in one of two ways. Either Zephyr creates a new CMake process (externalproject_add) or Zephyr invokes add_subdirectory(${OT_REPO}).

The integration is stronger and therefore the system as a whole faster and easier to use when add_subdirectory is used. But on the other hand when a new process is created the coupling is looser and therefore more independent and easier to maintain.

I'd suggest staying with a separate process, as it (had to be) for autotools, and then possibly migrating to supporting add_subdirectory when needed.

etc/cmake/arm-none-eabi.cmake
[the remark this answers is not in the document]
CiaranWoodward: Yes, I think this is a good option. If building without ot examples, the parent project provides the toolchain file.

examples/apps/cli/CMakeLists.txt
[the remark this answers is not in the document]
CiaranWoodward: I think I'll have to give this some more thought - double linking (1) is the simplest solution if we want to get something out quickly, but makes it more complex to use the openthread library, so we might want to reconsider later.

examples/platforms/posix/CMakeLists.txt
CiaranWoodward: Yeah, that looks good to me, just need to change
``if(HAVE_RT)``
to
``if(LIBRT)``
'HAVE_RT' doesn't mean anything in cmake

etc/cmake/arm-none-eabi.cmake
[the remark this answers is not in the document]
jwhui: Sounds good to me. Updated.

examples/apps/cli/CMakeLists.txt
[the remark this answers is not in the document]
jwhui: I went with (1) just to get things working.

examples/platforms/posix/CMakeLists.txt
jwhui: Thanks, updated.

jwhui:
I'd like to move this forward. What are people's thoughts on the status of this PR? Of course, we can modify as needed through future contributions.

**/cc/** @LuDuda @rlubos @CiaranWoodward @SebastianBoe

CiaranWoodward:
Looks good! I have some suggestions on my branch (https://github.com/CiaranWoodward/openthread/tree/cmake), with reasoning in the commit messages. Feel free to merge them into your branch if you agree. As you say, we can do further improvements in future with PRs.

Firstly, I have a build error when trying to compile your repo as is. This is part of a recurring error we keep seeing as there are dependencies from certain modules such as platform & platform-util back to the core. Because we have multiple versions of the core (ftd and mtd), it is difficult to properly represent this dependency. This is the same reason we are needing to double-link the platform at the add_executable point rather than allow cmake to figure it out automatically. I fixed this in the first commit for now, but this same problem may cause us other issues in the future that may cause it to be worth looking into other options for building mtd/ftd/radio seperately.

Two more points:
1. I'm generally against the 'globbing' that is done here, as it isn't run every time 'make' is run, so the user has to manually re-run 'cmake' if the files change. The recommendation I have generally read is to manually list out all files that are to be compiled.
2. I still want to look into the platform selection to verify it is correctly configured to work as a 'toolchain file'. I'll do this later today.

I have verified that this compiles on linux and windows with clang, gcc, make and ninja in various combinations. I think this is a great improvement vs autotools!

rlubos:
It looks really good!

I've spent some time today testing this solution, and it integrated with Zephyr pretty well (current solution turned out way cleaner than my previous attempt)! I could integrate OpenThread's CMake build system with (almost) no Zephyr-specific changes in existing files, which is great. You can check my branch [here](https://github.com/rlubos/openthread/commits/zephyr-cmake-integration-v2) (note that the integration is not complete, I still need to translate some Zephyr KConfig options like Joiner/Commissioner support).

The only thing I was missing, was `OT_BUILTIN_MBEDTLS` translation into `OPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS`, I did it like this:
```
if(OT_BUILTIN_MBEDTLS)
    list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS=1")
else()
    list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS=0")
endif()
```

Otherwise, I had no issues with integration. Probably we should consider applying @CiaranWoodward suggestions (like using CMake options instead of environmental variables for configuration).

I'm waiting eagerly to be able to use this solution 'officially' in Zephyr 👍

jwhui:
@CiaranWoodward, thanks for your review and contributions! I actually had the source files explicitly listed in the prior push. I just reverted to that. Would you be able to rebase your branch on this PR now so that I can cherry-pick your commits?

jwhui:
@rlubos, thank you for your review and contributions! I'd be happy to cherry-pick your `OT_BUILTIN_MBEDTLS` change into this PR.

CiaranWoodward:
Thanks! I have rebased - I'm just making a couple of changes to the platform integration then I'll be done.

Are you planning on adding travis integration before the merge?

jwhui:
@CiaranWoodward, I'll add basic travis tests for the posix and cc2538 platforms. I think we can enhance the travis tests further in future PRs.

CiaranWoodward:
@jhui I've updated my branch and tested building for the platform. I don't have the hardware set up to test the cc2538, but posix is working and all are building.

```bash
#!/bin/bash
set -e

git clone --branch cmake --depth=1 https://github.com/CiaranWoodward/openthread.git

mkdir ot-build
cd ot-build
cmake -DOT_PLATFORM=posix  ../openthread
make -j12
cd ../

mkdir ot-clang
cd ot-clang
CC=clang CXX=clang++ cmake -DOT_PLATFORM=posix ../openthread
make -j12
cd ../

mkdir ot-ninja
cd ot-ninja
cmake -GNinja -DOT_PLATFORM=posix ../openthread
ninja -j12
cd ../

mkdir ot-cc2538
cd ot-cc2538
cmake -DCMAKE_TOOLCHAIN_FILE=examples/platforms/cc2538/arm-none-eabi.cmake -DOT_PLATFORM="cc2538" ../openthread/
make -j12
cd ../
```

CiaranWoodward:
@googlebot I consent.

jwhui:
@CiaranWoodward @rlubos, thanks again for your help on this! I have pulled both of your commits into this PR. Please review and approve this PR if you think it is OK to merge.

.travis/script.sh
```diff
@@ -114,6 +114,13 @@ build_cc1352() {
 }
 
 build_cc2538() {
+    git checkout -- . || die
+    git clean -xfd || die
+    mkdir build && cd build || die
+    cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=examples/platforms/cc2538/arm-none-eabi.cmake -DOT_PLATFORM=cc2538 || die
```
CiaranWoodward: I believe you need to add a ``../openthread`` here (and to the posix one) to point to the source directory. If it works as-is then I'm wrong!

CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+option(OT_BUILTIN_MBEDTLS "Enable builtin mbedTLS" ON)
+
+include("${PROJECT_SOURCE_DIR}/etc/cmake/checks.cmake")
+include("${PROJECT_SOURCE_DIR}/etc/cmake/options.cmake")
+include("${PROJECT_SOURCE_DIR}/etc/cmake/functions.cmake")
+
+execute_process(
+    COMMAND "${PROJECT_SOURCE_DIR}/third_party/nlbuild-autotools/repo/scripts/mkversion" "-b" "`cat .default-version`" "."
+    OUTPUT_VARIABLE OT_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE
+)
```
CiaranWoodward: This doesn't seem to work properly, probably because it relies on a bash-style shell. Could try using the cmake ``FILE`` command to read .default-version instead of cat.

.travis/script.sh
jwhui: Good catch. Updated.

CMakeLists.txt
jwhui: Can you help suggest a command? :)

jwhui:
@LuDuda, thanks for your review and comments! I have added a mention to the `README.md`. I have also resolved the compile warnings for now by adding `-std=gnu99` for the `cc2538` platform.

rlubos:
@googlebot I consent.

CiaranWoodward:
@jwhui I have made 2 further commits to my branch to fix:
1. The map files being generated on top of each other on the cc2538
2. The version code was not being generated correctly

Note that the version code is still not being generated correctly on windows, but it is just set to blank. If the exact format isn't important I can put together a future PR in cmake-script to use git-describe in a cross-platform way. I think for now we are good to go!

etc/cmake/checks.cmake
```diff
@@ -0,0 +1,35 @@
...
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
+include(CheckFunctionExists)
+
+check_function_exists("strlcat" HAVE_STRLCAT)
+check_function_exists("strlcpy" HAVE_STRLCPY)
+check_function_exists("strnlen" HAVE_STRNLEN)
+
+configure_file(${PROJECT_SOURCE_DIR}/etc/cmake/openthread-config-generic.h.in ${PROJECT_SOURCE_DIR}/etc/cmake/openthread-config-generic.h)
```
rlubos: One final remark, a nit really. I've noticed that `openthread-config-generic.h` is the only file that is generated in-tree, when out-of-tree build is invoked (to reproduce simply create a `build` directory in the OT root and then invoke `cmake -GNinja -DOT_PLATFORM=none ..` from within).

If `${PROJECT_BINARY_DIR}` was used for the header destination, all of the build products would end up in the build directory. The include path for this file in the main `CMakeLists.txt` file would need to be updated accordingly (`list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_BINARY_DIR}/etc/cmake)`).

CiaranWoodward: Good catch! I agree with this change

jwhui:
@CiaranWoodward, thanks for the updates.

I am seeing the following output on macOS:
```
-- Looking for strnlen - found
bash: line 1: .: filename argument required
.: usage: . filename [arguments]
-- Version: 20180926-01075-gb07e7d45b
```
Is that expected?

etc/cmake/checks.cmake
jwhui: Thanks for the suggestion. I have incorporated it into this PR.

rlubos: Thanks 👍

CiaranWoodward:
@jwhui Version looks like it is being generated correctly, but the '.' is being processed as a second command. I fixed by removing the '.' at the end of the command. Mirrored branch of ``jwhui/cmake`` where I have edited history to fix is here: https://github.com/CiaranWoodward/openthread/tree/cmake

jwhui:
> @jwhui Version looks like it is being generated correctly, but the '.' is being processed as a second command. I fixed by removing the '.' at the end of the command. Mirrored branch of `jwhui/cmake` where I have edited history to fix is here: https://github.com/CiaranWoodward/openthread/tree/cmake

Thanks, @CiaranWoodward !

PR #4113, opened 2019-08-19
[sub-mac] relax the requirement for TxDone callback

abtink:
This commit changes the `SubMac::BeginTrasnmit()` by moving the call
to `Transmit()` on radio platform after the state change to
`kStateTransmit` (and pcap callback). This relaxes the requirement for
radio platform implementation and allow it to invoke `TxDone` callback
directly from the `Transmit()` call itself (instead of requiring
`TxDone` to be called only after the `Transmit()` call returns).

### DIFF
--- a/src/core/mac/sub_mac.cpp
+++ b/src/core/mac/sub_mac.cpp
@@ -286,15 +286,15 @@ void SubMac::BeginTransmit(void)
     error = Get<Radio>().Receive(mTransmitFrame.GetChannel());
     assert(error == OT_ERROR_NONE);
 
-    error = Get<Radio>().Transmit(mTransmitFrame);
-    assert(error == OT_ERROR_NONE);
-
     SetState(kStateTransmit);
 
     if (mPcapCallback)
     {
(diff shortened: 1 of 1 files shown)


src/core/mac/sub_mac.cpp
```diff
@@ -286,16 +286,16 @@ void SubMac::BeginTransmit(void)
     error = Get<Radio>().Receive(mTransmitFrame.GetChannel());
     assert(error == OT_ERROR_NONE);
 
-    error = Get<Radio>().Transmit(mTransmitFrame);
-    assert(error == OT_ERROR_NONE);
```
bukepo: If this must be successful, shall we make it no return?

abtink: `otPlatRadioTransmit()` can/should fail if the allowed radio state transitions are not followed. 
We follow the state transition in sub-mac/mac layer, so expect the call to be successful (therefor the `assert()` here). This is mainly intended to catch any platform implementation issues. I think we should keep it the same to catch/protect for this.

PR #4114, opened 2019-08-20
[mle] ensure to process repeated "Parent Response" from current parent candidate

abtink:
This PR contains two commits:

**[mle] process "Parent Response" from current parent candidate**
    
This commit updates MLE to continue to process a repeated
"Parent Response" from current parent candidate to ensure that the
challenge and frame counter are correctly updated.

**[toranj] test-case covering attach between single parent/child with poor link**

----------

Background of the issue:
- This applies when we have a single parent and child with poor link quality (or in case where radio platform does not provide RSSI/link quality info)
- The child goes through attach process
  - In the first phase child does "Send Parent Req to Routers" and hears from the parent, but does not select it (due to its poor link quality).
  - In the second phase child does "Send Parent Req to Routers and REEDs", and again gets a "Parent Response" from the same parent. 
  - Without the the change in this PR, the child would not fully process the new "Parent Response"  (from same parent) and and won't  update the "challenge".
  - At end of secon phase, child seletcs the parent, and sends "Child Id Request"
  - But parent rejects the "Child Id Request" from child (due to it containing incorrect challenge).
  - This whole process can repeat causing child not being able to attach (`troanj` test-case shows this behavior).


src/core/thread/mle.cpp
```diff
@@ -3282,9 +3282,14 @@ otError Mle::HandleParentResponse(const Message &aMessage, const Ip6::MessageInf
     }
 #endif
 
-    // if already have a candidate parent, only seek a better parent
-    if (mParentCandidate.GetState() == Neighbor::kStateParentResponse)
+    // Continue to process the "ParentResponse" if it is from current
+    // parent candidate to update the challenge and frame counters.
+
+    if ((mParentCandidate.GetState() == Neighbor::kStateParentResponse) &&
+        (mParentCandidate.GetRloc16() != sourceAddress.GetRloc16()))
```
jwhui: Should we match on extended address? RLOCs are only unique within a given partition.

PR #4118, opened 2019-08-20
[child-table] add support for more than 64 children

### LINKED ISSUE #3979: Configurable Number of Supported Children
When trying to simulate a POSIX network to hold the upper limit of child nodes, I hit an upper limit of 64 after which I observed an assert as below:
ot-cli-ftd: /home/work/src/openthread_190702/examples/../src/core/common/message.cpp:667: bool ot::Message::GetChildMask(uint8_t) const: Assertion `aChildIndex < sizeof(mBuffer.mHead.mInfo.mChildMask) * 8' failed.

It was clarified that currently the channel mask supports up to 64 children. To increase the limit, it was suggested that I should try to change 8 to a larger number in 
`https://github.com/openthread/openthread/blob/master/src/core/common/message.hpp#L96`

As per the specification the number of the supported end devices should be be up to 511. In order to support this, can the number of supported children be made configurable. 
[Original Discussion](https://groups.google.com/forum/#!topic/openthread-users/wpUPlhCEYlI)

### LINKED ISSUE #4100: Max allowed children
Hi all,

I'm trying to build a large scale Thread network which has about 500 children per router. According to the OpenThread webpage (https://openthread.io/guides/thread-primer/node-roles-and-types), within a single network, each router can have up to 511 children.

But the OpenThread implementation seems not to support that many children. When I checked the source code (include/openthread/thread_ftd.h), there are some related functions such as;

uint8_t otThreadGetMaxAllowedChildren(otInstance *aInstance);
otError otThreadSetMaxAllowedChildren(otInstance *aInstance, uint8_t aMaxChildren);

The issue is that these functions use uint8_t. I think to support more than 255 children it should use uint16_t.

Regards,
Seung

jwhui:
Fixes #3979 
Fixes #4100

### DIFF
--- a/src/core/thread/network_diagnostic.cpp
+++ b/src/core/thread/network_diagnostic.cpp
@@ -224,7 +224,7 @@ exit:
 otError NetworkDiagnostic::AppendChildTable(Message &aMessage)
 {
     otError         error   = OT_ERROR_NONE;
-    uint8_t         count   = 0;
+    uint16_t        count   = 0;
     uint8_t         timeout = 0;
     ChildTableTlv   tlv;
     ChildTableEntry entry;
@@ -232,7 +232,13 @@ otError NetworkDiagnostic::AppendChildTable(Message &aMessage)
     tlv.Init();
 
     count = Get<ChildTable>().GetNumChildren(ChildTable::kInStateValid);
-    tlv.SetLength(count * sizeof(ChildTableEntry));
+
+    if (count > (CHAR_BIT * sizeof(uint8_t) / sizeof(ChildTableEntry)))
+    {
+        count = CHAR_BIT * sizeof(uint8_t) / sizeof(ChildTableEntry);
+    }
+
+    tlv.SetLength(static_cast<uint8_t>(count * sizeof(ChildTableEntry)));
 
     SuccessOrExit(error = aMessage.Append(&tlv, sizeof(ChildTableTlv)));
 
--- a/examples/platforms/posix/radio.c
+++ b/examples/platforms/posix/radio.c
@@ -139,8 +139,8 @@ static bool     sPromiscuous = false;
 static bool     sTxWait      = false;
 static int8_t   sTxPower     = 0;
 
-static uint8_t      sShortAddressMatchTableCount = 0;
-static uint8_t      sExtAddressMatchTableCount   = 0;
+static uint16_t     sShortAddressMatchTableCount = 0;
+static uint16_t     sExtAddressMatchTableCount   = 0;
 static uint16_t     sShortAddressMatchTable[POSIX_MAX_SRC_MATCH_ENTRIES];
 static otExtAddress sExtAddressMatchTable[POSIX_MAX_SRC_MATCH_ENTRIES];
 static bool         sSrcMatchEnabled = false;
--- a/include/openthread/thread_ftd.h
+++ b/include/openthread/thread_ftd.h
@@ -103,7 +103,7 @@ typedef struct otEidCacheEntry
  * @sa otThreadSetMaxAllowedChildren
  *
  */
-uint8_t otThreadGetMaxAllowedChildren(otInstance *aInstance);
+uint16_t otThreadGetMaxAllowedChildren(otInstance *aInstance);
 
 /**
  * Set the maximum number of children currently allowed.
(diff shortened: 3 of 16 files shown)


jwhui:
@LuDuda, thanks for the review.

I tried simulating a 100 nodes, but the current simulator implementation runs **_really_** slow with 100 devices.

jsnayak:
@jwhui 
I was able to successfully create a network of 100+ nodes. 
Although when I merged this implementation with my implementation, I faced an error in src/core/address_resolver.cpp; I was able to resolve this by updating the data type in the address_resolver.cpp/address_resolver.hpp
Do you have any thoughts in this?
```
/home/pi/src/git/openthread_child_table/examples/../src/core/thread/address_resolver.cpp: In member function ‘void ot::AddressResolver::Clear()’:
/home/pi/src/git/openthread_child_table/examples/../src/core/thread/address_resolver.cpp:76:27: error: comparison is always true due to limited range of data type [-Werror=type-limits]
     for (uint8_t i = 0; i < kCacheEntries; i++)
                         ~~^~~~~~~~~~~~~~~
In file included from /home/pi/src/git/openthread_child_table/examples/../src/core/coap/coap_message.hpp:43:0,
                 from /home/pi/src/git/openthread_child_table/examples/../src/core/coap/coap.hpp:36,
                 from /home/pi/src/git/openthread_child_table/examples/../src/core/thread/address_resolver.hpp:39,
                 from /home/pi/src/git/openthread_child_table/examples/../src/core/thread/address_resolver.cpp:36:
/home/pi/src/git/openthread_child_table/examples/../src/core/thread/address_resolver.cpp: In member function ‘otError ot::AddressResolver::GetEntry(uint8_t, otEidCacheEntry&) const’:
/home/pi/src/git/openthread_child_table/examples/../src/core/thread/address_resolver.cpp:86:25: error: comparison is always true due to limited range of data type [-Werror=type-limits]
     VerifyOrExit(aIndex < kCacheEntries, error = OT_ERROR_INVALID_ARGS);
                  ~~~~~~~^~~
/home/pi/src/git/openthread_child_table/examples/../src/core/common/code_utils.hpp:111:15: note: in definition of macro ‘VerifyOrExit’
         if (!(aCondition))            \
               ^~~~~~~~~~
/home/pi/src/git/openthread_child_table/examples/../src/core/thread/address_resolver.cpp: In member function ‘void ot::AddressResolver::InvalidateCacheEntry(ot::AddressResolver::Cache&, ot::AddressResolver::InvalidationReason)’:
/home/pi/src/git/openthread_child_table/examples/../src/core/thread/address_resolver.cpp:211:35: error: large integer implicitly truncated to unsigned type [-Werror=overflow]
     aEntry.mAge   = kCacheEntries - 1;
                     ~~~~~~~~~~~~~~^~~
```

src/core/thread/network_diagnostic.cpp
```diff
@@ -224,15 +224,21 @@ otError NetworkDiagnostic::AppendIp6AddressList(Message &aMessage)
...
     uint8_t         timeout = 0;
     ChildTableTlv   tlv;
     ChildTableEntry entry;
 
     tlv.Init();
 
     count = Get<ChildTable>().GetNumChildren(ChildTable::kInStateValid);
-    tlv.SetLength(count * sizeof(ChildTableEntry));
+
+    if (count > (CHAR_BIT * sizeof(uint8_t) / sizeof(ChildTableEntry)))
```
LuDuda: @jwhui I believe this part of code produces incorrect TLV that affects Router 5.7.3 certification test case (TLV length different than payload size).

What was the purpose of this if statement? During review I thought the idea was to limit the TLV length to 255, but this code actually does not do it.

jwhui: @LuDuda, thanks for raising this issue. I'm not sure what's going on here :)

Would you like to submit a PR with the fix?

LuDuda: I see a couple of options:

a) make sure that TLV length never exceeds 255 (limits child table TLV to ~85 entries)
b) introduce extended TLV when needed (limits child table TLV to network MTU and so ~400 entries but hard to tell what is the exact limit)
c) implement a) or b) and file Thread JIRA on how to handle situation where there is e.g. 500 children attached (spec requires to append information about all the children).

What do you think about option c)?

abtink: Thanks @LuDuda for bringing this up.  I'd vote for doing (a) as first step, followed by (c). 

Thinking about this, we may face other similar situations which we need to address. For example running out of the spinel NCP buffer size when encoding the child table for `SPINEL_THREAD_CHILD_TABLE` property - maybe need to think of a solution to send partial table?.

jwhui: Agree with option (c) - implement (a) and file Thread Group JIRA for additional clarify.

PR #4119, opened 2019-08-21
[commissioner] make start fail early if device role is invalid or detached

simonlingoogle:
The commit fixed an issue that running "commissioner start" command when device role is detached cause commissioner to enter an invalid state.


src/core/meshcop/commissioner.cpp
```diff
@@ -144,6 +144,8 @@ otError Commissioner::Start(otCommissionerStateCallback  aStateCallback,
 {
     otError error = OT_ERROR_NONE;
 
+    otDeviceRole deviceRole = Get<Mle::MleRouter>().GetRole();
```
bukepo: use `Get<Mle::MleRouter>().IsAttached()` instread?

PR #4120, opened 2019-08-21
[commissioner] only allow attached device to become commissioner

simonlingoogle:
The PR fixes the issue that running "commissioner start" command when device role is detached will cause commissioner not be able to start or stop.

### DIFF
--- a/src/core/meshcop/commissioner.cpp
+++ b/src/core/meshcop/commissioner.cpp
@@ -144,6 +144,9 @@ otError Commissioner::Start(otCommissionerStateCallback  aStateCallback,
 {
     otError error = OT_ERROR_NONE;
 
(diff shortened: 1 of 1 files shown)


src/core/meshcop/commissioner.cpp
```diff
@@ -144,6 +144,9 @@ otError Commissioner::Start(otCommissionerStateCallback  aStateCallback,
 {
     otError error = OT_ERROR_NONE;
 
+    otDeviceRole deviceRole = Get<Mle::MleRouter>().GetRole();
```
bukepo: use `Get<Mle::MleRouter>().IsAttached()` instead?

simonlingoogle: modified

PR #4122, opened 2019-08-21
[nrf52xxx] update radio driver to version 1.5.0

hubertmis:


third_party/NordicSemiconductor/drivers/radio/platform/clock/nrf_802154_clock_sdk.c
```diff
@@ -65,12 +65,12 @@ static void clock_handler(nrf_drv_clock_evt_type_t event)
 
 void nrf_802154_clock_init(void)
 {
-    // Intentionally empty.
+    nrf_drv_clock_init();
```
LuDuda: This has to be reverted, same for uninit.

PR #4126, opened 2019-08-27
[posix] do not exit upon recv getting EINTR

gjc13:


examples/platforms/posix/radio.c
```diff
@@ -815,15 +817,17 @@ void platformRadioProcess(otInstance *aInstance, const fd_set *aReadFdSet, const
     {
         ssize_t rval = recvfrom(sRxFd, (char *)&sReceiveMessage, sizeof(sReceiveMessage), 0, NULL, NULL);
 
-        if (rval < 0)
+        if (rval > 0)
+        {
+            sReceiveFrame.mLength = (uint8_t)(rval - 1);
+
+            radioReceive(aInstance);
+        }
```
bukepo:
```cpp
else if (rval == 0)
{
    // socket is closed, which should not happen
    assert(false);
}
else if if (errno != EINTR && errno != EAGAIN)
{
    ....
}
```

PR #4127, opened 2019-08-27
[api] force linker failures when using disabled features

bukepo:
This PR helps misconfigurations and invalid usages be found earlier.

* Removed dummy implementations.
* Added assert NULL pointers.
* Changed return type to void for APIs always succeed.

### DIFF
--- a/include/openthread/channel_manager.h
+++ b/include/openthread/channel_manager.h
@@ -128,7 +128,6 @@ otError otChannelManagerSetDelay(otInstance *aInstance, uint16_t aDelay);
  * @retval OT_ERROR_NONE               Channel selection finished successfully.
  * @retval OT_ERROR_NOT_FOUND          Supported channel mask is empty, therefore could not select a channel.
  * @retval OT_ERROR_INVALID_STATE      Thread is not enabled or not enough data to select a new channel.
- * @retval OT_ERROR_DISABLED_FEATURE   `ChannelMonitor` feature is disabled by build-time configuration options.
  *
  */
 otError otChannelManagerRequestChannelSelect(otInstance *aInstance, bool aSkipQualityCheck);
--- a/include/openthread/channel_manager.h
+++ b/include/openthread/channel_manager.h
@@ -127,7 +127,6 @@ otError otChannelManagerSetDelay(otInstance *aInstance, uint16_t aDelay);
  *
  * @retval OT_ERROR_NONE               Channel selection finished successfully.
  * @retval OT_ERROR_NOT_FOUND          Supported channel mask is empty, therefore could not select a channel.
- * @retval OT_ERROR_INVALID_STATE      Thread is not enabled or not enough data to select a new channel.
  *
  */
 otError otChannelManagerRequestChannelSelect(otInstance *aInstance, bool aSkipQualityCheck);
--- a/include/openthread/coap_secure.h
+++ b/include/openthread/coap_secure.h
@@ -150,23 +150,20 @@ void otCoapSecureSetSslAuthMode(otInstance *aInstance, bool aVerifyPeerCertifica
  * This method sets the local device's X509 certificate with corresponding private key for
  * DTLS session with DTLS_ECDHE_ECDSA_WITH_AES_128_CCM_8.
  *
+ * @note This function requires `MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED=1`.
+ *
  * @param[in]  aInstance          A pointer to an OpenThread instance.
  * @param[in]  aX509Cert          A pointer to the PEM formatted X509 certificate.
  * @param[in]  aX509Length        The length of certificate.
  * @param[in]  aPrivateKey        A pointer to the PEM formatted private key.
  * @param[in]  aPrivateKeyLength  The length of the private key.
  *
- * @retval OT_ERROR_NONE              Successfully set the x509 certificate
- *                                    with his private key.
- * @retval OT_ERROR_DISABLED_FEATURE  Mbedtls config not enabled
- *                                    MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED.
- *
  */
-otError otCoapSecureSetCertificate(otInstance *   aInstance,
-                                   const uint8_t *aX509Cert,
-                                   uint32_t       aX509Length,
-                                   const uint8_t *aPrivateKey,
-                                   uint32_t       aPrivateKeyLength);
+void otCoapSecureSetCertificate(otInstance *   aInstance,
+                                const uint8_t *aX509Cert,
+                                uint32_t       aX509Length,
+                                const uint8_t *aPrivateKey,
+                                uint32_t       aPrivateKeyLength);
 
 /**
  * This method sets the trusted top level CAs. It is needed for validating the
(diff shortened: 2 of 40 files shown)


include/openthread/joiner.h
```diff
@@ -111,11 +111,8 @@ otError otJoinerStart(otInstance *     aInstance,
  *
  * @param[in]  aInstance  A pointer to an OpenThread instance.
  *
- * @retval OT_ERROR_NONE              Successfully disabled the Joiner role.
- * @retval OT_ERROR_DISABLED_FEATURE  The Joiner feature is not enabled in this build.
- *
  */
-otError otJoinerStop(otInstance *aInstance);
+void otJoinerStop(otInstance *aInstance);
```
gjc13: Shall we keep the return type untouched so that `SuccessOrExit(otJoinerStop(...));` pattern won't cause a compile error?

bukepo: Good question! @abtink @jwhui Thoughts?

librasungirl: Here just propose another topic might be a little broader.
Currently it seems there are different strategies regarding the APIs specific to the configurable options, for example
otJoinerXXX() APIs would have empty body when the option is not enabled, while 
all otServerXXX() APIs are just wrapped within the option macro 
https://github.com/openthread/openthread/blob/master/src/core/api/server_api.cpp#L36

Are there any specific rules to follow? @jwhui @bukepo

src/core/api/border_router_api.cpp
```diff
@@ -37,29 +37,28 @@
...
 
 using namespace ot;
 
 otError otBorderRouterGetNetData(otInstance *aInstance, bool aStable, uint8_t *aData, uint8_t *aDataLength)
 {
-    otError   error    = OT_ERROR_NONE;
     Instance &instance = *static_cast<Instance *>(aInstance);
 
-    VerifyOrExit(aData != NULL && aDataLength != NULL, error = OT_ERROR_INVALID_ARGS);
+    assert(aData != NULL && aDataLength != NULL);
```
librasungirl: add note in header file to clarify the pointers are not allowed to be NULL

src/core/thread/key_manager.cpp
```diff
@@ -129,10 +129,10 @@ otError KeyManager::SetMasterKey(const otMasterKey &aKey)
     ComputeKey(mKeySequence, mKey);
 
     // reset parent frame counters
-    routers = Get<Mle::MleRouter>().GetParent();
-    routers->SetKeySequence(0);
-    routers->SetLinkFrameCounter(0);
-    routers->SetMleFrameCounter(0);
+    parent = Get<Mle::MleRouter>().GetParent();
```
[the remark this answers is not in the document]
librasungirl: There was some discussion regarding this. Get the attaching parent is only required for reference device for certification. Sounds good to me to return null for non-child device if `REFERENCE_DEVICE` is not enabled.

src/core/api/border_router_api.cpp
bukepo: Then we must clarify for almost every API, `aInstance` is not allowed to be NULL. Thoughts?

bukepo:
@librasungirl Considered again, I think the `GetParent()` is another topic. I think we may enhance in separate PR.

src/core/api/coap_secure_api.cpp
```diff
@@ -52,98 +52,58 @@ otError otCoapSecureStart(otInstance *aInstance, uint16_t aPort)
...
 
     return instance.GetApplicationCoapSecure().SetCertificate(aX509Cert, aX509Length, aPrivateKey, aPrivateKeyLength);
-#else
-    OT_UNUSED_VARIABLE(aInstance);
-    OT_UNUSED_VARIABLE(aX509Cert);
-    OT_UNUSED_VARIABLE(aX509Length);
-    OT_UNUSED_VARIABLE(aPrivateKey);
-    OT_UNUSED_VARIABLE(aPrivateKeyLength);
-
-    return OT_ERROR_DISABLED_FEATURE;
```
zhanglongxia: Should we remove the return value `OT_ERROR_DISABLED_FEATURE` from the function comments?

src/core/api/coap_secure_api.cpp
```diff
@@ -52,98 +52,58 @@ otError otCoapSecureStart(otInstance *aInstance, uint16_t aPort)
...
-#endif // MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED
 }
 
 otError otCoapSecureSetCaCertificateChain(otInstance *   aInstance,
```
zhanglongxia: The only return value of this function  is `OT_ERROR_NONE` in the function comments, should we set the type of return value to `void`?

src/core/api/coap_secure_api.cpp
```diff
@@ -52,98 +52,58 @@ otError otCoapSecureStart(otInstance *aInstance, uint16_t aPort)
...
-    OT_UNUSED_VARIABLE(aX509CaCertificateChain);
-    OT_UNUSED_VARIABLE(aX509CaCertChainLength);
-
-    return OT_ERROR_DISABLED_FEATURE;
-#endif // MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED
 }
+#endif // MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED
 
+#ifdef MBEDTLS_KEY_EXCHANGE_PSK_ENABLED
 otError otCoapSecureSetPsk(otInstance *   aInstance,
```
zhanglongxia: Should we remove the return value `OT_ERROR_DISABLED_FEATURE` and `OT_ERROR_INVALID_ARGS ` from the function comments?

bukepo:
@abtink Totally agree with you! I will add the required feature in the API doc string.

include/openthread/logging.h
```diff
@@ -66,6 +66,8 @@ otLogLevel otLoggingGetLevel(void);
 /**
  * This function sets the log level.
  *
+ * @note This function requires `OPENTHREAD_CONFIG_ENABLE_DYNAMIC_LOG_LEVEL=1` to be enabled.
```
librasungirl: seems a little redundant, remove `to be enabled`?

bukepo:
@abtink do you have any more comments?

include/openthread/channel_manager.h
```diff
@@ -128,7 +128,6 @@ otError otChannelManagerSetDelay(otInstance *aInstance, uint16_t aDelay);
  * @retval OT_ERROR_NONE               Channel selection finished successfully.
  * @retval OT_ERROR_NOT_FOUND          Supported channel mask is empty, therefore could not select a channel.
  * @retval OT_ERROR_INVALID_STATE      Thread is not enabled or not enough data to select a new channel.
- * @retval OT_ERROR_DISABLED_FEATURE   `ChannelMonitor` feature is disabled by build-time configuration options.
```
jwhui: Should we remove `OT_ERROR_DISABLED_FEATURE` from `include/openthread/error.h`?

include/openthread/coap_secure.h
```diff
@@ -32,6 +32,9 @@
  *  This file defines the top-level functions for the OpenThread CoAP Secure implementation.
  *
  *  @note
+ *   The functions in this module requires the build-time feature `OPENTHREAD_CONFIG_COAP_SECURE_API_ENABLE=1`.
```
jwhui: nit: "requires" -> "require"

include/openthread/thread.h
```diff
@@ -46,6 +46,9 @@ extern "C" {
 /**
  * @addtogroup api-thread-general
  *
+ * @note
+ *   The functions in this module requires `OPENTHREAD_FTD=1` or `OPENTHREAD_MTD=1`.
```
jwhui: nit: "requires" -> "require"

include/openthread/thread_ftd.h
```diff
@@ -232,7 +232,7 @@ otError otThreadSetJoinerUdpPort(otInstance *aInstance, uint16_t aJoinerUdpPort)
  * Set Steering data out of band.
  *
  * Configuration option `OPENTHREAD_CONFIG_MLE_STEERING_DATA_SET_OOB_ENABLE` should be set to enable setting of steering
- * data out of band. Otherwise calling this function does nothing and it returns `OT_ERROR_DISABLED_FEATURE`
+ * data out of band.
  * error.
```
jwhui: Remove "error."

src/core/api/commissioner_api.cpp
```diff
@@ -40,14 +40,14 @@
 
 using namespace ot;
 
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_COMMISSIONER_ENABLE
```
jwhui: I think it would be simpler to have `OPENTHREAD_CONFIG_COMMISSIONER_ENABLE` here and check for `OPENTHREAD_FTD` in `openthread-core-config-check.h`. Thoughts?

bukepo: I think when COMMISSIONER=1 and --enable-ftd and --enable-mtd, both openthread_ftd.a and openthread_mtd.a will be compiled with OPENTHREAD_CONFIG_COMMISSIONER_ENABLE=1.

PR #4129, opened 2019-08-27
[mle] reset MLE adv timer when cost changes to/from infinite

simonlingoogle:
The test case 8.2.1 in thread certification is likely to fail with SiLab DUT JR on OT TB. The OT joiner might fail in the DTLS handshake process after it resend the `ClientHello(with cookie)` for multiple times. We found that the OT Commissioner lost its route to neighbor JR and failed to send `Server Hello, Server Key Exchange, Server Hello Done` in these failed test cases. 

OT Commissioner lost its route because it have received `MLE advertisement` from JR and updated its route accordingly. However, the link between Commissioner and JR is not restored fast enough which can cause messages to be dropped. 

This PR resets the `MLE advertisement` when the link cost to a neighbor is changed to or from infinite, so that link between Commissioner and JR is restored ASAP. Together with [PR 4130](https://github.com/openthread/openthread/pull/4130), the fail chance reduced from 44% to 0% for the 8.2.1 test case. 

This fix is also in line with the Spec. "5.9.7 Sending Advertisements" as shown bellow. 
> The Trickle timer is reset when the reported routing cost to one or more destinations changes to or from infinite (represented as value 0).

The PR also makes sure there is at most one call to `ResetAdvertiseInterval` in the `MleRouter::UpdateRoutes` method.

The detailed process causing this issue is stated as bellow:
- Joiner sends `ClientHello+cookie`
- JR forwards `ClientHello+cookie` to Commissioner, and Commissioner start processing the message (takes long time)
- Joiner resends `ClientHello+cookie` after 8s timeout
- JR loses its link to Commissioner because of transmission failure when forwarding `ClientHello+cookie`
- Commissioner update out link quality of JR to 0 (infinite) according to JR's `MLE advertisement`
- Commissioner is trying to send `ServerHello, ...` but think it had no route to JR, thus drops the message (multiple times)
- Joiner fails to receive `ServerHello, ...` for a long time and sends handshake failure

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -1427,7 +1427,8 @@ void MleRouter::UpdateRoutes(const RouteTlv &aRoute, uint8_t aRouterId)
     uint8_t cost;
     uint8_t linkQuality;
     bool    update;
-    bool    changed = false;
+    bool    resetAdvInterval = false;
+    bool    changed          = false;
 
     neighbor = mRouterTable.GetRouter(aRouterId);
     VerifyOrExit(neighbor != NULL);
@@ -1461,7 +1462,16 @@ void MleRouter::UpdateRoutes(const RouteTlv &aRoute, uint8_t aRouterId)
 
                 if (neighbor->GetLinkQualityOut() != linkQuality)
                 {
+                    uint8_t oldLinkCost = mRouterTable.GetLinkCost(*neighbor);
                     neighbor->SetLinkQualityOut(linkQuality);
+                    nextHop = mRouterTable.GetRouter(neighbor->GetNextHop());
+
+                    // reset MLE advertisement timer if neighbor route cost changed to or from infinite
+                    if (nextHop == NULL &&
+                        (oldLinkCost >= kMaxRouteCost) != (mRouterTable.GetLinkCost(*neighbor) >= kMaxRouteCost))
+                    {
+                        resetAdvInterval = true;
+                    }
                     update = true;
                 }
             }
@@ -1494,7 +1504,7 @@ void MleRouter::UpdateRoutes(const RouteTlv &aRoute, uint8_t aRouterId)
                         {
                             if (nextHop == NULL && mRouterTable.GetLinkCost(*router) >= kMaxRouteCost)
                             {
-                                ResetAdvertiseInterval();
+                                resetAdvInterval = true;
                             }
 
                             router->SetNextHop(aRouterId);
@@ -1505,7 +1515,7 @@ void MleRouter::UpdateRoutes(const RouteTlv &aRoute, uint8_t aRouterId)
                         {
                             if (mRouterTable.GetLinkCost(*router) >= kMaxRouteCost)
                             {
-                                ResetAdvertiseInterval();
+                                resetAdvInterval = true;
                             }
 
                             router->SetNextHop(kInvalidRouterId);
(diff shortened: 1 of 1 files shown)


src/core/thread/mle_router.cpp
```diff
@@ -1427,7 +1427,8 @@ void MleRouter::UpdateRoutes(const RouteTlv &aRoute, uint8_t aRouterId)
     uint8_t cost;
     uint8_t linkQuality;
     bool    update;
-    bool    changed = false;
+    bool    requireResetAdv = false;
```
jwhui: nit: how about rename to something like `resetAdvInterval` (we are resetting the interval, not the advertisement itself).

src/core/thread/mle_router.cpp
```diff
@@ -1461,7 +1462,14 @@ void MleRouter::UpdateRoutes(const RouteTlv &aRoute, uint8_t aRouterId)
 
                 if (neighbor->GetLinkQualityOut() != linkQuality)
                 {
+                    uint8_t oldRouteCost = mRouterTable.GetLinkCost(*neighbor);
                     neighbor->SetLinkQualityOut(linkQuality);
+
+                    // reset MLE advertisement timer if neighbor link cost changed to or from infinite
+                    if ((oldRouteCost >= kMaxRouteCost) != (mRouterTable.GetLinkCost(*neighbor) >= kMaxRouteCost))
```
jwhui: I think this should consider whether a route to the neighbor becomes available or unavailable. Note that it is possible to have a route even if a link has not been established. Rather than simply comparing link cost, we should also see if the route utilizes this link. Thoughts?

librasungirl: Agree completely 👍

simonlingoogle: Agree. I will update the code accordingly.

simonlingoogle: I have updated the code:
- when updating neighbor link quality out, reset adv interval only when there is no multihop route to the neighbor (i.e. `mRouterTable.GetRouter(neighbor->GetNextHop()) == NULL`)
- modify `if (cost + mRouterTable.GetLinkCost(*neighbor) <= kMaxRouteCost)` to `if (cost + mRouterTable.GetLinkCost(*neighbor) < kMaxRouteCost)` . 

Per my standing, the neighbor is used as route nexthop only when `route cost < kMaxRouteCost`, and many code in `UpdateRoutes` rely on this restriction. Please verify if there is a bug here.

src/core/thread/mle_router.cpp
```diff
@@ -1461,7 +1462,16 @@ void MleRouter::UpdateRoutes(const RouteTlv &aRoute, uint8_t aRouterId)
 
                 if (neighbor->GetLinkQualityOut() != linkQuality)
                 {
+                    uint8_t oldLinkCost = mRouterTable.GetLinkCost(*neighbor);
                     neighbor->SetLinkQualityOut(linkQuality);
+                    nextHop = mRouterTable.GetRouter(neighbor->GetNextHop());
+
+                    // reset MLE advertisement timer if neighbor route cost changed to or from infinite
```
librasungirl: update the comment here accordingly as well (for better readability in the future)?

simonlingoogle: The comment already said "route cost" which is accurate. It considers both multihop routing cost and direct link cost.

PR #4132, opened 2019-08-28
[mac] move static variable definition into Mac class

abtink:
----
Reason for this the change:
- Namespace/scope variables under `Mac` class only.
- Allow us in future to potentially break `mac.cpp` into multiple files (e.g. `mac_ftd`?).


abtink:
Travis test link: https://travis-ci.com/openthread/ot-nest-nordic/builds/125168328

PR #4136, opened 2019-08-30
 [data-poll-sender] always honor the default child timeout based poll interval

abtink:
This PR contains three related changes/commits.

It addresses an issue with child timing out on parent (kicked out of child table) if the data poll period is set (by user) to a value larger than the child's timeout.

    [network-diagnostics] use the child timeout directly
    
    This commit changes the `NetworkDiagnostic::FillRequestedTlvs()` to
    directly use the child timeout value in appended `TimeoutTlv`. This
    addresses an issue where a user-specified long data poll interval
    could cause the timeout TLV in diags response to be set higher than
    actual child's timeout value.

---------------

    [data-poll-sender] always honor the default child timeout based poll interval
    
    This commit changes the `DataPollSender` to ensure that a child device
    will always send a data poll within its timeout interval. This
    addresses an issue where a large user-specified poll period could
    cause the device to adopt a poll period longer than the default child
    timeout (causing parent to kick the child out of its child table).
    
    This commit also simplifies the selection code for a new poll period
    by adding a helper `UpdateIfLarger()` method to update the period only
    if it's larger than a given new value.

---------

    [toranj] add test to verify child sending data poll within its timeout
    
    This commit adds a `toranj` test-case. This test covers the situation
    where the data poll interval is set to a value larger than child's
    timeout interval. The test verifies that the sleepy child would send a
    data poll within its timeout interval (ensuring that child stays in
    parent's child table).

The new test does indeed fail without the change/fix from the second commit.

### DIFF
--- a/src/core/mac/data_poll_sender.hpp
+++ b/src/core/mac/data_poll_sender.hpp
@@ -105,15 +105,16 @@ public:
     /**
      * This method sets/clears a user-specified/external data poll period.
      *
+     * Value of zero for `aPeriod` clears the user-specified poll period.
+     *
      * If the user provides a non-zero poll period, the user value specifies the maximum period between data
      * request transmissions. Note that OpenThread may send data request transmissions more frequently when expecting
-     * a control-message from a parent or in case of data poll transmission failures or timeouts.
+     * a control-message from a parent or in case of data poll transmission failures or timeouts, or when the specified
+     * value is larger than the child timeout.
      *
-     * Minimal non-zero value should be `OPENTHREAD_CONFIG_MAC_MINIMUM_POLL_PERIOD` (10ms). Or zero to clear
-     * user-specified poll period.
-     *
-     * User-specified value should be no more than the maximal value 0x3FFFFFF ((1 << 26) - 1) allowed, otherwise it
-     * would be cilpped by the maximal value.
+     * A non-zero `aPeriod` should be larger than or equal to `OPENTHREAD_CONFIG_MAC_MINIMUM_POLL_PERIOD` (10ms) or
+     * this method returns `OT_ERROR_INVALID_ARGS`. If a non-zero `aPeriod` is larger than maximum value of
+     * `0x3FFFFFF ((1 << 26) - 1)`, it would be clipped to this value.
      *
      * @param[in]  aPeriod  The data poll period in milliseconds.
      *
@@ -222,10 +223,24 @@ public:
     /**
      * This method gets the maximum data polling period in use.
      *
-     * @returns the maximum data polling period in use.
+     * The maximum data poll period is determined based as the minimum of the user-specified poll interval and the
+     * default poll interval.
+     *
+     * @returns The maximum data polling period in use.
      *
      */
     uint32_t GetKeepAlivePollPeriod(void) const;
+
+    /**
+     * This method returns the default maximum poll period.
+     *
+     * The default poll period is determined based on the child timeout interval, ensuing the child would send data poll
+     * within the child's timeout.
+     *
+     * @returns The maximum default data polling interval (in msec).
+     *
+     */
+    uint32_t GetDefaultPollPeriod(void) const;
 
 private:
     enum // Poll period under different conditions (in milliseconds).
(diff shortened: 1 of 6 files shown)


abtink:
In  the new pushed commits, added PR number `(#4136)` to commit message title.

librasungirl:
@abtink , sorry for late response.

For the first commit, I am afraid it might be conflict with the Specification.

> 1.1 Spec 10.11.3
> Timeout For SEDs, indicates the maximum polling time period.

For the issue the first commit to resolve, I don't think it is valid to allow user to set a poll period larger than child's timeout. 

> Spec 1.1 Section 4.7.5
> If a Child does not receive a response to a keep-alive message from their Parent, the Child
> should retry three times before attempting to reattach to a new Parent. All four attempts
> MUST be completed within the child's timeout period.

How about adding the check when setting the poll period?

src/core/mac/data_poll_sender.hpp
```diff
@@ -105,15 +105,16 @@ class DataPollSender : public InstanceLocator
     /**
      * This method sets/clears a user-specified/external data poll period.
      *
+     * Value of  zero for `aPeriod` clears the user-specified poll period.
```
librasungirl: just notice that there are two spaces before `zero`

src/core/mac/data_poll_sender.hpp
```diff
@@ -222,11 +223,25 @@ class DataPollSender : public InstanceLocator
     /**
      * This method gets the maximum data polling period in use.
      *
-     * @returns the maximum data polling period in use.
+     * The maximum data poll period is determined based as the minimum of the user-specified poll interval  and the
```
librasungirl: nit: one extra space before `and`

abtink:
Thanks @librasungirl for helping review the PR. Addressed the typos from your comment in new pushed commit.

On first point, @jwhui  and I both read the spec before the change. The interpretation is that the value in this TLV should be the maximum polling interval device will always honor (i.e. we always have at least one poll within this interval). The child timeout is indeed the maximum poll interval that we always guarantee (independent of user's specified poll interval). The user may request different poll intervals and may change the value often, so it's not always ensured to remain the same. 

On second point, I think we should allow user to set the external poll interval to any value (even larger than child timeout): 
- Changing the API to return an error on value larger than timeout would directly impact how we use it in some of our internal projects:
  - The next layers (on some internal projects) do set the poll interval (user-specified poll interval) to some large value and instead the next layer directly controls when/how to send polls using `otLinkSendDataRequest()` API.
  - This is what exposed this issue being addressed here, i.e., if the next layer does not send polls frequently enough, OT core should still ensure that it honors the child timeout interval (send at least one poll within the child timeout).
- The fact that we need to poll at minimum within a child timeout's intevral, is a Thread requirement (which we ensure in core code implementation). The goal is for user to determine its poll interval from app layer requirements and would be easier to not have user be aware of child timeout.

Thoughts?

librasungirl:
@abtink Thanks for the explanation. Agree that it would not be a good idea to reject the larger value in the context. I vote for 'honor' solution here. Basically I am fine with this PR.  Here are some thoughts just for discussion:

Section 4.7.5 should indicate that in worst case four keep-alive messages(i.e. data poll for SED) should happen before `ChildTimeout` - that is, maximum poll interval should be the value of existing `DataPollSender::GetDefaultPollPeriod`, not `ChildTimeout`.

And Network Diagnostics should be designed to retrieve the state of the device or the network. The Timeout TLV in Network Diagnostic (maximum of poll interval, for SED) should be used to check the possible waiting time to deliver a message to a SED. I think it may make better sense to still fill TimeoutTLV with the real maximum poll period?

Thoughts?

jwhui:
Section 10.11.3 also states:
```
Same format and semantics as Timeout TLV (MLE TLV Type 2, Section 4.5.3).
```

Section 4.5.3 states:
```
The Timeout TLV (TLV Type 2) has a Value containing a 32-bit unsigned integer. The value
is the expected maximum interval between transmissions by the sender, in seconds. This
allows the receiver to more accurately time out a link to a neighbor that polls for its
incoming messages.
```

What is communicated in the MLE Timeout TLV is the child timeout value. So I think it makes sense that we use the child timeout value for the Network Diagnostics TLV as well.

Thoughts?

PR #4137, opened 2019-09-03
[cc2538] interrupt driven radio driver

sjlongland:
So, I have a confession to make… in PR#3609 I accidentally included some code that would have enabled the receive interrupt on the CC2538 in the NVIC.

Given the receive interrupt handler just blindly clears the interrupt flags, this is probably not a good plan.  So, a better option, implement the rest of it, with a compile-time switch to turn interrupt driven reception off if desired.  This allows for up to two frames to be stored until such time as they can be processed by OpenThread (one frame in the radio's buffer, and one in `sReceivedFrame`).

I'm doing some more thorough tests with the GRL test suite to ensure it hasn't broken anything but indications are it is working as expected (I'd be expecting failures left-right and centre if receive was broken by this).

### DIFF
--- a/examples/platforms/cc2538/radio.c
+++ b/examples/platforms/cc2538/radio.c
@@ -41,8 +41,6 @@
 #include "common/logging.hpp"
 #include "utils/code_utils.h"
 
-#define RFCORE_RXTX_INT (141)
-
 #define RFCORE_XREG_RFIRQM0 0x4008868C // RF interrupt masks
 #define RFCORE_XREG_RFIRQM1 0x40088690 // RF interrupt masks
 #define RFCORE_XREG_RFERRM 0x40088694  // RF error interrupt mask
--- a/examples/platforms/cc2538/radio.c
+++ b/examples/platforms/cc2538/radio.c
@@ -661,7 +661,7 @@ void otPlatRadioSetPromiscuous(otInstance *aInstance, bool aEnable)
     }
 }
 
-void readFrame(otInstance *aInstance)
+static void readFrame(void)
 {
     uint8_t length;
     uint8_t crcCorr;
@@ -678,7 +678,7 @@ void readFrame(otInstance *aInstance)
 #error Time sync requires the timestamp of SFD rather than that of rx done!
 #else
     // Timestamp
-    if (otPlatRadioGetPromiscuous(aInstance))
+    if (cc2538RadioGetPromiscuous())
 #endif
     {
         // The current driver only supports milliseconds resolution.
--- a/examples/platforms/cc2538/radio.c
+++ b/examples/platforms/cc2538/radio.c
@@ -743,11 +743,7 @@ void cc2538RadioProcess(otInstance *aInstance)
     HWREG(RFCORE_XREG_RFIRQM0) &= ~RFCORE_XREG_RFIRQM0_RXPKTDONE;
 #endif
 
-    if (sReceiveFrame.mLength == 0)
-    {
-        // sReceiveFrame is empty, so check for any received data.
-        readFrame();
-    }
+    readFrame();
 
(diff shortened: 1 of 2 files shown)


jwhui:
@sjlongland, thank you for this contribution and the additional testing! Let us know when you've completed your testing and think this PR can move forward.

sjlongland:
Two mistakes found:

1. I got the wrong register for the interrupt vector in the NVIC… so the interrupt service routine wasn't firing (confusingly, TI have _two_ interrupts mapped for RX/TX, and I'm not sure which one is being used)
2. In the interrupt service routine, I'm masking the constant `RFCORE_SFR_RFIRQF0` with another constant `RFCORE_SFR_RFIRQF0_RXPKTDONE` which basically means it's a no-op anyway.  `RFCORE_SFR_RFIRQF0` needs a `HWREG` macro wrapped around it.

I'll fix both of these problems. :-)

sjlongland:
Gah… as Maxwell Smart would put it… missed it by `→ ←` that much.  `make pretty` correction pushed.

sjlongland:
cc2538-rx-int-processing

If I surround parts of the radio driver, and comment out the call to `readFrame` from `otPlatRadioProcess` so that the only received traffic is via the interrupt, I get the above traces on a logic analyser.  In the above case:

* `PB7` was turned on just before [line 829](https://github.com/openthread/openthread/blob/4794c3fc3b23b43930a41ee89716daea841c7697/examples/platforms/cc2538/radio.c#L829) and off after [line 844](https://github.com/openthread/openthread/blob/4794c3fc3b23b43930a41ee89716daea841c7697/examples/platforms/cc2538/radio.c#L844).
* `PB2` was turned on before [line 740](https://github.com/openthread/openthread/blob/4794c3fc3b23b43930a41ee89716daea841c7697/examples/platforms/cc2538/radio.c#L740) and off after [line 824](https://github.com/openthread/openthread/blob/4794c3fc3b23b43930a41ee89716daea841c7697/examples/platforms/cc2538/radio.c#L824)
* `PB6` was turned on at [line 760](https://github.com/openthread/openthread/blob/4794c3fc3b23b43930a41ee89716daea841c7697/examples/platforms/cc2538/radio.c#L760) and off at [line 784](https://github.com/openthread/openthread/blob/4794c3fc3b23b43930a41ee89716daea841c7697/examples/platforms/cc2538/radio.c#L784)
* `PA7` was turned on at [line 786](https://github.com/openthread/openthread/blob/4794c3fc3b23b43930a41ee89716daea841c7697/examples/platforms/cc2538/radio.c#L786) and off after [line 817](https://github.com/openthread/openthread/blob/4794c3fc3b23b43930a41ee89716daea841c7697/examples/platforms/cc2538/radio.c#L817)

jwhui:
@sjlongland, thanks for the updates and thorough testing. Is this PR ready to move forward?

sjlongland:
I think it should be safe… receiving frames seems to be working just fine with it turned on and it seems to have a big (positive) impact on some test cases in the GRL test harness.  If we find issues, we can always turn the feature off in the OpenThread config.

examples/platforms/cc2538/radio.c
```diff
@@ -649,12 +661,21 @@ void otPlatRadioSetPromiscuous(otInstance *aInstance, bool aEnable)
...
     uint8_t crcCorr;
     int     i;
 
+    if (sReceiveFrame.mLength != 0)
+    {
+        /*
+         * There is already a frame present in the buffer, return early so
+         * we do not overwrite it (hopefully we'll catch it on the next run).
+         */
+        return;
```
jwhui: style: we prefer to have a single `return` statement. Can we use `otEXPECT()` here instead?

PR #4140, opened 2019-09-04
[crypto] don't add entropy source if it's done by mbedTLS

KaSroka:
If platform uses MbedTLS feature to add entropy source during context init (by platform entropy, HAVEGE or `MBEDTLS_ENTROPY_HARDWARE_ALT`) we shouldn't add second entropy source. This saves time and power as we don't have to poll two different entropy sources/poll same entropy source twice.

### DIFF
--- a/src/core/common/random_manager.hpp
+++ b/src/core/common/random_manager.hpp
@@ -45,6 +45,11 @@
 
 #include "utils/wrap_stdint.h"
 
+#if (!defined(MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES) && \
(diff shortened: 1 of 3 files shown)


src/core/common/random_manager.hpp
```diff
@@ -45,6 +45,11 @@
 
 #include "utils/wrap_stdint.h"
 
+#if (!defined(MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES) && \
+     (!defined(MBEDTLS_NO_PLATFORM_ENTROPY) || defined(MBEDTLS_HAVEGE_C) || defined(MBEDTLS_ENTROPY_HARDWARE_ALT)))
+#define MBEDTLS_STRONG_DEFAULT_ENTROPY_PRESENT
```
jwhui: This define appears to be specific to OpenThread. Should we prefix it with `OT_`?

PR #4142, opened 2019-09-04
[posix-sim] add method route_list & route_table to Node

simonlingoogle:
The commits implement the `router_list` and `router_table` methods for Node in simulator. The methods should be useful someday. 
The commits also include a simple test to verify correctness.

### DIFF
--- a/tests/scripts/thread-cert/test_route_table.py
+++ b/tests/scripts/thread-cert/test_route_table.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python
+#
+#  Copyright (c) 2017, The OpenThread Authors.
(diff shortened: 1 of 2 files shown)


tests/scripts/thread-cert/test_route_table.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python
+#
+#  Copyright (c) 2017, The OpenThread Authors.
```
librasungirl: nit: 2019

simonlingoogle: updated

PR #4143, opened 2019-09-04
[heap] export heap API

KaSroka:


jwhui:
@KaSroka, thanks for this contribution. Can you include the motivation for introducing this API in the commit message?

Vyrastas:
@KaSroka As part of this can you please also update https://github.com/openthread/openthread/blob/master/include/openthread/error.h with the defgroup for the Heap API?  That way it gets picked up by our Doxygen runs.

> @defgroup api-heap Heap

After api-factory-diagnostics in the Add Ons section would be best.

KaSroka:
@Vyrastas  Ok

@jwhui Here's the crypto library that I'm working on: https://github.com/KaSroka/openthread/commit/c0a488b5aa0078f1a07e558e5a4e057bdee9db3e
I don't want to submit a PR with it as it currently incorporates RC1 version of CC310 libs and I'm waiting for finall release. I hope that this would clarify my problem.

abtink:
Hi @KaSroka, 

I understand the reason you may need these APIs but in general I am not in favor of adding them as  public APIs into OT.  

My view is that OT is a networking stack implementing Thread protocol. We define a bunch of platform APIs to abstract functionality we need from a platform/OS (like timer, settings, heap, etc), but in terms of OT's public APIs, I think they all should be related to Thread and directly control OT's networking behavior (I always push back on adding public APIs that provide OS/platform like APIs from OT like tasklet, timer,  settings, etc functionality).

If I recall, we considered an idea to move the heap implementation in OT into a utility (like flash) that can be used by platform layer to provide the functionality (so the whole heap is delegated out to platform layer which can then either use the provided implementation and provide its own).  I think this way other modules can get it from platform layer directly. Can we consider this approach?

KaSroka:
@abtink I 100% agree with the idea that we should move it to the platform. Unfortunately, this would require some work as at least till now in utils we have only C code and heap implementation is in C++.

@abtink @jwhui Can we merge this as it is so we can also merge #4267 (it's quite big and hard to maintain out of tree)? And move heap to platform utils after that? If not, could someone help me with moving heap to platform/utils as I'm pretty busy right now?

abtink:
@KaSroka , I understand you want this merged to enable the use-case, but I think we agree that these public APIs should be removed  at later point.

As a solution, I suggest adding a  comment in description of the public Heap APIs mentioning that they are temporary and will be removed in future (maybe also removing doxygen documentation of APIs?). I am ok merging this if we document that it's a tempt solution.

Thoughts?

KaSroka:
@abtink I think I've changed it as you suggested.

PR #4144, opened 2019-09-04
[mac-frame] helper methods to copy ExtAddress in normal/reverse byte order

abtink:
This commit updates `Mac::ExtAddress` and `Mac::Address` to add
helper methods to copy address from/to a byte array in normal or
reverse bye order. This commit also adds a unit test for the two
Address classes.

### DIFF
--- a/src/core/mac/mac_frame.hpp
+++ b/src/core/mac/mac_frame.hpp
@@ -97,10 +97,33 @@ public:
     typedef String<kInfoStringSize> InfoString;
 
     /**
+     * This enumeration type specifies the copy byte order when Extended Address is being copied to/from a buffer.
+     *
+     */
+    enum CopyByteOrder
+    {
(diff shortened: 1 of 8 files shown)


src/core/mac/mac_frame.hpp
```diff
@@ -168,6 +191,17 @@ class ExtAddress : public otExtAddress
      */
     void ToggleLocal(void) { m8[0] ^= kLocalFlag; }
 
+    /**
+     * This method copies the Extended Address into a given buffer.
+     *
+     * @param[out]  A pointer to buffer to copy the Extended Address into in reverse order.
```
librasungirl: nit: whether or not `in reverse order` may depend on `aByteOrder`?
missing doxgen for `aByteOrder`

abtink: Thanks. Fixed them in new pushed commit.

PR #4149, opened 2019-09-08
[mac] add Mac::ExtendedPanId type (wrapping otExtendedPanId)

abtink:
The wrapper class `Mac::ExtendedPanId` mainly defines overloads of operators `==` and `!=` and `ToString()` method.

### DIFF
--- a/src/core/mac/mac.hpp
+++ b/src/core/mac/mac.hpp
@@ -414,20 +414,20 @@ public:
     void SetPanId(PanId aPanId);
 
     /**
-     * This method returns the IEEE 802.15.4 Extended PAN ID.
+     * This method returns the IEEE 802.15.4 Extended PAN Identifier.
      *
-     * @returns A pointer to the IEEE 802.15.4 Extended PAN ID.
+     * @returns The IEEE 802.15.4 Extended PAN Identifier.
      *
      */
-    const otExtendedPanId &GetExtendedPanId(void) const { return mExtendedPanId; }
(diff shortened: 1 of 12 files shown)


src/core/mac/mac.hpp
```diff
@@ -419,15 +419,26 @@ class Mac : public InstanceLocator
      * @returns A pointer to the IEEE 802.15.4 Extended PAN ID.
```
jwhui: nit: update docs since this does not return a "pointer"?

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -361,18 +362,21 @@ class ExtendedPanIdTlv : public Tlv
...
 
     /**
      * This method sets the Extended PAN ID value.
      *
-     * @param[in]  aExtendedPanId  A pointer to the Extended PAN ID value.
+     * @param[in]  aExtendedPanId  A Extended PAN ID value.
      *
      */
-    void SetExtendedPanId(const otExtendedPanId &aExtendedPanId) { mExtendedPanId = aExtendedPanId; }
+    void SetExtendedPanId(const otExtendedPanId &aExtendedPanId)
```
jwhui: nit: should we use `ExtendedPanId` here instead of `otExtendedPanId` to be consistent with `GetExtendedPanId()`? I realize this will likely require adding `static_cast` at some call sites.

PR #4150, opened 2019-09-08
[key-manager] add MasterKey class (wrapper over otMasterKey)

abtink:
The `MasterKey` class mainly provides overloads of operators `==`
and `!=` which help simplfy the code comparing keys.

### DIFF
--- a/src/core/meshcop/meshcop_tlvs.hpp
+++ b/src/core/meshcop/meshcop_tlvs.hpp
@@ -50,6 +50,7 @@
 #include "meshcop/timestamp.hpp"
 #include "net/ip6_address.hpp"
 #include "radio/radio.hpp"
+#include "thread/key_manager.hpp"
 
 namespace ot {
 namespace MeshCoP {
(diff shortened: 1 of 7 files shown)


src/core/thread/key_manager.cpp
```diff
@@ -69,7 +69,7 @@ static const otMasterKey kDefaultMasterKey = {{
 
 KeyManager::KeyManager(Instance &aInstance)
     : InstanceLocator(aInstance)
-    , mMasterKey(kDefaultMasterKey)
+    , mMasterKey()
```
jwhui: nit: can remove from initializer list since it is initialized in the body below

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -518,18 +519,21 @@ class NetworkMasterKeyTlv : public Tlv
...
 
     /**
      * This method sets the Network Master Key value.
      *
-     * @param[in]  aNetworkMasterKey  A pointer to the Network Master Key value.
+     * @param[in]  aMasterKey  The Network Master Key.
      *
      */
-    void SetNetworkMasterKey(const otMasterKey &aNetworkMasterKey) { mNetworkMasterKey = aNetworkMasterKey; }
+    void SetNetworkMasterKey(const otMasterKey &aMasterKey)
```
jwhui: nit: should we use `MasterKey` here instead of `otMasterKey` to be consistent with `GetNetworkMasterKey()`? I realize this will likely require adding `static_cast` at some call sites.

PR #4151, opened 2019-09-09
[tlv] simplify searching for a TLV or an Extended TLV within a message

abtink:
This commit contains the following:

- Adds `GetSize()`, `GetValue()` and `GetNext()` to `ExtendedTlv`
  overriding the inherited ones from `Tlv` (which would be
  incorrect for an `ExtendedTlv`).
- Defines a new common private static method `Tlv::Find()` which
  is then used to simplify `Tlv::Get()` and `Tlv::GetOffset()` and
  `Tlv::GetValueOffset()` implementations.
- This change also ensures `Tlv::Get()` (which finds and reads a
  TLV of a given type within a message) work correctly for Extended
  TLVs.
- Adds `Message::AppendTlv()` method to handle `ExtendedTlv`.

### DIFF
--- a/src/core/common/message.cpp
+++ b/src/core/common/message.cpp
@@ -407,6 +407,11 @@ otError Message::AppendTlv(const Tlv &aTlv)
     return Append(&aTlv, aTlv.GetSize());
 }
 
+otError Message::AppendTlv(const ExtendedTlv &aExtendedTlv)
+{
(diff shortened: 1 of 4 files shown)


abtink:
Closing this PR in favor of https://github.com/openthread/openthread/pull/4152.

PR #4152, opened 2019-09-10
[tlv] handle Extended TLV and simplify searching for TLV in a message

abtink:
This commit contains the following change in `Tlv` class:

- Changes `GetSize()`, `GetValue()` and `GetNext()` to ensure they
  work correctly independent of whether the TLV is an Extended TLV
  or not.
- Defines a new common private static method `Tlv::Find()` which
  is then used to simplify `Tlv::Get()` and `Tlv::GetOffset()` and
  `Tlv::GetValueOffset()` implementations.
- This change also ensures `Tlv::Get()` (which finds and reads a
  TLV of a given type within a message) work correctly for Extended
  TLVs.


src/core/common/tlvs.cpp
```diff
@@ -38,98 +38,120 @@
...
+        {
+            ExtendedTlv extTlv;
+
+            VerifyOrExit(offset + sizeof(ExtendedTlv) <= end);
+            aMessage.Read(offset, sizeof(ExtendedTlv), &extTlv);
+            size = extTlv.GetSize();
         }
 
-        VerifyOrExit(length <= end - offset);
+        VerifyOrExit(offset + size <= end);
```
jwhui: This change can result in overflow errors. Thoughts on reverting to the original logic (`size <= end - offset`)?

abtink: Thanks! Great point! will change it.

Issue #1476, opened 2017-03-16
Compiler vrs Compiler issues, vrs IDE issues.


DuaneEllis-TI:
Open thread relies upon a number of standard C header files that are different across compilers, and compiler releases.

For example:

#include <string.h> - depending on the compiler does not have some functions.
This also varies by compiler version, (ie: a fix was introduce in a later version)

In the GNU Autoconf world this is solved via either  "libiberty' or various "#ifdefs"
GCC supports an   #include_next <filename> scheme - other compilers do not.

a) This does not play nicely with IDEs that do not support Autoconf.
b) This also makes code "messy" with lots of #if/#else/#endif fixes.

I propose the following solution:

If a standard header files needs to be fixed for a specific tool.
Then - the following action is taken.

Step 1:

A "pseudo-replacement" header file is created in a new directory called: "fixed"
    #include <openthread/platform/fixed/fixed_FILENAME.H>

Alternatively it could be placed in   ${openthread}/src/<somewhere>/fixed/fixed_FILENAME.H
The concern here is if thread ever provids a "C++" api - with functions defined within a class.

Note: specifically the original filename should not be used "as is", but a standard prefix or suffix be added to the original filename to make it unique. In this example I propose: "fixed_FILENAME.h"

Example:  #include <stdio.h>
Becomes:     #include <openthread/platform/fixed/stdio.h 

Step 2:

This new <openthread/platform/fixed/fixed_FILENAME.H>
Must eventually include the tool chain provided FILENAME.H

But also contains any required compiler/library fixes in the form of:  #if/#else/#endif

Step 3:  Problematic Thread code then includes the "fixed" header file in lue of the standard header file.

aeliot:
These standard headers, and the function signatures, should be the same across compilers that support C99 (ISO/IEC 9899), shouldn't they?  Since we are using C99, I think as long as we are carful to only use standard functions that are from C99 and not a later revision, we should be safe.

DuaneEllis-TI:
Question: Is strnlen() c99? or was this added at a later date 

I believe this is an online copy of the C99 standard:

http://cs.nyu.edu/courses/summer12/CSCI-GA.2110-001/downloads/C99.pdf

the function is not described in that document.

Other places I have tried/looked.

I do not see it here: http://en.cppreference.com/w/c/string/byte
I do not see it here: https://en.wikibooks.org/wiki/C_Programming/Strings
[...]

PR #4154, opened 2019-09-10
[utils] remove stdint and stdbool wrappers

bukepo:
This PR removes the wrap_stdint.h and wrap_stdbool.h because OpenThread
enforces C99, which includes these headers in its standard library.


jwhui:
For background, these wrappers were introduced to address issues with Visual Studio (see #1476). Given that we have removed Windows support, I am fine with this PR unless someone else objects.

PR #4155, opened 2019-09-11
[posix-app] log filename when die

bukepo:
This PR aims at adding the filename when die. Other enhancements:

* Enable PLATFORM region by default for posix-app
* Disable warning for zero variadic macro arguments
* Use VerifyOrExit to implement other assert utilities
* Do not call fprintf() and print exit code in assert utility

To print dying message to stderr, add `-v` when launching the app.
To get the exit code, use `echo $?` just after the app died.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -233,7 +233,7 @@ AC_PATH_PROG(CMP, cmp)
 #
 
 PROSPECTIVE_CFLAGS="-Wall -Wextra -Wshadow -Werror -std=c99 -pedantic-errors -Wno-gnu-zero-variadic-macro-arguments"
-PROSPECTIVE_CXXFLAGS="-Wall -Wextra -Wshadow -Werror -std=gnu++98 -Wno-c++14-compat -fno-exceptions -Wno-gnu-zero-variadic-macro-arguments"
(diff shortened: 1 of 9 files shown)


configure.ac
```diff
@@ -232,8 +232,8 @@ AC_PATH_PROG(CMP, cmp)
 #   -Wall                        CC, CXX
 #
 
-PROSPECTIVE_CFLAGS="-Wall -Wextra -Wshadow -Werror -std=c99 -pedantic-errors"
-PROSPECTIVE_CXXFLAGS="-Wall -Wextra -Wshadow -Werror -std=gnu++98 -Wno-c++14-compat -fno-exceptions"
+PROSPECTIVE_CFLAGS="-Wall -Wextra -Wshadow -Werror -std=c99 -pedantic-errors -Wno-gnu-zero-variadic-macro-arguments"
+PROSPECTIVE_CXXFLAGS="-Wall -Wextra -Wshadow -Werror -std=gnu++98 -Wno-c++14-compat -fno-exceptions -Wno-gnu-zero-variadic-macro-arguments"
```
jwhui: I don't think `-Wno-gnu-zero-variadic-macro-arguments` is needed for `CXXFLAGS`.

PR #4156, opened 2019-09-11
[examples] provide APIs to manipulate 802.15.4 frames

bukepo:
This PR exposes several API for radio driver writers to manipulate MAC
frames.

### DIFF
--- a/examples/platforms/posix/radio.c
+++ b/examples/platforms/posix/radio.c
@@ -39,6 +39,7 @@
 #include <openthread/platform/time.h>
 
 #include "utils/code_utils.h"
+#include "utils/mac_frame.h"
 #include "utils/soft_source_match_table.h"
 
 // The IPv4 group for receiving packets of radio simulation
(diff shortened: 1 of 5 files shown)


examples/platforms/posix/radio.c
```diff
@@ -45,44 +45,11 @@
...
-    IEEE802154_SEC_LEVEL_MASK = 7 << 0,
-
-    IEEE802154_KEY_ID_MODE_0    = 0 << 3,
-    IEEE802154_KEY_ID_MODE_1    = 1 << 3,
-    IEEE802154_KEY_ID_MODE_2    = 2 << 3,
-    IEEE802154_KEY_ID_MODE_3    = 3 << 3,
-    IEEE802154_KEY_ID_MODE_MASK = 3 << 3,
-
-    IEEE802154_MACCMD_DATA_REQ = 4,
+    IEEE802154_FRAME_PENDING = 1 << 4,
```
Irving-cl: Shall we put all these constants `include/openthread/platform/radio.h ` so that they could be reused?

Irving-cl: For the contants removed above, I think I would use them later to implement other APIs, so maybe I would add them later into the public header.

include/openthread/platform/radio.h
```diff
@@ -169,6 +180,21 @@ typedef struct otRadioIeInfo
     uint8_t mTimeSyncSeq;       ///< The Time sync sequence.
 } otRadioIeInfo;
 
+/**
+ * This structure represents an IEEE 802.15.4 short or extended Address.
+ *
+ */
+typedef struct otMacAddress
```
abtink: Move this out of `radio.h` (since not used by any `otPlatRadio`)? 
Suggest adding this and `otMacAddressType` to utls/mac_frame.h`

src/core/mac/mac_frame.hpp
```diff
@@ -256,57 +256,43 @@ class Address
      */
     typedef ExtAddress::InfoString InfoString;
 
-    /**
-     * This enumeration specifies the IEEE 802.15.4 Address type.
-     *
-     */
-    enum Type
```
abtink: Suggeston: Keep this definition as is. 
Then convert from this type to `otMacAddress` (which I suggested to move out `radio.h` and into `util/mac_frame.h`) in util/mac-frame code.

bukepo: no problem, thanks for the suggestions

bukepo: @abtink if we don't let Mac::Address inherit `otMacAddress`, we have to use `reinterpret_cast`, are you OK with it?

abtink: If we can ensure it works correctly. Alternative can be to convert and re-pooluate the values/feidls accordingly.

examples/platforms/utils/mac_frame.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+bool otMacFrameIsAckRequested(const otRadioFrame *aFrame)
+{
+    return static_cast<const Mac::Frame *>(aFrame)->GetAckRequest();
+}
+
+otError otMacFrameGetSrcAddr(const otRadioFrame *aFrame, otMacAddress *aMacAddress)
+{
+    return static_cast<const Mac::Frame *>(aFrame)->GetSrcAddr(*reinterpret_cast<Mac::Address *>(aMacAddress));
```
abtink: Do we know for sure that the `reinterpret_cast<Mac::Address *>` would work on all toolchains/compliers?
To be safe, how about we add code to convert, something like this:
```
otError otMacFrameGetSrcAddr(const otRadioFrame *aFrame, otMacAddress *aMacAddress)
{
	otError error;
	Mac::Address address;

	error = static_cast<const Mac::Frame *>(aFrame)->GetSrcAddr(address);
	SuccessOrExit(error);

	switch (address.GetType())
	{
	case Mac::Address::kTypeNone:
		aMacAddress->mType = OT_MAC_ADDRESS_TYPE_NONE;
		break;

	case Mac::Address::kTypeShort:
	    aMacAddress->mType = OT_MAC_ADDRESS_TYPE_SHORT;
		aMacAddress->mAddress.mShortAddress = address.GetShort();
		break;

	case Mac::Address::kTypeExtended:
	    aMacAddress->mType = OT_MAC_ADDRESS_TYPE_EXTENDED
		aMacAddress->mAddress.mExtendedAddress = address.GetExtended();
		break;
	}

exit:
	return error;
}
```

examples/platforms/posix/radio.c
```diff
@@ -134,166 +103,55 @@ static otRadioFrame        sAckFrame;
...
 static bool sRadioCoexEnabled = true;
 #endif
 
-static inline bool isFrameTypeAck(const uint8_t *aFrame)
-{
-    return (aFrame[0] & IEEE802154_FRAME_TYPE_MASK) == IEEE802154_FRAME_TYPE_ACK;
-}
-
-static inline bool isFrameTypeMacCmd(const uint8_t *aFrame)
+static void RevertExtAddress(otExtAddress *aReverted, const otExtAddress *aOrigin)
```
jwhui: - `RevertExtAddress` -> `ReverseExtAddress`
- `aReverted` -> `aReversed`

examples/platforms/utils/mac_frame.cpp
```diff
@@ -0,0 +1,117 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "mac_frame.h"
+
+#include "mac/mac_frame.hpp"
+
+using namespace ot;
+
+bool otMacFrameIsMatched(const otRadioFrame *aFrame,
```
jwhui: nit: how about `otMacFrameDoesAddrMatch()` to better indicate this is address matching?

PR #4169, opened 2019-09-15
[mac] add Mac::NetworkName class

abtink:
This commit adds `Mac::NetworkName` class as C++ wrapper over the
`otNetworkName`. It also adds `NetworkName::Data` to represent the
network name as a pointer to a char array buffer (not necessarily null
terminated) with a given length. This representation is used by
`Mac::BeaconPlayload` and `MeshCop::NetworkNameTlv`. This commit also
adds unit test for `Mac::NetworkName` class.

### DIFF
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -97,6 +97,7 @@ Mac::Mac(Instance &aInstance)
     , mRadioChannel(OPENTHREAD_CONFIG_DEFAULT_CHANNEL)
     , mRadioChannelAcquisitionId(0)
     , mSupportedChannelMask(Get<Radio>().GetSupportedChannelMask())
+    , mNetworkName()
     , mScanChannel(Radio::kChannelMin)
     , mScanDuration(0)
     , mScanChannelMask()
@@ -117,7 +118,6 @@ Mac::Mac(Instance &aInstance)
 
     mCcaSuccessRateTracker.Reset();
     memset(&mCounters, 0, sizeof(otMacCounters));
-    memset(&mNetworkName, 0, sizeof(otNetworkName));
     memset(&mExtendedPanId, 0, sizeof(ExtendedPanId));
 
     mSubMac.Enable();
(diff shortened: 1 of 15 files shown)


abtink:
Updated this PR and removed "work-in-progress" label. It's ready for review.

src/core/mac/mac.cpp
```diff
@@ -446,17 +446,25 @@ void Mac::SetSupportedChannelMask(const ChannelMask &aMask)
     return;
 }
 
-otError Mac::SetNetworkName(const char *aBuffer, uint8_t aLength)
+otError Mac::SetNetworkName(const char *aNameString)
 {
-    otError error  = OT_ERROR_NONE;
-    uint8_t newLen = static_cast<uint8_t>(strnlen(aBuffer, aLength));
+    NetworkName::Data data(aNameString, NetworkName::kMaxSize + 1);
```
jwhui: Does `data.mLength` include the NULL character? Should we have the `+ 1` here?

abtink: Yes.  `data.mLength` does not include/count the null char.
But the `+1` here is intentional to ensure we reject a `aNameString` which has a length longer than `kMaxSize`.

- Basically `NetworkName::Data` has no restriction on what its `mLength` can be set to. 
- When setting from `aNameString` we only care about `kMaxSize + 1` chars in the string.
- When we use `SetNetworkName(data)` (which in turn calls `NetworkName::Set(data)` method), the `Set()` method will check for new length to be fit in `kMaxSize` (otherwise retursn `OT_ERROR_INVALID_ARGS`)
   -  `NetworkName::Set()` will check for any NULL chars within `data` and calculates the `newLen` for the name.

jwhui: Makes sense. Do you mind adding a short comment on why the `+ 1`?

src/core/mac/mac_frame.cpp
```diff
@@ -95,6 +95,46 @@ ExtendedPanId::InfoString ExtendedPanId::ToString(void) const
...
+    otError error  = OT_ERROR_NONE;
+    uint8_t newLen = static_cast<uint8_t>(strnlen(aNameData.GetBuffer(), aNameData.GetLength()));
+
+    VerifyOrExit(newLen <= kMaxSize, error = OT_ERROR_INVALID_ARGS);
+
+    // Ensure the new name does not match the current one.
+    VerifyOrExit(memcmp(m8, aNameData.GetBuffer(), newLen) || (m8[newLen] != 0), error = OT_ERROR_ALREADY);
+
+    memcpy(m8, aNameData.GetBuffer(), newLen);
+    m8[newLen] = 0;
```
jwhui: nit: prefer using`\0` for null characters.

src/core/mac/mac_frame.hpp
```diff
@@ -475,6 +475,110 @@ class ExtendedPanId : public otExtendedPanId
...
+    private:
+        const char *mBuffer;
+        uint8_t     mLength;
+    };
+
+    /**
+     * This constructor initializes the IEEE802.15.4 Network Name as an empty string.
+     *
+     */
+    NetworkName(void) { m8[0] = 0; }
```
jwhui: nit: use `\0`?

PR #4171, opened 2019-09-16
[posix-app] fix crash on os ifconfig down

bukepo:
This PR fixes a crash issue triggered by system command ifconfig down.
This is because OpenThread's tries removing addresses when the platform
network interface is already down.

### DIFF
--- a/src/posix/platform/netif.cpp
+++ b/src/posix/platform/netif.cpp
@@ -98,12 +98,24 @@ static void UpdateUnicast(otInstance *aInstance, const otIp6Address &aAddress, u
     ifr6.ifr6_ifindex   = static_cast<int>(sTunIndex);
     ifr6.ifr6_prefixlen = aPrefixLength;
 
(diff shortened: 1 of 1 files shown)


src/posix/platform/netif.cpp
```diff
@@ -98,12 +98,24 @@ static void UpdateUnicast(otInstance *aInstance, const otIp6Address &aAddress, u
...
+    {
+        VerifyOrExit(ioctl(sIpFd, SIOCDIFADDR, &ifr6) == 0, perror("ioctl"); error = OT_ERROR_FAILED);
+    }
 
 exit:
-    SuccessOrDie(error);
-    otLogInfoPlat("%s: %s", __func__, otThreadErrorToString(error));
+    if (error != OT_ERROR_NONE)
+    {
+        otLogInfoPlat("%s: %s", __func__, otThreadErrorToString(error));
```
jwhui: It seems INFO is used for the error case and WARN is used for success case. Is that intentional?

PR #4174, opened 2019-09-16
[key-manager] add Pskc class

abtink:
This commit adds a `Pskc` as a wrapper over `otPskc` and also
  renames all the uses of `PSK` into `Psk` to follow our variable
  and type naming convention which treats an acronym as a whole
  word.

### DIFF
--- a/src/core/thread/key_manager.hpp
+++ b/src/core/thread/key_manager.hpp
@@ -41,6 +41,7 @@
 #include <openthread/dataset.h>
 
 #include "common/locator.hpp"
+#include "common/random.hpp"
 #include "common/timer.hpp"
 #include "crypto/hmac_sha256.hpp"
 #include "mac/mac_frame.hpp"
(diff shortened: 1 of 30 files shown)


abtink:
One (sort of unrelated point) I realized when making this change:
Within OT our naming convention seems to treat acronyms as whole word and use camel-case model for entire acronym, e.g., `AesCcm`, `otCryptoEcdsaSign`, `mUartBuffer`, `NcpSpi`, 'HdlcBuffer`, `SetKek`, ...

The `PSKc` seems to not follow this model. How about in a future change we rename all the APIs and types to use `Pskc` (it looks a bit weird at first, but I guess we get used to it quickly)?

Thoughts?

jwhui:
Changing to `Pskc` works for me.

src/core/thread/key_manager.hpp
```diff
@@ -88,6 +88,38 @@ class MasterKey : public otMasterKey
...
+    /**
+     * This method evaluates whether or not the Thread PSKc values match.
+     *
+     * @param[in]  aOther  The Thread PSKc to compare.
+     *
+     * @retval TRUE   If the Thread PSKc values do not match.
+     * @retval FALSE  If the Thread PSKc values match.
+     *
+     */
+    bool operator!=(const PSKc &aOther) const { return memcmp(m8, aOther.m8, sizeof(PSKc)) != 0; }
```
wgtdkp: I think it would be better to implement as negative of `operator==()`:
```c++
bool operator!=(const PSKc &aOther) const { return !(*this == aOther); }
```

wgtdkp:
If the purpose of creating the new `Pskc` class that inherits from `otPSKc` is to add the comparing operators, I think another approach of overloading the `==` & `!=` operator for `otPSKc` would also work and doesn't need a new class:

```c++
bool operator==(const otPSKc &aLhs, const otPSKc &aRhs)
{
    return memcmp(aLhs.m8, aRhs.m8, sizeof(otPSKc)) == 0;
}

bool operator!=(const otPSKc &aLhs, const otPSKc &aRhs)
{
    return !(aLhs == aRhs);
}
```

src/core/thread/key_manager.hpp
abtink: I agree that this would be better way probably for more complex structures where `operator==` implementation itself may be more complex (to avoid repeating complex implementation).

But for simple structures and simple implementations like this, I am not sure if there is any advantages. Do you know of any advantages in this case?

abtink:
The class lets us add new methods (like `GenerateRandom()`) in future.

src/core/thread/key_manager.hpp
wgtdkp: To me, `return !(*this == aOther)` makes clearer that `operator!=` is implemented as negative of `operator==`, but don't need to verify if the two calling of `memcmp` are the same. This avoids subtle bugs (such as wrong arguments to `memcmp`).

wgtdkp:
> The class lets us add new methods (like `GenerateRandom()`) in future.

It makes sense.

PR #4175, opened 2019-09-17
[coap] finish cached coap response before sending

simonlingoogle:
The commits fixes the bug that Coap servers send invalid cached responses, specifically messages with Type=CONN, Code=0, MID=0. 
The cached responses should `Finish` before `Send`.

Please find the pcap traces regarding this issue in [pcap.zip](https://github.com/openthread/openthread/files/3619214/pcap.zip)

### DIFF
--- a/src/core/coap/coap.cpp
+++ b/src/core/coap/coap.cpp
@@ -601,6 +601,7 @@ void CoapBase::ProcessReceivedRequest(Message &aMessage, const Ip6::MessageInfo 
     switch (mResponsesQueue.GetMatchedResponseCopy(aMessage, aMessageInfo, &cachedResponse))
     {
     case OT_ERROR_NONE:
+        cachedResponse->Finish();
         error = Send(*cachedResponse, aMessageInfo);
         // fall through
         ;


src/core/coap/coap.cpp
```diff
@@ -601,6 +601,7 @@ void CoapBase::ProcessReceivedRequest(Message &aMessage, const Ip6::MessageInfo
     switch (mResponsesQueue.GetMatchedResponseCopy(aMessage, aMessageInfo, &cachedResponse))
     {
     case OT_ERROR_NONE:
+        cachedResponse->Finish();
```
jwhui: Great catch! Should this be inside `GetMatchedResponseCopy()`?

simonlingoogle: The enqueued responses are not `Finish`ed. 
If `GetMatchedResponseCopy` returns messages already `Finish`ed, wouldn't that be asymmetric?

jwhui: Should the enqueued responses be "finished" in `EnqueueResponse()`?

simonlingoogle: Yes, I think so. However, this PR was meant to be a quick fix for this bug.
@bukepo is working on a more comprehensive refactoring of coap message sending code.

PR #4176, opened 2019-09-17
[tlvs] add length check in Tlv::Find for extended tlvs

jwhui:


src/core/common/tlvs.cpp
```diff
@@ -128,6 +128,8 @@ otError Tlv::Find(const Message &aMessage, uint8_t aType, uint16_t *aOffset, uin
 
             VerifyOrExit(sizeof(ExtendedTlv) <= remainingLen);
             aMessage.Read(offset, sizeof(ExtendedTlv), &extTlv);
+
+            VerifyOrExit(extTlv.GetLength() <= (remainingLen - sizeof(ExtendedTlv)));
```
abtink: Maybe I am missing something here?
We have `VerifyOrExit(size <= remainingLen);` right after this which I think would check for same.

abtink: Ah! Figured it out! Issue with wrapping of `uint16_t` length...

PR #4178, opened 2019-09-17
[meshcop] enforce max length on commissioner id

jwhui:


src/core/meshcop/leader.cpp
```diff
@@ -107,6 +107,11 @@ void Leader::HandlePetition(Coap::Message &aMessage, const Ip6::MessageInfo &aMe
 
     mCommissionerId = commissionerId;
 
+    if (mCommissionerId.GetLength() > CommissionerIdTlv::kMaxLength)
```
bukepo: Should we just reject if the commissioner id exceeds the max length allowed?

jwhui: I think this is a bit of a grey area.

1. The maximum length of Commissioner ID is 64 bytes. However, the only field that specifies the length is the TLV's Length field.

2. Thread has a general rule of allowing future extensions to TLVs. If we want to apply that rule to the Commissioner ID TLV, then that means any Length value greater than 64 indicates a 64-byte Commissioner ID and additional payload afterwards.

I opted for this implementation to be more flexible to future extensions to Commissioner ID TLV.

Thoughts?

bukepo: Got it!

PR #33, opened 2016-05-15
Add radio method to check whether it supports ACK timeout or not

luwannest:
Some radio vendors may support ACK timeout, but some may not. This method is for checking this to determine the behavior of MAC layer.

PR #4191, opened 2019-09-19
[log] add proper log level output and LOG_COLOR option to posix.

turon:


src/core/config/logging.h
```diff
@@ -277,6 +277,16 @@
...
+/**
+ * @def OPENTHREAD_CONFIG_LOG_COLOR_ENABLE
+ *
+ * Define as 1 to enable color in logs.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_LOG_COLOR_ENABLE
+#define OPENTHREAD_CONFIG_LOG_COLOR_ENABLE 0
+#endif
+
```
abtink: While I like  the idea of adding color in posix logging, I am not sure about the configuration through the `config/logging.h` and common-switches makefile.
- In general, definitions under `core/config/` should configure the OT core behavior itself (and be independent of and applicable to all/multiple platforms).
- Other definitions in the `config/logging.h` do follow this (they impact how the logging is implemented with OT core). 
- However `OPENTHREAD_CONFIG_LOG_COLOR_ENABLE` does not change anything within core, and it needs to be handled by the platform code (it's used by `otPlatLog()` from `posix` example which is part of platform code).
- Also this config is only applicable to posix platform example.

My suggestion is to add this in `platform-config` specific to posix example (`example/platforms/posix/platform-config.h` file).

jwhui: Agree with @abtink's point above.

Some other points worth mentioning:

- rsyslog escapes control characters by default. There is a
 global `/etc/rsyslog.conf` configuration [EscapeControlCharactersOnReceive](https://www.rsyslog.com/doc/v8-stable/configuration/input_directives/rsconf1_escapecontrolcharactersonreceive.html), but that will affect all input received by rsyslog.
  ```
  Sep 20 21:54:26 558cfb0e1e95 ot-cli-ftd[11946]: #033[1;36m[1] [NOTE]-MLE-----: Role Disabled -> Detached#033[0m
  ```

- Why not leverage existing colorization tools such as [ccze](https://github.com/cornet/ccze)? Note that this does not require any change to OpenThread and rsyslog configuration. Example screenshot:
  Screen Shot 2019-09-20 at 3 11 03 PM

examples/platforms/posix/logging.c
```diff
@@ -73,8 +81,45 @@ OT_TOOL_WEAK void otPlatLog(otLogLevel aLogLevel, otLogRegion aLogRegion, const
...
+        logLevel = LOG_DEBUG;
+        break;
+    default:
+        assert(false);
+        logLevel = LOG_DEBUG;
+        break;
+    }
+
 exit:
-    syslog(LOG_CRIT, "%s", logString);
```
jwhui: The reason why everything was mapped to syslog's critical log level is because rsyslog defaults to notice log level. As a result, critical level ensures that all OpenThread logs can be seen without any change to `/etc/rsyslog.d`. Note that `examples/platforms/posix` is purely intended for simulating OpenThread and not to be used for anything beyond simulation. `syslog` was only introduced here as a convenient method for serializing logs between multiple posix simulation processes.

jwhui:
Closing stale PR.

PR #4194, opened 2019-09-19
[meshcop] improve HandleCommissioningGet implementation

jwhui:
- Fix insufficient length checks.
- Avoid using a large stack buffer.

### DIFF
--- a/src/core/thread/network_data_leader_ftd.cpp
+++ b/src/core/thread/network_data_leader_ftd.cpp
@@ -286,32 +286,19 @@ void Leader::HandleCommissioningGet(void *aContext, otMessage *aMessage, const o
 
 void Leader::HandleCommissioningGet(Coap::Message &aMessage, const Ip6::MessageInfo &aMessageInfo)
 {
-    MeshCoP::Tlv tlv;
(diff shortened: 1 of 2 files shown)


src/core/thread/network_data_leader_ftd.cpp
```diff
@@ -286,32 +286,19 @@ void Leader::HandleCommissioningGet(void *aContext, otMessage *aMessage, const o
...
-        aMessage.Read(offset, sizeof(tlv), &tlv);
-
-        if (tlv.GetType() == MeshCoP::Tlv::kGet)
-        {
-            length = tlv.GetLength();
-            aMessage.Read(offset + sizeof(MeshCoP::Tlv), length, tlvs);
-            break;
-        }
+    SuccessOrExit(Tlv::GetValueOffset(aMessage, MeshCoP::Tlv::kGet, offset, length));
+    aMessage.SetOffset(offset);
```
abtink: Changing offset on `aMessage` wont have any side-effects? (I think we are the only one processing the `aMessage`, right?).

jwhui: Yes, this should be safe. No other uses after this method completes.

PR #4195, opened 2019-09-20
[time] adding Time class

abtink:
This commit adds `Time` class which represents an instance of time (it is a simple wrapper over a `uint32_t` corresponding to a numerical time value). The `Time` class provides helpful operator overloads:
- Operators `+` and `-` with a `Time` instance and a `Duration` to get a `Time` instance after or before.
- Operator `-` with two `Time` instances to calculate the `Duration` between two time instances.
 - Operators `<`, `<=`, `>`,`>=`, `==`, and `!=` to compare two `Time` instances. They correctly handle the wrapping of the numeric time value.
    
The core modules are updated to use the new `Time` and `Duration` types which help make the code simpler. This commit also updates the unit test `test_timer` to add test cases for newly
added types.

-----
The main reason for adding a `Time` class (which simply wraps over the `uint32_t` time tick count) is to ensure the comparison operators (`<`, `>`, `<=` and `>=`) are overloaded and behave correctly when the raw time tick count value is wrapped. 

This blocks a (common) wrong use pattern in the code where two raw time (`uint32_t`) values `t1` and `t2` are compared using `t1 < t2` to determine if time `t1` is before `t2`. 
- For example, if `t1 = 0xfffffffe` and `t2` is 5 ticks ahead at `t2 = 0x3`, the check `t1 < t2` would be `false` where as `t1` should be considered before `t2`.
- Such a use would be wrong only when the raw values wrap. For a millisecond `uint32_t` tick time, the wrap happens every ~50 days, so for a large span of time things would work fine leading to bugs that are tricky to find.

The new `Time` class allows this use pattern by overloading the operator to handle the warping correctly which in turn help us avoid such errors in future code.

-----

Another aspect addressed by this PR is the code pattern used to compare two raw `uint32_t` time tick values using `static_cast<int32_t>(t1 - t2) < 0` to determine if `t1` is before `t2` ([example](https://github.com/openthread/openthread/blob/7d423506932df0e1027e2c9d920ef8e9cfab080b/src/core/thread/mle.hpp#L457)). The type casting from an unsigned to a signed integer is (from C++ standard) an undefined behavior [https://en.cppreference.com/w/cpp/language/implicit_conversion]. How this is handled is left to the implementation of compiler/toolchain. While most toolchains do what we expect (i.e. convert a large `uint32_t` value to a negative `int32_t` value), this is not guaranteed (have been bitten by a toolchain not behaving like this in a previous project). This PR ensures that we don't need to use the `static_cast<int32_t>` thus making the code safer.

### DIFF
--- a/src/core/api/thread_api.cpp
+++ b/src/core/api/thread_api.cpp
@@ -320,8 +320,8 @@ otError otThreadGetParentInfo(otInstance *aInstance, otRouterInfo *aParentInfo)
     aParentInfo->mPathCost       = parent->GetCost();
     aParentInfo->mLinkQualityIn  = parent->GetLinkInfo().GetLinkQuality();
     aParentInfo->mLinkQualityOut = parent->GetLinkQualityOut();
-    aParentInfo->mAge       = static_cast<uint8_t>(TimerMilli::MsecToSec(TimerMilli::Elapsed(parent->GetLastHeard())));
-    aParentInfo->mAllocated = true;
+    aParentInfo->mAge            = static_cast<uint8_t>(Time::MsecToSec(TimerMilli::GetNow() - parent->GetLastHeard()));
+    aParentInfo->mAllocated      = true;
     aParentInfo->mLinkEstablished = parent->GetState() == Neighbor::kStateValid;
 
 #if !OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE
--- a/src/core/common/time.hpp
+++ b/src/core/common/time.hpp
@@ -1,0 +1,263 @@
+/*
+ *  Copyright (c) 2019, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes definitions for time instance.
+ */
+
+#ifndef TIME_HPP_
+#define TIME_HPP_
+
+#include "openthread-core-config.h"
+
+#include <stddef.h>
+#include <stdint.h>
+
+namespace ot {
+
+/**
+ * @addtogroup core-timer
+ *
+ * @brief
+ *   This module includes definitions for the time instance.
+ *
+ * @{
+ *
+ */
+
+/**
+ * This class represents a time instance.
+ *
+ */
+class Time
+{
+public:
+    /**
+     * This constant defines a maximum time duration ensured to be longer than any other duration.
+     *
+     */
+    static const uint32_t kMaxDuration = ~static_cast<uint32_t>(0UL);
+
+    /**
+     * This is the default constructor for a `Time` object.
+     *
+     */
+    Time(void) {}
+
+    /**
+     * This constructor initializes a `Time` object with a given value.
+     *
+     * @param[in] aValue   The numeric time value to initialize the `Time` object.
+     *
+     */
+    explicit Time(uint32_t aValue) { SetValue(aValue); }
+
+    /**
+     * This method gets the numeric time value associated with the `Time` object.
+     *
+     * @returns The numeric `Time` value.
+     *
+     */
+    uint32_t GetValue(void) const { return mValue; }
+
+    /**
+     * This method sets the numeric time value.
+     *
+     * @param[in] aValue   The numeric time value.
+     *
+     */
+    void SetValue(uint32_t aValue) { mValue = aValue; }
+
+    /**
+     * This method calculates the time duration between two `Time` instances.
+     *
+     * @note Expression `(t1 - t2)` returns the duration of the interval starting from `t2` and ending at `t1`. When
+     * calculating the duration, `t2 is assumed to be in the past relative to `t1`. The duration calculation correctly
+     * takes into account the wrapping of numeric value of `Time` instances. The returned value can span the entire
+     * range of the `uint32_t` type.
+     *
+     * @param[in]   aOther  A `Time` instance to subtract from.
+     *
+     * @returns The duration of interval from @p aOther to this `Time` object.
+     *
+     */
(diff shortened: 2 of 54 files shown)


abtink:
The last commit in this PR makes the medta data types that are saved in `Message` (e.g., the `CoapMetadata` or `MplBufferedMessageMetadata`) non-packed structs.
- As we read/write the entire struct as a data blob in/form a `message`, the packed vs unpacked should not matter. 
- I think these structs were originally defined as packed to ensure we keep them as small as possible.
- It should not matter much though as the `Message` size is increased in chunks and these definitions already seem to align their member variables.
- Since we include `Time` type member variables in these meta data definitions, if we define them as packed, then 
   - Either the `Time` type itself need to be defined as packed.
   - Or we cannot use the `Time` type (instead use the raw `uint32_t` values to keep time stamps) - which is not ideal - I opted not to go for this.

In the first version of  this PR,  declared the `Time` class as packed, however, it caused a major increase in code size, due to fact that complier could no longer easily optimize and inline the simple methods in `Time` (since due to it being packed, complier needed take care of all cases even when a `Time` object would not be aligned in memory).

src/core/common/time.hpp
```diff
@@ -0,0 +1,359 @@
...
+     * This type represents the underlying unsigned int type used to store numeric value of a `Time` instance.
+     *
+     */
+    typedef uint32_t Value;
+
+    /**
+     * This constant defines a forever duration ensured to be longer than any other duration.
+     *
+     */
+    static const Duration kForeverDuration = ~static_cast<Duration>(0UL);
```
jwhui: Is this really "forever"? Is `kMaxDuration` more appropriate?

src/core/common/time.hpp
```diff
@@ -0,0 +1,359 @@
...
+     */
+    explicit Time(Value aValue) { Set(aValue); }
+
+    /**
+     * This method gets the numeric time value associated with the `Time` object.
+     *
+     * @returns The numeric `Time` value.
+     *
+     */
+    Value Get(void) const { return mValue; }
```
jwhui: Is this better as `GetValue()`? In calling code, things like `GetNow().Get()` seems a bit ambiguous to me.

src/core/common/time.hpp
```diff
@@ -0,0 +1,359 @@
...
+    /**
+     * This method sets the numeric time value.
+     *
+     * @param[in] aValue   The numeric time value.
+     *
+     */
+    void Set(Value aValue) { mValue = aValue; }
+
+    /**
+     * This method calculates the duration of time interval between two `Time` instances.
```
jwhui: suggestion: "duration of time interval" -> "time duration"

src/core/common/time.hpp
```diff
@@ -0,0 +1,359 @@
...
+     *
+     * @note Expression `(t1 - t2)` returns the duration of the interval starting from `t2` and ending at `t1`. The
```
jwhui: Clarify whether `t2` is assumed to be in the past?

abtink:
In the last push, added new smaller commits where we replace `memset` use with constructors which clear the object :
-  [mle] remove memset from DelayedResponseHeader constructors (#4195)
- [dns-client] remove memset(zero) from constructor (#4195)
- [core] make metadata (saved in `Message`) non-packed (#4195)

src/core/api/thread_api.cpp
```diff
@@ -320,7 +320,7 @@ otError otThreadGetParentInfo(otInstance *aInstance, otRouterInfo *aParentInfo)
     aParentInfo->mPathCost       = parent->GetCost();
     aParentInfo->mLinkQualityIn  = parent->GetLinkInfo().GetLinkQuality();
     aParentInfo->mLinkQualityOut = parent->GetLinkQualityOut();
-    aParentInfo->mAge       = static_cast<uint8_t>(TimerMilli::MsecToSec(TimerMilli::Elapsed(parent->GetLastHeard())));
+    aParentInfo->mAge       = static_cast<uint8_t>(TimeMilli::MsecToSec(TimeMilli::GetNow() - parent->GetLastHeard()));
```
jwhui: nit: any reason not to use `Time::MsecToSec`? `MsecToSec` seems to be defined as a part of `Time`.

src/core/common/time.hpp
```diff
@@ -0,0 +1,359 @@
...
+ *
+ */
+class Time
+{
+public:
+    /**
+     * This type represents a time duration.
+     *
+     */
+    typedef uint32_t Duration;
```
jwhui: What are the benefits of defining a different `Duration` type? Why not just use the `Time` type?

src/core/common/time.hpp
abtink: We have a `kMaxDuration` definition (in `Timer` class) which is the max duration we can use in a timer APIs `Start()` and `StartAt()`. Even though we can have a `Time:kMaxDuration` and a `Timer::kMaxDuration` I think it'd be better to use a different name for these two definitions.  I opted for `kForeverDuration` which was similar to `kForeverDt` we were using before. Suggestions for new names for these are most welcome. :)

src/core/common/time.hpp
abtink: I think they represent different things: One is a duration of time, the other is the raw value/storage associated with a `Time` instance. 

I went  for two types separate definition to allow future flexibility. For example, we can later change `Time` to store a `uint64_t` (or maybe some other way) without the need to change `Duration` type.

Ideally we want the user of `Time` (or `Timer`) classes to never be aware of how/what `Time` class stores (we still need `GetValue()` and `SetValue()` but would like in future to get rid of them if we can).

src/core/api/thread_api.cpp
abtink: Good point! Will change them.

src/core/common/time.hpp
jwhui: How about `Timer::kMaxDelay` vs. `Time::kMaxDuration`?

src/core/common/time.hpp
jwhui: If `Time` stores a `uint64_t` and `Duration` is a `uint32_t`, then the `-` operator that returns `Duration` needs to have special value for "out-of-range" along with associated code that may need to deal with it.

src/core/common/time.hpp
```diff
@@ -0,0 +1,348 @@
...
+    {
+    }
+
+    /**
+     * This static method returns the current time in milliseconds.
+     *
+     * @returns The current time in milliseconds.
+     *
+     */
+    static TimeMilli GetNow(void) { return Time(otPlatAlarmMilliGetNow()); }
```
jwhui: What was there reason of moving `GetNow()` from `TimerMilli` to here? I think `TimerMilli` is more appropriate since that is the component that has dependency on `otPlatAlarmMilli*`.

abtink:
The main reason for adding a `Time` class (which simply wraps over the `uint32_t` time tick count) is to ensure the comparison operators (`<`, `>`, `<=` and `>=`) are overloaded and behave correctly when the raw time tick count value is wrapped. 

This blocks a (common) wrong use pattern in the code where two raw time (`uint32_t`) values `t1` and `t2` are compared using `t1 < t2` to determine if time `t1` is before `t2`. 
- For example, if `t1 = 0xfffffffe` and `t2` is 5 ticks ahead at `t2 = 0x3`, the check `t1 < t2` would be `false` where as `t1` should be considered before `t2`.
- Such a use would be wrong only when the raw values wrap. For a millisecond `uint32_t` tick time, the wrap happens every ~50 days, so for a large span of time things would work fine leading to bugs that are tricky to find.

The new `Time` class allows this use pattern by overloading the operator to handle the warping correctly which in turn help us avoid such errors in future code.

-----

Another aspect addressed by this PR is the code pattern used to compare two raw `uint32_t` time tick values using `static_cast<int32_t>(t1 - t2) < 0` to determine if `t1` is before `t2` ([example](https://github.com/openthread/openthread/blob/7d423506932df0e1027e2c9d920ef8e9cfab080b/src/core/thread/mle.hpp#L457)). The type casting from an unsigned to a signed integer is (from C++ standard) an undefined behavior [https://en.cppreference.com/w/cpp/language/implicit_conversion]. How this is handled is left to the implementation of compiler/toolchain. While most toolchains do what we expect (i.e. convert a large `uint32_t` value to a negative `int32_t` value), this is not guaranteed (have been bitten by a toolchain not behaving like this in a previous project). This PR ensures that we don't need to use the `static_cast<int32_t>` thus making the code safer.

src/core/common/time.hpp
```diff
@@ -0,0 +1,263 @@
...
+ *
+ * @brief
+ *   This module includes definitions for the time instance.
+ *
+ * @{
+ *
+ */
+
+/**
+ * This class represent a time instance.
```
jwhui: nit: `represent` -> `represents`

src/core/common/time.hpp
```diff
@@ -0,0 +1,263 @@
...
+     * calculating the duration, `t2 is assumed to be in the past relative to `t1`. The duration calculation correctly
+     * takes into account the wrapping of numeric value of `Time` instances. The returned value can span the entire
+     * range of the `uint32_t` type.
+     *
+     * @param[in]   aOther  A `Time` instance to subtract from.
+     *
+     * @returns The duration of interval from @p aOther to this `Time` object.
+     *
+     */
+    uint32_t operator-(const Time &aOther) const { return mValue - aOther.mValue; }
```
bukepo: Can this return a signed value? This method disabled using result of comparison, which is not natural. Now we have to compare before `-`, which means `-` will actual be computed twice.

abtink: That's sort of intentional.

We want to encourage the use of `<` and `>` to determine if one time is before or after the other one.

The goal for `-` operator is to calculate the duration of interval between a _start_ to an _end_ point in time (which then by its definition assumes _start_ is before _end_) also this way we can use the entire `uint32_t` range when calculating interval/duration between two time stamps.

PR #4196, opened 2019-09-20
[commissioner] change GetNextJoinerInfo from c++ style to c style

huamenggg:


src/core/meshcop/commissioner.cpp
```diff
@@ -287,25 +287,25 @@ otError Commissioner::AddJoiner(const Mac::ExtAddress *aEui64, const char *aPskd
     return error;
 }
 
-otError Commissioner::GetNextJoinerInfo(uint16_t &aIterator, otJoinerInfo &aJoiner) const
+otError Commissioner::GetNextJoinerInfo(uint16_t *aIterator, otJoinerInfo *aJoiner) const
```
abtink: Is there a reason for this change?
The public OT APIs should be C but inside core, we use C++ and should prefer reference over pointers (from style-guide).

huamenggg:
Sorry for misunderstanding.

PR #4207, opened 2019-09-24
[commissioner]  update how Provisioning URL is stored

abtink:
This PR contains a collection of smaller commit changes/improvements in `Commissioner`.

_[commissioner] update how Provisioning URL is stored_
    
This commit updates how the `Commissioner` stores the Provisioning URL
(now saved as null terminated string). It also updates the related
APIs (including the public `otCommissionerGetProvisioningUrl`) and its
use in `NcpBase`.

_[commissioner] iterate over joiners using pointer
[commissioner] make simple getter methods inline
[commissioner] log commissioner state changes_


abtink:
In the new pushed commits, updated msg logs to include the PR number.

PR #4206, opened 2019-09-24
[nrf528xx] change directory structure for nrf528xx family

konradderda:
Merged nrf528xx sub-directories to simplify maintenance and porting of new nrf528xx family platforms in the future.

Crucial changes:
- Moved all source files to common 'src' directory
- Merged all *.c files and treated platform specific fragments of code with conditional compilation
- Created simple automake file [examples/platforms/nrf528xx/Makefile.am] that includes conditional-include blocks

To be done:
- Create single README.md file for all nrf528xx platforms

PR #4210, opened 2019-09-26
[nrf528xx] always enable NRF_MBEDTLS_AES_ALT_INTERRUPT_CONTEXT

konradderda:
If all of COMMISSIONER/JOINER/TIME_SYNC switches are not enabled and CC310 is not disabled, RNG always produces the same values (even after issuing a reset). This is caused by mbed TLS library which by default uses 256 keys internally but CC310 AES module can handle only 128 bit keys. We can enable NRF_MBEDTLS_AES_ALT_INTERRUPT_CONTEXT options for all configurations to enable software implementation when keys longer than 128 bits are used.


abtink:
Thanks for helping with this. 
Can we consider enable `NRF_MBEDTLS_AES_ALT_INTERRUPT_CONTEXT` for all configurations? Is there a down-side to it?

KaSroka:
> @KaSroka please review that.
> 
> I wonder if we need this define anymore since it is hardcoded to 1?

Yes, I think that we should completely remove this `NRF_MBEDTLS_AES_ALT_INTERRUPT_CONTEXT` define so we have this behavior enabled for all configurations, without any need for define. 

@abtink Removing the define completely as I and @LuDuda suggested would result in enabling it for all configurations. I cannot see any downside of this approach - as we would have exactly the same behavior as we have now for configurations with joiner or commissioner enabled.

konradderda:
@LuDuda @KaSroka We need this define to be set up to 1 because of #if directives in crypto library.

KaSroka:
@konradderda `NRF_MBEDTLS_AES_ALT_INTERRUPT_CONTEXT ` is used only in `_alt` files which are written by us so we can change it without any problems, it's not part of MbedTLS library.

konradderda:
@KaSroka I was misguided by the file location (i thought it was a part of nRF SDK) and thought that we would have to maintain #ifs with every update. It it's not a case I will remove #ifs and update the pull request.

konradderda:
@LuDuda @KaSroka I have removed #if directives and verified change with nRF52840 and nRF52811. It seems that everything is fine.

jwhui:
@konradderda, thanks for resolving the comments. FYI - #4206 caused conflicts which need to be resolved before merge.

konradderda:
@jwhui Merge conflict should be resolved now.

PR #4212, opened 2019-09-30
[nrf52840] add UART and BLE bootloader support

e-rk:
This commit introduces serial and BLE DFU bootloader support
* Added new linker script for serial and BLE DFU bootloaders
* Added new nRF52840 switches (BOOTLOADER=USB/UART/BLE)
* Deprecated `BOOTLOADER=1` switch

### DIFF
--- a/examples/Makefile-nrf52840
+++ b/examples/Makefile-nrf52840
@@ -85,8 +85,21 @@ COMMONCFLAGS                      := \
 include $(dir $(abspath $(lastword $(MAKEFILE_LIST))))/common-switches.mk
 
 ifeq ($(BOOTLOADER),1)
-configure_OPTIONS += --with-custom-linker-file=$(AbsTopSourceDir)/examples/platforms/nrf528xx/nrf52840/nrf52840_bootloader.ld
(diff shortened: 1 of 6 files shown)


examples/Makefile-nrf52840
```diff
@@ -85,10 +85,23 @@ COMMONCFLAGS                      := \
 include $(dir $(abspath $(lastword $(MAKEFILE_LIST))))/common-switches.mk
 
 ifeq ($(BOOTLOADER),1)
-configure_OPTIONS += --with-custom-linker-file=$(AbsTopSourceDir)/examples/platforms/nrf528xx/nrf52840/nrf52840_bootloader.ld
+$(info Warning: BOOTLOADER switch is deprecated. Defaulting to BOOTLOADER_USB=1.)
```
KaSroka: Maybe log as a warning? It shouldn't change much but `info Warning` looks strange.

e-rk: The only thing that `warning` does as an extra is print the line number. I don't think that is needed.

e-rk:
Thanks @Vyrastas, I have applied your suggestions.

Vyrastas:
Looks good!

PR #3922, opened 2019-06-14
[efr32] Added mg21 support and changes to mg12 radio inteface code

marvensilabs:
This commit covers two areas, one is the initial support for the MG21.  The other is radio.c changes for the MG12.  This includes a fix for  (#3768) "assert triggered in otPlatRadioTransmit()".

Moved efr32 directory into efr32mg12
New efr32mg21 directory for initial support of mg21

For the efr32mg12:
- changes to radio interface code, checks the txBusy flag before attempting to transmit a packet.  
- Cleaner auto state transition code and safer event handling
- Uses HAL_PA_VOLTAGE in all hal-config.h files

PR #4214, opened 2019-09-30
[efr32mg12] update the channel range of channel config entry

zhanglongxia:
The PR [#3922](https://github.com/openthread/openthread/commit/0bb0b7424#diff-e9cef7a8ce52bf573877847ae819d53dR212) checks if the first channel equals to the minimum channel for the sub-GHz. The minimum channel of sub-GHz is set to 1, the first channel in the channel config entry also should be set to 1 to match the minimum channel of sub-GHz.


zhanglongxia:
@mhallam91 @nigelg-silabs  Could help me review this PR at your convenience.

PR #4216, opened 2019-09-30
[tlvs] use uint32_t for TLV size to avoid overflows

jwhui:


bukepo:
Is this for fixing the situation `GetLength() + sizeof(tlv) > 65535`?

jwhui:
> Is this for fixing the situation `GetLength() + sizeof(tlv) > 65535`?

Yes.

src/core/common/message.cpp
```diff
@@ -404,7 +404,7 @@ otError Message::Append(const void *aBuf, uint16_t aLength)
 
 otError Message::AppendTlv(const Tlv &aTlv)
 {
-    return Append(&aTlv, aTlv.GetSize());
+    return Append(&aTlv, static_cast<uint16_t>(aTlv.GetSize()));
```
bukepo: If GetSize() really overflow uint16_t, it seems this will still cause issues. Shall we assert it is not greater than 65535?

jwhui: The primary issue this PR addresses is processing received TLVs that have length > 65535. I believe the vast majority of issues should be caught in `Tlv::Find()` before we get to `Message::AppendTlv()`. In any case, I have added the assert here to be safe.

PR #4225, opened 2019-10-08
[api] add API to get Thread Version

wgtdkp:
This PR adds the API that getting the Thread Version of current Thread interface.

`Thread Version` is required by border agent to publish its service through mDNS. The value is used to derive the value of key `tv` in a `DNS-SD TXT Answer Record` to tell a commissioner the Thread version of current border agent.

### DIFF
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -219,9 +219,9 @@ void otInstanceFactoryReset(otInstance *aInstance);
 otError otInstanceErasePersistentInfo(otInstance *aInstance);
 
 /**
- * This function gets the Thread stack version.
+ * This function gets the Thread protocol version.
  *
- * @returns the Thread stack version.
+ * @returns the Thread protocol version.
  *
  */
 uint16_t otGetThreadVersion(void);
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -219,6 +219,14 @@ void otInstanceFactoryReset(otInstance *aInstance);
 otError otInstanceErasePersistentInfo(otInstance *aInstance);
 
 /**
+ * This function gets the Thread stack version string.
+ *
+ * @returns A pointer to the Thread stack version.
+ *
+ */
+const char *otGetThreadVersionString(void);
+
+/**
  * This function gets the OpenThread version string.
  *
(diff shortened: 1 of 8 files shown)


wgtdkp:
Should we follow the [Semantic Versioning](https://semver.org/) and return the `Thread Version` in a format of `X.Y.Z`?

bukepo:
@wgtdkp How about adding API to get *Thread Version* instead of *Thread version string*? I think *Thread Version* is well defined and more accurate than the string.

wgtdkp:
> @wgtdkp How about adding API to get _Thread Version_ instead of _Thread version string_? I think _Thread Version_ is well defined and more accurate than the string.

makes sense.

include/openthread/instance.h
```diff
@@ -218,6 +218,14 @@ void otInstanceFactoryReset(otInstance *aInstance);
  */
 otError otInstanceErasePersistentInfo(otInstance *aInstance);
 
+/**
+ * This function gets the Thread stack version.
```
jwhui: nit: "Thread stack version" -> "Thread protocol version"

include/openthread/instance.h
```diff
@@ -218,6 +218,14 @@ void otInstanceFactoryReset(otInstance *aInstance);
...
+ *
+ * @returns the Thread stack version.
```
jwhui: nit: "Thread stack version" -> "Thread protocol version"

include/openthread/instance.h
```diff
@@ -218,6 +218,14 @@ void otInstanceFactoryReset(otInstance *aInstance);
  */
 otError otInstanceErasePersistentInfo(otInstance *aInstance);
 
+/**
+ * This function gets the Thread protocol version.
+ *
+ * @returns the Thread protocol version.
+ *
+ */
+uint16_t otGetThreadVersion(void);
```
jwhui: What about moving this to [include/openthread/thread.h](https://github.com/openthread/openthread/blob/master/include/openthread/thread.h) and be renamed to `otThreadGetVersion()`?

wgtdkp: done.

PR #4232, opened 2019-10-10
[posix] always set ACK FP for data requests when SrcMatch is disabled

simonlingoogle:
The commits fix the issue that when Src Match is disabled, ACK FramePending is never set for data requests , which on the contrary should always be set for data requests.

### DIFF
--- a/examples/platforms/posix/radio.c
+++ b/examples/platforms/posix/radio.c
@@ -177,7 +177,7 @@ static inline bool isDataRequestAndHasFramePending(const uint8_t *aFrame)
     const uint8_t *cur = aFrame;
     uint8_t        securityControl;
     bool           isDataRequest   = false;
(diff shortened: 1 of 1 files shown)


examples/platforms/posix/radio.c
```diff
@@ -177,7 +177,7 @@ static inline bool isDataRequestAndHasFramePending(const uint8_t *aFrame)
     const uint8_t *cur = aFrame;
     uint8_t        securityControl;
     bool           isDataRequest   = false;
-    bool           hasFramePending = false;
+    bool           hasFramePending = true;
```
bukepo: `bool hasFramePending = !sSrcMatchEnabled`?

simonlingoogle: Isn't ` = true` simpler? There is really no actual difference.

PR #4234, opened 2019-10-11
[mle] remove sleepy child in all roles

simonlingoogle:
The commits fix an issue that indirect messages for sleepy child are not correctly cleared in some case. 
In my posix-sim tests, this issue together with the issue fixed in PR #4232 can cause sleepy nodes unable to reattach for a long time, or even forever, as described in following steps:

1. Router R has a sleepy child S which has some indirect messages pending.
1. R changed role to Detached for any reason
2. R changed role to Child
3. R becomes router again with another Router ID
4. R removes the child S by calling `RemoveNeighbor`
    * because now R has `mRole == OT_DEVICE_ROLE_CHILD`, R does not call `ClearAllMessagesForSleepyChild`, failed to clear indirect messages
5. R got another child, which reused S's slot in ChildTable (thus having same ChildIndex)
6. S tries to reattach to R, sends Child ID Request, and get a different ChildIndex
7. S tries to send indirect message (Child ID Response) to S, calling `IndirectSender::AddMessageForSleepyChild -> SourceMatchController::IncrementMessageCount -> AddEntry`
8. In `SourceMatchController::AddEntry`, `AddAddress` returns `OT_ERROR_DUPLICATED` or `OT_ERROR_NO_BUFS`, because S's ext address already exists (was not correctly remove due to step 5). `AddEntry` calls `Enable(false)` to turn off src match in the code bellow.
    * `VerifyOrExit(AddAddress(aChild) == OT_ERROR_NONE, Enable(false));`
9. Since src match is turned off, issue #4232 causes all frame pending not to be set for data requests, causing sleepy child to go to sleep when ACK of Child ID Request is received, thus unable to receive Child ID Response.

Incorrect sleepy message removing also corrupts the data consistency of 
 `SourceMatchController`, causing its behavior unpredictable. 

I am still striving to have a better understanding of this issue and have many questions about removing neighbor. 
 * Beside `ClearAllMessagesForSleepyChild`, do we need to call other functions (listed bellow) too when role is not router/leader? 
    * `Signal(OT_NEIGHBOR_TABLE_EVENT_CHILD_REMOVED, aNeighbor);`
    * `Get<NetworkData::Leader>().SendServerDataNotification(aNeighbor.GetRloc16());`
    * `Get<AddressResolver>().Remove(aNeighbor.GetRloc16());`
    * `RemoveStoredChild(aNeighbor.GetRloc16());`
 * Should children be removed when router is detached?

Thoughts?

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -3132,19 +3132,20 @@ bool MleRouter::IsMinimalChild(uint16_t aRloc16)
 
 void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
 {
-    if (mRole == OT_DEVICE_ROLE_CHILD && &aNeighbor == &mParent)
+    if (&aNeighbor == &mParent)
     {
-        BecomeDetached();
+        if (mRole == OT_DEVICE_ROLE_CHILD)
+        {
+            BecomeDetached();
+        }
     }
-
-    if (!IsActiveRouter(aNeighbor.GetRloc16()))
+    else if (!IsActiveRouter(aNeighbor.GetRloc16()))
     {
         if (aNeighbor.IsStateValidOrRestoring())
         {
             Signal(OT_NEIGHBOR_TABLE_EVENT_CHILD_REMOVED, aNeighbor);
         }
 
-        aNeighbor.SetState(Neighbor::kStateInvalid);
         Get<IndirectSender>().ClearAllMessagesForSleepyChild(static_cast<Child &>(aNeighbor));
         Get<NetworkData::Leader>().SendServerDataNotification(aNeighbor.GetRloc16());
 
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -3132,14 +3132,12 @@ bool MleRouter::IsMinimalChild(uint16_t aRloc16)
 
 void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
 {
-    bool isActiveRouter = IsActiveRouter(aNeighbor.GetRloc16());
-
     if (mRole == OT_DEVICE_ROLE_CHILD && &aNeighbor == &mParent)
     {
         BecomeDetached();
     }
 
-    if (!isActiveRouter)
+    if (!IsActiveRouter(aNeighbor.GetRloc16()))
     {
         if (aNeighbor.IsStateValidOrRestoring())
         {
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -3132,49 +3132,36 @@ bool MleRouter::IsMinimalChild(uint16_t aRloc16)
 
 void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
 {
-    switch (mRole)
+    bool isActiveRouter = IsActiveRouter(aNeighbor.GetRloc16());
+
+    if (mRole == OT_DEVICE_ROLE_CHILD && &aNeighbor == &mParent)
     {
-    case OT_DEVICE_ROLE_DISABLED:
-    case OT_DEVICE_ROLE_DETACHED:
-        break;
+        BecomeDetached();
+    }
 
-    case OT_DEVICE_ROLE_CHILD:
-        if (&aNeighbor == &mParent)
+    if (!isActiveRouter)
+    {
+        if (aNeighbor.IsStateValidOrRestoring())
         {
-            BecomeDetached();
+            Signal(OT_NEIGHBOR_TABLE_EVENT_CHILD_REMOVED, aNeighbor);
         }
 
-        break;
+        aNeighbor.SetState(Neighbor::kStateInvalid);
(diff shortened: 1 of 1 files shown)


jwhui:
@simonlingoogle, thanks for identifying this issue and working to resolve it!

I agree that we need to move most of the other code out of the switch statement (see below).

> * Beside `ClearAllMessagesForSleepyChild`, do we need to call other functions (listed bellow) too when role is not router/leader?
>   
>   * `Signal(OT_NEIGHBOR_TABLE_EVENT_CHILD_REMOVED, aNeighbor);`

Yes.

>   * `Get<NetworkData::Leader>().SendServerDataNotification(aNeighbor.GetRloc16());`

This one is a bit tricky. If R is still a part of the same partition, then yes. If R is no longer a part of the same partition, then no.

>   * `Get<AddressResolver>().Remove(aNeighbor.GetRloc16());`

Yes.

>   * `RemoveStoredChild(aNeighbor.GetRloc16());`

Yes.

> * Should children be removed when router is detached?

No. Thread 1.1.1 Section 5.16.6 indicates that a router should try to keep its children attached, even when migrating to a different partition.

simonlingoogle:
> > * `Get<NetworkData::Leader>().SendServerDataNotification(aNeighbor.GetRloc16());`
> 
> This one is a bit tricky. If R is still a part of the same partition, then yes. If R is no longer a part of the same partition, then no.

@jwhui 
I didn't check if the child being removed belongs to the current partition (neither was before), because of following reasons:
- Implementing it would requires maintaining a partition id for each `Child` entry to record which partition the child belongs to, leading to more RAM footprints
- `SendServerDataNotification` will try to lookup the `rloc16` in the `Network Data`, and only send `/a/sd`  when the `rloc16` is found, which I think is highly unlikely if the child was not of the current partition.

src/core/thread/mle_router.cpp
```diff
@@ -3135,49 +3135,36 @@ bool MleRouter::IsMinimalChild(uint16_t aRloc16)
 
 void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
 {
-    switch (mRole)
+    bool isActiveRouter = IsActiveRouter(aNeighbor.GetRloc16());
```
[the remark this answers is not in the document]
simonlingoogle: nit fixed

src/core/thread/mle_router.cpp
```diff
@@ -3135,49 +3135,34 @@ bool MleRouter::IsMinimalChild(uint16_t aRloc16)
...
+            Get<AddressResolver>().Remove(aNeighbor.GetRloc16());
         }
 
-        break;
+        RemoveStoredChild(aNeighbor.GetRloc16());
+    }
+    else if (aNeighbor.GetState() == Neighbor::kStateValid)
+    {
+        Signal(OT_NEIGHBOR_TABLE_EVENT_ROUTER_REMOVED, aNeighbor);
+        mRouterTable.RemoveNeighbor(static_cast<Router &>(aNeighbor));
```
librasungirl: if the neighbor to be removed is the parent, we may not want to execute this `else` block?

src/core/thread/mle_router.cpp
```diff
@@ -3135,49 +3135,34 @@ bool MleRouter::IsMinimalChild(uint16_t aRloc16)
...
+    if (!IsActiveRouter(aNeighbor.GetRloc16()))
+    {
+        if (aNeighbor.IsStateValidOrRestoring())
         {
-            BecomeDetached();
+            Signal(OT_NEIGHBOR_TABLE_EVENT_CHILD_REMOVED, aNeighbor);
         }
 
-        break;
+        aNeighbor.SetState(Neighbor::kStateInvalid);
```
librasungirl: duplicate with [the last statement](https://github.com/openthread/openthread/pull/4234/files#diff-5a4f49e3956baf8690d055ae8a75a663R3169) 
could be removed?

simonlingoogle: It was duplicate in the original code. 
I didn't dared not to change it.

simonlingoogle: It was duplicate in the original code. 
I didn't dare to change it.

src/core/thread/mle_router.cpp
simonlingoogle: can you explain more about why we may not want to execute the `else` block?

librasungirl: first, original code doesn't do this, right?
second, [mParent](https://github.com/openthread/openthread/blob/master/src/core/thread/mle.hpp#L1648) is one variable independent from [mRouterTable](https://github.com/openthread/openthread/blob/master/src/core/thread/mle_router_ftd.hpp#L769)
It may not make sense to have this `else` block when removing parent.

Thoughts?

src/core/thread/mle_router.cpp
librasungirl: I could not see the possible risk so far :)

src/core/thread/mle_router.cpp
simonlingoogle: changed accordingly.

librasungirl:
@simonlingoogle resolve the conflicts?

PR #4241, opened 2019-10-14
[cli] add hex string payload for `udp send`

librasungirl:


src/cli/cli_udp.cpp
```diff
@@ -172,17 +158,63 @@ otError UdpExample::ProcessSend(int argc, char *argv[])
...
-        error = otMessageAppend(message, argv[curArg], static_cast<uint16_t>(strlen(argv[curArg])));
+        // Generally the rx buffer of uart would be 128.
+        // The length of the hexString in bytes would always be less than 128/2=64.
+        // Here uses one array of 70 bytes to store the payload.
+        uint8_t     buf[70];
+        uint8_t     bufLen  = 0;
+        uint8_t     index   = 0;
+        const char *payload = argv[curArg];
+
+        while (index < strlen(payload) && bufLen < sizeof(buf))
```
simonlingoogle: will `strlen` be calculated every loop?

librasungirl: updated, thanks.

src/cli/cli_udp.cpp
```diff
@@ -136,53 +136,86 @@ otError UdpExample::ProcessSend(int argc, char *argv[])
...
+    if (argc == 2 || argc == 4)
+    {
+        int typePos = curArg++;
+
+        if (strcmp(argv[typePos], "-s") == 0)
+        {
+            long value;
+            payloadType = kTypeAutoSize;
+            SuccessOrExit(error = Interpreter::ParseUnsignedLong(argv[curArg], value));
+            payloadLength = = static_cast<uint16_t>(value);
```
simonlingoogle: it wont compile

src/cli/cli_udp.cpp
```diff
@@ -172,17 +157,65 @@ otError UdpExample::ProcessSend(int argc, char *argv[])
...
+        // The length of the hexString in bytes would always be less than 128/2=64.
+        // Here uses one array of 70 bytes to store the payload.
+        uint8_t     buf[70];
+        uint8_t     bufLen  = 0;
+        uint8_t     index   = 0;
+        const char *payload = argv[curArg];
+
+        while (index < payloadLength && bufLen < sizeof(buf))
+        {
+            buf[bufLen] = ((*(payload + index)) - '0') << 4 | ((*(payload + index + 1)) - '0');
```
wgtdkp: 1. This doesn't cover `a-f` and `A-F`;
2. Should use `Interpreter::Hex2Bin()` instead;

src/cli/cli_udp.hpp
```diff
@@ -74,6 +74,13 @@ class UdpExample
         otError (UdpExample::*mCommand)(int argc, char *argv[]);
     };
 
+    enum PayloadType
```
wgtdkp: Is this necessary?

src/cli/cli_udp.cpp
```diff
@@ -172,17 +157,65 @@ otError UdpExample::ProcessSend(int argc, char *argv[])
...
 
+    if (argc == 2 || argc == 4)
+    {
+        int typePos = curArg++;
+
+        if (strcmp(argv[typePos], "-s") == 0)
+        {
+            unsigned long value;
+            payloadType = kTypeAutoSize;
+            SuccessOrExit(error = Interpreter::ParseUnsignedLong(argv[curArg], value));
```
simonlingoogle: when `argc == 3`, this code will access the end of `argv`, i.e. "\0", and ParseUnsignedLong("\0") will return OT_ERROR_NONE with value=0, which I think is also a minor issue.

src/cli/cli_udp.cpp
```diff
@@ -136,33 +136,18 @@ otError UdpExample::ProcessSend(int argc, char *argv[])
...
-    unsigned long autoGenPayloadLength = 0;
+    otMessage *   message       = NULL;
+    int           curArg        = 0;
+    uint16_t      payloadLength = 0;
+    PayloadType   payloadType   = kTypeText;
 
     memset(&messageInfo, 0, sizeof(messageInfo));
 
-    VerifyOrExit(argc == 1 || argc == 3 || argc == 4, error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(argc >= 1 && argc <= 4, error = OT_ERROR_INVALID_ARGS);
```
simonlingoogle: do we actually only support argc == 4 ?

src/cli/cli_udp.cpp
librasungirl: thanks, updated.

src/cli/cli_udp.cpp
librasungirl: I don't think it will happen as this block has the condition of `argc == 2 || argc == 4`.

src/cli/cli_udp.cpp
librasungirl: argc = [1,2,3,4]

src/cli/cli_udp.hpp
librasungirl: prefer to keep it is for readability

src/cli/cli_udp.cpp
librasungirl: updated the doc to reflect all supported `argc` scenarios. thanks

simonlingoogle:
LGTM+1

src/cli/README_UDP.md
```diff
@@ -113,20 +113,32 @@ Send a UDP message.
...
 
 Send a few bytes over UDP.
 
-* message_size: number of bytes to be send.
 * ip: the IPv6 destination address.
 * port: the UDP destination port.
+* type: the type of the message:
+        * `-t`: text payload in the `value`, same as without specifying the type. 
+        * `-s`: autogenerated payload with specified length indicated in the `value`.  
+        * `-h`: hex string payload in the `value`
```
bukepo: prefer "-x", "-h" is usually for help.

librasungirl: updated. thx

src/cli/cli_udp.cpp
```diff
@@ -172,17 +157,58 @@ otError UdpExample::ProcessSend(int argc, char *argv[])
...
+        break;
+    case kTypeAutoSize:
+        error = WriteCharToBuffer(message, payloadLength);
+        break;
+    case kTypeHexString:
+    {
+        // Generally the rx buffer of uart would be 128.
+        // The length of the hexString in bytes would always be less than 128/2=64.
+        // Here uses one array of 70 bytes to store the payload.
+        uint8_t buf[70];
```
jwhui: We could support arbitrary length hex string with a fixed buffer size by wrapping this in a loop and parsing/appending in chunks. Thoughts?

librasungirl: Updated. Please help to have a review. Thanks.

jwhui:
@librasungirl, thanks for the update. It seems that scan-build is failing, however.

src/cli/cli.cpp
```diff
@@ -261,16 +261,20 @@ Interpreter::Interpreter(Instance *aInstance)
 #endif // OPENTHREAD_CONFIG_DNS_CLIENT_ENABLE
 }
 
-int Interpreter::Hex2Bin(const char *aHex, uint8_t *aBin, uint16_t aBinLength)
+int Interpreter::Hex2Bin(const char *aHex, uint8_t *aBin, uint16_t aBinLength, bool aAllowTruncate)
```
bukepo: Is it necessary to add `aAllowTruncate`? I think we can always allow this. The caller can tell if it's truncated by from the value.

src/cli/cli_udp.cpp
```diff
@@ -172,17 +157,70 @@ otError UdpExample::ProcessSend(int argc, char *argv[])
...
+        break;
+    case kTypeHexString:
+    {
+        uint8_t     buf[50];
+        int8_t      bufLen;
+        const char *hexIndex         = argv[curArg];
+        uint16_t    remainHexLen     = static_cast<uint16_t>(strlen(argv[curArg]));
+        uint16_t    conversionHexLen = 0;
+
+        while (remainHexLen > 0)
```
bukepo: verify `remainHexLen` is even, otherwise error invalid argument?

src/cli/cli_udp.cpp
```diff
@@ -172,17 +157,70 @@ otError UdpExample::ProcessSend(int argc, char *argv[])
...
+            {
+                conversionHexLen -= 1;
+            }
+
+            hexIndex += conversionHexLen;
+            remainHexLen -= conversionHexLen;
+            SuccessOrExit(error = otMessageAppend(message, buf, static_cast<uint16_t>(bufLen)));
+        }
     }
+    break;
```
bukepo: move break into the `}`

src/cli/cli_udp.cpp
```diff
@@ -172,17 +157,70 @@ otError UdpExample::ProcessSend(int argc, char *argv[])
...
+    default:
```
bukepo: no need for this? The cases already covered all enumerations.

src/cli/README_UDP.md
```diff
@@ -135,4 +147,25 @@ Send a UDP message on a connected socket.
...
+```
+
+### send \<type\> \<value\>
+
+Send a few bytes over UDP.
+
+* type: the type of the message:
+        * `-t`: text payload in the `value`, same as without specifying the type.
+        * `-s`: autogenerated payload with specified length indicated in the `value`.
+        * `-x`: hex string payload in the `value`
```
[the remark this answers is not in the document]
librasungirl: updated

src/cli/cli_udp.cpp
librasungirl: the design of existing Hex2Bin is to allow odd and prepend 0 if odd

src/cli/cli_udp.cpp
librasungirl: updated.

src/cli/cli_udp.cpp
librasungirl: updated.

src/cli/cli.hpp
```diff
@@ -147,13 +147,16 @@ class Interpreter
     /**
      * This method converts a hex string to binary.
      *
-     * @param[in]   aHex        A pointer to the hex string.
-     * @param[out]  aBin        A pointer to where the binary representation is placed.
-     * @param[in]   aBinLength  Maximum length of the binary representation.
+     * @param[in]   aHex            A pointer to the hex string.
+     * @param[out]  aBin            A pointer to where the binary representation is placed.
+     * @param[in]   aBinLength      Maximum length of the binary representation.
+     * @param[in]   aAllowTruncate  Whether or not allow @aBin less than the required buffer
```
jwhui: suggestion: "TRUE if @p aBin must have even length, FALSE otherwise."

PR #4243, opened 2019-10-15
[timer] add helper methods for common timer use patterns and simplify…

abtink:
This PR adds helper methods in `Timer` class which cover commonly repeated use patterns by other modules.
    
Method `StartForFireTime()` allows a timer to be started with a given fire time. `RestartIfEarlierFireTime()` (re-)schedules the timer with  a given a fire time only if the timer is not running or if the new given fire time is earlier than the current fire time. This pattern is used in bunch of modules.
    
Modules mle, coap, ip6-mpl, joiner-router, dns-clinet, and sntp-client are updated to use the new methods.

-------

### DIFF
--- a/src/core/coap/coap.cpp
+++ b/src/core/coap/coap.cpp
@@ -259,41 +259,33 @@ void CoapBase::HandleRetransmissionTimer(Timer &aTimer)
 
 void CoapBase::HandleRetransmissionTimer(void)
 {
-    TimeMilli        now       = TimerMilli::GetNow();
-    uint32_t         nextDelta = TimeMilli::kMaxDuration;
+    TimeMilli        now      = TimerMilli::GetNow();
+    TimeMilli        nextTime = now.GetDistantFuture();
     CoapMetadata     coapMetadata;
-    Message *        message     = static_cast<Message *>(mPendingRequests.GetHead());
-    Message *        nextMessage = NULL;
+    Message *        message;
+    Message *        nextMessage;
     Ip6::MessageInfo messageInfo;
 
-    while (message != NULL)
+    for (message = static_cast<Message *>(mPendingRequests.GetHead()); message != NULL; message = nextMessage)
     {
         nextMessage = static_cast<Message *>(message->GetNext());
+
         coapMetadata.ReadFrom(*message);
 
-        if (coapMetadata.IsLater(now))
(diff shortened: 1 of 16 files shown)


src/core/common/timer.cpp
```diff
@@ -220,7 +233,12 @@ void                           TimerMicro::Start(uint32_t aDelay)
 void TimerMicro::StartAt(TimeMicro aStartTime, uint32_t aDelay)
 {
     assert(aDelay <= kMaxDelay);
-    mFireTime = aStartTime + aDelay;
+    StartForFireTime(aStartTime + aDelay);
+}
+
+void TimerMicro::StartForFireTime(TimeMicro aFireTime)
```
jwhui: nit: how about renaming to `FireAt()`?

src/core/common/timer.hpp
```diff
@@ -175,12 +175,29 @@ class TimerMilli : public Timer
...
     void Stop(void);
 
+    /**
+     * This method (re-)schedules the timer with a given a fire time only if the timer is not running or the new given
+     * fire time is earlier than the current fire time.
+     *
+     * @param[in]  aFireTime  The fire time.
+     *
+     */
+    void RestartIfEarlierFireTime(TimeMilli aFireTime);
```
jwhui: nit: how about `FireAtIfEarlier`

src/core/coap/coap.cpp
```diff
@@ -259,42 +259,34 @@ void CoapBase::HandleRetransmissionTimer(Timer &aTimer)
 
 void CoapBase::HandleRetransmissionTimer(void)
```
abtink: The diff here looks bigger than the actual change. Here are some points on what is changed:
- The `while()` iteration loop is changed to a ` for()`: this allows us to use `continue` to skip over entries.
- We move up the check for entries that are expired and use `continue` to skip over next entry in loop if no need to retx the entry.
- In the case that retx is required for an expired entry, the `mNextTimerShot` gets updated.
- The code which updates the next timer fire time (`nextTime` variable) is now moved at the end and is common for both cases (where entry is not expired or where the entry is expired and we need a retx).
- After the loop, we check if the we need to start the timer with ` nextTime`.

A similar pattern is used in other modules.

PR #4247, opened 2019-10-15
[coap] add CoAP(s) block-wise transfer support to CoAP(s) client and server

AuSchell:
Implemented CoAP/-s block-wise transfer according to RFC 7959. The implementation includes methods and functions available to the application layer to add Block1 and Block2 option calues to a coap message and an automatic block-wise transfer enabled by the compile switch "COAP_BLOCK".


sjlongland:
One thought on this, have you given any consideration to "larger than RAM" request entities?  I wonder if a call-back interface whereby a block handler routine is called with each block to allow the application to figure out where to stash or retrieve the next block.

Our biggest use case for block-wise transfer is OTA firmware updates, whereby the entire 512kB firmware image for the CC2538 is transferred block-wise over the mesh.  The idea being the downloaded blocks are written to a SPI flash chip which is then accessible to the boot-loader.

A 512kB firmware image is never going to fit in the 32kB of RAM available on a CC2538 no matter what you do, and on these devices, 4kB just set aside for block re-assembly is a lot of RAM!

We find we can support up to 5 devices transferring images in 256-byte blocks can do a full firmware update in about 10-15 minutes via this method, and the code has been very resilient to sudden power loss, tampering or corruption.

There's definitely merit in the approach taken here, would love to assist in seeing if we can generalise it just a little bit. :-)

src/core/coap/coap_message.cpp
```diff
@@ -192,6 +198,49 @@ otError Message::AppendProxyUriOption(const char *aProxyUri)
...
+        buf[1] = (optionValue & 0x0000ff00) >> 8;
+        buf[2] = optionValue & 0x000000ff;
+    }
+    else
+    {
+        error = OT_ERROR_INVALID_ARGS;
+        ExitNow();
+    }
+
+    error = AppendOption(aBlockType, length, buf);
```
sjlongland: I'd have a look at `AppendUintOption` for this:
https://github.com/openthread/openthread/blob/master/src/core/coap/coap_message.hpp#L286-L298

src/core/coap/coap_message.cpp
```diff
@@ -322,6 +382,137 @@ otError Message::GetOptionValue(void *aValue) const
...
+{
+    otError error                     = OT_ERROR_NONE;
+    uint8_t buf[kMaxOptionHeaderSize] = {0};
+
+    SuccessOrExit(error = GetOptionValue(buf));
+
+    *aBlockNumber = 0;
+    *aMoreBlocks  = false;
+
+    switch (aOptionLength)
```
sjlongland: This would cover most use cases, but it seems this is a repetitive way to go about it.

This routine could benefit greatly from https://github.com/openthread/openthread/pull/4396/commits/f17e2edb663cac04b8c61cee8dbf27fae04e3224.

src/core/coap/coap_secure.cpp
```diff
@@ -158,6 +158,13 @@ otError CoapSecure::SendMessage(Message &aMessage, otCoapResponseHandler aHandle
 
     VerifyOrExit(IsConnected(), error = OT_ERROR_INVALID_STATE);
 
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+    if (CoapBase::GetMaxBlockSize() > OT_COAP_OPTION_BLOCK_LENGTH_512)
```
sjlongland: Why are we limiting the block size to 512 bytes?  1024 byte blocks do work. :-)

sjlongland:
Just thinking about the handling of blocks… does this code as it stands now handle multiple simultaneous block-wise transfers?

If no, then perhaps a way forward is we add some routines that:
1. allow the application to declare a block of RAM somewhere in which to store a block-wise transferred entity.
2. allow the application to override the block retrieval/storage handlers for handling of larger entities that would otherwise fit in RAM.

There could be some benefit in defining some of those magic numbers in `include/openthread/coap.h` too.  That way someone looking through the code isn't left wondering what they are.

src/core/coap/coap_message.cpp
AuSchell: thanks for the tip, that saves some line of codes.

src/core/coap/coap_secure.cpp
AuSchell: That is actually a leftover from a test i've run, thanks, I've corrected it back to 1024 byte block-size.

src/core/coap/coap_message.cpp
AuSchell: That would have also saved some lines of code, unfortunatly, at the time of writing, this feature has already been removed again.

AuSchell:
@sjlongland Thank you for your review, I'm sorry for the long response time, I've been busy with other projects. As you can see, I've implemented your suggestions for the code.

AuSchell:
> One thought on this, have you given any consideration to "larger than RAM" request entities? I wonder if a call-back interface whereby a block handler routine is called with each block to allow the application to figure out where to stash or retrieve the next block.
>
> Our biggest use case for block-wise transfer is OTA firmware updates, whereby the entire 512kB firmware image for the CC2538 is transferred block-wise over the mesh. The idea being the downloaded blocks are written to a SPI flash chip which is then accessible to the boot-loader.
> 
> A 512kB firmware image is never going to fit in the 32kB of RAM available on a CC2538 no matter what you do, and on these devices, 4kB just set aside for block re-assembly is a lot of RAM!
> 
> We find we can support up to 5 devices transferring images in 256-byte blocks can do a full firmware update in about 10-15 minutes via this method, and the code has been very resilient to sudden power loss, tampering or corruption.

I've not thought of such a use case, but it sounds interesting. Many applications with constrained devices would benefit from such a feature. Unfortunately, I've no experience in this field. 

The 4kB restriction is just a guess so that i had something to work with. I was unsure about how to store the received blocks (static vs. dynamic memory allocation), so i've used static memory allocation and awaited the answer.

AuSchell:
> Just thinking about the handling of blocks… does this code as it stands now handle multiple simultaneous block-wise transfers?

No, it does not, at least not with just one Thread instance. The intention for this first version of the feature was to implement it as general as possible, while still being able to use it with our application. So I've focused on a minimal use case with just one block-wise transfer at the time to avoid complications.

> If no, then perhaps a way forward is we add some routines that:
> 
> 1. allow the application to declare a block of RAM somewhere in which to store a block-wise transferred entity.
> 2. allow the application to override the block retrieval/storage handlers for handling of larger entities that would otherwise fit in RAM.
> 
> There could be some benefit in defining some of those magic numbers in `include/openthread/coap.h` too. That way someone looking through the code isn't left wondering what they are.

Which numbers do you mean?

src/core/coap/coap_message.cpp
```diff
@@ -207,6 +228,154 @@ otError Message::AppendUriQueryOption(const char *aUriQuery)
...
+    SuccessOrExit(error = iterator.GetOptionValue(buf));
+
+    *aBlockNumber = 0;
+    *aMoreBlocks  = false;
+
+    switch (option->mLength)
+    {
+    case 1:
+        *aBlockNumber = (buf[0] & 0xf0) >> 4;
+        if ((buf[0] & 0x08) >> 3 == 1)
```
sjlongland: I mentioned about "magic numbers"…

`0x08`, `3`… there's a few numbers on this line that have no apparent meaning without reading the specs.  (_I_ know `0x08` here is the `M` bit, but a casual reader may not!).  PR #4396 adds a routine that will do the donkey work of converting that option to an unsigned integer so you'd be left with:

```
    otError Message::ReadBlockOptionValues(…)
    {
        uint64_t rawValue;
        OptionIterator iterator;
        // …

        SuccessOrExit(error = iterator.Init(this));
        VerifyOrExit((option = iterator.GetFirstOptionMatching(aBlockType)) != NULL, error = OT_ERROR_NOT_FOUND);
        SuccessOrExit(error = iterator.GetOptionValue(rawValue));

        *aBlockNumber = rawValue >> 4;
        if ((rawValue & 0x08) >> 3 == 1)
        {
            *aMoreBlocks = true;
        }
        *aBlockSize = (otCoapOptionBlockSize)(rawValue & 0x07); // and this should probably be a static_cast<>
    }
```

We cut some repetition, but note the magic numbers still there.

As a counter-example, have a look at this:
https://github.com/vrtsystems/openthread/commit/7f0e7f1b638d3faa7cde49425ebed2f891160a92

There, I've put the magic numbers in a macro which is named.  Arguably, I could do a bit more.  Those last few lines would collapse to this:

```
        *aBlockNumber = OT_COAP_BLOCK_NUM(rawValue);
        *aMoreBlocks = OT_COAP_BLOCK_MORE(rawValue) != 0;
        *aBlockSize = (otCoapOptionBlockSize)(OT_COAP_BLOCK_SZX(rawValue));
```

sjlongland:
Okay, so if we're just handling a single transfer at a time, then it's possibly acceptable to have a `otCoapBlockwiseReceiveHook` `typedef` which is a function that gets called when a block is received.

Then the developer can simply set the hook to point to a function of their choosing before initiating the request.  `malloc` or static allocation: their problem to deal with.  Maybe something like:

```
typedef otError (otCoapBlockwiseReceiveHook)(const uint8_t *aBlock, uint32_t aPosition, uint16_t aBlockLength, uint32_t aTotalLength);
```
(Yes, I might have my syntax wrong… but you get the gist I hope. :-)

* `aBlock` would contain the block that was just received
* `aPosition` would be the offset into the request entity where that block "belongs"
* `aBlockLength` would be the length of `aBlock`
* `aTotalLength` would be set from `Size2` if given, indicating the total size.

Similarly, for sending large request entities, you'd do a similar `typedef` (maybe `otCoapBlockwiseTransmitHook`), but `aBlock` would be a `uint8_t *` not a `const uint8_t *`… and maybe you'd do away with `aTotalLength`.  (The caller can do that themselves in the initial request.)

These pointers would probably live in `ot::Coap::Coap`, although one could also put them in `ot::Coap::Message` and thus be able to handle multiple parallel CoAP blockwise transfers with minimal effort.

In my OTA example, that receive hook would be writing to the external SPI flash.  Or it could be a (e)MMC card, etc, it doesn't matter. :-)

AuSchell:
> Okay, so if we're just handling a single transfer at a time, then it's possibly acceptable to have a otCoapBlockwiseReceiveHook typedef which is a function that gets called when a block is received.

I get the gist and I'm in for implementing, but there are some unsloved problems I would like to discuss:

For otCoapBlockwiseReceiveHook, I think there is no sophisticated way to conclude a blockwise transfer. Currently, the transfer is concluded on the CoAP layer by getting a new message from the message pool and attaching the content of the receive buffer and copy the header of the last block message (minus the blockwise option) to it. When the receive or response handler is called, it looks like the message has been received as a whole and not fragmented. 

If the hook is used, the CoAP layer has no knowledge where the blocks are stored and therefore the process can not conclude like described above. So. here's how I would implement it:

1. otCoapBlockwiseReceiveHook can be set when sending a GET/POST request or when creating a CoAP resource. The pointer to the otCoapBlockwiseReceiveHook is stored in the CoapMetadata instance associated with a sent request. This allows us to handle multiple parallel CoAP blockwise transfers. If no otCoapBlockwiseReceiveHook is set, the block messages will be forwarded like normal messages.

2. The buffers mReassemblyMessage and mDisassemblyMessage will be removed and the payload of a block message is forwarded to otCoapBlockwiseReceiveHook instead. The CoAP layer is no longer responsible for verifying that all blocks have been received and that the order is correct. The CoAP layer will send the appropriate CoAP error code if otCoapBlockwiseReceiveHook returns an error.

3. If the last block message of a transfer has been received, this block message is forwarded to the receive/response handler unaltered (after the payload has been forwarded to otCoapBlockwiseReceiveHook). 

In the end, the user is responsible to link the CoAP header of the last block message with the reassembled payload. And this is the point where I feel like something is missing. The user will have the payload wherever the user stored it and the complete last block message, with the last CoAP block option in its header and the content (that is already stored) of the last block in its payload. It just doesn't feel right to me. 

What do you think? Do you have an idea how we could cut this last edge or do you think this is something we can live with? Because everything else is fine, it will work. The only thing that won't work anymore is if an unexpected blockwise transfere occurs and no hook function is set, then the block messages are forwarded like normal messages.

sjlongland:
That sounds perfectly fine.  It's true that we can't do the nice thing of saying "here's your entire request entity" because not only do we not know where it was placed, but indeed it may not be in directly addressable memory.  (e.g. it may be on external SPI memory storage.)  As these chips typically don't have a MMU, there's basically no way for us to provide a pointer to such storage.

I think it is just fine to send to the original request handler, the final block.  Alternatively, that `typedef` could also include a `Bool aMore` parameter, which is set according to the `M` bit in the block header.  If that parameter is `false`, then we know this is the last block.

One thing I didn't mention was that in the suggested `typedef` signature above, that hook returned an `otError`, so there's scope there for the block-wise transfer to be aborted if it returns something other than `OT_ERROR_NONE`.  (By abort, we simply don't request the next block.  CoAP server should handle that, because a client reset could happen at any time.)

AuSchell:
Hello @sjlongland, I would like to ask for your help to solve the following problem:

How do we initiate the blockwise transfer on the device transmitting the payload? 

Here is the modified `SendMessage` function:

```
otError SendMessage(Message &aMessage, const Ip6::MessageInfo &aMessageInfo, otCoapResponseHandler aHandler = NULL, void *aContext = NULL, otCoapBlockwiseReceiveHook aReceiveHook = NULL, otCoapBlockwiseTransmitHook aTransmitHook = NULL)
```

One way i could think of is that aMessage includes just the header, including options (excluding blockwise option) with  `aReceiveHook` and/or `aTransmitHook` set respectivly. In the function `CoapBase::SendMessage` it is detected that an empyt message wants to be send and that a blockwise hook is set, therfore the payload to send is received via hook. 

Or the user has to set up the first block message by himself.

I prefer the first option, but where is the best place to explain how to use the modified `SendMessage` function?

sjlongland:
Apologies for not getting back to you… been busy lately.  Thinking about this, there's a can of worms where generating messages on the user's behalf comes in, and that's the fact that options must be given in a specific order.

The user can theoretically have any of them set in their initial message, and will likely assume the will be there on subsequent requests.  Unless we want to duplicate the interface for setting up that initial message, my thinking would be to use `otCoapSendRequest` as it is now.

Basically there'll be two blobs with "initial" options and "subsequent" options; which will sit either side of the `Block1` or `Block2` option.

For incoming block-wise requests (`Block2`), this isn't too bad.  User specifies the options in the first request, and the CoAP code can take it from there.

For outgoing block-wise requests (`Block1`) things get more tricky: who has responsibility of generating the blocks themselves?  We've got two options I can think of now:

1. let the user generate the first block, then the call-back hook takes it from there
2. let the user specify the request options, but omit the payload and let the code in `otCoapSendRequest` notice the `Block1` option, and query the call-back function for the first block.

I'm thinking the second option feels more "natural", you're telling it (via the message options) that there _will_ be a series of blocks to be sent, and it then uses that routine to begin requesting each block in turn.  How the block is retrieved doesn't change, whereas to go with option 1, you've got to re-implement that first step yourself.  That first message sent by the API caller just serves as the template, and container, for the options that will be used for the _actual_ messages sent to the CoAP peer.

AuSchell:
@sjlongland Hello, i don't know if this "re-request review"-button works, so i just let you know that my code is ready for review.

sjlongland:
@AuSchell No problems… I've been flat out at work (working from home) but I have a moment spare now so let's have a look. :-)

include/openthread/coap.h
```diff
@@ -755,6 +804,9 @@ otError otCoapOptionIteratorGetOptionUintValue(otCoapOptionIterator *aIterator,
  */
 otError otCoapOptionIteratorGetOptionValue(otCoapOptionIterator *aIterator, void *aValue);
 
+// TODO
```
sjlongland, line 807: `// TODO`?

src/core/coap/coap.cpp
```diff
@@ -456,15 +628,465 @@ void CoapBase::DequeueMessage(Message &aMessage)
...
+            response->SetBlockWiseBlockSize(OT_COAP_OPTION_BLOCK_SZX_32);
+            break;
+        case 16:
+            response->SetBlockWiseBlockSize(OT_COAP_OPTION_BLOCK_SZX_16);
+            break;
+        default:
+            error = OT_ERROR_INVALID_ARGS;
+            ExitNow();
+            break;
+        }
```
sjlongland, line 1015: Just thinking about this block… if I understand what this is doing correctly, this is allowing the transmit hook to change the block size.  I'm not sure if this is only intended to be used on the very first block, it does look like this would allow the hook to change the block size mid-flight, which could be a risky thing to do with some CoAP implementations.

It might be worth setting this some point earlier in the request sequence just once… that or if that's not possible, we should make it very clear to developers that they should not change `bufLen` from inside their transmit hooks after the first block has been sent (and thus block-size has been negotiated).

sjlongland:
I had a look, it's looking great so far.  There's a small conflict that needs fixing (looks like it'll be a one-liner).  I think this code could work real well.

On smaller devices, we might need to carefully consider `OPENTHREAD_CONFIG_COAP_MAX_BLOCK_LENGTH`: as `CoapBase::ProcessBlock2Request` allocates the block on the stack, and the block size defaults to 1kB, which if your entire stack is 4kB, that's huge.

I guess if that becomes a problem, we'll cross that bridge then.  In production we've found 256-byte blocks quite workable even on a busy Thread network, so on smaller ports, maybe we just limit it to that for now. :-)  OTA updates of a 512kB firmware image sent in 256-byte blocks takes about 5-10 minutes and usually we can update about 10 nodes at a time that way.

Looking forward to its final inclusion. :-)

include/openthread/coap.h
AuSchell: Was a reminder to do the documentation for the method... whose utilization has been implemented in the mean time XD.

src/core/coap/coap.cpp
AuSchell: I've added a warning to the description of the receive hook. Otherwise i would have to change the block size with no means of informing the user.

AuSchell:
> @AuSchell No problems… I've been flat out at work (working from home) but I have a moment spare now so let's have a look. :-)

same here. :-\
but did it the button work? there isn't really a feedback like a highliting, a short message, something that lets me know the request has been issued.

sjlongland:
Not sure… there's a lot I'm learning about Github.  I basically saw your comments and decided to have another look. :-)

Of course, it's worth noting my review is far from authoritative as I'm not one of the core OpenThread devs.

include/openthread/coap.h
```diff
@@ -875,6 +990,25 @@ void otCoapRemoveResource(otInstance *aInstance, otCoapResource *aResource);
  */
 void otCoapSetDefaultHandler(otInstance *aInstance, otCoapRequestHandler aHandler, void *aContext);
 
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+/**
+ * This method sets the maximum block size for CoAP/-s Block-Wise Transfer.
+ *
+ * @param[in]  aInstance  A pointer to an OpenThread instance.
+ * @param[in]  aBlockSize The maximal allowed block size.
+ */
```
jwhui, line 999: nit:
```suggestion
 *
 */
```

include/openthread/coap.h
```diff
@@ -875,6 +990,25 @@ void otCoapRemoveResource(otInstance *aInstance, otCoapResource *aResource);
...
+void otCoapSetMaxBlockSize(otInstance *aInstance, otCoapBlockSzx aBlockSize);
+
+/**
+ * This method gets the maximum block size for CoAP/-s Block-Wise Transfer.
+ *
+ * @param[in]  aInstance  A pointer to an OpenThread instance.
+ *
+ * @retval The current set maximum block size.
+ */
```
jwhui, line 1008: nit:
```suggestion
 *
 */
```

include/openthread/coap.h
```diff
@@ -875,6 +990,25 @@ void otCoapRemoveResource(otInstance *aInstance, otCoapResource *aResource);
...
+ * @param[in]  aBlockSize The maximal allowed block size.
+ */
+void otCoapSetMaxBlockSize(otInstance *aInstance, otCoapBlockSzx aBlockSize);
+
+/**
+ * This method gets the maximum block size for CoAP/-s Block-Wise Transfer.
+ *
+ * @param[in]  aInstance  A pointer to an OpenThread instance.
+ *
+ * @retval The current set maximum block size.
```
jwhui, line 1007: nit:
```suggestion
 * @returns The current set maximum block size.
```

src/core/api/coap_api.cpp
```diff
@@ -211,6 +211,38 @@ otError otCoapOptionIteratorGetOptionValue(otCoapOptionIterator *aIterator, void
...
+        *static_cast<Coap::Message *>(aMessage), *static_cast<const Ip6::MessageInfo *>(aMessageInfo),
+        Coap::TxParameters::From(aTxParameters), aHandler, aContext, aTransmitHook, aReceiveHook);
+}
+
+otError otCoapSendRequestWithParameters(otInstance *              aInstance,
+                                        otMessage *               aMessage,
+                                        const otMessageInfo *     aMessageInfo,
+                                        otCoapResponseHandler     aHandler,
+                                        void *                    aContext,
+                                        const otCoapTxParameters *aTxParameters)
```
jwhui, line 236: Should we validate `aTxParameters`?

src/core/api/coap_api.cpp
```diff
@@ -211,6 +211,38 @@ otError otCoapOptionIteratorGetOptionValue(otCoapOptionIterator *aIterator, void
...
+{
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
+    return instance.GetApplicationCoap().SendMessage(
```
jwhui, line 240: Do we need to special case this? It seems `SendMessage()` already has default `NULL` values for the blockwise params.

src/core/coap/coap.cpp
```diff
@@ -129,6 +134,165 @@ otError CoapBase::Send(ot::Message &aMessage, const Ip6::MessageInfo &aMessageIn
...
+        // As we do not retransmit non confirmable messages, create a
+        // copy of header only, for token information.
+        copyLength = aMessage.GetOptionStart();
+    }
+
+    if (copyLength > 0)
+    {
+        Metadata metadata;
+
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
```
jwhui, line 228: Can we avoid duplicating this code? Is it possible to integrate the blockwise code into the existing `SendMessage()` method?

src/core/coap/coap.cpp
```diff
@@ -456,15 +628,465 @@ void CoapBase::DequeueMessage(Message &aMessage)
...
+}
+
+otError CoapBase::CacheLastBlockResponse(Message *aResponse)
+{
+    // Save last response for block-wise transfer
+    FreeLastBlockResponse();
+
+    if ((mLastResponse = aResponse->Clone()) != NULL)
+    {
+        return OT_ERROR_NONE;
```
jwhui, line 659: nit: prefer single return statement at end.

src/core/coap/coap.cpp
```diff
@@ -456,15 +628,465 @@ void CoapBase::DequeueMessage(Message &aMessage)
...
+                  otCoapBlockSizeFromExponent(aMessage.GetBlockWiseBlockSize()), aMessage.IsMoreBlocksFlagSet());
+
+    // Conclude block-wise transfer if last block has been received
+    if (!aMessage.IsMoreBlocksFlagSet())
+    {
+        FinalizeCoapTransaction(aRequest, aCoapMetadata, &aMessage, &aMessageInfo, OT_ERROR_NONE);
+        ExitNow();
+    }
+
+    // Init request for next block
```
jwhui, line 816: Does this duplicate a bunch of code from `SendNextBlock1Request()`? If so, can we try to avoid the duplication?

src/core/coap/coap.cpp
```diff
@@ -456,15 +628,465 @@ void CoapBase::DequeueMessage(Message &aMessage)
...
+    SuccessOrExit(error = iterator.Init(&aRequest));
+
+    // Init request for next block
+    VerifyOrExit((request = NewMessage()) != NULL, error = OT_ERROR_NO_BUFS);
+    request->Init(OT_COAP_TYPE_CONFIRMABLE, aRequest.GetCode());
+
+    // Copy options from last response to next message
+    for (const otCoapOption *option = iterator.GetFirstOption(); option != NULL; option = iterator.GetNextOption())
+    {
+        VerifyOrExit((optionBuf = (uint8_t *)calloc(option->mLength, sizeof(uint8_t))) != NULL,
```
jwhui, line 713: We generally try to avoid requiring a heap. Is it possible to avoid here?

src/core/coap/coap.cpp
```diff
@@ -456,15 +628,465 @@ void CoapBase::DequeueMessage(Message &aMessage)
...
+    bufLen = aMessage.Read(aMessage.GetOffset(), aMessage.GetLength() - aMessage.GetOffset(), buf);
+    SuccessOrExit(error = aResource.mReceiveHook(aResource.mContext, buf,
+                                                 otCoapBlockSizeFromExponent(aMessage.GetBlockWiseBlockSize()) *
+                                                     aMessage.GetBlockWiseBlockNumber(),
+                                                 bufLen, aMessage.IsMoreBlocksFlagSet(), aTotalLength));
+
+    if (aMessage.IsMoreBlocksFlagSet())
+    {
+        // Set up next response
+        VerifyOrExit((response = NewMessage()) != NULL, error = OT_ERROR_FAILED);
```
jwhui, line 916: Is `OT_ERROR_NO_BUFS` more appropriate here?
```suggestion
        VerifyOrExit((response = NewMessage()) != NULL, error = OT_ERROR_NO_BUFS);
```

src/core/coap/coap.cpp
```diff
@@ -456,15 +628,465 @@ void CoapBase::DequeueMessage(Message &aMessage)
...
+    response->SetBlockWiseBlockNumber(
+        (otCoapBlockSizeFromExponent(aMessage.GetBlockWiseBlockSize()) * aMessage.GetBlockWiseBlockNumber()) /
+        (otCoapBlockSizeFromExponent(response->GetBlockWiseBlockSize())));
+
+    // Copy options from last response
+    SuccessOrExit(error = iterator.Init(mLastResponse));
+
+    for (const otCoapOption *option = iterator.GetFirstOption(); option != NULL; option = iterator.GetNextOption())
+    {
+        VerifyOrExit((optionBuf = (uint8_t *)calloc(option->mLength, sizeof(uint8_t))) != NULL,
```
jwhui, line 1034: Can we avoid requiring a heap?

src/core/coap/coap.hpp
```diff
@@ -165,6 +165,30 @@ class Resource : public otCoapResource, public LinkedListEntry<Resource>
...
     }
 
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+    /**
+     * This constructor initializes the resource.
+     *
+     * @param[in]  aUriPath  A pointer to a NULL-terminated string for the Uri-Path.
+     * @param[in]  aHandler  A function pointer that is called when receiving a CoAP message for @p aUriPath.
+     * @param[in]  aContext  A pointer to arbitrary context information.
+     * @param[in]  aHook     A function pointer that is called when receiving a CoAP block message for @p aUriPath.
```
jwhui, line 175: nit: doxygen does not match method declaration

src/core/coap/coap.hpp
```diff
@@ -451,6 +512,38 @@ class CoapBase : public InstanceLocator
...
+    otError SendRequestEntityIncomplete(const Message &aRequest, const Ip6::MessageInfo &aMessageInfo)
+    {
+        return SendHeaderResponse(OT_COAP_CODE_REQUEST_INCOMPLETE, aRequest, aMessageInfo);
+    }
+
+    /**
+     * This method sets the maximum block size for CoAP/-s Block-Wise Transfer.
+     *
+     * @param[in]  aBlockSize The maximal allowed block size.
+     */
```
jwhui, line 536: nit:
```suggestion
     *
     */
```

src/core/coap/coap.hpp
```diff
@@ -451,6 +512,38 @@ class CoapBase : public InstanceLocator
...
+    void SetMaxBlockSize(otCoapBlockSzx aBlockSize) { mCurrentMaxBlockSize = aBlockSize; }
+
+    /**
+     * This method returns the current set maximum block size for CoAP/-s Block-Wise Transfer.
+     *
+     * @retval The current set maximum block size.
+     */
```
jwhui, line 543: nit:
```suggestion
     * @returns The current set maximum block size.
     *
     */
```

src/core/coap/coap.hpp
```diff
@@ -451,6 +512,38 @@ class CoapBase : public InstanceLocator
...
+    otCoapBlockSzx GetMaxBlockSize(void) { return mCurrentMaxBlockSize; }
```
jwhui, line 544:
```suggestion
    otCoapBlockSzx GetMaxBlockSize(void) const { return mCurrentMaxBlockSize; }
```

src/core/coap/coap_message.hpp
```diff
@@ -411,6 +413,78 @@ class Message : public ot::Message
...
+     * @param[in]   aBlockType      Block1 or Block2 option value.
+     * @param[out]  aBlockNumber    The relative number of the block within a sequence of blocks.
+     * @param[out]  aMoreBlocks     More blocks are following.
+     * @param[out]  aBlockSize      One of the block lengths listed in otCoapBlockSzx above.
+     *
+     * @retval  OT_ERROR_NONE           The option has been found and is valid.
+     * @retval  OT_ERROR_NOT_FOUND      The option has not been found.
+     * @retval  OT_ERROR_INVALID_ARGS   The option is invalid.
+     */
+    otError ReadBlockOptionValues(otCoapOptionType aBlockType,
```
jwhui, line 428: Is this used somewhere?

src/core/coap/coap_message.cpp
```diff
@@ -225,6 +231,136 @@ otError Message::AppendUriQueryOption(const char *aUriQuery)
...
+        error = OT_ERROR_INVALID_ARGS;
+        break;
+    }
+
+exit:
+    return error;
+}
+#endif // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+
+otError Message::ReadBlockOptionValues(otCoapOptionType aBlockType,
```
jwhui, line 279: Is this used somewhere?

src/core/coap/coap_message.hpp
```diff
@@ -411,6 +413,78 @@ class Message : public ot::Message
...
+     */
+    uint16_t GetHeaderLength(void) const { return GetHelpData().mHeaderLength; }
+
+    /**
+     * This method returns the block number of a CoAP block-wise transfer message.
+     *
+     * @returns The block number.
+     *
+     */
+    uint32_t GetBlockWiseBlockNumber(void) { return GetHelpData().mBlockWiseData.mBlockNumber; }
```
jwhui, line 460:
```suggestion
    uint32_t GetBlockWiseBlockNumber(void) const { return GetHelpData().mBlockWiseData.mBlockNumber; }
```

src/core/coap/coap_message.hpp
```diff
@@ -411,6 +413,78 @@ class Message : public ot::Message
...
+
+    /**
+     * This method checks if the More Blocks flag is set.
+     *
+     * @retval TRUE   More Blocks flag is set.
+     * @retval FALSE  More Blocks flag is not set.
+     *
+     */
+    bool IsMoreBlocksFlagSet(void) { return GetHelpData().mBlockWiseData.mMoreBlocks; }
```
jwhui, line 469:
```suggestion
    bool IsMoreBlocksFlagSet(void) const { return GetHelpData().mBlockWiseData.mMoreBlocks; }
```

src/core/coap/coap_message.hpp
```diff
@@ -411,6 +413,78 @@ class Message : public ot::Message
...
+
+    /**
+     * This method returns the block size of a CoAP block-wise transfer message.
+     *
+     * @returns The block size.
+     *
+     */
+    otCoapBlockSzx GetBlockWiseBlockSize(void) { return GetHelpData().mBlockWiseData.mBlockSize; }
```
jwhui, line 477:
```suggestion
    otCoapBlockSzx GetBlockWiseBlockSize(void) const { return GetHelpData().mBlockWiseData.mBlockSize; }
```

src/core/coap/coap_message.hpp
```diff
@@ -411,6 +413,78 @@ class Message : public ot::Message
...
+#endif
+
+    /**
+     * This function reads and reassembles the URI path string and fills it into @p aUriPath.
+     *
+     * @retval  OT_ERROR_NONE       URI path string has been reassembled.
+     * @retval  OT_ERROR_NO_BUFS    URI path string is too long.
+     */
+    otError GetUriPath(char *aUriPath);
```
jwhui, line 486:
```suggestion
    otError GetUriPath(char *aUriPath) const;
```

src/core/coap/coap_message.hpp
```diff
@@ -411,6 +413,78 @@ class Message : public ot::Message
...
+     */
+    otCoapBlockSzx GetBlockWiseBlockSize(void) { return GetHelpData().mBlockWiseData.mBlockSize; }
+#endif
+
+    /**
+     * This function reads and reassembles the URI path string and fills it into @p aUriPath.
+     *
+     * @retval  OT_ERROR_NONE       URI path string has been reassembled.
+     * @retval  OT_ERROR_NO_BUFS    URI path string is too long.
+     */
```
jwhui, line 485: nit:
```suggestion
     *
     */
```

src/core/coap/coap_message.hpp
```diff
@@ -450,6 +524,33 @@ class Message : public ot::Message
...
+    void SetMoreBlocksFlag(bool aMoreBlocks) { GetHelpData().mBlockWiseData.mMoreBlocks = aMoreBlocks; }
+
+    /**
+     * This method sets the block size value in the message HelpData.
+     *
+     * @param[in]   aBlockSize    Block size value to set.
+     *
+     */
+    void SetBlockWiseBlockSize(otCoapBlockSzx aBlockSize) { GetHelpData().mBlockWiseData.mBlockSize = aBlockSize; }
+#endif
```
jwhui, line 552:
```suggestion
#endif // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```

src/core/coap/coap_message.hpp
```diff
@@ -791,6 +905,13 @@ class OptionIterator : public ::otCoapOptionIterator
      */
     otError GetOptionValue(void *aValue) const;
 
+    /**
+     * This function returns a pointer to the otCoapOption that matches aNumber
+     *
+     * @returns A pointer to the wanted option. If this option is not present NULL pointer is returned.
+     */
+    const otCoapOption *GetOptionByNumber(uint16_t aNumber);
```
jwhui, line 913:
```suggestion
    const otCoapOption *GetOptionByNumber(uint16_t aNumber) const;
```

src/core/coap/coap_secure.cpp
```diff
@@ -152,6 +152,35 @@ void CoapSecure::SetSslAuthMode(bool aVerifyPeerCertificate)
...
+                                ResponseHandler             aHandler,
+                                void *                      aContext,
+                                otCoapBlockwiseTransmitHook aTransmitHook,
+                                otCoapBlockwiseReceiveHook  aReceiveHook)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(IsConnected(), error = OT_ERROR_INVALID_STATE);
+
+    error = CoapBase::SendMessage(aMessage, mDtls.GetPeerAddress(), TxParameters::GetDefault(), aHandler, aContext,
```
jwhui, line 166: Rely on default parameter values to avoid special casing this?

src/core/coap/coap_secure.cpp
```diff
@@ -244,13 +274,10 @@ void CoapSecure::HandleTransmit(void)
...
     {
-        otLogNoteMeshCoP("CoapSecure Transmit: %s", otThreadErrorToString(error));
         message->Free();
     }
-    else
-    {
-        otLogDebgMeshCoP("CoapSecure Transmit: %s", otThreadErrorToString(error));
-    }
+
+    otLogNoteMeshCoP("CoapSecure Transmit: %s", otThreadErrorToString(error));
```
jwhui, line 280: Can we keep the original logging? `NOTE` level is intended for normal but significant events that require special handling (e.g., device getting detached) and are expected to be less frequent than INFO and DEBG level.

jwhui:
@AuSchell, can we also add a functional test for continuous integration testing? See [tests/scripts/thread-cert/test_coap_observe.py](https://github.com/openthread/openthread/blob/master/tests/scripts/thread-cert/test_coap_observe.py) for an example.

include/openthread/coap.h
```diff
@@ -325,6 +312,20 @@ typedef enum otCoapOptionContentFormat
     OT_COAP_OPTION_CONTENT_FORMAT_SENSML_XML = 311
 } otCoapOptionContentFormat;
 
+/**
+ * CoAP Block Size Exponents
+ */
+typedef enum otCoapBlockSzx
```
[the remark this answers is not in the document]
AuSchell: To differentiate between the actual block size in # of bytes and the exponent value which is used for the CoAP option.

src/core/coap/coap_secure.cpp
AuSchell: How would you suggest doing it? I'm a little lost on this one.

include/openthread/coap.h
```diff
@@ -798,6 +851,68 @@ otError otCoapSendRequestWithParameters(otInstance *              aInstance,
...
+ *
+ */
+otError otCoapSendRequestBlockWiseWithParameters(otInstance *                aInstance,
+                                                 otMessage *                 aMessage,
+                                                 const otMessageInfo *       aMessageInfo,
+                                                 otCoapResponseHandler       aHandler,
+                                                 void *                      aContext,
+                                                 const otCoapTxParameters *  aTxParameters,
+                                                 otCoapBlockwiseTransmitHook aTransmitHook,
+                                                 otCoapBlockwiseReceiveHook  aReceiveHook);
```
jwhui, line 882: The number of function parameters is getting quite large. I wonder if it would be useful to combine `aTxParameters`, `aTransmitHook`, and `aReceiveHook` into a single struct that we pass via pointer. That should also simplify extending with additional options in the future as well. Thoughts?

src/core/api/coap_api.cpp
```diff
@@ -218,9 +246,17 @@ otError otCoapSendRequestWithParameters(otInstance *              aInstance,
...
+
+    error = instance.GetApplicationCoap().SendMessage(*static_cast<Coap::Message *>(aMessage),
+                                                      *static_cast<const Ip6::MessageInfo *>(aMessageInfo),
+                                                      txParameters, aHandler, aContext, NULL, NULL);
+#else  // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+
     error = instance.GetApplicationCoap().SendMessage(*static_cast<Coap::Message *>(aMessage),
                                                       *static_cast<const Ip6::MessageInfo *>(aMessageInfo),
                                                       txParameters, aHandler, aContext);
+#endif // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
jwhui, line 259: Default parameters should avoid the need to special-case this.
```suggestion
    error = instance.GetApplicationCoap().SendMessage(*static_cast<Coap::Message *>(aMessage),
                                                      *static_cast<const Ip6::MessageInfo *>(aMessageInfo),
                                                      txParameters, aHandler, aContext);
```

src/core/api/coap_api.cpp
```diff
@@ -261,16 +297,60 @@ void otCoapSetDefaultHandler(otInstance *aInstance, otCoapRequestHandler aHandle
...
+
+otCoapBlockSzx otCoapGetMaxBlockSize(otInstance *aInstance)
+{
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
+    return instance.GetApplicationCoap().GetMaxBlockSize();
+}
+#endif // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
jwhui, line 316: Remove these lines and combine into the same if-else block?

src/core/coap/coap_secure.cpp
jwhui: Apologies, I misplaced this comment. Please ignore.

src/core/coap/coap.cpp
```diff
@@ -239,7 +312,11 @@ otError CoapBase::SendMessage(Message &               aMessage,
                               ResponseHandler         aHandler,
                               void *                  aContext)
 {
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+    return SendMessage(aMessage, aMessageInfo, TxParameters::GetDefault(), aHandler, aContext, NULL, NULL);
+#else  // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
     return SendMessage(aMessage, aMessageInfo, TxParameters::GetDefault(), aHandler, aContext);
+#endif // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
jwhui, line 319: nit: can avoid the comments on small blocks (less than 20 lines)
```suggestion
#endif
```

src/core/coap/coap.cpp
```diff
@@ -239,7 +312,11 @@ otError CoapBase::SendMessage(Message &               aMessage,
                               ResponseHandler         aHandler,
                               void *                  aContext)
 {
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+    return SendMessage(aMessage, aMessageInfo, TxParameters::GetDefault(), aHandler, aContext, NULL, NULL);
+#else  // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
jwhui, line 317: nit: can avoid the comments on small blocks (less than 20 lines)
```suggestion
#else
```

src/core/coap/coap.cpp
```diff
@@ -211,6 +280,10 @@ otError CoapBase::SendMessage(Message &               aMessage,
         metadata.mRetransmissionTimeout    = aTxParameters.CalculateInitialRetransmissionTimeout();
         metadata.mAcknowledged             = false;
         metadata.mConfirmable              = aMessage.IsConfirmable();
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+        metadata.mBlockwiseReceiveHook  = aReceiveHook;
+        metadata.mBlockwiseTransmitHook = aTransmitHook;
+#endif // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
jwhui, line 286: nit: can avoid the comments on small blocks (less than 20 lines)
```suggestion
#endif
```

src/core/coap/coap.hpp
```diff
@@ -166,6 +166,33 @@ class Resource : public otCoapResource, public LinkedListEntry<Resource>
...
+    /**
+     * This constructor initializes the resource.
+     *
+     * @param[in]  aUriPath         A pointer to a NULL-terminated string for the Uri-Path.
+     * @param[in]  aHandler         A function pointer that is called when receiving a CoAP message for @p aUriPath.
+     * @param[in]  aContext         A pointer to arbitrary context information.
+     * @param[in]  aReceiveHook     A function pointer that is called when receiving a CoAP block message for @p
+     * aUriPath.
+     * @param[in]  aTransmitHook    A function pointer that is called when transmitting a CoAP block message from @p
+     * aUriPath.
```
jwhui, line 179: nit: fix alignment here and two lines above

src/core/coap/coap.hpp
```diff
@@ -586,6 +717,12 @@ class CoapBase : public InstanceLocator, private NonCopyable
     void *         mDefaultHandlerContext;
 
     const Sender mSender;
+
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+    otCoapBlockSzx mCurrentMaxBlockSize;
+    Message *      mLastResponse;
+    bool           mHasLastBlockBeenReceived; // Server only
+#endif                                        // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
jwhui, line 725:
```suggestion
#endif
```

src/core/coap/coap_secure.cpp
```diff
@@ -152,6 +152,35 @@ void CoapSecure::SetSslAuthMode(bool aVerifyPeerCertificate)
 
 #endif // OPENTHREAD_CONFIG_COAP_SECURE_API_ENABLE
 
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
jwhui, line 155: Maintaining different methods based on whether block-wise feature is enabled or not ideal. Do you have an idea on the code size impact if we always included the block-wise specific parameters but never used them if the block-wise build feature is not enabled?

AuSchell: With block-wise transfer enabled, the code size of the different binaries increase as follows (all values are decimal and in bytes):
ot-cli-ftd:
text: +2736  bss: +214
ot-cli-mtd:
text: +2720  bss: +96
ot-ncp-ftd:
text: +2656  bss: +208
ot-ncp-mtd:
text: +2640  bss: +80

jwhui: Thanks for the information. What I meant is more of a middle ground, where we update the methods to always include the block-wise transfer parameters so that we do not need to maintain two different versions of them. However, the bulk of the implementation we can still feature flag to minimize code/memory. Thoughts?

AuSchell: I can't really see what feature would be beneficial to be included all the time, except for the method to add the coap-block-wise option to a coap message which already is. Everthing else i see as routines which depend on each other. But if you have suggestions, i will take a look at them gladly.

jwhui:
@AuSchell thanks for the updates. Can you add the necessary hooks so that the test script is executed as part of continuous integration?

src/core/coap/coap.cpp
```diff
@@ -129,25 +132,92 @@ otError CoapBase::Send(ot::Message &aMessage, const Ip6::MessageInfo &aMessageIn
...
 
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+otError CoapBase::SendMessage(Message &                   aMessage,
+                              const Ip6::MessageInfo &    aMessageInfo,
+                              const TxParameters &        aTxParameters,
+                              ResponseHandler             aHandler,
+                              void *                      aContext,
+                              otCoapBlockwiseTransmitHook aTransmitHook,
+                              otCoapBlockwiseReceiveHook  aReceiveHook)
+#else  // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
abtink, line 143:
```suggestion
#else 
```

src/core/coap/coap.cpp
```diff
@@ -129,25 +132,92 @@ otError CoapBase::Send(ot::Message &aMessage, const Ip6::MessageInfo &aMessageIn
...
+#endif // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
 {
     otError  error;
     Message *storedCopy = nullptr;
     uint16_t copyLength = 0;
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+    uint8_t  buf[kMaxBlockLength] = {0};
+    uint16_t bufLen               = kMaxBlockLength;
+    bool     moreBlocks           = false;
+#endif // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
abtink, line 158:
```suggestion
#endif 
```

src/core/coap/coap.cpp
```diff
@@ -129,25 +132,92 @@ otError CoapBase::Send(ot::Message &aMessage, const Ip6::MessageInfo &aMessageIn
...
     case OT_COAP_TYPE_ACKNOWLEDGMENT:
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+        // Check for block-wise transfer
+        if (aTransmitHook != nullptr)
+        {
+            // Check for Block2 option in header
+            if (aMessage.ReadBlockOptionValues(OT_COAP_OPTION_BLOCK2) == OT_ERROR_NONE)
+            {
+                // Check if it is the first block of the transfer
+                if (aMessage.GetBlockWiseBlockNumber() == 0)
```
abtink, line 171: How about combining the nested three `if`s into one with conditions added using `&&`.
```suggestion
        if ((aTransmitHook != nullptr) && (aMessage.ReadBlockOptionValues(OT_COAP_OPTION_BLOCK2) == OT_ERROR_NONE) && (aMessage.GetBlockWiseBlockNumber() == 0))
        {

```

src/core/coap/coap.cpp
```diff
@@ -129,25 +132,92 @@ otError CoapBase::Send(ot::Message &aMessage, const Ip6::MessageInfo &aMessageIn
...
     default:
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+        // Check for block-wise transfer
+        if (aTransmitHook != nullptr)
+        {
+            // Check for Block1 option in header
+            if (aMessage.ReadBlockOptionValues(OT_COAP_OPTION_BLOCK1) == OT_ERROR_NONE)
+            {
+                // Check if it is the first block of the transfer
+                if (aMessage.GetBlockWiseBlockNumber() == 0)
```
abtink, line 201: Same suggestion here. Combining the `if`s?

src/core/coap/coap.cpp
```diff
@@ -561,6 +1037,10 @@ void CoapBase::ProcessReceivedResponse(Message &aMessage, const Ip6::MessageInfo
 #if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
     bool responseObserve = false;
 #endif
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+    uint8_t  blockOptionType    = 0;
+    uint32_t totalTransfereSize = 0;
+#endif // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
abtink, line 1043:
```suggestion
#endif
```

src/core/coap/coap.cpp
```diff
@@ -691,6 +1250,10 @@ void CoapBase::ProcessReceivedRequest(Message &aMessage, const Ip6::MessageInfo
     Message *      cachedResponse = nullptr;
     otError        error          = OT_ERROR_NOT_FOUND;
     OptionIterator iterator;
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+    uint8_t  blockOptionType    = 0;
+    uint32_t totalTransfereSize = 0;
+#endif // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
abtink, line 1256:
```suggestion
#endif 
```

src/core/coap/coap.cpp
```diff
@@ -730,13 +1293,87 @@ void CoapBase::ProcessReceivedRequest(Message &aMessage, const Ip6::MessageInfo
...
+
+        case OT_COAP_OPTION_BLOCK2:
+            blockOptionType += 2;
+            break;
+
+        case OT_COAP_OPTION_SIZE1:
+            // ToDo: wait for method to read uint option values
+            totalTransfereSize = 0;
+            break;
+#endif // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
abtink, line 1309:
```suggestion
#endif 
```

src/core/coap/coap.cpp
```diff
@@ -730,13 +1293,87 @@ void CoapBase::ProcessReceivedRequest(Message &aMessage, const Ip6::MessageInfo
...
         }
     }
 
     curUriPath[0] = '\0';
 
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+    for (const Resource *resource = mResources.GetHead(); resource; resource = resource->GetNext())
+    {
+        if (strcmp(resource->mUriPath, uriPath) == 0)
+        {
```
abtink, line 1322: Suggest using `continue`  model to reduce multiple `if` indentations.
```
if (strcmp(resource->mUriPath, uriPath) != 0)
{
     continue;
}
```

src/core/coap/coap.hpp
```diff
@@ -457,6 +520,24 @@ class CoapBase : public InstanceLocator, private NonCopyable
...
+     *
+     * @retval OT_ERROR_NONE         Successfully enqueued the CoAP response message.
+     * @retval OT_ERROR_NO_BUFS      Insufficient buffers available to send the CoAP response.
+     *
+     */
+    otError SendRequestEntityIncomplete(const Message &aRequest, const Ip6::MessageInfo &aMessageInfo)
+    {
+        return SendHeaderResponse(OT_COAP_CODE_REQUEST_INCOMPLETE, aRequest, aMessageInfo);
+    }
+#endif // OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
```
abtink, line 539:
```suggestion
#endif 
```

AuSchell:
> @AuSchell thanks for the updates. Can you add the necessary hooks so that the test script is executed as part of continuous integration?

@jwhui what kind of hooks do you mean? I'm not familiar with your test enviroment, i've more or less copied what other test are doing.

AuSchell:
@abtink thank you for your review, i've applied the changes you suggested. For https://github.com/openthread/openthread/pull/4247#discussion_r468908248 i just added a comment that these members are only available with coap block-wise enabled, is that a vaild option?

jwhui:
> @jwhui what kind of hooks do you mean? I'm not familiar with your test enviroment, i've more or less copied what other test are doing.

Add the script to `EXTRA_DIST` and `check_SCRIPTS` in [tests/scripts/thread-cert/Makefile.am](https://github.com/openthread/openthread/blob/master/tests/scripts/thread-cert/Makefile.am).

AuSchell:
> > @jwhui what kind of hooks do you mean? I'm not familiar with your test enviroment, i've more or less copied what other test are doing.
> 
> Add the script to `EXTRA_DIST` and `check_SCRIPTS` in [tests/scripts/thread-cert/Makefile.am](https://github.com/openthread/openthread/blob/master/tests/scripts/thread-cert/Makefile.am).

The script has been added to that file.

src/core/coap/coap.cpp
abtink: Per the style guide, we don't add comments in cases where there is less than 20 lines between the `#if` and its related `#endif` to make it cleaner/easier to read (avoid unnecessary comments). We add it when there are   more than 20 lines in between.

src/core/coap/coap.cpp
```diff
@@ -129,25 +146,76 @@ otError CoapBase::Send(ot::Message &aMessage, const Ip6::MessageInfo &aMessageIn
...
                               const TxParameters &    aTxParameters,
                               ResponseHandler         aHandler,
                               void *                  aContext)
+#endif
 {
     otError  error;
     Message *storedCopy = nullptr;
     uint16_t copyLength = 0;
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+    uint8_t  buf[kMaxBlockLength] = {0};
```
abtink, line 169: Do we need the buffer to initialized to zero? Seems like we would be reading it from message below.
If not I would suggest removing `{0}`?

src/core/coap/coap.cpp
```diff
@@ -465,6 +541,410 @@ void CoapBase::DequeueMessage(Message &aMessage)
...
+
+    SuccessOrExit(error = SendMessage(*request, aMessageInfo, TxParameters::GetDefault(),
+                                      aCoapMetadata.mResponseHandler, aCoapMetadata.mResponseContext,
+                                      aCoapMetadata.mBlockwiseTransmitHook, aCoapMetadata.mBlockwiseReceiveHook));
+
+exit:
+    if (error != OT_ERROR_NONE && request != nullptr)
+    {
+        request->Free();
+    }
```
abtink, line 688: Can use helper function/macro here:
```suggestion
    FreeMessageOnError(request, error);
```

src/core/coap/coap.cpp
```diff
@@ -465,6 +541,410 @@ void CoapBase::DequeueMessage(Message &aMessage)
...
+
+    SuccessOrExit(error =
+                      SendMessage(*request, aMessageInfo, TxParameters::GetDefault(), aCoapMetadata.mResponseHandler,
+                                  aCoapMetadata.mResponseContext, nullptr, aCoapMetadata.mBlockwiseReceiveHook));
+
+exit:
+    if (error != OT_ERROR_NONE && request != nullptr)
+    {
+        request->Free();
+    }
```
abtink, line 753:
```suggestion
    FreeMessageOnError(request, error);
```

src/core/coap/coap.cpp
```diff
@@ -465,6 +541,410 @@ void CoapBase::DequeueMessage(Message &aMessage)
...
+    otLogInfoCoap("Send Block2 Nr. %d, Size: %d bytes, More Blocks Flag %d", response->GetBlockWiseBlockNumber(),
+                  otCoapBlockSizeFromExponent(response->GetBlockWiseBlockSize()), response->IsMoreBlocksFlagSet());
+
+    SuccessOrExit(error = SendMessage(*response, aMessageInfo));
+
+exit:
+    if (error != OT_ERROR_NONE && response != nullptr)
+    {
+        response->Free();
+    }
```
abtink, line 942:
```suggestion
    FreeMessageOnError(response, error);
```

src/core/coap/coap_message.hpp
```diff
@@ -820,6 +904,16 @@ class Message : public ot::Message
...
     };
 
+#if OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+    OT_TOOL_PACKED_BEGIN
+    struct BlockWiseData
+    {
+        uint32_t       mBlockNumber;
+        bool           mMoreBlocks;
+        otCoapBlockSzx mBlockSize;
+    } OT_TOOL_PACKED_END;
```
abtink, line 914: Do we need this to be packed? It seems to be used in `HelpData` only and we don't care about `HelpData` not being packed (not using packed when possible would help reduce code-size).

abtink (commented):
Looks good overall. Thanks.
Few smaller changes/suggestions below.

include/openthread/coap.h
```diff
@@ -367,6 +420,26 @@ typedef struct otCoapResource
...
+    otCoapBlockwiseReceiveHook mReceiveHook;   /* The callback for handling incoming blockwise transfer.
+                                                  This callback is available when
+                                                  OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+                                                  configuration is enabled. */
+    otCoapBlockwiseTransmitHook mTransmitHook; /* The callback for handling outgoing blockwise transfer.
+                                                  This callback is available when
+                                                  OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
+                                                  configuration is enabled. */
+    void *                          mContext;  ///< Application-specific context
+    struct otCoapBlockwiseResource *mNext;     ///< The next CoAP resource in the list
```
abtink, line 440: Suggest re-arranging the longer comment before the definition and start with `/** ` for doxygen to associate it correctly with the member variable.

```suggestion
    otCoapRequestHandler       mHandler;       ///< The callback for handling a received request

    /** The callback for handling incoming blockwise transfer.
     *  This callback is available when
     *  OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
     *  configuration is enabled.
     */
    otCoapBlockwiseReceiveHook mReceiveHook;

    /** The callback for handling outgoing blockwise transfer.
     *  This callback is available when
     *  OPENTHREAD_CONFIG_COAP_BLOCKWISE_TRANSFER_ENABLE
     *  configuration is enabled. 
     */
    otCoapBlockwiseTransmitHook mTransmitHook; 

    void *                          mContext;  ///< Application-specific context
    struct otCoapBlockwiseResource *mNext;     ///< The next CoAP resource in the list
```

abtink (approved):
Thanks @AuSchell. Looks great. 👍

jwhui:
@sjlongland , would you be willing to take another pass on this PR? It is getting close to merging.

sjlongland:
@jwhui I'll try to make a note to have a look soon.  Been flat out lately as we enter the end-of-year "silly season".

sjlongland (changes requested):
Aside from one bit of documentation that could use some clarification, this is looking really good.

sjlongland (approved):

abtink (approved):
LGTM. Thanks. 👍

jwhui (approved):
Thanks for all your work on this, @AuSchell !

PR #4250, opened 2019-10-16
[third-party] change end of line character from CRLF do LF

ktaborowski:
Changed end of line character form CRLF do LF in some third_party files


jwhui:
@ktaborowski, thanks for cleaning this up. I see a few modified files in `third_party/mbedtls/repo`. In general, those files are pristine from https://github.com/ARMmbed/mbedtls. Are you OK if we leave those unmodified?

ktaborowski:
> @ktaborowski, thanks for cleaning this up. I see a few modified files in `third_party/mbedtls/repo`. In general, those files are pristine from https://github.com/ARMmbed/mbedtls. Are you OK if we leave those unmodified?

Yes, I'm ok, because as far as I know there are Microsoft Visual Studio's files. I can't test if change to unix end of line characters is safe for it.

KaSroka:
@jwhui I think that line endings should be handled by git and as we have `.gitattributes` with:
```
# Auto normalize all files which appear to be text files.
*        text=auto
...
```
This setting keeps LF on the GitHub side - but it's not applied to third_party due to:
https://github.com/openthread/openthread/blob/master/third_party/.gitattributes
I don't know why we have this here. Also, MbedTLS doesn't have `.gitattributes` file so they have CRLF in their repos. I think that there is no harm in having Windows project files checked out with LF on Linux - they would convert to CRLF on Windows anyway.

jwhui:
@KaSroka, OK. The `third_party/.gitattributes` file was introduced to deal with issues when collaborating Windows and Linux developers. Given that we dropped Windows support, I am fine to remove `third_party/.gitattributes` as well. Can we do that in this PR?

ktaborowski:
@jwhui I removed .gitattributes from third_party

PR #4252, opened 2019-10-17
[spi-hdlc-adapter] remove unused variables

zhanglongxia:


tools/spi-hdlc-adapter/spi-hdlc-adapter.c
```diff
@@ -1722,7 +1713,12 @@ int main(int argc, char *argv[])
     argc -= optind;
     argv += optind;
 
-    if (argc >= 1)
+    if (argc == 0)
```
bukepo: prefer not change `argc` and `argv`, just compare `argc` and `optind` is good.

zhanglongxia: Updated.

tools/spi-hdlc-adapter/spi-hdlc-adapter.c
```diff
@@ -141,10 +141,6 @@ static int sMode = MODE_PTY;
 static int sMode = MODE_STDIO;
 #endif
 
-static const char *sSpiDevPath     = NULL;
-static const char *sIntGpioDevPath = NULL;
-static const char *sResGpioDevPath = NULL;
```
abtink: Even though these variable are not used in the code, I suggest we keep them as they are (they would be useful).
For example, in local tests/builds, we used to add these in log messages which was very helpful for debugging, or use gdb to check the value.

Actually I think it  may be helpful to log these as part of the code where the stat/info is being logged. 

Thoughts?

zhanglongxia: I just felt strange when I looked at the these variables. Because no one is using these variables.  For debugging, I think it is better to log these variables to syslog when the code is processing these variables.  Thoughts?

abtink: I am fine with that.

PR #4253, opened 2019-10-17
[cli] get preferred and supported channel mask

bukepo:


src/cli/cli.cpp
```diff
@@ -450,6 +450,14 @@ void Interpreter::ProcessChannel(int argc, char *argv[])
     {
         mServer->OutputFormat("%d\r\n", otLinkGetChannel(mInstance));
     }
+    else if (strcmp(argv[0], "supported") == 0)
+    {
+        mServer->OutputFormat("%x\r\n", otPlatRadioGetSupportedChannelMask(mInstance));
```
abtink: LGTM. I have one suggestion.
How about also converting to `Mac::ChannelMask` and then using `ChannelMask::ToString()` method here to log the mask (looks like we already include/use `Mac::ChannelMask` in this file for channel manager). 

This way we get the mask in more human-readable style: like `{ 11-26 }`, or `{ 11, 14-17, 20-22, 24, 25 }`.

Thoughts?

bukepo: Thank you! I think it's good idea to output human readable style in loggings. However I add this to support testing this feature in some internal testbeds. The hex number seems to be more friendly for writing test scripts, and outputting the number is consistent with wpanctl's behavior. I prefer just output program-readable style, what do you think?

bukepo: Just found wpanctl output decimals. Updated the output format.

jwhui: I find hex easier to understand in the case of mask. What about changing the channel mask output to hex and prepend with `0x`?

PR #4255, opened 2019-10-18
[coap] force set message id for CON and NON types

bukepo:
This commit force settings message id for CON and NON type CoAP messages.

### DIFF
--- a/src/core/coap/coap.cpp
+++ b/src/core/coap/coap.cpp
@@ -234,7 +234,6 @@ otError CoapBase::SendHeaderResponse(Message::Code aCode, const Message &aReques
 
     case OT_COAP_TYPE_NON_CONFIRMABLE:
         message->Init(OT_COAP_TYPE_NON_CONFIRMABLE, aCode);
-        message->SetMessageId(mMessageId++);
         break;
 
     default:
--- a/src/core/coap/coap.cpp
+++ b/src/core/coap/coap.cpp
@@ -147,15 +147,12 @@ otError CoapBase::SendMessage(Message &               aMessage,
     switch (aMessage.GetType())
     {
     case OT_COAP_TYPE_ACKNOWLEDGMENT:
-        assert(aMessage.IsMessageIdValid());
         mResponsesQueue.EnqueueResponse(aMessage, aMessageInfo);
         break;
     case OT_COAP_TYPE_RESET:
-        assert(aMessage.IsMessageIdValid());
         assert(aMessage.GetCode() == OT_COAP_CODE_EMPTY);
         break;
     default:
-        assert(!aMessage.IsMessageIdValid());
         aMessage.SetMessageId(mMessageId++);
         break;
     }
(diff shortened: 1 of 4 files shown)


src/core/coap/coap.cpp
```diff
@@ -151,8 +151,7 @@ otError CoapBase::SendMessage(Message &               aMessage,
     }
 
     // Set Message Id if it was not already set.
-    if (aMessage.GetMessageId() == 0 &&
-        (aMessage.GetType() == OT_COAP_TYPE_CONFIRMABLE || aMessage.GetType() == OT_COAP_TYPE_NON_CONFIRMABLE))
+    if (!aMessage.IsMessageIdValid())
```
simonlingoogle: Here, Message ID should always be valid for ACK messages, add a check for that?

src/core/coap/coap_message.cpp
```diff
@@ -358,11 +358,30 @@ otError Message::ParseHeader(void)
...
+{
+    const HelpData &helpData = GetHelpData();
+
+    // Invalid usage to get before set
+    assert(helpData.mIsMessageIdValid);
+
+    return HostSwap16(helpData.mHeader.mMessageId);
+}
+
+void Message::SetMessageId(uint16_t aMessageId)
```
simonlingoogle: In the file `border_agent.cpp`, there are two calls to `SetMessageId` I think might have issues.

 - `message->SetMessageId(aSeparate ? 0 : aRequest.GetMessageId());` in `SendErrorMessage`
 - `aMessage.SetMessageId(mSeparate ? 0 : mMessageId);` in `ToHeader`

These two `SetMessageId` might set the message ID to 0, which used to means `using the next message ID`. 
Please check if these two calls is still valid after the semantic change of message ID being zero.

bukepo: They are correct. The message id is set `0` when this is a separate response. For piggy-backed response, the message id should be identical to that of the original request.

bukepo: I got your idea.

src/core/coap/coap.cpp
```diff
@@ -144,17 +144,20 @@ otError CoapBase::SendMessage(Message &               aMessage,
...
-    if (aMessage.GetMessageId() == 0 &&
-        (aMessage.GetType() == OT_COAP_TYPE_CONFIRMABLE || aMessage.GetType() == OT_COAP_TYPE_NON_CONFIRMABLE))
-    {
+        break;
+    case OT_COAP_TYPE_RESET:
+        assert(aMessage.IsMessageIdValid());
+        assert(aMessage.GetCode() == OT_COAP_CODE_EMPTY);
+        break;
+    default:
+        assert(!aMessage.IsMessageIdValid());
```
simonlingoogle: I think this assert might fail because CON/NON COAP messages  can also have message ID already set. 
For example, `case OT_COAP_TYPE_NON_CONFIRMABLE` in `CoapBase::SendHeaderResponse`

simonlingoogle: I found that in the end, seems we don't actual need to check `IsMessageIdValid` at all...

PR #4256, opened 2019-10-18
[mac] add option to change direct/indirect TX retry number

e-rk:
This PR adds an option to set the number of direct/indirect number of TX retries on the MAC layer.

The PR covers:
* Public API
* NCP properties
* CLI commands


src/core/config/openthread-core-config-check.h
```diff
@@ -441,4 +441,9 @@
     "OPENTHREAD_CONFIG_PLATFORM_RADIO_COEX_METRICS_ENABLE was replaced by OPENTHREAD_CONFIG_PLATFORM_RADIO_COEX_ENABLE."
 #endif
 
+#ifdef OPENTHREAD_CONFIG_MAC_MAX_FRAME_RETRIES_DIRECT
```
LuDuda: Can we do the same for `OPENTHREAD_CONFIG_MAC_MAX_FRAME_RETRIES_INDIRECT` define?

src/core/mac/mac.hpp
```diff
@@ -425,6 +426,43 @@ class Mac : public InstanceLocator
...
+     *
+     * @returns The maximum number of retries during indirect transmission.
+     *
+     */
+    uint8_t GetMaxFrameRetriesIndirect(void) const { return mMaxFrameRetriesIndirect; }
+
+    /**
+     * This method sets the maximum number of frame retries during indirect transmission.
+     *
+     * @param[in]  aMaxFrameRetriesIndirect  The maximum number of retries induring indirect transmission.
```
LuDuda: nit: induring -> during

src/core/config/openthread-core-config-check.h
e-rk: I forgot about that one. Thanks.

LuDuda:
Any chance to get some initial feedback about this PR?

src/ncp/spinel.h
```diff
@@ -1927,6 +1927,22 @@ typedef enum
      */
     SPINEL_PROP_MAC_CCA_FAILURE_RATE = SPINEL_PROP_MAC_EXT__BEGIN + 9,
 
+    /// MAC Max direct retry number
+    /** Format: `C`
+     *
+     * The maximum (user-specified) number of direct frame transmission retries.
+     *
+     */
+    SPINEL_PROP_MAC_MAX_RETRY_NUMBER_DIRECT = SPINEL_PROP_MAC_EXT__BEGIN + 10,
```
abtink: nit: can you please add support for new porp definition in spinel.c `spinel_prop_key_to_cstr()` function? Thanks

src/ncp/spinel.h
```diff
@@ -1927,6 +1927,22 @@ typedef enum
...
+
+    /// MAC Max indirect retry number
+    /** Format: `C`
+     *
```
abtink: nit: how about adding  `Required capability: CAP_CONFIG_FTD` to document it can be used on FTD only?

e-rk:
Thanks @abtink.
In the meantime we have considered changing the `macretries` name into `mac retries` instead. That way, the `mac` related commands would be under one umbrella prefix. Would that be okay with you?

jwhui:
@e-rk, having a `mac` prefix works for me. We could also move `macfilter` to `mac filter` in a later PR.

PR #4258, opened 2019-10-21
[dataset] store only when changed

bukepo:


bukepo:
@abtink, good idea! How about implementing this check in the general `Settings::Save()` method?

abtink:
@bukepo sure. even better. :) 
I can help add it if you want.

jwhui:
@abtink, @bukepo - I actually think that avoiding duplicate writes is a platform driver optimization. How the optimization is implemented really depends on how the platform driver is implemented. For example, maybe the platform implements its own integrity checks and uses those integrity checks to determine whether any data has changed. Reading out into a buffer and doing a memcmp works but may not be the most efficient. Thoughts?

abtink:
I agree it would be technically better optimization in platform code itself, but considering it may not be present in the many platform implementation and it's sort of a  common pattern  we can add it in OT core to be safe.

Reading a previous saved value should be relatively cheap operation-wise (compared to re-saving the same value).

We can consider adding such a check in OT core as an optional feature (controlled with a CONFIG option - probably enabled by default) and then leave it to platform code to decide if they want to disable it and use their own implementation/optimization or keep it enabled in core.

Thoughts?

bukepo:
I'm good with optimizing on platform layer, just like ensuring operation atomic. OpenThread itself should more focus on the stack itself.

jwhui:
Resolved by #4304.

PR #4261, opened 2019-10-22
[harness-automation] save case failure times

FiveDimensions:
Save failure times for each case for judging if case passes randomly.

### DIFF
--- a/tools/harness-automation/autothreadharness/runner.py
+++ b/tools/harness-automation/autothreadharness/runner.py
@@ -106,10 +106,7 @@ class SimpleTestResult(unittest.TestResult):
             test (TestCase): The test just run
             passed (bool): Whether the case is passed
         """
-        if str(test.__class__.__name__) in self.result and 'fails' in self.result[str(test.__class__.__name__)]:
-            fails = int(self.result[str(test.__class__.__name__)]['fails'])
-        else:
-            fails = 0
+        fails = self.result.get(test.__class__.__name__, {}).get('fails', 0)
         if passed is False:
             fails += 1
         self.result[str(test.__class__.__name__)] = {
--- a/tools/harness-automation/autothreadharness/runner.py
+++ b/tools/harness-automation/autothreadharness/runner.py
@@ -106,10 +106,17 @@ class SimpleTestResult(unittest.TestResult):
             test (TestCase): The test just run
             passed (bool): Whether the case is passed
         """
+        if str(test.__class__.__name__) in self.result and 'fails' in self.result[str(test.__class__.__name__)]:
+            fails = int(self.result[str(test.__class__.__name__)]['fails'])
+        else:
+            fails = 0
+        if passed is False:
+            fails += 1
         self.result[str(test.__class__.__name__)] = {
             'started': self.started,
             'stopped': time.strftime('%Y-%m-%dT%H:%M:%S'),
             'passed': passed,
+            'fails': fails,
             'error': error,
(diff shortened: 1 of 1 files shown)


tools/harness-automation/autothreadharness/runner.py
```diff
@@ -106,10 +106,17 @@ def add_result(self, test, passed, error=None):
             test (TestCase): The test just run
             passed (bool): Whether the case is passed
         """
+        if str(test.__class__.__name__) in self.result and 'fails' in self.result[str(test.__class__.__name__)]:
```
simonlingoogle: is `str(test.__class__.__name__)` necessary? I believe `test.__class__.__name__` is assured to be `str`

tools/harness-automation/autothreadharness/runner.py
```diff
@@ -106,10 +106,17 @@ def add_result(self, test, passed, error=None):
...
+            fails = int(self.result[str(test.__class__.__name__)]['fails'])
+        else:
+            fails = 0
+        if passed is False:
```
simonlingoogle: nit: it's better just use `if not passed:` according to [PEP8](https://www.python.org/dev/peps/pep-0008/)
```
Don't compare boolean values to True or False using ==.

Yes:   if greeting:
No:    if greeting == True:
Worse: if greeting is True:
```

tools/harness-automation/autothreadharness/runner.py
```diff
@@ -106,10 +106,17 @@ def add_result(self, test, passed, error=None):
             test (TestCase): The test just run
             passed (bool): Whether the case is passed
         """
+        if str(test.__class__.__name__) in self.result and 'fails' in self.result[str(test.__class__.__name__)]:
+            fails = int(self.result[str(test.__class__.__name__)]['fails'])
```
simonlingoogle: what about `fails = self.result.get(test.__class__.__name__, {}).get('fails', 0)`?

tools/harness-automation/autothreadharness/runner.py
```diff
@@ -106,10 +106,17 @@ def add_result(self, test, passed, error=None):
...
+        else:
+            fails = 0
+        if passed is False:
+            fails += 1
         self.result[str(test.__class__.__name__)] = {
             'started': self.started,
             'stopped': time.strftime('%Y-%m-%dT%H:%M:%S'),
             'passed': passed,
+            'fails': fails,
```
simonlingoogle: Are the tests supposed to stop once we get an `Pass` ?
If we run 4 tests: 2 fail then 2 pass, we are going to get the result:
```
{
    ...
    'fails': 2, 
    'passed': True,
    ...
}
```

So we don't know how many pass we got. 
Why not also save the number of `Pass` we got?

tools/harness-automation/autothreadharness/runner.py
FiveDimensions: The reason I used "is False" is that "addError" will pass "None" to "passed" .  "passed" has 3 kinds of value : True, False and None. Only if passed is False, fails ++

tools/harness-automation/autothreadharness/runner.py
FiveDimensions: For current test strategy, once passed, automation tools will move to test next case.

tools/harness-automation/autothreadharness/runner.py
FiveDimensions: cool

PR #3785, opened 2019-04-25
[mesh-forwarder] verify data request is ack with fp=1

bukepo:
This PR adds the ability to verify if a data request is acknowledged with frame pending set. New radio drivers should set this flag when processing MAC data request.

Here's more background for this:

> SED may send MAC data-requests just after it sends a data message and expects a response. This is OK in single-chip mode. In dual-chip solution, since ACK is handled in the RCP, these immediate data poll could be processed by RCP before it processing the adding src match entry request, then the host chip processes the Data Request and start transmitting data frames before an ACK with frame pending set is actually sent to the SED. Data frames are dropped because SED is still expecting a correct ACK. As a result, several retransmissions are spotted. For example:
> 1. SED sends Child Update Request, and then starts sending Data Request fastly.
> 2. Router responds to Child Update Request and first Data Request with ACK and Frame Pending = 0.
> 3. Router processes Child Update Request, inserts Child Update Response into indirect transmission queue
> 4. Router tells radio driver to add source address match entry.
> 5. Before step 4 is completed, radio driver responds another Data Request with Frame Pending = 0,  and the Data Request is delivered to MAC layer.
> 6. Router starts transmitting the Child Update Response, which is dropped or not even received 
because SED didn't received an ACK with Frame Pending = 1 yet. Multiple retransmissions happens because radio driver cannot reply new ACK before this transmission finally fails.

PR #4263, opened 2019-10-22
[cc2538] set frame pending correctly

simonlingoogle:
The commits set `frame pending` correctly for cc2538 platform. 

When source match is enabled, `readFrame` reads the `SRC_MATCH_FOUND` flag from the register `RFCORE_SFR_RFIRQF0`. 

The `SRC_MATCH_FOUND` flag will be set whenever there is a message received and ACKed with FP=1 before `readFrame` is called. It is therefore read and cleared in `readFrame`. 

This is one of TODOs of PR #3785 .
**This RP conflicts with PR #4291 on `examples/platforms/cc2538/radio.c`**

### DIFF
--- a/examples/platforms/cc2538/radio.c
+++ b/examples/platforms/cc2538/radio.c
@@ -262,6 +262,21 @@ void setTxPower(int8_t aTxPower)
     }
 }
 
+static bool cc2538SrcMatchEnabled(void)
+{
+    return (HWREG(RFCORE_XREG_FRMCTRL1) & RFCORE_XREG_FRMCTRL1_PENDING_OR) == 0;
+}
+
+static bool cc2538GetSrcMatchFoundIntFlag(void)
+{
+    bool flag = (HWREG(RFCORE_SFR_RFIRQF0) & RFCORE_SFR_RFIRQF0_SRC_MATCH_FOUND) != 0;
+    if (flag)
+    {
+        HWREG(RFCORE_SFR_RFIRQF0) &= ~RFCORE_SFR_RFIRQF0_SRC_MATCH_FOUND;
+    }
+    return flag;
+}
+
 void otPlatRadioGetIeeeEui64(otInstance *aInstance, uint8_t *aIeeeEui64)
 {
     OT_UNUSED_VARIABLE(aInstance);
@@ -704,6 +719,14 @@ static void readFrame(void)
     {
         sReceiveFrame.mLength            = length;
         sReceiveFrame.mInfo.mRxInfo.mLqi = crcCorr & CC2538_LQI_BIT_MASK;
+
+        if (length > IEEE802154_ACK_LENGTH)
+        {
+            // Set ACK FP flag for the received frame according to whether SRC_MATCH_FOUND was triggered just before
+            // if SRC MATCH is not enabled, SRC_MATCH_FOUND is not triggered and all ACK FP is always set
+            sReceiveFrame.mInfo.mRxInfo.mAckedWithFramePending =
+                cc2538SrcMatchEnabled() ? cc2538GetSrcMatchFoundIntFlag() : true;
+        }
     }
     else
     {
@@ -753,9 +776,6 @@ void cc2538RadioProcess(otInstance *aInstance)
     if ((sState == OT_RADIO_STATE_RECEIVE && sReceiveFrame.mLength > 0) ||
         (sState == OT_RADIO_STATE_TRANSMIT && sReceiveFrame.mLength > IEEE802154_ACK_LENGTH))
     {
-        // TODO Set this flag only when the packet is really acknowledged with frame pending set.
-        // See https://github.com/openthread/openthread/pull/3785
-        sReceiveFrame.mInfo.mRxInfo.mAckedWithFramePending = true;
 #if OPENTHREAD_CONFIG_DIAG_ENABLE
 
         if (otPlatDiagModeGet())


examples/platforms/cc2538/radio.c
```diff
@@ -813,13 +826,19 @@ void cc2538RadioProcess(otInstance *aInstance)
...
 #endif
 }
 
 void RFCoreRxTxIntHandler(void)
 {
 #if OPENTHREAD_CONFIG_CC2538_USE_RADIO_RX_INTERRUPT
+
+    if (HWREG(RFCORE_SFR_RFIRQF0) & RFCORE_SFR_RFIRQF0_SRC_MATCH_FOUND)
+    {
+        sSrcMatchFound = true; // the next message received will have frame pending set
```
bukepo: Is it guaranteed the frame will be received and trigger the rx done interrupt? Will Rx done be triggered when FCS is incorrect? Maybe we should set clear the flag in rx_frame_accepted context.

examples/platforms/cc2538/radio.c
```diff
@@ -704,6 +716,8 @@ static void readFrame(void)
     {
         sReceiveFrame.mLength            = length;
         sReceiveFrame.mInfo.mRxInfo.mLqi = crcCorr & CC2538_LQI_BIT_MASK;
+        // sSrcMatchFound should have been set if this frame is acked with FP
+        sReceiveFrame.mInfo.mRxInfo.mAckedWithFramePending = cc2538SrcMatchEnabled() ? sSrcMatchFound : true;
```
bukepo: Is it possible to directly set `mAckedWithFramePending` during the src_matched interrupt?

simonlingoogle: No, the `SRC_MATCH_FOUND` interrupt comes before `RX` interrupt, so at that time, even `readFrame` has not executed, and there is no data in `sReceiveFrame`

examples/platforms/cc2538/radio.c
simonlingoogle: `sSrcMatchFound` is set to false even when crc check failed

examples/platforms/cc2538/radio.c
simonlingoogle: Maybe yes, need more thinking

examples/platforms/cc2538/radio.c
simonlingoogle: `RXPKTDONE A complete frame has been received.`
Currently, `readFrame` is triggered in the so called `rx_frame_accepted` context

examples/platforms/cc2538/radio.c
simonlingoogle: Setting `sReceiveFrame.mInfo.mRxInfo.mAckedWithFramePending` or `sSrcMatchFound` has no real difference because they are both static variables. Would be more understandable using a separate variable?

examples/platforms/cc2538/radio.c
simonlingoogle: According to cc2538 user guide, the order or interrupts is:
 - When source matched: `RX_FRM_ACCEPTED -> SRC_MATCH_FOUND -> RXPKTDONE`
 - When source not matched: `RX_FRM_ACCEPTED -> RXPKTDONE`

So, it is guaranteed that there will always be a `RXPKTDONE ` after `SRC_MATCH_FOUND`. 
For messages filtered by frame filtering, none of `RX_FRM_ACCEPTED -> SRC_MATCH_FOUND -> RXPKTDONE` will be triggered.

jwhui:
@sjlongland, can you help review and provide your thoughts?

librasungirl:
@srickardti would you please help to take a look?

sjlongland:
I note only the `OPENTHREAD_CONFIG_CC2538_USE_RADIO_RX_INTERRUPT=1` case is handled, I wonder if we ensured that the NVIC channel for the radio was left disabled (whilst enabling the interrupt flags on the radio module itself), would that enable the same logic to work for `OPENTHREAD_CONFIG_CC2538_USE_RADIO_RX_INTERRUPT=0`?

Is there a down-side to not implementing it for the pure polling case?

simonlingoogle:
> Is there a down-side to not implementing it for the pure polling case?

Yes and thanks for the idea :+1: 
I have included in the latest commit the implementation for the pure polling case. Please take a look. 
@sjlongland

jwhui:
@sjlongland, can you help review this again? Thanks!

examples/platforms/cc2538/radio.c
```diff
@@ -704,6 +729,14 @@ static void readFrame(void)
     {
         sReceiveFrame.mLength            = length;
         sReceiveFrame.mInfo.mRxInfo.mLqi = crcCorr & CC2538_LQI_BIT_MASK;
+
+        if (length > IEEE802154_ACK_LENGTH)
+        {
+            // Set ACK FP flag for the received frame according to whether SRC_MATCH_FOUND was triggered just before
+            // if SRC MATCH is not enabled, SRC_MATCH_FOUND is not triggered and all ACK FP is always set
+            sReceiveFrame.mInfo.mRxInfo.mAckedWithFramePending =
+                cc2538SrcMatchEnabled() ? cc2538GetSrcMatchFoundIntFlag() : true;
```
librasungirl: should it be set to false if `cc2538SrcMatchEnabled()` is not enabled?

librasungirl: recall previous comment as it is PENDING_OR when disabled

simonlingoogle: When source match table is not enabled, ack FP is always set according to datasheet

examples/platforms/cc2538/radio.c
```diff
@@ -262,6 +262,21 @@ void setTxPower(int8_t aTxPower)
...
+    return (HWREG(RFCORE_XREG_FRMCTRL1) & RFCORE_XREG_FRMCTRL1_PENDING_OR) == 0;
+}
+
+static bool cc2538GetSrcMatchFoundIntFlag(void)
+{
+    bool flag = (HWREG(RFCORE_SFR_RFIRQF0) & RFCORE_SFR_RFIRQF0_SRC_MATCH_FOUND) != 0;
+    if (flag)
+    {
+        HWREG(RFCORE_SFR_RFIRQF0) &= ~RFCORE_SFR_RFIRQF0_SRC_MATCH_FOUND;
+    }
```
librasungirl: is it necessary?

simonlingoogle: yes. the function is supposed to read and clear the flag

librasungirl: what's the impact if not?

would this flag would be refreshed per receive action by the PHY?

simonlingoogle: I don't think so

simonlingoogle: @librasungirl verify from test on CC2538 devices that if the flag is not cleared, it will be always 1 once it is set.

Issue #4200, opened 2019-09-22
[mac] assert seen while SED performed data poll


LuDuda:
In our stress tests where nodes constantly send some packets quickly (mostly ICMPv6 multicast ones) several SEDs end up with the same assert [here](https://github.com/openthread/openthread/blob/6fd3a24a43de602ac7b22a391a761beb107cf870/src/core/mac/mac.cpp#L1266). I believe this assert has been introduced with #3915. The SED devices have a polling interval of around 5s and they try to send frames with 200ms interval.

Any idea, how this could happen? Because it seems that the platform notifies about TX frame being sent that was not the Data Poll but rather the ICMPv6 ping with the 0xFFFF destination address (however take into account that the firmware was built with Os optimization).

GDB logs:
```
(gdb) target remote :54517
Remote debugging using :54517
0x00007e30 in ot::Mac::Mac::HandleTransmitDone (this=0x2000a240 <ot::gInstanceRaw+37976>, aFrame=..., aAckFrame=0x0 <__isr_vector>, aError=<optimized out>) at /build/openthread/examples/../src/core/mac/mac.cpp:1266
1266	/build/openthread/examples/../src/core/mac/mac.cpp: No such file or directory.
(gdb) where
#0  0x00007e30 in ot::Mac::Mac::HandleTransmitDone (this=0x2000a240 <ot::gInstanceRaw+37976>, aFrame=..., aAckFrame=0x0 <__isr_vector>, aError=<optimized out>) at /build/openthread/examples/../src/core/mac/mac.cpp:1266
#1  0x00007e7c in ot::Mac::Mac::HandleTransmitDone (this=0x2000a240 <ot::gInstanceRaw+37976>, aFrame=..., aAckFrame=0x0 <__isr_vector>, aError=<optimized out>) at /build/openthread/examples/../src/core/mac/mac.cpp:1307
#2  0x00017f9a in nrf5RadioProcess (aInstance=aInstance@entry=0x20000de8 <ot::gInstanceRaw>) at /build/openthread/examples/../examples/platforms/nrf52840/radio.c:626
#3  0x0001834c in otSysProcessDrivers (aInstance=aInstance@entry=0x20000de8 <ot::gInstanceRaw>) at /build/openthread/examples/../examples/platforms/nrf52840/system.c:137
#4  0x000002f8 in main (argc=0, argv=0x0 <__isr_vector>) at /build/openthread/examples/../examples/apps/cli/main.c:110
(gdb) info locals
[...]

PR #4264, opened 2019-10-22
[mac] check for no error on TxDone check for data poll frame

abtink:
This commit changes the `assert` check for frame to have ack-request
bit set in `HandleTransmitDone()` when performing the operation
`kOperationTransmitPoll` to ensure there is no error. This handles
the case where the frame would not be prepared when a poll tx is
aborted.

### DIFF
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -1286,7 +1286,7 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
         break;
 
     case kOperationTransmitPoll:
-        assert(aFrame.GetAckRequest());
+        assert((aError != OT_ERROR_NONE) || aFrame.GetAckRequest());
 
         if ((aError == OT_ERROR_NONE) && (aAckFrame != NULL))
         {


abtink:
Should help address issue: https://github.com/openthread/openthread/issues/4200

src/core/mac/mac.cpp
```diff
@@ -1286,7 +1286,7 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
         break;
 
     case kOperationTransmitPoll:
-        assert(aFrame.GetAckRequest());
+        assert((aError != OT_ERROR_NONE) || aFrame.GetAckRequest());
```
simonlingoogle: Do we also need to check for errors in the logic above because it also used `aFrame`?
```
    // Determine whether to re-transmit a broadcast frame.
    aFrame.GetDstAddr(dstAddr);	    aFrame.GetDstAddr(dstAddr);
    ...
```

abtink: Yes. But that' a bit tricker. Will need to think and we can probably address it from another PR. 

On other errors like CCA failure, we do want to resend a broadcast frame (that's the point of the code to ensure we send broadcast more than once to make it more reliable).

The problem I see is that we may get an `ABORT` from two different paths. Radio platform may give us an `ABORT` (in this case a broadcast frame is valid and prepared and I think should be re-transmitted) or the frame may not be prepared in first place where `Mac` itself calls `HandleTransmitDone()` with `ABORT` (this case the frame is not prepared and is the previous/older frame and should not be re-transmitted).

abtink: Closing this PR in favor of https://github.com/openthread/openthread/pull/4265.

PR #4265, opened 2019-10-22
[mac] set tx frame as empty when tx aborted during frame preparation

abtink:
The TxDone callbacks can be invoked with `OT_ERROR_ABORT` from two
different paths: Either when OpenThread itself cannot prepare the tx
frame (e.g., message was removed while waiting for MAC to handle a tx
request and ask next layers to prepare the frame) or when the radio
platform itself need to abort the tx.

This commit changes the code such that in the first case, the frame
length is set to zero to mark it as empty. The empty frame helps
differentiate between the two cases and ensures that a previously tx
frame is not incorrectly used from the TxDone callbacks.

The TxDone callback handlers (in `MeshForwarder`, `DataPollHanlder`,
`IndirectSender` and `DataPollHandler` are updated to check for frame
not being empty when processing the frame.

### DIFF
--- a/src/core/mac/data_poll_handler.cpp
+++ b/src/core/mac/data_poll_handler.cpp
@@ -250,7 +250,7 @@ void DataPollHandler::HandleSentFrame(const Mac::TxFrame &aFrame, otError aError
             ExitNow();
         }
 
-        if (aChild.GetIndirectTxAttempts() < kMaxPollTriggeredTxAttempts)
+        if ((aChild.GetIndirectTxAttempts() < kMaxPollTriggeredTxAttempts) && !aFrame.IsEmpty())
         {
             // We save the frame counter, key id, and data sequence number of
             // current frame so we use the same values for the retransmission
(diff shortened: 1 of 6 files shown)


abtink:
Should help address issue https://github.com/openthread/openthread/issues/4200.

abtink:
Some of the `toranj` tests seem to be failing with this change. Will investigate.

abtink:
Pushed new commit which should address the failing test issue (wrong comparison for `IsEmpty`).

simonlingoogle:
> Pushed new commit which should address the failing test issue (wrong comparison for `IsEmpty`).

What did you actually modify in the last commit ?

abtink:
@simonlingoogle, don't want to admit it ;).. but first draft `Frame::IsEmpty()` was incorrect (was returning `(mLength != 0)` instead of `(mLength == 0)`). Good that one of the `toranj` test cases caught it.

PR #2508, opened 2018-01-24
[posix] implement virtual time and update tests to utilize virtual time

jwhui:
Utilizing virtual time reduces Travis' test execution time by about 20 minutes.

This commit adds the following features:

- Example posix drivers to support virtual time simulation, allowing tests
  to run faster than real time.

- A python-based simulation driver that coordinates the virtual time
  simulations.  Each alarm event and message transmission across all nodes
  are scheduled in a single discrete time event queue.  As a result, only
  a single node is making forward progress at a time.

- Updated tests to utilize new virtual-time simulator, allowing them to run
  faster than real time.

- To enable virtual time feature, define VIRTUAL_TIME=1 env var and set
  CPPFLAGS=-DOPENTHREAD_POSIX_VIRTUAL_TIME=1.

PR #4268, opened 2019-10-23
[simulation] enhance simulation for OTNS

simonlingoogle:
The commits make OpenThread works with OTNS by:

- [x] raising the max number of simulating nodes to 999 if OTNS=1
- [x] **implementing the `status-push` mechanism**

Build:
```bash
$ make -f examples/Makefile-simulation OTNS=1
```

This PR have no change of OpenThread behaviors on platforms except `simulation`. 

I will create a Travis test for OpenThread to verify if OpenThread works well with OTNS after OTNS is published.


src/cli/cli_otns.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+    SuccessOrExit(error = otCoapMessageSetPayloadMarker(message));
+
+    sense_data_t sense_data;
+    assert(otns.mCollectionTreeSensorId > 0);
+    sense_data.sensor_id  = otns.mCollectionTreeSensorId;
+    sense_data.milli_time = otPlatAlarmMilliGetNow();
+    sense_data.sense_val  = 0xDDCCBBAA;
+
+    SuccessOrExit(error = otMessageAppend(message, &sense_data, sizeof(sense_data_t)));
```
wgtdkp: Is it possible that the receiver of this message could be another host? If true, I think there will be endianness problem. If false, I still don't think it is a good idea to send data in platform-dependent form through network.

examples/Makefile-posix
```diff
@@ -90,6 +90,10 @@ else
 COMMONCFLAGS                   += -DOPENTHREAD_CONFIG_NCP_UART_ENABLE=1
 endif # NCP_SPI == 1
 
+ifeq ($(OTNS),1)
+VIRTUAL_TIME ?= 1
```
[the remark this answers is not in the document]
simonlingoogle: I was tending to use option 2. Better to force in make, thanks.

examples/platforms/posix/platform-posix.h
```diff
@@ -66,6 +66,8 @@ enum
     OT_SIM_EVENT_RADIO_RECEIVED     = 1,
     OT_SIM_EVENT_UART_WRITE         = 2,
     OT_SIM_EVENT_RADIO_SPINEL_WRITE = 3,
+    OT_SIM_EVENT_POST_CMD           = 4,
```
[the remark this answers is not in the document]
simonlingoogle: cmd=4 is actually not introduced by OTNS. I learnt it from https://github.com/openthread/openthread/blob/c36fa453f1726f4e5f810264ede54a56b75252aa/tests/scripts/thread-cert/simulator.py#L123 which is submitted by @bukepo

examples/platforms/posix/sim/otns.c
```diff
@@ -0,0 +1,66 @@
...
+    va_list ap;
+    va_start(ap, fmt);
+
+    n = vsnprintf(maxline, sizeof(maxline), fmt, ap);
+
+    va_end(ap);
+
+    event.mDelay      = 0;
+    event.mEvent      = OT_SIM_EVENT_STATUS_PUSH;
+    event.mDataLength = (uint16_t)n;
```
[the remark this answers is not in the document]
simonlingoogle: By `fail`, do you mean the formatted string exceed capacity of `maxline`? When will `vsnprintf` return `n < 0`?

include/openthread/platform/otns.h
```diff
@@ -0,0 +1,54 @@
...
+#ifndef OPENTHREAD_PLATFORM_OTNS_H_
+#define OPENTHREAD_PLATFORM_OTNS_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void otPlatOtnsStatusPush(const char *aFmt, ...);
+
+#if OPENTHREAD_CONFIG_OTNS_ENABLE
```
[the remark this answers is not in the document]
simonlingoogle: This should be moved to internal header. 👎

examples/platforms/posix/sim/otns.c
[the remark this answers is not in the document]
wgtdkp: No. I am not sure when `vsnprintf` will fail. Maybe by providing invalid format string. But the function definition says it may fail, then it may fail. I believe it will not fail if you are providing with correct arguments. So, an assertion here is acceptable to me.

src/cli/cli_otns.cpp
```diff
@@ -0,0 +1,340 @@
...
+        ExitNow(error = OT_ERROR_NONE);
+    }
+
+exit:
+    return error;
+}
+
+otError OTNS::Process(int argc, char *argv[])
+{
+    otError error = OT_ERROR_PARSE;
```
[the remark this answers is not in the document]
simonlingoogle: I think `OT_ERROR_PARSE` is ok here. `udp` command uses `OT_ERROR_PARSE` too.

examples/platforms/simulation/platform-simulation.h
```diff
@@ -66,6 +66,8 @@ enum
     OT_SIM_EVENT_RADIO_RECEIVED     = 1,
     OT_SIM_EVENT_UART_WRITE         = 2,
     OT_SIM_EVENT_RADIO_SPINEL_WRITE = 3,
+    OT_SIM_EVENT_POST_CMD           = 4,
+    OT_SIM_EVENT_STATUS_PUSH        = 5,
```
wgtdkp:
```suggestion
    OT_SIM_EVENT_OTNS_STATUS_PUSH   = 5,
```

For a new reader, I'd like to know more context about this event (`STATUS` would be too general). Besides, this event is introduced by OTNS and used by `OTNS` exclusively, I think it is better to add this `OTNS` prefix. Thoughts?

We also need a clearer name for `OT_SIM_EVENT_POST_CMD`, but since it is not introduced by OTNS, we can leave it for future enhancements.

simonlingoogle: agree.

src/cli/cli_otns.cpp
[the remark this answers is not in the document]
wgtdkp: Not agree. Please consider in which case the final error value will be the initial value: when we can't find the input command in `sCommands`. This is definitely invalid arguments error, especially in the context of CLI.

for `cli_udp.cpp`, I can fire a PR to correct this.

examples/platforms/posix/sim/otns.c
```diff
@@ -0,0 +1,66 @@
...
+
+#include <openthread/platform/alarm-micro.h>
+#include <openthread/platform/alarm-milli.h>
+#include <openthread/platform/diag.h>
+
+#if OPENTHREAD_POSIX_VIRTUAL_TIME == 0
+#error "OTNS does not work with VIRTUAL_TIME=0"
+#endif
+
+void otPlatOtnsStatusPush(const char *fmt, ...)
```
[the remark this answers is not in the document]
simonlingoogle: changed to `aFmt`

src/cli/cli_otns.cpp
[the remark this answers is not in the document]
simonlingoogle: Resolving it in another PR sounds good to me

src/core/api/thread_api.cpp
```diff
@@ -97,6 +97,15 @@ otError otThreadGetLeaderRloc(otInstance *aInstance, otIp6Address *aLeaderRloc)
     return instance.Get<Mle::MleRouter>().GetLeaderAddress(*static_cast<Ip6::Address *>(aLeaderRloc));
 }
 
+otError otThreadGetLeaderAloc(otInstance *aInstance, otIp6Address *aLeaderAloc)
+{
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
+    assert(aLeaderAloc != NULL);
```
librasungirl: in the function declaration, here should return `OT_ERROR_INVALID_ARGS`?
https://github.com/openthread/openthread/pull/4268/files#diff-e01913b0351246189e44602cc26bcfabR338

src/cli/cli_otns.cpp
wgtdkp: Why mark this resolved?

src/cli/cli_otns.cpp
```diff
@@ -0,0 +1,345 @@
...
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file implements a simple CLI for the CoAP service.
```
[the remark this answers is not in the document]
simonlingoogle: fixed

src/cli/cli.cpp
```diff
@@ -2113,6 +2124,9 @@ void Interpreter::SendPing(void)
     SuccessOrExit(otMessageAppend(message, &timestamp, sizeof(timestamp)));
     SuccessOrExit(otMessageSetLength(message, mPingLength));
     SuccessOrExit(otIcmp6SendEchoRequest(mInstance, message, &messageInfo, mPingIdentifier));
+    OtnsStatusPush("ping_request=%s$%d$%lld",
+                   static_cast<Ip6::MessageInfo *>(&messageInfo)->GetPeerAddr().ToString().AsCString(), mPingLength,
+                   int64_t(HostSwap32(timestamp)));
```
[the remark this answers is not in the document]
simonlingoogle: I converted it to `int64_t`, so it is `%lld`. But I will change it to `%lu` and remove conversion to `int64_t`

src/cli/cli_otns.cpp
simonlingoogle: Sorry, only noticed 0xDDCCBBAA.

src/core/api/thread_api.cpp
simonlingoogle: Good catch!

simonlingoogle: We have same issue for `otThreadGetLeaderRloc` and `otThreadGetNextNeighborInfo`. I will just fix `otThreadGetLeaderAloc` here.

examples/platforms/simulation/virtual_time/otns.c
```diff
@@ -0,0 +1,69 @@
...
+#include <string.h>
+
+#include <openthread/platform/alarm-micro.h>
+#include <openthread/platform/alarm-milli.h>
+#include <openthread/platform/diag.h>
+#include <openthread/platform/otns.h>
+
+#if OPENTHREAD_SIMULATION_VIRTUAL_TIME == 0
+#error "OTNS must use VIRTUAL_TIME=1"
+#endif
```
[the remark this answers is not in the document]
simonlingoogle: Now `otPlatOtnsStatusPush` is defined only when virtual time is on.

simonlingoogle:
> Why do we choose little endian here? I believe it is more common to use big endian for network transformed data ([network order](https://tools.ietf.org/html/draft-newman-network-byte-order-01)). Besides, Thread TLVs all use big endian, it would be better we align to it.

So that I don't break current OTNS code because OTNS is expecting little-endian values.

simonlingoogle:
> Why do we choose little endian here? I believe it is more common to use big endian for network transformed data ([network order](https://tools.ietf.org/html/draft-newman-network-byte-order-01)). Besides, Thread TLVs all use big endian, it would be better we align to it.

Switched to big endian.

src/cli/cli_otns.cpp
```diff
@@ -0,0 +1,283 @@
...
+    assert(otCoapMessageGetType(aMessage) == OT_COAP_TYPE_CONFIRMABLE);
+
+    assert(otMessageGetOffset(aMessage) + sizeof(SenseData) == otMessageGetLength(aMessage));
+    VerifyOrExit(otMessageRead(aMessage, otMessageGetOffset(aMessage), &senseData, sizeof(SenseData)) ==
+                 sizeof(SenseData));
+
+    OtnsStatusPush("ctss=%lx:%lx:%lx", HostSwap32(senseData.sensorId), HostSwap32(senseData.milliTime),
+                   HostSwap32(senseData.senseVal));
+
+    responseCode = OT_COAP_CODE_VALID;
```
wgtdkp:
```suggestion
    responseCode = OT_COAP_CODE_CHANGED;
```

src/cli/cli_otns.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+    responseMessage = otCoapNewMessage(mInterpreter.mInstance, NULL);
+    VerifyOrExit(responseMessage != NULL, error = OT_ERROR_NO_BUFS);
+
+    otCoapMessageInitResponse(responseMessage, aMessage, OT_COAP_TYPE_ACKNOWLEDGMENT, responseCode);
+    SuccessOrExit(error = otCoapMessageSetToken(responseMessage, otCoapMessageGetToken(aMessage),
+                                                otCoapMessageGetTokenLength(aMessage)));
+
+    if (otCoapMessageGetCode(aMessage) == OT_COAP_CODE_GET)
```
wgtdkp:
```suggestion
    VerifyOrExit(otCoapMessageGetCode(aMessage) == OT_COAP_CODE_POST, error == OT_ERROR_PARSE);
```
And move this check to the beginning of this function?

src/cli/cli_otns.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+exit:
+
+    if (error != OT_ERROR_NONE)
+    {
+        if (responseMessage != NULL)
+        {
+            otMessageFree(responseMessage);
+        }
+        otLogCritCoap("coap send response error: %d %s", error, otThreadErrorToString(error));
```
wgtdkp:
```suggestion
        otLogCritCoap("coap handle %s failed: %d %s", OTNS_URI_PATH_COLLECTION_TREE, error, otThreadErrorToString(error));
```

src/cli/cli_otns.hpp
```diff
@@ -0,0 +1,118 @@
...
+namespace Cli {
+
+#define OTNS_URI_PATH_COLLECTION_TREE "otns/ct"
+
+/**
+ * This structure represents message payload used by Collection Tree
+ *
+ */
+OT_TOOL_PACKED_BEGIN
+struct SenseData
```
wgtdkp: There seems no necessary to put this structure in header file. How about moving to `cli_otns.cpp`?

tests/scripts/thread-cert/simulator.py
```diff
@@ -132,7 +132,7 @@ class VirtualTime(BaseSimulator):
     BASE_PORT = 9000
     MAX_NODES = 34
     MAX_MESSAGE = 1024
-    END_OF_TIME = 0x7FFFFFFF
+    END_OF_TIME = float('inf')
```
wgtdkp: It doesn't seem to matter to OTNS. make it another PR?

include/openthread/platform/otns.h
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @brief
+ *   This module includes the platform abstraction for OTNS.
+ *
+ * @{
+ *
+ */
+
+/**
+ * This function pushes device status to OTNS.
```
wgtdkp: Please document the format of OTNS status message, giving a few examples will be even better.

CMakeLists.txt
```diff
@@ -29,6 +29,8 @@
 cmake_minimum_required(VERSION 3.13.1)
 project(openthread)
 
+add_compile_definitions(OPENTHREAD_CONFIG_OTNS_ENABLE=1)
+
```
simonlingoogle:
```suggestion
```

examples/common-switches.mk
```diff
@@ -250,6 +251,15 @@ else
 COMMONCFLAGS += -DOPENTHREAD_SETTINGS_RAM=0
 endif
 
+ifeq ($(OTNS),1)
+
+ifneq ($(VIRTUAL_TIME),1)
+$(error OTNS only supports virtual time simulation.)
```
librasungirl: move option dependency check into https://github.com/openthread/openthread/blob/master/src/core/config/openthread-core-config-check.h?

examples/platforms/simulation/virtual_time/otns.c
```diff
@@ -0,0 +1,56 @@
...
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "platform-simulation.h"
+
+#if OPENTHREAD_CONFIG_OTNS_ENABLE && OPENTHREAD_SIMULATION_VIRTUAL_TIME
```
librasungirl: should be safe to only keep `OPENTHREAD_CONFIG_OTNS_ENABLE` if already checked?

examples/common-switches.mk
simonlingoogle: There are currently two checks required for OTNS:
- VIRTUAL_TIME must be 1
- VIRTUAL_TIME_UART must be 0

Currently, the default value of `OPENTHREAD_SIMULATION_VIRTUAL_TIME` and `OPENTHREAD_SIMULATION_VIRTUAL_TIME_UART` are both defined in `examples/simulation/platform-config.h`, which however is not included by `openthread-core-config-check.h`. So, we can not check these two MACROs in `openthread-core-config-check.h`. 

So, I put the check in `examples/simulation/platform-config.h`. 

Thoughts?

examples/common-switches.mk
```diff
@@ -69,6 +69,7 @@ SNTP_CLIENT         ?= 0
 THREAD_VERSION      ?= 1.1
 TIME_SYNC           ?= 0
 UDP_FORWARD         ?= 0
+OTNS                ?= 0
```
jwhui: nit: maintain alphabetical order

include/openthread/platform/otns.h
```diff
@@ -0,0 +1,77 @@
...
+ * e.x. ping_request=fdde:ad00:beef:0:459e:d7b4:b65e:5480,4,112000.
+ *
+ * New statuses should follow these conventions.
+ *
+ * @param[in]  aFmt   A pointer to the format string.
+ * @param[in]  ...    A matching list of arguments.
+ *
+ * Currently, OTNS only supports virtual time simulation.
+ */
+void otPlatOtnsStatusPush(const char *aFmt, ...);
```
jwhui: How about rename to `otPlatOtnsStatus()`? It should be clear that this is "pushing" information to OTNS.

include/openthread/platform/otns.h
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @brief
+ *   This module includes the platform abstraction for OTNS.
+ *
+ * @{
+ *
+ */
+
+/**
+ * OpenThread uses `OtnsStatusPush` to push node status to OTNS.
```
jwhui: First line should be a brief description that stands alone. How about something like: "This function exports status information to OTNS."

An empty line should follow the brief.

examples/common-switches.mk
librasungirl: For platform specific options check, one existing example I can find is in the Makefile
https://github.com/openthread/openthread/blob/master/examples/Makefile-nrf52840#L81-L83

Maybe here the check could be put into `Makefile-simulation`?
Besides, `OTNS` seems also a platform specific option - only for simulation, maybe also move to the  Makefile?

Thoughts? @jwhui @simonlingoogle

examples/common-switches.mk
simonlingoogle: will do

examples/common-switches.mk
simonlingoogle: Putting checks to `Makefile-simualtion` sounds good to me.

src/core/thread/mle.cpp
```diff
@@ -346,6 +347,8 @@ void Mle::SetRole(otDeviceRole aRole)
         mParent.SetState(Neighbor::kStateInvalid);
     }
 
+    OtnsStatus("role=%d", mRole);
```
abtink: I am not in favor of emitting `OtnsStatus()` from within the OT core modules (as an example here or from `SubMac` for the ext address change, etc).  I think this can create dependencies to the tests that would be hard to maintain in future. The core code can/will change but we try to keep the external/public APIs as before and backwards-compatible. I think instead it would be better to use the public APIs and/or existing eventing mechanism to get info.

For example here, can we consider using the existing notifier `OT_CHANGED_THREAD_ROLE` event model?
An OTNS module can register a callback to `Notifier` (e.g., using `otSetStateChangedCallback()` or a direct register defining `Notifier::Callback`) and process the `OT_CHANGED_THREAD_ROLE` event  and emit the `"role=%d", mRole` from callback. With this approach the otns event may not be as immediate as a direct emit from `Mle::SetRole()`  (due to `Notifier` emitting its events from a posted `Tasklet`), but I don't expect it would matter for building test cases.

I would say if there are some internal states that we currently don't expose out, it would be better to add the appropriate APIs or events for those instead.

Thoughts?

src/core/thread/mle.cpp
```diff
@@ -1018,6 +1023,8 @@ void Mle::SetLeaderData(uint32_t aPartitionId, uint8_t aWeighting, uint8_t aLead
     mLeaderData.SetPartitionId(aPartitionId);
     mLeaderData.SetWeighting(aWeighting);
     mLeaderData.SetLeaderRouterId(aLeaderRouterId);
+
+    OtnsStatus("parid=%x;lrid=%x", aPartitionId, aLeaderRouterId);
```
abtink: Can we use the `OT_CHANGED_THREAD_PARTITION_ID`?

src/core/thread/mle_router.cpp
```diff
@@ -988,6 +989,7 @@ otError MleRouter::HandleLinkAccept(const Message &         aMessage,
     router->SetKeySequence(aKeySequence);
 
     Signal(OT_NEIGHBOR_TABLE_EVENT_ROUTER_ADDED, *router);
+    OtnsStatus("router_added=%s", router->GetExtAddress().ToString().AsCString());
```
abtink: Can we use either `OT_NEIGHBOR_TABLE_EVENT_ROUTER_ADDED`? or `otThreadRegisterNeighborTableCallback()`

src/core/thread/mle_router.cpp
```diff
@@ -3261,6 +3263,7 @@ void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
         if (aNeighbor.IsStateValidOrRestoring())
         {
             Signal(OT_NEIGHBOR_TABLE_EVENT_CHILD_REMOVED, aNeighbor);
+            OtnsStatus("child_removed=%s", aNeighbor.GetExtAddress().ToString().AsCString());
```
abtink: Same as above

src/core/thread/mle.cpp
abtink: > 2.Notifier is using Tasklets which are posted rather than executed immediately. The virtual time simulator implemented in #2508 schedules all node events and transmissions to preserve the execution order in time. However, using Tasklets to send events breaks the strict ordering of events, thus could have impact on the correctness of simulation.

Can you please help explain why we need immediate eventing for OTNS to work (e.g., need to know exactly when RLOC16 or role value changes).
- The virtual time simulator in OT is emulating the platform layer and events are related to radio/timer, etc. 
- OpenThread core itself can be considered a big state machine which requires calls from platform layer and its public APIs to proceed, take any action and  move from one state to another.  
- So a simulator pretty much knows and is in control of when OT state changes (since it would be calling the platform/public APIs that cause OT to change). 
- A higher level simulator  can then use existing public APIs to query info/state from OT (e.g., check the current role using `otThreadGetDeviceRole()`). This can be done after every call that would change OT. 
- With such a model we should not need any immediate evening from OT. Can we  consider adopting such a model?

> The calls to OtnsStatus are placed exactly where callbacks are or should be. So, I would expect maintaining OtnsStatus calls simple because we will also have to maintain callback code as core code changes. We just need to make OtnsStatus follow corresponding callback signaling.

My reservation is for this exact purpose. For now the `OtnsStatus` may be placed where the callbacks are, but if we allow and use `OtnsStatus` directly in the core modules, it may be used in places where there is no callback and adds dependency in the code. 

I would highly suggest we funnel OTNS events through existing public APIs and/or adding new APIs/mechanism.

simonlingoogle: > Can you please help explain why we need immediate eventing for OTNS to work (e.g., need to know exactly when RLOC16 or role value changes).

One of the motivation for OTNS to use status emitting to acquire node status rather than calling APIs is because the **node status affects the simulation result immediately**. 
For example, OTNS uses node `rloc16` and `extaddr` for destination address filtering so that it only dispatches unicast messages to the right destinations. This is one of the reasons that `OTNS` outperforms existing virtual time simulator by tens of times. 

Using `Tasklets` for event emitting changes the order of events, e.x. message transmission, extaddr/rloc16 change, etc.  If we are to implement MAC filters in OTNS, the order of message transmission events and `extaddr` change events could cause `OTNS` to incorrectly drop messages. 

In general, I am worried about the disordering of events caused by `posted Tasklets` because order of events impacts `OTNS` simulation result. 

> A higher level simulator can then use existing public APIs to query info/state from OT (e.g., check the current role using otThreadGetDeviceRole()). This can be done after every call that would change OT.

This however implies that status emitting to `OTNS` must be implemented in one of apps or platforms. However, I would expect `OTNS` to work for cross apps/platforms in the future. For example, with `OTNS` enhancements and proper `otPlatOtnsStatus` implementation on `nrf528xx` platform, we can even use `OTNS` for simulating or visualizing real devices. Integrating `OTNS` mechanisms in the OpenThread core equips `OTNS` with more potentials. 

> but if we allow and use OtnsStatus directly in the core modules, it may be used in places where there is no callback and adds dependency in the code.

I would expect this never to happen that we call `OtnsStatus` in the places where there is no need to call callbacks. `OtnsStatus` should be called whenever the target status changes, in which case, we should also call callbacks. If the target status does not change, we don't need to call `OtnsStatus` or callbacks. Replacing `OtnsStatus` with callbacks should have the exact same overhead of maintaining code dependency as core code changes. 

---
With all these concerns, I am not sure using registered callbacks is suitable for status emitting to `OTNS`. For now, it seems best and logical to me to integrate `OTNS` utilities to OT core and expect future improvements when we gain more experience.

abtink: Thanks @simonlingoogle. My goal is understand the approach and help.

> One of the motivation for OTNS to use status emitting to acquire node status rather than calling APIs is because the node status affects the simulation result immediately.
> For example, OTNS uses node rloc16 and extaddr for destination address filtering so that it only dispatches unicast messages to the right destinations. This is one of the reasons that OTNS outperforms existing virtual time simulator by tens of times.

I think this is simulation of radio platform layer, and I agree filtering frames during simulation and dispatching the frame only to the actual dest  is a very good improvement and optimization. 

One thing that is not clear to me is why we would need the OT core to emit events about rloc16 and/or extended address changes to get such optimization. 
- We already have radio platform APIs that inform radio platform when ext address or short address changed (`otPlatRadioSetExtendedAddress()` and  `otPlatRadioSetShortAddress()`). 
- These APIs are indeed used by radio drivers to change and filter on actual hardware (btw, we have other parameters like channel and PAN Id which are also set by OT on radio platform layer and used for frame filtering during rx).
-  OTNS is expected to simulate the radio (and rest of platform layer). If OTNS needs events such as  "short/ext address changed" to update how it would filter/behave, I would say we should emit such events from the platform layer implementation of these APIs and not from OT core. 
- This would be actually immediate and no issue with ordering (or waiting for posted tasklets).
- When OT core calls `otPlatRadioSetShortAddress()` is the exact moment from which the radio simulation should start using the new given short address and node receive only frames for that address. So the ONTS event should be emitted from this function call vs. emitting it from MLE layer method `Mle::SetRlco16()` which is technically  many layers higher in the network stack.

Is there any reason why we cannot do this from the platform layer code and need it to be emitted from OT core?

-----

I would say for any state/parameter changes that is needed for simulation of platform layer (e.g., short/ext address, pan id, channel, etc), we should already have a platform API to set it on platform layer. We should be able to emit events to OTNS simulator from the platform implementation.

I guess/think other parameter/state changes (say, role or partition id, child attached etc) which are not directly used/needed for simulation of platform layer, are intended for visualization by OTNS. Is that correct? 
- If so, does the ordering of events matter for visualization? 
- I feel it should not and then I would suggest using existing notification model (`Notifier` and state changed callbacks) and/or public APIs to query for visualization purpose 
  -  Of course we should improve/add any missing features. 
  - For example if certain changes that would be useful for visualization don't have a related `OT_CHANGED` events we can add them;
  - Or enhance the Child/Neighbor table callbacks to allow more than one registred callback so that one can be used by OTNS for visualization).

simonlingoogle: > Is there any reason why we cannot do this from the platform layer code and need it to be emitted from OT core?

Yes, we can implement the status emitting to `OTNS` in platform code, after following enhancements to OT core:
* Add missing callbacks that is used by `OTNS`
* Enhance each callback to support multiple callbacks
* Enhance Child/Neighbor callbacks to be able to pass the corresponding `ExtAddress`

However, with such effort and increased code complexity, what we get is a strict binding of `OTNS` to the simulation platform, which makes it difficult to port `OTNS` to other platforms. 

Meanwhile, the current implementation has all logic code ready in OT core and requires only one platform function `otPlatOtnsStatus` to be implemented. For example, we can implement `otPlatOtnsStatus` for `NRF52840 Dongle` to emit the status information through UART to visualize devices that connected to the host computer. 

Overall, I think using callbacks in platform code does not have strong benefits, but impose restriction to future `OTNS` enhancements. 

---
Currently, `OtnsStatus` calls are scattered in OT core, which I think should be refactored.  **It would be better if we can use one core component just for `OTNS` related code.** However there is currently no mechanism for signalling synchronous events among core components.

examples/platforms/simulation/platform-simulation.h
```diff
@@ -66,6 +66,8 @@ enum
     OT_SIM_EVENT_RADIO_RECEIVED     = 1,
     OT_SIM_EVENT_UART_WRITE         = 2,
     OT_SIM_EVENT_RADIO_SPINEL_WRITE = 3,
+    OT_SIM_EVENT_POST_CMD           = 4,
```
librasungirl: not used?

simonlingoogle: It is not introduced by `OTNS`. However it is defined in existing virtual time simulator in python code. I add `OT_SIM_EVENT_POST_CMD           = 4,` here so that there won't be a gap between 3 and 5.

src/core/thread/mle.cpp
abtink: @simonlingoogle , I know it may feel that the current model of emitting `otnsStatus()` from different modules within OT is simple and works fine (and tested), and seems like changing the approach to use `Notifier` and other callbacks is adding complexity and more work, but my advice is it is always better to spend more time at start of a new solution and ensure to have a good approach/design (which is better suited for future) than to try to adapt/change it later on. 

> However, with such effort and increased code complexity, what we get is a strict binding of OTNS to the simulation platform, which **makes it difficult to port OTNS to other platforms**.

I don't follow this point about "hard to port on other platforms". From what I understand OTNS has two aspects, simulation (of platform layer) and visualization (or nodes, etc). I feel we should treat these differently.
- I expect OTNS simulation aspect to  only apply when we want to emulate/simulate the platform layer and in that case we would need to know when short/ext address changes, etc.  So it should be fine to emit the events from `otPlatRadio` implementation under simulation platform. 
- Visualization aspect I guess can be used with simulated nodes or with nodes running on actual hardware but if we are using actual hardware then their radio driver would provide MAC filtering and we would not need to know when short/ext address changes. The rest of events needed for visualization only can still use `otnsStatus` model and be common for all platforms.
  - If in future you are considering hybrid cases where some nodes are simulated and some are actual device, we need to update the radio platform code for this anyways and can still emit events for OTNS from the `otPlatRadio` functions. 

> Enhance Child/Neighbor callbacks to be able to pass the corresponding ExtAddress

The [`otNeighborTableCallback` mechanism](https://github.com/openthread/openthread/blob/35bb0027d0b6d2b1b2e54a289bb3e3f56def57d0/include/openthread/thread_ftd.h#L599-L631) already provides full info about neighbor/child being added or removed as `otChildInfo`/`otNeighborInfo` entry (which contains ext address and many other parameters) and it is an immediate callback. 


----------------

Here is my suggestions on how to hook OTNS into OT:

We should divide the events into the ones needed for simulation and the ones for visualization and use different approach for them.

For events required for simulation, I suggest two options:
1. We can use `otPlat...` calls and emit the event to OTNS from them  (this would be my preferred approach).
1. If you still want it to be in core, I would suggest to limit calls to OTNS for simulation from lowest layer of code in OT, for example from the `Radio.hpp` method (e.g., `Radio::SetShortAddress()` for emitting short address change event to OTNS).

For events for visualization (role change, joiner state change, child/neighbor added, etc)
- We can make OTNS a utility like module in under `src/core/util`. 
   - It can either register a callback to `Notifier` and from callback it can emit the events in the string format required by OTNS.
   - If we do really need immediate callback (cannot wait for posted tasklet of Notifier mechanism - which I expect we should not need for visualization), the approach I would suggest is to add a hook in `Notifier::Signal()` to call a method in `OTNS` which would then immediately process the event and emit the related string format event to OTNS. This way we limit the interaction of OTNS with other OT_CHANGED ones (which are tied to public APIs of OT).
-  For neighbor/child table updates, I see two options:
   - Update the [`otNeighborTableCallback` mechanism](https://github.com/openthread/openthread/blob/35bb0027d0b6d2b1b2e54a289bb3e3f56def57d0/include/openthread/thread_ftd.h#L599-L631) to allow more than one callback and register a callbacks from OTNS and process/emit events from there .
   - Or adding a hook in `MleRouter::Siganl()` method to call  into OTNS (note that `Signal()` already has a reference to `Neighbor` entry being added/removed).

examples/platforms/simulation/platform-simulation.h
librasungirl: How about using the consistent enumeration name `OT_SIM_EVENT_POSTCMD`?

simonlingoogle: I removed it because it's not introduced by `OTNS`

src/core/thread/mle.cpp
simonlingoogle: @abtink any further comments on the new implementation?

src/core/meshcop/joiner.cpp
```diff
@@ -75,11 +76,9 @@ void Joiner::GetJoinerId(Mac::ExtAddress &aJoinerId) const
 
 void Joiner::SetState(otJoinerState aState)
 {
-    VerifyOrExit(aState != mState);
+    SuccessOrExit(Get<Notifier>().Update(mState, aState, OT_CHANGED_JOINER_STATE));
 
     otLogInfoMeshCoP("JoinerState: %s -> %s", JoinerStateToString(mState), JoinerStateToString(aState));
```
abtink: Need to save the old value for logging to be correct. With changing the `mState` value first through `Update()` call, this log will now show the same value for `mState` and `aState`.

src/core/thread/mle.cpp
abtink: The new model looks great. Thanks 👍

src/core/utils/otns.hpp
```diff
@@ -0,0 +1,138 @@
...
+#include "thread/topology.hpp"
+
+namespace ot {
+namespace Utils {
+
+/**
+ * This class implements the OTNS Stub that interacts with OTNS.
+ *
+ */
+class OtnsStub : public InstanceLocator
```
abtink: Can we consider removing the `Stub` (how about calling the class `Otns`?). The "stub" term is usually used as a [temporary/stand-in](https://en.wikipedia.org/wiki/Method_stub) (rather than main implementation).

examples/Makefile-simulation
```diff
@@ -91,6 +91,19 @@ else
 COMMONCFLAGS                   += -DOPENTHREAD_CONFIG_NCP_UART_ENABLE=1
 endif # NCP_SPI == 1
 
+ifeq ($(OTNS),1)
+VIRTUAL_TIME ?= 1
+
+ifneq ($(VIRTUAL_TIME),1)
+$(error OTNS requires virtual time simulations)
```
jwhui: Should we move these checks to a `examples/simulation/openthread-core-simulation-config-check.h`, similar to [examples/platforms/cc2538/openthread-core-cc2538-config-check.h](https://github.com/openthread/openthread/blob/master/examples/platforms/cc2538/openthread-core-cc2538-config-check.h)?

This will help ensure that the checks are included even when a different build system is used.

examples/platforms/simulation/CMakeLists.txt
```diff
@@ -38,6 +38,16 @@ if(OT_SIMULATION_VIRTUAL_TIME_UART)
     list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_SIMULATION_VIRTUAL_TIME_UART=1")
 endif()
 
+if (OT_OTNS)
```
jwhui: See comment above about moving checks to header file.

include/openthread/platform/otns.h
```diff
@@ -0,0 +1,84 @@
...
+ *   This module includes the platform abstraction for OTNS.
+ *
+ * @{
+ *
+ */
+
+/**
+ * This function exports status information to OTNS.
+ *
+ * The status information is represented by a string with format recognizable by OTNS.
```
jwhui: Indicate whether the string needs to be null-terminated.

src/core/utils/otns.hpp
```diff
@@ -0,0 +1,138 @@
...
+     *
+     */
+    explicit OtnsStub(Instance &aInstance)
+        : InstanceLocator(aInstance)
+        , mNotifierCallback(aInstance, &OtnsStub::HandleStateChanged, this)
+    {
+    }
+
+    /**
+     * this function emits radio short address to OTNS when changed.
```
jwhui:
```suggestion
     * This function emits radio short address to OTNS when changed.
```

src/core/utils/otns.hpp
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @param[in]  aShortAddress  The new short address.
+     *
+     */
+    static void EmitShortAddress(uint16_t aShortAddress);
+
+    /**
+     * this function emits radio extended address to OTNS when changed.
```
jwhui:
```suggestion
     * This function emits radio extended address to OTNS when changed.
```

src/core/utils/otns.hpp
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @param[in]  aExtAddress  The new extended address.
+     *
+     */
+    static void EmitExtendedAddress(const Mac::ExtAddress &aExtAddress);
+
+    /**
+     * this function emits ping request information to OTNS when sending.
```
jwhui:
```suggestion
     * This function emits ping request information to OTNS when sending.
```

src/core/utils/otns.hpp
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @param[in]  aPingLength  The data length of the ping request.
```
jwhui: Missing `aPeerAddress`

src/core/utils/otns.hpp
```diff
@@ -0,0 +1,138 @@
...
+     */
+    static void EmitPingRequest(const Ip6::Address &aPeerAddress,
+                                uint16_t            aPingLength,
+                                uint32_t            aTimestamp,
+                                uint8_t             aHopLimit);
+
+    /**
+     * this function emits ping reply information to OTNS when received.
+     *
+     * @param[in]  aPingLength  The data length of the ping reply.
```
jwhui: Missing `aPeerAddress`

src/core/utils/otns.hpp
```diff
@@ -0,0 +1,138 @@
...
+     * @param[in]  aHopLimit    The hop limit of the ping request.
+     *
+     */
+    static void EmitPingRequest(const Ip6::Address &aPeerAddress,
+                                uint16_t            aPingLength,
+                                uint32_t            aTimestamp,
+                                uint8_t             aHopLimit);
+
+    /**
+     * this function emits ping reply information to OTNS when received.
```
jwhui:
```suggestion
     * This function emits ping reply information to OTNS when received.
```

src/core/utils/otns.hpp
```diff
@@ -0,0 +1,138 @@
...
+     * @param[in]  aHopLimit    The hop limit of the ping reply.
+     *
+     */
+    static void EmitPingReply(const Ip6::Address &aPeerAddress,
+                              uint16_t            aPingLength,
+                              uint32_t            aTimestamp,
+                              uint8_t             aHopLimit);
+
+    /**
+     * this function emits a neighbor table event to OTNS when a neighbor is added or removed.
```
jwhui:
```suggestion
     * This function emits a neighbor table event to OTNS when a neighbor is added or removed.
```

src/core/utils/otns.hpp
simonlingoogle: I named it to `OtnsStub` because this module serves as a client stub for interacting with `OTNS`, similar to a gRPC client stub that interacts with the gRPC server. Thoughts?

examples/Makefile-simulation
simonlingoogle: However we can not check `OPENTHREAD_SIMULATION_VIRTUAL_TIME` and `OPENTHREAD_SIMULATION_VIRTUAL_TIME_UART` in `openthread-core-simulation-config-check.h` because these 2 macros are defined in `platform-config.h`, which is not included by `openthread-core-config.h`. Thoughts?

jwhui: How about putting the check in `platform-config.h`?

src/core/utils/otns.hpp
```diff
@@ -0,0 +1,140 @@
...
+    static void EmitPingRequest(const Ip6::Address &aPeerAddress,
+                                uint16_t            aPingLength,
+                                uint32_t            aTimestamp,
+                                uint8_t             aHopLimit);
+
+    /**
+     * This function emits ping reply information to OTNS when received.
+     *
+     * @param[in]  aPeerAddress  The peer address of the ping request.
+     * @param[in]  aPingLength  The data length of the ping reply.
```
jwhui: nit: alignment

examples/Makefile-simulation
simonlingoogle: `platform-config.h` looks good to me. I originally putted the check there, but moved away.

src/core/utils/otns.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    if ((aFlags & OT_CHANGED_JOINER_STATE) != 0)
+    {
+        EmitStatus("joiner_state=%d", Get<MeshCoP::Joiner>().GetState());
+    }
+}
+
+void Otns::EmitNeighborChange(otNeighborTableEvent aEvent, Neighbor &aNeighbor)
+{
+    const char *extAddr = aNeighbor.GetExtAddress().ToString().AsCString();
```
abtink: This won't be correct.

Note that `ToString()` returns a temporary `String` object (think of it as returning an struct containing an array of chars), and `AsCString()` gets the pointer to the array.

A temp object stays valid for the whole enclosing statement/expression (in this case the whole statement is the `const char *extAddr = ...`). After this statement the temp object is destroyed/freed but we will keep a pointer to a freed char buffer in `extAddr`. From c++ [reference](https://en.cppreference.com/w/cpp/language/lifetime#Temporary_object_lifetime): "All temporary objects are destroyed as the last step in evaluating the full-expression that (lexically) contains the point where they were created"

src/core/utils/otns.cpp
```diff
@@ -0,0 +1,136 @@
...
+}
+
+void Otns::EmitPingReply(const Ip6::Address &aPeerAddress, uint16_t aPingLength, uint32_t aTimestamp, uint8_t aHopLimit)
+{
+    EmitStatus("ping_reply=%s,%u,%lu,%d", aPeerAddress.ToString().AsCString(), aPingLength, aTimestamp, aHopLimit);
+}
+
+void Otns::EmitStatus(const char *aFmt, ...)
+{
+    static char statusStr[kMaxStatusStringLength + 1];
```
abtink: Do we need the char buffer to be `static`?

simonlingoogle: Either it's allocated on heap or statically  does not matter. I was worried that if `kMaxStatusStringLength` was defined to a large value, this could overflow the stack.

src/cli/cli.cpp
```diff
@@ -2141,13 +2142,14 @@ void Interpreter::HandleIcmpReceive(otMessage *          aMessage,
                                     const otMessageInfo *aMessageInfo,
                                     const otIcmp6Header *aIcmpHeader)
 {
-    uint32_t timestamp;
+    uint32_t timestamp = 0;
+    uint16_t datasize;
```
wgtdkp:
```suggestion
    uint16_t dataSize;
```

PR #4273, opened 2019-10-24
[efr32] Constructors now called during startup.

marvensilabs:
Removed gcc flags --nostartfiles and __START definition.
Reset_Handler calls mainCRTStartup to call constructors instead of calling main directly.

### DIFF
--- a/examples/Makefile-efr32mg12
+++ b/examples/Makefile-efr32mg12
@@ -138,7 +138,6 @@ CXXFLAGS                       += \
 LDFLAGS                        += \
     $(COMMONCFLAGS)               \
     $(target_LDFLAGS)             \
-    -nostartfiles                 \
     -specs=nano.specs             \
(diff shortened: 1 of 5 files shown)


jwhui:
@marvensilabs, OpenThread should not require calling constructors before `main()` is called. Are you experiencing that OpenThread needs constructors called somewhere? Or is this to address something specific to efr32 platforms?

marvensilabs:
Ok, it is for the EFR32 OpenWeave door lock example app that appears to require constructors to be called prior to main.  I will check if the --nostartupfiles can be removed from just the OpenWeave door lock makefile and see if I need the -D__START=main in OpenThread's makefiles.

marvensilabs:
We will leave the current makefiles as is with --nostartfiles enabled and leave it up to the application to bring in the necessary startup code.  Closing pull request.

PR #4274, opened 2019-10-24
[notifier] add Update method

abtink:
This commit adds a new `Notifier::Update()` method which performs a
commonly used pattern in the code where a variable is updated with a
new given value and if the variable value does change a related
notifier `OT_CHANGED_<NAME>` is signaled.


jwhui:
Can/should we apply this to:

- `KeyManager::SetPskc()`
- `Mac::SetExtendedPanId()`

?

abtink:
Indeed! Will add them.

PR #4277, opened 2019-10-28
[efr32] fix LED gpio port and pin allocation for BRD4166A

marvensilabs:
Fix for Thunderboard Sense 2 (BRD4166A) LED GPIOs
This bug is inherited from SDK v2.6 bspconfig.h which is also incorrect.


gcbutt:
@googlebot I consent.

PR #4282, opened 2019-10-29
[address-resolver] update EID-to-RLOC cache when attaching MTD child

librasungirl:
This commits removes deprecated EID-to-RLOC cache if there is entry for
the unicast address registered by MTD child in case this child switches
its parent.

### DIFF
--- a/src/core/thread/address_resolver.hpp
+++ b/src/core/thread/address_resolver.hpp
@@ -115,6 +115,17 @@ public:
     otError UpdateCacheEntry(const Ip6::Address &aEid, Mac::ShortAddress aRloc16);
 
     /**
+     * This method removes the cache entry for the EID.
+     *
+     * @param[in]  aEid               A reference to the EID.
+     *
+     * @retval OT_ERROR_NONE          Successfully removed the cache entry.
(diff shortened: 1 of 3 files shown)


src/core/thread/address_resolver.hpp
```diff
@@ -114,6 +114,17 @@ class AddressResolver : public InstanceLocator
...
+    /**
+     * This method removes the cache entry for the EID.
+     *
+     * @param[in]  aEid               A reference to the EID.
+     *
+     * @retval OT_ERROR_NONE          Successfully removed the cache entry.
+     * @retval OT_ERROR_NOT_FOUND     No cache entry with @p aEid.
+     *
+     */
+    otError RemoveCacheEntry(const Ip6::Address &aEid);
```
jwhui: nit: how about renaming to `Remove()` and co-locating with the other `Remove()` methods above?

librasungirl: Thanks. Updated.

librasungirl:
@abtink Yes, please. Thanks for proposing to help to add test :)

PR #4285, opened 2019-10-29
[docs] add note that use of the public Heap API is discouraged

Vyrastas:
For visibility on https://openthread.io/reference/group/api-heap during future Doxygen runs, based on comments in #4143.

### DIFF
--- a/include/openthread/heap.h
+++ b/include/openthread/heap.h
@@ -47,6 +47,8 @@ extern "C" {
  * @brief
  *   This module includes functions that set the external OpenThread heap.
  *
+ * @note This API is deprecated and use of it is discouraged.
+ *
  * @{
  *
  */
--- a/include/openthread/heap.h
+++ b/include/openthread/heap.h
@@ -47,8 +47,6 @@ extern "C" {
  * @brief
  *   This module includes functions that set the external OpenThread heap.
  *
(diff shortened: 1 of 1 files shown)


include/openthread/heap.h
```diff
@@ -47,6 +47,8 @@ extern "C" {
  * @brief
  *   This module includes functions that set the external OpenThread heap.
  *
+ * @note This API is deprecated and use of it is discouraged.
```
jwhui: I believe `otHeapSetCAllocFree()` is OK to use, as it allows the platform to point OpenThread to an external heap. However, `otHeapCAlloc()` and `otHeapFree()` should **not** be used, as they allow external code to use OpenThread's internal heap.

Vyrastas: Ah ok, if we only want to mark those two as discouraged, I will move the note to both of those functions.  Currently they show up with no documentation on openthread.io.

jwhui: I think that makes sense.

Vyrastas:
Updated to place the note under the functions instead.

Also checking in some files related to #4250 per a conversation with @jwhui .  They appear that way in a fresh clone of OT.

jwhui:
@Vyrastas, I just merged #4292 to help address this Travis failure. Can you rebase to see if #4292 resolves the issue?

PR #4294, opened 2019-10-31
[ncp] add support in spinel/NCP to generate PSKc from pass-phrase 

abtink:
This PR also contains a smaller change in APIs:

[commissioner] change GeneratePskc to use otPskc/Pskc types (#4294)

----
Related PR in wpantund: https://github.com/openthread/wpantund/pull/443

### DIFF
--- a/src/core/api/commissioner_api.cpp
+++ b/src/core/api/commissioner_api.cpp
@@ -181,15 +181,12 @@ otCommissionerState otCommissionerGetState(otInstance *aInstance)
     return instance.Get<MeshCoP::Commissioner>().GetState();
 }
 
-otError otCommissionerGeneratePskc(otInstance *           aInstance,
-                                   const char *           aPassPhrase,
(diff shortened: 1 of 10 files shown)


src/core/api/commissioner_api.cpp
```diff
@@ -185,11 +185,11 @@ otError otCommissionerGeneratePskc(otInstance *           aInstance,
                                    const char *           aPassPhrase,
                                    const char *           aNetworkName,
                                    const otExtendedPanId *aExtPanId,
-                                   uint8_t *              aPskc)
+                                   otPskc *               aPskc)
 {
     OT_UNUSED_VARIABLE(aInstance);
```
jwhui: Should we remove `aInstance` from the argument list?

abtink: Good idea! Since we are changing the API anyway, good time to remove `aInstance` as well!

abtink: Changed it in new pushed commits.

PR #4300, opened 2019-11-03
[build] detect undefined macros to avoid misconfigurations

dismirlian:
Hi, this PR adds the -Wundef option to detect misconfigurations. Also, it fixes the following bugs:

- `-Wno-gnu-zero-variadic-macro-arguments` applied only to clang
- `kProtocolVersion` is undefined at pre-processing time; here: https://github.com/openthread/openthread/blob/16c01d58c04a8a572be40bcb35132a8236ec3e6d/src/core/mac/mac_frame.hpp#L1282
I actually don't know if the fix is ok, so I'd appreciate feedback.
- `openthread-core-default-config.h` requires definition of `OPENTHREAD_CONFIG_COAP_SECURE_API_ENABLE`, which is defined by default in `src/core/config/coap.h`:
https://github.com/openthread/openthread/blob/16c01d58c04a8a572be40bcb35132a8236ec3e6d/src/core/config/coap.h#L108
https://github.com/openthread/openthread/blob/16c01d58c04a8a572be40bcb35132a8236ec3e6d/src/core/config/openthread-core-default-config.h#L270
An ugly temporary fix is to include `coap.h` before `openthread-core-default-config.h` in `src/core/openthread-core-config.h`.
- Similar problem with `OPENTHREAD_CONFIG_TIMESYNC_ENABLE`. `config/time_sync.h` is placed before `config/mac.h`. An additional problem is that `make pretty` messes with the header order.
- Typos in `src/ncp/ncp_uart.hpp` (`OPENTHREAD_CONFIG_NCP_SPI_ENABLENEL_ENCRYPTER` instead of `OPENTHREAD_ENABLE_NCP_SPINEL_ENCRYPTER`).

Some platform code will probably report many warnings; I've removed -Wundef for `efr32`.


src/core/openthread-core-config.h
```diff
@@ -44,14 +44,16 @@
 
 #define OPENTHREAD_THREAD_VERSION (2)
 
+// This should be before openthread-core-default-config.h
+#include "config/coap.h"
```
abtink: One suggestion here: How about we include the header file from the file that needs/uses the other definitions
For example, `openthread-core-default-config.h` wants to use  `OPENTHREAD_CONFIG_COAP_SECURE_API_ENABLE`, so we add `#include "config/coap.h"` at the top of it.

Same way for `config/time_sync` and `config/mac`, i.e., if `config/mac.h` needs/uses `OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT`, we can add `#include "config/time_sync.h"` in `config/mac.h`.

The advantage I see with this approach is that the dependencies are handled by the header files themselves (instead of order of includes in this file), and if we ever happen  to have a loop of header file dependencies we would notice it.

Thoughts?

src/core/config/mac.h
```diff
@@ -162,7 +162,7 @@
  *
  */
 #ifndef OPENTHREAD_CONFIG_MAC_JOIN_BEACON_VERSION
-#define OPENTHREAD_CONFIG_MAC_JOIN_BEACON_VERSION kProtocolVersion
+#define OPENTHREAD_CONFIG_MAC_JOIN_BEACON_VERSION OPENTHREAD_THREAD_VERSION
```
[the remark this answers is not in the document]
dismirlian: Hi, @abtink, thanks for the suggestion. I don't fully understand the purpose of `kProtocolVersion`, but I think defining this as `2` makes sense and is consistent. I don't know how to handle the two occurrences in mac_frame.hpp, here:

https://github.com/dismirlian/openthread/blob/7e76e6f7c74825aac0bf12abc56cdfbdaf68f9ac/src/core/mac/mac_frame.hpp#L1190

https://github.com/dismirlian/openthread/blob/7e76e6f7c74825aac0bf12abc56cdfbdaf68f9ac/src/core/mac/mac_frame.hpp#L1282

src/core/openthread-core-config.h
dismirlian: Great, I agree.

src/core/config/mac.h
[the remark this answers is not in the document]
abtink: @dismirlian , I suggest keeping the use in `mac_frame.hpp` as before.

(if I recall correctly, this config option was added to allow beacon version to set to a different value when sending 'joinable' beacons - I expect typical use-cases would use the default value - this config I think was required to ensure OT emitted beacons were compatible with a legacy network solution).

dismirlian:
I've updated all platforms to either:
- Correct problems with undefined preprocessor symbols
or
- Silence warning (-Wundef), in case these come from code in SDKs (`third_party`).

Travis previously failed because the logs became too long (too many warnings). All platforms should now compile with no warnings.

src/core/config/mac.h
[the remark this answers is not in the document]
dismirlian: @abtink please check if `mac_frame.hpp` looks good to you. I'm not 100% sure. Thanks!

third_party/nlbuild-autotools/repo/autoconf/m4/ax_check_compiler.m4
```diff
@@ -31,8 +31,16 @@ AC_DEFUN([_AX_CHECK_COMPILER_OPTION_WITH_VAR],
     AC_MSG_CHECKING([whether the _AC_LANG compiler understands $3])
     SAVE_[]_AC_LANG_PREFIX[]FLAGS=${_AC_LANG_PREFIX[]FLAGS}
     SAVE_$2=${$2}
-    _AC_LANG_PREFIX[]FLAGS=$3
+    AS_LITERAL_IF([$3],
```
jwhui: The upstream for `third_party/nlbuild-autotools` is https://github.com/nestlabs/nlbuild-autotools. Could you try submitting these changes to upstream first?

dismirlian: Sure, I'll revert this commit and create a pull request to  https://github.com/nestlabs/nlbuild-autotools first.

I added these changes because without them gcc4 is fed the -Wno-c++14-compat flag, which it doesn't recognize, so CI fails when there are undefined macros (due to -Werror). gcc > 4 accepts this switch.

jwhui:
@dismirlian, any update on moving this PR forward?

dismirlian:
Hi @jwhui, the folks at https://github.com/nestlabs/nlbuild-autotools prefer not to include the fix. They argue that configure.ac should detect the compiler and check the positive version of each -Wno-* switch before adding the negative version to CFLAGS, instead of handling this in autotools.

I haven't had time yet to review how I can implement this. I'll look at this in a few days.

configure.ac
```diff
@@ -250,17 +255,60 @@ AC_CACHE_CHECK([whether $CC is Clang],
...
+        AX_CHECK_COMPILER_OPTIONS([C++], ${flags_to_check})
+        restore_negative_form_of_warning_flags "${CXXFLAGS}"
+        CXXFLAGS=${final_flags}
+        unset final_flags
+    fi
+}
 
-LT_INIT
-AC_PROG_LIBTOOL
+check_prospective_CFLAGS
```
jwhui: @gerickson, please review relative to https://github.com/nestlabs/nlbuild-autotools/pull/28#discussion_r349834306

gerickson: I'd make the same recommendation I made there which is that this is all assuming GCC or some GCC-compatible compiler.

I'd short circuit all of that logic there is the compiler isn't GCC or a GCC-compatible compiler. Given that this is in openthread's configure.ac, my feedback is advisory since this is project-specific.

dismirlian: I am assuming that OpenThread will be compiled by either gcc or clang. If this is not the case, I can include a test to short circuit the logic.

@jwhui, @gerickson, what do you think?

jwhui: I'm ok with this for now.

PR #4305, opened 2019-11-05
[spi-hdlc-adapter] add asserting C̅S̅ support

bukepo:
This PR adds a new argument to spi-hdlc-adapter to support asserting the
C̅S̅ pin before SPI transaction.

### DIFF
--- a/tools/spi-hdlc-adapter/README.md
+++ b/tools/spi-hdlc-adapter/README.md
@@ -35,6 +35,9 @@ protocol document.
     value is `0`.
 *   `--spi-speed[=hertz]`: Specify the SPI speed in hertz. Default
     value is `1000000` (1MHz).
+*   `--spi-cs[=gpio-path]`: Specify the a path to the Linux
+    sysfs-exported GPIO directory for the `C̅S̅` pin. If not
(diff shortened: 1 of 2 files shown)


abtink:
The code itself in this PR looks good.

I am not super familiar with the SPI related (linux) APIs but I think the CS may be already controlled by `spidev.h` itself through the `ioctl` calls from `do_spi_xfer()`. I know in some (internal) projects we do indeed rely/use the CS pin (on NCP side) and it's being controlled. Also we have `cs-delay` parameter and seems like we pass it as part of `spi_ioc_transfer` struct. 

@bukepo what's the use-case for adding control of CS by `spi-hdlc-adpater`? I think (if we can) we should delegate the control of CS pin to driver. If we want to control it from `spi-hdlc-adpater`, then we may need to handle cases where SPI is in use (i.e., what if the SPI is in use for another peripheral when we assert the CS pin, do we need to check for that?).

bukepo:
We'll hold off this PR until we know better the need for this.

jwhui:
Closing stale PR. Feel free to reopen if/when ready to move this forward.

PR #4307, opened 2019-11-05
[doc] clarify mac counters meaning

LuDuda:
This PR clarifies the meaning of MAC counters. This is driven by customer questions.

Also in the future, we should address the problem with the inconsistency for the `mTxErrCca` and `mTxRetry` counters. The meaning of the counters depends on the platform's radio capabilities.

To solve this problem, I believe we should add more fields to `otRadioFrame::mRxInfo` structure, so that whenever OT delegates e.g. CSMA/CA or Retransmission fnction to platform layer, there is a way to communicate e.g. a number of CCA failures or number of retransmissions to the higher layer to collect correct data.

### DIFF
--- a/include/openthread/link.h
+++ b/include/openthread/link.h
@@ -108,7 +108,7 @@ typedef struct otMacFilterEntry
 typedef struct otMacCounters
 {
     /**
-     * The total amount of unique MAC frame transmission requests.
+     * The total number of unique MAC frame transmission requests.
      *
      * Note that this counter is incremented for each MAC transmission request only by one,
      * regardless of the amount of CCA failures, CSMA-CA attempts, or retransmissions.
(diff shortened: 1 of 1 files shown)


include/openthread/link.h
```diff
@@ -107,38 +107,252 @@ typedef struct otMacFilterEntry
...
+     * The following equations are valid:
+     *     @p mTxTotal = @p mTxUnicast + @p mTxBroadcast
+     *     @p mTxTotal = @p mTxAckRequested + @p mTxNoAckRequested
+     *     @p mTxTotal = @p mTxData + @p mTxDataPoll + @p mTxBeacon + @p mTxBeaconRequest + @p mTxOther
+     *
+     */
+    uint32_t mTxTotal;
+
+    /**
+     * The total amount of unique unicast MAC frame transmission requests.
```
abtink: nit: suggest using `total number of` instead of `total amount of`.

PR #4312, opened 2019-11-08
[network-data] fix prefix comparison / use reference param instead of pointer (during iter)

abtink:
This PR contains two related commits in `NetworkData`: 

**[network-data] change iterator methods to use reference instead of pointer**
    
This commit contains t changes the methods in `NetworkData` class used for iterating over on-mesh prefix, external router, etc (e.g.,`GetNextOnMeshPrefix`()) to use reference input variables instead of
pointer type variables.

-----------------

**[network-data] fix how the OnMeshPrefixConfig structs are compared**
    
This commit ensures the entire struct size is used when comparing two `OnMeshPrefixConfig` (`otBorderRouterConfig`) instances from `NetworkData::ContainsOnMeshPrefixes()` method.

### DIFF
--- a/src/core/thread/network_data.cpp
+++ b/src/core/thread/network_data.cpp
@@ -80,14 +80,12 @@ exit:
     return error;
 }
 
-otError NetworkData::GetNextOnMeshPrefix(otNetworkDataIterator *aIterator, otBorderRouterConfig *aConfig)
+otError NetworkData::GetNextOnMeshPrefix(Iterator &aIterator, OnMeshPrefixConfig &aConfig)
 {
     return GetNextOnMeshPrefix(aIterator, Mac::kShortAddrBroadcast, aConfig);
 }
 
-otError NetworkData::GetNextOnMeshPrefix(otNetworkDataIterator *aIterator,
-                                         uint16_t               aRloc16,
-                                         otBorderRouterConfig * aConfig)
+otError NetworkData::GetNextOnMeshPrefix(Iterator &aIterator, uint16_t aRloc16, OnMeshPrefixConfig &aConfig)
 {
     otError             error = OT_ERROR_NOT_FOUND;
     NetworkDataIterator iterator(aIterator);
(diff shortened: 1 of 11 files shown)


abtink:
#### More details on second commit: 
The existing code uses the following: 
```
if (memcmp(&outerConfig, &innerConfig, (sizeof(outerConfig) - sizeof(outerConfig.mRloc16))) == 0))
```
I think the intention here was to compare all fields in the `otBorderRouterConfig` struct but exclude the last field which is `mRloc16`. 

Since the `otBorderRouterConfig` is not defined as a packed struct, compiler can choose to add padding to it and the `memcmp` is not necessarily correct in excluding `mRloc16`. I actually verified that on my local machine with `gcc 8.3.0 (Debian 8.3.0-6)` we see:
```
offsetof(otBorderRouterConfig, mRloc16) = 20
sizeof(otBorderRouterConfig) = 24
sizeof(config.mRloc16) = 2
```

Looking at the code for `ContainsOnMeshPrefixes()`: 
1. If we use `ContainsOnMeshPrefixes()` with a specific `aRloc16` then, all prefixes are restricted to the specific RLOC anyways, and including `mRloc16` in `memcmp` should be fine.
1. In the case where pass `aRloc16` as `kShortAddrBroadcast` I still think we should include the `mRloc16` as part of the check. @jwhui  what do you think?

In our current uses of `ContainsOnMeshPrefixes()` (which is from `network_data_local.cpp`) we always use it with a specific RLOC (our RLOC). So case 2 above is not currently used. 
https://github.com/openthread/openthread/blob/c8c26b7ac03d99481ebcc183319f8e6ea80d6fbb/src/core/thread/network_data_local.cpp#L329-L334

src/core/thread/network_data.cpp
```diff
@@ -378,19 +374,19 @@ otError NetworkData::GetNextServiceId(otNetworkDataIterator *aIterator, uint16_t
...
-        otBorderRouterConfig  innerConfig;
-        otError               error;
+        Iterator           innerIterator = kIteratorInit;
+        OnMeshPrefixConfig innerConfig;
+        otError            error;
 
-        while ((error = GetNextOnMeshPrefix(&innerIterator, aRloc16, &innerConfig)) == OT_ERROR_NONE)
+        while ((error = GetNextOnMeshPrefix(innerIterator, aRloc16, innerConfig)) == OT_ERROR_NONE)
         {
-            if (memcmp(&outerConfig, &innerConfig, (sizeof(outerConfig) - sizeof(outerConfig.mRloc16))) == 0)
```
jwhui: @librasungirl, can you help identify the reason for `- sizeof(outerConfig.mRloc16)`?

librasungirl: It really takes some time to recall the code several years ago. 

Now I may have thought of the intention at that time (when there might be some assumption that only one device serves as BR). However as @abtink noted, it may not take effect as the `otBorderRouterConfig` is not defined as a packed struct.

I am fine with the update here. 

Speaking of the original intention, the specification should have no rules to disallow the device which requests stable Thread Network Data to become a server, then we may still need to resolve the original concern (may in a separate PR) -- unexpected SRV_DATA.ntf from the device which acts as Server but require only stable network data.
For example,
one MTD (mode `rs`) serves as BR (`2001::/64 paosr`), due to the rules in 1.1 spec section 5.14,

> P_border_router_16 values for Border Routers that do not serve as DHCPv6 Agents are not used and set to the invalid value of 0xFFFE

This MTD server would always think its network data inconsistent and send SRV_DATA.ntf whenever there is `OT_CHANGED_THREAD_NETDATA` event in existing implementation.
https://github.com/openthread/openthread/blob/master/src/core/thread/mle.cpp#L1549

We may need to consider Server ALOCs/Border Router ALOCs etc for such device as well when determining the consistence.
And we may also need to move `Get<NetworkData::Local>().SendServerDataNotification()` when HandleAdvertisement from mle_router.cpp to mle.cpp 

Thoughts? @jwhui @abtink

jwhui: @librasungirl, I had guessed the intention had something to do with requesting only stable network data. In that case, I'm not sure how a device can determine if its information has been published to the leader. It can only determine if the prefixes exist, but not whether those prefixes were contributed by itself. That said, just determining existence may be useful in itself. Let's continue to discuss this issue separately.

PR #4318, opened 2019-11-12
[radio] process tx AES without otInstance

bukepo:
`otPlatRadioFrameUpdated()` is declared not to access any state within
OpenThread. However the current implementation does read the extended
address. This PR moves AES process into `Mac::TxFrame`, which is free of
`otInstance`, which also eliminates the current deep callback stacks
when the radio driver wants to process tx AES from interrupt context.

### DIFF
--- a/examples/platforms/utils/mac_frame.h
+++ b/examples/platforms/utils/mac_frame.h
@@ -143,7 +143,7 @@ uint8_t otMacFrameGetSequence(const otRadioFrame *aFrame);
 /**
  * This function performs AES CCM on the frame which is going to be sent.
  *
- * @param[in]  aFrame       A reference to the MAC frame buffer that is going to be sent.
+ * @param[in]  aFrame       A pointer to the MAC frame buffer that is going to be sent.
  * @param[in]  aExtAddress  A pointer to the extended address, which will be used to generate nonce
  *                          for AES CCM computation.
  *
--- a/examples/platforms/utils/mac_frame.h
+++ b/examples/platforms/utils/mac_frame.h
@@ -140,6 +140,16 @@ otError otMacFrameGetSrcAddr(const otRadioFrame *aFrame, otMacAddress *aMacAddre
  */
 uint8_t otMacFrameGetSequence(const otRadioFrame *aFrame);
 
+/**
+ * This function performs AES CCM on the frame which is going to be sent.
+ *
+ * @param[in]  aFrame       A reference to the MAC frame buffer that is going to be sent.
+ * @param[in]  aExtAddress  A pointer to the extended address, which will be used to generate nonce
+ *                          for AES CCM computation.
+ *
+ */
+void otMacFrameProcessTransmitAesCcm(otRadioFrame *aFrame, const otExtAddress *aExtAddress);
+
 #ifdef __cplusplus
 } // extern "C"
 #endif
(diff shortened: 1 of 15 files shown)


examples/platforms/utils/mac_frame.h
```diff
@@ -140,6 +140,16 @@ otError otMacFrameGetSrcAddr(const otRadioFrame *aFrame, otMacAddress *aMacAddre
  */
 uint8_t otMacFrameGetSequence(const otRadioFrame *aFrame);
 
+/**
+ * This function performs AES CCM on the frame which is going to be sent.
+ *
+ * @param[in]  aFrame       A reference to the MAC frame buffer that is going to be sent.
```
Irving-cl: A pointer here?

src/core/mac/mac_frame.hpp
```diff
@@ -1095,6 +1095,14 @@ class TxFrame : public Frame
      */
     void CopyFrom(const TxFrame &aFromFrame);
 
+    /**
+     * This method performs AES CCM on the frame which is going to be sent.
+     *
+     * @param[in]  aExtAddress  A pointer to the extended address, which will be used to generate nonce
```
Irving-cl: The parameter is a reference here.

src/core/mac/mac.cpp
```diff
@@ -966,7 +941,7 @@ void Mac::ProcessTransmitSecurity(TxFrame &aFrame, bool aProcessAesCcm)
 
     if (aProcessAesCcm)
     {
-        ProcessTransmitAesCcm(aFrame, extAddress);
+        aFrame.ProcessTransmitAesCcm(*extAddress);
```
Irving-cl: It seems to be possible that `extAddress` being `NULL` here, for example, in case `kKeyIdMode2`. Do we need to add a check here?

bukepo: Is it `null` in `kKeyIdMode2`?

src/core/mac/mac_frame.cpp
```diff
@@ -998,6 +1000,40 @@ void TxFrame::CopyFrom(const TxFrame &aFromFrame)
 #endif
 }
 
+void TxFrame::ProcessTransmitAesCcm(const ExtAddress &aExtAddress)
+{
+#if OPENTHREAD_RADIO
```
abtink: :+1:  for adding this method in `TxFrame` (moving it from `Mac`). I think this makes great sense.

One suggestion here: How about add `#ifdef OPENTHREAD_MTD || OPENTHREAD_FTD` to the whole definition/implementation of this method (i.e., not have it even defined under `OPENTHREAD_RADIO`).  This way, if it happens to be used under `OPENTHREAD_RADIO` (by mistake) we get a linker error.

bukepo: Thanks for the quick review. We cannot do that easily because the platform library is build without knowing the upper layer is rcp or not. All ftd, mtd, rcp link to the same platform library. We can introduce another special platform library for RCP, but i tend it to be done in a separate PR. Thoughts?

abtink: How about keeping the declaration but not having any implementation of this method under `OPENTHREAD_RADIO`? The platform lib should be happy seeing a declartion of this method and we still would get a linker error if used?

bukepo: I'm afraid not. ot-rcp is build together with ot-ncp and ot-cli. It must link all symbols.

abtink: Ok. sounds good.

PR #4322, opened 2019-11-12
[build] add a gn based build and corresponding travis script

paragdixit-g:
The travis script ensures that build of an openthread-core build with gn
and ninja goes through. The run time is about 1 min.

Additional reviewers : @vaas-krish @abtink

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,9 @@ jobs:
       dist: trusty
       python: "2.7" # The old Android build system only supports python2
       script: .travis/script.sh
+    - env: BUILD_TARGET="gn-build" VERBOSE=1
+      os: linux
+      script: .travis/script.sh
     - env: BUILD_TARGET="posix-distcheck" VERBOSE=1 VIRTUAL_TIME=1
       os: linux
       compiler: clang
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1,0 +1,221 @@
+#  Copyright (c) 2019, The OpenThread Authors.
+#  All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions are met:
+#  1. Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#  2. Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in the
+#     documentation and/or other materials provided with the distribution.
+#  3. Neither the name of the copyright holder nor the
+#     names of its contributors may be used to endorse or promote products
+#     derived from this software without specific prior written permission.
+#
+#  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+#  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+#  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+#  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+#  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+#  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+#  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
+static_library("lib-ot-core") {
+
+ cflags_cc = [ "-Wno-non-virtual-dtor" ]
+
+ sources = [
+    "src/core/api/border_agent_api.cpp",
+    "src/core/api/border_router_api.cpp",
+    "src/core/api/channel_manager_api.cpp",
+    "src/core/api/channel_monitor_api.cpp",
+    "src/core/api/child_supervision_api.cpp",
+    "src/core/api/coap_api.cpp",
+    "src/core/api/coap_secure_api.cpp",
+    "src/core/api/commissioner_api.cpp",
+    "src/core/api/crypto_api.cpp",
+    "src/core/api/dataset_api.cpp",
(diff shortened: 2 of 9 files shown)


BUILD.gn
```diff
@@ -0,0 +1,191 @@
+# Copyright 2019 Google Inc. All rights reserved.
```
jwhui: Can we use the standard OpenThread license header?

paragdixit-g: Sure, will fix this.

BUILD.gn
```diff
@@ -0,0 +1,204 @@
...
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
+static_library("lib-ot-core") {
+
+ cflags_cc = [ "-Wno-non-virtual-dtor" ]
+
+ sources = [
```
abtink: A question/suggestion: Is there any solution to have this list be auto generated (more generic), e.g., include all `cpp` files in `src/core/api`, and `src/core/common` and ...)?
If we can do this,  I think it would be more generic and robust against file renames and new src files added later.

Thoughts?

paragdixit-g: Currently we are matching how things are done in Android.mk file. 

It is possible to provide wildcards to include all cpp files within a given directory.  But not sure whether we want to do that, as it may produce unintentional side effects for Fuchsia specific directory.  For example, we have explicitly removed posix specific files which were there in Android.mk.  

This is just a starting point, and the file is expected to evolve anyway when new files/stubs are added for Fuchsia. We can make the change now (to include all cpp files within a directory now), or defer it to wait for things to progress a little bit.

paragdixit-g: Currently we are matching how things are done in Android.mk file.

It is possible to provide wildcards to include all cpp files within a given directory. But not sure whether we want to do that, as it may produce unintentional side effects for Fuchsia specific directory. For example, we have explicitly removed posix specific files which were there in Android.mk.

This is just a starting point, and the file is expected to evolve anyway when new files/stubs are added. We can make the change now (to include all cpp files within a directory now), or defer it to wait for things to progress a little bit.

paragdixit-g: Currently we are matching how things are done in Android.mk file.
Also, we are excluding some files within src/core/api , like : border_agent_api.cpp, similar to Android.mk

It is possible to provide wildcards to include all cpp files within a given directory. But not sure whether we want to do that right now.

This is just a starting point, and the file is expected to evolve anyway when new files/stubs are added. We're just matching Android.mk as a starting point and make the choice consciously of what files need to be included as required later.

BUILD.gn
```diff
@@ -0,0 +1,204 @@
...
+    "src/core/utils/missing_strlcpy.c",
+    "src/core/utils/missing_strlcat.c",
+    "src/core/utils/missing_strnlen.c",
+    "src/core/utils/parse_cmdline.cpp",
+    "src/core/utils/slaac_address.cpp",
+    "src/ncp/hdlc.cpp",
+    "src/ncp/spinel.c",
+    "src/ncp/spinel_decoder.cpp",
+    "src/ncp/spinel_encoder.cpp",
+    "third_party/mbedtls/repo/library/md.c",
```
gjc13: Shall we move the mbedtls sources to a separate BUILD.gn in third_party/mbedtls/?

BUILD.gn
abtink: I would still suggest that if we can use wildcards for certain folders, it would be great to include it from beginning and test it.

We can use the wildcards for certain core folders (the ones we know we always want to include all their src files). And for some other folders use a specific list. 

In general, regarding the OT core files/folder, we would want to include every file in a folder (there are `#if` checks at the top of files if certain features are not enabled), e.g., I see no advantage to not include the `border_agent_api.cpp`.

The main reason for this is to make it easier when new files are added/removed in code or there is a file rename..We want to reduce the number of makefiles/build files that developers need to maintain/update (makes it easier on devs and easier to test changes on travis). This also helps avoid the situation where unused/disabled features would go unnoticed if they are not included in the build/makefiles (we have faced this with andriod.mk in past where enabling feature later, caused build errors that certain src files were missing since they were not updated in andriod.mk during a change).

jwhui: It should be OK to always include all source files in `src/core`. Everything should be appropriately feature flagged. We intentionally support this to simplify maintenance of other build systems.

paragdixit-g: Sorry, I have a correction.   There doesn't seem to be a way to provide source names to include with wild cards. 

When I looked up help earlier (https://gn.googlesource.com/gn/+/master/docs/reference.md#how-to-use-patterns) , I thought patterns are used to include the source files.  However the example in the help shows how to filter out files matching a pattern (ie just provides an exclude list). I have also cross checked that Fuchsia tree which heavily uses GN lists all the source files in the BUILD.gn hierarchy. 

So we're probably stuck with listing out all the files explicitly.

paragdixit-g: I'm still checking to see if there is some trick to make it work, but surely there is no straight forward way.

BUILD.gn
paragdixit-g: It should be possible to do so. What is the motivation for that ? -   Is it for faster incremental build or something else ?

From faster speed perspective, we're already pretty fast given that it's a small project with limited number of files to build. I tried with a single and split BUILD.gn file that in both cases the runtime is about 5ms for gn step (which is what is speeded up because of parallel processing of BUILD.gn files) .  

I have also checked with tools team for Fuchsia project which uses GN about using a single vs hierarchical structure of BUILD.gn files : "In terms of performance, having separate BUILD.gn files for each directory is preferred because GN processes BULD.gn files in parallel, so by using a single file, you're reducing parallelism. On the other hand, if your project isn't too large and the current performance is fine for you, using a single BUILD.gn at this stage should be fine."

Using multiple files introduce some amount of complexity in BUILD.gn files as we have to properly propagate relevant header/configs. So I would like to avoid that if speed is only motivation for splitting.

BUILD.gn
paragdixit-g: After searching more, I don't think there is a way to provide wildcards for adding files.
I have added all source files under src/core to the BUILD.gn file as suggested above.

paragdixit-g: Based on above, I'm marking the conversation as resolved, please reopen if there is anything additional required.

jwhui:
@paragdixit-g, can we also add a `third_party/build_gn/README.md` file (see [third_party/mbedtls/README.md](https://github.com/openthread/openthread/blob/master/third_party/mbedtls/README.md) for an example).

.travis.yml
```diff
@@ -66,6 +66,10 @@ jobs:
       dist: trusty
       python: "2.7" # The old Android build system only supports python2
       script: .travis/script.sh
+    - env: BUILD_TARGET="gn-build" VERBOSE=1
+      os: linux
+      dist: trusty
```
jwhui: nit: is trusty required? if not, let's leave this out and use the global default (currently bionic).

paragdixit-g: Done

jwhui: I still see trusty specified here. Can you make sure you pushed the change?

paragdixit-g: sorry about that, I pushed the correct file today morning.

PR #4323, opened 2019-11-12
[travis] add arm gcc 9 to build tests

jwhui:
Also addresses potential unaligned pointer value flagged by Arm GCC 9 in cc1352, cc2650, and cc2652 example drivers.

### DIFF
--- a/examples/platforms/cc1352/radio.c
+++ b/examples/platforms/cc1352/radio.c
@@ -1575,14 +1575,15 @@ otError otPlatRadioAddSrcMatchExtEntry(otInstance *aInstance, const otExtAddress
 {
     OT_UNUSED_VARIABLE(aInstance);
 
-    otError error = OT_ERROR_NONE;
-    uint8_t idx   = rfCoreFindExtSrcMatchIdx((uint64_t *)aExtAddress);
+    otError  error      = OT_ERROR_NONE;
+    uint64_t extAddress = *(uint64_t *)aExtAddress;
+    uint8_t  idx        = rfCoreFindExtSrcMatchIdx(&extAddress);
 
     if (idx == CC1352_SRC_MATCH_NONE)
     {
         /* the entry does not exist already, add it */
         otEXPECT_ACTION((idx = rfCoreFindEmptyExtSrcMatchIdx()) != CC1352_SRC_MATCH_NONE, error = OT_ERROR_NO_BUFS);
-        sSrcMatchExtData.extAddrEnt[idx] = *((uint64_t *)aExtAddress);
+        sSrcMatchExtData.extAddrEnt[idx] = extAddress;
     }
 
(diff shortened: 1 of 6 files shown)


srickardti:
@jwhui I'm not really clear on why this is necessary. The pointer in question was used only to cast between the otExtAddress type and a uint64_t. The M4F or M3 can handle unaligned accesses, and the M0 radio controller is protected by the alignment of the structs it is sent.

examples/platforms/cc1352/radio.c
```diff
@@ -1575,14 +1575,18 @@ otError otPlatRadioAddSrcMatchExtEntry(otInstance *aInstance, const otExtAddress
 {
     OT_UNUSED_VARIABLE(aInstance);
 
-    otError error = OT_ERROR_NONE;
-    uint8_t idx   = rfCoreFindExtSrcMatchIdx((uint64_t *)aExtAddress);
+    otError  error = OT_ERROR_NONE;
+    uint8_t  idx;
+    uint64_t extAddress;
+
+    memcpy(&extAddress, aExtAddress, sizeof(extAddress));
```
srickardti: This could be done with a cast and assignment to a uint64_t. I think GCC is smart enough to optimize out the function call, so that is probably fine.

jwhui:
@srickardti, for full context, here is the compiler warning:
```
examples/platforms/cc1352/radio.c: In function 'otPlatRadioAddSrcMatchExtEntry':
examples/platforms/cc1352/radio.c:1579:5: error: converting a packed 'otExtAddress' {aka 'const struct otExtAddress'} pointer (alignment 1) to a 'uint64_t' {aka 'const long long unsigned int'} pointer (alignment 8) may result in an unaligned pointer value [-Werror=address-of-packed-member]
 1579 |     uint8_t idx   = rfCoreFindExtSrcMatchIdx((uint64_t *)aExtAddress);
      |     ^~~~~~~
examples/platforms/cc1352/radio.c: In function 'otPlatRadioClearSrcMatchExtEntry':
examples/platforms/cc1352/radio.c:1614:5: error: converting a packed 'otExtAddress' {aka 'const struct otExtAddress'} pointer (alignment 1) to a 'uint64_t' {aka 'const long long unsigned int'} pointer (alignment 8) may result in an unaligned pointer value [-Werror=address-of-packed-member]
 1614 |     otEXPECT_ACTION((idx = rfCoreFindExtSrcMatchIdx((uint64_t *)aExtAddress)) != CC1352_SRC_MATCH_NONE,
      |     ^~~~~~~~~~~~~~~
examples/platforms/cc1352/radio.c: At top level:
cc1: error: unrecognized command line option '-Wno-gnu-zero-variadic-macro-arguments' [-Werror]
```

examples/platforms/cc1352/radio.c
jwhui: Yes, cast-and-assignment seems to work here as well.

PR #4325, opened 2019-11-13
[mle] add enhanced keep-alive

bukepo:
This PR implements the enhanced keep-alive feature of Thread 1.2.
* Added version field into settings.


src/core/thread/mesh_forwarder.cpp
```diff
@@ -1096,6 +1096,17 @@ void MeshForwarder::HandleReceivedFrame(Mac::RxFrame &aFrame)
...
 
+#if OPENTHREAD_THREAD_VERSION >= OPENTHREAD_THREAD_VERSION_1_2 && OPENTHREAD_FTD
+        // From Thread 1.2, MAC Data Frame can also act as keep-alive message if child supports
+        {
+            Child *child = Get<ChildTable>().FindChild(macSource, ChildTable::kInStateAnyExceptInvalid);
+
+            if (child != NULL && child->IsEnhancedKeepAliveSupported())
+            {
+                child->SetLastHeard(TimerMilli::GetNow());
+            }
```
CiaranWoodward: The MAC Data Frame should only be used as a keep alive if the child is rx-off-when-idle, rx-on children are required to use the Child Update Request keep-alive.

src/core/thread/mesh_forwarder.cpp
```diff
@@ -1096,6 +1096,17 @@ void MeshForwarder::HandleReceivedFrame(Mac::RxFrame &aFrame)
             LogFrame("Received empty payload frame", aFrame, OT_ERROR_NONE);
         }
 
+#if OPENTHREAD_THREAD_VERSION >= OPENTHREAD_THREAD_VERSION_1_2 && OPENTHREAD_FTD
+        // From Thread 1.2, MAC Data Frame can also act as keep-alive message if child supports
+        {
```
abtink: Do we need to check if the received frame is secure (has link security)?

src/core/thread/mesh_forwarder.cpp
abtink: I am curios about why spec kept non-sleepy same as before.

If we receive a secure data frame from a child (sleepy or not), would it not indicate that child is still connected and ok (which I think was intention behind the keep-alive msg)?

src/core/thread/mle.cpp
```diff
@@ -3588,6 +3604,11 @@ otError Mle::HandleChildUpdateResponse(const Message &         aMessage,
     switch (mRole)
     {
     case OT_DEVICE_ROLE_DETACHED:
+        // Version
+        SuccessOrExit(error = Tlv::GetTlv(aMessage, Tlv::kVersion, sizeof(version), version));
+        VerifyOrExit(version.IsValid(), error = OT_ERROR_PARSE);
+        mParent.SetVersion(static_cast<uint8_t>(version.GetVersion()));
```
abtink: I am not sure but I think this may break compatibility in behavior between different OT builds (before and after this change)
- Here we expect in `HandleChildUpdateResponse()` that the message should include `VersionTlv`.
- On parent side the code that appends this TLV is added in this PR itself (from the change in `MleRouter::SendChildUpdateResponse()`)
- If we have a parent that is running a version of OT before this change, the child would reject its "Child Update Response".

abtink: The other places we add a check for VersionTlv in this PR are in `Mle::HandleParentResponse()` and `MleRouter::HandleChildIdRequest()`. Checking the src I see that in current OT master we do already include the version TLV in sent "parent response" or "child id request" msgs, so adding the check in those two should be fine.

jwhui: Thread 1.2.0 Specification does not currently indicate the use of Version TLV. Rather than querying the parent's version, how about saving the parent's version as as part of `ParentInfo` in settings (i.e. non-volatile). Alternatively, we'd need to propose an update to the Thread Specification.

src/core/thread/mesh_forwarder.cpp
jwhui: Yes, I think we sure make sure MAC frame security is enabled.

src/core/mac/mac.cpp
```diff
@@ -1265,6 +1265,12 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
         otDumpDebgMac("TX", aFrame.GetHeader(), aFrame.GetLength());
         FinishOperation();
         Get<MeshForwarder>().HandleSentFrame(aFrame, aError);
+#if OPENTHREAD_THREAD_VERSION >= OPENTHREAD_THREAD_VERSION_1_2
```
abtink: :+1:  on the `#if` condition model for the version check

src/core/thread/mle.cpp
bukepo: @abtink This PR explicitly requests the version TLV when sending child update request. I don't think this will break compatibility. We may choose to save the version, but requesting version in child update request should also work. Thoughts @jwhui

src/core/thread/mesh_forwarder.cpp
bukepo: Agree!

src/core/thread/mle.cpp
jwhui: Current implementation will never include a Version TLV in the Child Update Response. However, this PR currently requires Version TLV to be included in the Child Update Response.

bukepo: Oh, yes. I will make this optional.

jwhui: I think if we want to allow the possibility of requesting Version TLV, we need to make that more explicit in the Thread 1.2.0 Specification. I submitted [SPEC-909](https://threadgroup.atlassian.net/browse/SPEC-909) for further discussion.

src/core/thread/mle.cpp
```diff
@@ -423,6 +423,7 @@ otError Mle::Restore(void)
 
         mParent.Clear();
         mParent.SetExtAddress(*static_cast<Mac::ExtAddress *>(&parentInfo.mExtAddress));
+        mParent.SetVersion(parentInfo.mVersion);
```
bukepo: If `Restore()` fails, just go the full attach process.

src/core/thread/mesh_forwarder.cpp
CiaranWoodward: It is really down to a zero-compromises approach. Going back to Thread 1.1, the Child Update Request proves a fully functional bi-directional link including security, because the ChildUpdateRequest made it up to the parent, was processed, and the ChildUpdateResponse made it back to the child. The MAC Data request never did that, as a parent will Ack a MAC Data Request even if it can't process the security or doesn't consider it a child. This is why we have the sentence:

> Rx-off-when-idle Children SHOULD also periodically (at time intervals indicated by the application) resynchronize by sending a Child Update Request to their Parent to avoid cases when the MAC Data Request is being acknowledged erroneously by a receiver that is not a valid Parent.

In Thread 1.2, we introduced the Enhanced Frame Pending feature, which allows a child to accomplish the same thing as a MAC Data Request command (more or less) with a MAC Data frame - be aware of when the parent has data. Therefore using a MAC Data frame here is strictly an improvement vs Thread 1.1, whereas making it possible for Rx-On children would be a trade off.

abtink: Thanks @CiaranWoodward for explanation on this.

src/core/thread/mle.cpp
abtink: I think best approach for now is to include in request but make parsing of it optional. 
- We include in list of of requested TLVs we sent out in "Child Update Requset"
- Parent running 1.1 version (earlier OT build) would not include this TLV in the "Child Update Response"
- Here on the child, we check if Version TLV is present in the message, we parse it. Otherwise   we assume parent is 1.1.

We can also save version in `ParentInfo` (during parsing we can consider it 1.1 if it is not present), but I favor doing just an optional parsing of the TLV during the "Child Update Response". 


Thoughts?

jwhui: @abtink, I’m fine with the approach, but we need to ensure the Thread Specification explicitly supports querying the Version TLV (SPEC-909 was raised with the Thread Group).

bukepo: Updated to persist parent version and process ChildUpdateResponse as proposed by @abtink

jwhui: To add the ability to request Version TLV, we'll need to wait on the outcome of [SPEC-909](https://threadgroup.atlassian.net/browse/SPEC-909). This also raises another issue of what to do if a device upgrades from Thread 1.1 to Thread 1.2, since Thread 1.1 devices may/will not store Version of peers persistently. I have filed [SPEC-911](https://threadgroup.atlassian.net/browse/SPEC-911) to discuss this issue.

bukepo: I removed requesting Version TLV in Child Update Request.

src/core/mac/data_poll_sender.cpp
```diff
@@ -387,6 +387,15 @@ otError DataPollSender::StopFastPolls(void)
     return error;
 }
 
+void DataPollSender::ResetKeepAliveTimer(void)
+{
+    if (mTimer.IsRunning() && mPollPeriod == GetDefaultPollPeriod())
```
librasungirl: Here why `GetDefaultPollPeriod()`, but not `GetKeepAlivePollPeriod()`?

src/core/thread/mesh_forwarder.cpp
```diff
@@ -1096,6 +1096,18 @@ void MeshForwarder::HandleReceivedFrame(Mac::RxFrame &aFrame)
             LogFrame("Received empty payload frame", aFrame, OT_ERROR_NONE);
         }
 
+#if OPENTHREAD_THREAD_VERSION >= OPENTHREAD_THREAD_VERSION_1_2 && OPENTHREAD_FTD
```
librasungirl: The frame might be drop in HandleFragment() or HandleLowpanHC(), how about moving this block into [MeshForwarder::HandleDatagram](https://github.com/openthread/openthread/blob/master/src/core/thread/mesh_forwarder.cpp#L1366)?

src/core/mac/data_poll_sender.cpp
bukepo: `GetDefaultPollPeriod()` is the real default keep alive period. Otherwise, skipping the poll makes the poll rate unexpected.

src/core/mac/mac.cpp
```diff
@@ -1265,6 +1265,12 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
...
+        if (aError == OT_ERROR_NONE && aAckFrame != NULL)
+        {
+            Get<DataPollSender>().ResetKeepAliveTimer();
```
CiaranWoodward: As we are only using EKA for secured frames, we should confirm security was used here before resetting the timer.

bukepo: In 1.1, data poll is used as keep alive, the ack is not secured. I think this is also acceptable in 1.2. Thoughts?

jwhui: I believe @CiaranWoodward's point here is we should check if the sent frame has security enabled before resetting the keep alive timer. I agree with that comment.

bukepo: Then shall we make this clear in the spec, I remember it states **any** MAC Data Frame can be used as keep alive frame.

jwhui: Yes, this comment was raised a few weeks ago in this PR, [SPEC-908](https://threadgroup.atlassian.net/browse/SPEC-908) was submitted as a result, and you acknowledged it in https://github.com/openthread/openthread/pull/4325#discussion_r348293011.

bukepo: Oh, I see. Thank you!

tests/scripts/thread-cert/thread_cert.py
```diff
@@ -0,0 +1,93 @@
...
+            if params:
+                params = dict(DEFAULT_PARAMS, **params)
+            else:
+                params = DEFAULT_PARAMS.copy()
+            initial_topology[i] = params
+
+            self.nodes[i] = Node(i, params['is_mtd'], simulator=self.simulator)
+            self.nodes[i].set_panid(params['panid'])
+            self.nodes[i].set_mode(params['mode'])
+            self.nodes[i].set_addr64(format(i, '016x'))
```
librasungirl: extaddr should have u/l bit set?

librasungirl: In Thread 1.1 Specification Chapter 3.2

> After the Thread Network security credentials have been successfully obtained using either Thread Commissioning or out-of-band pre- configuration, the specific value provided by the device to the MAC sublayer to represent the aExtendedAddress MUST be a random generated 64-bit value different from the universally administered IEEE EUI-64 value that has been factory assigned to the Thread interface. The random value SHALL be generated from a true random number generation source to ensure a high probability of uniqueness.
> 
> If the 64-bit value resulting after either SHA-256 computation or random generation has the second least significant bit of octet 0 (also called U/L or X bit in IEEE EUI-64 terminology) set to 0, the value provided to the MAC sub-layer as the aExtendedAddress SHALL be modified to have the respective bit set to 1 in order to adhere to a locally administered EUI-64 format.

src/core/thread/mle.cpp
```diff
@@ -2289,9 +2291,11 @@ otError Mle::SendChildUpdateRequest(void)
     switch (mRole)
     {
     case OT_DEVICE_ROLE_DETACHED:
+    {
         Random::Crypto::FillBuffer(mParentRequest.mChallenge, sizeof(mParentRequest.mChallenge));
         SuccessOrExit(error = AppendChallenge(*message, mParentRequest.mChallenge, sizeof(mParentRequest.mChallenge)));
         break;
+    }
```
CiaranWoodward: micronit: Seems not common openthread style to have braces around cases

bukepo:
@CiaranWoodward Thanks for your suggestions. I added `mVersion` field into `NetworkInfo` and `ChildInfo`. A parent will re-attach for `ReadFixedSize()` failure.

src/core/common/settings.hpp
```diff
@@ -100,6 +100,7 @@ class SettingsBase : public InstanceLocator
         uint32_t        mPreviousPartitionId;    ///< PartitionId
         Mac::ExtAddress mExtAddress;             ///< Extended Address
         uint8_t         mMlIid[OT_IP6_IID_SIZE]; ///< IID from ML-EID
+        uint16_t        mVersion;                ///< Version
```
jwhui: Where is this used? I think this would be useful to detect if the device itself is changing versions. In this case, the device should remove any existing child/parent info and perform a full reattach.

src/core/thread/mle_router.cpp
```diff
@@ -3038,6 +3052,10 @@ void MleRouter::SendChildUpdateResponse(Child *                 aChild,
     {
         switch (aTlvs[i])
         {
+        case Tlv::kVersion:
+            SuccessOrExit(error = AppendVersion(*message));
```
jwhui: I don't think Thread 1.1/1.2 allows requesting the Version TLV. Since we are saving this information in non-volatile, can we simply assume the Thread Version is 1.1 when non-volatile does not include version information?

bukepo: Removed.

src/core/thread/mle.cpp
```diff
@@ -375,6 +375,9 @@ otError Mle::Restore(void)
 
     SuccessOrExit(error = Get<Settings>().ReadNetworkInfo(networkInfo));
 
+    // force re-attach when version mismatch.
+    VerifyOrExit(networkInfo.mVersion == kThreadVersion);
```
jwhui: I would move this check to just before the `switch (networkInfo.Role)` line below.

src/core/thread/mle_router.cpp
```diff
@@ -2444,6 +2451,13 @@ otError MleRouter::HandleChildUpdateResponse(const Message &         aMessage,
 
     LogMleMessage("Receive Child Update Response from child", aMessageInfo.GetPeerAddr(), child->GetRloc16());
 
+    // Version
+    if (Tlv::GetTlv(aMessage, Tlv::kVersion, sizeof(version), version) == OT_ERROR_NONE)
```
jwhui: Thread 1.2 does not currently specify inclusion of Version TLV in Child Update Response.

src/core/mac/mac.cpp
```diff
@@ -1274,6 +1274,12 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
         otDumpDebgMac("TX", aFrame.GetHeader(), aFrame.GetLength());
         FinishOperation();
         Get<MeshForwarder>().HandleSentFrame(aFrame, aError);
+#if OPENTHREAD_THREAD_VERSION >= OPENTHREAD_THREAD_VERSION_1_2
+        if (aFrame.GetSecurityEnabled() && aError == OT_ERROR_NONE && aAckFrame != NULL)
+        {
+            Get<DataPollSender>().ResetKeepAliveTimer();
```
CiaranWoodward: There needs to be an additional check here so that the timer is only reset when the parent is a Thread 1.2 device. (currently logic is only depending on child [own] version)

bukepo: Thank you! updated.

src/core/thread/topology.hpp
```diff
@@ -358,7 +358,10 @@ class Neighbor
      * @returns TRUE if Enhanced Keep-Alive is supported, FALSE otherwise.
      *
      */
-    bool IsEnhancedKeepAliveSupported(void) const { return mVersion >= OPENTHREAD_THREAD_VERSION_1_2; }
+    bool IsEnhancedKeepAliveSupported(void) const
+    {
+        return mState != kStateValid && mVersion >= OPENTHREAD_THREAD_VERSION_1_2;
```
CiaranWoodward: Is this supposed to be ``mState != kStateInvalid`` instead of ``mState != kStateValid``?

bukepo: Thank you! This also exposes the test didn't cover resetting timer, so I enhanced the test script to verify the timer is reset.

bukepo:
@abtink @jwhui any comments?

bukepo:
ping~~

src/core/common/settings.hpp
abtink: I think adding field to `NetworkInfo` struct  without a behavior change (like from https://github.com/openthread/openthread/pull/4499) would cause the read from previously written settings (by a previous version i.e. version without this change) to fail.

We have same situation with `ParentInfo` and `ChildInfo`, but I guess they are less critical. If we cannot read `NetworkInfo` device would not be able to connect back to network.

We need to follow

>      * 3. To support forward compatibility (rolling forward to a newer
>      *    software version), code reading and processing key values MUST
>      *    process key values that have shorter length.

@bukepo @jwhui  thoughts?

bukepo: According to the latest discussion, changing version should do full reattach. However, I just realized this new field doesn't mean the version is changed(i.e. users can still configure the version to 1.1 after this PR is merged). I will update the code so that only reattach when Thread version is really changed.

bukepo: I will wait the consensus of #4499.

abtink: @bukepo actually I think it may work out fine. Adding the field would cause in the very first run (with new firmware) not being able to read previously saved `NetworkInfo`, but we would still have Active Dataset. I think we would attach again with a new (randomly generated) extended address (I guess this is fine and we show up as a new node in network). One concern is that we won't recover our previous `mDeviceMode` (which user may have set on device last) but we can require next layers to set it again in such a case?

bukepo: @abtink Yes, I think it can reattach successfully.  Even a user use source code after this PR is merged, it can still configure the Thread version to 1.1. In such case, the version is not changed, is it necessary to reattach? @jwhui

jwhui: This issue should be resolved by #4499.

src/core/common/settings.cpp
```diff
@@ -50,10 +50,11 @@ namespace ot {
 
 void SettingsBase::LogNetworkInfo(const char *aAction, const NetworkInfo &aNetworkInfo) const
 {
-    otLogInfoCore("Non-volatile: %s NetworkInfo {rloc:0x%04x, extaddr:%s, role:%s, mode:0x%02x, keyseq:0x%x, ...",
-                  aAction, aNetworkInfo.GetRloc16(), aNetworkInfo.GetExtAddress().ToString().AsCString(),
-                  Mle::Mle::RoleToString(static_cast<otDeviceRole>(aNetworkInfo.GetRole())),
-                  aNetworkInfo.GetDeviceMode(), aNetworkInfo.GetKeySequence());
+    otLogInfoCore(
+        "Non-volatile: %s NetworkInfo {rloc:0x%04x, extaddr:%s, role:%s, mode:0x%02x, version: %hu, keyseq:0x%x, ...",
```
abtink:
```suggestion
        "Non-volatile: %s NetworkInfo {rloc:0x%04x, extaddr:%s, role:%s, mode:0x%02x, version:%hu, keyseq:0x%x, ...",
```
nit: removing the space between `:` and value.

src/core/thread/mesh_forwarder.cpp
```diff
@@ -1045,6 +1045,18 @@ void MeshForwarder::HandleReceivedFrame(Mac::RxFrame &aFrame)
...
```
abtink: Can we consider moving this whole check in `Mac::HandleReceivedFrame()` right before we call `Get<MeshForwarder>().HandleReceivedFrame(*aFrame);`.
The advantage I see with moving this is that we already have `neighbor` in the `Mac` (so no need to do `FindChild`).

Thoughts?

bukepo: Brilliant!

bukepo:
@LuDuda do you have any comments?

bukepo:
ping~

src/core/openthread-core-config.h
```diff
@@ -36,13 +36,19 @@
...
+
 #define OPENTHREAD_CORE_CONFIG_H_IN
 
 #ifdef OPENTHREAD_PROJECT_CORE_CONFIG_FILE
 #include OPENTHREAD_PROJECT_CORE_CONFIG_FILE
 #endif
 
-#define OPENTHREAD_THREAD_VERSION (2)
+#ifndef OPENTHREAD_THREAD_VERSION
+#define OPENTHREAD_THREAD_VERSION OPENTHREAD_THREAD_VERSION_1_1
```
jwhui: Should we call this `OPENTHREAD_CONFIG_THREAD_VERSION` for consistency?

src/core/openthread-core-config.h
```diff
@@ -36,13 +36,19 @@
 
 #include <openthread/config.h>
 
+#define OPENTHREAD_THREAD_VERSION_INVALID 0
```
jwhui: For other constants, we use the `OT_` prefix. Thoughts?

bukepo:
Updated.

PR #4326, opened 2019-11-13
[mle] fix unexpected data request

librasungirl:
This commit compares data version properly according to the device
mode (whether or not requiring full network data), thus avoiding
unexpected data request when receiving MLE advertisement.

### DIFF
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -2877,11 +2877,7 @@ otError Mle::HandleAdvertisement(const Message &aMessage, const Ip6::MessageInfo
         break;
     }
 
-    if (mRetrieveNewNetworkData ||
-        (IsFullNetworkData() &&
-         (static_cast<int8_t>(leaderData.GetDataVersion() - Get<NetworkData::Leader>().GetVersion()) > 0)) ||
-        (!IsFullNetworkData() &&
-         (static_cast<int8_t>(leaderData.GetStableDataVersion() - Get<NetworkData::Leader>().GetStableVersion()) > 0)))
+    if (mRetrieveNewNetworkData || IsNetworkDataNewer(leaderData))
     {
         delay = Random::NonCrypto::GetUint16InRange(0, kMleMaxResponseDelay);
         SendDataRequest(aMessageInfo.GetPeerAddr(), tlvs, sizeof(tlvs), delay);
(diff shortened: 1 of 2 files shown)


src/core/thread/mle.cpp
```diff
@@ -2878,7 +2878,10 @@ otError Mle::HandleAdvertisement(const Message &aMessage, const Ip6::MessageInfo
     }
 
     if (mRetrieveNewNetworkData ||
-        (static_cast<int8_t>(leaderData.GetDataVersion() - Get<NetworkData::Leader>().GetVersion()) > 0))
+        (IsFullNetworkData() &&
```
jwhui: We have similar code in `Mle:: HandleLeaderData`: https://github.com/openthread/openthread/blob/3dbd91aa2b70c7d5cc71b2c465ce3583a13dea79/src/core/thread/mle.cpp#L2955-L2963

Can we introduce a common method like:
```c++
bool Mle::IsNetworkDataNewer(const LeaderDataTlv &aLeaderData)
{
}
```
?

librasungirl: Updated. Thanks.

src/core/thread/mle.hpp
```diff
@@ -907,6 +907,17 @@ class Mle : public InstanceLocator
...
+    /**
+     * This method indicates whether or not there is newer available Network Data.
+     *
+     * @param[in]  aLeaderData  A reference to the leader data.
+     *
+     * @retval TRUE  If Network Data version in @p aLeaderDataa is newer.
+     * @retval FALSE If Network Data version in @p aLeaderDataa is not newer.
+     *
+     */
+    bool IsNetworkDataNewer(const LeaderDataTlv &aLeaderData);
```
jwhui: Make this private?

src/core/thread/mle.cpp
```diff
@@ -2919,6 +2918,22 @@ otError Mle::HandleDataResponse(const Message &         aMessage,
...
+    if (IsFullNetworkData())
+    {
+        diff = static_cast<int8_t>(aLeaderData.GetDataVersion() - Get<NetworkData::Leader>().GetVersion());
+    }
+    else
+    {
+        diff = static_cast<int8_t>(aLeaderData.GetStableDataVersion() - Get<NetworkData::Leader>().GetStableVersion());
+    }
+
+    return (diff > 0) ? true : false;
```
jwhui: nit: simplify to `return (diff > 0);`

PR #4329, opened 2019-11-14
[debug] override any existing assert definitions

gjc13:
This prevent build error if the platform drivers have already defined
the assert macro.

### DIFF
--- a/src/core/common/debug.hpp
+++ b/src/core/common/debug.hpp
@@ -56,6 +56,10 @@
 #define FILE_NAME __FILE__
 #endif
 
+#ifdef assert
+#undef assert
+#endif
+
 #define assert(cond)                               \
     do                                             \
     {                                              \
@@ -69,6 +73,10 @@
     } while (0)
 
 #else
+
+#ifdef assert
+#undef assert
+#endif
 
 #define assert(cond)  \
     do                \


src/core/common/debug.hpp
```diff
@@ -56,6 +56,7 @@
 #define FILE_NAME __FILE__
 #endif
 
+#ifndef assert
```
abtink: We have `OPENTHREAD_CONFIG_PLATFORM_ASSERT_MANAGEMENT` which requires platform to provide `otPlatAssertFail()` for OT core to use and I think in that case platform should not define `assert()` itself.

Also I am curious about something: `debug.hpp` is private header file intended only to use inside OT core. How would we get a platform definition of assert? Which include path would add this definition?

gjc13: The code has been updated to undefine assert so that we can keep the assert implementation in OpenThread code.

abtink: Thanks. This PR is fine with me.

We chatted offline on this and @gjc13  helped track the header file include chain: Seems like the issue happens through mbedtls header files (which are included in `"random_manager.hpp"`), in particular when project defines its own "alt" mbedtls header files to override some of the definitions. The "alt" header files may include platform specific assert definitions.

As a possible solution for longer-term (not in this PR), I suggest we define a macro `OT_ASSERT` and use it inside core (and map it to either `assert` from `#include <assert.h>` or  `otPlatAssertFail`). Thoughts?

bukepo: OpenThread's `assert()` mimic the standard `assert()`, which is also a macro function. Only define when it's not defined seems to be reasonable, which allows OpenThread use the platform's own assert. Thoughts?

abtink: @bukepo when `OPENTHREAD_CONFIG_PLATFORM_ASSERT_MANAGEMENT` is used we expect the platform to provide a `otPlatAssertFail` (which includes filename and line number as paramter) and we use this to define `assert()` to include the file name and line number (within OT core). We cannot be sure that an existing definition of `assert()` is doing the same.

PR #4331, opened 2019-11-14
[mle] add backbone-link link type

bukepo:
This PR introduces a new Thread link type: backbone-link link type,
which enables transmitting Thread packets over the backbone link, so
that Thread nodes that cannot hear from each other through 15.4 link can
attach the same Thread partition, thus Thread network fragments can be
avoided or reduced.

The backbone link is part of an external IPv4 or IPv6 network, either
physical or virtual. Here's a structure example.

```
                  Backbone Link          Backbone Link
                     Ethernet                Wi-Fi
                 ---------------        -----------------
                  |           |          |             |
Router     R1 -- R2          R3 ------- R4 --- R5 ..   |
                /  \         / \        /|\        .   |
 Child         C1  C2       C3 C4     C5 | \       ... C6
                                         |  \
 ToBLE                                  C7   C8

                Figure 1 - Example network
```

The backbone link can be seen as other PHYs of Thread. Each Thread node
can have multiple PHYs, and it can hear another node through multiple
PHYs. A node can only establish a single link with another node, either
via backbone-link link type or the 15.4 link type. This simplifies the
design and allows supporting multiple link types with only a few changes
in OpenThread's core stack.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,10 @@ stages:
 
 jobs:
   include:
+    - env: BUILD_TARGET="posix-backbone-link" VERBOSE=1 COVERAGE=1
+      os: linux
+      compiler: clang
+      script: script/test bblink
     - env: BUILD_TARGET="posix-app-cli" VERBOSE=1 VIRTUAL_TIME=1
       os: linux
       compiler: gcc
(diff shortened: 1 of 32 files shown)


jwhui:
Superseded by #4440

PR #4334, opened 2019-11-15
[mac] add enhanced frame pending

bukepo:
This PR adds enhanced frame pending feature of Thread 1.2. This replies
the platform radio driver to correctly set frame pending bit for Data
frames and MAC Data Request frames.


bukepo:
This will be rebased after #4325 is merged.

src/core/thread/mesh_forwarder.cpp
```diff
@@ -909,6 +911,19 @@ void MeshForwarder::HandleSentFrame(Mac::TxFrame &aFrame, otError aError)
         mMessageNextOffset = mSendMessage->GetLength();
 #endif
     }
+#if OPENTHREAD_THREAD_VERSION >= OPENTHREAD_THREAD_VERSION_1_2
+    else if (aAckFrame != NULL)
```
CiaranWoodward: I think it's necessary to check here that the parent is Thread Version 1.2 or greater. There is no guarantee that all Thread 1.1 devices are kind about setting their FP bit in response to data messages. Also resetting the keepalive for all data messages to 1.1 parent would break backwards compatibility.

src/core/thread/mesh_forwarder.cpp
```diff
@@ -909,6 +911,19 @@ void MeshForwarder::HandleSentFrame(Mac::TxFrame &aFrame, otError aError)
         mMessageNextOffset = mSendMessage->GetLength();
 #endif
     }
+#if OPENTHREAD_THREAD_VERSION >= OPENTHREAD_THREAD_VERSION_1_2
```
abtink: If we do this check in `MeshForwarder` we are restricting the check to only direct tx requests. 
If we do the same check in `Mac`, we can cover acks to any type of tx (in particular I am considering `kOperationTransmitOutOfBandFrame`).

Another suggestion: Maybe we can re-use the `DataPollSender::CheckFramePending(RxFrame)` method?

Thoughts?

examples/platforms/posix/radio.c
```diff
@@ -715,7 +714,13 @@ void radioSendAck(void)
     sAckFrame.mLength    = IEEE802154_ACK_LENGTH;
     sAckMessage.mPsdu[0] = IEEE802154_FRAME_TYPE_ACK;
 
-    if (isDataRequestAndHasFramePending(&sReceiveFrame))
+    if (
+#if OPENTHREAD_THREAD_VERSION >= OPENTHREAD_THREAD_VERSION_1_2
+        (otMacFrameIsDataRequest(&sReceiveFrame) || otMacFrameIsDataRequest(&sReceiveFrame))
```
dismirlian: Shouldn't this be like this?:

```suggestion
        (otMacFrameIsDataRequest(&sReceiveFrame) || otMacFrameIsData(&sReceiveFrame))
```

src/core/thread/mesh_forwarder.cpp
bukepo: Moved into MAC layer and changed to reuse CheckFramePending(), but leave out of band frame handling for future because we are considering remove it.

src/core/mac/data_poll_sender.cpp
```diff
@@ -317,6 +317,12 @@ void DataPollSender::CheckFramePending(Mac::RxFrame &aFrame)
     {
```
brianljt: Here, for OT1.1 devices, it will send out data poll whenever it received a frame with frame pending set. This means if peer OT1.2 parent send out ACK with frame pending set in response to MAC data frame, OT1.1 child will follow up with a data poll which is against spec.

src/core/mac/data_poll_sender.cpp
```diff
@@ -317,6 +317,12 @@ void DataPollSender::CheckFramePending(Mac::RxFrame &aFrame)
...
         SendDataPoll();
     }
+    else if (aIsAck)
```
brianljt: I think it is better to limit the function `CheckFramePending` to send data poll in accordance with its name. Maybe we can put this keep alive handling to a separate function or just add it in `Mac::HandleReceivedFrame` and `Mac::HandleTransmitDone`?

src/core/mac/data_poll_sender.hpp
```diff
@@ -172,7 +172,7 @@ class DataPollSender : public InstanceLocator
      * pending frame.
      *
      */
-    void CheckFramePending(Mac::RxFrame &aFrame);
```
brianljt: Please modify comments to reflect the new usage of this function if you want to use it this way.

src/core/mac/mac.cpp
```diff
@@ -1303,6 +1303,13 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
         otDumpDebgMac("TX", aFrame.GetHeader(), aFrame.GetLength());
         FinishOperation();
         Get<MeshForwarder>().HandleSentFrame(aFrame, aError);
+#if OPENTHREAD_THREAD_VERSION >= OPENTHREAD_THREAD_VERSION_1_2
+        if (aError == OT_ERROR_NONE && Get<Mle::Mle>().GetParent().IsEnhancedKeepAliveSupported() &&
+            aFrame.GetSecurityEnabled() && aAckFrame != NULL)
```
brianljt: Should check `aFrame.otMacFrameIsData()`?

src/core/mac/data_poll_sender.cpp
```diff
@@ -307,7 +307,7 @@ void DataPollSender::HandlePollTimeout(void)
     return;
 }
 
-void DataPollSender::CheckFramePending(Mac::RxFrame &aFrame)
+void DataPollSender::CheckFramePending(Mac::RxFrame &aFrame, bool aIsAck)
```
brianljt: For OT1.2, data poll is used to retrieve pending packets rather than check and retrieve data from parent. So the naming of this function does not really reflect its purpose here.

src/core/mac/mac.cpp
edmont: @brianljt yes, otherwise we would be sending MAC Data Requests one after each other.

src/core/thread/mesh_forwarder.cpp
```diff
@@ -1045,6 +1045,18 @@ void MeshForwarder::HandleReceivedFrame(Mac::RxFrame &aFrame)
             LogFrame("Received empty payload frame", aFrame, OT_ERROR_NONE);
         }
 
+#if OPENTHREAD_THREAD_VERSION >= OPENTHREAD_THREAD_VERSION_1_2 && OPENTHREAD_FTD
```
librasungirl: nit: `OPENTHREAD_THREAD_VERSION` -> `OPENTHREAD_CONFIG_THREAD_VERSION`
`OPENTHREAD_THREAD_VERSION_1_2`->`OT_THREAD_VERSION_1_2`

examples/platforms/posix/radio.c
```diff
@@ -715,7 +714,13 @@ void radioSendAck(void)
     sAckFrame.mLength    = IEEE802154_ACK_LENGTH;
     sAckMessage.mPsdu[0] = IEEE802154_FRAME_TYPE_ACK;
 
-    if (isDataRequestAndHasFramePending(&sReceiveFrame))
+    if (
+#if OPENTHREAD_THREAD_VERSION >= OPENTHREAD_THREAD_VERSION_1_2
```
librasungirl: nit: `OPENTHREAD_THREAD_VERSION` -> `OPENTHREAD_CONFIG_THREAD_VERSION`
`OPENTHREAD_THREAD_VERSION_1_2`->`OT_THREAD_VERSION_1_2`

src/core/mac/data_poll_sender.cpp
```diff
@@ -317,6 +317,12 @@ void DataPollSender::CheckFramePending(Mac::RxFrame &aFrame)
     {
         SendDataPoll();
     }
+    else if (aIsAck)
+    {
+#if OPENTHREAD_THREAD_VERSION >= OPENTHREAD_THREAD_VERSION_1_2
```
librasungirl: nit: `OPENTHREAD_THREAD_VERSION` -> `OPENTHREAD_CONFIG_THREAD_VERSION`
`OPENTHREAD_THREAD_VERSION_1_2`->`OT_THREAD_VERSION_1_2`

src/core/mac/data_poll_sender.cpp
jwhui: Sending an IEEE 802.15.4 Data Request message in response to receiving an IEEE 802.15.4 ACK with Frame Pending = 1 is intentional here to reduce latency of transferring messages. I don't think the Thread 1.1 and 1.2 specifications prohibit this behavior. If you believe this behavior is not conformant, can you provide references to the relevant portions of the spec?

jwhui: Yes, for Thread 1.1, we want to ignore the Frame Pending bit in ACKs that are **not** in response to IEEE 802.15.4 Data Request frames.

src/core/thread/mesh_forwarder.cpp
```diff
@@ -1045,6 +1045,18 @@ void MeshForwarder::HandleReceivedFrame(Mac::RxFrame &aFrame)
             LogFrame("Received empty payload frame", aFrame, OT_ERROR_NONE);
         }
 
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2 && OPENTHREAD_FTD
+        // From Thread 1.2, MAC Data Frame can also act as keep-alive message if child supports
```
jwhui: Wasn't this functionality added with #4325 

https://github.com/openthread/openthread/blob/fe6a7e1493ef8dcbccbe19b5bcb94e377a41a848/src/core/mac/mac.cpp#L1678-L1685

examples/platforms/posix/radio.c
```diff
@@ -715,7 +714,13 @@ void radioSendAck(void)
...
+        (otMacFrameIsData(&sReceiveFrame) || otMacFrameIsDataRequest(&sReceiveFrame))
```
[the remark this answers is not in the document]
jwhui: Agree that it is not strictly required by the specification, but useful to not set the Frame Pending bit in ACKs sent in response to Data frames.

src/core/mac/data_poll_sender.cpp
jwhui: I agree that we should separate keep alive handling from frame pending handling.

examples/platforms/posix/radio.c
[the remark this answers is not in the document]
bukepo: I chose the current implementation to make the code simpler, otherwise, the radio driver must know the peer's Thread version in the context of interrupt.
We could use something like the src match table for this, or verify the frame version is 2015.
However, I prefer the simpler solution as Thread 1.1 explicitly stated to ignore the bit. Thoughts?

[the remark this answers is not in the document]
brianljt: Please check the below comments. At least devices shipped with OT1.1 in current code will not ignore the Frame Pending bit in ACKs not in response to Data Request frames.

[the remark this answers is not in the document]
bukepo: If current 1.1 code doesn't ignore, then it's a bug. I think we should fix the bug as well.

src/core/mac/mac.cpp
bukepo: I think the operation is transmitting data only.

src/core/mac/data_poll_sender.cpp
bukepo: @abtink suggested reuse the method. And I also found it can avoid duplicate code. I agree to update the name as `ProcessFrame(aFrame)` to better reflect the functionality.

src/core/thread/mesh_forwarder.cpp
bukepo: rebase caused issue..

src/core/mac/data_poll_sender.cpp
librasungirl: @brianljt 
I think the OT1.1 behavior with this PR is the same as before this PR. Note that only 1.2 will pass AckFrame to `CheckFramePending()`.
https://github.com/openthread/openthread/pull/4334/files#diff-073e11a5a34cdfd560305449d7763fe3R1306

Anything I missed regarding 1.1 behavior?

abtink: Just adding a quick note. I think existing code (master) and 1.1 behavior is correct
- On master  `CheckFramePending()` is only called from rx path from `HandleReceivedFrame()`
- Received acks won't go through that path and are processed from tx callback `Mac::HandleTransmitDone()` 
- A pending bit in a received ack frame is only checked if the tx frame is data poll.

src/core/mac/mac_frame.hpp
```diff
@@ -360,6 +360,16 @@ class Frame : public otRadioFrame
      */
     uint8_t GetType(void) const { return GetPsdu()[0] & kFcfFrameTypeMask; }
 
+    /**
+     * This method returns whether the frame is an Ack frame.
+     *
+     * @retval TRUE   If this is an Ack.
+     * @retval FALSE  If this is not an Ack.
+     *
```
abtink:
```suggestion
 ```

File: src/core/mac/mac.cpp (hunk 7 above)
Comment (abtink) [reply src/core/mac/mac.cpp]: I agree with @bukepo . Data polls acks are handled under `case kOperationTransmitPoll` above. Here we are under   `case kOperationTransmitDataDirect` (which means this is a recieved ack for a data frame tx and not a data poll tx).

File: src/core/mac/mac_frame.hpp [hunk 15]
```diff
@@ -360,6 +360,16 @@ class Frame : public otRadioFrame
...
+     *
+     */
+    bool IsAck(void) const { return (GetPsdu()[0] & kFcfFrameTypeMask) == kFcfFrameAck; }
```
abtink:
```suggestion
    bool IsAck(void) const { return GetType() == kFcfFrameAck; }
```

src/core/mac/data_poll_sender.cpp
brianljt: Thank you @abtink @librasungirl for pointing this out. Then I think it is fine for OT1.1.

src/core/mac/data_poll_sender.hpp
```diff
@@ -171,8 +171,11 @@ class DataPollSender : public InstanceLocator
      * the received frame header and if it is set, data poll sender will send an immediate data poll to retrieve the
      * pending frame.
      *
+     * If enhanced keep-alive is supported, this method also resets keep-alive timer if the frame is an Ack.
+     *
+     * @param[in]   aFrame
      */
-    void CheckFramePending(Mac::RxFrame &aFrame);
+    void ProcessFrame(const Mac::RxFrame &aFrame);
```
CiaranWoodward: I think it would be useful to update the first line of the doxy comment here, as I'm not sure the usage is totally intuitive. It seems ``ProcessFrame`` is only called to notify the data poll sender in 2 situations:
1. Receiving a Data Frame indirectly, in both 1.1 and 1.2. (if FP is set we send another data poll.)
1. Receiving an Ack for a sent data frame in 1.2 only. (If FP we send a data poll, if not we reset the keepalive timer for sending next poll.)

tests/scripts/thread-cert/v1_2_test_enhanced_frame_pending.py
```diff
@@ -0,0 +1,113 @@
...
+
+    Leader ----- SED_1
+
+1. Set up topology a Leader and a SED.
+2. SED ping leader so a data poll is just sent.
+3. Leader sends a UDP packet to SED to put a pending frame in queue.
+4. Wait for half polling period, verify no packet received by SED.
+5. SED sends a UDP packet to Leader to solicit a ACK with frame pending set.
+   * Verify Leader receives a UDP packet
+   * Verify SED sent a data poll
```
edmont:
```suggestion
   * Verify SED sends a data poll
```

src/core/mac/data_poll_sender.hpp
bukepo: Thank you for reviewing. Would you like to contribute a GitHub suggestion  (e.g. https://github.com/openthread/openthread/pull/4334#discussion_r388277192) for the doxy?

tests/scripts/thread-cert/common.py
```diff
@@ -37,6 +37,8 @@
 # Map of 2 bits of parent priority.
 pp_map = {1: 1, 0: 0, 3: -1, 2: -2}
 
+UDP_TEST_PORT = 12345
```
librasungirl: move to config.py ?

tests/scripts/thread-cert/v1_2_test_enhanced_frame_pending.py
```diff
@@ -0,0 +1,112 @@
...
+
+Topology
+
+    Leader ----- SED_1
+
+1. Set up topology a Leader and a SED.
+2. SED pings leader so a data poll is just sent.
+3. Leader sends a UDP packet to SED to put a pending frame in queue.
+4. Wait for half polling period, verify no packet received by SED.
+5. SED sends a UDP packet to Leader to solicit a ACK with frame pending set.
```
librasungirl: nit: an Ack

tests/scripts/thread-cert/v1_2_test_enhanced_frame_pending.py
```diff
@@ -0,0 +1,112 @@
...
+
+        # 3 - Send to SED
+        self.nodes[LEADER].udp_send(UDP_BYTES_COUNT,
+                                    self.nodes[SED_1].get_rloc(),
+                                    common.UDP_TEST_PORT)
+
+        # 4 - Wait for half polling period
+        self.simulator.go(DEFAULT_POLL_PERIOD // 2)
+        with self.assertRaises(pexpect.TIMEOUT):
+            self.nodes[LEADER].udp_check_rx(UDP_BYTES_COUNT)
```
librasungirl: Here should be `SED_1`, not `LEADER`?

tests/scripts/thread-cert/v1_2_test_enhanced_frame_pending.py
```diff
@@ -0,0 +1,112 @@
...
+
+        # 5 - Send to Leader
+        self.nodes[SED_1].udp_send(UDP_BYTES_COUNT,
+                                   self.nodes[LEADER].get_rloc(),
+                                   common.UDP_TEST_PORT)
+        self.simulator.go(1)
+        self.nodes[LEADER].udp_check_rx(UDP_BYTES_COUNT)
+        sed_messages = self.simulator.get_messages_sent_by(SED_1)
+        sed_messages.next_data_poll()
```
librasungirl: add assertion?

Or update to be like `next_data_poll(assert_enabled=True)`?

tests/scripts/thread-cert/common.py
bukepo: config.py is imported by ipv6.py, causing circular imports.

src/core/mac/data_poll_sender.hpp
bukepo: @CiaranWoodward could you review the updated doc string?

CiaranWoodward: Apologies for not getting around to your request, but new text looks good to me!

NMzen:
Hi, I wanted to use this "Enhanced Frame Pending" feature of OT. I have ported nRF5 SDK for Thread and Zigbee v4.0.0. Can I recompile the openthread library with the above commit and use it as-is in nRF SDK? Or Do i need to wait for the next release of the nRF SDK?

bukepo:
@LuDuda

PR #4336, opened 2019-11-15
[posix-app] add log level argument

bukepo:


src/posix/main.c
```diff
@@ -191,6 +198,7 @@ static otInstance *InitInstance(int aArgCount, char *aArgVector[])
         exit(OT_EXIT_SUCCESS);
     }
 
+    otLoggingSetLevel(logLevel);
```
jwhui: Does this assume that `OPENTHREAD_CONFIG_ENABLE_DYNAMIC_LOG_LEVEL=1`? Is it necessary to support `OPENTHREAD_CONFIG_ENABLE_DYNAMIC_LOG_LEVEL=0`?

bukepo: Yes, this assumes dynamic logging is enabled. It's seems we'll have to make some assumptions unless openthread-core-config.h is exported for user, which is current not exported. Thoughts?

jwhui: If this is not optional, should we make the `DYNAMIC_LOG_LEVEL` build switch in `src/posix/Makefile-posix` mandatory?

bukepo: `DYNAMIC_LOG_LEVEL` is set to 1 in `src/posix/Makefile-posix`. If a user overrides this to `0`, he/she can modify the main.c accordingly.

PR #4337, opened 2019-11-15
[travis] enable posix-app-ncp test

### LINKED ISSUE #3635: [travis] posix-app-ncp test fails frequently
@bukepo, any thoughts?

jwhui:
The `posix-app-ncp` test seems to be stable again.

Resolves #3635

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,10 @@ jobs:
       os: linux
       compiler: gcc
       script: .travis/script.sh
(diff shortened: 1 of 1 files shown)


jwhui:
> I don’t remember we enhanced this. Maybe we can mark it as allowed failure in case it goes bad again. Thoughts?

FWIW, I have run 10 iterations without failure.

bukepo:
I remember it failed more often when there were many jobs running. However, we can go this way first and take action after it fails again.

PR #4338, opened 2019-11-18
[mle] clean up MleRouter::HandleAdvertisement

jwhui:
- Ensure there are no side effects if MLE frame counters have not yet
been synchronized.

- Do not remove stale neighbor until end of processing
Advertisement. This allows for complete processing of Advertisement
if frame counters have already been established.

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -600,12 +600,6 @@ otError MleRouter::HandleLinkRequest(const Message &aMessage, const Ip6::Message
 
         rloc16 = sourceAddress.GetRloc16();
 
-        if (aNeighbor && aNeighbor->IsStateValid() && aNeighbor->GetRloc16() != rloc16)
-        {
-            // remove stale neighbors
-            RemoveNeighbor(*aNeighbor);
-        }
-
         if (IsActiveRouter(rloc16))
         {
             Mac::ExtAddress macAddr;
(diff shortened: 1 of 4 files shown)


src/core/thread/mle_router.cpp
```diff
@@ -1215,6 +1210,8 @@ otError MleRouter::HandleAdvertisement(const Message &aMessage, const Ip6::Messa
     }
     else if (leaderData.GetLeaderRouterId() != GetLeaderId())
     {
+        VerifyOrExit(aNeighbor && aNeighbor->GetState() == Neighbor::kStateValid);
```
abtink: how about `aNeighbor->IsStateValid()` instead?

src/core/thread/mle_router.cpp
```diff
@@ -1333,17 +1332,25 @@ otError MleRouter::HandleAdvertisement(const Message &aMessage, const Ip6::Messa
...
-            router->ResetLinkFailures();
-            router->SetLastHeard(TimerMilli::GetNow());
-            router->SetState(Neighbor::kStateLinkRequest);
-            SendLinkRequest(router);
-            ExitNow(error = OT_ERROR_NO_ROUTE);
+            // MLE Advertisement not from parent, but from some other neighboring router
+            router = mRouterTable.GetRouter(routerId);
+            VerifyOrExit(router != NULL);
+
+            if (IsFullThreadDevice() && (router->GetState() != Neighbor::kStateValid) &&
```
abtink: how about using `router->IsState<State>()` here for `Valid` and `LinkRequest`?

PR #4340, opened 2019-11-19
[posix-app] remove function "HdlcInterface::IsDecoding()"

zhanglongxia:
The function "RadioSpinel::IsSafeToHandleNow()" is finally called by the
function "HdlcInterface::Decode()". However, the variable "mIsDecoding"
is set to TRUE before calling "RadioSpinel::IsSafeToHandleNow()" in
function "HdlcInterface::Decode()". So the function "IsDecoding()"
always return TURE in function "RadioSpinel::IsSafeToHandleNow()". This
CL removes the function "IsDecoding()" to make the logic of the code easier.

### DIFF
--- a/src/posix/platform/radio_spinel.hpp
+++ b/src/posix/platform/radio_spinel.hpp
@@ -603,7 +603,8 @@ private:
      */
     bool IsSafeToHandleNow(spinel_prop_key_t aKey) const
     {
-        return !(aKey == SPINEL_PROP_STREAM_RAW || aKey == SPINEL_PROP_MAC_ENERGY_SCAN_RESULT);
+        return !((mWaitingKey != SPINEL_PROP_LAST_STATUS) &&
+                 (aKey == SPINEL_PROP_STREAM_RAW || aKey == SPINEL_PROP_MAC_ENERGY_SCAN_RESULT));
     }
 
     void HandleNotification(HdlcInterface::RxFrameBuffer &aFrameBuffer);
--- a/src/posix/platform/radio_spinel.hpp
+++ b/src/posix/platform/radio_spinel.hpp
@@ -603,8 +603,7 @@ private:
      */
     bool IsSafeToHandleNow(spinel_prop_key_t aKey) const
     {
-        return !((mHdlcInterface.IsDecoding() || mWaitingKey != SPINEL_PROP_LAST_STATUS) &&
-                 (aKey == SPINEL_PROP_STREAM_RAW || aKey == SPINEL_PROP_MAC_ENERGY_SCAN_RESULT));
+        return !(aKey == SPINEL_PROP_STREAM_RAW || aKey == SPINEL_PROP_MAC_ENERGY_SCAN_RESULT);
     }
 
     void HandleNotification(HdlcInterface::RxFrameBuffer &aFrameBuffer);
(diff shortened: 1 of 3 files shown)


src/posix/platform/radio_spinel.hpp
```diff
@@ -603,8 +603,7 @@ class RadioSpinel : public HdlcInterface::Callbacks
      */
     bool IsSafeToHandleNow(spinel_prop_key_t aKey) const
     {
-        return !((mHdlcInterface.IsDecoding() || mWaitingKey != SPINEL_PROP_LAST_STATUS) &&
-                 (aKey == SPINEL_PROP_STREAM_RAW || aKey == SPINEL_PROP_MAC_ENERGY_SCAN_RESULT));
+        return !(aKey == SPINEL_PROP_STREAM_RAW || aKey == SPINEL_PROP_MAC_ENERGY_SCAN_RESULT);
```
bukepo: Is it better to keep `mWaitingKey != SPINEL_PROP_LAST_STATUS`?

zhanglongxia: To keep a line with the original logic of the code, we don't need to keep `mWaitingKey != SPINEL_PROP_LAST_STATUS`. @abtink How do you think about it?

abtink: @zhanglongxia , I agree we should not need it.

As you described in the commit log (I also read through the chain and verified it), whenever we get to `IsSafeToHandle()` the `IsDecoding()` should always return `true`. 

So 
`(mHdlcInterface.IsDecoding() || mWaitingKey != SPINEL_PROP_LAST_STATUS)` is always treated as `(true || mWaitingKey != SPINEL_PROP_LAST_STATUS)` which should become `true` (which is then `&&` with other checks).

Let me add the flow of calls:
- `IsSafeToHandleNow()` is only called from `HandleNotification(RxFrameBuffer)`.
- `HandleNotification(RxFrameBuffer)` is called from `HandleSpinelFrame()`
- This  itself is called from `HdlcInterface::Callbacks::HandleReceivedFrame()`
- The HDLC callback can be invoked when HDLC decoder `mHdlcDecoder` finishes decoding a full frame.
- This can only happen when `mHdlcDecoder.Decode()` is called and right before this we set `mIsDecoding` to `true`.

I think the reason we had this check in `IsSafeToHandle()` is that in earlier version `HandleNotification(RxFrameBuffer)` could be called to process queued frames. But in current master code we have two `HandleNotification()` methods (one with `RxFrameBuffer` and other one with a frame pointer and length), and the second is used for queued frames.

bukepo: I think when `mWaitingKey == SPINEL_PROP_LAST_STATUS`, it's safe to handle now, for performance consideration, I prefer keep the check for `mWaitingKey`. Thoughts?

abtink: @bukepo , I agree it should be fine to add such a check (to handle such frame more quickly instead of saving them in queue and processing them later).  

It would change the behavior of code though (as in the existing code we were never actually using this path in the code). This may change the order of processing of some of the received frames. It sounds ok/fine to me.

abtink: I think we may have a typo/error in the existing code and the `||` should have been `&&` in 
`(mHdlcInterface.IsDecoding() || mWaitingKey != SPINEL_PROP_LAST_STATUS)`.

The logic behind this is "if HDLC is decoding but we are not waiting for a response (indicated by `mWaitingKey` being `LAST_STATUS`), it is safe to handle", or saying it other way: "it is NOT safe to handle if HDLC is decoding AND we are waiting for a response.

Does it sound reasonable?

zhanglongxia: @abtink Thanks for your detailed explanation. I have updated the code.

PR #882, opened 2016-10-25
Windows API DLL

nibanks:
The main focus of this PR is to bring in support for a user mode API for OpenThread on Windows. This is done by exposing most all control path APIs in `openthread.h` via a new DLL, `otApi.dll`.

In order to export the functions in a DLL, every function that was to be exported needed to be prefixed with `__declspec(dllexport)` and `__declspec(dllimport)` for the code linking in the function. This was achieved with the use of the `OTAPI` define.

To allow for both x64 and x64 calling conventions on Windows, we use the `OTCALL` define, which syntactically, must come after the return type.

This causes a fairly ugly addition to the openthread headers. If someone has a better suggestion on how to do this, I am all ears.

The change also brings in the last week or two of driver changes.

PR #4344, opened 2019-11-19
[toolchain] remove unnecessary OT_TOOL_ALIGN

abtink:
This commit contains two small related commits:

**[thread] remove unnecessary use of OT_TOOL_ALIGN (#4344)**
    
This commit removes the use of `OT_TOOL_ALIGN` in definitions of `otLinkModeConfig` and `otRouterInfo` structures.

 **[toolchain] remove OT_TOOL_ALIGN (#4344)**

@jwhui , I am not sure if we need the `otLinkModeConfig` and `otRouterInfo` to be defined as `OT_TOOL_ALIGN(4)` (history of change shows that these were added during windows API DLL from #882). Do you agree?

### DIFF
--- a/include/openthread/platform/toolchain.h
+++ b/include/openthread/platform/toolchain.h
@@ -86,13 +86,6 @@ extern "C" {
  */
 
 /**
- * @def OT_TOOL_ALIGN
- *
- * Compiler-specific alignment modifier.
- *
- */
-
-/**
  * @def OT_TOOL_WEAK
  *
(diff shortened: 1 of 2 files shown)


jwhui:
> @jwhui , I am not sure if we need the `otLinkModeConfig` and `otRouterInfo` to be defined as `OT_TOOL_ALIGN(4)` (history of change shows that these were added during windows API DLL from #882). Do you agree?

Yes, I agree.

PR #4349, opened 2019-11-22
Best-effort to wait for the leave process to complete

gjc13:
This change adds a five second wait period for the NCP state to really
change to offline. This will avoid forming the network in an invalid
state which causes the test to fail.

### DIFF
--- a/tests/toranj/wpan.py
+++ b/tests/toranj/wpan.py
@@ -434,7 +434,12 @@ class Node(object):
         return self.wpanctl('status')
 
     def leave(self):
-        return self.wpanctl('leave')
+        res = self.wpanctl('leave')
+        for _ in range(5):
+            if self.get(WPAN_STATE) != STATE_OFFLINE:
+                print('***WARN***: Not in offline state')
+                time.sleep(1)
+        return res
 
     def form(
             self,


tests/toranj/wpan.py
```diff
@@ -434,7 +434,12 @@ def status(self):
         return self.wpanctl('status')
 
     def leave(self):
-        return self.wpanctl('leave')
+        res = self.wpanctl('leave')
+        for _ in range(5):
+            if self.get(WPAN_STATE) != STATE_OFFLINE:
+                print('***WARN***: Not in offline state')
+                time.sleep(1)
+        return res
```
abtink: @gjc13 I think this is related to discussions from https://github.com/openthread/wpantund/pull/444#discussion_r349445074


I am not sure if such a fix to wait for it in test script would be the right approach. I expect the `wpantund` itself would do the same and if it does not then I would say we'd need to address it in `wpantund` itself.

Checking the code, the wpantund leave task does seem to have code that waits for NCP to finish initialization:
https://github.com/openthread/wpantund/blob/9d31a884ef2618776df784e1eec1cc8c25f31943/src/ncp-spinel/SpinelNCPTaskLeave.cpp#L111-L132

Also in the wpantund form task, we seem to wait for some time for NCP to be ready before proceeding:
https://github.com/openthread/wpantund/blob/9d31a884ef2618776df784e1eec1cc8c25f31943/src/ncp-spinel/SpinelNCPTaskForm.cpp#L130-L135

gjc13:
It turned out to be a tricky case on how Linux is sending MLD messages on IPv6 multicast group changes. I'll update the detail in the original wpantund PR.

PR #4352, opened 2019-11-23
[posix-app] optimize the interface between radio_spinel and hdlc_interface

zhanglongxia:
This PR simplifies the API between RadioSpinel and HdlcInterface to improve the extensibility of
the API. So that we can easily add other low layer interfaces to send/receive spinel frames later.
The RadioSpinel is only responsible for encoding/decoding spinel frames. Lower layer interface is only responsible for sending/receiving spinel frames.
Some lower layer interfaces can't provide a file descriptor for RadioSpinel to wait for the file descriptor to become writable, so this PR removes the kStateTransmitPending state from RadioSpinel.c and directly calls the function Request() to send frames in method RadioSpinel::Transmit().

### DIFF
--- a/src/posix/platform/hdlc_interface.cpp
+++ b/src/posix/platform/hdlc_interface.cpp
@@ -271,7 +271,7 @@ otError HdlcInterface::WaitForFrame(struct timeval &aTimeout)
         assert(false);
         break;
     }
-#else // OPENTHREAD_POSIX_VIRTUAL_TIME
+#else  // OPENTHREAD_POSIX_VIRTUAL_TIME
     fd_set read_fds;
     fd_set error_fds;
     int rval;
@@ -306,7 +306,7 @@ otError HdlcInterface::WaitForFrame(struct timeval &aTimeout)
     {
         DieNowWithMessage("wait response", OT_EXIT_FAILURE);
     }
-#endif
+#endif // OPENTHREAD_POSIX_VIRTUAL_TIME
 
 exit:
     return error;
@@ -314,15 +314,15 @@ exit:
 
 void HdlcInterface::UpdateFdSet(fd_set &aReadFdSet, fd_set &aWriteFdSet, int &aMaxFd, struct timeval &aTimeout)
 {
+    OT_UNUSED_VARIABLE(aWriteFdSet);
+    OT_UNUSED_VARIABLE(aTimeout);
+
     FD_SET(mSockFd, &aReadFdSet);
 
     if (aMaxFd < mSockFd)
     {
         aMaxFd = mSockFd;
     }
-
-    OT_UNUSED_VARIABLE(aWriteFdSet);
-    OT_UNUSED_VARIABLE(aTimeout);
 }
 
 void HdlcInterface::Process(const fd_set &aReadFdSet, const fd_set &aWriteFdSet)
--- a/src/posix/platform/hdlc_interface.cpp
+++ b/src/posix/platform/hdlc_interface.cpp
@@ -327,12 +327,12 @@ void HdlcInterface::UpdateFdSet(fd_set &aReadFdSet, fd_set &aWriteFdSet, int &aM
 
 void HdlcInterface::Process(const fd_set &aReadFdSet, const fd_set &aWriteFdSet)
 {
+    OT_UNUSED_VARIABLE(aWriteFdSet);
+
     if (FD_ISSET(mSockFd, &aReadFdSet))
     {
         Read();
     }
-
-    OT_UNUSED_VARIABLE(aWriteFdSet);
 }
 
 otError HdlcInterface::WaitForWritable(void)
--- a/src/posix/platform/hdlc_interface.cpp
+++ b/src/posix/platform/hdlc_interface.cpp
@@ -223,6 +223,8 @@ otError HdlcInterface::Write(const uint8_t *aFrame, uint16_t aLength)
 #if OPENTHREAD_POSIX_VIRTUAL_TIME
     platformSimSendRadioSpinelWriteEvent(aFrame, aLength);
 #else
+    SuccessOrExit(error = WaitForWritable());
+
     while (aLength)
     {
         ssize_t rval = write(mSockFd, aFrame, aLength);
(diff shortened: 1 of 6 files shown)


src/posix/platform/hdlc_interface.cpp
```diff
@@ -247,6 +249,94 @@ otError HdlcInterface::Write(const uint8_t *aFrame, uint16_t aLength)
     return error;
 }
 
+otError HdlcInterface::WaitResponse(struct timeval &aTimeout)
```
bukepo: Is it possible to divide this into `WaitReadable()` and `Read()`? Such that `WaitResponse()` can be implemented in RadioSpinel.

src/posix/platform/radio_spinel.cpp
```diff
@@ -765,30 +768,13 @@ void RadioSpinel::Process(const fd_set &aReadFdSet, const fd_set &aWriteFdSet)
...
-#endif
-        {
-            otPlatRadioTxDone(mInstance, mTransmitFrame, (mAckRadioFrame.mLength != 0) ? &mAckRadioFrame : NULL,
-                              mTxError);
-        }
+        TransmitDone(mTransmitFrame, (mAckRadioFrame.mLength != 0) ? &mAckRadioFrame : NULL, mTxError);
     }
     else if (mState == kStateTransmitting && platformGetTime() >= mTxRadioEndUs)
     {
+        // Frame has been successfully passed to radio, but no `TransmitDone` event received wthin TX_WAIT_US.
```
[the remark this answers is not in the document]
zhanglongxia: Updated

src/posix/platform/hdlc_interface.cpp
```diff
@@ -223,6 +223,8 @@ otError HdlcInterface::Write(const uint8_t *aFrame, uint16_t aLength)
 #if OPENTHREAD_POSIX_VIRTUAL_TIME
     platformSimSendRadioSpinelWriteEvent(aFrame, aLength);
 #else
+    SuccessOrExit(error = WaitForWritable());
```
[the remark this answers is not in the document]
zhanglongxia: Updated

src/posix/platform/hdlc_interface.cpp
zhanglongxia: Yes. We can divide it into ` WaitReadable()` and `Read()`. But if we do it like this, it increases an API between RadioSpinel and HdlcInterface and reduces the flexibility of the low layer.

src/posix/platform/hdlc_interface.cpp
```diff
@@ -247,6 +247,94 @@ otError HdlcInterface::Write(const uint8_t *aFrame, uint16_t aLength)
...
+    FD_SET(mSockFd, &read_fds);
+    FD_SET(mSockFd, &error_fds);
+
+    rval = select(mSockFd + 1, &read_fds, NULL, &error_fds, &aTimeout);
+
+    if (rval > 0)
+    {
+        if (FD_ISSET(mSockFd, &read_fds))
+        {
+            Read();
```
abtink: We can possibly get part of a full HDLC frame from `Read()` and in that case we return from `WaitResponse()` without any full spinel frame received.  

Checking how we use `WaitForResponse()` in the `RadioSpinel`, I think this behavior is fine, but we should be careful of this case.

I would suggest maybe changing the `WaitForResponse()` description.

Thoughts?

zhanglongxia: For HDLC interface, it indeed may get part of a full HDLC frame from `read()`. I have updated the description for the function `WaitResponse()`.

src/posix/platform/hdlc_interface.cpp
bukepo: As discussed, we should rename this method, because request/response is spinel level concepts.

zhanglongxia: Updated.

src/posix/platform/hdlc_interface.hpp
```diff
@@ -170,6 +152,35 @@ class HdlcInterface
...
     otError SendFrame(const uint8_t *aFrame, uint16_t aLength);
 
+    /**
+     * This method waits for receiving part or all of spinel frame within specified interval.
+     *
+     * @retval OT_ERROR_NONE             Part or all of spinel frame is received.
+     * @retval OT_ERROR_RESPONSE_TIMEOUT No spinel frame is received within @p aTimeout.
+     *
+     */
+    otError WaitForFrame(struct timeval &aTimeout);
```
jwhui: Add docs for `aTimeout`.

src/posix/platform/hdlc_interface.cpp
```diff
@@ -247,6 +247,94 @@ otError HdlcInterface::Write(const uint8_t *aFrame, uint16_t aLength)
...
+void HdlcInterface::UpdateFdSet(fd_set &aReadFdSet, fd_set &aWriteFdSet, int &aMaxFd, struct timeval &aTimeout)
+{
+    FD_SET(mSockFd, &aReadFdSet);
+
+    if (aMaxFd < mSockFd)
+    {
+        aMaxFd = mSockFd;
+    }
+
+    OT_UNUSED_VARIABLE(aWriteFdSet);
```
jwhui: Prefer at top of method in own block (see [example](https://github.com/openthread/openthread/blob/2233ae150e78ed62c467fb494ac584d5ce50595b/src/core/meshcop/commissioner.cpp#L545)).

src/posix/platform/hdlc_interface.cpp
```diff
@@ -247,6 +247,94 @@ otError HdlcInterface::Write(const uint8_t *aFrame, uint16_t aLength)
...
+    }
+    else if (rval == 0)
+    {
+        ExitNow(error = OT_ERROR_RESPONSE_TIMEOUT);
+    }
+    else if (errno != EINTR)
+    {
+        DieNowWithMessage("wait response", OT_EXIT_FAILURE);
+    }
+#endif
```
jwhui: Add `// OPENTHREAD_POSIX_VIRTUAL_TIME`.

src/posix/platform/hdlc_interface.hpp
zhanglongxia: Updated

src/posix/platform/hdlc_interface.cpp
zhanglongxia: Updated

src/posix/platform/hdlc_interface.cpp
zhanglongxia: Updated

PR #4354, opened 2019-11-25
[nrf528xx] disable soft source match

KaSroka:
This causes all platforms to have source match both in utils and in platform code - which may lead to erroneous behaviour if the wrong one is included by the linker.

### DIFF
--- a/examples/platforms/nrf528xx/nrf52811/openthread-core-nrf52811-config.h
+++ b/examples/platforms/nrf528xx/nrf52811/openthread-core-nrf52811-config.h
@@ -224,6 +224,26 @@
 #define OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT 1
 #endif
 
+/**
+ * @def RADIO_CONFIG_SRC_MATCH_SHORT_ENTRY_NUM
+ *
+ * The number of short source address table entries.
+ *
+ */
+#ifndef RADIO_CONFIG_SRC_MATCH_SHORT_ENTRY_NUM
+#define RADIO_CONFIG_SRC_MATCH_SHORT_ENTRY_NUM 0
+#endif
+
+/**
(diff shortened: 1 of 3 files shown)


KaSroka:
Some of Travis checks for the POSIX platform calls `configure` directly instead of using Makefile.
Example:
```
./bootstrap
./configure --enable-ncp --enable-ftd --enable-mtd --with-examples=posix --disable-docs --disable-tests --with-vendor-extension=./src/core/common/extension_example.cpp
make -j8
```
So `OPENTHREAD_PROJECT_CORE_CONFIG_FILE` is not set properly for them - @jwhui @zhanglongxia @bukepo any idea how to solve this problem? One solution may be to pass this define as `CPPFLAGS` but it seems not elegant for me - we have to pass define and include path and we have few of those direct ./configure calls. Also when someone calls `./configure` directly - it won't build without mentioned additional flags.

KaSroka:
I have changed the approach to modify this for nRF528xx only - but I think it should be fixed in the future in a better way.

jwhui:
@KaSroka, thanks for identifying this issue and submitting this PR. While not the most elegant, using `CPPFLAGS` is what we do elsewhere (see [tests/toranj/build.sh](https://github.com/openthread/openthread/blob/af73358d3f946adf1171d2298ae1511636253948/tests/toranj/build.sh#L91)).

Do you want to update this PR to use `CPPFLAGS` where necessary? Or do you want to limit this PR to the nrf528xx platforms as an initial step?

KaSroka:
@jwhui let's merge this PR as it is as we need it for the development. I think that a better approach should be used for this particular configuration (`OPENTHREAD_PROJECT_CORE_CONFIG_FILE`) so we have something always included by default when using `./configure` directly.

PR #4357, opened 2019-11-26
[linked-list] adding `LinkedList` class

abtink:
This commit adds a linked list module which provides a template
implementation of a singly linked list. The new class is then used in
different core modules (netif (unicast/multicast addresses), UDP6
sockets/receiver, coap resources, etc). This commit also adds a
unit test `test_linked_list` for the linked list module.

### DIFF
--- a/src/core/common/linked_list.hpp
+++ b/src/core/common/linked_list.hpp
@@ -1,0 +1,317 @@
+/*
+ *  Copyright (c) 2019, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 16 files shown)


src/core/common/linked_list.hpp
```diff
@@ -0,0 +1,304 @@
...
+            Push(aEntry);
+        }
+
+        return error;
+    }
+
+    /**
+     * This method removes an entry from the linked list.
+     *
+     * @note This method does not change the removed entry @p aEntry itself (it is `const), i.e. the  entry next
```
jwhui:
```suggestion
     * @note This method does not change the removed entry @p aEntry itself (it is `const`), i.e. the  entry next
```

Issue #3268, opened 2018-11-04
OpenThread selects RLOC as a source IPv6 address


hubertmis:
I'm trying to implement a simple application using OpenThread. Node A transmits a multicast message that is received by B. After some time B transmits a response to A. I've noticed that A uses RLOC as a source address of the multicast request. Because of that, if the network topology is changed between transmission of the request and the response, the response includes outdated RLOC as destination IPv6 address. To prevent this problem nodes should use ML-EID instead. Moreover, Thread specification suggests that applications should not use RLOC as IPv6 addresses.

It looks that usage of RLOC as a source address by default is intentional: https://github.com/openthread/openthread/blob/9a1dc1eeaefe852aa3f51407741d638559e3be39/src/core/net/ip6.cpp#L1088 . I think the reason is that Thread specification requires usage of RLOC as a source address for Thread control messages.

However, I think that the source address selection algorithm should select RLOC for Thread control messages, while it should select ML-EID for application messages. Or should an application manually set the IPv6 source addresses using the `otMessageInfo` structure?

jwhui:
@hubertmis, thanks for raising this issue.

I agree that Thread control protocols should use RLOC whereas application protocols should use ML-EID for IPv6 source address.

One proposal:
1. Update Thread control protocols explicitly specify the RLOC as IPv6 source address, and
2. Update default IPv6 source address selection to prefer the ML-EID on realm-local or smaller scope multicast destinations.

Thoughts?

hubertmis:
That's one way to go. Another one would be to create a new message sub-type like `kSubTypeControl` and select appropriate source address based on the message sub-type.

PR #3284, opened 2018-11-11
[ip6] prefer ML-EID as realm-local scope source address

hubertmis:
Fixes #3268

PR #4362, opened 2019-11-26
[ip6] select RLOC source if destination is RLOC

librasungirl:
#3284 was introduced to prefer MLEID as realm-local scope source to resolve issue #3268.
However when destination is RLOC,  it would likely to be the Thread Management traffic, RLOC source should be preferred.


jwhui:
@librasungirl, thanks for identifying this issue and working to resolve it.

Rather than adding additional logic, how about we simply rely on longest-prefix matching to help choose between RLOC and ML-EID? See https://github.com/openthread/openthread/compare/master...jwhui:ip6 for proposal.

PR #4364, opened 2019-11-28
[thci] add configurable device connection type for OpenThread_WpanCtl

FiveDimensions:


librasungirl:
LGTM overall. However may I ask why so many style changes in this PR? @FiveDimensions

tools/harness-thci/OpenThread_WpanCtl.py
```diff
@@ -88,22 +71,24 @@ def __init__(self, **kwargs):
...
-            }
+            self.logStatus = {'stop': 'stop', 'running': 'running', 'pauseReq': 'pauseReq', 'paused': 'paused'}
             self.logThreadStatus = self.logStatus['stop']
-            self.connectType = (
-                (kwargs.get('Param5')).strip().lower()
-                if kwargs.get('Param5') is not None
-                else 'usb'
+            self.connectType = (kwargs.get('Param5')).strip().lower() if kwargs.get('Param5') is not None else 'usb'
+            (self.prompt, self.wpan_cmd_prefix, self.wpan_interface) = (
+                kwargs.get('Param8').strip().split(',') if kwargs.get('Param8') else ['#', 'wpanctl', 'wpan0']
```
simonlingoogle: nit: parameters like `Param8`, `Param5` are not very self-explanatory

tools/harness-thci/OpenThread_WpanCtl.py
```diff
@@ -88,22 +71,24 @@ def __init__(self, **kwargs):
...
+            self.wpan_cmd_prefix += ' '
+            self.precmd = (kwargs.get('Param9')).strip().split(',') if kwargs.get('Param9') else []
             if self.connectType == 'ip':
                 self.dutIpv4 = kwargs.get('TelnetIP')
                 self.dutPort = kwargs.get('TelnetPort')
                 self.port = self.dutIpv4 + ':' + self.dutPort
+                self.username = kwargs.get('Param6').strip() if kwargs.get('Param6') else None
+                self.password = kwargs.get('Param7').strip() if kwargs.get('Param7') else None
+                if not self.password:
+                    self.auth_password = False
```
simonlingoogle: nit: `self.auth_password` seems redundant ?
Why not just use `if not self.password:` instead of `if not self.auth_password:`

tools/harness-thci/OpenThread_WpanCtl.py
```diff
@@ -216,10 +201,7 @@ def __sendCommand(self, cmd):
         logging.info('%s: sendCommand[%s]', self.port, cmd)
         if self.logThreadStatus == self.logStatus['running']:
             self.logThreadStatus = self.logStatus['pauseReq']
-            while (
-                self.logThreadStatus != self.logStatus['paused']
-                and self.logThreadStatus != self.logStatus['stop']
-            ):
+            while self.logThreadStatus != self.logStatus['paused'] and self.logThreadStatus != self.logStatus['stop']:
```
simonlingoogle: nit: `while self.logThreadStatus not in (self.logStatus['paused'], self.logStatus['stop']):` is more consise

tools/harness-thci/OpenThread_WpanCtl.py
```diff
@@ -297,23 +259,17 @@ def __sendCommand(self, cmd):
                     logging.info('%s: the read line is[%s]', self.port, line)
                     if line:
                         response.append(line)
-                        if re.match(WPAN_CARRIER_PROMPT, line):
+                        if re.match(self.prompt, line):
```
simonlingoogle: nit: Is `self.prompt` a regular expression? If not, we should use `if line.startswith(self.prompt):`

FiveDimensions:
@librasungirl  I used "python3 -m black -l 119 -S " to format the file, is it correct?

tools/harness-thci/OpenThread_WpanCtl.py
FiveDimensions: After using "python3 -m black -l 119 -S ", the format changed

tools/harness-thci/OpenThread_WpanCtl.py
FiveDimensions: I am Ok with your suggestion. It is not really a re, but it supports part of the whole prompt like using "#" instead of "xxxx@xxx#".  @simonlingoogle @librasungirl @bukepo thoughts?

tools/harness-thci/OpenThread_WpanCtl.py
FiveDimensions: These are extra parameters for TH analyzing deviceInputFields.xml.

tools/harness-thci/OpenThread_WpanCtl.py
simonlingoogle: I don't think it supports  part of the whole prompt like using "#" instead of "xxxx@xxx#".
```python
>>> re.match('#', 'xxxx@xxx#') is None
True
```
re.match() checks for a match only at the beginning of the string. so the current code is effective same as `line.startswith(self.prompt)` when there is no special chars in prompt.

FiveDimensions: Yes, you are right, I mixed up with re.search.

librasungirl:
For code style update, better to use a separate commit, thus the reviewer can focus on the major changes for the new features you want to introduce, which would help ease the review process. 

Make sense?

PR #4368, opened 2019-11-30
[mle] MTD prioritizes ML-EID registration

librasungirl:
This commit ensures that the important MlEid would be the
first one in Address Registration TLV so that the parent
would always be able to store it.

(SPEC-899: A recipient MUST process address entries in an
Address Registration TLV sequentially from first entry to
last entry)

### DIFF
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -1342,7 +1342,7 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
     tlv.SetType(Tlv::kAddressRegistration);
     SuccessOrExit(error = aMessage.Append(&tlv, sizeof(tlv)));
 
-    // Prioritize MlEid
+    // Prioritize ML-EID
     entry.SetContextId(kMeshLocalPrefixContextId);
     entry.SetIid(GetMeshLocal64().GetIid());
     SuccessOrExit(error = aMessage.Append(&entry, entry.GetLength()));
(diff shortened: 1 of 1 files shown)


src/core/thread/mle.cpp
```diff
@@ -1333,35 +1333,58 @@ bool Mle::HasUnregisteredAddress(void)
...
+    uint8_t                         length            = 0;
+    uint8_t                         counter           = 0;
+    uint16_t                        startOffset       = aMessage.GetLength();
+    const Ip6::NetifUnicastAddress *next              = NULL;
 
     tlv.SetType(Tlv::kAddressRegistration);
     SuccessOrExit(error = aMessage.Append(&tlv, sizeof(tlv)));
 
-    for (const Ip6::NetifUnicastAddress *addr = Get<ThreadNetif>().GetUnicastAddresses(); addr; addr = addr->GetNext())
+    for (const Ip6::NetifUnicastAddress *addr = Get<ThreadNetif>().GetUnicastAddresses(); addr; addr = next)
```
abtink: How about before the `for` loop over all unicast addresses, we add an entry for `GetMeshLocal64()` address to the message and then go through the loop of all address (now skipping over the address matching ML64 ).

 I think this will also help simplify the case `mMode == kAppendMeshLocalOnly`.

Thoughts?

librasungirl: Make sense. Updated. Thanks.

src/core/thread/mle.cpp
```diff
@@ -1353,15 +1362,10 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
...
+        if (addr->GetAddress() == GetMeshLocal64())
         {
-            if (addr->GetAddress() != GetMeshLocal64())
-            {
-                continue;
-            }
-
-            // Set `done` to `true` to exit after the address is appended.
-            done = true;
+            // skip MlEid which was appended already.
```
jwhui: nit: should we combine with if-statement above?

src/core/thread/mle.cpp
```diff
@@ -1337,14 +1337,23 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
...
     Lowpan::Context          context;
-    bool                     done        = false;
     uint8_t                  length      = 0;
     uint8_t                  counter     = 0;
     uint16_t                 startOffset = aMessage.GetLength();
 
     tlv.SetType(Tlv::kAddressRegistration);
     SuccessOrExit(error = aMessage.Append(&tlv, sizeof(tlv)));
 
+    // Prioritize MlEid
```
jwhui:
```suggestion
    // Prioritize ML-EID
```

PR #3089, opened 2018-09-19
[ip6] add message info initialize API

bukepo:
Reported in #3077

PR #3077, opened 2018-09-16
[ip6] allow hop limit 0

bukepo:
Hop limit 0 is valid value if sent to the node itself. This PR treat 0 as a valid hop limit instead of replacing it with the `kDefaultHopLimit`. And `kDefaultHopLimit` is applied when creating the message info.

PR #4370, opened 2019-12-01
[ip6] allow hoplimit 0

librasungirl:
HopLimit 0 is required for TC defined in Thread 1.2 Test Plan. And also as pointed out by @bukepo in #3089, 0 is a valid value.
> From section IPV6_UNICAST_HOPS and IPV6_MULTICAST_HOPS of http://man7.org/linux/man-pages/man7/ipv6.7.html, 0 is a valid value, and -1 means using the default value

Here were some relative PRs and discussions regarding hop limit 0 (#3077, #3089, #3092).

The first commit in PR introduces one flag to indicate whether to apply hop limit 0 to meet the hoplimit 0 requirement, as well as keep previous usage as is. 

The second commit extends existing `ping` command to allow hoplimit configuration.

### DIFF
--- a/include/openthread/ip6.h
+++ b/include/openthread/ip6.h
@@ -139,13 +139,16 @@ typedef struct otSockAddr
  */
 typedef struct otMessageInfo
 {
-    otIp6Address mSockAddr;            ///< The local IPv6 address.
-    otIp6Address mPeerAddr;            ///< The peer IPv6 address.
-    uint16_t     mSockPort;            ///< The local transport-layer port.
-    uint16_t     mPeerPort;            ///< The peer transport-layer port.
-    const void * mLinkInfo;            ///< A pointer to link-specific information.
-    uint8_t      mHopLimit;            ///< The IPv6 hop limit.
-    bool         mIsHostInterface : 1; ///< TRUE if packets sent/received via host interface, FALSE otherwise.
+    otIp6Address mSockAddr;      ///< The local IPv6 address.
+    otIp6Address mPeerAddr;      ///< The peer IPv6 address.
+    uint16_t     mSockPort;      ///< The local transport-layer port.
+    uint16_t     mPeerPort;      ///< The peer transport-layer port.
+    const void * mLinkInfo;      ///< A pointer to link-specific information.
+    uint8_t      mHopLimit;      ///< The IPv6 Hop Limit value. Only applies if `mAllowZeroHopLimit` is FALSE.
+                                 ///< If `0`, IPv6 Hop Limit is default value `OPENTHREAD_CONFIG_IP6_HOP_LIMIT_DEFAULT`.
+                                 ///< Otherwise, specifies the IPv6 Hop Limit.
+    bool mIsHostInterface : 1;   ///< TRUE if packets sent/received via host interface, FALSE otherwise.
+    bool mAllowZeroHopLimit : 1; ///< TRUE to allow IPv6 Hop Limit 0 in `mHopLimit`, FALSE otherwise.
 } otMessageInfo;
 
 /**
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -2033,6 +2033,7 @@ void Interpreter::ProcessPing(int argc, char *argv[])
 
     VerifyOrExit(!mPingTimer.IsRunning(), error = OT_ERROR_BUSY);
 
+    mMessageInfo = Ip6::MessageInfo();
     SuccessOrExit(error = mMessageInfo.GetPeerAddr().FromString(argv[0]));
 
     mLength   = 8;
@@ -2057,6 +2058,16 @@ void Interpreter::ProcessPing(int argc, char *argv[])
             SuccessOrExit(error = ParsePingInterval(argv[index], interval));
             VerifyOrExit(0 < interval && interval <= Timer::kMaxDelay, error = OT_ERROR_INVALID_ARGS);
             mInterval = interval;
+            break;
+
+        case 4:
+            SuccessOrExit(error = ParseLong(argv[index], value));
+            VerifyOrExit(0 <= value && value <= 255, error = OT_ERROR_INVALID_ARGS);
+            mMessageInfo.mHopLimit = static_cast<uint8_t>(value);
+            if (value == 0)
+            {
+                mMessageInfo.mAllowZeroHopLimit = true;
+            }
             break;
 
         default:
--- a/src/core/net/ip6.cpp
+++ b/src/core/net/ip6.cpp
@@ -460,7 +460,15 @@ otError Ip6::SendDatagram(Message &aMessage, MessageInfo &aMessageInfo, IpProto 
     header.SetDscp(PriorityToDscp(aMessage.GetPriority()));
     header.SetPayloadLength(payloadLength);
     header.SetNextHeader(aIpProto);
-    header.SetHopLimit(aMessageInfo.mHopLimit ? aMessageInfo.mHopLimit : static_cast<uint8_t>(kDefaultHopLimit));
+
+    if (aMessageInfo.GetHopLimit() != 0 || aMessageInfo.ShouldAllowZeroHopLimit())
+    {
+        header.SetHopLimit(aMessageInfo.GetHopLimit());
+    }
+    else
+    {
+        header.SetHopLimit(static_cast<uint8_t>(kDefaultHopLimit));
+    }
 
     if (aMessageInfo.GetSockAddr().IsUnspecified() || aMessageInfo.GetSockAddr().IsMulticast())
     {
--- a/src/core/net/socket.hpp
+++ b/src/core/net/socket.hpp
@@ -183,6 +183,15 @@ public:
     bool IsHostInterface(void) const { return mIsHostInterface; }
 
     /**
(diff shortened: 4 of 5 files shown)


include/openthread/ip6.h
```diff
@@ -139,13 +139,17 @@ typedef struct otSockAddr
...
+    otIp6Address mPeerAddr;    ///< The peer IPv6 address.
+    uint16_t     mSockPort;    ///< The local transport-layer port.
+    uint16_t     mPeerPort;    ///< The peer transport-layer port.
+    const void * mLinkInfo;    ///< A pointer to link-specific information.
+    uint8_t      mHopLimit;    ///< The IPv6 hop limit.
+                               ///< For the default value `0`, `OPENTHREAD_CONFIG_IP6_HOP_LIMIT_DEFAULT` would be used
+                               ///< when passing down to IP layer if `mApplyHopLimit0` is FALSE by default;
+                               ///< Otherwise,  hop limit 0 is applied in IP layer if `mApplyHopLimit0` is TRUE.
+    bool mIsHostInterface : 1; ///< TRUE if packets sent/received via host interface, FALSE otherwise.
+    bool mApplyHopLimit0 : 1;  ///< TRUE if hop limit 0 is expected to take effect in IP layer.
```
bukepo: I'm fine with this design, but I prefer an initializing function for this structure.

abtink: Agree with @bukepo. An init sounds better to me.
How about changing `mHopLimit` to be `int16_t` type instead and then use `-1` (or any negative value) as use the default `OPENTHREAD_CONFIG_IP6_HOP_LIMIT_DEFAULT` in that case.

jwhui: One concern with using init function is that this breaks existing application/user code that relies on `memset()`.

LuDuda: I agree that `mApplyHopLimit0` looks like a hack, at the same time as mentioned in the previous PR/Issues the requirement to set `mHopLimit` will cause a lot of issues for OT users and existing applications (though we can argue that due to incorrect usage of API). I was leaning towards reverting Init function before due to not having a clear use case of using Hop Limit 0, which i understand we have now.

Personally then I think we should either:

1. Use @bukepo approach
2. Use a little bit different variant of @librasungirl approach:
- add `mOverrideDefaultHopLimit` variable which has to be set in order to make use of `mHopLimit`
- use `mHopLimit` value only when `mOverrideDefaultHopLimit` variable is set

src/cli/cli.cpp
```diff
@@ -2034,11 +2034,13 @@ void Interpreter::ProcessPing(int argc, char *argv[])
...
+    mMessageInfo = Ip6::MessageInfo();
     SuccessOrExit(error = mMessageInfo.GetPeerAddr().FromString(argv[0]));
 
-    mLength   = 8;
-    mCount    = 1;
-    mInterval = 1000;
+    mLength                = 8;
+    mCount                 = 1;
+    mInterval              = 1000;
+    mMessageInfo.mHopLimit = OPENTHREAD_CONFIG_IP6_HOP_LIMIT_DEFAULT;
```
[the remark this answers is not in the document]
librasungirl: not necessary indeed

include/openthread/ip6.h
librasungirl: Both Init() solution by @bukepo and option `2` mentioned by @LuDuda would impact developer's experience, and some existing sample code need to be revised accordingly as well. 

I don't know, but if it is agreed to allow such impact, I am fine with @bukepo 's solution.

jwhui: I think the primary consideration is whether we continue to support initialization using `memset()`. FWIW, the OpenThread APIs generally support struct initialization using `memset()`. For that reason, I would lean towards @LuDuda's suggestion of having a separate flag that indicates whether `mHopLimit` contains a valid value. Thoughts?

abtink: I also favor the @LuDuda approach (adding `mOverrideDefaultHopLimit` which would be by default false when using `memset`).

I share the concern of @librasungirl , that such a change would impact public APIs/behavior and existing code need to be revised. 
One suggestion to make it very clear that definition/behavior is changed: We can consider renaming the `mHopLimit` name to `mUserHopLimit` (or perhaps some other name). This way, existing code (still using `mHopLimit`) would break at compile time and developers/users would detect this and revise the use.

Thoughts?

LuDuda: +1 @abtink. I think that even if we break API, this will not affect current applications much since anyway it's rare to change the hop limit value.

I am also fine with changing the name of the `mOverrideDefaultHopLimit` flag.

librasungirl: @abtink The internal [`MessageInfo`](https://github.com/openthread/openthread/blob/master/src/core/net/socket.hpp#L55) inherits from `otMessageInfo`, do you mean all `mHopLimit `(APIs/core) changes to `mUserHopLImit`? Or introduce one more field `mHopLimit `in internal 'MessageInfo' to be distinguished from m`UserHopLimit`?

Agree with @LuDuda  that the proposal in this PR seems a little hack. Compared with 'mOverrideDefaultHopLimit ' option,  here I want to defend the proposal in this PR a bit:

As @LuDuda mentioned, it might be rare to change the hop limit value for the developer. 
And it would be **even more rare to use hop limit 0** for the developer (so far, only required in one test defined in 1.2 Test Plan), one advantage of this PR is that it won't impact user's experience in specifying a non-zero hoplimit.

If generally hoplimit 0 won't be used by developer (which should be true generally I guess), I would prefer this solution in this PR personally. However I am open to change if the other options are preferred as a more general solution.

Thoughts?

abtink: @librasungirl my suggestion was to rename the `mHopLimit` in `otMessageInfo` (which would then require us to change its interna/corel use as well (which I think should be mostly abstracted behind the getter we provide from `MessageInfo::GetHopLimit()`).

Thinking more about this, I agree with @librasungirl approach and argument, that the current proposal/solution in the PR has some clear advantages (it keeps the behavior same for any current user of APIs, and it works with `memset` of struct, while adding the new capability to use hop limit value of zero (which itself should not be commonly used)). I am fine with moving on with this solution.

src/core/net/socket.hpp
```diff
@@ -174,6 +174,15 @@ class MessageInfo : public otMessageInfo
...
     bool IsHostInterface(void) const { return mIsHostInterface; }
 
+    /**
+     * This method indicates whether or not to apply hop limit 0.
+     *
+     * @retval TRUE  if applying hop limit 0 when `mHopLimit` field is 0.
+     * @retval FALSE if applying default `OPENTHREAD_CONFIG_IP6_HOP_LIMIT_DEFAULT` when `mHopLimit` field is 0.
+     *
+     */
+    bool DoesApplyHopLimit0(void) const { return mApplyHopLimit0; }
```
abtink: Suggestion: how about `ShouldApplyZeroHopLimit`?

include/openthread/ip6.h
librasungirl: @abtink thanks for your clarification.

@LuDuda @jwhui thoughts?

librasungirl:
Refined this PR according to current discussions/comments.

Please have another review. @LuDuda @abtink @jwhui Thanks.

LuDuda:
Thank you @librasungirl for that!

src/core/net/ip6.cpp
```diff
@@ -460,7 +460,15 @@ otError Ip6::SendDatagram(Message &aMessage, MessageInfo &aMessageInfo, IpProto
     header.SetDscp(PriorityToDscp(aMessage.GetPriority()));
     header.SetPayloadLength(payloadLength);
     header.SetNextHeader(aIpProto);
-    header.SetHopLimit(aMessageInfo.mHopLimit ? aMessageInfo.mHopLimit : static_cast<uint8_t>(kDefaultHopLimit));
+
+    if (aMessageInfo.GetHopLimit() != 0 || aMessageInfo.ShouldApplyZeroHopLimit())
+    {
+        header.SetHopLimit(aMessageInfo.GetHopLimit());
```
jwhui: If `ShouldApplyZeroHopLimit()` is `true`, should we force IPv6 Hop Limit to 0 regardless of `GetHopLimit()` value?

librasungirl: @jwhui To serve the purpose you mentioned, `ShouldApplyZeroHopLimit()` should be updated to `ShouldApplyUserHopLimit()`, which would require the user to not only specify the mHopLimit but also set the flag (something like mApplyUserHopLimit)

However the code here is by design(as agreed with @abtink with least impact on user experience since hoplimit 0 is unlikely to be used generally) , with the `||` only handle the special  hoplimit value 0 regarding whether to apply 0 (if  `ShouldApplyZeroHopLimit() `is true), otherwise use the default `kDefaultHopLimit` as previously.

jwhui: With the current logic, if `ShouldApplyZeroHopLimit()` is `true`, the IPv6 Hop Limit is set to `GetHopLimit()`. My question is whether the IPv6 Hop Limit should be set to 0 whenever `ShouldApplyZeroHopLImit()` is `true`.

librasungirl: Updated as your suggestion to avoid confusion. Thanks.

PR #4374, opened 2019-12-03
[cmake] add OT_CONFIG to specify project core config header file

abtink:
This PR contains two commits:

**[cmake] add OT_CONFIG to specify project core config header file (#4374)**

This commit adds new parameter `OT_CONFIG` to main `CMakeLists.txt`
file. This parameter allows users to specify a project-specific core
config header file (mapped to `OPENTHREAD_PROJECT_CORE_CONFIG_FILE`)
during build configuration.
    
The default value for this parameter is "none". When not specified
(value is "none"), a platform cmake file may choose to change
`OT_COFNIG` variable to provide its own core config header file. This
commit updates platform 'posix' and 'cc2548' cmake files to provide
platform specific header files when `OT_CONFIG` is not specified by
user.

------

**[toranj] add cmake option to build script (#4374)**
    
This commit adds a `cmake` option to the `build.sh` to configure and
build OpenThread NCP/CLI for FTD/MTD/RCP modes using cmake/ninja.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,14 @@ if(NOT OT_PLATFORM IN_LIST OT_EXAMPLE_PLATFORMS)
     endif()
 endif()
 
+# OT_CONFIG allows users to specify the path to OpenThread project core
+# config header file. The default value of this parameter is empty string.
+# When not specified by user (value is ""), a platform cmake file may
+# choose to change this variable to provide its own core config header
+# file instead.
+
+set(OT_CONFIG "" CACHE STRING "OpenThread project-specific config header file chosen by user at configure time")
+
 list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_BINARY_DIR}/etc/cmake)
 list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_SOURCE_DIR}/etc/cmake)
 list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_SOURCE_DIR}/include)
@@ -69,6 +77,11 @@ if(OT_PLATFORM STREQUAL "posix-host")
 elseif(NOT OT_PLATFORM MATCHES "none")
     list(APPEND OT_PRIVATE_INCLUDES ${PROJECT_SOURCE_DIR}/examples/platforms/${OT_PLATFORM})
     add_subdirectory("${PROJECT_SOURCE_DIR}/examples/platforms/${OT_PLATFORM}")
+endif()
+
+if(OT_CONFIG)
+    list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_PROJECT_CORE_CONFIG_FILE=\"${OT_CONFIG}\"")
+    message(STATUS "Project core config: \"${OT_CONFIG}\"")
 endif()
 
 list(APPEND OT_PRIVATE_DEFINES ${OT_PLATFORM_DEFINES})
--- a/examples/platforms/cc2538/CMakeLists.txt
+++ b/examples/platforms/cc2538/CMakeLists.txt
@@ -28,12 +28,18 @@
 
 set(OT_PLATFORM_LIB "openthread-cc2538" PARENT_SCOPE)
 
+if(NOT OT_CONFIG)
+    set(OT_CONFIG "openthread-core-cc2538-config.h")
+    set(OT_CONFIG ${OT_CONFIG} PARENT_SCOPE)
+endif()
+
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -57,6 +57,14 @@ if(NOT OT_PLATFORM IN_LIST OT_EXAMPLE_PLATFORMS)
     message(FATAL_ERROR "Platform unknown: ${OT_PLATFORM}")
 endif()
 
+# OT_CONFIG allows users to specify the path to OpenThread project core
+# config header file. The default value of this parameter is "none".
+# When not specified by user (value is "none"), a platform cmake file may
+# choose to change this variable to provide its own core config header
+# file instead.
+
+set(OT_CONFIG "none" CACHE STRING "OpenThread project-specific config header file chosen by user at configure time")
```
CiaranWoodward: Personally I think I prefer "" to be the default value for variables in the case that they have not been selected. An empty string is treated as FALSE for if() statements, so slightly more intuitive.

examples/platforms/cc2538/CMakeLists.txt
```diff
@@ -28,13 +28,19 @@
 
 set(OT_PLATFORM_LIB "openthread-cc2538" PARENT_SCOPE)
 
+if(OT_CONFIG MATCHES "none")
+    set(OT_CONFIG "openthread-core-cc2538-config.h")
+    set(OT_CONFIG ${OT_CONFIG} PARENT_SCOPE)
```
CiaranWoodward: Note (not request for change): This doesn't change the CACHE variable, so that will still appear as 'none'.

examples/platforms/cc2538/CMakeLists.txt
```diff
@@ -28,13 +28,19 @@
...
+endif()
+
 list(APPEND OT_PLATFORM_DEFINES
-    "OPENTHREAD_PROJECT_CORE_CONFIG_FILE=\"openthread-core-cc2538-config.h\""
     "OPENTHREAD_CORE_CONFIG_PLATFORM_CHECK_FILE=\"openthread-core-cc2538-config-check.h\""
     "OPENTHREAD_CONFIG_NCP_UART_ENABLE=1"
 )
 set(OT_PLATFORM_DEFINES ${OT_PLATFORM_DEFINES} PARENT_SCOPE)
 
+list(APPEND OT_PLATFORM_DEFINES "OPENTHREAD_PROJECT_CORE_CONFIG_FILE=\"${OT_CONFIG}\"")
```
CiaranWoodward: Nit: Seems this isn't needed, as the root CMakeLists will set the OT_PRIVATE_DEFINES to include the config file using the OT_CONFIG that was just set in the parent scope.

CMakeLists.txt
abtink: Thanks! Great suggestion. Will change it.

examples/platforms/cc2538/CMakeLists.txt
abtink: I am still learning the cmake, but I think/guess that's the behavior we want (we just want to let platform to override this value in parent scope variable if not given by user through the cache variable).

examples/platforms/cc2538/CMakeLists.txt
abtink: Not sure if I follow. Here we are setting `OT_PLATFORM_DEFINES` but in root we set a different variable `OT_PRIVATE_DEFINES`.

examples/platforms/cc2538/CMakeLists.txt
CiaranWoodward: I agree, I think this produces more sensible behaviour than changing the cache directly.

examples/platforms/cc2538/CMakeLists.txt
CiaranWoodward: Ah, I saw the OT_PLATFORM_DEFINES added to the OT_PRIVATE_DEFINES in the higher scope, so made this comment because there is then a duplicate entry in the OT_PRIVATE_DEFINES list. I forgot the core config file was required for the platform layer, so not an issue.

jwhui:
Can we enable the cmake-based build on one of the two Travis toranj builds?

abtink:
@jwhui I like the idea (cmake builds are definitely faster) but not sure if we can do this at this point.  (If i am not mistaken) we don't have a way to enable coverage test `--enable-coverage` on cmake build and we enable coverage when tests are run on travis.

PR #4377, opened 2019-12-03
[ip6] prefer RLOC source with RLOC dest in SelectSourceAddress()

jwhui:
Prefer ML-EID for any mesh-local traffic that does not have RLOC
destination.

This addresses the issue @librasungirl identified in #4362.


jwhui:
> Can we add or plan to add a unit test / system test to check the source address selection algorithm? I have a feeling we hit this issue couple of times in the past.

@librasungirl may already have a test for this. @librasungirl, would you be able to provide a test for continuous integration?

librasungirl:
@jwhui added auto check on source selection and submitted #4380

PR #4379, opened 2019-12-04
[posix-host] add cmake support

bukepo:


jwhui:
@bukepo, thanks for submitting this.

I would like to align on CMake coding conventions. What about adopting something like [KDE's CMake Coding Style](https://community.kde.org/Policies/CMake_Coding_Style)?

bukepo:
@jwhui I updated indent and removed words inside of `else()` and `endif()`.

PR #4382, opened 2019-12-05
[mac] TxRetry histogram and maximal TxRetry expiry counter added

XPAWAN:


include/openthread/link.h
```diff
@@ -35,6 +35,7 @@
 #ifndef OPENTHREAD_LINK_H_
 #define OPENTHREAD_LINK_H_
 
+#include <config/mac.h>
```
abtink: Regarding the `#include <config/mac.h>`, in general we cannot add a config header in a public header file.

I understand it's added so that we can get `OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_DIRECT` and  `OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_INDIRECT` definitions  to use for the counter array but the current model won't work.

The `config/mac.h` contains only a default definition for the config options. User of OT can choose to override these default values from build system definition or through specifying a project specific config header file (`OPENTHREAD_PROJECT_CORE_CONFIG_FILE`).

In general, we do not include any config header file from the public OT header files (headers in `include/openthread` folder). In the case where user of OT provides its own project specific config header file, we technically don't own the header file and therefore cannot expose it from OT's public header files (note that other libs can include OT public headers to use OT APIs and they may have no access to the the user given project config header).

The approach/solution we have used so far is to keep the definitions in public OT header files generic:
- In cases where the API itself is tied to certain config feature, we keep the declaration of API and document the required feature in the API description (but then don't provide the implementation in the core). Example of this are [openthread/ip6.h slaac APIs](https://github.com/openthread/openthread/blob/bf4dae0c18dc84bb87091897d79d9b3d678fda66/include/openthread/ip6.h#L518-L575).
- For data members that are tied to some feature, we again include them and document that their use requires the conifg feature. Examples are "time sync" fields in [`otThreadLinkInfo`](https://github.com/openthread/openthread/blob/bf4dae0c18dc84bb87091897d79d9b3d678fda66/include/openthread/link.h#L56-L71), or `mFrameErrorRate` or `mMessageErrorRate` fields in [`otNeighborInfo`](https://github.com/openthread/openthread/blob/bf4dae0c18dc84bb87091897d79d9b3d678fda66/include/openthread/thread.h#L85-L109) definition.

include/openthread/link.h
```diff
@@ -216,6 +217,34 @@ typedef struct otMacCounters
...
+     */
+    uint32_t mTxDirectMaxRetryExpiry;
+
+    /**
+     * Histogram of number of retries for a single indirect packet until success
+     * [0 retry: packet count, 1 retries: packet count, 2 retries : packet count, ...,
+     * n retries : packet count (n == max retry limit)]
+     * The size of the array is OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_INDIRECT.
+     */
+    uint32_t mTxIndirectRetrySuccess[OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_INDIRECT];
```
abtink: How about if we provide APIs to get these counter info instead of array definition in `otMacCounter` to allow the size be flexible.

Something like this:
```
const uint32_t *otLinkGetTxDirectRetrySuccessCounter(uint8_t &aNumberOfEntries);
```
We get a pointer to array of counter and size of array is ouputeed from `aNumberOfEntries`.

include/openthread/link.h
```diff
@@ -216,6 +217,34 @@ typedef struct otMacCounters
      */
     uint32_t mTxRetry;
 
+    /**
+     * Histogram of number of retries for a single direct packet until success
+     * [0 retry: packet count, 1 retries: packet count, 2 retries : packet count, ...,
+     * n retries : packet count (n == max retry limit)]
+     * The size of the array is OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_DIRECT.
+     */
+    uint32_t mTxDirectRetrySuccess[OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_DIRECT];
```
abtink: We have recent APIs which allow user to change the retry numbers from at run-time `otLinkSetMaxFrameRetriesDirect()` and `otLinkSetMaxFrameRetriesIndirect()` from https://github.com/openthread/openthread/pull/4256 

`OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_DIRECT` is the default value but with the new API the user can change the retry count from default to a different and possibly larger value.

We need a solution on how to handle this. 
- Do we define an array of counters with some max length and ignore if user sets retry count larger than our array size?
- If user changes the retry count, do we reset the histogram counters?

include/openthread/link.h
```diff
@@ -216,6 +217,34 @@ typedef struct otMacCounters
...
+
+    /**
+     * The total number of unique MAC transmission packets that meet maximal retry limit for direct packets.
+     *
+     */
+    uint32_t mTxDirectMaxRetryExpiry;
+
+    /**
```
abtink: I suggest adding a config option to allow users to enable/disable the feature to track retry histogrram.
The reason for this is to avoid the RAM/code overhead on SED/MTD devices which may be resource-limited.

Thoughts?

src/ncp/ncp_base_mtd.cpp
```diff
@@ -2466,6 +2466,22 @@ template <> otError NcpBase::HandlePropertyGet<SPINEL_PROP_CNTR_ALL_MAC_COUNTERS
     SuccessOrExit(error = mEncoder.WriteUint32(counters->mTxErrCca));
     SuccessOrExit(error = mEncoder.WriteUint32(counters->mTxErrAbort));
     SuccessOrExit(error = mEncoder.WriteUint32(counters->mTxErrBusyChannel));
+    SuccessOrExit(error = mEncoder.WriteUint32(OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_DIRECT));
```
abtink: Suggest either adding a new property for the histogram (mainly since it may be optional) or adding the new feilds in a struct (to allow them to be empty when feature is not enabled and new field be added after them).
Also please update the `spinel.h` definitions/documentation for the property format change.

abtink:
Thanks @LuDuda. All sounds perfect. :)
I agree. It does make sense to include the max retry expiry ones in mac counters directly and have a new command for cli.

XPAWAN:
Thank You @abtink and @LuDuda for review.
 I have prepared a new version of KPI counters according to Your remarks.

src/core/mac/mac.cpp
```diff
@@ -1262,6 +1262,18 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
 
     case kOperationTransmitDataDirect:
         mCounters.mTxData++;
+#if OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_ENABLE
+        if (aError != OT_ERROR_NONE)
+        {
+            mCounters.mTxDirectMaxRetryExpiry++;
+        }
+        else if ((0 < mSubMac.GetTransmitRetries()) &&
```
LuDuda: Let's include also the situation where the packet was sent in the first try (without retries).

src/core/mac/mac.hpp
```diff
@@ -636,6 +666,29 @@ class Mac : public InstanceLocator
...
+        uint32_t mTxDirectRetrySuccess[OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_MAX_SIZE_COUNT_DIRECT];
+
+        /**
+         * Histogram of number of retries for a single indirect packet until success
+         * [0 retry: packet count, 1 retry: packet count, 2 retry : packet count ...
+         *  until max retry limit: packet count]
+         *
+         *  The size of the array is OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_SIZE_COUNT_INDIRECT.
+         */
+        uint32_t mTxIndirectRetrySuccess[OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_SIZE_COUNT_INDIRECT];
```
LuDuda: What if `OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_SIZE_COUNT_INDIRECT` is set to 0? I think with the new approach the table will have the size of `OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_SIZE_COUNT_INDIRECT` + 1, so we are fine.

src/core/mac/mac.hpp
```diff
@@ -636,6 +666,29 @@ class Mac : public InstanceLocator
...
+    } MacRetryHistogram;
```
LuDuda:
```suggestion
    } RetryHistogram;
```
I think that "Mac" is redundant.

src/ncp/ncp_base_mtd.cpp
```diff
@@ -2551,6 +2553,51 @@ template <> otError NcpBase::HandlePropertyGet<SPINEL_PROP_CNTR_ALL_IP_COUNTERS>
...
+    otError         error = OT_ERROR_NONE;
+    const uint32_t *histogramDirect;
+    const uint32_t *histogramIndirect;
+    uint8_t         histogramDirectEntries;
+    uint8_t         histogramIndirectEntries;
+
+    histogramDirect   = otLinkGetTxDirectRetrySuccessHistogramGet(mInstance, &histogramDirectEntries);
+    histogramIndirect = otLinkGetTxIndirectRetrySuccessHistogramGet(mInstance, &histogramIndirectEntries);
+
+    if (histogramDirectEntries)
```
LuDuda: Please use parenthesis for single if-statement:

```
if (histogramDirectEntries)
{
..
}
```

src/ncp/spinel.h
```diff
@@ -3849,6 +3852,35 @@ typedef enum
...
     SPINEL_PROP_CNTR_ALL_IP_COUNTERS = SPINEL_PROP_CNTR__BEGIN + 403,
 
+/// MAC retry histogram.
+/** Format: t(A(L))t(A(L))
+ *
+ * The contents include two structs, first one is histogram which corresponds to retries of direct transmission,
+ * second one provides the histogram of retries for indirect transmission.
+ *
+ * The first structure includes:
+ *   'L': DirectRetry[0]                   (The number of packets after 1 retry).
```
LuDuda: 1 retry -> 0 retry with the new approach

src/ncp/spinel.h
```diff
@@ -3849,6 +3852,35 @@ typedef enum
...
+ *   'L': IndirectRetry[1]                   (The number of packets after 2 retry).
+ *    ...
+ *   'L': IndirectRetry[m-1]                 (The number of packets after m retry).
+ *
+ * The size of the array is compile fixed to OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_SIZE_COUNT_INDIRECT.
+ *
+ * Writing to this property with any value would reset all MAC counters to zero.
+ *
+ */
+#if OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_ENABLE
```
LuDuda: The if-statement here is unnecessary, please remove it.

src/ncp/spinel.c
```diff
@@ -2179,6 +2179,12 @@ const char *spinel_prop_key_to_cstr(spinel_prop_key_t prop_key)
         ret = "CNTR_ALL_IP_COUNTERS";
         break;
 
+#if OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_ENABLE
```
LuDuda: The if-statement here is unnecessary, please remove it.

src/ncp/ncp_base_mtd.cpp
LuDuda: Also instead of if-statement, it's easier just to call:

assert(histogramDirect);

src/core/api/link_api.cpp
```diff
@@ -323,17 +323,18 @@ const uint32_t *otLinkGetTxDirectRetrySuccessHistogramGet(otInstance *aInstance,
...
 
+
 const uint32_t *otLinkGetTxIndirectRetrySuccessHistogramGet(otInstance *aInstance, uint8_t *aNumberOfEntries)
 {
 #if OPENTHREAD_FTD
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
+    return instance.Get<Mac::Mac>().IndirectRetrySuccessHistogramGet(aNumberOfEntries);
+#else
     (void)aInstance;
```
LuDuda: nit: prefer to use OT_UNUSED_VARIABLE(aInstance);

src/core/api/link_api.cpp
```diff
@@ -323,17 +323,18 @@ const uint32_t *otLinkGetTxDirectRetrySuccessHistogramGet(otInstance *aInstance,
     return instance.Get<Mac::Mac>().DirectRetrySuccessHistogramGet(aNumberOfEntries);
 }
 
+
```
LuDuda: nit: remove empty new line

src/ncp/spinel.h
```diff
@@ -3852,34 +3852,32 @@ typedef enum
...
+     *
+     * The second structure includes:
+     *   'L': IndirectRetry[0]                   (The number of packets after 0 retry).
+     *   'L': IndirectRetry[1]                   (The number of packets after 1 retry).
+     *    ...
+     *   'L': IndirectRetry[m]                   (The number of packets after m retry).
+     *
+     * The size of the array is compile fixed to OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_SIZE_COUNT_INDIRECT.
+     *
+     * Writing to this property with any value would reset all MAC counters to zero.
```
LuDuda: nit: Writing to this property with any value would reset MAC Retry histogram.

src/core/mac/mac.hpp
```diff
@@ -577,6 +577,38 @@ class Mac : public InstanceLocator
...
+     * @returns     A pointer to the histogram of retries (in a form of an array).
+     *              The n-th element indicates that the packet has been sent with n-th retry.
+     *
+     */
+#if OPENTHREAD_FTD
+    const uint32_t *IndirectRetrySuccessHistogramGet(uint8_t *aNumberOfEntries);
+#endif // OPENTHREAD_FTD
+
+    /**
+     * This method resets mac retry histogram.
```
LuDuda:
```suggestion
     * This method resets MAC retry histogram.
```

src/core/mac/mac.cpp
```diff
@@ -1262,6 +1262,22 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
...
+        if (aError != OT_ERROR_NONE)
+        {
+            mCounters.mTxDirectMaxRetryExpiry++;
+        }
+#if OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_ENABLE
+        else if (mSubMac.GetTransmitRetries() < OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_MAX_SIZE_COUNT_DIRECT)
+        {
+            mRetryHistogram.mTxDirectRetrySuccess[mSubMac.GetTransmitRetries()]++;
+        }
+        else
```
LuDuda: Please remove this else-statement.

src/core/mac/mac.hpp
```diff
@@ -577,6 +577,38 @@ class Mac : public InstanceLocator
...
+    /**
+     * This method returns the MAC retry histogram for indirect transmission.
+     *
+     * @param[out]  aNumberOfEntries    Size of returned histogram array.
+     *
+     * @returns     A pointer to the histogram of retries (in a form of an array).
+     *              The n-th element indicates that the packet has been sent with n-th retry.
+     *
+     */
+#if OPENTHREAD_FTD
```
LuDuda: Let's move this if-statement above the function documentation.

src/core/mac/mac.cpp
```diff
@@ -1271,6 +1287,22 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
...
+        if (aError != OT_ERROR_NONE)
+        {
+            mCounters.mTxIndirectMaxRetryExpiry++;
+        }
+#if OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_ENABLE
+        else if (mSubMac.GetTransmitRetries() < OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_SIZE_COUNT_INDIRECT)
+        {
+            mRetryHistogram.mTxIndirectRetrySuccess[mSubMac.GetTransmitRetries()]++;
+        }
+        else
```
LuDuda: Please remove this else statement.

src/core/mac/mac.cpp
```diff
@@ -1819,6 +1851,43 @@ void Mac::ResetCounters(void)
     memset(&mCounters, 0, sizeof(mCounters));
 }
 
+#if OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_ENABLE
+const uint32_t *Mac::DirectRetrySuccessHistogramGet(uint8_t *aNumberOfEntries)
+{
+    if (mMaxFrameRetriesDirect > kDefaultMaxFrameRetriesDirect)
```
LuDuda: I believe it should be minimum of OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_MAX_SIZE_COUNT_DIRECT and mMaxFrameRetriesDirect. Remember that first define is set to default number of retransmissions only by default, but user can change it.

include/openthread/link.h
```diff
@@ -955,6 +967,40 @@ uint8_t otLinkConvertRssToLinkQuality(otInstance *aInstance, int8_t aRss);
...
+const uint32_t *otLinkGetTxIndirectRetrySuccessHistogramGet(otInstance *aInstance, uint8_t *aNumberOfEntries);
+
+/**
+ * This method clears histogram statistics for direct and indirect transmissions.
+ * This function is valid when OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_ENABLE configuration is enabled.
+ *
+ * @param[in]   aInstance          A pointer to an OpenThread instance.
+ *
+ */
+void otLinkGetTxRetrySuccessHistogramClear(otInstance *aInstance);
```
LuDuda: Should be `otLinkResetTxRetrySuccessHistogram(otInstance`

XPAWAN:
Hi @jwhui and @abtink
The PR is ready for final review.

src/core/api/link_api.cpp
```diff
@@ -315,6 +315,36 @@ int8_t otLinkConvertLinkQualityToRss(otInstance *aInstance, uint8_t aLinkQuality
...
+{
+#if OPENTHREAD_FTD
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
+    return instance.Get<Mac::Mac>().IndirectRetrySuccessHistogramGet(aNumberOfEntries);
+#else
+    OT_UNUSED_VARIABLE(aInstance);
+    *aNumberOfEntries = 0;
+
+    return NULL;
```
abtink: nit: Per style guide, we aim to have a single `return` from any method/function, even when we have excluding `#if`` regions.
Suggest the following: 
```
const uint32_t *otLinkGetTxIndirectRetrySuccessHistogramGet(otInstance *aInstance, uint8_t *aNumberOfEntries)
{
    uint32_t *histogram = NULL;
#if OPENTHREAD_FTD

    histogram = instance.Get<Mac::Mac>().IndirectRetrySuccessHistogramGet(aNumberOfEntries);
#else
    *aNumberOfEntries = 0;
#endif

    return histogram;
}
```

```

File: src/core/config/mac.h [hunk 23]
```diff
@@ -83,6 +83,42 @@
...
+ */
+#ifndef OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_MAX_SIZE_COUNT_DIRECT
+#define OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_MAX_SIZE_COUNT_DIRECT \
+    (OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_DIRECT + 1)
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_MAX_SIZE_COUNT_INDIRECT
+ *
+ * The default size of MAC histogram array for indirect retry successed message transmission.
```
abtink: typo: successed -> success?

src/core/mac/mac.hpp
```diff
@@ -636,6 +668,29 @@ class Mac : public InstanceLocator
         kOperationTransmitOutOfBandFrame,
     };
 
+#if OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_ENABLE
+    typedef struct
+    {
+        /**
+         * Histogram of number of retries for a single direct packet until success
+         * [0 retry: packet count, 1 retry: packet count, 2 retry : packet count ...
+         *  until max retry limit: packet count]
```
abtink: With current implementation, the `0 retry` would include/count broadcast frames (which are always sent once by sub-mac layer). We need to be aware of this when the retry histogram is analyzed (e.g., if there are a lot of broadcast frame tx, it would make it look like a lot of frames were sent with 0 retry).

Alternative is to limit the retry histogram collection to unicast frames only (frames that require ack and can be retried).

Thoughts?

include/openthread/link.h
```diff
@@ -955,6 +967,43 @@ uint8_t otLinkConvertRssToLinkQuality(otInstance *aInstance, int8_t aRss);
...
+ *
+ * This function is valid when OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_ENABLE configuration is enabled.
+ *
+ * @param[in]   aInstance          A pointer to an OpenThread instance.
+ * @param[out]  aNumberOfEntries   A reference to where the size of returned histogram array is placed.
+ *
+ * @returns     A pointer to the histogram of retries (in a form of an array).
+ *              The n-th element indicates that the packet has been sent with n-th retry.
+ */
+const uint32_t *otLinkGetTxDirectRetrySuccessHistogram(otInstance *aInstance, uint8_t &aNumberOfEntries);
```
abtink: Public OT APIs are `extern "C"` , so we cannot use reference for them and need to keep `aNumberOfEntries` as pointer.
Inside OT core we have full c++ and prefer reference over pointer (to output from method/functions).

src/ncp/spinel.c
```diff
@@ -2600,6 +2604,10 @@ const char *spinel_capability_to_cstr(unsigned int capability)
         ret = "RADIO_COEX";
         break;
 
+    case SPINEL_CAP_MAC_RETRY_HISTOGRAM:
+        ret = "TX_RETRY_HISTOGRAM";
```
abtink: `TX_RETRY_HISTOGRAM` -> `MAC_RETRY_HISTOGRAM`?

src/core/api/link_api.cpp
```diff
@@ -315,6 +315,38 @@ int8_t otLinkConvertLinkQualityToRss(otInstance *aInstance, uint8_t aLinkQuality
...
+}
+
+const uint32_t *otLinkGetTxIndirectRetrySuccessHistogram(otInstance *aInstance, uint8_t *aNumberOfEntries)
+{
+    uint32_t *histogram = NULL;
+
+#if OPENTHREAD_FTD
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
+    histogram = (uint32_t *)instance.Get<Mac::Mac>().GetIndirectRetrySuccessHistogram(*aNumberOfEntries);
```
jwhui: Declare `histogram` as `const uint32_t *`? Then you can avoid the cast to non-const `uint32_t *` here.

src/core/mac/mac.hpp
jwhui: @XPAWAN @LuDuda, any thoughts on this?

XPAWAN: The number of broadcast packets are counted by the TxBroadcast MAC counter and included in the histogram calculation.
I have tested broadcast frames and see that they are also re-transmitted when CSMACA operation fails.
The number of broadcast frames re-transmission is configured by mMaxFrameRetriesDirect MAC parameter (The same as Unicast frames).

LuDuda: I'm okay with the proposed approach. Current TxRetry counter also counts retransmissions of broadcast frames triggered by CSMA/CA failures. So this implementation aligns with the current approach.

jwhui:
@XPAWAN, thanks for updating this PR and addressing the review comments. We recently merged #4409 to help reduce code size. Unfortunately, this PR now has conflicts. Can you help resolve the conflicts?

PR #4385, opened 2019-12-06
[posix-host] add spinel_interface to extend transport layers

zhanglongxia:
This commit abstracts the API between radio_spinel and hdlc_interface into a common interface `spinel_interface`.  Users can use this interface to add new transport layers to send/receive spinel frames.

### DIFF
--- a/src/posix/platform/hdlc_interface.cpp
+++ b/src/posix/platform/hdlc_interface.cpp
@@ -123,10 +123,10 @@ namespace ot {
 namespace PosixApp {
 
 HdlcInterface::HdlcInterface(Callbacks &aCallbacks)
-    : mCallbacks(aCallbacks)
+    : SpinelInterface()
+    , mCallbacks(aCallbacks)
     , mSockFd(-1)
-    , mRxFrameBuffer()
-    , mHdlcDecoder(mRxFrameBuffer, HandleHdlcFrame, this)
+    , mHdlcDecoder(GetRxFrameBuffer(), HandleHdlcFrame, this)
 {
 }
 
@@ -612,11 +612,11 @@ void HdlcInterface::HandleHdlcFrame(otError aError)
 {
     if (aError == OT_ERROR_NONE)
     {
-        mCallbacks.HandleReceivedFrame(*this);
+        mCallbacks.HandleReceivedFrame();
     }
     else
     {
-        mRxFrameBuffer.DiscardFrame();
+        GetRxFrameBuffer().DiscardFrame();
         otLogWarnPlat("Error decoding hdlc frame: %s", otThreadErrorToString(aError));
     }
 }
--- a/src/posix/platform/hdlc_interface.cpp
+++ b/src/posix/platform/hdlc_interface.cpp
@@ -126,7 +126,7 @@ HdlcInterface::HdlcInterface(Callbacks &aCallbacks)
     : SpinelInterface()
     , mCallbacks(aCallbacks)
     , mSockFd(-1)
-    , mHdlcDecoder(GetRxFrameBuffer(), HandleHdlcFrame, this)
+    , mHdlcDecoder(mRxFrameBuffer, HandleHdlcFrame, this)
 {
 }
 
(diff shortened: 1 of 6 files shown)


src/posix/platform/radio_spinel.hpp
```diff
@@ -513,11 +514,12 @@ class RadioSpinel : public HdlcInterface::Callbacks
...
     /**
-     *  This method processes a received Spinel frame.
+     * This method processes a received Spinel frame.
+     *
+     * The newly received frame is available in `RxFrameBuffer` from `SpinelInterface::GetRxFrameBuffer()`.
      *
-     * @param[in] aFrameBuffer The frame buffer constaining the newly received frame.
      */
-    void HandleSpinelFrame(HdlcInterface::RxFrameBuffer &aFrameBuffer);
+    void HandleReceivedFrame(void);
```
gjc13: Prefer to pass arguments from function parameters rather than implicitly passing via members.

src/posix/platform/hdlc_interface.cpp
```diff
@@ -123,10 +123,10 @@ namespace ot {
 namespace PosixApp {
 
 HdlcInterface::HdlcInterface(Callbacks &aCallbacks)
-    : mCallbacks(aCallbacks)
+    : SpinelInterface()
+    , mCallbacks(aCallbacks)
     , mSockFd(-1)
-    , mRxFrameBuffer()
-    , mHdlcDecoder(mRxFrameBuffer, HandleHdlcFrame, this)
+    , mHdlcDecoder(GetRxFrameBuffer(), HandleHdlcFrame, this)
```
gjc13: Can we move mRxFrameBuffer to protected?

src/posix/platform/radio_spinel.hpp
zhanglongxia: Are there any benefits to pass the arguments through function parameters? `SpinelInterface` already provide a function `GetRxFrameBuffer()` to up layer to get the received frames. So up layer can use this unified API to get the received frames.

src/posix/platform/hdlc_interface.cpp
zhanglongxia: Good idea. I have updated the code.

src/posix/platform/spinel_interface.hpp
```diff
@@ -0,0 +1,180 @@
...
+     * @note This method should be called before reading and sending spinel frames to the interface.
+     *
+     * @param[in]   aConfig  Parameters to be given to the device or executable.
+     *
+     * @retval OT_ERROR_NONE          The interface is initialized successfully
+     * @retval OT_ERROR_ALREADY       The interface is already initialized.
+     * @retval OT_ERROR_INVALID_ARGS  The device or executable cannot be found or failed to open/run.
+     *
+     */
+    virtual otError Init(const char *aRadioFile, const char *aRadioConfig) = 0;
```
jwhui: Missing doxygen for `aRadioFile` and `aRadioConfig`.

src/posix/platform/spinel_interface.hpp
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * This method waits for receiving part or all of spinel frame within specified interval.
+     *
+     * @param[in]  aTimeout  A reference to the timeout.
+     *
+     * @retval OT_ERROR_NONE             Part or all of spinel frame is received.
+     * @retval OT_ERROR_RESPONSE_TIMEOUT No spinel frame is received within @p aTimeout.
+     *
+     */
+    virtual otError WaitForFrame(struct timeval &aTimeout) = 0;
```
jwhui: Add `const` to `aTimeout`?

src/posix/platform/spinel_interface.hpp
zhanglongxia: Updated.

src/posix/platform/spinel_interface.hpp
zhanglongxia: Lower layer uses `aTimeout` as a parameter of the function `select()`. The type of the parameter is ` struct timeval *timeout` in function `select()`.  If we add `const` to `aTimeout`, the compiler will output errors.

PR #4388, opened 2019-12-07
[netif] allow addition of external multicast addresses in any state

abtink:
This PR contains three related commits.

**[linked-list] add Find() and GetTail() methods (#4388)**

The `LinkedList::Find()` method searches for a given entry in the
linked list and if found returns a pointer to the previous entry
behind it. The `LinkedList::GetTail()` returns the tail of the list
(last entry in the list). This commit also updates the unit test
`test_linked_list` to verify the behavior of newly added methods.

**[netif] allow addition of external multicast addresses in any state (#4388)**

This commit updates the Netif's handling of multicast addresses to
allow user to subscribe to (or unsubscribe from) an external multicast
address in any state. This aligns the behavior of external multicast
addresses with how external unicast addresses are handled (user can
already add/remove unicast addresses in any state).

This commit also adds checks to ensure user cannot subscribe to fixed
multicast addresses (e.g., link-local all nodes, realm-local all
routers, etc.) using `SubscribeExternalMulticast()`.

This commit also updates the implementation of methods dealing with
fixed multicast addresses to use the newly added `LinkedList` methods.

**[unit-test] add test covering netif multicast address methods (#4388)**

### DIFF
--- a/src/core/net/netif.cpp
+++ b/src/core/net/netif.cpp
@@ -119,79 +119,144 @@ exit:
     return rval;
 }
 
-void Netif::SubscribeAllNodesMulticast(void)
+otError Netif::SubscribeAllNodesMulticast(void)
 {
-    assert(mMulticastAddresses.IsEmpty());
+    otError                error = OT_ERROR_NONE;
+    NetifMulticastAddress *tail;
+    NetifMulticastAddress &linkLocalAllNodesAddress =
+        static_cast<NetifMulticastAddress &>(const_cast<otNetifMulticastAddress &>(kLinkLocalAllNodesMulticastAddress));
 
-    mMulticastAddresses.SetHead(static_cast<NetifMulticastAddress *>(
-        const_cast<otNetifMulticastAddress *>(&kLinkLocalAllNodesMulticastAddress)));
+    VerifyOrExit(!mMulticastAddresses.Contains(linkLocalAllNodesAddress), error = OT_ERROR_ALREADY);
 
-    if (mAddressCallback != NULL)
+    // Append the fixed chain of three multicast addresses to the
+    // tail of the list:
+    //
+    //    LinkLocalAll -> RealmLocalAll -> RealmLocalAllMpl.
+
+    tail = mMulticastAddresses.GetTail();
+
+    if (tail == NULL)
     {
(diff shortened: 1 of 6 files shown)


abtink:
This change should help address the failures discovered from https://github.com/openthread/wpantund/pull/444 (thanks to @gjc13 for discovering this and helping investigate the root cause of the issue).

src/core/net/netif.cpp
```diff
@@ -301,14 +365,20 @@ otError Netif::SubscribeExternalMulticast(const Address &aAddress)
...
+    NetifMulticastAddress &linkLocalAllRoutersAddress = static_cast<NetifMulticastAddress &>(
+        const_cast<otNetifMulticastAddress &>(kLinkLocalAllRoutersMulticastAddress));
 
-    VerifyOrExit(!mMulticastAddresses.IsEmpty(), error = OT_ERROR_INVALID_STATE);
+    // Check that the address is not one of the fixed addresses:
+    // LinkLocalAllRouters -> RealmLocalAllRouters -> LinkLocalAllNodes
+    // -> RealmLocalAllNodes -> RealmLocalAllMpl.
 
-    if (IsMulticastSubscribed(aAddress))
+    for (const NetifMulticastAddress *cur = &linkLocalAllRoutersAddress; cur; cur = cur->GetNext())
```
gjc13: Is it a must to check these addresses given that they will be checked in `IsMulticastSubscribed()`?
The linux kernel may subscribe ff02::2 on every interface so returning OT_ERROR_ALREADY may be more friendly.

abtink: The main change in this PR is we allow user to subscribe to multicast addresses at any point even before these addresses are added by the OpenThread core code to interface (so when calls these API, these addresses they may not yet be in the list). Returning `OT_ERROR_ALREADY` won't be correct then since the addresses are not really subscribed.

Note that OT core will add 3 addresses (link local and realm local all nodes and all MPL forwarders)  when netif is brought up. The other two router specific ones (link local and realm local all routers) are added only when device becomes a router. 

This behavior from this method is similar to how we don't let user add a unicast link local address, i.e. `AddExternalUnicastAddress()` returning `OT_ERROR_INVALID_ARGS` if the given address is link-local.

PR #4396, opened 2019-12-12
[coap] add support for RFC7641

sjlongland:
The following pull request implements support for [RFC7641 Observations](https://tools.ietf.org/html/rfc7641).

Observations was based on an earlier proof-of-concept which was tested against `node-coap`.

The implementation given here has been successfully been tested with `aiocoap-client` (with OpenThread implementing the server) and with Californium (with OpenThread implementing the client), as well as between two OpenThread nodes.

### DIFF
--- a/src/cli/README_COAP.md
+++ b/src/cli/README_COAP.md
@@ -55,12 +55,15 @@ coap response sent
 ## Command List
 
 * [help](#help)
+* [cancel](#cancel)
 * [delete](#delete-address-uri-path-type-payload)
 * [get](#get-address-uri-path-type)
+* [observe](#observe-address-uri-path-type)
 * [parameters](#parameters)
 * [post](#post-address-uri-path-type-payload)
 * [put](#put-address-uri-path-type-payload)
 * [resource](#resource-uri-path)
+* [set](#set-new-content)
 * [start](#start)
 * [stop](#stop)
 
@@ -71,18 +74,31 @@ coap response sent
 ```bash
 > coap help
 help
+cancel
 delete
 get
+observe
 parameters
 post
 put
 resource
+set
 start
 stop
 Done
 ```
 
 List the CoAP CLI commands.
+
+### cancel
+
+Request the cancellation of an existing observation subscription to a remote
+resource.
+
+```bash
+> coap cancel
+Done
+```
 
 ### delete \<address\> \<uri-path\> \[type\] \[payload\]
 
@@ -104,6 +120,20 @@ Done
 
 ```bash
 > coap get fdde:ad00:beef:0:2780:9423:166c:1aac test-resource
+Done
+```
+
+### observe \<address\> \<uri-path\> \[type\]
+
+This is the same a `get`, but the `Observe` parameter will be sent, set to 0
+triggering a subscription request.
+
+* address: IPv6 address of the CoAP server.
+* uri-path: URI path of the resource.
+* type: "con" for Confirmable or "non-con" for Non-confirmable (default).
+
+```bash
+> coap observe fdde:ad00:beef:0:2780:9423:166c:1aac test-resource
 Done
 ```
 
@@ -177,6 +207,15 @@ Sets the URI path for the test resource.
 Done
 > coap resource
 test-resource
+Done
+```
+
+### set \[new-content\]
+
+Sets the content sent by the test resource.  If a CoAP client is observing the resource, a notification is sent to that client.
+
+```bash
+> coap set Testing123
 Done
 ```
 
--- a/src/cli/cli_coap.cpp
+++ b/src/cli/cli_coap.cpp
@@ -100,6 +100,12 @@ otError Coap::CancelResourceSubscription(void)
     mRequestTokenLength = 0;
 
 exit:
+
+    if ((error != OT_ERROR_NONE) && (message != NULL))
+    {
+        otMessageFree(message);
+    }
+
     return error;
 }
 
--- a/src/cli/cli_coap.cpp
+++ b/src/cli/cli_coap.cpp
@@ -57,6 +57,10 @@ Coap::Coap(Interpreter &aInterpreter)
     : mInterpreter(aInterpreter)
     , mUseDefaultRequestTxParameters(true)
     , mUseDefaultResponseTxParameters(true)
+    , mObserveSerial(0)
+    , mRequestTokenLength(0)
+    , mSubscriberTokenLength(0)
+    , mSubscriberConfirmableNotifications(false)
 {
     memset(&mResource, 0, sizeof(mResource));
     memset(&mRequestAddr, 0, sizeof(mRequestAddr));
@@ -66,9 +70,6 @@ Coap::Coap(Interpreter &aInterpreter)
     memset(&mUriPath, 0, sizeof(mUriPath));
     memset(&mRequestUri, 0, sizeof(mRequestUri));
     strlcpy(mResourceContent, "0", kMaxBufferSize);
-    mRequestTokenLength    = 0;
-    mSubscriberTokenLength = 0;
-    mObserveSerial         = 0;
 }
 
 otError Coap::CancelResourceSubscription(void)
--- a/src/cli/cli_coap.cpp
+++ b/src/cli/cli_coap.cpp
@@ -45,33 +45,48 @@ namespace ot {
 namespace Cli {
 
 const struct Coap::Command Coap::sCommands[] = {
-    {"help", &Coap::ProcessHelp},         {"cancel", &Coap::ProcessCancel},
-    {"delete", &Coap::ProcessRequest},    {"get", &Coap::ProcessRequest},
-    {"observe", &Coap::ProcessRequest},   {"parameters", &Coap::ProcessParameters},
-    {"post", &Coap::ProcessRequest},      {"put", &Coap::ProcessRequest},
-    {"resource", &Coap::ProcessResource}, {"set", &Coap::ProcessSet},
-    {"start", &Coap::ProcessStart},       {"stop", &Coap::ProcessStop},
+    {"help", &Coap::ProcessHelp},
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
+    {"cancel", &Coap::ProcessCancel},
+#endif
+    {"delete", &Coap::ProcessRequest},
+    {"get", &Coap::ProcessRequest},
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
+    {"observe", &Coap::ProcessRequest},
+#endif
+    {"parameters", &Coap::ProcessParameters},
+    {"post", &Coap::ProcessRequest},
+    {"put", &Coap::ProcessRequest},
+    {"resource", &Coap::ProcessResource},
+    {"set", &Coap::ProcessSet},
+    {"start", &Coap::ProcessStart},
+    {"stop", &Coap::ProcessStop},
 };
 
 Coap::Coap(Interpreter &aInterpreter)
     : mInterpreter(aInterpreter)
     , mUseDefaultRequestTxParameters(true)
     , mUseDefaultResponseTxParameters(true)
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
     , mObserveSerial(0)
     , mRequestTokenLength(0)
     , mSubscriberTokenLength(0)
     , mSubscriberConfirmableNotifications(false)
+#endif
 {
     memset(&mResource, 0, sizeof(mResource));
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
     memset(&mRequestAddr, 0, sizeof(mRequestAddr));
     memset(&mSubscriberSock, 0, sizeof(mSubscriberSock));
     memset(&mRequestToken, 0, sizeof(mRequestToken));
     memset(&mSubscriberToken, 0, sizeof(mSubscriberToken));
-    memset(&mUriPath, 0, sizeof(mUriPath));
     memset(&mRequestUri, 0, sizeof(mRequestUri));
+#endif
+    memset(&mUriPath, 0, sizeof(mUriPath));
     strncpy(mResourceContent, "0", sizeof(mResourceContent) - 1);
 }
 
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
 otError Coap::CancelResourceSubscription(void)
 {
     otError       error   = OT_ERROR_NONE;
@@ -114,6 +129,7 @@ void Coap::CancelSubscriber(void)
     memset(&mSubscriberSock, 0, sizeof(mSubscriberSock));
     mSubscriberTokenLength = 0;
 }
+#endif // OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
 
 void Coap::PrintPayload(otMessage *aMessage) const
 {
@@ -141,6 +157,7 @@ void Coap::PrintPayload(otMessage *aMessage) const
     mInterpreter.mServer->OutputFormat("\r\n");
 }
 
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
 otError Coap::ProcessCancel(int argc, char *argv[])
 {
     OT_UNUSED_VARIABLE(argc);
@@ -148,6 +165,7 @@ otError Coap::ProcessCancel(int argc, char *argv[])
 
     return CancelResourceSubscription();
 }
+#endif
 
 otError Coap::ProcessHelp(int argc, char *argv[])
 {
@@ -188,15 +206,18 @@ exit:
 
 otError Coap::ProcessSet(int argc, char *argv[])
 {
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
     otMessage *   notificationMessage = NULL;
     otMessageInfo messageInfo;
-    otError       error = OT_ERROR_NONE;
+#endif
+    otError error = OT_ERROR_NONE;
 
     if (argc > 1)
     {
         VerifyOrExit(strlen(argv[1]) < (kMaxBufferSize - 1), error = OT_ERROR_INVALID_ARGS);
         strncpy(mResourceContent, argv[1], sizeof(mResourceContent) - 1);
 
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
         if (mSubscriberTokenLength > 0)
         {
             // Notify the subscriber
@@ -225,6 +246,7 @@ otError Coap::ProcessSet(int argc, char *argv[])
             SuccessOrExit(error = otCoapSendRequest(mInterpreter.mInstance, notificationMessage, &messageInfo,
                                                     &Coap::HandleNotificationResponse, this));
         }
+#endif // OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
     }
     else
     {
@@ -233,10 +255,12 @@ otError Coap::ProcessSet(int argc, char *argv[])
 
 exit:
 
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
     if ((error != OT_ERROR_NONE) && (notificationMessage != NULL))
     {
         otMessageFree(notificationMessage);
     }
+#endif
 
     return error;
 }
@@ -344,7 +368,9 @@ otError Coap::ProcessRequest(int argc, char *argv[])
     otCoapType   coapType               = OT_COAP_TYPE_NON_CONFIRMABLE;
     otCoapCode   coapCode               = OT_COAP_CODE_GET;
     otIp6Address coapDestinationIp;
-    bool         coapObserve = false;
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
+    bool coapObserve = false;
+#endif
 
     VerifyOrExit(argc > 0, error = OT_ERROR_INVALID_ARGS);
 
@@ -353,12 +379,14 @@ otError Coap::ProcessRequest(int argc, char *argv[])
     {
         coapCode = OT_COAP_CODE_GET;
     }
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
     else if (strcmp(argv[0], "observe") == 0)
     {
         // Observe request.  This is a GET with Observe=0
         coapCode    = OT_COAP_CODE_GET;
         coapObserve = true;
     }
+#endif
     else if (strcmp(argv[0], "post") == 0)
     {
         coapCode = OT_COAP_CODE_POST;
@@ -406,11 +434,13 @@ otError Coap::ProcessRequest(int argc, char *argv[])
         }
     }
 
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
     if (coapObserve && mRequestTokenLength)
     {
         // New observe request, cancel any existing observation
         SuccessOrExit(error = CancelResourceSubscription());
     }
+#endif
 
     message = otCoapNewMessage(mInterpreter.mInstance, NULL);
     VerifyOrExit(message != NULL, error = OT_ERROR_NO_BUFS);
@@ -418,10 +448,12 @@ otError Coap::ProcessRequest(int argc, char *argv[])
     otCoapMessageInit(message, coapType, coapCode);
     otCoapMessageGenerateToken(message, ot::Coap::Message::kDefaultTokenLength);
 
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
     if (coapObserve)
     {
         SuccessOrExit(error = otCoapMessageAppendObserveOption(message, 0));
     }
+#endif
 
     SuccessOrExit(error = otCoapMessageAppendUriPathOptions(message, coapUri));
 
@@ -445,6 +477,7 @@ otError Coap::ProcessRequest(int argc, char *argv[])
     messageInfo.mPeerAddr = coapDestinationIp;
     messageInfo.mPeerPort = OT_DEFAULT_COAP_PORT;
 
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
     if (coapObserve)
     {
         // Make a note of the message details for later so we can cancel it later.
@@ -453,6 +486,7 @@ otError Coap::ProcessRequest(int argc, char *argv[])
         memcpy(mRequestToken, otCoapMessageGetToken(message), mRequestTokenLength);
         strncpy(mRequestUri, coapUri, sizeof(mRequestUri) - 1);
     }
+#endif
 
     if ((coapType == OT_COAP_TYPE_CONFIRMABLE) || (coapCode == OT_COAP_CODE_GET))
     {
@@ -506,11 +540,13 @@ void Coap::HandleRequest(void *aContext, otMessage *aMessage, const otMessageInf
 
 void Coap::HandleRequest(otMessage *aMessage, const otMessageInfo *aMessageInfo)
 {
-    otError              error           = OT_ERROR_NONE;
-    otMessage *          responseMessage = NULL;
-    otCoapCode           responseCode    = OT_COAP_CODE_EMPTY;
-    uint64_t             observe         = 0;
-    bool                 observePresent  = false;
+    otError    error           = OT_ERROR_NONE;
+    otMessage *responseMessage = NULL;
+    otCoapCode responseCode    = OT_COAP_CODE_EMPTY;
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
+    uint64_t observe        = 0;
+    bool     observePresent = false;
+#endif
     otCoapOptionIterator iterator;
 
     mInterpreter.mServer->OutputFormat("coap request from ");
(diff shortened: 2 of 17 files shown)


sjlongland:
```
+curl -s https://openthread-size-report.glitch.me/size-report/1354027/bash
+chmod a+x size-report
+./size-report init OpenThread
./size-report: line 1: html: No such file or directory
./size-report: line 2: syntax error near unexpected token `<'
'
The command ".travis/check-size" exited with 2.
```
That smells awfully like a shell script got swapped with HTML for reasons unknown.

jwhui:
> ```
> +curl -s https://openthread-size-report.glitch.me/size-report/1354027/bash
> +chmod a+x size-report
> +./size-report init OpenThread
> ./size-report: line 1: html: No such file or directory
> ./size-report: line 2: syntax error near unexpected token `<'
> '
> The command ".travis/check-size" exited with 2.
> ```
> 
> That smells awfully like a shell script got swapped with HTML for reasons unknown.

Appears to be a momentary issue. I kicked the build again and it successfully completed.

src/cli/README_COAP.md
```diff
@@ -105,6 +121,20 @@ Done
 Done
 ```
 
+### obs \<address\> \<uri-path\> \[type\]
```
bukepo: shall we use `observe` for readability?

src/core/coap/coap_message.hpp
```diff
@@ -379,18 +379,42 @@ class Message : public ot::Message
...
      *
      */
     const otCoapOption *GetFirstOption(void);
 
+    /**
+     * This method returns a pointer to the next option matching the given option number.
+     *
+     * @param[in]   aOption         Option number to look for.
+     */
+    const otCoapOption *GetNextOptionMatching(otCoapOptionType aOption);
```
bukepo: Maybe we should Remove `Get` from `GetFirstOption()`, `GetNextOption()`, so that it's clear these methods are not ordinary getters. Thoughts? @jwhui

src/cli/README_COAP.md
sjlongland: We certainly can do that… I was thinking along the lines of keeping strings short but 4 bytes is hardly a big expense.

src/core/coap/coap_message.hpp
sjlongland: Another idea, if we're concerned about the state of the `otCoapMessage` object changing as options are iterated over… we have a routine that initialises a `otCoapOptionIterator` which contains the option iterator state and can scan back and forth without changing the state of the message object itself.  That iterator is then private to the caller manipulating it.

sjlongland:
```
/Users/travis/build/openthread/openthread/examples/../src/cli/cli_coap.cpp:211:90: error: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'uint16_t' (aka 'unsigned short') [-Werror,-Wconversion]

            SuccessOrExit(error = otMessageAppend(notificationMessage, mResourceContent, strlen(mResourceContent)));
```

Damn, so close.  I think I got that nailed this time.  Seems the MacOS X compiler toolchain really does want developers to be explicit on these things!

Is there a way of running these tests locally?  I do have a MacBook laying around, but the OS and compiler on there would be very long in the tooth now, so not representative of what's being used by Travis CI.

sjlongland:
I just realised this potentially overlaps with https://github.com/openthread/openthread/pull/4247.

That implementation of RFC7959 at least tries to actually implement the block-wise transfer logic so that from the application perspective, it's a single request.  I think @AuSchell has done a nice effort there.

The downside to that is one must have sufficient RAM to transmit or receive the entire request entity, which for the use cases I use it for is problematic (512kB of CC2538 flash image is never going to fit in 32kB of RAM).  The other implementation restricts the client to whatever `OPENTHREAD_CONFIG_COAP_MAX_ASSEMBLED_BODY` is set to (4kB in the pull request).

If it's helpful, I can pull the block-wise transfer bit out into a separate pull request.

jwhui:
@sjlongland, I think it would make sense to break out the RFC 7641 and RFC 7959 changes into separate PRs.

@sjlongland @AuSchell, can you provide your thoughts on how to reconcile the changes proposed in this PR vs #4247?

sjlongland:
No problems, that's chopped out the RFC7959 part out, it's now in its own branch: https://github.com/vrtsystems/openthread/commit/21f788a0fa1af3b09ad34b363a982f7547dca786

src/core/coap/coap_message.hpp
jwhui: I think it makes sense to introduce a coap option iterator.

sjlongland: Question is, do we do that here, or should that be a separate PR to separate concerns?

sjlongland: @jwhui @bukepo 

Had a crack at an iterator in https://github.com/vrtsystems/openthread/commits/feature/coap-msg-opt-iterator

I have not tested that code beyond seeing if it compiles as the hardware needed for testing is at work and the office is closing for the year (will be open again 6th January… I'm actually home a day earlier due to smoke-haze induced illness).

If that's getting close to the mark, we can make another pull request (and re-base this one). :-)

sjlongland: I've cleaned up that code and made a separate PR: #4452.

sjlongland:
I've just re-based this branch off PR#4452… testing on a couple of nRF52840 dongles (the Fireflys I was using seem to have grown legs and walked off during the Christmas break) suggests it is working as expected.

sjlongland:
Heads up, I've just re-based this off master to pick up the merged `otCoapOptionIterator`. :-)

sjlongland:
Silly question, following my rebase of this branch on `master`, is there any other steps needed from my end?

Does the rebased version of the pull request need another quick review?

src/cli/cli_coap.cpp
```diff
@@ -45,15 +45,63 @@ namespace ot {
...
 {
     memset(&mResource, 0, sizeof(mResource));
+    memset(&mRequestAddr, 0, sizeof(mRequestAddr));
+    memset(&mSubscriberSock, 0, sizeof(mSubscriberSock));
+    memset(&mRequestToken, 0, sizeof(mRequestToken));
+    memset(&mSubscriberToken, 0, sizeof(mSubscriberToken));
+    memset(&mUriPath, 0, sizeof(mUriPath));
+    memset(&mRequestUri, 0, sizeof(mRequestUri));
+    strlcpy(mResourceContent, "0", kMaxBufferSize);
+    mRequestTokenLength    = 0;
```
jwhui: nit: prefer using initializer list

src/cli/cli_coap.cpp
```diff
@@ -119,6 +175,55 @@ otError Coap::ProcessResource(int argc, char *argv[])
...
+otError Coap::ProcessSet(int argc, char *argv[])
+{
+    otMessage *   notificationMessage = NULL;
+    otMessageInfo messageInfo;
+    otError       error = OT_ERROR_NONE;
+
+    if (argc > 1)
+    {
+        VerifyOrExit(strlen(argv[1]) < (kMaxBufferSize - 1), error = OT_ERROR_INVALID_ARGS);
+        strlcpy(mResourceContent, argv[1], kMaxBufferSize);
```
jwhui: Prefer using `strncpy` - we are trying to move away from `strlpy` (see https://github.com/openthread/openthread/commit/708c250fe5ef7057e6395d9759fb341d150e89e8).

src/cli/cli_coap.cpp
```diff
@@ -119,6 +175,55 @@ otError Coap::ProcessResource(int argc, char *argv[])
...
+                                                    &Coap::HandleNotificationResponse, this));
+        }
+    }
+    else
+    {
+        mInterpreter.mServer->OutputFormat("%s\r\n", mResourceContent);
+    }
+
+exit:
+    return OT_ERROR_NONE;
```
jwhui: Should this return `error`?

src/cli/cli_coap.cpp
```diff
@@ -119,6 +175,55 @@ otError Coap::ProcessResource(int argc, char *argv[])
...
+
+            notificationMessage = otCoapNewMessage(mInterpreter.mInstance, NULL);
+            VerifyOrExit(notificationMessage != NULL, error = OT_ERROR_NO_BUFS);
+
+            otCoapMessageInit(
+                notificationMessage,
+                ((mSubscriberConfirmableNotifications) ? OT_COAP_TYPE_CONFIRMABLE : OT_COAP_TYPE_NON_CONFIRMABLE),
+                OT_COAP_CODE_CONTENT);
+
+            SuccessOrExit(error = otCoapMessageSetToken(notificationMessage, mSubscriberToken, mSubscriberTokenLength));
```
jwhui: Free `notificationMessage` if this or any of the calls below fail?

src/cli/cli_coap.cpp
```diff
@@ -45,15 +45,63 @@ namespace ot {
...
+    messageInfo.mPeerPort = OT_DEFAULT_COAP_PORT;
+
+    VerifyOrExit(mRequestTokenLength != 0, error = OT_ERROR_INVALID_STATE);
+
+    message = otCoapNewMessage(mInterpreter.mInstance, NULL);
+    VerifyOrExit(message != NULL, error = OT_ERROR_NO_BUFS);
+
+    otCoapMessageInit(message, OT_COAP_TYPE_CONFIRMABLE, OT_COAP_CODE_GET);
+
+    SuccessOrExit(error = otCoapMessageSetToken(message, mRequestToken, mRequestTokenLength));
```
jwhui: Free `notificationMessage` if this or any of the calls below fail?

src/cli/cli_coap.cpp
```diff
@@ -231,6 +355,15 @@ otError Coap::ProcessRequest(int argc, char *argv[])
     messageInfo.mPeerAddr = coapDestinationIp;
     messageInfo.mPeerPort = OT_DEFAULT_COAP_PORT;
 
+    if (coapObserve)
+    {
+        // Make a note of the message details for later so we can cancel it later.
+        memcpy(&mRequestAddr, &coapDestinationIp, sizeof(mRequestAddr));
+        mRequestTokenLength = otCoapMessageGetTokenLength(message);
+        memcpy(mRequestToken, otCoapMessageGetToken(message), mRequestTokenLength);
+        strlcpy(mRequestUri, coapUri, kMaxUriLength);
```
jwhui: Prefer using `strncpy` - we are trying to move away from `strlpy` (see https://github.com/openthread/openthread/commit/708c250fe5ef7057e6395d9759fb341d150e89e8).

src/cli/cli_coap.cpp
```diff
@@ -318,9 +461,38 @@ void Coap::HandleRequest(otMessage *aMessage, const otMessageInfo *aMessageInfo)
...
+                if (observe == 0)
+                {
+                    // New subscriber
+                    mInterpreter.mServer->OutputFormat("Subscribing client\r\n");
+                    mSubscriberSock.mAddress = aMessageInfo->mPeerAddr;
+                    mSubscriberSock.mPort    = aMessageInfo->mPeerPort;
+                    mSubscriberTokenLength   = otCoapMessageGetTokenLength(aMessage);
+                    memcpy(mSubscriberToken, otCoapMessageGetToken(aMessage), mSubscriberTokenLength);
+
+                    mSubscriberConfirmableNotifications = (otCoapMessageGetType(aMessage) == OT_COAP_TYPE_CONFIRMABLE);
```
jwhui: It seems we are determining whether notifications are sent as confirmable based on whether the request was sent as confirmable. RFC 7641 allows a server to choose independent of the request. Given that this is example code, I am fine with keeping this behavior. If so, it might be worth adding a comment indicating this policy choice and that RFC 7641 allows more flexibility.

src/cli/cli_coap.cpp
```diff
@@ -371,12 +579,30 @@ void Coap::HandleResponse(otMessage *aMessage, const otMessageInfo *aMessageInfo
...
 
-        PrintPayload(aMessage);
+            mInterpreter.mServer->OutputFormat("coap response from ");
+            mInterpreter.OutputIp6Address(aMessageInfo->mPeerAddr);
+
+            if (observeOpt != NULL)
+            {
+                uint64_t observeVal = 0;
+                otError  error      = otCoapOptionIteratorGetOptionUintValue(&iterator, &observeVal);
+                if (error == OT_ERROR_NONE)
```
jwhui: nit: prefer empty line after declaration block

src/core/coap/coap.hpp
```diff
@@ -106,18 +106,21 @@ class CoapMetadata
...
         , mAcknowledged(false)
-        , mConfirmable(false){};
+        , mConfirmable(false)
+        , mObserve(false){};
 
     /**
      * This constructor initializes the object with specific values.
      *
      * @param[in]  aConfirmable  Information if the request is confirmable or not.
+     * @param[in]  aObserve      Information if the request/response involves observations
```
jwhui: nit: end line with period.

src/core/coap/coap_message.cpp
```diff
@@ -412,6 +412,22 @@ otError OptionIterator::Init(const Message *aMessage)
     return err;
 }
 
+const otCoapOption *OptionIterator::GetFirstOptionMatching(otCoapOptionType aOption)
+{
+    const otCoapOption *rval = NULL;
+    for (const otCoapOption *option = GetFirstOption(); option != NULL; option = GetNextOption())
```
jwhui: nit: prefer empty line after declaration block

src/core/coap/coap_message.cpp
```diff
@@ -429,6 +445,22 @@ const otCoapOption *OptionIterator::GetFirstOption(void)
     return option;
 }
 
+const otCoapOption *OptionIterator::GetNextOptionMatching(otCoapOptionType aOption)
+{
+    const otCoapOption *rval = NULL;
+    for (const otCoapOption *option = GetNextOption(); option != NULL; option = GetNextOption())
```
jwhui: nit: prefer empty line after declaration block

src/core/coap/coap_message.cpp
```diff
@@ -512,6 +544,26 @@ const otCoapOption *OptionIterator::GetNextOption(void)
     return rval;
 }
 
+otError OptionIterator::GetOptionValue(uint64_t &aValue) const
+{
+    otError error = OT_ERROR_NONE;
+    uint8_t value[sizeof(aValue)];
+    uint8_t pos = 0;
+
+    VerifyOrExit(mOption.mLength < sizeof(aValue), error = OT_ERROR_NO_BUFS);
```
jwhui: Should this be `mOption.mLength <= sizeof(aValue)`?

src/core/coap/coap_message.hpp
```diff
@@ -599,20 +599,44 @@ class OptionIterator : public ::otCoapOptionIterator
      */
     otError Init(const Message *aMessage);
 
+    /**
+     * This method returns a pointer to the first option matching the given option number.
+     *
+     * @param[in]   aOption         Option number to look for.
+     */
```
jwhui: nit: document return value

src/core/coap/coap_message.hpp
```diff
@@ -599,20 +599,44 @@ class OptionIterator : public ::otCoapOptionIterator
...
      *
      * @returns A pointer to the first option. If no option is present NULL pointer is returned.
      */
     const otCoapOption *GetFirstOption(void);
 
+    /**
+     * This method returns a pointer to the next option matching the given option number.
+     *
+     * @param[in]   aOption         Option number to look for.
+     */
```
jwhui: nit: document return value

src/core/coap/coap_message.hpp
```diff
@@ -599,20 +599,44 @@ class OptionIterator : public ::otCoapOptionIterator
...
      * This method returns a pointer to the next option.
      *
      * @returns A pointer to the next option. If no more options are present NULL pointer is returned.
      */
     const otCoapOption *GetNextOption(void);
 
+    /**
+     * This function fills current option value into @p aValue.  The option is assumed to be an unsigned integer.
+     *
+     * @retval  OT_ERROR_NONE       Successfully filled value.
```
jwhui: nit: document `aValue` parameter

src/cli/cli_coap.cpp
sjlongland: Yep, I'll fix that… there's two other places but those are probably beyond the scope of this PR.
* `src/cli/cli_coap_secure.cpp`
* `src/core/meshcop/commissioner.cpp`

sjlongland:
I noticed there was a merge conflict before due to `strlcpy`→`strncpy`, so I've re-based to resolve that.

src/cli/cli_coap.cpp
```diff
@@ -119,6 +182,61 @@ otError Coap::ProcessResource(int argc, char *argv[])
...
+        }
+    }
+    else
+    {
+        mInterpreter.mServer->OutputFormat("%s\r\n", mResourceContent);
+    }
+
+exit:
+
+    if ((error != OT_ERROR_NONE) && (message != NULL))
```
jwhui:
```suggestion
    if ((error != OT_ERROR_NONE) && (notificationMessage != NULL))
```

src/core/coap/coap_message.hpp
```diff
@@ -597,23 +597,62 @@ class OptionIterator : public ::otCoapOptionIterator
      */
     otError Init(const Message *aMessage);
 
+    /**
+     * This method returns a pointer to the first option matching the given option number.
+     * The internal option pointer is advanced until matching option is seen, if no matching
```
jwhui: nit: prefer empty line between brief and body.

src/core/coap/coap_message.hpp
```diff
@@ -597,23 +597,62 @@ class OptionIterator : public ::otCoapOptionIterator
...
     /**
      * This method returns a pointer to the first option.
      *
      * @returns A pointer to the first option. If no option is present NULL pointer is returned.
      */
     const otCoapOption *GetFirstOption(void);
 
+    /**
+     * This method returns a pointer to the next option matching the given option number.
+     * The internal option pointer is advanced until matching option is seen, if no matching
```
jwhui: nit: prefer empty line between brief and body.

src/core/coap/coap_message.hpp
```diff
@@ -597,23 +597,62 @@ class OptionIterator : public ::otCoapOptionIterator
...
 
+    /**
+     * This method returns a pointer to the first option matching the given option number.
+     * The internal option pointer is advanced until matching option is seen, if no matching
+     * option is seen, the iterator will advance to the end of the options block.
+     *
+     * @param[in]   aOption         Option number to look for.
+     *
+     * @returns A pointer to the first matching option. If no option matching @p aOption is seen, NULL pointer is
+     * returned.
```
jwhui: nit: align with "A pointer" above.

src/core/coap/coap_message.hpp
```diff
@@ -597,23 +597,62 @@ class OptionIterator : public ::otCoapOptionIterator
...
 
+    /**
+     * This method returns a pointer to the next option matching the given option number.
+     * The internal option pointer is advanced until matching option is seen, if no matching
+     * option is seen, the iterator will advance to the end of the options block.
+     *
+     * @param[in]   aOption         Option number to look for.
+     *
+     * @returns A pointer to the next matching option (relative to current iterator position). If no option matching @p
+     * aOption is seen, NULL pointer is returned.
```
jwhui: nit: align with "A pointer" above.

src/core/coap/coap_message.hpp
```diff
@@ -597,23 +597,62 @@ class OptionIterator : public ::otCoapOptionIterator
...
+     * @retval  OT_ERROR_NO_BUFS    Value is too long to fit in a uint64_t.
+     *
+     */
+    otError GetOptionValue(uint64_t &aValue) const;
+
     /**
      * This function fills current option value into @p aValue.
      *
+     * @param[out]  aValue	    Buffer to store the option value.  This buffer is assumed to be sufficiently large (see
+     * @ref otCoapOption::mLength).
```
jwhui: nit: align with "Buffer to" above.

jwhui:
@bukepo any concerns with the code size increase in this PR?

bukepo:
@jwhui Yes, maybe we can make this feature optional at compile time.

jwhui:
> @jwhui Yes, maybe we can make this feature optional at compile time.

@sjlongland, thoughts about wrapping this feature with a build flag?

sjlongland:
I can look into it, I can't imagine it being too difficult to `#if` out some blocks.

jwhui:
@sjlongland, would you also be willing to add a continuous integration test?

sjlongland:
Okay, I've just got the compile-time flag in… it builds with and without.

I did think about CI tests, but I'll have to research how that is done as when I looked into it before, I didn't see any CI tests for CoAP so wasn't sure how to proceed with creating them.

jwhui:
> I did think about CI tests, but I'll have to research how that is done as when I looked into it before, I didn't see any CI tests for CoAP so wasn't sure how to proceed with creating them.

You can find an example of a CoAP-related functional test in [tests/scripts/thread-cert/test_coaps.py](https://github.com/openthread/openthread/blob/master/tests/scripts/thread-cert/test_coaps.py).

sjlongland:
Ahh right, I see how it works now… `pexpect` subprocess calling `ot-cli-${TYPE}`.  _That_ isn't so scary. :-)

sjlongland:
A heads up, working from home today due to illness using my laptop here which runs Gentoo…

* gcc 9.2 complains about the `strncpy(dest, src, sizeof(dest)-1);`, the work-around seems to follow that statement with `dest[sizeof(dest)-1] = '\0';`
* one of the unit tests fails to build in gcc 9.2 due to `memset(this, 0, sizeof(*this));`, replacing this with class assignments seems to have fixed that.
* `make pretty` seems to mandate `clang-format-6.0` which is no longer available in Gentoo.  `clang-format-8.0.1` is working just fine so far after making a wrapper script to "fool" `make pretty` into using it.

sjlongland:
It's taken me a little while to figure out how to run those tests outside of Travis CI…

`make -f examples/Makefile-posix COAP=1 COAP_OBSERVE=1 check` was the magic command in the end.  Not sure if the `COAP=1 COAP_OBSERVE=1` flags are needed yet… I might need to re-visit 7e1ac4a8e430d7590acf4e7b9c7cc8ab5d6aa7fc as the exact execution route was not clear.

For the above to work, I needed to install what was in `tests/scripts/thread-cert/requirements.txt`: mostly what I was missing was `pycryptodome`, I had `pycrypto`, and that bailed because it didn't support `MODE_CCM`.

It'd be worth noting how to set up one's environment and how to run unit tests in https://openthread.io/testing.

sjlongland:
Right, the test case is slightly intermittent, due to the fact that we've got a deliberate race condition and we're using `pexpect` to catch the events… but the test case tries (with both `CON` and `NON`) to subscribe to a point, receive some notifications, then cancel the subscription.

sjlongland:
```
> 2: coap observe fdde:ad00:beef:0:d086:e4d4:e1d3:655a test con
coap observe fdde:ad00:beef:0:d086:e4d4:e1d3:655a test con
Error 6: Parse
```

Okay, that's weird, I'll have to investigate this more thoroughly tomorrow morning.

jwhui:
> ```
> > 2: coap observe fdde:ad00:beef:0:d086:e4d4:e1d3:655a test con
> coap observe fdde:ad00:beef:0:d086:e4d4:e1d3:655a test con
> Error 6: Parse
> ```
> 
> Okay, that's weird, I'll have to investigate this more thoroughly tomorrow morning.

To resolve the `posix-app-cli` and `posix-app-ncp` failures, I believe you need to add `COAP_OBSERVE ?= 1` to https://github.com/openthread/openthread/blob/a5f2a41cb3f05d330269af6f86e4b7be05bc5c0e/src/posix/Makefile-posix#L40

To resolve the `posix-ncp` failure, I believe you need to add `test_coap_observe.py` to `XFAIL_NCP_TESTS`.
https://github.com/openthread/openthread/blob/a5f2a41cb3f05d330269af6f86e4b7be05bc5c0e/tests/scripts/thread-cert/Makefile.am#L285

sjlongland:
On 25/1/20 3:46 am, Jonathan Hui wrote:
> To resolve the `posix-app-cli` and `posix-app-ncp` failures, I believe you need to add `COAP_OBSERVE ?= 1` to https://github.com/openthread/openthread/blob/a5f2a41cb3f05d330269af6f86e4b7be05bc5c0e/src/posix/Makefile-posix#L40

Ahh right, so I *shouldn't* have reverted that earlier commit: the
answer is I need to enable that in both places.

> To resolve the `posix-ncp` failure, I believe you need to add `test_coap_observe.py` to `XFAIL_NCP_TESTS`.
> https://github.com/openthread/openthread/blob/a5f2a41cb3f05d330269af6f86e4b7be05bc5c0e/tests/scripts/thread-cert/Makefile.am#L285

Right, didn't think the NCP test suite would even look at this.  Anyway,
I've added it to the list and we'll see what CI does with it now. :-)

sjlongland:
A heads up, I just did a rebase of the branch to pick up the recently merged PR# 4513.

patrickc77:
@bukepo - Is there anything we can do to help get this PR reviewed and merged?

sjlongland:
I've just rebased the branch to pick up changes from PR #4481.

src/core/coap/coap_message.cpp
```diff
@@ -530,6 +564,26 @@ const otCoapOption *OptionIterator::GetNextOption(void)
...
 }
 
+otError OptionIterator::GetOptionValue(uint64_t &aValue) const
+{
+    otError error = OT_ERROR_NONE;
+    uint8_t value[sizeof(aValue)];
+    uint8_t pos = 0;
+
+    VerifyOrExit(mOption.mLength <= sizeof(aValue), error = OT_ERROR_NO_BUFS);
+    SuccessOrExit(GetOptionValue(value));
```
abtink: How about 
```suggestion
    SuccessOrExit(error = GetOptionValue(value));
```

src/core/coap/coap.cpp
```diff
@@ -478,22 +539,53 @@ void CoapBase::ProcessReceivedResponse(Message &aMessage, const Ip6::MessageInfo
         if (aMessage.IsEmpty())
         {
             // Empty acknowledgment.
-            if (coapMetadata.mConfirmable)
+#if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
+            if (coapMetadata.mObserve && (!request->IsRequest()))
```
abtink:
```suggestion
            if (coapMetadata.mObserve && !request->IsRequest())
```

sjlongland: Heh, yeah, I'm paranoid about operator precedence since I bounce between about 6 different languages and the only thing I lean on is the fact that brackets tend to rule the roost.

I'll make a note to tweak that as I re-base.

src/core/coap/coap_message.cpp
sjlongland: Yep, will fix that.

sjlongland: Fix is in cb96db62e9e5e2b92ea3da222b5c1a454af7facc.

src/core/coap/coap.cpp
sjlongland: Fix is in 0f64a1612379a0b052cdf08a761ec2a8fe5f099e.

src/cli/cli_coap.cpp
```diff
@@ -451,12 +728,31 @@ void Coap::HandleResponse(otMessage *aMessage, const otMessageInfo *aMessageInfo
...
-    else
+    else if ((aMessageInfo != NULL) && (aMessage != NULL))
     {
-        mInterpreter.mServer->OutputFormat("coap response from ");
-        mInterpreter.OutputIp6Address(aMessageInfo->mPeerAddr);
+        otCoapOptionIterator iterator;
+
+        if (otCoapOptionIteratorInit(&iterator, aMessage) == OT_ERROR_NONE)
+        {
+            const otCoapOption *observeOpt =
```
jwhui: nit: add `OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE` around the observe-specific code here?

jwhui: More generally, it seems this PR adds a few hundred bytes of code when the observe feature is not enabled. Just looking for ways to minimize that overhead.

sjlongland: Yep, understood, I found an extra spot where I was declaring and initialising an `otCoapOptionIterator`, but not using it because the offending code was `#if`'ed out by the `OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE` checks, so just widened those out a wee bit to capture the otherwise unused iterator as well.

sjlongland:
Re-based off `master` to resolve reported merge conflict in `examples/Makefile-posix`.

src/cli/cli_coap.hpp
```diff
@@ -82,18 +82,33 @@ class Coap
...
     otError ProcessRequest(int argc, char *argv[]);
     otError ProcessResource(int argc, char *argv[]);
+    otError ProcessSet(int argc, char *argv[]);
     otError ProcessStart(int argc, char *argv[]);
     otError ProcessStop(int argc, char *argv[]);
 
     static void HandleRequest(void *aContext, otMessage *aMessage, const otMessageInfo *aMessageInfo);
     void        HandleRequest(otMessage *aMessage, const otMessageInfo *aMessageInfo);
 
+    static void HandleNotificationResponse(void *               aContext,
```
jwhui: nit: wrap this with `OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE`?

PR #4397, opened 2019-12-12
[spi-hdlc-adapter] add HDLC Flag in front of the HDLC frame

zhanglongxia:
`spi-hdlc-adapter` forgot to add HDLC Flag in front of HDLC frames. RCP fortunately works with `spi-hdlc-adapter`, but in some special environments, they do not work properly. This commit fixes this issue.

### DIFF
--- a/tools/spi-hdlc-adapter/spi-hdlc-adapter.c
+++ b/tools/spi-hdlc-adapter/spi-hdlc-adapter.c
@@ -820,6 +820,8 @@ static int push_hdlc(void)
 
             unescaped_frame_len = sSpiRxPayloadSize;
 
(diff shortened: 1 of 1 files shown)


LuDuda:
@zhanglongxia Can you elaborate more on what are the special requirements, and what do you mean by "RCP fortunately uses spi-hdlc-adapter" - NCP also, right? How the current RCP and NCP work with SPI if spi-hdlc-adapter sends wrong frames?

zhanglongxia:
@LuDuda  I found the root cause of my issue is not caused by the format of the HDLC frame. I have submitted another PR https://github.com/openthread/openthread/pull/4407 to resolve my issue.  This PR will be closed.

PR #4398, opened 2019-12-12
[logging] remove otLog*Err macros and add MLE logs

jwhui:


src/core/thread/mle.cpp
```diff
@@ -2704,20 +2710,11 @@ void Mle::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessageIn
     {
         if (keySequence == neighbor->GetKeySequence())
         {
-            if (frameCounter < neighbor->GetMleFrameCounter())
-            {
-                otLogDebgMle("mle frame reject 1");
-                ExitNow();
-            }
+            VerifyOrExit(frameCounter >= neighbor->GetMleFrameCounter(), error = OT_ERROR_SECURITY);
```
abtink: How about returning `OT_ERROR_DUPLICATE` here and for key sequence instead to distinguish it from mismatching tag?

PR #4400, opened 2019-12-12
[mle] set appropriate Response TLV length in SendChildUpdateResponse

jwhui:


jwhui:
> LGTM👍 Is this handling the case when the TLV is longer than the maximum allowed challenge size?

Future versions of Thread spec may extend existing TLVs by adding new fields. This change ensures forward-compatibility.

PR #4401, opened 2019-12-12
[network-diagnostic] avoid reference binding to null pointer

jwhui:


jwhui:
> LGTM👍 I think this handles time out case.

Confirmed.

PR #4404, opened 2019-12-12
[commissioner] add network name length check in GeneratePskc()

jwhui:


src/core/meshcop/commissioner.cpp
```diff
@@ -1109,7 +1109,8 @@ otError Commissioner::GeneratePskc(const char *              aPassPhrase,
     uint16_t    saltLen = 0;
 
     VerifyOrExit((strlen(aPassPhrase) >= OT_COMMISSIONING_PASSPHRASE_MIN_SIZE) &&
-                     (strlen(aPassPhrase) <= OT_COMMISSIONING_PASSPHRASE_MAX_SIZE),
+                     (strlen(aPassPhrase) <= OT_COMMISSIONING_PASSPHRASE_MAX_SIZE) &&
```
abtink: I am fine with this change as is.
Was also considering if we can use `strnlen` instead of  `strlen` to protect against bad input from user? 
(something like `strnlen(aPassPhrase, OT_COMMISSIONING_PASSPHRASE_MAX_SIZE + 1)`))

jwhui: I updated this PR to use `strnlen`.

PR #4408, opened 2019-12-15
[posix-host] set logging level early

bukepo:
This PR sets logging level just after parsing command line arguments so
that dying messages during initialization can be printed.

### DIFF
--- a/src/posix/main.c
+++ b/src/posix/main.c
@@ -193,10 +193,9 @@ static otInstance *InitInstance(int aArgCount, char *aArgVector[])
 
     ParseArg(aArgCount, aArgVector, &config);
 
-#if OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED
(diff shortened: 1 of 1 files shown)


src/posix/main.c
```diff
@@ -193,10 +193,9 @@ static otInstance *InitInstance(int aArgCount, char *aArgVector[])
 
     ParseArg(aArgCount, aArgVector, &config);
 
-#if OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED
```
zhanglongxia: Why remove the line 196? Only the function "otPlatLog()" calls the function "syslog()".  If 'OPENTHREAD_CONFIG_LOG_OUTPUT' is set to other value. The function "otPlatLog()" won't be implemented and no one calls the function "syslog()". Then we don't need to call the function "openlog()" to initialize the  syslog.

bukepo: Actually, OpenThread's core configurations is not exposed as a user header file, so the posix-host assumes it's configured as platform defined way for logging. Actually, it also assumes dynamic logging is enabled.

PR #4409, opened 2019-12-15
[ncp] reduce size by searching property handlers

bukepo:
The big `switch` generates a lot of code on some compilers. This PR
drops the switch way to find getters and setters, instead a biinary
search is used to find property handlers. This requires all handlers are
defined in a **sorted** array.

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -511,7 +511,7 @@ build_samr21() {
 [ $BUILD_TARGET != posix-distcheck ] || {
     export ASAN_SYMBOLIZER_PATH=`which llvm-symbolizer` || die
     export ASAN_OPTIONS=symbolize=1 || die
-    export DISTCHECK_CONFIGURE_FLAGS= CPPFLAGS="-DOPENTHREAD_POSIX_VIRTUAL_TIME=1 -DOPENTHREAD_CHECK_NCP_HANDLER_ENTRIES_SORTED=1" || die
+    export DISTCHECK_CONFIGURE_FLAGS= CPPFLAGS=-DOPENTHREAD_POSIX_VIRTUAL_TIME=1 || die
     ./bootstrap || die
     REFERENCE_DEVICE=1 make -f examples/Makefile-posix distcheck || die
 }
@@ -572,7 +572,7 @@ build_samr21() {
 
 [ $BUILD_TARGET != posix-ncp ] || {
     ./bootstrap || die
-    REFERENCE_DEVICE=1 COVERAGE=1 PYTHONUNBUFFERED=1 NODE_TYPE=ncp-sim make -f examples/Makefile-posix check || die
+    CPPFLAGS=-DOPENTHREAD_CHECK_NCP_HANDLER_ENTRIES_SORTED=1 REFERENCE_DEVICE=1 COVERAGE=1 PYTHONUNBUFFERED=1 NODE_TYPE=ncp-sim make -f examples/Makefile-posix check || die
 }
 
 [ $BUILD_TARGET != toranj-test-framework ] || {
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -511,7 +511,7 @@ build_samr21() {
 [ $BUILD_TARGET != posix-distcheck ] || {
     export ASAN_SYMBOLIZER_PATH=`which llvm-symbolizer` || die
     export ASAN_OPTIONS=symbolize=1 || die
-    export DISTCHECK_CONFIGURE_FLAGS= CPPFLAGS=-DOPENTHREAD_POSIX_VIRTUAL_TIME=1 || die
+    export DISTCHECK_CONFIGURE_FLAGS= CPPFLAGS="-DOPENTHREAD_POSIX_VIRTUAL_TIME=1 -DOPENTHREAD_CHECK_NCP_HANDLER_ENTRIES_SORTED=1" || die
     ./bootstrap || die
     REFERENCE_DEVICE=1 make -f examples/Makefile-posix distcheck || die
 }
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -572,7 +572,7 @@ build_samr21() {
 
 [ $BUILD_TARGET != posix-ncp ] || {
     ./bootstrap || die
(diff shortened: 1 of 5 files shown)


src/ncp/ncp_base_dispatcher.cpp
```diff
@@ -1231,5 +735,37 @@ NcpBase::PropertyHandler NcpBase::FindRemovePropertyHandler(spinel_prop_key_t aK
...
+NcpBase::PropertyHandler NcpBase::FindPropertyHandler(const HandlerEntry *aHandlerEntries,
+                                                      size_t              aSize,
+                                                      spinel_prop_key_t   aKey)
+{
+    size_t l = 0;
+
+    assert(aSize > 0);
+
+#if OPENTHREAD_CHECK_NCP_HANDLER_ENTRIES_SORTED
+    for (size_t i = 1; i < aSize; ++i)
```
gjc13: Is this check duplicate?

bukepo: Removed.

src/ncp/ncp_base_dispatcher.cpp
```diff
@@ -37,1028 +37,532 @@ namespace Ncp {
...
+       &NcpBase::HandlePropertySet<SPINEL_PROP_THREAD_ALLOW_LOCAL_NET_DATA_CHANGE>},
+#endif
+      {SPINEL_PROP_THREAD_MODE, &NcpBase::HandlePropertySet<SPINEL_PROP_THREAD_MODE>},
+      {SPINEL_PROP_IPV6_ML_PREFIX, &NcpBase::HandlePropertySet<SPINEL_PROP_IPV6_ML_PREFIX>},
+      {SPINEL_PROP_IPV6_ICMP_PING_OFFLOAD, &NcpBase::HandlePropertySet<SPINEL_PROP_IPV6_ICMP_PING_OFFLOAD>},
+      {SPINEL_PROP_IPV6_ICMP_PING_OFFLOAD_MODE, &NcpBase::HandlePropertySet<SPINEL_PROP_IPV6_ICMP_PING_OFFLOAD_MODE>},
+      {SPINEL_PROP_STREAM_NET, &NcpBase::HandlePropertySet<SPINEL_PROP_STREAM_NET>},
+      {SPINEL_PROP_STREAM_NET_INSECURE, &NcpBase::HandlePropertySet<SPINEL_PROP_STREAM_NET_INSECURE>},
+#if OPENTHREAD_CONFIG_JOINER_ENABLE
+      {SPINEL_PROP_MESHCOP_JOINER_COMMISSIONING, &NcpBase::HandlePropertySet<SPINEL_PROP_MESHC
```
abtink: Looks like the array model is only used for Get/Set property handlers.
Should we change Insert/Remove handlers  as well? (I think it's better to have the same model all prop handlers).

bukepo: There are only a few handlers for insert and remove, so I intended leave them as is. However, I am fine if someone else want to change them to binary search. I make this PR higher priority because it is blocking our GitHub sync process.

.travis/script.sh
```diff
@@ -511,7 +511,7 @@ build_samr21() {
 [ $BUILD_TARGET != posix-distcheck ] || {
     export ASAN_SYMBOLIZER_PATH=`which llvm-symbolizer` || die
     export ASAN_OPTIONS=symbolize=1 || die
-    export DISTCHECK_CONFIGURE_FLAGS= CPPFLAGS=-DOPENTHREAD_POSIX_VIRTUAL_TIME=1 || die
+    export DISTCHECK_CONFIGURE_FLAGS= CPPFLAGS="-DOPENTHREAD_POSIX_VIRTUAL_TIME=1 -DOPENTHREAD_CHECK_NCP_HANDLER_ENTRIES_SORTED=1" || die
```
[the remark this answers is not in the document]
bukepo: Yes, it runs, but I think it only runs cli version. I've move the flag to another check. Thank you!

bukepo:
@abtink Thanks for your quick response! I didn't find a static check way in C++03 (I mentioned this in the code). I think we can further enhance the code after we switched to newer version C++, but before that, we have to keep an eye on spinel property changes and ensure the property handlers are sorted.

Besides, I'd like to keep this a pure implementation choice, so I keep most changes  in `ncp_base_dispatch.cpp`. Thoughts?

bukepo:
@abtink @jwhui This is now a blocking issue, could we accelerate reviewing this PR?

.travis/script.sh
```diff
@@ -572,7 +572,7 @@ build_samr21() {
 
 [ $BUILD_TARGET != posix-ncp ] || {
     ./bootstrap || die
-    REFERENCE_DEVICE=1 COVERAGE=1 PYTHONUNBUFFERED=1 NODE_TYPE=ncp-sim make -f examples/Makefile-posix check || die
+    CPPFLAGS=-DOPENTHREAD_CHECK_NCP_HANDLER_ENTRIES_SORTED=1 REFERENCE_DEVICE=1 COVERAGE=1 PYTHONUNBUFFERED=1 NODE_TYPE=ncp-sim make -f examples/Makefile-posix check || die
```
abtink: > It'd be good to have travis build which enables all possible options (such that all props are included in the arrays) and then run the check and ensure the array is sorted. (The trickiness is to maintain such a travis build and ensure all features are enabled in it).

Does this build cover all features?

bukepo: This is using posix sim, which we intentionally enabled all possible features in examples/Makefile-posix. However, I just realized some properties are only enabled on RCP build, so I will turn on the check on posix-app-ncp check as well.

jwhui: I don't think `examples/Makefile-posix` enables all build features that affects the NCP dispatch table (e.g. `OPENTHREAD_CONFIG_MLE_STEERING_DATA_SET_OOB_ENABLE`, `OPENTHREAD_CONFIG_LEGACY_ENABLE`).

bukepo:
@jwhui @abtink Added c++11 code to verify the array is sorted at build time. The negative test result can be found in https://travis-ci.org/bukepo/openthread/builds/625981224

abtink:
Thanks @bukepo. The new model to use c++ feature cons-expr to check if array is sorted at build -time if c++11 is supported looks great. We cover this in a bunch of travis build.

PR #4411, opened 2019-12-16
[neighbor-table] adding NeighborTable

abtink:
This PR contains a group of related commits with overall goal of adding a `NeighborTable` class to allow us to search for neighbors matching a given MAC/IPv6 address and a state filter (so we can find a neighbor entry in any state (not just valid or restoring)).

**[neighbor-table] adding NeighborTable (#4411)**

This commit moves the methods for finding a neighbor (based on short,
extedned MAC address or IPv6 address) from `Mle` and  `MleRouter`
into a newly added class `NeighborTable`. The methods in
`NeighborTable` allow finding a neighbor mathching a given
`Neighbor::StateFilter`.

**[router-table] add StateFilter to methods finding neighbor routers (#4411)**

**[child-table] move StatFilter definition to Neighbor (#4411)**

**[topology] move IsStateValidOrAttaching to Neighbor (#4411)**

**[data-poll-sender] add GetParent() (#4411)**

This commit changes the `Mle::GetParentCandidate()` to return a
reference to the parent candidate and moves the selection logic
between parent or parent candidate when sending data polls to
`DataPollSender` from the newly added `GetParent()` method.

**[mle] update Mle::GetParent() to return reference (#4411)**

**[mle] add Mle::IsAttaching() method (#4411)**

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -211,6 +211,7 @@ LOCAL_SRC_FILES                                          := \
     src/core/thread/mesh_forwarder_mtd.cpp                  \
     src/core/thread/mle.cpp                                 \
     src/core/thread/mle_router.cpp                          \
+    src/core/thread/neighbor_table.cpp                      \
     src/core/thread/network_data.cpp                        \
     src/core/thread/network_data_leader.cpp                 \
     src/core/thread/network_data_leader_ftd.cpp             \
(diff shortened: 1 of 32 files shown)


abtink:
Closing this in favor of #4419 (which is subset of commits from this PR).

PR #4412, opened 2019-12-16
[logging] reduce otThreadErrorToString code size

simonlingoogle:
This PR reduces the code size of `otThreadErrorToString` by about 900 bytes

### DIFF
--- a/src/core/common/logging.cpp
+++ b/src/core/common/logging.cpp
@@ -190,7 +190,7 @@ static const char *const sThreadErrorStrings[OT_NUM_ERRORS] = {
     "ReassemblyTimeout",          // OT_ERROR_REASSEMBLY_TIMEOUT = 30
     "NotTmf",                     // OT_ERROR_NOT_TMF = 31
     "NonLowpanDataFrame",         // OT_ERROR_NOT_LOWPAN_DATA_FRAME = 32
-    "UnknownErrorType",           // otError 33 is unknown
+    "ReservedError33",            // otError 33 is reserved
     "LinkMarginLow",              // OT_ERROR_LINK_MARGIN_LOW = 34
 };
 
--- a/src/core/common/logging.cpp
+++ b/src/core/common/logging.cpp
@@ -156,153 +156,90 @@ void otDump(otLogLevel, otLogRegion, const char *, const void *, const size_t)
 }
 #endif // OPENTHREAD_CONFIG_LOG_PKT_DUMP
 
+static const char *const sThreadErrorStrings[OT_NUM_ERRORS] = {
+    // OT_ERROR_NONE = 0,
+    "OK",
+    // OT_ERROR_FAILED = 1,
+    "Failed",
+    // OT_ERROR_DROP = 2,
+    "Drop",
+    // OT_ERROR_NO_BUFS = 3,
+    "NoBufs",
+    // OT_ERROR_NO_ROUTE = 4,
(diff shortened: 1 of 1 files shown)


src/core/common/logging.cpp
```diff
@@ -156,153 +156,90 @@ void otDump(otLogLevel, otLogRegion, const char *, const void *, const size_t)
...
+    // OT_ERROR_LINK_MARGIN_LOW = 34,
+    "LinkMarginLow",
+};
+
 const char *otThreadErrorToString(otError aError)
 {
     const char *retval;
-
-    switch (aError)
+    if (aError < OT_ARRAY_LENGTH(sThreadErrorStrings))
```
jwhui: nit: add extra line after variable declarations

src/core/common/logging.cpp
```diff
@@ -156,153 +156,56 @@ void otDump(otLogLevel, otLogRegion, const char *, const void *, const size_t)
...
+    "Detached",                   // OT_ERROR_DETACHED = 16
+    "FcsErr",                     // OT_ERROR_FCS = 17
+    "NoFrameReceived",            // OT_ERROR_NO_FRAME_RECEIVED = 18
+    "UnknownNeighbor",            // OT_ERROR_UNKNOWN_NEIGHBOR = 19
+    "InvalidSourceAddress",       // OT_ERROR_INVALID_SOURCE_ADDRESS = 20
+    "AddressFiltered",            // OT_ERROR_ADDRESS_FILTERED = 21
+    "DestinationAddressFiltered", // OT_ERROR_DESTINATION_ADDRESS_FILTERED = 22
+    "NotFound",                   // OT_ERROR_NOT_FOUND = 23
+    "Already",                    // OT_ERROR_ALREADY = 24
+    "UnknownErrorType",           // otError 25 is unknown
```
jwhui:
```suggestion
    "ReservedError25",           // otError 25 is reserved
```

src/core/common/logging.cpp
```diff
@@ -156,153 +156,56 @@ void otDump(otLogLevel, otLogRegion, const char *, const void *, const size_t)
...
+    "Ipv6AddressCreationFailure", // OT_ERROR_IP6_ADDRESS_CREATION_FAILURE = 26
+    "NotCapable",                 // OT_ERROR_NOT_CAPABLE = 27
+    "ResponseTimeout",            // OT_ERROR_RESPONSE_TIMEOUT = 28
+    "Duplicated",                 // OT_ERROR_DUPLICATED = 29
+    "ReassemblyTimeout",          // OT_ERROR_REASSEMBLY_TIMEOUT = 30
+    "NotTmf",                     // OT_ERROR_NOT_TMF = 31
+    "NonLowpanDataFrame",         // OT_ERROR_NOT_LOWPAN_DATA_FRAME = 32
+    "UnknownErrorType",           // otError 33 is unknown
```
jwhui:
```suggestion
    "ReservedErrorType",           // otError 33 is reserved
```

simonlingoogle: However this makes the behavior of `otThreadErrorToString` different from previously.

simonlingoogle: Why not "ReservedError33" ?

jwhui: OK, I am fine with unknown. However, would it be more useful to print the error number if unknown?

PR #4416, opened 2019-12-17
[ncp] conform with c++11 constexpr and static_assert

bukepo:
C++11 constexpr doesn't support local variables and loop. This PR uses
recursive way to assert handlers are sorted.
C++11 static_assert requires a message, this PR adds the missing
messages to static_assert calls.


src/ncp/ncp_base_dispatcher.cpp
```diff
@@ -38,17 +38,9 @@ namespace Ncp {
 #if __cplusplus >= 201103L
 constexpr bool NcpBase::IsHandlerEntriesSorted(const HandlerEntry *aHandlerEntries, size_t aSize)
```
jwhui:
```suggestion
constexpr bool NcpBase::AreHandlerEntriesSorted(const HandlerEntry *aHandlerEntries, size_t aSize)
```

PR #4419, opened 2019-12-17
[mle] change GetParent and GetParentCandidate

abtink:
This PR contains a group of related  smaller commits to simplify the handling of parent and parent candidate from MLE: 

**[child-table] move StateFilter definition to Neighbor**

**[topology] move IsStateValidOrAttaching to Neighbor**

**[data-poll-sender] add GetParent()**

This commit changes the `Mle::GetParentCandidate()` to return a
reference to the parent candidate and moves the selection logic
between parent or parent candidate when sending data polls to
`DataPollSender` from the newly added `GetParent()` method.

**[mle] update Mle::GetParent() to return reference**

**[mle] add Mle::IsAttaching() method**

### DIFF
--- a/src/core/thread/mle.hpp
+++ b/src/core/thread/mle.hpp
@@ -603,6 +603,19 @@ public:
     bool IsAttached(void) const;
 
     /**
+     * This method indicates whether device is currently attaching or not.
+     *
+     * Note that an already attached device may also be in attaching state. Examples of this include a leader/router
+     * trying to attach to a better partition, or a child trying to find a better parent (when feature
+     * `OPENTHREAD_CONFIG_PARENT_SEARCH_ENABLE` is enabled).
+     *
+     * @retval TRUE   Device is currently trying to attach.
+     * @retval FALSE  Device is not in middle of attach process.
+     *
+     */
+    bool IsAttaching(void) const { return (mAttachState != kAttachStateIdle); }
+
+    /**
      * This method returns the current Thread interface state.
      *
      * @returns The current Thread interface state.
--- a/src/core/thread/mle.hpp
+++ b/src/core/thread/mle.hpp
@@ -747,12 +747,12 @@ public:
     }
 
     /**
-     * This method returns a pointer to the parent when operating in End Device mode.
+     * This method gets the parent when operating in End Device mode.
      *
-     * @returns A pointer to the parent.
+     * @returns A reference to the parent.
      *
      */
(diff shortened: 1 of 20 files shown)


src/core/thread/mle.hpp
```diff
@@ -602,6 +602,19 @@ class Mle : public InstanceLocator
...
+     *
+     * Note that an already attached device may also be in attaching state. Examples of this include a leader/router
+     * trying to attach to a better partition, or a child trying to find a better parent (when feature
+     * `OPENTHREAD_CONFIG_PARENT_SEARCH_ENABLE` is enabled).
+     *
+     * @retval TRUE   Device is currently trying to attach.
+     * @retval FALSE  Device is not in middle of attach process.
+     *
+     */
+    bool IsAttaching(void) const { return (mAttachState != kAttachStateIdle); }
```
jwhui: How about using this method here:
https://github.com/openthread/openthread/blob/c740a053fd3181c085a7e13a2f4c6a66ed413b17/src/core/thread/mle.cpp#L627

src/core/thread/topology.hpp
```diff
@@ -157,6 +173,27 @@ class Neighbor
      */
     bool IsStateValidOrRestoring(void) const { return (mState == kStateValid) || IsStateRestoring(); }
 
+    /**
+     * This method indicates if the neighbor state is valid or being attached or being restored.
```
jwhui:
```suggestion
     * This method indicates if the neighbor state is valid, attaching, or restoring.
```

src/core/thread/topology.hpp
```diff
@@ -157,6 +173,27 @@ class Neighbor
...
+     *
+     * The states `kStateRestored`, `kStateChildIdRequest`, `kStateChildUpdateRequest`, `kStateValid`, and
+     * `kStateLinkRequest` are considered as attached or being restored.
```
jwhui:
```suggestion
     * `kStateLinkRequest` are considered as attached, attaching, or restored.
```

src/core/thread/topology.hpp
```diff
@@ -157,6 +173,27 @@ class Neighbor
...
+     *
+     * @returns TRUE if the neighbor is attached or being restored.
```
jwhui:
```suggestion
     * @returns TRUE if the neighbor is attached, attaching, or restored.
```

PR #4427, opened 2019-12-24
[mle] link sync enhancement

librasungirl:
There is chance for below scenario (as shown in below screenshot)
1) Device newly becomes router and sends multicast Link Request
   (challenge1),
2) Device receives MLE advertisement from one neighbor, and sends
   unicast Link Request (challenge2),
3) The neighbor ignores the unicast link Request in 2) and sends
   out Link Accept and Request message as response for 1)
4) Device would drop the Link Accept adn Request message
   in 3) as it compares the Response TLV value with challenge2, which
   in fact matches challenge1.
Screen Shot 2019-12-24 at 5 56 11 PM

The link between Device and the neighbor could not be established until
Device receives next MLE advertisement from the neighbor.

This commit suppresses unicast Link Request in 2) if there is multicast
Link Request ongoing. Thus the three-handshakes link sync could go well.

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -1397,8 +1397,8 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
         router = mRouterTable.GetRouter(routerId);
         VerifyOrExit(router != NULL);
 
-        // Send link request if no link to router
-        if (!router->IsStateValid() && !router->IsStateLinkRequest() &&
+        // Send unicast link request if no link to router and no unicast/multicast link request in progress
+        if (!router->IsStateValid() && !router->IsStateLinkRequest() && mChallengeTimeout == 0 &&
             (linkMargin >= OPENTHREAD_CONFIG_MLE_LINK_REQUEST_MARGIN_MIN))
         {
             router->SetExtAddress(macAddr);
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -1398,7 +1398,7 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
         VerifyOrExit(router != NULL);
 
         // Send unicast link request if no link to router and no unicast/multicast link request in progress
-        if (!router->IsStateValid() && !router->IsStateLinkRequest() && mChallengeTimeout == 0 &&
+        if (!router->IsStateValid() && !router->IsStateLinkRequest() && (mChallengeTimeout == 0) &&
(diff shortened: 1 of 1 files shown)


abtink:
Nice find...  :+1: 

Can we consider addressing this on the neighbor side? Have the neighbor not ignore the second unicast "Link Request" (from step 2)  it gets and respond to it with a "Link Accept" as well?

Another way to consider is to re-use the same challenge when sending the unicast "Link Request" (if `mChallengeTimeout` is non-zero)? 

The reason is to not suppress the unicast msg if possible (since unicasts are more reliable compared to broadcast tx).

Thoughts?

bukepo:
@abtink Good idea reusing the same challenge!  :+1:

librasungirl:
@abtink thanks for the proposals.

I like the idea to reuse the same challenge value, but I'm afraid it might not conform with the spec:

> challenge TLV
> The byte string MUST be at least 4 bytes in length and a new value MUST be chosen for each Challenge TLV transmitted

Ignoring the 'Link Request' is designed to avoid the redundant responses if there is already one as there might be retransmission for unicast message.

The corner case this PR tries to enhance is when there is unicast Link Request closely after multicast Link Request, which generally only happens when the device newly upgrade to Router role and the MLE advertisement from the neighbor arrives earlier than the response for the multicast Link Request. 
The worst result for this suppress solution - if the neighbor doesn't hear the multicast Link Request, link sync is postponed to next MLE advertisement from the neighbor,  is not worse than before.

A little hesitation after second thought - whether or not it is worthwhile to apply the 'suppress' solution might depend on the chance that the neighbor doesn't hear the multicast Link Request v.s. that the neighbor queues the MLE advertisement before Link Accept.

Thoughts? @abtink @jwhui

jwhui:
I agree with the proposed fix here. The device should not be sending additional Link Request messages if it is already waiting for Link Accept messages.

src/core/thread/mle_router.cpp
```diff
@@ -1397,8 +1397,8 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
         router = mRouterTable.GetRouter(routerId);
         VerifyOrExit(router != NULL);
 
-        // Send link request if no link to router
-        if (!router->IsStateValid() && !router->IsStateLinkRequest() &&
+        // Send unicast link request if no link to router and no unicast/multicast link request in progress
+        if (!router->IsStateValid() && !router->IsStateLinkRequest() && mChallengeTimeout == 0 &&
```
jwhui:
```suggestion
        if (!router->IsStateValid() && !router->IsStateLinkRequest() && (mChallengeTimeout == 0) &&
```

librasungirl:
Updated. Thanks.

PR #4430, opened 2019-12-26
[bbr] add backbone router service

librasungirl:
Refactored according to feedbacks.  
 ```
    This commit mainly includes:

    [Y] Introduce BACKBONE_ROUTER option for Backbone Router function.
    [Y] Implement Backbone Router service registration.
        [Y] Add basic Backbone Router service.
        [Y] Primary Backbone Router restores its Dataset when
            reattached after short reset, increases sequence
            number and re-register to Leader.
        [Y] Add configurable jitter for Backbone Router service
            registration.
    [Y] Add Backbone Router service test.

```

### CODE REVIEW

File: script/test [hunk 1]
```diff
@@ -141,8 +149,10 @@ main()
     export VERBOSE="${VERBOSE:-1}"
     export VIRTUAL_TIME="${VIRTUAL_TIME:-1}"
     export THREAD_VERSION="${THREAD_VERSION:-1.1}"
+
     if [ "${THREAD_VERSION}" = "1.2" ]; then
-        export top_builddir_1_1="build/${SYSTEM_TRIPLET}-1.1"
+        export top_builddir_1_1="${OT_BUILDDIR}/${OT_SYSTEM_TRIPLET}-1.1"
```
jwhui: This seems unrelated to BBR function. Move to separate PR?

src/core/Makefile.am
```diff
@@ -141,6 +143,8 @@ SOURCES_COMMON                             = \
     api/thread_api.cpp                       \
     api/thread_ftd_api.cpp                   \
     api/udp_api.cpp                          \
+    backbone_router/leader.cpp               \
+    backbone_router/local.cpp                 \
```
jwhui: nit: alignment

include/openthread/backbone_router_ftd.h
```diff
@@ -0,0 +1,160 @@
...
+ *
+ */
+otError otBackboneRouterGetLocal(otInstance *aInstance, otBackboneRouterConfig *aConfig);
+
+/**
+ * This function sets the local Backbone Router configuration.
+ *
+ * @param[in]  aInstance             A pointer to an OpenThread instance.
+ * @param[in]  aConfig               A pointer to the Backbone Router configuration to take effect.
+ * @param[in]  aConfigComponents     A pointer to the structure that specifies which componenents to
```
jwhui:
```suggestion
 * @param[in]  aConfigComponents     A pointer to the structure that specifies which components to
```

src/core/backbone_router/leader.hpp
```diff
@@ -0,0 +1,188 @@
...
+ * This class implements the basic Primary Backbone Router (PBbr) service operations.
+ *
+ */
+class Leader : public InstanceLocator
+{
+public:
+    // Primary Backbone Router Service State or State change.
+    enum State
+    {
+        kStateNone = 0,       // Not exist (trigger Backbnoe Router register its service).
```
jwhui:
```suggestion
        kStateNone = 0,       // Not exist (trigger Backbone Router register its service).
```

src/core/backbone_router/leader.hpp
```diff
@@ -0,0 +1,188 @@
...
+        kStateAdded,          // Newly added.
+        kStateRemoved,        // Newly removed (trigger Backbnoe Router register its service).
```
jwhui:
```suggestion
        kStateRemoved,        // Newly removed (trigger Backbone Router register its service).
```

src/core/backbone_router/leader.hpp
```diff
@@ -0,0 +1,188 @@
...
+        kStateToTriggerRereg, // Short address or sequence number chagnes (trigger re-registration).
```
jwhui:
```suggestion
        kStateToTriggerRereg, // Short address or sequence number changes (trigger re-registration).
```

src/core/backbone_router/local.hpp
```diff
@@ -0,0 +1,162 @@
...
+     * @param[out]  aConfig  A reference whether to put local Backbone Router configuration.
+     *
+     */
+    void GetConfig(BackboneRouterConfig &aConfig) const;
+
+    /**
+     * This method sets local Backbone Router configuration.
+     *
+     * @param[in]  aConfig              A reference of the configuration to set.
+     * @param[in]  aConfigComponents    A struct to specify which componenents to
```
jwhui:
```suggestion
     * @param[in]  aConfigComponents    A struct to specify which components to
```

src/core/thread/network_data_leader.cpp
```diff
@@ -67,6 +67,117 @@ void LeaderBase::Reset(void)
...
+            rvalServerTlv  = const_cast<ServerTlv *>(serverTlv);
+            rvalServerData = serverData;
+        }
+
+        subCur = subCur->GetNext();
+    }
+
+    error = OT_ERROR_NONE;
+
+    if (aConfig.mServer16 == Mac::kShortAddrInvalid)
```
jwhui: How about keeping this method to simply retrieving the primary bbr info and moving the state change logic to `Leader::Update()`?

src/cli/README.md
```diff
@@ -77,6 +78,106 @@ OpenThread test scripts use the CLI to execute test cases.
...
+Primary
+Done
+
+> bbr state
+Secondary
+Done
+```
+
+### bbr enable
+Enable Backbone Router Service for Thread 1.2 FTD (for test purpose).
```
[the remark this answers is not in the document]
librasungirl: yes. should also be useful in production. will update

src/cli/cli.cpp
```diff
@@ -434,6 +444,127 @@ void Interpreter::ProcessHelp(int argc, char *argv[])
...
+                }
+                argcur++;
+            }
+
+            if (!configComponents.mIsSequenceNumberPresent)
+            {
+                otBackboneRouterConfig currentConfig;
+                otBackboneRouterGetLocal(mInstance, &currentConfig);
+                configComponents.mIsSequenceNumberPresent = true;
+                config.mSequenceNumber                    = ++currentConfig.mSequenceNumber;
```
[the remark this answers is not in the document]
librasungirl: yes, some legacy code. will update.

src/core/backbone_router/leader.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+    VerifyOrExit(HasPrimary(), error = OT_ERROR_NOT_FOUND);
+
+    aConfig = mConfig;
+
+exit:
+    return error;
+}
+
+otError Leader::GetServiceId(uint8_t &aSID) const
```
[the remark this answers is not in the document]
librasungirl: seems better. will update

src/core/backbone_router/leader.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+Leader::Leader(Instance &aInstance)
+    : InstanceLocator(aInstance)
+{
+    mCallbacks = NULL;
+    memset(&mConfig, 0, sizeof(mConfig));
+    mConfig.mServer16 = Mac::kShortAddrInvalid;
+}
+
+void Leader::RegisterCallback(Callback &aCallback)
```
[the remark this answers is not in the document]
librasungirl: Different components (bbr local now, and further DUA/MA) may take actions on the primary bbr state change, we may directly call all relative components in one place when `Leader::Update()` here, however with the list of callbacks here, we can also let each component to decide independently whether or not they are interested in the state change.

Thoughts?

src/core/backbone_router/local.cpp
```diff
@@ -0,0 +1,284 @@
...
+        mState = OT_BBR_STATE_SECONDARY;
+        AddService();
+    }
+    else
+    {
+        if (mState == OT_BBR_STATE_PRIMARY)
+        {
+            RemoveService();
+            Get<ThreadNetif>().RemoveUnicastAddress(mBackboneRouterPrimaryAloc);
+            mState = OT_BBR_STATE_DISABLED;
```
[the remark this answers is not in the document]
librasungirl: make sense. will update.

src/core/backbone_router/local.cpp
```diff
@@ -0,0 +1,284 @@
...
+otError Local::RemoveService(void)
+{
+    otError error       = OT_ERROR_NONE;
+    uint8_t serviceData = NetworkData::ServiceTlv::kServiceDataBackboneRouter;
+
+    error = Get<NetworkData::Local>().RemoveService(THREAD_ENTERPRISE_NUMBER, &serviceData, sizeof(serviceData));
+    SuccessOrExit(error);
+    otLogDebgNetData("Removed Bbr Service from local Network Data");
+
+    error = Get<NetworkData::Local>().SendServerDataNotification();
```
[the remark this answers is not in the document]
librasungirl: yes, but maybe we can keep the return value for logging below?

src/core/backbone_router/local.cpp
```diff
@@ -0,0 +1,284 @@
...
+    switch (aState)
+    {
+    case Leader::kStateNone:
+        // fall through
+    case Leader::kStateRemoved:
+        VerifyOrExit(Get<Mle::MleRouter>().IsAttached());
+
+        if (!Get<BackboneRouter::Leader>().HasPrimary())
+        {
+            AddService();
```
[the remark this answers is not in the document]
librasungirl: make sense 👍  will update

src/core/backbone_router/local.cpp
```diff
@@ -0,0 +1,284 @@
...
+                Get<Notifier>().Signal(OT_CHANGED_THREAD_BACKBONE_ROUTER_STATE);
+            }
+        }
+        else
+        {
+            if (mState == OT_BBR_STATE_PRIMARY)
+            {
+                Reset();
+                mSequenceNumber++;
+                Store();
```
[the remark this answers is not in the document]
librasungirl: agree.

src/core/backbone_router/local.cpp
```diff
@@ -0,0 +1,284 @@
...
+            }
+        }
+        break;
+    default:
```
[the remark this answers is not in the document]
librasungirl: will update.

src/core/thread/network_data_leader.cpp
librasungirl: seems more reasonable in code structure. will update.

src/core/backbone_router/leader.cpp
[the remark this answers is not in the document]
jwhui: It makes sense if we need the flexibility. Otherwise, I think it's less code/memory to have direct calls.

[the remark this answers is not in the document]
librasungirl: removed callback

examples/Makefile-cc1352
```diff
@@ -44,7 +44,6 @@ BuildJobs                                                 ?= 10
 configure_OPTIONS                                          = \
     --enable-cli                                             \
     --enable-ftd                                             \
-    --enable-mtd                                             \
```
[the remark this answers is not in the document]
librasungirl: separated it into https://github.com/openthread/openthread/pull/4615 but abandoned as it seems not necessary for now.

src/core/thread/network_data_leader.cpp
```diff
@@ -67,6 +67,90 @@ void LeaderBase::Reset(void)
...
+             serverTlv->GetServer16() > rvalServerTlv->GetServer16()))
+        {
+            rvalServerTlv  = const_cast<ServerTlv *>(serverTlv);
+            rvalServerData = serverData;
+        }
+
+        subCur = subCur->GetNext();
+    }
+
+    aConfig.mServer16            = rvalServerTlv->GetServer16();
```
simonlingoogle: need to verify that `rvalServerTlv != NULL` here?

librasungirl: If there is no available service, it will exit in https://github.com/openthread/openthread/blob/adca16f505b568829c53ecbc774788d30a66ceed/src/core/thread/network_data_leader.cpp#L110-L114

When it goes here, rvalServerTlv should not be NULL.

src/core/backbone_router/local.cpp
```diff
@@ -0,0 +1,246 @@
...
+    memcpy(&mBackboneRouterPrimaryAloc.GetAddress(), &Get<Mle::MleRouter>().GetMeshLocal16(), 14);
+    mBackboneRouterPrimaryAloc.GetAddress().mFields.m16[7] = HostSwap16(Mle::kAloc16BackboneRouterPrimary);
+    return Get<ThreadNetif>().AddUnicastAddress(mBackboneRouterPrimaryAloc);
+}
+
+void Local::NotifyBackboneRouterPrimaryUpdate(Leader::State aState, const BackboneRouterConfig &aConfig)
+{
+    OT_UNUSED_VARIABLE(aConfig);
+
+    switch (aState)
```
simonlingoogle: for states `kStateNone` and `kStateRemoved, should the PBBR `Reset()` to be SBBR?

src/core/backbone_router/local.cpp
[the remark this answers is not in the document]
simonlingoogle: Can we move the current state transfer logic in `SetEnabled` and `Reset` to `SetState` and call `SetState` properly in these two functions?

src/core/thread/mle.cpp
```diff
@@ -746,6 +750,10 @@ bool Mle::IsAttached(void) const
 
 void Mle::SetStateDetached(void)
 {
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+    Get<BackboneRouter::Local>().Reset();
```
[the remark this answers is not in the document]
simonlingoogle: I see why it is `secondary` now.

src/core/backbone_router/local.cpp
```diff
@@ -0,0 +1,246 @@
...
+                    // Here original PBBR restores its Backbone Rouetr Service from Thread Network,
+                    // Intentionally skips the state update as PBBR will refresh its service.
+                    mSequenceNumber      = aConfig.mSequenceNumber + 1;
+                    mReregistrationDelay = aConfig.mReregistrationDelay;
+                    mMlrTimeout          = aConfig.mMlrTimeout;
+                    AddService(true /* Force registration to refresh and restore Primary state */);
+                }
+                else
+                {
+                    AddBackboneRouterPrimaryAloc();
```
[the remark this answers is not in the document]
librasungirl: The `local` and `leader` in namespace `BackboneRouter` are some kind of `cache` components, 
`local` cache for local bbr dataset with additionally logic to decide whether or not contribute to local server data as bbr service should be unique in general, the `leader` is to cache the primary backbone router service per network data update as the status of change of it may cause extra actions for other components.  The consistency check is left to the network_data_local in core per network data update or per MLE advertisement from neighbor.

 if local SeqNum is in local server data, the core will be responsible for the consistency check and register it if inconsistent. Why needs increase SeqNum and force registration?

src/core/thread/network_data_leader.cpp
librasungirl: for bbr service, I think it should contain server tlv. Will add the check to be safe as your suggestion.

src/core/api/backbone_router_api.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <openthread/backbone_router.h>
+#include "common/instance.hpp"
+
+using namespace ot;
+
+otError otBackboneRouterGetPrimary(otInstance *aInstance, otBackboneRouterConfig *aConfig)
+{
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
+    assert(aConfig != NULL);
```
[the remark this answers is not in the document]
simonlingoogle: I actually lean towards `assert`. But did saw some code that returns `OT_ERROR_INVALID_ARGS` for NULL.

src/core/backbone_router/local.cpp
simonlingoogle: Agree that `kStateNone` & `PBBR` should never happen. 
But `kStateRemoved` and `PBBR` could happen. Thoughts?

librasungirl: For what I can see, It may happen in some extreme scenario, for example, the non-leader PBBR could hear packets from the Thread Network, but any other devices in the Thread Network could not hear PBBR, then the Leader may remove the PBBR router id and remove its services. however in this case, PBBR itself should become detached when it detects its router id is not allocated any more in which it will reset local bbr state.
https://github.com/openthread/openthread/blob/adca16f505b568829c53ecbc774788d30a66ceed/src/core/thread/mle.cpp#L754

Any other possible such scenarios I missed?

simonlingoogle: I was thinking from the standpoint of considering all possible logical cases. 

Could `Leader` switching cause some `Network Data` to be lost and led to this situation?

src/cli/README.md
```diff
@@ -98,6 +99,132 @@ Done
...
+```bash
+> bbr local
+BBR local:
+seqno:    10
+delay:    120 secs
+timeout:  300 secs
+Done
+```
+
+### bbr local \[seqno \<seqno\>\] \[delay \<delay\>\] \[timeout \<timeout\>\]
```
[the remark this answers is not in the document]
simonlingoogle: > for your suggestion, how could it avoid the check on whether or not it is SBBR and have advantages in guarantee one SRV_DATA.ntf?

The check will be done by the `bbr local reg` command handler, so the CLI user does not need to check it.

include/openthread/backbone_router_ftd.h
```diff
@@ -0,0 +1,170 @@
...
+ * @retval OT_ERROR_NO_BUFS           Insufficient space to add the Backbone Router service.
+ * @retval OT_ERROR_NONE              Successfully queued a Server Data Request message for delivery.
+ *
+ * @sa otBackboneRouterSetEnabled
+ * @sa otBackboneRouterGetState
+ * @sa otBackboneRouterGetLocal
+ * @sa otBackboneRouterSetLocal
+ *
+ */
+otError otBackboneRouterRegisterLocal(otInstance *aInstance);
```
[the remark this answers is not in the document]
simonlingoogle: This method registers BBR dataset to local service forcely. This operation can change the current PBBR in the network. Just thought that choosing primary is the business of Thread network and users don't need to care about this process if not for testing or debugging. 

But I am fine to providing richer toolkit for users.

src/core/backbone_router/local.cpp
```diff
@@ -0,0 +1,246 @@
...
+void Local::SetEnabled(bool aEnable)
+{
+    VerifyOrExit(aEnable == (mState == OT_BACKBONE_ROUTER_STATE_DISABLED));
+
+    if (aEnable)
+    {
+        AddService();
+        SetState(OT_BACKBONE_ROUTER_STATE_SECONDARY);
+    }
+    else if (mState == OT_BACKBONE_ROUTER_STATE_PRIMARY)
```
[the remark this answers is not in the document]
simonlingoogle: Since we have `bbr register` CLI command and `otBackboneRouterRegisterLocal`, it should not be a rare case that `SBBR` has `BBR dataset` in local server data.

src/cli/README.md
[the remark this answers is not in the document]
librasungirl: Maybe I didn't get your point. Could you please explain a bit more on the usage of your proposal? 
now I know you want to check whether or not its SBBR in CLI app. One question is when the user need to add `reg`?

src/core/backbone_router/local.cpp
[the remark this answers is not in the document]
librasungirl: good point. will enhance.

src/cli/README.md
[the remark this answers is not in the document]
simonlingoogle: This command is almost same as `setBbrDataset` in THCI. Same as we have proposed to add a `forceSendSvrDataNtf` argument ot `setBbrDataset`, the `reg` argument for `bbr local` command is for the same purpose.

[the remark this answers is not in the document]
librasungirl: I think the conclusion in that JIRA is `setBbrDataset()` without `forceSendSvrDataNtf` and use fake SRV_DATA.ntf as it could meet existing TPs though it has some assumptions.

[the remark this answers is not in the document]
librasungirl: Even if there are two SRV_DATA.ntf, what's the impact? why is it so important to keep 1 SRV_DATA.ntf?

I think in general user scenarios, OpenThread (CLI or API) Users only need to config Bbr Dataset and leave other things to OpenThread Core. User should be aware of the impact of explicitly issue `bbr register` or call `otBackboneRouterRegisterLocal()`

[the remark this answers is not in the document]
librasungirl: The possible impact of the test case you concerns is that for SBBR, `setBbrDataset()` may only update local Bbr Dataset but not add it to local server data, when the fake SRV_DATA.ntf occurs, original PBBR will deregister due to lower sequence number. And SBBR would also `deregister` due to network data inconsistency as its own bbr dataset is not in local server dataset yet. So there might be one interim period that there will be no PBBR in the network. But afterwards, both SBBR and original PBBR will try to register and finally SBBR will become new PBBR due to higher sequence number, which should also be the expecting situations of TP/TH. 
Since it is decided to use the constructed fake SRV_DATA.ntf, I think Test Harness should be robust for such SRV_DATA.ntf message exchanges.
Thoughts?

src/core/backbone_router/local.cpp
librasungirl: If there is `new Leader` losing Network Data somehow, no matter whether the PBBR migrates to the new partition,  or forms new partition among with the devices in its communication range, or sticks to its old partition if `old leader` still in connection, it will either go through the same detach and attach process as above, or keep its primary state in old partition. 

Any more possible specific scenarios in your thoughts?

src/cli/README.md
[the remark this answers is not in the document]
simonlingoogle: I am fine with the current `bbr config` without `reg` argument since it is not essential anyway.

src/core/backbone_router/local.cpp
```diff
@@ -75,11 +68,9 @@ void Local::SetEnabled(bool aEnable)
         AddService();
         SetState(OT_BACKBONE_ROUTER_STATE_SECONDARY);
     }
-    else if (mState == OT_BACKBONE_ROUTER_STATE_PRIMARY)
+    else
     {
-        RemoveService();
-        Get<ThreadNetif>().RemoveUnicastAddress(mBackboneRouterPrimaryAloc);
-        SetState(OT_BACKBONE_ROUTER_STATE_DISABLED);
+        Reset();
```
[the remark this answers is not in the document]
librasungirl: right. updated.

src/core/backbone_router/local.cpp
```diff
@@ -0,0 +1,251 @@
...
+{
+}
+
+void Local::SetEnabled(bool aEnable)
+{
+    VerifyOrExit(aEnable == (mState == OT_BACKBONE_ROUTER_STATE_DISABLED));
+
+    if (aEnable)
+    {
+        AddService();
```
simonlingoogle: when `mState == OT_BACKBONE_ROUTER_STATE_DISABLED` and `aEnable == true`, this call to `AddService()` is useless because `VerifyOrExit(mState != OT_BACKBONE_ROUTER_STATE_DISABLED && Get<Mle::Mle>().IsAttached());` statement in `AddService`.

src/cli/cli.cpp
```diff
@@ -435,6 +445,139 @@ void Interpreter::ProcessHelp(int argc, char *argv[])
...
+        case OT_BACKBONE_ROUTER_STATE_PRIMARY:
+            mServer->OutputFormat("Primary\r\n");
+            break;
+        }
+    }
+    else if (strcmp(argv[0], "config") == 0)
+    {
+        if (argc == 1)
+        {
+            otBackboneRouterGetConfig(mInstance, &config);
```
simonlingoogle: Should also call `otBackboneRouterGetConfig(mInstance, &config);` when `argc > 1`?

librasungirl: yes, updated.

src/core/backbone_router/local.cpp
```diff
@@ -0,0 +1,258 @@
...
+        RemoveService();
+
+        if (mState == OT_BACKBONE_ROUTER_STATE_PRIMARY)
+        {
+            // Here increases sequence number if becomes Secondary from Primary.
+            mSequenceNumber++;
+            Get<Notifier>().Signal(OT_CHANGED_THREAD_BACKBONE_ROUTER_LOCAL);
+        }
+
+        SetState(OT_BACKBONE_ROUTER_STATE_SECONDARY);
```
simonlingoogle:
```suggestion
        Reset(); 
```
This is effectively same as `Reset();` ?

src/core/api/backbone_router_ftd_api.cpp
```diff
@@ -0,0 +1,97 @@
...
+    OT_ASSERT(aConfig != NULL);
+
+    instance.Get<BackboneRouter::Local>().SetConfig(*aConfig);
+}
+
+otError otBackboneRouterRegister(otInstance *aInstance)
+{
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
+    return instance.Get<BackboneRouter::Local>().AddService(true /* Forcely register local Backbone Router Service */);
```
jwhui:
```suggestion
    return instance.Get<BackboneRouter::Local>().AddService(true /* Force registration */);
```

src/core/backbone_router/local.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+void Local::Reset(void)
+{
+    VerifyOrExit(mState != OT_BACKBONE_ROUTER_STATE_DISABLED);
+
+    RemoveService();
+
+    if (mState == OT_BACKBONE_ROUTER_STATE_PRIMARY)
+    {
+        // Increase sequenence number when changing from primary to secondary.
```
jwhui:
```suggestion
        // Increase sequence number when changing from primary to secondary.
```

src/core/backbone_router/local.cpp
```diff
@@ -0,0 +1,249 @@
...
+        // Here uses the timer resource in Mle.
+        Get<Mle::MleRouter>().SetBackboneRouterRegistrationDelay(delay);
+    }
+    else if (aConfig.mServer16 != Get<Mle::MleRouter>().GetRloc16())
+    {
+        Reset(); 
+    }
+    else if (!mIsServiceAdded)
+    {
+        // Here original PBBR restores its Backbone Rouetr Service from Thread Network,
```
jwhui:
```suggestion
        // Here original PBBR restores its Backbone Router Service from Thread Network,
```

src/core/backbone_router/local.hpp
```diff
@@ -0,0 +1,182 @@
...
+     *
+     */
+    void SetConfig(const BackboneRouterConfig &aConfig);
+
+    /**
+     * This method registers Backbone Router Dataset to Leader.
+     *
+     * @param[in]  aForce True if to register forcely regardless of current BackboneRouterState.
+     *                    False if allow to decide whether or not to register according to
+     *                    current BackboneRouterState.
```
jwhui:
```suggestion
     * @param[in]  aForce True to force registration regardless of current BackboneRouterState.
     *                    False to decide based on current BackboneRouterState.
```

include/openthread/backbone_router_ftd.h
```diff
@@ -0,0 +1,170 @@
...
+    OT_BACKBONE_ROUTER_STATE_DISABLED  = 0, ///< Backbone function is disabled.
+    OT_BACKBONE_ROUTER_STATE_SECONDARY = 1, ///< Secondary Backbone Router.
+    OT_BACKBONE_ROUTER_STATE_PRIMARY   = 2, ///< The Primary Backbone Router.
+} otBackboneRouterState;
+
+/**
+ * This function enables or disables Backbone functionality.
+ *
+ * @param[in] aInstance A pointer to an OpenThread instance.
+ * @param[in] aEnable   TRUE if enable Backbone functionality, FALSE otherwise.
```
jwhui:
```suggestion
 * @param[in] aEnable   TRUE to enable Backbone functionality, FALSE otherwise.
```

src/core/backbone_router/leader.hpp
```diff
@@ -0,0 +1,146 @@
...
+     * @param[out]  aConfig        A reference whether to put the Primary Backbone Router information.
+     *
+     * @retval OT_ERROR_NONE       Successfully got the Primary Backbone Router information.
+     * @retval OT_ERROR_NOT_FOUND  No Backbone Router in the Thread Network.
+     *
+     */
+    otError GetConfig(BackboneRouterConfig &aConfig) const;
+
+    /**
+     * This method gets the Backbone Router service id.
```
jwhui:
```suggestion
     * This method gets the Backbone Router service ID.
```

src/core/backbone_router/leader.hpp
```diff
@@ -0,0 +1,146 @@
...
+     *
+     * @param[out]  aServiceId     A reference whether to put the Backbone Router service id.
```
jwhui:
```suggestion
     * @param[out]  aServiceId     The Backbone Router service ID.
```

src/core/backbone_router/local.hpp
```diff
@@ -0,0 +1,182 @@
...
+ */
+class Local : public InstanceLocator
+{
+public:
+    typedef otBackboneRouterState BackboneRouterState;
+
+    /**
+     * This constructor initializes the local Backbone Router.
+     *
+     * @param[in]  aNetif  A reference to the Thread network interface.
```
jwhui:
```suggestion
     * @param[in]  aInstance  A reference to the Thread instance.
```

src/core/backbone_router/local.hpp
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * This method resets the local Thread Network Data.
+     *
+     */
+    void Reset(void);
+
+    /**
+     * This method gets local Backbone Router configuration.
+     *
+     * @param[out]  aConfig  A reference whether to put local Backbone Router configuration.
```
jwhui:
```suggestion
     * @param[out]  aConfig  The local Backbone Router configuration.
```

src/core/backbone_router/local.hpp
```diff
@@ -0,0 +1,182 @@
...
+     *
+     */
+    void GetConfig(BackboneRouterConfig &aConfig) const;
+
+    /**
+     * This method sets local Backbone Router configuration.
+     *
+     * @param[in]  aConfig              A reference of the configuration to set.
```
jwhui:
```suggestion
     * @param[in]  aConfig. A reference of the configuration to set.
```

src/core/common/instance.hpp
```diff
@@ -73,6 +73,15 @@
 #if OPENTHREAD_CONFIG_CHANNEL_MONITOR_ENABLE
 #include "utils/channel_monitor.hpp"
 #endif
+
+#if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
+#include "backbone_router/leader.hpp"
```
jwhui: Add blank line before, to match blank line at end of block?

src/core/thread/mle.cpp
```diff
@@ -1560,6 +1594,25 @@ void Mle::HandleStateChanged(otChangedFlags aFlags)
...
+        if (Get<BackboneRouter::Local>().IsPrimary())
+        {
+            mBackboneRouterPrimaryAloc.GetAddress().SetLocator(kAloc16BackboneRouterPrimary);
+            mBackboneRouterPrimaryAloc.GetAddress().SetPrefix(GetMeshLocalPrefix());
+            Get<ThreadNetif>().AddUnicastAddress(mBackboneRouterPrimaryAloc);
+        }
+        else
+        {
+            Get<ThreadNetif>().RemoveUnicastAddress(mBackboneRouterPrimaryAloc);
+            mBackboneRouterPrimaryAloc.GetAddress().SetLocator(Mac::kShortAddrInvalid);
```
jwhui: Why not keep the locator always set to `kAloc16BackboneRouterPrimary`? Addition/removal of address should be sufficient, right?

src/core/thread/mle.hpp
```diff
@@ -730,9 +730,9 @@ class Mle : public InstanceLocator
     }
 
     /**
-     * This method retrieves the Service ALOC for given Service ID.
+     * This method retrieves the Service ALOC for given Service Id.
```
jwhui:
```suggestion
     * This method retrieves the Service ALOC for given Service ID.
```

src/core/thread/mle_types.hpp
```diff
@@ -210,21 +216,37 @@ enum AlocAllocation
...
     kAloc16CommissionerEnd             = 0xfc37,
+    kAloc16BackboneRouterPrimary       = 0xfc38,
     kAloc16CommissionerMask            = 0x0007,
     kAloc16NeighborDiscoveryAgentStart = 0xfc40,
     kAloc16NeighborDiscoveryAgentEnd   = 0xfc4e,
 };
 
 /**
- * Service IDs
+ * Service Ids
```
jwhui:
```suggestion
 * Service IDs
```

src/core/thread/network_data_leader.cpp
```diff
@@ -67,6 +67,92 @@ void LeaderBase::Reset(void)
...
 }
 
+otError LeaderBase::GetServiceId(uint32_t       aEnterpriseNumber,
+                                 const uint8_t *aServiceData,
+                                 uint8_t        aServiceDataLength,
+                                 bool           aServerStable,
+                                 uint8_t &      aServiceId)
+{
+    otError error = OT_ERROR_NOT_FOUND;
+
```
jwhui:
```suggestion
    otError error = OT_ERROR_NOT_FOUND;
```

src/core/thread/network_data_leader.hpp
```diff
@@ -229,6 +233,38 @@ class LeaderBase : public NetworkData
...
     otError GetRlocByContextId(uint8_t aContextId, uint16_t &aRloc16);
 
+    /**
+     * This method gets the Service Id for the specified service.
+     *
+     * @param[in]  aEnterpriseNumber  Enterprise Number (IANA-assigned) for Service TLV
+     * @param[in]  aServiceData       A pointer to the Service Data
+     * @param[in]  aServiceDataLength The length of @p aServiceData in bytes.
+     * @param[in]  aServerStable      The Stable flag value for Server TLV
+     * @param[out] aServiceId               A reference where to put the Service Id.
```
jwhui: nit: alignment

src/core/thread/network_data_leader.hpp
```diff
@@ -229,6 +233,38 @@ class LeaderBase : public NetworkData
...
+                         const uint8_t *aServiceData,
+                         uint8_t        aServiceDataLength,
+                         bool           aServerStable,
+                         uint8_t &      aServiceId);
+
+#if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
+    /**
+     * This method gets the Primary Backbone Router (PBBR) in the Thread Network.
+     *
+     * @param[out]  aConfig      A refernce where to put the Primary Backbone Router configuration.
```
jwhui:
```suggestion
     * @param[out]  aConfig      The Primary Backbone Router configuration.
```

src/core/thread/network_data_leader_ftd.cpp
```diff
@@ -751,7 +751,7 @@ otError Leader::RegisterNetworkData(uint16_t aRloc16, uint8_t *aTlvs, uint8_t aT
             rlocStable = true;
         }
 
-        // Store old Service IDs for given rloc16, so updates to server will reuse the same Service ID
+        // Store old Service Ids for given rloc16, so updates to server will reuse the same Service Id
```
jwhui:
```suggestion
        // Store old Service IDs for given rloc16, so updates to server will reuse the same Service ID
```

src/core/thread/network_data_tlvs.hpp
```diff
@@ -941,20 +946,20 @@ class ServiceTlv : public NetworkDataTlv
     }
 
     /**
-     * This method returns Service ID. It is in range 0x00-0x0f.
+     * This method returns Service Id. It is in range 0x00-0x0f.
```
jwhui:
```suggestion
     * This method returns Service ID. It is in range 0x00-0x0f.
```

src/core/thread/network_data_tlvs.hpp
```diff
@@ -1100,6 +1105,67 @@ class ServerTlv : public NetworkDataTlv
...
 } OT_TOOL_PACKED_END;
 
+OT_TOOL_PACKED_BEGIN
+class BackboneRouterServerData
+{
+public:
+    /**
+     * This method returns the sequence number of Backbone Router.
+     *
+     *  @returns  The sequence number of Backbone Router.
```
jwhui:
```suggestion
    *  @returns  The sequence number of the Backbone Router.
```

src/core/thread/network_data_tlvs.hpp
```diff
@@ -1100,6 +1105,67 @@ class ServerTlv : public NetworkDataTlv
...
+    /**
+     * This method sets the sequence number of Backbone Router.
+     *
+     *  @param[in]  aSequenceNumber  The sequence number of Backbone Router.
+     *
+     */
+    void SetSequenceNumber(uint8_t aSequenceNumber) { mSequenceNumber = aSequenceNumber; }
+
+    /**
+     * This method returns the Registration Delay (s) of Backbone Router.
```
jwhui:
```suggestion
     * This method returns the Registration Delay (in seconds) of Backbone Router.
```

tests/scripts/thread-cert/node.py
```diff
@@ -76,25 +82,45 @@ def __init__(self, nodeid, is_mtd=False, simulator=None, version=None):
 
     def __init_sim(self, nodeid, mode):
         """ Initialize a simulation node. """
-        if 'OT_CLI_PATH' in os.environ:
-            cmd = os.environ['OT_CLI_PATH']
-        elif (self.version == '1.1' and self.version != self.env_version):
+
+        # Default command if no match below, will be override if below conditions are met.
```
jwhui:
```suggestion
        # Default command if no match below, will be overridden if below conditions are met.
```

tests/scripts/thread-cert/v1_2_test_backbone_router_service.py
```diff
@@ -0,0 +1,235 @@
...
+            \      |
+              \    |
+                BBR_2
+
+ 1) Bring up Leader_1_1 and then BBR_1, BBR_1 becomes Primary Backbone Router.
+ 2) Reset BBR_1, if bring back soon, it could restore the Backbone Router Service
+    from the network, after increasing sequence number, it will reregister its
+    Backbone Router Service to the Leader and become Primary.
+ 3) Reset BBR_1, if bring back after it is released in the network, BBR_1 will
+    choose a random sequence number, register its Backbone Rouetr Service to
```
jwhui:
```suggestion
    choose a random sequence number, register its Backbone Router Service to
```

src/core/thread/mle.cpp
librasungirl: Yes,  it should also be sufficient. 
Previous thought in doing this way is that each addition/remove would iterate through the unicast address list to check whether or not it is contained by comparing address pointer. Here uses the last two bytes content to directly check whether or not the address is already added, which I thought might be a bit more efficient? Thoughts?

librasungirl: Yes, it should be sufficient and it might also be more simple.
Original thoughts is that as not per bbr state update would require addtion/removal for example from `disabled` to`secondary` or vice versa.
Here uses the last two bytes content to directly check whether or not the address is already added in advance to avoid unnecessary addition/removal as the unicast address list would always be iterated to check whether or not it is contained by comparing address pointer during addition/removal.

No strong opinion on this, I am fine to update to use the simpler way if you prefer it.

jwhui: I agree that execution cycles may be slightly reduced, but at the cost of increased code size. In general, I expect the addition/removal of addresses to occur relatively rarely. For that reason, I would have a preference for smaller code size. Thoughts?

librasungirl: 👍  make sense! will update.

src/cli/cli.cpp
```diff
@@ -435,6 +445,140 @@ void Interpreter::ProcessHelp(uint8_t aArgsLength, char *aArgs[])
     AppendResult(OT_ERROR_NONE);
 }
 
+#if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
+void Interpreter::ProcessBackboneRouter(uint8_t aArgsLength, char *aArgs[])
+{
+    otError                error = OT_ERROR_INVALID_COMMAND;
+    otBackboneRouterConfig config;
+    OT_UNUSED_VARIABLE(aArgs);
```
jwhui: nit: move to top with empty line after

src/core/backbone_router/leader.hpp
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * This method updates the cached Primary Backbone Router if any when new network data is available.
+     *
+     */
+    void Update(void);
+
+    /**
+     * This method gets the Primary Backbone Router in the Thread Network.
+     *
+     * @param[out]  aConfig        A reference whether to put the Primary Backbone Router information.
```
jwhui:
```suggestion
     * @param[out]  aConfig        The Primary Backbone Router information.
```

src/core/backbone_router/leader.hpp
```diff
@@ -0,0 +1,146 @@
...
+     * @retval OT_ERROR_NONE       Successfully got the Primary Backbone Router information.
+     * @retval OT_ERROR_NOT_FOUND  No Backbone Router in the Thread Network.
+     *
+     */
+    otError GetConfig(BackboneRouterConfig &aConfig) const;
+
+    /**
+     * This method gets the Backbone Router Service ID.
+     *
+     * @param[out]  aServiceId     The reference whether to put the Backbone Router Service ID.
```
jwhui:
```suggestion
     * @param[out]  aServiceId     The Backbone Router Service ID.
```

src/core/backbone_router/leader.hpp
```diff
@@ -0,0 +1,146 @@
...
+     * @retval OT_ERROR_NONE       Successfully got the Backbone Router Service ID.
+     * @retval OT_ERROR_NOT_FOUND  Backbone Router service doesn't exist.
+     *
+     */
+    otError GetServiceId(uint8_t &aServiceId) const;
+
+    /**
+     * This method gets the short address of the Primary Backbone Router.
+     *
+     * @retval short address of Primary Backbone Router, or Mac::kShortAddrInvalid if no Primary Backbone Router.
```
jwhui:
```suggestion
     * @returns short address of Primary Backbone Router, or Mac::kShortAddrInvalid if no Primary Backbone Router.
```

PR #4437, opened 2019-12-31
[mle] fix build error caused by GetParent return type change

gjc13:
The GetParent return type has been updated in PR#4419. Update
accordingly.


bukepo:
@gjc13 it seems there are some memory issues.

bukepo:
@abtink `StartParentSearchTimer()` is moved into `Enable()`. Any concerns?

abtink:
@bukepo I think it should be ok. LGTM.

PR #4104, opened 2019-08-18
[border-agent] fix appending extended UDP Encapsulation TLV

wgtdkp:
This PR fix the bug that border agent incorrectly appends the UDP Encapsulation TLV:
- _AppendTlv_ does not accept an Extended TLV;
- _AppendTlv_ expects a complete TLV object which includes the _value_ just behind the _length_.

PR #4438, opened 2019-12-31
[tlv] add AppendTo to resolve mutual dependency

bukepo:
This PR adds AppendTo(Message) to Tlv, so that Message doesn't depend
on Tlv any more. This allows a concrete Tlv overriding its
serialization. This was discussed in #4104.

### DIFF
--- a/src/core/common/message.cpp
+++ b/src/core/common/message.cpp
@@ -402,15 +402,6 @@ exit:
     return error;
 }
 
-otError Message::AppendTlv(const Tlv &aTlv)
-{
-    uint32_t size = aTlv.GetSize();
-
-    assert(size <= UINT16_MAX);
-
-    return Append(&aTlv, static_cast<uint16_t>(size));
-}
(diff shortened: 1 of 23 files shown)


abtink:
Thanks. This looks good. I like it mainly for removing dependency of `Message` to `Tlv`.

Regarding allowing `Tlv` sub-classes to override how they serialize (other than extended TLV), do we have a use-case for it? 

From PR https://github.com/openthread/openthread/pull/4152 the `Tlv` class itself now takes care of it being `Tlv` or `ExtendedTlv` (which makes its methods safe and correct under different cases).

bukepo:
One use case is overloading `UdpEncapsulationTlv::AppendTo()` with a additional parameter `aPayload` so that this method can be used in multiple places. However, I tend to do this later when native border agent is added.

abtink:
@bukepo , sounds good. Thanks.

PR #4439, opened 2020-01-01
[ncp] add timestamp to log metadata fields

konradderda:
This change adds an information about the timestamp to the log metadata that calculated with the formula: <timestamp_base> + <time_from_start_in_ms>. The <timestamp_base> value is set to 0 by default and can be changed with a new SPINEL property: SPINEL_PROP_DEBUG_LOG_TIMESTAMP_BASE formatted as X (uint64_t).


konradderda:
I have added a complementary PR to the wpantund:
https://github.com/openthread/wpantund/pull/452

src/ncp/ncp_base.cpp
```diff
@@ -2257,6 +2259,19 @@ template <> otError NcpBase::HandlePropertySet<SPINEL_PROP_DEBUG_NCP_LOG_LEVEL>(
 }
 #endif // OPENTHREAD_CONFIG_LOG_LEVEL_DYNAMIC_ENABLE
 
+template <> otError NcpBase::HandlePropertySet<SPINEL_PROP_DEBUG_LOG_TIMESTAMP_BASE>(void)
```
abtink: How about adding a getter as well?

abtink:
@LuDuda I like the idea of adding support for tracking a `uint64_t` clock/time (e.g., ms since start or since epoch) within OT core itself (maybe from a seperate PR?). We can then use this as the timestamp in the logs emitted by NCP.

jwhui:
@konradderda, thanks for updating this PR! Can you help resolve the conflicts as well?

src/ncp/ncp_base.cpp
```diff
@@ -2257,6 +2259,32 @@ template <> otError NcpBase::HandlePropertySet<SPINEL_PROP_DEBUG_NCP_LOG_LEVEL>(
...
+{
+    uint64_t timestampBase = 0;
+    otError  error         = OT_ERROR_NONE;
+    uint32_t currentTime   = otPlatAlarmMilliGetNow();
+
+    SuccessOrExit(error = mDecoder.ReadUint64(timestampBase));
+
+    if (timestampBase >= currentTime)
```
LuDuda: Super nit picking:use `VerifyOrExit` here instead?

src/ncp/ncp_base_dispatcher.cpp
```diff
@@ -342,7 +342,7 @@ NcpBase::PropertyHandler NcpBase::FindGetPropertyHandler(spinel_prop_key_t aKey)
       {SPINEL_PROP_DEBUG_TEST_ASSERT, &NcpBase::HandlePropertyGet<SPINEL_PROP_DEBUG_TEST_ASSERT>},
       {SPINEL_PROP_DEBUG_NCP_LOG_LEVEL, &NcpBase::HandlePropertyGet<SPINEL_PROP_DEBUG_NCP_LOG_LEVEL>},
       {SPINEL_PROP_DEBUG_TEST_WATCHDOG, &NcpBase::HandlePropertyGet<SPINEL_PROP_DEBUG_TEST_WATCHDOG>},
-    };
+      {SPINEL_PROP_DEBUG_LOG_TIMESTAMP_BASE, &NcpBase::HandlePropertyGet<SPINEL_PROP_DEBUG_LOG_TIMESTAMP_BASE>} };
```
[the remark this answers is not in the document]
konradderda: I wanted to but 'make pretty' has the last word here :)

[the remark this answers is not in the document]
LuDuda: Please try to make "," comma at the end ;)

src/ncp/ncp_base.cpp
konradderda: Done

src/ncp/ncp_base_dispatcher.cpp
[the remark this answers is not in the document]
konradderda: Done

src/ncp/ncp_base.cpp
konradderda: Good point. Done.

konradderda:
To summarize:
- All doubts of @LuDuda are addressed
- I also think that the addition of uint64_t timings on the platform level would be "nice to have" feature but with the current width, we still have around 50 days for this timer to overflow so I think that is something that may be handled by a user. I think that it would be good to discuss possible solutions in a separate issue.

@jwhui Merge conflicts are resolved now.

PR #4440, opened 2020-01-01
[radio-selector] multi radio and Thread Radio Encapsulation Link (TREL) 

abtink:
This PR includes a group of related commits adding support for **_multi radio link_** and 
**_Thread Radio Encapsulation Link_**  and related Travis and toranj tests.

**[radio-selector] adding multi radio link support**

This commit adds support for multi radio links in OpenThread core.

A set of `OPENTHREAD_CONFIG_RADIO_LINK_*` definitions (in header `config/radio_link.h`) determines the supported radio link types.

This commit adds a new class `RadioSelector` which selects the radio link for each message transmission. Per neighbor it tracks the supported radio types and a preference value for each type. `RadioSelector` updates the preference value based on the history of rx/tx events with the neighbor, e.g., a successful tx on a radio link increases the preference, whereas a failed tx attempt decreases it.

A new class `Mac::Links` is added defining a layer between `Mac` and different radio links (e.g., `SubMac` for 802.15.4 radio type). Broadcast frames are sent in parallel over all radio links. A unicast transmission is sent over a single radio link at a time.

This commit also adds the concept of deferred ack, allowing radio links report status of ack at a later time through a different callback.

**[trel] implementing Thread Radio Encapsulation Link (TREL)**

This commit implements TREL protocol to encapsulate and exchange IEEE 802.15.4 frames over a packet switched network (e.g., a wireless LAN). This commit also provides a TREL IPv6/UDP interface implementation (using link-local IPv6 unicast/multicast addresses to communicate over a subnet). It defines a platform abstraction (`otPlatTrelUdp6*`) for TREL over IPv6/UDP.

This commit also provides an implementation of the TREL platform under `examples/posix` focusing on simulating the behavior of the link.

**[posix-app] adding support for TREL over IPv6/UDP**

This commit implements TREL IPv6/UDP platform under POSIX App. The implementation uses datagram sockets to exchange packets over a given network interface (netif). The interface name can hard-coded in the build using `OPENTHREAD_CONFIG_POSIX_APP_TREL_INTERFACE_NAME` config or be specified as part of the input program arguments (using `--trel-interface` or `-t` option).

**[toranj] add support for multi radio and TREL**

This commit updates `toranj` test framework to support running tests in posix-app mode over a TREL radio link (a virtual Ethernet netif is used for TREL). It also updates the build script to easily set up different configurations, ncp/posix-app with any radio combination: 15.4 only, TREL only, or multi radio (15.4 + TREL).

**[mesh-forwarder] limit the Mesh Header frame MTU**

This commit allows us to use different MTU on different radio links. Mesh Header frames are forwarded by routers over multiple hops to reach a final destination. The forwarding path can have routers supporting different radio links with varying MTU sizes. Since the originator of the frame does not know the path and the MTU sizes of supported radio links by the routers in the path, we limit the max payload length of a Mesh Header frame to a fixed minimum value (derived from 15.4 radio) ensuring it can be handled by any radio link.

**[multi-radio] adding new public OT APIs for multi radio link config**

This commit adds new public OT definitions and API for radio link configuration in `openthread/multi_radio.h`. Mainly the function `otMultiRadioGetNeighborInfo()` is added which allows info about a neighbor to be retrieved (e.g., which radio links are supported by the neighbor and their preference value).

**[ncp] add new properties for multi radio and TREL support**

This commit adds new spinel properties (and their getters) related to multi radio link feature. `SPINEL_PROP_SUPPORTED_RADIO_LINKS` allows host to get the list of supported radio links by the device itself, `SPINEL_PROP_NEIGHBOR_TABLE_MULTI_RADIO_INFO` provides info about the radio links supported by all neighbors.

**[toranj] support for hybrid (multi radio) topology and test cases**

This commit extends toranj framework to allow tests for networks consisting of nodes supporting different combinations of radio links. Five new tests are added to cover multi radio behavior:

- test-700: attaching process between nodes with 15.4-only, trel-only and 15.4+trel combo and device reset recovery
- test-701: behavior of device after trel network is temporarily disabled and re-discovery of trel radio using probe mechanism.
- test-702: behavior of device after trel network is temporarily  disabled and re-discovery of trel radio through frame rx.
- test-703: "mesh header" messages exchange (over multi-hop) when the underlying devices (in the path) support different radio types with different frame MTU length.
- test-704/705: active/discover scan with nodes supporting different radios

This commit also adds a new travis test case covering all the above `toranj` tests.

-----

Two smaller related commits:
- **[topology] adding Neighbor::StateToString()**
- **[mle-router] adding FindNeighbor() to find a neighbor matching a state filter**

-----

Special thanks and credit to @bukepo for the original concept/idea and the implementation of a first prototype and @jwhui for a lot of great suggestions.


include/openthread/platform/trel-udp6.h
```diff
@@ -0,0 +1,137 @@
...
+#include <openthread/error.h>
+#include <openthread/instance.h>
+#include <openthread/ip6.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @addtogroup plat-trel
```
Vyrastas: Also please add this to the appropriate place in the Doxygen group hierarchy in include/openthread/error.h, if you could.

abtink:
Updated the PR's description itself to add the new pushed commits. In this note,  just summarizing the main changes/additions compared to previous version:

- The core and `RadioSelector` now support deferred ack model (for TREL) - this allows TREL over WiFi with clients operating in power-save mode (i.e, can work even with long packet exchange delay).
- `RadioSelector` now implements the probing mechanism which helps discover if a previously available radio link has again become available to use.
- `Mac::Links` and `Mac::TxFrames` now allow for parallel tx over multiple radio links (e.g., this is used for probing)
- `Trel` ack mechanism and packet format are changed (mainly to enable deferred ack model).
- Spinel and NCP code are updated to support multi-radio feature (e.g., get supported radio link info about all neighbors).
- `toranj` test framework is updated to support hybrid multi-radio topologies (nodes with different set of supported radio in the same test) . Also 6 multi-radio tests are added and ran in Travis CI.

src/core/mac/mac.cpp
```diff
@@ -1241,13 +1299,70 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
...
     }
+#endif // #if OPENTHREAD_CONFIG_RADIO_LINK_IEEE_802_15_4_ENABLE
+
+#if OPENTHREAD_CONFIG_MULTI_RADIO
+    if (!aFrame.IsEmpty())
+    {
+        RadioType  radio          = aFrame.GetRadioType();
+        RadioTypes requriedRadios = mLinks.GetTxFrames().GetRequiredRadioTypes();
+
+        Get<RadioSelector>().UpdateOnSendDone(aFrame, aError);
```
bukepo: Could we consider just selecting radio when establish a link? Thus, Thread core doesn't need to keep track all radio supported by itself and peers, and we might have a simpler solution. And we can also make switching to a better radio as an optional feature just like the better parent feature.

src/core/thread/radio_selector.hpp
```diff
@@ -0,0 +1,217 @@
...
+        /**
+         * This method retrieves the multi radio information `otMultiRadioNeighborInfo` associated with the neighbor.
+         *
+         * @param[out] aInfo  A reference to `MultiRadioInfo` to populate with neighbor info.
+         *
+         */
+        void PopulateMultiRadioInfo(MultiRadioInfo &aInfo);
+
+    private:
+        void AddSupportedRadioType(Mac::RadioType aType) { mSupportedRadioTypes.Add(aType); }
```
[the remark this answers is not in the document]
abtink: It's practically the same. Core module (other than maybe `radio selector`) treat the `RadioType` as an id that needs to be passed around with messages/frames (we just define it as an `enum` to have a specific type for it and  have a list of valid values).

src/core/mac/mac.cpp
abtink: It is a possible different approach but I feel more limiting (in the longer run).
We need to somehow track which neighbor support which radio to enable radio switch recovery feature. Having it in core and single module `RadioSelector` is not that complex and  allows us to send over multiple radios either in parallel (e.g., for broadcast and/or probe mechanism to discover/track new supported radios) or sequentially (e.g. if a tx over a radio fails and there are other radio options supported by device and destination neighbor, we can consider retrying the message/frame on another radio).

abtink:
The new push rebases this PR (same implementation as before) on current master. 
The travis tests are also migrated to GH actions.

abtink:
New pushed commit implements new behaviors/features
- New commit for HMAC Key Derivation Function (HKDF) using SHA-256. Used to create keys for TREL link (and a unit test for it).
- Different MAC frame counters per radio link (they are synced when saved in non-volatile or exchanged in an MLE message)
- Neighbor track the MAC frame counters per supported radio links and rx frames use the associated key and counter (per radio link)
- New mechanism to detect/suppress duplicate frames (rx over multiple radio links)
  - Adding first lowpan frag header to secure frame when multi-link is enabled
  - Tracking last rx frag tag per neighbor  (updated on neighbor reset and rx of specific MLE msg).
  - Using frag tag to detect/suppress duplicate rx frames.

Will be adding more test-cases to try to cover some corner case situations.

abtink:
Rebased on master (to fix conflicts).

abtink:
The new push contains rebased/modified implementation on current master.

There are multiple commits, each containing separate (but related) components of the overall  implementation: 
- HKDF along with unit test for it (key derivation).
-  multi-radio support in core (radio-selector)
- Trel protocol implementation (along with simulation support)
- Posix Trel implementation (for product use)
- Tests and CI (toranj and gh-action builds)

Main changes in the recent push are:
- How the sec frame counter (for all radio links) is managed 
- Addressing conflicts related to CSL implementation.
- Use new helper methods, from `NeighborTable`, i.e. `FindNeighbor()`, or new `Ip6::Address` helpers.
- Smaller changes (use `nullptr`, etc).

Irving-cl (commented):
Sorry, wrong operation here.

include/openthread/platform/trel-udp6.h
```diff
@@ -0,0 +1,154 @@
...
+ * This function updates the unicast IPv6 address for TREL IPv6/UDP interface.
+ *
+ * The interface should only have one unicast IPv6 address. Calling this function replaces any previously set unicast
+ * IPv6 address (during initialization from `otPlatTrelUdp6Init` or earlier calls to `otPlatTrelUdp6UpdateAddress()`).
+ *
+ * @param[in] aInstance        The OpenThread instance structure.
+ * @param[in] aUnicastAddress  The unicast address to add to interface and use for as tx source and rx destination.
+ *
+ */
+void otPlatTrelUdp6UpdateAddress(otInstance *aInstance, const otIp6Address *aUnicastAddress);
```
Irving-cl, line 82: Should we also allow the Udp Port to be updated in this API?

abtink: The port number would be practically constant/same throughout device operation, so we set it from init function (no use-case/situation where we would need to change it afterwards).  The address however can potentially change during operation so we need a platform API to be able to change it.

src/core/radio/trel_link.cpp
```diff
@@ -0,0 +1,489 @@
...
+
+exit:
+    return;
+}
+
+void Link::SendAck(Packet &aRxPacket)
+{
+    Packet ackPacket;
+
+    ackPacket.Init(mAckPacketBuffer, sizeof(mAckPacketBuffer));
```
zhanglongxia, line 424: Why do we define a new ACK instead of using 15.4 ACK？

src/core/radio/trel_packet.hpp
```diff
@@ -0,0 +1,373 @@
...
+    };
+
+    // All header fields are big-endian.
+
+    uint8_t         mControl;
+    uint8_t         mChannel;
+    uint16_t        mPanId;
+    uint32_t        mPacketNumber;
+    Mac::ExtAddress mSource;
+    Mac::ExtAddress mDestination; // Present in `kTypeAck` or `kTypeUnicast` packet types.
```
zhanglongxia, line 255: The 15.4 frame has already included the panid, source and destination address. Why do we define these fields in the  TREL radio packet?

include/openthread/multi_radio.h
```diff
@@ -0,0 +1,103 @@
...
+#define OPENTHREAD_MULTI_RADIO_H_
+
+#include <openthread/platform/radio.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @addtogroup api--multi-radio
```
Vyrastas, line 45: Should be api-multi-radio

src/core/common/message.hpp
```diff
@@ -174,8 +174,14 @@ struct MessageMetadata
     bool    mDoNotEvict : 1;   ///< Indicates whether or not this message may be evicted.
 #if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
     bool    mTimeSync : 1;      ///< Indicates whether the message is also used for time sync purpose.
-    uint8_t mTimeSyncSeq;       ///< The time sync sequence.
     int64_t mNetworkTimeOffset; ///< The time offset to the Thread network time, in microseconds.
+    uint8_t mTimeSyncSeq;       ///< The time sync sequence.
+#endif
+#if OPENTHREAD_CONFIG_MULTI_RADIO
+    uint8_t mRadioType : 2;      ///< The radio link type the message was received on, or should be sent on.
```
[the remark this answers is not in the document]
abtink: Adding a new `Invalid` enumerator to `RadioType` enum can potentially help only here with the `MessageMetaData` but it adds the overhead that in all other places where the `RadioType` enum values are used/compared, we would need to check against an extra enumerator value. It is easier/better to handle it directly here over adding overhead everywhere else. 

Also in this case (in `MessageMetadata`), it's not that the radio type is invalid, but that the radio link is not yet determined (for an outbound message) and therefore it is not yet set on the message.

src/core/mac/mac_links.cpp
```diff
@@ -0,0 +1,207 @@
...
+#if OPENTHREAD_CONFIG_RADIO_LINK_IEEE_802_15_4_ENABLE
+    if (aRadioTypes.Contains(kRadioTypeIeee802154) && mTxFrames.mTxFrame802154.IsEmpty())
+    {
+        mTxFrames.mTxFrame802154.CopyFrom(aFrame);
+    }
+#endif
+#if OPENTHREAD_CONFIG_RADIO_LINK_TREL_ENABLE
+    if (aRadioTypes.Contains(kRadioTypeTrel) && mTxFrames.mTxFrameTrel.IsEmpty())
+    {
+        mTxFrames.mTxFrameTrel.CopyFrom(aFrame);
```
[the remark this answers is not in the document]
abtink: No, there would be no duplicate copy. 

- In the code here we check if the frame (e.g. `mTxFrameTrel`) is empty before we copy (if it's already prepared by layers above we don't do anything).
- The code in the `mac.cpp` is used only when the frame requires link-layer security. It's then copying the frame and processing security for each radio types on which the frame would need to be sent. Note that the security processing would be different for each frame (different keys and counters).
- For frame that don't use link/mac security (e.g. MLE message, active scan, etc) MAC layer would not need to make any copy and it is up to sub-MAC.

src/core/thread/radio_selector.cpp
```diff
@@ -0,0 +1,420 @@
...
+    // `aTxFrames`. We allow probe on TREL radio link if it is not
+    // currently usable (thus not selected) but is/was supported by
+    // the neighbor (i.e., we did rx/tx on TREL link from/to this
+    // neighbor in the past). The probe process helps detect whether
+    // the TREL link is usable again allowing us to switch over
+    // faster.
+
+#if OPENTHREAD_CONFIG_RADIO_LINK_TREL_ENABLE
+    if (!selections.Contains(Mac::kRadioTypeTrel) && neighbor->GetSupportedRadioTypes().Contains(Mac::kRadioTypeTrel) &&
+        (Random::NonCrypto::GetUint8InRange(0, 100) < kTrelProbeProbability))
```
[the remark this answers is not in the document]
abtink: We would like to enable this in future, but I intentionally left it out for now from this PR (to reduce its complexity)
- The probing on 15.4 is much less useful, since if the neighbor link supports both TREL/WiFi and 15.4, we would in general prefer TREL for frame tx.
- We have the receive mechanism to discover the radio linkes (e.g. if/when a 15.4 link becomes available). For a router neighbor MLE Link Adv would be send on all radio links and potentially received by neighbors on multiple radio links as well). 
- Probing is mainly intended for cases where we are actively sending a lot of data/frames and want to quickly discover if a better radio link (TREL/WIFi in this case) has become available to switch to it quickly.

There are some other points to consider for probing on 15.4 links:
- When probing on 15.4 we don't want to use the same max tx retries (e.g., we use 15 tx attempts for direct tx). 
- Otherwise if the 15.4 link is not available, we would spend/waste a lot of time retx-ing the frame 15 times over the 15.4 link and waiting for ack timeout.
- So for this, we need to use mechanism to set a different tx retry count for probe frame vs normal data per radio type (we have this behavior in sub-MAC layer but need to make sure the radio platform also supports this to cover the cases where radio platform layer is handling the frame retx logic).

src/core/thread/radio_selector.cpp
```diff
@@ -0,0 +1,420 @@
...
+#endif
+
+exit:
+    return aTxFrames.GetTxFrame(selections);
+}
+
+Mac::RadioType RadioSelector::SelectPollFrameRadio(const Neighbor &aParent)
+{
+    // This array defines the order in which different radio link types
+    // are selected for data poll frame tx.
```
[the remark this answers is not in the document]
abtink: We do this for data frame from `SelectRadio`.
But data polls are intended for energy-constrained devices (sleepy devices). So we prefer the 15.4 over TREL/WiFi for such frames. In general, I don't think having a SED with WiFi/TREL radio is practical or useful.

src/core/mac/mac.cpp
```diff
@@ -1349,7 +1581,12 @@ void Mac::RecordFrameTransmitStatus(const TxFrame &aFrame,
 
 void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError)
 {
-    if (!aFrame.IsEmpty())
+#if OPENTHREAD_CONFIG_RADIO_LINK_IEEE_802_15_4_ENABLE
```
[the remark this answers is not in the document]
abtink: Thanks. Yes, we can. Will change it in next push.

For longer term, I like to actually move the broadcast retx logic to sub-MAC layer (since we already do all other retx in sub-MAC or radio). This way it would be directly tied to 15.4 link.

src/core/mac/mac_frame.cpp
```diff
@@ -970,25 +972,130 @@ const TimeIe *Frame::GetTimeIe(void) const
...
+#elif !OPENTHREAD_CONFIG_RADIO_LINK_IEEE_802_15_4_ENABLE
+
+uint16_t Frame::GetMtu(void) const
+{
+    return
+#if OPENTHREAD_CONFIG_RADIO_LINK_TREL_ENABLE
+        Trel::Link::kMtuSize;
+#elif OPENTHREAD_CONFIG_RADIO_LINK_TOBLE_ENABLE
+        Toble::Link::kMtuSize;
+#endif
```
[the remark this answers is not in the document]
abtink: This is actually intentional. We want the `GetMtu()` to be inlined and threrefore defined in the header file in the simple and common case where we only have 15.4 radio link (this is to avoid code size overhead in the simple case).

Note that  moving the implementation to header in general case is not as easy/straight-forward since it creates a header dependency loop:
- We would need to include `trel_link.hpp` in `mac_frame.hpp` to see the `Trel::Link::kMtuSize` constant
- At the same time `trel_link.hpp` would need to see full definitions of `Mac::RxFrame` and `Mac::TxFrame` so it needs to include `mac_frame.hpp`.

src/core/radio/trel_link.cpp
abtink: TREL ack packet serves a very different purpose/functionality compared to 15.4 ACK.
- It is used to indicate whether TREL packet was received by a known neighbor (to help update the neighbor radio link state).
- The TREL ack can be requested independent of whether the embedded/encapsulated 15.4 frame within the TREL packet requires ack or not.
    - For example, this is helpful when converting a broadcast 15.4 frame to a bunch of unicast TREL packets sent to each neighbor (the original 15.4 frame does not require ack but the TREL packets do).
    - We require  a TREL ack in response to a TREL packet destined to a known neighbor (even if the embedded 15.4 frame does not require ack).
- The TREL header include a "TREL packet number" which is used to associate a TREL packet with its TREL ack.
   - The packet number is different per neighbor and is tracked/managed by the TREL link layer.
   - The packet number is used in deferred ack mechansim which is used by TREL to ensure it can be used on WiFi links using 802.11 power-save mode.
   - 15.4 ack frame on the other hand has a rather limited format since it only contains a seq number byte value.

src/core/mac/mac_links.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    mSelectedRadioTypes.Add(aRadioType);
+
+    return *frame;
+}
+
+TxFrame &TxFrames::GetTxFrame(RadioTypes aRadioTypes)
+{
+    // Return the TxFrame among all set of `aRadioTypes` with the smallest MTU.
+
```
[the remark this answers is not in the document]
abtink: Will add a comment in the next push.

Note that this is a `TxFrame`. Basically we want to prepare a frame to send over a subset of radio links in parallel, so we need to make sure to prepare the frame such that it fits in the most restricted radio (i.e., choose the `TxFrame` for the radio link with smallest supported MTU size).

src/core/radio/trel_packet.hpp
abtink: Note that we do need a TREL header to include a version and TREL packet number. We considered whether to include these fields (PAN Id, src and dest) in the TREL header as well or parse them from embedded 15.4 frame. 

At the end, decided to include these and have a clear and separate (well-defined) TREL header. There are multiple benefits to this. It not only helps simplify the code, it also adds flexibility to the design:
- On receive side we just need to parse the TREL header to filter TREL packet. No need to look into the encapsulated 15.4 frame.
- It adds flexibility since we can then embed any type of 15.4 frame into TREL packet, even the ones that may not have the src/dest or PAN Id info
   - For example, we can embed a 15.4 ack frame into TREL packet (note that ack frame does not contain PAN Id and/or src/dest) and ensure it is correctly filtered at the receiver at TREL link layer.
- In a 15.4 frame the src/dest adress may or may not be present and can be either short or extended. In TREL layer we would like to know the source and destination as extended addresses (since these will directly map to the (IPv6) addresses in TREL interface layer).
    - Having them as extended address in the TREL packet header would help simplify the TREL interface layer code (it does not need to look into encapsulated 15.4 frame again and be aware of neighbor table to convert from short to extended address).
   - It helps address some corner cases where it gets tricky to handle otherwise
       - We can receive a TREL packet (requiring a TREL ack) from a node which we don't yet have in our neighbor table. Having the src/dest as extended address in the TREL packet header decouples sending of TREL ack (we can prepare and send it immediately even if the neighbor is not yet added in neighbor table).
- The only benefit for not including these fields in the TREL header is to reduce the overall TREL packet size, but since TREL is run over infrastructure link with high throughput and high PSDU, optimizing few bytes does not matter much.

src/core/mac/mac_frame.hpp
```diff
@@ -950,21 +949,53 @@ class Frame : public otRadioFrame
...
      * This method returns the FCS size.
      *
      * @returns This method returns the FCS size.
      *
      */
-    uint16_t GetFcsSize(void) const { return kFcsSize; }
+    uint8_t GetFcsSize(void) const
+#if !OPENTHREAD_CONFIG_MULTI_RADIO && OPENTHREAD_CONFIG_RADIO_LINK_IEEE_802_15_4_ENABLE
+    {
+        return sizeof(uint16_t);
```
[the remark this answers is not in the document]
abtink: Good suggestion. Will change it in next push.

include/openthread/multi_radio.h
abtink: Thanks. Will change it in the next push.

src/core/thread/key_manager.cpp
```diff
@@ -66,6 +67,18 @@ const otMasterKey KeyManager::kDefaultMasterKey = {{
     0xff,
 }};
 
+#if OPENTHREAD_CONFIG_RADIO_LINK_TREL_ENABLE
+const uint8_t KeyManager::kHkdfExtractSaltString[] = {'T', 'h', 'r', 'e', 'a', 'd', 'S', 'e', 'q', 'u', 'e', 'n',
+                                                      'c', 'e', 'M', 'a', 's', 't', 'e', 'r', 'K', 'e', 'y'};
+
+const uint8_t KeyManager::kTrelInfoString[] = {'T', 'h', 'r', 'e', 'a', 'd', 'O', 'v', 'e',
+                                               'r', 'W', 'i', 'F', 'i', 'K', 'e', 'y'};
```
jwhui, line 75: nit: change to ThreadOverInfraKey?

src/core/thread/radio_selector.cpp
```diff
@@ -0,0 +1,391 @@
...
+    uint8_t        selectedPreference = 0;
+    bool           found              = false;
+
+    // Select the first radio links with preference higher than
+    // threshold `kHighPreference`. The radio links are checked in the
+    // order defined by the `sRadioSelectionOrder` array. If no radio
+    // link has preference higher then threshold, select the one with
+    // highest preference.
+
+    for (uint8_t index = 0; index < OT_ARRAY_LENGTH(sRadioSelectionOrder); index++)
```
jwhui, line 245: nit: change to range-based for loop?

src/core/thread/radio_selector.cpp
```diff
@@ -0,0 +1,391 @@
...
+                        const char *    aActionText,
+                        Mac::RadioType  aRadioType,
+                        const Neighbor &aNeighbor)
+{
+    String<kRadioPreferenceStringSize> preferenceString;
+    bool                               isFirstEntry = true;
+
+    VerifyOrExit(otLoggingGetLevel() >= aLogLevel, OT_NOOP);
+
+    for (uint8_t index = 0; index < OT_ARRAY_LENGTH(sRadioSelectionOrder); index++)
```
jwhui, line 388: nit: change to range-based for loop?

src/core/mac/mac.cpp
```diff
@@ -1166,47 +1351,94 @@ void Mac::BeginTransmit(void)
...
-        ProcessTransmitSecurity(sendFrame);
+#if OPENTHREAD_CONFIG_MULTI_RADIO
+        // Go through all selected radio link types for this tx and
+        // copy the frame into correct `TxFrame`  for each radio type
+        // (if it is not already prepared), then process security for
+        // each radio type separately. This allows radio links to
+        // handle security differently, e.g., with different keys or
+        // link frame counters.
+
+        for (uint8_t index = 0; index < OT_ARRAY_LENGTH(RadioTypes::kAllRadioTypes); index++)
```
jwhui, line 1376: nit: change to range-based for loop?

src/core/thread/radio_selector.cpp
```diff
@@ -0,0 +1,391 @@
...
+            aNeighbor);
+
+        // In case of deferred ack timeout, we check if the neighbor
+        // has any other radio link (with high preference) for future
+        // tx. If it it does, we set `aAllowNeighborRemove` to `false`
+        // to ensure neighbor is not removed yet.
+
+        VerifyOrExit(aTxError != OT_ERROR_NONE, OT_NOOP);
+
+        for (const Mac::RadioType *radio = &sRadioSelectionOrder[0]; radio < OT_ARRAY_END(sRadioSelectionOrder);
```
jwhui, line 208: nit: change to range-based for loop?

src/core/thread/radio_selector.cpp
```diff
@@ -0,0 +1,391 @@
...
+        Mac::kRadioTypeToble,
+#endif
+#if OPENTHREAD_CONFIG_RADIO_LINK_TREL_ENABLE
+        Mac::kRadioTypeTrel,
+#endif
+    };
+
+    Mac::RadioType selection = selectionOrder[0];
+
+    for (uint8_t index = 0; index < OT_ARRAY_LENGTH(selectionOrder); index++)
```
jwhui, line 360: nit: change to range-based for loop?

src/core/radio/trel_link.cpp
```diff
@@ -0,0 +1,489 @@
...
+    }
+}
+
+// LCOV_EXCL_START
+
+const char *Link::StateToString(State aState)
+{
+    const char *str = "Unknown";
+
+    switch (aState)
```
jwhui, line 462: nit: use static const char * array?

src/core/radio/trel_packet.hpp
```diff
@@ -0,0 +1,373 @@
...
+ * This class represent a TREL radio link packet.
+ *
+ */
+class Packet
+{
+public:
+    /**
+     * This method initializes the `Packet` with a given buffer and length.
+     *
+     * @param[in] A pointer to a buffer containing the entire packet (header and payload).
```
jwhui, line 268: nit: missing param names

src/core/radio/trel_packet.hpp
```diff
@@ -0,0 +1,373 @@
...
+    uint16_t GetLength(void) const { return mLength; }
+
+    /**
+     * This method validates the packet header.
+     *
+     * @retval OT_ERROR_NONE    Successfully parsed and validated the packet header.
+     * @retval OT_ERROR_PARSE   Failed to parse the packet header
+     *
+     */
+    otError ValidateHeader(void) const;
```
jwhui, line 321:
```suggestion
    bool IsHeaderValid(void) const;
```

abtink:
Thanks for the reviews. I will first rebase the existing code on current master (to address all the conflicts) and ensure everything builds/works as before. Then update the code based on the review comments.

abtink:
New pushed commits are rebased on master and also address review comments.

The recent push  also removes the TOBLE related definitions in multi-radio implementation within OT core, NCP, and public APIs. We can add support for it if/when we have a solution/use for ToBLE (the multi-radio implementation allows for adding support for more radio-link types in future and is extendable)

abtink:
There is a new wpanund PR https://github.com/openthread/wpantund/pull/475 corresponding to this PR (the new one  also removes TOBLE related definition (from spinel)). The last commit in this PR is also updated to point to the wpantund branch from the above PR.

bukepo (approved):
LGTM:+1:

Irving-cl (approved):
LGTM

zhanglongxia (approved):
LGTM

jwhui (approved):

PR #4443, opened 2020-01-02
[network-data] fix compile error when building service without border router

dismirlian:


src/core/thread/network_data_local.cpp
```diff
@@ -370,8 +370,11 @@ otError Local::SendServerDataNotification(void)
 #if OPENTHREAD_CONFIG_BORDER_ROUTER_ENABLE
         !IsOnMeshPrefixConsistent() || !IsExternalRouteConsistent()
 #endif
+#if OPENTHREAD_CONFIG_BORDER_ROUTER_ENABLE && OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE
```
jwhui: Thanks for identifying this issue and proposing a fix! Would it be cleaner to simply break this `VerifyOrExit()` up into two separate statements to simplify the `#if`s?

dismirlian: Hi @jwhui, I agree, much cleaner.

dismirlian:
I'll close and re-open to kick the failed Travis CI again.

src/core/thread/network_data_local.cpp
jwhui: @dismirlian, I think you want something like:
```c++
    bool isConsistent =	true;

#if OPENTHREAD_CONFIG_BORDER_ROUTER_ENABLE
    isConsistent &= IsOnMeshPrefixConsistent() && IsExternalRouteConsistent();
#endif
#if OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE
    isConsistent &= IsServiceConsistent();
#endif

    VerifyOrExit(!isConsistent, ClearResubmitDelayTimer());
```
This should help resolve the Travis test failures.

dismirlian: @jwhui, I'm sorry, I thought it was the usual Travis timeouts... I'll check the Travis failures and your suggestions. Thanks!!

dismirlian: I see why it failed the tests (the ...OrExit macros are hard to get used to...). Implemented a minor variation your suggestion, so as not to mix bitwise and logical ORs and allow for lazy evaluation. Thanks for your help!

PR #4446, opened 2020-01-03
[posix-host] add posix parameter to get dataset from NCP and save to file

brianljt:
[dataset] Add posix app bootup parameter to retrieve dataset from NCP and save to file

There are cases when NCP platform was used at first. Then need to change to use RCP platform later. The PAN active and pending dataset needs to be retrieved from NCP firmware to avoid repairing again.

This change,
1. Add one bootup parameter to posix app (--ncp-dataset).
2. Utilize spinel property SPINEL_PROP_THREAD_ACTIVE_DATASET and SPINEL_PROP_THREAD_PENDING_DATASET to retrieve dataset from NCP.


src/posix/platform/radio_spinel.cpp
```diff
@@ -301,6 +311,20 @@ otError RadioSpinel::CheckRadioCapabilities(void)
     return error;
 }
 
+otError RadioSpinel::GetSaveNcpDataset(void)
```
bukepo: -> `RestoreDatasetFromNCP()?`?

src/posix/main.c
```diff
@@ -204,6 +212,13 @@ static otInstance *InitInstance(int aArgCount, char *aArgVector[])
         printf("%s\n", otPlatRadioGetVersionString(instance));
     }
 
+    if (config.mGetNcpDataset)
+    {
+        uint8_t error = OT_EXIT_SUCCESS;
+        error = ((platformGetSaveNcpDataset() == OT_ERROR_NONE) ? OT_EXIT_SUCCESS : OT_EXIT_FAILURE);
```
bukepo: This should be handled inside platform/system.c

src/posix/platform/radio_spinel.cpp
```diff
@@ -488,7 +664,12 @@ void RadioSpinel::HandleWaitingResponse(uint32_t          aCommand,
         }
         else
         {
-            if (aCommand == mExpectedCommand)
+            if (aKey == SPINEL_PROP_THREAD_ACTIVE_DATASET ||
```
bukepo: How about making a special format to indicate the caller wants to decode the payload itself? I find `"."` is proper for this purpose. see:https://github.com/openthread/openthread/blob/master/src/ncp/spinel.h#L144

src/posix/platform/system.c
```diff
@@ -64,6 +64,14 @@ otInstance *otSysInit(otPlatformConfig *aPlatformConfig)
     platformUdpInit(aPlatformConfig->mInterfaceName);
 #endif
 
+    if (aPlatformConfig->mGetNcpDataset)
+    {
+        uint8_t error = OT_EXIT_SUCCESS;
```
bukepo: nit: remove this?

src/posix/platform/radio_spinel.hpp
```diff
@@ -519,6 +519,14 @@ class RadioSpinel : public SpinelInterface::Callbacks
      */
     void HandleReceivedFrame(void);
 
+    /**
+     * This function get dataset from NCP radio and save it.
```
abtink: nit: `get` -> `gets`

src/posix/platform/radio_spinel.cpp
```diff
@@ -452,6 +478,157 @@ void RadioSpinel::HandleResponse(const uint8_t *aBuffer, uint16_t aLength)
...
+
+        default:
+            break;
+        }
+
+        SuccessOrExit(error = decoder.CloseStruct());
+    }
+
+    dataset.mActiveTimestamp                      = 0;
+    dataset.mComponents.mIsActiveTimestampPresent = true;
```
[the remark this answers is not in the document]
brianljt: "We may want to consider initially saving with Active Timestamp value set to 0. This will allow us to still join the network yet retrieve the full Active Dataset from a neighboring device if one exists."

I will add a comment in code to explain this.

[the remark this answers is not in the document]
abtink: Do we always want to retrieve it from neighbor again? If we already have it and if it has a higher time stamp than neighbor's, I think we should use ours.

Another point is about pending dataset, I think we need pending time stamp to be set for it?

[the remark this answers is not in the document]
jwhui: Spinel does not currently allow retrieving the full Operational Dataset (e.g. attributes that are not supported by spinel itself). If we updated spinel (and this solution) to retrieve the Operational Dataset as a binary blob, then I agree we should just keep the Operational Dataset as-is, including the timestamp.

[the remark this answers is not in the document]
abtink: Got it. Thanks. Makes sense as a solution for now.

.travis/script.sh
```diff
@@ -601,6 +601,26 @@ build_samr21() {
     .travis/check-posix-app-pty || die
 }
 
+[ $BUILD_TARGET != posix-app-migrate ] || {
+    # check daemon mode
```
bukepo: I don't think these ninja checks are necessary.

src/posix/platform/openthread-system.h
```diff
@@ -96,6 +96,7 @@ typedef struct otPlatformConfig
     const char *mRadioFile;     ///< Radio file path.
     const char *mRadioConfig;   ///< Radio configurations.
     bool        mResetRadio;    ///< Whether to reset RCP when initializing.
+    bool        mGetNcpDataset; ///< Whether to retrieve dataset from NCP and save to file.
```
bukepo: Is `mRestoreDatasetFromNcp` better?

src/posix/platform/radio_spinel.cpp
```diff
@@ -301,6 +311,22 @@ otError RadioSpinel::CheckRadioCapabilities(void)
     return error;
 }
 
+otError RadioSpinel::RestoreDatasetFromNcp(void)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(!mIsRcp);
+
```
bukepo: Is it possible to call otPlatSettingsInit() and otPlatSettingsDeinit() to directly save dataset in this method? This method can be called by Init().

.travis/check-ncp-rcp-migrate
```diff
@@ -0,0 +1,144 @@
...
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
+set -e
+set -x
+
+die() {
```
bukepo: remove this?

.travis/check-ncp-rcp-migrate
```diff
@@ -0,0 +1,144 @@
...
+send "masterkey\r\n"
+expect ${MASTER_KEY}
+expect "Done"
+send "exit\r\n"
+expect eof
+EOF
+}
+
+main() {
+    case $1 in
```
bukepo: quote `"$1"`.

src/posix/platform/radio_spinel.hpp
```diff
@@ -538,8 +546,10 @@ class RadioSpinel : public SpinelInterface::Callbacks
         kStateTransmitDone, ///< Radio indicated frame transmission is done.
     };
 
+    typedef otError (RadioSpinel::*ResponseHandler)(const uint8_t *aBuffer, uint16_t aLength);
+
     otError CheckSpinelVersion(void);
-    otError CheckCapabilities(void);
+    otError CheckCapabilities(bool *aIsRcp);
```
bukepo: use reference instead?

src/posix/platform/radio_spinel.hpp
```diff
@@ -519,6 +519,14 @@ class RadioSpinel : public SpinelInterface::Callbacks
      */
     void HandleReceivedFrame(void);
 
+    /**
+     * This function gets dataset from NCP radio and saves it.
+     *
+     * @retval  OT_ERROR_NONE               Succeeded.
+     * @retval  OT_ERROR_FAILED             Failed.
+     */
+    otError RestoreDatasetFromNcp(void);
```
bukepo: Make this private?

src/core/common/settings.hpp
```diff
@@ -152,6 +151,7 @@ class SettingsBase : public InstanceLocator
         kKeySlaacIidSecretKey = 0x0007, ///< Secret key used by SLAAC module for generating semantically opaque IID
     };
 
+protected:
```
bukepo: is this change necessary?

brianljt: This is needed as radio_spinel is not inherited from settings

bukepo:
@jwhui the travis tests passed.

src/posix/platform/CMakeLists.txt
```diff
@@ -68,4 +68,5 @@ target_include_directories(openthread-posix PRIVATE
     ${PROJECT_SOURCE_DIR}/src
     ${PROJECT_SOURCE_DIR}/src/core
     ${PROJECT_SOURCE_DIR}/src/posix/platform
+    ${PROJECT_SOURCE_DIR}/third_party/mbedtls/repo/include
```
jwhui: Is this necessary? Is this not getting included as part of `OT_PRIVATE_INCLUDES`?

https://github.com/openthread/openthread/blob/73e30d1d7b1454e0d82607381b6b9c1f79ff4cec/CMakeLists.txt#L91

brianljt: I think it's because this is not included by the platform library.(see the test failure: https://travis-ci.org/openthread/openthread/jobs/634169345).

src/posix/platform/radio_spinel.cpp
```diff
@@ -452,6 +488,160 @@ void RadioSpinel::HandleResponse(const uint8_t *aBuffer, uint16_t aLength)
...
+        }
+
+        case SPINEL_PROP_NET_NETWORK_NAME:
+        {
+            const char *name;
+            size_t      len;
+
+            SuccessOrExit(error = decoder.ReadUtf8(name));
+            len = strnlen(name, OT_NETWORK_NAME_MAX_SIZE + 1);
+            memcpy(opDataset.mNetworkName.m8, name, len + 1);
```
jwhui: What happens when `len` is `OT_NETWORK_NAME_MAX_SIZE + 1`?

src/posix/platform/radio_spinel.cpp
```diff
@@ -452,6 +488,160 @@ void RadioSpinel::HandleResponse(const uint8_t *aBuffer, uint16_t aLength)
...
+        }
+
+        case SPINEL_PROP_IPV6_ML_PREFIX:
+        {
+            const otIp6Address *addr;
+            uint8_t             prefixLen;
+
+            SuccessOrExit(error = decoder.ReadIp6Address(addr));
+            SuccessOrExit(error = decoder.ReadUint8(prefixLen));
+            VerifyOrExit(prefixLen == 64, error = OT_ERROR_INVALID_ARGS);
```
jwhui: nit: avoid magic number

src/posix/platform/radio_spinel.cpp
```diff
@@ -452,6 +488,160 @@ void RadioSpinel::HandleResponse(const uint8_t *aBuffer, uint16_t aLength)
...
+    /*
+     * Initially set Active Timestamp to 0. This is to allow the node to join the network
+     * yet retrieve the full Active Dataset from a neighboring device if one exists.
+     */
+    opDataset.mActiveTimestamp                      = 0;
+    opDataset.mComponents.mIsActiveTimestampPresent = true;
+
+    SuccessOrExit(error = dataset.Set(opDataset));
+    SuccessOrExit(error = otPlatSettingsSet(mInstance,
+                                            (mWaitingKey == SPINEL_PROP_THREAD_ACTIVE_DATASET)
```
jwhui: Store dataset type in local var and use both here and in `dataset` constructor call above?

src/posix/platform/radio_spinel.cpp
brianljt: You are right. It will probably copy a name string without tailing \0. I think the network name should not exceed OT_NETWORK_NAME_MAX_SIZE, otherwise, an assert should happen here.

src/posix/platform/CMakeLists.txt
bukepo: @jwhui the platform library is added before that path is added into `OT_PRIVATE_INCLUDES`.

src/posix/platform/radio_spinel.cpp
```diff
@@ -452,6 +488,159 @@ void RadioSpinel::HandleResponse(const uint8_t *aBuffer, uint16_t aLength)
...
+            break;
+        }
+
+        case SPINEL_PROP_NET_NETWORK_NAME:
+        {
+            const char *name;
+            size_t      len;
+
+            SuccessOrExit(error = decoder.ReadUtf8(name));
+            len = strlen(name);
```
jwhui: I would keep the `strnlen` to properly bound the scan to `OT_NETWORK_NAME_MAX_SIZE + 1` chars.

PR #4447, opened 2020-01-03
[efr32] update to Silicon Labs SDK v2.7

marvensilabs:
Updated makefiles for v2.7 SDK
Replaced RAIL timers with sleeptimers.


marvensilabs:
Remove function call to disable execution from RAM in latest commit.

PR #4452, opened 2020-01-06
[coap] add otCoapOptionIterator and related API functions

sjlongland:
As discussed in https://github.com/openthread/openthread/pull/4396#discussion_r357525060, this adds a `otCoapOptionIterator` structure which the user can use to iterate over the CoAP options of a `ot::Coap::Message` object without influencing any of the properties of that object.

Multiple `otCoapOptionIterator` structures may operate on the same `ot::Coap::Message` object without interference.

The code has been tested on nRF52840.

### DIFF
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -583,8 +583,8 @@ const uint8_t *otCoapMessageGetToken(const otMessage *aMessage);
  * @param[inout]  aIterator A pointer to the CoAP message option iterator.
  * @param[in]     aMessage  A pointer to the CoAP message.
  *
- * @retval  OT_ERROR_NONE   Successfully initialised
- * @retval  OT_ERROR_PARSE  Message state is inconsistent
+ * @retval  OT_ERROR_NONE   Successfully initialised.
+ * @retval  OT_ERROR_PARSE  Message state is inconsistent.
  *
  */
 otError otCoapOptionIteratorInit(otCoapOptionIterator *aIterator, const otMessage *aMessage);
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -578,36 +578,45 @@ uint8_t otCoapMessageGetTokenLength(const otMessage *aMessage);
 const uint8_t *otCoapMessageGetToken(const otMessage *aMessage);
 
 /**
+ * This function initialises an iterator for the options in the given message.
+ *
+ * @param[inout]  aIterator A pointer to the CoAP message option iterator.
+ * @param[in]     aMessage  A pointer to the CoAP message.
+ *
+ */
+void otCoapOptionIteratorInit(otCoapOptionIterator *aIterator, const otMessage *aMessage);
+
+/**
  * This function returns a pointer to the first option.
  *
- * @param[in]  aMessage  A pointer to the CoAP message.
+ * @param[inout]  aIterator A pointer to the CoAP message option iterator.
  *
  * @returns A pointer to the first option. If no option is present NULL pointer is returned.
  *
  */
-const otCoapOption *otCoapMessageGetFirstOption(otMessage *aMessage);
+const otCoapOption *otCoapOptionIteratorGetFirstOption(otCoapOptionIterator *aIterator);
 
 /**
  * This function returns a pointer to the next option.
  *
- * @param[in]  aMessage  A pointer to the CoAP message.
+ * @param[inout]  aIterator A pointer to the CoAP message option iterator.
  *
  * @returns A pointer to the next option. If no more options are present NULL pointer is returned.
  *
  */
-const otCoapOption *otCoapMessageGetNextOption(otMessage *aMessage);
+const otCoapOption *otCoapOptionIteratorGetNextOption(otCoapOptionIterator *aIterator);
 
 /**
  * This function fills current option value into @p aValue.
  *
- * @param[in]  aMessage  A pointer to the CoAP message.
- * @param[out] aValue    A pointer to a buffer to receive the option value.
+ * @param[inout]  aIterator A pointer to the CoAP message option iterator.
+ * @param[out]    aValue    A pointer to a buffer to receive the option value.
  *
  * @retval  OT_ERROR_NONE       Successfully filled value.
  * @retval  OT_ERROR_NOT_FOUND  No current option.
  *
  */
-otError otCoapMessageGetOptionValue(otMessage *aMessage, void *aValue);
+otError otCoapOptionIteratorGetOptionValue(otCoapOptionIterator *aIterator, void *aValue);
 
 /**
  * This function creates a new CoAP message.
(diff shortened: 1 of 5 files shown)


sjlongland:
Okay, this has me confused:

```
coap/coap_message.cpp:245:45: warning: The left operand of '>' is a garbage value
    VerifyOrExit(iterator.mNextOptionOffset > 0, error = OT_ERROR_PARSE);
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
./common/code_utils.hpp:112:15: note: expanded from macro 'VerifyOrExit'
        if (!(aCondition))            \
              ^~~~~~~~~~
1 warning generated.
```

Just a few lines above that is `iterator.Init(this);` which calls `otCoapOptionIterator::GetFirstOption` which sets `mNextOptionOffset = message.GetHelpData().mHeaderOffset + message.GetOptionStart();`.  How is this "garbage"?

jwhui:
> Just a few lines above that is `iterator.Init(this);` which calls `otCoapOptionIterator::GetFirstOption` which sets `mNextOptionOffset = message.GetHelpData().mHeaderOffset + message.GetOptionStart();`. How is this "garbage"?

The compiler believes that there are code paths in the call to `iterator.Init(this)` that do not set `iterator.mNextOptionOffset` - see the `VerifyOrExit()` statement just before setting `mNextOptionOffset` [here](https://github.com/openthread/openthread/pull/4452/files#diff-65fd928e47b2c24bd2dc1dbbe9662d8fR405).

sjlongland:
Ahh, gotcha… sometimes those `VerifyOrExit` macros can seem a little opaque, but no problems.  I think the best plan of attack would be to make `Init(const Message*)` return `OT_ERROR_PARSE` if this condition is true, and for the calling code to check for that.

Probably more robust than relying on `GetFirstOption` returning `NULL` (which can happen legitimately if there are no options).

sjlongland:
I'm open to suggestions on function naming… I took the rationale of being predictable in the name changes rather than being terse, so the function names in the C API particularly, are a bit on the long side.

src/core/coap/coap_message.hpp
```diff
@@ -607,6 +587,44 @@ class Message : public ot::Message
...
+
+    /**
+     * This method returns a pointer to the first option.
+     *
+     */
+    const otCoapOption *GetFirstOption(void);
+
+    /**
+     * This method returns a pointer to the next option.
+     *
```
jwhui: nit: add "@returns A pointer to the next option."

src/core/coap/coap_message.hpp
```diff
@@ -607,6 +587,44 @@ class Message : public ot::Message
...
+     * @retval  OT_ERROR_NONE   Successfully initialised
+     * @retval  OT_ERROR_PARSE  Message state is inconsistent
+     *
+     */
+    otError Init(const Message *aMessage);
+
+    /**
+     * This method returns a pointer to the first option.
+     *
```
jwhui: nit: add "@returns A pointer to the first option."

src/core/coap/coap_message.cpp
```diff
@@ -502,5 +389,140 @@ const char *Message::CodeToString(void) const
...
+{
+    otError err = OT_ERROR_NONE;
+
+    /*
+     * Check that:
+     *   Length - Offset: the length of the payload
+     * is greater than
+     *   Start position of optionsa
+     *
+     * → Assert options start before the message ends
```
jwhui: Do we really assert?

src/core/coap/coap_message.cpp
```diff
@@ -502,5 +389,140 @@ const char *Message::CodeToString(void) const
...
 
+otError OptionIterator::Init(const Message *aMessage)
+{
+    otError err = OT_ERROR_NONE;
+
+    /*
+     * Check that:
+     *   Length - Offset: the length of the payload
+     * is greater than
+     *   Start position of optionsa
```
jwhui: nit: docs seems to have formatting/typo issues

include/openthread/coap.h
```diff
@@ -566,37 +577,49 @@ uint8_t otCoapMessageGetTokenLength(const otMessage *aMessage);
  */
 const uint8_t *otCoapMessageGetToken(const otMessage *aMessage);
 
+/**
+ * This function initialises an iterator for the options in the given message.
+ *
+ * @param[inout]  aIterator A pointer to the CoAP message option iterator.
+ * @param[in]     aMessage  A pointer to the CoAP message.
+ *
+ * @retval  OT_ERROR_NONE   Successfully initialised
```
jwhui: nit: prefer to end doc lines with period '.'

src/core/coap/coap_message.hpp
```diff
@@ -607,6 +587,44 @@ class Message : public ot::Message
...
+     * This method returns a pointer to the next option.
+     *
+     */
+    const otCoapOption *GetNextOption(void);
+
+    /**
+     * This function fills current option value into @p aValue.
+     *
+     * @retval  OT_ERROR_NONE       Successfully filled value.
+     * @retval  OT_ERROR_NOT_FOUND  No more options, aIterator->mNextOptionOffset is set to offset of payload.
```
jwhui:
```suggestion
     * @retval  OT_ERROR_NOT_FOUND  No more options, mNextOptionOffset is set to offset of payload.
```

src/core/coap/coap_message.cpp
sjlongland: Maybe "assert" is the wrong verb (it's a "soft assert": if condition isn't met we bail the `Init` function with `OT_ERROR_PARSE` as the reason).

I put the comment there because it took me a few minutes to decipher what the `VerifyOrExit` check was doing and what it would do in the "no options" case (i.e. did I need to make `>=` a `>`, check for `==` separately, etc).

jwhui:
@sjlongland, thanks for making the updates. It seems the docker check failed. Can you push again to kick the docker test?

sjlongland:
It actually looks like the usual PGP keyserver time-out issues that we face with our own Docker stuff internally:

```
If it reports error message of conflict to gcc-arm-none-eabi, which is likely if upgrading from 4.x to 5+, please uninstall it first with:
       "sudo apt-get remove gcc-arm-none-eabi"

Questions should be asked at https://answers.launchpad.net/gcc-arm-embedded

Bug can be filed at https://bugs.launchpad.net/gcc-arm-embedded/+filebug. It is highly encouraged to ask question first before filing a bug.
 More info: https://launchpad.net/~team-gcc-arm-embedded/+archive/ubuntu/ppa

Error: retrieving gpg key timed out.

 *** ERROR: 

Removing intermediate container bbf2b266875d

The command '/bin/sh -c ./script/bootstrap' returned a non-zero code: 1
```

Not sure what key server is being used, but lately the usual keyserver pool has been particularly unreliable lately.  Using the Ubuntu one helps, but isn't a complete solution.

I've just "amended" the last commit and force-pushed it, no code changes.  So we'll see how it goes this time around. :-)

PR #4453, opened 2020-01-07
[lowpan] simplify MeshHeader/FragHeader and their use in MeshForwarder

abtink:
This PR contains a group of related commits changing how `Lowpan::MeshHeader` is parsed and generated and its use in `MeshForwarder`. 

**[lowpan] simplify MeshHeader and its use in MeshForwarder**

This commit updates the `Lowpan::MeshHeader` class to provide methods to parse/write the Mesh Header from/to a given frame or message. The `MeshHeader` no longer provides a direct representation of the header (instead, it contains all fields required to generate the header). This is mainly due to fact the header format itself may vary depending on the "hops left" value. This change also removes the need for the `MeshHeader` to be a packed `OT_TOOL_PACKED` structure.

**[mesh-forwarder] skip mesh header immediately after parsing it**

This commit updates `HandleMesh()` method to skip the mesh header in the frame immediately after parsing it. This in turn removes the need for other methods (`CheckReachability` or `UpdateRoutes()` calling through `GetIp6Header()` to `DecompressIp6Header()`) to redo the parsing to skip the mesh header. With this change we can also remove the now unused `SkipMeshHeader()`.

**[mesh-forwarder] simplify GetForwardFramePriority() and remove GetMeshHeader()**
    
Before passing the frame to `GetForwardFramePriority()` the mesh header is skipped (to avoid re-reading the mesh header). Also the src/dest parameters are renamed to indicate they are mesh source and mesh destination (also the order of parameters changed to start with source and then dest). With this change we can remove `GetMeshHeader()` (since no longer used/needed).

**[unit-test] add unit test for Lowpan::MeshHeader class**

----------

This PR also has few smaller changes in `MeshForwarder`:

- **[mesh-forwarder] use pointer to iterate through mFragmentEntries**
- **[mesh-forwarder] use SuccessOrExit, update comments**

### DIFF
--- a/src/core/thread/lowpan.cpp
+++ b/src/core/thread/lowpan.cpp
@@ -1319,19 +1319,40 @@ uint16_t MeshHeader::WriteTo(Message &aMessage, uint16_t aOffset) const
     return headerLength;
 }
 
-otError FragmentHeader::Init(const uint8_t *aFrame, uint16_t aFrameLength)
+//---------------------------------------------------------------------------------------------------------------------
+// FragmentHeader
+
+void FragmentHeader::Init(uint16_t aSize, uint16_t aTag, uint16_t aOffset)
+{
+    mSize   = (aSize & kSizeMask);
+    mTag    = aTag;
+    mOffset = (aOffset & kOffsetMask);
+}
+
+bool FragmentHeader::IsFragmentHeader(const uint8_t *aFrame, uint16_t aFrameLength)
+{
+    return (aFrameLength >= kFirstFragmentHeaderSize) && ((*aFrame & kDispatchMask) == kDispatch);
+}
+
+otError FragmentHeader::ParseFrom(const uint8_t *aFrame, uint16_t aFrameLength, uint16_t &aHeaderLength)
 {
     otError error = OT_ERROR_PARSE;
 
-    VerifyOrExit(aFrameLength >= sizeof(mDispatchSize) + sizeof(mTag));
-    memcpy(reinterpret_cast<uint8_t *>(&mDispatchSize), aFrame, sizeof(mDispatchSize) + sizeof(mTag));
-    aFrame += sizeof(mDispatchSize) + sizeof(mTag);
-    aFrameLength -= sizeof(mDispatchSize) + sizeof(mTag);
+    VerifyOrExit(IsFragmentHeader(aFrame, aFrameLength));
 
-    if (IsOffsetPresent())
+    mSize = ReadUint16(aFrame + kSizeIndex) & kSizeMask;
+    mTag  = ReadUint16(aFrame + kTagIndex);
+
+    if ((*aFrame & kOffsetFlag) == kOffsetFlag)
     {
-        VerifyOrExit(aFrameLength >= sizeof(mOffset));
-        mOffset = *aFrame++;
+        VerifyOrExit(aFrameLength >= kSubsequentFragmentHeaderSize);
+        mOffset       = aFrame[kOffsetIndex] * 8;
+        aHeaderLength = kSubsequentFragmentHeaderSize;
+    }
+    else
+    {
+        mOffset       = 0;
+        aHeaderLength = kFirstFragmentHeaderSize;
     }
 
     error = OT_ERROR_NONE;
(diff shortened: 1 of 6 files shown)


src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -548,7 +547,7 @@ void MeshForwarder::HandleMesh(uint8_t *               aFrame,
         meshHeader.SetHopsLeft(meshHeader.GetHopsLeft() - 1);
         meshHeader.AppendTo(aFrame);
 
-        GetForwardFramePriority(aFrame, aFrameLength, meshDest, meshSource, priority);
+        GetForwardFramePriority(aFrame, aFrameLength, meshHeader, meshDest, meshSource, priority);
```
jwhui: The addition of `meshHeader` argument here seems incorrect.

src/core/thread/lowpan.cpp
```diff
@@ -1206,58 +1207,115 @@ int Lowpan::Decompress(Message &           aMessage,
...
-    mDispatchHopsLeft = *aFrame++;
-    aFrameLength--;
+    VerifyOrExit((dispatch & (kDispatchMask | kSourceShort | kDestShort)) == (kDispatch | kSourceShort | kDestShort));
 
-    if (IsDeepHopsLeftField())
+    mHopsLeft = (dispatch & kHopsLeftMask);
+
+    if (mHopsLeft == kDeepHopsLeft)
     {
-        VerifyOrExit(aFrameLength >= 1, error = OT_ERROR_PARSE);
```
jwhui: Where do we check that the frame length is `>=` than expected header size in the case that Deep Hops Left field is used?

src/core/thread/lowpan.cpp
```diff
@@ -1206,58 +1207,115 @@ int Lowpan::Decompress(Message &           aMessage,
...
-    uint16_t offset = 0;
-    uint16_t bytesRead;
+    uint16_t headerLength;
+
+    return ParseFrom(aMessage, aOffset, headerLength);
+}
+
+otError MeshHeader::ParseFrom(const Message &aMessage, uint16_t aOffset, uint16_t &aHeaderLength)
+{
+    uint8_t  frame[kDeepHopsLeft];
```
jwhui: Why size the buffer based on `kDeepHopsLeft`?

src/core/thread/lowpan.hpp
```diff
@@ -399,174 +398,151 @@ class MeshHeader
...
      */
-    otError Init(const Message &aMessage);
+    otError ParseFrom(const Message &aMessage, uint16_t aOffset);
 
     /**
-     * This method indicates whether or not the header is a Mesh Header.
-     *
-     * @retval TRUE   If the header matches the Mesh Header dispatch value.
-     * @retval FALSE  If the header does not match the Mesh Header dispatch value.
+     * This method parses the Mesh Header from a message and a given offset.
```
jwhui: `message and a given offset` -> `message at a given offset`

src/core/thread/lowpan.hpp
```diff
@@ -399,174 +398,151 @@ class MeshHeader
...
      */
-    bool IsValid(void) const { return (mDispatchHopsLeft & kSourceShort) && (mDispatchHopsLeft & kDestinationShort); }
+    otError ParseFrom(const Message &aMessage, uint16_t aOffset, uint16_t &aHeaderLength);
 
     /**
-     * This method indicates whether or not the header contains Deep Hops Left field.
-     *
-     * @retval TRUE   If the header does contain Deep Hops Left field.
-     * @retval FALSE  If the header does not contain Deep Hops Left field.
+     * This method returns the the Mesh Header length when written to a frame/message.
```
jwhui: nit: `frame/message` -> `frame`, to be consistent with RFC 4944.

src/core/thread/lowpan.hpp
```diff
@@ -399,174 +398,151 @@ class MeshHeader
...
+    uint16_t GetDestination(void) const { return mDestination; }
 
     /**
-     * This method sets the Mesh Destination address.
+     * This method writes the Mesh Header into a given frame.
      *
-     * @param[in]  aDestination  The Mesh Destination address.
+     * @note This method method expect the frame buffer to have enough space for entire Mesh Header.
+     *
+     * @param[in]  aFrame  The pointer to the frame buffer to write to.
```
jwhui: Document as `inout`?

src/core/thread/lowpan.hpp
```diff
@@ -399,174 +398,151 @@ class MeshHeader
...
 
     /**
-     * This method appends Mesh Header to the @p aFrame frame.
+     * This method writes the Mesh Header to a message at a given offset.
      *
-     * @param[in]  aFrame  The pointer to the frame.
+     * @note This method expects the @p aMessage length to be already set such there is enough space for entire Mesh
+     * Header to be written.
+     *
+     * @param[in] aMessage  A message to write the Mesh Header into.
```
jwhui: Document as `inout`?

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -846,14 +851,14 @@ otError MeshForwarder::LogMeshFragmentHeader(MessageAction       aAction,
     bool                   shouldLogRss;
     Lowpan::MeshHeader     meshHeader;
     Lowpan::FragmentHeader fragmentHeader;
+    uint16_t               meshHeaderLength;
 
-    SuccessOrExit(meshHeader.Init(aMessage));
-    VerifyOrExit(meshHeader.IsMeshHeader());
+    SuccessOrExit(meshHeader.ParseFrom(aMessage, /* aOffset */ 0, meshHeaderLength));
```
jwhui: Is `MeshHeader::ParseFrom` ever called with `aOffset != 0`?

tests/unit/test_lowpan.cpp
```diff
@@ -1832,11 +1832,83 @@ void TestLowpanIphc(void)
     testFreeInstance(sInstance);
 }
 
+void TestLowpanMeshHeader(void)
```
jwhui: Can we add some tests to ensure `ParseFrom` does not read beyond `headerLength`?

src/core/thread/lowpan.cpp
abtink: Good point. Thanks. Will fix it.

src/core/thread/lowpan.cpp
abtink: It's typo. Will change to `kDeepHopsHeaderLength`

src/core/thread/mesh_forwarder_ftd.cpp
abtink: Not currently (and I don't expect we would need it).
We can keep it simple (as it was) and assume we read MeshHeader at zero offset.

tests/unit/test_lowpan.cpp
abtink: Good idea. Will add it.

abtink:
In new push added two new commits:

**[lowpan] simplify FragmentHeader and its use in MeshForwarder** (https://github.com/openthread/openthread/pull/4453/commits/2e946d065d65ecd02ce101f4fefb7f3f70b10135)
    
This commit updates the `Lowpan::FragmentHeader` class to provide methods to parse/write the Fragment Header from/to a given frame or message.

**[lowpan] update BufferWriter to use end pointer instead of length** (https://github.com/openthread/openthread/pull/4453/commits/57b85d75f2ba090683eca81b0ba86617292cd952)
    
This commit also updates some of `Lowpan` (decompression) methods to use an end pointer instead of tracking remaining length.

----

Also added unit test for `FragmentHeader` (updated the related commit https://github.com/openthread/openthread/pull/4453/commits/6c713f7fa7be5fc6067e747202bb23967b41b5a5)

src/core/thread/lowpan.hpp
```diff
@@ -550,158 +550,125 @@ class MeshHeader
...
-        {
-            mDispatchSize = HostSwap16(HostSwap16(mDispatchSize) & ~kOffset);
-        }
-        else
-        {
-            mDispatchSize = HostSwap16(HostSwap16(mDispatchSize) | kOffset);
-            mOffset       = (aOffset >> 3) & kOffsetMask;
-        }
-    }
+    uint16_t WriteTo(uint8_t *aFrame) const;
```
jwhui: Missing documentation.

abtink: Thanks. Added in new pushed.

PR #4463, opened 2020-01-10
[ncp] remove dead code in SPINEL_PROP_STREAM_NET_INSECURE set handler

jwhui:


src/ncp/ncp_base_mtd.cpp
```diff
@@ -2921,11 +2921,6 @@ template <> otError NcpBase::HandlePropertySet<SPINEL_PROP_STREAM_NET_INSECURE>(
     message = NULL;
```
abtink: :+1:. With this change,  we can also remove the `message = NULL` and update the comment above it?
```suggestion
    
```

PR #4458, opened 2020-01-09
[misc] fix undefined (config) macro use warnings 

abtink:
This PR is a follow up on PR #4300. It contains a group of smaller commits fixing the warnings for use of undefined (config) macros in different modules.

- [posix-app] fix undefined config macro warnings 
- [ncp/cli/posix-app] disable coverage test when macro not defined 
  This change helps suppress the undefined macro warnings.
- [utils-settings] add a header file for settings config definitions 
    This commit adds a header file `settings.h` to include all the
    default configuration definitions related to `settings` platform
    utility. It also adds `OPENTHREAD_SETTINGS_RAM` to this file.
    This helps address the undefined macro warnings.

- [example-posix] add missing config definitions 
    This commit updates `platform-config.h` to add default value for
    config definitions. This helps suppress the undefined macro
    warnings.
- [toranj] fix undefined config macro warnings 
- [wrap-strings] fix undefined config macro warnings 
- [spi-hdlc-adapter] fix undefined config macro warnings

PR #4469, opened 2020-01-11
[build] define OPENTHREAD_FTD/MTD/RADIO in posix-app and unit test makefiles

abtink:
This helps address undefined config macro warnings.


bukepo:
We intended to only support FTD on posix-host. Will this change also enable MTD and RCP?

abtink:
@bukepo good point. Will change it to assume FTD only in posix-app makefiles.
The goal for this is to address the issue with "undefined macro used" warnings (follow up on https://github.com/openthread/openthread/pull/4458)

PR #3575, opened 2019-02-10
[mbedtls] guard allocator with accurate macro

bukepo:
This PR guard using mbedTLS's buffer allocator API with a more accurate
macro, so that even in single instance mode, we can use other memory
allocator.

PR #4473, opened 2020-01-13
[mbedtls] split builtin mbedtls management option from mbedtls build option

amosolgo-lutron:
Add a new `OPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS_MANAGEMENT` flag to configure builtin mbedtls buffer/debug management separately from the existing flag which configures the inclusion of the builtin mbedtls library in the build.

This adjusts the conditional compilation sections added in [#3575](https://github.com/openthread/openthread/pull/3575) to use the new config flag so that it may be controlled separately from the existing flag. The default behavior should remain unchanged: if `--enable-builtin-mbedtls=yes` is used then builtin mbedtls management is enabled by default, and if `--enable-builtin-mbedtls=no` is used then builtin mbedtls management is disabled by default.


amosolgo-lutron:
Hi @bukepo, @rlubos, @jwhui, as participants in #3575, you may be interested in this PR.  I saw that early commits on #3575 used an implementation similar to this.

src/core/config/openthread-core-default-config.h
```diff
@@ -240,6 +240,26 @@
...
+ *
+ * OPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS_MANAGEMENT determines whether to manage mbedTLS memory
+ * allocation and debug config internally.  If not configured, the default is to enable builtin
+ * management if builtin mbedtls is enabled and disable it otherwise.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS_MANAGEMENT
+/* Make sure the build system config has provided OPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS before it
+ * is used as a default value. */
+#ifndef OPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS
```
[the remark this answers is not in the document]
amosolgo-lutron: You're right! I had originally removed the default `OPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS` so the check made sense, but I forgot to remove the check after I returned the default definition above.  I'll take out this check.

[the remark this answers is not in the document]
amosolgo-lutron: Removed in 02551f1

examples/common-switches.mk
```diff
@@ -209,6 +209,10 @@ ifeq ($(DISABLE_BUILTIN_MBEDTLS),1)
 configure_OPTIONS              += --disable-builtin-mbedtls
 endif
 
+ifdef BUILTIN_MBEDTLS_MANAGEMENT
+COMMONCFLAGS                   += -DOPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS_MANAGEMENT=$(BUILTIN_MBEDTLS_MANAGEMENT)
```
bukepo:
```suggestion
COMMONCFLAGS                   += -DOPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS_MANAGEMENT=1
```

examples/common-switches.mk
```diff
@@ -209,6 +209,10 @@ ifeq ($(DISABLE_BUILTIN_MBEDTLS),1)
 configure_OPTIONS              += --disable-builtin-mbedtls
 endif
 
+ifdef BUILTIN_MBEDTLS_MANAGEMENT
```
bukepo: is it possible to use the same style as other flags?

amosolgo-lutron: The goal with using `ifdef` is to allow the definition for `DOPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS_MANAGEMENT` to be overridden either with zero or non-zero.  Most other flags are less flexible and only allow setting the value to 1.

I now see another flag with similar behavior, and I could use similar `ifneq` logic for more style consistency:

```
ifneq ($(LOG_OUTPUT),)
COMMONCFLAGS                   += -DOPENTHREAD_CONFIG_LOG_OUTPUT=OPENTHREAD_CONFIG_LOG_OUTPUT_$(LOG_OUTPUT)
endif
```

PR #4480, opened 2020-01-16
[posix-host] explicitly ignore return value on write

dismirlian:
Build failed because `write` was declared with attribute `warn_unused_result`.

### DIFF
--- a/src/posix/client.c
+++ b/src/posix/client.c
@@ -55,6 +55,8 @@
 #include <unistd.h>
 
 #include "code_utils.h"
+#include "common/code_utils.hpp"
+
 #include "platform-posix.h"
 
 static int sSessionFd = -1;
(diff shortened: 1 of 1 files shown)


dismirlian:
@googlebot I fixed it.

dismirlian:
@googlebot I consent.

src/posix/client.c
```diff
@@ -154,7 +156,7 @@ int main(int argc, char *argv[])
             }
             else
             {
-                write(STDOUT_FILENO, buffer, rval);
+                IgnoreReturnValue(write(STDOUT_FILENO, buffer, rval));
```
jwhui: @bukepo any reason to check the return value?

bukepo: Yes, I think so. It may require multiple writes if the return value is less than `rval`. We can enhance in this PR or later.

bukepo: Besides we also need enable this check in travis.

PR #4482, opened 2020-01-17
[efr32] Enhancements to radio driver (1/3)

dismirlian:
This is the first of 3 PR addressing the efr32mg12 radio driver.

PR# 1: 
- Features:
   * 18b70b2 Implemented get/set CcaEnergyDetectThreshold.

- Sniffer mode fix:
   * 981ee39 Implemented RX timestamping (at the SFD reception).
   * 21f442a Handled promiscuous mode for ACK packets (ACK packets were discarded for sniffer mode).

- Misc bug corrections:
   * 1042eda Corrected macAckWaitDuration (from 894us to 864us; it was a typo).
   * 18b70b2 Honor mMaxCsmaBackoffs.
   * f4dbf62 Signalled event pending when failed to start TX (event pending was not signalled for this case, even if TX operation was actually finished).

- Minor enhancements
   * 1027145 Removed duplicate calls of RAIL functions which were called once per band, and did not need to (the call was exactly the same for all bands).
   * f2e2755 Correct hardware capabilities.
   * f2e2755 Correct soft source match table size.
   * f9c65d0 Used assert instead of otEXPECT to detect issues when interacting with RAIL.

@andrasbiro-silabs please review.

Thanks

### DIFF
--- a/examples/platforms/efr32mg12/openthread-core-efr32-config.h
+++ b/examples/platforms/efr32mg12/openthread-core-efr32-config.h
@@ -93,7 +93,7 @@
  * Define to 1 if you want to enable software CSMA-CA backoff logic.
  *
  */
-#define OPENTHREAD_CONFIG_SOFTWARE_CSMA_BACKOFF_ENABLE 1
+#define OPENTHREAD_CONFIG_SOFTWARE_CSMA_BACKOFF_ENABLE 0
 
 /**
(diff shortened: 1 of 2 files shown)


dismirlian:
> Thanks for the simplification, especially for the TOC in the PR text!

@andrasbiro-silabs No problem! Thanks for reviewing.

PR #4483, opened 2020-01-17
[utils] remove "missing" impls of strlcat, strlcpy, strnlen

jwhui:


src/core/common/string.cpp
```diff
@@ -35,6 +35,18 @@
 
 namespace ot {
 
+uint16_t StringLength(const char *aString, uint16_t aMaxLength)
```
bukepo: Is it possible to provide as a weak implementation of `strnlen()`?

jwhui: How would you propose to keep the current unit tests for `StringLength()`? Would you define a weak strnlen that uses `StringLength()`?

bukepo: Got it!

Issue #4358, opened 2019-11-26
[radio] properly set mAckedWithFramePending on received frame


jwhui:
The received frame structure includes an `mAckedWithFramePending` field that must be set if and only if the ACK sent in response to the received frame has the Frame Pending bit set. #4131 relies on radio driver to properly implement this. However, not all platforms implement this.

Without proper implementation, testing against the Thread Test Harness may result in test case failures due to redundant empty data frames caused by PR #4131.

This issue is to track the current example driver implementation status for properly setting the `mAckedWithFramePending` field in a received frame.

Platform status:

- [ ] cc1352
- [x] cc2538
- [ ] cc2650
- [ ] cc2652
- [x] efr32mg12
- [x] efr32mg13
- [x] efr32mg21
- [ ] gp712
- [ ] kw41z
- [x] nrf52811
- [x] nrf52833
- [x] nrf52840
- [x] posix
- [ ] qpg6095
- [ ] samr21

jwhui:
Platform-specific code has moved into platform-specific repos.

PR #4484, opened 2020-01-17
[efr32] Enhancements to radio driver (2/3)

dismirlian:
This is the second of 3 PRs addressing the efr32mg12 radio driver.

PR# 2: Implements correctly setting `mAckedWithFramePending` (See #4358).

**The idea is the following:** when receiving an IEEE802.15.4 Data Request Command, RAIL generates an event and calls `RAILCb_Generic` as early as possible (even before the packet is fully received; i.e. the CRC could be still wrong), from interrupt context.

In `RAILCb_Generic`, the driver calls `ieee802154DataRequestCommand`, which in turn decides whether to set the Frame Pending flag for the outgoing ACK, by analysing the soft source match table. If the Frame Pending flag is set, we have to report `mAckedWithFramePending = true` when we pass the Data Request Command packet to the upper layers.

Since the Data Request Command packet is held in the callback and passed to the upper layers in `processNextRxPacket`, and the Parent can receive multiple Data Request Commands from different children back-to-back before calling `processNextRxPacket`, one cannot simply set a flag in the callback to signal `processNextRxPacket` that the Data Request Command was acked with Frame Pending.

The proposed solution is to insert a "unique identifier" of the Data Request Command packet into a FIFO when the ACK was sent with the Frame Pending flag is set. Then, for each Data Request Command packet that `processNextRxPacket` will pass to the upper layer, it will pop the "unique identifiers" from the FIFO searching for a match. If the identifier is found, `mAckedWithFramePending` will be set to `true`.

The "unique identifier" is actually the contents of the packet being received (PHR + MAC Header), because the `RAIL_GetRxIncomingPacketInfo` does not return a `RAIL_RxPacketHandle_t`.

When `processNextRxPacket` pops a unique identifier from the FIFO:
- It pops the Data Request Command packets in sync with the RAIL RX FIFO, so that the packets are correctly matched.
- It discards the packets which could be inserted to the FIFO without passing the CRC check.

The FIFO size is set to 16 entries, which should be the maximum that RAIL can hold in its internal metadata FIFO (see https://docs.silabs.com/rail/latest/efr32-main).

Note: I think this method (with some modifications) could be compatible with the Thread 1.2 Enhanced Frame Pending feature.

Without this PR (empty data packets):
image

With this PR (no empty data packets):
image

@andrasbiro-silabs please review. If you have a better idea on how to uniquely identify the packets, please let me know; you are the RAIL expert.

Thanks!

### DIFF
--- a/examples/platforms/efr32mg12/radio.c
+++ b/examples/platforms/efr32mg12/radio.c
@@ -805,9 +805,7 @@ static void processNextRxPacket(otInstance *aInstance)
     packetInfo.packetBytes--;
 
     // read packet
-    memcpy(sReceiveFrame.mPsdu, packetInfo.firstPortionData, packetInfo.firstPortionBytes);
-    memcpy(sReceiveFrame.mPsdu + packetInfo.firstPortionBytes, packetInfo.lastPortionData,
-           packetInfo.packetBytes - packetInfo.firstPortionBytes);
+    RAIL_CopyRxPacket(sReceiveFrame.mPsdu, &packetInfo);
 
     status = RAIL_ReleaseRxPacket(gRailHandle, packetHandle);
     if (status == RAIL_STATUS_NO_ERROR)
--- a/examples/platforms/efr32mg12/radio.c
+++ b/examples/platforms/efr32mg12/radio.c
@@ -59,14 +59,19 @@
 
 enum
 {
-    IEEE802154_MIN_LENGTH      = 5,
-    IEEE802154_MAX_LENGTH      = 127,
-    IEEE802154_ACK_LENGTH      = 5,
-    IEEE802154_FRAME_TYPE_MASK = 0x7,
-    IEEE802154_FRAME_TYPE_ACK  = 0x2,
-    IEEE802154_FRAME_PENDING   = 1 << 4,
-    IEEE802154_ACK_REQUEST     = 1 << 5,
-    IEEE802154_DSN_OFFSET      = 2,
+    IEEE802154_MIN_LENGTH = 5,
+    IEEE802154_MAX_LENGTH = 127,
+    IEEE802154_ACK_LENGTH = 5,
+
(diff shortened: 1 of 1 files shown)


dismirlian:
I also considered using the source address as the "unique identifier", but that would require more parsing. What do you think?

dismirlian:
Hi @andrasbiro-silabs, thanks again for reviewing.

> I agree, we don't have a better solution at the moment - either this, or decode the source address in wasAckedWithFramePending. However, I think RAIL could support this functionality better, I created an internal ticket to make the packetHandler accessible from the data request event - once we have that, we can simplify this code a lot.

That would be great!
Also, are there any plans to open source RAIL? That would be even better ;).
 
> For thread 1.2 frame pending, you'll probably need RAIL_IEEE802154_EnableDataFramePending and maybe RAIL_IEEE802154_EnableEarlyFramePending as well.

Yes, I saw them, once Thread 1.2 is out, we can improve the driver.

andrasbiro-silabs:
As far as I'm aware, there's no plan to opensource RAIL. Though I normally prefer and support open source, I personally agree with keeping RAIL closed (I'm an applications engineer working mainly on firmware support).

dismirlian:
That's a pity. I really like the quality of Silabs SDK software, especially emlib/emdrv; I assume RAIL would be no different. In Silabs Thread 19Q2 release document, there is a note that states:

> Silicon Labs Thread has been deprecated in favor of the popular OpenThread implementation. [...] Due to **low customer interest in vendor-specific implementations**, Silicon Labs has decided to remove SL-Thread and there are no plans to expand its functionality.

This is a trend that is not going away: customers will always prefer to have access to source code, especially high quality implementations such as Silabs'. High-quality open software is nowadays as imporant a decisive factor as the capabilities of the ICs you manufacture, or maybe even more important given the amount of time and work that goes into firmware development of complex applications.

Issue #4472, opened 2020-01-13
Packets sent to Service ALOC not reaching destination


mwilniewiec:
Lately an issue with Service ALOCs was found and fixed #4424 

Currently we observe second part of the same problem. 
A child service registers its RLOC in the Leader and it is added to the list but all old RLOCs are not removed. As a result we have 1 server node but multiple Server TLV (0d 02 RLOC16) entries in the network data mostly pointing to some old RLOCs.

During ALOC routing some routers route packet to a valid RLOC but some pick an invalid one and packets are lost.

Im missing something or specification is not addressing this problem too.

librasungirl:
@mwilniewiec would you please describe the process a bit more when the issue happens? e.g. Did this child server attach to a different parent?

mwilniewiec:
Yes the issue arrises when mesh topology is changing. Our server becomes a child of a different parent or changes chlid->router or router->child (generally changes its RLOC). 
When mesh boots up for the first time - server registers itself and leader has only 1 RLOC in network data. When server changes its RLOC it notifies Leader about the change and a new Server TLV is added to the network data.
We ended up with 1 server having 8 different RLOCs is network data. Some garbage collection is needed as there is no chance to route a packet properly in such situation.

librasungirl:
@mwilniewiec Would you please share what the 8 different RLOCs look like?

RLOC is considered temporary in the Thread Mesh Network. It may take some time for the outdated network data info to be removed from the Leader Network Data.

1) If child attaches to a different router, the old parent would deregister the outdated service when timing out the child (up to the ChildTimeout reported during attach, 240s by default).
    [To remove the depracated service sooner, proper shorter Child Timeout might help]

2) If child->router or router->child, when it registers its service, it would include its old RLOC16 when updating its service (https://github.com/openthread/openthread/blob/master/src/core/thread/network_data.cpp#L998) , thus the old server entry would be removed by the Leader before new entry is added.

mwilniewiec:
@librasungirl Sure, this is the network data I've taken from one node:
`0b23
000000c6ac0100
0d 02 b809
0d 02 c812
0d 02 9411
0d 02 f81a
0d 02 ac0c
0d 02 8c0c
0d 02 8c17
080b
0b02 b852
0801 ff
0902 8c17`
You can notice 7 RLOCs here here only one is valid:
0d 02 c812
0d 02 9411
0d 02 f81a
0d 02 ac0c
0d 02 8c0c
0d 02 8c17

[...]

PR #4485, opened 2020-01-17
[network-data] remove stale network data with best efforts

librasungirl:
Rebase and refine previous updates, also introduce network data check for invalid child entry when network data changes according to feedbacks.

See per commit message below for details.

commit1

    [core] clear cached coap requests with rloc source when rloc changes

    This commit helps to reduce unnecessary CoAP retransmission when rloc
    changes. When comes to SRV_DATA.ntf, it also helps to avoid invalid
    network data registration. Take the issue reported in #4472 as example,
    in a Thread Network with packet loss, the child server may switch parents
    frequently, there might be cached SRV_DATA.ntf including old rloc16,
    both the cached one and new one might be sent after switched, causing
    Leader contain two server entries in the Network Data.

commit2

    [network_data] reuse `mLastAttempt` for the function of `mLastAttemptWait`

    This enhancement is based on that `mLastAttempt` with value `0` is
    equivalent to `mLastAttemptWait` with value `false`

commit3

    [network_data] remove `ClearResubmitDelayTimer()` in `UpdateRloc()`

    `UpdateRloc()`, which happens before checking network data consistence,
    should not `ClearResubmitDelayTimer()` to guarantee the DATA_RESUBMIT_DELAY
    before resending when there is inconsistence.

commit4

    [network_data] fix to ensure DATA_RESUBMIT_DELAY between successive SRV_DATA.ntf for server

commit5

    [network_data] introduce 5s delay between successive SRV_DATA.ntf when deregister on behalf of the child server

commit6

    [network-data] deregister for invalid child server if any when receiving new network data

commit7

    [network-data] deregister stale network data when receiving packet destined to ALOC.

    This commit assists the previous commit to help to remove invalid
    server entry in some extreme scenario, for example if child server
    switches parent in a Thread Network with packet loss, and
      a) SRV_DATA.ntf from the old parent to deregister fails
      b) the first attempt of SRV_DATA.ntf from the child fails
    When the child tries to resend SRV_DATA.ntf due to network data inconsistence,
    there will be no old rloc16, that is, the network data may have invalid server entry.

    Though previous commit introduces the invalid child server detection when there is
    new network data, however if the deregister there fails and there is no new
    network data further, there is chance there invalid server entry would
    stick there.
    Here provides one invalid child server detection and recover mechanism
    according to the ALOC destined traffic.

commit8

    [service] fix the minimum cost server selection

commit9

    [router_table] exit earlier when `GetRouter()` for invalid router id

---
> [outdated]
> This PR aims to resolve issues in high packet loss network reported in #4472 
> 
> On one hand, this PR aims to reduce invalid SRV_DATA.ntf, for example
> First commit is used to clear cached CoAP message when RLOC changes in case before successfully register SRV_DATA, the child server switches to a different parent.
> Second commit ensures the SRV_DATA.ntf intervals when detecting network data inconsistence after hearing neighboring MLE Advertisement
> 
> On the other hand, this PR aims to remove stale network data. In the high packet loss network, there is chance that both SRV_DATA.ntf from the server child after switching parent and the SRV_DATA.ntf from old parent for deregister and lost, when the child tries to register again (without old child rloc16 now), the old child rloc16 server entry would be still there. This PR introduce a feedback mechanism - the parent
> of the old child rloc16 would explicitly deregister the service again if destination unreachable when receiving ALOC destined packet.
> However as to how long would it take to recover fresh network data, it may depend on Service Aloc destined traffic.

### DIFF
--- a/src/core/coap/coap.cpp
+++ b/src/core/coap/coap.cpp
@@ -79,6 +79,24 @@ void CoapBase::ClearRequestsAndResponses(void)
     mResponsesQueue.DequeueAllResponses();
 }
 
+void CoapBase::ClearRequests(const Ip6::Address &aAddress)
+{
+    Message *nextMessage;
+
+    // Remove pending messages with the specified source.
+    for (Message *message = static_cast<Message *>(mPendingRequests.GetHead()); message != NULL; message = nextMessage)
+    {
+        CoapMetadata coapMetadata;
+        nextMessage = static_cast<Message *>(message->GetNext());
+        coapMetadata.ReadFrom(*message);
+
+        if (coapMetadata.mSourceAddress == aAddress)
+        {
+            FinalizeCoapTransaction(*message, coapMetadata, NULL, NULL, OT_ERROR_ABORT);
+        }
+    }
+}
+
 otError CoapBase::AddResource(Resource &aResource)
 {
     return mResources.Add(aResource);
--- a/src/core/thread/mesh_forwarder_ftd.cpp
+++ b/src/core/thread/mesh_forwarder_ftd.cpp
@@ -783,7 +783,8 @@ otError MeshForwarder::GetDestinationRlocByServiceAloc(uint16_t aServiceAloc, ui
     {
         NetworkData::NetworkDataTlv *cur = serviceTlv->GetSubTlvs();
         NetworkData::NetworkDataTlv *end = serviceTlv->GetNext();
-        NetworkData::ServerTlv *     server;
+        Neighbor *                   neighbor;
+        uint16_t                     server16;
         uint8_t                      bestCost = Mle::kMaxRouteCost;
         uint8_t                      curCost  = 0x00;
         uint16_t                     bestDest = Mac::kShortAddrInvalid;
@@ -793,12 +794,45 @@ otError MeshForwarder::GetDestinationRlocByServiceAloc(uint16_t aServiceAloc, ui
             switch (cur->GetType())
             {
             case NetworkData::NetworkDataTlv::kTypeServer:
-                server  = static_cast<NetworkData::ServerTlv *>(cur);
-                curCost = Get<Mle::MleRouter>().GetCost(server->GetServer16());
+                server16 = static_cast<NetworkData::ServerTlv *>(cur)->GetServer16();
+
+                // Path cost
+                curCost = Get<Mle::MleRouter>().GetCost(server16);
+
+                if (!Get<Mle::MleRouter>().IsActiveRouter(server16))
+                {
+                    // Assume best link between remote child server and its parent.
+                    curCost += 1;
+                }
+
+                // Cost if the server is direct neighbor.
+                neighbor = Get<Mle::MleRouter>().GetNeighbor(server16);
+
+                if (neighbor != NULL && neighbor->IsStateValid())
+                {
+                    uint8_t cost;
+
+                    if (!Get<Mle::MleRouter>().IsActiveRouter(server16))
+                    {
+                        // Cost calculated only from Link Quality In as the parent only maintains
+                        // one-direction link info.
+                        cost = Mle::MleRouter::LinkQualityToCost(neighbor->GetLinkInfo().GetLinkQuality());
+                    }
+                    else
+                    {
+                        cost = Get<Mle::MleRouter>().GetLinkCost(Mle::Mle::RouterIdFromRloc16(server16));
+                    }
+
+                    // Choose the minimum cost
+                    if (cost < curCost)
+                    {
+                        curCost = cost;
+                    }
+                }
 
                 if ((bestDest == Mac::kShortAddrInvalid) || (curCost < bestCost))
                 {
-                    bestDest = server->GetServer16();
+                    bestDest = server16;
                     bestCost = curCost;
                 }
 
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -969,6 +969,12 @@ void Mle::SetRloc16(uint16_t aRloc16)
     if (aRloc16 != oldRloc16)
     {
         otLogNoteMle("RLOC16 %04x -> %04x", oldRloc16, aRloc16);
+
+        // Clear cached CoAP with old RLOC source
+        if (oldRloc16 != Mac::kShortAddrInvalid)
+        {
+            Get<Coap::Coap>().ClearRequests(mMeshLocal16.GetAddress());
+        }
     }
 
     Get<ThreadNetif>().RemoveUnicastAddress(mMeshLocal16);
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -3821,6 +3821,11 @@ otError MleRouter::CheckReachability(uint16_t aMeshSource, uint16_t aMeshDest, I
         {
             ExitNow();
         }
+        else if (IsAnycastLocator(aIp6Header.GetDestination()))
+        {
+            // Proactively notify Leader of the expired child to de-register the stale service if any.
+            Get<NetworkData::Leader>().SendServerDataNotification(aMeshDest);
+        }
     }
     else if (GetNextHop(aMeshDest) != Mac::kShortAddrInvalid)
     {
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -2606,6 +2606,24 @@ void MleRouter::HandleNetworkDataUpdateRouter(void)
 
     SynchronizeChildNetworkData();
 
+    // Detect if any server entry with invalid child RLOC16
+    {
+        NetworkData::Iterator iterator = NetworkData::kIteratorInit;
+        uint16_t              rloc16   = Mac::kShortAddrInvalid;
+
+        while (Get<NetworkData::Leader>().GetNextServer(iterator, rloc16) == OT_ERROR_NONE)
+        {
+            if (!IsActiveRouter(rloc16) && RouterIdMatch(GetRloc16(), rloc16) &&
+                mChildTable.FindChild(rloc16, Child::kInStateValid) == NULL)
+            {
+                Get<NetworkData::Leader>().SendServerDataNotification(rloc16);
+                // In Thread 1.1 Specification 5.15.6.1, only one RLOC16 TLV entry may appear in SRV_DATA.ntf.
+                // So break here.
+                break;
+            }
+        }
+    }
+
 exit:
     return;
 }
--- a/src/core/thread/network_data.cpp
+++ b/src/core/thread/network_data.cpp
@@ -977,8 +977,16 @@ otError NetworkData::SendServerDataNotification(uint16_t aRloc16)
     Coap::Message *  message = NULL;
     Ip6::MessageInfo messageInfo;
 
-    VerifyOrExit(mLastAttempt.GetValue() == 0 || ((TimerMilli::GetNow() - mLastAttempt) > kDataResubmitDelay),
-                 error = OT_ERROR_ALREADY);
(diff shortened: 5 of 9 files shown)


src/core/thread/mle_router.cpp
```diff
@@ -3802,6 +3802,11 @@ otError MleRouter::CheckReachability(uint16_t aMeshSource, uint16_t aMeshDest, I
         {
             ExitNow();
         }
+        else if (IsAnycastLocator(aIp6Header.GetDestination()))
```
simonlingoogle: :+1: I think this is a great idea.

src/core/thread/network_data.hpp
```diff
@@ -537,7 +537,6 @@ class NetworkData : public InstanceLocator
     };
 
     const Type mType;
-    bool       mLastAttemptWait;
```
simonlingoogle: Is `mLastAttemptWait` useful when the node is just started and current time is within `kDataResubmitDelay` (300s) ?

src/core/thread/network_data_local.cpp
```diff
@@ -344,8 +344,6 @@ void Local::UpdateRloc(void)
             break;
         }
     }
-
-    ClearResubmitDelayTimer();
```
simonlingoogle: Question: what is the reason to remove this statement?

src/core/thread/mle_router.cpp
simonlingoogle: Should we also limit the frequency of sending the server data notification here?

src/core/thread/network_data.hpp
librasungirl: `mLastAttempt` with value 0 would be equivalent to  `mLastAttemptWait` with value`false`.

See the former condition in https://github.com/openthread/openthread/pull/4485/files#diff-c3f068e31220c794c63fd26918c84a85R981.

src/core/thread/mle_router.cpp
librasungirl: Good suggestion. Updated in new patch

src/core/thread/network_data_local.cpp
librasungirl: If initiated by user (e.g. by issuing `netdataregister`), `ClearResubmitDelayTimer` is already done when addinng/removing prefix or service, no necessary to do it here;
If triggered by hearing MLE advertisement, it should be fine to `ClearResubmitDelayTimer()` when network data is consistent (https://github.com/openthread/openthread/pull/4485/files#diff-e518b44189ad2bc7319b004676d0c29fL377), other than in `UpdateRloc()`. Otherwise, it will break the requirement of `DATA_RESUBMIT_DELAY seconds before resending`

librasungirl:
Other thoughts after discussing the issue with the team:
1) Add response handler for the CON CoAP message SRV_DATA.ntf, resend the message if timeout to ensure de-register happens successfullly. 
    - the advantage is to ensure service (de-)register sooner, and always Leader Network Data fresh asap.
    - the disadvantage is that such resending indicates more traffic injection in the original network of higher packet loss

2) Routers may check periodically whether or not the leader network data has child servers which matches its routerid but not a valid child any more, and issue de-register on behalf if there is.
   - The advantage is it may deregister invalid server earlier, compared with the deregister triggered by ALOC destined traffic as in this PR.
   - The disadvantage is a) Router itself may have outdated leader network data, it may be no significant for such de-register; b) the computing cost for such period check especially in a large scale network, whether or not worth to add burdens to all routers for one child server?

@jwhui thoughts?

simonlingoogle:
I would suggest that queued all messages of source address being RLOC should be removed when RLOC16 changes.

jwhui:
In general, I think the device that is authoritative for a given RLOC16 space should be responsible for cleaning up any stale entries in the network data. This includes:

1. The router should remove any network data entries that match the same router id but the child id is invalid.
1. The leader should remove any network data entries that include invalid router ids. **OpenThread already implements this.**

I believe any other solutions will run the risk of having stale entries in the network data.

src/core/coap/coap.cpp
```diff
@@ -79,6 +79,25 @@ void CoapBase::ClearRequestsAndResponses(void)
     mResponsesQueue.DequeueAllResponses();
 }
 
+void CoapBase::ClearRequests(const Ip6::Address &aAddress)
+{
+    Message *    message;
```
jwhui: nit: declare inside for-loop?

src/core/coap/coap.cpp
```diff
@@ -79,6 +79,25 @@ void CoapBase::ClearRequestsAndResponses(void)
...
+    Message *    nextMessage;
+    CoapMetadata coapMetadata;
```
jwhui: nit: declare inside for-loop?

src/core/thread/mle.cpp
```diff
@@ -973,6 +973,11 @@ void Mle::SetRloc16(uint16_t aRloc16)
 
     Get<ThreadNetif>().RemoveUnicastAddress(mMeshLocal16);
 
+    if (Get<Mac::Mac>().GetShortAddress() != Mac::kShortAddrInvalid && Get<Mac::Mac>().GetShortAddress() != aRloc16)
```
jwhui: We have a similar check on RLOC16 above on L969. Can we combine?

src/core/thread/network_data.cpp
```diff
@@ -1010,8 +1009,7 @@ otError NetworkData::SendServerDataNotification(uint16_t aRloc16)
 
     if (mType == kTypeLocal)
     {
-        mLastAttempt     = TimerMilli::GetNow();
-        mLastAttemptWait = true;
+        mLastAttempt = TimerMilli::GetNow();
```
jwhui: What is `TimerMilli::GetNow()` returns `0`?

src/core/thread/network_data.cpp
```diff
@@ -977,7 +977,9 @@ otError NetworkData::SendServerDataNotification(uint16_t aRloc16)
     Coap::Message *  message = NULL;
     Ip6::MessageInfo messageInfo;
 
-    VerifyOrExit(mLastAttempt.GetValue() == 0 || ((TimerMilli::GetNow() - mLastAttempt) > kDataResubmitDelay),
+    VerifyOrExit(mLastAttempt.GetValue() == 0 ||
+                     ((mType == kTypeLocal) && (TimerMilli::GetNow() - mLastAttempt) > kDataResubmitDelay) ||
```
jwhui: Use local variable to store `TimerMilli::GetNow() - mLastAttempt` to avoid computing twice?

src/core/thread/network_data.hpp
```diff
@@ -304,6 +304,18 @@ class NetworkData : public InstanceLocator
      */
     bool ContainsService(uint8_t aServiceId, uint16_t aRloc16);
 
+    /**
+     * This method provides the next server RLOC16 in the Thread Network Data.
+     *
+     * @param[inout]  aIterator  A reference to the Network Data iterator.
+     * @param[out]     aRloc16    The RLOC16 value.
```
jwhui: nit: align `aRloc16`

src/core/thread/network_data.hpp
```diff
@@ -304,6 +304,18 @@ class NetworkData : public InstanceLocator
...
+     *
+     * @retval OT_ERROR_NONE       Successfully found the next server.
+     * @retval OT_ERROR_NOT_FOUND  No subsequent On Mesh prefix exists in the Thread Network Data.
```
jwhui:
```suggestion
     * @retval OT_ERROR_NOT_FOUND  No subsequent server exists in the Thread Network Data.
```

src/core/thread/router_table.cpp
```diff
@@ -383,16 +383,23 @@ Router *RouterTable::GetNeighbor(const Mac::ExtAddress &aExtAddress)
...
     const Router *router = NULL;
-    uint16_t      rloc16 = Mle::Mle::Rloc16FromRouterId(aRouterId);
 
-    for (router = GetFirstEntry(); router != NULL; router = GetNextEntry(router))
+    // Skip if invalid router id is passed.
+    VerifyOrExit(aRouterId < Mle::kInvalidRouterId);
+
     {
-        if (router->GetRloc16() == rloc16)
+        uint16_t rloc16 = Mle::Mle::Rloc16FromRouterId(aRouterId);
```
jwhui: nit: declare `rloc16` at top of method to avoid additional nesting.

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -793,12 +794,45 @@ otError MeshForwarder::GetDestinationRlocByServiceAloc(uint16_t aServiceAloc, ui
...
+                    // assume best link between remote child server and its parent.
+                    curCost += 1;
+                }
+
+                // Cost if the server is direct neighbor.
+                neighbor = Get<Mle::MleRouter>().GetNeighbor(server16);
+
+                if (neighbor != NULL && neighbor->IsStateValid())
+                {
+                    uint8_t cost = Mle::kMaxRouteCost;
```
jwhui: Is it necessary to initialize here?

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -781,9 +781,10 @@ otError MeshForwarder::GetDestinationRlocByServiceAloc(uint16_t aServiceAloc, ui
 
     if (serviceTlv != NULL)
     {
-        NetworkData::NetworkDataTlv *cur = serviceTlv->GetSubTlvs();
-        NetworkData::NetworkDataTlv *end = serviceTlv->GetNext();
-        NetworkData::ServerTlv *     server;
+        NetworkData::NetworkDataTlv *cur      = serviceTlv->GetSubTlvs();
+        NetworkData::NetworkDataTlv *end      = serviceTlv->GetNext();
+        Neighbor *                   neighbor = NULL;
```
jwhui: Is it necessary to initialize here?

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -793,12 +794,45 @@ otError MeshForwarder::GetDestinationRlocByServiceAloc(uint16_t aServiceAloc, ui
...
+                        // info.
+                        cost = Mle::MleRouter::LinkQualityToCost(neighbor->GetLinkInfo().GetLinkQuality());
+                    }
+                    else
+                    {
+                        cost = Get<Mle::MleRouter>().GetLinkCost(Mle::Mle::RouterIdFromRloc16(server16));
+                    }
+
+                    // Choose the minimum cost
+                    if (cost <= curCost)
```
jwhui:
```suggestion
                    if (cost < curCost)
```

src/core/thread/network_data.cpp
librasungirl: In case there is wrap and here returns 0 after ~49days, the impact is that there's chance that be two successive SRV_DATA.ntf in a period shorter than the expected delay.
I am fine to recover if you prefer previous way.

jwhui: I'm OK with one of the following:

1. Check for 0 and set it to 1 if so.

2. Add a comment that shorter delay may occur if `TimerMilli::GetNow()` returns `0`.

src/core/thread/mesh_forwarder_ftd.cpp
jwhui: Yes, I think direct neighbor should be preferred. How does the suggested change affect this? If `cost == curCost` then there is no need to execute `curCost = cost` again.

librasungirl: you are right. will update.

src/core/thread/network_data.cpp
librasungirl: Thanks for the suggestion, I like option 1. will update.

PR #4486, opened 2020-01-20
[mle] misc - log msg change, static method renames

abtink:
This PR contains a group of smaller changes/enhancements mainly in `Mle`, adding/updating log  messages and renaming the static methods which convert from one simple type/value to another:

- [mle] log rlco16 changes 

- [mle] update route table log message 
This commit updates the route table log message to include if link
is established with router or not (router is a neighbor). Also the
`0xffff` is printed as next hop if it is invalid and we indicate
if the entry is for the device itself.

- [mle-router] use routerId as variable name 

- [mle] rename static method to Rloc16FromRouterId() from GetRloc16() 
This commit renames the static `Mle` method which derives RLOC16
from a given Router ID to `Rloc16FromRouterId()` from `GetRloc16()`.
This helps distinguish this method from other `GetRloc16()` methods.

- [mle] rename static method to RouterIdFromRloc16() from GetRouterId() 

- [mle] rename static methods converting between service ALOC and Id 

- [mle] rename static method to ChildIdFromRloc16() from GetChildId() 

- [mesh-forwarder] use Mle::RouterIdMatch() helper method 

- [address-resolver] rename static method to InvalidationReasonToString

### DIFF
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -423,7 +423,7 @@ otError Mle::Restore(void)
         mParent.SetExtAddress(*static_cast<Mac::ExtAddress *>(&parentInfo.mExtAddress));
         mParent.SetDeviceMode(DeviceMode(DeviceMode::kModeFullThreadDevice | DeviceMode::kModeRxOnWhenIdle |
                                          DeviceMode::kModeFullNetworkData | DeviceMode::kModeSecureDataRequest));
-        mParent.SetRloc16(Rloc16FromRouterId(GetRouterId(networkInfo.mRloc16)));
+        mParent.SetRloc16(Rloc16FromRouterId(RouterIdFromRloc16(networkInfo.mRloc16)));
         mParent.SetState(Neighbor::kStateRestored);
 
 #if OPENTHREAD_CONFIG_MLE_INFORM_PREVIOUS_PARENT_ON_REATTACH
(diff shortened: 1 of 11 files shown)


src/core/thread/mle.cpp
```diff
@@ -962,6 +962,13 @@ uint16_t Mle::GetRloc16(void) const
 
 void Mle::SetRloc16(uint16_t aRloc16)
 {
+    uint16_t oldRloc16 = GetRloc16();
```
jwhui: nit: fix typo in commit message "rlco16" -> "rloc16"

src/core/thread/mle_router.cpp
```diff
@@ -1440,19 +1440,19 @@ void MleRouter::UpdateRoutes(const RouteTlv &aRoute, uint8_t aRouterId)
     changed = UpdateLinkQualityOut(aRoute, *neighbor, resetAdvInterval);
 
     // update routes
-    for (uint8_t i = 0, routeCount = 0; i <= kMaxRouterId; i++)
+    for (uint8_t routerId = 0, routeCount = 0; routerId <= kMaxRouterId; routerId++)
```
jwhui: There are other instances of using `i` rather than `routerId` in this file. Should we update those as well?

jwhui:
I assume the code size increase is due to the logging enhancements?

abtink:
Yes. The code size increase is from the new/changed logs.

PR #4487, opened 2020-01-20
[meshcop] fix premature free of finalize message

wgtdkp:
This PR fix the bug that the `finalize message` will be prematurely freed if the joiner is started at the second time without waiting for the finish of the first initiation.

### DIFF
--- a/src/core/meshcop/joiner.cpp
+++ b/src/core/meshcop/joiner.cpp
@@ -450,13 +450,10 @@ exit:
 
 void Joiner::FreeJoinerFinalizeMessage(void)
 {
-    VerifyOrExit(mState == OT_JOINER_STATE_IDLE);
+    VerifyOrExit(mState == OT_JOINER_STATE_IDLE && mFinalizeMessage != NULL);
 
-    if (mFinalizeMessage != NULL)
-    {
-        mFinalizeMessage->Free();
-        mFinalizeMessage = NULL;
-    }
+    mFinalizeMessage->Free();
+    mFinalizeMessage = NULL;
 
 exit:
     return;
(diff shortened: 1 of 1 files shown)


src/core/meshcop/joiner.cpp
```diff
@@ -129,7 +129,10 @@ otError Joiner::Start(const char *     aPskd,
     if (error != OT_ERROR_NONE)
     {
         otLogInfoMeshCoP("Error %s while starting joiner", otThreadErrorToString(error));
-        FreeJoinerFinalizeMessage();
+        if (mState == OT_JOINER_STATE_IDLE)
```
librasungirl: How about initializing `error` as `OT_ERROR_NONE` at the begining?

Line#101 has the status check already, here not necessary to check again if `error` properly initiazlied.

jwhui: How about moving this check into `FreeJoinerFinalizeMessage()`? This change would also require moving the call to `FreeJoinerFinalizeMessage()` in `Joiner::Finish()` to after the `SetState(OT_JOINER_STATE_IDLE)` call. Thoughts?

wgtdkp: Makes sense to me.
Done.

src/core/meshcop/joiner.cpp
```diff
@@ -450,11 +450,16 @@ otError Joiner::PrepareJoinerFinalizeMessage(const char *aProvisioningUrl,
 
 void Joiner::FreeJoinerFinalizeMessage(void)
 {
+    VerifyOrExit(mState == OT_JOINER_STATE_IDLE);
```
jwhui: nit: include the NULL check in this `VerifyOrExit`?

wgtdkp: made it in separate `VerifyOrExit` to make it clearer.

PR #4489, opened 2020-01-21
[codecov] ignore third_party code

bukepo:
From https://docs.codecov.io/docs/ignoring-paths, `third_party/*` will not recursively ignore all files in `third_party`. This PR changes `.codecov.yml` to use the correct pattern `third_party/**/*`.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,7 +5,7 @@ coverage:
 
 ignore:
   - "tests/*"
-  - "third_party/*"
+  - "third_party/**/*"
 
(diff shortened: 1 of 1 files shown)


jwhui:
> From https://docs.codecov.io/docs/ignoring-paths, `third_party/*` will not recursively ignore all files in `third_party`. This PR changes `.codecov.yml` to use the correct pattern `third_party/**/*`.

Good catch. It appears the codecov docs at the link above was updated about a month ago.

PR #4490, opened 2020-01-21
[ncp] simplify unimplemented commands

bukepo:


src/ncp/ncp_base_mtd.cpp
```diff
@@ -129,7 +129,7 @@ static uint8_t ExternalRoutePreferenceToFlagByte(int aPreference)
         break;
 
     default:
-        flags = SPINEL_ROUTE_PREFERENCE_MEDIUM;
+        assert(false);
```
jwhui: I'm not sure we want an assert here. We should be careful not to assert simply because of unexpected information received from the network.

PR #4493, opened 2020-01-21
[codecov] cover swapping flash block of settings

bukepo:
This PR adds a self test to cover swapping block functionality. Since
find will include the `.o` file generated by the unittest program, how
the platform library find objects of the utils library is also updated.

### DIFF
--- a/examples/platforms/cc1352/Makefile.am
+++ b/examples/platforms/cc1352/Makefile.am
@@ -77,9 +77,8 @@ PRETTY_FILES                                                                    
     $(PLATFORM_SOURCES)                                                                   \
     $(NULL)
 
-Dash = -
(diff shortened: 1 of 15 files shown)


bukepo:
@jwhui any comments?

bukepo:
Ping~

jwhui:
@bukepo, I would like to move the `settings_flash.c` implementation into core. A build config can choose whether the OpenThread library calls `otPlatSettings*` APIs or something like `otPlatFlash*` APIs. Are you OK with waiting to merge this PR until after this move is complete?

bukepo:
Yes, thanks for the information.

bukepo:
I don't think the new implementation needs this change anymore. I will close this for now.

PR #4494, opened 2020-01-21
[meshcop] fix commissioner's invalid call to SetState

simonlingoogle:
Commissioner should not call `SetState(OT_COMMISSIONER_STATE_DISABLED)` directly, rather it should call `Stop()` so that joiners and resources are cleaned up.

This PR ...
 - Fixes this issue
 - Adds argument name `/* aResign */` to Stop() calls
 - Ignores petition/keep-alive response if commissioner is not in PETITION/ACTIVE state
 - Calls `CoapSecure::Stop` when `Start` failed (suggested by @bukepo )
 - Responds to `Leader Petition Response (accept)` by sending `KeepAlive(reject)` if commissioner is in disabled state

### DIFF
--- a/src/core/meshcop/commissioner.cpp
+++ b/src/core/meshcop/commissioner.cpp
@@ -730,8 +730,6 @@ void Commissioner::HandleLeaderPetitionResponse(Coap::Message *         aMessage
         ExitNow();
     }
 
-    mSessionId = sessionId.GetCommissionerSessionId();
-
     Get<Mle::MleRouter>().GetCommissionerAloc(mCommissionerAloc.GetAddress(), mSessionId);
     Get<ThreadNetif>().AddUnicastAddress(mCommissionerAloc);
 
@@ -763,11 +761,9 @@ otError Commissioner::SendKeepAlive(void)
 
 otError Commissioner::SendKeepAlive(uint16_t aSessionId)
 {
-    otError                  error   = OT_ERROR_NONE;
-    Coap::Message *          message = NULL;
-    Ip6::MessageInfo         messageInfo;
-    StateTlv                 state;
-    CommissionerSessionIdTlv sessionId;
+    otError          error   = OT_ERROR_NONE;
+    Coap::Message *  message = NULL;
+    Ip6::MessageInfo messageInfo;
 
     VerifyOrExit((message = NewMeshCoPMessage(Get<Coap::Coap>())) != NULL, error = OT_ERROR_NO_BUFS);
 
@@ -778,7 +774,7 @@ otError Commissioner::SendKeepAlive(uint16_t aSessionId)
         error = Tlv::AppendUint8Tlv(*message, Tlv::kState,
                                     (mState == OT_COMMISSIONER_STATE_ACTIVE) ? StateTlv::kAccept : StateTlv::kReject));
 
-    SuccessOrExit(error = Tlv::AppendUint16Tlv(*message, Tlv::kCommissionerSessionId, mSessionId));
+    SuccessOrExit(error = Tlv::AppendUint16Tlv(*message, Tlv::kCommissionerSessionId, aSessionId));
 
     messageInfo.SetSockAddr(Get<Mle::MleRouter>().GetMeshLocal16());
     SuccessOrExit(error = Get<Mle::MleRouter>().GetLeaderAloc(messageInfo.GetPeerAddr()));
--- a/src/core/meshcop/commissioner.cpp
+++ b/src/core/meshcop/commissioner.cpp
@@ -164,27 +164,42 @@ otError Commissioner::Start(otCommissionerStateCallback  aStateCallback,
     SetState(OT_COMMISSIONER_STATE_PETITION);
 
 exit:
+    if (error != OT_ERROR_NONE)
+    {
+        Get<Coap::CoapSecure>().Stop();
+    }
     return error;
 }
 
-otError Commissioner::Stop(void)
+otError Commissioner::Stop(bool aResign)
 {
-    otError error = OT_ERROR_NONE;
+    otError error      = OT_ERROR_NONE;
+    bool    needResign = false;
 
     VerifyOrExit(mState != OT_COMMISSIONER_STATE_DISABLED, error = OT_ERROR_INVALID_STATE);
 
     Get<Coap::CoapSecure>().Stop();
 
-    Get<ThreadNetif>().RemoveUnicastAddress(mCommissionerAloc);
-    RemoveCoapResources();
-    ClearJoiners();
-    mTransmitAttempts = 0;
+    if (mState == OT_COMMISSIONER_STATE_ACTIVE)
+    {
+        Get<ThreadNetif>().RemoveUnicastAddress(mCommissionerAloc);
+        RemoveCoapResources();
+        ClearJoiners();
+        needResign = true;
+    }
+    else if (mState == OT_COMMISSIONER_STATE_PETITION)
+    {
+        mTransmitAttempts = 0;
+    }
 
     mTimer.Stop();
 
     SetState(OT_COMMISSIONER_STATE_DISABLED);
 
-    SendKeepAlive();
+    if (needResign && aResign)
+    {
+        SendKeepAlive();
+    }
 
 exit:
     return error;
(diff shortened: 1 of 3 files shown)


src/core/meshcop/commissioner.cpp
```diff
@@ -706,15 +715,15 @@ void Commissioner::HandleLeaderPetitionResponse(Coap::Message *         aMessage
     CommissionerSessionIdTlv sessionId;
     bool                     retransmit = false;
 
-    VerifyOrExit(mState == OT_COMMISSIONER_STATE_PETITION, SetState(OT_COMMISSIONER_STATE_DISABLED));
+    VerifyOrExit(mState == OT_COMMISSIONER_STATE_PETITION, Stop(false));
```
wgtdkp: Why should the commissioner stop itself if it received a `COMM_PET.rsp` in an invalid state? Why not ignore the unexpected `COMM_PET.rsp`?

src/core/meshcop/commissioner.cpp
```diff
@@ -735,7 +744,7 @@ void Commissioner::HandleLeaderPetitionResponse(Coap::Message *         aMessage
     {
         if (mTransmitAttempts >= kPetitionRetryCount)
         {
-            SetState(OT_COMMISSIONER_STATE_DISABLED);
+            Stop(true);
```
wgtdkp: We are not in active state, so I think the commissioner should not send the keep alive message (whether to keep it self alive or stop itself gracefully).

src/core/meshcop/commissioner.cpp
```diff
@@ -800,26 +809,20 @@ void Commissioner::HandleLeaderKeepAliveResponse(Coap::Message *         aMessag
 
     StateTlv state;
 
-    VerifyOrExit(mState == OT_COMMISSIONER_STATE_ACTIVE, SetState(OT_COMMISSIONER_STATE_DISABLED));
-    VerifyOrExit(aResult == OT_ERROR_NONE && aMessage->GetCode() == OT_COAP_CODE_CHANGED,
-                 SetState(OT_COMMISSIONER_STATE_DISABLED));
+    VerifyOrExit(mState == OT_COMMISSIONER_STATE_ACTIVE, Stop(false));
```
wgtdkp: No need to stop the commissioner. Maybe ignore the message.

src/core/meshcop/commissioner.cpp
wgtdkp: nit: add comment to the boolean argument of `Stop`? Like this `Stop(/* aSendKeepAlive */ false)`.

simonlingoogle: > nit: add comment to the boolean argument of `Stop`? Like this `Stop(/* aSendKeepAlive */ false)`.

IDEs like `Clion` can actually do the job for you by showing you the argument names. 
I think the comment is helpful, but I haven't seen it been used in `openthread` code.

src/core/meshcop/commissioner.cpp
simonlingoogle: I agree with what you said. The current code is same as it was.

src/core/meshcop/commissioner.cpp
simonlingoogle: The current code will cause a petitioning commissioner to stop if KeepAlive response is received. 
 I think the chance to happen is negligible.

src/core/meshcop/commissioner.cpp
wgtdkp: Search `(/*` to find example code. The comment is only useful for arguments that can't easily tell its meaning ( such as a boolean argument).

src/core/meshcop/commissioner.cpp
wgtdkp: yes, this will rarely happen if all things are going well. But, we would say, "stop the commissioner when receiving an unexpected message" is wrong. Agree?

src/core/meshcop/commissioner.hpp
```diff
@@ -84,11 +84,13 @@ class Commissioner : public InstanceLocator
...
      * This method stops the Commissioner service.
      *
+     * @param[in]  aSendKeepAlive      Whether send LEAD_KA.req to resign as Commissioner
+     *
      * @retval OT_ERROR_NONE           Successfully stopped the Commissioner service.
      * @retval OT_ERROR_INVALID_STATE  Commissioner is already stopped.
      *
      */
-    otError Stop(void);
+    otError Stop(bool aSendKeepAlive);
```
bukepo:
```suggestion
    otError Stop(bool aResign);
```

src/core/meshcop/commissioner.cpp
```diff
@@ -705,15 +714,15 @@ void Commissioner::HandleLeaderPetitionResponse(Coap::Message *         aMessage
     CommissionerSessionIdTlv sessionId;
     bool                     retransmit = false;
 
-    VerifyOrExit(mState == OT_COMMISSIONER_STATE_PETITION, SetState(OT_COMMISSIONER_STATE_DISABLED));
+    VerifyOrExit(mState == OT_COMMISSIONER_STATE_PETITION);
```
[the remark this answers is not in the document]
xiaonanln: If the commissioner is stoped during petition, it should have already called `Stop(/* aResign */ true)` when executing `otCommissionerStop`, thus already sent the `KeepAlive (reject)`. No need to send one more. Thoughts?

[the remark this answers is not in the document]
simonlingoogle: If the commissioner is stoped during petition, it should have already called Stop(/* aResign */ true) when executing otCommissionerStop, thus already sent the KeepAlive (reject). No need to send one more. Thoughts?

src/core/meshcop/commissioner.cpp
```diff
@@ -164,27 +164,40 @@ otError Commissioner::Start(otCommissionerStateCallback  aStateCallback,
     SetState(OT_COMMISSIONER_STATE_PETITION);
 
 exit:
+    if (error != OT_ERROR_NONE)
+    {
+        Get<Coap::CoapSecure>().Stop();
```
simonlingoogle: @bukepo Would calling `Get<Coap::CoapSecure>().Stop()` cause any issue if `Get<Coap::CoapSecure>().Start` is not executed. This could happen if `Commissioner::Start` failed in the first two lines of `VerifyOrExit`

bukepo: If it would, it should be fixed. I think we Stop() should be a idempotent operation. Thoughts?

src/core/meshcop/commissioner.cpp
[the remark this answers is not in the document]
bukepo: If a node hasn't gotten a petition response, it doesn't have the leader allocated commissioner session id to be included in LEAD_KA .req yet.

src/core/meshcop/commissioner.cpp
simonlingoogle: Agree that `CoapSecure::Stop` should be idempotent. Couldn't find anything non-idempotent along the call tree.

src/core/meshcop/commissioner.cpp
```diff
@@ -164,27 +164,40 @@ otError Commissioner::Start(otCommissionerStateCallback  aStateCallback,
...
+    }
     return error;
 }
 
-otError Commissioner::Stop(void)
+otError Commissioner::Stop(bool aResign)
```
librasungirl: is it necessary to have the `bool aResign` parameter?
shouldn't the resign only be required when stopping an active commissioner?

if the commissioner is stopped before receiving the PET.rsp with accept status, the commissioner itself won't enter active mode, and the leader will expire commissioner session automatically after keep-alive timeout.

src/core/meshcop/commissioner.cpp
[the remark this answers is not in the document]
simonlingoogle: @bukepo added sending KA reject in disabled state accordingly

src/core/meshcop/commissioner.cpp
```diff
@@ -164,27 +164,40 @@ otError Commissioner::Start(otCommissionerStateCallback  aStateCallback,
...
+    {
+        mTransmitAttempts = 0;
+    }
 
     mTimer.Stop();
 
     SetState(OT_COMMISSIONER_STATE_DISABLED);
 
-    SendKeepAlive();
+    if (aResign)
```
librasungirl: the resign keep alive should only required when stopping an active commissioner, not necessary if it is still candidate.

simonlingoogle: There are two cases when an active commissioner can be stopped: 
* When commissioner session is rejected in KeepAlive response, the commissioner need to stop, but no need to `SendKeepAlive (reject)`. 
* When commissioner session is stopped by API, the commissioner need to send `SendKeepAlive (reject)`.

So, the commissioner can not determine if it's necessary to `SendKeepAlive()` only by checking state. `aResign` argument is required at least for the active state.

librasungirl: I mean this resign here -  if stopped by API, there should not be `SendKeepAlive(reject)` if it is still in petition. 

for the existing code here, may introduce a bool variable to record whether or not it was an  active commissioner, and here the condition should be (aResign && boolean variable)
make sense?

simonlingoogle: Yes, we can use more conditions to avoid this `KeepAlive reject`. 
If commissioner is stopped by API in petition state, it will send `KeepAlive reject` with current `mSessionId` which I believe should be 0 or last active session Id. Either case, the `KeepAlive reject` has incorrect session ID.

wgtdkp: @librasungirl I think the commissioner state perfectly indicates whether it is active. No need for extra flags. right? We just need to change commissioner state to `DISABLED` later.

librasungirl: In exiting implementation, `SendKeepAlive()` would has state `kAccept` or `kReject` according to the commissioner state.
https://github.com/openthread/openthread/blob/master/src/core/meshcop/commissioner.cpp#L760

`DISABLED` should be set before call `SendKeepAlive()`

wgtdkp: @librasungirl You are right about the implementation. But I don't think it make sense to have a commissioner send a keep alive message when it is in `DISABLED` state. Once a commissioner became `DISABLED`, it shouldn't be able  to send any messages. Agree?

So I would suggest changing signature `SendKeepAlive()` to  `SendKeepAlive(bool aAccept)`, and moving `SetState(...)` to the end of this function. Thoughts?

simonlingoogle: > So I would suggest changing signature `SendKeepAlive()` to `SendKeepAlive(bool aAccept)`, and moving `SetState(...)` to the end of this function. 

@wgtdkp I think you are right. The current `SendKeepAlive` implementation is **implicitly** using current commissioner state to determine if it's a reject or accept. Could be a surprise for people read the code.

simonlingoogle: I think it's better to make this PR a dedicated fix for handling commissioning messages. 
@wgtdkp would you create another PR for refactoring `SendKeepAlive` prototype? 
Thoughts? @librasungirl @wgtdkp @bukepo

wgtdkp: SGTM.

simonlingoogle: > for the existing code here, may introduce a bool variable to record whether or not it was an active commissioner, and here the condition should be (aResign && boolean variable)

added variable `needResign` for this purpose.

simonlingoogle:
@librasungirl @bukepo , 
Any further comments for this PR ?

librasungirl:
@simonlingoogle some conflicts to resolve~

PR #4499, opened 2020-01-22
[settings] zero out any trailing bytes in buffer

jwhui:
If the setting value read from non-volatile is smaller than the
expected size, make sure to zero out the remainder of the buffer to
ensure that the additional fields are initialized.

### DIFF
--- a/src/core/common/settings.cpp
+++ b/src/core/common/settings.cpp
@@ -119,11 +119,12 @@ otError Settings::SaveOperationalDataset(bool aIsActive, const MeshCoP::Dataset 
 
 otError Settings::ReadOperationalDataset(bool aIsActive, MeshCoP::Dataset &aDataset) const
 {
-    otError  error = OT_ERROR_NONE;
-    uint16_t length;
+    otError  error  = OT_ERROR_NONE;
+    uint16_t length = MeshCoP::Dataset::kMaxSize;
 
-    SuccessOrExit(error = Read(aIsActive ? kKeyActiveDataset : kKeyPendingDataset, aDataset.GetBytes(),
-                               MeshCoP::Dataset::kMaxSize, length));
+    SuccessOrExit(error = Read(aIsActive ? kKeyActiveDataset : kKeyPendingDataset, aDataset.GetBytes(), length));
+    VerifyOrExit(length <= MeshCoP::Dataset::kMaxSize, error = OT_ERROR_NOT_FOUND);
+
     aDataset.SetSize(length);
 
 exit:
@@ -141,10 +142,11 @@ otError Settings::DeleteOperationalDataset(bool aIsActive)
 
 otError Settings::ReadNetworkInfo(NetworkInfo &aNetworkInfo) const
 {
-    otError error;
+    otError  error;
+    uint16_t length = sizeof(NetworkInfo);
 
     aNetworkInfo.Init();
-    SuccessOrExit(error = ReadFixedSize(kKeyNetworkInfo, &aNetworkInfo, sizeof(NetworkInfo)));
+    SuccessOrExit(error = Read(kKeyNetworkInfo, &aNetworkInfo, length));
     LogNetworkInfo("Read", aNetworkInfo);
 
 exit:
@@ -155,8 +157,9 @@ otError Settings::SaveNetworkInfo(const NetworkInfo &aNetworkInfo)
 {
     otError     error = OT_ERROR_NONE;
     NetworkInfo prevNetworkInfo;
+    uint16_t    length = sizeof(prevNetworkInfo);
 
-    if ((ReadFixedSize(kKeyNetworkInfo, &prevNetworkInfo, sizeof(NetworkInfo)) == OT_ERROR_NONE) &&
+    if ((Read(kKeyNetworkInfo, &prevNetworkInfo, length) == OT_ERROR_NONE) && (length == sizeof(NetworkInfo)) &&
         (memcmp(&prevNetworkInfo, &aNetworkInfo, sizeof(NetworkInfo)) == 0))
     {
         LogNetworkInfo("Re-saved", aNetworkInfo);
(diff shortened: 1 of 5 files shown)


src/core/common/settings.cpp
```diff
@@ -305,6 +305,8 @@ otError Settings::ReadFixedSize(Key aKey, void *aBuffer, uint16_t aExpectedSize)
     SuccessOrExit(error = otPlatSettingsGet(&GetInstance(), aKey, 0, reinterpret_cast<uint8_t *>(aBuffer), &size));
     VerifyOrExit(size >= aExpectedSize, error = OT_ERROR_NOT_FOUND);
 
+    memset(aBuffer, 0, aExpectedSize - size);
```
bukepo:
```suggestion
    memset(aBuffer, 0, size - aExpectedSize);
```

src/core/common/settings.cpp
```diff
@@ -305,6 +305,8 @@ otError Settings::ReadFixedSize(Key aKey, void *aBuffer, uint16_t aExpectedSize)
...
+
 exit:
```
abtink: On return from `otPlatSettingsGet()`, the `size` variable should be updated to give the size of `aKey` in the settings.
And we should get here only if `size >= aExpectedSize` (from the `VerifyOrExit` line above), so I think the entire size of `aBuffer` (which is `aExpectedSize`) should already be filled from `otPlatSettingsGet()` call? .
 Maybe I am missing something?

jwhui: This PR is incomplete. In summary, we need to relax the size constraint to allow for SWU where the new version has appended values to the settings structure. If you agree, I will update this PR.

abtink: Got it. Sounds reasonable to me.

We can consider adding a new method to handle such a case (i.e., allow the saved value to be smaller than what we expect to read) or we can use/update `Read()` to `memset` the left over in the given buffer and use it instead?. The advantage I see by having a different method is that we can indicate which keys we would allow incomplete saved data and which ones we won't.

Thoughts?

bukepo: How about requiring the structure of `aBuffer` is already correctly initialized? So even the read length is smaller, there will be no uninitialized value or **illegal** values? Zero is not guaranteed to be legal values.

bukepo: @jwhui Thoughts?

jwhui: We could treat zero as "uninitialized". If the field is zero after reading, we can then initialize it to some valid value. Given that the `PACKED` declaration was just merged, treating zero as "uninitialized" would also handle cases where there was padding before `PACKED`. Thoughts?

bukepo: It seems this way the structure's logic will be scattered in the reading method. If we provide a Init() method for each structure and call it before reading from non-volatile, we can keep the read method logic clear and prevent from adding a lot extra initializing task in future. Thoughts?

jwhui: This is in fact what I am implementing. However, the initial implementation will simply memset the structure to 0 for simplicity. In the future, we can revisit if there is a need to initialize to non-zero values.

bukepo: OK, I will initialize the newly added `mVersion` to a proper value in #4325.

src/core/common/settings.hpp
```diff
@@ -584,7 +584,9 @@ class Settings : public SettingsBase
      */
     otError ReadSlaacIidSecretKey(Utils::Slaac::IidSecretKey &aKey)
     {
-        return ReadFixedSize(kKeySlaacIidSecretKey, &aKey, sizeof(Utils::Slaac::IidSecretKey));
+        uint16_t length = sizeof(Utils::Slaac::IidSecretKey);
+
+        return Read(kKeySlaacIidSecretKey, &aKey, length);
```
bukepo: Should it return error if the read length is smaller?

jwhui: The length check has been removed from all the other reads, which implies that we trust the settings driver implementation to have some value and length that we expect. Is there a use case here that does not apply to the other settings?

bukepo: It can be helpful when debugging, but I'm fine with this assumption.

src/core/common/settings.cpp
```diff
@@ -154,8 +155,9 @@ otError Settings::SaveNetworkInfo(const NetworkInfo &aNetworkInfo)
 {
     otError     error = OT_ERROR_NONE;
     NetworkInfo prevNetworkInfo;
+    uint16_t    length;
```
abtink:
```suggestion
    uint16_t    length = sizeof(NetworkInfo);
```
I think we need to init `length` since it would be used by `Read()`.

src/core/common/settings.cpp
```diff
@@ -119,11 +119,10 @@ otError Settings::SaveOperationalDataset(bool aIsActive, const MeshCoP::Dataset
...
 otError Settings::ReadOperationalDataset(bool aIsActive, MeshCoP::Dataset &aDataset) const
 {
-    otError  error = OT_ERROR_NONE;
-    uint16_t length;
+    otError  error  = OT_ERROR_NONE;
+    uint16_t length = MeshCoP::Dataset::kMaxSize;
 
-    SuccessOrExit(error = Read(aIsActive ? kKeyActiveDataset : kKeyPendingDataset, aDataset.GetBytes(),
-                               MeshCoP::Dataset::kMaxSize, length));
+    SuccessOrExit(error = Read(aIsActive ? kKeyActiveDataset : kKeyPendingDataset, aDataset.GetBytes(), length));
```
abtink: Just to be safe here, should we do a `VerifyOrExit(length < MeshCoP::Dataset::kMaxSize);`? (or maybe `assert()` instead)?
The `length` upon return from `Read()` would be number of bytes available for the given key (not the number of bytes actually read and copied into the buffer).

PR #4500, opened 2020-01-23
[cmake] add additional compiler options

jwhui:


CiaranWoodward:
3 minor issues I see here - but glad to see cmake getting continued attention!

#### Breaks non-gcc-compatible compilers
This will cause all non-gcc compatible compilers to fail due to unsupported arguments. This could be worked around by using either an if:
```
if(${CMAKE_C_COMPILER_ID} STREQUAL "GNU" OR ${CMAKE_C_COMPILER_ID} STREQUAL "Clang")
    set(OT_CFLAGS -Wall -Wextra -Wshadow -Werror -std=c99 -pedantic-errors)
    set(OT_CXXFLAGS -Wall -Wextra -Wshadow -Werror -std=gnu++98 -Wno-c++14-compat -fno-exceptions)
endif()
```
or using even deeper generator expressions:
```
set(OT_CFLAGS "$<$<AND:$<COMPILE_LANGUAGE:C>,$<OR:$<C_COMPILER_ID:GNU>,$<C_COMPILER_ID:Clang>>:-Wall -Wextra -Wshadow -Werror -std=c99 -pedantic-errors">)
```
Personally I prefer the first option because ifs are easier to read, and the compiler will not change between configure and generate time. 

Also minor style, but it might be neater for future maintenance to put the generator expression into the variable, like:
```
set(OT_CFLAGS "$<$<COMPILE_LANGUAGE:C>:-Wall -Wextra -Wshadow -Werror -std=c99 -pedantic-errors>")
set(OT_CXXFLAGS "$<$<COMPILE_LANGUAGE:CXX>:-Wall -Wextra -Wshadow -Werror -std=gnu++98 -Wno-c++14-compat -fno-exceptions")
```

#### CMake convention to set c standard is by using target properties
This allows cmake to do the menial work of figuring out appropriate flags to add for different compilers. eg:
```
set_property(TARGET openthread-ftd PROPERTY CXX_STANDARD 98)
set_property(TARGET openthread-ftd PROPERTY C_STANDARD 99)
```

#### -Werror
I guess this is down to developer preference, but while I agree Werror is useful to have in CI, it can be frustrating to develop with. Would prefer if it was something that had to be explicitly enabled (or at least have an option to disable).

jwhui:
> I think we are using cmake build to check NCP handlers are sorted because it supports c++11. With this change we may need a specific build target for it.

Good point. I removed the C++ standard specification from NCP for now. We can rely on the autotools-based build for the C++ standard check for now.

jwhui:
@CiaranWoodward, thanks for the review and suggestions. I believe I have applied your suggestions to this PR. Let me know what you think.

src/core/CMakeLists.txt
```diff
@@ -30,6 +30,13 @@ add_library(openthread-ftd)
 add_library(openthread-mtd)
 add_library(openthread-radio)
 
+set_property(TARGET openthread-ftd PROPERTY C_STANDARD 99)
+set_property(TARGET openthread-ftd PROPERTY CXX_STANDARD 98)
+set_property(TARGET openthread-mtd PROPERTY C_STANDARD 99)
+set_property(TARGET openthread-mtd PROPERTY CXX_STANDARD 98)
+set_property(TARGET openthread-radio PROPERTY C_STANDARD 99)
+set_property(TARGET openthread-radio PROPERTY CXX_STANDARD 98)
```
CiaranWoodward: Potential, optional style:
```
set_target_properties(
    openthread-ftd openthread-mtd openthread-radio
    PROPERTIES
        C_STANDARD 99
        CXX_STANDARD 98
)
```

src/core/CMakeLists.txt
```diff
@@ -45,6 +52,21 @@ target_compile_definitions(openthread-radio PRIVATE
     OPENTHREAD_RADIO=1
 )
 
+target_compile_options(openthread-ftd PRIVATE
+    ${OT_CFLAGS}
+    ${OT_CXXFLAGS}
```
CiaranWoodward: Potentially might be worth combining ``OT_CFLAGS`` and ``OT_CXXFLAGS`` into one ``OT_FLAGS`` just to reduce boilerplate, but no strong feelings.

PR #4503, opened 2020-01-23
[mac] simplify scan handlers

abtink:
This commit changes the energy and active scan callback handlers in
`Mac` class to make them follow the model used by external OpenThread
scan APIs. This in turn simplify their use and removes the need to
save energy/active callbacks and their context in `Instance` object.

### DIFF
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -106,6 +106,7 @@ Mac::Mac(Instance &aInstance)
     , mMaxFrameRetriesIndirect(kDefaultMaxFrameRetriesIndirect)
 #endif
     , mActiveScanHandler(NULL) // Initialize `mActiveScanHandler` and `mEnergyScanHandler` union
+    , mScanHandlerContext(NULL)
     , mSubMac(aInstance)
     , mOperationTask(aInstance, &Mac::HandleOperationTask, this)
     , mTimer(aInstance, &Mac::HandleTimer, this)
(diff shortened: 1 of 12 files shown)


src/core/mac/mac.cpp
```diff
@@ -275,11 +278,31 @@ void Mac::PerformActiveScan(void)
     {
         mSubMac.SetPanId(mPanId);
         FinishOperation();
-        mActiveScanHandler(GetInstance(), NULL);
+        ReportActiveScanResult(NULL);
         PerformNextOperation();
     }
 }
 
+void Mac::ReportActiveScanResult(RxFrame *aBeaconFrame)
```
bukepo: Is it possible to
```suggestion
void Mac::ReportActiveScanResult(const RxFrame *aBeaconFrame)
```

abtink: 👍  was thinking of the same. 
It requires adding some changes in `Beacon` class and `ConvertBeaconToActiveScanResult` (missing  some `const` defs/method). Will add it later as a separate commit.

src/core/mac/mac.hpp
```diff
@@ -162,12 +155,13 @@ class Mac : public InstanceLocator
      * @param[in]  aScanDuration     The time in milliseconds to spend scanning each channel. If the duration is set to
      *                               zero, a single RSSI sample will be taken per channel.
      * @param[in]  aHandler          A pointer to a function called to pass on scan result or indicate scan completion.
+     * @param[in]  aContext          A pointer to an arbitrary context (used when invoking `aHandler` callback).
```
jwhui:
```suggestion
     * @param[in]  aContext          A pointer to an arbitrary context (used when invoking @p aHandler callback).
```

src/core/mac/mac_frame.hpp
```diff
@@ -1164,13 +1164,21 @@ class Beacon
     }
 
     /**
-     * This method returns the pointer to the beacon payload address.
+     * This method returns the pointer to the beacon payload.
      *
-     * @retval A pointer to the beacon payload address.
+     * @retval A pointer to the beacon payload.
```
jwhui:
```suggestion
     * @returns A pointer to the beacon payload.
```

src/core/mac/mac_frame.hpp
```diff
@@ -1164,13 +1164,21 @@ class Beacon
...
      *
      */
     uint8_t *GetPayload(void) { return reinterpret_cast<uint8_t *>(this) + sizeof(*this); }
 
+    /**
+     * This method returns the pointer to the beacon payload.
+     *
+     * @retval A pointer to the beacon payload.
```
jwhui:
```suggestion
     * @returns A pointer to the beacon payload.
```

src/core/mac/mac.cpp
abtink: Added it in the main commit itself.

PR #4506, opened 2020-01-23
[efr32] Enhancements to radio driver (3/3)

dismirlian:
This is the last of 3 PRs addressing the efr32mg12 radio driver.

PR# 3:

- Features:
  - Allow per-band PA configurations:
    - Supply selection (DCDC or VBAT).
    - Ramp time configuration.
    - Voltage configuration.

- Bug fixes:
  - Corrected PA supply selection for BRD4166A (is supplied from DCDC instead of VBAT). This could cause an incorrect output power setting.

@andrasbiro-silabs please review.

Thanks

### DIFF
--- a/examples/platforms/efr32mg12/radio.c
+++ b/examples/platforms/efr32mg12/radio.c
@@ -186,7 +186,11 @@ static const RAIL_IEEE802154_Config_t sRailIeee802154Config = {
     .isPanCoordinator = false,
 };
 
+#if RADIO_CONFIG_PA_USES_DCDC
+RAIL_DECLARE_TX_POWER_DCDC_CURVES(piecewiseSegments, curvesSg, curves24Hp, curves24Lp);
+#else
 RAIL_DECLARE_TX_POWER_VBAT_CURVES(piecewiseSegments, curvesSg, curves24Hp, curves24Lp);
+#endif
 
 static int8_t sTxPowerDbm = OPENTHREAD_CONFIG_DEFAULT_TRANSMIT_POWER;
 
--- a/examples/platforms/efr32mg12/brd4161a/board_config.h
+++ b/examples/platforms/efr32mg12/brd4161a/board_config.h
@@ -45,4 +45,6 @@
 #define RADIO_CONFIG_DMP_SUPPORT 0            /// Set to 1 to enable Dynamic Multi-Protocol support in radio.c
 #endif
 
+#define RADIO_CONFIG_PA_USES_DCDC 0           /// The PA(s) is(are) fed from VBAT
+
 #endif // __BOARD_CONFIG_H__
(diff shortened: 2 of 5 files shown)


examples/platforms/efr32mg12/radio.c
```diff
@@ -186,7 +185,28 @@ static const RAIL_IEEE802154_Config_t sRailIeee802154Config = {
...
+
+#if RADIO_CONFIG_SUBGIG_PA_USES_DCDC
+RAIL_DECLARE_PA_SG_DCDC;
+static const RAIL_TxPowerCurves_t curvesSg[1] = {
+    {RAIL_PA_CURVES_SG_DCDC_MAX_POWER, RAIL_PA_CURVES_SG_DCDC_MIN_POWER, subgigDcdcCurves}};
+#else
+RAIL_DECLARE_PA_SG_VBAT;
+static const RAIL_TxPowerCurves_t curvesSg[1] = {
+    {RAIL_PA_CURVES_SG_VBAT_MAX_POWER, RAIL_PA_CURVES_SG_VBAT_MIN_POWER, subgigVbatCurves}};
+#endif
```
dismirlian: It would be ideal if these declarations could be provided by the RAIL header. In fact, the header provides `RAIL_DECLARE_TX_POWER_VBAT_CURVES` and `RAIL_DECLARE_TX_POWER_DCDC_CURVES`, but those macros assume that both PAs are from by the same power rail.

andrasbiro-silabs:
Though this is technically correct, I don't see the point this way. All of the silabs kits use the same PA voltage configuration for subGig and 2.4. For custom hw, we recommend to use RAIL_InitTxPowerCurves() directly with the curves you measured for your hw (see AN1127). 

Granted, the design of the current radio.c doesn't allow this, as it calls RAIL_DECLARE_TX_POWER_VBAT_CURVES, which essentially locks you in with the macro magic that is only actually required if you need to support the various silabs kits from simplicity studio.

So I could accept this as is, but I think it would be probably better to move RAIL_TxPowerConfig_t or its elements out from radio.c into the board folder.

dismirlian:
The idea is to allow for custom boards to use radio.c without having to modify it.

> So I could accept this as is, but I think it would be probably better to move RAIL_TxPowerConfig_t or its elements out from radio.c into the board folder.

I like this idea. I'll try to implement this and update the PR.

dismirlian:
While trying to implement the idea proposed by @andrasbiro-silabs (move RAIL_TxPowerConfig_t or its elements out from radio.c into the board folder) I came across a bug in nlbuild-autotools (it won't expand variables in _SOURCES), so the changes required would not justify the effort.

I ended up only fixing the bug (BRD4166 PA supply was wrong), and leaving out the "features" (per-band configuration of PA supply, ramp-time and voltage).

PR #4510, opened 2020-01-24
[posix-app] update config header file includes

abtink:
This PR contains a group of (related) commits for posix-app.

**[posix-app] add missing config header file includes**    
 This commit updates how posix-app source files include the config header file to follow a similar policy as for core modules. All header files and `cpp` files that do not have a corresponding `hpp` file should include "platform-config.h" first (which itself will include the core config header). The "cpp" files which have a related "hpp" should include the related hpp file first. This way we ensure all config macro definitions are correctly included in all modules.

**[posix-app-config] add UART/SPI config macros and their default value**    
This addresses undefined macro warnings when building posix-app.

**[toranj] move posix-app config defs into header file from build script**

**[hdlc] add missing <string.h> include (for memmove())**

**[posix-app] add error if no RCP interface (SPI/UART) is enabled**

### DIFF
--- a/src/posix/client.c
+++ b/src/posix/client.c
@@ -26,7 +26,7 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include "openthread-core-config.h"
+#include "platform/openthread-posix-config.h"
 
 #include <openthread/platform/toolchain.h>
 
--- a/src/posix/main.c
+++ b/src/posix/main.c
@@ -26,6 +26,8 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "platform/openthread-posix-config.h"
+
 #include <openthread/config.h>
 
 #include <assert.h>
--- a/tests/toranj/build.sh
+++ b/tests/toranj/build.sh
@@ -137,7 +137,7 @@ case ${build_config} in
         ./bootstrap || die
         cd "${top_builddir}"
         ${top_srcdir}/configure                 \
-            CPPFLAGS="$cppflags_config -DOPENTHREAD_CONFIG_POSIX_APP_ENABLE_PTY_DEVICE=1 -DOPENTHREAD_POSIX_RCP_UART_ENABLE=1" \
+            CPPFLAGS="$cppflags_config"         \
             --enable-posix-app                  \
             $configure_options || die
         make -j 8 || die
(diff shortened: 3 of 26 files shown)


src/posix/platform/openthread-system.h
```diff
@@ -35,6 +35,8 @@
 #ifndef OPENTHREAD_SYSTEM_H_
 #define OPENTHREAD_SYSTEM_H_
 
+#include "platform-config.h"
```
bukepo: Is this necessary? This file is exported as a public header file.

src/posix/main.c
```diff
@@ -26,6 +26,8 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "platform/platform-config.h"
```
bukepo: Only openthread-system.h is exported.

src/posix/platform/openthread-system.h
abtink: Good point. Did not notcie that we expose this as public header for the posix-app plat lib. We currently don't need/use any config option in this file. I think for future we can follow the same model as OT public headers (i.e. do not use any `#if SOME_CONFIG` checks in public headers).
I will change the commit such that "platform-config.h" is only included form private source files under posix-app platform.

src/posix/main.c
abtink: I kept this include in new push since this file is not exported (is private) - and having all config definitions would be useful.

src/posix/client.c
```diff
@@ -26,7 +26,7 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include "openthread-core-config.h"
+#include "platform/platform-config.h"
```
jwhui: Should we be renaming this include to something like `openthread-posix-config.h`?

tests/toranj/build.sh
```diff
@@ -137,7 +137,7 @@ case ${build_config} in
         ./bootstrap || die
         cd "${top_builddir}"
         ${top_srcdir}/configure                 \
-            CPPFLAGS="$cppflags_config -DOPENTHREAD_CONFIG_POSIX_APP_ENABLE_PTY_DEVICE=1 -DOPENTHREAD_POSIX_RCP_UART_ENABLE=1" \
+            CPPFLAGS="$cppflags_config" \
```
jwhui: nit: align `\`

src/posix/client.c
abtink: Good idea. 
Renamed it to `openthread-posix-config.h` (also updated the first commit message) in the new push.

PR #4511, opened 2020-01-28
[build] Support building platform code without examples and adapt generic sdk_config.h

e-rk:
- [build] Added --with-platform configure option to enable building the platform libraries
without building the example applications or any supporting libraries used solely by the
examples.

- [nrf528xx] Modified the nrf528xx-specific makefiles such that Nordic SDK code located in
third_party/NordicSemiconductor is only built/referenced when building the example
applications.  When building just the platform libraries, the caller of configure is
expected to provide the correct arguments to use an external copy of the Nordic nRF5 SDK.
An exception to this is the 802.15.4 radio driver, which is always built using the sources
in third_party/NordicSemiconductor/drivers/radio.

- [nrf528xx] Made various minor edits to #include statements in the nrf528xx platform code.
This was necessary because of the difference in directory structure between the code in
third_party/NordicSemiconductor and that in the formally published Nordic nRF5 SDK.

- [nrf528xx] adapt generic sdk_config.h file from nRF5 SDK
Platform code configuration definitions can now be supplied in "app_config.h" file.


e-rk:
This PR supersedes https://github.com/openthread/openthread/pull/4497 and resolves conflict with 609665d017.

configure.ac
```diff
@@ -861,10 +868,88 @@ AM_CONDITIONAL([OPENTHREAD_EXAMPLES_SAMR21],    [test "${with_examples}" = "samr
 
 AM_COND_IF([OPENTHREAD_EXAMPLES_POSIX], CPPFLAGS="${CPPFLAGS} -DOPENTHREAD_EXAMPLES_POSIX=1", CPPFLAGS="${CPPFLAGS} -DOPENTHREAD_EXAMPLES_POSIX=0")
 
-AC_MSG_CHECKING([whether to enable examples])
-AC_MSG_RESULT(${with_examples})
-
 AM_CONDITIONAL([OPENTHREAD_EXAMPLES_NRF528XX], [test OPENTHREAD_EXAMPLES_NRF52811 || test OPENTHREAD_EXAMPLES_NRF52833 || test OPENTHREAD_EXAMPLES_NRF52840])
```
e-rk: It seems that this line does not work as one would expect. In `if` statements in `Makefile.am` the `OPENTHREAD_EXAMPLES_NRF528XX` always evaluates to true.

bukepo:
Currently we have disable-excutables to disable building example apps. Can it solve the part of the problem? If so, it might be possible to keep changes within nordic platform itself. Thoughts?

e-rk:
@bukepo I think it could be done, however that would change the behaviour of the `--enable-executable` flag, because it only disables the examples.

Perhaps the `--with-platform` flags for each platform could be consolidated into one, for example `--with-external-dependencies` and the `--enable-executable` would be set to `no` to achieve the same effect.

jaylogue:
--with-platform is intended to specify which platform layer code to build.  Used by itself, it builds the target's platform layer code only, without building any target-specific support code present in third_party.  This is intended as a general purpose feature for all targets, for when developers are building applications using different external components and different configurations than those chosen for the OpenThread examples.

--enable-executable merely controls whether to build executables of any kind.  In my opinion it would obscure its purpose to have it also control whether third_party code is built.

I don't see how --with-external-dependencies (with no argument) would work.  By itself there would be no indication of which platform layer code to build.  --with-external-dependencies=target could work, but its meaning is super unclear; e.g. if you specify --with-external-dependencies=nrf52840 by itself, what gets built?  platform code?  the example apps?  third_party libraries?  The name of the option gives no clue.

bukepo:
Is it necessary to let both --with-example and --with-platform accept a platform name? How about:
1. Rename --with-example to --with-platform
2. Remove --enable-excutable, add --enable-example-app, --enable-tools

jaylogue:
> Rename --with-example to --with-platform
> Remove --enable-excutable, add --enable-example-app, --enable-tools

This will break all down-stream build systems, which seems highly undesirable.

configure.ac
jaylogue: This line should read as follows:

```
AM_CONDITIONAL([OPENTHREAD_EXAMPLES_NRF528XX], [test "${with_examples}" = "nrf52811" -o "${with_examples}" = "nrf52833" -o "${with_examples}" = "nrf52840"])
```

Issue #4213, opened 2019-09-30
[coap/-s block-wise transfer] Implementation of CoAP/-s Block-Wise Transfer (RFC7959))


AuSchell:
Hello everyone,

I'm currently implementing CoAP Block-Wise-Transfer for my fork of the OpenThread stack for a project and i would like to share my code with this community afterwards, as proposed in #3323. My goal is that the Block-Wise Transfer is handled at the CoAP layer similar to IPv6 fragmentation, should the user define that automatic Block-Wise Transfer should be used. 

So, my current challenge is to cache a CoAP message for the duration of a Block-Wise Transfer, for either disassembling or reassambling. For this problem, i thought of two solutions:

1. An array buffer
- At the sender: The sender which attempts to send a CoAP message with a long payload to a receiver caches the payload of the message (called body) in an array with a defined length in the instance of the CoapBase class used. The sender reads the payload of the blocks sent to the receiver from that array.

- At the receiver: The receiver caches each payload of the received blocks in the same array (or there are different arrays for disassembling and reassembling) in its instance of the CoapBase class. The array is appended to an otMessage at the receiver once the Block-Wise Transfer is completed.

2. The message is cached in the CoAP message pool
- At the sender: The original message is cached in the same message pool from which all CoAP messages are taken. The pointer to the original message is part of the CoapBase class. Each block message is derived from this original message. The original message is freed after the Block-Wise Transfer is completed.

- At the receiver: The receiver also aquires a new message from the coap message pool and copies the header of the first block message it receives (leaving out the block-wise option) and appends the payload of each received block message to that new message. The reassembled message is returned to the application layer after the Block-Wise Transfer is finished.

Which of those sollutions do you think are better or are there better solutions to cache the message for the duration of the Block-Wise Transfer? 

Thank you for your answers.

AuSchell:
PR is released: #4247

PR #4513, opened 2020-01-28
[coap] Add helper functions to encode coap block options

nealjack:
This small feature addition is two helper functions to encode the Block1 and Block2 uint options described in RFC7959. This should help the effort in #4213 and #4247.

### DIFF
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -143,6 +143,18 @@ typedef enum otCoapOptionType
     OT_COAP_OPTION_SIZE1          = 60, ///< Size1
 } otCoapOptionType;
 
+/**
+ * CoAP Block1/Block2 Types
+ */
+typedef enum otCoapBlockType
+{
+    OT_COAP_BLOCK_TYPE_1 = 1,
+    OT_COAP_BLOCK_TYPE_2 = 2,
+} otCoapBlockType;
+
+/**
+ * CoAP Block Size Exponents
+ */
 typedef enum otCoapBlockSize
 {
     OT_COAP_BLOCK_SIZE_16   = 0,
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -144,15 +144,6 @@ typedef enum otCoapOptionType
 } otCoapOptionType;
 
 /**
- * CoAP Block1/Block2 Types
- */
-typedef enum otCoapBlockType
-{
-    OT_COAP_BLOCK_TYPE_1 = 1,
-    OT_COAP_BLOCK_TYPE_2 = 2,
-} otCoapBlockType;
-
-/**
  * CoAP Block Size Exponents
  */
 typedef enum otCoapBlockSize
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -496,12 +496,22 @@ otError otCoapMessageAppendObserveOption(otMessage *aMessage, uint32_t aObserve)
 otError otCoapMessageAppendUriPathOptions(otMessage *aMessage, const char *aUriPath);
 
 /**
+ * This function converts a CoAP Block option SZX field to the actual block size
+ *
+ * @param[in]     aSize     Block size exponent.
+ *
+ * @returns The actual size exponent value.
+ *
+ */
+uint16_t otCoapBlockSizeFromExponent(otCoapBlockSize aSize);
+
+/**
  * This function appends a Block2 option
  *
  * @param[inout]  aMessage  A pointer to the CoAP message.
  * @param[in]     aNum      Current block number.
  * @param[in]     aMore     Boolean to indicate more blocks are to be sent.
- * @param[in]     aSize     Maximum block size.
+ * @param[in]     aSize     Block Size Exponent.
  *
  * @retval OT_ERROR_NONE          Successfully appended the option.
  * @retval OT_ERROR_INVALID_ARGS  The option type is not equal or greater than the last option type.
@@ -516,7 +526,7 @@ otError otCoapMessageAppendBlock2Option(otMessage *aMessage, uint32_t aNum, bool
  * @param[inout]  aMessage  A pointer to the CoAP message.
  * @param[in]     aNum      Current block number.
  * @param[in]     aMore     Boolean to indicate more blocks are to be sent.
- * @param[in]     aSize     Maximum block size.
+ * @param[in]     aSize     Block Size Exponent.
  *
  * @retval OT_ERROR_NONE          Successfully appended the option.
  * @retval OT_ERROR_INVALID_ARGS  The option type is not equal or greater than the last option type.
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -134,6 +134,8 @@ typedef enum otCoapOptionType
     OT_COAP_OPTION_URI_QUERY      = 15, ///< Uri-Query
     OT_COAP_OPTION_ACCEPT         = 17, ///< Accept
     OT_COAP_OPTION_LOCATION_QUERY = 20, ///< Location-Query
+    OT_COAP_OPTION_BLOCK2         = 23, ///< Block1 (RFC7959)
+    OT_COAP_OPTION_BLOCK1         = 27, ///< Block2 (RFC7959)
     OT_COAP_OPTION_PROXY_URI      = 35, ///< Proxy-Uri
     OT_COAP_OPTION_PROXY_SCHEME   = 39, ///< Proxy-Scheme
     OT_COAP_OPTION_SIZE1          = 60, ///< Size1
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -141,6 +141,17 @@ typedef enum otCoapOptionType
     OT_COAP_OPTION_SIZE1          = 60, ///< Size1
 } otCoapOptionType;
 
+typedef enum otCoapBlockSize
+{
+    OT_COAP_BLOCK_SIZE_16   = 0,
+    OT_COAP_BLOCK_SIZE_32   = 1,
+    OT_COAP_BLOCK_SIZE_64   = 2,
+    OT_COAP_BLOCK_SIZE_128  = 3,
+    OT_COAP_BLOCK_SIZE_256  = 4,
+    OT_COAP_BLOCK_SIZE_512  = 5,
+    OT_COAP_BLOCK_SIZE_1024 = 6,
+} otCoapBlockSize;
+
 /**
  * This structure represents a CoAP option.
  *
@@ -483,7 +494,7 @@ otError otCoapMessageAppendUriPathOptions(otMessage *aMessage, const char *aUriP
  * @retval OT_ERROR_NO_BUFS       The option length exceeds the buffer size.
  *
  */
-otError otCoapMessageAppendBlock2Option(otMessage *aMessage, uint32_t aNum, bool aMore, uint16_t aSize);
+otError otCoapMessageAppendBlock2Option(otMessage *aMessage, uint32_t aNum, bool aMore, otCoapBlockSize aSize);
 
 /**
  * This function appends a Block1 option
@@ -498,7 +509,7 @@ otError otCoapMessageAppendBlock2Option(otMessage *aMessage, uint32_t aNum, bool
  * @retval OT_ERROR_NO_BUFS       The option length exceeds the buffer size.
  *
  */
-otError otCoapMessageAppendBlock1Option(otMessage *aMessage, uint32_t aNum, bool aMore, uint16_t aSize);
+otError otCoapMessageAppendBlock1Option(otMessage *aMessage, uint32_t aNum, bool aMore, otCoapBlockSize aSize);
 
 /**
  * This function appends a Proxy-Uri option.
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -471,6 +471,36 @@ otError otCoapMessageAppendObserveOption(otMessage *aMessage, uint32_t aObserve)
 otError otCoapMessageAppendUriPathOptions(otMessage *aMessage, const char *aUriPath);
 
 /**
+ * This function appends a Block2 option
+ *
+ * @param[inout]  aMessage  A pointer to the CoAP message.
+ * @param[in]     aNum      Current block number.
+ * @param[in]     aMore     Boolean to indicate more blocks are to be sent.
+ * @param[in]     aSize     Maximum block size.
+ *
+ * @retval OT_ERROR_NONE          Successfully appended the option.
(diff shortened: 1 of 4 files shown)


src/core/coap/coap_message.cpp
```diff
@@ -187,6 +187,74 @@ otError Message::AppendUriPathOptions(const char *aUriPath)
...
+        SuccessOrExit(error = AppendUintOption(OT_COAP_OPTION_BLOCK1, encoded));
+    }
+    else if (aType == 2)
+    {
+        SuccessOrExit(error = AppendUintOption(OT_COAP_OPTION_BLOCK2, encoded));
+    }
+    else
+    {
+        SuccessOrExit(error = OT_ERROR_INVALID_ARGS);
+    }
```
abtink: Suggest to replace entire `if/else` block with:
```
VerifyOrExit(aType == 1 || aType == 2, error = OT_ERROR_INVALID_ARGS);

error = AppendUintOption((aType == 1) ? OT_COAP_OPTION_BLOCK1 : OT_COAP_OPTION_BLOCK2, encoded);
```

include/openthread/coap.h
```diff
@@ -134,6 +134,8 @@ typedef enum otCoapOptionType
     OT_COAP_OPTION_URI_QUERY      = 15, ///< Uri-Query
     OT_COAP_OPTION_ACCEPT         = 17, ///< Accept
     OT_COAP_OPTION_LOCATION_QUERY = 20, ///< Location-Query
+    OT_COAP_OPTION_BLOCK2         = 23, ///< Block1 (RFC7959)
+    OT_COAP_OPTION_BLOCK1         = 27, ///< Block2 (RFC7959)
```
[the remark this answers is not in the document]
nealjack: You're correct! I crossed the references here and in the function definitions

src/core/coap/coap_message.cpp
```diff
@@ -187,6 +187,74 @@ otError Message::AppendUriPathOptions(const char *aUriPath)
...
+                                   const bool     aMore,
+                                   const uint16_t aSizeExponent)
+{
+    otError  error   = OT_ERROR_NONE;
+    uint32_t encoded = 0;
+
+    if ((aNum > 0xFFFFF))
+    {
+        SuccessOrExit(error = OT_ERROR_INVALID_ARGS);
+    }
```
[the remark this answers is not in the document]
nealjack: Done! I wasn't aware of these utilities.

src/core/coap/coap_message.cpp
```diff
@@ -187,6 +187,74 @@ otError Message::AppendUriPathOptions(const char *aUriPath)
...
+
+    case 512:
+        encoded = 5;
+        break;
+
+    case 1024:
+        encoded = 6;
+        break;
+
+    case 2048:
```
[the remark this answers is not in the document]
nealjack: Removed

include/openthread/coap.h
```diff
@@ -468,6 +470,36 @@ otError otCoapMessageAppendObserveOption(otMessage *aMessage, uint32_t aObserve)
...
+ * @param[in]     aNum      Current block number.
+ * @param[in]     aMore     Boolean to indicate more blocks are to be sent.
+ * @param[in]     aSize     Maximum block size.
+ *
+ * @retval OT_ERROR_NONE          Successfully appended the option.
+ * @retval OT_ERROR_INVALID_ARGS  The option type is not equal or greater than the last option type.
+ * @retval OT_ERROR_NO_BUFS       The option length exceeds the buffer size.
+ *
+ */
+otError otCoapMessageAppendBlock2Option(otMessage *aMessage, uint32_t aNum, bool aMore, uint16_t aSize);
```
abtink: The implementation expects `aSize` to be certain values 16, 32, 64, ..., 1024. All other values would cause a `OT_ERROR_INVALID_ARGS` error.

How about instead of using a `uint16_t aSize`, we define an `enum otCoapBlockSize` type:
```
typedef enum otCoapBlockSize
{
    OT_COAP_BLOCK_SIZE_16 = 0,
    OT_COAP_BLOCK_SIZE_32 = 1,
    ...
} otCoapBlockSize;
```
I feel this way it is more clear how the API is intended to be used.  Also the enum values  already map to encoded value which would also simplify the encoding logic in `AppendBlockOption()`

Thoughts?

nealjack: Yeah, I like that solution much better than passing in an arbitrary int. I'll make the change!

nealjack: I just thought of a few downsides to using an enum. Most applications will still need to have an additional integer value of the block size around in order to perform data segmentation and block number calculations, which is kind of annoying. Also, with a C compiler, a user won't get a warning/error when supplying an invalid integer instead of an enum member.

I still think your proposed solution is more elegant, however.

jwhui:
@sjlongland - given that you've been driving a bunch of CoAP enhancements, can you review/comment on this PR?

include/openthread/coap.h
sjlongland: The other way would be for the application to specify the size exponent directly, and instead provide a means (macro, inline function) to convert between that and and nearest size… something like:

```
#define OT_COAP_MIN_SZX (0)
#define OT_COAP_MAX_SZX (7)

uint16_t otCoapBlockSizeFromExponent(uint8_t aSzx)
{
    return 1 << (aSzx + 4);
}

uint8_t otCoapNearestSizeExponent(uint16_t aBlockSize)
{
    uint8_t szx;
    for (szx = OT_COAP_MIN_SZX; szx < OT_COAP_MAX_SZX; szx++)
    {
        if (otCoapBlockSizeFromExponent(szx) >= aBlockSize)
        {
            break;
        }
    }

    return szx;
}
```

That will probably need some tweaking, and there may be optimisations that can be made (I took the naïve iterative approach to illustrate), but that will handle arbitrary sizes like 192 bytes, by rounding up to the nearest block size (256 in that case).

Then application writers can bounce either way.  The C compiler may even be smart enough if given a constant to `otCoapNearestSizeExponent`, that the result never changes and so replace the function call with a constant.  The return value of `otCoapNearestSizeExponent` could possibly be cast to that `enum` as well.

src/core/coap/coap_message.cpp
```diff
@@ -187,6 +187,23 @@ otError Message::AppendUriPathOptions(const char *aUriPath)
...
+{
+    otError  error   = OT_ERROR_NONE;
+    uint32_t encoded = aSize;
+
+    VerifyOrExit(aNum < 0xFFFFF, error = OT_ERROR_INVALID_ARGS);
+
+    encoded |= static_cast<uint32_t>((aMore) << 3);
+    encoded |= (aNum) << 4;
+
+    VerifyOrExit(aType == 1 || aType == 2, error = OT_ERROR_INVALID_ARGS);
```
sjlongland: I wonder if an `enum` might be a clearer way of ensuring this… maybe in `coap_message.hpp` (since this isn't exposed to the C API):

```
typedef enum otCoapBlockType {
    OT_COAP_BLOCK_TYPE_1 = 1,
    OT_COAP_BLOCK_TYPE_2 = 2
} otCoapBlockType;
```

Then the compiler will enforce type 1 or type 2 for you.

include/openthread/coap.h
nealjack: Yeah, I was thinking at the very least that providing a helper function to convert to/from an encoded size exponent would be helpful. I think a user providing a non-existent size should result in an error. I don't think we should increase/decrease size to a valid number. The interface can and should be well defined and limited - a user should supply a valid size.

sjlongland:
Nothing really jumped out at me, it looks good.  There are a few small improvements that can be made, but even as-is, it looks great. :-)

src/core/coap/coap_message.cpp
nealjack: Fair, that's a good change. Although only a C++ compiler will enforce that. A C compiler doesn't differentiate an integer or an enum. In practice a user could provide any integer and not get an error.

include/openthread/coap.h
sjlongland: Fair enough, thankfully there aren't lots of options (8 possible values, the powers of two between 16 and 2048… we might omit 2048 since the MTU won't support it).

The complicating matter is you then can't do something like…
```
otCoapMessageAppendBlock2Option(mymessage, mynum, mynum < mylast, otCoapGimmieMySzx(1024));
```
…which would work as expected otherwise, unless we added a `OT_COAP_BLOCK_SIZE_INVALID` entry to the size `enum` (which has its own problems).

I'd say the case where the application chooses the `SZX` directly and needs to know the block size is more probable though, and for that, `otCoapBlockSizeFromExponent` alone would suffice.

So one approach is we provide a convenience function to go one way, and let the application developer beat a path back if they want to do it the other way.

src/core/coap/coap_message.hpp
```diff
@@ -97,6 +97,16 @@ class Message : public ot::Message
      */
     typedef otCoapCode Code;
 
+    /**
+     * CoAP Block1/Block2 Types
+     *
+     */
+    typedef enum otCoapBlockType
```
abtink: @nealjack thanks for all the changes. Some small naming suggestions below:

In OT, core is c++, but all public APIs as defined in C.

Since this definition is restricted to core (c++ source), we don't need the `typedef` anymore (`enum Type` already defines a type named `Type`).

Also the `ot{Name}` convention is typically used for public C OpenThread definitions. Within core modules, we don't use the `ot` prefix (since everything is already under `namespace ot`).

For constants, the `OT_{NAME}` is typically used for public C OT definitions. Within core (c++) the convention is `k{NameInCamelCase`}`.


So suggest the following:
```
enum BlockType  
{
    kBlockType1 = 1,
    kBlockType2 = 2,
};
```
(this `enum` type is already under `class Coap` so it would get namespaced as `Coap::BlockType`).

nealjack: Sounds good - I moved the enum definition without thinking about naming

src/core/coap/coap_message.cpp
```diff
@@ -197,7 +197,7 @@ otError Message::AppendBlockOption(Message::otCoapBlockType aType, uint32_t aNum
     encoded |= static_cast<uint32_t>((aMore) << 3);
     encoded |= (aNum) << 4;
 
-    VerifyOrExit(aType == OT_COAP_BLOCK_TYPE_1 || aType == OT_COAP_BLOCK_TYPE_2, error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(aType == kBlockType1 || aType == kBlockType2, error = OT_ERROR_INVALID_ARGS);
     error = AppendUintOption((aType == 1) ? OT_COAP_OPTION_BLOCK1 : OT_COAP_OPTION_BLOCK2, encoded);
```
abtink:
```suggestion
    error = AppendUintOption((aType == kBlockType1) ? OT_COAP_OPTION_BLOCK1 : OT_COAP_OPTION_BLOCK2, encoded);
```

src/core/api/coap_api.cpp
```diff
@@ -110,6 +110,21 @@ otError otCoapMessageAppendUriPathOptions(otMessage *aMessage, const char *aUriP
     return static_cast<Coap::Message *>(aMessage)->AppendUriPathOptions(aUriPath);
 }
 
+uint16_t otCoapBlockSizeFromExponent(otCoapBlockSize aSize)
+{
+    return static_cast<uint16_t>(1 << (aSize + 4));
```
LuDuda: nit: magic number

src/core/coap/coap_message.cpp
```diff
@@ -187,6 +187,23 @@ otError Message::AppendUriPathOptions(const char *aUriPath)
...
 }
 
+otError Message::AppendBlockOption(Message::BlockType aType, uint32_t aNum, bool aMore, otCoapBlockSize aSize)
+{
+    otError  error   = OT_ERROR_NONE;
+    uint32_t encoded = aSize;
+
+    VerifyOrExit(aNum < 0xFFFFF, error = OT_ERROR_INVALID_ARGS);
+
+    encoded |= static_cast<uint32_t>((aMore) << 3);
```
LuDuda: nit: magic number here and below

src/core/coap/coap_message.cpp
```diff
@@ -187,6 +187,23 @@ otError Message::AppendUriPathOptions(const char *aUriPath)
...
+    encoded |= (aNum) << 4;
+
+    VerifyOrExit(aType == kBlockType1 || aType == kBlockType2, error = OT_ERROR_INVALID_ARGS);
```
LuDuda: nit: move check to line 196, before processing anything?

src/core/coap/coap_message.cpp
```diff
@@ -187,6 +187,23 @@ otError Message::AppendUriPathOptions(const char *aUriPath)
     return error;
 }
 
+otError Message::AppendBlockOption(Message::BlockType aType, uint32_t aNum, bool aMore, otCoapBlockSize aSize)
+{
+    otError  error   = OT_ERROR_NONE;
+    uint32_t encoded = aSize;
+
+    VerifyOrExit(aNum < 0xFFFFF, error = OT_ERROR_INVALID_ARGS);
```
LuDuda: nit: magic number

src/core/coap/coap_message.cpp
```diff
@@ -187,6 +187,23 @@ otError Message::AppendUriPathOptions(const char *aUriPath)
...
+
+    encoded |= static_cast<uint32_t>((aMore) << 3);
+    encoded |= (aNum) << 4;
```
LuDuda: nit: unnecessary bracket

src/core/coap/coap_message.cpp
LuDuda: nit: unnecessary bracket around `aMore`

src/core/api/coap_api.cpp
nealjack: Where is the appropriate place to define these constants? Are defines or enums preferable? Are there any existing examples that you can point me to?

jwhui: enums are generally preferred over defines for scoping purposes. You can find an example here: https://github.com/openthread/openthread/blob/a84fea58e10beaa0d8cb775915c1ec57e221c0a1/src/core/thread/mle_tlvs.hpp#L1740-L1745

src/core/coap/coap_message.cpp
nealjack: This file actually contains many definitions of similar masks. Should these eventually be changed too, or are masks generally acceptable?

https://github.com/openthread/openthread/blob/3954b549dba3fe2ad0a01b8fd8df4f41124825b7/src/core/coap/coap_message.cpp#L105
https://github.com/openthread/openthread/blob/3954b549dba3fe2ad0a01b8fd8df4f41124825b7/src/core/coap/coap_message.cpp#L169

PR #4518, opened 2020-01-30
[posix-host] add hardware flow control capability to hdlc interface

dismirlian:


src/posix/platform/hdlc_interface.cpp
```diff
@@ -549,6 +550,20 @@ int HdlcInterface::OpenFile(const char *aFile, const char *aConfig)
...
+        switch (flow)
+        {
+        case 'N':
+            break;
+        case 'H':
+            tios.c_cflag |= CRTSCTS;
+            break;
+        default:
+            // not supported
+            assert(false);
```
bukepo: Remove this for this is not logic error?

dismirlian: Hi, I duplicated the logic that was implemented for parity, speed and stop bits:

https://github.com/openthread/openthread/blob/cd0ef26c0257e82c263d0649366311dd46de4c95/src/posix/platform/hdlc_interface.cpp#L440-L448

https://github.com/openthread/openthread/blob/cd0ef26c0257e82c263d0649366311dd46de4c95/src/posix/platform/hdlc_interface.cpp#L455-L462

https://github.com/openthread/openthread/blob/cd0ef26c0257e82c263d0649366311dd46de4c95/src/posix/platform/hdlc_interface.cpp#L542-L550

Do you think I should remove the `assert`s from all?

bukepo: Yes, I think so.  Thank you for finding these mistakes.

dismirlian: Done.

PR #4519, opened 2020-01-31
[ip6-address] multiple enhancements

abtink:
This PR contains a group of smaller commits all in `Ip6::Address`.

**[ip6-address] use larger data fields when checking against constants**

This commit changes methods in `Ip6::Address` class to use larger
fields (`m16` or `m32`) for comparing portions of address sequence
against constant values when possible.

**[ip6-address] simplify IsAnycastServiceLocator()**

This change ensures we avoid comparing values with host-swapped
constants.

**[ip6-address] remove undefined/unused method declaration**

**[ip6-address] inline very simple methods**

This commit inlines the very simple methods (containing a single
basic operation) in `Ip6::Address` class.

------------

Adding a note regarding the second commit: The original code for `IsAnycastServiceLocator()` contained:

```
   (mFields.m16[7] >= HostSwap16(Mle::kAloc16ServiceStart)) &&
   (mFields.m16[7] <= HostSwap16(Mle::kAloc16ServiceEnd)
```

with `kAloc16ServiceStart = 0xfc10` and `kAloc16ServiceEnd = 0xfc2f` and the intention to allow values within the start to end range.

Comparing a `uint16_t` with a `BigEndian::HostSwap16()` constant does not necessarily cover the same range of values. For example, if the platform is little-endian, and the bye sequence is `0x00`, `0x11`, the `uint16_t` value would be read as `0x1100` which would then be within the range of host-swapped constants, i.e. `0x10fc <= 0x1100 <= 0x2ffc` but it should not be accepted.

The behavior of the method in this case was actually fine since the `IsAnycastRoutingLocator()` already checked the `mFields.m8[14] == kAloc16Mask` (where `kAloc16Mask = 0xfc`), so not marking this as a bug. :)

### DIFF
--- a/src/core/net/ip6_address.cpp
+++ b/src/core/net/ip6_address.cpp
@@ -129,8 +129,8 @@ bool Address::IsAnycastRoutingLocator(void) const
 
 bool Address::IsAnycastServiceLocator(void) const
 {
-    return IsAnycastRoutingLocator() && (mFields.m16[7] >= HostSwap16(Mle::kAloc16ServiceStart)) &&
-           (mFields.m16[7] <= HostSwap16(Mle::kAloc16ServiceEnd));
+    return IsAnycastRoutingLocator() && (mFields.m8[15] >= (Mle::kAloc16ServiceStart & 0xff)) &&
+           (mFields.m8[15] <= (Mle::kAloc16ServiceEnd & 0xff));
 }
 
 bool Address::IsSubnetRouterAnycast(void) const
--- a/src/core/net/ip6_address.cpp
+++ b/src/core/net/ip6_address.cpp
@@ -62,7 +62,7 @@ bool Address::IsLoopback(void) const
 
 bool Address::IsLinkLocal(void) const
 {
-    return (mFields.m8[0] == 0xfe) && ((mFields.m8[1] & 0xc0) == 0x80);
+    return (mFields.m16[0] & HostSwap16(0xffc0)) == HostSwap16(0xfe80);
 }
 
 bool Address::IsMulticast(void) const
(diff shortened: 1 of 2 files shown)


abtink:
Thanks @bukepo.

I think/expect very likely most toolchains would already do `HostSwap32(0xff)` in compile-time since all functions in `encoding.hpp` are defined as inline.

Checking the assembly output for `arm-gcc-7` it seems like it does convert the constants already
(`0x000000ff` -> `0xff000000`):
```
00013a74 <_ZNK2ot3Ip67Address16IsRoutingLocatorEv>:
   13a74:       6883            ldr     r3, [r0, #8]
   13a76:       f1b3 4f7f       cmp.w   r3, #4278190080 ; 0xff000000
   13a7a:       d10a            bne.n   13a92 <_ZNK2ot3Ip67Address16IsRoutingLocatorEv+0x1e>
   13a7c:       8983            ldrh    r3, [r0, #12]
   13a7e:       2bfe            cmp     r3, #254        ; 0xfe
   13a80:       d107            bne.n   13a92 <_ZNK2ot3Ip67Address16IsRoutingLocatorEv+0x1e>
   13a82:       7b80            ldrb    r0, [r0, #14]
   13a84:       28fb            cmp     r0, #251        ; 0xfb
   13a86:       d804            bhi.n   13a92 <_ZNK2ot3Ip67Address16IsRoutingLocatorEv+0x1e>
   13a88:       f080 0002       eor.w   r0, r0, #2
   13a8c:       f3c0 0040       ubfx    r0, r0, #1, #1
   13a90:       4770            bx      lr
   13a92:       2000            movs    r0, #0
   13a94:       4770            bx      lr
```

PR #4522, opened 2020-02-01
[ip6] fix conversion of enum value to uint8_t

abtink:
This commit updates how the `otMessagePriority` enumeration variable
is updated in `otMessageSettings` structure. This addresses an issue
where the enum variable was being treated as an `uint8_t` which may
not necessarily be correct (since complier can choose any integral
type which covers all the enumerator values as the underlying storage
for an unscoped enumeration type).

------

`*reinterpret_cast<uint8_t *>(&settings.mPriority)` should work ok on most platforms/toolchains. I think it would be incorrect only if the platform is big-endian and complier uses a larger integral type (than `uint8_t`) for the enum variable.  A bug but not that critical :)

### DIFF
--- a/src/core/net/ip6.cpp
+++ b/src/core/net/ip6.cpp
@@ -76,13 +76,15 @@ Message *Ip6::NewMessage(const uint8_t *aData, uint16_t aDataLength, const otMes
 {
     otMessageSettings settings = {true, OT_MESSAGE_PRIORITY_NORMAL};
     Message *         message  = NULL;
+    uint8_t           priority;
 
     if (aSettings != NULL)
     {
         settings = *aSettings;
     }
 
-    SuccessOrExit(GetDatagramPriority(aData, aDataLength, *reinterpret_cast<uint8_t *>(&settings.mPriority)));
+    SuccessOrExit(GetDatagramPriority(aData, aDataLength, priority));
+    settings.mPriority = static_cast<otMessagePriority>(priority);
     VerifyOrExit((message = Get<MessagePool>().New(Message::kTypeIp6, 0, &settings)) != NULL);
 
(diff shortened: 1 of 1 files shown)


src/core/net/ip6.cpp
```diff
@@ -76,13 +76,15 @@ Message *Ip6::NewMessage(const uint8_t *aData, uint16_t aDataLength, const otMes
...
     Message *         message  = NULL;
+    uint8_t           priority;
 
     if (aSettings != NULL)
     {
         settings = *aSettings;
     }
 
-    SuccessOrExit(GetDatagramPriority(aData, aDataLength, *reinterpret_cast<uint8_t *>(&settings.mPriority)));
+    SuccessOrExit(GetDatagramPriority(aData, aDataLength, priority));
```
bukepo: Is it possible to change the prototype of GetDatagramPriority() to directly manipulate  otMessagePriority instead of uint8?

abtink: God idea. I think it may work in this case.

 Checking the code it looks like in general we seem to use a `uint8_t` for priority inside core and not the `enum` type.  I guess the main reason for this was that we only expose 3 priority levels (low, normal, high) to outside (so `enum` has 3 values) but inside core we have 4 levels (we use `kPriorityNet` for MLE).

abtink: Actually thinking more about this, I think it'd be better to not use `otMessagePriority` type as the parameter type for the `GetDatagramPriority()` method (which is a private core method). This way, the logic stays consistent and clear: Inside core we always represent priority as `uint8_t` type. The enum type `otMessagePriority` is only for used for public OT APIs. If we mix and match I feel it may cause confusion in future.

Thoughts?

bukepo: SGTM

PR #4527, opened 2020-02-04
[settings] encode as packed little endian structs

### LINKED ISSUE #4461: Provide settings serializer?
Currently settings are persisted as opaque bytes, which is making restoring dataset from NCP not easy. I think this may also cause problem when switching from a 32-bit mode to a 64-bit mode. The alignment may be different.

Implement a serializer and deserializer for settings structures, so that the data is independent of CPU architecture.

Allow platform to provide serializer and deserializer.

**Additional context**
Some products may migrate from NCP to RCP solution, some may switch from 32bit to 64bit.

jwhui:
The choice of little-endian was made for the following reasons:

- The existing encoding remains unchanged for settings unchanged on little-endian platforms.

- Eliminates any code overhead for little-endian platforms.

Resolves #4461

### DIFF
--- a/src/core/common/settings.cpp
+++ b/src/core/common/settings.cpp
@@ -51,26 +51,29 @@ namespace ot {
 void SettingsBase::LogNetworkInfo(const char *aAction, const NetworkInfo &aNetworkInfo) const
 {
     otLogInfoCore("Non-volatile: %s NetworkInfo {rloc:0x%04x, extaddr:%s, role:%s, mode:0x%02x, keyseq:0x%x, ...",
-                  aAction, aNetworkInfo.mRloc16, aNetworkInfo.mExtAddress.ToString().AsCString(),
-                  Mle::Mle::RoleToString(static_cast<otDeviceRole>(aNetworkInfo.mRole)), aNetworkInfo.mDeviceMode,
-                  aNetworkInfo.mKeySequence);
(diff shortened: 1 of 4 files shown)


jwhui:
@bukepo @abtink here is a proposal to resolve #4461, let me know what you think of the approach. if you agree, I will clean this up further.

bukepo:
That’s great! I think the TLVs in operational dataset should also be defined as little endian.

:+1:

LuDuda:
@jwhui just to make it clear, this PR will not break settings representation in NVM for little-enidan platforms like nRF528xx and over the air upgrade which won't touch persistent data still can be easily done?

jwhui:
> That’s great! I think the TLVs in operational dataset should also be defined as little endian.

Operational Datasets are stored using Thread's TLV encoding. For non-volatile storage, we simply treat it as a variable-length binary blob.

jwhui:
> @jwhui just to make it clear, this PR will not break settings representation in NVM for little-endian platforms like nRF528xx and over the air upgrade which won't touch persistent data still can be easily done?

@LuDuda yes, this PR should be transparent to existing little-endian platforms. However, big-endian platforms will not be able to restore `NetworkInfo`, `ParentInfo`, and `ChildInfo`.  To be clear, not being able to restore that information is not fatal - it simply means a device needs to reattach with a new Extended Address and Mesh-Local Address.

Issue #4531, opened 2020-02-06
ot::Cli::Interpreter Cause Fault when ICMP Echo Reply Received


braiden:
**Describe the bug**

If an ICMP echo reply is received after `otCliConsoleInit` is called, but before there's been any user interaction with the CLI, a fault occurs. In my use case such a situation is common. I'm using openthread in zephyr, I start a cli, but rarely interact with it. Someone could craft an ICMP packet to crash my device, and the device crashes whenever I use zephyr's stack to ping.

**To Reproduce**

`otCliConsoleInit` registers an `Interpreter::HandleIcmpReceive` callback in its constructor. This callback tries to do `mServer->OutputFormat()`. `mServer` is null. It only gets a non-null value when `Interpreter::ProcessLine` is called.

Hope that's clear... happy to test.

**Expected behavior**

Don't log ICMP packet when `mServer` is null?

**Console/log output**

**Additional context**

jwhui:
@braiden thanks for raising this issue! We’ll work to propose a fix soon.

jwhui:
Resolved by #4534.

PR #4532, opened 2020-02-06
[cli] ensure mServer is not NULL from ICMP/Diag callbacks

abtink:
----

This should help address Issue https://github.com/openthread/openthread/issues/4531.

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -1994,6 +1994,7 @@ void Interpreter::HandleIcmpReceive(Message &               aMessage,
 {
     uint32_t timestamp = 0;
 
+    VerifyOrExit(mServer != NULL);
     VerifyOrExit(aIcmpHeader.mType == OT_ICMP6_TYPE_ECHO_REPLY);
 
     mServer->OutputFormat("%u bytes from ",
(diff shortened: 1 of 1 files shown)


src/cli/cli.cpp
```diff
@@ -3716,6 +3717,8 @@ void Interpreter::HandleDiagnosticGetResponse(Message &aMessage, const Ip6::Mess
     uint16_t bytesPrinted = 0;
     uint16_t length       = aMessage.GetLength() - aMessage.GetOffset();
 
+    VerifyOrExit(mServer != NULL);
```
jwhui: I don't think this case is possible. This is a CoAP response handler, which requires a CoAP request to be in the past.

src/cli/cli.cpp
```diff
@@ -1994,6 +1994,7 @@ void Interpreter::HandleIcmpReceive(Message &               aMessage,
 {
     uint32_t timestamp = 0;
 
+    VerifyOrExit(mServer != NULL);
```
jwhui: It's probably more appropriate to only handle ICMPv6 Echo Reply messages if an ICMPv6 Echo Request message was sent.

abtink: I see. Make sense (and more correct this way). I went for quick fix for this.
Let me close this PR.

PR #4534, opened 2020-02-06
[cli] use ICMPv6 Echo Request ID to match Responses

abtink:
This commit updates/enhances the CLI `ping` command. It adds the
feature to track and match the ICMPv6 echo requests and responses
using an identifier. Each `ping` command will use a new value as the
identifier. A received ICMPv6 echo response not matching the current
identifier is ignored. This commit also makes the following (smaller)
changes:
- renames the ping related member variables as `mPing{Name}`,
- defines  constants for default ping parameters,
- changes `ping` to use public OT APIs/definition (instead of core
  methods).

---

This should help address https://github.com/openthread/openthread/issues/4531.

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -225,9 +225,12 @@ Interpreter::Interpreter(Instance *aInstance)
     : mUserCommands(NULL)
     , mUserCommandsLength(0)
     , mServer(NULL)
-    , mLength(8)
-    , mCount(1)
-    , mInterval(1000)
+    , mPingLength(kDefaultPingLength)
+    , mPingCount(kDefaultPingCount)
+    , mPingInterval(kDefaultPingInterval)
+    , mPingHopLimit(0)
+    , mPingAllowZeroHopLimit(false)
+    , mPingIdentifier(0)
     , mPingTimer(*aInstance, &Interpreter::HandlePingTimer, this)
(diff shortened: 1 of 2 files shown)


src/cli/cli.cpp
```diff
@@ -2083,42 +2077,56 @@ void Interpreter::ProcessPing(int argc, char *argv[])
         index++;
     }
 
-    HandlePingTimer();
+    mPingIdentifier++;
 
-    return;
+    if (mPingIdentifier == 0)
+    {
+        mPingIdentifier++;
```
simonlingoogle: Should mPingIdentifier increase for each ping caused by the same ping command?

abtink: ICMP echo request/response has its own seq counter `icmp_seq` which is increased per message and is included in output/print message to user. 

The `mPingIndentifier` is tied to every `ping` command issued by user. Note that this is only used internally (the `id` is not printed and known to user).

abtink:
@LuDuda I agree this will change how issuing back-to-back `ping` would behave. CLI will handle a single outstanding `ping`command  at a time (the `ping` command itself can specify a `count` parameter for number of pings to send). But once we issue a new `ping` (or `ping stop`) we won't get any output for any previous `ping` commands even if we receive ICMP echo replies from those later.

I don't have any strong preference here. If there is a need to support parallel `ping`s from CLI we can change the logic to support it.

Thoughts?

PR #4536, opened 2020-02-08
[nrf528xx] provide the option to read factory assigned EUI-64 from UICR

LuDuda:
This PR adds another way of defining the factory-assigned EUI64, by reading it from the UICR register.

### DIFF
--- a/examples/platforms/nrf528xx/nrf52833/openthread-core-nrf52833-config.h
+++ b/examples/platforms/nrf528xx/nrf52833/openthread-core-nrf52833-config.h
@@ -276,6 +276,46 @@
 #define RADIO_CONFIG_SRC_MATCH_EXT_ENTRY_NUM 0
 #endif
 
+/**
+ * @def #define OPENTHREAD_CONFIG_PLATFORM_EUI64_CUSTOM_SOURCE_ENABLE
+ *
+ * Define to 1 to enable custom otPlatRadioGetIeeeEui64 function implementation.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_PLATFORM_EUI64_CUSTOM_SOURCE_ENABLE
+#define OPENTHREAD_CONFIG_PLATFORM_EUI64_CUSTOM_SOURCE_ENABLE 0
+#endif
+
+/**
+ * @def #define OPENTHREAD_CONFIG_PLATFORM_EUI64_UICR_ENABLE
+ *
+ * Define to 1 to enable reading the IEEE EUI-64 address from the UICR.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_PLATFORM_EUI64_UICR_ENABLE
+#define OPENTHREAD_CONFIG_PLATFORM_EUI64_UICR_ENABLE 0
+#endif
+
+/**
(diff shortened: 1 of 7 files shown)


examples/platforms/nrf528xx/nrf52833/openthread-core-nrf52833-config.h
```diff
@@ -276,6 +276,46 @@
...
 #endif
 
+/**
+ * @def #define OPENTHREAD_CONFIG_PLATFORM_EUI64_CUSTOM_SOURCE_ENABLE
+ *
+ * Define to 1 to enable custom otPlatRadioGetIeeeEui64 function implementation.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_PLATFORM_EUI64_CUSTOM_SOURCE_ENABLE
+#define OPENTHREAD_CONFIG_PLATFORM_EUI64_CUSTOM_SOURCE_ENABLE 0
```
bukepo:
```suggestion
#define OPENTHREAD_CONFIG_NRF528XX_EUI64_CUSTOM_SOURCE_ENABLE 0
```

Shall we use a more descriptive prefix for platform specific configurations? @jwhui

jwhui: I could see this specific configuration being used across other platforms - it simply indicates whether to compile in `otPlatRadioGetIeeeEui64()` from `examples/platforms`.

However, we could consider using the `OPENTHREAD_CONFIG_NRF528XX` prefix for the configurations below, since those seem specific to nRF528xx platforms.

@LuDuda, thoughts?

LuDuda: Platform-specific configuration is also in the `platform-config.h` file, so I see the inconsistency. The `OPENTHREAD_CONFIG_PLATFORM_EUI64_CUSTOM_SOURCE_ENABLE` was added by the community and accepted before. I do believe that this config should not be present in `openthread-core-config.h`. Will make proper changes soon. Thanks!

jwhui: I agree that `OPENTHREAD_CONFIG_PLATFORM_EUI64_CUSTOM_SOURCE_ENABLE` can keep the more generic prefix.

However, for the other configurations, like `OPENTHREAD_CONFIG_PLATFORM_EUI64_CUSTOM_SOURCE_ENABLE`, those seem more nRF specific. I think @bukepo's question was more about those. Thoughts?

LuDuda: Oh, what I meant to say but probably failed, is that I agree with the @bukepo comment, and will move all platform dependent configuration to the `platform-config.h` with the changed prefix. Regarding keeping or not `OPENTHREAD_CONFIG_PLATFORM_EUI64_CUSTOM_SOURCE_ENABLE`, i think it can be implemented by each platform, though, it is not strictly related to `OPENTHREAD_CONFIG_PLATFORM` like microseconds timer or some radio features.

jwhui: @LuDuda, sounds good. Thanks for the clarification!

PR #4540, opened 2020-02-09
[address-resolver] resolve several issues

### LINKED ISSUE #4535: [adress-query] Address Error Notification
When I enabled all logs during stress tests on 24 NCP-FTD boards (where devices often create different partitions and merge again due to too heavy traffic), I found a lot of this kind of logs:

`Error while processing address error notification: Drop`

Couple conclusions:

1. This particular error is printed since we don't accept non-confirmable CoAP type [here](https://github.com/openthread/openthread/blob/master/src/core/thread/address_resolver.cpp#L567). This is a bug since according to the specification we should be able to parse Address Notification which is sent as a multicast packet with Non-Confirmable CoAP type.

Proposed solution: Just accept NON and CON.

2. The Address Error of the multicast type is only sent in this [line](https://github.com/openthread/openthread/blob/master/src/core/thread/address_resolver.cpp#L470). This means that in the network there are at least two devices which share the same ML-EID - which is not the case - another bug.

I spent some time analyzing the code to understand how it could be possible to have such a problem since it does not seem obvious. The only theory I have is that the memcmp checks ML-EID (from Address Notification) with some garbage address. For example, I can imagine, that there was a previous cache entry created by a proactive inspection of mesh header [here](https://github.com/openthread/openthread/blob/master/src/core/thread/mesh_forwarder_ftd.cpp#L609), which at the end [created the entry](https://github.com/openthread/openthread/blob/master/src/core/thread/address_resolver.cpp#L269) without any ML-EID modification. This came with #4091.

After the stress test, and some stabilization phase, we often saw that certain devices need a couple of ICMPv6 Ping Req/Rsp iterations before they could communicate with all the other devices in the network. I think that the problem was hidden in the Address Resolver module, which actually might be related to the reported above issue.

LuDuda:
This PR includes four individual commits associated with the address resolver module. It should resolve #4535. It's the easiest to review them one-by-one.

**[address-resolver] accept multicast Address Error Notification**

This commit enables processing of Address Error Notification sent as non-confirmable CoAP multicast message. The code change relaxes the verification of CoAP type.
_____________________________________________________

**[address-resolver] fix last transaction time of entry created by packet inspection**

For cached entries created based on snooping message, the last transaction time was incorrectly set.
_____________________________________________________

**[address-resolver] dequeue a message from resolving queue before invalidate the entry**

During code inspection, I found that in two places, a cache entry may be invalidated in the query state. That may result in a packet permanently being stuck in resolving queue in the Mesh Forwarder layer.

The fast test, to emulate erroneous behavior without this PR:

Topology: Leader <--> Router

```
Leader:
> reset

Router:
> ping Leader_ML-EID
> state leader   (emulate moving to another partition, command needs to be executed within 3 seconds)
Done
> bufferinfo
...
arp: 1 1 (stays forever)
...
```

_____________________________________________________

**[address-resolver] introduce MoveAt functionality**

This commit introduces generic MoveCacheEntryAt functionality. This also helps to address @abtink request to put cache entries which are based on snooping messages, right after cached entries.

_____________________________________________________

I run some manual and stress tests on this PR. However, I think we should have more system/unit tests for address resolver/address error functionalities. Unfortunately, I don't have the bandwidth to add them now.

### DIFF
--- a/src/core/thread/address_resolver.cpp
+++ b/src/core/thread/address_resolver.cpp
@@ -248,8 +248,8 @@ otError AddressResolver::UpdateCacheEntry(const Ip6::Address &aEid, Mac::ShortAd
 
             if (mCache[i].mState != Cache::kStateCached)
             {
-                mCache[i].mRetryTimeout        = 0;
                 mCache[i].mLastTransactionTime = static_cast<uint32_t>(kLastTransactionTimeInvalid);
+                mCache[i].mRetryTimeout        = 0;
                 mCache[i].mTimeout             = 0;
                 mCache[i].mFailures            = 0;
                 mCache[i].mState               = Cache::kStateCached;
@@ -273,11 +273,13 @@ otError AddressResolver::AddCacheEntry(const Ip6::Address &aEid, Mac::ShortAddre
 
     VerifyOrExit(entry != NULL, error = OT_ERROR_NO_BUFS);
 
-    entry->mTarget   = aEid;
-    entry->mRloc16   = aRloc16;
-    entry->mTimeout  = 0;
-    entry->mFailures = 0;
-    entry->mState    = Cache::kStateCached;
+    entry->mTarget              = aEid;
+    entry->mLastTransactionTime = static_cast<uint32_t>(kLastTransactionTimeInvalid);
+    entry->mRloc16              = aRloc16;
+    entry->mRetryTimeout        = 0;
+    entry->mTimeout             = 0;
+    entry->mFailures            = 0;
+    entry->mState               = Cache::kStateCached;
 
     MarkCacheEntryAsUsed(*entry);
 
--- a/src/core/thread/address_resolver.cpp
+++ b/src/core/thread/address_resolver.cpp
@@ -564,8 +564,7 @@ void AddressResolver::HandleAddressError(Coap::Message &aMessage, const Ip6::Mes
     Mac::ExtAddress       macAddr;
     Ip6::Address          destination;
 
-    VerifyOrExit(aMessage.GetType() == OT_COAP_TYPE_CONFIRMABLE && aMessage.GetCode() == OT_COAP_CODE_POST,
-                 error = OT_ERROR_DROP);
+    VerifyOrExit(aMessage.GetCode() == OT_COAP_CODE_POST, error = OT_ERROR_DROP);
 
     otLogInfoArp("Received address error notification");
 
--- a/src/core/thread/address_resolver.cpp
+++ b/src/core/thread/address_resolver.cpp
@@ -345,8 +345,9 @@ otError AddressResolver::AddCacheEntry(const Ip6::Address &aEid, Mac::ShortAddre
     entry->mTimeout             = 0;
     entry->mFailures            = 0;
     entry->mState               = Cache::kStateCached;
-
     MoveCacheEntryBehindCached(*entry);
+
+    otLogNoteArp("Cache entry added (snoop): %s, 0x%04x", aEid.ToString().AsCString(), aRloc16);
 
 exit:
     return error;
--- a/src/core/thread/address_resolver.cpp
+++ b/src/core/thread/address_resolver.cpp
@@ -170,17 +170,71 @@ AddressResolver::Cache *AddressResolver::NewCacheEntry(void)
     return rval;
 }
 
-void AddressResolver::MarkCacheEntryAsUsed(Cache &aEntry)
+void AddressResolver::MoveCacheEntryAt(Cache &aEntry, uint8_t aAge)
 {
-    for (int i = 0; i < kCacheEntries; i++)
+    VerifyOrExit(aEntry.mAge != aAge);
+
+    if (aEntry.mAge > aAge)
     {
-        if (mCache[i].mAge < aEntry.mAge)
+        for (int i = 0; i < kCacheEntries; i++)
         {
-            mCache[i].mAge++;
+            if (mCache[i].mAge >= aAge && mCache[i].mAge < aEntry.mAge)
+            {
+                mCache[i].mAge++;
+            }
+        }
+    }
+    else
+    {
+        for (int i = 0; i < kCacheEntries; i++)
+        {
+            if (mCache[i].mAge > aEntry.mAge && mCache[i].mAge <= aAge)
+            {
+                mCache[i].mAge--;
+            }
         }
     }
 
-    aEntry.mAge = 0;
+    aEntry.mAge = aAge;
+
+exit:
+    return;
+}
+
+void AddressResolver::MoveCacheEntryBehindCached(Cache &aEntry)
+{
+    bool    found             = false;
+    uint8_t lastInCachedState = 0;
+
+    for (int i = 0; i < kCacheEntries; i++)
+    {
+        if (mCache[i].mState != Cache::kStateCached || mCache[i].mAge == aEntry.mAge)
+        {
+            continue;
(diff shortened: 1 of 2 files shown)


src/core/thread/address_resolver.cpp
```diff
@@ -170,17 +170,64 @@ AddressResolver::Cache *AddressResolver::NewCacheEntry(void)
...
+
+        if (mCache[i].mAge > lastInCachedState)
         {
-            mCache[i].mAge++;
+            lastInCachedState = mCache[i].mAge;
         }
     }
 
-    aEntry.mAge = 0;
+    if (i == kCacheEntries)
```
abtink: There is no `break` in the above `for` loop, so when we come out of it, wouldn't `i` always be `kCacheEntries`?

I think we may need to track if we found any entry in cached state or not (with a bool var maybe?) so we can know if `lastInCachedState was set or not?

src/core/thread/address_resolver.cpp
```diff
@@ -248,8 +305,8 @@ otError AddressResolver::UpdateCacheEntry(const Ip6::Address &aEid, Mac::ShortAd
 
             if (mCache[i].mState != Cache::kStateCached)
             {
-                mCache[i].mRetryTimeout        = 0;
                 mCache[i].mLastTransactionTime = static_cast<uint32_t>(kLastTransactionTimeInvalid);
+                mCache[i].mRetryTimeout        = 0;
```
abtink: One suggestion: How about if we call `MoveCacheEntryBehindCached(mChache[i])` from `UpdateCacheEntry()` when we transition the entry from query to cached state?

src/core/thread/address_resolver.cpp
```diff
@@ -79,11 +79,27 @@ void AddressResolver::Clear(void)
...
+void AddressResolver::Clear(void)
+{
+    for (uint8_t i = 0; i < kCacheEntries; i++)
+    {
+        if (mCache[i].mState != Cache::kStateQuery)
+        {
+            continue;
+        }
+
+        Get<MeshForwarder>().HandleResolved(mCache[i].mTarget, OT_ERROR_DROP);
```
abtink: Nice find! 👍

src/core/thread/address_resolver.cpp
```diff
@@ -273,13 +330,15 @@ otError AddressResolver::AddCacheEntry(const Ip6::Address &aEid, Mac::ShortAddre
...
+    entry->mTarget              = aEid;
+    entry->mLastTransactionTime = static_cast<uint32_t>(kLastTransactionTimeInvalid);
+    entry->mRloc16              = aRloc16;
+    entry->mRetryTimeout        = 0;
+    entry->mTimeout             = 0;
+    entry->mFailures            = 0;
+    entry->mState               = Cache::kStateCached;
 
-    MarkCacheEntryAsUsed(*entry);
+    MoveCacheEntryBehindCached(*entry);
```
abtink: Suggestion: Should we add a log here? We have a log everywhere else address cache entries are changed.
```
otLogNoteArp("Cache entry added (rx inspection): %s, 0x%04x", aEid.ToString().AsCString(), aRloc16);
```

src/core/thread/address_resolver.cpp
jwhui: Why wouldn't we call `MoveCacheEntryAtFront()`? If this entry was in `kStateQuery`, I assume that this entry is actually being used.

src/core/thread/address_resolver.cpp
jwhui: In the case where there are no more invalid entries, the "snoop optimization" could have limited value if the device is receiving lots of messages. On the other hand, we wouldn't want the "snoop optimization" to flush entries that are actually in use.

It would add complexity, but one thought is we could have up to X newly-created "snoop optimization" entries. When they are actually used, then we mark them as such. If they are not used after some timeout, then they may be reclaimed. The reason for the timeout is to allow for response delays (especially in the case of responses coming from a sleepy end device).

This can be worked on in a separate PR, but I thought I'd raise the discussion here. Thoughts?

librasungirl: "snoop optimization" always uses the invalid entry or the cached entry of oldest age, this PR here (`MoveCacheEntryBehindCached()`) already ensures that the entry newly added by "snoop optimization" has the oldest age.
For cached entry either added by "snoop optimization" or address query, if used (`Resolve()`) for messages initiated by the device, would have `[fresh](https://github.com/openthread/openthread/blob/master/src/core/thread/address_resolver.cpp#L364)` age.
That is, if the entries added by "snoop optimization" are not used, the last entry added this way would always be preferred to be evicted if required, thus it should have limited impact on the cached entries in use, unless in some extreme scenarios, the device would send messages to the dests (up to `[OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_ENTRIES](https://github.com/openthread/openthread/blob/5d09ce7124be70edc9b635ca9b8670979dbe9835/src/core/config/tmf.h#L45)`) continuously , and all the dests diff from the sources of messages it receives continuously - which I think should be rare to happen.

Besides for performance consideration, it should always be suggested to configure an appropriate value for `[OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_ENTRIES](https://github.com/openthread/openthread/blob/5d09ce7124be70edc9b635ca9b8670979dbe9835/src/core/config/tmf.h#L45)` according to the use case.

Thoughts?

jwhui: In the case where there are no more invalid entries, the "snoop optimization" only gets to use one entry. Having only one entry can be limiting in certain scenarios. For example, if an FTD receives a number of request messages from different sources for its children, only the response to one of those sources will get to utilize the "snoop" entry. The rest will need to incur the overhead of Address Query again. Do you agree?

jwhui: As for sizing the number of cache entries, that is difficult to do in the case that you do not know what applications or networks you will be supporting.

librasungirl: Agree. 
One question for the `X newly-created "snoop optimization" entries`, is it allocated from heap or some fixed ram space?
If latter, with the special scenario mentioned, we'll have another question regarding the size of `the newly-created "snoop optimization" entries` similar as the size issue of cache entries - we never know the proper value without knowing the use cases, and without proper size, the overhead of AQ still could not be avoided.

To relieve the `only one entry` for "snoop optimization" noted above, we may let the entry newly added this way has age like 'SNOOP_CACHE_AGE'  (= kCachesEntries - X), where X is number of entries which could be used for "snoop optimization", when no available invalid entry.

Thoughts?

src/core/thread/address_resolver.cpp
LuDuda: Thanks, heh, before filing this PR I actually changed implementation from your proposed approach to reduce one variable.. apparently, it's not always good to optimize everything :)

src/core/thread/address_resolver.cpp
LuDuda: I agree as well. I think this will be covered with the #4542.

LuDuda:
Thank you for review. I've updated two commits to resolve comments.

src/core/thread/address_resolver.cpp
abtink: Nice discussions an suggestions. I agree and like the idea of allowing some max number of snoop entries and after a new one is added not allowing them to be reclaimed for some timeout interval (giving them a chance to be used).

PR #4544, opened 2020-02-10
[network-data] add check for invalid router IDs

jwhui:
It is posisble for a leader to receive SVR_DATA.ntf messages
attempting to register with a router ID that is not valid on the
leader.

This commit does not allow registering network data with invalid
router IDs to avoid having stale network data that persists.

Addresses the issue raised by @mwilniewiec in https://github.com/openthread/openthread/issues/4472#issuecomment-584217141.


mwilniewiec:
LGTM :)

PR #4547, opened 2020-02-11
[posix-host] use c++ for all platform code

bukepo:
* Switched to C++ for all drivers.
* Removed C utilities.
* Renamed sim.c to virtual_time.cpp.
* Changed platformSim prefix to virtualTime.


src/posix/platform/platform-posix.h
```diff
@@ -305,13 +305,13 @@ void platformNetifProcess(const fd_set *aReadFdSet, const fd_set *aWriteFdSet, c
  * This function initialize virtual time simulation.
  *
  */
-void platformSimInit(void);
+void virtualTimeInit(void);
```
librasungirl: how about moving `virtualTimeXXX()` after `platformXXX()` in the header file?

bukepo: There will be duplicated function names.

PR #4521, opened 2020-01-31
[efr32] add nvm3 support

garyButt:
silabs nvm3 support is selected by setting following macro in
openthread-core-efr32-config.h :-
#define OPENTHREAD_CONFIG_PLATFORM_FLASH_API_ENABLE 0

PR #4552, opened 2020-02-11
[flash] move flash storage driver into src/core/utils

jwhui:
This is a proposal to move the flash storage driver from `examples/platforms/utils` to `src/core/utils`. This effectively promotes the flash storage driver from being an example to a supported OpenThread feature.

This proposal introduces a new build config `OPENTHREAD_CONFIG_PLATFORM_FLASH_API_ENABLE` that controls whether the OpenThread library calls `otPlatSettings*` APIs or the newly-introduces `otPlatFlash*` APIs.

This proposal also includes a clean rewrite of the flash storage implementation that retains the same flash storage format and compatibility as the existing implementation.


src/core/utils/flash.cpp
```diff
@@ -0,0 +1,268 @@
...
+            otPlatFlashWrite(&GetInstance(), mSwapIndex, offset, &record, sizeof(record));
+            error = OT_ERROR_NONE;
+        }
+
+        if ((index == 1) && (aIndex == 0))
+        {
+            record.SetFirst();
+            otPlatFlashWrite(&GetInstance(), mSwapIndex, offset, &record, sizeof(record));
+            ExitNow();
+        }
```
abtink: Reading this logic I think in some (rare) corner cases, we may get to bad state. Here is situation I can think of:
- Suppose user has already called `Add` with a bunch of values for a key. Say there are 3 records: `r0(first), r1, r2` (first one is marked as `IsFirst`).
- User now calls `Set` for the same key, it would add a new record at the end with `IsFirst` set -> `r0(first), r1, r2, n0(first)`.
- Now user issues a `Delete` for index 0 (we can have a vulnerability here):
  - We run through the algo: `r0` is marked as deleted, `r1` is marked as first. We get to `n0` and it is marked as deleted.
  - Now the records in the flash are `r1(fist), r2.`
  - So we get some old entries back in the list.

The only way I can think of for us to get into such situation is if user mixes  `Set` and `Add` for the same key. I think we currently don't use this from core, but would be good to be aware of this limitation.

src/core/common/instance.hpp
```diff
@@ -323,9 +323,10 @@ class Instance : public otInstance
     // Notifier, Settings, and MessagePool are initialized  before
     // other member variables since other classes/objects from their
     // constructor may use them.
-    Notifier    mNotifier;
-    Settings    mSettings;
-    MessagePool mMessagePool;
+    Notifier       mNotifier;
+    Settings       mSettings;
+    SettingsDriver mSettingsDriver;
```
abtink: [commit 2]: Can we consider making `mSettingsDriver` be a member variable of `Settings` class instead?
I think `Settings` would be the only user of the driver. We can then consider not providing `Instance::Get<SettingsDriver>`.

Thoughts?

src/core/common/settings.hpp
```diff
@@ -49,6 +50,102 @@ namespace MeshCoP {
...
+class SettingsDriver : public InstanceLocator
+{
+public:
+    /**
+     * Constructor.
+     */
+    explicit SettingsDriver(Instance &aInstance);
+
+    /**
+     * This method initialize the settings storage driver.
```
abtink: [commit 2] nit: 
```suggestion
     * This method initializes the settings storage driver.
```

src/core/common/settings.hpp
```diff
@@ -49,6 +50,102 @@ namespace MeshCoP {
...
+     *
+     * @retval OT_ERROR_NONE     The value was added.
+     * @retval OT_ERROR_NO_BUFS  Not enough space to store the value.
+     *
+     */
+    otError Add(uint16_t aKey, const uint8_t *aValue, uint16_t aValueLength);
+
+    /**
+     * This method removes a value from @p aKey.
+     *
```
abtink: [commit 2] - nit: extra line.
```suggestion
```

src/core/utils/flash.cpp
```diff
@@ -0,0 +1,268 @@
...
+                if (readLength > record.GetLength())
+                {
+                    readLength = record.GetLength();
+                }
+
+                otPlatFlashRead(&GetInstance(), mSwapIndex, offset + sizeof(record), aValue, readLength);
+            }
+
+            valueLength = record.GetLength();
+            error       = OT_ERROR_NONE;
```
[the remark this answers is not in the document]
jwhui: Correct. It would have been nice to have all superseded entries marked with the "delete" flag to avoid unnecessary `memcpy`, but depending on that now would break compatibility with the existing implementation.

src/core/utils/flash.cpp
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        index++;
+    }
+
+    if (aValueLength)
+    {
+        *aValueLength = valueLength;
+    }
```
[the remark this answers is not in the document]
jwhui: Updating `aValueLength` inside the for-loop could cause the read length to be shorter that desired in subsequent for-loop iterations.

src/core/common/instance.hpp
jwhui: I'm not sure this would work without additional changes. This would result in multiple instances of `mSettingsDriver` given that `ChildInfoIterator` inherits from `SettingsBase`. Correct?

src/core/utils/flash.cpp
jwhui: Yes, thanks for catching that! I updated the implementation and unit tests to properly handle/check this scenario.

src/core/utils/flash.cpp
```diff
@@ -0,0 +1,266 @@
...
+        if ((aIndex == index) || (aIndex == -1))
+        {
+            record.SetDeleted();
+            otPlatFlashWrite(&GetInstance(), mSwapIndex, offset, &record, sizeof(record));
+            error = OT_ERROR_NONE;
+        }
+
+        if ((index == 1) && (aIndex == 0))
+        {
+            record.SetFirst();
```
srickardti: Is it necessary to mark an item as "first", shouldn't it being the first found indicate that it is the first?

Write and erase operations can be pretty expensive in time, power, and line wearing.

examples/platforms/cc1352/flash.c
```diff
@@ -28,43 +28,21 @@
...
-/*
- * The settings configuration page number _SHOULD_ be defined in the core
- * config file.
- */
-#ifndef SETTINGS_CONFIG_PAGE_NUM
-#warn "SETTINGS_CONFIG_PAGE_NUM not defined in the OpenThread Core Config"
-#endif /* SETTINGS_CONFIG_PAGE_NUM */
+#define FLASH_BASE_ADDRESS 0x52000
+#define FLASH_PAGE_SIZE 0x2000
+#define FLASH_PAGE_NUM 2
```
srickardti: This value must be a multiple of `2`. Might want to add a comment

examples/platforms/cc1352/flash.c
```diff
@@ -158,117 +136,84 @@ static void restoreFlashCache(uint32_t mode)
...
-    uint32_t fsmRet;
-    otError  ret;
 
-    otEXPECT_ACTION(checkVoltage(), ret = OT_ERROR_FAILED);
+    otEXPECT(checkVoltage());
 
     mode = disableFlashCache();
 
-    fsmRet = FlashSectorErase(aAddress);
+    FlashSectorErase(mapAddress(aSwapIndex, 0));
```
srickardti: This will only erase the page pointed at by `mapAddress`. If `FLASH_PAGE_NUM` is `!=2` this will not erase all necessary pages in the swap index.

examples/platforms/cc2652/flash.c
```diff
@@ -28,43 +28,21 @@
...
-/*
- * The settings configuration page number _SHOULD_ be defined in the core
- * config file.
- */
-#ifndef SETTINGS_CONFIG_PAGE_NUM
-#warn "SETTINGS_CONFIG_PAGE_NUM not defined in the OpenThread Core Config"
-#endif /* SETTINGS_CONFIG_PAGE_NUM */
+#define FLASH_BASE_ADDRESS 0x52000
+#define FLASH_PAGE_SIZE 0x2000
+#define FLASH_PAGE_NUM 2
```
srickardti: cc1352 and cc2652 function the same in this case, the comments would also apply here.

src/core/utils/flash.cpp
jwhui: The existing `Set()` implementation does not mark older records as deleted/invalid. Instead, `Set()` uses the "first" flag to indicate that all older records should be ignored. Does that make sense?

src/core/common/instance.hpp
abtink: You are right. That won't work.

src/core/utils/flash.cpp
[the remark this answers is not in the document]
abtink: Very good point! Thanks.

src/core/utils/flash.cpp
abtink: How about if we later (future PR?) change the requirement for `Set` and `Add` (in `otPlatSettings{name}` as well) such we don't allow user to mix and match `Set` and `Add` (considering it as incorrect use and undefined behavior (not supported)). 

Basically expect a key to either be used as a list (use `Add`) or as a single value (use `Set`).

Thoughts?

jwhui: We can consider this in a future PR. Just curious, how would you enforce this?

abtink: Not sure if there'd be an easy way to enforce it in the code. 
It would be just changing the documentation (and making sure we follow the new policy in core - which we already do).

src/core/utils/flash.cpp
```diff
@@ -0,0 +1,266 @@
...
+            continue;
+        }
+
+        otPlatFlashRead(&GetInstance(), mSwapIndex, srcOffset, &record, record.GetSize());
+        otPlatFlashWrite(&GetInstance(), dstIndex, dstOffset, &record, record.GetSize());
+        dstOffset += record.GetSize();
+    }
+
+exit:
+    otPlatFlashWrite(&GetInstance(), dstIndex, 0, &sSwapActive, sizeof(sSwapActive));
```
bukepo: shall we store this in fixed little-endian?

jwhui: Is there a use case that concerns dealing with different architectures?

bukepo: None for now, but ARM supports to run on either big/little endian, I. prefer persist in fixed endian just for some unexpected cases. I don't have strong opinion on this.

bukepo:
Should the `src/utils` in title be `src/core/utils`?

dismirlian:
@jwhui I think this implementation is much cleaner than the old one, thanks!!

I would like to propose a minor change in order to support multiple instances of the settings area (as in the PR #4589, now closed).

The use case I foresee is that the application running alongside OT would probably benefit from a key-value store, and it would be ideal to take advantage of this implementation instead of duplicating the effort/code space. Of course, the application could use the same settings area as the OT core, but this would mean that the OT factory reset would also wipe the application data. So I propose to have the possibility of having to separate settings areas.

This change should come with zero (or almost zero) overhead for those who don't want multiple instances.

One possible idea to implement this would be to change `Flash` to accept a pointer to an `ot::Instance` instead of a reference. A `NULL` pointer would identify the application settings area. This has the advantage of being very simple, but the disadvantage of not allowing more than one application settings area (I don't know if anyone would need more than one). Also, it's not as elegant ;).

examples/platforms/efr32mg12/flash.c
```diff
@@ -31,114 +31,66 @@
...
 {
-    return aAddress + FLASH_DATA_START_ADDR;
-}
+    uint32_t address;
 
-static otError returnTypeConvert(int32_t aStatus)
-{
-    otError error = OT_ERROR_NONE;
+    address = FLASH_DATA_START_ADDR + aOffset;
+    ;
```
dismirlian: extra `;`?

examples/platforms/efr32mg12/flash.c
```diff
@@ -31,114 +31,66 @@
...
 {
-    MSC_Init();
-    return OT_ERROR_NONE;
-}
+    OT_UNUSED_VARIABLE(aInstance);
 
-uint32_t utilsFlashGetSize(void)
-{
-    return FLASH_DATA_END_ADDR - FLASH_DATA_START_ADDR;
+    MSC_Init();
```
dismirlian: This unlocks the flash for writing, leaving it unlocked. We could introduce a `otPlatFlashLock/Unlock` API, to reduce unintended writes. `otPlatFlashInit` would do nothing in this case. What do you think?

src/core/utils/flash.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+        otPlatFlashRead(&GetInstance(), mSwapIndex, 0, &swapMarker, sizeof(swapMarker));
+
+        if (swapMarker == sSwapActive)
+        {
+            break;
+        }
+    }
+
+    for (mSwapUsed = kSwapMarkerSize; mSwapUsed < mSwapSize; mSwapUsed += record.GetSize())
```
dismirlian: This loop does not guarantee that `otPlatFlashRead` will not read past the end of the swap block. Maybe `mSwapUsed < mSwapSize - sizeof(record)`?

src/core/utils/flash.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+otError Flash::Get(uint16_t aKey, int aIndex, uint8_t *aValue, uint16_t *aValueLength) const
+{
+    otError      error       = OT_ERROR_NOT_FOUND;
+    uint16_t     valueLength = 0;
+    int          index       = 0;
+    uint32_t     offset;
+    RecordHeader record;
+
+    for (offset = kSwapMarkerSize; offset < mSwapUsed; offset += record.GetSize())
```
dismirlian: This loop does not guarantee that `otPlatFlashRead` will not read past the end of the swap block. Maybe `mSwapUsed < mSwapSize - sizeof(record)`?

src/core/utils/flash.cpp
```diff
@@ -0,0 +1,266 @@
...
+exit:
+    return error;
+}
+
+bool Flash::DoesValidRecordExist(uint32_t aOffset, uint16_t aKey) const
+{
+    RecordHeader record;
+    bool         rval = false;
+
+    for (; aOffset < mSwapUsed; aOffset += record.GetSize())
```
dismirlian: This loop does not guarantee that `otPlatFlashRead` will not read past the end of the swap block. Maybe `aOffset < mSwapSize - sizeof(record)`?

src/core/utils/flash.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+void Flash::Swap(void)
+{
+    uint8_t  dstIndex  = !mSwapIndex;
+    uint32_t dstOffset = kSwapMarkerSize;
+    Record   record;
+
+    otPlatFlashErase(&GetInstance(), dstIndex);
+
+    for (uint32_t srcOffset = kSwapMarkerSize; srcOffset < mSwapUsed; srcOffset += record.GetSize())
```
dismirlian: This loop does not guarantee that `otPlatFlashRead` will not read past the end of the swap block. Maybe `srcOffset < mSwapSize - sizeof(record)`?

src/core/utils/flash.cpp
```diff
@@ -0,0 +1,266 @@
...
+    mSwapUsed  = dstOffset;
+}
+
+otError Flash::Delete(uint16_t aKey, int aIndex)
+{
+    otError      error = OT_ERROR_NOT_FOUND;
+    int          index = 0;
+    RecordHeader record;
+
+    for (uint32_t offset = kSwapMarkerSize; offset < mSwapUsed; offset += record.GetSize())
```
dismirlian: This loop does not guarantee that `otPlatFlashRead` will not read past the end of the swap block. Maybe `offset < mSwapSize - sizeof(record)`?

src/core/utils/flash.cpp
jwhui: `mSwapUsed` is intended to mark where the records end within a swap, not the end of the swap. Can you give an example of how we might loop past the end of the swap block?

examples/platforms/efr32mg12/flash.c
dismirlian: Also, instead of the proposed APIs, we could just lock/unlock in `otPlatFlashErase/Write`.

jwhui: I agree we should look into this further. However, given that this PR does not change this behavior, I would prefer to address this in a separate PR. Does that work for you?

dismirlian: Yes, of course.

dismirlian:
If I understand the code correctly, I think there may some small problems regarding power-failure robustness:

1. We always assume that the record with `kFlagAddBegin` set will also have its `mLength` correctly set. If the firmware fails to write the whole header (for instance, if it resets after writing `mFlags`), then the record is left in an inconsistent state. For example, `mFlags` with `kFlagAddBegin` but `mLength` erased (=65535). This would have negative side effects, because it impacts all the methods that iterate over the records, specifically, the `Init` method only checks for `record.IsAddBeginSet()` and then reads the length:
https://github.com/openthread/openthread/blob/381f4deda9a6a3d986261046fc2618743ac2422d/src/core/utils/flash.cpp#L69-L73
After `Init` initializes `mSwapUsed` incorrectly, all other methods would iterate over garbage.
The same happens in `Swap`.

2. We always assume that the space after the last record is "writeable" (that is to say, it is erased). This may not be the case if the `mKey` was written but the `mFlags` wasn't.

A possible solution to both issues would be to enhance the `Init` method to "sanitize" the settings area:
- If a record is found whose `mFlags` has `kFlagAddBegin` but not `kFlagComplete`, we can assume an interrupted operation, should not count that record in `mSwapUsed`, and trigger a `Swap`. This should get rid of the invalid record.
- After the last (valid) record, `Init` should scan the remaining space in the swap block and trigger a `Swap` if any byte is != 0xff. This will ensure that the remaining space is "writable".

dismirlian:
There are several places in the code where the following happens:

```
for (uint32_t someOffset = kSwapMarkerSize; someOffset < mSwapUsed; someOffset += record.GetSize())
{
    otPlatFlashRead(&GetInstance(), mSwapIndex, someOffset, &record, sizeof(record));
```

This loop does not guarantee that `otPlatFlashRead` will not read past the end of the swap block. I think we should use `someOffset < mSwapUsed - sizeof(record)`. What do you think?

jwhui:
> A possible solution to both issues would be to enhance the `Init` method to "sanitize" the settings area:
> 
> * If a record is found whose `mFlags` has `kFlagAddBegin` but not `kFlagComplete`, we can assume an interrupted operation, should not count that record in `mSwapUsed`, and trigger a `Swap`. This should get rid of the invalid record.
> * After the last (valid) record, `Init` should scan the remaining space in the swap block and trigger a `Swap` if any byte is != 0xff. This will ensure that the remaining space is "writable".

Thanks for calling out these issues! I agree we should address these issues and your proposed solution sounds good to me. We can address them in a follow-on PR once this PR has been approved and merged.

dismirlian:
It sounds reasonable, I'll propose a PR to address these issues once this one is merged. Thanks!

nigelg-silabs:
Hi @jwhui, Can we merge this yet - we would like to add our NV PR and this is still blocking it

LuDuda:
We are currently reviewing this. It should be finished soon.

dismirlian:
I think there may be an issue with the efr32 platform:

The MG12 datasheet states:
> It is possible to write words twice between each erase by keeping at 1 the bits that are not to be changed. [...] Note that there is a maximum of two writes to the same word between each erase due to a physical limitation of the flash.

The MG21 datasheet says something similar:

> The Flash memory is organized into 64-bit wide double-words. Each 64-bit double-word can be written only twice between erase cycles

The current implementation writes the `RecordHeader` twice during `Add` and once more during `Delete`.

I don't know what implications this may have. Maybe increased wear or decreased reliability/retention? Perhaps someone from Silabs can shed some light on this... @nigelg-silabs? Also, I think the original implementation does the same thing.

jwhui:
> I don't know what implications this may have. Maybe increased wear or decreased reliability/retention? Perhaps someone from Silabs can shed some light on this... @nigelg-silabs? Also, I think the original implementation does the same thing.

@dismirlian, I believe the EFR32 will be moving away from this implementation. See https://github.com/openthread/openthread/pull/4521

examples/platforms/nrf528xx/src/flash.c
```diff
@@ -67,7 +73,7 @@ otError utilsFlashInit(void)
         sFlashDataEnd = pageSize * codeSize;
     }
 
-    sFlashDataStart = sFlashDataEnd - (pageSize * SETTINGS_CONFIG_PAGE_NUM);
+    sFlashDataStart = sFlashDataEnd - (pageSize * FLASH_PAGE_NUM);
```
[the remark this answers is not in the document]
jwhui: @LuDuda, thanks for catching this!

To clarify, by "first option", you meant updating the linker script?

Unfortunately, we don't have any real data that reflects flash write/erase frequency. Of course, it largely depends on a given deployment's dynamics (i.e. stable or not). If you would rather be conservative, we can increase `FLASH_PAGE_NUM` back to 4 for nrf52833 and nrf52840.

Thoughts?

src/core/utils/flash.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+otError Flash::Get(uint16_t aKey, int aIndex, uint8_t *aValue, uint16_t *aValueLength) const
+{
+    otError      error       = OT_ERROR_NOT_FOUND;
+    uint16_t     valueLength = 0;
+    int          index       = 0;
+    uint32_t     offset;
+    RecordHeader record;
+
+    for (offset = kSwapMarkerSize; offset < mSwapUsed; offset += record.GetSize())
```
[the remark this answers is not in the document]
jwhui: The assumption is that the `otPlatFlashRead()` in the body of the for-loop will populate `record`. Am I missing something?

examples/platforms/nrf528xx/src/flash.c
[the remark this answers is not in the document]
jwhui: @LuDuda, I updated this PR with `FLASH_PAGE_NUM` set to 4 for nrf52833 and nrf52840 to match the existing behavior.

src/core/utils/flash.cpp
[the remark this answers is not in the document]
LuDuda: True, I knew I was missing something. Thank you for reminding me how the "for" loop works :+1: I checked other places, and I don't see any problems.

examples/platforms/nrf528xx/src/flash.c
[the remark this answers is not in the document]
LuDuda: That works, thank you!

dismirlian:
Hi @jwhui, thanks for your response. Yes, I saw that PR. Too bad for those who have devices in the field; either they will update to PR #4521 and break compatibility with those devices, or they stay with the original/this implementation which violates the datasheet recommendations. Thanks!

jwhui:
> Hi @jwhui, thanks for your response. Yes, I saw that PR. Too bad for those who have devices in the field; either they will update to PR #4521 and break compatibility with those devices, or they stay with the original/this implementation which violates the datasheet recommendations. Thanks!

@dismirlian, can you open a new GitHub issue to address this concern?

dismirlian:
> > Hi @jwhui, thanks for your response. Yes, I saw that PR. Too bad for those who have devices in the field; either they will update to PR #4521 and break compatibility with those devices, or they stay with the original/this implementation which violates the datasheet recommendations. Thanks!
> 
> @dismirlian, can you open a new GitHub issue to address this concern?

Of course. Anyways, maybe no one is impacted by this (in our case, we are conducting some feasability tests, so there's no impact for us). I could open an issue to see to what extent this is important.

PR #4553, opened 2020-02-11
[coap] fix comparison with block size validity check

nealjack:
Fix for #4513

Accidentally introduced bug when checking for valid block size. My bad!

I've tested this fix with an outgoing block1 transfer.

### DIFF
--- a/src/core/coap/coap_message.cpp
+++ b/src/core/coap/coap_message.cpp
@@ -193,7 +193,7 @@ otError Message::AppendBlockOption(Message::BlockType aType, uint32_t aNum, bool
     uint32_t encoded = aSize;
 
     VerifyOrExit(aType == kBlockType1 || aType == kBlockType2, error = OT_ERROR_INVALID_ARGS);
-    VerifyOrExit(aSize >= OT_COAP_BLOCK_SIZE_32 && aSize <= OT_COAP_BLOCK_SIZE_1024, error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(aSize <= OT_COAP_BLOCK_SIZE_1024, error = OT_ERROR_INVALID_ARGS);
     VerifyOrExit(aNum < kBlockNumMax, error = OT_ERROR_INVALID_ARGS);
 
     encoded |= static_cast<uint32_t>(aMore << kBlockMOffset);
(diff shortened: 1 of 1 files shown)


src/core/coap/coap_message.cpp
```diff
@@ -193,7 +193,7 @@ otError Message::AppendBlockOption(Message::BlockType aType, uint32_t aNum, bool
     uint32_t encoded = aSize;
 
     VerifyOrExit(aType == kBlockType1 || aType == kBlockType2, error = OT_ERROR_INVALID_ARGS);
-    VerifyOrExit(aSize < OT_COAP_BLOCK_SIZE_32 || aSize > OT_COAP_BLOCK_SIZE_1024, error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(aSize > OT_COAP_BLOCK_SIZE_32 || aSize < OT_COAP_BLOCK_SIZE_1024, error = OT_ERROR_INVALID_ARGS);
```
sjlongland: Not sure that equation is quite right…

* if `aSize=OT_COAP_BLOCK_SIZE_32`:
** `(aSize > OT_COAP_BLOCK_SIZE_32) = false`
** `(aSize < OT_COAP_BLOCK_SIZE_1024) = true`
** resultant is `true`

* if `aSize=OT_COAP_BLOCK_SIZE_1024`:
** `(aSize > OT_COAP_BLOCK_SIZE_32) = true`
** `(aSize < OT_COAP_BLOCK_SIZE_1024) = false`
** resultant is `true`

* if `aSize=OT_COAP_BLOCK_SIZE_16`:
** `(aSize > OT_COAP_BLOCK_SIZE_32) = false`
** `(aSize < OT_COAP_BLOCK_SIZE_1024) = true`
** resultant is `true`

I think you might mean: `aSize >= OT_COAP_BLOCK_SIZE_32 && aSize <= OT_COAP_BLOCK_SIZE_1024`

nealjack: Yep, my bad!

sjlongland: For what its worth, block size=16 is legitimate too (`szx=0`):
```
   The block size is encoded using a three-bit unsigned integer (0 for
   2**4 bytes to 6 for 2**10 bytes), which we call the "SZX" ("size
   exponent"); the actual block size is then "2**(SZX + 4)".
```
-- https://tools.ietf.org/html/rfc7959#page-8

The need for such a small block size might seem frivolous, but if you've got really patchy communications, it could be handy to have anyway.

nealjack: Yeah, so in that case we would just check for `<= OT_COAP_BLOCK_SIZE_1024`, seeing as the enum is interpreted as unsigned.

PR #4554, opened 2020-02-12
[coap] smaller enhancements

abtink:
**Edit**: New pushed commit rebased all changes on master and squashes them into a single commit. There are some new changes as well:

This commits contains the following changes:
- Declare `Metadata` as a private inner class of `CoapBase`. Also replace constructors with `Init()` method and removes the default empty constructor (since it is not needed).
- Add `ResponseMetadata` as private inner struct in `ResposeQueue`.
- Simplify `ResponsesQueue::FindMatchedResponse()`.
- Declare `CoapBase::Sender` as a protected type.
- Use initializer list in `CoapBase` constructor.
- Use `for` loop for iterating over messages in a message queue.
- Define `ResponseHandler` and `RequestHandler` types.
- Update `TxParameters` class. Add helper method for common calculation of span window, and remove unused constant definitions.
- Move more complex method implementation from header file into `cpp` file (helps reduce code size).
- Update comments and documentations.

----------------

This PR contain a bunch of smaller changes/enhancements in `coap` module:

**[coap] declare CoapMetadata as a private inner class of CoapBase**

This commit also replaces `CoapMetadata` constructors with `Init()`
methods and removes the default empty constructor (since it is not
needed).

**[coap] add ResponseMetadata as private inner struct in ResposeQueue**

This commit renames and moves `EnqueuedResponseHeader` definition as
`ResponseMetadata` declared as a private inner `struct` in the class
`ResponseQueue`. It also removes the unnecessary constructors for this
type (replacing them with `Init()` method) and simplifies the
implementation of `GetRemainingTime()`.

**[coap] simplify ResponsesQueue::FindMatchedResponse**

**[coap] declare CoapBase::Sender as a protected type**

Note that the `Sender` type is intended for use by sub-classes of
`CoapBase` only. This commit also declares `mSender` as a `const`
to emphasis that it would not change (this may help with tool-chain
code optimization).

**[coap] use initializer list in CoapBase constructor**

**[coap] use for loop for iterating over message in the queue**

This commit also include other smaller changes in `coap.cpp`.

**[coap] simplify calculation of constants kMaxTransmitSpan and kMaxTransmitWait**

**[coap] define ResponseHandler and RequestHandler types**

**[coap] update comments**

### DIFF
--- a/src/core/coap/coap.cpp
+++ b/src/core/coap/coap.cpp
@@ -48,6 +48,8 @@ namespace Coap {
 
 CoapBase::CoapBase(Instance &aInstance, Sender aSender)
     : InstanceLocator(aInstance)
+    , mPendingRequests()
+    , mMessageId(Random::NonCrypto::GetUint16())
     , mRetransmissionTimer(aInstance, &Coap::HandleRetransmissionTimer, this)
     , mResources()
     , mContext(NULL)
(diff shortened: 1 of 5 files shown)


src/core/coap/coap.cpp
```diff
@@ -620,35 +620,39 @@ void CoapBase::ProcessReceivedRequest(Message &aMessage, const Ip6::MessageInfo
...
+void CoapBase::CoapMetadata::ReadFrom(const Message &aMessage)
+{
+    uint16_t length;
+
+    length = aMessage.Read(aMessage.GetLength() - sizeof(*this), sizeof(*this), this);
+    assert(length == sizeof(*this));
 
-    mAcknowledged = false;
-    mConfirmable  = aConfirmable;
+    OT_UNUSED_VARIABLE(length);
```
jwhui: nit: prefer at top of method body in own block

src/core/coap/coap.cpp
```diff
@@ -781,36 +782,48 @@ void ResponsesQueue::HandleTimer(Timer &aTimer)
...
+otError ResponsesQueue::ResponseMetadata::AppendTo(Message &aMessage) const
+{
+    return aMessage.Append(this, sizeof(*this));
+}
+
+void ResponsesQueue::ResponseMetadata::ReadFrom(const Message &aMessage)
+{
+    uint16_t length = aMessage.Read(aMessage.GetLength() - sizeof(*this), sizeof(*this), this);
+    assert(length == sizeof(*this));
+    OT_UNUSED_VARIABLE(length);
```
jwhui: nit: prefer at top of method body

src/core/coap/coap.cpp
```diff
@@ -63,18 +63,16 @@ CoapBase::CoapBase(Instance &aInstance, Sender aSender)
 
 void CoapBase::ClearRequestsAndResponses(void)
 {
-    Message *    message = static_cast<Message *>(mPendingRequests.GetHead());
-    Message *    messageToRemove;
+    Message *    message;
```
jwhui: nit: move declaration into for-loop initializer?

src/core/coap/coap.cpp
abtink: I actually changed it fully to remove the `OT_UNUSED_VARIABLE()` (by using the `length`).

I was not sure about the value of `assert(length == sizeof(*this))` in the original code, since it is basically verifying the behavior of `Message::Read()`.
```
    {
        uint16_t length = aMessage.Read(aMessage.GetLength() - sizeof(*this), sizeof(*this), this);
        assert(length == sizeof(*this));
        OT_UNUSED_VARIABLE(length);
    }
```

I changed it to this which instead `assert`s that `aMessage` is long enough to contain the metadata before reading it.
```
void CoapBase::CoapMetadata::ReadFrom(const Message &aMessage)
{
    uint16_t length = aMessage.GetLength();

    assert(length >= sizeof(*this));
    aMessage.Read(length - sizeof(*this), sizeof(*this), this);
}
```

src/core/coap/coap.hpp
```diff
@@ -461,6 +448,19 @@ class CoapBase : public InstanceLocator
...
 protected:
+    /**
+     * This function pointer is called to send a CoAP message.
+     *
+     * @param[in]  aCoapBase     A reference to the CoAP agent.
+     * @param[in]  aMessage      A reference to the message to send.
+     * @param[in]  aMessageInfo  A reference to the message info associated with @p aMessage.
+     *
+     * @retval OT_ERROR_NONE     Successfully sent CoAP message.
+     * @retval OT_ERROR_NO_BUFS  Failed to allocate retransmission data.
```
librasungirl: though not relative to this PR, noticing `retransmission data` here seems improper.

PR #4555, opened 2020-02-12
[build] rename example platform POSIX to simulation

bukepo:
People are confused about the two POSIX platforms: example POSIX and
POSIX-app. This PR renames the example platform POSIX to
sim(simulation), and after this is merge, another PR will rename the
current POSIX-app to POSIX.


simonlingoogle:
> People are confused about the two POSIX platforms: example POSIX and POSIX-app. 

Why is `src/posix/Makefile-posix`'s location so special? Why not `examples/Makefile-sim` and `examples/Makefile-posix`?

bukepo:
All production-level code are supposed to be under src/. @simonlingoogle

Vyrastas:
FYI after this goes in we'll have to update our Simulation Codelab and all references on openthread.io for a Posix build (which there are many).

simonlingoogle:
> FYI after this goes in we'll have to update our Simulation Codelab and all references on openthread.io for a Posix build (which there are many).

After Docs are updated, it might be misleading for users who prefer to use [releases](https://github.com/openthread/openthread/releases).

bukepo:
@Vyrastas Yes, we have to update openthread.io and codelab accordingly. However, as mentioned in description, this PR resolves confusions in longer term. Thoughts? @jwhui

jwhui:
> @Vyrastas Yes, we have to update openthread.io and codelab accordingly. However, as mentioned in description, this PR resolves confusions in longer term. Thoughts? @jwhui

Should we stage the required doc changes so that we can update source and docs at nearly the same time?

/cc/ @Vyrastas

Vyrastas:
If you can wait a day or two, I can do the documentation updates ahead of merging this.

jwhui:
> If you can wait a day or two, I can do the documentation updates ahead of merging this.

Sounds good to me. Thanks, @Vyrastas !

Vyrastas:
There is some example output on openthread.io that has this:

`wpantund[139]: NCP is running "OPENTHREAD/20170716-00967-g677d49b4; POSIX; Oct 18 2018 08:19:20"`

Is that version string going to change as part of this or will it still say "POSIX"?

jwhui:
> There is some example output on openthread.io that has this:
> 
> `wpantund[139]: NCP is running "OPENTHREAD/20170716-00967-g677d49b4; POSIX; Oct 18 2018 08:19:20"`
> 
> Is that version string going to change as part of this or will it still say "POSIX"?
> 
> From here: https://github.com/openthread/wpantund/blob/master/src/wpantund/NCPInstanceBase.cpp#L264

@Vyrastas, good catch. @bukepo, we should also change the platform info string to something like "SIM" or "SIMULATION".

https://github.com/openthread/openthread/blob/cec297475f3af986022b701751622a0604db7cd7/examples/platforms/posix/openthread-core-posix-config.h#L48

bukepo:
@Vyrastas @jwhui This PR uses `sim`, but I don't know whether we should use full `simulation`.  I'm fine with either way, but if you have any preference, I'm happy to change it.

librasungirl:
no strong preference. `simulation` might be easier in keeping wording consistence in code/comment/document.

Vyrastas:
Will the existing Docker images pick up this change too?

jwhui:
> Will the existing Docker images pick up this change too?

Yes, I believe so.

PR #4457, opened 2020-01-08
[posix-host] add command mode for ot-ctl

gjc13:
ot-ctl will now send the command line arguments to ot-daemon and print
the reply.

PR #4556, opened 2020-02-13
[cli] always output command result

bukepo:
Always output command result, i.e. "Done" on success, error message
otherwise.

Factory commands also follow the same pattern. Their existing error
messages are not removed for backward compatibility consideration.

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -430,6 +430,8 @@ void Interpreter::ProcessHelp(int argc, char *argv[])
     {
         mServer->OutputFormat("%s\r\n", mUserCommands[i].mName);
     }
+
+    AppendResult(OT_ERROR_NONE);
 }
 
 void Interpreter::ProcessBufferInfo(int argc, char *argv[])
@@ -1023,10 +1025,11 @@ void Interpreter::ProcessDiscover(int argc, char *argv[])
     mServer->OutputFormat("| J | Network Name     | Extended PAN     | PAN  | MAC Address      | Ch | dBm | LQI |\r\n");
     mServer->OutputFormat("+---+------------------+------------------+------+------------------+----+-----+-----+\r\n");
 
-    return;
-
 exit:
-    AppendResult(error);
+    if (error != OT_ERROR_NONE)
+    {
+        AppendResult(error);
+    }
 }
 
 #if OPENTHREAD_CONFIG_DNS_CLIENT_ENABLE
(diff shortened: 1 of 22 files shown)


bukepo:
This is a follow-up of #4457.

src/cli/cli.cpp
```diff
@@ -2013,6 +2014,10 @@ void Interpreter::HandleIcmpReceive(otMessage *          aMessage,
     mServer->OutputFormat("\r\n");
 
 exit:
+    if (!mPingTimer.IsRunning())
+    {
+        AppendResult(OT_ERROR_NONE);
```
simonlingoogle: if `ping` from node 1 to node 2, node 2 is also printing `Done` because it's `HandleIcmpReceive` is called. Should make sure `aIcmpHeader->mType == OT_ICMP6_TYPE_ECHO_REPLY`

src/cli/cli.cpp
```diff
@@ -2013,6 +2014,10 @@ void Interpreter::HandleIcmpReceive(otMessage *          aMessage,
     mServer->OutputFormat("\r\n");
 
 exit:
+    if (!mPingTimer.IsRunning())
```
simonlingoogle: It is logical to print `Done` after `ping` have received ICMP replies. However there are some issues regarding to this behavior:
- `ping` is not guaranteed to receive ICMP echo reply, violating the rule that `CLI commands always output command result`
- There can be a long delay between `ping ...` command and `Done`. I run `ping count=5` and `scan` at the same time, got following output:
```
> ping fdde:ad00:beef:0:6926:9463:d7be:21e6 4 5 1 64
12 bytes from fdde:ad00:beef:0:6926:9463:d7be:21e6: icmp_seq=24 hlim=64 time=253ms
12 bytes from fdde:ad00:beef:0:6926:9463:d7be:21e6: icmp_seq=25 hlim=64 time=1ms
12 bytes from fdde:ad00:beef:0:6926:9463:d7be:21e6: icmp_seq=26 hlim=64 time=1ms
> scan
| J | Network Name     | Extended PAN     | PAN  | MAC Address      | Ch | dBm | LQI |
+---+------------------+------------------+------+------------------+----+-----+-----+
| 0 | OpenThread       | dead00beef00cafe | face | 221f8796816c1847 | 11 | -20 |   0 |
Done // scan
12 bytes from fdde:ad00:beef:0:6926:9463:d7be:21e6: icmp_seq=27 hlim=64 time=4067ms
Done // what is this?
12 bytes from fdde:ad00:beef:0:6926:9463:d7be:21e6: icmp_seq=28 hlim=64 time=3070ms
Done // ping
```
It becomes become difficult to track which `Done` is for which command, since commands and `Done`s are not one after another.  Also there are more than 2 `Done`s, should be a bug.

simonlingoogle: > Also there are more than 2 Dones, should be another issue.

Should be due to that the pings are delayed by `scan`, causing node have receive multiple ping replies after `mPingTimer` stops.

bukepo:
@abtink @jwhui any comments?

PR #4558, opened 2020-02-13
[build] move spinel into src/utils

bukepo:
Move spinel from src/ncp into src/utils and remove dependency on
OpenThread's core configurations.


src/utils/spinel_encoder.hpp
```diff
@@ -33,18 +33,15 @@
...
 
-#include "openthread-core-config.h"
-#include "ncp/ncp_buffer.hpp"
-#include "ncp/spinel.h"
+#include "spinel.h"
+#include "spinel_buffer.hpp"
 
 namespace ot {
-namespace Ncp {
+namespace Utils {
```
abtink: How about using `Spinel` instead of `Utils` as namespace name for all the spinel modules?
This way we distinguish it from `ot:Utils` namespace from `src/core/utils` modules.

src/utils/spinel_encoder.hpp
```diff
@@ -541,7 +538,7 @@ class SpinelEncoder
      */
     otError WriteDataWithLen(const uint8_t *aData, uint16_t aDataLen);
 
-#if OPENTHREAD_MTD || OPENTHREAD_FTD
+#if OPENTHREAD_MESSAGE_IN_SPINEL_BUFFER
```
abtink: Can we consider adding a new header file `src/utils/ot-spinel-config.h` for all config options for spinel modules/library. This way we can add documentation (describing what the config is intended to do) and have a default value for each config (which can also help with "undefined marco warning").

Thoughts?

src/ncp/ncp_base.hpp
```diff
@@ -208,7 +207,7 @@ class NcpBase
         NcpBase::PropertyHandler mHandler;
     };
 
-    NcpFrameBuffer::FrameTag GetLastOutboundFrameTag(void);
+    ot::Spinel::Buffer::FrameTag GetLastOutboundFrameTag(void);
```
abtink: Just curios: do we need the `ot::`?

src/posix/platform/radio_spinel.cpp
```diff
@@ -66,6 +66,8 @@
 #define TX_WAIT_US (5 * US_PER_S)
 #endif
 
+using ot::Spinel::Decoder;
```
abtink: Just curios. Why do we need this?

src/spinel/spinel_decoder.hpp
```diff
@@ -33,22 +33,21 @@
...
 namespace ot {
-namespace Ncp {
+namespace Spinel {
 
 /**
  * This class defines a spinel decoder.
  *
  */
-class SpinelDecoder
+class Decoder
```
abtink: :+1:  Thanks for renaming and removing `Spinel` prefix.

src/ncp/example_vendor_hook.cpp
```diff
@@ -34,6 +34,8 @@
 
 #include "ncp_base.hpp"
 
+using namespace ot::Utils;
```
jwhui: Is this necessary? We generally try to discourage `using namespace` (with the exception of `src/core/api`).

PR #4564, opened 2020-02-14
[dtls] fix allocation and deallocation of ssl cookies

mnp222:
Cookies were being freed on the client when they were never allocated,
causing a hard fault.

### DIFF
--- a/src/core/meshcop/dtls.cpp
+++ b/src/core/meshcop/dtls.cpp
@@ -110,10 +110,7 @@ Dtls::Dtls(Instance &aInstance, bool aLayerTwoSecurity)
 void Dtls::FreeMbedtls(void)
 {
 #if defined(MBEDTLS_SSL_SRV_C) && defined(MBEDTLS_SSL_COOKIE_C)
-    if (mCookieCtx.hmac_ctx.hmac_ctx != NULL)
-    {
-        mbedtls_ssl_cookie_free(&mCookieCtx);
-    }
+    mbedtls_ssl_cookie_free(&mCookieCtx);
 #endif
 #if OPENTHREAD_CONFIG_COAP_SECURE_API_ENABLE
 #ifdef MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED
@@ -263,6 +260,17 @@ otError Dtls::Setup(bool aClient)
 #endif // MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED
 #endif // OPENTHREAD_CONFIG_COAP_SECURE_API_ENABLE
 
+#if defined(MBEDTLS_SSL_SRV_C) && defined(MBEDTLS_SSL_COOKIE_C)
+    mbedtls_ssl_cookie_init(&mCookieCtx);
+    if (!aClient)
+    {
+        rval = mbedtls_ssl_cookie_setup(&mCookieCtx, mbedtls_ctr_drbg_random, Random::Crypto::MbedTlsContextGet());
+        VerifyOrExit(rval == 0);
+
+        mbedtls_ssl_conf_dtls_cookies(&mConf, mbedtls_ssl_cookie_write, mbedtls_ssl_cookie_check, &mCookieCtx);
+    }
+#endif
+
     rval = mbedtls_ssl_config_defaults(&mConf, aClient ? MBEDTLS_SSL_IS_CLIENT : MBEDTLS_SSL_IS_SERVER,
                                        MBEDTLS_SSL_TRANSPORT_DATAGRAM, MBEDTLS_SSL_PRESET_DEFAULT);
     VerifyOrExit(rval == 0);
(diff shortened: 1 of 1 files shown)


src/core/meshcop/dtls.cpp
```diff
@@ -287,7 +290,7 @@ otError Dtls::Setup(bool aClient)
     mbedtls_ssl_conf_handshake_timeout(&mConf, 8000, 60000);
     mbedtls_ssl_conf_dbg(&mConf, HandleMbedtlsDebug, this);
 
-#ifdef MBEDTLS_SSL_SRV_C
+#if defined(MBEDTLS_SSL_SRV_C) && defined(MBEDTLS_SSL_COOKIE_C)
     if (!aClient)
     {
         mbedtls_ssl_cookie_init(&mCookieCtx);
```
wgtdkp: We will still get hard fault if we failed before reaching here. Because we will call `FreeMbedtls` before  exiting if there are errors. see https://github.com/openthread/openthread/blob/master/src/core/meshcop/dtls.cpp#L343

Suggestion: init cookie at the beginning of this function.

romacdon: @wgtdkp - Perhaps I'm missing something but I believe the changes @mnp222 made to FreeMbedtls will prevent the hard fault you mention.

wgtdkp: @romacdon cookie will be freed before getting initialized if [this line](https://github.com/openthread/openthread/blob/master/src/core/meshcop/dtls.cpp#L263) fails.
We should do all `mbedtls_*_init(...)` stuff at the beginning of this function.

romacdon: @wgtdkp I see.  @mnp222 made fix to address an issue associated with client side cookies.  Your additional request is to fix a potential issue with the server side cookie.

wgtdkp: Since the bug is freeing cookie before initialization, I'd like to fix it (both client and server side) in this PR.

romacdon: @wgtdkp yes, understood and I agree.  Thank you.

mnp222: Hi Kang. You are correct and I agree we need to make this safe in all cases.  I just committed a change that only calls free when there is something to be freed, so it should take care of all cases, including the server and including when a failure happens before allocation.

wgtdkp: @mnp222 Thanks for you update.
```c++
if (mCookieCtx.hmac_ctx.hmac_ctx != NULL)
{
    mbedtls_ssl_cookie_free(&mCookieCtx);
}
```
But I don't think it is a good idea to reference to the implementation of `mbedtls_ssl_cookie_ctx` and depend on it. We should stick to the mbedtls cookie APIs (`mbedtls_ssl_cookie_*`) and they should be enough.

Besides, OpenThread coding style puts declarations and initialization at the beginning of a function, we should do so with cookie.

mnp222: I agree Kang.  I spent more time investigating the reason for the fault, and I discovered the real issue in the documentation for mbedtls md.h:

```
/**                                                                                                                                                                  
 * \brief           This function clears the internal structure of \p ctx and                                                                                        
 *                  frees any embedded internal structure, but does not free                                                                                         
 *                  \p ctx itself.                                                                                                                                   
 *                                                                                                                                                                   
 *                  If you have called mbedtls_md_setup() on \p ctx, you must                                                                                        
 *                  call mbedtls_md_free() when you are no longer using the                                                                                          
 *                  context.                                                                                                                                         
 *                  Calling this function if you have previously                                                                                                     
 *                  called mbedtls_md_init() and nothing else is optional.                                                                                           
 *                  You must not call this function if you have not called                                                                                           
 *                  mbedtls_md_init().                                                                                                                               
 */
void mbedtls_md_free( mbedtls_md_context_t *ctx );
```

So the issue happens when mbedtls_ssl_cookie_free() (which calls mbedtls_md_free()) is called without calling mbedtls_ssl_cookie_init() (which calls mbedtsl_md_init()). 

I fixed the dtls.cpp code so that mbedtls_ssl_cookie_init() is called at the beginning of Setup(), before any failure can occur, and for either client or server.  That way in FreeMbedtls(), mbedtls_ssl_cookie_free() can be called to cleanup regardless of what happened. This fixes the fault for me.

Unfortunately some Travis tests are failing, so I'll have to look into that.

src/core/meshcop/dtls.cpp
```diff
@@ -260,6 +260,17 @@ otError Dtls::Setup(bool aClient)
 #endif // MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED
 #endif // OPENTHREAD_CONFIG_COAP_SECURE_API_ENABLE
 
+#if defined(MBEDTLS_SSL_SRV_C) && defined(MBEDTLS_SSL_COOKIE_C)
+    mbedtls_ssl_cookie_init(&mCookieCtx);
+    if (!aClient)
+    {
+        rval = mbedtls_ssl_cookie_setup(&mCookieCtx, mbedtls_ctr_drbg_random, Random::Crypto::MbedTlsContextGet());
```
jwhui: This needs to move after the call to `mbedtls_ssl_config_defaults()`, which includes the following:
https://github.com/openthread/openthread/blob/b339a8946bfa8aa6fbf5ed967471f18a85fbad87/third_party/mbedtls/repo/library/ssl_tls.c#L9093-L9096

src/core/meshcop/dtls.cpp
wgtdkp: @mnp222 it seems the travis failures are all related to commissioning, which is the only feature depends on dtls. I think it is cause by the issue here https://github.com/openthread/openthread/pull/4564#discussion_r384184881

src/core/meshcop/dtls.cpp
```diff
@@ -259,11 +259,24 @@ otError Dtls::Setup(bool aClient)
...
+#if defined(MBEDTLS_SSL_SRV_C) && defined(MBEDTLS_SSL_COOKIE_C)
+    mbedtls_ssl_cookie_init(&mCookieCtx);
+#endif
 
     rval = mbedtls_ssl_config_defaults(&mConf, aClient ? MBEDTLS_SSL_IS_CLIENT : MBEDTLS_SSL_IS_SERVER,
                                        MBEDTLS_SSL_TRANSPORT_DATAGRAM, MBEDTLS_SSL_PRESET_DEFAULT);
     VerifyOrExit(rval == 0);
 
+#if defined(MBEDTLS_SSL_SRV_C) && defined(MBEDTLS_SSL_COOKIE_C)
+    if (!aClient)
```
wgtdkp: We should put this back to its original place. So that there won't be unnecessary diff :)

jwhui:
@mnp222, thanks for identifying this issue and working to resolve it!

mnp222:
Thank you @wgtdkp and @jwhui for the help in finding the best fix!  I'm impressed with your thoroughness and speedy responses.

PR #4568, opened 2020-02-17
[posix-host] enable platform udp/netif by default in daemon mode

gjc13:


src/posix/platform/openthread-core-posix-config.h
```diff
@@ -81,4 +81,11 @@
 #ifndef OPENTHREAD_CONFIG_PLATFORM_RADIO_COEX_ENABLE
 #define OPENTHREAD_CONFIG_PLATFORM_RADIO_COEX_ENABLE 1
 #endif
+
+// Platform udp and platform netif is necessary for daemon mode to work.
+#if OPENTHREAD_ENABLE_POSIX_APP_DAEMON
+#define OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE 1
```
bukepo:
```suggestion
#ifndef OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE
#define OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE 1
#endif
```

src/posix/platform/openthread-core-posix-config.h
```diff
@@ -81,4 +81,18 @@
 #ifndef OPENTHREAD_CONFIG_PLATFORM_RADIO_COEX_ENABLE
 #define OPENTHREAD_CONFIG_PLATFORM_RADIO_COEX_ENABLE 1
 #endif
+
+// Platform udp and platform netif is necessary for daemon mode to work.
```
bukepo:
```suggestion
```
Daemon mode doesn't rely on these twos features. But I think it a good idea to enable them by default in this mode.

PR #4569, opened 2020-02-18
[locator] adding InstanceLocatorInit (providing default constructor)

abtink:
This PR contains a group of related commits:

**[locator] adding InstanceLocatorInit**

This commit adds a new class `InstanceLocatorInit` which is
similar to `InstanceLoator` but provides a default  constructor
(instead of a parameterized one) and allows an inheriting class to
initialize the object (set the OpenThread Instance) post constructor
call using the `Init()` method. This class is intended for types that
require a default constructor and cannot use a parameterized one. An
example of such a class is one that is used as type of a C array
element (e.g., `Neighbor`/`Child`/`Router` classes which are used as
array in `ChildTable`/`RouterTable`). The inheriting class from
`InstanceLocatorInit` should ensure that object is properly
initialized after the object is created and more importantly that it
is re-initialized when/if it is ever cleared or reset.

**[topology] have Neighbor/Child/Router inherit from InstanceLocatorInit**

This commit changes the `Neighbor` class (and therefore its sub-classes
`Child` and `Router`) to inherit from `InstanceLoatorInit`. It adds
`Init()` method to them as well and ensures that they are initialized
before use from `ChildTable`, `RouterTable` and `Mle` classes.

**[child] remove instance parameter from Ip6Address related methods**

Remove the now unnecessary `aInstance` parameter in `Child` methods
(since `Child` is now itself an `InstanceLocator`).

**[link-quality] have LinkQualityInfo inherit from IntsanceLocatorInit**

**[link-quality] remove aNoiseFloor param from LinkQualityInfo methods**

### DIFF
--- a/src/core/thread/child_table.cpp
+++ b/src/core/thread/child_table.cpp
@@ -102,7 +102,11 @@ ChildTable::ChildTable(Instance &aInstance)
     : InstanceLocator(aInstance)
     , mMaxChildrenAllowed(kMaxChildren)
 {
-    Clear();
+    for (Child *child = &mChildren[0]; child < OT_ARRAY_END(mChildren); child++)
+    {
+        child->Init(aInstance);
+        child->Clear();
+    }
 }
 
 void ChildTable::Clear(void)
--- a/src/core/common/locator.hpp
+++ b/src/core/common/locator.hpp
@@ -71,6 +71,8 @@ extern uint64_t gInstanceRaw[];
  */
 class InstanceLocator
 {
+    friend class InstanceLocatorInit;
+
 public:
     /**
      * This method returns a reference to the parent OpenThread Instance.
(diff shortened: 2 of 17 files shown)


src/core/thread/child_table.cpp
```diff
@@ -102,7 +102,11 @@ ChildTable::ChildTable(Instance &aInstance)
     : InstanceLocator(aInstance)
     , mMaxChildrenAllowed(kMaxChildren)
 {
-    Clear();
+    for (Child *child = &mChildren[0]; child < OT_ARRAY_END(mChildren); child++)
+    {
+        child->Init(aInstance);
+        child->Clear();
```
jwhui: Does this actually work when `OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE=1`? The current implementation of `Child::Clear()` simply calls `memset()`.
https://github.com/openthread/openthread/blob/0823114dfff39dd7d27eeaf1ba4ac3d16002ba86/src/core/thread/topology.cpp#L106-L110

PR #4572, opened 2020-02-18
[mle] invalidate parent if not child

bukepo:
Current Thread certification needs Parent information even after the
node already become a router or leader. This PR sets parent state to
invalid to indicate information in mParent should not be used in normal
cases.


librasungirl:
it should only for Router upgraded from child, no need for Leader or Child device to perform the `if` here.
How about invalidating parent when received a/as rsp? https://github.com/openthread/openthread/blob/master/src/core/thread/mle_router.cpp#L3992?

PR #4573, opened 2020-02-18
[posix-host] set network interface name

bukepo:
When platform netif is enabled, the TUN device's name should be able to
set from command. This PR use the existing `-I` to allow user provide a
wanted name.


src/posix/platform/netif.cpp
```diff
@@ -406,7 +406,17 @@ void platformNetifInit(otInstance *aInstance)
...
+
+    if (aInterfaceName)
+    {
+        VerifyOrDie(strlen(aInterfaceName) < IFNAMSIZ, OT_EXIT_INVALID_ARGUMENTS);
+
+        strncpy(ifr.ifr_name, aInterfaceName, IFNAMSIZ);
+    }
+    else
+    {
+        strncpy(ifr.ifr_name, "wpan%d", IFNAMSIZ);
```
abtink: I think with `strncpy` the `ifr_name` would be exactly `"wpan%d"` (with `%d` in the string). 
Is the `%d` replaced somewhere else with some value?

bukepo: Kernel will replace %d.

wgtdkp: It looks cool that the kernel can automatically suggest an available PAN interface. On the other hand, it complicated the posix platform code that there are multiple source of truth of the TUN interface name `gNetifName`:
1. `otPlatformConfig.mInterfaceName` if it's not `NULL`
2. The TUN interface suggested by the `ioctl` kernel API when initializing the `netif.cpp` module

While there are other modules depending on the TUN interface name and we need to maintain the order carefully to make sure the TUN interface is always initialized before other those modules. It also causes problem when `OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE` is disabled - for example, we need to set the TUN interface name separately in `Udp::Init()` which is adding the 3rd SoT of the TUN interface name.

I suggest we remove this pattern and always use the `otPlatformConfig.mInterfaceName`. For platforms where the default interface name is already used, we are already allowing set a different name via command line options.

@bukepo

wgtdkp: We may want to use a more Thread-specific interface name like `thread-wpan`. For example, Android bluetooth uses the fixed name `bt-pan` (https://android.googlesource.com/platform/packages/modules/Bluetooth/+/refs/heads/master/system/btif/include/btif_pan_internal.h#39).

bukepo: I don't think there are multiple source of truth. The `gNetifName` is always the single source of truth.

On Linux platform, we can customize the tun interface name, but I don't think this is supported on macOS. That's the reason we added the command and API to get the finalized network interface name.

see https://github.com/openthread/openthread/blob/main/src/posix/main.c#L361

wgtdkp: > I don't think there are multiple source of truth. The gNetifName is always the single source of truth.

`gNetifName` is set from `otPlatformConfig.mInterfaceName` or by kernel, so it's not the source.

> On Linux platform, we can customize the tun interface name, but I don't think this is supported on macOS.

Okay, that could be a reason to keep it.

PR #4574, opened 2020-02-18
[posix-host] explicitly ignore return value on write

dismirlian:
This PR fixes an "Ignored return value" warning by effectively marking the return value as ignored on purpose.

However, I think that the return value should be checked and dealt with. There are 2 instances where this happens:

https://github.com/openthread/openthread/blob/fbada4d21845e1d725079d8c003333c99894f20a/src/posix/client.cpp#L166-L169

https://github.com/openthread/openthread/blob/fbada4d21845e1d725079d8c003333c99894f20a/src/posix/client.cpp#L263-L267

### DIFF
--- a/src/posix/client.cpp
+++ b/src/posix/client.cpp
@@ -164,7 +164,7 @@ static void SendBlockingCommand(int aArgc, char *aArgv[])
         ssize_t rval = read(sSessionFd, buffer, sizeof(buffer));
 
         VerifyOrExit(rval >= 0);
-        write(STDOUT_FILENO, buffer, static_cast<size_t>(rval));
+        IgnoreReturnValue(write(STDOUT_FILENO, buffer, static_cast<size_t>(rval)));
         for (ssize_t i = 0; i < rval; i++)
         {
             if (FindDone(&doneState, buffer[i]) || FindError(&errorState, buffer[i]))


src/posix/client.cpp
```diff
@@ -164,7 +164,7 @@ static void SendBlockingCommand(int aArgc, char *aArgv[])
         ssize_t rval = read(sSessionFd, buffer, sizeof(buffer));
 
         VerifyOrExit(rval >= 0);
-        write(STDOUT_FILENO, buffer, static_cast<size_t>(rval));
+        IgnoreReturnValue(write(STDOUT_FILENO, buffer, static_cast<size_t>(rval)));
```
jwhui: @dismirlian, thanks for addressing this (again). This seems to be a follow-on to #4480.

Note that the need to properly handle `write()` return values was previously noted in https://github.com/openthread/openthread/pull/4480#discussion_r367690447

And as noted, we should have a separate PR for addressing  those issues.

dismirlian: Hi @jwhui, yes in fact I wanted to re-surface this issue. If you want, I could prepare a PR addressing this.

jwhui: @dismirlian, that would be great!

dismirlian: @jwhui please see #4580

PR #4431, opened 2019-12-27
[posix-host] add SPI interface support

zhanglongxia:

PR #4578, opened 2020-02-19
[spi] add parameter to reset gpio device for spi

Irving-cl:
This PR is related to #4431 
Currently some GPIO is exported in some scripts to do the hardware reset. However, to use SPI, these GPIOs cannot be exported. So a parameter is added to the command to do the reset so that GPIOs don't need to be exported.

### DIFF
--- a/src/posix/platform/openthread-system.h
+++ b/src/posix/platform/openthread-system.h
@@ -114,7 +114,7 @@ typedef struct otPlatformConfig
     bool        mResetRadio;            ///< Whether to reset RCP when initializing.
     bool        mRestoreDatasetFromNcp; ///< Whether to retrieve dataset from NCP and save to file.
 
-    bool     mSpiGpioReset;       ///< Whether to reset GPIO device only.
+    bool     mSpiGpioReset;       ///< Whether to use Reset Pin to reset RCP only.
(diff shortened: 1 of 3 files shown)


Irving-cl:
@bukepo @zhanglongxia

src/posix/platform/openthread-system.h
```diff
@@ -114,6 +114,7 @@ typedef struct otPlatformConfig
     bool        mResetRadio;            ///< Whether to reset RCP when initializing.
     bool        mRestoreDatasetFromNcp; ///< Whether to retrieve dataset from NCP and save to file.
 
+    bool     mSpiGpioReset;       ///< Whether to reset GPIO device only.
```
zhanglongxia: nit: ///< Whether to use Reset Pin to reset RCP chip only.

src/posix/platform/spi_interface.cpp
```diff
@@ -107,6 +107,12 @@ otError SpiInterface::Init(const otPlatformConfig &aPlatformConfig)
     // Reset RCP chip.
     TrigerReset();
 
+    // Exit if only to reset GPIO device.
```
zhanglongxia: nit: Exit if only to reset RCP chip.

Irving-cl:
Discussed with @xyk, this seems to be unnecessary.

PR #4583, opened 2020-02-20
[network-data] avoid buffer overrun in HandleCommissioningSet()

jwhui:
Reject Commissioner Dataset TLVs with extended TLV length.

### DIFF
--- a/src/core/thread/network_data_leader_ftd.cpp
+++ b/src/core/thread/network_data_leader_ftd.cpp
@@ -206,7 +206,7 @@ void Leader::HandleCommissioningSet(Coap::Message &aMessage, const Ip6::MessageI
     {
         MeshCoP::Tlv::Type type;
 
-        VerifyOrExit((cur + 1) <= end && cur->GetNext() <= end);
+        VerifyOrExit(((cur + 1) <= end) && !cur->IsExtended() && (cur->GetNext() <= end));
 
         type = cur->GetType();
 
--- a/tests/fuzz/fuzzer_platform.c
+++ b/tests/fuzz/fuzzer_platform.c
@@ -26,6 +26,8 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "openthread-core-config.h"
+
 #include <string.h>
 
 #include <openthread/platform/alarm-micro.h>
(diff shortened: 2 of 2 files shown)


src/core/thread/network_data_leader_ftd.cpp
```diff
@@ -206,7 +206,7 @@ void Leader::HandleCommissioningSet(Coap::Message &aMessage, const Ip6::MessageI
     {
         MeshCoP::Tlv::Type type;
 
-        VerifyOrExit((cur + 1) <= end && cur->GetNext() <= end);
+        VerifyOrExit(((cur + sizeof(MeshCoP::Tlv)) <= end) && !cur->IsExtended() && (cur->GetNext() <= end));
```
wgtdkp: pointer arithmetic already multiplies the size of the object, which means `cur + 1` points to memory `sizeof(MeshCoP::Tlv)` bytes after `cur`  while `cur + sizeof(MeshCoP::Tlv)` points to `sizeof(MeshCoP::Tlv) * sizeof(MeshCoP::Tlv)` bytes after `cur`. I think this is not we want.

abtink: I agree. I think `cur + 1` is correct.

Is the `!cur->IsExtended()` for the following corner case:
- If  `cur + 1` is exactly equal to `end` and the bye sequence happens to make it look like the TLV is extended, `cur->GetNext()` would read beyond the `end` to get the extended length. Is this correct?
(The `cur->GetNext()` should work correctly independent of whether the TLV is regular or extended but the `cur + 1 <= end` would not cover the extended tlv case).

We use same check(`cur + 1 <= end && cur->GetNext() <= end`) in other places and I think may face similar situation.

How about (in a future PR)  adding a new method to `ot::Tlv` class itself that checks if the entire TLV fits correctly up to a given `end` pointer (which should be after start of `Tlv` itself). We can write this to work correctly if tlv is extended or not.
```
bool Tlv::CatFit(const uint8_t *aEnd) const // better name?
{
    const uint8_t *start = reinterpret_cast<const uint8_t *>(this);

    return (start + sizeof(Tlv) <= aEnd) && (!IsExtended() || (start + sizeof(ExtendedTlv) <= aEnd)) &&
           (start + GetSize() <= aEnd);
}
```

Thoughts?

jwhui: > pointer arithmetic already multiplies the size of the object, which means `cur + 1` points to memory `sizeof(MeshCoP::Tlv)` bytes after `cur` while `cur + sizeof(MeshCoP::Tlv)` points to `sizeof(MeshCoP::Tlv) * sizeof(MeshCoP::Tlv)` bytes after `cur`. I don't think it is what we want.

Yes, it should be `cur + 1`.

PR #4589, opened 2020-02-21
[utils] settings: move to object, to be able to have multiple instances

dismirlian:
This PR modifies the implementation of Flash Settings to be able to have multiple instances.

This may be useful for applications that want to store its own settings and retain the capability of wiping the OT settings area and the APP settings area independently.

### DIFF
--- a/examples/platforms/utils/Makefile.am
+++ b/examples/platforms/utils/Makefile.am
@@ -47,7 +47,9 @@ libopenthread_platform_utils_a_SOURCES  = \
     mac_frame.h                           \
     settings.h                            \
     settings_ram.c                        \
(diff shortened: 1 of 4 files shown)


jwhui:
@dismirlian, thank you for this contribution (as always).

In case you haven't notice, #4552 proposes to move the settings flash implementation into the OpenThread library as a build option. Can you take a look at #4552 and see how you might support your use case in that model?

dismirlian:
Hi @jwhui, somehow I missed your PR. I'll check it and come back to you. Thanks!!

dismirlian:
Closing because this implementation is going away (#4552)

Issue #4413, opened 2019-12-17
[code] create src/utils for common utilities


bukepo:
**Is your feature request related to a problem? Please describe.**
create a folder `src/utils` for utilities that can be used by multiple applications. I think spinel, hdlc, the upcoming performance test utilities are examples of this requirement.

**Additional context**
#4117
https://github.com/openthread/openthread/pull/3975#discussion_r301717735

bukepo:
@jwhui  @abtink @simonlingoogle @zhanglongxia Thoughts?

abtink:
Sounds good to me.

jwhui:
SGTM

zhanglongxia:
SGTM

PR #4590, opened 2020-02-24
[build] move hdlc to lib/hdlc

bukepo:
This commit moves hdlc from src/ncp to src/lib/hdlc, src/spinel is also
moved to src/lib/spinel.


bukepo:
This fixes #4413

src/ncp/ncp_base.hpp
```diff
@@ -51,10 +51,10 @@
...
 #include "common/instance.hpp"
 #include "common/tasklet.hpp"
-#include "spinel/spinel.h"
-#include "spinel/spinel_buffer.hpp"
-#include "spinel/spinel_decoder.hpp"
-#include "spinel/spinel_encoder.hpp"
+#include "lib/spinel/spinel.h"
+#include "lib/spinel/spinel_buffer.hpp"
+#include "lib/spinel/spinel_decoder.hpp"
+#include "lib/spinel/spinel_encoder.hpp"
```
abtink: Can we consider having `src/lib` folder in the include path? This way we won't need `lib/` when calculating the header files.

src/lib/hdlc/hdlc.hpp
```diff
@@ -33,13 +33,12 @@
 #ifndef HDLC_HPP_
 #define HDLC_HPP_
 
-#include "openthread-core-config.h"
```
abtink: Can we consider keeping this? (even though I see that we don't use any configs in hdlc implementation would be good for consistency and future use).
We already include the config header in other libs: e.g. `spinel_encoder.hpp` and `ncp_base.hpp`

Thoughts?

src/core/mac/mac_frame.cpp
```diff
@@ -1031,7 +1031,7 @@ void TxFrame::ProcessTransmitAesCcm(const ExtAddress &aExtAddress)
 
 exit:
     return;
-#endif // OPENTHREAD_MTD || OPENTHREAD_FTD
+#endif // OPENTHREAD_RADIO
```
abtink: How about moving this to a different PR or commit? Seems very unrelated to rest :)

src/ncp/ncp_base.hpp
bukepo: I am trying to follow google's c++ style as a second guide if not covered by OpenThread's style guide. I prefer using a full path because it works better if there are duplicate file names inside a project.
https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

src/lib/hdlc/hdlc.hpp
bukepo: I didn't see spinel_encoder.hpp includes the core config file. I intended to remove dependency on core config, which seems to be focusing on `src/core` config. Also this ease third party utilization of these libraries.

abtink: Yes. You are right. I miss-read `"openthread-spinel-config.h"` in `spinel-encoder.hpp`. I forgot you already helped add a spinel specific config. I agree and think not using core config is definitely the right way for all lib/utils modules.

Android.mk
```diff
@@ -401,7 +401,7 @@ LOCAL_SRC_FILES                            := \
     src/ncp/ncp_base_ftd.cpp                  \
     src/ncp/ncp_base_dispatcher.cpp           \
     src/ncp/ncp_uart.cpp                      \
-    src/spinel/spinel_buffer.cpp              \
+    src/lib/spinel/spinel_buffer.cpp          \
```
jwhui: nit: maintain alphabetical order?

src/Makefile.am
```diff
@@ -35,14 +35,14 @@ DIST_SUBDIRS                            = \
     cli                                   \
     ncp                                   \
     posix                                 \
-    spinel                                \
+    lib                                   \
```
jwhui: nit: maintain alphabetical order (here and below)

PR #4596, opened 2020-02-25
[posix-host] fix warnings caused by SPI arguments

bukepo:
* signed/unsigned conversion error
* undefined behavior of embedding macros in string literal

> embedding a directive within macro arguments has undefined behavior

### DIFF
--- a/src/posix/main.c
+++ b/src/posix/main.c
@@ -122,7 +122,6 @@ static const struct option kOptions[] = {{"debug-level", required_argument, NULL
                                          {"ncp-dataset", no_argument, NULL, ARG_RESTORE_NCP_DATASET},
                                          {"time-speed", required_argument, NULL, 's'},
                                          {"verbose", no_argument, NULL, 'v'},
-#if OPENTHREAD_POSIX_RCP_SPI_ENABLE
                                          {"gpio-int-dev", required_argument, NULL, ARG_SPI_GPIO_INT_DEV},
                                          {"gpio-int-line", required_argument, NULL, ARG_SPI_GPIO_INT_LINE},
                                          {"gpio-reset-dev", required_argument, NULL, ARG_SPI_GPIO_RESET_DEV},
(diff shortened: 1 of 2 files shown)


src/posix/main.c
```diff
@@ -150,7 +148,6 @@ static void PrintUsage(const char *aProgramName, FILE *aStream, int aExitCode)
             "        --ncp-dataset             Retrieve and save NCP dataset to file.\n"
             "    -s  --time-speed factor       Time speed up factor.\n"
             "    -v  --verbose                 Also log to stderr.\n"
-#if OPENTHREAD_POSIX_RCP_SPI_ENABLE
```
zhanglongxia: If the macro "OPENTHREAD_POSIX_RCP_SPI_ENABLE" is removed,  posix-host will output SPI help information regardless of whether SPI interface is enabled. Could we print SPI and HDLC information separately, and then use macros to control which information is printed?

src/posix/main.c
```diff
@@ -133,48 +132,62 @@ static const struct option kOptions[] = {{"debug-level", required_argument, NULL
...
+    "        --spi-cs-delay[=usec]     Specify the delay after C̅S̅ assertion, in µsec.\n"
+    "        --spi-reset-delay[=ms]    Specify the delay after R̅E̅S̅E̅T̅ assertion, in milliseconds.\n"
+    "        --spi-align-allowance[=n] Specify the maximum number of 0xFF bytes to clip from start of\n"
+    "                                  MISO frame. Max value is 16.\n"
+    "        --spi-small-packet=[n]    Specify the smallest packet we can receive in a single transaction.\n"
+    "                                  (larger packets will require two transactions). Default value is 32.\n"
+    "    -h  --help                    Display this usage information.\n";
+#else
+static const char *sHelpString =
+    "Syntax:\n"
```
zhanglongxia: The code between line 139 and 149 and the code between line 174 and 184 are the same. Could we define a common help string to represent it? Otherwise, we need to update two places when we want to modify the common help string in the future.

Issue #4542, opened 2020-02-09
[address-resolver] change the age model


abtink:
The issue https://github.com/openthread/openthread/issues/4535 triggered a deeper look into the `AddressResolver` implementation.

First let me say, the existing approach and the fix by @LuDuda for the issue all look good and fine.

I was thinking of a possible future enhancement to simplify the logic and replace the age model. 
Describing it here to get feedback on it: 
- We use a cache entry array as a pool.
- We use single linked list to keep the entries in order we want.
  - For linked list we can use a next index (into array) instead of a next pointer to be more memory efficient.
- We can use 4 different lists: 
  - List of cached entries : We keep the list sorted based on order of use - when an entry is used, we move it to the head of the list.
  - List of query-first-time: First time queries (we wait for the timeout).
  - List of query-retry-mode: Queries that failed at least once and are in retry timeout mode. We keep them in order the user requested them.
  - List of unused entries
- This way there is no need to keep state in entry (the list you are in indicates your state). 
- When trying to allocate a new entry, we first check in the unused list, then query-retry-list and finally the cached list from back of the list (oldest entry).

I can help add it later if the approach sounds good.

In the current design I think we may use entries inefficiently in some cases, for example:
- We query for an address, it grabs the highest age one (last one)
- Suppose it times out (no response), we increase failure count and update retry time (age remains the same).
- Now if we query for another address, the same entry (since it's still highest age) would be used even though we may have unused entries we could use.

Thoughts?

jwhui:
@abtink, thanks for proposing this enhancement.

I think single-linked list using index to array is a good optimization - it avoids needing to update the age value for all affected entries when inserting/removing entries.

I do **not** think we should reclaim entries that are in `kStateQuery`. The current implementation does not allocate new entries that are in `kStateQuery`: https://github.com/openthread/openthread/blob/b62d42f0ef8cba720d2d90a98a88a322380b6a03/src/core/thread/address_resolver.cpp#L138

Note that in a long-running stable network, I think we should assume that there are no "invalid" entries, since we do not actually have a mechanism that ages out entries (which is intentional).

Also, we may want to consider how we treat entries that were created due to "snooping" messages - see my comment in https://github.com/openthread/openthread/pull/4540#discussion_r376828382.

LuDuda:
@abtink I really like your proposal. It would simplify the current implementation a lot! I just followed the current model while bug-fixing.

[...]

PR #4599, opened 2020-02-26
[address-resolver] use linked-list and enable snooped entry timeout

abtink:
This PR updates the `AddressResolver` and how cache entries are managed. This is related to Issue https://github.com/openthread/openthread/issues/4542. It contains the following related commits:

**[linked-list] change PopAfter() parameter to be a pointer**

This commit changes `LinkedList::PopAfter()` method to accept a pointer to a previous entry as a parameter. If the previous entry pointer is NULL the entry at the head of the list is popped. 

**[address-resolver] use linked-list and enable snooped entry timeout**

This commit contains multiple enhancements in `AddressResolver`:

- Cache entries are stored in different lists: Cached list (entries in use), query list (entries actively querying and waiting for address notification response), query-retry list (entries in delay wait mode due to prior query failing to get a response), and snooped list (entries discovered through received message inspection aka snoop optimization). Singly linked list is used which helps keep the entries sorted based on the order of use (whenever a cache entry is used to resolve an address, it is moved to the head of the list). This replaces and simplifies the aging model. 
- This commit updates the cache entry class definition to use `union` for member variables (like timeout or transaction time) that are tied to entry being in different lists. This help reduce the memory requirement for cache table.
- This commit also adds a new mechanism to manage snooped entries. When a new snooped entry is added, we do not allow it to be evicted for a short timeout. This allows some delay for a response message to use the entry (if entry is used it will be moved to the cached list). If a snooped entry is not used after the timeout, we allow it to be evicted. To ensure snooped entries do not overwrite other cached entries, we limit the number of snooped entries that are in timeout mode and cannot be evicted by a (build-time configurable) threshold. 

**[address-resolver] update how the cache table changes are logged**

This commit adds a new helper method `LogCacheEntryChange()` which logs any change (entry added/removed/updated) and its reason in address cache table. 

**[address-resolver] update public OT API for getting cache table**

This commit updates the public OpenThread APIs to get the address cache table. The new APIs allows to iterate through all entries (including in query or retry mode) and provides more info about the entries, e.g., current timeout, retry delay (if entry is in query/retry state), or the mesh-local EID, last transaction time (if entry is in cached state). 

**[spinel/ncp] update address cache table property to include additional info**

This commit expands the `SPINEL_PROP_THREAD_ADDRESS_CACHE_TABLE` property format to include more info about address cache table. This change keeps the definition backward compatible by not changing existing format and only adding new (optional) fields. 

**[toranj] remove unused age property when parsing cache table**

-----------------------
This PR is related to two other PRs. The change is broken due to dependency we have between new spinel property definitions in OT, wpantund changes and then the new `toranj` test-case.
- This PR (https://github.com/openthread/openthread/pull/4599) is the first one (should be merged first). 
- The wpantund PR: https://github.com/openthread/wpantund/pull/462
- The follow up OpenThread PR (which adds the test): https://github.com/openthread/openthread/pull/4600

### DIFF
--- a/include/openthread/thread_ftd.h
+++ b/include/openthread/thread_ftd.h
@@ -79,16 +79,45 @@ typedef struct
 typedef uint16_t otChildIp6AddressIterator; ///< Used to iterate through IPv6 addresses of a Thread Child entry.
 
 /**
+ * This enumeration defines the EID cache entry state.
+ *
+ */
+typedef enum otCacheEntryState
+{
+    OT_CACHE_ENTRY_STATE_CACHED      = 0, // Entry is cached and in-use.
+    OT_CACHE_ENTRY_STATE_SNOOPED     = 1, // Entry is created by snoop optimization (inspection of received msg).
+    OT_CACHE_ENTRY_STATE_QUERY       = 2, // Entry represents an ongoing query for the EID.
+    OT_CACHE_ENTRY_STATE_RETRY_QUERY = 3, // Entry is in retry wait mode (a prior query did not get a response).
+} otCacheEntryState;
+
+/**
  * This structure represents an EID cache entry.
  *
  */
-typedef struct otEidCacheEntry
+typedef struct otCacheEntryInfo
 {
-    otIp6Address   mTarget;    ///< Target
-    otShortAddress mRloc16;    ///< RLOC16
-    uint8_t        mAge;       ///< Age (order of use, 0 indicates most recently used entry)
-    bool           mValid : 1; ///< Indicates whether or not the cache entry is valid
-} otEidCacheEntry;
+    otIp6Address      mTarget;             ///< Target EID
+    otShortAddress    mRloc16;             ///< RLOC16
+    otCacheEntryState mState;              ///< Entry state
+    bool              mCanEvict : 1;       ///< Indicates whether the entry can be evicted.
+    bool              mValidLastTrans : 1; ///< Indicates whether last transaction time and ML-EID are valid.
+    uint32_t          mLastTransTime;      ///< Last transaction time (applicable in cached state).
+    otIp6Address      mMeshLocalEid;       ///< Mesh Local EID (applicable if entry in cached state).
+    uint16_t          mTimeout;            ///< Timeout in seconds (applicable if in snooped/query/retry-query states).
+    uint16_t          mRetryDelay;         ///< Retry delay in seconds (applicable if in query-retry state).
+} otCacheEntryInfo;
+
+/**
+ * This type represents an iterator used for iterating through the EID cache table entries.
+ *
+ * To initialize the iterator and start from the first entry in the cache table, set all its fields in the structure to
+ * zero (e.g., `memset` the iterator to zero).
+ *
+ */
+typedef struct otCacheEntryIterator
+{
+    const void *mData[2]; ///< Opaque data used by the core implementation. Should not be changed by user.
+} otCacheEntryIterator;
 
 /**
  * Get the maximum number of children currently allowed.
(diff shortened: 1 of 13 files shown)


src/core/thread/address_resolver.cpp
```diff
@@ -69,370 +74,471 @@ AddressResolver::AddressResolver(Instance &aInstance)
     Get<Ip6::Icmp>().RegisterHandler(mIcmpHandler);
 }
 
-void AddressResolver::Init(void)
+void AddressResolver::InitCacheEntries(void)
 {
-    memset(&mCache, 0, sizeof(mCache));
+    mUnusedList.Clear();
```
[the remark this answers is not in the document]
abtink: Indeed. Yes. Thanks.

src/core/thread/address_resolver.cpp
```diff
@@ -69,370 +74,471 @@ AddressResolver::AddressResolver(Instance &aInstance)
...
-    aEntry.mState = Cache::kStateInvalid;
-    MoveCacheEntryAtBack(aEntry);
+    LogCacheEntryChange(kEntryUpdated, kReasonSnoop, *entry);
+
+exit:
+    return error;
 }
 
-otError AddressResolver::UpdateCacheEntry(const Ip6::Address &aEid, Mac::ShortAddress aRloc16)
+otError AddressResolver::AddCacheEntry(const Ip6::Address &aEid, Mac::ShortAddress aRloc16)
```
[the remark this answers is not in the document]
abtink: Good idea. Thanks.

abtink:
@LuDuda thanks so much for reviewing this. :)
I know it's a bigger change and not super easy to read and review.

On calling `Init()`, my idea for `CacheEntry::Init()` was to only set the `Instance` on all entries. Technically, we would only need to do it once from constructor. After that entries should always be in one of the linked-lists, so if we remove/evict it from a list, we should immediately push the entry onto a new list. I will try to see if I can add comments in the code on this.

include/openthread/thread_ftd.h
```diff
@@ -507,17 +536,19 @@ uint8_t otThreadGetMaxRouterId(otInstance *aInstance);
...
- * This function gets an EID cache entry.
+ * This function gets the next EID cache entry (using an iterator).
  *
- * @param[in]   aInstance A pointer to an OpenThread instance.
- * @param[in]   aIndex    An index into the EID cache table.
- * @param[out]  aEntry    A pointer to where the EID information is placed.
+ * @param[in]    aInstance   A pointer to an OpenThread instance.
+ * @param[out]   aEntryInfo  A pointer to where the EID cache entry information is placed.
+ * @param[inout] aIterator   A pointer to an iterator. It will be updated to point to next entry on success. To get the
+ *                           first entry initialize the iterator by setting all its fields to zero (e.g., `memset` the
```
jwhui:
```suggestion
 *                           first entry, initialize the iterator by setting all its fields to zero (e.g., `memset` the
```

src/core/thread/address_resolver.hpp
```diff
@@ -75,24 +88,26 @@ class AddressResolver : public InstanceLocator
     void Clear(void);
 
     /**
-     * This method gets an EID cache entry.
+     * This methods gets the information about the next EID cache entry (using an iterator).
```
jwhui:
```suggestion
     * This method gets the information about the next EID cache entry (using an iterator).
```

src/core/thread/address_resolver.hpp
```diff
@@ -75,24 +88,26 @@ class AddressResolver : public InstanceLocator
...
      *
-     * @param[in]   aIndex  An index into the EID cache table.
-     * @param[out]  aEntry  A reference to where the EID information is placed.
+     * @param[out]   aInfo       An `EntryInfo` where the EID cache entry information is placed.
+     * @param[inout] aIterator   An iterator. It will be updated to point to the next entry on success.
+     *                           To get the first entry initialize the iterator by setting all its fields to zero (e.g.,
```
jwhui:
```suggestion
     *                           To get the first entry, initialize the iterator by setting all its fields to zero (e.g.,
```

abtink:
Rebased and updated this PR and its related one https://github.com/openthread/openthread/pull/4600 adding the related test.

jwhui:
@abtink, thanks for the update. It seems many of the comments are not addressed, can you double check that you pushed your latest changes?

abtink:
Thanks @jwhui. Yes, made a mistake and pushed the wrong branch. Pushed the right one now.

src/core/config/tmf.h
```diff
@@ -45,6 +45,17 @@
...
 
+/**
+ * @def OPENTHREAD_CONFIG_TMF_MAX_SNOOP_ADDRESS_CACHE_ENTRIES
+ *
+ * The maximum number of EID-to-RLOC cache entries that can be used for "snoop optimization" where an entry is created
+ * by inspecting a received message.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_TMF_MAX_SNOOP_ADDRESS_CACHE_ENTRIES
+#define OPENTHREAD_CONFIG_TMF_MAX_SNOOP_ADDRESS_CACHE_ENTRIES 2
```
LuDuda: Just spotted one more nit (can be ignored):
 ```suggestion
#define OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_MAX_SNOOP_ENTRIES 2
```

LuDuda: If yes.. then also change the order to define `OPENTHREAD_CONFIG_TMF_SNOOP_CACHE_ENTRY_TIMEOUT`.

abtink: Great suggestion @LuDuda. Thanks.

PR #4602, opened 2020-02-27
[network-data] add source address of SVR_DATA.ntf check

simonlingoogle:
This PR makes sure that SVR_DATA.ntf messages are using RLOC source addresses according to Spec. 5.15.6.1. Otherwise `RegisterNetworkData` might be called with wrong RLOC16.

### DIFF
--- a/src/core/thread/network_data_leader_ftd.cpp
+++ b/src/core/thread/network_data_leader_ftd.cpp
@@ -152,6 +152,8 @@ void Leader::HandleServerData(Coap::Message &aMessage, const Ip6::MessageInfo &a
 
     otLogInfoNetData("Received network data registration");
 
+    VerifyOrExit(aMessageInfo.GetPeerAddr().IsRoutingLocator());
+
     if (ThreadTlv::GetTlv(aMessage, ThreadTlv::kRloc16, sizeof(rloc16), rloc16) == OT_ERROR_NONE)
     {
         VerifyOrExit(rloc16.IsValid());


src/core/thread/network_data_leader_ftd.cpp
```diff
@@ -152,6 +152,8 @@ void Leader::HandleServerData(Coap::Message &aMessage, const Ip6::MessageInfo &a
 
     otLogInfoNetData("Received network data registration");
 
+    VerifyOrExit(aMessageInfo.GetPeerAddr().IsRoutingLocator());
```
librasungirl: One concern is that if putting the check here, in case there is SRV_DATA.ntf with non RLOC source, there would be no ACK for the CON Coap message, thus there might be meaningless CoAP retransmission.
In case there is some coincidence which you cares, maybe move the check before trying to retrieve the short address from source?

jwhui:
@simonlingoogle, thank you for this submission.

Couple questions:

1. Are we seeing SVR_DATA.ntf messages that do not have RLOC as IPv6 source address in practice?

1. Even if this does occur, would the recent changes in #4485 eventually recover this situation?

/cc/ @librasungirl

src/core/thread/network_data_leader_ftd.cpp
simonlingoogle: There are two concerns regarding this PR:
1. Should we check before `RegisterNetworkData`?
2. Should we send empty ACK?

For 1), I think we should not check before `RegisterNetworkData` because the caller would thought that the request is fully processed however it is not. 

For 2), I think we should not reply with `ACK 2.04 Changed`, but reply with `ACK 4.00 Bad Request`. However current OT codebase rarely does that for invalid COAP requests, but only exit skip quietly. 

Thoughts?

librasungirl:
@jwhui 
1) In general I think it won't happen. It is just observed when there is some fake SRV_DATA.ntf constructed on OTBR host, as wpantund interface only have MLEID added, the linux would select it as source.
2) Yes,  agree that #4485 should be able to recover if there is some coincidence. 

I am fine to add RLOC source check before deriving RLOC16 from the source and using it.

simonlingoogle:
> * Are we seeing SVR_DATA.ntf messages that do not have RLOC as IPv6 source address in practice?

Yes it happened in certification as @librasungirl have described.  

It would happen in production only if there is bug in the implementation. I think it is generally better to validate requests according to Spec, especially if the request could cause logical errors.

src/core/thread/network_data_leader_ftd.cpp
jwhui: > There are two concerns regarding this PR:
> 
> 1. Should we check before `RegisterNetworkData`?
> 2. Should we send empty ACK?
> 
> For 1), I think we should not check before `RegisterNetworkData` because the caller would thought that the request is fully processed however it is not.

I think the current location is fine. As noted, sending with non-RLOC IPv6 Source Address is unexpected behavior and we should catch that as early as possible.

> For 2), I think we should not reply with `ACK 2.04 Changed`, but reply with `ACK 4.00 Bad Request`. However current OT codebase rarely does that for invalid COAP requests, but only skip quietly.

Yes, the existing codebase generally does not respond with Bad Request for TMF messages. We can consider changing that in a separate PR if we find that useful.

simonlingoogle: I think always sending ACK for CON is helpful to avoid meaningless CoAP retransmission as @librasungirl said.

librasungirl: It makes sense to catch it earlier if invalid.

Generally SRV_DATA.ntf in Thread-conformance implementation should have the right RLOC source. And the issue we observed so far only happens in the special test scenario which constructs fake SRV_DATA.ntf and sends it over the wpantund which filters out RLOC address. we may keep it here and change to support Bad Request to avoid the impact of CoAP retransmission when required in the future? @simonlingoogle

simonlingoogle: > we may keep it here and change to support Bad Request to avoid the impact of CoAP retransmission when required in the future? @simonlingoogle

Agree.

PR #4603, opened 2020-02-27
[travis] move package check into lint stage

bukepo:
This PR moves the package check into the first stage so that if there
are any cmake build errors jobs in stage 2 will be skipped.

### DIFF
--- a/script/test
+++ b/script/test
@@ -113,10 +113,6 @@ EXAMPLES:
 }
 
 do_package() {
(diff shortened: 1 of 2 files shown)


script/test
```diff
@@ -113,6 +113,10 @@ EXAMPLES:
 }
 
 do_package() {
+    if ! cmake --version | grep -q '(?<=cmake version \(\d+.+d)'; then
```
jwhui:
```
grep: Unmatched ( or \(
```

PR #4604, opened 2020-02-27
[build] enhancements for OpenThread border router

bukepo:
This PR makes a few small changes so that the cmake build files can be
used by project depending on this project.


third_party/mbedtls/CMakeLists.txt
```diff
@@ -28,14 +28,23 @@
...
-    ${OT_PRIVATE_INCLUDES}
-    ${PROJECT_SOURCE_DIR}/src/core
+target_include_directories(mbedcrypto
+    PUBLIC
+        ${PROJECT_SOURCE_DIR}/third_party/mbedtls
+        ${PROJECT_SOURCE_DIR}/third_party/mbedtls/repo/include
+    PRIVATE
+        ${OT_PUBLIC_INCLUDES}
+        ${OT_PRIVATE_INCLUDES}
+        ${PROJECT_SOURCE_DIR}/src/core
```
wgtdkp: The compiling option at here https://github.com/openthread/openthread/pull/4604/files#diff-13df1eb8811b9d6002a25bf0bbddd046R35 is public, it adds the macro `MBEDTLS_CONFIG_FILE` to a target that is depending on `mbedcrypto`. But it should fail because `mbedcrypto`  privately adds `${PROJECT_SOURCE_DIR}/src/core` which includes `openthread-core-config.h` (included by `MBEDTLS_CONFIG_FILE`). It works for now because all targets depending on `mbedcrypto` has already manually added `${PROJECT_SOURCE_DIR}/src/core`.

But, to work with modern CMake, I think this needs to be public.

src/ncp/CMakeLists.txt
```diff
@@ -94,6 +94,23 @@ target_sources(openthread-ncp-ftd PRIVATE ${COMMON_SOURCES})
 target_sources(openthread-ncp-mtd PRIVATE ${COMMON_SOURCES})
 target_sources(openthread-rcp PRIVATE ${COMMON_SOURCES})
 
-target_link_libraries(openthread-ncp-ftd openthread-spinel-ncp)
-target_link_libraries(openthread-ncp-mtd openthread-spinel-ncp)
-target_link_libraries(openthread-rcp openthread-spinel-rcp)
+target_link_libraries(openthread-ncp-ftd
+    PRIVATE
```
wgtdkp: follow the pattern that `public` comes first?

.travis/script.sh
```diff
@@ -42,7 +42,7 @@ python --version || die
     export CPPFLAGS="-DMBEDTLS_DEBUG_C"
     export CPPFLAGS="${CPPFLAGS} -I${TRAVIS_BUILD_DIR}/third_party/mbedtls"
     export CPPFLAGS="${CPPFLAGS} -I${TRAVIS_BUILD_DIR}/third_party/mbedtls/repo/include"
-    export CPPFLAGS="${CPPFLAGS} -DMBEDTLS_CONFIG_FILE=\\\"mbedtls-config.h\\\""
+    export CPPFLAGS="${CPPFLAGS} -DMBEDTLS_CONFIG_FILE=\\\"ot-mbedtls-default-config.h\\\""
```
jwhui: Can you provide some background for the file name change?

bukepo: mbedtls-config.h is too common as a default header file provided by OpenThread. By adding a prefix, users may use the common name when they provide their own mbedtls config file. Besides, the one we provided depends on core config so it cannot be a public header which is included by mbedtls's public headers.

jwhui: > mbedtls-config.h is too common as a default header file provided by OpenThread. By adding a prefix, users may use the common name when they provide their own mbedtls config file.

In the case of a user providing their own mbedtls config, we should remove `third_party/mbedtls` from the include path.

> Besides, the one we provided depends on core config so it cannot be a public header which is included by mbedtls's public headers.

Yes, `third_party/mbedtls/mbedtls-config.h` should only be used if mbedTLS is being built only for OpenThread.

bukepo: reverted.

third_party/mbedtls/CMakeLists.txt
```diff
@@ -75,6 +92,7 @@ target_sources(mbedcrypto PRIVATE
     repo/library/ssl_ticket.c
     repo/library/ssl_tls.c
     repo/library/threading.c
+    repo/library/timing.c
```
jwhui: What is the use case for including `timing.c`?

bukepo: otbr-commissioner uses the timer API in mbedtls. Adding this without defining `TIMING_C` in mbedtls-config.h will not affect OpenThread.

PR #4607, opened 2020-02-27
[posix] fix handling SPI interrupts

bukepo:
When SPI is already interrupted before select(), the wait timeout will
be set to 0, and select() will return 0 immediately if there are no new
events. As a result, cause an incorrect OT_ERROR_RESPONSE_TIMEOUT error.

This PR fixes this issue by check if data is ready before checking the
select() result.

### DIFF
--- a/src/posix/platform/spi_interface.cpp
+++ b/src/posix/platform/spi_interface.cpp
@@ -97,6 +97,9 @@ otError SpiInterface::Init(const otPlatformConfig &aPlatformConfig)
     {
         // If the interrupt pin is not set, SPI interface will use polling mode.
         InitIntPin(aPlatformConfig.mSpiGpioIntDevice, aPlatformConfig.mSpiGpioIntLine);
+    }
+    else
+    {
         otLogNotePlat("SPI interface enters polling mode.");
     }
 
@@ -656,12 +659,13 @@ otError SpiInterface::WaitForFrame(const struct timeval &aTimeout)
     struct timeval timeout = {kSecPerDay, 0};
     fd_set         readFdSet;
     int            ret;
+    bool           isDataReady = false;
 
     FD_ZERO(&readFdSet);
 
     if (mIntGpioValueFd >= 0)
     {
-        if (CheckInterrupt())
+        if ((isDataReady = CheckInterrupt()))
         {
             // Interrupt pin is asserted, set the timeout to be 0.
             timeout.tv_sec  = 0;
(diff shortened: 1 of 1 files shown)


src/posix/platform/spi_interface.cpp
```diff
@@ -695,14 +699,13 @@ otError SpiInterface::WaitForFrame(const struct timeval &aTimeout)
...
+            isDataReady = CheckInterrupt();
         }
+    }
 
-        // If we can receive a packet.
-        if (CheckInterrupt())
-        {
-            otLogDebgPlat("WaitForFrame(): Interrupt.");
-            PushPullSpi();
-        }
```
zhanglongxia: How about update the code like this? Then we don't need to define new variable and the code logic is easier.
```
if (ret >= 0) 
{ 
    if ((ret == 0) && ((timeout.tv_sec != 0) || (timeout.tv_usec != 0)))
    {
        ExitNow(error = OT_ERROR_RESPONSE_TIMEOUT);
    }
  
    if (FD_ISSET(mIntGpioValueFd, &readFdSet))
    {     
        struct gpioevent_data event;
    
        // Read event data to clear interrupt.
        VerifyOrDie(read(mIntGpioValueFd, &event, sizeof(event)) != -1, OT_EXIT_FAILURE);
    }     
              
    // If we can receive a packet.  
    if (CheckInterrupt())
    {     
        otLogDebgPlat("WaitForFrame(): Interrupt.");
        PushPullSpi();
    }     
}      
else if (errno != EINTR)
{             
    DieNow(OT_EXIT_ERROR_ERRNO);
}
```

src/posix/platform/spi_interface.cpp
```diff
@@ -656,12 +659,13 @@ otError SpiInterface::WaitForFrame(const struct timeval &aTimeout)
...
     fd_set         readFdSet;
     int            ret;
+    bool           isDataReady = false;
 
     FD_ZERO(&readFdSet);
 
     if (mIntGpioValueFd >= 0)
     {
-        if (CheckInterrupt())
+        if ((isDataReady = CheckInterrupt()))
```
abtink: Can we consider to do push-pull here and exit and skip the `select()` fully?
```
        if (CheckInterrupt())
        {
             // Clear the interrupt ...(read the fd to clear it)...
             PushPullSpi();
             ExitNow();
        }
 ```

Issue #4598, opened 2020-02-26
[cli] Not possible to set a Network Name using spaces


edmont:
The Network Name is defined in Thread as:

> A human-readable utf-8 string to identify the network over-the-air, similar to the Service Set Identifier (SSID) in Wi-Fi. The maximum length of this field is 16 bytes.

However the space character (UTF-8 `0020`) is not possible to be assigned to a Network Name through the [CLI](https://github.com/openthread/openthread/blob/thread-reference-20191113/src/cli/cli.cpp#L1822) since it's used as argument separator.

A possible solution would be to require quotation when entering the Network Name through the CLI.

I found this when trying to reproduce the proposed derivation of PSKc in the Thread 1.1 Specification (8.4.1.2.2), in which "Test Network" is given as example of Network Name.

jwhui:
@edmont, thank you for raising this issue.

I agree it would be nice to support space characters in the Network Name. Would you like to submit a PR with the proposed change?

simonlingoogle:
This is an issue I am aware of recently. 
I think there are generally 3 options:

1. Use quotation for CLI arguments that contain whitespaces
    - e.x. `networkname "hello world"`
2. Use windows style `\<space>` to escape whitespaces in commands
    - e.x. `networkname hello\ world`
3. Add new command that receives a hex network name
    - e.x. `networknamex 68656c6c6f20776f726c64`

Option 1 and 2 would general solutions for all commands, meanwhile option 3 only handles commands like `networkname`, `domainname` specifically.

Thoughts? @librasungirl @jwhui

edmont:
I would say the first option is cleaner for the user.

But notice that values of Network Name containing quotes should be allowed as well, so the logic would be: a single string argument is the one between the two farthermost quote characters. This disallows commands taking two consecutive UTF-8 string arguments, which I don't think are needed now anyway.

In that sense, option 2 is more versatile.

edmont:
Fixed by https://github.com/openthread/openthread/pull/4608.

Issue #4597, opened 2020-02-26
[meshcop] Commissioner Credential should be used to generate PSKc


edmont:
Thread 1.1 Specification section 8.4.1.2 defines how a passphrase (the Commissioner Credential) is used to derive a key (PSKc) that is used to establish a DTLS session between the Commissioner and the Border Agent.

This Commissioner Credential is defined as (8.2):

> A human-scaled passphrase for use in authenticating that a device may petition to become the Commissioner of the network. This passphrase is encoded in utf-8 format and has a length of six (6) bytes minimum and 255 bytes maximum. This credential can be thought of as the network administrator password for a Thread Network.
> The first device in a network, typically the initial Leader, MUST be out-of-band commissioned to inject the correct user generated Commissioning Credential into the Thread Network, or provide a known default Commissioning Credential to be changed later. This credential is used to derive an enhanced key using key stretching called the PSKc (Pre-Shared Key for the Commissioner) which is used to establish the Commissioner Session.

At the moment, I believe OT (API + CLI + docs) is failing to provide this functionality since only the PSKc can be configured when modifying the datasets.

A [tool](https://openthread.io/guides/border-router/tools#pskc_generator) is provided in the OT-BR project to derive the PSKc from the Commissioner Credential, but it's not even mentioned in the OpenThread documentation. And even the Commissioner Credential [is unknown](https://github.com/openthread/openthread/pull/3769) until a _handmade_ PSKc is configured.

In my opinion OpenThread should have an easy way to setup a Thread network with a given Commissioner Credential that can be used later in a Commissioner UI, and it should probably have a default Commissioner Credential in case the user doesn't provide one. At the moment, being able to handle just the PSKc generates a lot of confusion among developers about the terminology, I think. Users shouldn't need to see or use the PSKc at any moment.

jwhui:
@edmont, thank you for raising this issue.

I agree we could improve on the use of Commissioner Credential when setting up the network. Perhaps we could enhance the "dataset" CLI command to support generation of datasets with the a Commissioner Passphrase, rather than a random PSKc.

OpenThread already has an API for generating the PSKc [otCommissionerGeneratePskc()](https://github.com/openthread/openthread/blob/91b0dba2a0e212adc9fa2a02860761ec3bada999/include/openthread/commissioner.h#L419), but that is only included when the commissioner feature is enabled.

Of course, in terms of getting/reading datasets, we have no option but to provide the PSKc itself.

[...]

Issue #4327, opened 2019-11-13
Thread 1.2.0 Core Implementation


jwhui:
This issue tracks Thread 1.2.0 core feature implementation.

- [x] Thread 1.2 feature flag (#4325)
- [x] Parent Selection rules (#4423)
- [x] Enhanced Keep Alive (#4325)
- [x] Enhanced Frame Pending (#4334)
- [x] Ack Frame Generator Functions (#4884)
- [x] Move Transmit AES-CCM* processing to sub-mac (#4919)
- [x] Link Metrics
  - [x] Single Probe (#5481)
  - [x] Forward Tracking Series (#5608)
  - [x] Enhanced ACK Probing (#5780)
- [x] Coordinated Sampled Listening (#4557)
- [x] Backbone Router service (#4430)
- [x] All Network Backbone Routers Multicast Address (#4755)
- [x] Domain Name (#4815)
- [x] Domain Unicast Address Registration
  - [x] Domain Prefix (#4793)
  - [x] All Domain Backbone Routers Multicast Address (#4795)
  - [x] Domain Unicast Address generation (#4854)
  - [x] Prioritize DUA in Address Registration TLV (#4862)
  - [x] Allow manual DUA IID configuration (#4900)
  - [x] Store DUA DAD counter in non-volatile (#4904)
  - [x] Register Domain Unicast Addresses (#5276)
- [x] Multicast Address Registration
  - [x] MED Multicast Registration to Parent (#4837)
  - [x] Register local multicast addresses (#5248)
  - [x] Register child multicast addresses (#5236)
  - [x] Backbone Router multicast listeners table (#5292)
  - [x] Add multicast forwarding for MLR Listeners (https://github.com/openthread/ot-br-posix/pull/532)
  - [x] Add Commissioner API to register Multicast Listeners (#5346)
  - [x] Configure MLR.req response status (#5350)
  - [x] Send BMLR.ntf on Backbone link (#5388)
- [x] Backbone TMF Agent (#5471)

zhifeng1756:
Does Openthread supports Commissioning and configuration via BLE as stated in Thread [v1.2](https://www.threadgroup.org/Portals/0/documents/support/ThreadInCommercialWhitePaper_2542_1.pdf)? If not yet, any plan?

Thanks,

jwhui:
[...]

PR #4608, opened 2020-02-27
[cli] allow to escape separators

edmont:
Fixing #4598

Now it's allowed to escape the space character `' '` using backslash `'\'`.

For instance:
```
> networkname Test\ Network                                                     
Done                                                                            
> networkname                                                                   
Test Network                                                                    
Done                                                                            
>
```

### DIFF
--- a/src/cli/README.md
+++ b/src/cli/README.md
@@ -5,6 +5,27 @@ command line interface. Use the CLI to play with OpenThread, which
 can also be used with additional application code. The
 OpenThread test scripts use the CLI to execute test cases.
 
+## Separator and escaping characters
+
+The whitespace character (`' '`) is used to delimitate the command name
+and the different arguments, toghether with tab (`'\t'`) and new line
+characters (`'\r'`, `'\n'`).
+
+Some arguments might require to accept whitespaces on them. For those
+cases the backslash character (`'\'`) can be used to escape whitespace
+or the backslash itself.
+
+Example:
+
+```bash
+> networkname Test\ Network
+Done
+> networkname
+Test Network
+Done
+>
+```
+
 ## OpenThread Command List
 
 * [bufferinfo](#bufferinfo)
@@ -718,7 +739,7 @@ Done
 
 ### logfilename \<filename\>
 
-- Note: POSIX Platform Only, ie: `OPENTHREAD_EXAMPLES_POSIX`
+- Note: Simulation Only, ie: `OPENTHREAD_EXAMPLES_SIMULATION`
 - Requires `OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_DEBUG_UART`
 
 Specifies filename to capture otPlatLog() messages, useful when
--- a/src/cli/README.md
+++ b/src/cli/README.md
@@ -7,12 +7,12 @@ OpenThread test scripts use the CLI to execute test cases.
 
 ## Separator and escaping characters
 
-The whitespace character (`' '`) is used to delimitate the command name
+The whitespace character (`' '`) is used to delimit the command name
 and the different arguments, together with tab (`'\t'`) and new line
 characters (`'\r'`, `'\n'`).
 
 Some arguments might require to accept whitespaces on them. For those
-cases the backslash character (`'\'`) can be used to escape whitespace
+cases the backslash character (`'\'`) can be used to escape separators
 or the backslash itself.
 
 Example:
--- a/src/cli/README.md
+++ b/src/cli/README.md
@@ -8,7 +8,7 @@ OpenThread test scripts use the CLI to execute test cases.
 ## Separator and escaping characters
 
 The whitespace character (`' '`) is used to delimitate the command name
-and the different arguments, toghether with tab (`'\t'`) and new line
+and the different arguments, together with tab (`'\t'`) and new line
 characters (`'\r'`, `'\n'`).
 
 Some arguments might require to accept whitespaces on them. For those
--- a/src/core/utils/parse_cmdline.cpp
+++ b/src/core/utils/parse_cmdline.cpp
@@ -45,6 +45,11 @@ static bool IsSpaceOrNewLine(char aChar)
     return (aChar == ' ') || (aChar == '\t') || (aChar == '\r') || (aChar == '\n');
 }
 
+static bool IsEscapable(char aChar)
+{
+    return (aChar == ' ') || (aChar == '\\');
+}
+
 otError CmdLineParser::ParseCmd(char *aString, uint8_t &aArgc, char *aArgv[], uint8_t aArgcMax)
 {
     otError error = OT_ERROR_NONE;
@@ -54,14 +59,15 @@ otError CmdLineParser::ParseCmd(char *aString, uint8_t &aArgc, char *aArgv[], ui
 
     for (cmd = aString; *cmd; cmd++)
     {
-        if ((*cmd == '\\') && (*(cmd + 1) == ' '))
+        if ((*cmd == '\\') && IsEscapable(*(cmd + 1)))
         {
-            strcpy(cmd, cmd + 1);
+            memmove(cmd, cmd + 1, strlen(cmd));
         }
         else if (IsSpaceOrNewLine(*cmd))
         {
             *cmd = '\0';
         }
+
         if ((*cmd != '\0') && ((aArgc == 0) || (*(cmd - 1) == '\0')))
         {
             VerifyOrExit(aArgc < aArgcMax, error = OT_ERROR_INVALID_ARGS);
--- a/src/core/utils/parse_cmdline.cpp
+++ b/src/core/utils/parse_cmdline.cpp
@@ -31,6 +31,8 @@
  *   This file implements the command line parser.
  */
 
+#include <string.h>
+
 #include "parse_cmdline.hpp"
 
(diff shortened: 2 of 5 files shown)


src/core/utils/parse_cmdline.cpp
```diff
@@ -50,21 +52,17 @@ otError CmdLineParser::ParseCmd(char *aString, uint8_t &aArgc, char *aArgv[], ui
...
+        if ((*cmd == '\\') && (*(cmd + 1) == ' '))
+        {
+            strcpy(cmd, cmd + 1);
+        }
+        else if (IsSpaceOrNewLine(*cmd))
         {
             *cmd = '\0';
         }
-        else if (*(cmd - 1) == '\0')
+        if ((*cmd != '\0') && ((aArgc == 0) || (*(cmd - 1) == '\0')))
```
wgtdkp: nit: add a new line before this to separate if from the `else if (...) {...}` block.

src/core/utils/parse_cmdline.cpp
```diff
@@ -50,21 +52,17 @@ otError CmdLineParser::ParseCmd(char *aString, uint8_t &aArgc, char *aArgv[], ui
...
-    if (*cmd)
+    for (cmd = aString; *cmd; cmd++)
     {
-        aArgv[aArgc++] = cmd++; // the first argument
-    }
-
-    for (; *cmd; cmd++)
-    {
-        if (IsSpaceOrNewLine(*cmd))
+        if ((*cmd == '\\') && (*(cmd + 1) == ' '))
```
bukepo: Should we also escape `\` itself in case a string argument ends with `\`, but there is another argument follows?

src/core/utils/parse_cmdline.cpp
```diff
@@ -50,21 +52,17 @@ otError CmdLineParser::ParseCmd(char *aString, uint8_t &aArgc, char *aArgv[], ui
...
+        {
+            strcpy(cmd, cmd + 1);
```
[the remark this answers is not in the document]
edmont: Agree, will do.

src/core/utils/parse_cmdline.cpp
edmont: Agree, will do.

src/core/utils/parse_cmdline.cpp
edmont: Agree, will do.

@bukepo  @simonlingoogle 
Should we escape other separator characters as well (`'\t'`, `'\r'`, `'\n'`)?

simonlingoogle: I think it would be good to support other white spaces too. And maybe also `\\`?

edmont: @simonlingoogle take a look to the latest commits. We can extend `IsEscapable()` as desired.

src/cli/README.md
```diff
@@ -5,6 +5,27 @@ command line interface. Use the CLI to play with OpenThread, which
 can also be used with additional application code. The
 OpenThread test scripts use the CLI to execute test cases.
 
+## Separator and escaping characters
+
+The whitespace character (`' '`) is used to delimitate the command name
+and the different arguments, toghether with tab (`'\t'`) and new line
```
bukepo: typo:
```suggestion
and the different arguments, together with tab (`'\t'`) and new line
```

src/cli/README.md
```diff
@@ -5,6 +5,27 @@ command line interface. Use the CLI to play with OpenThread, which
 can also be used with additional application code. The
 OpenThread test scripts use the CLI to execute test cases.
 
+## Separator and escaping characters
+
+The whitespace character (`' '`) is used to delimitate the command name
```
bukepo:
```suggestion
The whitespace character (`' '`) is used to delaminate the command name
```

edmont: https://www.thefreedictionary.com/delimitate

src/core/utils/parse_cmdline.cpp
simonlingoogle: Do you mean `\t` or `\<tab>`? 
I think we can implement `IsEscapable` by calling `IsSpaceOrNewLine` directly:
```
return IsSpaceOrNewLine(c) || c == '\\'
```

edmont:
> Will you also enhance uses of `networkname` command in Python scripts? I mean we should escape names that contain whitespaces.
> @edmont

Are there more cases apart from `networkname`?

src/core/utils/parse_cmdline.cpp
edmont: @simonlingoogle yes, I was thinking about that. Will do.

edmont:
> > Are there more cases apart from `networkname`?

So far I identified:

- `networkname`
- `dataset networkname`
- `commissioner provisioningurl` (but URLs shouldn't contain spaces)

Other UTF-8 strings, not currently handled by the CLI:

- Commissioner Credential (#4597)
- Vendor Name
- Model Name
- Thread Domain Name (#4327)

bukepo:
There's a UTF-8 validation function: https://github.com/openthread/openthread/blob/master/src/lib/spinel/spinel.c#L176
We may consider reuse it later.

src/core/utils/parse_cmdline.cpp
```diff
@@ -43,28 +45,30 @@ static bool IsSpaceOrNewLine(char aChar)
...
     {
-        aArgv[aArgc++] = cmd++; // the first argument
-    }
-
-    for (; *cmd; cmd++)
-    {
-        if (IsSpaceOrNewLine(*cmd))
+        if ((*cmd == '\\') && IsEscapable(*(cmd + 1)))
+        {
+            memmove(cmd, cmd + 1, strlen(cmd));
```
wgtdkp: nit: `memmove(cmd, cmd + 1, strlen(cmd+1) + 1)` would be clearer to us. We got two reviewers thinking this is wrong at first glance :)

tools/harness-thci/OpenThread.py
```diff
@@ -626,6 +626,20 @@ def __getCommissionerSessionId(self):
...
+
+        Args:
+            string (str): UTF-8 input string.
+
+        Returns:
+            [str]: The modified string with escaped characters.
+        """
+        escapable_chars = ['\\', ' ', '\t', '\r', '\n']
+        for char in escapable_chars:
+            string = string.replace(char, '\ %s' % char)
```
librasungirl:
```suggestion
            string = string.replace(char, '\%s' % char)
```

edmont: Oops, typo, thanks!

src/core/utils/parse_cmdline.cpp
edmont: @wgtdkp :)

Well, yes, it would be clearer but take more program memory as well. Maybe it's better to just add a comment clarifying that we want to move everything from the next character until the null terminator, both included?

wgtdkp: it works to me.

edmont: Done.

wgtdkp:
Thanks @edmont 
LGTM 👍

edmont:
Should the solution provided for `tools/harness-thci/OpenThread.py` be extended to `tools/harness-thci/OpenThread_WpanCtl.py` and `tests/scripts/thread-cert/node.py` as well?

tools/harness-thci/OpenThread.py
```diff
@@ -626,6 +626,20 @@ def __getCommissionerSessionId(self):
         print('%s call getCommissionerSessionId' % self.port)
         return self.__sendCommand('commissioner sessionid')[0]
 
+    def __escape_escapable(self, string):
```
wgtdkp:
```suggestion
    def _escape_escapable(self, string):
```

we use single underline prefix for private methods.

edmont: @wgtdkp it seems there is mixed usage within this file, but with a majority of double underline.

I would name it as `__escapeEscapable` (camel case) to keep consistency. What do you think?

wgtdkp: SGTM

tests/scripts/thread-cert/node.py
```diff
@@ -446,6 +446,20 @@ def set_network_id_timeout(self, network_id_timeout):
...
+    def _escape_escapable(self, string):
+        """Escape CLI escapable characters in the given string.
+
+        Args:
+            string (str): UTF-8 input string.
+
+        Returns:
+            [str]: The modified string with escaped characters.
+        """
+        escapable_chars = ['\\', ' ', '\t', '\r', '\n']
```
simonlingoogle: nit: 
```suggestion
        escapable_chars = '\\ \t\r\n'
```

tools/harness-thci/OpenThread.py
```diff
@@ -626,6 +626,20 @@ def __getCommissionerSessionId(self):
...
+    def __escapeEscapable(self, string):
+        """Escape CLI escapable characters in the given string.
+
+        Args:
+            string (str): UTF-8 input string.
+
+        Returns:
+            [str]: The modified string with escaped characters.
+        """
+        escapable_chars = ['\\', ' ', '\t', '\r', '\n']
```
simonlingoogle: nit:
```suggestion
        escapable_chars = '\\ \t\r\n'
```

PR #4612, opened 2020-02-28
[coap] reduce code size by avoiding using 64bit division

bukepo:
* Removed OpenThread's default CoAP configuration macro, because they are fixed values by Thread spec.
* Use 32bit multiple only to detect overflow of application CoAP transmission parameters.
* Added documentation describing conditions of valid CoAP transmission parameters.

### DIFF
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -785,7 +785,7 @@ otMessage *otCoapNewMessage(otInstance *aInstance, const otMessageSettings *aSet
  * @param[in]  aTxParameters    A pointer to transmission parameters for this request. Use NULL for defaults.
  *                              Otherwise, parameters given must meet the following conditions:
  *                              1. mMaxRetransmit is no more than OT_COAP_MAX_RETRANSMIT.
- *                              2. mAckRandomFactorNumerator / mAckRandomFactorDenominator must not below 1.0.
+ *                              2. mAckRandomFactorNumerator / mAckRandomFactorDenominator must not be below 1.0.
  *                              3. The calculated exchange life time must not overflow uint32_t.
  *
  * @retval OT_ERROR_INVALID_ARGS    @p aTxParameters is invalid.
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -62,6 +62,8 @@ extern "C" {
 
 #define OT_COAP_MAX_RETRANSMIT 30 ///< Max retransmit supported by OpenThread.
 
+#define OT_COAP_MIN_ACK_TIMEOUT 1000 ///< Minimal ACK timeout in milliseconds supported by OpenThread.
+
 /**
  * CoAP Type values.
  *
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -780,8 +780,9 @@ otMessage *otCoapNewMessage(otInstance *aInstance, const otMessageSettings *aSet
  * @param[in]  aContext         A pointer to arbitrary context information. May be NULL if not used.
  * @param[in]  aTxParameters    A pointer to transmission parameters for this request. Use NULL for defaults.
  *
- * @retval OT_ERROR_NONE    Successfully sent CoAP message.
- * @retval OT_ERROR_NO_BUFS Failed to allocate retransmission data.
+ * @retval OT_ERROR_INVALID_ARGS    @p aTxParameters is invalid.
+ * @retval OT_ERROR_NONE            Successfully sent CoAP message.
+ * @retval OT_ERROR_NO_BUFS         Failed to allocate retransmission data.
  *
  */
 otError otCoapSendRequestWithParameters(otInstance *              aInstance,
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -60,6 +60,8 @@ extern "C" {
 
 #define OT_COAP_MAX_TOKEN_LENGTH 8 ///< Max token length as specified (RFC 7252).
 
+#define OT_COAP_MAX_RETRANSMIT 30 ///< Max retransmit supported by OpenThread.
+
 /**
  * CoAP Type values.
  *
@@ -779,6 +781,10 @@ otMessage *otCoapNewMessage(otInstance *aInstance, const otMessageSettings *aSet
  * @param[in]  aHandler         A function pointer that shall be called on response reception or timeout.
  * @param[in]  aContext         A pointer to arbitrary context information. May be NULL if not used.
  * @param[in]  aTxParameters    A pointer to transmission parameters for this request. Use NULL for defaults.
+ *                              Otherwise, parameters given must meet the following conditions:
+ *                              1. mMaxRetransmit is no more than OT_COAP_MAX_RETRANSMIT.
+ *                              2. mAckRandomFactorNumerator / mAckRandomFactorDenominator must not below 1.0.
+ *                              3. The calculated exchange life time must not overflow uint32_t.
  *
  * @retval OT_ERROR_INVALID_ARGS    @p aTxParameters is invalid.
  * @retval OT_ERROR_NONE            Successfully sent CoAP message.
--- a/src/core/api/coap_api.cpp
+++ b/src/core/api/coap_api.cpp
@@ -218,20 +218,17 @@ otError otCoapSendRequestWithParameters(otInstance *              aInstance,
                                         void *                    aContext,
                                         const otCoapTxParameters *aTxParameters)
 {
-    otError            error;
-    Instance &         instance = *static_cast<Instance *>(aInstance);
-    Coap::TxParameters txParameters;
+    otError   error;
+    Instance &instance = *static_cast<Instance *>(aInstance);
 
-    VerifyOrExit(aTxParameters == NULL ||
-                     (aTxParameters->mAckRandomFactorNumerator >= aTxParameters->mAckRandomFactorDenominator &&
-                      ((aTxParameters->mAckTimeout * ((1ULL << aTxParameters->aMaxRetx + 1) - 1) /
-                        aTxParameters->mAckRandomFactorDenominator * aTxParameters->mAckRandomFactorNumerator) >>
-                       32) == 0),
-                 error = OT_ERROR_INVALID_ARGS);
+    if (aTxParameters != NULL)
+    {
+        VerifyOrExit(Coap::TxParameters::Validate(*aTxParameters), error = OT_ERROR_INVALID_ARGS);
+    }
 
     error = instance.GetApplicationCoap().SendMessage(*static_cast<Coap::Message *>(aMessage),
                                                       *static_cast<const Ip6::MessageInfo *>(aMessageInfo),
-                                                      txParameters, aHandler, aContext);
+                                                      Coap::TxParameters::From(aTxParameters), aHandler, aContext);
 
 exit:
     return error;
--- a/src/core/api/coap_api.cpp
+++ b/src/core/api/coap_api.cpp
@@ -218,11 +218,23 @@ otError otCoapSendRequestWithParameters(otInstance *              aInstance,
                                         void *                    aContext,
                                         const otCoapTxParameters *aTxParameters)
 {
-    Instance &instance = *static_cast<Instance *>(aInstance);
+    otError            error;
+    Instance &         instance = *static_cast<Instance *>(aInstance);
+    Coap::TxParameters txParameters;
 
-    return instance.GetApplicationCoap().SendMessage(*static_cast<Coap::Message *>(aMessage),
-                                                     *static_cast<const Ip6::MessageInfo *>(aMessageInfo),
-                                                     Coap::TxParameters::From(aTxParameters), aHandler, aContext);
+    VerifyOrExit(aTxParameters == NULL ||
+                     (aTxParameters->mAckRandomFactorNumerator >= aTxParameters->mAckRandomFactorDenominator &&
+                      ((aTxParameters->mAckTimeout * ((1ULL << aTxParameters->aMaxRetx + 1) - 1) /
+                        aTxParameters->mAckRandomFactorDenominator * aTxParameters->mAckRandomFactorNumerator) >>
+                       32) == 0),
+                 error = OT_ERROR_INVALID_ARGS);
+
+    error = instance.GetApplicationCoap().SendMessage(*static_cast<Coap::Message *>(aMessage),
+                                                      *static_cast<const Ip6::MessageInfo *>(aMessageInfo),
+                                                      txParameters, aHandler, aContext);
+
+exit:
+    return error;
 }
 
 otError otCoapStart(otInstance *aInstance, uint16_t aPort)
--- a/src/core/api/coap_api.cpp
+++ b/src/core/api/coap_api.cpp
@@ -218,17 +218,18 @@ otError otCoapSendRequestWithParameters(otInstance *              aInstance,
                                         void *                    aContext,
                                         const otCoapTxParameters *aTxParameters)
 {
-    otError   error;
-    Instance &instance = *static_cast<Instance *>(aInstance);
+    otError                   error;
+    Instance &                instance     = *static_cast<Instance *>(aInstance);
+    const Coap::TxParameters &txParameters = Coap::TxParameters::From(aTxParameters);
 
     if (aTxParameters != NULL)
     {
-        VerifyOrExit(Coap::TxParameters::Validate(*aTxParameters), error = OT_ERROR_INVALID_ARGS);
+        VerifyOrExit(txParameters.IsValid(), error = OT_ERROR_INVALID_ARGS);
     }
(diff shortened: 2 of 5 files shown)


src/core/coap/coap.cpp
```diff
@@ -981,8 +981,13 @@ uint32_t TxParameters::CalculateMaxTransmitWait(void) const
 
 uint32_t TxParameters::CalculateSpan(uint32_t aMaxRetx) const
 {
+#if OPENTHREAD_CONFIG_COAP_API_ENABLE
```
abtink: This is fine as a quick fix to reduce code size.

Some other suggestion
- Do we really need 64 bit calculation for calculating the coap parameters values? Can we consider using 32 bit in all cases?
- The size increase is coming from configurable coap tx parameters feature. We can cosnider adding a config/build option around it to be able to enable/disable the feature/APIs if not needed (it would reduce code size even more if we really use only constant values and not need to calculate them at run time).

I personally prefer option 1. Thoughts? @bukepo, @jwhui

dismirlian: Hi all, I think for normal use cases we could use 32 bit calculations. For very long delays and/or high ReTx count this would result in overflow.

Two comments, besides those by @abtink:

1. We could reverse the order of the division and muliplication, to extend the range a bit.
2. We *should* state clearly this overflow risk in the docs.

bukepo: Yes, this is a quick fix. I  also prefer just using 32bit  for all cases if possible. If that's not possible, I think we need to do all @abtink and @dismirlian suggested.

bukepo:
@dismirlian any more comments?

src/core/coap/coap.cpp
jwhui: I'm fine with limiting the range if there are no compelling use cases. I agree with @dismirlian that the limitations should be clearly documented.

dismirlian: @bukepo, no, it seems fine!

include/openthread/coap.h
```diff
@@ -779,9 +781,14 @@ otMessage *otCoapNewMessage(otInstance *aInstance, const otMessageSettings *aSet
  * @param[in]  aHandler         A function pointer that shall be called on response reception or timeout.
  * @param[in]  aContext         A pointer to arbitrary context information. May be NULL if not used.
  * @param[in]  aTxParameters    A pointer to transmission parameters for this request. Use NULL for defaults.
- *
- * @retval OT_ERROR_NONE    Successfully sent CoAP message.
- * @retval OT_ERROR_NO_BUFS Failed to allocate retransmission data.
+ *                              Otherwise, parameters given must meet the following conditions:
+ *                              1. mMaxRetransmit is no more than OT_COAP_MAX_RETRANSMIT.
+ *                              2. mAckRandomFactorNumerator / mAckRandomFactorDenominator must not below 1.0.
```
simonlingoogle:
```suggestion
 *                              2. mAckRandomFactorNumerator >= mAckRandomFactorDenominator.
```

src/core/coap/coap.cpp
```diff
@@ -962,6 +962,37 @@ uint32_t ResponsesQueue::ResponseMetadata::GetRemainingTime(void) const
...
+static uint32_t Multiple(uint32_t aValueA, uint32_t aValueB)
+{
+    uint32_t result = aValueA * aValueB;
+
+    return (result / aValueA == aValueB) ? result : 0;
+}
+
+bool TxParameters::Validate(const otCoapTxParameters &aTxParameters)
+{
+    bool rval;
```
simonlingoogle:
```suggestion
    bool rval = false;
```

src/core/coap/coap.cpp
```diff
@@ -962,6 +962,37 @@ uint32_t ResponsesQueue::ResponseMetadata::GetRemainingTime(void) const
...
+
+        tmp /= aTxParameters.mAckRandomFactorDenominator;
+        tmp = Multiple(tmp, aTxParameters.mAckRandomFactorNumerator);
+
+        rval = (tmp != 0 && (tmp + aTxParameters.mAckTimeout + 2 * kDefaultMaxLatency) > tmp);
+    }
+    else
+    {
+        rval = false;
+    }
```
simonlingoogle:
```suggestion
```

src/core/coap/coap.cpp
```diff
@@ -979,10 +1010,10 @@ uint32_t TxParameters::CalculateMaxTransmitWait(void) const
     return CalculateSpan(mMaxRetransmit + 1);
 }
 
-uint32_t TxParameters::CalculateSpan(uint32_t aMaxRetx) const
+uint32_t TxParameters::CalculateSpan(uint8_t aMaxRetx) const
 {
-    return static_cast<uint32_t>(mAckTimeout * ((1ULL << aMaxRetx) - 1) * mAckRandomFactorNumerator /
-                                 mAckRandomFactorDenominator);
+    return static_cast<uint32_t>(mAckTimeout * ((1U << aMaxRetx) - 1) / mAckRandomFactorDenominator *
```
simonlingoogle: could incorrectly calculate to 0 if `mAckTimeout * ((1U << aMaxRetx) - 1)` < `mAckRandomFactorDenominator`?

src/core/coap/coap.cpp
```diff
@@ -962,6 +962,37 @@ uint32_t ResponsesQueue::ResponseMetadata::GetRemainingTime(void) const
     return (mDequeueTime > now) ? mDequeueTime - now : 0;
 }
 
+/// Return product of @p aValueA and @p aValueB if no overflow otherwise 0.
+static uint32_t Multiple(uint32_t aValueA, uint32_t aValueB)
```
simonlingoogle: Should we make this function a member of `TxParameters` since it has special rules that only applies to `TxParameters`?

bukepo: I prefer keep implementation in .cpp file if possible.

src/core/coap/coap.cpp
bukepo: added verification of `mAckTimeout` is no less than 1.0.

include/openthread/coap.h
bukepo: this is copied from RFC.

simonlingoogle:
Is it necessary to add a check to make sure `64bit division` is not added to OpenThread accidentally in the future? @bukepo

bukepo:
@simonlingoogle I think it's doable, but we should only check objects generated by OpenThread, so that if platforms support 64bit division, they will not be affected. And I think it can be enhanced in another PR. I expect this PR be merged soon.

include/openthread/coap.h
```diff
@@ -779,9 +783,14 @@ otMessage *otCoapNewMessage(otInstance *aInstance, const otMessageSettings *aSet
...
```
abtink:
```suggestion
 *                              2. mAckRandomFactorNumerator / mAckRandomFactorDenominator must not be below 1.0.
```

src/core/coap/coap.hpp
```diff
@@ -113,17 +123,17 @@ class TxParameters : public otCoapTxParameters
 private:
     enum
     {
-        kDefaultAckTimeout                 = OPENTHREAD_CONFIG_COAP_ACK_TIMEOUT_MILLIS,
-        kDefaultAckRandomFactorNumerator   = OPENTHREAD_CONFIG_COAP_ACK_RANDOM_FACTOR_NUMERATOR,
-        kDefaultAckRandomFactorDenominator = OPENTHREAD_CONFIG_COAP_ACK_RANDOM_FACTOR_DENOMINATOR,
-        kDefaultMaxRetransmit              = OPENTHREAD_CONFIG_COAP_MAX_RETRANSMIT,
+        kDefaultAckTimeout                 = 2000,
```
abtink:
```suggestion
        kDefaultAckTimeout                 = 2000, // in millisecond
```

src/core/coap/coap.cpp
```diff
@@ -962,6 +962,33 @@ uint32_t ResponsesQueue::ResponseMetadata::GetRemainingTime(void) const
...
+    bool rval = false;
+
+    if (aTxParameters.mAckRandomFactorNumerator >= aTxParameters.mAckRandomFactorDenominator &&
+        aTxParameters.mAckTimeout >= OT_COAP_MIN_ACK_TIMEOUT && aTxParameters.mMaxRetransmit <= OT_COAP_MAX_RETRANSMIT)
+    {
+        // Calulate exchange lifetime step by step and verify no overflow.
+        uint32_t tmp = Multiple(aTxParameters.mAckTimeout, (1U << (aTxParameters.mMaxRetransmit + 1)) - 1);
+
+        tmp /= aTxParameters.mAckRandomFactorDenominator;
+        tmp = Multiple(tmp, aTxParameters.mAckRandomFactorNumerator);
```
abtink: This is ok but should be aware that in some cases the order (whether multiply first or divide first) may matter (result in different rounding happening). I don't think it really impacts though.
(1001 * 8 / 3)  = 2669
(1001 / 3 * 8)  = 2664

bukepo: Yes, this is intentionally as suggested by @dismirlian

jwhui:
@sjlongland, can you review and provide your thoughts?

src/core/coap/coap.cpp
sjlongland: If we're worried about OpenThread application code supplying parameters that overflow, one prospect is we change that `#if` to `#if OPENTHREAD_CONFIG_COAP_LONG_TIMEOUT`… which we define in the configuration:

```
/**
 * @def OPENTHREAD_CONFIG_COAP_LONG_TIMEOUT
 *
 * Enable support for long time-out values in CoAP requests.  In ordinary circumstances, the time-out
 * calculations use 32-bit arithmetic which can lead to overflows when computing the total request time
 * span.  Set this to 1 to use 64-bit arithmetic to avoid these overflows.
 */
#ifndef OPENTHREAD_CONFIG_COAP_LONG_TIMEOUT
#define OPENTHREAD_CONFIG_COAP_LONG_TIMEOUT 0
#endif
```

sjlongland:
About the only use case I see with longer time-outs would be sleepy end devices… the question is, what sort of figures do we need to put in before we start hitting overflows?

If the numbers are obscenely large, then we could possibly just document the problem as suggested.  Otherwise I'd look at maybe making the 64-bit arithmetic switchable at compile time so in the "normal" use case, we can avoid the bloat whilst still allowing people to use longer time-outs if desired.

bukepo:
@sjlongland Is it possible to support 64bit timeout here when we really hit the case that 32bit cannot do? The 32bit gives a maximum delay of ~49 days. I think it can cover most normal use cases.

sjlongland:
On 3/3/20 4:17 pm, Yakun Xu wrote:
> @sjlongland Is it possible to support 64bit timeout here when we really hit the case that 32bit cannot do? The 32bit gives a maximum delay of ~49 days. I think it can cover most normal use cases. 
> 

49 days is enormous… On that basis unless someone is doing Thread
long-distance networking via homing pigeon, I think we just drop the
64-bit arithmetic entirely.

src/core/coap/coap.cpp
```diff
@@ -962,6 +962,33 @@ uint32_t ResponsesQueue::ResponseMetadata::GetRemainingTime(void) const
     return (mDequeueTime > now) ? mDequeueTime - now : 0;
 }
 
+/// Return product of @p aValueA and @p aValueB if no overflow otherwise 0.
+static uint32_t Multiple(uint32_t aValueA, uint32_t aValueB)
```
jwhui:
```suggestion
static uint32_t Multiply(uint32_t aValueA, uint32_t aValueB)
```

src/core/api/coap_api.cpp
```diff
@@ -218,11 +218,20 @@ otError otCoapSendRequestWithParameters(otInstance *              aInstance,
...
 {
+    otError   error;
     Instance &instance = *static_cast<Instance *>(aInstance);
 
-    return instance.GetApplicationCoap().SendMessage(*static_cast<Coap::Message *>(aMessage),
-                                                     *static_cast<const Ip6::MessageInfo *>(aMessageInfo),
-                                                     Coap::TxParameters::From(aTxParameters), aHandler, aContext);
+    if (aTxParameters != NULL)
+    {
+        VerifyOrExit(Coap::TxParameters::Validate(*aTxParameters), error = OT_ERROR_INVALID_ARGS);
```
jwhui: What about making this a non-static method and validating after the static cast?

bukepo: That would result in checking the default parameters again and again. I prefer  avoid the unnecessary checking. Thoughts?

src/core/coap/coap.cpp
```diff
@@ -962,6 +962,33 @@ uint32_t ResponsesQueue::ResponseMetadata::GetRemainingTime(void) const
...
+    bool rval = false;
+
+    if (aTxParameters.mAckRandomFactorNumerator >= aTxParameters.mAckRandomFactorDenominator &&
+        aTxParameters.mAckTimeout >= OT_COAP_MIN_ACK_TIMEOUT && aTxParameters.mMaxRetransmit <= OT_COAP_MAX_RETRANSMIT)
+    {
+        // Calulate exchange lifetime step by step and verify no overflow.
+        uint32_t tmp = Multiple(aTxParameters.mAckTimeout, (1U << (aTxParameters.mMaxRetransmit + 1)) - 1);
+
+        tmp /= aTxParameters.mAckRandomFactorDenominator;
+        tmp = Multiple(tmp, aTxParameters.mAckRandomFactorNumerator);
```
bukepo:
```suggestion
        tmp = Multiply(tmp, aTxParameters.mAckRandomFactorNumerator);
```

src/core/coap/coap.cpp
```diff
@@ -962,6 +962,33 @@ uint32_t ResponsesQueue::ResponseMetadata::GetRemainingTime(void) const
...
+
+bool TxParameters::Validate(const otCoapTxParameters &aTxParameters)
+{
+    bool rval = false;
+
+    if (aTxParameters.mAckRandomFactorNumerator >= aTxParameters.mAckRandomFactorDenominator &&
+        aTxParameters.mAckTimeout >= OT_COAP_MIN_ACK_TIMEOUT && aTxParameters.mMaxRetransmit <= OT_COAP_MAX_RETRANSMIT)
+    {
+        // Calulate exchange lifetime step by step and verify no overflow.
+        uint32_t tmp = Multiple(aTxParameters.mAckTimeout, (1U << (aTxParameters.mMaxRetransmit + 1)) - 1);
```
bukepo:
```suggestion
        uint32_t tmp = Multiply(aTxParameters.mAckTimeout, (1U << (aTxParameters.mMaxRetransmit + 1)) - 1);
```

src/core/api/coap_api.cpp
jwhui: You can avoid validating if `aTxParameters` is NULL.

bukepo: Oh, yes! I will update soon.

include/openthread/coap.h
```diff
@@ -60,6 +60,10 @@ extern "C" {
 
 #define OT_COAP_MAX_TOKEN_LENGTH 8 ///< Max token length as specified (RFC 7252).
 
+#define OT_COAP_MAX_RETRANSMIT 30 ///< Max retransmit supported by OpenThread.
+
+#define OT_COAP_MIN_ACK_TIMEOUT 1000 ///< Minimal ACK timeout in milliseconds supported by OpenThread.
```
jwhui: @sjlongland, does a minimum ACK timeout of 1 second fit with your use case(s)?

sjlongland: I think it should… I'd imagine you really don't want to be pummelling a low-rate mesh network at millisecond intervals with retransmissions.  I was thinking something in the realm of ~5 seconds or so.

A minimum of one would be fine.

bukepo: The 1 second value comes from CoAP RFC7252 4.8.1.

simonlingoogle: I think someone might requires a smaller timeout for the first retransmission. We should allow extreme use cases when the user need it. 1000ms is too much larger than the minimal timeout that OpenThread can support theoretically. Thoughts?

sjlongland: How probable is it that a response to a request is received within a sub-second interval?

On big and very busy mesh networks (30~50 nodes), a simple ICMP echo request takes ~2000ms to return.  A sub-second ACK time-out would have expired by then.

bukepo: I tend to leave this when we really hit a use case that requires sub-second ACK timeout. @simonlingoogle Thoughts?

simonlingoogle: > The 1 second value comes from CoAP RFC7252 4.8.1.

I am fine with using 1000 ms as min timeout and changing to smaller timeout if we find it necessary in the future.

PR #4615, opened 2020-03-01
[build] make MTD and RADIO_ONLY build configurable

librasungirl:
Some feature like Backbone Router in #4430 requires FTD capability, however previously FTD/MTD/RADIO_ONLY are bundled in Makefile recipe. 

This PR introduces MTD and RADIO_ONLY option to allow the flexibility to decide whether or not to build MTD/RADIO_ONLY when `make`.

### DIFF
--- a/examples/Makefile-cc1352
+++ b/examples/Makefile-cc1352
@@ -44,7 +44,6 @@ BuildJobs                                                 ?= 10
 configure_OPTIONS                                          = \
     --enable-cli                                             \
     --enable-ftd                                             \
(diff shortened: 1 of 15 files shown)


bukepo:
Is it possible to just ignore the bbr flag when building mtd? Commissioner is an [example](https://github.com/openthread/openthread/blob/master/src/core/meshcop/commissioner.cpp#L52).

librasungirl:
yes, it is also an option. will close this PR as it seems not mandatory so far.

PR #4616, opened 2020-03-01
Bugfix/struct alignment

richarddonkin:
Fix compilation errors on GCC 9.2.0 due to struct and class member alignment.

### DIFF
--- a/src/core/thread/network_diagnostic_tlvs.hpp
+++ b/src/core/thread/network_diagnostic_tlvs.hpp
@@ -157,7 +157,6 @@ public:
  * This class implements Extended Address TLV generation and parsing.
  *
  */
-OT_TOOL_PACKED_BEGIN
 class ExtMacAddressTlv : public NetworkDiagnosticTlv
 {
 public:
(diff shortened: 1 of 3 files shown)


src/core/thread/network_diagnostic_tlvs.hpp
```diff
@@ -157,7 +157,6 @@ class NetworkDiagnosticTlv : public ot::Tlv
  * This class implements Extended Address TLV generation and parsing.
  *
  */
-OT_TOOL_PACKED_BEGIN
```
wgtdkp: We make it packed in this PR, could you explain why we need to remove it?

richarddonkin: This class takes an address of its member mMacAddr in the function
GetMacAddr(). If the class is packed, the address of its member
may be misaligned, which results in a compilation error.

This helps fix compilation with arm-none-eabi-gcc 9.2.0.

richarddonkin: I closed this pull request because I'm still learning how to use Github and Travis. Please refer to https://github.com/openthread/openthread/pull/4617 for the open pull request.

PR #4617, opened 2020-03-01
[cc1352,cc26xx] fix unaligned access of ext address in radio.c

richarddonkin:
Fix compilation errors on GCC 9.2.0 due to struct and class member alignment.

### DIFF
--- a/examples/platforms/utils/encoding.h
+++ b/examples/platforms/utils/encoding.h
@@ -31,8 +31,8 @@
  *   This file includes macros for validating runtime conditions.
  */
 
-#ifndef ENCODING_H
-#define ENCODING_H
+#ifndef PLATFORM_UTILS_ENCODING_H
+#define PLATFORM_UTILS_ENCODING_H
 
 #include <stdint.h>
 
@@ -40,7 +40,11 @@
  * Converts a 64-bit unsigned integer stored as a little-endian byte
  * array to uint64_t.
  *
- * @parma[in]  aSource  The byte array.
+ * Allows for the array to be byte-aligned, so that loading the
+ * uint64_t value via this function avoids a processor exception due
+ * to unaligned access.
+ *
+ * @param[in]  aSource  The byte array.
  *
  * @returns The 64-bit value as a uint64_t.
  *
@@ -61,4 +65,4 @@ static inline uint64_t otEncodingReadUint64Le(const uint8_t *aSource)
     return value;
 }
 
-#endif // ENCODING_H
+#endif // PLATFORM_UTILS_ENCODING_H
--- a/examples/platforms/utils/encoding.h
+++ b/examples/platforms/utils/encoding.h
@@ -1,0 +1,64 @@
+/*
+ *  Copyright (c) 2020, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes macros for validating runtime conditions.
+ */
+
+#ifndef ENCODING_H
+#define ENCODING_H
+
+#include <stdint.h>
+
+/**
+ * Converts a 64-bit unsigned integer stored as a little-endian byte
+ * array to uint64_t.
+ *
+ * @parma[in]  aSource  The byte array.
+ *
+ * @returns The 64-bit value as a uint64_t.
+ *
+ */
+static inline uint64_t otEncodingReadUint64Le(const uint8_t *aSource)
+{
+    uint64_t value = 0;
+
+    value |= (uint64_t)aSource[0];
+    value |= ((uint64_t)aSource[1]) << 8;
+    value |= ((uint64_t)aSource[2]) << 16;
+    value |= ((uint64_t)aSource[3]) << 24;
+    value |= ((uint64_t)aSource[4]) << 32;
+    value |= ((uint64_t)aSource[5]) << 40;
+    value |= ((uint64_t)aSource[6]) << 48;
+    value |= ((uint64_t)aSource[7]) << 56;
+
+    return value;
+}
+
+#endif // ENCODING_H
--- a/include/openthread/platform/radio.h
+++ b/include/openthread/platform/radio.h
@@ -161,6 +161,29 @@ struct otExtAddress
 typedef struct otExtAddress otExtAddress;
 
 /**
+ * Converts an array representing an IEEE 802.15.4 Extended Address into a
+ * 64-bit value.
+ *
+ * This is necessary as the otExtAddress array is not necessary aligned for
+ * access as a 64-bit value.
+ */
+static inline uint64_t otExtAddressLoadUint64(const otExtAddress *address)
+{
+    uint64_t alignedValue = 0;
+    /* The following loads assume that the array is storing the 64-bit word
+     * in little-endian format. */
+    alignedValue |= (uint64_t)address->m8[0];
+    alignedValue |= ((uint64_t)address->m8[1]) << 8;
+    alignedValue |= ((uint64_t)address->m8[2]) << 16;
+    alignedValue |= ((uint64_t)address->m8[3]) << 24;
+    alignedValue |= ((uint64_t)address->m8[4]) << 32;
+    alignedValue |= ((uint64_t)address->m8[5]) << 40;
+    alignedValue |= ((uint64_t)address->m8[6]) << 48;
+    alignedValue |= ((uint64_t)address->m8[7]) << 56;
+    return alignedValue;
+}
+
+/**
  * This structure represents the IEEE 802.15.4 Header IE (Information Element) related information of a radio frame.
  */
 typedef struct otRadioIeInfo
--- a/include/openthread/platform/radio.h
+++ b/include/openthread/platform/radio.h
@@ -161,29 +161,6 @@ struct otExtAddress
 typedef struct otExtAddress otExtAddress;
 
 /**
- * Converts an array representing an IEEE 802.15.4 Extended Address into a
- * 64-bit value.
- *
- * This is necessary as the otExtAddress array is not necessary aligned for
- * access as a 64-bit value.
- */
-static inline uint64_t otExtAddressLoadUint64(const otExtAddress *address)
-{
-    uint64_t alignedValue = 0;
-    /* The following loads assume that the array is storing the 64-bit word
-     * in little-endian format. */
-    alignedValue |= (uint64_t)address->m8[0];
-    alignedValue |= ((uint64_t)address->m8[1]) << 8;
-    alignedValue |= ((uint64_t)address->m8[2]) << 16;
-    alignedValue |= ((uint64_t)address->m8[3]) << 24;
-    alignedValue |= ((uint64_t)address->m8[4]) << 32;
-    alignedValue |= ((uint64_t)address->m8[5]) << 40;
-    alignedValue |= ((uint64_t)address->m8[6]) << 48;
-    alignedValue |= ((uint64_t)address->m8[7]) << 56;
-    return alignedValue;
-}
-
-/**
  * This structure represents the IEEE 802.15.4 Header IE (Information Element) related information of a radio frame.
  */
 typedef struct otRadioIeInfo
--- a/examples/platforms/cc1352/radio.c
+++ b/examples/platforms/cc1352/radio.c
@@ -1577,7 +1577,7 @@ otError otPlatRadioAddSrcMatchExtEntry(otInstance *aInstance, const otExtAddress
     OT_UNUSED_VARIABLE(aInstance);
 
     otError  error      = OT_ERROR_NONE;
-    uint64_t extAddress = *(uint64_t *)aExtAddress;
+    uint64_t extAddress = otExtAddressLoadUint64(aExtAddress);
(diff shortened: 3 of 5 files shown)


wgtdkp:
@richarddonkin thanks for reporting this. Could you paste the compiling error here? It would be very helpful. Thanks.

richarddonkin:
There are two errors, corresponding to the two commits.
First error:

    /mnt/home-hdd/richard/projects/embedded/security-button/ot-rtos/third_party/openthread/repo/examples/../examples/platforms/cc1352/radio.c: In function 'otPlatRadioAddSrcMatchExtEntry':
    /mnt/home-hdd/richard/projects/embedded/security-button/ot-rtos/third_party/openthread/repo/examples/../examples/platforms/cc1352/radio.c:1579:5: error: converting a packed 'otExtAddress' {aka 'const struct otExtAddress'} pointer (alignment 1) to a 'uint64_t' {aka 'const long long unsigned int'} pointer (alignment 8) may result in an unaligned pointer value [-Werror=address-of-packed-member]
     1579 |     uint8_t idx   = rfCoreFindExtSrcMatchIdx((uint64_t *)aExtAddress);
          |     ^~~~~~~
    /mnt/home-hdd/richard/projects/embedded/security-button/ot-rtos/third_party/openthread/repo/examples/../examples/platforms/cc1352/radio.c: In function 'otPlatRadioClearSrcMatchExtEntry':
    /mnt/home-hdd/richard/projects/embedded/security-button/ot-rtos/third_party/openthread/repo/examples/../examples/platforms/cc1352/radio.c:1614:5: error: converting a packed 'otExtAddress' {aka 'const struct otExtAddress'} pointer (alignment 1) to a 'uint64_t' {aka 'const long long unsigned int'} pointer (alignment 8) may result in an unaligned pointer value [-Werror=address-of-packed-member]
     1614 |     otEXPECT_ACTION((idx = rfCoreFindExtSrcMatchIdx((uint64_t *)aExtAddress)) != CC1352_SRC_MATCH_NONE,
          |     ^~~~~~~~~~~~~~~

Second error:

    In file included from /mnt/home-hdd/richard/workspace/openthread/examples/../src/core/thread/network_diagnostic.hpp:42,
                     from /mnt/home-hdd/richard/workspace/openthread/examples/../src/core/thread/thread_netif.hpp:73,
                     from /mnt/home-hdd/richard/workspace/openthread/examples/../src/core/common/instance.hpp:69,
                     from /mnt/home-hdd/richard/workspace/openthread/examples/../src/core/api/coap_api.cpp:39:
    /mnt/home-hdd/richard/workspace/openthread/examples/../src/core/thread/network_diagnostic_tlvs.hpp: In member function 'const ot::Mac::ExtAddress* ot::NetworkDiagnostic::ExtMacAddressTlv::GetMacAddr() const':
    /mnt/home-hdd/richard/workspace/openthread/examples/../src/core/thread/network_diagnostic_tlvs.hpp:189:60: error: taking address of packed member of 'ot::NetworkDiagnostic::ExtMacAddressTlv' may result in an unaligned pointer value [-Werror=address-of-packed-member]
      189 |     const Mac::ExtAddress *GetMacAddr(void) const { return &mMacAddr; }
          |                                                            ^~~~~~~~~

wgtdkp:
hi @richarddonkin Thanks for the error message.

For the first error, can this PR https://github.com/openthread/openthread/pull/4323 fix it?

For the second error, it seems can be avoided by returning a reference, like this:
```c++
const Mac::ExtAddress &GetMacAddr(void) const { return mMacAddr; }
```

richarddonkin:
Hi @wgtdkp 

PR https://github.com/openthread/openthread/pull/4323 still dereferences unaligned pointers. The compiler error is gone because the compiler is not sophisticated enough to detect the problem in this case, but the problem is the same. Please see https://blog.quarkslab.com/unaligned-accesses-in-cc-what-why-and-solutions-to-do-it-properly.html for a nicely written explanation of a similar case that fails even though it generated no warning.

In order to access the buffer safely as a pointer or any aligned value, the buffer must also be aligned.

Unfortunately an unaligned reference has the same problems as an unaligned pointer. Your suggestion of returning a reference instead of a pointer also just removes the compiler error due to the compiler not being able to detect the problem in this case. Please see https://stackoverflow.com/questions/17146792/why-do-i-get-a-warning-for-an-unaligned-pointer-but-not-a-reference?rq=1 for an explanation of a similar case.

As far as I know, unaligned accesses cannot be handled on the Cortex M4 processor unless there is special operating system code to handle the processor exceptions and perform byte-for-byte copies. I imagine most real-time projects will not consider this an acceptable solution in terms of predictable behaviour and timing.

wgtdkp:
@richarddonkin Thanks for your research. It seems dereferencing unaligned pointer on MCUs is really dangerous and I would prefer not using it:

- Extended MAC Address is represented as 8 bytes array in openthread, it doesn't make sense to make it aligned by 8 bytes.
- The solutions suggested by https://blog.quarkslab.com/unaligned-accesses-in-cc-what-why-and-solutions-to-do-it-properly.html sounds good to me. Especially the **multiple loads**, it is alignment independent and endianness independent (and I don't think there are too much overhead of this).

For the second error:
- First, the `ExtMacAddressTlv` is used to map directly to a Extended Address TLV in Network Diagnostic message payload, so it should be packed.
- Second, we don't use such reinterpret-casting on return value of `const Mac::ExtAddress &GetMacAddr(void) const { return mMacAddr; }`, so it should be ok.

TLV structures like `ExtMacAddressTlv` are all packed and one can always cast them to pointers of other type. I think the best way to avoid the alignment problem is to void reinterpret-casting to type with greater alignment.

Thoughts?

richarddonkin:
@wgtdkp Thanks for your good suggestions.

On the first point, I think using multiple loads to construct a uint64_t value from otExtAddress array is a great idea. It can even be done using an inline function defined the same header that defines otExtAddress.

On your second point, I understand now the reason that ExtMacAddressTlv needs to be packed. However returning a reference to a member of a packed class can still cause an unaligned access, regardless of whether it is used in a reinterpret cast. There is an implicit cast from "Mac::ExtAddress packed &" to "Mac::ExtAddress&" (using my own pseudo-code). It is always possible that "Mac::ExtAddres packed &" is misaligned; that is the nature of packing.

One solution could be that the function should return Mac::ExtAddress; not a reference nor a pointer. This will cause the value to be copied when it is returned, avoiding the issue of misaligned pointers or references.

wgtdkp:
@richarddonkin 
>There is an implicit cast from Mac::ExtAddress packed & to Mac::ExtAddress& (using my own pseudo-code).

Not sure how gcc packed attribute works, but `Mac::ExtAddress` itself is also packed. Will this make `Mac::ExtAddress packed &` and `MacExtAddress &` have the same semantic?

richarddonkin:
@wgtdkp 
> Not sure how gcc packed attribute works, but `Mac::ExtAddress` itself is also packed. Will this make `Mac::ExtAddress packed &` and `MacExtAddress &` have the same effect?

Unfortunately not. The GCC packed attribute packs the members inside the struct, relative to the start of the struct. It doesn't allow the struct as a whole to start at a misaligned address.

include/openthread/platform/radio.h
```diff
@@ -150,7 +150,9 @@ typedef uint16_t otShortAddress;
 OT_TOOL_PACKED_BEGIN
 struct otExtAddress
 {
-    uint8_t m8[OT_EXT_ADDRESS_SIZE]; ///< IEEE 802.15.4 Extended Address bytes
+    OT_TOOL_ALIGNED_FIELD_BEGIN(OT_EXT_ADDRESS_SIZE)
```
abtink: Reading [EWARM](http://supp.iar.com/FilesPublic/UPDINFO/004916/arm/doc/EWARM_DevelopmentGuide.ENU.pdf), for `_Pragma("data_alignment")` I think this would request the struct filed to be aligned certain way but we actually want the whole `otExtAddress` to be packed. How would this work?

richarddonkin: The alignment of a plain struct is determined by the alignment of its first member.

However I am planning rather to approach things differently based on the discussion above. The otExtAddress struct does not need its member aligned, if the bytes are loaded explicitly from the array to form a 64-bit value. See https://blog.quarkslab.com/unaligned-accesses-in-cc-what-why-and-solutions-to-do-it-properly.html . I'll add commits to the pull request when I've made the changes.

jwhui: I think we want to keep the packed attribute because we do want unaligned access.

In the case of https://github.com/openthread/openthread/blob/150aa9b6db6f241bdc6056ff6c45c039cabc90b0/examples/platforms/cc1352/radio.c#L1580 we could simply use `memcpy()` or something equivalent.

abtink: We use packed definition basically to allow unaligned access (being able to cast any pointer to the packed type). This is for example used in TLV parsing.

I would also echo @jwhui  suggestion to do copying of data/value and trying to avoid direct casting to `unit` pointer and deference.

include/openthread/platform/radio.h
```diff
@@ -159,6 +159,29 @@ struct otExtAddress
...
 typedef struct otExtAddress otExtAddress;
 
+/**
+ * Converts an array representing an IEEE 802.15.4 Extended Address into a
+ * 64-bit value.
+ *
+ * This is necessary as the otExtAddress array is not necessary aligned for
+ * access as a 64-bit value.
+ */
+static inline uint64_t otExtAddressLoadUint64(const otExtAddress *address)
```
jwhui: @srickardti, is this necessary?

In https://github.com/openthread/openthread/pull/4323#issuecomment-553178110 you had commented:

> I'm not really clear on why this is necessary. The pointer in question was used only to cast between the otExtAddress type and a uint64_t. The M4F or M3 can handle unaligned accesses, and the M0 radio controller is protected by the alignment of the structs it is sent.

In https://github.com/openthread/openthread/pull/4323#discussion_r345513880, you had commented:

> This could be done with a cast and assignment to a uint64_t. I think GCC is smart enough to optimize out the function call, so that is probably fine.

Thoughts?

srickardti: I'm a little unclear what the point of this function is, and the doxygen comment is as well. This function is not aligning access, it is loading an array of little-endian order bytes into a `uint64_t`. Which is then returned to the caller. Alignment in that case is actually enforced by the caller.

Since both the CC1352 and CC2652 are little endian it is equivalent to the cast and de-reference hack that was there before. And GCC seems to be intelligent enough to figure that out: https://godbolt.org/z/qGXJc6. My only concern would be the TI and IAR compilers being as clever.

richarddonkin:
@srickardti

I disagree with the statement that the Cortex M4F supports unaligned accesses. Please refer to https://developer.arm.com/docs/dui0553/a/the-cortex-m4-instruction-set/about-the-instruction-descriptions/address-alignment

Basically the compiler may emit an instruction that assumes the address is aligned and causes a UsageFault if it is not aligned, or the memory attached to the bus may generate a bus fault if it does not support unaligned access. 

This function loads the bytes into a 64-bit word, but ensures that the compiler will only use instructions that support byte-aligned access. This is necessary because the struct may be misaligned with the 32-bit memory bus. 

Please see the discussion above in previous comments and the web links referred to, which explain unaligned access and how to fix it. 


-------- Original Message --------

srickardti:
@richarddonkin , you are correct, I was mistaken about a statement in our TRM. I thought this would generate two `ldr` instructions, but it generates an `ldrd` which does not allow non-word aligned accesses. GCC seems to be smart enough to optimize for space with two `ldr` instructions. 

I have no issues with this PR. This does also have a side-effect of byte reversing the address on big-endian systems.

include/openthread/platform/toolchain.h
```diff
@@ -103,6 +118,8 @@ extern "C" {
 #define OT_TOOL_PACKED_FIELD __attribute__((packed))
 #define OT_TOOL_PACKED_END __attribute__((packed))
 #define OT_TOOL_WEAK __attribute__((weak))
+#define OT_TOOL_ALIGNED_FIELD_BEGIN(size)
```
jwhui: Given that these macros are no longer needed in this PR, can we avoid making changes to this file in this PR?

dismirlian:
I think that one can always define a pointer to an unaligned location:

Eg. in GCC:
```
typedef int packedint __attribute__((aligned(1)));

packedint a = &packed_struct.member;
```

Full example:
```
struct  __attribute__((__packed__)) ps {
    char c;
    int i;
};
volatile struct ps ps_test;

struct rs {
    char c;
    int i;
};
volatile struct rs rs_test;


int f(volatile int *a) {
    return *a;
}

//function g treats a as packed
typedef int packedint __attribute__((aligned(1)));
int g(volatile packedint  *a) {
    return *a;
}

int main(void) {
    volatile int j;
    
    j = f(&ps_test.i);      //compiler (correctly) complains here
    j = g(&ps_test.i);      //compiler doesn't complain
    j = f(&rs_test.i);      //compiler doesn't complain
    j = g(&rs_test.i);      //compiler doesn't complain

    (void)j;
    return 0;
}
```

My compiler version:
`arm-none-eabi-gcc.exe (GNU Tools for Arm Embedded Processors 9-2019-q4-major) 9.2.1 20191025 (release) [ARM/arm-9-branch revision 277599]`

With Keil: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0491c/CJAFJHJD.html
`__packed int *p;                    /* pointer to a __packed int */`

dismirlian:
More importantly, enabling the `-Wcast-align` GCC warning reveals some potential problems (not necessarily true, but maybe worth analyzing). Maybe the compilers are not currently generating instructions that don't support unaligned accesses in Cortex-M4F.

Some examples:

```
openthread/src/core/utils/heap.hpp:96:115: error: cast from 'const uint8_t*' {aka 'const unsigned char*'} to 'const uint16_t*' {aka 'const short unsigned int*'} increases required alignment of target type [-Werror=cast-align]
   96 |         return *reinterpret_cast<const uint16_t *>(reinterpret_cast<const uint8_t *>(this) + sizeof(mSize) + mSize);
      |                                                                                                                   ^
```
```
openthread/src/core/thread/network_data_tlvs.hpp:891:90: error: cast from 'uint8_t*' {aka 'unsigned char*'} to 'uint32_t*' {aka 'long unsigned int*'} increases required alignment of target type [-Werror=cast-align]
  891 |             return HostSwap32(*reinterpret_cast<uint32_t *>(GetEnterpriseNumberLocation()));
```

include/openthread/platform/toolchain.h
richarddonkin: I've removed the commit with the uneccessary macros.

include/openthread/platform/radio.h
jwhui, line 169: AFAICT, this function is only needed by the TI platforms, currently. Can we move this to TI-specific platform code?

richarddonkin: I'm not against that idea. Technically it could be needed by other platforms if they ever want to load the address into a 64-bit word, but that is not the case at the moment.

However I would like to ask that someone else take over the pull request in that case. I don't see one location where I should put this function so that it will be TI-specific and still follow the overall structure of the openthread repo. I foresee some bikeshedding around the best location for the function and I don't know the project well enough to have a strong opinion.

jwhui: @srickardti, thoughts?

abtink: @srickardti , I agree with @jwhui in that I feel this function does not really belong in `radio.h`. OT radio platform definition treat the extended MAC address as an array of 8 bytes in `otExtAddress`.

To me, this is more of a generic helper function that reads the value from a pointer (to an array/buffer) as a `uint64_t` with a specific encoding (little endian). 

- In core we have `common/encoding.hpp` which provides different `ReadUint16/ReadUint32` flavors (for big/little endian).
- How about adding this function (with maybe other similar functions for reading `uint16_t` and `uint32_t`) for all platform layer code under `example/platforms/utils` folder in a `encoding.h` header (this way it can shared/use by all radios/platform, similar to how `utils/code_utils.h` header is used).

jwhui: @richarddonkin, @srickardti, @abtink - I updated this PR as @abtink suggested above. Please review.

examples/platforms/utils/encoding.h
```diff
@@ -0,0 +1,64 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes macros for validating runtime conditions.
+ */
+
+#ifndef ENCODING_H
+#define ENCODING_H
```
richarddonkin, line 35: It might be better to use a longer macro name here, such as `PLATFORM_UTILS_ENCODING_H_`. The name `ENCODING_H` seems quite short so it could easily clash with other headers, including in user applications.

examples/platforms/utils/encoding.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include <stdint.h>
+
+/**
+ * Converts a 64-bit unsigned integer stored as a little-endian byte
+ * array to uint64_t.
+ *
+ * @parma[in]  aSource  The byte array.
```
richarddonkin, line 43: This should be `@param`.

Also consider adding something like the following to the description of the function above:
`Allows for the array to be byte-aligned, so that loading the uint64_t value
via this function avoids a processor exception due to unaligned access.`

This explains the purpose of the function, which is perhaps more important to document in comments than simply what the function does.

jwhui:
> It's a good function name and a good header location.

@richarddonkin, thanks for your review and comments. I pushed another commit. Please take a look!

PR #4620, opened 2020-03-02
[posix] log version on start

bukepo:


bukepo:
depends on https://github.com/openthread/openthread/pull/4629

PR #4621, opened 2020-03-02
[coap] Fix CoAP option data type in C/C++ API

tlifschitz:
Hi all! In a recent pull request (#4396) new methods have been added to parse CoAP options using an option's number. The type these new functions receive as parameter is `otCoapOptionType`, which is an enumeration found in coap.h:

```
/**
 * CoAP Option Numbers
 */
typedef enum otCoapOptionType
{
    OT_COAP_OPTION_IF_MATCH       = 1,  ///< If-Match
    OT_COAP_OPTION_URI_HOST       = 3,  ///< Uri-Host
    ...
    OT_COAP_OPTION_PROXY_SCHEME   = 39, ///< Proxy-Scheme
    OT_COAP_OPTION_SIZE1          = 60, ///< Size1
} otCoapOptionType;
```
This makes impossible to use vendor-specific options since the underlying type of the enum is implementation-defined, and in my particular case the option number gets casted to an 8 bit integer. I simply replaced the argument type in all protoypes with uint16_t.

### DIFF
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -691,8 +691,7 @@ otError otCoapOptionIteratorInit(otCoapOptionIterator *aIterator, const otMessag
  * @returns A pointer to the first matching option. If no matching option is present NULL pointer is returned.
  *
  */
-const otCoapOption *otCoapOptionIteratorGetFirstOptionMatching(otCoapOptionIterator *aIterator,
-                                                               uint16_t              aOption);
+const otCoapOption *otCoapOptionIteratorGetFirstOptionMatching(otCoapOptionIterator *aIterator, uint16_t aOption);
 
 /**
  * This function returns a pointer to the first option.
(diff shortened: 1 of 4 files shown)


tlifschitz:
@googlebot I signed it!

tlifschitz:
@googlebot I fixed it.

jwhui:
@sjlongland, can you review and provide your thoughts? Thanks.

sjlongland:
Hi, yeah I guess I should have looked closer at `otCoapMessageAppendUintOption`.

The use of that type is more a hint to the developer about what values should be used, but if you want to use your own I was thinking static casting to `otCoapOptionType` was doable.

No gripe with making it a generic `uint16_t` on the parameters.  It shouldn't make much difference.

https://travis-ci.org/openthread/openthread/jobs/657332818?utm_medium=notification&utm_source=github_status might need to be given a kick though as that build seems to have fallen off the perch prematurely.

PR #4626, opened 2020-03-03
[android] add switch to use SPI in Android.mk

Irving-cl:
@bukepo @zhanglongxia


Android.mk
```diff
@@ -70,6 +69,12 @@ else
 OPENTHREAD_COMMON_FLAGS += -DOPENTHREAD_CONFIG_UDP_FORWARD_ENABLE=1
 endif
 
+ifeq ($(USE_RCP_SPI), 1)
```
bukepo:
```suggestion
ifeq ($(USE_OT_RCP_BUS), spi)
```

PR #4629, opened 2020-03-03
[tests] enhance matching pattern

bukepo:
This PR enhances the way to match patterns during tests, so that it can
also support with readline enabled.

### DIFF
--- a/tests/scripts/thread-cert/node.py
+++ b/tests/scripts/thread-cert/node.py
@@ -176,6 +176,17 @@ class Node:
                     raise
 
     def _prepare_pattern(self, pattern):
+        """Build a new pexpect pattern matching line by line.
+
+        Adds lookahead and lookbehind to make each pattern match a whole line,
+        and add 'Done' as the first pattern.
+
+        Args:
+            pattern: a single regex or a list of regex.
+
+        Returns:
+            A list of regex.
+        """
         EXPECT_LINE_FORMAT = r'(?<=[\r\n])%s(?=[\r\n])'
 
         if isinstance(pattern, list):
@@ -186,11 +197,25 @@ class Node:
         return [EXPECT_LINE_FORMAT % 'Done'] + pattern
 
     def _expect_result(self, pattern, *args, **kwargs):
+        """Expect a single matching result.
+
+        The arguments are identical to pexpect.expect().
+
+        Returns:
+            The matched line.
+        """
         results = self._expect_results(pattern, *args, **kwargs)
         assert len(results) == 1
         return results[0]
 
     def _expect_results(self, pattern, *args, **kwargs):
+        """Expect multiple matching results.
+
+        The arguments are identical to pexpect.expect().
+
+        Returns:
+            The matched lines.
+        """
         results = []
         pattern = self._prepare_pattern(pattern)
 
--- a/tests/scripts/thread-cert/node.py
+++ b/tests/scripts/thread-cert/node.py
@@ -40,6 +40,9 @@ import time
 import unittest
 import binascii
 
+EXPECT_LINE_START = r'(?<=[\r\n])'
+EXPECT_LINE_END = r'(?=[\r\n])'
+
 
 class Node:
 
@@ -175,6 +178,31 @@ class Node:
                 if timeout <= 0:
                     raise
 
+    def _prepare_pattern(self, pattern):
+        if isinstance(pattern, list):
+            pattern = [EXPECT_LINE_START + p + EXPECT_LINE_END for p in pattern]
+        else:
+            pattern = [EXPECT_LINE_START + pattern + EXPECT_LINE_END]
+
+        return [EXPECT_LINE_START + r'Done' + EXPECT_LINE_END] + pattern
+
+    def _expect_result(self, pattern, *args, **kwargs):
+        results = self._expect_results(pattern, *args, **kwargs)
+        assert len(results) == 1
+        return results[0]
+
+    def _expect_results(self, pattern, *args, **kwargs):
+        results = []
+        pattern = self._prepare_pattern(pattern)
+
+        while True:
+            if self._expect(pattern, *args, **kwargs):
+                results.append(self.pexpect.match.group(0).decode('utf8'))
+            else:
+                break
+
+        return results
+
     def __init_soc(self, nodeid):
         """ Initialize a System-on-a-chip node connected via UART. """
         import fdpexpect
@@ -264,14 +292,7 @@ class Node:
     def get_commands(self):
         self.send_command('?')
         self._expect('Commands:')
-        commands = []
-        while True:
-            i = self._expect(['Done', r'(\S+)'])
-            if i != 0:
-                commands.append(self.pexpect.match.groups()[0])
-            else:
-                break
-        return commands
+        return self._expect_results(r'\S+')
 
     def set_mode(self, mode):
         cmd = 'mode %s' % mode
@@ -357,11 +378,8 @@ class Node:
 
     def get_addr16(self):
         self.send_command('rloc16')
-        i = self._expect('([0-9a-fA-F]{4})')
-        if i == 0:
-            addr16 = int(self.pexpect.match.groups()[0], 16)
-        self._expect('Done')
-        return addr16
+        rloc16 = self._expect_result(r'[0-9a-fA-F]{4}')
+        return int(rloc16, 16)
 
     def get_router_id(self):
         rloc16 = self.get_addr16()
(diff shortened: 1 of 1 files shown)


bukepo:
depends on https://github.com/openthread/pyspinel/pull/77

tests/scripts/thread-cert/node.py
```diff
@@ -357,53 +378,32 @@ def remove_whitelist(self, addr):
...
+        return self._expect_result('[0-9a-fA-F]{16}')
 
     def get_channel(self):
         self.send_command('channel')
-        i = self._expect(r'(\d+)\r?\n')
-        if i == 0:
-            channel = int(self.pexpect.match.groups()[0])
-        self._expect('Done')
-        return channel
+        return self._expect_result(r'\d+')
```
bukepo:
```suggestion
        return int(self._expect_result(r'\d+'))
```

tests/scripts/thread-cert/node.py
```diff
@@ -40,6 +40,9 @@
 import unittest
 import binascii
 
+EXPECT_LINE_START = r'(?<=[\r\n])'
+EXPECT_LINE_END = r'(?=[\r\n])'
```
wgtdkp: Limit them inside the class or method?

tests/scripts/thread-cert/node.py
```diff
@@ -175,6 +178,31 @@ def _expect(self, pattern, timeout=-1, *args, **kwargs):
...
+    def _expect_result(self, pattern, *args, **kwargs):
+        results = self._expect_results(pattern, *args, **kwargs)
+        assert len(results) == 1
+        return results[0]
+
+    def _expect_results(self, pattern, *args, **kwargs):
+        results = []
+        pattern = self._prepare_pattern(pattern)
+
+        while True:
```
wgtdkp:
```suggestion
        while self._expect(pattern, *args, **kwargs):
            results.append(self.pexpect.match.group(0).decode('utf8'))
```

tests/scripts/thread-cert/node.py
bukepo: They are constants used by multiple methods. If we declare as class member, access them would look like `node._EXPECT_LINE_START`, which looks really weird. So I tend to keep them a module global constants.

tests/scripts/thread-cert/node.py
```diff
@@ -637,7 +575,10 @@ def get_eidcaches(self):
         self.send_command('eidcache')
 
         while True:
-            i = self._expect([r'([a-fA-F0-9\:]+) ([a-fA-F0-9]+)\r?\n', 'Done'])
+            i = self._expect([
+                _EXPECT_LINE_START + r'([a-fA-F0-9\:]+) ([a-fA-F0-9]+)' +
+                _EXPECT_LINE_END, 'Done'
+            ])
```
simonlingoogle: What about something like below
```suggestion
            i = self._expect_results(r'(([a-fA-F0-9\:]+) ([a-fA-F0-9]+)|Done)')
```

Or if we just stop matching  when we found `Done` or `Error: ...`, we can even simplify this to 
`i = self._expect_results(r'([a-fA-F0-9\:]+) ([a-fA-F0-9]+)')`

Thoughts?

tests/scripts/thread-cert/node.py
```diff
@@ -175,6 +175,30 @@ def _expect(self, pattern, timeout=-1, *args, **kwargs):
                 if timeout <= 0:
                     raise
 
+    def _prepare_pattern(self, pattern):
```
wgtdkp: nit: document.

tests/scripts/thread-cert/node.py
```diff
@@ -175,6 +175,55 @@ def _expect(self, pattern, timeout=-1, *args, **kwargs):
...
+        Adds lookahead and lookbehind to make each pattern match a whole line,
+        and add 'Done' as the first pattern.
+
+        Args:
+            pattern: a single regex or a list of regex.
+
+        Returns:
+            A list of regex.
+        """
+        EXPECT_LINE_FORMAT = r'(?<=[\r\n])%s(?=[\r\n])'
```
simonlingoogle: This regex is not totally correct. If `pattern` == `''` (empty string) , then the line prepared pattern will be `r'(?<=[\r\n])(?=[\r\n])'`. We are expecting it to match an empty line between 2 EOLs, e.x. `'\n\n'`, `'\r\n\r\n'`. However, it will also match `'\r\n'`, which is not an empty line between two EOLs.

tests/scripts/thread-cert/node.py
```diff
@@ -175,6 +175,55 @@ def _expect(self, pattern, timeout=-1, *args, **kwargs):
...
+
+        if isinstance(pattern, list):
+            pattern = [EXPECT_LINE_FORMAT % p for p in pattern]
+        else:
+            pattern = [EXPECT_LINE_FORMAT % pattern]
+
+        return [EXPECT_LINE_FORMAT % 'Done'] + pattern
```
simonlingoogle: Should we also add `Error:.*` along side with `Done` ?

tests/scripts/thread-cert/node.py
```diff
@@ -636,14 +597,11 @@ def get_eidcaches(self):
...
-        while True:
-            i = self._expect([r'([a-fA-F0-9\:]+) ([a-fA-F0-9]+)\r?\n', 'Done'])
-            if i == 0:
-                eid = self.pexpect.match.groups()[0].decode("utf-8")
-                rloc = self.pexpect.match.groups()[1].decode("utf-8")
-                eidcaches.append((eid, rloc))
-            elif i == 1:
-                break
+        pattern = self._prepare_pattern(r'([a-fA-F0-9\:]+) ([a-fA-F0-9]+)')
+        while self._expect(pattern):
```
simonlingoogle: I think in long term we can eliminate all calls to `_expect`, leaving only `_expect_results` and other methods that are based on `_expect_results`.
For example, current calls to `self._expect('Done')` are not strictly correct. It is just expecting string `Done`, however the correct behavior is to expect a line of `Done`. 
Thoughts.

tests/scripts/thread-cert/node.py
```diff
@@ -805,18 +759,9 @@ def panid_query(self, panid, mask, ipaddr):
...
-                r'\|\s(\S+)\s+\|\s(\S+)\s+\|\s([0-9a-fA-F]{4})\s\|\s([0-9a-fA-F]{16})\s\|\s(\d+)\r?\n',
-                'Done',
-            ])
-            if i == 0:
-                results.append(self.pexpect.match.groups())
-            else:
-                break
-
-        return results
+        return self._expect_results(
```
simonlingoogle: Is the type of return value changed from `List[Tuple[str]]` to `List[str]`?

tests/scripts/thread-cert/node.py
bukepo: Do we have a use case to expect a empty line? There is a limitation that look behind has to be with fixed size. And with readline enabled, there are mixed \r\n, \n, even \r as new line delimiter.

tests/scripts/thread-cert/node.py
bukepo: I don't think so. Error is not expected on success, and the test should fail for unexpected results.

tests/scripts/thread-cert/node.py
bukepo: Agree. Maybe replacing send_command with execute_command and let it return expected results is even cleaner. But I think it's beyond the purpose of this PR.

tests/scripts/thread-cert/node.py
simonlingoogle: Do you think `r'(?<=[\n])%s(?=\r?\n)' ` will be better? Since `node` is reading from CLI executive what is written by us, we can be sure that newline always ends with `\n`.

tests/scripts/thread-cert/node.py
simonlingoogle: > I don't think so. Error is not expected on success, and the test should fail for unexpected results.

Agree that we should fail for `Error: ...`. That was what I mean by adding `Error: ...`. Maybe we can enhance it later.

tests/scripts/thread-cert/node.py
bukepo: IIRC when readline is enabled, sometimes a line starts with a `\r`.

simonlingoogle: SGTM since we have no use case to expect an empty line yet.

PR #4631, opened 2020-03-04
[coap] enum type conversion to int in otCoapBlockSizeFromExponent

piotrkoziar:
There is "enumerated type mixed with another enumerated type" error during the iar compilation. This PR fixes that.

### DIFF
--- a/src/core/api/coap_api.cpp
+++ b/src/core/api/coap_api.cpp
@@ -112,7 +112,7 @@ otError otCoapMessageAppendUriPathOptions(otMessage *aMessage, const char *aUriP
 
 uint16_t otCoapBlockSizeFromExponent(otCoapBlockSize aSize)
 {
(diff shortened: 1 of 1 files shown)


src/core/api/coap_api.cpp
```diff
@@ -112,7 +112,7 @@ otError otCoapMessageAppendUriPathOptions(otMessage *aMessage, const char *aUriP
 
 uint16_t otCoapBlockSizeFromExponent(otCoapBlockSize aSize)
 {
-    return static_cast<uint16_t>(1 << (aSize + Coap::Message::kBlockSzxBase));
+    return static_cast<uint16_t>(1 << (static_cast<int>(aSize) + static_cast<int>(Coap::Message::kBlockSzxBase)));
```
jwhui: nit: should we cast to `uint8_t` to better reflect the values we expect?

PR #4634, opened 2020-03-04
[spinel] change spinel.h output dir include/spinel

FiveDimensions:


src/lib/spinel/Makefile.am
```diff
@@ -101,6 +107,7 @@ spinel_test_CFLAGS                                = \
     -I$(top_srcdir)/include                         \
     $(NULL)
 
+
```
bukepo:
```suggestion
```

src/lib/spinel/Makefile.am
```diff
@@ -59,8 +59,14 @@ COMMON_SOURCES                                    = \
     spinel_encoder.cpp                              \
     $(NULL)
 
+ot_spinel_headers                                   = \
+    spinel.h                                       \
```
bukepo: not aligned.

Issue #4611, opened 2020-02-28
[error] semantic of OT_ERROR_PARSE and OT_ERROR_INVALID_ARGS


wgtdkp:
Hi guys, this PR is opened for discussion of the semantic of `otError`, specifically, `OT_ERROR_PARSE` and `OT_ERROR_INVALID_ARGS`.

https://github.com/openthread/openthread/blob/c36fa453f1726f4e5f810264ede54a56b75252aa/include/openthread/error.h#L90-L98

It says that `OT_ERROR_PARSE` stands  for both message and arguments parsing error, while **I think `OT_ERROR_PARSE` should only represent message parsing error**. **Because arguments parsing error is a sub-category of invalid arguments error while message parsing error is another error category with the same level of invalid arguments error.**

Current semantic leading to confusing that we can't tell if there are problems with the arguments or we has received malformed messages. `INVALID_ARGS` errors are often caused by the user who doesn't understand the API or CLI argument requirements well. When there is argument parsing error, returning `OT_ERROR_INVALID_ARGS` clearly tells the caller/user of the API/CLI that they are providing bad arguments and the user got a chance to fix it, while returning `OT_ERROR_PARSE` doesn't tell whether the user is using the API/CLI wrong, or there is internal message processing error.

To make the conclusion, I think arguments parsing error is invalid arguments error (errors that can be avoided) and must be corrected by the user of the API or CLI, while message parsing error has its own category (errors that can't be avoided) and openthread should continue working when it happens. Thus, **I think we shouldn't mix them into `OT_ERROR_PARSE`**.

What to do:
1. openthread doesn't give details of each kind of error, we may define another error `OT_ERROR_INVALID_ARGS_PARSE` if it helps on resolving the actual arguments error (but I don't think it is necessary).
2. For openthread CLI, a user may input a command that cannot be recognized by the CLI. it is better to define and return `OT_ERROR_INVALID_COMMAND` other than  `OT_ERROR_PARSE`. Conceptually, `INVALID_COMMAND` errors are one kind of error which should and can be corrected by the user (just like `INVALID_ARGS`).
3. For API, `OT_ERROR_INVALID_ARGS` should be returned for arguments parsing error.

I'd like to hear from more.
Thoughts? @jwhui @librasungirl @bukepo @abtink  @simonlingoogle @gjc13 @Irving-cl

Irving-cl:
It makes sense to me. I think **OT_ERROR_INVALID_ARGS** is better for API/CLI.

[...]

PR #4639, opened 2020-03-05
[cli] return OT_ERROR_INVALID_ARGS for invalid arguments

wgtdkp:
This PR addresses https://github.com/openthread/openthread/issues/4611 with three changes:
1. Remove argument parsing failure from semantic of `OT_ERROR_PARSE`.
2. Add `OT_ERROR_INVALID_COMMAND` to indicate invalid CLI command error.
3. return `OT_ERROR_INVALID_ARGS` for invalid CLI arguments.

### DIFF
--- a/include/openthread/error.h
+++ b/include/openthread/error.h
@@ -88,7 +88,7 @@ typedef enum otError
     OT_ERROR_BUSY = 5,
 
     /**
-     * Failed to parse message or arguments.
+     * Failed to parse message.
      */
     OT_ERROR_PARSE = 6,
 
@@ -222,6 +222,11 @@ typedef enum otError
      * The link margin was too low.
      */
     OT_ERROR_LINK_MARGIN_LOW = 34,
+
+    /**
+     * Input (CLI) command is invalid.
+     */
+    OT_ERROR_INVALID_COMMAND = 35,
 
     /**
      * The number of defined errors.
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -1015,7 +1015,7 @@ void Interpreter::ProcessDiscover(int argc, char *argv[])
 
     if (argc > 0)
     {
-        SuccessOrExit(error = ParseLong(argv[0], value) == OT_ERROR_NONE);
+        SuccessOrExit(error = ParseLong(argv[0], value));
         VerifyOrExit((0 <= value) && (value < static_cast<long>(sizeof(scanChannels) * CHAR_BIT)),
                      error = OT_ERROR_INVALID_ARGS);
         scanChannels = 1 << value;
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -2602,6 +2602,7 @@ otError Interpreter::ProcessRouteRemove(int argc, char *argv[])
 
     {
         unsigned long length;
+
         SuccessOrExit(error = ParseUnsignedLong(prefixLengthStr, length));
         prefix.mLength = static_cast<uint8_t>(length);
     }
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -361,14 +361,14 @@ otError Interpreter::ParseLong(char *aString, long &aLong)
 {
     char *endptr;
     aLong = strtol(aString, &endptr, 0);
-    return (*endptr == '\0') ? OT_ERROR_NONE : OT_ERROR_PARSE;
+    return (*endptr == '\0') ? OT_ERROR_NONE : OT_ERROR_INVALID_ARGS;
 }
 
 otError Interpreter::ParseUnsignedLong(char *aString, unsigned long &aUnsignedLong)
 {
     char *endptr;
     aUnsignedLong = strtoul(aString, &endptr, 0);
-    return (*endptr == '\0') ? OT_ERROR_NONE : OT_ERROR_PARSE;
+    return (*endptr == '\0') ? OT_ERROR_NONE : OT_ERROR_INVALID_ARGS;
 }
 
 otError Interpreter::ParsePingInterval(const char *aString, uint32_t &aInterval)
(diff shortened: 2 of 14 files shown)


include/openthread/error.h
```diff
@@ -223,6 +223,11 @@ typedef enum otError
      */
     OT_ERROR_LINK_MARGIN_LOW = 34,
 
+    /**
+     * Input (CLI) command is invalid.
+     */
+    OT_ERROR_INVALID_COMMAND = 35,
```
simonlingoogle: I think there are many other command parsing code that we should change from returning `OT_ERROR_PARSE` to returning `OT_ERROR_INVALID_COMMAND`. Still `work-in-progress` ?

wgtdkp: Thanks for mention, I think sub-commands like `udp send` should also return `OT_ERROR_COMMAND`. Will do.

src/cli/cli.cpp
```diff
@@ -1121,12 +1121,13 @@ void Interpreter::ProcessEidCache(int argc, char *argv[])
...
     OT_UNUSED_VARIABLE(argc);
     OT_UNUSED_VARIABLE(argv);
+    otError error = OT_ERROR_NONE;
 
     otEidCacheEntry entry;
 
     for (uint8_t i = 0;; i++)
     {
-        SuccessOrExit(otThreadGetEidCacheEntry(mInstance, i, &entry));
+        SuccessOrExit(error = otThreadGetEidCacheEntry(mInstance, i, &entry));
```
abtink: I think not checking error here is intentional, otherwise will always get to `AppendResult()` with an error
The `for` loop keeps going till we get an error from `otThreadGetEidCacheEntry()` which indicates the index `i` is out of bound and we exit the loop.
https://github.com/openthread/openthread/blob/a75b412ecfe3e3c286c57517b08171eb05fef0c4/include/openthread/thread_ftd.h#L509-L520

Just as a note, I am changing the public API for address cache table from this open PR https://github.com/openthread/openthread/pull/4599 which will use an iterator and return `OT_ERROR_NOT_FOUND` when we reach last cache entry in the list.

src/cli/cli.cpp
```diff
@@ -2391,8 +2386,9 @@ otError Interpreter::ProcessPrefixList(void)
 {
     otNetworkDataIterator iterator = OT_NETWORK_DATA_ITERATOR_INIT;
     otBorderRouterConfig  config;
+    otError               error;
 
-    while (otBorderRouterGetNextOnMeshPrefix(mInstance, &iterator, &config) == OT_ERROR_NONE)
+    while ((error = otBorderRouterGetNextOnMeshPrefix(mInstance, &iterator, &config)) == OT_ERROR_NONE)
```
abtink: Same situation here. I don't think we want to return `error` from `ProcessPrefixList()`. 
The `otBorderRouterGetNextOnMeshPrefix()` returns `OT_ERROR_NOT_FOUND` to indicate it reached end of the list.

src/cli/cli.cpp
```diff
@@ -2612,8 +2603,9 @@ otError Interpreter::ProcessRouteList(void)
 {
     otNetworkDataIterator iterator = OT_NETWORK_DATA_ITERATOR_INIT;
     otExternalRouteConfig config;
+    otError               error;
 
-    while (otBorderRouterGetNextRoute(mInstance, &iterator, &config) == OT_ERROR_NONE)
+    while ((error = otBorderRouterGetNextRoute(mInstance, &iterator, &config)) == OT_ERROR_NONE)
```
abtink: Same here.

src/cli/cli.cpp
```diff
@@ -2549,11 +2541,10 @@ otError Interpreter::ProcessRouteAdd(int argc, char *argv[])
 
     SuccessOrExit(error = otIp6AddressFromString(argv[argcur], &config.mPrefix.mPrefix));
 
-    config.mPrefix.mLength = static_cast<uint8_t>(strtol(prefixLengthStr, &endptr, 0));
-
-    if (*endptr != '\0')
     {
-        ExitNow(error = OT_ERROR_PARSE);
+        unsigned long length;
+        SuccessOrExit(error = ParseUnsignedLong(prefixLengthStr, length));
```
bukepo:
```suggestion

        SuccessOrExit(error = ParseUnsignedLong(prefixLengthStr, length));
```

src/cli/cli.cpp
```diff
@@ -2609,11 +2599,10 @@ otError Interpreter::ProcessRouteRemove(int argc, char *argv[])
 
     SuccessOrExit(error = otIp6AddressFromString(argv[argcur], &prefix.mPrefix));
 
-    prefix.mLength = static_cast<uint8_t>(strtol(prefixLengthStr, &endptr, 0));
-
-    if (*endptr != '\0')
     {
-        ExitNow(error = OT_ERROR_PARSE);
+        unsigned long length;
```
bukepo:
```suggestion
        unsigned long length;

```

src/cli/cli.hpp
```diff
@@ -153,8 +153,9 @@ class Interpreter
      * @param[in]   aAllowTruncate  TRUE if @p aBinLength may be less than what is required
      *                              to convert @p aHex to binary representation, FALSE otherwise.
      *
+     * @returns -1         Indicates @p aHex is a invalid hex string.
+     * @returns Otherwise  The number of bytes in the binary representation.
```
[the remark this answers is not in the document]
librasungirl: didn't see `returns Otherwise`, here you marked as resolved, are you going to apply the suggestion or reject it? @wgtdkp

src/cli/cli.cpp
```diff
@@ -1015,7 +1015,7 @@ void Interpreter::ProcessDiscover(int argc, char *argv[])
 
     if (argc > 0)
     {
-        SuccessOrExit(error = ParseLong(argv[0], value));
+        VerifyOrExit(ParseLong(argv[0], value) == OT_ERROR_NONE, error = OT_ERROR_INVALID_ARGS);
```
jwhui: What about reverting to the original `SuccessOrExit()` given that `ParseLong` will also return `OT_ERROR_INVALID_ARGS`?

src/cli/cli.cpp
```diff
@@ -1015,7 +1015,7 @@ void Interpreter::ProcessDiscover(int argc, char *argv[])
 
     if (argc > 0)
     {
-        SuccessOrExit(error = ParseLong(argv[0], value));
+        SuccessOrExit(error = ParseLong(argv[0], value) == OT_ERROR_NONE);
```
jwhui:
```suggestion
        SuccessOrExit(error = ParseLong(argv[0], value));
```

PR #4645, opened 2020-03-06
[network-data] fix stable version update issue for service

librasungirl:


src/core/thread/network_data_leader_ftd.cpp
```diff
@@ -674,11 +676,12 @@ bool Leader::IsStableUpdated(uint8_t *aTlvs, uint8_t aTlvsLength, uint8_t *aTlvs
                             NetworkDataTlv *curServerBase = serviceBase->GetSubTlvs();
                             NetworkDataTlv *endServerBase = serviceBase->GetNext();
 
-                            while (curServerBase <= endServerBase)
+                            while (curServerBase < endServerBase)
```
librasungirl: note: the extra `=` may cause exit when checking the length, thus no chance for below to flag stable update if no matching server.

https://github.com/openthread/openthread/blob/master/src/core/thread/network_data_leader_ftd.cpp#L695-L698

PR #4647, opened 2020-03-08
[cli] add `preferrouterid` command

librasungirl:


src/cli/cli.cpp
```diff
@@ -2480,6 +2481,20 @@ void Interpreter::ProcessPrefix(int argc, char *argv[])
...
 
 #if OPENTHREAD_FTD
+void Interpreter::ProcessPreferRouterId(int argc, char *argv[])
+{
+    otError error = OT_ERROR_NONE;
+    long    value;
+
+    VerifyOrExit(argc > 0, error = OT_ERROR_INVALID_ARGS);
+
+    SuccessOrExit(error = ParseLong(argv[0], value));
```
simonlingoogle: Use `ParseUnsignedLong`?

librasungirl: updated. thanks.

src/cli/cli.cpp
```diff
@@ -2480,6 +2481,19 @@ void Interpreter::ProcessPrefix(int argc, char *argv[])
 #endif // OPENTHREAD_CONFIG_BORDER_ROUTER_ENABLE
 
 #if OPENTHREAD_FTD
+void Interpreter::ProcessPreferRouterId(int argc, char *argv[])
+{
+    otError       error = OT_ERROR_NONE;
+    unsigned long value;
+
+    VerifyOrExit(argc == 1, error = OT_ERROR_INVALID_ARGS);
```
simonlingoogle: show the current `prefered router id` when `argc == 0` ?

librasungirl: may add it in a separate PR when there is use scenario?

simonlingoogle: I think it is a convention for commands to show the current value when no argument is given. Better to have if won't involve much effort. 
But it's fine to me to add this feature when it's really used.

Issue #4530, opened 2020-02-05
[harness-automation] v50 select case doesn't work to choose specific test cases


doublemis1:
**Describe the bug**
During execution test ex. Router 5.1.7 autothreadharness crash because of `action_chains.perform()` in harness_case.py file.
It happens every time if any previous test case is not available (not display) in test case tree

**To Reproduce**
Information to reproduce the behavior, including:
1. Openthread master
2. GRL 2.0.0 (v50)
3. Google chrome 80
4. start.bat -t -r 3 --blacklist rfbox_cases -p "Router_5_1_7*.py"

**Console/log output**
```
ERROR:root:case[Router_5_1_7] error
--->    INFO : Traceback (most recent call last):
--->    INFO :   File "C:\Python27\lib\unittest\case.py", line 329, in run
--->    INFO :     testMethod()
--->    INFO :   File "autothreadharness\harness_case.py", line 1106, in test
--->    INFO :     self._select_case(self.role, self.case)
--->    INFO :   File "autothreadharness\harness_case.py", line 839, in _select_case
--->    INFO :     action_chains.perform()
--->    INFO :   File "C:\Python27\lib\site-packages\selenium\webdriver\common\action_chains.py", line 80, in perform
--->    INFO :     self.w3c_actions.perform()
--->    INFO :   File "C:\Python27\lib\site-packages\selenium\webdriver\common\actions\action_builder.py", line 76, in perform
--->    INFO :     self.driver.execute(Command.W3C_ACTIONS, enc)
--->    INFO :   File "C:\Python27\lib\site-packages\selenium\webdriver\remote\webdriver.py", line 321, in execute
--->    INFO :     self.error_handler.check_response(response)
--->    INFO :   File "C:\Python27\lib\site-packages\selenium\webdriver\remote\errorhandler.py", line 242, in check_response
--->    INFO :     raise exception_class(message, screen, stacktrace)
--->    INFO : JavascriptException: Message: javascript error: Failed to execute 'elementsFromPoint' on 'Document': The provided double value is non-finite.
--->    INFO :   (Session info: chrome=80.0.3987.87)
```

**Additional context**
I am not sure if it is problem with new certification tool version or just my environment.

**Fix**
In method `_select_case` add click checkbox with id `display-all-devices`, before getting elements from test case tree.
```
self._browser.find_element_by_id('display-all-devices').click()
checkbox = None
[...]

PR #4648, opened 2020-03-08
[harness-automation] fix case selection issue when using Chrome 80

FiveDimensions:


jwhui:
@FiveDimensions, can we move this PR forward? Please rebase on latest master as well.

tools/harness-automation/autothreadharness/harness_case.py
```diff
@@ -882,14 +882,18 @@ def _select_case(self, role, case):
...
-            action_chains = ActionChains(self._browser)
-            action_chains.move_to_element(elem)
-            action_chains.perform()
-            logger.debug(elem.text)
-            if finder.match(elem.text) or finder_dotted.match(elem.text):
-                parent = elem.find_element_by_xpath('..')
-                checkbox = parent.find_element_by_class_name('tree-checkbox')
-                break
+            if elem.text:
+                self._browser.execute_script('arguments[0].scrollIntoView();',
```
librasungirl, line 886: @FiveDimensions would  you please add comments to illustrate when `elem.text` is null and what `arguments[0]` is?

Also noticed your comment of issue #4530:

> This issue happens on the testbeds with mixed vendor devices using Chrome 80 and Chromedriver 80. The ActionChains.perform() will throw an exception when ActionChains move to the element without text. 

Would you please help to explain why it only happens on mixed testbed? what's the difference compared with CV testbed?

If possible, why it occurs when using the Chrome 80?

FiveDimensions: @librasungirl "arguments[0].scrollIntoView();" is a javascript and will take the reference of the "elem" as arguments[0] to scroll the window to it. 
There are 2 issues using Chrome and chromedriver 80:
1. Cases will not display when the selected vendor devices are not enough, their web classes are tagged as "NotToBeEnabled" and their web elements' text is null. When elem.text is null, move_to_element(elem) and perform() will throw exceptions. 
2. move_to_element(elem) will not work when the "elem" class is not displayed in the current window and need to scroll to it.
 For case Router 5.1.7,  it needs 11 devices,  in CV testbed, all previous cases (Router 5.1.1, Router 5.1.2 ....) in the list can display since their needed device numbers are less than 11 which means move_to_element(elem) and perform() work. In IV (mixed vendor) testbed, several previous cases do not show in the list since the vendor devices are not enough, which will cause the issue #1.

jwhui: I read @librasungirl's comment as adding a comment/documentation to the code - @librasungirl, can you confirm?

librasungirl: yes, add comments into code for readability.

librasungirl: Thanks for the explanation. Still I have one question:
> Cases will not display when the selected vendor devices are not enough, their web classes are tagged as "NotToBeEnabled" and their web elements' text is null. 

For CV testbed, there are also chances that the reference devices may be not enough, right? it should be a common issue for both IV/CV testbeds?

FiveDimensions: Yes, it should be

tools/harness-automation/autothreadharness/harness_case.py
```diff
@@ -882,14 +882,22 @@ def _select_case(self, role, case):
...
         for elem in elems:
-            action_chains = ActionChains(self._browser)
-            action_chains.move_to_element(elem)
-            action_chains.perform()
-            logger.debug(elem.text)
-            if finder.match(elem.text) or finder_dotted.match(elem.text):
-                parent = elem.find_element_by_xpath('..')
-                checkbox = parent.find_element_by_class_name('tree-checkbox')
-                break
+            # perform() will throw exceptions if elem.text is null since Chrome
```
librasungirl, line 885:
```suggestion
            # elem.txt might be null when the required reference devices could
            # not be met (either due to quantity or the type) for specific test.
            # perform() will throw exceptions if elem.text is null since Chrome
```

tools/harness-automation/autothreadharness/harness_case.py
```diff
@@ -882,14 +882,24 @@ def _select_case(self, role, case):
...
-            if finder.match(elem.text) or finder_dotted.match(elem.text):
-                parent = elem.find_element_by_xpath('..')
-                checkbox = parent.find_element_by_class_name('tree-checkbox')
-                break
+            # elem.txt might be null when the required reference devices could
+            # not be met (either due to the quantity or the type) for specific test.
+            # perform() will throw exceptions if elem.text is null since Chrome
+            # and chromedriver 80. If elem is not shown in current case list
+            # window, move_to_element() will not work either.
+            if elem.text:
```
bukepo, line 890: To avoid indent.
```suggestion
            if not elem.text:
                continue
```

tools/harness-automation/autothreadharness/harness_case.py
```diff
@@ -882,14 +882,24 @@ def _select_case(self, role, case):
...
+                # execute a javascript to scroll the window to the elem
+                self._browser.execute_script('arguments[0].scrollIntoView();',
+                                             elem)
+                action_chains = ActionChains(self._browser)
+                action_chains.move_to_element(elem)
```
bukepo, line 895: Is this still necessary with line 892 added?

FiveDimensions: no, 
ERROR:root:case[Router_5_1_11] error
  File "C:\GRL\Thread1.1\Python27\lib\unittest\case.py", line 329, in run
    testMethod()
  File "autothreadharness\harness_case.py", line 1177, in test
    self._select_case(self.role, self.case)
  File "autothreadharness\harness_case.py", line 899, in _select_case
    raise Exception('Failed to find the case')
Exception: Failed to find the case

PR #4652, opened 2020-03-09
[nrf528xx] add volatile mark to nrf uart read/write buffers

gjc13:
These buffers are touched in volatile and should be marked as volatile.

### DIFF
--- a/examples/platforms/nrf528xx/src/transport/uart.c
+++ b/examples/platforms/nrf528xx/src/transport/uart.c
@@ -56,16 +56,16 @@ bool sUartEnabled = false;
 /**
  *  UART TX buffer variables.
  */
-static const uint8_t *sTransmitBuffer = NULL;
-static uint16_t       sTransmitLength = 0;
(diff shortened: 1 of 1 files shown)


examples/platforms/nrf528xx/src/uart.c
```diff
@@ -56,16 +56,16 @@ bool sUartEnabled = false;
...
 
 /**
  *  UART RX ring buffer variables.
  */
-static uint8_t  sReceiveBuffer[UART_RX_BUFFER_SIZE];
-static uint16_t sReceiveHead = 0;
-static uint16_t sReceiveTail = 0;
+static volatile uint8_t  sReceiveBuffer[UART_RX_BUFFER_SIZE];
+static volatile uint16_t sReceiveHead = 0;
+static volatile uint16_t sReceiveTail = 0;
```
zhanglongxia: `sReceiveTail` is not changed in the ISR, do we need to add `volatile` before it?

PR #4655, opened 2020-03-09
[mle] only register preferred addresses for MTD

librasungirl:


src/core/thread/mle.cpp
```diff
@@ -1385,7 +1385,8 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
     for (const Ip6::NetifUnicastAddress *addr = Get<ThreadNetif>().GetUnicastAddresses(); addr; addr = addr->GetNext())
     {
         if (addr->GetAddress().IsLinkLocal() || IsRoutingLocator(addr->GetAddress()) ||
-            IsAnycastLocator(addr->GetAddress()) || addr->GetAddress() == GetMeshLocal64())
+            IsAnycastLocator(addr->GetAddress()) || addr->GetAddress() == GetMeshLocal64() ||
+            !addr->IsPreferred() /*skip deprecated address*/)
```
abtink: Who is responsible for setting the preferred flag on an address?
Is it required to not register the non-preferred addresses or is it an optimization?

jwhui: - For SLAAC addresses, the "preferred" flag is set based on the on-mesh prefix information contained in the network data. https://github.com/openthread/openthread/blob/c54238fdd612cce027bc7b095954f0229b031608/src/core/utils/slaac_address.cpp#L232

- For DHCPv66 addresses, the "preferred" flag is set based on the DHCPv6 response.
https://github.com/openthread/openthread/blob/c54238fdd612cce027bc7b095954f0229b031608/src/core/net/dhcp6_client.cpp#L593

- For manual addresses, the "preferred" flag is set by the application.

abtink: Thanks. 

I feel this changes the existing behavior and we want to be careful about its impact. In particular, about the manual addresses and if in all (e.g., our internal) use-cases the preferred flag is set by application. If not this will impact behavior (may require a change in application code?).

I checked the `wpantund` code and it does indeed set it correctly for manually added addresses and for the added on-mesh-prefixes. Though `wpantund` is not really intended/used for MTD type devices.

librasungirl: Thanks for the feedbacks.

Besides the scenarios @jwhui mentioned, another scenario in my thought is that for the unicast addresses registered in Address Registration Tlv, if not echoed back (excluding MLEID), the unicast addresses should be marked as `not preferred` as the parent would not help to respond address query or forward to it.

jwhui: Looking at this again, I think we do want to register deprecated addresses (i.e. valid but not preferred).

From [RFC 4862 Section 2](https://tools.ietf.org/html/rfc4862#section-2):
```

Thoughts?
   deprecated address -  An address assigned to an interface whose use
      is discouraged, but not forbidden.  A deprecated address should no
      longer be used as a source address in new communications, but
      packets sent from or to deprecated addresses are delivered as
      expected.  A deprecated address may continue to be used as a
      source address in communications where switching to a preferred
      address causes hardship to a specific upper-layer activity (e.g.,
      an existing TCP connection).
```

librasungirl: 👍make sense!
Will close this PR.

PR #4657, opened 2020-03-09
[joiner-router] prepare Joiner Entrust msg after the delay

abtink:

**[joiner-router] move JoinerEntrustMetadata as private inner type of JoinerRouter**

**[joiner-router] prepare Joiner Entrust msg after the delay**
    
 This commit changes how the tx of Joiner Entrust is delayed. Instead of preparing the message (and appending metadata to it) and enqueueing it to be sent later after a delay, this commit changes the code such that only the metadata is saved and the Joiner Entrust message itself is prepared after the delay (when it is being sent).

----
I think it would be easier to check/review the commits separately.

### DIFF
--- a/src/core/meshcop/joiner_router.cpp
+++ b/src/core/meshcop/joiner_router.cpp
@@ -234,15 +234,120 @@ exit:
 
 otError JoinerRouter::DelaySendingJoinerEntrust(const Ip6::MessageInfo &aMessageInfo, const uint8_t *aKek)
 {
-    otError          error;
-    Coap::Message *  message = NULL;
-    Ip6::MessageInfo messageInfo;
-    Dataset          dataset(MeshCoP::Tlv::kActiveTimestamp);
+    otError               error   = OT_ERROR_NONE;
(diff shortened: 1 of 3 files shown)


src/core/meshcop/joiner_router.cpp
```diff
@@ -232,18 +232,15 @@ void JoinerRouter::HandleRelayTransmit(Coap::Message &aMessage, const Ip6::Messa
     }
 }
 
-otError JoinerRouter::DelaySendingJoinerEntrust(const Ip6::MessageInfo &aMessageInfo, const uint8_t *aKek)
+Coap::Message *JoinerRouter::PrepareJoinerEntrustMessage(void)
```
jwhui: nit: how about moving `PrepareJoinerEntrustMessage()` closer to `SendJoinerEntrust()` where it is used?

abtink: Sounds good. Kept it here to minimize the git diff. Will move it.

PR #4658, opened 2020-03-09
[toranj] make test-13-off-mesh-route more robust

abtink:
This commit changes the test to wait till network data is updated
on all nodes and they all see all the added off-mesh routes (instead
of waiting for fixed 0.5 seconds).

### DIFF
--- a/tests/toranj/test-013-off-mesh-route-traffic.py
+++ b/tests/toranj/test-013-off-mesh-route-traffic.py
@@ -107,6 +107,9 @@ sed2.set(wpan.WPAN_POLL_INTERVAL, '1500')
 # -----------------------------------------------------------------------------------------------------------------------
 # Test implementation
 
+WAIT_TIME = 10
+NUM_ROUTES = 3
+NUM_ROUTES_LOCAL = 1
 ON_MESH_PREFIX = "fd00:1234::"
 OFF_MESH_ROUTE_1 = "fd00:abba::"
 OFF_MESH_ROUTE_2 = "fd00:cafe::"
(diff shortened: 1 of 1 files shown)


tests/toranj/test-013-off-mesh-route-traffic.py
```diff
@@ -132,7 +134,23 @@
 fed1.add_route(OFF_MESH_ROUTE_3)
 fed1.add_ip6_address_on_interface(OFF_MESH_ADDR_3)
 
-time.sleep(0.5)
+# Wait till network data is updated on all nodes and they all see all
```
jwhui: Comment says "all nodes", yet `fed1` is not included in the check below. Should it be?

tests/toranj/test-013-off-mesh-route-traffic.py
```diff
@@ -132,7 +134,23 @@
...
+def check_off_mesh_routes():
+    verify(
+        len(wpan.parse_list(r1.get(wpan.WPAN_THREAD_OFF_MESH_ROUTES))) ==
+        NUM_ROUTES + 1)
+    verify(
+        len(wpan.parse_list(r2.get(wpan.WPAN_THREAD_OFF_MESH_ROUTES))) ==
+        NUM_ROUTES + 1)
+    verify(
+        len(wpan.parse_list(sed2.get(wpan.WPAN_THREAD_OFF_MESH_ROUTES))) ==
+        NUM_ROUTES)
```
jwhui: The above checks have `NUM_ROUTES + 1`, why not here?

tests/toranj/test-013-off-mesh-route-traffic.py
abtink: We can add it but I excluded  it mainly since we only care about the nodes that will be initiating tx and sending traffic to one of the off-mesh routes. Let me update the comment to be clear about this.

tests/toranj/test-013-off-mesh-route-traffic.py
abtink: If a node itself adds a route, the route entry will be seen twice in `"Thread:OffMeshRoutes"` wpantund list (one time as part of leader/network-wide network data and again as part of the local network data ("IsLocal" flag is set)). The `+1` is to account for that.  In the test network we add off-mesh routes on `r1`, `r2` and `fed` but not on `sed1`. So when things are synced and stable, `r1 `and `r2` should see 4 entries in their list while `sed` should only see `3`.

jwhui: How about having a named value like NUM_ROUTES_LOCAL instead of magic number.

PR #4659, opened 2020-03-10
[ncp] notify wpantund when RLOC changes

simonlingoogle:
Notify `wpantund` when `RLOC` changes so that `wpantund` has the chance to update `RLOC` address.


src/ncp/ncp_base_mtd.cpp
```diff
@@ -3657,6 +3657,8 @@ void NcpBase::ProcessThreadChangedFlags(void)
         {OT_CHANGED_THREAD_PANID, SPINEL_PROP_MAC_15_4_PANID},
         {OT_CHANGED_THREAD_NETWORK_NAME, SPINEL_PROP_NET_NETWORK_NAME},
         {OT_CHANGED_THREAD_EXT_PANID, SPINEL_PROP_NET_XPANID},
+        {OT_CHANGED_THREAD_RLOC_ADDED, SPINEL_PROP_IPV6_ADDRESS_TABLE},
```
librasungirl: update the comment to reflect this
https://github.com/openthread/openthread/blob/master/src/ncp/ncp_base_mtd.cpp#L3742

bukepo: Since we are touching this part, how about aggregating: OT_CHANGED_THREAD_RLOC_ADDED and OT_CHANGED_THREAD_RLOC_REMOVED into a single OT_CHANGED_THREAD_RLOC

simonlingoogle: > update the comment to reflect this

Sure. Changed to `OT_CHANGED_SECURITY_POLICY`.

simonlingoogle: > Since we are touching this part, how about aggregating: OT_CHANGED_THREAD_RLOC_ADDED and OT_CHANGED_THREAD_RLOC_REMOVED into a single OT_CHANGED_THREAD_RLOC

We also have `OT_CHANGED_IP6_ADDRESS_REMOVED` and `OT_CHANGED_IP6_ADDRESS_ADDED`. Strictly speaking, we should just use `OT_CHANGED_IP6_ADDRESS` because I don't see why `RLOC` is not an `IP6 ADDRESS`. 
The purpose of this PR is to just update RLOC on `wpantund`. Maybe we can enhance it in another PR. Thoughts?

abtink:
@simonlingoogle  the RLOC IP address is actually by default filtred by `wpantund` (not added on linux interface). The idea is that RLOC is intended for use by the Thread layer and not application side so not needed on `wpan` interface.  The filtering behavior on wpantund can be changed through the wpan `"Thread:Config:FilterRLOCAddresses"` property (by default it is on). Same is true with ALOC addresses. What is the use-case that requires the RLOC address on `wpantund`?

librasungirl:
@abtink , 
The filter switch only lies in wpantund side now, not passed into NCP. So here it should be reasonable to pass all addresses changes to wpantund side and let it to decide whether or not add RLOC on `wpan` interface according to the switch.

Maybe another possible option is to pass the switch configuration to NCP, and let the NCP decides the action according to the switch.


More background for introducing the update here is that currently in 1.2 certification, there is some fake CoAP message (specifically, SRV_DATA.ntf) constructed on the OTBR host side, before this PR, linux will select MLEID as the source before passing this message to NCP, however it is expected that RLOC would be the source of SRV_DATA.ntf. 

we have configured "Thread:Config:FilterRLOCAddresses"  as your tip and tested the update, now it works as expected.

abtink:
Thanks @librasungirl . This change LGTM. Was just curious what use-case needs the RLOC on host side (and give the hint that we need to disable the filter if we want to use it).

simonlingoogle:
> Thanks @librasungirl . This change LGTM. Was just curious what use-case needs the RLOC on host side (and give the hint that we need to disable the filter if we want to use it).

As @librasungirl said, the 1.2 certification requires the host playing a `SBBR` role to send a fake `SVR_DATA.ntf` to the leader. The current way of achieving it is to send a UDP message from host with crafted payload that mimics a real `SVR_DATA.ntf` message. Since `SVR_DATA.ntf` must use `RLOC` as source address, we have to disable the `RLOC` filter on `wpantund` and use this PR to update `RLOC` when changed. 

Generally, I think `RLOC` should be used as source address when the host is sending messages to devices by `RLOC` destination addresses so as to have better source address selection.

PR #4660, opened 2020-03-10
[script] simplify pretty

bukepo:
* No need to bootstrap and configure for make pretty
* Use parallel to speed up make pretty


abtink:
This is really nice and helpful. Thanks @bukepo . :)

PR #4662, opened 2020-03-10
[tests] initialize _initialized

wgtdkp:
This PR initializes `_initialized` of `Node` before doing initialization. Otherwise, we will have an `AttributeError` in desctructor when initialization failed:
```text
Exception ignored in: <function Node.__del__ at 0x7fd3a90b6560>
  File "/usr/local/google/home/wgtdkp/openthread/tests/scripts/thread-cert/node.py", line 187, in __del__
    self.destroy()
  File "/usr/local/google/home/wgtdkp/openthread/tests/scripts/thread-cert/node.py", line 190, in destroy
    if not self._initialized:
AttributeError: 'Node' object has no attribute '_initialized'
```

### DIFF
--- a/tests/scripts/thread-cert/node.py
+++ b/tests/scripts/thread-cert/node.py
@@ -48,6 +48,7 @@ class Node:
         self.verbose = int(float(os.getenv('VERBOSE', 0)))
         self.node_type = os.getenv('NODE_TYPE', 'sim')
         self.env_version = os.getenv('THREAD_VERSION', '1.1')
+        self._initialized = False
 
         if version is not None:
             self.version = version


tests/scripts/thread-cert/node.py
```diff
@@ -48,6 +48,7 @@ def __init__(self, nodeid, is_mtd=False, simulator=None, version=None):
         self.verbose = int(float(os.getenv('VERBOSE', 0)))
         self.node_type = os.getenv('NODE_TYPE', 'sim')
         self.env_version = os.getenv('THREAD_VERSION', '1.1')
+        self._initialized = False
```
simonlingoogle: We can add `_initialized = False` as a class member of `Node`.

wgtdkp: Is this based on the truth that `Simulator` is used as a singleton?

simonlingoogle: No. `self. _initialized` will simply gets `Simulator._initialized` if `_initialized` is not defined on the instance.

simonlingoogle: This is minor. I am fine with both approach.

PR #4664, opened 2020-03-10
[posix] refactoring of netif code, mostly to accommodate BSDs

rob-the-dude:
This code (likely a first attempt, as I'm sure there's some stuff in here that
folks will want to change) does the following:

- provides an implementation to manage networking notices from a PF_ROUTE socket,
	instead of (the Linux-specific) PF_NETLINK

- breaks out some of the platform-specific mechanisms for configuring a tunnel
	device (Linux, mac OS, and NetBSD are all different in some ways)

- provides handling of I/O to the tun drivers on BSDs (where there's a 4-byte
	header prepended to identify the IP version)

- adds some debugging info and more deliberate error handling (I suspect
	there are some more OpenThread-friendly ways to do some of the
	debugging, but I didn't try to figure them out)

I've used most of this code for several months on mac OS and NetBSD, and I
wanted to get a version of it shared for review.  I'd love to see it
merged in so that I don't have to maintain it completely by myself (and perhaps
others will want to use it on those platforms), but I suspect it'll need
some changes first.

### DIFF
--- a/src/posix/platform/netif.cpp
+++ b/src/posix/platform/netif.cpp
@@ -558,6 +558,11 @@ static void logAddrEvent(bool isAdd, bool isUnicast, struct sockaddr_in6 &addr6,
 {
     char addressString[INET6_ADDRSTRLEN + 1];
 
+    // these parameters may not be used if logging is disabled at compile time
+    OT_UNUSED_VARIABLE(isUnicast);
+    OT_UNUSED_VARIABLE(addr6);
+    OT_UNUSED_VARIABLE(addressString);
+
     if ((error == OT_ERROR_NONE) || ((isAdd) && (error == OT_ERROR_ALREADY)) ||
         ((!isAdd) && (error == OT_ERROR_NOT_FOUND)))
     {
@@ -582,9 +587,6 @@ static void processNetifAddrEvent(otInstance *aInstance, struct nlmsghdr *aNetli
     size_t              rtaLength;
     otError             error = OT_ERROR_NONE;
     struct sockaddr_in6 addr6;
-    char                addressString[INET6_ADDRSTRLEN + 1];
-
-    (void)addressString; // to silence warnings if otLog...Plat is defined as a no-op
 
     VerifyOrExit(ifaddr->ifa_index == static_cast<unsigned int>(sTunIndex) && ifaddr->ifa_family == AF_INET6, OT_NOOP);
 
@@ -829,7 +831,7 @@ static void processNetifAddrEvent(otInstance *aInstance, struct rt_msghdr *rtm)
                         struct in6_aliasreq ifr6;
                         char                addressString[INET6_ADDRSTRLEN + 1];
 
-                        (void)addressString; // if otLog*Plat is disabled, we'll get a warning
+                        OT_UNUSED_VARIABLE(addressString); // if otLog*Plat is disabled, we'll get a warning
 
                         memset(&ifr6, 0, sizeof(ifr6));
                         strlcpy(ifr6.ifra_name, sTunName, sizeof(ifr6.ifra_name));
--- a/src/posix/platform/netif.cpp
+++ b/src/posix/platform/netif.cpp
@@ -63,22 +63,24 @@
 
 #if USE_TUN && USE_APPLE_UTUN
 #error "You may not use both tun and utun."
-#endif
+#endif // USE_TUN && USE_APPLE_UTUN
 
-#endif
+#endif // defined(__APPLE__)
 
 #include <arpa/inet.h>
 #include <assert.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <ifaddrs.h>
 #ifdef __linux__
 #include <linux/if_tun.h>
 #include <linux/netlink.h>
 #include <linux/rtnetlink.h>
-#endif
+#endif // __linux__
 #include <net/if.h>
 #include <net/if_arp.h>
 #include <stdio.h>
+#include <string.h>
 #include <sys/ioctl.h>
 #include <sys/select.h>
 #include <sys/socket.h>
@@ -90,7 +92,7 @@
 #include <netinet/in.h>
 #if defined(__APPLE__) || defined(__FreeBSD__)
 #include <net/if_var.h>
-#endif
+#endif // defined(__APPLE__) || defined(__FreeBSD__)
 #include <net/route.h>
 #include <netinet6/in6_var.h>
 //
@@ -103,35 +105,30 @@
 #include <netinet6/nd6.h> // ND6_INFINITE_LIFETIME
 #else
 #define ND6_INFINITE_LIFETIME 0xffffffff
-#endif
+#endif // !(defined(__APPLE__) || defined(__FreeBSD__))
+
 #include <net/if_dl.h> // struct sockaddr_dl
 
 #ifdef __APPLE__
 #if USE_APPLE_UTUN
 #include <net/if_utun.h>
-#endif
+#endif // USE_APPLE_UTUN
 
 #if USE_TUN
 #include <sys/ioccom.h>
 // FIX ME: include the tun_ioctl.h file (challenging, as it's location depends on where the developer puts it)
 #define TUNSIFHEAD _IOW('t', 96, int)
 #define TUNGIFHEAD _IOR('t', 97, int)
-#endif
+#endif // USE_TUN
 
 #include <sys/kern_control.h>
-#endif
+#endif // defined(__APPLE__)
 
-#ifdef __NetBSD__
+#if defined(__NetBSD__) || defined(__FreeBSD__)
 #include <net/if_tun.h>
-#endif
+#endif // defined(__NetBSD__) || defined(__FreeBSD__)
 
-#ifdef __FreeBSD__
-#include <net/if_tun.h>
-#endif
-
-#include <errno.h>
-#include <string.h>
-#endif
+#endif // defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
 
 #include <openthread/icmp6.h>
 #include <openthread/instance.h>
@@ -172,7 +169,7 @@ struct in6_ifreq
     __u32           ifr6_prefixlen;
     int             ifr6_ifindex;
 };
-#endif
+#endif // defined(__linux__) && !defined(_IPV6_H) && !defined(_UAPI_IPV6_H)
 
 #if defined(RTM_NEWMADDR) || defined(__NetBSD__)
 // on some BSDs (mac OS, FreeBSD), we get RTM_NEWMADDR/RTM_DELMADDR messages, so we don't need to monitor using MLD
@@ -184,7 +181,7 @@ struct in6_ifreq
 // MLDv2 messages to know when mulicast memberships change
 // 		https://stackoverflow.com/questions/37346289/using-netlink-is-it-possible-to-listen-whenever-multicast-group-membership-is-ch
 #define OPENTHREAD_POSIX_USE_MLD_MONITOR 1
-#endif
+#endif // defined(RTM_NEWMADDR) || defined(__NetBSD__)
 
 // some platforms (like NetBSD) do not have RTM_NEWMADDR/RTM_DELMADDR messages, and they ALSO lack
 // working MLDv2 support.  for those platforms, we must tell the OpenThread interface to
@@ -244,56 +241,10 @@ enum
 
 static const size_t kMaxIp6Size = 1536;
 
-static bool gLogPackets = false;
-
-static void dump_hex(const void *buffer, size_t len)
-{
-    char           ascii_buffer[sizeof("xxxx xxxx") + 1];
-    const uint8_t *buf = (const uint8_t *)buffer;
-    size_t         i;
-    ascii_buffer[0] = 0;
-    for (i = 0; i < len; i++)
-    {
-        uint16_t val = (*buf) & 0x00FF;
-        buf++;
-
-        ascii_buffer[i % 8]     = isprint(val) ? (char)val : '.';
-        ascii_buffer[i % 8 + 1] = 0;
-
-        if ((i % 8) == 0)
-        {
-            if (i != 0)
-            {
-                fprintf(stderr, "\r\n");
-            }
-            fprintf(stderr, "\t");
-        }
-        if ((i % 4) == 0)
-            fprintf(stderr, "\t");
-
-        fprintf(stderr, "%02hX ", val);
-
-        if (((i + 1) % 8) == 0)
-        {
-            fprintf(stderr, "\t%s", ascii_buffer);
-            ascii_buffer[0] = 0;
-        }
-    }
-    while ((i % 8) != 0)
-    {
-        if ((i % 4) == 0)
-        {
-            fprintf(stderr, "\t");
-        }
-
-        fprintf(stderr, "   ");
-        i++;
-    }
-    fprintf(stderr, "\t%s\r\n", ascii_buffer);
-}
+#define OPENTHREAD_POSIX_LOG_TUN_PACKETS 0
 
 #if !defined(__linux__)
-static bool unicastAddressIsSubscribed(otInstance *aInstance, const otNetifAddress *netAddr)
+static bool UnicastAddressIsSubscribed(otInstance *aInstance, const otNetifAddress *netAddr)
 {
     const otNetifAddress *address = otIp6GetUnicastAddresses(aInstance);
 
@@ -313,52 +264,27 @@ static bool unicastAddressIsSubscribed(otInstance *aInstance, const otNetifAddre
 #endif
 
 #if defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
-static inline void apply_mask(struct in6_addr *address, uint8_t mask)
+static const uint8_t allOnes[] = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+                                  0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
+
+static void InitNetaskWithPrefixLength(struct in6_addr *address, uint8_t prefixLen)
 {
-    if (mask > 128)
+#define MAX_PREFIX_LENGTH (OT_IP6_ADDRESS_SIZE * CHAR_BIT)
+    if (prefixLen > MAX_PREFIX_LENGTH)
     {
-        mask = 128;
+        prefixLen = MAX_PREFIX_LENGTH;
     }
 
-    memset((void *)(address->s6_addr + ((mask + 7) / 8)), 0, 16 - ((mask + 7) / 8));
+    ot::Ip6::Address addr;
 
-    if (mask % 8)
-    {
-        address->s6_addr[mask / 8] &= ~(0xFF >> (mask % 8));
-    }
+    addr.Clear();
+    addr.SetPrefix(allOnes, prefixLen);
+    memcpy(address, addr.mFields.m8, sizeof(addr.mFields.m8));
 }
 
-static unsigned int calculate_prefix_len(const struct sockaddr_in6 *sa6)
+static uint8_t NetmaskToPrefixLength(const struct sockaddr_in6 *netmask)
 {
-    unsigned int result = 0;
-
-    unsigned int i;
-
-    for (i = 0; i < sizeof(sa6->sin6_addr.s6_addr); i++)
-    {
-        if (sa6->sin6_addr.s6_addr[i] == 0xFF)
-        {
-            result += 8;
-            continue;
-        }
-        else
-        {
-            int     j;
-            uint8_t byte = sa6->sin6_addr.s6_addr[i];
-
-            for (j = 7; j >= 0; j--)
-            {
-                if (byte & (1 << j))
-                    result++;
-                else
-                    break;
-            }
-
-            break;
-        }
-    }
-
-    return result;
+    return ot::Ip6::Address::PrefixMatch(netmask->sin6_addr.s6_addr, allOnes, 128);
 }
 #endif
 
@@ -399,8 +325,7 @@ static void UpdateUnicast(otInstance *aInstance, const otIp6Address &aAddress, u
         memcpy(&ifr6.ifra_addr.sin6_addr, &aAddress, sizeof(struct in6_addr));
         ifr6.ifra_prefixmask.sin6_family = AF_INET6;
         ifr6.ifra_prefixmask.sin6_len    = sizeof(ifr6.ifra_prefixmask);
-        memset(&ifr6.ifra_prefixmask.sin6_addr, 0xFF, 16);
-        apply_mask(&ifr6.ifra_prefixmask.sin6_addr, aPrefixLength);
+        InitNetaskWithPrefixLength(&ifr6.ifra_prefixmask.sin6_addr, aPrefixLength);
         ifr6.ifra_lifetime.ia6t_vltime    = ND6_INFINITE_LIFETIME;
         ifr6.ifra_lifetime.ia6t_pltime    = ND6_INFINITE_LIFETIME;
 
@@ -444,18 +369,21 @@ static void UpdateMulticast(otInstance *aInstance, const otIp6Address &aAddress,
 
     int err;
     err = setsockopt(sIpFd, IPPROTO_IPV6, (aIsAdded ? IPV6_JOIN_GROUP : IPV6_LEAVE_GROUP), &mreq, sizeof(mreq));
+#if defined(__APPLE__) || defined(__FreeBSD__)
     if ((err != 0) && (errno == EINVAL) && (IN6_IS_ADDR_MC_LINKLOCAL(&mreq.ipv6mr_multiaddr)))
     {
         // FIX ME
         // on mac OS (and FreeBSD), the first time we run (but not subsequently), we get a failure on this particular
         // join. do we need to bring up the interface at least once prior to joining?
-
+        // we need to figure out why so we can get rid of this workaround
         char addressString[INET6_ADDRSTRLEN + 1];
+
         inet_ntop(AF_INET6, mreq.ipv6mr_multiaddr.s6_addr, addressString, sizeof(addressString));
         otLogWarnPlat("ignoring %s failure (EINVAL) for MC LINKLOCAL address (%s)",
                       aIsAdded ? "IPV6_JOIN_GROUP" : "IPV6_LEAVE_GROUP", addressString);
         err = 0;
     }
+#endif
 
     if (err != 0)
     {
@@ -473,8 +401,8 @@ static void UpdateLink(otInstance *aInstance)
 {
     otError      error = OT_ERROR_NONE;
     struct ifreq ifr;
-    bool         if_state = false;
-    bool         ot_state = false;
+    bool         ifState = false;
+    bool         otState = false;
 
     assert(sInstance == aInstance);
 
@@ -483,14 +411,14 @@ static void UpdateLink(otInstance *aInstance)
     strncpy(ifr.ifr_name, sTunName, sizeof(ifr.ifr_name));
     VerifyOrExit(ioctl(sIpFd, SIOCGIFFLAGS, &ifr) == 0, perror("ioctl"); error = OT_ERROR_FAILED);
 
-    if_state = ((ifr.ifr_flags & IFF_UP) == IFF_UP) ? true : false;
-    ot_state = otIp6IsEnabled(aInstance);
+    ifState = ((ifr.ifr_flags & IFF_UP) == IFF_UP) ? true : false;
+    otState = otIp6IsEnabled(aInstance);
 
-    otLogNotePlat("changing interface state to %s%s.", ot_state ? "UP" : "DOWN",
-                  (if_state == ot_state) ? " (already done, ignoring)" : "");
-    if (if_state != ot_state)
+    otLogNotePlat("changing interface state to %s%s.", otState ? "UP" : "DOWN",
+                  (ifState == otState) ? " (already done, ignoring)" : "");
+    if (ifState != otState)
     {
-        ifr.ifr_flags = ot_state ? (ifr.ifr_flags | IFF_UP) : (ifr.ifr_flags & ~IFF_UP);
+        ifr.ifr_flags = otState ? (ifr.ifr_flags | IFF_UP) : (ifr.ifr_flags & ~IFF_UP);
         VerifyOrExit(ioctl(sIpFd, SIOCSIFFLAGS, &ifr) == 0, perror("ioctl"); error = OT_ERROR_FAILED);
     }
 
@@ -528,11 +456,12 @@ static void processStateChange(otChangedFlags aFlags, void *aContext)
 static void processReceive(otMessage *aMessage, void *aContext)
 {
     char     packet[kMaxIp6Size + 4];
-    otError  error      = OT_ERROR_NONE;
-    uint16_t length     = otMessageGetLength(aMessage);
-    size_t   offset     = 0;
-    uint16_t max_length = sizeof(packet) - 4;
+    otError  error     = OT_ERROR_NONE;
+    uint16_t length    = otMessageGetLength(aMessage);
+    size_t   offset    = 0;
+    uint16_t maxLength = sizeof(packet) - 4;
 #if defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
+    // BSD tunnel drivers use (for legacy reasons) a 4-byte header to determine the address family of the packet
     offset += 4;
 #endif
 
@@ -540,7 +469,12 @@ static void processReceive(otMessage *aMessage, void *aContext)
 
     VerifyOrExit(sTunFd > 0, OT_NOOP);
 
-    VerifyOrExit(otMessageRead(aMessage, 0, &packet[offset], max_length) == length, error = OT_ERROR_NO_BUFS);
+    VerifyOrExit(otMessageRead(aMessage, 0, &packet[offset], maxLength) == length, error = OT_ERROR_NO_BUFS);
+
+#if OPENTHREAD_POSIX_LOG_TUN_PACKETS
+    otLogInfoPlat("Packet from NCP (%hu bytes)", static_cast<uint16_t>(length));
+    otDumpInfo(OT_LOG_REGION_PLATFORM, "", &packet[offset], length);
+#endif
 
 #if defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
     packet[0] = 0;
@@ -549,13 +483,6 @@ static void processReceive(otMessage *aMessage, void *aContext)
     packet[3] = (PF_INET6 << 0) & 0xFF;
     length += 4;
 #endif
-
-    if (gLogPackets)
-    {
-        fprintf(stderr, "Packet from NCP (%hu bytes)\r\n", (uint16_t)(length - 4));
-        dump_hex(&packet[4], length - 4);
-        fflush(stdout);
-    }
 
     VerifyOrExit(write(sTunFd, packet, length) == length, perror("write"); error = OT_ERROR_FAILED);
 
@@ -589,6 +516,7 @@ static void processTransmit(otInstance *aInstance)
     VerifyOrExit(message != NULL, error = OT_ERROR_NO_BUFS);
 
 #if defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
+    // BSD tunnel drivers have (for legacy reasons), may have a 4-byte header on them
     if ((rval >= 4) && (packet[0] == 0) && (packet[1] == 0))
     {
         rval -= 4;
@@ -596,12 +524,10 @@ static void processTransmit(otInstance *aInstance)
     }
 #endif
 
-    if (gLogPackets)
-    {
-        fprintf(stderr, "Packet to NCP (%ld bytes)\r\n", (long)rval);
-        dump_hex(&packet[offset], (size_t)rval);
-        fflush(stdout);
-    }
+#if OPENTHREAD_POSIX_LOG_TUN_PACKETS
+    otLogInfoPlat("Packet to NCP (%hu bytes)", static_cast<uint16_t>(rval));
+    otDumpInfo(OT_LOG_REGION_PLATFORM, "", &packet[offset], static_cast<size_t>(rval));
+#endif
 
     SuccessOrExit(error = otMessageAppend(message, &packet[offset], static_cast<uint16_t>(rval)));
 
@@ -908,9 +834,9 @@ static void processNetifAddrEvent(otInstance *aInstance, struct rt_msghdr *rtm) 
                 bool           subscribed;
 
                 netAddr.mAddress      = addr;
-                netAddr.mPrefixLength = (uint8_t)calculate_prefix_len(&netmask);
+                netAddr.mPrefixLength = NetmaskToPrefixLength(&netmask);
 
-                subscribed = unicastAddressIsSubscribed(aInstance, &netAddr);
+                subscribed = UnicastAddressIsSubscribed(aInstance, &netAddr);
 
                 if (subscribed)
                 {
@@ -933,8 +859,7 @@ static void processNetifAddrEvent(otInstance *aInstance, struct rt_msghdr *rtm) 
                         memcpy(&ifr6.ifra_addr.sin6_addr, &addr6.sin6_addr, sizeof(struct in6_addr));
                         ifr6.ifra_prefixmask.sin6_family = AF_INET6;
                         ifr6.ifra_prefixmask.sin6_len    = sizeof(ifr6.ifra_prefixmask);
-                        memset(&ifr6.ifra_prefixmask.sin6_addr, 0xFF, 16);
-                        apply_mask(&ifr6.ifra_prefixmask.sin6_addr, netAddr.mPrefixLength);
+                        InitNetaskWithPrefixLength(&ifr6.ifra_prefixmask.sin6_addr, netAddr.mPrefixLength);
                         ifr6.ifra_lifetime.ia6t_vltime = ND6_INFINITE_LIFETIME;
                         ifr6.ifra_lifetime.ia6t_pltime = ND6_INFINITE_LIFETIME;
 
@@ -1374,7 +1299,6 @@ static otError destroyTunnel(void)
     error = OT_ERROR_NONE;
 
 exit:
-
     return error;
 }
 #endif
@@ -1468,10 +1392,7 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
     sIpFd = SocketWithCloseExec(AF_INET6, SOCK_DGRAM, IPPROTO_IP, kSocketNonBlock);
     VerifyOrDie(sIpFd >= 0, OT_EXIT_ERROR_ERRNO);
 
-    otIcmp6SetEchoMode(aInstance, OT_ICMP6_ECHO_HANDLER_DISABLED);
-
     platformConfigureNetLink();
-
     platformConfigureTunDevice(aInstance, aInterfaceName, sTunName, sizeof(sTunName));
 
     sTunIndex = if_nametoindex(sTunName);
@@ -1484,10 +1405,10 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
     mldListenerInit();
 #endif
 
+    otIcmp6SetEchoMode(aInstance, OT_ICMP6_ECHO_HANDLER_DISABLED);
     otIp6SetReceiveCallback(aInstance, processReceive, aInstance);
     otIp6SetAddressCallback(aInstance, processAddressChange, aInstance);
     otSetStateChangedCallback(aInstance, processStateChange, aInstance);
-
 #if OPENTHREAD_POSIX_MULTICAST_PROMISCUOUS_REQUIRED
     otIp6SetMulticastPromiscuousEnabled(aInstance, true);
 #endif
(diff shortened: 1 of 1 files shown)


bukepo:
@rob-the-dude I think multicast is a requirement for Thread. Do we need to support Apple utun if multicast is not supported?

I find it pretty straightforward to install tuntap through homebrew:

```bash
brew cask install tuntap
```

Then allow the extension according to the formula of [tuntap](https://formulae.brew.sh/cask/tuntap).

bukepo:
I just played on macOS 10.15.3.

```bash
➜  openthread git:(pr/4664) ✗ sudo ./output/posix/x86_64-apple-darwin/bin/ot-cli output/x86_64-apple-darwin/bin/ot-rcp 1
> panid 0xface
Done
> ifconfig up
IPV6_JOIN_GROUP failure (22)
Done
> 	thread: ADD [M] ff03:0000:0000:0000:0000:0000:0000:0001 failure (7)
	thread: ADD [M] ff03:0000:0000:0000:0000:0000:0000:00fc failure (7)
	thread:         fe80:0000:0000:0000:aede:48ff:fe00:1122 (remove stack-added link-local)
	thread: ADD [M] ff02:0000:0000:0000:0000:0001:ff00:1122
	thread: ADD [M] ff02:0000:0000:0000:0000:0000:0000:0001 failure (7)
	thread: ADD [M] ff02:0000:0000:0000:0000:0002:ff97:242f
	thread: ADD [M] ff01:0013:0000:0000:0000:0000:0000:0001 failure (3)
	thread: ADD [U] fe80:0000:0000:0000:d4cb:52b7:c36e:74db (already subscribed, ignored)
	thread: ADD [M] ff02:0000:0000:0000:0000:0001:ff6e:74db failure (3)
	thread: DEL [U] fe80:0000:0000:0000:aede:48ff:fe00:1122 (not found, ignored)
	thread: DEL [M] ff02:0000:0000:0000:0000:0001:ff00:1122

>
>
>
>
> thread start
Done
> 	thread: ADD [U] fdde:ad00:beef:0000:85c7:f4f0:0459:6a43 (already subscribed, ignored)
	thread: ADD [M] ff02:0000:0000:0000:0000:0001:ff59:6a43
	thread: ADD [M] ff32:0000:fdde:ad00:beef:0000:0000:0001 failure (3)
	thread: ADD [M] ff33:0040:fdde:ad00:beef:0000:0000:0001 (already subscribed, ignored)

>
>
>
>
> 	thread: ADD [U] fdde:ad00:beef:0000:0000:00ff:fe00:f400 (already subscribed, ignored)
	thread: ADD [M] ff02:0000:0000:0000:0000:0001:ff00:f400 failure (3)
	thread: ADD [U] fdde:ad00:beef:0000:0000:00ff:fe00:fc00 (already subscribed, ignored)
	thread: ADD [M] ff02:0000:0000:0000:0000:0001:ff00:fc00 failure (3)
	thread: ADD [M] ff02:0000:0000:0000:0000:0000:0000:0002 failure (7)
	thread: ADD [M] ff03:0000:0000:0000:0000:0000:0000:0002 failure (7)

>
>
>
> ipaddr
fdde:ad00:beef:0:0:ff:fe00:f400
fdde:ad00:beef:0:85c7:f4f0:459:6a43
fe80:0:0:0:d4cb:52b7:c36e:74db
Done
> ipmaddr
ff02:0:0:0:0:1:ff59:6a43
ff33:40:fdde:ad00:beef:0:0:1
ff32:40:fdde:ad00:beef:0:0:1
ff02:0:0:0:0:2:ff97:242f
ff02:0:0:0:0:0:0:2
ff03:0:0:0:0:0:0:2
ff02:0:0:0:0:0:0:1
ff03:0:0:0:0:0:0:1
ff03:0:0:0:0:0:0:fc
Done
```

There are some failures, I'm not sure they matters. Besides, I hope to verify if multicast addresses are subscribed, and I tried `netstat -r -f inet6`, but I didn't find addresses listed in OpenThread's `ipmaddr`. I don't know whether I'm using the correct command to check multicast subscriptions.

rob-the-dude:
> @rob-the-dude I think multicast is a requirement for Thread. Do we need to support Apple utun if multicast is not supported?

Fair point.  I certainly don't care to support Apple utun until it supports multicast.  Would you prefer me to remove it?

> I find it pretty straightforward to install tuntap through homebrew:
> 
> ```shell
> brew cask install tuntap
> ```
> 
> Then allow the extension according to the formula of [tuntap](https://formulae.brew.sh/cask/tuntap).

It's nice to hear that works.  I don't use brew on my system(s).  I use very few command line tools or system-wide additions that I cannot build from source.

Third-party kernel drivers (like tun-tap) are on the decline on mac OS.  At WWDC last year, they introduced many new ways of doing USB and networking devices outside the kernel -- but for the moment, tun-tap still works.  I don't know if that will remain true after the next OS release or not, but if it isn't, then I'm hopeful that either (a) utun will gain support for multicast, or (b) there'll be a new way to do what tun-tap does.

rob-the-dude:
> There are some failures, I'm not sure they matters. Besides, I hope to verify if multicast addresses are subscribed, and I tried `netstat -r -f inet6`, but I didn't find addresses listed in OpenThread's `ipmaddr`. I don't know whether I'm using the correct command to check multicast subscriptions.

If you've got a sequence of commands that form a test, I'd be happy to perform them and examine the output on both Linux and mac OS.  During my development, I found that certain address manipulation was also problematic on Linux, but the errors were ignored.  As I point out, I added additional logging -- partially in the hopes that the Linux code would ALSO stop generating errors (not by adding the code, but because the errors now being logged would be corrected by others using the code).  But it's been awhile since I tried debugging the code on Linux.

Again, if you've got a quick sanity test, I'd be happy to run it on both platforms.  Your comment mentions this:

sudo ./output/posix/x86_64-apple-darwin/bin/ot-cli output/x86_64-apple-darwin/bin/ot-rcp 1
> panid 0xface
> ifconfig up
> thread start
> ipaddr
> ipmaddr

Is that enough?  If so, I can run those commands on an existing Linux build, a new Linux build, and a mac OS build and report back on any differences (I might add the output of "netstat -n -r -f inet6", but that seems to show route information instead of subscribed multicast addresses -- though in practicality, I think I can see how they might end up being the same).

rob-the-dude:
Did some deeper testing and debugging into the error messages.  Would like feedback on a couple cases from @bukepo and @jwhui.

This error message:

     thread: ADD [M] ff02:0000:0000:0000:0000:0001:ff00:f400 failure (3)

Is happening because otIp6SubscribeMulticastAddress is returning OT_ERROR_NO_BUFS.   That's because the default number of external multicast addresses is small (2), and a mac OS system generally has a handful of system processes that want to bind to multicast addresses when they show up in the system and get configured for IFF_UP (i.e., mDNSResponder is one of them).  On Linux, this more or less gets "eaten" (ignored) by the "SuccessOrExit" call inside "processNetIfAddrEvent".  I'm certainly unhappy with that outcome -- silently failing to bind to a multicast address would make system processes fail with practically no hint as to what's wrong.  I'd be happy to special case the handling of  OT_ERROR_NO_BUFS with a particular log message ("no space for address") and a comment in the code that the developer might need to increase OPENTHREAD_CONFIG_IP6_MAX_EXT_MCAST_ADDRS.  But in reality, such a notice might be better placed in "Netif::SubscribeExternalMulticast".

Additionally, I'd ask if, on mac OS specifically, we should increase the default value of OPENTHREAD_CONFIG_IP6_MAX_EXT_MCAST_ADDRS?

rob-the-dude:
Similar to my last comment, I'd like input from @bukepo and @jwhui -- these error messages:

	thread: ADD [M] ff02:0000:0000:0000:0000:0000:0000:0002 failure (7)
	thread: ADD [M] ff03:0000:0000:0000:0000:0000:0000:0002 failure (7)

Are happening because otIp6SubscribeMulticastAddress (technically, Netif::SubscribeExternalMulticast) recognizes them as one of the "fixed" addresses, and returns OT_ERROR_INVALID_ARGS.

I'm not a fan of this error code from this particular function, mainly because those "fixed" addresses aren't exposed externally -- they're defined in the Netif implementation file and not available elsewhere.  So I can't say, "oh, I got back OT_ERROR_INVALID_ARGS -- did I pass in one of the special addresses?" (or for that matter, detect it's a magic address and not call the function anyway), as opposed to a truly invalid argument (usually, I think of invalid arguments as me passing in something that's a bogus file descriptor, a bogus pointer, or, in this case, something that wasn't actually a valid multicast address).  Instead, I have to know that OT_ERROR_INVALID_ARGS means something special to this particular method, and know that I might sometimes get this error, and it doesn't mean that the address is bad -- it means that it's a special address.  (Indeed, from a quick perusal of the code, it looks like I could totally pass a bogus multicast address (say, a unicast address) to otIp6SubscribeMulticastAddress, and -- as long as it wasn't one of the special ones -- it wouldn't complain at all.  (That seems like a bug to me.)

In any event, I'd probably prefer that otIp6SubscribeMulticastAddress return OT_ERROR_ALREADY for the "special" addresses -- at least I'd know that, "hey, this address is already in our list of multicast address we'll receive packets for" and know that I'm OK.  If that's not going to happen, then I'd like the "special" addresses exposed (or give me a helper function to know it's one of them).  Otherwise, I really have no way to protecting myself from OT_ERROR_INVALID_ARGS actually meaning what I think it might mean, should future changes to the underlying code ever happen.

Finally, (I think) the reason this happens on mac OS and not Linux is that it seems like Linux may filter out PF_NETLINK notifications for changes that our process makes to the device (at least it seems that way for these addresses -- I don't see a PF_NETLINK notification for them, even though they do get added).  The mac OS code doesn't filter by PID (I could try that, but I'm not certain it's correct behavior).  In any event, the error message in this case is telling us something is happening that's more of a programmer error than an actual system misbehavior.  The particular multicast address that you're attempting to add to the interface is indeed already in the list (albeit a special "wouldn't you like to know" list).

bukepo:
> Similar to my last comment, I'd like input from @bukepo and @jwhui -- these error messages:
> 
> ```
> thread: ADD [M] ff02:0000:0000:0000:0000:0000:0000:0002 failure (7)
> thread: ADD [M] ff03:0000:0000:0000:0000:0000:0000:0002 failure (7)
> ```
> 
> Are happening because otIp6SubscribeMulticastAddress (technically, Netif::SubscribeExternalMulticast) recognizes them as one of the "fixed" addresses, and returns OT_ERROR_INVALID_ARGS.
> 
> I'm not a fan of this error code from this particular function, mainly because those "fixed" addresses aren't exposed externally -- they're defined in the Netif implementation file and not available elsewhere. So I can't say, "oh, I got back OT_ERROR_INVALID_ARGS -- did I pass in one of the special addresses?" (or for that matter, detect it's a magic address and not call the function anyway), as opposed to a truly invalid argument (usually, I think of invalid arguments as me passing in something that's a bogus file descriptor, a bogus pointer, or, in this case, something that wasn't actually a valid multicast address). Instead, I have to know that OT_ERROR_INVALID_ARGS means something special to this particular method, and know that I might sometimes get this error, and it doesn't mean that the address is bad -- it means that it's a special address. (Indeed, from a quick perusal of the code, it looks like I could totally pass a bogus multicast address (say, a unicast address) to otIp6SubscribeMulticastAddress, and -- as long as it wasn't one of the special ones -- it wouldn't complain at all. (That seems like a bug to me.)
> 
> In any event, I'd probably prefer that otIp6SubscribeMulticastAddress return OT_ERROR_ALREADY for the "special" addresses -- at least I'd know that, "hey, this address is already in our list of multicast address we'll receive packets for" and know that I'm OK. If that's not going to happen, then I'd like the "special" addresses exposed (or give me a helper function to know it's one of them). Otherwise, I really have no way to protecting myself from OT_ERROR_INVALID_ARGS actually meaning what I think it might mean, should future changes to the underlying code ever happen.
> 
> Finally, (I think) the reason this happens on mac OS and not Linux is that it seems like Linux may filter out PF_NETLINK notifications for changes that our process makes to the device (at least it seems that way for these addresses -- I don't see a PF_NETLINK notification for them, even though they do get added). The mac OS code doesn't filter by PID (I could try that, but I'm not certain it's correct behavior). In any event, the error message in this case is telling us something is happening that's more of a programmer error than an actual system misbehavior. The particular multicast address that you're attempting to add to the interface is indeed already in the list (albeit a special "wouldn't you like to know" list).

Hi @rob-the-dude, I agree that we should return `OT_ERROR_ALREADY` when subscribing *fixed* multicast addresses. Thank you for the detailed analysis! Would you like to submit a PR for this issue, or just piggyback on this PR?

abtink:
@rob-the-dude good to see you here  :) .... Adding some thoughts below. 
 
- Returning `OT_ERROR_ALREADY` for the special/internal multicast addresses may indicate that the address is already available on netif which may be true for "link-local all nodes" but is not necessarily the case for some of the other addresses like "link-local all routers", etc. (which are only added by OT when device is FTD and a router/REED). 
- On unicast addresses, `otIp6AddUnicastAddress()` also restricts user from adding/changing some internal addresses, for example, user cannot add any link-local IPv6 address (I think the error is same `INVALID_ARGS` here). 
- On `wpantund` such addresses (for both unicast and multicast) are filtered and not pushed down to NCP.  I agree it requires the knowledge of code and what addresses are considered internal. I would suggest we add documentation of what "internal addresses" are in these OT public APIs.

About running out of unicast/multicast address space on OT core, that can lead to some tricky situations in `wpantund` as well. User can add an unicast IPv6 address on "wpan0" interface and it would seemingly succeed but later on when `wpantund` gets the netlink msg about the newly added address and tries to push the same address to NCP, it may get an error `NO_BUFS`. What to do then? 
- Do we just log it as an error and let it be? (address would be on "wpan0" but not on NCP side).
- Do we remove the address from "wpan0" (how to report to user that adding the address failed?)
- Not let user add address on "wpan0" interface and require user to always go through driver and a `wpantund` command? (not very user-friendly). 
- The current approach in `wpantund` (which to be honest I am not a fan of, but I agree is the safest and most noticeable) is to assert/die in `wpantund` when we cannot sync addresses between host and NCP. This practically causes continuous wpantund restarts and brings down the wpan network and makes it unusable (with the hope that user would then notice).

Thoughts?

rob-the-dude:
> @rob-the-dude good to see you here :)

Nice to be here.

> * Returning `OT_ERROR_ALREADY` for the special/internal multicast addresses may indicate that the address is already available on netif which may be true for "link-local all nodes" but is not necessarily the case for some of the other addresses like "link-local all routers", etc. (which are only added by OT when device is FTD and a router/REED).

Actually, perhaps this is a good argument for returning OT_ERROR_ALREADY ONLY when those addresses are bound, and returning OT_ERROR_INVALID_ARGS when they are not but still prohibited.

In actual correct practice, no one is going to bind to those addresses -- the only reason the mac OS POSIX daemon is seeing the notification is because OpenThread itself did the bind, and the PF_ROUTE socket told it about it -- in which case, returning OT_ERROR_ALREADY would be "correct".  If you're not a router/REED, then the POSIX daemon won't bind to those address, and if someone else tried to do so, then returning OT_ERROR_INVALID_ARGS seems quite valid -- this address argument is "invalid", at least in the current state.

> * On unicast addresses, `otIp6AddUnicastAddress()` also restricts user from adding/changing some internal addresses, for example, user cannot add any link-local IPv6 address (I think the error is same `INVALID_ARGS` here).

I'd argue for the same behavior -- if I pass down the SAME link-local address that is bound, then the error ought to be OT_ERROR_ALREADY, and otherwise OT_ERROR_INVALID_ARGS is OK.  But at least in this case, there's a simple macro to check if something's a link-local address, and it's easy to document that you can't bind to link-local addresses.  The special multicast addresses require a deeper knowledge of how Thread itself works -- they're not really special in any other context.

> * On `wpantund` such addresses (for both unicast and multicast) are filtered and not pushed down to NCP.  I agree it requires the knowledge of code and what addresses are considered internal. I would suggest we add documentation of what "internal addresses" are in these OT public APIs.

Ugh <...shiver...>

> About running out of unicast/multicast address space on OT core, that can lead to some tricky situations in `wpantund` as well. User can add a unicast IPv6 address on "wpan0" interface and it would seemingly succeed but later on when `wpantund` gets the netlink msg about the newly added address and tries to push the same address to NCP, it may get an error `NO_BUFS`. What to do then?
> 
> * Do we just log it as an error and let it be? (address would be on "wpan0" but not on NCP side).
> * Do we remove the address from "wpan0" (how to report to user that adding the address failed?)
> * Not let user add address on "wpan0" interface and require user to always go through driver and a `wpantund` command? (not very user-friendly).

Seems like that would also break a lot of IP software that uses ifreq to do such things.

> * The current approach in `wpantund` (which to be honest I am not a fan of, but I agree is the safest and most noticeable) is to assert/die in `wpantund` when we cannot sync addresses between host and NCP. This practically causes continuous wpantund restarts and brings down the wpan network and makes it unusable (with the hope that user would then notice).
> 
> Thoughts?

I can't say that I have anything better, other than documenting -- perhaps in the log message, but at least in the code -- that a common reason for getting OT_ERROR_NO_BUFS is the build-time configuration parameter for max address count.  NO_BUFS is a generic name, and the first time you get that, you're like, "oh, crap, how do I increase the number of buffers" and you start thinking about a problem you probably don't have.  Having a log message that specifically says, "No room for additional addresses" and a comment that specifically says "if you are hitting OT_ERROR_NO_BUFS in this situation, it may mean that your deployment needs to increase the maximum (unicast/multicast) address constants OPENTHREAD_CONFIG_IP6_MAX_EXT_UCAST_ADDRS/OPENTHREAD_CONFIG_IP6_MAX_EXT_MCAST_ADDRS".

jwhui:
> Actually, perhaps this is a good argument for returning OT_ERROR_ALREADY ONLY when those addresses are bound, and returning OT_ERROR_INVALID_ARGS when they are not but still prohibited.
> 
> In actual correct practice, no one is going to bind to those addresses -- the only reason the mac OS POSIX daemon is seeing the notification is because OpenThread itself did the bind, and the PF_ROUTE socket told it about it -- in which case, returning OT_ERROR_ALREADY would be "correct". If you're not a router/REED, then the POSIX daemon won't bind to those address, and if someone else tried to do so, then returning OT_ERROR_INVALID_ARGS seems quite valid -- this address argument is "invalid", at least in the current state.

Makes sense to me.

> I can't say that I have anything better, other than documenting -- perhaps in the log message, but at least in the code -- that a common reason for getting OT_ERROR_NO_BUFS is the build-time configuration parameter for max address count. NO_BUFS is a generic name, and the first time you get that, you're like, "oh, crap, how do I increase the number of buffers" and you start thinking about a problem you probably don't have. Having a log message that specifically says, "No room for additional addresses" and a comment that specifically says "if you are hitting OT_ERROR_NO_BUFS in this situation, it may mean that your deployment needs to increase the maximum (unicast/multicast) address constants OPENTHREAD_CONFIG_IP6_MAX_EXT_UCAST_ADDRS/OPENTHREAD_CONFIG_IP6_MAX_EXT_MCAST_ADDRS".

Agree with documentation and a clear log message at a minimum. The API documentation does have the information, but maybe we could explicitly mention `OPENTHREAD_CONFIG_IP6_MAX_EXT_MCAST_ADDRS` in the `OT_ERROR_NO_BUFS` return value documentation.

https://github.com/openthread/openthread/blob/bf1cf97b0ffbf77d52bc8c9346e3280f5bbaf845/include/openthread/ip6.h#L184-L192

rob-the-dude:
> > Actually, perhaps this is a good argument for returning OT_ERROR_ALREADY ONLY when those addresses are bound, and returning OT_ERROR_INVALID_ARGS when they are not but still prohibited.
> > In actual correct practice, no one is going to bind to those addresses -- the only reason the mac OS POSIX daemon is seeing the notification is because OpenThread itself did the bind, and the PF_ROUTE socket told it about it -- in which case, returning OT_ERROR_ALREADY would be "correct". If you're not a router/REED, then the POSIX daemon won't bind to those address, and if someone else tried to do so, then returning OT_ERROR_INVALID_ARGS seems quite valid -- this address argument is "invalid", at least in the current state.
> 
> Makes sense to me.

Should I file an issue against otIp6SubscribeMulticastAddress/Netif::SubscribeExternalMulticast to implement exactly that behavior (i.e., returning OT_ERROR_ALREADY when bound and OT_ERROR_INVALID_ARGS otherwise)?

I MIGHT be able to provide a patch, but I'm a little nervous about it -- I am not deeply familiar with when those addresses are actually bound and when they aren't.  Right now, the code checks against them in basically a static list -- I'll need to figure out when that list is "active" and when it isn't in order to implement the change.  Regardless, I think that's a separate PR.

BTW, I'm leaning towards withdrawing this PR until after #4687 is done -- and probably until after the correct behavior from otIp6SubscribeMulticastAddress is also implemented.

gjc13:
See https://github.com/openthread/openthread/pull/4693 for the multicast subscribe return value isse.

rob-the-dude:
> See #4693 for the multicast subscribe return value issue.

Thanks; tried them out locally and got the expected behavior.  I'll rebase against them once they're merged.  Also eager to get #4687.

bukepo:
@rob-the-dude did you get a chance to update this PR?

rob-the-dude:
> @rob-the-dude did you get a chance to update this PR?

I was waiting for #4804 to get merged (even though I'd made a suggestion but apparently didn't submit it -- doh, my bad), and was doing a little bit of additional cleanup based on changes that have gone in since I did the original.  I see now that you've got a change dependent on this.  I'll step it up, but I think it would be better if #4804 could make it in first.

bukepo:
@rob-the-dude #4804 has been merged.

src/posix/platform/netif.cpp
```diff
@@ -79,13 +174,43 @@ struct in6_ifreq
...
 #endif
 
-static otInstance * sInstance     = NULL;
-static int          sTunFd        = -1; ///< Used to exchange IPv6 packets.
-static int          sIpFd         = -1; ///< Used to manage IPv6 stack on Thread interface.
-static int          sNetlinkFd    = -1; ///< Used to receive netlink events.
-static int          sMLDMonitorFd = -1; ///< Used to receive MLD events.
-static unsigned int sTunIndex     = 0;
+#if defined(RTM_NEWMADDR) || defined(__NetBSD__)
+// on some BSDs (mac OS, FreeBSD), we get RTM_NEWMADDR/RTM_DELMADDR messages, so we don't need to monitor using MLD
```
bukepo, line 178: that's great to see we can handle multicast address on macOS.

src/posix/platform/netif.cpp
```diff
@@ -596,18 +1316,180 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
...
+    int         err;
+    const char *last_slash;
+    const char *path;
+
+    (void)aInterfaceName;
+    (void)aInstance;
+
+    path = OPENTHREAD_POSIX_TUN_DEVICE;
+
+    sTunFd = open(path, O_RDWR | O_NONBLOCK);
```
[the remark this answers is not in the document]
rob-the-dude: I cannot find a way to do that on mac OS or NetBSD (the main platforms I'm using).  FreeBSD has a way to rename interfaces, but that mechanism does not exist in mac OS (even though they share a lot of their stack).  NetBSD is also missing that feature.  Apple's 'utun' interface has a way to dynamically obtain an unused interface (and the code here for utun makes use of it), but it does not provide you a way to arbitrarily name the interface.

rob-the-dude:
> @rob-the-dude not sure if you have time to contribute some test scripts on macOS. currently the macos check is very simple. Current the macOS only checks build. But with this change, we can verify the platform netif and platform udp on macOS with the script/check-posix-pty.

I would love to do that.  Can I do it in a second PR, since it's adding additional functionality, and the existing functionality is being tested on Linux?  There's a lot of changes here, and I get nervous about anyone else touching netif before I can get them in :-).

I've already tried using script/check-posix-pty on mac OS, but only indirectly -- manually running the tests that it automates.  Some things it uses (socat, libcoap) aren't on mac OS by default.  Both appear to be available via homebrew, but I don't know if homebrew is available for GitHub mac OS actions?  I don't use homebrew on my own development systems, so I'll have to be doing this work "live" on my GitHub fork.  I need to come up to speed on what the capabilities of GitHub actions for mac OS look like.

src/posix/platform/netif.cpp
```diff
@@ -269,7 +588,22 @@ static void processTransmit(otInstance *aInstance)
...
+    if ((rval >= 4) && (packet[0] == 0) && (packet[1] == 0))
+    {
+        rval -= 4;
+        offset = 4;
+    }
+#endif
+
+    if (gLogPackets)
+    {
+        fprintf(stderr, "Packet to NCP (%ld bytes)\r\n", (long)rval);
```
gjc13, line 601: Use `otLog` rather than stderr. (Applies to other `fprintf` as well).

src/posix/platform/netif.cpp
```diff
@@ -309,10 +649,15 @@ static void processNetifAddrEvent(otInstance *aInstance, struct nlmsghdr *aNetli
         case IFA_LOCAL:
         case IFA_BROADCAST:
         case IFA_ANYCAST:
+        case IFA_MULTICAST:
```
gjc13, line 652: Have you checked whether the linux kernel will issue link events on multicast join? Based on my previous experience on wpantund and platform netif, it won't and that's why we need the mld listener.

src/posix/platform/netif.cpp
```diff
@@ -115,31 +240,185 @@ enum
...
+        }
+
+        fprintf(stderr, "   ");
+        i++;
+    }
+    fprintf(stderr, "\t%s\r\n", ascii_buffer);
+}
+
+#if !defined(__linux__)
+static bool unicastAddressIsSubscribed(otInstance *aInstance, const otNetifAddress *netAddr)
```
gjc13, line 296:
```
static bool UnicastAddressIsSubscribed(otInstance *aInstance, const otNetifAddress *aNetAddr)
```

src/posix/platform/netif.cpp
```diff
@@ -379,6 +793,284 @@ static void processNetifLinkEvent(otInstance *aInstance, struct nlmsghdr *aNetli
...
+                if (subscribed)
+                {
+                    inet_ntop(AF_INET6, addr6.sin6_addr.s6_addr, addressString, sizeof(addressString));
+                    otLogNotePlat("ADD [U] %s (already subscribed, ignored)", addressString);
+                }
+                else
+                {
+                    if (is_link_local)
+                    {
+                        // remove the stack-added link-local address
```
gjc13, line 924: Why shall we remove stack-added link local address from the interface address?

src/posix/platform/netif.cpp
```diff
@@ -379,6 +793,284 @@ static void processNetifLinkEvent(otInstance *aInstance, struct nlmsghdr *aNetli
...
+
+        if (rtm->rtm_type == RTM_NEWADDR
+#ifdef RTM_NEWMADDR
+            || rtm->rtm_type == RTM_NEWMADDR
+#endif
+        )
+        {
+            if (!addr.IsMulticast())
+            {
+                otNetifAddress netAddr;
```
gjc13, line 907: Move this long code block to a separate function?

src/posix/platform/netif.cpp
```diff
@@ -379,6 +793,284 @@ static void processNetifLinkEvent(otInstance *aInstance, struct nlmsghdr *aNetli
...
+                                "ADD [U] %s failed (%s)",
+                                inet_ntop(AF_INET6, addr6.sin6_addr.s6_addr, addressString, sizeof(addressString)),
+                                otThreadErrorToString(error));
+                        }
+                    }
+                }
+            }
+            else
+            {
+                otNetifMulticastAddress netAddr;
```
gjc13, line 986: A new function as well?

src/posix/platform/netif.cpp
```diff
@@ -321,22 +666,89 @@ static void processNetifAddrEvent(otInstance *aInstance, struct nlmsghdr *aNetli
 
                     netAddr.mAddress      = addr;
                     netAddr.mPrefixLength = ifaddr->ifa_prefixlen;
-                    SuccessOrExit(error = otIp6AddUnicastAddress(aInstance, &netAddr));
+
+                    error = otIp6AddUnicastAddress(aInstance, &netAddr);
+
+                    if ((error == OT_ERROR_NONE) || (error == OT_ERROR_ALREADY))
+                    {
+                        otLogNotePlat(
```
gjc13, line 674: These logging code seems duplicate. Can we add a macro or function for them?

src/posix/platform/netif.cpp
```diff
@@ -379,6 +793,284 @@ static void processNetifLinkEvent(otInstance *aInstance, struct nlmsghdr *aNetli
...
+                {
+                    otLogWarnPlat("DEL [M] %s failure (%d)",
+                                  inet_ntop(AF_INET6, addr6.sin6_addr.s6_addr, addressString, sizeof(addressString)),
+                                  error);
+                }
+            }
+        }
+    }
+
+exit:;
```
gjc13, line 1050:
```
exit:
    return;
```
Or we may do the logging here to avoid duplicate code?

src/posix/platform/netif.cpp
```diff
@@ -469,6 +1201,7 @@ static void processMLDEvent(otInstance *aInstance)
     size_t              offset;
     uint8_t             type;
     struct ifaddrs *    ifAddrs = NULL;
+    char                addressString[INET6_ADDRSTRLEN + 1];
```
gjc13, line 1204: Is this variable used?

src/posix/platform/netif.cpp
```diff
@@ -596,18 +1319,179 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
...
+    otError      error;
+    struct ifreq ifr;
+
+    memset(&ifr, 0, sizeof(ifr));
+    strncpy(ifr.ifr_name, sTunName, sizeof(ifr.ifr_name));
+    VerifyOrExit(ioctl(sIpFd, SIOCIFDESTROY, &ifr) == 0, perror("ioctl"); error = OT_ERROR_FAILED);
+    error = OT_ERROR_NONE;
+
+exit:
+
```
gjc13, line 1377: No new line here.

src/posix/platform/netif.cpp
```diff
@@ -596,18 +1319,179 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
...
+#endif // defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
+}
+
+void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
+{
+    sIpFd = SocketWithCloseExec(AF_INET6, SOCK_DGRAM, IPPROTO_IP, kSocketNonBlock);
+    VerifyOrDie(sIpFd >= 0, OT_EXIT_ERROR_ERRNO);
+
+    otIcmp6SetEchoMode(aInstance, OT_ICMP6_ECHO_HANDLER_DISABLED);
+
```
gjc13, line 1472: Do we need so many new lines here?

src/posix/platform/netif.cpp
```diff
@@ -596,18 +1319,179 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
...
+#if defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
+    {
+        int status;
+#ifdef ROUTE_FILTER
+        unsigned int msgfilter = ROUTE_FILTER(RTM_IFINFO) | ROUTE_FILTER(RTM_NEWADDR) | ROUTE_FILTER(RTM_DELADDR) |
+                                 ROUTE_FILTER(RTM_NEWMADDR) | ROUTE_FILTER(RTM_DELMADDR);
+#define FILTER_CMD ROUTE_MSGFILTER
+#define FILTER_ARG msgfilter
+#define FILTER_ARG_SZ sizeof(msgfilter)
```
gjc13, line 1448: Are `FILTER_ARG_SZ` and `FILTER_ARG` the same under both cases?

gjc13:
Great work for supporting MacOS/BSD. I'm wondering if we can move some of the functions that are very different on linux and BSD to different files so that we don't need so many macro checks.

src/posix/platform/netif.cpp
rob-the-dude: Yes, in inet_ntop call for the log messages.

src/posix/platform/netif.cpp
```diff
@@ -53,6 +86,53 @@
...
+#include <sys/kern_control.h>
+#endif
+
+#ifdef __NetBSD__
+#include <net/if_tun.h>
+#endif
+
+#ifdef __FreeBSD__
+#include <net/if_tun.h>
+#endif
```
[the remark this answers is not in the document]
rob-the-dude: Done.

src/posix/platform/netif.cpp
```diff
@@ -53,6 +86,53 @@
...
+
+#include <errno.h>
```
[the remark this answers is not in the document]
rob-the-dude: Done (I think).

src/posix/platform/netif.cpp
```diff
@@ -115,31 +240,185 @@ enum
...
+
+        address = address->mNext;
+    }
+
+    return false;
+}
+#endif
+
+#if defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
+static inline void apply_mask(struct in6_addr *address, uint8_t mask)
```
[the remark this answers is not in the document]
rob-the-dude: Actually, I'm replacing these two utility routines in my next commit (and I'll fix the names).  They now make use of existing OpenThread utility functions for prefix application and length calculation.

src/posix/platform/netif.cpp
```diff
@@ -229,15 +527,35 @@ static void processStateChange(otChangedFlags aFlags, void *aContext)
...
-    char     packet[kMaxIp6Size];
-    otError  error  = OT_ERROR_NONE;
-    uint16_t length = otMessageGetLength(aMessage);
+    char     packet[kMaxIp6Size + 4];
+    otError  error      = OT_ERROR_NONE;
+    uint16_t length     = otMessageGetLength(aMessage);
+    size_t   offset     = 0;
+    uint16_t max_length = sizeof(packet) - 4;
+#if defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
+    offset += 4;
```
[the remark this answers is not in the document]
rob-the-dude: See the man pages for tun/tap on BSDs.  The kernel/userland interface uses the four bytes to distinguish IPv6 from IPv4 -- while it's not technically necessary for it to do that (it could look at the packet), that's how the interface was originally developed and it's now there for legacy reasons.  wpantund has the same code.  I'll add a comment.

src/posix/platform/netif.cpp
rob-the-dude: You are correct that it does not presently.  I had added this case during some debugging, as it seems correct -- if we did get a multicast address, then this code would handle it properly.  But the MLD listener would not be needed in that case.  Have you seen the other flavors of addressing here?  Or only IFA_ADDRESS?

src/posix/platform/netif.cpp
rob-the-dude: Fixed.

src/posix/platform/netif.cpp
rob-the-dude: Because it breaks things :-).

Stacks normally add their own link-local IPv6 address, though they may have options to avoid doing so -- but unfortunately, some stacks (mac OS in particular) don't.

When we get notified that the address has been added (as is happening here), openthread will not allow you to add this (second) link-local address to its list of addresses (I believe openthread rejects all addition of link-local addresses, as they are internally handled).  

So if we do NOT remove it, then the stack will have an address that openthread doesn't know about.

But worse than that, the stack will create packets with that address as a source address, and send them down to openthread for sending onto the network.  Other nodes will attempt to respond, but the openthread stack will not "receive" these packets because they're not bound for a known address.

src/posix/platform/netif.cpp
```diff
@@ -115,31 +240,185 @@ enum
     kICMPv6MLDv2RecordChangeToExcludeType = 3,
     kICMPv6MLDv2RecordChangeToIncludeType = 4,
 };
+#endif
 
 static const size_t kMaxIp6Size = 1536;
 
+static bool gLogPackets = false;
+
+static void dump_hex(const void *buffer, size_t len)
```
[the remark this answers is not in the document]
rob-the-dude: Done.

src/posix/platform/netif.cpp
```diff
@@ -115,31 +240,185 @@ enum
     kICMPv6MLDv2RecordChangeToExcludeType = 3,
     kICMPv6MLDv2RecordChangeToIncludeType = 4,
 };
+#endif
 
 static const size_t kMaxIp6Size = 1536;
 
+static bool gLogPackets = false;
```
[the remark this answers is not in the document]
rob-the-dude: I've added a macro in the next version.  I've also switched to using otDumpInfo.  I don't think we need it there or enabled in general, but I don't want to remove it because it is very useful in debugging interactions, especially on a new platform.

src/posix/platform/netif.cpp
```diff
@@ -115,31 +240,185 @@ enum
...
+    }
+
+    return false;
+}
+#endif
+
+#if defined(__APPLE__) || defined(__NetBSD__) || defined(__FreeBSD__)
+static inline void apply_mask(struct in6_addr *address, uint8_t mask)
+{
+    if (mask > 128)
```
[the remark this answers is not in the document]
rob-the-dude: Done.

src/posix/platform/netif.cpp
```diff
@@ -115,31 +240,185 @@ enum
...
 static void UpdateUnicast(otInstance *aInstance, const otIp6Address &aAddress, uint8_t aPrefixLength, bool aIsAdded)
 {
-    struct in6_ifreq ifr6;
-    otError          error = OT_ERROR_NONE;
+    otError error = OT_ERROR_NONE;
 
     assert(sInstance == aInstance);
 
-    VerifyOrExit(sIpFd > 0, error = OT_ERROR_INVALID_STATE);
+    VerifyOrExit(sIpFd >= 0, error = OT_ERROR_INVALID_STATE);
```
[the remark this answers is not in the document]
rob-the-dude: Why?  And if so, why don't we reserve stdout and stderr?

It seems to me that all the descriptors here (sTun, sNetlinkFd, sIpFd) should be checked if they are valid descriptors.  Indeed, when we decide to close them, that's what we check (i.e., we don't check for >0, but for != -1).

It seems possible that someone might use the posix netif code but with an alternative command-line wrapper which did choose to close stdin (for example, it only wants to read from a config file and doesn't need to keep that descriptor open).  If it did so, it is also possible that the underlying C library and/or kernel might return a value of 0 as the socket descriptor.  (I just tested such code on my Mac, and when I close stdin, and then call open, I indeed get a descriptor with a value of 0).

src/posix/platform/netif.cpp
```diff
@@ -115,31 +240,185 @@ enum
...
 
-    memcpy(&ifr6.ifr6_addr, &aAddress, sizeof(ifr6.ifr6_addr));
+#if defined(__linux__)
```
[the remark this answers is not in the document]
rob-the-dude: I thought about that, but I'm really nervous about that.  There are some functions that are indeed different, but for the most part, big chunks of the code are very similar.  I'd like to avoid significant duplication because duplicated code always diverges, which means means that at least SOME of the code needs to be in a file that's common.  I worry that if the mac/BSD implementation is separate, people will carry one of the implementations forward without the other.  By putting it here, it's easier to see when doing something Linux-specific (or Mac-specific) will break something.

src/posix/platform/netif.cpp
```diff
@@ -115,31 +240,185 @@ enum
...
+#endif
+
+        err = ioctl(sIpFd, aIsAdded ? SIOCAIFADDR_IN6 : SIOCDIFADDR_IN6, &ifr6);
+        if ((err == -1) && (errno == EALREADY))
+        {
+            err = 0;
+        }
+        VerifyOrExit(err == 0, perror("ioctl"); error = OT_ERROR_FAILED);
     }
+#endif
```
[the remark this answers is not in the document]
rob-the-dude: What are the guidelines for adding such comments after #endif ?  I appreciate their value, but I don't see them used consistently in other places.  (Indeed, in the existing version of netif.cpp, I count three #endif with comments and two without.  I am happy to add to all cases, but if I'm supposed to only do some and not the others, then I need some guidance.

src/posix/platform/netif.cpp
```diff
@@ -159,14 +438,31 @@ static void UpdateMulticast(otInstance *aInstance, const otIp6Address &aAddress,
...
-        setsockopt(sIpFd, IPPROTO_IPV6, (aIsAdded ? IPV6_JOIN_GROUP : IPV6_LEAVE_GROUP), &mreq, sizeof(mreq)) == 0,
-        perror("setsockopt");
-        error = OT_ERROR_FAILED);
+    int err;
+    err = setsockopt(sIpFd, IPPROTO_IPV6, (aIsAdded ? IPV6_JOIN_GROUP : IPV6_LEAVE_GROUP), &mreq, sizeof(mreq));
+    if ((err != 0) && (errno == EINVAL) && (IN6_IS_ADDR_MC_LINKLOCAL(&mreq.ipv6mr_multiaddr)))
+    {
+        // FIX ME
+        // on mac OS (and FreeBSD), the first time we run (but not subsequently), we get a failure on this particular
+        // join. do we need to bring up the interface at least once prior to joining?
```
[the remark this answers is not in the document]
rob-the-dude: Yes.  But I don't yet know if that's the underlying work-around or not.  On mac, I see this error the very first time I run ot-daemon.  But I do not see it on subsequent runs.  I've not had time to figure out why.

src/posix/platform/netif.cpp
```diff
@@ -159,14 +438,31 @@ static void UpdateMulticast(otInstance *aInstance, const otIp6Address &aAddress,
...
+
+        char addressString[INET6_ADDRSTRLEN + 1];
```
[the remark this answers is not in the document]
rob-the-dude: Done.

src/posix/platform/netif.cpp
```diff
@@ -159,14 +438,31 @@ static void UpdateMulticast(otInstance *aInstance, const otIp6Address &aAddress,
...
     memcpy(&mreq.ipv6mr_multiaddr, &aAddress, sizeof(mreq.ipv6mr_multiaddr));
     mreq.ipv6mr_interface = sTunIndex;
 
-    VerifyOrExit(
-        setsockopt(sIpFd, IPPROTO_IPV6, (aIsAdded ? IPV6_JOIN_GROUP : IPV6_LEAVE_GROUP), &mreq, sizeof(mreq)) == 0,
-        perror("setsockopt");
-        error = OT_ERROR_FAILED);
+    int err;
+    err = setsockopt(sIpFd, IPPROTO_IPV6, (aIsAdded ? IPV6_JOIN_GROUP : IPV6_LEAVE_GROUP), &mreq, sizeof(mreq));
+    if ((err != 0) && (errno == EINVAL) && (IN6_IS_ADDR_MC_LINKLOCAL(&mreq.ipv6mr_multiaddr)))
```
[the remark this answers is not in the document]
rob-the-dude: We already have that -- it's two lines below.  But if I don't squelch this error case on Mac and FreeBSD, the code will never proceed.  And because it dies before bringing up the interface, it will actually die every time.

I'm happy to wrap this special handing for Mac and FreeBSD (and eventually, I'd like to remove the workaround).

src/posix/platform/netif.cpp
rob-the-dude: Done.

src/posix/platform/netif.cpp
rob-the-dude: I've tried to get rid of some of them, but I like code that is logically performing a set of related steps to be separated from code that is performing a different set of steps.

src/posix/platform/netif.cpp
rob-the-dude: Yes, but that might not always be the case, if the platform's routing filter mechanism was different.

src/posix/platform/netif.cpp
```diff
@@ -177,25 +473,27 @@ static void UpdateLink(otInstance *aInstance)
 {
     otError      error = OT_ERROR_NONE;
     struct ifreq ifr;
+    bool         if_state = false;
```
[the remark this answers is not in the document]
rob-the-dude: Done.

src/posix/platform/netif.cpp
rob-the-dude: I think this is better now.

src/posix/platform/netif.cpp
rob-the-dude: Done.

src/posix/platform/netif.cpp
rob-the-dude: I consolidated a lot of the logging that was making this so complicated.  Take a peek.

src/posix/platform/netif.cpp
rob-the-dude: I consolidated a lot of the logging that was making this so complicated.  Take a peek.

src/posix/platform/netif.cpp
rob-the-dude: I consolidated a lot of the logging that was making this so complicated.  Take a peek.

src/posix/platform/netif.cpp
rob-the-dude: I consolidated a lot of the logging -- I think it's substantially better now.

src/posix/platform/netif.cpp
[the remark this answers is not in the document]
gjc13: If the `#endif` is 20 lines away from `#if` then we need to add such comments.

src/posix/platform/netif.cpp
gjc13: In fact I've only found IFA_ADDRESS and IFA_LOCAL.

PR #4665, opened 2020-03-10
[debug] adding OT_ASSERT() macro (allow disabling asserts)

abtink:

**[debug] adding OT_ASSERT() macro**

This commit adds a new OpenThread specific `OT_ASSERT()` macro
along with a new config OPENTHREAD_CONFIG_ASSERT_ENABLE option which
allows the assert to be enabled or disabled. This allows a release
build to consider disabling the assert reducing code size.

**[travis] add new simul platform build with asserts disabled**

### DIFF
--- a/src/core/common/debug.hpp
+++ b/src/core/common/debug.hpp
@@ -39,9 +39,13 @@
 #include <ctype.h>
 #include <stdio.h>
 
+#if OPENTHREAD_CONFIG_ASSERT_ENABLE
+
 #if defined(OPENTHREAD_TARGET_DARWIN) || defined(OPENTHREAD_TARGET_LINUX)
 
 #include <assert.h>
+
+#define OT_ASSERT(cond) assert(cond)
 
 #elif OPENTHREAD_CONFIG_PLATFORM_ASSERT_MANAGEMENT
 
@@ -55,11 +59,7 @@
 #define FILE_NAME __FILE__
 #endif
 
-#ifdef assert
-#undef assert
-#endif
-
-#define assert(cond)                               \
+#define OT_ASSERT(cond)                            \
     do                                             \
     {                                              \
         if (!(cond))                               \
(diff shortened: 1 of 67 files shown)


src/core/common/debug.hpp
```diff
@@ -71,23 +71,25 @@
...
+            while (1)   \
+            {           \
+            }           \
+        }               \
+    } while (0)
 
-#ifdef assert
-#undef assert
-#endif
+#endif // #elif OPENTHREAD_CONFIG_PLATFORM_ASSERT_MANAGEMENT
```
jwhui: nit: we generally do not include the `#elif` in the comment

bukepo:
Could we mimic the standard `assert()` behavior? i.e. use the same name and disabled by the same definition `NDEBUG`?

src/core/common/debug.hpp
```diff
@@ -73,21 +73,23 @@
...
+            }           \
+        }               \
     } while (0)
 
-#endif
+#endif // OPENTHREAD_CONFIG_PLATFORM_ASSERT_MANAGEMENT
+
+#else
+
+#define OT_ASSERT(cond)
```
simonlingoogle:
```suggestion
#define OT_ASSERT(cond) ((void)0)
```
otherwise some lint program may complain about empty statements.

abtink:
> Could we mimic the standard assert() behavior? i.e. use the same name and disabled by the same definition NDEBUG?

The main reason for adding an OT specific assert `OT_ASSERT()` macro (instead of the `assert()`) is to avoid any macro/function name collisions with other definitions from third-party or platform layer code (e.g., the issue that was fixed from https://github.com/openthread/openthread/pull/4329 where platform layer asserts (intended for mbed-tls module) were being used instead). 

With an `OT_ASSERT()` macro, it is clear it is OT specific assert and we allow platform layer to customize the behavior (using `OPENTHREAD_CONFIG_PLATFORM_ASSERT_MANAGEMENT`) on how to  handle asserts in OT core may be even differently from asserts from other modules.

Thoughts?

src/core/common/debug.hpp
abtink: We actually do use a lot of empty macros for OT logging (`otLogDebg`, etc.) already.
I would suggest we keep this as is and use empty macro here as well. We can change them all if/when there is an issue with using empty macros.

PR #4672, opened 2020-03-11
[energy-scan] avoid buffer overrun when receiving energy results

jwhui:


src/core/thread/energy_scan_server.cpp
```diff
@@ -140,6 +140,7 @@ void EnergyScanServer::HandleScanResult(Mac::EnergyScanResult *aResult)
 
     if (aResult)
     {
+        VerifyOrExit(mScanResultsLength < OPENTHREAD_CONFIG_TMF_ENERGY_SCAN_MAX_RESULTS);
```
bukepo: should we use `%` to overwrite an oldest scan result?

PR #4674, opened 2020-03-11
[key-manager] add Kek class

abtink:
This commit adds a class `Kek` in `KeyManager` representing a Key Encryption Key.

### DIFF
--- a/src/core/thread/key_manager.hpp
+++ b/src/core/thread/key_manager.hpp
@@ -138,6 +138,55 @@ public:
 } OT_TOOL_PACKED_END;
 
 /**
+ *
+ * This class represents a Key Encryption Key (KEK).
+ *
+ */
+class Kek
+{
+    friend class KeyManager;
+
+public:
+    enum
+    {
+        kSize = 16, // KEK size in bytes.
+    };
+
+    /**
+     * This method returns the KEK.
+     *
+     * @returns A pointer to buffer containing the KEK.
+     *
+     */
+    const uint8_t *GetKey(void) const { return m8; }
+
+    /**
+     * This method evaluates whether or not two KEKs match.
+     *
+     * @param[in]  aOther  The KEK to compare.
+     *
+     * @retval TRUE   If the KEKs match.
+     * @retval FALSE  If the KEKs do not match.
+     *
+     */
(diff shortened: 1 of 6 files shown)


src/core/thread/key_manager.hpp
```diff
@@ -137,6 +137,55 @@ class Pskc : public otPskc
 
 } OT_TOOL_PACKED_END;
 
+/**
+ *
+ * This class represents a Key Encryption Key (KEK).
+ *
+ */
+class Kek
```
wgtdkp: nit: should it be packed? since we will map TLV value to it.

jwhui: This class is not used in a TLV, so I don't think it needs to be packed.

wgtdkp: @jwhui we read KEK value drectly to class `Kek` at here: https://github.com/openthread/openthread/pull/4674/files#diff-463fc3bfeedf929024c4d31de1707e85R221

I think this has a requirement of being packed or we should read to `ket.GetKek()`.

jwhui: In what case would `&kek` and `kek.GetKek()` not return the same address?

wgtdkp: Yes, they always return the same address and being packed makes no difference here. But, I am not sure if it is a convention in openthread that a struct _should_(even if it is not necessary when look at the struct definition) be marked packed when it is used to directly map to/from byte array(`void*` or `uint8_t*`)?

abtink: I agree. I don't think we need to define it as  packed.  We actually use packed for two different reasons: 
- First to ensure there is no padding bytes in between member variables.
- Second is to allow us to cast any pointer to our packed  type and not worry about or be restricted by the alignment requirements.

For the `Kek` type, we won't have either situation:
- Since the type is PoD (plain old data), C++ standard requires that a pointer to it to be same as pointer to to its first member variable. (so `&kek` and `kek.m8` should always be same).
- The `Tlv::ReadTlv()` method would not do any casting/mapping of types or pointers. It will copy the TLV content/value (from the message) as a byte sequence into the given buffer/pointer.

wgtdkp: clear :+1: Thanks @jwhui @abtink

PR #4676, opened 2020-03-12
[spinel] Include pre-defines before verify defines

kghost:
Signed-off-by: Zang MingJie <mingjiez@google.com>

### DIFF
--- a/src/lib/spinel/openthread-spinel-config.h
+++ b/src/lib/spinel/openthread-spinel-config.h
@@ -34,6 +34,8 @@
 #ifndef OPENTHREAD_SPINEL_CONFIG_H_
 #define OPENTHREAD_SPINEL_CONFIG_H_
 
+#include <openthread-core-config.h>
+
 /**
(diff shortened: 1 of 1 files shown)


src/lib/spinel/openthread-spinel-config.h
```diff
@@ -34,6 +34,8 @@
 #ifndef OPENTHREAD_SPINEL_CONFIG_H_
 #define OPENTHREAD_SPINEL_CONFIG_H_
 
+#include <openthread-core-config.h>
```
bukepo: We intended to remove dependency on core config.

kghost:
This is causing problems, since the define being checked in this file should be defined before this file.

How to resolve the problem ?

bukepo:
I think you can define in CPPFLAGS.

jwhui:
@kghost, any plan to move this PR forward?

jwhui:
Closing stale PR.

PR #4677, opened 2020-03-12
[ip6-address] adding SetPrefix() to set the IPv6 address prefix

abtink:
This commit adds a new method `SetPrefix()` in `Ip6::Address` class to
 allow the prefix in an IPv6 address to be updated. Unit test
`test_ip6_address` is updated to verify behavior of new method. The
`Mle` and `Lowpan` classes are updated to use the new `SetPrefix()`
method.

### DIFF
--- a/src/core/net/ip6_address.cpp
+++ b/src/core/net/ip6_address.cpp
@@ -144,6 +144,28 @@ bool Address::IsIidReserved(void) const
     return IsSubnetRouterAnycast() || IsReservedSubnetAnycast() || IsAnycastRoutingLocator();
 }
 
+void Address::SetPrefix(const uint8_t *aPrefix, uint8_t aPrefixLength)
+{
+    uint8_t extraBits = (aPrefixLength % CHAR_BIT);
+
+    OT_ASSERT(aPrefixLength <= sizeof(Address) * CHAR_BIT);
+
+    memcpy(mFields.m8, aPrefix, (aPrefixLength - extraBits) / CHAR_BIT);
+
+    if (extraBits > 0)
+    {
+        uint8_t index = aPrefixLength / CHAR_BIT;
+        uint8_t mask  = ((0x80 >> (extraBits - 1)) - 1);
+
+        // `mask` has its higher (msb) `extraBits` bits as `0` and the reminaing as `1`.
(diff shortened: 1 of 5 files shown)


src/core/net/ip6_address.cpp
```diff
@@ -144,6 +144,26 @@ bool Address::IsIidReserved(void) const
     return IsSubnetRouterAnycast() || IsReservedSubnetAnycast() || IsAnycastRoutingLocator();
 }
 
+void Address::SetPrefix(const uint8_t *aPrefix, uint8_t aPrefixLength)
+{
+    uint8_t extraBits = (aPrefixLength % CHAR_BIT);
+
+    memcpy(mFields.m8, aPrefix, (aPrefixLength - extraBits) / CHAR_BIT);
```
simonlingoogle: need to make sure `aPrefixLength <= 128`?

tests/unit/test_ip6_address.cpp
```diff
@@ -134,8 +136,92 @@ void TestIp6AddressFromString(void)
     }
 }
 
+bool CheeckPrefix(const ot::Ip6::Address &aAddress, const uint8_t *aPrefix, uint8_t aPrefixLength)
```
simonlingoogle:
```suggestion
bool CheckPrefix(const ot::Ip6::Address &aAddress, const uint8_t *aPrefix, uint8_t aPrefixLength)
```

tests/unit/test_ip6_address.cpp
```diff
@@ -134,8 +136,92 @@ void TestIp6AddressFromString(void)
...
+        {
+            matches = false;
+            break;
+        }
+    }
+
+    return matches;
+}
+
+bool CheckAddress(const ot::Ip6::Address &aAddress1, const ot::Ip6::Address &aAddress2, uint8_t aPrefixLength)
```
simonlingoogle: Should be renamed to `CheckInterfaceId`? I think the name `CheckAddress` is misleading.

src/core/net/ip6_address.cpp
abtink: Added an assert

PR #4681, opened 2020-03-13
[core/common] Add an option to move instance on heap

kghost:
There is limit data segment on some FreeRTOS platform, because the
instance is very large, we may want to move the instance from static
data segment to dynamic allocated heap.

The patch introduced an new define option:

OPENTHREAD_CONFIG_INSTANCE_ON_HEAP

When it is defined, the instance will be allocated dynamically on heap.

Signed-off-by: Zang MingJie <mingjiez@google.com>

### DIFF
--- a/src/core/common/instance.cpp
+++ b/src/core/common/instance.cpp
@@ -43,7 +43,11 @@ namespace ot {
 #if !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
 
 // Define the raw storage used for OpenThread instance (in single-instance case).
+#if !OPENTHREAD_CONFIG_INSTANCE_ON_HEAP
 OT_DEFINE_ALIGNED_VAR(gInstanceRaw, sizeof(Instance), uint64_t);
+#else
+Instance *gInstanceRaw = NULL;
+#endif
 
 #endif
 
@@ -116,11 +120,21 @@ Instance &Instance::InitSingle(void)
 {
     Instance *instance = &Get();
 
+#if !OPENTHREAD_CONFIG_INSTANCE_ON_HEAP
     VerifyOrExit(!instance->mIsInitialized);
 
     instance = new (&gInstanceRaw) Instance();
 
     instance->AfterInit();
+#else
+    VerifyOrExit(instance == NULL);
+
+    gInstanceRaw = new Instance();
+
+    instance = gInstanceRaw;
+
+    instance->AfterInit();
+#endif
 
 exit:
     return *instance;
@@ -128,9 +142,11 @@ exit:
 
 Instance &Instance::Get(void)
 {
-    void *instance = &gInstanceRaw;
-
-    return *static_cast<Instance *>(instance);
+#if !OPENTHREAD_CONFIG_INSTANCE_ON_HEAP
+    return *static_cast<Instance *>((void*)&gInstanceRaw);
+#else
+    return *gInstanceRaw;
+#endif
 }
 
 #else // #if !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
--- a/src/core/common/locator.hpp
+++ b/src/core/common/locator.hpp
@@ -45,7 +45,11 @@ namespace ot {
 class Instance;
 
 #if !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
+#if !OPENTHREAD_CONFIG_INSTANCE_ON_HEAP
 extern uint64_t gInstanceRaw[];
+#else
+extern Instance *gInstanceRaw;
+#endif
 #endif
 
 /**
@@ -83,7 +87,11 @@ public:
 #if OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
     Instance &GetInstance(void) const { return *mInstance; }
 #else
+#if !OPENTHREAD_CONFIG_INSTANCE_ON_HEAP
     Instance &GetInstance(void) const { return *reinterpret_cast<Instance *>(&gInstanceRaw); }
+#else
+    Instance &GetInstance(void) const { return *gInstanceRaw; }
+#endif
 #endif
 
     /**
--- a/src/core/config/openthread-core-default-config.h
+++ b/src/core/config/openthread-core-default-config.h
@@ -108,6 +108,16 @@
 #endif
 
 /**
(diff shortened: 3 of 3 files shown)


abtink:
Wouldn't the data (or heap) segment commonly be configurable on platforms? 

Having statically allocated instance has the advantage that the used/required memory/RAM by OT library can be tracked and verified during build time, vs. the heap/dynamic allocation where the code may need to add checks at the run-time.

kghost:
At least, it is not configurable on ESP32 platform.

simonlingoogle:
> At least, it is not configurable on ESP32 platform.

This is actually true according to [ESP32 Reference](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/mem_alloc.html#available-heap): 
> Due to a technical limitation, the maximum statically allocated DRAM usage is 160KB. The remaining 160KB (for a total of 320KB of DRAM) can only be allocated at runtime as heap.
@abtink

src/core/common/instance.cpp
```diff
@@ -116,21 +120,33 @@ Instance &Instance::InitSingle(void)
...
+#if !OPENTHREAD_CONFIG_INSTANCE_ON_HEAP
     VerifyOrExit(!instance->mIsInitialized);
 
     instance = new (&gInstanceRaw) Instance();
 
     instance->AfterInit();
+#else
+    VerifyOrExit(instance == NULL);
+
+    gInstanceRaw = new Instance();
```
simonlingoogle: Should make sure `gInstanceRaw` is aligned on 64bit boundary?

src/core/common/instance.cpp
```diff
@@ -116,21 +120,33 @@ Instance &Instance::InitSingle(void)
...
+
+    instance = gInstanceRaw;
+
+    instance->AfterInit();
```
simonlingoogle: move this out of `#if #else #endif` ?

src/core/common/instance.cpp
```diff
@@ -43,7 +43,11 @@ namespace ot {
 #if !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
 
 // Define the raw storage used for OpenThread instance (in single-instance case).
+#if !OPENTHREAD_CONFIG_INSTANCE_ON_HEAP
 OT_DEFINE_ALIGNED_VAR(gInstanceRaw, sizeof(Instance), uint64_t);
+#else
+Instance *gInstanceRaw = NULL;
```
simonlingoogle: Use `uint64_t *gInstanceRaw` for more consistency?

jwhui:
If a platform must use dynamically allocated memory for the OpenThread instance, how about enabling multiple instance support and using: https://github.com/openthread/openthread/blob/132df25505604a0e9793c26d61002ca428e8634f/include/openthread/instance.h#L80

src/core/common/locator.hpp
```diff
@@ -83,7 +87,11 @@ class InstanceLocator
 #if OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
     Instance &GetInstance(void) const { return *mInstance; }
 #else
+#if !OPENTHREAD_CONFIG_INSTANCE_ON_HEAP
     Instance &GetInstance(void) const { return *reinterpret_cast<Instance *>(&gInstanceRaw); }
+#else
+    Instance &GetInstance(void) const { return *gInstanceRaw; }
```
simonlingoogle: Should able to use same implementation for both case when `gInstanceRaw` is defined to `uint64_t *`.

src/core/common/locator.hpp
```diff
@@ -45,7 +45,11 @@ namespace ot {
 class Instance;
 
 #if !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
+#if !OPENTHREAD_CONFIG_INSTANCE_ON_HEAP
 extern uint64_t gInstanceRaw[];
+#else
+extern Instance *gInstanceRaw;
```
simonlingoogle:
```suggestion
extern uint64_t *gInstanceRaw;
```

src/core/common/instance.cpp
```diff
@@ -116,21 +120,33 @@ Instance &Instance::InitSingle(void)
...
 
 Instance &Instance::Get(void)
 {
-    void *instance = &gInstanceRaw;
-
-    return *static_cast<Instance *>(instance);
+#if !OPENTHREAD_CONFIG_INSTANCE_ON_HEAP
+    return *static_cast<Instance *>((void*)&gInstanceRaw);
+#else
+    return *gInstanceRaw;
```
simonlingoogle: Should able to use same implementation for both case when `gInstanceRaw` is defined to `uint64_t *`.

src/core/config/openthread-core-default-config.h
```diff
@@ -107,6 +107,16 @@
 #define OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE 0
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
```
abtink, line 111: nit: change to `OPENTHREAD_CONFIG_INSTANCE_ON_HEAP`

src/core/common/instance.cpp
```diff
@@ -116,21 +120,33 @@ Instance &Instance::InitSingle(void)
 {
     Instance *instance = &Get();
 
+#if !OPENTHREAD_CONFIG_INSTANCE_ON_HEAP
     VerifyOrExit(!instance->mIsInitialized);
```
wgtdkp, line 124: The instance is referenced before it is allocated at below.

src/core/common/instance.cpp
```diff
@@ -116,21 +120,33 @@ Instance &Instance::InitSingle(void)
...
     return *instance;
 }
 
 Instance &Instance::Get(void)
 {
-    void *instance = &gInstanceRaw;
-
-    return *static_cast<Instance *>(instance);
+#if !OPENTHREAD_CONFIG_INSTANCE_ON_HEAP
+    return *static_cast<Instance *>((void*)&gInstanceRaw);
```
wgtdkp, line 146:
```suggestion
    return *static_cast<Instance *>(gInstanceRaw);
```

wgtdkp:
@kghost @jwhui 
With https://github.com/openthread/openthread/pull/4915, I think we can close this PR now.

kghost:
Yes, multiple instance is better solution for this

PR #4683, opened 2020-03-13
[mle] define Metadata (for delaying response) as inner type of Mle class

abtink:
This PR contains two small related commits. Kept them seperate to make it easier to read/review (but would suggest to squash for merge).

- Define `Metadata` (for delaying MLE responses) as innter type of `Mle` class
- Simplify `Header` class methods (move `enum` definitions to top of `class`).

### DIFF
--- a/src/core/thread/mle.hpp
+++ b/src/core/thread/mle.hpp
@@ -91,6 +91,43 @@ class Header
 {
 public:
     /**
+     * MLE Command Types.
+     *
+     */
+    enum Command
+    {
+        kCommandLinkRequest          = 0,  ///< Link Request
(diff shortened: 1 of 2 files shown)


src/core/thread/mle.hpp
```diff
@@ -1775,6 +1686,16 @@ class Mle : public InstanceLocator
         kDataRequestActive, ///< Data Request has been sent, Data Response is expected.
     };
 
+    struct Metadata // for delayed response
```
jwhui: How about something more descriptive, like `ResponseInfo`?

abtink: Sounds good. I changed it to `DelayedResponseMetadata`.

PR #4687, opened 2020-03-16
[posix] add support for multicast group join

gjc13:
Listen for mld reports sent by kernel to capture user multicast group
join and forward the addresses to OpenThread interface.

See https://github.com/openthread/wpantund/pull/444 for more background.

### DIFF
--- a/src/posix/platform/netif.cpp
+++ b/src/posix/platform/netif.cpp
@@ -438,28 +438,24 @@ void platformNetifDeinit(void)
     if (sMLDMonitorFd != -1)
     {
         close(sMLDMonitorFd);
-        sNetlinkFd = -1;
+        sMLDMonitorFd = -1;
     }
 
     sTunIndex = 0;
 }
 
-static otError mldListenerInit(void)
+static void mldListenerInit(void)
 {
     struct ipv6_mreq mreq6;
-    otError          err = OT_ERROR_NONE;
 
     sMLDMonitorFd          = SocketWithCloseExec(AF_INET6, SOCK_RAW | SOCK_NONBLOCK, IPPROTO_ICMPV6);
     mreq6.ipv6mr_interface = sTunIndex;
     memcpy(&mreq6.ipv6mr_multiaddr, kMLDv2MulticastAddress.mFields.m8, sizeof(kMLDv2MulticastAddress.mFields.m8));
 
-    VerifyOrExit(setsockopt(sMLDMonitorFd, IPPROTO_IPV6, IPV6_JOIN_GROUP, &mreq6, sizeof(mreq6)) == 0,
-                 err = OT_ERROR_FAILED);
-    VerifyOrExit(setsockopt(sMLDMonitorFd, SOL_SOCKET, SO_BINDTODEVICE, sTunName,
-                            static_cast<socklen_t>(strnlen(sTunName, IFNAMSIZ))) == 0,
-                 err = OT_ERROR_FAILED);
-exit:
-    return err;
+    VerifyOrDie(setsockopt(sMLDMonitorFd, IPPROTO_IPV6, IPV6_JOIN_GROUP, &mreq6, sizeof(mreq6)) == 0, OT_EXIT_FAILURE);
+    VerifyOrDie(setsockopt(sMLDMonitorFd, SOL_SOCKET, SO_BINDTODEVICE, sTunName,
+                           static_cast<socklen_t>(strnlen(sTunName, IFNAMSIZ))) == 0,
+                OT_EXIT_FAILURE);
 }
 
 static void processMLDEvent(otInstance *aInstance)
@@ -565,10 +561,10 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
     struct ifreq ifr;
 
     sIpFd = SocketWithCloseExec(AF_INET6, SOCK_DGRAM, IPPROTO_IP);
-    VerifyOrExit(sIpFd >= 0);
+    VerifyOrDie(sIpFd >= 0, OT_EXIT_ERROR_ERRNO);
 
     sNetlinkFd = socket(AF_NETLINK, SOCK_DGRAM, NETLINK_ROUTE);
-    VerifyOrExit(sNetlinkFd > 0);
+    VerifyOrDie(sNetlinkFd > 0, OT_EXIT_ERROR_ERRNO);
 
     otIcmp6SetEchoMode(aInstance, OT_ICMP6_ECHO_HANDLER_DISABLED);
 
@@ -578,11 +574,11 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
         memset(&sa, 0, sizeof(sa));
         sa.nl_family = AF_NETLINK;
         sa.nl_groups = RTMGRP_LINK | RTMGRP_IPV6_IFADDR;
-        VerifyOrExit(bind(sNetlinkFd, reinterpret_cast<struct sockaddr *>(&sa), sizeof(sa)) == 0);
+        VerifyOrDie(bind(sNetlinkFd, reinterpret_cast<struct sockaddr *>(&sa), sizeof(sa)) == 0, OT_EXIT_ERROR_ERRNO);
     }
 
     sTunFd = open(OPENTHREAD_POSIX_TUN_DEVICE, O_RDWR | O_CLOEXEC);
-    VerifyOrExit(sTunFd > 0, otLogCritPlat("Unable to open tun device %s", OPENTHREAD_POSIX_TUN_DEVICE));
+    VerifyOrDie(sTunFd > 0, OT_EXIT_ERROR_ERRNO);
 
     memset(&ifr, 0, sizeof(ifr));
     ifr.ifr_flags = IFF_TUN | IFF_NO_PI;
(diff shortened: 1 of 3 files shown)


.travis/check-posix-pty
```diff
@@ -129,6 +130,10 @@ EOF
     extaddr=$(awk '/extaddr/{getline; print}' $OT_OUTPUT | tr -d '\r\n')
     echo "Extended address is: ${extaddr}"
 
+    # Test join multicast group
+    NETIF_INDEX=$(ip link show "${VALID_NETIF_NAME}" | cut -f 1 -d ":" | head -n 1)
+    python3 "$(pwd)/.travis/test_multicast_join.py" "${NETIF_INDEX}" || die 'multicast group join failed'
```
bukepo: It seems this doesn't verify whether the multicast address is added to OpenThread's external multicast address list.

src/posix/platform/netif.cpp
```diff
@@ -79,12 +79,55 @@ struct in6_ifreq
...
+};
+
+struct MLDv2Header
+{
+    uint8_t  mType;
+    uint8_t  _rsv0;
+    uint16_t mChecksum;
+    uint16_t _rsv1;
+    uint16_t mNumRecords;
+} __attribute__((packed));
```
bukepo: Use `OT_TOOL_PACKED_BEGIN`

src/posix/platform/netif.cpp
```diff
@@ -79,12 +79,55 @@ struct in6_ifreq
...
+static int          sMLDMonitorFd = -1; ///< Used to receive MLD events.
+static unsigned int sTunIndex     = 0;
 static char         sTunName[IFNAMSIZ];
+// ff02::16
+static const otIp6Address kMLDv2MulticastAddress = {
+    {{0xff, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x16}}};
+
+static const otIp6Address kFilterMulticastAddresses[] = {
+    kMLDv2MulticastAddress,
+    // ff03::fc
```
bukepo: Could you add comment for each address, explaining why they must be filtered?

src/posix/platform/netif.cpp
```diff
@@ -384,6 +427,122 @@ static void processNetifEvent(otInstance *aInstance)
...
+                {
+                    otIp6SubscribeMulticastAddress(aInstance, &address);
+                }
+                else if (record->mRecordType == kICMPv6MLDv2RecordChangeToExcludeType)
+                {
+                    otIp6UnsubscribeMulticastAddress(aInstance, &address);
+                }
+            }
+
+            offset += sizeof(MLDv2Record) + sizeof(in6_addr) * record->mNumSources;
```
abtink: Do we need to care about the encoding (little/big endian) of the `mNumSources` field? 
We cast a received buffer pointer to `MLDv2Record` and then read the field as `uint16_t` (I guess the complier/platform endian-ness would come into picture when we read it).

src/posix/platform/netif.cpp
```diff
@@ -384,6 +427,122 @@ static void processNetifEvent(otInstance *aInstance)
...
+                break;
+            }
+        }
+    }
+    VerifyOrExit(fromSelf);
+
+    hdr    = reinterpret_cast<MLDv2Header *>(buffer);
+    offset = sizeof(MLDv2Header);
+
+    for (size_t i = 0; i < hdr->mNumRecords && offset < static_cast<size_t>(bufferLen); i++)
```
abtink: Same question about encoding (big/little endian) here for `mNumRecords`?

src/posix/platform/netif.cpp
```diff
@@ -384,6 +427,122 @@ static void processNetifEvent(otInstance *aInstance)
...
+    {
+        if (static_cast<size_t>(bufferLen) >= (sizeof(MLDv2Record) + offset))
+        {
+            MLDv2Record *record = reinterpret_cast<MLDv2Record *>(&buffer[offset]);
+
+            if (!isMulticastAddressFiltered(record->mMulticastAddress))
```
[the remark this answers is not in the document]
rob-the-dude: FWIW, in the mac OS code I included in #4664, I do catch the failures and AT LEAST log them.  I really hate that these things fail silently.  I don't mind if the system tries to keep running in a "not completely working" state, but we should at least complain about it at runtime.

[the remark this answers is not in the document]
jwhui: I agree that we should log any errors returned from `otIp6SubscribeMulticastAddress()` and `otIp6UnsubscribeMulticastAddress()`.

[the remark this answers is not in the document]
rob-the-dude: Also FWIW, I need to develop a better understanding of how OpenThread's logging mechanisms work.  My code just spits stuff to stderr using fprintf -- since this code is on POSIX, that turns out to be OK.  But I'm suspicious there's a way to use some ot logging APIs and get the same results (i.e., errors and interesting results get spit to stderr).  Using that seems "more correct" for this codebase.  And I'm hopeful that I wouldn't have to provide my own "IPv6-address-to-string" function if I was using the preferred way.  But back when I started on this effort (almost a year ago), the ot logging interfaces were more challenging and less full-featured that I could deal with.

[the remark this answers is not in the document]
gjc13: I filter these addresses to prevent subscribing ff02::02 when thread is enabled. I'm not quit sure about  the current codes behavior under such circumstance but I decided to filter them to be safe.

src/posix/platform/netif.cpp
gjc13: You are right, we need to add ntohs. It was not causing a crash because of the data length guard.

src/posix/platform/netif.cpp
```diff
@@ -434,6 +604,7 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
 #if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
     platformUdpInit(sTunName);
 #endif
+    mldListenerInit();
```
rob-the-dude: Seems like we should also have a check for:

if (sMLDMonitorFd != -1)
{
     close(sMLDMonitorFd);
     sMLDMonitorFd = -1;
}

at the end of this routine, right?

src/posix/platform/netif.cpp
```diff
@@ -384,6 +429,131 @@ static void processNetifEvent(otInstance *aInstance)
...
+                otError      err;
+                otIp6Address address;
+
+                memcpy(&address.mFields.m8, &record->mMulticastAddress, sizeof(address.mFields.m8));
+                if (record->mRecordType == kICMPv6MLDv2RecordChangeToIncludeType)
+                {
+                    err = otIp6SubscribeMulticastAddress(aInstance, &address);
+                    if (err != OT_ERROR_NONE)
+                    {
+                        otLogCritPlat("Not subscribing multicast address: %s", otThreadErrorToString(err));
```
bukepo:
```suggestion
                        otLogWarnPlat("Failed to subscribe %s: %s", ip, otThreadErrorToString(err));
```

src/posix/platform/netif.cpp
```diff
@@ -384,6 +429,131 @@ static void processNetifEvent(otInstance *aInstance)
...
+                    }
+                }
+                else if (record->mRecordType == kICMPv6MLDv2RecordChangeToExcludeType)
+                {
+                    err = otIp6UnsubscribeMulticastAddress(aInstance, &address);
+                    if (err != OT_ERROR_NONE)
+                    {
+                        otLogCritPlat("Not subscribing multicast address: %s", otThreadErrorToString(err));
```
bukepo:
```suggestion
                        otLogWarnPlat("Failed to unsubscribe %s: %s", ip, otThreadErrorToString(err));
```

src/posix/platform/netif.cpp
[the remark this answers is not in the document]
abtink: Logging the error from the `otIp6SubscribeMulticastAddress()` and `otIp6UnsubscribeMulticastAddress()` sounds good to me, but the question for me is if we want to do the filtering of addresses in this module or not.

I would say, unless we change the code in this module to `assert` on an error,  it may be better not to filter/exclude any multicast address and log the error. Particularly since the change in https://github.com/openthread/openthread/pull/4693 makes the returned error more appropriate when address is already present on the list. We then treat `OT_ERROR_ALREADY` and `OT_ERROR_NONE` as ok/success cases.

About `ff02::02` the OT API itself should not allow an external subscription for it and I expect would return the appropriate error code (OT_ERROR_ALREADY if we are FTD (router/reed), and OT_ERROR_INVALID_ARGS if we are MTD). 

Thoughts?

src/posix/platform/netif.cpp
gjc13: In `mldListenerInit`, we die on errors.

src/posix/platform/netif.cpp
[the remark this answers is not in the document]
gjc13: If OT will do the filtering than that looks fine to me.

src/posix/platform/netif.cpp
rob-the-dude: Fair enough, though it seems like that violates the symmetry of the code.  Some routines succeed or jump to exit on error.  Some die.  Why the difference?  Why not make them all the same, and clean up after all of them on exit?

gjc13: The difference lies on whether it is critical for further code executions.
For instance, an error in parsing the MLD report won't affect the future packet parsing while not being able to setup the listener will prevent you from parsing any MLD packets.

gjc13: @rob-the-dude I found you are right when I was working on https://github.com/openthread/openthread/pull/4726.
Would you mind to wait until https://github.com/openthread/openthread/pull/4726 is merged so that we can elegantly handle deinitialization?

rob-the-dude: Sure.  I think #4726 is a good solution to making this cleaner and easier to understand.  Thank you.

src/posix/platform/netif.cpp
```diff
@@ -598,17 +601,16 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
         strncpy(ifr.ifr_name, "wpan%d", IFNAMSIZ);
     }
 
-    VerifyOrExit(ioctl(sTunFd, TUNSETIFF, static_cast<void *>(&ifr)) == 0,
+    VerifyOrExit(ioctl(sTunFd, TUNSETIFF, static_cast<void *>(&ifr)) == 0, err = OT_ERROR_FAILED;
                  otLogCritPlat("Unable to configure tun device %s", OPENTHREAD_POSIX_TUN_DEVICE));
```
bukepo:
```suggestion
    VerifyOrDie(ioctl(sTunFd, TUNSETIFF, static_cast<void *>(&ifr)) == 0, OT_EXIT_ERROR_ERRNO);
```

src/posix/platform/netif.cpp
```diff
@@ -598,17 +601,16 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
...
-    VerifyOrExit(ioctl(sTunFd, TUNSETLINK, ARPHRD_VOID) == 0,
+    VerifyOrExit(ioctl(sTunFd, TUNSETLINK, ARPHRD_VOID) == 0, err = OT_ERROR_FAILED;
                  otLogCritPlat("Unable to set link type of tun device %s", OPENTHREAD_POSIX_TUN_DEVICE));
```
bukepo:
```suggestion
    VerifyOrDie(ioctl(sTunFd, TUNSETLINK, ARPHRD_VOID) == 0, OT_EXIT_ERROR_ERRNO);
```

src/posix/platform/netif.cpp
```diff
@@ -598,17 +601,16 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
...
 
     strncpy(sTunName, ifr.ifr_name, sizeof(sTunName));
-    SuccessOrExit(mldListenerInit());
-
     sTunIndex = if_nametoindex(ifr.ifr_name);
-    VerifyOrExit(sTunIndex > 0);
+    VerifyOrExit(sTunIndex > 0, err = OT_ERROR_FAILED);
```
bukepo:
```suggestion
    VerifyOrDie(sTunIndex > 0, OT_EXIT_ERROR_ERRNO);
```

gjc13:
@rob-the-dude I've updated all the error check in initialization to die instantly.

gjc13:
@bukepo @abtink @rob-the-dude 
Rebased to latest master and fixed the check errors.

bukepo:
Similar to https://github.com/openthread/openthread/pull/5098, I think we should intercept the MLD going out via the Thread interface and translate them into Thread API calls. Thus, the MLD message are filtered out and won't bother other Thread nodes in the network. Thoughts?

jwhui:
> Once the host listens on a multicast address on the interface, the kernel will send an MLDv2 packet to `ff02::16%wpan0`. Since we are forwarding this group join information via `otIp6SubscribeMulticastAddress`, there is no need to forward this multicast packet into the Thread network. Thus this proposal is quite reasonable.

I realize now that @bukepo is proposing to move this logic down the stack where we process packets coming in via the Thread network interface. I think that makes sense.

I was confused because I thought what @bukepo proposed ("intercept the MLD going out via the Thread interface and translate them into Thread API calls") was what this PR already achieved.

PR #4693, opened 2020-03-17
[netif] enhance return value of adding external addresses

gjc13:
This commit changes the return value of subscribing a pre-defined address when it's already subscribed. Now returns OT_ERROR_ALREADY instead of OT_ERROR_INVALID_ARGS.

Resolves the return value issue in https://github.com/openthread/openthread/pull/4664

### DIFF
--- a/src/core/net/netif.cpp
+++ b/src/core/net/netif.cpp
@@ -363,6 +363,8 @@ otError Netif::SubscribeExternalMulticast(const Address &aAddress)
     NetifMulticastAddress &linkLocalAllRoutersAddress = static_cast<NetifMulticastAddress &>(
         const_cast<otNetifMulticastAddress &>(kLinkLocalAllRoutersMulticastAddress));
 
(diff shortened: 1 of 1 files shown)


src/core/net/netif.cpp
```diff
@@ -367,13 +367,13 @@ otError Netif::SubscribeExternalMulticast(const Address &aAddress)
     // LinkLocalAllRouters -> RealmLocalAllRouters -> LinkLocalAllNodes
     // -> RealmLocalAllNodes -> RealmLocalAllMpl.
 
+    VerifyOrExit(!IsMulticastSubscribed(aAddress), error = OT_ERROR_ALREADY);
```
bukepo: move above line 366?

jwhui:
> I tried these changes locally, and they address my needs in #4664.

@rob-the-dude, thanks for testing and reporting back!

rob-the-dude:
> Should we also change unicast as well to behave similarly? 

That's my preference, for whatever that's worth.

PR #4694, opened 2020-03-17
[build] cmake force C++ warning as error

wgtdkp:


CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "(Apple)?[Cc]lang")
 
     set(OT_CFLAGS
         $<$<COMPILE_LANGUAGE:C>:${OT_CFLAGS} -Wall -Wextra -Wshadow -Werror -pedantic-errors>
-        $<$<COMPILE_LANGUAGE:CXX>:${OT_FLAGS} -Wall -Wextra -Wshadow -Wno-c++14-compat -fno-exceptions>
+        $<$<COMPILE_LANGUAGE:CXX>:${OT_FLAGS} -Wall -Wextra -Wshadow -Werror -Wno-c++14-compat -fno-exceptions>
```
jwhui: The `-Werror` was intentionally left out from https://github.com/openthread/openthread/pull/4500.

@CiaranWoodward made the following comment in https://github.com/openthread/openthread/pull/4500#pullrequestreview-347974498:

> -Werror
I guess this is down to developer preference, but while I agree Werror is useful to have in CI, it can be frustrating to develop with. Would prefer if it was something that had to be explicitly enabled (or at least have an option to disable).

As a result, `-Werror` is included via option here:
https://github.com/openthread/openthread/blob/bf1cf97b0ffbf77d52bc8c9346e3280f5bbaf845/CMakeLists.txt#L41-L44

wgtdkp: The automake tool enables `-Werror` for both C and C++ by default, so I think there should be no exception for cmake? The cmake option seems add `-Werror` to only `OT_CFLAGS` and it gives warning but not error for C++ code when I build with `cmake -DOT_COMPILE_WARNING_AS_ERROR=ON ..`. 

@CiaranWoodward Why we always force warning as error for only C code but allows warning for only C++ code?
https://github.com/openthread/openthread/blob/ab9c0a40917be2223c8beebcd48a97296b644dab/CMakeLists.txt#L53

jwhui: `OT_CFLAGS` should also be included for C++, there seems to be a typo here:
https://github.com/openthread/openthread/blob/ab9c0a40917be2223c8beebcd48a97296b644dab/CMakeLists.txt#L54

The reasoning is as @CiaranWoodward mentioned in https://github.com/openthread/openthread/pull/4500#pullrequestreview-347974498, that having to deal with warnings during some quick development iterations can be frustrating. However, by enforcing `-Werror` in Travis, we ensure any code in the repo does not generate warnings.

wgtdkp: >However, by enforcing -Werror in Travis, we ensure any code in the repo does not generate warnings.

Yes, but we are delaying errors which can be corrected at compile-time to CI-time. It is even worse that openthread travis CI runs slow and we has to wait for travis builds.

@CiaranWoodward  What about add an option to disable `-Werror` for C++ but keep `-Werror` by default, just as automake tools do? Is that acceptable to you?

jwhui: > Yes, but we are delaying errors which can be corrected at compile-time to CI-time. It is even worse that openthread travis CI runs slow and we has to wait for travis builds.

You can (and should) enable Travis on your fork to run tests independent of the OpenThread org queue. See [CONTRIBUTING.md](https://github.com/openthread/openthread/blob/master/CONTRIBUTING.md)

Also, you can run the checks locally before submitting if you don’t want to use Travis at all.

I’d like to hear @CiaranWoodward’s thoughts on changing the default behavior.

CiaranWoodward: So ``Werror`` is useful to prevent warnings slipping in to the openthread codebase. This is important for:
1. Contributors to the openthread repo who want their code merged
2. The CI tool to verify
Any other users of the openthread repo will find it to be an annoyance, as anything from upgrading compiler to changing the target can generate warnings that aren't necessarily important. I know openthread extensively checks different compilers in CI, but it still isn't impossible. Also if a contributor wants to try something out in a quick-hack kind of way before dedicating the time to a proper implementation, then warnings can happen and just get in the way.

So really I think it is better to have an opt-in method of ``Werror`` like we have now, where the onus is on the CI tool to be configured correctly, and the contributors to use ccmake/cmake-gui/``-DOT_COMPILE_WARNING_AS_ERROR=ON`` to enable it.

abtink: > Any other users of the openthread repo will find it to be an annoyance, as anything from upgrading compiler to changing the target can generate warnings that aren't necessarily important.

I agree and I think this is a very good reason to keep the default `Werror` disabled. Particularly since cmake enables easier porting and use of new tool-chains. 

Also want to echo (and highly recommend) @jwhui comment about enabling and using Travis on a developer's fork. Travis covers so many OT configurations and tests  that when writing code would not be practical to check them on local machine (even small change may have side-effect that are not easily noticeable in another configurations). The other benefit is to help reduce the load on the main OT Travis queue.

wgtdkp: Okay, will close this PR and fix the typo in another PR.

PR #4696, opened 2020-03-17
[cmake] fix build with external mbedTLS

rlubos:
`mbedcrypto` target is only available when `OT_BUILTIN_MBEDTLS` is set,
therefore linking it unconditionally when external mbed TLS instance is
used results in a linking error.

Signed-off-by: Robert Lubos <robert.lubos@nordicsemi.no>

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -233,12 +233,14 @@ target_sources(openthread-radio PRIVATE
     utils/parse_cmdline.cpp
 )
 
-target_link_libraries(openthread-ftd
-    PUBLIC
-        mbedcrypto
-)
+if(OT_BUILTIN_MBEDTLS)
+    target_link_libraries(openthread-ftd
+        PUBLIC
+            mbedcrypto
+    )
 
-target_link_libraries(openthread-mtd
-    PUBLIC
-        mbedcrypto
-)
+    target_link_libraries(openthread-mtd
+        PUBLIC
+            mbedcrypto
+    )
+endif()


rlubos:
I've encountered the issue when I tried to update the OT version used in Zephyr.

bukepo:
Should external mbedtls also provides mbedcrypto? Removing this dependence results OpenThread no where to find mbedtls headers.

rlubos:
I'm open for discussion here, in Zephyr we took a different approach where we simply append the OpenThread inclusion list:
https://github.com/zephyrproject-rtos/openthread/blob/zephyr/zephyr/CMakeLists.txt#L72

Another option would be to create a dummy `mbedcrypto` target which would only specify include paths. mbed TLS in Zephyr is already in a different target, so re-defining it is not an option.

But IMO, if we allow providing an external mbed TLS, it should be up to the provided how it's done, no silent assumptions should be made.

bukepo:
> I'm open for discussion here, in Zephyr we took a different approach where we simply append the OpenThread inclusion list:
> https://github.com/zephyrproject-rtos/openthread/blob/zephyr/zephyr/CMakeLists.txt#L72
> 
> Another option would be to create a dummy `mbedcrypto` target which would only specify include paths. mbed TLS in Zephyr is already in a different target, so re-defining it is not an option.
> 
> But IMO, if we allow providing an external mbed TLS, it should be up to the provided how it's done, no silent assumptions should be made.

Thank you for sharing the facts. And I like the idea to create an interface target `mbedcrypto` in such case so that OpenThread's dependency is **complete**, and targets(inside OpenThread or outside of OpenThread) depends on OpenThread will not have to manipulate private variables(such as `OT_PRIVATE_INCLUDES`). I think this way is more aligned with modern cmake.

Issue #4689, opened 2020-03-16
[mac] inefficient use of radio in SED mode on nRF52840


e-rk:
**Describe the bug**
While conducting current measurements it was observed that during SED operation the radio on nRF52840 is inefficiently used, because before the data request, the OT stack explicitly requests the radio to enter RX state. This consumes extra power and is unnecessary on nRF52840 platform.

**To Reproduce**
Information to reproduce the behavior, including:
1. Git commit id: `8a1992e2e42fb707babe9491a6a1456e553490e8`, present on master as well
2. IEEE 802.15.4 hardware platform: nRF52840
3. Build steps: any configuration
4. Network topology: One parent and one SED

**Expected behavior**
The RX state is requested only if the platform requires it.

**Console/log output**
Current data request implementation
![current_unopt_label](https://user-images.githubusercontent.com/39553874/76778820-bfe86080-67aa-11ea-9979-ee70b758b738.png)

Expected behaviour:
![current_fix_label](https://user-images.githubusercontent.com/39553874/76778864-cd9de600-67aa-11ea-8232-1d2b5bc172b2.png)

**Additional context**
After commenting out the `Get<Radio>().Receive` call in `void SubMac::BeginTransmit(void)` the charge required to do the data request was lower by 30%, as seen on the second picture.

abtink:
`radio.h` platform defines the following state machine with regards to radio transitions and how OT core would/should use radio platform APIs:
https://github.com/openthread/openthread/blob/bf1cf97b0ffbf77d52bc8c9346e3280f5bbaf845/include/openthread/platform/radio.h#L236-L248
Per this, we would need to go to "receive" first before we can "send".   

The  [`otPlatRadioTransmit()`](https://github.com/openthread/openthread/blob/bf1cf97b0ffbf77d52bc8c9346e3280f5bbaf845/include/openthread/platform/radio.h#L554-L571) also mentions in its documentation that it can be called when radio is  "receive" state and should return `OT_ERROR_INVALID_STATE ` if called from other states.  I think some other radio platforms expect and require this pattern/behavior, so would not recommend changing this (in the core common Mac/SubMAC `BeginTransmit()`).

I am wondering why calling `otPlatRadioReceive()` on nRF52840 would cause this longer wait time before tx starts. Doesn't the `otRadioReceive()` return quickly (or does it take some action)?

e-rk:
[...]

PR #4697, opened 2020-03-17
[mac] add capability to transition from Sleep to Transmit during data request

e-rk:
Proposed fix to https://github.com/openthread/openthread/issues/4689


LuDuda:
Thank you @e-rk 👍

PR #4700, opened 2020-03-17
[coap] use Coap::Message helper methods checking CoAP type

abtink:
This commit changes code to use `IsConfirmable()`,
`IsNonConfirmable()`, `IsAck()` helper methods.

### DIFF
--- a/src/core/meshcop/border_agent.cpp
+++ b/src/core/meshcop/border_agent.cpp
@@ -193,7 +193,7 @@ static void SendErrorMessage(Coap::CoapSecure &   aCoapSecure,
 
     VerifyOrExit((message = NewMeshCoPMessage(aCoapSecure)) != NULL, error = OT_ERROR_NO_BUFS);
 
-    if (aRequest.GetType() == OT_COAP_TYPE_NON_CONFIRMABLE || aSeparate)
+    if (aRequest.IsNonConfirmable() || aSeparate)
     {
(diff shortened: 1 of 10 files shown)


jwhui:
What about:
https://github.com/openthread/openthread/blob/13fa5adc5e78fff41e56f393158b9694999d948b/src/core/meshcop/joiner.cpp#L523

src/core/meshcop/border_agent.cpp
```diff
@@ -453,8 +453,7 @@ void BorderAgent::HandleRelayReceive(const Coap::Message &aMessage)
     Coap::Message *message = NULL;
     otError        error;
 
-    VerifyOrExit(aMessage.GetType() == OT_COAP_TYPE_NON_CONFIRMABLE && aMessage.GetCode() == OT_COAP_CODE_POST,
-                 error = OT_ERROR_DROP);
+    VerifyOrExit(aMessage.IsNonConfirmable() && aMessage.GetCode() == OT_COAP_CODE_POST, error = OT_ERROR_DROP);
```
simonlingoogle: nit: Add `IsPost()` too?

abtink: This PR is using the existing helper methods in `Coap::Message` that check the CoAP "type". The `POST` is a CoAP "code" (not a "type"), and we currently don't have helper methods like `IsPost()`. 

There are only 4 CoAP types and we already did have helper methods defined for all 4 types, but there are many more CoAP codes so I am not sure if we want helper methods for checking the code values.

PR #4701, opened 2020-03-17
[coap-message] add Coap::MessageQueue class

abtink:
This commit adds `Coap::MessageQueue` class as a subclass of generic
`ot::MessageQueue` as a queue of `Coap::Message` only. This type is
used in `coap.cpp` helping simplify the code (by removing the type
casting).

### DIFF
--- a/src/core/coap/coap.cpp
+++ b/src/core/coap/coap.cpp
@@ -76,11 +76,11 @@ void CoapBase::ClearRequests(const Ip6::Address *aAddress)
 {
     Message *nextMessage;
 
-    for (Message *message = static_cast<Message *>(mPendingRequests.GetHead()); message != NULL; message = nextMessage)
+    for (Message *message = mPendingRequests.GetHead(); message != NULL; message = nextMessage)
     {
         Metadata metadata;
 
-        nextMessage = static_cast<Message *>(message->GetNext());
+        nextMessage = message->GetNextCoapMessage();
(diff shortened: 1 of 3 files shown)


src/core/coap/coap.cpp
```diff
@@ -76,11 +76,11 @@ void CoapBase::ClearRequests(const Ip6::Address *aAddress)
 {
     Message *nextMessage;
 
-    for (Message *message = static_cast<Message *>(mPendingRequests.GetHead()); message != NULL; message = nextMessage)
+    for (Message *message = mPendingRequests.GetHead(); message != NULL; message = nextMessage)
     {
         Metadata metadata;
 
-        nextMessage = static_cast<Message *>(message->GetNext());
+        nextMessage = MessageQueue::GetNext(*message);
```
abtink: I went for a static method of `Coap::MessageQueue` to get the next entry from a previous method instead of overloading the `GetNext()` of `Coap::Message` (since a CoAP message can still be added in an `ot::MessageQueue` along with other `ot::Message` instances). 

Thoughts? Suggestions?

simonlingoogle: Add method `Coap::GetNextCoap` that returns `Coap::Message` ?

abtink: Added `Coap::Message::GetNextCoapMessage()`.

PR #4702, opened 2020-03-18
[nrf528xx] fix IAR undefined order of volatile accesses error

piotrkoziar:
After #4652 IAR compiles with warning: 
warning[Pa082]: undefined behavior: the order of volatile accesses is undefined in this statement


examples/platforms/nrf528xx/src/transport/uart.c
```diff
@@ -97,21 +98,23 @@ static void processReceive(void)
 {
     // Set head position to not be changed during read procedure.
     uint16_t head = sReceiveHead;
-
+    uint8_t *position;
```
LuDuda: nit: please add a new line below.

PR #4703, opened 2020-03-18
[tlv] add IsBoundedBy() helper method

abtink:
The new `Tlv::IsBoundedBy()` method checks whether the entire TLV is
bounded by (i.e., `<`) a given end pointer. The implementation works
correctly independent of TLV being an Extended TLV or not. It also
ensures not to read beyond the given end pointer by first checking the
Type and Length fields are bounded by the end address before reading
the Length value.

### DIFF
--- a/src/core/common/tlvs.hpp
+++ b/src/core/common/tlvs.hpp
@@ -164,6 +164,19 @@ public:
     }
 
     /**
+     * This method indicates whether the TLV and its Value are bounded by (i.e., `<`) a given end pointer.
+     *
+     * This method can be used independent of whether TLV is an Extended TLV or not.
+     *
+     * @param[in] aEndPointer  An end pointer to check against.
+     *
+     * @retval TRUE   The TLV is bounded by the @p aEndPointer (i.e., the end of TLV < @p aEndPointer)
+     * @retval FALSE  The TLV is not bounded by the @p aEndPointer (i.e., the end of TLV >= @p aEndPointer).
+     *
+     */
+    bool IsBoundedBy(const void *aEndPointer) const;
+
+    /**
      * This method appends a TLV to the end of the message.
      *
      * On success, this method grows the message by the size of the TLV.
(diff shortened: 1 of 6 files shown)


src/core/common/tlvs.hpp
```diff
@@ -163,6 +163,19 @@ class Tlv
         return reinterpret_cast<const Tlv *>(reinterpret_cast<const uint8_t *>(this) + GetSize());
     }
 
+    /**
+     * This method indicates whether the TLV and its Value are bounded by (i.e., `<=`) a given end pointer.
```
jwhui:
```suggestion
     * This method indicates whether the TLV and its Value are bounded by (i.e., `<`) a given end pointer.
```

I think we generally want the last byte of the TLV to be before the end pointer, right?

src/core/common/tlvs.hpp
```diff
@@ -163,6 +163,19 @@ class Tlv
...
+     *
+     * This method can be used independent of whether TLV is an Extended TLV or not.
+     *
+     * @param[in] aEndPointer  An end pointer to check against.
+     *
+     * @retval TRUE   The TLV is bounded by the @p aEndPointer (i.e., the end of TLV <= @p aEndPointer)
+     * @retval FLASE  The TLV is not bounded by the @p aEndPointer (i.e., the end of TLV > @p aEndPointer).
```
jwhui:
```suggestion
     * @retval FALSE  The TLV is not bounded by the @p aEndPointer (i.e., the end of TLV >= @p aEndPointer).
```
See comment above.

src/core/common/tlvs.hpp
abtink: Yes. Indeed. Thanks.
The implementing is correct in using `<=` (since we go to the next TLV which is after end of current one) but considering current TLV itself we should expect all its bytes are strictly before  (`<`) the given end pointer.

abtink:
Closing this in favor of https://github.com/openthread/openthread/pull/4743

PR #4708, opened 2020-03-19
[script] let alias to take effect

librasungirl:


script/clang-format
```diff
@@ -34,6 +34,10 @@ die() {
     exit 1
 }
 
+# Aliases are not expanded when the shell is not interactive, unless the
+# expand_aliases shell option is set using shopt.
+shopt -s expand_aliases
```
bukepo: If this is necessary, should we add this to all scripts?

PR #4714, opened 2020-03-20
[cc1352,cc26xx] include `-pedantic-errors` in platform code

jwhui:
- Also fix a pointer arithmetic issue that was flagged.


FiveDimensions:
verified that no more compilation errors with -pedantic-errors enabled

Issue #4710, opened 2020-03-19
[flash] improve robustness to power failure


jwhui:
**Describe the bug**

As reported by @dismirlian in https://github.com/openthread/openthread/pull/4552#issuecomment-592943770

> If I understand the code correctly, I think there may some small problems regarding power-failure robustness:
> 
> 1. We always assume that the record with `kFlagAddBegin` set will also have its `mLength` correctly set. If the firmware fails to write the whole header (for instance, if it resets after writing `mFlags`), then the record is left in an inconsistent state. For example, `mFlags` with `kFlagAddBegin` but `mLength` erased (=65535). This would have negative side effects, because it impacts all the methods that iterate over the records, specifically, the `Init` method only checks for `record.IsAddBeginSet()` and then reads the length:
>    https://github.com/openthread/openthread/blob/381f4deda9a6a3d986261046fc2618743ac2422d/src/core/utils/flash.cpp#L69-L73
>    
>    After `Init` initializes `mSwapUsed` incorrectly, all other methods would iterate over garbage.
>    The same happens in `Swap`.
> 2. We always assume that the space after the last record is "writeable" (that is to say, it is erased). This may not be the case if the `mKey` was written but the `mFlags` wasn't.
> 
> A possible solution to both issues would be to enhance the `Init` method to "sanitize" the settings area:
> 
> * If a record is found whose `mFlags` has `kFlagAddBegin` but not `kFlagComplete`, we can assume an interrupted operation, should not count that record in `mSwapUsed`, and trigger a `Swap`. This should get rid of the invalid record.
> * After the last (valid) record, `Init` should scan the remaining space in the swap block and trigger a `Swap` if any byte is != 0xff. This will ensure that the remaining space is "writable".

jwhui:
@dismirlian, would you like to submit a PR to address this issue?

dismirlian:
Hi @jwhui, yes, no problem. I'm currently preparing a PR to fix a problem with the new efr32 flash platform driver.

I can submit another one addressing this issue afterwards.

jwhui:
> Hi @jwhui, yes, no problem. I'm currently preparing a PR to fix a problem with the new efr32 flash platform driver.
> 
> I can submit another one addressing this issue afterwards.

@dismirlian, great, thanks!

PR #4717, opened 2020-03-20
[flash] improve power failure robustness

dismirlian:
This PR improves the flash implementation robustness to power failure. The idea is the following:

The `Init` method now checks if a record has been partially written by seeing if the `kFlagAddBegin` flag is set, but the `kFlagComplete` flag is not. In this case, `Init` stops incrementing `mSwapUsed`. The new `SanitizeFreeSpace` method checks the (supposedly) free space, to see if it is writeable. If it isn't, it triggers a `Swap`, which sanitizes the flash area.

Please see #4710 and https://github.com/openthread/openthread/pull/4552#issuecomment-592943770.

There is an additional subtle detail: the `Delete` method, when called with `aIndex = 0`, erases the records without first marking the `index = 1` record with `kFlagFirst`. If the operation is interrupted at this point, no record will have the `kFlagFirst` flag set. At first, this struck me as a possible issue, but actually it's not: all the methods that iterate over the settings area initialize `index = 0` without expecting a record with the `kFlagFirst` flag set. Maybe we could add a comment, so that this doesn't get lost during a future code refactoring, or maybe it's obvious for the casual reader (it wasn't obvious to me...).

### DIFF
--- a/src/core/utils/flash.cpp
+++ b/src/core/utils/flash.cpp
@@ -183,6 +183,8 @@ otError Flash::Add(uint16_t aKey, bool aFirst, const uint8_t *aValue, uint16_t a
     record.Init(aKey, aFirst);
     record.SetData(aValue, aValueLength);
 
+    OT_ASSERT((mSwapSize - record.GetSize()) >= kSwapMarkerSize);
+
     if ((mSwapSize - record.GetSize()) < mSwapUsed)
     {
         Swap();
--- a/src/core/utils/flash.cpp
+++ b/src/core/utils/flash.cpp
@@ -69,11 +69,47 @@ void Flash::Init(void)
     for (mSwapUsed = kSwapMarkerSize; mSwapUsed <= mSwapSize - sizeof(record); mSwapUsed += record.GetSize())
     {
         otPlatFlashRead(&GetInstance(), mSwapIndex, mSwapUsed, &record, sizeof(record));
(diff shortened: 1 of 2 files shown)


jwhui:
Thanks for this contribution!

> There is an additional subtle detail: the `Delete` method, when called with `aIndex = 0`, erases the records without first marking the `index = 1` record with `kFlagFirst`. If the operation is interrupted at this point, no record will have the `kFlagFirst` flag set. At first, this struck me as a possible issue, but actually it's not: all the methods that iterate over the settings area initialize `index = 0` without expecting a record with the `kFlagFirst` flag set. Maybe we could add a comment, so that this doesn't get lost during a future code refactoring, or maybe it's obvious for the casual reader (it wasn't obvious to me...).

I agree this is subtle. Would you mind adding the comment to this PR?

dismirlian:
> I agree this is subtle. Would you mind adding the comment to this PR?

No problem, done!

jwhui:
> > I agree this is subtle. Would you mind adding the comment to this PR?
> 
> No problem, done!

Thanks!

jwhui:
@dismirlian, thank you for this submission!

Can you run `clang-format` to resolve format checks? Thanks.

dismirlian:
@jwhui sorry for the delay. It seems I forgot to format my latest commit. Thanks!

PR #4720, opened 2020-03-21
[network-data] use explicit signed int for bit field

jwhui:


include/openthread/netdata.h
```diff
@@ -65,7 +65,7 @@ typedef struct otBorderRouterConfig
     /**
      * A 2-bit signed integer indicating router preference as defined in RFC 4191.
      */
-    int mPreference : 2;
+    signed int mPreference : 2;
```
bukepo: Is this just highlight this is a signed integer for readability?

jwhui: https://en.cppreference.com/w/cpp/language/bit_field

> The following properties of bit fields are _implementation-defined_
> 
> * Whether char, short, int, long, and long long bit fields that aren't explicitly signed or unsigned are signed or unsigned.
> 
>   * For example, int b:3; may have the range of values 0..7 or -4..3.
>

PR #4725, opened 2020-03-23
[lowpan] avoid backward goto statement in Compress()

jwhui:


jwhui:
> What's the reason for this change? We have one other place where we use `goto` to a start in `MeshForwarder`.

This is mainly for readability. `Lowpan::Compress` has both forward and backward goto statements, which makes it a bit less readable than in `MeshForwarder::PrepareDataFrame`. I'll admit I have not found a way to remove goto without any code overhead. As a result, I'm OK with simply closing this PR if we think the added benefit does not outweigh the cost.

abtink:
I would favor readability and would recommend keeping it.

PR #4730, opened 2020-03-24
[mle] avoid unnecessary method calls on MTD

jwhui:


src/core/thread/router_table.hpp
```diff
@@ -29,6 +29,8 @@
 #ifndef ROUTER_TABLE_HPP_
 #define ROUTER_TABLE_HPP_
 
+#if OPENTHREAD_FTD
```
abtink: Not related to this PR, but reading this notcied that this file seems to be missing `#include "openthread-core-config.h"` at the top.
I think we should include it before the `#if OPENTHREAD_FTD` check?

src/core/thread/src_match_controller.cpp
```diff
@@ -31,6 +31,8 @@
  *   This file implements source address match controller.
  */
 
+#if OPENTHREAD_FTD
```
abtink: Should we move this after the `#include "src_match_controller.hpp"` to give chance for config header to be included?

src/core/thread/child_table.hpp
```diff
@@ -34,15 +34,15 @@
 #ifndef CHILD_TABLE_HPP_
 #define CHILD_TABLE_HPP_
 
+#if OPENTHREAD_FTD
```
abtink: Same here? Moving this after the `#include "openthread-core-config.h"`?

src/core/thread/mle.cpp
```diff
@@ -2713,50 +2735,32 @@ void Mle::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessageIn
...
         break;
 
+    case Header::kCommandAnnounce:
+        HandleAnnounce(aMessage, aMessageInfo);
+        break;
+
     case Header::kCommandChildUpdateRequest:
         if (mRole == OT_DEVICE_ROLE_LEADER || mRole == OT_DEVICE_ROLE_ROUTER)
         {
+#if OPENTHREAD_FTD
```
librasungirl: nit: should be no chance for non-FTD device to act as Leader or Router role, how about wrapping `if... else` with the `OPENTHREAD_FTD` macro?

src/core/thread/mle.cpp
```diff
@@ -2768,7 +2772,9 @@ void Mle::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessageIn
     case Header::kCommandChildUpdateResponse:
         if (mRole == OT_DEVICE_ROLE_LEADER || mRole == OT_DEVICE_ROLE_ROUTER)
         {
+#if OPENTHREAD_FTD
```
librasungirl: ditto

PR #4731, opened 2020-03-25
[travis] disable build checks that exist in GitHub Actions

jwhui:
If/when we are comfortable with GitHub Actions, we can start disabling redundant Travis tests.


simonlingoogle:
Looking forward to delete `.travis.yml` :+1:

PR #4749, opened 2020-03-28
[address-resolver] remove invalid caches earlier when applicable

librasungirl:


src/core/thread/router_table.cpp
```diff
@@ -564,6 +564,7 @@ void RouterTable::ProcessTlv(const ThreadRouterMaskTlv &aTlv)
         else
         {
             mAllocatedRouterIds.Remove(routerId);
+            Get<AddressResolver>().Remove(routerId);
```
simonlingoogle: Can we even remove the neighbor earlier?

librasungirl: Existing implementation has already updated router table accordingly, see https://github.com/openthread/openthread/blob/master/src/core/thread/router_table.cpp#L539

simonlingoogle: Why is the logic here not exactly same as in function `void RouterTable::ProcessTlv(const Mle::RouteTlv &aTlv)` ?

src/core/thread/router_table.cpp
```diff
@@ -531,6 +531,7 @@ void RouterTable::ProcessTlv(const Mle::RouteTlv &aTlv)
             RemoveNeighbor(*router);
 
             mAllocatedRouterIds.Remove(routerId);
+            Get<AddressResolver>().Remove(routerId);
```
simonlingoogle: `Get<AddressResolver>().Remove(routerId);` is already called by `RemoveNeighbor(*router);` ?

simonlingoogle: The current code of handling `RouterTLV` and `ThreadRouterMaskTlv` are not exactly same. Should we just use one same code for processing both TLVs ?

librasungirl: Agree that it is not necessary here and there are redundancies in router id set relative process. will close this one submit a separate one for enhancements.

PR #4753, opened 2020-03-29
[mesh-forwarder] do not remove router on link failures

jwhui:
A router coming out of reset sends a Link Request to re-establish its
router role and links to neighboring routers. The existing
implementation only responds to such Link Requests when a valid
neighbor entry is found. This check is too strict and can cause
partitions to remain for a couple minutes. The Thread Specification
only requires a router to check that the Router ID is currently
allocated.

This commit updates the implementation to be conformant with the
Thread Specification and allows routers coming out of reset to
resynchronize even when a neighboring router has detect a link
failure.

### DIFF
--- a/src/core/thread/mesh_forwarder.cpp
+++ b/src/core/thread/mesh_forwarder.cpp
@@ -810,7 +810,7 @@ Neighbor *MeshForwarder::UpdateNeighborOnSentFrame(Mac::TxFrame &aFrame, otError
 
         if (neighbor->GetLinkFailures() >= Mle::kFailedRouterTransmissions)
         {
-            Get<Mle::MleRouter>().RemoveNeighbor(*neighbor);
+            Get<Mle::MleRouter>().RemoveRouterLink(*static_cast<Router *>(neighbor));
         }
     }
 
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -3267,6 +3267,29 @@ bool MleRouter::IsMinimalChild(uint16_t aRloc16)
     return rval;
 }
 
+void MleRouter::RemoveRouterLink(Router &aRouter)
+{
+    switch (mRole)
+    {
+    case OT_DEVICE_ROLE_CHILD:
+        if (&aRouter == &mParent)
+        {
+            BecomeDetached();
+        }
+        break;
+
+#if OPENTHREAD_FTD
+    case OT_DEVICE_ROLE_ROUTER:
+    case OT_DEVICE_ROLE_LEADER:
+        mRouterTable.RemoveNeighbor(aRouter);
+        break;
+#endif
+
+    default:
+        break;
+    }
+}
+
 void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
 {
     if (&aNeighbor == &mParent)
--- a/src/core/thread/mle_router.hpp
+++ b/src/core/thread/mle_router.hpp
@@ -328,6 +328,14 @@ public:
     void RemoveNeighbor(Neighbor &aNeighbor);
 
     /**
+     * This method invalidates a direct link to a neighboring router (due to failed link-layer acks).
+     *
+     * @param[in]  aRouter  A reference to the router object.
+     *
+     */
+    void RemoveRouterLink(Router &aRouter);
+
+    /**
      * This method restores children information from non-volatile memory.
      *
      */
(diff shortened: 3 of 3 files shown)


src/core/thread/mesh_forwarder.cpp
```diff
@@ -810,7 +810,16 @@ Neighbor *MeshForwarder::UpdateNeighborOnSentFrame(Mac::TxFrame &aFrame, otError
 
         if (neighbor->GetLinkFailures() >= Mle::kFailedRouterTransmissions)
         {
-            Get<Mle::MleRouter>().RemoveNeighbor(*neighbor);
+#if OPENTHREAD_FTD
+            if (Mle::Mle::IsActiveRouter(neighbor->GetRloc16()))
```
simonlingoogle: always `true` here since `VerifyOrExit(Mle::Mle::IsActiveRouter(neighbor->GetRloc16()));` above?

src/core/thread/mesh_forwarder.cpp
```diff
@@ -806,12 +806,15 @@ Neighbor *MeshForwarder::UpdateNeighborOnSentFrame(Mac::TxFrame &aFrame, otError
...
     {
         neighbor->IncrementLinkFailures();
+
+#if OPENTHREAD_FTD
         VerifyOrExit(Mle::Mle::IsActiveRouter(neighbor->GetRloc16()));
 
         if (neighbor->GetLinkFailures() >= Mle::kFailedRouterTransmissions)
         {
-            Get<Mle::MleRouter>().RemoveNeighbor(*neighbor);
+            Get<RouterTable>().RemoveNeighbor(static_cast<Router &>(*neighbor));
```
abtink: Should we do? `Signal(OT_NEIGHBOR_TABLE_EVENT_ROUTER_REMOVED, aNeighbor);`?

---

(unrelated to this: but at some point, I'd think it'd be good to move such notification events to the router table itself).

jwhui: Note that this change simply removes the direct neighbor link (by setting the link quality to 0). Do we want to notify when the link quality transitions to 0 but the router entry has not yet been removed?

src/core/thread/mesh_forwarder.cpp
```diff
@@ -806,12 +806,15 @@ Neighbor *MeshForwarder::UpdateNeighborOnSentFrame(Mac::TxFrame &aFrame, otError
     else if (aError == OT_ERROR_NO_ACK)
     {
         neighbor->IncrementLinkFailures();
+
+#if OPENTHREAD_FTD
```
abtink: Let me double-check if we do not need this on MTD?
For example I think a SED child not getting acks from its parent would go through this code path... and then would remove its parent (which would cause it to detach)?

jwhui: Yes, you are correct. I'll change it to be based on role.

src/core/thread/mesh_forwarder.cpp
abtink: I see. The `Neighbor` state remains as before, so it will still show up in the "neighbor table". 
Where/how would we remove this neighbor (until the router id entry is removed?)

jwhui: Router ID entry is removed or timeout occurs due to not receiving MLE Advertisements.

src/core/thread/mesh_forwarder.cpp
```diff
@@ -810,7 +810,7 @@ Neighbor *MeshForwarder::UpdateNeighborOnSentFrame(Mac::TxFrame &aFrame, otError
 
         if (neighbor->GetLinkFailures() >= Mle::kFailedRouterTransmissions)
         {
-            Get<Mle::MleRouter>().RemoveNeighbor(*neighbor);
+            Get<Mle::MleRouter>().RemoveRouterLink(*static_cast<Router *>(neighbor));
```
abtink: Nice. Thanks for renaming it. 👍

src/core/thread/mesh_forwarder.cpp
abtink: Sounds good to me. 👍 I double-checked that in the case of `OPENTHREAD_CONFIG_MLE_SEND_LINK_REQUEST_ON_ADV_TIMEOUT` we don't expect for link failures to remove the neighbor and use the timeout.

PR #4756, opened 2020-03-30
[core] use the common `RouterIdSet` class

librasungirl:
Currently there are four similar copies regarding Router Id set separately in 1) `RouteTlv` in mle_tlvs.hpp; 2) `ThreadRouterMaskTlv` in thread_tlvs.hpp; 3) `RouterIdSet` in router_table.hpp; 4)`RouteTlv` in network_diagnostic_tlvs.hpp.

This PR extracts the common `RouterIdSet` class and apply it in the four components, meanwhile keep one copy for `RouteTlv` and `ThreadRouterMaskTlv` process in route_table (see [comments in #4749](https://github.com/openthread/openthread/pull/4749#discussion_r399736521))

### DIFF
--- a/src/core/thread/mle_types.hpp
+++ b/src/core/thread/mle_types.hpp
@@ -614,10 +614,21 @@ public:
      * @retval FALSE  If the Router ID sets are not equal.
      *
      */
-    bool operator!=(const RouterIdSet &aOther) const
+    bool operator==(const RouterIdSet &aOther) const
     {
-        return memcmp(mRouterIdSet, aOther.mRouterIdSet, sizeof(mRouterIdSet)) != 0;
+        return memcmp(mRouterIdSet, aOther.mRouterIdSet, sizeof(mRouterIdSet)) == 0;
     }
+
+    /**
+     * This method returns whether or not the Router ID sets are not equal.
+     *
+     * @param[in]  aOther The other Router ID Set to compare with.
+     *
+     * @retval TRUE   If the Router ID sets are not equal.
+     * @retval FALSE  If the Router ID sets are equal.
+     *
+     */
+    bool operator!=(const RouterIdSet &aOther) const { return !(*this == aOther); }
 
 private:
     uint8_t mRouterIdSet[BitVectorBytes(Mle::kMaxRouterId + 1)];
--- a/src/core/thread/mle_types.hpp
+++ b/src/core/thread/mle_types.hpp
@@ -42,6 +42,7 @@
 
 #include <openthread/thread.h>
 
+#include "common/encoding.hpp"
 #include "common/string.hpp"
 #include "mac/mac_types.hpp"
 
@@ -566,6 +567,61 @@ public:
      */
     void SetLeaderRouterId(uint8_t aRouterId) { mLeaderRouterId = aRouterId; }
 };
+
+OT_TOOL_PACKED_BEGIN
+class RouterIdSet
+{
+public:
+    /**
+     * This method clears the Router Id Set.
+     *
+     */
+    void Clear(void) { memset(mRouterIdSet, 0, sizeof(mRouterIdSet)); }
+
+    /**
+     * This method indicates whether or not a Router ID bit is set.
+     *
+     * @param[in]  aRouterId  The Router ID.
+     *
+     * @retval TRUE   If the Router ID bit is set.
+     * @retval FALSE  If the Router ID bit is not set.
+     *
+     */
+    bool Contains(uint8_t aRouterId) const { return (mRouterIdSet[aRouterId / 8] & (0x80 >> (aRouterId % 8))) != 0; }
+
+    /**
+     * This method sets a given Router ID.
+     *
+     * @param[in]  aRouterId  The Router ID to set.
+     *
+     */
+    void Add(uint8_t aRouterId) { mRouterIdSet[aRouterId / 8] |= 0x80 >> (aRouterId % 8); }
+
+    /**
+     * This method removes a given Router ID.
+     *
+     * @param[in]  aRouterId  The Router ID to remove.
+     *
+     */
+    void Remove(uint8_t aRouterId) { mRouterIdSet[aRouterId / 8] &= ~(0x80 >> (aRouterId % 8)); }
+
+    /**
+     * This method returns whether or not the Router ID sets are equal.
+     *
+     * @param[in]  aOther The other Router ID Set to compare with.
+     *
+     * @retval TRUE   If the Router ID sets are equal.
+     * @retval FALSE  If the Router ID sets are not equal.
+     *
+     */
+    bool operator!=(const RouterIdSet &aOther) const
+    {
+        return memcmp(mRouterIdSet, aOther.mRouterIdSet, sizeof(mRouterIdSet)) != 0;
+    }
+
+private:
+    uint8_t mRouterIdSet[BitVectorBytes(Mle::kMaxRouterId + 1)];
+} OT_TOOL_PACKED_END;
 
 /**
  * @}
--- a/src/core/thread/router_table.cpp
+++ b/src/core/thread/router_table.cpp
@@ -306,7 +306,7 @@ exit:
     return error;
 }
 
-void RouterTable::RemoveNeighbor(Router &aRouter)
+void RouterTable::RemoveRouterLink(Router &aRouter)
 {
     aRouter.SetLinkQualityOut(0);
     aRouter.SetLastHeard(TimerMilli::GetNow());
@@ -502,23 +502,25 @@ exit:
     return rval;
 }
 
-void RouterTable::ProcessTlv(const Mle::RouteTlv &aTlv)
+void RouterTable::ProcessRouterIdSet(uint8_t aRouterIdSequence, const Mle::RouterIdSet &aRouterIdSet)
 {
     bool allocationChanged = false;
 
-    mRouterIdSequence            = aTlv.GetRouterIdSequence();
+    mRouterIdSequence            = aRouterIdSequence;
     mRouterIdSequenceLastUpdated = TimerMilli::GetNow();
+
+    VerifyOrExit(mAllocatedRouterIds != aRouterIdSet);
 
     for (uint8_t routerId = 0; routerId <= Mle::kMaxRouterId; routerId++)
     {
-        if (aTlv.IsRouterIdSet(routerId) == IsAllocated(routerId))
+        if (aRouterIdSet.Contains(routerId) == IsAllocated(routerId))
         {
             continue;
         }
 
         allocationChanged = true;
 
-        if (aTlv.IsRouterIdSet(routerId))
+        if (aRouterIdSet.Contains(routerId))
         {
             mAllocatedRouterIds.Add(routerId);
         }
@@ -528,7 +530,7 @@ void RouterTable::ProcessTlv(const Mle::RouteTlv &aTlv)
 
             OT_ASSERT(router != NULL);
             router->SetNextHop(Mle::kInvalidRouterId);
-            RemoveNeighbor(*router);
+            RemoveRouterLink(*router);
 
             mAllocatedRouterIds.Remove(routerId);
         }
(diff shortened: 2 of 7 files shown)


src/core/thread/mle_types.hpp
```diff
@@ -567,6 +568,61 @@ class LeaderData : public otLeaderData
...
+    /**
+     * This method removes the Router ID.
+     *
+     * @param[in]  aRouterId  The Router ID to remove.
+     *
+     */
+    void Remove(uint8_t aRouterId) { mRouterIdSet[aRouterId / 8] &= ~(0x80 >> (aRouterId % 8)); }
+
+    /**
+     * This method returns whether or not two Router ID sets equal.
```
jwhui:
```suggestion
     * This method returns whether or not two Router ID sets are equal.
```

src/core/thread/mle_types.hpp
```diff
@@ -567,6 +568,61 @@ class LeaderData : public otLeaderData
...
+     *
+     * @param[in]  aOther The other Router ID Set to compare with.
+     *
+     * @retval TRUE   If two Router ID sets equal.
```
jwhui:
```suggestion
     * @retval TRUE   If the two Router ID sets are equal.
```

src/core/thread/mle_types.hpp
```diff
@@ -567,6 +568,61 @@ class LeaderData : public otLeaderData
...
+     * @retval FALSE  If two Router ID sets don't equal.
```
jwhui:
```suggestion
     * @retval FALSE  If the Router ID sets are not equal.
```

src/core/thread/mle_types.hpp
```diff
@@ -567,6 +568,61 @@ class LeaderData : public otLeaderData
...
+     * @param[in]  aRouterId  The Router ID.
+     *
+     * @retval TRUE   If the Router ID bit is set.
+     * @retval FALSE  If the Router ID bit is not set.
+     *
+     */
+    bool Contains(uint8_t aRouterId) const { return (mRouterIdSet[aRouterId / 8] & (0x80 >> (aRouterId % 8))) != 0; }
+
+    /**
+     * This method sets the Router ID.
```
jwhui:
```suggestion
     * This method sets a given Router ID.
```

src/core/thread/mle_types.hpp
```diff
@@ -567,6 +568,61 @@ class LeaderData : public otLeaderData
...
+     *
+     * @param[in]  aRouterId  The Router ID to set.
+     *
+     */
+    void Add(uint8_t aRouterId) { mRouterIdSet[aRouterId / 8] |= 0x80 >> (aRouterId % 8); }
+
+    /**
+     * This method removes the Router ID.
```
jwhui:
```suggestion
     * This method removes a given Router ID.
```

src/core/thread/mle_types.hpp
```diff
@@ -567,6 +568,61 @@ class LeaderData : public otLeaderData
...
+     *
+     * @param[in]  aOther The other Router ID Set to compare with.
+     *
+     * @retval TRUE   If two Router ID sets equal.
+     * @retval FALSE  If two Router ID sets don't equal.
+     *
+     */
+    bool operator!=(const RouterIdSet &aOther) const
+    {
+        return memcmp(mRouterIdSet, aOther.mRouterIdSet, OT_ARRAY_LENGTH(mRouterIdSet)) != 0;
```
jwhui:
```suggestion
        return memcmp(mRouterIdSet, aOther.mRouterIdSet, sizeof(mRouterIdSet)) != 0;
```
I think better to use `sizeof()` since we care more about the size rather than the number of elements.

src/core/thread/mle_types.hpp
```diff
@@ -567,6 +568,61 @@ class LeaderData : public otLeaderData
...
 
+OT_TOOL_PACKED_BEGIN
+class RouterIdSet
+{
+public:
+    /**
+     * This method clears the Router Id Set.
+     *
+     */
+    void Clear(void) { memset(mRouterIdSet, 0, OT_ARRAY_LENGTH(mRouterIdSet)); }
```
jwhui:
```suggestion
    void Clear(void) { memset(mRouterIdSet, 0, sizeof(mRouterIdSet)); }
```
I think better to use `sizeof()` since we care more about the size rather than the number of elements.

src/core/thread/router_table.cpp
```diff
@@ -306,7 +306,7 @@ otError RouterTable::Release(uint8_t aRouterId)
     return error;
 }
 
-void RouterTable::RemoveNeighbor(Router &aRouter)
```
simonlingoogle: Is this change related to this PR?

librasungirl: It might be not so related to the intention of this PR, however `RemoveNeighbor()` name does not exactly reflect what it does when updating `ProcessTlv()` to `ProcessRouterIdSet()`.
It is just a small piece of code, prefer to keep it as is.

simonlingoogle: Agree that `RouterTable::RemoveNeighbor` does not reflect what it actually do. But, neither do I think `RemoveRouter` reflects exactly. 
`RemoveRouter` is basically same as `RemoveNeighbor`, because every `Neighbor` in `RouterTable` is `Router`. However, what this function actually removes is the **ROUTES** that use the router as the next hop. 

So, maybe this function should be renamed to `RemoveRoutes` or `RemoveRoutesFrom`? Thoughts? @jwhui @bukepo

jwhui: FWIW, `RemoveNeighbor()` was closer than `RemoveRouter()`. This method removes direct link to the neighboring device (effectively removing the neighbor). Of course, as a result, this also removes any routes that utilize the direct neighbor link.

librasungirl: The removed entry is what was in the router table but now not when learning the latest `RouterIdSet` either from `RouteTlv` or `ThreadRouterMaskTlv`, it may be the device's direct neighbor, but may also be some other routers which were routable.

As @jwhui , when removing one router entry, the other entries which was routed via this entry should be updated along with.

Thoughts?

jwhui: This method is also in the call chain initiated by `MeshForwarder()` to invalidate a link to a neighboring device.
https://github.com/openthread/openthread/blob/d85c0007a13b05ceac5f2b070417d9f191bc5cdc/src/core/thread/mesh_forwarder.cpp#L813

If we want to rename, we could adopt the same name `RemoveRouterLink()`.

librasungirl: Thought the neighboring router also belongs router, not conflicting with `RemoveRouter()` name.
Not sure but the `Link` in `RemoveRouterLink()` gives me the impression is it is removing direct neighbor, is it wrong?

jwhui: This method only sets the link quality to zero and invalidates any routes that utilize the link. It does not actually remove any other neighbor or router state. `RemoveRouter()` might give the impression that the frame counters and other state are being removed. Thoughts?

librasungirl: ok. that makes sense. will update.

src/core/thread/mle_types.hpp
```diff
@@ -567,6 +568,61 @@ class LeaderData : public otLeaderData
     void SetLeaderRouterId(uint8_t aRouterId) { mLeaderRouterId = aRouterId; }
 };
 
+OT_TOOL_PACKED_BEGIN
+class RouterIdSet
```
jwhui: Would a more natural location for this be in `router_table.hpp`?

src/core/thread/mle_types.hpp
```diff
@@ -567,6 +568,61 @@ class LeaderData : public otLeaderData
...
+     * @param[in]  aOther The other Router ID Set to compare with.
+     *
+     * @retval TRUE   If the Router ID sets are equal.
+     * @retval FALSE  If the Router ID sets are not equal.
+     *
+     */
+    bool operator!=(const RouterIdSet &aOther) const
+    {
+        return memcmp(mRouterIdSet, aOther.mRouterIdSet, sizeof(mRouterIdSet)) != 0;
+    }
```
abtink: How about if we also add `operator==` as well and implement `!=` from it instead?

src/core/thread/mle_types.hpp
```diff
@@ -567,6 +568,61 @@ class LeaderData : public otLeaderData
...
+     *
+     */
+    void Remove(uint8_t aRouterId) { mRouterIdSet[aRouterId / 8] &= ~(0x80 >> (aRouterId % 8)); }
+
+    /**
+     * This method returns whether or not the Router ID sets are equal.
+     *
+     * @param[in]  aOther The other Router ID Set to compare with.
+     *
+     * @retval TRUE   If the Router ID sets are equal.
```
abtink: nit: update doc, since this is `!=` and returns `TRUE` if not equal.

src/core/thread/mle_types.hpp
abtink: I also prefer the definition in `router_table.hpp`, but I think it may cause include dependency loops. If not easy to move it, I'd say ok to keep it in `mle_type.hpp`.

src/core/thread/router_table.cpp
```diff
@@ -502,23 +502,25 @@ uint8_t RouterTable::GetLinkCost(Router &aRouter)
...
+
     for (uint8_t routerId = 0; routerId <= Mle::kMaxRouterId; routerId++)
     {
-        if (aTlv.IsRouterIdSet(routerId) == IsAllocated(routerId))
+        if (aRouterIdSet.Contains(routerId) == IsAllocated(routerId))
         {
             continue;
         }
 
         allocationChanged = true;
```
abtink: Can we consider removing the `allocationChanged` variable (and the related `if` below), since we know after the newly added `VerifyOrExit(mAllocatedRouterIds != aRouterIdSet)` check, some router id has definitely changed.

src/core/thread/router_table.cpp
```diff
@@ -502,23 +502,25 @@ uint8_t RouterTable::GetLinkCost(Router &aRouter)
...
-    mRouterIdSequence            = aTlv.GetRouterIdSequence();
+    mRouterIdSequence            = aRouterIdSequence;
     mRouterIdSequenceLastUpdated = TimerMilli::GetNow();
 
+    VerifyOrExit(mAllocatedRouterIds != aRouterIdSet);
+
     for (uint8_t routerId = 0; routerId <= Mle::kMaxRouterId; routerId++)
     {
-        if (aTlv.IsRouterIdSet(routerId) == IsAllocated(routerId))
+        if (aRouterIdSet.Contains(routerId) == IsAllocated(routerId))
```
abtink: Can we remove this check as well? I think it does not serve any purpose if we remove `allocationChanged`?

src/core/thread/router_table.cpp
```diff
@@ -502,23 +502,25 @@ uint8_t RouterTable::GetLinkCost(Router &aRouter)
...
         {
             continue;
         }
 
         allocationChanged = true;
 
-        if (aTlv.IsRouterIdSet(routerId))
+        if (aRouterIdSet.Contains(routerId))
         {
             mAllocatedRouterIds.Add(routerId);
```
abtink: How about if remove the `mAllocatedRouterIds.Add(routerId)` and the `mAllocatedRouterIds.Remove(routerId);` in the `else`?  And instead after the loop, do a `mAllocatedRouterIds = aRouterIdSet` to copy all the bits at the same time.

This should hep make the the loop simple: We go through all the router id values and check the ones that are removed (are contained in the old `mAllocatedRouterIds` but not in the new `aRouterIdSet`) and do the `SetNextHop` and `RemoveRouterLink` for those...

src/core/thread/router_table.cpp
```diff
@@ -539,39 +541,9 @@ void RouterTable::ProcessTlv(const Mle::RouteTlv &aTlv)
...
-
-        allocationChanged = true;
-
-        if (aTlv.IsAssignedRouterIdSet(routerId))
-        {
-            mAllocatedRouterIds.Add(routerId);
-        }
-        else
-        {
-            mAllocatedRouterIds.Remove(routerId);
```
abtink: There is a difference in behavior here where we don't do 
```
     {
            Router *router = GetRouter(routerId);

            OT_ASSERT(router != NULL);
            router->SetNextHop(Mle::kInvalidRouterId);
            RemoveRouterLink(*router);
     }
```
compared to the new common `ProcessRouterIdSet()`. 

I am not sure if it matters. I think it's fine to use common method. @jwhui , @librasungirl  do you agree?

src/core/thread/router_table.hpp
```diff
@@ -312,20 +312,21 @@ class RouterTable : public InstanceLocator
...
-     * This method updates the router table with a received Route TLV.
+     * This method updates the Router ID allocation.
      *
-     * @param[in]  aTlv  A reference to the Route TLV.
+     * @param[in]  aRouterIdSequence  The Router Id Sequence.
+     * @param[in]  aRouterIdSet       A reference to the Router Id Set.
      *
      */
-    void ProcessTlv(const Mle::RouteTlv &aTlv);
+    void ProcessRouterIdSet(uint8_t aRouterIdSequence, const Mle::RouterIdSet &aRouterIdSet);
```
abtink: How about renaming this to `UpdateRouterIdSet()`?
(it is fine for me as is, just a suggestion).

src/core/thread/router_table.cpp
jwhui: Good point. I think the added logic in this case is OK.

src/core/thread/mle_types.hpp
librasungirl: yes, like @abtink mentioned, putting it here was to avoid include dependency loops.

src/core/thread/router_table.cpp
librasungirl: yes, you are right!  updated.

src/core/thread/router_table.hpp
librasungirl: renamed as your suggestion.

src/core/thread/router_table.cpp
librasungirl: good point!

src/core/thread/mle_types.hpp
librasungirl: updated as your suggestion.

Issue #4758, opened 2020-03-30
Problem with compilation of anonymous structs


piotrkoziar:
**Describe the bug**
After recent changes, there is code in OpenThread that uses anonymous structs (include/openthread/netdiag.h). Since not every C++ compiler supports this feature, I got errors during the compilation (Keil):
```
..\..\..\..\openthread\include\openthread/netdiag.h(260): error:  #3093: anonymous structs are only supported in --gnu mode, or when enabled with #pragma anon_unions

          };

..\..\..\..\openthread\include\openthread/netdiag.h(241): warning:  #1165-D: types cannot be declared in anonymous unions
```

**To Reproduce**
Information to reproduce the behavior, including:
1. Commit with #4619 
2. Build OpenThread with Keil compiler

**Solution**
I believe the best solution is to replace anonymous structs in include/openthread/__netdiag.h__ file with unnamed structs instantiated as members.

PR #4760, opened 2020-03-30
[network-diagnostic] replace anonymous structs in netdiag.h

piotrkoziar:
My proposition on #4758


piotrkoziar:
I forget about the "types cannot be declared in anonymous unions" warning. I pushed commit with named union.

PR #4516, opened 2020-01-30
[toranj] update test-012-multi-hop-traffic

abtink:
This commit updates tornaj `test-012-multi-hop-traffic` to make it
more robust by waiting for entire chain topology to be discovered
(router table to be fully updated) before starting message exchange.

This should help with rare failures of this test, which happen when an
address query response is received before the "next hop" for the
destination router is discovered (through Link Advertisement
exchanges) on the first router `r1`.

PR #4761, opened 2020-03-30
[toranj] change start.sh to remove variable check to retry failed test

abtink:
This commit changes the `start.sh` script to remove the check for the
environment variable (which is specific to travis) before allowing
retry of failed tests. This should help with robustness of the tests
on GitHub Actions.

### DIFF
--- a/tests/toranj/start.sh
+++ b/tests/toranj/start.sh
@@ -58,15 +58,13 @@ run() {
             return
         fi
 
-        # On Travis, we allow a failed test to be retried up to 3 attempts.
-        if [ "$BUILD_TARGET" = "toranj-test-framework" ]; then
-            if [ "$counter" -lt 2 ]; then
-                counter=$((counter+1))
-                echo Attempt $counter running "$1" failed. Trying again.
-                cleanup
-                sleep 10
-                continue
-            fi
+        # We allow a failed test to be retried up to 3 attempts.
+        if [ "$counter" -lt 2 ]; then
+            counter=$((counter+1))
+            echo Attempt $counter running "$1" failed. Trying again.
+            cleanup
+            sleep 10
+            continue
         fi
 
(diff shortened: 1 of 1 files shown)


tests/toranj/start.sh
```diff
@@ -58,15 +58,13 @@ run() {
             return
         fi
 
-        # On Travis, we allow a failed test to be retried up to 3 attempts.
-        if [ "$BUILD_TARGET" = "toranj-test-framework" ]; then
-            if [ "$counter" -lt 2 ]; then
```
simonlingoogle: Can we make the `toranj` tests not fail in the first place?

abtink: @simonlingoogle, that's always the main goal and and  in many cases when investigating why some test fail rarely, we find some issue/bug in the some part of the the code that we fix (in some cases it's an issue in core OT, or sometime in the platform layer simulation or driver/wpantund). 
In some cases it can be assumption in the way the test script is written which we address (a good example of this case is #4516).

That said, the retries are for issues/failures that are usually beyond our immediate control and  may be related to OS/processes/sockets/dbus/pipes, etc. For example, socket tx (which we use for posix simulation) is not guaranteed to be 100% reliable, or IP messages can be dropped/delayed by netif (used by wpantund), or port number being blocked (test cannot bind to randomly selected port number). Many of these are rare but possible (e.g., perfectly valid for linux `netif` to drop an IPv6 message). 

Note that some of the tests run with very high speed-up generating a lot of events/actions that can overflow system/processes (e.g., the `test-602-channel-manager-channel-select.py` emulates about 12 hours of network operation within 4.5s, the test can be made more robust by running it at slower speed up  but at the end, it made more sense to run it fast (to help make the average test run time shorter) and allow retry for occasional failures.

PR #4764, opened 2020-03-31
[posix] fix mld process invalid arguments

gjc13:


src/posix/platform/netif.cpp
```diff
@@ -464,7 +464,7 @@ static void processMLDEvent(otInstance *aInstance)
     uint8_t             buffer[kMaxMLDEvent];
     ssize_t             bufferLen = -1;
     struct sockaddr_in6 srcAddr;
-    socklen_t           addrLen;
+    socklen_t           addrLen = sizeof(srcAddr);
```
bukepo:
```suggestion
    socklen_t           addrLen  = sizeof(srcAddr);
```

PR #4765, opened 2020-03-31
[mac] remove disable CSMA-CA on the last transmit

bukepo:
This PR simply removes this feature which doesn't conforms with specs.


src/core/config/openthread-core-config-check.h
```diff
@@ -394,11 +389,6 @@
 #error "OPENTHREAD_CONFIG_ENABLE_DYNAMIC_MPL_INTERVAL was replaced by OPENTHREAD_CONFIG_MPL_DYNAMIC_INTERVAL_ENABLE."
 #endif
 
-#ifdef OPENTHREAD_CONFIG_DISABLE_CSMA_CA_ON_LAST_ATTEMPT
-#error \
-    "OPENTHREAD_CONFIG_DISABLE_CSMA_CA_ON_LAST_ATTEMPT was replaced by OPENTHREAD_CONFIG_MAC_DISABLE_CSMA_CA_ON_LAST_ATTEMPT."
-#endif
-
```
abtink: How about if we keep an error that says `OPENTHREAD_CONFIG_MAC_DISABLE_CSMA_CA_ON_LAST_ATTEMPT` was removed? (just to be safe if it is used in a project?)

PR #4766, opened 2020-03-31
[meshcop] generate PSKc from passphrase

bukepo:


src/core/meshcop/meshcop.hpp
```diff
@@ -62,6 +66,25 @@ inline Coap::Message *NewMeshCoPMessage(Coap::CoapBase &aCoap)
     return aCoap.NewMessage(&settings);
 }
 
+/**
+ * This static method generates PSKc.
```
bukepo:
```suggestion
 * This function generates PSKc.
```

src/cli/cli.cpp
```diff
@@ -1716,7 +1716,20 @@ void Interpreter::ProcessPskc(uint8_t aArgsLength, char *aArgs[])
...
         otPskc pskc;
 
-        VerifyOrExit(Hex2Bin(aArgs[0], pskc.m8, sizeof(pskc)) == OT_PSKC_MAX_SIZE, error = OT_ERROR_INVALID_ARGS);
+        if (aArgsLength == 1)
+        {
+            VerifyOrExit(Hex2Bin(aArgs[0], pskc.m8, sizeof(pskc)) == OT_PSKC_MAX_SIZE, error = OT_ERROR_INVALID_ARGS);
+        }
+        else if (!strcmp(aArgs[0], "-p"))
+        {
+            otDatasetGeneratePskc(aArgs[1], otThreadGetNetworkName(mInstance), otThreadGetExtendedPanId(mInstance),
```
jwhui, line 1725: In the case that the dataset includes a Thread Network Name and/or Extended PAN ID, I think it is confusing that the PSKc comes from the current state in the stack. How about extracting the value of those fields if they are already defined in the dataset. If not, fall back to using `otThreadGetNetworkName()` and `otThreadGetExtendedPanId()`?

bukepo: otThreadGetNetworkName() and otThreadGetExtendedPanId() is the same as the values in dataset because mle restores dataset. And pskc cannot be changed after the device is already attached.

src/cli/README_DATASET.md
```diff
@@ -374,13 +374,16 @@ Done
 
 ### pskc
 
-Usage: `dataset pskc <key>`
+Usage: `dataset pskc [-p] <key>|<passphrase>`
```
jwhui, line 377: This is not what's implemented below.

src/cli/cli.cpp
jwhui: Thie docs you wrote above confused me. Are you looking to also support updating the PSKc via the `dataset` CLI command?

include/openthread/dataset.h
```diff
@@ -426,6 +426,25 @@ otError otDatasetSendMgmtPendingSet(otInstance *                aInstance,
...
+ * @param[in]  aNetworkName  The network name for PSKc computation.
+ * @param[in]  aExtPanId     The extended PAN ID for PSKc computation.
+ * @param[out] aPskc         A pointer to variable to output the generated PSKc.
+ *
+ * @retval OT_ERROR_NONE          Successfully generate PSKc.
+ * @retval OT_ERROR_INVALID_ARGS  If any of the input arguments is invalid.
+ *
+ */
+otError otDatasetGeneratePskc(const char *           aPassPhrase,
+                              const char *           aNetworkName,
```
jwhui, line 444: What about using `otNetworkName` defined in this file?

bukepo: otThreadGetNetworkName() returns `const char*`. I think keeping `const char*` is simpler. Thoughts?

src/cli/cli.cpp
bukepo: dataset can be changed at runtime by FTD and MTD, however MTD doesn't support computing t PSKc from passphrase. To reduce the code size of MTD, I prefer not add this into dataset command, throught?

include/openthread/dataset.h
edmont: I would prefer using defined types when available. Makes code better understandable. Specially useful when searching for occurrences of this among the whole project files.

include/openthread/dataset.h
```diff
@@ -426,6 +426,25 @@ otError otDatasetSendMgmtPendingSet(otInstance *                aInstance,
                                     const uint8_t *             aTlvs,
                                     uint8_t                     aLength);
 
+/**
+ * This function generates PSKc from a given pass-phrase, network name, and extended PAN Id.
```
edmont, line 430:
```suggestion
 * This function generates PSKc from a given pass-phrase, network name, and extended PAN ID.
```

src/core/meshcop/meshcop.hpp
```diff
@@ -62,6 +66,25 @@ inline Coap::Message *NewMeshCoPMessage(Coap::CoapBase &aCoap)
...
 }
 
+/**
+ * This function generates PSKc.
+ *
+ * PSKc is used to establish the Commissioner Session.
+ *
+ * @param[in]  aPassPhrase   The commissioning passphrase.
+ * @param[in]  aNetworkName  The network name for PSKc computation.
+ * @param[in]  aExtPanId     The extended pan id for PSKc computation.
```
edmont, line 76:
```suggestion
 * @param[in]  aExtPanId     The extended PAN ID for PSKc computation.
```

tests/unit/test_pskc.cpp
```diff
@@ -89,12 +90,12 @@ int main(void)
     return 0;
```
edmont, line 90: What about adding here a test for the example test vector provided in the Thread Spec section 8.4.1.2.2?

src/cli/README.md
```diff
@@ -1144,6 +1144,20 @@ Set the customized data poll period for sleepy end device (milliseconds >= 10ms)
 Done
 ```
 
+### pskc
+
+Usage: `pskc [-p] <key>|<passphrase>`
+
+Set pskc with hex format or passphrase if `-p` is given when the pskc is
```
edmont, line 1173: "Commissioner Credential" term should be explicitly mentioned here. And UTF-8 format specified.

src/cli/README.md
```diff
@@ -80,6 +80,7 @@ Done
 * [preferrouterid](#preferrouterid-routerid)
 * [prefix](#prefix-add-prefix-pvdcsr-prf)
 * [promiscuous](#promiscuous)
+* [pskc](#pskc--p-key-passphrase)
```
edmont, line 83: This anchor link doesn't work.

src/cli/README.md
```diff
@@ -1166,6 +1167,17 @@ Set the customized data poll period for sleepy end device (milliseconds >= 10ms)
 Done
 ```
 
+### pskc [-p] \<key\>|\<passphrase\>
+
+with `-p` generate pskc from \<passphrase\>(UTF-8 encoded) together with current network name and extended PAN ID, otherwise set pskc as \<key\>(hex format).
```
edmont, line 1172:
```suggestion
With `-p` generate pskc from \<passphrase\> (UTF-8 encoded) together with current network name and extended PAN ID, otherwise set pskc as \<key\> (hex format). 
```

src/cli/cli.cpp
jwhui: The recommended way for configuring Thread Network parameters is via the Operational Dataset (see [src/cli/README_DATASET.md](https://github.com/openthread/openthread/blob/master/src/cli/README_DATASET.md#quick-start)).

If we are concerned about code size, can we simply exclude the PSKc computation from MTD? If no one calls `otDatasetGeneratePskc()`, the linker should strip the dead code as well.

src/cli/README.md
bukepo:
```suggestion
* [pskc](#pskc--p-keypassphrase)
```

edmont:
There is a problem with this PR. It's good to be able to configure the PSKc based on the Commissioner Credential, but subsequent changes in Network Name or Extended PAN ID would lead to a Commissioner Candidate being unable to derive the matching PSKc based on the three input parameters.

This means the PSKc should be derivated whenever some of the input parameters (Commissioner Credential, Extended PAN ID, Network Name) change. But Thread doesn't define a way to propagate Commissioner Credential (probably it's not secure to do it) so that other devices can do it.

But maybe this is more an issue from the application which handles the datasets than from OpenThread itself.

@bukepo @jwhui thougths?

bukepo:
@edmont I think this feature is added in CLI  only for convenience. Users should be aware of this command should only be called after network name and extended pan id are set. Besides, it's not allowed to change pskc after the device is attached through the `pskc` command(see [the api](https://github.com/openthread/openthread/blob/master/src/core/api/thread_ftd_api.cpp#L349)), same for `networkname` and `extpanid`. They can only be changed through `dataset` command, which will propagate new dataset to leader.

bukepo:
@edmont since this is a CLI command, so the README mentioned it uses the current network name and extended panid. Do you think it adequate as a notice for users?

edmont:
@bukepo maybe we should add a note in `networkname` and `extpanid` commands warning about the effect of modifying them on the Commissioning Credential used for the PSKc derivation.

edmont:
Yeah, those notes look good. What about mentioning Commissioning Credential with upper case? It's Thread terminology (Table 8-2).

PR #4773, opened 2020-04-01
[nrf528xx] update nrf security

piotrkoziar:
* enabled new functionalities in nrf_security module
* updated mbedtls config file for nrf52833 and nrf52840


piotrkoziar:
@LuDuda @e-rk does it look ok for you?

e-rk:
Thatks @piotrkoziar, after discussion we decided to remove additional features from built-in mbedTLS and leave only the nrf_security library changed.

LuDuda:
@e-rk Would that change the overall size?

e-rk:
@LuDuda Effectively reverting built-in mbedtls config affects nRF52833 only. nRF52840 will still have the same increased size.

PR #4776, opened 2020-04-01
[mesh-forwarder] include address resolving queue in message eviction

jwhui:
The existing implementation only considered the forwarding queue when
looking for a message to evict. This ignores message buffers that may
be buffered in other message queues.

This commit adds the ability to evict messages in the address resolver
queue. The lowest priority message from either the forwarding or
address resolver queue is selected for eviction.

### DIFF
--- a/src/core/thread/mesh_forwarder_ftd.cpp
+++ b/src/core/thread/mesh_forwarder_ftd.cpp
@@ -184,36 +184,69 @@ void MeshForwarder::HandleResolved(const Ip6::Address &aEid, otError aError)
 
 otError MeshForwarder::EvictMessage(uint8_t aPriority)
 {
-    otError  error = OT_ERROR_NOT_FOUND;
-    Message *message;
+    otError        error    = OT_ERROR_NOT_FOUND;
+    PriorityQueue *queues[] = {&mResolvingQueue, &mSendQueue};
+    Message *      evict    = NULL;
 
-    VerifyOrExit((message = mSendQueue.GetTail()) != NULL);
-
-    if (message->GetPriority() < aPriority)
+    // search for a lower priority message to evict (choose lowest priority message among all queues)
+    for (uint8_t index = 0; index < OT_ARRAY_LENGTH(queues); index++)
     {
-        VerifyOrExit(!message->GetDoNotEvict());
-        RemoveMessage(*message);
+        for (uint8_t priority = 0; priority < aPriority; priority++)
+        {
+            for (Message *message = queues[index]->GetHeadForPriority(priority); message; message = message->GetNext())
+            {
+                if (message->GetPriority() != priority)
+                {
+                    break;
+                }
+
+                if (message->GetDoNotEvict())
+                {
+                    continue;
+                }
+
+                evict     = message;
+                aPriority = priority;
+                break;
+            }
+        }
+    }
+
+    if (evict != NULL)
+    {
         ExitNow(error = OT_ERROR_NONE);
     }
-    else
+
+    for (uint8_t priority = aPriority; priority < Message::kNumPriorities; priority++)
     {
-        while (aPriority <= Message::kPriorityNet)
+        // search for an equal or higher priority indirect message to evict
+        for (Message *message = mSendQueue.GetHeadForPriority(aPriority); message; message = message->GetNext())
         {
-            for (message = mSendQueue.GetHeadForPriority(aPriority); message && (message->GetPriority() == aPriority);
-                 message = message->GetNext())
+            if (message->GetPriority() != priority)
             {
-                if (message->IsChildPending())
-                {
-                    RemoveMessage(*message);
-                    ExitNow(error = OT_ERROR_NONE);
-                }
+                break;
             }
 
-            aPriority++;
+            if (message->GetDoNotEvict())
+            {
+                continue;
+            }
+
+            if (message->IsChildPending())
+            {
+                evict = message;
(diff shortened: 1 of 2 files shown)


src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -184,36 +184,71 @@ void MeshForwarder::HandleResolved(const Ip6::Address &aEid, otError aError)
...
-    Message *message;
+    otError        error    = OT_ERROR_NOT_FOUND;
+    PriorityQueue *queues[] = {&mResolvingQueue, &mSendQueue};
+    Message *      evict    = NULL;
 
-    VerifyOrExit((message = mSendQueue.GetTail()) != NULL);
+    // search for a lower priority message to evict (choose lowest priority message among all queues)
+    for (uint8_t index = 0; index < OT_ARRAY_LENGTH(queues); index++)
+    {
+        for (uint8_t priority = 0; priority < aPriority; priority++)
```
simonlingoogle, line 194: nit: small optimization
```suggestion
        for (uint8_t priority = 0; priority < (evict ? evict->GetPriority() : aPriority); priority++)
```

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -184,36 +184,71 @@ void MeshForwarder::HandleResolved(const Ip6::Address &aEid, otError aError)
...
+                if (message->GetDoNotEvict())
+                {
+                    continue;
+                }
+
+                if ((evict == NULL) || (message->GetPriority() < evict->GetPriority()))
+                {
+                    evict = message;
+                    break;
+                }
```
simonlingoogle, line 212:
```suggestion
                evict = message;
                break;
```
due to optimization above.

src/core/thread/mesh_forwarder_ftd.cpp
jwhui: @simonlingoogle, great idea! I implemented this optimization in a different way. Please review.

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -184,36 +184,69 @@ void MeshForwarder::HandleResolved(const Ip6::Address &aEid, otError aError)
...
+                ExitNow(error = OT_ERROR_NONE);
+            }
         }
     }
 
 exit:
+
+    if (error == OT_ERROR_NONE)
+    {
+        RemoveMessage(*evict);
```
simonlingoogle, line 247: evict can be NULL here?

jwhui: If `error == OT_ERROR_NONE`, then `evict` should not be NULL. Am I missing something?

simonlingoogle: Sorry, didn't realize that error is initialized to `OT_ERROR_NOT_FOUND`.

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -184,36 +184,69 @@ void MeshForwarder::HandleResolved(const Ip6::Address &aEid, otError aError)
...
 
-    if (message->GetPriority() < aPriority)
+                evict     = message;
+                aPriority = priority;
+                break;
+            }
+        }
+    }
+
+    if (evict != NULL)
```
simonlingoogle, line 215: nit: 
```suggestion
    VerifyOrExit(evict == NULL, error = OT_ERROR_NONE);
```

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -184,36 +184,69 @@ void MeshForwarder::HandleResolved(const Ip6::Address &aEid, otError aError)
...
+            {
+                if (message->GetPriority() != priority)
+                {
+                    break;
+                }
+
+                if (message->GetDoNotEvict())
+                {
+                    continue;
+                }
```
Irving-cl, line 206: Can we change the two layer for loop to one like:
```suggestion
    for (Message *message = queues[index]->GetHead(); message && message->GetPriority() < aPriority; message = message->GetNext())
    {
        if (message->GetDoNotEvict())
        {
            continue;
        }
```
I checked some code of `PriorityQueue`:
```
Message *PriorityQueue::GetHeadForPriority(uint8_t aPriority) const
{
    Message *head;
    Message *previousTail;

    if (mTails[aPriority] != NULL)
    {
        previousTail = FindFirstNonNullTail(PrevPriority(aPriority));

        OT_ASSERT(previousTail != NULL);

        head = previousTail->Next();
    }
    else
    {
        head = NULL;
    }

    return head;
}
```
So it seems that the tail of one priority links to the head of next priority. Then I think we can traverse it from low priority to high priority in this way, a little simpler.

jwhui: I don't think that works.

`PriorityQueue::GetHead()` returns the highest priority message. Instead, we want to start with the lowest priority messages.

Note that the `PriorityQueue` implementation relies on a circular linked list:
https://github.com/openthread/openthread/blob/64e03b834741d025aa575f46f692225869031b4d/src/core/common/message.cpp#L799-L802
So it finds the tail of the lowest priority message, then returns the next message, which should be the highest priority message (due to the circularly linked list).

Thoughts?

Irving-cl: Ok. That's true.

Issue #4769, opened 2020-04-01
Make it possible for otPlatReset to be a callback


rob-the-dude:
**Is your feature request related to a problem? Please describe.**

On NetBSD, I'm using "crunch" to build an image (it's similar to Linux's "busy box") that effectively rebuilds all programs as a single executable.  To do this, it generally uses libraries as-is (the openthread libraries work fine), but it runs each executable's symbols through a script that tweaks the names to prevent conflicts.  Unfortunately, the openthread libraries link against a "otPlatReset" which is built into the executable -- but because the executable's symbols have been tweaked, the link cannot happen.

**Describe the solution you'd like**
I have a patch that allows you to OPTIONALLY (at compile time, via a new switch) build the libraries and executables in a way that work.  With this patch, "otPlatReset" is implemented by the library but it calls a function pointer that must be provided via a new API, "otSetPlatResetCallback".  The executable calls that API at startup and everything works fine.  It does add some code space and a variable for the function pointer, but I'm only using it on POSIX where that's OK for my application (and again, the whole feature is optional).

**Describe alternatives you've considered**
Haven't really tried any.

**Additional context**

jwhui:
@rob-the-dude, thanks for raising this issue.

I am generally OK with the approach. How about defining `otPlatReset()` somewhere in the platform driver layer (i.e. [src/posix/platform](https://github.com/openthread/openthread/tree/master/src/posix/platform)) and defining an API specific to the posix platform for use in [src/posix/main.c](https://github.com/openthread/openthread/blob/master/src/posix/main.c). In this way, we don't need to make any changes to the existing `otPlat*` APIs. Thoughts?

bukepo:
Hi @rob-the-dude , I don't quite understand the problem yet.  do you mean openthread libraries are dynamically linked? If so can static link solve this issue?

rob-the-dude:
Hi, @bukepo.  No, static linking doesn't solve the issue -- everything in crunch is statically linked.

Here's how crunch works:

#1) First, you build all the libraries you want in your system.  All of the openthread libraries that run on the host are built during that stage, and they work more or less as you would expect them in non-crunch builds.

#2) Next, you COMPILE the sources for every executable in the system (this is where "ot-cli" or "ot-daemon" get built).  But before you link, you run all the object files through a name-mangling process that pre-pends the executable's name onto the symbol.  This is necessary, because many executables have functions with the same name (indeed, ALL of them have at least one conflict: "main").

[...]

PR #4778, opened 2020-04-01
allow otPlatReset to be linked as part of platform library

rob-the-dude:
Addresses #4769.  Provides an optional compile-time flag (OPENTHREAD_LINK_PLAT_RESET_VIA_CALLBACK).  If set, otPlatReset function will be defined in the platform layer, and a new API to set a callback into the application layer will be exposed.

### DIFF
--- a/include/openthread/platform/misc.h
+++ b/include/openthread/platform/misc.h
@@ -61,6 +61,11 @@ extern "C" {
  */
 void otPlatReset(otInstance *aInstance);
 
+#if OPENTHREAD_LINK_PLAT_RESET_VIA_CALLBACK
+typedef void (*otPlatResetCallbackFunction)(otInstance *aInstance);
+void otSetPlatResetCallback(otPlatResetCallbackFunction f);
+#endif
+
 /**
  * Enumeration of possible reset reason codes.
  *
--- a/src/posix/platform/misc.cpp
+++ b/src/posix/platform/misc.cpp
@@ -43,6 +43,22 @@
 static otPlatResetReason   sPlatResetReason   = OT_PLAT_RESET_REASON_POWER_ON;
 static otPlatMcuPowerState gPlatMcuPowerState = OT_PLAT_MCU_POWER_STATE_ON;
 
+#if OPENTHREAD_LINK_PLAT_RESET_VIA_CALLBACK
+static otPlatResetCallbackFunction otPlatResetCallback = NULL;
+
+void otSetPlatResetCallback(otPlatResetCallbackFunction f)
+{
+    otPlatResetCallback = f;
+}
+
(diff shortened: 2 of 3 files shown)


src/posix/platform/misc.cpp
```diff
@@ -43,6 +43,22 @@
 static otPlatResetReason   sPlatResetReason   = OT_PLAT_RESET_REASON_POWER_ON;
 static otPlatMcuPowerState gPlatMcuPowerState = OT_PLAT_MCU_POWER_STATE_ON;
 
+#if OPENTHREAD_LINK_PLAT_RESET_VIA_CALLBACK
```
jwhui, line 46: How about removing this build option and always defining `otPlatReset()` here?

include/openthread/platform/misc.h
```diff
@@ -61,6 +61,11 @@ extern "C" {
  */
 void otPlatReset(otInstance *aInstance);
 
+#if OPENTHREAD_LINK_PLAT_RESET_VIA_CALLBACK
+typedef void (*otPlatResetCallbackFunction)(otInstance *aInstance);
+void otSetPlatResetCallback(otPlatResetCallbackFunction f);
```
jwhui, line 66: How about moving this to [src/posix/platform/include/openthread/openthread-system.h](https://github.com/openthread/openthread/blob/master/src/posix/platform/include/openthread/openthread-system.h)?

If yes, we could rename something like `otSysSetResetCallback()`.

rob-the-dude: When I move it, do I need to keep the OPENTHREAD_LINK_PLAT_RESET_VIA_CALLBACK build option?  Or should I just move it there, but no other platforms will implement it?

rob-the-dude: Oh, wait, never mind.  I see that I'm moving it to a posix-only header file.  Got it.

bukepo:
@rob-the-dude did you get a chance to update this PR?

rob-the-dude:
Actually, given that there are other APIs that do this (they just don't impact my particular application), and the external solution that crunch has (thanks for pushing me to find it!), I'm fine with closing this and leaving it as is.  I think there's a reasonable argument for providing callbacks for all such APIs, but I personally don't think doing it on a case-by-case basis sets a good example.

If you guys disagree and WANT this change, then I'm happy to tweak, rebase and fix.  But otherwise, I can close and abandon.  Thoughts?

bukepo:
I'm fine with either way.

rob-the-dude:
After realizing that there are other APIs that would need this, I think it's better to tackle them all via a separate feature.  But I don't think that's a high priority.  I'm going to close this one.

PR #4779, opened 2020-04-01
[network-data] periodically check for stale child ids

jwhui:
A parent is responsible for removing stale child entries from the
network data. The existing implementation only triggered a
SVR_DATA.ntf message when new network data is received. In some cases,
the SVR_DATA.ntf message may not be sent due to rate limiting or lack
of message buffers.

This commit turns this process into a periodic check to ensure that
stale child entires are removed even if a failure in sending the
SVR_DATA.ntf occurs.

### DIFF
--- a/src/core/thread/network_data_leader_ftd.cpp
+++ b/src/core/thread/network_data_leader_ftd.cpp
@@ -1504,6 +1504,23 @@ void Leader::HandleTimer(void)
     }
 }
 
+void Leader::RemoveStaleChildEntries(void)
+{
+    Iterator iterator = kIteratorInit;
+    uint16_t rloc16;
+
+    while (GetNextServer(iterator, rloc16) == OT_ERROR_NONE)
+    {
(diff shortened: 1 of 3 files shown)


src/core/thread/network_data_leader_ftd.cpp
```diff
@@ -1504,6 +1504,23 @@ void Leader::HandleTimer(void)
     }
 }
 
+void Leader::RemoveStaleChildEntries(void)
+{
+    Iterator iterator = kIteratorInit;
+    uint16_t rloc16;
+
+    while (GetNextServer(iterator, rloc16) == OT_ERROR_NONE)
```
abtink, line 1512: LGTM. 

I think moving this to `network_data_leader_ftd` is great. With this change, later we can remove the `GetNextServer()` and have a direct implementation of going through all the entries here instead.

mwilniewiec:
Looks good, Will apply to our setup, thanks.

PR #4782, opened 2020-04-03
[ip6] disable security for link-local packets from unsecure ports

gjc13:
When a device allows unsecure ports for new joining devices, it needs
to reply with unsecure port as well.  Once the device has securely
joined, we should remove the unsecure port.


src/core/net/ip6_filter.cpp
```diff
@@ -173,6 +173,23 @@ otError Filter::RemoveUnsecurePort(uint16_t aPort)
     return error;
 }
 
+bool Filter::HasUnsecurePort(uint16_t aPort)
```
bukepo, line 176:
```suggestion
bool Filter::IsPortUnsecure(uint16_t aPort)
```

src/core/net/ip6.cpp
```diff
@@ -1230,6 +1231,22 @@ otError Ip6::HandleDatagram(Message &aMessage, Netif *aNetif, const void *aLinkM
...
+            uint16_t dstPort;
+
+            // TCP/UDP shares header uint16_t srcPort, uint16_t dstPort
+            VerifyOrExit(aMessage.Read(aMessage.GetOffset() + sizeof(uint16_t), sizeof(dstPort), &dstPort) ==
+                             sizeof(dstPort),
+                         error = OT_ERROR_PARSE);
+            dstPort = HostSwap16(dstPort);
+            if (aMessage.IsLinkSecurityEnabled() && Get<ot::Ip6::Filter>().HasUnsecurePort(dstPort))
+            {
+                otLogInfoIp6("Will remove unsecure port %d", dstPort);
```
bukepo, line 1245: Move to after the line below?
```suggestion
                otLogInfoIp6("Removed unsecure port %u", dstPort);
```

src/core/net/ip6.cpp
```diff
@@ -1268,6 +1285,22 @@ otError Ip6::HandleDatagram(Message &aMessage, Netif *aNetif, const void *aLinkM
...
+            if (nextHeader == kProtoTcp || nextHeader == kProtoUdp)
+            {
+                uint16_t sourcePort;
+
+                VerifyOrExit(aMessage.Read(aMessage.GetOffset(), sizeof(sourcePort), &sourcePort) == sizeof(sourcePort),
+                             error = OT_ERROR_PARSE);
+                sourcePort = HostSwap16(sourcePort);
+                if (Get<ot::Ip6::Filter>().HasUnsecurePort(sourcePort))
+                {
+                    otLogInfoIp6("Will disable link security for packet to %s",
```
bukepo, line 1298:
```suggestion
                    otLogInfoIp6("Disabled link security from unsecure port %u",
```

gjc13:
@librasungirl @bukepo @abtink 
Shall we add a feature flag to enable/disable this behavior and disable it by default?

librasungirl:
@gjc13 personally,  I am still not in favor of adding such application required tricks into OT core even with some feature flag. It's the responsibility of the upper layer to ensure the removal of unsecure Port and properly disable Link layer security for the message to send.

bukepo:
@librasungirl Unsecured port feature is implemented in OpenThread's IP stack. I treat auto removing an unsecured port on receiving a secured packet a feature of unsecured port itself. So I think this PR should update the document for unsecured port feature. @jwhui

src/core/net/ip6.cpp
```diff
@@ -1230,6 +1231,22 @@ otError Ip6::HandleDatagram(Message &aMessage, Netif *aNetif, const void *aLinkM
...
+
+            // TCP/UDP shares header uint16_t srcPort, uint16_t dstPort
+            VerifyOrExit(aMessage.Read(aMessage.GetOffset() + sizeof(uint16_t), sizeof(dstPort), &dstPort) ==
+                             sizeof(dstPort),
+                         error = OT_ERROR_PARSE);
+            dstPort = HostSwap16(dstPort);
+            if (aMessage.IsLinkSecurityEnabled() && Get<ot::Ip6::Filter>().IsPortUnsecure(dstPort))
+            {
+                Get<ot::Ip6::Filter>().RemoveUnsecurePort(dstPort);
+                otLogInfoIp6("Removed unsecure port %d", dstPort);
```
abtink, line 1246: This practically changes and stricts the use of the unsecure port number:
- We can have only one node using unsecure port (during pairing) - since as soon as the node attaches and send traffic with security, we remove the unsecure port.
- We cannot use the same port number for communication to other already attached node (with security enabled).

If we go for this, I think we should ensure to update the documentation to mention the change.

src/core/net/ip6.cpp
```diff
@@ -1268,6 +1285,22 @@ otError Ip6::HandleDatagram(Message &aMessage, Netif *aNetif, const void *aLinkM
...
+                uint16_t sourcePort;
+
+                VerifyOrExit(aMessage.Read(aMessage.GetOffset(), sizeof(sourcePort), &sourcePort) == sizeof(sourcePort),
+                             error = OT_ERROR_PARSE);
+                sourcePort = HostSwap16(sourcePort);
+                if (Get<ot::Ip6::Filter>().IsPortUnsecure(sourcePort))
+                {
+                    aMessage.SetLinkSecurityEnabled(false);
+                    otLogInfoIp6("Disabled link security for packet to %s",
+                                 header.GetDestination().ToString().AsCString());
```
abtink, line 1300: This is rather big change in how the unsecure port is used compared to before.

Unseucre port was only related to **allowing rx** of IPv6 message (link-local address) from the port number (without link layer security). This change now applies the port to the **IPv6 message transmissions** (to same port) as well.

I still have the same concern on this that this code now overrides (removes) the link security on the message even if the user has explicitly requested it (if the port matches the unsecure port).  I personally feel we should follow what host/app layer and not override/reduce the security-level (keep the logic as before).

I this it may be ok  if we do it the other way (add link-layer security when user did not request it), i.e. check if  user has asked the message to be sent without link-security but the destination is an attached neighbor, then require security.  

Thoughts?

abtink: One idea here is to add a new flag/option in `otMessageSettings` related to link-layer security. Basically have 3 options:
- Host/user requires link-layer security
- Host/user explicitly asks for no link-layer security
- Host/user delegates it to lower level stack

We can then apply the rule about the unsecure port only if user delegated it to stack. 

Thoughts?

src/core/net/ip6.cpp
bukepo: I  agree better document is required.

gjc13:
@abtink @librasungirl @bukepo 
We've decided to add a feature flag for it and have added the feature description.

src/core/config/ip6.h
```diff
@@ -160,4 +160,20 @@
...
+/**
+ * @def OPENTHREAD_CONFIG_DYNAMIC_UNSECURE_PORT
+ *
+ * Define as 1 to enable dynamic unsecure port management
+ *
+ * If this feature is enabled, OpenThread will automaitically disable link-level
+ * security for packets sent with unsecure source ports. Once we receive a secure
+ * packet sent to the unsecure port, this port will be removed from the unsecure
+ * port list.
+ * Enable only if you know well about its behavior.
```
bukepo, line 172:
```suggestion
  *
 * Enable only if you know well about its behavior.
```

src/core/net/ip6_filter.cpp
```diff
@@ -173,6 +173,23 @@ otError Filter::RemoveUnsecurePort(uint16_t aPort)
...
+bool Filter::IsPortUnsecure(uint16_t aPort)
+{
+    bool found = false;
+
+    for (int i = 0; i < kMaxUnsecurePorts; i++)
+    {
+        if (mUnsecurePorts[i] == aPort)
+        {
+            found = true;
+            ExitNow();
```
bukepo, line 185:
```suggestion
            break;
```

src/core/config/ip6.h
```diff
@@ -160,4 +160,20 @@
...
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_DYNAMIC_UNSECURE_PORT
+#define OPENTHREAD_CONFIG_DYNAMIC_UNSECURE_PORT 0
```
bukepo, line 176: Enable this feature in Android.mk?

src/core/config/ip6.h
```diff
@@ -160,4 +160,21 @@
...
+ *
+ * If this feature is enabled, OpenThread will automaitically disable link-level
+ * security for packets sent with unsecure source ports. Once we receive a secure
+ * packet sent to the unsecure port, this port will be removed from the unsecure
+ * port list.
+ *
+ * Enable only if you know well about its behavior.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_DYNAMIC_UNSECURE_PORT
```
bukepo, line 176:
```suggestion
#ifndef OPENTHREAD_CONFIG_UNSECURE_PORT_REMOVE_ON_RX_SECURE_ENABLE
```

src/core/net/ip6_filter.hpp
```diff
@@ -97,6 +97,16 @@ class Filter
...
 
+    /**
+     * This method checks whether a port is in the unsecure port list.
+     *
+     * @param[in]  aPort  The port value.
+     *
+     * @returns Whether the given port is in the unsecure port list.
+     *
+     */
+    bool IsPortUnsecure(uint16_t aPort);
```
bukepo, line 108:
```suggestion
    bool IsUnsecurePort(uint16_t aPort);
```

Android.mk
```diff
@@ -66,6 +66,7 @@ ifeq ($(USE_OTBR_DAEMON), 1)
 OPENTHREAD_PUBLIC_CFLAGS                                         += \
     -DOPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE=1                     \
     -DOPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE=1                       \
+    -DOPENTHREAD_CONFIG_UNSECURE_TRAFFIC_MANAGED_BY_STACK_ENABLE=1                     \
```
jwhui, line 69:
```suggestion
    -DOPENTHREAD_CONFIG_UNSECURE_TRAFFIC_MANAGED_BY_STACK_ENABLE=1  \
```

src/core/net/ip6.cpp
```diff
@@ -1230,6 +1231,24 @@ otError Ip6::HandleDatagram(Message &aMessage, Netif *aNetif, const void *aLinkM
...
+
+            // TCP/UDP shares header uint16_t srcPort, uint16_t dstPort
+            VerifyOrExit(aMessage.Read(aMessage.GetOffset() + sizeof(uint16_t), sizeof(dstPort), &dstPort) ==
+                             sizeof(dstPort),
+                         error = OT_ERROR_PARSE);
+            dstPort = HostSwap16(dstPort);
+            if (aMessage.IsLinkSecurityEnabled() && Get<ot::Ip6::Filter>().IsUnsecurePort(dstPort))
+            {
+                Get<ot::Ip6::Filter>().RemoveUnsecurePort(dstPort);
+                otLogInfoIp6("Removed unsecure port %d", dstPort);
```
jwhui, line 1247: Should this log be moved inside `RemoveUnsecurePort()`?

Issue #3880, opened 2019-05-29
Move language standard baseline to C++11?


jwhui:
The current language standard baseline is C++03 (see [STYLE_GUIDE.md](https://github.com/openthread/openthread/blob/master/STYLE_GUIDE.md#standards)).

Thoughts on whether we should move to supporting C++11?

One use case was raised in #3877.

abtink:
I am in favor in general. We should consider the list of the features that would be helpful (`auto`, ranged-`for` loops, `constexpr`,  `enum class`, etc.) and if there are any features we may want to limit the use of?
(new function syntax with `->` to specify return type, lambda with capture?)

KaSroka:
As C++ 20 is coming I think that we may consider newer standards as well (if they have features that may help us). Also we are building with std=gnu++98 configured currently, not 03. And the Android build have no standard defined (I cannot see any std= flag when running Android build on Travis)

KaSroka:
Here is a nice brief of new features in each version: https://github.com/AnthonyCalandra/modern-cpp-features

jwhui:
@KaSroka, we can consider adopting newer versions of the C++ language standard. At the same time, we should also consider compiler support as well. If there are good use cases for using newer language features that are well supported by compilers, we can discuss further.

KaSroka:
After looking into features of C++14 and 17 i cannot see any features that may help us maybe except relaxed requirements of constexpr in C++14 as well as no implicit const when using them.

gjc13:
Personally I found such code snippet to be appealing if we can adopt C++11 (which introduces argument pack in templates and std::forward):
```
#include <utility>

extern void LogInfo(const char *format, ...);

template <typename classname, typename U, typename RetType, typename ... ArgType>
class CtxBinder {
public:
    typedef RetType (classname::*Funct)(ArgType...);

    CtxBinder(U ctx, Funct f)
        :mctx(ctx), mf(f) {
    }

    RetType operator()(ArgType... args) {
       return (mctx->*mf)(std::forward<ArgType...>(args...));
    }

private:
    U mctx;
    Funct mf;
};

class A {
public:
    A();

    void f(int x);

    void trigger();

private:
    CtxBinder<A, A*, void, int> mf;
};

A::A() 
    :mf (this, &A::f) {
}

void A::f(int x) {
    LogInfo("%d\n", x);
}

void A::trigger() {
    mf(1);
}
```
[...]

PR #4784, opened 2020-04-03
[error] force error checking

wgtdkp:
This PR forces error checking by applying the `warn_unused_result` attribute on the `otError` enum (See https://github.com/openthread/ot-commissioner/pull/34 for more detail).

There are hundreds cases in openthread codebase that ignore error silently. To make this PR easier to be reviewed, I made no changes to the logic but always wrap function calls with `IgnoreError` to get rid of the warning/error. One exception is that, `RemoveRouterLink` has different signature btween FTD and MTD, this PR unify it to return `void`.

**NOTE**
**This PR also changes openthread public APIs. That is, applications building atop of the openthread library will not compile (if specified -Werror) when it silently ignores the returned `otError` of any openthread public APIs** (in case the application is compiled with clang). But the application can always get rid of the `unused-return-value` warning by using `-Wno-unused-result`. @jwhui @librasungirl @bukepo @abtink  How do you think about this?

### DIFF
--- a/include/openthread/platform/toolchain.h
+++ b/include/openthread/platform/toolchain.h
@@ -71,6 +71,8 @@ extern "C" {
  *
  * @note This is currently only available with clang (C++17 implements it
  *       as attribute [[nodiscard]]).
+ * @note To suppress the 'unused-result' warning/error, please use the
+ *       '-Wno-unused-result' compiler option.
  *
  */
 #if defined(__clang__)
--- a/include/openthread/platform/toolchain.h
+++ b/include/openthread/platform/toolchain.h
@@ -64,6 +64,22 @@ extern "C" {
 #endif
 
 /**
+ * @def OT_MUST_USE_RESULT
+ *
+ * Compiler-specific indication that a class or enum must be used when it is
+ * the return value of a function.
+ *
+ * @note This is currently only available with clang (C++17 implements it
+ *       as attribute [[nodiscard]]).
+ *
+ */
+#if defined(__clang__)
+#define OT_MUST_USE_RESULT __attribute__((warn_unused_result))
+#else
+#define OT_MUST_USE_RESULT
+#endif
+
+/**
  * @def OT_TOOL_PACKED_BEGIN
  *
  * Compiler-specific indication that a class or struct must be byte packed.
--- a/src/core/meshcop/dataset.cpp
+++ b/src/core/meshcop/dataset.cpp
@@ -135,8 +135,8 @@ void Dataset::ConvertTo(otOperationalDataset &aDataset) const
             break;
 
         case Tlv::kNetworkName:
-            static_cast<Mac::NetworkName &>(aDataset.mNetworkName)
-                .Set(static_cast<const NetworkNameTlv *>(cur)->GetNetworkName());
+            IgnoreError(static_cast<Mac::NetworkName &>(aDataset.mNetworkName)
+                .Set(static_cast<const NetworkNameTlv *>(cur)->GetNetworkName()));
             aDataset.mComponents.mIsNetworkNamePresent = true;
             break;
 
@@ -194,7 +194,7 @@ otError Dataset::SetFrom(const otOperationalDataset &aDataset)
         tlv.Init();
         tlv.SetSeconds(aDataset.mActiveTimestamp);
         tlv.SetTicks(0);
-        SetTlv(tlv);
+        IgnoreError(SetTlv(tlv));
     }
 
     if (aDataset.mComponents.mIsPendingTimestampPresent)
@@ -203,12 +203,12 @@ otError Dataset::SetFrom(const otOperationalDataset &aDataset)
         tlv.Init();
         tlv.SetSeconds(aDataset.mPendingTimestamp);
         tlv.SetTicks(0);
-        SetTlv(tlv);
+        IgnoreError(SetTlv(tlv));
     }
 
     if (aDataset.mComponents.mIsDelayPresent)
     {
-        SetUint32Tlv(Tlv::kDelayTimer, aDataset.mDelay);
+        IgnoreError(SetUint32Tlv(Tlv::kDelayTimer, aDataset.mDelay));
     }
 
     if (aDataset.mComponents.mIsChannelPresent)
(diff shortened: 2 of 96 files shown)


src/core/meshcop/dataset.cpp
```diff
@@ -263,7 +263,7 @@ otError Dataset::Set(const otOperationalDataset &aDataset)
         tlv.Init();
         tlv.SetSeconds(aDataset.mActiveTimestamp);
         tlv.SetTicks(0);
-        Set(tlv);
+        IgnoreError(error = Set(tlv));
```
simonlingoogle, line 266: Seems logic changed here. If so, `IgnoreError` is still necessary here?
```suggestion
        error = Set(tlv);
```

src/core/meshcop/dataset_manager.cpp
```diff
@@ -160,7 +160,7 @@ otError DatasetManager::Save(const Dataset &aDataset)
 
     if (isMasterkeyUpdated || compare > 0)
     {
-        mLocal.Save(aDataset);
+        SuccessOrExit(error = mLocal.Save(aDataset));
```
simonlingoogle, line 163: logic changed here.

wgtdkp: Thanks @simonlingoogle, that's very careful of you!

include/openthread/platform/toolchain.h
```diff
@@ -63,6 +63,22 @@
...
+ *
+ * Compiler-specific indication that a class or enum must be used when it is
+ * the return value of a function.
+ *
+ * @note This is currently only available with clang (C++17 implements it
+ *       as attribute [[nodiscard]]).
+ *
+ */
+#if defined(__clang__)
+#define OT_MUST_USE_RESULT __attribute__((warn_unused_result))
```
simonlingoogle, line 77: Can we allow OT API user to turn this feature off somehow? I think the major target of this PR is the OT core code.

wgtdkp: I don't see a strategy to make this happen, only if we add a new error class and use the new error class in OT core code. Do you have any suggestion? @simonlingoogle

include/openthread/platform/toolchain.h
```diff
@@ -63,6 +63,22 @@
...
+#else
+#define OT_MUST_USE_RESULT
+#endif
```
simonlingoogle, line 80:
```suggestion
#ifndef OT_MUST_USE_RESULT
#if defined(__clang__)
#define OT_MUST_USE_RESULT __attribute__((warn_unused_result))
#else
#define OT_MUST_USE_RESULT
#endif
#endif
```
A simple solution is allowing users to redefine OT_MUST_USE_RESULT so that they can turn off this feature?

wgtdkp: Is feature switching in public headers a good idea?

wgtdkp: @bukepo @abtink Any idea about this?

abtink: @wgtdkp , yes we cannot use any config options in public headers (since we allow projects to define their own project header and since we don't own the file, we cannot make it part of public headers of OT). 

One way I can think of to allow user decide their own `MUST_USE_RESULT` is the following: (though I should say, it's complex and won't recommend it, just mentioning it as an option):
- We can add a core mirror of `otError` enumeration type (say `ot::Error`) which re-defiens the same enumeration values and use `OT_MUST_USE_RESULT` for `ot::Error` (similar to https://github.com/openthread/openthread/pull/4794 for `otDeviceRole` and `ot::Mle::DeviceRole`).  I think this can work (since the two types would be different and one can have `OT_MUST_USE_RESULT` and other don't).  This will  be big change and require a lot of `static_cast` in all the `{module}_api.cpp` and platform files to convert between `ot::Error` and `otError`.

wgtdkp: Thanks @abtink , it seems to be a big overhead. I come up with another solution: Instead of defining `OT_MUST_USE_RESULT` in public header, we can define it inside OT code and decorate `otError` with `OT_MUST_USE_RESULT` within only OT code. The price is, all OT core source file should always include a wrapper file (like `error_nodicard.hpp`) instead of `openthread/error.h` (This is not a big change, because we don't really explicitly includes `openthread/error.h` for each source file but inherit it) . The `error_nodiscard.hpp` file looks like this:

```c++
#if defined(__clang__)
#define OT_MUST_USE_RESULT __attribute__((warn_unused_result))
#else
#define OT_MUST_USE_RESULT
#endif

class OT_MUST_USE_RESULT otError;

#include <openthread/error.h>
```
In this way, we add `OT_MUST_USE_RESULT` attribute for only OT core code. Thoughts?

simonlingoogle: I think this is the best solution so far.

wgtdkp: @abtink @jwhui thoughts?

abtink: @wgtdkp , if it can be made to work reliably on all tool-chains, it is fine.

Note that OT public APIs are declared in C (not C++), so `otError` is a C `enum` declaration (not a `class` or `enum class`).  I think we'd need at least `C++11` to be able to use forward declaration of `enum` types.   OT currently requires/supports `C++03` and not yet on `C++11` (please see #https://github.com/openthread/openthread/issues/3880).
- Also If I remember correctly in C++11 we can only forward-declare an `enum` with a specified underlying (integral) type/size (or if it is a `enum class` and not `enum`).

Another point to conisder is if we can change the attribute on a forward declaration and not include the same attribute in the main definition. Would this work on all tool-chains?

wgtdkp: @abtink I will verify if `enum OT_MUST_USE_RESULT otError;` can work. To my understanding, the `warn_unused_result` attribute is specific to compiler and doesn't change generated code. So I think it is okay for application to not see this attribute with `otError`.

wgtdkp: @abtink , I verified that `enum OT_MUST_USE_RESULT otError;` works.

wgtdkp:
@simonlingoogle @bukepo @librasungirl This PR is ready for review now~

wgtdkp:
@jwhui We may need to re-run GitHub Actions for other ongoing PRs after merging this PR, because those PRs may do ignore errors silently.

PR #4786, opened 2020-04-03
[mesh-forwarder] fix bug in evicting from address resolver queue

jwhui:
Fixes a bug introduced in #4776 

The existing MeshForwarder::RemoveMessage() implementation is specific
to evicting messages from mSendQueue. This commit generalizes the
method by retrieving the queue from the message itself.

### DIFF
--- a/src/core/common/message.hpp
+++ b/src/core/common/message.hpp
@@ -723,6 +723,17 @@ public:
         return (!mBuffer.mHead.mInfo.mInPriorityQ) ? mBuffer.mHead.mInfo.mQueue.mMessage : NULL;
     }
 
+    /**
+     * This method returns a pointer to the priority message queue (if any) where this message is queued.
+     *
+     * @returns A pointer to the priority queue or NULL if not in any priority queue.
+     *
+     */
+    PriorityQueue *GetPriorityQueue(void) const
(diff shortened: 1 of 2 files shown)


src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -244,7 +247,16 @@ otError MeshForwarder::EvictMessage(uint8_t aPriority)
 
     if (error == OT_ERROR_NONE)
     {
-        RemoveMessage(*evict);
+        if (evictQueue == &mSendQueue)
```
abtink, line 250: Good catch or `RemoveMessage()`.

How about asking the message itself about the queue it is in instead of saving the `evictQueue`?
`Message` tracks its own queue and whether it's a normal or a priority queue?

jwhui: Updated as suggested. I had to make `Message::GetPriorityQueue()` public. Also, the current method only supports priority queues for now given that's all it needs to operate on.

jwhui:
> Am I guessing correctly, that previous code triggered some "assert" due to freeing the message without dequeuing it?

@LuDuda, yes.

PR #4600, opened 2020-02-26
[toranj] add new test-case for cache table snoop

abtink:
This is related and a follow up to PR https://github.com/openthread/openthread/pull/4599.
It adds a toranj test-case verifying behavior of `AddressResolver` and management of cache table.

This PR is related to two other PRs. The change is broken due to dependency we have between new spinel property definitions in OT, wpantund changes and then the new `toranj` test-case.
- The main PR https://github.com/openthread/openthread/pull/4599 (should be merged first). 
- The wpantund PR: https://github.com/openthread/wpantund/pull/462
- This PR https://github.com/openthread/openthread/pull/4600 (which adds the test)

PR #4789, opened 2020-04-04
[address-resolver] ensure prev pointer stays valid on entry removal

abtink:
This commit simplifies the model for iterating through a linked
list by using a `prev` pointer as the iterator variable. This helps
fix an issue with the `prev` pointer not staying  valid when entry
is removed from the list during iteration.

### DIFF
--- a/src/core/thread/address_resolver.cpp
+++ b/src/core/thread/address_resolver.cpp
@@ -189,6 +189,11 @@ void AddressResolver::Remove(uint16_t aRloc16)
     Remove(aRloc16, /* aMatchRouterId */ false);
 }
 
+AddressResolver::CacheEntry *AddressResolver::GetEntryAfter(CacheEntry *aPrev, CacheEntryList &aList)
+{
+    return (aPrev == NULL) ? aList.GetHead() : aPrev->GetNext();
+}
+
 void AddressResolver::Remove(Mac::ShortAddress aRloc16, bool aMatchRouterId)
 {
     CacheEntryList *lists[] = {&mCachedList, &mSnoopedList};
@@ -196,23 +201,23 @@ void AddressResolver::Remove(Mac::ShortAddress aRloc16, bool aMatchRouterId)
     for (uint8_t index = 0; index < OT_ARRAY_LENGTH(lists); index++)
     {
         CacheEntryList *list = lists[index];
-        CacheEntry *    prev;
+        CacheEntry *    prev = NULL;
         CacheEntry *    entry;
-        CacheEntry *    next;
 
-        for (prev = NULL, entry = list->GetHead(); entry != NULL; prev = entry, entry = next)
+        while ((entry = GetEntryAfter(prev, *list)) != NULL)
         {
-            // We keep track of the next entry in the list before
-            // checking the entry's RLOC16 or Router Id since the
-            // entry may be removed from the list if it matches.
-
-            next = entry->GetNext();
-
             if ((aMatchRouterId && Mle::Mle::RouterIdMatch(entry->GetRloc16(), aRloc16)) ||
                 (!aMatchRouterId && (entry->GetRloc16() == aRloc16)))
             {
                 RemoveCacheEntry(*entry, *list, prev, aMatchRouterId ? kReasonRemovingRouterId : kReasonRemovingRloc16);
                 mUnusedList.Push(*entry);
+
+                // If the entry is removed from list, we keep the same
+                // `prev` pointer.
+            }
+            else
+            {
+                prev = entry;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


LuDuda:
@jwhui @abtink side comment. Recently we have found a couple of P1/P2 bugs which OpenThread CI didn't find, but rather (most likely) yours and our longer stress tests. What do you think about addressing this problem on OT CI?

What I'm thinking about is to have tests that stress the network for a greater amount of time (generating running out of message buffers, cache entries and emulating some loss of the packets), so we can detect asserts/faults faster. Because of simulated time, I think, it should be viable to have such tests running on multiple simulated devices possibly for "hours".

jwhui:
> @jwhui @abtink side comment. Recently we have found a couple of P1/P2 bugs which OpenThread CI didn't find, but rather (most likely) yours and our longer stress tests. What do you think about addressing this problem on OT CI?
> 
> What I'm thinking about is to have tests that stress the network for a greater amount of time (generating running out of message buffers, cache entries and emulating some loss of the packets), so we can detect asserts/faults faster. Because of simulated time, I think, it should be viable to have such tests running on multiple simulated devices possibly for "hours".

@LuDuda, I completely agree. In fact, we are working on an improved OpenThread Network Simulator (OTNS) that should allow for much faster simulation speedup by rewriting the simulation driver (see #4268 for the initial evidence of this). We are planning to open source OTNS very soon and use this as a basis for expanded testing scenarios, including stress tests.

/cc/ @simonlingoogle

abtink:
@LuDuda I agree that more testing is always good, the _key in my opinion is in addition to stress testing, to write more test-cases and covering more situations_. At the end, we should be ready to take quick action and investigate issues as no test framework can truly cover all aspects and corner cases (this is nature of active development and product integration). 

This issue itself was found by oss-fuzz testing which I believe is run everyday on master and sort of performs the stress testing that you are mentioning. 

We did also add a a rather detailed test-case covering the behavior of the address query/cache along with the original #4599 from PR #4600 . This test went through [a lot of the behaviors of address reolsver module](https://github.com/openthread/openthread/blob/master/tests/toranj/test-039-address-cache-table-snoop.py) and many corner-cases:
- **Verify query timeout and retry behavior**:
  - Multiple address query to non-existing addresses and verify cache entries start in "query" state.
  - Verify that entries to time out and check that they enter "retry" state.
  - Check that in "retry" state, during retry delay interval no new address query sent for the same tragets
  - Verify that after retry delay passed, new address queries are sent correctly again.
  - Verify that on first time address query the entry cannot be evicted, but in second/next attempts it is allowed to be evicted.
- **Verify snoop behavior:**
  - Verify snoop behavior and that snoop entries are added correctly on rx packet inspection
  - Verify that only MAX_SNOOPED_NON_EVICTABLE of snooped entries are non-evictable.
  - Verify that snoop entries are used and move to "cached" state on tx to same address.
- **Verify cache entry update:**
  - Verify address cache entry update on successful address notification from  routers
  - Verify address cache entry update for request to a sleepy child (and that parent is the rloc) and last transaction in cache entry are updated correctly.
  - Verify cache entry moving to top of list on subsequent use
- **Verify behavior when cache table is full:**
  - Verify that snoop optimization uses at most `MAX_SNOOPED_NON_EVICTABLE` entries and not overwrite other cached entries
  - Verify that snooped entries are converted to "cached" on use.
  - Verify that entries are moved to top of list on use.
  - Verify that on full table, oldest entry is evicted.

As you can see this already covers a lot of things. I did actually spent some time to figure out how come the test did not catch the issue being addressed by this PR (it happens to be rather unfortunate order of events which masked this). I am going to update the test to cover the situation being addressed here.

LuDuda:
Thank you @abtink for the detailed response.

I'm sorry if I badly formulated my intention of the above comment. I just wanted to present some idea I had for a while, and even started to implement something, though the OTNS sounds great.

I agree that this feature is well verified :+1: In general truely in my opinion you do the best job in community with testing e.g. toranj or unit testing. So my comment wasn't directed to you :)

abtink:
@LuDuda no worries at all man...  :)...

I agree and I think we'd need combination of different approaches: more targeted test-cases (to cover corner cases that may be hard to happen by fuzz/stress test), along with more stress/fuzz testing to cover situations we cannot think of. At the end, I am sure bugs may still slip through once in while... but we'd be ready for it... :)

LuDuda:
Let's be ready! :smile:

simonlingoogle:
> What I'm thinking about is to have tests that stress the network for a greater amount of time (generating running out of message buffers, cache entries and emulating some loss of the packets), so we can detect asserts/faults faster. Because of simulated time, I think, it should be viable to have such tests running on multiple simulated devices possibly for "hours".

There has been some work on building stress tests for OpenThread. One outcome of the effort is OTNS, which is to be released soon, that can simulate large number of nodes efficiently. 
The current stress tests covered some basic connectivity tests and commissioning tests, which have revealed a number of bugs in OT core. I think one difference between stress tests and functional tests (e.x. unittests, tranj) is that stress tests tries to make sure application level functionalities are achievable despite the harsh environment.
I think one way to integrate this into OT CI is to run comprehensive stress tests for each master branch commit so that we reveal issues as soon as possible. 
@LuDuda @jwhui

jwhui:
@simonlingoogle, I would prefer to do as much testing as possible before we get into master branch. As @abtink mentioned, we already employ a number of techniques, including unit, functional, fuzzing, static analysis, etc.

I see stress testing as a form of functional testing. Stress tests may cover application scenarios, or something completely random (i.e. monte carlo). It would be great to add these additional test scenarios to our testing mix.

PR #4793, opened 2020-04-05
[bbr] support domain prefix addition and removal

librasungirl:


include/openthread/backbone_router_ftd.h
```diff
@@ -158,6 +159,54 @@ uint8_t otBackboneRouterGetRegistrationJitter(otInstance *aInstance);
...
+ *
+ * @retval OT_ERROR_NONE          Successfully set the Domain Prefix.
+ * @retval OT_ERROR_INVALID_ARGS  @p aConfig is invalid.
+ * @retval OT_ERROR_NO_BUFS       Insufficient space to add the Domain Prefix entry to local server data.
+ *
+ * @sa otBackboneRouterGetDomainPrefix
+ * @sa otBackboneRouterRemoveDomainPrefix
+ *
+ */
+otError otBackboneRouterSetDomainPrefix(otInstance *aInstance, otBorderRouterConfig *aConfig);
```
jwhui, line 208: How about generalize this to `otBackboneRouterAddDomainPrefix()`? The limitation to one domain prefix is currently a "SHOULD".

include/openthread/netdata.h
```diff
@@ -102,6 +102,18 @@ typedef struct otBorderRouterConfig
      */
     bool mStable : 1;
 
+    /**
+     * TRUE, if this border router is able to supply DNS infomration obtained via ND.  FALSE, otherwise.
+     */
+    bool mNdDns : 1;
+
+    /**
+     * TRUE, if this prefix is Thread Domain Prefix.  FALSE, otherwise.
```
jwhui, line 111:
```suggestion
     * TRUE, if this prefix is a Thread Domain Prefix.  FALSE, otherwise.
```

src/cli/README.md
```diff
@@ -1193,29 +1193,42 @@ Done
...
 
-### prefix add \<prefix\> [pvdcsr] [prf]
+### prefix add \<prefix\> [padcrosnD] [prf]
 
 Add a valid prefix to the Network Data.
 
+Note: The Domain Prefix flag(`D`) is only available for Thread 1.2.
+      If the domain prefix is configured to the local backbone router server.
+      And it would not be added to local server data unless backbone router
+      feature is enabled.
```
jwhui, line 1213: I don't think these three lines are necessary. Implementation below shows that CLI will return error if `D` flag is specified.

include/openthread/backbone_router_ftd.h
```diff
@@ -158,6 +159,54 @@ uint8_t otBackboneRouterGetRegistrationJitter(otInstance *aInstance);
...
+ * @param[in]  aPrefix   A pointer to an IPv6 Domain prefix.
+ *
+ * @retval OT_ERROR_NONE          Successfully removed the Domain Prefix configuration.
+ * @retval OT_ERROR_INVALID_ARGS  @p aPrefix is invalid.
+ * @retval OT_ERROR_NOT_FOUND     No Domain Prefix was configured or @p aPrefix doesn't matches.
+ *
+ * @sa otBackboneRouterGetDomainPrefix
+ * @sa otBackboneRouterSetDomainPrefix
+ */
+otError otBackboneRouterRemoveDomainPrefix(otInstance *aInstance, const otIp6Prefix *aPrefix);
```
jwhui, line 190: Can we reuse the existing `otBorderRouterAddPrefix()` and `otBorderRouterRemovePrefix()` methods, rather than defining new ones here? The `D` flag should be sufficient, right?

src/core/thread/network_data_tlvs.hpp
```diff
@@ -634,6 +638,50 @@ class BorderRouterEntry
...
     void SetOnMesh(void) { mFlags |= kOnMeshFlag; }
 
+    /**
+     * This method indicates whether or not the Nd-Dns flag is set.
+     *
+     * @retval TRUE   If the Nd-Dns flag is set.
+     * @retval FALSE  If the Nd-Dns flag is not set.
+     *
+     */
+    bool IsNdDns(void) const { return (mReserved & kNdDnsFlag) != 0; }
```
jwhui, line 648: We should rename `mReserved`.

include/openthread/backbone_router_ftd.h
librasungirl: make sense. removed these two APIs.

src/cli/README.md
librasungirl: removed.

src/core/thread/network_data_tlvs.hpp
librasungirl: will update.

librasungirl: Will update. Any suggestion on the name? @jwhui

src/core/thread/network_data.cpp
```diff
@@ -217,6 +217,10 @@ otError NetworkData::GetNextOnMeshPrefix(Iterator &aIterator, uint16_t aRloc16,
                     aConfig.mOnMesh         = borderRouterEntry->IsOnMesh();
                     aConfig.mStable         = borderRouter->IsStable();
                     aConfig.mRloc16         = borderRouterEntry->GetRloc();
+                    aConfig.mNdDns          = borderRouterEntry->IsNdDns();
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
```
jwhui, line 221: I think we should keep this even for Thread 1.1. Being able to get the Domain Prefix flag may be useful.

src/core/backbone_router/local.cpp
```diff
@@ -242,6 +245,134 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+
+exit:
+    return error;
+}
+
+otError Local::RemoveDomainPrefix(const otIp6Prefix &aPrefix)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(aPrefix.mLength > 0, error = OT_ERROR_INVALID_ARGS);
```
jwhui, line 264: Need to check of `aPrefix.mLength == mDomainPrefixConfig.mPrefix.mLength`?

src/core/backbone_router/local.cpp
```diff
@@ -242,6 +245,134 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+                                                     mDomainPrefixConfig.mPrefix.mLength);
+    }
+
+    mDomainPrefixConfig = aDomainPrefixConfig;
+
+    if (IsEnabled())
+    {
+        uint8_t flags = 0;
+
+        if (mDomainPrefixConfig.mPreferred)
```
jwhui, line 301: This seems to share a lot of code with `otBorderRouterAddOnMeshPrefix()`. Maybe `NetworkData::Local::AddOnMeshPrefix()` should take `OnMeshPrefixConfig` and include this code to avoid duplication?

src/core/thread/network_data.cpp
librasungirl: updated.

src/core/backbone_router/local.cpp
librasungirl: added the check.

src/core/backbone_router/local.cpp
librasungirl: updated as your suggestion.

src/cli/cli.cpp
```diff
@@ -2575,68 +2584,94 @@ otError Interpreter::ProcessPrefixRemove(uint8_t aArgsLength, char *aArgs[])
...
+
+    while (otBorderRouterGetNextOnMeshPrefix(mInstance, &iterator, &config) == OT_ERROR_NONE)
+    {
+        OutputPrefix(config);
+    }
+
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+    if (otBackboneRouterGetState(mInstance) == OT_BACKBONE_ROUTER_STATE_DISABLED)
+    {
+        if (otBackboneRouterGetDomainPrefix(mInstance, &config) == OT_ERROR_NONE)
```
simonlingoogle, line 2668: nit: 
```suggestion
        SuccessOrExit(otBackboneRouterGetDomainPrefix(mInstance, &config))
```

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +241,92 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+    return error;
+}
+
+otError Local::RemoveDomainPrefix(void)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(mDomainPrefixConfig.mPrefix.mLength > 0, error = OT_ERROR_NOT_FOUND);
+
+    SuccessOrExit(error = RemoveDomainPrefix(mDomainPrefixConfig.mPrefix));
```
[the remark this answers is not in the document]
simonlingoogle: This cause `bbr disable` unable to remove domain prefix from network data.

src/cli/cli.cpp
```diff
@@ -2575,68 +2584,94 @@ otError Interpreter::ProcessPrefixRemove(uint8_t aArgsLength, char *aArgs[])
...
+        {
+            OutputPrefix(config);
```
simonlingoogle, line 2670: Since the domain prefix here is not in network data, I would suggest we somehow mark this prefix as ineffective to make it easier for CLI users to tell the difference.

include/openthread/netdata.h
```diff
@@ -102,6 +102,18 @@ typedef struct otBorderRouterConfig
...
+     *
+     * Note: Only available since Thread 1.2.
+     */
+    bool mDp : 1;
```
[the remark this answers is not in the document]
librasungirl: It is `P_dp` in Thread Specification, here follows the existing naming rule, like  `mOnMesh` represents `P_on_mesh` flag in spec.

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +241,92 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+
+otError Local::SetDomainPrefix(const NetworkData::OnMeshPrefixConfig &aConfig)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(aConfig.mPrefix.mLength > 0, error = OT_ERROR_INVALID_ARGS);
+
+    if (IsEnabled())
+    {
+        Get<NetworkData::Local>().RemoveOnMeshPrefix(mDomainPrefixConfig.mPrefix.mPrefix.mFields.m8,
```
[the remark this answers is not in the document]
librasungirl: good catch. will update.

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +241,92 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+                                                     mDomainPrefixConfig.mPrefix.mLength);
+    }
+
+    mDomainPrefixConfig = aConfig;
```
simonlingoogle, line 302:
```
> prefix
prefix
2001:dead:beef:cafe::/64 paros med
fd00:7d03:7d03:7d03::/64 prosD med
------------------------------------
fd00:7d03:7d03:7d03::/64 prosD med [ot::BackboneRouter::Local]
Done
> ipaddr mleid
ipaddr mleid
fdde:ad00:beef:0:63b4:ac98:130:649c
Done
> prefix add fdde:ad00:beef:0::/64 prosD med
prefix add fdde:ad00:beef:0::/64 prosD med
Error 7: InvalidArgs
> prefix
prefix
2001:dead:beef:cafe::/64 paros med
------------------------------------
fdde:ad00:beef:0::/64 prosD med [ot::BackboneRouter::Local]
Done
```

`prefix add <mleid> prosD med` can remove the domain prefix in local data and overwrite `mDomainPrefixConfig`, but eventually failed to add to local data. Should never allow using `mleid` as domain prefix?

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +241,92 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+
+    if (IsEnabled())
+    {
+        SuccessOrExit(error = Get<NetworkData::Local>().AddOnMeshPrefix(mDomainPrefixConfig));
```
simonlingoogle, line 306: Any `aConfig` that fail in this step will leave `mDomainPrefixConfig` in an inconsistent state because `mDomainPrefixConfig.mPrefix.mLength > 0` but it's not in local data, like the MLEID `fdde:ad00:beef:0::/64` above.

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +241,92 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+
+        Get<NetworkData::Notifier>().HandleServerDataUpdated();
```
simonlingoogle: Should be called as long as one of `Get<NetworkData::Local>().RemoveOnMeshPrefix` and `Get<NetworkData::Local>().AddOnMeshPrefix` is successful?

src/core/backbone_router/local.cpp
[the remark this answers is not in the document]
librasungirl: good catch!

src/core/thread/network_data_local.cpp
```diff
@@ -54,9 +54,54 @@ Local::Local(Instance &aInstance)
...
-otError Local::AddOnMeshPrefix(const uint8_t *aPrefix, uint8_t aPrefixLength, int8_t aPrf, uint8_t aFlags, bool aStable)
+otError Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
 {
-    return AddPrefix(aPrefix, aPrefixLength, NetworkDataTlv::kTypeBorderRouter, aPrf, aFlags, aStable);
+    uint16_t flags = 0;
+
+    if (aConfig.mPreferred)
+    {
+        flags |= BorderRouterEntry::kPreferredFlag;
+    }
```
simonlingoogle, line 64: nit: can we increase code coverage using code below?
```suggestion
    flags |= aConfig.mPreferred ? BorderRouterEntry::kPreferredFlag : 0;
```

src/core/api/border_router_api.cpp
```diff
@@ -54,52 +54,47 @@ otError otBorderRouterGetNetData(otInstance *aInstance, bool aStable, uint8_t *a
...
 otError otBorderRouterRemoveOnMeshPrefix(otInstance *aInstance, const otIp6Prefix *aPrefix)
 {
+    otError   error    = OT_ERROR_NONE;
     Instance &instance = *static_cast<Instance *>(aInstance);
 
     OT_ASSERT(aPrefix != NULL);
 
-    return instance.Get<NetworkData::Local>().RemoveOnMeshPrefix(aPrefix->mPrefix.mFields.m8, aPrefix->mLength);
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+    VerifyOrExit((error = instance.Get<BackboneRouter::Local>().RemoveDomainPrefix(*aPrefix)) != OT_ERROR_NONE);
```
jwhui, line 88: nit: how about using `if-else` construction to avoid having needing an additional #if statement below.

src/core/thread/network_data_tlvs.hpp
```diff
@@ -644,8 +598,7 @@ class BorderRouterEntry
 
 private:
     uint16_t mRloc;
-    uint8_t  mFlags;
-    uint8_t  mReserved;
+    uint16_t mFlags;
```
simonlingoogle, line 601: The code above regarding `mFlags` is difficult to reasoning. Should we still use 2 bytes like below?
```suggestion
    uint8_t mFlagsHB;
    uint8_t mFlagsLB;
```

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +251,90 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+    if (IsEnabled())
+    {
+        SuccessOrExit(error = Get<NetworkData::Local>().AddOnMeshPrefix(mDomainPrefixConfig));
+    }
+
+exit:
+    return error;
+}
+
+otError Local::RemoveDomainPrefix(void)
```
jwhui, line 313: I think it would be more clear if we do not overload the method name. Can we have something like:
- `RemoveDomainPrefixFromNetworkData()`
- `RemoveDomainPrefix()`

Something similar with `AddPrefix()`?

src/core/backbone_router/local.cpp
jwhui: Yes, Domain Prefix must not share the same prefix as Mesh Local. Should add a check for this.

simonlingoogle: If so, we should also add check for modifying `MLEID`.

simonlingoogle: Furthermore, is it possible that a `BBR` capable node could join a Thread network with `mleid` equals to its previously configured domain prefix?  Should we clear `mDomainPrefixConfig` when node attaches in such case?

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +251,90 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(mDomainPrefixConfig.mPrefix.mLength > 0, error = OT_ERROR_NOT_FOUND);
+
+    SuccessOrExit(error = Get<NetworkData::Local>().RemoveOnMeshPrefix(mDomainPrefixConfig.mPrefix.mPrefix.mFields.m8,
+                                                                       mDomainPrefixConfig.mPrefix.mLength));
+
+exit:
+    return error;
```
jwhui, line 323: Do we actually use this error? If not, just return void.

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +251,90 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+otError Local::AddDomainPrefixToNetworkData(void)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(mDomainPrefixConfig.mPrefix.mLength > 0, error = OT_ERROR_NOT_FOUND);
+
+    SuccessOrExit(error = Get<NetworkData::Local>().AddOnMeshPrefix(mDomainPrefixConfig));
+
+exit:
+    return error;
```
jwhui, line 335: Do we actually use this error? If not, just return void.

src/core/backbone_router/local.cpp
jwhui: It might be simpler to not allow enabling BBR service if Domain Prefix matches mesh-local prefix.

Note that it's not just the ML-EID, it's anything that matches mesh-local prefix.

simonlingoogle: Thanks for clarification. No enabling BBR service when `ML` is a prefix of domain prefix seems simplest to implement.

src/core/backbone_router/local.cpp
```diff
@@ -170,14 +181,13 @@ otError Local::AddService(bool aForce)
 
 otError Local::RemoveService(void)
 {
-    otError error       = OT_ERROR_NONE;
+    otError error       = OT_ERROR_NOT_FOUND;
```
abtink, line 184: Can we consider not initializing this variable since the first use is actually `SuccessOrExit( error = ...)` which would assign a value to it. In general, I personally prefer not assigning default values when they server a purpose.

My main reason for this is first it shows the intention of developer that it would set by some other method, and if we later on add  another exit path in the code before setting the variable and the new code does not set it, we get a complier warning which would ensure the developer (adding new exit path) consider the variable/error value.

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +251,90 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+
+otError Local::RemoveDomainPrefix(const otIp6Prefix &aPrefix)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(aPrefix.mLength > 0, error = OT_ERROR_INVALID_ARGS);
+
+    VerifyOrExit(mDomainPrefixConfig.mPrefix.mLength == aPrefix.mLength, error = OT_ERROR_NOT_FOUND);
+
+    VerifyOrExit(otIp6PrefixMatch(&mDomainPrefixConfig.mPrefix.mPrefix, &aPrefix.mPrefix) >=
```
abtink, line 274: nit: use `Ip6::Address::PrefixMatch` instead?

src/core/api/border_router_api.cpp
```diff
@@ -54,52 +54,52 @@ otError otBorderRouterGetNetData(otInstance *aInstance, bool aStable, uint8_t *a
...
     OT_ASSERT(aPrefix != NULL);
 
-    return instance.Get<NetworkData::Local>().RemoveOnMeshPrefix(aPrefix->mPrefix.mFields.m8, aPrefix->mLength);
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+    if ((error = instance.Get<BackboneRouter::Local>().RemoveDomainPrefix(*aPrefix)) == OT_ERROR_NONE)
+    {
+        // do nothing
+    }
+    else
+    // Continue to try to remove on mesh prefix if it is not the Domain Prefix.
```
jwhui, line 96:
```suggestion
    error = instance.Get<BackboneRouter::Local>().RemoveDomainPrefix(*aPrefix);

    if (error == OT_ERROR_NOT_FOUND)
```

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +251,77 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+
+    VerifyOrExit(Ip6::Address::PrefixMatch(mDomainPrefixConfig.mPrefix.mPrefix.mFields.m8, aPrefix.mPrefix.mFields.m8,
+                                           BitVectorBytes(aPrefix.mLength)) >= aPrefix.mLength,
+                 error = OT_ERROR_NOT_FOUND);
+
+    mDomainPrefixConfig.mPrefix.mLength = 0;
+
+    if (IsEnabled())
+    {
+        SuccessOrExit(error =
```
jwhui, line 282: Should this be an assert?

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +251,77 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+
+exit:
+    return error;
+}
+
+void Local::SetDomainPrefix(const NetworkData::OnMeshPrefixConfig &aConfig)
+{
+    if (IsEnabled() && mDomainPrefixConfig.mPrefix.mLength > 0)
+    {
+        Get<NetworkData::Local>().RemoveOnMeshPrefix(mDomainPrefixConfig.mPrefix.mPrefix.mFields.m8,
```
jwhui, line 294: Assert on error?

src/core/backbone_router/local.cpp
jwhui: I don't think we can reject an operational dataset - only the Leader can do that.

How about we simply disable BBR service if the new active operational dataset has a conflicting ML prefix?

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +251,75 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+    {
+        Get<NetworkData::Local>().AddOnMeshPrefix(mDomainPrefixConfig);
+    }
+}
+
+void Local::RemoveDomainPrefixFromNetworkData(void)
+{
+    if (mDomainPrefixConfig.mPrefix.mLength > 0)
+    {
+        Get<NetworkData::Local>().RemoveOnMeshPrefix(mDomainPrefixConfig.mPrefix.mPrefix.mFields.m8,
```
jwhui, line 310: Assert on error?

src/core/backbone_router/local.cpp
```diff
@@ -66,14 +67,18 @@ void Local::SetEnabled(bool aEnable)
     if (aEnable)
     {
         SetState(OT_BACKBONE_ROUTER_STATE_SECONDARY);
+        AddDomainPrefixToNetworkData();
```
jwhui, line 70: Looking at this again, should we not enable BBR service if we fail to add necessary items to network data?

src/core/backbone_router/local.cpp
simonlingoogle: I found that it is possible to enable BBR service without a domain prefix. If a BBR service without domain prefix is valid and useful, maybe we should just turnoff domain prefix when it conflicts with ML prefix and leave BBR service on? Seems better to have minimal surprising impact.

src/core/backbone_router/local.cpp
librasungirl: One BBR Service takes 16B, one Domain Prefix TLV takes 20B, when fails to add domain prefix, it might be still possible to add BBR service successfully? Should Multicast across Thread Networks be independent regardless of whether or not there is DUA feature and whether or not DUA is enabled successfully?

In general I don't think there is big chance for the local server data be full,  otherwise, bigger chance for the leader network data to be much more insufficient.

The size limitation of Thread Network Data should be a general issue when there are too much servers - like > 12 (=254/20) different prefixes configuration, which may be assumed to not happen generally?

Thoughts?

src/core/backbone_router/local.cpp
librasungirl: Sorry when I was saying `reject such pending dataset at the first beginning`, I mean rejection by leader, not allowing such dataset change.

I am fine with the simple option to disable BBR service if ML prefix conflicts when applying operational dataset.

src/core/backbone_router/local.cpp
jwhui: How about adding log messages if adding to network data fails?

simonlingoogle: Also, if we not enable BBR service when we failed to add domain prefix to network data, then we are not allowing empty domain prefix.

jwhui: > Also, if we not enable BBR service when we failed to add domain prefix to network data, then we are not allowing empty domain prefix.

@simonlingoogle, I didn't mean any error. If the error is `OT_ERROR_NO_BUFS`, that could be an issue for the leader.

src/core/backbone_router/local.cpp
simonlingoogle: > Sorry when I was saying reject such pending dataset at the first beginning, I mean rejection by leader, not allowing such dataset change.

I think it is not easy to implement this correctly. When we configure active/pending dataset on the Leader, the `SVR_DATA.ntf` message from the commissioner could be already transmitting in the air, but the Leader haven't received it. The Leader would need extra logic to reject this `SVR_DATA.ntf`, and cleanup current Network Data if necessary.

jwhui: We have to assume that the Leader may be a Thread 1.1 device.

librasungirl: Thanks for the valuable discussions here. I would summarize below scenarios if only speaking of possible coincidence:

Note: below wording `on-mesh prefix` indicates non-mesh local prefix for easier description.

1) For any Thread 1.1 or Thread 1.2 leader, would it be allowed to
     a) accept on mesh prefix registration via SRV_DATA.ntf which may conflicts with the one  in pending dataset?
      b) accept pending operation dataset which may have mesh local prefix the same as the other configured on mesh prefix which was not mesh local prefix from some authorized source via MGMT_PENDING_SET ?

2) For any Thread 1.1/1.2 server, when it joins the Thread Network, if the local on-mesh prefix configuration is the same as mesh local prefix of the network it joins, should it somehow notify upper layer and remove or suppress its SRV_DATA.ntf? 

3) Similar as 2) , suppose a Thread 1.1/1.2 servers migrates to a Network of which the pending dataset has the same mesh local prefix as the on-mesh prefix configuration of the server, what's the expected behavior?

Domain Prefix in Thread 1.2  belongs to the on-mesh prefixes, I think we may need to consider the common coincidence issue between on-mesh prefix  and mesh local prefix.

As for whether or not disable BBR Service when domain prefix conflicts with mesh local prefix, I think there is no clear definition in Thread Specification, it could be flexible for the implementation. We may keep BBR service if we think the multicast feature should be supported if possible.

Thoughts? @jwhui @simonlingoogle

jwhui: @librasungirl, thank you for summarizing the questions. I agree that handling all the situations properly is complex. I am fine if OpenThread allows such configuration and effectively says that configuring on-mesh/domain prefix the same as mesh-local prefix is a user/admin configuration problem. If this becomes an issue later, we can address it then.

librasungirl: agree. thanks.

src/core/thread/network_data_local.cpp
simonlingoogle: Fine with `if ...`

src/core/backbone_router/local.cpp
```diff
@@ -275,49 +271,88 @@ otError Local::RemoveDomainPrefix(const otIp6Prefix &aPrefix)
...
-
     if (IsEnabled())
     {
-        SuccessOrExit(error =
-                          Get<NetworkData::Local>().RemoveOnMeshPrefix(aPrefix.mPrefix.mFields.m8, aPrefix.mLength));
+        error = Get<NetworkData::Local>().RemoveOnMeshPrefix(aPrefix.mPrefix.mFields.m8, aPrefix.mLength);
+        LogDomainPrefix("Remove", error);
     }
 
+    mDomainPrefixConfig.mPrefix.mLength = 0;
```
simonlingoogle, line 280: It seems possible that this function can return some error but still have `mDomainPrefixConfig.mPrefix.mLength = 0;
` executed. Normally we expect nothing has changed when a function returns error; Should we just set `error = OT_ERROR_NONE` here? Or should we change this function to return `void` just like `Local::SetDomainPrefix` ?

src/core/backbone_router/local.cpp
```diff
@@ -275,49 +271,88 @@ otError Local::RemoveDomainPrefix(const otIp6Prefix &aPrefix)
...
 
-    mDomainPrefixConfig.mPrefix.mLength = 0;
-
     if (IsEnabled())
     {
-        SuccessOrExit(error =
-                          Get<NetworkData::Local>().RemoveOnMeshPrefix(aPrefix.mPrefix.mFields.m8, aPrefix.mLength));
+        error = Get<NetworkData::Local>().RemoveOnMeshPrefix(aPrefix.mPrefix.mFields.m8, aPrefix.mLength);
+        LogDomainPrefix("Remove", error);
     }
```
simonlingoogle: Just call `RemoveDomainPrefixFromNetworkData`?

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,114 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+    VerifyOrExit(mDomainPrefixConfig.mPrefix.mLength == aPrefix.mLength, error = OT_ERROR_NOT_FOUND);
+
+    VerifyOrExit(Ip6::Address::PrefixMatch(mDomainPrefixConfig.mPrefix.mPrefix.mFields.m8, aPrefix.mPrefix.mFields.m8,
+                                           BitVectorBytes(aPrefix.mLength)) >= aPrefix.mLength,
+                 error = OT_ERROR_NOT_FOUND);
+
+    if (IsEnabled())
+    {
+        error = Get<NetworkData::Local>().RemoveOnMeshPrefix(aPrefix.mPrefix.mFields.m8, aPrefix.mLength);
+        LogDomainPrefix("Remove", error);
```
simonlingoogle, line 277:
```suggestion
        RemoveDomainPrefixFromNetworkData();
```

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,114 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+{
+    otError error; // only used for logging.
+
+    if (IsEnabled() && mDomainPrefixConfig.mPrefix.mLength > 0)
+    {
+        error = Get<NetworkData::Local>().RemoveOnMeshPrefix(mDomainPrefixConfig.mPrefix.mPrefix.mFields.m8,
+                                                             mDomainPrefixConfig.mPrefix.mLength);
+
+        LogDomainPrefix("Remove", error);
+    }
```
simonlingoogle, line 296:
```suggestion
    if (IsEnabled())
    {
        RemoveDomainPrefixFromNetworkData();
    }
```

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,114 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+
+    mDomainPrefixConfig = aConfig;
+    LogDomainPrefix("Set", OT_ERROR_NONE);
+
+    if (IsEnabled())
+    {
+        error = Get<NetworkData::Local>().AddOnMeshPrefix(mDomainPrefixConfig);
+        LogDomainPrefix("Add", error);
```
simonlingoogle, line 304:
```suggestion
        AddDomainPrefixToNetworkData();
```

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,114 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+void Local::RemoveDomainPrefixFromNetworkData(void)
+{
+    otError error = OT_ERROR_NOT_FOUND; // only used for logging.
+
+    if (mDomainPrefixConfig.mPrefix.mLength > 0)
+    {
+        error = Get<NetworkData::Local>().RemoveOnMeshPrefix(mDomainPrefixConfig.mPrefix.mPrefix.mFields.m8,
+
+                                                             mDomainPrefixConfig.mPrefix.mLength);
+        OT_ASSERT(error == OT_ERROR_NONE);
```
simonlingoogle, line 317: I think `OT_ASSERT` here is too strict because we can not guarantee that `AddDomainPrefixToNetworkData` always succeed. `AddDomainPrefixToNetworkData` can fail due to `NetworkDataTlv ` overflow, or even conflicting with `ML prefix`. So it is possible that backbone router is enabled and `mDomainPrefixConfig.mPrefix.mLength > 0`, but domain prefix is not in network data?

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,114 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+    }
+
+    LogDomainPrefix("Add", error);
+}
+
+void Local::LogDomainPrefix(const char *aAction, otError aError)
+{
+    if (mDomainPrefixConfig.mPrefix.mLength > 0)
+    {
+        otLogDebgNetData("prefix: %s/%d",
```
simonlingoogle, line 339: Why is it `DEBG` here but `INFO` in `else` block?

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,114 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+                         (*static_cast<Ip6::Address *>(&mDomainPrefixConfig.mPrefix.mPrefix)).ToString().AsCString(),
+                         mDomainPrefixConfig.mPrefix.mLength);
+    }
+    else
+    {
+        otLogInfoNetData("No available prefix");
+    }
+
+    otLogInfoNetData("%s Domain Prefix: %s%s", aAction, otThreadErrorToString(aError));
```
simonlingoogle, line 348: This function generate 2 lines of log. I think 1 line of log that contains all information is better. Thoughts?

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,114 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+}
+
+void Local::LogBackboneRouterService(const char *aAction, otError aError)
+{
+    otLogDebgNetData("local BBR: seqno (%d), delay (%ds), timeout (%ds)", mSequenceNumber, mReregistrationDelay,
+                     mMlrTimeout);
+    otLogInfoNetData("%s BBR Service: %s", aAction, otThreadErrorToString(aError));
```
simonlingoogle, line 355: Merge to 1 line of log?

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,114 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+
+    if (mDomainPrefixConfig.mPrefix.mLength > 0)
+    {
+        error = Get<NetworkData::Local>().AddOnMeshPrefix(mDomainPrefixConfig);
+    }
+
+    LogDomainPrefix("Add", error);
+}
+
+void Local::LogDomainPrefix(const char *aAction, otError aError)
```
simonlingoogle, line 335: Got compile warnings:
```
warning: unused parameter ‘aAction’ [-Wunused-parameter]
warning: unused parameter ‘aError’ [-Wunused-parameter]
```

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,114 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+    }
+    else
+    {
+        otLogInfoNetData("No available prefix");
+    }
+
+    otLogInfoNetData("%s Domain Prefix: %s%s", aAction, otThreadErrorToString(aError));
+}
+
+void Local::LogBackboneRouterService(const char *aAction, otError aError)
```
simonlingoogle, line 351: Got compile warnings:
```
warning: unused parameter ‘aAction’ [-Wunused-parameter]
warning: unused parameter ‘aError’ [-Wunused-parameter]
```

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,114 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+{
+    otLogDebgNetData("local BBR: seqno (%d), delay (%ds), timeout (%ds)", mSequenceNumber, mReregistrationDelay,
```
simonlingoogle, line 353: Need new region: `otLogDebgBbr` ?

librasungirl: planned to add BBR region log in a separate PR, let's keep this one simple for now?

src/core/backbone_router/local.cpp
librasungirl: yes, should recover error state. 
Different from SetDomainPrefix() before which the prefix has been pre-checked. 
Prefer to return error here.

src/core/backbone_router/local.cpp
librasungirl: hesitated as there are some conditions already met here. but agree it might be simpler to reuse the code. will update.

src/core/backbone_router/local.cpp
librasungirl: there are already some pre-checks when set domain prefix - already checked whether or not conflicts or overflow, so in general, the prefix should always be added successfully when backbone functionality was enabled if there are not many other server configurations before enable.

src/core/backbone_router/local.cpp
librasungirl: I took the detail prefix content as useful information for debugging purpose, and action and result as the INFO level.

@jwhui any guidance on `DEBG` and `INFO` level logging?

src/core/backbone_router/local.cpp
librasungirl: will fix.

src/core/backbone_router/local.cpp
simonlingoogle: I am fine with another PR.

src/core/backbone_router/local.cpp
librasungirl: added the log

src/core/backbone_router/local.cpp
librasungirl: removed the assert

src/core/backbone_router/local.cpp
simonlingoogle: > Different from SetDomainPrefix() before which the prefix has been pre-checked.

No string opinion on return values. But is it better for callee not to assume how it is called by callers?

src/core/backbone_router/local.cpp
jwhui: We log sent/received IPv6 messages using INFO level. I think it's fine to output the domain prefix using INFO level as well.

src/core/backbone_router/local.cpp
jwhui: Agree that we should keep any update to the domain prefix to one log line.

src/core/backbone_router/local.cpp
librasungirl: ok. will update

librasungirl: updated.

src/core/backbone_router/local.cpp
librasungirl: updated.

src/core/backbone_router/local.cpp
librasungirl: updated.

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,100 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+    if (IsEnabled())
+    {
+        AddDomainPrefixToNetworkData();
+    }
+}
+
+void Local::RemoveDomainPrefixFromNetworkData(void)
+{
+    otError error = OT_ERROR_NOT_FOUND; // only used for logging.
+
```
simonlingoogle, line 304:
```suggestion

    VerifyOrExit(IsEnabled(), error = OT_ERROR_INVALID_STATE); 
```
Since `RemoveDomainPrefixFromNetworkData` is always called when enabled, what about move checking `IsEnabled()` here to make calling `RemoveDomainPrefixFromNetworkData()` even simplier? No need to check have `if (IsEnabled()) { ... }`.

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,100 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+                                                             mDomainPrefixConfig.mPrefix.mLength);
+    }
+
+    LogDomainPrefix("Remove", error);
+}
+
+void Local::AddDomainPrefixToNetworkData(void)
+{
+    otError error = OT_ERROR_NOT_FOUND; // only used for logging.
+
```
simonlingoogle, line 318:
```suggestion

    VerifyOrExit(IsEnabled(), error = OT_ERROR_INVALID_STATE);
```
Since `AddDomainPrefixToNetworkData` is always called when enabled, what about move checking `IsEnabled()` here to make calling `AddDomainPrefixToNetworkData()` even simplier? No need to check have `if (IsEnabled()) { ... }`.

src/core/backbone_router/local.cpp
librasungirl: also hesitated here... decided to do as is now with the considerations - 1) In `SetEnabled()`, it is determined for the Add or Remove action, 2) it might be easier to understand the logical when reading the code like `SetDomainPrefix()` .

https://github.com/openthread/openthread/pull/4793/files#diff-f0e917ba333de39b51d15d69eef03a3eR287

prefer a bit to keep it as is.

simonlingoogle: Fine with keeping it as is for better readability.

src/core/backbone_router/local.cpp
```diff
@@ -238,6 +247,100 @@ void Local::UpdateBackboneRouterPrimary(Leader::State aState, const BackboneRout
...
+}
+
+void Local::RemoveDomainPrefixFromNetworkData(void)
+{
+    otError error = OT_ERROR_NOT_FOUND; // only used for logging.
+
+    if (mDomainPrefixConfig.mPrefix.mLength > 0)
+    {
+        error = Get<NetworkData::Local>().RemoveOnMeshPrefix(mDomainPrefixConfig.mPrefix.mPrefix.mFields.m8,
+
```
abtink, line 308: nit: I think there is an extra new line here?

src/core/backbone_router/local.hpp
```diff
@@ -162,9 +163,49 @@ class Local : public InstanceLocator
...
+     * @param[in]  aConfig A reference to the Domain Prefix configuration.
+     *
+     */
+    void SetDomainPrefix(const NetworkData::OnMeshPrefixConfig &aConfig);
+
 private:
     void    SetState(BackboneRouterState aState);
     otError RemoveService(void);
+    void    AddDomainPrefixToNetworkData(void);
+    void    RemoveDomainPrefixFromNetworkData(void);
```
abtink, line 201: suggestion: How about since these are method in `NetworkData::Local` class, we omit the `ToNetworkData`, `FromNetworkData` in the method names and use `AddDomainPrefix`/`RemoveDomainPrefix`?

src/core/thread/network_data_tlvs.hpp
```diff
@@ -505,7 +512,8 @@ class BorderRouterEntry
      */
     void SetPreference(int8_t aPrf)
     {
-        mFlags = (mFlags & ~kPreferenceMask) | ((static_cast<uint8_t>(aPrf) << kPreferenceOffset) & kPreferenceMask);
+        mFlags = HostSwap16((HostSwap16(mFlags) & ~kPreferenceMask) |
+                            ((static_cast<uint16_t>(aPrf) << kPreferenceOffset) & kPreferenceMask));
```
abtink, line 516: nit: use `GetFlags()` here to replace the `(HostSwap16(mFlags) & ~kPreferenceMask)`?

src/core/backbone_router/local.hpp
simonlingoogle: I would prefer the current names which are more descriptive and accurate. Thoughts @librasungirl

jwhui: @abtink, this is the `BackboneRouter::Local` class.

src/core/thread/network_data_tlvs.hpp
librasungirl: updated.

src/core/backbone_router/local.cpp
librasungirl: removed

src/core/backbone_router/local.hpp
librasungirl: Let's keep the existing names for readability as clarified by @simonlingoogle ? @abtink

abtink: SGTM. Let's keep the name. I mistook this file as `NetworkData::Local`... :)

PR #4795, opened 2020-04-06
[bbr] (Un)Subscribe AllDomainBBRs multicast address

librasungirl:
This PR mainly includes
    - cache domain prefix in the Thread Network
    - subscribe/unsubscribe AllDomainBBRs Multicast address according to Backbone functionality and Domain Prefix state
    - update test to cover AllNetworkBBRs and AllDomainBBRs subscription


src/core/backbone_router/leader.cpp
```diff
@@ -186,6 +235,64 @@ void Leader::Update(void)
...
+    else if (config.mPrefix.mLength == mDomainPrefix.mLength &&
+             otIp6PrefixMatch(&mDomainPrefix.mPrefix, &config.mPrefix.mPrefix) >= mDomainPrefix.mLength)
+    {
+        state = kDomainPrefixUnchanged;
+    }
+    else
+    {
+        mDomainPrefix = config.mPrefix;
+
+        if (mDomainPrefix.mLength == 0)
```
jwhui, line 278: This is effectively the same as `if (config.mPrefix == 0)`. Is this what was intended?

src/core/backbone_router/leader.cpp
```diff
@@ -186,6 +235,64 @@ void Leader::Update(void)
...
+            break;
+        }
+    }
+
+    if (!found)
+    {
+        if (mDomainPrefix.mLength != 0)
+        {
+            // Domain Prefix does not exist any more.
+            memset(&mDomainPrefix, 0, sizeof(mDomainPrefix));
```
jwhui, line 261: Just set `mDomainPrefix.mLength` to zero for simplicity?

src/core/backbone_router/leader.cpp
librasungirl: Good catch!

librasungirl:
The changes to resolve the comments are in the last commit, please review. @jwhui

src/core/net/ip6_address.hpp
```diff
@@ -325,14 +325,25 @@ class Address : public otIp6Address
...
         SetMulticastNetworkPrefix(aMeshLocalPrefix.m8, Mle::MeshLocalPrefix::kLength);
     }
 
+    /**
+     * This method sets the prefix content of Prefix-Based Multicast Address.
+     *
+     * @param[in]  aPrefix A pointer to a IPv6 Prefix.
+     *
+     */
+    void SetMulticastNetworkPrefix(const otIp6Prefix *aPrefix)
```
simonlingoogle, line 342: use alias?
```suggestion
    void SetMulticastNetworkPrefix(const otIp6Prefix &aPrefix)
```

src/core/thread/mle.hpp
```diff
@@ -666,6 +666,17 @@ class Mle : public InstanceLocator
...
+
+    /**
+     * This method returns a reference to the All Domain Backbone Routers Multicast Address.
+     *
+     * @returns A reference to the All Domain Backbone Routers Multicast Address.
+     *
+     */
+    const Ip6::Address &GetAllDomainBackboneRoutersAddress(void) const
+    {
+        return mAllDomainBackboneRouters.GetAddress();
```
simonlingoogle, line 678: `mAllDomainBackboneRouters` is not properly set up when BBR service is not enabled or when there is no domain prefix in network. 
I think it's not good to return a wrong multicast address. Should we return error instead in those cases?

src/core/backbone_router/leader.hpp
```diff
@@ -126,11 +136,37 @@ class Leader : public InstanceLocator
      */
     bool HasPrimary(void) const { return mConfig.mServer16 != Mac::kShortAddrInvalid; }
 
+    /**
+     * This method gets the Domain Prefix in the Thread Network.
+     *
+     * @retval A pointer to the Domain Prefix or NULL if there is no Domain Prefix.
+     *
+     */
+    const otIp6Prefix *GetDomainPrefix(void) const { return (mDomainPrefix.mLength == 0) ? NULL : &mDomainPrefix; }
```
simonlingoogle, line 145: nit: return alias and leave it for the callers to check if the domain prefix is valid?
```suggestion
    const otIp6Prefix &GetDomainPrefix(void) const { return mDomainPrefix; }
```

src/core/thread/mle.hpp
librasungirl: Note that the `AllNetworkBBRs`/`AllDomainBBRs` are only used for BTMF:
- `AllNetworkBBRs` is for `b/bmr` when BBR service is enabled and accepts `n/mr`.
- `AllDomainBBRs` is for `b/bq` when BBR service is enabled, domain prefix is available and accepts `n/dr`.

I would prefer to keep the code simple as is considering the use scenarios in core so far, and open to do possible enhancements if required in the future. 

Thoughts?

simonlingoogle: > `AllDomainBBRs` is for `b/bq` when BBR service is enabled, domain prefix is available and accepts `n/dr`.

Guarantee these conditions by `OT_ASSERT`?

librasungirl: Like said above, bbr may not process n/dr if no domain prefix, it won't use AllDomainBBRs.
Besides, like existing [Mle::GetMeshLocal16()](https://github.com/openthread/openthread/blob/master/src/core/thread/mle.hpp#L744), it simply return the address without assert the attachment status.
I prefer to keep the code simple as is when we know clearly the address is there when it is used.

src/core/net/ip6_address.hpp
simonlingoogle: Looking at `SetMulticastNetworkPrefix` alone, it does not actually accept `aPrefix == NULL`. So, should we prefer reference to pointer in this case?

librasungirl: updated to use reference as the suggestion.

src/core/backbone_router/leader.cpp
```diff
@@ -57,8 +57,11 @@ Leader::Leader(Instance &aInstance)
 
 void Leader::Reset(void)
 {
-    memset(&mConfig, 0, sizeof(mConfig));
+    // Invalid server short address indicates that no available Backbone Router service in the Thread Network.
```
jwhui, line 60:
```suggestion
    // Invalid server short address indicates no available Backbone Router service in the Thread Network.
```

src/core/backbone_router/leader.cpp
```diff
@@ -57,8 +57,11 @@ Leader::Leader(Instance &aInstance)
...
     mConfig.mServer16 = Mac::kShortAddrInvalid;
+
+    //  Domain Prefix Length 0 indicates that no available Domain Prefix in the Thread network.
```
jwhui, line 63:
```suggestion
    //  Domain Prefix Length 0 indicates no available Domain Prefix in the Thread network.
```

src/core/backbone_router/leader.cpp
```diff
@@ -137,7 +149,44 @@ const char *Leader::StateToString(State aState)
...
 }
 
+const char *Leader::StateToString(DomainPrefixState aState)
+{
+    const char *logString = "Unknown";
+
+    switch (aState)
+    {
+    case kDomainPrefixNone:
+        logString = "Domain Prefix: None";
```
jwhui, line 159: In `LogDomainPrefix` above, we already include `Domain Prefix: `. Should we remove redundancy?

src/core/backbone_router/leader.cpp
```diff
@@ -186,6 +235,62 @@ void Leader::Update(void)
...
+            mDomainPrefix.mLength = 0;
+            state                 = kDomainPrefixRemoved;
+        }
+        else
+        {
+            state = kDomainPrefixNone;
+        }
+    }
+    else if (config.mPrefix.mLength == mDomainPrefix.mLength &&
+             otIp6PrefixMatch(&mDomainPrefix.mPrefix, &config.mPrefix.mPrefix) >= mDomainPrefix.mLength)
```
jwhui, line 268: Prefer calling `Ip6::Address::PrefixMatch()` rather than using public API internally.

src/core/backbone_router/leader.hpp
```diff
@@ -68,6 +69,15 @@ class Leader : public InstanceLocator
         kStateUnchanged,      // No change on Primary Backbone Router information (only for logging).
     };
 
+    enum DomainPrefixState
+    {
+        kDomainPrefixNone = 0,  // No Domain Prefix.
```
jwhui, line 74:
```suggestion
        kDomainPrefixNone = 0,  ///< No Domain Prefix.
```
Use doxygen end-of-line comments.

src/core/backbone_router/leader.cpp
librasungirl: Good catch!

src/core/thread/mle.cpp
```diff
@@ -1662,6 +1669,18 @@ void Mle::HandleStateChanged(otChangedFlags aFlags)
         }
     }
 
+    if (aFlags & (OT_CHANGED_THREAD_BACKBONE_ROUTER_STATE | OT_CHANGED_THREAD_DOMAIN_PREFIX))
+    {
+        Get<ThreadNetif>().UnsubscribeMulticast(mAllDomainBackboneRouters);
```
simonlingoogle, line 1674: Can we unexpectedly remove other multicast addresses (e.x. `All Network Backbone Routers Multicast Address`) here, even though the chance is low. 
In the current implementation, `mAllDomainBackboneRouters` will contain the network prefix of last domain prefix when the device had BBR service enabled. When BBR service is turned off, `mAllDomainBackboneRouters` never change. Whenever this code runs, it will try to remove `mAllDomainBackboneRouters` which could equal to other subscribed multicast address. 
For example, if the current mesh local prefix equals to the old domain prefix set in `mAllDomainBackboneRouters` (a totally valid prefix setting), this code will unsubscribe `All Network Backbone Routers Multicast Address`.

I suggest a simple fix that we always set up `mAllDomainBackboneRouters` as long as `Get<BackboneRouter::Leader>().HasDomainPrefix()`.

librasungirl: I don't think it will remove other irrelevant multicast addresses unexpectedly given that:
`mMulticastAddresses` is a linked-list, when [Netif::UnsubscribeMulticast()](https://github.com/openthread/openthread/blob/master/src/core/net/netif.cpp#L321), it finally goes to [LinkedList::Remove(const Type &aEntry)](https://github.com/openthread/openthread/blob/master/src/core/common/linked_list.hpp#L291) which accepts the reference as the formal parameter.

src/core/backbone_router/leader.cpp
```diff
@@ -100,34 +104,42 @@ void Leader::LogBackboneRouterPrimary(State aState, const BackboneRouterConfig &
...
         break;
 
     case kStateRemoved:
-        logString = "PBBR: Removed";
+        logString = "Removed";
         break;
 
     case kStateToTriggerRereg:
-        logString = "PBBR: To trigger re-registration";
+        logString = "Rereg triggered";
```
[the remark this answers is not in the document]
librasungirl: The `logstring` for other state uses the passive tension. I'd prefer a bit to keep the tension in log consistent.

src/core/thread/mle.cpp
simonlingoogle: Won't have this issue if it's comparing by reference. :+1:

src/core/backbone_router/leader.hpp
```diff
@@ -68,6 +69,15 @@ class Leader : public InstanceLocator
...
     };
 
+    enum DomainPrefixState
+    {
+        kDomainPrefixNone = 0,  ///< No Domain Prefix.
+        kDomainPrefixAdded,     // Newly added.
+        kDomainPrefixRemoved,   // Newly removed.
+        kDomainPrefixRefreshed, // Domain Prefix changed.
+        kDomainPrefixUnchanged, // No change on Domain Prefix.
+    };
```
abtink, line 79: Can we define it as `private` in this class?

I think we only use it in this class and for logging purpose. How about calling it `DomainPrefixChange` as it tracks the last change to domain prefix and not a current state.

src/core/backbone_router/leader.cpp
```diff
@@ -186,6 +235,62 @@ void Leader::Update(void)
...
+
+        mDomainPrefix = config.mPrefix;
+    }
+
+    LogDomainPrefix(state, mDomainPrefix);
+
+    if (state == kDomainPrefixAdded || state == kDomainPrefixRefreshed || state == kDomainPrefixRemoved)
+    {
+        Get<Notifier>().Signal(OT_CHANGED_THREAD_DOMAIN_PREFIX);
+    }
```
abtink, line 291: Do we know if the users of OT would need to be notified on the "Domain Prefix" change?

If not,  can we consider not using the `Notifier` for this?  I see that we currently only monitor this "changed flag" from `Mle` (where it does update `mAllDomainBackboneRouters`). We can consider adding a  method in `Mle` that does that and call it directly from here.

Thoughts?

src/core/backbone_router/leader.cpp
abtink: Same suggestion as @jwhui ..

src/core/backbone_router/leader.cpp
```diff
@@ -186,6 +235,62 @@ void Leader::Update(void)
...
+    DomainPrefixState               state;
+    bool                            found = false;
+
+    while (Get<NetworkData::Leader>().GetNextOnMeshPrefix(iterator, config) == OT_ERROR_NONE)
+    {
+        if (config.mDp)
+        {
+            found = true;
+            break;
+        }
```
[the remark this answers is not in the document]
jwhui: Thread 1.2 specification states:

> There SHOULD be at most one prefix configured with P_dp set 'true'; future versions of this specification and special use cases may allow multiple.

src/core/backbone_router/leader.hpp
```diff
@@ -68,6 +69,15 @@ class Leader : public InstanceLocator
         kStateUnchanged,      // No change on Primary Backbone Router information (only for logging).
     };
 
+    enum DomainPrefixState
+    {
+        kDomainPrefixNone = 0,  ///< No Domain Prefix.
+        kDomainPrefixAdded,     // Newly added.
```
jwhui, line 75: nit: be consistent with using `//` or `///<` in end-of-line comments.

include/openthread/instance.h
```diff
@@ -150,6 +150,7 @@ enum
     OT_CHANGED_THREAD_BACKBONE_ROUTER_STATE = 1 << 25, ///< Backbone Router state changed
     OT_CHANGED_THREAD_BACKBONE_ROUTER_LOCAL = 1 << 26, ///< Local Backbone Router configuration changed
     OT_CHANGED_JOINER_STATE                 = 1 << 27, ///< Joiner state changed
+    OT_CHANGED_THREAD_DOMAIN_PREFIX         = 1 << 28, ///< Domain Prefix changed, only available since Thread 1.2
```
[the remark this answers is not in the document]
librasungirl: Thanks for the reminder. Removed and use direct call instead.

src/core/backbone_router/leader.hpp
```diff
@@ -126,11 +136,37 @@ class Leader : public InstanceLocator
...
 private:
+    void UpdateBackboneRouterPrimary(void);
+    void UpdateDomainPrefixConfig(void);
+#if (OPENTHREAD_CONFIG_LOG_LEVEL >= OT_LOG_LEVEL_INFO) && (OPENTHREAD_CONFIG_LOG_NETDATA == 1)
     void               LogBackboneRouterPrimary(State aState, const BackboneRouterConfig &aConfig) const;
+    void               LogDomainPrefix(DomainPrefixState aState, const otIp6Prefix &aPrefix) const;
     static const char *StateToString(State aState);
-
-    BackboneRouterConfig mConfig; ///< Primary Backbone Router information.
+    static const char *StateToString(DomainPrefixState aState);
```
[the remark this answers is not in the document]
librasungirl: make sense. updated as your suggestion.

src/core/backbone_router/leader.hpp
librasungirl: Along with the change from previous notifier way to direct call way, now other components need to access this enumeration.  So keep it public here.

src/core/backbone_router/leader.cpp
librasungirl: Currently not mandatory to use notification way, changed to direct call as your suggestion.
Please take a review.

src/core/backbone_router/leader.cpp
librasungirl: missed previously. updated now.

src/core/backbone_router/leader.hpp
```diff
@@ -56,18 +56,28 @@ typedef otBackboneRouterConfig BackboneRouterConfig;
...
+        kStateAdded,          ///< Newly added.
+        kStateRemoved,        ///< Newly removed (trigger Backbone Router register its service).
+        kStateToTriggerRereg, ///< Short address or sequence number changes (trigger re-registration).
+                              ///< May also have ReregistrationDelay or MlrTimeout update.
+        kStateRefreshed,      ///< Only ReregistrationDelay or MlrTimeout changes.
+        kStateUnchanged,      ///< No change on Primary Backbone Router information (only for logging).
     };
 
+    // Domain Prefix state or state change.
+    typedef enum DomainPrefixState
```
abtink, line 72:
```suggestion
   enum DomainPrefixState
```
I think no need for `typedef` here. `enum DomainPrefixState` would already define the type (since this is C++ only header).

src/core/backbone_router/leader.hpp
```diff
@@ -56,18 +56,28 @@ typedef otBackboneRouterConfig BackboneRouterConfig;
...
+    {
+        kDomainPrefixNone = 0,  ///< Not available.
+        kDomainPrefixAdded,     ///< Added.
+        kDomainPrefixRemoved,   ///< Removed.
+        kDomainPrefixRefreshed, ///< Changed.
+        kDomainPrefixUnchanged, ///< Nothing changed.
+    } DomainPrefixState;
```
abtink, line 79:
```suggestion
    };
```

PR #4797, opened 2020-04-06
[android] export CFLAGS and include directories

gjc13:
This allows libraries depending on ot-core not to specify OpenThread
path and share the OpenThread configs.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -32,8 +32,7 @@ OPENTHREAD_DEFAULT_VERSION := $(shell cat $(LOCAL_PATH)/.default-version)
 OPENTHREAD_SOURCE_VERSION := $(shell git -C $(LOCAL_PATH) describe --always --match "[0-9].*" 2> /dev/null)
 OPENTHREAD_PROJECT_CFLAGS ?= -DOPENTHREAD_PROJECT_CORE_CONFIG_FILE=\"openthread-core-posix-config.h\"
 
-OPENTHREAD_COMMON_FLAGS                                          := \
-    -DMBEDTLS_CONFIG_FILE=\"mbedtls-config.h\"                      \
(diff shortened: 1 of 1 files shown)


Android.mk
```diff
@@ -116,10 +119,24 @@ LOCAL_C_INCLUDES                                         := \
...
 
 LOCAL_CFLAGS                                                                := \
-    $(OPENTHREAD_COMMON_FLAGS)                                                 \
+    $(OPENTHREAD_PUBLIC_CFLAGS)                                                \
+    $(OPENTHREAD_PRIVATE_CFLAGS)                                               \
     $(OPENTHREAD_PROJECT_CFLAGS)                                               \
     $(NULL)
 
+LOCAL_EXPORT_CFLAGS                                        := \
+    -DOPENTHREAD_CONFIG_FILE=\<openthread-config-android.h\>  \
```
bukepo, line 128: Already in public cflags.
```suggestion
```

Android.mk
```diff
@@ -116,10 +119,24 @@ LOCAL_C_INCLUDES                                         := \
...
+    $(OPENTHREAD_PUBLIC_CFLAGS)                               \
+    $(OPENTHREAD_PROJECT_CFLAGS)                              \
+    $(NULL)
+
+LOCAL_EXPORT_C_INCLUDE_DIRS     := \
+    $(OPENTHREAD_PROJECT_INCLUDES) \
+    $(LOCAL_PATH)/include          \
+    $(LOCAL_PATH)/src              \
+    $(LOCAL_PATH)/src/core         \
```
bukepo, line 137: is core headers necessary to be exported?

gjc13: Yes, we need to export `openthread-core-config.h` so that users can check which features are supported. This header is requires core headers.

PR #4800, opened 2020-04-08
[icmp] do not generate errors in response to errors

jwhui:


src/core/thread/mesh_forwarder.cpp
```diff
@@ -1257,23 +1245,54 @@ bool MeshForwarder::UpdateReassemblyList(void)
...
 }
 
-void MeshForwarder::HandleLowpanHC(uint8_t *               aFrame,
+otError MeshForwarder::FrameToMessage(const uint8_t *&    aFrame,
+                                      uint16_t &          aFrameLength,
+                                      uint16_t            aDatagramSize,
+                                      const Mac::Address &aMacSource,
+                                      const Mac::Address &aMacDest,
+                                      Message *&          aMessage)
+{
```
abtink, line 1254: This is indeed a nice common method. 👍 

One question: Do we need the `aFrame` and `aFrameLength` to be reference and updated? Cannot see if we use the frame after any of the calls to this method.

PR #4803, opened 2020-04-09
[mbedtls] leverage cmake provided by mbedtls

jwhui:


examples/platforms/cc1352/CMakeLists.txt
```diff
@@ -75,6 +76,7 @@ target_compile_definitions(openthread-cc1352
         ${OT_PLATFORM_DEFINES}
     PRIVATE
         ${OT_PRIVATE_DEFINES}
+        "MBEDTLS_CONFIG_FILE=\"mbedtls-config.h\""
```
wgtdkp, line 79: Why added this?

third_party/mbedtls/CMakeLists.txt
```diff
@@ -26,73 +26,57 @@
...
 
-add_library(mbedcrypto)
-
-set(OT_MBEDTLS_DEFAULT_CONFIG_FILE "\"mbedtls-config.h\"")
+set(OT_MBEDTLS_DEFAULT_CONFIG_FILE \"${CMAKE_CURRENT_SOURCE_DIR}/mbedtls-config.h\")
 
 set(OT_MBEDTLS_CONFIG_FILE "" CACHE STRING "The mbedTLS config file")
 
-target_compile_definitions(mbedcrypto
+set(ENABLE_TESTING OFF CACHE BOOL "Disable mbedtls test")
```
bukepo, line 33: Not sure if we should add `FORCE`, because I remember sometimes this doesn't take effect without `FORCE`.
```suggestion
set(ENABLE_TESTING OFF CACHE BOOL "Disable mbedtls test" FORCE)
```

src/core/CMakeLists.txt
```diff
@@ -246,11 +246,15 @@ target_sources(openthread-radio PRIVATE
 if(OT_BUILTIN_MBEDTLS)
     target_link_libraries(openthread-ftd
         PUBLIC
```
wgtdkp, line 248: It seems `mbedtls` names are not exposed by the public headers of target `openthread-ftd`, so this may should be `PRIVATE`? Other targets depending on `mbedtls` may explicitly link it by themselves.

src/core/CMakeLists.txt
```diff
@@ -246,11 +246,15 @@ target_sources(openthread-radio PRIVATE
 if(OT_BUILTIN_MBEDTLS)
```
wgtdkp, line 246: whether we use builtin `mbedtls` in third_party or `mbedtls` provided by application, we need the mbedtls targets to compile. So I think this should be removed?

third_party/mbedtls/CMakeLists.txt
```diff
@@ -26,73 +26,57 @@
...
-
-target_include_directories(mbedcrypto
+target_include_directories(mbedtls
     PUBLIC
-        ${PROJECT_SOURCE_DIR}/third_party/mbedtls/repo/include
+        ${CMAKE_CURRENT_SOURCE_DIR}/repo/include
     PRIVATE
         ${OT_PUBLIC_INCLUDES}
         ${OT_PRIVATE_INCLUDES}
+        ${PROJECT_SOURCE_DIR}/src/core
```
wgtdkp, line 51: I think this needs to be `PUBLIC`. Because the `mbedtls` target includes `mbedtls-config.h` in its public headers (header files in `${CMAKE_CURRENT_SOURCE_DIR}/repo/include`), and `mbedtls-config.h` includes a core config file in `${PROJECT_SOURCE_DIR}/src/core`. So targets depending on `mbedtls` will not be able to find the core config file if we make `${PROJECT_SOURCE_DIR}/src/core` PRIVATE.

even worse, `${OT_PRIVATE_DEFINES}` below also needs to be PUBLIC. Otherwise, targets other than `openthread-ftd` and `openthread-mtd` will see different mbedtls header files.

To avoid exposing `${PROJECT_SOURCE_DIR}/src/core` and `${OT_PRIVATE_DEFINES}`, one option is to provide mbedtls macros by `target_compile_definitions(...)` rather than `MBEDTLS_CONFIG_FILE`. thoughts?

src/core/CMakeLists.txt
jwhui: This was added with #4696 to support linking with external mbedTLS. Are you sure we want to revert this?

examples/platforms/cc1352/CMakeLists.txt
jwhui: MBEDTLS_CONFIG_FILE needs to point to the proper config header when building platform-specific crypto drivers. Suggestions on how best to support this?

src/core/CMakeLists.txt
jwhui: We can move mbed libs private here. CLI and NCP target will also need to explicitly list mbed libs as a result.

src/core/CMakeLists.txt
wgtdkp: @rlubos Are you using external mbedtls but not using the canonical mbedtls targets name (`mbedtls`, `mbedcrypto` and `mbedx509`)?

examples/platforms/cc1352/CMakeLists.txt
wgtdkp: It seems the `openthread-cc1352` target needs to include mbedtls headers, then we should expect that 
```cmake
target_link_libraries(openthread-cc1352
    PRIVATE mbedtls
)
```
should do all we need, and we can also remove `${OT_PRIVATE_DEFINES}` here (this seems only needed by mbedTLS headers), if we make `${OT_PRIVATE_DEFINES}` PUBLIC definitions of mbedTLS targets.

src/core/CMakeLists.txt
wgtdkp: Yes, I think this is the way to do it. Ideally, they should just need to do:
```cmake
target_link_libraries(xxx-cli PRIVATE mbedtls ... )
```

third_party/mbedtls/CMakeLists.txt
jwhui: @wgtdkp, thoughts?

third_party/mbedtls/CMakeLists.txt
jwhui: This seems to work as is. Note that `MBEDTLS_CONFIG_FILE` is PUBLIC with absolute path. Thoughts?

third_party/mbedtls/CMakeLists.txt
wgtdkp: `ENABLE_TESTING` is a global CMake VAR. If it is set by other modules before, then the value will not be changed  here. I think if we want to force disable mbedtls tests, we need 
 `FORCE` (also need to be fixed in OT Commissioner).

For better ways to excludes unwanted targets in submodules, we can add submodules with
```cmake
add_subdirectories(repo EXCLUDE_FROM_ALL)
```
This will not build any targets in `repo` but only those linked or explicitly specified by `ninja <target-name>`.

third_party/mbedtls/CMakeLists.txt
wgtdkp: Yes,  `MBEDTLS_CONFIG_FILE` has absolute path, so we don't need to include `${CMAKE_CURRENT_SOURCE_DIR}` to find `mbedtls-config.h`. But `mbedtls-config.h` is not self-contained, because it includes `penthread-core-config.h`.

It works now, because we manually add the required definitions to all targets (e.g. openthread-ftd) that depends on the mbedTLS targets. But I don't think this is how modern CMake should work. Ideally, I would expect any target `A` depending on target  `B` just need to do `target_link_libraries(A PRIVATE B)` and inherit include directories and definitions, but don't need to manually add them. thoughts?

jwhui: The dependency on `openthread-core-config.h` generally applies to all targets in the OpenThread project. It's not clear to me that the medtls target should be the one responsible for exporting where `openthread-core-config.h` is located.

wgtdkp: Agree that mbedTLS targets should not export the location of `openthread-core-config.h`. But we are exporting `MBEDTLS_CONFIG_FILE ` and it includes `openthread-core-config.h`, which force we to export location of `openthread-core-config.h` to make it work with modern cmake.

I think currently it is ok. But make openthread works with modern cmake and easier to be used by applications, we need to fix this in the future.

jwhui: OK, if I understand correctly, the real problem is the inclusion of `openthread-core-config.h` in `mbedtls-config.h` and you'd like to see that removed. That makes sense to me. At the same time, `openthread-core-config.h` is what determines what features are needed by mbedTLS. Given that this is an existing issue not related to this PR, I'd prefer to address this separately.

src/core/CMakeLists.txt
jwhui: @wgtdkp, just wanted to make sure you saw the comment from @rlubos in https://github.com/openthread/openthread/pull/4696#issuecomment-600489939.

rlubos: Zephyr specifies it's own CMakeLists.txt for mbedTLS and the library is called `mbedTLS`.

But, as I wrote in the comment linked, if OpenThread prefers to link mbedTLS here unconditionally, I can work it around in Zephyr (or convince Zephyr community to switch to canonical name).

third_party/mbedtls/CMakeLists.txt
wgtdkp: Agree.

src/core/CMakeLists.txt
rlubos: I can see that other OT libraries (ncp, cli)  have also gained `mbedtls` target dependency in this commit, so it would likely make sense to give up on this condition for consistency.

jwhui: @rlubos, thanks for your input! I have removed the condition in this PR.

wgtdkp: @jwhui Thanks for the link, I was careless. @rlubos Now I understand the situation.

I think, leaving openthread linking against nothing doesn't make sense. I believe you need to manually write:
```cmake
target_link_libraries(openthread-ftd PRIVATE mbedTLS)
```
to compile openthread libraries.

I have some options (not necessary be addressed in this PR):
1. Openthread always use canonical mbedTLS target names, and for your case, you can add interface libraries or library ALIAS to adapt the target name `mbedTLS` to one of the caonical mbedtls target names.
2. replace `OT_BUILTIN_MBEDTLS` with `OT_EXTERNAL_MBEDTLS` which gives the external mbedtls library name, and link against `OT_EXTERNAL_MBEDTLS` if it is provided.

thoughts?

rlubos: Thanks for the feedback @wgtdkp, 1 sounds like a viable solution to me, I'll try to add an alias name for the mbedTLS during the next OT update in Zephyr. 

In case I meet some serious blocker, I'll come back to you guys, we might consider the second solution then (which also sounds fine, but that would add some extra complexity one the OpenThread side which might not necessarily be needed).

wgtdkp: @rlubos Thanks, please don't hesitate to reach out if you have any issue with it.

Issue #4770, opened 2020-04-01
SOCK_NONBLOCK (passed to SocketWithCloseExec in posix netif) is Linux-specific


rob-the-dude:
**Is your feature request related to a problem? Please describe.**

SOCK_NONBLOCK is a Linux-specific feature.  Some BSDs, including macOS, don't have it.

**Describe the solution you'd like**

It looks like the posix code is trying to wrap the socket calls with the SocketWithCloseExec API (likely because SOCK_CLOEXEC is also non-portable).  Perhaps a new bool parameter indicating that the socket should be marked non-blocking?

**Describe alternatives you've considered**
I can also #ifdef around the issue and make the socket non-blocking in the code, but it just looks uglier.

gjc13:
Would adding `fcntl(fd, O_NONBLOCK)` suffice for  BSD?

rob-the-dude:
Yes, that's what is needed.

However, I was looking at this again.  Why do we mark it non-blocking?  We never call "read/recv" on it unless "select" has indicated it's ready for reading.  And we never write to the socket, so we will never stall for such a case.

We don't mark the other sockets/descriptors as non-blocking.  In particular, the AF_NETLINK socket (sNetlinkFd) is used almost exactly the same way as the MLD socket (we poll for readability on select, and then read).  Why is one non-blocking and the other blocking?  The IP ioctl socket (sIpFd) is left in blocking mode, but all we use it for is ioctl, so that seems right.  And the tunnel descriptor (sTunFd) is also in blocking mode -- that may be intended (the main consequence is that "processReceive" may block in "write" while sending packets to the kernel, though in practice that's likely extremely unlikely -- we do not handle the EWOULDBLOCK error (or any error) from "write" particularly gracefully, but we don't fail -- the packet is simply lost).

Personally, it seems the easiest thing to do may simply be to remove SOCK_NONBLOCK and don't worry about.  I don't think the socket needs to be non-blocking for the use case we have here. But I'm willing to be to be educated why that's wrong, if I can also be educated why the other sockets are properly blocking.

gjc13:
https://stackoverflow.com/questions/5351994/will-read-ever-block-after-select

Chances are that the read after select will block so we add NONBLOCK option to be safe.

gjc13:
I found some of the fds in netif.cpp not opened with NONBLOCK option. I'll also update these fds.

PR #4804, opened 2020-04-09
[posix] open netif fds with O_NONBLOCK in a compatible way

gjc13:
Resolves https://github.com/openthread/openthread/issues/4770

### DIFF
--- a/src/posix/platform/misc.cpp
+++ b/src/posix/platform/misc.cpp
@@ -83,7 +83,7 @@ otPlatMcuPowerState otPlatGetMcuPowerState(otInstance *aInstance)
     return gPlatMcuPowerState;
 }
 
-int SocketWithCloseExec(int aDomain, int aType, int aProtocol, bool aNonBlock)
+int SocketWithCloseExec(int aDomain, int aType, int aProtocol, SocketBlockOption aBlockOption)
 {
     int rval = 0;
     int fd   = -1;
@@ -92,10 +92,10 @@ int SocketWithCloseExec(int aDomain, int aType, int aProtocol, bool aNonBlock)
     VerifyOrExit((fd = socket(aDomain, aType, aProtocol)) != -1, perror("socket(SOCK_CLOEXEC)"));
 
     VerifyOrExit((rval = fcntl(fd, F_GETFD, 0)) != -1, perror("fcntl(F_GETFD)"));
-    rval |= aNonBlock ? O_NONBLOCK | FD_CLOEXEC : FD_CLOEXEC;
+    rval |= aBlockOption == kSocketNonBlock ? O_NONBLOCK | FD_CLOEXEC : FD_CLOEXEC;
     VerifyOrExit((rval = fcntl(fd, F_SETFD, rval)) != -1, perror("fcntl(F_SETFD)"));
 #else
-    aType |= aNonBlock ? SOCK_CLOEXEC | SOCK_NONBLOCK : SOCK_CLOEXEC;
+    aType |= aBlockOption == kSocketNonBlock ? SOCK_CLOEXEC | SOCK_NONBLOCK : SOCK_CLOEXEC;
     VerifyOrExit((fd = socket(aDomain, aType, aProtocol)) != -1, perror("socket(SOCK_CLOEXEC)"));
 #endif
 
--- a/src/posix/platform/misc.cpp
+++ b/src/posix/platform/misc.cpp
@@ -83,7 +83,7 @@ otPlatMcuPowerState otPlatGetMcuPowerState(otInstance *aInstance)
     return gPlatMcuPowerState;
 }
 
-int SocketWithCloseExec(int aDomain, int aType, int aProtocol)
+int SocketWithCloseExec(int aDomain, int aType, int aProtocol, bool aNonBlock)
 {
     int rval = 0;
     int fd   = -1;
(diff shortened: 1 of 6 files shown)


rob-the-dude:
I understand that the tunnel descriptor is not a socket, but should we open it in non-blocking mode as well?  It's opened in platformNetifInit -- I think you could add " | O_NONBLOCK" to the open call to get the desired behavior.

gjc13:
> I understand that the tunnel descriptor is not a socket, but should we open it in non-blocking mode as well? It's opened in platformNetifInit -- I think you could add " | O_NONBLOCK" to the open call to get the desired behavior.

Yes, see https://stackoverflow.com/questions/5351994/will-read-ever-block-after-select for why we need to specify O_NONBLOCK even if we are using it with select.

gjc13:
@bukepo Looks like virtual time doesn't work well with O_NONBLOCK so I still need to leave it in blocking mode.

src/posix/platform/misc.cpp
```diff
@@ -107,6 +107,31 @@ int SocketWithCloseExec(int aDomain, int aType, int aProtocol)
...
+
+    VerifyOrExit((rval = fcntl(fd, F_GETFD, 0)) != -1, perror("fcntl(F_GETFD)"));
+    VerifyOrExit((rval = fcntl(fd, F_SETFD, rval | FD_CLOEXEC | O_NONBLOCK)) != -1, perror("fcntl(F_SETFD)"));
+#else
+    VerifyOrExit((fd = socket(aDomain, aType | SOCK_CLOEXEC | SOCK_NONBLOCK, aProtocol)) != -1,
+                 perror("socket(SOCK_CLOEXEC)"));
+#endif
+
+exit:
+    if (rval == -1 && fd != -1)
```
simonlingoogle, line 126: nit: `fd` must not be -1 if `rval == -1` ?
```suggestion
    if (rval == -1)
```

src/posix/platform/platform-posix.h
```diff
@@ -442,6 +442,20 @@ void platformUdpUpdateFdSet(otInstance *aInstance, fd_set *aReadFdSet, int *aMax
...
+ * @param[in]   aDomain     The communication domain.
+ * @param[in]   aType       The semantics of communication.
+ * @param[in]   aProtocol   The protocol to use.
+ *
+ * @returns The file descriptor of the created socket.
+ *
+ * @retval  -1  Failed to create socket.
+ *
+ */
+int NonBlockSocketWithCloseExec(int aDomain, int aType, int aProtocol);
```
bukepo, line 457: Is it possible to merge into SocketWithCloseExec() by adding an additional argument for nonblocking?

src/posix/platform/misc.cpp
```diff
@@ -92,13 +92,21 @@ int SocketWithCloseExec(int aDomain, int aType, int aProtocol)
...
+    if (aNonBlock)
+    {
+        rval |= O_NONBLOCK;
+    }
+    VerifyOrExit((rval = fcntl(fd, F_SETFD, rval)) != -1, perror("fcntl(F_SETFD)"));
 #else
-    VerifyOrExit((fd = socket(aDomain, aType | SOCK_CLOEXEC, aProtocol)) != -1, perror("socket(SOCK_CLOEXEC)"));
+    int type;
+
+    type = aNonBlock ? aType | SOCK_CLOEXEC | SOCK_NONBLOCK : aType | SOCK_CLOEXEC;
```
bukepo, line 104: use `aType` instead of defining a new variable?

src/posix/platform/misc.cpp
```diff
@@ -92,13 +92,19 @@ int SocketWithCloseExec(int aDomain, int aType, int aProtocol)
     VerifyOrExit((fd = socket(aDomain, aType, aProtocol)) != -1, perror("socket(SOCK_CLOEXEC)"));
 
     VerifyOrExit((rval = fcntl(fd, F_GETFD, 0)) != -1, perror("fcntl(F_GETFD)"));
-    VerifyOrExit((rval = fcntl(fd, F_SETFD, rval | FD_CLOEXEC)) != -1, perror("fcntl(F_SETFD)"));
+    rval |= FD_CLOEXEC;
+    if (aNonBlock)
+    {
+        rval |= O_NONBLOCK;
+    }
```
simonlingoogle, line 99:
```suggestion
    rval |= aNonBlock ? O_NONBLOCK | FD_CLOEXEC : FD_CLOEXEC;
```
Why are we using `?:` operator in `#else` block but not here?

src/posix/platform/misc.cpp
```diff
@@ -92,13 +92,19 @@ int SocketWithCloseExec(int aDomain, int aType, int aProtocol)
...
+    VerifyOrExit((rval = fcntl(fd, F_SETFD, rval)) != -1, perror("fcntl(F_SETFD)"));
 #else
-    VerifyOrExit((fd = socket(aDomain, aType | SOCK_CLOEXEC, aProtocol)) != -1, perror("socket(SOCK_CLOEXEC)"));
+    aType = aNonBlock ? aType | SOCK_CLOEXEC | SOCK_NONBLOCK : aType | SOCK_CLOEXEC;
```
simonlingoogle, line 102:
```suggestion
    aType |= aNonBlock ? SOCK_CLOEXEC | SOCK_NONBLOCK : SOCK_CLOEXEC;
```

src/posix/platform/platform-posix.h
```diff
@@ -434,13 +434,14 @@ void platformUdpUpdateFdSet(otInstance *aInstance, fd_set *aReadFdSet, int *aMax
...
  * @param[in]   aProtocol   The protocol to use.
+ * @param[in]   aNonBlock   Whether to add nonblock flags.
  *
  * @returns The file descriptor of the created socket.
  *
  * @retval  -1  Failed to create socket.
  *
  */
-int SocketWithCloseExec(int aDomain, int aType, int aProtocol);
+int SocketWithCloseExec(int aDomain, int aType, int aProtocol, bool aNonBlock);
```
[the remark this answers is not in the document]
gjc13: Not sure whether it's a good idea @bukepo @jwhui . Is it common practice to add such definitions in OpenThread?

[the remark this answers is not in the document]
jwhui: Named values are generally preferred for the reasons that @rob-the-dude mentioned. Some examples in existing code:

https://github.com/openthread/openthread/blob/af4837d122c9da0eb6dcb9f2a1b19b1e37b738cc/src/core/mac/data_poll_sender.hpp#L274-L278

https://github.com/openthread/openthread/blob/af4837d122c9da0eb6dcb9f2a1b19b1e37b738cc/src/core/mac/mac_types.hpp#L111-L115

https://github.com/openthread/openthread/blob/af4837d122c9da0eb6dcb9f2a1b19b1e37b738cc/src/core/thread/mle.hpp#L1049-L1053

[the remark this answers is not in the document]
gjc13: Done

PR #4805, opened 2020-04-09
[network-data-notifier] fix flag for child removed notification

piotrkoziar:
This PR fixes IAR compilation error:
Error[Pa089]: enumerated type mixed with another enumerated type

### DIFF
--- a/src/core/thread/network_data_notifier.cpp
+++ b/src/core/thread/network_data_notifier.cpp
@@ -108,12 +108,14 @@ void Notifier::HandleStateChanged(ot::Notifier::Callback &aCallback, otChangedFl
 
 void Notifier::HandleStateChanged(otChangedFlags aFlags)
 {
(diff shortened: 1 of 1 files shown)


piotrkoziar:
@LuDuda thank you, I completely missed that!

src/core/thread/network_data_notifier.cpp
```diff
@@ -108,12 +108,14 @@ void Notifier::HandleStateChanged(ot::Notifier::Callback &aCallback, otChangedFl
 
 void Notifier::HandleStateChanged(otChangedFlags aFlags)
 {
-    if (aFlags & (OT_CHANGED_THREAD_ROLE | OT_NEIGHBOR_TABLE_EVENT_CHILD_REMOVED))
+    if (aFlags &
+        (static_cast<uint32_t>(OT_CHANGED_THREAD_ROLE) | static_cast<uint32_t>(OT_CHANGED_THREAD_CHILD_REMOVED)))
```
LuDuda, line 112: I don't think we need to cast it anymore.

Issue #4796, opened 2020-04-06
[build] Unaligned load/store compiler warning


dismirlian:
Hi all, GCC/Clang provide the -Wcast-align warning switch, which tells the compiler to emit warnings whenever a pointer is cast such that the required alignment of the target is increased. For instance, `*reinterpret_cast<uint32_t *>(GetEnterpriseNumberLocation())` would emit a warning.

This issue is related to #4617.

In the former example, the compiler could emit instructions which access unaligned memory. Cortex M4 mostly supports unaligned accesses, but store/load double/multiple instructions (LDM, STM, LDRD and STRD) will always generate a usage fault. In general, I guess that the compiler doesn't generate these instructions frequently, so the code works. But I think we should enable the warning and fix the potential issues because the code would break with a new compiler version or different optimization levels.

I enabled -Wcast-align, and I found these instances of possible problems in (core) code.

In `heap.hpp`:
https://github.com/openthread/openthread/blob/c6a258e3a5bd90aa26d8245c435c0ae5797027f4/src/core/utils/heap.hpp#L94-L97

https://github.com/openthread/openthread/blob/c6a258e3a5bd90aa26d8245c435c0ae5797027f4/src/core/utils/heap.hpp#L107-L110

These are not really issues, because `CAlloc` always aligns `size`. In this case, there are two ways to tell the compiler that this isn't a problem:
1) Use GCC/Clang `__builtin_assume_aligned`:
```
uint16_t GetNext(void) const
{
    const uint8_t *ptr = reinterpret_cast<const uint8_t *>(this) + sizeof(mSize) + mSize;
    return *reinterpret_cast<const uint16_t *>(__builtin_assume_aligned(ptr, 2));
}
```
(of course, wrapped in a `OT_TOOL...` macro). This alternative doesn't have any downside, except possibly for compiler support.

2) Something like this:
```
uint16_t GetNext(void) const
{
    const uint8_t *ptr = ;
    return *(reinterpret_cast<const uint16_t *>(this) + sizeof(mSize) / 2 + mSize / 2);
}
```
This alternative increases the code size by 16 bytes on my test platform.

In `network_data_tlvs.hpp`:

https://github.com/openthread/openthread/blob/c6a258e3a5bd90aa26d8245c435c0ae5797027f4/src/core/thread/network_data_tlvs.hpp#L927-L931

[...]

PR #4806, opened 2020-04-09
[network-data] simplify ServiceTlv

abtink:
This commit updates the `ServiceTlv` implementation and its use.

- Set the `ServiceTlv` fields through a new `Init()`
- Use union for managing presence of Enterprise Number field
- Add static helper `GetSize()` to get the size of a Service TLV
  with a given Enterprise number and Service Data length.
- Add `ServiceTlv::kThreadEnterpriseNumber` constant replacing the
  `THREAD_ENTERPRISE_NUMBER` definition.

### DIFF
--- a/src/core/thread/network_data_tlvs.hpp
+++ b/src/core/thread/network_data_tlvs.hpp
@@ -36,15 +36,17 @@
 
 #include "openthread-core-config.h"
 
+#include <openthread/netdata.h>
+
+#include "common/debug.hpp"
 #include "common/encoding.hpp"
 #include "net/ip6_address.hpp"
-
-#define THREAD_ENTERPRISE_NUMBER 44970
 
 namespace ot {
 namespace NetworkData {
 
 using ot::Encoding::BigEndian::HostSwap16;
+using ot::Encoding::BigEndian::HostSwap32;
 
 /**
  * @addtogroup core-netdata-tlvs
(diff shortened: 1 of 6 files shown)


src/core/thread/network_data_tlvs.hpp
```diff
@@ -1000,148 +1020,62 @@ class ServiceTlv : public NetworkDataTlv
...
-        if (IsThreadEnterprise())
-        {
-            return THREAD_ENTERPRISE_NUMBER;
-        }
-        else
-        {
-            // This memory access most likely will not be aligned to 4 bytes
-            return HostSwap32(*reinterpret_cast<const uint32_t *>(GetEnterpriseNumberLocation()));
-        }
+        return IsThreadEnterprise() ? static_cast<uint32_t>(kThreadEnterpriseNumber) : HostSwap32(mEnterpriseNumber);
```
jwhui, line 1045: @dismirlian, I believe this addresses one of the items called out in #4796 .

src/core/thread/network_data_tlvs.hpp
```diff
@@ -1000,148 +1020,62 @@ class ServiceTlv : public NetworkDataTlv
...
-            mTResSId |= kTMask;
-        }
-        else
-        {
-            mTResSId &= ~kTMask;
-
-            // This memory access most likely will not be aligned to 4 bytes
-            *reinterpret_cast<uint32_t *>(GetEnterpriseNumberLocation()) = HostSwap32(aEnterpriseNumber);
-        }
+        return IsThreadEnterprise() ? mServiceDataLengthThreadEnterprise : mServiceDataLength;
```
jwhui, line 1056: @dismirlian, I believe this addresses one of the items called out in #4796 .

src/core/thread/network_data_tlvs.hpp
abtink: Yes. The new model should remove the need for casting the pointer and possible un-aligned access. I personally don't like the pointer casting approach for reading/writing any integer (other then `uint8_t`). The behavior is tool-chain dependent.

We have methods in `Encoding` like `ReadUint32`, `WriteUint32`, etc (for both little/big endian encodings) that do the same and should be safer and work with any pointer value (independent of its alignment).

dismirlian: @jwhui yes it is one of those instances. I agree with both of you that `ReadUint32` (and similar) are the best way forward.  @abtink's refactoring gets rid of the possible unaligned access, so I'll hold my PR until this is merged. Thanks!

src/core/thread/network_data_tlvs.hpp
```diff
@@ -1188,33 +1110,59 @@ class ServiceTlv : public NetworkDataTlv
...
+        kMinLength            = sizeof(uint8_t) + sizeof(uint8_t), // Flags and Service Data length.
+    };
+
+    // When `kThreadEnterpriseFlag is set in the `mFlagsServiceId`, the
+    // `mEnterpriseNumber` field is elided and `mFlagsServiceId` is
+    // immediately followed by the Service Data length field (which is
+    // represented by `mServiceDataLengthThreadEnterprise`)
+
+    uint8_t mFlagsServiceId;
+    union OT_TOOL_PACKED_FIELD
```
jwhui, line 1159: Avoid anonymous unions.

abtink: Changed in new pushed commit.

PR #4808, opened 2020-04-09
[compiler] avoid use of compiler extensions

jwhui:
- avoid use of GNU extension `##__VA_ARGS__`
- avoid use of zero variadic macro arguments
- avoid use of flexible array members
- enforce above by adding pedantic-errors flag to cmake build files
- use C++11 in cmake only because code still includes some C++11 features

### DIFF
--- a/src/core/common/code_utils.hpp
+++ b/src/core/common/code_utils.hpp
@@ -99,21 +99,27 @@
     } while (false)
 
 /**
+ * Use this macro in conjunction with `VerifyOrExit()` when no action is specified.
+ *
+ */
+#define OT_NOOP
+
+/**
  * This macro checks for the specified condition, which is expected to commonly be true, and both executes @a ... and
  * branches to the local label 'exit' if the condition is false.
  *
  * @param[in]  aCondition  A Boolean expression to be evaluated.
- * @param[in]  ...         An expression or block to execute when the assertion fails.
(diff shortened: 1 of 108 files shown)


src/core/common/code_utils.hpp
```diff
@@ -98,22 +98,28 @@
         }                      \
     } while (false)
 
+/**
+ * Use this macro in conjunction with `VerifyOrExit()` when no action is specified.
+ *
+ */
+#define OT_NOOP
```
simonlingoogle, line 105:
```suggestion
#define Nop
```
`OT_NOOP` is always used together with `VerifyOrExit`. So, should it follow the same naming convention of `VerifyOrExit`?
Besides, `nop` is a common assembly instruction and should be familiar with embedded developers. 
FYI: https://en.wikipedia.org/wiki/NOP_(code)

jwhui: The wiki article shows "NOP", "no-op", and "NOOP" as all valid abbreviations. At the same time, "OT_NOOP" is not the same as it does not compile to anything.

As for uppercase vs camel-case, the OpenThread project generally uses all upper case for macros with no arguments.

I originally had this as "OT_NO_ACTION". Would you prefer that?

simonlingoogle: I would prefer `OT_NOOP` to `OT_NO_ACTION`.

PR #4811, opened 2020-04-10
[address-resolver] use RLOC dest when send `a/ae` to child

simonlingoogle:
Besides, we found that the `/a/ae` COAP message with link-local destination wouldn't be processed by by child's COAP layer. 
Thoughts on why it is unreachable by link-local address? @librasungirl @jwhui

### DIFF
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -1108,23 +1108,17 @@ exit:
     return error;
 }
 
-otError Mle::GetAlocAddress(Ip6::Address &aAddress, uint16_t aAloc16) const
+otError Mle::GetLocatorAddress(Ip6::Address &aAddress, uint16_t aLoc16) const
 {
     otError error = OT_ERROR_NONE;
 
     VerifyOrExit(GetRloc16() != Mac::kShortAddrInvalid, error = OT_ERROR_DETACHED);
 
     memcpy(&aAddress, &mMeshLocal16.GetAddress(), 14);
-    aAddress.SetLocator(aAloc16);
+    aAddress.SetLocator(aLoc16);
 
 exit:
     return error;
-}
-
-otError Mle::GetRlocAddress(Ip6::Address &aAddress, uint16_t aRloc16) const
-{
-    // Same logic with `GetAlocAddress`
-    return GetAlocAddress(aAddress, aRloc16);
 }
 
 otError Mle::GetServiceAloc(uint8_t aServiceId, Ip6::Address &aAddress) const
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -1108,14 +1108,14 @@ exit:
     return error;
 }
 
-otError Mle::GetLocatorAddress(Ip6::Address &aAddress, uint16_t aLoc16) const
+otError Mle::GetLocatorAddress(Ip6::Address &aAddress, uint16_t aLocator) const
 {
     otError error = OT_ERROR_NONE;
 
     VerifyOrExit(GetRloc16() != Mac::kShortAddrInvalid, error = OT_ERROR_DETACHED);
 
     memcpy(&aAddress, &mMeshLocal16.GetAddress(), 14);
-    aAddress.SetLocator(aLoc16);
+    aAddress.SetLocator(aLocator);
 
 exit:
     return error;
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -1121,6 +1121,12 @@ exit:
     return error;
 }
 
+otError Mle::GetRlocAddress(Ip6::Address &aAddress, uint16_t aRloc16) const
+{
+    // Same logic with `GetAlocAddress`
+    return GetAlocAddress(aAddress, aRloc16);
+}
+
 otError Mle::GetServiceAloc(uint8_t aServiceId, Ip6::Address &aAddress) const
 {
     otError error = OT_ERROR_NONE;
--- a/src/core/thread/mle.hpp
+++ b/src/core/thread/mle.hpp
@@ -779,7 +779,7 @@ public:
      * @retval OT_ERROR_DETACHED  The Thread interface is not currently attached to a Thread Partition.
      *
      */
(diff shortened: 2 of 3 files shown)


jwhui:
> Besides, we found that the `/a/ae` COAP message with link-local destination wouldn't be processed by by child's COAP layer.
> Thoughts on why it is unreachable by link-local address? @librasungirl @jwhui

See the rules documented here:

https://github.com/openthread/openthread/blob/2f618bd4a72501a2f31e23d50a4f388b9a8c6b9b/src/core/thread/thread_netif.cpp#L210-L213

But we force the source address to RLOC16 in `SendAddressError()` here:

https://github.com/openthread/openthread/blob/2f618bd4a72501a2f31e23d50a4f388b9a8c6b9b/src/core/thread/address_resolver.cpp#L691

simonlingoogle:
> But we force the source address to RLOC16 in `SendAddressError()` here:
> 
> https://github.com/openthread/openthread/blob/2f618bd4a72501a2f31e23d50a4f388b9a8c6b9b/src/core/thread/address_resolver.cpp#L691

I think we should fix `SendAddressError` too, by code like below:
```cpp
    if (aMessageInfo.GetSockAddr().IsLinkLocal() || aMessageInfo.GetSockAddr().IsLinkLocalMulticast())
    {
        messageInfo.SetSockAddr(Get<Mle::MleRouter>().GetLinkLocalAddress());
    }
    else
    {
        messageInfo.SetSockAddr(Get<Mle::MleRouter>().GetMeshLocal16());
    }
```
There should be other methods that have similar issue, e.x. `EnergyScanClient::SendQuery`:
https://github.com/openthread/openthread/blob/2f618bd4a72501a2f31e23d50a4f388b9a8c6b9b/src/core/meshcop/energy_scan_client.cpp#L61

jwhui:
> > But we force the source address to RLOC16 in `SendAddressError()` here:
> > https://github.com/openthread/openthread/blob/2f618bd4a72501a2f31e23d50a4f388b9a8c6b9b/src/core/thread/address_resolver.cpp#L691
> 
> I think we should fix `SendAddressError` too, by code like below:
> 
> ```c++
>     if (aMessageInfo.GetSockAddr().IsLinkLocal() || aMessageInfo.GetSockAddr().IsLinkLocalMulticast())
>     {
>         messageInfo.SetSockAddr(Get<Mle::MleRouter>().GetLinkLocalAddress());
>     }
>     else
>     {
>         messageInfo.SetSockAddr(Get<Mle::MleRouter>().GetMeshLocal16());
>     }
> ```

Agree. We could also simplify to:

```c++
    if (!(aMessageInfo.GetSockAddr().IsLinkLocal() || aMessageInfo.GetSockAddr().IsLinkLocalMulticast()))
    {
        messageInfo.SetSockAddr(Get<Mle::MleRouter>().GetMeshLocal16());
    }
```

And let default source address selection handle the link-local case.

> There should be other methods that have similar issue, e.x. `EnergyScanClient::SendQuery`:
> https://github.com/openthread/openthread/blob/2f618bd4a72501a2f31e23d50a4f388b9a8c6b9b/src/core/meshcop/energy_scan_client.cpp#L61

I think the assumption was that these messages are typically generated by a Commissioner. But I am fine to generalize this as well.

librasungirl:
For `ADDR_ERR.ntf`, Thread Specification 1.1 Section 5.6.3 says

> The IPv6 Source address MUST be the RLOC of the originator. The IPv6 Destination address
> is either the Realm-Local All-Routers multicast address (FF03::2) when sent via multicast or
> the RLOC of the destination when sent via unicast

For `MGMT_ED_SCAN.qry` or `MGMT_PANID_QUERY.qry`,  there is no clear text regarding the dest or source. Fine to select the source properly according to the dest passed in.

jwhui:
> For `ADDR_ERR.ntf`, Thread Specification 1.1 Section 5.6.3 says
> 
> > The IPv6 Source address MUST be the RLOC of the originator. The IPv6 Destination address
> > is either the Realm-Local All-Routers multicast address (FF03::2) when sent via multicast or
> > the RLOC of the destination when sent via unicast

Ah, yes. Thanks for the reminder.

simonlingoogle:
> > For `ADDR_ERR.ntf`, Thread Specification 1.1 Section 5.6.3 says
> > > The IPv6 Source address MUST be the RLOC of the originator. The IPv6 Destination address
> > > is either the Realm-Local All-Routers multicast address (FF03::2) when sent via multicast or
> > > the RLOC of the destination when sent via unicast
> 
> Ah, yes. Thanks for the reminder.

Thanks. This PR is justified now.

src/core/thread/mle.cpp
```diff
@@ -1121,6 +1121,12 @@ otError Mle::GetAlocAddress(Ip6::Address &aAddress, uint16_t aAloc16) const
     return error;
 }
 
+otError Mle::GetRlocAddress(Ip6::Address &aAddress, uint16_t aRloc16) const
```
librasungirl, line 1124: How about using a unique `GetLocatorAddress()` for both ALOC and RLOC address construction?

simonlingoogle: Will revise accordingly.

simonlingoogle: Created new function `GetLocatorAddress`

src/core/thread/mle.hpp
```diff
@@ -973,6 +973,18 @@ class Mle : public InstanceLocator
...
+     * This method gets the RLOC or ALOC of a given RLOC16 or ALOC16.
+     *
+     * @param[out]  aAddress  A reference to the RLOC or ALOC.
+     * @param[in]   aLoc16    RLOC16 or ALOC16.
+     *
+     * @retval OT_ERROR_NONE      If got the RLOC or ALOC successfully.
+     * @retval OT_ERROR_DETACHED  If device is detached.
+     *
+     */
+    otError GetLocatorAddress(Ip6::Address &aAddress, uint16_t aLoc16) const;
```
librasungirl, line 986: nit: `aLoc16`->`aLocator`?

simonlingoogle: renamed

PR #4812, opened 2020-04-10
[cmake] enable user specified mbedtls target

wgtdkp:
This PR allows users to specify the mbedtls target by a `OT_EXTERNAL_MBEDTLS` option. 

The native cmake build system of `mbedtls` is not so "modern" (for backward compatibility, it maybe hard to change it soon) and applications using `mbedtls` are likely to write wrapper cmake files and targets by themselves. We have see this problem in https://github.com/openthread/openthread/pull/4696 and also in the [esp-idf](https://github.com/espressif/esp-idf/blob/master/components/mbedtls/CMakeLists.txt) SDK.

This PR implements the second proposal at https://github.com/openthread/openthread/pull/4803#discussion_r406020566.


CMakeLists.txt
```diff
@@ -30,8 +30,8 @@ cmake_minimum_required(VERSION 3.13.1)
 project(openthread)
 
 option(OT_BUILD_EXECUTABLES "Build executables" ON)
-option(OT_BUILTIN_MBEDTLS "Enable builtin mbedTLS" ON)
 option(OT_COVERAGE "enable coverage" OFF)
+set(OT_EXTERNAL_MBEDTLS "" CACHE STRING "Specify external mbedtls library")
```
FiveDimensions, line 34: default value OFF?

third_party/CMakeLists.txt
```diff
@@ -26,6 +26,12 @@
...
 
-if(OT_BUILTIN_MBEDTLS)
+if(NOT OT_EXTERNAL_MBEDTLS)
     add_subdirectory(mbedtls)
 endif()
+
+add_library(openthread-mbedtls INTERFACE)
+
+target_link_libraries(openthread-mbedtls
+    INTERFACE $<IF:$<BOOL:${OT_EXTERNAL_MBEDTLS}>,${OT_EXTERNAL_MBEDTLS}, mbedtls>
```
FiveDimensions, line 36: new line for the expression $<IF:$<BOOL:${OT_EXTERNAL_MBEDTLS}>,${OT_EXTERNAL_MBEDTLS}, mbedtls>

CMakeLists.txt
wgtdkp: this is not an option, but a global string with default value `""`.

wgtdkp:
> Thank you for handling this. I can confirm that the proposed solution works nicely with Zephyr

@rlubos Thanks for your feedback!

src/ncp/CMakeLists.txt
```diff
@@ -105,7 +105,7 @@ target_link_libraries(openthread-ncp-ftd
     PUBLIC
         openthread-ftd
     PRIVATE
-        mbedtls
+        openthread-mbedtls
```
bukepo, line 108: Is this necessary? The core already depends on mbedtls.

src/core/CMakeLists.txt
```diff
@@ -245,10 +245,10 @@ target_sources(openthread-radio PRIVATE
 
 target_link_libraries(openthread-ftd
     PRIVATE
-        mbedtls
+        openthread-mbedtls
```
bukepo, line 248: Make this public?

CMakeLists.txt
bukepo, line 34: Can this be OT_MBEDTLS with default value mbedtls. And use this variable instead of adding an interface target?

src/core/CMakeLists.txt
wgtdkp: I think it should not be public, because `mbedtls` headers are not exported by public headers of `openthread-ftd`.

CMakeLists.txt
wgtdkp: We need a way to tell whether the user has set `OT_EXTERNAL_MBEDTLS` or not, so the default value has to be an invalid target name , such as `""`.

An [ALIAS target](https://cmake.org/cmake/help/latest/command/add_library.html#alias-libraries) should be more suitable for our usage case. But the `OT_EXTERNAL_MBEDTLS` target itself could be an `ALIAS target` and cmake doesn't allow ALIAS target on aliass.

bukepo: So how about use OT_MBEDTLS as an internal variable whose value is mbedtls and can be overridden by OT_EXTERNAL_MBEDTLS?

src/core/CMakeLists.txt
bukepo: Maybe this should be public even though mbedtls headers is not exported by openthread-ftd, otherwise every target depends on openthread-ftd must know what it depends on, otherwise there will be link errors, which I think is not what cmake dependency system designs for. Thoughts?

wgtdkp: > otherwise every target depends on openthread-ftd must know what it depends on, otherwise there will be link errors

I don't think this is true for CMake. When we write
```cmake
target_link_libraries(foo PRIVATE bar)
```

I think CMake will link `bar` and all its dependencies for `foo`.

bukepo: yes, you are right. Thanks.

PR #4817, opened 2020-04-12
[network-data] enhance/simplify updating of network data on leader

abtink:
This commit changes the model for updating Network Data on a leader.

It adds `ChangedFlags` type which is used to track whether full or stable version of Network Data gets changed as TLVs/sub-TLVs are being added or removed. This is then used to update version and stable version accordingly.

When registering new received Network Data on the leader, the new model updates the Network Data in place. First, newly added `Validate()` method is called to verify that the received Network Data contains well-formed TLVs and sub-TLVs (e.g., no duplicate Prefix/Service TLVs) and all sub-TLVs/entries match the sender's RLOC16. Then, all entries in the current Network Data associated with the sender's RLOC16 which are not present in the newly received data are removed. Afterwards, any new entry in the received Network Data is added. This approach helps simplify the code and ensures existing TLVs/sub-TLVs remain as before (e.g., no need to keep a copy of previous data to ensure same Service IDs are used when adding/removing Service TLVs).

This commit also adds `UpdatePrefix()` and `UpdateService()` methods which ensure a Prefix or Service TLV is marked correctly as stable or not depending on its sub-TLVs (e.g., if all stable sub-TLVs are removed the enclosing TLV is marked as not stable).

-----------

**[toranj] add test-case verifying Network Data update**

This commit adds a new test-case to `toranj` which verifies Network Data update and version changes (stable only vs. full version).

The test creates a simple network of a leader with 3 children where one of the children is SED and is configured to request only stable Network Data. The test covers the following:
- Adding/removing prefixes (stable or temporary) on the first child.
- Verifying that Network Data is updated on all nodes accordingly.
- Ensuring correct update to Network Data version and stable version.

The above steps are then repeated over many different situations:
- Where the same prefixes are also added by other nodes.
- Or the same prefixes are added as off-mesh routes by other nodes.

### DIFF
--- a/src/core/thread/network_data_leader_ftd.cpp
+++ b/src/core/thread/network_data_leader_ftd.cpp
@@ -104,6 +104,14 @@ void Leader::IncrementVersionAndStableVersion(void)
     }
 }
 
+void Leader::IncrementVersions(const ChangedFlags &aFlags)
+{
+    if (aFlags.DidChange())
+    {
+        IncrementVersions(aFlags.DidStableChange());
+    }
+}
+
 void Leader::IncrementVersions(bool aIncludeStable)
 {
     if (aIncludeStable)
@@ -117,16 +125,10 @@ void Leader::IncrementVersions(bool aIncludeStable)
 
 void Leader::RemoveBorderRouter(uint16_t aRloc16, MatchMode aMatchMode)
 {
-    bool rlocIn;
-    bool rlocStable;
+    ChangedFlags flags;
 
-    RlocLookup(aRloc16, rlocIn, rlocStable, mTlvs, mLength, aMatchMode);
-    VerifyOrExit(rlocIn);
-    RemoveRloc(aRloc16, aMatchMode);
-    IncrementVersions(rlocStable);
-
-exit:
-    return;
+    RemoveRloc(aRloc16, aMatchMode, flags);
+    IncrementVersions(flags);
 }
 
 void Leader::HandleServerData(void *aContext, otMessage *aMessage, const otMessageInfo *aMessageInfo)
@@ -402,19 +404,14 @@ bool Leader::RlocMatch(uint16_t aFirstRloc16, uint16_t aSecondRloc16, MatchMode 
     return matched;
 }
 
-otError Leader::RlocLookup(uint16_t       aRloc16,
-                           bool &         aIn,
-                           bool &         aStable,
-                           const uint8_t *aTlvs,
-                           uint8_t        aTlvsLength,
-                           MatchMode      aMatchMode,
-                           bool           aAllowOtherEntries)
+otError Leader::Validate(const uint8_t *aTlvs, uint8_t aTlvsLength, uint16_t aRloc16)
 {
+    // Validate that the `aTlvs` contains well-formed TLVs, sub-TLVs,
+    // and entries all matching `aRloc16` (no other entry for other
+    // RLOCs and no duplicates TLVs).
+
     otError               error = OT_ERROR_NONE;
     const NetworkDataTlv *end   = reinterpret_cast<const NetworkDataTlv *>(aTlvs + aTlvsLength);
-
-    aIn     = false;
-    aStable = false;
 
     for (const NetworkDataTlv *cur = reinterpret_cast<const NetworkDataTlv *>(aTlvs); cur < end; cur = cur->GetNext())
     {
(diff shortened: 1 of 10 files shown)


src/core/thread/network_data_leader_ftd.cpp
```diff
@@ -402,20 +404,15 @@ bool Leader::RlocMatch(uint16_t aFirstRloc16, uint16_t aSecondRloc16, MatchMode
     return matched;
 }
 
-otError Leader::RlocLookup(uint16_t       aRloc16,
```
abtink, line 405: Adding a note about the change here. The diff here is not easy to follow, so I suggest reading the new code only.

Basically here we are removing two larger methods `RlocLookup()` and `IsStableUpdated()` that are no longer needed in the new approach. Instead we have a group of `Validate()` methods  followed by a group of static helper `ContainsMatchingEntry` and then we have `UpdatePrefix/Service()` methods. There is a short comment at the beginning of each method describing what it does.

src/core/thread/network_data_leader_ftd.cpp
```diff
@@ -1135,13 +1081,15 @@ void Leader::RemoveRloc(uint16_t aRloc16, MatchMode aMatchMode)
...
+                FindPrefix(prefix->GetPrefix(), prefix->GetPrefixLength(), aExcludeTlvs, aExcludeTlvsLength);
 
-            RemoveRloc(*prefix, aRloc16, aMatchMode);
+            RemoveRlocInPrefix(*prefix, aRloc16, aMatchMode, excludePrefix, aChangedFlags);
 
-            if (prefix->GetSubTlvsLength() == 0)
+            if (UpdatePrefix(*prefix) == kTlvRemoved)
             {
-                RemoveTlv(prefix);
+                // Do not update `cur` when TLV is removed.
```
jwhui, line 1092: Not introduced by this PR, but this skips the log message below. Should we improve the logging?

src/core/thread/network_data_tlvs.hpp
```diff
@@ -1133,14 +1174,20 @@ class ServerTlv : public NetworkDataTlv
     };
 
     /**
-     * This method initializes the TLV.
+     * This method initializes the Server TLV.
+     *
+     * @param[in] aServer16          The Server16 value.
+     * @param[in] aServerData        A Pointer to the Server Data.
```
jwhui, line 1180:
```suggestion
     * @param[in] aServerData        The Server Data.
```

src/core/thread/network_data_leader_ftd.cpp
```diff
@@ -886,221 +852,186 @@ otError Leader::AddHasRoute(PrefixTlv &aPrefix, HasRouteTlv &aHasRoute)
...
-            otLogInfoNetData("Allocated Service ID = %d", i);
+        dstBorderRouter = static_cast<BorderRouterTlv *>(aDstPrefix.GetNext());
+        Insert(dstBorderRouter, sizeof(BorderRouterTlv));
+        aDstPrefix.IncreaseLength(sizeof(BorderRouterTlv));
+        dstBorderRouter->Init();
 
-            VerifyOrExit(i <= Mle::kServiceMaxId, error = OT_ERROR_NO_BUFS);
+        if (aBorderRouter.IsStable())
+        {
+            dstBorderRouter->SetStable();
```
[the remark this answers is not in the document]
abtink: Not sure if we can use this.  The `UpdateTlv()` updates an enclosing TLV's stable flag based on its sub-TLVs, but here the `dstBorderRouter` is the sub-TLV itself.

src/core/thread/network_data_leader_ftd.cpp
```diff
@@ -424,130 +421,122 @@ otError Leader::RlocLookup(uint16_t       aRloc16,
...
 
-                switch (subCur->GetType())
-                {
-                case NetworkDataTlv::kTypeServer:
-                {
-                    const ServerTlv *server = static_cast<const ServerTlv *>(subCur);
+            foundEntry = false;
 
-                    VerifyOrExit(server->IsValid(), error = OT_ERROR_PARSE);
+            for (const BorderRouterEntry *entry = borderRouter->GetFirstEntry(); entry <= borderRouter->GetLastEntry();
```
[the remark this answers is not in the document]
abtink: Good to know. Thanks. It can help simplify the `Validate()` and adding the entry.

src/core/thread/network_data_leader_ftd.cpp
abtink: I agree.  I actually prefer if we remove some of the debg level logs (e.g. from `RemoveRloc()` we just log at the end after all related TLVs/sub-TLvs/entries are removed (and not log per prefix/entry). We already do the same when adding new entries. What do you think?

jwhui: Sounds good to me!

abtink: Changed this in the latest pushed commit.

PR #4818, opened 2020-04-13
[otns] allow virtual time UART

simonlingoogle:
This PR allows use to build `OTNS=1` with or without `VIRTUAL_TIME_UART`.

Later I will enhance OTNS to support both virtual time UART and non virtual time UART.


simonlingoogle:
Hi @jwhui , this PR now does not have to be merged together with [OTNS #5](https://github.com/openthread/ot-ns/pull/5) because OTNS is going to support both.

PR #4819, opened 2020-04-13
[logging] add bbr region

librasungirl:
This PR depends on #4795 , please only review the last commit which includes the update.


src/core/backbone_router/leader.cpp
```diff
@@ -87,47 +83,56 @@ otError Leader::GetServiceId(uint8_t &aServiceId) const
...
     OT_UNUSED_VARIABLE(aConfig);
 
-    otLogInfoNetData("BBR state %s", StateToString(aState));
+    otLogInfoBbr("PBBR state: %s", StateToString(aState));
 
     if (aState != kStateRemoved && aState != kStateNone)
     {
-        otLogInfoNetData("Rloc16: 0x%4X, seqno: %d, delay: %d, timeout %d", aConfig.mServer16, aConfig.mSequenceNumber,
-                         aConfig.mReregistrationDelay, aConfig.mMlrTimeout);
+        otLogInfoBbr("Rloc16: 0x%4X, seqno: %d, delay: %d, timeout %d", aConfig.mServer16, aConfig.mSequenceNumber,
```
simonlingoogle, line 95: merge logs to one line ?

librasungirl: This PR serves only for logging update, I'd prefer not to introduce unrelated changes into this one.

Besides, regarding the comment, there is no valid dataset when the state is `kStateRemoved` or `kStateNone`,  one line for all scenarios would make it not easy to read the code in my view.

Any suggestion to achieve one line log? @simonlingoogle

librasungirl:
rebased as the dependent #4795 gets merged.

src/core/backbone_router/leader.cpp
simonlingoogle: We can simply use `if else` with 2 different lines of `otLogInfoBbr`, but only one will be executed at runtime.

librasungirl: It is an option. No strong opinion. Anyway feel free to submit a PR for it if you prefer this way. Thanks.

Will mark this as resolved as it is not related to this PR.

src/core/common/logging.hpp
```diff
@@ -89,6 +89,9 @@ extern "C" {
 #define _OT_REGION_CLI_PREFIX "-CLI-----: "
 #define _OT_REGION_CORE_PREFIX "-CORE----: "
 #define _OT_REGION_UTIL_PREFIX "-UTIL----: "
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
```
jwhui, line 92: nit: how about we avoid feature-flagging these defines (here and below)?

librasungirl: removed.

PR #4823, opened 2020-04-14
[travis] replace tab with 4 spaces in bash script

simonlingoogle:


.travis/before_install.sh
```diff
@@ -30,8 +30,8 @@
 [ -n "$BUILD_TARGET" ] || exit 0
 
 die() {
-	echo " *** ERROR: " $*
-	exit 1
+    echo " *** ERROR: " $*
```
bukepo, line 33:
```suggestion
    echo " *** ERROR:  $*"
```

jwhui:
FYI - Travis-specific scripts will be removed once the migration to GitHub Actions is complete (#4824).

.travis/before_install.sh
simonlingoogle: I would prefer this PR to only fix indentation issue.

jwhui:
This file no longer exists.

Issue #4827, opened 2020-04-14
REED unable to synchronize network data to Leader when Leader has smaller `routerupgradethreshold`


simonlingoogle:
**Describe the bug**
If the Leader has smaller `routerupgradethreshold` than the REED, the REED will keep sending `AddressSolicit` and being rejected by the Leader. In the meantime, the REED won't be able to synchronize local network data to Leader, e.x. it won't be able to become Primary BBR. 

**To Reproduce**

1. Git commit id: master
2. IEEE 802.15.4 hardware platform: simulation
3. Build steps: 
```
make -f examples/Makefile-simulation BACKBONE_ROUTER=1 THREAD_VERSION=1.2
```
4. Network topology
```
Leader---REED
```
Execution steps:
  1. Start the leader and wait for it to become leader
  2. Set `routerupgradethreshold` to 1 on Leader: `routerupgradethreshold 1`
  3. Start the REED
  4. Start BBR service on REED: `bbr enable`
  5. Wait for a long time, but REED never become Primary BBR:
```
node 2> bbr enable
Done
node 2> 
node 2> bbr
BBR Primary: None
Done
```

**Expected behavior**
The REED should become Primary BBR. 

**Proposed fix**
A simple fix can be if the REED receives `a/as` ACK with `status == 1 (No Address Available)`, it can update its `routerupgradethreshold` like below:
```cpp
routerupgradethreshold = min(routerupgradethreshold, active_router_count)
``` 
This can help the node exit `REED` status and return back to child state. 

Thoughts? @librasungirl @jwhui

librasungirl:
It should be a common issue regardless of the Thread version. And it may happen in user scenarios when Thread devices from different vendors have different `routerupgradethreshold` configuration (though default 16, but it is configurable).

@simonlingoogle in your proposal, will it change the [mRouterUpgradeThreshold](https://github.com/openthread/openthread/blob/master/src/core/thread/mle_router.hpp#L827)?

I may suggest to introduce a variable like `mRunningRouterUpgradeThreshold`, thus when the device need to become Leader due to lost of Leader, it still have the chance to adopt its original `mRouterUpgradeThreshold` value when deciding whether or not to allocate Router ID for the others.

simonlingoogle:
> It should be a common issue regardless of the Thread version.

Yes, it is a common issue. I am using `BBR service` mainly for easy illustration. The issue can happen whenever the Leader has smaller `routerupgradethreshold` than the REED.

[...]

PR #4834, opened 2020-04-15
[mle] REED to handle rejected ADDR_SOL.rsp

simonlingoogle:
Currently, a REED which expects to become Router soon does not send any `SVR_DATA.ntf`. 
This PR handles rejecting `ADDR_SOL.rsp` to allow REED to send `SVR_DATA.ntf` after it's rejected. 
- [x] Commit 1 wraps existing logic within `MleRouter::ExpectBecomeRouterSoon`
- [x] Commit 2 adds a test that should fail without fix
- [x] Commit 3 fixes this issue

This PR addresses #4827 .

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -4050,10 +4050,10 @@ void MleRouter::HandleAddressSolicitResponse(Coap::Message *         aMessage,
 
     SuccessOrExit(Tlv::ReadUint8Tlv(*aMessage, ThreadTlv::kStatus, status));
 
-    mAddressSolicitRejected = (status != ThreadStatusTlv::kSuccess);
-
     if (status != ThreadStatusTlv::kSuccess)
     {
+        mAddressSolicitRejected = true;
+
         if (IsRouterIdValid(mPreviousRouterId))
         {
             if (HasChildren())
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -74,6 +74,7 @@ MleRouter::MleRouter(Instance &aInstance)
     , mFixedLeaderPartitionId(0)
     , mRouterEligible(true)
     , mAddressSolicitPending(false)
+    , mAddressSolicitRejected(false)
     , mPreviousPartitionIdRouter(0)
     , mPreviousPartitionId(0)
     , mPreviousPartitionRouterIdSequence(0)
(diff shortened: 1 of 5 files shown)


src/core/thread/mle_router.hpp
```diff
@@ -319,6 +319,15 @@ class MleRouter : public Mle
...
     void SetRouterDowngradeThreshold(uint8_t aThreshold) { mRouterDowngradeThreshold = aThreshold; }
 
+    /**
+     * This method returns if the REED is expected to become Router soon.
+     *
+     * @retval TRUE   If the REED is going to become Router.
+     * @retval FALSE  Otherwise.
+     *
+     */
+    bool ExpectBecomeRouterSoon(void) const;
```
abtink, line 329: suggestion: how about `IsExpectedToBecomeRouter()`?

simonlingoogle: renamed, thanks.

src/core/thread/mle_router.cpp
```diff
@@ -4046,6 +4050,8 @@ void MleRouter::HandleAddressSolicitResponse(Coap::Message *         aMessage,
 
     SuccessOrExit(Tlv::ReadUint8Tlv(*aMessage, ThreadTlv::kStatus, status));
 
+    mAddressSolicitRejected = (status != ThreadStatusTlv::kSuccess);
```
librasungirl, line 4053: move it into below `if` which does status check?

simonlingoogle: This also sets `mAddressSolicitRejected` to `TRUE` when status == `kSuccess`. 
Even though I think it might  not be necessary.

librasungirl: you should mean `mAddressSolicitRejected` is `FALSE` when status is `kSuccess`, which should be done already when the device becomes child, that is, before address solicit request 

https://github.com/openthread/openthread/blob/4cb5d1b30acbdebebe8e7ee9543afe2168c1494d/src/core/thread/mle_router.cpp#L232-L237

Given it is not necessary to clear here, prefer to put a simple sentence in below `if`

simonlingoogle: Revised accordingly. 

I meant that when `status == kSuccess`, the original code sets `mAddressSolicitRejected` to `FALSE`, which reflects the fact that it is not rejected by the Leader. 
However, after the REED upgrade to Router, the `mAddressSolicitRejected` flag is not used anywhere. So, whether or not setting it to `FALSE` on `status=kSuccess` has no impact.

PR #2618, opened 2018-03-15
[time-sync] network-wide time synchronization service

chshu:
OpenThread network-wide time synchronization service is an **experimental** feature, not part of the standard Thread protocol.

Feature Overview:
 * All the nodes in the same Thread partition sync to the same Thread network-wide time;
 * Microsecond level time synchronization precision;
 * Flexible time accuracy and time sync period configuration;
 * APIs for application layer use case, support both CLI and NCP version.

The feature is wrapped by `OPENTHREAD_CONFIG_ENABLE_TIME_SYNC`, there's no change to current Thread 1.1 implementation if `OPENTHREAD_CONFIG_ENABLE_TIME_SYNC` is not enabled.

If `OPENTHREAD_CONFIG_ENABLE_TIME_SYNC` is enabled, the node could:
 * Attach to a time sync enabled network, otherwise
 * Attach to a standard Thread 1.1 network, otherwise
 * Form a new time sync enabled network

In addition, if `OPENTHREAD_CONFIG_TIME_SYNC_ESSENTIAL` is also enable, the node could only:
 * Attach to a time sync enabled network, otherwise
 * Form a new time sync enabled network

Note:
Currently, the feature is only supported on nRF52840 and posix platforms. And the network time
is only synced among Routers and REEDs, the SEDs will be supported later as an optional feature.

PR #4836, opened 2020-04-15
[github-actions] run cli-time-sync with Thread version 1.2

Irving-cl:
I found that in our github-actions, `cli-time-sync` is run under default Thread version. Time sync needs `OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT`. According to **IEEE Std 802.15.4-2015**:

> 7.2.1.7 IE Present field
> The IE Present field shall be set to one if IEs are contained in the frame. This field shall be set to zero
> otherwise. If the Frame Version field is 0b00 or 0b01, the IE Present field shall be zero.

So the time sync feature shouldn't be enabled in Thread version 1.1.


Irving-cl:
Seems that we have some problems now. Just like I said above, since **TIME_SYNC** requires **HEADER_IE** and **HEADER_IE** shouldn't be used in 1.1 devices, **TIME_SYNC** should only be tested with 1.2 devices. 

In the current testing, case 8_1_1 has a check which compares the version with fixed value(2, Thread version 1.1). That caused a failure now. So should all these **Cert** cases work for both 1.1 and 1.2 devices? Or we should remove this `TIME_SYNC` test?

@librasungirl @bukepo @jwhui

jwhui:
@Irving-cl, the time-sync service was originally developed against Thread 1.1 in #2618. Time-sync is currently an experimental feature.

I think the real question here is whether we want to continue maintaining the time-sync feature as it was originally conceived moving forward.

@librasungirl, thoughts?

librasungirl:
IIRC, TIME_SYNC feature was designed that the device without TIME_SYNC would only attach as child in a mixed network with TIME_SYNC, which should be the same even in Thread  1.2  network I think?

A second thought on this topic, since TIME_SYNC is explicitly declared as an **experimental** as @jwhui mentioned, I think it may be independently enabled regardless of Thread Version according to the adopter's use scenario at their own risk.

I may suggest to enable HEADER_IE_SUPPORT when CSL is enabled, and use this flag to wrap IEEE802.15.4-2015 specific code.

Thoughts?

Irving-cl:
Ok I see. Then I would close this PR. So:

* We would allow **TIME_SYNC** to run on 1.1 device.
* We would use the macro **HEADER_IE_SUPPORT** to control the code related to handling versio n 2015 frame.

librasungirl:
@Irving-cl Let's wait confirmation from @jwhui .

Irving-cl:
Ok sure.

PR #4837, opened 2020-04-15
[mle] 1.2 MED registers MA (scope > 3) to its parent via AR TLV

librasungirl:


src/core/thread/mle.cpp
```diff
@@ -1446,15 +1446,27 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
         VerifyOrExit(counter < OPENTHREAD_CONFIG_MLE_IP_ADDRS_TO_REGISTER);
     }
 
-    // For sleepy end device, register external multicast addresses to the parent for indirect transmission
+    // For sleepy end device, register external multicast addresses to the parent for indirect transmission.
+    // Since Thread 1.2, non-sleepy MED would depend on its Thread 1.2 parent as the proxy to perform
+    // Multicast Listener Report, it registers external multicast addresses of scope larger than realm
+    // local as well.
+#if (OPENTHREAD_CONFIG_THREAD_VERSION == OT_THREAD_VERSION_1_1)
```
jwhui, line 1453: To reduce the number of preprocessor statements and simplify the code, how about removing this if-statement and combine into code within loop?

src/core/thread/mle.cpp
```diff
@@ -1446,15 +1446,27 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
...
     {
         uint8_t      iterator = 0;
         Ip6::Address address;
 
         // append external multicast address
         while (Get<ThreadNetif>().GetNextExternalMulticast(iterator, address) == OT_ERROR_NONE)
         {
+#if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
+            // For Thread 1.2 MED, skip multicast address with scope not larger than realm local.
+            if (IsRxOnWhenIdle() && !address.IsMulticastLargerThanRealmLocal())
```
jwhui, line 1465: Should we also have a check for Thread 1.1 vs 1.2 parent?

src/core/thread/mle.cpp
```diff
@@ -1583,7 +1595,14 @@ void Mle::HandleStateChanged(otChangedFlags aFlags)
 
     if ((aFlags & (OT_CHANGED_IP6_MULTICAST_SUBSCRIBED | OT_CHANGED_IP6_MULTICAST_UNSUBSCRIBED)) != 0)
     {
+#if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
```
simonlingoogle, line 1598: What about we add a new method for judging if the node needs to register MAs so that we can avoid using different conditions for different thread versions, making the code more compact?

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+        },
+    }
+    """All nodes are created with default configurations"""
+
+    def test(self):
+
+        # 1) Bring up Leader_1_2.
+        self.nodes[LEADER_1_2].start()
+        WAIT_TIME = WAIT_ATTACH
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 115:
```suggestion
        self.simulator.go(WAIT_ATTACH)
```
Variable `WAIT_TIME` seems redundant. Many same issues follows.

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+        self.assertEqual(self.nodes[LEADER_1_2].get_state(), 'leader')
+
+        # 2) Bring up MED_1_2.
+        self.nodes[MED_1_2].start()
+        WAIT_TIME = WAIT_ATTACH
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 121:
```suggestion
        self.simulator.go(WAIT_ATTACH)
```

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+        self.assertEqual(self.nodes[MED_1_2].get_state(), 'child')
+
+        # 2a) add MA1_LINKLOCAL which would NOT be registered in AddressRegistrationTLV of Child Update Request.
+        # Flush relative message queues.
+        self.flush_nodes([MED_1_2])
+        self.nodes[MED_1_2].add_ipmaddr(MA1_LINKLOCAL)
+        WAIT_TIME = WAIT_REDUNDANCE
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 129:
```suggestion
        self.simulator.go(WAIT_REDUNDANCE)
```

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+            msg, MA1_LINKLOCAL
+        ), 'Error: Unexpected %s in AddressRegistrationTLV'.format(
+            MA1_LINKLOCAL)
+
+        # 2b) add MA2_ADMINSCOPE which would be registered in AddressRegistrationTLV of Child Update Request.
+        # Flush relative message queues.
+        self.flush_nodes([MED_1_2])
+        self.nodes[MED_1_2].add_ipmaddr(MA2_ADMINSCOPE)
+        WAIT_TIME = WAIT_REDUNDANCE
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 142:
```suggestion
        self.simulator.go(WAIT_REDUNDANCE)
```

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+        msg = messages.next_mle_message(mle.CommandType.CHILD_UPDATE_REQUEST)
+        assert command.check_address_registration_tlv(
+            msg, MA2_ADMINSCOPE
+        ), 'Error: Expected %s in AddressRegistrationTLV not found'.format(
+            MA2_ADMINSCOPE)
+
+        # 3) Bring up SED_1_2.
+        self.nodes[SED_1_2].start()
+        WAIT_TIME = WAIT_ATTACH
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 153:
```suggestion
        self.simulator.go(WAIT_ATTACH)
```

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+        self.assertEqual(self.nodes[SED_1_2].get_state(), 'child')
+
+        # 3a) add MA1_LINKLOCAL which would be registered in AddressRegistrationTLV of Child Update Request.
+        # Flush relative message queues.
+        self.flush_nodes([SED_1_2])
+        self.nodes[SED_1_2].add_ipmaddr(MA1_LINKLOCAL)
+        WAIT_TIME = WAIT_REDUNDANCE
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 161:
```suggestion
        self.simulator.go(WAIT_REDUNDANCE)
```

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+            msg, MA1_LINKLOCAL
+        ), 'Error: Expected %s in AddressRegistrationTLV not found'.format(
+            MA1_LINKLOCAL)
+
+        # 3b) add MA2_ADMINSCOPE which would be registered in AddressRegistrationTLV of Child Update Request.
+        # Flush relative message queues.
+        self.flush_nodes([SED_1_2])
+        self.nodes[SED_1_2].add_ipmaddr(MA2_ADMINSCOPE)
+        WAIT_TIME = WAIT_REDUNDANCE
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 174:
```suggestion
        self.simulator.go(WAIT_REDUNDANCE)
```

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+        msg = messages.next_mle_message(mle.CommandType.CHILD_UPDATE_REQUEST)
+        assert command.check_address_registration_tlv(
+            msg, MA2_ADMINSCOPE
+        ), 'Error: Expected %s in AddressRegistrationTLV not found'.format(
+            MA2_ADMINSCOPE)
+
+        # 4) Bring up MED_1_1.
+        self.nodes[MED_1_1].start()
+        WAIT_TIME = WAIT_ATTACH
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 185:
```suggestion
        self.simulator.go(WAIT_ATTACH)
```

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+        self.assertEqual(self.nodes[MED_1_1].get_state(), 'child')
+
+        # 4a) add MA1_LINKLOCAL which would NOT be registered in AddressRegistrationTLV of Child Update Request.
+        # Flush relative message queues.
+        self.flush_nodes([MED_1_1])
+        self.nodes[MED_1_1].add_ipmaddr(MA1_LINKLOCAL)
+        WAIT_TIME = WAIT_REDUNDANCE
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 193:
```suggestion
        self.simulator.go(WAIT_REDUNDANCE)
```

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+                msg, MA1_LINKLOCAL
+            ), 'Error: Unexpected %s in AddressRegistrationTLV'.format(
+                MA1_LINKLOCAL)
+
+        # 4b) add MA2_ADMINSCOPE which would NOT be registered in AddressRegistrationTLV of Child Update Request.
+        # Flush relative message queues.
+        self.flush_nodes([MED_1_1])
+        self.nodes[MED_1_1].add_ipmaddr(MA2_ADMINSCOPE)
+        WAIT_TIME = WAIT_REDUNDANCE
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 209:
```suggestion
        self.simulator.go(WAIT_REDUNDANCE)
```

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+        if msg:
+            assert not command.check_address_registration_tlv(
+                msg, MA2_ADMINSCOPE
+            ), 'Error: Unexpected %s in AddressRegistrationTLV'.format(
+                MA2_ADMINSCOPE)
+
+        # 5) Bring up SED_1_1.
+        self.nodes[SED_1_1].start()
+        WAIT_TIME = WAIT_ATTACH
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 223:
```suggestion
        self.simulator.go(WAIT_ATTACH)
```

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+        self.assertEqual(self.nodes[SED_1_1].get_state(), 'child')
+
+        # 5a) add MA1_LINKLOCAL which would be registered in AddressRegistrationTLV of Child Update Request.
+        # Flush relative message queues.
+        self.flush_nodes([SED_1_1])
+        self.nodes[SED_1_1].add_ipmaddr(MA1_LINKLOCAL)
+        WAIT_TIME = WAIT_REDUNDANCE
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 231:
```suggestion
        self.simulator.go(WAIT_REDUNDANCE)
```

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+            msg, MA1_LINKLOCAL
+        ), 'Error: Expected %s in AddressRegistrationTLV not found'.format(
+            MA1_LINKLOCAL)
+
+        # 5b) add MA2_ADMINSCOPE which would be registered in AddressRegistrationTLV of Child Update Request.
+        # Flush relative message queues.
+        self.flush_nodes([SED_1_1])
+        self.nodes[SED_1_1].add_ipmaddr(MA2_ADMINSCOPE)
+        WAIT_TIME = WAIT_REDUNDANCE
+        self.simulator.go(WAIT_TIME)
```
simonlingoogle, line 244:
```suggestion
        self.simulator.go(WAIT_REDUNDANCE)
```

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
```diff
@@ -0,0 +1,254 @@
...
+        self.flush_nodes([MED_1_2])
+        self.nodes[MED_1_2].add_ipmaddr(MA1_LINKLOCAL)
+        WAIT_TIME = WAIT_REDUNDANCE
+        self.simulator.go(WAIT_TIME)
+        messages = self.simulator.get_messages_sent_by(MED_1_2)
+        msg = messages.next_mle_message(mle.CommandType.CHILD_UPDATE_REQUEST)
+        assert not command.check_address_registration_tlv(
+            msg, MA1_LINKLOCAL
+        ), 'Error: Unexpected %s in AddressRegistrationTLV'.format(
+            MA1_LINKLOCAL)
```
simonlingoogle, line 135: There are many code blocks like this in this script. What about we wrap in a method following logic and call the method each time for better code reuse:

**def checkMulticastReigstation(id, ...):**
- flush node messages
- add an MA
- wait for `WAIT_REDUNDANCE`
- get next `Child Update Request` message
- assert if the `AR TLV` contains or not contains the `MA`

librasungirl: note that for some scenario, child update request is mandatory, but optional for the others.

Agree that there are the some code repeats.
However considering that it is just a unit test serving for the functions introduced in this PR, and per section has clear comments stating the goal, I may prefer existing way a bit as it seems to me easier for readers to know what happened per step.

librasungirl: revise my previous comment. a well-named method should have no impact on the readability and understanding of the test steps. will upate.

src/core/thread/mle.cpp
simonlingoogle: According to Spec 5.24.3.3, if an MTD attaches to a 1.1 parent, it should perform both actions:

- Registering with MLR.req 
- Registering with MLE Address Registration TLV

Thoughts @jwhui @librasungirl

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
simonlingoogle: Any reason to make `go()` line same?

src/core/thread/mle.cpp
librasungirl: @simonlingoogle  
 I am trying to split into small relative independent PRs to ease the review process. 
MLR would be in a separate PR. Please only focus on the purpose as the subject reveals. Thanks.

librasungirl: @jwhui added parent version check. thanks.

src/core/thread/mle.cpp
librasungirl: added a private method.

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
librasungirl: just personal preference, no strong specific reason. if have to say one, `WAIT_TIME = ` is shorter than `self.simulator.go(`, a bit less chance to cause line break.

src/core/thread/mle.cpp
```diff
@@ -4469,5 +4487,22 @@ void Mle::UpdateAllDomainBackboneRouters(BackboneRouter::Leader::DomainPrefixSta
 }
 #endif // OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
 
+// Since Thread 1.2, MTD end device (including rx-on-when-idle or rx-off-when-idle) should
+// register or de-register their multicast address larger than realm local to the parent via
+// MLE Child Update Request.
+bool Mle::ShouldUpdateMulticastRegistration(void) const
```
simonlingoogle, line 4493:
```suggestion
bool Mle::ShouldRegisterMulticastToParent(void) const
```
I think `ShouldUpdateMulticastRegistration` is not a good name because we can argue that FED sending `MLR.req` is also a way of updating Multicast Registration?

src/core/thread/mle.cpp
```diff
@@ -1446,24 +1448,40 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
...
-            length += entry.GetLength();
-
-            counter++;
-            // only continue to append if there is available entry.
-            VerifyOrExit(counter < OPENTHREAD_CONFIG_MLE_IP_ADDRS_TO_REGISTER, OT_NOOP);
+#if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
+            // For Thread 1.2 MED, skip MA registration if the parent is Thread 1.1 device.
+            if (GetParent().IsThreadVersion2())
+            {
+                break;
```
simonlingoogle, line 1464: In Spec 5.24.3.3, it seems to me that a Thread 1.2 MED should not skip MA registration even if the parent is a Thread 1.1 device. 
```
Once an MTD is attached to such a Parent, it performs both these registration actions:
• Registering with MLR.req as described in Section 5.24.2.3, Registration as Multicast Listener (for FTDs)
• Registering with MLE Address Registration TLV as described in Section 5.24.3.1, Registration as Multicast Listener.
```
Note that the Spec says **BOTH** actions are required. 
So, 1.2 MED should not skip MA registration if parent is 1.1?

tests/scripts/thread-cert/v1_2_test_multicast_registration.py
simonlingoogle: Personal preference too. I am fine to keep the code as it is.

src/core/thread/topology.hpp
```diff
@@ -351,6 +351,14 @@ class Neighbor : public InstanceLocatorInit
      */
     void SetRloc16(uint16_t aRloc16) { mRloc16 = aRloc16; }
 
+    /**
+     * This method indicates whether or not it is a valid Thread 1.1 neighbor.
+     *
+     * @returns TRUE if it is a valid Thread 1.1 neighbor, FALSE otherwise.
+     *
+     */
+    bool IsThreadVersion2(void) const { return mState != kStateInvalid && mVersion == OT_THREAD_VERSION_1_1; }
```
simonlingoogle, line 360:
```suggestion
    bool IsThreadVersion1_1(void) const { return mState != kStateInvalid && mVersion == OT_THREAD_VERSION_1_1; }
```
Prefer `IsThreadVersion1_1` because we commonly say version 1.1, 1.2, we never say version 2, 3 in our conversations. I think `IsThreadVersion1_1` is more understandable to code readers.

src/core/thread/mle.cpp
librasungirl: Good catch! 

However I don't see any value for a 1.2 MED registers MA to the Thread 1.1 parent. 
On the contrary, it may cause one extra Child Update Request/Response exchange and  may consume one of the limited address storage capability on the  parent side.

Not sure the background of the text in spec, should we file a ticket to Thread  Group? @jwhui

src/core/thread/mle.cpp
librasungirl: renamed.

src/core/thread/topology.hpp
librasungirl: no strong preference. renamed as your suggestion.

src/core/thread/mle.cpp
jwhui: Yes, please file a Thread Group JIRA to discuss this.

simonlingoogle: Do we need to choose for this PR for now?  
I would prefer to just sending `MA reg` to 1.1 parent too because it does what Spec. says.

src/core/thread/mle.cpp
```diff
@@ -4469,5 +4487,22 @@ void Mle::UpdateAllDomainBackboneRouters(BackboneRouter::Leader::DomainPrefixSta
...
+// MLE Child Update Request.
+bool Mle::ShouldRegisterMulticastToParent(void) const
+{
+    bool update = false;
+
+    VerifyOrExit(IsChild() && !IsFullThreadDevice(), OT_NOOP);
+#if (OPENTHREAD_CONFIG_THREAD_VERSION == OT_THREAD_VERSION_1_1)
+    VerifyOrExit(!IsRxOnWhenIdle(), OT_NOOP);
+#endif
+    update = true;
```
simonlingoogle, line 4501: If we choose the strategy that 1.2 MED does not send MA registration to 1.1 parent, should we check for 1.1 parent here?

src/core/thread/mle.cpp
librasungirl: @simonlingoogle 
is there any test case in 1.2 TP that requires that 1.2 MED sends CUR to register MA to 1.1 parent?

simonlingoogle: Verified that in the 5.10.20 test case, a MTD is required to register MAs to its 1.1 parent too, which is in conformance with the current Spec.

librasungirl: Thanks for the information @simonlingoogle 

Filed BHC-612 to raise the discussion in Thread Group. Also mentioned possible impact on test 5.10.20.

As we are not switching to the public repo for certification recently, I would prefer to keep the code as is before BHC-612 is resolved.

Thoughts?
@jwhui @simonlingoogle

simonlingoogle: I think following the current Spec. actually lead to less code. So I would suggest follow the Spec.

src/core/thread/mle.cpp
librasungirl: updated. please take a review.

src/core/thread/mle.cpp
jwhui: I don't have a strong opinion either way. I am fine to implement the spec for now and change when BHC-612 is resolved.

src/core/thread/mle.cpp
```diff
@@ -4469,5 +4487,27 @@ void Mle::UpdateAllDomainBackboneRouters(BackboneRouter::Leader::DomainPrefixSta
...
+// register or de-register their multicast address larger than realm local to its Thread 1.2
+// parent via MLE Child Update Request.
+bool Mle::ShouldRegisterMulticastToParent(void) const
+{
+    bool update = false;
+
+    VerifyOrExit(IsChild() && !IsFullThreadDevice(), OT_NOOP);
+
+#if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
+    // When multicast subscription changes, fifferent from 1.2 SED, 1.2 MED would not trigger
```
simonlingoogle, line 4500:
```suggestion
    // When multicast subscription changes, different from 1.2 SED, 1.2 MED would not trigger
```

src/core/thread/mle.cpp
librasungirl: One note is that specification suggests 1.2 MED only should register MA with scope larger than 3 when it attaches to 1.2 parent and rely on its parent to do MLR. 

For my understanding, the intention of such design itself is to reduce the unnecessary CUR. And this PR follows the same design principle.

I would personally think there is chance for the spec to distinguish MED and SED when mentioning `MTD` in the paragraph quoted by @simonlingoogle .
Even if BHC-612 is rejected, it would be easier to reduce the code to follow the spec.

If no strong objection, I personally tend to keep this PR as is.

src/core/thread/mle.cpp
```diff
@@ -1446,24 +1448,40 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
...
     }
 
-    // For sleepy end device, register external multicast addresses to the parent for indirect transmission
-    if (!IsRxOnWhenIdle())
+    // Append external multicast address.
+    // For sleepy end device, register external multicast addresses to the parent for indirect transmission.
+    // Since Thread 1.2, non-sleepy MED would depend on its Thread 1.2 parent as the proxy to perform
+    // Multicast Listener Report, it registers external multicast addresses of scope larger than realm
+    // local as well.
+    while (Get<ThreadNetif>().GetNextExternalMulticast(iterator, address) == OT_ERROR_NONE)
```
simonlingoogle, line 1456: We can use `ShouldRegisterMulticastToParent` to further simplify this code make it it more readable. 
```suggestion

    VerifyOrExit(ShouldRegisterMulticastToParent(), OT_NOOP);

    while (Get<ThreadNetif>().GetNextExternalMulticast(iterator, address) == OT_ERROR_NONE)
    {
#if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
        if (IsRxOnWhenIdle() && !address.IsMulticastLargerThanRealmLocal())
        {
            // For Thread 1.2 MED, skip multicast address with scope not larger than realm local when registering.
            continue;
        }
#endif

        entry.SetUncompressed();
        entry.SetIp6Address(address);
        SuccessOrExit(error = aMessage.Append(&entry, entry.GetLength()));
        length += entry.GetLength();

        counter++;
        // only continue to append if there is available entry.
        VerifyOrExit(counter < OPENTHREAD_CONFIG_MLE_IP_ADDRS_TO_REGISTER, OT_NOOP);
    }

#endif

```

librasungirl: Note that `Mle::AppendAddressRegistration()` itself is an indication that it is MTD end device, no need to perform the below check in `Mle::ShouldRegisterMulticastToParent()`

https://github.com/openthread/openthread/blob/210223e9af2e748f41c72ec5cf7de6bf2b0fb1e0/src/core/thread/mle.cpp#L4493-L4497

When it goes to here, the only decision to make is whether or not to append the multicast address. As for readability, I think per situation is well commented in existing code, which seems to me more clear to understand when reading.

https://github.com/openthread/openthread/blob/210223e9af2e748f41c72ec5cf7de6bf2b0fb1e0/src/core/thread/mle.cpp#L1458-L1474

Prefer to keep it as it.

simonlingoogle: > Note that `Mle::AppendAddressRegistration()` itself is an indication that it is MTD end device, no need to perform the below check in `Mle::ShouldRegisterMulticastToParent()`

Yes, `Mle::ShouldRegisterMulticastToParent()` do some redundant checks here. 

> For `Mle::AppendAddressRegistration()`, when it goes here, the only decision to make is whether or not to append the multicast address. As for readability, I think per situation is well commented in existing code, which seems to me more clear to understand when reading.
> 
> https://github.com/openthread/openthread/blob/210223e9af2e748f41c72ec5cf7de6bf2b0fb1e0/src/core/thread/mle.cpp#L1458-L1474
> 

The code here does same checks of `Mle::ShouldRegisterMulticastToParent()`:
- Skip 1.1 MED
- Skip 1.2 MED with 1.1 parent

So, I would suggest to use `Mle::ShouldRegisterMulticastToParent()` for better code reuse. 

Besides, after `VerifyOrExit(ShouldRegisterMulticastToParent(), OT_NOOP);`, we are facing only 2 conditions:
1. SED
2. 1.2 MED with 1.2 Parent
This can help code readers easier to understand the code?

librasungirl: Personally prefer existing code a bit, per condition here well matches the proceeding summary comment when appending external multicast address - which seems to me more friendly to understand.

https://github.com/openthread/openthread/blob/210223e9af2e748f41c72ec5cf7de6bf2b0fb1e0/src/core/thread/mle.cpp#L1451-L1455

simonlingoogle: Fine with current code since it's only personal preference.

src/core/thread/mle.cpp
```diff
@@ -4469,5 +4487,27 @@ void Mle::UpdateAllDomainBackboneRouters(BackboneRouter::Leader::DomainPrefixSta
...
+bool Mle::ShouldRegisterMulticastToParent(void) const
+{
+    bool update = false;
+
+    VerifyOrExit(IsChild() && !IsFullThreadDevice(), OT_NOOP);
+
+#if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
+    // When multicast subscription changes, different from 1.2 SED, 1.2 MED would not trigger
+    // Child Update Request if the parent is Thread 1.1.
+    VerifyOrExit(!(GetParent().IsThreadVersion1_1()) || !IsRxOnWhenIdle(), OT_NOOP);
```
jwhui, line 4502:
```suggestion
    VerifyOrExit(!GetParent().IsThreadVersion1_1() || !IsRxOnWhenIdle(), OT_NOOP);
```

src/core/thread/mle.cpp
```diff
@@ -1434,24 +1436,40 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
...
+            }
+            else if (!address.IsMulticastLargerThanRealmLocal())
+            {
+                // For Thread 1.2 MED, skip multicast address with scope not larger than realm local when registering.
+                continue;
+            }
+#else
+            // For Thread 1.1 MED, skip any multicast address.
+            break;
+#endif
```
abtink, line 1462: Can we consider moving some the common checks outside of the `while` loop.

I think something like this may work:
```
    // Append external multicast addresses.  For sleepy end device,
    // register all external multicast addresses with the parent for
    // indirect transmission. Since Thread 1.2, non-sleepy MED should
    // also register external multicast addresses of scope larger than
    // realm with a 1.2 or higher parent.

    if (!IsRxOnWhenIdle()  
    #if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
        || !GetParent().IsThreadVersion1_1()
    #endif
        )
    {
        while (Get<ThreadNetif>().GetNextExternalMulticast(iterator, address) == OT_ERROR_NONE)
        {
    #if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
            // For Thread 1.2 MED, skip multicast address with scope not
            // larger than realm local when registering
            if (IsRxOnWhenIdle() && !address.IsMulticastLargerThanRealmLocal())
            {
                continue;
            }
    #endif  

        ...

        }
    }
```

src/core/thread/mle.cpp
```diff
@@ -4446,5 +4464,27 @@ void Mle::UpdateAllDomainBackboneRouters(BackboneRouter::Leader::DomainPrefixSta
 }
 #endif // OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
 
+// Since Thread 1.2, MTD end device (including rx-on-when-idle or rx-off-when-idle) should
+// register or de-register their multicast address larger than realm local to its Thread 1.2
+// parent via MLE Child Update Request.
+bool Mle::ShouldRegisterMulticastToParent(void) const
+{
+    bool update = false;
```
abtink, line 4472: How about this:
(under 1.2 we are practically adding one more check for non-sleepy cases that parent is 1.2 or higher).
```
     return IsChild() && !IsFullThreadDevice() &&
           (!IsRxOnWhenIdle()
#if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
            || !GetParent().IsThreadVersion1_1()
#endif
           ); 
```

Also since we seem to be using `ShouldRegisterMultciast` only in one place, we can consider inlining the same check in the code and remve `ShouldRegister` method. Thoughts?

src/core/thread/mle.cpp
```diff
@@ -4446,5 +4464,27 @@ void Mle::UpdateAllDomainBackboneRouters(BackboneRouter::Leader::DomainPrefixSta
 }
 #endif // OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
 
+// Since Thread 1.2, MTD end device (including rx-on-when-idle or rx-off-when-idle) should
+// register or de-register their multicast address larger than realm local to its Thread 1.2
```
abtink, line 4468:
```suggestion
// register or de-register their multicast address larger than realm local with its Thread 1.2
```

src/core/thread/topology.hpp
```diff
@@ -351,6 +351,14 @@ class Neighbor : public InstanceLocatorInit
      */
     void SetRloc16(uint16_t aRloc16) { mRloc16 = aRloc16; }
 
+    /**
+     * This method indicates whether or not it is a valid Thread 1.1 neighbor.
+     *
+     * @returns TRUE if it is a valid Thread 1.1 neighbor, FALSE otherwise.
+     *
+     */
+    bool IsThreadVersion1_1(void) const { return mState != kStateInvalid && mVersion == OT_THREAD_VERSION_1_1; }
```
abtink, line 360: General point here to discuss and consider our naming style and how to handle numbers and whether we want to use underscoer `_` in camel-case style names.
- I personally favor avoiding `_` (as much as possible) in any camel case style name.
- We can consider `IsThreadVersion1d1`  ("one dot one") or `IsThreadVersion1p1` ("one point one").
  -  We use something similar for "2.4GHz" in `OPENTHREAD_CONFIG_RADIO_2P4GHZ_OQPSK_SUPPORT`.

Thoughts?

src/core/thread/mle.cpp
librasungirl: no strong preference. update as your suggestion.

src/core/thread/mle.cpp
librasungirl: SGTM. Removed.

src/core/thread/topology.hpp
librasungirl: I am fine with `IsThreadVersion1p1`. Updated.

PR #4840, opened 2020-04-16
[random] use crypto RNG in places where non-trival random is required

jwhui:


abtink:
One thing to consider is that since with this change we would use the `Crypto` API from the `Instance` and related constructors in more places, on some platforms the entropy source may not yet be ready to provide a lot of TRNG.  I don't know if we will face this but if we do, I would say this should probably be addressed at the platform layer (e.g., by ensuring that we wait for entropy source from the underlying calls when required). 

Thoughts?

jwhui:
> One thing to consider is that since with this change we would use the `Crypto` API from the `Instance` and related constructors in more places, on some platforms the entropy source may not yet be ready to provide a lot of TRNG. I don't know if we will face this but if we do, I would say this should probably be addressed at the platform layer (e.g., by ensuring that we wait for entropy source from the underlying calls when required).

@abtink, I agree that the platform should ensure there is sufficient random output from the entropy source. At the same time, it is probably useful to move things that can fail out of the constructor so that we can reasonably return those errors from `otInstanceInit()`. Probably a task for a separate PR. Thoughts?

abtink:
> @abtink, I agree that the platform should ensure there is sufficient random output from the entropy source. At the same time, it is probably useful to move things that can fail out of the constructor so that we can reasonably return those errors from otInstanceInit(). Probably a task for a separate PR. Thoughts?

@jwhui  SGTM.

PR #4847, opened 2020-04-17
[github-actions] add test to verify OTNS works

simonlingoogle:
This PR imports unit tests from OTNS.

### DIFF
--- a/.github/workflows/simulation.yml
+++ b/.github/workflows/simulation.yml
@@ -62,6 +62,34 @@ jobs:
         ./bootstrap
         make -f examples/Makefile-simulation distcheck
 
+  cli-ftd-otns:
+    runs-on: ubuntu-18.04
+    env:
+      CFLAGS: -m32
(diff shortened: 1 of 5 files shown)


.github/workflows/simulation.yml
```diff
@@ -62,6 +62,34 @@ jobs:
...
+        sudo apt update
+        sudo apt --no-install-recommends install -y g++-multilib
+        ./bootstrap
+    - name: Run OTNS Tests
+      run: |
+        git clone https://github.com/openthread/ot-ns.git --depth 1 --branch master ~/otns
+        export OTDIR=$PWD
+        cd ~/otns
+        ./script/test py-unittests py-examples
+    - name: Codecov
```
jwhui, line 90: Should we also increase the value of `after_n_builds`? https://github.com/openthread/openthread/blob/fa96bcd10fef0e4ff9340bb331a5c7b10b96addf/.codecov.yml#L12

PR #4848, opened 2020-04-17
[logging] fix compile error for BBR region

simonlingoogle:
There is no `aFormat` parameter any more.

### DIFF
--- a/src/core/common/logging.hpp
+++ b/src/core/common/logging.hpp
@@ -470,11 +470,11 @@ extern "C" {
  *
  */
 #if OPENTHREAD_CONFIG_LOG_BBR == 1
-#define otLogCritBbr(...) otLogCrit(OT_LOG_REGION_BBR, _OT_REGION_BBR_PREFIX aFormat __VA_ARGS__)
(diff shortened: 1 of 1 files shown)


jwhui:
@simonlingoogle, thanks for fixing this. Can we also add a continuous-integration build that includes BBR and full logs?

simonlingoogle:
> @simonlingoogle, thanks for fixing this. Can we also add a continuous-integration build that includes BBR and full logs?

I am on it actually. I am about to create another PR for the full log build test. Do you think it should be included in this PR?

jwhui:
> I am on it actually. I am about to create another PR for the full log build test. Do you think it should be included in this PR?

I don't have a preference on whether it's in this PR or a fast-follow separate PR. Just want to make sure we prevent any build failures due to Thread 1.2 with logging going forward.

simonlingoogle:
> I don't have a preference on whether it's in this PR or a fast-follow separate PR. Just want to make sure we prevent any build failures due to Thread 1.2 with logging going forward.

I will make this PR just a simple fix for the compile error.

PR #4850, opened 2020-04-17
[tests] add Thread 1.2 all features build test

simonlingoogle:
with or without full logs. 

**This PR can not pass checks before #4848 is merged.**

### DIFF
--- a/script/check-simulation-build-autotools
+++ b/script/check-simulation-build-autotools
@@ -83,13 +83,25 @@ build_all_features()
         "-DOPENTHREAD_CONFIG_UDP_FORWARD_ENABLE=1"
     )
 
+    local options_1_2=(
+        "-DOPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE=1"
+    )
+
     export CPPFLAGS="${options[*]} -DOPENTHREAD_CONFIG_LOG_LEVEL=OT_LOG_LEVEL_NONE"
     reset_source
     make -f examples/Makefile-simulation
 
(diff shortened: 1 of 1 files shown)


jwhui:
> And `PROTOCOL_VERSION` is `2` defined in `config.py`. So when we use Thread 1.2 device, this would fail. As we are moving forward to 1.2, shall we support all these original cases for 1.2?

Is checking that the Version value is set to 2 a part of the Thread Test Specification for Thread 1.1?

simonlingoogle:
> As we are moving forward to 1.2, shall we support all these original cases for 1.2?

I think we need to make sure that 1.2 pass all 1.1 test cases.

simonlingoogle:
> I think with Thread version 1.2, not all functional tests can be passed.

BTW, this PR is just for adding build tests for Thread 1.2, not for functional tests.

Irving-cl:
> > And `PROTOCOL_VERSION` is `2` defined in `config.py`. So when we use Thread 1.2 device, this would fail. As we are moving forward to 1.2, shall we support all these original cases for 1.2?
> 
> Is checking that the Version value is set to 2 a part of the Thread Test Specification for Thread 1.1?

Yes, it's a part of the hread Test Specification for Thread 1.1.

Irving-cl:
> > I think with Thread version 1.2, not all functional tests can be passed.
> 
> BTW, this PR is just for adding build tests for Thread 1.2, not for functional tests.

Ok then. So the issue I proposed won't affect this PR. But we'll need to solve it to introduce 1.2.

PR #4854, opened 2020-04-17
 [dua] Domain Unicast Address generation

librasungirl:
This is one of the PR series for DUA features. PRs for non-volatile storing DADCounter into non-volatile memory, prioritizing DUA in AR Tlv, allowing manual DUA address configuration, DUA registration etc will come afterwards. Please only focus on simplified subject this PR serves when review. Thanks.

As DUA features require non-volatile storage for DADCounter to ensure the consistency of Domain Unicast Address generated, in this PR

- The first commit updates existing `Utils:: Slaac:: GenerateIid()` to allow customized Network_ID and DadCounter maintained by the other components.

- The second commit introduces the DUA feature flag and DUA generation, and update existing test script to verify the communication via DUA.

 _Note that `aNetworkID` would be the domain name for CCM network, empty for Thread 1.2_


src/core/thread/mle.cpp
```diff
@@ -4418,6 +4429,53 @@ void Mle::DelayedResponseMetadata::RemoveFrom(Message &aMessage) const
...
+        mDomainUnicastAddress.GetAddress().SetPrefix(prefix->mPrefix.mFields.m8, prefix->mLength);
+        mDomainUnicastAddress.GetAddress().SetIid(address.GetIid());
+        mDomainUnicastAddress.mPrefixLength = prefix->mLength;
+
+        Get<ThreadNetif>().AddUnicastAddress(mDomainUnicastAddress);
+    }
+    else if (aState == BackboneRouter::Leader::kDomainPrefixAdded)
+    {
+        mDomainUnicastAddress.GetAddress().SetPrefix(prefix->mPrefix.mFields.m8, prefix->mLength);
+        mDomainUnicastAddress.mPrefixLength = prefix->mLength;
```
simonlingoogle, line 4461: Set `mDomainUnicastAddress.mPrefixLength` after SLAAC succeed ?

src/core/thread/mle.cpp
```diff
@@ -4418,6 +4429,53 @@ void Mle::DelayedResponseMetadata::RemoveFrom(Message &aMessage) const
...
+
+        if (Get<Utils::Slaac>().GenerateIid(mDomainUnicastAddress, NULL, 0, &mDadCounter,
+                                            false /* not allow random IID if no valid IID */))
+        {
+            Get<ThreadNetif>().AddUnicastAddress(mDomainUnicastAddress);
+        }
+        else
+        {
+            otLogWarnMle("Failed to generate valid DUA");
```
simonlingoogle, line 4470: In this case, `mDomainUnicastAddress` does not have valid `IID`.

src/core/thread/mle.cpp
```diff
@@ -4418,6 +4429,53 @@ void Mle::DelayedResponseMetadata::RemoveFrom(Message &aMessage) const
...
+
+    VerifyOrExit(prefix != NULL, OT_NOOP);
+
+    if (aState == BackboneRouter::Leader::kDomainPrefixRefreshed)
+    {
+        Ip6::Address address = mDomainUnicastAddress.GetAddress();
+
+        mDomainUnicastAddress.GetAddress().Clear();
+        mDomainUnicastAddress.GetAddress().SetPrefix(prefix->mPrefix.mFields.m8, prefix->mLength);
+        mDomainUnicastAddress.GetAddress().SetIid(address.GetIid());
```
simonlingoogle, line 4453: Seems we can't assure that `address` has valid `IID` here because in `kDomainPrefixAdded` state, SLAAC might fail.

examples/common-switches.mk
```diff
@@ -47,6 +47,7 @@ DIAGNOSTIC          ?= 0
 DISABLE_DOC         ?= 0
 DISABLE_TOOLS       ?= 0
 DNS_CLIENT          ?= 0
+DUA_REGISTRATION    ?= 0
```
simonlingoogle, line 50: Enable this by default for Makefile-simulation ?

src/core/utils/slaac_address.hpp
```diff
@@ -120,6 +120,29 @@ class Slaac : public InstanceLocator
...
 
+    /**
+     * This methods generates the IID of an IPv6 address.
+     *
+     * @param[inout]  aAddress        A reference to the address that will be filled with the IID generated.
+     *                                Note the prefix of the address was already filled and will be used to
+     *                                generate the IID.
+     * @param[in]  aNetworkId         A pointer to a byte array of Network_ID to generate IID.
+     * @param[in]  aNetworkIdLength   The size of array @p aNetworkId.
+     * @param[inout]  aDadCounter     A pointer to the DAD_Counter that is employed to resolve Duplicate
```
abtink, line 131: nit: align the `a{ArgName}`?

src/core/thread/mle.cpp
```diff
@@ -4418,6 +4429,53 @@ void Mle::DelayedResponseMetadata::RemoveFrom(Message &aMessage) const
...
+        mDomainUnicastAddress.mPrefixLength = prefix->mLength;
+
+        Get<ThreadNetif>().AddUnicastAddress(mDomainUnicastAddress);
```
abtink, line 4456: Do we need to remove it before adding it here in "refreshed case"

abtink: We also need to somehow track if the "add case" did add it before or not? e.g., can `clear()` the address to indicate this.

src/core/thread/mle.cpp
abtink, line 4453: Can we have smaller than 64 bit prefix length?

src/core/utils/slaac_address.cpp
```diff
@@ -262,47 +272,68 @@ void Slaac::GenerateIid(Ip6::NetifUnicastAddress &aAddress) const
...
     }
 
     otLogWarnUtil("SLAAC: Failed to generate a non-reserved IID after %d attempts", dadCounter);
-    Random::Crypto::FillBuffer(hash, Ip6::Address::kInterfaceIdentifierSize);
-    aAddress.GetAddress().SetIid(&hash[0]);
+
+    if (aAllowLastRandomIid)
+    {
+        Random::Crypto::FillBuffer(hash, Ip6::Address::kInterfaceIdentifierSize);
+        aAddress.GetAddress().SetIid(&hash[0]);
```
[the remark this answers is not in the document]
librasungirl: might be nice to have. However this PR doesn't want to touch unnecessary logical updates.

src/core/utils/slaac_address.cpp
```diff
@@ -262,47 +272,68 @@ void Slaac::GenerateIid(Ip6::NetifUnicastAddress &aAddress) const
...
     uint8_t        hash[Crypto::Sha256::kHashSize];
+    bool           validIid = false;
 
     OT_STATIC_ASSERT(sizeof(hash) >= Ip6::Address::kInterfaceIdentifierSize,
                      "SHA-256 hash size is too small to use as IPv6 address IID");
 
     GetIidSecretKey(secretKey);
 
-    for (dadCounter = 0; dadCounter < kMaxIidCreationAttempts; dadCounter++)
+    for (; dadCounter < kMaxIidCreationAttempts; dadCounter++)
```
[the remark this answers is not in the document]
librasungirl: updated.

src/core/thread/mle.cpp
librasungirl: use `mValid` to indicate whether or not there is available valid DUA.

src/core/utils/slaac_address.cpp
```diff
@@ -262,47 +272,83 @@ void Slaac::GenerateIid(Ip6::NetifUnicastAddress &aAddress) const
...
+        {
+            *aDadCounter = dadCounter;
+        }
 
-        VerifyOrExit(aAddress.GetAddress().IsIidReserved(), OT_NOOP);
+        // Exit and return the address if the IID is not reserved,
+        ExitNow(error = OT_ERROR_NONE);
     }
 
     otLogWarnUtil("SLAAC: Failed to generate a non-reserved IID after %d attempts", dadCounter);
```
simonlingoogle, line 341: Strictly speaking, `dadCounter` no longer equals to the number of attempts. The number of attempts should always equal to `kMaxIidCreationAttempts`?

src/core/thread/mle.cpp
```diff
@@ -4418,6 +4428,66 @@ void Mle::DelayedResponseMetadata::RemoveFrom(Message &aMessage) const
...
+        }
+    }
+
+    if (aState == BackboneRouter::Leader::kDomainPrefixAdded || attempt)
+    {
+        mDomainUnicastAddress.GetAddress().Clear();
+        mDomainUnicastAddress.GetAddress().SetPrefix(prefix->mPrefix.mFields.m8, prefix->mLength);
+
+        if (Get<Utils::Slaac>().GenerateIid(mDomainUnicastAddress, NULL, 0, &mDadCounter,
+                                            false /* not allow random IID if no valid IID */) == OT_ERROR_NONE)
```
simonlingoogle, line 4474: any background on why not allow random IID ?

src/core/utils/slaac_address.cpp
```diff
@@ -262,47 +272,83 @@ void Slaac::GenerateIid(Ip6::NetifUnicastAddress &aAddress) const
...
+        dadCounter = *aDadCounter;
+
+        if ((UINT16_MAX - kMaxIidCreationAttempts) >= dadCounter)
+        {
+            endDadCounter = UINT16_MAX;
+        }
+        else
+        {
+            endDadCounter = dadCounter + kMaxIidCreationAttempts;
+        }
```
simonlingoogle, line 304:
```suggestion
```
see comment below

src/core/thread/mle.cpp
librasungirl: now mle component owns mDadCounter, which is used for Duplicate detection, I think mle should control and ensure the relationship of IID and mDadCounter.

src/core/utils/slaac_address.cpp
librasungirl: good catch. will update.

src/core/utils/slaac_address.cpp
```diff
@@ -262,47 +272,83 @@ void Slaac::GenerateIid(Ip6::NetifUnicastAddress &aAddress) const
...
+    }
+
     GetIidSecretKey(secretKey);
 
-    for (dadCounter = 0; dadCounter < kMaxIidCreationAttempts; dadCounter++)
+    for (; dadCounter < endDadCounter; dadCounter++)
```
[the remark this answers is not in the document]
librasungirl: update in a different way with a local variable option. PTAL.

src/core/thread/mle.cpp
simonlingoogle: `SLAAC` only generates random IID after failed using `mDadCounter` counter. In such extreme situation, would it better to just use a random IID than to not having DUA?

src/core/utils/slaac_address.cpp
```diff
@@ -280,29 +291,49 @@ void Slaac::GenerateIid(Ip6::NetifUnicastAddress &aAddress) const
...
+        }
+
+        // Exit and return the address if the IID is not reserved,
+        ExitNow(error = OT_ERROR_NONE);
     }
 
-    otLogWarnUtil("SLAAC: Failed to generate a non-reserved IID after %d attempts", dadCounter);
-    Random::Crypto::FillBuffer(hash, Ip6::Address::kInterfaceIdentifierSize);
-    aAddress.GetAddress().SetIid(&hash[0]);
+    otLogWarnUtil("SLAAC: Failed to generate a non-reserved IID after %d attempts", kMaxIidCreationAttempts);
```
simonlingoogle, line 326: int: Include the value of `dadCounter` in the log for more information?

src/core/thread/mle.cpp
librasungirl: 1.2 Specification 5.23.1 says

> Domain Unicast Address (DUA), an IPv6 address for a Thread
> Device that remains stable even after roaming to a new Thread Network configured with the same Thread Domain information

I think the determinacy is expected more than a random one which is unpredictable. Besides, note that existing SLAAC utils tries up to 1024 to avoid the Reserved IID which is just a limited sets

https://github.com/openthread/openthread/blob/master/src/core/net/ip6_address.cpp#L158-L161

I would take a random one to be a `nice to have` feature in some extreme situations that all the IID generated by all the `1024` generates are reserved ones, and the random IID could stored to ensure the consistency when roaming.

This PR might not be so perfection to handle all the possible scenarios in some view, however could we focus on the majority scenarios when review and leave enhancements on some possible extreme scenarios in the future?

simonlingoogle: Fine to not use random IID since the situation hardly ever happen.

src/core/thread/mle.cpp
```diff
@@ -4418,6 +4428,66 @@ void Mle::DelayedResponseMetadata::RemoveFrom(Message &aMessage) const
     OT_UNUSED_VARIABLE(error);
 }
 
+#if OPENTHREAD_CONFIG_TMF_DUA_REGISTRATION_ENABLE
+void Mle::UpdateDomainUnicastAddress(BackboneRouter::Leader::DomainPrefixState aState)
```
jwhui, line 4432: Why is this logic in MLE? Can we put Domain Unicast Address management in a separate source file? Maybe in `backbone_router` given that Domains are related to BBRs?

src/core/config/openthread-core-config-check.h
```diff
@@ -476,4 +476,8 @@
 
 #endif // OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
 
+#if OPENTHREAD_CONFIG_TMF_DUA_REGISTRATION_ENABLE && (OPENTHREAD_THREAD_VERSION < OPENTHREAD_THREAD_VERSION_1_2)
+#error "At least Thread Version 1.2 is required for OPENTHREAD_CONFIG_TMF_DUA_REGISTRATION_ENABLE"
```
abtink, line 480:
```suggestion
#error "Thread Version 1.2 or later is required for OPENTHREAD_CONFIG_TMF_DUA_REGISTRATION_ENABLE"
```

src/core/config/tmf.h
```diff
@@ -172,4 +172,14 @@
 #define OPENTHREAD_CONFIG_TMF_NETWORK_DIAG_MTD_ENABLE 0
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_TMF_DUA_REGISTRATION_ENABLE
+ *
+ * Define as 1 to support Thread 1.2 Domain Unicast Address Registration features.
```
abtink, line 178:
```suggestion
 * Define as 1 to support Thread 1.2 Domain Unicast Address Registration feature.
```

etc/cmake/options.cmake
```diff
@@ -114,6 +114,11 @@ if(OT_ECDSA)
     list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_CONFIG_ECDSA_ENABLE=1")
 endif()
 
+option(OT_DUA_REGISTRATION "enable domain unicast address registration feature for Thread 1.2")
```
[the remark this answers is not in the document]
librasungirl: apply this suggestion along with the adjustment of the location to host address

script/test
```diff
@@ -121,6 +127,12 @@ build_posix()
         "-DOT_THREAD_VERSION=${version}"
     )
 
+    if [[ "${version}" == "1.2" ]]; then
+        options+=(
+        "-DOT_DUA_REGISTRATION=ON"
```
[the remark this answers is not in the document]
librasungirl: updated.

script/test
```diff
@@ -78,6 +78,12 @@ build_simulation()
...
```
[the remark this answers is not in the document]
librasungirl: updated.

src/core/config/openthread-core-config-check.h
librasungirl: Updated.

src/core/config/tmf.h
librasungirl: Applied.

src/core/thread/mle.cpp
librasungirl: Moved to the separate dua_manager module as discussed. Please review.

librasungirl:
Pushed the third commit to remove the reuse of DUA IID when domain prefix changes, with reasons as below:

Though reuse may save one `GenerateIid()` computation, however since we are not planning to store the iid so far, if the device reboots, a different iid would be generated as the domain prefix is different from what was used to generate the iid.

Since the consistency of DUA before/after reboot are always preferred, and no matter reuse or not, DUA registration is required when domain prefix changes, now I prefer to simply regenerate the DUA when domain prefix changes.

Thoughts?  @jwhui  @abtink @simonlingoogle

src/core/thread/dua_manager.cpp
```diff
@@ -0,0 +1,99 @@
...
+    prefix = Get<BackboneRouter::Leader>().GetDomainPrefix();
+
+    OT_ASSERT(prefix != NULL);
+
+    mDomainUnicastAddress.GetAddress().Clear();
+    mDomainUnicastAddress.GetAddress().SetPrefix(prefix->mPrefix.mFields.m8, prefix->mLength);
+    mDomainUnicastAddress.mPrefixLength = prefix->mLength;
+
+    if (Get<Utils::Slaac>().GenerateIid(mDomainUnicastAddress, NULL, 0, &mDadCounter,
+                                        false /* not allow random IID if no valid IID */) == OT_ERROR_NONE)
```
simonlingoogle, line 82: nit:
```suggestion
                                        /* aAllowLastRandomIid */ false) == OT_ERROR_NONE)
```

src/core/thread/dua_manager.cpp
```diff
@@ -0,0 +1,99 @@
...
+    mDomainUnicastAddress.mPreferred          = true;
+    mDomainUnicastAddress.mScopeOverride      = Ip6::Address::kGlobalScope;
+    mDomainUnicastAddress.mScopeOverrideValid = true;
+}
+
+void DuaManager::UpdateDomainUnicastAddress(BackboneRouter::Leader::DomainPrefixState aState)
+{
+    const otIp6Prefix *prefix;
+
+    if (aState == BackboneRouter::Leader::kDomainPrefixRemoved ||
```
jwhui, line 63:
```suggestion
    if ((aState == BackboneRouter::Leader::kDomainPrefixRemoved) ||
```
Be consistent with the parens.

src/core/thread/dua_manager.cpp
```diff
@@ -0,0 +1,99 @@
...
+        (aState == BackboneRouter::Leader::kDomainPrefixRefreshed))
+    {
+        Get<ThreadNetif>().RemoveUnicastAddress(mDomainUnicastAddress);
+    }
+
+    VerifyOrExit(aState == BackboneRouter::Leader::kDomainPrefixAdded ||
```
jwhui, line 69:
```suggestion
    VerifyOrExit((aState == BackboneRouter::Leader::kDomainPrefixAdded) ||
```

src/core/utils/slaac_address.cpp
```diff
@@ -280,29 +291,49 @@ void Slaac::GenerateIid(Ip6::NetifUnicastAddress &aAddress) const
...
+
+    if (aAllowLastRandomIid)
+    {
+        Random::Crypto::FillBuffer(hash, Ip6::Address::kInterfaceIdentifierSize);
```
jwhui, line 330: [RFC 7217](https://tools.ietf.org/html/rfc7217) states: "In the event that a DAD conflict cannot be solved (possibly after trying a number of different addresses), address configuration would fail." Should we really continue to support this?

src/core/thread/dua_manager.cpp
```diff
@@ -0,0 +1,99 @@
...
+    prefix = Get<BackboneRouter::Leader>().GetDomainPrefix();
+
+    OT_ASSERT(prefix != NULL);
+
+    mDomainUnicastAddress.GetAddress().Clear();
+    mDomainUnicastAddress.GetAddress().SetPrefix(prefix->mPrefix.mFields.m8, prefix->mLength);
+    mDomainUnicastAddress.mPrefixLength = prefix->mLength;
+
+    if (Get<Utils::Slaac>().GenerateIid(mDomainUnicastAddress, NULL, 0, &mDadCounter,
+                                        /* aAllowLastRandomIid */ false) == OT_ERROR_NONE)
```
jwhui, line 82: nit: generally prefer named values to avoid the comment

src/core/thread/dua_manager.hpp
```diff
@@ -0,0 +1,92 @@
...
+     * This method updates Domain Unicast Address.
+     *
+     * @param[in]  aState  The Domain Prefix state or state change.
+     *
+     */
+    void UpdateDomainUnicastAddress(BackboneRouter::Leader::DomainPrefixState aState);
+
+private:
+    Ip6::NetifUnicastAddress mDomainUnicastAddress;
+    uint16_t                 mDadCounter;
```
jwhui, line 86: Just curious - why 16 bits? The likelihood of incrementing beyond 255 should be exceedingly low.

src/core/utils/slaac_address.cpp
librasungirl: This PR simply respects previous code which already uses random in the last attempt. 
Besides, in the loop, it checks whether the IID generated is reserved, is it also counted as `duplicate address`, my previous impression is the address (not iid) conflicts with the others on the device itself or in the device's network.

Thoughts?

src/core/thread/dua_manager.hpp
librasungirl: Agree. Here simply respects existing dad counter in SLAAC module, and generate iid operation may try 1024 which exceeds the range of one byte.

I am fine to update it to be one byte if you prefer.

Thoughts?

src/core/thread/dua_manager.cpp
librasungirl: not required now if the last random iid attempt is removed as in the last commit patch.

librasungirl:
Append a new commit to resolve your concern @jwhui , specifically

>     - remove the last random iid attempt
>     - update dad counter from two bytes to one byte
>     - change the max generate op attempts in slaac from 1024 to 256

Please take a review.

src/core/utils/slaac_address.cpp
```diff
@@ -291,7 +290,7 @@ otError Slaac::GenerateIid(Ip6::NetifUnicastAddress &aAddress,
 
     GetIidSecretKey(secretKey);
 
-    for (uint16_t count = 0; count < kMaxIidCreationAttempts; count++, dadCounter++)
+    for (uint8_t count = 0; count < kMaxIidCreationAttempts; count++, dadCounter++)
```
simonlingoogle, line 293:
```suggestion
    for (int count = 0; count < kMaxIidCreationAttempts; count++, dadCounter++)
```

`count < kMaxIidCreationAttempts` is always true ?

librasungirl: good catch 👍

src/core/CMakeLists.txt
```diff
@@ -176,6 +176,7 @@ set(COMMON_SOURCES
     thread/announce_begin_server.cpp
     thread/announce_sender.cpp
     thread/child_table.cpp
+    thread/dua_manager.cpp
```
abtink, line 179: Shall we also add this in `Andriod.mk` and `BUILD.gn`?

src/core/thread/dua_manager.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * @file
+ *   This file implements managing DUA.
+ */
+
+#include "dua_manager.hpp"
+
+#if OPENTHREAD_FTD || OPENTHREAD_MTD
+#if OPENTHREAD_CONFIG_DUA_ENABLE
```
abtink, line 37: How about combining all `#if` checks?
```suggestion
#if (OPENTHREAD_FTD || OPENTHREAD_MTD) && OPENTHREAD_CONFIG_DUA_ENABLE
```

src/core/thread/dua_manager.cpp
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+exit:
+    return;
+}
+
+} // namespace ot
+
+#endif // OPENTHREAD_CONFIG_DUA_ENABLE
+#endif // OPENTHREAD_FTD || OPENTHREAD_MTD
```
abtink, line 98:
```suggestion
#endif // (OPENTHREAD_FTD || OPENTHREAD_MTD) && OPENTHREAD_CONFIG_DUA_ENABLE
```

src/core/utils/slaac_address.hpp
```diff
@@ -120,10 +120,30 @@ class Slaac : public InstanceLocator
      */
     void SetFilter(otIp6SlaacPrefixFilter aFilter);
 
+    /**
+     * This methods generates the IID of an IPv6 address.
```
abtink, line 124:
```suggestion
     * This method generates the IID of an IPv6 address.
```

src/core/utils/slaac_address.hpp
```diff
@@ -120,10 +120,30 @@ class Slaac : public InstanceLocator
...
+     *
+     * @param[inout]  aAddress             A reference to the address that will be filled with the IID generated.
+     *                                     Note the prefix of the address was already filled and will be used to
```
abtink, line 127:
```suggestion
     *                                     Note the prefix of the address must already be filled and will be used to
```

src/core/CMakeLists.txt
librasungirl: added

src/core/thread/dua_manager.cpp
librasungirl: updated.

librasungirl:
Rebased and reorganized into two separate clean commits, without any logical changes. @jwhui

PR #4855, opened 2020-04-17
[network-data] fix length overflow in ServiceTlv::IsValid()

jwhui:


src/core/thread/network_data_tlvs.hpp
```diff
@@ -1012,10 +1012,11 @@ class ServiceTlv : public NetworkDataTlv
...
     bool IsValid(void) const
     {
-        uint8_t length = GetLength();
+        uint8_t length           = GetLength();
+        uint8_t enterpriseLength = IsThreadEnterprise() ? 0 : sizeof(uint32_t);
 
-        return (length >= sizeof(mFlagsServiceId)) &&
-               (length >= kMinLength + (IsThreadEnterprise() ? 0 : sizeof(uint32_t))) && (length >= GetFieldsLength());
+        return (length >= sizeof(mFlagsServiceId)) && (length >= kMinLength + enterpriseLength) &&
+               (length - enterpriseLength >= GetServiceDataLength()) && (length >= GetFieldsLength());
```
abtink, line 1019: Nice find @jwhui . Thanks.  👍 
This addresses possible overflow of `GetFieldsLength()` return value `uint8_t`.

This is fine as is.  I was thinking about this. What do you think of replacing the `length >= GetFiledsLength()` with this check:
```
(static_cast<uint16_t>(length) + sizeof(NetworkDataTlv) >= CalculateSize(GetEnterpriseNumber(), GetServiceDataLength()))`
```
The `CalculateSize()` does return `uint16_t`. 

I am also considering to change `NetworkDataTlv::GetSize()` to also return `uint16_t` (in a next PR)? Then we can do `GetSize() >= CalculateSize(...)`?

jwhui: SGTM. I'll keep this PR as is. We can improve in a future PR.

PR #4859, opened 2020-04-19
[cmake] add long routes options

simonlingoogle:
This PR adds `OT_MLE_LONG_ROUTES` option for cmake and `MLE_LONG_ROUTES` option for Makefile.


simonlingoogle:
More comments on this? @bukepo @jwhui

etc/cmake/options.cmake
```diff
@@ -161,6 +161,11 @@ if(OT_MAC_FILTER)
     list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_CONFIG_MAC_FILTER_ENABLE=1")
 endif()
 
+option(OT_MLE_LONG_ROUTES "enable MLE long routes extension")
```
jwhui, line 164: Can we add a note here that this breaks Thread conformance, something like:
```suggestion
option(OT_MLE_LONG_ROUTES "enable MLE long routes extension (experimental, breaks Thread conformance)")
```

examples/common-switches.mk
```diff
@@ -199,6 +200,10 @@ ifeq ($(MAC_FILTER),1)
 COMMONCFLAGS                   += -DOPENTHREAD_CONFIG_MAC_FILTER_ENABLE=1
 endif
 
+ifeq ($(MLE_LONG_ROUTES),1)
+COMMONCFLAGS                   += -DOPENTHREAD_CONFIG_MLE_LONG_ROUTES_ENABLE=1
```
jwhui, line 204: Add a note here that this breaks Thread conformance?

etc/cmake/options.cmake
simonlingoogle: Good idea. Will do.

examples/common-switches.mk
simonlingoogle: noted

etc/cmake/options.cmake
simonlingoogle: noted

PR #4860, opened 2020-04-19
[mle] fix MLE Router handling errors

### LINKED ISSUE #4508: Not stable Thread network topology in a big > 100 nodes network
Hi, first I must say it is very critical issue for us and I would need any urgent help if possible.

### The Setup:
We have a hardware setup of 50x nRF5280 + one NCP and a border router.
Currently we test them in a line topology:
`BR (NCP) --- NODE --- NODE --- NODE --- ... --- NODE`
Each link to a nearest neighbour is between 70 and 80 dbm, Nodes can see around 3-4 neighbour nodes (RSSI: 80-95).
All nodes send keep-alive confirmable coap to the NCP Mesh Local address every 1 minute.

### Observations:
Thread network topology becomes stable quite quickly (within single minutes),
Keep alives from all nodes reach border router which is fine.
Communication can be stable for hours with rare keep alives being lost.

When we push some messages into the network ( Mesh multicast + confirmable command to every node Mesh local address ) everything goes crazy.
Keep alives are all gone or almost all gone, we observed NCP parent router to change its RLOC (d000 ->bc00), looks like thread topology is changing/partitioning?.

Short higher traffic can break the network within seconds but it takes around 10 minutes to recover from that situation. After several minutes everything gets back to normal. 

### Notes
We didn't observe similar issue when topology wasn't linear (less hops).
We will provide more details ASAP like sniffer pcap, some test results, charts and anything that could help.

### Questions
Do you have any thoughts why this is happening?
Are there any timeouts we could tune to make it better?
Have you worked with similar hardware setups with many nodes and many hops?
@jwhui I would be greatful if you could look into the issue

simonlingoogle:
This PR tries to fix the [assertion failure](https://github.com/openthread/openthread/issues/4508#issuecomment-615899675) that occurred when simulating 8x8 nodes forming one partition by [OTNS form_partition.py](https://github.com/openthread/ot-ns/blob/master/pylibs/examples/form_partition.py). 

This bug was found when trying to resolve an issue in #4508.

### DIFF
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -2993,6 +2993,9 @@ otError Mle::HandleLeaderData(const Message &aMessage, const Ip6::MessageInfo &a
     uint16_t            pendingDatasetOffset = 0;
     bool                dataRequest          = false;
     Tlv                 tlv;
+    RouterIdSet         routerIdSet;
+
+    OT_UNUSED_VARIABLE(routerIdSet);
 
     // Leader Data
     SuccessOrExit(error = ReadLeaderData(aMessage, leaderData));
@@ -3004,6 +3007,16 @@ otError Mle::HandleLeaderData(const Message &aMessage, const Ip6::MessageInfo &a
         {
             SetLeaderData(leaderData.GetPartitionId(), leaderData.GetWeighting(), leaderData.GetLeaderRouterId());
             mRetrieveNewNetworkData = true;
+
+#if OPENTHREAD_FTD
+            if (IsFullThreadDevice())
+            {
+                routerIdSet.Clear();
+                routerIdSet.Add(GetLeaderId());
+                routerIdSet.Add(GetParent().GetRouterId());
+                Get<RouterTable>().UpdateRouterIdSet(0, routerIdSet);
+            }
+#endif
         }
         else
         {
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -2993,9 +2993,6 @@ otError Mle::HandleLeaderData(const Message &aMessage, const Ip6::MessageInfo &a
     uint16_t            pendingDatasetOffset = 0;
     bool                dataRequest          = false;
     Tlv                 tlv;
-    RouterIdSet         routerIdSet;
-
-    OT_UNUSED_VARIABLE(routerIdSet);
 
     // Leader Data
     SuccessOrExit(error = ReadLeaderData(aMessage, leaderData));
@@ -3005,18 +3002,14 @@ otError Mle::HandleLeaderData(const Message &aMessage, const Ip6::MessageInfo &a
     {
         if (IsChild())
         {
+#if OPENTHREAD_FTD
+            // An FTD skips handling LeaderData of a different partition.
+            VerifyOrExit(!IsFullThreadDevice() || (leaderData.GetPartitionId() == mLeaderData.GetPartitionId() &&
+                                                   leaderData.GetLeaderRouterId() == GetLeaderId()),
+                         error = OT_ERROR_DROP);
+#endif
             SetLeaderData(leaderData.GetPartitionId(), leaderData.GetWeighting(), leaderData.GetLeaderRouterId());
             mRetrieveNewNetworkData = true;
-
-#if OPENTHREAD_FTD
-            if (IsFullThreadDevice())
-            {
-                routerIdSet.Clear();
-                routerIdSet.Add(GetLeaderId());
-                routerIdSet.Add(GetParent().GetRouterId());
-                Get<RouterTable>().UpdateRouterIdSet(0, routerIdSet);
-            }
-#endif
         }
         else
         {
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -1284,6 +1284,7 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
         if (processRouteTlv)
         {
             SuccessOrExit(error = ProcessRouteTlv(route));
+            aNeighbor = NULL; // aNeighbor is no longer valid after `ProcessRouteTlv`
         }
     }
 
@@ -2237,6 +2238,7 @@ void MleRouter::HandleChildIdRequest(const Message &         aMessage,
     router = mRouterTable.GetRouter(macAddr);
     if (router != NULL)
     {
+        // The `router` here can be invalid
         RemoveNeighbor(*router);
     }
 
@@ -3330,6 +3332,8 @@ void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
     }
     else if (aNeighbor.IsStateValid())
     {
+        OT_ASSERT(mRouterTable.GetRouterIndex(static_cast<Router &>(aNeighbor)) < kMaxRouters);
+
         Signal(OT_NEIGHBOR_TABLE_EVENT_ROUTER_REMOVED, aNeighbor);
         mRouterTable.RemoveRouterLink(static_cast<Router &>(aNeighbor));
     }
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -3300,7 +3300,7 @@ void MleRouter::RemoveRouterLink(Router &aRouter)
 
 void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
 {
-    VerifyOrExit(!aNeighbor.IsStateInvalid());
+    VerifyOrExit(!aNeighbor.IsStateInvalid(), OT_NOOP);
 
     if (&aNeighbor == &mParent || &aNeighbor == &mParentCandidate)
     {
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -1284,7 +1284,10 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
         if (processRouteTlv)
         {
             SuccessOrExit(error = ProcessRouteTlv(route));
-            aNeighbor = NULL; // aNeighbor is no longer valid after `ProcessRouteTlv`
+            if (Get<RouterTable>().GetRouterIndex(static_cast<Router &>(*aNeighbor)) < kMaxRouters)
+            {
+                aNeighbor = NULL; // aNeighbor is no longer valid after `ProcessRouteTlv`
+            }
         }
     }
 
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -1423,7 +1423,7 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
     UpdateRoutes(route, routerId);
 
 exit:
-    if (sourceAddress != 0xFFFF && aNeighbor && aNeighbor->GetRloc16() != sourceAddress)
+    if (aNeighbor && aNeighbor->GetRloc16() != sourceAddress)
     {
         // Remove stale neighbors
         RemoveNeighbor(*aNeighbor);
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -1166,7 +1166,7 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
     const otThreadLinkInfo *linkInfo = static_cast<const otThreadLinkInfo *>(aMessageInfo.GetLinkInfo());
     uint8_t linkMargin = LinkQualityInfo::ConvertRssToLinkMargin(Get<Mac::Mac>().GetNoiseFloor(), linkInfo->mRss);
     Mac::ExtAddress macAddr;
-    uint16_t        sourceAddress;
+    uint16_t        sourceAddress = 0xFFFF;
     LeaderData      leaderData;
     RouteTlv        route;
     uint32_t        partitionId;
@@ -1423,7 +1423,7 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
     UpdateRoutes(route, routerId);
 
 exit:
-    if (aNeighbor && aNeighbor->GetRloc16() != sourceAddress)
+    if (sourceAddress != 0xFFFF && aNeighbor && aNeighbor->GetRloc16() != sourceAddress)
     {
         // Remove stale neighbors
         RemoveNeighbor(*aNeighbor);
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -3300,6 +3300,8 @@ void MleRouter::RemoveRouterLink(Router &aRouter)
 
 void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
 {
+    VerifyOrExit(!aNeighbor.IsStateInvalid());
+
     if (&aNeighbor == &mParent || &aNeighbor == &mParentCandidate)
     {
         if (IsChild())
@@ -3334,6 +3336,9 @@ void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
 
     aNeighbor.GetLinkInfo().Clear();
     aNeighbor.SetState(Neighbor::kStateInvalid);
+
+exit:
+    return;
 }
 
 Neighbor *MleRouter::GetNeighbor(uint16_t aAddress)
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -1166,7 +1166,7 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
     const otThreadLinkInfo *linkInfo = static_cast<const otThreadLinkInfo *>(aMessageInfo.GetLinkInfo());
     uint8_t linkMargin = LinkQualityInfo::ConvertRssToLinkMargin(Get<Mac::Mac>().GetNoiseFloor(), linkInfo->mRss);
     Mac::ExtAddress macAddr;
-    uint16_t        sourceAddress = 0xFFFF;
+    uint16_t        sourceAddress = Mac::kShortAddrInvalid;
     LeaderData      leaderData;
     RouteTlv        route;
     uint32_t        partitionId;
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -1284,7 +1284,7 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
         if (processRouteTlv)
         {
             SuccessOrExit(error = ProcessRouteTlv(route));
-            if (Get<RouterTable>().GetRouterIndex(static_cast<Router &>(*aNeighbor)) < kMaxRouters)
+            if (Get<RouterTable>().Contains(*aNeighbor))
             {
                 aNeighbor = NULL; // aNeighbor is no longer valid after `ProcessRouteTlv`
             }
@@ -3339,7 +3339,7 @@ void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
     }
     else if (aNeighbor.IsStateValid())
     {
-        OT_ASSERT(mRouterTable.GetRouterIndex(static_cast<Router &>(aNeighbor)) < kMaxRouters);
+        OT_ASSERT(mRouterTable.Contains(aNeighbor));
 
         Signal(OT_NEIGHBOR_TABLE_EVENT_ROUTER_REMOVED, aNeighbor);
         mRouterTable.RemoveRouterLink(static_cast<Router &>(aNeighbor));
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -3304,12 +3304,16 @@ void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
 {
     VerifyOrExit(!aNeighbor.IsStateInvalid(), OT_NOOP);
 
-    if (&aNeighbor == &mParent || &aNeighbor == &mParentCandidate)
+    if (&aNeighbor == &mParent)
     {
         if (IsChild())
         {
             IgnoreError(BecomeDetached());
         }
+    }
+    else if (&aNeighbor == &mParentCandidate)
+    {
+        mParentCandidate.Clear();
     }
     else if (!IsActiveRouter(aNeighbor.GetRloc16()))
     {
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -3300,7 +3300,7 @@ void MleRouter::RemoveRouterLink(Router &aRouter)
 
 void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
 {
-    if (&aNeighbor == &mParent)
+    if (&aNeighbor == &mParent || &aNeighbor == &mParentCandidate)
     {
         if (IsChild())
         {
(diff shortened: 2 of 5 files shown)


src/core/thread/mle.cpp
```diff
@@ -3073,6 +3076,16 @@ otError Mle::HandleLeaderData(const Message &aMessage, const Ip6::MessageInfo &a
...
             SetLeaderData(leaderData.GetPartitionId(), leaderData.GetWeighting(), leaderData.GetLeaderRouterId());
             mRetrieveNewNetworkData = true;
+
+#if OPENTHREAD_FTD
+            if (IsFullThreadDevice())
+            {
+                routerIdSet.Clear();
+                routerIdSet.Add(GetLeaderId());
+                routerIdSet.Add(GetParent().GetRouterId());
+                Get<RouterTable>().UpdateRouterIdSet(0, routerIdSet);
```
simonlingoogle, line 3086: It is possible that Child can receive from parent a `Data Response` with different partition ID (in LeaderData) from the Child's current LeaderData. The original code only `SetLeaderData`, but didn't update the `RouterTable`, which I think can lead to inconsistent states, more specifically, `Get<RouterTable>().GetLeader() == NULL`. 

Since there is no `RouteTLV` in `Data Response`, here the fix creates a minimal `RouterIdSet` and updates the RouterTable. 

Another fix is call `BecomeDetached()` here? Thoughts? @librasungirl @jwhui

src/core/thread/mle_router.cpp
```diff
@@ -3327,6 +3329,8 @@ void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
     }
     else if (!IsActiveRouter(aNeighbor.GetRloc16()))
     {
+        OT_ASSERT(mChildTable.GetChildIndex(static_cast<Child &>(aNeighbor)) < kMaxChildren);
```
simonlingoogle, line 3332: Make sure the neighbor is a valid child.

src/core/thread/mle_router.cpp
```diff
@@ -3318,7 +3318,9 @@ void MleRouter::RemoveRouterLink(Router &aRouter)
 
 void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
 {
-    if (&aNeighbor == &mParent)
+    VerifyOrExit(!aNeighbor.IsStateInvalid());
```
simonlingoogle, line 3321: It is possible that `aNeighbor` is invalid here when called from `MleRouter::HandleAdvertisement`: 
https://github.com/openthread/openthread/blob/af4837d122c9da0eb6dcb9f2a1b19b1e37b738cc/src/core/thread/mle_router.cpp#L1425-L1429

I think it might be a side effect of the fix of this PR in `Mle::HandleLeaderData`.

src/core/thread/mle_router.cpp
```diff
@@ -3318,7 +3318,9 @@ void MleRouter::RemoveRouterLink(Router &aRouter)
...
+
+    if (&aNeighbor == &mParent || &aNeighbor == &mParentCandidate)
```
simonlingoogle, line 3323: It is possible that `aNeighbor` is `mParentCandidate` here when called from `MleRouter::HandleAdvertisement`: 
https://github.com/openthread/openthread/blob/af4837d122c9da0eb6dcb9f2a1b19b1e37b738cc/src/core/thread/mle_router.cpp#L1425-L1429

Handle it same as the case `&aNeighbor == &mParent`.

src/core/thread/mle_router.cpp
```diff
@@ -3309,7 +3309,9 @@ void MleRouter::RemoveRouterLink(Router &aRouter)
 
 void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
 {
-    if (&aNeighbor == &mParent)
+    VerifyOrExit(!aNeighbor.IsStateInvalid(), OT_NOOP);
```
simonlingoogle, line 3312: An invalid neighbor with RLOC16 == 0xFFFF will be incorrectly identified as a `Child` in `mChildTable` in the original code.

src/core/thread/mle_router.cpp
```diff
@@ -3335,12 +3341,17 @@ void MleRouter::RemoveNeighbor(Neighbor &aNeighbor)
     }
     else if (aNeighbor.IsStateValid())
     {
+        OT_ASSERT(mRouterTable.GetRouterIndex(static_cast<Router &>(aNeighbor)) < kMaxRouters);
```
simonlingoogle, line 3344: Make sure the neighbor is a valid `Router`

src/core/thread/mle_router.cpp
simonlingoogle: Verified together with @librasungirl that this situation could happen in which a Router is attaching to a better parent (REED), but the REED changed it's RLOC16 during attachment and the REED also sent `MLE Advertisement`.

src/core/thread/mle_router.cpp
```diff
@@ -3309,15 +3311,23 @@ void MleRouter::RemoveRouterLink(Router &aRouter)
...
     if (&aNeighbor == &mParent)
     {
         if (IsChild())
         {
             BecomeDetached();
         }
     }
+    else if (&aNeighbor == &mParentCandidate)
+    {
+        mParentCandidate.Clear();
```
simonlingoogle, line 3325: How to properly handle this? Thoughts? @jwhui @librasungirl

src/core/thread/mle_router.cpp
```diff
@@ -1284,6 +1284,7 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
         if (processRouteTlv)
         {
             SuccessOrExit(error = ProcessRouteTlv(route));
+            aNeighbor = NULL; // aNeighbor is no longer valid after `ProcessRouteTlv`
```
abtink, line 1287: Can you please explain the reason for this?

 We have check with `aNeighbor` below for example at `exit` which would not happen (if we go through this code path). Do we know if this would be ok?
```
if (aNeighbor && aNeighbor->GetRloc16() != sourceAddress)
    {
        // Remove stale neighbors
        RemoveNeighbor(*aNeighbor);
    }
```

src/core/thread/mle_router.cpp
abtink: What situation did you have in mind here for this check (and the one for router)?

I think when we have `aNeighbor` pointer entry we have found/acquired it from either child table or router table.  Either case I think, it should be already in the related table with a valid index.

src/core/thread/mle_router.cpp
simonlingoogle: The original code has bug in it:

Let's assume that `aNeighbor` references to a `Router` instance in `mRouterTable`. 
- After `ProcessRouteTlv` is executed, `aNeighbor`  can point to a different `Router`, because `ProcessRouteTlv` can move Routers in `mRouterTable`. So, after `ProcessRouteTlv`, the `Router` instance referenced by `aNeighbor` is not what we expect it to be, the sender of the `MLE Advertisement`. So, any later handling of `aNeighbor` can be a potential bug. 
- If we are not lucky, `aNeighbor` now references to a `Router` instance with `RLOC16 == 0xFFFF` (Router ID not allocated). In such case, `RemoveNeighbor(*aNeighbor);` at the `exit` will be executed and the original `RemoveNeighbor` will treat this neighbor as a `Child` instance which it is not. 

One way of addressing it would be reassigning `aNeighbor` to the correct `Router`. Meanwhile assigning `aNeighbor` to `NULL` seems also OK to me according to analysis of the following code in this function.

src/core/thread/mle_router.cpp
simonlingoogle: We found two conditions that can incorrectly enter these two code blocks in the simulation. 

- `HandleAdvertisement` can incorrectly call `RemoveNeighbor` with `Router` from `mRouterTable`, which has `RLOC16 == 0xFFFF`. `RemoveNeighbor` will incorrectly handle this `aNeighbor` as a `Child`. 
- `HandleAdvertisement` can also call `RemoveNeighor` with `aNeighbor` being `mParentCandidate`. The original `RemoveNeighor` will incorrectly handle `aNeighbor` as `Child` or `Router` based on its `RLOC16`. 

In case there are more undiscovered bugs, I add these two `OT_ASSERT` statements for catching code bugs earlier.

src/core/thread/mle_router.cpp
abtink: Thanks.

simonlingoogle: This PR is to fix issues found in a 8x8 nodes simulation. The bugs are subtle and difficult to comprehend and unclear on how to handle. The current code changes are more for discussion than for fixing.

abtink: This is indeed a good find! 👍

src/core/thread/mle_router.cpp
```diff
@@ -1422,7 +1423,7 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
     UpdateRoutes(route, routerId);
 
 exit:
-    if (aNeighbor && aNeighbor->GetRloc16() != sourceAddress)
+    if (sourceAddress != 0xFFFF && aNeighbor && aNeighbor->GetRloc16() != sourceAddress)
```
simonlingoogle, line 1426: `sourceAddress` can be initialized if `Tlv::ReadUint16Tlv(aMessage, Tlv::kSourceAddress, sourceAddress)` failed according to @librasungirl 's analysis.

src/core/thread/mle.cpp
jwhui: Maybe go one step further - ignore MLE Data Response from different partition on FTD. Wait for MLE Advertisement as @librasungirl suggested.

src/core/thread/mle_router.cpp
jwhui: This check should not be necessary: a valid neighbor should not have RLOC16 = 0xffff.

src/core/thread/mle_router.cpp
jwhui: Do we call `RemoveNeighbor()` on neighbors with state invalid?

src/core/thread/mle_router.cpp
jwhui: Setting `aNeighbor = NULL` is probably sufficient here. Wherever we need a router entry below, we perform another lookup.

src/core/thread/mle_router.cpp
simonlingoogle: This check is to make sure that `sourceAddress` was properly initialized. Otherwise, `sourceAddress` can be of any value.

src/core/thread/mle_router.cpp
simonlingoogle: Yes, we do. Here is the place where we found that the `RemoveNeighbor` can be removing an invalid neighbor:
https://github.com/openthread/openthread/blob/e839100be10e7c3a4119936c2d0bdfaf11effa94/src/core/thread/mle_router.cpp#L2238-L2243

src/core/thread/mle_router.cpp
jwhui: If `sourceAddress` can be any value, then checking against `0xffff` is insufficient. I think a better solution is to set `aNeighbor = NULL` if reading the Source Address TLV fails.

simonlingoogle: I have initialized `sourceAddress` to `0xFFFF` in this PR. Now if reading `Source Address TLV` fails, `sourceAddress` is 0xFFFF;

jwhui: Then I return to my previous statement, if `sourceAddress` here is `0xffff`, then the check should not be necessary: a vaoid neighbor should not have RLOC16 = 0xffff.

jwhui: I think it would be safer if we simply set `aNeighbor = NULL` after failing to find source address. Note that the previous version of TLV read used to modify the buffer, even on failure. Some TLV reads continue to do that. We should not assume the buffer is unmodified on failure for now.

src/core/thread/mle_router.cpp
jwhui: Should we consider fully clearing the neighbor entry when setting to invalid so that we do not match on it?

src/core/thread/mle_router.cpp
jwhui: Also, not including a Source Address TLV in the message is an error by the neighbor. Removing the neighbor here might not be such a bad idea.

simonlingoogle: I will remove the condition `sourceAddress != 0xFFFF` here.

src/core/thread/mle.cpp
simonlingoogle: The solution sounds good to me. Will update according.

src/core/thread/mle_router.cpp
simonlingoogle: I was a little confused too. 
Per my understanding, an `Router` in `Invalid` state is not actually **INVALID**. An actual **INVALID** `Router` is the one with `RLOC16 == 0xFFFF`. The implementation of `RouterTable::GetNextEntry` has confirmed it too: 
```cpp
const Router *RouterTable::GetNextEntry(const Router *aRouter) const
{
    VerifyOrExit(aRouter != NULL, OT_NOOP);
    aRouter++;
    VerifyOrExit(aRouter < &mRouters[Mle::kMaxRouters], aRouter = NULL);
    VerifyOrExit(aRouter->GetRloc16() != 0xffff, aRouter = NULL);

exit:
    return aRouter;
}
```

So, a `Router` of `Invalid` state would just mean that the `Router` is totally unreachable, but all those link information is still sound. Maybe these information can be useful later?

abtink: That's all correct.

Here is how I usually treat the types (sort of my cheat sheet :))
- A `Neighbor` is a node we can directly talk to. `InvalidState` means it is not a neighbor (i.e. can not directly talk to it)
- `RouterTable` gives list of all routers within Thread network. Some `Router`s may be our neighbor and some may not be (e.g, two hops away).  `Router` inherits from `Neighbor` and uses `State` to indicate this. 
- A `Child` (`ChidTable`) is the list of devices own children and a `Child` is always a `Neighbor` (invalid state mean it is no longer a child).

`RouterTable` provides `GetNeighbor()` vs `GetRouter()` methods, first one returns a `Neighbor` (state is not invalid) vs second one returning a `Router` entry (may not be a neighbor).  

`ChildTable` provides `FindChild()` methods, returned entry is always a neighbor (state is not invalid). Let's caller to define a `StateFilter` to find `Child` in specific states.

`Mle` provides methods to `GetNeighbor()` (matching rloc, ext address, etc), returned value should be always a neighbor (i.e. not in invalid state). May be an entry from `ChildTable` or  `RouterTable` or one of special entries `Parent` or `ParentCandidate`.

jwhui: Yes, that is a good summary @abtink .

src/core/thread/mle.cpp
simonlingoogle: Updated accordingly.

src/core/thread/mle_router.cpp
```diff
@@ -1166,7 +1166,7 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
     const otThreadLinkInfo *linkInfo = static_cast<const otThreadLinkInfo *>(aMessageInfo.GetLinkInfo());
     uint8_t linkMargin = LinkQualityInfo::ConvertRssToLinkMargin(Get<Mac::Mac>().GetNoiseFloor(), linkInfo->mRss);
     Mac::ExtAddress macAddr;
-    uint16_t        sourceAddress;
+    uint16_t        sourceAddress = 0xFFFF;
```
jwhui, line 1169: How about using `Mac::kShortAddrInvalid`?

src/core/thread/mle_router.cpp
simonlingoogle: > Since aNeighbor is reset to NULL now, this scenario would not occur any more, right?

Yes, it should not happen any more. But I think we need to keep the `OT_ASSERT` because the very purpose of using `OT_ASSERT` is to catch something that **would not occur**.

simonlingoogle: > How could the Router receive an MLE advertisement from the REED who has already changed its RLOC in such short 1.25s considering the time consumption of the attach better process of REED itself and the at least 1s jitter

Could it be that the REED was a Router and had just downgraded from being a Router? 

However without comprehensive debugging, we can't not know for sure why it happened.

simonlingoogle:
This PR has been simulated for more than 200 days without any assertion failure. 
Any more comments ? @jwhui @abtink @librasungirl

simonlingoogle:
It's strange that simulation tests failed a lot for this PR. I am marking this PR as `WIP` for further investigation. @abtink @jwhui @librasungirl

src/core/thread/mle_router.cpp
simonlingoogle, line 1287: Turnout this is problematic, causing  `Cert_6_1_04_REEDAttachConnectivity.py` to fail

simonlingoogle: Fixed: set `aNeighbor` to `NULL` only when it's in RouterTable.

src/core/thread/mle_router.cpp
```diff
@@ -1284,6 +1284,10 @@ otError MleRouter::HandleAdvertisement(const Message &         aMessage,
         if (processRouteTlv)
         {
             SuccessOrExit(error = ProcessRouteTlv(route));
+            if (Get<RouterTable>().GetRouterIndex(static_cast<Router &>(*aNeighbor)) < kMaxRouters)
+            {
+                aNeighbor = NULL; // aNeighbor is no longer valid after `ProcessRouteTlv`
```
mwilniewiec, line 1289: Could you please give me short explanation why aNeighbor is invalid after ProcessRouteTlv?
Im trying to apply your patch to older OT code that we use currently, I don't have aNeighbor parameter in HandleAdvertisement (added by Jonathan) and Im trying to understand the fix. Thanks

simonlingoogle: The condition `Get<RouterTable>().GetRouterIndex(static_cast<Router &>(*aNeighbor)) < kMaxRouters` is satisfied when `aNeighbor` references to a `Router` in the router table `mRouterTable`. 

It is all about how router table organize `Routers` in `mRouters`. **Note that the `Routers` in `mRouters` are ordered by `Router ID`.**
Let's assume that the current valid Router ID set is {1, 3, 5} and `aNeighbor` points to `Router 5`, i.e. `aNeighbor == &mRouters[2]`. If the new Router ID set is {1, 5} in the RouterTLV, after `ProcessRouteTlv`, `Router 5` is moved from `mRouters[2]` to `mRouters[1]`, however the value of `aNeighbor` is not updated accordingly. So, after `ProcessRouterTLV`, `aNeighbor` is still pointing to `&mRouters[2]`, which is actually an **INVALID** `Router`. 

This fix sets `aNeighbor = NULL` in such case so that the `aNeighbor` handling code after the exit label won't be executed.

mwilniewiec: Thanks, this wasnt obvious :)
Adding routerTableChanged result to a ProcessRouteTlv function would improve readability of this code.

Don't we have a problem here that the whole block of code for a kRoleChild:
if (aNeighbor == &mParent) { ... } (line 1301) will never execute?
From what I see processRouteTlv is true when a child receives an advertisement from its parent.

simonlingoogle: @mwilniewiec
Why do you think `if (aNeighbor == &mParent) { ... } (line 1301)` will never execute? 
Note that when `aNeighbor == &mParent`, `aNeighbor` is not pointing to a `Router`, so it won't be set to `NULL` after `ProcessRouteTLV`.

mwilniewiec: Ok, after looking deeper now I see.
GetNeighbor will return mParent for a child node (even though it is an FTD and has the same entry in its router table) but for a router it returns a pointer to router table entry.
In that case this should work as expected.

simonlingoogle: Thanks for your investigation.

src/core/thread/router_table.hpp
```diff
@@ -265,6 +265,16 @@ class RouterTable : public InstanceLocator
...
 
+    /**
+     * This method returns the router table index for a given `Router` instance.
+     *
+     * @param[in]  aRouter  A reference to a `Router`
+     *
+     * @returns The index corresponding to @p aRouter.
+     *
+     */
+    size_t GetRouterIndex(const Router &aRouter) const { return static_cast<uint16_t>(&aRouter - mRouters); }
```
abtink, line 276: I have a suggestion here:
How about instead of this, we add a method like 
```
    /**
     * This method indicates whether a given neighbor is in the router table.
     *
     * @note This method checks whether the `aNeighbor` instance itself is an entry from the router table.
     *
     * @param[in] aNeighbor   A reference to a neighbor.
     *
     * @returns TRUE if the neighbor is present in the router table, FALSE otherwise.
     *
     */
    bool ContainsNeighbor(const Neighbor &aNeighbor) const;
```

Two reason for this:
- I think it will help simplify the code where we use this to check.
- The fact that `RouterTable` uses an array to store router entries is a private info in the `RouterTable`. Publicly, the `RouterTable` provides methods to search/find a entry and an `Iterator` to go through the entries (which does now show what the internal data structure is). The whole notion of router table "index" (which is only applicable when we know we have an array) is not part of public APIs of `RouterTable`.  Using `ContainsNeighbor` would be more generic (it allows us in future to consider changing internal data structure used to store routers from array).

simonlingoogle: Agree. Will revise. I have actually done it but reverted back to have less code changes. 

Do you think the method name should be `Contains` rather than `ContainsNeighbor` because the argument type already says it's a Neighbor.

abtink: Thanks. 
Yes `Contains` sounds good to me.

simonlingoogle: revised accordingly

simonlingoogle:
any further comments? @mwilniewiec

src/core/thread/router_table.hpp
```diff
@@ -265,6 +265,20 @@ class RouterTable : public InstanceLocator
...
+     *
+     * @param[in]  aNeighbor  A reference to a `Neighbor`.
+     *
+     * @returns Whether the router table contains @p aNeighbor.
+     *
+     */
+    bool Contains(const Neighbor &aNeighbor) const
+    {
+        return mRouters <= &static_cast<const Router &>(aNeighbor) &&
+               &static_cast<const Router &>(aNeighbor) < mRouters + Mle::kMaxRouters;
```
librasungirl, line 279: It seems to me always return True. When would it be false? @simonlingoogle 

```
return mRouters <= &static_cast<const Router &>(aNeighbor) &&
               &static_cast<const Router &>(aNeighbor) < mRouters + Mle::kMaxRouters;
```

simonlingoogle: when aNeighbor is not a Router in mRouters, e.x. mParent, mParentCandidate

librasungirl: how about adding note to illustrate False scenarios for easier readability?

simonlingoogle: comment added

src/core/thread/router_table.hpp
```diff
@@ -265,6 +265,22 @@ class RouterTable : public InstanceLocator
...
     Router *GetRouter(const Mac::ExtAddress &aExtAddress);
 
+    /**
+     * This method returns if the router table contains a given `Neighbor` instance.
+     *
+     * @param[in]  aNeighbor  A reference to a `Neighbor`.
+     *
+     * @retval TRUE  if @p aNeighbor is a `Router` in the router table.
+     * @retval FALSE if @p aNeighbor is not a `Router` in the router table
+     *               (e.x. mParent, mParentCandidate, a `Child` of the child table).
```
librasungirl, line 275: nit: `e.x.` -> `e.g.` or `i.e.`?
it seems `e.x.` only appear here or in `include/openthread/platform/otns.h`.

simonlingoogle: fixed

PR #4862, opened 2020-04-20
[dua] MTD prioritizes DUA in AR Tlv

librasungirl:
Thread 1.2 spec requires prioritizing DUA in AR TLV to ensure DUA could be stored in the limited addresses space for Child on the parent.

This PR depends on #4854 , please only review the last commit which is the purpose of this PR.

### DIFF
--- a/src/core/backbone_router/leader.hpp
+++ b/src/core/backbone_router/leader.hpp
@@ -166,7 +166,7 @@ private:
 #endif
 
     BackboneRouterConfig mConfig;       ///< Primary Backbone Router information.
-    otIp6Prefix          mDomainPrefix; ///< Domain Prefix on the Thread Network.
+    otIp6Prefix          mDomainPrefix; ///< Domain Prefix in the Thread network.
 };
 
 } // namespace BackboneRouter
--- a/src/core/thread/dua_manager.hpp
+++ b/src/core/thread/dua_manager.hpp
@@ -82,15 +82,12 @@ public:
     void UpdateDomainUnicastAddress(BackboneRouter::Leader::DomainPrefixState aState);
 
     /**
-     * This method returns a pointer to the Domain Unicast Address.
+     * This method returns a reference to the Domain Unicast Address.
      *
-     * @returns A pointer to the Domain Unicast Address or NULL if no valid Domain Unicast Address.
+     * @returns A reference to the Domain Unicast Address.
      *
      */
-    const Ip6::Address *GetDomainUnicastAddress(void) const
-    {
-        return mDomainUnicastAddress.mValid ? &mDomainUnicastAddress.GetAddress() : NULL;
-    }
+    const Ip6::Address &GetDomainUnicastAddress(void) const { return mDomainUnicastAddress.GetAddress(); }
 
 private:
     Ip6::NetifUnicastAddress mDomainUnicastAddress;
--- a/src/core/thread/dua_manager.hpp
+++ b/src/core/thread/dua_manager.hpp
@@ -81,6 +81,17 @@ public:
      */
     void UpdateDomainUnicastAddress(BackboneRouter::Leader::DomainPrefixState aState);
 
+    /**
+     * This method returns a pointer to the Domain Unicast Address.
+     *
+     * @returns A pointer to the Domain Unicast Address or NULL if no valid Domain Unicast Address.
+     *
+     */
+    const Ip6::Address *GetDomainUnicastAddress(void) const
+    {
+        return mDomainUnicastAddress.mValid ? &mDomainUnicastAddress.GetAddress() : NULL;
+    }
+
 private:
     Ip6::NetifUnicastAddress mDomainUnicastAddress;
     uint8_t                  mDadCounter;
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -1362,17 +1362,18 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
 
 #if OPENTHREAD_CONFIG_DUA_ENABLE
     {
-        const Ip6::Address *dua = Get<DuaManager>().GetDomainUnicastAddress();
+        // Cache Domain Unicast Address.
+        address = Get<DuaManager>().GetDomainUnicastAddress();
 
-        if (dua && Get<ThreadNetif>().IsUnicastAddress(*dua))
+        if (Get<ThreadNetif>().IsUnicastAddress(address))
         {
-            error = Get<NetworkData::Leader>().GetContext(*dua, context);
+            error = Get<NetworkData::Leader>().GetContext(address, context);
 
             OT_ASSERT(error == OT_ERROR_NONE);
 
             // Prioritize DUA, compressed entry
             entry.SetContextId(context.mContextId);
-            entry.SetIid(dua->GetIid());
+            entry.SetIid(address.GetIid());
             SuccessOrExit(error = aMessage.Append(&entry, entry.GetLength()));
             length += entry.GetLength();
             counter++;
@@ -1390,7 +1391,7 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
 
 #if OPENTHREAD_CONFIG_DUA_ENABLE
         // Here skips DUA which was appended already.
-        if (&addr->GetAddress() == Get<DuaManager>().GetDomainUnicastAddress())
+        if (addr->GetAddress() == address)
         {
             continue;
         }
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -1360,6 +1360,26 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
     VerifyOrExit(aMode != kAppendMeshLocalOnly, OT_NOOP);
     counter++;
 
+#if OPENTHREAD_CONFIG_DUA_ENABLE
+    {
+        const Ip6::Address *dua = Get<DuaManager>().GetDomainUnicastAddress();
+
+        if (dua && Get<ThreadNetif>().IsUnicastAddress(*dua))
+        {
+            error = Get<NetworkData::Leader>().GetContext(*dua, context);
+
+            OT_ASSERT(error == OT_ERROR_NONE);
+
+            // Prioritize DUA, compressed entry
+            entry.SetContextId(context.mContextId);
+            entry.SetIid(dua->GetIid());
+            SuccessOrExit(error = aMessage.Append(&entry, entry.GetLength()));
+            length += entry.GetLength();
+            counter++;
+        }
+    }
+#endif // OPENTHREAD_CONFIG_DUA_ENABLE
+
     for (const Ip6::NetifUnicastAddress *addr = Get<ThreadNetif>().GetUnicastAddresses(); addr; addr = addr->GetNext())
     {
         if (addr->GetAddress().IsLinkLocal() || IsRoutingLocator(addr->GetAddress()) ||
@@ -1367,6 +1387,14 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
         {
             continue;
         }
+
+#if OPENTHREAD_CONFIG_DUA_ENABLE
+        // Here skips DUA which was appended already.
+        if (&addr->GetAddress() == Get<DuaManager>().GetDomainUnicastAddress())
+        {
+            continue;
+        }
+#endif
 
         if (Get<NetworkData::Leader>().GetContext(addr->GetAddress(), context) == OT_ERROR_NONE)
         {
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -1361,23 +1361,21 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
     counter++;
 
 #if OPENTHREAD_CONFIG_DUA_ENABLE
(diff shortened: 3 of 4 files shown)


src/core/thread/mle.cpp
```diff
@@ -1424,6 +1443,14 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
             continue;
         }
 
+#if OPENTHREAD_CONFIG_TMF_DUA_REGISTRATION_ENABLE
+        // Here skips DUA which was appended already.
+        if (addr->GetAddress() == GetDomainUnicastAddress())
```
simonlingoogle, line 1448: Better just compare by pointer address?
```suggestion
        if (&(addr->GetAddress()) == &GetDomainUnicastAddress())
```

src/core/thread/mle.cpp
```diff
@@ -1416,6 +1416,25 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
...
+        if (Get<ThreadNetif>().IsUnicastAddress(address))
+        {
+            OT_ASSERT(Get<NetworkData::Leader>().GetContext(address, context) == OT_ERROR_NONE);
+
+            // Prioritize DUA, compressed entry
+            entry.SetContextId(context.mContextId);
+            entry.SetIid(address.GetIid());
+            SuccessOrExit(error = aMessage.Append(&entry, entry.GetLength()));
+            length += entry.GetLength();
+            counter++;
```
simonlingoogle, line 1433:
```suggestion
            counter++;
            VerifyOrExit(counter < OPENTHREAD_CONFIG_MLE_IP_ADDRS_TO_REGISTER, OT_NOOP);
```
is it necessary ?

src/core/thread/mle.cpp
```diff
@@ -1416,6 +1416,25 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
...
     counter++;
 
+#if OPENTHREAD_CONFIG_TMF_DUA_REGISTRATION_ENABLE
+    if (Get<BackboneRouter::Leader>().HasDomainPrefix())
+    {
+        const Ip6::Address &address = GetDomainUnicastAddress();
+
+        if (Get<ThreadNetif>().IsUnicastAddress(address))
+        {
+            OT_ASSERT(Get<NetworkData::Leader>().GetContext(address, context) == OT_ERROR_NONE);
```
simonlingoogle, line 1426: Should not put real logic code in `OT_ASSERT` because `OT_ASSERT` can be turned off.

src/core/thread/mle.cpp
librasungirl: I don't think it is necessary. 
`OPENTHREAD_CONFIG_MLE_IP_ADDRS_TO_REGISTER` equals to `OPENTHREAD_CONFIG_MLE_IP_ADDRS_PER_CHILD`, which is at least `4` according to Thread Conformance spec.
https://github.com/openthread/openthread/blob/779da7c858445d11df571dabfe344f11670c4109/src/core/config/mle.h#L87-L89

IIRC, there should be no test case requiring MTD to register less than maximal 4 entries in AR TLV, is there?

src/core/thread/mle.cpp
librasungirl: separated `OT_ASSERT` out.

src/core/backbone_router/leader.hpp
```diff
@@ -166,7 +166,7 @@ class Leader : public InstanceLocator
 #endif
 
     BackboneRouterConfig mConfig;       ///< Primary Backbone Router information.
-    otIp6Prefix          mDomainPrefix; ///< Domain Prefix on the Thread Network.
+    otIp6Prefix          mDomainPrefix; ///< Domain Prefix in the Thread Network.
```
simonlingoogle, line 169:
```suggestion
    otIp6Prefix          mDomainPrefix; ///< Domain Prefix in the Thread network.
```

src/core/thread/dua_manager.hpp
```diff
@@ -81,6 +81,17 @@ class DuaManager : public InstanceLocator
      */
     void UpdateDomainUnicastAddress(BackboneRouter::Leader::DomainPrefixState aState);
 
+    /**
+     * This method returns a pointer to the Domain Unicast Address.
+     *
+     * @returns A pointer to the Domain Unicast Address or NULL if no valid Domain Unicast address.
```
simonlingoogle, line 87:
```suggestion
     * @returns A pointer to the Domain Unicast Address or NULL if no valid Domain Unicast Address.
```

librasungirl:
Updated along with the dependent #4854 getting merged. Please take a review. @abtink @jwhui

src/core/thread/mle.cpp
```diff
@@ -1368,6 +1388,14 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
             continue;
         }
 
+#if OPENTHREAD_CONFIG_DUA_ENABLE
+        // Here skips DUA which was appended already.
+        if (&addr->GetAddress() == Get<DuaManager>().GetDomainUnicastAddress())
```
abtink, line 1393: We are comparing pointers here. While this works ok (since from `GetDomainUnicastAddress()` we are returning a pointer to `mAddress` field in the same `Ip6::NetifUnicastAddress` instance as the one we would be adding to `Netif` and later get in `addr` as we iterate through the `Netif` address entries) it relies on this knowledge of internal/private  implementation/behavior of `DuaManager::GetDomainUnicastAddress()` and `Netif`. If any of these gets changed in future this code would become incorrect (and such issue would not be easily noticed).

To be safe and correct, I would suggest we do check the IP address value directly instead of checking the pointers. 

Thoughts?

simonlingoogle: Our existing usage of `RemoveUnicastAddress` and `UnsubscribeMulticast` rely on the fact that the `Ip6::NetifUnicastAddress` instances in `NetIf` are the same members. 

For example, this code rely on the fact that only `&mAllDomainBackboneRouters` can be unsubscribed because the actual content of `mAllDomainBackboneRouters` is not correctly set in some cases.  
https://github.com/openthread/openthread/blob/7b0cb984455731bbaf2eb51188baf5d225a2ee41/src/core/backbone_router/local.cpp#L369-L375

It would be worthless if we only check IP address here but other code are not properly revised.

That being said, I would prefer pointer comparing for its efficiency. Thoughts? @librasungirl

abtink: @simonlingoogle you raise a good question but my initial point is different (not directly related to `NetifUnicastAddress` entries being same as the one added earlier (which in my opinion is fine)).

Note that `Ip6::Address` and `Ip6::NetifUnicastAddress` are two different types:
- The `DuaManager::GetDomainUnicastAddress()` is returning an `Ip6::Address *` 
- The fact that this pointer is derived from a `Ip6::NetifUnicastAddress` instance is an internal/private implementation of `DuaManager` class.
  - Note that `mDomainUnicastAddress` is a private member of `DuaManager`.
- So using this fact from another class (`Mle` here) is not good/safe.
- If in future for any reason the `DuaManager` implementation gets changed (e.g., we save the IP address differently), it would not be easy for developer making the change to realize that that `Ip6::Address *` returned from `GetDomainUnicastAddress()` may be used  like this in other modules.

----------

Regarding `Ip6::NetifUnicastAddress` and `Add/RemoveUnicastAddress()` (and multicast version), the fact that they add/remove the `NetifUnicastAddress` instance itself (and not copy it) is intentional and part of the design and public behavior of `Netif` class methods:
- We have two types of methods for adding/removing unicast addresses (similarly for multicast):
   -  `AddUnicastAddress()`/`RemoveUnicastAddress()`  (which are intended for core/internal use) and will directly use the passed-in `NetifUnicastAddress` instance itself.
   - `AddExternalUnicastAddress()`/`RemoveExternalUnicastAddress()` which will not use the passed-in `NetifUnicastAddress` instance and will create a copy of it.
- So if in other modules/classes, we rely on the fact that `Add/RemoveUnciastAddress()` is using the entry, it is consistent and part of the intended known public behavior of API/methods of `Netif` (though I would say, we have room for improvement in documenting this more clearly in the `Netif` module).

simonlingoogle: Thanks for your clarification. What you said makes much sense.

If we are going to use address comparing rather than pointer comparing, we need to make sure that `DuaManager::GetDomainUnicastAddress()` always return `NULL` when DUA is not added to `Netif`, meaning we need to correctly set `mDomainUnicastAddress.mValid` at all cases, which I think is not true for now:

- When `DUA` is removed from `Netif` due to domain prefix change,  `mDomainUnicastAddress.mValid` is not set to `false`, thus  `DuaManager::GetDomainUnicastAddress()` can still return the `DUA` address which is not correct. 
https://github.com/openthread/openthread/blob/7b0cb984455731bbaf2eb51188baf5d225a2ee41/src/core/thread/dua_manager.cpp#L62-L66

- When `SLAAC` failed to generate a `DUA`, `mDomainUnicastAddress.mValid` is not set to `false` either. 
https://github.com/openthread/openthread/blob/7b0cb984455731bbaf2eb51188baf5d225a2ee41/src/core/thread/dua_manager.cpp#L80-L88

simonlingoogle: Can we just compare `addr` with `&Get<DuaManager>().mDomainUnicastAddress` ?

librasungirl: @abtink  @simonlingoogle Thanks for the discussion here! 
Agree that it would be better to check the address value directly for safe. 
To keep the consistence of core/internal unicast address usage, I pushed the second commit to resolve your concerns:
1) Simply return the reference of the address in `DuaManager().GetDomainUnicastAddress()`;
2) Use `Get<ThreadNetif>().IsUnicastAddress()` to determine whether it is a valid address successfully added to the interface.

Please take a look.

src/core/thread/mle.cpp
```diff
@@ -1360,6 +1360,27 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
     VerifyOrExit(aMode != kAppendMeshLocalOnly, OT_NOOP);
     counter++;
 
+#if OPENTHREAD_CONFIG_DUA_ENABLE
+    {
+        // Cache Domain Unicast Address.
+        address = Get<DuaManager>().GetDomainUnicastAddress();
```
librasungirl, line 1366: One note is that here does not define a new local reference variable but reuse the `IP6::Address address` already defined in the beginning of the body, so it could be used directly in the below for loop.

src/core/thread/mle.cpp
```diff
@@ -1362,17 +1362,18 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
 
 #if OPENTHREAD_CONFIG_DUA_ENABLE
     {
-        const Ip6::Address *dua = Get<DuaManager>().GetDomainUnicastAddress();
+        // Cache Domain Unicast Address.
+        address = Get<DuaManager>().GetDomainUnicastAddress();
 
-        if (dua && Get<ThreadNetif>().IsUnicastAddress(*dua))
+        if (Get<ThreadNetif>().IsUnicastAddress(address))
```
simonlingoogle, line 1368: One concern is that when `DUA` is not added to `Netif`, this address can be any value, i.e. can be equal a non-DUA address. Might incorrectly prioritize a non-DUA address despite a very low chance?

librasungirl: For what I can see, there are 3 possible situations for the address
1) valid DUA
2) valid domain prefix, but reserved iid(when generate iid fails)
3) All 0,  unspecified address (when there is no available domain prefix)

For 1), it should be added to Thread interface already;
For 2), first little chance for GenerateIid fails; and even fail, I don't think it will be coincident with other addresses added to the Thread interface - note there should be only one DUA either manually configured or generated according to the spec.
For 3), I don't think there would be such address added to Thread interface.

Any other possible scenarios I missed? @simonlingoogle

librasungirl: Submitted a new patch to clear the address when possible so that generally only above 1) and 3) may happen.

Any more concerns? @simonlingoogle

simonlingoogle: My concerns have been resolved with this patch. 

My major concern was actually about 1). Since we are going to have the fixed `DUA` feature, a valid DUA, which is not added to `Netif`, can be set to almost any address. 

We should have no such issue  since now we guarantee that when `DUA` is not in `Netif`, it is always cleared to `::`.

src/core/thread/mle.cpp
abtink: Sounds great and simple. Thanks

src/core/thread/mle.cpp
```diff
@@ -1360,6 +1360,27 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
     VerifyOrExit(aMode != kAppendMeshLocalOnly, OT_NOOP);
     counter++;
 
+#if OPENTHREAD_CONFIG_DUA_ENABLE
+    {
```
jwhui, line 1364: nit: is this paren necessary?

src/core/thread/mle.cpp
```diff
@@ -1368,6 +1389,14 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
             continue;
         }
 
+#if OPENTHREAD_CONFIG_DUA_ENABLE
+        // Here skips DUA which was appended already.
```
jwhui, line 1393:
```suggestion
        // Skip DUA that was already appended above.
```

PR #4866, opened 2020-04-21
[size-report] support local check

bukepo:


jwhui:
> How would it work on pushed branch? Do we go to GH action "size" and see the repot in the build logs? Thanks.

I think you have to rely on build logs in this case. The size-report service has no where to post a comment when it's just a branch.

bukepo:
@abtink Yes, the size report will be printed in the log of GitHub Actions.

Issue #4869, opened 2020-04-21
OT_CONFIG and OT_PRIVATE_DEFINES specified too late to impact platform


rob-the-dude:
**Describe the bug**
In the top level CMakeLists.txt, if OT_CONFIG is defined, then it is added to OT_PRIVATE_DEFINES only AFTER the OT_PLATFORM directory has been added.  Similarly, any externally provided options (OT_PLATFORM_DEFINES) are also added to OT_PRIVATE_DEFINES only AFTER the OT_PLATFORM directory has been added.  While the platform CMakeLists.txt files generally only add values to the OT_PLATFORM_DEFINES, they could also look at switches already included to modify behavior.  Likewise, at least one CMakeLists.txt file (simulation) checks to see if OT_CONFIG has been specified -- and in this case, it hasn't, but it WILL be, in the top-level CMakeLists.txt file, about three lines later.

Is this intentional?  If so, why?

If not, then the following change seems to "fix" it for me (it's basically moving the "if (OT_CONFIG)" check and the addition of OT_PLATFORM_DEFINES up about 8 lines) -- I'm happy to open a PR and provide the change if others think it's correct.

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 9605fbf7..1c6e7c6f 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,13 @@ list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_BINARY_DIR}/etc/cmake)
 list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_SOURCE_DIR}/etc/cmake)
 list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_SOURCE_DIR}/include)
 
+if(OT_CONFIG)
+    list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_PROJECT_CORE_CONFIG_FILE=\"${OT_CONFIG}\"")
+    message(STATUS "Project core config: \"${OT_CONFIG}\"")
+endif()
+
+list(APPEND OT_PRIVATE_DEFINES ${OT_PLATFORM_DEFINES})
+
 if(OT_PLATFORM STREQUAL "posix")
     list(APPEND OT_PRIVATE_INCLUDES ${PROJECT_SOURCE_DIR}/src/posix/platform)
     add_subdirectory("${PROJECT_SOURCE_DIR}/src/posix/platform")
[...]

PR #4870, opened 2020-04-21
move OT_CONFIG and OT_PRIVATE_DEFINES check

rob-the-dude:
Move OT_CONFIG and OT_PRIVATE_DEFINES check prior to platform subdirectory addition.  Allows subdirectory CMakeLists.txt files to pick up on their contents.  Addresses #4869.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,13 @@ list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_BINARY_DIR}/etc/cmake)
 list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_SOURCE_DIR}/etc/cmake)
 list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_SOURCE_DIR}/include)
 
+if(OT_CONFIG)
+    list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_PROJECT_CORE_CONFIG_FILE=\"${OT_CONFIG}\"")
+    message(STATUS "Project core config: \"${OT_CONFIG}\"")
+endif()
+
+list(APPEND OT_PRIVATE_DEFINES ${OT_PLATFORM_DEFINES})
+
 if(OT_PLATFORM STREQUAL "posix")
     list(APPEND OT_PRIVATE_INCLUDES ${PROJECT_SOURCE_DIR}/src/posix/platform)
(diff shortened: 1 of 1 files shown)


rob-the-dude:
There's clearly something unhappy about it -- it looks like the simulation projects break because, by moving the config earlier, they become dependent on "otPlatAlarmMicroFired".  I haven't looked at why yet, but will.  Feedback welcome.

CMakeLists.txt
```diff
@@ -99,6 +99,13 @@ list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_BINARY_DIR}/etc/cmake)
 list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_SOURCE_DIR}/etc/cmake)
 list(APPEND OT_PUBLIC_INCLUDES ${PROJECT_SOURCE_DIR}/include)
 
+if(OT_CONFIG)
+    list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_PROJECT_CORE_CONFIG_FILE=\"${OT_CONFIG}\"")
+    message(STATUS "Project core config: \"${OT_CONFIG}\"")
+endif()
+
+list(APPEND OT_PRIVATE_DEFINES ${OT_PLATFORM_DEFINES})
```
jwhui, line 107: It appears that the core openthread lib is not pulling in the project config header.

I think wee need to:

1. Leave this line as it was before.

1. Swap the order of these two lines 

https://github.com/openthread/openthread/blob/c034fd0c0cceeaa9a1d74e384ef0d472666bfd98/examples/platforms/simulation/CMakeLists.txt#L53-L55

rob-the-dude: I do now understand what's going on with OT_CONFIG.  And I see that the platform files are supposed to be setting OT_PLATFORM_DEFINES so it can be used later.

However, I think the thing that's bugging me is that I can't pass something "down" from the top level into the platform layer to use as it sets things up for the remaining CMake list files.

jwhui: Moving `OPENTHREAD_PROJECT_CORE_CONFIG_FILE` up as you have proposed is all you need, right? We just need to move the `list()` command back to where it was before and swap a couple lines in `examples/platforms/simulation/CMakeLists.txt`. Would that work for you?

rob-the-dude:
Many changes have been made that may make this request obsolete.  I'm going to live with them for a bit and see if there's anything additional I need.

PR #4871, opened 2020-04-22
[posix] use enumerations for RCP bus

bukepo:
This PR changes the way to define RCP bus type. It now defines two
enumerations for UART and SPI. This ensures only a single bus type is
enabled, and helps keep compatible with projects using UART bus.


src/posix/platform/openthread-posix-config.h
```diff
@@ -58,33 +58,37 @@
...
+/**
+ * @def OPENTHREAD_POSIX_RCP_BUS
+ *
+ * This setting configures what type of RCP bus to use.
  *
  */
-#ifndef OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE
-#define OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE 0
+#ifndef OPENTHREAD_POSIX_RCP_BUS
+#define OPENTHREAD_POSIX_RCP_BUS OT_POSIX_RCP_BUS_UART
```
jwhui, line 91: nit: any reason why `OPENTHREAD_POSIX_RCP_BUS` rather than `OPENTHREAD_POSIX_CONFIG_RCP_BUS`?

Do we want a config check to make sure this is set to either SPI or UART?

bukepo: Currently this is checked in radio.cpp, does it working for you?

jwhui: > Currently this is checked in radio.cpp, does it work for you?

Yes, I think that is fine. Thanks.

PR #4872, opened 2020-04-22
[script] remove 'tmp' directory for each simulation

simonlingoogle:
Currently we remove `tmp` directory in `script/cert` or `test-driver` before running each test case. However, there is nothing prevent us to have multiple tests in one unit tests, e.x. [test_reed_address_solicit_rejected.py](https://github.com/openthread/openthread/blob/master/tests/scripts/thread-cert/test_reed_address_solicit_rejected.py).

This PR removes `tmp` directory to give each simulation a clean start.

**[test_reed_address_solicit_rejected.py](https://github.com/openthread/openthread/blob/master/tests/scripts/thread-cert/test_reed_address_solicit_rejected.py) has a small chance to fail without this fix.**

### DIFF
--- a/tests/scripts/thread-cert/config.py
+++ b/tests/scripts/thread-cert/config.py
@@ -577,4 +577,6 @@ def create_default_simulator():
 
 
 def _clear_tmp_files():
-    os.system(f"rm -f tmp/{PORT_OFFSET}_*.flash tmp/{PORT_OFFSET}_*.data tmp/{PORT_OFFSET}_*.swap")
+    os.system(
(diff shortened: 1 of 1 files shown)


tests/scripts/thread-cert/config.py
```diff
@@ -106,6 +106,8 @@
 
 PROTOCOL_VERSION = 2
 
+PORT_OFFSET = int(os.getenv('PORT_OFFSET', '0'))
```
bukepo, line 109: as discussed offline, prefer solving this issue in the particular test case only. And use a new solution for migrating to cmake.

PR #4875, opened 2020-04-23
[cmake] move cc1352 & cc2652 libs to third-party cmake file

FiveDimensions:
move thirdparty related in cc1352&cc2652 cmake files into their respective cmake files

### DIFF
--- a/third_party/ti/CMakeLists.txt
+++ b/third_party/ti/CMakeLists.txt
@@ -38,7 +38,7 @@ if(OT_PLATFORM STREQUAL "cc2650")
         INTERFACE
             -T${CMAKE_CURRENT_SOURCE_DIR}/devices/cc26x0/linker_files/cc26x0f128.lds
     )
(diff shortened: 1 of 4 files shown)


third_party/ti/CMakeLists.txt
```diff
@@ -38,4 +38,16 @@ if(OT_PLATFORM STREQUAL "cc2650")
         INTERFACE
             -T${CMAKE_CURRENT_SOURCE_DIR}/devices/cc26x0/linker_files/cc26x0f128.lds
     )
+elseif(OT_PLATFORM MATCHES "cc(13|26)[0-9]2")
```
librasungirl, line 41: It may be fine for now, but we are not sure if there would be other platforms matching the regex but requiring other drivers.
How about explicitly matching the exact platforms and combining with `OR`?

FiveDimensions: Ok, will compare the specific platforms name

PR #4879, opened 2020-04-23
[mesh-forwarder] ensure mesh dst checks for dst unreach

jwhui:
This PR contains the following commits:

- [mesh-forwarder] send all message fragments to the same mesh dest
    
    All fragments for a given IPv6 datagram must be sent to the same
    destination. Otherwise, there is not much point in transmitting the
    messages.

- [mesh-forwarder] ensure mesh dst checks for dst unreach
    
    An out-dated EID-to-RLOC map cache may lead a device to send an IPv6
    datagram to a device that does not have the IPv6 Destination Address
    assigned to its interface. However, the existing implementation would
    not send back an ICMPv6 Destination Unreachable message necessary to
    invalidate the originator's EID-to-RLOC cache entry.
    
    This implementation adds reachability checks on the mesh destination
    receive path to ensure that reachability checks are performed in the
    above situation.

- [toranj] update address cache table test
    
    Updated test-021-address-cache-table.py to reflect cache entry
    invalidation when sending to a neighbor that does not have the
    destination.

### DIFF
--- a/src/core/thread/mesh_forwarder.cpp
+++ b/src/core/thread/mesh_forwarder.cpp
@@ -1102,8 +1102,11 @@ void MeshForwarder::HandleFragment(const uint8_t *         aFrame,
         message->SetNetworkTimeOffset(aLinkInfo.mNetworkTimeOffset);
 #endif
 
-        // Security Check
         VerifyOrExit(Get<Ip6::Filter>().Accept(*message), error = OT_ERROR_DROP);
+
+#if OPENTHREAD_FTD
+        SendIcmpErrorIfDstUnreach(*message, aMacSource, aMacDest);
+#endif
 
         // Allow re-assembly of only one message at a time on a SED by clearing
         // any remaining fragments in reassembly list upon receiving of a new
@@ -1301,8 +1304,11 @@ void MeshForwarder::HandleLowpanHC(const uint8_t *         aFrame,
     message->SetNetworkTimeOffset(aLinkInfo.mNetworkTimeOffset);
 #endif
 
-    // Security Check
     VerifyOrExit(Get<Ip6::Filter>().Accept(*message), error = OT_ERROR_DROP);
+
+#if OPENTHREAD_FTD
+    SendIcmpErrorIfDstUnreach(*message, aMacSource, aMacDest);
+#endif
 
 exit:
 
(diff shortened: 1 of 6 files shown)


src/core/thread/mesh_forwarder.cpp
```diff
@@ -1102,9 +1102,12 @@ void MeshForwarder::HandleFragment(const uint8_t *         aFrame,
         message->SetNetworkTimeOffset(aLinkInfo.mNetworkTimeOffset);
 #endif
 
-        // Security Check
         VerifyOrExit(Get<Ip6::Filter>().Accept(*message), error = OT_ERROR_DROP);
 
+#if OPENTHREAD_FTD
+        CheckReachability(*message, aMacSource, aMacDest);
```
abtink, line 1108: Do I get it correctly? We don't care about returned `error` here (we want to send the unreachable error only). 
(can we add `IgnoreReturnValue` here).

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -404,13 +404,17 @@ otError MeshForwarder::UpdateMeshRoute(Message &aMessage)
...
+otError MeshForwarder::UpdateIp6RouteFtd(Ip6::Header &ip6Header, Message &aMessage)
 {
     Mle::MleRouter &mle   = Get<Mle::MleRouter>();
     otError         error = OT_ERROR_NONE;
     Neighbor *      neighbor;
 
-    if (mle.IsRoutingLocator(ip6Header.GetDestination()))
+    if (aMessage.GetOffset() > 0)
+    {
+        mMeshDest = aMessage.GetMeshDest();
```
abtink, line 415: This is used for direct msg tx only, correct?

tests/toranj/test-021-address-cache-table.py
```diff
@@ -226,22 +226,148 @@ def check_c2_is_removed_from_r2_child_table():
 wpan.Node.perform_async_tx_rx()
 verify(sender.was_successful and recver.was_successful)
 
-# The address cache table on r1 should still be the same as before.
+# The address cache table on r1 should have c2's address removed.
```
abtink, line 229: 👍

src/core/thread/mesh_forwarder.cpp
jwhui: Changed to return void. Also renamed to `SendIcmpErrorIfDstUnreach()`.

src/core/thread/mesh_forwarder_ftd.cpp
jwhui: Yes, correct.

PR #4885, opened 2020-04-24
[build] fix warnings for ti-cgt

srickardti:
The ti-cgt warns for unreachable code after asserts or infinite loops. This PR adds unreachable code macros to the sections of code after the asserts.


include/openthread/link_raw.h
```diff
@@ -267,7 +267,7 @@ otError otLinkRawSrcMatchEnable(otInstance *aInstance, bool aEnable);
  * @retval OT_ERROR_INVALID_STATE    If the raw link-layer isn't enabled.
  *
  */
-otError otLinkRawSrcMatchAddShortEntry(otInstance *aInstance, uint16_t aShortAddress);
+otError otLinkRawSrcMatchAddShortEntry(otInstance *aInstance, const uint16_t aShortAddress);
```
jwhui, line 270: Did you mean to add `const` here? Other similar instances below.

src/core/coap/coap.cpp
```diff
@@ -309,7 +309,7 @@ otError CoapBase::SendHeaderResponse(Message::Code aCode, const Message &aReques
 
     default:
         ExitNow(error = OT_ERROR_INVALID_ARGS);
-        break;
+        OT_UNREACHABLE_CODE(break;)
```
jwhui, line 312: I wonder if it is necessary to keep the `break` after `ExitNow()` statements.

src/core/config/openthread-core-config-check.h
```diff
@@ -42,6 +42,10 @@
 #error "OPENTHREAD_ENABLE_DHCP6_MULTICAST_SOLICIT requires OPENTHREAD_CONFIG_DHCP6_CLIENT_ENABLE to be also set."
 #endif
 
+#if OPENTHREAD_ENABLE_DHCP6_MULTICAST_SOLICIT && OPENTHREAD_CONFIG_DHCP6_SERVER_ENABLE
```
jwhui, line 45: Is the check appropriate? `OPENTHREAD_ENABLE_DHCP6_MULTICAST_SOLICIT` is primarily for the client side implementation.

include/openthread/link_raw.h
srickardti: Yep, the header didn't match the source file which caused linktime errors. I was a little confused why a `uint6_t` was being marked const, but I figured adding consts was better than removing them.

src/core/coap/coap.cpp
srickardti: I had figured it was a code style requirement in any switch statement to avoid possible fall throughs.

src/core/config/openthread-core-config-check.h
srickardti: We thought it would be incorrect to have both enabled at the same time. This was a way to catch our customers to remind them to enable dnsmasq on the border router. That is admittedly a special case for our examples.

include/openthread/link_raw.h
jwhui: Thanks for pointing that out. In this case, I think we should remove const. Indeed, we removed some of them in 71604f6cd0e19b6f3999ed949924c8691e922e17. However, after looking, it seems there are other instances. I'll submit a separate PR to clean this up more generally.

src/core/config/openthread-core-config-check.h
jwhui: Makes sense. I am OK to keep this check.

include/openthread/link_raw.h
jwhui: @srickardti, just submitted #4901. Please review.

src/core/coap/coap.cpp
jwhui: Yeah, I think I remember needing to include these to avoid static code analysis warnings as well.

src/core/mac/mac_frame.cpp
```diff
@@ -836,7 +836,7 @@ uint8_t Frame::SkipAddrFieldIndex(void) const
 
     default:
         ExitNow(index = kInvalidIndex);
-        break;
+        OT_UNREACHABLE_CODE(break;);
```
jwhui, line 839: nit: in some cases, I see:

- `OT_UNREACHABLE_CODE(break;);`

In others, I see:

- `OT_UNREACHABLE_CODE(break;)`

We should be consistent. I would prefer a single semi-colon and after the closing parent, like:

- `OT_UNREACHABLE_CODE(break);`.

Thoughts?

srickardti: Done. The original code looked like this:
```
OT_UNREACHABLE_CODE(
    break;
);
```
but make pretty compacted it to one line.

PR #4886, opened 2020-04-25
[pretty] add support for shell scripts

jwhui:


script/check-size
```diff
@@ -128,67 +128,67 @@ size_nrf52840()
...
     fi
 
-    mkdir -p "${OT_TMP_DIR}/a"
-    git archive "${OT_SHA_OLD}" | tar x -C "${OT_TMP_DIR}/a"
-    (cd "${OT_TMP_DIR}/a" \
-        && ./bootstrap \
-        && make -f examples/Makefile-nrf52840 "${flags[@]}")
+    mkdir -p "$OT_TMP_DIR/a"
+    git archive "$OT_SHA_OLD" | tar x -C "$OT_TMP_DIR/a"
+    (cd "$OT_TMP_DIR/a" &&
```
bukepo, line 148: Google's shell style guide suggests the original style guide. https://google.github.io/styleguide/shellguide.html#pipelines
Not sure why.

script/test
```diff
@@ -239,23 +244,23 @@ do_expect()
...
     fi
 
     while read -r script; do
         rm -rf tmp
-        OT_COMMAND="${ot_command}" "${script}" || {
+        OT_COMMAND="$ot_command" "$script" || {
             local exit_code=$?
-            echo -e "${COLOR_FAIL}FAIL${COLOR_NONE} ${script}"
-            exit "${exit_code}"
+            echo -e "${COLOR_FAIL}FAIL$COLOR_NONE $script"
```
bukepo, line 257: prefer `${var}` over `$var` from https://google.github.io/styleguide/shellguide.html#variable-expansion

jwhui: `shellharden` seems to want this change. Unless you can suggest how to configure `shellharden` otherwise, we can either:

1. Remove `shellharden` from `script/make-pretty`

2. Accept this style.

Thoughts?

bukepo: Since bash is not easy to debug, so I prefer the curly style which is less error prone.

I tested with the following script, no warnings are given.

```bash
a=1
ab=1
a_b=1
echo "$a_b"
echo "$a"
echo "$ab"
```

Thoughts?

jwhui:
Superseded by #4966.

PR #4884, opened 2020-04-24
[mac-frame] add util functions to generate ack

Irving-cl:
This PR adds mac frame util functions to generate acks(both Imm and Enh acks) from a received frame:
```
    /**
     * Generate Imm-Ack in this frame object.
     *
     * @param[in]    aFrame             A pointer to the incoming frame.
     * @param[in]    aIsFramePending    If the ACK's frame pending bit should be set.
     *
     */
    void GenerateImmAck(const Frame *aFrame, bool aIsFramePending);

    /**
     * Generate Enh-Ack in this frame object.
     *
     * @param[in]    aFrame             A pointer to the frame.
     * @param[in]    aIsFramePending    If the ACK's frame pending bit should be set.
     * @param[in]    aIeData            A pointer to the IE data pattern of the ACK to be sent.
     * @param[in]    aIeLength          The length of IE data pattern of the ACK to be sent.
     *
     */
    void GenerateEnhAck(const Frame *aFrame, bool aIsFramePending, const uint8_t *aIeData, uint8_t aIeLength);
```
These funtions would be used as utils API for platforms which don't compose ack themselves to generate acks. And in 1.2, our simulation would use these functions to generate acks.

And in this PR, the compile options for CSL is introduced to control the CSL IE related part.

PR #4887, opened 2020-04-25
[mac-frame] helper methods to calculate size of headers/MIC

abtink:
This commit adds new helper methods in `Mac::Frame` to calculate the
size (number of bytes) of headers or security MIC. These are used by
different methods to simplify the code and avoid repeated patterns:
- `CalculateAddrFieldSize(aFcf)` to calculate the size of Address
  Fields given a Frame Control value.
- `CalculateSecurityHeaderSize(aSecurityControl)` to calculate size
  of Security Header given a Security Control value.
- `CalculateMicSize(aSecurityControl)` to get the size of MIC given
  a Security Control value.

### DIFF
--- a/src/core/mac/mac_frame.cpp
+++ b/src/core/mac/mac_frame.cpp
@@ -49,101 +49,33 @@ namespace Mac {
 void Frame::InitMacHeader(uint16_t aFcf, uint8_t aSecurityControl)
 {
     uint8_t *bytes = GetPsdu();
-
-    mLength = 0;
+    uint8_t  length;
 
     // Frame Control Field
     Encoding::LittleEndian::WriteUint16(aFcf, bytes);
-    mLength += kFcfSize;
 
-    // Sequence Number
-    mLength += kDsnSize;
-
-    // Destination PAN
-    if (IsDstPanIdPresent(aFcf))
-    {
-        mLength += sizeof(PanId);
-    }
-
-    // Destination Address
-    switch (aFcf & kFcfDstAddrMask)
-    {
-    case kFcfDstAddrNone:
-        break;
-
-    case kFcfDstAddrShort:
-        mLength += sizeof(ShortAddress);
-        break;
-
-    case kFcfDstAddrExt:
-        mLength += sizeof(ExtAddress);
-        break;
-
-    default:
-        OT_ASSERT(false);
-    }
-
-    // Source PAN
-    if (IsSrcPanIdPresent(aFcf))
-    {
-        mLength += sizeof(PanId);
-    }
-
-    // Source Address
-    switch (aFcf & kFcfSrcAddrMask)
-    {
-    case kFcfSrcAddrNone:
-        break;
-
-    case kFcfSrcAddrShort:
-        mLength += sizeof(ShortAddress);
-        break;
-
-    case kFcfSrcAddrExt:
-        mLength += sizeof(ExtAddress);
-        break;
-
-    default:
-        OT_ASSERT(false);
-    }
+    length = CalculateAddrFieldSize(aFcf);
+    OT_ASSERT(length != kInvalidSize);
 
     // Security Header
     if (aFcf & kFcfSecurityEnabled)
     {
-        bytes[mLength] = aSecurityControl;
+        bytes[length] = aSecurityControl;
 
-        if (aSecurityControl & kSecLevelMask)
-        {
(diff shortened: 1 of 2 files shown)


src/core/mac/mac_frame.cpp
```diff
@@ -769,104 +706,125 @@ void Frame::SetPayloadLength(uint16_t aLength)
 
 uint8_t Frame::SkipSecurityHeaderIndex(void) const
```
abtink, line 707: Suggest to use the "Hide whitespace changes" option to look at the diff here and below.

abtink:
@Irving-cl just checking your PR https://github.com/openthread/openthread/pull/4884 (adding enhanced ack) noticed you have added some similar methods (e.g., Get MIC size or security header size) from that PR. Can you please help review this PR?

src/core/mac/mac_frame.cpp
```diff
@@ -769,104 +706,125 @@ void Frame::SetPayloadLength(uint16_t aLength)
...
-    uint8_t  index = kFcfSize + kDsnSize; // Frame Control field and Sequence Number field
-    uint16_t fcf;
+    uint8_t size = kFcfSize + kDsnSize;
 
-    VerifyOrExit((index + GetFcsSize()) <= GetPsduLength(), index = kInvalidIndex);
-    fcf = GetFrameControlField();
+    // This static method calculates the size (number of bytes) of
+    // Address header filed for a given Frame Control `aFcf` value.
+    // The size includes the Frame Control and Sequence Number fields
+    // along with Destination and Source PAN ID and Short/Extended
```
Irving-cl, line 776: What about only excluding the `Frame Control` and `Sequence Number` from this size? 
* First I think in this way its behavior fits its name more. 
* Secondly this function currently has the same behavior with `SkipAddrFieldIndex`, the only difference is one takes a parameter `aFcf`. And I think this is different from the case, for example, `IsVersion2015()` has two versions, where `IsVersion2015(void)` is public and will be used externally. Both `SkipAddrFieldIndex` and `CalculateAddrFieldSize` are only used internally. So I think we can make them for different behaviors.

src/core/mac/mac_frame.hpp
```diff
@@ -528,7 +528,7 @@ class Frame : public otRadioFrame
      * @returns TRUE if the Src PanId is present, FALSE otherwise.
      *
      */
-    bool IsSrcPanIdPresent(uint16_t aFcf) const;
+    static bool IsSrcPanIdPresent(uint16_t aFcf);
```
Irving-cl, line 531: Shall we move it into private, the same position as `IsDstPanIdPresent`? This function is actually not used externally.

Irving-cl:
On a second thought, I think there's another difference for `SkipAddrFieldIndex` and `CalculateAddrFieldSize`:
`CalculateAddrFieldSize` could be called when the frame is not fully constructed while `SkipAddrFieldIndex`should be only called when the frame is complete, the length should be correctly set. 
Our helper functions are used in both cases to handle a received frame and to generate a frame. Maybe we can divide these helper functions: some can be used when the frame is still constructing and some can only be used to deal a complete frame(it will check any errors for the frames). Thoughts? @abtink

abtink:
Thanks @Irving-cl for review. :)

> Our helper functions are used in both cases to handle a received frame and to generate a frame. Maybe we can divide these helper functions: some can be used when the frame is still constructing and some can only be used to handle a complete frame(it will check any errors for the frames). Thoughts? @abtink 

Fully agree. The `static` helper methods should be generic and can be used safely when preparing a frame (or parsing it). The (non-`static`) helper methods (like `SkipAddrFieldIndex` or `SkipSecurityHeaderIndex`) should be used only when frame is fully formed (e.g., parsing a frame). 

> On a second thought, I think there's another difference for SkipAddrFieldIndex and CalculateAddrFieldSize:
> CalculateAddrFieldSize could be called when the frame is not fully constructed while SkipAddrFieldIndexshould be only called when the frame is complete, the length should be correctly set.
> 

Again agre.  I would suggest we keep the behavior in terms of including FCF and DSN the same for these (i.e either both include FCF/DSN or both exclude them) and main difference in how/where we can use each.

Irving-cl:
> Again agre. I would suggest we keep the behavior in terms of including FCF and DSN the same for these (i.e either both include FCF/DSN or both exclude them) and main difference in how/where we can use each.

Ok. I think it's good to do it in this way.

PR #4889, opened 2020-04-26
[mac] fix warning in mac_types.hpp

bukepo:


abtink:
Thanks @bukepo for fix.

PR #4892, opened 2020-04-26
[cmake] detect big endian

bukepo:
Endianness detection can't work with `-Wl,--gc-sections`, so move it out from toolchain configuration.


examples/platforms/cc1352/CMakeLists.txt
```diff
@@ -79,6 +79,7 @@ target_link_options(openthread-cc1352
 target_link_options(openthread-cc1352
     PUBLIC
         -Wl,-Map=$<TARGET_PROPERTY:NAME>.map
+        -Wl,--gc-sections
```
jwhui, line 82:
```suggestion
        -Wl,--gc-sections
        -Wl,-Map=$<TARGET_PROPERTY:NAME>.map
```

PR #4894, opened 2020-04-27
[aloc] add ApplyMeshLocalPrefix() to applicable ALOCs

librasungirl:
MeshLocalPrefix may change when applying Operational Dataset.
This PR applies mesh local prefix to the ALOCs which are not in `Mle` component.

### DIFF
--- a/src/core/net/dhcp6_server.cpp
+++ b/src/core/net/dhcp6_server.cpp
@@ -64,7 +64,7 @@ otError Dhcp6Server::UpdateService(void)
     Lowpan::Context                 lowpanContext;
 
     // remove dhcp agent aloc and prefix delegation
-    for (int i = 0; i < OPENTHREAD_CONFIG_DHCP6_SERVER_NUM_PREFIXES; i++)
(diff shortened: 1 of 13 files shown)


src/core/net/dhcp6_server.cpp
```diff
@@ -64,7 +64,7 @@ otError Dhcp6Server::UpdateService(void)
     Lowpan::Context                 lowpanContext;
 
     // remove dhcp agent aloc and prefix delegation
-    for (int i = 0; i < OPENTHREAD_CONFIG_DHCP6_SERVER_NUM_PREFIXES; i++)
+    for (uint8_t i = 0; i < OT_ARRAY_LENGTH(mPrefixAgents); i++)
```
bukepo, line 67: Size_t instead of uint8 to avoid conversation

librasungirl: Pushed the second commit to unify the type of local variables when `OT_ARRAY_LENGTH` is used. Please review.

PR #4896, opened 2020-04-27
[spinel] fix time narrowing conversion

Moandor-y:
This makes narrowing conversions of time types explicit to fix build errors.

Also added a CI build to check for any errors like this. The new build uses clang because GCC's `-Wconversion` is too strict, making a huge number of compiling errors in our codebase.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -137,6 +137,28 @@ jobs:
         script/check-simulation-build
         script/check-posix-build
 
+  clang-m32:
+    name: clang-m32-${{ matrix.clang_ver }}
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        clang_ver: [9]
+    env:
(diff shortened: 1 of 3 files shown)


jwhui:
> This makes narrowing conversions of time types explicit to fix build errors.

@Moandor-y, thank you for this fix. Our continuous-integration checks typically catch these issues. Can we add a test for this specific scenario?

Moandor-y:
> > This makes narrowing conversions of time types explicit to fix build errors.
> 
> @Moandor-y, thank you for this fix. Our continuous-integration checks typically catch these issues. Can we add a test for this specific scenario?

This issue only happens when building a 32-bit target with `-Wconversion` on. Pushed the commit https://github.com/openthread/openthread/pull/4896/commits/2f31e876f22ff130a26e93c4459af756f08eb7a3 to add a build test for this scenario.

PR #4899, opened 2020-04-27
[commissioner] fix duplicate commissioner start

simonlingoogle:
This PR fixes the bug that duplicate `commissioner start` will incorrectly close `CoapSecure`.

### DIFF
--- a/src/core/meshcop/commissioner.cpp
+++ b/src/core/meshcop/commissioner.cpp
@@ -170,7 +170,7 @@ otError Commissioner::Start(otCommissionerStateCallback  aStateCallback,
     SetState(OT_COMMISSIONER_STATE_PETITION);
 
 exit:
-    if (error != OT_ERROR_NONE)
+    if (error != OT_ERROR_NONE && error != OT_ERROR_ALREADY)
     {
         otLogWarnMeshCoP("Failed to start commissioner: %s", otThreadErrorToString(error));
         Get<Coap::CoapSecure>().Stop();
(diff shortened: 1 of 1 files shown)


src/core/meshcop/commissioner.cpp
```diff
@@ -170,7 +170,7 @@ otError Commissioner::Start(otCommissionerStateCallback  aStateCallback,
     SetState(OT_COMMISSIONER_STATE_PETITION);
 
 exit:
-    if (error != OT_ERROR_NONE)
+    if (error != OT_ERROR_NONE && error != OT_ERROR_ALREADY)
```
LuDuda, line 173: Good catch. Can we, however, apply this if statement for CoapSecure.Stop() method and leave logs as they are now both in start and stop?

simonlingoogle: Don't have strong opinion on removing the log for `OT_ERROR_ALREADY`.  
But if we are treating `OT_ERROR_ALREADY` same as success, should we better keep this rule even in logging?

LuDuda: Also don't have very strong opinion, but leaning towards leaving the log, so while debugging we know someone tried to "rerun" commissioner.

simonlingoogle: Log kept for `OT_ERROR_ALREADY`

simonlingoogle:
@LuDuda any further comments?

LuDuda:
@simonlingoogle Thank you so much!

PR #4901, opened 2020-04-27
[style] remove const from uint16_t function/method params

jwhui:
As identified by @srickardti in #4885.


srickardti:
@jwhui There was one other const in the PR that isn't caught here. ` src/core/common/logging.hpp:1879` has a const on a `size_t`. would you like to get that one as well?

jwhui:
> @jwhui There was one other const in the PR that isn't caught here. ` src/core/common/logging.hpp:1879` has a const on a `size_t`. would you like to get that one as well?

@srickardti, I believe that was addressed in 71604f6cd0e19b6f3999ed949924c8691e922e17.

srickardti:
> @srickardti, I believe that was addressed in [71604f6](https://github.com/openthread/openthread/commit/71604f6cd0e19b6f3999ed949924c8691e922e17).

Yep, you're right. I just hadn't updated my branch.

PR #4902, opened 2020-04-28
[commissioner] remove duplicate MGMT_COMMISSISONER_SET.req

simonlingoogle:
This PR:
1. Removes the unnecessary `MGMT_COMMISSISONER_SET.req` sent by Commissioner to Leader when a Joiner is removed only to be added back. 
2. Use proper error codes for `MleRouter::HandleDiscoveryRequest`

### DIFF
--- a/src/core/meshcop/commissioner.cpp
+++ b/src/core/meshcop/commissioner.cpp
@@ -348,11 +348,11 @@ exit:
     return error;
 }
 
-otError Commissioner::RemoveJoiner(const Mac::ExtAddress *aEui64, uint32_t aDelay, JoinerOpFlag flags)
+otError Commissioner::RemoveJoiner(const Mac::ExtAddress *aEui64, uint32_t aDelay, JoinerOpFlag aFlags)
 {
     otError error = OT_ERROR_NOT_FOUND;
 
-    OT_ASSERT(!(flags & kJoinerOpFlagNotNotifyLeader) || aDelay == 0);
+    OT_ASSERT(!(aFlags & kJoinerOpFlagNotNotifyLeader) || aDelay == 0);
 
     VerifyOrExit(mState == OT_COMMISSIONER_STATE_ACTIVE, error = OT_ERROR_INVALID_STATE);
 
@@ -391,7 +391,7 @@ otError Commissioner::RemoveJoiner(const Mac::ExtAddress *aEui64, uint32_t aDela
 
             joiner->mValid = false;
             UpdateJoinerExpirationTimer();
-            if ((flags & kJoinerOpFlagNotNotifyLeader) == 0)
+            if ((aFlags & kJoinerOpFlagNotNotifyLeader) == 0)
             {
                 SendCommissionerSet();
             }
--- a/src/core/meshcop/commissioner.cpp
+++ b/src/core/meshcop/commissioner.cpp
@@ -284,7 +284,7 @@ otError Commissioner::AddJoiner(const Mac::ExtAddress *aEui64, const char *aPskd
 
     VerifyOrExit(StringLength(aPskd, Dtls::kPskMaxLength + 1) <= Dtls::kPskMaxLength, error = OT_ERROR_INVALID_ARGS);
 
-    RemoveJoiner(aEui64, 0); // remove immediately
+    RemoveJoiner(aEui64, 0, /* aNotifyLeader */ false); // remove immediately
 
     for (Joiner *joiner = &mJoiners[0]; joiner < OT_ARRAY_END(mJoiners); joiner++)
     {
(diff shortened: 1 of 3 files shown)


src/core/thread/mle_router.cpp
```diff
@@ -2836,7 +2836,8 @@ otError MleRouter::HandleDiscoveryRequest(const Message &aMessage, const Ip6::Me
         case MeshCoP::Tlv::kExtendedPanId:
             aMessage.Read(offset, sizeof(extPanId), &extPanId);
             VerifyOrExit(extPanId.IsValid(), error = OT_ERROR_PARSE);
-            VerifyOrExit(Get<Mac::Mac>().GetExtendedPanId() != extPanId.GetExtendedPanId(), OT_NOOP);
+            VerifyOrExit(Get<Mac::Mac>().GetExtendedPanId() != extPanId.GetExtendedPanId(),
+                         error = OT_ERROR_ADDRESS_FILTERED);
```
wgtdkp, line 2840: use `OT_ERROR_DROP`?

simonlingoogle: good idea

src/core/meshcop/commissioner.cpp
```diff
@@ -284,7 +284,7 @@ otError Commissioner::AddJoiner(const Mac::ExtAddress *aEui64, const char *aPskd
 
     VerifyOrExit(StringLength(aPskd, Dtls::kPskMaxLength + 1) <= Dtls::kPskMaxLength, error = OT_ERROR_INVALID_ARGS);
 
-    RemoveJoiner(aEui64, 0); // remove immediately
+    RemoveJoiner(aEui64, 0, /* aNotifyLeader */ false); // remove immediately
```
jwhui, line 287: nit: generally prefer having an enum rather than bool here to avoid the need for having the comment

src/core/meshcop/commissioner.cpp
```diff
@@ -389,7 +391,10 @@ otError Commissioner::RemoveJoiner(const Mac::ExtAddress *aEui64, uint32_t aDela
 
             joiner->mValid = false;
             UpdateJoinerExpirationTimer();
-            SendCommissionerSet();
+            if (aNotifyLeader)
+            {
+                SendCommissionerSet();
```
jwhui, line 396: Long term, I think it is more appropriate to pull this out of `RemoveJoiner()`. Doing so would allow for cases where multiple joiners are removed at the same time. In that case, only one message needs to be sent.

src/core/meshcop/commissioner.cpp
simonlingoogle: revised the code to use `enum JoinerOperationFlag`, please check

src/core/meshcop/commissioner.cpp
simonlingoogle: I was thinking about using a new private method like `RemoveJoinerImpl` for only modifying the `mJoiners`. But later I was convinced by myself to only add a bool argument.

librasungirl: Agree it would be better to aggregate `MGMT_COMMISSIONER_SET.req` for the addition/removal of multiple joiners into one.

Some options in my mind are
1) split the `MGMT_COMMISSIONER_SET.req` from Add/Remove Joiner operation, specifically
    -  no MGMT_COMMISSIONER_SET.req when adding/removing Joiner;
    -  introduce one API like `otCommissionerDatasetRegister()` which will call `Commissioner::SendCommissionerSet(void)` to construct one MGMT_COMMISSIONER_SET.req

     The side effect is that we should add note for the user that this extra API should be explicitly called to let the addition/removal of joiner to take effect.

2)  introduce some delay for `MGMT_COMMISSIONER_SET.req` per addition/removal operation, similar as existing aggregation solution for Child Update Request.

Thoughts?

simonlingoogle: I think splitting the functionalities of one `commissioner joiner add` command into multiple commands make things unnecessarily more complex for users. 
I think 2) is a good idea, which we can enhance accordingly in another separate PR.

src/core/meshcop/commissioner.hpp
```diff
@@ -128,15 +138,16 @@ class Commissioner : public InstanceLocator
     /**
      * This method removes a Joiner entry.
      *
-     * @param[in]  aEui64          A pointer to the Joiner's IEEE EUI-64 or NULL for any Joiner.
-     * @param[in]  aDelay          The delay to remove Joiner (in seconds).
+     * @param[in]  aEui64         A pointer to the Joiner's IEEE EUI-64 or NULL for any Joiner.
+     * @param[in]  aDelay         The delay to remove Joiner (in seconds).
+     * @param[in]  aNotifyLeader  Whether or not to send MGMT_COMMISSISONER_SET.req to notify Leader.
```
librasungirl, line 143: update the parameter doxygen

PR #2688, opened 2018-04-27
[MAC] Provide a mechanism to start/stop the MAC according to the state of the Interface.

pvanhorn:
To ensure proper and expedient shutdown of Openthread when the Interface is brought down, this change provides a mechanism to gracefully stop the MAC layer operations.  Scans are stopped by preventing any new channel change.  New Transmissions are stopped. Existing transmissions are allowed to complete. Any pending operations are cleared.

PR #4903, opened 2020-04-28
[mac] disable SubMac on instance finalization

e-rk:
When OT is initialized after boot, finalized, and initialized again without reboot, assertion in SubMac::Enable is triggered. The reason is that SubMac::Disable is never called during finalization and platform radio is in unexpected state.

### DIFF
--- a/src/core/common/instance.cpp
+++ b/src/core/common/instance.cpp
@@ -196,6 +196,8 @@ void Instance::Finalize(void)
     Get<Settings>().Deinit();
 #endif
 
+    Get<Mac::SubMac>().Disable();
+
 #if !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
 
     /**
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -73,7 +73,7 @@ const char Mac::sDomainNameInit[] = "Thread";
 
 Mac::Mac(Instance &aInstance)
     : InstanceLocator(aInstance)
-    , mEnabled(true)
+    , mEnabled(false)
     , mPendingActiveScan(false)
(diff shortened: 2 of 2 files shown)


abtink:
Thanks @e-rk  for this change.

A bit of background of the `Mac::SetEnabled()` and `otLinkSetEnabled()`: This was added originally from #2688 with the goal to just stop the MAC layer operations while keeping the radio in same state as before (this was used in an internal projects to get control of radio temporarily from OT stack).  This PR would change that behavior and would disable/stop the sub-mac/radio as well from `otLinkSetEnabled()`. Another approach to this PR is to call `SubMac::Disable()` from `Finalize()`?

Having said that, I personally don't mind if we change the behavior of `otLinkSetEnabled()` but should be careful about its use/impact (which may later require some changes in our internal code). Thoughts? @jwhui @bukepo

e-rk:
@abtink Would a reference counter (incremented after every `SubMac::Enable`) be a suitable solution here?

abtink:
@e-rk that can work, but I'd say when finalizing and restarting OT, it may be better to also enable/disable the radio again as well. 

> When OT is initialized after boot, finalized, and initialized again without reboot, assertion in SubMac::Enable is triggered. The reason is that SubMac::Disable is never called during finalization and platform radio is in unexpected state

I would say the quick fix for the above issue is to call `Get<Mac::SubMac>().Disable();` in `Instance::Finalize()` (it can be outside `#if OPENTHREAD_MTD || OPENTHREAD_FTD` check since it would apply in all situations). Thoughts?

e-rk:
@e-rk I see. Some nasty stuff could happen with careless refcounting. I can go with disabling `SubMac` in `Instance::Finalize()`. Thanks for giving me context on the situation.

PR #4904, opened 2020-04-28
[dua] save dad counter into non volatile memory

librasungirl:
This PR depends on #4900 (<-- #4862 <-- #4854).

Please review the last commit which serves the purpose of storing dad counter.

Due to that DUA DAD is not submitted yet, for this PR, just manually verified the dad counter storage and consistency of DUA address generated after reset on another fork branch [librasungirl:12_dua_store_dadcounter_test_only](https://github.com/librasungirl/openthread/commits/12_dua_store_dadcounter_test_only) which introduces some test APIs to manually update (and store) the dad counter.

### DIFF
--- a/src/core/common/settings.cpp
+++ b/src/core/common/settings.cpp
@@ -81,7 +81,7 @@ void SettingsBase::LogChildInfo(const char *aAction, const ChildInfo &aChildInfo
 #if OPENTHREAD_CONFIG_DUA_ENABLE
 void SettingsBase::LogDadInfo(const char *aAction, const DadInfo &aDadInfo) const
 {
-    otLogInfoCore("Non-volatile: %s DadInfo {DadCounter:0x%2d}", aAction, aDadInfo.mDadCounter);
+    otLogInfoCore("Non-volatile: %s DadInfo {DadCounter:%2d}", aAction, aDadInfo.mDadCounter);
 }
 #endif
 
--- a/src/core/common/settings.cpp
+++ b/src/core/common/settings.cpp
@@ -78,6 +78,13 @@ void SettingsBase::LogChildInfo(const char *aAction, const ChildInfo &aChildInfo
                   aChildInfo.GetMode(), aChildInfo.GetVersion());
 }
 
+#if OPENTHREAD_CONFIG_DUA_ENABLE
+void SettingsBase::LogDadInfo(const char *aAction, const DadInfo &aDadInfo) const
+{
+    otLogInfoCore("Non-volatile: %s DadInfo {DadCounter:0x%2d}", aAction, aDadInfo.mDadCounter);
+}
+#endif
+
 #endif // #if (OPENTHREAD_CONFIG_LOG_LEVEL >= OT_LOG_LEVEL_INFO)
 
 #if (OPENTHREAD_CONFIG_LOG_LEVEL >= OT_LOG_LEVEL_WARN)
@@ -394,6 +401,54 @@ void Settings::ChildInfoIterator::Read(void)
 exit:
     mIsDone = (error != OT_ERROR_NONE);
 }
+
+#if OPENTHREAD_CONFIG_DUA_ENABLE
+otError Settings::ReadDadInfo(DadInfo &aDadInfo) const
+{
+    otError  error;
+    uint16_t length = sizeof(DadInfo);
+
+    aDadInfo.Init();
+    SuccessOrExit(error = Read(kKeyDadInfo, &aDadInfo, length));
+    LogDadInfo("Read", aDadInfo);
+
+exit:
+    return error;
+}
+
+otError Settings::SaveDadInfo(const DadInfo &aDadInfo)
+{
+    otError  error = OT_ERROR_NONE;
+    DadInfo  prevDadInfo;
+    uint16_t length = sizeof(DadInfo);
+
+    if ((Read(kKeyDadInfo, &prevDadInfo, length) == OT_ERROR_NONE) && (length == sizeof(DadInfo)) &&
+        (memcmp(&prevDadInfo, &aDadInfo, sizeof(DadInfo)) == 0))
+    {
+        LogDadInfo("Re-saved", aDadInfo);
+        ExitNow();
+    }
+
+    SuccessOrExit(error = Save(kKeyDadInfo, &aDadInfo, sizeof(DadInfo)));
+    LogDadInfo("Saved", aDadInfo);
+
+exit:
+    LogFailure(error, "saving DadInfo", false);
+    return error;
+}
+
+otError Settings::DeleteDadInfo(void)
+{
+    otError error;
+
+    SuccessOrExit(error = Delete(kKeyDadInfo));
+    otLogInfoCore("Non-volatile: Deleted DadInfo");
+
+exit:
+    LogFailure(error, "deleting DadInfo", true);
+    return error;
+}
+#endif // OPENTHREAD_CONFIG_DUA_ENABLE
 
 otError Settings::Read(Key aKey, void *aBuffer, uint16_t &aSize) const
 {
--- a/src/core/common/settings.hpp
+++ b/src/core/common/settings.hpp
@@ -539,6 +539,40 @@ public:
     } OT_TOOL_PACKED_END;
 
     /**
+     * This structure represents the duplicate address detection information for settings storage.
+     *
+     */
+    OT_TOOL_PACKED_BEGIN
+    class DadInfo
+    {
+    public:
+        /**
+         * This method clears the struct object (setting all the fields to zero).
+         *
+         */
+        void Init(void) { memset(this, 0, sizeof(*this)); }
+
+        /**
+         * This method returns the Dad Counter.
+         *
+         * @returns The Dad Counter value.
+         *
+         */
+        uint8_t GetDadCounter(void) const { return mDadCounter; }
+
+        /**
+         * This method sets the Dad Counter.
+         *
+         * @param[in] aDadCounter The Dad Counter value.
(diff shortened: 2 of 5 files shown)


src/core/thread/dua_manager.cpp
```diff
@@ -104,11 +105,18 @@ void DuaManager::UpdateDomainUnicastAddress(BackboneRouter::Leader::DomainPrefix
...
 otError DuaManager::GenerateDomainUnicastAddressIid(void)
 {
-    otError error;
+    otError  error;
+    uint16_t dadCounter = mDadCounter;
 
     if ((error = Get<Utils::Slaac>().GenerateIid(mDomainUnicastAddress, NULL, 0, &mDadCounter,
                                                  false /* not allow random IID if no valid IID */)) == OT_ERROR_NONE)
     {
+        if (dadCounter != mDadCounter)
```
simonlingoogle, line 114: Why change `mDadCounter` back to `dadCounter`?
Should pass `&dadCounter` to `Get<Utils::Slaac>().GenerateIid` instead of `&mDadCounter`?

src/core/common/settings.hpp
```diff
@@ -550,6 +589,7 @@ class SettingsBase : public InstanceLocator
         kKeyChildInfo         = 0x0005, ///< Child information
         kKeyReserved          = 0x0006, ///< Reserved (previously auto-start)
         kKeySlaacIidSecretKey = 0x0007, ///< Secret key used by SLAAC module for generating semantically opaque IID
+        kKeyDadInfo           = 0x0008, ///< Duplicate Address Detection (DAD) information.
```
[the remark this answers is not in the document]
librasungirl: Not sure if it is a good thing to name it too specific.
Note that there is also duplicate address detection mechanism in existing OT (Address Query), however when conflicts happens, only remove the address from Thread interface, no future avoidance mechanism when generating the address via SLAAC again.
Though it is not so clear how to achieve such avoidance mechanism for the other on-mesh prefixes so far (need to match the prefix and their dad counter), maybe we can keep the name to be generic as is now to reserve the possibility to extend the duplicate avoidance mechanism?

src/core/thread/dua_manager.cpp
librasungirl: oh, yes, careless mistake. will update.

src/core/common/settings.hpp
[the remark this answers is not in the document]
simonlingoogle: Fine to extend in the future.

src/core/common/settings.cpp
```diff
@@ -395,6 +402,54 @@ void Settings::ChildInfoIterator::Read(void)
...
+{
+    otError  error;
+    uint16_t length = sizeof(DadInfo);
+
+    aDadInfo.Init();
+    SuccessOrExit(error = Read(kKeyDadInfo, &aDadInfo, length));
+    LogDadInfo("Read", aDadInfo);
+
+exit:
+    return error;
```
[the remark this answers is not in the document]
librasungirl: no `LogFailure` for Read Op in existing settings.
Guess there is no information stored for a fresh device, so `LogFailure` is skipped for Read as may not care so much as the other operations.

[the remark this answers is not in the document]
simonlingoogle: Should call `ReadDadInfo` whenever node is booted? 

Noted that other `ReadXxxxInfo` methods do not have `LogFailure` either, not knowing why. @abtink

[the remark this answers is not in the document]
abtink: It can be added but as @librasungirl mentioned it does not add much new info/value to log error on read operation. 

When reading a setting entry, it's legitimate to not have the setting entry and thus returning `OT_ERROR_NOT_FOUND` (which is not really an error and more an indication that such entry does not exist). We need to handle this error as a special case in the logs since it is not useful to log such error (it'd just bloat the logs).

librasungirl:
Rebased

src/core/common/settings.cpp
```diff
@@ -78,6 +78,13 @@ void SettingsBase::LogChildInfo(const char *aAction, const ChildInfo &aChildInfo
                   aChildInfo.GetMode(), aChildInfo.GetVersion());
 }
 
+#if OPENTHREAD_CONFIG_DUA_ENABLE
+void SettingsBase::LogDadInfo(const char *aAction, const DadInfo &aDadInfo) const
+{
+    otLogInfoCore("Non-volatile: %s DadInfo {DadCounter:0x%2d}", aAction, aDadInfo.mDadCounter);
```
abtink, line 84:
```suggestion
    otLogInfoCore("Non-volatile: %s DadInfo {DadCounter:0x%02x}", aAction, aDadInfo.mDadCounter);
```
Sicne we have `0x` logging it as hex.
(or we should remove the `0x` and log as `%d`?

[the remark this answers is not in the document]
librasungirl: Note that `Core` region is generally used in `settings`, in which no specific region per feature. @simonlingoogle

librasungirl: good catch. will remove `0x`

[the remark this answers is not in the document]
simonlingoogle: sure, thanks for the tip

librasungirl:
Updated this PR on top of current #4904 and added the last commit to resolve the comments here.

src/core/common/settings.cpp
```diff
@@ -395,6 +402,54 @@ void Settings::ChildInfoIterator::Read(void)
...
+otError Settings::SaveDadInfo(const DadInfo &aDadInfo)
+{
+    otError  error = OT_ERROR_NONE;
+    DadInfo  prevDadInfo;
+    uint16_t length = sizeof(DadInfo);
+
+    if ((Read(kKeyDadInfo, &prevDadInfo, length) == OT_ERROR_NONE) && (length == sizeof(DadInfo)) &&
+        (memcmp(&prevDadInfo, &aDadInfo, sizeof(DadInfo)) == 0))
+    {
+        LogDadInfo("Re-saved", aDadInfo);
```
jwhui, line 428: nit: since nothing was written to non-volatile, how about something like:
```suggestion
        LogDadInfo("Save skipped", aDadInfo);
```

src/core/common/settings.hpp
```diff
@@ -563,11 +598,17 @@ class SettingsBase : public InstanceLocator, private NonCopyable
     void LogNetworkInfo(const char *aAction, const NetworkInfo &aNetworkInfo) const;
     void LogParentInfo(const char *aAction, const ParentInfo &aParentInfo) const;
     void LogChildInfo(const char *aAction, const ChildInfo &aChildInfo) const;
-#else
+#if OPENTHREAD_CONFIG_DUA_ENABLE
+    void LogDadInfo(const char *aAction, const DadInfo &aDadInfo) const;
+#endif // OPENTHREAD_CONFIG_DUA_ENABLE
```
jwhui, line 603:
```suggestion
#endif
```

src/core/common/settings.hpp
```diff
@@ -563,11 +598,17 @@ class SettingsBase : public InstanceLocator, private NonCopyable
...
+#else  // (OPENTHREAD_CONFIG_LOG_LEVEL >= OT_LOG_LEVEL_INFO) && (OPENTHREAD_CONFIG_LOG_UTIL != 0)
     void LogNetworkInfo(const char *, const NetworkInfo &) const {}
     void LogParentInfo(const char *, const ParentInfo &) const {}
     void LogChildInfo(const char *, const ChildInfo &) const {}
-#endif
+#if OPENTHREAD_CONFIG_DUA_ENABLE
+    void LogDadInfo(const char *, const DadInfo &) const {}
+#endif // OPENTHREAD_CONFIG_DUA_ENABLE
```
jwhui, line 610:
```suggestion
#endif
```

src/core/common/settings.hpp
```diff
@@ -864,6 +905,42 @@ class Settings : public SettingsBase
...
+     */
+    otError SaveDadInfo(const DadInfo &aDadInfo);
+
+    /**
+     * This method reads duplicate address detection information.
+     *
+     * @param[out]   aDadInfo         A reference to a `DadInfo` structure to output the read content.
+     *
+     * @retval OT_ERROR_NONE              Successfully read the duplicate address detection information.
+     * @retval OT_ERROR_NOT_FOUND         No corresponding value in the setting store.
```
jwhui, line 927: Is the `NOT_FOUND` error useful?

src/core/thread/dua_manager.hpp
```diff
@@ -124,8 +124,18 @@ class DuaManager : public InstanceLocator
      */
     const Ip6::InterfaceIdentifier &GetFixedDuaInterfaceIdentifier(void) const { return mFixedDuaInterfaceIdentifier; }
 
+    /*
+     * This method restores duplicate address detection information from non-volatile memory.
+     *
+     * @retval OT_ERROR_NONE       Successfully restored the duplicatee address detection information.
+     * @retval OT_ERROR_NOT_FOUND  There is no valid Dad Counter stored in non-volatile memory.
```
jwhui, line 131: Is returning an error useful?

src/core/common/settings.hpp
librasungirl: `NOT_FOUND` is generally used to indicate no corresponding key storage in existing `Read` methods and  the underlying SettingDriver's `Get` method in `settings` module.
Better to keep it?

src/core/thread/dua_manager.hpp
librasungirl: could be removed. will update

PR #4909, opened 2020-04-29
[radio] use otDataset APIs instead of otPlatSettings APIs

wgtdkp:
This PR replaces use of `otPlatSettingsSet()` in `radio_spinel_impl.hpp` to equivalent `otDatasetSetActive()` and `otDatasetSetPending()`, because `otPlatSettings` APIs are not mandatory for all platforms while `otDataset` APIs are provided by openthread itself.


wgtdkp:
Closed this PR because using `otPlatSettings` API is essential and `otDataset` API seems doesn't work.

PR #4911, opened 2020-04-29
[continuous-integration] recover verbose check

bukepo:
The error logs of `check` will only be printed when VERBOSE=1.

### DIFF
--- a/.github/workflows/simulation.yml
+++ b/.github/workflows/simulation.yml
@@ -84,7 +84,7 @@ jobs:
         make -f examples/Makefile-simulation
     - name: Run
       run: |
-        make -f examples/Makefile-simulation check
+        VERBOSE=1 make -f examples/Makefile-simulation check
     - name: Codecov
       uses: codecov/codecov-action@v1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/simulation.yml
```diff
@@ -236,7 +234,7 @@ jobs:
         make -f src/posix/Makefile-posix
     - name: Run
       run: |
-        OT_NCP_PATH="$(pwd)/$(ls output/posix/*/bin/ot-ncp)" RADIO_DEVICE="$(pwd)/$(ls output/*/bin/ot-rcp)" make -f src/posix/Makefile-posix check
+        VERBOSE=1 OT_NCP_PATH="$(pwd)/$(ls output/posix/*/bin/ot-ncp)" RADIO_DEVICE="$(pwd)/$(ls output/*/bin/ot-rcp)" make -f src/posix/Makefile-posix check
```
simonlingoogle, line 237: `VERBOSE` in env not working?

bukepo: It is only used by the `make check`, so I prefer adding it here.

PR #4913, opened 2020-04-29
[continuous-integration] add POSIX SPI build

Moandor-y:
This adds a CI build for POSIX platform with SPI enabled.

Also fixed a build error in `spi_interface.cpp`.


src/posix/platform/CMakeLists.txt
```diff
@@ -38,15 +38,15 @@ if(OT_POSIX_VIRTUAL_TIME)
     list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_POSIX_VIRTUAL_TIME=1")
 endif()
 
+set(OT_POSIX_CONFIG_RCP_BUS "OT_POSIX_RCP_BUS_UART" CACHE STRING "RCP bus type")
+list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_POSIX_CONFIG_RCP_BUS=${OT_POSIX_CONFIG_RCP_BUS}")
```
bukepo, line 42:
```suggestion
set(OT_POSIX_CONFIG_RCP_BUS "UART" CACHE STRING "RCP bus type")
list(APPEND OT_PLATFORM_DEFINES "OPENTHREAD_POSIX_CONFIG_RCP_BUS=OT_POSIX_CONFIG_RCP_${OT_POSIX_CONFIG_RCP_BUS}")
```

.github/workflows/build.yml
```diff
@@ -159,6 +159,19 @@ jobs:
...
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Bootstrap
+      run: |
+        sudo apt-get update
+        sudo apt-get --no-install-recommends install -y cmake ninja-build
+    - name: Build
+      run: |
+        script/check-posix-spi-build
```
bukepo, line 173:
```suggestion
        mkdir build && cd build && cmake -GNinja -DOT_COMPILE_WARNING_AS_ERROR=on-DOT_PLATFORM=posix && ninja
```

.github/workflows/build.yml
```diff
@@ -159,6 +159,19 @@ jobs:
...
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Bootstrap
+      run: |
+        sudo apt-get update
+        sudo apt-get --no-install-recommends install -y cmake ninja-build
+    - name: Build
+      run: |
+        mkdir build && cd build && cmake -GNinja -DOT_COMPILE_WARNING_AS_ERROR=on -DOT_PLATFORM=posix -DOT_POSIX_CONFIG_RCP_BUS=SPI .. && ninja
```
jwhui, line 173: Can we create `script/check-posix-build-cmake` and use that here?

Moandor-y: Updated. Also fixed another error ([scan format type mismatch](https://github.com/openthread/openthread/pull/4913/files#diff-19d73e39e7f769ebb868e2472105e02fR438)) found by CMake build.

Moandor-y:
> We should also fix/update:
> https://github.com/openthread/openthread/blob/81fb1210c46b19b0c3213d32e6b64c2fa817f0cd/script/check-posix-build-autotools#L45

Building with SPI on macOS will fail so I disabled the SPI build for macOS.

src/posix/platform/CMakeLists.txt
```diff
@@ -38,15 +38,15 @@ if(OT_POSIX_VIRTUAL_TIME)
     list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_POSIX_VIRTUAL_TIME=1")
 endif()
 
+set(OT_POSIX_CONFIG_RCP_BUS "UART" CACHE STRING "RCP bus type")
```
bukepo, line 41: according to suggestion from @abtink, we should make default value `""`, and only set when the variable is set `""`.

src/posix/platform/CMakeLists.txt
```diff
@@ -38,15 +38,15 @@ if(OT_POSIX_VIRTUAL_TIME)
...
+list(APPEND OT_PLATFORM_DEFINES "OPENTHREAD_POSIX_CONFIG_RCP_BUS=OT_POSIX_RCP_BUS_${OT_POSIX_CONFIG_RCP_BUS}")
```
bukepo, line 42: according to suggestion from @abtink, we should make default value `""`, and only set when the variable is set `""`.
```suggestion
set(OT_POSIX_CONFIG_RCP_BUS "" CACHE STRING "RCP bus type")
if(OT_POSIX_CONFIG_RCP_BUS)
    list(APPEND OT_PLATFORM_DEFINES "OPENTHREAD_POSIX_CONFIG_RCP_BUS=OT_POSIX_RCP_BUS_${OT_POSIX_CONFIG_RCP_BUS}")
endif()
```

PR #4916, opened 2020-04-29
[test] fix script/test on macOS

simonlingoogle:
`./script/test build` does not work on macOS without this PR.

### DIFF
--- a/script/test
+++ b/script/test
@@ -378,7 +378,7 @@ envsetup()
     fi
 
     read -r -a ot_extra_options <<< "${OT_OPTIONS-}"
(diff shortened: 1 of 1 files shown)


script/test
```diff
@@ -378,6 +378,7 @@ envsetup()
     fi
 
     read -r -a ot_extra_options <<< "${OT_OPTIONS-}"
+    ot_extra_options=${ot_extra_options:-}
```
bukepo, line 381: Is this still an array?

simonlingoogle: revised, please check

PR #3262, opened 2018-11-02
[mac] adding sub-MAC layer

abtink:
This commit introduces sub-MAC layer (class `Mac::SubMac`) in OpenThread implementing a subset of IEEE802.15.4 MAC primitives which are shared by both MAC layer (in FTD/MTD modes) and Raw Link (Radio only mode).

The sub-MAC layer handles the following (if not provided by radio platform):

  - Ack timeout for frame transmission,
  - CSMA backoff logic,
  - Frame re-transmissions,
  - Energy scan on a single channel and RSSI sampling.

It also act as the interface (to radio platform) for setting/getting radio configurations such as short or extended addresses or PAN Id.

In addition, this commit contains following improvements in `Mac`:
- The implementation of `STAY_AWAKE_BETWEEN_FRAGMENTS` (delay sleep)   feature is simplified by adding a new state variable tracking if  sleep is being delayed and canceling sleep delay on start of any operation.
- Member variables in `Mac` class are re-ordered to avoid alignment gaps.
- `Mac` timers are combined into one which is used for Active Scan,  Data Poll timeout, and delaying sleep.

PR #1248, opened 2017-02-02
Optional otLinkRaw Features

nibanks:
When a Host uses the Spinel interface to control the `otLinkRaw*` interface it practically requires the device to support of all the `otRadioCaps`. Many of these features are not supported in hardware by our current example platforms. This change adds support for each feature in software, controllable by compile-time flags. All of this is still gated on the compile time flag that controls the `otLinkRaw*` API.

In summary, with this change, all example platforms (excluding posix) should fully support a Host using the `otLinkRaw*` APIs via Spinel.

PR #4919, opened 2020-05-01
[mac] put transmission AES-CCM* to SubMac layer

brianljt:
In Thread 1.2, it is possible to receive an IEEE 802.15.4-2015 packet. Per specification, receiver should acknowledge this packet with an IEEE 802.15.4-2015 ACK(Enh-ACK). This Enh-ACK can include header IE with it and requires `security enabled` bit in `FCF` be set to true. It is impractical for the host to generate the Enh-ACK and send to RCP for transmission within `AIFS` time(192us). So RCP should prepare the Enh-ACK by itself, which requires it to fill in the frame counter and do the encryption/authentication. This PR tries to address the need of transmission security/authentication by including the following changes,

1. Move Key ID mode 1 AES-CCM* related functions from MAC layer to SubMac layer, which is mirrored in RCP.
2. Distribute the MAC key and MAC key ID to RCP in posix app using newly added spinel properties.
3. Make it possible for radio(either in radio driver or hardware) to do transmission AES-CCM* if the platform supports by adding radio capability `OT_RADIO_CAPS_TRANSMIT_SEC`.
4. Enable this for RCP mode on simulation platform.


src/ncp/ncp_base_dispatcher.cpp
```diff
@@ -411,6 +411,12 @@ NcpBase::PropertyHandler NcpBase::FindSetPropertyHandler(spinel_prop_key_t aKey)
 #if OPENTHREAD_CONFIG_MAC_RETRY_SUCCESS_HISTOGRAM_ENABLE
         OT_NCP_SET_HANDLER_ENTRY(SPINEL_PROP_CNTR_MAC_RETRY_HISTOGRAM),
 #endif
+#endif // OPENTHREAD_MTD || OPENTHREAD_FTD
+#if OPENTHREAD_RADIO || OPENTHREAD_CONFIG_LINK_RAW_ENABLE
+        OT_NCP_SET_HANDLER_ENTRY(SPINEL_PROP_RCP_MAC_KEY),
+        OT_NCP_SET_HANDLER_ENTRY(SPINEL_PROP_RCP_MAC_KEY_ID),
+#endif
```
Irving-cl, line 418: What about putting these lines after line 428? Thus we can separate the `OPENTHREAD_FTD || OPENTHREAD_MTD` part from the `OPENTHREAD_RADIO` part.

src/ncp/ncp_base_radio.cpp
```diff
@@ -421,6 +429,50 @@ otError NcpBase::HandlePropertySet_SPINEL_PROP_STREAM_RAW(uint8_t aHeader)
...
+{
+    otError        error = OT_ERROR_NONE;
+    uint8_t        keyIdMode;
+    uint16_t       keySize;
+    const uint8_t *prevKey;
+    const uint8_t *currKey;
+    const uint8_t *nextKey;
+
+    SuccessOrExit(error = mDecoder.ReadUint8(keyIdMode));
+    VerifyOrExit(keyIdMode == Mac::Frame::kKeyIdMode1, error = OT_ERROR_INVALID_ARGS);
```
Irving-cl, line 442: Since `keyIdMode` could only be `1`, shall we remove the parameter from the functions and always let `keyIdMode` to be `1` on RCP? Or it is possible to extend further in the future?

include/openthread/platform/radio.h
```diff
@@ -433,6 +435,14 @@ bool otPlatRadioGetPromiscuous(otInstance *aInstance);
  */
 void otPlatRadioSetPromiscuous(otInstance *aInstance, bool aEnable);
 
+void otPlatRadioSetMacKey(otInstance *   aInstance,
+                          uint8_t        aKeyMode,
+                          uint8_t        aKeySize,
+                          const uint8_t *aPrevKey,
+                          const uint8_t *aCurrKey,
+                          const uint8_t *aNextKey);
+void otPlatRadioSetMacKeyId(otInstance *aInstance, uint8_t aKeyMode, uint8_t aKeyId);
```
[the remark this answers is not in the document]
brianljt: Sure.

src/ncp/ncp_base_dispatcher.cpp
brianljt: I think there is a check to see if these properties are in order with their value. Moving these around will throw an expect error.

src/ncp/ncp_base_radio.cpp
brianljt: I think it is better to have the ability to extend this property in the future although only one type of key ID mode is used right now.

include/openthread/platform/radio.h
```diff
@@ -433,6 +435,34 @@ bool otPlatRadioGetPromiscuous(otInstance *aInstance);
  */
 void otPlatRadioSetPromiscuous(otInstance *aInstance, bool aEnable);
 
+/**
+ * Update MAC keys.
+ *
```
abtink, line 440: How about adding a comment here like:
```
  * This function is used when radio provides OT_RADIO_CAPS_TRANSMIT_SEC capability.
```

include/openthread/platform/radio.h
```diff
@@ -433,6 +435,34 @@ bool otPlatRadioGetPromiscuous(otInstance *aInstance);
...
+void otPlatRadioSetMacKey(otInstance *   aInstance,
+                          uint8_t        aKeyMode,
+                          uint8_t        aKeySize,
+                          const uint8_t *aPrevKey,
+                          const uint8_t *aCurrKey,
+                          const uint8_t *aNextKey);
+
+/**
+ * Update MAC key index.
+ *
```
abtink, line 458: Similar comment about `CAPS_TRANSMIT_SEC`?

src/core/mac/mac.hpp
```diff
@@ -720,15 +720,12 @@ class Mac : public InstanceLocator
...
-     * (depends on @p aProcessAesCcm). If aProcessAesCcm is False, it probably means that some content in the frame
-     * will be updated just before transmission, so AES CCM will be performed after that (before transmission).
+     * This method prepares the frame and fills Mac auxiliary header.
      *
      * @param[in]  aFrame          A reference to the MAC frame buffer which is going to be sent.
-     * @param[in]  aProcessAesCcm  TRUE to perform AES CCM immediately, FALSE otherwise.
      *
      */
-    void ProcessTransmitSecurity(TxFrame &aFrame, bool aProcessAesCcm);
+    void ProcessTransmitSecurity(TxFrame &aFrame);
```
abtink, line 728: Can we consider removing the doxygen description for this method since it is a `private` method?

src/core/mac/sub_mac.cpp
```diff
@@ -565,6 +644,59 @@ void SubMac::SetState(State aState)
...
+        memcpy(mNextKey, aNextKey, kMacKeySize);
+
+        break;
+
+    default:
+        OT_ASSERT(false);
+        break;
+    }
+
+    VerifyOrExit(!ShouldHandleTransmitSecurity(), OT_NOOP);
```
Irving-cl, line 668: Just a question, not a suggestion:
If we run a RCP instance and try to set Mac Key, this method would be called twice: first time on host and this function wouldn't stop at this line, the second time on RCP and this function would stop here, does it work in this way?

src/core/mac/sub_mac.hpp
```diff
@@ -347,6 +348,50 @@ class SubMac : public InstanceLocator
      */
     int8_t GetNoiseFloor(void);
 
+    /**
+     * This method sets previous, current and next MAC key.
```
abtink, line 352:
```suggestion
     * This method sets the previous, current, and next MAC key.
```

src/core/mac/sub_mac.cpp
brianljt: Yes. It is intended this way. The radio function in line 670 is intended to set MAC key to Radio. In RCP case, on host side, the radio is RCP. So in posix app, the radio_spinel will be the radio layer and this key will be passed to RCP. In RCP side, the radio layer is the real platform, if the platform driver or hardware supports Tx encryption, this key will be passed to it, if not, Submac will handle the Tx security and the key will not be passed down.

For NCP mode, this line is used to decide if submac need to pass the key the real radio driver.

Irving-cl: Got it, thanks!

src/core/mac/sub_mac.cpp
```diff
@@ -565,6 +644,59 @@ void SubMac::SetState(State aState)
...
+
+    Get<Radio>().SetMacKey(aKeyIdMode, kMacKeySize, aPrevKey, aCurrKey, aNextKey);
+
+exit:
+    return;
+}
+
+void SubMac::SetMacKeyId(uint8_t aKeyIdMode, uint8_t aKeyId)
```
abtink, line 676: I noticed in all these methods, we pass in `aKeyIdMode` but I think we practically only call it with `kKeyMode1` and only in this mode the code does something (saves the keys, etc). Do we think in future we may use other key id modes? Can we consider removing the `aKeyIdMode` parameter?

src/ncp/ncp_base_dispatcher.cpp
abtink: Yes, this list needs be sorted since binary search is used to find the matching handler for a property key.

src/core/mac/sub_mac.cpp
brianljt: I leave the key ID mode in the function parameter for future expansion/code reorg. I think if we keep key ID mode 0 and 2 handling in SubMac, too, we can remove the function `Mac::ProcessTransmitSecurity` from `Mac` layer. This will keep a cleaner code architecture so that the security related code does not scatter around `Mac` and `SubMac`. How do you think?

abtink: SGTM.

include/openthread/platform/radio.h
```diff
@@ -433,6 +435,36 @@ bool otPlatRadioGetPromiscuous(otInstance *aInstance);
  */
 void otPlatRadioSetPromiscuous(otInstance *aInstance, bool aEnable);
 
+/**
+ * Update MAC keys.
```
bukepo, line 439:
```suggestion
 * Update MAC keys.
 *
```

include/openthread/platform/radio.h
```diff
@@ -433,6 +435,36 @@ bool otPlatRadioGetPromiscuous(otInstance *aInstance);
...
+ */
+void otPlatRadioSetMacKey(otInstance *   aInstance,
+                          uint8_t        aKeyMode,
+                          uint8_t        aKeySize,
+                          const uint8_t *aPrevKey,
+                          const uint8_t *aCurrKey,
+                          const uint8_t *aNextKey);
+
+/**
+ * Update MAC key index.
```
bukepo, line 458:
```suggestion
 * Update MAC key index.
 *
```

src/core/config/link_raw.h
```diff
@@ -78,7 +78,17 @@
  *
  */
 #ifndef OPENTHREAD_CONFIG_SOFTWARE_CSMA_BACKOFF_ENABLE
-#define OPENTHREAD_CONFIG_SOFTWARE_CSMA_BACKOFF_ENABLE 0
+#define OPENTHREAD_CONFIG_SOFTWARE_CSMA_BACKOFF_ENABLE 1
```
bukepo, line 81: is it possible to keep the original value?

include/openthread/platform/radio.h
```diff
@@ -196,6 +197,7 @@ typedef struct otRadioFrame
             uint8_t        mMaxFrameRetries;   ///< Maximum number of retries allowed after a transmission failure.
             bool           mIsARetx : 1;       ///< True if this frame is a retransmission (ignored by radio driver).
             bool           mCsmaCaEnabled : 1; ///< Set to true to enable CSMA-CA for this packet, false otherwise.
+            bool           mIsOob : 1;         ///< True if this frame is an out of band frame.
```
bukepo, line 200: Are all out of band frames of key id mode 1? I hope we can avoid introducing this flag into spinel protocol.

src/core/config/link_raw.h
brianljt: This was mistakenly changed. Will change it back.

include/openthread/platform/radio.h
brianljt: Per description of function `otLinkOutOfBandTransmitRequest`,

> An Out of Band frame is one that was generated outside of OpenThread

So OT should not modify or make assumption of the content of the frame. Which means it is possible that the frame is using key ID mode 1 and OT stack should not encrypt the content again.

bukepo: As we were considering removing this feature, so if these frames are only of key id mode 1, we should avoid introducing this flag. Besides, we should also add a note the existing API that this feature maybe removed. Thoughts? @abtink @jwhui

bukepo: Agreed to use a more neutral flag name.

abtink: @bukepo I agree and would also definitely favor removing the OOB feature later on.
A more neutral flag name sounds good to me.

third_party/mbedtls/Makefile.am
```diff
@@ -21,7 +21,10 @@ EXTRA_DIST                                    = \
     repo/include                                \
     $(NULL)
 
-lib_LIBRARIES                                 = libmbedcrypto.a
+lib_LIBRARIES                                 = \
+    libmbedcrypto.a                             \
+    libmbedcrypto_radio.a                       \
+	$(NULL)
```
jwhui, line 27:
```suggestion
    $(NULL)
```

src/core/config/link_raw.h
```diff
@@ -81,6 +81,16 @@
...
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_SOFTWARE_TX_SECURITY_ENABLE
+ *
+ * Define to 1 if you want to enable software transmission security logic.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_SOFTWARE_TX_SECURITY_ENABLE
+#define OPENTHREAD_CONFIG_SOFTWARE_TX_SECURITY_ENABLE 0
```
jwhui, line 91: Should this belong in `src/core/config/mac.h`? Also, should this name include `MAC` or `SUB_MAC`?

brianljt: This follows the similar behavior like `OPENTHREAD_CONFIG_SOFTWARE_CSMA_BACKOFF_ENABLE` and `OPENTHREAD_CONFIG_SOFTWARE_RETRANSMIT_ENABLE`. This flag is to indicate whether Tx security is done at OT(no matter which layer) or at platform(radio driver or hardware). If platform provide capability `OT_RADIO_CAPS_TRANSMIT_SEC`, then OT does not do so called software CSMA/retransmission/Tx security and will leave it to platform(considered as hardware CSMA/Retx/Tx Sec). If not, OT will do this.

This flag is configured per platform, so it resides in platform's openthread-core-xxx-config.h.

jwhui: I think the current pattern is out of date. These options were created in #1248 and specific to LINK_RAW feature at the time. Now these features have been combined into the sub-MAC in #3262 and included as part of RCP, independent of whether LINK_RAW features is enabled. As a result, I think these options are likely better scoped under the MAC component.

brianljt: I see. I will move all these flags to `src/core/config/mac.h`.

src/core/radio/radio.hpp
```diff
@@ -247,6 +247,34 @@ class Radio : public InstanceLocator
      */
     void SetShortAddress(Mac::ShortAddress aShortAddress);
 
+    /**
+     * This method sets previous, current and next MAC key.
+     *
+     * @param[in] aKeyIdMode  MAC key ID mode.
+     * @param[in] aKeySize    MAC key length.
```
librasungirl, line 254:
```suggestion
     * @param[in] aKeySize    MAC key size in bytes
```

src/lib/spinel/spinel.h
```diff
@@ -3909,6 +3914,34 @@ enum
...
+
+    /// MAC Key
+    /** Format: `Cddd`.
+     *
+     *  `C`: MAC key ID mode
+     *  `d`: previous MAC key material data
+     *  `d`: current MAC key material data
+     *  `d`: next MAC key material data
+     *
+     * The Spinel property is used to set/get MAC key materials to and from RCP.
```
librasungirl, line 3927: Just curious - 
1) there is only set handler in the implementation, will this be a read/write property or a write-only property?

2) I see there are some benefits of splitting KeyId from the Keys, but if `SPINEL_PROP_RCP_MAC_KEY` and `SPINEL_PROP_RCP_MAC_KEY_ID` are write-only, maybe combining key and key id may reduce the amount of APIs?

src/core/mac/sub_mac.cpp
```diff
@@ -222,6 +231,58 @@ otError SubMac::Send(void)
...
+    VerifyOrExit(ShouldHandleTransmitSecurity(), OT_NOOP);
+    VerifyOrExit(mTransmitFrame.GetSecurityEnabled(), OT_NOOP);
+    VerifyOrExit(!mTransmitFrame.ShouldSkipAes(), OT_NOOP);
+
+#if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
+    if (mTransmitFrame.GetTimeIeOffset() != 0)
+    {
+        // Transmit security will be processed after time IE content is updated.
+        processTransmitAesCcm = false;
+    }
```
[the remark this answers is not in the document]
brianljt: For TimerIE included frame, we still need to set the KeyId, frame counter, MAC key here. Just skip the AES encryption and authentication part. So it is intended not to skip the rest parts before `exit`.

src/lib/spinel/spinel.h
brianljt: 1. Currently it is a write only property. Host does not have the need to get MAC key from RCP for now. We can implement it later if needed. I would prefer to keep it as write only as MAC keys are sensitive information.

2. This is out of naming and future expansion consideration. What if we need to read this property from RCP later or need to change only Key ID.

librasungirl: MAC keys are derived from masterkey and sequence number, so if sequence number (key id) changes, MAC keys should always change as well, right?

brianljt: That is true. I was also considering expanding this property to be used for Key ID mode 0 and 2. For mode 0 case, the Key ID is not there.

brianljt: On a second thought, combining MAC key and MAC key ID into one property might help address the racing condition when MAC key is updated and Key ID is not, an Enh-ACK is transmitted in between with wrong key used. I will change the implementation.

src/core/mac/mac_frame.cpp
```diff
@@ -1045,7 +1043,7 @@ void TxFrame::ProcessTransmitAesCcm(const ExtAddress &aExtAddress)
 
 exit:
     return;
-#endif // OPENTHREAD_RADIO
+#endif // OPENTHREAD_RADIO && !OPENTHREAD_CONFIG_MAC_SOFTWARE_TX_SECURITY_ENABLE
```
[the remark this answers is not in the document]
brianljt: This is paired with `#if OPENTHREAD_RADIO && !OPENTHREAD_CONFIG_MAC_SOFTWARE_TX_SECURITY_ENABLE`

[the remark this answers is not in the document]
librasungirl: wrong match, removed.

include/openthread/platform/radio.h
```diff
@@ -196,6 +197,7 @@ typedef struct otRadioFrame
             uint8_t        mMaxFrameRetries;   ///< Maximum number of retries allowed after a transmission failure.
             bool           mIsARetx : 1;       ///< True if this frame is a retransmission (ignored by radio driver).
             bool           mCsmaCaEnabled : 1; ///< Set to true to enable CSMA-CA for this packet, false otherwise.
+            bool           mSkipAes : 1;       ///< True if SubMac should skip the AES processing of this frame.
```
bukepo, line 200:
```suggestion
            bool           mIsSecurityProcessed : 1;       ///< True if the frame already processed security.
```

src/core/mac/mac_frame.cpp
```diff
@@ -1063,7 +1061,7 @@ void TxFrame::ProcessTransmitAesCcm(const ExtAddress &aExtAddress)
 
 exit:
     return;
```
bukepo, line 1063: mark the frame security processed by setting mIsSecurityProcessed?

src/core/mac/sub_mac.cpp
```diff
@@ -222,6 +231,58 @@ otError SubMac::Send(void)
...
+#endif
+
+    IgnoreError(mTransmitFrame.GetKeyIdMode(keyIdMode));
```
bukepo, line 252: Use assert here because this is the frame is generated by us?

src/core/mac/sub_mac.cpp
```diff
@@ -222,6 +231,58 @@ otError SubMac::Send(void)
...
+
+    switch (keyIdMode)
+    {
+    case Frame::kKeyIdMode0:
+    case Frame::kKeyIdMode2:
+        // For key Id mode 0 and 2, the security is handled by Mac
+        processTransmitAesCcm = false;
```
bukepo, line 259: if we mark the frame is processed already, we don't need this block anymore.

src/core/mac/mac.cpp
```diff
@@ -982,9 +989,10 @@ void Mac::ProcessTransmitSecurity(TxFrame &aFrame, bool aProcessAesCcm)
         {
             aFrame.SetFrameCounter(keyManager.GetMacFrameCounter());
             keyManager.IncrementMacFrameCounter();
-            aFrame.SetKeyId((keyManager.GetCurrentKeySequence() & 0x7f) + 1);
         }
 
+        // For MAC key ID mode 1, the AES CCM* is done at SubMac or Radio if supported
+        processTransmitAesCcm = false;
```
bukepo, line 995: Verify the radio supports transmit encryption?

brianljt: No need. It will be passed down to `SubMac` and check radio capability there.

src/core/mac/sub_mac.cpp
```diff
@@ -222,6 +231,44 @@ otError SubMac::Send(void)
     return error;
 }
 
+void SubMac::ProcessTransmitSecurity(void)
+{
+    const ExtAddress *extAddress            = NULL;
+    bool              processTransmitAesCcm = true;
```
bukepo, line 237:
```suggestion
```

src/core/mac/sub_mac.cpp
```diff
@@ -222,6 +231,44 @@ otError SubMac::Send(void)
...
+    VerifyOrExit(mTransmitFrame.GetSecurityEnabled(), OT_NOOP);
+    VerifyOrExit(!mTransmitFrame.IsSecurityProcessed(), OT_NOOP);
+
+#if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
+    if (mTransmitFrame.GetTimeIeOffset() != 0)
+    {
+        // Transmit security will be processed after time IE content is updated.
+        processTransmitAesCcm = false;
+    }
+#endif
```
bukepo, line 250:
```suggestion
#if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
VerifyOrExit(mTransmitFrame.GetTimeIeOffset() == 0);
#endif
```
And move to the line just before `mTransmitFrame.ProcessTransmitAesCcm(*extAddress);`

src/core/mac/mac.cpp
bukepo: ExitNow();

src/core/mac/mac.cpp
```diff
@@ -944,14 +945,23 @@ bool Mac::IsJoinable(void) const
...
 
     VerifyOrExit(aFrame.GetSecurityEnabled(), OT_NOOP);
 
+#if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
+    if (aFrame.GetTimeIeOffset() != 0)
+    {
+        // Transmit security will be processed after time IE content is updated.
+        processTransmitAesCcm = false;
+    }
+#endif
```
bukepo, line 963:
```suggestion
#if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
VerifyOrExit(aFrame.GetTimeIeOffset() == 0);
#endif
```
And move to just before: `aFrame.ProcessTransmitAesCcm(*extAddress);`

src/core/mac/mac.cpp
```diff
@@ -944,14 +945,23 @@ bool Mac::IsJoinable(void) const
...
 }
 
-void Mac::ProcessTransmitSecurity(TxFrame &aFrame, bool aProcessAesCcm)
+void Mac::ProcessTransmitSecurity(TxFrame &aFrame)
 {
     KeyManager &      keyManager = Get<KeyManager>();
     uint8_t           keyIdMode;
-    const ExtAddress *extAddress = NULL;
+    const ExtAddress *extAddress            = NULL;
+    bool              processTransmitAesCcm = true;
```
bukepo, line 953:
```suggestion
```

PR #4923, opened 2020-05-03
[build] add -Wcast-align to detect potential alignment issues when casting

dismirlian:
Hi all, this PR enables the -Wcast-align warning for the OT core, and for some platforms (very few that don't issue warnings).

For background, see #4796.


jwhui:
@dismirlian, thanks!

Issue #4706, opened 2020-03-19
[efr32] Settings


dismirlian:
Recently, a rework of the settings storage was merged (see PR #4552), which retains format compatibility with the original implementation. This settings format (both the original and new implementations) has an issue with the efr32 platform:

  The MG12 datasheet states:

  > It is possible to write words twice between each erase by keeping at 1 the bits that are not to be changed. [...] Note that there is a maximum of two writes to the same word between each erase due to a physical limitation of the flash.

  The MG21 datasheet says something similar:
  > The Flash memory is organized into 64-bit wide double-words. Each 64-bit double-word can be written only twice between erase cycles

  The current implementation writes the `RecordHeader` twice during `Add` and once more during `Delete`.

Additionally, PR #4521 provides an alternate implementation of the settings area, specifically for the efr32 platforms. This will not retain compatibility with the old format.

I wanted to "expose" a potential issue, for those with devices on the field:
- Either they update to #4521, and break format compatibility or
- They stay with the original format which violates the datasheet recommendations. I don't know what implications this may have. Maybe increased wear or decreased reliability/retention? Perhaps someone from Silabs can shed some light on this.

In our case, we are not impacted by this potential problem, because we don't have devices on the field. I am opening this issue to see to what extent this is an actual issue or not. If it's not, then we can close it.

Thanks!

yuping-xiao:
Diego,

You are correct that the maximum number of writes on a flash word in between erase cycles is 2 on EFR32MG21. It is a statistically safe recommendation according to our testing. Writing a 3rd time may not be a problem immediately but it’s risky and violates our recommendation, in which case we can no longer guarantee that the flash always perform to its datasheet specifications such as write/erase cycle endurance and data retention. 
 
Two questions about PR #4552:
1.	I can see the two writes in Add and one write in Delete but it’s not obvious to me that they are writing to the exact same flash location. Can you confirm/elaborate? This question is due to my lack of familiarity with the code.
2.	Permitting a small number of writes between erase cycles is a typical limitation on flash, not an exception, although the exact number of such guaranteed writes may vary. Assuming answer to question 1 is positive, then I’m curious to know which chip families this implementation is intended for, and if the respective datasheets have been consulted with to confirm it’s a sound implementation.

I should add that we ask those using EFR32 to adopt Silicon Labs NVM3 solution instead of the OT default NV manager. One thing to note is the token values cannot be expected to be preserved during firmware upgrade if switching from OT default NV manager to NVM3.

[...]

PR #4926, opened 2020-05-04
[flash] avoid more than two writes to the same 32-bit word

dismirlian:
Hi all, this PR updates the Flash driver and associated unit tests to avoid the issues explained in #4706.

I think it would be best to review this as separate commits. The changes are:
  - 9e7e920: Update the flash driver to avoid marking with `kFlagFirst` the record with `index == 1`. This is the most "critical" commit, because it allows the use of the `mReserved` half-word for the `kFlagFirst` and `kFlagDelete` flags. Please review it carefully; it slightly changes the underlying logic for calculating index 0.
  - cdf7a2a: Update the unit test to check the maximum number of 32-bit words written between erase cycles. This is fixed to 2. See #4706.
  - d6a1a4e: Implements the new flash format, optionally keeping backward compatibility. The new format uses the `mReserved` half-word to write the `kFlagFirst` and `kFlagDeleted`, but (optionally) understands the old format. The backward compatibility is enabled by default.
  - fa6ebef: Updates the unit test to check compatibility: writing and reading with the old driver, writing and reading with the new driver, and writing with the old driver and reading with the new one.

Comments are welcome...

Thanks.

### DIFF
--- a/src/core/Makefile.am
+++ b/src/core/Makefile.am
@@ -352,6 +352,7 @@ HEADERS_COMMON                             = \
     config/diag.h                            \
     config/dns_client.h                      \
     config/ip6.h                             \
+    config/flash.h                           \
     config/joiner.h                          \
     config/link_quality.h                    \
     config/link_raw.h                        \
--- a/src/core/config/flash.h
+++ b/src/core/config/flash.h
@@ -1,0 +1,71 @@
+/*
+ *  Copyright (c) 2020, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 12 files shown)


jwhui:
@dismirlian , thank you for this contribution! Were you able to get an answer to https://github.com/openthread/openthread/issues/4706#issuecomment-612017804 from Silicon Labs to ensure that considering 32-bit words is sufficient?

dismirlian:
Hi @jwhui, no, I haven't had any response. What do you suggest?

jwhui:
> Hi @jwhui, no, I haven't had any response. What do you suggest?

@dismirlian, let me try to get an answer from Silicon Labs.

PR #4927, opened 2020-05-06
[tests] fix key_id_mode handling and add exception handle

Irving-cl:
Sorry for my mistake in #3448 . This PR fixes `key_id_mode` handling and added exception handling when `key_id_mode` is `0`.

According to **IEEE Std 802.15.4-2015**, when Key Id Mode is `0x00`:

> Key is determined implicitly from the originator and recipient(s) of the frame, as indicated in the frame header.

So in our test scripts we cannot actually handle such packets. In case 8_1_01, as I remember, @wgtdkp checked such packets(`HandshakeType.SERVER_KEY_EXCHANGE` and `HandshakeType.CLIENT_KEY_EXCHANGE`) by other means. When we parse such packet we can simply drop them.

However if we leave `key_id_mode = 0` case unhandled, we would get some exceptions which makes people feel that something goes wrong. However dropping packets in this case is a normal behavior. So I added the handling to such exceptions.

### DIFF
--- a/tests/scripts/thread-cert/simulator.py
+++ b/tests/scripts/thread-cert/simulator.py
@@ -41,6 +41,8 @@ import mesh_cop
 import message
 import pcap
 
+from message import DropPacketException
+
 
 def dbg_print(*args):
     if False:
@@ -183,6 +185,10 @@ class VirtualTime(BaseSimulator):
             messages = self._message_factory.create(io.BytesIO(message))
             self.devices[addr]['msgs'] += messages
 
+        except DropPacketException:
+            print(
+                "Drop current packet because it cannot be handled in test scripts"
+            )
         except Exception as e:
             # Just print the exception to the console
             print("EXCEPTION: %s" % e)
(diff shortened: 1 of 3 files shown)


tests/scripts/thread-cert/simulator.py
```diff
@@ -41,6 +41,8 @@
 import message
 import pcap
 
+from message import DropPacketException
+
```
bukepo, line 45:
```suggestion
```

tests/scripts/thread-cert/simulator.py
```diff
@@ -175,14 +175,18 @@ def stop(self):
...
 
         # Ignore any exceptions
         try:
-            messages = self._message_factory.create(io.BytesIO(message))
+            messages = self._message_factory.create(io.BytesIO(message_obj))
             self.devices[addr]['msgs'] += messages
 
+        except message.DropPacketException:
+            print(
+                "Drop current packet because it cannot be handled in test scripts"
```
jwhui, line 188: nit: use `'` instead of `"`?

PR #4929, opened 2020-05-06
[cmake] add project version

bukepo:
* use an older version for better compatibility
* add version to project
* only use cpack when OpenThread is the top-level project
* exclude posix apps when BUILD_EXECUTABLES is OFF
* exclude mbedtls to prevent installing headers


third_party/mbedtls/CMakeLists.txt
```diff
@@ -33,7 +33,7 @@ set(OT_MBEDTLS_CONFIG_FILE "" CACHE STRING "The mbedTLS config file")
 set(ENABLE_TESTING OFF CACHE BOOL "Disable mbedtls test" FORCE)
 set(ENABLE_PROGRAMS OFF CACHE BOOL "Disable mbetls program" FORCE)
 
-add_subdirectory(repo)
+add_subdirectory(repo EXCLUDE_FROM_ALL)
```
wgtdkp, line 36: better use `EXCLUDE_FROM_ALL` on the whole `third_party` directory.

CMakeLists.txt
```diff
@@ -26,8 +26,13 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
-cmake_minimum_required(VERSION 3.13.1)
-project(openthread)
+cmake_policy(SET CMP0048 NEW)
+cmake_minimum_required(VERSION 3.11.4)
```
wgtdkp, line 30: How is the minimum version determined? Are there dependent features on version 3.11.4?

bukepo: It is the cmake in openwrt 18.06, the version we support in ot-br-posix.

third_party/mbedtls/CMakeLists.txt
bukepo: agree.

PR #4931, opened 2020-05-06
[error] add missing IgnoreError

abtink:
---------

Follow up on #4784 adding missing `IgnoreError` in different modules.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -176,8 +176,26 @@ jobs:
         export PATH=/tmp/bin:$PATH
         script/check-gn-build
 
-  macos:
+  macos-clang:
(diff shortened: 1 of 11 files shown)


jwhui:
@abtink, thanks for submitting this. Can we update the continuous-integration checks to ensure we have the appropriate coverage?

abtink:
Looks like we did not cover `script/check-simulation-build` under clang build workflow (on GH actions). And I think `macos` build was using gcc (not clang)? 

I added it in a new commit https://github.com/openthread/openthread/pull/4931/commits/26d720232e6d6e8e2941e190e5b2024f4a0335f3 which also enables "treat warning as error" on toranj build configs.

Issue #4734, opened 2020-03-25
[error] provide perror-like API to get a description of last error.


wgtdkp:
**Is your feature request related to a problem? Please describe.**
OT's error code is too generic to give more specific information about the failure. It is painful for developers & users to see only a error code in the CLI output (or log) when something is going wrong.

**Describe the solution you'd like**
provide an additional public API to get a descriptive message of the (last) error, for example:
```c++
const char *otErrorGetLastMessage(void);
```

**Describe alternatives you've considered**
OT Commissioner has another implementation of adapting [abseil/status](https://github.com/abseil/abseil-cpp/blob/master/absl/status/status.h) to include the message along with the Error class, since it provides only C++ APIs, but it seems to be a too large change to OT to apply this approach.

**Additional context**

PR #4934, opened 2020-05-07
[joiner] add validation of joiner PSKd

wgtdkp:
This PR adds validation of joiner PSKd to conform to Thread specifiction:

> A Joining Device Credential is encoded as uppercase alphanumeric characters (base32-thread: 0-9,A-Y excluding I,O,Q, and Z for readability) with a minimum length of 6 such characters and a maximum length of 32 such characters.

Related PR: https://github.com/openthread/ot-commissioner/pull/101

### DIFF
--- a/src/core/meshcop/joiner.cpp
+++ b/src/core/meshcop/joiner.cpp
@@ -87,10 +87,10 @@ exit:
     return;
 }
 
-otError Joiner::ValidatePskd(const char *aPskd)
+bool Joiner::IsPskdValid(const char *aPskd)
 {
-    otError error      = OT_ERROR_INVALID_ARGS;
-    size_t  pskdLength = StringLength(aPskd, kPskdMaxLength + 1);
+    bool   valid      = false;
+    size_t pskdLength = StringLength(aPskd, kPskdMaxLength + 1);
 
     OT_STATIC_ASSERT(static_cast<uint8_t>(kPskdMaxLength) <= static_cast<uint8_t>(Dtls::kPskMaxLength),
                      "The maximum length of DTLS PSK is smaller than joiner PSKd");
@@ -105,10 +105,10 @@ otError Joiner::ValidatePskd(const char *aPskd)
         VerifyOrExit(c != 'I' && c != 'O' && c != 'Q' && c != 'Z', OT_NOOP);
     }
 
-    error = OT_ERROR_NONE;
+    valid = true;
 
 exit:
-    return error;
+    return valid;
 }
 
 otError Joiner::Start(const char *     aPskd,
@@ -127,7 +127,7 @@ otError Joiner::Start(const char *     aPskd,
 
     VerifyOrExit(mState == OT_JOINER_STATE_IDLE, error = OT_ERROR_BUSY);
 
-    SuccessOrExit(error = ValidatePskd(aPskd));
+    VerifyOrExit(IsPskdValid(aPskd), error = OT_ERROR_INVALID_ARGS);
 
     // Use random-generated extended address.
     randomAddress.GenerateRandom();
--- a/src/core/meshcop/joiner.cpp
+++ b/src/core/meshcop/joiner.cpp
@@ -86,6 +86,27 @@ exit:
     return;
 }
 
+otError Joiner::ValidatePskd(const char *aPskd)
+{
+    otError error      = OT_ERROR_INVALID_ARGS;
+    size_t  pskdLength = strlen(aPskd);
+
+    VerifyOrExit(pskdLength >= kMinPskdLength && pskdLength <= kMaxPskdLength, OT_NOOP);
+
+    for (size_t i = 0; i < pskdLength; i++)
+    {
+        char c = aPskd[i];
+
+        VerifyOrExit(isdigit(c) || isupper(c), OT_NOOP);
+        VerifyOrExit(c != 'I' && c != 'O' && c != 'Q' && c != 'Z', OT_NOOP);
+    }
+
+    error = OT_ERROR_NONE;
+
+exit:
+    return error;
+}
+
 otError Joiner::Start(const char *     aPskd,
                       const char *     aProvisioningUrl,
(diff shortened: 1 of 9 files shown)


src/core/meshcop/joiner.hpp
```diff
@@ -109,12 +109,28 @@ class Joiner : public InstanceLocator
      */
     void GetJoinerId(Mac::ExtAddress &aJoinerId) const;
 
+    /**
+     * This method validates the PSKd.
+     *
+     * Per Thread specification, a Joining Device Credential is encoded as
+     * uppercase alphanumeric characters (base32-thread: 0-9,A-Y excluding
+     * I,O,Q, and Z for readability) with a minimum length of 6 such
```
jwhui, line 117:
```suggestion
     * uppercase alphanumeric characters (base32-thread: 0-9, A-Z excluding
     * I, O, Q, and Z for readability) with a minimum length of 6 such
```
nit: if we say we are excluding "Z", then we should include it in the range A-Z.

src/core/meshcop/joiner.hpp
```diff
@@ -109,12 +109,28 @@ class Joiner : public InstanceLocator
...
+     * characters and a maximum length of 32 such characters.
+     *
```
jwhui, line 119:
```suggestion
     *
     * param[in]  aPskd  The PSKd to validate.
```

src/core/meshcop/joiner.hpp
```diff
@@ -109,12 +109,28 @@ class Joiner : public InstanceLocator
...
+     * @retval OT_ERROR_NONE          The PSKd is valid.
+     * @retval OT_ERROR_INVALID_ARGS  The PSKd is invalid.
+     *
+     */
+    static otError ValidatePskd(const char *aPskd);
```
jwhui, line 124: nit: how about:
```suggestion
    static bool IsPskdValid(const char *aPskd);
```

src/core/meshcop/joiner.cpp
```diff
@@ -86,6 +86,30 @@ void Joiner::SetState(otJoinerState aState)
...
+                     "The maximum length of DTLS PSK is smaller than joiner PSKd");
+
+    VerifyOrExit(pskdLength >= kPskdMinLength && pskdLength <= kPskdMaxLength, OT_NOOP);
+
+    for (size_t i = 0; i < pskdLength; i++)
+    {
+        char c = aPskd[i];
+
+        VerifyOrExit(isdigit(c) || isupper(c), OT_NOOP);
+        VerifyOrExit(c != 'I' && c != 'O' && c != 'Q' && c != 'Z', OT_NOOP);
```
Irving-cl, line 104: What about using `switch` here to make it a little bit faster?

src/core/meshcop/joiner.hpp
wgtdkp: Agree.

src/core/meshcop/joiner.hpp
wgtdkp: I have considered this myself. But, if we can implement https://github.com/openthread/openthread/issues/4734, returning `otError` can give detials about why the `PSKd` is considered invalid (e.g. whether it exceed the length range or it includes invalid characters).

src/core/meshcop/joiner.cpp
wgtdkp: I don't think performance matters here, since the joiner API is not supposed to be called frequently.

src/core/meshcop/joiner.hpp
```diff
@@ -109,12 +109,29 @@ class Joiner : public InstanceLocator
...
 
+    /**
+     * This method validates the PSKd.
+     *
+     * Per Thread specification, a Joining Device Credential is encoded as
+     * uppercase alphanumeric characters (base32-thread: 0-9, A-Z excluding
+     * I, O, Q, and Z for readability) with a minimum length of 6 such
+     * characters and a maximum length of 32 such characters.
+     *
+     * param[in]  aPskd  The PSKd to validate.
```
simonlingoogle, line 120:
```suggestion
     * param[in]  aPskd  The PSKd to validate.
     *
```
Add blank line between params and retvals

src/core/meshcop/joiner.cpp
```diff
@@ -86,6 +86,30 @@ void Joiner::SetState(otJoinerState aState)
     return;
 }
 
+otError Joiner::ValidatePskd(const char *aPskd)
+{
+    otError error      = OT_ERROR_INVALID_ARGS;
+    size_t  pskdLength = strlen(aPskd);
```
simonlingoogle, line 92: use strnlen?

wgtdkp: 1. we use `strlen` more in our code (by searching `strlen` vs `strnlen`).
2. `strnlen` hides the true length of `aPSKd`.
3. `strnlen` is not the C++ standard before C++11 (`std::strnLen_s`).

abtink: Can use `ot::StringLength()` which is our internal `strnlen` 

We use it here in `commissioner` code for checking PSKd: https://github.com/openthread/openthread/blob/master/src/core/meshcop/commissioner.cpp#L288
(BTW, can we use the helper `Joiner::ValidatePskd()` in the above code as well?)

wgtdkp: @abtink 

> Can use ot::StringLength() which is our internal strnlen

Will do.

>BTW, can we use the helper Joiner::ValidatePskd() in the above code as well?

Yes, we did this. https://github.com/openthread/openthread/pull/4934/files#diff-adad1a583f73a28505dd375d3aaadd93

src/core/meshcop/joiner.hpp
wgtdkp: updated to return`bool`. We can consider returning `otError` when we are going to implement https://github.com/openthread/openthread/issues/4734. @jwhui please review.

PR #4935, opened 2020-05-07
[posix] add radio url

gjc13:
Radio url provides a unified interface for users to specify the radio
peripheral device and arguments to the OpenThread core stack.

### DIFF
--- a/script/check-ncp-rcp-migrate
+++ b/script/check-ncp-rcp-migrate
@@ -93,12 +93,12 @@ EOF
 
     echo "Step 2. Start retrieving dataset from Radio..."
     RADIO_NCP_PATH="$(pwd)/$(ls output/*linux*/bin/ot-ncp-ftd)"
-    "$(pwd)/$(ls output/posix/*linux*/bin/ot-ncp)" -n --radio-version --ncp-dataset -- "${RADIO_NCP_PATH}" 1
+    "$(pwd)/$(ls output/posix/*linux*/bin/ot-ncp)" -n --radio-version "spinel+hdlc+forkpty://${RADIO_NCP_PATH}?arg=1&ncp-dataset=1"
 
     echo "Step 3. Start posix app and check whether PAN dataset is the same..."
     RADIO_RCP_PATH="$(pwd)/$(ls output/*linux*/bin/ot-rcp)"
 
-    OT_CLI_CMD="$(pwd)/$(ls output/posix/*linux*/bin/ot-cli) ${RADIO_RCP_PATH} 1"
+    OT_CLI_CMD="$(pwd)/$(ls output/posix/*linux*/bin/ot-cli) spinel+hdlc+forkpty://${RADIO_RCP_PATH}?arg=1"
 
     expect <<EOF
 spawn ${OT_CLI_CMD}
@@ -126,8 +126,9 @@ expect eof
 EOF
 
     echo "Step 4. Start posix app and check whether it can get radio firmware version..."
-    RADIO_VERSION="$("$(pwd)/$(ls output/posix/*linux*/bin/ot-cli)" -n --radio-version --ncp-dataset -- "${RADIO_RCP_PATH}" 1)" || true
-    test -n "${RADIO_VERSION}"
+    RADIO_VERSION="$("$(pwd)/$(ls output/posix/*linux*/bin/ot-cli)" -n --radio-version "spinel+hdlc+forkpty://${RADIO_RCP_PATH}?arg=1&ncp-dataset=1")" || true
+    echo "${RADIO_VERSION}"
+    test -n "{RADIO_VERSION}"
 }
 
 main()
--- a/src/posix/main.c
+++ b/src/posix/main.c
@@ -75,6 +75,7 @@
 #error "Unknown posix app type!"
 #endif
 #include <common/code_utils.hpp>
+#include <common/logging.hpp>
 #include <lib/platform/exit_code.h>
 #include <openthread/openthread-system.h>
 
@@ -102,88 +103,31 @@ void __gcov_flush();
 enum
 {
     ARG_PRINT_RADIO_VERSION = 1001,
-    ARG_NO_RADIO_RESET      = 1002,
-    ARG_RESTORE_NCP_DATASET = 1003,
-    ARG_SPI_GPIO_INT_DEV    = 1011,
-    ARG_SPI_GPIO_INT_LINE   = 1012,
-    ARG_SPI_GPIO_RESET_DEV  = 1013,
-    ARG_SPI_GPIO_RESET_LINE = 1014,
-    ARG_SPI_MODE            = 1015,
-    ARG_SPI_SPEED           = 1016,
-    ARG_SPI_CS_DELAY        = 1017,
-    ARG_SPI_RESET_DELAY     = 1018,
-    ARG_SPI_ALIGN_ALLOWANCE = 1019,
-    ARG_SPI_SMALL_PACKET    = 1020,
-    ARG_MAX_POWER_TABLE     = 1021,
-
 };
 
 static const struct option kOptions[] = {{"debug-level", required_argument, NULL, 'd'},
                                          {"dry-run", no_argument, NULL, 'n'},
                                          {"help", no_argument, NULL, 'h'},
                                          {"interface-name", required_argument, NULL, 'I'},
-                                         {"no-reset", no_argument, NULL, ARG_NO_RADIO_RESET},
                                          {"radio-version", no_argument, NULL, ARG_PRINT_RADIO_VERSION},
-                                         {"ncp-dataset", no_argument, NULL, ARG_RESTORE_NCP_DATASET},
                                          {"time-speed", required_argument, NULL, 's'},
                                          {"verbose", no_argument, NULL, 'v'},
-#if OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE
-                                         {"max-power-table", required_argument, NULL, ARG_MAX_POWER_TABLE},
-#endif
-#if OPENTHREAD_POSIX_CONFIG_RCP_BUS == OT_POSIX_RCP_BUS_SPI
-                                         {"gpio-int-dev", required_argument, NULL, ARG_SPI_GPIO_INT_DEV},
-                                         {"gpio-int-line", required_argument, NULL, ARG_SPI_GPIO_INT_LINE},
-                                         {"gpio-reset-dev", required_argument, NULL, ARG_SPI_GPIO_RESET_DEV},
-                                         {"gpio-reset-line", required_argument, NULL, ARG_SPI_GPIO_RESET_LINE},
-                                         {"spi-mode", required_argument, NULL, ARG_SPI_MODE},
-                                         {"spi-speed", required_argument, NULL, ARG_SPI_SPEED},
-                                         {"spi-cs-delay", required_argument, NULL, ARG_SPI_CS_DELAY},
-                                         {"spi-reset-delay", required_argument, NULL, ARG_SPI_RESET_DELAY},
-                                         {"spi-align-allowance", required_argument, NULL, ARG_SPI_ALIGN_ALLOWANCE},
-                                         {"spi-small-packet", required_argument, NULL, ARG_SPI_SMALL_PACKET},
-#endif // OPENTHREAD_POSIX_CONFIG_RCP_BUS == OT_POSIX_RCP_BUS_SPI
                                          {0, 0, 0, 0}};
 
 static void PrintUsage(const char *aProgramName, FILE *aStream, int aExitCode)
 {
     fprintf(aStream,
             "Syntax:\n"
-            "    %s [Options] NodeId|Device|Command [DeviceConfig|CommandArgs]\n"
+            "    %s [Options] RadioURL\n"
             "Options:\n"
             "    -d  --debug-level             Debug level of logging.\n"
             "    -h  --help                    Display this usage information.\n"
             "    -I  --interface-name name     Thread network interface name.\n"
             "    -n  --dry-run                 Just verify if arguments is valid and radio spinel is compatible.\n"
-            "        --no-reset                Do not send Spinel reset command to RCP on initialization.\n"
             "        --radio-version           Print radio firmware version.\n"
-            "        --ncp-dataset             Retrieve and save NCP dataset to file.\n"
             "    -s  --time-speed factor       Time speed up factor.\n"
             "    -v  --verbose                 Also log to stderr.\n",
             aProgramName);
-#if OPENTHREAD_POSIX_CONFIG_RCP_BUS == OT_POSIX_RCP_BUS_SPI
-    fprintf(aStream,
-            "        --gpio-int-dev[=gpio-device-path]\n"
-            "                                  Specify a path to the Linux sysfs-exported GPIO device for the\n"
-            "                                  `I̅N̅T̅` pin. If not specified, `SPI` interface will fall back to\n"
-            "                                  polling, which is inefficient.\n"
-            "        --gpio-int-line[=line-offset]\n"
-            "                                  The offset index of `I̅N̅T̅` pin for the associated GPIO device.\n"
-            "                                  If not specified, `SPI` interface will fall back to polling,\n"
-            "                                  which is inefficient.\n"
-            "        --gpio-reset-dev[=gpio-device-path]\n"
-            "                                  Specify a path to the Linux sysfs-exported GPIO device for the\n"
-            "                                  `R̅E̅S̅` pin.\n"
-            "        --gpio-reset-line[=line-offset]"
-            "                                  The offset index of `R̅E̅S̅` pin for the associated GPIO device.\n"
-            "        --spi-mode[=mode]         Specify the SPI mode to use (0-3).\n"
-            "        --spi-speed[=hertz]       Specify the SPI speed in hertz.\n"
-            "        --spi-cs-delay[=usec]     Specify the delay after C̅S̅ assertion, in µsec.\n"
-            "        --spi-reset-delay[=ms]    Specify the delay after R̅E̅S̅E̅T̅ assertion, in milliseconds.\n"
-            "        --spi-align-allowance[=n] Specify the maximum number of 0xFF bytes to clip from start of\n"
-            "                                  MISO frame. Max value is 16.\n"
-            "        --spi-small-packet=[n]    Specify the smallest packet we can receive in a single transaction.\n"
-            "                                  (larger packets will require two transactions). Default value is 32.\n");
-#endif
 #if OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE
     fprintf(aStream,
             "        --max-power-table         Max power for channels in ascending order separated by commas,\n"
@@ -191,6 +135,7 @@ static void PrintUsage(const char *aProgramName, FILE *aStream, int aExitCode)
             "                                  the last value will be applied to all remaining channels.\n"
             "                                  Special value 0x7f disables a channel.\n");
 #endif
+    fprintf(aStream, "%s", otSysGetRadioUrlHelpString());
     exit(aExitCode);
 }
 
@@ -198,15 +143,8 @@ static void ParseArg(int aArgCount, char *aArgVector[], PosixConfig *aConfig)
 {
     memset(aConfig, 0, sizeof(*aConfig));
 
-    aConfig->mPlatformConfig.mSpeedUpFactor      = 1;
-    aConfig->mPlatformConfig.mResetRadio         = true;
-    aConfig->mPlatformConfig.mSpiSpeed           = OT_PLATFORM_CONFIG_SPI_DEFAULT_SPEED_HZ;
-    aConfig->mPlatformConfig.mSpiCsDelay         = OT_PLATFORM_CONFIG_SPI_DEFAULT_CS_DELAY_US;
-    aConfig->mPlatformConfig.mSpiResetDelay      = OT_PLATFORM_CONFIG_SPI_DEFAULT_RESET_DELAY_MS;
-    aConfig->mPlatformConfig.mSpiAlignAllowance  = OT_PLATFORM_CONFIG_SPI_DEFAULT_ALIGN_ALLOWANCE;
-    aConfig->mPlatformConfig.mSpiSmallPacketSize = OT_PLATFORM_CONFIG_SPI_DEFAULT_SMALL_PACKET_SIZE;
-    aConfig->mPlatformConfig.mSpiMode            = OT_PLATFORM_CONFIG_SPI_DEFAULT_MODE;
-    aConfig->mLogLevel                           = OT_LOG_LEVEL_CRIT;
+    aConfig->mPlatformConfig.mSpeedUpFactor = 1;
+    aConfig->mLogLevel                      = OT_LOG_LEVEL_CRIT;
 
     optind = 1;
 
@@ -253,47 +191,6 @@ static void ParseArg(int aArgCount, char *aArgVector[], PosixConfig *aConfig)
         case ARG_PRINT_RADIO_VERSION:
             aConfig->mPrintRadioVersion = true;
             break;
-        case ARG_NO_RADIO_RESET:
-            aConfig->mPlatformConfig.mResetRadio = false;
-            break;
-        case ARG_RESTORE_NCP_DATASET:
-            aConfig->mPlatformConfig.mRestoreDatasetFromNcp = true;
-            break;
-        case ARG_SPI_GPIO_INT_DEV:
-            aConfig->mPlatformConfig.mSpiGpioIntDevice = optarg;
-            break;
-        case ARG_SPI_GPIO_INT_LINE:
-            aConfig->mPlatformConfig.mSpiGpioIntLine = (uint8_t)atoi(optarg);
-            break;
-        case ARG_SPI_GPIO_RESET_DEV:
-            aConfig->mPlatformConfig.mSpiGpioResetDevice = optarg;
-            break;
-        case ARG_SPI_GPIO_RESET_LINE:
-            aConfig->mPlatformConfig.mSpiGpioResetLine = (uint8_t)atoi(optarg);
-            break;
-        case ARG_SPI_MODE:
-            aConfig->mPlatformConfig.mSpiMode = (uint8_t)atoi(optarg);
-            break;
-        case ARG_SPI_SPEED:
-            aConfig->mPlatformConfig.mSpiSpeed = (uint32_t)atoi(optarg);
-            break;
-        case ARG_SPI_CS_DELAY:
-            aConfig->mPlatformConfig.mSpiCsDelay = (uint16_t)atoi(optarg);
-            break;
-        case ARG_SPI_RESET_DELAY:
-            aConfig->mPlatformConfig.mSpiResetDelay = (uint32_t)atoi(optarg);
-            break;
-        case ARG_SPI_ALIGN_ALLOWANCE:
-            aConfig->mPlatformConfig.mSpiAlignAllowance = (uint8_t)atoi(optarg);
-            break;
-        case ARG_SPI_SMALL_PACKET:
-            aConfig->mPlatformConfig.mSpiSmallPacketSize = (uint8_t)atoi(optarg);
-            break;
-#if OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE
-        case ARG_MAX_POWER_TABLE:
-            aConfig->mPlatformConfig.mMaxPowerTable = optarg;
-            break;
-#endif
         case '?':
             PrintUsage(aArgVector[0], stderr, OT_EXIT_INVALID_ARGUMENTS);
             break;
@@ -307,13 +204,7 @@ static void ParseArg(int aArgCount, char *aArgVector[], PosixConfig *aConfig)
     {
         PrintUsage(aArgVector[0], stderr, OT_EXIT_INVALID_ARGUMENTS);
     }
-
-    aConfig->mPlatformConfig.mRadioFile = aArgVector[optind];
-
-    if (optind + 1 < aArgCount)
-    {
-        aConfig->mPlatformConfig.mRadioConfig = aArgVector[optind + 1];
-    }
+    aConfig->mPlatformConfig.mRadioUrl = aArgVector[optind];
 }
 
 static otInstance *InitInstance(int aArgCount, char *aArgVector[])
--- a/src/posix/platform/hdlc_interface.cpp
+++ b/src/posix/platform/hdlc_interface.cpp
@@ -138,30 +138,30 @@ HdlcInterface::HdlcInterface(SpinelInterface::ReceiveFrameCallback aCallback,
 {
 }
 
-otError HdlcInterface::Init(const otPlatformConfig &aPlatformConfig)
+otError HdlcInterface::Init(Arguments &aArguments)
 {
     otError     error = OT_ERROR_NONE;
     struct stat st;
 
     VerifyOrExit(mSockFd == -1, error = OT_ERROR_ALREADY);
 
-    VerifyOrDie(stat(aPlatformConfig.mRadioFile, &st) == 0, OT_EXIT_INVALID_ARGUMENTS);
+    VerifyOrDie(stat(aArguments.GetPath(), &st) == 0, OT_EXIT_INVALID_ARGUMENTS);
 
     if (S_ISCHR(st.st_mode))
     {
-        mSockFd = OpenFile(aPlatformConfig.mRadioFile, aPlatformConfig.mRadioConfig);
+        mSockFd = OpenFile(aArguments.GetPath(), aArguments);
         VerifyOrExit(mSockFd != -1, error = OT_ERROR_INVALID_ARGS);
     }
 #if OPENTHREAD_POSIX_CONFIG_RCP_PTY_ENABLE
     else if (S_ISREG(st.st_mode))
     {
-        mSockFd = ForkPty(aPlatformConfig.mRadioFile, aPlatformConfig.mRadioConfig);
+        mSockFd = ForkPty(aArguments.GetPath(), aArguments.GetValue("arg"));
         VerifyOrExit(mSockFd != -1, error = OT_ERROR_INVALID_ARGS);
     }
 #endif // OPENTHREAD_POSIX_CONFIG_RCP_PTY_ENABLE
     else
     {
-        otLogCritPlat("Radio file '%s' not supported", aPlatformConfig.mRadioFile);
+        otLogCritPlat("Radio file '%s' not supported", aArguments.GetPath());
         ExitNow(error = OT_ERROR_INVALID_ARGS);
     }
 
@@ -407,10 +407,12 @@ exit:
     return error;
 }
 
-int HdlcInterface::OpenFile(const char *aFile, const char *aConfig)
+int HdlcInterface::OpenFile(const char *aFile, Arguments &aArguments)
 {
-    int fd   = -1;
-    int rval = 0;
+    int         fd       = -1;
+    int         rval     = 0;
+    const char *parity   = aArguments.GetValue("uart-parity");
+    uint32_t    baudrate = 115200;
 
     fd = open(aFile, O_RDWR | O_NOCTTY | O_NONBLOCK | O_CLOEXEC);
     if (fd == -1)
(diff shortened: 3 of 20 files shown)


src/posix/platform/radio_url.hpp
```diff
@@ -0,0 +1,27 @@
+#ifndef POSIX_PLATFORM_RADIO_URL_HPP_
```
bukepo, line 1: license header

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,224 @@
+#include "posix/platform/radio_url.hpp"
```
bukepo, line 1: license header

src/posix/platform/spi_interface.hpp
```diff
@@ -49,6 +49,20 @@
 namespace ot {
 namespace Posix {
 
+struct SpiInterfaceConfig
```
bukepo, line 52: can these be members of the SpiInterface?

src/posix/platform/include/openthread/openthread-system.h
```diff
@@ -69,25 +69,8 @@ enum
...
-    uint8_t  mSpiGpioIntLine;     ///< Line index of the `I̅N̅T̅` pin for the associated GPIO character device.
-    uint8_t  mSpiGpioResetLine;   ///< Line index of the `R̅E̅S̅E̅T̅` pin for the associated GPIO character device.
-    uint8_t  mSpiMode;            ///< SPI mode to use (0-3).
-    uint32_t mSpiSpeed;           ///< SPI speed in hertz.
-    uint32_t mSpiResetDelay;      ///< The delay after R̅E̅S̅E̅T̅ assertion, in miliseconds.
-    uint16_t mSpiCsDelay;         ///< The delay after SPI C̅S̅ assertion, in µsec.
-    uint8_t  mSpiAlignAllowance;  ///< Maximum number of 0xFF bytes to clip from start of MISO frame.
-    uint8_t  mSpiSmallPacketSize; ///< Smallest SPI packet size we can receive in a single transaction.
+    uint32_t    mSpeedUpFactor; ///< Speed up factor.
+    const char *mInterfaceName; ///< Thread network interface name.
```
bukepo, line 73: add mRadioUrl?

src/posix/main.c
```diff
@@ -253,47 +193,9 @@ static void ParseArg(int aArgCount, char *aArgVector[], PosixConfig *aConfig)
...
-        case ARG_SPI_MODE:
-            aConfig->mPlatformConfig.mSpiMode = (uint8_t)atoi(optarg);
-            break;
-        case ARG_SPI_SPEED:
-            aConfig->mPlatformConfig.mSpiSpeed = (uint32_t)atoi(optarg);
-            break;
-        case ARG_SPI_CS_DELAY:
-            aConfig->mPlatformConfig.mSpiCsDelay = (uint16_t)atoi(optarg);
+        case ARG_RADIO_URL:
+            *aRadioUrl = optarg;
```
bukepo, line 197: memcpy()?

script/check-ncp-rcp-migrate
```diff
@@ -90,12 +90,12 @@ EOF
 
     echo "Step 2. Start retrieving dataset from Radio..."
     RADIO_NCP_PATH="$(pwd)/$(ls output/*linux*/bin/ot-ncp-ftd)"
-    "$(pwd)/$(ls output/posix/*linux*/bin/ot-ncp)" -n --radio-version --ncp-dataset -- ${RADIO_NCP_PATH} 1
+    "$(pwd)/$(ls output/posix/*linux*/bin/ot-ncp)" -n --radio-version --radio-url "spinel+exec://${RADIO_NCP_PATH}?args=1&restoredataset=1"
```
bukepo, line 93:
```suggestion
    "$(pwd)/$(ls output/posix/*linux*/bin/ot-ncp)" -n --radio-version --radio-url "spinel+hdlc+forkpty://${RADIO_NCP_PATH}?arg=1&restoredataset=1"
```

script/check-ncp-rcp-migrate
```diff
@@ -90,12 +90,12 @@ EOF
...
 
     echo "Step 3. Start posix app and check whether PAN dataset is the same..."
     RADIO_RCP_PATH="$(pwd)/$(ls output/*linux*/bin/ot-rcp)"
 
-    OT_CLI_CMD="$(pwd)/$(ls output/posix/*linux*/bin/ot-cli) ${RADIO_RCP_PATH} 1"
+    OT_CLI_CMD="$(pwd)/$(ls output/posix/*linux*/bin/ot-cli) --radio-url spinel+exec://${RADIO_RCP_PATH}?args=1"
```
bukepo, line 98:
```suggestion
    OT_CLI_CMD="$(pwd)/$(ls output/posix/*linux*/bin/ot-cli) --radio-url spinel+hdlc+forkpty://${RADIO_RCP_PATH}?arg=1"
```

script/check-ncp-rcp-migrate
```diff
@@ -123,7 +123,7 @@ expect eof
 EOF
 
     echo "Step 4. Start posix app and check whether it can get radio firmware version..."
-    RADIO_VERSION="$("$(pwd)/$(ls output/posix/*linux*/bin/ot-cli)" -n --radio-version --ncp-dataset -- ${RADIO_RCP_PATH} 1)" || true
+    RADIO_VERSION="$("$(pwd)/$(ls output/posix/*linux*/bin/ot-cli)" -n --radio-version --radio-url "spinel+exec://${RADIO_RCP_PATH}?args=1&restoredataset=1" )" || true
```
bukepo, line 126:
```suggestion
    RADIO_VERSION="$("$(pwd)/$(ls output/posix/*linux*/bin/ot-cli)" -n --radio-version --radio-url "spinel+hdlc+forkpty://${RADIO_RCP_PATH}?arg=1&restoredataset=1" )" || true
```

src/posix/main.c
```diff
@@ -75,13 +75,18 @@
...
 #include <common/code_utils.hpp>
+#include <common/logging.hpp>
 #include <lib/platform/exit_code.h>
 #include <openthread/openthread-system.h>
 
 #ifndef OPENTHREAD_ENABLE_COVERAGE
 #define OPENTHREAD_ENABLE_COVERAGE 0
 #endif
 
+#ifndef OPENTHREAD_RADIO_URL_MAX_SIZE
```
bukepo, line 86: move this into platform and copy the url there

src/posix/platform/radio.cpp
```diff
@@ -94,18 +94,22 @@ void otPlatRadioSetPromiscuous(otInstance *aInstance, bool aEnable)
     SuccessOrDie(sRadioSpinel.SetPromiscuous(aEnable));
 }
 
-void platformRadioInit(const otPlatformConfig *aPlatformConfig)
+void platformRadioInit(otPosixRadioArguments *aArguments)
 {
+    ot::Posix::Arguments *args           = reinterpret_cast<ot::Posix::Arguments *>(aArguments);
+    bool                  resetRadio     = (args->GetValue("noreset") == NULL);
```
bukepo, line 100: @jwhui @abtink May I have your input about the parameter name style, should the name include `-` or not? ```suggestion
    bool                  resetRadio     = (args->GetValue("no-reset") == NULL);
```

File: src/posix/main.c [hunk 11]
```diff
@@ -153,37 +129,12 @@ static void PrintUsage(const char *aProgramName, FILE *aStream, int aExitCode)
             "    -d  --debug-level             Debug level of logging.\n"
             "    -h  --help                    Display this usage information.\n"
             "    -I  --interface-name name     Thread network interface name.\n"
+            "        --radio-url url           Radio url.\n"
```
FiveDimensions, line 132: if posix need to parse protocols, radio dev, args from the url, is it better to set url as a required argument instead of option "--radio-url"?

src/posix/platform/radio.cpp
gjc13: @jwhui @abtink Friendly reminder

src/posix/main.c
```diff
@@ -102,94 +103,74 @@ void __gcov_flush();
...
+                     " spi-cs-delay[=usec]       Specify the delay after C̅S̅ assertion, in µsec.\n"
+                     " spi-reset-delay[=ms]      Specify the delay after R̅E̅S̅E̅T̅ assertion, in milliseconds.\n"
+                     " spi-align-allowance[=n]   Specify the maximum number of 0xFF bytes to clip from start of\n"
+                     "                           MISO frame. Max value is 16.\n"
+                     " spi-small-packet=[n]      Specify the smallest packet we can receive in a single transaction.\n"
+                     "                           (larger packets will require two transactions). Default value is 32.\n"
+
+    );
+#else
+    fprintf(aStream, "spinel+uart://${PATH_TO_UART_DEVICE}?${arguments} for real uart device\n"
```
bukepo, line 166:
```suggestion
    fprintf(aStream, "spinel+hdlc+uart://${PATH_TO_UART_DEVICE} for real uart device\n"
```

src/posix/main.c
```diff
@@ -102,94 +103,74 @@ void __gcov_flush();
...
+                     "arguments:\n"
+                     " parity[=even|odd]         Uart parity config, optional.\n"
```
bukepo, line 168:
```suggestion
                     " tty-parity[=even|odd]         TTY parity config, optional.\n"
```

src/posix/platform/spi_interface.cpp
```diff
@@ -91,32 +91,78 @@ SpiInterface::SpiInterface(SpinelInterface::ReceiveFrameCallback aCallback,
...
+    uint16_t    spiCsDelay         = OT_PLATFORM_CONFIG_SPI_DEFAULT_CS_DELAY_US;
+    uint8_t     spiAlignAllowance  = OT_PLATFORM_CONFIG_SPI_DEFAULT_ALIGN_ALLOWANCE;
+    uint8_t     spiSmallPacketSize = OT_PLATFORM_CONFIG_SPI_DEFAULT_SMALL_PACKET_SIZE;
 
-    mSpiCsDelayUs       = aPlatformConfig.mSpiCsDelay;
-    mSpiSmallPacketSize = aPlatformConfig.mSpiSmallPacketSize;
-    mSpiAlignAllowance  = aPlatformConfig.mSpiAlignAllowance;
+    spiGpioIntDevice   = aArguments.GetValue("gpio-int-device");
+    spiGpioResetDevice = aArguments.GetValue("gpio-reset-device");
+    if (aArguments.GetValue("gpio-int-line"))
```
bukepo, line 109: Is this a required argument?

src/posix/platform/spi_interface.hpp
```diff
@@ -49,6 +49,10 @@
...
```
bukepo, line 52: Remove this?

src/posix/platform/radio_url.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+namespace ot {
+namespace Posix {
+
+class Arguments : public otPosixRadioArguments
+{
+public:
+    Arguments(const char *aUrl);
+    const char *GetPath(void) const { return mDevice; }
+    const char *GetValue(const char *aName, const char *aLastValue = NULL);
```
bukepo, line 53: missing doc string.

src/posix/platform/system.cpp
```diff
@@ -44,15 +44,27 @@
...
     otInstance *instance = NULL;
 
 #if OPENTHREAD_POSIX_VIRTUAL_TIME
     virtualTimeInit();
 #endif
+
     platformAlarmInit(aPlatformConfig->mSpeedUpFactor);
-    platformRadioInit(aPlatformConfig);
+    {
+        char                 radioUrl[OPENTHREAD_RADIO_URL_MAX_SIZE];
```
bukepo, line 61: remove this?

src/posix/platform/system.cpp
```diff
@@ -44,15 +44,27 @@
 
 uint64_t gNodeId = 0;
 
+#ifndef OPENTHREAD_RADIO_URL_MAX_SIZE
```
bukepo, line 47: remove?

src/posix/platform/radio_url.hpp
```diff
@@ -0,0 +1,66 @@
...
+#define POSIX_PLATFORM_RADIO_URL_HPP_
+
+#include <stdlib.h>
+
+#include <openthread/openthread-system.h>
+
+extern "C" {
+typedef struct otPosixRadioArguments
+{
+    const char *mDevice;
```
bukepo, line 39: Add doc string.
```suggestion
    const char *mPath; ///< The path to the executable or device.
```

src/posix/platform/spi_interface.cpp
gjc13: Yes.

src/posix/main.c
```diff
@@ -102,94 +103,74 @@ void __gcov_flush();
...
-#endif
 #if OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE
     fprintf(aStream,
             "        --max-power-table         Max power for channels in ascending order separated by commas,\n"
             "                                  If the number of values is less than that of supported channels,\n"
             "                                  the last value will be applied to all remaining channels.\n"
             "                                  Special value 0x7f disables a channel.\n");
+#endif
+#if OPENTHREAD_POSIX_CONFIG_RCP_BUS == OT_POSIX_RCP_BUS_SPI
+    fprintf(aStream, "Radio URL:\n"
```
bukepo, line 139: I think we should make this string an API of the platform layer. so that otbr-agent can use this as well.

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,208 @@
...
+           " spi-mode[=mode]           Specify the SPI mode to use (0-3).\n"
+           " spi-speed[=hertz]         Specify the SPI speed in hertz.\n"
+           " spi-cs-delay[=usec]       Specify the delay after C̅S̅ assertion, in µsec.\n"
+           " spi-reset-delay[=ms]      Specify the delay after R̅E̅S̅E̅T̅ assertion, in milliseconds.\n"
+           " spi-align-allowance[=n]   Specify the maximum number of 0xFF bytes to clip from start of\n"
+           "                           MISO frame. Max value is 16.\n"
+           " spi-small-packet=[n]      Specify the smallest packet we can receive in a single transaction.\n"
+           "                           (larger packets will require two transactions). Default value is 32.\n";
+#else
+    return "spinel+hdlc+uart://${PATH_TO_UART_DEVICE}?${arguments} for real uart device\n"
```
bukepo, line 67: missing "Radio URL:\n"

src/posix/platform/include/openthread/openthread-system.h
```diff
@@ -156,6 +142,14 @@ int otSysMainloopPoll(otSysMainloopContext *aMainloop);
  */
 void otSysMainloopProcess(otInstance *aInstance, const otSysMainloopContext *aMainloop);
 
+/**
+ * This method returns the radio url help string
```
bukepo, line 146:
```suggestion
 * This method returns the radio url help string.
```

src/posix/platform/hdlc_interface.cpp
```diff
@@ -421,150 +422,132 @@ int HdlcInterface::OpenFile(const char *aFile, const char *aConfig)
...
 
         VerifyOrExit((rval = tcgetattr(fd, &tios)) == 0, OT_NOOP);
 
         cfmakeraw(&tios);
 
         tios.c_cflag = CS8 | HUPCL | CREAD | CLOCAL;
 
-        // example: 115200N1H
-        if (aConfig != NULL)
+        if (parity)
```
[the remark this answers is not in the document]
bukepo: @gjc13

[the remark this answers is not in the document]
gjc13: I've checked in the code block. It can be only "even" or "odd".

src/posix/platform/hdlc_interface.cpp
```diff
@@ -423,40 +425,38 @@ int HdlcInterface::OpenFile(const char *aFile, const char *aConfig)
...
+        {
+            tios.c_cflag |= PARENB;
+            if (strncmp(parity, "odd", 3) == 0)
```
bukepo, line 440:
```suggestion
            if (strncmp(parity, "odd", sizeof("odd") - 1) == 0)
```

src/posix/platform/radio.cpp
```diff
@@ -94,18 +94,22 @@ void otPlatRadioSetPromiscuous(otInstance *aInstance, bool aEnable)
...
 {
+    ot::Posix::Arguments *args           = reinterpret_cast<ot::Posix::Arguments *>(aArguments);
+    bool                  resetRadio     = (args->GetValue("no-reset") == NULL);
+    bool                  restoreDataset = (args->GetValue("restore-dataset") != NULL);
 #if OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE
-    uint8_t channel = ot::Radio::kChannelMin;
-    int8_t  power   = ot::Posix::MaxPowerTable::kPowerDefault;
+    uint8_t     channel       = ot::Radio::kChannelMin;
+    int8_t      power         = ot::Posix::MaxPowerTable::kPowerDefault;
+    const char *maxPowerTable = args->GetValue("powertable");
```
bukepo, line 105:
```suggestion
    const char *maxPowerTable = args->GetValue("max-power-table");
```

src/posix/platform/hdlc_interface.cpp
```diff
@@ -423,40 +425,39 @@ int HdlcInterface::OpenFile(const char *aFile, const char *aConfig)
...
+            tios.c_cflag |= PARENB;
+            tios.c_cflag &= !PARODD;
+            if (strncmp(parity, "odd", 3) == 0)
+            {
+                tios.c_cflag |= PARODD;
+            }
+            else if (strncmp(parity, "even", 4))
+            {
+                DieNow(OT_EXIT_INVALID_ARGUMENTS);
+            }
```
bukepo:
```suggestion
            if (strncmp(parity, "odd", sizeof("odd") - 1) == 0)
            {
                tios.c_cflag |= PARODD | PARENB;
            }
            else if (strncmp(parity, "even", sizeof("even") - 1) == 0)
            {
                tios.c_cflag |= PARENB;
                tios.c_cflag &= ~PARODD;
            }
            else
            {
                DieNow(OT_EXIT_INVALID_ARGUMENTS);
            }
```

src/posix/platform/hdlc_interface.cpp
```diff
@@ -423,40 +425,39 @@ int HdlcInterface::OpenFile(const char *aFile, const char *aConfig)
...
+        }
+        else
         {
-            sscanf(aConfig, "%u%c%d%c", &speed, &parity, &cstopb, &flow);
+            tios.c_cflag &= (!(PARENB | PARODD));
```
bukepo, line 456:
```suggestion
            tios.c_cflag &= (~(PARENB | PARODD));
```

src/posix/platform/hdlc_interface.hpp
```diff
@@ -74,14 +74,14 @@ class HdlcInterface
...
-     * @param[in]  aPlatformConfig  Platform configuration structure.
+     * @param[in]  aArguments         Arguments parsed from radio url.
      *
      * @retval OT_ERROR_NONE          The interface is initialized successfully
      * @retval OT_ERROR_ALREADY       The interface is already initialized.
      * @retval OT_ERROR_INVALID_ARGS  The UART device or executable cannot be found or failed to open/run.
      *
      */
-    otError Init(const otPlatformConfig &aPlatformConfig);
+    otError Init(ot::Posix::Arguments &aArguments);
```
bukepo, line 84:
```suggestion
    otError Init(Arguments &aArguments);
```

src/posix/platform/hdlc_interface.hpp
```diff
@@ -198,7 +198,7 @@ class HdlcInterface
     static void HandleHdlcFrame(void *aContext, otError aError);
     void        HandleHdlcFrame(otError aError);
 
-    static int OpenFile(const char *aFile, const char *aConfig);
+    static int OpenFile(const char *aFile, ot::Posix::Arguments &aArguments);
```
bukepo, line 201:
```suggestion
    static int OpenFile(const char *aFile, Arguments &aArguments);
```

src/posix/platform/include/openthread/openthread-system.h
```diff
@@ -69,25 +70,9 @@ enum
...
-    uint8_t  mSpiGpioResetLine;   ///< Line index of the `R̅E̅S̅E̅T̅` pin for the associated GPIO character device.
-    uint8_t  mSpiMode;            ///< SPI mode to use (0-3).
-    uint32_t mSpiSpeed;           ///< SPI speed in hertz.
-    uint32_t mSpiResetDelay;      ///< The delay after R̅E̅S̅E̅T̅ assertion, in miliseconds.
-    uint16_t mSpiCsDelay;         ///< The delay after SPI C̅S̅ assertion, in µsec.
-    uint8_t  mSpiAlignAllowance;  ///< Maximum number of 0xFF bytes to clip from start of MISO frame.
-    uint8_t  mSpiSmallPacketSize; ///< Smallest SPI packet size we can receive in a single transaction.
+    uint32_t    mSpeedUpFactor; ///< Speed up factor.
+    const char *mInterfaceName; ///< Thread network interface name.
+    const char *mRadioUrl;      ///< Radio url.
```
bukepo, line 75:
```suggestion
    const char *mInterfaceName; ///< Thread network interface name.
    const char *mRadioUrl;      ///< Radio url.
    uint32_t    mSpeedUpFactor; ///< Speed up factor.
```

src/posix/platform/radio.cpp
```diff
@@ -94,18 +94,22 @@ void otPlatRadioSetPromiscuous(otInstance *aInstance, bool aEnable)
     SuccessOrDie(sRadioSpinel.SetPromiscuous(aEnable));
 }
 
-void platformRadioInit(const otPlatformConfig *aPlatformConfig)
+void platformRadioInit(otPosixRadioArguments *aArguments)
 {
+    ot::Posix::Arguments *args           = reinterpret_cast<ot::Posix::Arguments *>(aArguments);
+    bool                  resetRadio     = (args->GetValue("no-reset") == NULL);
+    bool                  restoreDataset = (args->GetValue("restore-dataset") != NULL);
```
bukepo, line 101: use the original parameter name?
```suggestion
    bool                  restoreDataset = (args->GetValue("ncp-dataset") != NULL);
```

src/posix/platform/spi_interface.hpp
```diff
@@ -79,14 +79,14 @@ class SpiInterface
...
-     * @param[in]  aPlatformConfig  Platform configuration structure.
+     * @param[in]  aArguments         Arguments parsed from radio url.
      *
      * @retval OT_ERROR_NONE          The interface is initialized successfully.
      * @retval OT_ERROR_ALREADY       The interface is already initialized.
      * @retval OT_ERROR_INVALID_ARGS  The UART device or executable cannot be found or failed to open/run.
      *
      */
-    otError Init(const otPlatformConfig &aPlatformConfig);
+    otError Init(ot::Posix::Arguments &aArguments);
```
bukepo, line 89:
```suggestion
    otError Init(Arguments &aArguments);
```

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,209 @@
...
+namespace ot {
+namespace Posix {
+
+Arguments::Arguments(const char *aUrl)
+{
+    char *url = &mUrl[0];
+
+    VerifyOrDie(aUrl != NULL, OT_EXIT_INVALID_ARGUMENTS);
+
+    strncpy(mUrl, aUrl, sizeof(mUrl));
```
bukepo, line 88:
```suggestion
    mUrl[sizeof(mUrl) - 1] = '\0';
    strncpy(mUrl, aUrl, sizeof(mUrl) - 1);
```

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,209 @@
...
+    url = strstr(url, "://");
+    url += sizeof("://") - 1;
```
bukepo, line 90: die if url is null?

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,209 @@
...
+}
+
+namespace ot {
+namespace Posix {
+
+Arguments::Arguments(const char *aUrl)
+{
+    char *url = &mUrl[0];
+
+    VerifyOrDie(aUrl != NULL, OT_EXIT_INVALID_ARGUMENTS);
```
bukepo, line 86: How about adding a IsValid() method to validate the URL so that this utility doesn't rely on *Die()?

src/posix/platform/hdlc_interface.hpp
```diff
@@ -74,14 +74,14 @@ class HdlcInterface
...
-     * @param[in]  aPlatformConfig  Platform configuration structure.
+     * @param[in]  aArguments         Arguments parsed from radio url.
      *
      * @retval OT_ERROR_NONE          The interface is initialized successfully
      * @retval OT_ERROR_ALREADY       The interface is already initialized.
      * @retval OT_ERROR_INVALID_ARGS  The UART device or executable cannot be found or failed to open/run.
      *
      */
-    otError Init(const otPlatformConfig &aPlatformConfig);
+    otError Init(Arguments &aArguments);
```
Irving-cl, line 84: Shall we use `const`?

src/posix/platform/include/openthread/openthread-system.h
```diff
@@ -97,6 +82,7 @@ typedef struct otPlatformConfig
  *       when initialization of OpenThread's drivers is most appropriate.
  *
  * @param[in]  aPlatformConfig  Platform configuration structure.
+ * @param[in]  aRadioUrl        Radio url.
```
Irving-cl, line 85: The below function `otSysInit` doesn't take this parameter. Does the function need this parameter?

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+#include <stdio.h>
+#include <string.h>
+
+#include <openthread/openthread-system.h>
+
+#include "core/common/code_utils.hpp"
+#include "posix/platform/platform-posix.h"
+
+const char *otSysGetRadioUrlHelpString(void)
```
zhanglongxia, line 39: How about move the implementation of this function to `spi_interface.cpp` and `hdlc_interface.cpp` respectively? When we add a new interface to posix in the future, we can just implement this function in the new interface.

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+exit:
+    return rval;
+}
+
+} // namespace Posix
+} // namespace ot
+
+#if OT_UNIT_TEST_RADIO_URL
```
zhanglongxia, line 151: "OT_UNIT_TEST_RADIO_URL" is not defined, compiler will generate a warning.
How about move this unit test to directory `tests/unit` ?

src/posix/platform/radio_url.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <stdlib.h>
+
+#include <openthread/openthread-system.h>
+
+extern "C" {
+typedef struct otPosixRadioArguments
+{
+    const char *mPath; //< The path to the executable or device
+    char *      mStart;
+    char *      mEnd;
```
zhanglongxia, line 42: Should we also add comments after `mStart` and `mEnd`?

src/posix/platform/radio_url.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <openthread/openthread-system.h>
+
+extern "C" {
+typedef struct otPosixRadioArguments
+{
+    const char *mPath; //< The path to the executable or device
```
zhanglongxia, line 40: nit: ///<

src/posix/platform/radio_url.hpp
```diff
@@ -0,0 +1,84 @@
...
+     * @returns The path in url.
+     *
+     */
+    const char *GetPath(void) const { return mPath; }
+
+    /**
+     * This method returns the url agument value.
+     *
+     * @param[in] aName       Argument name.
+     * @param[in] aLastValue  The last iterated argument value, NULL for first value.
```
zhanglongxia, line 66: Do we really need the argument `aLastValue`. It seems only the unit test uses this argument.

src/posix/platform/spi_interface.hpp
```diff
@@ -79,14 +79,14 @@ class SpiInterface
...
-     * @param[in]  aPlatformConfig  Platform configuration structure.
+     * @param[in]  aArguments         Arguments parsed from radio url.
      *
      * @retval OT_ERROR_NONE          The interface is initialized successfully.
      * @retval OT_ERROR_ALREADY       The interface is already initialized.
      * @retval OT_ERROR_INVALID_ARGS  The UART device or executable cannot be found or failed to open/run.
      *
      */
-    otError Init(const otPlatformConfig &aPlatformConfig);
+    otError Init(Arguments &aArguments);
```
zhanglongxia, line 89: nit: const

src/posix/platform/radio_url.hpp
```diff
@@ -0,0 +1,84 @@
...
+     *
+     * @returns The argument value.
+     *
+     */
+    const char *GetValue(const char *aName, const char *aLastValue = NULL);
```
zhanglongxia, line 71: nit: add `const` to the end of the function

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,215 @@
...
+Arguments::Arguments(const char *aUrl)
+{
+    char *url = &mUrl[0];
+
+    mPath  = NULL;
+    mStart = NULL;
+    mEnd   = NULL;
+
+    VerifyOrExit(aUrl != NULL, OT_NOOP);
+    strncpy(mUrl, aUrl, sizeof(mUrl) - 1);
```
zhanglongxia, line 91: Should we compare the length of `aUrl ` and `mUrl `? The length of the `aUrl ` may larger than the size of `mUrl`. In that case, there is no need to deal with `aUrl`.

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,215 @@
...
+    {
+        char *last = NULL;
+
+        for (char *cur = strtok(start, "&"); cur != NULL; cur = strtok(NULL, "&"))
+        {
+            if (!strncmp(aName, cur, len))
+            {
+                if (cur[len] == '=')
+                {
+                    ExitNow(rval = &cur[len + 1]);
```
zhanglongxia, line 131: The return value is a string which is ended by '?' rather than '\0' most of the time. Will it cause a fault when we add a new argument that is a string value in the future?

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,215 @@
...
+#include <string.h>
+
+#include <openthread/openthread-system.h>
+
+#include "core/common/code_utils.hpp"
+#include "posix/platform/platform-posix.h"
+
+const char *otSysGetRadioUrlHelpString(void)
+{
+#if OPENTHREAD_POSIX_CONFIG_RCP_BUS == OT_POSIX_RCP_BUS_SPI
```
zhanglongxia, line 41: The URL represents the UART or SPI bus parameters, how about change the file name from `radio_url.cpp` to `bus_url.cpp`?

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,215 @@
...
+    url += sizeof("://") - 1;
+    mPath = url;
+
+    mStart = strstr(url, "?");
+
+    if (mStart != NULL)
+    {
+        mStart[0] = '\0';
+
+        mStart += sizeof("?") - 1;
```
zhanglongxia, line 103: How about change the line 103 to `mStart += strlen("?");`or `mStart++;`  to make it easier for reader to understand?

src/posix/platform/radio_url.hpp
gjc13: This is reserved for future use of passing multiple arguments with the same key.

src/posix/platform/radio_url.cpp
gjc13: `strtok` will replace the delimiter with `\0`.

src/posix/platform/spi_interface.hpp
gjc13: Since `Arguments` uses `strtok`, we cannot pass it as const here.

src/posix/platform/radio_url.hpp
gjc13: Since Arguments uses strtok, we cannot pass it as const here.

src/posix/platform/hdlc_interface.hpp
gjc13: Since Arguments uses strtok, we cannot pass it as const here.

src/posix/platform/radio_url.cpp
gjc13: Since we are also including arguments like `no-reset` and `max-power-table` in radio url. We have to gather all the help strings here.

src/posix/platform/radio_url.cpp
gjc13: Since we are also including arguments like no-reset and max-power-table in radio url. I'd like to keep this name.

gjc13:
@jwhui Ready for merging. This change is required for supporting spi in otbr-agent.

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,224 @@
...
+{
+    const char * rval  = NULL;
+    const size_t len   = strlen(aName);
+    char *       start = (aLastValue == NULL ? mStart : (const_cast<char *>(aLastValue) + strlen(aLastValue) + 1));
+
+    while (start < mEnd)
+    {
+        char *last = NULL;
+
+        for (char *cur = strtok(start, "&"); cur != NULL; cur = strtok(NULL, "&"))
```
abtink, line 129: I assume we should be able to call `GetValue()` with different `aName` sequentially and in any order we want. Say  we have a URL string such that `mStart` is `"first=1&second=2&third=3"`, we should be able to call `GetValue("third")` first and then `GetValue("second")`.

If I am not mistaken, with current code this would not work.
- The `strtok` would actually update the original string itself and replace the delimiter `&` with `\0`
- So I think first time going through this, we would get all tokens but now the original string does not have `"&"` anymore between each toekn/arg
  - The original string pointed by `mStart` would become `"first=1\0second=2\0third=3"`
- Next call to `GetValue()` we again start from `mStart` but this time it cannot go through the full list since now `strtok` would only see `"first=1"`

src/posix/platform/spi_interface.cpp
```diff
@@ -91,32 +91,91 @@ SpiInterface::SpiInterface(SpinelInterface::ReceiveFrameCallback aCallback,
...
+    {
+        spiCsDelay = static_cast<uint16_t>(atoi(aArguments.GetValue("spi-cs-delay")));
+    }
+    if (aArguments.GetValue("spi-align-allowance"))
+    {
+        spiAlignAllowance = static_cast<uint8_t>(atoi(aArguments.GetValue("spi-align-allowance")));
+    }
+    if (aArguments.GetValue("spi-small-packet"))
+    {
+        spiSmallPacketSize = static_cast<uint8_t>(atoi(aArguments.GetValue("spi-small-packet")));
```
abtink, line 152: Another thing that `getopt_long()` would help is to detect missing required value for an option when the `option` struct specified the `required_argument` for an option entry. I think in this case `getopt_long()` returns `?` which we handle as error in the code and exit.

I think with the `GetValue()` we cannot detech that and fallback to empty string as value if none is given.

src/posix/platform/radio_url.cpp
```diff
@@ -0,0 +1,224 @@
...
+    if (mStart != NULL)
+    {
+        mStart[0] = '\0';
+        mStart++;
+
+        mEnd = mStart + strlen(mStart);
+    }
+    else
+    {
+        mEnd = url + strlen(url);
```
abtink, line 113: From this `else` path, we will have `mStart` as NULL and `mEnd` not NULL.
Later from `GetValue()` we can comapre `start < mEnd` where `start` is set as `mStart` and therefore is `NULL`
IIRC doing `NULL < mEnd` is not well-defined (undefined behavior). Can we consider avoiding this situation? Why do we need `mEnd` when `mStart` is NULL?

src/posix/platform/radio_url.cpp
abtink: I re-read the code and realized how it handles the changed the `mStart` string (by remembering `last` string and skipping over it and the `\0` from the `while (start < mEnd)` loop). So the code should work correctly as is.

How about we do the `strtok` from the constructor (which then replaces all `&` with `\0` in string once) and then in `GetValue()` just use the iteration from `start` to `mEnd` skipping over each string by its length? I feel this helps make the code simpler and avoid confusion (like my mistake reading it last night... :)...).

src/posix/platform/spi_interface.cpp
gjc13: The current behavior is that we return `nullptr` for keys not specified in the url. The consumer of the arguments takes the responsibility of ensuring the required ones. It will be a fancy feature to report un-recognized arguments but maybe we can leave for future improvements since ot-br-posix is pending on this feature to fully support all possible radio interfaces.

PR #4936, opened 2020-05-07
[mac] count out-of-band Tx frames

### LINKED ISSUE #4459: MAC counters doesn't count out of band tx frames
tx out of band frames are also 15.4 frames, no matter how it is secured, the frame type can be retrieved. Shall we count these frames?

count frames according to the frame type instead of tx operation.

count these frames as other frames because there's other rx frame counter.

**Additional context**
This is helpful to analyze power consumption.

FiveDimensions:
Fix #4459 : MAC counters doesn't count out of band tx frames


include/openthread/platform/radio.h
```diff
@@ -217,6 +217,7 @@ typedef struct otRadioFrame
 
             // Flags
             bool mAckedWithFramePending : 1; /// This indicates if this frame was acknowledged with frame pending set.
+            bool mIsOobFrame;                ///< True if this frame is set to be an out-of-band frame for counting.
```
Irving-cl, line 220:
```suggestion
            bool mIsOobFrame : 1;            ///< True if this frame is set to be an out-of-band frame for counting.
```
Use bit field here?

src/core/mac/#mac.cpp#
```diff
@@ -0,0 +1,2112 @@
+/*
```
librasungirl, line 1: what's the purpose of this file?

FiveDimensions: occasionally generated invalid file, deleted

src/core/mac/mac.cpp
```diff
@@ -1375,7 +1377,48 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
         break;
 #endif
 
+        // count Oob frames based on type
+
     case kOperationTransmitOutOfBandFrame:
+        switch (aFrame.GetType())
```
abtink, line 1383: I won't recommend parsing the OOB frame to update the counters (e.g., determine if is if MAC command etc).
Would suggest just incrementing the `mTxOther` to keep it simple.

The intention behind the OOB frame was to send a custom frame (application/vensor specific) frame  which may not even follow full 802.15.4 frame/header format and/or may be encrypted differently.

In longer term I think it would be good to remove the whole OOB feature (it was intended to solve a problem for a legacy networking solution). If such a use is needed in future, I would recommend handling it at platform layer and radio driver level (instead of having an API which allows a frame to be injected through OT code).

FiveDimensions: @abtink Thanks for the suggestion, the simple way is increasing mTxOther.  In the discussion for issue[#4459](https://github.com/openthread/openthread/issues/4459) ,  the recommended solution is counting Oob frames based on the frame type.  Thoughts? @bukepo

librasungirl: Agree that counting OoB into mTxOther is a simple solution.
@bukepo thoughts?

librasungirl: In #4459
> Describe alternatives you've considered
> count these frames as other frames because there's other rx frame counter.

If `rx` is a typo for `tx`, the alternative should be the simple solution proposed here? @bukepo @FiveDimensions

PR #4938, opened 2020-05-08
[style] fix unused variables in release build

bukepo:
* change package check build Release


jwhui:
@bukepo , thanks for this. Can we add a continuous-integration check to avoid this issue in the future?

bukepo:
It's enabled in package check

PR #4939, opened 2020-05-08
[tasklet] change Post() to return void

abtink:
This commit changes the `Tasklet::Post()` method to return `void`
instead of `otError`. This helps simplify its use. The error indicated
whether the tasklet was already posted and was always ignored
(required `IgnoreError()` when posting tasklets). `IsPosted()` method
can be used to determine this when/if needed.

### DIFF
--- a/src/core/coap/coap_secure.cpp
+++ b/src/core/coap/coap_secure.cpp
@@ -179,7 +179,7 @@ otError CoapSecure::Send(ot::Message &aMessage, const Ip6::MessageInfo &aMessage
     otError error;
 
     SuccessOrExit(error = mTransmitQueue.Enqueue(aMessage));
-    IgnoreError(mTransmitTask.Post());
+    mTransmitTask.Post();
 
 exit:
     return error;
@@ -236,7 +236,7 @@ void CoapSecure::HandleTransmit(void)
 
     if (mTransmitQueue.GetHead() != NULL)
     {
-        IgnoreError(mTransmitTask.Post());
+        mTransmitTask.Post();
     }
 
     SuccessOrExit(error = mDtls.Send(*message, message->GetLength()));
--- a/src/core/common/notifier.cpp
+++ b/src/core/common/notifier.cpp
@@ -124,7 +124,7 @@ void Notifier::Signal(otChangedFlags aFlags)
 {
     mFlagsToSignal |= aFlags;
     mSignaledFlags |= aFlags;
-    IgnoreError(mTask.Post());
+    mTask.Post();
 }
 
 void Notifier::SignalIfFirst(otChangedFlags aFlags)
--- a/src/core/common/tasklet.cpp
+++ b/src/core/common/tasklet.cpp
@@ -49,15 +49,12 @@ Tasklet::Tasklet(Instance &aInstance, Handler aHandler, void *aOwner)
 {
 }
 
-otError Tasklet::Post(void)
+void Tasklet::Post(void)
 {
-    otError error = OT_ERROR_NONE;
-
-    VerifyOrExit(!IsPosted(), error = OT_ERROR_ALREADY);
-    Get<TaskletScheduler>().PostTasklet(*this);
-
-exit:
-    return error;
+    if (!IsPosted())
+    {
+        Get<TaskletScheduler>().PostTasklet(*this);
+    }
 }
 
 TaskletScheduler::TaskletScheduler(void)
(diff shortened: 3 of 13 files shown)


jwhui:
> LGTM 👍
> We should try to avoid using IgnoreError everywhere else.

This would be nice. One challenge is with methods that are exposed via OpenThread API. While we can guarantee that core code calls methods appropriately, we cannot assume users do the same and need to propagate errors back to the users. One example is `MessageQueue::Enqueue()`. One possibility is to have two separate methods:
- `otError MessageQueue::EnqueueExternal()`
- `void MessageQueue::EnqueueInternal()`

Thoughts?

simonlingoogle:
We have many methods that might return `OT_ERROR_ALREADY`. 
For these methods, generally speaking, I tends to not return `OT_ERROR_ALREADY`, but rather return `OT_ERROR_NONE` to indicate a success. 
One reason is that handling `OT_ERORR_ALREADY` is error-prone. For example #4788 was trying to handle `OT_ERROR_ALREADY` the same like `OT_ERROR_NONE`, but it introduced a bug (despite being reviewed by 5 reviewers) which was later fixed by #4899 
Do you think we should generally avoid using `OT_ERROR_ALREADY`? Thoughts? @abtink @jwhui @wgtdkp

wgtdkp:
@jwhui 

>One challenge is with methods that are exposed via OpenThread API. While we can guarantee that core code calls methods appropriately, we cannot assume users do the same and need to propagate errors back to the users. One example is MessageQueue::Enqueue(). One possibility is to have two separate methods:
otError MessageQueue::EnqueueExternal()
void MessageQueue::EnqueueInternal()
But there are a lot of methods that we expose via user APIs.

Sounds good to me.

Another option is that we keep a single `otError MessageQueue::Enqueue(...)` and use `assert` on all internal usage.

wgtdkp:
@simonlingoogle 

> Do you think we should generally avoid using OT_ERROR_ALREADY? Thoughts? @abtink @jwhui @wgtdkp

Agree that in many cases, `OT_ERROR_ALREADY` doesn't indicate a failure but should be treated as `OT_ERROR_NONE`. An example would be `Coap::AddResource(...)`. Its semantic should be like a `map` and should always success (if we don't consider `out of memory` error). In such case, maybe we should return a boolean to indicate whether a new resource is inserted.

jwhui:
> @simonlingoogle
> 
> > Do you think we should generally avoid using OT_ERROR_ALREADY? Thoughts? @abtink @jwhui @wgtdkp
> 
> Agree that in many cases, `OT_ERROR_ALREADY` doesn't indicate a failure but should be treated as `OT_ERROR_NONE`. An example would be `Coap::AddResource(...)`. Its semantic should be like a `map` and should always success (if we don't consider `out of memory` error). In such case, maybe we should return a boolean to indicate whether a new resource is inserted.

In many cases, we can probably remove `OT_ERROR_ALREADY`. I'll take a look and see how far we can take it.

abtink:
I generally agree/favor that if we don't need/use a returned `otError` from a method it may be better to make the method `void` (in some cases maybe adding `OT_ASSERT` in the implementation instead). 
The `Tasklet::Post()` was the easy low-hanging fruit :)

----

> This would be nice. One challenge is with methods that are exposed via OpenThread API. While we can guarantee that core code calls methods appropriately, we cannot assume users do the same and need to propagate errors back to the users. One example is `MessageQueue::Enqueue()`. One possibility is to have two separate methods:
> 
> `otError MessageQueue::EnqueueExternal()`
> `void MessageQueue::EnqueueInternal()`
> But there are a lot of methods that we expose via user APIs.
> 
> Thoughts?

I agree and think it'd be better to keep OT API as before (i.e. keep the error in the public OT API). One way I can see is to add check for error case in `api` file, e.g., in this particular case:
- We can convert the `MessageQueue::Enqueue()` to `void` and use `OT_ASSERT()` in its implementation.
- We can then make `Message::IsInAQueue()` public and then use it from `otMessageQueueEnqueue()` to detect the error in `message_api.cpp`.
- The advantage is that single implementation internally, the downside I see is that the `api` code would have more logic in it, than a simple call to a matching method.

Thoughts?

----

On `OT_ERROR_ALREADY`, I agree removing it would be trickier, but would be good in some cases (also we can consider `NOT_FOUND` error for remove).

jwhui:
> On `OT_ERROR_ALREADY`, I agree removing it would be trickier, but would be good in some cases (also we can consider `NOT_FOUND` error for remove).

I just submitted #4942 that starts to address this.

PR #4940, opened 2020-05-08
[ip6] fix and enhance IPv6 fragmentation

LuDuda:
This PR fixes and enhances the handling of IPv6 fragmentation.

It consists of three commits:

**[ncp] ensure IPv6 packet length does not exceed maximum size of SPINEL command payload**

With the current code, it is possible that NCP tries to send IPv6 packet larger than the maximum size of SPINEL command payload. This results in the following error in the wpantund:

```
wpantund[31934]: SpinelNCPInstance-DataPump.cpp:239: Requirement Failed (mInboundFrameSize < sizeof(mInboundFrame))
```

This commit adds additional defines to spinel.h and proper checks in NCP codebase.

**[ip6] do not free the same message twice while handling IPv6 fragment**

This commit fixes the serious bug when IPv6 fragmentation is enabled. It is possible that `HandleDatagram` which is called on complete reassembly, returns the error (e.g. because ICMPv6 Echo Response can't be generated due to lack of buffers). In such case both `HandleDatagram` method and `HandleFragment` tries to free the same message which finally ends up in a crash of the system.

To reproduce it:
- Set up a network with two nodes A and B.
- Execute the following command on node A: `ping ML-EID_B 1950 1000 0.01`

**[ip6] enhance handling of IPv6 fragmentation**

This commit introduces a few enhancements:
- In NCP architecture forward the IPv6 fragments instead of the full reassembled IPv6 packet (which in the default configuration can't be forwarder due to SPINEL limitations). Without this fix the fragmentation does not work if Thread device wants to send more than MTU data through the Border Router (to the IPv6 address of NCP). 
- Use `OPENTHREAD_CONFIG_IP6_MAX_ASSEMBLED_DATAGRAM` only for the receiving path. For transmitting path, relay on the message buffers pool. This is especially important for some of the use-cases we have, and also reduce a need for regenerating OpenThread libraries.
- In case receiving IPv6 packet is bigger than supported `OPENTHREAD_CONFIG_IP6_MAX_ASSEMBLED_DATAGRAM`, exit before trying to allocate the next message buffer.

### DIFF
--- a/src/core/net/ip6.cpp
+++ b/src/core/net/ip6.cpp
@@ -646,8 +646,6 @@ otError Ip6::FragmentDatagram(Message &aMessage, uint8_t aIpProto)
         FragmentHeader::MakeDivisibleByEight(kMinimalMtu - aMessage.GetOffset() - sizeof(fragmentHeader));
     uint16_t payloadLeft = aMessage.GetLength() - aMessage.GetOffset();
 
-    VerifyOrExit(aMessage.GetLength() <= kMaxAssembledDatagramLength, error = OT_ERROR_NO_BUFS);
-
     VerifyOrExit(aMessage.Read(0, sizeof(header), &header) == sizeof(header), error = OT_ERROR_PARSE);
     header.SetNextHeader(kProtoFragment);
 
@@ -699,6 +697,7 @@ otError Ip6::FragmentDatagram(Message &aMessage, uint8_t aIpProto)
 
         otLogInfoIp6("Fragment %d with %d bytes sent", fragmentCnt, payloadFragment);
     }
+
     aMessage.Free();
 
 exit:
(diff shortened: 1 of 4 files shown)


LuDuda:
Jonathan, I see that `ncp_base_mtd.cpp` was not updated correctly. Let me send one more commit.

jwhui:
> Jonathan, I see that `ncp_base_mtd.cpp` was not updated correctly. Let me send one more commit.

@LuDuda, thanks! I updated the commits to incorporate your details and PR# as well.

LuDuda:
Okay, now the change is applied again. Just verified again that it works as it should. So it's ready to be reviewed and merged. Thank you for updating the commits, I will try to do that next time as well.

PR #4941, opened 2020-05-09
Change methods to return void when return value is not used

jwhui:
This PR contains a number of small commits that should make it easier to review.

More will come in subsequent PRs.

### DIFF
--- a/src/core/coap/coap.cpp
+++ b/src/core/coap/coap.cpp
@@ -375,7 +375,7 @@ void CoapBase::HandleRetransmissionTimer(void)
                 messageInfo.SetPeerPort(metadata.mDestinationPort);
                 messageInfo.SetSockAddr(metadata.mSourceAddress);
 
-                IgnoreError(SendCopy(*message, messageInfo));
+                SendCopy(*message, messageInfo);
             }
         }
 
@@ -465,7 +465,7 @@ void CoapBase::DequeueMessage(Message &aMessage)
     // the timer would just shoot earlier and then it'd be setup again.
 }
 
-otError CoapBase::SendCopy(const Message &aMessage, const Ip6::MessageInfo &aMessageInfo)
+void CoapBase::SendCopy(const Message &aMessage, const Ip6::MessageInfo &aMessageInfo)
 {
     otError  error;
     Message *messageCopy = NULL;
@@ -478,12 +478,15 @@ otError CoapBase::SendCopy(const Message &aMessage, const Ip6::MessageInfo &aMes
 
 exit:
 
-    if (error != OT_ERROR_NONE && messageCopy != NULL)
+    if (error != OT_ERROR_NONE)
     {
-        messageCopy->Free();
-    }
+        otLogWarnCoap("Failed to send copy: %s", otThreadErrorToString(error));
 
-    return error;
+        if (messageCopy != NULL)
+        {
+            messageCopy->Free();
+        }
+    }
 }
 
 Message *CoapBase::FindRelatedRequest(const Message &         aResponse,
--- a/src/core/meshcop/commissioner.cpp
+++ b/src/core/meshcop/commissioner.cpp
@@ -209,7 +209,7 @@ otError Commissioner::Stop(bool aResign)
 
     if (needResign && aResign)
     {
-        IgnoreError(SendKeepAlive());
+        SendKeepAlive();
     }
 
 exit:
@@ -451,7 +451,7 @@ void Commissioner::HandleTimer(void)
         break;
 
     case OT_COMMISSIONER_STATE_ACTIVE:
-        IgnoreError(SendKeepAlive());
+        SendKeepAlive();
         break;
     }
 }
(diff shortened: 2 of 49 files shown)


src/core/coap/coap.cpp
```diff
@@ -478,12 +478,15 @@ otError CoapBase::SendCopy(const Message &aMessage, const Ip6::MessageInfo &aMes
 
 exit:
 
-    if (error != OT_ERROR_NONE && messageCopy != NULL)
+    if (error != OT_ERROR_NONE)
     {
-        messageCopy->Free();
-    }
+        otLogNoteCoap("Failed to send copy: %s", otThreadErrorToString(error));
```
simonlingoogle, line 483: otLogWarnCoap ?

src/core/meshcop/commissioner.cpp
```diff
@@ -312,7 +316,7 @@ otError Commissioner::AddJoiner(const Mac::ExtAddress *aEui64, const char *aPskd
 
         UpdateJoinerExpirationTimer();
 
-        IgnoreError(SendCommissionerSet());
+        SendCommissionerSet();
```
simonlingoogle, line 319: If `SendCommissionerSet()` failed here, do we have any mechanism to assure that the Commissioner information is eventually synchronized to Leader?

src/core/thread/indirect_sender.cpp
```diff
@@ -84,23 +84,22 @@ void IndirectSender::Stop(void)
...
 }
 
-otError IndirectSender::AddMessageForSleepyChild(Message &aMessage, Child &aChild)
+void IndirectSender::AddMessageForSleepyChild(Message &aMessage, Child &aChild)
 {
-    otError  error = OT_ERROR_NONE;
     uint16_t childIndex;
 
-    VerifyOrExit(!aChild.IsRxOnWhenIdle(), error = OT_ERROR_INVALID_STATE);
+    VerifyOrExit(!aChild.IsRxOnWhenIdle(), OT_NOOP);
```
simonlingoogle, line 91:
```suggestion
    OT_ASSERT(!aChild.IsRxOnWhenIdle());
```

src/core/thread/mle_router.cpp
```diff
@@ -3264,8 +3252,6 @@ otError MleRouter::SendDataResponse(const Ip6::Address &aDestination,
     {
         message->Free();
```
simonlingoogle, line 3253: log fail here?

src/core/thread/mle_router.cpp
```diff
@@ -467,8 +469,6 @@ otError MleRouter::SendAdvertisement(void)
...
```
simonlingoogle, line 470: log error?

src/core/meshcop/commissioner.cpp
jwhui: Yes, that would require more changes and beyond the scope of this PR.

src/core/thread/mle_router.cpp
```diff
@@ -374,6 +374,12 @@ void MleRouter::SetStateLeader(uint16_t aRloc16)
     otLogNoteMle("Leader partition id 0x%x", mLeaderData.GetPartitionId());
 }
 
+void MleRouter::AddLeaderAloc(void)
+{
+    IgnoreError(GetLeaderAloc(mLeaderAloc.GetAddress()));
+    IgnoreError(Get<ThreadNetif>().AddUnicastAddress(mLeaderAloc));
+}
```
abtink, line 381: I verified that we don't need `VerifyOrExit(IsLeader())`
Noticed it is being used only from `SetStateLeader()`. Can we consider removing this method (inlining it)?

src/core/thread/mle_router.cpp
```diff
@@ -3260,12 +3253,15 @@ otError MleRouter::SendDataResponse(const Ip6::Address &aDestination,
 
 exit:
 
-    if (error != OT_ERROR_NONE && message != NULL)
+    if (error != OT_ERROR_NONE)
     {
-        message->Free();
-    }
+        otLogWarnMle("Failed to send Data Response: %s", otThreadErrorToString(error));
```
simonlingoogle, line 3258: Can we further reduce the code size by using same formatting string for all send failures?
```suggestion
        otLogWarnMle("Failed to send %s: %s", "Data Response", otThreadErrorToString(error));
```

src/core/thread/mle.cpp
```diff
@@ -2508,8 +2494,6 @@ otError Mle::SendAnnounce(uint8_t aChannel, bool aOrphanAnnounce, const Ip6::Add
     {
         message->Free();
```
simonlingoogle, line 2495: log failure? 

Would it be better if we use the same statement for both successful log and failure log, like:
```suggestion

if (error == OT_ERROR_NONE) {
  otLogInfoMle("Send Announce on channel %d: %s", aChannel, otThreadErrorToString(error));
} else {
  otLogWarnMle("Send Announce on channel %d: %s", aChannel, otThreadErrorToString(error));
}

        message->Free();
```
We can define a new MACRO for this purpose.

jwhui: We can consider this, but probably better for a separate PR that covers all instances that are not within scope of this PR.

simonlingoogle: Agree

src/core/thread/mle_router.cpp
jwhui: We can consider this, but probably better for a separate PR that covers all instances that are not within scope of this PR.

PR #4942, opened 2020-05-09
Avoid returning OT_ERROR_ALREADY and OT_ERROR_NOT_FOUND

jwhui:
See #4939 for context.

This PR contains a number of small commits that should make it easier to review.

More will come in subsequent PRs.

### DIFF
--- a/include/openthread/message.h
+++ b/include/openthread/message.h
@@ -321,11 +321,8 @@ void otMessageQueueInit(otMessageQueue *aQueue);
  * @param[in]  aQueue    A pointer to the message queue.
  * @param[in]  aMessage  The message to add.
  *
- * @retval OT_ERROR_NONE     Successfully added the message to the queue.
- * @retval OT_ERROR_ALREADY  The message is already enqueued in a queue.
- *
  */
-otError otMessageQueueEnqueue(otMessageQueue *aQueue, otMessage *aMessage);
+void otMessageQueueEnqueue(otMessageQueue *aQueue, otMessage *aMessage);
 
 /**
  * This function adds a message at the head/front of the given message queue.
@@ -333,11 +330,8 @@ otError otMessageQueueEnqueue(otMessageQueue *aQueue, otMessage *aMessage);
  * @param[in]  aQueue    A pointer to the message queue.
  * @param[in]  aMessage  The message to add.
  *
- * @retval OT_ERROR_NONE     Successfully added the message to the queue.
- * @retval OT_ERROR_ALREADY  The message is already enqueued in a queue.
- *
  */
-otError otMessageQueueEnqueueAtHead(otMessageQueue *aQueue, otMessage *aMessage);
+void otMessageQueueEnqueueAtHead(otMessageQueue *aQueue, otMessage *aMessage);
 
 /**
  * This function removes a message from the given message queue.
--- a/include/openthread/message.h
+++ b/include/openthread/message.h
@@ -339,11 +339,8 @@ void otMessageQueueEnqueueAtHead(otMessageQueue *aQueue, otMessage *aMessage);
  * @param[in]  aQueue    A pointer to the message queue.
  * @param[in]  aMessage  The message to remove.
  *
- * @retval OT_ERROR_NONE       Successfully removed the message from the queue.
- * @retval OT_ERROR_NOT_FOUND  The message is not enqueued in this queue.
- *
  */
-otError otMessageQueueDequeue(otMessageQueue *aQueue, otMessage *aMessage);
+void otMessageQueueDequeue(otMessageQueue *aQueue, otMessage *aMessage);
 
 /**
  * This function returns a pointer to the message at the head of the queue.
--- a/src/core/common/message.cpp
+++ b/src/core/common/message.cpp
@@ -372,7 +372,7 @@ otError Message::SetPriority(uint8_t aPriority)
     if (mBuffer.mHead.mInfo.mInPriorityQ)
     {
         priorityQueue = mBuffer.mHead.mInfo.mQueue.mPriority;
-        IgnoreError(priorityQueue->Dequeue(*this));
+        priorityQueue->Dequeue(*this);
     }
 
     mBuffer.mHead.mInfo.mPriority = aPriority;
@@ -832,12 +832,9 @@ void MessageQueue::Enqueue(Message &aMessage, QueuePosition aPosition)
     }
 }
 
-otError MessageQueue::Dequeue(Message &aMessage)
+void MessageQueue::Dequeue(Message &aMessage)
 {
-    otError error = OT_ERROR_NONE;
-
-    VerifyOrExit(aMessage.GetMessageQueue() == this, error = OT_ERROR_NOT_FOUND);
-
+    OT_ASSERT(aMessage.GetMessageQueue() == this);
     OT_ASSERT((aMessage.Next() != NULL) && (aMessage.Prev() != NULL));
 
     if (&aMessage == GetTail())
@@ -857,9 +854,6 @@ otError MessageQueue::Dequeue(Message &aMessage)
     aMessage.Next() = NULL;
 
     aMessage.SetMessageQueue(NULL);
-
-exit:
-    return error;
 }
 
 void MessageQueue::GetInfo(uint16_t &aMessageCount, uint16_t &aBufferCount) const
@@ -970,13 +964,12 @@ void PriorityQueue::Enqueue(Message &aMessage)
     mTails[priority] = &aMessage;
 }
 
-otError PriorityQueue::Dequeue(Message &aMessage)
+void PriorityQueue::Dequeue(Message &aMessage)
 {
-    otError  error = OT_ERROR_NONE;
     uint8_t  priority;
     Message *tail;
 
-    VerifyOrExit(aMessage.GetPriorityQueue() == this, error = OT_ERROR_NOT_FOUND);
+    OT_ASSERT(aMessage.GetPriorityQueue() == this);
 
     priority = aMessage.GetPriority();
 
@@ -1000,9 +993,6 @@ otError PriorityQueue::Dequeue(Message &aMessage)
     aMessage.Prev()         = NULL;
 
     aMessage.SetMessageQueue(NULL);
-
-exit:
-    return error;
 }
 
 void PriorityQueue::GetInfo(uint16_t &aMessageCount, uint16_t &aBufferCount) const
--- a/src/core/common/message.cpp
+++ b/src/core/common/message.cpp
@@ -379,7 +379,7 @@ otError Message::SetPriority(uint8_t aPriority)
 
     if (priorityQueue != NULL)
     {
-        IgnoreError(priorityQueue->Enqueue(*this));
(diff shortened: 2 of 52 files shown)


include/openthread/message.h
```diff
@@ -321,23 +321,17 @@ void otMessageQueueInit(otMessageQueue *aQueue);
  * @param[in]  aQueue    A pointer to the message queue.
  * @param[in]  aMessage  The message to add.
  *
- * @retval OT_ERROR_NONE     Successfully added the message to the queue.
- * @retval OT_ERROR_ALREADY  The message is already enqueued in a queue.
- *
  */
-otError otMessageQueueEnqueue(otMessageQueue *aQueue, otMessage *aMessage);
+void otMessageQueueEnqueue(otMessageQueue *aQueue, otMessage *aMessage);
```
abtink, line 325: Thoughts about the approach from https://github.com/openthread/openthread/pull/4939#issuecomment-626108171 to keep the public API same (and do error handling in the `message_api.cpp`)?

src/core/common/message.cpp
```diff
@@ -801,11 +801,9 @@ Message *MessageQueue::GetHead(void) const
     return (GetTail() == NULL) ? NULL : GetTail()->Next();
 }
 
-otError MessageQueue::Enqueue(Message &aMessage, QueuePosition aPosition)
+void MessageQueue::Enqueue(Message &aMessage, QueuePosition aPosition)
 {
-    otError error = OT_ERROR_NONE;
-
-    VerifyOrExit(!aMessage.IsInAQueue(), error = OT_ERROR_ALREADY);
+    VerifyOrExit(!aMessage.IsInAQueue(), OT_NOOP);
```
simonlingoogle, line 806:
```suggestion
    VerifyOrExit(aMessage.GetMessageQueue() != this, OT_NOOP);
```

If `aMessage` belongs to another `MessageQueue`, better to just let it fail in the `OT_ASSERT` below?

src/core/common/message.cpp
```diff
@@ -835,14 +833,12 @@ otError MessageQueue::Enqueue(Message &aMessage, QueuePosition aPosition)
...
+    return;
 }
 
-otError MessageQueue::Dequeue(Message &aMessage)
+void MessageQueue::Dequeue(Message &aMessage)
 {
-    otError error = OT_ERROR_NONE;
-
-    VerifyOrExit(aMessage.GetMessageQueue() == this, error = OT_ERROR_NOT_FOUND);
+    VerifyOrExit(aMessage.GetMessageQueue() == this, OT_NOOP);
```
simonlingoogle, line 841:
```suggestion
    VerifyOrExit(aMessage.IsInAQueue(), OT_NOOP);
    
    OT_ASSERT(aMessage.GetMessageQueue() == this);
```
If `aMessage` is in any queue but not the current message queue, better just let it fail in the `OT_ASSERT` below?

src/core/common/message.cpp
```diff
@@ -944,14 +940,13 @@ Message *PriorityQueue::GetTail(void) const
...
-otError PriorityQueue::Enqueue(Message &aMessage)
+void PriorityQueue::Enqueue(Message &aMessage)
 {
-    otError  error = OT_ERROR_NONE;
     uint8_t  priority;
     Message *tail;
     Message *next;
 
-    VerifyOrExit(!aMessage.IsInAQueue(), error = OT_ERROR_ALREADY);
+    VerifyOrExit(!aMessage.IsInAQueue(), OT_NOOP);
```
simonlingoogle, line 949:
```suggestion
    VerifyOrExit(aMessage.GetPriorityQueue() != this, OT_NOOP);

    OT_ASSERT(!aMessage.IsInAQueue());
```
Just let the cases that the message is no another queue to fail.

src/core/common/message.cpp
```diff
@@ -977,16 +972,15 @@ otError PriorityQueue::Enqueue(Message &aMessage)
...
 
-otError PriorityQueue::Dequeue(Message &aMessage)
+void PriorityQueue::Dequeue(Message &aMessage)
 {
-    otError  error = OT_ERROR_NONE;
     uint8_t  priority;
     Message *tail;
 
-    VerifyOrExit(aMessage.GetPriorityQueue() == this, error = OT_ERROR_NOT_FOUND);
+    VerifyOrExit(aMessage.GetPriorityQueue() == this, OT_NOOP);
```
simonlingoogle, line 983:
```suggestion
    VerifyOrExit(aMessage.IsInAQueue(), OT_NOOP);
    
    OT_ASSERT(aMessage.GetPriorityQueue() == this);
```
Just let the cases that the message is in another queue to fail.

src/core/meshcop/border_agent.cpp
```diff
@@ -751,10 +751,11 @@ void BorderAgent::ApplyMeshLocalPrefix(void)
 {
     VerifyOrExit(mState == OT_BORDER_AGENT_STATE_ACTIVE, OT_NOOP);
 
-    if (Get<ThreadNetif>().RemoveUnicastAddress(mCommissionerAloc) == OT_ERROR_NONE)
+    if (Get<ThreadNetif>().IsUnicastAddressAdded(mCommissionerAloc) == OT_ERROR_NONE)
```
simonlingoogle, line 754:
```suggestion
    if (Get<ThreadNetif>().IsUnicastAddressAdded(mCommissionerAloc))
```

src/core/net/netif.hpp
```diff
@@ -205,22 +205,27 @@ class Netif : public InstanceLocator, public LinkedListEntry<Netif>
...
+    void RemoveUnicastAddress(const NetifUnicastAddress &aAddress);
+
+    /**
+     * This method indicates whether a unicast address is added to the network interface.
+     *
+     * @param[in]  aAddress  A reference to the unicast address.
      *
      */
-    otError RemoveUnicastAddress(const NetifUnicastAddress &aAddress);
+    bool IsUnicastAddressAdded(const NetifUnicastAddress &aAddress) const
```
simonlingoogle, line 225:
```suggestion
    bool IsUnicastAddress(const NetifUnicastAddress &aAddress) const
```

src/core/net/netif.cpp
```diff
@@ -219,6 +217,7 @@ otError Netif::SubscribeAllRoutersMulticast(void)
     // Ensure that the `LinkLocalAll` was found on the list.
 
     OT_ASSERT(error == OT_ERROR_NONE);
+    OT_UNUSED_VARIABLE(error);
```
simonlingoogle, line 220: Not in this PR, but if we define `OT_ASSERT(x)` to `((void)(x))` when it's **turned off**, we can get rid of this `OT_UNUSED_VARIABLE(error)`.

include/openthread/message.h
jwhui: I see both sides. On one hand, it is desirable to keep public APIs stable. On the other hand, it was argued in https://github.com/openthread/openthread/pull/4939#issuecomment-626086485 that returning `OT_ERROR_ALREADY` leads to calling code handling errors incorrectly. Thoughts?

jwhui:
> LGTM 👍

@simonlingoogle , it seems apply your suggested changes causes [some v1_2 tests to fail](https://github.com/openthread/openthread/runs/659211493?check_suite_focus=true). Any ideas?

include/openthread/message.h
abtink: I fully agree that handling error in higher layer gets tricky (we needed to handle for unicast/multicast address add/remove). I do also prefer that in such cases that we change both the private and  public APIs. 

But for the `MessageQueueEnqueue` case, I think it may be worth treating it differently:
- In other cases, like addresses or coap resource, ..., there is a single container/queue where the entries are being added/removed, so if user tries to add/remove same thing again, it should be fine to report `ERROR_NONE`.
- But for `Message` it can be in a different queue than the one user is calling `Enqueue` on, and I think we should consider this pattern a bug (incorrect use).
- Forgetting to dequeue a message and then trying to add it to another/same queue is a pattern that I feel can happen by mistake.

I would suggest for internal use we explicitly require that the `Message` passed to `Enqueue` to NOT be in any other queue and `OT_ASSERT` if it is (even if it is already in the same queue). 

For external API, I say we either provide an error in this case, or we can apply/require the same rule (I would favor we apply same rule to keep things simpler).

Thoughts?

src/core/net/netif.cpp
abtink: @simonlingoogle I would like this (considered the same when changing `OT_ASSERT`) but realized that it may not work in all cases/toolchain. 

The  `(void)(x)` does not seem work on some tool-chains. That's why we have specific implementations of `OT_UNUSED_VARIABLE()` in `toolchain.h` which uses the address of the variable and I don't think we can apply the same trick to the `condition` check in `OT_ASSERT(condition)` (since it's not a variable).

We need a different approach/idea here :)

abtink: Maybe something like:
```
#define OT_ASSERT(cond) \
    do                  \
    {                   \
        if (!(cond))    \
        {               \
        }               \
    } while (0)  
```

simonlingoogle: I think your approach should be better. @abtink

simonlingoogle:
I see the Simulation workflow is passed now. It's fixed?

Simon Lin | Software Engineer, Nest Shanghai | simonlin@google.com | +86
13656630312

jwhui:
> > @simonlingoogle , it seems applying your suggested changes causes some v1_2 tests to fail. Any ideas?
> 
> 
> 
> I see the Simulation workflow is passed now. It's fixed? @jwhui 

Yes, it is fixed. Thanks.

simonlingoogle:
> > > @simonlingoogle , it seems applying your suggested changes causes some v1_2 tests to fail. Any ideas?
> > 
> > 
> > 
> > I see the Simulation workflow is passed now. It's fixed? @jwhui 
> 
> Yes, it is fixed. Thanks.

May I ask what was the issue?

jwhui:
> > Yes, it is fixed. Thanks.
> 
> May I ask what was the issue?

Seems to be a momentary issue in GitHub Actions. Rerunning the job passes now.

include/openthread/message.h
jwhui: I generally prefer to be more strict as well. I updated this PR as suggested. Please review.

src/core/common/message.cpp
```diff
@@ -975,18 +963,15 @@ otError PriorityQueue::Enqueue(Message &aMessage)
...
-otError PriorityQueue::Dequeue(Message &aMessage)
+void PriorityQueue::Dequeue(Message &aMessage)
 {
-    otError  error = OT_ERROR_NONE;
     uint8_t  priority;
     Message *tail;
 
-    VerifyOrExit(aMessage.GetPriorityQueue() == this, error = OT_ERROR_NOT_FOUND);
+    OT_ASSERT(aMessage.IsInAQueue());
+    OT_ASSERT(aMessage.GetPriorityQueue() == this);
```
simonlingoogle, line 974:
```suggestion
    OT_ASSERT(aMessage.GetPriorityQueue() == this);
```

PR #4944, opened 2020-05-10
[style] fix unused result

bukepo:
* add pacakge-clang to detect this kind of errors


bukepo:
The original error: https://travis-ci.org/github/bukepo/ot-br-posix/jobs/685194417#L627

bukepo:
use #4946 instead.

PR #4946, opened 2020-05-11
[continuous-integration] enable most features on build check

Moandor-y:
This PR enables as many features as possible when checking POSIX building to find more issues. Also fixes some compiling errors.

`-Wno-gnu-folding-constant` is added to mac compile flags because mac's `CMSG_SPACE` implementation has an issue triggering this warning.


.github/workflows/build.yml
```diff
@@ -181,6 +181,8 @@ jobs:
     env:
       CC: clang
       CXX: clang++
+      CFLAGS: -Wno-gnu-folding-constant
```
simonlingoogle, line 184: Do we need to apply this flag not only in GitHub Workflow, but also in normal cmake builds?

Moandor-y: Yes. I'm going to use [this fix](https://github.com/openthread/openthread/pull/4944/files#diff-c1d434eb0e2a6df91b54b43e9ac19fc4) which applies this flag to one line of code only.

src/posix/platform/udp.cpp
```diff
@@ -125,7 +126,7 @@ static otError transmitPacket(int aFd, uint8_t *aPayload, uint16_t aLength, cons
         memcpy(CMSG_DATA(cmsg), &hopLimit, sizeof(int));
 
         cmsg = CMSG_NXTHDR(&msg, cmsg);
-        controlLength += CMSG_SPACE(sizeof(int));
+        controlLength += OT_APPLE_IGNORE_GNU_FOLDING_CONSTANT(CMSG_SPACE(sizeof(int)));
```
bukepo, line 129: not necessary?

src/posix/platform/netif.cpp
```diff
@@ -1357,7 +1357,7 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
     otIcmp6SetEchoMode(aInstance, OT_ICMP6_ECHO_HANDLER_DISABLED);
     otIp6SetReceiveCallback(aInstance, processReceive, aInstance);
     otIp6SetAddressCallback(aInstance, processAddressChange, aInstance);
-    otSetStateChangedCallback(aInstance, processStateChange, aInstance);
+    IgnoreError(otSetStateChangedCallback(aInstance, processStateChange, aInstance));
```
simonlingoogle, line 1360: Use `VerifyOrDie` ?

PR #4949, opened 2020-05-12
[message] change SetOffset()/MoveOffset() to return void

jwhui:


src/core/common/message.cpp
```diff
@@ -308,31 +308,17 @@ uint8_t Message::GetBufferCount(void) const
...
-    return error;
 }
 
-otError Message::SetOffset(uint16_t aOffset)
+void Message::SetOffset(uint16_t aOffset)
 {
-    otError error = OT_ERROR_NONE;
-
     OT_ASSERT(aOffset <= GetLength());
-    VerifyOrExit(aOffset <= GetLength(), error = OT_ERROR_INVALID_ARGS);
```
abtink, line 330: Did we have both assert and `VerifyOrExit` in earlier code?

abtink: Seems like we did. :)

PR #4950, opened 2020-05-12
[posix] fix inconsistent function definitions issue

zhanglongxia:
The arguments of the function `otPlatDiagProcess` in posix radio drivers different from 
the one defined in the `diag.h`. It causes the diag module uses the weak definition of
the function `otPlatDiagProcess`. Then all platform defined diag commands are not
available. This PR fixes this issue.

### DIFF
--- a/src/posix/platform/radio.cpp
+++ b/src/posix/platform/radio.cpp
@@ -337,7 +337,11 @@ exit:
 #endif
 
 #if OPENTHREAD_CONFIG_DIAG_ENABLE
-otError otPlatDiagProcess(otInstance *aInstance, int argc, char *argv[], char *aOutput, size_t aOutputMaxLen)
+otError otPlatDiagProcess(otInstance *aInstance,
+                          uint8_t     aArgsLength,
+                          char *      aArgs[],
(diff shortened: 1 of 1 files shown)


src/posix/platform/radio.cpp
```diff
@@ -337,17 +337,21 @@ otError otPlatRadioGetCoexMetrics(otInstance *aInstance, otRadioCoexMetrics *aCo
...
+                          size_t      aOutputMaxLen)
 {
     // deliver the platform specific diags commands to radio only ncp.
     OT_UNUSED_VARIABLE(aInstance);
     char  cmd[OPENTHREAD_CONFIG_DIAG_CMD_LINE_BUFFER_SIZE] = {'\0'};
     char *cur                                              = cmd;
     char *end                                              = cmd + sizeof(cmd);
 
-    for (int index = 0; index < argc; index++)
+    for (int index = 0; index < aArgsLength; index++)
```
jwhui, line 352:
```suggestion
    for (uint8_t index = 0; index < aArgsLength; index++)
```

PR #4951, opened 2020-05-12
[nrf528xx] add `OT_UNUSED_VARIABLE` to avoid compiler warning

zhanglongxia:
When the macro `SPIS_AS_SERIAL_TRANSPORT` is enabled,
`UART_AS_SERIAL_TRANSPORT` and `USB_CDC_AS_SERIAL_TRANSPORT`
are disabled, no one uses the argument `aPseudoReset`. The compiler generates a warning.

### DIFF
--- a/examples/platforms/nrf528xx/src/transport/transport.c
+++ b/examples/platforms/nrf528xx/src/transport/transport.c
@@ -31,6 +31,7 @@
  *   This file implements the nrf5 platform transport initialization functions.
  *
  */
+#include <openthread/platform/toolchain.h>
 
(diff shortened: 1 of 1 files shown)


examples/platforms/nrf528xx/src/transport/transport.c
```diff
@@ -31,6 +31,7 @@
  *   This file implements the nrf5 platform transport initialization functions.
  *
  */
+#include <openthread/platform/toolchain.h>
```
jwhui, line 34:
```suggestion

#include <openthread/platform/toolchain.h>
```

PR #4952, opened 2020-05-12
[posix] add platform udp send retry

gjc13:
The kernel will report EINVAL when we've just added/removed addresses
on the interface. Add retry to be tolerate.

### DIFF
--- a/src/posix/platform/udp.cpp
+++ b/src/posix/platform/udp.cpp
@@ -155,17 +155,17 @@ static otError transmitPacket(int aFd, uint8_t *aPayload, uint16_t aLength, cons
     for (size_t i = 0; i < OPENTHREAD_CONFIG_UDP_SEND_RETRY; i++)
     {
         rval = sendmsg(aFd, &msg, 0);
-        if (rval > 0 || errno != EINVAL)
+        if (rval != -1 || errno != EINVAL)
         {
             break;
         }
-        otLogWarnPlat("sendmsg(), retry: %s", strerror(errno));
+        otLogWarnPlat("sendmsg(): %s", strerror(errno));
         sleep(0);
     }
-    VerifyOrExit(rval > 0, perror("sendmsg"));
+    VerifyOrExit(rval != -1, perror("sendmsg"));
 
 exit:
-    return rval > 0 ? OT_ERROR_NONE : OT_ERROR_FAILED;
+    return rval != -1 ? OT_ERROR_NONE : OT_ERROR_FAILED;
 }
 
 static otError receivePacket(int aFd, uint8_t *aPayload, uint16_t &aLength, otMessageInfo &aMessageInfo)
--- a/src/posix/platform/udp.cpp
+++ b/src/posix/platform/udp.cpp
@@ -152,8 +152,16 @@ static otError transmitPacket(int aFd, uint8_t *aPayload, uint16_t aLength, cons
 #else
     msg.msg_controllen = controlLength;
 #endif
-
-    rval = sendmsg(aFd, &msg, 0);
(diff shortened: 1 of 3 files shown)


src/posix/platform/udp.cpp
```diff
@@ -152,8 +152,16 @@ static otError transmitPacket(int aFd, uint8_t *aPayload, uint16_t aLength, cons
...
-    rval = sendmsg(aFd, &msg, 0);
+    for (size_t i = 0; i < OPENTHREAD_CONFIG_PLATFORM_UDP_SEND_RETRY; i++)
+    {
+        rval = sendmsg(aFd, &msg, 0);
+        if (rval > 0)
+        {
+            break;
+        }
+        otLogWarnPlat("sendmsg failed, retry: %s", strerror(errno));
+        sleep(0);
```
wgtdkp, line 163: Is this necessary? Why would it make difference to retry multiple times continuously?

src/core/config/platform.h
```diff
@@ -87,6 +87,16 @@
...
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_PLATFORM_UDP_SEND_RETRY
+ *
+ * Number of times the stack will retry to send udp packets via kernel.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_PLATFORM_UDP_SEND_RETRY
+#define OPENTHREAD_CONFIG_PLATFORM_UDP_SEND_RETRY 3
```
bukepo, line 97: Move to openthread-posix-config.h
```suggestion
#define OPENTHREAD_POSIX_CONFIG_UDP_SEND_RETRY 3
```

src/posix/platform/udp.cpp
```diff
@@ -152,8 +152,16 @@ static otError transmitPacket(int aFd, uint8_t *aPayload, uint16_t aLength, cons
 #else
     msg.msg_controllen = controlLength;
 #endif
-
-    rval = sendmsg(aFd, &msg, 0);
+    for (size_t i = 0; i < OPENTHREAD_CONFIG_PLATFORM_UDP_SEND_RETRY; i++)
+    {
+        rval = sendmsg(aFd, &msg, 0);
+        if (rval > 0)
```
bukepo, line 158:
```suggestion
        if (rval > 0 || rval != EINVAL)
```

src/posix/platform/udp.cpp
```diff
@@ -152,8 +152,16 @@ static otError transmitPacket(int aFd, uint8_t *aPayload, uint16_t aLength, cons
...
+        {
+            break;
+        }
+        otLogWarnPlat("sendmsg failed, retry: %s", strerror(errno));
```
bukepo, line 162: because this may be the last attempt.
```suggestion
        otLogWarnPlat("sendmsg(): %s", strerror(errno));
```

src/posix/platform/udp.cpp
gjc13: Looks like the kernel becomes unavailable for sending for some time. So we yield and retry.

wgtdkp: But why `sleep(0)`? I am also curious why would it gives `EINVAL`?

bukepo: I think this is caused by adding address is not taking effect immediatelly.

src/posix/platform/udp.cpp
bukepo: Why is this resolved?

gjc13:
Looks like we should throw the error here and handle OT_ERROR_FAILED in the callers. Abandon.

PR #4953, opened 2020-05-12
[posix] ot-cli be indepedent of openthread-ncp-ftd

bukepo:


tests/toranj/openthread-core-toranj-config-posix.h
```diff
@@ -63,4 +63,12 @@
 
 #define OPENTHREAD_POSIX_CONFIG_RCP_BUS OT_POSIX_RCP_BUS_UART
 
+/**
+ * @def OPENTHREAD_CONFIG_LOG_OUTPUT
+ *
+ * Selects if, and where the LOG output goes to.
+ *
+ */
+#define OPENTHREAD_CONFIG_LOG_OUTPUT OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED
```
abtink, line 72: We intentionally pass all logs through NCP to make it simple to collect all logs (OT and `wpantund`) per simulated device during test run. This is super helpful for debugging issues (all logs are at same file per device).

I am not sure with this change, we will get the same behavior. Let me try it out by running on local machine to see how it would work.

abtink: Yes. It's as I thought/expected. With this change the OT logs from all simulated nodes will go to syslog and won't be collected as part of device logs in `toranj` (only the wpantund logs are saved in the log file). 

I think we should keep the log output as is (to NCP) for the simulations. 

I understand/guess that you changed this for `ot-cli` build (where there is no ncp and therefore no `otPlatLog`). I'd say we need a different solution here.
- Can we add a new file with the implementation `otPlatLog` which we include for the `ot-cli` build only?
- Another approach is to add a weak implementation of `otPlatLog` somewhere (though I like the specific ot-cli one better).

bukepo: Thanks for the proposals.

I think using spinel log is for NCP(ot-ncp). And toranj doesn't use ot-cli, maybe we disable building CLI in toranj.

Besides, since wpantund is also using syslog, and I think the syslog can keep order of logs better, maybe we can also consider changing the toranj tests to collect logs from syslog.

I will try just building NCP in toranj first.

abtink: > I will try just building NCP in toranj first.

Sounds ok to me.

But in general, I think we need a solution on how to build `ot-cli` if/when project config sets the log output to `NCP` independent of toranj case itself. I feel with such a config  the logs in `ot-cli` should be disabled (since there is no NCP/spinel to use for output). Thoughts?

> Besides, since wpantund is also using syslog, and I think the syslog can keep order of logs better, maybe we can also consider changing the toranj tests to collect logs from syslog.

`toranj` collects the logs from `stderr` of `wpantund` process (`wpantund` configure syslog to emit logs to `stderr` as well). Changing `toranj` to collect logs directly from syslog output is probably possible but would be much larger change (not sure if it is worth it).

bukepo: I agree. I think there is a APP log output, which goes to CLI in CLI build, and spinel in NCP build. Maybe we can remove NCP_SPINEL output definition, and make it the APP behavior in NCP mode.

abtink: Good idea. SGTM. 👍

PR #4954, opened 2020-05-12
[test] add expect tests for extaddr, ping, promiscuous

Moandor-y:
This adds some `expect` test cases for CLI commands `extaddr`, `ping`, `promiscuous`.


tests/scripts/expect/cli-extaddr.exp
```diff
@@ -0,0 +1,40 @@
...
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
+spawn $env(OT_COMMAND) 1
+set timeout 1
+expect_after {
+    timeout { exit 1 }
+}
+send "extaddr 99aabbccddeeff00\n"
```
simonlingoogle, line 35: add more cases?
- extaddr startswith '0x'
- invalid extaddr
- extaddr shorter or longer than expected

Moandor-y: The CLI doesn't seem able to handle hex string starts with `0x`, is it intended behavior?

jwhui:
This is great. Just curious - why focus on this set of commands?

Moandor-y:
> This is great. Just curious - why focus on this set of commands?

These commands are not covered in codecov reports. There are many other commands to be added, just did the 3 first. Most codes in `cli.cpp` shall be covered finally.

tests/scripts/expect/cli-extaddr.exp
Moandor-y: Found that shorter `extaddr` would cause uninitialized memory to be used. Fixed by `memset`.

tests/scripts/expect/cli-extaddr.exp
```diff
@@ -0,0 +1,50 @@
...
+expect "Done\r\n"
+send "extaddr 99AABBCCDDEEFF00\n"
+expect "Done\r\n"
+send "extaddr\n"
+expect "99aabbccddeeff00\r\n"
+send "extaddr 1\n"
+expect "Done\r\n"
+send "extaddr\n"
+expect "0100000000000000\r\n"
```
jwhui, line 44: In response to setting "1", should we expect something like "0000000000000001" instead? Alternatively, should we require user input to include all 16 hex characters?

simonlingoogle: `extaddr 1`, if allowed, should definitely be `extaddr 0000000000000001` . 
I would prefer being more strict **allowing only 16 hex characters** to make OT CLI simpler.

Moandor-y: Agree. Changed to require the user to input exactly 16 characters.

Moandor-y:
Added a `ping` test case having 2 nodes making inter-node pings.

tests/scripts/expect/cli-ping.exp
```diff
@@ -0,0 +1,127 @@
...
+spawn $env(OT_COMMAND) 2
+set spawn_2 $spawn_id
+expect_after {
+    timeout { exit 1 }
+}
+
+set spawn_id $spawn_1
+send "panid 0x1234\n"
+expect "Done\r\n"
+send "masterkey 00112233445566778899aabbccddeeff\n"
```
jwhui, line 59: How about using the preferred network setup flow as described in [src/cli/README_COMMISSIONING.md](https://github.com/openthread/openthread/blob/master/src/cli/README_COMMISSIONING.md)?

src/cli/cli.cpp
```diff
@@ -1364,7 +1364,10 @@ void Interpreter::ProcessExtAddress(uint8_t aArgsLength, char *aArgs[])
     {
         otExtAddress extAddress;
 
-        VerifyOrExit(Hex2Bin(aArgs[0], extAddress.m8, sizeof(otExtAddress)) >= 0, error = OT_ERROR_INVALID_ARGS);
+        memset(&extAddress, 0, sizeof(extAddress));
+        VerifyOrExit(strlen(aArgs[0]) == 2 * sizeof(extAddress.m8), error = OT_ERROR_INVALID_ARGS);
```
librasungirl, line 1368: is this necessary?
fyi, there is length check in Hex2Bin
https://github.com/openthread/openthread/blob/72109da6f1a0bdc1ec91e0281f9ac53269a5e978/src/cli/cli.cpp#L296-L299

Moandor-y: `aAllowTruncate` is checking if the input string is longer than needed, but doesn't prevent shorter strings.
Also, checking the return value of `Hex2Bin` isn't sufficient as it allows odd lengths of input strings.
Is it better to change `Hex2Bin` to match exactly `hexLength == 2 * aBinLength` if `!aAllowTruncate`?

tests/scripts/expect/cli-ping.exp
Moandor-y: Updated to commissioning setup flow.

src/cli/cli.cpp
librasungirl: The implementation of Hex2Bin is allowing the first preceding 0 skipped.  Do you want to forbid such usage?

Moandor-y: Yes. If we don't allow truncation, then truncation on neither the input string nor the output binary should be allowed.
It's confusing if the conversion is done left-to-right but allows left-most 0's. E.g. `1234567` -> `01234567` but `12345` -> `01234500`.

librasungirl: `aAllowTruncate` was introduced purely for parsing a long input string with a buffer of 50 bytes in #4241 without sacrificing the flexibility of skipping first preceding 0 character.

@Moandor-y ,  when will `12345->01234500` happen? Do you mean the buffer (reset  with 0) is longer than the what the input string could be converted? If so, I think the user should check the return value of Hex2Bin to see how many bytes are converted from the input string when using the buffer.

Moandor-y: Updated to leave `Hex2Bin` untouched and made all callers to check the return value match exactly the binary length. This should prevent the uninitialized memory error and retain the flexibility.

PR #4955, opened 2020-05-12
[diag] check RCP diag commands

bukepo:


src/core/diags/factory_diags.cpp
```diff
@@ -557,6 +557,14 @@ otError Diags::ProcessCmd(uint8_t aArgsLength, char *aArgs[], char *aOutput, siz
 {
     otError error = OT_ERROR_NONE;
 
+#ifndef NDEBUG
```
abtink, line 560: Looks good to me overall.
@bukepo , can you please explain the reason for `#ifndef NDEBUG`? Thanks.

bukepo: This code is just for testing or debug purpose, so guard this with `#ifndef NDEBUG` to remove it from release build.

bukepo:
This is a follow up of #4950

PR #4961, opened 2020-05-14
[codecov] only comment after a number of builds

simonlingoogle:
Immature comments from Codecov bot shows big reduce in code coverage and is confusing. 
This PR prevent Codecov comment until a number of builds are finished. 

However the number of `after_n_builds` is hardcoded and has to be revised whenever we add a new `codecov/codecov-action@v1` step. 

Do you think it's better to have `after_n_builds` defined? @wgtdkp @bukepo @jwhui

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -9,3 +9,4 @@ ignore:
 
 comment:
   layout: "diff, flags, files"
+  after_n_builds: 15


jwhui:
@simonlingoogle, thanks for addressing this.

Rather than defining a number of builds (which you pointed out may need to be modified as we add/remove builds), have you looked into one of the following?

- [require_ci_to_pass](https://docs.codecov.io/docs/codecovyml-reference#codecovrequire_ci_to_pass)
- [notify.wait_for_ci](https://docs.codecov.io/docs/codecovyml-reference#codecovnotifywait_for_ci)

simonlingoogle:
Did some investigation:

> * [notify.wait_for_ci](https://docs.codecov.io/docs/codecovyml-reference#codecovnotifywait_for_ci)

This option is only for the [`notification` feature](https://docs.codecov.io/v4.3.0/docs/notifications), which supports Gitter, Hipchat, IRC, Slack, but has nothing to do with commenting on PR. 

> * [require_ci_to_pass](https://docs.codecov.io/docs/codecovyml-reference#codecovrequire_ci_to_pass)

Also not work as we wanted. This option is actually `yes` by default. I think this is also for the notification only. 

So, for now, `comment.after_n_builds` is the only option that works. @jwhui

PR #4963, opened 2020-05-14
[posix] add pseudo-reset

FiveDimensions:
Add posix pseudo reset

### DIFF
--- a/src/posix/main.c
+++ b/src/posix/main.c
@@ -386,11 +386,9 @@ pseudo_reset:
 #endif
 #endif
 
-    while (!otSysPseudoResetWasRequested())
+    while (true)
     {
         otSysMainloopContext mainloop;
-
-        otTaskletsProcess(instance);
 
         FD_ZERO(&mainloop.mReadFdSet);
         FD_ZERO(&mainloop.mWriteFdSet);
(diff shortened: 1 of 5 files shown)


src/posix/platform/system.cpp
```diff
@@ -42,12 +42,18 @@
...
 
 otInstance *otSysInit(otPlatformConfig *aPlatformConfig)
 {
     otInstance *instance = NULL;
 
+    if (gPlatformPseudoResetWasRequested)
+    {
+        gPlatformPseudoResetWasRequested = false;
+    }
+
```
simonlingoogle, line 56:
```suggestion
    gPlatformPseudoResetWasRequested = false;

```

src/posix/main.c
```diff
@@ -392,7 +385,7 @@ int main(int argc, char *argv[])
 #endif
 #endif
 
-    while (true)
+    while (!otSysPseudoResetWasRequested())
```
bukepo, line 388: can we make otSysMainloopPoll() return `-2` to indicate requesting pseudo-reset? Thus we don't need a new API.

FiveDimensions: ok

bukepo:
ping~

simonlingoogle:
> ping~

Could any one answer why we need pseudo-reset on posix ?

jwhui:
@FiveDimensions , any plans to move this PR forward?

PR #4965, opened 2020-05-14
[posix] allow bind to Thread interface only

bukepo:
This commit adds an option to UdpSocket::Bind to allow it bind to Thread interface only.


src/posix/platform/udp.cpp
```diff
@@ -275,6 +276,19 @@ otError otPlatUdpBind(otUdpSocket *aUdpSocket)
...
+    if (!aAllInterfaces)
+    {
+#if __linux__
+        struct ifreq interface;
+
+        strncpy(interface.ifr_ifrn.ifrn_name, sPlatNetifName, IFNAMSIZ);
+        VerifyOrExit(setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, &interface, sizeof(interface)) == 0,
+                     error = OT_ERROR_FAILED);
+#else // __NetBSD__ || __FreeBSD__ || __APPLE__
+        setsockopt(fd, IPPROTO_IP, IP_BOUND_IF, &sPlatNetifIndex, sizeof(sPlatNetifIndex))
```
bukepo, line 288:
```suggestion
        setsockopt(fd, IPPROTO_IP, IP_BOUND_IF, &sPlatNetifIndex, sizeof(sPlatNetifIndex));
```

simonlingoogle:
> This commit adds an option to UdpSocket::Bind to allow it bind to Thread interface only.

I did the same thing in #5376 . Should we always binds to Thread interface only? Why was this PR closed previously?

src/core/net/udp6.cpp
```diff
@@ -202,7 +207,7 @@ otError Udp::Connect(SocketHandle &aSocket, const SockAddr &aSockAddr)
 
     if (!aSocket.IsBound())
     {
-        SuccessOrExit(error = Bind(aSocket, aSocket.GetSockName()));
+        SuccessOrExit(error = Bind(aSocket, aSocket.GetSockName(), false));
```
simonlingoogle, line 210:
```suggestion
        SuccessOrExit(error = Bind(aSocket, aSocket.GetSockName(), true));
```
Should bind to Thread only by default?

src/posix/platform/udp.cpp
```diff
@@ -283,6 +284,19 @@ otError otPlatUdpBind(otUdpSocket *aUdpSocket)
     VerifyOrExit(0 == setsockopt(fd, IPPROTO_IPV6, IPV6_MULTICAST_IF, &sPlatNetifIndex, sizeof(sPlatNetifIndex)),
                  error = OT_ERROR_FAILED);
 
+    if (aThreadOnly)
+    {
+#if __linux__
+        struct ifreq interface;
+
+        strncpy(interface.ifr_ifrn.ifrn_name, sPlatNetifName, IFNAMSIZ);
+        VerifyOrExit(setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, &interface, sizeof(interface)) == 0,
```
simonlingoogle, line 293: We should do it before calling `bind`?

src/posix/platform/udp.cpp
```diff
@@ -283,6 +284,19 @@ otError otPlatUdpBind(otUdpSocket *aUdpSocket)
     VerifyOrExit(0 == setsockopt(fd, IPPROTO_IPV6, IPV6_MULTICAST_IF, &sPlatNetifIndex, sizeof(sPlatNetifIndex)),
                  error = OT_ERROR_FAILED);
 
+    if (aThreadOnly)
```
simonlingoogle, line 287: if `aThreadOnly = false` , should we use `SO_REUSEPORT`?

src/posix/platform/udp.cpp
```diff
@@ -408,6 +422,8 @@ void platformUdpInit(const char *aIfName)
     {
         perror("if_nametoindex");
     }
+
+    sPlatNetifName = aIfName;
```
simonlingoogle, line 426: Platform netif name is used or going to be used by a number of platform modules (e.g. udp, uart).
But saving the pointer to the Netif name isn't the safest way. The pointer `aIfName` does point to static memory in the current implementation, but not guaranteed so. 
I added a posix API `platformGetNetifName()` in #5376 so that `udp` and `uart` can use this interface to get the Netif name. Do you think it's a better way? @bukepo

src/posix/platform/udp.cpp
```diff
@@ -320,7 +334,7 @@ otError otPlatUdpConnect(otUdpSocket *aUdpSocket)
         // We create new socket to disconnect.
         SuccessOrExit(error = otPlatUdpClose(aUdpSocket));
         SuccessOrExit(error = otPlatUdpSocket(aUdpSocket));
-        SuccessOrExit(error = otPlatUdpBind(aUdpSocket));
+        SuccessOrExit(error = otPlatUdpBind(aUdpSocket, false));
```
simonlingoogle, line 337:
```suggestion
        SuccessOrExit(error = otPlatUdpBind(aUdpSocket, ???));
```
How do we know this socket was previous binded to Thread only or not?

src/posix/platform/udp.cpp
```diff
@@ -283,6 +284,19 @@ otError otPlatUdpBind(otUdpSocket *aUdpSocket)
...
+    if (aThreadOnly)
+    {
+#if __linux__
+        struct ifreq interface;
+
+        strncpy(interface.ifr_ifrn.ifrn_name, sPlatNetifName, IFNAMSIZ);
+        VerifyOrExit(setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, &interface, sizeof(interface)) == 0,
+                     error = OT_ERROR_FAILED);
+#else  // __NetBSD__ || __FreeBSD__ || __APPLE__
+        setsockopt(fd, IPPROTO_IP, IP_BOUND_IF, &sPlatNetifIndex, sizeof(sPlatNetifIndex))
```
simonlingoogle, line 296: use VerifyOrExit ?

src/core/thread/mle.cpp
```diff
@@ -165,7 +165,7 @@ otError Mle::Enable(void)
 
     UpdateLinkLocalAddress();
     SuccessOrExit(error = mSocket.Open(&Mle::HandleUdpReceive, this));
-    SuccessOrExit(error = mSocket.Bind(kUdpPort));
+    SuccessOrExit(error = mSocket.Bind(kUdpPort, true));
```
abtink, line 168: We seem to already filter based on MLE port number in `Bind()` implementation.
```
     else if (!IsMlePort(aSocket.mSockName.mPort))
     {
         error = otPlatUdpBind(&aSocket);
```
 So do we even need this?

-------

In general, I am not sure if it is a good idea to add `bool aThreadOnly` falg to all `Bind()` methods, mainly since it seems to be only applicable when `OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE` is enabled.

I think by default, from the perspective of OT core and users, there should only one netif and it is the thread netif and OT sockets are tied to that interface only.

bukepo: No, it's not necessary here. However, I think it still meaningful to mark this should be a service on Thread interface only. This change is mainly for allowing running multiple OpenThread instances on a single host. Both TMF and MLE should run on Thread interface only. But border agent should run on all interfaces. The backbone agent should run on the backbone link interface(s) only.

I think we might need revert https://github.com/openthread/openthread/pull/5316, so that an OpenThread service can be bound to any host interface.

bukepo:
> I am trying to understand this issue more. My general question would be "When do we need to bind to all interfaces rather than Thread only?".
> Currently we have these sockets in OpenThread:
> 
> 1. MLE
> 2. COAP
> 3. DTLS
> 4. JOINER ROUTER
> 5. DHCP6 SERVER
> 6. DHCP6 CLIENT
> 7. DNS CLIENT
> 8. SNTP CLIENT
> 
> What does it mean for multiple OT instances to bind these Udp Sockets to all interface at the same time. I think only one will succeed.

The border agent should run on all network interfaces.

src/core/thread/mle.cpp
simonlingoogle: BBR features will needs to use sockets that bind to all interfaces or backbone interface only. So, I think it's useful to use flags to indicate which interface to bind to.

simonlingoogle:
> > I am trying to understand this issue more. My general question would be "When do we need to bind to all interfaces rather than Thread only?".
> > Currently we have these sockets in OpenThread:
> > 
> > 1. MLE
> > 2. COAP
> > 3. DTLS
> > 4. JOINER ROUTER
> > 5. DHCP6 SERVER
> > 6. DHCP6 CLIENT
> > 7. DNS CLIENT
> > 8. SNTP CLIENT
> > 
> > What does it mean for multiple OT instances to bind these Udp Sockets to all interface at the same time. I think only one will succeed.
> 
> The border agent should run on all network interfaces.

Seems most sockets should bind to Thread only. I think it's safer to use `Thread only` by default. For those sockets that purposely binds to all interface, they can explicitly specify so using `aThreadOnly = false`.

src/core/thread/mle.cpp
abtink: Here is how I see this:
- An OpenThread instance itself provides a "thread" netif to the user.
- All the socket APIs on an OT instance like `Socket::Bind()` should be tied to the provided "thread" netif only.
- I think OT core (representing an OT instance) should not be in general aware of any other netif (that is provided on device by host or other  instances of OT) and I think OT core should not be involved in managing (e.g., adding addresses, etc) on any other netif.
- So having a flag `aThreadOnly` in `Socket::Bind()` to indicate to bind  to thread netif vs all netifs inherently assumes an OT instance is aware of (and managing/binding to) more than one netif.

We have `OT_CONFIG_PLATFORM_UDP_ENABLE` which allow platforms to instead provide and handle UDP related socket APIs.
- Only in this context (when this feature is enabled), we sort of know of two netifs (one internal to OT core and one indirectly provided by platform UDP APIs).
- So only in this context, the `aThreadOnly` flags have any meaning and even in this case there seems to be no change in behavior in OT core code (related to `aThreadOnly` flag) and the flags is only passed down to platform UDP APIs.

I understand the use-case and reason what this PR is aiming to do, but I'd like to suggest that we try to add in a way that it is limited/used to when `OT_CONFIG_PLATFORM_UDP_ENABLE` feature is enabled.
- For example, as I pointed we already have in code logic that does filter MLE message based on the port number
```
     else if (!IsMlePort(aSocket.mSockName.mPort))
     {
         error = otPlatUdpBind(&aSocket);
```
- Instead of adding `aThreadOnly` to different OT APIs, can we consider adding additional/similar checks like the one for MLE which are scoped under `#if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE` and take care of the special cases.

Thoughts?

bukepo: The `aThreadOnly` is not added as an internal argument which is not exposed via OT APIs.

Here's the use cases, it seems the core still need to know existence of other network interfaces.
* The backbone agent need to know the index of the backbone link interface.
* The Thread border agent should be listening on all network interfaces(Wi-Fi, Ethernet and so on) so that external commissioners can reach the agent through all PHYs.

Actually, in core/net/ip6, whether a datagram comes from the Thread network interface is also checked([code](https://github.com/openthread/openthread/blob/master/src/core/net/ip6.cpp#L1161)).

Besides, one motivation of platform UDP was to allow developers develop their apps/services directly on OpenThread if they find OpenThread's infrastructure is easier to write the code(For example, OpenThread has a good CoAP implementation) and the apps/services may be operating on other network interfaces just like the Thread border agent. Another motivation is trying to integrate OpenThread's IP stack with the host's, so that if OpenThread has a socket bound to 1000, the host should not bind to that port again and vice versa.

Thoughts?

simonlingoogle: We do have the need to bind the the Backbone link so I think it is proper to have the argument for `Socket::Bind()` to choose which interface to bind to. 
But I also agree that "OT core (representing an OT instance) should not be in general aware of any other netif". 
That's why I suggested to use default argument `bool aThreadOnly = true` for the `Socket::Bind`.

simonlingoogle:
Any further thoughts for the comments of this PR? @bukepo @abtink 
MLR/DUA Backbone related features are depending on how this PR is resolved.
Thanks

abtink:
> Any further thoughts for the comments of this PR? @bukepo @abtink
> MLR/DUA Backbone related features are depending on how this PR is resolved.
> Thanks

I still prefer if we can make it work such that we do not change the `Bind()` in general situation, and instead add the related code  under `#if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE` scope (like MLE example I mentioned above) . We only use the `aThreadOnly` flag when this feature is enabled.  If we need dynamic/run-time check we can add methods in other modules to call from inside `#if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE` scope to decide the flag value.  I feel this would be cleaner model.

---

If we do decide to go for the `aThreadFlag` parameter model, I would suggest we use  `enum` definition (like `kThreadInterfaceOnly`, etc)  rather a `bool` (it helps make the use/caller side more readable).

simonlingoogle:
> > Any further thoughts for the comments of this PR? @bukepo @abtink
> > MLR/DUA Backbone related features are depending on how this PR is resolved.
> > Thanks
> 
> I still prefer if we can make it work such that we do not change the `Bind()` in general situation, and instead add the related code under `#if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE` scope (like MLE example I mentioned above) . We only use the `aThreadOnly` flag when this feature is enabled. If we need dynamic/run-time check we can add methods in other modules to call from inside `#if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE` scope to decide the flag value. I feel this would be cleaner model.
> 
> If we do decide to go for the `aThreadFlag` parameter model, I would suggest we use `enum` definition (like `kThreadInterfaceOnly`, etc) rather a `bool` (it helps make the use/caller side more readable).

Some thought for the current implementation:

**For Udp Foward:**
`Udp Foward` forwards the message to Host if the message has a `IsHostInterface` flag. So, conceptually speaking, UDP sockets are always binded to `All Interfaces` in `Udp Forward` mode. 

**For Platform Udp:**
- MLE UDP are actually not bounded to any Host interface. 
- Other UDP are bounded to all Host interfaces

So, I think the **Socket binding strategy** is a valid option for UDP binding, and there are actually more options we have used implicitly than we have defined:

1. Bind to Threadk stack **(Note that Thread stack is not Host Thread interface, it's OpenThread itself)**
2. Bind to all Host interfaces
3. Bind to one Host interface (Thread / Backbone / other)

1# is useful for MLE and TMF sockets that are supposed to only send diagrams in the Thread network and should not interact with Host interfaces. 
2# is useful for otbr-agent and 3# is useful to support multiple ot-daemon instances or multiple backbone interfaces. 

Currently, `Udp Forward` actually implicitly binds all sockets to both `Thread stack` and `all Host interfaces`. 
`Platform Udp` binds MLE socket to Thread stack by hardcoding, and other sockets to all Host interfaces. 

So, I think it's better and more general for us to call out these binding strategies that we have actually used and both `Udp Forward` and `Platform UDP` implementations should respect the binding strategy of a socket. 

- `Platform Udp` should bind MLE socket to `Thread stack`, instead of by hardcoding
- `Udp Forward` should respect the binding of sockets and only forward messages to Host if the socket is not bounded to `Thread stack`.

**For `IsHostInterface` flag:**
Furthermore, I think both `Udp Forward` and `Platform Udp` should respect the `IsHostInterface` flag and only forward for messages with `IsHostInterface=TRUE`  to Host. And sockets bind to Thread stack should not send any message with `IsHostInterface=TRUE`. 

In this way, we can make the socket binding strategies general for both `Udp Forward` and `Platform Udp`. 

Thoughts? @abtink @bukepo

src/core/net/udp6.cpp
```diff
@@ -100,6 +100,13 @@ otError Udp::Socket::Bind(uint16_t aPort)
     return Bind(SockAddr(aPort));
 }
 
+#if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
+otError Udp::Socket::BindToNetif(otNetifIdentifier aNetifIdentifier)
+{
+    return otPlatUdpBindToNetif(this, aNetifIdentifier);
+}
+#endif // OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
```
abtink, line 108:
```suggestion
#endif 
```

src/core/thread/tmf.cpp
```diff
@@ -0,0 +1,82 @@
...
+namespace Tmf {
+
+otError TmfAgent::Start(void)
+{
+    otError error;
+
+    SuccessOrExit(error = Coap::Start(kUdpPort));
+#if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
+    error = mSocket.BindToNetif(OT_NETIF_THREAD);
+    VerifyOrExit(error = OT_ERROR_NONE, IgnoreError(mSocket.Close()));
```
abtink, line 48:
```suggestion
    VerifyOrExit(error == OT_ERROR_NONE, IgnoreError(mSocket.Close()));
```

src/core/net/udp6.hpp
```diff
@@ -167,21 +168,28 @@ class Udp : public InstanceLocator
...
-         * @retval OT_ERROR_NONE    Successfully bound the socket.
-         * @retval OT_ERROR_FAILED  Failed to bind UDP Socket.
+         * @retval OT_ERROR_NONE            Successfully bound the socket.
+         * @retval OT_ERROR_INVALID_ARGS    Unable to bind to Thread network interface with the given address.
+         * @retval OT_ERROR_FAILED          Failed to bind UDP Socket.
          *
          */
         otError Bind(const SockAddr &aSockAddr);
 
+        /**
```
abtink, line 180: How about adding `#if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE` here as well?

src/core/thread/tmf.hpp
```diff
@@ -0,0 +1,79 @@
...
+class TmfAgent : public Coap::Coap
+{
+public:
+    explicit TmfAgent(Instance &aInstance)
+        : Coap::Coap(aInstance)
+    {
+        SetInterceptor(&Filter, this);
+    }
+
+    otError Start(void);
```
abtink, line 60: Missing doxygen documentation?

src/core/thread/tmf.hpp
```diff
@@ -0,0 +1,79 @@
...
+};
+
+/**
+ * This class implements functionality of the Thread TMF agent.
+ *
+ */
+class TmfAgent : public Coap::Coap
+{
+public:
+    explicit TmfAgent(Instance &aInstance)
```
abtink, line 54: Same here?

src/core/thread/tmf.cpp
```diff
@@ -0,0 +1,82 @@
...
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file implements TMF functionality.
```
simonlingoogle, line 31: nit: 
```suggestion
 *   This file implements Thread Management Framework (TMF) functionalities.
```

src/posix/platform/netif.cpp
```diff
@@ -147,6 +147,9 @@ extern int
 #include "common/logging.hpp"
 #include "net/ip6_address.hpp"
 
+unsigned int gNetifIndex = 0;
+char         gNetifName[IFNAMSIZ];
```
simonlingoogle, line 151: nit:
```suggestion
char         gNetifName[IFNAMSIZ] = "";
```

src/posix/platform/udp.cpp
```diff
@@ -280,18 +278,54 @@ otError otPlatUdpBind(otUdpSocket *aUdpSocket)
...
 
-    VerifyOrExit(0 == setsockopt(fd, IPPROTO_IPV6, IPV6_MULTICAST_IF, &sPlatNetifIndex, sizeof(sPlatNetifIndex)),
+    VerifyOrExit(0 == setsockopt(fd, IPPROTO_IPV6, IPV6_MULTICAST_IF, &gNetifIndex, sizeof(gNetifIndex)),
                  error = OT_ERROR_FAILED);
 
 exit:
     if (error == OT_ERROR_FAILED)
     {
-        perror("otPlatUdpBind");
+        otLogWarnPlat("Failed to bind UDP socket: %s", otThreadErrorToString(error));
```
simonlingoogle, line 287:
```suggestion
        otLogCritPlat("Failed to bind UDP socket: %s", strerror(errno));
```

src/posix/platform/udp.cpp
```diff
@@ -257,9 +255,9 @@ otError otPlatUdpBind(otUdpSocket *aUdpSocket)
     otError error = OT_ERROR_NONE;
     int     fd;
 
-    assert(sPlatNetifIndex != 0);
+    assert(gNetifIndex != 0);
     assert(aUdpSocket->mHandle != nullptr);
-    VerifyOrExit(sPlatNetifIndex != 0, error = OT_ERROR_INVALID_STATE);
+    VerifyOrExit(gNetifIndex != 0, error = OT_ERROR_INVALID_STATE);
```
simonlingoogle, line 260: duplicate check for `gNetifIndex != 0` ?

src/posix/platform/udp.cpp
```diff
@@ -402,9 +436,13 @@ void platformUdpInit(const char *aIfName)
         DieNow(OT_EXIT_INVALID_ARGUMENTS);
     }
 
-    sPlatNetifIndex = if_nametoindex(aIfName);
+    if (aIfName != gNetifName)
+    {
+        VerifyOrDie(strlen(aIfName) < sizeof(gNetifName) - 1, OT_EXIT_INVALID_ARGUMENTS);
+        strcpy(gNetifName, aIfName);
+    }
```
simonlingoogle, line 443: Don't quite get the intention here. Any explanation? 

`gNetifIndex` is initialized by `platform/netif.cpp`, so it depends on `PLATFORM_NETIF`. 
But if `PLATFORM_NETIF` is `OFF`, how can we get `PLATFORM_UDP` work since it uses `gNetifIndex` in so many places.

src/core/net/udp6.cpp
```diff
@@ -269,8 +274,7 @@ otError Udp::SendTo(SocketHandle &aSocket, Message &aMessage, const MessageInfo
     messageInfoLocal.SetSockPort(aSocket.GetSockName().mPort);
 
 #if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
-    if (!IsMlePort(aSocket.mSockName.mPort) &&
-        !(aSocket.mSockName.mPort == ot::kCoapUdpPort && aMessage.GetSubType() == Message::kSubTypeJoinerEntrust))
```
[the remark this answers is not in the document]
bukepo: Because it uses different key mode from normal TMF frames.

src/posix/platform/udp.cpp
bukepo: assert() becomes empty in release build.

simonlingoogle: But it's very common to use OT_ASSERT without addition VerifyOrExit in OpenThread core code.

assert already means it should not happen unless there is a logical error? Do we need to use `VerifyOrExit` for logically error conditions?

src/posix/platform/udp.cpp
```diff
@@ -402,19 +436,22 @@ void platformUdpInit(const char *aIfName)
         DieNow(OT_EXIT_INVALID_ARGUMENTS);
     }
 
-    sPlatNetifIndex = if_nametoindex(aIfName);
-
-    if (sPlatNetifIndex == 0)
+    if (aIfName != gNetifName)
```
simonlingoogle, line 439: nit:
```suggestion
    if (strcmp(aIfName, gNetifName))
```

src/posix/platform/udp.cpp
```diff
@@ -402,19 +436,22 @@ void platformUdpInit(const char *aIfName)
...
+
+    assert(gNetifIndex != 0);
 }
 
 void platformUdpProcess(otInstance *aInstance, const fd_set *aReadFdSet)
 {
     otMessageSettings msgSettings = {false, OT_MESSAGE_PRIORITY_NORMAL};
 
-    VerifyOrExit(sPlatNetifIndex != 0, OT_NOOP);
+    VerifyOrExit(gNetifIndex != 0, OT_NOOP);
```
simonlingoogle, line 454: if `gNetifIndex == 0`, the program should have exited in `platformUdpInit`, 
Is it still necessary to check here?

src/posix/platform/udp.cpp
```diff
@@ -402,19 +436,22 @@ void platformUdpInit(const char *aIfName)
...
     }
 
-    sPlatNetifIndex = if_nametoindex(aIfName);
-
-    if (sPlatNetifIndex == 0)
+    if (aIfName != gNetifName)
     {
-        perror("if_nametoindex");
+        VerifyOrDie(strlen(aIfName) < sizeof(gNetifName) - 1, OT_EXIT_INVALID_ARGUMENTS);
+        strcpy(gNetifName, aIfName);
```
simonlingoogle, line 442: nit: just to be safer
```suggestion
        assert(gNetifIndex == 0);
        strcpy(gNetifName, aIfName);
```

simonlingoogle (approved):
LGTM :+1: 

with some nits

src/core/thread/tmf.hpp
```diff
@@ -0,0 +1,92 @@
...
+        : Coap::Coap(aInstance)
+    {
+        SetInterceptor(&Filter, this);
+    }
+
+    /**
+     * This method starts the TMF agent.
+     *
+     * @retval OT_ERROR_NONE    Successfully started the CoAP service.
+     * @retval OT_ERROR_ALREADY Already started.
```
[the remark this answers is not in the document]
bukepo: The error inherits from Coap::Start(). Why changing it to void?

src/posix/platform/udp.cpp
bukepo: This is intended, PLATFORM_NETIF pass gNetifName to to PLATFORM_UDP to initialize it.

src/core/thread/tmf.hpp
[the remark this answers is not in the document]
simonlingoogle: maybe we should modify `Coap::Start()` (in a separate PR).

simonlingoogle:
https://github.com/openthread/openthread/blob/9aa86a114cc6e7a13d4f79194a11fa85e8fa6cfb/src/posix/platform/udp.cpp#L353-L358

Here the socket is closed and recreated and rebound. How do we restore the UDP socket's original Netif binding ?

src/posix/platform/udp.cpp
bukepo:
```suggestion
```

bukepo:
> https://github.com/openthread/openthread/blob/9aa86a114cc6e7a13d4f79194a11fa85e8fa6cfb/src/posix/platform/udp.cpp#L353-L358
> 
> Here the socket is closed and recreated and rebound. How do we restore the UDP socket's original Netif binding ? @bukepo

Good point! Reading the bound device is not supported before Linux 3.8. We can only try to restore the original binding, but allow failures.

include/openthread/platform/udp.h
```diff
@@ -65,12 +65,13 @@ otError otPlatUdpClose(otUdpSocket *aUdpSocket);
  * This function binds the UDP socket by platform.
  *
  * @param[in]   aUdpSocket  A pointer to the UDP socket.
+ * @param[in]   aThreadOnly True to bind to Thread network interface only, otherwise do not bind to network interfaces.
```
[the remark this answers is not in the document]
simonlingoogle: I will add `BACKBONE` later.

src/posix/platform/udp.cpp
```diff
@@ -316,11 +348,27 @@ otError otPlatUdpConnect(otUdpSocket *aUdpSocket)
...
         // We create new socket to disconnect.
         SuccessOrExit(error = otPlatUdpClose(aUdpSocket));
         SuccessOrExit(error = otPlatUdpSocket(aUdpSocket));
         SuccessOrExit(error = otPlatUdpBind(aUdpSocket));
+
+        if (len > 0 && netifName[0] != '\0')
+        {
+            fd = FdFromHandle(aUdpSocket->mHandle);
+            VerifyOrExit(setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, &netifName, len) == 0, error = OT_ERROR_FAILED);
+        }
```
simonlingoogle, line 370: put before `SuccessOrExit(error = otPlatUdpBind(aUdpSocket));`

simonlingoogle:
> > https://github.com/openthread/openthread/blob/9aa86a114cc6e7a13d4f79194a11fa85e8fa6cfb/src/posix/platform/udp.cpp#L353-L358
> > 
> > Here the socket is closed and recreated and rebound. How do we restore the UDP socket's original Netif binding ? @bukepo
> 
> Good point! Reading the bound device is not supported before Linux 3.8. We can only try to restore the original binding, but allow failures.

Another way maybe saving the NetifIdentifier as a member field in aUdpSocket so that we can read it here directly. 
But I think `getsockopt` approach should be better.

src/posix/platform/udp.cpp
bukepo: What's the difference?

src/core/thread/tmf.hpp
```diff
@@ -0,0 +1,92 @@
...
+    otError Start(void);
+
+    /**
+     * This method returns whether Thread Management Framework Addressing Rules are met.
+     *
+     * @retval TRUE   if Thread Management Framework Addressing Rules are met.
+     * @retval FALSE  if Thread Management Framework Addressing Rules are not met.
+     *
+     */
+    bool IsTmfMessage(const Ip6::MessageInfo &aMessageInfo);
```
jwhui, line 82: Declare as `const`?

src/posix/platform/netif.cpp
jwhui: @bukepo , can we resolve this comment?

bukepo: global array is already initialized to all zeros.

romacdon:
@bukepo - can you provide some more background on this change?  I ask because we have some tests that seem to have broken with this change.  The tests try to bind to certain addresses like multicast addresses on SoCs.

fyi: @lmnotran @suveshpratapa

jwhui:
> @bukepo and/or @jwhui - can you provide some more background on this change? I ask because we have some tests that seem to have broken with this change.

@romacdon , Thread 1.2 Backbone Router requires supporting Thread protocols on both the Thread and Backbone links. We needed a way to control the binding of a given protocol endpoint to a specific interface. This change provides that level of control.

> The tests try to bind to certain addresses like multicast addresses on SoCs.

I don't think binding to IPv6 multicast addresses is correct - multicast addresses can only be used in the IPv6 Destination Address and not the IPv6 Source Address. There is a related discussion in https://github.com/openthread/ot-br-posix/issues/538.

romacdon:
Thanks @jwhui - I will follow up on the multicast issue in the other discussion.  We do also get the "invalid args" error when we execute the following command `udp bind fd46:db8:0:0:0:ff:fe00:2400 4242`.  What's the reason for that failure?

jwhui:
> Thanks @jwhui - I will follow up on the multicast issue in the other discussion. We do also get the "invalid args" error when we execute the following command `udp bind fd46:db8:0:0:0:ff:fe00:2400 4242`. What's the reason for that failure?

@romacdon , this PR also added a check to require that the IPv6 address is assigned to the network interface:
https://github.com/openthread/openthread/blob/b3c21f086c77a9b02b527770b06c4d9cdd4705a1/src/core/net/udp6.cpp#L296-L297

Is `fd46:db8:0:0:0:ff:fe00:2400` assigned to the Thread interface? You can check this by using the `ipaddr` command.

romacdon:
> 
> Is `fd46:db8:0:0:0:ff:fe00:2400` assigned to the Thread interface? You can check this by using the `ipaddr` command.

@jwhui Here's the output from one of our SQA tests:

[10.4.168.143:4901:I]ipaddr
[10.4.168.143:4901:O]> ipaddr
[10.4.168.143:4901:O]fd46:db8:0:0:0:ff:fe00:f000
[10.4.168.143:4901:O]fe80:0:0:0:cc31:dc12:d964:4732
[10.4.168.143:4901:O]fd46:db8:0:0:c17f:4f0:a75e:dd78
[10.4.168.143:4901:O]Done
[10.4.168.143:4901:O]> 
[10.4.168.142:4901:I]udp open
[10.4.168.142:4901:O]> udp open
[10.4.168.142:4901:O]Done
[10.4.168.142:4901:O]> 
[10.4.168.142:4901:I]udp bind fd46:db8:0:0:0:ff:fe00:f000 4242
[10.4.168.142:4901:O]> udp bind fd46:db8:0:0:0:ff:fe00:f000 4242
[10.4.168.142:4901:O]Error 7: InvalidArgs

simonlingoogle:
> > 
> 
> > Is `fd46:db8:0:0:0:ff:fe00:2400` assigned to the Thread interface? You can check this by using the `ipaddr` command.
> 
> 
> 
> @jwhui Here's the output from one of our SQA tests:
> 
> 
> 
> [10.4.168.143:4901:I]ipaddr
> 
> [10.4.168.143:4901:O]> ipaddr
> 
> [10.4.168.143:4901:O]fd46:db8:0:0:0:ff:fe00:f000
> 
> [10.4.168.143:4901:O]fe80:0:0:0:cc31:dc12:d964:4732
> 
> [10.4.168.143:4901:O]fd46:db8:0:0:c17f:4f0:a75e:dd78
> 
> [10.4.168.143:4901:O]Done
> 
> [10.4.168.143:4901:O]> 
> 
> [10.4.168.142:4901:I]udp open
> 
> [10.4.168.142:4901:O]> udp open
> 
> [10.4.168.142:4901:O]Done
> 
> [10.4.168.142:4901:O]> 
> 
> [10.4.168.142:4901:I]udp bind fd46:db8:0:0:0:ff:fe00:f000 4242
> 
> [10.4.168.142:4901:O]> udp bind fd46:db8:0:0:0:ff:fe00:f000 4242
> 
> [10.4.168.142:4901:O]Error 7: InvalidArgs
> 
> 

Are you binding to the same port on the same host? This PR binds the udp socket to the host udp port, so you can't use same port for multiple udp sockets. 

Can you use two different ports to avoid port conflict?

romacdon:
> Are you binding to the same port on the same host? This PR binds the udp socket to the host udp port, so you can't use same port for multiple udp sockets.
> 
> Can you use two different ports to avoid port conflict?

Hi @simonlingoogle. In our case it is an SoC build .  There is no host.  The test is using the ot-cli-ftd app.  The test worked prior to the changes introduced in this PR.

simonlingoogle:
> Hi @simonlingoogle. In our case it is an SoC build . There is no host. The test is using the ot-cli-ftd app. The test worked prior to the changes introduced in this PR.
> 

Sorry for not realizing that it's on an SoC build. 

As @jwhui has mentioned, this PR requires that the address to bind, if specified, must be one of local unicast addresses. 

```c++
VerifyOrExit(aSockAddr.GetAddress().IsUnspecified() || Get<ThreadNetif>().HasUnicastAddress(aSockAddr.GetAddress()),
                 error = OT_ERROR_INVALID_ARGS);
```
I tested your steps locally, and it worked just fine:
```bash
node 1> ipaddr
fdde:ad00:beef:0:0:ff:fe00:bc00
fdde:ad00:beef:0:57d3:3846:b105:f1c5
fe80:0:0:0:50e0:ab36:4525:7421
Done
node 1> udp open
Done
node 1> udp bind fdde:ad00:beef:0:0:ff:fe00:bc00 4242
Done
```

So, it's unclear why it failed in your testbed. I wonder if it's possible that the SoC device changed it's RLOC before `udp bind`.

PR #4973, opened 2020-05-18
[instance] disable instance member copying

Moandor-y:
This PR makes `Instance` and its members non-copyable to prevent issues like https://github.com/openthread/openthread/pull/4970.

### DIFF
--- a/src/core/Makefile.am
+++ b/src/core/Makefile.am
@@ -331,6 +331,7 @@ HEADERS_COMMON                             = \
     common/logging.hpp                       \
     common/message.hpp                       \
     common/new.hpp                           \
+    common/non_copyable.hpp                  \
     common/notifier.hpp                      \
     common/random.hpp                        \
     common/random_manager.hpp                \
--- a/src/core/coap/coap.hpp
+++ b/src/core/coap/coap.hpp
@@ -38,6 +38,7 @@
 #include "common/linked_list.hpp"
 #include "common/locator.hpp"
 #include "common/message.hpp"
+#include "common/non_copyable.hpp"
 #include "common/timer.hpp"
 #include "net/ip6.hpp"
 #include "net/netif.hpp"
@@ -268,7 +269,7 @@ private:
  * This class implements the CoAP client and server.
  *
  */
-class CoapBase : public InstanceLocator
+class CoapBase : public InstanceLocator, private NonCopyable
 {
     friend class ResponsesQueue;
 
--- a/src/core/common/extension.hpp
+++ b/src/core/common/extension.hpp
@@ -37,6 +37,7 @@
 #include "openthread-core-config.h"
 
 #include "common/locator.hpp"
+#include "common/non_copyable.hpp"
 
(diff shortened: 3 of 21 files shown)


src/core/utils/non_copyable.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * This class makes any class that derives from it non-copyable. It is
+ * intended to be used as a private base class.
+ */
+class NonCopyable
+{
+protected:
+    NonCopyable() {}
+    ~NonCopyable() {}
```
bukepo, line 48:
```suggestion
    NonCopyable(void) {}
    ~NonCopyable(void) {}
```

src/core/utils/non_copyable.hpp
```diff
@@ -0,0 +1,58 @@
...
+#define NON_COPYABLE_HPP_
+
+namespace ot {
+namespace Utils {
+
+/**
+ * This class makes any class that derives from it non-copyable. It is
+ * intended to be used as a private base class.
+ */
+class NonCopyable
```
bukepo, line 44:
```suggestion
class Uncopyable
```

src/core/utils/non_copyable.hpp
```diff
@@ -0,0 +1,58 @@
...
+{
+protected:
+    NonCopyable(void) {}
+    ~NonCopyable(void) {}
```
abtink, line 48: Can we consider removing the destructor? Do we need it?

src/core/utils/non_copyable.hpp
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * @file
+ *   This file includes a common base class for disabling copying.
+ */
+
+#ifndef NON_COPYABLE_HPP_
+#define NON_COPYABLE_HPP_
+
+namespace ot {
+namespace Utils {
```
abtink, line 38: What do you think of moving this to `common/non_copyable.hpp` from `utils/non_copyable.hpp`?
I feel this is a common type used by all core classes.

I generally consider `utils` folder is for utility modules that are adding some new/extra feature which is not directly part of core Thread implementation, e.g., jam detector, channel monitor, flash, etc.

Thoughts?

src/core/utils/non_copyable.hpp
Moandor-y: It's a common guideline that destructors for base classes must be either public and virtual or protected and non-virtual. Also, if someone mistakenly uses it as a public base class, this prevents it from being deleted by a pointer to base class.

src/core/utils/non_copyable.hpp
Moandor-y: Sounds reasonable. I'm OK with both. @bukepo @jwhui thoughts?

bukepo: I agree with the proposal.

Moandor-y: Updated.

PR #4974, opened 2020-05-18
[continuous-integration] fix oss-fuzz version

Moandor-y:
Newest https://github.com/google/oss-fuzz introduced something that breaks the CI. Point to an old version before it's fixed.


Moandor-y:
This can't be fixed by checking out old versions. `oss-fuzz` always uses the latest Docker image from gcr.io/oss-fuzz-base/cifuzz-base.

PR #4976, opened 2020-05-18
[mac] move auxiliary frame counter from Mac to SubMac

brianljt:
In Thread 1.2, it is possible to receive an IEEE 802.15.4-2015 packet. Per specification, receiver should acknowledge this packet with an IEEE 802.15.4-2015 ACK(Enh-ACK). This Enh-ACK can include header IE with it and requires `security enabled` bit in `FCF` be set to true. It is impractical for the host to generate the Enh-ACK and send to RCP for transmission within `AIFS` time(192us). So RCP should prepare the Enh-ACK by itself, which requires it to fill in the frame counter and do the encryption/authentication. This PR follows up with PR#4919 and tries to address the need of filling auxiliary frame counter in RCP by including the following changes,

1. Move frame counter related functions from MAC layer to SubMac layer, which is mirrored in RCP.
2. Set `mMacFrameCounter` to RCP using newly added spinel properties.
3. RCP reports last used frame counter to host in TxDone and RxDone.
4. Add a simulation test for reset verification.


src/core/radio/radio.hpp
```diff
@@ -147,6 +147,18 @@ class Radio : public InstanceLocator
...
+         * This callback method handles "store MAC frame counter" event from radio platform.
+         *
+         * This method is used when radio provides OT_RADIO_CAPS_TRANSMIT_SEC capability. It is called from
+         * `otPlatRadioMacFrameCounterStore()`.
+         *
+         * @param[in]  aInstance           The OpenThread instance structure.
+         * @param[in]  aMacFrameCounter    The current MAC frame counter value.
+         *
+         */
+        void HandleMacFrameCounterStore(uint32_t aMacFrameCounter);
```
Irving-cl, line 160: No `aInstance` in the function arguments?

src/ncp/ncp_base_radio.cpp
```diff
@@ -459,6 +480,42 @@ template <> otError NcpBase::HandlePropertySet<SPINEL_PROP_RCP_MAC_KEY>(void)
...
+}
+
+template <> otError NcpBase::HandlePropertySet<SPINEL_PROP_RCP_MAC_FRAME_COUNTER>(void)
+{
+    otError  error = OT_ERROR_NONE;
+    uint32_t frameCounter;
+
+    SuccessOrExit(error = mDecoder.ReadUint32(frameCounter));
+
+    error = otLinkRawSetMacFrameCounter(mInstance, frameCounter, &NcpBase::LinkRawMacFrameCounterStore);
```
Irving-cl, line 500: Can we set the store callback at initialization instead of passing it every time we want to set and store the frame counter?

src/ncp/ncp_base_radio.cpp
```diff
@@ -123,6 +123,10 @@ void NcpBase::LinkRawTransmitDone(otRadioFrame *aFrame, otRadioFrame *aAckFrame,
 
         SuccessOrExit(mEncoder.BeginFrame(header, SPINEL_CMD_PROP_VALUE_IS, SPINEL_PROP_LAST_STATUS));
         SuccessOrExit(mEncoder.WriteUintPacked(ThreadErrorToSpinelStatus(aError)));
+
+        // Transmit frame header
+        SuccessOrExit(mEncoder.WriteDataWithLen(aFrame->mPsdu, static_cast<Mac::TxFrame *>(aFrame)->GetHeaderLength()));
```
zhanglongxia, line 128: How about move line 128 to line 152, then the spinel frame will be compatible with the original spinel frame format.

src/core/mac/link_raw.cpp
```diff
@@ -220,6 +221,50 @@ otError LinkRaw::SetMacKey(uint8_t        aKeyIdMode,
...
+}
+
+otError LinkRaw::SetMacFrameCounter(uint32_t aMacFrameCounter, otLinkRawMacFcStore aCallback)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(IsEnabled(), error = OT_ERROR_INVALID_STATE);
+    mSubMac.SetMacFrameCounter(aMacFrameCounter);
+
+    mMacFcStoreCallback = aCallback;
```
zhanglongxia, line 241: Both the funtion 'LinkRaw::SetMacFrameCounter'  and 'LinkRaw::SetStoredMacFrameCounterr' are using the same variable 'mMacFcStoreCallback' to record the callback function. What if these two functions are set with two different callback functions?

src/core/mac/sub_mac.cpp
```diff
@@ -657,6 +661,74 @@ void SubMac::SetMacKey(uint8_t        aKeyIdMode,
...
+{
+    mStoredMacFrameCounter = aStoredMacFrameCounter;
+
+    VerifyOrExit(!ShouldHandleTransmitSecurity(), OT_NOOP);
+
+    Get<Radio>().SetStoredMacFrameCounter(aStoredMacFrameCounter);
+
+exit:
+    return;
+}
```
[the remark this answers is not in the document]
brianljt: Usually, the caller to this function only sets `MacFrameCounter` or `StoredMacFrameCounter` but not both. For example, `Mle::Store()` sets `StoredMacFrameCounter` only while `Mle::Restore()` and `KeyManager::SetCurrentKeySequence()` sets `MacFrameCounter` only.

src/core/mac/sub_mac.cpp
```diff
@@ -60,6 +60,8 @@ SubMac::SubMac(Instance &aInstance)
     , mPcapCallback(NULL)
     , mPcapCallbackContext(NULL)
     , mKeyId(0)
+    , mMacFrameCounter(0)
```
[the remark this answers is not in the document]
brianljt: Why change it this way. This member was moved from class `KeyManager` to `SubMac`. There are member called `mMleFrameCounter` and `mStoredMleFrameCounter`. The naming of `mMacFrameCounter` makes it easy for reader to understand which frame counter this member holds.

include/openthread/link_raw.h
```diff
@@ -351,6 +351,62 @@ otError otLinkRawSetMacKey(otInstance *   aInstance,
...
+
+/**
+ * Gets the current MAC frame counter value.
+ *
+ * @param[in]  aInstance  A pointer to an OpenThread instance.
+ *
+ * @returns The MAC frame counter.
+ *
+ */
+uint32_t otLinkRawGetMacFrameCounter(otInstance *aInstance);
```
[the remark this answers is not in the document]
brianljt: Yes, this is needed. `Mle::Store()` will be called not only because `MacFrameCounter` has reaches the threshold, but could be other reason. For example, when state changed, it will be called to save all the network info to NVM. So this API is needed to get the latest counter in RCP.

Another function is `Mle::AppendLinkFrameCounter`. This will fill the Link Frame Counter TLV in several MLE frames. This requires to get the latest counter in use, too.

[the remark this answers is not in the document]
bukepo: Can this API guarantee getting the latest frame counter? What if an enh-ack is transmitted just after the API returned?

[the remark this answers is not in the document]
brianljt: Not guaranteed. But I think it is OK. The differences should be small and host will save (current value + 1000) to NVM. I think it is big enough to cover it.

src/core/mac/sub_mac.cpp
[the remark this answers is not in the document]
bukepo: This is a member of MAC layer itself. I don't think the `Mac` prefix is necessary. Otherwise, we'll need `mMacState`. Thoughts? The `mMleFrameCounter` in KeyManager, which doesn't belong to mac layer or MLE layer.

include/openthread/link_raw.h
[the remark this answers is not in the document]
bukepo: I think we can include the frame counter used in `mRxInfo` in `otRadioFrame`, just like `mAckedWithFramePending`. Thus we can save a API, and the mechanism to notify host. Thoughts?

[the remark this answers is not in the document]
brianljt: Yes, that could be another way to handle this. Frame counter should be increased by

1. Transmitted secured frame.
2. Transmitted Enh-ACK in response to a received frame.

So we can put the frame counter in TxInfo(from transmitted frame) and RxInfo(for transmitted enh-ack) so that RCP does not need to track the stored value. Host will read each frame info and save it to NVM when necessary.

src/core/mac/sub_mac.cpp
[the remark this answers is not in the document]
brianljt: Sound reasonable. Will change that. Thanks!

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -865,12 +865,27 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::ParseRadioFrame(otRadioF
...
 
         aFrame.mInfo.mRxInfo.mAckedWithFramePending = ((flags & SPINEL_MD_FLAG_ACKED_FP) != 0);
+
+        if (mRadioCaps & OT_RADIO_CAPS_TRANSMIT_SEC)
+        {
+            aFrame.mInfo.mRxInfo.mAckedWithSecEnhAck = ((flags & SPINEL_MD_FLAG_ACKED_SEC) != 0);
+
+            unpacked = spinel_datatype_unpack(aBuffer, aLength, SPINEL_DATATYPE_UINT8_S SPINEL_DATATYPE_UINT32_S,
+                                              &aFrame.mInfo.mRxInfo.mAckKeyId, &aFrame.mInfo.mRxInfo.mAckFrameCounter);
+            VerifyOrExit(unpacked > 0, error = OT_ERROR_PARSE);
```
abtink, line 883: Can you please help update `spinel.h` documentation for the newly added fields?

I would also suggest in the spinel frame format we put the new fields (keyid and ack frame counter) into a spinel struct. This way if more fields need to be added in future to spinel to frame format, for cases that we won't need these security related fields it can be added as an empty struct and skipped over during parsing (this would help with backward compatibility).

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1445,30 +1472,35 @@ void RadioSpinel<InterfaceType, ProcessContextType>::HandleTransmitDone(uint32_t
...
     {
         error = SpinelStatusToOtError(status);
     }
 
+    if (mRadioCaps & OT_RADIO_CAPS_TRANSMIT_SEC && static_cast<Mac::TxFrame *>(mTransmitFrame)->GetSecurityEnabled())
+    {
+        uint8_t  keyId;
+        uint32_t frameCounter;
+
+        // Replace transmit frame auxiliary key index and frame counter with the one filled by RCP
```
abtink, line 1496:
```suggestion
        // Replace transmit frame security key index and frame counter with the one filled by RCP
```

src/ncp/ncp_base_radio.cpp
```diff
@@ -51,20 +51,12 @@ namespace Ncp {
 // MARK: Raw Link-Layer Datapath Glue
 // ----------------------------------------------------------------------------
 
-void NcpBase::LinkRawReceiveDone(otInstance *, otRadioFrame *aFrame, otError aError)
-{
-    sNcpInstance->LinkRawReceiveDone(aFrame, aError);
-}
-
-void NcpBase::LinkRawReceiveDone(otRadioFrame *aFrame, otError aError)
+otError NcpBase::LinkRawPackRadioFrame(otRadioFrame *aFrame, otError aError)
```
abtink, line 54: Suggest renaming to `PackRadioFrame`? 
I think the other methods with `LinkRaw` prefix have `LinkRaw` in the name since they are callbacks from `LinkRaw` module.

src/lib/spinel/radio_spinel_impl.hpp
brianljt: Thank you Abtin for the review. Accept and resolve all comments.

For this one. Looks like in `spinel.h`, `SPINEL_PROP_LAST_STATUS` only specify the first status code for host-to-NCP case. The information passed along with this property in `TxDone` for host-to-RCP case is not specified. Do you want me to summarize it and add it in `spinel.h`?

src/core/mac/sub_mac.hpp
```diff
@@ -167,6 +167,14 @@ class SubMac : public InstanceLocator
...
          */
         void EnergyScanDone(int8_t aMaxRssi);
+
+        /**
+         * This method notifies user of `SubMac` that MAC frame counter is updated.
+         *
+         * @param[in]  aFrameCounter  The MAC frame counter value.
+         *
+         */
+        void FrameCounterUpdated(uint32_t);
```
bukepo, line 177:
```suggestion
        void FrameCounterUpdated(uint32_t aFrameCounter);
```

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1445,30 +1463,37 @@ void RadioSpinel<InterfaceType, ProcessContextType>::HandleTransmitDone(uint32_t
...
-            mAckRadioFrame.mLength = 0;
-        }
+        aLength -= static_cast<uint16_t>(unpacked);
     }
     else
     {
         error = SpinelStatusToOtError(status);
     }
 
+    if (mRadioCaps & OT_RADIO_CAPS_TRANSMIT_SEC && static_cast<Mac::TxFrame *>(mTransmitFrame)->GetSecurityEnabled())
```
bukepo, line 1484:
```suggestion
    if ((mRadioCaps & OT_RADIO_CAPS_TRANSMIT_SEC) && static_cast<Mac::TxFrame *>(mTransmitFrame)->GetSecurityEnabled())
```

src/lib/spinel/radio_spinel_impl.hpp
abtink: Thanks @brianljt. 👍 
I see! This is `LAST_STTUS` (I thought it was `STREAM_RAW`). I think it's fine.  

As a future enhancement, I think we should conside re-designing/changing the  the RCP spinel model (maybe add new commands and callback that are more related to the radio APIs). I never liked the idea of using `LAST_STATUS` for tx done callback. IIRC this was quick solution (more of a hack) from when link-raw was added quickly.

src/lib/spinel/spinel.h
```diff
@@ -343,7 +343,7 @@
 // ----------------------------------------------------------------------------
 
 #define SPINEL_PROTOCOL_VERSION_THREAD_MAJOR 4
-#define SPINEL_PROTOCOL_VERSION_THREAD_MINOR 3
+#define SPINEL_PROTOCOL_VERSION_THREAD_MINOR 4
```
abtink, line 346: I think we should not change the `SPINEL_PROTOCOL_VERSION_THREAD_MINOR`.

- The policy for spinel is to ensure backward compatibility when adding new fields and properties (i.e. newer NCP image should stay compatible with older host side drive).
https://github.com/openthread/openthread/blob/f8f75d6d05064fd63e81a3ad4518e5b3728303f8/src/lib/spinel/spinel.h#L295-L306

- We don't have a explicit policy for updating the spinel version number, but the idea for it is that it would change when there is bigger change that does not gurantee backward comatibility.
- **Changing the version can add new restrictions in product integration and firmware update (between host firmware and NCP/RCP firmware ).**


Here is the challenge with Posix/RCP model if we change the `VERSION_THREAD_MINOR`.
- The host code (radio spinel) does check the version of RCP to match its own version and otherwise it would exit with an error.
https://github.com/openthread/openthread/blob/f8f75d6d05064fd63e81a3ad4518e5b3728303f8/src/lib/spinel/radio_spinel_impl.hpp#L232-L252
- So with this existing code, changing the VERSION_MINOR would mean, whenever POSIX code is updated on host side, we need to also update the RCP firmware along with it.

We have similar check in `wpantund` which checks the version (and also in other internal drivers we use on products (RTOS driver)) which mean we would need to update `wpantund` on host with any change on NCP firmware (which is not desirable).

brianljt: I see. I will add the new fields at the end of the spinel to maintain backward compatibility. However, I don't quite like current implementation of TxDone metadata. It is appended at the end of `SPINEL_PROP_LAST_STATUS`. It is not well organized and not easy to add new info. Maybe we can consider changing it later when the version of spinel protocol is updated someday.

abtink: Me too. I very much dislike the use of `LAST_STATUS` for tx done as well. Mentioned the same here https://github.com/openthread/openthread/pull/4976#discussion_r433343963. (I would be very happy if/when we change/re-design this in future... :)...).

src/core/mac/sub_mac.cpp
```diff
@@ -201,6 +202,11 @@ void SubMac::HandleReceiveDone(RxFrame *aFrame, otError aError)
         mPcapCallback(aFrame, false, mPcapCallbackContext);
     }
 
+    if (!ShouldHandleTransmitSecurity() && aFrame != NULL && aFrame->mInfo.mRxInfo.mAckedWithSecEnhAck)
```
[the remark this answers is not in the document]
abtink: That's a good question.

In general, how would we handle the enhanced ack for a received frame in the case where OT (`SubMac`) is handling the security frame counter and the counter is not given to the radio platform layer (from `SetFrameCounter()` below). 
Do we not allow any enhanced ack in that case?

src/core/mac/sub_mac.cpp
```diff
@@ -390,10 +390,15 @@ void SubMac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aEr
 
     if (!ShouldHandleTransmitSecurity() && aFrame.GetSecurityEnabled())
     {
+        uint8_t  keyIdMode;
         uint32_t frameCounter = 0;
 
-        OT_ASSERT(aFrame.GetFrameCounter(frameCounter) == OT_ERROR_NONE);
-        UpdateFrameCounter(frameCounter);
+        IgnoreError(aFrame.GetKeyIdMode(keyIdMode));
+        if (keyIdMode == Frame::kKeyIdMode1)
```
abtink, line 397: 👍

src/core/mac/sub_mac.cpp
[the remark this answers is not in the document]
abtink: I think we can think more on this and address it later (after merging this PR).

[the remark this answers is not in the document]
brianljt: There are two cases:

1. Radio or spinel_radio supports `OPENTHREAD_CONFIG_MAC_SOFTWARE_TX_SECURITY_ENABLE`, then the frame counter is maintained by radio layer. In `SubMac`, this metainfo should be used to update the counter and invoke NVM save if needed.

2. Radio or spinel_radio does not supports `OPENTHREAD_CONFIG_MAC_SOFTWARE_TX_SECURITY_ENABLE`, then the frame counter is maintained by `SubMac`. Enh-Ack is also generated at `SubMac` which should be using and incrementing the frame counter there. `SubMac` should be the source of the truth rather than rely on the information from lower layer.

include/openthread/instance.h
```diff
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (1)
+#define OPENTHREAD_API_VERSION (2)
```
librasungirl, line 56: @bukepo what would be expected if multiple PRs update API version? 

Like the other one in [#4900](https://github.com/openthread/openthread/pull/4900/commits/76be02688ceb79b98c26a2113c18ab98494d969f#diff-68d9781e82e4802a218d26d0595157a7R56)

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1445,30 +1461,37 @@ void RadioSpinel<InterfaceType, ProcessContextType>::HandleTransmitDone(uint32_t
     aBuffer += unpacked;
     aLength -= static_cast<uint16_t>(unpacked);
 
+    unpacked = spinel_datatype_unpack(aBuffer, aLength, SPINEL_DATATYPE_BOOL_S, &framePending);
+    OT_UNUSED_VARIABLE(framePending);
```
librasungirl, line 1465: nit: move to the top of the function?

[STYLE_GUILD.md](https://github.com/openthread/openthread/blob/master/STYLE_GUIDE.md)
> All unused parameters shall be declared as such using the OT_UNUSED_VARIABLE macro at the top of a function or method before all local variable declarations.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1433,8 +1448,9 @@ void RadioSpinel<InterfaceType, ProcessContextType>::HandleTransmitDone(uint32_t
...
                                                                         uint16_t          aLength)
 {
-    otError         error  = OT_ERROR_NONE;
-    spinel_status_t status = SPINEL_STATUS_OK;
+    otError         error        = OT_ERROR_NONE;
+    spinel_status_t status       = SPINEL_STATUS_OK;
+    bool            framePending = false;
     spinel_ssize_t  unpacked;
 
     VerifyOrExit(aCommand == SPINEL_CMD_PROP_VALUE_IS && aKey == SPINEL_PROP_LAST_STATUS, error = OT_ERROR_FAILED);
```
librasungirl, line 1456: not relative to this PR, but the doxygen for `SPINEL_PROP_LAST_STATUS` should be updated? @bukepo 

https://github.com/openthread/openthread/blob/6de8d778b784d2ef6af409e6b85d511cbc8fc3b5/src/lib/spinel/spinel.h#L1149-L1163

librasungirl: revise a bit - also relative to this PR. @brianljt

src/lib/spinel/spinel.h
```diff
@@ -3952,6 +3953,16 @@ enum
      */
     SPINEL_PROP_RCP_MAC_KEY = SPINEL_PROP_RCP__BEGIN + 0,
 
+    /// MAC Frame Counter
+    /** Format: `L`.
+     *
+     *  `L`: MAC frame counter
+     *
+     * The Spinel property is used to set/get MAC frame counter to and from RCP.
```
librasungirl, line 3961: It seems a `write-only` property? update the doxygen?

include/openthread/instance.h
jwhui: I think we need to simply flag these during the review process (e.g. add the `api: user` and `api: platform` labels) and ensure such PRs are up-to-date with latest master before merging.

include/openthread/platform/radio.h
```diff
@@ -234,11 +234,14 @@ typedef struct otRadioFrame
              */
             uint64_t mTimestamp;
 
-            int8_t  mRssi; ///< Received signal strength indicator in dBm for received frames.
-            uint8_t mLqi;  ///< Link Quality Indicator for received frames.
+            int8_t   mRssi;            ///< Received signal strength indicator in dBm for received frames.
+            uint8_t  mLqi;             ///< Link Quality Indicator for received frames.
+            uint8_t  mAckKeyId;        ///< ACK security key index (applicable when `mAckedWithSecEnhAck` is set).
+            uint32_t mAckFrameCounter; ///< ACK security frame counter (applicable when `mAckedWithSecEnhAck` is set).
```
jwhui, line 240: nit: can we reorder to minimize padding?

include/openthread/platform/radio.h
```diff
@@ -475,6 +478,17 @@ void otPlatRadioSetMacKey(otInstance *    aInstance,
                           const otMacKey *aCurrKey,
                           const otMacKey *aNextKey);
 
+/**
+ * This method sets the current MAC frame counter value.
+ *
+ * This function is used when radio provides OT_RADIO_CAPS_TRANSMIT_SEC capability.
```
jwhui, line 484:
```suggestion
 * This function is used when radio provides `OT_RADIO_CAPS_TRANSMIT_SEC` capability.
```

include/openthread/platform/radio.h
```diff
@@ -635,6 +649,9 @@ extern void otPlatRadioTxStarted(otInstance *aInstance, otRadioFrame *aFrame);
  * The radio driver calls this function to notify OpenThread that the transmit operation has completed,
  * providing both the transmitted frame and, if applicable, the received ack frame.
  *
+ * When radio provides `OT_RADIO_CAPS_TRANSMIT_SEC` capability, radio platform layer updates `aFrame`
+ * with the security frame counter and key index values used in the transmission of the frame.
```
jwhui, line 653:
```suggestion
 * When radio provides `OT_RADIO_CAPS_TRANSMIT_SEC` capability, radio platform layer updates @p aFrame
 * with the security frame counter and key index values maintained by the radio.
```

src/core/mac/mac.cpp
```diff
@@ -970,19 +970,7 @@ void Mac::ProcessTransmitSecurity(TxFrame &aFrame)
...
-        // manager.
-
-        if (!aFrame.IsARetransmission())
-        {
-            aFrame.SetFrameCounter(keyManager.GetMacFrameCounter());
-            keyManager.IncrementMacFrameCounter();
-        }
-
-        // For MAC key ID mode 1, the AES CCM* is done at SubMac or Radio if supported
+        // For MAC key ID mode 1, the security framec counter update and CCM* is done at SubMac or Radio if supported
```
jwhui, line 973:
```suggestion
        // For MAC Key ID Mode 1, the security frame counter update and CCM* is done at SubMac or Radio depending on `OT_RADIO_CAPS_TRANSMIT_SEC`.
```

src/core/thread/key_manager.hpp
```diff
@@ -260,15 +260,15 @@ class KeyManager : public InstanceLocator
      * @returns The current MAC Frame Counter value.
      *
      */
-    uint32_t GetMacFrameCounter(void) const { return mMacFrameCounter; }
+    uint32_t GetMacFrameCounter(void);
```
jwhui, line 263: Declare as `const`?

src/lib/spinel/radio_spinel_impl.hpp
brianljt: Please check the discussion here. https://github.com/openthread/openthread/pull/4976#discussion_r433343963

The TxDone info is appended after `SPINEL_PROP_LAST_STATUS`

PR #4980, opened 2020-05-18
[posix] allow configuration of macOS tunnel driver selection outside of code

rob-the-dude:
In order to enable automated testing of mac OS, we're likely to need the
ability to build for utun instead of tun, so make that option configurable via
external options, and make the default utun, which will be easier for most developers
to use (the caveats are documented in netif.cpp).


src/posix/platform/openthread-posix-config.h
```diff
@@ -101,4 +101,28 @@
...
+#define OPENTHREAD_POSIX_CONFIG_MACOS_TUN 2
+
+/**
+ * @def OPENTHREAD_POSIX_MACOS_TUN_OPTION
+ *
+ * This setting configures which tunnel driver to use.
+ *
+ */
+#ifndef OPENTHREAD_POSIX_MACOS_TUN_OPTION
+#define OPENTHREAD_POSIX_MACOS_TUN_OPTION OPENTHREAD_POSIX_CONFIG_MACOS_UTUN
```
jwhui, line 123:
```suggestion
#define OPENTHREAD_POSIX_CONFIG_MACOS_TUN_OPTION OPENTHREAD_POSIX_CONFIG_MACOS_UTUN
```
nit: to be consistent with other config options in this file?

bukepo: Should OPENTHREAD_POSIX_CONFIG_MACOS_UTUN be renamed to `OT_*`?

rob-the-dude: Made suggested changes.  Didn't see (didn't actually look for) guidance for why the options should be OT_POSIX_* instead of OPENTHREAD_POSIX_*, but I see that's being done for the RCP bus configuration, so I'm happy to be consistent with that.

PR #4983, opened 2020-05-19
[cmake] add NCP vendor hook option from Zephyr

Philipp-Wohlgenannt-Tridonic-com:
Add NCP vendor hook option from Zephyr.


Philipp-Wohlgenannt-Tridonic-com:
@googlebot I signed it!

src/ncp/CMakeLists.txt
```diff
@@ -85,6 +85,13 @@ set(COMMON_SOURCES
     ncp_uart.cpp
 )
 
+if("OPENTHREAD_ENABLE_NCP_VENDOR_HOOK=1" IN_LIST OT_PRIVATE_DEFINES)
+    if(NOT OPENTHREAD_NCP_VENDOR_HOOK_SOURCE)
+        message(FATAL_ERROR "OPENTHREAD_NCP_VENDOR_HOOK_SOURCE must be defined when OPENTHREAD_ENABLE_NCP_VENDOR_HOOK=1")
+    endif()
+    list(APPEND COMMON_SOURCES ${OPENTHREAD_NCP_VENDOR_HOOK_SOURCE})
+endif()
```
bukepo, line 93: can OPENTHREAD_NCP_VENDOR_HOOK_SOURCE and OT_NCP_VENDOR_HOOK be a single option?
```suggestion
if(OPENTHREAD_NCP_VENDOR_HOOK_SOURCE)
    list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_ENABLE_NCP_VENDOR_HOOK=1")
    list(APPEND COMMON_SOURCES ${OPENTHREAD_NCP_VENDOR_HOOK_SOURCE})
endif()
```

Philipp-Wohlgenannt-Tridonic-com: Yes they can. I will change it and i will make new commits for that in this pull request and the related Zephyr pull request.

etc/cmake/options.cmake
```diff
@@ -176,6 +176,11 @@ if(OT_PLATFORM_NETIF)
     list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE=1")
 endif()
 
+option(OT_NCP_VENDOR_HOOK "Enable vendor hook for NCP")
```
bukepo, line 179: Is this still necessary?

Philipp-Wohlgenannt-Tridonic-com: The keyword OPENTHREAD_ENABLE_NCP_VENDOR_HOOK is still necessary and defined here to have a additional abstraction layer. Other defines are done in the same way (e.g. OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE). It is used in /src/ncp/ncp_uart.cpp.

bukepo: from https://cmake.org/cmake/help/latest/command/option.html, it seems `option()` is just for defining `ON`/`OFF`, so I think we should move this to `src/ncp/CMakefile`. Thoughts?

markus-becker-tridonic-com: Originally we wanted to use the same variables that were used in https://github.com/openthread/openthread/blob/master/configure.ac#L688-L707, but were asked to combine it into one option/variable. We try to do the same as `--with-ncp-vendor-hook-source=<VENDOR_HOOK.CPP>` but via cmake and not via configure. Suggestions?

bukepo: I see the problem. I'm fine with this PR if it's working and I'm considering introducing a `ot-config` target to better support change OpenThread configuration in a deep subdirectory.

Philipp-Wohlgenannt-Tridonic-com: I've changed the name of OT_NCP_VENDOR_HOOK to OT_NCP_VENDOR_HOOK_SOURCE to align with https://github.com/openthread/openthread/blob/master/configure.ac#L688-L707.

I've tested it locally with my openthread/zephyr device and it worked fine.

PR #4985, opened 2020-05-20
[cli] add command for getting the network interface name and index

rob-the-dude:
On some platform (mostly BSDs), we do not have the option to specify the tunnel name, although most of those platforms do have a way to dynamically obtain a unique interface name if desired.  Add a CLI command -- only enabled on platforms that have PLATFORM_NETIF enabled (which is basically just POSIX right now) -- that will return the name of the tunnel interface, as well as it's index.  Both are useful in doing automated tests on non-Linux systems.


src/posix/platform/netif.cpp
```diff
@@ -1456,4 +1457,20 @@ void platformNetifProcess(const fd_set *aReadFdSet, const fd_set *aWriteFdSet, c
     return;
 }
 
+otError otPlatGetNetif(otInstance *aInstance, const char **outNetIfName, unsigned int *outNetIfIndex)
+{
+    OT_UNUSED_VARIABLE(aInstance);
+
+    otError error;
+
+    VerifyOrExit(sTunIndex != 0, error = OT_ERROR_FAILED);
```
bukepo, line 1466: If netif failed to initialize, the should already died. So is it necessary to check this?

include/openthread/platform/misc.h
```diff
@@ -198,6 +198,23 @@ otPlatMcuPowerState otPlatGetMcuPowerState(otInstance *aInstance);
...
+ *
+ * @param[out] outNetIfName  A pointer for the returned network interface name.
+ *
+ * @param[in] outNetIfIndex  A pointer for the returned network interface index (i.e., if_nametoindex).
+ *
+ * @retval OT_ERROR_NONE     Successfully returned the network interface and index.
+ * @retval OT_ERROR_FAILED   The network interface is not enabled or is unknown.
+ *
+ */
+otError otPlatGetNetif(otInstance *aInstance, const char **outNetIfName, unsigned int *outNetIfIndex);
```
bukepo, line 215: @jwhui we have a `otCliSetUserCommands()`, should we add a otCliSetPlatCommands() as well so that we can keep platform specific commands in platform layer.

jwhui: @bukepo, I generally agree. I think we can clean this up in a separate PR.

include/openthread/platform/misc.h
```diff
@@ -198,6 +198,23 @@ otPlatMcuPowerState otPlatGetMcuPowerState(otInstance *aInstance);
...
 
+#if defined(OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE) && OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE
+/**
+ * This function gets the name and index of the platform's network interface (if it exists).
+ *
+ * @param[in] aInstance  A pointer to OpenThread instance.
+ *
+ * @param[out] outNetIfName  A pointer for the returned network interface name.
+ *
+ * @param[in] outNetIfIndex  A pointer for the returned network interface index (i.e., if_nametoindex).
```
jwhui, line 209:
```suggestion
 * @param[in]  aInstance      A pointer to OpenThread instance.
 * @param[out] outNetIfName   A pointer for the returned network interface name.
 * @param[out] outNetIfIndex  A pointer for the returned network interface index (i.e., if_nametoindex).
```

src/posix/platform/netif.cpp
rob-the-dude: It may not be strictly necessary, but it seems safe and relatively low-cost -- if the initialization logic ever changed to NOT die, this code would continue working.  OK to leave?

include/openthread/platform/misc.h
rob-the-dude: I'm OK with additional cleanup in another PR :-).

PR #4989, opened 2020-05-20
[script] install shfmt

bukepo:
This commit installs shfmt in script/bootstrap.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -60,7 +60,8 @@ install_packages_apt()
     python3 -m pip install mdv || echo 'Failed to install markdown render. Install it manually if you need.'
 
     # add shfmt for shell pretty, try brew because snap does not support home directory not being /home
-    command -v shfmt || brew install shfmt || (sudo apt-get install --no-install-recommends -y snapd && sudo snap install shfmt)
+    command -v shfmt || brew install shfmt || (sudo apt-get install --no-install-recommends -y snapd \
+        && (sudo systemctl daemon-reload || true) && sudo service snapd start && sudo snap install shfmt)
 }
 
 install_packages_opkg()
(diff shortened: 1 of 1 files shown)


simonlingoogle:
> This commit installs shfmt in script/bootstrap.

why not just install it in `make-pretty`?

script/bootstrap
```diff
@@ -58,6 +58,9 @@ install_packages_apt()
 
     # add mdv for local size report
     python3 -m pip install mdv || echo 'Failed to install markdown render. Install it manually if you need.'
+
+    # add shfmt for shell pretty
+    command -v shfmt || brew install shfmt || (sudo apt-get install --no-install-recommends -y snapd && sudo snap install shfmt)
```
jwhui, line 63: Does attempting to use `brew install` make sense in `install_packages_apt()`?

bukepo: Yes, snap doesn't support home dir not being at /home. Linuxbrew works perfect. So try brew first.

bukepo:
> > This commit installs shfmt in script/bootstrap.
> 
> 
> 
> why not just install it in `make-pretty`?

I don't see other style tools are installed in that script.

script/bootstrap
jwhui: Makes sense. Can we add a comment since having `brew install` inside `install_packages_apt()` is not intuitive?

bukepo: added.

bukepo:
There are no logs of the docker build failures. And I cannot re-trigger the build. I will close and re-open and see if docker build is rebuilt and hopefully I can get the error logs.

bukepo:
close and reopen to trigger docker hub build.

PR #4990, opened 2020-05-20
[mac-frame] fix problems to generate enh ack

Irving-cl:
This PR fixes small problems in #4884 
1. I used to assume security is always enabled in enhanced ack, however that's not always the case. So add security part only when it's required.
2. `FindHeaderIeIndex` would return `kInvalidIndex` when IE is not present. It's possible that no IE is in an enh ack, So I use `SkipSecurityHeaderIndex` instead.

### DIFF
--- a/src/core/mac/mac_frame.cpp
+++ b/src/core/mac/mac_frame.cpp
@@ -1165,16 +1165,20 @@ otError TxFrame::GenerateEnhAck(const RxFrame &aFrame, bool aIsFramePending, con
         SetDstAddr(address);
     }
 
-    // Set security header
-    SuccessOrExit(error = aFrame.GetSecurityControlField(securityControlField));
-    SuccessOrExit(error = aFrame.GetFrameCounter(frameCounter));
-    SuccessOrExit(error = aFrame.GetKeyId(keyId));
-
     SetPsduLength(kMaxPsduSize); // At this time the length of ACK hasn't been determined, set it to
                                  // `kMaxPsduSize` to call methods that check frame length.
-    SetSecurityControlField(securityControlField);
-    SetFrameCounter(frameCounter);
-    SetKeyId(keyId);
+
+    // Set security header
+    if (aFrame.GetSecurityEnabled())
+    {
+        SuccessOrExit(error = aFrame.GetSecurityControlField(securityControlField));
+        SuccessOrExit(error = aFrame.GetFrameCounter(frameCounter));
+        SuccessOrExit(error = aFrame.GetKeyId(keyId));
+
+        SetSecurityControlField(securityControlField);
+        SetFrameCounter(frameCounter);
+        SetKeyId(keyId);
+    }
 
     // Set header IE
     if (aIeLength > 0)
@@ -1186,7 +1190,7 @@ otError TxFrame::GenerateEnhAck(const RxFrame &aFrame, bool aIsFramePending, con
     // Set frame length
     footerLength = GetFooterLength();
     OT_ASSERT(footerLength != kInvalidIndex);
-    mLength = FindHeaderIeIndex() + aIeLength + GetFooterLength();
+    mLength = SkipSecurityHeaderIndex() + aIeLength + GetFooterLength();
 
 exit:
     return error;
--- a/tests/unit/test_mac_frame.cpp
+++ b/tests/unit/test_mac_frame.cpp
@@ -574,7 +574,7 @@ void TestMacFrameAckGeneration(void)
     uint8_t     ie_data[6] = {0x04, 0x0d, 0x21, 0x0c, 0x35, 0x0c};
     Mac::CslIe *csl;
 
-    ackFrame.GenerateEnhAck(receivedFrame, false, ie_data, sizeof(ie_data));
+    IgnoreError(ackFrame.GenerateEnhAck(receivedFrame, false, ie_data, sizeof(ie_data)));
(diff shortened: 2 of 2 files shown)


src/core/mac/mac_frame.cpp
```diff
@@ -1165,16 +1165,20 @@ otError TxFrame::GenerateEnhAck(const RxFrame &aFrame, bool aIsFramePending, con
...
-    SuccessOrExit(error = aFrame.GetKeyId(keyId));
-
     SetPsduLength(kMaxPsduSize); // At this time the length of ACK hasn't been determined, set it to
                                  // `kMaxPsduSize` to call methods that check frame length.
-    SetSecurityControlField(securityControlField);
-    SetFrameCounter(frameCounter);
-    SetKeyId(keyId);
+
+    // Set security header
+    if (aFrame.GetSecurityEnabled())
```
brianljt, line 1172: Per 802.15.4-2015 6.7.2

> If the Enh-Ack frame contains IEs and/or a Frame Payload and it is in responseto a secured frame, then the Enh-Ack frame shall be secured. If the Enh-Ack frame is in response to asecured frame and does not contain either IEs or a Frame Payload, then the Enh-Ack may be secured.

So only when Rx frame is a secured frame and the enh-ack contains IEs/Payloads, it needs to be secured. This code only check for Rx frame security. Are there any modification in Thread spec to specify this behavior? I.E, enh-ack should have same security with regard to the received frame?

src/core/mac/mac_frame.cpp
```diff
@@ -1165,16 +1165,20 @@ otError TxFrame::GenerateEnhAck(const RxFrame &aFrame, bool aIsFramePending, con
...
+    {
+        SuccessOrExit(error = aFrame.GetSecurityControlField(securityControlField));
+        SuccessOrExit(error = aFrame.GetFrameCounter(frameCounter));
```
brianljt, line 1175: For enh-ack, the sequence number should be the one from received frame. But the Auxiliary header frame counter should be increased monotonously. Any reason here you use the counter from received frame? I think it breaks the criterial that each key material should be used only once per frame.

src/core/mac/mac_frame.cpp
Irving-cl: Thread spec doesn't specify this. Actually I implemented this referring to the enh ack generation implemented in nrf radio driver. But I think it doesn't violate the spec to always secure the Enh-Ack when the Rx frame is secured no matter if the Enh-Ack contains IE or Payload. Will this cause any other problems?

brianljt: From the 15.4 spec, I think `may be secured` means we can always use secured enh-ACK for received secure 15.4-2015 frame. But I'm considering something looking paradoxical to me.

15.4-2015 specify that retransmission should be using the exactly the same frame

> 6.7.4.3 When not using TSCH mode and a frame with the Security Enabled field set to one is retransmitted, the frame shall be retransmitted without changes and without passing through the outgoing frame security procedure, as defined in 9.2.1.

However, there are cases when this limitation does not hold true. For example, the CSL IE should be updated so that the CSL phase is correct.

Another case is, when doing retransmission, there are backoff between retransmission. What if peer send a secured 2015 frame to this node and this node response with an Enh-ACK with auxiliary frame counter increased. Then the old frame counter in the retransmission is not valid any more?

src/core/mac/mac_frame.cpp
Irving-cl: It should use the device's own frame counter. I will change it after #4976 gets merged since it updates a new way to get the frame counter.

src/core/mac/mac_frame.cpp
brianljt: As discussed. It is OK to always secure enh-ack for received secure 15.4-2015 frame. Set to resolved.

src/core/mac/mac_frame.cpp
brianljt: Thanks. Let's fix this after #4976 got merged.

tests/unit/test_mac_frame.cpp
```diff
@@ -574,7 +574,7 @@ void TestMacFrameAckGeneration(void)
     uint8_t     ie_data[6] = {0x04, 0x0d, 0x21, 0x0c, 0x35, 0x0c};
     Mac::CslIe *csl;
 
-    ackFrame.GenerateEnhAck(receivedFrame, false, ie_data, sizeof(ie_data));
+    IgnoreError(ackFrame.GenerateEnhAck(receivedFrame, false, ie_data, sizeof(ie_data)));
```
jwhui, line 577: Why wasn't this captured by continuous integration?

Irving-cl: https://github.com/openthread/openthread/blob/master/include/openthread/platform/toolchain.h#L72
I think the reason is that `IgnoreError` only works for `clang`. And it seems that in the matrices there's no building using `clang` with `THREAD_VERSION=1.2` and this check is wrapped in `THREAD_VERSION=1.2`.

jwhui: Makes sense. Can we add a clang build for Thread 1.2 (can be in a separate PR)?

Irving-cl: Yes, just plan to do so. I'll do it.

PR #4993, opened 2020-05-21
[continuous-integration] add item to build and run 1_2 using clang

Irving-cl:
Add a check to build and run 1_2 using clang.

### DIFF
--- a/.github/workflows/simulation.yml
+++ b/.github/workflows/simulation.yml
@@ -430,14 +430,21 @@ jobs:
       uses: codecov/codecov-action@v1
 
   thread-1-2:
+    name: thread-1-2-${{ matrix.compiler.c }}-${{ matrix.arch }}
     runs-on: ubuntu-18.04
     env:
-      CFLAGS: -m32
-      CXXFLAGS: -m32
-      LDFLAGS: -m32
+      CFLAGS: -${{ matrix.arch }}
+      CXXFLAGS: -${{ matrix.arch }}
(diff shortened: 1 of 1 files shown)


.github/workflows/simulation.yml
```diff
@@ -427,3 +427,30 @@ jobs:
         ./script/test cert_suite tests/scripts/thread-cert/v1_2_*
     - name: Codecov
       uses: codecov/codecov-action@v1
+
+  thread-1-2-clang:
```
bukepo, line 431: use matrix instead?

.github/workflows/simulation.yml
```diff
@@ -430,14 +430,32 @@ jobs:
...
+      CFLAGS: -${{ matrix.arch_flags }}
+      CXXFLAGS: -${{ matrix.arch_flags }}
+      LDFLAGS: -${{ matrix.arch_flags }}
       COVERAGE: 1
       THREAD_VERSION: 1.2
       VIRTUAL_TIME: 1
+      CC: ${{ matrix.c_compiler }}
+      CXX: ${{ matrix.cpp_compiler }}
+    strategy:
+      matrix:
```
bukepo, line 445: Does the following syntax work?
```suggestion
      matrix:
        compiler: [{c: 'gcc', cxx: 'g++'}, { c: 'clang', cxx: 'clang++'}],
        arch: ['m32', 'm64']
```

PR #4994, opened 2020-05-21
[posix] tolerate SIGPIPE under daemon mode

gjc13:
Chances are that the ot-ctl client exits between read error check and
write. This will lead to a SIGPIPE crash. We can ignore this signal and
check the error code instead.

### DIFF
--- a/src/posix/platform/uart.cpp
+++ b/src/posix/platform/uart.cpp
@@ -297,7 +297,12 @@ void platformUartProcess(const fd_set *aReadFdSet, const fd_set *aWriteFdSet, co
 
     if ((sWriteLength > 0) && (FD_ISSET(fd, aWriteFdSet)))
     {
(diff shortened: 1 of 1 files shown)


src/posix/platform/uart.cpp
```diff
@@ -297,7 +297,12 @@ void platformUartProcess(const fd_set *aReadFdSet, const fd_set *aWriteFdSet, co
 
     if ((sWriteLength > 0) && (FD_ISSET(fd, aWriteFdSet)))
     {
+#if OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE
+        // Don't die on SIGPIPE
+        rval = send(fd, sWriteBuffer, sWriteLength, MSG_NOSIGNAL);
```
simonlingoogle, line 302: ignore signal SIGPIPE?

bukepo: ignore a single is process-wide, I don't think it proper for a library.

PR #4996, opened 2020-05-23
[posix] some platforms don't support MSG_NOSIGNAL

rob-the-dude:
On platforms that have SO_NOSIGPIPE, use that to achieve the same result.

On platforms that have neither, emit a compile-time warning -- you may
still receive the signal on those systems.

### DIFF
--- a/src/posix/platform/uart.cpp
+++ b/src/posix/platform/uart.cpp
@@ -299,10 +299,27 @@ void platformUartProcess(const fd_set *aReadFdSet, const fd_set *aWriteFdSet, co
     {
 #if OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE
         // Don't die on SIGPIPE
+#if !defined(MSG_NOSIGNAL)
+        // some platforms (mac OS, Solaris) don't have MSG_NOSIGNAL
+        // SOME of those (mac OS, but NOT Solaris) support SO_NOSIGPIPE
+        // if we have SO_NOSIGPIPE, then set it.  otherwise, we're going
+        // to simply ignore it.
+#if defined(SO_NOSIGPIPE)
+        int err;
+        int flag;
+
+        flag = 1;
+        err  = setsockopt(fd, SOL_SOCKET, SO_NOSIGPIPE, &flag, sizeof(flag));
+        VerifyOrDie(err == 0, OT_EXIT_ERROR_ERRNO);
+#else
+#warning "no support for MSG_NOSIGNAL or SO_NOSIGPIPE"
+#endif
+#define MSG_NOSIGNAL 0
+#endif // !defined(MSG_NOSIGNAL)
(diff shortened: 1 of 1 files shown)


src/posix/platform/uart.cpp
```diff
@@ -299,10 +299,27 @@ void platformUartProcess(const fd_set *aReadFdSet, const fd_set *aWriteFdSet, co
...
+        // SOME of those (mac OS, but NOT Solaris) support SO_NOSIGPIPE
+        // if we have SO_NOSIGPIPE, then set it.  otherwise, we're going
+        // to simply ignore it.
+#if defined(SO_NOSIGPIPE)
+        int err;
+        int flag;
+
+        flag = 1;
+        err  = setsockopt(fd, SOL_SOCKET, SO_NOSIGPIPE, &flag, sizeof(flag));
+        VerifyOrDie(err == 0, OT_EXIT_ERROR_ERRNO);
```
gjc13, line 313: Shall we die on this error?

src/posix/platform/uart.cpp
```diff
@@ -299,10 +299,27 @@ void platformUartProcess(const fd_set *aReadFdSet, const fd_set *aWriteFdSet, co
...
+#else
+#warning "no support for MSG_NOSIGNAL or SO_NOSIGPIPE"
+#endif
+#define MSG_NOSIGNAL 0
```
gjc13, line 317: If we have neither `MSG_NOSIGNAL` nor `SO_NOSIGPIPE`, shall we fallback to `write` rather than manually define `MSG_NOSIGNAL`?

src/posix/platform/uart.cpp
```diff
@@ -299,10 +299,27 @@ void platformUartProcess(const fd_set *aReadFdSet, const fd_set *aWriteFdSet, co
     {
 #if OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE
         // Don't die on SIGPIPE
+#if !defined(MSG_NOSIGNAL)
```
gjc13, line 302: `#ifdef` `#ifndef`?

src/posix/platform/uart.cpp
rob-the-dude: Not sure if there's an advantage -- send with flags set to zero is basically the same as write.  Either way, that particular platform is doing something different than ones we normally use (hence the warning so the developer notices).  I don't have a strong preference.

src/posix/platform/uart.cpp
rob-the-dude: Is there a style preference for #ifdef/#ifndef when there's only a single check?  I somewhat prefer "defined" because it's usable in ALL situations, regardless of how many things you're checking, but I don't care, especially if there's a documented preference in OpenThread.

PR #4999, opened 2020-05-25
[low-power] encapsulate method to update header IE

Irving-cl:
OpenThread now have **TIME_SYNC** IE. And the presence of IE would also effect the frame version and IE present field. This PR encapsulate the process to update these things so that it could be easily extended to support other IEs that would be introduced later.

### DIFF
--- a/src/core/common/message.hpp
+++ b/src/core/common/message.hpp
@@ -856,6 +856,7 @@ public:
 
     /**
      * This method indicates whether or not the message is also used for time sync purpose.
+     *
      * When OPENTHREAD_CONFIG_TIME_SYNC_ENABLE is 0, this method always return false.
      *
      * @retval TRUE   If the message is also used for time sync purpose.
--- a/src/core/common/message.hpp
+++ b/src/core/common/message.hpp
@@ -854,16 +854,17 @@ public:
         return (GetMetadata().mInPriorityQ) ? GetMetadata().mQueue.mPriority : NULL;
     }
 
-#if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
     /**
      * This method indicates whether or not the message is also used for time sync purpose.
+     * When OPENTHREAD_CONFIG_TIME_SYNC_ENABLE is 0, this method always return false.
      *
      * @retval TRUE   If the message is also used for time sync purpose.
      * @retval FALSE  If the message is not used for time sync purpose.
      *
      */
-    bool IsTimeSync(void) const { return GetMetadata().mTimeSync; }
+    bool IsTimeSync(void) const;
 
+#if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
     /**
      * This method sets whether or not the message is also used for time sync purpose.
      *
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -832,8 +832,8 @@ otError Mac::PrepareDataRequest(TxFrame &aFrame)
     SuccessOrExit(error = Get<DataPollSender>().GetPollDestinationAddress(dst));
     VerifyOrExit(!dst.IsNone(), error = OT_ERROR_ABORT);
 
-    fcf = Frame::kFcfFrameMacCmd | Frame::kFcfPanidCompression | Frame::kFcfFrameVersion2006 | Frame::kFcfAckRequest |
-          Frame::kFcfSecurityEnabled;
+    fcf = Frame::kFcfFrameMacCmd | Frame::kFcfPanidCompression | Frame::kFcfAckRequest | Frame::kFcfSecurityEnabled;
+    UpdateFrameControlField(NULL, fcf);
 
     if (dst.IsExtended())
     {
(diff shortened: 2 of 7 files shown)


src/core/mac/mac.hpp
```diff
@@ -663,6 +664,33 @@ class Mac : public InstanceLocator
...
+     * This method updates frame version. If the frame would contain header IEs,
+     * IE present field would be set. If this is a csl transmission frame or
+     * header IE is present in this frame, the version should be set to 2015.
+     * Otherwise, the version would be set to 2006.
+     *
+     * @param[in]   aMessage   A pointer to the message to send, could be `NULL`.
+     * @param[out]  aFcf       A reference to the frame control field to set.
+     *
+     */
+    void UpdateFrameControlField(Message *aMessage, uint16_t &aFcf) const;
```
bukepo, line 692: static?

src/core/mac/mac.cpp
```diff
@@ -2061,5 +2061,69 @@ uint8_t Mac::GetTimeIeOffset(const Frame &aFrame)
 }
 #endif
 
+#if OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT
+otError Mac::AppendHeaderIe(Message *aMessage, TxFrame &aFrame) const
```
abtink, line 2065: I would suggest not using `Message *aMessage` as input for these new methods
- `Mac` generally deals with `Frame` and is not aware of `Message` (which are handled by next layer `MeshForwarder`).
- Seems like `aMessage` is only used to detect if it is time-sync or not `aMessage->IsTimeSync()`.

How about adding a `bool aIsTimeSync` as input parameter to this method instead?

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -349,11 +349,16 @@ void MeshForwarder::RemoveDataResponseMessages(void)
 void MeshForwarder::SendMesh(Message &aMessage, Mac::TxFrame &aFrame)
 {
     uint16_t fcf;
+#if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
+    bool timeSync = aMessage.IsTimeSync();
```
bukepo, line 353: how about make Message::IsTimeSync() always return false when OPENTHREAD_CONFIG_TIME_SYNC_ENABLE is not 1.

src/core/mac/mac.hpp
```diff
@@ -663,6 +663,33 @@ class Mac : public InstanceLocator
...
+     *
+     * @param[in]      aIsTimeSync  A boolean indicates if time sync is being used.
+     * @param[in,out]  aFrame       A reference to the TX-frame to which the IEs will be appended.
+     *
+     * @retval OT_ERROR_NONE    If append header IEs successfully.
+     * @retval OT_ERROR_FAILED  If cannot find header IE position in the frame.
+     *
+     */
+    otError AppendHeaderIe(bool aIsTimeSync, TxFrame &aFrame) const;
+#endif // OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT
```
abtink, line 679:
```suggestion
#endif 
```
Taking advantage of style guide 20 line limit   :)

>  - All `#endif` directives equal to or greater than 20 lines away from the `#if` or `#ifdef` directive to which they are related shall be decorated by language comment indicating the conditional they are associated with.

src/core/mac/mac.cpp
```diff
@@ -832,8 +832,8 @@ otError Mac::PrepareDataRequest(TxFrame &aFrame)
     SuccessOrExit(error = Get<DataPollSender>().GetPollDestinationAddress(dst));
     VerifyOrExit(!dst.IsNone(), error = OT_ERROR_ABORT);
 
-    fcf = Frame::kFcfFrameMacCmd | Frame::kFcfPanidCompression | Frame::kFcfFrameVersion2006 | Frame::kFcfAckRequest |
-          Frame::kFcfSecurityEnabled;
+    fcf = Frame::kFcfFrameMacCmd | Frame::kFcfPanidCompression | Frame::kFcfAckRequest | Frame::kFcfSecurityEnabled;
+    UpdateFrameControlField(false, fcf);
```
abtink, line 836:
```suggestion
    UpdateFrameControlField(/* aIsTimeSync */ false, fcf);
```

jwhui: Alternatively, should `aIsTimeSync` be an enum rather than bool? I assume we are preparing for additional header IE types?

Irving-cl: For now this method is intended to call once to determine the IE present field and frame version when preparing a new frame, so it needs to take all the information about what IEs would be included as input. In predicable future, we would have CSL IE and whether to have it would be inferred from information in mac(so no other input parameters like `aIsTimeSync` is needed). However, if someday we have another new IE and we may probably need to add another input parameter indicating if it would exist. So I think enum may not be used here.

src/core/common/message.hpp
```diff
@@ -854,16 +854,17 @@ class Message : public Buffer
         return (GetMetadata().mInPriorityQ) ? GetMetadata().mQueue.mPriority : NULL;
     }
 
-#if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
     /**
      * This method indicates whether or not the message is also used for time sync purpose.
+     * When OPENTHREAD_CONFIG_TIME_SYNC_ENABLE is 0, this method always return false.
```
jwhui, line 859: nit: add blank line between brief and description.

src/core/mac/mac.hpp
```diff
@@ -663,6 +663,33 @@ class Mac : public InstanceLocator
...
+     *
+     * @retval OT_ERROR_NONE    If append header IEs successfully.
+     * @retval OT_ERROR_FAILED  If cannot find header IE position in the frame.
+     *
+     */
+    otError AppendHeaderIe(bool aIsTimeSync, TxFrame &aFrame) const;
+#endif
+
+    /**
+     * This method updates frame control field. If the frame would contain header IEs,
```
jwhui, line 682: nit: add blank line between brief and description.

src/core/mac/mac.hpp
```diff
@@ -663,6 +663,33 @@ class Mac : public InstanceLocator
...
+#if OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT
+    /**
+     * This method appends header IEs to a TX-frame according to its
+     * frame control field and if time sync is enabled.
+     *
+     * @param[in]      aIsTimeSync  A boolean indicates if time sync is being used.
+     * @param[in,out]  aFrame       A reference to the TX-frame to which the IEs will be appended.
+     *
+     * @retval OT_ERROR_NONE    If append header IEs successfully.
+     * @retval OT_ERROR_FAILED  If cannot find header IE position in the frame.
```
jwhui, line 675: nit: I generally prefer to avoid generic error like `OT_ERROR_FAILED`. Should this be `OT_ERROR_NOT_FOUND` instead?

src/core/mac/mac.hpp
```diff
@@ -663,6 +663,33 @@ class Mac : public InstanceLocator
...
+     *
+     */
+    otError AppendHeaderIe(bool aIsTimeSync, TxFrame &aFrame) const;
```
jwhui, line 678: Declare as static?

PR #5000, opened 2020-05-25
[posix] get node id from command line argument

bukepo:
This simplify manually running posix virtual time mode when debugging.

### DIFF
--- a/src/posix/platform/system.cpp
+++ b/src/posix/platform/system.cpp
@@ -49,7 +49,7 @@ otInstance *otSysInit(otPlatformConfig *aPlatformConfig)
     otInstance *instance = NULL;
 
 #if OPENTHREAD_POSIX_VIRTUAL_TIME
-    virtualTimeInit();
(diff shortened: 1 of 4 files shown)


src/posix/platform/system.cpp
```diff
@@ -49,7 +49,7 @@ otInstance *otSysInit(otPlatformConfig *aPlatformConfig)
     otInstance *instance = NULL;
 
 #if OPENTHREAD_POSIX_VIRTUAL_TIME
-    virtualTimeInit();
+    virtualTimeInit(static_cast<uint16_t>(atoi(aPlatformConfig->mRadioConfig)));
```
rob-the-dude, line 52: This looks weird to me.  I went back, and I guess it makes sense, but it doesn't seem clearly documented that "mRadioConfig" is the node ID (or at least, it's a white-space delineated list that starts with node ID).  I feel like that ought to be made more clear (perhaps even renamed), although I don't think the code here is "wrong".

PR #5003, opened 2020-05-25
[cmake] add ot-config

bukepo:
This commit add an interface target `ot-config` for collecting compile definitions.


etc/cmake/options.cmake
```diff
@@ -32,228 +32,223 @@ option(OT_APP_RCP "enable RCP app" ON)
...
 endif()
 
 option(OT_PLATFORM_NETIF "enable platform netif support")
 if(OT_PLATFORM_NETIF)
-    list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE=1")
-endif()
-
-option(OT_NCP_VENDOR_HOOK_SOURCE "Enable vendor hook for NCP")
-if(OT_NCP_VENDOR_HOOK_SOURCE)
-    list(APPEND OT_PRIVATE_DEFINES "OPENTHREAD_ENABLE_NCP_VENDOR_HOOK=1")
```
Irving-cl, line 190: Is it intentional to remove the `OPENTHREAD_ENABLE_NCP_VENDOR_HOOK` definition?

bukepo: yes

third_party/mbedtls/CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ target_compile_definitions(mbedtls
     PUBLIC
         "MBEDTLS_CONFIG_FILE=$<IF:$<BOOL:${OT_MBEDTLS_CONFIG_FILE}>,${OT_MBEDTLS_CONFIG_FILE},${OT_MBEDTLS_DEFAULT_CONFIG_FILE}>"
     PRIVATE
-        ${OT_PRIVATE_DEFINES}
+        $<TARGET_PROPERTY:ot-config,INTERFACE_COMPILE_DEFINITIONS>
```
wgtdkp, line 42: Why don't we use `target_link_libraries`?

bukepo: I forgot to mention, the mbedtls is using an old cmake version, which doesn't support this yet.

PR #5006, opened 2020-05-27
[docker] remove wpantund from environment image

jwhui:
To help reduce processing times in Docker hub continuous-integration build checks.


etc/docker/environment/Dockerfile
```diff
@@ -2,26 +2,30 @@
 FROM ubuntu:18.04
 
 ENV DEBIAN_FRONTEND noninteractive
+ENV LANG en_US.utf8
```
bukepo, line 5:
```suggestion
ENV LANG en_US.UTF-8
```

PR #5007, opened 2020-05-27
[logging] add newline to the end of NCP logs

simonlingoogle:
This PR adds `\n` to the end of NCP logs so that `wpantund` can print each log line properly.

### DIFF
--- a/src/ncp/ncp_base.cpp
+++ b/src/ncp/ncp_base.cpp
@@ -2409,6 +2409,8 @@ extern "C" void otNcpPlatLogv(otLogLevel aLogLevel, otLogRegion aLogRegion, cons
             charsWritten = static_cast<int>(sizeof(logString) - 1);
         }
 
+        logString[charsWritten++] = '\n';
+
         IgnoreError(otNcpStreamWrite(0, reinterpret_cast<uint8_t *>(logString), charsWritten));
     }
 }


src/ncp/ncp_base.cpp
```diff
@@ -2409,6 +2409,8 @@ extern "C" void otNcpPlatLogv(otLogLevel aLogLevel, otLogRegion aLogRegion, cons
             charsWritten = static_cast<int>(sizeof(logString) - 1);
         }
 
+        logString[charsWritten++] = '\n';
```
abtink, line 2412: @simonlingoogle 

We have `OPENTHREAD_CONFIG_LOG_SUFFIX` config which can be used for this purpose (a suffix that to be added to end of all log messages by the logging modules/macros).
https://github.com/openthread/openthread/blob/75a382d292b2271c0a6e202324550091753467f0/src/core/config/logging.h#L291-L299

The OpenThread NCP build may not be used with `wpantund` (e.g. it may be used by an RTOS-based driver) and always including `\n` may not be desired by all host drivers, so I won't recommend adding it here.

Also if you are using NCP build with `wpantund` I would highly suggest using the config
`OPENTHREAD_CONFIG_LOG_OUTPUT` to `OPENTHREAD_CONFIG_LOG_OUTPUT_NCP_SPINEL`.

simonlingoogle: Thanks for your suggestions. This PR is not thoughtful enough.

PR #5008, opened 2020-05-27
 [tlvs] add ReadTlv (ReadUint8Tlv, etc) to read a TLV from a message at given offset

abtink:
This PR contains 3 related commits:

**[tlvs] rename helper methods to FindTlv (FindUint8Tlv, etc)**

This commit renames the helper `static` methods in `Tlv` class. It
harmonizes the naming model to use `FindTlv()` for different flavors
of `Tlv` methods which search within a given `Message`/bufer to find a
TLV. It also removes the re-definition of same static methods in `Tlv`
sub-classes. This helps simplify the use of such methods and avoids
the shadowing of methods due to re-definition in the sub-class (the
sub-class re-definition allows more specific TLV `Type` enumeration
parameter to be used, however the base case version will continue to
work due to implicit conversion of `enum` to integer).

**[tlvs] change Tlv::AppendTlv value param to be `void *`**

This commit changes the `Tlv::AppendTlv()` value parameter to use 
`void *` (which simplify its use). This commit also updates the use of 
this method from different modules.

**[tlvs] add ReadTlv (ReadUint8Tlv, etc) helper methods**

This commit adds new helper static method in `Tlv` class to read a
(simple) TLV in a `Message` at a given offset with TLV's value being
an `uint` type or a fixed length data blob. These methods are use
in MLE and network-diagnostics modules and also internally by other
TLV helper methods.

### DIFF
--- a/src/core/common/tlvs.hpp
+++ b/src/core/common/tlvs.hpp
@@ -347,7 +347,7 @@ public:
      * @retval OT_ERROR_NO_BUFS  Insufficient available buffers to grow the message.
      *
      */
-    static otError AppendTlv(Message &aMessage, uint8_t aType, const uint8_t *aValue, uint8_t aLength);
+    static otError AppendTlv(Message &aMessage, uint8_t aType, const void *aValue, uint8_t aLength);
 
 protected:
     enum
--- a/src/core/common/tlvs.hpp
+++ b/src/core/common/tlvs.hpp
@@ -191,7 +191,7 @@ public:
      * @retval OT_ERROR_NOT_FOUND  Could not find the TLV with Type @p aType.
      *
      */
-    static otError Get(const Message &aMessage, uint8_t aType, uint16_t aMaxSize, Tlv &aTlv);
+    static otError FindTlv(const Message &aMessage, uint8_t aType, uint16_t aMaxSize, Tlv &aTlv);
 
     /**
      * This static method obtains the offset of a TLV within @p aMessage.
@@ -206,7 +206,7 @@ public:
      * @retval OT_ERROR_NOT_FOUND  Could not find the TLV with Type @p aType.
      *
      */
-    static otError GetOffset(const Message &aMessage, uint8_t aType, uint16_t &aOffset);
+    static otError FindTlvOffset(const Message &aMessage, uint8_t aType, uint16_t &aOffset);
 
     /**
      * This static method finds the offset and length of a given TLV type.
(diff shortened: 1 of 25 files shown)


src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -158,26 +158,34 @@ class Tlv : public ot::Tlv
...
+     * an expected minimum value, and then reads its value into a given buffer.
      *
-     * @param[in]   aMessage    A reference to the message.
-     * @param[in]   aType       The Type value to search for.
-     * @param[out]  aOffset     The offset where the value starts.
-     * @param[out]  aLength     The length of the value.
+     * If the TLV length is smaller than the minimum length @p aLength, the TLV is considered invalid. In this case,
+     * this method returns `OT_ERROR_PARSE` and the @p aValue buffer is not updated.
+     *
+     * If the TLV is length is larger than @p aLength, the TLV is considered valid, but only the @aLength first bytes
```
jwhui, line 173:
```suggestion
     * If the TLV length is larger than @p aLength, the TLV is considered valid, but only the @p aLength first bytes
```

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -158,26 +158,34 @@ class Tlv : public ot::Tlv
...
+     * of the value are read and copied into the @p aValue buffer.
+     *
+     * @param[in]    aMessage    A reference to the message.
+     * @param[in]    aType       The TLV type to search for.
+     * @param[out]   aValue      A buffer to output the value (must contain at least @p aLength bytes).
+     * @param[in]    aLength     The expected (minimum) length of the TLV value.
      *
-     * @retval OT_ERROR_NONE       Successfully found the TLV.
+     * @retval OT_ERROR_NONE       The TLV was found and read successfully. @p @aValue is updated.
```
jwhui, line 181:
```suggestion
     * @retval OT_ERROR_NONE       The TLV was found and read successfully. @p aValue is updated.
```

src/core/common/tlvs.hpp
```diff
@@ -177,6 +176,59 @@ class Tlv
...
+     *
+     */
+    static otError ReadUint32Tlv(const Message &aMessage, uint16_t aOffset, uint32_t &aValue);
+
+    /**
+     * This static method reads a TLV in a message at a given offset expecting a minimum length for the value.
+     *
+     * @param[in]   aMessage    The message to read from.
+     * @param[in]   aOffset     The offset into the message pointing to the start of the TLV.
+     * @param[out]  aValue      A buffer to output the TLV's value, must contains (at least) @p aMinLength bytes.
```
jwhui, line 223:
```suggestion
     * @param[out]  aValue      A buffer to output the TLV's value, must contain (at least) @p aMinLength bytes.
```

PR #5009, opened 2020-05-27
[bootstrap] remove PPA on Ubuntu 20.04 and later

dsyx:
My host:

```
Distributor ID: Ubuntu
Description:    Ubuntu 20.04 LTS
Release:        20.04
Codename:       focal
```

The PPA source does not include the package `gcc-arm-embedded` for Ubuntu 20.04.
In fact, users can directly use the toolchain provided by official sources.

Has been tested and available on my machine.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -40,9 +40,15 @@ install_packages_apt()
     sudo apt-get --no-install-recommends install -y automake g++ libtool lsb-release make cmake ninja-build
 
     PLATFORM=$(lsb_release -is)
+    RELEASE=$(lsb_release -rs)
+    UBUNTU2004=20.04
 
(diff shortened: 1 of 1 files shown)


dsyx:
@googlebot I signed it!

dsyx:
I am confused about `script/check-size`. Since the ARM toolchain is already configured at `script/bootstrap`, why does the `script/check-size` need to pull another ARM toolchain?

bukepo:
> I am confused about `script/check-size`. Since the ARM toolchain is already configured at `script/bootstrap`, why does the `script/check-size` need to pull another ARM toolchain?

Using a fixed version so that the size change will not be affected by different tool chain versions.

PR #5011, opened 2020-05-28
[ot-ctl] fix reset in non-interactive mode

bukepo:
* set close on exec
* explicitly close socket on reset
* enhance ot-ctl to exit on peer close

### DIFF
--- a/src/posix/client.cpp
+++ b/src/posix/client.cpp
@@ -76,74 +76,74 @@ static void InputCallback(char *aLine)
 }
 #endif // OPENTHREAD_USE_READLINE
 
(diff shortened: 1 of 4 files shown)


src/posix/client.cpp
```diff
@@ -170,41 +170,23 @@ static bool DoWrite(int aFile, const void *aBuffer, size_t aSize)
 
 static void SendBlockingCommand(int aArgc, char *aArgv[])
```
jwhui, line 171: nit: Is this method name still appropriate given that we are no longer waiting for a response?

PR #5012, opened 2020-05-28
[api] add a monotonic version number

bukepo:


include/openthread/instance.h
```diff
@@ -45,6 +45,14 @@
 extern "C" {
 #endif
 
+/**
+ * The OpenThread API monotonic version number.
+ *
+ * The number should be increased everytime the OpenThread API changes.
+ *
+ */
+#define OPENTHREAD_API_VERSION (1)
```
abtink, line 54: LGTM.
Just to clarify the rules around this:
- We increment it on any change to public OT APIs 
  - Addition/removal or parameter/return type changes on functions
  - Also on any OT public struct/type definition changes (e.g. new fields in a struct, change type of a struct field, , removing a field, etc).
- Do we want to increment this version for changes to OT platform layer APIs? (I would suggest no, maybe a different version for that later?).

Thoughts?

include/openthread/instance.h
```diff
@@ -45,6 +45,14 @@
 extern "C" {
 #endif
 
+/**
+ * The OpenThread API monotonic version number.
+ *
+ * The number should be increased everytime the OpenThread API changes.
```
simonlingoogle, line 51: how can we make sure it's not violated?

include/openthread/instance.h
bukepo: Agree! I actually expect an increment on any changes to the public header files. No matter the definition or just documentation. After this is merged, I can add a check to verify the version number is increased in a PR which changed public header files.

Thoughts?

include/openthread/instance.h
wgtdkp: what about adding a GitHub Action check or git hook to verify that this version number is increased if the PR includes a change in `include/openthread`?

include/openthread/instance.h
abtink: Sounds good to me.
Also great idea on adding a check to ensure version is increased in any PR that changes any of public header files. 👍

jwhui: > I can add a check to verify the version number is increased in a PR which changed public header files.

Would this check include any change? Or just change to function declarations, types, etc.? Would we require a change on doc typos, for example?

simonlingoogle: what if there are 2 PRs both changes version from 1 to 2. They will both pass the checks. When one PR is merged, will the other PR be conflicted?  If not, the API version won't be correctly changed to 3.

bukepo: I was planning to include any changes. It seems not easy to just detect real changes. And if we treat documentation also a part of the API, the number should be increased.

include/openthread/instance.h
```diff
@@ -45,6 +45,14 @@
 extern "C" {
 #endif
 
+/**
+ * The OpenThread API monotonic version number.
+ *
+ * The number SHALL be increased by one everytime the content of OpenThread API files changed.
```
jwhui, line 51:
```suggestion
 * This number MUST increase by one each time any content of public OpenThread API files changes.
```

include/openthread/instance.h
jwhui: Does this version number cover both user and platform APIs? We probably want to add that clarification to the documentation as well.

bukepo: Yes, I think so.

bukepo: added a note below.

PR #5017, opened 2020-05-28
[test] add various CLI command tests

Moandor-y:
This adds `expect` tests for many CLI commands. Some simple commands are placed in `cli-misc.exp`.

### DIFF
--- a/tests/scripts/expect/cli-routereligible.exp
+++ b/tests/scripts/expect/cli-routereligible.exp
@@ -1,0 +1,54 @@
+#!/usr/bin/expect -f
+#
+#  Copyright (c) 2020, The OpenThread Authors.
+#  All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


tests/scripts/expect/cli-routereligible.exp
```diff
@@ -0,0 +1,47 @@
...
+send "routereligible\n"
+expect "Disabled"
+expect "Done"
+send "routereligible enable\n"
+expect "Done"
+send "routereligible\n"
+expect "Enabled"
+expect "Done"
+send "routereligible something_invalid\n"
+expect "Error 7: InvalidArgs"
```
jwhui, line 47: How about adding a test when mode is set to MTD? In this case, the routereligible command should return `OT_ERROR_NOT_CAPABLE`.

Moandor-y: Done.

PR #5019, opened 2020-05-28
[ip6-address] add a type flag to addresses

markus-becker-tridonic-com:
Add a type flag to addresses so that applications are able to behave
differently on the type of address.

Signed-off-by: Markus Becker <markus.becker@tridonic.com>

### DIFF
--- a/include/openthread/ip6.h
+++ b/include/openthread/ip6.h
@@ -116,6 +116,18 @@ struct otIp6Prefix
 typedef struct otIp6Prefix otIp6Prefix;
 
 /**
+ * IPv6 Address origins
+ *
+ */
+enum
+{
+    OT_ADDRESS_ORIGIN_THREAD = 0, ///< Thread assigned address (ALOC, RLOC, MLEID, etc)
+    OT_ADDRESS_ORIGIN_SLAAC  = 1, ///< SLAAC assigned address
+    OT_ADDRESS_ORIGIN_DHCPV6 = 2, ///< DHCPv6 assigned address
+    OT_ADDRESS_ORIGIN_MANUAL = 3, ///< Manually assigned address
+};
+
+/**
  * This structure represents an IPv6 network interface unicast address.
  *
  */
@@ -123,6 +135,7 @@ typedef struct otNetifAddress
 {
     otIp6Address           mAddress;                ///< The IPv6 unicast address.
     uint8_t                mPrefixLength;           ///< The Prefix length (in bits).
+    uint8_t                mAddressOrigin;          ///< The IPv6 address origin.
     bool                   mPreferred : 1;          ///< TRUE if the address is preferred, FALSE otherwise.
     bool                   mValid : 1;              ///< TRUE if the address is valid, FALSE otherwise.
     bool                   mScopeOverrideValid : 1; ///< TRUE if the mScopeOverride value is valid, FALSE otherwise.
--- a/src/core/backbone_router/local.cpp
+++ b/src/core/backbone_router/local.cpp
@@ -62,6 +62,7 @@ Local::Local(Instance &aInstance)
     mBackboneRouterPrimaryAloc.Clear();
 
     mBackboneRouterPrimaryAloc.mPrefixLength       = Mle::MeshLocalPrefix::kLength;
+    mBackboneRouterPrimaryAloc.mAddressOrigin      = OT_ADDRESS_ORIGIN_THREAD;
     mBackboneRouterPrimaryAloc.mPreferred          = true;
     mBackboneRouterPrimaryAloc.mValid              = true;
     mBackboneRouterPrimaryAloc.mScopeOverride      = Ip6::Address::kRealmLocalScope;
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (4)
+#define OPENTHREAD_API_VERSION (5)
 
 /**
  * @addtogroup api-instance
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -1556,10 +1556,11 @@ otError Interpreter::ProcessIpAddrAdd(uint8_t aArgsLength, char *aArgs[])
     VerifyOrExit(aArgsLength > 0, error = OT_ERROR_INVALID_ARGS);
 
     SuccessOrExit(error = otIp6AddressFromString(aArgs[0], &aAddress.mAddress));
-    aAddress.mPrefixLength = 64;
-    aAddress.mPreferred    = true;
-    aAddress.mValid        = true;
-    error                  = otIp6AddUnicastAddress(mInstance, &aAddress);
+    aAddress.mPrefixLength  = 64;
+    aAddress.mPreferred     = true;
+    aAddress.mValid         = true;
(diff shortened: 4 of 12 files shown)


include/openthread/ip6.h
```diff
@@ -97,6 +97,21 @@ struct otIp6Prefix
...
+/**
+ * IPv6 Address types
+ *
+ */
+enum
+{
+    kLinkLocal = 0, ///< Link-local address
+    kSlaac     = 1, ///< SLAAC assigned address
+    kDhcpv6    = 2, ///< DHCPv6 assigned address
+    kManual    = 3, ///< Manually assigned address
```
simonlingoogle, line 109: I don't think these are all valid address types. 

SLAAC, DHCP, MANUAL are actually types of methods for generating addresses. 
LinkLocal, Aloc, MeshLocal, DUA are address types, but they can be generally determined by checking network prefixes or interface ids.

markus-becker-tridonic-com: How about calling the variable "mAddressSource", "mAddressMechnism", or "mAddressMethod" instead? And then have the values kLink, kSlaac, kDhcpv6, kManual, and possibly kMesh?

abtink: How about using `mAddressOrigin`?

include/openthread/ip6.h
```diff
@@ -97,6 +97,21 @@ struct otIp6Prefix
...
+    kAloc      = 4, ///< Aloc address
+    kMeshLocal = 5, ///< Mesh-local address
+    kDua       = 6, ///< DUA address
+};
```
abtink, line 113: Since this is public OT API/definition, we need to follow C style constants (capital case with `_`) and starting with prefix `OT_`. Thanks.

include/openthread/ip6.h
simonlingoogle: I think we should use proper checking for prefixes and locators to determine the address types. 
For the address origin, what about just defining four origin types (using only 2 bits):

```c++
    kSystem = 0; // System generated addresses including ALOCs, MLEID, RLOC, etc
    kSlaac = 1; // SLAAC
    kDhcpv6 = 2; // DHCPv6
    kManual = 3; // Manually configured
```

Thoughts? @markus-becker-tridonic-com @abtink

include/openthread/ip6.h
markus-becker-tridonic-com: Like this?
```
enum
{
    OT_ADDRESSORIGIN_SYSTEM = 0, ///< Address assigned by the system
    OT_ADDRESSORIGIN_SLAAC  = 1, ///< SLAAC assigned address
    OT_ADDRESSORIGIN_DHCPV6 = 2, ///< DHCPv6 assigned address
    OT_ADDRESSORIGIN_MANUAL = 3, ///< Manually assigned address
};
```

include/openthread/ip6.h
markus-becker-tridonic-com: > I think we should use proper checking for prefixes and locators to determine the address types.

Seemed pretty obvious which type/origin it is in the code. When in dhcp6_server.hpp, it is a DHCP address type/origin, etc.

include/openthread/ip6.h
abtink: Yes. Thanks. I would aslo suggest adding `_` in `ADDRESS_ORIGIN`?

markus-becker-tridonic-com: Reworked accordingly.

src/core/backbone_router/local.cpp
```diff
@@ -62,6 +62,7 @@ Local::Local(Instance &aInstance)
     mBackboneRouterPrimaryAloc.Clear();
 
     mBackboneRouterPrimaryAloc.mPrefixLength       = Mle::MeshLocalPrefix::kLength;
+    mBackboneRouterPrimaryAloc.mAddressOrigin      = OT_ADDRESS_ORIGIN_SYSTEM;
```
simonlingoogle, line 65: Assigning to `OT_ADDRESS_ORIGIN_SYSTEM` is actually unnecessary since `mAddressOrigin` is already 0 after cleared ?

include/openthread/ip6.h
```diff
@@ -105,6 +117,7 @@ typedef struct otNetifAddress
 {
     otIp6Address           mAddress;                ///< The IPv6 unicast address.
     uint8_t                mPrefixLength;           ///< The Prefix length (in bits).
+    uint8_t                mAddressOrigin;          ///< The IPv6 address origin.
```
simonlingoogle, line 120:
```suggestion
    uint8_t                mAddressOrigin: 2;          ///< The IPv6 address origin.
```

src/core/backbone_router/local.cpp
markus-becker-tridonic-com: Better assigning it explicitly than implicitly? E.g. if the enum is re-ordered, this might change the value unintentionally at this location.

include/openthread/ip6.h
markus-becker-tridonic-com: Would you change your suggestion to 
```suggestion
    uint8_t                mAddressOrigin: 2;       ///< The IPv6 address origin.
```
so that the alignment is correct and I can just apply it.

markus-becker-tridonic-com: nevermind. github allowed to modify the suggestion.

src/core/backbone_router/local.cpp
simonlingoogle: Inded explicitness is better.

include/openthread/ip6.h
```diff
@@ -105,6 +117,7 @@ typedef struct otNetifAddress
 {
     otIp6Address           mAddress;                ///< The IPv6 unicast address.
     uint8_t                mPrefixLength;           ///< The Prefix length (in bits).
+    uint8_t                mAddressOrigin: 2;       ///< The IPv6 address origin.
```
jwhui, line 120: Needed to pass pretty check:
```suggestion
    uint8_t                mAddressOrigin : 2;      ///< The IPv6 address origin.
```

include/openthread/ip6.h
```diff
@@ -97,6 +97,18 @@ struct otIp6Prefix
  */
 typedef struct otIp6Prefix otIp6Prefix;
 
+/**
+ * IPv6 Address origins
+ *
+ */
+enum
+{
+    OT_ADDRESS_ORIGIN_SYSTEM = 0, ///< Address assigned by the system
```
jwhui, line 106: nit: can we be more specific than "system"?
```suggestion
    OT_ADDRESS_ORIGIN_SYSTEM = 0, ///< Address assigned by Thread control protocols
```

markus-becker-tridonic-com: I'd prefer:
```suggestion
    OT_ADDRESS_ORIGIN_SYSTEM = 0, ///< Thread assigned address (ALOC, RLOC, MLEID, etc)

```

simonlingoogle: > nit: can we be more specific than "the system"?

I think `OT_ADDRESS_ORIGIN_THREAD` might be a more suitable name. Thoughts?

jwhui: @markus-becker-tridonic-com , "Thread assigned address" sounds good to me.

markus-becker-tridonic-com: done

markus-becker-tridonic-com:
i've addressed latest review comments and rebased to solve merge conflict because of API version change in master.

PR #5026, opened 2020-05-30
[ot-ctl] no output prompt in non-interactive mode

bukepo:


src/posix/client.cpp
```diff
@@ -76,6 +76,27 @@ static void InputCallback(char *aLine)
 }
 #endif // OPENTHREAD_USE_READLINE
 
+static bool FindPrompt(int &aState, char aChar)
+{
+    switch (aChar)
+    {
+    case '>':
+        aState = aState == 0 ? 1 : -1;
```
simonlingoogle, line 84: define state enum type?

bukepo: I tend to enhance this later because this needs a rewrite of multiple functions which is beyond the purpose of this PR.

simonlingoogle: I am fine to enhance later.

PR #5029, opened 2020-05-30
[core] adding Equatable and Clearable subclasses

abtink:
This PR contains two commits:

 **[core] add Equatable class providing overloads of `==` and `!=` operator**
    
 This commit adds new class `Equatable` (inherited by other types)
 providing overloads of operators `==` and `!=`. The implementation
 simply compares all the bytes of two instances of same type to be
 equal (using `memcmp`).

**[core] add Clearable class providing simple Clear() method**
    
This commit adds new class `Clearable` (inherited by other types)
providing simple implementation of `Clear()` which sets all the
instance bytes to zero (using `memset()`).

### DIFF
--- a/src/core/common/clearable.hpp
+++ b/src/core/common/clearable.hpp
@@ -1,0 +1,60 @@
+/*
+ *  Copyright (c) 2020, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes definitions for Clearable class for OpenThread objects.
+ */
+
+#ifndef CLEARABLE_HPP_
+#define CLEARABLE_HPP_
+
+#include "openthread-core-config.h"
+
+#include <string.h>
+
+namespace ot {
+
+/**
+ * This template class defines a Clearable object which provides `Clear()` method.
+ *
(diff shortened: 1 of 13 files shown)


src/core/common/clearable.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * The `Clear` implementation simply sets all the bytes of a `Type` instance to zero.
+ *
+ * Users of this class should follow CRTP-style inheritance, i.e., the `Type` class itself should publicly inherit
+ * from `Clearable<Type>`.
+ *
+ */
+template <typename Type> class Clearable
+{
+public:
+    void Clear(void) { memset(reinterpret_cast<void *>(this), 0, sizeof(Type)); }
```
Moandor-y, line 55:
```suggestion
    void Clear(void) { *this = Type(); }
```
Safer than `memset`.

src/core/common/equatable.hpp
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * This method overloads operator `==` to evaluate whether or not two instances of `Type` are equal.
+     *
+     * @param[in]  aOther  The other `Type` instance to compare with.
+     *
+     * @retval TRUE   If the two `Type` instances are equal.
+     * @retval FALSE  If the two `Type` instances are not equal.
+     *
+     */
+    bool operator==(const Type &aOther) const { return memcmp(this, &aOther, sizeof(Type)) == 0; }
```
Moandor-y, line 64: `memcmp` is typically only meaningful for trivially-copyable objects with no padding. For type `struct{char c; int n;}` it will compare the padding bytes whose values may differ when the values of `c` and `n` are the same.

src/core/common/clearable.hpp
simonlingoogle: If Type() calls `Clear` and `Clear` calls Type(), would that be a infinite loop?

Moandor-y: `Type()` shouldn't call `Clear`. Is there any case that this is required?

src/core/mac/channel_mask.hpp
```diff
@@ -62,7 +63,7 @@ namespace Mac {
  * It is a wrapper class around a `uint32_t` bit vector representing a set of channels.
  *
  */
-class ChannelMask
+class ChannelMask : public Equatable<ChannelMask>
```
simonlingoogle, line 66: skip `ChannelMask`? 
I think inheriting `Euquatable` makes `ChannelMask` less efficient.

src/core/common/time.hpp
```diff
@@ -55,7 +57,7 @@ namespace ot {
  * This class represents a time instance.
  *
  */
-class Time
+class Time : public Equatable<Time>
```
simonlingoogle, line 60: skip `Time` for efficiency?

src/core/common/clearable.hpp
simonlingoogle: see https://github.com/openthread/openthread/blob/8e69b5acb7003223f517dff2bc9ef720ac4c9ff8/src/core/net/socket.hpp#L215

Moandor-y: In this case the constructor can be changed to `SockAddr(void) : otSockAddr() {}` without changing any behaviors.
But these changes are not trivial. If it's impractical to change all such calls, I'm OK with the `memset` solution.

abtink: This PR is only refactoring existing code by adding a common sub-class to avoid repeating the same logic in many different classes. Existing code uses `memset` already.

Here are some reasons why I would say `memset` is preferred:
- I think `Type()` does ensure zero initialization since C++11 (https://en.cppreference.com/w/cpp/language/zero_initialization). Note that OT still only requires C++03.
- It adds a restriction for the `Type` to have a public default constructor `Type()` so that it can inherit from `Clearable` (unnecessarily additional requirement) and that the default constructor should set all fields to zero. 
- OT is network stack and therefore is being built with a lot of older tool-chains/compliers. An older complier (not performing good optimization) may treat `*this = Type()` to create a new instance and zero initialize it and then copy it to `*this` (which adds an extra copying). `memset` approach would not rely on compiler optimization.

src/core/common/equatable.hpp
abtink: Yes. That's clear.

The intention for `Equatable` is not to handle all situation. It should be used when we do want to check all bytes in the objects (e.g. when it is packed, etc) and `memcmp` makes sense. If code uses `Equatable` for struct as your example, then I would say it is incorrect use and actually a bug (such struct should probably define its own `==` overload if wants one).

This PR is only refactoring existing code (where we already use `memcmp`) by defining a new common base class so that we don't repeat the same code pattern in different classes.

src/core/mac/channel_mask.hpp
abtink: I was actually considering this. I checked the code size impact and noticed that it did not matter (which I did not expect to be honest). I think there are efficient `memcmp` implementations which handle small size differently so the complier may be optimizaing it already.

It just helps avoding the definitions (src code reduction), but I agree it may better/safer to not use it for `ChannelMask` and `Time`. Will change it.

src/core/common/time.hpp
abtink: Changed it in new pushed commits.

bukepo:
Annot sure why the code size increased.

bukepo:
Not sure why the code size increased.

abtink:
@bukepo I know (or better say have a theory) for the reason for the small code size increase. It's because previously in some types, we strategically defined `operator==` in `cpp` file rather in the header (to avoid it being inlined), but with this it will be always inlined. The increase is not really significant and practically we are trading off speed (inlining) vs size.

PR #5030, opened 2020-05-31
[test] cover micro timer in unittest

bukepo:


tests/unit/test_timer.cpp
```diff
@@ -83,11 +83,11 @@ void InitCounters(void)
  * `TestTimer` sub-classes `ot::TimerMilli` and provides a handler and a counter to keep track of number of times timer
  * gets fired.
  */
-class TestTimer : public ot::TimerMilli
+template <typename TimerType> class TestTimer : public TimerType
```
abtink, line 86: Nice! 👍

PR #5034, opened 2020-06-01
[continuous-integration] add version number check

bukepo:


script/check-api-version
```diff
@@ -0,0 +1,64 @@
...
+die()
+{
+    echo >&2 "ERROR: $*"
+    exit 1
+}
+
+mkdir -p tmp
+
+if git diff --quiet "${OT_SHA_OLD}" -- include; then
+    echo 'No OpenThread public APIs updates.'
```
simonlingoogle, line 44: check that API version is not changed?

script/check-api-version
```diff
@@ -0,0 +1,64 @@
...
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
+set -euo pipefail
+
+readonly OT_SHA_OLD="$(git cat-file -p HEAD | grep 'parent ' | head -n1 | cut -d' ' -f2)"
+readonly OT_VERSIONS_FILE=tmp/versions
```
simonlingoogle, line 33: remove OT_VERSIONS_FILE eventually?

simonlingoogle, line 33:
```suggestion
readonly OT_VERSIONS_FILE=tmp/api_versions
```

script/check-api-version
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+#
+#  Copyright (c) 2019, The OpenThread Authors.
```
wgtdkp, line 3:
```suggestion
#  Copyright (c) 2020, The OpenThread Authors.
```

script/check-api-version
```diff
@@ -0,0 +1,64 @@
...
+
+old_version=$(grep -aoP '(?<=-#define OPENTHREAD_API_VERSION ).+' "${OT_VERSIONS_FILE}") || die 'Failed to find old version!'
+new_version=$(grep -aoP '(?<=\+#define OPENTHREAD_API_VERSION ).+' "${OT_VERSIONS_FILE}") || die 'Failed to find new version!'
+
+echo -e '\n--------------------------\n'
+
+echo "Old version: ${old_version}"
+echo "New version: ${new_version}"
+
+[[ $((new_version - old_version)) -gt 0 ]] || die 'Version is not increased!'
```
wgtdkp, line 60: Should `new_version` be exactly `old_version + 1`?

.github/workflows/version.yml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  cancel-previous-runs:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: rokroskar/workflow-run-cleanup-action@master
+      env:
+        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+      if: "github.ref != 'refs/heads/master'"
+
+  version:
```
wgtdkp, line 42:
```suggestion
  api-version:
```

.github/workflows/version.yml
```diff
@@ -0,0 +1,49 @@
+#
```
wgtdkp, line 1: rename the file to `api-version.yml` to avoid mixing with existing OT version.

bukepo: I tend to keep this name for extensibility. Thoughts?

script/check-api-version
bukepo: is it necessary? during running checks, a lot of files are generated. We don't remove them before.

script/check-api-version
bukepo: the api definition is in side the include directory.

bukepo:
> If there are two or more PRs that both changed API version from 1 to 2, seems they are **NOT** going to conflict with each other, thus can all be merged successfully.
> 
> After one PR is merged, we need to manually rebase other PRs so that they can be re-checked against the version check.
> 
> However, there seems to be no simple solution for this issue.

In this case, I think the maintainer can provide a suggest and submit according to the merge order. Thoughts?

script/check-api-version
bukepo: No, I think we should allow increase by two for maintainer's convenience. Let's say there are multiple PRs ready to merge, and all the version number must be increased. I think the maintainer should be able to assign each PR a version number and let them re-run again, then wait for a while to merge them all at once. Thoughts?

simonlingoogle:
> In this case, I think the maintainer can provide a suggest and submit according to the merge order. Thoughts?

Might still error-prone, but I am not aware of any other better solution.

script/check-api-version
simonlingoogle: I am considering the case that it's executed locally rather than in github actions. 
Fine to keep the file.

script/check-api-version
wgtdkp: okay with it.

wgtdkp: Is it possible to enforce `new_version == old_version + 1` for only master branch?

simonlingoogle: > Is it possible to enforce `new_version == old_version + 1` for only master branch?

seems not, [pre-receive-hooks](https://help.github.com/en/enterprise/2.19/admin/developer-workflow/using-pre-receive-hooks-to-enforce-policy) can achieve this purpose, but it's only available on GitHub Enterprise Server.

PR #5036, opened 2020-06-01
[mle] restore operational datasets when becoming leader

jwhui:
The leader is authoritative in the operational datasets that it
distributes. This commit ensures that a device becoming a leader
restores the operational dataset from non-volatile. This ensures that
the operational datasets that the leader attempts to distribute are
the same as in its non-volatile storage.

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -337,6 +337,8 @@ void MleRouter::SetStateRouter(uint16_t aRloc16)
 
 void MleRouter::SetStateLeader(uint16_t aRloc16)
 {
+    IgnoreError(Get<MeshCoP::ActiveDataset>().Restore());
+    IgnoreError(Get<MeshCoP::PendingDataset>().Restore());
     SetRloc16(aRloc16);
 
     SetRole(kRoleLeader);


src/core/thread/mle_router.cpp
```diff
@@ -337,6 +337,8 @@ void MleRouter::SetStateRouter(uint16_t aRloc16)
 
 void MleRouter::SetStateLeader(uint16_t aRloc16)
 {
+    IgnoreError(Get<MeshCoP::ActiveDataset>().Restore());
```
wgtdkp, line 340: Instead of using `IgnoreError`, I think we should return the error or log it in-place. thoughts?

Generally, I think we should not use `IgnoreError` in new code and try to remove existing `IgnoreError` where possible.

bukepo: I think this is ignored because the error case is not found only.

jwhui: Adding a log message in place of every place where we might otherwise use `IgnoreError()` will blow up the code size when logging is enabled. As @bukepo mentioned, adding a log message here is not useful because there is a similar log message that will tell us the same information in `ActiveDataset::GenerateLocal()` (which is also called in this path). Thoughts?

wgtdkp: You mean failed to restore operational dataset is okay in context of `SetStateLeader`?

jwhui: Yes, not having an active and/or pending operational dataset is acceptable in the current implementation. `ActiveDataset::GenerateLocal()` will generate one if no active operational dataset exists.

wgtdkp: I agree that this error can be ignored because the semantic seems to be `try to restore operational dataset`, in current context.

PR #5038, opened 2020-06-02
[tests] skip running expected fail tests

bukepo:
This commit skip running expected fail tests so that it will be easier
to find real failures and saves some time.

### DIFF
--- a/tests/scripts/thread-cert/Cert_5_3_10_AddressQuery.py
+++ b/tests/scripts/thread-cert/Cert_5_3_10_AddressQuery.py
@@ -43,6 +43,8 @@ MED1 = 5
 
 
 class Cert_5_3_10_AddressQuery(thread_cert.TestCase):
+    support_ncp = False
+
     topology = {
         LEADER: {
             'mode': 'rsdn',
--- a/tests/scripts/thread-cert/thread_cert.py
+++ b/tests/scripts/thread-cert/thread_cert.py
@@ -28,6 +28,7 @@
 #
 
 import os
+import sys
 import unittest
 
 import config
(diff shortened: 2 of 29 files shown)


tests/scripts/thread-cert/Cert_5_3_10_AddressQuery.py
```diff
@@ -43,6 +43,8 @@
 
 
 class Cert_5_3_10_AddressQuery(thread_cert.TestCase):
+    support_ncp = False
```
simonlingoogle, line 46:
```suggestion
    SUPPORT_NCP = False
```

tests/scripts/thread-cert/thread_cert.py
```diff
@@ -50,7 +51,21 @@
...
+    """
+
+    support_ncp = True
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if os.getenv('NODE_TYPE', 'sim') == 'ncp-sim' and not self.support_ncp:
+            # 77 means skip this test case in automake tests
+            sys.exit(77)
```
simonlingoogle, line 65: nit: move `super().__init__(*args, **kwargs)` below `sys.exit` ?

tests/scripts/thread-cert/Cert_5_3_10_AddressQuery.py
bukepo: this will be done in a separate PR together with the topology attribute.

simonlingoogle: fine to apply style changes in a separate PR

PR #5040, opened 2020-06-02
[meshcop] adding SteeringData class

abtink:
This commit adds a new class `MeshCoP::SteeringData` which manages
the Steering Data and bloom filter generating and filtering.

### DIFF
--- a/src/core/meshcop/meshcop.hpp
+++ b/src/core/meshcop/meshcop.hpp
@@ -37,6 +37,8 @@
 
 #include "openthread-core-config.h"
 
+#include <limits.h>
+
 #include <openthread/instance.h>
 
 #include "coap/coap.hpp"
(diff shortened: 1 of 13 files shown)


src/core/meshcop/meshcop.hpp
```diff
@@ -55,6 +57,155 @@ enum
...
+     *
+     * @param[in] aJoinderId  A Joiner ID.
+     *
+     * @returns TRUE if the bloom filter accepts @p aJoinerId, FALSE otherwise.
+     *
+     */
+    bool Accepts(const Mac::ExtAddress &aJoinerId) const;
+
+    /**
+     * This method indicates whether the bloom filter accepts a hash bit indexes (derived from a Joiner ID).
```
jwhui, line 171:
```suggestion
     * This method indicates whether the bloom filter contains the hash bit indexes (derived from a Joiner ID).
```

src/core/meshcop/meshcop.hpp
```diff
@@ -55,6 +57,155 @@ enum
...
+     */
+    bool IsEmpty(void) const { return DoesAllMatch(0); }
+
+    /**
+     * This method indicates whether the bloom filter accepts all Joiner IDs (all the bits are set).
+     *
+     * @returns TRUE if the bloom filter accepts all Joiners IDs, FALSE otherwise.
+     *
+     */
+    bool AcceptsAll(void) const { return (mLength > 0) && DoesAllMatch(kAcceptAll); }
```
jwhui, line 158: nit: since a bloom filter probabilistically indicates members of a set, how about using "Contains" instead of "Accepts" (here and everywhere).

abtink: Thanks. SGTM. Changed in new pushed commit.

PR #5049, opened 2020-06-04
[tests] add dns client tests

bukepo:


script/test
```diff
@@ -121,7 +126,7 @@ do_build()
 
 do_clean()
 {
-    rm -rfv "${OT_BUILDDIR}"
+    rm -rfv "${OT_BUILDDIR}" || sudo rm -rfv "${OT_BUILDDIR}"
```
Irving-cl, line 129: What about directly using `sudo`?

bukepo: Mostly sudo is not needed. Try it without sudo first to avoid input password.

PR #5051, opened 2020-06-04
[cli] add output for RCP version string

jwhui:


gabekassel:
👏

bukepo:
The client needs parsing to get versions. How about adding a command rcp? Rcp has a subcommand version. This is helpful because we are also going to add command to change rcp log level.

jwhui:
> The client needs parsing to get versions. How about adding a command rcp? Rcp has a subcommand version. This is helpful because we are also going to add command to change rcp log level.

Updated to introduce a separate `rcp` command. Please review.

PR #5052, opened 2020-06-05
[notifier] add Event and Events and Receiver types & simplify handler

abtink:
This PR contains three related commits: 

**[notifier] add Event and Events types & simplify Notifier handler**

This class adds `Event` enumeration type representing `Notifier` events (mirroring `OT_CHANGED_{EVENT}` constants). It also adds `Events` class which represents a collection of events (i.e., signaled from `Notifier` to different modules). The `Events` class provides helper methods to check if a specific event has happened (`Events::Contains()` or if any (`Events::ContainsAny()`) or all (`Events::ContainsAll()`) of a given subset of events are present in the collection.
    
This commit also renames and harmonizes the `Notifier` handler method in different core classes to use `HandleNotififerEvents()`.

**[notifier] add Notifier::Receiver**

This commit adds `Notifier::Receiver` class which is inherited by OpenThread core types/classes to register themselves as a receiver of `Notifier` events. This change helps simplify and replace the previous `Notifier` callback model.

**[posix] rename struct Event to VirtualTimeEvent**

This `struct` is defined in a C header file and therefore is not scoped in a namespae. It is being used for radio/timer simulating (virtual time mode) in POISX platform. The rename is to avoid name conflicts with OpenThread core thus allowing `ot::Event` to be defined/used.

### DIFF
--- a/src/posix/platform/platform-posix.h
+++ b/src/posix/platform/platform-posix.h
@@ -83,7 +83,7 @@ enum
 };
 
 OT_TOOL_PACKED_BEGIN
-struct Event
+struct VirtualTimeEvent
 {
     uint64_t mDelay;
     uint8_t  mEvent;
@@ -331,7 +331,7 @@ void virtualTimeSendRadioSpinelWriteEvent(const uint8_t *aData, uint16_t aLength
  * @param[out]  aEvent  A pointer to the event receiving the event.
  *
  */
-void virtualTimeReceiveEvent(struct Event *aEvent);
+void virtualTimeReceiveEvent(struct VirtualTimeEvent *aEvent);
 
 /**
  * This function sends sleep event through virtual time simulation.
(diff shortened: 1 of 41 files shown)


src/posix/platform/platform-posix.h
```diff
@@ -83,7 +83,7 @@ enum
 };
 
 OT_TOOL_PACKED_BEGIN
-struct Event
+struct VirtualTimeEvent
```
jwhui, line 86: Can we turn this file in to a C++ header file? That would allow us to use namespaces.

abtink: 100% agree. I would say we should change the entire posix code to c++.
I would say we should tackle it in future PR. Here just renamed it to fix build errors (due to name conflicts).

PR #5053, opened 2020-06-05
[posix] use enum for ot-ctl CLI parsing states

simonlingoogle:
One commit for each state:
- PromptState
- DoneState
- ErrorState

### DIFF
--- a/src/posix/client.cpp
+++ b/src/posix/client.cpp
@@ -76,25 +76,33 @@ static void InputCallback(char *aLine)
 }
 #endif // OPENTHREAD_USE_READLINE
 
-static bool FindPrompt(int &aState, char aChar)
+enum PromptState
+{
+    kPromptStateNewline,
+    kPromptStateChar1Received,
+    kPromptStatePrompted,
+    kPromptStateWaitNewline,
+};
+
+static bool FindPrompt(PromptState &aState, char aChar)
 {
     switch (aChar)
     {
     case '>':
-        aState = aState == 0 ? 1 : -1;
+        aState = aState == kPromptStateNewline ? kPromptStateChar1Received : kPromptStateWaitNewline;
         break;
     case ' ':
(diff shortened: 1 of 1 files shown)


src/posix/client.cpp
```diff
@@ -76,95 +76,132 @@ static void InputCallback(char *aLine)
...
-static bool FindPrompt(int &aState, char aChar)
+enum PromptState
+{
+    kPromptStateNewline,
+    kPromptStateChar1Received,
+    kPromptStatePrompted,
+    kPromptStateWaitNewline,
+};
+
+static bool FindPrompt(PromptState &aState, char aChar)
```
bukepo, line 87: how about implement a generic finder?

```cpp
bool Find(int &aState, const char* aPattern, char aChar)
{
    // ...
    if (aPattern[aState] == aChar) // move the next state
    else if (aPattern[aState == '\r' || '\n') // state to zero
    else // state to invalid
    // ...
}
```

simonlingoogle: what do you mean by `generic finder` ?

src/posix/client.cpp
```diff
@@ -76,95 +76,132 @@ static void InputCallback(char *aLine)
...
 {
     switch (aChar)
     {
     case '>':
-        aState = aState == 0 ? 1 : -1;
+        aState = aState == kPromptStateNewline ? kPromptStateChar1Received : kPromptStateWaitNewline;
```
jwhui, line 92: For state machines, I would prefer if we "pivot" this switch statement to `switch (aState)`. It makes it easier to ensure we handle each state appropriately and the conditions for state transmissions from each state.

simonlingoogle: Agree.

simonlingoogle:
Close for now to favor a generic implementation.

PR #5054, opened 2020-06-05
[test] add expect tests for router, scan, and more

Moandor-y:
This adds `expect` tests for `router`, `scan`, and some other simple commands.


Moandor-y:
> Better to split the cli-2-nodes.exp into multiple files each of which contains a single test suites for a single command.

Yes. I'll do this in a separate PR.

PR #5055, opened 2020-06-05
[meshcop] support get/set operational dataset via raw TLVs

### LINKED ISSUE #5028: Support setting multiple dataset properties at once and changing dataset init default values
**Problem Statement**
As of May 2020, initializing an operational dataset uses a pre-made default dataset. Separately, operational datasets can only be updated one property/attribute as a time (e.g. panid, masterkey, etc). 

**Potential Solution**
Support setting multiple dataset properties in one command and potentially tie to dataset init to "seed" the dataset fully upon init.

jwhui:
**Problem**
Thread stores and distributes Active and Pending Operational Datasets as opaque binary objects. To support forward compatibility with newer Thread versions, Thread devices must store and propagate TLVs even when they are unknown. The existing dataset API only allows getting/setting dataset values that are defined.

**Solution**
Introduce new APIs (and associated CLI commands) that allow getting/setting Active/Pending Operational Datasets via binary-encoded TLVs. This allows higher layers to copy Operational Datasets from one device to another without losing any information.

Example CLIs:

Print Active Operational Dataset as hex-encoded TLVs.

```bash
> dataset active binary
0e080000000000010000000300001035060004001fffe002084eb74ab03c56e6d00708fdc7fe165c83a67805108e2104f183e698da87e96efc1e45aa51030f4f70656e5468726561642d383631310102861104108d6273023d82c841eff0e68db86f35740c030000ff
Done
```

Set the Active Operational Dataset using hex-encoded TLVs.

```bash
> dataset set active 0e080000000000010000000300001035060004001fffe002084eb74ab03c56e6d00708fdc7fe165c83a67805108e2104f183e698da87e96efc1e45aa51030f4f70656e5468726561642d383631310102861104108d6273023d82c841eff0e68db86f35740c030000ff
Done
```

Resolves #5028

### DIFF
--- a/include/openthread/dataset.h
+++ b/include/openthread/dataset.h
@@ -352,12 +352,38 @@ otError otDatasetGetActiveTlvs(otInstance *aInstance, otOperationalDatasetTlvs *
  * @param[in]  aInstance A pointer to an OpenThread instance.
  * @param[in]  aDataset  A pointer to the Active Operational Dataset.
  *
- * @retval OT_ERROR_NONE          Successfully set the Active Operational Dataset.
- * @retval OT_ERROR_NO_BUFS       Insufficient buffer space to set the Active Operational Dataset.
- * @retval OT_ERROR_INVALID_ARGS  @p aDataset was NULL.
+ * @retval OT_ERROR_NONE             Successfully set the Active Operational Dataset.
+ * @retval OT_ERROR_NO_BUFS          Insufficient buffer space to set the Active Operational Dataset.
+ * @retval OT_ERROR_NOT_IMPLEMENTED  The platform does not implement settings functionality.
  *
  */
 otError otDatasetSetActive(otInstance *aInstance, const otOperationalDataset *aDataset);
+
+/**
+ * This function sets the Active Operational Dataset.
+ *
+ * If the dataset does not include an Active Timestamp, the dataset is only partially complete.
+ *
+ * If Thread is enabled on a device that has a partially complete Active Dataset, the device will attempt to attach to
+ * an existing Thread network using any existing information in the dataset. Only the Thread Master Key is needed to
+ * attach to a network.
+ *
+ * If channel is not included in the dataset, the device will send MLE Announce messages across different channels to
+ * find neighbors on other channels.
+ *
+ * If the device successfully attaches to a Thread network, the device will then retrieve the full Active Dataset from
+ * its Parent. Note that a router-capable device will not transition to the Router or Leader roles until it has a
+ * complete Active Dataset.
+ *
+ * @param[in]  aInstance A pointer to an OpenThread instance.
+ * @param[in]  aDataset  A pointer to the Active Operational Dataset.
+ *
+ * @retval OT_ERROR_NONE             Successfully set the Active Operational Dataset.
+ * @retval OT_ERROR_NO_BUFS          Insufficient buffer space to set the Active Operational Dataset.
+ * @retval OT_ERROR_NOT_IMPLEMENTED  The platform does not implement settings functionality.
+ *
+ */
+otError otDatasetSetActiveTlvs(otInstance *aInstance, const otOperationalDatasetTlvs *aDataset);
 
 /**
  * This function gets the Pending Operational Dataset.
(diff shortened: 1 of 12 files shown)


include/openthread/dataset.h
```diff
@@ -234,6 +234,24 @@ typedef struct otOperationalDataset
...
+/**
+ * Maximum length of Operational Dataset in bytes.
+ *
+ */
+#define OT_OPERATIONAL_DATASET_MAX_LENGTH 254
+
+/**
+ * This structure represents an Active or Pending Operational Dataset.
+ *
+ * The operational dataset is TLV encoded as specified by Thread.
```
librasungirl, line 246:
```suggestion
 * The Operational Dataset is TLV encoded as specified by Thread.
```

include/openthread/dataset.h
```diff
@@ -234,6 +234,24 @@ typedef struct otOperationalDataset
...
+ *
+ */
+typedef struct otOperationalDatasetTlvs
+{
+    uint8_t mTlvs[OT_OPERATIONAL_DATASET_MAX_LENGTH]; ///< Operational dataset TLVs.
```
librasungirl, line 251:
```suggestion
    uint8_t mTlvs[OT_OPERATIONAL_DATASET_MAX_LENGTH]; ///< OperationalDataset TLVs.
```

include/openthread/dataset.h
```diff
@@ -234,6 +234,24 @@ typedef struct otOperationalDataset
...
+    uint8_t mLength;                                  ///< Size of operational dataset in bytes.
```
librasungirl, line 252:
```suggestion
    uint8_t mLength;                                  ///< Size of Operational Dataset in bytes.
```

include/openthread/dataset.h
```diff
@@ -354,13 +354,41 @@ otError otDatasetGetActiveTlvs(otInstance *aInstance, otOperationalDatasetTlvs *
...
+ * If the device successfully attaches to a Thread network, the device will then retrieve the full Active Dataset from
+ * its Parent. Note that a router-capable device will not transition to the Router or Leader roles until it has a
+ * complete Active Dataset.
+ *
+ * @param[in]  aInstance A pointer to an OpenThread instance.
+ * @param[in]  aDataset  A pointer to the Active Operational Dataset.
+ *
+ * @retval OT_ERROR_NONE             Successfully set the Active Operational Dataset.
+ * @retval OT_ERROR_NO_BUFS          Insufficient buffer space to set the Active Operational Dataset.
+ * @retval OT_ERROR_INVALID_ARGS     @p aDataset was NULL.
```
librasungirl, line 386: `OT_ASSERT` is used when aDataset is NULL, there should be no retval `OT_ERROR_INVALID_ARGS`?

https://github.com/openthread/openthread/blob/3915bda983155b9051619c23641ae70f5e25c0f9/src/core/api/dataset_api.cpp#L87-L93

src/core/meshcop/dataset.hpp
```diff
@@ -51,16 +51,16 @@ namespace MeshCoP {
  * This class represents MeshCop Dataset.
  *
  */
-class Dataset
+class Dataset : private otOperationalDatasetTlvs
```
abtink, line 54: I have a suggestion here (to resolve the casting issue). 
- Let's not have `Dataset` inherit from `otOperationalDatasetTlvs`
- Instead we can add method to convert from one type to the other.
   - A `Dataset::ConvertTo(otOperationalDatasetTlv &aDatasetTlvs) const;`
   - And maybe a constructor `Dataset::Dataset(Type aType, otOperationalDatasetTlv &aDatasetTlvs)` (or an `Init`).

We will have extra copying but looking at  `DatasetLocal::Read(otOperationalDatasetTlvs &aDataset` we already do need copying.

-------
Alternate (maybe longer term) is to change `Dataset` class to not track the `mType` and `mUpdateTime` and have maybe `DatasetManager` or `DatsetLocal` do it?

jwhui: I was trying to avoid the copy, but that would require more work as you pointed out. I reverted to the suggested simpler approach. Please review.

PR #5058, opened 2020-06-05
[cmake] use the correct define for external heap

markus-becker-tridonic-com:
In options.cmake `OPENTHREAD_CONFIG_`**EXTERNAL_HEAP**`_ENABLE` is used, while
in common-switches.mk and the source code the slightly different
`OPENTHREAD_CONFIG_`**HEAP_EXTERNAL**`_ENABLE` is used.

Signed-off-by: Markus Becker <markus.becker@tridonic.com>

### DIFF
--- a/etc/cmake/options.cmake
+++ b/etc/cmake/options.cmake
@@ -132,7 +132,7 @@ endif()
 
 option(OT_EXTERNAL_HEAP "enable external heap support")
 if(OT_EXTERNAL_HEAP)
-    target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE=1")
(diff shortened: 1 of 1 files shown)


etc/cmake/options.cmake
```diff
@@ -132,7 +132,7 @@ endif()
 
 option(OT_EXTERNAL_HEAP "enable external heap support")
 if(OT_EXTERNAL_HEAP)
-    target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE=1")
+    target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE=1")
```
simonlingoogle, line 135: Should also change the cmake option to `OT_HEAP_EXTERNAL`?

markus-becker-tridonic-com: That would have an impact of users of the config API, e.g. Zephyr. So it would need some sort of deprecation warning.

PR #5063, opened 2020-06-06
[test] split out 1.2 and posix from simulation

bukepo:
This commit splits out 1.2 and posix workflows from simulation workflow
so that when an unstable job failed, just a few jobs need rerun.


simonlingoogle:
Do you think we should add badge for this workflow to the readme.md?

bukepo:
> Do you think we should add badge for this workflow to the readme.md? 

Is it necessary?

simonlingoogle:
> > Do you think we should add badge for this workflow to the readme.md?
> 
> Is it necessary?

It's nothing big deal. 
We used to have fewer workflows and we have badges for `Build` and `Simulation`. 
Now that we are having more and more workflows, including `Toranj`, `POSIX`, `Simulation`, `Simulation 1.2`, the question is going to be why do we only add badge for `Simulation` ?

bukepo:
> > > Do you think we should add badge for this workflow to the readme.md?
> > 
> > 
> > Is it necessary?
> 
> It's nothing big deal.
> We used to have fewer workflows and we have badges for `Build` and `Simulation`.
> Now that we are having more and more workflows, including `Toranj`, `POSIX`, `Simulation`, `Simulation 1.2`, the question is going to be why do we only add badge for `Simulation` ?

There's already other workflows(CIFuzzer, Toranj, Size) not listed. I tend to leave this to the maintainer to decide which to list and which not.

PR #5065, opened 2020-06-08
[meshcop] move Joiner::IsPskdValid() to MeshCoP::IsPskdValid()

wgtdkp:
`Joiner::IsPskdValid` is called by the `Commissioner`. In case of building with `JOINER=0 COMMISSIONER=1`, it gives linking error of `undefined Joiner::IsPskdValid` (relevant comment https://github.com/openthread/openthread/commit/092eb186e17ffbcab21178b14c090377e4a44c84#commitcomment-39737571).

This PR fixes this issue by including implementation of `Joiner::IsPskdValid` when COMMISSIONER feature is enabled.


src/core/meshcop/joiner.cpp
```diff
@@ -697,3 +673,37 @@ void Joiner::LogCertMessage(const char *aText, const Coap::Message &aMessage) co
 } // namespace ot
 
 #endif // OPENTHREAD_CONFIG_JOINER_ENABLE
+
+#if OPENTHREAD_CONFIG_JOINER_ENABLE || OPENTHREAD_CONFIG_COMMISSIONER_ENABLE
+
+namespace ot {
+namespace MeshCoP {
+
+bool Joiner::IsPskdValid(const char *aPskd)
```
jwhui, line 682: Given that this method is used by both Commissioner and Joiner, should we move this to a more neutral location? For example, [src/core/meshcop/meshcop.hpp](https://github.com/openthread/openthread/blob/master/src/core/meshcop/meshcop.hpp).

wgtdkp: Good idea! will do.

wgtdkp: @jwhui Relocated to [meshcop.hpp](https://github.com/openthread/openthread/blob/master/src/core/meshcop/meshcop.hpp), please review.

PR #5070, opened 2020-06-09
[posix] fix otPlatAlarmMicroGetNow to use scaled time

paragdixit-g:
This makes it consistent with MilliGetNow() conterpart


src/posix/platform/alarm.cpp
```diff
@@ -98,7 +98,7 @@ void otPlatAlarmMilliStop(otInstance *aInstance)
 #if OPENTHREAD_CONFIG_PLATFORM_USEC_TIMER_ENABLE
 uint32_t otPlatAlarmMicroGetNow(void)
 {
-    return (uint32_t)(otPlatTimeGet());
+    return (uint32_t)(platformAlarmGetNow());
```
bukepo, line 101:
```suggestion
    return static_cast<uint32_t>(platformAlarmGetNow());
```

PR #5071, opened 2020-06-09
[cmake] add posix in platform list

bukepo:
* rename "none" to NO, which computes to false in `if()`
* sort platform names
* add "posix" into the platform list so that it can be selected from cmake-gui
* remove "utils" from the platform list


bukepo:
> Stylistically it might make more sense to use the empty string ``""`` instead of ``"NO"``, but overall looks like a good improvement to me!

I agree. Unfortunately, the cmake gui doesn't support empty options.

PR #5075, opened 2020-06-09
[discover-scanner] adding Mle::DiscoverScanner

abtink:
This commit adds new class `DiscoverScanner` which implements the MLE
Discover Scan process (moving and simplifying all the code related to
this from `Mle` and `MeshForwarder` into same file/class).

----
New commits:

**[toranj] enhance discover scan test - cover filtering, channel-mask**

This commit enhances `test-005-discover-scan` to cover discover
scan over subset of channels, PAN ID filtering, and joiner only,
bloom filter scan behavior.

**[discover-scanner] allow filtering based on a user given ID**
    
This commit enhances the filtering behavior of Discover Scan. When
filtering is enabled, MLE Discovery Responses with steering data
(bloom filter) not containing a given ID (MAC Extended Address) are
filtered. This commit updates the `DiscoverScanner::Discover()` to
allow the ID used for filtering to be optionally specified by the
caller. It can be set to NULL (for default behavior) where the hash
of factory-assigned EUI64 of the device would be used.

### DIFF
--- a/src/core/thread/discover_scanner.cpp
+++ b/src/core/thread/discover_scanner.cpp
@@ -62,6 +62,7 @@ otError DiscoverScanner::Discover(const Mac::ChannelMask &aScanChannels,
                                   uint16_t                aPanId,
                                   bool                    aJoiner,
                                   bool                    aEnableFiltering,
+                                  const FilterIndexes *   aFilterIndexes,
                                   Handler                 aCallback,
                                   void *                  aContext)
 {
@@ -76,12 +77,18 @@ otError DiscoverScanner::Discover(const Mac::ChannelMask &aScanChannels,
 
     if (mEnableFiltering)
     {
-        Mac::ExtAddress extAddress;
+        if (aFilterIndexes == NULL)
+        {
+            Mac::ExtAddress extAddress;
 
-        Get<Radio>().GetIeeeEui64(extAddress);
-        MeshCoP::ComputeJoinerId(extAddress, extAddress);
-
-        MeshCoP::SteeringData::CalculateHashBitIndexes(extAddress, mFilterIndexes);
+            Get<Radio>().GetIeeeEui64(extAddress);
+            MeshCoP::ComputeJoinerId(extAddress, extAddress);
+            MeshCoP::SteeringData::CalculateHashBitIndexes(extAddress, mFilterIndexes);
+        }
+        else
+        {
+            mFilterIndexes = *aFilterIndexes;
+        }
     }
 
     mHandler            = aCallback;
--- a/src/core/thread/discover_scanner.cpp
+++ b/src/core/thread/discover_scanner.cpp
@@ -1,0 +1,351 @@
+/*
+ *  Copyright (c) 2016-2020, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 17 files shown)


src/core/thread/mesh_forwarder.hpp
```diff
@@ -438,6 +441,9 @@ class MeshForwarder : public InstanceLocator
 
     otError GetDestinationRlocByServiceAloc(uint16_t aServiceAloc, uint16_t &aMeshDest);
 
+    void PuaseMessageTransmissions(void) { mTxPaused = true; }
```
jwhui, line 444:
```suggestion
    void PauseMessageTransmissions(void) { mTxPaused = true; }
```

src/core/thread/discover_scanner.cpp
```diff
@@ -0,0 +1,351 @@
...
+    case kStateIdle:
+        break;
+
+    case kStateScanning:
+        // Mark the Discovery Request message for direct tx to ensure it
+        // is not dequeued and freed by `MeshForwarder` and is ready for
+        // the next scan channel. Also pause message tx on `MeshForwarder`
+        // while listening to receive Discovery Responses.
+        aMessage.SetDirectTransmission();
+        Get<MeshForwarder>().PuaseMessageTransmissions();
```
jwhui, line 181:
```suggestion
        Get<MeshForwarder>().PauseMessageTransmissions();
```

src/core/thread/mesh_forwarder.hpp
```diff
@@ -519,12 +524,8 @@ class MeshForwarder : public InstanceLocator
...
 
     Tasklet mScheduleTransmissionTask;
-    bool    mEnabled;
-
-    Mac::ChannelMask mScanChannels;
-    uint8_t          mScanChannel;
-    uint16_t         mRestorePanId;
-    bool             mScanning;
+    bool    mEnabled : 1;
+    bool    mTxPaused : 1;
```
jwhui, line 528: nit: co-locate all the bools?

src/core/thread/discover_scanner.hpp
```diff
@@ -0,0 +1,157 @@
...
+    static void HandleTimer(Timer &aTimer);
+    void        HandleTimer(void);
+
+    Handler                               mHandler;
+    void *                                mHandlerContext;
+    TimerMilli                            mTimer;
+    MeshCoP::SteeringData::HashBitIndexes mFilterIndexes;
+    Mac::ChannelMask                      mScanChannels;
+    uint8_t                               mScanChannel;
+    State                                 mState;
```
jwhui, line 153: nit: to avoid padding? Also, with C++11, we can specify a type for enums.
```suggestion
    State                                 mState;
    uint8_t                               mScanChannel;
```

src/core/thread/discover_scanner.hpp
```diff
@@ -109,6 +112,7 @@ class DiscoverScanner : public InstanceLocator
                      Mac::PanId              aPanId,
                      bool                    aJoiner,
                      bool                    aEnableFiltering,
+                     const Mac::ExtAddress * aFilterId,
```
jwhui, line 115: Should we allow for something more general? Either a `uint64_t` or even a byte array?

abtink: I actually changed this in new pushed commit to use bloom filter (steering data) `FilterIndexes` directly.

PR #5077, opened 2020-06-10
[style] advance C++ language standard to C++11

### LINKED ISSUE #3880: Move language standard baseline to C++11?

jwhui:
Resolves #3880.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -227,7 +227,7 @@ AC_PROG_LIBTOOL
 #
 
 PROSPECTIVE_CFLAGS="-Wall -Wextra -Wshadow -Wundef -Wcast-align -Werror -Wno-error=undef -std=c99 -pedantic-errors"
-PROSPECTIVE_CXXFLAGS="-Wall -Wextra -Wshadow -Wundef -Wcast-align -Werror -Wno-error=undef -std=gnu++98 -Wno-c++14-compat -fno-exceptions"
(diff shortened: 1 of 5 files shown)


configure.ac
```diff
@@ -227,7 +227,7 @@ AC_PROG_LIBTOOL
 #
 
 PROSPECTIVE_CFLAGS="-Wall -Wextra -Wshadow -Wundef -Wcast-align -Werror -Wno-error=undef -std=c99 -pedantic-errors"
-PROSPECTIVE_CXXFLAGS="-Wall -Wextra -Wshadow -Wundef -Wcast-align -Werror -Wno-error=undef -std=gnu++98 -Wno-c++14-compat -fno-exceptions"
+PROSPECTIVE_CXXFLAGS="-Wall -Wextra -Wshadow -Wundef -Wcast-align -Werror -Wno-error=undef -std=c++11 -Wno-c++14-compat -fno-exceptions -pedantic-errors"
```
wgtdkp, line 230: Should we enforce `-std=c++11` for CMake as well?

jwhui: I think that's already covered:
https://github.com/openthread/openthread/blob/4d1197a34af2616c2a00865221d73b47cbfbbbaf/src/core/CMakeLists.txt#L33-L38

wgtdkp: oh, I missed that.

PR #5087, opened 2020-06-11
[posix] fix ot-ctl CLI output missing

simonlingoogle:
This PR resolves https://github.com/openthread/ot-br-posix/issues/468.

This issue occurs when ot-ctl reads the CLI output in multiple chunks in which the last line of a chunk does not ends with a '\r\n'. The last line is then dropped incorrectly. 

This PR fixes this issue:
- Output the last line (without `\r\n`) if it's not `">"` (have to drop `'>'` because we can not determine whether or not to output `'>'`)
- Output `">"` if the first char is not `' '` but the previous `promptState` is 1 (meaning `'>'` was dropped previous but should have been printed).

I think the fix is difficult to understand and in long term we should switch to a more readable implementation that just reads and processes one complete line at a time without these state transfers. Thoughts? @bukepo @gjc13 @jwhui

### DIFF
--- a/src/posix/client.cpp
+++ b/src/posix/client.cpp
@@ -301,6 +301,7 @@ int main(int argc, char *argv[])
                 for (ssize_t i = 0; i < rval; i++)
                 {
                     int prevPromptState = promptState;
+
                     if (FindPrompt(promptState, buffer[i]))
                     {
                         doneState  = 0;
--- a/src/posix/client.cpp
+++ b/src/posix/client.cpp
@@ -326,14 +326,10 @@ int main(int argc, char *argv[])
                     }
                 }
 
-                if (lineStart < rval)
+                if (lineStart < rval && promptState != 1)
                 {
                     assert(promptState != 0 && promptState != 2);
-                    if (promptState == -1)
-                    {
-                        VerifyOrExit(DoWrite(STDOUT_FILENO, buffer + lineStart, rval - lineStart),
-                                     ret = OT_EXIT_FAILURE);
-                    }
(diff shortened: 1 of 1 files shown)


src/posix/client.cpp
```diff
@@ -320,6 +326,16 @@ int main(int argc, char *argv[])
                         ret        = OT_EXIT_SUCCESS;
                     }
                 }
+
+                if (lineStart < rval)
+                {
+                    OT_ASSERT(promptState != 0 && promptState != 2);
```
bukepo, line 332: directly use `assert()`?

src/posix/client.cpp
```diff
@@ -295,17 +296,22 @@ int main(int argc, char *argv[])
...
                     if (FindPrompt(promptState, buffer[i]))
                     {
                         doneState  = 0;
                         errorState = 0;
                         lineStart  = i + 1;
                         continue;
                     }
+                    else if (prevPromptState == 1 && i == 0)
+                    {
+                        VerifyOrExit(DoWrite(STDOUT_FILENO, ">", 1), ret = OT_EXIT_FAILURE);
```
bukepo, line 313: can this be implemented in FindPrompt?

simonlingoogle: FindPrompt can not check `i == 0`

jwhui:
> I think the fix is difficult to understand and in long term we should switch to a more readable implementation that just reads and processes one complete line at a time without these state transfers. Thoughts? @bukepo @gjc13 @jwhui

Thanks for addressing this. I agree that we should clean this up in a future PR.

PR #5088, opened 2020-06-11
[docs] add English source for the Thread Primer

Vyrastas:
For translation purposes.

The directory structure mirrors what we have on openthread.io, for easy import.  For now, I didn't bother with images, so those are all broken, but I can add them later if we really want.


Vyrastas:
@jwhui This is failing the pretty check, is there a way to disable for a PR?  I can eventually tweak the source so it passes and doesn't mess with how its rendered on openthread.io, but I'd rather not mess with it at the moment.

jwhui:
> @jwhui This is failing the pretty check, is there a way to disable for a PR? I can eventually tweak the source so it passes and doesn't mess with how its rendered on openthread.io, but I'd rather not mess with it at the moment.

@Vyrastas , I think this PR should add `doc/en` to `OT_EXCLUDE_DIRS` so that this and future PRs exclude these files during the check.
https://github.com/openthread/openthread/blob/2c258a3f3e0a09315fe41ea2c32e8b0ecab8176b/script/make-pretty#L64

Vyrastas:
> > @jwhui This is failing the pretty check, is there a way to disable for a PR? I can eventually tweak the source so it passes and doesn't mess with how its rendered on openthread.io, but I'd rather not mess with it at the moment.
> 
> @Vyrastas , I think this PR should add `doc/en` to `OT_EXCLUDE_DIRS` so that this and future PRs exclude these files during the check.
> https://github.com/openthread/openthread/blob/2c258a3f3e0a09315fe41ea2c32e8b0ecab8176b/script/make-pretty#L64

Great, so that we don't have to update that script for each language folder, I've moved the content in a doc/site folder and excluded that.

Vyrastas:
Also, to make the file look cleaner, I've removed project/book YAML tags at the beginning of each.  I can have my import script add those, they aren't necessary to have here on GitHub.

dsyx:
Would it be better to define a shared resource directory for various translations (eg: `doc/share` for storing images/styles/compressed packages, etc.)?

Vyrastas:
> Would it be better to define a shared resource directory for various translations (eg: `doc/share` for storing images/styles/compressed packages, etc.)?

Let me take a look because now that you mention it, I'm not sure the images would work with the same links on openthread.io if they only exist in the en/ language directory.  For now I'd just leave the links.

Styles are built into the openthread.io authoring platform, so that's nothing we'd host here.

Vyrastas:
Regarding a shared resource directory, this will be handled by the openthread.io publishing platform.  For the Thread Primer we will use the existing images we have with the English text.

Ideally people should go to openthread.io to view the Primer and any translated pages.  If there comes a need to host images here on GitHub, we can do that later.

dsyx:
So I don’t need to modify the image address references in the original text, right?

Vyrastas:
> So I don’t need to modify the image address references in the original text, right?

You do not, leave them as they are.

PR #5089, opened 2020-06-12
[posix] get time offset between host and rcp

brianljt:
To support CSL on RCP mode, POSIX host needs to know current time on RCP. This commit adds function to periodically calculate the time offset between host and RCP so that host can get an estimated RCP time by adding its current time with this offset.


examples/platforms/nrf528xx/src/alarm.c
```diff
@@ -688,7 +688,8 @@ void RTC_IRQ_HANDLER(void)
     }
 }
 
-#if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
+#if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE || OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE || \
```
bukepo, line 691: can we drop these guards, and let the linker to drop unused symbols?

src/core/radio/radio_platform.cpp
```diff
@@ -145,3 +145,10 @@ OT_TOOL_WEAK void otPlatRadioSetMacFrameCounter(otInstance *aInstance, uint32_t
     OT_UNUSED_VARIABLE(aInstance);
     OT_UNUSED_VARIABLE(aMacFrameCounter);
 }
+
+OT_TOOL_WEAK uint64_t otPlatRadioGetNow(otInstance *aInstance)
+{
+    OT_UNUSED_VARIABLE(aInstance);
+
```
bukepo, line 152: return otPlatTimeGet()?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1253,6 +1261,26 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Get(spinel_prop_key_t aK
     return error;
 }
 
+template <typename InterfaceType, typename ProcessContextType>
+otError RadioSpinel<InterfaceType, ProcessContextType>::GetWithParam(spinel_prop_key_t aKey,
```
bukepo, line 1265: is this going to be used by other functions? If not, should it be embedded into CalcRcpTimeOffset() for now?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1696,5 +1724,71 @@ otRadioState RadioSpinel<InterfaceType, ProcessContextType>::GetState(void) cons
...
+    localTxTimestamp = otPlatTimeGet();
+
+    // Dummy timestamp payload to make request length same as response
+    error = GetWithParam(SPINEL_PROP_RCP_TIMESTAMP, buffer, packed, SPINEL_DATATYPE_UINT64_S, &remoteTimestamp);
+
+    localRxTimestamp = otPlatTimeGet();
+
+    VerifyOrExit(error == OT_ERROR_NONE, mRadioTimeRecalcStart = localRxTimestamp);
+
+    mRadioTimeOffset      = remoteTimestamp - ((localRxTimestamp / 2) + (localTxTimestamp / 2));
```
bukepo, line 1778:
```suggestion
    mRadioTimeOffset      = remoteTimestamp - ((localRxTimestamp + localTxTimestamp) / 2));
```

brianljt:
Measured accuracy is about +300us. This should be good enough for RCP usage in which case, host will queue frame to RCP several millisecond in advance.

Test platform,
Raspberry Pi 3 B+(host) and nRF52840(RCP)

Test method,
Use an interrupt line to get accurate timestamp on both size. The offset get by interrupt is called Dint. Use this commit to get the offset, called Dcalc. The accuracy is,
Accuracy = Dcalc - Dint.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1696,5 +1724,71 @@ otRadioState RadioSpinel<InterfaceType, ProcessContextType>::GetState(void) cons
...
+
+exit:
+    LogIfFail("Error calculating RCP time offset: %s", error);
+#endif
+}
+
+template <typename InterfaceType, typename ProcessContextType>
+uint64_t RadioSpinel<InterfaceType, ProcessContextType>::GetNow(void)
+{
+    return mIsTimeSynced ? (otPlatTimeGet() + static_cast<uint64_t>(mRadioTimeOffset)) : 0;
```
bukepo, line 1790: `0` is a valid time, how about using UINT64_MAX
```suggestion
    return mIsTimeSynced ? (otPlatTimeGet() + static_cast<uint64_t>(mRadioTimeOffset)) : UINT64_MAX;
```

src/ncp/ncp_base_radio.cpp
```diff
@@ -221,6 +222,20 @@ template <> otError NcpBase::HandlePropertyGet<SPINEL_PROP_MAC_SRC_MATCH_ENABLED
     return mEncoder.WriteBool(mSrcMatchEnabled);
 }
 
+template <> otError NcpBase::HandlePropertyGet<SPINEL_PROP_RCP_TIMESTAMP>(void)
+{
+    otError error = OT_ERROR_NONE;
+
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE || OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
```
bukepo, line 229: drop the conditions?

include/openthread/platform/radio.h
```diff
@@ -489,6 +489,16 @@ void otPlatRadioSetMacKey(otInstance *    aInstance,
  */
 void otPlatRadioSetMacFrameCounter(otInstance *aInstance, uint32_t aMacFrameCounter);
 
+/**
+ * Get the current estimated time (64bits width) of the radio chip.
+ *
+ * @param[in]   aInstance    A pointer to an OpenThread instance.
+ *
+ * @returns The current time in microseconds.
```
bukepo, line 497: Document the special value when the RCP doesn't support or out of sync with Host?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1696,5 +1724,71 @@ otRadioState RadioSpinel<InterfaceType, ProcessContextType>::GetState(void) cons
...
+    VerifyOrExit(packed > 0 && static_cast<size_t>(packed) <= sizeof(buffer), error = OT_ERROR_NO_BUFS);
+
+    localTxTimestamp = otPlatTimeGet();
+
+    // Dummy timestamp payload to make request length same as response
+    error = GetWithParam(SPINEL_PROP_RCP_TIMESTAMP, buffer, packed, SPINEL_DATATYPE_UINT64_S, &remoteTimestamp);
+
+    localRxTimestamp = otPlatTimeGet();
+
+    VerifyOrExit(error == OT_ERROR_NONE, mRadioTimeRecalcStart = localRxTimestamp);
```
bukepo, line 1776: Record failure times so that we can check liveness of RCP and take actions when it's dead?

src/core/radio/radio_platform.cpp
brianljt: Not all platform supports this API. Currently, only nrf528xx and simulation supports this. Others support `otPlatAlarmMilliGetNow`. So need to return a special value here as a weak implementation.

src/lib/spinel/radio_spinel_impl.hpp
brianljt: OK.

src/lib/spinel/radio_spinel_impl.hpp
brianljt: This is to avoid overflow

src/lib/spinel/radio_spinel_impl.hpp
brianljt: Good idea!

src/lib/spinel/radio_spinel_impl.hpp
brianljt: `WaitResponse` already handle the case when there are no response from RCP. It will die with `OT_EXIT_RADIO_SPINEL_NO_RESPONSE`

bukepo: yes.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1696,5 +1724,71 @@ otRadioState RadioSpinel<InterfaceType, ProcessContextType>::GetState(void) cons
...
+
+exit:
+    LogIfFail("Error calculating RCP time offset: %s", error);
+#endif
+}
+
+template <typename InterfaceType, typename ProcessContextType>
+uint64_t RadioSpinel<InterfaceType, ProcessContextType>::GetNow(void)
+{
+    return mIsTimeSynced ? (otPlatTimeGet() + static_cast<uint64_t>(mRadioTimeOffset)) : UINT64_MAX;
```
zhanglongxia, line 1790: If the "mRadioTimeOffset" is a negative value, line 1790 converts it to a positive value. Will it affect the calculated result?

zhanglongxia:
@brianljt  Could you share the synchronization accuracy you tested?

src/lib/spinel/radio_spinel_impl.hpp
brianljt: No, without static_cast, compiler with cast int64_t to uint64_t, too. For negative value, it is just complement add. So `((uint64_t)100 + (int64_t)(-50)) = 50`

For accuracy, please refer to this comment https://github.com/openthread/openthread/pull/5089#issuecomment-643041967

src/lib/spinel/radio_spinel.hpp
```diff
@@ -653,6 +653,21 @@ template <typename InterfaceType, typename ProcessContextType> class RadioSpinel
      */
     otError RestoreDatasetFromNcp(void);
 
+    /**
+     * This method returns the next timepoint to recalculate RCP time offset.
+     *
+     * @returns The timepoint to start the recalculation of RCP time offset.
+     *
+     */
+    uint64_t GetNextRadioTimeRecalcStart(void) { return mRadioTimeRecalcStart; }
```
jwhui, line 662:
```suggestion
    uint64_t GetNextRadioTimeRecalcStart(void) const { return mRadioTimeRecalcStart; }
```

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1696,5 +1724,71 @@ otRadioState RadioSpinel<InterfaceType, ProcessContextType>::GetState(void) cons
...
+
+    VerifyOrExit(error == OT_ERROR_NONE, mRadioTimeRecalcStart = localRxTimestamp);
+
+    mRadioTimeOffset      = remoteTimestamp - ((localRxTimestamp / 2) + (localTxTimestamp / 2));
+    mIsTimeSynced         = true;
+    mRadioTimeRecalcStart = localRxTimestamp + RCP_TIME_OFFSET_CHECK_INTERVAL;
+
+exit:
+    LogIfFail("Error calculating RCP time offset: %s", error);
+#endif
```
jwhui, line 1784:
```suggestion
#endif // OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE || OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
```

include/openthread/link_raw.h
```diff
@@ -363,6 +363,16 @@ otError otLinkRawSetMacKey(otInstance *    aInstance,
  */
 otError otLinkRawSetMacFrameCounter(otInstance *aInstance, uint32_t aMacFrameCounter);
 
+/**
+ * Get current platform time (64bits width) of the radio chip.
+ *
+ * @param[in]  aInstance    A pointer to an OpenThread instance.
+ *
+ * @retval     The current platform time in microseconds.
```
jwhui, line 371: "platform" time? or "radio" time?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1253,6 +1261,26 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Get(spinel_prop_key_t aK
...
+                                                                     const char *      aFormat,
+                                                                     ...)
+{
+    otError error;
+
+    assert(mWaitingTid == 0);
+
+    mPropertyFormat = aFormat;
+    va_start(mPropertyArgs, aFormat);
+    error = Request(true, SPINEL_CMD_PROP_VALUE_GET, aKey, SPINEL_DATATYPE_DATA_S, aParam, aParamSize);
```
abtink, line 1277: This is emitting ` VALUE_GET` command for a spinel property with some extra parameters.
`VALUE_GET` command expects only the spinel property id.
I understand you are adding it to make the spinel request and response frame same length specifically for the timestap usecase.

We should be careful about adding this as it is not a supported model for `VALUE_GET` command and should not be used in general.

brianljt: OK. I will add some comments before `GetWithParam`.

src/lib/spinel/radio_spinel.hpp
```diff
@@ -653,6 +653,21 @@ template <typename InterfaceType, typename ProcessContextType> class RadioSpinel
...
+     * @returns The timepoint to start the recalculation of RCP time offset.
+     *
+     */
+    uint64_t GetNextRadioTimeRecalcStart(void) const { return mRadioTimeRecalcStart; }
+
+    /**
+     * This method gets the current estimated time on RCP.
+     *
+     * @returns The current estimated RCP time in microseconds.
+     */
```
jwhui, line 668:
```suggestion
     * @returns The current estimated RCP time in microseconds.
     *
     */
```

PR #5093, opened 2020-06-14
[tests] ignore otError in spinel tests

simonlingoogle:


jwhui:
@simonlingoogle , thanks for addressing this. Can we add a continuous integration check to ensure this is covered?

simonlingoogle:
> @simonlingoogle , thanks for addressing this. Can we add a continuous integration check to ensure this is covered?

These code are ncp test code. If we are to ensure this is covered in CI, we need a ncp test using `CC=clang`.

tests/unit/test_spinel_buffer.cpp
```diff
@@ -524,14 +524,14 @@ void TestBuffer(void)
         WriteTestFrame3(ncpBuffer, Spinel::Buffer::kPriorityHigh);
 
         ncpBuffer.InFrameBegin((j % 2) == 0 ? Spinel::Buffer::kPriorityHigh : Spinel::Buffer::kPriorityLow);
-        ncpBuffer.InFrameFeedData(sHelloText, sizeof(sHelloText));
+        IgnoreError(ncpBuffer.InFrameFeedData(sHelloText, sizeof(sHelloText)));
```
wgtdkp, line 527:
```suggestion
        SuccessOrQuit(ncpBuffer.InFrameFeedData(sHelloText, sizeof(sHelloText)));
```
Since we are in unittest, is `SuccessOrQuit` better than ignoring?

simonlingoogle: Yes. Will revise accordingly.

PR #5095, opened 2020-06-15
[settings] fix DUA logging

simonlingoogle:


src/core/common/settings.cpp
```diff
@@ -81,7 +81,7 @@ void SettingsBase::LogChildInfo(const char *aAction, const ChildInfo &aChildInfo
 #if OPENTHREAD_CONFIG_DUA_ENABLE
 void SettingsBase::LogDadInfo(const char *aAction, const DadInfo &aDadInfo) const
 {
-    otLogInfoCore("Non-volatile: %s DadInfo {DadCounter:%2d}", aAction, aDadInfo.mDadCounter);
+    otLogInfoCore("Non-volatile: %s DadInfo {DadCounter:%2d}", aAction, aDadInfo.GetDadCounter());
```
wgtdkp, line 84: `DUA` features are never enabled in CI?

simonlingoogle: Maybe DUA and logging are never enabled together in CI.

wgtdkp: Seems we need to at a GitHub Action job to at least build the `DUA` feature.

jwhui: We should have a Thread 1.2 build with full logs enabled.

simonlingoogle: Actually, it was because `DUA` feature is not included in the 1.2 simulation build test. 
I have fixed it in this PR too.

Issue #433, opened 2016-08-19
Implement Energy Scan


luwannest:
Certificate 9.2.13 is used to verify the device is able to properly accept and process energy scan process.

This feature need to be implemented to pass the certificate test.

PR #5099, opened 2020-06-15
Fix Keil packed compilation error.

piotrkoziar:
After the recent approach with Clearable and Equatable classes, there are some errors.
 
When building with Keil,  there are compilation errors like the example below:
`error:  #1308: Base class "ot::Equatable<ot::Mac::ExtAddress>" of __packed class "ot::Mac::ExtAddress" must be __packed
`
The error occurs if Clearable or Equatable is a base class for a packed class.
Keil requires that all base classes of a packed class have to be packed.

Below I presented ways to deal with that problem:

1. Make classes Clearable and Equatable packed.
After during so 2 types of errors in IAR compilation (Keil compilation is fixed) appear:

i) 
`no instance of overloaded function "ot::Ip6::NetifUnicastAddress::GetNext" matches the argument list and object (the object has type qualifiers that prevent a match)
            object type is: ot::Ip6::NetifUnicastAddress const __packed`

`...\src\core\net\netif.hpp(106) : Error[Pe167]: argument of type "ot::Ip6::NetifUnicastAddress __packed *" is incompatible with parameter of type "ot::Ip6::NetifUnicastAddress *"`

`...\src\core\net\netif.hpp(159) : Error[Pe513]: a value of type "ot::Ip6::NetifMulticastAddress __packed *" cannot be assigned to an entity of type "otNetifMulticastAddress const *"`

The solution is to make struct otNetifAddress and struct otNetifMulticastAddress from ip6.h packed.
Then error
 ` ...\src\ncp\ncp_base_mtd.cpp(1625): error:  #433: qualifiers dropped in binding reference of type "std::uint8_t &" to initializer of type "__packed std::uint8_t"`
 appers but it can be easily resolved by creating a local variable and passing it to `ReadUint8()`.

ii)
Several errors like that: `src\core\net\ip6_address.cpp(48) : Error[Pe147]: declaration is incompatible with "__interwork __softfp bool ot::Ip6::Address::IsUnspecified() const" (declared <SRCREF line=116 file="...\src\core\net\ip6_address.hpp">at line 116 of "...\src\core\net\ip6_address.hpp"</SRCREF>)`

I did not found a good solution to this.

2. Remove packed from all classes that inherit from Clearable or Equatable.

Doing so requires a lot of changes and I end up with two types of errors (with GCC compiler, did not check Keil or IAR):
i) 
`.../src/core/thread/network_data_tlvs.hpp:343:90: error: cast from 'uint8_t* {aka unsigned char*}' to 'ot::NetworkData::HasRouteEntry*' increases required alignment of target type [-Werror=cast-align]
         return reinterpret_cast<HasRouteEntry *>(GetValue() + (i * sizeof(HasRouteEntry)));`

Error occurs in every place where uint8_t pointer is casted to pointer of non-packed class. 
ii)
`.../src/core/thread/network_diagnostic_tlvs.hpp:695:22: error: ignoring packed attribute because of unpacked non-POD field 'ot::Mle::RouterIdSet ot::NetworkDiagnostic::RouteTlv::mRouterIdMask' [-Werror] `

A lot of changes is required to fix this, so I decided to go another way.

3. Do not make Clearable and Equatable classes packed for IAR.

This approach is implemented in this PR.

### DIFF
--- a/src/core/common/clearable.hpp
+++ b/src/core/common/clearable.hpp
@@ -38,6 +38,8 @@
 
 #include <string.h>
 
+#include <openthread/platform/toolchain.h>
+
 namespace ot {
 
 /**
@@ -53,7 +55,11 @@ template <typename Type> class Clearable
 {
 public:
     void Clear(void) { memset(reinterpret_cast<void *>(this), 0, sizeof(Type)); }
-};
+}
+#if defined(__CC_ARM)
+OT_TOOL_PACKED_END
+#endif
+    ;
 
 } // namespace ot
 
--- a/src/core/common/equatable.hpp
+++ b/src/core/common/equatable.hpp
@@ -38,6 +38,8 @@
 
 #include <string.h>
 
+#include <openthread/platform/toolchain.h>
+
 namespace ot {
 
 /**
@@ -73,7 +75,11 @@ public:
      *
      */
     bool operator!=(const Type &aOther) const { return !(*this == aOther); }
-};
+}
+#if defined(__CC_ARM)
+OT_TOOL_PACKED_END
+#endif
+    ;
 
 } // namespace ot
 


src/core/common/clearable.hpp
```diff
@@ -53,7 +55,7 @@ template <typename Type> class Clearable
 {
 public:
     void Clear(void) { memset(reinterpret_cast<void *>(this), 0, sizeof(Type)); }
-};
+} OT_TOOL_PACKED_END;
```
abtink, line 58: Is there any build option provided by Keil toolchain that allows user to relax the requirement of packed base-classes (specifically in the cases where base-class does not define any member variable).

If not, I would say the the approach to define these as packed only under Keil toolchain is the best approch among 3 options you mentioned.

Is there an `#ifdef` we can add here to limit this line to only Keil tool chain?

abtink:
@piotrkoziar another option I was consdiering is to define a packed version (along with unpacked version)  of `Clearable`/`Equatable`, and then use the specific base class based on whether the sub-class type itself is packed or not.

I added it in PR  https://github.com/openthread/openthread/pull/5102.

Can you please help check if this addresses the issues on Keil and IAR? Thanks.

src/core/common/clearable.hpp
piotrkoziar: Unfortunately Keil compiler documentation does not specify such option.

piotrkoziar:
@abtink I run build tests on #5102 and there is the same issue as I described in option 1.ii:
`...\src\core\net\ip6_address.cpp(67) : Error[Pe147]: declaration is incompatible with "__interwork __softfp bool ot::Ip6::Address::IsUnspecified() const" (declared <SRCREF line=164 file="...\src\core\net\ip6_address.hpp">at line 164 of "...\src\core\net\ip6_address.hpp"</SRCREF>)`
There is more errors like this in the build log.

IAR fails at the compilation of functions from ot::Ip6::Address class that are defined in ip6_address.cpp file. (the similar error applies to mac_type.cpp and mle_type.cpp).

abtink:
Thanks @piotrkoziar for testing.

That's a weird error (not sure if I understand it). Seems like compiler does not think the method declaration and implementation  match (somehow caused with addition of a packed base (template) class?).

abtink:
@piotrkoziar  do you happen to have any idea what is the error we get from IAR? (why is complier not happy with the decleration?)

piotrkoziar:
@abtink I agree that the implementation and the declaration of class methods do not match after __packed is added to it but I don't know the exact reason for this error. I do not have any ideas on how to fix that right now. I asked IAR technical support about it, but I think that the answer would be rather to change the compiler version (which would not help in my case) than to make some nice adjustments in the code.

src/core/common/clearable.hpp
LuDuda: But there is the compile time check that you can use, see [here](https://github.com/openthread/openthread/blob/master/examples/platforms/nrf528xx/src/flash.c#L72).

```
#if defined(__CC_ARM)
```

piotrkoziar: @LuDuda I used it, see the commit https://github.com/openthread/openthread/commit/02af093fa69ddb3bd8dca510d56d6df8a9845b84. This solution passes all build tests (GCC, IAR, KEIL).

piotrkoziar:
@abtink Currently this PR passes all build tests. Clearable and Equatable classes are declared packed only under Keil compiler. Are you ok with this solution?

abtink:
Thanks @piotrkoziar for your help with this and checking further on IAR error.

I agree that this fix works but I feel it is more of workaround (just for Keil toolchain).

Keil seem to require all base classes of a packed type to also be declared as packed (even if they are empty base classes). I believe this is not required by gcc in the case of empty base class. That said, such a requirement seems reasonable to me and would expect that other embedded tool-chains may require the same. So a solution like in https://github.com/openthread/openthread/pull/5102 which addresses this fully on all tool-chains would be better in my opinion.

That's why I'd say it may be worth trying to understand and fix the IAR error with  https://github.com/openthread/openthread/pull/5102 change. In that PR:
- We define a `PackedClearable` and regular (un-packed) `Clearable` 
- If the sub-class type itself is packed, we inherit from `PackedClearable`, otherwise from the `Clearable`

The IAR error seems to be related to cpp implementation of `ot::Ip6::Address::IsUnspecified() const` method not being associated with header file definition. 
```
Error[Pe147]: declaration is incompatible with "__interwork __softfp bool ot::Ip6::Address::IsUnspecified() const" (declared <SRCREF line=164 file="...\src\core\net\ip6_address.hpp">at line 164 of "...\src\core\net\ip6_address.hpp"</SRCREF>)
```
Here are some observations/theories:
- The `Ip6::Address` class itself is defined as packed.
- It already inherits from `otIp6Address` which is itself packed and IAR toolchain is ok with this before.
- In current master it inherits from `Clearable` (which is not declared as packed) and IAR toolchain again seems fine with this.
- The change in the PR is that it now also inherits from `PackedClearable` which causes the compilation error.
- IAR uses the prefix `__packed` (in `OT_TOOL_PACKED_BEGIN`) to declare a type as packed (only tool chain currently that uses a prefix `OT_TOOL_PACKED_BEGIN` and not a suffix `OT_TOOL_PACKED_END).
- Maybe in the template definition of `PackedClearable` the order of `__packed` matters for the IAR tool-chain
   - We have `template <typename Type> OT_TOOL_PACKED_BEGIN class PackedClearable` currently in the PR (first `template` then `__packed`)
  - Maybe chanding the order `OT_TOOL_PACKED_BEGIN template <typename Type> class PackedClearable` would help address the issue (@piotrkoziar  would you be able to please help check if this may help or not?).

Here is my proposal on how to proceed:
- If we can find a way to make https://github.com/openthread/openthread/pull/5102 work on IAR   that would be best way to go.
- If we cannot address the IAR error, I would say we adopt the approach in https://github.com/openthread/openthread/pull/5102 but add code to handle IAR as a special-case (adding a workaround for IAR in the code), i.e., we add `#if` checks and if tool-chain is IAR we declare the `PacekdClearable` same `Clearable` (not packed) only under IAR toolchain.
  - If we need to special-case some tool-chain, I prefer we do for IAR compared to Keil. Keil has a reasonable requirement (of base class of packed type being packed), over the IAR seemingly having a weird limitation/error.

Thoughts?

piotrkoziar:
@abtink Thank you for the analysis. I agree with your points regarding the #5102 and procedure.

Here are my observations:
- Regarding the `OT_TOOL_PACKED_BEGIN template <typename Type> class PackedClearable`: I have already checked and it is not allowed - it produces `Error[Pe463]: "template" is not allowed`.
- Just to make clear, there are errors produced not only for `ot::Ip6::Address::IsUnspecified() const` implementation but for all method implementations from the Address class. 
- Removing `__packed` from Address class did not solve the problem with IAR.
- Removing `__packed` from Address class and from struct `otIp6Address` did not solve the problem with IAR.
- I have tried changing the inheritance order (no effect).
- I have also tried to use some typedefs for Address class (no effect).

As it was said, the next actions depend on if the solution to IAR error is found. There are chances (not high in my opinion) that the response from IAR technical support will put some light on the error and allow us to address it. Otherwise, let's adopt the #5102 approach with #ifdef.

It is hard to tell exactly when but I am expecting the answer from the support in the next week (there is some holiday in Sweden at the end of this week). But I have nothing against merging #5102 earlier.

abtink:
Thanks @piotrkoziar. Seems like you already tested a lot of case.  :)

I added a workaround for IAR toolchain (as a separate commit) in #5102.

piotrkoziar:
This fix is not critical.

jwhui:
> As it was said, the next actions depend on if the solution to IAR error is found. There are chances (not high in my opinion) that the response from IAR technical support will put some light on the error and allow us to address it. Otherwise, let's adopt the #5102 approach with #ifdef.
> 
> It is hard to tell exactly when but I am expecting the answer from the support in the next week (there is some holiday in Sweden at the end of this week). But I have nothing against merging #5102 earlier.

@piotrkoziar , any updates on this?

piotrkoziar:
> > As it was said, the next actions depend on if the solution to IAR error is found. There are chances (not high in my opinion) that the response from IAR technical support will put some light on the error and allow us to address it. Otherwise, let's adopt the #5102 approach with #ifdef.
> > It is hard to tell exactly when but I am expecting the answer from the support in the next week (there is some holiday in Sweden at the end of this week). But I have nothing against merging #5102 earlier.
> 
> @piotrkoziar , any updates on this?

Nothing. I do not know if anyone would have this problem right now because there were changes in C++ contribution policy (C++11) - the IAR compiler version that I was using regarding this issue would not handle the code anyway.

jwhui:
> @jwhui nothing new. I do not know if anyone would have this problem right now because there were changes in C++ contribution policy (C++11) - the IAR compiler version that I was using regarding this issue would not handle the code anyway.

@piotrkoziar , is this still an issue to address? If not, should we close this PR?

piotrkoziar:
> > @jwhui nothing new. I do not know if anyone would have this problem right now because there were changes in C++ contribution policy (C++11) - the IAR compiler version that I was using regarding this issue would not handle the code anyway.
> 
> @piotrkoziar , is this still an issue to address? If not, should we close this PR?

@jwhui I see it that way: so far we found more like a workaround to an issue that is very unlikely to meet again (after certain commit it would be simply not valid). So yes, I recommend closing the PR.

jwhui:
> @jwhui I see it that way: so far we found more like a workaround to an issue that is very unlikely to meet again (after certain commit it would be simply not valid). So yes, I recommend closing the PR.

Thanks, @piotrkoziar

PR #5103, opened 2020-06-16
[github-actions] fix version workflow's name

simonlingoogle:


.github/workflows/version.yml
```diff
@@ -26,7 +26,7 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
-name: Build
+name: Version
```
jwhui, line 29: Can we be more descriptive? Something like "API Version"?

simonlingoogle: The yml file was intentionally named to be `version.yml`. 
@bukepo described about it in https://github.com/openthread/openthread/pull/5034#discussion_r433592601
So I think the name should also be `Version` for the same reason.

jwhui: We have a lot of versions, some are not related to API, like Thread 1.1/1.2. I think we should be clear about what this workflow actually checks. @bukepo @simonlingoogle , thoughts?

bukepo: I was considering further extensions, for example when we finally have some lts versions, say 1.0, we may add checks for it interop with master. And we may add it in this workflow. However, I am fine with renaming it to API version given there is only a single check for now.

simonlingoogle, line 29:
```suggestion
name: API Version
```
Let's use `API Version` for now.

simonlingoogle: renamed

PR #5104, opened 2020-06-16
[logging] remove NCP_SPINEL output

bukepo:
This commit removes log output NCP_SPINEL and update the original log
output APP to use NCP_SPINEL behavior. The issue was raised in #4953.


src/ncp/ncp_base.cpp
```diff
@@ -2393,29 +2393,9 @@ otError otNcpStreamWrite(int aStreamId, const uint8_t *aDataPtr, int aDataLen)
...
-        }
-
-        IgnoreError(otNcpStreamWrite(0, reinterpret_cast<uint8_t *>(logString), charsWritten));
-    }
-}
-
-#if (OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_NCP_SPINEL)
-
-extern "C" void otPlatLog(otLogLevel aLogLevel, otLogRegion aLogRegion, const char *aFormat, ...)
+extern "C" OT_TOOL_WEAK void otPlatLog(otLogLevel aLogLevel, otLogRegion aLogRegion, const char *aFormat, ...)
```
abtink, line 2398: I think this does not need to be `OT_TOOL_WEAK`.

bukepo: I hope to allow developers to override the behavior of APP log output, just like the CLI version. Thoughts?

abtink: If we keep this as weak, my concern is how to ensure this weak `otPlatLog()` is the one linked/used over the other weak implementations we may have (in platform code)?

bukepo: Platform should not provide weak implementation for the app output. It should provide platform defined output instead.

abtink: I agree that it would be not correct way but IIRC we used to define a weak `otPlatLog` (without any config check) in some internal project platform layer (letting it to be overridden if there was another strong definition somewhere).

I think if project config sets `OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_APP` then app layer (which would be NCP) should provide the logging functionality and if there is any other (non-weak) `otPlatLog` somewhere it would be better/safer to get a build error (so we detect if there is an error in config). If a project wants its own logging then I'd say `OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM` is better way to go (compared to allowing `LOG_OUTPUT_APP` be overridden).

bukepo: I think there are 3 layers in OpenThread: platform, core and application. The platform layer should not define app output, if that is defined, we should correct. The app output is defined in main object. NCP and CLI are application layer lib, so I think it makes sense to provide a weak log output implementation. I we define the app output in platform layer, is it breaking the layer boundaries?

abtink: I definitely agree on the having 3 layers: platform, core, app.

I'd say the whole `LOG_OUTPUT` config definition is rather unfortunate and causing this issue (mixing layers). I would be very happy if we get rid of it at some point. :)

In longer term, I prefer if all `OPENTHREAD_CONFIG_` definitions change/config something in core itself only. In this case, from core perspective in terms of logging, all it expects is that there is a platform API `otPlatLog()` which core would then use.  By the `otPlat` prefix in the name `otPlatLog`() we are indicating it should be a platform layer API. If we go by this model, the whole option of `LOG_OUTPUT_APP` is mixing the layers.

To me `LOG_OUTPUT_APP` is a way of saying, NCP or CLI executables would handle/provide the logging (over any implementation added at platform layer).  I do not interpret is as allowing more general app layer code (that is not just an NCP or CLI executable) to define `otPlatLog()`.

bukepo: I agree and I will make the otPlatLog() strong in NCP.

src/ncp/ncp_base.cpp
```diff
@@ -2393,27 +2393,7 @@ otError otNcpStreamWrite(int aStreamId, const uint8_t *aDataPtr, int aDataLen)
...
-        {
-            charsWritten = static_cast<int>(sizeof(logString) - 1);
-        }
-
-        IgnoreError(otNcpStreamWrite(0, reinterpret_cast<uint8_t *>(logString), charsWritten));
-    }
-}
-
-#if (OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_NCP_SPINEL)
+#if (OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_APP)
```
jwhui, line 2396:
```suggestion

#if (OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_APP)
```

examples/platforms/efr32mg12/system.c
```diff
@@ -59,9 +59,8 @@
 #include "fem-control.h"
 #endif
 
-#define USE_EFR32_LOG                                                                   \
-    ((OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED) || \
-     (OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_NCP_SPINEL))
+#define USE_EFR32_LOG \
+    ((OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED)
```
romacdon, line 63: @bukepo there's a typo in the about change.  There's an extra '(' at the beginning which results in this compilation error:

openthread/examples/../examples/platforms/efr32mg12/system.c:63:5: error: missing ')' in expression
     ((OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED)
     ^
openthread/examples/../examples/platforms/efr32mg12/system.c: In function 'otSysDeinit':
openthread/examples/../examples/platforms/efr32mg12/system.c:63:5: error: missing ')' in expression

examples/platforms/efr32mg13/system.c
```diff
@@ -59,9 +59,8 @@
 #include "fem-control.h"
 #endif
 
-#define USE_EFR32_LOG                                                                   \
-    ((OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED) || \
-     (OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_NCP_SPINEL))
+#define USE_EFR32_LOG \
+    ((OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED)
```
romacdon, line 63: same issue as above

examples/platforms/efr32mg21/system.c
```diff
@@ -59,9 +59,8 @@
 #include "fem-control.h"
 #endif
 
-#define USE_EFR32_LOG                                                                   \
-    ((OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED) || \
-     (OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_NCP_SPINEL))
+#define USE_EFR32_LOG \
+    ((OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED)
```
romacdon, line 63: same issue as above

examples/platforms/efr32mg12/system.c
romacdon: I created https://github.com/openthread/openthread/pull/5135 to address the issue.

PR #5105, opened 2020-06-16
[encoding] add uint64_t read/write methods for big/little endian encodings

abtink:


src/core/common/encoding.hpp
```diff
@@ -239,6 +274,22 @@ inline uint32_t ReadUint32(const uint8_t *aBuffer)
...
+ *
+ * @returns The `uint64_t` value read from buffer.
+ *
+ */
+inline uint64_t ReadUint64(const uint8_t *aBuffer)
+{
+    return ((static_cast<uint64_t>(aBuffer[0]) << 0) | (static_cast<uint64_t>(aBuffer[1]) << 8) |
+            (static_cast<uint64_t>(aBuffer[2]) << 16) | (static_cast<uint64_t>(aBuffer[3]) << 24) |
+            (static_cast<uint64_t>(aBuffer[4]) << 32) | (static_cast<uint64_t>(aBuffer[5]) << 40) |
+            (static_cast<uint64_t>(aBuffer[6]) << 48) | (static_cast<uint64_t>(aBuffer[7]) << 56));
```
wgtdkp, line 290: nit:
```suggestion
    return ReadUint32(aBuffer) | (static_cast<uint64_t>(ReadUint32(aBuffer + 4)) << 32);
```

abtink: We don't use this model for other methods here (e.g., we don't implement `ReadUint32()` in terms of `ReadUint16()`). I think this is mainly to help with compilers (particularly older toolchains) doing better optimization of these inline methods.

While I like the smaller/shorter implementation, to keep it consistent with other methods and help with compiler optimization, I prefer/suggest to keep it as is.

wgtdkp: okay.

PR #5068, opened 2020-06-08
[posix] fix spinel radio state transition

wgtdkp:
Assetion failure triggered in `SubMac` by below commands:
```shell
## Build ot-cli
make -f src/posix/Makefile-posix

## Build ot-rcp
make -f examples/Makefile-nrf52840
arm-none-eabi-objcopy -O ihex output/nrf52840/bin/ot-rcp ot-rcp.hex
nrfjprog -f nrf52 --chiperase --reset --program ot-rcp.hex
nrfjprog -f nrf52 --pinresetenable
nrfjprog -f nrf52 --reset

# Disable MSD
expect <<EOF
spawn JLinkExe
expect "J-Link>"
send "msddisable\n"
expect "Probe configured successfully."
exit
EOF

## Run ot-rcp
./output/posix/x86_64-unknown-linux-gnu/bin/ot-cli 'spinel+hdlc+uart:///dev/ttyACM0?uart-baudrate=115200'
# > scan
# | J | Network Name     | Extended PAN     | PAN  | MAC Address      | Ch | dBm | LQI |
# +---+------------------+------------------+------+------------------+----+-----+-----+
# > ot-cli: /usr/local/google/home/wgtdkp/openthread/src/posix/../../src/core/mac/sub_mac.cpp:336: void ot::Mac::SubMac::BeginTransmit(): Assertion `error == OT_ERROR_NONE' failed.
# [1]    144072 abort      ./output/posix/x86_64-unknown-linux-gnu/bin/ot-cli 
```

The root cause is that `RadioSpinel` doesn't handle the situation that transition from state `Sleep` to `Transmit` is allowed if it has `OT_RADIO_CAPS_SLEEP_TO_TX`  capability.

https://github.com/openthread/openthread/blob/fb39910402c6074379bc5788fd8f454c6b5e683a/include/openthread/platform/radio.h#L262-L277

Related issue: https://github.com/openthread/ot-esp32/issues/6.

PR #5108, opened 2020-06-16
[link-raw] set receive done callback when enabled

bukepo:
This fixes openthread/ot-esp32#6.

### DIFF
--- a/examples/platforms/simulation/radio.c
+++ b/examples/platforms/simulation/radio.c
@@ -117,6 +117,8 @@ static bool sSrcMatchEnabled = false;
 static bool sRadioCoexEnabled = true;
 #endif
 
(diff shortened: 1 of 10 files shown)


bukepo:
This also fixes the scan issue in https://groups.google.com/g/openthread-users/c/rfMMH178CvI/m/BrvcRal9BgAJ

bukepo:
Should be introduced by #5068, because transmit will not call receive now if the radio supports.

jwhui:
> Should be introduced by #5068, because transmit will not call receive now if the radio supports.

Can we add a continuous integration check to make sure the scan function does not break in the future?

bukepo:
> > Should be introduced by #5068, because transmit will not call receive now if the radio supports.
> 
> Can we add a continuous integration check to make sure the scan function does not break in the future?

Added the check as an expect test.

PR #4981, opened 2020-05-18
[meshcop] do not send MGMT_*_SET.req while one is outstanding

jwhui:
This commit helps ensure the following:

- At most one MGMT_ACTIVE_SET.req message is outstanding.

- At most one MGMT_PENDING_SET.req message is outstanding.

PR #5110, opened 2020-06-17
[commissioner] prefer entry with Joiner ID match over one accepting any

abtink:
This commit includes a group of smaller enhancements in commissioner
module:

It adds `FindBestMatchingJoinerEntry()` which finds the best matching
joiner entry preferring an entry with full Joiner ID match over an
entry accepting any joiner.

`AddJoiner()` is changed to first try to find and update an existing
entry matching the given EUI64 before allocating a new entry. This
change in turn removes the need to call `RemoveJoiner()` from
`AddJoiner()` and the use of `JoinerOpFlag` to skip notifying the
leader of the unnecessary remove.

This commit also simplifies updating of joiner expiration timer,
changes the `SendCommissionerSet()` to be `void` (to avoid
`IgnoreError()`), adopts the use of `Ip6::InterfaceIdentifier` type to
store Joiner IID, and adds method `LogJoinerEntry()` to help with
logging of Joiner List updates.

----------

Kept some of the commits separate to help with the review, but would suggest to squash them all. The first commit  contains the full log message (for all the changes).

### DIFF
--- a/src/core/meshcop/commissioner.cpp
+++ b/src/core/meshcop/commissioner.cpp
@@ -139,10 +139,110 @@ void Commissioner::HandleCoapsConnected(bool aConnected)
 
     event = aConnected ? OT_COMMISSIONER_JOINER_CONNECTED : OT_COMMISSIONER_JOINER_END;
 
-    joinerId.Set(mJoinerIid);
+    joinerId.Set(mJoinerIid.m8);
     joinerId.ToggleLocal();
 
     SignalJoinerEvent(event, joinerId);
+}
+
+Commissioner::Joiner *Commissioner::GetUnusedJoinerEntry(void)
+{
+    Joiner *joiner;
+
+    for (joiner = &mJoiners[0]; joiner < OT_ARRAY_END(mJoiners); joiner++)
+    {
+        if (!joiner->mValid)
+        {
+            ExitNow();
+        }
+    }
+
+    joiner = nullptr;
+
+exit:
+    return joiner;
+}
+
+Commissioner::Joiner *Commissioner::FindJoinerEntry(const Mac::ExtAddress *aEui64)
+{
+    Joiner *joiner;
+
+    for (joiner = &mJoiners[0]; joiner < OT_ARRAY_END(mJoiners); joiner++)
+    {
+        if (!joiner->mValid)
+        {
+            continue;
+        }
+
+        if (aEui64 == nullptr)
+        {
+            if (joiner->mAny)
+            {
+                ExitNow();
+            }
+        }
+        else
+        {
+            if (!joiner->mAny && (joiner->mEui64 == *aEui64))
+            {
+                ExitNow();
+            }
+        }
+    }
+
+    joiner = nullptr;
+
+exit:
+    return joiner;
+}
+
+Commissioner::Joiner *Commissioner::FindBestMatchingJoinerEntry(const Mac::ExtAddress &aReceivedJoinerId)
+{
+    Joiner *best = nullptr;
+
+    // Prefer a full Joiner ID match, if not found use the entry
+    // accepting any joiner.
+
+    for (Joiner *joiner = &mJoiners[0]; joiner < OT_ARRAY_END(mJoiners); joiner++)
+    {
+        if (!joiner->mValid)
+        {
+            continue;
+        }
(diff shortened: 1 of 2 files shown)


src/core/meshcop/commissioner.cpp
```diff
@@ -322,14 +322,13 @@ otError Commissioner::Stop(bool aResign)
     return error;
 }
 
-otError Commissioner::SendCommissionerSet(void)
+void Commissioner::SendCommissionerSet(void)
```
jwhui, line 325: Not directly related to this PR, but we will want to consider the case where COMMISSIONER_SET was not delivered to the Leader for some reason.

src/core/meshcop/commissioner.cpp
```diff
@@ -1169,6 +1169,18 @@ const char *Commissioner::StateToString(otCommissionerState aState)
...
+                     aJoiner.mPsk);
+}
+
+#else
+
+void Commissioner::LogJoinerEntry(const char *, const Joiner &) const
+{
+}
+
 #endif // (OPENTHREAD_CONFIG_LOG_LEVEL >= OT_LOG_LEVEL_INFO) && (OPENTHREAD_CONFIG_LOG_MLE == 1)
```
jwhui, line 1182: Not related to this PR, but should this conditional be on `OPENTHREAD_CONFIG_LOG_MESHCOP`?

src/core/meshcop/commissioner.cpp
```diff
@@ -145,6 +145,106 @@ void Commissioner::HandleCoapsConnected(bool aConnected)
     SignalJoinerEvent(event, joinerId);
 }
 
+Commissioner::Joiner *Commissioner::AllocateJoinerEntry(void)
```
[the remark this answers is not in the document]
abtink: Good point! Will change it.

src/core/meshcop/commissioner.cpp
abtink: Shall we keep the the `otError`? I guess we probably would handle the failure (retx maybe)  from the method itself, so may be fine to keep it `void`. What do you think?

src/core/meshcop/commissioner.cpp
abtink: Indeed. It does not seem to be correct. Will fix it. Thanks.

abtink: Turns out we don't actually have `OPENTHREAD_CONFIG_LOG_MESHCOP` and the  meshcop logging macros are tied to the `OPENTHREAD_CONFIG_LOG_MLE` config:
https://github.com/openthread/openthread/blob/3b362071e594375035e063c0c25988b1aa52bd5f/src/core/common/logging.hpp#L291-L303

I think we may want to address this, but probably from a different PR.

src/core/meshcop/commissioner.cpp
jwhui: I think changing to `void` is fine.

src/core/meshcop/commissioner.cpp
jwhui: Good catch. We should clean this up in a separate PR.

src/core/meshcop/commissioner.cpp
simonlingoogle: We can grant only one outstanding `SendCommissionerSet` just like what @jwhui  did for `MGMT_*_SET.req` in #4981 .

src/core/meshcop/commissioner.cpp
```diff
@@ -258,10 +357,10 @@ otError Commissioner::SendCommissionerSet(void)
 
     dataset.mIsSteeringDataSet = true;
 
-    SuccessOrExit(error = SendMgmtCommissionerSetRequest(dataset, nullptr, 0));
+    IgnoreError(SendMgmtCommissionerSetRequest(dataset, nullptr, 0));
```
wgtdkp, line 360: Instead of ignoring the error, what about logging it?

abtink: Updated it in the new pushed commit to log the error. 
We later probably need a mechanism on to handle the failure (retry, etc). Perhaps in a future PR.

PR #5113, opened 2020-06-17
[autoconf] add back OPENTHREAD_TARGET_LINUX

gjc13:
OPENTHREAD_TARGET_LINUX is checked in platform makefiles to determie ld
libraries.

https://github.com/openthread/openthread/blob/f5d7f50f44fce9e872bdda2a040a49ab91e36623/examples/platforms/simulation/Makefile.platform.am#L37-L41


gjc13:
Close since this change is not resolving the issue we've encountered.

PR #5115, opened 2020-06-17
[docker] build ot-daemon/ot-ctl in codelab_otsim

Vyrastas:
To support the updated Docker Simulation Codelab.

### DIFF
--- a/etc/docker/codelab_otsim/Dockerfile
+++ b/etc/docker/codelab_otsim/Dockerfile
@@ -37,3 +37,6 @@ RUN cd ~/src && \
     cd openthread && \
     ./bootstrap && \
     make -f examples/Makefile-simulation
+
+# Install OpenThread Daemon and ot-ctl
+RUN make -f src/posix/Makefile-posix DAEMON=1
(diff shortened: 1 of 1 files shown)


Vyrastas:
> Should we remove `wpantund` from the codelab?

Since `wpantund` has been removed from the Codelab, I suppose we can just remove from here too.  I wasn't sure whether we wanted to keep it for legacy purposes.

Is `wpantund` going to be completely deprecated and not used at all in the future, or will it still be used as long as we have the NCP design?

If I do remove, should I remove all of lines 9-32 then?

jwhui:
> Since `wpantund` has been removed from the Codelab, I suppose we can just remove from here too. I wasn't sure whether we wanted to keep it for legacy purposes.
> 
> Is `wpantund` going to be completely deprecated and not used at all in the future, or will it still be used as long as we have the NCP design?

I think the general preference is to prefer RCP where applicable. The primary use cases I see for NCP is where the host processor needs to sleep (seems like a rare use case at the moment) or where the host processor is an RTOS and does not want the complexity of also running OpenThread on the same chip (does not use wpantund).

> If I do remove, should I remove all of lines 9-32 then?

In general, I think that's right. We may need to keep some packages for building OpenThread. Running `./script/bootstrap` instead may be the better approach.

Vyrastas:
I've removed `wpantund` and update the packages accordingly to make the build work with ./script/bootstrap.  PTAL

I had to add the ca-certificates / update-ca-certificates part to get it to git clone on my local system, not sure if that's required for CI.

PR #5116, opened 2020-06-18
[network-data] only send ADDR_QRY.qry for on-mesh prefixes

### LINKED ISSUE #5112: Should not send ADDR_QRY.qry for an off-mesh destination
**Describe the bug**
OT sends `ADDR_QRY.qry` for destination address of `off-mesh` prefix. 

**To Reproduce**
1. Git commit id: master
2. IEEE 802.15.4 hardware platform: OTNS
3. Steps: 
```bash
make -f example/Makefile-simulation OTNS=1

$ otns
> add router x 100 y 100
1
Done
> add router x 100 y 200
2
Done
> node 1 "prefix add 2001::/64 pars med"
Done
> node 1 "netdataregister"
Done
> node 1 "ipaddr"
2001:0:0:0:59eb:2331:ac03:2b7e
fdde:ad00:beef:0:0:ff:fe00:3c00
fdde:ad00:beef:0:cad2:dcc0:d14d:163c
fe80:0:0:0:bcdf:1bb4:f60f:fbdd
Done
> node 2 "ipaddr"
2001:0:0:0:e5a0:be90:273:71c
fdde:ad00:beef:0:0:ff:fe00:c000
fdde:ad00:beef:0:4228:c5af:aa9c:2e97
fe80:0:0:0:2c81:d84d:47e:820d
Done
> ping 1 "2001:0:0:0:e5a0:be90:273:71c"
Done
> exit
```

Use wireshark to find the `ADDR_QRY.qry` in the pcap: [current.pcap.zip](https://github.com/openthread/openthread/files/4793285/current.pcap.zip)

4. Network topology
```Leader---Router```

**Expected behavior**
OT should not send `ADDR_QRY.qry` for `off-mesh` destination according to `Spec. 5.15.7 Router Behavior`:
```
22 If the IPv6 Destination Address is not on-mesh, then the Router uses the IPv6 Source
23 Address to determine the set of Border Routers that offer routes associated with the same
24 Provisioning Domain as the prefix of the IPv6 Source Address.
```

I also check the case when the `off-mesh` prefix is **NOT** using SLAAC, `ADDR_QRY.qry` is still sent for a fake `off-mesh` destination, as shown in pcap: [current_no_slaac.pcap.zip](https://github.com/openthread/openthread/files/4793320/current.pcap.zip)

Is this expected behavior for off-mesh destinations? 
Thoughts? @jwhui @bukepo @librasungirl

### PR DESCRIPTION
**Pull request by jwhui**
This PR includes the following fixes.

- fix `NetworkData::Leader::IsOnMesh()` to check on-mesh flags.

- fix tests to specify on-mesh flag in prefix 3.

- fix SLAAC to set prefix length to /128 when prefix is not on mesh.

Resolves #5112

### DIFF
--- a/src/core/thread/network_data_leader.cpp
+++ b/src/core/thread/network_data_leader.cpp
@@ -260,12 +260,25 @@ bool LeaderBase::IsOnMesh(const Ip6::Address &aAddress) const
 
     while ((prefix = FindNextMatchingPrefix(aAddress, prefix)) != nullptr)
     {
-        if (FindBorderRouter(*prefix) == nullptr)
+        // check both stable and temporary Border Router TLVs
+        for (int i = 0; i < 2; i++)
         {
-            continue;
-        }
+            const BorderRouterTlv *borderRouter = FindBorderRouter(*prefix, /* aStable */ (i == 0));
 
-        ExitNow(rval = true);
+            if (borderRouter == nullptr)
+            {
+                continue;
+            }
+
+            for (const BorderRouterEntry *entry = borderRouter->GetFirstEntry(); entry <= borderRouter->GetLastEntry();
+                 entry                          = entry->GetNext())
+            {
+                if (entry->IsOnMesh())
+                {
+                    ExitNow(rval = true);
+                }
+            }
+        }
     }
 
 exit:
--- a/src/core/utils/slaac_address.cpp
+++ b/src/core/utils/slaac_address.cpp
@@ -138,6 +138,14 @@ exit:
     return;
 }
 
+bool Slaac::DoesConfigMatchNetifAddr(const NetworkData::OnMeshPrefixConfig &aConfig,
+                                     const Ip6::NetifUnicastAddress &       aAddr)
+{
+    return (((aConfig.mOnMesh && (aAddr.mPrefixLength == aConfig.mPrefix.mLength)) ||
+             (!aConfig.mOnMesh && (aAddr.mPrefixLength == 128))) &&
+            (aAddr.GetAddress().PrefixMatch(aConfig.mPrefix.mPrefix) >= aConfig.mPrefix.mLength));
+}
+
 void Slaac::Update(UpdateMode aMode)
 {
     NetworkData::Iterator           iterator;
@@ -165,16 +173,13 @@ void Slaac::Update(UpdateMode aMode)
 
                 while (Get<NetworkData::Leader>().GetNextOnMeshPrefix(iterator, config) == OT_ERROR_NONE)
                 {
-                    otIp6Prefix &prefix = config.mPrefix;
-
                     if (config.mDp)
                     {
                         // Skip domain prefix which is processed in MLE.
                         continue;
                     }
 
-                    if (config.mSlaac && !ShouldFilter(prefix) && (prefix.mLength == slaacAddr->mPrefixLength) &&
-                        (slaacAddr->GetAddress().PrefixMatch(prefix.mPrefix) >= prefix.mLength))
+                    if (config.mSlaac && !ShouldFilter(config.mPrefix) && DoesConfigMatchNetifAddr(config, *slaacAddr))
                     {
                         found = true;
                         break;
@@ -212,8 +217,7 @@ void Slaac::Update(UpdateMode aMode)
             for (const Ip6::NetifUnicastAddress *netifAddr = Get<ThreadNetif>().GetUnicastAddresses();
                  netifAddr != nullptr; netifAddr           = netifAddr->GetNext())
             {
-                if ((netifAddr->mPrefixLength == prefix.mLength) &&
-                    (netifAddr->GetAddress().PrefixMatch(prefix.mPrefix) >= prefix.mLength))
+                if (DoesConfigMatchNetifAddr(config, *netifAddr))
                 {
                     found = true;
                     break;
@@ -234,7 +238,7 @@ void Slaac::Update(UpdateMode aMode)
                     slaacAddr->Clear();
                     memcpy(&slaacAddr->mAddress, &prefix.mPrefix, BitVectorBytes(prefix.mLength));
 
-                    slaacAddr->mPrefixLength  = prefix.mLength;
+                    slaacAddr->mPrefixLength  = config.mOnMesh ? prefix.mLength : 128;
                     slaacAddr->mAddressOrigin = OT_ADDRESS_ORIGIN_SLAAC;
                     slaacAddr->mPreferred     = config.mPreferred;
                     slaacAddr->mValid         = true;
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: src/core/thread/network_data_leader.cpp [hunk 1]
```diff
@@ -260,12 +260,26 @@ bool LeaderBase::IsOnMesh(const Ip6::Address &aAddress) const
...
+            {
+                continue;
+            }
 
-        ExitNow(rval = true);
+            for (uint8_t entry = 0; entry < borderRouter->GetNumEntries(); entry++)
+            {
+                const BorderRouterEntry *borderRouterEntry = borderRouter->GetEntry(entry);
+
+                if (borderRouterEntry->IsOnMesh())
```
abtink, line 277: nit: Works fine as is. Suggestion to use `entry` pointer?
```suggestion
            for (const BorderRouterEntry *entry = borderRouter->GetFirstEntry(); entry <= borderRouter->GetLastEntry();
                 entry                          = entry->GetNext())
            {
                if (entry->IsOnMesh())
```

src/core/thread/network_data_leader.cpp
```diff
@@ -260,12 +260,26 @@ bool LeaderBase::IsOnMesh(const Ip6::Address &aAddress) const
 
     while ((prefix = FindNextMatchingPrefix(aAddress, prefix)) != NULL)
     {
-        if (FindBorderRouter(*prefix) == NULL)
+        // check both stable and full Border Router TLVs
```
simonlingoogle, line 263:
```suggestion
        // check both stable and unstable Border Router TLVs
```
Do we call unstable `full` ?

src/core/utils/slaac_address.cpp
```diff
@@ -234,7 +236,7 @@ void Slaac::Update(UpdateMode aMode)
                     slaacAddr->Clear();
                     memcpy(&slaacAddr->mAddress, &prefix.mPrefix, BitVectorBytes(prefix.mLength));
 
-                    slaacAddr->mPrefixLength  = prefix.mLength;
+                    slaacAddr->mPrefixLength  = config.mOnMesh ? prefix.mLength : 128;
```
simonlingoogle, line 239: The code and comment seems to mean that OT only generate SLAAC addresses for `on-mesh` prefixes. But the actual runtime behavior is that OT also do SLAAC for non `on-mesh` prefixes.

tests/scripts/thread-cert/Cert_5_6_05_NetworkDataRegisterAfterAttachRouter.py
```diff
@@ -109,7 +109,7 @@ def test(self):
             if addr[0:3] == '200':
                 self.assertTrue(self.nodes[LEADER].ping(addr))
 
-        self.nodes[ROUTER].add_prefix('2001:2:0:3::/64', 'pacs')
+        self.nodes[ROUTER].add_prefix('2001:2:0:3::/64', 'paos')
```
simonlingoogle, line 112: Add some test to verify that OT handles `off-mesh` prefixes properly?

src/core/utils/slaac_address.cpp
simonlingoogle, line 239: What if this prefix config is `off-mesh` but there is another same prefix afterwards but is `on-mesh`. I think we need to generate SLAAC address of prefix  of `prefix.mLength` rather than 128 for this prefix. 
Seems we need the full information of a prefix to generate the correct address?

tests/scripts/thread-cert/test_on_mesh_prefix.py
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env python3
+#
+#  Copyright (c) 2016, The OpenThread Authors.
```
simonlingoogle, line 3:
```suggestion
#  Copyright (c) 2020, The OpenThread Authors.
```

src/core/thread/network_data_leader.cpp
jwhui: Changed to "temporary", which is the term the Thread Specification uses.

tests/scripts/thread-cert/Cert_5_6_05_NetworkDataRegisterAfterAttachRouter.py
jwhui: Added a test in a new commit.

src/core/utils/slaac_address.cpp
jwhui: I think there are two definitions for on-mesh:

1. A prefix that is assigned to the Thread network.

1. A prefix that requires using ADDR_QRY.qry when forwarding.

Both are proper uses of on-mesh. I don't have a good term that disambiguates them.

tests/scripts/thread-cert/test_on_mesh_prefix.py
```diff
@@ -0,0 +1,142 @@
...
+
+LEADER = 1
+ROUTER = 2
+ED1 = 3
+SED1 = 4
+
+MTDS = [ED1, SED1]
+
+
+class Test_OnMeshPrefixr(thread_cert.TestCase):
```
simonlingoogle, line 43:
```suggestion
class Test_OnMeshPrefix(thread_cert.TestCase):
```

src/core/utils/slaac_address.cpp
jwhui: Added a check above to ensure we will generate a new address.

simonlingoogle: I think we should follow the same definition of `on-mesh` as Thread Spec., in which quote `A prefix is "on-mesh" if packets with on-mesh destinations are routed within the Thread subnet`. 
So 2) is more correct according to this definition? 

Currently, `GetNextOnMeshPrefix` visits both `on-mesh` prefixes and `off-mesh` prefixes. I think we can simply rename it to `GetNextPrefix` to reduce confusion?

jwhui: If the destination is outside the Thread network, the destination does not match any prefix assigned to the Thread network. As a result, I don't consider any prefix assigned to the Thread network as "off-mesh".

This is equivalent to traditional IPv6. A prefix may be assigned to a link. Separately, a prefix may be declared as "on-link", which indicates to use IPv6 ND to discover the destination. A prefix not declared as "on-link" just means "don't use IPv6 ND to discover the destination". However, packets destined to such prefixes do not leave the link (since the prefix is assigned to the link).

src/core/utils/slaac_address.cpp
```diff
@@ -173,7 +173,9 @@ void Slaac::Update(UpdateMode aMode)
                         continue;
                     }
 
-                    if (config.mSlaac && !ShouldFilter(prefix) && (prefix.mLength == slaacAddr->mPrefixLength) &&
+                    if (config.mSlaac && !ShouldFilter(prefix) &&
+                        ((config.mOnMesh && (slaacAddr->mPrefixLength == prefix.mLength)) ||
+                         (!config.mOnMesh && (slaacAddr->mPrefixLength == 128))) &&
```
simonlingoogle, line 178: Define a new method like `OnMeshPrefixConfig::Match` to check if the address is matched by the Prefix to encapsulate this condition?

PR #5119, opened 2020-06-18
[posix] only update timeout with a lower value

gjc13:


src/posix/platform/alarm.cpp
```diff
@@ -159,8 +159,11 @@ void platformAlarmUpdateTimeout(struct timeval *aTimeout)
             remaining = 1;
         }
 
-        aTimeout->tv_sec  = (time_t)(remaining / US_PER_S);
-        aTimeout->tv_usec = remaining % US_PER_S;
+        if (remaining < aTimeout->tv_sec * US_PER_S + aTimeout->tv_usec)
+        {
+            aTimeout->tv_sec  = (time_t)(remaining / US_PER_S);
```
bukepo, line 164:
```suggestion
            aTimeout->tv_sec  = static_cast<time_t>(remaining / US_PER_S);
```

PR #5122, opened 2020-06-19
[posix] use real time signal for microsecond timer

bukepo:


src/posix/main.c
```diff
@@ -194,6 +205,18 @@ static void ParseArg(int aArgCount, char *aArgVector[], PosixConfig *aConfig)
...
             break;
+#if __linux__
+        case OT_POSIX_OPT_REAL_TIME_SIGNAL:
+            if (optarg[0] == '+')
+            {
+                aConfig->mPlatformConfig.mRealTimeSignal = SIGRTMIN + atoi(&optarg[1]);
+            }
+            else
+            {
+                aConfig->mPlatformConfig.mRealTimeSignal = atoi(optarg);
```
Irving-cl, line 216: I checked the previous code, the micro timer signal is always set to `OPENTHREAD_CONFIG_MICRO_TIMER_SIGNAL` and is by default set to `SUGRTMIN`.  And it is determined at compile time. Now it is determined at run-time. What is the consideration of this? I think the signal won't change after the program starts to run.
And shall we ensure that the assigned signal number doesn't collide with other signals? Like the interrupt signal. Will it cause some problems?

bukepo: If we want to distribute OpenThread POSIX as a library, we should allow users to pass in this value, which can be compile time defined or run-time given in the application code.

src/posix/platform/alarm.cpp
```diff
@@ -70,9 +91,43 @@ static uint64_t platformAlarmGetNow(void)
...
+        sa.sa_sigaction = microTimerHandler;
+        sigemptyset(&sa.sa_mask);
+
+        VerifyOrDie(sigaction(aRealTimeSignal, &sa, NULL) != -1, OT_EXIT_ERROR_ERRNO);
+
+        sev.sigev_notify          = SIGEV_SIGNAL;
+        sev.sigev_signo           = aRealTimeSignal;
+        sev.sigev_value.sival_ptr = &sMicroTimer;
+
+        VerifyOrDie(timer_create(CLOCK_REALTIME, &sev, &sMicroTimer) != -1, OT_EXIT_ERROR_ERRNO);
```
brianljt, line 120: Should we use `CLOCK_MONOTONIC_RAW` or `CLOCK_MONOTONIC` as `otPlatTimeGet`?

PR #5125, opened 2020-06-19
[dtls] increase maximum length of DTLS application data

wgtdkp:
This PR increases the maximum allowed length of DTLS application data from `512` bytes to `572` bytes.

The maximum possible length of application data is just around 512 bytes, and we observed that a `Client Hello` with cookie encoded in RLY_RX.ntf message could be 520 bytes.

Update:
The maximum/minimum application data length is agreed to be 1152 bytes.


simonlingoogle:
Is this considered as a bug? Any impact on certification?

wgtdkp:
> Is this considered as a bug? Any impact on certification?

I didn't see a failure in certification, but encountered it when doing external commissioning with latest OTBR (RCP) solutions, by following instructions at https://openthread.io/guides/border-router/external-commissioning?comm=ot-commissioner

jwhui:
> I didn't see a failure in certification, but encountered it when doing external commissioning with latest OTBR (RCP) solution, by following instructions at https://openthread.io/guides/border-router/external-commissioning?comm=ot-commissioner

Any idea on what the difference may be? Why is there a difference in the packet size between the different cases?

wgtdkp:
> > I didn't see a failure in certification, but encountered it when doing external commissioning with latest OTBR (RCP) solution, by following instructions at https://openthread.io/guides/border-router/external-commissioning?comm=ot-commissioner
> 
> Any idea on what the difference may be? Why is there a difference in the packet size between the different cases?

@jwhui certification test cases are using on-mesh commissioner and there is no Border Agent used there (so no DTLS session).

jwhui:
> @jwhui certification test cases are using on-mesh commissioner and there is no Border Agent used there (so no DTLS session).

@wgtdkp , we have also tested with Thread Group Android App as external commissioner and that seemed to work as well. Any difference between the Thread Group Commissioning App and ot-commissioner that can account for the difference? Part of the question here is how do we know whether increasing to 572 bytes is enough?

wgtdkp:
> we have also tested with Thread Group Android App as external commissioner and that seemed to work as well. Any difference between the Thread Group Commissioning App and ot-commissioner that can account for the difference? Part of the question here is how do we know whether increasing to 572 bytes is enough?

@jwhui This reminds me that integration tests in OT Commissioner has been always successful. I will debug to see what is the cause of the difference. Mark this PR to `WIP` for now, will get back soon.

wgtdkp:
@jwhui  I created https://github.com/openthread/openthread/issues/5142 to provide more bakcground/details.

> Part of the question here is how do we know whether increasing to 572 bytes is enough?

Increasing the value to `572` doesn't seem to be a good fix because a DTLS handshake message can be as long as `2^24 -1` bytes.
I am not sure why the original max length is set to 512, maybe because it is the minimum value that works with our mbedtls configuration?

edmont:
> Increasing the value to `572` doesn't seem to be a good fix because a DTLS handshake message can be as long as `2^24 -1` bytes.
> I am not sure why the original max length is set to 512, maybe because it is the minimum value that works with our mbedtls configuration?

Maybe this should be addressed at Thread Specification? Some definition could be added in Chapter 7 to limit the size of DTLS messages to a reasonable value for Thread Networks.

jwhui:
> Maybe this should be addressed at Thread Specification? Some definition could be added in Chapter 7 to limit the size of DTLS messages to a reasonable value for Thread Networks.

I agree. @wgtdkp , can you submit a Thread Group JIRA for this?

wgtdkp:
> > Maybe this should be addressed at Thread Specification? Some definition could be added in Chapter 7 to limit the size of DTLS messages to a reasonable value for Thread Networks.
> 
> I agree. @wgtdkp , can you submit a Thread Group JIRA for this?

Will do.

jwhui:
This issue is being discussed in [SPEC-942](https://threadgroup.atlassian.net/browse/SPEC-942). Current proposal is to specify 1152 bytes.

@wgtdkp , would you like to update this PR with the 1152 byte value?

wgtdkp:
> This issue is being discussed in [SPEC-942](https://threadgroup.atlassian.net/browse/SPEC-942). Current proposal is to specify 1152 bytes.
> 
> @wgtdkp , would you like to update this PR with the 1152 byte value?

Sure, updated.

PR #5130, opened 2020-06-20
[script] use real time when run expect tests

bukepo:


script/test
```diff
@@ -328,6 +327,16 @@ envsetup()
         fi
     fi
 
+    # All expect tests only works in real time mode.
+    VIRTUAL_TIME=1
```
simonlingoogle, line 331: this will override env set by user? e.g. VIRTUAL_TIME=0 ./script/test build

script/test
```diff
@@ -328,6 +327,16 @@ envsetup()
...
+    for arg in "$@"; do
+        if [[ $arg == expect ]]; then
+            VIRTUAL_TIME=0
```
simonlingoogle, line 334: it's convenient to set VIRTUAL_TIME=0 by default for `expect` tests. But this also bring extra complexity to this `test` script. Maybe just using`VIRTUAL_TIME=0 ./script/test expect` is more consistent?

PR #5131, opened 2020-06-20
[mle] add discovery request callback

bukepo:
This commit allows notifying the application a discovery request is received, which is useful if the application wants to be more active when a joiner comes.


src/core/thread/mle_router.cpp
```diff
@@ -2842,6 +2847,17 @@ void MleRouter::HandleDiscoveryRequest(const Message &aMessage, const Ip6::Messa
 
     error = SendDiscoveryResponse(aMessageInfo.GetPeerAddr(), aMessage.GetPanId());
 
+    if (mDiscoveryRequestCallback != nullptr && hasDiscoveryRequest)
```
simonlingoogle, line 2850: Can we give `mDiscoveryRequestCallback` a chance to determine how the node will handle the `DiscoveryRequest` ? For example, if `mDiscoveryRequestCallback` returns `false`, we don't `SendDiscoveryResponse`.

bukepo: I also considered this, but I don't have a use case of it for now. I prefer not adding a feature before we are going to use it, otherworldly it only increases maintenance burden and decreases code coverage. Thoughts?

include/openthread/thread.h
```diff
@@ -781,6 +781,37 @@ void otThreadRegisterParentResponseCallback(otInstance *                   aInst
...
+/**
+ * This structure represents the Thread Discovery Request data.
+ *
+ */
+typedef struct otThreadDiscoveryRequestInfo
+{
+    const otExtAddress *mExtAddress;   ///< IEEE 802.15.4 Extended Address of the requester
+    uint8_t             mVersion : 4;  ///< Thread version.
+    bool                mIsJoiner : 1; ///< Whether is from joiner.
+} otThreadDiscoveryRequestInfo;
```
abtink, line 793: @bukepo, the implementation overall looks good.

Can you please help give a bit more detail on how we can use this (how it would help)?

One thing is that, in current code when a joiner requests a discover scan, it uses a randomly generated ext address and not the Joiner ID (was recently changed from #4920) which I guess required by spec (?). So we cannot really tell apart joiners when they issue scan from their ext-address.

include/openthread/thread.h
```diff
@@ -781,6 +781,37 @@ void otThreadRegisterParentResponseCallback(otInstance *                   aInst
...
+
+/**
+ * This function pointer is called every time an MLE Discovery Request message is received.
+ *
+ * @param[in]  aInfo     A pointer to a location on stack holding the Discovery Request info data.
```
Irving-cl, line 798:
```suggestion
 * @param[in]  aInfo     A pointer to the Discovery Request info data.
```

include/openthread/thread.h
```diff
@@ -781,6 +781,37 @@ void otThreadRegisterParentResponseCallback(otInstance *                   aInst
...
+ * @param[in]  aContext  A pointer to callback application-specific context.
+ *
+ */
+typedef void (*otThreadDiscoveryRequestCallback)(const otThreadDiscoveryRequestInfo *aInfo, void *aContext);
+
+/**
+ * This function sets a callback to receive MLE Discovery Request data.
```
Irving-cl, line 805:
```suggestion
 * This function sets a callback called when MLE Discovery Request data is received.
```

include/openthread/thread.h
simonlingoogle: Yes, #4920 was required by the Spec.

bukepo: This is useful when https://github.com/openthread/openthread/issues/5175 is implemented.

include/openthread/thread.h
```diff
@@ -781,6 +781,37 @@ void otThreadRegisterParentResponseCallback(otInstance *                   aInst
                                             otThreadParentResponseCallback aCallback,
                                             void *                         aContext);
 
+/**
+ * This structure represents the Thread Discovery Request data.
+ *
+ */
+typedef struct otThreadDiscoveryRequestInfo
+{
+    const otExtAddress *mExtAddress;   ///< IEEE 802.15.4 Extended Address of the requester
```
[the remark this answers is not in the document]
bukepo: updated.

bukepo:
I reverted the url changes. @abtink @lanyuwen @jwhui

src/core/thread/mle_router.cpp
```diff
@@ -2853,7 +2855,7 @@ void MleRouter::HandleDiscoveryRequest(const Message &aMessage, const Ip6::Messa
     otError                      error = OT_ERROR_NONE;
     Tlv                          tlv;
     MeshCoP::Tlv                 meshcopTlv;
-    MeshCoP::DiscoveryRequestTlv discoveryRequest;
+    MeshCoP::DiscoveryRequestTlv discoveryRequest{};
```
abtink, line 2858: I guess you are using `{ }` to zero-initialze the object.

I would suggest not doing this and instead setting the legth explicitly to zero in the code.

Two reasons:
- It's more explicit and clear in the code (also we just care about length and not rest of field to be zero)
- I don't  know in detail the rules on different initializations in different version of c++, but from what I recall, it's a bit messy/tricky. I think if we  add a default constructor (or just another constructor) to this `Tlv` or any of its sub-classes, the behavior may change and we may not get the zero initialization of object anymore.

PR #5134, opened 2020-06-20
[posix] ot-ctl process lines for non-interactive mode

simonlingoogle:
This PR enhance `ot-ctl` to process only after a full line is read in non-interactive mode. 

This eliminates the complex and error-prone states handling, which still has potential issues (despite of small chance):

- If `Done\r\n` is received in two chunks: `Done\r` and `\n`, the original implementation won't output `\n`
- If `Error 7: InvalidArgs` is received in two chunks: `Error ` and `7: InvalidArgs`, the original implementation won't output `7: InvalidArgs`

### DIFF
--- a/src/posix/client.cpp
+++ b/src/posix/client.cpp
@@ -59,7 +59,11 @@
 
 #include "platform-posix.h"
 
-#define MAXLINE 4096
+const size_t LINE_BUFFER_SIZE = 256;
+
+OT_STATIC_ASSERT(LINE_BUFFER_SIZE >= sizeof("> "), "LINE_BUFFER_SIZE is too small");
+OT_STATIC_ASSERT(LINE_BUFFER_SIZE >= sizeof("Done\r\n"), "LINE_BUFFER_SIZE is too small");
+OT_STATIC_ASSERT(LINE_BUFFER_SIZE >= sizeof("Error "), "LINE_BUFFER_SIZE is too small");
 
 static int sSessionFd = -1;
 
@@ -106,8 +110,9 @@ int main(int argc, char *argv[])
     int    ret;
     bool   isInteractive = true;
     bool   isFinished    = false;
-    char   lineBuffer[MAXLINE + 1];
+    char   lineBuffer[LINE_BUFFER_SIZE];
     size_t lineBufferWritePos = 0;
+    bool   isBeginOfLine      = true;
 
     sSessionFd = socket(AF_UNIX, SOCK_STREAM, 0);
     VerifyOrExit(sSessionFd != -1, perror("socket"); ret = OT_EXIT_FAILURE);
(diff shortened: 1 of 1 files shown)


src/posix/client.cpp
```diff
@@ -59,6 +59,12 @@
 
 #include "platform-posix.h"
 
+const size_t LINE_BUFFER_SIZE = 256;
+
+static_assert(LINE_BUFFER_SIZE >= sizeof("> "), "LINE_BUFFER_SIZE is too small");
+static_assert(LINE_BUFFER_SIZE >= sizeof("Done\r\n"), "LINE_BUFFER_SIZE is too small");
+static_assert(LINE_BUFFER_SIZE >= sizeof("Error "), "LINE_BUFFER_SIZE is too small");
```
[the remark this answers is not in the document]
simonlingoogle: In case we change it. static_assert make sure we are not assigning an invalid value to `LINE_BUFFER_SIZE`.

src/posix/client.cpp
```diff
@@ -296,44 +211,40 @@ int main(int argc, char *argv[])
...
-                    {
-                        VerifyOrExit(DoWrite(STDOUT_FILENO, ">", 1), ret = OT_EXIT_FAILURE);
-                    }
+                        size_t skipPrefix = 0;
 
-                    if (buffer[i] == '\r' || buffer[i] == '\n')
-                    {
-                        VerifyOrExit(DoWrite(STDOUT_FILENO, buffer + lineStart, static_cast<size_t>(i - lineStart + 1)),
+                        // read one line successfully or line buffer is full
+                        lineBuffer[lineBufferWritePos] = '\0';
```
Irving-cl, line 224: Is this line necessary? It seems that each time to call `DoWrite`, this position won't be output. And things like `strlen` is not used here.

simonlingoogle: Yes it is.
It's necessary for `strncmp`, for example in case the line in `lineBuffer` is shorter than `Done\r\n`

src/posix/client.cpp
```diff
@@ -59,6 +59,12 @@
 
 #include "platform-posix.h"
 
+const size_t LINE_BUFFER_SIZE = 256;
```
wgtdkp, line 62:
```suggestion
static const size_t kLineBufferSize = 256;
```

simonlingoogle: any difference when using `static` ?

wgtdkp: `static` limits the visibility of the name inside current `.cpp` file.

simonlingoogle: Make sense. Thanks.

simonlingoogle: revised to use enum

PR #5135, opened 2020-06-21
[efr32] remove extra parenthesis causing compilation problems

romacdon:
Remove extra parenthesis introduced with commit "[logging] remove NCP_SPINEL output #5104" causing compilation error for all efr32mg builds.


romacdon:
size-report failed but it looks like it has nothing to do with my change and more likely due to the environment at the time of the run.

```
./size-report: line 1: html: No such file or directory
./size-report: line 2: syntax error near unexpected token `<'
./size-report: line 2: `<head><title>504 Gateway Time-out</title></head>
'
##[error]Process completed with exit code 2.
```

jwhui:
> size-report failed but it looks like it has nothing to do with my change and more likely due to the environment at the time of the run.
> 
> ```
> ./size-report: line 1: html: No such file or directory
> ./size-report: line 2: syntax error near unexpected token `<'
> ./size-report: line 2: `<head><title>504 Gateway Time-out</title></head>
> '
> ##[error]Process completed with exit code 2.
> ```

@romacdon , yes, the [Glitch](http://glitch.me/) service we use has had stability issues recently. After triggering again, the size-report check was successful.

PR #5137, opened 2020-06-22
[meshcop] adding Joiner Discerner and new mechanism for commissioning

abtink:
This PR contains 3 related commits:

**[meshcop] adding `otJoinerDiscerner`**

The `otJoinerDiscerner` enables a new mechanism for Thread commissioning. The traditional Thread commissioning process uses factory assigned EUI-64 of the device to derive the Joiner ID and identify/filter a joiner (through steering data bloom filter). The Joiner Discerner (which is an unsigned value along with a user-specified bit length up to 64 bits) allows users to have more control and do not rely on factory-assigned EUI-64.

On joiner side, when a a Joiner Discerner value is provided, the Joiner code uses the discerner value to derive Joiner ID (appending a random prefix to extend the value to 64 bits) and bloom filter.

On commissioner side, users can add different joiners providing either an EUI-64 or an associated Joiner Discerner and the code accordingly match the Joiner IDs and compute steering data bloom filter.

**[joiner] adding `otJoinerPskd` and `JoinerPskd` types**

**[ncp] add support for Joiner Discerner and getting list of joiners**

This commit adds/updates spinel properties to add Joiner Discerner support on joiner or commissioner side. It also adds support for getting the list of joiners on commissioner by implementing property get handler for spinel property `MESHCOP_COMMISSIONER_JOINERS`.

--------

The code in this PR is done so posted it to start the review, but would like to add test-cases to cover the behavior of the new Joiner Discerner commissioning model (before merging). So next steps are 
- PR in `wpantund` to support the new spinel properties, then 
- PR to add `toranj` test-cases for Discerner based commissioning.

---
wpantund PR: https://github.com/openthread/wpantund/pull/472
PR to add `toranj` test-case https://github.com/openthread/openthread/pull/5154


src/core/meshcop/meshcop.cpp
```diff
@@ -44,11 +44,143 @@
...
+    uint8_t  reminaing = mLength;
+    uint64_t value     = mValue;
+
+    OT_ASSERT(IsValid());
+
+    // Write full bytes
+    while (reminaing >= CHAR_BIT)
+    {
+        *cur = static_cast<uint8_t>(value & 0xff);
+        value >>= 8;
```
wgtdkp, line 138:
```suggestion
        value >>= CHAR_BIT;
```

src/core/coap/coap_secure.cpp
```diff
@@ -115,6 +115,15 @@ otError CoapSecure::SetPsk(const uint8_t *aPsk, uint8_t aPskLength)
     return mDtls.SetPsk(aPsk, aPskLength);
 }
 
+void CoapSecure::SetPsk(const MeshCoP::JoinerPskd &aPskd)
+{
+    static_assert(static_cast<uint16_t>(MeshCoP::JoinerPskd::kMaxLength) <=
+                      static_cast<uint16_t>(MeshCoP::Dtls::kPskMaxLength),
+                  "The maximum length of DTLS PSK is smaller than joiner PSKd");
+
+    IgnoreError(mDtls.SetPsk(reinterpret_cast<const uint8_t *>(aPskd.GetAsCString()), aPskd.GetLength()));
```
wgtdkp, line 124: If we are sure that `mDtls.SetPsk` will success, I think assertion is better at here.

src/core/meshcop/commissioner.hpp
```diff
@@ -112,7 +112,27 @@ class Commissioner : public InstanceLocator
...
+     * @param[in]  aDiscerner  A Joiner Discerner.
+     * @param[in]  aPskd       A pointer to the PSKd.
+     * @param[in]  aTimeout    A time after which a Joiner is automatically removed, in seconds.
+     *
+     * @retval OT_ERROR_NONE           Successfully added the Joiner.
+     * @retval OT_ERROR_NO_BUFS        No buffers available to add the Joiner.
+     * @retval OT_ERROR_INVALID_STATE  Commissioner service is not started.
+     *
+     */
+    otError AddJoiner(const JoinerDiscerner &aDiscerner, const char *aPskd, uint32_t aTimeout)
```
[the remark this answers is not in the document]
abtink: This is existing design where we used a `nullptr` passed as `aEui64` to indicate "Any Joiner". I don't like it as well. Will add a new API to add "any joiner" and change this to use reference.

include/openthread/commissioner.h
```diff
@@ -112,19 +113,42 @@ typedef struct otCommissioningDataset
...
-    char         mPsk[OT_PSKD_MAX_SIZE + 1]; ///< Joiner pskd
-    uint32_t     mExpirationTime;            ///< Joiner expiration time in msec
-
-    bool mAny : 1; /// TRUE if eui64 isn't set, FALSE otherwise.
+    otJoinerInfoType mType; ///< Joiner type.
+    union
+    {
+        otExtAddress      mEui64;     ///< Joiner EUI64 (when `mType` is `OT_JOINER_INFO_TYPE_EUI64`)
+        otJoinerDiscerner mDiscerner; ///< Joiner Discerner (when `mType` is `OT_JOINER_INFO_TYPE_DISCERNER`)
+    } mShared;                        ///< Shared fields
```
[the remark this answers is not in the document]
abtink: We use `mShared` to indicate `union` fields.  I used "Joiner Discerner" and avoided the  "discriminator" term to ensure no confusion with CHIP project use of this term (also it has the side benefit of not using ["discriminate"](https://en.wikipedia.org/wiki/Discrimination) term  which is generally associated with not-positive stuff).. :)

src/core/meshcop/commissioner.cpp
```diff
@@ -1173,8 +1292,25 @@ const char *Commissioner::StateToString(otCommissionerState aState)
 
 void Commissioner::LogJoinerEntry(const char *aAction, const Joiner &aJoiner) const
 {
-    otLogInfoMeshCoP("%s Joiner (%s, %s)", aAction, aJoiner.mAny ? "*" : aJoiner.mEui64.ToString().AsCString(),
-                     aJoiner.mPsk);
+    switch (aJoiner.mType)
+    {
+    case Joiner::kTypeUnused:
```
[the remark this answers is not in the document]
abtink: We can, but since this method is only for logging/debugging I think it's ok to not assert to avoid code size increase and also avoid the messiness of assert in switch requiring `OT_UNREACHABLE_CODE(break)` (to make it work for some toolchains)

src/core/meshcop/commissioner.hpp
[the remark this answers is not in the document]
abtink: Added this in new pushed commits.

src/core/meshcop/commissioner.hpp
```diff
@@ -137,7 +157,26 @@ class Commissioner : public InstanceLocator
...
+     *
+     * @param[in]  aDiscerner     A Joiner Discerner.
+     * @param[in]  aDelay         The delay to remove Joiner (in seconds).
+     *
+     * @retval OT_ERROR_NONE           Successfully added the Joiner.
+     * @retval OT_ERROR_NOT_FOUND      The Joiner specified by @p aEui64 was not found.
+     * @retval OT_ERROR_INVALID_STATE  Commissioner service is not started.
+     *
+     */
+    otError RemoveJoiner(const JoinerDiscerner &aDiscerner, uint32_t aDelay)
```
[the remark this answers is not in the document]
abtink: Added this in new pushed commits.

src/core/meshcop/commissioner.cpp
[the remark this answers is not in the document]
wgtdkp: Makes sense to me.

include/openthread/commissioner.h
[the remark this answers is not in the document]
wgtdkp: `mEui64` and `mDiscerner` are both objects represent an identity, so I was thinking that the union should be a common / abstract name describes the meaning of the two fields. While discriminator may not be a good name,  what about `mSharedId` we used in `commissioner.h`?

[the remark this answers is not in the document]
wgtdkp: Another question is, do we need to distinguish `EUI64` and general `discerner`? I don't see how we will process differently for EUI64 and discerner, since they are just sources of joiner ID, which is used for steering data. It seems we can remove `mType` and `mEui64` and keep only `mDiscerner`. Thoughts?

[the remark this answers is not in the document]
abtink: I am ok renaming the `union` field to `mSharedId`. Will do in the next push.

I think we do need to distinguish between EUI64 and discerner (I think you may be thinking that we can treat EUI64 as a discerner with 64 bit length which would not be the case). 
- In terms of how they are used there are differences: 
   - For EUI64 we use a hash of it to add to bloom filter and also use same as Joiner Id.
   - For Joiner Discerner, there is no hashing for bloom filter (the actual discerner value is directly used for updating bloom filter). For link-local IID (Joiner Id) we currently do add a random prefix to extend it to 64 bit (if not already at 64) and no hashing. We may consider later to update this, e.g., use hash of a random prefix + discerner value to hide the actual discerner value when deriving the link-local IID (note that this would be only for IID derivation and cannnot be used in bloom filter).
- Also in terms of representation (in `JoinerInfo`),  the discerner is an unsigned int `uint64_t` value so its byte order is machine/platform dependent (little/big endian), compared to `EUI64` which is a MAC Ext Address (byte order is fixed).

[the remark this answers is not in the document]
wgtdkp: > For Joiner Discerner, there is no hashing for bloom filter

I missed this. Agree we need to keep both `EUI64` and `Discerner`.

include/openthread/commissioner.h
```diff
@@ -114,17 +115,31 @@ typedef struct otCommissioningDataset
 
 #define OT_PSKD_MAX_SIZE 32 ///< Size of a Joiner PSKd (bytes)
 
+/**
+ * This enumeration defines a Joiner Info Typer.
+ *
+ */
+typedef enum otJoinerInfoType
```
jwhui, line 131: With C++11, should we be explicit with the type?

include/openthread/commissioner.h
```diff
@@ -141,12 +156,14 @@ typedef void (*otCommissionerStateCallback)(otCommissionerState aState, void *aC
 /**
  * This function pointer is called whenever the joiner state changes.
  *
- * @param[in]  aEvent     The joiner event type.
- * @param[in]  aJoinerId  A pointer to the Joiner ID.
- * @param[in]  aContext   A pointer to application-specific context.
+ * @param[in]  aEvent       The joiner event type.
+ * @param[in]  aJoinerInfo  A pointer to the Joiner Info.
+ * @param[in]  aJoinerId    A pointer to the Joiner ID (if not known Id is set to all zeros).
```
jwhui, line 170: What about passing NULL if not known?

src/core/meshcop/meshcop.cpp
```diff
@@ -50,6 +50,88 @@ enum
...
+
+void JoinerDiscerner::CopyTo(Mac::ExtAddress &aExtAddress) const
+{
+    // Copies the discerner value up to its bit length to `aExtAddress`
+    // array, assuming big-endian encoding (i.e., the discerner lowest bits
+    // are copied at end of `aExtAddress.m8[]` array). Any initial/remaining
+    // bits of `aExtAddress` array remain unchanged.
+
+    uint8_t *cur       = &aExtAddress.m8[sizeof(Mac::ExtAddress) - 1];
+    uint8_t  reminaing = mLength;
```
jwhui, line 129:
```suggestion
    uint8_t  remaining = mLength;
```

src/core/meshcop/meshcop.hpp
```diff
@@ -58,6 +61,97 @@ enum
...
+     * @retval OT_ERROR_NONE           The PSKd was updated successfully.
+     * @retval OT_ERROR_INVALID_ARGS   The given PSKd C string is not valid.
+     *
+     */
+    otError SetFrom(const char *aPskdString);
+
+    /**
+     * This method gets the PSKd as a null terminated C string.
+     *
+     * This method must be used when PSKd is verified to be valid, otherwise its behavior is undefined.
```
jwhui, line 103:
```suggestion
     * This method must be used after the PSKd is validated, otherwise its behavior is undefined.
```

src/core/meshcop/meshcop.hpp
```diff
@@ -58,6 +61,97 @@ enum
...
+     *
+     * @returns The PSKd as a C string.
+     *
+     */
+    const char *GetAsCString(void) const { return m8; }
+
+    /**
+     * This method gets the PSKd string length.
+     *
+     * This method must be used when PSKd is verified to be valid, otherwise its behavior is undefined.
```
jwhui, line 113:
```suggestion
     * This method must be used after the PSKd is validated, otherwise its behavior is undefined.
```

src/lib/spinel/spinel.h
```diff
@@ -3143,6 +3150,32 @@ enum
      */
     SPINEL_PROP_MESHCOP_COMMISSIONER_SESSION_ID = SPINEL_PROP_MESHCOP__BEGIN + 5,
 
+    /// Thread Joiner Discerner
+    /** Format `CX`  - Read-write
+     *
+     * Required capability: SPINEL_CAP_THREAD_JOINER
+     *
+     * This property represent a Joiner Discerner.
```
jwhui, line 3158:
```suggestion
     * This property represents a Joiner Discerner.
```

src/lib/spinel/spinel.h
```diff
@@ -3112,15 +3112,22 @@ enum
     SPINEL_PROP_MESHCOP_COMMISSIONER_STATE = SPINEL_PROP_MESHCOP__BEGIN + 2,
 
     // Thread Commissioner Joiners
-    /** Format `A(t(E)UL)` - insert or remove only
+    /** Format `A(t(t(E|CX)UL))` - get, insert or remove.
      *
      * Required capability: SPINEL_CAP_THREAD_COMMISSIONER
      *
-     * Data per item is:
+     * Data per arrat entry is:
```
jwhui, line 3119:
```suggestion
     * Data per array entry is:
```

abtink:
> Looks good overall. Are we planning any continuous integration tests with the new Joiner Discerner feature?

Thanks @jwhui . Yes, I want to add CI `toranj` tests for it (with the `wpanatund` changes) and see it all passing, before we merge this. Will submit the CI test as a seperate PR (due to its dependency to `wpantund` changes).

include/openthread/commissioner.h
abtink: Good idea. I agree we should use the explicit type in all core `enum` definitions.

In the case of public OT APIs/definitions, we still need to support `C` (not `C++`) so I am not sure if we can use the explicit type for these `enum` definitions.

include/openthread/commissioner.h
abtink: Makes sense. Thanks. Will change it.

abtink:
Pushed the updated changes. It includes one additional commit at the end: 

**[commissioner] use c++11 ranged-base `for` and explicitly typed `enum`**

PR #5139, opened 2020-06-22
[posix] ensure backward compatible with older RCP image

brianljt:


abtink:
I think we need to define and document the policy we want to enforce related to RCP and (posix) host compatibility. 

During early RCP approach development, we talked about the compatibility question related to RCP and host fimrware. IIRC at the time we went with no explicit compatibility requirement (expecting the RCP and posix host to be updated together).

In spinel definition (related to host side driver/`wpantund`) we require new NCP firmware to remain compatible with older driver (i.e. older versions `wpantund` or other host drivers should continue to work with newer NCP OT builds).
```
 * ---------------------------------------------------------------------------
 *
 *   Spinel definition guideline:
 *
 *   New NCP firmware should work with an older host driver, i.e., NCP
 *   implementation should remain backward compatible.
 *
 *    - Existing fields in the format of an already implemented spinel
 *      property or command cannot change.
 *
 *    - New fields may be appended at the end of the format (or the end of
 *      a struct) as long as the NCP implementation treats the new fields as
 *      optional (i.e., a driver not aware of and therefore not using the
 *      new fields should continue to function as before).
 *
 * ---------------------------------------------------------------------------
```

With this PR, it seems like for RCP and host we want it such that the new version on the host side code to remain compatible with older RCP firmware (i.e. the opposite way of NCP/driver case).

bukepo:
> I think we need to define and document the policy we want to enforce related to RCP and (posix) host compatibility. 
> 
> 
> 
> During early RCP approach development, we talked about the compatibility question related to RCP and host fimrware. IIRC at the time we went with no explicit compatibility requirement (expecting the RCP and posix host to be updated together).
> 
> 
> 
> In spinel definition (related to host side driver/`wpantund`) we require new NCP firmware to remain compatible with older driver (i.e. older versions `wpantund` or other host drivers should continue to work with newer NCP OT builds).
> 
> ```
> 
>  * ---------------------------------------------------------------------------
> 
>  *
> 
>  *   Spinel definition guideline:
> 
>  *
> 
>  *   New NCP firmware should work with an older host driver, i.e., NCP
> 
>  *   implementation should remain backward compatible.
> 
>  *
> 
>  *    - Existing fields in the format of an already implemented spinel
> 
>  *      property or command cannot change.
> 
>  *
> 
>  *    - New fields may be appended at the end of the format (or the end of
> 
>  *      a struct) as long as the NCP implementation treats the new fields as
> 
>  *      optional (i.e., a driver not aware of and therefore not using the
> 
>  *      new fields should continue to function as before).
> 
>  *
> 
>  * ---------------------------------------------------------------------------
> 
> ```
> 
> 
> 
> With this PR, it seems like for RCP and host we want it such that the new version on the host side code to remain compatible with older RCP firmware (i.e. the opposite way of NCP/driver case).
> 
> 
> 
> 
> 
> 

We received several incompatibility issues from internal and external users. I agree it is not an requirement for this kind of compatibility, but since we didn't increase the version, I think it better to keep compatibility in two directions if it is not difficult. We can remove the the code one day we increase the version of spinel. Thoughts?

abtink:
> We received several incompatibility issues from internal and external users. I agree it is not an requirement for this kind of compatibility, but since we didn't increase the version, I think it better to keep compatibility in two directions if it is not difficult. We can remove the the code one day we increase the version of spinel. Thoughts?

It seems reasonable to me if we have different directions for RCP/host vs NCP/host. It makes sense based on how they are being used. .

simonlingoogle:
I have had assertion failures on host when using an older RCP firmware. 

Unlike NCP firmwares, developers would expect the RCP firmware to update much less often. So, they might expect an very old RCP firmware to work just fine on latest Host. 

Thus, IMO, we should keep the Host software backward compatible with older RCP firmware whenever possible.

brianljt:
> In spinel definition (related to host side driver/wpantund) we require new NCP firmware to remain compatible with older driver

I think we can still keep the principle that older driver **should** work with newer radio firmware, but **not necessary** that newer driver works with older radio firmware. Because it is not always doable to keep newer driver work with older radio firmware. For example, if radio appends more meta info into TxDone or RxDone event but these new fields are not indicated by any `RADIO_CAP_XXX` or Spinel version change, host does not know which structure the meta info is using and does not know how to parse it.

For this change, it is easy to do so. So we can implement it as a better to have feature to maintain compatibility. But I think it does not change the principle on backward compatibility regarding host driver and radio firmware.

PR #5143, opened 2020-06-24
[posix] wait for writable before next try

bukepo:
This commit waits fd to be writable before try writing next time.

### DIFF
--- a/src/posix/platform/hdlc_interface.cpp
+++ b/src/posix/platform/hdlc_interface.cpp
@@ -235,19 +235,16 @@ otError HdlcInterface::Write(const uint8_t *aFrame, uint16_t aLength)
     {
         ssize_t rval = write(mSockFd, aFrame, aLength);
 
-        if (rval > 0)
+        if (rval >= 0)
         {
             aLength -= static_cast<uint16_t>(rval);
             aFrame += static_cast<uint16_t>(rval);
-            continue;
+            SuccessOrExit(error = WaitForWritable());
         }
-
-        if ((rval < 0) && (errno != EAGAIN) && (errno != EWOULDBLOCK) && (errno != EINTR))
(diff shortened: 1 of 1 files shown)


src/posix/platform/hdlc_interface.cpp
```diff
@@ -235,19 +235,16 @@ otError HdlcInterface::Write(const uint8_t *aFrame, uint16_t aLength)
     {
         ssize_t rval = write(mSockFd, aFrame, aLength);
 
-        if (rval > 0)
+        if (rval >= 0)
         {
             aLength -= static_cast<uint16_t>(rval);
             aFrame += static_cast<uint16_t>(rval);
-            continue;
+            SuccessOrExit(error = WaitForWritable());
```
abtink, line 242: Looks good as is.
One suggestion: Can we add a check to see if `aLength` is now zero after the `-=` and `break` from while loop to avoid an extra call to `WaitForWritable()` when all bytes are sent.

bukepo: updated.

src/posix/platform/hdlc_interface.cpp
```diff
@@ -235,16 +235,18 @@ otError HdlcInterface::Write(const uint8_t *aFrame, uint16_t aLength)
...
             aLength -= static_cast<uint16_t>(rval);
             aFrame += static_cast<uint16_t>(rval);
-            continue;
         }
-
-        if ((rval < 0) && (errno != EAGAIN) && (errno != EWOULDBLOCK) && (errno != EINTR))
+        else if (rval < 0)
         {
-            DieNow(OT_EXIT_ERROR_ERRNO);
+            VerifyOrDie((errno == EAGAIN) || (errno == EWOULDBLOCK) || (errno == EINTR), OT_EXIT_ERROR_ERRNO);
```
abtink, line 249: Nice. Thanks. Much simpler and more readable.

PR #5144, opened 2020-06-24
[cmake] silence CMP0048 warning in mbedtls

FiveDimensions:
CMake Warning (dev) at third_party/mbedtls/repo/CMakeLists.txt:5 (project):
  Policy CMP0048 is not set: project() command manages VERSION variables.
  Run "cmake --help-policy CMP0048" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.


bukepo:
I also noticed the warnings, but since this is from third_party code, are we going to modify this file?

jwhui:
> I also noticed the warnings, but since this is from third_party code, are we going to modify this file?

I generally agree that we should not modify code we pull from third party unless necessary.

FiveDimensions:
Considering repo is from third party and leave the code as it is, close the pr.

PR #5151, opened 2020-06-25
[mac-filter] misc enhancements

abtink:
This commit contains some smaller enhancements in `Mac::Filter`:

- Use C++11 range-based `for` loop when iterating over filter entries
- Define `Filter:Mode` enum (mirror of `otMacFilterAddressMode`)
- Change `Filter::SetMode()` and `otLinkFilterSetAddressMode()` to
  return `void` (update its use from CLI/NCP code).
- Simplify `Filter::Apply()` implementation.
 - Add new separate public OT and internal APIs to set/clear default
    RSS-In (e.g., `SetDefaultRssIn()`/`ClearDefaultRssIn()`).
- Return no error (instead of `ALREADY`) from `Filter::AddAddress()`
   and its corresponding public `otLinkFilter` API when entry is already
   present in the list.
- Do not return `NOT_FOUND` error from `Filter::Remove{Address/RssIn}()`.
- Update public `otLinkFilter{API}` documentation.
- Update NCP and CLI code to adopt new `otLinkFilter` APIs.

### DIFF
--- a/include/openthread/link.h
+++ b/include/openthread/link.h
@@ -722,19 +722,11 @@ void otLinkSetMaxFrameRetriesIndirect(otInstance *aInstance, uint8_t aMaxFrameRe
 /**
  * This function gets the address mode of MAC filter.
  *
+ * This function is available when OPENTHREAD_CONFIG_MAC_FILTER_ENABLE configuration is enabled.
+ *
  * @param[in]  aInstance  A pointer to an OpenThread instance.
  *
  * @returns  the address mode.
- *
- * @sa otLinkFilterSetAddressMode
- * @sa otLinkFilterAddAddress
- * @sa otLinkFilterRemoveAddress
- * @sa otLinkFilterClearAddresses
- * @sa otLinkFilterGetNextAddress
- * @sa otLinkFilterAddRssIn
- * @sa otLinkFilterRemoveRssIn
- * @sa otLinkFilterClearRssIn
- * @sa otLinkFilterGetNextRssIn
  *
  */
 otMacFilterAddressMode otLinkFilterGetAddressMode(otInstance *aInstance);
@@ -742,45 +734,25 @@ otMacFilterAddressMode otLinkFilterGetAddressMode(otInstance *aInstance);
 /**
  * This function sets the address mode of MAC filter.
  *
+ * This function is available when OPENTHREAD_CONFIG_MAC_FILTER_ENABLE configuration is enabled.
+ *
  * @param[in]  aInstance  A pointer to an OpenThread instance.
  * @param[in]  aMode      The address mode to set.
  *
- * @retval OT_ERROR_NONE           Successfully set the address mode.
- * @retval OT_ERROR_INVALID_ARGS   @p aMode is not valid.
- *
- * @sa otLinkFilterGetAddressMode
- * @sa otLinkFilterAddAddress
- * @sa otLinkFilterRemoveAddress
- * @sa otLinkFilterClearAddresses
- * @sa otLinkFilterGetNextAddress
- * @sa otLinkFilterAddRssIn
- * @sa otLinkFilterRemoveRssIn
- * @sa otLinkFilterClearRssIn
- * @sa otLinkFilterGetNextRssIn
- *
  */
-otError otLinkFilterSetAddressMode(otInstance *aInstance, otMacFilterAddressMode aMode);
+void otLinkFilterSetAddressMode(otInstance *aInstance, otMacFilterAddressMode aMode);
 
 /**
  * This method adds an Extended Address to MAC filter.
  *
+ * This function is available when OPENTHREAD_CONFIG_MAC_FILTER_ENABLE configuration is enabled.
+ *
  * @param[in]  aInstance    A pointer to an OpenThread instance.
- * @param[in]  aExtAddress  A reference to the Extended Address.
+  * @param[in]  aExtAddress  A pointer to the Extended Address (MUST NOT be NULL).
  *
  * @retval OT_ERROR_NONE           Successfully added @p aExtAddress to MAC filter.
  * @retval OT_ERROR_ALREADY        If @p aExtAddress was already in MAC filter.
- * @retval OT_ERROR_INVALID_ARGS   If @p aExtAddress is NULL.
  * @retval OT_ERROR_NO_BUFS        No available entry exists.
- *
- * @sa otLinkFilterGetAddressMode
- * @sa otLinkFilterSetAddressMode
- * @sa otLinkFilterRemoveAddress
- * @sa otLinkFilterClearAddresses
- * @sa otLinkFilterGetNextAddress
- * @sa otLinkFilterAddRssIn
- * @sa otLinkFilterRemoveRssIn
- * @sa otLinkFilterClearRssIn
- * @sa otLinkFilterGetNextRssIn
  *
  */
 otError otLinkFilterAddAddress(otInstance *aInstance, const otExtAddress *aExtAddress);
@@ -788,22 +760,13 @@ otError otLinkFilterAddAddress(otInstance *aInstance, const otExtAddress *aExtAd
 /**
  * This method removes an Extended Address from MAC filter.
  *
+ * This function is available when OPENTHREAD_CONFIG_MAC_FILTER_ENABLE configuration is enabled.
+ *
  * @param[in]  aInstance    A pointer to an OpenThread instance.
- * @param[in]  aExtAddress  A reference to the Extended Address.
+ * @param[in]  aExtAddress  A pointer to the Extended Address (MUST NOT be NULL).
  *
  * @retval OT_ERROR_NONE           Successfully removed @p aExtAddress from MAC filter.
- * @retval OT_ERROR_INVALID_ARGS   If @p aExtAddress is NULL.
  * @retval OT_ERROR_NOT_FOUND      @p aExtAddress is not in MAC filter.
- *
- * @sa otLinkFilterGetAddressMode
- * @sa otLinkFilterSetAddressMode
- * @sa otLinkFilterAddAddress
- * @sa otLinkFilterClearAddresses
- * @sa otLinkFilterGetNextAddress
- * @sa otLinkFilterAddRssIn
- * @sa otLinkFilterRemoveRssIn
- * @sa otLinkFilterClearRssIn
- * @sa otLinkFilterGetNextRssIn
  *
  */
 otError otLinkFilterRemoveAddress(otInstance *aInstance, const otExtAddress *aExtAddress);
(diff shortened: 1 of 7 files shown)


include/openthread/link.h
```diff
@@ -722,227 +722,159 @@ void otLinkSetMaxFrameRetriesIndirect(otInstance *aInstance, uint8_t aMaxFrameRe
...
+void otLinkFilterSetAddressMode(otInstance *aInstance, otMacFilterAddressMode aMode);
 
 /**
  * This method adds an Extended Address to MAC filter.
  *
+ * This function is available when OPENTHREAD_CONFIG_MAC_FILTER_ENABLE configuration is enabled.
+ *
  * @param[in]  aInstance    A pointer to an OpenThread instance.
- * @param[in]  aExtAddress  A reference to the Extended Address.
+ * @param[in]  aExtAddress  A pointer to the Extended Address (MUST not be NULL).
```
jwhui, line 751:
```suggestion
  * @param[in]  aExtAddress  A pointer to the Extended Address (MUST NOT be NULL).
```

include/openthread/link.h
```diff
@@ -722,227 +722,159 @@ void otLinkSetMaxFrameRetriesIndirect(otInstance *aInstance, uint8_t aMaxFrameRe
...
 otError otLinkFilterAddAddress(otInstance *aInstance, const otExtAddress *aExtAddress);
 
 /**
  * This method removes an Extended Address from MAC filter.
  *
+ * This function is available when OPENTHREAD_CONFIG_MAC_FILTER_ENABLE configuration is enabled.
+ *
  * @param[in]  aInstance    A pointer to an OpenThread instance.
- * @param[in]  aExtAddress  A reference to the Extended Address.
+ * @param[in]  aExtAddress  A pointer to the Extended Address (MUST not be NULL).
```
jwhui, line 766:
```suggestion
 * @param[in]  aExtAddress  A pointer to the Extended Address (MUST NOT be NULL).
```

include/openthread/link.h
```diff
@@ -722,227 +722,159 @@ void otLinkSetMaxFrameRetriesIndirect(otInstance *aInstance, uint8_t aMaxFrameRe
...
  * This method gets an in-use address filter entry.
  *
+ * This function is available when OPENTHREAD_CONFIG_MAC_FILTER_ENABLE configuration is enabled.
+ *
  * @param[in]     aInstance  A pointer to an OpenThread instance.
  * @param[inout]  aIterator  A pointer to the MAC filter iterator context. To get the first in-use address filter entry,
- *                           it should be set to OT_MAC_FILTER_ITERATOR_INIT.
- * @param[out]    aEntry     A pointer to where the information is placed.
+ *                           it should be set to OT_MAC_FILTER_ITERATOR_INIT. MUST not be NULL.
+ * @param[out]    aEntry     A pointer to where the information is placed. MUST not be NULL.
```
jwhui, line 792:
```suggestion
 *                           it should be set to OT_MAC_FILTER_ITERATOR_INIT. MUST NOT be NULL.
 * @param[out]    aEntry     A pointer to where the information is placed. MUST NOT be NULL.
```

include/openthread/link.h
```diff
@@ -722,227 +722,159 @@ void otLinkSetMaxFrameRetriesIndirect(otInstance *aInstance, uint8_t aMaxFrameRe
...
+ * This method adds a fixed received signal strength (in dBm) entry for the messages from a given Extended Address in
+ * MAC Filter.
+ *
+ * This function is available when OPENTHREAD_CONFIG_MAC_FILTER_ENABLE configuration is enabled.
  *
  * @param[in]  aInstance    A pointer to an OpenThread instance.
- * @param[in]  aExtAddress  A pointer to the IEEE 802.15.4 Extended Address, or NULL to set the default received signal
- *                          strength.
- * @param[in]  aRss         The received signal strength (in dBm) to set.
+ * @param[in]  aExtAddress  A pointer to the IEEE 802.15.4 Extended Address. MUST not be NULL.
```
jwhui, line 807:
```suggestion
 * @param[in]  aExtAddress  A pointer to the IEEE 802.15.4 Extended Address. MUST NOT be NULL.
```

include/openthread/link.h
```diff
@@ -722,227 +722,159 @@ void otLinkSetMaxFrameRetriesIndirect(otInstance *aInstance, uint8_t aMaxFrameRe
...
+ * @param[in]  aRss         A received signal strength (in dBm)..
```
jwhui, line 808:
```suggestion
 * @param[in]  aRss         A received signal strength (in dBm).
```

include/openthread/link.h
```diff
@@ -722,227 +722,159 @@ void otLinkSetMaxFrameRetriesIndirect(otInstance *aInstance, uint8_t aMaxFrameRe
...
- * @sa otLinkFilterClearRssIn
- * @sa otLinkFilterGetNextRssIn
- *
  */
 otError otLinkFilterAddRssIn(otInstance *aInstance, const otExtAddress *aExtAddress, int8_t aRss);
 
 /**
- * This method removes the received signal strength setting for the received messages from the Extended Address or
- * removes the default received signal strength setting if no Extended Address is specified.
+ * This method removes an MAC Filter entry for fixed received signal strength setting for a given Extended Address.
```
jwhui, line 817:
```suggestion
 * This method removes a MAC Filter entry for fixed received signal strength setting for a given Extended Address.
```

include/openthread/link.h
```diff
@@ -722,227 +722,159 @@ void otLinkSetMaxFrameRetriesIndirect(otInstance *aInstance, uint8_t aMaxFrameRe
...
- * @sa otLinkFilterGetAddressMode
- * @sa otLinkFilterSetAddressMode
- * @sa otLinkFilterAddAddress
- * @sa otLinkFilterRemoveAddress
- * @sa otLinkFilterClearAddresses
- * @sa otLinkFilterGetNextAddress
- * @sa otLinkFilterAddRssIn
- * @sa otLinkFilterClearRssIn
- * @sa otLinkFilterGetNextRssIn
+ * @param[in]  aExtAddress  A pointer to the IEEE 802.15.4 Extended Address. MUST not be NULL.
```
jwhui, line 824:
```suggestion
 * @param[in]  aExtAddress  A pointer to the IEEE 802.15.4 Extended Address. MUST NOT be NULL.
```

include/openthread/link.h
```diff
@@ -722,227 +722,159 @@ void otLinkSetMaxFrameRetriesIndirect(otInstance *aInstance, uint8_t aMaxFrameRe
...
 /**
  * This method gets an in-use RssIn filter entry.
  *
+ * This function is available when OPENTHREAD_CONFIG_MAC_FILTER_ENABLE configuration is enabled.
+ *
  * @param[in]     aInstance  A pointer to an OpenThread instance.
- * @param[inout]  aIterator  A reference to the MAC filter iterator context. To get the first in-use RssIn Filter entry,
- *                           it should be set to OT_MAC_FILTER_ITERATOR_INIT.
- * @param[out]    aEntry     A reference to where the information is placed. The last entry would have the extended
+ * @param[inout]  aIterator  A pointer to the MAC filter iterator context. MUST not be NULL.
```
jwhui, line 869:
```suggestion
 * @param[inout]  aIterator  A pointer to the MAC filter iterator context. MUST NOT be NULL.
```

include/openthread/link.h
```diff
@@ -722,227 +722,159 @@ void otLinkSetMaxFrameRetriesIndirect(otInstance *aInstance, uint8_t aMaxFrameRe
...
+ *                           To get the first entry, it should be set to OT_MAC_FILTER_ITERATOR_INIT.
+ * @param[out]    aEntry     A pointer to where the information is placed. The last entry would have the extended
  *                           address as all 0xff to indicate the default received signal strength if it was set.
+                             @p aEntry MUST not be NULL.
```
jwhui, line 873:
```suggestion
                             @p aEntry MUST NOT be NULL.
```

src/core/mac/mac_filter.hpp
```diff
@@ -73,10 +73,20 @@ class Filter
      */
     typedef otMacFilterIterator Iterator;
 
+    /**
+     * This enumeration type represents the MAC Filter mode.
+     *
+     */
+    enum Mode
```
jwhui, line 80: Add explicit type?

src/core/mac/mac_filter.hpp
```diff
@@ -73,10 +73,20 @@ class Filter
...
+        kModeRssInOnly = OT_MAC_FILTER_ADDRESS_MODE_DISABLED,  ///< No address filtering. RSS-In update only.
+        kModeWhitelist = OT_MAC_FILTER_ADDRESS_MODE_WHITELIST, ///< Enable whitelist address filter mode.
+        kModeBlacklist = OT_MAC_FILTER_ADDRESS_MODE_BLACKLIST, ///< Enable blacklist address filter mode.
+    };
+
     enum
     {
-        kMaxEntries       = OPENTHREAD_CONFIG_MAC_FILTER_SIZE, ///< The maximum number of filter entries.
-        kFixedRssDisabled = OT_MAC_FILTER_FIXED_RSS_DISABLED,  ///< Value to indicate no fixed RSS is set.
+        kFixedRssDisabled = OT_MAC_FILTER_FIXED_RSS_DISABLED, ///< Value to indicate no fixed RSS is set.
```
jwhui, line 89: Add explicit type?

src/core/mac/mac_filter.cpp
```diff
@@ -41,61 +41,46 @@ namespace ot {
...
+    FilterEntry *rval = nullptr;
 
-    for (entry = &mFilterEntries[0]; entry < OT_ARRAY_END(mFilterEntries); entry++)
+    for (FilterEntry &entry : mFilterEntries)
     {
-        VerifyOrExit(entry->IsInUse(), OT_NOOP);
+        if (!entry.IsInUse())
+        {
+            ExitNow(rval = &entry);
+        }
```
simonlingoogle, line 79:
```suggestion
        VerifyOrExit(entry->IsInUse(), rval = &entry);
```
can we still use VerifyOrExit?

include/openthread/link.h
```diff
@@ -722,227 +722,159 @@ void otLinkSetMaxFrameRetriesIndirect(otInstance *aInstance, uint8_t aMaxFrameRe
...
  * This method adds an Extended Address to MAC filter.
  *
+ * This function is available when OPENTHREAD_CONFIG_MAC_FILTER_ENABLE configuration is enabled.
+ *
  * @param[in]  aInstance    A pointer to an OpenThread instance.
- * @param[in]  aExtAddress  A reference to the Extended Address.
+ * @param[in]  aExtAddress  A pointer to the Extended Address (MUST not be NULL).
  *
  * @retval OT_ERROR_NONE           Successfully added @p aExtAddress to MAC filter.
  * @retval OT_ERROR_ALREADY        If @p aExtAddress was already in MAC filter.
```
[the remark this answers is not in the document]
abtink: Good idea. Will change this later.

src/core/mac/mac_filter.cpp
abtink: From what I recall, we had a discussion about this (in earlier days of OT development).

For cases where we are searching within a list/array/container for an entry (while we can use `VerifyOrExit` and it would be fewer lines) the preferred approach is `"if (entry satisfies certain conditions) then exit"` which reads much more clearly (and therefore less chance of a mistake), compared to `"verify that entry does not satisfy conditions or exit"`.

`VerifyOrExit` suits well for checking and verifying that some conditions are true before continuing on with next lines of code and logic.

In general, I personally like the syntax sugars and using fewer lines wherever possible, but in this case, I also do favor readability of the style "if (check) then exit" approach.

simonlingoogle: I agree with you that `if (condition) exit` is more readable in some loops.

abtink:
Kept the new changes in a new commit to help with review (but would suggest to squash all before merge - the first commit has the updated/full commit message). 

The change in new commit is mainly removing `OT_ERROR_ALREADY`/`OT_ERROR_NOT_FOUND` from `Add`/`Remove{Entry}` methods in `Mac::Filter` and related public `otLinkFilter` APIs.

PR #5155, opened 2020-06-27
[otns] add default implementation for otPlatOtnsStatus

ctan-g:
This PR adds a default implementation for `otPlatOtnsStatus`, which emits OTNS status messages to log, so that `OTNS=1` could be used with real devices. This is required for the Silk-OTNS integration work, as Silk will pick up these messages and reroute them to OTNS for visualization purposes.

### DIFF
--- a/src/core/common/logging.hpp
+++ b/src/core/common/logging.hpp
@@ -1159,6 +1159,19 @@ extern "C" {
 #endif
 
 /**
+ * @def otLogOtns
+ *
+ * This method generates a log with level none for the Core region,
+ * and is specifically for OTNS visualization use.
+ *
+ * @param[in]  ...       Arguments for the format specification.
+ *
+ */
+#if (OPENTHREAD_MTD || OPENTHREAD_FTD) && OPENTHREAD_CONFIG_OTNS_ENABLE
+#define otLogOtns(...) _otLogFormatter(OT_LOG_LEVEL_NONE, OT_LOG_REGION_CORE, _OT_LEVEL_NONE_PREFIX __VA_ARGS__, NULL)
+#endif
+
+/**
  * @def otDumpCrit
  *
  * This method generates a memory dump with log level critical.
--- a/src/core/utils/otns.cpp
+++ b/src/core/utils/otns.cpp
@@ -136,7 +136,7 @@ void Otns::EmitNeighborChange(otNeighborTableEvent aEvent, Neighbor &aNeighbor)
 
 OT_TOOL_WEAK void otPlatOtnsStatus(const char *aStatus)
 {
-    _otLogFormatter(OT_LOG_LEVEL_NONE, OT_LOG_REGION_CORE, "[OTNS] %s", aStatus, NULL);
+    otLogOtns("[OTNS] %s", aStatus);
 }
 
 #endif // (OPENTHREAD_MTD || OPENTHREAD_FTD) && OPENTHREAD_CONFIG_OTNS_ENABLE
--- a/src/core/utils/otns.cpp
+++ b/src/core/utils/otns.cpp
@@ -38,6 +38,7 @@
 
 #include "common/debug.hpp"
 #include "common/locator-getters.hpp"
+#include "common/logging.hpp"
 
 namespace ot {
 namespace Utils {
@@ -132,5 +133,10 @@ void Otns::EmitNeighborChange(otNeighborTableEvent aEvent, Neighbor &aNeighbor)
 
 } // namespace Utils
 } // namespace ot
+
+OT_TOOL_WEAK void otPlatOtnsStatus(const char *aStatus)
+{
+    _otLogFormatter(OT_LOG_LEVEL_NONE, OT_LOG_REGION_CORE, "[OTNS] %s", aStatus, NULL);
+}
 
 #endif // (OPENTHREAD_MTD || OPENTHREAD_FTD) && OPENTHREAD_CONFIG_OTNS_ENABLE


src/core/utils/otns.cpp
```diff
@@ -133,4 +134,10 @@ void Otns::EmitNeighborChange(otNeighborTableEvent aEvent, Neighbor &aNeighbor)
 } // namespace Utils
 } // namespace ot
 
+OT_TOOL_WEAK void otPlatOtnsStatus(const char *aStatus)
+{
+    OT_UNUSED_VARIABLE(aStatus);
+    otLogInfoPlat("%s\r\n", aStatus);
```
simonlingoogle, line 140:
```suggestion
    _otLogFormatter(OT_LOG_LEVEL_NONE, OT_LOG_REGION_CORE, "[OTNS] %s", aStatus, NULL);
```
- use log level `NONE` to make sure the log is printed despite the logging level? 
- Don't append `\r\n` to the end of format string. (logs are not supposed to contain the EOLs)
- Use some specific tags to make it easier for log parsers to identify OTNS status strings

ctan-g: This will output the logs to `wpantund` at `[EMRG]` level. Would this be okay?

ctan-g:
Running `OT_OPTIONS=-DOT_READLINE=OFF VIRTUAL_TIME=0 NODE_MODE=rcp ./script/test clean build expect` seems to pass locally

src/core/utils/otns.cpp
```diff
@@ -133,4 +134,9 @@ void Otns::EmitNeighborChange(otNeighborTableEvent aEvent, Neighbor &aNeighbor)
 } // namespace Utils
 } // namespace ot
 
+OT_TOOL_WEAK void otPlatOtnsStatus(const char *aStatus)
+{
+    _otLogFormatter(OT_LOG_LEVEL_NONE, OT_LOG_REGION_CORE, "[OTNS] %s", aStatus, NULL);
```
abtink, line 139: I would advice against using `_otLogFormatter()` from another file.

All macors starting with prefix `_` should be considered private to the header file itself and not used by anyone else.

abtink: If we need specific logging behavior (e.g. logging with `[OTNS]` or at certain level) for `otns`, I suggest adding specific new macros for that in `logging.hpp` directly.

ctan-g: Ok, that makes sense.

src/core/common/logging.hpp
```diff
@@ -110,6 +110,21 @@ extern "C" {
...
+ *
+ * @param[in]  aRegion   The log region.
+ * @param[in]  ...       Arguments for the format specification.
+ *
+ */
+#if OPENTHREAD_CONFIG_LOG_LEVEL >= OT_LOG_LEVEL_NONE
+#define otLogNone(aRegion, ...) _otLogFormatter(OT_LOG_LEVEL_NONE, aRegion, _OT_LEVEL_NONE_PREFIX __VA_ARGS__, NULL)
+#else
+#define otLogNone(aRegion, ...)
+#endif
```
abtink, line 126: Log level "none" indicates that user does not want any logging (i.e., disable all logging). So having an `otLogNone()` and `otLogNoneCore()` macros to emit logs at "none" level does seem good to me.

Also note that by using `otLog{Level}Core` you are tied to `OPENTHREAD_CONFIG_LOG_CORE` config and if that is disabled, then OTNS logs wont be emitted as well.

I guess you want to emit logs for OTNS independent of log level and all other configs (this is similar to cert log I'd say).
So I would suggest adding  a specific macro for OTNS, something like this:
```
#if (OPENTHREAD_MTD || OPENTHREAD_FTD) && OPENTHREAD_CONFIG_OTNS_ENABLE
#define otLogOtns(...)   _otLogFormatter(OT_LOG_LEVEL_NONE, OT_LOG_REGION_CORE, _OT_LEVEL_NONE_PREFIX __VA_ARGS__, NULL)
#endif
```
(would be good to add it conditionally when OTNS config is enabled)

ctan-g: Hmm, I was under the impression that "none" is the highest level of logging. In `wpantund` it is interpreted as "emergency" logging level. Your solution does make more sense however.

abtink: Edit: Noticed I had a typo -> "to emit logs at "none" level does seem good to me" -> "to emit logs at "none" level does **_not_** seem good to me". 

"none" is not really a log level to be used within OpenThread itself. It indicates that user wants no logging (disable all logging).

ctan-g: Maybe I can settle with critical level then?

ctan-g: Discovered that I need to avoid using debug level logging, since the verbose logging exceeds the NCP buffer size and will cause OTNS logs to be lost. However there's no easy CLI build flags to set logging levels yet. Falling back to use none level as previously suggested.

abtink: - There is a CLI command `"log level"` to get or change log level dynamically. 
- Note that you need `OPENTHREAD_CONFG_LOG_LEVEL_DYNMIAC_ENABLE` enabled to allow dynamic log level change (during run time) 
- Otherwise the log level is determined through the config `OPENTHREAD_CONFIG_LOG_LEVEL`. 
- You can check `config/logging.h` for all config related to logging behavior. 
- Some of these configs are also exposed through cmake options (`OT_LOG_LEVEL_DYNMIAC`, `OT_FULL_LOGS`) and autoconf switches.

ctan-g: Do you suggest I use critical level for this log? The only purpose for this is to enable real devices to emit status logs to wpantund where it will be redirected to OTNS by Silk to allow for network visualization when running test cases. Ideally it should be turned on whenever I set `OTNS=1`. A change of `OPENTHREAD_CONFIG_LOG_LEVEL` isn't supported as one of the make command flags, right?
Also `OT_FULL_LOGS` does flood the NCP buffer and cause loss of logs as I mention above.

ctan-g: Although I think the proper way is to have a new log region for OTNS, and thus enable an easy control of OTNS logs. But I feel like that's a part of the specification and adding a region will be a much more arbitrated process.

abtink: As I mentioned before, to me this looks like cert logging (`CONFIG_REFERNCE_DEVICE_ENABLE`). 

So similar approach to that makes sense, which is close to what the latest code in this PR is doing, i.e.,
- adding specific logging macro for this use-case (and we can do whatever level/regsion/etc in macro definition)
- and that the macro is only provided when OTNS is enabled (and not tied to any other configs).

jwhui: I think having a log region specific to OTNS makes the most sense. Can you clarify the concern about adding a log region for OTNS? This is an OpenThread-specific concept and nothing to do with the Thread Specification.

jwhui: As @abtink , I'm fine with the current approach as well (given that it is similar to the existing reference device logs).

ctan-g: I am not concerned per se, just wasn't sure that I felt comfortable making the decision of adding a new region unless other maintainers agree with it, since I am very new to the codebase. Now that this is merged, a change could be made in the future to add the region if we do have the need to have finer grained control of OTNS logs, or if we are using OTNS logs in more places. For now I think this is fine, as OTNS logs won't be used beyond reporting these status messages and radio events, as far as I can tell.

PR #5157, opened 2020-06-28
[github-actions] fix expects fail by chance

simonlingoogle:
This PR tries to reduce expects fails in GitHub actions:
- `wait_for` only sends command when it's not empty
- increases the default `expect` timeout from 1s to 3s.


simonlingoogle:
> This change looks good.
> But `> Done` shouldn't have caused any failures since a partial match would be accepted. Is there a failed case because of this?

I think you are right. I think the changes are still legitimate.

jwhui:
@simonlingoogle , can you help fix conflicts?

PR #5165, opened 2020-06-30
[continuous-integration] add SPI build check

Moandor-y:
Enables NCP SPI in build checks. Also fixes some build errors in SPI codes.


src/core/common/debug.hpp
```diff
@@ -57,7 +57,7 @@
  */
 #ifndef FILE_NAME
 #define FILE_NAME __FILE__
-#endif
+#endif // FILE_NAME
```
abtink, line 60: Per our style guide, we add the comment on `#endif` if the number of lines between `#if` and `#endif` is more than 20 lines. Would be better to not add it in cases like this (does not really help with code readability)
See this in style guide: 
https://github.com/openthread/openthread/blob/d65f8a3812611a89f851a18df2d4b8219a6fbca2/STYLE_GUIDE.md#conventions-and-best-practices

src/core/common/debug.hpp
```diff
@@ -86,9 +86,11 @@
 
 #endif // OPENTHREAD_CONFIG_PLATFORM_ASSERT_MANAGEMENT
 
-#else
+#else // OPENTHREAD_CONFIG_ASSERT_ENABLE
 
-#define OT_ASSERT(cond)
+#include "openthread/platform/toolchain.h"
+
+#define OT_ASSERT(cond) OT_UNUSED_VARIABLE(cond)
```
abtink, line 93: This won't work for all toolchains. Note that `OT_UNUSED_VARIABLE()` has different implemenations per toolchain limitation.
For example:
https://github.com/openthread/openthread/blob/d65f8a3812611a89f851a18df2d4b8219a6fbca2/include/openthread/platform/toolchain.h#L208-L214

The `cond` is a not necessarily a variable but a condition (statement) and the above logic won't work.

--------

Also this change seems to be unrelated to this PR and adding  SPI build check to CI. I would suggest to submit the assert changes as a separate PR. Thanks.

Moandor-y: This change is here because the assertion at https://github.com/openthread/openthread/blob/d65f8a3812611a89f851a18df2d4b8219a6fbca2/src/ncp/ncp_spi.cpp#L279 is the only place using the variable `readLength`. Build fails because of the "set but not used" error in some cases.
I changed this to use `OT_UNUSED_VARIABLE` after the assertion.

src/core/common/debug.hpp
Moandor-y: Done.

src/core/common/debug.hpp
```diff
@@ -47,7 +47,7 @@
 
 #define OT_ASSERT(cond) assert(cond)
 
-#elif OPENTHREAD_CONFIG_PLATFORM_ASSERT_MANAGEMENT
+#elif OPENTHREAD_CONFIG_PLATFORM_ASSERT_MANAGEMENT // defined(__APPLE__) || defined(__linux__)
```
abtink, line 50: Same suggestion as before. This is less than 20 lines from `#if`

src/core/common/debug.hpp
abtink: Yes. That's the better/correct way (we already use this approach in other parts (where a variable is only used in an `OT_ASSERT()`))

src/core/common/debug.hpp
Moandor-y: The `#endif` is more than 20 lines from `#if`. I think the `if-elif-else-endif` block should be considered as a whole.

abtink: The rational behind this is to help readers of code. Basically if we can see the `#if` condition itself and the comment line corresponding to `#if` in the same window/screen (i.e., they are less than 20 lines apart) we don't add the comment.

In this case, the `#if defined(__APPLE__) || defined(__linux__)` is just 6 lines above the newly added comment.
```
#if defined(__APPLE__) || defined(__linux__)

#include <assert.h>

#define OT_ASSERT(cond) assert(cond)

#elif OPENTHREAD_CONFIG_PLATFORM_ASSERT_MANAGEMENT // defined(__APPLE__) || defined(__linux__)
```

Moandor-y: Sounds reasonable. Updated.

abtink: Thanks 👍

PR #5168, opened 2020-06-30
[netif] misc enhancement

abtink:
This commit contains some smaller enhancements in netif:

- Add method to check if unicast/multicast address is external or
  internal (e.g., `IsUnicastAddressExternal`)
- Simplify iteration over (external) multicast addresses and remove
  `GetNextExternalMulticast()`.
- Use range-based `for` loop for iteration over external address
  arrays.
- Rename `IsUnicastAddress()` to `HasUnicastAddress()`.
- Clarify method documentation specially between internal and
  external addresses (how the passed-in address parameter is used,
  i.e., copied into another entry from an address pool for external
  vs. directly used for internal addresses).

### DIFF
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -1226,13 +1226,17 @@ bool Mle::HasUnregisteredAddress(void)
 
     if (!IsRxOnWhenIdle())
     {
-        uint8_t      iterator = 0;
-        Ip6::Address address;
-
         // For sleepy end-device, we register any external multicast
         // addresses.
 
-        retval = (Get<ThreadNetif>().GetNextExternalMulticast(iterator, address) == OT_ERROR_NONE);
+        for (const Ip6::NetifMulticastAddress *address = Get<ThreadNetif>().GetMulticastAddresses(); address != nullptr;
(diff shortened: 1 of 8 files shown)


src/core/thread/mle.cpp
```diff
@@ -1248,8 +1252,9 @@ otError Mle::AppendAddressRegistration(Message &aMessage, AddressRegistrationMod
     uint8_t                  length      = 0;
     uint8_t                  counter     = 0;
     uint16_t                 startOffset = aMessage.GetLength();
-    uint8_t                  iterator    = 0; // used to iterate external multicast addresses.
-    Ip6::Address             address;
+#if OPENTHREAD_CONFIG_DUA_ENABLE
+    Ip6::Address address;
```
jwhui, line 1256: nit: rename to `dua` or `domainUnicastAddress`?

abtink: Good suggestion. Thanks. Changed in new pushed commit.

abtink:
Added another small change in the new pushed commit:

Renamed  `Netif::IsUnicastAddress()` to  `Netif::HasUnicastAddress()`.

PR #5169, opened 2020-06-30
[dhcp6-client] only close socket when no agents and update service RLOC

### LINKED ISSUE #5166: DHCP client closes the socket and stops solicitation if two consecutive Data response message are received in a short time
If two messages with the same prefix arrive in a short time ( See void Dhcp6Client::UpdateAddresses(void) in  dhcp6_client.cpp)
The second message closes the socket before the first message (DHCP solicit) is handled. This happens because "newAgent" variable in the function will stay false.

The result is that a Thread device will not do the address solicitation successfully and does not get the global ipv6 address.

**To Reproduce** 

1. Git commit id: c8efee95f3be7aa3f35a6da50bca737d2e53ddc3 (src/core/net/dhcp6_client.cpp)
2. IEEE 802.15.4 hardware platform: Atmel at86rf233
3. Build steps: Reference platform should work fine for testing.
4. Network topology: Not relevant

**Expected behavior** 

If consecutive Data response messages are received with the same prefix, the second one should not close the socket.

### LINKED ISSUE #5167: When BR is a REED and its RLOC changes, DHCP client in node does not update RLOC in identity associations.
When BR is REED, it can keep changing its RLOC16 address frequently 
DHCP client in nodes does not detect this change when they get the Data response message.

**To Reproduce** 

1. Git commit id: c8efee9 (src/core/net/dhcp6_client.cpp)
2. IEEE 802.15.4 hardware platform: Atmel at86rf233
3. Build steps: should be reproducible in the reference platform.
4. Network topology: RLOC seems to change more often when there are 23 devices in the network (on the downgrade threshold). 

**Expected behavior** 

DHCP client should update RLOC in the funcion "void Dhcp6Client::UpdateAddresses(void)"
(mIdentityAssociations[i].mPrefixAgentRloc)

jwhui:
This PR includes the following fixes:
- only close socket if no agents exist (resolves #5166)
- update DHCPv6 service RLOC on new network data (resolves #5167)

### DIFF
--- a/src/core/net/dhcp6_client.cpp
+++ b/src/core/net/dhcp6_client.cpp
@@ -70,8 +70,8 @@ bool Dhcp6Client::MatchNetifAddressWithPrefix(const Ip6::NetifUnicastAddress &aN
 
 void Dhcp6Client::UpdateAddresses(void)
 {
-    bool                            found    = false;
-    bool                            newAgent = false;
+    bool                            found          = false;
+    bool                            doesAgentExist = false;
     NetworkData::Iterator           iterator;
     NetworkData::OnMeshPrefixConfig config;
 
@@ -121,7 +121,8 @@ void Dhcp6Client::UpdateAddresses(void)
             continue;
         }
 
-        found = false;
+        doesAgentExist = true;
+        found          = false;
 
         for (size_t i = 0; i < OT_ARRAY_LENGTH(mIdentityAssociations); i++)
         {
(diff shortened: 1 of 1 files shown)


src/core/net/dhcp6_client.cpp
```diff
@@ -144,21 +145,24 @@ void Dhcp6Client::UpdateAddresses(void)
...
             else
             {
                 otLogWarnIp6("Insufficient memory for new DHCP prefix");
+                continue;
             }
         }
+
+        ia->mPrefixAgentRloc = config.mRloc16;
+
+        doesAgentExist = true;
```
librasungirl, line 162: nit: `doesAgentExist` seems only related to whether or not there is DHCP prefix, regardless of the `found` result.
How about putting it directly after the flag check?
https://github.com/openthread/openthread/blob/c03a5628f742bb67bb8a1623db06da53f2e363af/src/core/net/dhcp6_client.cpp#L119-L122

PR #5171, opened 2020-07-01
[otns] fix simulation to use strong otPlatOtnsStatus definition

### LINKED ISSUE #5170: [github-actions] cli-ftd-otns failing in master

cli-ftd-otns GitHub Action is failing in master

1. Git commit id: 5ce4046
2. IEEE 802.15.4 hardware platform: N/A
3. Build steps: N/A
4. Network topology: N/A

**Expected behavior** A clear and concise description of what you expected to happen.

cli-ftd-otns GitHub Action passes.

**Console/log output** If applicable, add console/log output to help explain your problem.

https://github.com/openthread/openthread/runs/824271448

[logs_9382.zip](https://github.com/openthread/openthread/files/4854580/logs_9382.zip)

**Additional context** Add any other context about the problem here.

Started failing after merging #5155

simonlingoogle:
This PR fixes #5170 

This PR also fixes the issue that OTNS tests failed to use openthread of the current PR, but used openthread master instead. 

@ctan-g could you verify if this fix works for Silk?

### DIFF
--- a/examples/platforms/utils/otns_utils.cpp
+++ b/examples/platforms/utils/otns_utils.cpp
@@ -35,7 +35,7 @@
 
 /*
  * Implementation note:
- *   These are all "weak" so that a platform may if it chooses  override the instance.
+ *   These are all "weak" so that a platform may if it chooses to override the instance.
  */
 
 #if OPENTHREAD_CONFIG_OTNS_ENABLE
--- a/examples/platforms/utils/otns_utils.cpp
+++ b/examples/platforms/utils/otns_utils.cpp
@@ -1,0 +1,49 @@
+/*
+ *  Copyright (c) 2020, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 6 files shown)


examples/platforms/utils/otns_utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <openthread/config.h>
+
+#include <openthread/platform/otns.h>
+#include <openthread/platform/toolchain.h>
+
+#include "common/logging.hpp"
+
+/*
+ * Implementation note:
+ *   These are all "weak" so that a platform may if it chooses  override the instance.
```
ctan-g, line 38: Looks like an extra space after "chooses".

simonlingoogle:
> I am just curios, what was the issue with strong definition not being linked.

I am not clear why the `weak` definition is used. 
Actually if both definitions are strong, it still compiles without any complain. 
I suspect that the linker just found the first definition for the symbol and stopped searching any more even though the definition was weak. 
[Some](https://stackoverflow.com/a/13100465) say that `weak` symbols should be defined in libraries to be effective, which is consistent with what we have seen. 
@abtink @ctan-g

PR #5172, opened 2020-07-01
[test] add expect tests for coap and coaps

Moandor-y:
Also adds command `coaps set` to make it consistent with `coap set`.

### DIFF
--- a/src/cli/cli_coap.cpp
+++ b/src/cli/cli_coap.cpp
@@ -83,7 +83,8 @@ Coap::Coap(Interpreter &aInterpreter)
     memset(&mRequestUri, 0, sizeof(mRequestUri));
 #endif
     memset(&mUriPath, 0, sizeof(mUriPath));
-    strncpy(mResourceContent, "0", sizeof(mResourceContent) - 1);
+    strncpy(mResourceContent, "0", sizeof(mResourceContent));
+    mResourceContent[sizeof(mResourceContent) - 1] = '\0';
 }
 
 #if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
@@ -214,8 +215,9 @@ otError Coap::ProcessSet(uint8_t aArgsLength, char *aArgs[])
 
     if (aArgsLength > 1)
     {
-        VerifyOrExit(strlen(aArgs[1]) < (kMaxBufferSize - 1), error = OT_ERROR_INVALID_ARGS);
-        strncpy(mResourceContent, aArgs[1], sizeof(mResourceContent) - 1);
+        VerifyOrExit(strlen(aArgs[1]) < sizeof(mResourceContent), error = OT_ERROR_INVALID_ARGS);
+        strncpy(mResourceContent, aArgs[1], sizeof(mResourceContent));
+        mResourceContent[sizeof(mResourceContent) - 1] = '\0';
 
 #if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
         if (mSubscriberTokenLength > 0)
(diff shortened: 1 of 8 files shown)


src/cli/cli_coap.cpp
```diff
@@ -214,8 +215,9 @@ otError Coap::ProcessSet(uint8_t aArgsLength, char *aArgs[])
 
     if (aArgsLength > 1)
     {
-        VerifyOrExit(strlen(aArgs[1]) < (kMaxBufferSize - 1), error = OT_ERROR_INVALID_ARGS);
-        strncpy(mResourceContent, aArgs[1], sizeof(mResourceContent) - 1);
+        VerifyOrExit(strlen(aArgs[1]) < sizeof(mResourceContent), error = OT_ERROR_INVALID_ARGS);
```
bukepo, line 218:
```suggestion
        VerifyOrExit(strlen(aArgs[1]) < sizeof(mResourceContent) - 1, error = OT_ERROR_INVALID_ARGS);
```

src/cli/cli_coap_secure.cpp
```diff
@@ -135,6 +138,25 @@ otError CoapSecure::ProcessResource(uint8_t aArgsLength, char *aArgs[])
     return error;
 }
 
+otError CoapSecure::ProcessSet(uint8_t aArgsLength, char *aArgs[])
+{
+    otError error = OT_ERROR_NONE;
+
+    if (aArgsLength > 1)
+    {
+        VerifyOrExit(strlen(aArgs[1]) < sizeof(mResourceContent), error = OT_ERROR_INVALID_ARGS);
```
bukepo, line 147:
```suggestion
        VerifyOrExit(strlen(aArgs[1]) < sizeof(mResourceContent) - 1, error = OT_ERROR_INVALID_ARGS);
```

src/cli/cli_coap.cpp
Moandor-y: This isn't needed. The string length can be equal to `size - 1` without any issues.

src/cli/cli_coap.cpp
```diff
@@ -83,7 +83,8 @@ Coap::Coap(Interpreter &aInterpreter)
     memset(&mRequestUri, 0, sizeof(mRequestUri));
 #endif
     memset(&mUriPath, 0, sizeof(mUriPath));
-    strncpy(mResourceContent, "0", sizeof(mResourceContent) - 1);
+    strncpy(mResourceContent, "0", sizeof(mResourceContent));
+    mResourceContent[sizeof(mResourceContent) - 1] = '\0';
```
jwhui, line 87: nit: is explicitly adding NULL character necessary in this case?

src/cli/cli_coap_secure.cpp
```diff
@@ -70,6 +71,8 @@ CoapSecure::CoapSecure(Interpreter &aInterpreter)
     memset(&mResource, 0, sizeof(mResource));
     memset(&mPsk, 0, sizeof(mPsk));
     memset(&mPskId, 0, sizeof(mPskId));
+    strncpy(mResourceContent, "0", sizeof(mResourceContent));
+    mResourceContent[sizeof(mResourceContent) - 1] = '\0';
```
jwhui, line 75: nit: is explicitly adding NULL character necessary in this case?

src/cli/cli_coap.cpp
Moandor-y: It's not necessary. But I want to make sure the array is null-terminated in an explicit way, because codes may be changing in some unexpected ways, e.g. `"0"` being changed to another string or size of `mResourceContent` being changed, which may not be so obvious. I think it's better to be extra cautious when dealing with things like buffer overflow which may cause serious security issues.

PR #5178, opened 2020-07-01
[efr32] remove unnecessary nvm3 calls

lmnotran:
This removes all unnecessary calls to nvm3 open and close that were being made when handling nvm operations.

In a normal mode of operation, the nvm open call should be made only once during initialization.

The only way these changes could be verified was by building a binary, flashing it on a node and making sure that the node's network operations (like forming a network) were okay.


romacdon:
@googlebot I signed it!

PR #5181, opened 2020-07-02
[test] print debug info when expect tests fail

Moandor-y:
This stores debug logs from `expect` and prints them if a test fails to enable easier debugging.

### DIFF
--- a/script/test
+++ b/script/test
@@ -239,6 +239,9 @@ do_expect()
             echo -e "${COLOR_FAIL}FAIL${COLOR_NONE} ${script}"
             exit "${exit_code}"
         }
+        if [[ ${VERBOSE} == 1 ]]; then
+            cat "${log_file}" >&2
+        fi
         echo -e "${COLOR_PASS}PASS${COLOR_NONE} ${script}"
     done < <(
         if [[ $# != 0 ]]; then
--- a/script/test
+++ b/script/test
@@ -219,13 +219,10 @@ do_expect()
         test_patterns=(-name 'cli-*.exp' -o -name 'simulation-*.exp')
     fi
 
-    local temp_dir
-    temp_dir=$(mktemp -d)
-    trap 'rm -r "$temp_dir"' EXIT
-
-    local log_file="${temp_dir}/log_expect"
+    local log_file="tmp/log_expect"
     while read -r script; do
         sudo rm -rf tmp
+        mkdir tmp
         {
             if [[ ${OT_NATIVE_IP} == 1 ]]; then
                 OT_COMMAND="${ot_command}" RCP_COMMAND="${rcp_command}" sudo -E expect -df "${script}" 2>"${log_file}"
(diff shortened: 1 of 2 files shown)


script/test
```diff
@@ -219,16 +219,23 @@ do_expect()
...
+    trap 'rm -r "$temp_dir"' EXIT
+
+    local log_file
+    log_file="${temp_dir}/log_expect"
     while read -r script; do
         sudo rm -rf tmp
         {
             if [[ ${OT_NATIVE_IP} == 1 ]]; then
-                OT_COMMAND="${ot_command}" RCP_COMMAND="${rcp_command}" sudo -E "${script}"
+                OT_COMMAND="${ot_command}" RCP_COMMAND="${rcp_command}" sudo -E expect -df "${script}" 2>"${log_file}"
```
simonlingoogle, line 232: `test` script has the `VERBOSE` option. 
What about we print the debug log when `VERBOSE=1` and then use `VERBOSE=1` in GitHub actions.

Moandor-y: Changed to always print debug logs when `VERBOSE=1`.
But didn't turn on verbose mode in GitHub actions. That will cause too many outputs. We only need these logs when tests fail.

simonlingoogle: If we just use `VERBOSE=1` in github actions, we don't need all these logic of redirecting log to a tmp file and printing it when failed.
Besides, having more debug logs in github action does not seems to be an issue for me.

Moandor-y: The concern about the huge number of logs is that in GitHub Action when it fails we have to scroll to the bottom manually to see what happened, the loading and scrolling lags with too many logs.
Also when running locally, I think it's nice to have detailed logs printed for failed cases by default, so we don't have to run again with debug mode if something fails.
Another benefit of redirecting is that `stdout` and `stderr` won't interleave, which makes them easier to read.

simonlingoogle: > The concern about the huge number of logs is that in GitHub Action when it fails we have to scroll to the bottom manually to see what happened, the loading and scrolling lags with too many logs.

This can be solved by using `View raw logs` or `Download log archive`. BTW, building openthread itself already generates massive logs.

> Also when running locally, I think it's nice to have detailed logs printed for failed cases by default, so we don't have to run again with debug mode if something fails.

It's nice to always print logs if failed but it's also very simple to run it again locally. 
I am fine with the current strategy but I do prefer a simpler solution which has no obvious disadvantages.

bukepo: @simonlingoogle The certification tests only print logs of failed cases. I don't think it's necessary to print passed tests. Too many unnecessary logs increases the difficulty to debug when bad things happen.

script/test
```diff
@@ -219,19 +219,28 @@ do_expect()
         test_patterns=(-name 'cli-*.exp' -o -name 'simulation-*.exp')
     fi
 
+    local temp_dir
+    temp_dir=$(mktemp -d)
```
bukepo, line 223: create a temp file instead.

Moandor-y: Done. Now using current directory to save the log file.

PR #5184, opened 2020-07-02
[nrf528xx] add platform radio support for transmit security

brianljt:


examples/platforms/nrf528xx/src/radio.c
```diff
@@ -824,6 +921,24 @@ void nrf_802154_received_timestamp_raw(uint8_t *p_data, int8_t power, uint8_t lq
...
+        receivedFrame->mInfo.mRxInfo.mAckedWithSecEnhAck = sAckedWithSecEnhAck;
+        receivedFrame->mInfo.mRxInfo.mAckFrameCounter    = sAckFrameCounter;
+        receivedFrame->mInfo.mRxInfo.mAckKeyId           = sAckKeyId;
+    }
+    else
+    {
+        receivedFrame->mInfo.mRxInfo.mAckedWithSecEnhAck = false;
+        receivedFrame->mInfo.mRxInfo.mAckFrameCounter    = 0;
+        receivedFrame->mInfo.mRxInfo.mAckKeyId           = 0;
+    }
```
bukepo, line 937: already initialized to at the beginning.
```suggestion
```

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -191,6 +204,86 @@ static inline void clearPendingEvents(void)
     } while (__STREXW(pendingEvents, (uint32_t *)&sPendingEvents));
 }
 
+#if OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT
+static void nrf_802154_tx_process_security(uint8_t *p_ack_frame)
```
bukepo, line 208: It seems there are very little common code between tx ack and tx other frames. It seems reasonable to separate code into corresponding functions: nrf_802154_tx_started() and nrf_802154_tx_ack_started(). Thoughts?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -971,3 +1076,30 @@ uint32_t nrf_802154_random_get(void)
...
+                          uint8_t         aKeyIdMode,
+                          uint8_t         aKeyId,
+                          const otMacKey *aPrevKey,
+                          const otMacKey *aCurrKey,
+                          const otMacKey *aNextKey)
+{
+    OT_UNUSED_VARIABLE(aInstance);
+    OT_UNUSED_VARIABLE(aKeyIdMode);
+
+    assert(aPrevKey != NULL && aCurrKey != NULL && aNextKey != NULL);
```
LuDuda, line 1091: Do we ensure that Radio is in the Sleep state while we change the keys? Otherwise, we may experience some race conditions (e.g. when encrypting the ACK). The alternative would be to put this code into the critical section.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -429,7 +493,11 @@ otRadioCaps otPlatRadioGetCaps(otInstance *aInstance)
     OT_UNUSED_VARIABLE(aInstance);
 
     return (otRadioCaps)(OT_RADIO_CAPS_ENERGY_SCAN | OT_RADIO_CAPS_ACK_TIMEOUT | OT_RADIO_CAPS_CSMA_BACKOFF |
-                         OT_RADIO_CAPS_SLEEP_TO_TX);
+                         OT_RADIO_CAPS_SLEEP_TO_TX
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+                         | OT_RADIO_CAPS_TRANSMIT_SEC
+#endif
+    );
```
bukepo, line 500: nit
```suggestion
#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
                         OT_RADIO_CAPS_TRANSMIT_SEC |
#endif
                         OT_RADIO_CAPS_SLEEP_TO_TX);
```

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -191,6 +204,57 @@ static inline void clearPendingEvents(void)
     } while (__STREXW(pendingEvents, (uint32_t *)&sPendingEvents));
 }
 
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+static void txAckProcessSecurity(uint8_t *p_ack_frame)
```
bukepo, line 208: nit:
```suggestion
static void txAckProcessSecurity(uint8_t *aAckFrame)
```

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -191,6 +204,57 @@ static inline void clearPendingEvents(void)
...
+    {
+        key = &sPrevKey;
+    }
+    else if (keyId == sKeyId + 1)
+    {
+        key = &sNextKey;
+    }
+    else
+    {
+        otEXPECT(false);
```
bukepo, line 239: Add log here, or assert(false)? It seems there is a logic error if this happens.

examples/platforms/nrf528xx/src/radio.c
brianljt: Yes, setting MAC key will be a problem if ACK is generated in interrupt. I would prefer putting it into critical sections rather than deferring the setting until radio is in sleep state.

So this means I have to expose `nrf_802154_critical_section_enter` and `nrf_802154_critical_section_exit` to radio.c

examples/platforms/nrf528xx/src/radio.c
brianljt: As discussed, it is possible for a node to received a frame with "bad" key index. It should just drop it. This implementation follows `Mac::ProcessReceiveSecurity`.

>From 15.4-2015 section 7.3.3
If the Security Enabled field in the Frame Control field is set to one, the Enh-Ack frame is protected using the procedure described in 9.2.1. The SecurityLevel, KeyIdMode, KeySource and KeyIndex shall be set to match the corresponding fields of the frame that is being acknowedged.

>From Thread 1.2 section 7.1.5
Typically, three keys are stored ... This storage of multiple keys allows migration to a new key while still retaining the old key.

>From 15.4-2015 section 9.2.1
c)Obtain KeyDescriptor. The procedure shall obtain the  KeyDescriptor using the KeyDescriptor lookup procedure as described in 9.2.2 using the KeyIdMode, KeyIndex, and KeySource with the DeviceAddressingMode  set  to  Destination  Addressing  Mode  field, DevicePanId  set  to  theDestination PAN ID field, and DeviceAddress set to the Destination Address field. If that procedure fails, the procedure shall return with a Status of UNAVAILABLE_KEY.

So if key could not be found, stack will escape the encryption procedure and the frame will be sent out unencrypted.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -971,3 +1076,34 @@ uint32_t nrf_802154_random_get(void)
...
+                          const otMacKey *aPrevKey,
+                          const otMacKey *aCurrKey,
+                          const otMacKey *aNextKey)
+{
+    OT_UNUSED_VARIABLE(aInstance);
+    OT_UNUSED_VARIABLE(aKeyIdMode);
+
+    assert(aPrevKey != NULL && aCurrKey != NULL && aNextKey != NULL);
+
+    if (nrf_802154_critical_section_enter())
```
LuDuda, line 1093: We should not use this function. I do see two possibilities, either:

1) Have a double buffering for keys, and make simple switch between them after keys are coppied.
2) Use `CRITICAL_REGION_ENTER();` and `CRITICAL_REGION_EXIT();` macros, which will handle also cases where SoftDevice is used. I believe the proper headers are already included.

brianljt: Thank you for reminding me. Updated.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -971,3 +1075,34 @@ uint32_t nrf_802154_random_get(void)
...
+    memcpy(sNextKey.m8, aNextKey->m8, OT_MAC_KEY_SIZE);
+
+    CRITICAL_REGION_EXIT();
+}
+
+void otPlatRadioSetMacFrameCounter(otInstance *aInstance, uint32_t aMacFrameCounter)
+{
+    OT_UNUSED_VARIABLE(aInstance);
+
+    sMacFrameCounter = aMacFrameCounter;
```
abtink, line 1106: Do we need to protect the `sMacFrameCounter = aMacFrameCounter;` here , i.e., wrap it in a critical section?

I see that we can do  `sMacFrameCounter++ ` from `nrf_802154_tx_started()`. If this is called from ISR, then we need to make sure the write to `sMacFrameCounter` is atomic.

brianljt: It is one line assignment statement, I think itself has atomicity thus did not add critical section protection.

brianljt: OK. I see your point, the assignment might not be translated to single assembly code, so critical section protection is still needed.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -857,14 +936,22 @@ void nrf_802154_receive_failed(nrf_802154_rx_error_t error)
...
 
-void nrf_802154_tx_ack_started(const uint8_t *p_data)
+void nrf_802154_tx_ack_started(uint8_t *p_data)
 {
     // Check if the frame pending bit is set in ACK frame.
     sAckedWithFramePending = p_data[FRAME_PENDING_OFFSET] & FRAME_PENDING_BIT;
+
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+    // Update IE and secure Enh-ACK.
+    txAckProcessSecurity(p_data);
```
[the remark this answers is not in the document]
LuDuda: I've consulted with @hubertmis and it seems we don't have any exact numbers now. If ACK is sent correctly we are on the good side - IRQ has the highest priority. As i mentioned the future version of Radio Driver should entirely solve this.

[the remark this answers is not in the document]
jwhui: Sounds good. Thanks @LuDuda !

Issue #5142, opened 2020-06-23
[meshcop] Nordic devices cannot join with external commissioner


wgtdkp:
**Describe the bug** A clear and concise description of what the bug is.

Thread Commissioning with nrf52840 devices failed because OTBR cannot forward `Client Hello` sent by the joiner.

**To Reproduce** Information to reproduce the behavior, including:
Follow the guide https://openthread.io/guides/border-router/external-commissioning?comm=ot-commissioner.

1. Git commit id: openthread master HEAD
2. IEEE 802.15.4 hardware platform: nrf52840
3. Build steps
4. Network topology: External Commissioner <-----> OTBR <------> Joiner

**Expected behavior** A clear and concise description of what you expected to happen.
The joining process shall success and "Join success" is printed by the joiner.

**Console/log output** If applicable, add console/log output to help explain your problem.
- The joining process failed with `Security` error:
    ```text
    > Join failed [Security]
    ```
- in OTBR log, `NoBufs` error is reported:
    ```text
    otbr-agent[8408]: [INFO]-PLAT----: processReceive: OK
    otbr-agent[8408]: [INFO]-MESH-CP-: JoinerRouter::HandleUdpReceive
    otbr-agent[8408]: [INFO]-MESH-CP-: Sent relay rx
    otbr-agent[8408]: [INFO]-PLAT----: processTransmit: OK
    otbr-agent[8408]: [INFO]-MESH-CP-: Sent to commissioner
    otbr-agent[8408]: [INFO]-MESH-CP-: Sent to commissioner on c/rx
    otbr-agent[8408]: [NOTE]-MESH-CP-: CoapSecure Transmit: NoBufs
    ```

**Additional context** Add any other context about the problem here.
The same commissioning process is executed with simulated devices (OTBR and Joiner) and it works well.  The traffic of the nrf52840 case shows that it is keep retransmiting the `Client Hello` message when including cookie. This is because the OTBR failed to forward this `Client Hello` message to the external commissioner (the `NoBufs` error). The `Client Hello` message is too long to be encoded within the `RLY_RX` message with a maximum length of `kApplicationDataMaxLength = 512` defined at:
https://github.com/openthread/openthread/blob/5c722a770c6f3a0627c93ec5ae3bb3e13e491bbe/src/core/meshcop/dtls.hpp#L74-L76

The simulated joiner works because it includes less support ECP group extension in `Client Hello` message:

`Client Hello` sent by nrf52840 joiner:
```text
[...]

PR #5192, opened 2020-07-03
[dtls] explicitly configure ECJPAKE parameters

edmont:
This PR uses the mbedtls API to explicitly configure DTLS to use only the `secp256r1` curve and no signature hash algorithms when using ECJPAKE cipher suite.

This fixes https://github.com/openthread/openthread/issues/5142

### DIFF
--- a/src/core/meshcop/dtls.cpp
+++ b/src/core/meshcop/dtls.cpp
@@ -56,6 +56,11 @@
 namespace ot {
 namespace MeshCoP {
 
+const mbedtls_ecp_group_id Dtls::sCurves[] = {MBEDTLS_ECP_DP_SECP256R1, MBEDTLS_ECP_DP_NONE};
+#if defined(MBEDTLS_KEY_EXCHANGE__WITH_CERT__ENABLED)
+const int Dtls::sHashes[] = {MBEDTLS_MD_NONE};
+#endif
+
 Dtls::Dtls(Instance &aInstance, bool aLayerTwoSecurity)
     : InstanceLocator(aInstance)
     , mState(kStateClosed)
@@ -97,10 +102,6 @@ Dtls::Dtls(Instance &aInstance, bool aLayerTwoSecurity)
 #endif
 
     memset(mCipherSuites, 0, sizeof(mCipherSuites));
-    memset(mCurves, 0, sizeof(mCurves));
-#if defined(MBEDTLS_KEY_EXCHANGE__WITH_CERT__ENABLED)
-    memset(mHashes, 0, sizeof(mHashes));
-#endif
     memset(mPsk, 0, sizeof(mPsk));
     memset(&mSsl, 0, sizeof(mSsl));
     memset(&mConf, 0, sizeof(mConf));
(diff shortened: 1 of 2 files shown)


src/core/meshcop/dtls.cpp
```diff
@@ -285,7 +286,17 @@ otError Dtls::Setup(bool aClient)
     mbedtls_ssl_conf_max_version(&mConf, MBEDTLS_SSL_MAJOR_VERSION_3, MBEDTLS_SSL_MINOR_VERSION_3);
 
     OT_ASSERT(mCipherSuites[1] == 0);
+    OT_ASSERT(mCurves[1] == MBEDTLS_ECP_DP_NONE);
     mbedtls_ssl_conf_ciphersuites(&mConf, mCipherSuites);
+    mbedtls_ssl_conf_curves(&mConf, mCurves);
+#if defined(MBEDTLS_KEY_EXCHANGE__WITH_CERT__ENABLED)
+    if (mCipherSuites[0] == MBEDTLS_TLS_ECJPAKE_WITH_AES_128_CCM_8)
+    {
+        int hashes[1];
```
bukepo, line 295: From the implementation of mbedtls_ssl_conf_sig_hashes(), it seems it cannot accept a list in a stack.

edmont: @bukepo you are right, I'll add mHashes to Dtls class.

And I guess it would be nice to restrict the SHAs for ECDSA as well, but not sure which are required...

bukepo: I think these restrictions should only be applied to the dtls socket used for Thread commissioning. Could we keep the base Dtls not restricted? Maybe we can move the current `ot::MeshCoP::Dtls` to `ot::Net::Dtls`, and the `ot::MeshCoP::Dtls` inherits the basic one and adds these restrictions. Thoughts?

edmont: @bukepo I would say that makes sense, however it's a bigger change.

By the moment I'll redo this PR just to match its description regarding ECJPAKE only.

src/core/meshcop/dtls.hpp
```diff
@@ -409,9 +409,13 @@ class Dtls : public InstanceLocator
...
 
-    int     mCipherSuites[2];
-    uint8_t mPsk[kPskMaxLength];
-    uint8_t mPskLength;
+    int                  mCipherSuites[2];
+    uint8_t              mPsk[kPskMaxLength];
+    uint8_t              mPskLength;
+    mbedtls_ecp_group_id mCurves[2];
+#if defined(MBEDTLS_KEY_EXCHANGE__WITH_CERT__ENABLED)
+    int mHashes[1];
```
bukepo, line 417: can this be static const?

src/core/meshcop/dtls.hpp
```diff
@@ -409,9 +409,13 @@ class Dtls : public InstanceLocator
 
     State mState;
 
-    int     mCipherSuites[2];
-    uint8_t mPsk[kPskMaxLength];
-    uint8_t mPskLength;
+    int                  mCipherSuites[2];
+    uint8_t              mPsk[kPskMaxLength];
+    uint8_t              mPskLength;
+    mbedtls_ecp_group_id mCurves[2];
```
bukepo, line 415: Can this be static const?

src/core/meshcop/dtls.cpp
```diff
@@ -56,6 +56,11 @@
 namespace ot {
 namespace MeshCoP {
 
+const mbedtls_ecp_group_id Dtls::sCurves[] = {MBEDTLS_ECP_DP_SECP256R1, MBEDTLS_ECP_DP_NONE};
+#if defined(MBEDTLS_KEY_EXCHANGE__WITH_CERT__ENABLED)
```
jwhui, line 60: nit: the rest of this file uses `#ifdef`
```suggestion
#ifdef MBEDTLS_KEY_EXCHANGE__WITH_CERT__ENABLED
```

src/core/meshcop/dtls.cpp
```diff
@@ -286,6 +291,13 @@ otError Dtls::Setup(bool aClient)
 
     OT_ASSERT(mCipherSuites[1] == 0);
     mbedtls_ssl_conf_ciphersuites(&mConf, mCipherSuites);
+    if (mCipherSuites[0] == MBEDTLS_TLS_ECJPAKE_WITH_AES_128_CCM_8)
+    {
+        mbedtls_ssl_conf_curves(&mConf, sCurves);
+#if defined(MBEDTLS_KEY_EXCHANGE__WITH_CERT__ENABLED)
```
jwhui, line 297: nit: the rest of this file uses `#ifdef`
```suggestion
#ifdef MBEDTLS_KEY_EXCHANGE__WITH_CERT__ENABLED
```

PR #5193, opened 2020-07-03
[pool] introducing a generic object pool class

abtink:
This commit adds a template `Pool<Type, kPoolSize>` class representing
an object pool. `Message`, `Netif`, and `AddressResolver` classes are
respectively updated to use `Pool` for managing their buffer pool,
external unicast/multicast address pool, and address cache pool. This
commit also adds a unit test for the `Pool` class.

### DIFF
--- a/src/core/common/pool.hpp
+++ b/src/core/common/pool.hpp
@@ -1,0 +1,185 @@
+/*
+ *  Copyright (c) 2020, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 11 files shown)


src/core/common/pool.hpp
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * This method frees a previously allocated object.
+     *
+     * The @p aEntry MUST be an entry from the pool previously allocated using `Allocate()` method and not yet freed.
+     * An already freed entry MUST not be freed again.
+     *
+     * @param[in]  aEntry   The pool object entry to free.
+     *
+     */
+    void Free(Type &aEntry) { mFreeList.Push(aEntry); }
```
simonlingoogle, line 116: should `assert` that `aEntry` was allocated from this pool?

src/core/common/pool.hpp
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * This method indicates whether or not a given `Type` object is from the pool.
+     *
+     * @param[in]  aObject   A reference to a `Type` object.
+     *
+     * @retval TRUE if @p aObject is from the pool.
+     * @retval FALSE if @p aObject is not from the pool.
+     *
+     */
+    bool Contains(const Type &aObject) const { return (&mPool[0] <= &aObject) && (&aObject < OT_ARRAY_END(mPool)); }
```
simonlingoogle, line 135:
```suggestion
    bool IsPoolEntry(const Type &aObject) const { return (&mPool[0] <= &aObject) && (&aObject < OT_ARRAY_END(mPool)); }
```
It's reasonable to think that only unallocated entries are `Contained` in the pool.
Better to rename it?

abtink: Good suggestion. Thanks. Changed it to `IsPoolEntry()` in the new pushed commit.

src/core/common/pool.hpp
abtink: I initially considered adding such an `assert` (also adding an `assert` check to ensure entry is not already freed), but decided against adding them:
- Main reason is to avoid the extra overhead (in term of code size and run-time checks) that such asserts would add (in all usages of these methods), particularly since the template methods are typically inlined, so these extra checks would be included everywhere.
- I feel incorrect use of the `Pool` methods would be rather obvious error (and therefore unlikely).
   - In most use-cases, we would simply have a pool of objects that we just allocate entries from, and then later free, and incorrect use of the `Pool` APIs would be almost not possible (or a very obvious coding error).
   - If there are more complex usage where we may benefit from the extra `assert` checks, I think better approach would be to leave it to the users of the `Pool` method to add it (e.g., we can define a subclass of a specific `Pool<Type>` and re-define `Free()` in it to add any extra checks/asserts, and this would apply to that sub-class type).
- Also the existing implementations (which are being replaced by `Pool` class methods) do not include such `assert` checks.

simonlingoogle: Thanks for explanation. Your concerns are reasonable to me.

src/core/common/pool.hpp
```diff
@@ -0,0 +1,185 @@
...
+ *   This module includes definitions for OpenThread object pool.
+ *
+ * @{
+ *
+ */
+
+/**
+ * This template class represents an object pool.
+ *
+ * The template type `Type` should provide `GetNext() and `SetNext()` so that it can be added to a linked list.
```
jwhui, line 59: nit: should we start using `@tparam` to document template parameters?

https://www.doxygen.nl/manual/commands.html#cmdtparam

abtink: Good idea. Did not know about this.

PR #5194, opened 2020-07-05
[ip6-address] update and use Ip6::InterfaceIdentifier class

abtink:
This commit updates `Ip6::InterfaceIdentifier` class adding new helper methods to it (e.g., generate random IID, or convert IID to/from a MAC Extended Address). It also updates the `Ip6::Address` methods (related to getting/setting IID) to use the `Ip6::InterfaceIdentifier` type.

The larger change in this commit is to update other core classes and modules (`Lowpan`, `Mle`, `Settings`, `AddressResolver`, `Child`, etc.) to use the `Ip6::InterfaceIdentifier`/`Ip6::Address` type and their new/changed methods.

This commit also adds a new `struct` type `otIp6AddressComponents` to the public OT API header `ip6.h`. This type represents the components of a (unicast) IPv6 address, i.e. a network prefix and an IID. The `otIp6Address` uses the new type as one of its `union` representations of an IPv6 address.

### DIFF
--- a/include/openthread/ip6.h
+++ b/include/openthread/ip6.h
@@ -66,7 +66,12 @@ extern "C" {
 OT_TOOL_PACKED_BEGIN
 struct otIp6InterfaceIdentifier
 {
-    uint8_t m8[OT_IP6_IID_SIZE]; ///< The Interface Identifier of an IPv6 address.
+    union OT_TOOL_PACKED_FIELD
+    {
+        uint8_t  m8[OT_IP6_IID_SIZE];                     ///< 8-bit fields
+        uint16_t m16[OT_IP6_IID_SIZE / sizeof(uint16_t)]; ///< 16-bit fields
+        uint32_t m32[OT_IP6_IID_SIZE / sizeof(uint32_t)]; ///< 32-bit fields
+    } mFields;                                            ///< The Interface Identifier accessor fields
 } OT_TOOL_PACKED_END;
 
 /**
@@ -74,6 +79,43 @@ struct otIp6InterfaceIdentifier
  *
  */
 typedef struct otIp6InterfaceIdentifier otIp6InterfaceIdentifier;
+
+/**
+ * @struct otIp6NetworkPrefix
+ *
+ * This structure represents the Network Prefix of an IPv6 address (most significant 64 bits of the address).
+ *
+ */
+OT_TOOL_PACKED_BEGIN
+struct otIp6NetworkPrefix
+{
+    uint8_t m8[OT_IP6_PREFIX_SIZE]; ///< The Network Prefix.
+} OT_TOOL_PACKED_END;
+
+/**
+ * This structure represents the Network Prefix of an IPv6 address (most significant 64 bits of the address).
+ *
+ */
+typedef struct otIp6NetworkPrefix otIp6NetworkPrefix;
+
+/**
+ * @struct otIp6AddressComponents
+ *
+ * This structure represents the components of an IPv6 address.
+ *
+ */
+OT_TOOL_PACKED_BEGIN
+struct otIp6AddressComponents
+{
+    otIp6NetworkPrefix       mNetworkPrefix; ///< The Network Prefix (most significant 64 bits of the address)
+    otIp6InterfaceIdentifier mIid;           ///< The Interface Identifier (least significant 64 bits of the address)
+} OT_TOOL_PACKED_END;
+
+/**
+ * This structure represents the components of an IPv6 address.
+ *
+ */
+typedef struct otIp6AddressComponents otIp6AddressComponents;
 
 /**
  * @struct otIp6Address
@@ -86,10 +128,11 @@ struct otIp6Address
 {
     union OT_TOOL_PACKED_FIELD
     {
-        uint8_t  m8[OT_IP6_ADDRESS_SIZE];                     ///< 8-bit fields
-        uint16_t m16[OT_IP6_ADDRESS_SIZE / sizeof(uint16_t)]; ///< 16-bit fields
-        uint32_t m32[OT_IP6_ADDRESS_SIZE / sizeof(uint32_t)]; ///< 32-bit fields
-    } mFields;                                                ///< IPv6 accessor fields
+        uint8_t                m8[OT_IP6_ADDRESS_SIZE];                     ///< 8-bit fields
+        uint16_t               m16[OT_IP6_ADDRESS_SIZE / sizeof(uint16_t)]; ///< 16-bit fields
+        uint32_t               m32[OT_IP6_ADDRESS_SIZE / sizeof(uint32_t)]; ///< 32-bit fields
+        otIp6AddressComponents mComponents;                                 ///< IPv6 address components
+    } mFields;                                                              ///< IPv6 accessor fields
 } OT_TOOL_PACKED_END;
 
 /**
--- a/src/core/net/ip6_address.hpp
+++ b/src/core/net/ip6_address.hpp
@@ -66,10 +66,13 @@ class InterfaceIdentifier : public otIp6InterfaceIdentifier,
                             public Equatable<InterfaceIdentifier>,
                             public Clearable<InterfaceIdentifier>
 {
+    friend class Address;
+
 public:
     enum
     {
-        kInfoStringSize = 17, // Max chars for the info string (`ToString()`).
+        kSize           = OT_IP6_IID_SIZE, ///< Size of an IPv6 Interface Identifier (in bytes).
+        kInfoStringSize = 17,              ///< Max chars for the info string (`ToString()`).
     };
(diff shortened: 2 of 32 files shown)


src/core/net/ip6_address.hpp
```diff
@@ -484,66 +532,59 @@ class Address : public otIp6Address, public Equatable<Address>, public Clearable
...
-     * This method sets the Interface Identifier.
+     * This method sets the Interface Identifier from a given byte array.
      *
-     * @param[in]  aIid  A reference to the Interface Identifier.
+     * @param[in] aBuffer    Pointer to an array containing the Interface Identifier. `InterfaceIdentifier::kSize`
+     *                       bytes from the buffer are copied to form the Interface Identifier.
      *
      */
-    void SetIid(const InterfaceIdentifier &aIid) { SetIid(aIid.m8); }
+    void SetIidFromByteArray(const uint8_t *aIidBuffer) { GetIid().SetFromByteArray(aIidBuffer); }
```
simonlingoogle, line 579: nit: Can we remove these `SetIidFrom...` methods because typing `GetIid().SetFromByteArray(aIidBuffer)` is not much more difficult than typing `SetIidFromByteArray(aIidBuffer)`?

src/core/net/ip6_address.hpp
```diff
@@ -96,6 +97,53 @@ class InterfaceIdentifier : public otIp6InterfaceIdentifier,
...
+     */
+    void GenerateRandom(void);
+
+    /**
+     * This method gets the Interface Identifier as a pointer to a byte array.
+     *
+     * @returns A pointer to a byte array (of size `kSize`) containing the Interface Identifier.
+     *
+     */
+    const uint8_t *GetAsByteArray(void) const { return mFields.m8; }
```
simonlingoogle, line 112: nit:
```suggestion
    const uint8_t *GetBytes(void) const { return mFields.m8; }
```

src/core/net/ip6_address.hpp
```diff
@@ -96,6 +97,53 @@ class InterfaceIdentifier : public otIp6InterfaceIdentifier,
...
+     */
+    void SetFromByteArray(const uint8_t *aBuffer);
+
+    /**
+     * This method sets the Interface Identifier from a given IEEE 802.15.4 Extended Address.
+     *
+     * @param[in] aExtAddress  An Extended Address.
+     *
+     */
+    void SetFromExtAddress(const Mac::ExtAddress &aExtAddress);
```
simonlingoogle, line 129: nit: rename to `FromExtAddress` so that it's symmetric with `ToExtAddress`?

```suggestion
    void FromExtAddress(const Mac::ExtAddress &aExtAddress);
```

src/core/net/ip6_address.hpp
```diff
@@ -96,6 +97,53 @@ class InterfaceIdentifier : public otIp6InterfaceIdentifier,
...
+    const uint8_t *GetAsByteArray(void) const { return mFields.m8; }
+
+    /**
+     * This method sets the Interface Identifier from a given byte array.
+     *
+     * @param[in] aBuffer    Pointer to an array containing the Interface Identifier. `kSize` bytes from the buffer
+     *                       are copied to form the Interface Identifier.
+     *
+     */
+    void SetFromByteArray(const uint8_t *aBuffer);
```
jwhui, line 121: Just curious - when do we make the method name explicit vs. use method overloading?

src/core/net/ip6_address.hpp
abtink: Nice idea. I agree it may be cleaner to not repeat the same methods over in the `Ip6::Address`

src/core/net/ip6_address.hpp
abtink: Even though I think we don't explicitly mention it in our style guide, in terms of method names, I like that we start the name with a "verb/action" (I feel it reads better from the caller side - i.e., it's clear what action we are performing on the object). 

The only exception is `To{SomeType}()`, even then in some cases I actually prefer using `ConvertTo{SomeType}()` name. 

Here is my suggestion for this:
- We use `To{SomeType}()`name in cases the method converts and then returns the given type (like `ToString()` which returns a `String` instance).
- We use `ConvertTo{SomeType}(SomeType &)` method name in cases where the method convertsto a new type and outputs the converted value in a passed-in reference parameter.
- i.e.,
```
class Type
{
    OtherType ToOtherType(void) const;                    // First option (preferred)     
    void ConvertToOtherType(OtherType &aOtherType) const; // Second option (can return `otError` if conversion can fail)
    
    void ToOtherType(OtherType &aOtherType) const;        // Third option (not preferred).
};
```

I feel statement like 
```
    macAddr = meshLocalIid.ToExtAddress();
```
is more readable/clear compared to 
``` 
    meshLocalIid.ToExtAddress(macAddr);
```

With OT moving to C++11 and almost all toolchains supporting "return value optimization" and ["copy elision"](https://en.cppreference.com/w/cpp/language/copy_elision), the first option should not add any copying overhead and should be as efficient as the second option.

Thoughts? @jwhui?

src/core/net/ip6_address.hpp
abtink: Here are my thoughts/suggestions on this:

- For `set`ting an object in cases where a conversion/parsing is needed, I would say it'd better to require and use an explicit method name:
   - `Iid::SetFromExtAddress()`  is a good example (where we toggle the "local" bit to convert `ExtAddress` to IID),
   -  or `Dataset::SetFrom(const otOperationalDataset &)` is another good example. 
   - [in light of this, I would suggest renaming `Dataset::Set(const Message &, ...);` to `ReadFromMessage()`?]
- For cases where the value is directly used (no change/conversion), either way (use explicit method name or overload) is fine.
   - I personally prefer explicit name in this case as well (feel it is more readable and avoid any incorrect use), but we can leave it to developer.

Thoughts? @jwhui ?

src/core/net/ip6_address.hpp
simonlingoogle: I would also prefer `OtherType ToOtherType(void) const;` if there is no performance cost.

src/core/net/ip6_address.hpp
abtink: Updated this in the new pushed commit. 
Also moved more of the methods related to IID from `Address` to `InterfaceIdentifier` class.

src/core/net/ip6_address.hpp
jwhui: Agree. Use `CovertTo()` if the method needs to return error for any reason.

src/core/net/ip6_address.hpp
jwhui: I generally prefer being explicit as well.

include/openthread/ip6.h
```diff
@@ -75,6 +80,44 @@ struct otIp6InterfaceIdentifier
...
+ */
+OT_TOOL_PACKED_BEGIN
+struct otIp6AddressComponents
+{
+    otIp6NetworkPrefix       mNetworkPrefix; ///< The Network Prefix (most significant 64 bits of the address)
+    otIp6InterfaceIdentifier mIid;           ///< The Interface Identifier (least significant 64 bits of the address)
+} OT_TOOL_PACKED_END;
+
+/**
+ *
```
jwhui, line 115:
```suggestion
```

src/core/net/ip6_address.hpp
```diff
@@ -96,13 +99,162 @@ class InterfaceIdentifier : public otIp6InterfaceIdentifier,
...
+     * @param[in]  aLocator    RLOC16 or ALOC16.
+     *
+     */
+    void SetToLocator(uint16_t aLocator);
+
+    /**
+     * This method indicates whether or not the Interface Identifier matches the locator pattern `0000:00ff:fe00:xxxx`.
+     *
+     * @retval TRUE   If the IID matches the locator pattern
+     * @retval FALSE  If the IID does not match the locator pattern
```
jwhui, line 180:
```suggestion
     * @retval TRUE   If the IID matches the locator pattern.
     * @retval FALSE  If the IID does not match the locator pattern.
```

src/core/net/ip6_address.hpp
```diff
@@ -401,7 +468,7 @@ class Address : public otIp6Address, public Equatable<Address>, public Clearable
      * @param[in]  aLocator   RLOC16 or ALOC16.
      *
      */
-    void SetLocator(uint16_t aLocator) { mFields.m16[7] = HostSwap16(aLocator); }
+    // void GetIid().SetLocator(uint16_t aLocator) { mFields.m16[7] = HostSwap16(aLocator); }
```
jwhui, line 471: Remove?

abtink: Thanks.

PR #5197, opened 2020-07-06
[efr32] process received packet immediately

lmnotran:
When analyzing the energy mode transitions for a sleepy node, running the `sleepy-demo-mtd`  app, it was observed that, the current implementation of the platform radio driver (`efr32RadioProcess()`) although correct, can be optimized by processing the next received packet (`processNextRxPacket()`) before other actions being carried out in this routine. For example, in a situation where a 802.15.4 MAC Ack needs to be processed in response to a sent packet,  adding this call towards the end of this routine then delays stack notification (that a Tx was done) until the next call to `efr32RadioProcess()`.

### DIFF
--- a/examples/platforms/efr32mg12/sleepy-demo/README.md
+++ b/examples/platforms/efr32mg12/sleepy-demo/README.md
@@ -1,6 +1,6 @@
 # EFR32MG12 Sleepy Demo Example
 
-The EFR32 Sleepy applications demonstrates Sleepy End Device behaviour using the EFR32's low power EM2 mode. The steps below will take you through the process of building and running the demo
+The EFR32 Sleepy applications demonstrates Sleepy End Device behavior using the EFR32's low power EM2 mode. The steps below will take you through the process of building and running the demo
 
 For setting up the build environment refer to [examples/platforms/efr32mg12/README.md](../README.md).
 
@@ -35,7 +35,7 @@ sleepy-demo-ftd started
 sleepy-demo-ftd changed to leader
 ```
 
-When the sleepy-demo-mtd device starts it joins the preconfigured Thread network before disabling Rx-On-Idle to become a Sleepy-End-Device.
+When the sleepy-demo-mtd device starts it joins the pre-configured Thread network before disabling Rx-On-Idle to become a Sleepy-End-Device.
 
 Use the command "child table" in the FTD console and observe the R flag of the child is 0.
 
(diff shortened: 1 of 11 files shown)


romacdon:
@lmnotran there is another PR about to get merged that adds support for efr32mg1.  Once that PR goes in you'll need to make the same changes to the radio.c in the new efr32mg1 directory.  The PR is https://github.com/openthread/openthread/pull/5191.

examples/platforms/efr32mg13/sleepy-demo/README.md
```diff
@@ -1,6 +1,6 @@
 # EFR32MG13 Sleepy Demo Example
 
-The EFR32 Sleepy applications demonstrates Sleepy End Device behaviour using the EFR32's low power EM2 mode. The steps below will take you through the process of building and running the demo
```
romacdon, line 3: Please m ake all the same changes to the efr32mg12 and efr32mg21 versions of this file.

examples/platforms/efr32mg12/sleepy-demo/README.md
```diff
@@ -1,6 +1,6 @@
 # EFR32MG12 Sleepy Demo Example
```
romacdon, line 1: I see there are different changes to the efr32mg12, efr32mg13, and efr32mg21 versions of the file which indicates to me the contents are different.  Why are they different?  Shouldn't they all be the same?

bukepo:
I don't understand how this PR fixes this issue. If there is a pending response message from its parent, the MAC layer should be in kOperationWaitingForData state, once it receives a frame with frame pending cleared, it will put the radio to sleep mode. But it will not trigger another transmission in kOperationWaitingForData state. So I don't understand why the order or process rx or tx matter, although I agree it's better to process rx before tx.

lmnotran:
CC: @schinchani

schinchani:
> I don't understand how this PR fixes this issue. If there is a pending response message from its parent, the MAC layer should be in kOperationWaitingForData state, once it receives a frame with frame pending cleared, it will put the radio to sleep mode. But it will not trigger another transmission in kOperationWaitingForData state. So I don't understand why the order or process rx or tx matter, although I agree it's better to process rx before tx.

schinchani:
> I don't understand how this PR fixes this issue. If there is a pending response message from its parent, the MAC layer should be in kOperationWaitingForData state, once it receives a frame with frame pending cleared, it will put the radio to sleep mode. But it will not trigger another transmission in kOperationWaitingForData state. So I don't understand why the order or process rx or tx matter, although I agree it's better to process rx before tx.

Hi @bukepo. I agree to your comment. We realized that details on this Pull Request did not accurately reflect the issue we were trying to address here. This is change is just an optimization, in the way the Radio driver handles processing of the incoming received packet and the notification sent to the stack.

examples/platforms/efr32mg12/sleepy-demo/README.md
lmnotran: It looks like the `sleepy-demo-mtd` main.c file has been augmented to automatically wake the device every 5 seconds by setting `sButtonPressed` to true when 5000ms has elapsed on some timer. As for the motivation, the only thing I can think of is it makes the testing process contactless so you don't have to press a button. I don't know.

Looks like Marven Gilhespie from the High Wycombe office made the change in Jan 2020 https://github.com/openthread/openthread/commit/73e30d1d7b1454e0d82607381b6b9c1f79ff4cec

schinchani: Hi Bob - I was expecting the behavior to be the same too, but looks like, it is not. Based on the read-me, it appears that the button press functionality. has been incorporated differently for the 12,13 vs the 21 part

jwhui:
Kick CLA.

jwhui:
@lmnotran , it seems that the CLA checker had an issue reporting on this PR. Can you simply comment with "I signed it!" to trigger it again?

lmnotran:
I signed it!

jwhui:
@googlebot

PR #5198, opened 2020-07-06
[logging] fix benign typecast issue in otDump

tarpmaster:
Fixed a benign typecast issue found when using the IAR compiler.


jwhui:
CLA kick.

Issue #5186, opened 2020-07-03
[cli] add a netstat command


bukepo:
**Is your feature request related to a problem? Please describe.** A clear and concise description of what the problem is.
I hope to have a command to list all sockets, which can be helpful for debugging issues.

**Describe the solution you'd like** A clear and concise description of what you want to happen.
 provide a simple implementation of netstat-like command, including bound address, port and status.

**Describe alternatives you've considered** A clear and concise description of any alternative solutions or features you've considered.
simply list all open ports.

**Additional context** Add any other context or screenshots about the feature request here.

bukepo:
@jwhui @abtink what do you think?

jwhui:
I assume this is an OpenThread CLI? Sounds like a good idea to me.

bukepo:
Yes, an OpenThread CLI.

abtink:
Sounds great to me.

PR #5202, opened 2020-07-07
[cli] add `netstat` command

Moandor-y:
This fixes https://github.com/openthread/openthread/issues/5186 by adding `netstat` command to CLI.


src/cli/cli.cpp
```diff
@@ -2062,6 +2065,66 @@ void Interpreter::ProcessNetif(uint8_t aArgsLength, char *aArgs[])
 }
 #endif
 
+#if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
```
jwhui, line 2068: What's the motivation for using `OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE` feature flag?

src/cli/cli.hpp
```diff
@@ -177,8 +177,12 @@ class Interpreter
      * Write an IPv6 address to the CLI console.
      *
      * @param[in]  aAddress  A reference to the IPv6 address.
+     *
+     * @returns The number of bytes placed in the output queue.
+     *
+     * @retval  -1  Driver is broken.
```
jwhui, line 183: minor nit to be consistent:
```suggestion
     * @retval  -1  Driver is broken.
     *
```

src/cli/cli.cpp
```diff
@@ -2062,6 +2065,66 @@ void Interpreter::ProcessNetif(uint8_t aArgsLength, char *aArgs[])
...
+    }
+
+    AppendResult(OT_ERROR_NONE);
+}
+
+int Interpreter::OutputSocketAddress(const otSockAddr &aAddress)
+{
+    int outputLength = 0;
+
+    outputLength += OutputIp6Address(aAddress.mAddress);
```
jwhui, line 2112: Check negative return value here and below?

src/cli/cli.cpp
Moandor-y: `otUdpGetSockets` is defined only if `OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE=1`.

jwhui: Yes, but why? 😄 @bukepo , thoughts?

abtink: I noticed the same in `udp_api.cpp` (though it does not explicitly mention this in the `udp.h` documentation of this API). I guess when it was added it was only used/intended when `PLATFORM_UDP` feature was enabled. 

I also favor that we do not restrict this API to this config and make available in general.

src/cli/cli.cpp
Moandor-y: Done.

src/cli/cli.cpp
bukepo: Abtin helped answer this question, and yes, agree not to restrict this API.

Moandor-y: Removed this flag requirement from API/CLI.

src/core/net/udp6.cpp
```diff
@@ -124,12 +124,19 @@ otError Udp::Socket::Connect(const SockAddr &aSockAddr)
 
     mPeerName = aSockAddr;
 
+    if (!IsBound())
+    {
+        SuccessOrExit(error = Bind(GetSockName()));
+    }
```
abtink, line 130: 👍

PR #5203, opened 2020-07-07
[udp/tcp/icmp] simplify Header parsing, and declare related types as nested

abtink:
This commit contains two main changes in `Ip6` modules: `Udp`, `Tcp`,
and `Icmp`. First, it simplifies the `Header` types defining constant
`enum` for byte offset of different fields in the header (these
replace the static methods and avoid use of `offsetof` macro). Second
change is it declare related types in each module as nested types of
the the main class (place them under the namespace of the main type).
For example `Icmp::Header`, `Udp::Socket` respectively replace
`IcmpHeader`, `UdpSocket` types.

### DIFF
--- a/src/core/net/icmp6.hpp
+++ b/src/core/net/icmp6.hpp
@@ -38,6 +38,7 @@
 
 #include <openthread/icmp6.h>
 
+#include "common/clearable.hpp"
 #include "common/encoding.hpp"
 #include "common/linked_list.hpp"
 #include "common/locator.hpp"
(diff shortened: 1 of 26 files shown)


src/core/net/icmp6.hpp
```diff
@@ -58,193 +59,173 @@ using ot::Encoding::BigEndian::HostSwap16;
...
+         * @returns The ICMPv6 message code.
+         *
+         */
+        Code GetCode(void) const { return static_cast<Code>(mCode); }
+
+        /**
+         * This method sets the ICMPv6 message code.
+         *
+         * @param[in]  aCode  The ICMPv6 message code.
+         */
```
jwhui, line 146:
```suggestion
         *
         */
```

PR #3540, opened 2019-01-31
[core] add Set class

abtink:
This commit adds a new common class `Set` which represents a set of
integers with a given (template) size parameter. The `Set` class is
implemented using a bit-vector mask array. This commit also adds a
unit test for the new `Set` class. The `ChannelMask`, `RouterTable`,
and `Ncp::ChangedPropsSet` classes are modified to use the new
`Set` class.

PR #5206, opened 2020-07-08
[utils] add ChildMask class

simonlingoogle:
This PR adds the `ChildMask` class so that it can be reused by other modules.


simonlingoogle:
> LGTM.
> 
> However I'm thinking that this should be made into something more general. In common dir, the classes and functions seem to be very generic. So I mean it should not have the concept 'Child`. Maybe we can change the name of it.
> 
> I think it's basically a bit mask and very similar to [bitset](http://www.cplusplus.com/reference/bitset/bitset/) in cpp. Shall we rename the class name and api names?

it's actually related to `Child` because the size of `ChildMask` is calculated by:
```c++
kChildMaskBytes = BitVectorBytes(OPENTHREAD_CONFIG_MLE_MAX_CHILDREN)
```

Irving-cl:
> > LGTM.
> > However I'm thinking that this should be made into something more general. In common dir, the classes and functions seem to be very generic. So I mean it should not have the concept 'Child`. Maybe we can change the name of it.
> > I think it's basically a bit mask and very similar to [bitset](http://www.cplusplus.com/reference/bitset/bitset/) in cpp. Shall we rename the class name and api names?
> 
> it's actually related to `Child` because the size of `ChildMask` is calculated by:
> 
> ```c++
> kChildMaskBytes = BitVectorBytes(OPENTHREAD_CONFIG_MLE_MAX_CHILDREN)
> ```

Is this class only used for child mask? If another module wants to use the bitset like

abtink:
I had a detailed template `Set` class (bitset) implementation long ago in https://github.com/openthread/openthread/pull/3540.

Unfortunately, it added a lot of code size overhead on most tool-chains. Investigated different options on this. The challenge was/is even in cases where we mark a simple bit check method as inline, some compliers did not generate as efficient of assembly code as the actual direct inlined code, so decided to close this PR.

I think having smaller "set" like types for specific size and usage (like `Mac::ChannelMask` or `RouterIdSet` (in `mle_types.hpp`) may be better way to go. So adding a `ChildMask` type specific for child sounds reasnable to me.

src/core/common/child_mask.hpp
```diff
@@ -0,0 +1,140 @@
...
+ * @brief
+ *   This module includes definitions for OpenThread Child Mask.
+ *
+ * @{
+ *
+ */
+
+enum
+{
+    kChildMaskBytes = BitVectorBytes(OPENTHREAD_CONFIG_MLE_MAX_CHILDREN)
```
abtink, line 59: How about moving this `enum` to `ChildMask` class and declaring it as `private`?

src/core/common/child_mask.hpp
```diff
@@ -0,0 +1,140 @@
...
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes definitions for a child mask.
```
abtink, line 31: I agree that `common/` folder may not be best place for this header file.

How about moving it under `thread/` (which is where `child_table.hpp` and `topology.hpp` (which defines `Child`) are located)?

src/core/common/child_mask.hpp
```diff
@@ -0,0 +1,140 @@
...
+        mMask[aChildIndex / 8] &= ~(0x80 >> (aChildIndex % 8));
+    }
+
+    /**
+     * This method returns if any Child mask is set.
+     *
+     * @returns Whether any Child mask is set.
+     *
+     */
+    bool GetAny() const
```
abtink, line 113: How about `HasAny()`?
Also adding `void` when no param, i.e. `HasAny(void)`?

simonlingoogle:
> Also where (other than `Message` metadata) do you have in mind for using the `ChildMask` type?

We are going to use `ChildMask` for the the 1.2 MLR feature in which a parent uses a `ChildMask` to mask all Children that subscribed to a multicast address.

src/core/common/child_mask.hpp
simonlingoogle: `ChildMask` is currently used by `common/message.hpp`. 
I don't feel right to include `thread/child_mask.hpp` in `common/message.hpp`. 
I think `thread` can depend on `common`, but `common` **should not** depend on `thread`.

src/core/common/child_mask.hpp
simonlingoogle: moved

src/core/common/child_mask.hpp
simonlingoogle: renamed

src/core/common/child_mask.hpp
abtink: Not sure if we need to worry about dependency between core folders. The folders are just a way of organizing source files - and we are not building each folder as a different libs (they are all tied and built together).

We already include `"mac/mac_types.hpp"` and `"thread/link_quality.hpp"` in `message.hpp`.

And actually `common/instance.hpp` includes almost every other header (directly or through other ones).

simonlingoogle: good point. I will revise.

simonlingoogle: moved

src/core/thread/child_mask.hpp
```diff
@@ -0,0 +1,140 @@
...
+    void Clear(uint16_t aChildIndex)
+    {
+        OT_ASSERT(aChildIndex < OPENTHREAD_CONFIG_MLE_MAX_CHILDREN);
+        mMask[aChildIndex / 8] &= ~(0x80 >> (aChildIndex % 8));
+    }
+
+    /**
+     * This method returns if any Child mask is set.
+     *
+     * @returns Whether any Child mask is set.
```
jwhui, line 105:
```suggestion
     * @retval TRUE   If any Child index is set.
     * @retval FALSE  If all Child indexes are clear.
```

src/core/thread/child_mask.hpp
```diff
@@ -0,0 +1,140 @@
...
+ */
+class ChildMask
+{
+public:
+    /**
+     * This method returns if a given Child index is masked.
+     *
+     * @param aChildIndex  The Child index.
+     *
+     * @returns Whether the given Child index is masked.
```
jwhui, line 69:
```suggestion
     * @retval TRUE   If the given Child index is set.
     * @retval FALSE  If the given Child index is clear.
```

src/core/thread/child_mask.hpp
```diff
@@ -0,0 +1,140 @@
...
+    bool Get(uint16_t aChildIndex) const
+    {
+        OT_ASSERT(aChildIndex < OPENTHREAD_CONFIG_MLE_MAX_CHILDREN);
+        return (mMask[aChildIndex / 8] & (0x80 >> (aChildIndex % 8))) != 0;
+    }
+
+    /**
+     * This method sets the mask of a given Child index.
+     *
+     * @param aChildIndex  The Child index.
```
jwhui, line 81:
```suggestion
     * @param[in] aChildIndex  The Child index.
```

src/core/thread/child_mask.hpp
```diff
@@ -0,0 +1,140 @@
...
+    void Set(uint16_t aChildIndex)
+    {
+        OT_ASSERT(aChildIndex < OPENTHREAD_CONFIG_MLE_MAX_CHILDREN);
+        mMask[aChildIndex / 8] |= 0x80 >> (aChildIndex % 8);
+    }
+
+    /**
+     * This method clears the mask of a given Child index.
+     *
+     * @param aChildIndex  The Child index.
```
jwhui, line 93:
```suggestion
     * @param[in] aChildIndex  The Child index.
```

PR #5212, opened 2020-07-09
[string] add String::AppendHexBytes()

abtink:
This method appends a given array of bytes in hex representation
(`%02x` style) to a `String` object. This method helps simplify the
`ToString()` methods in simpler types (e.g. `Mac::ExtAddress`).


jwhui:
@googlebot

PR #5213, opened 2020-07-09
[ip6] define NetworkPrefix type

abtink:
This commit adds `Ip6::NetworkPrefix` mirroring `otIp6NetworkPrefix`
which represents the Network Prefix (most significant 64 bits) of an
IPv6 address. Methods in `Ip6::Address` are updated to use the new
type. The `otMeshLocalPrefix` and `Mle::MeshLocalPrefix` are also
updated to use the `NetworkPrefix` type.

### DIFF
--- a/include/openthread/dataset.h
+++ b/include/openthread/dataset.h
@@ -99,23 +99,13 @@ struct otExtendedPanId
  */
 typedef struct otExtendedPanId otExtendedPanId;
 
-#define OT_MESH_LOCAL_PREFIX_SIZE 8 ///< Size of the Mesh Local Prefix (bytes)
+#define OT_MESH_LOCAL_PREFIX_SIZE OT_IP6_PREFIX_SIZE ///< Size of the Mesh Local Prefix (bytes)
 
 /**
  * This structure represents a Mesh Local Prefix.
  *
  */
-OT_TOOL_PACKED_BEGIN
-struct otMeshLocalPrefix
-{
-    uint8_t m8[OT_MESH_LOCAL_PREFIX_SIZE]; ///< Byte values
-} OT_TOOL_PACKED_END;
-
-/**
- * This structure represents a Mesh Local Prefix.
- *
- */
-typedef struct otMeshLocalPrefix otMeshLocalPrefix;
+typedef otIp6NetworkPrefix otMeshLocalPrefix;
 
 #define OT_PSKC_MAX_SIZE 16 ///< Maximum size of the PSKc (bytes)
 
--- a/src/core/net/ip6_address.cpp
+++ b/src/core/net/ip6_address.cpp
@@ -47,6 +47,16 @@ namespace ot {
 namespace Ip6 {
 
 //---------------------------------------------------------------------------------------------------------------------
+// NetworkPrefix methods
+
+otError NetworkPrefix::GenerateRandomUla(void)
+{
+    m8[0] = 0xfd;
+
+    return Random::Crypto::FillBuffer(&m8[1], kSize - 1);
+}
+
+//---------------------------------------------------------------------------------------------------------------------
 // InterfaceIdentifier methods
 
 bool InterfaceIdentifier::IsUnspecified(void) const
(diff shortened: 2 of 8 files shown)


jwhui:
@googlebot

src/core/net/ip6_address.cpp
```diff
@@ -249,6 +249,11 @@ void Address::SetToRealmLocalAllMplForwarders(void)
     *this = GetRealmLocalAllMplForwarders();
 }
 
+void Address::SetPrefix(const NetworkPrefix &aNetworkPrefix)
+{
+    mFields.mComponents.mNetworkPrefix = aNetworkPrefix;
+}
+
 void Address::SetPrefix(const uint8_t *aPrefix, uint8_t aPrefixLength)
```
jwhui, line 257: Just testing our method name policy here: to overload or not? Thoughts?

include/openthread/dataset.h
```diff
@@ -99,23 +99,13 @@ struct otExtendedPanId
...
-{
-    uint8_t m8[OT_MESH_LOCAL_PREFIX_SIZE]; ///< Byte values
-} OT_TOOL_PACKED_END;
-
-/**
- * This structure represents a Mesh Local Prefix.
- *
- */
-typedef struct otMeshLocalPrefix otMeshLocalPrefix;
+typedef otIp6NetworkPrefix otMeshLocalPrefix;
```
jwhui, line 108: Should we replace `otMeshLocalPrefix` with `otIp6NetworkPrefix`?

src/core/net/ip6_address.cpp
abtink: I initially had  this as `SetNetworkPrefix()` but then decided in favor of overloading the `SetPrefix()` method.
- I think the overloads of `SetPrefix()` are distinct (and easily distinguishable) on the caller side, one requires a single reference argument, and the other one has two arguments, a buffer pointer and a length.
- Also in both cases, we don't do any conversion and directly copy/use the provided input value.

In future we may add a new overload of the form `SetPrefix(const otIp6Prefix &)`. I feel this will still be ok and distinct (and again value would be copied as is - no conversion).

Thoughts?

include/openthread/dataset.h
abtink: I think it's a good idea to remove the public  `otMeshLocalPrefix` type definition itself (and use `otIp6NetworkPrefix` in its place directly). 

What do you think about keeping or removing `Mle::MeshLocalPrefix` class?
- I favor keeping it.
- With this PR, it inherits from `Ip6::NetworkPrefix` and adds one method `SetFromExtendedPanId()`.
- We can consider moving this method to `NetworkPrefix` but I feel it's  MLE specific.

src/core/net/ip6_address.cpp
jwhui: I'm personally fine with the method overloading here. I'm just trying to reconcile this with what you stated in https://github.com/openthread/openthread/pull/5194#discussion_r450412604 :

> - For cases where the value is directly used (no change/conversion), either way (use explicit method name or overload) is fine.
>   - I personally prefer explicit name in this case as well (feel it is more readable and avoid any incorrect use), but we can leave it to developer.

Just want to ensure consistency.

include/openthread/dataset.h
jwhui: I'm fine with keeping `Mle::MeshLocalPrefix` for the reasons you mentioned.

src/core/net/ip6_address.cpp
abtink: Very good and fair point indeed. 😄 

Thinking more about this (and actually somewhat changing my mind on this), here is what I would suggest.
- For cases where there is a conversion/parsing happening we prefer/recommend using explicit method name vs overloading (to make it super clear that it being converted).
- For cases where there is no conversion (value is used directly), we prefer/recommend using overloading over explicit names.
- In both case, it's a recommendation (not a requirement) so we keep it flexible for handling special situations.

Thoughts?

jwhui: Sounds good to me!

include/openthread/dataset.h
abtink: Updated in the new push. 

Also added a `GenerateRandomUla()` method to generate a ULA based random prefix of `fdxx:xxxx:xxxx:xxxx:` pattern.

abtink: Per offline chat with @jwhui , decided to keep the public `otMeshLocalPrefix` definition (we define it as a `typedef` of the `otIp6NetworkPrefix`).

PR #5221, opened 2020-07-10
[bbr] validate BBR configuration

simonlingoogle:
This PR makes sure BBR configuration is valid according to Thread Spec. 5.21.3.3.

### DIFF
--- a/src/core/backbone_router/local.cpp
+++ b/src/core/backbone_router/local.cpp
@@ -143,8 +143,7 @@ otError Local::SetConfig(const BackboneRouterConfig &aConfig)
     bool    update = false;
 
     // Validate configuration according to Thread 1.2.1 Specification 5.21.3.3:
-    // " The Reregistration Delay in seconds MUST be lower than (0.5 * MLR Timeout). 
-    // It MUST be at least 1."
+    // "The Reregistration Delay in seconds MUST be lower than (0.5 * MLR Timeout). It MUST be at least 1."
     VerifyOrExit(aConfig.mReregistrationDelay >= 1, error = OT_ERROR_INVALID_ARGS);
     VerifyOrExit(aConfig.mReregistrationDelay * 2 < aConfig.mMlrTimeout, error = OT_ERROR_INVALID_ARGS);
 
--- a/src/core/backbone_router/local.cpp
+++ b/src/core/backbone_router/local.cpp
@@ -142,7 +142,9 @@ otError Local::SetConfig(const BackboneRouterConfig &aConfig)
     otError error  = OT_ERROR_NONE;
     bool    update = false;
 
-    // Validate configuration according to Thread Spec. 5.21.3.3
+    // Validate configuration according to Thread 1.2.1 Specification 5.21.3.3:
+    // " The Reregistration Delay in seconds MUST be lower than (0.5 * MLR Timeout). 
+    // It MUST be at least 1."
     VerifyOrExit(aConfig.mReregistrationDelay >= 1, error = OT_ERROR_INVALID_ARGS);
     VerifyOrExit(aConfig.mReregistrationDelay * 2 < aConfig.mMlrTimeout, error = OT_ERROR_INVALID_ARGS);
 
--- a/src/core/backbone_router/local.cpp
+++ b/src/core/backbone_router/local.cpp
@@ -145,6 +145,8 @@ otError Local::SetConfig(const BackboneRouterConfig &aConfig)
     // Validate configuration according to Thread 1.2.1 Specification 5.21.3.3:
     // "The Reregistration Delay in seconds MUST be lower than (0.5 * MLR Timeout). It MUST be at least 1."
     VerifyOrExit(aConfig.mReregistrationDelay >= 1, error = OT_ERROR_INVALID_ARGS);
+    static_assert(sizeof(aConfig.mReregistrationDelay) < sizeof(aConfig.mMlrTimeout),
+                  "the calculation below might overflow");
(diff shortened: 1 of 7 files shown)


src/core/backbone_router/local.hpp
```diff
@@ -104,8 +104,11 @@ class Local : public InstanceLocator
      *
      * @param[in]  aConfig  The configuration to set.
      *
+     * @retval OT_ERROR_NONE          Successfully updated configuration.
+     * @retval OT_ERROR_INVALID_ARGS  If the specified configuration is invalid.
```
jwhui, line 108:
```suggestion
     * @retval OT_ERROR_INVALID_ARGS  The configuration in @p aConfig is invalid.
```

src/core/backbone_router/local.cpp
```diff
@@ -137,9 +137,15 @@ void Local::GetConfig(BackboneRouterConfig &aConfig) const
...
+otError Local::SetConfig(const BackboneRouterConfig &aConfig)
 {
-    bool update = false;
+    otError error  = OT_ERROR_NONE;
+    bool    update = false;
+
+    // Validate configuration according to Thread 1.2.1 Specification 5.21.3.3:
+    // "The Reregistration Delay in seconds MUST be lower than (0.5 * MLR Timeout). It MUST be at least 1."
+    VerifyOrExit(aConfig.mReregistrationDelay >= 1, error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(aConfig.mReregistrationDelay * 2 < aConfig.mMlrTimeout, error = OT_ERROR_INVALID_ARGS);
```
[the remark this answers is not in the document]
simonlingoogle: Great catch!

However with more analysis, I found that the original condition is actually correct because `mReregistrationDelay` is of type `uint16_t`. In the expression, `mReregistrationDelay` will be promoted to `uint32_t` so it will never overflow. 

That being said, this expression is a potential bug if one day we changed `mReregistrationDelay` to be of type `uint32_t`. In such case, the suggested conditions are however incorrect. I tend to keep the code unchanged since it's already correct.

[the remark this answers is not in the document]
simonlingoogle: I found it interesting that it's actually difficult to write such a simple expression correctly. 

I think this expression is correct no matter `mReregistrationDelay` is of type `uint16_t` or `uint32_t`. 
```c++
    VerifyOrExit(static_cast<uint32_t>(aConfig.mReregistrationDelay) <= std::numeric_limits<uint32_t>::max() / 2 &&
                     aConfig.mReregistrationDelay < (aConfig.mMlrTimeout + 1) / 2,
                 error = OT_ERROR_INVALID_ARGS);
```

Thoughts? @jwhui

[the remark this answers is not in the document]
jwhui: Rather than complicating run-time check, how about we use `static_assert` to ensure the types are appropriate for the given logic?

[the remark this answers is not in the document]
simonlingoogle: I have added the `static_assert` check.

simonlingoogle:
Added one more check that **MLR timeout >= MLR_TIMEOUT_MIN (300)** required by Thread Spec.
@librasungirl @jwhui

src/core/thread/mle_types.hpp
```diff
@@ -258,9 +258,13 @@ enum
 {
     kRegistrationDelayDefault         = 1200, //< In seconds.
     kMlrTimeoutDefault                = 3600, //< In seconds.
+    kMlrTimeoutMin                    = 300,  //< In seconds.
     kBackboneRouterRegistrationJitter = 5,    //< In seconds.
 };
 
+static_assert(kMlrTimeoutDefault >= kMlrTimeoutMin,
```
librasungirl, line 265: Generally user should not change this internal header file which defines  `kMlrTimeoutDefault` and `kMlrTimeoutMin`,  is it necessary for this `static_assert`?

src/core/backbone_router/local.cpp
```diff
@@ -145,6 +145,7 @@ otError Local::SetConfig(const BackboneRouterConfig &aConfig)
     // Validate configuration according to Thread 1.2.1 Specification 5.21.3.3:
     // "The Reregistration Delay in seconds MUST be lower than (0.5 * MLR Timeout). It MUST be at least 1."
     VerifyOrExit(aConfig.mReregistrationDelay >= 1, error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(aConfig.mMlrTimeout >= Mle::kMlrTimeoutMin, error = OT_ERROR_INVALID_ARGS);
```
librasungirl, line 148: nit: how about adjusting the order a bit to put `mReregistrationDelay` relative checks together and match the comment?

src/core/thread/mle_types.hpp
simonlingoogle: I agree we will hardly change these values.

src/core/backbone_router/local.cpp
simonlingoogle: The current intention is to check two individual values first, and then check the relationship between them. Does it make sense to you?

simonlingoogle: revised according

src/core/thread/mle_types.hpp
simonlingoogle: I tends to keep this `static_assert` because it satisfies logical completeness to guarantee that `MLR timeout >= 300` is always true for OT devices.

Issue #5218, opened 2020-07-10
[ci] "external-commissioner" test failures - dependency between ot-br and openthread repo PRs 


abtink:
Noticed that in PR https://github.com/openthread/openthread/pull/5213 that the "external-commissioner" CI test (under simulation workflow and merged recently from https://github.com/openthread/openthread/pull/5187) is failing [[link]](https://github.com/openthread/openthread/pull/5213/checks?check_run_id=855929502).

This seems to be due to the test trying to build ot-br repo as well (with the new changes from OT PR pulled in):
```
FAILED: src/dbus/server/CMakeFiles/otbr-dbus-server.dir/dbus_thread_object.cpp.o 
../../src/dbus/server/dbus_thread_object.cpp: In member function ‘otError otbr::DBus::DBusThreadObject::SetMeshLocalPrefixHandler(DBusMessageIter&)’:
2126
../../src/dbus/server/dbus_thread_object.cpp:441:5: error: ‘otMeshLocalPrefix’ was not declared in this scope
2127
     otMeshLocalPrefix                         prefix;
2128
     ^~~~~~~~~~~~~~~~~
2129
compilation terminated due to -Wfatal-errors.
```

Note that this PR (per comments) is actually changing some of the public OT definitions (e.g., replacing`otMeshLocalPrefix` definition with `otIp6NetworkPrefix`). 

Looks like `ot-br` build is pulling in OT from PR (rather then the version of OT it is snapped to). This is adding dependency between repos (which we should avoid).

Digging a bit more into it, I think this recent merged PR https://github.com/openthread/ot-commissioner/pull/125 may be the cause of this.

@Moandor-y can you please help look into this? Thanks.

cc/ @bukepo

Moandor-y:
Yes. It will fail if there are any incompatible changes. This made the two repo depend on each other, and as discussed with @bukepo , @simonlingoogle and @wgtdkp , the only way is to change both repos at the same time. Since GitHub doesn't support PR dependencies, it has to be done manually. This does not only happen with `ot-br`, both `ot-commissioner` and `ot-br` are pulling `openthread` from the PR. Each time we encounter this we have to modify multiple repos and merge the PRs together (with some checks in failed state).
Or is there any better approach?

abtink:
I would highly advice against the approach to allow PRs to be merged with failed CI tests. 

I think we should avoid creating any such dependencies in the CI of openthread. This would block development of OT.

[...]

PR #5222, opened 2020-07-10
[github-actions] allow external commissioner tests to fail

Moandor-y:
This allows the CI test `external-commissioner` to fail. It removes the dependencies of `ot-commissioner` and `ot-br-posix` so that a breaking change can be merged. Fixes https://github.com/openthread/openthread/issues/5218.

### DIFF
--- a/.github/workflows/simulation.yml
+++ b/.github/workflows/simulation.yml
@@ -204,6 +204,7 @@ jobs:
       uses: codecov/codecov-action@v1
 
   external-commissioner:
+    continue-on-error: true
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/simulation.yml
+++ b/.github/workflows/simulation.yml
@@ -209,11 +209,9 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Bootstrap
-      continue-on-error: true
       run: |
         sudo apt-get --no-install-recommends install -y ninja-build
         git clone https://github.com/openthread/ot-commissioner.git /tmp/ot-commissioner --depth 1 --branch master
-        exit 1
     - name: Build
       run: |
         cd /tmp/ot-commissioner
--- a/.github/workflows/simulation.yml
+++ b/.github/workflows/simulation.yml
@@ -211,9 +211,9 @@ jobs:
     - name: Bootstrap
       continue-on-error: true
       run: |
-        exit 1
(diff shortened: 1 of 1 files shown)


.github/workflows/simulation.yml
```diff
@@ -204,6 +204,7 @@ jobs:
       uses: codecov/codecov-action@v1
 
   external-commissioner:
+    continue-on-error: true
```
jwhui, line 207: I believe GitHub Actions will always mark this test with a green check. The only way to see if this test fails is by digging into the logs. Is there something more visible?

simonlingoogle: Just some random thoughts: if we have a comment bot, we can use it to comment in the PR and @ related people.

Moandor-y: Failed tests will be shown like this: https://github.com/Moandor-y/openthread/actions/runs/166971709. The workflow succeeded but it still shows the job failure. So I guess in the PR UI it will show something like "58 successful, 1 failing, 1 expected, and 2 neutral checks", we don't have to dig into the logs.

Moandor-y: Tried 2 options:
1. Add `continue-on-error` to `jobs`: This looks the same as before. The check fails, and has a red cross icon in the PR.
1. Add `continue-on-error` to `steps`: This makes all things green. The only way to see failures is by logs.
I think it's better to use option 2 with something making it more visible.

jwhui: @Moandor-y , do you have a proposal for "something making it more visible"?

Moandor-y: I think commenting on PR is a good idea. I'll find a way to do this.

simonlingoogle: I investigated on commenting on PR. Basically we need to build a simple GitHub App similar to `openthread-size-report` for doing this.

bukepo: I think we can remove dependency on ot-br-posix by using openthread-daemon. https://github.com/openthread/openthread/tree/master/src/posix

wgtdkp: > I think we can remove dependency on ot-br-posix by using openthread-daemon.

Agree.

Moandor-y: Created https://github.com/openthread/ot-commissioner/pull/127 to remove the dependency.

jwhui:
Closing in favor of https://github.com/openthread/ot-commissioner/pull/127

PR #5228, opened 2020-07-13
[mle] add `mHasRestored` flag

librasungirl:
-- edited --
This PR introduces the `mHasRestored ` flag and a common `HasRestored()` method as devices may decide to introduce some jitter/randomization to avoid the burst of traffic in cases a bunch of devices nearby resets/reboots together. One typical usage defined in Thread 1.2 Specification 5.23.4.1 is that
> In the following cases, a random time delay chosen in the interval [0, Reregistration Delay], a random time delay chosen in the interval [0, Reregistration Delay] MUST be applied before sending the DUA.req
> - The Device reregisters its DUA due to a Device reboot/restart.

### DIFF
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -86,6 +86,7 @@ Mle::Mle(Instance &aInstance)
     , mDataRequestAttempts(0)
     , mDataRequestState(kDataRequestNone)
     , mAddressRegistrationMode(kAppendAllAddresses)
+    , mHasRestored(false)
     , mParentLinkMargin(0)
     , mParentIsSingleton(false)
     , mReceivedResponseFromParent(false)
@@ -438,6 +439,9 @@ otError Mle::Restore(void)
         Get<MleRouter>().RestoreChildren();
     }
 #endif
+
+    // Sucessfully restored the network information from non-volatile settings after boot.
+    mHasRestored = true;
 
 exit:
     return error;
--- a/src/core/thread/mle.hpp
+++ b/src/core/thread/mle.hpp
@@ -915,6 +915,16 @@ public:
      */
     otError GetLocatorAddress(Ip6::Address &aAddress, uint16_t aLocator) const;
 
+    /**
+     * This method indicates whether or not the device has restored the network information from
+     * non-volatile settings after boot.
+     *
+     * @retval true  Sucessfully restored the network information.
+     * @retval false No valid network information was found.
+     *
+     */
+    bool HasRestored(void) const { return mHasRestored; }
+
 protected:
     /**
      * States during attach (when searching for a parent).
@@ -1734,6 +1744,7 @@ private:
 
     AddressRegistrationMode mAddressRegistrationMode;
 
+    bool       mHasRestored;
     uint8_t    mParentLinkMargin;
     bool       mParentIsSingleton;
     bool       mReceivedResponseFromParent;


src/core/thread/mle.cpp
```diff
@@ -439,6 +440,9 @@ otError Mle::Restore(void)
     }
 #endif
 
+    // Successful restore indicates the device was attached to some Thread Network.
+    mIsFreshBoot = false;
```
jwhui, line 444: I'm not clear on the purpose of this flag. Which of the following cases is this flag intended to capture?

1. Device has been factory reset, but has not yet attached to any Thread network since factory reset.

1. Device has been rebooted, but has not yet attached to any Thread network since last reboot.

librasungirl: 2. the second scenario.

jwhui: If case 2, should we be setting `mIsFreshBoot` to `false` here? Shouldn't we wait until the role has changed to Child, Router, or Leader?

librasungirl: Please refer to above first conversion for the background. Not sure but is the variable name or the comment bringing the confusion?

simonlingoogle:
> This PR introduces the `mIsFreshBoot` flag and a common `IsFreshBoot()` method as devices may decide to introduce some jitter/randomization to avoid the burst of traffic in cases a bunch of devices nearby resets/reboots together. 

I agree that it follows the Thread Spec to behave differently by checking if the device is a freshly booted.
But I doubt if such logic is meaningful in real products. 
For example, if devices registers DUA without delay when it's a fresh boot and with delay < `REREG_DELAY` when it's not a fresh boot. Devices in real life will **almost always** delay registering DUA because the `refresh boot` logic only runs when devices are booted for the first time or factory-reseted. 
I would argue that optimizations for such a rare case can hardly be of much value. So it would be equal in functionalities and simpler in code to use the same logic for all boots. 

Another potential issue of logics that check `fresh boots` is that we will be running a different set of code for development and products. We have so much more fresh boots in development and CI tests than in products' lifetime, that we are always testing the code that is not executed in real-life products, but not the code that actually runs.

src/core/thread/mle.cpp
jwhui: I think it would help if there was more detail on what this flag is intended to represent and how it is used. This is the only place where I see `mIsFreshBoot` set to `false`. Are there other places where `mIsFreshBoot` should be set to `false`?

librasungirl: Sorry for the confusion, to be accurate, the flag was designed to represent `Device joined one Thread Network, it had network info stored, and device is rebooted/restarted now, it restores the network info from flash and tries to reattach` and `leaving the role check to the caller who cares for fresh boot or not`.

For the DUA reregistration usage example - `The Device reregisters its DUA due to a Device reboot/restart in a random time delay chosen in the interval [0, Reregistration Delay]` , it should not re-register if it has not become attached and got Thread Network Data with Backbone Router Service.

I had considered to check whether it is `reboot` in `DuaManager::Restore()`, however there might be no DadCounter stored if  the DUA Interface Identifier is manually specified by user. And another consideration to put it here is that in practical scenarios, there might be some application level transactions which may cause burst of traffic in short time once multiple devices reboot together( and I am thinking whether or not providing one otAPI so that developer may decide to randomize the time in their application accordingly...)

Currently it is designed to reset this flag only accord to whether or not there is valid network info stored when bring up, there should be no other places to reset this value in my mind so far.

Thoughts?

librasungirl:
@simonlingoogle I also had similar concerns when reading the spec, here are some thoughts with the above DUA reregistration scenario as example, just fyi:
Generally the new thread devices would be commissioned/joined one by one, that is, there should be no burst of traffic due to DUA registration transaction or other upper layer application.
Things becomes different if a bunch of devices leave and come back at the same time (power off/on), there might be chance for the burst of traffic. For DUA feature, the possible situations are
1) PBBR is alive all the time
2) PBBR also has experienced power off/on, same as the Thread device

For 1), the introduce of the delay has some advantage in scattering the re-registration traffic without impacting the ND-Proxy functionality of PBBR;  However I would agree the proxy performance may depend on the value of `mReregistraionDelay` of Primary BBR's Dataset due to the introduce of the delay.

Comments? @jwhui @simonlingoogle

simonlingoogle:
My opinion is that devices can always delay for 0~mReregistraionDelay seconds before registering DUA when the device joins the network. 
Allowing a newly commissioned device to register DUA immediately can be beneficial if we are using some device/service discovery mechanism that requires DUA readability.
But I don't think it matters that much since the device will need to delay registering DUA ever since commissioned. The max delay from booting device to having DUA readability is always `mReregistraionDelay`. 
BTW, I would expect `mReregistraionDelay` to be set to a relatively small value (e.g. 5s ~ 60s) for home networks.

librasungirl:
Regarding the choice of `ReregistrationDelay`, 1.2 Spec 5.21.3.3 says

```
The Reregistration Delay value is vendor-specific and can be chosen by the BBR either as a
fixed (configured) value or a dynamic value based on detected network size or
circumstances. If dynamic, the rate of change should be low i.e. rather change every few
hours/days, but not every few minutes. The Reregistration Delay in seconds MUST be lower
than (0.5 * MLR Timeout). It MUST be at least 1.
```

Agree that it could be configured as  5s ~ 60s for home network, as it is `vendor-specific`, however DUA feature is designed
for beyond home,  the implementation should be more general, other than with such assumption.

simonlingoogle:
> Agree that it could be configured as 5s ~ 60s for home network, as it is `vendor-specific`, however DUA feature is designed
> for beyond home, the implementation should be more general, other than with such assumption.

My point had always been that if a device would need to delay for `t < DUA reregistration delay`  in most of time, except rare cases like it's just commissioned, being able to eliminate the delay for the rare case seems not that useful to me. 

For example if a vendor sets reregistration delay to 1 hour, then the vendor application should be fine with no available DUA on devices in 1 hour.

That being said. If we are to check fresh boot, what about checking if `mDadCounter == 0` for DUA registration? There seems no obvious difference between `mDadCounter == 0` and `IsFreshBoot()`.

jwhui:
I wonder if there's some confusion here:

> My point had always been that if a device would need to delay for `t < DUA reregistration delay` in most of time, except rare cases like it's just commissioned, being able to eliminate the delay for the rare case seems not that useful to me.

I thought the purpose of this PR is to address the use case of any reboot, not just the first boot after being commissioned.

> For example if a vendor sets reregistration delay to 1 hour, then the vendor application should be fine with no available DUA on devices in 1 hour.

I don't think that's true. I think the vendor expects that the device will have its DUA available relatively soon after attaching to a network.

> That being said. If we are to check fresh boot, what about checking if `mDadCounter == 0` for DUA registration? There seems no obvious difference between `mDadCounter == 0` and `IsFreshBoot()`.

If we are talking about any reboot (not just first boot after reattach), then I'm not sure `mDadCounter == 0` and `IsFreshBoot()` are checking the same thing.

src/core/thread/mle.cpp
jwhui: How about renaming the flag to `mIsRestoring`?

`Mle::Restore()` is also called as part of `otInstanceInit()` and that may be called for reasons not related to reboot.

simonlingoogle:
> I don't think that's true. I think the vendor expects that the device will have its DUA available relatively soon after attaching to a network.

Thread 1.2.1 Spec 5.23.4.1 states: 
```
In the following cases, a random time delay chosen in the interval [0, Reregistration Delay]
MUST be applied before sending the DUA.req:

* The Device reregisters its DUA due to a Device reboot/restart.
```
In such case, does the device that rebooted have DUA? My understanding is that it does not. So the device won't have DUA reachability for the interval [0, Reregistration Delay], which is vendor specific (can be hours). 
Thoughts? @jwhui

src/core/thread/mle.cpp
abtink: I also think `IsRestoring` or may be even `HasRestored` are name for this... this is to indicate that device restored network ifno (dataset, etc) from non-volatile settings after OT initialization (device boot).

librasungirl: I would prefer `HasRestored` a bit as the suffix of `ing` looks like something ongoing.
Will update.

src/core/thread/mle.hpp
```diff
@@ -1588,6 +1588,16 @@ class Mle : public InstanceLocator, public Notifier::Receiver
...
 
+    /**
+     * This method indicates whether or not the device has restored the network information from
+     * non-volatile settings after boot.
+     *
+     * @retval true  Sucessfully restored the network information.
+     * @retval false No valid network information was found.
+     *
+     */
+    bool HasRestored(void) { return mHasRestored; }
```
jwhui, line 1599:
```suggestion
    bool HasRestored(void) const { return mHasRestored; }
```

Issue #5162, opened 2020-06-29
[udp6] limitation of otUdpSocket down-casted to Ip6::UdpSocket (being an InstanceLocator)


abtink:
This is not a critical issue but documenting here so we can address.

Here are the details:
- We declare `otUdpSocket` as public OT definition and provide a bunch of public `otUdp{Open/Close/Bind/Connect/Send}` APIs to user of OT.
- In the core, we define `Ip6::UdpSocket` classs which inherits from `otUdpSocket` but it is also declared as a sub-class of `InstanceLocator`. 
- In `udp_api.cpp` we `static_cast` a given `otUdpSocket` to a `Ip6::UdpSocket`: 
`Ip6::UdpSocket &socket = *static_cast<Ip6::UdpSocket *>(aSocket);`
- Or In `otUdpOpen()` we do
```Ip6::UdpSocket &socket   = *new (aSocket) Ip6::UdpSocket(instance.Get<Ip6::Udp>());```

This currently works since `InstanceLocator` is an empty base class (for single instance case) but breaks in multi-instance case and in general not a reliable/good approach.

The current solution I have in mind for this is:
- Move the current `UdpSocket` methods (like `SendTo()`) to the `Udp` class itself with socket being passed as one of the parameters.
- Also we can consider distinguishing between public and internal UDP sockets (defining two different types/mechanism for how to use sockets, similar to how `Notifier` uses different ways for core-internal vs. public notifications).

Thoughts?

PR #5231, opened 2020-07-14
[udp6] update UDP socket definitions

abtink:
This commit updates the public OT `otUdp{}` (like `otUdpConnect` or `otUpdSend()`) functions to requires a pointer the OpenThread instance to be passed as their first parameter (this harmonizes the definitions with other public APIs and removes the need for `otUdpSocket` to track the instance).

The core implementation in `udp6` module is updated to define `SocketHandle` as an internal mirror of `otUdpSocket` structure. The socket related methods `Open`(), `Bind()`, `Connect()`, etc., are moved to `Udp` class itself (with a `SocketHandle` passed as a parameter). For core internal use `Socket` class is defined as a sub-class of `SocketHandle` and `InstanceLocator` providing same helper methods. The separation between `SocketHandle` and `Socket` type addresses the problem that can be caused by the treating/casting publicly provided `otUdpSockt` objects as `InstanceLocator`.

--------

This should help address issue https://github.com/openthread/openthread/issues/5162.


src/core/net/udp6.hpp
```diff
@@ -348,20 +358,66 @@ class Udp : public InstanceLocator
...
     /**
-     * This method adds a UDP socket.
+     * This method opens a UDP socket.
+     *
+     * @param[in]  aSocket   A reference to the socket.
+     * @param[in]  aHandler  A pointer to a function that is called when receiving UDP messages.
+     * @param[in]  aContext  A pointer to arbitrary context information.
+     *
+     * @retval OT_ERROR_NONE     Successfully opened the socket.
+     * @retval OT_ERROR_ALREADY  The socket is already open.
```
jwhui, line 368: Replace with `OT_ERROR_FAILED`?

src/core/net/udp6.hpp
```diff
@@ -348,20 +358,66 @@ class Udp : public InstanceLocator
...
-     * This method removes a UDP socket.
+     * This method sends a UDP message using a socket.
      *
-     * @param[in]  aSocket  A reference to the UDP socket.
+     * @param[in]  aSocket       A reference to the socket.
+     * @param[in]  aMessage      The message to send.
+     * @param[in]  aMessageInfo  The message info associated with @p aMessage.
+     *
+     * @retval OT_ERROR_NONE          Successfully sent the UDP message.
+     * @retval OT_ERROR_INVALID_ARGS  If no peer is specified in @p aMessageInfo or by connect().
```
jwhui, line 416:
```suggestion
     * @retval OT_ERROR_INVALID_ARGS  If no peer is specified in @p aMessageInfo or by Connect().
```

include/openthread/udp.h
```diff
@@ -157,68 +157,51 @@ otMessage *otUdpNewMessage(otInstance *aInstance, const otMessageSettings *aSett
...
+ * @retval OT_ERROR_FAILED Failed to connect UDP socket.
  *
  */
-otError otUdpConnect(otUdpSocket *aSocket, otSockAddr *aSockName);
+otError otUdpConnect(otInstance *aInstance, otUdpSocket *aSocket, const otSockAddr *aSockName);
 
 /**
  * Send a UDP/IPv6 message.
  *
+ * @param[in]  aInstance  A pointer to an OpenThread instance.
```
jwhui, line 204:
```suggestion
 * @param[in]  aInstance     A pointer to an OpenThread instance.
```

romacdon:
@abtink and @jwhui - this commit breaks the efr32 builds.  I'll have someone fix up the efr32 builds but whats the normal policy on making an incompatible public API change like this?

abtink:
@romacdon we track any public OT API changes (even a change in a documentation) with an API version number increase (this is away for users of OT APIs to be aware of such changes) -> https://github.com/openthread/openthread/blob/b88f4fccba5bd00c4f49bc6bc0d77ebc3d23a2ad/include/openthread/instance.h#L48-L56

We also have a CI test/check which ensures any PRs that change any public OT API source files, does indeed update the API version number as well.

romacdon:
Thanks @abtink - I didn't know about the rules around OPENTHREAD_API_VERSION. Good to know thats in place.  Was there any thought about deprecating old APIs before actually removing them?

abtink:
We generally do not remove OT public APIs. We may add/enhance or fix some of the APIs. 

For example, this PR addressed the issue where some of the `otUdp` APIs that they did not provide an `otInstance *` as a parameter - which is typically done/expected for all OT public APIs. This could lead to issues in some situations/configs, explained here #5162.

romacdon:
I've worked on projects before that even when APIs are enhanced or fixed the old signature of the API is still supported for some time.  Anyway, I'm OK with this change and will have someone address the issues introduced by this change.  Thanks for your explanations, @abtink .

abtink:
@romacdon sounds good. Thanks.

Is the efr32 example code (using this API) part of OT repo itself? If so, I highly suggest adding CI tests for it. This would help detect any change that would break the example code. I would have fixed/updated it as part of this PR itself, if I knew that the code was impacted and was part of OT repo.

romacdon:
@abtink yes the code is part of the OT repo.  The difficulty of adding CI tests is the inclusion of the Silabs SDK.  Until that is in an easily downloadable location its won't be possible to add the CI tests.  We are working on making it easily downloadable.

PR #5232, opened 2020-07-14
[tests] ignore unknown TLV

simonlingoogle:
This PR ignores unknown TLVs when parsing a message and continues to parse the rest parts of the message.

An unknown TLV can cause the simulator to drop the whole packet, which could lead to packet parsing issues. 
For example, if a `Child ID Response` message is dropped, the dissector will fail to add the Short Address and Extended Address to the Device Descriptors, causing subsequent secure packets unable to be decrypted. 

**PR #5219 requires this PR to pass tests stably.**

### DIFF
--- a/tests/scripts/thread-cert/mesh_cop.py
+++ b/tests/scripts/thread-cert/mesh_cop.py
@@ -934,7 +934,7 @@ class MeshCopCommandFactory:
             return self._tlvs_factories[_type]
         except KeyError:
             logging.error(
-                "Could not find TLV factory. Unsupported TLV type: {}".format(
+                'Could not find TLV factory. Unsupported TLV type: {}'.format(
                     _type))
             return UnknownTlvFactory(_type)
 
--- a/tests/scripts/thread-cert/mesh_cop.py
+++ b/tests/scripts/thread-cert/mesh_cop.py
@@ -30,10 +30,12 @@
 from binascii import hexlify
 from enum import IntEnum
 import io
+import logging
 import struct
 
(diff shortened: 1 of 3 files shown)


tests/scripts/thread-cert/mle.py
```diff
@@ -452,15 +454,13 @@ def __eq__(self, other):
...
-        return (
-            "LeaderData(partition_id={}, weighting={}, data_version={}, stable_data_version={},",
-            "leader_router_id={})").format(
-                self.partition_id,
-                self.weighting,
-                self.data_version,
-                self.stable_data_version,
-                self.leader_router_id,
-            )
+        return "LeaderData(partition_id={}, weighting={}, data_version={}, stable_data_version={},leader_router_id={}".format(
```
bukepo, line 457: use single quote for consistence.

simonlingoogle: accepted.

tests/scripts/thread-cert/mesh_cop.py
```diff
@@ -30,10 +30,12 @@
 from binascii import hexlify
 from enum import IntEnum
 import io
+import logging
 import struct
 
 from network_data import SubTlvsFactory
 import common
+from tlvs_parsing import UnknownTlvFactory
```
Irving-cl, line 38: put this line after `from network_data import SubTlvsFactory` to keep headers order?

simonlingoogle: accepted.

PR #5236, opened 2020-07-14
[mlr] Parent send MLR.req on behalf of Children

simonlingoogle:
This PR implements parents sending `MLR.req` on behave of Children.
  - Parent sends MLR.req for Children after delaying for a random time < PARENT_AGGREGATE_DELAY (5s).
  - Update test script to check for proxied `MLR.req`
      - Parent proxy MLR.req for Children after a delay t in [0, 5]

**MLR PR List:**
- #5236 Parent send MLR.req on behave of Children **<== This PR**
- #5292 Add MulticastListenersTable on BBR 
- ~~#5248 Send MLR.req for locally Netfi MAs **MERGED**~~
- ~~#5250 Add the `MLR` log region  **MERGED**~~

### DIFF
--- a/src/core/common/bit_vector.hpp
+++ b/src/core/common/bit_vector.hpp
@@ -78,24 +78,16 @@ public:
     }
 
     /**
-     * This method sets or clears the mask of a given index.
+     * This method sets the mask of a given index.
      *
      * @param[in] aIndex  The index.
-     * @param[in] aSet    TRUE to sets the mask, or FALSE to clear the mask.
      *
      */
-    void Set(uint16_t aIndex, bool aSet = true)
+    void Set(uint16_t aIndex)
     {
         OT_ASSERT(aIndex < N);
 
-        if (aSet)
-        {
-            mMask[aIndex / 8] |= 0x80 >> (aIndex % 8);
-        }
-        else
-        {
-            mMask[aIndex / 8] &= ~(0x80 >> (aIndex % 8));
-        }
+        mMask[aIndex / 8] |= 0x80 >> (aIndex % 8);
     }
 
     /**
@@ -104,7 +96,12 @@ public:
      * @param[in] aIndex  The index.
      *
      */
-    void Clear(uint16_t aIndex) { Set(aIndex, false); }
+    void Clear(uint16_t aIndex)
+    {
+        OT_ASSERT(aIndex < N);
+
+        mMask[aIndex / 8] &= ~(0x80 >> (aIndex % 8));
+    }
 
     /**
      * This method clears all indexes.
--- a/src/core/common/bit_vector.hpp
+++ b/src/core/common/bit_vector.hpp
@@ -78,15 +78,24 @@ public:
     }
 
     /**
-     * This method sets the mask of a given index.
+     * This method sets or clears the mask of a given index.
      *
      * @param[in] aIndex  The index.
+     * @param[in] aSet    TRUE to sets the mask, or FALSE to clear the mask.
      *
      */
-    void Set(uint16_t aIndex)
+    void Set(uint16_t aIndex, bool aSet = true)
     {
         OT_ASSERT(aIndex < N);
-        mMask[aIndex / 8] |= 0x80 >> (aIndex % 8);
+
+        if (aSet)
+        {
+            mMask[aIndex / 8] |= 0x80 >> (aIndex % 8);
+        }
+        else
+        {
+            mMask[aIndex / 8] &= ~(0x80 >> (aIndex % 8));
+        }
     }
 
     /**
@@ -95,11 +104,7 @@ public:
      * @param[in] aIndex  The index.
      *
      */
-    void Clear(uint16_t aIndex)
-    {
-        OT_ASSERT(aIndex < N);
-        mMask[aIndex / 8] &= ~(0x80 >> (aIndex % 8));
-    }
+    void Clear(uint16_t aIndex) { Set(aIndex, false); }
 
     /**
      * This method clears all indexes.
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -2055,13 +2055,12 @@ otError MleRouter::UpdateChildAddresses(const Message &aMessage, uint16_t aOffse
     // Retrieve registered multicast addresses of the Child
     if (aChild.HasAnyMlrRegisteredAddress())
     {
-        auto childAddresses = aChild.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal);
-
-        for (auto iter = childAddresses.begin(); iter != childAddresses.end(); iter++)
+        for (const Ip6::Address &childAddress :
+             aChild.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal))
         {
-            if (aChild.GetAddressMlrState(iter.GetAsIndex()) == kMlrStateRegistered)
+            if (aChild.GetAddressMlrState(childAddress) == kMlrStateRegistered)
             {
-                oldMlrRegisteredAddresses[oldMlrRegisteredAddressNum++] = *iter.GetAddress();
+                oldMlrRegisteredAddresses[oldMlrRegisteredAddressNum++] = childAddress;
             }
         }
     }
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -2055,6 +2055,8 @@ otError MleRouter::UpdateChildAddresses(const Message &aMessage, uint16_t aOffse
     // Retrieve registered multicast addresses of the Child
     if (aChild.HasAnyMlrRegisteredAddress())
     {
+        OT_ASSERT(aChild.IsStateValid());
+
         for (const Ip6::Address &childAddress :
              aChild.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal))
         {
@@ -4651,6 +4653,11 @@ void MleRouter::SetChildStateToValid(Child &aChild)
 
     aChild.SetState(Neighbor::kStateValid);
     IgnoreError(StoreChild(aChild));
+
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    Get<MlrManager>().UpdateProxiedSubscriptions(aChild, nullptr, 0);
+#endif
+
     Signal(OT_NEIGHBOR_TABLE_EVENT_CHILD_ADDED, aChild);
 
 exit:
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -2040,11 +2040,33 @@ otError MleRouter::UpdateChildAddresses(const Message &aMessage, uint16_t aOffse
     uint16_t                 offset          = 0;
     uint16_t                 end             = 0;
 
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    Ip6::Address oldMlrRegisteredAddresses[OPENTHREAD_CONFIG_MLE_IP_ADDRS_PER_CHILD - 1];
+    uint16_t     oldMlrRegisteredAddressNum = 0;
+#endif
+
     VerifyOrExit(aMessage.Read(aOffset, sizeof(tlv), &tlv) == sizeof(tlv), error = OT_ERROR_PARSE);
     VerifyOrExit(tlv.GetLength() <= (aMessage.GetLength() - aOffset - sizeof(tlv)), error = OT_ERROR_PARSE);
 
     offset = aOffset + sizeof(tlv);
     end    = offset + tlv.GetLength();
+
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    // Retrieve registered multicast addresses of the Child
+    if (aChild.HasAnyMlrRegisteredAddress())
+    {
+        auto childAddresses = aChild.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal);
+
+        for (auto iter = childAddresses.begin(); iter != childAddresses.end(); iter++)
+        {
+            if (aChild.GetAddressMlrState(iter.GetAsIndex()) == kMlrStateRegistered)
+            {
+                oldMlrRegisteredAddresses[oldMlrRegisteredAddressNum++] = *iter.GetAddress();
+            }
+        }
+    }
+#endif
+
     aChild.ClearIp6Addresses();
 
     while (offset < end)
@@ -2133,6 +2155,10 @@ otError MleRouter::UpdateChildAddresses(const Message &aMessage, uint16_t aOffse
         // Clear EID-to-RLOC cache for the unicast address registered by the child.
         Get<AddressResolver>().Remove(address);
     }
+
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    Get<MlrManager>().UpdateProxiedSubscriptions(aChild, oldMlrRegisteredAddresses, oldMlrRegisteredAddressNum);
+#endif
 
     if (registeredCount == 0)
     {
--- a/src/core/thread/mlr_manager.cpp
+++ b/src/core/thread/mlr_manager.cpp
@@ -139,7 +139,7 @@ bool MlrManager::IsAddressMlrRegisteredByAnyChildExcept(const Ip6::Address &aAdd
 
     OT_ASSERT(aAddress.IsMulticastLargerThanRealmLocal());
 
-    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
+    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValid))
     {
         if (&child != aExceptChild && child.HasMlrRegisteredAddress(aAddress))
         {
@@ -155,7 +155,7 @@ void MlrManager::UpdateProxiedSubscriptions(Child &             aChild,
                                             const Ip6::Address *aOldMlrRegisteredAddresses,
                                             uint16_t            aOldMlrRegisteredAddressNum)
 {
-    VerifyOrExit(aChild.IsStateValidOrRestoring(), OT_NOOP);
+    VerifyOrExit(aChild.IsStateValid(), OT_NOOP);
 
     // Search the new multicast addresses and set its flag accordingly
     for (const Ip6::Address &address : aChild.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal))
@@ -194,7 +194,7 @@ exit:
 
 void MlrManager::SetChildMulticastAddressMlrState(MlrState aFromState, MlrState aToState)
 {
-    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
+    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValid))
     {
         for (const Ip6::Address &address : child.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal))
         {
@@ -274,7 +274,7 @@ void MlrManager::SendMulticastListenerRegistration(void)
 
 #if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
     // Append Child multicast addresses
-    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
+    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValid))
     {
         if (addressesNum >= kIPv6AddressesNumMax)
         {
@@ -495,7 +495,7 @@ void MlrManager::LogMulticastAddresses(void)
 #endif
 
 #if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
-    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
+    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValid))
     {
         for (const Ip6::Address &address : child.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal))
         {
--- a/src/core/thread/mlr_manager.cpp
+++ b/src/core/thread/mlr_manager.cpp
@@ -155,15 +155,12 @@ void MlrManager::UpdateProxiedSubscriptions(Child &             aChild,
                                             const Ip6::Address *aOldMlrRegisteredAddresses,
                                             uint16_t            aOldMlrRegisteredAddressNum)
 {
-    auto childAddresses = aChild.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal);
-
     VerifyOrExit(aChild.IsStateValidOrRestoring(), OT_NOOP);
 
     // Search the new multicast addresses and set its flag accordingly
-    for (auto iter = childAddresses.begin(); iter != childAddresses.end(); iter++)
+    for (const Ip6::Address &address : aChild.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal))
     {
-        bool         isMlrRegistered = false;
-        Ip6::Address address         = *iter.GetAddress();
+        bool isMlrRegistered = false;
 
         // Check if it's a new multicast address against old addresses
         for (size_t i = 0; i < aOldMlrRegisteredAddressNum; i++)
@@ -183,7 +180,7 @@ void MlrManager::UpdateProxiedSubscriptions(Child &             aChild,
         isMlrRegistered = isMlrRegistered || IsAddressMlrRegisteredByAnyChildExcept(address, &aChild);
 
         // Set MLR state accordingly
-        aChild.SetAddressMlrState(iter.GetAsIndex(), isMlrRegistered ? kMlrStateRegistered : kMlrStateToRegister);
+        aChild.SetAddressMlrState(address, isMlrRegistered ? kMlrStateRegistered : kMlrStateToRegister);
     }
 
 exit:
@@ -199,13 +196,11 @@ void MlrManager::SetChildMulticastAddressMlrState(MlrState aFromState, MlrState 
 {
     for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
     {
-        auto addresses = child.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal);
-
-        for (auto iter = addresses.begin(); iter != addresses.end(); iter++)
+        for (const Ip6::Address &address : child.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal))
         {
-            if (child.GetAddressMlrState(iter.GetAsIndex()) == aFromState)
+            if (child.GetAddressMlrState(address) == aFromState)
             {
-                child.SetAddressMlrState(iter.GetAsIndex(), aToState);
+                child.SetAddressMlrState(address, aToState);
             }
         }
     }
@@ -281,8 +276,6 @@ void MlrManager::SendMulticastListenerRegistration(void)
     // Append Child multicast addresses
     for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
     {
-        auto childAddresses = child.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal);
-
         if (addressesNum >= kIPv6AddressesNumMax)
         {
             break;
@@ -293,17 +286,17 @@ void MlrManager::SendMulticastListenerRegistration(void)
             continue;
         }
 
-        for (auto iter = childAddresses.begin(); iter != childAddresses.end(); iter++)
+        for (const Ip6::Address &address : child.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal))
         {
             if (addressesNum >= kIPv6AddressesNumMax)
             {
                 break;
             }
 
-            if (child.GetAddressMlrState(iter.GetAsIndex()) == kMlrStateToRegister)
+            if (child.GetAddressMlrState(address) == kMlrStateToRegister)
             {
-                AppendToUniqueAddressList(addresses, addressesNum, *iter.GetAddress());
-                child.SetAddressMlrState(iter.GetAsIndex(), kMlrStateRegistering);
+                AppendToUniqueAddressList(addresses, addressesNum, address);
+                child.SetAddressMlrState(address, kMlrStateRegistering);
             }
         }
     }
@@ -497,22 +490,17 @@ void MlrManager::LogMulticastAddresses(void)
 #if OPENTHREAD_CONFIG_MLR_ENABLE
     for (const Ip6::ExternalNetifMulticastAddress &addr : Get<ThreadNetif>().IterateExternalMulticastAddresses())
     {
-        MlrState state = addr.GetMlrState();
-
-        otLogDebgMlr("%-32s%c", addr.GetAddress().ToString().AsCString(), "-rR"[state]);
+        otLogDebgMlr("%-32s%c", addr.GetAddress().ToString().AsCString(), "-rR"[addr.GetMlrState()]);
     }
 #endif
 
 #if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
     for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
     {
-        auto childAddresses = child.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal);
-
-        for (auto iter = childAddresses.begin(); iter != childAddresses.end(); iter++)
+        for (const Ip6::Address &address : child.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal))
         {
-            MlrState state = child.GetAddressMlrState(iter.GetAsIndex());
-
-            otLogDebgMlr("%-32s%c %04x", iter.GetAddress()->ToString().AsCString(), "-rR"[state], child.GetRloc16());
+            otLogDebgMlr("%-32s%c %04x", address.ToString().AsCString(), "-rR"[child.GetAddressMlrState(address)],
+                         child.GetRloc16());
         }
     }
 #endif
--- a/src/core/thread/mlr_manager.cpp
+++ b/src/core/thread/mlr_manager.cpp
@@ -33,7 +33,7 @@
 
 #include "mlr_manager.hpp"
 
-#if OPENTHREAD_CONFIG_MLR_ENABLE
+#if OPENTHREAD_CONFIG_MLR_ENABLE || OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
 
 #include "common/code_utils.hpp"
 #include "common/instance.hpp"
@@ -58,10 +58,12 @@ MlrManager::MlrManager(Instance &aInstance)
 
 void MlrManager::HandleNotifierEvents(Events aEvents)
 {
+#if OPENTHREAD_CONFIG_MLR_ENABLE
     if (aEvents.Contains(kEventIp6MulticastSubscribed))
     {
-        ScheduleSend(0);
+        UpdateLocalSubscriptions();
     }
+#endif
 
     if (aEvents.Contains(kEventThreadRoleChanged) && Get<Mle::MleRouter>().IsChild())
     {
@@ -80,6 +82,136 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
 
     UpdateReregistrationDelay(needRereg);
 }
+
+#if OPENTHREAD_CONFIG_MLR_ENABLE
+void MlrManager::UpdateLocalSubscriptions(void)
+{
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    // Check multicast addresses are newly listened against Children
+    for (Ip6::ExternalNetifMulticastAddress &addr : Get<ThreadNetif>().IterateExternalMulticastAddresses())
+    {
+        if (addr.GetAddress().IsMulticastLargerThanRealmLocal() && addr.GetMlrState() == kMlrStateToRegister &&
+            IsAddressMlrRegisteredByAnyChild(addr.GetAddress()))
+        {
+            addr.SetMlrState(kMlrStateRegistered);
+        }
+    }
+#endif // OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+
+    ScheduleSend(0);
+}
+
+bool MlrManager::IsAddressMlrRegisteredByNetif(const Ip6::Address &aAddress)
+{
+    bool ret = false;
+
+    OT_ASSERT(aAddress.IsMulticastLargerThanRealmLocal());
+
+    for (const Ip6::ExternalNetifMulticastAddress &addr : Get<ThreadNetif>().IterateExternalMulticastAddresses())
+    {
+        if (addr.GetAddress() == aAddress && addr.GetMlrState() == kMlrStateRegistered)
+        {
+            ExitNow(ret = true);
+        }
+    }
+
+exit:
+    return ret;
+}
+
+void MlrManager::SetNetifMulticastAddressMlrState(MlrState aFromState, MlrState aToState)
+{
+    for (Ip6::ExternalNetifMulticastAddress &addr : Get<ThreadNetif>().IterateExternalMulticastAddresses())
+    {
+        if (addr.GetAddress().IsMulticastLargerThanRealmLocal() && addr.GetMlrState() == aFromState)
+        {
+            addr.SetMlrState(aToState);
+        }
+    }
+}
+#endif // OPENTHREAD_CONFIG_MLR_ENABLE
+
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+
+bool MlrManager::IsAddressMlrRegisteredByAnyChildExcept(const Ip6::Address &aAddress, const Child *aExceptChild)
+{
+    bool ret = false;
+
+    OT_ASSERT(aAddress.IsMulticastLargerThanRealmLocal());
+
+    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
+    {
+        if (&child != aExceptChild && child.HasMlrRegisteredAddress(aAddress))
+        {
+            ExitNow(ret = true);
+        }
+    }
+
+exit:
+    return ret;
+}
+
+void MlrManager::UpdateProxiedSubscriptions(Child &             aChild,
+                                            const Ip6::Address *aOldMlrRegisteredAddresses,
+                                            uint16_t            aOldMlrRegisteredAddressNum)
+{
+    auto childAddresses = aChild.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal);
+
+    VerifyOrExit(aChild.IsStateValidOrRestoring(), OT_NOOP);
+
+    // Search the new multicast addresses and set its flag accordingly
+    for (auto iter = childAddresses.begin(); iter != childAddresses.end(); iter++)
+    {
+        bool         isMlrRegistered = false;
+        Ip6::Address address         = *iter.GetAddress();
+
+        // Check if it's a new multicast address against old addresses
+        for (size_t i = 0; i < aOldMlrRegisteredAddressNum; i++)
+        {
+            if (aOldMlrRegisteredAddresses[i] == address)
+            {
+                isMlrRegistered = true;
+                break;
+            }
+        }
+
+#if OPENTHREAD_CONFIG_MLR_ENABLE
+        // Check if it's a new multicast address against parent Netif
+        isMlrRegistered = isMlrRegistered || IsAddressMlrRegisteredByNetif(address);
+#endif
+        // Check if it's a new multicast address against other Children
+        isMlrRegistered = isMlrRegistered || IsAddressMlrRegisteredByAnyChildExcept(address, &aChild);
+
+        // Set MLR state accordingly
+        aChild.SetAddressMlrState(iter.GetAsIndex(), isMlrRegistered ? kMlrStateRegistered : kMlrStateToRegister);
+    }
+
+exit:
+    LogMulticastAddresses();
+
+    if (aChild.HasAnyMlrToRegisterAddress())
+    {
+        static_assert(Mle::kParentAggregateDelay > 1, "kParentAggregateDelay should be larger than 1 second");
+        ScheduleSend(Random::NonCrypto::GetUint16InRange(1, Mle::kParentAggregateDelay));
+    }
+}
+
+void MlrManager::SetChildMulticastAddressMlrState(MlrState aFromState, MlrState aToState)
+{
+    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
+    {
+        auto addresses = child.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal);
+
+        for (auto iter = addresses.begin(); iter != addresses.end(); iter++)
+        {
+            if (child.GetAddressMlrState(iter.GetAsIndex()) == aFromState)
+            {
+                child.SetAddressMlrState(iter.GetAsIndex(), aToState);
+            }
+        }
+    }
+}
+#endif // OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
 
 void MlrManager::ScheduleSend(uint16_t aDelay)
 {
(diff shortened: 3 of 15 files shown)


src/core/thread/mlr_manager.cpp
```diff
@@ -117,6 +117,50 @@ void MlrManager::UpdateLocalSubscriptions(void)
...
+
+    // Add new multicast address of this Child
+    while (aChild.GetNextIp6Address(childAddrIter, address) == OT_ERROR_NONE)
+    {
+        if (!address.IsMulticastLargerThanRealmLocal())
+        {
+            continue;
+        }
+
+        mTable.SetChildProxied(address, childIndex, true);
```
jwhui, line 154: Continuing the conversation from https://github.com/openthread/openthread/pull/5219#discussion_r455473665:

> As for the proxied MLR.req, we analyzed the logic of sending MLR.req without using any table and found that the time complexity could be O(ChildNum * IpNumPerChild * IP6AddressesTlvSize). On the other hand, the current implementation using a table has a time complexity of O(MulticastIpNum). I think we can discuss more in the proxied MLR.req PR #5236 .

I think there are ways to reduce the overhead. For example, we could maintain a bit-vector with each bit corresponding to a child's address in the child table. A bit is set if the corresponding child's address is to be used when sending an MLR.req.

When adding a child multicast address, check to see if any of the addresses marked for use in MLR.req match (i.e. the address entries with the corresponding bit set).

1. If there is a match, clear the bit corresponding to the address being added. This indicates that the newly added address is already included in MLR.req messages.

1. If there is no match, set the bit corresponding to the address being added. This indicates that the newly added address needs to be added to MLR.req messages.

When removing a child multicast address:

1. If the corresponding bit is set, check to see if any other child has a matching address. If no matching address, clear the bit. This indicates that the address can be removed from the MLR.req message.

1. If the corresponding bit is clear, simply remove the address as this indicates that other children are still using the address.

When sending an MLR.req message, simply send the multicast addresses that have the bit set.

If we wanted to further optimize memory, then we could distribute the bit vector into the child table - I believe there is some padding that we could use to avoid any additional memory overhead. If we wanted to avoid scanning each child table entry, we could have a top-level bit vector that indicates whether a child entry has any such bit set.

Thoughts?

simonlingoogle: Yes we only need to check Child addresses that was marked as newly added.
We have a similar flag defined in #5236 : `MlrState` whose value can be `ToRegister (0)` or `Registering (1)` or `Registered (2)`

I have created PR #5248 which has less RAM cost and code size (to my supprise). 
The reduction in code size was mainly due to the compaction of `MlrManager::UpdateLocalSubscriptions` because it does not need to copy addresses anymore.

simonlingoogle: @jwhui @librasungirl @abtink 
Some further thoughts on this:

Let's first assume that Child has a flag for each Address (domain >= 4) to indicate if it's registered (1) or not (0). 
I found that following invariant needs to be satisfied:

- For a MA, the flag should not be both 1 and 0 for different Children. MA with both flag 1 and 0 **should not** be included in the `MLR.req.`  If the flag can be both 1 and 0 on different Children, when sending `MLR.req`, for each MA with flag = 0, we will have to check if has flag 1 on other children. This lead to a time complexity of `O(TotalIpNum*TotalIpNum)`, which I think is something we should avoid.

This lead to the further consideration that when a Child adds a MA, we need to check all MAs on all Children to check if this MA is indeed newly listened. If this MA was registered (with flag=1) by the any Child, the MA's flag should be set to 1. If the MA was not listened or was listened but not registered (flag = 0), the MA's flag should be set to 0. 

Thus for time complexity, we can have folowings:
- Adding a MA to Child has a time complexity of O(TotalIpNum) to search for same MA with flag=1
- Sending MLR.req has a time complexity of O(TotalIpNum * IP6AddressesTlvSize). 

This however has not taken into consideration the bit mask optimization tricks @jwhui  have suggested. 
- To efficiently searching for MAs with flag=0, we can have a `mToRegisterMask` bit-vector for each Child. 
- To efficiently searching for MAs with flag=1, we can have a `mRegisteredMask` bit-vector for each Child. 

These bit masks will allow us to skip Child that has no flag=0 or flag=1 MAs.

jwhui: I think the important point is we want to quickly identify the set of unique multicast addresses. My proposal was to have a bit for each address that indicates whether the given multicast address is "authoritative" or "duplicate". When sending MR.req message, we only consider and maintain state for "authoritative" IPv6 addresses. This should avoid the O(N^2) time complexity and result in O(N) time complexity. Of course, it means that adding/removing a multicast address also has O(N) time complexity.

The bit-vector simply makes it a bit more efficient by avoiding the walk across all child address entries, but it's still O(N) time complexity.

Thoughts?

simonlingoogle: image

Now that the proxying is more or less implemented, we can see the size change to compare the current `table-less` approach (in green) with the earlier `table` approach (in red).

- The `table-less` implemented is MTD-friendly to have less code size and RAM footprint. 
- For FTD, the `table-less` approaches does have code size advantage nor disadvantage but it can proxy more Child addresses (30 v.s. 15) with same RAM footprint.

src/core/config/ip6.h
```diff
@@ -177,4 +177,12 @@
 #define OPENTHREAD_CONFIG_UNSECURE_TRAFFIC_MANAGED_BY_STACK_ENABLE 0
 #endif
 
+/**
+ *
+ * This setting configures the Multicast Listener Registration parent proxing in Thread 1.2.
+ * This is compulsory for 1.2 FTD.
+ *
+ */
+#define OPENTHREAD_CONFIG_MLR_PROXY_ENABLE (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2) && OPENTHREAD_FTD
```
librasungirl, line 186: Maybe more proper to put in src/core/config/tmf.h? as it mainly related to the CoAP transaction.

simonlingoogle: moved

simonlingoogle:
Hi @librasungirl, @abtink , @jwhui 

This PR is ready for reviewing. The most part of logic is to check if each MA is already registered by local Netif or other Children. 
Please take a look when you have time.  
Thanks,

src/core/thread/mlr_manager.cpp
```diff
@@ -81,6 +83,126 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
...
+    return ret;
+}
+
+void MlrManager::UpdateProxiedSubscriptions(Child &             aChild,
+                                            const Ip6::Address *aRegisteredAddresses,
+                                            uint16_t            aRegisteredAddressNum)
+{
+    Ip6::Address address;
+
+    VerifyOrExit(aChild.IsStateValidOrRestoring(), OT_NOOP);
```
librasungirl, line 146:
```suggestion
    VerifyOrExit(aChild.IsStateValid(), OT_NOOP);
```

src/core/thread/mlr_manager.cpp
```diff
@@ -81,6 +83,126 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
...
+
+    // Search for the new multicast addresses and set it's flag accordingly
```
librasungirl, line 148:
```suggestion
    // Search the new multicast addresses and set its flag accordingly
```

src/core/thread/topology.hpp
```diff
@@ -743,6 +753,53 @@ class Child : public Neighbor, public IndirectSender::ChildInfo, public DataPoll
...
+     *
+     */
+    void SetAddressMlrState(uint16_t aAddressIndex, MlrState aState);
+
+    /**
+     * This method returns if the Child has Ip6 address @p aAddress of MLR state `kMlrStateRegistered`.
+     *
+     * @param[in] aAddress  The Ip6 address.
+     *
+     * @returns Whether the Child has Ip6 address @p aAddress of MLR state `kMlrStateRegistered`.
```
librasungirl, line 781: nit here and other places as
```
* @retval true  ......
* @retval false ......
```

src/core/thread/topology.cpp
```diff
@@ -266,6 +270,90 @@ void Child::GenerateChallenge(void)
...
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+bool Child::HasMlrRegisteredAddress(const Ip6::Address &aAddress)
+{
+    bool         has = false;
+    Ip6::Address address;
+
+    if (!mMlrRegisteredMask.HasAny())
+    {
+        ExitNow();
+    }
```
librasungirl, line 282:
```suggestion
    VerifyOrExit(mMlrRegisteredMask.HasAny(), OT_NOOP);   
```

src/core/thread/mlr_manager.hpp
```diff
@@ -86,6 +88,21 @@ class MlrManager : public InstanceLocator, public Notifier::Receiver
...
                                            const BackboneRouter::BackboneRouterConfig &aConfig);
 
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    /**
+     * This function updates the Multicast Subscription Table according to the Child information.
+     *
+     * @param[in]   aChild A reference to the child information.
+     *
+     * @retval  OT_ERROR_NONE          Successfully updated.
+     * @retval  OT_ERROR_NO_BUFS       No available entry in ListenerTable.
```
librasungirl, line 98: update doxygen?

src/core/thread/mlr_manager.hpp
```diff
@@ -113,8 +130,34 @@ class MlrManager : public InstanceLocator, public Notifier::Receiver
...
+#if OPENTHREAD_CONFIG_MLR_ENABLE
+        SetNetifMulticastAddressMlrState(aFromState, aToState);
+#endif
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+        SetChildMulticastAddressMlrState(aFromState, aToState);
+#endif
+    }
+
+    void AppendToUniqueAddressList(Ip6::Address (&aAddresses)[kIPv6AddressesNumMax],
+                                   uint8_t &           aAddressNum,
```
librasungirl, line 159: nit: remove `&`?

src/core/thread/mlr_manager.cpp
```diff
@@ -367,6 +529,29 @@ void MlrManager::SetNetifMulticastAddressMlrState(MlrState aFromState, MlrState
         }
     }
 }
+#endif // OPENTHREAD_CONFIG_MLR_ENABLE
```
librasungirl, line 532: nit: move to wrap `OPENTHREAD_CONFIG_MLR_ENABLE` relative methods together?

src/core/thread/mlr_manager.cpp
```diff
@@ -81,6 +83,126 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
...
+        // Set MLR state accordingly
+        aChild.SetAddressMlrState(iter.Get(), isMlrRegistered ? kMlrStateRegistered : kMlrStateToRegister);
+    }
+
+exit:
+    if (aChild.HasAnyMlrToRegisterAddress())
+    {
+        LogMulticastAddresses();
+
+        static_assert(Mle::kParentAggregateDelay > 1, "kParentAggregateDelay should be larger than 1 second");
```
librasungirl, line 184: `kParentAggregateDelay` is a constant, if check, better to put it to the header file `mle_types.hpp` where it is defined?

src/core/thread/mlr_manager.cpp
```diff
@@ -81,6 +83,126 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
...
+        ScheduleSend(Random::NonCrypto::GetUint16InRange(1, Mle::kParentAggregateDelay));
+    }
+}
+
+void MlrManager::SetChildMulticastAddressMlrState(MlrState aFromState, MlrState aToState)
+{
+    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
```
librasungirl, line 191:
```suggestion
    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValid))
```

src/core/thread/mlr_manager.cpp
```diff
@@ -121,44 +243,77 @@ void MlrManager::SendMulticastListenerRegistration(void)
...
-            SuccessOrExit(error = message->Append(&addr.GetAddress(), sizeof(Ip6::Address)));
+            AppendToUniqueAddressList(addresses, addressesNum, addr.GetAddress());
             addr.SetMlrState(kMlrStateRegistering);
+        }
+    }
+#endif
+
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    // Append Child multicast addresses
+    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
```
librasungirl, line 280:
```suggestion
    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValid))
```

src/core/thread/mlr_manager.cpp
```diff
@@ -81,6 +83,126 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
...
+}
+
+bool MlrManager::IsAddressMlrRegisteredByNetif(const Ip6::Address &aAddress)
+{
+    bool ret = false;
+
+    for (const Ip6::ExternalNetifMulticastAddress &addr : Get<ThreadNetif>().IterateExternalMulticastAddresses())
+    {
+        if (addr.GetAddress().IsMulticastLargerThanRealmLocal() && addr.GetMlrState() == kMlrStateRegistered &&
+            addr.GetAddress() == aAddress)
```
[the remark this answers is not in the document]
simonlingoogle: Great idea! Then we don't need to check `addr.GetAddress().IsMulticastLargerThanRealmLocal()` right?

src/core/thread/mlr_manager.cpp
```diff
@@ -121,44 +243,77 @@ void MlrManager::SendMulticastListenerRegistration(void)
...
+                child.GetAddressMlrState(iter.Get()) == kMlrStateToRegister)
+            {
+                AppendToUniqueAddressList(addresses, addressesNum, address);
+                child.SetAddressMlrState(iter.Get(), kMlrStateRegistering);
+            }
         }
     }
+#endif
+
+    VerifyOrExit(addressesNum > 0, error = OT_ERROR_NOT_FOUND);
```
librasungirl, line 311: nit:  move this before CoAP message buffer allocation at Line 254?

src/core/thread/mlr_manager.cpp
```diff
@@ -330,6 +486,7 @@ void MlrManager::LogMulticastAddresses(void)
 #if OPENTHREAD_CONFIG_LOG_BBR && OPENTHREAD_CONFIG_LOG_LEVEL >= OT_LOG_LEVEL_DEBG
```
librasungirl, line 486:
```suggestion
#if OPENTHREAD_CONFIG_LOG_MLR && OPENTHREAD_CONFIG_LOG_LEVEL >= OT_LOG_LEVEL_DEBG
```

src/core/thread/mlr_manager.cpp
```diff
@@ -340,23 +497,28 @@ void MlrManager::LogMulticastAddresses(void)
...
+        Ip6::Address address;
+
+        for (Child::Ip6AddressIterator iter; child.GetNextIp6Address(iter, address) == OT_ERROR_NONE;)
         {
-            count++;
+            if (address.IsMulticastLargerThanRealmLocal())
+            {
+                MlrState state = child.GetAddressMlrState(iter.Get());
+
+                otLogDebgMlr("%-32s%c", address.ToString().AsCString(), "-rR"[state]);
```
librasungirl, line 512: also log the child short address?

src/core/thread/mlr_manager.cpp
```diff
@@ -81,6 +83,126 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
...
+}
+#endif // OPENTHREAD_CONFIG_MLR_ENABLE
+
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+
+bool MlrManager::IsAddressMlrRegisteredByAnyChildExcept(const Ip6::Address &aAddress, const Child *aExceptChild)
+{
+    bool ret = false;
+
+    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
```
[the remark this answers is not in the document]
simonlingoogle: Any thoughts on whether or not we should consider a restoring Child?

src/core/thread/mlr_manager.cpp
simonlingoogle: Sounds reasonable. 
But here is the actual logic that imposes such restriction. So, putting it here might be more logical?

src/core/thread/mlr_manager.cpp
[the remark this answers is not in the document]
librasungirl: When the parent is up after reboot, it will set its children as [kStateRestored](https://github.com/openthread/openthread/blob/master/src/core/thread/mle_router.cpp#L3688); after it re-establish link with neighboring routers, it will try to restore its children via ChildUpdateRequest() and set the state as [kStateChildUpdateRequest](https://github.com/openthread/openthread/blob/master/src/core/thread/mle_router.cpp#L3100).

The parent won't store the MAs for its children,  it only knows the MLEID of the child before restore the link with its child.
It would be useless to iterate the children in restoring state.

thoughts?

src/core/thread/mlr_manager.cpp
librasungirl: You put the other check in header file, any rules to follow?
https://github.com/openthread/openthread/blob/d14a14a494c4ca2379ec61a919e2c8c7335ca637/src/core/thread/mle_types.hpp#L266

simonlingoogle: `static_assert(kMlrTimeoutDefault >= kMlrTimeoutMin` is now put under definition of the two constants, which I think is reasonable. 

For `Mle::kParentAggregateDelay`, defining it to 1 may seem fine. If we put the `static_assert(Mle::kParentAggregateDelay > 1)` right after its definition, we don't have the context of why. 

So, I think it's better to be put where `Mle::kParentAggregateDelay > 1` is actually required, which is the statement:
```c++
Random::NonCrypto::GetUint16InRange(1, Mle::kParentAggregateDelay)
```

Thoughts?

src/core/thread/mlr_manager.cpp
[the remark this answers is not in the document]
simonlingoogle: I digged into the code for a little bit:

**`UpdateChildAddresses` is called in 3 places:**

1. `HandleChildIdRequest` which can change Child state from `Invalid` to `ChildIdRequest` or 
2. `HandleChildUpdateRequest` which can change Child state from `Restoring` to `Valid`
3. `HandleChildUpdateResponse` which can change Child state from `ChildUpdateRequest` to `Valid`

One thing I noticed is that in both case 2 and 3, `UpdateChildAddresses` is called before `SetChildStateToValid`. So, I think it necessary to also consider `Restoring` state since it just going to be `Valid` immediately. 

For case 1 (`HandleChildIdRequest`),  `UpdateChildAddresses` seems to be called when the Child state is still invalid. In such case, no address will be processed by `MlrManager`, which I think is an issue. 

Thoughts? @librasungirl

src/core/thread/mlr_manager.hpp
```diff
@@ -86,6 +88,20 @@ class MlrManager : public InstanceLocator, public Notifier::Receiver
     void HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State               aState,
                                            const BackboneRouter::BackboneRouterConfig &aConfig);
 
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    /**
+     * This function updates the Multicast Subscription Table according to the Child information.
```
abtink, line 93:
```suggestion
     * This method updates the Multicast Subscription Table according to the Child information.
```

src/core/thread/topology.cpp
```diff
@@ -274,6 +278,87 @@ void Child::GenerateChallenge(void)
...
+        mlrState = kMlrStateToRegister;
+    }
+    else if (registered)
+    {
+        mlrState = kMlrStateRegistered;
+    }
+    else
+    {
+        mlrState = kMlrStateRegistering;
+    }
```
abtink, line 321:
```suggestion
    mlrState = toRegister ? kMlrStateToRegister : (registered ? kMlrStateRegistered : kMlrStateRegistering);
```

src/core/thread/topology.cpp
```diff
@@ -274,6 +278,87 @@ void Child::GenerateChallenge(void)
...
+        toRegister = true;
+        break;
+    case kMlrStateRegistered:
+        registered = true;
+        break;
+    }
+
+    if (toRegister)
+    {
+        mMlrToRegisterMask.Set(aAddressIndex);
```
abtink, line 345: How about adding `Set(index, bool)` in `BitVector` to set a bit at a given index to a given value?
(maybe in future PR?).

src/core/thread/mle_router.cpp
```diff
@@ -2040,11 +2040,33 @@ otError MleRouter::UpdateChildAddresses(const Message &aMessage, uint16_t aOffse
...
     offset = aOffset + sizeof(tlv);
     end    = offset + tlv.GetLength();
+
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    // Retrieve registered multicast addresses of the Child
+    if (aChild.HasAnyMlrRegisteredAddress())
+    {
+        auto childAddresses = aChild.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal);
+
+        for (auto iter = childAddresses.begin(); iter != childAddresses.end(); iter++)
```
abtink, line 2060: Suggest using the iterator directly to avoid using hidden/private type (through use of `auto`).

```
for (Child::AddressIterator iter(aChild, Ip6::Address::kTypeMulticastLargerThanRealmLocal); !iter.IsDone(); iter++)
```

I think we want to allow `IterateIp6Addresses()`  for use in range-based `for` loop only.

src/core/thread/mle_router.cpp
```diff
@@ -2134,6 +2156,10 @@ otError MleRouter::UpdateChildAddresses(const Message &aMessage, uint16_t aOffse
         Get<AddressResolver>().Remove(address);
     }
 
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    Get<MlrManager>().UpdateProxiedSubscriptions(aChild, oldMlrRegisteredAddresses, oldMlrRegisteredAddressNum);
```
abtink, line 2160: If I understand it correctly, in the code above, we go through all child addresses and find all (old) registered addresses, copy them into an array and then pass the array here to `UpdateProxiedSubscriptions()` along with the `aChild`.

Can we consider having the `UpdateProxiedSubscriptions()` perform this (since we are passing `aChild` to it already).

src/core/thread/mlr_manager.cpp
```diff
@@ -81,6 +83,141 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
...
+                                            const Ip6::Address *aRegisteredAddresses,
+                                            uint16_t            aRegisteredAddressNum)
+{
+    Ip6::Address address;
+    auto         childAddresses = aChild.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal);
+
+    VerifyOrExit(aChild.IsStateValidOrRestoring(), OT_NOOP);
+
+    // Search the new multicast addresses and set its flag accordingly
+    for (auto iter = childAddresses.begin(); iter != childAddresses.end(); iter++)
```
abtink, line 164: Same as earlier comment. Suggest using `Iterator` type directly here (avoid `auto` which would use hidden/private type)?

src/core/thread/mlr_manager.cpp
```diff
@@ -81,6 +83,141 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
...
+
+        // Check if it's a new multicast address against old addresses
+        for (size_t i = 0; i < aRegisteredAddressNum; i++)
+        {
+            if (aRegisteredAddresses[i] == address)
+            {
+                isMlrRegistered = true;
+                break;
+            }
+        }
```
abtink, line 181: Can this work instead?
```
isMlrRegistered = (aChild.GetAddressMlrState(iter.GetAsIndex()) == kMlrStateRegistered);
```

src/core/thread/topology.hpp
```diff
@@ -862,6 +872,56 @@ class Child : public Neighbor, public IndirectSender::ChildInfo, public DataPoll
...
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    /**
+     * This method returns MLR state of an Ip6 address.
+     *
+     * @param[in] aAddressIndex  The Ip6 address index.
+     *
+     * @returns MLR state of the Ip6 address.
+     *
+     */
+    MlrState GetAddressMlrState(uint16_t aAddressIndex);
```
abtink, line 884: I have an idea which I think can help simplify the code.

How about instead of passing `aAddressIndex` to the `Get/SetMlrState` methods, we require the caller to pass a reference to the `aAddress` entry itself.
-  We require `aAddress` reference to be the same reference as the one user does get from iterating over `Child` addresses (i.e. same reference that we get when we loop over addresses using `IterateIp6Addresses()`).
- Then inside `Child` class, we simply check the address of `aAddress` against `mIp6Address[]` to find an index which we can use to get to two masks `mMlrToRegisterMask` and `mMlrRegisteredMask`

I feel this model may be more straight-forward than current model of using an `AddressIterator` to go through addresses, converting the iterator to an index, then passing the index to `Child` methods get/set more info/state about the address of the child.

src/core/thread/mlr_manager.cpp
simonlingoogle: No. `aChild` is the current Child whose addresses are updated via `Child Update Requests`. 
This code is executed after `aChild` has updated all its addresses to new address.  So we should check current address against the original MLR registered addresses, rather than current addresses.

src/core/thread/topology.hpp
```diff
@@ -862,6 +872,56 @@ class Child : public Neighbor, public IndirectSender::ChildInfo, public DataPoll
...
+
```
abtink, line 885: Follow up on previous approach:

What we want is to get/set the MLR state on a specific "child IPv6 address" as we iterate through the addresses.

How about the following
- We can consider defining a new class/type `Child::Ip6Address` which represents an IPv6 address on a `Child`
  - It provides `const Ip6::Address &GetAddress()` to get the associated IPv6 address, 
  -  `Child &GetChild()` for the associated `Child`, etc
  - It can also provide `SetMlrState()` and `GetMlrState()` to allow MLR state to be set on a specific address of the child 
     - The implementation of these can call some private methods of `Child` to update the masks but that all of this would be hidden from caller).
- We change iterator such that when we iterate using `IterateIp6Addresses()`, we get a `Child::Ip6Address` instead of ab `Ip6::Address`, i.e.
```
   for (Child::Ip6Address &addr : aChild.IterateIp6Addresses(filter))
   {
        // Can use addr.GetAddress() to get the actual `Ip6::Address &`
        // Can use `addr.GetMlrState()` or `addr.SetMleState()` directly on `addr`
        ...
   }
````

File: src/core/thread/mle_router.cpp (hunk 22 above)
Comment (simonlingoogle) [reply src/core/thread/mle_router.cpp:2160]: You understood correctly.

We must save old registered addresses and pass them to  `UpdateProxiedSubscriptions` because they are rewritten in this function. 

If we decide to `UpdateProxiedSubscriptions` before old addresses are cleared in `Child`, then we need to retrieve all new addresses and pass them to  `UpdateProxiedSubscriptions`. I think this is complicated because the new addresses could be compressed so we need to extract them. We also can not rewrite the two old masks (i.e. `mMlrRegisteredMask` and `mMlrToRegisterMask`) because they are still useful.

File: src/core/thread/topology.cpp (hunk 20 above)
Comment (simonlingoogle) [reply src/core/thread/topology.cpp:344]: I also felt the need to have this method. Let's do it now!

File: src/core/thread/topology.hpp (hunk 26 above)
Comment (simonlingoogle) [reply src/core/thread/topology.hpp:885]: This sounds good to me overall. My only concern is that having many different `XxxAddress` could make the code less straightforward to understand?

File: src/core/thread/mle_router.cpp (hunk 21 above)
Comment (simonlingoogle) [reply src/core/thread/mle_router.cpp:2060]: I think there is no such code now.

File: src/core/thread/topology.hpp (hunk 26 above)
Comment (simonlingoogle) [reply src/core/thread/topology.hpp:885]: If `IterateIp6Addresses` iterates on `Child::Ip6Address &addr`, we will need to construct a fake `Child::Ip6Address` from MLEID.

Comment (simonlingoogle) [reply src/core/thread/topology.hpp:885]: I tried to use `Ip6Address` directly:
```c++
Ip6Address               mIp6Address[kNumIp6Addresses]; ///< Registered IPv6 addresses
``` 

Turnout it's not workable because `Ip6Address` can not contain any additional information (we don't want to increase the memory consumption of Child). 

Then I realized that `Ip6Address` is essential the `AddressIterator` itself. We can even define `GetMlrState` and `SetMlrState` in `AddressIterator` class and it will work.

That being said, I think `IterateIp6Addresses()` should return `Ip6::Address` to have a straightforward interface. Now I tend to keep the code as it is so as to make `AddressIterator` less complex.

src/core/thread/mle_router.cpp
abtink: I see. We deleter the addresses. got it.

src/core/thread/topology.hpp
abtink: We don't want to change the `mIp6Address` it should still be `Ip6::Address mIp6Address`.
The `Child::Ip6Address` type would use indireation, i.e. has a pointer/reference to an `Ip6::Address` (which we get from `mIpAddress[]` or it is from iterator `mMeshLocalAddress`). I agree the `AddressIterator` itself can be made to look like `Child::Ip6Address`.

While I like this approach, my concern with it is that it will increase code size (due to all indirection and new type). So I think we can consider it for future and try it out later after this PR is merged.

I would suggest for this PR, we follow the https://github.com/openthread/openthread/pull/5236/files#r459920104 (from comment above), i.e. pass the `aAddress` reference directly to `Get/SetMlrState()` (instead of iter index).

src/core/thread/mlr_manager.cpp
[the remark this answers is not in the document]
librasungirl: Generally speaking, I think only the keep-alive child update request could for sure include all unicast/multicast that should be registered.

For case 3,  ChildUpdateResponse from child only include MLEID.
https://github.com/openthread/openthread/blob/c5e7252fed253760e76097179a1078f410bcd641/src/core/thread/mle.cpp#L2324

For case 2, when the child is `Restoring`, if it is a fragmented keep-alive CUR from child, it may have no chance to go up to MLE layer.  
If tracing back to #1853 which introduced this change, it is a Child Update Request from child which reboots, which in theory should be better to only include mesh local eid (#5295 is submitted for this).

Thoughts?

src/core/thread/mlr_manager.cpp
librasungirl: No strong preference, but may suggest to put it to the header file to avoid repeated check in other components as whenever using it, here in MLR, or in DUA as well, it should be a similar random generation statement.

> `PARENT_AGGREGATE_DELAY` The longest time period that a Parent Router can wait to collect multicast subscriptions and/or DUA registrations from its MTD Children before registering these to the BBR Server

src/core/thread/mlr_manager.cpp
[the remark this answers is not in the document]
simonlingoogle: > For case 3, ChildUpdateResponse from child only include MLEID.
We should also consider other implementations (e.g. Kirale)? We can't assure that other implementations have same behavior as OpenThread. 

> For case 2, when the child is Restoring, if it is a fragmented keep-alive CUR from child, it may have no chance to go up to MLE layer.

I am not clear about the background. The CUR must be be fragmented in such case?

Moreover, what do you have in mind for case 1? I think we can't assume that `CIR` contains only MLEID.

src/core/thread/mlr_manager.cpp
simonlingoogle: moved

src/core/thread/topology.hpp
simonlingoogle: I will resolve this PR and revise according to https://github.com/openthread/openthread/pull/5236#discussion_r459920104

src/core/thread/mlr_manager.cpp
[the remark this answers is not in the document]
jwhui: Thread requires fragmented MLE messages to enable MAC frame security. However, if the receiving device does not yet have the necessary MAC frame counters (e.g. after reboot), then fragmented MLE messages are dropped.

Implicitly, this means any message that includes a Challenge and/or Response TLV must be small enough to fit within a single IEEE 802.15.4 frame. At the same time, if an Address Registration TLV with multiple IPv6 addresses can fit within a single IEEE 802.15.4 frame, then it can carry more than just the ML-EID.

src/core/thread/mlr_manager.cpp
simonlingoogle: Let's discuss in one place.

src/core/thread/mlr_manager.cpp
simonlingoogle: Let's discuss in one place.

src/core/thread/mlr_manager.cpp
simonlingoogle: Let's discuss in one place.

src/core/thread/mlr_manager.cpp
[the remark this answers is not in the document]
librasungirl: @simonlingoogle 
For case 1, the filter `kInStateValidOrRestoring` itself does already exclude child in `kStateParentResponse` (before receiving CIR) or `kStateChildIdRequest` (after receiving CIR), right?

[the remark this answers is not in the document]
simonlingoogle: > @simonlingoogle
> For case 1, the filter `kInStateValidOrRestoring` itself does already exclude child in `kStateParentResponse` (before receiving CIR) or `kStateChildIdRequest` (after receiving CIR), right?

I believe the Child is mostly likely in state `kStateParentRequest` when the Parent receives a `Child Id Request` (I couldn't find Parent set Child state to `kStateParentResponse` anywhere in the code). The Parent will call `UpdateChildAddresses` in the  `kStateParentRequest` state before transferring it to `kStateChildIdRequest`. You are right that both states are filtered-out by `IsStateValidOrRestoring`. I think this is a potential issue that we need to fix.

src/core/thread/topology.hpp
simonlingoogle: updated accordingly

src/core/thread/mlr_manager.cpp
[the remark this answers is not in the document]
simonlingoogle: Analyzed a Child's state transition: 
image

I believe that the filtered state actually does not matters that much, whether it's `Valid` only or `ValidOrRestoring` . What we need to do is two things:

- When the Child enters the filtered state (`Valid` or `ValidOrRestoring`), we need the check the current multicast addresses and schedule MLR.req if necessary.
- When the Child sends Address Registration TLV (by ChildUpdateRequest or ChildUpdateResponse) in the filtered state, we update the multicast addresses and schedule MLR.req if necessary. 

So, if we filter `Valid` state, then we need to update multicast addresses when a Child's state becomes `kStateValid`. Then, only update multicast addresses in `kStateValid` state. 

If we filter `ValidOrRestoring` state, then we need to update multicast addresses when a Child's state becomes `kStateValid` or `kStateChildUpdateRequest`. Then, update multicast addresses in these two states.  

With this analysis, maybe filtering `Valid` state is a better idea because we already have a clear entrance for a Child to become valid: `MleRouter::SetChildStateToValid` which calls `Signal(OT_NEIGHBOR_TABLE_EVENT_CHILD_ADDED, aChild);`. We can update the initial multicast addresses in these two `SetChildStateToValid` or `Signal`. 

Thoughts? @librasungirl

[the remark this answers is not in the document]
simonlingoogle: Revised accordingly in the last commit. @librasungirl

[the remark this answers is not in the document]
librasungirl: Thanks for your excellent efforts for this minor enhancement. SGTM.

src/core/thread/mlr_manager.cpp
```diff
@@ -81,6 +83,130 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
     UpdateReregistrationDelay(needRereg);
 }
 
+#if OPENTHREAD_CONFIG_MLR_ENABLE
+void MlrManager::UpdateLocalSubscriptions(void)
+{
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    // Check multicast addresses are newly listened against Children
+    for (Ip6::ExternalNetifMulticastAddress &addr : Get<ThreadNetif>().IterateExternalMulticastAddresses())
```
abtink, line 91: Use filter `kTypeMulticastLargerThanRealmLocal`?

src/core/thread/mlr_manager.cpp
```diff
@@ -81,6 +83,130 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
...
+    {
+        if (addr.GetAddress().IsMulticastLargerThanRealmLocal() && addr.GetMlrState() == kMlrStateToRegister &&
+            IsAddressMlrRegisteredByAnyChild(addr.GetAddress()))
+        {
+            addr.SetMlrState(kMlrStateRegistered);
+        }
+    }
+#endif // OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
```
abtink, line 99:
```suggestion
#endif
```

src/core/thread/mlr_manager.cpp
```diff
@@ -81,6 +83,130 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
...
+        }
+    }
+
+exit:
+    return ret;
+}
+
+void MlrManager::SetNetifMulticastAddressMlrState(MlrState aFromState, MlrState aToState)
+{
+    for (Ip6::ExternalNetifMulticastAddress &addr : Get<ThreadNetif>().IterateExternalMulticastAddresses())
```
abtink, line 124: Use filter `kTypeMulticastLargerThanRealmLocal`?

src/core/thread/mlr_manager.cpp
```diff
@@ -81,6 +83,130 @@ void MlrManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State
...
+        }
+
+#if OPENTHREAD_CONFIG_MLR_ENABLE
+        // Check if it's a new multicast address against parent Netif
+        isMlrRegistered = isMlrRegistered || IsAddressMlrRegisteredByNetif(address);
+#endif
+        // Check if it's a new multicast address against other Children
+        isMlrRegistered = isMlrRegistered || IsAddressMlrRegisteredByAnyChildExcept(address, &aChild);
+
+        // Set MLR state accordingly
```
abtink, line 182: Remove comment? (I feel clear from the code).

src/core/thread/mlr_manager.cpp
```diff
@@ -121,19 +247,62 @@ void MlrManager::SendMulticastListenerRegistration(void)
...
     VerifyOrExit(mle.IsAttached(), error = OT_ERROR_INVALID_STATE);
     VerifyOrExit(mle.IsFullThreadDevice() || mle.GetParent().IsThreadVersion1p1(), error = OT_ERROR_INVALID_STATE);
     VerifyOrExit(Get<BackboneRouter::Leader>().HasPrimary(), error = OT_ERROR_INVALID_STATE);
 
-    num = static_cast<uint8_t>(CountNetifMulticastAddressesToRegister());
-    VerifyOrExit(num > 0, error = OT_ERROR_NOT_FOUND);
-    if (num > kIPv6AddressesNumMax)
+#if OPENTHREAD_CONFIG_MLR_ENABLE
+    // Append Netif multicast addresses
+    for (Ip6::ExternalNetifMulticastAddress &addr : Get<ThreadNetif>().IterateExternalMulticastAddresses())
```
abtink, line 260: Use filter for multicast larger than realm local?

src/core/thread/topology.hpp
```diff
@@ -862,6 +862,56 @@ class Child : public Neighbor, public IndirectSender::ChildInfo, public DataPoll
...
+    bool HasMlrRegisteredAddress(const Ip6::Address &aAddress);
+
+    /**
+     * This method returns if the Child has any Ip6 address of MLR state `kMlrStateRegistered`.
+     *
+     * @retval true   If the Child has any Ip6 address of MLR state `kMlrStateRegistered`.
+     * @retval false  If the Child does not have any Ip6 address of MLR state `kMlrStateRegistered`.
+     *
+     */
+    bool HasAnyMlrRegisteredAddress() { return mMlrRegisteredMask.HasAny(); }
```
abtink, line 903:
```suggestion
    bool HasAnyMlrRegisteredAddress(void) { return mMlrRegisteredMask.HasAny(); }
```

src/core/thread/topology.hpp
```diff
@@ -862,6 +862,56 @@ class Child : public Neighbor, public IndirectSender::ChildInfo, public DataPoll
...
+
+    /**
+     * This method returns if the Child has any Ip6 address of MLR state `kMlrStateToRegister`.
+     *
+     * @retval true   If the Child has any Ip6 address of MLR state `kMlrStateToRegister`.
+     * @retval false  If the Child does not have any Ip6 address of MLR state `kMlrStateToRegister`.
+     *
+     */
+    bool HasAnyMlrToRegisterAddress() { return mMlrToRegisterMask.HasAny(); }
```
abtink, line 912:
```suggestion
    bool HasAnyMlrToRegisterAddress(void) { return mMlrToRegisterMask.HasAny(); }
```

src/core/common/bit_vector.hpp
```diff
@@ -95,11 +104,7 @@ template <uint16_t N> class BitVector : public Equatable<BitVector<N>>
      * @param[in] aIndex  The index.
      *
      */
-    void Clear(uint16_t aIndex)
-    {
-        OT_ASSERT(aIndex < N);
-        mMask[aIndex / 8] &= ~(0x80 >> (aIndex % 8));
-    }
+    void Clear(uint16_t aIndex) { Set(aIndex, false); }
```
abtink, line 107: How about removing `Clear()` in light of now having `Set(aIndex, false)`?
I think it'd good to provide one way for this.

src/core/thread/topology.hpp
```diff
@@ -862,6 +862,56 @@ class Child : public Neighbor, public IndirectSender::ChildInfo, public DataPoll
 
 #endif // #if OPENTHREAD_CONFIG_CHILD_SUPERVISION_ENABLE
 
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    /**
+     * This method returns MLR state of an Ip6 multicast address.
+     *
```
abtink, line 868: Let's also add a comment to mention the reference should be one which we get from the iterator.
```suggestion
     * 
     * @note The @p aAdddress reference MUST be from `IterateIp6Addresses()` or `AddressIterator`.
```

src/core/common/bit_vector.hpp
```diff
@@ -78,15 +78,24 @@ template <uint16_t N> class BitVector : public Equatable<BitVector<N>>
...
     /**
-     * This method sets the mask of a given index.
+     * This method sets or clears the mask of a given index.
      *
      * @param[in] aIndex  The index.
+     * @param[in] aSet    TRUE to sets the mask, or FALSE to clear the mask.
      *
      */
-    void Set(uint16_t aIndex)
+    void Set(uint16_t aIndex, bool aSet = true)
```
jwhui, line 87: There are two models here: (1) `Set()` with a param for the value and (2) `Set()`/`Clear()`. Option 1 here would seem to add code for those cases where the caller does not conditionally call `Set()` vs. `Clear()`. However, it may not make any difference if the caller always has its own if-statement. Thoughts?

src/core/thread/mlr_manager.hpp
```diff
@@ -113,8 +130,34 @@ class MlrManager : public InstanceLocator, public Notifier::Receiver
                                                      const Ip6::MessageInfo *aMessageInfo,
                                                      otError                 aResult);
 
-    uint16_t CountNetifMulticastAddressesToRegister(void) const;
-    void     SetNetifMulticastAddressMlrState(MlrState aFromState, MlrState aToState);
+#if OPENTHREAD_CONFIG_MLR_ENABLE
+    void UpdateLocalSubscriptions(void);
+    void SetNetifMulticastAddressMlrState(MlrState aFromState, MlrState aToState);
+    bool IsAddressMlrRegisteredByNetif(const Ip6::Address &aAddress);
```
jwhui, line 136: nit: declare as `const`?
```suggestion
    bool IsAddressMlrRegisteredByNetif(const Ip6::Address &aAddress) const;
```

src/core/thread/mlr_manager.hpp
```diff
@@ -113,8 +130,34 @@ class MlrManager : public InstanceLocator, public Notifier::Receiver
...
+#endif
+
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    void SetChildMulticastAddressMlrState(MlrState aFromState, MlrState aToState);
+    bool IsAddressMlrRegisteredByAnyChild(const Ip6::Address &aAddress)
```
jwhui, line 141: nit: declare as `const`?
```suggestion
    bool IsAddressMlrRegisteredByAnyChild(const Ip6::Address &aAddress) const
```

src/core/thread/mlr_manager.hpp
```diff
@@ -113,8 +130,34 @@ class MlrManager : public InstanceLocator, public Notifier::Receiver
...
+    {
+        return IsAddressMlrRegisteredByAnyChildExcept(aAddress, nullptr);
+    }
+    bool IsAddressMlrRegisteredByAnyChildExcept(const Ip6::Address &aAddress, const Child *aExceptChild);
```
jwhui, line 145: nit: declare as `const`?
```suggestion
    bool IsAddressMlrRegisteredByAnyChildExcept(const Ip6::Address &aAddress, const Child *aExceptChild) const;
```

src/core/thread/mlr_manager.hpp
```diff
@@ -86,6 +88,21 @@ class MlrManager : public InstanceLocator, public Notifier::Receiver
     void HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State               aState,
                                            const BackboneRouter::BackboneRouterConfig &aConfig);
 
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    /**
+     * This method updates the Multicast Subscription Table according to the Child information.
+     *
+     * @param[in]  aChild                       A reference to the child information.
+     * @param[in]  aOldMlrRegisteredAddresses   A pointer to an array of the Child's previously registered Ip6
+     * addresses.
```
jwhui, line 97: nit: alignment

src/core/thread/topology.hpp
```diff
@@ -862,6 +862,56 @@ class Child : public Neighbor, public IndirectSender::ChildInfo, public DataPoll
...
+    /**
+     * This method returns if the Child has Ip6 address @p aAddress of MLR state `kMlrStateRegistered`.
+     *
+     * @param[in] aAddress  The Ip6 address.
+     *
+     * @retval true   If the Child has Ip6 address @p aAddress of MLR state `kMlrStateRegistered`.
+     * @retval false  If the Child does not have Ip6 address @p aAddress of MLR state `kMlrStateRegistered`.
+     *
+     */
+    bool HasMlrRegisteredAddress(const Ip6::Address &aAddress);
```
jwhui, line 894: nit: declare as `const`?
```suggestion
    bool HasMlrRegisteredAddress(const Ip6::Address &aAddress) const;
```

src/core/thread/topology.hpp
```diff
@@ -862,6 +862,56 @@ class Child : public Neighbor, public IndirectSender::ChildInfo, public DataPoll
...
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    /**
+     * This method returns MLR state of an Ip6 multicast address.
+     *
+     * @param[in] aAddress  The Ip6 multicast address.
+     *
+     * @returns MLR state of the Ip6 multicast address.
+     *
+     */
+    MlrState GetAddressMlrState(const Ip6::Address &aAddress);
```
jwhui, line 874: nit: declare as `const`?
```suggestion
    MlrState GetAddressMlrState(const Ip6::Address &aAddress) const;
```

src/core/thread/topology.hpp
jwhui, line 903: nit: declare as `const`?
```suggestion
    bool HasAnyMlrRegisteredAddress(void) const { return mMlrRegisteredMask.HasAny(); }
```

src/core/thread/topology.hpp
jwhui, line 912: nit: declare as `const`?
```suggestion
    bool HasAnyMlrToRegisterAddress(void) const { return mMlrToRegisterMask.HasAny(); }
```

src/core/thread/topology.hpp
```diff
@@ -892,6 +944,10 @@ class Child : public Neighbor, public IndirectSender::ChildInfo, public DataPoll
     uint8_t                  mNetworkDataVersion;           ///< Current Network Data version
```
jwhui, line 944: nit: move this down to avoid additional padding?

src/core/common/bit_vector.hpp
simonlingoogle: It makes the caller code simpler in some cases, as in this PR. 

If the caller has its own if-statement, it can call `Set()` and `Clear()` for different conditions. 

I agree it is a little bit confusing to have `Set(false)` and `Clear()` do the same job. 
Maybe we should remove `Clear()`  and always use `Set(bool)`. If so, we can let BitVector inherit Clearable and use `Clear()` instead of `ClearAll()`. 
Thoughts? @jwhui @abtink @librasungirl

src/core/common/bit_vector.hpp
simonlingoogle: We can let BitVector inherit Clearable again and use Clear(void) instead of ClearAll(void). Thoughts?

src/core/common/bit_vector.hpp
simonlingoogle: I think it's too much change for `BitVector` in this PR. 
I will revert the BitVector to have separate `Set()` and `Clear()` in this PR. 

But let's continue discussing on this subject here to find out the best solution. 
Thoughts?

jwhui: Moving the `BitVector()` change to a separate PR sounds good. It will allow us to better evaluate the pros/cons of the change.

My only concern with having a single `Set()` method is that this copies the code everywhere (even in cases where the conditional is not necessary). Is this true given that the methods are implemented in a header? Or will the compiler properly optimize this?

simonlingoogle: I think it's fine if `Set(true)` and `Set(false)` is as efficient as current `Set()` and `Clear()` (optimized by compiler).
We can see from `size-report` if this is too in the BitVector refactoring PR.

src/core/common/bit_vector.hpp
simonlingoogle: Will resolve it in a separate PR.

PR #5241, opened 2020-07-15
[size-report] set Thread version to 1.2

simonlingoogle:


script/check-size
```diff
@@ -126,6 +126,7 @@ size_nrf52840()
         "SNTP_CLIENT=1"
         "TIME_SYNC=1"
         "UDP_FORWARD=1"
+        "THREAD_VERSION=1.2"
```
jwhui, line 129: I wonder if it would be worthwhile to have separate size reports for Thread 1.1 and 1.2. Thoughts?

/cc/ @simonlingoogle @bukepo @abtink

jwhui, line 129: The size report is not showing any differences. Is that expected?

simonlingoogle: I misunderstood how `check-size` works. 
Turn out with this PR, it's comparing 1.2 with 1.2, so no change at all.

simonlingoogle: I misunderstood how `check-size` works. 
Turn out with this PR, it's comparing 1.2 with 1.2, so no change at all. 
Maybe we should close this PR since it's useless.

simonlingoogle: I think it’s not very useful unless we enhance size-report. 
Most PR contains changes to either 1.1 features or 1.2 features, but not both. For these PRs, it is not useful to print the two reports. 
- 1.1 PRs will have same size increases in both reports 
- 1.2 PRs will have 0 size increases in 1.1 report

From this point, I think it’s better to have 1.2 features turned on for size-report to support 1.2 features. 

However I think it’s valuable if we can show sizes and increases of different 1.2/1.1 components, e.g. Bbr, Dua, Mlr, Csl, so that we know the cost of each features.

jwhui: I was hoping to catch situations where adding Thread 1.2 code (intentionally or unintentionally) changes Thread 1.1 code. Should we flag situations where adding a Thread 1.2 feature significantly increases Thread 1.1 code?

PR #5242, opened 2020-07-15
[test] move expect timeout settings to _common.exp

Moandor-y:
This moves all settings of `timeout` to one place. Also changes the default timeout to 10 seconds, can probably solve the occasional timeout when `dataset commit active`.

### DIFF
--- a/tests/scripts/expect/_common.exp
+++ b/tests/scripts/expect/_common.exp
@@ -29,6 +29,7 @@
 
 proc wait_for {command expected} {
     set result 0
+    set timeout 1
     for {set i 0} {$i < 20} {incr i} {
         if {$command != ""} {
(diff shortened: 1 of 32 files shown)


tests/scripts/expect/_common.exp
```diff
@@ -29,6 +29,7 @@
 
 proc wait_for {command expected} {
     set result 0
+    set timeout 1
```
simonlingoogle, line 32: need to restore timeout to 10s when this proc exit?

Moandor-y: No. The variable will be scoped to the function.

simonlingoogle: Thanks

PR #5246, opened 2020-07-16
[thread-cert] deprecate certification test case 5.5.8, 5.6.8, 5.8.1

FiveDimensions:
test case 5.5.8, 5.6.8 and 5.8.1 were deprecated in Thread Certification testplan


simonlingoogle:
Could there be any value to still keep them?

PR #5247, opened 2020-07-16
[legacy] move legacy api to core include

gjc13:
This allow users to use legacy apis without ncp mode enabled.


include/openthread/legacy.h
```diff
@@ -0,0 +1,115 @@
...
+ */
+void otLegacyInit(void);
+
+/**
+ * This method starts the legacy processing.
+ *
+ * Will be called when Thread network is enabled.
+ *
+ */
+void otLegacyStart(void);
```
abtink, line 68: I don't think we want to add these as OT APIs (under `openthread/legacy.h`). OT APIs are for all functionality that is supported by OT core itself (these are not) and APIs provided for others to use. 

These are more like platform API, which OT declares and expects platform layer (or some other layer) to provide an implementation and then OT code would use these APIs

include/openthread/legacy.h
```diff
@@ -0,0 +1,115 @@
...
+ *
+ */
+
+#define OT_LEGACY_ULA_PREFIX_LENGTH 8 ///< Legacy ULA size (in bytes)
+
+/**
+ * This method initializes the legacy processing.
+ *
+ */
+void otLegacyInit(void);
```
abtink, line 60: @gjc13  can you please provide more background on this whole PR? Thanks.

include/openthread/legacy.h
gjc13: Done

include/openthread/legacy.h
abtink: @gjc13 reached out to explain the reason for moving these definitions is to allow users/projects that don't go through NCP interface (i.e. want to control through OT APIs directly) can use the same legacy APIs as in NCP case.

Thinking more about this, I feel we may not want to include these in OT directly. 
In this case (where NCP interface is not being used by the user of OT i.e., project/product integrator): 
- We are declaring a bunch of new APIs for some external module to provide.
- They are not used by core of OT.
- We expect user of OT to directly call them to use them (from the app/next layer above OT).

We are unnecessarily acting as a middle-man here. No need for OT to be aware of these APIs in this case and they can be fully defined and used by the next layer and user of OT, i.e. project/product/platform related code.

In case the user wants the code to run along with OT, we provide "extension" mechanism, but in this case, I don't think user needs to add these as extension.

Regarding NCP use, we are limited in that `npc` code it the interface for users to issue commands to OT.

Thoughts?

gjc13: Correct. We can directly call those extensions in integration code.

PR #5248, opened 2020-07-16
[mlr] send MLR.req for local multicast addresses

simonlingoogle:
In order to make CR easier, the scope of this PR is limited to:
- [x] Adding compile configurations necessary for MLR features
- [x] Adding source/header files necessary for MLR features
- [x] Send `MLR.req` for **local** MAs on `ThreadNetif`
  - [x] At most one `MLR.req` is outstanding
  - [x] Reregistration according to PBBR configuration
  - [x] Renewing periods according to PBBR configuration
- [x] BBRs receive `MLR.req` with minimal processing and send `MLR.rsp`
- [x] Add a test to verify that MLR.req is sent for local MAs.

Please note that:
- Parent send MLR.req in behave of Children is implemented in another PR #5236 .

**MLR PR List:**
- #5248 Send MLR.req for locally Netfi MAs **<== This PR**
- #5236 Parent send MLR.req on behave of Children 
- ~~#5250 Add the `MLR` log region **MERGED**~~


src/core/thread/mlr_manager.cpp
```diff
@@ -0,0 +1,406 @@
...
+        ScheduleSend(0);
+    }
+    else
+    {
+        otBackboneRouterConfig config;
+        uint16_t               reregDelay;
+
+        SetNetifMulticastAddressMlrState(kMulticastAddressFilterRegistering, kMlrStateToRegister);
+
+        // Schedule reregistration if failed in t < Reregistration Delay
```
librasungirl, line 244: Thread 1.2 Specification 5.24.2.3 should be the source for the test requirement.
update the comment?

src/core/thread/mlr_manager.cpp
```diff
@@ -0,0 +1,406 @@
...
+        messageInfo.GetPeerAddr().SetToRoutingLocator(mle.GetMeshLocalPrefix(),
+                                                      Get<BackboneRouter::Leader>().GetServer16());
+    }
+
+    messageInfo.SetPeerPort(kCoapUdpPort);
+    messageInfo.SetSockAddr(mle.GetMeshLocal16());
+
+    SuccessOrExit(error = Get<Coap::Coap>().SendMessage(
+                      *message, messageInfo, &MlrManager::HandleMulticastListenerRegistrationResponse, this));
+
```
librasungirl, line 188: Enable fast poll if the rx-off device has stoped CSL? thus avoiding the delay to receive MLR.rsp and possible unnecessary COAP retransmission?

src/core/thread/mlr_manager.cpp
```diff
@@ -0,0 +1,406 @@
...
+exit:
+    if (error == OT_ERROR_NONE)
+    {
+        mMlrPending = true;
```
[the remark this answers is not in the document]
simonlingoogle: I think it's better to put it after `exit` like this because it makes sure that `mMlrPending` is always correctly set to `true`. 
Currently it's fine for this function to put `mMlrPending = true;` before `exit:`. But if we add any statement like `VerifyOrExit(..., OT_NOOP)` or `ExitNow()` in this function, we are going to incorrectly skip setting `mMlrPending = true;`

src/core/thread/mlr_manager.cpp
```diff
@@ -0,0 +1,406 @@
...
+}
+
+void MlrManager::HandleNotifierEvents(Notifier::Receiver &aReceiver, Events aEvents)
+{
+    static_cast<MlrManager &>(aReceiver).HandleNotifierEvents(aEvents);
+}
+
+void MlrManager::HandleNotifierEvents(Events aEvents)
+{
+    if (aEvents.ContainsAny(kEventIp6MulticastSubscribed | kEventIp6MulticastUnsubscribed))
```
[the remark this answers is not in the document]
simonlingoogle: It used to be necessary because we needed to update the table. 
Now it's not necessary any more.

src/core/backbone_router/bbr_manager.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include "bbr_manager.hpp"
+
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+
+#include "common/code_utils.hpp"
+#include "common/instance.hpp"
+#include "common/locator-getters.hpp"
+#include "common/logging.hpp"
+#include "common/random.hpp"
+#include "thread/mle_types.hpp"
```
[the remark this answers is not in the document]
simonlingoogle: I think it's ordered by clang-format.

src/core/backbone_router/bbr_manager.cpp
```diff
@@ -0,0 +1,119 @@
...
+            Get<Coap::Coap>().AddResource(mMulticastListenerRegistration);
+        }
+    }
+}
+
+void Manager::HandleMulticastListenerRegistration(const Coap::Message &aMessage, const Ip6::MessageInfo &aMessageInfo)
+{
+    bool                       isPrimary = Get<BackboneRouter::Local>().IsPrimary();
+    ThreadStatusTlv::MlrStatus status    = ThreadStatusTlv::kMlrSuccess;
+
```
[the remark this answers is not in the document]
simonlingoogle: revised

src/core/thread/mlr_manager.cpp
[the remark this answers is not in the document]
simonlingoogle: fixed.

src/core/thread/mlr_manager.cpp
simonlingoogle: added, please check

script/check-size
```diff
@@ -120,11 +120,13 @@ size_nrf52840()
         "JOINER=1"
         "LINK_RAW=1"
         "MAC_FILTER=1"
+        "MLR=1"
         "MTD_NETDIAG=1"
         "SERVICE=1"
         "SLAAC=1"
         "SNTP_CLIENT=1"
         "TIME_SYNC=1"
+        "THREAD_VERSION=1.2"
```
jwhui, line 129: Continuing the discussion from https://github.com/openthread/openthread/pull/5241#discussion_r455884634:

> I was hoping to catch situations where adding Thread 1.2 code (intentionally or unintentionally) changes Thread 1.1 code. Should we flag situations where adding a Thread 1.2 feature significantly increases Thread 1.1 code?

src/core/common/instance.hpp
```diff
@@ -745,6 +750,13 @@ template <> inline DuaManager &Instance::Get(void)
 }
 #endif
 
+#if (OPENTHREAD_MTD || OPENTHREAD_FTD) && OPENTHREAD_CONFIG_MLR_ENABLE
+template <> inline MlrManager &Instance::Get(void)
+{
+    return mThreadNetif.mMlrManager;
+}
+#endif // (OPENTHREAD_MTD || OPENTHREAD_FTD) && OPENTHREAD_CONFIG_MLR_ENABLE
```
jwhui, line 758: nit: do not need closing comment for small blocks
```suggestion
#endif
```

src/core/net/netif.hpp
```diff
@@ -152,8 +153,16 @@ class NetifMulticastAddress : public otNetifMulticastAddress,
         return static_cast<NetifMulticastAddress *>(const_cast<otNetifMulticastAddress *>(mNext));
     }
 
+#if (OPENTHREAD_FTD || OPENTHREAD_MTD) && OPENTHREAD_CONFIG_MLR_ENABLE
+    MlrState GetMlrState() const { return mMlrState; }
```
jwhui, line 157:
```suggestion
    MlrState GetMlrState(void) const { return mMlrState; }
```

src/core/thread/mlr_types.hpp
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @{
+ *
+ */
+
+/**
+ * Multicast Listener Registration state for multicast addresses.
+ *
+ */
+typedef enum MlrState
```
jwhui, line 54: nit: prefer to define explicit type

src/core/thread/mlr_manager.cpp
```diff
@@ -0,0 +1,392 @@
...
+    , mTimer(aInstance, MlrManager::HandleTimer, this)
+    , mReregistrationDelay(0)
+    , mSendDelay(0)
+    , mMlrPending(false)
+{
+}
+
+void MlrManager::HandleNotifierEvents(Notifier::Receiver &aReceiver, Events aEvents)
+{
+    static_cast<MlrManager &>(aReceiver).HandleNotifierEvents(aEvents);
```
jwhui, line 61: Inline in `mlr_manager.hpp` to be consistent with `HandleMulticastListenerRegistrationResponse()`?

src/core/thread/mlr_manager.cpp
```diff
@@ -0,0 +1,392 @@
...
+        {
+            auto state = addr->GetMlrState();
+
+            otLogDebgMlr("%-32s%c", addr->GetAddress().ToString().AsCString(), "-rR"[state]);
+        }
+    }
+#endif
+}
+
+size_t MlrManager::CountNetifMulticastAddressesToRegister()
```
jwhui, line 360:
```suggestion
size_t MlrManager::CountNetifMulticastAddressesToRegister(void) const
```

src/core/thread/mlr_manager.hpp
```diff
@@ -0,0 +1,135 @@
...
+    {
+        static_cast<MlrManager *>(aContext)->HandleMulticastListenerRegistrationResponse(
+            static_cast<Coap::Message *>(aMessage), static_cast<const Ip6::MessageInfo *>(aMessageInfo), aResult);
+    }
+
+    void HandleMulticastListenerRegistrationResponse(Coap::Message *         aMessage,
+                                                     const Ip6::MessageInfo *aMessageInfo,
+                                                     otError                 aResult);
+
+    size_t CountNetifMulticastAddressesToRegister();
```
jwhui, line 113:
```suggestion
    size_t CountNetifMulticastAddressesToRegister(void) const;
```

src/core/thread/mlr_manager.cpp
```diff
@@ -0,0 +1,392 @@
...
+{
+    size_t count = 0;
```
jwhui, line 362: Just curious, why `size_t` and not something like `uint8_t` or `uint16_t`?

script/check-size
abtink: I fully agree with this. Having a 1.2 version size vs  1.1 version would be useful (to see the impact of 1.2 features on 1.1 code size).

src/core/net/netif.hpp
abtink, line 157: nit: missing documentation for the new method?

src/core/net/netif.hpp
```diff
@@ -152,8 +153,16 @@ class NetifMulticastAddress : public otNetifMulticastAddress,
...
+    void     SetMlrState(MlrState aState) { mMlrState = aState; }
+#endif
 private:
     bool Matches(const Address &aAddress) const { return GetAddress() == aAddress; }
+
+#if (OPENTHREAD_FTD || OPENTHREAD_MTD) && OPENTHREAD_CONFIG_MLR_ENABLE
+    MlrState mMlrState : 2;
+#endif
```
abtink, line 165: We are adding new member variable in `NetifMulticastAddress` class which is inheriting from `otNetifMulticastAddress`

We should be careful that we are not getting an `otNetifMulticastAddress` object (not really being `NetifMulticastAddress`) and then down-casting/treating it as `NetifMulticastAddress`.  I think this pattern is not safe. 

I know we do  have some fixed/constant multicast addresses (like "link-local all nodes", "link-local all routers", etc.) which are defined as constant `otNetifMulticastAddress` objects (and stored in flash) - defined in `netif.cpp` source file.

I undestand that the way `mMleState` is being used later in the code (we do check if it is external address), it may work ok but I think using this pattern may set a bad precedence. 

So I feel we may be better not add `mMleState` to the  `NetifMulticastAddress` class.

Thoughts?

src/core/net/netif.hpp
```diff
@@ -313,6 +322,7 @@ class Netif : public InstanceLocator, public LinkedListEntry<Netif>, private Non
      *
      */
     const NetifMulticastAddress *GetMulticastAddresses(void) const { return mMulticastAddresses.GetHead(); }
+    NetifMulticastAddress *      GetMulticastAddresses(void) { return mMulticastAddresses.GetHead(); }
```
abtink, line 325: (Same as above) we do have some multicast entries in the multicast list which are literally `const` (stored in flash), again I know that the current use is checking for this and is ok, but providing a method that allows user to get non-const `NetifMulticastAddress` seems unsafe (allows possible incorrect use in future).

src/core/thread/mlr_manager.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+void MlrManager::HandleTimer(void)
+{
+    if (mSendDelay > 0)
+    {
+        if (--mSendDelay == 0)
+        {
+            SendMulticastListenerRegistration();
+        }
+    }
```
abtink, line 273: How about:
```suggestion
    if ((mSendDelay > 0) && (--mSendDelay == 0))
    {
        SendMulticastListenerRegistration();
    }
```

src/core/thread/mlr_manager.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+    if (mReregistrationDelay > 0)
+    {
+        if (--mReregistrationDelay == 0)
+        {
+            Reregister();
+        }
+    }
```
abtink, line 281:
```suggestion
    if ((mReregistrationDelay > 0) && (--mReregistrationDelay == 0))
    {
        Reregister();
    }
```

src/core/thread/mlr_manager.cpp
```diff
@@ -0,0 +1,392 @@
...
+    bool needSendMlr = (mle.IsFullThreadDevice() || mle.GetParent().IsThreadVersion1p1()) &&
+                       Get<BackboneRouter::Leader>().HasPrimary();
+
+    if (!needSendMlr)
+    {
+        mReregistrationDelay = 0;
+    }
+    else
+    {
+        otBackboneRouterConfig config;
```
abtink, line 308: How about using `BackboneRouter::BackboneRouterConfig` ?

src/core/thread/mlr_manager.cpp
```diff
@@ -0,0 +1,392 @@
...
+#if OPENTHREAD_CONFIG_LOG_BBR && OPENTHREAD_CONFIG_LOG_LEVEL >= OT_LOG_LEVEL_DEBG
+    otLogDebgMlr("-------- Multicast Addresses --------");
+
+    for (const Ip6::NetifMulticastAddress *addr = Get<ThreadNetif>().GetMulticastAddresses(); addr != nullptr;
+         addr                                   = addr->GetNext())
+    {
+        if (Get<ThreadNetif>().IsMulticastAddressExternal(*addr) &&
+            addr->GetAddress().IsMulticastLargerThanRealmLocal())
+        {
+            auto state = addr->GetMlrState();
```
abtink, line 352: I suggest  we limit use of `auto` (use it only in cases where it adds clear benefit). 
```suggestion
            MlrState state = addr->GetMlrState();
```
Thoughts?

src/core/net/netif.hpp
abtink: Thinking about this I can suggest a possible different approach:
- We keep `NetifMultciastAddress` as is.
- We define a sub-class of it which then adds the new member variable and getter/setter for this (maybe calling it `ExternalNetifMultciastAddress` or `MlrNetifMultciastAddress`?)
- We can then add a new method to `Netif` to iterate through all external/MLR multicast addresses (only external/MLR ones) and return a pointer/reference to new sub-class type.
- User of this method can then safely get/set `MleState`.

Thoughts?

script/check-size
simonlingoogle: But it's beyond the scope of this PR. Shall we just turn off size check for this PR?

src/core/thread/mlr_manager.cpp
simonlingoogle: uint16_t should be more than enough.

src/core/thread/mlr_manager.cpp
simonlingoogle: I agree with you.

src/core/net/netif.hpp
simonlingoogle: > I know we do have some fixed/constant multicast addresses (like "link-local all nodes", "link-local all routers", etc.) which are defined as constant `otNetifMulticastAddress` objects (and stored in flash) - defined in `netif.cpp` source file.

Implementing a link list with some elements being constant is challenging. Do we make sure that the constant `otNetifMulticastAddress`s are always at the tail of the list? @abtink

simonlingoogle: > * We define a sub-class of it which then adds the new member variable and getter/setter for this (maybe calling it `ExternalNetifMultciastAddress` or `MlrNetifMultciastAddress`?)

I actually thought of it too. It should even save us some memory (MlrState on constant `otNetifMulticastAddress` is redundant).

src/core/thread/mlr_types.hpp
simonlingoogle: Defining it to explicit type `uint8_t` could lead to compile issues:

```
2020-07-21T02:24:12.9702646Z /tmp/ot-size-report/b/examples/../src/core/net/netif.hpp:177:26: error: 'ot::Ip6::NetifMulticastAddress::mMlrState' is too small to hold all values of 'ot::MlrState {aka enum ot::MlrState}' [-Werror]
2020-07-21T02:24:12.9702794Z      MlrState mMlrState : 2;
```

Same issue happened to one of @librasungirl 's PR too.

src/core/net/netif.hpp
abtink: > Implementing a link list with some elements being constant is challenging. Do we make sure that the constant otNetifMulticastAddresss are always at the tail of the list? @abtink 
 
Indeed. They are always at the tail of the list. This whole model is used mainly to reduce RAM use.

https://github.com/openthread/openthread/blob/ceb63f97d57a42f15359c2fb1a6b912c2a963221/src/core/net/netif.cpp#L45-L55
You can check the implementation of `SubscribeAllNodesMulticast()` and `SubscribeAllRoutersMulticast()`, etc for more details (there are comments to help make the code easier to read and follow).

librasungirl: Another thought is that we may define one MlrState array with the size same as `mExtMulticastAddressPool`, per MA that requires MLR.req, we can get the index, and use the index to sync the MlrState correspondingly.

Comparatively, this option might bring less code changes on the multicast address class between different Thread versions, though  the subclass option is more C++ style.

abtink: This can work as well but we need to define the `MlrState` array in the `Netif` itself, since the fact that we use an array/pool for external addresses and therefore can have an index associated with each address  is internal to `Netif` class (not part of `public` definitions of `Netif`).

Thinking more about this, I feel having a way to iterate over the external (only) addresses would be useful in general. We can use this in some place in the code (e.g., in `Mle::AppendAddressRegistration()` where we go through the list of multicast address but do skip over entries that are not external).

- We can define `ExtNetifMulticastAddress` as sub-class of `NetifMulticastAddress`.
- We can overload the `GetNext()` method in the sub-class `ExtNetifMulticastAddress` to get the next external address entry (basically have `GetNext()` implementation skip over linked-list entries that are not external).
- This would be simple for user of `Netif`, they can the head of a linked-list of all `NetifMulticastAddress` entries or the head of linked-list of `ExtNetifMulticastAddress` and in both cases, they iterate through it like a linked-list.

Thoughts?

simonlingoogle: My opinion on the two solutions: 

> 1# Add `ExternalNetifMulticastAddress` as a new subclass of `NetifMulticastAddress` with `MlrState` being it's member. 
> 2# Add an of MlrState with equal size of  mExtMulticastAddressPool.

First of all, I think they are similar solutions without much practical difference. 
- 1# is more object oriented by encapsulating MlrState in the `ExternalNetifMulticastAddress` class, but it requires significant code refactoring. For example,  `ExternalNetifMulticastAddress` will be put to a `LinkedList<ExternalNetifMulticastAddress>` but `ExternalNetifMulticastAddress.GetNext()` returns a `NetifMulticastAddress`, which violates the protocol of `LinkedList`. 
- 2# introduces much less change to existing code but it assumes that a unique index can be get for each external multicast address (e.g. allocated from a array of a Pool). 2# also allows using a bit-vector with 2 bits for each address to further reduce the memory consumption, even though the benefit is minor (e.g. 8 multicast addresses consumes 8B in solution 1# but only 2B in solution 2#). 

Overall I prefer 2# for a little bit because it involves less code change. 

Thoughts? @abtink @librasungirl @jwhui

src/core/common/instance.hpp
```diff
@@ -745,6 +750,13 @@ template <> inline DuaManager &Instance::Get(void)
 }
 #endif
 
+#if (OPENTHREAD_MTD || OPENTHREAD_FTD) && OPENTHREAD_CONFIG_MLR_ENABLE
```
librasungirl, line 753: nit: remove `(OPENTHREAD_MTD || OPENTHREAD_FTD)`

This is within `#if OPENTHREAD_MTD || OPENTHREAD_FTD` block already.

https://github.com/openthread/openthread/blob/3a0569e96542c8a428ee13ba4f7961099f316d42/src/core/common/instance.hpp#L410-L769

src/core/net/netif.hpp
simonlingoogle: > * We can overload the `GetNext()` method in the sub-class `ExtNetifMulticastAddress` to get the next external address entry (basically have `GetNext()` implementation skip over linked-list entries that are not external).

Overloading should be necessary, but skipping over non-external entries seems unnecessary. `ExtNetifMulticastAddress.GetNext()` is only used for the free list of the pool, so when it's called, the next entry **MUST** be an ExtNetifMulticastAddress entry or nullptr.

src/core/thread/mlr_manager.hpp
```diff
@@ -0,0 +1,138 @@
...
+ * @file
+ *   This file includes definitions for managing Multicast Listener Registration feature defined in Thread 1.2.
+ */
+
+#ifndef MLR_MANAGER_HPP_
+#define MLR_MANAGER_HPP_
+
+#include "openthread-core-config.h"
+
+#if (OPENTHREAD_FTD || OPENTHREAD_MTD) && OPENTHREAD_CONFIG_MLR_ENABLE
```
librasungirl, line 39: nit: `(OPENTHREAD_FTD || OPENTHREAD_MTD) && ` could be removed as well

src/core/net/netif.hpp
abtink: @simonlingoogle , all options are fine for me. 

> It requires significant code refactoring. For example, `ExternalNetifMulticastAddress` will be put to a `LinkedList<ExternalNetifMulticastAddress>` but `ExternalNetifMulticastAddress.GetNext()` returns a NetifMulticastAddress, which violates the protocol of LinkedList.

What I had in mind  was not to create a new actual linked-list of `LinkedList<ExternalNetifMulticastAddress>`. I agree that would not be good and would require a lot of changes.

The idea was that we really have one actual linked-list in memory which is a list of all `NetifMulticastAddress` objects. We define a new sub-class `ExtNetifMulticastAdderss` where we just overload the `GetNext()` to go through the same linked-list but skip over all entries that are not external (i.e., make it look like as if the external addresses are in a linked-list of their own).

 But considering how to implement this, I realized it may not be as easy as I initially thought - main thing is that we cannot tell in `(Ext)NetifMulticastAdderss` class itself whether an entry is external or not. For this, we need to have have access to `Netif` class - which we can get if we make `ExtNetifMulticastAddress` be an `InstanceLocatorInit` (all doable  but requires more changes).

src/core/thread/mlr_manager.hpp
simonlingoogle: you mean that without FTD or MTD, MlrManager is not going to be built?

src/core/net/netif.hpp
simonlingoogle: I pushed in the last commit an implementation of `ExternalNetifMulticastAddress` class (actually I have got this code finished before @librasungirl suggested the alternative solution). @abtink @librasungirl 

It allows us to use for-loops like this to iterate all external multicast addresses:
```c++
    for (Ip6::ExternalNetifMulticastAddress &addr : Get<ThreadNetif>().IterateExternalMulticastAddresses()) { ... }
```
I think this code can be useful for other purposes too.

src/core/net/netif.hpp
simonlingoogle: removed

tests/scripts/thread-cert/v1_2_test_multicast_listener_registration.py
```diff
@@ -0,0 +1,400 @@
...
+     |     /   \                       |
+     |    /     \                      |
+   ROUTER_1_2  ROUTER_1_1
+    |     |  \                         |
+    |     |   \                        |
+  SED_1  MED_1 FED_1
+"""
+
+
+class TestDomainUnicastAddress(thread_cert.TestCase):
```
librasungirl, line 69: nit: rename

src/core/net/netif.hpp
```diff
@@ -164,8 +206,141 @@ class Netif : public InstanceLocator, public LinkedListEntry<Netif>, private Non
...
+            const NetifMulticastAddress *addr = mNetif.GetMulticastAddresses();
+
+            while (addr != nullptr && !aNetif.IsMulticastAddressExternal(*addr))
+            {
+                addr = addr->GetNext();
+            }
+
+            mMulticastAddress =
+                const_cast<ExternalNetifMulticastAddress *>(static_cast<const ExternalNetifMulticastAddress *>(addr));
+        }
```
abtink, line 240: This is not a small method, do we want to keep it inlined or maybe move it to `cpp` file?

abtink: This seems to share some code with `Advance()`... how about if we refactor as follows:

We can add a common method like `Update(bool mShouldAdavce)`. If `mShouldAdvance` is set to true,  we first advances the pointer (`const NetifMulticastAddress *addr = mMulticastAddress->GetNext()`  before searching for the next external address (the `while` loop).

Thoughts?

src/core/net/netif.hpp
```diff
@@ -164,8 +206,141 @@ class Netif : public InstanceLocator, public LinkedListEntry<Netif>, private Non
...
+
+    public:
+        /**
+         * This constructor initializes an `ExternalMulticastAddressIterator` instance to start from the first external
+         * multicast address.
+         *
+         * @param[in] aNetif  A reference to the Netif instance.
+         *
+         */
+        explicit ExternalMulticastAddressIterator(Netif &aNetif)
```
abtink, line 228: How about defining `ExternalMulticastAddressIterator` as an `InstanceLocator` and passing in `Instance` instead of `Netif`?

script/check-size
jwhui: I agree this should be in a separate PR. Can we take a look at what it would take to support size reports for both 1.1 and 1.2?

/cc/ @bukepo @librasungirl

simonlingoogle: One thing I can imagine is that the size-report will be 2x longer if we simply add 1.2 results under 1.1 results. 

I think a better idea might be put 1.2 columns after current 1.1 columns:
- Remove the `total` column to save space for 3 more columns. (I think summing up `text` increase and `memory` increases does not make much sense. )
- Add `text`, `data`, `bss` columns for 1.2 to the right of 1.1 columns.

src/core/net/netif.hpp
```diff
@@ -156,6 +157,47 @@ class NetifMulticastAddress : public otNetifMulticastAddress,
     bool Matches(const Address &aAddress) const { return GetAddress() == aAddress; }
 };
 
+class ExternalNetifMulticastAddress : public NetifMulticastAddress
+{
+    friend class Netif;
+    friend class LinkedList<ExternalNetifMulticastAddress>;
```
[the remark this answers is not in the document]
simonlingoogle: `LinkedList<ExternalNetifMulticastAddress>` is necessary not because we have an extra list of `ExternalNetifMulticastAddress`, but because of the fact that **Pool<ExternalNetifMulticastAddress> uses a free list of `ExternalNetifMulticastAddress` for allocation**.

src/core/net/netif.hpp
simonlingoogle: A method we usually use might be setting `mMulticastAddress` to the first `NetifMulticastAddress` on the linked list and call `Advance()` if it's not external. 
I think this should work well only for the fact that we are assigning a `NetifMulticastAddress` to `mMulticastAddress` (even thought we never access fields not of `NetifMulticastAddress` ).
Do you think we should use this method instead?

src/core/net/netif.hpp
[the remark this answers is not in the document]
abtink: I see. Makes sense. Thanks.

Is that why we also need the `GetNext()` implementation (to cast the `mNext` to `ExtenralMultiAddr` in this class?

src/core/net/netif.hpp
simonlingoogle: I tried `InstanceLocator` initially, but changed to `Netif &aNetif`. 

The problem with iterator being `InstanceLocator` is that we can't write `Get<Netif>()`, but we can only use `Get<ThreadNetif>()`.

This cod however has nothing to do with `ThreadNetif`, it is totally for `Netif`. Even thought we only have one `Netif` instance that is `ThreadNetif`, we might be having multiple `Netif` instances in the future. So to allow the iterator to get the corresponding `Netif` instance, we must pass `Netif` instance to it rather than being InstanceLocator.

src/core/net/netif.hpp
```diff
@@ -156,6 +157,47 @@ class NetifMulticastAddress : public otNetifMulticastAddress,
...
+public:
+    /**
+     * This method returns the next `ExternalNetifMulticastAddress` instance in the a linked list.
+     *
+     * @returns A pointer to the next `ExternalNetifMulticastAddress` instance.
+     *
+     */
+    ExternalNetifMulticastAddress *GetNext(void)
+    {
+        return static_cast<ExternalNetifMulticastAddress *>(const_cast<otNetifMulticastAddress *>(mNext));
```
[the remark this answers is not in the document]
abtink: Do we also need the `GetNext()` here for the use by `Pool` class?
If so, can we see if we can define it as `private` here (since `LinkedList<Ext...>` is a friend already)?

src/core/net/netif.hpp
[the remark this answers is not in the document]
simonlingoogle: Yes. But I defined `GetNext` in public so `friend class` is actually not necessary. 
Agree that it's better to  `GetNext()` to private.

src/core/net/netif.hpp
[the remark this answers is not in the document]
simonlingoogle: Yes, we need this for Pool to compile, otherwise we get this compile error:
```c++
In file included from openthread/src/core/net/netif.cpp:34:
In file included from openthread/src/core/net/netif.hpp:41:
openthread/src/core/common/linked_list.hpp:191:28: error: assigning to 'ot::Ip6::ExternalNetifMulticastAddress *' from incompatible type 'ot::Ip6::NetifMulticastAddress *'
            mHead = mHead->GetNext();
                    ~~~~~~~^~~~~~~~~
openthread/src/core/common/pool.hpp:105:45: note: in instantiation of member function 'ot::LinkedList<ot::Ip6::ExternalNetifMulticastAddress>::Pop' requested here
    Type *Allocate(void) { return mFreeList.Pop(); }
                                            ^
openthread/src/core/net/netif.cpp:335:38: note: in instantiation of member function 'ot::Pool<ot::Ip6::ExternalNetifMulticastAddress, 2>::Allocate' requested here
    entry = mExtMulticastAddressPool.Allocate();
                                     ^
```

Another fix might be changing the code of `Pop()` to use `static_cast`:
```c++
    Type *Pop(void)
    {
        Type *entry = mHead;

        if (mHead != nullptr)
        {
            mHead = static_cast<Type *>(mHead->GetNext());
        }

        return entry;
    }
```

Does it sounds logical to you? @abtink

src/core/net/netif.hpp
simonlingoogle: encapsulated the advance logic in `AdvanceFrom` method.

src/core/net/netif.hpp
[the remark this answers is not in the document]
simonlingoogle: moved to private.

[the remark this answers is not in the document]
abtink: I think, if defining `GetNext()` on `ExternalMulticastAddress` as `private` works out, it is a good solution. We let the `Pool`'s linked list to use this (being a `friend`) but not any one else.

I am not sure about changing `Pop()`. 

----

This actually exposes a limitation of the `Pool` class that its template `Type` needs to be linked-listable. I think there may be way to remove such a restriction (and make `Pool` even more flexible), but that's probably for a future PR/solution. :)

src/core/net/netif.hpp
abtink: Keeping `Netif &mNetif` is fine.

The small advantage of `InstanceLocator` is that it would be empty class and we won't need a new `Netif &` to store and therefore reduces the (RAM) size of iterator, but I guess since `Iterator`s are generally temporary/stack variables, it may not matter. So we can keep the ref to `Netif`.

src/core/net/netif.hpp
```diff
@@ -164,8 +200,125 @@ class Netif : public InstanceLocator, public LinkedListEntry<Netif>, private Non
...
+        }
+
+        /**
+         * This method overloads `++` operator (pre-increment) to advance the iterator.
+         *
+         * The iterator is moved to point to the next entry.  If there are no more entries matching the iterator becomes
+         * empty.
+         *
+         */
+        void operator++(void) { AdvanceFrom(mCurrent->GetNext()); }
```
abtink, line 235: AdvanceFrom(const_cast<NetifMulticastAddress *>(mNetif.GetMulticastAddresses()));

src/core/net/netif.hpp
```diff
@@ -164,8 +200,125 @@ class Netif : public InstanceLocator, public LinkedListEntry<Netif>, private Non
...
+            kEndIterator,
+        };
+
+        ExternalMulticastAddressIterator(Netif &aNetif, IteratorType)
+            : mNetif(aNetif)
+            , mCurrent(nullptr)
+        {
+        }
+
+        void AdvanceFrom(NetifMulticastAddress *aAddr)
```
abtink, line 307: Can we consider defining `aAddr` as `const NetifMulticastAddress *`?

It would then help with `AdvanceFrom(const_cast<NetifMulticastAddress *>(mNetif.GetMulticastAddresses()));` (no need for the `const_cast`) and I think it should work ok with `AdvanceFrom(mCurrent->GetNext());` as well?

src/core/net/netif.hpp
simonlingoogle: I intentionally didn't check it. I think we should never call Advance when the iterator is empty because it is a logical error. We will never have any issue if we just use it in c++11 loops.

src/core/net/netif.hpp
simonlingoogle: revised.

src/core/net/netif.hpp
simonlingoogle: The document of [std::advance](https://en.cppreference.com/w/cpp/iterator/advance) defines increasing the `past-the-end` iterator to be undefined. 
```
The behavior is undefined if the specified sequence of increments or decrements would require that 
a non-incrementable iterator (such as the past-the-end iterator) is incremented, ...
```

So I think it's not required to check for `mCurrent == nullptr` so that we have a more concise implementation.

abtink: SGTM.

script/check-size
jwhui: Adding columns sounds like a good idea.

We now know that this PR adds 2048 bytes code and 32 bytes RAM. I think we should remove this flag in this PR and have a separate PR that updates size-report to include 1.2 changes.

Thoughts?

simonlingoogle: Example:
# Size Report of **OpenThread**
> Merging [#5268](https://github.com/openthread/openthread/commit/7497659a1b11ef1aea9b6774c9f6e9239a4b7440) into [master](https://github.com/openthread/openthread/commit/57ee45be21c2b6cbba0dc53cea235fca33772cbc)(57ee45be21c2b6cbba0dc53cea235fca33772cbc).

|  name  |  branch  |  text  | data  | bss  | total | text  | data  | bss  | total |
| :----: | :------: | -----: | ----: | ---: | ----: |  ----: | ----: | ----: | ----: |
| ot-cli-ftd | master | 396720 | 736 | 72828 | 470284 | 396720 | 736 | 72828 | 470284 |
|  | #5268 | 396720 | 736 | 72828 | 470284 | 396720 | 736 | 72828 | 470284 |
|  | +/- | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 |
| ot-cli-mtd | master | 319360 | 736 | 64180 | 384276 | 319360 | 736 | 64180 | 384276 |
|  | #5268 | 319360 | 736 | 64180 | 384276 | 319360 | 736 | 64180 | 384276 |
|  | +/- | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 |

librasungirl: sgtm to add columns for 1.2 size. I may suggest to keep `total` as it is more visible to learn the overall consumption.
The other option is two reports separately for 1.1 and 1.2 if there is width concern/issue for the 'add columns' option

src/core/net/netif.cpp
```diff
@@ -336,6 +336,9 @@ otError Netif::SubscribeExternalMulticast(const Address &aAddress)
     VerifyOrExit(entry != nullptr, error = OT_ERROR_NO_BUFS);
 
     entry->mAddress = aAddress;
+#if (OPENTHREAD_FTD || OPENTHREAD_MTD) && OPENTHREAD_CONFIG_MLR_ENABLE
```
librasungirl, line 339: is `(OPENTHREAD_FTD || OPENTHREAD_MTD)` necessary?

src/core/net/netif.hpp
```diff
@@ -156,6 +157,41 @@ class NetifMulticastAddress : public otNetifMulticastAddress,
     bool Matches(const Address &aAddress) const { return GetAddress() == aAddress; }
 };
 
+class ExternalNetifMulticastAddress : public NetifMulticastAddress
+{
+    friend class Netif;
+    friend class LinkedList<ExternalNetifMulticastAddress>;
+
+public:
+#if (OPENTHREAD_FTD || OPENTHREAD_MTD) && OPENTHREAD_CONFIG_MLR_ENABLE
```
librasungirl, line 166: is `(OPENTHREAD_FTD || OPENTHREAD_MTD)` necessary?

src/core/net/netif.hpp
```diff
@@ -156,6 +157,41 @@ class NetifMulticastAddress : public otNetifMulticastAddress,
...
+    void SetMlrState(MlrState aState) { mMlrState = aState; }
+#endif
+
+private:
+    ExternalNetifMulticastAddress *GetNext(void)
+    {
+        return static_cast<ExternalNetifMulticastAddress *>(const_cast<otNetifMulticastAddress *>(mNext));
+    }
+
+#if (OPENTHREAD_FTD || OPENTHREAD_MTD) && OPENTHREAD_CONFIG_MLR_ENABLE
```
librasungirl, line 190: is `(OPENTHREAD_FTD || OPENTHREAD_MTD)` necessary?

src/core/net/netif.cpp
simonlingoogle: removed

src/core/net/netif.hpp
simonlingoogle: removed

src/core/net/netif.hpp
simonlingoogle: removed

script/check-size
simonlingoogle: @jwhui 
check-size change has been removed from this PR. Let's discuss this further in https://github.com/openthread/openthread/issues/5274.

PR #5256, opened 2020-07-17
[mle] work around MLE Link Accept messages that are too large

### LINKED ISSUE #5152: Router losing its role after reset in bigger network
My setup: 20x NRF52840 hardware nodes with LONG_ROUTES enabled (still Nordic's SDK 2.0: 41d8e4a with multiple different patches)
Note: This cannot be reproduced with 10 nodes.

1. enable LONG_ROUTES
2. take 20 nodes
3. take node N, make sure it is a router
4. reset node N
5. node N will be a child after startup (instead of keeping previous router RLOC)

I investigated the problem a bit.
1. Node N wakes up and sends Link Request to all routers as expected.
2. Neighbor routers answer with Link Accept providing Address16 TLV as expected
3. Because of big Route64 TLV (Length = 33) whole Link Accept does not fit into one message and is splitted into 2 messages
4. Node N do not receive those messages properly ( drops them ?, HandleLinkAccept(), HandleUdpReceive() are not called )

Screenshot from 2020-06-25 16-48-08

PS. For 10 nodes Route64 lenght is 22 bytes and Link Accept fits into 1 message.
For 140 nodes where I make more tests Route64 length is 57

Update: I thought that fragmented MLE is just something which is not supported but I can see that I have Child  ID Responses divided into 2 fragments and it looks like everything works fine there.

jwhui:
This commit addresses an issue where the MLE Link Accept message may
be too large to fit within a single IEEE 802.15.4 frame. Thread
requires any MLE message transported using 6LoWPAN fragmentation to
also enable IEEE 802.15.4 MAC frame security. However, the purpose of
MLE Link Accept messages is to establish a secure link and cannot
itself be transported with IEEE 802.15.4 MAC frame security enabled.

An MLE Link Accept message may carry a Route64 TLV that has a length
dependent on the number of active routers in the network. When there
are many active routers, the Route64 TLV is large and the MLE Link
Accept message can no longer fit within a single IEEE 802.15.4
frame. The side effect is that a router coming out of reset will not
be able to diretly restore its router role and must first attach as an
end device.

This commit limits the size of a Route64 TLV when included in MLE Link
Accept messages. When the number of active routers is above a
threshold, some routers will be elided from the Route64 TLV and the
Router ID Sequence value is decremented. This allows the device coming
out of reset to restore its router role and retrieve the current
Route64 TLV in a subsequent message exchange.

Resolves #5152

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -725,7 +725,7 @@ otError MleRouter::SendLinkAccept(const Ip6::MessageInfo &aMessageInfo,
         switch (aRequestedTlvs.mTlvs[i])
         {
         case Tlv::kRoute:
-            SuccessOrExit(error = AppendRoute(*message));
+            SuccessOrExit(error = AppendRoute(*message, aNeighbor));
             break;
 
         case Tlv::kAddress16:
@@ -4139,15 +4139,62 @@ exit:
     return error;
 }
 
-void MleRouter::FillRouteTlv(RouteTlv &aTlv)
+void MleRouter::FillRouteTlv(RouteTlv &aTlv, Neighbor *aNeighbor)
 {
-    uint8_t routerCount = 0;
+    uint8_t     routerIdSequence = mRouterTable.GetRouterIdSequence();
+    RouterIdSet routerIdSet      = mRouterTable.GetRouterIdSet();
+    uint8_t     routerCount;
 
-    aTlv.SetRouterIdSequence(mRouterTable.GetRouterIdSequence());
-    aTlv.SetRouterIdMask(mRouterTable.GetRouterIdSet());
+    if (aNeighbor && IsActiveRouter(aNeighbor->GetRloc16()))
+    {
+        // Sending a Link Accept message that may require truncation
+        // of Route64 TLV
+
+        routerCount = mRouterTable.GetActiveRouterCount();
+
+        if (routerCount > RouteTlv::kLinkAcceptMaxRouters)
+        {
+            for (uint8_t routerId = 0; routerId <= kMaxRouterId; routerId++)
+            {
+                if (routerCount <= RouteTlv::kLinkAcceptMaxRouters)
+                {
+                    break;
+                }
+
+                if ((routerId == RouterIdFromRloc16(GetRloc16())) || (routerId == aNeighbor->GetRouterId()) ||
+                    (routerId == GetLeaderId()))
+                {
+                    // Route64 TLV must contain this device and the
+                    // neighboring router to ensure that at least this
+                    // link can be established.
+                    continue;
+                }
+
+                if (routerIdSet.Contains(routerId))
+                {
+                    routerIdSet.Remove(routerId);
+                    routerCount--;
+                }
+            }
+
+            // Ensure that the neighbor will process the current
+            // Route64 TLV in a subsequent message exchange
+            routerIdSequence -= RouteTlv::kLinkAcceptSequenceRollback;
+        }
+    }
+
+    aTlv.SetRouterIdSequence(routerIdSequence);
+    aTlv.SetRouterIdMask(routerIdSet);
+
+    routerCount = 0;
 
     for (Router &router : Get<RouterTable>().Iterate())
     {
+        if (!routerIdSet.Contains(router.GetRouterId()))
+        {
+            continue;
+        }
+
         if (router.GetRloc16() == GetRloc16())
         {
             aTlv.SetLinkQualityIn(routerCount, 0);
@@ -4193,12 +4240,12 @@ void MleRouter::FillRouteTlv(RouteTlv &aTlv)
     aTlv.SetRouteDataLength(routerCount);
 }
 
-otError MleRouter::AppendRoute(Message &aMessage)
+otError MleRouter::AppendRoute(Message &aMessage, Neighbor *aNeighbor)
 {
     RouteTlv tlv;
 
     tlv.Init();
-    FillRouteTlv(tlv);
+    FillRouteTlv(tlv, aNeighbor);
 
     return tlv.AppendTo(aMessage);
 }
--- a/tests/scripts/thread-cert/test_router_reattach.py
+++ b/tests/scripts/thread-cert/test_router_reattach.py
@@ -1,0 +1,281 @@
+#!/usr/bin/env python3
+#
+#  Copyright (c) 2020, The OpenThread Authors.
+#  All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions are met:
+#  1. Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#  2. Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in the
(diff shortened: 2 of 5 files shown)


jwhui:
@mwilniewiec , this is one proposal to address #5256. Please review.

src/core/thread/mle_router.cpp
```diff
@@ -4472,17 +4472,63 @@ otError MleRouter::AppendChildAddresses(Message &aMessage, Child &aChild)
...
+
+    aTlv.SetRouterIdSequence(routerIdSequence);
+    aTlv.SetRouterIdMask(routerIdSet);
+
+    routerCount = 0;
 
-    for (RouterTable::Iterator iter(GetInstance()); !iter.IsDone(); iter++, routerCount++)
+    for (RouterTable::Iterator iter(GetInstance()); !iter.IsDone(); iter++)
     {
         Router &router = *iter.GetRouter();
```
abtink, line 4525:
```suggestion
    for (Router &router : Get<RouterTable>().Iterate()))
    {
```

simonlingoogle:
Could there be any side-effects of this solution? Is it possible that a Router would receive this "faked" Route TLV and rewrite the correct version it already has?

jwhui:
> Could there be any side-effects of this solution? Is it possible that a Router would receive this "faked" Route TLV and rewrite the correct version it already has?

@simonlingoogle , the Router ID Sequence field should prevent this. Devices that have already process and accepted a Route64 TLV should only accept new Route64 TLVs that have a higher Router ID Sequence value.

simonlingoogle:
> > Could there be any side-effects of this solution? Is it possible that a Router would receive this "faked" Route TLV and rewrite the correct version it already has?
> 
> @simonlingoogle , the Router ID Sequence field should prevent this. Devices that have already process and accepted a Route64 TLV should only accept new Route64 TLVs that have a higher Router ID Sequence value.

A Router's Router ID Sequence might fall behind if it failed to receive Advertisements for a period of time. In such case it might receive and process this `faked` Route TLV, causing it to remove Routers that actually exist. 

Would it be safer if we decrease `routerIdSequence` by a larger value instead of `routerIdSequence--`? 
For example, `routerIdSequence -= 64`.

jwhui:
> Would it be safer if we decrease `routerIdSequence` by a larger value instead of `routerIdSequence--`?
> For example, `routerIdSequence -= 64`.

@simonlingoogle , that's a good point. I applied the suggested value for now. I'm hoping through some testing we can see if this overall approach is workable.

src/core/thread/mle_router.cpp
```diff
@@ -4468,15 +4468,61 @@ otError MleRouter::AppendChildAddresses(Message &aMessage, Child &aChild)
...
+                    // Route64 TLV must contain this device and the
+                    // neighboring router to ensure that at least this
+                    // link can be established.
+                    continue;
+                }
+
+                if (routerIdSet.Contains(routerId))
+                {
+                    routerIdSet.Remove(routerId);
+                    routerCount -= RouteTlv::kLinkAcceptSequenceRollback;
```
simonlingoogle, line 4504:
```suggestion
                    routerCount--;
```

src/core/thread/mle_router.cpp
```diff
@@ -4468,15 +4468,61 @@ otError MleRouter::AppendChildAddresses(Message &aMessage, Child &aChild)
...
+                }
+            }
+
+            // Ensure that the neighbor will process the current
+            // Route64 TLV in a subsequent message exchange
+            routerIdSequence--;
```
simonlingoogle, line 4510:
```suggestion
            routerIdSequence -= RouteTlv::kLinkAcceptSequenceRollback;
```

src/core/thread/mle_router.cpp
```diff
@@ -4448,15 +4448,61 @@ otError MleRouter::AppendChildAddresses(Message &aMessage, Child &aChild)
...
 
-void MleRouter::FillRouteTlv(RouteTlv &aTlv)
+void MleRouter::FillRouteTlv(RouteTlv &aTlv, Neighbor *aNeighbor)
 {
-    uint8_t routerCount = 0;
+    uint8_t     routerIdSequence = mRouterTable.GetRouterIdSequence();
+    RouterIdSet routerIdSet      = mRouterTable.GetRouterIdSet();
+    uint8_t     routerCount;
+
+    if (aNeighbor && IsActiveRouter(aNeighbor->GetRloc16()))
```
simonlingoogle, line 4457: Should we have a test to verify that this work around works?

jwhui: I have a test, but it fails the `cli-time-sync` check.
https://github.com/openthread/openthread/runs/905142732?check_suite_focus=true

I haven't had a chance to figure out why. Let me know if you have any ideas :)

src/core/thread/mle_router.cpp
```diff
@@ -4448,15 +4448,61 @@ otError MleRouter::AppendChildAddresses(Message &aMessage, Child &aChild)
...
+                }
 
-    aTlv.SetRouterIdSequence(mRouterTable.GetRouterIdSequence());
-    aTlv.SetRouterIdMask(mRouterTable.GetRouterIdSet());
+                if ((routerId == RouterIdFromRloc16(GetRloc16())) || (routerId == aNeighbor->GetRouterId()))
+                {
+                    // Route64 TLV must contain this device and the
+                    // neighboring router to ensure that at least this
+                    // link can be established.
+                    continue;
```
simonlingoogle, line 4478: We should also keep the Leader?
According to your discussion in https://github.com/openthread/openthread/pull/5301#discussion_r463497948, I think we should make sure Leader ID is included in the shrinked RouteTLV.
Thoughts?

src/core/thread/mle_router.cpp
librasungirl: When Time Sync feature is enabled, TimeHeader IE would be appended, thus bringing extra 17 Bytes (sizeof(TimeHeaderIE) 15B + IE Termination 2 B), and causing fragmentation.
https://github.com/openthread/openthread/blob/6488b25484aeed042a72962d1b3e3b5f2074f688/src/core/thread/mesh_forwarder.cpp#L570-L572

One possible option is when Time Sync is enabled, define a smaller `kLinkAcceptMaxRouters` (verified that value 3 could work, fyi)

tests/scripts/thread-cert/test_router_reattach.py
```diff
@@ -0,0 +1,281 @@
+#!/usr/bin/env python3
+#
+#  Copyright (c) 2016, The OpenThread Authors.
```
simonlingoogle, line 3:
```suggestion
#  Copyright (c) 2020, The OpenThread Authors.
```

tests/scripts/thread-cert/test_router_reattach.py
```diff
@@ -0,0 +1,281 @@
...
+            'router_upgrade_threshold': 32
+        },
+    }
+
+    def test(self):
+        self.nodes[1].start()
+        self.simulator.go(5)
+        self.assertEqual(self.nodes[1].get_state(), 'leader')
+
+        for i in range(2, 21):
```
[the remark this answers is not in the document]
librasungirl: Agree that totally 20 active routers should be fine for this test, if so, may not necessary to declare so much nodes in Topology above?

PR #5261, opened 2020-07-18
[mle] enhance IsExpectedToBecomeRouter()

librasungirl:


src/core/thread/mle_router.cpp
```diff
@@ -4118,8 +4118,8 @@ void MleRouter::HandleAddressSolicitResponse(Coap::Message *         aMessage,
 
 bool MleRouter::IsExpectedToBecomeRouter(void) const
 {
-    return IsRouterEligible() && !IsRouterOrLeader() &&
-           (Get<RouterTable>().GetActiveRouterCount() < GetRouterUpgradeThreshold()) && !mAddressSolicitRejected;
+    return IsRouterEligible() && !IsRouterOrLeader() && !mAddressSolicitRejected &&
+           (GetRouterSelectionJitterTimeout() != 0 || mAddressSolicitPending);
```
simonlingoogle, line 4122: Could you explain more on why we need to check these conditions? @librasungirl 

More thoughts on this:
- We used `!IsRouterOrLeader()`, but do you think we should change it to `IsChild()` ?
- Should we also check `mRouterTable.GetActiveRouterCount() < mRouterUpgradeThreshold` ?

simonlingoogle, line 4122: I have a concern about the `kRouterSelectionJitter` setting. It's 120s by default and can be increased up to 255s. In the worst case, the Child node is going to delay for 255s before sending `/a/as`, and in this period the device is not going to register DUA, thus has no DUA reachability. 
But the devices are expected to get DUA quickly. Considering the default jitter is 120s and on average nodes will need to wait for 60s without DUA, do you think we should solve this problem?

librasungirl: Regarding the changes, firstly `GetRouterSelectionJitterTimeout() != 0` should be equivalent with ` Get<RouterTable>().GetActiveRouterCount() < GetRouterUpgradeThreshold()`, right?

The real addition in this PR is `mAddressSolicitPending` which represent the device is in the process of becoming router (waiting for response) in case 1) there might be CoAP retransmissions and it might take several seconds for the response depends on the communication 2) the Router Eligible device is trying to become router due to `kHaveChildIdRequest` or `kParentPartitionChange` regardless the active router number v.s. mRouterUpgradeThreshold.

Agree `IsChild()` could be applied here.

librasungirl: Thanks for sharing your concern. 
I would assume your concern comes from the possible optimization on BBR `It MAY update its EID-to-RLOC Map Cache based on the pair {DUA, RLOC16-source} which is gleaned from the DUA.req packet` and we may want to suppress the DUA.req if the RouterEligible device will become router soon.

Two options I can see:
1) not send DUA.req before the device become router, like you said, in the worst case, the device might not be reachable via DUA within `kRouterSelectionJitter`,  we may add note that the admin should be aware of the impact of changing this value.

2) send DUA.req if `GetRouterSelectionJitterTimeout()` is larger than a window (e.g. 20 seconds), thus the device could be reached via DUA sooner.  And the device could refresh its registration when it becomes router.

Thoughts?

simonlingoogle: I think option 2 is a good idea. DUA reregistration when it becomes Router (RLOC16 changed) can help reduce failed message transactions, which can easily compensate for the extra DUA.req when becoming Router.
Similarly, do you think we should also send DUA.req when a Router downgrades to a REED (RLOC16 changed)?

simonlingoogle: Agree that in case 2), `IsExpectedToBecomeRouter()` should return true. 

> Regarding the changes, firstly `GetRouterSelectionJitterTimeout() != 0` should be equivalent with ` Get<RouterTable>().GetActiveRouterCount() < GetRouterUpgradeThreshold()`, right?

It's not? When the REED is sending ADDR_SOL.req, `GetRouterSelectionJitterTimeout() should return 0 but `Get<RouterTable>().GetActiveRouterCount() < GetRouterUpgradeThreshold()` should still be true. 

Overall I agree the new code is more better.

librasungirl: yes, you are right. My saying is not so precise, I intended to say they are equivalent before sending ADDR_SOL.req if the REED finally would become router.
Thanks.

librasungirl: Good point. I think it could be assumed that the downgrade conditions could still be met when the Router downgrades to REED successfully, that is, `IsExpectedToBecomeRouter()` would be false, the same logic, just like a new attached REED, would be applied automatically.

simonlingoogle: > Good point. I think it could be assumed that the downgrade conditions could still be met when the Router downgrades to REED successfully, that is, `IsExpectedToBecomeRouter()` would be false, the same logic, just like a new attached REED, would be applied automatically.

Wouldn't `IsExpectedToBecomeRouter()` return true when a Router downgrades to REED? 

When a Router downgrades to a REED, it won't send `DUA.req` because it's not required to to do so? It already has the DUA, only the RLOC16 is wrong on PBBR.

simonlingoogle: Maybe we can simply remove this `DUA.req suppressing within Routter Jitter window` feature?
Just don't feel it's that useful, but instead complicates the logic and could cause DUA issues.

librasungirl: Reference device should have routerselectionjitter as 1 in THCI. may need to evaluate possible impact on REED DUT when not enable AUTO_DUT.
Anyway, it could be resolved when issue occurs.

simonlingoogle: Looks fine to me.

PR #5264, opened 2020-07-20
[address-resolver] TimeSinceLastTransaction should be in seconds

librasungirl:
Thread Specification 5.4.2.4
> The Time Since Last Transaction TLV contains the number of seconds since a keep-alive message (as defined in Section 4.6.3 Timing Out Children in Chapter 4, Mesh Link Establishment) was last received from the Child.

### DIFF
--- a/src/core/thread/address_resolver.cpp
+++ b/src/core/thread/address_resolver.cpp
@@ -42,6 +42,7 @@
 #include "common/instance.hpp"
 #include "common/locator-getters.hpp"
 #include "common/logging.hpp"
+#include "common/time.hpp"
 #include "mac/mac_types.hpp"
 #include "thread/mesh_forwarder.hpp"
 #include "thread/mle_router.hpp"
@@ -787,7 +788,7 @@ void AddressResolver::HandleAddressQuery(Coap::Message &aMessage, const Ip6::Mes
 
         if (child.HasIp6Address(target))
         {
-            lastTransactionTime = TimerMilli::GetNow() - child.GetLastHeard();
+            lastTransactionTime = Time::MsecToSec(TimerMilli::GetNow() - child.GetLastHeard());
             SendAddressQueryResponse(target, child.GetMeshLocalIid(), &lastTransactionTime, aMessageInfo.GetPeerAddr());
             ExitNow();
         }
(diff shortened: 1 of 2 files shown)


src/core/thread/address_resolver.cpp
```diff
@@ -787,7 +788,7 @@ void AddressResolver::HandleAddressQuery(Coap::Message &aMessage, const Ip6::Mes
 
         if (child.HasIp6Address(target))
         {
-            lastTransactionTime = TimerMilli::GetNow() - child.GetLastHeard();
+            lastTransactionTime = Time::MsecToSec(TimerMilli::GetNow() - child.GetLastHeard());
```
abtink, line 791: @librasungirl , this rounds down the value, thus breaking the test `tests/toranj/test-039-address-cache-table-snoop.py`. In particular here (where a non-zero last trans is expected).
```
# We expect to see the cache entries for c3 addresses pointing to r3.

cache_table = wpan.parse_address_cache_table_result(
    r1.get(wpan.WPAN_THREAD_ADDRESS_CACHE_TABLE))

for index in range(NUM_ADDRESSES):
    verify(cache_table[index].address == PREFIX + "c3:" +
           str(NUM_ADDRESSES - index - 1))
    verify(cache_table[index].rloc16 == r3_rloc)
    verify(cache_table[index].state == wpan.ADDRESS_CACHE_ENTRY_STATE_CACHED)
    verify(cache_table[index].last_trans > 0)
```

Here are some suggsted solutions:
- We can round up at zero - i.e. not allowing `lastTransactionTime` to be set to zero (if zero set it to `1`). Specially since value 0 is used when there is no such TLV appended.
- Change the test to wait for 1 second before checking the cache table.
- Remove this check from test.

I prefer first option. Thoughts?

librasungirl: Thanks @abtink for looking into this and the suggestions here.
I would prefer to allow it to be 0 as the definition for `lastTransactionTime` is `the number of seconds since a keep-alive message was last received from the Child.`.

The poll period is configured as 200ms in this test, i.e. the keep-alive message period is 200ms, so the last_trans would always be 0 seconds. Given the poll period, I think we can change the check to be `verify(cache_table[index].last_trans == 0)`, or remove this check in case POLL_PERIOD may be changed someday.

Thoughts?

librasungirl: Updated the test script to verify it is `last_trans == 0` tentatively as there might be little chance to change POLL_PERIOD in this test.

abtink: SGTM

PR #5267, opened 2020-07-21
[core] add BitVector template class

simonlingoogle:
This adds the `BitVector` template class and changes `ChildMask` to inherit from `BitVector`.
This can help reduce redundant code for bit mask manipulations.

### DIFF
--- a/src/core/common/bit_vector.hpp
+++ b/src/core/common/bit_vector.hpp
@@ -1,0 +1,139 @@
+/*
+ *  Copyright (c) 2020, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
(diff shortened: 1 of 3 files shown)


src/core/thread/child_mask.hpp
```diff
@@ -36,11 +36,7 @@
 
 #include "openthread-core-config.h"
 
-#include <openthread/error.h>
-
-#include "common/code_utils.hpp"
-#include "common/debug.hpp"
-#include "common/encoding.hpp"
+#include "utils/bit_vector.hpp"
```
abtink, line 39: I suggest we move `bit_vector.hpp` under `common` folder. The `common` is for all common helper modules used in OT core.

The  `utils` folder is mainly for extra features/functionalities that may not be directly part of Thread spec (and are not directly used by other OT core modules) but are useful additional/extra features we want to include in OT, for example, jam-detector, channel-monitor, heap managment, flash managment, etc.

src/core/utils/bit_vector.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * This class represents a bit-vector.
+ *
+ */
+template <uint16_t N> class BitVector
+{
+public:
+    /**
+     * This method returns if a given index is masked.
```
abtink, line 61:
```suggestion
     * This method indicates whether a given index is included in the mask or not.
```

src/core/thread/child_mask.hpp
```diff
@@ -58,78 +54,8 @@ namespace ot {
  * This class represents a bit-vector of child mask.
  *
  */
-class ChildMask
+class ChildMask : public BitVector<OPENTHREAD_CONFIG_MLE_MAX_CHILDREN>
```
abtink, line 57: How about using `typedef` here?
```
typedef BitVector<OPENTHREAD_CONFIG_MLE_MAX_CHILDREN> ChildMask;
```

We can make it a sub-class later if/when we have more APIs that we want to add to `ChildMask`?

src/core/thread/child_mask.hpp
simonlingoogle: I moved it from `common/` to `utils/`. Now I feel sorry for that.

simonlingoogle:
#5236 depends on this PR to be merged to rebase. @librasungirl @jwhui 
Please review when you have time. Thanks

src/core/common/bit_vector.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * @file
+ *   This file includes definitions for a bit-vector.
+ */
+
+#ifndef BIT_VECTOR_HPP_
+#define BIT_VECTOR_HPP_
+
+#include "../openthread-core-config.h"
```
librasungirl, line 37:
```suggestion
#include "openthread-core-config.h"
```

simonlingoogle:
> LGTM
> 
> As in #5276, ChildMask is required to be comparable and clearable, do you plan to enhance this one or a separate one later?

`BitVector` has a `ClearAll()` method to clear all indexes. 
```
void ClearAll(void) { memset(mMask, 0, sizeof(mMask)); }
```

I intentionally didn't inherit Clearable<BitVector> to avoid using the same method `Clear` for two different methods to reduce confusion. 
Thoughts? @librasungirl 

BTW, Comparable is a good idea, I will add.

src/core/common/bit_vector.hpp
```diff
@@ -0,0 +1,139 @@
...
+ * @brief
+ *   This module includes definitions for bit-vector.
+ *
+ * @{
+ *
+ */
+
+/**
+ * This class represents a bit-vector.
+ *
```
jwhui, line 55: nit: add `@tparam` doc

PR #5268, opened 2020-07-21
[otns] emit device mode

simonlingoogle:
This PR allows [OTNS](https://github.com/openthread/ot-ns) to correctly visualize node mode when it changes.

### DIFF
--- a/src/core/utils/otns.cpp
+++ b/src/core/utils/otns.cpp
@@ -153,6 +153,12 @@ void Otns::EmitTransmit(const Mac::TxFrame &aFrame)
     }
 }
 
+void Otns::EmitDeviceMode(Mle::DeviceMode aMode)
+{
+    EmitStatus("mode=%s%s%s%s", aMode.IsRxOnWhenIdle() ? "r" : "", aMode.IsSecureDataRequest() ? "s" : "",
+               aMode.IsFullThreadDevice() ? "d" : "", aMode.IsFullNetworkData() ? "n" : "");
+}
+
 } // namespace Utils
 } // namespace ot
 
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -778,6 +778,10 @@ otError Mle::SetDeviceMode(DeviceMode aDeviceMode)
     VerifyOrExit(mDeviceMode != aDeviceMode, OT_NOOP);
     mDeviceMode = aDeviceMode;
 
+#if OPENTHREAD_CONFIG_OTNS_ENABLE
+    Get<Utils::Otns>().EmitDeviceMode(mDeviceMode);
+#endif
+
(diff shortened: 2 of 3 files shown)


src/core/utils/otns.cpp
```diff
@@ -153,6 +153,12 @@ void Otns::EmitTransmit(const Mac::TxFrame &aFrame)
     }
 }
 
+void Otns::EmitDeviceMode(Mle::DeviceMode aMode)
+{
+    EmitStatus("mode=%s%s%s%s", aMode.IsRxOnWhenIdle() ? "r" : "", aMode.IsSecureDataRequest() ? "s" : "",
+               aMode.IsFullThreadDevice() ? "d" : "", aMode.IsFullNetworkData() ? "n" : "");
```
ctan-g, line 159: Isn't mode represented by an 8-bit int? What's the advantage of using this string instead of sending the int (and parsing it in OTNS)?

simonlingoogle: This code uses the same format used by the CLI [mode](https://github.com/openthread/openthread/blob/master/src/cli/README.md#mode) command which is familiar to OpenThread developers. 
Since we don't care that much about performance here, I prefer using a more human-readable status string.   

Now that real devices could also emit OTNS status via logging, this also makes the device log more readable.

ctan-g: Sounds good!

abtink: This is fine as is. 
We can consider using `DeviceMode::ToString()` here. The string would be longer, but then more readable, something like: `"rx-on:no sec-poll:yes ftd:no full-net:yes"`
We can also emit both raw value and string, like `"mode=0x%02x (%s)", aMode.Get(), aMode.ToString().AsCString()`.

simonlingoogle: > We can consider using `DeviceMode::ToString()` here. The string would be longer, but then more readable, something like: `"rx-on:no sec-poll:yes ftd:no full-net:yes"`
> We can also emit both raw value and string, like `"mode=0x%02x (%s)", aMode.Get(), aMode.ToString().AsCString()`.

Being human-readable as it is, the status string should also be easy to parse. OTNS can just iterate the characters in the mode string (e.g. `rsdn`, `sn`) and set each flag accordingly. 

I would even think that `DeviceMode::ToString()` should return strings like `rsdn`, `sn` so that we have a uniform string representation for device modes across the OpenThread code (i.e. `DeviceMode::ToString()` and CLI `mode`).

Thougths?

abtink: Sounds fine to me to use character flags for OTNS here.

With regards to `DeviceMode::ToString()`, it is mainly intended for logging. For this I would personally prefer a more readable string like `"rx-on:no sec-poll:yes ftd:no full-net:yes"` over `"rsdn"` char flags in the logs. I know in similar cases I have forgotten the association of chars with actual flags (their meaning) and then need to look it up the code, so I generally favor more readable logs.

simonlingoogle: More readable sound good to me.

PR #5269, opened 2020-07-21
[child] enhance Child::AddressIterator

abtink:
This commit enhances the iterator `Child::AddressIterator` which is
used to iterate over IPv6 address list of a `Child`. The new model
enables range-based `for` loops to be used. It also enables filtering
addresses based on type (e.g., iterate over multicast addresses only).
The unit test `test_child.cpp` is also updated to verify the newly
added functionalities of `AddressIterator`.

### DIFF
--- a/src/core/api/thread_ftd_api.cpp
+++ b/src/core/api/thread_ftd_api.cpp
@@ -273,22 +273,28 @@ otError otThreadGetChildInfoByIndex(otInstance *aInstance, uint16_t aChildIndex,
 
 otError otThreadGetChildNextIp6Address(otInstance *               aInstance,
                                        uint16_t                   aChildIndex,
-                                       otChildIp6AddressIterator *aIterator,
+                                       otChildIp6AddressIterator *aIterIndex,
                                        otIp6Address *             aAddress)
 {
-    otError                   error    = OT_ERROR_NONE;
-    Instance &                instance = *static_cast<Instance *>(aInstance);
-    Child::Ip6AddressIterator iterator;
-    Ip6::Address *            address;
+    otError      error    = OT_ERROR_NONE;
+    Instance &   instance = *static_cast<Instance *>(aInstance);
+    const Child *child;
 
-    OT_ASSERT(aIterator != nullptr && aAddress != nullptr);
+    OT_ASSERT(aIterIndex != nullptr && aAddress != nullptr);
 
-    address = static_cast<Ip6::Address *>(aAddress);
-    iterator.Set(*aIterator);
+    child = instance.Get<ChildTable>().GetChildAtIndex(aChildIndex);
+    VerifyOrExit(child != nullptr, error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(child->IsStateValidOrRestoring(), error = OT_ERROR_INVALID_ARGS);
 
-    SuccessOrExit(error = instance.Get<Mle::MleRouter>().GetChildNextIp6Address(aChildIndex, iterator, *address));
+    {
+        Child::AddressIterator iter(*child, *aIterIndex);
 
-    *aIterator = iterator.Get();
+        VerifyOrExit(!iter.IsDone(), error = OT_ERROR_NOT_FOUND);
+        *aAddress = *iter.GetAddress();
+
+        iter++;
+        *aIterIndex = iter.GetAsIndex();
+    }
 
 exit:
     return error;
--- a/src/core/net/ip6_address.cpp
+++ b/src/core/net/ip6_address.cpp
@@ -369,6 +369,31 @@ uint8_t Address::PrefixMatch(const otIp6Address &aOther) const
     return PrefixMatch(mFields.m8, aOther.mFields.m8, sizeof(Address));
 }
 
+bool Address::MatchesFilter(TypeFilter aFilter) const
+{
+    bool matches = true;
+
+    switch (aFilter)
+    {
+    case kTypeAny:
+        break;
+
+    case kTypeUnicast:
+        matches = !IsUnspecified() && !IsMulticast();
+        break;
+
+    case kTypeMulticast:
+        matches = IsMulticast();
+        break;
+
+    case kTypeMulticastLargerThanRealmLocal:
+        matches = IsMulticastLargerThanRealmLocal();
+        break;
+    }
+
+    return matches;
+}
+
 otError Address::FromString(const char *aBuf)
 {
     otError     error  = OT_ERROR_NONE;
--- a/src/core/net/ip6_address.hpp
+++ b/src/core/net/ip6_address.hpp
@@ -325,6 +325,18 @@ public:
     };
 
     /**
+     * This enumeration defines IPv6 address type filter.
+     *
+     */
+    enum TypeFilter : uint8_t
+    {
+        kTypeAny,                           ///< Accept any IPv6 address (unicast or multicast).
+        kTypeUnicast,                       ///< Accept unicast IPv6 addresses only.
+        kTypeMulticast,                     ///< Accept multicast IPv6 addresses only.
+        kTypeMulticastLargerThanRealmLocal, ///< Accept multicast IPv6 addresses with scope larger than Realm Local.
+    };
+
+    /**
      * This type defines the fixed-length `String` object returned from `ToString()`.
      *
      */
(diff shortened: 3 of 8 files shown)


src/core/net/ip6_address.hpp
```diff
@@ -324,6 +324,17 @@ class Address : public otIp6Address, public Equatable<Address>, public Clearable
         kGlobalScope         = 14, ///< Global scope
     };
 
+    /**
+     * This enumeration defines IPv6 address type filter.
+     *
+     */
+    enum TypeFilter : uint8_t
+    {
+        kTypeAny,           ///< Accept any IPv6 address (unicast or multicast).
```
simonlingoogle, line 333:
```suggestion
        kTypeFilterAny,           ///< Accept any IPv6 address (unicast or multicast).
```
Put type name in the value names?

src/core/net/ip6_address.hpp
```diff
@@ -324,6 +324,17 @@ class Address : public otIp6Address, public Equatable<Address>, public Clearable
...
+        kTypeUnicastOnly,   ///< Accept unicast IPv6 addresses only.
+        kTypeMulticastOnly, ///< Accept multicast IPv6 addresses only.
```
simonlingoogle, line 335:
```suggestion
        kTypeMulticastOnly, ///< Accept multicast IPv6 addresses only.
        kTypeMulticastLargerThanRealmLocal, ///< Accept multicast IPv6 addresses with scope larger than Realm Local.
```
kTypeMulticastLargerThanRealmLocal can be very useful for MLR features.

src/core/thread/topology.cpp
```diff
@@ -111,6 +111,84 @@ void Neighbor::GenerateChallenge(void)
...
+    Update(action);
+}
+
+otChildIp6AddressIterator Child::AddressIterator::ConvertToChildIp6AddressIterator(void) const
+{
+    otChildIp6AddressIterator iterIndex;
+
+    if (mAddress == &mMeshLocalAddress)
+    {
+        iterIndex = 0;
```
simonlingoogle, line 140: What about save the iter index as `mIterIndex` and then just return it. The code here would be much simpler. 
**The index is actually useful for MLR features because it's used to index in a bit mask.**

src/core/thread/topology.hpp
```diff
@@ -497,60 +497,165 @@ class Neighbor : public InstanceLocatorInit
...
+         * @param[in]  aOther  The other `Iterator` to compare with.
+         *
+         * @retval TRUE   If the two `Iterator` objects are equal.
+         * @retval FALSE  If the two `Iterator` objects are not equal.
+         *
+         */
+        bool operator==(const AddressIterator &aOther) const
+        {
+            OT_ASSERT(&mChild == &aOther.mChild);
+            return ((mAddress == &mMeshLocalAddress) && (aOther.mAddress == &aOther.mMeshLocalAddress)) ||
```
simonlingoogle, line 617: comparing iter index should make this function more efficient and even simpler.

src/core/net/ip6_address.hpp
abtink: I avoided this mainly since I think/feel `kTypeFilterUnicastOnly` reads as if we are filtering out the unicast addresses.

src/core/net/ip6_address.hpp
abtink: Will add it later

src/core/thread/topology.cpp
abtink: Good suggestion.  
It makes the `GetAddress()` a bit more complex, but can simplify the rest of the methods. Overall I think would be good. Will change it later.

src/core/net/ip6_address.hpp
simonlingoogle: I am fine with kTypeXxx

src/core/net/ip6_address.hpp
abtink: Changed in new pushed commit.

src/core/thread/topology.cpp
abtink: Changed in new pushed commit.

src/core/api/thread_ftd_api.cpp
```diff
@@ -273,22 +273,28 @@ otError otThreadGetChildInfoByIndex(otInstance *aInstance, uint16_t aChildIndex,
 
 otError otThreadGetChildNextIp6Address(otInstance *               aInstance,
                                        uint16_t                   aChildIndex,
-                                       otChildIp6AddressIterator *aIterator,
+                                       otChildIp6AddressIterator *aIterIndex,
```
simonlingoogle, line 276: Should be still named `aIterator` since the type is still `otChildIp6AddressIterator`? 
API users don't need to know that the iterator is an index?

src/core/net/ip6_address.cpp
```diff
@@ -369,6 +369,31 @@ uint8_t Address::PrefixMatch(const otIp6Address &aOther) const
...
+{
+    bool matches = true;
+
+    switch (aFilter)
+    {
+    case kTypeAny:
+        break;
+
+    case kTypeUnicast:
+        matches = !IsMulticast();
```
simonlingoogle, line 382:
```suggestion
        matches = !IsUnspecified() && !IsMulticast();
```

src/core/thread/topology.hpp
```diff
@@ -593,16 +706,22 @@ class Child : public Neighbor, public IndirectSender::ChildInfo, public DataPoll
...
 
     /**
-     * This method gets the next IPv6 address in the list.
+     * This method enables range-based `for` loop iteration over all IPv6 addresses.
+     *
+     * This method should be used as follows:
      *
-     * @param[inout] aIterator           A reference to an IPv6 address iterator.
-     * @param[out]   aAddress            A reference to an IPv6 address to provide the next address (if any).
+     *     for (const Ip6::Address &address : child.IterateIp6Addresses()) { ... }
```
simonlingoogle, line 713: Add another example that uses a filter?

src/core/net/ip6_address.hpp
```diff
@@ -324,6 +324,18 @@ class Address : public otIp6Address, public Equatable<Address>, public Clearable
         kGlobalScope         = 14, ///< Global scope
     };
 
+    /**
+     * This enumeration defines IPv6 address type filter.
+     *
+     */
+    enum TypeFilter : uint8_t
```
simonlingoogle, line 331:
```suggestion
    enum Filter : uint8_t
```
nit: Filters `kTypeMulticast` and `kTypeMulticastLargerThanRealmLocal` are filtering addresses of same type (Multicast) but with different scope range? Maybe rename to `Filter` to make it more general?

simonlingoogle:
Where is the 128B code size increase from ? @abtink

src/core/api/thread_ftd_api.cpp
abtink: The public definition in header file uses the `aIterator` (which is where API user would look for the documentation). 

This is the core internal implementation of this method. I renamed this to avoid confusion with `iter` variable and also make it easier and explicit (when reading the code) that it is being used as the index.

simonlingoogle: Fine to use it for internal implementation.

src/core/net/ip6_address.hpp
abtink: I prefer to keep it `TypeFilter` to be clear that it is a "type filter"  (i.e. filtering based on address type) and avoid confusion with address filtering and other definition (e.g., we have a `Ip6::Filter` class).

src/core/thread/topology.hpp
abtink: Added in new push.

src/core/net/ip6_address.hpp
simonlingoogle: Fine with me.

abtink:
> Where is the 128B code size increase from ? @abtink 

In short, it's from all the new changes in the PR, i.e., the new type filtering mechanism and the new iterator model.

Some chunk of it is coming from the fact that iterator is a larger (more complex) object and contains an `Ip6::Address` (we need this for mesh-local address or we go back to model of caller needing to provide the address object for us to copy into which is not idle).  The larger size of `Iterator` seem to hinder complier optimizing/inlining some the logic.

PR #5277, opened 2020-07-22
[netif] enhance IterateExternalMulticastAddresses 

simonlingoogle:
This PR enhances `IterateExternalMulticastAddresses` to use further filter external multicast addresses.

**Should be rebased after #5269 is merged**

### DIFF
--- a/src/core/net/netif.hpp
+++ b/src/core/net/netif.hpp
@@ -214,13 +214,15 @@ public:
     public:
         /**
          * This constructor initializes an `ExternalMulticastAddressIterator` instance to start from the first external
-         * multicast address.
+         * multicast address that matches a given Ip6 address type filter.
          *
-         * @param[in] aNetif  A reference to the Netif instance.
+         * @param[in] aNetif   A reference to the Netif instance.
+         * @param[in] aFilter  The Ip6 address type filter.
          *
          */
-        explicit ExternalMulticastAddressIterator(const Netif &aNetif)
+        explicit ExternalMulticastAddressIterator(const Netif &aNetif, Address::TypeFilter aFilter)
             : mNetif(aNetif)
+            , mFilter(aFilter)
         {
             AdvanceFrom(mNetif.GetMulticastAddresses());
         }
@@ -315,7 +317,8 @@ public:
 
         void AdvanceFrom(const NetifMulticastAddress *aAddr)
         {
-            while (aAddr != nullptr && !mNetif.IsMulticastAddressExternal(*aAddr))
+            while (aAddr != nullptr &&
+                   !(mNetif.IsMulticastAddressExternal(*aAddr) && aAddr->GetAddress().MatchesFilter(mFilter)))
             {
                 aAddr = aAddr->GetNext();
             }
@@ -326,6 +329,7 @@ public:
 
         const Netif &                  mNetif;
         ExternalNetifMulticastAddress *mCurrent;
+        Address::TypeFilter            mFilter;
     };
 
     /**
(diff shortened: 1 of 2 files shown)


src/core/net/netif.hpp
```diff
@@ -163,6 +163,16 @@ class ExternalNetifMulticastAddress : public NetifMulticastAddress
...
 public:
+    /**
+     * This enumeration defines external Netif multicast address filter.
+     *
+     */
+    enum Filter : uint8_t
+    {
+        kFilterAny,                  ///< Accept any external Netif multicast address.
+        kFilterLargerThanRealmLocal, ///< Accept external Netif multicast addresses with scope larger than Realm Local.
+    };
```
abtink, line 174: The idea of adding this sounds good, but I suggest waiting for https://github.com/openthread/openthread/pull/5269 to be merged and then using the common definition in `Ip6::Address` for filter type.

simonlingoogle: Using the same filter type sounds good to me.

simonlingoogle: rebased on #5269  to use the same type filter

src/core/net/netif.hpp
```diff
@@ -214,13 +214,15 @@ class Netif : public InstanceLocator, public LinkedListEntry<Netif>, private Non
...
-         * multicast address.
+         * multicast address that matches a given Ip6 address type filter.
          *
-         * @param[in] aNetif  A reference to the Netif instance.
+         * @param[in] aNetif   A reference to the Netif instance.
+         * @param[in] aFilter  The Ip6 address type filter.
          *
          */
-        explicit ExternalMulticastAddressIterator(Netif &aNetif)
+        explicit ExternalMulticastAddressIterator(Netif &aNetif, Address::TypeFilter aFilter)
```
abtink, line 223: How about adding a default for `aFilter` as `aFilter = Ip6::Address::kTypeMulticast`?

src/core/net/netif.hpp
```diff
@@ -214,13 +214,15 @@ class Netif : public InstanceLocator, public LinkedListEntry<Netif>, private Non
...
             : mNetif(aNetif)
+            , mFilter(aFilter)
```
abtink, line 225: Not related to this PR, but an idea/suggestion maybe for a possible future enhancement:

We can consider declaring the iterator and iterator-builder classes as nested types within the `ExteranlNetifExternalMulticast` class. 
- This way we can use shorter names `Iterator` and `IteratorBuilder` for them in the implementation 
- But for users they are  properly name-spaced as `ExternalNetifMulticastAddress::Iterator`.

src/core/net/netif.hpp
simonlingoogle: `IterateExternalMulticastAddresses` already has default value for filter. 

Since we hardly creates `ExternalMulticastAddressIterator` by hand, it not useful to add a default  value for `aFilter`?

src/core/net/netif.hpp
```diff
@@ -554,19 +558,28 @@ class Netif : public InstanceLocator, public LinkedListEntry<Netif>, private Non
...
+     * This method enables range-based `for` loop iteration over external multicast addresses on the Netif that matches
+     * a given Ip6 address type filter.
      *
-     * This method should be used like follows:
+     * This method should be used like follows: to iterate over all external multicast addresses
      *
      *     for (Ip6::ExternalNetifMulticastAddress &addr : Get<ThreadNetif>().IterateExternalMulticastAddresses())
      *     { ... }
      *
+     * or to iterate over a subset of extern multicast addresses determined by a given address type filter
```
jwhui, line 569:
```suggestion
     * or to iterate over a subset of external multicast addresses determined by a given address type filter
```

jwhui:
@simonlingoogle , please resolve conflicts.

src/core/net/netif.hpp
simonlingoogle: I think it's a good idea. 
Even though usually `iterator` is an internal class of the container rather than entry (e.g. std::vector::iterator), I believe we don't have this option here.

PR #5286, opened 2020-07-23
[mle] use newly added Netif::IterateExternalMulticastAddresses

abtink:
This commit updates `Mle` class to use the newly added `Netif`
iterator for external multicast addresses. It also adds a helper
method in `Netif` to indicate if it is subscribed to any external
multicast address.

### DIFF
--- a/src/core/net/netif.hpp
+++ b/src/core/net/netif.hpp
@@ -219,11 +219,20 @@ public:
          * @param[in] aNetif  A reference to the Netif instance.
          *
          */
-        explicit ExternalMulticastAddressIterator(Netif &aNetif)
+        explicit ExternalMulticastAddressIterator(const Netif &aNetif)
             : mNetif(aNetif)
         {
             AdvanceFrom(mNetif.GetMulticastAddresses());
         }
+
+        /**
+         * This method indicates whether the iterator has reached end of the list.
+         *
+         * @retval TRUE   There are no more entries in the list (reached end of the list).
+         * @retval FALSE  The current address entry is valid.
+         *
+         */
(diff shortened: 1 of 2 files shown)


src/core/net/netif.hpp
```diff
@@ -569,6 +578,15 @@ class Netif : public InstanceLocator, public LinkedListEntry<Netif>, private Non
...
     }
 
+    /**
+     * This method indicates whether or not the network interfaces is subscribed to any external multicast address.
+     *
+     * @retval TRUE  The network interface is subscribed to at least one external multicast address.
+     * @retval FALSE The network interface is not subscribed to any external multicast address.
+     *
+     */
+    bool HasAnyExtenralMulticastAddress(void) const { return !ExternalMulticastAddressIterator(*this).IsDone(); }
```
jwhui, line 588:
```suggestion
    bool HasAnyExternalMulticastAddress(void) const { return !ExternalMulticastAddressIterator(*this).IsDone(); }
```

abtink: Thanks

abtink:
> Adds 32 bytes?

@jwhui , seems like iterating using linked list model produces less code compared to using the iterator.  I am actually fine closing this PR (not merging it).

Thoughts?

simonlingoogle:
> > Adds 32 bytes?
> 
> @jwhui , seems like iterating using linked list model produces less code compared to using the iterator. I am actually fine closing this PR (not merging it).
> 
> Thoughts?

The new code is more readable, which should compensate for the code increase?  
If we used `IterateExternalMulticastAddresses` more the future, we might get better code reuse and smaller code size overall?

PR #5288, opened 2020-07-24
[data-poll-handler] fix doc typo for DataPollHandler::Callbacks

Irving-cl:


simonlingoogle:
Is `SendFastPolls` also a typo? Should it be `StartFastPolls`

Irving-cl:
> Is `SendFastPolls` also a typo? Should it be `StartFastPolls`? @Irving-cl

I'm not sure. I think it's fine. But it seems to be unrelated with this change.

PR #5250, opened 2020-07-16
[logging] add MLR log region

simonlingoogle:
Adds the `MLR` log region, which is going to be used by MLR features.

**MLR PR List:**
- #5248 Send MLR.req for locally Netfi MAs 
- #5236 Parent send MLR.req on behave of Children 
- #5250 Add the `MLR` log region  **<== This PR**

PR #5292, opened 2020-07-24
[mlr] add MulticastListenersTable on BBR

simonlingoogle:
This PR adds the Multicast Listener Table for BBR:

- [x] Handles MLR.req
  - [x] Adds registered MAs to `MulticastListenerTable`
  - [x] Respond with unsuccessful status along with failed addresses 
- [x] Removes expired Listeners from `MulticastListenerTable` per second (using `Heap` for better performance
- [ ] Add a test to verify

**MLR PR List:**
- #5236 Parent send MLR.req on behave of Children
- #5292 Add MulticastListenersTable on BBR **<== This PR**
- ~~#5248 Send MLR.req for locally Netfi MAs **MERGED**~~
- ~~#5250 Add the `MLR` log region  **MERGED**~~


simonlingoogle:
Hi @librasungirl @abtink @jwhui 

This PR implements the Multicast Listeners Table on BBR. 
For now, the PBBR only saves MAs to the table and respond with failed addresses. 
BMLR on backbone and other MLR related backbone features on BBR will be implemented upcoming PRs. 

Please take a look. 

Thanks,

src/core/backbone_router/multicast_listeners_table.hpp
```diff
@@ -0,0 +1,226 @@
...
+    /**
+     * This method removes a given Ip6 multicast address.
+     *
+     * @param[in] aAddress  The Ip6 multicast address to remove.
+     *
+     */
+    void Remove(const Ip6::Address &aAddress);
+
+    /**
+     * This method removes expired Ip6 multicast addresses.
```
librasungirl, line 138:
```suggestion
     * This method removes the expired Ip6 multicast addresses.
```

src/core/backbone_router/multicast_listeners_table.hpp
```diff
@@ -0,0 +1,226 @@
...
+
+#if OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE
+    otError GetNext(otBackboneRouterMulticastListenerIterator &aIterator,
+                    otBackboneRouterMulticastListenerInfo &    aListenerInfo);
+#endif
+
+private:
+    enum
+    {
+        kMulticastListenersTableSize = OPENTHREAD_CONFIG_MAX_MULTICAST_LISTENERS
```
librasungirl, line 184:
```suggestion
        kMulticastListenersTableSize = OPENTHREAD_CONFIG_MAX_MULTICAST_LISTENERS,
```

src/core/backbone_router/multicast_listeners_table.hpp
```diff
@@ -0,0 +1,226 @@
...
+    };
```
librasungirl, line 185: assert that `kMulticastListenersTableSize` must be no less than 75 which is the minimal requirement defined in Conformance Specification?

src/core/config/backbone_router.h
```diff
@@ -45,4 +45,16 @@
 #define OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE 0
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_MAX_MULTICAST_LISTENERS
+ *
+ * This setting configures the maximum number of Multicast Listeners on a Backbone Router.
```
librasungirl, line 51: add note regarding the minimal requirement defined in spec?

src/cli/cli.cpp
```diff
@@ -514,14 +514,96 @@ void Interpreter::ProcessBackboneRouter(uint8_t aArgsLength, char *aArgs[])
...
+        {
+            PrintMulticastListenersTable();
+        }
+        else if (!strcmp(aArgs[1], "clear"))
+        {
+            otBackboneRouterMulticastListenerClear(mInstance);
+        }
+        else if (!strcmp(aArgs[1], "list"))
+        {
+            PrintMulticastListenersTable();
```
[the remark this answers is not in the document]
simonlingoogle: will remove `list`

include/openthread/backbone_router_ftd.h
```diff
@@ -174,6 +174,112 @@ void otBackboneRouterSetRegistrationJitter(otInstance *aInstance, uint8_t aJitte
...
+ * @param[in] aAddress       The Ip6 multicast address of the Multicast Listner.
+ * @param[in] aTimeoutMilli  The timeout (in milliseconds) of the Multicast Listener.
+ *
+ * @sa otBackboneRouterMulticastListenerAdd
+ * @sa otBackboneRouterGetNextMulticastListener
+ *
+ */
+otError otBackboneRouterMulticastListenerAdd(otInstance *        aInstance,
+                                             const otIp6Address *aAddress,
+                                             uint32_t            aTimeoutMilli);
```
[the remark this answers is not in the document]
simonlingoogle: It's not used right now. Prefer to add only when its used. 
BTW, it currently only used by test scripts.
Thoughts?

src/core/backbone_router/multicast_listeners_table.hpp
simonlingoogle: There could be multiple expired addresses. No need to use `the` ?

src/core/thread/mle_types.hpp
```diff
@@ -257,22 +257,24 @@ enum
...
-    kParentAggregateDelay             = 5,    ///< In seconds.
-    kNoBufDelay                       = 5,    ///< In seconds.
-    kImmediateReRegisterDelay         = 1,    ///< In seconds.
-    KResponseTimeoutDelay             = 30,   ///< In seconds.
-    kDuaDadPeriod                     = 100,  ///< In seconds. Time period after which the address
-                                              ///< becomes "Preferred" if no duplicate address error.
+    kRegistrationDelayDefault         = 1200,              ///< In seconds.
+    kMlrTimeoutDefault                = 3600,              ///< In seconds.
+    kMlrTimeoutMin                    = 300,               ///< In seconds.
+    kMlrTimeoutMax                    = 0x7fffffff / 1000, ///< In seconds (about 24 days).
```
simonlingoogle, line 263: Here, I limited the max MLR timeout to be about 24 days, which is not required by the Spec. 
This make expire time calculation simple, not having to cope with overflow, because the MLR Timeout in Milliseconds will not exceed the `kDistantFuture`. 

However it has two side-effects:

- Users can never set MLR timeout larger than `kMlrTimeoutMax` on a OpenThread BBR 
- OT devices that registers to BBR of another vendor would use `kMlrTimeoutMax` if the actual MLR timeout is larger than `kMlrTimeoutMax`. 

Personally I believe both side-effects are acceptable because a MLR timeout of 24 days should be more than enough for most situations. 

If MLR timeout is allowed to exceed the defined `kMlrTimeoutMax`, we can't not use `TimeMilli` for maintaining expire times because the expire time will be larger than `kDistantFuture`. We will have to use a per-second timestamp for expire time. However, currently we don't have an working implementation of per-second timer or timestamp. 

Thoughts? @jwhui @librasungirl

include/openthread/backbone_router_ftd.h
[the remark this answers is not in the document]
librasungirl: +1

src/core/backbone_router/multicast_listeners_table.hpp
librasungirl: yes

src/core/backbone_router/bbr_manager.hpp
```diff
@@ -117,6 +135,9 @@ class Manager : public InstanceLocator, public Notifier::Receiver
     ThreadStatusTlv::DuaStatus mDuaResponseStatus;
     bool                       mDuaResponseIsSpecified : 1;
 #endif
+
+    MulticastListenersTable mMulticastListenersTable;
```
jwhui, line 139: nit: move this above `#if OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE` above to keep always-included members variables together?

src/core/backbone_router/multicast_listeners_table.cpp
```diff
@@ -0,0 +1,307 @@
...
+    Listener saveElem;
+
+    OT_ASSERT(aIndex < mNumValidListeners);
+
+    saveElem = mListeners[aIndex];
+
+    for (;;)
+    {
+        uint16_t i = (j - 1) / 2;
+        if (i == j || !Listener::IsAfter(mListeners[i], saveElem))
```
jwhui, line 231: nit: empty line after variable declarations

src/core/backbone_router/multicast_listeners_table.cpp
```diff
@@ -0,0 +1,307 @@
...
+/**
+ * @file
+ *   This file implements the Multicast Listeners Table.
+ */
+
+#include "multicast_listeners_table.hpp"
+
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+
+#include <algorithm>
```
jwhui, line 38: nit: put a comment here to indicate this is only used with `CheckInvariants()` for debugging purposes?

src/core/backbone_router/multicast_listeners_table.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    /**
+     * This method removes expired Ip6 multicast addresses.
+     *
+     */
+    void Expire(void);
+
+    /**
+     * This method counts the number of valid Ip6 multicast addresses.
+     * @return
```
jwhui, line 147: nit: missing `@return` description.

src/core/backbone_router/multicast_listeners_table.hpp
```diff
@@ -0,0 +1,231 @@
...
+     */
+    uint16_t Count(void) const { return mNumValidListeners; }
+
+    /**
+     * This method enables range-based `for` loop iteration over all Listener entries.
+     *
+     * This method should be used as follows:
+     *
+     *     for (Child &child : aChildTable.Iterate(aFilter)) { ... }
```
jwhui, line 156: nit: update the example

src/core/backbone_router/multicast_listeners_table.hpp
```diff
@@ -0,0 +1,231 @@
...
+     *
+     * @param[in] aFilter  A child state filter.
```
jwhui, line 158: nit: remove

src/core/backbone_router/multicast_listeners_table.hpp
```diff
@@ -0,0 +1,231 @@
...
+    /**
+     * This method clears the Multicast Listeners Table.
+     *
+     */
+    void Clear(void);
+
+    /**
+     * This method sets the Multicast Listener callback.
+     *
+     * @param aCallback  The callback function.
```
jwhui, line 174: nit: document `aContext`

src/core/backbone_router/multicast_listeners_table.hpp
```diff
@@ -0,0 +1,231 @@
...
+#endif
+
+private:
+    enum
+    {
+        kMulticastListenersTableSize = OPENTHREAD_CONFIG_MAX_MULTICAST_LISTENERS,
+    };
+
+    static_assert(kMulticastListenersTableSize >= 75,
+                  "Multicast Listener Table size must be larger than or equal to 75.");
```
jwhui, line 190:
```suggestion
                  "Thread 1.2 Conformance requires the Multicast Listener Table size to be larger than or equal to 75.");
```

src/core/thread/mlr_manager.cpp
```diff
@@ -370,25 +349,41 @@ void MlrManager::HandleMulticastListenerRegistrationResponse(Coap::Message *
...
 
+    if (ThreadTlv::FindTlvValueOffset(*aMessage, IPv6AddressesTlv::kIPv6Addresses, addressesOffset, addressesLength) ==
+        OT_ERROR_NONE)
+    {
+        VerifyOrExit(addressesLength % sizeof(Ip6::Address) == 0, error = OT_ERROR_PARSE);
+        VerifyOrExit(addressesLength / sizeof(Ip6::Address) <= kIPv6AddressesNumMax, error = OT_ERROR_PARSE);
+
+        for (uint16_t offset = 0; offset < addressesLength; offset += sizeof(Ip6::Address))
+        {
+            VerifyOrExit(aMessage->Read(addressesOffset + offset, sizeof(Ip6::Address),
```
jwhui, line 371: Is it possible for `Read()` to return an unexpected length here?

src/core/thread/mlr_manager.hpp
```diff
@@ -168,6 +164,12 @@ class MlrManager : public InstanceLocator, public Notifier::Receiver
     void        HandleTimer(void);
 
     void LogMulticastAddresses(void);
+    void CheckInvariants(void);
```
jwhui, line 167:
```suggestion
    void CheckInvariants(void) const;
```

src/core/thread/mlr_manager.cpp
```diff
@@ -531,6 +591,84 @@ void MlrManager::AppendToUniqueAddressList(Ip6::Address (&aAddresses)[kIPv6Addre
...
+
+        for (uint8_t i = 0; i < aFailedAddressNum; i++)
+        {
+            otLogWarnBbr("MLR failed: %s", aFailedAddresses[i].ToString().AsCString());
+        }
+    }
+#endif
+}
+
+void MlrManager::CheckInvariants(void)
```
jwhui, line 644:
```suggestion
void MlrManager::CheckInvariants(void) const
```

src/core/backbone_router/multicast_listeners_table.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+void MulticastListenersTable::FixHeap(uint16_t aIndex)
+{
+    if (!SiftHeapElemDown(aIndex))
+    {
+        SiftHeapElemUp(aIndex);
+    }
+}
+
+void MulticastListenersTable::CheckInvariants(void)
```
jwhui, line 171:
```suggestion
void MulticastListenersTable::CheckInvariants(void) const
```

src/core/backbone_router/multicast_listeners_table.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    void LogMulticastListenersTable(const char *        aAction,
+                                    const Ip6::Address &aAddress,
+                                    TimeMilli           aExpireTime,
+                                    otError             aError);
+
+    void FixHeap(uint16_t aIndex);
+    bool SiftHeapElemDown(uint16_t aIndex);
+    void SiftHeapElemUp(uint16_t aIndex);
+    void CheckInvariants(void);
```
jwhui, line 212:
```suggestion
    void CheckInvariants(void) const;
```

src/core/backbone_router/multicast_listeners_table.cpp
simonlingoogle: I think I should put this include in the `#if OPENTHREAD_EXAMPLES_SIMULATION ... #endif` block.

src/core/thread/mle_types.hpp
jwhui: I am fine with limiting MLR timeout to 24 days for now.

tests/unit/test_multicast_listeners_table.cpp
```diff
@@ -0,0 +1,211 @@
...
+    otBackboneRouterMulticastListenerIterator iter = OT_BACKBONE_ROUTER_MULTICAST_LISTENER_ITERATOR_INIT;
+    otBackboneRouterMulticastListenerInfo     info;
+    size_t                                    table_size = 0;
+
+    while (otBackboneRouterMulticastListenerGetNext(aInstance, &iter, &info) == OT_ERROR_NONE)
+    {
+        VerifyOrQuit(false, "Table should be empty");
+    }
+
+    SuccessOrQuit(otBackboneRouterMulticastListenerAdd(aInstance, &MA401, 30000), "Add failed");
```
abtink, line 176: These APIs are only defined under OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE.

I think should add a `#if OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE` around this block of code (to avoid linker error).

tests/unit/test_multicast_listeners_table.cpp
```diff
@@ -0,0 +1,211 @@
...
+void testMulticastListenersTableAPIs(Instance *aInstance);
+
+void TestMulticastListenersTable(void)
+{
+    InitTestTimer();
+
+    sInstance = testInitInstance();
+    VerifyOrQuit(sInstance != nullptr, "Null OpenThread instance");
+
+    MulticastListenersTable table = sInstance->Get<MulticastListenersTable>();
```
abtink, line 79: This creates a copy of table. If it is not intentional, I suggest using a reference instead?

(may want to consider declaring `MulticastListenersTable` as `NonCopyable`?)

src/core/backbone_router/multicast_listeners_table.cpp
```diff
@@ -0,0 +1,310 @@
...
+    uint16_t i = aIndex;
+    Listener saveElem;
+
+    OT_ASSERT(aIndex < mNumValidListeners);
+
+    saveElem = mListeners[aIndex];
+
+    for (;;)
+    {
+        uint16_t j = 2 * i + 1;
```
abtink, line 191: how about calling it `left = 2 * index + 1` and `right = 2 * index + 2` (for left and right node indexes of binary tree)?

src/core/backbone_router/multicast_listeners_table.cpp
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    mNumValidListeners = 0;
+
+    CheckInvariants();
+}
+
+void MulticastListenersTable::SetCallback(otBackboneRouterMulticastListenerCallback aCallback, void *aContext)
+{
+    OT_ASSERT(mNumValidListeners == 0);
```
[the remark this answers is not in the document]
simonlingoogle: Calling `SetCallback` when `mNumValidListeners > 0` is a potential issue for API users, in which case, the API user is not aware of existing Multicast Listeners in the table.

I will revise it to emit existing Multicast Listeners by `OT_BACKBONE_ROUTER_MULTICAST_LISTENER_ADDED` events to the callback in `SetCallback`.

src/core/backbone_router/multicast_listeners_table.cpp
```diff
@@ -0,0 +1,310 @@
...
+    if (!SiftHeapElemDown(aIndex))
+    {
+        SiftHeapElemUp(aIndex);
+    }
+}
+
+void MulticastListenersTable::CheckInvariants(void) const
+{
+#if OPENTHREAD_EXAMPLES_SIMULATION
+    OT_ASSERT(std::is_heap(&mListeners[0], &mListeners[mNumValidListeners], Listener::IsAfter));
```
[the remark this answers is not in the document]
simonlingoogle: ##### For `CheckInvariants()`:
This code take advantage of the simulation platform to run more consistency  checks (with higher overhead).
I believe it is a good practice:
- It allows us to catch bugs more quickly, and gives us much more information when bugs are found. Unittests do cover the basic usage but might unable to cover some corner cases, or it might not be updated properly in future code refactoring. 
- It does not introduce any overhead on any platform except simulation, on which platform we don't care about performance much.

##### For `std::is_heap`
I used `std::is_heap` to avoid implementing it manually. We upgraded to C++11 standard recently so that we can take advantage of modern C++ features. Maybe we can have a further discussion on which std libraries can we use and which we can't. 
**For now, I will remove usage of `std:is_heap` and write our own version.**

src/core/backbone_router/multicast_listeners_table.cpp
simonlingoogle: renamed `j` to `child`. Didn't use `left` because `child` might be pointing to the right child.

tests/unit/test_multicast_listeners_table.cpp
simonlingoogle: added `#if OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE`

src/core/backbone_router/multicast_listeners_table.cpp
```diff
@@ -0,0 +1,310 @@
...
+    {
+        mListeners[i] = saveElem;
+    }
+
+    return i > aIndex;
+}
+
+void MulticastListenersTable::SiftHeapElemUp(uint16_t aIndex)
+{
+    uint16_t j = aIndex;
```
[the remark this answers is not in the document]
simonlingoogle: renamed

src/core/backbone_router/multicast_listeners_table.cpp
```diff
@@ -0,0 +1,310 @@
...
+    Listener saveElem;
+
+    OT_ASSERT(aIndex < mNumValidListeners);
+
+    saveElem = mListeners[aIndex];
+
+    for (;;)
+    {
+        uint16_t i = (j - 1) / 2;
```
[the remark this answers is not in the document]
simonlingoogle: renamed

src/core/backbone_router/multicast_listeners_table.cpp
```diff
@@ -0,0 +1,310 @@
...
+void MulticastListenersTable::CheckInvariants(void) const
+{
+#if OPENTHREAD_EXAMPLES_SIMULATION
+    OT_ASSERT(std::is_heap(&mListeners[0], &mListeners[mNumValidListeners], Listener::IsAfter));
+#endif
+}
+
+bool MulticastListenersTable::SiftHeapElemDown(uint16_t aIndex)
+{
+    uint16_t i = aIndex;
```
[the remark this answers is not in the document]
simonlingoogle: renamed

src/core/backbone_router/multicast_listeners_table.cpp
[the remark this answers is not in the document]
simonlingoogle: Changed `SetCallback` to trigger `OT_BACKBONE_ROUTER_MULTICAST_LISTENER_ADDED` for existing Multicast Listeners.

src/core/backbone_router/bbr_leader.cpp
```diff
@@ -217,6 +217,14 @@ void Leader::UpdateBackboneRouterPrimary(void)
         state = kStateUnchanged;
     }
 
+    // Restrain the range of MLR timeout to be always valid
+    config.mMlrTimeout = config.mMlrTimeout < static_cast<uint32_t>(Mle::kMlrTimeoutMin)
+                             ? static_cast<uint32_t>(Mle::kMlrTimeoutMin)
+                             : config.mMlrTimeout;
+    config.mMlrTimeout = config.mMlrTimeout > static_cast<uint32_t>(Mle::kMlrTimeoutMax)
+                             ? static_cast<uint32_t>(Mle::kMlrTimeoutMax)
+                             : config.mMlrTimeout;
```
librasungirl, line 226: May add log when `config.mMlrTimeout` is changed and different from the value in Thread Network Data?

src/core/backbone_router/bbr_manager.cpp
```diff
@@ -90,18 +111,63 @@ void Manager::HandleMulticastListenerRegistration(const Coap::Message &aMessage,
...
+        case OT_ERROR_NONE:
+            failed = false;
+            break;
+        case OT_ERROR_INVALID_ARGS:
+            if (status == ThreadStatusTlv::kMlrSuccess)
+            {
+                status = ThreadStatusTlv::kMlrInvalid;
+            }
+            break;
+        case OT_ERROR_NO_BUFS:
```
librasungirl, line 142: When `NO_BUFS` happens, may skip further `mMulticastListenersTable.Add` and just take following addresses as failed?

src/core/backbone_router/bbr_manager.cpp
```diff
@@ -90,18 +111,63 @@ void Manager::HandleMulticastListenerRegistration(const Coap::Message &aMessage,
...
     // TODO: (MLR) handle Timeout TLV
-    // TODO: (MLR) handle addresses, updated listener groups, send BMLR.req
+
+    VerifyOrExit(ThreadTlv::FindTlvValueOffset(aMessage, IPv6AddressesTlv::kIPv6Addresses, addressesOffset,
+                                               addressesLength) == OT_ERROR_NONE,
+                 error = OT_ERROR_PARSE);
+    VerifyOrExit(addressesLength % sizeof(Ip6::Address) == 0, error = OT_ERROR_PARSE);
+    VerifyOrExit(addressesLength / sizeof(Ip6::Address) <= kIPv6AddressesNumMax, error = OT_ERROR_PARSE);
+
+    SuccessOrExit(error = Get<BackboneRouter::Leader>().GetConfig(config));
```
librasungirl, line 121: assert error is `OT_ERROR_NONE`?

src/core/backbone_router/bbr_manager.cpp
```diff
@@ -90,18 +111,63 @@ void Manager::HandleMulticastListenerRegistration(const Coap::Message &aMessage,
     // TODO: (MLR) send configured MLR response for Reference Device
     // TODO: (MLR) handle Commissioner Session TLV
     // TODO: (MLR) handle Timeout TLV
-    // TODO: (MLR) handle addresses, updated listener groups, send BMLR.req
+
+    VerifyOrExit(ThreadTlv::FindTlvValueOffset(aMessage, IPv6AddressesTlv::kIPv6Addresses, addressesOffset,
+                                               addressesLength) == OT_ERROR_NONE,
+                 error = OT_ERROR_PARSE);
+    VerifyOrExit(addressesLength % sizeof(Ip6::Address) == 0, error = OT_ERROR_PARSE);
+    VerifyOrExit(addressesLength / sizeof(Ip6::Address) <= kIPv6AddressesNumMax, error = OT_ERROR_PARSE);
```
librasungirl, line 119: Just for discussion here - how about responding with `kMlrGeneralFailure` for these scenarios to suppress CoAP retransmission from the requester?

src/core/backbone_router/bbr_manager.cpp
```diff
@@ -90,18 +111,63 @@ void Manager::HandleMulticastListenerRegistration(const Coap::Message &aMessage,
...
+
+    SuccessOrExit(error = Get<BackboneRouter::Leader>().GetConfig(config));
+    expireTime = TimerMilli::GetNow() + TimeMilli::SecToMsec(config.mMlrTimeout);
+
+    for (uint16_t offset = 0; offset < addressesLength; offset += sizeof(Ip6::Address))
+    {
+        bool failed = true;
+
+        IgnoreReturnValue(aMessage.Read(addressesOffset + offset, sizeof(Ip6::Address), &address) ==
+                          sizeof(Ip6::Address));
```
librasungirl, line 129: nit: the `==` should be unnecessary?

```suggestion
        IgnoreReturnValue(aMessage.Read(addressesOffset + offset, sizeof(Ip6::Address), &address));
```

src/core/backbone_router/bbr_manager.cpp
librasungirl: Checked MATN-TC-21, Step 20, `kMlrGeneralFailure` is one of the three flexible responses.

src/core/backbone_router/bbr_leader.cpp
```diff
@@ -217,6 +217,14 @@ void Leader::UpdateBackboneRouterPrimary(void)
         state = kStateUnchanged;
     }
 
+    // Restrain the range of MLR timeout to be always valid
+    config.mMlrTimeout = config.mMlrTimeout < static_cast<uint32_t>(Mle::kMlrTimeoutMin)
```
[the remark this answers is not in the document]
simonlingoogle: Yes, otherwise come compiler gives warnings

src/core/backbone_router/bbr_manager.cpp
simonlingoogle: I prefer the current code:

- To implement what you have suggested, we need extra code to stop the loop and append left addresses (unprocessed) to `failedAddresses`, which will increase code size
- The current code can already do that (even though with a little more CPU cycles)
- If there is a valid address that was already registered, the current code can correctly handle it without adding it to the `failedAddresses`, even if the table is running out of memory space.

src/core/backbone_router/bbr_manager.cpp
simonlingoogle: This is impossible to fail here. Will use `IgnoreError` to ignore it.

src/core/backbone_router/bbr_manager.cpp
simonlingoogle: `ST_MLR_GENERAL_FAILURE` is a good idea.

include/openthread/backbone_router_ftd.h
```diff
@@ -193,6 +193,110 @@ void otBackboneRouterConfigNextDuaRegistrationResponse(otInstance *
...
+
+/**
+ * This method adds a Multicast Listener.
+ *
+ * Note: available only when `OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE` is enabled.
+ *       Only used for test and certification.
+ *
+ * @param[in] aInstance  A pointer to an OpenThread instance.
+ * @param[in] aAddress   The Multicast Listener address.
+ * @param[in] aTimeout   The timeout (in milliseconds) of the Multicast Listener, or 0 to use the default MLR timeout.
```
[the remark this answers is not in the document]
simonlingoogle: Changed API to use seconds.

include/openthread/backbone_router_ftd.h
```diff
@@ -193,6 +193,110 @@ void otBackboneRouterConfigNextDuaRegistrationResponse(otInstance *
...
+typedef uint16_t otBackboneRouterMulticastListenerIterator; ///< Used to iterate through Multicast Listeners.
+
+/**
+ * This structure represents a Backbone Router Multicast Listener info.
+ *
+ */
+typedef struct otBackboneRouterMulticastListenerInfo
+{
+    otIp6Address mAddress;
+    uint32_t     mTimeout;
```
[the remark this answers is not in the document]
simonlingoogle: Added doc to specify the unit to seconds.

src/cli/cli.cpp
```diff
@@ -538,18 +540,84 @@ void Interpreter::ProcessBackboneRouter(uint8_t aArgsLength, char *aArgs[])
...
+
+void Interpreter::PrintMulticastListenersTable(void)
+{
+    otBackboneRouterMulticastListenerIterator iter = OT_BACKBONE_ROUTER_MULTICAST_LISTENER_ITERATOR_INIT;
+    otBackboneRouterMulticastListenerInfo     listenerInfo;
+
+    while (otBackboneRouterMulticastListenerGetNext(mInstance, &iter, &listenerInfo) == OT_ERROR_NONE)
+    {
+        OutputIp6Address(listenerInfo.mAddress);
+        mServer->OutputFormat(" %lu\r\n", listenerInfo.mTimeout);
```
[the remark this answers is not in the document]
simonlingoogle: revised

include/openthread/backbone_router_ftd.h
```diff
@@ -193,6 +193,110 @@ void otBackboneRouterConfigNextDuaRegistrationResponse(otInstance *
...
+ *
+ * @param[in] aInstance  A pointer to an OpenThread instance.
+ * @param[in] aAddress   The Multicast Listener address.
+ * @param[in] aTimeout   The timeout (in milliseconds) of the Multicast Listener, or 0 to use the default MLR timeout.
+ *
+ * @sa otBackboneRouterMulticastListenerClear
+ * @sa otBackboneRouterMulticastListenerGetNext
+ *
+ */
+otError otBackboneRouterMulticastListenerAdd(otInstance *aInstance, const otIp6Address *aAddress, uint32_t aTimeout);
```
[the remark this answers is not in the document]
simonlingoogle: documented.

src/core/api/backbone_router_ftd_api.cpp
```diff
@@ -121,6 +130,37 @@ void otBackboneRouterConfigNextDuaRegistrationResponse(otInstance *
...
+{
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
+    instance.Get<BackboneRouter::MulticastListenersTable>().Clear();
+}
+
+otError otBackboneRouterMulticastListenerAdd(otInstance *aInstance, const otIp6Address *aAddress, uint32_t aTimeout)
+{
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
```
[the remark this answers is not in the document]
simonlingoogle: Added assert

src/core/api/backbone_router_ftd_api.cpp
```diff
@@ -121,6 +130,37 @@ void otBackboneRouterConfigNextDuaRegistrationResponse(otInstance *
...
+    return instance.Get<BackboneRouter::MulticastListenersTable>().Add(static_cast<const Ip6::Address &>(*aAddress),
+                                                                       TimerMilli::GetNow() + aTimeout);
+}
+
+otError otBackboneRouterMulticastListenerGetNext(otInstance *                           aInstance,
+                                                 otChildIp6AddressIterator *            aIterator,
+                                                 otBackboneRouterMulticastListenerInfo *aListenerInfo)
+{
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
```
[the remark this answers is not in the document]
simonlingoogle: Added assert.

src/core/backbone_router/bbr_leader.cpp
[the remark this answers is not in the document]
simonlingoogle: Yes, some compilers give warnings if not using `static_cast<uint32_t> `

src/core/backbone_router/bbr_leader.cpp
simonlingoogle: Added a warning log

src/core/backbone_router/bbr_manager.cpp
simonlingoogle: changed to return `ST_MLR_GENERAL_FAILURE`.

src/core/thread/mlr_manager.cpp
```diff
@@ -368,25 +347,40 @@ void MlrManager::HandleMulticastListenerRegistrationResponse(Coap::Message *
...
+    if (ThreadTlv::FindTlvValueOffset(*aMessage, IPv6AddressesTlv::kIPv6Addresses, addressesOffset, addressesLength) ==
+        OT_ERROR_NONE)
+    {
+        VerifyOrExit(addressesLength % sizeof(Ip6::Address) == 0, error = OT_ERROR_PARSE);
+        VerifyOrExit(addressesLength / sizeof(Ip6::Address) <= kIPv6AddressesNumMax, error = OT_ERROR_PARSE);
+
+        for (uint16_t offset = 0; offset < addressesLength; offset += sizeof(Ip6::Address))
+        {
+            IgnoreReturnValue(aMessage->Read(addressesOffset + offset, sizeof(Ip6::Address),
+                                             &failedAddresses[failedAddressNum]) == sizeof(Ip6::Address));
```
librasungirl, line 370:
```suggestion
            IgnoreReturnValue(aMessage->Read(addressesOffset + offset, sizeof(Ip6::Address),
                                             &failedAddresses[failedAddressNum]));
```

src/core/backbone_router/multicast_listeners_table.cpp
```diff
@@ -0,0 +1,318 @@
...
+        }
+    }
+
+    VerifyOrExit(mNumValidListeners < OT_ARRAY_LENGTH(mListeners), error = OT_ERROR_NO_BUFS);
+
+    mListeners[mNumValidListeners].SetAddress(aAddress);
+    mListeners[mNumValidListeners].SetExpireTime(aExpireTime);
+    mNumValidListeners++;
+
+    FixHeap(mNumValidListeners - 1);
```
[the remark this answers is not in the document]
simonlingoogle: I think it might also work, but the current code make sure all Heap operations operate on a heap with well defined boundaries.   
- `mNumValidListeners` is always the number of elements in the Heap.
- `FixHeap` fixes the element indexed by a given index in the heap. So heap operations actually requires that the operating index  must be valid:
```python3
OT_ASSERT(aIndex < mNumValidListeners);
```

So, in the current code, `FixHeap(mNumValidListeners); mNumValidListeners++` would cause assertion failure.

src/core/thread/mlr_manager.cpp
```diff
@@ -529,6 +588,84 @@ void MlrManager::AppendToUniqueAddressList(Ip6::Address (&aAddresses)[kIPv6Addre
     return;
 }
 
+bool MlrManager::AddressListContains(const Ip6::Address *aAddressList,
+                                     uint8_t             aAddressListSize,
+                                     const Ip6::Address &aAddress)
+{
+    bool contains = false;
+
+    VerifyOrExit(aAddressListSize > 0, contains = true);
```
librasungirl, line 597: add comment like below to avoid confusion, for  better readability as well?

> For non `kMlrSuccess` response which doesn't have IPv6AddressesTlv,  we take all MAs in registering state as failed address .

src/core/thread/mlr_manager.cpp
```diff
@@ -529,6 +588,84 @@ void MlrManager::AppendToUniqueAddressList(Ip6::Address (&aAddresses)[kIPv6Addre
...
+    for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValid))
+    {
+        for (const Ip6::Address &address : child.IterateIp6Addresses(Ip6::Address::kTypeMulticastLargerThanRealmLocal))
+        {
+            registeringNum += (child.GetAddressMlrState(address) == kMlrStateRegistering);
+        }
+    }
+#endif
+
+    OT_ASSERT(registeringNum == 0 || mMlrPending);
```
librasungirl, line 665:
```suggestion
    OT_ASSERT(registeringNum == 0 || mSendDelay != 0 || mMlrPending);
```

tests/scripts/thread-cert/node.py
```diff
@@ -528,6 +530,36 @@ def clear_dua_iid(self):
...
+
+        return table
+
+    def multicast_listener_clear(self):
+        cmd = f'bbr mgmt mlr listener clear'
+        self.send_command(cmd)
+        self._expect("Done")
+
+    def multicast_listener_add(self, ip: Union[ipaddress.IPv6Address, str], timeout: float = 0):
+        timeout_ms = int(timeout * 1000)
```
librasungirl, line 555: update to use `s`, other than `ms` accordingly.

src/core/backbone_router/multicast_listeners_table.cpp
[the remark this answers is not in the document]
librasungirl: make sense.

src/core/thread/mlr_manager.cpp
simonlingoogle: This assert makes sure that if there is some MA in `Registering` state, then the device must be sending MLR.req and waiting for the response (or timeout). 
It's not directly related to `mSendDelay`. If there is a send delay but no pending MLR.req, there should be no `Registering` MAs. 
So, `registeringNum == 0 || mMlrPending` should always hold true, no need to add more conditions.

librasungirl: right.

jwhui:
@simonlingoogle , please help resolve conflicts.

PR #5299, opened 2020-07-27
[joiner-adv] initial implementation of Joiner Advertisement

lanyuwen:
This PR implements Thread based Joiner Advertisement from Joiner side, including:

- A new MeshCop TLV: Joiner Advertisement TLV which includes IEEE OUI and the Advertisement Data 
- New APIs to set Joiner Advertisement which will be sent along with Discovery Request messages

### DIFF
--- a/include/openthread/thread.h
+++ b/include/openthread/thread.h
@@ -274,7 +274,7 @@ bool otThreadIsDiscoverInProgress(otInstance *aInstance);
  * @param[in]  aInstance        A pointer to an OpenThread instance.
  * @param[in]  aOui             The Vendor IEEE OUI value that will be included in the Joiner Advertisement. Only the
  *                              least significant 3 bytes will be used, and the most significant byte will be ignored.
- * @param[in]  aAdvData         The AdvData that will be included in the Joiner Advertisement.
+ * @param[in]  aAdvData         A pointer to the AdvData that will be included in the Joiner Advertisement.
  * @param[in]  aAdvDataLength   The length of AdvData in bytes.
  *
  * @retval OT_ERROR_NONE         Successfully set Joiner Advertisement.
--- a/include/openthread/thread.h
+++ b/include/openthread/thread.h
@@ -286,7 +286,7 @@ otError otThreadSetJoinerAdvertisement(otInstance *   aInstance,
                                        const uint8_t *aAdvData,
                                        uint8_t        aAdvDataLength);
 
-#define OT_JOINER_ADVDATA_MAX_LENGTH 64 ///<Maximum AdvData Length of Joiner Advertisement
+#define OT_JOINER_ADVDATA_MAX_LENGTH 64 ///< Maximum AdvData Length of Joiner Advertisement
 
 /**
  * Get the Thread Child Timeout used when operating in the Child role.
--- a/include/openthread/thread.h
+++ b/include/openthread/thread.h
@@ -265,6 +265,28 @@ otError otThreadDiscover(otInstance *             aInstance,
 bool otThreadIsDiscoverInProgress(otInstance *aInstance);
 
 /**
+ * This method sets the Thread Joiner Advertisement when discovering Thread network.
+ *
+ * Thread Joiner Advertisement is used to allow a Joiner advertise its own application-specific information
+ * (such as Vendor ID, Product ID, Discriminator, etc.) unsolicitedly via a newly-proposed Joiner Advertisement TLV,
+ * and to make this information available to Commissioners or Commissioner Candidates without human interactions.
+ *
+ * @param[in]  aInstance        A pointer to an OpenThread instance.
+ * @param[in]  aOui             The Vendor IEEE OUI value which will be included in the Joiner Advertisement. Only the
+ *                              least significant 3 bytes will be used, and the most significant byte will be ignored.
+ * @param[in]  aAdvData         The AdvData which will be included in the Joiner Advertisement.
+ * @param[in]  aAdvDataLength   The length of AdvData
+ *
+ * @retval OT_ERROR_NONE         Successfully set Joiner Advertisement.
+ * @retval OT_ERROR_INVALID_ARGS Invalid AdvData.
+ *
+ */
+otError otThreadSetJoinerAdvertisement(otInstance *   aInstance,
+                                       uint32_t       aOui,
+                                       const uint8_t *aAdvData,
+                                       uint8_t        aAdvDataLength);
+
+/**
  * Get the Thread Child Timeout used when operating in the Child role.
  *
  * @param[in]  aInstance A pointer to an OpenThread instance.
--- a/include/openthread/thread.h
+++ b/include/openthread/thread.h
@@ -286,6 +286,8 @@ otError otThreadSetJoinerAdvertisement(otInstance *   aInstance,
                                        const uint8_t *aAdvData,
                                        uint8_t        aAdvDataLength);
 
+#define OT_JOINER_ADVDATA_MAX_LENGTH 64 ///<Maximum AdvData Length of Joiner Advertisement
+
 /**
  * Get the Thread Child Timeout used when operating in the Child role.
  *
--- a/include/openthread/thread.h
+++ b/include/openthread/thread.h
@@ -267,15 +267,15 @@ bool otThreadIsDiscoverInProgress(otInstance *aInstance);
 /**
  * This method sets the Thread Joiner Advertisement when discovering Thread network.
  *
- * Thread Joiner Advertisement is used to allow a Joiner advertise its own application-specific information
- * (such as Vendor ID, Product ID, Discriminator, etc.) unsolicitedly via a newly-proposed Joiner Advertisement TLV,
- * and to make this information available to Commissioners or Commissioner Candidates without human interactions.
+ * Thread Joiner Advertisement is used to allow a Joiner to advertise its own application-specific information
+ * (such as Vendor ID, Product ID, Discriminator, etc.) via a newly-proposed Joiner Advertisement TLV,
+ * and to make this information available to Commissioners or Commissioner Candidates without human interaction.
  *
  * @param[in]  aInstance        A pointer to an OpenThread instance.
- * @param[in]  aOui             The Vendor IEEE OUI value which will be included in the Joiner Advertisement. Only the
+ * @param[in]  aOui             The Vendor IEEE OUI value that will be included in the Joiner Advertisement. Only the
  *                              least significant 3 bytes will be used, and the most significant byte will be ignored.
- * @param[in]  aAdvData         The AdvData which will be included in the Joiner Advertisement.
- * @param[in]  aAdvDataLength   The length of AdvData
+ * @param[in]  aAdvData         The AdvData that will be included in the Joiner Advertisement.
+ * @param[in]  aAdvDataLength   The length of AdvData in bytes.
  *
  * @retval OT_ERROR_NONE         Successfully set Joiner Advertisement.
  * @retval OT_ERROR_INVALID_ARGS Invalid AdvData.
--- a/src/core/meshcop/meshcop_tlvs.hpp
+++ b/src/core/meshcop/meshcop_tlvs.hpp
@@ -2365,7 +2365,8 @@ class JoinerAdvertisementTlv : public Tlv
 public:
     enum
     {
-        kType = kJoinerAdvertisement, ///< The TLV Type.
+        kType             = kJoinerAdvertisement,         ///< The TLV Type.
+        kAdvDataMaxLength = OT_JOINER_ADVDATA_MAX_LENGTH, ///< The Max Length of AdvData
     };
 
     /**
@@ -2436,20 +2437,15 @@ public:
      */
     void SetAdvData(const uint8_t *aAdvData, uint8_t aAdvDataLength)
     {
-        OT_ASSERT((aAdvData != nullptr) && (aAdvDataLength > 0) && (aAdvDataLength <= kMaxLength));
+        OT_ASSERT((aAdvData != nullptr) && (aAdvDataLength > 0) && (aAdvDataLength <= kAdvDataMaxLength));
 
         SetLength(aAdvDataLength + sizeof(mOui));
         memcpy(mAdvData, aAdvData, aAdvDataLength);
     }
 
 private:
-    enum
-    {
-        kMaxLength = 64,
-    };
-
     uint8_t mOui[3];
-    uint8_t mAdvData[kMaxLength];
+    uint8_t mAdvData[kAdvDataMaxLength];
 } OT_TOOL_PACKED_END;
 
 } // namespace MeshCoP
--- a/src/core/meshcop/meshcop_tlvs.hpp
+++ b/src/core/meshcop/meshcop_tlvs.hpp
@@ -2432,7 +2432,8 @@ public:
     /**
      * This method sets the Adv Data value.
      *
-     * @param[in]  aAdvData  A pointer to the Adv Data value.
+     * @param[in]  aAdvData        A pointer to the AdvData value.
+     * @param[in]  aAdvDataLength  The length of AdvData in bytes.
      *
      */
     void SetAdvData(const uint8_t *aAdvData, uint8_t aAdvDataLength)
--- a/src/core/meshcop/meshcop_tlvs.hpp
+++ b/src/core/meshcop/meshcop_tlvs.hpp
@@ -111,6 +111,7 @@ public:
         kEnergyList              = OT_MESHCOP_TLV_ENERGY_LIST,              ///< Energy List TLV
         kDiscoveryRequest        = OT_MESHCOP_TLV_DISCOVERYREQUEST,         ///< Discovery Request TLV
         kDiscoveryResponse       = OT_MESHCOP_TLV_DISCOVERYRESPONSE,        ///< Discovery Response TLV
+        kJoinerAdvertisement     = OT_MESHCOP_TLV_JOINERADVERTISEMENT,      ///< Joiner Advertisement TLV
     };
 
     /**
@@ -2352,6 +2353,103 @@ private:
     };
     uint8_t mFlags;
     uint8_t mReserved;
+} OT_TOOL_PACKED_END;
+
+/**
+ * This class implements Joiner Advertisement TLV generation and parsing.
+ *
+ */
+OT_TOOL_PACKED_BEGIN
+class JoinerAdvertisementTlv : public Tlv
+{
+public:
+    enum
+    {
+        kType = kJoinerAdvertisement, ///< The TLV Type.
+    };
+
+    /**
+     * This method initializes the TLV.
+     *
+     */
+    void Init(void)
+    {
+        SetType(kJoinerAdvertisement);
+        SetLength(sizeof(*this) - sizeof(Tlv));
+    }
+
+    /**
+     * This method indicates whether or not the TLV appears to be well-formed.
+     *
+     * @retval TRUE   If the TLV appears to be well-formed.
+     * @retval FALSE  If the TLV does not appear to be well-formed.
+     *
+     */
+    bool IsValid(void) const { return GetLength() >= sizeof(mOui) && GetLength() <= sizeof(mOui) + sizeof(mAdvData); }
+
+    /**
+     * This method returns the Vendor OUI value.
+     *
+     * @returns The Vendor OUI value.
+     *
+     */
+    uint32_t GetOui(void) const
+    {
+        return (static_cast<uint32_t>(mOui[0]) << 16) | (static_cast<uint32_t>(mOui[1]) << 8) |
+               static_cast<uint32_t>(mOui[2]);
+    }
+
+    /**
+     * This method returns the Vendor OUI value.
+     *
+     * @param[in]  aOui The Vendor OUI value.
+     *
+     */
+    void SetOui(uint32_t aOui)
+    {
+        mOui[0] = (aOui >> 16) & 0xff;
+        mOui[1] = (aOui >> 8) & 0xff;
+        mOui[2] = aOui & 0xff;
+    }
+
+    /**
+     * This method returns the Adv Data length.
+     *
+     * @returns The AdvData length.
+     *
+     */
+    uint8_t GetAdvDataLength(void) const { return GetLength() - sizeof(mOui); }
+
+    /**
+     * This method returns the Adv Data value.
+     *
+     * @returns A pointer to the Adv Data value.
+     *
+     */
+    const uint8_t *GetAdvData(void) const { return mAdvData; }
+
+    /**
+     * This method sets the Adv Data value.
+     *
+     * @param[in]  aAdvData  A pointer to the Adv Data value.
+     *
+     */
+    void SetAdvData(const uint8_t *aAdvData, uint8_t aAdvDataLength)
+    {
+        OT_ASSERT((aAdvData != nullptr) && (aAdvDataLength > 0) && (aAdvDataLength <= kMaxLength));
+
+        SetLength(aAdvDataLength + sizeof(mOui));
+        memcpy(mAdvData, aAdvData, aAdvDataLength);
+    }
+
+private:
+    enum
+    {
+        kMaxLength = 64,
+    };
+
+    uint8_t mOui[3];
+    uint8_t mAdvData[kMaxLength];
 } OT_TOOL_PACKED_END;
 
 } // namespace MeshCoP
--- a/src/core/meshcop/meshcop_tlvs.hpp
+++ b/src/core/meshcop/meshcop_tlvs.hpp
@@ -2401,7 +2401,7 @@ public:
     }
 
     /**
-     * This method returns the Vendor OUI value.
+     * This method sets the Vendor OUI value.
      *
      * @param[in]  aOui The Vendor OUI value.
      *
--- a/src/core/thread/discover_scanner.cpp
+++ b/src/core/thread/discover_scanner.cpp
@@ -69,7 +69,6 @@ otError DiscoverScanner::Discover(const Mac::ChannelMask &aScanChannels,
 {
     otError                         error   = OT_ERROR_NONE;
     Message *                       message = nullptr;
-    uint16_t                        startOffset;
     Tlv                             tlv;
     Ip6::Address                    destination;
     MeshCoP::DiscoveryRequestTlv    discoveryRequest;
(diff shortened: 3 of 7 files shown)


src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -2354,6 +2355,108 @@ class DiscoveryResponseTlv : public Tlv
...
+     *
+     * @returns The AdvData length.
+     *
+     */
+    uint8_t GetAdvDataLength(void) const { return mAdvDataLength; }
+
+    /**
+     * This method returns the Adv Data value.
+     *
+     * @returns The Adv Data value.
```
simonlingoogle, line 2426:
```suggestion
     * @returns A pointer to the Adv Data value.
```

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -2354,6 +2355,108 @@ class DiscoveryResponseTlv : public Tlv
...
+
+    /**
+     * This method sets the Adv Data value.
+     *
+     * @param[in]  aAdvData  A pointer to the Adv Data value.
+     *
+     */
+    void SetAdvData(const uint8_t *aAdvData, uint8_t aAdvDataLength)
+    {
+        uint8_t len = (((aAdvData == nullptr) || (aAdvDataLength > kMaxLength)) ? 0 : aAdvDataLength);
```
simonlingoogle, line 2439:
```suggestion
        OT_ASSERT((aAdvData == nullptr && aAdvDataLength == 0) || (aAdvData != nullptr && (aAdvDataLength > 0 && aAdvDataLength <= kMaxLength) ));
```
Be more strict on arguments?

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -2354,6 +2355,108 @@ class DiscoveryResponseTlv : public Tlv
...
+
+        if (len > 0)
+        {
+            memcpy(mAdvData, aAdvData, len);
+            mAdvDataLength = aAdvDataLength;
+            SetLength(len + sizeof(mOui));
+        }
```
simonlingoogle, line 2446:
```suggestion
        mAdvDataLength = len;
        SetLength(len + sizeof(mOui));
        memcpy(mAdvData, aAdvData, len);
```

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -2354,6 +2355,108 @@ class DiscoveryResponseTlv : public Tlv
...
+
+private:
+    enum
+    {
+        kMaxLength = 64,
+    };
+
+    uint8_t mOui[3];
+    uint8_t mAdvData[kMaxLength];
+    uint8_t mAdvDataLength;
```
simonlingoogle, line 2457: Not necessary to have this field since `mAdvDataLength` always equals to `GetLength() - sizeof(mOui)`?

src/core/thread/discover_scanner.hpp
```diff
@@ -163,6 +180,10 @@ class DiscoverScanner : public InstanceLocator
     uint8_t          mScanChannel;
     bool             mEnableFiltering : 1;
     bool             mShouldRestorePanId : 1;
+    uint32_t         mOui;
+    const uint8_t *  mAdvData;
+    uint8_t          mAdvDataLength;
```
simonlingoogle, line 185:
```suggestion
    JoinerAdvertisementTlv mJoinerAdvertisement;
```
Use a `JoinerAdvertisementTlv` instead?

src/core/thread/discover_scanner.cpp
```diff
@@ -144,6 +157,22 @@ otError DiscoverScanner::Discover(const Mac::ChannelMask &aScanChannels,
...
 
+otError DiscoverScanner::SetJoinerAdvertisement(uint32_t aOui, const uint8_t *aAdvData, uint8_t aAdvDataLength)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit((aAdvData != nullptr) && (aAdvDataLength != 0) && (aAdvDataLength <= kMaxLength),
+                 error = OT_ERROR_INVALID_ARGS);
+
+    mOui                    = aOui;
+    mAdvData                = aAdvData;
```
simonlingoogle, line 168: Should `memcpy` the data to `mAdvData` instead of just copying the pointer? 
`SetJoinerAdvertisement` is called by `otThreadSetJoinerAdvertisement` API. `aAdvData` might be a pointer to stack data.

src/core/thread/discover_scanner.cpp
```diff
@@ -144,6 +157,22 @@ otError DiscoverScanner::Discover(const Mac::ChannelMask &aScanChannels,
...
+    mAdvDataLength          = aAdvDataLength;
+    mHasJoinerAdvertisement = true;
```
simonlingoogle, line 170: Should we also have API for turning off `Joiner Advertisement`?

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -2354,6 +2355,108 @@ class DiscoveryResponseTlv : public Tlv
...
+
+    /**
+     * This method returns the Vendor OUI value.
+     *
+     * @param[in]  aOui The Vendor OUI value.
+     *
+     */
+    void SetOui(uint32_t aOui)
+    {
+        mOui[0] = (aOui >> 16) & 0xff;
```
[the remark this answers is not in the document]
lanyuwen: I think justing ignoring it should be fine.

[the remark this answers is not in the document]
simonlingoogle: Fine to me.

include/openthread/thread.h
```diff
@@ -264,6 +264,22 @@ otError otThreadDiscover(otInstance *             aInstance,
...
+ *
+ * @retval OT_ERROR_NONE         Successfully set Joiner Advertisement.
+ * @retval OT_ERROR_INVALID_ARGS Invalid AdvData.
+ *
+ */
+otError otThreadSetJoinerAdvertisement(otInstance *   aInstance,
+                                       uint32_t       aOui,
+                                       const uint8_t *aAdvData,
+                                       uint8_t        aAdvDataLength);
+
```
abtink, line 282: Looks like th current code assumes that the caller is provding the buffer for `aAdvData` (i.e. the content is not copied an just the pointer is saved).

There are pros/cons for this model (mainly that it makes it more complex for caller/user of OT), but I think it is good and flexible model for the initial implementation.


_Assuming we choose to go for this model (i.e., caller need to provides the buffer/object)_, I suggest we change it as follows:
- We define a new type `otJoinerAdvData` 
```
/**
 * This structure represents a Joiner Advertisement Data.
 *
 */
typedef struct otJoinerAdvData
{
    uint32_t                      mOui;        ///< Vendor OUI
    const uint8_t *               mData;       ///< Pointer to buffer containing the Adv Data
    uint8_t                       mDataLength; ///< The length of data (in bytes)
    const struct otJoinerAdvData *mNext;       ///< For internal OT core use.
} otJoinerAdvData;
```
- We define `otThreadAddJoinerAdvData(otInstance *, const otJoinAdvData *)` and `otThreadRemoveAdvData()` to allow user to add and remove entries (e.g., associated with different OUIs)
- We require the `otJoinerAdvData` instance that the caller provdies to remain valid.
- Within OT core just put the given objects into a linked-list.

Thoughts?

include/openthread/thread.h
```diff
@@ -264,6 +264,22 @@ otError otThreadDiscover(otInstance *             aInstance,
...
+ *
+ * @param[in]  aInstance A pointer to an OpenThread instance.
+ * @param[in]  aOui      The Vendor OUI value which will be included in the Joiner Advertisement.
+ * @param[in]  aAdvData  The AdvData which will be included in the Joiner Advertisement.
+ *
+ * @retval OT_ERROR_NONE         Successfully set Joiner Advertisement.
+ * @retval OT_ERROR_INVALID_ARGS Invalid AdvData.
+ *
+ */
+otError otThreadSetJoinerAdvertisement(otInstance *   aInstance,
```
abtink, line 278: Can we consider moving/defining the Joiner Adv related APIs in `joiner.h` (since they seem to be applicable to a joiner).

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -2354,6 +2355,108 @@ class DiscoveryResponseTlv : public Tlv
     uint8_t mReserved;
 } OT_TOOL_PACKED_END;
 
+/**
+ * This class implements Joiner Advertisement TLV generation and parsing.
+ *
+ */
+OT_TOOL_PACKED_BEGIN
+class JoinerAdvertisementTlv : public Tlv
```
abtink, line 2363: If we have more than one OUI Adv data, do we want to append multiple `JoinerAdvertisementTlv`s (each with its own OUI and data), or a single `JoinerAdvertisementTlv` with multiple entries in it?

I would suggest going for multiple `JoinerAdvertisementTlv`. 

Thoughts?

src/core/thread/discover_scanner.cpp
lanyuwen: Agree, fixed.

src/core/meshcop/meshcop_tlvs.hpp
lanyuwen: I'm not thinking to support either one in this initial implementation. Maybe we can add it later when there is a use case.

include/openthread/thread.h
lanyuwen: Thanks for the suggestion Abtin. I've changed the code to copy the content rather than just save the pointer as the latter approach could be problematic if the user fails to keep the data valid until the TLV is sent. Please take a look.

src/core/meshcop/meshcop_tlvs.hpp
abtink: The  main reason I bring it up is that the TLV format would impact this. And I don't think we want to change the TLV format later on.

If we want the multiple entries in same TLV, then TLV format needs to track the length of each data portion. If we go for multiple TLVs then TLV format can omit the length (derive it from TLV len itself as suggested by @simonlingoogle).

lanyuwen: Fair point Abtin. I actually prefer to go for multiple TLVs if we want to support the mentioned use case later (multiple OUIs/AdvData on a single deivce). Agree that we can omit the `mAdvDataLength` from the TLV implementation, I've updated the code to include this change. Please take a look.

bukepo: If we are going to support multiple data, maybe we can borrow the idea in BLE's advertisement. We don't use OUI, but service type. And users should request the service type in Thread group for new type of data. Thoughts?

include/openthread/thread.h
lanyuwen: The Joiner Advertisement is not related to `joiner.h` much actually, though they seem to be related. Technically we can move the APIs into `joiner.h`, but I was hesitating to move these APIs to `joiner.h` for a couple reasons:

- moving user facing APIs to `joiner.h` also implies that the internal APIs should better be moved from `class Mle::DiscoverScanner `to `class MeshCoP::Joiner`
- the `AdvData` however needs sent out within `DiscoverScanner` eventually where the discovery request message will be sent out

As results of above two facts:
- the parameters (such as `AdvData` and `Oui`)need passed from `Joiner` to `DiscoverScanner`, so we'll have to either change the existing `DiscoverScanner` APIs or allowing `DiscoverScanner` to access all of `Joiner`'s the private members
- the DiscoverScanner code will have to be fragmented because `OPENTHREAD_CONFIG_JOINER_ENABLE` flag will be introduced to make sure the new code only included when `MeshCoP::Joiner` code is included

And more importantly:
- doing so will disable the Joiner Advertisement function when `MeshCoP::Joiner` is not enabled, which is not always desired. E.g., the application protocol may want to use Joiner Advertisement for rendezvous purpose, but do not want to use the Thread Commissioning for provisioning.

Thoughts?

src/core/meshcop/meshcop_tlvs.hpp
lanyuwen: Hi Yakun, I think we had some discussions on "OUI" v.s. "Type" and have reached to an agreement on using OUI. The main reason is that "While OUI is 3 bytes, it alleviates the need for Thread Group to maintain a number authority.".

src/core/thread/discover_scanner.hpp
lanyuwen: this is intended to save memory.

simonlingoogle: Fine with both approach. 
But actually only saved 1 byte (i.e. Type of TLV) ?

src/core/thread/discover_scanner.hpp
```diff
@@ -163,6 +181,10 @@ class DiscoverScanner : public InstanceLocator
     uint8_t          mScanChannel;
     bool             mEnableFiltering : 1;
     bool             mShouldRestorePanId : 1;
+    uint32_t         mOui;
+    uint8_t          mAdvData[kMaxLength];
+    uint8_t          mAdvDataLength;
+    bool             mHasJoinerAdvertisement : 1;
```
simonlingoogle, line 187: nit: use better data alignment?

- Move bool flags together ?
- Move uint32_t mOui up ?

src/core/thread/discover_scanner.cpp
```diff
@@ -144,6 +157,24 @@ otError DiscoverScanner::Discover(const Mac::ChannelMask &aScanChannels,
...
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit((aAdvData != nullptr) && (aAdvDataLength != 0) && (aAdvDataLength <= kMaxLength),
+                 error = OT_ERROR_INVALID_ARGS);
+
+    mOui           = aOui;
+    mAdvDataLength = aAdvDataLength;
+
+    memcpy(mAdvData, aAdvData, aAdvDataLength);
```
simonlingoogle, line 170:
```suggestion
if (aAdvDataLength > 0)
{
    memcpy(mAdvData, aAdvData, aAdvDataLength);
}
```
Passing `nullptr` to memcpy is undefined?

lanyuwen: - maybe I miss something, but what's the intention to check if an `uint8_t` variable is greater than 0 or not?
- `nullptr` has been disallowed and won't be passed to `memcpy`, do you mean `nullptr` should be supported?

src/core/thread/discover_scanner.hpp
lanyuwen: thanks, tried my best to make them look tidy again. :)

src/core/thread/discover_scanner.cpp
simonlingoogle: Didn't notice that `nullptr` is prohibited.

lanyuwen:
@bukepo @librasungirl @jwhui any comments?

include/openthread/thread.h
abtink: SGTM

src/core/thread/discover_scanner.hpp
```diff
@@ -142,6 +155,11 @@ class DiscoverScanner : public InstanceLocator
         kStateScanDone,
     };
 
+    enum
+    {
+        kMaxLength = 64,
```
abtink, line 160: Suggest renaming this constant to be explicit that it is adv data max length.

abtink: We are adding two `enum` definitions for the same constant, in here and also in the TLV definition. 
Suggest defining it once in TLV as a public `enum` which then allows it to used by `DiscoverScanner` as well.

src/core/thread/discover_scanner.cpp
```diff
@@ -113,6 +116,16 @@ otError DiscoverScanner::Discover(const Mac::ChannelMask &aScanChannels,
 
     SuccessOrExit(error = Tlv::AppendTlv(*message, Tlv::kDiscovery, &discoveryRequest, sizeof(discoveryRequest)));
 
+    if (mHasJoinerAdvertisement)
+    {
+        // Append MLE Discovery TLV with sub-TLV (MeshCoP Joiner Advertisement).
```
abtink, line 121: This comment does not seem to be valid here. In the code below it, we are not appending Discovery TLV with a sub-TLV as its value and we are just  appending a `JoinerAdv` TLV to the message.

@lanyuwen did you intend for the `JoinerAdv` TLV to be added as one of the sub-TLVs of the `DiscoveryRequestTlv` or as its own TLV? If the former, then the code need to update the length of enclosing `DiscoveryTlv` (as it now has two sub-tlvs as its value).

include/openthread/thread.h
abtink: SGTM.

src/core/thread/discover_scanner.cpp
lanyuwen: The intention is to append the TLV as a sub-TLV. I'll get it fixed later, and thanks for pointing this out!

src/core/thread/discover_scanner.hpp
lanyuwen: > Suggest renaming this constant to be explicit that it is adv data max length.

Sure, I can rename it. But just wanted to keep the naming aligned with what's being used everywhere else in the `meshcop_tlvs.hpp`

> We are adding two `enum` definitions for the same constant, in here and also in the TLV definition.
> Suggest defining it once in TLV as a public `enum` which then allows it to used by `DiscoverScanner` as well.

I don't think we can easily share the enum to `discover_scanner.hpp` or other code piece without declaring an object of the TLV class even it's defined as a public member. Do you mean we should make something like `uint8_t mAdvData[MeshCoP::JoinerAdvertisementTlv::kAdvDataMaxLength]`) or am I missing anything?

abtink: > I don't think we can easily share the enum to `discover_scanner.hpp` or other code piece without declaring an object of the TLV class even it's defined as a public member. Do you mean we should make something like `uint8_t mAdvData[MeshCoP::JoinerAdvertisementTlv::kAdvDataMaxLength])` or am I missing anything?

Yes. We can use any public `enum` constants defined by another `class`/`struct` (key is that it should be public).
So yes, this definition would be valid in `DiscoverScanner`:
```
uint8_t mAdvData[MeshCoP::JoinerAdvertisementTlv::kAdvDataMaxLength])
```

The `enum` definition in this case is a nested type of `MeshCoP::JoinerAdvertisementTlv` class. And using nested types, in general, does not require us to have or declare an object of the enclosing type. The enclosing type is practically acting as a `namespace`.

lanyuwen: Great, I'll give a try. Thank you Abtin!

src/core/thread/discover_scanner.hpp
bukepo: can mAdvDataLength serve the purpose of mHasJoinerAdvertisement?

src/core/thread/discover_scanner.hpp
lanyuwen: fixed

src/core/thread/discover_scanner.cpp
lanyuwen: fixed

src/core/thread/discover_scanner.cpp
```diff
@@ -106,12 +110,30 @@ otError DiscoverScanner::Discover(const Mac::ChannelMask &aScanChannels,
...
+        // Append sub-TLV MeshCoP Joiner Advertisement.
+        joinerAdvertisement.Init();
+        joinerAdvertisement.SetOui(mOui);
+        joinerAdvertisement.SetAdvData(mAdvData, mAdvDataLength);
+
+        SuccessOrExit(error = joinerAdvertisement.AppendTo(*message));
+    }
+
+    tlv.SetLength(static_cast<uint8_t>(message->GetLength() - startOffset));
+    message->Write(startOffset - sizeof(tlv), sizeof(tlv), &tlv);
```
abtink, line 136: Thanks @lanyuwen. This code works fine. I have a suggestion to simplify it a bit.

Since we create the sub-TLV(s) of `DiscoveryTlv` and therefore their sizes is known beforehand, we can write the Discovery TLV with the correct length in the `message` from the beginning. No need to write it with an unspecified length and then remember the offset in the `message` to re-write the TLV and update its length later. 

```
    discoveryRequest.Init();
    discoveryRequest.SetVersion(kThreadVersion);
    discoveryRequest.SetJoiner(aJoiner);

    if (mAdvDataLength != 0)
    {
        joinerAdvertisement.Init();
        joinerAdvertisement.SetOui(mOui);
        joinerAdvertisement.SetAdvData(mAdvData, mAdvDataLength);
    }

    // Append Discovery TLV with one or two sub-TLVs.
    tlv.SetType(Tlv::kDiscovery);
    tlv.SetLength(static_cast<uint8_t>(discoveryRequest.GetSize() + 
                 ((mAdvDataLength != 0) ? joinerAdvertisement.GetSize() : 0)));

    SuccessOrExit(error = message->Append(&tlv, sizeof(tlv)));
    SuccessOrExit(error = discoveryRequest.AppendTo(*message));

    if (mAdvDataLength != 0)
    {
        SuccessOrExit(error = joinerAdvertisement.AppendTo(*message));
    }
```

lanyuwen: Great, this is indeed much more simplified. Thank you Abtin! :)

lanyuwen: Just updated as @abtink  suggested.

include/openthread/thread.h
```diff
@@ -264,6 +264,30 @@ otError otThreadDiscover(otInstance *             aInstance,
  */
 bool otThreadIsDiscoverInProgress(otInstance *aInstance);
 
+/**
+ * This method sets the Thread Joiner Advertisement when discovering Thread network.
+ *
+ * Thread Joiner Advertisement is used to allow a Joiner advertise its own application-specific information
+ * (such as Vendor ID, Product ID, Discriminator, etc.) unsolicitedly via a newly-proposed Joiner Advertisement TLV,
+ * and to make this information available to Commissioners or Commissioner Candidates without human interactions.
```
jwhui, line 272:
```suggestion
 * Thread Joiner Advertisement is used to allow a Joiner to advertise its own application-specific information
 * (such as Vendor ID, Product ID, Discriminator, etc.) via a newly-proposed Joiner Advertisement TLV,
 * and to make this information available to Commissioners or Commissioner Candidates without human interaction.
```

include/openthread/thread.h
```diff
@@ -264,6 +264,30 @@ otError otThreadDiscover(otInstance *             aInstance,
...
+ *
+ * @param[in]  aInstance        A pointer to an OpenThread instance.
+ * @param[in]  aOui             The Vendor IEEE OUI value which will be included in the Joiner Advertisement. Only the
```
jwhui, line 275:
```suggestion
 * @param[in]  aOui             The Vendor IEEE OUI value that will be included in the Joiner Advertisement. Only the
```

include/openthread/thread.h
```diff
@@ -264,6 +264,30 @@ otError otThreadDiscover(otInstance *             aInstance,
...
+ *                              least significant 3 bytes will be used, and the most significant byte will be ignored.
+ * @param[in]  aAdvData         The AdvData which will be included in the Joiner Advertisement.
```
jwhui, line 277:
```suggestion
 * @param[in]  aAdvData         The AdvData that will be included in the Joiner Advertisement.
```

include/openthread/thread.h
```diff
@@ -264,6 +264,30 @@ otError otThreadDiscover(otInstance *             aInstance,
...
+ * @param[in]  aAdvDataLength   The length of AdvData
```
jwhui, line 278:
```suggestion
 * @param[in]  aAdvDataLength   The length of AdvData in bytes.
```

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -2354,6 +2355,99 @@ class DiscoveryResponseTlv : public Tlv
...
+
+    /**
+     * This method returns the Vendor OUI value.
+     *
+     * @returns The Vendor OUI value.
+     *
+     */
+    uint32_t GetOui(void) const
+    {
+        return (static_cast<uint32_t>(mOui[0]) << 16) | (static_cast<uint32_t>(mOui[1]) << 8) |
```
jwhui, line 2399: Should we add `ReadUint24()` and `WriteUint24()` to [src/core/common/encoding.hpp](https://github.com/openthread/openthread/blob/master/src/core/common/encoding.hpp)? I'm fine leaving this for a separate PR.

src/core/thread/discover_scanner.cpp
```diff
@@ -144,6 +166,23 @@ otError DiscoverScanner::Discover(const Mac::ChannelMask &aScanChannels,
     return error;
 }
 
+otError DiscoverScanner::SetJoinerAdvertisement(uint32_t aOui, const uint8_t *aAdvData, uint8_t aAdvDataLength)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit((aAdvData != nullptr) && (aAdvDataLength != 0) &&
```
jwhui, line 173: Should we allow `aAdvLength == 0` to allow the user to clear the advertisement data?

src/core/meshcop/meshcop_tlvs.hpp
lanyuwen: Agree, let me file an issue to follow up on this later.

src/core/thread/discover_scanner.cpp
lanyuwen: I'm a bit hesitated to allow the user to clear the advertisement data as this may bring in some additional implications to border agent and commissioners. Maybe we can introduce an enhancement later when we have better understanding on the use cases?

src/core/meshcop/meshcop_tlvs.hpp
lanyuwen: FYI - #5322

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -2354,6 +2355,99 @@ class DiscoveryResponseTlv : public Tlv
...
+               static_cast<uint32_t>(mOui[2]);
+    }
+
+    /**
+     * This method returns the Vendor OUI value.
```
librasungirl, line 2404:
```suggestion
     * This method sets the Vendor OUI value.
```

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -2354,6 +2355,99 @@ class DiscoveryResponseTlv : public Tlv
...
+     *
+     * @returns A pointer to the Adv Data value.
+     *
+     */
+    const uint8_t *GetAdvData(void) const { return mAdvData; }
+
+    /**
+     * This method sets the Adv Data value.
+     *
+     * @param[in]  aAdvData  A pointer to the Adv Data value.
```
librasungirl, line 2435: add doxygen:  
` * @param[in]  aAdvDataLength  The length of Adv Data @p aAdvData.`

src/core/thread/discover_scanner.hpp
```diff
@@ -134,6 +134,19 @@ class DiscoverScanner : public InstanceLocator
      */
     bool IsInProgress(void) const { return (mState != kStateIdle); }
 
+    /**
+     * This method sets Joiner Advertisement.
+     *
+     * @param[in]  aOui             The Vendor OUI for Joiner Advertisement
+     * @param[in]  aAdvData         A pointer to AdvData for Joiner Advertisement
+     * @param[in]  aAdvDataLength   The length of AdvData
```
librasungirl, line 142: nits
```suggestion
     * @param[in]  aOui             The Vendor OUI for Joiner Advertisement.
     * @param[in]  aAdvData         A pointer to AdvData for Joiner Advertisement.
     * @param[in]  aAdvDataLength   The length of AdvData @p aAdvData.
```

src/core/meshcop/meshcop_tlvs.hpp
lanyuwen: updated, thanks. 👍

src/core/meshcop/meshcop_tlvs.hpp
lanyuwen: updated, thanks. 👍

src/core/thread/discover_scanner.hpp
lanyuwen: updated, thanks. 👍

PR #5301, opened 2020-07-28
[mle] fix FTD unable to restore link to parent

### LINKED ISSUE #5293: [mle] re-synchronization with FTD child
Hello,

I'm currently on [this](https://github.com/openthread/openthread/commit/e653478c503d5b13207b01938fa1fa494a8b87d3) git commit using nRF52840.

I noticed, that sometimes FTD child after reset can't re-attach to the old parent quite fast. The reason for this is that Child drops several `MLE Child Update Response` from the parent, and finally after few retries it starts regular attachment using `MLE Parent Request`.

```
-MLE-----: Failed to process Child Update Response: Drop
```

The reason for this is that `HandleChildUpdateResponse` method calls `HandleLeaderData` which returns `OT_ERROR_DROP`.
This is because `mLeaderData` (local Leader Data) is empty (i.e. partitionId = 0, weighting=0, leaderId = 0 etc.), and so it is different from the Parent's Leader Data.

The problem does not occur when `MLE Advertisement` message is received somewhere in the re-attachment process - since `mLeaderData` is updated.

@jwhui Is this expected behaviour? If not, would it make sense to accept new Leader Data from the parent?

simonlingoogle:
This PR fixes #5293 

Testing if two `OT_ASSERT(leader != nullptr);` statements can be kept since they seems correct:
- OT Build: 
  - **FTD Leader Data dropping removed**
  - **`OT_ASSERT(leader != nullptr);` statements kept**
  - kMinDowngradeNeighbors = 4
  - OPENTHREAD_CONFIG_MLE_LONG_ROUTES_ENABLE = 1
- OTNS Stress Tests
  - [x] [Large Network Forming](https://github.com/openthread/ot-ns/blob/master/pylibs/stress_tests/large_network_forming.py) (simulated 100 days successfully)
  - [x] [MLEID Connectivity test](https://github.com/simonlingoogle/ot-ns/blob/stress-test-mleid-connectivity/pylibs/stress_tests/mleid_connectivity.py) (simulated 100 days successfully)

### DIFF
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -2839,12 +2839,6 @@ otError Mle::HandleLeaderData(const Message &aMessage, const Ip6::MessageInfo &a
     {
         if (IsChild())
         {
-#if OPENTHREAD_FTD
-            // An FTD skips handling LeaderData of a different partition.
-            VerifyOrExit(!IsFullThreadDevice() || (leaderData.GetPartitionId() == mLeaderData.GetPartitionId() &&
-                                                   leaderData.GetLeaderRouterId() == GetLeaderId()),
-                         error = OT_ERROR_DROP);
-#endif
             SetLeaderData(leaderData.GetPartitionId(), leaderData.GetWeighting(), leaderData.GetLeaderRouterId());
             mRetrieveNewNetworkData = true;
         }


src/core/thread/mle.cpp
```diff
@@ -2918,12 +2918,6 @@ otError Mle::HandleLeaderData(const Message &aMessage, const Ip6::MessageInfo &a
     {
         if (IsChild())
         {
-#if OPENTHREAD_FTD
-            // An FTD skips handling LeaderData of a different partition.
-            VerifyOrExit(!IsFullThreadDevice() || (leaderData.GetPartitionId() == mLeaderData.GetPartitionId() &&
-                                                   leaderData.GetLeaderRouterId() == GetLeaderId()),
-                         error = OT_ERROR_DROP);
-#endif
```
abtink, line 2926: We know that `HandeLeaderData()` is called from 3 places:
- `Mle::HandleDataResponse()`
- `Mle::HandleChildUpdateRequest()`
- `Mle::HandleChildUpdateResponse`

This PR is aimed to fix the initial child sync (i.e. Child Update msgs). Just want us to make sure and verify that in other uses of the method, it is still safe/fine to  remove the partition and id check?

simonlingoogle: It's very thoughtful of you. 

Note that this code removed by this PR is added by https://github.com/openthread/openthread/pull/4860 to fix assertion failures when `mRouterTable.GetLeader() == nullptr`. So, this PR restores OT back to the original behavior that OT had  always been so until https://github.com/openthread/openthread/pull/4860, which I believe is safe. 

Thoughts?

abtink: Thanks for the info/background on this. Sounds all good to me. 👍

librasungirl, line 2926: @simonlingoogle ,  

When recalling #4860 and the [assertion issue](https://github.com/openthread/openthread/issues/4508#issuecomment-615899675) it tried to fix, I become confused on how the assertion happened after checking OT code a bit more:

Without this, if partition id is different

https://github.com/openthread/openthread/blob/82e7e32bb5526ffa3b0cca91ff7d37636417fa75/src/core/thread/mle.cpp#L951-L957

And the RouterTable will be cleared here
https://github.com/openthread/openthread/blob/82e7e32bb5526ffa3b0cca91ff7d37636417fa75/src/core/thread/mle_router.cpp#L101-L109

If it is cleared, in `MleRouter::HandleParentRequest()`, before going to the assertion, it will exit already at Line 1628.

https://github.com/openthread/openthread/blob/82e7e32bb5526ffa3b0cca91ff7d37636417fa75/src/core/thread/mle_router.cpp#L1628-L1632

If it is not the root cause for assertion failure you found, besides removing this block, I would prefer keeping `OT_ASSERT(leader != nullptr);` which could help detecting issues earlier.

Thoughts?

simonlingoogle: I agree with you that if FTD switched to another partition with only `LeaderData` but without `RouteTLV`, the condition `mRouterTable.GetLeaderAge() < mNetworkIdTimeout` should be false since the router table will be empty until a `RouteTLV` is processed. In such case the assertion should not fail. 

However, consider such case:
* The FTD falls behind in `Router ID Sequence` because it failed to receive some `MLE Advertisements`
* A new Router joined the network, so FTD does not the Router ID in Router Table
* The new Router become Leader (the FTD still don't know)
* FTD receives an Data Response with new Leader ID, but its Router Table is still outdated. 
* FTD updates the Leader Data with new Leader ID

Note that now the FTD would have `mRouterTable.GetLeader() == nullptr` and `mRouterTable.GetLeaderAge() < mNetworkIdTimeout` both satisfied. 

With such case in mind, I think that we would have to recognize `mRouterTable.GetLeader() == nullptr` as a rare corner case which still could happen. 

Thoughts? @librasungirl

librasungirl: >FTD receives an Data Response with new Leader ID, but its Router Table is still outdated, without new Leader Id.
>FTD updates the Leader Data with new Leader ID

When the FTD decides to accept the DataResponse, it only accept it in mLeaderData and it will clear its own Router Table as code mentioned in previous comment, right?

simonlingoogle: > > FTD receives an Data Response with new Leader ID, but its Router Table is still outdated, without new Leader Id.
> > FTD updates the Leader Data with new Leader ID
> 
> When the FTD decides to accept the DataResponse, it only accept it in mLeaderData and it will clear its own Router Table as code mentioned in previous comment, right?
> How could it keep the old router table when partition changes?

You are right. If a FTD has  a non-empty Router Table, it must have `mRouterTable.GetLeader() != nullptr`. 
So, there should be no assertion failure in  `HandleParentRequest`. 

There is also another assertion `OT_ASSERT(leader != nullptr)` in `HandleAddressSolicitResponse`, but this should also never  fail because  the Router Table is updated with a RouterMaskTLV which must have Leader in it.  

Back then when we were resolving #4860 , there are not just one bug but a number of bugs. For example, we found 
a bug that Router in Router Table could be removed unexpectedly. This might be the real cause of the `Leader == nullptr` issue. 

I will do some more work to try to reproduce this assertion failure. If can't I would prefer keeping these `assert` too. 

Thoughts ? @jwhui @librasungirl

librasungirl: SGTM @simonlingoogle

jwhui: SGTM as well. Thanks.

simonlingoogle: @librasungirl @jwhui 

The two stress tests finished successfully, each has simulated a network of significant size without any issue for 100 days despite fault injections. 

  - [x] [Large Network Forming](https://github.com/openthread/ot-ns/blob/master/pylibs/stress_tests/large_network_forming.py) (simulated 100 days successfully)
  - [x] [MLEID Connectivity test](https://github.com/simonlingoogle/ot-ns/blob/stress-test-mleid-connectivity/pylibs/stress_tests/mleid_connectivity.py) (simulated 100 days successfully)

I think this does give us more hint that these two asserts are valid and should not be removed.

Thus I have reverted the removal of the two asserts and now this PR only eliminates the dropping of LeaderData on FTD.

But we should be aware that `Leader` still might be `nullptr` before a valid `RouteTLV` or `RouteMaskTLV` is processed.

PR #5302, opened 2020-07-28
[bit-vector] refactor methods

simonlingoogle:
This PR refactors BitVector to use one method for setting or clearing masks.

### DIFF
--- a/src/core/common/bit_vector.hpp
+++ b/src/core/common/bit_vector.hpp
@@ -59,7 +59,7 @@ namespace ot {
  * @tparam N  Specifies the number of bits.
  *
  */
-template <uint16_t N> class BitVector : public Equatable<BitVector<N>>
+template <uint16_t N> class BitVector : public Equatable<BitVector<N>>, public Clearable<BitVector<N>>
 {
 public:
     /**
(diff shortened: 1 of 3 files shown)


simonlingoogle:
@abtink @jwhui Do you think this is a better implementation than original?

src/core/common/bit_vector.hpp
```diff
@@ -81,34 +81,23 @@ template <uint16_t N> class BitVector : public Equatable<BitVector<N>>
      * This method sets the mask of a given index.
      *
      * @param[in] aIndex  The index.
+     * @param[in] aSet    TRUE to set the mask, or FALSE to clear the mask.
      *
      */
-    void Set(uint16_t aIndex)
+    void Set(uint16_t aIndex, bool aSet)
```
jwhui, line 87: nit: how about `aValue`?
```suggestion
    void Set(uint16_t aIndex, bool aValue)
```

simonlingoogle:
> LGTM. Thanks. 👍
> 
> One suggestion about title of commit. How about `[bit-vector] refactor methods`? (since we already have `bit-vector` as a module in the code we can refer to it).

Thanks for the suggestion.

PR #5305, opened 2020-07-28
[posix] fix netif use of IPv6 Prefix Match method

abtink:


src/posix/platform/netif.cpp
```diff
@@ -292,7 +292,8 @@ static void InitNetaskWithPrefixLength(struct in6_addr *address, uint8_t prefixL
 
 static uint8_t NetmaskToPrefixLength(const struct sockaddr_in6 *netmask)
 {
-    return ot::Ip6::Address::PrefixMatch(netmask->sin6_addr.s6_addr, allOnes, 128);
+    return otIp6PrefixMatch(static_cast<const otIp6Address *>(netmask->sin6_addr.s6_addr),
+                            static_cast<const otIp6Address *>(allOnes));
```
abtink, line 296: The last parameter in `Ip6::Address::PrefixMatch()` is the size of buffers contaning the addresses/prefixes in bytes (not in bits).
It's not a critical issue as there is code in `PrefixMtach()` that checks the value  for size parameter and limits it to 16 (max size of an IPv6 address). But we may change this check to an assert soon.

Changed the code to use the public ot API instead here.

PR #5306, opened 2020-07-29
[ip6] adding Ip6::Prefix class

abtink:
This commit adds `Ip6::Prefix` class as sub-class of `otIp6Prefix`
representing an IPv6 Prefix. Helper methods are provided in `Prefix`
and `Address` to perform prefix matching.

Different OT core modules such as `network-data`, `lowpan`, `dhcp6`,
`slaac`, and `backbone-router` are updated to use the new `Prefix` and
its helper methods.

This commit also updates the unit test `test_ip6_address` to verify
behavior of `Prefix` and its helper methods.


src/core/thread/network_data_tlvs.hpp
```diff
@@ -479,6 +491,42 @@ class PrefixTlv : public NetworkDataTlv
...
+     * This method indicates whether the Prefix from TLV matches (same length and same prefix bytes) a given
+     * `Ip6::Prefix`.
+     *
+     * @param[in] aPrefix  A Prefix to match with.
+     *
+     * @retval TRUE   The TLV's Prefix matches @p aPrefix.
+     * @retval FALSE  The TLV's Prefix does not match @p aPrefix.
+     *
+     */
+    bool MatchesPrefix(Ip6::Prefix &aPrefix) const { return aPrefix.Matches(GetPrefix(), GetPrefixLength()); }
```
jwhui, line 512: Given length has to match, how about something like `IsEqual()`?
```suggestion
    bool IsEqual(Ip6::Prefix &aPrefix) const { return aPrefix.Matches(GetPrefix(), GetPrefixLength()); }
```

src/core/thread/network_data_tlvs.hpp
```diff
@@ -479,6 +491,42 @@ class PrefixTlv : public NetworkDataTlv
...
+     * This method indicates whether the Prefix from TLV matches (same length and same prefix bytes) a given Prefix.
+     *
+     * @param[in]  aPrefix        A pointer to an IPv6 prefix.
+     * @param[in]  aPrefixLength  The prefix length pointed to by @p aPrefix (in bits).
+     *
+     * @retval TRUE   The TLV's Prefix matches @p aPrefix.
+     * @retval FALSE  The TLV's Prefix does not match @p aPrefix.
+     *
+     */
+    bool MatchesPrefix(const uint8_t *aPrefix, uint8_t aPrefixLength) const
```
jwhui, line 524: Given length has to match, how about something like `IsEqual()`?
```suggestion
    bool IsEqual(const uint8_t *aPrefix, uint8_t aPrefixLength) const
```

src/core/thread/network_data.hpp
```diff
@@ -486,13 +525,36 @@ class NetworkData : public InstanceLocator
...
      *
      */
     const PrefixTlv *FindPrefix(const uint8_t *aPrefix, uint8_t aPrefixLength) const;
 
+    /**
+     * This method returns a pointer to a Prefix TLV.
+     *
+     * @param[in]  aPrefix        An IPv6 prefix.
+     *
+     * @returns A pointer to the Prefix TLV is one is found or nullptr if no matching Prefix TLV exists.
```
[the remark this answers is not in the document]
abtink: Thanks. There are many more instances same typo. Fixed them all.

src/core/net/ip6_address.hpp
```diff
@@ -82,6 +82,185 @@ class NetworkPrefix : public otIp6NetworkPrefix, public Equatable<NetworkPrefix>
...
+     * This method set the prefix from a given Network Prefix
+     *
+     */
+    void Set(const NetworkPrefix &aNetworkPrefix) { Set(aNetworkPrefix.m8, NetworkPrefix::kLength); }
+
+    /**
+     * This method clears the prefix by setting the length to zero.
+     *
+     */
+    void Clear(void) { mLength = 0; }
```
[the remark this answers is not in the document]
abtink: Good point. Would actually then add `SetLength(aLength)` and let the user of `Prefix` choose to set it to zero.

src/core/thread/network_data.cpp
```diff
@@ -686,18 +681,6 @@ const PrefixTlv *NetworkData::FindPrefix(const uint8_t *aPrefix,
...
-{
-    uint8_t matchedLength;
-
-    // Note that he `Ip6::Address::PrefixMatch` expects the prefix
-    // length to be in bytes unit.
-
-    matchedLength = Ip6::Address::PrefixMatch(a, b, BitVectorBytes(aLength));
-
-    return (matchedLength >= aLength) ? static_cast<int8_t>(matchedLength) : -1;
-}
```
abtink, line 699: Thanks @jwhui  for reviewing this. I'd like to ask for your help checking something.

In this commit/PR, I removed the `NetworkData` version of ` PrefixMatch` and used the new helper methods where it was being used before.

The `NetworkData::PrefixMatch()` calculates the `matchedLength` and if larger than given `aLength` it would return the `matchedLength` which can potentially be larger than `aLength`.

There is one place that removing this may change the existing behavior. 

- In `LeaderBase::ExternalRouteLookup()` we go through all the `PrefixTlv` from network data and try to match them with a given `aDestination` address and keep track of the best match so far.
- I changed the code in this PR to use `Address::MatchPrefix()` and then use the prefix length itself as the match length.
- This would be different from `NetworkData::PrefixMatch` which can potentially use a longer match length than  prefix length. 
- I think the new model is better and correct but just want to double-check with you as well.  

Thanks. :)

src/core/net/ip6_address.hpp
```diff
@@ -82,6 +82,188 @@ class NetworkPrefix : public otIp6NetworkPrefix, public Equatable<NetworkPrefix>
...
+     *
+     * @param[in] aPrefix  A pointer to buffer containing the prefix bytes.
+     * @param[in] aLength  The length or prefix in bits.
+     *
+     */
+    void Set(const uint8_t *aPrefix, uint8_t aLength);
+
+    /**
+     * This method set the prefix from a given Network Prefix
+     *
```
jwhui, line 145: Missing doc for `aNetworkPrefix`

src/core/thread/network_data.cpp
jwhui: I agree. Thanks for catching this!

PR #5307, opened 2020-07-29
[cmake] fix linking coverage library

Irving-cl:
`--coverage` looks more like a option than a library. Though somehow it works. I think `gcov` is better.


Irving-cl:
It seems this doesn't work.

PR #5309, opened 2020-07-29
[cmake] fix link issues of unit tests

Irving-cl:
I think `openthread-ncp-ftd` is added to unit tests' link libraries by mistake. It's not required and as it depends on `OT_PLATFORM`, it would cause `openthread-simulation` linked in unit test executables, which would lead to a redefinition as some methods are defined in `test-platform.cpp`.

### DIFF
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -33,7 +33,6 @@ set(COMMON_SOURCES
 
 set(COMMON_LIBS
     openthread-ftd
-    openthread-ncp-ftd
     ot-config
     ${OT_MBEDTLS}
     util


tests/unit/CMakeLists.txt
```diff
@@ -33,7 +33,6 @@ set(COMMON_SOURCES
 
 set(COMMON_LIBS
     openthread-ftd
-    openthread-ncp-ftd
```
simonlingoogle, line 36: why does `openthread-ncp-ftd` include simulation platform?

Irving-cl: I think so. I haven't checked the code. But I checked the `build.ninja` and I'm sure it's this one that causes `simulation` linked together.

Irving-cl: https://github.com/openthread/openthread/blob/master/src/ncp/CMakeLists.txt#L112

`openthread-ncp-ftd` added `OT_PLATFORM_LIB` into its link targets.

PR #5310, opened 2020-07-29
[posix] increase UART tx buffer size for posix

d1ee:
This fixes the issue for `ot-br-posix` when running command `ot-ctl` with large outputs over 1024 bytes.
In addition, I fixed the macro name in the comment.

### DIFF
--- a/src/posix/platform/openthread-core-posix-config.h
+++ b/src/posix/platform/openthread-core-posix-config.h
@@ -118,4 +118,14 @@
 
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_CLI_TX_BUFFER_SIZE
+ *
+ * The size of CLI message buffer in bytes.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_CLI_UART_TX_BUFFER_SIZE
+#define OPENTHREAD_CONFIG_CLI_UART_TX_BUFFER_SIZE 8192
+#endif
+
 #endif // OPENTHREAD_CORE_POSIX_CONFIG_H_
--- a/src/posix/platform/openthread-core-posix-config.h
+++ b/src/posix/platform/openthread-core-posix-config.h
@@ -119,7 +119,7 @@
 #endif
 
 /**
- * @def OPENTHREAD_CONFIG_CLI_TX_BUFFER_SIZE
+ * @def OPENTHREAD_CONFIG_CLI_UART_TX_BUFFER_SIZE
  *
  * The size of CLI message buffer in bytes.
  *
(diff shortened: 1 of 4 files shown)


d1ee:
@googlebot I signed it!

src/posix/platform/openthread-core-posix-config.h
```diff
@@ -118,4 +118,14 @@
...
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_CLI_UART_TX_BUFFER_SIZE
+ *
+ * The size of CLI message buffer in bytes.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_CLI_UART_TX_BUFFER_SIZE
+#define OPENTHREAD_CONFIG_CLI_UART_TX_BUFFER_SIZE 8192
```
simonlingoogle, line 128: do we really need such large TX buffer?

jwhui: On a POSIX environment, is there a practical downside of an 8 kB buffer?

1024 bytes was not enough even for modest networks. Some of the COI output depends on how many children you have attached.

bukepo: Can we utilize the flush api when buffer is full?

jwhui: We would need to implement `otPlatUartFlush()` to support that:
https://github.com/openthread/openthread/blob/b49ee0879ac1c264d947f2dc8c3bbc276f58f868/src/posix/platform/uart.cpp#L165-L168

simonlingoogle: > me of the COI output depends on how many children you have attached.

Yes, agree that 8KB is fine for POSIX.

bukepo: I tend to use less memory, openwrt is also posix, which has very limited resource.

jwhui: > I tend to use less memory, openwrt is also posix, which has very limited resource.

OpenWrt platforms have MBs of memory, right? Several kB is pretty small in comparison.

I don't think even 2kB is enough to print a full router table. At 8kB, the current child table CLI output is limited to somewhere around 100 children. We have already seen deployments with routers that support more than 100 children.

d1ee:
Fixing the problem would be a better option than increasing the buffer size. Unfortunately, I am not a C/C++ developer. I could try to implement `otPlatUartFlush()` and try to use it properly.

Does it sound ok for you? If yes, I'll close this pull-request and start implementing it.

bukepo:
> Fixing the problem would be a better option than increasing the buffer size. Unfortunately, I am not a C/C++ developer. I could try to implement `otPlatUartFlush()` and try to use it properly.
> 
> Does it sound ok for you? If yes, I'll close this pull-request and start implementing it.

Hi I submitted a PR adding the missing implementation: https://github.com/openthread/openthread/pull/5312

PR #5317, opened 2020-08-01
[commissioner] add 'state' CLI command

jwhui:

PR #5321, opened 2020-08-01
[tests] fix external commissioner building options

wgtdkp:
This PR aligns the external commissioner buliding options with OT commissioner.

Fixes test failures revealed in https://github.com/openthread/openthread/pull/5317/checks?check_run_id=934695903.


wgtdkp:
A long term fix it to have some building scripts with OT commissioner (e.g. `script/build-for-tests`, `script/build-for-android`, `script/build-for-java-binding` ...)

So that we don't need to maintain those options across repos. Will do it soon.

PR #5326, opened 2020-08-01
[mle-router] simplify 'GetNeighbor(const Ip6::Address &)'

abtink:
This commit simplifies `MleRouter:GetNeighbor(const Ip6::Address &)`
handling of mesh-local router locator IPv6 addresses.

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -3512,8 +3512,7 @@ Neighbor *MleRouter::GetNeighbor(const Mac::Address &aAddress)
 
 Neighbor *MleRouter::GetNeighbor(const Ip6::Address &aAddress)
 {
-    Lowpan::Context context;
-    Neighbor *      rval = nullptr;
+    Neighbor *rval = nullptr;
 
     if (aAddress.IsLinkLocal())
     {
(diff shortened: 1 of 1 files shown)


src/core/thread/mle_router.cpp
```diff
@@ -3523,32 +3522,25 @@ Neighbor *MleRouter::GetNeighbor(const Ip6::Address &aAddress)
...
     {
-        context.mContextId = 0xff;
+        uint16_t rloc16 = aAddress.GetIid().GetLocator();
+
+        rval = mChildTable.FindChild(rloc16, Child::kInStateValidOrRestoring);
+        VerifyOrExit(rval == nullptr, OT_NOOP);
+
+        rval = mRouterTable.GetNeighbor(rloc16);
+        ExitNow();
     }
```
abtink, line 3534: Seems like the `Context` was only used to determine if the address is mesh-local. The method `Mle::IsRoutingLocator()` would perform similar check to verify that the prefix to match the mesh-local prefix and the IID is locator pattern.

------

There is one corner case that I can see where the new code changes the behavior:
If we happen to have a prefix with a length strictly larger than 64 bit and its first 64 bits happen to match the mesh-local prefix and rest of bits happen to match the locator IID pattern, when calling `GetContext` we can get a different `Conext` matching `aAddress` than the mesh-local prefix one (since `GetContext()` tries to find the longest match length). So with existing code we won't treat the address as mesh-local (though it very much looks like it) but with new code (in this PR) we would. I don't know if this situation is realistic/valid and/or we need to handle.

@jwhui any thoughts?

abtink: Actually digging more, realized the corner-case situation I described above, cannot happen and would not be allowed.

When adding new prefixes, we do already check in `NetworkData` to ensure  that the new prefix against mesh-local prefix (i.e. it should not contain mesh-local prefix as a sub-prefix).
https://github.com/openthread/openthread/blob/3b6eb7de4282829bf8a74d1e4ab1687387505581/src/core/thread/network_data_local.cpp#L145

PR #5338, opened 2020-08-05
[coap] reduce OT_COAP_MAX_RETRANSMIT to 20

### LINKED ISSUE #5336: Overflow leads to invalid error
https://github.com/openthread/openthread/blob/481f69703b6478d461590ac6b036c7c7d5b6a6b8/src/core/coap/coap.cpp#L974

For number `mMaxRetransmit >= 22` and for any number `mAckTimeout >= 1000`, the multiplication leads to overflow and returns a `false` value for this method.

It wasn't trivial to find out why `IsValid()` method was returning `false` value. Maybe it would help to increase the `tmp`'s size to `uint64_t`.

jwhui:
Also add a note to clarify the set of valid CoAP transmission
parameters.

Resolves #5336

### DIFF
--- a/include/openthread/coap.h
+++ b/include/openthread/coap.h
@@ -60,7 +60,7 @@ extern "C" {
 
 #define OT_COAP_MAX_TOKEN_LENGTH 8 ///< Max token length as specified (RFC 7252).
 
-#define OT_COAP_MAX_RETRANSMIT 30 ///< Max retransmit supported by OpenThread.
+#define OT_COAP_MAX_RETRANSMIT 20 ///< Max retransmit supported by OpenThread.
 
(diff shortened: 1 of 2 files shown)


include/openthread/coap.h
```diff
@@ -368,6 +368,10 @@ typedef struct otCoapResource
 /**
  * This structure represents the CoAP transmission parameters.
  *
+ * @note mAckTimeout * ((2 ** (mMaxRetransmit + 1)) - 1) * (mAckRandomFactorNumerator / mAckRandomFactorDenominator)
+ *       must not exceed what can be represented by a uint32_t (0xffffffff). This limitation allows OpenThread to
```
simonlingoogle, line 372: Add a static_assert for it?

d1ee:
Also looks good to me, thanks 👍

include/openthread/coap.h
jwhui: These values are specified at run time.

PR #4501, opened 2020-01-23
[style] apply google python style guide

bukepo:
This PR applies and forces Google's python style for tests.

PR #5339, opened 2020-08-05
[style] change python yapf column_limit to 119

simonlingoogle:
Before #4501 changed python formatting tool to `yapf`, we agreed on using a column limit of 119. 

https://github.com/openthread/openthread/blob/3a25f3a020cf7b0925bfe717a16f3289f6ff7d5f/script/pystyle.cfg#L7-L8

In my opinion, a column limit of 80 is neither readable nor necessary.

Thoughts? @bukepo @librasungirl @jwhui @FiveDimensions


librasungirl:
@simonlingoogle IIRC, it was agreed to relax the length check (the pystyle.cfg is just used by flake8 to check the style, but not as a format style) considering the length of some method names.

Indeed talked with @bukepo long time ago that it might be better to format with column length 119. Thanks for making it happen.

PR #5342, opened 2020-08-05
[mac-frame] fix/enhance parsing of Header IEs

abtink:
This commit changes `MacFrame::FindPayloadIndex()`:

- Ensure that at least one Header IE is included when FCF indicates
  that Header IE is present.

- Detect error case where Header IE(s) are partially included.

- Handle the case where frame contains Header IE(s) but no data
  payload (spec does not require Header IE termination in this
  case).

### DIFF
--- a/src/core/mac/mac_frame.cpp
+++ b/src/core/mac/mac_frame.cpp
@@ -837,22 +837,24 @@ exit:
 uint8_t Frame::FindPayloadIndex(void) const
 {
     uint8_t index = SkipSecurityHeaderIndex();
-#if OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT
-    const uint8_t *cur    = nullptr;
-    const uint8_t *footer = GetFooter();
-#endif
 
     VerifyOrExit(index != kInvalidIndex, OT_NOOP);
 
 #if OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT
-    cur = GetPsdu() + index;
-
     if (IsIePresent())
     {
-        while (cur + sizeof(HeaderIe) <= footer)
+        const uint8_t *cur    = &mPsdu[index];
+        const uint8_t *footer = GetFooter();
+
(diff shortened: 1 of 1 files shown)


src/core/mac/mac_frame.cpp
```diff
@@ -837,22 +837,24 @@ uint8_t Frame::CalculateAddrFieldSize(uint16_t aFcf)
...
+        const uint8_t *footer = GetFooter();
+
+        do
         {
-            const HeaderIe *ie  = reinterpret_cast<const HeaderIe *>(cur);
-            uint8_t         len = static_cast<uint8_t>(ie->GetLength());
+            const HeaderIe *ie;
+            uint8_t         len;
+
+            VerifyOrExit(cur + sizeof(HeaderIe) <= footer, index = kInvalidIndex);
```
Irving-cl, line 854:
```suggestion
            VerifyOrExit(cur + sizeof(HeaderIe) < footer, index = kInvalidIndex);
```
The `ie->GetLength()` shouldn't be `0` as I know. So here if it reaches `footer`, I think it's invalid.

abtink: I think an IE header can have no content so zero can be a valid value for `ie->GetLength()` (which is the content length). The main case for this is the "Header Termination IE 1/2" which have no content and are just the IE header (with specific IDs).

abtink: The "Header Termination IE" does indeed have zero length. Here is code preparing it to be appended to a frame:
https://github.com/openthread/openthread/blob/3c132d46200b26fbdea72d39d2c3c7bdcc3b5f29/src/core/mac/mac.cpp#L2077-L2079

Irving-cl: Ok. Sorry for the mistake.

abtink: No worries. :)

PR #5344, opened 2020-08-05
[cli] support reading individual operational dataset values

jwhui:


ndrake:
👍

src/cli/cli_dataset.cpp
```diff
@@ -445,14 +550,29 @@ otError Dataset::ProcessMeshLocalPrefix(uint8_t aArgsLength, char *aArgs[])
...
+    if (aArgsLength == 0)
+    {
+        if (sDataset.mComponents.mIsNetworkNamePresent)
+        {
+            mInterpreter.mServer->OutputFormat("%.*s\r\n", static_cast<uint16_t>(sizeof(sDataset.mNetworkName)),
+                                               sDataset.mNetworkName.m8);
+        }
+        else
+        {
+            mInterpreter.mServer->OutputFormat("none\r\n");
```
simonlingoogle, line 563: How can we tell if there is no `Network Name` in dataset, or the `Network Name` is "none" ?

simonlingoogle: I would prefer (slightly) not to output anything if the corresponding field is not present in the dataset (not just `networkname`, but all fields)
Thoughts? @jwhui

src/cli/cli_dataset.cpp
```diff
@@ -98,7 +98,7 @@ otError Dataset::Print(otOperationalDataset &aDataset)
 
     if (aDataset.mComponents.mIsChannelMaskPresent)
     {
-        mInterpreter.mServer->OutputFormat("Channel Mask: %08x\r\n", aDataset.mChannelMask);
+        mInterpreter.mServer->OutputFormat("Channel Mask: 0x%08x\r\n", aDataset.mChannelMask);
```
wgtdkp, line 101: Although `Channel Mask` is implemented as an `uint32_t`, but it is defined as a bit mask with variable length. Do we have a need to make this change?

jwhui: Thread 1.2 Section 8.10.1.18.1 defines the Channel Mask Entry for IEEE 802.15.4-2006 2.4 GHz PHY as a 4 byte value. If a new Channel Page is defined in the future, we will need to more more changes than just the Channel Mask. Thoughts?

wgtdkp: okay, I am fine with it.

PR #5346, opened 2020-08-06
[mlr] add Commissioner API to register Multicast Listeners

simonlingoogle:
This PR adds a new Commissioner API (**in the last commit**) for registering Multicast Listeners using MLR.req (based on #5292).

This is introduced by `Thread Spec 1.2.1 5.24.5 Commissioner Operation`.

- [x] Add API `otIp6RegisterMulticastListeners` for commissioners
- [x] Add CLI command: `mlr reg <ipaddr> [<ipaddr> ...] [<timeout>]`
- [x] Add registering tests
- [x] Revise Doxygen
- **~~Persistent registration is not supported by this PR.~~**

**MLR PR List:**
- https://github.com/openthread/ot-br-posix/pull/532 add multicast forwarding for MLR Listeners
- https://github.com/openthread/openthread/pull/5346 Add Commissioner API to register Multicast Listeners **<== This PR**
- https://github.com/openthread/openthread/pull/5388 send BMLR.ntf on Backbone link


src/core/backbone_router/bbr_manager.cpp
```diff
@@ -98,61 +101,104 @@ void Manager::HandleMulticastListenerRegistration(const Coap::Message &aMessage,
...
+                Get<NetworkData::Leader>().GetCommissioningDataSubTlv(MeshCoP::Tlv::kCommissionerSessionId));
+
+        VerifyOrExit(commissionerSessionIdTlv != nullptr &&
+                         commissionerSessionIdTlv->GetCommissionerSessionId() == commissionerSessionId,
+                     error = OT_ERROR_PARSE);
+
+        hasCommissionerSessionIdTlv = true;
+    }
+
+    hasTimeoutTlv = hasCommissionerSessionIdTlv &&
```
librasungirl, line 136: rename `hasTimeoutTlv` as `processTimeoutTlv`?
As here indicates the appearance of `ThreadTlv::kTimeout` is not equivalent to process it, right?

src/core/backbone_router/bbr_manager.cpp
```diff
@@ -98,61 +101,104 @@ void Manager::HandleMulticastListenerRegistration(const Coap::Message &aMessage,
...
+    }
+    else
     {
-        bool failed = true;
+        VerifyOrExit(timeout < std::numeric_limits<uint32_t>::max(), status = ThreadStatusTlv::kMlrNoPersistent);
+
+        if (timeout != 0)
+        {
+            timeout = std::min(timeout, static_cast<uint32_t>(Mle::kMlrTimeoutMax));
+            timeout = std::max(timeout, static_cast<uint32_t>(Mle::kMlrTimeoutMin));
```
librasungirl, line 158: In theory, the `timeout` value in `TimeoutTlv` should not be changed in the recipient I think.
When changing it, better to add some log?

Besides, is it necessary to require it `>= kMlrTimeoutMin` here?
For MLR from general devices, it was already pre-processed when getting PBBR Dataset.
For MLR from commissioner which is allowed to take TimeoutTlv, could we simply respect the value inside?

src/core/backbone_router/bbr_manager.cpp
```diff
@@ -98,61 +101,104 @@ void Manager::HandleMulticastListenerRegistration(const Coap::Message &aMessage,
...
+
+    expireTime = TimerMilli::GetNow() + TimeMilli::SecToMsec(timeout);
 
+    for (uint16_t offset = 0; offset < addressesLength; offset += sizeof(Ip6::Address))
+    {
         IgnoreReturnValue(aMessage.Read(addressesOffset + offset, sizeof(Ip6::Address), &address) ==
                           sizeof(Ip6::Address));
 
-        switch (mMulticastListenersTable.Add(address, expireTime))
+        if (timeout == 0)
```
librasungirl, line 169:
```suggestion
        if (hasTimeoutTlv && timeout == 0)
```

include/openthread/ip6.h
```diff
@@ -653,6 +653,58 @@ typedef bool (*otIp6SlaacPrefixFilter)(otInstance *aInstance, const otIp6Prefix
...
 
+/**
+ * This function pointer is called with results of `otIp6RegisterMulticastListeners`.
+ *
+ * @param[in]  aContext  A pointer to the user context.
+ * @param[in]  aError    OT_ERROR_NONE when successfully sent MLR.req and received MLR.rsp,
+ *                       OT_ERROR_BUSY when a previous registration was ongoing,
+ *                       OT_ERROR_INVALID_ARGS when the one or more arguments were invalid,
+ *                       OT_ERROR_INVALID_STATE when the device had no active Commissioner session,
+ *                       OT_ERROR_INVALID_STATE when there was no Primary Backbone Router in Thread network,
```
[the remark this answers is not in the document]
simonlingoogle: Yes it is. 
We conventionally returns  `OT_ERROR_INVALID_STATE` when there is no PBBR in the network. 

I was intended to add a new `OT_ERROR_NO_PRIMARY_BACKBONE_ROUTER` error, but thought it's better added in a separate PR because other code needs to be revised accordingly too.

src/cli/cli.cpp
```diff
@@ -2018,6 +2021,97 @@ void Interpreter::ProcessMasterKey(uint8_t aArgsLength, char *aArgs[])
     AppendResult(error);
 }
 
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
```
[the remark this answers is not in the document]
simonlingoogle: I think `mlr` command could be having more sub-commands in the future.

src/core/backbone_router/bbr_manager.cpp
simonlingoogle: Here the code handles TimeoutTLV which is for Commissioners only.
Even if we allow Commissioner to configure Timeout TLV < `kMlrTimeoutMin`, it's useless because devices will treat `MLR Timeout < kMlrTimeoutMin` as `kMlrTimeoutMin` which is what Thread Spec. requires. 
So, I would prefer to avoid such "invalid" configuration to be applied in the first place.

It's a good idea to add a log here.

src/core/backbone_router/bbr_manager.cpp
simonlingoogle: No need to check `processTimeoutTlv` since `timeout` is assigned `config.mMlrTimeout` when  `!processTimeoutTlv`, and `config.mMlrTimeout` must be non-zero since we check that this there must be a PBBR previously.

include/openthread/backbone_router_ftd.h
```diff
@@ -193,6 +193,110 @@ void otBackboneRouterConfigNextDuaRegistrationResponse(otInstance *
...
+typedef uint16_t otBackboneRouterMulticastListenerIterator; ///< Used to iterate through Multicast Listeners.
+
+/**
+ * This structure represents a Backbone Router Multicast Listener info.
+ *
+ */
+typedef struct otBackboneRouterMulticastListenerInfo
+{
+    otIp6Address mAddress;
+    uint32_t     mTimeout;
```
wgtdkp:
```suggestion
    uint32_t     mTimeoutInMilliseconds;
```

I think putting the scale with the timeout value is more readable. So the reader don't need to do a research to find out the scale by themselves. Thoughts?

src/core/thread/mlr_manager.cpp
```diff
@@ -304,6 +283,115 @@ void MlrManager::SendMulticastListenerRegistration(void)
...
+    VerifyOrExit(aContext == nullptr || aCallback != nullptr, error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(Get<MeshCoP::Commissioner>().IsActive(), error = OT_ERROR_INVALID_STATE);
+
+    // Only allow one outstanding registration if callback is specified.
+    VerifyOrExit(aCallback == nullptr || mRegisterMulticastListenersCallback == nullptr, error = OT_ERROR_BUSY);
+
+    if (aCallback != nullptr)
+    {
+        SuccessOrExit(
+            error = SendMulticastListenerRegistrationMessage(
```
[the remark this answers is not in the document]
simonlingoogle: I am following our convention to only trigger callback if the `Send` call was successful. 
For example, Coap::Send only triggers the response handler if the message was sent successfully. 
We have similar implementation for `PanIdQueryClient::SendQuery`. 

Moreover, when `otIp6RegisterMulticastListeners` returns an error, users already has a chance to handle it. They don't need to handle the error in callback.  

Thoughts?

include/openthread/backbone_router_ftd.h
wgtdkp: By the way, I remember that MLR.req is measuring the timeout value in `Seconds`. Is there a good reason to use `Milliseconds` at here?

src/cli/cli.cpp
```diff
@@ -2057,6 +2060,92 @@ void Interpreter::ProcessMasterKey(uint8_t aArgsLength, char *aArgs[])
...
+}
+
+void Interpreter::HandleMlrRegResult(otError             aError,
+                                     uint8_t             aMlrStatus,
+                                     const otIp6Address *aFailedAddresses,
+                                     uint8_t             aFailedAddressNum)
+{
+    if (aError == OT_ERROR_NONE)
+    {
+        OutputFormat("status %d, failed %d\r\n", aMlrStatus, aFailedAddressNum);
```
wgtdkp, line 2135:
```suggestion
        OutputFormat("status %d, %d failed\r\n", aMlrStatus, aFailedAddressNum);
```

src/cli/cli.hpp
```diff
@@ -354,6 +354,23 @@ class Interpreter
     void ProcessLeaderWeight(uint8_t aArgsLength, char *aArgs[]);
 #endif
     void ProcessMasterKey(uint8_t aArgsLength, char *aArgs[]);
+#if OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+    void ProcessMlr(uint8_t aArgsLength, char *aArgs[]);
+
+#if OPENTHREAD_CONFIG_COMMISSIONER_ENABLE
```
wgtdkp, line 360: Yes, `MLR.req` is a message in scope of `MGMT`. So I think the command should be defined under `commissioner_cli.hpp`, right?

src/core/backbone_router/bbr_manager.cpp
```diff
@@ -119,39 +142,70 @@ void Manager::HandleMulticastListenerRegistration(const Coap::Message &aMessage,
...
-            OT_ASSERT(false);
         }
+    }
+
+    expireTime = TimerMilli::GetNow() + TimeMilli::SecToMsec(timeout);
 
-        if (failed)
+    for (uint16_t offset = 0; offset < addressesLength; offset += sizeof(Ip6::Address))
+    {
+        IgnoreReturnValue(aMessage.Read(addressesOffset + offset, sizeof(Ip6::Address), &address) ==
```
wgtdkp, line 173: Do you mean `VerifyOrExit`?

src/core/config/tmf.h
```diff
@@ -189,7 +189,12 @@
  * This is compulsory for 1.2 FTD.
  *
  */
+#ifdef OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE
+#error \
+    "Don't define `OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE` manually. It's an alias of `((OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2) && OPENTHREAD_FTD)`."
```
wgtdkp, line 194:
```suggestion
    "Don't define `OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE` manually. It's already defined when `((OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2) && OPENTHREAD_FTD)`."
```

src/core/thread/mlr_manager.cpp
```diff
@@ -403,6 +496,43 @@ void MlrManager::HandleMulticastListenerRegistrationResponse(Coap::Message *
...
+
+    if (ThreadTlv::FindTlvValueOffset(*aMessage, IPv6AddressesTlv::kIPv6Addresses, addressesOffset, addressesLength) ==
+        OT_ERROR_NONE)
+    {
+        VerifyOrExit(addressesLength % sizeof(Ip6::Address) == 0, error = OT_ERROR_PARSE);
+        VerifyOrExit(addressesLength / sizeof(Ip6::Address) <= kIPv6AddressesNumMax, error = OT_ERROR_PARSE);
+
+        for (uint16_t offset = 0; offset < addressesLength; offset += sizeof(Ip6::Address))
+        {
+            IgnoreReturnValue(aMessage->Read(addressesOffset + offset, sizeof(Ip6::Address),
```
wgtdkp, line 523: ditto.

include/openthread/backbone_router_ftd.h
```diff
@@ -193,6 +193,110 @@ void otBackboneRouterConfigNextDuaRegistrationResponse(otInstance *
...
+ *
+ */
+typedef enum
+{
+    OT_BACKBONE_ROUTER_MULTICAST_LISTENER_ADDED   = 0, ///< Multicast Listener was added.
+    OT_BACKBONE_ROUTER_MULTICAST_LISTENER_REMOVED = 1, ///< Multicast Listener was removed or expired.
+} otBackboneRouterMulticastListenerEvent;
+
+/**
+ * This function pointer is called whenever the Multicast Listeners change.
```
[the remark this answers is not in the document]
simonlingoogle: We used `This function pointer ...`  most of time in the current code base. 
I think it's best to just follow conventions.

include/openthread/ip6.h
```diff
@@ -653,6 +653,58 @@ typedef bool (*otIp6SlaacPrefixFilter)(otInstance *aInstance, const otIp6Prefix
...
+/**
+ * This function registers Multicast Listeners to Primary Backbone Router.
+ *
+ * Note: only available when both `OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE` and
+ * `OPENTHREAD_CONFIG_COMMISSIONER_ENABLE` are enabled)
+ *
+ * @param[in]  aInstance    A pointer to an OpenThread instance.
+ * @param[in]  aAddresses   A Multicast Address Array to register.
+ * @param[in]  aAddressNum  The number of Multicast Address to register (0 if @p aAddresses is NULL).
+ * @param[in]  aTimeout     A pointer to the timeout value to be included in MLR.req. A timeout value of 0 removes the
```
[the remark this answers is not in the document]
simonlingoogle: it's in seconds

src/core/backbone_router/bbr_manager.cpp
simonlingoogle: We already made sure there is enough data to read. 
But should revise to 
```c++
IgnoreReturnValue(aMessage.Read(addressesOffset + offset, sizeof(Ip6::Address), &address));
```

src/core/config/tmf.h
simonlingoogle: I would prefer the current wording. 
Thoughts @librasungirl

simonlingoogle: Your wording sounds like `OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE` is not defined if not `((OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2) && OPENTHREAD_FTD)`
But in fact `OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE` is always pre-defined and users should never define it manually.

src/core/thread/mlr_manager.cpp
simonlingoogle: It should be 
```
IgnoreReturnValue(
                aMessage->Read(addressesOffset + offset, sizeof(Ip6::Address), &aFailedAddresses[aFailedAddressNum]));
```

Note that it can't fail here.

src/cli/cli.cpp
simonlingoogle: modified

src/cli/cli.hpp
simonlingoogle: I think you meant `cli_commissioner.hpp`. 
`cli_commissioner.hpp` only contains definitions for `commissioner ...` command. But `mlr ...` command will very likely be extended to have other commands that are not only for commissioners. So I would prefer to have `mlr` command defined in the `cli.hpp`.

src/core/config/tmf.h
wgtdkp: I am fine to keep it.

include/openthread/backbone_router_ftd.h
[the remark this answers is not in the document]
wgtdkp: Okay, it just reads strange to me that a `function pointer` is being called. It is defined like a pointer because of limitation of the `C` language.

src/core/thread/mlr_manager.cpp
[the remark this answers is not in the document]
librasungirl: It seems not consistent with error doxygen of the callback in this PR

https://github.com/openthread/openthread/blob/865fa26bd6b0dfcb473454540d618529d33affd5/include/openthread/ip6.h#L664-L665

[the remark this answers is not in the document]
simonlingoogle: Good catch! I think I need to remove errors that couldn't happen in the callback.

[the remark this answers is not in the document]
simonlingoogle: revised

include/openthread/ip6.h
```diff
@@ -653,6 +653,55 @@ typedef bool (*otIp6SlaacPrefixFilter)(otInstance *aInstance, const otIp6Prefix
...
+ * @param[in]  aAddresses   A Multicast Address Array to register.
+ * @param[in]  aAddressNum  The number of Multicast Address to register (0 if @p aAddresses is NULL).
+ * @param[in]  aTimeout     A pointer to the timeout value (in seconds) to be included in MLR.req. A timeout value of 0
+ *                          removes the corresponding Multicast Listener. If NULL, MLR.req would have no Timeout Tlv by
+ *                          default.
+ * @param[in]  aCallback    A pointer to the callback function.
+ * @param[in]  aContext     A pointer to the user context.
+ *
+ * @retval OT_ERROR_NONE          Successfully sent MLR.req.
+ * @retval OT_ERROR_INVALID_ARGS  The Multicast Addresses pointed by @p aAddresses is NULL and there is no subscribed
```
librasungirl, line 692: nit: add document for extra error retval
`OT_ERROR_INVALID_STATE`
`OT_ERROR_BUSY`
`OT_ERROR_NO_BUFS`

src/core/thread/mlr_manager.cpp
```diff
@@ -289,9 +399,25 @@ void MlrManager::SendMulticastListenerRegistration(void)
...
+        SuccessOrExit(error = ThreadTlv::AppendUint16Tlv(*message, ThreadTlv::kCommissionerSessionId,
+                                                         Get<MeshCoP::Commissioner>().GetSessionId()));
+    }
+
+    if (aTimeout != nullptr)
+    {
+        SuccessOrExit(error = Tlv::AppendUint32Tlv(*message, ThreadTlv::kTimeout, *aTimeout));
+    }
+#else
+    OT_UNUSED_VARIABLE(aTimeout);
```
librasungirl, line 418: nit: move to the beginning.

https://github.com/openthread/openthread/blob/master/STYLE_GUIDE.md

> Unused parameters
All unused parameters shall be declared as such using the OT_UNUSED_VARIABLE macro at the top of a function or method before all local variable declarations.

wgtdkp (approved):
LGTM 👍

librasungirl (approved):
LGTM 👍

src/core/backbone_router/bbr_manager.cpp
```diff
@@ -33,6 +33,9 @@
 
 #include "bbr_manager.hpp"
 
+#include <algorithm>
```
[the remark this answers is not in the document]
simonlingoogle: It's for std::min, and std::max.

src/core/thread/mlr_manager.cpp
```diff
@@ -280,6 +281,117 @@ void MlrManager::SendMulticastListenerRegistration(void)
...
+            error = SendMulticastListenerRegistrationMessage(
+                aAddresses, aAddressNum, aTimeout, &MlrManager::HandleRegisterMulticastListenersResponse, this));
+
+        mRegisterMulticastListenersCallback = aCallback;
+        mRegisterMulticastListenersContext  = aContext;
+    }
+    else
+    {
+        SuccessOrExit(
+            error = SendMulticastListenerRegistrationMessage(aAddresses, aAddressNum, aTimeout, nullptr, nullptr));
```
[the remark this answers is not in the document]
simonlingoogle: The practical question is what values should we set if arguments `aCallback` and `aContext` are both nullptr ?
We can't set `mRegisterMulticastListenersCallback` to `aCallback` in such case, and had to choose another value to represent the null callback. 
Personally I prefer to always set `mRegisterMulticastListenersCallback`. But I was not sure what value to set. 
Thoughts? @jwhui

src/cli/cli.cpp
```diff
@@ -2057,6 +2060,92 @@ void Interpreter::ProcessMasterKey(uint8_t aArgsLength, char *aArgs[])
...
+{
+    otError      error = OT_ERROR_NONE;
+    otIp6Address addresses[kIPv6AddressesNumMax];
+
+    VerifyOrExit(aArgsLength >= 1, error = OT_ERROR_INVALID_ARGS);
+
+    for (uint8_t i = 0; i < aArgsLength; i++)
+    {
+        error = otIp6AddressFromString(aArgs[i], &addresses[i]);
+        VerifyOrExit(error == OT_ERROR_NONE || (i > 0 && i == aArgsLength - 1), error = OT_ERROR_INVALID_ARGS);
```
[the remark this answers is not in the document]
simonlingoogle: refactored

src/cli/cli.cpp
```diff
@@ -2057,6 +2060,92 @@ void Interpreter::ProcessMasterKey(uint8_t aArgsLength, char *aArgs[])
...
+
+            SuccessOrExit(error = ParseUnsignedLong(aArgs[i], value));
+
+            timeout = static_cast<uint32_t>(value);
+            error = otIp6RegisterMulticastListeners(mInstance, addresses, i, &timeout, Interpreter::HandleMlrRegResult,
+                                                    this);
+        }
+        else if (i == aArgsLength - 1)
+        {
+            error = otIp6RegisterMulticastListeners(mInstance, addresses, aArgsLength, nullptr,
```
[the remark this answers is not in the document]
simonlingoogle: refactored

src/cli/cli.cpp
```diff
@@ -2057,6 +2060,92 @@ void Interpreter::ProcessMasterKey(uint8_t aArgsLength, char *aArgs[])
...
+void Interpreter::ProcessMlrReg(uint8_t aArgsLength, char *aArgs[])
+{
+    otError      error = OT_ERROR_NONE;
+    otIp6Address addresses[kIPv6AddressesNumMax];
+
+    VerifyOrExit(aArgsLength >= 1, error = OT_ERROR_INVALID_ARGS);
+
+    for (uint8_t i = 0; i < aArgsLength; i++)
+    {
+        error = otIp6AddressFromString(aArgs[i], &addresses[i]);
```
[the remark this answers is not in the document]
simonlingoogle: refactored

src/core/backbone_router/bbr_manager.cpp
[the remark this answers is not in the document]
simonlingoogle: Defined `OT_MIN` and `OT_MAX` in common/code_utils.hpp`.

Using <algorithm> had been fine. The only issue that could happen is conflict of `operator new` since it's both defined in STL and `common/new.hpp`. Do you think simple utilities like `std::min`, `std::max` should be allowed ?

src/core/thread/mlr_manager.cpp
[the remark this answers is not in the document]
jwhui: Would always setting `mRegisterMulticastListenersCallback` to `aCallback` work? If `aCallback` is `nullptr`, we simply avoid calling it in the callback.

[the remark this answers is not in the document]
simonlingoogle: Currently the code uses `mRegisterMulticastListenersCallback != nullptr` to check for an outstanding registration. We are going to need another (bool) variable for that purpose if `mRegisterMulticastListenersCallback` might be nullptr.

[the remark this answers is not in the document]
jwhui: Is the desire here that we want to call `RegisterMulticastListeners()` with `aCallback = nullptr` even when there is an outstanding registration with callback?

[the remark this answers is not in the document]
simonlingoogle: > Is the desire here that we want to call `RegisterMulticastListeners()` with `aCallback = nullptr` even when there is an outstanding registration with callback?

I don't think so. 

The current code allows calling `RegisterMulticastListeners()` with `aCallback = nullptr` even when there is an outstanding registration with callback. But it's nothing more than a side-effect of the decision to use `mRegisterMulticastListenersCallback != nullptr` to check for the outstanding registration. Maybe it's an advantage, but  hardly a useful one. 

I think using another bool variable to represent whether there is an outstanding registration should be a better idea.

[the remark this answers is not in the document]
simonlingoogle: revised accordingly.

PR #5355, opened 2020-08-07
[mle] log handling parent request failures

bukepo:
This commit adds a few logs when handling MLE Parent Request, so that when a child failed to receive a response, the logs can tell what happened.

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -1673,7 +1673,7 @@ void MleRouter::HandleParentRequest(const Message &aMessage, const Ip6::MessageI
 
     if (child == nullptr)
     {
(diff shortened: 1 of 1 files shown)


src/core/thread/mle_router.cpp
```diff
@@ -1997,6 +1997,7 @@ void MleRouter::SendParentResponse(Child *aChild, const Challenge &aChallenge, b
     if (error != OT_ERROR_NONE && message != nullptr)
     {
         message->Free();
+        otLogWarnMle("Failed to send Parent Response: %s", otThreadErrorToString(error));
```
simonlingoogle, line 2000: Should also log when `message == nullptr`?

PR #5358, opened 2020-08-07
[build] update GN build file for OpenThread

mspang:
Add the CLI & configuration to the build file.


include/openthread/instance.h
```diff
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (22)
+#define OPENTHREAD_API_VERSION (23)
```
bukepo, line 56: submitted https://github.com/openthread/openthread/pull/5370 to avoid this change.

mspang: Done

PR #5362, opened 2020-08-08
[otns] fix otns status emission built by cmake

simonlingoogle:
Seems the strong `otPlatOtnsStatus` is not used, but the weak `otPlatOtnsStatus` was used in the original code if built by cmake. 
This PR fixes this issue. Now both `cmake` and `make` builds correctly.


examples/platforms/simulation/virtual_time/platform-sim.c
```diff
@@ -312,4 +312,25 @@ void otSysProcessDrivers(otInstance *aInstance)
 #endif
 }
 
+#if OPENTHREAD_CONFIG_OTNS_ENABLE
+
+#include <openthread/platform/alarm-micro.h>
```
wgtdkp, line 317: Looks like we don't actually need this header but include it to force the linker using the `otPlatOtnsStatus` defined here, right?

As a work-around, I think we should leave a comment (`TODO` or `FIXME`) here to make the intention clear.

simonlingoogle: I copied it from the original `virtual_time/otns.c` to here. There is no intention to use it to force linker to use `otPlatOtnsStatus` defined here. I think I can remove it.

simonlingoogle: removed.

PR #5363, opened 2020-08-08
[notifier] declare 'Notifier::Receiver' constructor as protected

abtink:
This commit declares the `Notifier::Receiver` constructor and
`Handler` type as `protected` (ensures `Receiver` can only be
sub-classed). It also changes the subclasses to privately inherit from
 `Notifier::Receiver` (this helps hide the `Handler` definition).


abtink:
Closing this in favor of PR https://github.com/openthread/openthread/pull/5365.

PR #5364, opened 2020-08-09
[build] add option to build 1.2 alone

Irving-cl:
Currently `script/test` can only build 1.2 together with 1.1. This PR add supports to build 1.2 alone. And when `THREAD_VERSION=1.2` and `INTER_OP=1` are both set, both versions would be built. Sometimes we only want to build 1.2. This would save some time for us.

### DIFF
--- a/.github/workflows/simulation-1.2.yml
+++ b/.github/workflows/simulation-1.2.yml
@@ -50,6 +50,7 @@ jobs:
       COVERAGE: 1
       THREAD_VERSION: 1.2
       VIRTUAL_TIME: 1
+      INTER_OP: 1
       CC: ${{ matrix.compiler.c }}
       CXX: ${{ matrix.compiler.cxx }}
     strategy:
--- a/script/test
+++ b/script/test
@@ -43,6 +43,7 @@ readonly NODE_MODE="${NODE_MODE:-standalone}"
 readonly NODE_TYPE="${NODE_TYPE:-sim}"
 readonly OT_NATIVE_IP="${OT_NATIVE_IP:-0}"
 readonly THREAD_VERSION="${THREAD_VERSION:-1.1}"
+readonly INTER_OP="${INTER_OP:-0}"
 readonly VERBOSE="${VERBOSE:-0}"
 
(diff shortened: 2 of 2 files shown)


.github/workflows/simulation-1.2.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       CXXFLAGS: -${{ matrix.arch }}
       LDFLAGS: -${{ matrix.arch }}
       COVERAGE: 1
-      THREAD_VERSION: 1.2
+      THREAD_VERSION: inter-op
```
simonlingoogle, line 51: Better not change the original behavior? 
Since the valid Thread versions are 1.1 and 1.2, we can use another variable to specify the intention to only build 1.2?

Irving-cl: I think it's more intuitive that if `THREAD_VERSION` is `1.1` then we only build 1.1 and if it's `1.2` then we only build `1.2`. I think it's a little confusing if `1.1` builds only `1.1`, `1.2` builds them all, and another variable makes it build only `1.2`. Thoughts? There won't be big changes on our test configurations.

Irving-cl: I think it would be better to add a VARIABLE `INTER-OP`. If it's `1` and `THREAD_VERSION` is `1.2` we build both versions. Otherwise, it only builds `1.2`.

bukepo (approved):
LGTM:+1:

script/test
```diff
@@ -269,6 +276,7 @@ ENVIRONMENTS:
     VERBOSE         1 to build or test verbosely. The default is 0.
     VIRTUAL_TIME    1 for virtual time, otherwise real time. The default is 1.
     THREAD_VERSION  1.1 for Thread 1.1 stack, 1.2 for Thread 1.2 stack. The default is 1.1.
+    INTER_OP        1 to build 1.1 together. Only works when THREAD_VERSION is 1.2.
```
jwhui, line 279: Document default value here.

PR #5365, opened 2020-08-10
[notifier] simplify how 'Notifier' emits events to core/internal modules

abtink:
This commit changes the way the `Notifier` class emits events to other
OpenThread core/internal modules by having `Notifier` directly invoke
their callbacks, i.e., `Notifier::EmitEvents()` uses a hard-coded list
of all core modules/objects that want to receive events and it then
directly calls `HandleNotifierEvents()` on each object.

This change replaces/removes the `Notifier::Receiver` model and
simplifies the approach (reduces code-size and memory/RAM use).

For external modules (or in case the new model cannot be used by an
internal module), `Notifier` still provides the `RegisterCallback()`
model.

### DIFF
--- a/src/core/backbone_router/bbr_manager.cpp
+++ b/src/core/backbone_router/bbr_manager.cpp
@@ -51,7 +51,6 @@ namespace BackboneRouter {
 
 Manager::Manager(Instance &aInstance)
     : InstanceLocator(aInstance)
-    , Notifier::Receiver(aInstance, Manager::HandleNotifierEvents)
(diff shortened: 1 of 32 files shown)


simonlingoogle:
> This commit changes the way the `Notifier` class emits events to other
> OpenThread core/internal modules by having `Notifier` directly invoke
> their callbacks, i.e., `Notifier::EmitEvents()` uses a hard-coded list
> of all core modules/objects that want to receive events and it then
> directly calls `HandleNotifierEvents()` on each object.

It's a good idea with less ram and code size. Thanks!

PR #5374, opened 2020-08-11
[mesh-forwarder] simplify 'FragmentPriorityList'

abtink:
This commit simplifies management of "Fragment Priority Entries"
defining a new class `FragmentPriorityList` with helper method to
allocate/find entries in/from the list. The `Entry` types is also
moved/defined as a nested type of `FragmentPriorityList`.

### DIFF
--- a/src/core/thread/mesh_forwarder_ftd.cpp
+++ b/src/core/thread/mesh_forwarder_ftd.cpp
@@ -726,17 +726,17 @@ exit:
     return;
 }
 
-bool MeshForwarder::UpdateFragmentLifetime(void)
+bool MeshForwarder::FragmentPriorityList::UpdateOnTimeTick(void)
 {
     bool shouldRun = false;
 
-    for (FragmentPriorityEntry &entry : mFragmentEntries)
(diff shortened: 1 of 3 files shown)


src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -751,46 +751,47 @@ void MeshForwarder::UpdateFragmentPriority(Lowpan::FragmentHeader &aFragmentHead
                                            uint16_t                aSrcRloc16,
                                            Message::Priority       aPriority)
 {
-    FragmentPriorityEntry *entry;
+    FragmentPriorityList::Entry *entry;
```
abtink, line 754: Adding a note on the change here:

This commit changes the order the check.

- We first check if there is an existing entry with a matching source RLOC16 and tag.
   - This handles the case where we may receive duplicate first fragments (this can, for example, happen with MLE msgs since they don't use MAC layer security and therefore won't be filtered as duplicate by `Mac`).
- Only if there is no existing entry, we create a new one when datagram offset is zero (i.e., it's a first frag).

PR #5375, opened 2020-08-11
[github-action] change the way to upload coverage report for 1_2 tests

Irving-cl:
I found that when running `code-cov` for `thread-1-2` in our github actions, the uploaded `.gcov` report file is not correct(Some code that will surely be executed is not hit in the report). 

The possible reason is that in `thread-1-2`, we have 3 different build directories. Thus we have 3 `.gcda` and `gcno` files for one source file. And `codecov-action@v1` isn't smart enough to include all these into one `.gcov` and then upload. Though I don't know the underlying process(how does `codecov-action@v1` generate `.gcov` when there are 3 data files for one source file), I've verified that, uploading report when there's only one build directory and repeating three times for different build directories take effect.

This PR adds a function in `script/test` to tar and untar the build directory. And in the `thread-1.2` action, `codecov-action@v1` is called for 3 times. Before each call, I use tar and untar to make sure currently there's only one valid build directory.

### DIFF
--- a/.github/workflows/simulation-1.2.yml
+++ b/.github/workflows/simulation-1.2.yml
@@ -71,5 +71,21 @@ jobs:
       run: |
         ./script/test unit
         ./script/test cert_suite tests/scripts/thread-cert/v1_2_*
+    - name: Keep-1-2-only
+      run: |
+        ./script/test tar 1.1
+        ./script/test tar 1.2-bbr
+    - name: Codecov
+      uses: codecov/codecov-action@v1
+    - name: Keep-1-2-bbr-only
+      run: |
+        ./script/test tar 1.2
+        ./script/test untar 1.2-bbr
+    - name: Codecov
+      uses: codecov/codecov-action@v1
+    - name: Keep-1-1-only
+      run: |
+        ./script/test tar 1.2-bbr
+        ./script/test untar 1.1
     - name: Codecov
       uses: codecov/codecov-action@v1
--- a/script/test
+++ b/script/test
@@ -332,6 +332,23 @@ do_package()
     ls "${builddir}"/openthread-daemon-*.deb
 }
 
+do_tar()
+{
+    local target_name="openthread-simulation-$1"
+    local build_dir="${OT_BUILDDIR}/cmake"
+    tar -cf "${target_name}.tar" -C "${build_dir}" "${target_name}"
+    mv "${target_name}.tar" "${build_dir}/"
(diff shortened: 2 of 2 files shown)


bukepo:
Not sure if this issue existing in other tests, which may cause the unstable code coverage results. @Moandor-y

Moandor-y:
> Not sure if this issue existing in other tests, which may cause the unstable code coverage results. @Moandor-y

In expect tests for RCP mode, we are building both `simulation` and `posix` which could have the same issue.

Irving-cl:
@Moandor-y 

I think we can merge this PR first and then apply it to other github actions.

Moandor-y:
> @Moandor-y
> 
> I think we can merge this PR first and then apply it to other github actions.

Yes. I'll have a look at that after this is merged.

.github/workflows/simulation-1.2.yml
```diff
@@ -71,5 +71,21 @@ jobs:
       run: |
         ./script/test unit
         ./script/test cert_suite tests/scripts/thread-cert/v1_2_*
+    - name: Keep-1-2-only
+      run: |
+        ./script/test tar 1.1
+        ./script/test tar 1.2-bbr
+    - name: Codecov
+      uses: codecov/codecov-action@v1
```
simonlingoogle, line 79: Have you checked the arguments of  `Codecov` to see if there is any argument we can use to address this problem?

Irving-cl: I found some documents and started an [issue](https://github.com/codecov/codecov-action/issues/106) at `codecov-action`, but unfortunately for now I didn't get anyhing that helps. Anyway I would follow up this question.

simonlingoogle: > I found some documents and started an [issue](https://github.com/codecov/codecov-action/issues/106) at `codecov-action`, but unfortunately for now I didn't get anyhing that helps. Anyway I would follow up this question.

Have you tried the directory parameter? I suspect that it might solve our problem.

Irving-cl: I couldn't find the related document. Is there a document showing how to use the parameter?

Irving-cl: I think it's not the one we want. It's the directory to search for coverage reports, e.g., generated one. What we want is to specify the build directory to search for `*.gcno` and `.gcda` which are used to generate the reports.

PR #5378, opened 2020-08-11
[flash] Implement new settings format

dismirlian:
Hi, this PR addresses issue #4706. It is a work in progress, but I'd appreciate feedback.

"Preparatory" commits (they don't change the format):
- 453c5d1: Hides implementation of record format (moves some internal classes from flash.hpp to flash.cpp)
- 4f6e2c1e0e6bfd11ad67f9f429edb3b5f041dd18: Introduces `SwapHeader` class, in preparation for the new format.
- c5049a3: Refactoring
- 3da2888: Further Refactoring

Tests:
- 2497121 & 5d23dc5: Add tests to check for the issues identified in #4706. The original implementation is preserved in the `tests` directory, in preparation to check for legacy compatibility.

Format v2:
- d087e45: This implements the new flash format, which avoids the identified issues.
- b6ecbcd: This makes final changes for legacy compatibility (writes with v1, reads with v2).
- c5f19da: Fixes an issue identified by @jwhui
- 981eccb: Implements CRC and additional checks to verify the integrity of records.
- 2d334bc: Swap erase counters

Missing:
- [x] Fix atomic setting of record
- [x] Compile switches to opt-in/out legacy compatibility
- [x] Robustness: CRC and additional checks
- [x] Add flash erase counters
- [ ] Cleanup

TBD in a separate PR: Add API and CLI command to get swap erase counters.

Comments/Suggestions?

Thanks,
Diego.


src/core/utils/flash.cpp
```diff
@@ -39,44 +39,281 @@
...
+        kFlagAddComplete = 1 << 1, ///< 0 indicates record write has completed, 1 otherwise.
+        kFlagDelete      = 1 << 2, ///< 0 indicates record was deleted, 1 otherwise.
+        kFlagFirst       = 1 << 3, ///< 0 indicates first record for key, 1 otherwise.
+
+        /* The record's format version is in the upper nibble of mFlags */
+        kFormatTop   = 0xf,
+        kFormatShift = 12,
+
+        /* Initial value of flags: kFlagAddBegin, kFlagAddComplete and format version */
+        kFlagsInit = ((kFormatTop - kFormatV2) << kFormatShift) | (0xfff & ~(kFlagAddBegin | kFlagAddComplete)),
```
jwhui, line 199: It seems we are not attempting to maintain backward compatibility in that the V1 flash implementation will not be able to read the V2 flash implementation records. If that is the case, do we care to maintain the begin/complete flags as well?

src/core/utils/flash.cpp
```diff
@@ -39,44 +39,281 @@
...
+    {
+        OT_ASSERT(aDataLength <= kMaxDataSize);
+        RecordHeader::Init(aKey, aDataLength);
+
+        memcpy(mData, aData, aDataLength);
+
+        /* fill slack space with 0xff to avoid writing 0 more than once to each bit */
+        memset(mData + aDataLength, 0xff, GetDataSize() - aDataLength);
+
+        // TODO: calculate CRC
```
jwhui, line 223: What is the current thinking on where the CRC value will go? In the record header `mReserved` field?

src/core/utils/flash.cpp
```diff
@@ -167,53 +394,45 @@ otError Flash::Get(uint16_t aKey, int aIndex, uint8_t *aValue, uint16_t *aValueL
 
 otError Flash::Set(uint16_t aKey, const uint8_t *aValue, uint16_t aValueLength)
 {
-    return Add(aKey, true, aValue, aValueLength);
+    Delete(aKey, -1);
```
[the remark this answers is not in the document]
dismirlian: Excellent point. We could:
1. Delete after adding (we have to think what to do in some situations, I think it's not as easy as it seems)
2. Restore the `first` flag.
3. Think of another alternative!

src/core/utils/flash.cpp
dismirlian: Actually, we don't need to maintain both flags. This is to simplify some methods which would depend on the format version. For example, see the `IsFree` and `IsIntegrityOk` methods. I think the code is a lot more complex/uglier than it should be, because of legacy compatibility.

src/core/utils/flash.cpp
dismirlian: Yes, that's the idea.

src/core/utils/flash.cpp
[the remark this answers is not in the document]
jwhui: With "delete after adding", we would also need to handle the case that there would be multiple entries after we write the new entry but before we delete the old entry.

I do not have a better proposal than the `first` flag right now, but I am open to other ideas if you have them.

src/core/utils/flash.cpp
jwhui: I understand the need to understand the old format to support forward migration. However, if we are not looking to support backward migration (i.e. writing old format), then I think we should do whatever is most clean going forward.

src/core/utils/flash.cpp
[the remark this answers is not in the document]
dismirlian: Yes, that's the case I was thinking of. Every solution I can imagine seems to use something similar to the `first` flag.

src/core/utils/flash.cpp
dismirlian: Yes, I was hoping to do some "cleanup" after the main functions were implemented.

src/core/utils/flash.cpp
dismirlian: Done.

src/core/utils/flash.cpp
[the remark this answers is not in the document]
dismirlian: Implemented an idea similar to the `First` flag, but instead of meaning "first", it means "invalidate previous" records. 

This is a sublte difference, but avoids the need to mark "first" the record `index == 1` when deleting record `index == 0`. This retains compatibility with the old format's `First` flag. Also, the `InvalidatePrevious` flag implicitly resets the iterator's index to 0. Please see c5f19dac689266ad9f04e2df69cd4cbd3593bebf.

dismirlian:
@jwhui, please review the current status. I'll finish the cleanup of this code based on your suggestions.

src/core/config/flash.h
```diff
@@ -0,0 +1,62 @@
+/*
+ *  Copyright (c) 2019, The OpenThread Authors.
```
jwhui, line 2:
```suggestion
 *  Copyright (c) 2020, The OpenThread Authors.
```

src/core/utils/flash.cpp
```diff
@@ -165,55 +476,41 @@ otError Flash::Get(uint16_t aKey, int aIndex, uint8_t *aValue, uint16_t *aValueL
...
-    if ((mSwapSize - record.GetSize()) < mSwapUsed)
+    if ((mSwapSize - size) < mSwapUsed)
     {
         Swap();
-        VerifyOrExit((mSwapSize - record.GetSize()) >= mSwapUsed, error = OT_ERROR_NO_BUFS);
+        VerifyOrExit((mSwapSize - size) >= mSwapUsed, error = OT_ERROR_NO_BUFS);
     }
 
-    otPlatFlashWrite(&GetInstance(), mSwapIndex, mSwapUsed, &record, record.GetSize());
+    record.Save(&GetInstance(), mSwapIndex, mSwapUsed);
```
jwhui, line 496: Should we validate integrity of record after writing? If integrity check fails, we can try append a new record?

src/core/utils/flash.cpp
```diff
@@ -118,39 +429,39 @@ otError Flash::Get(uint16_t aKey, int aIndex, uint8_t *aValue, uint16_t *aValueL
...
             {
-                uint16_t readLength = *aValueLength;
-
-                if (readLength > record.GetLength())
-                {
-                    readLength = record.GetLength();
-                }
-
-                otPlatFlashRead(&GetInstance(), mSwapIndex, offset + sizeof(record), aValue, readLength);
+                recordHeader.ReadData(&GetInstance(), mSwapIndex, offset, aValue, *aValueLength);
```
jwhui, line 461: Should we validate integrity of record?

src/core/utils/flash.cpp
```diff
@@ -225,71 +522,97 @@ bool Flash::DoesValidRecordExist(uint32_t aOffset, uint16_t aKey) const
...
             continue;
         }
 
-        otPlatFlashRead(&GetInstance(), mSwapIndex, srcOffset, &record, record.GetSize());
-        otPlatFlashWrite(&GetInstance(), dstIndex, dstOffset, &record, record.GetSize());
+        record.LoadData(&GetInstance(), mSwapIndex, srcOffset);
+#if OPENTHREAD_CONFIG_FLASH_LEGACY_COMPATIBILITY
+        record.Fixup();
+#endif
+        record.Save(&GetInstance(), dstIndex, dstOffset);
```
jwhui, line 556: Should we validate integrity of record after writing? If integrity check fails, we can try append a new record?

src/core/utils/flash.cpp
```diff
@@ -225,71 +522,97 @@ bool Flash::DoesValidRecordExist(uint32_t aOffset, uint16_t aKey) const
...
-            error = OT_ERROR_NONE;
+            continue;
         }
 
-        /* [Note] If the operation gets interrupted here and aIndex is 0, the next record (index == 1) will never get
-         * marked as first. However, this is not actually an issue because all the methods that iterate over the
-         * settings area initialize the index to 0, without expecting any record to be effectively marked as first. */
-
-        if ((index == 1) && (aIndex == 0))
+        if ((aIndex == index) || (aIndex == -1))
```
jwhui, line 611: In the case of `aIndex == -1`, I think it would be better to atomically delete all existing records. I think the best way to do this is by append a record that indicates all previous are invalidated. Thoughts?

src/core/utils/flash.cpp
```diff
@@ -29,54 +29,360 @@
...
+        kFormatTop   = 0xb, // This matches the first nibble of the swap header format v1.
+        kFormatShift = 28,
+
+        kMarkerInit = ((uint32_t)(kFormatTop - kFormatV2) << kFormatShift) | kActive,
+    };
+    static_assert(kActive >> kFormatShift == 0, "wrong kActive");
+    static_assert(kActiveMask == (uint32_t)((1 << kFormatShift) - 1), "wrong kActiveMask");
+
+    uint32_t mMarker;
+    uint16_t mEraseCounter;
```
jwhui, line 117: Are we only interested in the first 65k erase cycles? I think it is common for flash components to support 100k flash cycles, for example.

tests/unit/test_flash.cpp
```diff
@@ -183,14 +288,121 @@ void TestFlash(void)
...
+        if (counter < 0xffff)
+        {
+            counter++;
+        }
+
+        VerifyOrQuit(flash.GetEraseCounter() == counter, "GetEraseCounter() did not return expected value");
+    }
+}
+
+void TestFlash(void)
```
jwhui, line 340: Can we add some tests to cover the integrity checking?

jwhui (commented):
@dismirlian , thanks for continuing to drive this forward! This is looking pretty good. Added some comments.

jwhui:
@dismirlian , any updates on this PR?

dismirlian:
@jwhui, sorry for the lack of updates; I've been busy... I'll update this during the following weeks. Thanks!

romacdon:
@dismirlian - I stumbled across [https://github.com/openthread/openthread/issues/4706](url) while searching for open efr32 issues and found it tied to this PR.  Is there any update on this?  Thanks !

dismirlian:
Hi @romacdon, I'm sorry I still have no updates; I'll try to look at this during the next weeks.

jwhui:
@dismirlian , any plans to move this PR forward?

dismirlian:
Sorry for the lack of advance. I can try to take a look and move it forward in the following weeks.

jwhui:
Closing stale PR.

PR #5380, opened 2020-08-12
[github-actions] run packet verification for multiple times

simonlingoogle:
This PR makes packet verification to run for multiple times.

Since now we are developing many packet verification scripts, this PR can help us reduce the chance of merging a verification code that could fail by chance, which is very often for packet verification code.

**Successfully ran for 10 times in 13 minutes.**

### DIFF
--- a/.github/workflows/simulation.yml
+++ b/.github/workflows/simulation.yml
@@ -102,7 +102,7 @@ jobs:
       - name: Bootstrap
         run: |
           sudo rm /etc/apt/sources.list.d/* && sudo apt-get update
-          sudo apt-get --no-install-recommends install -y g++-multilib python3-setuptools python3-wheel
+          sudo apt-get --no-install-recommends install -y g++-multilib python3-setuptools python3-wheel ninja-build
           python3 -m pip install -r tests/scripts/thread-cert/requirements.txt
       - name: Build
         run: |
--- a/.github/workflows/simulation.yml
+++ b/.github/workflows/simulation.yml
@@ -106,11 +106,13 @@ jobs:
           python3 -m pip install -r tests/scripts/thread-cert/requirements.txt
       - name: Build
         run: |
-          ./bootstrap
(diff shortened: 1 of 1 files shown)


jwhui:
> Since now we are developing many packet verification scripts, this PR can help us reduce the chance of merging a verification code that could fail by chance, which is very often for packet verification code.

@simonlingoogle , can you provide some insight on why tests may fail? We need to be careful that we are not hiding bugs by retrying tests.

simonlingoogle:
> > Since now we are developing many packet verification scripts, this PR can help us reduce the chance of merging a verification code that could fail by chance, which is very often for packet verification code.
> 
> @simonlingoogle , can you provide some insight on why tests may fail? We need to be careful that we are not hiding bugs by retrying tests.

@jwhui , just to be clear that this PR does not hide bugs but expose potential bugs because it requires each packet verification test to pass for 10 times consistently. 

As for the reason why packet verification (PV) code might fail, I think it is indeed challenging to verify packets correctly according the the Test Spec strictly:
- The order of packets can be different for each run. If the PV code depends on a specific ordering of packets, it could fail by chance. 
- The packets can be different for different DUT devices. For example, an SED DUT sends more `Child Update Request` messages when attaching. A PV code that finds `Child Update Request` and looks for `AddressRegistrationTLV` could work well for other DUTs but fail for SED devices. 
- If the PV code does not use enough filters for a packet, it might finds a wrong packet in some runs depends on what actually happens in the network. 

In summary, I think it's easy to write PV code that could pass or fail by chance. So, I submitted this PR to be more strict on PV code so that we can finds potential issues easier.

jwhui:
> @jwhui , just to be clear that this PR does not hide bugs but expose potential bugs because it requires each packet verification test to pass for 10 times consistently.

@simonlingoogle , that makes sense to me. We are effectively applying Monte Carlo method here by selecting random seeds each time. I misinterpreted your original comment. Thanks for the clarification.

PR #5382, opened 2020-08-12
[mle] add helper 'InitNetifUnicastAddress()'

abtink:


src/core/thread/mle.cpp
```diff
@@ -200,6 +173,17 @@ Mle::Mle(Instance &aInstance)
     // to the Link- and Realm-Local All Thread Nodes multicast addresses.
 }
 
+void Mle::InitNetifUnicastAddress(Ip6::NetifUnicastAddress &aAddress)
```
jwhui, line 176: Should we also leverage this for other components? For example:
https://github.com/openthread/openthread/blob/7dc3b2696ade4a524b8e2e33cd73bed2668af408/src/core/meshcop/border_agent.cpp#L355-L361

src/core/thread/mle.hpp
```diff
@@ -1622,6 +1622,7 @@ class Mle : public InstanceLocator
         TimeMilli    mSendTime;    // Time when the message shall be sent.
     };
 
+    void        InitNetifUnicastAddress(Ip6::NetifUnicastAddress &aAddress);
```
simonlingoogle, line 1625: make this method static?

src/core/thread/mle.cpp
abtink: Thinking more on this, I think may be best to add one or more such `Init()` methods in `NetifUnicastAddress` class itself. This way can be used by all other classes.

abtink:
Closing this PR in favor of PR https://github.com/openthread/openthread/pull/5390.

PR #5383, opened 2020-08-12
[netif] invoke address callbacks for external multicast addresses

simonlingoogle:
Address callback should be called for external multicast addresses but seems to be missing.

### DIFF
--- a/src/core/net/netif.cpp
+++ b/src/core/net/netif.cpp
@@ -343,6 +343,9 @@ otError Netif::SubscribeExternalMulticast(const Address &aAddress)
     mMulticastAddresses.Push(*entry);
     Get<Notifier>().Signal(kEventIp6MulticastSubscribed);
 
+    VerifyOrExit(mAddressCallback != nullptr, OT_NOOP);
+    mAddressCallback(&aAddress, kMulticastPrefixLength, /* IsAdded */ true, mAddressCallbackContext);
+
 exit:
     return error;
 }
@@ -363,6 +366,9 @@ otError Netif::UnsubscribeExternalMulticast(const Address &aAddress)
     mExtMulticastAddressPool.Free(static_cast<ExternalNetifMulticastAddress &>(*entry));
 
     Get<Notifier>().Signal(kEventIp6MulticastUnsubscribed);
+
+    VerifyOrExit(mAddressCallback != nullptr, OT_NOOP);
+    mAddressCallback(&aAddress, kMulticastPrefixLength, /* IsAdded */ false, mAddressCallbackContext);
 
 exit:
     return error;


src/core/net/netif.cpp
```diff
@@ -364,6 +367,9 @@ otError Netif::UnsubscribeExternalMulticast(const Address &aAddress)
 
     Get<Notifier>().Signal(kEventIp6MulticastUnsubscribed);
 
+    VerifyOrExit(mAddressCallback != nullptr, OT_NOOP);
+    mAddressCallback(&aAddress, kMulticastPrefixLength, /* IsAdded */ false, mAddressCallbackContext);
+
```
abtink, line 372: This behavior is intentional (not a bug).

The `otIp6AddressCallback` should be called when an *internal* IPv6 address is added or removed (not on external addresses).
https://github.com/openthread/openthread/blob/7dc3b2696ade4a524b8e2e33cd73bed2668af408/include/openthread/ip6.h#L418-L441

If I recall @bukepo  added this feature. He may be able to add more info about this behavior.

From what I recall, the idea for this behavior is as follows:
Users of OT would be the ones adding/removing external addresses (through OT public APIs), so they should be aware of address being added/removed and there is no need to issue a callback to the user about the address that they added/removed.

gjc13: The case here is what if someone calls `otIp6SubscribeMulticastAddress` on the host?
Currently the `mAddressCallback` won't be called and this address won't be forwarded to the linux interface. The reverse path works well. If you call `IP6_JOIN_GROUP` with a multicast address on wpan0, you'll find this address both on the interface and the OpenThread stack.
Shall we make this path the only supported way to manipulate multicast groups on the host or shall we allow `otIp6SubscribeMulticastAddress` as well?

simonlingoogle: What you said make sense to some extent. 

I was using ot-daemon and  added multicast address `ff04::abcd` via CLI command `ipmaddr add ff04::abcd`, but the address does not show up in the `wpan0` interface.

Either we use the fix of this PR, or we fix the behavior of `ipmaddr add`. But `ip maddr` is part of the CLI app, meanwhile the address callback is registered by posix platform. It seems difficult for CLI app to notify the posix platform of the new multicast address.

That being said, I think it is consistent to always trigger address callback in the Thread core whenever a multicast address is subscribed or unsubscribed.

Thoughts? @abtink @bukepo

simonlingoogle: > The case here is what if someone calls `otIp6SubscribeMulticastAddress` on the host?
> Currently the `mAddressCallback` won't be called and this address won't be forwarded to the linux interface. The reverse path works well. If you call `IP6_JOIN_GROUP` with a multicast address on wpan0, you'll find this address both on the interface and the OpenThread stack.
> Shall we make this path the only supported way to manipulate multicast groups on the host or shall we allow `otIp6SubscribeMulticastAddress` as well?

What happens if we failed to add the multicast address to the Thread? I mean the number of external addresses that can be subscribed in Thread is limited (only 2 by default). 
Should we leave the multicast group if failed to add the MA in Thread?

abtink: The idea for this behavior of address callback (if I recall) is to make the user/host side code simpler on what to do on receiving the `callback`:
- Anytime the callback is received, we know it's an internal OT address, so should be safe to add/remove the same address on host (linux) interface.
- Anytime an external address is being added, host would know the origin (if it is already from linux interface or not) and can take the proper action (add it on both OT and linux interface or just on OT).

If we change the code to invoke the callback on all addresses (external and internal), the code handling it on host side would be more complex and host on rx of callback needs to decide whether it would need to update the linux interface or not (there are race timing issues (adding/removing address on linux netif can take time) to consider as well).

simonlingoogle: > The idea for this behavior of address callback (if I recall) is to make the user/host side code simpler on what to do on receiving the `callback`:
> 
> * Anytime the callback is received, we know it's an internal OT address, so should be safe to add/remove the same address on host (linux) interface.
> * Anytime an external address is being added, host would know the origin (if it is already from linux interface or not) and can take the proper action (add it on both OT and linux interface or just on OT).
> 
> If we change the code to invoke the callback on all addresses (external and internal), the code handling it on host side would be more complex and host on rx of callback needs to decide whether it would need to update the linux interface or not (there are race timing issues (adding/removing address on linux netif can take time) to consider as well).

I can see why the current code can make Host logic simpler. 

However I think the current code is problematic. 

**Using normal operations should never cause OT to enter a inconsistent state.** But with the current code we can have inconsistent MA subscription in Thread and Host:

- Adding MA by `ipmaddr add` adds MA to Thread but not on Host.
- Add MA by IP6_JOIN_GROUP and then remove it by `ipmaddr del` have the same effect. 

Same issue with `ipaddr` for Unicast addresses.

Are we planing to ban `ipaddr ` and `ipmaddr` commands on Hosts? 

Furthermore, since the Host program is single threaded. I don't think it's difficult to handle race timing issues.

abtink: In my view, OpenThread CLI is a mirror/representation of OT public APIs, so all it should control is the OT itself.
So the CLI `ipmaddr` command should behave same way as `otIp6SubscribeMulticastAddress()` API.

OT core itself is not managing/creating any host/linux interface so should not be aware of it and/or control it in any way (same way that it would not be aware of any RTOS interfaces if used on an RTOS-based platform).

So if a user uses OT CLI command, it should be treated same as if public OT APIs are being called.

-------

The way I see it, is that syncing between host (interfaces) and OT is the job of a "host driver".

I would say "ot-daemon" is a host driver.

`wpantund` is also a  "host driver" and we have mechanisms in place in its design to ensure that it syncs addresses/routes/prefixes/etc between host and OT/NCP (it took some iterations and many bugs/issues with earlier approaches till we converged on this design in `wpantund`).

- In short, `wpantund` has a module which tracks all such entries (address/route/prefix/etc) and also tracks their origin (i.e. who added the entry, is it added by OT/NCP, or host interface, or user through other APIs, etc).
-  It has code that syncs all entries (e.g., if entry added by host interface push it to NCP, and vice versa). 
- This also allows us to protect entries, for example an address added by OT core cannot be removed by host/user. 
- Note that `wpantund` does not actually use the `address` callback feature we are discussing here and always gets the full list of addresses from NCP/OT.

On how to handle failures (when syncing entries), this is how `wpantund` behaves:
- We do have retry/wait mechanism if entry cannot be added on host/linux interface (we keep track of addresses to be added/removed and wait till host linux interface is ready/up to add/remove them)
- On OT/NCP side, we have the situation that we may run out of address space (as you mentioned there is a build-time configurable but fixed number of unicast/multicast addresses that is supported by OT).
  - `wpantund` behavior is that it would retry and then after some failures would assert (restart `wpantund`)
  - So while there are more addresses on host linux interface than what can be supported by the OT/NCP, `wpantund` would be in an assert loop (practically disabling Thread use).
  - While this is not perfect, choosing this behavior was very much intentional (in our internal projects) as it was the best way to indicate that there is something wrong with how the network is configured (too many addresses)

We considered other options: 
- If we ran out of address space on OT, just log it as error, and ignore it and let things proceed
   -  Not good, since user may want to use the added address and it won't work since it is not really added/supported on OT/NCP.
- If the address from host interface origin cannot be added on OT/NCP, go back and try to remove it on host interface     
    - Not good, since user adds the address on host interface and it succeeds but later it disappears.
- Provide API in driver to add/remove entries (addresses) and don't let user add/remove address on host linux interface - this would allow us to return error from API if out of space in OT. 
   - While I was fine with this approach, it is indeed limiting as it disallows user to treat Thread `wpan` interface like other interfaces (this was far from desirable by folks writing next layer (application) code).

abtink: Adding one more note on this:

If I remember correctly, the address callback feature may be used on some of our internal RTOS based projects. If the behavior is changed to invoke callback on all address (internal and external), we may need to check and update its use in our projects.

bukepo: As @abtink pointed out, the behavior is intentional. I agree with Abtin and tend not to change the behavior of the address callback.

To prevent the addresses diverging between host and OpenThread core and keep OpenThread's CLI command consistent, how about we allow specializing the `ipaddr` and `ipmaddr` implementation in the CLI library so that in POSIX daemon, it can be implemented based on Linux/macOS system APIs?

For error handling, the address callback was designed to keep addresses identical between host and OpenThread core, and if failed, I think it should be treated as a fatal error.

Thoughts? @abtink @simonlingoogle

simonlingoogle: > how about we allow specializing the ipaddr and ipmaddr implementation in the CLI library so that in POSIX daemon, it can be implemented based on Linux/macOS system APIs?

It could work but I think it's not elegant to insert `ot-daemon` specific code to the CLI app. To implement it, we need to call syscall like `setsockopt(sIpFd, IPPROTO_IPV6, (aIsAdded ? IPV6_JOIN_GROUP : IPV6_LEAVE_GROUP), &mreq, sizeof(mreq))` from CLI app. Not sure how CLI app can get the `sIpFd` descriptor from Posix. 

BTW, If we are not going to change the behavior of address callbacks, I would recommend to change the API from `otIp6SetAddressCallback` to `otIp6SetInternalAddressCallback`, so as to reflect it's actual functionality.

simonlingoogle: > For error handling, the address callback was designed to keep addresses identical between host and OpenThread core, and if failed, I think it should be treated as a fatal error.

I think you are talking about adding internal addresses to Host interface failed in address callback. In such case, I would agree that it's a fatal error and something should have gone wrong. 

But for the cases when user adds more address than OT can support on the `wpan0` interface, the current `ot-daemon` is printing a error log and leaving the address on wpan0.  It neither removes the IP address from wpan interface, nor crashes like `wpantund`. 

Personally I think this behavior is fine because it's kind like a best-effort strategy, even though I think removing the address would be a better choice. For `wpantund`'s strategy of crashing in such case, I think it's fine for internal projects (maybe even helpful for debugging), but not a proper choice for `ot-daemon` to be used by OT community. **The ip addresses added by user to wpan interface are indeed user inputs which should (and can) be handled gracefully.**

Thoughts? @abtink @bukepo

simonlingoogle:
Close since it's intended behavior.

PR #5389, opened 2020-08-13
[mac] change address filter and seek function to accept frame without PAN ID

brianljt:
In 802.15.4-2015 spec Table 7-2, it is possible to receive frame with destination address but without destination PAN ID. This PR changes the address filter function to avoid dropping such frames.

image


abtink:
Thanks @brianljt.
One question on this:  is there any use of such a frame format within Thread?

jwhui:
> One question on this: is there any use of such a frame format within Thread?

@abtink , I don't think the Thread 1.2 specification prohibits eliding the Destination PAN field.

src/core/mac/mac_frame.cpp
```diff
@@ -972,6 +972,46 @@ const TimeIe *Frame::GetTimeIe(void) const
...
+    Mac::Address dst;
+    Mac::PanId   panid;
+
+    SuccessOrExit(GetDstAddr(dst));
+
+    // Verify address match
+    switch (dst.GetType())
+    {
+    case Address::kTypeShort:
+        VerifyOrExit((aRxOnWhenIdle && dst.IsBroadcast()) || dst.GetShort() == aShortAddress, OT_NOOP);
```
abtink, line 990: I am not sure about moving this logic in a method in `Mac::Frame` class. I suggest we keep it in `Mac` class (as before).
- `Mac:Frame` class is a representation of a frame.
- Make sense for it to provide methods to parse the frame, and get/set different fields in the frame.
- Doing such checks that are tied to whether `aRxOnWhenIdle` does not fit well here.
- The `aRxOnWhenIdle` is really a `Mac` state variable and `Frame` should not be aware of it or care about it. 

I understand that you moved this this here so that it can be reused from `examples/platform/utils` library code. I think the `platform/util` code is behaving more like MAC layer itself than just frame parser.  In this case, I would argue that having a clean model in what `Mac::Mac` class should do, and what `Mac::Frame` class should provide,  is more valuable over maybe not being able to share a bit of code between platform util  library and OT core.

brianljt: That make sense to me. But the cost is some duplication of code. I would like to get more feedback on this. @bukepo @jwhui , how do you think?

brianljt: Another way of thinking this problem, we can consider this address filter as a `Mac` method and is only needed in `Mac` layer.

For `Radio` which needs to do address filtering, most of the case, it will be done at Chip hardware which does not need this utility function. For simulation radio, as it just simulate a hardware behavior, we can add this filtering(with some code duplication) at simulation radio which is dedicate for simulation radio and **delete it from `Mac:Frame` class**. This way, we can save the code space in `Mac:Frame` class and does not increase the code size for `Mac` layer. How do you think? @abtink

abtink: This sounds good to me. Thanks @brianljt .

src/core/mac/mac.cpp
```diff
@@ -1590,12 +1587,18 @@ void Mac::HandleReceivedFrame(RxFrame *aFrame, otError aError)
...
 
     case Address::kTypeExtended:
-        IgnoreError(aFrame->GetDstPanId(panid));
-        VerifyOrExit(panid == mPanId && dstaddr.GetExtended() == GetExtAddress(),
-                     error = OT_ERROR_DESTINATION_ADDRESS_FILTERED);
+        VerifyOrExit(dstaddr.GetExtended() == GetExtAddress(), error = OT_ERROR_DESTINATION_ADDRESS_FILTERED);
         break;
     }
 
+    // Verify destination Pan ID if exist
```
bukepo, line 1594:
```suggestion
    // Verify destination PAN ID if present
```

jwhui:
> Could you add tests to cover the changes?

@brianljt , thoughts on this?

brianljt:
> > Could you add tests to cover the changes?
> 
> @brianljt , thoughts on this?

Updated. Commit ca9f578 remove the constrain on 2015 frame to forcefully use destination PAN ID so that this case should be covered in time sync. I also add commit 0a8d5dc to refine all the address field search related function in `Mac:Frame` class to be able to handle 2015 frame format. @jwhui @bukepo

examples/platforms/utils/mac_frame.cpp
```diff
@@ -41,29 +41,33 @@ bool otMacFrameDoesAddrMatch(const otRadioFrame *aFrame,
...
         break;
     }
 
+    if (frame.IsDstPanIdPresent())
+    {
+        Mac::PanId panid;
+
+        SuccessOrExit(frame.GetDstPanId(panid));
+        VerifyOrExit(panid == Mac::kPanIdBroadcast || panid == aPanId, OT_NOOP);
+    }
```
bukepo, line 67: avoid duplicate check? GetDstPanId() already checks frame.IsDstPanIdPresent().
```suggestion
    {
        Mac::PanId panid;

        VerifyOrExit(OT_ERROR_NONE != frame.GetDstPanId(panid) || panid == Mac::kPanIdBroadcast || panid == aPanId, OT_NOOP);
    }
```

src/core/mac/mac.cpp
```diff
@@ -1590,12 +1596,18 @@ void Mac::HandleReceivedFrame(RxFrame *aFrame, otError aError)
...
-                     error = OT_ERROR_DESTINATION_ADDRESS_FILTERED);
+        VerifyOrExit(dstaddr.GetExtended() == GetExtAddress(), error = OT_ERROR_DESTINATION_ADDRESS_FILTERED);
         break;
     }
 
+    // Verify destination PAN ID if present
+    if (aFrame->IsDstPanIdPresent())
+    {
+        PanId panid;
+        SuccessOrExit(aFrame->GetDstPanId(panid));
```
bukepo, line 1607: avoid duplicate check?

PR #5390, opened 2020-08-13
[netif] add helper methods to init 'NetifUnicastAddress'

abtink:
This commit adds a group of `Init` methods in `NetifUnciastAddress`
for common initializations of a unicast address with different
parameters (origin, scope, prefix length, and preferred/valid flags).

### DIFF
--- a/src/core/net/netif.cpp
+++ b/src/core/net/netif.cpp
@@ -495,5 +495,37 @@ bool Netif::IsUnicastAddressExternal(const NetifUnicastAddress &aAddress) const
     return mExtUnicastAddressPool.IsPoolEntry(aAddress);
 }
 
+void NetifUnicastAddress::InitAsThreadOrigin(void)
+{
+    Clear();
+    mPrefixLength  = NetworkPrefix::kLength;
+    mAddressOrigin = OT_ADDRESS_ORIGIN_THREAD;
+    mPreferred     = true;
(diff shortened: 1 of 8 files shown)


src/core/net/netif.cpp
```diff
@@ -495,5 +495,37 @@ bool Netif::IsUnicastAddressExternal(const NetifUnicastAddress &aAddress) const
...
+}
+
+void NetifUnicastAddress::InitAsSlaacOrigin(uint8_t aPrefixLength, bool aPreferred)
+{
+    Clear();
+    mPrefixLength  = aPrefixLength;
+    mAddressOrigin = OT_ADDRESS_ORIGIN_SLAAC;
+    mPreferred     = aPreferred;
+    mValid         = true;
+}
```
simonlingoogle, line 528: Can we use one method and pass `address origin`, `override` information as arguments?

abtink: I tried that. This would actually increase code size (and I feel it is not much nicer at the caller side than the existing code).
 Here is branch/commit if you are interested to see it: https://github.com/abtink/openthread/commit/400694f1e6c104bf6d9f50f01d67b21401d57a70

simonlingoogle: Thanks for clarification.

PR #5393, opened 2020-08-14
[security] verify received Enh-ACK integrity

brianljt:
With Thread 1.2, it is possible to receive an Enh-ACK with security enabled. This PR,

1. Move receive AES CCM related code to `Frame` class.
2. Verify received Enh-ACK in `Mac` layer. If MIC check failed, the transmitted frame will be marked as `OT_ERROR_NO_ACK`.


brianljt:
Another possible approach of handling Enh-ACK verification is at `Radio` layer. This requires frame extended source address to do AES CCM.

One way for `Radio` layer to get Enh-ACK extended source address is to get it from Enh-ACK source address or transmitted frame destination address. However, there are case when transmitted frame destination address and Enh-ACK source address are both not in extended format. So in order to do so, some change in `Mac` layer is needed to force to set destination address to extended format for ack required packet and frame version is 2015.

It is more nature to handle this in `Radio` layer as the retransmission/CSMA are all handled at `SubMac` or `Radio` layer. So that if one Enh-ACK has wrong MIC, `SubMac` or `Radio` can do retransmission without chaning its current logic.

I'm sending out this PR as a prototype to collect more feedback on this. @bukepo @abtink @jwhui

brianljt:
> Is it possible to add tests to cover these changes?

This can be verified once Li's CSL change is merged. So I did not add specific test cases to cover this but manually verify it.

src/core/mac/mac_frame.cpp
```diff
@@ -1155,6 +1155,54 @@ otError TxFrame::GenerateEnhAck(const RxFrame &aFrame, bool aIsFramePending, con
 }
 #endif // OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
 
+otError RxFrame::ProcessReceiveAesCcm(const ExtAddress &aExtAddress, const Key &aMacKey)
```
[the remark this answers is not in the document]
brianljt: The `RxFrame` will be modified in this method if the data payload is encrypted, so cannot be a `const` method.

src/core/mac/mac.cpp
```diff
@@ -1502,26 +1506,7 @@ otError Mac::ProcessReceiveSecurity(RxFrame &aFrame, const Address &aSrcAddr, Ne
...
-    // For fuzz tests, execute AES but do not alter the payload
-    uint8_t fuzz[OT_RADIO_FRAME_MAX_SIZE];
-    aesCcm.Payload(fuzz, aFrame.GetPayload(), aFrame.GetPayloadLength(), Crypto::AesCcm::kDecrypt);
-#endif
-    aesCcm.Finalize(tag);
-
-#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
-    VerifyOrExit(memcmp(tag, aFrame.GetFooter(), tagLength) == 0, OT_NOOP);
-#endif
+    VerifyOrExit(OT_ERROR_NONE == aFrame.ProcessReceiveAesCcm(*extAddress, *macKey), OT_NOOP);
```
bukepo, line 1509:
```suggestion
    SuccessOrExit(aFrame.ProcessReceiveAesCcm(*extAddress, *macKey));
```

src/core/mac/mac.cpp
```diff
@@ -1269,6 +1269,14 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
 
             mBroadcastTransmitCount = 0;
         }
+
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+        // Verify Enh-ACK integrity by checking its MIC
+        if ((aError == OT_ERROR_NONE) && (aAckFrame != nullptr))
+        {
+            aError = (ProcessEnhAckSecurity(aFrame, *aAckFrame) == OT_ERROR_NONE) ? OT_ERROR_NONE : OT_ERROR_NO_ACK;
```
Irving-cl, line 1277: Just have a question here: if the ack is not authenticated successfully(maybe there are attackers), shall we directly jump to the end or let upper layer handle this with a error code? I think there should be the right ACK soon. If we handle this TX here, then this TX has finished. If we got the right ACK later, we still got failure.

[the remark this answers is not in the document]
brianljt: If a malformed enh-ack is injected before a genuine enh-ack, the radio state machine will jump to some state like tx frame acked. So the following genuine enh-ack will just be ignored. Also, this case will not happen because ack sending does not go through CSMA/CA process. For example, A is the data frame transmitter, B is the intended genuine receiver, C is an intruder. When A sends out a data frame. B and C both receives this. B will send out genuine enh-ack after AIFS(192us) without doing any CCA. If C sends out a malformed ack, this ack will collide with B's ack and results in an NO_ACK for A.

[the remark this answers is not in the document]
Irving-cl: So if there'an intruder, then A couldn't get the genuine enh-ack, right?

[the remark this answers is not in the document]
brianljt: Yes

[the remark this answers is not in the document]
Irving-cl: Ok, make sense to me.

src/core/mac/mac.cpp
```diff
@@ -1369,6 +1369,14 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
...
+        }
```
abtink, line 1378: The code as is looks fine. A suggestion below to combine the check in `if`?
```suggestion
        if ((aError == OT_ERROR_NONE) && (aAckFrame != nullptr) && (ProcessEnhAckSecurity(aFrame, *aAckFrame) != OT_ERROR_NONE))
        {
            aError = OT_ERROR_NO_ACK;
        }
```

src/core/mac/mac.cpp
```diff
@@ -1664,6 +1649,97 @@ otError Mac::ProcessReceiveSecurity(RxFrame &aFrame, const Address &aSrcAddr, Ne
...
+    {
+        neighbor = Get<NeighborTable>().FindNeighbor(srcAddr);
+
+        if (neighbor != nullptr)
+        {
+            srcAddr.SetExtended(neighbor->GetExtAddress());
+        }
+    }
+
+    if (srcAddr.GetType() != Address::kTypeExtended)
```
abtink, line 1691: We can use helper methods for type check: `IsExtended()`, `IsShort()` and `IsNone()`:
```suggestion
    if (!srcAddr.IsExtended())
```

src/core/mac/mac.cpp
```diff
@@ -1664,6 +1649,97 @@ otError Mac::ProcessReceiveSecurity(RxFrame &aFrame, const Address &aSrcAddr, Ne
...
+    {
+        // Get Enh-ACK source address from transmitted frame destination address
+        IgnoreError(aTxFrame.GetDstAddr(srcAddr));
+
+        if (srcAddr.GetType() == Address::kTypeShort)
```
abtink, line 1696:
```suggestion
        if (srcAddr.IsShort())
```

src/core/mac/mac.cpp
```diff
@@ -1664,6 +1649,97 @@ otError Mac::ProcessReceiveSecurity(RxFrame &aFrame, const Address &aSrcAddr, Ne
...
+            neighbor = Get<NeighborTable>().FindNeighbor(srcAddr);
+
+            if (neighbor != nullptr)
+            {
+                srcAddr.SetExtended(neighbor->GetExtAddress());
+            }
+        }
+    }
+
+    VerifyOrExit(srcAddr.GetType() == Address::kTypeExtended, OT_NOOP);
```
abtink, line 1707:
```suggestion
    VerifyOrExit(srcAddr.IsExtended(), OT_NOOP);
```

src/core/mac/mac.cpp
```diff
@@ -1664,6 +1649,97 @@ otError Mac::ProcessReceiveSecurity(RxFrame &aFrame, const Address &aSrcAddr, Ne
...
+    VerifyOrExit(keyIdMode == Frame::kKeyIdMode1, error = OT_ERROR_NONE);
+
+    IgnoreError(aTxFrame.GetKeyId(txKeyId));
+    IgnoreError(aAckFrame.GetKeyId(ackKeyId));
+
+    VerifyOrExit(txKeyId == ackKeyId, OT_NOOP);
+
+    IgnoreError(aAckFrame.GetSrcAddr(srcAddr));
+
+    if (srcAddr.GetType() == Address::kTypeShort)
```
abtink, line 1681:
```suggestion
    if (srcAddr.IsShort())
```

examples/platforms/utils/mac_frame.h
```diff
@@ -161,6 +161,20 @@ uint8_t otMacFrameGetSequence(const otRadioFrame *aFrame);
...
+ * @param[in]  aFrame       A pointer to the MAC frame buffer that is received.
+ * @param[in]  aExtAddress  A pointer to the extended address, which will be used to generate nonce
+ *                          for AES CCM computation.
+ * @param[in]  aMacKey      A pointer to the MAC key to decrypt the received frame.
+ *
+ * @retval OT_ERROR_NONE      Process of received frame AES CCM succeeded.
+ * @retval OT_ERROR_SECURITY  Received frame MIC check failed.
+ *
+ */
+otError otMacFrameProcessReceiveAesCcm(otRadioFrame *aFrame, const otExtAddress *aExtAddress, const otMacKey *aMacKey);
```
jwhui, line 176: What is the expected use case for this utility function?

brianljt: Currently it is not used in Radio, but might need this later if we want to do decryption in Radio. I will remove this for now.

PR #5394, opened 2020-08-14
[posix] fix typo in error handling on macOS

wgtdkp:


src/posix/platform/uart.cpp
```diff
@@ -238,7 +238,7 @@ static void InitializeSessionSocket(void)
     // to simply ignore it.
 #if defined(SO_NOSIGPIPE)
     rval = 1;
-    VerifyOrExit((rval = setsockopt(sSessionSocket, SOL_SOCKET, SO_NOSIGPIPE, &rval, sizeof(rval)) != -1, OT_NOOP));
+    VerifyOrExit(rval = setsockopt(sSessionSocket, SOL_SOCKET, SO_NOSIGPIPE, &rval, sizeof(rval)) != -1, OT_NOOP);
```
simonlingoogle, line 241: Can the original code compile?

simonlingoogle, line 241:
```suggestion
    VerifyOrExit((rval = setsockopt(sSessionSocket, SOL_SOCKET, SO_NOSIGPIPE, &rval, sizeof(rval))) != -1, OT_NOOP);
```
use () to have explicit evaluation order?

wgtdkp: The error is revealed when I tried to compile otbr on my macOS catalina.

src/posix/platform/uart.cpp
```diff
@@ -238,7 +238,7 @@ static void InitializeSessionSocket(void)
     // to simply ignore it.
 #if defined(SO_NOSIGPIPE)
     rval = 1;
-    VerifyOrExit((rval = setsockopt(sSessionSocket, SOL_SOCKET, SO_NOSIGPIPE, &rval, sizeof(rval)) != -1, OT_NOOP));
+    VerifyOrExit((rval = setsockopt(sSessionSocket, SOL_SOCKET, SO_NOSIGPIPE, &rval, sizeof(rval))) != -1, OT_NOOP);
```
abtink, line 241: Nice find. This s tricky bug (do we not get compile error on this?) 👍 

How about breaking the statement into two lines?
- One line setting the `rval = setsockopt(...)` and 
- The other just a simple `VerifyOrExit(rval != -1, OT_NOOP)`.

I feel makes it more readable.

(also do we need the `rval = 1;` before this. Can we remove that as well?)

wgtdkp: @abtink 

> Nice find. This is a tricky bug (do we not get compile error on this?) +1

This actually will give a compiling error when building posix targets with `-DOT_DAEMON=ON`, added a build check for posix daemon.

>How about breaking the statement into two lines?
One line setting the rval = setsockopt(...) and
The other just a simple VerifyOrExit(rval != -1, OT_NOOP).

Changed, please review.

src/posix/platform/uart.cpp
```diff
@@ -237,8 +237,8 @@ static void InitializeSessionSocket(void)
     // if we have SO_NOSIGPIPE, then set it. Otherwise, we're going
     // to simply ignore it.
 #if defined(SO_NOSIGPIPE)
-    rval = 1;
-    VerifyOrExit((rval = setsockopt(sSessionSocket, SOL_SOCKET, SO_NOSIGPIPE, &rval, sizeof(rval)) != -1, OT_NOOP));
+    rval = setsockopt(sSessionSocket, SOL_SOCKET, SO_NOSIGPIPE, &rval, sizeof(rval));
+    VerifyOrExit(rval != -1, OT_NOOP);
```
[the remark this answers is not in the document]
wgtdkp: > Should we close(sSessionSocket) before sSessionSocket = -1 after exit: ?

I think yes and it needs to be refactored. Will do in a separate PR.

> Should we ignore this setsockopt(SO_NOSIGPIPE) (i.e. not set rval) even if it failed because it seems non-critical according to the comments above?

I think if `SO_NOSIGPIPE ` is defined, we expect `setsockopt(... SO_NOSIGPIPE ...)` to success. The comment means that we can just not set it if `SO_NOSIGPIPE ` is not defined.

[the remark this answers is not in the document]
wgtdkp: See https://github.com/openthread/openthread/pull/5405.

PR #5395, opened 2020-08-14
[pretty] shfmt only git files

simonlingoogle:
This PR enhanced `make-pretty` to only shfmt git files.

### DIFF
--- a/script/make-pretty
+++ b/script/make-pretty
@@ -133,7 +133,8 @@ do_shell_format()
     echo -e '     format shell'
     echo -e '====================='
 
-    shfmt -f . | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
+    local git_files=($(git ls-files))
+    shfmt -f "${git_files[@]}" | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
         | xargs -n10 -P"$OT_BUILD_JOBS" shfmt -i 4 -bn -ci -fn -s -w
 }
 
@@ -143,10 +144,11 @@ do_shell_check()
     echo -e '     check shell'
     echo -e '====================='
 
-    shfmt -f . | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
+    local git_files=($(git ls-files))
+    shfmt -f "${git_files[@]}" | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
         | xargs -n10 -P"$OT_BUILD_JOBS" shfmt -i 4 -bn -ci -fn -s -d
 
-    shfmt -f . | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
+    shfmt -f "${git_files[@]}" | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
         | xargs -n10 -P"$OT_BUILD_JOBS" shellcheck
 }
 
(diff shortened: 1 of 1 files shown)


simonlingoogle:
I realized that the actual issue is that `shfmt` does not use `git ls-files` to filter only repository files. But I think it is still valid to change the exclude patterns. 

An alternative fix would be use `git ls-files` together with `shfmt`. But the current fix seems even simpler. 

Thoughts? @jwhui

script/make-pretty
```diff
@@ -143,10 +143,10 @@ do_shell_check()
     echo -e '     check shell'
     echo -e '====================='
 
-    shfmt -f . | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
+    shfmt -f . | grep -v -E "^($(echo "${OT_EXCLUDE_PATTERNS[@]}" | tr ' ' '|'))" \
```
bukepo, line 146: I think using git ls-files is better which avoid changing untracked files.

simonlingoogle: Agree.

abtink:
@simonlingoogle this seems reasonable and a good enhancement to me. 👍 
I am just curious about the background on this. Is there a problem caused with formatting all matching source files?

script/make-pretty
```diff
@@ -133,7 +133,8 @@ do_shell_format()
     echo -e '     format shell'
     echo -e '====================='
 
-    shfmt -f . | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
+    local git_files=($(git ls-files))
+    shfmt -f "${git_files[@]}" | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
```
bukepo, line 137:
```suggestion
    git ls-files | xargs shfmt -f | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
```

script/make-pretty
```diff
@@ -143,10 +144,11 @@ do_shell_check()
     echo -e '     check shell'
     echo -e '====================='
 
-    shfmt -f . | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
+    local git_files=($(git ls-files))
+    shfmt -f "${git_files[@]}" | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
```
bukepo, line 148:
```suggestion
    git ls-files | xargs shfmt -f | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
```

script/make-pretty
```diff
@@ -143,10 +144,11 @@ do_shell_check()
...
         | xargs -n10 -P"$OT_BUILD_JOBS" shfmt -i 4 -bn -ci -fn -s -d
 
-    shfmt -f . | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
+    shfmt -f "${git_files[@]}" | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
```
bukepo, line 151:
```suggestion
    git ls-files | xargs shfmt -f | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
```

script/make-pretty
simonlingoogle: Thought of it too. But not sure if it's a good idea to have 2 `xargs` commands

simonlingoogle:
> @simonlingoogle this seems reasonable and a good enhancement to me. 👍
> I am just curious about the background on this. Is there a problem caused with formatting all matching source files?

When executed locally on my workstation, `script/make-pretty check shell` gives false fail with massive output because it checked shell files in the build/ directory and the `configure` file. I think it's confusing for OT developers.

PR #5398, opened 2020-08-14
[scripts] remove 1.2-bbr build from script/test

simonlingoogle:
This PR simplifies 1.2 builds by removing 1.2-bbr builds from `script/test`. 
With this PR, `script/test` always build one Thread 1.2 version with `BACKBONE_ROUTER` feature. 
**Non-BBR devices simple do not issue `bbr enable` command.**

### DIFF
--- a/script/test
+++ b/script/test
@@ -65,7 +65,7 @@ build_simulation()
     fi
 
     if [[ ${version} == "1.2" ]]; then
-        options+=("-DOT_CSL_RECEIVER=ON")
+        options+=("-DOT_CSL_RECEIVER=ON" "-DOT_BACKBONE_ROUTER=ON")
     fi
 
     if [[ ${ot_extra_options[*]+x} ]]; then
@@ -73,14 +73,6 @@ build_simulation()
     fi
 
     OT_CMAKE_BUILD_DIR="${OT_BUILDDIR}/cmake/openthread-simulation-${version}" "${OT_SRCDIR}"/script/cmake-build simulation "${options[@]}"
-
-    if [[ ${version} == "1.2" ]]; then
-
-        options+=("-DOT_BACKBONE_ROUTER=ON")
-
-        OT_CMAKE_BUILD_DIR="${OT_BUILDDIR}/cmake/openthread-simulation-${version}-bbr" "${OT_SRCDIR}"/script/cmake-build simulation "${options[@]}"
-
-    fi
 }
 
 build_posix()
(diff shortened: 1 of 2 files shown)


librasungirl:
Firmware built with BACKBONE_ROUTER flag without `bbr enable` is indeed could used to replace Firmware built without BACKBONE_ROUTER flag, however they are indeed different firmwares for individual device types, I think per device type should be tested.

Note that we do have plan to add a new 1.2 FTD build without DUA/MLR feature itself and verify its proxy MLR/DUA.req functionality.

Apart from simplifying the test, do you have other more considerations?

simonlingoogle:
> Firmware built with BACKBONE_ROUTER flag without `bbr enable` is indeed could used to replace Firmware built without BACKBONE_ROUTER flag, however they are indeed different firmwares for individual device types, I think per device type should be tested.
> 
> Note that we do have plan to add a new 1.2 FTD build without DUA/MLR feature itself and verify its proxy MLR/DUA.req functionality.
> 
> Apart from simplifying the test, do you have other more considerations?

The major consideration is to speed up building for 1.2 tests. Previously we need to build 3 versions for 1.2 (i.e. 1.2, 1.2-bbr, 1.1) and this PR reduced it to 2 versions. 

I do realized that 1.2 BBR=1 firmware without `bbr enable` is different from 1.2 BBR=0 firmware. But they shouldn't have any functional difference  as long as BBR functionalities are implemented correctly. 

OT has tens of configurations. I think it's impossible for us to test all combination of configurations. For example, we now test firmwares with REFERENCE_DEVICE=1, but `Reference Device` firmware is different from non Reference Device firmware. I guess we just accept such risks.  

I think it's valuable to test 1.2 FTD build without DUA/MLR feature to verify its proxy MLR/DUA.req functionality, realizing the fact that the code is not throughly tested for now.  But this also highlight the need to reduce the versions to build so that we are not going to have to build too many version just for testing.

Thoughts? @librasungirl

librasungirl:
Agree that it is not possible to enumerate all combination and 1.2 BBR=1, 1.2 BBR=0, 1.1 might looks complex.
However if just for speeding up build time for 1.2 test, just curious: how long would a build cost? how does it become a concern?

simonlingoogle:
> Agree that it is not possible to enumerate all combination and 1.2 BBR=1, 1.2 BBR=0, 1.1 might looks complex.
> However if just for speeding up build time for 1.2 test, just curious: how long would a build cost? how does it become a concern?

It depends on the performance of the computer. Recently, since I usually use `Macbook Pro` for working rather than more powerful Linux workstation, the building took much longer time than before.

librasungirl:
I may prefer a bit to keep existing builds.

PR #5399, opened 2020-08-14
[mesh-forwarder] add new class 'ThreadLinkInfo'

abtink:
The `ThreadLinkInfo` mirrors the `otThreadLinkInfo` definition.
It is mostly used in `MeshForwarder` and `Mle` class. Here are
some related helper methods:
- New helper methods in `Ip6::MessageInfo` to get and set the
  link info as `ThreadLinkInfo`.
- Helper method `ThreadLinkInfo::SetFrom()` allows it to be set
  from a given `Mac::RxFrame`.
- Helper method `Message::SetLinkInfo()` allows message link info
  related fields (in message metadata) to be set/updated from a
  given `ThreadLinkInfo`.

### DIFF
--- a/src/core/common/message.hpp
+++ b/src/core/common/message.hpp
@@ -53,6 +53,8 @@
 
 namespace ot {
 
+class ThreadLinkInfo;
+
 /**
  * @addtogroup core-message
  *
@@ -797,6 +799,14 @@ public:
      *
      */
     const RssAverager &GetRssAverager(void) const { return GetMetadata().mRssAverager; }
+
+    /**
+     * This method sets the message's link info properties (PAN ID, link security, RSS) from a given `ThreadLinkInfo`.
+     *
+     * @param[in] aLinkInfo   The `ThreadLinkInfo` instance from which to set message's related properties.
+     *
+     */
+    void SetLinkInfo(const ThreadLinkInfo &aLinkInfo);
 
     /**
      * This static method updates a checksum.
--- a/src/core/net/socket.hpp
+++ b/src/core/net/socket.hpp
@@ -40,6 +40,9 @@
 #include "net/ip6_address.hpp"
 
 namespace ot {
+
+class ThreadLinkInfo;
+
 namespace Ip6 {
 
 /**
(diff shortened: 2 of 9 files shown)


src/core/net/socket.hpp
```diff
@@ -174,6 +177,22 @@ class MessageInfo : public otMessageInfo, public Clearable<MessageInfo>
      */
     void SetLinkInfo(const void *aLinkInfo) { mLinkInfo = aLinkInfo; }
 
+    /**
+     * This method returns a pointer to the Link Info as a `ThreadLinkInfo`.
+     *
+     * @returns A pointer to Thread Link Info.
+     *
+     */
+    const ThreadLinkInfo *GetThreadLinkInfo(void) const { return reinterpret_cast<const ThreadLinkInfo *>(mLinkInfo); }
```
bukepo, line 186: does static_cast works?

abtink: yes, I think it would work. In this case, since we are casting from ` void *` we can use `static_cast` as well. But we generally use `static_cast` for conversion between parent/sub-classes. Here we are really casting  a `void *` as a new pointer type, so I felt `reinterpret_cast` would be indicate that better.

src/core/common/message.hpp
```diff
@@ -798,6 +800,14 @@ class Message : public Buffer
      */
     const RssAverager &GetRssAverager(void) const { return GetMetadata().mRssAverager; }
 
+    /**
+     * This method sets the link info fields (PAN ID, link security, RSS) from a given `ThreadLinkInfo`.
+     *
+     * @param[in] aLinkInfo   A Thread Link Info.
```
simonlingoogle, line 806:
```suggestion
     * @param[in] aLinkInfo   The Thread Link info.
```
Not sure if we should use `Link` or `link`. But I think we should be consistent in Doxygen.

src/core/net/socket.hpp
simonlingoogle: I tend to use `reinterpret_cast` whenever applicable.

src/core/net/socket.hpp
```diff
@@ -174,6 +177,22 @@ class MessageInfo : public otMessageInfo, public Clearable<MessageInfo>
...
+
+    /**
+     * This method sets the pointer to the Thread Link Info.
+     *
+     * @param[in]  aLinkInfo  A pointer to the Thread Link Info.
```
simonlingoogle, line 191:
```suggestion
     * @param[in]  aLinkInfo  A pointer to the Thread Link info.
```

src/core/net/socket.hpp
```diff
@@ -174,6 +177,22 @@ class MessageInfo : public otMessageInfo, public Clearable<MessageInfo>
...
+    /**
+     * This method returns a pointer to the Link Info as a `ThreadLinkInfo`.
+     *
+     * @returns A pointer to Thread Link Info.
+     *
+     */
+    const ThreadLinkInfo *GetThreadLinkInfo(void) const { return reinterpret_cast<const ThreadLinkInfo *>(mLinkInfo); }
+
+    /**
+     * This method sets the pointer to the Thread Link Info.
```
simonlingoogle, line 189:
```suggestion
     * This method sets the pointer to the Thread Link info.
```

abtink:
Thanks @jwhui  for merging the fix in (just came in to update branch and found it already merged)...  :)

PR #5400, opened 2020-08-14
[test] make POSIX pty check stable

bukepo:
This commits enhance paring extended address from CLI, so that it will
not be affected by unexpected prompt or tty control characters.

### DIFF
--- a/script/check-posix-pty
+++ b/script/check-posix-pty
@@ -139,7 +139,7 @@ EOF
 
     netstat -an | grep -q 61631 || die 'TMF port is not available!'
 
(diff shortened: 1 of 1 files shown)


script/check-posix-pty
```diff
@@ -139,7 +139,7 @@ EOF
 
     netstat -an | grep -q 61631 || die 'TMF port is not available!'
 
-    extaddr=$(awk '/extaddr/{getline; print}' $OT_OUTPUT | tr -d '\r\n')
+    extaddr=$(grep -aoE '[0-9a-z]{16}' $OT_OUTPUT)
```
simonlingoogle, line 142: I think we need to find (and fix) the root cause. 
FYI: I tested it locally on my Ubuntu 18.04 PC and it didn't fail.

Issue #5402, opened 2020-08-15
[github-actions] ot-ftd-otns is failing


jwhui:
https://github.com/openthread/openthread/runs/987574386?check_suite_focus=true
https://github.com/openthread/openthread/runs/987569802?check_suite_focus=true
https://github.com/openthread/openthread/runs/987588745?check_suite_focus=true

PR #5403, opened 2020-08-15
[github-actions] fix otns test failure

simonlingoogle:
This PR fixes OTNS test fails reported by #5402 @jwhui 

The OTNS test of OpenThread CI is supposed to be a subset of OTNS CI tests. So, Openthread's OTNS test should never fail if the merged OTNS PRs has all passed OTNS CI.

Turnout in OTNS CI, a setup is used to further setup the Python3 env which seems to make the Python env more reliable. 
```yml
    - name: Set up Python 3.6
      uses: actions/setup-python@v1
      with:
        python-version: 3.6
```

Adding this step in the OpenThread's OTNS test job fixes the issue.

### DIFF
--- a/.github/workflows/simulation.yml
+++ b/.github/workflows/simulation.yml
@@ -76,10 +76,10 @@ jobs:
     - uses: actions/setup-go@v1
       with:
         go-version: '1.13'
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: Set up Python 3.6
       uses: actions/setup-python@v1
       with:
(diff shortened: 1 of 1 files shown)


ctan-g:
Yes, I think that step is required if any Python package installation is to be performed.

simonlingoogle:
> Yes, I think that step is required if any Python package installation is to be performed.

I think this step downloads a pre built python version that already has some package installed (e.g. python3-wheel).
It should also be possible not to use this step but to install each required library.

PR #5406, opened 2020-08-15
[joiner-advertisement] add Joiner Advertisement CLI command

wgtdkp:


include/openthread/thread.h
```diff
@@ -286,6 +286,18 @@ otError otThreadSetJoinerAdvertisement(otInstance *   aInstance,
...
+ * This method returns the Thread Joiner Advertisement.
+ *
+ * @param[in]   aInstance       A pointer to an OpenThread instance.
+ * @param[out]  aOui            The Vendor IEEE OUI value that will be included in the Joiner Advertisement.
+ * @param[out]  aAdvDataLength  The length of Advertisement data in bytes.
+ *
+ * @retval A pointer to the Advertisement data, or NULL if none is set.
+ *
+ */
+const uint8_t *otThreadGetJoinerAdvertisement(otInstance *aInstance, uint32_t *aOui, uint8_t *aAdvDataLength);
```
lanyuwen, line 299: Could you help explain what's the use case of these "Get" APIs and why do we need them?

lanyuwen, line 299: Could you help explain what's the use case of these "Get" APIs and why do we need them?

src/cli/cli_joiner.hpp
```diff
@@ -79,6 +79,7 @@ class Joiner
         otError (Joiner::*mCommand)(uint8_t aArgsLength, char *aArgs[]);
     };
 
+    otError ProcessAdvertisement(uint8_t aArgsLength, char *aArgs[]);
```
lanyuwen, line 82: Maybe I don't have the context of why we want to introduce these APIs, but please note that the Joiner Advertisement "Set" APIs are available even when the Joiner flag is not enabled. That means they are general APIs for any roles but not only MeshCOP Joiners. Do we want align the scenarios between "Get" and "Set" APIs?

src/cli/cli_joiner.cpp
```diff
@@ -43,10 +45,55 @@ namespace ot {
...
+            mInterpreter.OutputFormat("\r\n");
+        }
+    }
+    else
+    {
+        unsigned long oui;
+        uint8_t       advData[OT_JOINER_ADVDATA_MAX_LENGTH];
+        int           advDataLength;
+
+        VerifyOrExit(aArgsLength >= 3, error = OT_ERROR_INVALID_ARGS);
```
simonlingoogle, line 81:
```suggestion
        VerifyOrExit(aArgsLength == 3, error = OT_ERROR_INVALID_ARGS);
```

include/openthread/thread.h
wgtdkp: A `getter` usually comes with a `setter` (It is common to have `write-only` API, but is unusual to have `read-only` API). For debugging, the getter is helpful to validate that we have the advertisement data set as expected.

simonlingoogle: If only for debugging/testing purpose, we can put this in the `REFERENCE DEVICE` scope. 
Thoughts? @wgtdkp @librasungirl @lanyuwen

src/cli/cli_joiner.hpp
wgtdkp: >  the Joiner Advertisement "Set" APIs are available even when the Joiner flag is not enabled. That means they are general APIs for any roles but not only MeshCOP Joiners.

Do you mean we should not implement those CLI sub-commands under the `joiner` command? Do you have a recommendation of the CLI commands format?

> Do we want align the scenarios between "Get" and "Set" APIs?

Yes, they are setter-getter pair.

include/openthread/thread.h
wgtdkp: @simonlingoogle  I think the user of OpenThread have the needs to verify that their product have correct advertisement data set.

src/cli/cli_joiner.cpp
```diff
@@ -43,10 +45,55 @@ namespace ot {
...
+        uint8_t        advDatalength;
+        const uint8_t *advData;
+
+        advData = otThreadGetJoinerAdvertisement(mInterpreter.mInstance, &oui, &advDatalength);
+        if (advData != nullptr)
+        {
+            mInterpreter.OutputFormat("oui: 0x%06x\r\n", oui);
+            mInterpreter.OutputFormat("advData: ");
+            mInterpreter.OutputBytes(advData, advDatalength);
+            mInterpreter.OutputFormat("\r\n");
```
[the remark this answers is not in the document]
wgtdkp: The `Output` method requires an additional `aBufLength` arguments. it requires counting string length by the caller, so I choose to stick with `OutputFormat` as it is simpler.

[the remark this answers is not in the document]
simonlingoogle: Worth to write a new version of `Output(const char *s)` ?

include/openthread/thread.h
simonlingoogle: > @simonlingoogle I think the user of OpenThread have the needs of verifying that their products have correct advertisement data set.

We can have the API always available and CLI commands reference device only.

src/cli/cli_joiner.cpp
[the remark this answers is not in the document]
wgtdkp: Maybe, but better not in this PR. Because there are many other usages of `OutputFormat` with the same cause, and it is better to have they all using the new version of `Output` once at all. So I think it is better to do this in a separate PR,  Agree?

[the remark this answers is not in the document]
simonlingoogle: agree

include/openthread/thread.h
wgtdkp: @lanyuwen @simonlingoogle Decided to remove the setter as it seems not critical for now.

src/cli/cli_joiner.hpp
wgtdkp: @lanyuwen Changed to use a CLI command `advdata`. Please review.

src/cli/cli.cpp
```diff
@@ -714,6 +717,28 @@ void Interpreter::ProcessDua(uint8_t aArgsLength, char *aArgs[])
...
+    unsigned long oui;
+    uint8_t       advData[OT_JOINER_ADVDATA_MAX_LENGTH];
+    int           advDataLength;
+
+    VerifyOrExit(aArgsLength == 2, error = OT_ERROR_INVALID_ARGS);
+
+    SuccessOrExit(error = ParseUnsignedLong(aArgs[0], oui));
+    VerifyOrExit(oui <= std::numeric_limits<uint32_t>::max(), error = OT_ERROR_INVALID_ARGS);
+
+    advDataLength = Hex2Bin(aArgs[1], advData, sizeof(advData));
```
lanyuwen, line 732: How about just taking the input as text string for verification purpose? Or else we may need to take additional step to make sure the input is a legal hex string?

wgtdkp: > Or else we may need to take additional step to make sure the input is a legal hex string?

@lanyuwen The [Hex2Bin](https://github.com/openthread/openthread/blob/6886ff1e4c5dcf1bdc6ac7a067a81debff113938/src/cli/cli.cpp#L307) function already validated that `advData` must be a hex string (it also makes sure that the hex string does exceeds the length of the buffer). It returns `-1` in case of getting invalid arguments and we captured that error at just below.

lanyuwen: I see, then this looks good to me.

lanyuwen (approved):
Thanks for submitting this PR, LGTM 👍

simonlingoogle (approved):
LGTM :+1:

jwhui:
Closing stale PR.

PR #5387, opened 2020-08-12
[cli] remove multiple servers support

bukepo:

PR #5408, opened 2020-08-16
[cli] fix crash when full logs is on

simonlingoogle:
It is possible that log is printed before the Interpreter (UART or CONSOLE) object is created when `FULL LOGS` is enabled. In such case, OT crashes. 
This PR fixes this bug. 

This bug was introduced by #5387

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -251,6 +251,8 @@ const struct Command Interpreter::sCommands[] = {
     {"version", &Interpreter::ProcessVersion},
 };
 
+Interpreter *Interpreter::sInterpreter = nullptr;
+
 Interpreter::Interpreter(Instance *aInstance)
     : mUserCommands(nullptr)
     , mUserCommandsLength(0)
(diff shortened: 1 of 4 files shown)


src/cli/cli.cpp
```diff
@@ -4648,39 +4650,63 @@ int Interpreter::OutputFormatV(const char *aFormat, va_list aArguments)
...
+    VerifyOrExit(Interpreter::IsInitialized(), OT_NOOP);
+
     Interpreter::GetInterpreter().Output(aString, aLength);
+exit:
+    return;
 }
 
 extern "C" void otCliAppendResult(otError aError)
 {
+    VerifyOrExit(Interpreter::IsInitialized(), OT_NOOP);
```
bukepo, line 4692: This is only needed in `Output()`?

simonlingoogle: The function that crashed was `otCliPlatLogv` only. 

Maybe other functions do not need this checking. Your thoughts?

simonlingoogle:
> is it possible to add test to cover this situation?

We can have test with `full logs`. I think our current `thread-cert` and `expect` tests do not work under `full logs`. 
Maybe we can add a github action of OTNS test with `full logs`. 
Thoughts? @bukepo

src/cli/cli.cpp
simonlingoogle: Besides `otCliPlatLogv`, other `otCliXxxXxx` APIs are called by user so the user should be responsible for initializing CLI first. So it changed to use assert in `GetInterpreter`.
Please check @bukepo

bukepo:
> > is it possible to add test to cover this situation?
> 
> We can have test with `full logs` to catch such issue. I think our current `thread-cert` and `expect` tests do not work under `full logs`.
> Maybe we can add a github action of OTNS test with `full logs`.
> Thoughts? @bukepo

SGTM

PR #5413, opened 2020-08-17
[ip6] simplify `Ip6:Header` and misc. enhancements

abtink:
This commit contain a list of smaller enhancements in `Ip6` modules:
- Simplify `Ip6::Header` removing the PoD definition and providing
  `enum` constants for the byte offset to different fields in the header
  (replacing/removing static methods).
- Declare some of the `Ip6` methods as `private`.
- Remove error logs (not used/applicable anymore).
- Simplify checking of `destPort` in `ProcessReceiveCallback()`.
- Specify `int` size of `enum` definitions (declare internally-used
  `enum`s a `private`).
- Add unit test `TestIp6Header()`.

### DIFF
--- a/src/core/net/ip6.cpp
+++ b/src/core/net/ip6.cpp
@@ -458,11 +458,10 @@ void Ip6::EnqueueDatagram(Message &aMessage)
 
 otError Ip6::SendDatagram(Message &aMessage, MessageInfo &aMessageInfo, uint8_t aIpProto)
 {
-    otError                    error = OT_ERROR_NONE;
-    Header                     header;
-    uint16_t                   payloadLength = aMessage.GetLength();
-    uint16_t                   checksum;
-    const NetifUnicastAddress *source;
+    otError  error = OT_ERROR_NONE;
+    Header   header;
+    uint16_t payloadLength = aMessage.GetLength();
+    uint16_t checksum;
 
     header.Init();
     header.SetDscp(PriorityToDscp(aMessage.GetPriority()));
@@ -480,7 +479,9 @@ otError Ip6::SendDatagram(Message &aMessage, MessageInfo &aMessageInfo, uint8_t 
 
     if (aMessageInfo.GetSockAddr().IsUnspecified() || aMessageInfo.GetSockAddr().IsMulticast())
     {
-        VerifyOrExit((source = SelectSourceAddress(aMessageInfo)) != nullptr, error = OT_ERROR_INVALID_SOURCE_ADDRESS);
+        const NetifUnicastAddress *source = SelectSourceAddress(aMessageInfo);
+
+        VerifyOrExit(source != nullptr, error = OT_ERROR_INVALID_SOURCE_ADDRESS);
         header.SetSource(source->GetAddress());
     }
     else
@@ -497,7 +498,6 @@ otError Ip6::SendDatagram(Message &aMessage, MessageInfo &aMessageInfo, uint8_t 
 
     SuccessOrExit(error = aMessage.Prepend(&header, sizeof(header)));
 
-    // compute checksum
     checksum = ComputePseudoheaderChecksum(header.GetSource(), header.GetDestination(), payloadLength, aIpProto);
 
     switch (aIpProto)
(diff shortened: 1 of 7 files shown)


src/core/net/ip6.cpp
```diff
@@ -536,20 +536,17 @@ otError Ip6::SendDatagram(Message &aMessage, MessageInfo &aMessageInfo, uint8_t
         SuccessOrExit(error = AddTunneledMplOption(aMessage, header, aMessageInfo));
     }
```
abtink, line 538: Adding a note on the `diff` below:
- Instead of after `exit` checking `if (error == OT_ERROR_NONE)` and then doing the queuing/fragmentation, we do it before the `exit` label.
- In the code above, all jumps to `exit` indicate a failure and would set the `error`.

src/core/net/ip6.cpp
```diff
@@ -1058,42 +1057,36 @@ otError Ip6::ProcessReceiveCallback(const Message &    aMessage,
...
-            default:
-#if OPENTHREAD_FTD
-                if (udp.GetDestinationPort() == Get<MeshCoP::JoinerRouter>().GetJoinerUdpPort())
-                {
-                    ExitNow(error = OT_ERROR_NO_ROUTE);
-                }
+            }
 #endif
-                break;
+#if OPENTHREAD_FTD
```
abtink, line 1084: A quick not here:
I did not use `else if` in the code checking the `dstPort` values to cover the (unlikely by I think possible) situation that `GetJoinerUdpPort()` happen to match one of the ones above.

src/core/net/ip6.cpp
```diff
@@ -1103,26 +1096,18 @@ otError Ip6::ProcessReceiveCallback(const Message &    aMessage,
...
-    switch (error)
+    if (messageCopy == nullptr)
     {
-    case OT_ERROR_NO_BUFS:
         otLogWarnIp6("Failed to pass up message (len: %d) to host - out of message buffer.", aMessage.GetLength());
-        break;
-
-    case OT_ERROR_DROP:
-        otLogNoteIp6("Dropping message (len: %d) from local host since next hop is the host.", aMessage.GetLength());
-        break;
```
abtink, line 1120: Removed this log.
- It is not longer valid (the error returned in this situation is actually `OT_ERROR_NO_ROUTE` and not `DROP`)
- I think I added this log line in early development days of OT. This was helpful back then to detect a bunch of failure situations (e.g., when another node is offline and a message to it is being bounced back to host). We now have a lot more useful logs that provide more details in different situations, so I don't think this log adds any value.

src/core/net/ip6.cpp
```diff
@@ -1058,42 +1057,36 @@ otError Ip6::ProcessReceiveCallback(const Message &    aMessage,
...
                 }
-
-                break;
-
-#if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE == 0
-            case kCoapUdpPort:
-
+            }
+#if !OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
+            if (destPort == kCoapUdpPort)
```
jwhui, line 1075: Not using `else-if` here is also intentional?

src/core/net/ip6.cpp
```diff
@@ -1103,26 +1096,18 @@ otError Ip6::ProcessReceiveCallback(const Message &    aMessage,
...
         otLogWarnIp6("Failed to pass up message (len: %d) to host - out of message buffer.", aMessage.GetLength());
-        break;
-
-    case OT_ERROR_DROP:
-        otLogNoteIp6("Dropping message (len: %d) from local host since next hop is the host.", aMessage.GetLength());
-        break;
-
-    default:
-        break;
+        ExitNow();
```
jwhui, line 1104:
```suggestion
        ExitNow(error = OT_ERROR_NO_BUFS);
```

src/core/net/ip6.cpp
abtink: Not really, we can use `else` here. 
I just changed all of them to `if (dstPort == {something})` but we can keep the `else` here. Will change it.

abtink: Changed this in new pushed commit. Aslo combined the indented two `if`s into one single `if` check (for each case).

PR #5415, opened 2020-08-17
[message] add helper methods to get first/next data chunks

abtink:
The `Message` data is stored in a sequence of linked-listed `Buffer`
instances. When reading or writing data in the `Message`, the internal
implementation iterates over the `Buffer` list and updates the data in
chunks. This commit adds two helper methods `GetFirstChunk()` and
`GetNextChunk()` to get first/next data chunk (contiguous buffer)
corresponding to a given offset and length into the message. These
helper methods are then used to simplify `Read()`, `Write()`, and
`UpdateChecksum()` methods. The unit test `test_message` is also
updated to verify reading and writing data at different offsets and
lengths.

### DIFF
--- a/src/core/common/message.cpp
+++ b/src/core/common/message.cpp
@@ -429,16 +429,14 @@ void Message::RemoveHeader(uint16_t aLength)
     }
 }
 
-uint16_t Message::Read(uint16_t aOffset, uint16_t aLength, void *aBuf) const
+void Message::GetFirstChunk(uint16_t aOffset, uint16_t &aLength, Chunk &aChunk) const
 {
-    const Buffer *curBuffer;
-    uint16_t      bytesCopied = 0;
-    uint16_t      bytesToCopy;
-
-    if (aOffset >= GetLength())
-    {
-        ExitNow();
-    }
+    // This method gets the first message chunk (contiguous data
+    // buffer) corresponding to a given offset and length. On exit
+    // `aChunk` is updated such that `aChunk.GetData()` gives the
+    // pointer to the start of chunk and `aChunk.GetLength()` gives
+    // its length. The `aLength` is also decreased by the chunk
+    // length.
 
     if (aOffset + aLength >= GetLength())
     {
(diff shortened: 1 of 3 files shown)


src/core/common/message.cpp
```diff
@@ -447,138 +449,118 @@ uint16_t Message::Read(uint16_t aOffset, uint16_t aLength, void *aBuf) const
...
 
-    if (aOffset + aLength >= GetLength())
-    {
-        aLength = GetLength() - aOffset;
-    }
+    aChunkBuffer = aChunkBuffer->GetNextBuffer();
+    OT_ASSERT(aChunkBuffer != nullptr);
 
-    aOffset += GetReserved();
+    aChunkData   = aChunkBuffer->GetData();
```
Irving-cl, line 509: I think in `GetNextChunk`, the outcome `aChunkData` actually always points to the start of `aChunkBuffer`. This function  doesn't actually depend on `aChunkData` of the previous chunk. Maybe we can remove this parameter.

src/core/common/message.cpp
```diff
@@ -447,138 +449,118 @@ uint16_t Message::Read(uint16_t aOffset, uint16_t aLength, void *aBuf) const
...
-    while (aOffset >= kBufferDataSize)
+    while (chunkLength > 0)
     {
-        OT_ASSERT(curBuffer != nullptr);
-
-        curBuffer = curBuffer->GetNextBuffer();
-        aOffset -= kBufferDataSize;
+        memcpy(bufPtr, chunkData, chunkLength);
+        bufPtr += chunkLength;
+        GetNextChunk(aLength, chunkBuffer, chunkData, chunkLength);
```
Irving-cl, line 538: Following the previous comment, what about changing this section as:
```
    while (chunkLength > 0)
    {
        memcpy(bufPtr, chunkData, chunkLength);
        bufPtr += chunkLength;
        GetNextChunk(aLength, chunkBuffer, chunkLength);
        chunkData = chunkBuffer->GetData();
    }
```

   
```

src/core/common/message.cpp
abtink: The primary goal of adding the `GetFirst/NextChunk()` is to help encapsulate all the common code that is used by `Read()`, `Write()` and `UpdateChecksum()` into the same helper method (so we don't repeat same code). 

They all need to get the pointer to the start of the chunk data (from the related `Buffer`), so I think it makes sense for this to be part of the common helper method itself (than each of them doing the same).

Also keeping `aChunkData` as a parameter helps keep the behavior of `GetFirstChunk()` and `GetNextChunk()` to be similar.

Irving-cl: Make sense.

abtink:
New pushed commit defines a private `struct Chunk`  (containing `mData` as pointer to chunk data buffer, `mLength` as  its length, and `mBuffer` for the `Buffer` containing the chunk). This helps simply the `GetFirst/NextChunk()` parameters. For non-const version of these method a sub-class `WritableChunk` is defined and used.

PR #5417, opened 2020-08-18
[nrf528xx] add received ACK timestamp info to mRxInfo

brianljt:
To enable CSL, it is necessary to get received Enh-ACK timestamp so that the next sample time could be calculated with this timestamp and CSL phase in CSL IE.


bukepo (approved):
LGTM:+1:

Irving-cl (approved):
LGTM👍

LuDuda (approved):

PR #5421, opened 2020-08-19
[cmake] use OT_CONFIG for OPENTHREAD_CONFIG_FILE

bukepo:
OT_CONFIG was used to specify a customized OpenThread core config file
which is only included by OpenThread core library. This commit changes
the behavior so that it specifies the OPENTHREAD_CONFIG_FILE which is
used by all components of OpenThread including the core library.


Irving-cl (approved):
LGTM👍

Issue #5274, opened 2020-07-22
Enhance size-report to include 1.2 features


simonlingoogle:
Continuing discussion from https://github.com/openthread/openthread/pull/5248#discussion_r458561842 

> I was hoping to catch situations where adding Thread 1.2 code (intentionally or unintentionally) changes Thread 1.1 code. Should we flag situations where adding a Thread 1.2 feature significantly increases Thread 1.1 code?

Here are some examples by adding more 1.2 columns to `size-report`.  

### Example 1:

|  name  |  branch  | 1.1 text  | 1.1 data  | 1.1 bss  | 1.1 total | 1.2 text  | 1.2 data  | 1.2 bss  | 1.2 total |
| :----: | :------: | -----: | ----: | ---: | ----: |  ----: | ----: | ----: | ----: |
| libopenthread-ncp-ftd.a | master | 396720 | 736 | 72828 | 470284 | 396720 | 736 | 72828 | 470284 |
|  | #5268 | 396720 | 736 | 72828 | 470284 | 396720 | 736 | 72828 | 470284 |
|  | +/- | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 |
| libopenthread-ncp-mtd.a| master | 319360 | 736 | 64180 | 384276 | 319360 | 736 | 64180 | 384276 |
|  | #5268 | 319360 | 736 | 64180 | 384276 | 319360 | 736 | 64180 | 384276 |
|  | +/- | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 |

### Example 2 (Only add 1.1/1.2 for total column):
|  name  |  branch  | text  | data  | bss  | 1.1 total | text  | data  | bss  | 1.2 total |
| :----: | :------: | -----: | ----: | ---: | ----: |  ----: | ----: | ----: | ----: |
| libopenthread-ncp-ftd.a | master | 396720 | 736 | 72828 | 470284 | 396720 | 736 | 72828 | 470284 |
|  | #5268 | 396720 | 736 | 72828 | 470284 | 396720 | 736 | 72828 | 470284 |
|  | +/- | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 |
| libopenthread-ncp-mtd.a | master | 319360 | 736 | 64180 | 384276 | 319360 | 736 | 64180 | 384276 |
|  | #5268 | 319360 | 736 | 64180 | 384276 | 319360 | 736 | 64180 | 384276 |
[...]

PR #5424, opened 2020-08-19
[size-report] add size-report for Thread 1.2

simonlingoogle:
This PR adds size-report for Thread 1.2 features in the simplest way that **does not require any change to [openthread-size-report](https://glitch.com/edit/#!/openthread-size-report) on Glitch**. 

Size reports for both Thread 1.1 and 1.2 are posted in the same comment with 1.2 report below 1.1 report.

This PR does not fully address style related comments in #5274, which we can leave to future enhancements.

**QUESTION: should we put 1.2 report above 1.1 report instead?**

### DIFF
--- a/script/check-size
+++ b/script/check-size
@@ -98,7 +98,7 @@ markdown()
     esac
 }
 
-size_nrf52840()
+size_nrf52840_version()
 {
     local flags=(
         "BORDER_AGENT=1"
@@ -128,6 +128,18 @@ size_nrf52840()
         "UDP_FORWARD=1"
     )
 
+    local thread_version=$1
+
+    if [[ ${thread_version} == "1.2" ]]; then
+        flags+=(
+            "THREAD_VERSION=1.2"
+            "BACKBONE_ROUTER=1"
+            "DUA=1"
+            "MLR=1"
+            # "CSL_RECEIVER=1" not supported on nrf52840 yet
+        )
+    fi
+
     rm -rf "${OT_TMP_DIR}"
 
     # new commit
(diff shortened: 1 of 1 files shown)


script/check-size
```diff
@@ -149,6 +161,39 @@ size_nrf52840()
         && ./bootstrap \
         && make -f examples/Makefile-nrf52840 "${flags[@]}")
 
+    # rename the generated files to be ready for size-report
+    # shellcheck disable=SC2011
```
simonlingoogle, line 165: **clarification for the shellcheck disable:**
shellcheck suggested to replace `ls` with `find .. -print0 | ...` so as to handle file names with special characters correctly. 
But I think it's fine here. BTW, it seems quite difficult to write the command in the `find .. -print0 | ...` way.

simonlingoogle:
Hi all @bukepo @abtink @jwhui @librasungirl , 

Any comments on this PR. 
It would be really help to be able to see the code and ram size costs of 1.2 PRs

Thanks,

librasungirl:
`libopenthread-cli-ftd.a_1.2`,  such naming with the version suffix seems to me a bit change.
I may prefer to add a separate `version` column a bit.

simonlingoogle:
> `libopenthread-cli-ftd.a_1.2`, such naming with the version suffix seems to me a bit change.
> I may prefer to add a separate `version` column a bit.

Yes, it's a little bit strange. 
But we can't have a separate `version` column without enhancing the [openthread-size-report](https://glitch.com/edit/#!/openthread-size-report) bot. 
This PR however tries to add the Thread 1.2 size report with only OpenThread local changes.
I prefer to enhance the size report to follow whatever style we are going to agree in #5274 in a future PR. 

Thoughts?

librasungirl (approved):

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #5428, opened 2020-08-20
[low-power] fix CSL_RECEIVER build option

Irving-cl:
This PR removes the config setting `OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE`automatically when it's 1.2 and MTD. The reason is that it's in core config and it's also used in platform(radio) which cannot access core config. 

If `CSL_RECEIVER` is not set in options in cmake and it's 1.2, then in core config, `OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE` would be turned on if it's MTD. While in platform, it's not turned on. And then there would be a link issue.


src/core/config/mac.h
```diff
@@ -314,18 +314,6 @@
...
- * This setting configures the CSL receiver feature in Thread 1.2.
- * This is compulsory for 1.2 MTD, optional for 1.2 FTD.
- *
- */
-#ifndef OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
-#define OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE \
-    (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2) && OPENTHREAD_MTD
-#endif
-
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
```
simonlingoogle, line 317: seems to have many warnings?

simonlingoogle (commented):
We should catch the error with compile errors rather than leave it to link error?

src/core/config/mac.h
Irving-cl: Has defined it as 0 to avoid warnings.

src/core/config/mac.h
```diff
@@ -318,12 +318,12 @@
  * @def OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
  *
  * This setting configures the CSL receiver feature in Thread 1.2.
- * This is compulsory for 1.2 MTD, optional for 1.2 FTD.
+ * Set it to 0 by default because it is also used in 'platform' which cannot access the core config. To enable this
+ * option, it has to be set in build options when building.
```
simonlingoogle, line 322:
```suggestion
```
I think users already know they need to specify build options when building.

simonlingoogle (approved):
LGTM :+1:

bukepo (approved):
LGTM:+1:

src/core/config/mac.h
Irving-cl: Make sense. Has removed it.

librasungirl (approved):

PR #5432, opened 2020-08-20
[low-power] keep configured CSL channel for child

edmont:
A SSED uses the Child Update Request to start CSL synchronization. If the CSL Channel TLV is not present in the message, the parent assigns the Thread Network channel as the CSL channel for this child during the attachment.

However, it may happen that a child changes its CSL channel to a different one with a Child Update Request and decides to keep using it. In a subsequent Child Update Request it could omit the CSL Channel TLV (no changes) but the parent would change its CSL channel to the Thread Network channel. This PR avoids that.

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -2540,7 +2540,7 @@ void MleRouter::HandleChildUpdateRequest(const Message &         aMessage,
         {
             child->SetCslChannel(static_cast<uint8_t>(cslChannel.GetChannel()));
         }
-        else
+        else if (child->GetCslChannel() == 0)
         {
             child->SetCslChannel(Get<Mac::Mac>().GetPanChannel());
         }


src/core/thread/mle_router.cpp
```diff
@@ -2540,7 +2540,7 @@ void MleRouter::HandleChildUpdateRequest(const Message &         aMessage,
         {
             child->SetCslChannel(static_cast<uint8_t>(cslChannel.GetChannel()));
         }
-        else
+        else if (child->GetCslChannel() == 0)
```
Irving-cl, line 2543: I think it's allowed for a child to change and update its CSL channel. I didn't find descriptions in spec to forbid so. What's the reason doing so?

Irving-cl (approved):
LGTM

edmont:
BTW, currently OT SSED always appends CSL TLVs to the Child Update Request, but it could omit them if they didn't change since the last synchronization:

https://github.com/openthread/openthread/blob/master/src/core/thread/mle.cpp#L2248-L2254

src/core/thread/mle_router.cpp
zhanglongxia, line 2543: According to the 802.15.4-2015 spec, channel 0 is a valid channel. The frequency of the channel 0 is 868.3MHz. If someone implements the channel 0 in the future, there will be a problem here.

Irving-cl: Here `child->GetCslChannel() == 0` means it's unset. I think we define a invalid value instead of 0 to replace it later.

zhanglongxia: It sounds reasonable.

zhanglongxia (approved):

PR #5433, opened 2020-08-20
[cmake] support setting log level

bukepo:


script/check-simulation-build-cmake
```diff
@@ -41,6 +41,8 @@ build_all_features()
     reset_source
     "$(dirname "$0")"/cmake-build simulation
     reset_source
+    "$(dirname "$0")"/cmake-build simulation -DOT_LOG_LEVEL=INFO
+    reset_source
```
simonlingoogle, line 45: `LOG_LEVEL=INFO` build test seems not that necessary? Note that each test we add makes our GitHub action slower. 

I think it's better to have LOG_LEVEL=NONE test and LOG_LEVEL=DEBG test to cover most cases (not sure if we already have them).

bukepo: I think this only checks building.

simonlingoogle: Currently, build tests are the slowest github actions we have. 
That's why I think we should add most representative building tests (i.e. full log building and none log building), but need not to add intermediate building tests. 

Thoughs?

bukepo: The new test is only for validating the cmake option OT_LOG_LEVEL works, nothing beyond this purpose.

simonlingoogle (approved):
lgtm, thanks

Issue #5414, opened 2020-08-17
[message] potential issue with checksum calculation


abtink:
I think we may have a potential issue with checksum calculation.

The checksum uses `uint16` values as it updates the value (padding zero at the end if total number of bytes is odd).
We have different flavors of `Message::UpdateChecksum()`

1. A static method `UpdatChecksum(value) adding `uint16_t` value.
1. A static method `UpdateChecksum(void *aBuffer, uint16_t aLength)` adding an array of bytes (given as buffer and length)
1. A `Message` method `UpdateChecksum(uint16_t offset, uint16_t aLength)` adding content of `message` from given offset up to specified length. 

The problem is with the third method. 
- The `Message` content is internally stored in chunks of data in a sequence of linked-listed allocated `Buffer` instances. 
  - The first `Buffer` stores the `Metada` as well, so it can store smaller chunk of the message data.
- The `UpdateCheckum()` implementation goes through all data chunks from different `Buffer` instances and updates the checksum for each data chunk (calling second flavor of `UpdateChecksum` on each data chunk).
- The issue arises if any of the data chunks happen to have an odd length, then when we add the next chunk its first byte will be treated as new `uint16_t` rather than MSB of the previous `uint16_t`.

I don't think this is critical. I think it all current place where we calculate checksum the offsets are  even and also the metda data length (which impact the data in first `Buffer`) is also even.

jwhui:
@abtink , thanks for flagging this issue. As you mentioned, the current implementation assumes that each check has even offset and even length. Maybe it would be sufficient to add some static checks in rather than updating the implementation to support odd offset/length?

abtink:
@jwhui , I have a refactoring and simplification of all checksum related method in the works (adding a `Checksum` class/module that provides all related functionality). It would also address this.

The challenge I see is if something is added to message metadata to make it odd sized, we may face an issue (the first chunk can be of odd length).  Also buffer size itself can be customized by project config (and can be odd value), so I think it may be better be safe in all cases.

jwhui:
> The challenge I see is if something is added to message metadata to make it odd sized, we may face an issue (the first chunk can be of odd length). Also buffer size itself can be customized by project config (and can be odd value), so I think it may be better to be safe in all cases.

I think the offset would generally be OK - the structures are not declared as packed, so I assume there would be padding due to word alignment.

As for buffer size itself, I think we could have a compile-time check to ensure it is not odd. That way we do not incur and code overhead for dealing with the odd size case.

[...]

PR #5434, opened 2020-08-20
[net] add 'Checksum' class

abtink:
This commit adds `Checksum` class which implements IP checksum
calculation and verification (all related code from other modules is
now moved to `Checksum` class). The new checksum calculation allows
any message chunk size (removes a limitation with previous model
of requiring even sized chunks). This commit also adds a unit test
`test_checksum` covering the behavior of `Checksum` methods.

----

This should help address Issue https://github.com/openthread/openthread/issues/5414.

### DIFF
--- a/src/core/net/udp6.cpp
+++ b/src/core/net/udp6.cpp
@@ -41,9 +41,8 @@
 #include "common/encoding.hpp"
 #include "common/instance.hpp"
 #include "common/locator-getters.hpp"
+#include "net/checksum.hpp"
 #include "net/ip6.hpp"
-
-using ot::Encoding::BigEndian::HostSwap16;
 
 namespace ot {
 namespace Ip6 {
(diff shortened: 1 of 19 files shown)


src/core/net/udp6.cpp
```diff
@@ -353,27 +352,16 @@ otError Udp::SendDatagram(Message &aMessage, MessageInfo &aMessageInfo, uint8_t
...
 {
-    otError  error = OT_ERROR_NONE;
-    Header   udpHeader;
-    uint16_t payloadLength;
-    uint16_t checksum;
-
-    payloadLength = aMessage.GetLength() - aMessage.GetOffset();
-
-    // check length
-    VerifyOrExit(payloadLength >= sizeof(Header), error = OT_ERROR_PARSE);
```
abtink, line 364: Adding a note about `diff` here:
The new code performs the reading of `udpHeader` from message first before checksum check. 
With this we can then remove the check `VerifyOrExit(payloadLength >= sizeof(Header)` (it would be covered by header read).

simonlingoogle (approved):
LGTM :+1:

tests/unit/test_checksum.cpp
```diff
@@ -0,0 +1,280 @@
...
+        Checksum::UpdateMessageChecksum(*message, messageInfo.GetSockAddr(), messageInfo.GetPeerAddr(),
+                                        Ip6::kProtoIcmp6);
+
+        VerifyOrQuit(message->Read(message->GetOffset(), sizeof(icmp6Header), &icmp6Header) == sizeof(icmp6Header),
+                     "Message::Read() failed");
+        VerifyOrQuit(icmp6Header.GetChecksum() != 0, "Failed to update checksum");
+
+        // Verify that the calculated ICMP6 checksum is valid.
+
+        VerifyOrQuit(CalculateChecksum(messageInfo.GetSockAddr(), messageInfo.GetPeerAddr(), Ip6::kProtoIcmp6,
```
jwhui, line 237: Can we add some negative test cases to validate that the checksum catches corrupt data?

abtink: Good idea. Will add it in next push.

abtink: Added in new pushed commit.

PR #5435, opened 2020-08-20
[neighbor-table] add 'StateFilter' parameter in 'Find' methods

abtink:
---

The filtering is not currently used in any of existing code, however it adds flexibility without adding much of (code size or run-time) overhead. 

It is a functionality that is needed/used by the multi-link (TREL) PR (which earlier added it only when feature is enabled, but I think it make sense to add the filtering in general case).

### DIFF
--- a/src/core/thread/neighbor_table.hpp
+++ b/src/core/thread/neighbor_table.hpp
@@ -85,63 +85,75 @@ public:
      * address.
      *
      * @param[in]  aShortAddress  A short address.
+     * @param[in]  aFilter        A neighbor state filter
      *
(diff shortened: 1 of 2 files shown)


src/core/thread/neighbor_table.hpp
```diff
@@ -85,75 +85,89 @@ class NeighborTable : public InstanceLocator
      * address.
      *
      * @param[in]  aShortAddress  A short address.
+     * @param[in]  aFilter        A neighbor state filter
```
simonlingoogle, line 88:
```suggestion
     * @param[in]  aFilter        A Neighbor state filter
```
I think `Neighbor` is a Thread network concept. 

Same issue for many other places.

PR #5436, opened 2020-08-20
Fix OpenThread GN build with -Wundef

mspang:
Always define all of OPENTHREAD_{FTD,MTD,RADIO} and add a define for
OPENTHREAD_POSIX.

### DIFF
--- a/etc/gn/openthread.gni
+++ b/etc/gn/openthread.gni
@@ -55,6 +55,9 @@ declare_args() {
   # Use external mbedtls. If blank, internal mbedtls will be used.
   openthread_external_mbedtls = ""
 
+  # Enable building for POSIX platforms.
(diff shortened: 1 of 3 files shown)


etc/gn/openthread.gni
```diff
@@ -55,6 +55,9 @@ declare_args() {
   # Use external mbedtls. If blank, internal mbedtls will be used.
   openthread_external_mbedtls = ""
 
+  # Enable building for POSIX platforms.
+  openthread_posix = (current_os == "linux" || current_os == "fuchsia")
```
bukepo, line 59: Is fuchsia POSIX?

mspang: It isn't although it is building as if it were Android for OpenThread. I think you are right that it's safer to leave this off there, though, removed it.

PR #5437, opened 2020-08-21
[github-action] fail on Codedov uploading errors

Moandor-y:
This makes the CI pipeline fail when Codedov uploading runs into errors.

PR #5441, opened 2020-08-22
[message] change 'Write()' to be a 'void' method

abtink:
This commit changes `Message::Write()` to return `void` instead of
returning the number of bytes written. The `Write()` method does not
resize the message and requires that the given data can be fully
written into the message within its existing length. The `Write()`
already contains an `OT_ASSERT()` to verify this.

This commit also removes the redundant `OT_ASSERT()` checks in few
places where `Write()` method's returned value was being used.


Moandor-y:
> I wonder why code coverage dropped by 1.25% in this PR. Apparently this has nothing to do with this PR.
> @abtink @Moandor-y

It seems after merging https://github.com/openthread/openthread/pull/5437 everything has a significant drop. Both this PR and https://github.com/openthread/openthread/pull/5440 show +1k lines of code. I think this is because those codes are not compiled by GCC, so they weren't uploaded before, but now we changed from `gcov` to `llvm-cov` and they are reported properly.

PR #5442, opened 2020-08-23
[thread-cert] refactor case 9.2.12 and 9.2.13 using pktverify

FiveDimensions:


tests/scripts/thread-cert/Cert_9_2_13_EnergyScan.py
```diff
@@ -105,6 +112,24 @@ def test(self):
...
+    def verify(self, pv):
+        pkts = pv.pkts
+        pv.summary.show()
+
+        ED = pv.vars['ED']
+        COMMISSIONER = pv.vars['COMMISSIONER']
+        _ed_pkts = pkts.filter_wpan_src64(ED)
+
+        # Step 3: ED MUST send MGMT_ED_REPORT.ans to the Commissioner and report energy measurements
+        _ed_pkts.filter_coap_request(MGMT_ED_REPORT).must_next()
```
simonlingoogle, line 124: should check for the destination?

tests/scripts/thread-cert/Cert_9_2_13_EnergyScan.py
```diff
@@ -105,6 +112,24 @@ def test(self):
...
+
+        # Step 5: ED MUST send MGMT_ED_REPORT.ans to the Commissioner and report energy measurements
+        _ed_pkts.filter_coap_request(MGMT_ED_REPORT).must_next()
```
simonlingoogle, line 127: should check for the destination?

tests/scripts/thread-cert/Cert_9_2_13_EnergyScan.py
FiveDimensions: added

tests/scripts/thread-cert/Cert_9_2_13_EnergyScan.py
FiveDimensions: added

tests/scripts/thread-cert/Cert_9_2_12_Announce.py
```diff
@@ -129,14 +120,56 @@ def test(self):
...
+        pkts.filter_wpan_src64(MED).filter_wpan_dst64(LEADER_2).filter_mle_cmd(
+            MLE_CHILD_ID_REQUEST).must_next().must_verify(lambda p: p.wpan.dst_pan == DATASET1_PANID)
+
+        pkts.filter_wpan_src64(MED).filter_ipv6_dst(LINK_LOCAL_ALL_NODES_MULTICAST_ADDRESS).filter_mle_cmd(
+            MLE_ANNOUNCE).must_next().must_verify(
+                lambda p: {CHANNEL_TLV, PAN_ID_TLV, ACTIVE_TIMESTAMP_TLV} == set(p.mle.tlv.type) and p.wpan.dst_pan ==
+                0xffff and p.wpan.aux_sec.key_id_mode == 0x2 and p.wpan.aux_sec.key_source == 0x00000000ffffffff)
+
+        # Step 6: MED MUST respond with an ICMPv6 Echo Reply
+        pkts.filter_ping_reply().filter_ipv6_src_dst(MED_RLOC, LEADER_2_RLOC)
```
simonlingoogle, line 171:
```suggestion
        pkts.filter_ping_reply().filter_ipv6_src_dst(MED_RLOC, LEADER_2_RLOC).must_next()
```

tests/scripts/thread-cert/Cert_9_2_13_EnergyScan.py
```diff
@@ -42,23 +44,27 @@ class Cert_9_2_13_EnergyScan(thread_cert.TestCase):
...
         },
         ROUTER1: {
+            'name': 'ROUTER',
             'mode': 'rsdn',
             'panid': 0xface,
             'router_selection_jitter': 1,
             'allowlist': [LEADER, ED1]
         },
         ED1: {
+            'name': 'ED',
```
simonlingoogle, line 67: In this case, the ED DUT can be MED or SED. How do we verify both role within one case?

FiveDimensions: There is no special packet to verify for MED/SED in the test plan

simonlingoogle: But we need to verify that both MED/SED works in this test case.
The simplest method is to have two almost identical test cases, with different mods for the ED device. But this brings too much redundant code. 

A better way I can think of is to define a base test class and create two subclasses (in two files) one for MED DUT and another for SED DUT. Each subclass define it's own topology with correct modes for SED/MED. 

Thoughts? @FiveDimensions @librasungirl

FiveDimensions: Currently, there are several cases with different DUT scenarios, implement this as a example to refactor the others

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #5425, opened 2020-08-19
[scripts] add 1.2 build test without additional features

simonlingoogle:
This PR adds a Thread 1.2 build test without any additional feature (i.e. DUA/MLR/CSL). 
- [x] Fix 1.2 build error with DUA=0

~Depends on #5428~ Merged

PR #5445, opened 2020-08-24
[dua] fix debg output

Irving-cl:
`mRegistrationDelay` only exists when `OPENTHREAD_CONFIG_DUA_ENABLE` is set. If `OPENTHREAD_CONFIG_DUA_ENABLE` is not set and log level is DEBUG, there would be an error.


Irving-cl:
Fixed in #5424

simonlingoogle:
#5425 is already merged.

PR #4557, opened 2020-02-13
[low-power] add csl feature for Thread 1.2

Irving-cl:
This PR implements the CSL feature in Thread 1.2.
It is based on enhanced keep alive(#4325) and enhanced frame pending(#4334).
In more detail, this PR does these things:
*  Add macro definitions for low power to control the compiling of source code
*  Add data and methods for running CSL in `Mac` and `SubMac`. This mainly includes setting CSL parameters, starting/stopping CSL, and the timer handling process.
*  Add otPlatTimeGetAPI and implementation.
*  Add CSL transmission implementation. CSL transmission is a new kind of transmission, the related definition and implementaion for the whole transmitting process is added.
*  Add calling of start/stop CSL in certain cases.
*  Implement CSL synchronization maintainence. If a CSL cordinator didn't get a frame containing CSL IE for `CSLTimeout`, the CSL receiver is regarded as de-synchronized.
*  Add Cli interface for using CSL.
*  Implement enhanced Ack with IE. The original code can only generate auto ack for Imm-Ack. As CSL requires CSL IE included in enhanced ack. This PR implements it.
*  Add basic functional test for CSL transmission. More tests corresponding to test plan would be added later.

The above items are corresponded to the commits in this PR.

PR #5449, opened 2020-08-24
[low-power] add CSL implementations on nrf52840

Irving-cl:
This PR adds CSL support on nrf52840. I have done some simple verifications on nrf52840. 

## Issue
However we've got an issue now: according to a couple times of tests, currently the `OPENTHREAD_CONFIG_CSL_SAMPLE_WINDOW` is set to **40**(6.4ms). Otherwise, the CSL transmission wouldn't succeed(Less than 40, for example, 30). It's much bigger than we have expected. And it's quite difficult to analyze the timeline for two devices together(I cannot tell why doesn't the window perfectly matches the csl transmission). I'm planning to add a feature that uses GPIO in two devices to sync their tx/rx state to another device(a oscillometer or a development board) so that we can see the internal states for two devices together. I think it would be helpful for our later enhancement and debug for CSL feature. I would discuss and track this in another issue.

## Simple verification
First build it on nrf52840 with CSL features(For dongle):
```
make -f examples/Makefile-nrf52840 THREAD_VERSION=1.2 CSL_RECEIVER=1 BOOTLOADER=USB USB=1
```
To run a simple test, we can use exp scripts:
**csl-leader.exp**:
```
#!/usr/bin/expect -f

set dev_name [lindex $argv 0];

puts "dev_name: $dev_name"

spawn screen -L -Logfile leader.log $dev_name

send "ifconfig down\n"
expect "Done"
send "factoryreset\n"
sleep 1
send "\n"
send "ifconfig up\n"
expect "Done"
send "panid 0x1234\n"
expect "Done"
send "channel 23\n"
expect "Done"
send "thread start\n"
expect "Done"

interact
```
**csl-ssed.exp**:
```
#!/usr/bin/expect -f

set dev_name [lindex $argv 0];

puts "dev_name: $dev_name"

spawn screen -L -Logfile ssed.log $dev_name

send "ifconfig down\n"
expect "Done"
send "factoryreset\n"
sleep 1
send "ifconfig up\n"
expect "Done"
send "panid 0x1234\n"
expect "Done"
send "channel 23\n"
expect "Done"
send "mode s\n"
expect "Done"
send "csl period 3125\n"
expect "Done"
send "csl channel 23\n"
expect "Done"
send "thread start\n"
expect "Done"

interact
```

Run `csl-leader.exp` first with a parameter for device name, for example:
```
./csl-leader.exp /dev/ttyACM0
```
Then run `csl-ssed.exp`:
```
./csl-ssed.exp /dev/ttyACM1
```
Then we can do some ping tests.


Irving-cl:
Upd:
After a discussion, we think it would be simpler to just add a debug option for CSL. When the option, the CSL receiver would keep receiving when it should sleep so that it could compare the actual receiving time and its sample window time.

edmont:
Hi @Irving-cl,

This implementation uses the `alarm.c` timer which is not so precise due to latency problems, since it depends on other processes: https://github.com/openthread/openthread/blob/b66ac5d9071d1ae6334cdc2947422c7980dee793/examples/platforms/nrf528xx/src/system.c#L146

It would be better to implement CSL for Nordic devices using `nrf_802154_transmit_raw_at()` and `nrf_802154_receive_at()` APIs, as it was proposed here: https://github.com/openthread/openthread/pull/4557#issuecomment-657916603

> We can add a new capability like OT_RADIO_CAP_TRANSMIT_TIMING to indicate whether platform supports precise transmission timing or not. If so, SubMac will not hold the packet and directly handle it to radio. Otherwise, SubMac handle the transmission timing, just like what we have for CSMA, retries and tx security.

Irving-cl:
> Hi @Irving-cl,
> 
> This implementation uses the `alarm.c` timer which is not so precise due to latency problems, since it depends on other processes:
> 
> https://github.com/openthread/openthread/blob/b66ac5d9071d1ae6334cdc2947422c7980dee793/examples/platforms/nrf528xx/src/system.c#L146
> 
> It would be better to implement CSL for Nordic devices using `nrf_802154_transmit_raw_at()` and `nrf_802154_receive_at()` APIs, as it was proposed here: [#4557 (comment)](https://github.com/openthread/openthread/pull/4557#issuecomment-657916603)
> 
> > We can add a new capability like OT_RADIO_CAP_TRANSMIT_TIMING to indicate whether platform supports precise transmission timing or not. If so, SubMac will not hold the packet and directly handle it to radio. Otherwise, SubMac handle the transmission timing, just like what we have for CSMA, retries and tx security.

@edmont That makes sense. I'll have a discussion with others and change the implementation.

Irving-cl:
@edmont 

I have a few questions about `nrf_802154_transmit_raw_at()` and `nrf_802154_receive_at()`:
1. We can only schedule one transmission at the same time, right? If I have already scheduled one tx using `nrf_802154_transmit_raw_at()`, and now(before it is sent) I call it again with another time, would the first one be overriden?
2. The same question with above for `nrf_802154_receive_at`.
3. Would `nrf_802154_tx_started` also be called when the transmission starts?
4. If the tx time for `transmit_raw_at` comes, and at this moment, the radio is doing another tx, what would happen? According to the comment, does it mean `nrf_802154_transmit_failed` would be called?
5. Similar question for `receive`. If at the moment `receive` starts, and the radio is currently doing tx, what would happen?
6. Does `receive_at` mean that, starting from the time specified, the radio would keep receiving until it gets other instructions? I suppose it is since it's 'a delayed version of `nrf_802154_version`'.
7. What is param `timeout` in `nrf_802154_receive_at` for? I cannot quite understand the usage for it.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -109,6 +111,13 @@ static uint32_t sEnergyDetectionTime;
 static uint8_t  sEnergyDetectionChannel;
 static int8_t   sEnergyDetected;
 
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+static uint32_t      sCslPeriod;
```
bukepo, line 115: Should this be of type uint16_t?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -943,13 +955,34 @@ void nrf_802154_receive_failed(nrf_802154_rx_error_t error)
     setPendingEvent(kPendingEventReceiveFailed);
 }
 
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+uint16_t getCslPhase()
```
bukepo, line 959: style
```suggestion
uint16_t GetCslPhase(void)
```

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -1118,3 +1158,52 @@ void otPlatRadioSetMacFrameCounter(otInstance *aInstance, uint32_t aMacFrameCoun
...
+    {
+        parentExtAddr[i] = aExtAddr->m8[sizeof(parentExtAddr) - i - 1];
+    }
+
+    updateIeData(aInstance, sParentShortAddress, parentExtAddr);
+
+    return error;
+}
+
+void otPlatRadioUpdateCslSampleTime(otInstance *aInstance, uint32_t aCslSampleTime)
```
bukepo, line 1203: should aCslSampleTime use 10 symbols unit as the period?

examples/platforms/nrf528xx/src/radio.c
Irving-cl: My consideration is, in `getCslPhase`:
```
uint32_t cslPeriodInUs = sCslPeriod * OT_US_PER_TEN_SYMBOLS;
```
If we use `uint16_t` here, I think it's possible that it overflows in the calculation process. Thoughts?

examples/platforms/nrf528xx/src/radio.c
Irving-cl: In this file, function names are like: `dataInit`, `convertShortAddress`, isPendingEventSet`, etc. Shall we keep the consistency here?

brianljt:
> Hi @Irving-cl,
> 
> This implementation uses the `alarm.c` timer which is not so precise due to latency problems, since it depends on other processes:
> 
> https://github.com/openthread/openthread/blob/b66ac5d9071d1ae6334cdc2947422c7980dee793/examples/platforms/nrf528xx/src/system.c#L146
> 
> It would be better to implement CSL for Nordic devices using `nrf_802154_transmit_raw_at()` and `nrf_802154_receive_at()` APIs, as it was proposed here: [#4557 (comment)](https://github.com/openthread/openthread/pull/4557#issuecomment-657916603)
> 
> > We can add a new capability like OT_RADIO_CAP_TRANSMIT_TIMING to indicate whether platform supports precise transmission timing or not. If so, SubMac will not hold the packet and directly handle it to radio. Otherwise, SubMac handle the transmission timing, just like what we have for CSMA, retries and tx security.

From OT stack, the timer provided is scheduled. This is when Radio platform does not provide something like `transmit_raw_at()` API, CSL can still work by the timer in `SubMac`(although might not be precise enough). In PR#5456, a new Radio capability `OT_RADIO_CAPS_TRANSMIT_TIMING` is added. So if radio platform like Nordic provides precise transmission timing control API, then `SubMac` will skip the delay and deliver the frame to Radio and radio will control the transmission timing. How do you think? @Irving-cl @edmont

examples/platforms/nrf528xx/src/radio.c
Irving-cl: Make sense. This would also cause a change in simulation radio. Shall we do it in another PR as an enhancement?

Irving-cl:
@brianljt 

Agree. That's what we discussed before. Just try to figure out some questions I mentioned above before we can use the APIs.

edmont:
@Irving-cl 

>     1. We can only schedule one transmission at the same time, right? If I have already scheduled one tx using `nrf_802154_transmit_raw_at()`, and now(before it is sent) I call it again with another time, would the first one be overriden?
>     2. The same question with above for `nrf_802154_receive_at`.
That's right, only one scheduled transmission at a time is supported. A consecutive call while the operation is pending will return `false` and the previous operation will continue. `nrf_802154_transmit_at_cancel()` or `nrf_802154_receive_at_cancel()` need to be used in order to cancel an ongoing operation.

>     3. Would `nrf_802154_tx_started` also be called when the transmission starts?
Yes, `nrf_802154_tx_started()` is called for both normal transmissions and delayed transmissions.

>     4. If the tx time for `transmit_raw_at` comes, and at this moment, the radio is doing another tx, what would happen? According to the comment, does it mean `nrf_802154_transmit_failed` would be called?
>     5. Similar question for `receive`. If at the moment `receive` starts, and the radio is currently doing tx, what would happen?
When the scheduled operation time comes, the ongoing operation at that time is terminated and it's notified with `nrf_802154_transmit_failed(pointer to psdu being transmitted)`. So the scheduled operation takes precedence over the normal one.

>     6. Does `receive_at` mean that, starting from the time specified, the radio would keep receiving until it gets other instructions? I suppose it is since it's 'a delayed version of `nrf_802154_version`'.
>     7. What is param `timeout` in `nrf_802154_receive_at` for? I cannot quite understand the usage for it.
The `timeout` param defines the reception window duration. When the window expires `nrf_802154_receive_failed()` will be called with the `NRF_802154_RX_ERROR_DELAYED_TIMEOUT` argument.
However, the receiver is not automatically disabed upon this timeout expiration, the upper layer would need to call `nrf_802154_sleep()` to do so.
Also notice that `nrf_802154_receive_failed(NRF_802154_RX_ERROR_DELAYED_TIMEOUT)` could occur even after a successful reception with `nrf_802154_received()` when the reception window is extended after the preable is received close to the end of the timeout.

Irving-cl:
@edmont 

Thanks for the kind explanation. I'd like to clarify on the last point:

> The timeout param defines the reception window duration. When the window expires nrf_802154_receive_failed() will be called with the NRF_802154_RX_ERROR_DELAYED_TIMEOUT argument.
> However, the receiver is not automatically disabed upon this timeout expiration, the upper layer would need to call nrf_802154_sleep() to do so.

Let's say, we set the timeout to 20ms. Then the receiving started, and 20ms passed. At this moment, `nrf_802154_receive_failed(NRF_802154_RX_ERROR_DELAYED_TIMEOUT)` would be called right? But the radio would still keep receiving. Could it still receive packets as normal and pass them to upper layer?

If we want to avoid `nrf_802154_receive_failed(NRF_802154_RX_ERROR_DELAYED_TIMEOUT)` being called, we should call `nrf_802154_sleep()`or other operations before receiving gets timeout, right?

brianljt (approved):
LGTM. I think we can have this for now and add enhancement for smaller sample window later.

For the sample window issue, I think we can,
1. Use receive timestamp debug to check if the start of frame is not far from the start of sample window. If it is delayed, then we should check what is the cause of the delay on CSL transmitter side, timer scheduled delay, software processing time or CSMA/CA not turned off.
2. Another possibility is, when CSL frame arrives at SSED within sample window, the SSED is doing receiving. However, `SubMac` does not know radio is receiving and will put it to sleep at the end of sample window regardless of the radio operation(Radio may be still doing reception for the frame). In my previous test, I've changed the sleep implementation to not terminate current operation. You can have a try with that.

```
--- a/third_party/NordicSemiconductor/drivers/radio/nrf_802154.c
+++ b/third_party/NordicSemiconductor/drivers/radio/nrf_802154.c
@@ -323,7 +323,7 @@ bool nrf_802154_sleep(void)

     nrf_802154_log(EVENT_TRACE_ENTER, FUNCTION_SLEEP);

-    result = nrf_802154_request_sleep(NRF_802154_TERM_802154);
+    result = nrf_802154_request_sleep(NRF_802154_TERM_NONE);
```

edmont:
@Irving-cl 
> Let's say, we set the timeout to 20ms. Then the receiving started, and 20ms passed. At this moment, `nrf_802154_receive_failed(NRF_802154_RX_ERROR_DELAYED_TIMEOUT)` would be called right? But the radio would still keep receiving. Could it still receive packets as normal and pass them to upper layer?

Actually in the case when the radio is receiving at the moment the timeout expires nothing happens yet. The radio will keep receiving and when finished there will be two consecutive calls: `nrf_802154_received()` and `nrf_802154_receive_failed(NRF_802154_RX_ERROR_DELAYED_TIMEOUT)`.

> If we want to avoid `nrf_802154_receive_failed(NRF_802154_RX_ERROR_DELAYED_TIMEOUT)` being called, we should call `nrf_802154_sleep()`or other operations before receiving gets timeout, right?

We should call `nrf_802154_receive_at_cancel()` to avoid the timeout error.

Irving-cl:
Thanks @brianljt for pointing out a mistake and suggestion for enhancement. Now the window value could be decreased to **20**.

examples/platforms/nrf528xx/src/radio.c
bukepo: fine with the name.
```suggestion
static uint16_t getCslPhase(void)
```

bukepo (approved):

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -41,6 +41,8 @@
 #include <stdint.h>
 #include <string.h>
 
+#include <openthread/link.h>
```
abtink, line 44: This is fine. Just wondering what definition/API from `link.h` is used?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -943,13 +955,34 @@ void nrf_802154_receive_failed(nrf_802154_rx_error_t error)
...
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+uint16_t getCslPhase()
+{
+    uint32_t curTime       = otPlatAlarmMicroGetNow();
+    uint32_t cslPeriodInUs = sCslPeriod * OT_US_PER_TEN_SYMBOLS;
+    uint32_t diff = ((sCslSampleTime % cslPeriodInUs) - (curTime % cslPeriodInUs) + cslPeriodInUs) % cslPeriodInUs;
+
+    return (uint16_t)(diff / OT_US_PER_TEN_SYMBOLS);
+}
+#endif // OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
```
abtink, line 967:
```suggestion
#endif
```

examples/platforms/nrf528xx/src/radio.c
Irving-cl: It's `OT_US_PER_TEN_SYMBOLS` defined in `link.h`. It's currently used in both simulation and nrf528xx platforms.

src/core/mac/sub_mac.cpp
```diff
@@ -253,10 +253,11 @@ void SubMac::HandleReceiveDone(RxFrame *aFrame, otError aError)
 #if OPENTHREAD_CONFIG_MAC_CSL_DEBUG_ENABLE
```
brianljt, line 253:
```suggestion
#if OPENTHREAD_CONFIG_MAC_CSL_DEBUG_ENABLE && OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
```

src/core/mac/sub_mac.cpp
```diff
@@ -253,10 +253,11 @@ void SubMac::HandleReceiveDone(RxFrame *aFrame, otError aError)
...
     if (aFrame != nullptr && aError == OT_ERROR_NONE)
     {
-        otLogDebgMac(
-            "Received frame in state (SubMac %s, CSL %d), timestamp %lu, target sample start time %u, time drift %d",
-            StateToString(mState), mCslState, aFrame->mInfo.mRxInfo.mTimestamp, mCslSampleTime.GetValue(),
-            static_cast<uint32_t>(aFrame->mInfo.mRxInfo.mTimestamp) - mCslSampleTime.GetValue());
+        // Split the log into two lines for RTT to output
+        otLogDebgMac("Received frame in state (SubMac %s, CSL %d), timestamp %u", StateToString(mState),
```
brianljt, line 257:
```suggestion
        otLogDebgMac("Received frame in state (SubMac %s, CSL %s), timestamp %u", StateToString(mState),
```

src/core/mac/sub_mac.cpp
Irving-cl: It would be checked that, when `OPENTHREAD_CONFIG_MAC_CSL_DEBUG_ENABLE`  is enabled, `OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE` has to be enabled when compiling: https://github.com/openthread/openthread/blob/master/src/core/config/openthread-core-config-check.h#L520

examples/platforms/nrf528xx/nrf52840/openthread-core-nrf52840-config.h
```diff
@@ -268,6 +268,16 @@
...
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_CSL_SAMPLE_WINDOW
+ *
+ * CSL sample window in units of 10 symbols.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_CSL_SAMPLE_WINDOW
+#define OPENTHREAD_CONFIG_CSL_SAMPLE_WINDOW 20
```
edmont, line 278: My testing results in a considerable number of MAC retries required for delivering frames. A value of 30 would be much safer by the moment.

edmont (approved):
LGTM as a starting point, but there is still room for low-power improvement.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -109,6 +111,13 @@ static uint32_t sEnergyDetectionTime;
 static uint8_t  sEnergyDetectionChannel;
 static int8_t   sEnergyDetected;
 
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+static uint32_t      sCslPeriod;
+static uint32_t      sCslSampleTime;
+static const uint8_t sCslIeHeader[OT_IE_HEADER_SIZE] = {0x04, 0x0d};
+static uint8_t       sParentShortAddress[SHORT_ADDRESS_SIZE];
```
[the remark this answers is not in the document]
Irving-cl: Used `nrf_802154_pib_short_address_get()`. 

Do we consider adding a method `nrf_802154_short_address_get` in `nrf_802154.h`? As I notice, for `nrf_802154_pib_short_address_set()`, we could call `nrf_802154_short_address_set()`in `nrf_802154.h`, which then calls `nrf_802154_pib_short_address_set()`. Should we do the same practice for `nrf_802154_short_address_get`?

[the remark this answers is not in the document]
edmont: Yes, that'd be better in order to keep consistency.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -109,6 +111,13 @@ static uint32_t sEnergyDetectionTime;
 static uint8_t  sEnergyDetectionChannel;
 static int8_t   sEnergyDetected;
 
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+static uint32_t      sCslPeriod;
+static uint32_t      sCslSampleTime;
+static const uint8_t sCslIeHeader[OT_IE_HEADER_SIZE] = {0x04, 0x0d};
```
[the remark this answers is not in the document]
edmont: Now this can be applied to the simulation platform as well.

brianljt:
Please check `void SubMac::BeginTransmit(void)` line 408. I think you made a mistake here, this line should be
```
mTransmitFrame.SetCsmaCaEnabled(mTransmitFrame.mInfo.mTxInfo.mPeriod == 0);
```
to turn off CSMA/CA for CSL transmission.

In my tests, after this change, the CSL frame will arrive at the SSED within 1ms window. You can change the window size to 10 or smaller. Please have a try and update the PR accordingly.

edmont:
@brianljt good catch!

However, the testing is not still smooth on my side with the window size set to 10. See the attached captures in which I send 20 pings from the parent at a random interval between 1 and 10 seconds. We shouldn't see MAC retransmissions in this kind of testing in order to consider an acceptable window size. Even with a window size set to 20 I observed one retransmission.

[csl-captures.zip](https://github.com/openthread/openthread/files/5160887/csl-captures.zip)

brianljt:
> @brianljt good catch!
> 
> However, the testing is not still smooth on my side with the window size set to 10. See the attached captures in which I send 20 pings from the parent at a random interval between 1 and 10 seconds. We shouldn't see MAC retransmissions in this kind of testing in order to consider an acceptable window size. Even with a window size set to 20 I observed one retransmission.
> 
> 

@edmont , I had a try on my setup with the default window size of 10(1.6ms) works fine and 20 pings from parent to SSED will result in no retransmission. [csl-captures.zip](https://github.com/openthread/openthread/files/5160887/csl-captures.zip)

To narrow down this issue, you may want to enable macro `OPENTHREAD_CONFIG_MAC_CSL_DEBUG_ENABLE` for SSED and check the RTT log at SSED side for function `SubMac::HandleReceiveDone` to see when are CSL frame received(how much delay is introduced). Before it is sorted out, I think we can keep the default window size of 30 for safety. @Irving-cl 

I can think of one possible cause to this,
The clock source is not match at both side. One clock is faster than another. In my setup, both side use same batch of nrf52840dk boards. I'm working on an enhancement on SSED side to take timer accuracy into consideration so that sample window size and sample time can be dynamically adjusted.

Irving-cl:
> @brianljt good catch!
> 
> However, the testing is not still smooth on my side with the window size set to 10. See the attached captures in which I send 20 pings from the parent at a random interval between 1 and 10 seconds. We shouldn't see MAC retransmissions in this kind of testing in order to consider an acceptable window size. Even with a window size set to 20 I observed one retransmission.
> 
> [csl-captures.zip](https://github.com/openthread/openthread/files/5160887/csl-captures.zip)

@edmont 
Have changed it back to `30`. Will continue to enhance it.

edmont (approved):
Thanks!

jwhui (approved):

PR #5451, opened 2020-08-24
[k32w] rtos uart fixes

doru91:
This PR adds fixes which are needed for running Project CHIP on K32W.

Fixes are related to UART handling, radio interrupts and Free RTOS support.

### DIFF
--- a/examples/platforms/k32w/src/uart.c
+++ b/examples/platforms/k32w/src/uart.c
@@ -272,7 +272,6 @@ static void USART0_IRQHandler(USART_Type *base, usart_handle_t *handle)
         }
 
 #if USE_RTOS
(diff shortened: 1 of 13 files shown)


examples/platforms/k32w/src/uart.c
```diff
@@ -163,6 +175,7 @@ otError otPlatUartFlush(void)
  * @param[in] aBuf             Pointer to the character buffer
  * @param[in] len              Length of the character buffer
  */
+
```
jwhui, line 178: nit: is this line addition intentional?

doru91: I removed the empty line.

PR #5454, opened 2020-08-25
[message] update and fix 'CopyTo()'

abtink:
This commit updates `Message::CopyTo()` method to use data chunks
(this copies data directly from the source message to the destination
avoiding the need for an extra buffer copy). It also addresses an
issue where if the requested copy length is more than the available
bytes in the source message, incorrect bytes may be copied into the
destination message, and the returned value (i.e. number of bytes
copied) can be invalid.

This commit also updates the unit test `test_message` covering
behavior of `CopyTo()` for different offsets and copy lengths. It also
checks the behavior of `Read()` when the read length goes beyond the
end of the message (trying to read more bytes than available in the
message).

-------------

Checked all places where `CopyTo()` is currently used and the "read length" 
in  all current cases is smaller than source message length.  So the issue/bug is
not really critical (we can consider asserting instead, but I feel it's better to keep 
its behavior similar to how `Read()` behaves, i.e. copy as many bytes as possible
and the return the number of bytes copied).

### DIFF
--- a/src/core/common/message.cpp
+++ b/src/core/common/message.cpp
@@ -434,6 +434,8 @@ void Message::GetFirstChunk(uint16_t aOffset, uint16_t &aLength, Chunk &aChunk) 
     // its length. The `aLength` is also decreased by the chunk
     // length.
 
+    VerifyOrExit(aOffset < GetLength(), aChunk.mLength = 0);
+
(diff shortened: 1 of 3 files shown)


wgtdkp (approved):
LGTM 👍

Issue #5460, opened 2020-08-26
[continuous-integration] fuzz failure in `Ip6::InsertMplOption`


Irving-cl:
**Describe the bug**
Some PRs now get failure in fuzzers: #5455 #5458 #5459 

I think the reason is that, in `src/core/net/ip6.cpp:283`, it's called like this:
```
    aMessage.CopyTo(8, 0, hbhLength, aMessage);
```
This message copies to itself.

And in the newly updated `CopyTo` method, it uses `memcpy`, when `hbhLength` is greater than `8`, it would cause a 'overlap', which causes the error indicated in fuzzer:

> ==11==ERROR: AddressSanitizer: memcpy-param-overlap: memory ranges [0x00000131bf10,0x00000131bf20) and [0x00000131bf18, 0x00000131bf28) overlap
> SCARINESS: 10 (memcpy-param-overlap)
>     #0 0x51fce4 in __asan_memcpy /src/llvm-project/compiler-rt/lib/asan/asan_interceptors_memintrinsics.cpp:22:3
>     #1 0x55d010 in ot::Message::Write(unsigned short, unsigned short, void const*) /src/openthread/src/core/common/message.cpp:540:9
>     #2 0x55e066 in ot::Message::CopyTo(unsigned short, unsigned short, unsigned short, ot::Message&) const /src/openthread/src/core/common/message.cpp:555:18
>     #3 0x592bf0 in ot::Ip6::Ip6::InsertMplOption(ot::Message&, ot::Ip6::Header&, ot::Ip6::MessageInfo&) /src/openthread/src/core/net/ip6.cpp:283:22
>     #4 0x5958fc in ot::Ip6::Ip6::SendRaw(ot::Message&) /src/openthread/src/core/net/ip6.cpp:1108:9
>     #5 0x554960 in otIp6Send /src/openthread/src/core/api/ip6_api.cpp:167:37
>     #6 0x553e7b in LLVMFuzzerTestOneInput /src/openthread/tests/fuzz/ip6_send.cpp:72:13
>     #7 0x458621 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:559:15
[...]

PR #5462, opened 2020-08-26
[message] update 'Write()' to use 'memmove()' instead of 'memcpy()'

abtink:
This change allows `CopyTo()` method to be used to copy bytes backwards
within the same message (i.e., src and dest messages are the same and
the dest offset is before src offset). Unit test `test_message` is
updated in this commit to cover such a use.

However, the `CopyTo()` implementing  can still potentially overwrite
the data when bytes are being copied forward within the same message,
i.e., source and destination messages are the same, and source offset
is smaller than the destination offset. This commit adds an assert not
allowing such a use and updates the method documentation mentioning
this restriction.

----------

This change should address https://github.com/openthread/openthread/issues/5460 and unblock the fuzzer error on PRs.

Checking the `Message::CopyTo()` uses in the code, we seem to be copying within same 
message only from `Ip6::InsertMplOption()` and it's backward copy which should be addressed
by using `memmove` (other places we seem to copy between different source and destination 
messages which is fine). For now added an assert to restrict the `CopyTo()`'s use to forward 
copy within the same message 

I think the overlap write was an issue with the previous implementation (before https://github.com/openthread/openthread/pull/5454)  as well but it
was being masked (not detected by fuzzer) by the fact that `CopyTo()` would read content into a 
separate small 16 bytes buffer from  source message and the write them in destination.

### DIFF
--- a/src/core/common/message.cpp
+++ b/src/core/common/message.cpp
@@ -537,7 +537,7 @@ void Message::Write(uint16_t aOffset, uint16_t aLength, const void *aBuf)
 
     while (chunk.GetLength() > 0)
     {
-        memcpy(chunk.GetData(), bufPtr, chunk.GetLength());
+        memmove(chunk.GetData(), bufPtr, chunk.GetLength());
         bufPtr += chunk.GetLength();
         GetNextChunk(aLength, chunk);
     }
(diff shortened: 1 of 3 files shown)


src/core/common/message.cpp
```diff
@@ -548,6 +548,13 @@ uint16_t Message::CopyTo(uint16_t aSourceOffset, uint16_t aDestinationOffset, ui
     uint16_t bytesCopied = 0;
     Chunk    chunk;
 
+    // The implementing can potentially overwrite the data when bytes are
```
jwhui, line 551:
```suggestion
    // This implementation can potentially overwrite the data when bytes are
```

PR #5465, opened 2020-08-27
[diag] clear output in case no output given

bukepo:
The `stop` command doesn't write anything to the output buffer.


simonlingoogle (approved):
LGTM :+1:

PR #5466, opened 2020-08-27
[ot-ctl] auto reconnect session 

simonlingoogle:
This PR includes: 
- [x] ot-ctl automatically reconnect session when disconnected
- [x] add test for reset/factoryreset

Fixes https://github.com/openthread/ot-br-posix/issues/474

### DIFF
--- a/src/posix/client.cpp
+++ b/src/posix/client.cpp
@@ -28,6 +28,8 @@
 
 #include "platform/openthread-posix-config.h"
 
+#include <chrono>
+
 #include <openthread/platform/toolchain.h>
 
 #ifndef HAVE_LIBEDIT
@@ -46,6 +48,7 @@
 #include <string.h>
 #include <sys/socket.h>
 #include <sys/un.h>
+#include <time.h>
 #include <unistd.h>
 
 #if HAVE_LIBEDIT
@@ -108,17 +111,17 @@ exit:
     return ret;
 }
 
-int main(int argc, char *argv[])
+static int ConnectSession(void)
 {
-    int    ret;
-    bool   isInteractive = true;
-    bool   isFinished    = false;
-    char   lineBuffer[kLineBufferSize];
-    size_t lineBufferWritePos = 0;
-    bool   isBeginOfLine      = true;
+    int ret = 0;
+
+    if (sSessionFd != -1)
+    {
+        close(sSessionFd);
+    }
 
     sSessionFd = socket(AF_UNIX, SOCK_STREAM, 0);
-    VerifyOrExit(sSessionFd != -1, perror("socket"); ret = OT_EXIT_FAILURE);
+    VerifyOrExit(sSessionFd != -1, ret = -1);
 
     {
         struct sockaddr_un sockname;
@@ -128,13 +131,45 @@ int main(int argc, char *argv[])
         strncpy(sockname.sun_path, OPENTHREAD_POSIX_DAEMON_SOCKET_NAME, sizeof(sockname.sun_path) - 1);
 
         ret = connect(sSessionFd, reinterpret_cast<const struct sockaddr *>(&sockname), sizeof(struct sockaddr_un));
-
-        if (ret == -1)
-        {
-            fprintf(stderr, "OpenThread daemon is not running.\n");
-            ExitNow(ret = OT_EXIT_FAILURE);
-        }
     }
+
+exit:
+    if (ret == -1 && sSessionFd != -1)
+    {
+        close(sSessionFd);
+        sSessionFd = -1;
+    }
+    return ret;
+}
+
+static bool ReconnectSession(void)
+{
+    bool     ok    = false;
+    uint32_t delay = 100000; // 100ms
+
+    for (int i = 0; i < 5; i++) // delay for 3.1s in total
+    {
+        usleep(delay);
+
+        VerifyOrExit(ConnectSession() == -1, ok = true);
+
+        delay *= 2;
+    }
+
+exit:
+    return ok;
+}
+
+int main(int argc, char *argv[])
+{
+    int    ret;
+    bool   isInteractive = true;
+    bool   isFinished    = false;
+    char   lineBuffer[kLineBufferSize];
+    size_t lineBufferWritePos = 0;
+    bool   isBeginOfLine      = true;
+
+    VerifyOrExit(ConnectSession() != -1, perror("connect session failed"); ret = OT_EXIT_FAILURE);
 
     if (argc > 1)
     {
@@ -205,6 +240,11 @@ int main(int argc, char *argv[])
             if (rval == 0)
             {
                 // daemon closed sSessionFd
+                if (isInteractive && ReconnectSession())
+                {
+                    continue;
+                }
+
                 ExitNow(ret = isInteractive ? OT_EXIT_FAILURE : OT_EXIT_SUCCESS);
             }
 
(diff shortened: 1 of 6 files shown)


src/posix/client.cpp
```diff
@@ -128,14 +129,44 @@ int main(int argc, char *argv[])
...
+    }
+    return ret;
+}
+
+static bool ReconnectSession(void)
+{
+    bool    ok       = false;
+    clock_t deadline = clock() + 3 * CLOCKS_PER_SEC;
+
+    while (clock() <= deadline)
```
romacdon, line 148: 2 points on this line - 1) you should include "&& !ok". and 2) there's an issue if clock() is less than 3 seconds from wrapping when setting deadline.  In that case you won't enter the while loop because deadline will have been set to a value less than clock().

simonlingoogle: > 1) you should include "&& !ok".

Actually it not necessary to check `!ok` in while because `VerifyOrExit(ConnectSession() == -1, ok = true);` sets `ok` to true and `goto exit`. So, we always have `!ok` satisfied in the loop. 

> 2) there's an issue if clock() is less than 3 seconds from wrapping when setting deadline. In that case you won't enter the while loop because deadline will have been set to a value less than clock().

Good catch! I think we should use `std::chrono` utilities instead. Any suggestions?

romacdon: > > 1. you should include "&& !ok".
> 
> Actually it not necessary to check `!ok` in while because `VerifyOrExit(ConnectSession() == -1, ok = true);` sets `ok` to true and `goto exit`. So, we always have `!ok` satisfied in the loop.
> 
Oh right.  Its not necessary, nice!

> > 1. there's an issue if clock() is less than 3 seconds from wrapping when setting deadline. In that case you won't enter the while loop because deadline will have been set to a value less than clock().
> 
> Good catch! I think we should use `std::chrono` utilities instead. Any suggestions?

I was thinking some like:

simonlingoogle: changed to just use retry. please check.

src/posix/client.cpp
```diff
@@ -28,6 +28,8 @@
 
 #include "platform/openthread-posix-config.h"
 
+#include <chrono>
```
bukepo, line 31: Is this used?

src/posix/client.cpp
```diff
@@ -128,14 +131,46 @@ int main(int argc, char *argv[])
...
+static bool ReconnectSession(void)
+{
+    bool     ok    = false;
+    uint32_t delay = 100000; // 100ms
+
+    for (int i = 0; i < 5; i++) // delay for 3.1s in total
+    {
+        usleep(delay);
+
+        VerifyOrExit(ConnectSession() == -1, ok = true);
```
bukepo, line 154: print error information of each failure?

bukepo (commented):
will this cause intended exit waiting 3s?

src/posix/client.cpp
simonlingoogle: The current behavior is mostly like OT CLI. 

Note that the first delay is 100ms. It is possible that ot-daemon hasn't finished restarting in 100ms, so I would prefer `ot-ctl` to simply wait for more time without printing any error. 

Furthermore, if all 5 reconnecting retry failed, it's very likely that ot-daemon has really exited, instead of being reset. In such case, I am fine to print error before `ot-ctl` exits, but exiting quietly should also be fine?

src/posix/client.cpp
simonlingoogle: No it's not, will remove

simonlingoogle:
> will this cause intended exit waiting 3s?

It will. We need to fix this.

src/posix/client.cpp
romacdon: > changed to just use retry. please check.

LGTM

simonlingoogle:
**Labeled as WIP since it's unclear how to exit immediately when `exit` command is issued.**
The current implementation would cause `ot-ctl` to also wait for about 3 seconds and keep reconnecting in such condition, which I believe should be fixed. 

Currently, the best solution I have in mind:

- First, enhance `ot-daemon` to always remove `openthread.sock` when exit and not remove when reset. 
- `ot-ctl` checks if the file `openthread.sock` exists. If not, it quit immediately.

The main idea is for `ot-ctl` to find a way to determine if `ot-daemon` (or otbr-agent) process really exit. 

Welcome to input. @bukepo @romacdon @jwhui

romacdon:
> **Labeled as WIP since it's unclear how to exit immediately when `exit` command is issued.**
> The current implementation would cause `ot-ctl` to also wait for about 3 seconds and keep reconnecting in such condition, which I believe should be fixed.
> 
> Currently, the best solution I have in mind:
> 
> * First, enhance `ot-daemon` to always remove `openthread.sock` when exit and not remove when reset.
> * `ot-ctl` checks if the file `openthread.sock` exists. If not, it quit immediately.
> 
> The main idea is for `ot-ctl` to find a way to determine if `ot-daemon` (or otbr-agent) process really exit.
> 
> Welcome to input. @bukepo @romacdon @jwhui

I poked around in the quote a bit to get a better understanding how this works.  I thought perhaps we could do something by not "closing on exec" the socket to ot-ctl but in the end that seemed too complicated.  I think your approach suggested above is probably the simplest to implement.  

In my investigation I did notice OPENTHREAD_POSIX_DAEMON_SOCKET_LOCK  too and was wondering if that could somehow be used but I never did figure out exactly what that was used for.

romacdon:
Hi @simonlingoogle, do you think we are going move ahead with the solution you proposed with regards to exiting immediately when the "exit" command is invoked or are you waiting for additional input from others?

romacdon (changes requested):

src/posix/platform/misc.cpp
```diff
@@ -40,14 +40,14 @@
 #include "common/code_utils.hpp"
 #include "common/logging.hpp"
 
-static otPlatResetReason   sPlatResetReason   = OT_PLAT_RESET_REASON_POWER_ON;
+otPlatResetReason          gPlatResetReason   = OT_PLAT_RESET_REASON_POWER_ON;
```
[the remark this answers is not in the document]
simonlingoogle: I tried to not use a global variable but it seems not that simple to do so. 

I think `otPlatGetResetReason` are supposed to be defined by platform code and used by core code. So, it not proper to define a new method (i.e. otPlatSetResetReason) only to be called by core code. 

We are already using other global variables to be accessed by multiple platform sources. That's why I think using global variable is acceptable here.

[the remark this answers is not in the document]
romacdon: > I tried to not use a global variable but it seems not that simple to do so.
> 
> I think `otPlatGetResetReason` are supposed to be defined by platform code and used by core code. So, it not proper to define a new method (i.e. otPlatSetResetReason) only to be called by platform code.
> 
> We are already using other global variables to be accessed by multiple platform sources. That's why I think using global variable is acceptable here.

I see and since the precedent has already been set I'm good with the change.

romacdon (approved):

simonlingoogle:
Now it quits immediately if `exit` command is issued in `ot-ctl`. 
Please check. @bukepo @gjc13

gjc13 (approved):

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #5468, opened 2020-08-27
[backbone] add API to configure Backbone netif

simonlingoogle:


simonlingoogle:
close infavor of https://github.com/openthread/openthread/pull/5471

PR #5469, opened 2020-08-27
[ip6] skip copying message from 'ProcessReceiveCallback()'

abtink:
This commit adds a new parameter to `Ip6::ProcessReceiveCallback()`
allowing caller to indicate whether the method can take custody of the
passed-in message or whether it needs to make a copy of it (to use in
receive callback). This is indicated by a new `enum` definition
`Message::Ownership`. In case of `kTakeCustody` value, the caller
transfers custody of message to the called method (i.e. the method can
`Enqueue()` the message and would need to `Free()` the message when no
longer needed).

The `Ip6::HandleDatagram()` uses this to avoid copying the message in
cases where it is not needed (e.g. for a message that is passed to
host through callback but does not need to be processed by Thread
core).

### DIFF
--- a/src/core/net/ip6.cpp
+++ b/src/core/net/ip6.cpp
@@ -950,7 +950,7 @@ otError Ip6::HandleExtensionHeaders(Message &    aMessage,
 
         case kProtoFragment:
             // Always forward IPv6 fragments to the Host.
-            IgnoreError(ProcessReceiveCallback(aMessage, aMessageInfo, aNextHeader, aFromNcpHost));
+            IgnoreError(ProcessReceiveCallback(aMessage, aMessageInfo, aNextHeader, aFromNcpHost, Message::kCopyToUse));
 
             SuccessOrExit(error = HandleFragment(aMessage, aNetif, aMessageInfo, aFromNcpHost));
             break;
@@ -1002,13 +1002,14 @@ otError Ip6::HandlePayload(Message &aMessage, MessageInfo &aMessageInfo, uint8_t
     return error;
 }
 
-otError Ip6::ProcessReceiveCallback(const Message &    aMessage,
+otError Ip6::ProcessReceiveCallback(Message &          aMessage,
                                     const MessageInfo &aMessageInfo,
                                     uint8_t            aIpProto,
-                                    bool               aFromNcpHost)
+                                    bool               aFromNcpHost,
+                                    Message::Ownership aMessageOwnership)
 {
-    otError  error = OT_ERROR_NONE;
-    Message *messageCopy;
+    otError  error   = OT_ERROR_NONE;
+    Message *message = &aMessage;
 
     VerifyOrExit(!aFromNcpHost, error = OT_ERROR_NO_ROUTE);
     VerifyOrExit(mReceiveIp6DatagramCallback != nullptr, error = OT_ERROR_NO_ROUTE);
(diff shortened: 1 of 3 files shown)


Irving-cl (approved):
LGTM👍

src/core/net/ip6.cpp
```diff
@@ -1237,17 +1254,13 @@ otError Ip6::HandleDatagram(Message &aMessage, Netif *aNetif, const void *aLinkM
...
-
-            ExitNow();
+            error = ProcessReceiveCallback(aMessage, messageInfo, nextHeader, aFromNcpHost, Message::kTakeCustody);
+            ExitNow(shouldFreeMessage = false);
         }
 
         if (aNetif != nullptr)
         {
-            VerifyOrExit(mForwardingEnabled, forward = false);
+            VerifyOrExit(mForwardingEnabled, OT_NOOP);
```
simonlingoogle, line 1263:
```suggestion
            VerifyOrExit(mForwardingEnabled, foward = false);
```

why remove `forward = false` ?

abtink: `forward` flag was earlier used at `exit` label to determine whether to free the message or not (earlier we had a rather complex set of checks `(!tunnel && (error != OT_ERROR_NONE || !forward))` to decide whether to free or not). From here, the code was setting `forward` to `false` to ensure message is freed (since `error` is not being set). 

With the changes in this commit, we don't need this anymore. This commit adds a new simpler flag `shouldFreeMessage` to track this.
- It's set by default to `true` (i.e. we need to free the message by default),
- From two places in code where we know message's custody is successfully given to another method, we set the `shouldFreeMessage` to `false`

simonlingoogle: I think you are right about the message freeing logic and whether or not we set `forward` to false does not matter any more in the current code. 

But I was hoping that we should set both `receive` and `forward` to the correct value, i.e. `forward` is `true` iff the message was forwarded. Later, someone might add some code at exit that checks `forward` to determine if the message was forwarded. 

Thoughts?

abtink: `forward` variable is used to indicate whether the message should be forwarded or not and not whether the message was actually forwarded. 

There are other checks before and after this part of the code where we can exit (e.g., we have the `if (!ShouldForwardToThread(messageInfo))` check before and the hop limit check after) and in both cases the message won't be forwarded but the `forward` variable remains `true` as before.

If we want to change the meaning of `forward` for possible future use, we would need to change all other instances as well. I would say it's better not to add additional code that serves no immediate purpose or use. 

The reason for the existing code setting `forward` to `false` here seems to be to ensure message is freed.

If in future there is a need to determine whether the message was successfully forwarded, the best way to do it, would be to add it right after the `SuccessOrExit(error = Get<ThreadNetif>().SendMessage(aMessage))` statement.

simonlingoogle: I think we can interpret `forward` in both ways in the original code. 

The `(!ShouldForwardToThread(messageInfo))` case can ben seen as **forwarding** to Host.  
The hop limit checking failed case can be seen as a `forwarding` failure. 

The `!mForwardingEnabled` case is a little bit different in which it decided not to forward since it's not enabled. 

But it's fine to me since `forward` is not used at exit.

simonlingoogle (approved):
LGTM 👍 

This can be a great performance boost for Thread-Backbone traffic.

PR #5470, opened 2020-08-28
[mle] simplify logging and add helper log methods

abtink:
This commit simplifies the logging in `Mle` modules. It renames and
simplifies the existing MLE logging methods. They are now defined as
`static` methods and the main implementation uses `#if` guards
checking the proper log level (when logs are not enabled they are
defined as empty inline methods).

This commit also adds new helper methods to log failure of sending or
processing of an MLE message, namely `Mle::LogSendError()` and
`Mle::LogProcessError()`. These are then used to simplify logging in
many methods in `Mle` and other related modules.

### DIFF
--- a/src/core/thread/discover_scanner.cpp
+++ b/src/core/thread/discover_scanner.cpp
@@ -154,7 +154,7 @@ otError DiscoverScanner::Discover(const Mac::ChannelMask &aScanChannels,
     mScanChannel = Mac::ChannelMask::kChannelIteratorFirst;
     mState       = (mScanChannels.GetNextChannel(mScanChannel) == OT_ERROR_NONE) ? kStateScanning : kStateScanDone;
 
-    otLogInfoMle("Send Discovery Request (%s)", destination.ToString().AsCString());
+    Mle::Log("Send Discovery Request", destination);
 
 exit:
 
@@ -302,7 +302,7 @@ void DiscoverScanner::HandleDiscoveryResponse(const Message &aMessage, const Ip6
     uint16_t                      end;
     bool                          didCheckSteeringData = false;
 
-    otLogInfoMle("Receive Discovery Response (%s)", aMessageInfo.GetPeerAddr().ToString().AsCString());
+    Mle::Log("Receive Discovery Response", aMessageInfo.GetPeerAddr());
 
     VerifyOrExit(mState == kStateScanning, error = OT_ERROR_DROP);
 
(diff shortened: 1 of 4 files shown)


src/core/thread/discover_scanner.cpp
```diff
@@ -154,7 +154,7 @@ otError DiscoverScanner::Discover(const Mac::ChannelMask &aScanChannels,
     mScanChannel = Mac::ChannelMask::kChannelIteratorFirst;
     mState       = (mScanChannels.GetNextChannel(mScanChannel) == OT_ERROR_NONE) ? kStateScanning : kStateScanDone;
 
-    otLogInfoMle("Send Discovery Request (%s)", destination.ToString().AsCString());
+    Mle::Log("Send Discovery Request", destination);
```
simonlingoogle, line 157: Further reduce code size ?
```suggestion
    Mle::Log("Send", "Discovery Request", destination);
```
That is to say, `Log` has two arguments: `aAction` and `aWhat`. 
We can save code size because most `aAction` would be `Send` or `Receive` so compiler will use the same strings in `.rodata` section. 
Many `aWhat` arguments  would also be equal so they can be reused.

abtink: I noticed the same... :)... that is that the MLE message string is being repeated in different logs. My idea for this is to introduce the "MLE message type" as an `enum` (or use the existing `Mle::Header::Command` enumeration) that is then converted to a string. This would be bigger change and probably better for a future PR.

I would suggest not optimizing by using same literal strings as we would then rely on the compiler/toolchain to do this optimization (i.e. use one actual instance for all same string literal definitions). While most advanced compilers would do this, the behavior is toolchain dependent.

simonlingoogle: Using `MLE message type` enum might not be a good fit for the problem? For example, we can't convert this code just using enum. 
```c++
Log("Send Child ID Request - short", destination);
```

But we can convert it to 
```c++
Log("Send", "Child ID Request - short", destination);
```
Furthermore, this optimization should be common for compilers. 
But I am fine to apply this optimization in a future PR.

abtink: > Using MLE message type enum might not be a good fit for the problem

I also noticed that we have some special cases  :)....(for example for "Child Update" as well)
That's why I mentioned to possibly introduce a new "MLE message type" enum (may be a super-set of `Header::Command`) for logging purposes.

> Furthermore, this optimization should be common for compilers.

I have been surprised in past (many times) by optimizations/behaviors that I expected a toolchain would do to realize later that was not supported. I would always suggest not to rely on toolchain optimizations as much as possible and try to optimize the code itself. Specially since OT is a generic network stack firmware and can be built by vendor/hardware-specific toolchains.

> That being said, I am fine to apply this optimization in a future PR.

👍

simonlingoogle (approved):
LGTM 👍

PR #5388, opened 2020-08-13
[mlr] send BMLR.ntf on Backbone link

simonlingoogle:
This PR enhances Backbone Router to send BMLR.ntf on the backbone link for Multicast Listeners.

- [x] Send `BMLR.ntf` to the Backbone link
- [x] Add Backbone test with packet verification
- [x] Includes some revision to the Thread 1.2 Backbone CI scripts to make it more stable 

**Receiving and processing `BMLR.ntf` on the Backbone link is not implemented in this PR**

**MLR PR List:**
- https://github.com/openthread/ot-br-posix/pull/532 add multicast forwarding for MLR Listeners
- https://github.com/openthread/openthread/pull/5388 send BMLR.ntf on Backbone link **<== This PR**

PR #5471, opened 2020-08-28
[tmf] add Backbone TMF agent

simonlingoogle:
This PR adds the Backbone TMF agent

### DIFF
--- a/src/core/backbone_router/backbone_tmf.cpp
+++ b/src/core/backbone_router/backbone_tmf.cpp
@@ -55,7 +55,7 @@ otError BackboneTmfAgent::Filter(const ot::Coap::Message &aMessage,
                                                                                          : OT_ERROR_NOT_TMF;
 }
 
-bool BackboneTmfAgent::IsBackboneTmfMessage(const Ip6::MessageInfo &aMessageInfo)
+bool BackboneTmfAgent::IsBackboneTmfMessage(const Ip6::MessageInfo &aMessageInfo) const
 {
     const Ip6::Address &dst = aMessageInfo.GetSockAddr();
     const Ip6::Address &src = aMessageInfo.GetPeerAddr();
--- a/src/core/backbone_router/backbone_tmf.cpp
+++ b/src/core/backbone_router/backbone_tmf.cpp
@@ -28,46 +28,48 @@
 
 /**
  * @file
- *   This file implements Thread Management Framework (TMF) functionalities.
+ *   This file implements Backbone Thread Management Framework (TMF) functionalities.
  */
 
-#include "thread/tmf.hpp"
+#include "backbone_tmf.hpp"
+
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
 
 #include "common/locator-getters.hpp"
 
 namespace ot {
-namespace Tmf {
+namespace BackboneRouter {
 
-otError TmfAgent::Start(void)
+otError BackboneTmfAgent::Start(void)
 {
-    return Coap::Start(kUdpPort);
+    return Coap::Start(kBackboneUdpPort, OT_NETIF_BACKBONE);
 }
 
-otError TmfAgent::Filter(const ot::Coap::Message &aMessage, const Ip6::MessageInfo &aMessageInfo, void *aContext)
+otError BackboneTmfAgent::Filter(const ot::Coap::Message &aMessage,
+                                 const Ip6::MessageInfo & aMessageInfo,
+                                 void *                   aContext)
 {
     OT_UNUSED_VARIABLE(aMessage);
 
-    return static_cast<TmfAgent *>(aContext)->IsTmfMessage(aMessageInfo) ? OT_ERROR_NONE : OT_ERROR_NOT_TMF;
+    return static_cast<BackboneTmfAgent *>(aContext)->IsBackboneTmfMessage(aMessageInfo) ? OT_ERROR_NONE
+                                                                                         : OT_ERROR_NOT_TMF;
 }
 
-bool TmfAgent::IsTmfMessage(const Ip6::MessageInfo &aMessageInfo) const
+bool BackboneTmfAgent::IsBackboneTmfMessage(const Ip6::MessageInfo &aMessageInfo)
 {
-    bool rval = true;
+    const Ip6::Address &dst = aMessageInfo.GetSockAddr();
+    const Ip6::Address &src = aMessageInfo.GetPeerAddr();
 
-    // A TMF message must comply with following rules:
-    // 1. The destination is a Mesh Local Address or a Link-Local Multicast Address or a Realm-Local Multicast Address,
-    //    and the source is a Mesh Local Address. Or
-    // 2. Both the destination and the source are Link-Local Addresses.
-    VerifyOrExit(
-        ((Get<Mle::MleRouter>().IsMeshLocalAddress(aMessageInfo.GetSockAddr()) ||
-          aMessageInfo.GetSockAddr().IsLinkLocalMulticast() || aMessageInfo.GetSockAddr().IsRealmLocalMulticast()) &&
-         Get<Mle::MleRouter>().IsMeshLocalAddress(aMessageInfo.GetPeerAddr())) ||
-            ((aMessageInfo.GetSockAddr().IsLinkLocal() || aMessageInfo.GetSockAddr().IsLinkLocalMulticast()) &&
-             aMessageInfo.GetPeerAddr().IsLinkLocal()),
-        rval = false);
-exit:
-    return rval;
+    // A Backbone TMF message must comply with following rules:
+    // The destination must be one of:
+    //     1. All Network BBRs (Link-Local scope)
+    //     2. All Domain BBRs (Link-Local scope)
+    //     3. A Backbone Link-Local address
+    // The source must be a Backbone Link-local address.
+    return (dst.IsLinkLocal() || dst.IsLinkLocalMulticast()) && src.IsLinkLocal();
 }
 
-} // namespace Tmf
+} // namespace BackboneRouter
 } // namespace ot
+
+#endif // OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
--- a/src/core/backbone_router/backbone_tmf.cpp
+++ b/src/core/backbone_router/backbone_tmf.cpp
@@ -66,7 +66,9 @@ bool BackboneTmfAgent::IsBackboneTmfMessage(const Ip6::MessageInfo &aMessageInfo
     //     2. All Domain BBRs (Link-Local scope)
     //     3. A Backbone Link-Local address
     // The source must be a Backbone Link-local address.
-    return (dst.IsLinkLocal() || dst.IsLinkLocalMulticast()) && src.IsLinkLocal();
+    return (Get<BackboneRouter::Local>().IsEnabled() && src.IsLinkLocal() &&
+            (dst.IsLinkLocal() || dst == Get<BackboneRouter::Local>().GetAllNetworkBackboneRoutersAddress() ||
+             dst == Get<BackboneRouter::Local>().GetAllDomainBackboneRoutersAddress()));
 }
 
 } // namespace BackboneRouter
--- a/src/core/backbone_router/backbone_tmf.hpp
+++ b/src/core/backbone_router/backbone_tmf.hpp
@@ -73,7 +73,7 @@ public:
     otError Start(void);
 
     /**
-     * This method returns whether Thread Management Framework Addressing Rules are met.
+     * This method returns whether @p aMessageInfo meets Backbone Thread Management Framework Addressing Rules.
      *
      * @retval true   Thread Management Framework Addressing Rules are met.
      * @retval false  Thread Management Framework Addressing Rules are not met.
--- a/src/core/backbone_router/backbone_tmf.hpp
+++ b/src/core/backbone_router/backbone_tmf.hpp
@@ -75,11 +75,11 @@ public:
     /**
      * This method returns whether Thread Management Framework Addressing Rules are met.
      *
-     * @retval TRUE   if Thread Management Framework Addressing Rules are met.
-     * @retval FALSE  if Thread Management Framework Addressing Rules are not met.
+     * @retval true   Thread Management Framework Addressing Rules are met.
+     * @retval false  Thread Management Framework Addressing Rules are not met.
      *
      */
-    static bool IsBackboneTmfMessage(const Ip6::MessageInfo &aMessageInfo);
+    bool IsBackboneTmfMessage(const Ip6::MessageInfo &aMessageInfo);
 
 private:
     static otError Filter(const ot::Coap::Message &aMessage, const Ip6::MessageInfo &aMessageInfo, void *aContext);
--- a/src/core/backbone_router/backbone_tmf.hpp
+++ b/src/core/backbone_router/backbone_tmf.hpp
@@ -79,7 +79,7 @@ public:
      * @retval false  Thread Management Framework Addressing Rules are not met.
      *
      */
-    bool IsBackboneTmfMessage(const Ip6::MessageInfo &aMessageInfo);
+    bool IsBackboneTmfMessage(const Ip6::MessageInfo &aMessageInfo) const;
 
 private:
     static otError Filter(const ot::Coap::Message &aMessage, const Ip6::MessageInfo &aMessageInfo, void *aContext);
--- a/src/core/backbone_router/backbone_tmf.hpp
+++ b/src/core/backbone_router/backbone_tmf.hpp
@@ -28,46 +28,64 @@
 
 /**
  * @file
- *   This file implements Thread Management Framework (TMF) functionalities.
+ *   This file includes definitions for Backbone TMF functionality.
  */
 
-#include "thread/tmf.hpp"
+#ifndef OT_CORE_THREAD_BACKBONE_TMF_HPP_
+#define OT_CORE_THREAD_BACKBONE_TMF_HPP_
 
-#include "common/locator-getters.hpp"
+#include "coap/coap.hpp"
 
 namespace ot {
-namespace Tmf {
+namespace BackboneRouter {
 
-otError TmfAgent::Start(void)
+enum
 {
-    return Coap::Start(kUdpPort);
-}
+    kBackboneUdpPort = 61631, ///< Backbone TMF UDP Port
+};
 
-otError TmfAgent::Filter(const ot::Coap::Message &aMessage, const Ip6::MessageInfo &aMessageInfo, void *aContext)
+/**
+ * This class implements functionality of the Backbone TMF agent.
+ *
+ */
+class BackboneTmfAgent : public Coap::Coap
 {
-    OT_UNUSED_VARIABLE(aMessage);
+public:
+    /**
+     * This constructor initializes the object.
+     *
+     * @param[in] aInstance      A reference to the OpenThread instance.
+     *
+     */
+    explicit BackboneTmfAgent(Instance &aInstance)
+        : Coap::Coap(aInstance)
+    {
+        SetInterceptor(&Filter, this);
+    }
 
-    return static_cast<TmfAgent *>(aContext)->IsTmfMessage(aMessageInfo) ? OT_ERROR_NONE : OT_ERROR_NOT_TMF;
-}
+    /**
+     * This method starts the TMF agent.
+     *
+     * @retval OT_ERROR_NONE    Successfully started the CoAP service.
+     * @retval OT_ERROR_ALREADY Already started.
+     *
+     */
+    otError Start(void);
 
-bool TmfAgent::IsTmfMessage(const Ip6::MessageInfo &aMessageInfo) const
-{
-    bool rval = true;
+    /**
+     * This method returns whether Thread Management Framework Addressing Rules are met.
+     *
+     * @retval TRUE   if Thread Management Framework Addressing Rules are met.
+     * @retval FALSE  if Thread Management Framework Addressing Rules are not met.
+     *
+     */
+    static bool IsBackboneTmfMessage(const Ip6::MessageInfo &aMessageInfo);
 
-    // A TMF message must comply with following rules:
-    // 1. The destination is a Mesh Local Address or a Link-Local Multicast Address or a Realm-Local Multicast Address,
-    //    and the source is a Mesh Local Address. Or
-    // 2. Both the destination and the source are Link-Local Addresses.
-    VerifyOrExit(
-        ((Get<Mle::MleRouter>().IsMeshLocalAddress(aMessageInfo.GetSockAddr()) ||
-          aMessageInfo.GetSockAddr().IsLinkLocalMulticast() || aMessageInfo.GetSockAddr().IsRealmLocalMulticast()) &&
-         Get<Mle::MleRouter>().IsMeshLocalAddress(aMessageInfo.GetPeerAddr())) ||
-            ((aMessageInfo.GetSockAddr().IsLinkLocal() || aMessageInfo.GetSockAddr().IsLinkLocalMulticast()) &&
-             aMessageInfo.GetPeerAddr().IsLinkLocal()),
-        rval = false);
-exit:
-    return rval;
-}
+private:
+    static otError Filter(const ot::Coap::Message &aMessage, const Ip6::MessageInfo &aMessageInfo, void *aContext);
+};
 
-} // namespace Tmf
+} // namespace BackboneRouter
 } // namespace ot
+
+#endif //  OT_CORE_THREAD_BACKBONE_TMF_HPP_
--- a/src/core/backbone_router/bbr_manager.cpp
+++ b/src/core/backbone_router/bbr_manager.cpp
@@ -56,6 +56,7 @@ Manager::Manager(Instance &aInstance)
     , mNdProxyTable(aInstance)
     , mMulticastListenersTable(aInstance)
     , mTimer(aInstance, Manager::HandleTimer, this)
+    , mBackboneTmfAgent(aInstance)
 #if OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE
     , mDuaResponseStatus(ThreadStatusTlv::kDuaSuccess)
     , mMlrResponseStatus(ThreadStatusTlv::kMlrSuccess)
@@ -67,6 +68,8 @@ Manager::Manager(Instance &aInstance)
 
 void Manager::HandleNotifierEvents(Events aEvents)
 {
+    otError error;
+
     if (aEvents.Contains(kEventThreadBackboneRouterStateChanged))
     {
         if (Get<BackboneRouter::Local>().GetState() == OT_BACKBONE_ROUTER_STATE_DISABLED)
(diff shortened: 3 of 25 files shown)


simonlingoogle:
Labeld as `P2` because many fore-coming 1.2 Backbone features including MLR, DUA, CCM requires this backbone TMF feature. @wgtdkp @librasungirl @bukepo @jwhui @abtink

src/core/thread/tmf.hpp
```diff
@@ -57,8 +57,9 @@ class TmfAgent : public Coap::Coap
      * @param[in] aInstance      A reference to the OpenThread instance.
      *
      */
-    explicit TmfAgent(Instance &aInstance)
+    explicit TmfAgent(Instance &aInstance, otNetifIdentifier aNetifIdentifier)
```
bukepo, line 60: Is it possible to let this class inherit CoapBase and managing two sockets, one for Thread interface and one for Backbone link interface. I guess this will save some resources.

simonlingoogle: It will save some RAM resource but the memory saved should be trivial for backbone routers which tends to have megabytes of RAM. 
With one COAP handling two sockets, we are going to make the COAP logic more complex with increased code size (e.g. choosing UDP sockets for sending/receiving messages based on message attributes)
Also with the security issue in mind, we will have to determine whether a resource should handle Backbone requests. 
I would prefer to use separate COAP objects since it's simpler. 
Thoughts?

src/core/thread/tmf.hpp
```diff
@@ -82,8 +83,10 @@ class TmfAgent : public Coap::Coap
     bool IsTmfMessage(const Ip6::MessageInfo &aMessageInfo) const;
 
 private:
-    otError        Start(uint16_t aPort);
+    otError        Start(uint16_t aPort, otNetifIdentifier aNetifIdentifier);
     static otError Filter(const ot::Coap::Message &aMessage, const Ip6::MessageInfo &aMessageInfo, void *aContext);
+
+    otNetifIdentifier mNetifIdentifier;
```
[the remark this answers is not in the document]
simonlingoogle: Yes it can. I don't have strong opinion on this and there is no significant difference. 

If it's a parameter of `Start()`, we need to change `GetInstance().GetBackboneTmfAgent().Start()` to `GetInstance().GetBackboneTmfAgent().Start(OT_NETIF_BACKBONE)`. It seems a little redundant because `GetInstance().GetBackboneTmfAgent().Start(OT_NETIF_THREAD)` is illegal.

src/core/net/udp6.hpp
```diff
@@ -123,6 +123,10 @@ class Udp : public InstanceLocator
         {
             mHandler(mContext, &aMessage, &aMessageInfo);
         }
+
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+        bool mBoundToThreadStack;
```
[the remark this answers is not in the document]
simonlingoogle: Yes. Can't pass checks without this. 

The problem was on Backbone Routers. Since a Backbone Router has two Udp Sockets both bound to port 61631, it does not know which one is the correct receiver for a received UDP message. This flag thus help Backbone Router to skip the Udp sockets bound to Backbone interface that are not supposed to process any UDP messages received from the Thread stack.

src/core/thread/tmf.hpp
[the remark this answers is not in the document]
bukepo: I think we should have separate class for BackboneTmfAgent, which not only has a different port and netif, but also different filter rules. Thoughts?

src/core/net/udp6.hpp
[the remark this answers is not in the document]
bukepo: This adds the overhead to all sockets. As `MessageInfo::mIstHostInterface` tells whether this message comes from Thread network interface, is it possible to add special handling for the backbone tmf port here?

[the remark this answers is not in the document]
simonlingoogle: This adds the overhead to all sockets, **but only on Backbone Routers**. Note that Backbone Routers are usually way more powerful than normal SOC devices. 

The problem is in the UDP socket matching where UDP sockets know nothing about up layer COAP objects. It does not know which Udp Socket should receive a UDP message if they share the same port. We can add special handling to skip the TMF socket by comparing Socket object references. 

However I think we should not keep adding hardcoding and special handling for Netif binding issues, but to accept that OT core should be aware of different network interfaces (at least on Backbone Routers).

src/core/thread/tmf.hpp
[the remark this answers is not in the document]
simonlingoogle: Note that the default port for Backbone TMF is 61631, same as the Thread TMF port. So, by default they share the same port number. 

I haven't investigated the filter rules. If they differ for Thread TMF and Backbone TMF, I would agree that it's better to have separate classes. 
But I think it's not the major issue. I will investigate and revise accordingly.

src/core/net/udp6.hpp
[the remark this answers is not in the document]
bukepo: Since this field is only added into the internal class, I'm wondering if this will affect upper layer users who don't have such internal classes from C API. `otUdpSocket` is directly used in the CLI library, will it cause issues when being cast to the internal `SocketHandle`?

src/core/thread/tmf.hpp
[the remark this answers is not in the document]
bukepo: I guess the backbone TMF should only allow link local scope destinations?

src/core/net/udp6.hpp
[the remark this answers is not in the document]
simonlingoogle: This is a good point. Adding `bool mBoundToThreadStack` to `otUdpSocket` brings unnecessary cost that should be avoided.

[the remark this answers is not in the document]
bukepo: How about managing sockets segment by segment.

```bash

socket1 -> socket2 - > ... -> socket[n] -> socket[n+1] -> ...
     ^                            ^
     |                            |
Head of all sockets(not bound)  First backbone link socket
```

src/core/common/linked_list.hpp
```diff
@@ -409,13 +445,17 @@ template <typename Type> class LinkedList
      * @returns A pointer to the matching entry if one is found, or nullptr if no matching entry was found.
      *
      */
-    template <typename Inidcator> const Type *FindMatching(const Inidcator &aIndicator, const Type *&aPrevEntry) const
+    template <typename Inidcator>
+    const Type *FindMatching(const Type *     aBegin,
+                             const Type *     aEnd,
```
bukepo, line 450: Avoid touching unrelated code by making `aEnd` optional?

src/posix/platform/udp.cpp
```diff
@@ -276,9 +276,6 @@ otError otPlatUdpBind(otUdpSocket *aUdpSocket)
         VerifyOrExit(0 == setsockopt(fd, IPPROTO_IPV6, IPV6_RECVPKTINFO, &on, sizeof(on)), error = OT_ERROR_FAILED);
     }
 
-    VerifyOrExit(0 == setsockopt(fd, IPPROTO_IPV6, IPV6_MULTICAST_IF, &gNetifIndex, sizeof(gNetifIndex)),
```
simonlingoogle, line 279: Some explanation for removing this line of code:

- For sockets that bound to Thread network interface, this line of code is useless since the multicast network interface is already fixed to Thread interface only. 
- For sockets that bound to Backbone network interface, this line of code is wrong (will always fail) and thus should be removed. 
- For sockets that bound to unspecified (all) interfaces (e.g. mApplicationCoap), it could be a security pitfall to bind `MULTICAST_IF` to Thread interface because we might accidentally send multicast in the Thread network. I prefer to add `MUTLICAST_IF` configuration when we really need it.

src/core/net/udp6.cpp
```diff
@@ -205,6 +221,35 @@ otError Udp::Bind(SocketHandle &aSocket, const SockAddr &aSockAddr)
...
+
+    if (mPrevBackboneSockets != nullptr)
+    {
+        mSockets.PushAfter(aSocket, *mPrevBackboneSockets);
+    }
+    else
+    {
+        IgnoreError(mSockets.Add(aSocket));
+    }
+}
```
abtink, line 245: I think it may be actually much easier and a lot cleaner to use two linked-lists, one as before and one for `BackboneSocket`. This makes all the logic in the core for this much simpler.

I think you mentioned the only limitation is with public `otUdpGetSockets()` behavior.

How about adding a new public API that gets the list of backbone sockets in addition to `otUdpGetSockets()` which would not include the backbone sockets.

We can also later change the public API to an iterator based API to be more flexible.

src/core/thread/tmf.hpp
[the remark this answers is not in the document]
simonlingoogle: this is resolved since there is now a separate class `BackboneTmfAgent` with customized TMF filter rules.

src/core/net/udp6.hpp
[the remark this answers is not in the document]
simonlingoogle: segmented linked list implemented. Please check.

src/core/common/linked_list.hpp
```diff
@@ -409,13 +453,17 @@ template <typename Type> class LinkedList
...
+                             const Inidcator &aIndicator,
+                             const Type *&    aPrevEntry) const
```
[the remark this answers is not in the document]
simonlingoogle: Thanks for the two approaches, both of which should work. 
We have thought about `Approach 1` but decided to add a ranged `FindMatching` method to `LinkedList`:
- It's a valid use case to search in a given range of linked list. The method might also be useful for other purposes. 
- It does not increase any code size  if ranged `FindMatching` is not used. (from the size-report we can find that the overall code size actually decreased by 16B)

For `approach 2`, I think it will increase code size so it does not seem to be the best choice. 

Thoughts?

src/core/common/linked_list.hpp
```diff
@@ -348,33 +348,74 @@ template <typename Type> class LinkedList
     /**
      * This method searches within the linked list to find an entry and if found returns a pointer to previous entry.
      *
-     * @param[in]  aEntry      A reference to an entry to find.
+     * @param[in]  aEntry      A pointer to an entry to find. If nullptr, this method finds the tail of the list.
```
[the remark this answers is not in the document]
simonlingoogle: You are right. 
I was too focused on recalculating `mPrefBackboneSockets` using `mSockets.Find(backboneSockets, mPrevBackboneSockets))`.

src/core/net/udp6.cpp
simonlingoogle: The segmented linked list just works. It's largest advantage is that it does not affect the public API at all.
I would prefer to switch to use multiple linked lists of sockets when it's really required.

Thoughts? @abtink @bukepo

bukepo: It seems we are using the segmented list for managing addresses. I like its simplicity for implementation and also allows providing simpler public APIs(introduce another iterator).

abtink: Using segmented lists sounds fine to me.

src/core/common/linked_list.hpp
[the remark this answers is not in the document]
abtink: SGTM. Thanks.

src/core/net/udp6.cpp
```diff
@@ -205,6 +221,35 @@ otError Udp::Bind(SocketHandle &aSocket, const SockAddr &aSockAddr)
...
+{
+    RemoveSocket(aSocket);
+
+    if (mPrevBackboneSockets != nullptr)
+    {
+        mSockets.PushAfter(aSocket, *mPrevBackboneSockets);
+    }
+    else
+    {
+        IgnoreError(mSockets.Add(aSocket));
```
abtink, line 243: Use  `mSocket.Push(aSocket)` here?

`Add()` checks for it not to be on the list already before adding/pushing it to the list but since we know it's not in the list (we did `RemoveSocke()` already) we can use `Push` (which does not check).

src/core/net/udp6.cpp
```diff
@@ -297,12 +342,40 @@ otError Udp::SendTo(SocketHandle &aSocket, Message &aMessage, const MessageInfo
...
+
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+    if (&aSocket == mPrevBackboneSockets)
+    {
+        SocketHandle *backboneSockets = aSocket.GetNext();
+        otError       error;
+
+        if (backboneSockets != nullptr)
+        {
+            IgnoreError(error = mSockets.Find(*backboneSockets, mPrevBackboneSockets));
```
abtink, line 369: Remove `IgnoreError()`?

src/core/net/udp6.cpp
```diff
@@ -297,12 +342,40 @@ otError Udp::SendTo(SocketHandle &aSocket, Message &aMessage, const MessageInfo
...
+        mPrevBackboneSockets = &aSocket;
+    }
+#endif
+exit:
+    return;
 }
 
 void Udp::RemoveSocket(SocketHandle &aSocket)
 {
     SuccessOrExit(mSockets.Remove(aSocket));
```
abtink, line 359: I think we can do a small optimization here.
 
If instead of `Remove()` we first search for entry in list (using `Find()`) we get the `prev` entry before it, we can then remove it using `PopAfter()` and use it to see update `mPrevBackboneSockets` (if needed).
```
     SocketHandle *prev;

     SuccessOrExit(mSockets.Find(aSocket , prev));
     mSocket.PopAfter(prev);
     
 #if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
     if (&aSocket == mPrevBackboneSockets)
     {
           mPrevBackboneSockets = prev;
     }
#endif 
```

src/core/thread/tmf.hpp
simonlingoogle: I think we can refactor in the future if `one COAP base two sockets` is proven to be a better idea.

src/core/net/udp6.cpp
simonlingoogle: If we remove `IgnoreError()` , we need to add `OT_UNUSED_VARIABLE(error)` (note that `OT_ASSERT` can be turned off). 
I am fine with both, but `IgnoreError()` reduced one line of code.

src/core/net/udp6.cpp
simonlingoogle: I didn't put much thought into it. The code you suggested is indeed more efficient. Thanks.

src/posix/platform/udp.cpp
bukepo: I'm fine get back to this issue later.

src/core/net/udp6.cpp
simonlingoogle: revised

src/core/net/udp6.cpp
simonlingoogle: **Just for discussion:** 
If we have a method like 
```c++
static inline void OT_ASSERT_OK(otError aError) 
{
    OT_UNUSED_VARIABLE(aError);
    OT_ASSERT(aError == OT_ERROR_NONE);
}
```
 , we can avoid using `IgnoreError` or `OT_UNUSED_VARIABLE`. It's quite easy (and a little annoying) to miss these statements and get build failures.

simonlingoogle:
**Beyond the scope of this PR:**

I found that allowing `Find()` of linked list to get the tail with `nullptr` seems even reduce the code size. 
**I tend to think that using `nullptr` to get the tail make sense since the next of tail is indeed `nullptr`.**
Then we can implement `GetTail()` by using `Find()` directly, which should further reduce code size. 

Thought? @abtink

abtink:
> Beyond the scope of this PR:
> 
> I found that allowing Find() of linked list to get the tail with nullptr seems even reduce the code size.
> I tend to think that using nullptr to get the tail make sense since the next of tail is indeed nullptr.
> Then we can implement GetTail() by using Find() directly, which should further reduce code size.
> 
> Thought? @abtink 

Sounds reasonable to me. For a future PR. 
I would suggest that we define the version of `Find()` which uses a pointer as input as `private` and use it from different places (e.g. from `GetTail()` and the public `Find()`). This way the public methods are clearly defined (not complex) but common code is shared privately.

src/core/net/udp6.cpp
abtink: I don' think that would work on all toolchains (the `OT_UNUSED_VARIABLE` can apply to local parameter variable `aError` and not the caller one). 

The trickiness in general is that `OT_UNUSED_VARIABLE` is tool-chain dependent.

 I would recommend using `OT_UNUSED_VARIABLE` directly to silence the unused-var warning from such `OT_ASEERT()` use and not use `IgnoreError` for this. `IgnoreError` is to ensure returned value is not ignored and it just happens to work in this case to use the variable (silence the warning), but may not work in all situations and toolchains.

simonlingoogle: **Note that the code we are commenting is already removed, so resolving this.**

src/core/backbone_router/backbone_tmf.cpp
```diff
@@ -0,0 +1,75 @@
...
+    const Ip6::Address &dst = aMessageInfo.GetSockAddr();
+    const Ip6::Address &src = aMessageInfo.GetPeerAddr();
+
+    // A Backbone TMF message must comply with following rules:
+    // The destination must be one of:
+    //     1. All Network BBRs (Link-Local scope)
+    //     2. All Domain BBRs (Link-Local scope)
+    //     3. A Backbone Link-Local address
+    // The source must be a Backbone Link-local address.
+    return (dst.IsLinkLocal() || dst.IsLinkLocalMulticast()) && src.IsLinkLocal();
```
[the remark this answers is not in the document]
simonlingoogle: I was thinking about this too, but thought it might not matter that much. 
 
Note that if the code here runs, the destination of the message **MUST** be a subscribed Multicast Address on the Backbone interface. The other possible Multicast Addresses are `Solicited-node Multicast Addresses` which are also link-local. 
We should never receive TMF messages with `Solicited-node Multicast Addresses` as destinations. But even if we did receive a message like that, it does not harm to process it? 

Thoughts?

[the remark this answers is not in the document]
librasungirl: If it doesn't cost much to be accurate, why leave it vague?
I may prefer a bit to update the code to be consistent with the rules `1.`, `2.` you defined in the comments.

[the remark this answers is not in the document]
simonlingoogle: revised accordingly.

src/core/backbone_router/bbr_manager.cpp
```diff
@@ -86,6 +89,33 @@ void Manager::HandleNotifierEvents(Events aEvents)
                 mTimer.Start(kTimerInterval);
             }
         }
+
+        if (Get<BackboneRouter::Local>().GetState() == OT_BACKBONE_ROUTER_STATE_PRIMARY)
```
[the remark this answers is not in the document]
simonlingoogle: revised accordingly.

src/core/backbone_router/backbone_tmf.hpp
```diff
@@ -0,0 +1,91 @@
...
+     * @retval OT_ERROR_ALREADY Already started.
+     *
+     */
+    otError Start(void);
+
+    /**
+     * This method returns whether Thread Management Framework Addressing Rules are met.
+     *
+     * @retval TRUE   if Thread Management Framework Addressing Rules are met.
+     * @retval FALSE  if Thread Management Framework Addressing Rules are not met.
```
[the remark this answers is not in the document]
simonlingoogle: revised accordingly.

bukepo (approved):
LGTM:+1:

src/core/net/udp6.cpp
```diff
@@ -156,7 +172,10 @@ otError Udp::RemoveReceiver(Receiver &aReceiver)
 
 otError Udp::Open(SocketHandle &aSocket, otUdpReceive aHandler, void *aContext)
 {
-    otError error = OT_ERROR_NONE;
+    otError       error = OT_ERROR_NONE;
+    SocketHandle *prev;
+
+    VerifyOrExit(mSockets.Find(aSocket, prev) == OT_ERROR_NOT_FOUND, error = OT_ERROR_ALREADY);
```
simonlingoogle, line 178: Turn out `Udp::Open` is not **reentrant**. 
Currently `BackboneTmf` agent would call `Coap::Start` when BBR primary state changes, which in turn calls `Udp::Open`.
So I make `Udp::Open` **reentrant** as the code shows. 
@librasungirl  @abtink @bukepo

src/core/coap/coap.cpp
```diff
@@ -1020,9 +1020,11 @@ Coap::Coap(Instance &aInstance)
 
 otError Coap::Start(uint16_t aPort, otNetifIdentifier aNetifIdentifier)
 {
-    otError error;
+    otError error        = OT_ERROR_NONE;
     bool    socketOpened = false;
 
+    VerifyOrExit(!mSocket.IsBound(), OT_NOOP);
```
librasungirl, line 1026: According to the available return values in the header file,  `OT_ERROR_ALREADY` might be more appropriate here?

https://github.com/openthread/openthread/blob/25ae9b073955a4fa00d8b0b2ed368c1ac77fec8f/src/core/coap/coap.hpp#L607-L616

librasungirl (approved):
👍  LGTM with one minor inline comment

src/posix/platform/system.cpp
```diff
@@ -63,6 +63,8 @@ otInstance *otSysInit(otPlatformConfig *aPlatformConfig)
     instance = otInstanceInitSingle();
     assert(instance != nullptr);
 
+    platformBackboneInit(instance, aPlatformConfig->mBackboneInterfaceName);
```
librasungirl, line 66: add `OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE` condition?

src/posix/platform/backbone.cpp
```diff
@@ -0,0 +1,54 @@
...
+    VerifyOrExit(aInterfaceName != nullptr, OT_NOOP);
+
+    VerifyOrDie(strnlen(aInterfaceName, IFNAMSIZ) <= IFNAMSIZ - 1, OT_EXIT_INVALID_ARGUMENTS);
+    strcpy(gBackboneNetifName, aInterfaceName);
+
+    gBackboneNetifIndex = if_nametoindex(gBackboneNetifName);
+    VerifyOrDie(gBackboneNetifIndex > 0, OT_EXIT_FAILURE);
+exit:
+    return;
+}
```
librasungirl, line 54: add `OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE` condition?

src/posix/platform/platform-posix.h
```diff
@@ -395,6 +395,15 @@ enum SocketBlockOption
...
 int SocketWithCloseExec(int aDomain, int aType, int aProtocol, SocketBlockOption aBlockOption);
 
+/**
+ * This function initializes platform Backbone network.
+ *
+ * @param[in]   aInstance       A pointer to the OpenThread instance.
+ * @param[in]   aInterfaceName  A pointer to Thread network interface name.
+ *
+ */
+void platformBackboneInit(otInstance *aInstance, const char *aInterfaceName);
```
librasungirl, line 405: add `OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE` condition?

src/core/coap/coap.cpp
simonlingoogle: Agree that it's better for this PR to return `OT_ERROR_ALREADY`. 

But I think in general we should avoid using `OT_ERROR_ALREADY`. We can refactor it in a separate PR.

src/posix/platform/system.cpp
simonlingoogle: It's actually not a must. 
- It's not a strong requirement of posix platform code to reduce the code size, so we can have this code running even when it's not a Backbone Router. 
- Further more, it's possible (and perfectly fine) that other COAP objects (other than Backbone TMF agent) might need to bind on the Backbone interface.

I don't have a strong opinion for it, but I also think it's fine to keep the code available on non Backbone Routers. 

Thoughts? @librasungirl

src/core/backbone_router/backbone_tmf.hpp
```diff
@@ -0,0 +1,91 @@
...
+    otError Start(void);
+
+    /**
+     * This method returns whether Thread Management Framework Addressing Rules are met.
+     *
+     * @retval true   Thread Management Framework Addressing Rules are met.
+     * @retval false  Thread Management Framework Addressing Rules are not met.
+     *
+     */
+    bool IsBackboneTmfMessage(const Ip6::MessageInfo &aMessageInfo);
```
jwhui, line 82: Declare as const?

src/posix/platform/system.cpp
librasungirl: I may prefer to add the macro condition: 
- follow the same style as the others, like `platformNetifProcess()` has `OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE` wrapper
- As similar rule in another PR https://github.com/openthread/openthread/pull/5346#discussion_r475161377 agreed -better to only add required functionalities unless we have clear use case for the `possible (and perfectly fine) that other COAP objects (other than Backbone TMF agent) might need to bind on the Backbone interface`, though we don't concern code size on posix so much as in embedded platform.

src/core/backbone_router/backbone_tmf.hpp
```diff
@@ -0,0 +1,91 @@
...
+     * This method starts the TMF agent.
+     *
+     * @retval OT_ERROR_NONE    Successfully started the CoAP service.
+     * @retval OT_ERROR_ALREADY Already started.
+     *
+     */
+    otError Start(void);
+
+    /**
+     * This method returns whether Thread Management Framework Addressing Rules are met.
```
Irving-cl, line 76:
```suggestion
     * This method returns whether @p aMessageInfo meets Thread Management Framework Addressing Rules.
```
What about saying like this?

src/core/common/instance.hpp
```diff
@@ -763,10 +763,16 @@ template <> inline BackboneRouter::MulticastListenersTable &Instance::Get(void)
...
+
 template <> inline BackboneRouter::NdProxyTable &Instance::Get(void)
 {
     return mThreadNetif.mBackboneRouterManager.GetNdProxyTable();
 }
+
+template <> inline BackboneRouter::BackboneTmfAgent &Instance::Get(void)
+{
+    return mThreadNetif.mBackboneRouterManager.GetBackboneTmfAgent();
+}
```
Irving-cl, line 775: This seems not used anywhere.

src/core/net/udp6.cpp
```diff
@@ -99,13 +99,26 @@ otError Udp::Socket::Bind(uint16_t aPort)
...
-}
+    OT_UNUSED_VARIABLE(aNetifIdentifier);
+
+    otError error = OT_ERROR_NONE;
+
+#if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
+    SuccessOrExit(error = otPlatUdpBindToNetif(this, aNetifIdentifier));
+#endif
+
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
```
Irving-cl, line 112: Change to `#if OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE` and add config check to see whether this is enabled when `OPENTHREAD_FTD` is not enabled in `config-check.h`?

Example: https://github.com/openthread/openthread/blob/master/src/core/config/openthread-core-config-check.h#L526

src/core/net/udp6.cpp
```diff
@@ -296,14 +341,34 @@ otError Udp::SendTo(SocketHandle &aSocket, Message &aMessage, const MessageInfo
...
 {
-    SuccessOrExit(mSockets.Remove(aSocket));
+    SocketHandle *prev;
+
+    SuccessOrExit(mSockets.Find(aSocket, prev));
+
+    mSockets.PopAfter(prev);
     aSocket.SetNext(nullptr);
 
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
```
Irving-cl, line 365: The same as mentioned above.

src/core/net/udp6.cpp
```diff
@@ -397,7 +462,27 @@ void Udp::HandlePayload(Message &aMessage, MessageInfo &aMessageInfo)
     SocketHandle *socket;
     SocketHandle *prev;
 
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
```
Irving-cl, line 465: The same as mentioned above.

src/core/common/instance.hpp
simonlingoogle: It will be used soon in #5388

src/core/net/udp6.cpp
simonlingoogle: This is indeed a good idea. But I prefer to do it in a separate PR because there are so many places that are using `#ifdef OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE` and they should all be fixed (but unrelated to this PR).

Irving-cl: SGTM.

Irving-cl (approved):
LGTM👍

jwhui (approved):

PR #5475, opened 2020-08-29
[sub-mac] do not assert when parsing raw-link frames

abtink:
This commit relaxes the assert checks when parsing raw-link frame from
the radio tx-done callback to update the security frame counter. In an
FTD/MTD build, if/when raw-link is enabled, the `TxFrame` is prepared
and given by user and may not necessarily follow 15.4 frame format
(raw-link can be used with vendor-specific format), so we allow
failure when parsing the frame. In other cases (in an RCP build or in
an FTD/MTD build without raw-link) since the `TxFrame` should be
prepared by OpenThread core, we expect no error and therefore assert
if parsing fails.

-------

This PR should help address an issue discovered by oss-fuzz testing.


bukepo (approved):

brianljt (approved):
LGTM. Thanks!

PR #5482, opened 2020-09-02
[coap] misc enhancements

abtink:
This commit contains a group of smaller changes/renames in `Coap` modules:
- Add enumeration (c++ style) constants for CoAP `Type` and `Code` and `OptionType`.
- Add helpers `Message::Is{Method}Request()` e.g., `IsPostRequest()`
- Define Thread URI Paths (in `uri_paths.hpp/cpp`) as char array.
- Add helper method in `Coap::Message` for common initializations,  e.g., `InitAsConfirmablePost()`.
- Add helper methods for common `Coap:Message` type and code checks, e.g., `IsConfirmablePost()`.
- Re-define `OT_COAP_TYPE` values to use 2-bit unsigned int values (per RFC-7252).


src/core/thread/uri_paths.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include "openthread-core-config.h"
+
+namespace ot {
+
+/**
+ *
+ * This structure contains Thread URI Path string definitions.
+ *
+ */
+struct UriPath
```
wgtdkp, line 46: What about using  `namespace UriPath`? Seems there is no necessary to define a `struct` with no fields.

src/core/thread/uri_paths.hpp
```diff
@@ -0,0 +1,89 @@
...
+{
+    static const char kAddressQuery[];           ///< The URI Path for Address Query ("a/aq").
```
wgtdkp, line 48: By changing `struct UriPath` to `namespace UriPath`, I think we can defines the value directly at here:
```suggestion
    static const char kAddressQuery[] = "a/aq";
```

src/core/thread/uri_paths.hpp
abtink: I did not go for `namespace` for two reasons:
- First we generally use `namespace` in OpenThread for larger components, grouping of multiple related modules (e.g, `Mac`, `Mle`, `Coap`, etc). I don't feel that `UriPath` can be considered a compont.
- Second is for future use (I did not add it in this PR) of `UriPath` itself to represent a fixed-length URI path as a type, along with some helper methods.

src/core/thread/uri_paths.hpp
abtink: Yes. This can work/compile (with or even without `static` since `const` definitions should have "internal linkage" model - I think after C++11), but I intentionally avoided this pattern:
- These array definitions are in a header file which is included by a bunch of `cpp` files.
- With this we will have the same const variable array definitions in each `cpp` file (scoped to that file).
- While most toolchain would optimize this while linking and will have a single common version of the array string, I don't think this is necessarily required or ensured (it's tool-chain dependent).
- The code in this PR is explicit about this and ensures that we will have a single instance of each array which is shared by all users.

src/core/meshcop/panid_query_client.cpp
```diff
@@ -71,9 +71,8 @@ otError PanIdQueryClient::SendQuery(uint16_t                            aPanId,
     VerifyOrExit(Get<MeshCoP::Commissioner>().IsActive(), error = OT_ERROR_INVALID_STATE);
     VerifyOrExit((message = MeshCoP::NewMeshCoPMessage(Get<Tmf::TmfAgent>())) != nullptr, error = OT_ERROR_NO_BUFS);
 
-    SuccessOrExit(error =
-                      message->Init(aAddress.IsMulticast() ? OT_COAP_TYPE_NON_CONFIRMABLE : OT_COAP_TYPE_CONFIRMABLE,
-                                    OT_COAP_CODE_POST, OT_URI_PATH_PANID_QUERY));
+    SuccessOrExit(error = message->Init(aAddress.IsMulticast() ? Coap::kTypeNonConfirmable : Coap::kTypeConfirmable,
```
simonlingoogle, line 74: There are 3 places that `Init` message according to if the address is multicast. Worth to define a method for that?

simonlingoogle (approved):
LGTM 👍 

Great to have code size reduced.

src/core/meshcop/panid_query_client.cpp
abtink: Good idea. I noticed the same but was at first thinking that passing the `aAddress` to the `Message::Init()` may not be good/clear. Thinking more about it, I feel it can be fine (it's the destination based on which the CoAP type (confirmable vs non-confirmable is determined).

wgtdkp (approved):
LGTM 👍

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #5486, opened 2020-09-02
[thread-cert] fix case 6.1.9 random fail

FiveDimensions:


Cert_6_1_09_EDSynchronization.json
```diff
@@ -0,0 +1,83 @@
+{
```
jwhui, line 1: Is adding this file (and pcap file below) intentional?

jwhui:
I noticed that 6.1.9 still failed with this PR.

Cert_6_1_09_EDSynchronization.json
FiveDimensions: sorry, wrong commit, deleted

FiveDimensions:
> I noticed that 6.1.9 still failed with this PR.

yes, works locally, looking into it

jwhui (approved):

tests/scripts/thread-cert/Cert_6_1_09_EDSynchronization.py
```diff
@@ -58,7 +58,8 @@ class Cert_6_1_9_EDSynchronization(thread_cert.TestCase):
         ED: {
             'name': 'ED',
             'panid': 0xface,
-            'whitelist': [LEADER, ROUTER1, ROUTER2, ROUTER3]
+            'router_upgrade_threshold': 0,
```
librasungirl, line 61: should use `routereligible disable`?

PR #5488, opened 2020-09-03
[mle] declare 'Header' as a private nested type of 'Mle'

abtink:
This commit declares MLE `Header` as a `private` nested type of `Mle`
class. It also  moves the `Command` enumeration definition into the
`Mle` class (from `Header`).

### DIFF
--- a/src/core/thread/mle.hpp
+++ b/src/core/thread/mle.hpp
@@ -85,224 +85,6 @@ namespace Mle {
  */
 
 /**
- * This class implements MLE Header generation and parsing.
- *
- */
-OT_TOOL_PACKED_BEGIN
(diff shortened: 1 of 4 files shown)


src/core/thread/mle.hpp
```diff
@@ -934,6 +716,33 @@ class Mle : public InstanceLocator
     bool HasRestored(void) const { return mHasRestored; }
 
 protected:
+    /**
+     * MLE Command Types.
+     *
+     */
+    enum Command
```
jwhui, line 723:
```suggestion
    enum Command : uint8_t
```

jwhui (approved):

PR #4586, opened 2020-02-20
[efr32] Radio: don't abort current operation on switching to idle

dismirlian:
On the RAIL callback, received packets are held and passed to OT core in the event loop.

On a sleepy device, if there is nothing more to do, the OT core calls `otPlatRadioSleep` which in turn calls `RAIL_Idle(gRailHandle, RAIL_IDLE_ABORT, true)`.

If the time between `RAIL_EVENT_RX_PACKET_RECEIVED` and the call to `RAIL_Idle` is less than the turnaround + the ACK tx time, the call to `RAIL_Idle` aborts the auto-ack that was being transmitted by the radio, so there are received packets without ACK, like this:

image

<pre>
[0012727105] [WARN]-PLAT----: RX_ACK_RECEIVED
[0012727111] [WARN]-PLAT----: RX_PACKET_RECEIVED
[0012727112] [WARN]-PLAT----: <b>TXACK_ABORTED</b>
[0012727998] [WARN]-PLAT----: RX_ACK_RECEIVED
[0012728008] [WARN]-PLAT----: RX_ACK_RECEIVED
[0012728019] [WARN]-PLAT----: RX_PACKET_RECEIVED
[0012728020] [WARN]-PLAT----: TXACK_PACKET_SENT
[0012728030] [WARN]-PLAT----: RX_PACKET_RECEIVED
[0012728030] [WARN]-PLAT----: TXACK_PACKET_SENT
[0012728042] [WARN]-PLAT----: RX_PACKET_RECEIVED
[0012728042] [WARN]-PLAT----: TXACK_PACKET_SENT
[0012728055] [WARN]-PLAT----: RX_ACK_RECEIVED
[0012728153] [WARN]-PLAT----: RX_ACK_RECEIVED
[0012728159] [WARN]-PLAT----: RX_PACKET_RECEIVED
[0012728160] [WARN]-PLAT----: <b>TXACK_ABORTED</b>
[0012736995] [WARN]-PLAT----: RX_ACK_RECEIVED
</pre>

This PR changes that call so that `RAIL_Idle` waits for the last operation to finish. With this change, the ACK is always sent.

PR #5494, opened 2020-09-03
[efr32] Enhancements to EFR32MG21 radio

lmnotran:
When working on a recent Frame Pending issue observed with BRD4180A, it was observed that the radio implementation for efr32mg21 misses several enhancements, which were added to the efr32mg1x parts. We need to make sure that the supported features in radio.c align across all parts.

The following GitHub PRs captures these:
https://github.com/openthread/openthread/pull/4482
https://github.com/openthread/openthread/pull/4484
https://github.com/openthread/openthread/pull/4506

### DIFF
--- a/examples/platforms/efr32mg21/radio.c
+++ b/examples/platforms/efr32mg21/radio.c
@@ -55,22 +55,22 @@
 #include "rail_config.h"
 #include "rail_ieee802154.h"
 
-#define IEEE802154_MIN_LENGTH   5
-#define IEEE802154_MAX_LENGTH   127
-#define IEEE802154_ACK_LENGTH   5
+#define IEEE802154_MIN_LENGTH 5
+#define IEEE802154_MAX_LENGTH 127
+#define IEEE802154_ACK_LENGTH 5
 
 // FCF + DSN + dest PANID + dest addr + src PANID + src addr (without security header)
-#define IEEE802154_MAX_MHR_LENGTH           2 + 1 + 2 + 8 + 2 + 8
+#define IEEE802154_MAX_MHR_LENGTH 2 + 1 + 2 + 8 + 2 + 8
 
-#define IEEE802154_FRAME_TYPE_MASK          0x07
-#define IEEE802154_FRAME_TYPE_ACK           0x02
-#define IEEE802154_FRAME_TYPE_MAC_COMMAND   0x03
-#define IEEE802154_ACK_REQUEST              0x20
-#define IEEE802154_DSN_OFFSET               2
-#define IEEE802154_FCF_OFFSET               0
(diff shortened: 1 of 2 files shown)


dismirlian:
Hi, please also check #4586, I don't know if you've already considered it.

romacdon:
Marked this as a work-in-progress as there are some things we need to address including the point that @dismirlian made with regards to needing to include PR #4586.

examples/platforms/efr32mg21/radio.c
```diff
@@ -136,7 +136,7 @@ static const RAIL_IEEE802154_Config_t sRailIeee802154Config = {
     {
         // ackConfig
         true, // ackConfig.enable
-        894,  // ackConfig.ackTimeout
+        672,  // ackConfig.ackTimeout
```
dismirlian, line 139: Why is this 672us? Acoording to my understanding:

For beacon-enabled PANs:
macAckWaitDuration = aUnitBackoffPeriod + aTurnaroundTime + phySHRDuration + 6 × phySymbolsPerOctet = 54 symbols = 864us

For beacon-disabled PANs:
macAckWaitDuration = aTurnaroundTime + phySHRDuration + 6 × phySymbolsPerOctet = 34 symbols = 544us

schinchani: Hi @dismirlian 

802.15.4 defines its macAckWaitDuration from the end of the transmitted packet to complete reception of the ACK. RAIL's ackTimeout only covers sync word detection of the ACK. Therefore,  we have to subtract the ACK's PHY header and payload time to get RAIL's ackTimeout setting. For 2.4 GHz OQPSK, macAckWaitDuration is specified as 54 symbols; subtracting 2-symbol PHY header and 10-symbol payload yields a RAIL ackTimeout of 42 symbols or 672 microseconds at 16 microseconds/symbol.

This has also been documented in the Silabs RAIL document here:
https://docs.silabs.com/rail/latest/group-i-e-e-e802-15-4

lmnotran: https://docs.silabs.com/rail/latest/group-i-e-e-e802-15-4#details:~:text=802.15.4%20defines%20its%20macAckWaitDuration%20from%20the,or%20672%20microseconds%20at%2016%20microseconds%2Fsymbol.

dismirlian: Great! Thanks for your answers. I see this note was added in RAIL's docs >= 2.8, but I assume it's valid for the previous versions also.

So the EFR32xG1x platform requires an update, right?

schinchani: Yes, we understand that there are inconsistencies between the different platform implementations. We will are planning to align them soon.

PR #5495, opened 2020-09-03
[mle] further simplifications of logging in MLE

abtink:
This commit updates the logging in `Mle` modules by adding two
enumerations `MessageAction` and `MessageType` to indicate the action
(send, receive, delay, etc.) and the message type in the updated
`Log()` methods.

### DIFF
--- a/src/core/thread/discover_scanner.cpp
+++ b/src/core/thread/discover_scanner.cpp
@@ -154,7 +154,7 @@ otError DiscoverScanner::Discover(const Mac::ChannelMask &aScanChannels,
     mScanChannel = Mac::ChannelMask::kChannelIteratorFirst;
     mState       = (mScanChannels.GetNextChannel(mScanChannel) == OT_ERROR_NONE) ? kStateScanning : kStateScanDone;
 
-    Mle::Log("Send Discovery Request", destination);
+    Mle::Log(Mle::kMessageSend, Mle::kTypeDiscoveryRequest, destination);
 
 exit:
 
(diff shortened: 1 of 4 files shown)


simonlingoogle (approved):
LGTM 👍 

This is wonderful! Thanks.

simonlingoogle:
BTW, I think this can also be extended to more than just for MLE module. It's actually very common for other modules to print similar logs like `Failed to send/receive/process Xxxx Response: error`. Thoughts?

abtink:
Thanks @simonlingoogle. 😃 

> BTW, I think this can also be extended to more than just for MLE module. It's actually very common for other modules to print similar logs like Failed to send/receive/process Xxxx Response: error. Thoughts?

Yes, I agree. I think we can use similar model in other components as well (e.g., `MeshCop` modules seems like a good next candidate). I would say best way is for each component (set of modules) to define its own specific helper `Log()` methods suited for its needs.

simonlingoogle:
> Thanks @simonlingoogle. 😃 
> 
> 
> 
> > BTW, I think this can also be extended to more than just for MLE module. It's actually very common for other modules to print similar logs like Failed to send/receive/process Xxxx Response: error. Thoughts?
> 
> 
> 
> Yes, I agree. I think we can use similar model in other components as well (e.g., `MeshCop` modules seems like a good next candidate). I would say best way is for each component (set of modules) to define its own specific helper `Log()` methods suited for its needs. 

The problem with each component defining its own helper log is that they cannot share the same string constants, unless we provide a universal module for converting action/type to string.

librasungirl (approved):

jwhui (approved):

PR #5496, opened 2020-09-04
[k32w] use a custom Persistent Data Manager

doru91:
This commit add support for a custom PDM (Persistent Data Manager) which is also used inside Project CHIP.

Signed-off-by: Doru Gucea <doru-cristian.gucea@nxp.com>


doru91:
Hi, @jwhui, @librasungirl 

There are OT NS tests which are failing. However, the files touched by this pull-request are platform-related so I don't see why those tests (which should be stack-related) are failing. I also checked the OT NS logs from here (https://github.com/openthread/openthread/pull/5496/checks?check_run_id=1077504647) and the message error is "could not find factory to build UDP payload".

Thanks,
Doru

jwhui (approved):

PR #5499, opened 2020-09-08
[mac-frame] handle exception case for payload index of mac cmd

Irving-cl:
In current `Mac::Frame` implementation, when the frame is a Mac command, the `CommandId` is always not included as its **private payload**, which means we won't encrypt the `CommandId`.

According to _ieee802.15.4-2015_, 9.2.1:
> f) Secure the frame. For the frames specified in Table 9-1, the Private Payload field and Open Payload field shall be set as indicated in the table. For frames not specified in Table 9-1, the Private Payload shall be set to the MAC Payload field and Open Payload field shall be empty. The procedure shall then use the Private Payload field, the Open Payload field, the macExtendedAddress, the Frame Counter field (if TSCH is not being used), the ASN (if TSCH is being used), the SecurityLevel parameter, and the secKey element of the KeyDescriptor to produce the secured frame according to the CCM* transformation process defined in 9.3.4.

Table 9-1:
Screen Shot 2020-09-08 at 2 38 56 PM

Mac command frame format:
Screen Shot 2020-09-08 at 2 41 30 PM

As I understand, these things indicate that, in most cases the Private Payload(things we need to encrypt) shall be set to the MAC Payload field(For Mac cmd, this field includes CommandId, shown in Figure 7-17) and the Open Payload shall be empty. The only 2 exception cases are listed in Table 9-1. 

So for Mac command with version >= 2, its CommandId should be included in Private Payload and should be encrypted when necessary. For Mac command with version < 2, CommandId should be in included in Open Payload and shouldn't be encrypted.

This PR fixes this problem and adds unit tests for this.

### DIFF
--- a/script/test
+++ b/script/test
@@ -163,7 +163,6 @@ do_unit()
 
     if [[ ${THREAD_VERSION} == "1.2" ]]; then
         do_unit_version "1.2-bbr"
-        do_unit_version "1.1"
     fi
 }
 
--- a/src/core/mac/mac_frame.cpp
+++ b/src/core/mac/mac_frame.cpp
@@ -582,10 +582,9 @@ otError Frame::GetCommandId(uint8_t &aCommandId) const
 {
     otError error = OT_ERROR_NONE;
     uint8_t index = FindPayloadIndex();
-
     VerifyOrExit(index != kInvalidIndex, error = OT_ERROR_PARSE);
 
-    aCommandId = mPsdu[index - 1];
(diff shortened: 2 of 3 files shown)


brianljt (approved):
LGTM. Thanks!

bukepo (approved):
LGTM:+1:

abtink (approved):
LGTM. 👍

jwhui (approved):

PR #5501, opened 2020-09-08
[low-power] fix csl channel

Irving-cl:
Current implementation is a different of spec regarding CSL channel:
> 3.2.6.3.2    CSL Channel
> A CSL Receiver MAY configure macCslChannelMask to indicate an additional channel distinct from phyCurrentChannel. The macCslChannelMask is relayed by a CSL Receiver SSED Child to a CSL Transmitter Parent Router via the CSL Channel TLV sent in the Child Update Request as described in Section 4.6, Child Link Communication in Chapter 4, Mesh Link Establishment. A CSL Transmitter Parent Router MUST be capable of storing as many CSL Channel values as supported CSL Receiver SSED children.
> If configured, CSL data transmission and reception using the additional channel operates as described in Section 6.12.2.6 in [IEEE802154-2015]. The CSL Channel for a Thread Interface operating as a CSL Receiver or as a CSL Transmitter is either:
>  • The value of phyCurrentChannel when macCSLChannelMask has value 0.
>  • The channel value selected by macCSLChannelMask when macCSLChannelMask is different from 0.

Basically, if `CSL Channel` is never specified, the actual CSL channel should always be the phy channel that the device is using. If `CSL Channel` hasn't been specified and radio channel changes, the actual CSL channel should also change. In current implementation, `CSL Channel` is by default a valid value(not specified by upper layer) and wouldn't change when radio channel changes even if it's never specified.

This PR fixes this issue. On CSL receiver, a bit is extracted from `mCslChannel` to be used as a flag `mIsCslChannelSpecified`. If it is `false`, `mCslChannel` would keep synchronized to the value of `mRadioChannel` in `Mac` and `Csl Channel TLV` wouldn't be appended in Child Update Request to its parent. On CSL transmitter, `CSL channel` info in `child` is allowed to be `0`. If it's `0`, the channel info in frame would be set using `Mac::GetPanChannel`.

### DIFF
--- a/include/openthread/link.h
+++ b/include/openthread/link.h
@@ -1019,7 +1019,8 @@ uint8_t otLinkCslGetChannel(otInstance *aInstance);
  * This function sets the CSL channel.
  *
  * @param[in]  aInstance      A pointer to an OpenThread instance.
- * @param[in]  aChannel       The CSL sample channel.
+ * @param[in]  aChannel       The CSL sample channel. Channel value should be `0` (Set CSL Channel unspecified) or
+ *                            within the valid range [11, 26];
  *
  * @retval OT_ERROR_NONE           Successfully set the CSL parameters.
  * @retval OT_ERROR_INVALID_ARGS   Invalid @p aChannel.
--- a/include/openthread/link.h
+++ b/include/openthread/link.h
@@ -1020,7 +1020,7 @@ uint8_t otLinkCslGetChannel(otInstance *aInstance);
  *
  * @param[in]  aInstance      A pointer to an OpenThread instance.
  * @param[in]  aChannel       The CSL sample channel. Channel value should be `0` (Set CSL Channel unspecified) or
- *                            within the valid range [11, 26];
+ *                            within the range [1, 10] (if 915-MHz supported) and [11, 26] (if 2.4 GHz supported).
  *
  * @retval OT_ERROR_NONE           Successfully set the CSL parameters.
  * @retval OT_ERROR_INVALID_ARGS   Invalid @p aChannel.
--- a/src/core/api/link_api.cpp
+++ b/src/core/api/link_api.cpp
@@ -493,8 +493,7 @@ otError otLinkCslSetChannel(otInstance *aInstance, uint8_t aChannel)
     otError   error    = OT_ERROR_NONE;
     Instance &instance = *static_cast<Instance *>(aInstance);
 
-    VerifyOrExit(aChannel == 0 || ((Radio::kChannelMin <= aChannel) && (aChannel <= Radio::kChannelMax)),
-                 error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(Radio::IsCslChannelValid(aChannel), error = OT_ERROR_INVALID_ARGS);
 
     instance.Get<Mac::Mac>().SetCslChannel(aChannel);
 
--- a/src/core/api/link_api.cpp
+++ b/src/core/api/link_api.cpp
@@ -493,7 +493,8 @@ otError otLinkCslSetChannel(otInstance *aInstance, uint8_t aChannel)
     otError   error    = OT_ERROR_NONE;
     Instance &instance = *static_cast<Instance *>(aInstance);
 
-    VerifyOrExit((Radio::kChannelMin <= aChannel) && (aChannel <= Radio::kChannelMax), error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(aChannel == 0 || ((Radio::kChannelMin <= aChannel) && (aChannel <= Radio::kChannelMax)),
+                 error = OT_ERROR_INVALID_ARGS);
 
     instance.Get<Mac::Mac>().SetCslChannel(aChannel);
 
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -434,6 +434,13 @@ otError Mac::SetPanChannel(uint8_t aChannel)
 
     mRadioChannel = mPanChannel;
 
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+    if (!mSubMac.IsCslChannelSpecified())
+    {
+        mSubMac.SetCslChannel(mRadioChannel);
+    }
+#endif
+
     UpdateIdleMode();
 
 exit:
@@ -658,7 +665,7 @@ void Mac::UpdateIdleMode(void)
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
         if (IsCslEnabled())
         {
-            IgnoreError(mSubMac.CslSample());
+            IgnoreError(mSubMac.CslSample(mRadioChannel));
             ExitNow();
         }
 #endif
@@ -2269,6 +2276,7 @@ void Mac::SetCslChannel(uint8_t aChannel)
     VerifyOrExit(GetCslChannel() != aChannel, OT_NOOP);
 
     mSubMac.SetCslChannel(aChannel);
+    mSubMac.SetIsCslChannelSpecified(true);
 
     if (IsCslEnabled())
     {
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -434,13 +434,6 @@ otError Mac::SetPanChannel(uint8_t aChannel)
 
     mRadioChannel = mPanChannel;
 
-#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
-    if (!mSubMac.IsCslChannelSpecified())
-    {
-        mSubMac.SetCslChannel(mRadioChannel);
-    }
-#endif
-
     UpdateIdleMode();
 
 exit:
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -2276,7 +2276,7 @@ void Mac::SetCslChannel(uint8_t aChannel)
     VerifyOrExit(GetCslChannel() != aChannel, OT_NOOP);
 
     mSubMac.SetCslChannel(aChannel);
-    mSubMac.SetIsCslChannelSpecified(true);
+    mSubMac.SetCslChannelSpecified(aChannel != 0 ? true : false);
 
     if (IsCslEnabled())
     {
--- a/src/core/mac/sub_mac.hpp
+++ b/src/core/mac/sub_mac.hpp
@@ -305,7 +305,7 @@ public:
      * started, `mState` will become `kStateCslSample`. But it could be doing `Sleep` or `Receive` at this moment
      * (depending on `mCslState`).
      *
-     * @param[in]  aPanChannel  The current phy channel used by the device. This param is passed for the case when CSL
+     * @param[in]  aPanChannel  The current phy channel used by the device. This param will only take effect when CSL
      *                          channel hasn't been explicitly specified.
      *
      * @retval OT_ERROR_NONE          Successfully entered CSL operation (sleep or receive according to CSL timer).
@@ -387,7 +387,7 @@ public:
     /**
      * This method sets the CSL channel.
      *
-     * @param[in]  aChannel  The CSL channel.
+     * @param[in]  aChannel  The CSL channel. `0` to set CSL Channel unspecified.
      *
      */
     void SetCslChannel(uint8_t aChannel);
--- a/src/core/mac/sub_mac.hpp
+++ b/src/core/mac/sub_mac.hpp
@@ -404,7 +404,7 @@ public:
      * This method sets the flag representing if CSL channel has been specified.
      *
      */
-    void SetIsCslChannelSpecified(bool aIsSpecified) { mIsCslChannelSpecified = aIsSpecified; }
+    void SetCslChannelSpecified(bool aIsSpecified) { mIsCslChannelSpecified = aIsSpecified; }
 
     /**
      * This method gets the CSL period.
@@ -624,10 +624,10 @@ private:
     uint32_t  mCslTimeout;     ///< The CSL synchronized timeout in seconds.
     TimeMicro mCslSampleTime;  ///< The CSL sample time of the current period.
     uint16_t  mCslPeriod;      ///< The CSL sample period, in units of 10 symbols (160 microseconds).
-    uint8_t   mCslChannel : 7; ///< The actually CSL sample channel. If `mCslChannelIsSet` is 0, this should be equal to
-                               ///< the Pan channel of `Mac`.
+    uint8_t   mCslChannel : 7; ///< The actually CSL sample channel. If `mIsCslChannelSpecified` is 0, this should be
+                               ///< equal to the Pan channel of `Mac`.
     uint8_t
-        mIsCslChannelSpecified : 1; ///< Indicates that if CSL channel has been explicitly specified by the upper layer.
+        mIsCslChannelSpecified : 1; ///< Indicates whether or not the CSL channel was explicitly specified by the user.
 
     CslState mCslState;
 
--- a/src/core/mac/sub_mac.hpp
+++ b/src/core/mac/sub_mac.hpp
@@ -398,7 +398,7 @@ public:
      * @returns If CSL channel has been specified.
      *
      */
-    bool IsCslChannelSpecified() const { return mIsCslChannelSpecified; }
+    bool IsCslChannelSpecified(void) const { return mIsCslChannelSpecified; }
 
     /**
      * This method sets the flag representing if CSL channel has been specified.
@@ -626,8 +626,8 @@ private:
     uint16_t  mCslPeriod;      ///< The CSL sample period, in units of 10 symbols (160 microseconds).
     uint8_t   mCslChannel : 7; ///< The actually CSL sample channel. If `mIsCslChannelSpecified` is 0, this should be
                                ///< equal to the Pan channel of `Mac`.
-    uint8_t
-        mIsCslChannelSpecified : 1; ///< Indicates whether or not the CSL channel was explicitly specified by the user.
+    uint8_t mIsCslChannelSpecified : 1; ///< Indicates whether or not the CSL channel was explicitly specified by
+                                        ///< the user.
 
     CslState mCslState;
 
--- a/src/core/mac/sub_mac.hpp
+++ b/src/core/mac/sub_mac.hpp
@@ -305,12 +305,15 @@ public:
      * started, `mState` will become `kStateCslSample`. But it could be doing `Sleep` or `Receive` at this moment
      * (depending on `mCslState`).
      *
+     * @param[in]  aPanChannel  The current phy channel used by the device. This param is passed for the case when CSL
+     *                          channel hasn't been explicitly specified.
+     *
      * @retval OT_ERROR_NONE          Successfully entered CSL operation (sleep or receive according to CSL timer).
      * @retval OT_ERROR_BUSY          The radio was transmitting.
      * @retval OT_ERROR_INVALID_STATE The radio was disabled.
      *
      */
-    otError CslSample(void);
+    otError CslSample(uint8_t aPanChannel);
 #endif
 
     /**
@@ -388,6 +391,20 @@ public:
      *
      */
     void SetCslChannel(uint8_t aChannel);
+
+    /**
+     * This method indicates if CSL channel has been explicitly specified by the upper layer.
+     *
+     * @returns If CSL channel has been specified.
+     *
+     */
+    bool IsCslChannelSpecified() const { return mIsCslChannelSpecified; }
+
+    /**
+     * This method sets the flag representing if CSL channel has been specified.
+     *
+     */
+    void SetIsCslChannelSpecified(bool aIsSpecified) { mIsCslChannelSpecified = aIsSpecified; }
 
     /**
      * This method gets the CSL period.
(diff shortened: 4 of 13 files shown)


src/core/mac/sub_mac.hpp
```diff
@@ -389,6 +392,20 @@ class SubMac : public InstanceLocator
...
+     * @returns If CSL channel has been specified.
+     *
+     */
+    bool IsCslChannelSpecified() const { return mIsCslChannelSpecified; }
+
+    /**
+     * This method sets the flag representing if CSL channel has been specified.
+     *
+     */
+    void SetIsCslChannelSpecified(bool aIsSpecified) { mIsCslChannelSpecified = aIsSpecified; }
```
librasungirl, line 407:
```suggestion
    void SetCslChannelSpecified(bool aIsSpecified) { mIsCslChannelSpecified = aIsSpecified; }
```

src/core/mac/sub_mac.hpp
```diff
@@ -602,10 +619,13 @@ class SubMac : public InstanceLocator
...
 
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
-    uint32_t  mCslTimeout;    ///< The CSL synchronized timeout in seconds.
-    TimeMicro mCslSampleTime; ///< The CSL sample time of the current period.
-    uint16_t  mCslPeriod;     ///< The CSL sample period, in units of 10 symbols (160 microseconds).
-    uint8_t   mCslChannel;    ///< The CSL sample channel.
+    uint32_t  mCslTimeout;     ///< The CSL synchronized timeout in seconds.
+    TimeMicro mCslSampleTime;  ///< The CSL sample time of the current period.
+    uint16_t  mCslPeriod;      ///< The CSL sample period, in units of 10 symbols (160 microseconds).
+    uint8_t   mCslChannel : 7; ///< The actually CSL sample channel. If `mCslChannelIsSet` is 0, this should be equal to
```
librasungirl, line 625:
```suggestion
    uint8_t   mCslChannel : 7; ///< The actually CSL sample channel. If `mIsCslChannelSpecified ` is 0, this should be equal to
```

src/core/mac/sub_mac.hpp
```diff
@@ -602,10 +619,13 @@ class SubMac : public InstanceLocator
...
+                               ///< the Pan channel of `Mac`.
+    uint8_t
+        mIsCslChannelSpecified : 1; ///< Indicates that if CSL channel has been explicitly specified by the upper layer.
```
librasungirl, line 628: Mac layer is also upper layer,  update to be more accurate?

```suggestion
        mIsCslChannelSpecified : 1; ///< Indicates whether or not the CSL channel was explicitly specified by the user.
```

src/core/thread/csl_tx_scheduler.cpp
```diff
@@ -188,7 +188,8 @@ otError CslTxScheduler::HandleFrameRequest(Mac::TxFrame &aFrame)
         aFrame.SetIsARetransmission(false);
     }
 
-    aFrame.SetChannel(mCslTxChild->GetCslChannel());
+    aFrame.SetChannel(mCslTxChild->GetCslChannel() == 0 ? Get<Mac::Mac>().GetPanChannel()
```
librasungirl, line 191: Instead of executing this condition statement per frame to send, for efficiency, how about setting the child's CSL channel to be pan channel directly?

tests/scripts/thread-cert/v1_2_test_csl_transmission.py
```diff
@@ -71,6 +72,16 @@ def test(self):
         self.assertTrue(self.nodes[LEADER].ping(self.nodes[SSED_1].get_rloc()))
         self.simulator.go(5)
 
+        self.nodes[SSED_1].set_csl_channel(SECOND_CHANNEL)
+        self.nodes[LEADER].set_csl_channel(SECOND_CHANNEL)
```
librasungirl, line 76: is it necessary to configure the CSL channel for leader?

tests/scripts/thread-cert/v1_2_test_csl_transmission.py
```diff
@@ -36,7 +36,9 @@
 
 CSL_PERIOD = 500 * 6.25  # 500ms
 CSL_TIMEOUT = 30  # 30s
-CSL_CHANNEL = 12
+CSL_CHANNEL = 13
+
+SECOND_CHANNEL = 12
```
librasungirl, line 41:
```suggestion
CSL_CHANNEL_1 = 13  # different from the default pan channel 11
CSL_CHANNEL_2 = 12
```

librasungirl: is it necessary to test two different CSL channels?

src/core/thread/csl_tx_scheduler.cpp
Irving-cl: I think this would make the logic more complex: When the parent changes channel, it might need to change `csl channel` for each of its child. And it doesn't know if the `csl channel` has been specified by the user of CSL receiver.(since we have set child's CSL channel to be a non-zero value).

tests/scripts/thread-cert/v1_2_test_csl_transmission.py
Irving-cl: A mistake here. It should be `set_channel`. The purpose is to test, without specifying csl channel, csl tx works properly when both devices change their channels. Changing channels can only be done when the devices are stopped. I added a logic to stop, set channel and restart.

tests/scripts/thread-cert/v1_2_test_csl_transmission.py
Irving-cl: As above mentioned.

src/core/thread/csl_tx_scheduler.cpp
librasungirl: make sense.

src/core/thread/mle.cpp
```diff
@@ -1396,7 +1396,7 @@ otError Mle::AppendCslChannel(Message &aMessage)
     otError       error = OT_ERROR_NONE;
     CslChannelTlv cslChannel;
 
-    VerifyOrExit(Get<Mac::Mac>().GetPanChannel() != Get<Mac::Mac>().GetCslChannel(), OT_NOOP);
+    VerifyOrExit(Get<Mac::Mac>().IsCslChannelSpecified(), OT_NOOP);
```
librasungirl, line 1399: What if the specified csl channel is the same as the pan channel?

src/core/thread/mle_router.cpp
```diff
@@ -2515,10 +2515,6 @@ void MleRouter::HandleChildUpdateRequest(const Message &         aMessage,
         {
             child->SetCslChannel(static_cast<uint8_t>(cslChannel.GetChannel()));
         }
```
librasungirl, line 2517: If `kCslChannel` Tlv. could not be found, should the parent clears the child's csl channel?

tests/scripts/thread-cert/v1_2_test_csl_transmission.py
```diff
@@ -71,6 +72,28 @@ def test(self):
         self.assertTrue(self.nodes[LEADER].ping(self.nodes[SSED_1].get_rloc()))
         self.simulator.go(5)
```
librasungirl, line 73: should we check that there is no CslChannel Tlv in Child Update Request?

tests/scripts/thread-cert/v1_2_test_csl_transmission.py
```diff
@@ -71,6 +72,28 @@ def test(self):
...
+
+        self.nodes[SSED_1].start()
+        self.simulator.go(7)
+        self.assertEqual(self.nodes[SSED_1].get_state(), 'child')
+
+        self.assertTrue(self.nodes[LEADER].ping(self.nodes[SSED_1].get_rloc()))
+        self.simulator.go(5)
+
+        self.nodes[SSED_1].set_csl_channel(CSL_CHANNEL)
+        self.simulator.go(1)
```
librasungirl, line 93: Similarly, check there is csl channel tlv in CUR with value `CSL_CHANNEL`?

src/core/mac/mac.cpp
```diff
@@ -2269,6 +2276,7 @@ void Mac::SetCslChannel(uint8_t aChannel)
     VerifyOrExit(GetCslChannel() != aChannel, OT_NOOP);
 
     mSubMac.SetCslChannel(aChannel);
+    mSubMac.SetCslChannelSpecified(true);
```
librasungirl, line 2279: is there chance that the user may want to clear the csl channel he/she specified and let csl work on the same pan channel?

tests/scripts/thread-cert/v1_2_test_csl_transmission.py
librasungirl: `stop, set_channel and then start` doesn't make any difference with the procedure before stop as by default both devices work on the same default channel 11.

To test the csl channel (when unspecified) changes along with pan channel, the proper way is to change the channel of running stack via pending dataset.

src/core/thread/mle.cpp
Irving-cl: I think it doesn't matter. It's not forbidden in spec to specify the CSL channel to current pan channel. Though there's no effect in terms of CSL transmission, but the internal state(if CSL channel has been specified) should be changed in this scenario. Thoughts?

src/core/thread/mle_router.cpp
Irving-cl: I think it shouldn't.

After reading this again:
> If configured, CSL data transmission and reception using the additional channel operates as described in Section 6.12.2.6 in [IEEE802154-2015]. The CSL Channel for a Thread Interface operating as a CSL Receiver or as a CSL Transmitter is either:
> • The value of phyCurrentChannel when macCSLChannelMask has value 0.
> • The channel value selected by macCSLChannelMask when macCSLChannelMask is different from 0.

I think it's ok to set for SSED to set CSL Channel to 0 again if it wants so. So when there's no CSL Channel TLV, we don't need to clear child's CSL channel.

src/core/mac/mac.cpp
Irving-cl: Make sense. As I mentioned above, I think users can do this by setting CSL channel to 0 again.

Irving-cl:
@librasungirl 

Made some changes to allow users to set CSL channel to be 0(which means unspecified):
1. When CSL Receiver changes its CSL channel, it would always send CSL Channel TLV. Even if it's 0.
2. On Transmitter side, no changes is required. It would always set child's CSL channel to the new value, including 0.
3. On Receiver side, if CSL Channel is set to 0, `IsSpecified` would be set to `false` again.

src/core/thread/mle.cpp
librasungirl: > The Child Update Request SHALL include the CSL Channel TLV if the SSED selected a separate channel for CSL operation. Otherwise, the Child Update Request MAY include the CSL Channel TLV

agree that thread spec does not forbid the csl channel same as pan channel.

src/core/thread/mle_router.cpp
librasungirl: See CSL Channel TLV definition in 4.4.21
> IEEE 802.15.4 CslChannelPage: An unsigned 8-bit value that specifies the channel page.
> IEEE 802.15.4 CslChannel: An unsigned 16-bit value that identifies the channel within the channel page.

I am afraid channel 0 is is not a valid channel within the channel page 0, which have 16 channels [11, 26] in the 2450 MHz band

src/core/mac/mac.cpp
```diff
@@ -434,6 +434,13 @@ otError Mac::SetPanChannel(uint8_t aChannel)
 
     mRadioChannel = mPanChannel;
 
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+    if (!mSubMac.IsCslChannelSpecified())
+    {
+        mSubMac.SetCslChannel(mRadioChannel);
+    }
+#endif
```
librasungirl, line 442: is it necessary given `mSubMac.CslSample(mRadioChannel)` below?

src/core/thread/mle.cpp
```diff
@@ -1396,8 +1396,6 @@ otError Mle::AppendCslChannel(Message &aMessage)
     otError       error = OT_ERROR_NONE;
     CslChannelTlv cslChannel;
 
-    VerifyOrExit(Get<Mac::Mac>().GetPanChannel() != Get<Mac::Mac>().GetCslChannel(), OT_NOOP);
```
librasungirl, line 1399: change to only append when csl channel is specified, other than remove the condition here?

tests/scripts/thread-cert/v1_2_test_csl_transmission.py
```diff
@@ -71,6 +73,26 @@ def test(self):
         self.assertTrue(self.nodes[LEADER].ping(self.nodes[SSED_1].get_rloc()))
         self.simulator.go(5)
 
+        ssed_messages = self.simulator.get_messages_sent_by(SSED_1)
+        msg = ssed_messages.next_mle_message(mle.CommandType.CHILD_UPDATE_REQUEST)
+        msg.assertMleMessageContainsTlv(mle.CslChannel)
```
librasungirl, line 78: should assert there is no csl channel tlv  as no csl channel is specified and it is the same as pan channel

tests/scripts/thread-cert/v1_2_test_csl_transmission.py
```diff
@@ -71,6 +73,26 @@ def test(self):
...
+
+        # Change channel
+        self.nodes[LEADER].send_mgmt_pending_set(channel=SECOND_CHANNEL)
```
librasungirl, line 81: For non-commissioner thread device, it should include full pending dataset in MGMT_PENDING_SET, otherwise, the pending dataset in MLE Data Response would be incomplete.
To allow partial dataset from simplification's perspective, I would suggest you to make Leader become active commissioner.

src/core/thread/mle_router.cpp
Irving-cl: @librasungirl 

After reading the spec again, I have some interesting findings and questions now. Correct me if I'm wrong.

First I must say, currently my description of **'CSL channel has been specified or not'** is inaccurate. The spec actually means:
1. When *macCSLChannelMask* has value `0`, CSL tx uses `phyCurrentChannel`.
2. When *macCSLChannelMask* isn't `0`, CSL tx uses `phyCurrentChannel`.

**I would call these two conditions case 1 and case 2 below.**

*macCSLChannelMask* is a PIB attribute. So if we set CSL Channel first(move to case 2) and then want to get back to case 1, we just need to change *macCSLChannelMask* back to 0. I believe that `The value of phyCurrentChannel when macCSLChannelMask has value 0.` indicates *macCSLChannelMask* could equal to 0.

The spec also mentions that(**3.2.6.3.2**):

> The macCslChannelMask is relayed by a CSL Receiver SSED Child to a CSL Transmitter Parent Router via the CSL Channel TLV sent in the Child Update Request as described in Section 4.6

However, what you mentioned(CSL Channel TLV in 4.4.31) is right. If we set `CslChannelPage` to `0`, then we cannot set `CslChannel` to `0` in the TLV.

So here is my question:
1. I found it's mentioned in **8.10.1.1.1 Supported Channel Pages**, currently only `0` is a valid value, `1-255` are reserved. Can we use `1-255` for our implementation usage?
2. If by no means we can relay `0` in CSL Channel TLV, then I think we cannot use a 'Thread' way to move from case 2 to case 1. We have to set *macCSLChannelMask* on both sides if we want to do so. Actually I think it's fine even if we don't support to move from case 2 to case 1. Just ask for your opinion.

For now, I would prefer implement it all conforming to the spec. That is, we cannot move from case 2 to case 1. Thoughts?

librasungirl: channel page 1-255 for 2.4Ghz conflicts IEEE802.15.4 spec.

Not sure how to exactly interprete the `relay` here,  macCSLChannelMask is not supported in your CSL implementation, in fact, even supported, the mask should also only represent channels [11, 26], and choose one within these channels. I think so far we may simply focus on the CSL Channel TLV and may assume user can specify any channel valid for channel page 0?

I think it should be fine to just not append CSL channel Tlv when unspecified, and clear the SSED's csl channel when the parent receives CUR without CSL channel TLV.

Thoughts?

Irving-cl: > I think it should be fine to just not append CSL channel Tlv when unspecified, and clear the SSED's csl channel when the parent receives CUR without CSL channel TLV.

SGTM.

src/core/mac/sub_mac.hpp
```diff
@@ -305,12 +305,15 @@ class SubMac : public InstanceLocator
      * started, `mState` will become `kStateCslSample`. But it could be doing `Sleep` or `Receive` at this moment
      * (depending on `mCslState`).
      *
+     * @param[in]  aPanChannel  The current phy channel used by the device. This param is passed for the case when CSL
```
librasungirl, line 308: the comment seems to me not accurate - @p aPanChannel is always passed, but only take effect when CslChannel is not specified.

src/core/mac/sub_mac.hpp
```diff
@@ -389,6 +392,26 @@ class SubMac : public InstanceLocator
      */
     void SetCslChannel(uint8_t aChannel);
```
librasungirl, line 393: add comment that value `0` is to clear the specified Csl Channel?

src/core/mac/sub_mac.hpp
```diff
@@ -389,6 +392,26 @@ class SubMac : public InstanceLocator
...
 
+    /**
+     * This method indicates if CSL channel has been explicitly specified by the upper layer.
+     *
+     * In other words, this indicates that if the PIB attribute `macCSLChannelMask` is 0. In current implementation,
+     * it's allowed to set this to 0 again.
```
librasungirl, line 399: `mIsCslChannelSpecified` is only related to `mCslChannel`, and it is `mCslChannel` that is required to be a valid channel value within `macCSLChannelMask`. 
Why brings the new `macCSLChannelMask` concept, which is not implemented or well documented, when introducing `mIsCslChannelSpecified` related APIs?

For simplicity, or at least before `macCSLChannelMask` is introduced, I think it should be enough to only specifically document that `0` is to clear the Csl Channel specified.

Thoughts?

tests/scripts/thread-cert/v1_2_test_csl_transmission.py
```diff
@@ -29,13 +29,14 @@
...
 
+import mle
 import thread_cert
 
 LEADER = 1
 SSED_1 = 2
 
 CSL_PERIOD = 500 * 6.25  # 500ms
-CSL_TIMEOUT = 30  # 30s
+CSL_TIMEOUT = 305  # 305s
```
librasungirl, line 39: restore the smaller value as you are not using pending dataset for test?

src/core/mac/sub_mac.hpp
Irving-cl: SGTM.

tests/scripts/thread-cert/v1_2_test_csl_transmission.py
```diff
@@ -75,16 +73,14 @@ def test(self):
...
-        msg.assertMleMessageContainsTlv(mle.CslChannel)
-
-        # Change channel
-        self.nodes[LEADER].send_mgmt_pending_set(channel=SECOND_CHANNEL)
-        self.simulator.go(301)
+        msg.assertMleMessageDoesNotContainTlv(mle.CslChannel)
 
+        self.nodes[SSED_1].set_csl_channel(CSL_CHANNEL)
+        self.simulator.go(1)
         self.assertTrue(self.nodes[LEADER].ping(self.nodes[SSED_1].get_rloc()))
```
librasungirl, line 80: assert there is CSL channel TLV when CSL channel is different from pan channel?

librasungirl (approved):

src/core/api/link_api.cpp
```diff
@@ -493,7 +493,8 @@ otError otLinkCslSetChannel(otInstance *aInstance, uint8_t aChannel)
     otError   error    = OT_ERROR_NONE;
     Instance &instance = *static_cast<Instance *>(aInstance);
 
-    VerifyOrExit((Radio::kChannelMin <= aChannel) && (aChannel <= Radio::kChannelMax), error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(aChannel == 0 || ((Radio::kChannelMin <= aChannel) && (aChannel <= Radio::kChannelMax)),
```
librasungirl, line 496: update the doxygen for `otLinkCslSetChannel()` to allow value 0 to clear the csl channel?

include/openthread/link.h
```diff
@@ -1019,7 +1019,8 @@ uint8_t otLinkCslGetChannel(otInstance *aInstance);
  * This function sets the CSL channel.
  *
  * @param[in]  aInstance      A pointer to an OpenThread instance.
- * @param[in]  aChannel       The CSL sample channel.
+ * @param[in]  aChannel       The CSL sample channel. Channel value should be `0` (Set CSL Channel unspecified) or
+ *                            within the valid range [11, 26];
```
jwhui, line 1023: With the experimental sub-GHz support, we allow channels 1 through 10.

src/core/mac/sub_mac.hpp
```diff
@@ -384,11 +387,25 @@ class SubMac : public InstanceLocator
...
      */
     void SetCslChannel(uint8_t aChannel);
 
+    /**
+     * This method indicates if CSL channel has been explicitly specified by the upper layer.
+     *
+     * @returns If CSL channel has been specified.
+     *
+     */
+    bool IsCslChannelSpecified() const { return mIsCslChannelSpecified; }
```
jwhui, line 401:
```suggestion
    bool IsCslChannelSpecified(void) const { return mIsCslChannelSpecified; }
```

src/core/mac/sub_mac.hpp
```diff
@@ -604,10 +621,13 @@ class SubMac : public InstanceLocator
...
-    TimeMicro mCslSampleTime; ///< The CSL sample time of the current period.
-    uint16_t  mCslPeriod;     ///< The CSL sample period, in units of 10 symbols (160 microseconds).
-    uint8_t   mCslChannel;    ///< The CSL sample channel.
+    uint32_t  mCslTimeout;     ///< The CSL synchronized timeout in seconds.
+    TimeMicro mCslSampleTime;  ///< The CSL sample time of the current period.
+    uint16_t  mCslPeriod;      ///< The CSL sample period, in units of 10 symbols (160 microseconds).
+    uint8_t   mCslChannel : 7; ///< The actually CSL sample channel. If `mIsCslChannelSpecified` is 0, this should be
+                               ///< equal to the Pan channel of `Mac`.
+    uint8_t
+        mIsCslChannelSpecified : 1; ///< Indicates whether or not the CSL channel was explicitly specified by the user.
```
jwhui, line 630: nit: can we avoid the newline after `uint8_t` by adjusting the end-of-line comment?

src/core/api/link_api.cpp
edmont, line 496: What about an `IsCslChannelValid()` function to take into account sub-GHz support as @jwhui mentions?

src/core/thread/mle.cpp
```diff
@@ -1396,7 +1396,11 @@ otError Mle::AppendCslChannel(Message &aMessage)
...
+    // In current implementation, it's allowed to set CSL Channel unspecified again. As `0` is not valid for
```
edmont, line 1399:
```suggestion
    // In current implementation, it's allowed to set CSL Channel unspecified. As `0` is not valid for
```

edmont (approved):
LGTM.

jwhui (approved):

PR #5503, opened 2020-09-08
[coap] update/simplify CoAP Options generation/processing

abtink:
This commit simplifies the generation/processing of CoAP Options in a
CoAP message. It adds helper methods `ReadExtendedOptionField()`,
`WriteExtendedOptionField` to decode/encode extended Option Header
fields (Option Number Delta or Option Length). It simplifies
`Option::Iterator` (using `Init()`/`Advance()` to parse and iterate
through all CoAP Options in a message). Note that the public OT
`otCoap` APIs and their behavior remain as before and unchanged.

### DIFF
--- a/src/core/coap/coap.cpp
+++ b/src/core/coap/coap.cpp
@@ -171,18 +171,18 @@ otError CoapBase::SendMessage(Message &               aMessage,
 
 #if OPENTHREAD_CONFIG_COAP_OBSERVE_API_ENABLE
         // Whether or not to turn on special "Observe" handling.
-        OptionIterator iterator;
-        bool           observe;
+        Option::Iterator iterator;
+        bool             observe;
 
-        SuccessOrExit(error = iterator.Init(&aMessage));
-        observe = (iterator.GetFirstOptionMatching(kOptionObserve) != nullptr);
+        SuccessOrExit(error = iterator.Init(aMessage, kOptionObserve));
+        observe = !iterator.IsDone();
 
         // Special case, if we're sending a GET with Observe=1, that is a cancellation.
         if (observe && aMessage.IsGetRequest())
         {
             uint64_t observeVal = 0;
 
-            SuccessOrExit(error = iterator.GetOptionValue(observeVal));
+            SuccessOrExit(error = iterator.ReadOptionValue(observeVal));
 
             if (observeVal == 1)
             {
@@ -570,10 +570,10 @@ void CoapBase::ProcessReceivedResponse(Message &aMessage, const Ip6::MessageInfo
     if (metadata.mObserve && request->IsRequest())
     {
         // We sent Observe in our request, see if we received Observe in the response too.
-        OptionIterator iterator;
+        Option::Iterator iterator;
 
-        SuccessOrExit(error = iterator.Init(&aMessage));
-        responseObserve = (iterator.GetFirstOptionMatching(kOptionObserve) != nullptr);
+        SuccessOrExit(error = iterator.Init(aMessage, kOptionObserve));
+        responseObserve = !iterator.IsDone();
     }
 #endif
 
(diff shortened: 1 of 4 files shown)


src/core/coap/coap.cpp
```diff
@@ -714,29 +714,26 @@ void CoapBase::ProcessReceivedRequest(Message &aMessage, const Ip6::MessageInfo
...
-        case kOptionUriPath:
-            if (curUriPath != uriPath)
-            {
-                *curUriPath++ = '/';
-            }
+            *curUriPath++ = '/';
+        }
 
-            VerifyOrExit(option->mLength < sizeof(uriPath) - static_cast<size_t>(curUriPath + 1 - uriPath), OT_NOOP);
+        VerifyOrExit(curUriPath + optionLength < OT_ARRAY_END(uriPath), OT_NOOP);
```
abtink, line 728: Adding  a note here about the change so that it can also be checked/verified by others/reviewers.

I think the condition in the original code was too restrictive (potentially leaving one byte/char in the array unused). In the new code, we just check if after adding `optionLength` to `cur` we are still before the array end (which should still leave space for the null `\0` char to be added).

src/core/coap/coap_message.hpp
```diff
@@ -911,88 +917,178 @@ class MessageQueue : public ot::MessageQueue
...
+        otError ReadOptionValue(void *aValue) const;
+
+        /**
+         * This method read the current Option Value which is assumed to be an unsigned integer.
+         *
+         * @param[out]  aValue          A reference to `uint64_t` to output the read Option Value.
+         *
+         * @retval  OT_ERROR_NONE       Successfully read the Option value.
+         * @retval  OT_ERROR_NO_BUFS    Value is too long to fit in an `uint64_t`.
+         * @retval OT_ERROR_NOT_FOUND   Iterator is done (not pointing to any option).
```
jwhui, line 1047: nit: alignment

src/core/coap/coap_message.hpp
```diff
@@ -911,88 +917,178 @@ class MessageQueue : public ot::MessageQueue
...
-     * @returns A pointer to the next option. If no more options are present nullptr pointer is returned.
-     */
-    const otCoapOption *GetNextOption(void);
+        /**
+         * This method advances the iterator to the next CoAP Option in the header.
+         *
+         * The iterator is updated to point to the next option or marked as done when there are no more options.
+         *
+         * @retval  OT_ERROR_NONE   Successfully advanced the iterator.
+         * @retval  OT_ERROR_PARSE  CoAP Option header in @p aMessage is not well-formed.
```
jwhui, line 994: `aMessage` is not a parameter.

src/core/coap/coap_message.hpp
```diff
@@ -911,88 +917,178 @@ class MessageQueue : public ot::MessageQueue
...
+        otError Advance(void);
+
+        /**
+         * This method advances the iterator to the next CoAP Option in the header matching a given Option Number value.
+         *
+         * The iterator is updated to point to the next matching option or marked as done when there are no more
+         * matching options.
+         *
+         * @retval  OT_ERROR_NONE   Successfully advanced the iterator.
+         * @retval  OT_ERROR_PARSE  CoAP Option header in @p aMessage is not well-formed.
```
jwhui, line 1006: `aMessage` is not a parameter.

src/core/coap/coap_message.hpp
```diff
@@ -911,88 +917,178 @@ class MessageQueue : public ot::MessageQueue
...
+         *
+         */
+        otError Advance(uint16_t aNumber) { return InitOrAdvance(nullptr, aNumber); }
```
jwhui, line 1009: Document `aNumber`.

jwhui (approved):

PR #5283, opened 2020-07-23
[continuous-integration] add code coverage threshold

Moandor-y:
This adds 2 coverage checks to pull requests:
1. The overall coverage must not drop more than 1% against the parent commit.
1. ~~The diff coverage must be at least 85%.~~

PR #5509, opened 2020-09-10
[meshcop] add helper function 'LogError()'

abtink:


src/core/meshcop/commissioner.cpp
```diff
@@ -968,16 +958,12 @@ void Commissioner::SendKeepAlive(uint16_t aSessionId)
...
-        otLogWarnMeshCoP("Failed to send keep alive: %s", otThreadErrorToString(error));
-
-        if (message != nullptr)
-        {
-            message->Free();
-        }
+        message->Free();
     }
+
+    LogError("send keep alive", error);
```
simonlingoogle, line 966:
```suggestion
    LogError("send", "keep alive", error);
```
Should we split `action` and `message type` ? The actions can be `send`, `start`, `process`, etc.

abtink: Having more function parameters would increase code size on caller side (now requiring two arguments to be prepared (loaded into registers, etc.) before method/function call). The code-size reduction from potential sharing of strings "send", "start", etc would need to be balanced by this. 

Also the sharing of common string definitions is not always done by all toolchains. For example,  some compilers (particularly arm architecture) can prefer (under certain optimization flags) to include all constant (arrays/strings/pointers) in the region directly at the end of each method so that instructions that load relative to `PC` (program counter) register can be used to access/load them quickly.

If the goal is to reduce code size by sharing string, it would help if we can do this explicitly (e.g. , define an `Action` enum to covert to string within `LogError` itself) and/or share the message type (`"keep alive"`) strings as well.

 I would propose to leave this for a separate PR (future change)

simonlingoogle:
Note that the diff coverage of this PR is only 52%. How should we cover this PR if #5283 were merged?

@abtink @Moandor-y @jwhui

abtink:
> Note that the diff coverage of this PR is only 52%. How should we cover this PR if #5283 were merged?

Here are my thoughts:
- I am not in favor of adding strict thresholds for coverage until we know it's robust and correct (I don't monitor the code coverage reports in detail, but from observations/samples over time, I get the feeling that the coverage values are not always correct/accurate).
- It would be good to understand why for example, this PR is giving low diff coverage. Perhaps logging is not enabled/covered in the coverage builds, if so, we should see if we can add/enable it.
- I know we have used `// LCOV_EXCL_START` and `STOP` in  the code (which I think are excluding some potion of code from coverage). I guess it can be an option (but personally don't prefer excluding from coverage).

jwhui:
> * I am not in favor of adding strict thresholds for coverage until we know it's robust and correct (I don't monitor the code coverage reports in detail, but from observations/samples over time, I get the feeling that the coverage values are not always correct/accurate).

We have done quite a bit of work to make code coverage more stable. See #5283 for the discussion.

> * It would be good to understand why for example, this PR is giving low diff coverage. Perhaps logging is not enabled/covered in the coverage builds, if so, we should see if we can add/enable it.

About half of the diff is error conditions - see the [diff coverage report](https://codecov.io/gh/openthread/openthread/pull/5509/diff). We currently do not have a good mechanism to test all of these error conditions. 

> * I know we have used `// LCOV_EXCL_START` and `STOP` in  the code (which I think are excluding some potion of code from coverage). I guess it can be an option (but personally don't prefer excluding from coverage).

I generally prefer not to use this, but maybe it is OK for logging.

bukepo:
I don't think we can guarantee the code will work as expect if it's not covered, even for logging code. I remember there were multiple times that enabling `FULL_LOGS` exposed build errors.

So I think it meaningful to improve code coverage as much as possible.

simonlingoogle:
> > * It would be good to understand why for example, this PR is giving low diff coverage. Perhaps logging is not enabled/covered in the coverage builds, if so, we should see if we can add/enable it.
> 
> About half of the diff is error conditions - see the [diff coverage report](https://codecov.io/gh/openthread/openthread/pull/5509/diff). We currently do not have a good mechanism to test all of these error conditions.

I am worried about this too. I agree that it's meaningful to cover even error handling code. But it can be very difficult, especially for some small corner case error handling. For example, #5432 is a one line fix with diff coverage 0%.
Another example is that my Backbone TMF PR #5471 does not have enough code coverage either. It's more difficult to test this feature in this PR itself, but much easier to test it in #5388 and other Backbone related PRs. Sometime we split one feature in multiple PRs and might be preferable to test it when it's more complete. 

I think it's all about the compromise between productivity and stability. **Writing too much test code harms productivity without enough benefit.**
Furthermore, I don't think that we can guarantee that all covered code is bug free. It only gives us more confidence. 

> 
> > * I know we have used `// LCOV_EXCL_START` and `STOP` in  the code (which I think are excluding some potion of code from coverage). I guess it can be an option (but personally don't prefer excluding from coverage).
> 
> I generally prefer not to use this, but maybe it is OK for logging.

Using `LCOV_EXCL_START` and `STOP` can be a solution. But it will certainly lead to discussions in PRs about whether or not it's allowed to use for the enclosed code. 

> I remember there were multiple times that enabling FULL_LOGS exposed build errors.

I think the building tests are used to reveal build errors. It has nothing to do with code coverage.

src/core/meshcop/commissioner.cpp
simonlingoogle: > Having more function parameters would increase code size on caller side (now requiring two arguments to be prepared (loaded into registers, etc.) before method/function call). The code-size reduction from potential sharing of strings "send", "start", etc would need to be balanced by this.

We can easily compensate for the code size increase for one more parameter if the action strings are reused? Especially when the action strings are not only shared in meshcop, but also reused for other modules (e.g. MLE). 

> Also the sharing of common string definitions is not always done by all toolchains. For example, some compilers (particularly arm architecture) can prefer (under certain optimization flags) to include all constant (arrays/strings/pointers) in the region directly at the end of each method so that instructions that load relative to `PC` (program counter) register can be used to access/load them quickly.

But since `-Os` is used, compilers should prefer smaller code than than higher performance? I think we can have the benefit of reduced code size on majority of compilers (at least GCC) even without using string enum. The benefit should already overrun the potential risk. 

> If the goal is to reduce code size by sharing string, it would help if we can do this explicitly (e.g. , define an `Action` enum to covert to string within `LogError` itself) and/or share the message type (`"keep alive"`) strings as well.
> 
> I would propose to leave this for a separate PR (future change)

I think we should do it. Fine to me to leave this for a separate PR.

abtink: > We can easily compensate for the code size increase for one more parameter if the action strings are reused? Especially when the action strings are not only shared in meshcop, but also reused for other modules (e.g. MLE).

In general yes, but not necessarily in all cases. There are counter-intuitive cases.
It depends on different factors, e.g., size of strings being shared, how the new arg is added. 

Let me give an example:

If we share `"send"` string,  we would save 4 chars or 4 bytes (per its use). But adding a new parameter/argument in the `LogError()` call can add more than 4 bytes (6 bytes or more on arm-gcc) to the generated code (again per use).

Here is disassembly of code with two args in `LogError()` like `LogError("send", "scan results", error)`
(this is from arm-gcc-7 and same config as the one we use in our size-report)

```
# EnegyScanServer::SendReport()

00011868 <_ZN2ot16EnergyScanServer10SendReportEv>:
   11868:       b5f0            push    {r4, r5, r6, r7, lr}
   ....

   # Loading the two pointer values into r0 and r1 registers 
   # before branching to `MeshCoP::LogError`

   1194c:       490a            ldr     r1, [pc, #40]   ; (11978 <_ZN2ot16EnergyScanServer10SendReportEv+0x110>)
   1194e:       480b            ldr     r0, [pc, #44]   ; (1197c <_ZN2ot16EnergyScanServer10SendReportEv+0x114>)
   11950:       f7fd fa5a       bl      ee08 <_ZN2ot7MeshCoP8LogErrorEPKcS2_7otError>
      ...
   11966:       e7f0            b.n     1194a <_ZN2ot16EnergyScanServer10SendReportEv+0xe2>
   11968:       2000a344        .word   0x2000a344
   1196c:       00046403        .word   0x00046403
   11970:       2000ac2c        .word   0x2000ac2c
   11974:       00045694        .word   0x00045694
   11978:       000456aa        .word   0x000456aa
   1197c:       00045930        .word   0x00045930
```

Note that we see two pointers being loaded into `r0` and `r1` before calling `LogError()`. We need two bytes for `ldr  r1, [pc, #40]` instruction for the second argument, but in addition to this, the actual pointer value is stored right at the end of function at `[pc, #40]` or at address `11978` (which stores the string address `00456aa`). This adds another 4 bytes to the code. 

---

Another good (counter-intuitive) example is our logging:
- We have the feature to append log level and/or log region prefix string to the log.
- The prefix string are literally added to the `format` string, so the strings (in the image) are like:
  - `"[NOTE]-MLE-----: RLOC16 %04x -> %04x"`, or `"[NOTE]-MLE-----: Downgrade to REED"`
- At first it seems to be inefficient and makes sense to change it to reuse the `"[NOTE]"` and `"-MLE----"` strings.
- We tried this, which boils down to a call like this:
   - `otPlatLog("%s%s: " aFormat, sLevelPrefix[aLevel], sRegionPrefix[aRegion], ...)`
   - `sLevelPrefix[]` is const array of strings (for each log level), same way for `sRegionPrefix[]` string array.
- This change actually increased the code size largely (by ~2K if I recall).
- The reason is the overhead of adding two arguments to `otPlatLog()` call (which is variable arg function itself) was more than the bytes/chars saved by sharing the string literals.

-----

> But since -Os is used, compilers should prefer smaller code than than higher performance? I think we can have the benefit of reduced code size on majority of compilers (at least GCC) even without using string enum. The benefit should already overrun the potential risk.

Over the years, I have been surprised many times by toolchain behaviors (specially in embedded sw where we have many custom/specialized compilers). My approach/advice is that it is always better not to rely on toolchains and/or assume any behavior. Better to be explicit in the code (when/if we can).

abtink:
> About half of the diff is error conditions - see the diff coverage report. We currently do not have a good mechanism to test all of these error conditions.

I checked the diff coverage. I see and agree that it may not be easy to cover all error conditions.

One thing that is not clear to me is how the "commit diff" coverage value is calculated and whether it takes into account the previous code (being replaced by the change and if it was previously covered or not).

For example, the change in this PR in [`joiner_router.cpp`](https://github.com/openthread/openthread/pull/5509/files#diff-8f7d8f73c697b87503f94346449ccf9c) is combining the two `if` conditions which then causes it to remove indentation  of `message->Free()`.

Looking at [coverage report](https://codecov.io/gh/openthread/openthread/pull/5509/diff), the commit diff coverage for this file is 67%. Looks to me that the `message->Free()` line is considered as a new line and marked as not covered. Technically it is not really new code and is just a tab/indentation change and I think it is not covered in the original version as well.

Or another example is `border_agent.cpp` changes, where the  report gives a 0% "commit diff" coverage, but I guess the reason is that none of the original border-agent code is covered. 

All of these make the overall "commit diff" coverage value to a low `52%`. 

It may be good to see if we can change/configure the diff coverage calculation to take into account previous code. This sounds tricky to me though.

I think we want to use the "commit diff" coverage to indicate how much of new code/feature (in a PR) is being covered by tests. It seems to me the way it is currently calculated,  if a PR moves some exiting code (e.g., from one file to another),all of it will be considered as new and we can get a low "commit diff" coverage for the PR if the existing code (being moved) had low coverage.

Irving-cl:
@abtink 

The diff coverage only takes the new code into account. Say, if we have a line `A` which is not hit before, and then we replace it with a new line `B` in a new PR. Then the diff coverage for this PR would be 0%. And if we take a look at `border_agent.cpp` for its overall coverage in the latest commit: https://codecov.io/gh/openthread/openthread/src/21b49e2e2b9be05b996ad83fa96e4100746269d0/src/core/meshcop/border_agent.cpp It's indeed very low. So I believe it's not this PR that lower the code coverage. However I'm not sure what should we do to handle this case. Maybe we need the authors for those parts(which have low coverage) to enhance the code coverage.

abtink:
Thanks @Irving-cl . 

Going back to original question: I don't think it is a good idea to add any restriction on "diff coverage" to the PRs (like from https://github.com/openthread/openthread/pull/5283 that the diff coverage must be at least 85%.) at least until we have much higher coverage in all the existing code or if the "diff coverage" can take into account existing code's coverage somehow and not count changes where existing code is also not covered. Otherwise I feel it would be too restrictive.

src/core/meshcop/commissioner.cpp
simonlingoogle: Thanks for the explanation.

jwhui:
> * if the "diff coverage" can take into account existing code's coverage somehow and not count changes where existing code is also not covered.

I think this is addressed by the change in "total code coverage". Although, I think this metric should somehow be relative to the number of lines being changed, rather than all lines in the project.

simonlingoogle (approved):
LGTM 👍

jwhui (approved):

PR #5511, opened 2020-09-11
[joiner] add 'Joiner::State' enumeration (c++ style constants)

abtink:
Also changes discover scan callback to use `ScanResult` as its input
parameter type.


simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #5489, opened 2020-09-03
[thread-cert] Thread 1.2 CI with OTBR and Backbone link

simonlingoogle:
**Thread 1.2 CI with OTBR and Backbone link**
- [x] Run OTBR in Dockers with Backbone link
- [x] Enhance node.py to work with OTBR Docker
- [x] Add Packet Verification for 5.11.1
    - incomplete, some steps can not pass yet, just to make sure Backbone traffic verification works
- [x] Support running multiple tests simultaneously (default: each test run 3 instances)
- [x] Build OTBR Docker using OpenThread PR code
- [x] Upload code coverage in Docker
  - **Verified that Codecov in Docker works**:
    image

**Some implementation details:**
- Most existing code of `node.py` is shared by OTBR Docker
- Backbone related test scripts are found in `tests/scripts/thread-cert/backbone`
- A new script `tests/scripts/thread-cert/run_bbr_tests.py` is added to manage multiple running tests
- Test configuration differs according to `PORT_OFFSET`
  - Backbone interface name: `Backbone{PORT_OFFSET}`
  - Backbone network prefix: `91{PORT_OFFSET:02x}::/64`
  - Docker instance name: `otbr_{PORT_OFFSET}_{nodeid}`
  - Output Files:
    - Pcap: 
      - Thread: `{test_name}_{PORT_OFFSET}.pcap`
      - Backbone: `{test_name}_{PORT_OFFSET}_backbone.pcap`
      - Merged: `{test_name}_{PORT_OFFSET}_merged.pcap`
    - Log: `{test_name}_{PORT_OFFSET}.log`

PR #5512, opened 2020-09-11
[thread-1.2-cert] setup radvd service on Host

simonlingoogle:
Host need to use `radvd` service to advertise the `Domain Prefix` on the Backbone link. 

Different from Host in TestHarness certification, Host in Thread 1.2 CI does not need to advertise the Backbone prefix (910b::/64) because it's already handled by Docker network.  

todo:
- [ ] Add Backbone test to verify that Host advertises `Domain Prefix` with correct flags periodically 

Depends on #5489


simonlingoogle:
No longer needed since the functionality has been provided by #5664 (merged)

PR #5523, opened 2020-09-14
[dua] handle coap error code for DUA.rsp

simonlingoogle:
This PR:
- [x] Configure next response to 5.00 for Reference Device
- [x] Handle 5.00 (and other COAP error codes) in DUA.rsp (**required by 1.2 Test Spec**)
- [x] Update DUA test to handle COAP code 5.00

### DIFF
--- a/src/core/coap/coap.hpp
+++ b/src/core/coap/coap.hpp
@@ -437,13 +437,14 @@ public:
      *
      * @param[in]  aRequest        A reference to the CoAP Message that was used in CoAP request.
      * @param[in]  aMessageInfo    The message info corresponding to the CoAP request.
+     * @param[in]  aCode           The CoAP code of the dummy CoAP response.
      *
      * @retval OT_ERROR_NONE          Successfully enqueued the CoAP response message.
      * @retval OT_ERROR_NO_BUFS       Insufficient buffers available to send the CoAP response.
      * @retval OT_ERROR_INVALID_ARGS  The @p aRequest header is not of confirmable type.
      *
      */
-    otError SendEmptyAck(const Message &aRequest, const Ip6::MessageInfo &aMessageInfo);
+    otError SendEmptyAck(const Message &aRequest, const Ip6::MessageInfo &aMessageInfo, Code aCode = kCodeChanged);
 
     /**
      * This method sends a header-only CoAP message to indicate no resource matched for the request.
--- a/src/core/thread/dua_manager.cpp
+++ b/src/core/thread/dua_manager.cpp
@@ -529,7 +529,7 @@ void DuaManager::HandleDuaResponse(Coap::Message &aMessage, const Ip6::MessageIn
     VerifyOrExit(aMessage.GetCode() == Coap::kCodeChanged || aMessage.GetCode() >= Coap::kCodeBadRequest,
                  error = OT_ERROR_PARSE);
 
-    error = ProcessDuaResponse(aMessage, &mRegisteringDua);
+    error = ProcessDuaResponse(aMessage);
 
 exit:
     if (error != OT_ERROR_RESPONSE_TIMEOUT)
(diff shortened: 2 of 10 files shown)


tests/scripts/thread-cert/node.py
```diff
@@ -544,7 +544,13 @@ def remove_domain_prefix(self, prefix):
         self.remove_prefix(prefix)
         self.register_netdata()
 
-    def set_next_dua_response(self, status, iid=None):
+    def set_next_dua_response(self, status: Union[str, int], iid=None):
+        # Convert 5.00 to COAP CODE 160
+        if isinstance(status, str):
+            assert '.' in status
+            status = status.split('.')
+            status = (int(status[0]) << 5) + int(status[1])
```
Irving-cl, line 552: Without much background about coap code, does this formula only works for `5.00 -> 160`? If `status` is other value, is it also correct? If this only works for `5.00`, I think we can just compare the value and set it to `160`.

src/core/thread/dua_manager.cpp
```diff
@@ -520,6 +521,8 @@ void DuaManager::HandleDuaResponse(Coap::Message &aMessage, const Ip6::MessageIn
     OT_UNUSED_VARIABLE(aMessageInfo);
     otError error;
 
+    otLogCritDua("HandleDuaResponse: aResult=%s, code=%d", otThreadErrorToString(aResult), aMessage.GetCode());
```
Irving-cl, line 524: `otLogCrit` is usually used when some error happen. Use `Info` here?

Irving-cl (approved):
LGTM👍

Small suggestions.

tests/scripts/thread-cert/node.py
simonlingoogle: Actually it works for all COAP code, but only 5.00 is useful. I tend to make the code general rather than only handling 5.00 because it's just simple to implement in Python.

src/core/thread/dua_manager.cpp
simonlingoogle: This is debug log. Will remove. Thanks.

src/core/coap/coap.hpp
```diff
@@ -443,7 +443,7 @@ class CoapBase : public InstanceLocator, private NonCopyable
      * @retval OT_ERROR_INVALID_ARGS  The @p aRequest header is not of confirmable type.
      *
      */
-    otError SendEmptyAck(const Message &aRequest, const Ip6::MessageInfo &aMessageInfo);
+    otError SendEmptyAck(const Message &aRequest, const Ip6::MessageInfo &aMessageInfo, Code aCode = kCodeChanged);
```
librasungirl, line 446: add intro for `aCode`?

src/core/thread/dua_manager.cpp
```diff
@@ -528,9 +529,11 @@ void DuaManager::HandleDuaResponse(Coap::Message &aMessage, const Ip6::MessageIn
         ExitNow(error = aResult);
     }
 
-    VerifyOrExit(aResult == OT_ERROR_NONE && aMessage.GetCode() == Coap::kCodeChanged, error = OT_ERROR_PARSE);
+    VerifyOrExit(aResult == OT_ERROR_NONE, error = OT_ERROR_PARSE);
+    VerifyOrExit(aMessage.GetCode() == Coap::kCodeChanged || aMessage.GetCode() >= Coap::kCodeBadRequest,
+                 error = OT_ERROR_PARSE);
 
-    error = ProcessDuaResponse(aMessage);
+    error = ProcessDuaResponse(aMessage, &mRegisteringDua);
```
librasungirl, line 536: It might not be necessary to pass `mRegisteringDua`?

simonlingoogle: It seems not necessary in most cases. 

In the current implementation, `mRegisteringDua` is used in the `ProcessDuaResponse` method only when the COAP response code is not `2.04`. Most of time `ProcessDuaResponse` handles `DUA.rsp` in such case `mRegisteringDua` must be the DUA that is registering. 

But note that `ProcessDuaResponse` is also called by `HandleDuaNotification` which handles proactive DUA notification (`n/dn`). For `n/dn` message with code != 2.04, even though very unlikely to receive, we can't assume that the corresponding DUA is `mRegisteringDua`. 

Thoughts?

simonlingoogle: I realized that the COAP code for `n/dn` must be POST (0.02). So, there is a bug int the current code. Will fix.

simonlingoogle: fixed, please check

librasungirl (approved):

jwhui (approved):

PR #5528, opened 2020-09-14
[cli] misc enhancements

abtink:
This commit contains a group of smaller enhancements in `Cli` modules:

- Add and use `OutputLine()` helper method which outputs a formatted 
  string  line (appends newline `\r\n`).
- Use binary search when finding the command in `sCommands` array (also
  adds`static_assert` check to ensure `sCommands` array is sorted).
- Define `Command` struct as a `private` nested type.
- Use `enum` constants (replacing `#define`).
- Rename method to `OutputResult()` (from `AppendReult`) to harmonize
  with rest of a `Output{Something}()` methods.
- Remove comments at `#endif` when within 20 lines of `#if`

--------

I kept the commits separate for easier review but would suggest to squash them all before merge.

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -366,7 +366,7 @@ exit:
     return rval;
 }
 
-void Interpreter::AppendResult(otError aError)
+void Interpreter::OutputResult(otError aError)
 {
     if (aError == OT_ERROR_NONE)
     {
@@ -493,7 +493,7 @@ void Interpreter::ProcessHelp(uint8_t aArgsLength, char *aArgs[])
         OutputFormat("%s\r\n", mUserCommands[i].mName);
     }
 
-    AppendResult(OT_ERROR_NONE);
+    OutputResult(OT_ERROR_NONE);
 }
 
 #if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
@@ -561,7 +561,7 @@ void Interpreter::ProcessBackboneRouter(uint8_t aArgsLength, char *aArgs[])
 
 exit:
 #endif // OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
-    AppendResult(error);
+    OutputResult(error);
 }
 
 #if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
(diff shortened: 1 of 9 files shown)


simonlingoogle (approved):
LGTM :+1: 
Happy to see code size reduced.

bukepo (approved):
LGTM:+1:

src/cli/cli.cpp
```diff
@@ -4416,58 +4256,79 @@ void Interpreter::ProcessDiag(uint8_t aArgsLength, char *aArgs[])
...
-                mUserCommands[i].mCommand(aArgsLength - 1, &aArgs[1]);
-                break;
-            }
-        }
-
-        if (i == mUserCommandsLength)
-        {
-            AppendResult(OT_ERROR_INVALID_COMMAND);
+            mUserCommands[i].mCommand(aArgsLength - 1, &aArgs[1]);
+            ExitNow();
```
Irving-cl, line 4326: Change `FindCommand` so that it could also be applied to `UserCommands`?

Irving-cl (approved):
LGTM👍

src/cli/cli.cpp
abtink: We cannot easily extend `FindCommand()` to use for both  `sCommand` and `mUserCommands` arrays:
- Unlike the `sCommands` we cannot know if the `mUserCommands` list is sorted or not at compile time (it's given to us at run-time), so cannot always use binary search for it.
- Another point is that their underlying types are different (one uses `Command` struct and the other one uses `otCliCommand`), most notably their handler type `mCommand` are different:
   - `Command` uses an `Interpreter` method pointer 
   ```void (Interpreter::*mCommand)(uint8_t aArgsLength, char *aArgs[])``` 
   -  but `otCliCommand` uses a normal function pointer `void (*mCommand)(uint8_t aArgsLength, char *aArgs[])`

Irving-cl: Got it. Thanks.

src/cli/cli.cpp
```diff
@@ -485,17 +332,19 @@ void Interpreter::ProcessHelp(uint8_t aArgsLength, char *aArgs[])
     OT_UNUSED_VARIABLE(aArgsLength);
     OT_UNUSED_VARIABLE(aArgs);
 
+    static_assert(IsArraySorted(sCommands, OT_ARRAY_LENGTH(sCommands)), "Command list is not sorted");
```
jwhui, line 335: Can/should we apply these optimizations to sub-commands (e.g. `dataset`)?

abtink: Good idea. I think we should. Will add this later.

I am also considering (for future) to add a bunch of helper methods in `CmdLineParser` (arg lookup based on a table, string match, parse as number, etc). This way they can be shared/used by CLI and factory diags.

jwhui (approved):

PR #5531, opened 2020-09-15
[github-actions] break up the 'build' and 'simulation' workflows

abtink:
This commit breaks up the `build` and `simulation` github-actions
workflows into multiple separate smaller workflows. This allows re-run
of a smaller set of builds/tests in case of random/spurious failures.


.github/workflows/simulation.yml
```diff
@@ -1,436 +0,0 @@
...
-#  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
-#  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
-#  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
-#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
-#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
-#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
-#  POSSIBILITY OF SUCH DAMAGE.
-#
-
-name: Simulation
```
simonlingoogle, line 29: What would happen to the badge icon: Simulation in the README.md

.github/workflows/build.yml
```diff
@@ -26,7 +26,7 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
-name: Build
```
simonlingoogle, line 29: What would happen to the Build badge in README.md?

.github/workflows/simulation.yml
abtink: Ideally I think we want it to indicate aggregation of all `Build` related workflows (show it `passing` if al of them arel passing). I don't know if this is possible or not.

Thoughts? Suggestions?

jwhui: I don't think GItHub Actions provides an ability to aggregate status into a single badge.

abtink: I could not find any immediate solution to enable aggression 

We can consider including badges for a subset of the workflows in `README.md` file?

Thoughts?

simonlingoogle: My suggestion would be:

Build (arm-gcc)/badge.svg) Build (gcc)/badge.svg) Build (clang)/badge.svg) Build (macos)/badge.svg) Simulation (cli)/badge.svg) Simulation (cli)/badge.svg) Simulation 1.2 POSIX Docker

Is it too many ?

abtink:
I am putting this on hold for now (monitoring how often different workflow builds fail and need re-kick).

simonlingoogle:
> I am putting this on hold for now (monitoring how often different workflow builds fail and need re-kick).

The chance each workflow might fail varies over time? So it's less meaningful to divide workflows by fail chance, but more meaningful to divide by testing area?

PR #5532, opened 2020-09-15
[cli] implement setting/getting cca threshold

dismirlian:
Hi all, this PR adds a command to the CLI to be able to set (and get) the CCA threshold. This is useful when there is a LNA in the radio's RX path.

This function is present in wpantund, but wasn't in ot-ctl.

Thanks!


brianljt (approved):
LGTM. I want to point out that, in IEEE 802.15.4 spec, it points out the RF power measurement should be done at antenna port.

> 10.1.4 RF power measurement
Unless otherwise stated, all RF power measurements, either transmit or receive, shall be made at the appropriate transceiver to antenna connector. The measurements shall be made with equipment that is either matched to the impedance of the antenna connector or corrected for any mismatch. For devices without an antenna connector, the measurements shall be interpreted as effective isotropic radiated power (EIRP) and any radiated measurements shall be corrected to compensate for the antenna gain in the implementation.

So when using cli to set/get CCA threshold, I think user would expect to set/get the real CCA measurement on antenna port and will not need to take into consideration of any external FEM used or board attenuation. I.E. for two boards with and without external FEM, setting CCA threshold of -74dBm should have similar effects on both boards. This FEM related gain/attenuation should be handled in `Radio` layer and reflected in `otPlatRadioSetCcaEnergyDetectThreshold`.

src/cli/cli.cpp
```diff
@@ -674,6 +674,29 @@ void Interpreter::ProcessBufferInfo(uint8_t aArgsLength, char *aArgs[])
...
+        int8_t cca;
+
+        SuccessOrExit(error = otPlatRadioGetCcaEnergyDetectThreshold(mInstance, &cca));
+        OutputLine("%d dBm", cca);
+    }
+    else
+    {
+        long value;
+
+        SuccessOrExit(error = ParseLong(aArgs[0], value));
```
simonlingoogle, line 692: Validate the value range of `value`?

src/cli/cli.hpp
```diff
@@ -542,6 +543,7 @@ class Interpreter
         {"bbr", &Interpreter::ProcessBackboneRouter},
 #endif
         {"bufferinfo", &Interpreter::ProcessBufferInfo},
+        {"ccathreshold", &Interpreter::ProcessCcaThreshold},
```
simonlingoogle, line 546: Also update doc in cli/README.md ?

bukepo:
I'm fine to add this command. But is adding a separate API to set the LNA gain a better approach to solve the issue in description?

I think there may be different CCA threshold requirements in different regulatory domain, setting a fixed LNA value allows OpenThread to automatically adjust the threshold value according to the regulatory domain settings.

Thoughts? @jwhui @brianljt

brianljt:
> I'm fine to add this command. But is adding a separate API to set the LNA gain a better approach to solve the issue in description?
> 
> I think there may be different CCA threshold requirements in different regulatory domain, setting a fixed LNA value allows OpenThread to automatically adjust the threshold value according to the regulatory domain settings.
> 
> Thoughts? @jwhui @brianljt

I'm thinking whether we need to expose LNA gain to OpenThread. In my opinion, this information should be handled at `Radio` layer. So for different regulatory domain, OT just needs to set the target CCA measured at antenna port, `Radio` driver then will offset the necessary gain from it and set it to its chip registers. For development phase, it is OK to use this cli command to adjust the threshold on the fly, but once the gain is measured in products/boards, it should be handled at lower layer so this get/set command reflects only the target threshold.

bukepo:
> > I'm fine to add this command. But is adding a separate API to set the LNA gain a better approach to solve the issue in description?
> > I think there may be different CCA threshold requirements in different regulatory domain, setting a fixed LNA value allows OpenThread to automatically adjust the threshold value according to the regulatory domain settings.
> > Thoughts? @jwhui @brianljt
> 
> I'm thinking whether we need to expose LNA gain to OpenThread. In my opinion, this information should be handled at `Radio` layer. So for different regulatory domain, OT just needs to set the target CCA measured at antenna port, `Radio` driver then will offset the necessary gain from it and set it to its chip registers. For development phase, it is OK to use this cli command to adjust the threshold on the fly, but once the gain is measured in products/boards, it should be handled at lower layer so this get/set command reflects only the target threshold.

Not all OpenThread users will build the firmware from source code. They may choose a vendor's 15.4 chip and directly and use the vendor's **certified release of OpenThread** for simplicity. But they may choose any FEM whose LNA gain is unknown at first. Having an API to tell the radio driver the FEM's LNA gain saves the trouble of developers to deal with the regulatory domain issue because the CCA threshold values should be fixed in each regulatory domain. Thoughts?

dismirlian:
> > I'm fine to add this command. But is adding a separate API to set the LNA gain a better approach to solve the issue in description?
> > I think there may be different CCA threshold requirements in different regulatory domain, setting a fixed LNA value allows OpenThread to automatically adjust the threshold value according to the regulatory domain settings.
> > Thoughts? @jwhui @brianljt
> 
> I'm thinking whether we need to expose LNA gain to OpenThread. In my opinion, this information should be handled at `Radio` layer. So for different regulatory domain, OT just needs to set the target CCA measured at antenna port, `Radio` driver then will offset the necessary gain from it and set it to its chip registers. For development phase, it is OK to use this cli command to adjust the threshold on the fly, but once the gain is measured in products/boards, it should be handled at lower layer so this get/set command reflects only the target threshold.

I agree with @brianljt that this should be best handled at platform level. For me, however, the main purpose of this command is to allow the evaluation of OT's performance with different hardware configurations (with/without FEM, etc).

Also, please note that there are similar issues with the `txpower` command.

dismirlian:
> Not all OpenThread users will build the firmware from source code. They may choose a vendor's 15.4 chip and directly and use the vendor's **certified release of OpenThread** for simplicity. But they may choose any FEM whose LNA gain is unknown at first. Having an API to tell the radio driver the FEM's LNA gain saves the trouble of developers to deal with the regulatory domain issue because the CCA threshold values should be fixed in each regulatory domain. Thoughts?

This is also a good point. Bottomline, IMHO it should be up to the platform to cap the TX power and handle the CCA threshold appropriately.

src/cli/cli.cpp
dismirlian: Yes, I modeled this exactly as the `ProcessTxPower` (it's almost a copy & paste):
https://github.com/openthread/openthread/blob/b735193857cdbd490cb71f274c9794e4b8995e39/src/cli/cli.cpp#L3760-L3765

If you want I can change both methods to validate this cast/truncation.

src/cli/cli.hpp
```diff
@@ -296,6 +296,7 @@ class Interpreter
 
 #endif // (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
 
+    void ProcessCcaThreshold(uint8_t aArgsLength, char *aArgs[]);
```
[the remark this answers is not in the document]
dismirlian: Done

src/cli/cli.hpp
dismirlian: Done

brianljt:
> > > I'm fine to add this command. But is adding a separate API to set the LNA gain a better approach to solve the issue in description?
> > > I think there may be different CCA threshold requirements in different regulatory domain, setting a fixed LNA value allows OpenThread to automatically adjust the threshold value according to the regulatory domain settings.
> > > Thoughts? @jwhui @brianljt
> > 
> > 
> > I'm thinking whether we need to expose LNA gain to OpenThread. In my opinion, this information should be handled at `Radio` layer. So for different regulatory domain, OT just needs to set the target CCA measured at antenna port, `Radio` driver then will offset the necessary gain from it and set it to its chip registers. For development phase, it is OK to use this cli command to adjust the threshold on the fly, but once the gain is measured in products/boards, it should be handled at lower layer so this get/set command reflects only the target threshold.
> 
> Not all OpenThread users will build the firmware from source code. They may choose a vendor's 15.4 chip and directly and use the vendor's **certified release of OpenThread** for simplicity. But they may choose any FEM whose LNA gain is unknown at first. Having an API to tell the radio driver the FEM's LNA gain saves the trouble of developers to deal with the regulatory domain issue because the CCA threshold values should be fixed in each regulatory domain. Thoughts?

Thank you @bukepo, this makes sense to me. If so, I think I agree to add two cli commands. One is get/set CCA threshold and another is get/set FEM PA/LNA gain. Corresponding `otPlatRadioXXX` APIs should be added to pass this FEM information to `Radio`.

bukepo (approved):
Agree to add address the LNA gain value in a separate PR.

bukepo:
The code is not covered, could you please add a tests in `tests/scripts/expect`, you may reference https://github.com/openthread/openthread/blob/master/tests/scripts/expect/cli-log-level.exp

dismirlian:
> The code is not covered, could you please add a tests in `tests/scripts/expect`, you may reference https://github.com/openthread/openthread/blob/master/tests/scripts/expect/cli-log-level.exp

Added.

bukepo (approved):
LGTM:+1:

src/cli/cli.cpp
simonlingoogle: Fine to change both or stay in the scope of this PR.

simonlingoogle (approved):
LGTM :+1: 
with a nit

jwhui (approved):

src/cli/cli.cpp
dismirlian: Ok, if you agree we can merge this as it is, and I'll open an issue to address the range validation in `ccathreshold` and `txpower` (and possibly others?).

PR #5540, opened 2020-09-17
[cli] return error from 'Process{Cmd}()' methods, adding 'OT_ERROR_PENDING'

abtink:
This commit changes the CLI `Process{Cmd}()` handler methods to return
an `otError` (instead of outputting the result). This helps harmonize
the `Process{Cmd}()` handlers in different CLI modules. It also
enables the result of CLI command to be outputted from `ProcessLine()`
after the command is looked up in the table and the corresponding
`Process{Cmd}()` handler is invoked.

This commit also adds a new error code `OT_ERROR_PENDING`. This
special error code indicates the success or error status of an
operation is not yet known and is pending. This is then used by some
of the CLI `Process{Cmd}()` methods which perform an async operation,
e.g., `ProcessDiscover()` returns `OT_ERROR_PENDING` if the discover
scan starts successfully, and later the final result is outputted from
scan callback. The `OutputResult(aError)` method is changed to output
nothing if the error is `OT_ERROR_PENDING`.

### DIFF
--- a/include/openthread/cli.h
+++ b/include/openthread/cli.h
@@ -109,6 +109,7 @@ void otCliUartInit(otInstance *aInstance);
  *
  * @param[in]  aUserCommands  A pointer to an array with user commands.
  * @param[in]  aLength        @p aUserCommands length.
+ *
  */
 void otCliSetUserCommands(const otCliCommand *aUserCommands, uint8_t aLength);
 
@@ -117,6 +118,7 @@ void otCliSetUserCommands(const otCliCommand *aUserCommands, uint8_t aLength);
  *
  * @param[in]  aBytes   A pointer to data which should be printed.
  * @param[in]  aLength  @p aBytes length.
+ *
  */
 void otCliOutputBytes(const uint8_t *aBytes, uint8_t aLength);
 
@@ -125,6 +127,7 @@ void otCliOutputBytes(const uint8_t *aBytes, uint8_t aLength);
  *
  * @param[in]  aFmt   A pointer to the format string.
  * @param[in]  ...    A matching list of arguments.
+ *
  */
 void otCliOutputFormat(const char *aFmt, ...);
 
@@ -133,13 +136,17 @@ void otCliOutputFormat(const char *aFmt, ...);
  *
  * @param[in]  aString  A pointer to the string, which may not be null-terminated.
  * @param[in]  aLength  Number of bytes.
+ *
  */
 void otCliOutput(const char *aString, uint16_t aLength);
 
 /**
  * Write error code to the CLI console
  *
+ * If the @p aError is `OT_ERROR_PENDING` nothing will be outputted.
+ *
  * @param[in]  aError Error code value.
+ *
  */
 void otCliAppendResult(otError aError);
 
@@ -150,6 +157,7 @@ void otCliAppendResult(otError aError);
  * @param[in]  aLogRegion  The log region.
  * @param[in]  aFormat     A pointer to the format string.
  * @param[in]  aArgs       va_list matching aFormat.
+ *
  */
 void otCliPlatLogv(otLogLevel aLogLevel, otLogRegion aLogRegion, const char *aFormat, va_list aArgs);
 
--- a/include/openthread/error.h
+++ b/include/openthread/error.h
@@ -229,6 +229,12 @@ typedef enum OT_MUST_USE_RESULT otError
     OT_ERROR_INVALID_COMMAND = 35,
 
     /**
+     * Special error code used to indicate success/error status is pending and not yet known.
+     *
+     */
+    OT_ERROR_PENDING = 36,
+
+    /**
      * The number of defined errors.
      */
     OT_NUM_ERRORS,
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (28)
+#define OPENTHREAD_API_VERSION (29)
 
 /**
  * @addtogroup api-instance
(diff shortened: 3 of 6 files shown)


simonlingoogle (approved):
LGTM :+1:

simonlingoogle:
Beyond the scope of this PR, I think this PR actually bring us one step closer to implement **synchronized commands**. 
By synchronized commands, I mean OT CLI only execute one outstanding command at any time, and no subsequent command will be executed before the previous command output result (i.e. `Done` or `Error ...`).
I think this is helpful for programs to interact with OT CLI and parse command outputs. Currently, when multiple **async** commands are running, it's impossible to tell which `Done` / `Error ...` is of which command.

There should be two ways to achieve this:
1. Block other commands if there is an `async` command pending. This is now easier to implement base on this PR.
2. All commands should output `Done` / `Error ...` immediately and other `async` events (e.g. ping result, scan result) is emitted asynchronously, but they are not seen as part of the command output.

Approach 1 is easier to use if the `async` operations finished within a short period of time. But if a command blocks for a long duration (e.g. waiting for COAP retransmissions and timeout), no other commands can be issued and OT CLI is less responsive in such situation. 
Approach 2 is easy to implement but OT CLI users needs to be able to extract and handle `async` events from CLI output. 

Personally I prefer approach 2 because I think it's more scalable. 

As OT CLI becomes more widely used, I think it's more of an public OpenThread interface rather than some code example for invoking OpenThread APIs. From such point, we should make the OT CLI output more consistent (no mixed output of multiple commands).

Thoughts? @abtink @bukepo @jwhui

gjc13 (approved):

abtink:
@simonlingoogle , in general having one outstanding command being executed at each point sound reasonable to me. The main thing is how to block a next command: we either need a queue of commands or a mechanism to not allow user to issue next command if one is ongoing.

simonlingoogle:
> @simonlingoogle , in general having one outstanding command being executed at each point sound reasonable to me. The main thing is how to block a next command: we either need a queue of commands or a mechanism to not allow user to issue next command if one is ongoing.

Yes. And the pending command queue can be indefinitely long. One solution might be ignoring the next command if one is ongoing and relying on  CLI users to always send next command when the previous command is done. 

This is a reason why I prefers Approach 2 of https://github.com/openthread/openthread/pull/5540#issuecomment-693777489

jwhui (approved):

wgtdkp:
> This commit also adds a new error code OT_ERROR_PENDING. This
special error code indicates the success or error status of an
operation is not yet known and is pending. This is then used by some
of the CLI Process{Cmd}() methods which perform an async operation,
e.g., ProcessDiscover() returns OT_ERROR_PENDING if the discover
scan starts successfully, and later the final result is outputted from
scan callback. The OutputResult(aError) method is changed to output
nothing if the error is OT_ERROR_PENDING.

@abtink This makes me feel that we should rename `OT_ERROR_XXX` to `OT_STATUS_XXX`.
Besides`OT_ERROR_PENDING`, `OT_STATUS_OK` also looks better than `OT_ERROR_NONE` to me.

abtink:
> @abtink This makes me feel that we should rename OT_ERROR_XXX to OT_STATUS_XXX.
BesidesOT_ERROR_PENDING, OT_STATUS_OK also looks better than OT_ERROR_NONE to me.

While I agree renaming makes these two cases read semantically better, changing the `OT_ERROR_XXX` -> `OT_STATUS_XXX` would not only require a lot of changes inside the `core` module, but since `otError` definition is part of public definition of OpenThread and is used by a lot of public APIs, it would also impact the users and platform provides of OpenThread (all the code/modules where OT APIs are being used and/or all OT platform layer implementation need to be updated, e.g. I know internally we do have a lot of modules which would be impacted, also I expect a lot of third party code needs to be changed). 

This would be rather disruptive change and I think we should consider it only if/when there are very good justifications/reasons for it.  In my view, having some error cases sound/read better semantically may not be a good enough justification for such a big/disruptive change.

In general since OT is being used/integrated in more user's code, we should aim to keep OT public definitions and API remain untouched (as much as possible) to ensure higher level of backward compatibility with user's existing code.

wgtdkp:
> > @abtink This makes me feel that we should rename OT_ERROR_XXX to OT_STATUS_XXX.
> > BesidesOT_ERROR_PENDING, OT_STATUS_OK also looks better than OT_ERROR_NONE to me.
> 
> While I agree renaming makes these two cases read semantically better, changing the `OT_ERROR_XXX` -> `OT_STATUS_XXX` would not only require a lot of changes inside the `core` module, but since `otError` definition is part of public definition of OpenThread and is used by a lot of public APIs, it would also impact the users and platform providers of OpenThread (all the code/modules where OT APIs are being used and/or all OT platform layer implementation need to be updated, e.g. I know internally we do have a lot of modules which would be impacted, also I expect a lot of third party code needs to be changed).
> 
> This would be rather disruptive change and I think we should consider it only if/when there are very good justifications/reasons for it. In my view, having some error cases sound/read better semantically may not be a good enough justification for such a big/disruptive change.
> 
> In general since OT is being used/integrated in more user's code, we should aim to keep OT public definitions and API remain untouched (as much as possible) to ensure higher level of backward compatibility with user's existing code.

@abtink Totally agree with you. Just feeling that `otStatus` could be a better name :)

simonlingoogle:
> > > Besides OT_ERROR_PENDING, OT_STATUS_OK also looks better than OT_ERROR_NONE to me.

> @abtink Totally agree with you. Just feeling that `otStatus` could be a better name :)

Agree with @abtink that `reading semantically better` is not a strong enough reason to change public APIs in such a large scale.

Furthermore, I think it's arguable whether or not `OT_STATUS_OK` is better than `OT_ERROR_NONE`. Neither does `IgnoreStatus` read semantically better than `IgnoreError`. 

`OT_STATUS_PENDING` does read better than `OT_ERROR_PENDING`, but that's simply because `OT_ERROR_PENDING` is not a true error. I thought about suggesting not to introduce `OT_ERROR_PENDING` in this PR but a second thought was that it's just simpler to implement with `OT_ERROR_PENDING`.

PR #5544, opened 2020-09-18
[cli] add api version command

Moandor-y:
This adds a CLI command to get the API version number.


bukepo:
there's a `version` command, how about adding a subcommand `api` to get the API version number?

Moandor-y:
> there's a `version` command, how about adding a subcommand `api` to get the API version number?

Sounds good. Will change to `version` command.

tests/scripts/expect/cli-misc.exp
```diff
@@ -111,6 +111,11 @@ expect "Done"
 
 send "version\n"
 expect "Done"
+send "version api\n"
+expect "28"
```
simonlingoogle, line 115: this will fail when API version is increased to 29.
```suggestion
expect -re {(\d+)}
```

Moandor-y: I think we should also change here when the API version is increased.

simonlingoogle: Would it be too strict? Having to change two places for API version change also seems redundant. Thoughts?

Moandor-y: Indeed, that's somewhat redundant. I was thinking about checking the correctness of the CLI function implementation, but with a function simple as that it seems unnecessary.

simonlingoogle (approved):
LGTM :+1: 
Thanks

src/cli/cli.cpp
```diff
@@ -3835,12 +3835,26 @@ void Interpreter::ProcessUnsecurePort(uint8_t aArgsLength, char *aArgs[])
...
+        ExitNow();
+    }
+
+    if (strcmp(aArgs[0], "api") == 0)
+    {
+        OutputLine("%d", OPENTHREAD_API_VERSION);
+    }
+    else
+    {
+        ExitNow(error = OT_ERROR_INVALID_ARGS);
```
librasungirl, line 3853: Should the error be `OT_ERROR_INVALID_COMMAND`?

Moandor-y: I think both are OK. The difference is to consider this as a subcommand or as an argument. I've seen some commands using mixed ideas, e.g. `mac` outputs invalid args but `mac something_invalid` outputs invalid command.

Irving-cl (approved):
LGTM👍

bukepo (approved):
LGTM:+1:

src/cli/cli.cpp
```diff
@@ -3835,12 +3835,26 @@ void Interpreter::ProcessUnsecurePort(uint8_t aArgsLength, char *aArgs[])
...
-    OT_UNUSED_VARIABLE(aArgs);
+    otError error = OT_ERROR_NONE;
 
-    const char *version = otGetVersionString();
-    OutputLine("%s", static_cast<const char *>(version));
-    OutputResult(OT_ERROR_NONE);
+    if (aArgsLength == 0)
+    {
+        const char *version = otGetVersionString();
+        OutputLine("%s", static_cast<const char *>(version));
```
abtink, line 3843: I know it's not part the change in this PR/commit, but since these lines are being included (and tabbed) in a new`if`, we can improve the code as well (I think there is no need to `static_cast` from same type and can inline the call).
```suggestion
        OutputLine("%s", otGetVersionString());
```

abtink (approved):
LGTM.
One suggestion below. Thanks

wgtdkp (approved):

src/cli/cli.cpp
librasungirl: My understanding is if the command or sub-command accepts parameters and the parameters are invalid, the error code would be `OT_ERROR_INVALID_ARGS`, otherwise `OT_ERROR_INVALID_COMMAND` should be generally used to indicate the command or sub-command is not supported.

The command `version`,  it doesn't have any parameters; it could support sub-command, any following word after `version` should be interpreted as a sub-command, and check whether or not it's a valid sub-command in the valid sub-command set (here only one `api` element).

I think we'd better to define a general rule to follow, other than the ambiguity in existing usages in OT like the example of `mac`that @Moandor-y  mentioned.

@wgtdkp your thoughts? is `mac` intentionally designed as is now?

Moandor-y: I think you are right. In the documentations something marked with brackets (like `<port>`) should be considered as an argument, otherwise it would have a separate entry in the command list and should be considered as a sub-command. In this case `version api` is a sub-command.

librasungirl (approved):

src/cli/cli.cpp
```diff
@@ -3685,13 +3685,25 @@ otError Interpreter::ProcessUnsecurePort(uint8_t aArgsLength, char *aArgs[])
...
+    {
+        OutputLine("%d", OPENTHREAD_API_VERSION);
+    }
+    else
+    {
+        ExitNow(error = OT_ERROR_INVALID_COMMAND);
+    }
+
+exit:
+    OutputResult(error);
```
librasungirl, line 3706: need return value.

jwhui, line 3706: Please address the check failures below.

Moandor-y: Done.

src/cli/cli.cpp
wgtdkp: @librasungirl Agree that we should follow a common rule of returning INVALID_COMMAND or INVALID_ARGS. The `mac` command returns `OT_INVALID_ARGS` is not intentional but an escaped case when introduced `OT_INVALID_COMMAND` (Will fix it soon).

@Moandor-y proposed a nice way to distinguish between an argument and sub-command:

> In the documentations something marked with brackets (like <port>) should be considered as an argument, otherwise it would have a separate entry in the command list and should be considered as a sub-command.

So, in the documentation:
1. Arguments should always be surrounded by `<>`.
2. Commands and sub-commands should always use non-escaped plain text in lowercase.

In the CLI interpreter:
1. Always return `OT_INVALID_COMMAND` for commands & sub-commands.

simonlingoogle: Prefer `OT_ERROR_INVALID_COMMAND` slightly because we are searching for a sub command here.

jwhui (approved):

PR #5545, opened 2020-09-18
[low-power] enhance CSL transmission on NRF52840 using `transmit_at`

Irving-cl:
This PR adds a new platform api `otPlatRadioTransmitAfter` which allows user to transmit frame after a specified delay. This api is currently only implemented on `nrf52` platform. And this new api is used in CSL transmission code.

After hundreds of testing, I found that if I added some delay when doing the CSL tx like:
```
Get<Radio>().TransmitAfter(mTransmitFrame, 
        (phaseDesired - phaseNow) * kUsPerTenSymbols + kCslTransmitAfterCalibration);
```
It's more likely to succeed. Otherwise it rarely succeeded. 

I tried using CSL debug option and I found that the tx did occur in the CSL sample period. I guess the possible reason is `nrf_802154_trasmit_raw_at` so accurate and the frame was sent a little before the receiver starts to sample(due to the some delay on its side). But current implementation still has failed tx sometimes and I cannot find the reasons. Could you take a look and share your thoughts on this? 

@brianljt @edmont


include/openthread/platform/radio.h
```diff
@@ -672,6 +672,26 @@ otRadioFrame *otPlatRadioGetTransmitBuffer(otInstance *aInstance);
  */
 otError otPlatRadioTransmit(otInstance *aInstance, otRadioFrame *aFrame);
 
+/**
+ * Schedule a transmission after a specified delay @p aDelay.
+ *
+ * The caller must form the IEEE 802.15.4 frame in the buffer provided by `otPlatRadioGetTransmitBuffer()` before
+ * requesting transmission.  The channel and transmit power are also included in the otRadioFrame structure.
```
edmont, line 679:
```suggestion
 * requesting transmission. The channel and transmit power are also included in the otRadioFrame structure.
```

src/core/radio/radio.hpp
```diff
@@ -466,6 +466,23 @@ class Radio : public InstanceLocator, private NonCopyable
      */
     otError Transmit(Mac::TxFrame &aFrame);
 
+#if OPENTHREAD_CONFIG_CSL_USE_DELAY_TRANSMIT
+    /**
+     * This method starts the transmit sequence on the radio after a specified delay time.
+     *
+     * The caller must form the IEEE 802.15.4 frame in the buffer provided by `GetTransmitBuffer()` before
+     * requesting transmission.  The channel and transmit power are also included in the frame.
```
edmont, line 474:
```suggestion
     * requesting transmission. The channel and transmit power are also included in the frame.
```

src/core/mac/sub_mac.cpp
```diff
@@ -355,11 +355,21 @@ void SubMac::StartCsmaBackoff(void)
...
+                                           (phaseDesired - phaseNow) * kUsPerTenSymbols + kCslTransmitAfterCalibration);
+#else
                 mTimer.Start((phaseDesired - phaseNow) * kUsPerTenSymbols);
+#endif
             }
             else if (phaseNow > phaseDesired)
             {
+#if OPENTHREAD_CONFIG_CSL_USE_DELAY_TRANSMIT
+                Get<Radio>().TransmitAfter(mTransmitFrame,
+                                           (phaseDesired - phaseNow) * kUsPerTenSymbols + kCslTransmitAfterCalibration);
```
edmont, line 369: `mPeriod` should be added here as well.

src/core/mac/sub_mac.hpp
```diff
@@ -501,6 +501,9 @@ class SubMac : public InstanceLocator
         kEnergyScanRssiSampleInterval = 128, ///< RSSI sample interval during energy scan, 128 usec
 #else
         kEnergyScanRssiSampleInterval = 1, ///< RSSI sample interval during energy scan, 1 ms
+#endif
+#if OPENTHREAD_CONFIG_CSL_USE_DELAY_TRANSMIT
+        kCslTransmitAfterCalibration = 10, ///< Calibration time in units of microseconds when using `TransmitAfter`
```
edmont, line 506: If needed, I would keep this value platform specific.

edmont (commented):
Thanks @Irving-cl, I'll perform some testing and see what could be happening.

abtink:
> This PR adds a new platform api otPlatRadioTransmitAfter which allows user to transmit frame after a specified delay. This api is currently only implemented on nrf52 platform. And this new api is used in CSL transmission code.

@Irving-cl , I have two suggestions here:
- I would suggest instead of adding new API `otPlatRadioTransmitAfter` we use the existing TX API and add delay behavior through the paramters/fields in `otRadioFrame`. We can add a `mTxDelay` in `mTxInfo` part of `otRadioFrame` structure to request a delay for the frame tx (it can be set to zero by default).
- Follow on though on this: I think we already have mechanism to request the radio platform  to tx a frame at given time (used for CSL tx). I guess this is done by parameters in `mTxInfo` part of `otRadioFrame` structure (`mPhase/mPeriod`).  Can we consider making the model more generic such that we just specify an exact tx time in `otRadioFrame`. This way the delay can be accounted for in the calculated tx-time. I feel if this can work, it would cleaner model since I don't think radio platform would need/want to be aware of details of CSL parameter such as phase/period and can simply provide a mechanism to send a frame at given specific time.

brianljt:
> I tried using CSL debug option and I found that the tx did occur in the CSL sample period. I guess the possible reason is nrf_802154_trasmit_raw_at so accurate and the frame was sent a little before the receiver starts to sample(due to the some delay on its side). But current implementation still has failed tx sometimes and I cannot find the reasons. Could you take a look and share your thoughts on this?

What is the timestamp/diff you get when you enable debug option? I think current SSED wake up time is already a little bit ahead of sample window. Could you please enlarge this a little to make it wake up earlier and see if SSED now can receive the frame?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -1216,4 +1216,19 @@ void otPlatRadioUpdateCslSampleTime(otInstance *aInstance, uint32_t aCslSampleTi
...
 }
+
+#if OPENTHREAD_CONFIG_CSL_USE_DELAY_TRANSMIT
+otError otPlatRadioTransmitAfter(otInstance *aInstance, otRadioFrame *aFrame, uint32_t aDelay)
+{
+    otError error = OT_ERROR_NONE;
+
+    aFrame->mPsdu[-1] = aFrame->mLength;
+
+    error = nrf_802154_transmit_raw_at(&aFrame->mPsdu[-1], false, (uint32_t)nrf5AlarmGetCurrentTime(), aDelay,
```
hubertmis, line 1227: I'm not sure if using `nrf5AlarmGetCurrentTime()` in precise TX function is the best approach. Calling functions like this decreases precision, because it is prone to software latency.

The idea behind `nrf_802154_transmit_raw_at()` is to provide precise TX time. The program flow should be like this:

```
nrf_802154_received_raw(frame, ..., timestamp)
{
  m_rx_frame = frame;
  m_rx_timestamp = timestamp;
}

otPlatRadioTransmitPrecisely(aFrame, ...)
{
  nrf_802154_transmit_raw_at(aFrame, ..., m_rx_timestamp, getCslPhaseInUs(m_rx_frame) + x * getCslPeriodInUs(m_rx_frame), ...);
}
```

The result is precise TX timing, related only to RX timestamp. Software latency has no impact on radio timings with such approach.

Would it be possible to change proposed functions to support code flow similar to what I suggested?

Irving-cl:
Thanks for all the input. I'm working on this for a good solution. Will update later.

Irving-cl:
Upated:

I have a general idea and some questions:
1. I agree to change the origin API `otPlatRadioTransmit` to use `nrf_802154_transmit_raw_at`, instead of adding a new API. This won't affect the feasibility of the suggestions about the way to use `nrf_802154_transmit_raw_at`.
2. Regarding to use `rx_timestamp` in `nrf_802154_transmit_raw_at`:
    * It seems that we cannot do the CSL TX **entirely** based on `rx_timestamp`and `phase` in CSL IE using `nrf_802154_transmit_raw_at`, without the timer in Openthread. As in OpenThread, operations like 'Receive', 'Transmit' are scheduled and are processed one by one in 'Mac' layer. And when one operation is ongoing, other operations are blocked. So I think we cannot directly use API that contains `nrf_802154_transmit_raw_at` in high layer (above `Mac` layer) as it would break the rule of operations in `Mac`, I.E. when the transmission by `nrf_802154_transmit_raw_at` starts, another operation is being processed, and then something unexpected would happen. So I think at least it's inevitable to use transmitter's timer in `Mac` layer so that it starts a CSL TX operation at an approximate time that's a little ahead of the accurate CSL time, as we're doing in current code.
    * However we can try to use `rx_timestamp` and `phase` with `nrf_802154_transmit_raw_at` in radio, in `otPlatRadioTransmit`. 
        * Let's first review the whole CSL TX process: 1. When the transmitter has a frame for the receiver, the `CslTxScheduler` requests the `Mac` layer to send the frame at an approximate time. (As I mentioned above, the reason for doing this is to keep the rule in `Mac` layer without blocking other operations for too much time.) 2. And then it goes to `SubMac::Send`. Currently we haven't used `nrf_802154_transmit_raw_at`, so we wait in `SubMac` until the accurate TX time comes. 3. And then when accurate time comes, it goes to the radio layer(`otPlatRadioTransmit`). 
        * Back to our case: since now we have `nrf_802154_transmit_raw_at`, we won't wait in `SubMac`, instead we directly enter the radio layer and we're ready to use `nrf_802154_transmit_raw_at` here. And I think it's feasible to use it based on `rx_timestamp` (for example, we can pass the information in `mTxInfo` in the frame). As a contrast, we can use `nrf_802154_transmit_raw_at` based on current time and the `phase` used on transmitter side(Note that this is not the one in CSL IE). I think it seems to be better if we use it based on `rx_timestamp` here. However, how do we know `x` here mentioned in @hubertmis 's comment? We want to find the nearest `x`. Then we have to get it by knowing the current time (using `nrf5AlarmGetCurrentTime()` again). How can we avoid that?

So as a summary:
1. I can only think of this way to do CSL Tx, using `nrf_802154_transmit_raw_at` in radio with `rx_timestamp`.
2. If we use `nrf_802154_transmit_raw_at` in radio layer, how can we know `x` without calling `nrf5AlarmGetCurrentTime()`? Or maybe since we're just trying to calculate `x`, it's ok for the precision problem.

Thoughts?
@abtink @brianljt @edmont @hubertmis

hubertmis:
@Irving-cl , thanks for detailed explanation. Probably I was too brief with my previous comment, but I meant something similar to what you described.

I think that OpenThread MeshForwarder or Mac should use OpenThread Timer to roughly schedule operations and calculate `x` value. However, when parameters to `nrf_802154_transmit_at()` are calculated, we should not use data from OpenThread Timer, but only from radio driver timestamps and offsets defined by protocol specification.

If rough scheduling of radio operations from OpenThread Timer is off by too much value (could be caused by unexpected software latency) and it is too late to transmit a requested frame, `nrf_802154_transmit_at()` would report and error that should be handled by the caller (calculate again with incremented `x`, perhaps use greater timing margin to request transmission a little earlier this time).

Irving-cl:
I'd like to describe the new design in a little bit detail first before I implement it:

The main concern is about the `t0` that passed to `nrf_802154_transmit_raw_at` as the base time. When using this method, the actual tx time would be `t0 + delay`. As @hubertmis mentioned, in CSL tx, it would be more accurate if we pass the `rx_timestamp` of the frame received from SSED as `t0` than choosing current time (getting by `nrf5AlarmGetCurrentTime`).

So what about we implement it like this:
1. Add a new member `mLastRxTimestamp` in `CslTxScheduler::ChildInfo`. And we update this field in `Mac::ProcessCsl`, we directly set it to `aFrame.GetTimestamp()`. And `mLastRxTimestamp` would be used as the base time `t0` for `nrf_802154_transmit_raw_at`.
    * I think this field somewhat duplicates with `mCslLastHeard`. But `mLastRxTimestamp` is in **us** precision while `mCslLastHeard` is in **ms**. And `mCslLastHeard` is mainly used to determine CSL timeout. We can consider to use `mLastRxTimestamp` to replace `mCslLastHeard`. This is not important anyway.
2. In `Mac::ProcessCsl`, we set `mCslPhase` to the raw value we get from `CslIe`.
3. We enhance `CslTxScheduler::GetNextCslTransmissionDelay` method to make it calculate the target tx delay (based on `mLastRxTimestamp`) and return. Calculating process:
```
T_now = GetNow()
t_1 = (mLastRxTimestamp + mCslPhase) % mCslPeriod  // SSED's phase in transmitter's period
t_2 = T_now % mCslPeriod

if (t_2 + TIME_AHEAD <= t_1)
    T_tx = T_now - t_2 + t_1
else
    T_tx = T_now - t_2 + t_1 + mCslPeriod

tx_delay = T_tx - mLastRxTimestamp

// * Note that we don't minus TIME_AHEAD from T_tx here because T_tx should be precise time. T_tx is used to calculate the precise delay.
```
4. We already have `tx_delay`, which is based on `mLastRxTimestamp`. We may need to add another field `mTxDelay` in `CslTxScheduler::ChildInfo`. And after we get the best child in `CslTxScheduler::RescheduleCslTx`, we set `mTxDelay` to the value we just calculated.
5. We add two new field `mLastRxTimestamp` and `mTxDelay` to `otRadioFrame.mInfo.mTxInfo`. And in `CslTxScheduler::HandleFrameRequest`, we passed the two fields from `ChildInfo` to the frame.
6. After going through some process in `Mac`, we now directly call `Radio::Transmit` in `SubMac`.
7. Finally we goes to `otPlatRadioTransmit`. And we find that the two fields `mLastRxTimestamp` and `mTxDelay` in the frame are set, so we call `nrf_802154_transmit_raw_at` with the two value.
    * I think we can then replace `mPhase` and `mPeriod` in frame with the two new fields, including that for ncp.

Thoughts? @abtink @brianljt @brianljt 

>  Can we consider making the model more generic such that we just specify an exact tx time in otRadioFrame. 

I think in this way I mentioned, we can specify the exact tx time by providing the two fields `mLastRxTimestamp` and `mTxDelay`.

hubertmis:
@Irving-cl , Thanks for updated proposal. I don't know OpenThread internals well enough to understand it fully. I just would like to make sure that at the end: `mTxDelay = cslToUs(cslPhase + x * cslPeriod)`, where `x = (GetNow() - getNowTakenWhenReceivedFrame) / clsToUs(clsPeriod)`.  Your calculation proposal is a little more complicated and I'm not sure if the result is identical.

Irving-cl:
@hubertmis 

Thanks for the information. In my calculation I just skip `x` and directly calculate the delay (based on rx_timestamp). I think our purpose for calculating `x` is just to calculate the delay, so we can directly get the delay.

hubertmis:
@Irving-cl , I'm a little confused now.
Could you explain to me values from your code snipped? Are `mCslPeriod` and `mCslPhase` values extracted from CslIe from last frame received from SSED? Are they stored in CSL units or microseconds? Is `mLastRxTimestamp` and `T_now` in microseconds?

hubertmis:
@Irving-cl , I think I understood now and it seems to me that potentially imprecise value of `T_now` can affect `tx_delay` value.

How about a little change in calcuation?
```
now = GetNow()
first_tx_window = mLastRxTimestamp + cslToUs(mCslPhase)
periods_elapsed = (now - first_tx_window) % cslToUs(mCslPeriod)

next_tx_window = first_tx_window + (periods_elapsed + 1) * cslToUs(mCslPeriod)

while (now + TIME_AHEAD >= next_tx_window) next_tx_window += cslToUs(mCslPeriod)

tx_delay = next_tx_window - mLastRxTimestamp
```
Here `periods_elapsed` is `x`.

Irving-cl:
@hubertmis Thanks very much for the suggestion. I think our algorithm are basically the same. But as I mentioned, I think we can directly get the `next_tx_window` without first calculating `periods_elapsed`.

> periods_elapsed = (now - first_tx_window) % cslToUs(mCslPeriod)

I think you mean `periods_elapsed = (now - first_tx_window) / cslToUs(mCslPeriod)` here.

So our target is to calculate `next_tx_window`, which is the nearest sample time starting from `now`. The most important property of `next_tx_window` is that **it has the same offset, position with `first_tx_window` within a period**. And we want to know the difference between `now`'s position within a period and `first_tx_window`'s position. So we can do like this:
```
next_tx_window = now - (now % cslToUs(mCslPeriod)) + (first_tx_window % cslToUs(mCslPeriod))

// now - (now % cslToUs(mCslPeriod)) is the 'start position' within a period
// Then, + (first_tx_window % cslToUs(mCslPeriod)) moves it to the same position with `first_tx_window`
```
I think so long as we have the same `now`, this would give the exactly same result of `next_tx_window` since all the operations are integer operations.(You calculation maybe one period ahead of mine, but this doesn't matter. We will calibrate it in the end). Let's have an example:
```
mLastRxTimestamp = 12345
cslToUs(mCslPhase) = 10000
now = 56789
```
Then in your algorithm:
```
first_tx_window = 22345
periods_elapsed = (56789 - 22345) / 10000 = 34444 / 10000 = 3
next_tx_window = 22345 + (3 + 1) * 10000 = 62345
// Then we calibrate next_tx_window, add a few periods possibly
```
In my algorithm:
```
next_tx_window = 56789 - (56789 % 10000) + (22345 % 10000) = 56789 - 6789 + 2345 = 52345
// Then calibrate
```
So I think our calculation gets the exactly the same position. Then we would add it by a few periods to calibrate it.

Whatever, I think our idea is the same. I'm happy to have discussions about the implementation to make sure we get the precise result.

hubertmis:
@Irving-cl , thanks for the explanation. That's right: I meant `/` instead of `%`. I've edited my previous message.

I still don't fully understand your algorithm, but I just wanted to make sure it gives the same results as what I proposed. You've shown in the example that the result is the same. Because of that I'm OK with it.

brianljt:
Hi @Irving-cl @hubertmis , I did not go though the whole discussion above, Please correct me if I miss anything.

I think the current fields `otRadioFrame.mInfo.mTxInfo.mPeriod/mPhase` in transmit frame should be accurate enough for this transmission purpose. From your discussion above, I think the problem is not `nrf5AlarmGetCurrentTime` itself(it has a fixed software delay to get the current time) but the scheduling delay at OT stack(`SubMac` layer and above).

When `otPlatRadioTransmit` is called, it is some time before the target transmission time(T<sub>target</sub>). The `mPhase` is already aligned to platform's clock(or received frame timestamp).

```
Mac::ProcessCsl()
{
...
child->SetCslPhase(((aFrame.GetTimestamp() + kUsPerTenSymbols - 1) / kUsPerTenSymbols + csl->GetPhase()) %
                   csl->GetPeriod());
...
}
```

So the precise transmission time should be the nearest T where T % `mPeriod` == `mPhase` . so we can set the base time T<sub>base</sub> to the time where T % `mPeriod` == 0 and target delay D<sub>t</sub> to `mPhase` with some simple algo.

Add below check at `otPlatRadioTransmit`.
```
if (aFrame->mInfo.mTxInfo.mPeriod != 0)
{
    uint32_t now = nrf5AlarmGetCurrentTime();
    uint32_t start;
    uint8_t  phase = aFrame->mInfo.mTxInfo.mPhase;

    // Get previous closest time which could be divided by CSL period
    start = now - (now % aFrame->mInfo.mTxInfo.mPeriod);
    if (start + phase <= now)
    {
        // This sample window is already passed, start from next closest time
        start += aFrame->mInfo.mTxInfo.mPeriod;
    }

    // return error if the target transmission delay is bigger than some threshold
    if (start + phase - now >= kDelayThreshold)
    {
        return OT_ERROR_FAILURE;
    }

    return nrf_802154_transmit_raw_at(&aFrame->mPsdu[-1], true, start, phase, channel);
}
```

hubertmis:
@brianljt My point was that apart from delays caused by scheduling of OT stack, `nrf5AlarmGetCurrentTime()` accuracy is not deterministic. Value returned by this function may vary depending on ISRs in the system, scheduling of OS and other potential sources of latency. Because of that I suggested to remove result of `nrf5AlarmGetCurrentTime()` from parameters passed to `nrf_802154_transmit_at()`.

From the discussion it seems that it is feasible and this way we can achieve deterministic precision of radio timings.

brianljt:
> @brianljt My point was that apart from delays caused by scheduling of OT stack, `nrf5AlarmGetCurrentTime()` accuracy is not deterministic. Value returned by this function may vary depending on ISRs in the system, scheduling of OS and other potential sources of latency. Because of that I suggested to remove result of `nrf5AlarmGetCurrentTime()` from parameters passed to `nrf_802154_transmit_at()`.
> 
> From the discussion it seems that it is feasible and this way we can achieve deterministic precision of radio timings.

I see. Thanks for the explanation. So the timestamp in received frame is trust-worthy as it is captured using PPI to be saved in Timer capture register, right?

hubertmis:
@brianljt That's right. With proposed approach it is possible to remove any software delays from calculation. Timestamps of received frames can be captured by hardware (capture registers) and transmission can be initialized by hardware timer (compare registers) as well.

Irving-cl:
I found an issue when I implement it with the new way.

In `nrf_802154_transmit_raw_at`, the base time `t0` is of type `uint32_t`. However, the timestamp of a received frame has type `uint64_t` ([nrf528xx/src/radio.c:939](https://github.com/openthread/openthread/blob/master/examples/platforms/nrf528xx/src/radio.c#L939)). If we pass `RxTimestamp` as `t0` when using this api, we would probably get overflow. `uint32_t` has the max value `4294967295`. `4294967295`us is about `4295`s which is less than `2` hours. This means if one device runs for more than one hour, `nrf_802154_transmit_raw_at`will probably go wrong.

Shall we set `t0` in `nrf_802154_transmit_raw_at` as a `uint64_t`?
@hubertmis @LuDuda

hubertmis:
@Irving-cl , this 32-bit overflow is expected. 

`nrf_802154_transmit_at()` expects timing parameters from now up to approximately 40 minutes in the future (half of 32-bit number in microseconds). Timing range [now ... 40 min later] should be enough for all applications. If you try to schedule an operation further in the future than 40 min from now, it would be considered as past, because of 32-bit number overflow, and denied. If you try to schedule an operation 100 min from now, it would be wrongly scheduled approx 10 min from now, due to 32-bit overflow.

Irving-cl:
@hubertmis Got it. Thanks.

hubertmis:
@Irving-cl , BTW, the timestamp of received frame in `nrf_802154_received_timestamp_raw()` is also `uint32_t time` with 32-bit overflow. This `time` shall be used as `t0`.

Irving-cl:
> BTW, the timestamp of received frame in `nrf_802154_received_timestamp_raw()` is also `uint32_t time` with 32-bit overflow. This `time` shall be used as `t0`.

@hubertmis 

The timestamp of received frame is:
```
receivedFrame->mInfo.mRxInfo.mTimestamp = nrf5AlarmGetCurrentTime() - offset;
```
I'm a little confused here. Do you mean the value of `mTimestamp` is always within the range of `uint32_t` even though `nrf5AlarmGetCurrentTime()` returns `uint64_t` value?

hubertmis:
@Irving-cl , I meant argument `uint32_t time` that is used in `uint32_t offset` calculation. I see it is later used in `uint64_t mTimestamp` calculation. But using 64-bit timestamp captured from alarm module is not necessary here, as we discussed above.
I think 64-bit arithmetic could be avoided as an optimization. But I don't think we should optimize now. Let's make it work precisely first :)

On the other hand, probably 64-bit timestamp from alarm module is needed to schedule next transmission. Because of that we most probably cannot fully avoid it.

Irving-cl:
@hubertmis 

Thanks a lot. I have implemented it using the new way and I'm debugging it. Currently it seems that **`nrf_802154_transmit_raw_at` transmits the frame about 1000us earlier than expected**. And I really cannot tell why.

Details:
* I have two nodes (using nrf52840-dk): a leader and an SSED. I start the leader first and then start the SSED. The first message that contains CSL IE sent by SSED is a `Child Update Request`. And then the leader would reply this message using CSL transmission.
* Part of the capture (using ot sniffer):
Screen Shot 2020-10-20 at 4 09 41 PM

* We can see `Child Update Response` are re-transmitted, which means the child doesn't receive the message successfully. We can see that **the CSL phase is `1918`** in `Child Update Request(No.30)`. So the expected time of `Child Update Response` should be: `326.900543 + (1918 * 160 / 1000000) = 327.207423`. But actually the first `Child Update Response(No.32)` is `327.206420`. (about 1000us earlier)
* I added some log in the leader. And I verified that when calling `nrf_802154_transmit_raw_at`, `t0` is set to the receive timestamp of the `Child Update Request(No.30)` and `dt` is set to `1918 * 160 = 306880` here. 

I really cannot figure out how would this happen. I also tried like this: always adding `1000` to the `dt` when calling `nrf_802154_transmit_raw_at` and it worked. The CSL transmissions are successful then.

Any idea about this? The code I run with is the same as current PR. The code with my logs is at: https://github.com/Irving-cl/openthread/tree/enhance_csl_nordic_test

hubertmis:
@Irving-cl , I did some calculations based on the capture you provided. I assumed you used nRF 802.15.4 sniffer that gives timestamps of received frames with 1us accuracy, captured at the end of last symbol of the frame.

It looks that the frame 30 started its transmission at 326.900543 - 104*0.000032 = 326.897215.
In CSL IE it states that it expects response 1918 CSL units after start of this frame: 326.897215 + 1918*0.000160 = 327.204095

The peer is expected to start its response transmission at 327.204095

Response timestamp (frame 32) is 327.206420. It is the timestamp of the end of the last transmitted symbol. The first transmitted symbol was at 327.206420 - 71*0.000032 = 327.204148

Response is transmitted 53 us later than expected. I think it is within expected precision. However it looks that receiver was not enabled at this time, but according to your description it was enabled 1ms too late.

Irving-cl:
@hubertmis 

I didn't use a nRF 802.15.4 sniffer. I used an `ot-sniffer` which is basically an ot-rcp running on a nrf52840 dongle. It just keeps receiving and forwards the received packets to wireshark. So the timestamp of frames in the capture just come from `receivedFrame->mInfo.mRxInfo.mTimestamp = nrf5AlarmGetCurrentTime() - offset;`. So I think these timestamps in the pcap are **the time of the first symbol received**.

On the other hand, I think `nrf_802154_transmit_raw_at` did send the frame at expected time. (I printed some timestamps on transmitter and think this is true)

Do you have any idea? I noticed the calculation of received timestamp:
```
      uint32_t offset =   int otPlatAlarmMicroGetNow()                                                                                                                                                                                                     
          (int32_t)otPlatAlarmMicroGetNow() - (int32_t)nrf_802154_first_symbol_timestamp_get(time, p_data[0]);                                                                                                                                             
      receivedFrame->mInfo.mRxInfo.mTimestamp = nrf5AlarmGetCurrentTime() - offset;   
```
Is it an approximate conversion or the conversion is absolutely accurate?

hubertmis:
@Irving-cl , I recommend to use nRF 802.15.4 sniffer: https://github.com/NordicSemiconductor/nRF-Sniffer-for-802.15.4 . We created it to solve CSL timing mysteries. It provides accurate timings of known point in received frame.

I've run some calculations for Acks in the posted capture and it seems that the timestamps are taken at the end of received frames, but I it looks that accuracy of timestamps is not perfect. I hope it is good enough for rough calculations.

I don't know details of OT platform and sniffer implementations. I assume that `otPlatAlarmMicroGetNow()` and `nrf5AlarmGetCurrentTime()` return the same value (possibly one is limited to 32 LSB bits). Usually given conversion should be accurate. There is a race condition when RTC tick happens between a call to one function and a call to the other one.
However, it is possible that sniffer implementation reports other timestamp to wireshark than value of `receivedFrame->mInfo.mRxInfo.mTimestamp`. I think that using other timestamp source would explain why we see timestamps of the end of the frame and why accuracy is a little worse than 1 microsecond.

Irving-cl:
@hubertmis 

Thanks very much. I tried the nRF 802.15.4 sniffer:
Screen Shot 2020-10-21 at 9 58 02 AM

So the timestamps captured are the time when last symbol received, right? Then I think the time is correct:

Timestamp of frame No.21: **24.421090**
Timestamp of first symbol of frame No.21: 24.421090 - 130 * 0.000032 = **24.416930**
Expected Timestamp of first symbol of frame No.23: 24.416930 + 456 * 0.000160 = **24.489890**

Timestamp of frame No.23: **24.493009**
Timestamp of first symbol of frame No.23: 24.493009 - 97 * 0.000032 = **24.489905**

The difference is just **15us** and the precision seems good. I'd like to have your confirmation for my calculation above.

hubertmis:
@Irving-cl , your calculation is perfect.

As a clarification: "first symbol of the frame" in this calculation could be first symbol of PSDU, or PHR. I'm not sure if it is actually first symbol of SHR. However, because we are measuring deltas between starts of frames, as long as all "first symbols" are first symbol of PHR for all frames, we are good with these calculations.

To be sure if it is first symbol of SHR, we would need to verify what actually is counted in length field in wireshark.

edmont:
Hi, I've been doing some testing on the current PR status. I'm using this little method to calculate the next expected transmission window timestamps based on CSL parameters and frames sizes:

```
def next_timestamps(period, phase, rx_len, tx_len, windows_nr = 1):
    first_expected = phase * 0.000160 - rx_len * 0.000032
    first_timestamp = first_expected + tx_len * 0.000032
    for i in range(windows_nr):
        print(first_timestamp + i * period * 0.000160)
```
Screenshot from 2020-10-22 13-04-03

For the frame 15 we have:
```
next_timestamps(3125, 2336, 102, 69)
0.37270400000000004
```
So that frame 17 is delayed 255 us.

For the frame 18 we have:
```
next_timestamps(3125, 3197, 18, 48, 3)
0.51248
1.01248
1.51248
```
So that frame 19 is delayed 29 ms.

For the frame 21 we have:
```
next_timestamps(3125, 3180, 56, 48, 13)
0.508544
[...]
6.0085440000000006
6.5085440000000006
```
So that frame 24 is delayed 159 ms.

Are my observations right?

Irving-cl:
@edmont 

I think the calculation is correct. However, the result seems quite different on my side:
Screen Shot 2020-10-22 at 8 54 25 PM

I calculated some results based on your calculation and the precision is always within 1000us. What sniffer did you use? I ran it with 2 nrf52840 DKs (1.1.0 PCA10059). Did you use DKs or dongle? And what options did you enable to build the firmware?

edmont:
@Irving-cl I used dongle as sniffer and DKs as Leader and SSED. Built with:

`make -f examples/Makefile-nrf52840 THREAD_VERSION=1.2 CSL_RECEIVER=1`

Irving-cl:
@edmont 

That's exactlly the same build options with mine. Did you use the nRF sniffer that mentioned above: https://github.com/NordicSemiconductor/nRF-Sniffer-for-802.15.4 ? Could you double check that you build with the latest commit?

And how is the CSL transmission going at your side? Does it miss frequently (no reply)?

edmont:
You are right, I was not using the latest commit. Now it works perfectly for me, and I believe the sample window can be greatly reduced. I was able to deliver 200 echo requests at a random 1-10 seconds interval with just 2 retransmissions with `#define OPENTHREAD_CONFIG_CSL_SAMPLE_WINDOW 2`. It's probably worth reducing this value in order to improve battery life on the SSED side.


[200-pings.pcap.zip](https://github.com/openthread/openthread/files/5423876/200-pings.pcap.zip)

Irving-cl:
@edmont 

Thanks for the great testing. I'm so happy to see that.

 Regarding the `SAMPLE_WINDOW`, it would be great make it as little as possible. However I just want to mention that, for a normal IEEE 802.15.4, say 50 bytes (a default ping request is 48 bytes long), the receiving would last for `50 * 32 = 1600us` (1 octet is 2 symbols,  and we got 160us/10symbols). And that is `10` for `SAMPLE_WINDOW`. Currently we ensure that the receiving won't be interrupted by the sleep by using `nrf_802154_sleep_if_idle()`. So I think when the `SAMPLE_WINDOW` is small enough, the power cost won't decrease further, especially for large frames, because the radio would spend the time to complete receiving. Of course it would still make a difference when receiving small frames. What about setting the `SAMPLE_WINDOW` to `5` here?

examples/platforms/nrf528xx/nrf52840/openthread-core-nrf52840-config.h
```diff
@@ -285,7 +285,34 @@
...
+ */
+#define OPENTHREAD_CONFIG_PLATFORM_DELAY_TRANSMIT_SUPPORTED 1
+
+/**
+ * @def OPENTHREAD_CONFIG_CSL_USE_PLATFORM_DELAY_TRANSMIT
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_CSL_USE_PLATFORM_DELAY_TRANSMIT
+#define OPENTHREAD_CONFIG_CSL_USE_PLATFORM_DELAY_TRANSMIT (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
+#endif
```
brianljt, line 305: Suggest to combine this with `OPENTHREAD_CONFIG_MAC_SOFTWARE_TX_TIMING_ENABLE`. Better to remove this define and use `OPENTHREAD_CONFIG_MAC_SOFTWARE_TX_TIMING_ENABLE` like other radio capabilities, like `OPENTHREAD_CONFIG_MAC_SOFTWARE_RETRANSMIT_ENABLE` and `OPENTHREAD_CONFIG_MAC_SOFTWARE_CSMA_BACKOFF_ENABLE`

examples/platforms/nrf528xx/nrf52840/openthread-core-nrf52840-config.h
```diff
@@ -285,7 +285,34 @@
...
+#define OPENTHREAD_CONFIG_CSL_SAMPLE_WINDOW 10
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_PLATFORM_DELAY_TRANSMIT_SUPPORTED
+ *
+ * The platform radio has the ability to transmit after a specified delay time.
+ *
+ */
+#define OPENTHREAD_CONFIG_PLATFORM_DELAY_TRANSMIT_SUPPORTED 1
```
brianljt, line 297: Suggest to remove this define and let radio return capability `OT_RADIO_CAPS_TRANSMIT_TIMING`

src/core/thread/csl_tx_scheduler.cpp
```diff
@@ -202,8 +201,10 @@ otError CslTxScheduler::HandleFrameRequest(Mac::TxFrame &aFrame)
 
     aFrame.SetChannel(mCslTxChild->GetCslChannel() == 0 ? Get<Mac::Mac>().GetPanChannel()
                                                         : mCslTxChild->GetCslChannel());
-    aFrame.SetTxPhase(mCslTxChild->GetCslPhase());
-    aFrame.SetTxPeriod(mCslTxChild->GetCslPeriod());
+    aFrame.SetTxDelay(mCslTxChild->GetCslTxDelay());
```
brianljt, line 204: Recalculate the delay with `GetNextCslTransmissionDelay`? I think recalculate the delay here will help,

1. Reduce memory usage as `mCslTxDelay` will not be needed?
2. Avoid the corner case when the code reaches here, the current time has just passed the saved `mCslTxDelay` and a new delay should be calculated.

examples/platforms/nrf528xx/nrf52840/openthread-core-nrf52840-config.h
```diff
@@ -285,7 +285,34 @@
...
+
+/**
+ * @def OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD
+ *
+ * For some reasons, CSL receivers wake up a little later than expected. This variable specifies how much time that
+ * CSL receiver would wake up earlier than the expected sample window. The time is in unit of 10 symbols.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD
+#define OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD 3
```
brianljt, line 315: I'm thinking of using unit of 1 symbol or just microsecond instead of 10 symbols for `OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD` and `OPENTHREAD_CONFIG_CSL_SAMPLE_WINDOW`. This way, we can get finer granularity of time control as CSL is time critical feature.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -452,19 +452,35 @@ otError otPlatRadioReceive(otInstance *aInstance, uint8_t aChannel)
...
-    bool result = true;
+    bool    result = true;
+    otError error  = OT_ERROR_NONE;
 
     aFrame->mPsdu[-1] = aFrame->mLength;
 
     nrf_802154_channel_set(aFrame->mChannel);
 
-    if (aFrame->mInfo.mTxInfo.mCsmaCaEnabled)
+#if OPENTHREAD_CONFIG_CSL_USE_PLATFORM_DELAY_TRANSMIT
```
brianljt, line 462: Check OT version >= 1.2 instead here and return correct radio capability according to OT version.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -475,7 +491,10 @@ otError otPlatRadioTransmit(otInstance *aInstance, otRadioFrame *aFrame)
         setPendingEvent(kPendingEventChannelAccessFailure);
     }
 
-    return OT_ERROR_NONE;
+#if OPENTHREAD_CONFIG_CSL_USE_PLATFORM_DELAY_TRANSMIT
```
brianljt, line 494: ditto

src/core/mac/sub_mac.cpp
```diff
@@ -349,26 +350,26 @@ void SubMac::StartCsmaBackoff(void)
...
+            if (otPlatRadioGetNow(&GetInstance()) <
+                mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime + mTransmitFrame.mInfo.mTxInfo.mTxDelay)
             {
-                mTimer.Start((phaseDesired + mTransmitFrame.mInfo.mTxInfo.mPeriod - phaseNow) * kUsPerTenSymbols);
+#if OPENTHREAD_CONFIG_CSL_USE_PLATFORM_DELAY_TRANSMIT
+                BeginTransmit();
+#else
+                mTimer.StartAt(Time(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime),
+                               mTransmitFrame.mInfo.mTxInfo.mTxDelay);
+#endif
```
brianljt, line 367: When radio supports `OT_RADIO_CAPS_TRANSMIT_TIMING`, the direct transmission should falls to else part of `ShouldHandleTransmitTargetTime`.
```suggestion
                mTimer.StartAt(Time(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime),
                               mTransmitFrame.mInfo.mTxInfo.mTxDelay);
```

brianljt:
> @edmont
> 
> Thanks for the great testing. I'm so happy to see that.
> 
> Regarding the `SAMPLE_WINDOW`, it would be great make it as little as possible. However I just want to mention that, for a normal IEEE 802.15.4, say 50 bytes (a default ping request is 48 bytes long), the receiving would last for `50 * 32 = 1600us` (1 octet is 2 symbols, and we got 160us/10symbols). And that is `10` for `SAMPLE_WINDOW`. Currently we ensure that the receiving won't be interrupted by the sleep by using `nrf_802154_sleep_if_idle()`. So I think when the `SAMPLE_WINDOW` is small enough, the power cost won't decrease further, especially for large frames, because the radio would spend the time to complete receiving. Of course it would still make a difference when receiving small frames. What about setting the `SAMPLE_WINDOW` to `5` here?

I think user should set `SAMPLE_WINDOW` as small as possible as SSED could benefits from smaller sample window regarding power consumption. Parent will not send packet to SSED in every sample. For example, parent sends out one packet to SSED every 10 sample. This means the SSED could still have very small sample window for 9 samples and only one sample is made larger because of ongoing receiving.

However, I would still suggest to keep the default sample window to be bigger than test results(5 or 10 here looks good to me) to make it more tolerant as the CSL accuracy TLV is not implemented yet(Related discussion: [SPEC-961](https://threadgroup.atlassian.net/browse/SPEC-961)). User of OT might have customized board with not so accurate crystal than DK. Another possible way is to make `SAMPLE_WINDOW` runtime configurable, like, expose it as a csl cli command so that user can test multiple settings and settle on a final golden value.

edmont:
I've being doing more testing with increased ping sizes and the value 2 still gives great results. I wouldn't make it bigger than 5 for this PR.

Having a runtime configurable sample window seems like a good idea, together with auto-configuring it depending on the parent precision. And I also think having a mechanism for an SSED to select the most accurate parent should be defined in the spec.

However, is the current CSL Accuracy TLV valid as defined today?

> _The accuracy of the clock that is used by the Device for channel sampling timings in units of ppm._

With this definition both implementations for Nordic (pre- and post- this PR) would indicate the same clock accuracy, but the achievable receive window has one order of magnitude difference. This should be something to discuss within the LPWG as well.

brianljt:
@edmont , I've already proposed to include CSL accuracy TLV in parent response and take it into account when choosing best parent. Please take a look and feel free to add comment. https://threadgroup.atlassian.net/browse/SPEC-961

edmont:
Yes, I saw that, thanks. That's why I'm questioning the validity of CSL Accuracy TLV as defined today.

brianljt:
> However, is the current CSL Accuracy TLV valid as defined today?
> 
> > _The accuracy of the clock that is used by the Device for channel sampling timings in units of ppm._
> 
> With this definition both implementations for Nordic (pre- and post- this PR) would indicate the same clock accuracy, but the achievable receive window has one order of magnitude difference. This should be something to discuss within the LPWG as well.

I think the definition should be something that takes into account of crystal accuracy and maximum software delay. What CSL receiver really cares is how big the possible lag is between the target sample window compared to the first symbol of a CSL frame is detected(T<sub>lag</sub> = T<sub>elapse</sub> * PPM + T<sub>sw_delay</sub>). Looks like the current definition does not take into account of the software scheduling delay. So before this PR, the accuracy TLV should be set bigger than the crystal PPM.

hubertmis:
`nrf_802154_receive_at()` function is designed to enable RX window for specified time. If 802.15.4 driver receives at least part of preamble, SFD and PHR during this receive window, it automatically extends this window to make sure that detected frame fits within. Because of that it should be OK to have RX window shorter than 1ms if crystals are accurate enough.

Moreover, using `nrf_802154_sleep_if_idle()` is not required. When you call `nrf_802154_sleep()` during open RX window, this call is ignored (I think it returns an error code to indicate that). 
To abort active RX window you should call `nrf_802154_receive_at_cancel()` followed with `nrf_802154_sleep()`.

src/core/thread/csl_tx_scheduler.cpp
Irving-cl: I think it's necessary to have both `TxDelay` and `TxDelayBaseTime` in a frame because when we finally call `nrf_802154_transmit_raw_at`, we need both parameters. Note that in this updated implementation, the base time for delay transmission is not `now`. I think there would be some problems if we do something like: add one to another and call the api with `0` and the sum.

Irving-cl:
@hubertmis 

Currently it's kind of difficult to use `nrf_802154_receive_at()` in our implementation because of the `Mac` and `SubMac` model. I think we would still use timer to switch between sleep and receiving for now.

hubertmis:
@Irving-cl , thanks for clarification. I wasn't aware of that `nrf_802154_receive_at()` is not used.
It looks that `nrf_802154_receive_at()` could be used as a further power optimization.

src/core/thread/csl_tx_scheduler.cpp
hubertmis: That's correct. Currently `nrf_802154_transmit_raw_at()` requires two parameters. I think it could work correctly if you implement a hack and pass:
`t0 = t0 + dt - 5000`
`dt = 5000`

But if this optimization is not critical for you now, I would recommend to use both parameters and wait until we optimize `nrf_802154_transmit_raw_at()` API. We plan to do that to reduce memory consumption inside the driver itself.

edmont (approved):

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -452,19 +452,35 @@ otError otPlatRadioReceive(otInstance *aInstance, uint8_t aChannel)
...
+    if (aFrame->mInfo.mTxInfo.mTxDelay != 0)
     {
-        nrf_802154_transmit_csma_ca_raw(&aFrame->mPsdu[-1]);
+        if (!nrf_802154_transmit_raw_at(&aFrame->mPsdu[-1], aFrame->mInfo.mTxInfo.mCsmaCaEnabled,
+                                        aFrame->mInfo.mTxInfo.mTxDelayBaseTime, aFrame->mInfo.mTxInfo.mTxDelay,
+                                        aFrame->mChannel))
+        {
+            error = OT_ERROR_INVALID_STATE;
+        }
+        goto exit;
```
LuDuda, line 471: nit: maybe we can introduce `otExitNow` or much less preferably use `otExpect(false);`

src/core/mac/mac_frame.hpp
```diff
@@ -1311,8 +1311,8 @@ class TxFrame : public Frame
     otError GenerateEnhAck(const RxFrame &aFrame, bool aIsFramePending, const uint8_t *aIeData, uint8_t aIeLength);
 
 #if OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
-    void SetTxPhase(uint16_t aPhase) { mInfo.mTxInfo.mPhase = aPhase; }
-    void SetTxPeriod(uint16_t aPeriod) { mInfo.mTxInfo.mPeriod = aPeriod; }
+    void SetTxDelay(uint32_t aTxDelay) { mInfo.mTxInfo.mTxDelay = aTxDelay; }
```
LuDuda, line 1314: nit: can we add proper docstring for those functions?

LuDuda (approved):
This looks great! Thank you for your effort :+1:

examples/platforms/nrf528xx/nrf52840/openthread-core-nrf52840-config.h
Irving-cl: I think it's a good idea. Shall do it in another separate PR? I think that would be easier.

src/core/thread/csl_tx_scheduler.cpp
brianljt: What I meant is the member `mCslTxDelay` in class `ChildInfo` is not needed. It only exist because it avoids recalculation. But it brings the possibility of a corner case as stated above. So I would suggest to remove `mCslTxDelay` in `ChildInfo` and use `GetNextCslTransmissionDelay` to do recalculation here.

examples/platforms/nrf528xx/nrf52840/openthread-core-nrf52840-config.h
brianljt: Sure. Sounds good to me.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -452,19 +452,35 @@ otError otPlatRadioReceive(otInstance *aInstance, uint8_t aChannel)
...
+    bool    result = true;
+    otError error  = OT_ERROR_NONE;
 
     aFrame->mPsdu[-1] = aFrame->mLength;
 
     nrf_802154_channel_set(aFrame->mChannel);
 
-    if (aFrame->mInfo.mTxInfo.mCsmaCaEnabled)
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+    if ((otPlatRadioGetCaps(NULL) & OT_RADIO_CAPS_TRANSMIT_TIMING) != 0 && aFrame->mInfo.mTxInfo.mTxDelay != 0)
```
brianljt, line 463:
```suggestion
    if (aFrame->mInfo.mTxInfo.mTxDelay != 0)
```
I think there are no need to check radio cap in radio layer.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -452,19 +452,35 @@ otError otPlatRadioReceive(otInstance *aInstance, uint8_t aChannel)
...
     {
-        nrf_802154_transmit_csma_ca_raw(&aFrame->mPsdu[-1]);
+        if (!nrf_802154_transmit_raw_at(&aFrame->mPsdu[-1], aFrame->mInfo.mTxInfo.mCsmaCaEnabled,
+                                        aFrame->mInfo.mTxInfo.mTxDelayBaseTime, aFrame->mInfo.mTxInfo.mTxDelay,
+                                        aFrame->mChannel))
+        {
+            error = OT_ERROR_INVALID_STATE;
+        }
+        otEXPECT(false);
```
brianljt, line 471: Why jump to exit here? In this way, a CSL transmission will not call `otPlatRadioTxStarted()` callback.

hubertmis:
> I think the definition should be something that takes into account of crystal accuracy and maximum software delay

I think that transmitter software delay should be taken into account by the transmitter implementation and transmitter shall start TX operation early enough to transmit at specified time. If this delay is nondeterministic (could be from 500ms to 700ms), transmitter shall add this inaccuracy into Accuracy TLV (the transmitter should start TX 600ms earlier, and inform receiver that TX time varies +/- 100 ms).

In similar way, RX software delay should be taken into account by the receiver implementation. If RX start delay is nondeterministic in range 100-300 ms, the receiver should start opening its RX window 300ms earlier than according to the spec.

hubertmis (approved):
Using 802.15.4 API looks correct. I can't comment about other parts of the review, because I don't know how MAC layer works now.

I was thinking how is handled the case when CSL child is desynchronized with its parent. The parent shall store the message that is going to be retrieved by the child either using data poll option, or CSL after resynchronization later. The queue used to retrieve messages both ways seems like something interesting. But it looks it is out of the scope of this review. Is is already implemented?

Irving-cl:
> Using 802.15.4 API looks correct. I can't comment about other parts of the review, because I don't know how MAC layer works now.
> 
> I was thinking how is handled the case when CSL child is desynchronized with its parent. The parent shall store the message that is going to be retrieved by the child either using data poll option, or CSL after resynchronization later. The queue used to retrieve messages both ways seems like something interesting. But it looks it is out of the scope of this review. Is is already implemented?

@hubertmis 

Yeah, the queue is shared between data poll and CSL transmission and there is some logic to handle them. For example, a CSL transmission has already been scheduled but not yet sent, and then receiving a data poll. And at this moment, the message would be directly sent out and the CSL transmission would be aborted.

hubertmis:
> Yeah, the queue is shared between data poll and CSL transmission and there is some logic to handle them. For example, a CSL transmission has already been scheduled but not yet sent, and then receiving a data poll. And at this moment, the message would be directly sent out and the CSL transmission would be aborted.

Thanks for the details. I'll have a look at the code to find it. I'm just curious how it works :)

src/core/mac/sub_mac.hpp
```diff
@@ -547,6 +547,15 @@ class SubMac : public InstanceLocator, private NonCopyable
         kCslSampleWindow = OPENTHREAD_CONFIG_CSL_SAMPLE_WINDOW * kUsPerTenSymbols,
     };
 
+    /**
+     * CSL receivers would wake up `kCslReceiveTimeAhead` earlier than expected sample window. The time is in unit of 10
+     * symbols.
+     *
+     */
+    enum {
+        kCslReceiveTimeAhead = OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD,
```
brianljt, line 556: Put `kCslReceiveTimeAhead` and `kCslSampleWindow` in same enum block?

brianljt (approved):
Thanks for the effort. LGTM. 👍 One nit suggestion.

src/core/mac/sub_mac.hpp
```diff
@@ -540,11 +540,16 @@ class SubMac : public InstanceLocator, private NonCopyable
 
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
     /**
-     * The SSED sample window in units of 10 symbols.
+     *
      *
      */
```
brianljt, line 545: Remove this comment block? Or some comments about this enum block is about CSL parameters?

jwhui (approved):

src/core/mac/sub_mac.cpp
```diff
@@ -439,6 +432,13 @@ void SubMac::BeginTransmit(void)
     }
 
     error = Get<Radio>().Transmit(mTransmitFrame);
+    if (error == OT_ERROR_INVALID_STATE && mTransmitFrame.mInfo.mTxInfo.mTxDelay > 0)
```
bukepo, line 435: @Irving-cl Why we need to give it another try? It looks like this is some platform specific workaround?

PR #5546, opened 2020-09-18
[scripts] output cert test log to stderr

simonlingoogle:
This PR allows thread-cert tests to output OT log (`otLogXxx`) to stderr which is very helpful for debugging.


simonlingoogle:
seems `VERBOSE=1` just do the job. :100:

PR #5548, opened 2020-09-18
[posix] fix options dependency

### LINKED ISSUE #5520: [posix] cannot build posix with default options
When I build posix using autotools:
```
./bootstrap
make -f src/posix/Makefile-posix
```

I got the error:
```
ot_cli-main.o: In function `main':
/home/openthread/src/posix/../../src/posix/main.c:329: undefined reference to `otCliUartInit`
```

The reason is that, in `Makefile-posix:112`, `CLI_TRANSPORT` was evaluated to `CONSOLE`  and the code in `cli_uart.cpp` is not compiled. However in `src/posix/main.c:70`, `OPENTHREAD_USE_CONSOLE` is not evaluated to `1` because neither `HAVE_LIBEDIT` nor `HAVE_LIBREADLINE` is defined.

* First, I think we shouldn't allow users to configure `CLI_TRANSPORT` while we evaluate it again in `Makefile-posix`. It's not clear to me why we add the line`CLI_TRANSPORT                   = $(if $(and $(filter-out no,$(READLINE)),$(filter 0,$(DAEMON))),CONSOLE,UART)`. If we need to verify that `READLINE` and `DAEMON` are properly set when it's `CONSOLE`, we can add such a check and detect at configuring and building. Instead of setting the `CLI_TRANSPORT` value again. I think we should remove this line and add a check instead. Thoughts?
* `HAVE_LIBEDIT` and `HAVE_LIBREADLINE` seem to only exist when we use CMake(`src/posix/cli.cmake:52`). So when using autotools, they would never be evaluated to `1` and hence `OPENTHREAD_USE_CONSOLE` is always 0. So I think we cannot build posix with `CLI_TRANSPORT=CONSOLE` on posix. How should we handle this?
* On github action now, take this [posix-cli](https://github.com/openthread/openthread/pull/5518/checks?check_run_id=1104651093) build as an example, we can see `-DOPENTHREAD_CONFIG_CLI_TRANSPORT=OT_CLI_TRANSPORT_UART` in its config. However in its environment, `READLINE=readline` is set and `DAEMON` would be 0 by default. This should make `CLI_TRANPORT` evaluate to `CONSOLE` in `Makefile-posix:112`. I don't understand. Why it's `UART` in this case?

Irving-cl:
This PR fixes the problem #5520 

The root cause of #5520 turned out to be the absence of LIBREADLINE in the environment. However the error I got is confusing and doesn't indicate that because the options dependencies are not good. I think the options' dependency and relationship here should be:
1. When `DEAMON=0`(ot-cli, ot-ncp will be built and ot-cli depends on console), `CLI_TRANSPORT=CONSOLE` MUST be used. And otherwise (`DEAMON=1`), `CLI_TRANSPORT=UART`MUST be used. **So when users build posix, they don't need to consider the option `CLI_TRANSPORT`.**  It's reasonable to add the evaluation for `CLI_TRANSPORT` in `Makefile-posix`.
2. When `CLI_TRANSPORT=CONSOLE`, `(HAVE_LIBEDIT || HAVE_LIBREADLINE)` MUST be defined and `DAEMON=0` MUST be satisfied. I think `(HAVE_LIBEDIT || HAVE_LIBREADLINE)` is a result of `CLI_TRANSPORT` instead of the reason. The reason for `CLI_TRANSPORT` should be `DEAMON`.

Now I set `CLI_TRANSPORT` option based on `DAEMON` and make `(HAVE_LIBEDIT || HAVE_LIBREADLINE)` as a check for `CLI_TRANSPORT=CONSOLE`.

Going back to issue #5520, if READLINE lib is not found, in this implementation, there will be an error indicating that it's not found.

Resolves #5520

### DIFF
--- a/src/posix/main.c
+++ b/src/posix/main.c
@@ -66,12 +66,8 @@
 #elif OPENTHREAD_POSIX_APP_TYPE == OT_POSIX_APP_TYPE_CLI
 #include <openthread/cli.h>
 
-#include "cli/cli_config.h"
-#if (HAVE_LIBEDIT || HAVE_LIBREADLINE) && !OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE && \
-    (OPENTHREAD_CONFIG_CLI_TRANSPORT == OT_CLI_TRANSPORT_CONSOLE)
-#define OPENTHREAD_USE_CONSOLE 1
 #include "console_cli.h"
-#endif
+#include "cli/cli_config.h"
 #else
 #error "Unknown posix app type!"
 #endif
--- a/src/posix/console_cli.cpp
+++ b/src/posix/console_cli.cpp
@@ -28,6 +28,8 @@
 
 #include "console_cli.h"
 
+#if OPENTHREAD_CONFIG_CLI_TRANSPORT == OT_CLI_TRANSPORT_CONSOLE
+
 #include <openthread/config.h>
 
 #include <assert.h>
@@ -129,3 +131,5 @@ void otxConsoleProcess(const otSysMainloopContext *aMainloop)
 }
 
 #endif // HAVE_LIBEDIT || HAVE_LIBREADLINE
+
+#endif // OPENTHREAD_CONFIG_CLI_TRANSPORT == OT_CLI_TRANSPORT_CONSOLE
--- a/src/posix/console_cli.h
+++ b/src/posix/console_cli.h
@@ -35,6 +35,24 @@
 
 #include <openthread/openthread-system.h>
 
+#include "cli/cli_config.h"
+
+#ifndef HAVE_LIBEDIT
+#define HAVE_LIBEDIT 0
+#endif
+
+#ifndef HAVE_LIBREADLINE
+#define HAVE_LIBREADLINE 0
+#endif
+
+#if OPENTHREAD_CONFIG_CLI_TRANSPORT == OT_CLI_TRANSPORT_CONSOLE
+#define OPENTHREAD_USE_CONSOLE 1
+#if !(HAVE_LIBEDIT || HAVE_LIBREADLINE) || OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE
+#error \
+    "When OPENTHREAD_CONFIG_CLI_TRANSPORT=OT_CLI_TRANSPORT_CONSOLE, HAVE_LIBEDIT or HAVE_LIBREADLINE MUST be defined and OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE MUST be 0"
+#endif
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif


src/posix/main.c
```diff
@@ -67,9 +67,12 @@
 #include <openthread/cli.h>
 
 #include "cli/cli_config.h"
-#if (HAVE_LIBEDIT || HAVE_LIBREADLINE) && !OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE && \
-    (OPENTHREAD_CONFIG_CLI_TRANSPORT == OT_CLI_TRANSPORT_CONSOLE)
+#if OPENTHREAD_CONFIG_CLI_TRANSPORT == OT_CLI_TRANSPORT_CONSOLE
 #define OPENTHREAD_USE_CONSOLE 1
+#if !(HAVE_LIBEDIT || HAVE_LIBREADLINE) || OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE
+#error \
+    "When OPENTHREAD_CONFIG_CLI_TRANSPORT=OT_CLI_TRANSPORT_CONSOLE,  HAVE_LIBEDIT or HAVE_LIBREADLINE MUST be defined and OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE MUST be 0"
```
bukepo, line 74:
```suggestion
    "When OPENTHREAD_CONFIG_CLI_TRANSPORT=OT_CLI_TRANSPORT_CONSOLE, HAVE_LIBEDIT or HAVE_LIBREADLINE MUST be defined and OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE MUST be 0"
```

src/posix/main.c
```diff
@@ -67,9 +67,12 @@
 #include <openthread/cli.h>
 
 #include "cli/cli_config.h"
-#if (HAVE_LIBEDIT || HAVE_LIBREADLINE) && !OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE && \
-    (OPENTHREAD_CONFIG_CLI_TRANSPORT == OT_CLI_TRANSPORT_CONSOLE)
+#if OPENTHREAD_CONFIG_CLI_TRANSPORT == OT_CLI_TRANSPORT_CONSOLE
 #define OPENTHREAD_USE_CONSOLE 1
+#if !(HAVE_LIBEDIT || HAVE_LIBREADLINE) || OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE
```
bukepo, line 72: move this block into console_cli.h?

bukepo (approved):
LGTM:+1:

src/posix/main.c
Irving-cl: SGTM. I also moved the above default definition of `HAVE_LIBEDIT` and `HAVE_LIBREADLINE` into `console_cli.h` to avoid warnings.

jwhui (approved):

PR #5551, opened 2020-09-19
[script] fix output when no arguments are specified

dismirlian:
When no arguments are provided to `script/test`, the script prints `$1: unbound variable` and exits without printing the usage information. This is due to the `set -u` at the top. 

This PR has a small fix so that a newbie like me is presented with the usage information. Also, it prints an error when it detect an unknown command.

Please note that the arguments to the commands themselves are not validated.

Thanks!

### DIFF
--- a/script/test
+++ b/script/test
@@ -471,8 +471,7 @@ main()
                 ;;
             *)
                 echo
-                echo -e "${COLOR_FAIL}Unknown command${COLOR_NONE}: '$1'"
-                exit 1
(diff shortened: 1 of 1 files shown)


dismirlian:
Actually, the "command not found" thing breaks the script because the commands don't consume the arguments they use.

What do you think?

bukepo (approved):
LGTM:+1:

jwhui (approved):
Thanks! 👍

PR #5555, opened 2020-09-21
[message] add helper macros to free message (on error)

abtink:
This commit adds macros `FreeMessage()`, `FreeMessageOnError(`' and
`FreeAndNullMessageOnError()` which would free a given `Message`
pointer if not `nullptr` (conditionally on an `otError`).

These macros implement small yet common code patterns used in many of
the core modules. They are intentionally defined as macros instead of
inline methods/functions to ensure that they are fully inlined. Note
that an `inline` method/function is not necessarily always inlined by
the toolchain, and not inlining such small implementations could add a
rather large code-size overhead.

### DIFF
--- a/src/core/common/message.hpp
+++ b/src/core/common/message.hpp
@@ -53,8 +53,6 @@
 
 namespace ot {
 
-class ThreadLinkInfo;
-
 /**
  * @addtogroup core-message
  *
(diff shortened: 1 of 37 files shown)


src/core/common/message.hpp
```diff
@@ -65,6 +63,61 @@ class ThreadLinkInfo;
...
+ * @param[in] aMessage    A pointer to a `Message` to free (can be nullptr).
+ * @param[in] aError      The `otError` to check.
+ *
+ */
+#define FreeMessageOnError(aMessage, aError)                        \
+    do                                                              \
+    {                                                               \
+        if (((aError) != OT_ERROR_NONE) && ((aMessage) != nullptr)) \
+        {                                                           \
+            (aMessage)->Free();                                     \
```
simonlingoogle, line 100:
```suggestion
            (aMessage)->Free();                                     \
            (aMessage) = nullptr;                                    \
```

If this does not increase code size, maybe we can avoid defining `FreeAndNullMessageOnError` ?

src/core/common/message.hpp
```diff
@@ -65,6 +63,61 @@ class ThreadLinkInfo;
...
+/**
+ * This macro frees a given message buffer if a given `otError` indicates an error.
+ *
+ * The parameter @p aMessage can be nullptr in which case this macro does nothing.
+ *
+ * @param[in] aMessage    A pointer to a `Message` to free (can be nullptr).
+ * @param[in] aError      The `otError` to check.
+ *
+ */
+#define FreeMessageOnError(aMessage, aError)                        \
```
simonlingoogle, line 95: What about using static inline function for better type checking. 

```suggestion
static inline void FreeMessageOnError(const Message *&aMessage, otError aError) 
```

simonlingoogle:
@abtink 
Any clue on why code size increased by 16B?

src/core/common/message.hpp
abtink: We cannot use it. It does impact the code size.

src/core/common/message.hpp
abtink: I used `static inline` first. It can impact code-size (majorly). `inline` is recommendation to the complier. We cannot have such simple method not be inlined (the overhead of calling a function is more than the actual assembly code of the implementation).

Please see here (repeated the same in commit/PR message).
```
/**
 * This macro frees a given message buffer if not nullptr.
 *
 * This macro and the ones that follow contain small but common code patterns used in many of the core modules. They
 * are intentionally defined as macros instead of inline methods/functions to ensure that they are fully inlined.
 * Note that an `inline` method/function is not necessarily always inlined by the toolchain and not inlining such
 * small implementations can add a rather large code-size overhead.
 *
 * @param[in] aMessage    A pointer to a `Message` to free (can be nullptr).
 *
 */```

Comment (simonlingoogle) [reply src/core/common/message.hpp:95]: Thanks for the info.

File: src/core/common/message.hpp (hunk 1 above)
Comment (simonlingoogle) [reply src/core/common/message.hpp:100]: I tested it with nrf52840 build and it didn't increase code size. But not sure if all compilers can do such optimization (dead code elimination).

Comment (abtink) [reply src/core/common/message.hpp:100]: I know some toolchain could optimize the unused assignment and remove the code but we would then need to rely on compiler/toolchain behavior/optimization. 
My recommendation/approach is not to rely on toolchain behavior. Better be explicit and clear when/if we can.

Comment (simonlingoogle) [reply src/core/common/message.hpp:100]: SGTM

**Review by simonlingoogle — APPROVED** [review]
LGTM

**Comment by abtink** [comment]
> @abtink Any clue on why code size increased by 16B?

I was curious about this as well. I verified it and it seems to due to the change/fix in this PR in  `src/core/net/dhcp6_client.cpp`.
```
-    if (message != nullptr)
     {
+        FreeMessage(message);
         otLogWarnIp6("Failed to send DHCPv6 Solicit: %s", otThreadErrorToString(error));
-
-        if (error != OT_ERROR_NONE)
-        {
-            message->Free();
-        }
     }
 }
``` 
The existing code does not seem to be correct as we would log "Failed to send ..." on a non-null `message` with/without an error.

The change/diff is very small (practically switching the two `if` checks to first check `error` and then `message`).  I verified that this change is adding the 16 bytes to code size. I briefly looked at the disassembled generated code for this method and  it does indeed look rather different (towards the end of the method). Not clear to me why arm-gcc generated so different of a assembly code due to this rather small change. I guess it may be tied to certain optimization that the arg-gcc tried to do

Anyways, 16 bytes really does not matter and the original/existing code does not seem to be correct.

jwhui (approved):

PR #5563, opened 2020-09-22
[cli] add command to set/get external FEM's LNA gain

### LINKED ISSUE #5561: Support for FEMs
Currently, OT provides two API functions (and associated CLI commands) to set the TX power and the CCA threshold. As per IEEE802.15.4, those values should be measured at the antenna (or antenna connector). When a FEM is used in front of the radio, the user must correct for the PA and LNA gains by hand. Please see #5532 for background.

As per the discussion in #5532, an API to set the PA and LNA gains, so that the radio platform driver can account for the FEM, (and maybe also cap the total TX power).

Please see #5532.

brianljt:
Per discussion here https://github.com/openthread/openthread/pull/5532#issuecomment-693775116, 

> There may be different CCA threshold requirements in different regulatory domain, setting a fixed LNA value allows OpenThread to automatically adjust the threshold value according to the regulatory domain settings.

This PR adds a cli command to set/get the external FEM's LNA gain

Resolves #5561


src/cli/README.md
```diff
@@ -807,6 +808,27 @@ Delete all stored settings, and signal a platform reset.
 > factoryreset
 ```
 
+### femparam
```
bukepo, line 811:
```suggestion
### fem lnagain
```

src/cli/README.md
```diff
@@ -807,6 +808,27 @@ Delete all stored settings, and signal a platform reset.
...
+
+Get the external FEM parameters.
+
+```bash
+> femparam
+LNA gain 11 dBm
+Done
+```
+
+### femparam \<LNA gain\>
```
bukepo, line 821:
```suggestion
### fem lnagain \<LNA gain\>
```

include/openthread/platform/radio.h
```diff
@@ -456,6 +456,31 @@ otError otPlatRadioGetCcaEnergyDetectThreshold(otInstance *aInstance, int8_t *aT
  */
 otError otPlatRadioSetCcaEnergyDetectThreshold(otInstance *aInstance, int8_t aThreshold);
 
+/**
+ * Get the external FEM's Rx LNA gain in dBm.
+ *
+ * @param[in] aInstance  The OpenThread instance structure.
```
bukepo, line 462:
```suggestion
 * @param[in]  aInstance  The OpenThread instance structure.
```

src/cli/README.md
```diff
@@ -807,6 +808,27 @@ Delete all stored settings, and signal a platform reset.
...
 ```
 
+### femparam
+
+Get the external FEM parameters.
+
+```bash
+> femparam
+LNA gain 11 dBm
+Done
```
bukepo, line 818:
```suggestion
> fem lnagain
11
Done
```

bukepo: I know there are examples in CLI that input is a number without unit, and the output is a number with unit. 
I prefer input and output to be the same for consistency. Thoughts? @jwhui @abtink @brianljt

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -701,22 +702,54 @@ otError otPlatRadioSetCcaEnergyDetectThreshold(otInstance *aInstance, int8_t aTh
...
+
+otError otPlatRadioSetFemLnaGain(otInstance *aInstance, int8_t aGain)
+{
+    OT_UNUSED_VARIABLE(aInstance);
+
+    int8_t threshold;
+
+    otEXPECT(otPlatRadioGetCcaEnergyDetectThreshold(aInstance, &threshold) == OT_ERROR_NONE);
+    otEXPECT(otPlatRadioSetCcaEnergyDetectThreshold(aInstance, threshold) == OT_ERROR_NONE);
+    sLnaGain = aGain;
```
Irving-cl, line 747: It's unclear to me here. Should `sLnaGain = aGain` come before `otEXPECT(otPlatRadioSetCcaEnergyDetectThreshold(aInstance, threshold) == OT_ERROR_NONE);` so that the set method could check the new `sLnaGain`?

brianljt: You are right. Made a mistake here.

Irving-cl (approved):
LGTM👍

examples/platforms/simulation/radio.c
```diff
@@ -983,6 +984,28 @@ otError otPlatRadioSetCcaEnergyDetectThreshold(otInstance *aInstance, int8_t aTh
     return OT_ERROR_NONE;
 }
 
+otError otPlatRadioGetFemLnaGain(otInstance *aInstance, int8_t *aGain)
+{
+    OT_UNUSED_VARIABLE(aInstance);
+
+    assert(aInstance != NULL);
```
zhanglongxia, line 991: assert(aGain != NULL);

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -701,22 +702,57 @@ otError otPlatRadioSetCcaEnergyDetectThreshold(otInstance *aInstance, int8_t aTh
...
+
+otError otPlatRadioSetFemLnaGain(otInstance *aInstance, int8_t aGain)
+{
+    OT_UNUSED_VARIABLE(aInstance);
+
+    int8_t  threshold;
+    otError error = OT_ERROR_NONE;
+
+    error = otPlatRadioGetCcaEnergyDetectThreshold(aInstance, &threshold);
+    otEXPECT(error == OT_ERROR_NONE && (threshold + aGain) < NRF528XX_MIN_CCA_ED_THRESHOLD);
```
zhanglongxia, line 747: According to condition of the line 705,  `(threshold + aGain) < NRF528XX_MIN_CCA_ED_THRESHOLD` means the `aGain` is invalid. Should we return an error here?

src/cli/README.md
brianljt: This is inline with ccathreshold and txpower. I think the unit related to power is dBm for most of the cases so there are no need to add unit to the input number. Also, the unit is clarified in the README.md file for cli.

zhanglongxia (approved):
LGTM

dismirlian:
Looks good. I would only point out that most FEMs have a PA in addition to the LNA. I think that an OT user would expect to be able to set the PA gain using a similar command, e.g. `fem pagain`.

brianljt:
> Looks good. I would only point out that most FEMs have a PA in addition to the LNA. I think that an OT user would expect to be able to set the PA gain using a similar command, e.g. `fem pagain`.

Yes. But unlike LNA, most of the time, the PA gain is not a fixed value. It could be adjusted by different Vcc voltage or configured via register setting. So I did not add a pagain setting for now.

dismirlian:
> > Looks good. I would only point out that most FEMs have a PA in addition to the LNA. I think that an OT user would expect to be able to set the PA gain using a similar command, e.g. `fem pagain`.
> 
> Yes. But unlike LNA, most of the time, the PA gain is not a fixed value. It could be adjusted by different Vcc voltage or configured via register setting. So I did not add a pagain setting for now.

I'm fine with this, it's just that the text "_Set the radio's transmit power in dBm measured **at antenna connector**..."_ could be misleading.

Thanks!

brianljt:
> > > Looks good. I would only point out that most FEMs have a PA in addition to the LNA. I think that an OT user would expect to be able to set the PA gain using a similar command, e.g. `fem pagain`.
> > 
> > 
> > Yes. But unlike LNA, most of the time, the PA gain is not a fixed value. It could be adjusted by different Vcc voltage or configured via register setting. So I did not add a pagain setting for now.
> 
> I'm fine with this, it's just that the text "_Set the radio's transmit power in dBm measured **at antenna connector**..."_ could be misleading.
> 
> Thanks!

Thanks for reviewing. I will remove the line for `txpower` before `pagain` setting is supported. I will keep the line for `ccathreshold` since `lnagain` setting is supported in the PR.

src/cli/README.md
```diff
@@ -807,6 +808,25 @@ Delete all stored settings, and signal a platform reset.
...
 
+### fem
+
+Get the external FEM parameters.
+
+```bash
+> fem
+LNA gain 11 dBm
+Done
+```
```
bukepo, line 819:
```suggestion
### fem lnagain

Get LNA gain in dBm of the external FEM.

```bash
> fem lnagain
11
Done
```
```
```suggestion
### fem

Get the external FEM parameters.

```bash
> fem
LNA gain 11 dBm
Done
```
```

File: src/cli/README.md [hunk 9]
```diff
@@ -807,6 +808,25 @@ Delete all stored settings, and signal a platform reset.
...
+
+### fem lnagain \<LNA gain\>
+
+Set the external FEM parameters.
```
bukepo, line 823:
```suggestion
Set the LNA gain in dBm of the external FEM.
```

src/cli/README.md
brianljt: For this, I would like to use `fem` command to retrieve all the fem parameters so no subcommand is needed. How do you think?

bukepo: Is it more consistent to have getter and setter the same pattern? I treat the CLI another protocol to control OpenThread. As you mentioned, the unit is already documented in API and README of the CLI. `LNA gain 11 dBm` uses more bytes and is harder to parse than `11` for scripts.

bukepo (approved):
LGTM:+1:

src/cli/cli.cpp
```diff
@@ -1587,6 +1587,43 @@ otError Interpreter::ProcessFake(uint8_t aArgsLength, char *aArgs[])
 }
 #endif
 
+otError Interpreter::ProcessFem(uint8_t aArgsLength, char *aArgs[])
+{
+    otError error = OT_ERROR_NONE;
+
+    if (aArgsLength == 0)
+    {
+        int8_t lna_gain;
```
LuDuda, line 1596:
```suggestion
        int8_t lnaGain;
```

LuDuda (commented):
@brianljt thank you for this.

Shouldn't we reference to power on the antenna connector for each functionality, and so add/decrement the offset (LNA Gain) in the following functions:

`nrf_802154_energy_detected`, `nrf_802154_received_timestamp_raw`, `nrf_802154_transmitted_raw`, `otPlatRadioGetRssi`?

For example:
```
void nrf_802154_energy_detected(uint8_t result)
{
    sEnergyDetected = nrf_802154_dbm_from_energy_level_calculate(result) - sLnaGain;

    setPendingEvent(kPendingEventEnergyDetected);
}
```

include/openthread/platform/radio.h
```diff
@@ -456,6 +456,31 @@ otError otPlatRadioGetCcaEnergyDetectThreshold(otInstance *aInstance, int8_t *aT
...
+ * Set the external FEM's Rx LNA gain in dBm.
+ *
+ * @param[in] aInstance  The OpenThread instance structure.
+ * @param[in] aGain      The external FEM's Rx LNA gain in dBm.
+ *
+ * @retval OT_ERROR_NONE             Successfully set the external FEM's LNA gain.
+ * @retval OT_ERROR_NOT_IMPLEMENTED  External FEM's LNA gain setting is not implemented.
+ *
+ */
+otError otPlatRadioSetFemLnaGain(otInstance *aInstance, int8_t aGain);
```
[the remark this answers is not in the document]
brianljt: Thank you @LuDuda for the review. PR updated.

We are planning to use this `fem` cli command to set/get several FEM related parameter later. For this, I would keep current name for simplicity. User can count any board/antenna gain/lost to this parameters.

dismirlian:
@brianljt and @LuDuda, regarding the RSSI measurement, I think the driver should not substract the LNA gain (or the offset). The IEEE802.15.4-2015 standard states (Table 8-77):

>  The Received Signal Strength Indicator is a measure of the RF power level at the input of the transceiver measured during the PHR and is valid after the SFD is detected.

I think this makes sense, because the received signal's power is effectively higher due to the LNA, and this is probably what a user expects to see.

Also (7.5.21):

> ... and the use of the Average RSSI result by the next higher layer is not specified in this standard.

I don't know how OT uses the RSSI value, and there may be arguments in favor of taking the RSSI at the antenna/connector.

brianljt:
> @brianljt and @LuDuda, regarding the RSSI measurement, I think the driver should not substract the LNA gain (or the "rssi offset"). The IEEE802.15.4-2015 standard states (Table 8-77):
> 
> > The Received Signal Strength Indicator is a measure of the RF power level at the input of the transceiver measured during the PHR and is valid after the SFD is detected.
> 
> I think this makes sense, because the received signal's power is effectively higher due to the LNA, and this is probably what a user expects to see.
> 
> Also (7.5.21):
> 
> > ... and the use of the Average RSSI result by the next higher layer is not specified in this standard.
> 
> I don't know how OT uses the RSSI value, and there may be arguments in favor of taking the RSSI at the antenna/connector.

Thank you @dismirlian for pointing this out. I checked the code, looks like the RSSI/Energy detection is used to,

1. Calculate the link margin (RSSI - noise floor). For this, as long as the RSSI and noise floor are measured at the same point. Then it should be fine.
2. Use Energy detection to select a channel with relatively lower interference. This should be fine to get the channel energy with or without offsetting the FEM LNA gain.
3. Show the PAN signal quality with cli command `scan`. In this case, if we offset the RSSI with the LNA gain, then user might see a lower signal strength.

As the spec has indicated the RSSI measurement to be at the transceiver pin, I would guess the spec would like the RSSI to include any gain/loss introduces by the board/FEM/antenna. So for above case 3 and what the spec has indicated, I would prefer to NOT offset RSSI/energy detection with this LNA gain but only use it for CCA threshold setting. Thoughts? @dismirlian @LuDuda @Irving-cl

dismirlian:
> 1. Calculate the link margin (RSSI - noise floor). For this, as long as the RSSI and noise floor are measured at the same point. Then it should be fine.

If the LNA's gain is high enough, the noise figure will be dominated by it (Friis' formula), so the RX sensitivity should be enhanced (in general, the performance of the external LNA should be better than the internal's). So, to calculate the link margin I think we should use the RSSI at the receiver and the modified receiver sensitivity (for which we don't have a correction parameter yet, because it would depend on the NF of the LNAs).

> 2. Use Energy detection to select a channel with relatively lower interference. This should be fine to get the channel energy with or without offsetting the FEM LNA gain.

I agree.

> 3. Show the PAN signal quality with cli command `scan`. In this case, if we offset the RSSI with the LNA gain, then user might see a lower signal strength.

I would prefer to show the real RSSI (without offsetting) in this case.

Thanks!

LuDuda:
It's okay for me as well.

brianljt:
Thank you @dismirlian @LuDuda for helping with the review. I've updated the PR as discussed(do not offset RSSI and energy detection result with LNA gain). Please help review again. Thank you!

include/openthread/platform/radio.h
```diff
@@ -717,7 +742,7 @@ extern void otPlatRadioTxDone(otInstance *aInstance, otRadioFrame *aFrame, otRad
 extern void otPlatDiagRadioTransmitDone(otInstance *aInstance, otRadioFrame *aFrame, otError aError);
 
 /**
- * Get the most recent RSSI measurement.
+ * Get the most recent RSSI measurement measured at antenna connector.
```
dismirlian, line 745: Shouldn't this be: "measured at the receiver"?

include/openthread/platform/radio.h
```diff
@@ -747,7 +772,7 @@ otError otPlatRadioEnergyScan(otInstance *aInstance, uint8_t aScanChannel, uint1
  * This function is used when radio provides OT_RADIO_CAPS_ENERGY_SCAN capability.
  *
  * @param[in]  aInstance           The OpenThread instance structure.
- * @param[in]  aEnergyScanMaxRssi  The maximum RSSI encountered on the scanned channel.
+ * @param[in]  aEnergyScanMaxRssi  The maximum RSSI encountered on the scanned channel measured at antenna connector.
```
dismirlian, line 775: Shouldn't this be: "measured at the receiver"?

dismirlian (commented):
@brianljt, sorry for the delay. Looks good, apart from those two comments.

Thanks!

brianljt:
Looks like there are no more comments, I think we can merge this PR. @jwhui please help take a look. Thank you!

src/cli/README.md
```diff
@@ -868,6 +869,35 @@ Note: Only for certification test.
...
+### fem
+
+Get external FEM parameters.
+
+```bash
+> fem lnagain
+LNA gain 11 dBm
+Done
+```
+
```
jwhui, line 881: Is there a use case for this? Are we planning to have other FEM parameters here?

If we want to keep this, is this covered by the expect tests?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1168,6 +1168,15 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::GetCcaEnergyDetectThresh
     return error;
 }
 
+template <typename InterfaceType, typename ProcessContextType>
+otError RadioSpinel<InterfaceType, ProcessContextType>::GetFemLnaGain(int8_t &aGain)
+{
+    otError error = Get(SPINEL_PROP_PHY_FEM_LNA_GAIN, SPINEL_DATATYPE_INT8_S, &aGain);
+
+    LogIfFail("Get CCA ED threshold failed", error);
```
jwhui, line 1176: Is this the intended log message?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1252,6 +1261,14 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::SetCcaEnergyDetectThresh
     return error;
 }
 
+template <typename InterfaceType, typename ProcessContextType>
+otError RadioSpinel<InterfaceType, ProcessContextType>::SetFemLnaGain(int8_t aGain)
+{
+    otError error = Set(SPINEL_PROP_PHY_FEM_LNA_GAIN, SPINEL_DATATYPE_INT8_S, aGain);
+    LogIfFail("Set CCA ED threshold failed", error);
```
jwhui, line 1268: Is this the intended log message?

abtink (approved):
LGTM. Thanks.

src/cli/README.md
brianljt: Per #5561, there are needs to set FEM gain to radio so that when setting/getting CCA threshold, user does not need to offset the LNA gain anymore. There might be other FEM parameters added later, like PA gain to cap the tx power.

This cli is covered in expect tests in cli-misc.exp

jwhui: I see `fem lnagain` is covered, but not `fem`.

Note that the doc above seems incorrect for the example output. It should be `fem` with zero args, correct?

brianljt: Sorry, my mistake. Updated.

LuDuda (approved):

jwhui (approved):

PR #5564, opened 2020-09-22
[gn] add missing files to BUILD.gn

brianljt:


examples/apps/cli/main.c
```diff
@@ -27,6 +27,7 @@
  */
 
 #include <assert.h>
+#include <cli/cli_config.h>
```
Irving-cl, line 30: Not sure if we should put this header here. After checking our style guide again:

> Preprocessor #include directives should be grouped, ordered, or sorted as follows:
> *  If the unit is a core/private header file, "openthread-core-config.h" should be the first header file included.
> *  If the unit is a core/private .c or .cpp file:
>     *  If the unit has a corresponding header file, the unit's corresponding header file should be included before any other header file.
>     *  If the unit has no corresponding header file, then it should directly include "openthread-core-config.h" before any other header file.
> *  C++ Standard Library headers
> *  C Standard Library headers
> *  Third-party library headers
> *  First-party library headers
> *  Private or local headers
> *  Alphanumeric order within each subgroup

First I think `cli/main.c` is not a core/private source. And maybe we should regard `cli/cli_config.h`as **Private or local header**. So the original order looks good to me. Still, I'm not 100% sure. No strong preference on this.

abtink (approved):
LGTM

Irving-cl (approved):
LGTM👍 Thanks.

examples/apps/cli/main.c
brianljt: Thanks. Resolved.

bukepo (approved):
LGTM:+1:

jwhui (approved):

Issue #5568, opened 2020-09-24
[nRF52840] device easily hangs for 1.2 build when continuous traffic since #5184


librasungirl:
**Describe the bug** 
When testing continuous Multicast traffic (MA with scope = 5), nRF52840 DK with 1.2 firmware easily hang
```
(gdb) bt
#0  0x0003f958 in mutex_lock ()
#1  0x00031508 in ProcessAesDrv ()
#2  0x00030ebc in cc310_mbedtls_internal_aes_encrypt ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

J-Link RTT Viewer logs
```
00> [0000050596] [DEBG]-M[0000050603] [DEBG]-MAC-----: Request to start operation "TransmitDataDirect"
00> [0000050603] [DEBG]-MAC-----: Starting operation "TransmitDataDirect"
00> [0000050603] [DEBG]-MAC-----: RadioState: Receive -> CsmaBackoff
00> [0000050603] [DEBG]-MAC-----: RadioState: CsmaBackoff -> Transmit
00> [0000050604] [DEBG]-MAC-----: Received frame from short address 0x1400
00> [0000050604] [DEBG]-MAC-----: Rx security - frame counter 41128
```

**To Reproduce** Information to reproduce the behavior, including
1. Git commit id
`master branch since #5184`

2. IEEE 802.15.4 hardware platform
`nRF52840 DK`

3. Build steps
`make -f examples/Makefile-nrf52840 USB=1 THREAD_VERSION=1.2 DUA=1 MLR=1 FULL_LOGS=1 LOG_OUTPUT=PLATFORM_DEFINED`

**Additional context** 
The issue occurs since #5184

brianljt:
Thanks for pointing this out @librasungirl . From backtrace, looks like it is caused by duplicated access to cc310. When a frame is received, it will use CC310 to do decryption. However, in #5184, the enh-ACK will be generated in interrupt context and will also use CC310 to do encryption. A possible solution is to use software AES instead of CC310 if the mutex cannot be acquired.

simonlingoogle:
> A possible solution is to use software AES instead of CC310 if the mutex cannot be acquired.

Is it possible not to do decryption in interrupt handler but post it to the main loop?

brianljt:
> > A possible solution is to use software AES instead of CC310 if the mutex cannot be acquired.
> 
> Is it possible not to do decryption in interrupt handler but post it to the main loop?

[...]

PR #2719, opened 2018-05-17
[nrf52840] support aes usage in interrupt context

chshu:
These are some use cases to do AES in interrupt context in some new features (for example: #2618 network-wide time sync service), it works well with software AES provided by mbedtls, but there are some problems when hardware acceleration is enabled: 
- Some platform's hardware AES has only one AES engine, so it doesn't support concurrent access;
- Some platform's hardware AES doesn't support interrupt context (for example, nRF52840).

This PR resolves the issue by introducing a simple software AES under platform layer:
* Enhance AesCcm to indicate whether it is used in interrupt context
* Introduce a simple software AES under nRF52840 platform layer for AES usage in interrupt context

PR #5569, opened 2020-09-24
[security] disable CC310 and use buildin mbedtls for nrf52840 in OT1.2

brianljt:
This PR fixes issue #5568.

In OT1.2, the Enh-ACK and Tx frame encryption will be done in interrupt context. However, the decryption is still done in `Mac` layer in normal context. CC310 crypto cell library does not support multi-threading. Change to use software builtin mbedtls library.

### DIFF
--- a/examples/Makefile-nrf52840
+++ b/examples/Makefile-nrf52840
@@ -43,7 +43,14 @@ GCCVersion                      = $(shell expr `$(CC) -dumpversion | cut -f1 -d.
 
 # Disable built-in mbedTLS by default allowing for nrf_security implementation with hardware acceleration.
 # When set to 0 built-in, software implementation would be used.
+ifeq ($(THREAD_VERSION), 1.2)
+ifeq ($(DISABLE_BUILTIN_MBEDTLS), 1)
+$(error for THREAD_VERSION>=1.2, cannot use CC310, please use DISABLE_BUILTIN_MBEDTLS=0)
+endif
+DISABLE_BUILTIN_MBEDTLS        = 0
+else
 DISABLE_BUILTIN_MBEDTLS        ?= 1
+endif
 
 ifeq ($(DISABLE_CC310), 1)
 $(error DISABLE_CC310=1 is deprecated, please use DISABLE_BUILTIN_MBEDTLS=0)


examples/Makefile-nrf52840
```diff
@@ -43,7 +43,14 @@ GCCVersion                      = $(shell expr `$(CC) -dumpversion | cut -f1 -d.
 
 # Disable built-in mbedTLS by default allowing for nrf_security implementation with hardware acceleration.
 # When set to 0 built-in, software implementation would be used.
+ifeq ($(THREAD_VERSION), 1.2)
+ifeq ($(DISABLE_BUILTIN_MBEDTLS), 1)
+$(error for THREAD_VERSION>=1.2, cannot use CC310, please use DISABLE_BUILTIN_MBEDTLS=0)
+endif
+DISABLE_BUILTIN_MBEDTLS        = 0
+else
 DISABLE_BUILTIN_MBEDTLS        ?= 1
```
simonlingoogle, line 52: Being unfamiliar with related aspects, I have some general questions:
- How are CC310 and builtin mbedtls different in performance? 
- Can we solve this issue by prohibiting ISRs when the main loop is using CC130

brianljt: > How are CC310 and builtin mbedtls different in performance?

In my test, the software approach is fast enough to encrypt the biggest packet(127 bytes) on time. However, according to @bukepo , the software approach is slow at calculating the eclipse curves. As only AES has the possibility of being called in interrupt context, I'm planning to only use builtin mbedtls for AES but leave all the other security calculation to CC310.

> Can we solve this issue by prohibiting ISRs when the main loop is using CC130

Unfortunately not. It is because the delayed handling of ACK will cause encryption problem to out going frame. The current sequence is like,
1. Transmit frame (or transmit ACK)
2. Frame header has been set, interrupt generated
3. Do encryption in this interrupt

brianljt:
PR #2719 has fixed this before. But cannot do so as the encryption library on Nordic is a binary lib rather than source code now.

simonlingoogle (approved):
This PR reduced 1.2 code size by 65KB. LGTM :+1:

librasungirl (approved):

Irving-cl (approved):
LGTM👍

LuDuda (approved):
Hello @brianljt,

Thank you for this PR. I would say, we should merge it as a first step. The rest steps are on us.

We do plan to replace the current `nrf_security` library with the newest one, which won't use CC310 for AES but either vanilla mbedtls or nRF Oberon backend. Possibly in the next iteration also detect the IRQ context as in #2719 to utilize CC310 anyway.

Also, we do plan to have AES encryption feature embedded to our radio driver which should also solve the problem.

brianljt:
> Hello @brianljt,
> 
> Thank you for this PR. I would say, we should merge it as a first step. The rest steps are on us.
> 
> We do plan to replace the current `nrf_security` library with the newest one, which won't use CC310 for AES but either vanilla mbedtls or nRF Oberon backend. Possibly in the next iteration also detect the IRQ context as in #2719 to utilize CC310 anyway.
> 
> Also, we do plan to have AES ACK encryption feature embedded to our radio driver which should also solve the problem.

Awesome! Thanks for the updates. Please keep us updated once the `nrf_security` library got updated. The current eclipse curve calculation took too long so that the commissioning process is not user friendly if OT1.2 is enabled. Maybe we should still keep eclipse curve calculation which is only called in normal context(no multi-threading) to crypto cell in the new library? @bukepo please take a look and see if we can merge this as a first step?

brianljt:
Mark it as work in progress. Trying to use CC310 for eclipse curve calculation.

examples/Makefile-nrf52840
simonlingoogle: Thanks for explaination.

LuDuda:
Hello @brianljt 

Yes, this is the plan, to use CC310 for each crypto operation except AES. We should be able to provide a new `nrf_security` library within the next week.

brianljt:
> Hello @brianljt
> 
> Yes, this is the plan, to use CC310 for each crypto operation except AES. We should be able to provide a new `nrf_security` library within the next week.

Thank you @LuDuda for the quick fix.👍  I think we can stop working on this work around PR and wait for your enhancements in `nrf_security` for issue #5568

piotrkoziar:
> Thank you @LuDuda for the quick fix.+1 I think we can stop working on this work around PR and wait for your enhancements in `nrf_security` for issue #5568

Hello,

https://github.com/openthread/openthread/pull/5596 with updated nrf_security is introduced. We will verify it solves the problem soon.

jwhui:
Superseded by #5596

PR #5579, opened 2020-09-26
[github-actions] run packet verification multiple times and allow it to fail

simonlingoogle:
I suggested to run packet verification tests for multiple times because it's easy to write PV code that fails by chance.
But this seems to bring difficulties on other PRs to pass checks.

By this PR, I am trying to find a balance between catching packet verification bugs and passing checks. The solution is:

- Run PV test once without `continue-on-error: true`, which will fail the whole job if fails. 
- Run PV test more times with `continue-on-error: true`, which won't fail the whole job if fails. 
  - Remove *.pcap and *.json if success, keep otherwise. 
- Collect failed *.pcap and *.json to artifacts

The benefits of the solution is two fold:

- PRs will be easier to pass PV checks
- We can still run PV tests for multiple times to catch the bugs and upload the failed artifacts

An example in which `Run More Times` failed (got error annotations and pcaps of failed tests):
image

Thoughts? @FiveDimensions @Irving-cl @librasungirl @jwhui

### DIFF
--- a/.github/workflows/simulation-1.2.yml
+++ b/.github/workflows/simulation-1.2.yml
@@ -122,16 +122,30 @@ jobs:
       - name: Get Thread-Wireshark
         run: |
           ./script/test get_thread_wireshark
-      - name: Run
+      - name: Run Once
         run: |
-          for i in {1..10}
-          do
-            ./script/test cert_suite ./tests/scripts/thread-cert/v1_2_LowPower*.py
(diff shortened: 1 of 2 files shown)


Irving-cl:
Sounds good to me👍  Thanks!

simonlingoogle:
Close for now since we'd better fix these fails.

PR #5581, opened 2020-09-28
[cli] style fixes and smaller enhancements

abtink:
This commit contains a group of style fixes and smaller enhancements
in CLI modules:
- Remove unnecessary use of `struct` when declaring variable.
- Remove use of single `"%s"` in `Output` methods.
- Fix order of `#include` headers.
- Remove unnecessary use of `OT_UNUSED_VARIABLE`.
- Remove unnecessary temp variables and casts.
- Add comments to `#endif` directives than are more than  than 20
  lines away from the corresponding `#if` they are associated with.

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -35,41 +35,30 @@
 
 #include <stdio.h>
 #include <stdlib.h>
-#include "mac/channel_mask.hpp"
-#include "utils/parse_cmdline.hpp"
 
+#include <openthread/diag.h>
 #include <openthread/icmp6.h>
 #include <openthread/link.h>
+#include <openthread/logging.h>
 #include <openthread/ncp.h>
 #include <openthread/thread.h>
+#include <openthread/platform/uart.h>
 #if OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
 #include <openthread/network_time.h>
 #endif
-
 #if OPENTHREAD_FTD
 #include <openthread/dataset_ftd.h>
 #include <openthread/thread_ftd.h>
 #endif
-
 #if OPENTHREAD_CONFIG_BORDER_ROUTER_ENABLE
 #include <openthread/border_router.h>
 #endif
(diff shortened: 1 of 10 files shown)


src/cli/cli_udp.cpp
```diff
@@ -317,8 +317,8 @@ void UdpExample::HandleUdpReceive(void *aContext, otMessage *aMessage, const otM
 
 void UdpExample::HandleUdpReceive(otMessage *aMessage, const otMessageInfo *aMessageInfo)
 {
-    uint8_t buf[1500];
-    int     length;
+    char buf[1500];
```
simonlingoogle, line 320: Maybe not in the scope of this PR, but this local variable take so much stack memory. Could it be a potential threat to cause stack overflow?

abtink: I agree. It would be better to avoid/minimize use of large stack variables. I think it may be a good future fix (in a future PR). That said, I think it's not very critical in this case since it is being used  in CLI UDP examples which I guess is mainly intended for quick testing.

src/cli/cli.cpp
```diff
@@ -35,41 +35,30 @@
...
-#include "mac/channel_mask.hpp"
-#include "utils/parse_cmdline.hpp"
 
+#include <openthread/diag.h>
 #include <openthread/icmp6.h>
 #include <openthread/link.h>
+#include <openthread/logging.h>
 #include <openthread/ncp.h>
 #include <openthread/thread.h>
+#include <openthread/platform/uart.h>
```
Irving-cl, line 45: This line should be put before `<openthread/thread.h>`?

Irving-cl (approved):
LGTM👍 

A small nit suggestion.

src/cli/cli.cpp
Irving-cl: Make sense to me. Thanks!

jwhui (approved):

PR #5582, opened 2020-09-28
[scripts] set FED routereligible disable

simonlingoogle:
Instead of setting [routerupgradethreshold](https://github.com/openthread/openthread/blob/master/src/cli/README.md#routerupgradethreshold) to 0, it's more correct for `FED` to turn off [routereligible](https://github.com/openthread/openthread/blob/master/src/cli/README.md#routereligible).

- [ ] WIP: support `routereligible` on NCP (spinel-cli.py)


simonlingoogle:
Will do it when have time.

PR #5583, opened 2020-09-28
[low-power] add an Enh-ACK frame link layer counter

brianljt:
This PR is to fix the issue discussed in Thread group [SPEC-939](https://threadgroup.atlassian.net/browse/SPEC-939). The approach is,

On the transmitter side, only maintain a single frame counter, and never reuse it.

On the receiver side, maintain a second incoming frame counter `eFC` alongside `FC` for each device, where `eFC >= FC >= 0`. When an Enh-ACK is received, verify it against `eFC`. Update `eFC` to be equal to the received frame counter. Do not touch `FC`. When a frame other than Enh-ACK is received, verify it against `FC`. Update both `eFC` and `FC` to be equal to the received frame counter.


include/openthread/thread.h
```diff
@@ -90,21 +90,24 @@ typedef struct otLinkModeConfig
...
-    bool         mSecureDataRequest : 1; ///< Secure Data Requests
-    bool         mFullThreadDevice : 1;  ///< Full Thread Device
-    bool         mFullNetworkData : 1;   ///< Full Network Data
-    bool         mIsChild : 1;           ///< Is the neighbor a child
+    otExtAddress mExtAddress;       ///< IEEE 802.15.4 Extended Address
+    uint32_t     mAge;              ///< Time last heard
+    uint16_t     mRloc16;           ///< RLOC16
+    uint32_t     mLinkFrameCounter; ///< Link Frame Counter
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+    uint32_t mLinkAckFrameCounter; ///< Link ACK Frame Counter
```
abtink, line 98: Do we need to provide the `mLinkAckFrameCounter` in the public APIs? If there is no use-case I suggest not changing the public definitions/API.

src/core/mac/mac.cpp
```diff
@@ -1637,6 +1637,12 @@ otError Mac::ProcessReceiveSecurity(RxFrame &aFrame, const Address &aSrcAddr, Ne
         }
 
         aNeighbor->SetLinkFrameCounter(frameCounter + 1);
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+        if (aNeighbor->GetLinkFrameCounter() > aNeighbor->GetLinkAckFrameCounter())
```
abtink, line 1641: How about the follwoing: (I am thinking of this to help reduce the potenrtial conflicts with the multi-radio code and its  rebase if/when this is merged.).
```suggestion
        if (frameCounter + 1 > aNeighbor->GetLinkAckFrameCounter())
```

src/ncp/ncp_base_mtd.cpp
```diff
@@ -182,6 +182,9 @@ otError NcpBase::EncodeNeighborInfo(const otNeighborInfo &aNeighborInfo)
     SuccessOrExit(error = mEncoder.WriteUint32(aNeighborInfo.mLinkFrameCounter));
     SuccessOrExit(error = mEncoder.WriteUint32(aNeighborInfo.mMleFrameCounter));
     SuccessOrExit(error = mEncoder.WriteInt8(aNeighborInfo.mLastRssi));
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+    SuccessOrExit(error = mEncoder.WriteUint32(aNeighborInfo.mLinkAckFrameCounter));
```
abtink, line 186: If we update this, I think we should also update the `spinel.h` and the documentation of this property

abtink (commented):
Looks good overall. Thanks @brianljt 
Happy to see the approach/fix being used by spec (this is what we wanted). "This is the way" 😃

include/openthread/thread.h
brianljt: Thank you for the suggestions. Remove the link ACK frame counter from neighbor info which is used for debug only and there are no user case for now.

abtink (approved):
LGTM. 👍

librasungirl (approved):
LGTM 👍

Irving-cl (approved):
LGTM👍

src/core/mac/mac.cpp
```diff
@@ -1677,35 +1685,29 @@ otError Mac::ProcessEnhAckSecurity(TxFrame &aTxFrame, RxFrame &aAckFrame)
...
         neighbor = Get<NeighborTable>().FindNeighbor(srcAddr);
-
-        if (neighbor != nullptr)
-        {
-            srcAddr.SetExtended(neighbor->GetExtAddress());
-        }
     }
 
-    if (!srcAddr.IsExtended())
+    if (neighbor == nullptr && !dstAddr.IsNone())
```
bukepo, line 1699: Any case response source is not None and still need to find the neighbor according to the original destination address?

brianljt: Updated as suggested

bukepo (approved):
LGTM:+1:

src/core/mac/mac.cpp
```diff
@@ -1677,35 +1685,28 @@ otError Mac::ProcessEnhAckSecurity(TxFrame &aTxFrame, RxFrame &aAckFrame)
 
     VerifyOrExit(txKeyId == ackKeyId, OT_NOOP);
 
+    IgnoreError(aAckFrame.GetFrameCounter(frameCounter));
+    otLogDebgMac("Rx security - Ack frame counter %u", frameCounter);
+
     IgnoreError(aAckFrame.GetSrcAddr(srcAddr));
+    IgnoreError(aTxFrame.GetDstAddr(dstAddr));
```
jwhui, line 1692: Should we move this into the else block at L1698 below?

src/core/thread/topology.hpp
```diff
@@ -582,6 +607,9 @@ class Neighbor : public InstanceLocatorInit
         {
             uint32_t mLinkFrameCounter; ///< The Link Frame Counter
             uint32_t mMleFrameCounter;  ///< The MLE Frame Counter
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+            uint32_t mLinkAckFrameCounter; ///< The Link Ack Frame Counter
```
jwhui, line 611: Maybe in a future PR, we could save some memory here by not storing the full 32 bits. For example, I expect the difference between `mLinkFrameCounter` and `mLinkAckFrameCounter` to be small.

brianljt: Maybe we can use uint8_t to save the offset between two counter. However, in current structure, alignment padding might result in similar memory usage as uint32_t?

jwhui: I think it is fine to leave for now. In future, we may want the additional bytes for other features.

jwhui (approved):

PR #5584, opened 2020-09-28
[thread-cert] enhance case 9.2.18 for SED packet order

FiveDimensions:
copy the pkt to verify Step 13 and Step 14 SED packets since the packets could be in front of the packet in Step 12


simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #5585, opened 2020-09-28
[posix] fix MLE sends ICMPv6 destination unreachable

simonlingoogle:
Fixes https://github.com/openthread/ot-br-posix/issues/491

### DIFF
--- a/src/core/net/ip6.cpp
+++ b/src/core/net/ip6.cpp
@@ -978,9 +978,11 @@ otError Ip6::ProcessReceiveCallback(Message &          aMessage,
     if (mIsReceiveIp6FilterEnabled)
     {
         // do not pass messages sent to an RLOC/ALOC, except Service Locator
+#if !OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE
         VerifyOrExit(!aMessageInfo.GetSockAddr().GetIid().IsLocator() ||
                          aMessageInfo.GetSockAddr().GetIid().IsAnycastServiceLocator(),
                      error = OT_ERROR_NO_ROUTE);
+#endif
 
         switch (aIpProto)
         {
@@ -1011,13 +1013,13 @@ otError Ip6::ProcessReceiveCallback(Message &          aMessage,
                 // do not pass MLE messages
                 ExitNow(error = OT_ERROR_NO_ROUTE);
             }
-#if !OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
             else if ((destPort == Tmf::kUdpPort) && Get<Tmf::TmfAgent>().IsTmfMessage(aMessageInfo))
             {
-                // do not pass TMF messages
+                // pass TMF messages only if `OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE`
+#if !OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
                 ExitNow(error = OT_ERROR_NO_ROUTE);
-            }
 #endif
+            }
 
 #if OPENTHREAD_FTD
             if (destPort == Get<MeshCoP::JoinerRouter>().GetJoinerUdpPort())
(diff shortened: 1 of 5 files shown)


abtink (approved):
LGTM. 👍

src/core/net/ip6.cpp
```diff
@@ -984,10 +984,12 @@ otError Ip6::ProcessReceiveCallback(Message &          aMessage,
 
     if (mIsReceiveIp6FilterEnabled)
     {
+#if !OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
```
bukepo, line 987:
```suggestion
#if !OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE
```

tests/scripts/thread-cert/pktverify/packet_filter.py
```diff
@@ -25,7 +25,7 @@
 #  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 #  POSSIBILITY OF SUCH DAMAGE.
 #
-
+import logging
```
bukepo, line 28: keep the blank line?

simonlingoogle, line 28:
```suggestion

import logging
```

src/core/net/ip6.cpp
simonlingoogle: Considering our current OTBR solutions. 

| OTBR Solutions                   | PLAT_NETIF | PLAT_UDP | Should Pass RLOC/ALOC ? |
|----------------------------------|------------|----------|-------------------------|
| OT-Daemon + RCP-Dongle           | Yes        | Yes      | Yes                     |
| Wpantund + NCP-Dongle            | No         | No       | No                      |
| Wpantund + NCP-Host + RCP-Dongle | No         | Yes      | TMF Yes, Other No       |

Do you agree on this table? @bukepo @abtink @jwhui @librasungirl 

If yes, I think we should move the `RLOC/ALOC filtering code` down to be below `TMF filtering` so that we can pass TMF message but not other messages with RLOC/ALOC destinations when PLAT_UDP is ON.

src/core/net/ip6.cpp
```diff
@@ -978,9 +978,11 @@ otError Ip6::ProcessReceiveCallback(Message &          aMessage,
     if (mIsReceiveIp6FilterEnabled)
     {
         // do not pass messages sent to an RLOC/ALOC, except Service Locator
+#if !OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE
```
bukepo, line 981: Can RLOC/ALOC be determined by the IID only?

src/posix/platform/netif.cpp
```diff
@@ -1457,6 +1457,7 @@ void platformNetifInit(otInstance *aInstance, const char *aInterfaceName)
     mldListenerInit();
 #endif
 
+    otIp6SetReceiveFilterEnabled(aInstance, true);
```
bukepo, line 1460: How about enhancing the otIp6SetReceiveFilterEnabled() API to a API which configures the detailed filter rules instead of just enable/disable?

simonlingoogle: Currently, the changes introduced by this PR is just what we need, which fixes https://github.com/openthread/ot-br-posix/issues/491 . 
I think it's unclear whether or not an enhanced implementation like `otIp6SetReceiveFilterEnabled` is necessary for now, which also could unnecessarily increase code size. 

I prefer to enhance it when we really have the requirement in the future. 

Thoughts?

src/core/net/ip6.cpp
simonlingoogle: Using `!OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE` should be OK if we want to pass RLOC/ALOC dst messages when PLAT_NETIF is ON.

src/posix/platform/netif.cpp
bukepo: wpantund has a option to enable/disable mapping RLOC/ALOC addresses, which means even when platform netif is not enabled, the messages may still need to be delivered to host IP stack. I'm fine with enhancing this later.

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #5587, opened 2020-09-28
[cli] add context for the user commands

gjc13:
This makes using a class method as the cli handler easier.

### DIFF
--- a/src/posix/main.c
+++ b/src/posix/main.c
@@ -56,6 +56,7 @@
 #define OT_POSIX_APP_TYPE_NCP 1
 #define OT_POSIX_APP_TYPE_CLI 2
 
+#include <openthread/cli.h>
 #include <openthread/diag.h>
 #include <openthread/logging.h>
 #include <openthread/tasklet.h>
@@ -250,24 +251,30 @@ static void ParseArg(int aArgCount, char *aArgVector[], PosixConfig *aConfig)
     aConfig->mPlatformConfig.mRadioUrl = aArgVector[optind];
 }
 
-static otInstance *InitInstance(int aArgCount, char *aArgVector[])
+#if OPENTHREAD_POSIX_APP_TYPE == OT_POSIX_APP_TYPE_CLI
+static void PrintRadioUrl(void *aContext, uint8_t aArgsLength, char *aArgs[])
 {
-    PosixConfig config;
+    (void)aArgsLength;
+    (void)aArgs;
+
+    otPlatformConfig *config = (otPlatformConfig *)aContext;
+    otCliOutputFormat("%s\r\nDone\r\n", config->mRadioUrl);
+}
+#endif // OPENTHREAD_POSIX_APP_TYPE == OT_POSIX_APP_TYPE_CLI
+
+static otInstance *InitInstance(PosixConfig *aConfig)
+{
     otInstance *instance = NULL;
 
-    ParseArg(aArgCount, aArgVector, &config);
-
-    openlog(aArgVector[0], LOG_PID | (config.mIsVerbose ? LOG_PERROR : 0), LOG_DAEMON);
-    setlogmask(setlogmask(0) & LOG_UPTO(LOG_DEBUG));
     syslog(LOG_INFO, "Running %s", otGetVersionString());
     syslog(LOG_INFO, "Thread version: %hu", otThreadGetVersion());
-    IgnoreError(otLoggingSetLevel(config.mLogLevel));
+    IgnoreError(otLoggingSetLevel(aConfig->mLogLevel));
 
-    instance = otSysInit(&config.mPlatformConfig);
+    instance = otSysInit(&aConfig->mPlatformConfig);
 
     atexit(otSysDeinit);
 
-    if (config.mPrintRadioVersion)
+    if (aConfig->mPrintRadioVersion)
     {
         printf("%s\n", otPlatRadioGetVersionString(instance));
     }
@@ -276,7 +283,7 @@ static otInstance *InitInstance(int aArgCount, char *aArgVector[])
         syslog(LOG_INFO, "RCP version: %s", otPlatRadioGetVersionString(instance));
     }
 
-    if (config.mIsDryRun)
+    if (aConfig->mIsDryRun)
     {
         exit(OT_EXIT_SUCCESS);
     }
@@ -304,6 +311,10 @@ int main(int argc, char *argv[])
 {
     otInstance *instance;
     int         rval = 0;
+    PosixConfig config;
+#if OPENTHREAD_POSIX_APP_TYPE == OT_POSIX_APP_TYPE_CLI
+    otCliCommand radioUrlCommand = {"radiourl", PrintRadioUrl};
+#endif
 
 #ifdef __linux__
     // Ensure we terminate this process if the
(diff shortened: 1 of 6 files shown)


bukepo (approved):
LGTM:+1:

abtink:
Thanks @gjc13. In general adding `mContext` sounds ok to me, but since we already need to change the public CLI definition (which I guess also requires an API version increase in `instance.h`, and breaks compatibility with older code, i.e.  requires changes wherever these definitions were being used before), I have a suggestion for a different approach which I think would give us more flexibility in future:

We change the definition to use a linked-list model. Also we can change the callback handler function pointer type to pass a pointer to the command entry ` struct otCliCommand *aCommand)` itself (from the list) as a parameter:
```
typedef struct otCliCommand
{
    const struct otCliCommand *mNext;                     ///< A pointer to the next command.
    const char *mName;                                    ///< A pointer to the command string.
    void (*mCommand)(uint8_t aArgsLength, char *aArgs[], const struct otCliCommand *aCommand); 
                                                          ///< A function pointer to process the command.
} otCliCommand;
```

We then also change `otCliSetUserCommands()` to maybe `otCliRegisterUserCommands()` and allow it to be called from different modules 

This model has some advantages:
- Using a linked-list (instaed of array) makes it possible to have multiple modules register user CLI commands (e.g., we can have radio platform layer add its own CLI commands and app layer its own, etc). In the code we just link all the list given to us.
- Passing the `struct otCliCommand *aCommand` in the callback gives us more flexibility in its use. It allows the caller/user to sub-class `otCliCommand` and add any number of additional info/state/context variables associated the command (it can even be different sub-classes for different commands). In the callback, the user can cast the `aCommand` paramter to the sub-class type and access the member variables/info.

gjc13:
> Thanks @gjc13. In general adding `mContext` sounds ok to me, but since we already need to change the public CLI definition (which I guess also requires an API version increase in `instance.h`, and breaks compatibility with older code, i.e. requires changes wherever these definitions were being used before), I have a suggestion for a different approach which I think would give us more flexibility in future:
> 
> We change the definition to use a linked-list model. Also we can change the callback handler function pointer type to pass a pointer to the command entry ` struct otCliCommand *aCommand)` itself (from the list) as a parameter:
> 
> ```
> typedef struct otCliCommand
> {
>     const struct otCliCommand *mNext;                     ///< A pointer to the next command.
>     const char *mName;                                    ///< A pointer to the command string.
>     void (*mCommand)(uint8_t aArgsLength, char *aArgs[], const struct otCliCommand *aCommand); 
>                                                           ///< A function pointer to process the command.
> } otCliCommand;
> ```
> 
> We then also change `otCliSetUserCommands()` to maybe `otCliRegisterUserCommands()` and allow it to be called from different modules
> 
> This model has some advantages:
> 
> * Using a linked-list (instaed of array) makes it possible to have multiple modules register user CLI commands (e.g., we can have radio platform layer add its own CLI commands and app layer its own, etc). In the code we just link all the list given to us.
> * Passing the `struct otCliCommand *aCommand` in the callback gives us more flexibility in its use. It allows the caller/user to sub-class `otCliCommand` and add any number of additional info/state/context variables associated the command (it can even be different sub-classes for different commands). In the callback, the user can cast the `aCommand` paramter to the sub-class type and access the member variables/info.
> 
> **Edit**: One limitation/challenge I see with allowing multiple linked-lists to be appended to each other is that it would require the `mNext` in the entry to be changeable which would not allow user to define the entries/list as `const` or `constexpr` (place them flash instead of RAM). So I think it may be better to allow lists to be seperate and constant. We can allow some `x` number of user CLI command lists to be registered (we can have `x` to be a CLI speicifc OT config parmeters)

Agree with the linked list approach to make possible register handlers from multiple modules. I'm not very comfortable with forcing users to subclass `otCliCommand` especially when it's declared as an `extern "C"` struct. I'd like to add context to `otCliCommand` and leave linked list an an inner implementation. Does it look fine to you?

gjc13:
Another advantage of add an array to store the user commands is to eliminate the implicit requirement for the lifetime of the `otCliCommand` struct.

src/cli/cli_config.h
```diff
@@ -99,4 +99,14 @@
...
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_MAX_USER_COMMAND_NUMBER
+ *
+ * The maxium number of supported user commands.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_MAX_USER_COMMAND_NUMBER
+#define OPENTHREAD_CONFIG_MAX_USER_COMMAND_NUMBER 10
```
abtink, line 109: Suggest using a smaller default number here in the config to avoid using memory? I would say 2 maybe?

abtink (approved):
LGTM. 👍

src/cli/cli_config.h
jwhui, line 109:
```suggestion
#ifndef OPENTHREAD_CONFIG_CLI_MAX_USER_COMMANDS
#define OPENTHREAD_CONFIG_CLI_MAX_USER_COMMANDS 10
```

src/posix/main.c
```diff
@@ -250,24 +251,29 @@ static void ParseArg(int aArgCount, char *aArgVector[], PosixConfig *aConfig)
...
 }
 
-static otInstance *InitInstance(int aArgCount, char *aArgVector[])
+#if OPENTHREAD_POSIX_APP_TYPE == OT_POSIX_APP_TYPE_CLI
+static void PrintRadioUrl(void *aContext, uint8_t aArgsLength, char *aArgs[])
 {
-    PosixConfig config;
-    otInstance *instance = NULL;
+    (void)aArgsLength;
+    (void)aArgs;
```
jwhui, line 258:
```suggestion
    OT_UNUSED_VARIABLE(aArgsLength);
    OT_UNUSED_VARIABLE(aArgs);
```

src/posix/main.c
```diff
@@ -250,24 +251,29 @@ static void ParseArg(int aArgCount, char *aArgVector[], PosixConfig *aConfig)
...
+    otPlatformConfig *config = (otPlatformConfig *)aContext;
+    otCliOutputFormat("%s\r\nDone\r\n", config->mRadioUrl);
+}
+#endif // OPENTHREAD_POSIX_APP_TYPE == OT_POSIX_APP_TYPE_CLI
 
-    openlog(aArgVector[0], LOG_PID | (config.mIsVerbose ? LOG_PERROR : 0), LOG_DAEMON);
-    setlogmask(setlogmask(0) & LOG_UPTO(LOG_DEBUG));
+static otInstance *InitInstance(PosixConfig *aConfig)
+{
+    otInstance *instance = NULL;
```
jwhui, line 267:
```suggestion
    otInstance *instance = NULL;

```

gjc13:
@abtink Moved back to the plain context mode.

src/posix/main.c
```diff
@@ -250,24 +251,30 @@ static void ParseArg(int aArgCount, char *aArgVector[], PosixConfig *aConfig)
...
+    (void)aArgsLength;
+    (void)aArgs;
 
-    ParseArg(aArgCount, aArgVector, &config);
+    otPlatformConfig *config = (otPlatformConfig *)aContext;
+    otCliOutputFormat("%s\r\nDone\r\n", config->mRadioUrl);
+}
+#endif // OPENTHREAD_POSIX_APP_TYPE == OT_POSIX_APP_TYPE_CLI
+
+static otInstance *InitInstance(PosixConfig *aConfig)
```
bukepo, line 265: Is this change necessary?

src/posix/main.c
```diff
@@ -330,6 +344,7 @@ int main(int argc, char *argv[])
 #else
     otCliUartInit(instance);
 #endif
+    otCliSetUserCommands(&radioUrlCommand, 1, &config.mPlatformConfig);
```
bukepo, line 347: This is just for coverage purpose for now. I tend to keep `PosixConfig config` a temporary variable inside the `InitInstance()` function and revert the change to `InitInstance()`.

```suggestion
    otCliSetUserCommands(&radioUrlCommand, 1, argv[optind]);
```

abtink (approved):
LGTM

bukepo (approved):

jwhui (approved):

PR #5588, opened 2020-09-28
[cli] add command `uptime` to cli

zhanglongxia:
This PR adds a command `uptime` to print how long the device has been running.

### DIFF
--- a/src/cli/README.md
+++ b/src/cli/README.md
@@ -99,6 +99,7 @@ Done
 - [thread](#thread-start)
 - [txpower](#txpower)
 - [unsecureport](#unsecureport-add-port)
+- [uptime](#uptime)
 - [version](#version)
 
 ## OpenThread Command Details
@@ -1935,6 +1936,16 @@ Print all ports from the allowed unsecured port list.
 ```bash
 > unsecureport get
 1234
+Done
+```
+
+### uptime
+
+Print how long the device has been running.
+
+```bash
+> uptime
+00:00:51 up 0 days
 Done
 ```
 
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -155,6 +155,8 @@ Interpreter::Interpreter(Instance *aInstance)
 #if OPENTHREAD_CONFIG_DNS_CLIENT_ENABLE
     memset(mResolvingHostname, 0, sizeof(mResolvingHostname));
 #endif
+
+    mStartTimeUs = otPlatTimeGet();
 }
 
 int Interpreter::Hex2Bin(const char *aHex, uint8_t *aBin, uint16_t aBinLength, bool aAllowTruncate)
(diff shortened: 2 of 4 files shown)


src/cli/cli.cpp
```diff
@@ -3767,6 +3769,34 @@ otError Interpreter::ProcessTxPower(uint8_t aArgsLength, char *aArgs[])
     return error;
 }
 
+otError Interpreter::ProcessUptime(uint8_t aArgsLength, char *aArgs[])
+{
+    OT_UNUSED_VARIABLE(aArgs);
+
+    otError  error = OT_ERROR_NONE;
+    uint32_t now;
+    uint32_t seconds;
```
simonlingoogle, line 3778:
```suggestion
    uint8_t seconds;
```

`uint8_t` is enough for 0 ~ 59 ?

src/cli/cli.cpp
```diff
@@ -3767,6 +3769,34 @@ otError Interpreter::ProcessTxPower(uint8_t aArgsLength, char *aArgs[])
...
+    uint32_t minutes;
```
simonlingoogle, line 3779:
```suggestion
    uint8_t minutes;
```
ditto

src/cli/cli.cpp
```diff
@@ -3767,6 +3769,34 @@ otError Interpreter::ProcessTxPower(uint8_t aArgsLength, char *aArgs[])
...
+    uint32_t hours;
```
simonlingoogle, line 3780:
```suggestion
    uint8_t hours;
```
ditto

src/cli/cli.cpp
```diff
@@ -3767,6 +3769,34 @@ otError Interpreter::ProcessTxPower(uint8_t aArgsLength, char *aArgs[])
...
+    now = static_cast<uint32_t>((otPlatTimeGet() - mStartTimeUs) / 1000000);
+
+    seconds = now % 60;
+    now     = now / 60;
+    minutes = now % 60;
+    now     = now / 60;
+    hours   = now % 24;
+    days    = now / 24;
+
+    OutputLine("%02d:%02d:%02d up %d days", hours, minutes, seconds, days);
```
simonlingoogle, line 3794:
```suggestion
    OutputLine("%dd%dh%dm%ds", days, hours, minutes, seconds);
```
nit: formats like `3d20h30m59s` is easier to parse?

src/cli/README.md
```diff
@@ -1938,6 +1939,16 @@ Print all ports from the allowed unsecured port list.
 Done
 ```
 
+### uptime
+
+Print how long the device has been running.
```
simonlingoogle, line 1944: nit: 
```suggestion
Print how long the device has been running since started or reset.
```

abtink (approved):
LGTM. 👍

bukepo (commented):
If this is for debugging, should we provide a option to condition out this feature?

simonlingoogle:
> If this is for debugging, should we provide a option to condition out this feature?

We can use `OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE` if this is for debugging only. 
But personally I think this is a good command to have for all CLI devices.

zhanglongxia:
@bukepo @simonlingoogle The command "uptime" is used to check how long the device has started or reset. Sometimes, the device is accidentally reset. We can use this command to check this case. All platforms are able to support it. I thought it is a common command,  it may be a good choice to enable it on all platforms. Comments?

PR #5589, opened 2020-09-28
[continuous-integration] fix API change detector

Moandor-y:
API version check always passes with "no updates", this fixes it.


bukepo (approved):
good catch!

simonlingoogle (approved):
LGTM 👍 

Good Catch!

jwhui (approved):

PR #5602, opened 2020-10-01
[utils] adding 'LookupTable' class and use it in CLI

abtink:
This commit adds `LookupTable` module which provides helper methods to
perform binary search in a generic sorted lookup table for specific
entry matching a given name string. It also provides `constexpr`
methods to check (at build-time) whether a lookup table array is
sorted. This commit also updates CLI modules to use the `LookupTable`
methods for command lookup. It also adds a unit test for the lookup
table.

### DIFF
--- a/src/core/utils/lookup_table.hpp
+++ b/src/core/utils/lookup_table.hpp
@@ -1,0 +1,151 @@
+/*
+ *  Copyright (c) 2020, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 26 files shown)


src/core/utils/lookup_table.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+    template <class EntryType> static constexpr bool IsSorted(const EntryType *aTable, uint16_t aLength)
+    {
+        return (aLength <= 1) ? true
+                              : AreInOrder(aTable[0].mName, aTable[1].mName) && IsSorted(aTable + 1, aLength - 1);
+    }
+
+    constexpr static bool AreInOrder(const char *aFirst, const char *aSecond)
+    {
+        return (*aFirst < *aSecond) ? true : ((*aFirst > *aSecond) ? false : AreInOrder(aFirst + 1, aSecond + 1));
```
jwhui, line 136: Do we properly handle strings that are equal? In other words, what if two adjacent strings are the same?

abtink: Good point. I will update it (will add a case for it in unit test as well).

abtink: Added this on new push (we ensure that there are no duplicate names in the table).

jwhui (approved):

eujean:
Getting a few undefined references when building an example for cc1352:

> src/core/utils/lookup_table.hpp:117: undefined reference to `ot::Utils::LookupTable::Find(char const*, void const*, unsigned short, unsigned short, char const* (&)(void const*))'

jwhui:
> Getting a few undefined references when building an example for cc1352:
> 
> > src/core/utils/lookup_table.hpp:117: undefined reference to `ot::Utils::LookupTable::Find(char const*, void const*, unsigned short, unsigned short, char const* (&)(void const*))'

@eujean , did you try running `./bootstrap` again? Can you provide the build commands that you are using? Note that the cc1352 build check is [currently passing](https://github.com/openthread/openthread/runs/1206649393?check_suite_focus=true).

PR #5603, opened 2020-10-01
[cli] add new helper 'ParseIp6Prefix()'

abtink:


src/cli/cli.cpp
```diff
@@ -277,6 +277,26 @@ otError Interpreter::ParseJoinerDiscerner(char *aString, otJoinerDiscerner &aDis
...
 }
 
+otError Interpreter::ParseIp6Prefix(char *aString, otIp6Prefix &aPrefix)
+{
+    otError       error = OT_ERROR_NONE;
+    char *        prefixLengthStr;
+    unsigned long length;
+
+    prefixLengthStr = strchr(aString, '/');
+    VerifyOrExit(prefixLengthStr != nullptr, error = OT_ERROR_INVALID_ARGS);
```
simonlingoogle, line 287: use default 64 if not specified?

src/cli/cli.cpp
```diff
@@ -277,6 +277,26 @@ otError Interpreter::ParseJoinerDiscerner(char *aString, otJoinerDiscerner &aDis
...
+
+    *prefixLengthStr++ = '\0';
+
+    SuccessOrExit(error = otIp6AddressFromString(aString, &aPrefix.mPrefix));
+
+    SuccessOrExit(error = ParseUnsignedLong(prefixLengthStr, length));
+    aPrefix.mLength = static_cast<uint8_t>(length);
```
simonlingoogle, line 294: check that `length` is in the valid range (0 ~ 128 ? )

abtink: We can add such a check but may be redundant since the OT APIs (that would use with the prefix) would actually verify the prefix length(e.g., `Local::AddPrefix()` method in `network_data_local.cpp`). In my view, it's better to perform such checks at API call interface such that it's common.

In general, I noticed that CLI modules do not perform checks  on validity of input value (e.g., in many places we cast from `ulong/long` to `uint8_t/uint16_t` without checking if it is within valid range). I think this is actually fine since CLI is intended for testing and in general for (advanced) users who know what they want do. Adding checks would make it safer but will impact code size.

src/cli/cli.cpp
abtink: I think it may be better to require the length to be given explicitly here and not use a default value (since this is used for an IPv6 prefix (not an Pv6 address)).

simonlingoogle: Fine to be more strict.

src/cli/cli.cpp
simonlingoogle: > We can add such a check but may be redundant since the OT APIs (that would use with the prefix) would actually verify the prefix length(e.g., `Local::AddPrefix()` method in `network_data_local.cpp`). In my view, it's better to perform such checks at API call interface such that it's common.

Fine to me if `AddPrefix` has further validation. 

> I think this is actually fine since CLI is intended for testing and in general for (advanced) users who know what they want do. Adding checks would make it safer but will impact code size.

Not really. Currently, OT CLI is the public interface of `ot-ctl`, thus also OTBR. 


> In general, I noticed that CLI modules do not perform checks on validity of input value (e.g., in many places we cast from `ulong/long` to `uint8_t/uint16_t` without checking if it is within valid range). 

Actually I think other CLI commands should fix these issues too (treating 256 as 0 is simply incorrect). But maybe we can fix them in a uniform way. e.g. introduce a `ParseUint8` method to parse uint8 integers.

simonlingoogle (approved):
LGTM 👍

jwhui (approved):

src/cli/cli.cpp
abtink: > Actually I think other CLI commands should fix these issues too (treating 256 as 0 is simply incorrect). But maybe we can fix them in a uniform way. e.g. introduce a ParseUint8 method to parse uint8 integers.

I have something like this in an older branch -> https://github.com/abtink/openthread/commit/3ef428998e11b7dd1b2b65a2ee2250126894f154
It does impact code size, but may be worth the safty it would add in the use of CLI and checking the paramter

PR #5604, opened 2020-10-02
[message] add new flavors of Read/Write/Append/Prepend methods

abtink:
This commit adds new helper methods in `Message` class. It adds new
template methods to allow an object (of any type) to be read from or
written/appended/prepended to the message. The `Read<ObjectType>()`
method returns a parse `otError` if there are not enough bytes
remaining in the message to read the entire object.

The template methods are used in other core modules which help
simplify the code (there is no need to specify the template type since
it can be deduced by the complier from the passed-in argument). Also
since the template methods directly use the `sizeof` the object type,
they help reduce the chance of incorrect use (using incorrect
read/write length when calling the method).

This commit also updates and renames the methods which read/write raw
bytes from/to the message to `ReadBytes()`/`WriteBytes()`. The order
of parameters in these methods are changed to follow the common
pattern used by other modules and the OT public APIs (i.e., the
pointer to the buffer is given first, followed by its length).

### DIFF
--- a/src/core/common/message.hpp
+++ b/src/core/common/message.hpp
@@ -633,9 +633,25 @@ public:
     uint16_t ReadBytes(uint16_t aOffset, void *aBuf, uint16_t aLength) const;
 
     /**
+     * This method reads a given number of bytes from the message.
+     *
+     * If there are fewer bytes available in the message than the requested read length, the available bytes will be
+     * read and copied into @p aBuf. In this case `OT_ERROR_PARSE` will be returned.
+     *
+     * @param[in]  aOffset  Byte offset within the message to begin reading.
+     * @param[out] aBuf     A pointer to a data buffer to copy the read bytes into.
+     * @param[in]  aLength  Number of bytes to read.
+     *
+     * @retval OT_ERROR_NONE     @p aLength bytes were successfully read from message.
+     * @retval OT_ERROR_PARSE    Not enough bytes remaining in message to read the entire object.
+     *
+     */
+    otError Read(uint16_t aOffset, void *aBuf, uint16_t aLength) const;
+
+    /**
      * This method reads an object from the message.
      *
-     * If there are less bytes available in the message than the requested object size, the available bytes will be
+     * If there are fewer bytes available in the message than the requested object size, the available bytes will be
      * read and copied into @p aObject (@p aObject will be read partially). In this case `OT_ERROR_PARSE` will
      * be returned.
      *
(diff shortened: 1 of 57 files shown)


src/core/common/message.hpp
```diff
@@ -553,14 +553,32 @@ class Message : public Buffer
      *
      * On success, this method grows the message by @p aLength bytes.
      *
-     * @param[in]  aBuf     A pointer to a data buffer.
+     * @param[in]  aBuf     A pointer to a data buffer (can be `nullptr` to grow message without writing bytes)
```
jwhui, line 556:
```suggestion
     * @param[in]  aBuf     A pointer to a data buffer (can be `nullptr` to grow message without writing bytes).
```

src/core/common/message.hpp
```diff
@@ -582,19 +600,58 @@ class Message : public Buffer
...
+     * read and copied into @p aObject (@p aObject will be read partially). In this case `OT_ERROR_PARSE` will
+     * be returned.
+     *
+     * @tparam     ObjectType   The object type to read from the message.
+     *
+     * @param[in]  aOffset      Byte offset within the message to begin reading.
+     * @param[out] aObject      A reference to the object to read into.
+     *
+     * @retval OT_ERROR_NONE     Object @p aObject was successfully read from message.
+     * @retval OT_ERRRO_PARSE    Not enough bytes remaining in message to read the entire object.
```
jwhui, line 648:
```suggestion
     * @retval OT_ERROR_PARSE    Not enough bytes remaining in message to read the entire object.
```

src/core/thread/mesh_forwarder.cpp
```diff
@@ -1292,38 +1292,37 @@ otError MeshForwarder::ParseIp6UdpTcpHeader(const Message &aMessage,
                                             uint16_t &     aDestPort)
 {
     otError error = OT_ERROR_PARSE;
-    union
```
jwhui, line 1295: Does this change increase stack overhead?

src/core/common/message.hpp
```diff
@@ -582,19 +600,58 @@ class Message : public Buffer
      * @retval OT_ERROR_NO_BUFS  Insufficient available buffers to grow the message.
      *
      */
-    otError Append(const void *aBuf, uint16_t aLength);
+    otError AppendBytes(const void *aBuf, uint16_t aLength);
```
jwhui, line 603: Document whether `aBuf` can be `nullptr`?

src/core/thread/mesh_forwarder.cpp
abtink: I think it may. It depends on compiler/toolchain.
Will change it back.

abtink:
In new push, added a new commit (to be squashed) on top of the previous change/commit. Noticed that we have the pattern where reading bytes/object from message and then checking that the number of bytes read to equal the request read length. We can use the non-templated `Read()` method (which was earlier `private`) for this purpose.

jwhui (approved):

PR #5608, opened 2020-10-06
[low-power] implement forward tracking series

Irving-cl:
This PR implements the **Forward Tracking Series** feature in Link Metrics. Here are a few things to mention in this PR:
1. Two new apis are added: 
    * `otLinkMetricsSendMgmtRequestForwardTrackingSeries`, which is used to send an MLE Link Metrics Management Request
    * `otLinkMetricsSendLinkProbe`, which is used to send an MLE Link Probe
2. A new class `LinkMetricsSeriesInfo` is used to maintain the data of one Series configured by a neighbor. This class inherits `LinkedListEntry` and each `Neighbor` has a list of `LinkMetricsSeriesInfo`. All `LinkedListEntry`s are allocated and freed by a `Pool` in `LinkMetrics`.
3. To specify `SeriesFlags` in cli command for Forward Tracking Series, a similar approach with `LinkMetricsFlags` is used. Another character flags is used to represent `SeriesFlags`.
4. Whenever the node receives a frame (including ACK) from a neighbor, it would call `Neighbor::AggregateLinkMetrics` which goes through each `LinkMetricsSeriesInfo` entry in the neighbor's list and aggregate the data if the frame type matches the entry.
5. Two test scripts are added to test this function:
    * `v1_2_LowPower_7_2_01_ForwardTrackingSeries.py`. This case in test harness is used to test **Forward Tracking Series**. So I implemented it in this PR for test purpose. However there's some problems in the test plan.(I proposed in TESTPLAN-485) So I slightly changed the content of the case in this PR. We may not need to be very strict for the content in this PR. I would refine it once this case is discussed by the working group.
    * `v1_2_LowPower_test_forward_tracking_series.py`. This case is an extra test as a complement to the above case. It's intended to hit most new code in this PR.


edmont (commented):
Overall it looks good to me, but still need to test on hardware.

See just some minor doc related comments.

src/cli/README.md
```diff
@@ -1134,6 +1134,56 @@ Done
...
+> linkmetrics query fe80:0:0:0:3092:f334:1455:1ad2 forward 1
+Done
+> Received Link Metrics Report from: fe80:0:0:0:3092:f334:1455:1ad2
+
+ - LQI: 76 (Exponential Moving Average)
+ - Margin: 82 (dB) (Exponential Moving Average)
+ - RSSI: -18 (dBm) (Exponential Moving Average)
+```
+
+### linkmetrics mgmt \<ipaddr\> forward \<seriesid\> [ldraX][pqmr]
```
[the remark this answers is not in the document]
Irving-cl: I tried to add a space here however it's reverted by `make-pretty markdown`.

src/core/thread/link_metrics.hpp
```diff
@@ -58,6 +61,99 @@ namespace ot {
...
+     * accounted.
+     * @param[in]  aLinkMetricsFlags    A reference to flags specifying what metrics to query.
+     *
+     */
+    void Init(uint8_t aSeriesId, const SeriesFlags &aSeriesFlags, const otLinkMetrics &aLinkMetrics);
+
+    /**
+     * This method gets the PDU count.
+     *
+     * @retval  The PDU count.
```
abtink, line 103:
```suggestion
     * @returns  The PDU count.
```

src/core/thread/link_metrics.hpp
```diff
@@ -58,6 +61,99 @@ namespace ot {
...
+     *
+     */
+    uint32_t GetPduCount() const { return mPduCount; }
+
+    /**
+     * This method gets the average LQI.
+     *
+     * @retval  The average LQI.
```
abtink, line 111:
```suggestion
     * @returns  The average LQI.
```

src/core/thread/link_metrics.hpp
```diff
@@ -58,6 +61,99 @@ namespace ot {
...
+     *
+     */
+    uint8_t GetAverageLqi() const { return mLqiAverager.GetAverage(); }
+
+    /**
+     * This method gets the average RSS.
+     *
+     * @retval  The average RSS.
```
abtink, line 119:
```suggestion
     * @returns  The average RSS.
```

src/core/thread/link_metrics.hpp
```diff
@@ -85,20 +181,81 @@ class LinkMetrics : public InstanceLocator, private NonCopyable
      */
     otError LinkMetricsQuery(const Ip6::Address & aDestination,
                              uint8_t              aSeriesId,
-                             const otLinkMetrics &aLinkMetricsFlags);
+                             const otLinkMetrics *aLinkMetricsFlags);
+
+    /**
+     * This function sends an MLE Link Metrics Management Request to configure a Forward Tracking Series.
```
abtink, line 187:
```suggestion
     * This method sends an MLE Link Metrics Management Request to configure a Forward Tracking Series.
```

src/core/thread/mle.cpp
```diff
@@ -2420,6 +2421,53 @@ otError Mle::SendOrphanAnnounce(void)
...
+
+    VerifyOrExit((message = NewMleMessage()) != nullptr, error = OT_ERROR_NO_BUFS);
+    SuccessOrExit(error = AppendHeader(*message, kCommandLinkMetricsManagementResponse));
+
+    tlv.SetType(Tlv::kLinkMetricsManagement);
+    statusSubTlv.SetType(kLinkMetricsStatus);
+    statusSubTlv.SetLength(sizeof(aStatus));
+    tlv.SetLength(statusSubTlv.GetSize());
+
+    SuccessOrExit(error = message->AppendBytes(&tlv, sizeof(tlv)));
```
abtink, line 2440: We can use `Message::Append()` here and all the uses below.

Anywhere we have a pattern like `AppendByte(&someObject, sizeof(someObject))` we can instead use `Append(someObject)`.

abtink (approved):
Looks great. 👍 
Few small suggestions below.

examples/platforms/nrf528xx/nrf52840/openthread-core-nrf52840-config.h
```diff
@@ -288,6 +288,16 @@
 #define OPENTHREAD_CONFIG_CSL_SAMPLE_WINDOW 30
 #endif
 
+/**
```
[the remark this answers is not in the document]
Irving-cl: I was thinking that, since this variable would effect the memory usage, we can configure different values on different platforms. For example, on some platforms we have very limited memory and we set this smaller. However, I think for `nrf52840` and `simulation` we can just use the default value in `mle.h`. I would remove the defines in `nrf52840` and `simulation`.

include/openthread/link_metrics.h
```diff
@@ -115,6 +138,46 @@ otError otLinkMetricsQuery(otInstance *                aInstance,
...
+ * @param[in] aSeriesId          The ID of the Series to operate with.
+ * @param[in] aSeriesFlags       A pointer to the Series Flags that specifies which frames are to be accounted.
+ * @param[in] aLinkMetricsFlags  A pointer to flags specifying what metrics to query.
+ *
+ * @retval OT_ERROR_NONE          Successfully sent a Link Metrics Management Request message.
+ * @retval OT_ERROR_NO_BUFS       Insufficient buffers to generate the MLE Link Metrics Management Request message.
+ * @retval OT_ERROR_INVALID_ARGS  @p aSeriesId is not within the valid range.
+ *
+ */
+otError otLinkMetricsSendMgmtRequestForwardTrackingSeries(otInstance *                    aInstance,
```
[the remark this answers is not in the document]
Irving-cl: I would prefer to present them as two separate api to the users. The main reason is that the parameters for them are different: Forward Probing requires `Forward Series ID` and `Forward Series Flags` while Enhanced Ack  requires `Enh-ACK Flags`. In addition, I think, from a user's perspective, mgmt request can be transparent. The users only needs to know Forward Tracking Series and Enhanced Ack Probing.

Speaking of this, I would prefer renaming this api to `otLinkMetricsConfigureForwardTrackingSeries`. Thoughts?

librasungirl:
@Irving-cl , size report should be updated to reflect the size changes along with this PR.

include/openthread/link_metrics.h
```diff
@@ -115,6 +138,46 @@ otError otLinkMetricsQuery(otInstance *                aInstance,
                            otLinkMetricsReportCallback aCallback,
                            void *                      aCallbackContext);
 
+/**
+ * This function sends an MLE Link Metrics Management Request to configure a Forward Tracking Series.
+ *
+ * @param[in] aInstance          A pointer to an OpenThread instance.
+ * @param[in] aDestination       A pointer to the destination address.
+ * @param[in] aSeriesId          The ID of the Series to operate with.
+ * @param[in] aSeriesFlags       A pointer to the Series Flags that specifies which frames are to be accounted.
```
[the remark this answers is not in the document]
Irving-cl: It shouldn't be `null`. I think it would be difficult for users to use using `uint8_t` here (That was Abtin's suggestion in the previous Link Metrics PR) because users need to play with the bits then. I think `aSeriesFlags` is just similar with `aLinkMetricsFlags` and we can use it in the same way.

src/core/thread/link_metrics.cpp
```diff
@@ -44,39 +44,171 @@
...
 otError LinkMetrics::LinkMetricsQuery(const Ip6::Address & aDestination,
                                       uint8_t              aSeriesId,
-                                      const otLinkMetrics &aLinkMetricsFlags)
+                                      const otLinkMetrics *aLinkMetricsFlags)
 {
     otError                error;
     LinkMetricsTypeIdFlags typeIdFlags[kMaxTypeIdFlags];
-    uint8_t                typeIdFlagsCount = TypeIdFlagsFromLinkMetricsFlags(typeIdFlags, aLinkMetricsFlags);
+    uint8_t                typeIdFlagsCount = 0;
+
```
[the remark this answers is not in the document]
Irving-cl: Added below.

src/cli/cli.cpp
```diff
@@ -1902,14 +1902,27 @@ otError Interpreter::ProcessLeaderWeight(uint8_t aArgsLength, char *aArgs[])
...
 #if OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE
+enum : uint8_t
+{
+    kLinkMetricsStatusSuccess                   = 0,
+    kLinkMetricsStatusCannotSupportNewSeries    = 1,
+    kLinkMetricsStatusSeriesIdAlreadyRegistered = 2,
+    kLinkMetricsStatusSeriesIdNotRecognized     = 3,
+    kLinkMetricsStatusNoMatchingFramesReceived  = 4,
+    kLinkMetricsStatusOtherError                = 254,
+};
```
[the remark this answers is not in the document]
Irving-cl: Updated.

include/openthread/link_metrics.h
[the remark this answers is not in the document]
librasungirl: Add note that the pointer could not be null or maybe simply pass the object `otLinkMetricsSeriesFlags  aSeriesFlags` other than a pointer?

include/openthread/link_metrics.h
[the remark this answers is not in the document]
librasungirl: sgtm for the new name, or  maybe a bit shorter version `otLinkMetricsConfigForwardTrackingSeries`?

include/openthread/link_metrics.h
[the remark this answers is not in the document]
Irving-cl: I would prefer to add some note here because all arguments for structs in public API are passed by pointer, I think it would be better to keep consistent.

src/core/thread/link_metrics.cpp
```diff
@@ -164,83 +299,205 @@ otError LinkMetrics::AppendLinkMetricsReport(Message &aMessage, const Message &a
...
+        default:
+            break;
+        }
+
+        index += tlv.GetSize();
+    }
+
+    if (doForwardProbingRegistration)
+    {
+        aStatus = ConfigureForwardTrackingSeries(seriesId, seriesFlags, linkMetrics, aNeighbor);
```
edmont, line 347: `aStatus` is not being set when the request is malformed.

src/core/thread/mle.cpp
```diff
@@ -3689,6 +3751,50 @@ void Mle::HandleAnnounce(const Message &aMessage, const Ip6::MessageInfo &aMessa
...
+                                             Neighbor *              aNeighbor)
+{
+    otError             error = OT_ERROR_NONE;
+    otLinkMetricsStatus status;
+
+    Log(kMessageReceive, kTypeLinkMetricsManagementRequest, aMessageInfo.GetPeerAddr());
+
+    VerifyOrExit(aNeighbor != nullptr, error = OT_ERROR_INVALID_STATE);
+
+    error = Get<LinkMetrics>().HandleLinkMetricsManagementRequest(aMessage, *aNeighbor, status);
```
edmont, line 3766: We are not doing anything with this `error` here.

edmont (commented):
I'm observing an invalid value of `0x20` for the Link Metrics Status Sub-TLV after a well formed Link Metrics Management Request.

Screenshot from 2020-10-23 15-25-35

src/core/thread/link_metrics.cpp
Irving-cl: If the request is malformed, this function would exit earlier (in the before `while loop`). So if the code reaches here, the message must be well-formed. And if the request is malformed, nothing would be done further and no response would be sent so `aStatus` is not required in this case.

Irving-cl: Upd: I think this comment is actually caused by the next comment. I have updated so that when the request is malformed it would be simply discarded. So the `aStatus` is not required.

Irving-cl:
> I'm observing an invalid value of `0x20` for the Link Metrics Status Sub-TLV after a well formed Link Metrics Management Request.
> 
> Screenshot from 2020-10-23 15-25-35

@edmont 

Thanks for finding the problem. Could you provide the steps to reproduce this?

src/cli/README.md
```diff
@@ -1143,6 +1143,44 @@ Set the Thread Leader Weight.
 Done
 ```
 
+### linkmetrics mgmt \<ipaddr\> forward \<seriesid\> [ldraX][pqmr]
+
+Send a Link Metrics Management Request to configure a Forward Tracking Series.
+
+- ipaddr: Peer address.
+- seriesid: The ID of the series.
```
librasungirl, line 1151:
```suggestion
- seriesid: The Series ID.
```

src/cli/README.md
```diff
@@ -1164,6 +1202,23 @@ Done
...
+- seriesid: The Series ID.
+
+```bash
+> linkmetrics query fe80:0:0:0:3092:f334:1455:1ad2 forward 1
+Done
+> Received Link Metrics Report from: fe80:0:0:0:3092:f334:1455:1ad2
+
+ - LQI: 76 (Exponential Moving Average)
+ - Margin: 82 (dB) (Exponential Moving Average)
+ - RSSI: -18 (dBm) (Exponential Moving Average)
```
librasungirl, line 1219: If it is the report for the forward tracking series `1` configured above
`linkmetrics mgmt fe80:0:0:0:3092:f334:1455:1ad2 forward 1 dra pqmr`

why there is no sub report for `p`?

src/core/api/link_metrics_api.cpp
```diff
@@ -49,12 +49,40 @@ otError otLinkMetricsQuery(otInstance *                aInstance,
...
+
+otError otLinkMetricsConfigForwardTrackingSeries(otInstance *                      aInstance,
+                                                 const otIp6Address *              aDestination,
+                                                 uint8_t                           aSeriesId,
+                                                 const otLinkMetricsSeriesFlags *  aSeriesFlags,
+                                                 const otLinkMetrics *             aLinkMetricsFlags,
+                                                 otLinkMetricsMgmtResponseCallback aCallback,
+                                                 void *                            aCallbackContext)
+{
+    OT_ASSERT(aDestination != nullptr && aSeriesFlags != nullptr && aLinkMetricsFlags != nullptr);
```
librasungirl, line 68: `aLinkMetricsFlags` could be null if `*aSeriesFlags` is 0?

1.2 spec 4.11.3.4.4.4
> This field MUST NOT be included when Forward Series Flags='0'

The API design should conform with spec to avoid unnecessary confusion?

src/core/thread/link_metrics.cpp
```diff
@@ -44,39 +44,177 @@
...
+
+bool LinkMetricsSeriesInfo::IsFrameTypeMatch(uint8_t aFrameType) const
+{
+    bool match = false;
+
+    switch (aFrameType)
+    {
+    case kSeriesTypeLinkProbe:
+        VerifyOrExit(!mSeriesFlags.IsMacDataFlagSet()); // Ignore this when Mac Data is accounted
+        match = mSeriesFlags.IsLinkProbeFlagSet();
```
librasungirl, line 88: 1.2 spec 4.11.3.4.4.4 says
> Bit 0: Track metrics of forward MAC data frames for MLE Link Probes that contain a Link Probe TLV with a ‘Series ID’ field that matches the ‘Forward Series ID’ of this series.
 
> Bit 1: Track metrics of forward data frames for all Link Layer data frames. If this bit is selected, Bit 0 SHALL be ignored.

My understanding is 
- if `IsMacDataFlagSet()` is false,  match should not be true unless the `Serial ID` in `MLE Link Probe` matches the one in the `LinkMetricsSeriesInfo`; 
- if `IsMacDataFlagSet()` is true, any MLE Link Probe should be valid MAC Data frame, regardless of the `Series ID` inside.

However here I didn't see check on Series ID match, this code block ignores MLE link probe, not the expecting behavior to ignore `bit 0` of the series flags.

src/core/thread/link_metrics.hpp
```diff
@@ -85,20 +181,81 @@ class LinkMetrics : public InstanceLocator, private NonCopyable
      */
     otError LinkMetricsQuery(const Ip6::Address & aDestination,
                              uint8_t              aSeriesId,
-                             const otLinkMetrics &aLinkMetricsFlags);
+                             const otLinkMetrics *aLinkMetricsFlags);
+
+    /**
+     * This method sends an MLE Link Metrics Management Request to configure a Forward Tracking Series.
```
librasungirl, line 187:
```suggestion
     * This method sends an MLE Link Metrics Management Request to configure/clear a Forward Tracking Series.
```

include/openthread/link_metrics.h
```diff
@@ -115,6 +152,47 @@ otError otLinkMetricsQuery(otInstance *                aInstance,
                            otLinkMetricsReportCallback aCallback,
                            void *                      aCallbackContext);
 
+/**
+ * This function sends an MLE Link Metrics Management Request to configure a Forward Tracking Series.
```
librasungirl, line 156:
```suggestion
 * This function sends an MLE Link Metrics Management Request to configure/clear a Forward Tracking Series.
```

src/core/thread/link_metrics.cpp
```diff
@@ -44,39 +44,177 @@
...
-    Tlv           tlv;
-    uint8_t       queryId;
-    bool          hasQueryId  = false;
-    uint8_t       length      = 0;
-    uint16_t      startOffset = aMessage.GetLength();
-    uint16_t      offset;
-    uint16_t      endOffset;
-    otLinkMetrics linkMetrics;
+    otError     error = OT_ERROR_NONE;
+    Tlv         forwardProbingRegistrationSubTlv;
```
librasungirl, line 145: is it necessary? may use a pointer of type TLV to the start of `subTLVs` buffer to avoid below unnecessary memcpy below?

src/core/thread/link_metrics.cpp
```diff
@@ -44,39 +44,177 @@
...
+        seriesFlags.SetMacDataRequestFlag();
+    }
+    if (aSeriesFlags.mMacAck)
+    {
+        seriesFlags.SetMacAckFlag();
+    }
+
+    memcpy(subTlvs, &forwardProbingRegistrationSubTlv, sizeof(forwardProbingRegistrationSubTlv));
+    memcpy(subTlvs + sizeof(forwardProbingRegistrationSubTlv), &aSeriesId, sizeof(aSeriesId));
+    memcpy(subTlvs + sizeof(forwardProbingRegistrationSubTlv) + sizeof(aSeriesId), &aSeriesFlags, sizeof(aSeriesFlags));
```
librasungirl, line 183: may use a `offset` variable to iterate subTLVs and avoid `memcpy`?

src/core/thread/link_metrics.cpp
```diff
@@ -44,39 +44,177 @@
...
+otError LinkMetrics::SendLinkProbe(const Ip6::Address &aDestination, uint8_t aSeriesId, uint8_t aLength)
+{
+    otError error = OT_ERROR_NONE;
+    uint8_t buf[kLinkProbeMaxLen];
 
-    memset(&linkMetrics, 0, sizeof(linkMetrics));
+    VerifyOrExit(aLength <= LinkMetrics::kLinkProbeMaxLen && aSeriesId != 0, error = OT_ERROR_INVALID_ARGS);
+
+    error = Get<Mle::MleRouter>().SendLinkProbe(aDestination, aSeriesId, buf, aLength);
+exit:
```
librasungirl, line 201: add log for debug purpose?

src/cli/README.md
Irving-cl: The result isn't actually related with the previous example. But I can add the sub report for `p`.

src/cli/cli.cpp
```diff
@@ -1959,49 +2021,152 @@ otError Interpreter::ProcessLinkMetricsQuery(uint8_t aArgsLength, char *aArgs[])
...
 
-            case 'r':
-                linkMetrics.mRssi = 1;
+            case 'a':
+                seriesFlags.mMacAck = true;
+                break;
+
+            case 'X':
+                VerifyOrExit(arg == aArgs[3] && *(arg + 1) == '\0',
+                             error = OT_ERROR_INVALID_ARGS); // Ensure the flags only contain 'X'
```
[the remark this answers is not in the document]
Irving-cl: Both `Xldra` and `ldraX` are invalid, only `X` is valid.

```
                VerifyOrExit(arg == aArgs[3] && *(arg + 1) == '\0',
                             error = OT_ERROR_INVALID_ARGS); // Ensure the flags only contain 'X'
```
This checks if `X` is the first character and the only character in the string.

src/core/thread/link_metrics.cpp
Irving-cl: I think some compiler would complain about such practice: https://github.com/openthread/openthread/issues/5647. Besides, I think code like this is not quite readable.

src/core/thread/link_metrics.cpp
Irving-cl: Updated to add match for `SeriesID` for Link Probe.

src/cli/README.md
```diff
@@ -59,7 +59,7 @@ Done
 - [leaderdata](#leaderdata)
 - [leaderpartitionid](#leaderpartitionid)
 - [leaderweight](#leaderweight)
-- [linkmetrics](#linkmetrics-query-ipaddr-single-pqmr)
+- [linkmetrics](#linkmetrics-probe-ipaddr-seriesid-length)
```
edmont, line 62:
```suggestion
- [linkmetrics](#linkmetrics-mgmt-ipaddr-forward-seriesid-ldraxpqmr)
```

edmont (commented):
> Thanks for finding the problem. Could you provide the steps to reproduce this?

```
    def harness_join_network(self, dev, mode : str = 'rdn'):
        self.cmd(dev, 'dataset channelmask 0x7fff800')
        self.cmd(dev, 'dataset securitypolicy 3600 onrcb')
        self.cmd(dev, 'channel 12')
        self.cmd(dev, 'dataset channel 12')
        self.cmd(dev, 'panid 0xface')
        self.cmd(dev, 'dataset panid 0xface')
        self.cmd(dev, 'extpanid 000db80000000000')
        self.cmd(dev, 'dataset extpanid 000db80000000000')
        self.cmd(dev, 'networkname GRL')
        self.cmd(dev, 'dataset networkname GRL')
        self.cmd(dev, 'masterkey 00112233445566778899aabbccddeeff')
        self.cmd(dev, 'dataset masterkey 00112233445566778899aabbccddeeff')
        self.cmd(dev, 'dataset meshlocalprefix FD00:0DB8:0000:0000::')
        self.cmd(dev, 'dataset pskc 00000000000000000000000000000000')
        self.cmd(dev, 'dataset activetimestamp 1')
        self.cmd(dev, 'routerupgradethreshold 0')
        self.cmd(dev, 'mode %s' % mode)
        self.cmd(dev, 'dataset commit active')
        self.cmd(dev, 'routerselectionjitter 1')
        self.cmd(dev, 'ifconfig up')
        self.cmd(dev, 'thread start')

    def test_lm_forward(self):
        self.harness_join_network(self.dut)
        time.sleep(3)

        self.cmd(self.node, 'pollperiod 500')
        self.harness_join_network(self.node, mode='-')
        time.sleep(3)

        dut_ip = self.cmd(self.dut, 'ipaddr linklocal')

        self.cmd(self.node, 'linkmetrics mgmt %s forward %s' % (dut_ip, '1 l p'))
        time.sleep(2)

        self.cmd(self.node, 'linkmetrics query %s forward %s' % (dut_ip, '1'))
        time.sleep(2)
```

Irving-cl:
@edmont 

I have found the problem: A variable is not initialized before processing: https://github.com/openthread/openthread/commit/6ec22b04b8abeaabe64723a4ccbb1238f3a91974#diff-9efeb98d3167009731304c661cd6f5eb9aab1c6cd027f866f3d13e17c1ac8f94R659  And then in the parsing function, it would return an error. That's not detected on our simulation platform because uninitialized variable have undefined behavior. Thanks very much for finding the issue! Please check again with the latest commit. I have already verified it on dk.

src/core/thread/mle.cpp
```diff
@@ -2420,6 +2421,53 @@ otError Mle::SendOrphanAnnounce(void)
...
+
+    VerifyOrExit((message = NewMleMessage()) != nullptr, error = OT_ERROR_NO_BUFS);
+    SuccessOrExit(error = AppendHeader(*message, kCommandLinkProbe));
+
+    tlv.SetType(Tlv::kLinkProbe);
+    tlv.SetLength(sizeof(aSeriesId) + aLength);
+
+    SuccessOrExit(error = message->Append(tlv));
+    SuccessOrExit(error = message->Append(aSeriesId));
+    SuccessOrExit(error = message->Append(aBuf));
```
librasungirl, line 2463:
```suggestion
    SuccessOrExit(error = message->AppendBytes(aBuf, aLength));
```

src/core/thread/mle.cpp
```diff
@@ -3689,6 +3751,57 @@ void Mle::HandleAnnounce(const Message &aMessage, const Ip6::MessageInfo &aMessa
...
+}
+
+void Mle::HandleLinkProbe(const Message &aMessage, const Ip6::MessageInfo &aMessageInfo, Neighbor *aNeighbor)
+{
+    otError error = OT_ERROR_NONE;
+    uint8_t seriesId;
+
+    Log(kMessageReceive, kTypeLinkProbe, aMessageInfo.GetPeerAddr());
+
+    SuccessOrExit(error = Get<LinkMetrics>().HandleLinkProbe(aMessage, seriesId));
```
librasungirl, line 3796: 1.2 spec 4.4.37 Link Probe TLV
> Series ID ... Valid values are in the range 1-254

Should add check and ignore invalid Series ID?

src/core/thread/mle.hpp
```diff
@@ -1485,6 +1494,38 @@ class Mle : public InstanceLocator, private NonCopyable
...
+     */
+    otError SendLinkMetricsManagementRequest(const Ip6::Address &aDestination,
+                                             const uint8_t *     aSubTlvs,
+                                             uint8_t             aLength);
+
+    /**
+     * This method sends an MLE Link Probe message.
+     *
+     * @param[in]  aDestination  A reference to the IPv6 address of the destination.
+     * @param[in]  aSeriesId     The ID of the Series which the Probe message targets at.
```
librasungirl, line 1517:
```suggestion
     * @param[in]  aSeriesId     The Series ID [1, 254] which the Probe message targets at.
```

src/core/thread/mle.hpp
```diff
@@ -1485,6 +1494,38 @@ class Mle : public InstanceLocator, private NonCopyable
...
+     * @param[in]  aBuf          A pointer to the data payload.
+     * @param[in]  aLength       The length of the data payload in Link Probe TLV, [0, 64].
+     *
+     * @retval OT_ERROR_NONE     Successfully sent a Link Metrics Management Request.
+     * @retval OT_ERROR_NO_BUFS  Insufficient buffers to generate the MLE Link Metrics Management Request message.
```
librasungirl, line 1522: update doxygen for `OT_ERROR_INVALID_ARGS` if Series ID or aLength is invalid

include/openthread/link_metrics.h
```diff
@@ -115,6 +152,48 @@ otError otLinkMetricsQuery(otInstance *                aInstance,
...
+                                                 const otLinkMetrics *             aLinkMetricsFlags,
+                                                 otLinkMetricsMgmtResponseCallback aCallback,
+                                                 void *                            aCallbackContext);
+
+/**
+ * This function sends an MLE Link Probe message.
+ *
+ * @param[in] aInstance       A pointer to an OpenThread instance.
+ * @param[in] aDestination    A pointer to the destination address.
+ * @param[in] aSeriesId       The Series ID which the Probe message aims at.  * @param[in] aLength         The length of
```
librasungirl, line 184: add note for valid Series ID[1, 254]?

src/core/thread/link_metrics.cpp
```diff
@@ -44,39 +44,180 @@
...
+    otLogDebgMle("SendMgmtRequestForwardTrackingSeries, error:%s", otThreadErrorToString(error));
+    return error;
+}
+
+otError LinkMetrics::SendLinkProbe(const Ip6::Address &aDestination, uint8_t aSeriesId, uint8_t aLength)
+{
+    otError error = OT_ERROR_NONE;
+    uint8_t buf[kLinkProbeMaxLen];
+
+    VerifyOrExit(aLength <= LinkMetrics::kLinkProbeMaxLen && aSeriesId != 0, error = OT_ERROR_INVALID_ARGS);
```
librasungirl, line 200:
```suggestion
    VerifyOrExit(aLength <= LinkMetrics::kLinkProbeMaxLen && aSeriesId != 0  && aSeriesId != 255, error = OT_ERROR_INVALID_ARGS);
```

src/core/thread/mle_tlvs.hpp
```diff
@@ -75,37 +75,39 @@ class Tlv : public ot::Tlv
...
+        kPendingTimestamp      = 23, ///< Pending Timestamp TLV
+        kActiveDataset         = 24, ///< Active Operational Dataset TLV
+        kPendingDataset        = 25, ///< Pending Operational Dataset TLV
+        kDiscovery             = 26, ///< Thread Discovery TLV
+        kCslChannel            = 80, ///< CSL Channel TLV
+        kCslTimeout            = 85, ///< CSL Timeout TLV
+        kLinkMetricsQuery      = 87, ///< Link Metrics Query TLV
+        kLinkMetricsManagement = 88, ///< Link Metrics Management TLV
+        kLinkMetricsReport     = 89, ///< Link Metrics Report TLV
+        kLinkProbe             = 90, /// Link Probe TLV
```
librasungirl, line 110:
```suggestion
        kLinkProbe             = 90, ///< Link Probe TLV
```

src/core/thread/topology.hpp
```diff
@@ -589,6 +595,57 @@ class Neighbor : public InstanceLocatorInit
...
+
+    /**
+     * This method removes a specific LinkMetricsSeriesInfo by Series ID.
+     *
+     * @param[in] aSeriesId    A reference to the Series ID.
+     *
+     * @returns The pointer to the LinkMetricsSeriesInfo. `nullptr` if not found.
+     *
+     */
+    LinkMetricsSeriesInfo *RemoveForwardTrackingSeriesInfo(const uint8_t &aSeriesId);
```
librasungirl, line 639:
```suggestion
    LinkMetricsSeriesInfo *RemoveForwardTrackingSeriesInfo(const uint8_t aSeriesId);
```

src/core/thread/topology.hpp
```diff
@@ -589,6 +595,57 @@ class Neighbor : public InstanceLocatorInit
...
+     *
+     * @returns The pointer to the LinkMetricsSeriesInfo. `nullptr` if not found.
+     *
+     */
+    LinkMetricsSeriesInfo *GetForwardTrackingSeriesInfo(const uint8_t &aSeriesId);
+
+    /**
+     * This method removes a specific LinkMetricsSeriesInfo by Series ID.
+     *
+     * @param[in] aSeriesId    A reference to the Series ID.
```
librasungirl, line 634:
```suggestion
     * @param[in] aSeriesId    The Series ID to remove.
```

src/core/thread/topology.hpp
```diff
@@ -629,6 +686,10 @@ class Neighbor : public InstanceLocatorInit
 #endif
     uint8_t         mVersion;  ///< The MLE version
     LinkQualityInfo mLinkInfo; ///< Link quality info (contains average RSS, link margin and link quality)
+#if OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE
+    LinkedList<LinkMetricsSeriesInfo> mLinkMetricsSeriesInfoList; ///< A list of Link Metrics Forward Tracking Series
```
librasungirl, line 690: When a neighbor is removed, the forward tracking series configured by it should be cleared?

tests/scripts/thread-cert/v1_2_LowPower_7_2_01_ForwardTrackingSeries.py
```diff
@@ -0,0 +1,340 @@
...
+        self.simulator.go(5)
+
+        # Step 4 - SED_1 requests a Forward Tracking Series by sending a Link Metrics Management Request
+        # Forward Series Flags = 0x04:
+        # - Bit 2 = 1, MAC Data Request
+        # - Bits 0,1 & 3-7 = 0
+        # Concatenation of Link Metric Type ID Flags = 0x00:
+        # - Item1: (0)(1)(000)(000) = 0x40
+        # -- E = 0
+        # -- L = 1
```
librasungirl, line 104: does it conflict with Test Plan r08?

It seems that for `L 0/1` we discussed, Test Plan indicates the value of PDU (Metrics Enum 0) is of 1 byte (Step 4 ~8) if L bit is 0, and it is 4 bytes. (Step 11 ~ 14) if L bit is 1;  however OT would always report 4 bytes PDU, regardless of `L` bit. I am afraid existing OT implementation may fail this test case.

Thoughts?

Irving-cl: The test case is problematic as I mentioned in PR description: https://github.com/openthread/openthread/pull/5608#issue-498391452. They haven't completed Link Metrics cases.

tests/scripts/thread-cert/v1_2_LowPower_7_2_01_ForwardTrackingSeries.py
```diff
@@ -0,0 +1,340 @@
...
+        # Step 21 - SSED_1 clears the Forward Series Link Metrics
+        # Forward Series Flags = 0x00:
+        # - Bits 0-7 = 0
+        # Concatenation of Link Metric Type ID Flags = NULL
+        self.nodes[SSED_1].link_metrics_mgmt_req_forward_tracking_series(leader_addr, SERIES_ID_2, 'X', '')
+
+        self.simulator.go(5)
+
+        # Step 23 - SSED_1 sends an MLE Data Request to retrieve aggregated Forward Series Results
+        self.nodes[SSED_1].link_metrics_query_forward_tracking_series(leader_addr, SERIES_ID_2)
```
librasungirl, line 161: If following Test Plan exactly (`Query ID = [same value as set in step 9]`), here should use `SERIES_ID`.
Or maybe the Series/Query ID in step 19/21/23 in TP be updated to the same one as Step 16?

src/core/thread/link_metrics.cpp
```diff
@@ -44,39 +44,180 @@
...
+    memcpy(subTlvs, &forwardProbingRegistrationSubTlv, sizeof(forwardProbingRegistrationSubTlv));
+    memcpy(subTlvs + sizeof(forwardProbingRegistrationSubTlv), &aSeriesId, sizeof(aSeriesId));
+    memcpy(subTlvs + sizeof(forwardProbingRegistrationSubTlv) + sizeof(aSeriesId), &aSeriesFlags, sizeof(aSeriesFlags));
+
+    error = Get<Mle::MleRouter>().SendLinkMetricsManagementRequest(aDestination, subTlvs,
+                                                                   forwardProbingRegistrationSubTlv.GetSize());
 
-    memset(&linkMetrics, 0, sizeof(linkMetrics));
+exit:
+    otLogDebgMle("SendMgmtRequestForwardTrackingSeries, error:%s", otThreadErrorToString(error));
```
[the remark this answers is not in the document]
Irving-cl: > should here be match = mSeriesFlags.IsMacDataFlagSet() || mSeriesFlags.IsLinkProbeFlagSet()?

If `IsMacDataFlagSet` is `true`, this message would be accounted as a mac data frame. `VerifyOrExit(!mSeriesFlags.IsMacDataFlagSet());` is used to avoid duplication. 

For example, when `IsMacDataFlagSet` is true and we receive a Link Probe message. The Link Probe message would be accounted first at `AggregateLinkMetrics` in `Mac::HandleReceivedFrame`. And then accounted again at `AggregateLinkMetrics` in `Mle::HandleLinkProbe`. I did this because Link Probe are not link layer messages so I cannot directly count it in `Mac`. If `IsMacDataFlagSet` is true, Link Probe would be accounted in Mac layer and won't be accounted in Mle. If `IsMacDataFlagSet` is false and `IsLinkProbeFlagSet()` is true, Link Probe would be accounted in Mle.

tests/scripts/thread-cert/v1_2_LowPower_7_2_01_ForwardTrackingSeries.py
```diff
@@ -0,0 +1,340 @@
...
+            .must_next()
+
+        # Step 4 - SED_1 requests a Forward Series Link Metrics by sending a Link Metrics Management Request
+        # Forward Series Flags = 0x04:
+        # - Bit 2 = 1, MAC Data Request
+        # - Bits 0,1 & 3-7 = 0
+        # Concatenation of Link Metric Type ID Flags = 0x40:
+        # - Item1: (0)(1)(000)(000) = 0x40
+        # -- E = 0
+        # -- L = 1
```
librasungirl, line 189: `step 4` in TP r08  doesn't mention `L`

src/core/thread/link_metrics.cpp
```diff
@@ -299,52 +579,166 @@ otError LinkMetrics::SendLinkMetricsQuery(const Ip6::Address &          aDestina
...
+
+    return count;
+}
+
+otError LinkMetrics::ReadTypeIdFlagsFromMessage(const Message &aMessage,
+                                                uint8_t        aStartPos,
+                                                uint8_t        aEndPos,
+                                                otLinkMetrics &aLinkMetrics)
+{
+    otError error = OT_ERROR_NONE;
```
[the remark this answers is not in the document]
Irving-cl: Has updated in next line.

tests/scripts/thread-cert/v1_2_LowPower_7_2_01_ForwardTrackingSeries.py
```diff
@@ -0,0 +1,340 @@
...
+        # - Link Metrics Query TLV
+        # -- Link Metrics Query ID Sub-TLV
+        # --- Query ID: (ID set in step 16)
+        # - Verify that the Link Metrics Query Options Sub-TLV is NOT present
+        pkts.filter_wpan_src64(SSED_1) \
+            .filter_wpan_dst64(LEADER) \
+            .filter_mle_cmd(consts.MLE_DATA_REQUEST) \
+            .filter(lambda p: TlvType.LINK_METRICS_QUERY in p.mle.tlv.type) \
+            .filter(lambda p: p.mle.tlv.query_id == SERIES_ID_2) \
+            .must_next()
```
librasungirl, line 275: `# - Verify that the Link Metrics Query Options Sub-TLV is NOT present` is not verified

tests/scripts/thread-cert/v1_2_LowPower_7_2_01_ForwardTrackingSeries.py
```diff
@@ -0,0 +1,340 @@
...
+
+        # Step 16 - SSED_1 requests a Forward Series Link Metrics by sending a Link Metrics Management Request
+        # Forward Series Flags = 0x02:
+        # - Bit 1 = 1, (all Link Layer Data Frames)
+        # - Bits 0, 2-7 = 0
+        # Concatenation of Link Metric Type ID Flags = 0x0a:
+        # - Item1: (0)(0)(001)(010) = 0x0a
+        # -- E = 0
+        # -- L = 0
```
librasungirl, line 245: not checked

tests/scripts/thread-cert/v1_2_LowPower_7_2_01_ForwardTrackingSeries.py
```diff
@@ -0,0 +1,340 @@
...
+            .filter(lambda p: p.mle.tlv.query_id == SERIES_ID_2) \
+            .must_next()
+
+        # Step 20 - Leader MUST reply to SSED_1 with an MLE Data Response with the following:
+        # - Link Metrics Report TLV
+        # -- Link Metrics Report Sub-TLV
+        # --- Metric Type ID Flags
+        # ---- Type / Average Enum = 1 (Exponential Moving Average)
+        # ---- Metric Enum = 2 (Link Margin)
+        # --- Value (1 bytes)
```
librasungirl, line 283: size not checked

src/core/thread/topology.hpp
Irving-cl: There is a reason for using reference. The underlying template method `RemoveMatching` (linked_list.hpp) requires to use const reference.

src/core/thread/topology.hpp
Irving-cl: It should. Where are the calls should we clear it? I'll update it.

src/core/thread/link_metrics.cpp
Irving-cl: @librasungirl

If `IsMacDataFlagSet` is true, this message would be accounted as a mac data frame. `VerifyOrExit(!mSeriesFlags.IsMacDataFlagSet());` is used to avoid duplication.

For example, when `IsMacDataFlagSet` is `true` and we receive a Link Probe message. The Link Probe message would be accounted first at `AggregateLinkMetrics` in `Mac::HandleReceivedFrame` because it is itself a link layer data frame. And then accounted again at `AggregateLinkMetrics` in `Mle::HandleLinkProbe`. I did this because Link Probe are not link layer messages so I cannot directly count it in Mac. If `IsMacDataFlagSet` is true, Link Probe would be accounted in Mac layer and won't be accounted in Mle. If `IsMacDataFlagSet` is false and `IsLinkProbeFlagSet()` is `true`, Link Probe would be accounted in Mle.

src/core/thread/mle.cpp
Irving-cl: Added check in `LinkMetrics::HandleLinkProbe`.

src/core/thread/link_metrics.cpp
librasungirl: yes, you are right.

src/core/thread/mle.cpp
```diff
@@ -2420,6 +2421,53 @@ otError Mle::SendOrphanAnnounce(void)
     return error;
 }
 
+#if OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE
+otError Mle::SendLinkMetricsManagementResponse(const Ip6::Address &aDestination, otLinkMetricsStatus aStatus)
+{
+    otError  error = OT_ERROR_NONE;
+    Message *message;
+    Tlv      tlv;
+    ot::Tlv  statusSubTlv;
```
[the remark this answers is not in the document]
Irving-cl: It is necessary. Without `ot::`, `Tlv` would be regarded as `Mle::Tlv` whose `SetType` method takes an `Mle::Type as parameter. However, `kLinkMetricsStatus` is defined in `Mle` namespace and its type doesn't match `Mle::Type`. Removing `ot::` would cause compiling error.

src/core/thread/topology.hpp
Irving-cl: Updated.

include/openthread/link_metrics.h
```diff
@@ -80,17 +80,54 @@ typedef struct otLinkMetricsValues
...
+/**
+ * This structure represents which frames are accounted in a Forward Tracking Series.
+ *
+ */
+typedef struct otLinkMetricsSeriesFlags
+{
+    bool mLinkProbe : 1;      ///< MLE Link Probe.
+    bool mMacData : 1;        ///< MAC Data frame.
+    bool mMacDataRequest : 1; ///< MAC Data Request.
+    bool mMacAck : 1;         ///< Mac Ack.
```
edmont, line 92:
```suggestion
    bool mMacAck : 1;         ///< MAC Ack.
```

edmont (approved):
@Irving-cl thanks for fixing the issue. Now it looks good to me.

src/core/thread/link_metrics.cpp
Irving-cl: Have updated.

include/openthread/link_metrics.h
```diff
@@ -115,6 +152,48 @@ otError otLinkMetricsQuery(otInstance *                aInstance,
...
+/**
+ * This function sends an MLE Link Metrics Management Request to configure/clear a Forward Tracking Series.
+ *
+ * @param[in] aInstance          A pointer to an OpenThread instance.
+ * @param[in] aDestination       A pointer to the destination address.
+ * @param[in] aSeriesId          The Series ID to operate with.
+ * @param[in] aSeriesFlags       A pointer to the Series Flags that specifies which frames are to be accounted. The
+ * pointer MUST NOT be `nullptr`.
+ * @param[in] aLinkMetricsFlags  A pointer to flags specifying what metrics to query. Should be `NULL` when
+ * `aSeriesFlags` is `NULL`.
```
librasungirl, line 164: conflict - whether or not `aSeriesFlags` could be NULL? 
From the context, `aLinkMetricsFlags` should be `nullptr` when `@p aSeriesFlags` is a pointer to a Series Flags with value 0.

src/core/thread/link_metrics.cpp
```diff
@@ -164,92 +303,237 @@ otError LinkMetrics::AppendLinkMetricsReport(Message &aMessage, const Message &a
...
+    while (index < length)
+    {
+        SuccessOrExit(aMessage.Read(offset + index, tlv));
+
+        switch (tlv.GetType())
+        {
+        case kLinkMetricsStatus:
+            VerifyOrExit(!hasStatus, error = OT_ERROR_PARSE);
+            hasStatus = true;
+            SuccessOrExit(aMessage.Read(offset + index + sizeof(tlv), status));
```
librasungirl, line 382:
```suggestion
            SuccessOrExit(aMessage.Read(offset + index + sizeof(tlv), status));
            hasStatus = true;
```

src/core/thread/link_metrics.cpp
```diff
@@ -164,92 +303,237 @@ otError LinkMetrics::AppendLinkMetricsReport(Message &aMessage, const Message &a
...
+    SuccessOrExit(error = Tlv::FindTlvValueOffset(aMessage, Mle::Tlv::Type::kLinkMetricsManagement, offset, length));
+
+    while (index < length)
+    {
+        SuccessOrExit(aMessage.Read(offset + index, tlv));
+
+        switch (tlv.GetType())
+        {
+        case kLinkMetricsStatus:
+            VerifyOrExit(!hasStatus, error = OT_ERROR_PARSE);
```
librasungirl, line 380: tlv length validation check before read?

src/core/thread/link_metrics.cpp
```diff
@@ -164,92 +303,237 @@ otError LinkMetrics::AppendLinkMetricsReport(Message &aMessage, const Message &a
...
-            metricsValues.mLinkMarginValue =
-                metricsRawValue * 130 / 255; // Reverse operation for linear scale, map from [0, 255] to [0, 130]
-            pos += sizeof(uint8_t);
-            otLogDebgMle(" - Margin: %d (dB) (Exponential Moving Average)", metricsValues.mLinkMarginValue);
+        case kLinkMetricsStatus:
+            VerifyOrExit(!hasStatus && !hasReport,
+                         error = OT_ERROR_DROP); // There should be either: one Status TLV or some Report-Sub TLVs
+            hasStatus = true;
+            VerifyOrExit(tlv.GetLength() == sizeof(status), error = OT_ERROR_PARSE);
+            SuccessOrExit(aMessage.Read(pos, status));
```
librasungirl, line 437:
```suggestion
            VerifyOrExit(tlv.GetLength() == sizeof(status), error = OT_ERROR_PARSE);
            SuccessOrExit(aMessage.Read(pos, status));
            hasStatus = true;
```

src/core/thread/link_metrics.cpp
```diff
@@ -164,92 +303,237 @@ otError LinkMetrics::AppendLinkMetricsReport(Message &aMessage, const Message &a
...
 }
 
+otError LinkMetrics::HandleLinkProbe(const Message &aMessage, uint8_t &aSeriesId)
+{
+    otError  error = OT_ERROR_NONE;
+    uint16_t offset;
+    uint16_t length;
+
+    SuccessOrExit(error = Tlv::FindTlvValueOffset(aMessage, Mle::Tlv::Type::kLinkProbe, offset, length));
+    SuccessOrExit(error = aMessage.Read(offset, aSeriesId));
```
librasungirl, line 517: tlv validation check before read?

librasungirl (approved):
LGTM with some more minor comments.

include/openthread/link_metrics.h
Irving-cl: Sorry I made a mistake. It shouldn't be `NULL` anytime. The comment actually means `aSeriesFlags` is `0`. I agree with your previous comment to pass `otLinkMetricsSeriesFlags` in this method instead of a pointer so that it could not be `NULL`.

include/openthread/link_metrics.h
```diff
@@ -115,6 +152,48 @@ otError otLinkMetricsQuery(otInstance *                aInstance,
...
                            void *                      aCallbackContext);
 
+/**
+ * This function sends an MLE Link Metrics Management Request to configure/clear a Forward Tracking Series.
+ *
+ * @param[in] aInstance          A pointer to an OpenThread instance.
+ * @param[in] aDestination       A pointer to the destination address.
+ * @param[in] aSeriesId          The Series ID to operate with.
+ * @param[in] aSeriesFlags       The Series Flags that specifies which frames are to be accounted.
+ *
```
jwhui, line 162: nit: remove blank line?

include/openthread/link_metrics.h
```diff
@@ -115,6 +152,48 @@ otError otLinkMetricsQuery(otInstance *                aInstance,
...
+ * @retval OT_ERROR_INVALID_ARGS  @p aSeriesId is not within the valid range.
+ *
+ */
+otError otLinkMetricsConfigForwardTrackingSeries(otInstance *                      aInstance,
+                                                 const otIp6Address *              aDestination,
+                                                 uint8_t                           aSeriesId,
+                                                 otLinkMetricsSeriesFlags          aSeriesFlags,
+                                                 const otLinkMetrics *             aLinkMetricsFlags,
+                                                 otLinkMetricsMgmtResponseCallback aCallback,
+                                                 void *                            aCallbackContext);
```
jwhui, line 177: nit: missing docs

src/cli/README.md
```diff
@@ -1143,6 +1143,44 @@ Set the Thread Leader Weight.
...
+Send a Link Metrics Management Request to configure a Forward Tracking Series.
+
+- ipaddr: Peer address.
+- seriesid: The Series ID.
+- ldraX: This specifies which frames are to be accounted.
+- l: MLE Link Probe.
+- d: MAC Data.
+- r: MAC Data Request.
+- a: MAC Ack.
+- X: This represents none of the above flags, i.e., to stop accounting and remove the series. This can only be used without any other flags.
```
jwhui, line 1157: nit: indent?

src/cli/README.md
```diff
@@ -1143,6 +1143,44 @@ Set the Thread Leader Weight.
...
+- pqmr: This specifies what metrics to query.
+- p: Layer 2 Number of PDUs received.
+- q: Layer 2 LQI.
+- m: Link Margin.
+- r: RSSI.
```
jwhui, line 1162: nit: indent?

src/cli/cli.cpp
```diff
@@ -1918,24 +1921,75 @@ void Interpreter::HandleLinkMetricsReport(const otIp6Address *aAddress, const ot
...
+        OutputLine("Series ID not recognized");
+        break;
+    case OT_LINK_METRICS_STATUS_NO_MATCHING_FRAMES_RECEIVED:
+        OutputLine("No matching series ID");
+        break;
+    case OT_LINK_METRICS_STATUS_OTHER_ERROR:
+        OutputLine("Other error");
+        break;
+    default:
+        OutputLine("Unknown error");
```
jwhui, line 1991: nit: does having a single `OutputLine()` call after the switch statement reduce code size?

src/core/thread/link_metrics.cpp
```diff
@@ -44,39 +44,181 @@
...
+        sizeof(Tlv) + sizeof(uint8_t) * 2; // Forward Probing Registration sub-TLV header + SeriesId + SeriesFlags
+    uint8_t      subTlvs[sizeof(Tlv) + sizeof(uint8_t) * 2 + sizeof(LinkMetricsTypeIdFlags) * kMaxTypeIdFlags];
+    Tlv *        forwardProbingRegistrationSubTlv = reinterpret_cast<Tlv *>(subTlvs);
+    SeriesFlags *seriesFlags = reinterpret_cast<SeriesFlags *>(subTlvs + sizeof(Tlv) + sizeof(aSeriesId));
+    uint8_t      typeIdFlagsCount =
+        aLinkMetricsFlags == nullptr
+            ? 0
+            : TypeIdFlagsFromLinkMetricsFlags(reinterpret_cast<LinkMetricsTypeIdFlags *>(subTlvs + typeIdFlagsOffset),
+                                              *aLinkMetricsFlags); // Directly transform `aLinkMetricsFlags` into
+                                                                   // LinkMetricsTypeIdFlags and put them into `subTlvs`
```
jwhui, line 155: nit: this declaration block is hard to read - can we clean this up?

src/core/thread/link_metrics.hpp
```diff
@@ -58,8 +61,111 @@ namespace ot {
...
+class LinkMetricsSeriesInfo : public LinkedListEntry<LinkMetricsSeriesInfo>
+{
+    friend class LinkedList<LinkMetricsSeriesInfo>;
+    friend class LinkedListEntry<LinkMetricsSeriesInfo>;
+
+public:
+    enum
+    {
+        kSeriesTypeLinkProbe =
+            0, ///< This represents Link Probe when filtering frames to be accounted using Series Flag. There's
```
jwhui, line 84: nit: put this doc string above the enum

src/core/thread/link_metrics.hpp
```diff
@@ -58,8 +61,111 @@ namespace ot {
...
+               ///< already `kFcfFrameData`, `kFcfFrameAck` and `kFcfFrameMacCmd`. This item is added so that we can
+               ///< filter a Link Probe for series in the same way as other frames.
+    };
+
+    /**
+     * This method initializes the object.
+     *
+     * @param[in]  aSeriesId            The Series ID.
+     * @param[in]  aSeriesFlags         A reference to the Series Flags which specify what types of frames are to be
+     * accounted.
```
jwhui, line 94: nit: alignment

src/core/thread/link_metrics.hpp
```diff
@@ -85,20 +191,81 @@ class LinkMetrics : public InstanceLocator, private NonCopyable
...
-                             const otLinkMetrics &aLinkMetricsFlags);
+                             const otLinkMetrics *aLinkMetricsFlags);
+
+    /**
+     * This method sends an MLE Link Metrics Management Request to configure/clear a Forward Tracking Series.
+     *
+     * @param[in] aDestination       A reference to the IPv6 address of the destination.
+     * @param[in] aSeriesId          The Series ID.
+     * @param[in] aSeriesFlags       A reference to the Series Flags which specify what types of frames are to be
+     * accounted.
```
jwhui, line 202: nit: alignment

src/core/thread/link_metrics.hpp
```diff
@@ -114,43 +281,89 @@ class LinkMetrics : public InstanceLocator, private NonCopyable
...
      * @param[in]  aCallbackContext  A pointer to application-specific context.
      *
      */
     void SetLinkMetricsReportCallback(otLinkMetricsReportCallback aCallback, void *aCallbackContext);
 
+    /**
+     * This method registers a callback to handle Link Metrics Management Response received.
+     *
+     * @param[in]  aCallback         A pointer to a function that is called when a Link Metrics Management Response is
+     * received.
```
jwhui, line 309: nit: alignment

src/core/thread/link_metrics.hpp
```diff
@@ -114,43 +281,89 @@ class LinkMetrics : public InstanceLocator, private NonCopyable
...
+
 private:
     enum
     {
         kMaxTypeIdFlags = 4,
 
         kTypeIdFlagPdu =
-            0x40, ///< 0x0_1_000_000 -> 0x40 ==> L bit set, type = 0 (count/summation), metric-enum = 0 (PDU rxed)
-        kTypeIdFlagLqi = 0x09, ///< 0x0_0_001_001 -> 0x00 ==> L bit not set, type = 1 (exp ave), metric-enum = 1 (LQI)
+            0x40, ///< 0x0_1_000_000 -> 0x40 ==> L bit set, type = 0 (count/summation), metric-enum = 0 (PDU rxed).
```
jwhui, line 321: nit: place doc string above enum to avoid wrap?

src/cli/cli.cpp
Irving-cl: Updated to use a `LinkMetricsStatusToStr` method to get static string and `OutputLine` with it

jwhui (approved):

PR #5614, opened 2020-10-07
[thread-cert] enhance case 9.2.4 to avoid reverifying coap retransmission packets

FiveDimensions:
replace must_verify with filter to avoid possible verification of coap retransmission packets which could cause filter incorrect packets.


simonlingoogle (approved):
LGTM :+1:

jwhui:
The code coverage check is failing. @simonlingoogle @Moandor-y , thoughts?

Issue #5554, opened 2020-09-20
"PRIx64" compile error on Ubuntu 20


nealjack:
**Describe the bug** 
Compiling openthread libraries and examples fails with the following error:
```
/openthread/examples/../src/cli/cli_joiner.cpp: In member function 'otError ot::Cli::Joiner::ProcessDiscerner(uint8_t, char**)':
/openthread/examples/../src/cli/cli_joiner.cpp:76:38: error: expected ')' before 'PRIx64'
   76 |         mInterpreter.OutputLine("0x%" PRIx64 "/%u", discerner->mValue, discerner->mLength);
      |                                ~     ^~~~~~~
      |                                      )
  CXX      libopenthread_cli_ftd_a-cli_network_data.o
```

It appears to be an issue with the PRIx64 macro and inttypes.h/cinttypes. Replacing the define with `<cinttypes>`, as well as defining the macro `__STDC_FORMAT_MACROS` as mentioned [here ](https://github.com/Microsoft/rDSN/issues/197) did not fix this issue.

**To Reproduce**

1. Git commit id: **0d8a99b**
2. IEEE 802.15.4 hardware platform: **NRF52840**
3. Build steps:
I am building on a newly created Ubuntu 20 VM, using these commands:
```
./script/bootstrap
./bootstrap
make -f examples/Makefile-nrf52840 BORDER_AGENT=1 BORDER_ROUTER=1 COAP=1 COMMISSIONER=1 DISABLE_BUILTIN_MBEDTLS=1 DNS_CLIENT=1 DIAGNOSTIC=1 EXTERNAL_HEAP=1 JOINER=1 LINK_RAW=1 MAC_FILTER=1 MTD_NETDIAG=1 SERVICE=1 UDP_FORWARD=1 ECDSA=1 SNTP_CLIENT=1 COAPS=1 DHCP6_SERVER=1 DHCP6_CLIENT=1
```
4. Network topology: N/A

**Expected behavior** 
I would expect the build to complete successfully.

bukepo:
What compiler and version are you using?

nealjack:
I am using gcc/g++:
```
$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (15:9-2019-q4-0ubuntu1) 9.2.1 20191025 (release) [ARM/arm-9-branch revision 277599]

$ arm-none-eabi-g++ --version
[...]

PR #5615, opened 2020-10-07
[cli] replace 'PRIx64' use

abtink:
---
This should help with https://github.com/openthread/openthread/issues/5554.

In earlier projects, we faced similar problems with `PRIx64` use with some embedded toolchains. 
In general, I would suggest we try to avoid using it if/when we can.


Irving-cl (approved):
LGTM👍

jwhui (approved):

PR #5624, opened 2020-10-09
[debug] prefer user provided assert to system assert

gjc13:
This allows users to enable OT_ASSERT in release builds on posix systems.

### DIFF
--- a/src/posix/platform/misc.cpp
+++ b/src/posix/platform/misc.cpp
@@ -76,6 +76,14 @@ otError otPlatSetMcuPowerState(otInstance *aInstance, otPlatMcuPowerState aState
     return error;
 }
 
+void otPlatAssertFail(const char *aFilename, int aLineNumber)
+{
+    otLogCritPlat("assert failed at %s:%d", aFilename, aLineNumber);
+    // For debug build, use assert to genreate a core dump
+    assert(false);
+    exit(1);
+}
+
 otPlatMcuPowerState otPlatGetMcuPowerState(otInstance *aInstance)
 {
     OT_UNUSED_VARIABLE(aInstance);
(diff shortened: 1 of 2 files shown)


abtink (approved):
LGTM. 👍

src/posix/platform/misc.cpp
```diff
@@ -76,6 +76,12 @@ otError otPlatSetMcuPowerState(otInstance *aInstance, otPlatMcuPowerState aState
     return error;
 }
 
+void otPlatAssertFail(const char *aFilename, int aLineNumber)
+{
+    otLogCritPlat("assert failed at %s:%d", aFilename, aLineNumber);
+    exit(1);
```
simonlingoogle, line 82: `exit(1)` won't dump core. Can we use `assert(false)` here?

wgtdkp: @simonlingoogle I believe this PR is to enable assertion in release build (assertion is disabled by `NDEBUG`), so `assert(false)` is not going to work.

wgtdkp (approved):
LGTM 👍

bukepo (approved):
LGTM:+1:

Irving-cl (approved):
LGTM👍

src/posix/platform/misc.cpp
gjc13: Use `abort()` which will generate a core dump instead. Thanks for pointing out.

simonlingoogle: We can use `assert(false)` before `exit(1)`. 
In release build with `NDEBUG`, `assert(false)` will be ignored and exit(1) is effective. 
In debug build, `assert(false)` generates the core dump. 

Actually I am not quite clear about the intention of this change. In what case will this `otPlatAssertFail` be better than the default `assert` behavior?

gjc13: Done

simonlingoogle (approved):
LGTM :+1:

PR #5626, opened 2020-10-09
[tests] init joiner with different masterkey before start

librasungirl:
For 8.x.x, there is validation that joiner has the same keys as the commissioner after commissioning, it should be configured with a different master key before start commissioning.


wgtdkp (approved):
LGTM :+1:

simonlingoogle (approved):
LGTM :+1:

FiveDimensions (approved):

jwhui (approved):

PR #5627, opened 2020-10-09
[harness-automation] add more test info options for gencsv

FiveDimensions:


tools/harness-automation/gencsv.py
```diff
@@ -15,14 +38,101 @@ def get_key(item):
...
+
+        o.write('Test Case,DUT Role,Status,Logs,Started,Stopped\n')
+        if test_info is not None:
+            o.write('%s' % test_info)
+        for k, v in sorted(result.items(), key=get_key):
+            o.write('%s,%s,%s,%s,%s,%s\n' % (
+                get_case_id(k),
+                get_case_role(k),
+                (v['passed'] and 'Pass') or 'Fail',
+                (v['error'] or '').replace('\n', ' '),
```
priyankaCh22, line 129: In case of error: this line writes error info into log column

tools/harness-automation/gencsv.py
```diff
@@ -15,14 +38,101 @@ def get_key(item):
...
+                v['started'],
```
priyankaCh22, line 130: Here we should add a column for "Reruns" given by "fails" in result.json

tools/harness-automation/gencsv.py
```diff
@@ -15,14 +38,101 @@ def get_key(item):
...
+                v['stopped'],
```
priyankaCh22, line 131: We don't need test start stop test time so we can drop these entries from the result.csv

PR #5629, opened 2020-10-10
[posix] map address with rtnetlink API on Linux

bukepo:
* no DAD for anycast address.
* add anycast address as non-preferred address.
* update address callback to provide more address info.

### DIFF
--- a/include/openthread/ip6.h
+++ b/include/openthread/ip6.h
@@ -416,18 +416,28 @@ typedef void (*otIp6ReceiveCallback)(otMessage *aMessage, void *aContext);
 void otIp6SetReceiveCallback(otInstance *aInstance, otIp6ReceiveCallback aCallback, void *aCallbackContext);
 
 /**
+ * @struct otIp6AddressInfo
+ *
+ * This structure represents IPv6 address information.
+ *
+ */
+typedef struct otIp6AddressInfo
+{
+    const otIp6Address *mAddress;       ///< A pointer to the IPv6 address.
+    uint8_t             mPrefixLength;  ///< The prefix length of mAddress if it is a unicast address.
+    uint8_t             mScope : 4;     ///< The scope of this address.
+    bool                mIsAnycast : 1; ///< Whether this is an anycast address.
+} otIp6AddressInfo;
+
+/**
  * This function pointer is called when an internal IPv6 address is added or removed.
  *
- * @param[in]   aAddress            A pointer to the IPv6 address.
- * @param[in]   aPrefixLength       The prefix length if @p aAddress is unicast address, and 128 for multicast address.
+ * @param[in]   aAddressInfo        A pointer to the IPv6 address information.
  * @param[in]   aIsAdded            TRUE if the @p aAddress was added, FALSE if @p aAddress was removed.
  * @param[in]   aContext            A pointer to application-specific context.
  *
  */
-typedef void (*otIp6AddressCallback)(const otIp6Address *aAddress,
-                                     uint8_t             aPrefixLength,
-                                     bool                aIsAdded,
-                                     void *              aContext);
+typedef void (*otIp6AddressCallback)(const otIp6AddressInfo *aAddressInfo, bool aIsAdded, void *aContext);
 
 /**
  * This function registers a callback to notify internal IPv6 address changes.
--- a/src/core/net/netif.cpp
+++ b/src/core/net/netif.cpp
@@ -42,6 +42,30 @@
 namespace ot {
 namespace Ip6 {
 
+class AddressInfo : public otIp6AddressInfo
+{
+public:
+    AddressInfo(const NetifUnicastAddress &aAddress, Instance &aInstance)
+    {
+        mAddress      = &aAddress.mAddress;
+        mPrefixLength = aAddress.mPrefixLength;
+        mScope        = aAddress.GetScope();
+        mIsAnycast    = aInstance.Get<Mle::MleRouter>().IsAnycastLocator(aAddress.GetAddress());
+    }
+
+    AddressInfo(const NetifMulticastAddress &aAddress)
+    {
+        mAddress      = &aAddress.GetAddress();
+        mPrefixLength = kMulticastPrefixLength;
+        mScope        = aAddress.GetAddress().GetScope();
+        mIsAnycast    = false;
+    }
+
+private:
+    static constexpr uint8_t kMulticastPrefixLength =
+        128; ///< Multicast prefix length used to notify internal address changes.
+};
+
 /*
  * Certain fixed multicast addresses are defined as a set of chained (linked-list) constant `otNetifMulticastAddress`
  * entries:
@@ -128,7 +152,10 @@ void Netif::SubscribeAllNodesMulticast(void)
 
     for (const NetifMulticastAddress *entry = &linkLocalAllNodesAddress; entry; entry = entry->GetNext())
     {
-        mAddressCallback(&entry->GetAddress(), kMulticastPrefixLength, /* IsAdded */ true, mAddressCallbackContext);
+        AddressInfo addressInfo(*entry);
+
+        mAddressCallback(&addressInfo,
+                         /* IsAdded */ true, mAddressCallbackContext);
     }
 
 exit:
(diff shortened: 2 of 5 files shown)


simonlingoogle (commented):
This PR looks good.

But I think we should determine if we wan't to add RLOC and ALOC to the Host wpan interface. 
Maybe we can just hide it from Host and make RLOC and ALOC only seen by the OpenThread stack.

Thoughts? @librasungirl @bukepo @jwhui

src/posix/platform/netif.cpp
```diff
@@ -298,36 +325,83 @@ static uint8_t NetmaskToPrefixLength(const struct sockaddr_in6 *netmask)
...
+
+        memcpy(RTA_DATA(rta), &cacheinfo, sizeof(cacheinfo));
+
+        req.nh.nlmsg_len += rta->rta_len;
+    }
+
+    if (send(sNetlinkFd, &req, req.nh.nlmsg_len, 0) != -1)
+    {
+        otLogInfoPlat("Sent request#%u to %s %s/%u", sNetlinkSequence, (aIsAdded ? "add" : "remove"),
+                      static_cast<const Ip6AddressString &>(Ip6AddressString(aAddressInfo.mAddress)).ToString(),
```
gjc13, line 381: The const cast is not required.

gjc13 (approved):

src/posix/platform/netif.cpp
```diff
@@ -152,6 +152,40 @@ char         gNetifName[IFNAMSIZ];
 
 #if OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE
 
+namespace {
+
+/**
+ * This utility class converts binary IPv6 address to text format.
+ *
+ */
+class Ip6AddressString
```
wgtdkp, line 161: There is already a `ToString` method in the internal `Ip6::Address` class:
https://github.com/openthread/openthread/blob/fa57212b82e34a95c6b3302ee9ec2047388ab342/src/core/net/ip6_address.hpp#L880-L886

Is it applicable at here?

src/posix/platform/netif.cpp
```diff
@@ -298,36 +325,81 @@ static uint8_t NetmaskToPrefixLength(const struct sockaddr_in6 *netmask)
...
+    }
+}
+#endif // defined(__linux__)
+
+static void UpdateUnicast(otInstance *aInstance, const otIp6AddressInfo &aAddressInfo, bool aIsAdded)
+{
+    OT_UNUSED_VARIABLE(aInstance);
+
+    assert(sInstance == aInstance);
+    assert(sIpFd >= 0);
```
wgtdkp, line 396: Should we use `OT_ASSERT` instead?

src/posix/platform/netif.cpp
bukepo: as you see, it's in the core library.

src/posix/platform/netif.cpp
```diff
@@ -298,36 +325,81 @@ static uint8_t NetmaskToPrefixLength(const struct sockaddr_in6 *netmask)
...
+    req.nh.nlmsg_len   = NLMSG_LENGTH(sizeof(struct ifaddrmsg));
+    req.nh.nlmsg_flags = NLM_F_REQUEST | NLM_F_ACK | NLM_F_CREATE | NLM_F_EXCL;
+    req.nh.nlmsg_type  = aIsAdded ? RTM_NEWADDR : RTM_DELADDR;
+    req.nh.nlmsg_pid   = 0;
+    req.nh.nlmsg_seq   = ++sNetlinkSequence;
 
-#if defined(__linux__)
+    req.ifa.ifa_family    = AF_INET6;
+    req.ifa.ifa_prefixlen = aAddressInfo.mPrefixLength;
+    req.ifa.ifa_flags     = aAddressInfo.mIsAnycast ? IFA_F_NODAD : 0;
```
simonlingoogle, line 350: Should use `IFA_F_NODAD` for all Thread addresses? I think OpenThread should manage it's addresses by itself. Thus, no DAD is required for these addresses.

src/posix/platform/netif.cpp
```diff
@@ -152,6 +152,40 @@ char         gNetifName[IFNAMSIZ];
...
+class Ip6AddressString
+{
+public:
+    /**
+     * The constructor of this converter.
+     *
+     * @param[in]   aAddress    A pointer to a buffer holding an IPv6 address.
+     *
+     */
+    Ip6AddressString(const void *aAddress)
```
simonlingoogle, line 170:
```suggestion
    Ip6AddressString(const otIp6Address *aAddress)
```

include/openthread/ip6.h
```diff
@@ -415,19 +415,30 @@ typedef void (*otIp6ReceiveCallback)(otMessage *aMessage, void *aContext);
...
+ *
+ * This structure represents IPv6 address information.
+ *
+ */
+typedef struct otIp6AddressInfo
+{
+    const otIp6Address *mAddress;       ///< A pointer to the IPv6 address.
+    uint8_t             mPrefixLength;  ///< The prefix length of mAddress if it is a unicast address.
+    uint8_t             mScope : 4;     ///< The scope of this address.
+    bool                mIsAnycast : 1; ///< Whether this is an anycast address.
```
simonlingoogle, line 429:
```suggestion
    bool                mIsAloc : 1; ///< Whether this is an ALOC address.
```

src/core/net/netif.cpp
```diff
@@ -42,6 +42,32 @@
...
+class AddressInfo : public otIp6AddressInfo
+{
+public:
+    AddressInfo(const NetifUnicastAddress &aAddress, Instance &aInstance)
+    {
+        mAddress      = &aAddress.mAddress;
+        mPrefixLength = aAddress.mPrefixLength;
+        mScope        = aAddress.GetScope();
+        mIsAnycast    = aInstance.Get<Mle::MleRouter>().IsAnycastLocator(aAddress.GetAddress());
+        mIsRloc       = aAddress.mRloc;
```
simonlingoogle, line 54: nit: Instead of using mutliple bit filelds (i.e. `mIsAnycast`, `mIsRloc`), what about define address types: e.g.

- RLOC
- ALOC
- MLEID
- OTHER

So, we can save more information in the two bits.

simonlingoogle (approved):
LGTM :+1: 

with nits

src/posix/platform/netif.cpp
bukepo: There are multiple instances of assert() in posix platform. I tend to address this issue together in a separate PR. Thoughts?

src/posix/platform/netif.cpp
```diff
@@ -152,6 +152,40 @@ char         gNetifName[IFNAMSIZ];
...
+        VerifyOrDie(inet_ntop(AF_INET6, aAddress, mBuffer, sizeof(mBuffer)) != nullptr, OT_EXIT_ERROR_ERRNO);
+    }
+
+    /**
+     * This method returns the text format of the IPv6 address.
+     *
+     * @returns A pointer to the converted text format of the IPv6 address.
+     *
+     */
+    const char *Get(void) const { return mBuffer; }
```
simonlingoogle, line 181: nit:
```suggestion
    const char *AsCString(void) const { return mBuffer; }
```

include/openthread/ip6.h
bukepo: We care about whether it is a anycast or not, not whether it is a ALOC.

src/core/net/netif.cpp
bukepo: I will remove the mIsRloc, which is not used.

src/posix/platform/netif.cpp
bukepo: This is for all IPv6 address structures or objects following the inet_ntop API.

simonlingoogle: It is. But we only use it for `const otIp6Address *aAddress`? Use `const otIp6Address *aAddress` for better type checking?

src/core/net/netif.cpp
simonlingoogle: SGTM

src/posix/platform/netif.cpp
simonlingoogle: `OT_ASSERT` is defined in OT core. We can't use it in platform code.

wgtdkp (approved):
LGTM 👍

jwhui (approved):

jwhui:
@bukepo , please resolve failing checks.

PR #5630, opened 2020-10-10
[indirect-sender] no replacing transmitting supervision message

bukepo:
If MAC layer is transmitting a supervision message, and then another
message is delivered to the child, the supervision message will be dequeue from the
mesh forwarder send queue. And the supervision message is transmitted,
the dequeue operation will fail because the message is not in the queue.

### DIFF
--- a/src/core/mac/data_poll_handler.hpp
+++ b/src/core/mac/data_poll_handler.hpp
@@ -260,6 +260,12 @@ public:
      */
     void RequestFrameChange(FrameChange aChange, Child &aChild);
 
+    /**
+     * This method returns the current indirect tx child.
+     *
+     */
+    const Child *GetIndrectTxChild(void) const { return mIndirectTxChild; }
+
 private:
     // Callbacks from MAC
     void    HandleDataPoll(Mac::RxFrame &aFrame);
--- a/src/core/thread/indirect_sender.cpp
+++ b/src/core/thread/indirect_sender.cpp
@@ -276,6 +276,9 @@ void IndirectSender::RequestMessageUpdate(Child &aChild)
 
     VerifyOrExit(!aChild.IsWaitingForMessageUpdate(), OT_NOOP);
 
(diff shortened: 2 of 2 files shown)


gjc13 (approved):

abtink:
@bukepo, very nice find indeed! Thanks. 👍 

- The current approach/fix of checking `VerifyOrExit(Get<DataPollHandler>().GetIndrectTxChild() != &aChild, OT_NOOP);` does indeed work but it adds a new API and interaction between `DataPollHanlder` and `IndirectSender`. 
- The idea behind separating these into two modules was to be able to later push the `DataPollHandler` into RCP. 
- I think checking state of `Get<DataPollHandler>().GetIndrectTxChild()`  from `IndirectSender` (when/if it is running on RCP) would not be easy (there may be delays/race situation between RCP and next layer when asking about it current state/tx-child).
- This is partly the reason we have the whole async purge/update change call request and callback model between the two.

I have a suggestion for another potential fix:
- The issue seem to happen due to `FindIndirectMessage()` removing a message from child (without calling `RemoveMessageFromSleepyChild()` to purge/change the frame properly on `DataPollHandler`).
- How about when a new message is being added for a child,  from `AddMessageForSleepyChild()`, we check if there is an existing "supervision" message in the queue for the same child and if so, we remove it then? 

Thoughts?

abtink:
@bukepo , I wrote the alternate fix here from this branch https://github.com/abtink/openthread/commit/c911af216f3266e3a974962595be1e44e763de98. We can submit this instead of as fix

There is one thing which we can optimize later related to this fix. With this we would search in the message queue for existing sup msgs even if there are none queued. We can add mecansim to track in `Child` whether there is a child sup already enqueued to optimize the search - I did not add it for now to keep the fix smaller and easier to cherry-pick.

bukepo:
@abtink Thanks for the suggestion! I agree to submit the alternative fix.

PR #5632, opened 2020-10-12
[ip6] no anycast as source address

bukepo:
* Check source address of packets from users
* Determine anycast address by MLE API for accuracy
* Replace anycast address with a valid unicast address when sending with platform UDP in UDP layer
* Replace anycast address with a valid unicast address in IP layer

### DIFF
--- a/src/core/net/udp6.cpp
+++ b/src/core/net/udp6.cpp
@@ -327,6 +327,15 @@ otError Udp::SendTo(SocketHandle &aSocket, Message &aMessage, const MessageInfo 
     if (!IsMlePort(aSocket.mSockName.mPort) &&
         !(aSocket.mSockName.mPort == Tmf::kUdpPort && aMessage.GetSubType() == Message::kSubTypeJoinerEntrust))
     {
+        // Change anycast address to unicast address here because host may be unclear about the anycast address
+        if (Get<Mle::Mle>().IsAnycastLocator(messageInfoLocal.GetSockAddr()))
+        {
+            const NetifUnicastAddress *netifAddr = Get<Ip6>().SelectSourceAddress(messageInfoLocal);
+
+            VerifyOrExit(netifAddr != nullptr, error = OT_ERROR_INVALID_ARGS);
+            messageInfoLocal.SetSockAddr(netifAddr->GetAddress());
+        }
+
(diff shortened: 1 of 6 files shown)


librasungirl (approved):

simonlingoogle (approved):
LGTM :+1:

abtink (approved):
LGTM. 👍

jwhui:
@bukepo , thread-1-2-backbone test seems to be failing consistently. Can you take a look?

src/core/net/udp6.cpp
```diff
@@ -327,6 +327,15 @@ otError Udp::SendTo(SocketHandle &aSocket, Message &aMessage, const MessageInfo
     if (!IsMlePort(aSocket.mSockName.mPort) &&
         !(aSocket.mSockName.mPort == Tmf::kUdpPort && aMessage.GetSubType() == Message::kSubTypeJoinerEntrust))
     {
+        // Change anycast address to unicast address here because host may be unclear about the anycast address
```
jwhui, line 330: Just curious - in what cases can this occur?

bukepo: On-mesh commissioner sends petition to leader with ALOC, and since the TMF is running on host IP stack which doesn't know it is an ALOC, so it will respond with the ALOC as source address.

jwhui: Can/should we update TMF to not use ALOC as source address?

bukepo: I think so. But it seems preventing use anycast as source address is a general issue. To prevent this issue occurring again, I think it might be solved here. Thoughts?

jwhui: I'm wondering if it is better to return error here rather than trying to fix it automatically and "hide it under the covers". Thoughts?

simonlingoogle: It's quite common for us to handle TMF and send response using the same `aMessageInfo`, in which the peer address might be an ALOC.

jwhui: SGTM - can we update the comment to something like: "Replace anycast address with a valid unicast address since response messages typically copy the peer address."?

bukepo:
```suggestion
        // Replace anycast address with a valid unicast address since response messages typically copy the peer address
```

jwhui (approved):

PR #5636, opened 2020-10-12
[cmd-parser] remove dependency of <limits>

Moandor-y:
We don't have C++ STL on some platforms, so this replaces it.

### DIFF
--- a/src/core/utils/parse_cmdline.cpp
+++ b/src/core/utils/parse_cmdline.cpp
@@ -36,69 +36,8 @@
 #include <string.h>
 
 #include "common/code_utils.hpp"
+#include "common/numeric_limits.hpp"
 #include "net/ip6_address.hpp"
-
-namespace {
-template <typename Type> class NumericLimits
-{
-};
-
-template <> class NumericLimits<int8_t>
-{
-public:
-    static constexpr int8_t Min() { return INT8_MIN; }
-    static constexpr int8_t Max() { return INT8_MAX; }
-};
-
-template <> class NumericLimits<int16_t>
-{
-public:
-    static constexpr int16_t Min() { return INT16_MIN; }
-    static constexpr int16_t Max() { return INT16_MAX; }
-};
-
-template <> class NumericLimits<int32_t>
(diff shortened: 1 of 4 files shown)


bukepo (approved):
LGTM:+1:

Irving-cl (approved):
LGTM👍

src/core/utils/parse_cmdline.cpp
```diff
@@ -33,12 +33,73 @@
 
 #include "parse_cmdline.hpp"
 
-#include <limits>
 #include <string.h>
 
 #include "common/code_utils.hpp"
 #include "net/ip6_address.hpp"
 
+namespace {
```
abtink, line 41: How about adding this as a new common header e.g., `common/numeric_limits.hpp`?
This way it can also be used by other core modules.

src/core/utils/parse_cmdline.cpp
```diff
@@ -33,12 +33,73 @@
...
+template <typename Type> class NumericLimits
+{
+};
+
+template <> class NumericLimits<int8_t>
+{
+public:
+    static constexpr int8_t Min() { return INT8_MIN; }
```
abtink, line 49: Adding `void` ...Same for all other definitions...
```suggestion
    static constexpr int8_t Min(void) { return INT8_MIN; }
```

abtink (commented):
Look good overall. One suggestion below on moving the definition under `common` folder.

simonlingoogle (approved):
It's great to have. Thanks!
LGTM :+1:

src/core/utils/parse_cmdline.cpp
Moandor-y: Done.

src/core/utils/parse_cmdline.cpp
Moandor-y: Done.

abtink (approved):
LGTM. Thanks 👍

jwhui (approved):

PR #5637, opened 2020-10-12
[docs] update BBB guide formatting

Vyrastas:
Fixing a few formatting things I should've caught when reviewing the update to the BBB guide:

- The callouts should only have a single > on the first line
- The ### within a code block was causing issues with my import script... easier to just change them to a single #

### DIFF
--- a/doc/site/en/guides/border-router/beaglebone-black.md
+++ b/doc/site/en/guides/border-router/beaglebone-black.md
@@ -20,9 +20,9 @@ Steps to enable
 1. Set up a Wi-Fi access point
 
 > Note: The BBB does not have built-in Wi-Fi support. This guide was built and
-> tested with a BBONE-GATEWAY-CAPE for Wi-Fi AP operation. Some BeagleBone
-> variants have onboard Wi-Fi capability, and some of this guide may be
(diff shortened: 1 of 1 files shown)


doc/site/en/guides/border-router/beaglebone-black.md
```diff
@@ -152,10 +152,10 @@ with the command `sudo vim /boot/uEnv.txt` and make sure the following lines
 match:
 
 ```
-###Custom Cape
+#Custom Cape
```
srickardti, line 155: This won't match what is in the file `/boot/uEnv.txt`, but I think people will understand anyway.

srickardti (approved):

doc/site/en/guides/border-router/beaglebone-black.md
Vyrastas: I would hope so, if it's a problem we can address later.

Vyrastas:
@srickardti FYI your guide update is live on the site:

https://openthread.io/guides/border-router/beaglebone-black

jwhui (approved):

PR #5645, opened 2020-10-14
[coap] allow max URI path length

simonlingoogle:
This PR:
- allow max URI path length of 32
- change error to `OT_ERROR_PARSE` (used to be `OT_ERROR_NOT_FOUND`)

### DIFF
--- a/src/core/coap/coap.cpp
+++ b/src/core/coap/coap.cpp
@@ -666,7 +666,7 @@ exit:
 
 void CoapBase::ProcessReceivedRequest(Message &aMessage, const Ip6::MessageInfo &aMessageInfo)
 {
-    char             uriPath[Resource::kMaxReceivedUriPath];
+    char             uriPath[Resource::kMaxReceivedUriPath + 1];
     char *           curUriPath     = uriPath;
     Message *        cachedResponse = nullptr;
     otError          error          = OT_ERROR_NOT_FOUND;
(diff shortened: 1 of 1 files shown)


wgtdkp (approved):
LGTM :+1:

bukepo (approved):
LGTM:+1:

src/core/coap/coap.cpp
```diff
@@ -704,7 +704,7 @@ void CoapBase::ProcessReceivedRequest(Message &aMessage, const Ip6::MessageInfo
             *curUriPath++ = '/';
         }
 
-        VerifyOrExit(curUriPath + optionLength < OT_ARRAY_END(uriPath), OT_NOOP);
+        VerifyOrExit(curUriPath + optionLength + 1 < OT_ARRAY_END(uriPath), error = OT_ERROR_PARSE);
```
abtink, line 707: I don't follow why we would need `+1` here in the `VerifyOrExit` check. Can you please explain? Thanks.

Here is my reading of the code (maybe I am missing something)?
- `curUriPath` points to current location to write to.
- When reading the value we would write `optionLength` chars from `curUriPath[0]` up to `curUriPath[optionLength-1]`
- And update the `curUriPath` by `curUriPath += optionLength` 
- The check `curUriPath + optionLength  < OT_ARRAY_END(uriPath)` ensures the updated `curUriPath`  still points to somewhere within the array (not passed the end of array). 
- And we should still be able to do `*curUriPath = '\0';` (since `curUriPath` should still point to a valid byte within array).

abtink (approved):
LGTM. Thanks

src/core/coap/coap.cpp
simonlingoogle: I think you are right. 
I changed this PR to just introduce two minor enhancements. 
Thanks.

wgtdkp (approved):

librasungirl (approved):

jwhui (approved):

Issue #5648, opened 2020-10-14
[github-actions] posix-pty fails frequently


jwhui:
**Describe the bug** A clear and concise description of what the bug is.

posix-pty fails frequently.

**Console/log output** If applicable, add console/log output to help explain your problem.

```
Keep POSIX Only
Run rm -rf build/x86_64-unknown-linux-gnu/
  rm -rf build/x86_64-unknown-linux-gnu/
  tar -xf build/posix.tar -C build/
  shell: /bin/bash -e {0}
  env:
    COVERAGE: 1
tar: posix/x86_64-unknown-linux-gnu/third_party/Makefile: Cannot open: Permission denied
tar: posix/x86_64-unknown-linux-gnu/third_party/NordicSemiconductor: Cannot mkdir: Permission denied
tar: posix/x86_64-unknown-linux-gnu/third_party/NordicSemiconductor: Cannot mkdir: Permission denied
tar: posix/x86_64-unknown-linux-gnu/third_party/NordicSemiconductor/libraries: Cannot mkdir: No such file or directory
tar: posix/x86_64-unknown-linux-gnu/third_party/NordicSemiconductor: Cannot mkdir: Permission denied
tar: posix/x86_64-unknown-linux-gnu/third_party/NordicSemiconductor/libraries/utf_converter: Cannot mkdir: No such file or directory
tar: posix/x86_64-unknown-linux-gnu/third_party/NordicSemiconductor: Cannot mkdir: Permission denied
tar: posix/x86_64-unknown-linux-gnu/third_party/NordicSemiconductor/libraries/utf_converter/.deps: Cannot mkdir: No such file or directory
tar: posix/x86_64-unknown-linux-gnu/third_party/NordicSemiconductor: Cannot mkdir: Permission denied
tar: posix/x86_64-unknown-linux-gnu/third_party/NordicSemiconductor/libraries/utf_converter/.deps/libnordicsemi_nrf52840_sdk_a-utf.Po: Cannot open: No such file or directory
tar: posix/x86_64-unknown-linux-gnu/third_party/NordicSemiconductor: Cannot mkdir: Permission denied
tar: posix/x86_64-unknown-linux-gnu/third_party/NordicSemiconductor/libraries/utf_converter/.deps/libnordicsemi_nrf52811_sdk_a-utf.Po: Cannot open: No such file or directory
tar: posix/x86_64-unknown-linux-gnu/third_party/NordicSemiconductor: Cannot mkdir: Permission denied
[...]

PR #5654, opened 2020-10-15
[continuous-integration] kill ot-rcp and ot-cli in pty check

Moandor-y:
This adds kill commands to POSIX pty check, to make sure all processes are stopped before the check completes. Fixes https://github.com/openthread/openthread/issues/5648.

### DIFF
--- a/script/check-posix-pty
+++ b/script/check-posix-pty
@@ -43,6 +43,8 @@ at_exit()
     sudo killall expect || true
     sudo killall ot-ctl || true
     sudo killall ot-daemon || true
+    sudo killall ot-cli || true
+    sudo killall ot-rcp || true
     sudo killall socat || true
 
     exit $EXIT_CODE


bukepo (approved):
LGTM:+1:

abtink (approved):

Irving-cl (approved):
LGTM👍

jwhui (approved):

simonlingoogle:
I think the real issue is why do we have dangling ot-rcp processes when the test is already finished. 

Should we fix the root cause instead?

simonlingoogle (approved):
LGTM 👍 
since it's not ot-rcp issue.

Moandor-y:
> I think the real issue is why do we have ot-rcp processes running when the test is already finished.
> 
> Should we fix the root cause instead?

The `ot-rcp` process will stop when `socat` is killed, but there may be a short delay.

PR #5657, opened 2020-10-15
[notifier] signal change to active/pending dataset + simplify 'Notifier::EventToString'

abtink:

**[notifier] signal change to active or pending Operational Dataset**

**[notifier] simplify 'EventToString()'**
    
This commit changes `EventToString()` to use an array instead of
`switch`. This helps with code size due to enumerator cases being
`uint32_t` bitmask values and therefore compliers may not be able to
optimize the `switch` statement as a table.

### DIFF
--- a/src/core/common/notifier.hpp
+++ b/src/core/common/notifier.hpp
@@ -62,7 +62,7 @@ namespace ot {
  * This enumeration type represents events emitted from OpenThread Notifier.
  *
  */
-enum Event
+enum Event : uint32_t
 {
     kEventIp6AddressAdded                  = OT_CHANGED_IP6_ADDRESS_ADDED,            ///< IPv6 address was added
     kEventIp6AddressRemoved                = OT_CHANGED_IP6_ADDRESS_REMOVED,          ///< IPv6 address was removed
--- a/src/core/common/notifier.hpp
+++ b/src/core/common/notifier.hpp
@@ -92,6 +92,8 @@ enum Event
     kEventThreadBackboneRouterStateChanged = OT_CHANGED_THREAD_BACKBONE_ROUTER_STATE, ///< Backbone Router state changed
     kEventThreadBackboneRouterLocalChanged = OT_CHANGED_THREAD_BACKBONE_ROUTER_LOCAL, ///< Local Backbone Router changed
     kEventJoinerStateChanged               = OT_CHANGED_JOINER_STATE,                 ///< Joiner state changed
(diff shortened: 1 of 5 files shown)


src/core/common/notifier.hpp
```diff
@@ -92,6 +92,8 @@ enum Event
     kEventThreadBackboneRouterStateChanged = OT_CHANGED_THREAD_BACKBONE_ROUTER_STATE, ///< Backbone Router state changed
     kEventThreadBackboneRouterLocalChanged = OT_CHANGED_THREAD_BACKBONE_ROUTER_LOCAL, ///< Local Backbone Router changed
     kEventJoinerStateChanged               = OT_CHANGED_JOINER_STATE,                 ///< Joiner state changed
+    kEventActiveDatasetChanged             = OT_CHANGED_ACTIVE_DATASET,               ///< Active Dataset changed
+    kEventPendingDatasetChanged            = OT_CHANGED_PENDING_DATASET,              ///< Pending Dataset changed
```
Irving-cl, line 96: Just a question: It seems that these events are always used alone when calling `Signal` instead of a bit mask way. So why not defining these values simply as 0-29? For `uint32_t` we can add at most 2 more events in the future.

Irving-cl (approved):
LGTM👍

src/core/common/notifier.hpp
abtink: We do use them as bitmask (which can contain multiple events) in the callbacks, e.g., in `otStateChangedCallback()` or in `HandleNotifierEvents()`.

`Signal()` method uses a single event but multiple events can be collected and emitted together (`Notifier` 
has its own `Tasklet` from which it would emit the collected events).

------------

I agree we may face a limit due to use of `uint32_t`. We have been careful about adding events and only adding ones that would be useful. In the case here, I think knowing when Datasets get changed would be helpful (for different situations).

We have intentionally defined `otChangedFlags` type which is used as parameter type in `otStateChangedCallback()`. It is currentlly defined  in `instance.h` (public header) as:
```
typedef uint32_t otChangedFlags;

typedef void (*otStateChangedCallback)(otChangedFlags aFlags, void *aContext);
```

This would allow us in future to change the `otChangedFlags` to `uint64_t` to get more possible `OT_CHANGED` events and hopefully minimize impact on users of the API.

Inside the core, we have `Events` type which acts as a set/list of events and provides helper methods to check whether it `Contains(Event)` and/or `ContainsAny()` or `ContainsAll()` of a collection of events.

jwhui (approved):

PR #5663, opened 2020-10-19
[border-agent] misc enhancements

abtink:
This commit contains smaller enhancements in `BorderAgent` module

- define `ForwardContext` as a private sub-type of BorderAgent
- change `ForwardContext` to be `InstanceLocatorInit` and avoid
  `new` (use `Init()`)
- add `HandleCoapResponse()` method which is called from `static`
  method `HandleCoapResponse()`.
- convert file `static` functions `SendErrorMessage()` to `BorderAgent`
  methods
- Add `State` enum and inline/remove `SetState()`.
- Smaller style/documentation fixes (e.g., remove unnecessary `{ }`).

### DIFF
--- a/src/core/meshcop/border_agent.cpp
+++ b/src/core/meshcop/border_agent.cpp
@@ -37,7 +37,6 @@
 #include "common/instance.hpp"
 #include "common/locator-getters.hpp"
 #include "common/logging.hpp"
-#include "common/new.hpp"
 #include "meshcop/meshcop.hpp"
 #include "meshcop/meshcop_tlvs.hpp"
 #include "thread/thread_netif.hpp"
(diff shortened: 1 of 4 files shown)


src/core/meshcop/border_agent.cpp
```diff
@@ -218,57 +156,68 @@ void BorderAgent::HandleCoapResponse(void *               aContext,
...
-    const Coap::Message *response       = static_cast<const Coap::Message *>(aMessage);
-    Coap::Message *      message        = nullptr;
-    otError              error;
+    ForwardContext &forwardContext = *static_cast<ForwardContext *>(aContext);
+
+    forwardContext.Get<BorderAgent>().HandleCoapResponse(forwardContext, static_cast<const Coap::Message *>(aMessage),
+                                                         aResult);
+}
+
+void BorderAgent::HandleCoapResponse(ForwardContext &aForwardContext, const Coap::Message *aResponse, otError aResult)
```
jwhui, line 165: Can `aResponse` be set to `nullptr`? If not, pass by reference?

abtink: Was thinking of the same, but checked the callback and its use, and it can be set to  `nullptr` in case of error (e.g. `aResult == OT_ERROR_RESPONSE_TIMEOUT`)  when no response is received.

jwhui (approved):

PR #5670, opened 2020-10-20
[backbone-router] cleanup DomainPrefixState

simonlingoogle:
This commit removes two Domain Prefix states that are not actually useful.
 - kDomainPrefixNone
 - kDomainPrefixUnchanged


librasungirl (approved):

simonlingoogle:
It seems to be a difficult change. Will figure it out when I have time later.

PR #5664, opened 2020-10-19
[dua] implement Backbone query and answer

simonlingoogle:
This PR implements DUA features of Thread 1.2 DUA that involves sending and processing Backbone query and answer:

- [x] Extend address query to the Backbone Link
  - [x] Send BB.qry for Thread address query
  - [x] Send ADDR_ANS.ntf for BB.ans 
- [x] DAD process
  - [x] PBBR to send BB.qry on the Backbone link (3 times if not answered)
  - [x] PBBR to receive and handle BB.qry and sends BB.ans
  - [x] PBBR to receive BB.ans and sends ADDR_ERR.ntf if duplicated
- [x] PRO_BB.ntf
  - [x] Send PRO_BB.ntf when DUA registration is updated
  - [x] Handle PRO_BB.ntf
- [x] Add test_dua_dad.py to test the DAD process when it was successful or duplicated
    - [x] Verify the normal DAD process
    - [x] Verify Address Query can be extended to the Backbone link
    - [x] Verify DAD duplicate is handled correctly
    - [x] Verify PRO_BB.ntf for duplicated DUA is handled correctly

PR #5671, opened 2020-10-20
[dua] extend ADDR.qry to BB.qry

simonlingoogle:
This commit extends ADDR.qry in the Thread network to BB.qry on the Backbone link:
 - [x] PBBR to send BB.qry whenver ADDR.qry is received or sent by PBBR itself
 - [x] Handle BB.ans with src RLOC16 and sends the corresponding ADDR_NTF.ans

Based on #5664


simonlingoogle:
Close this PR since it's merged to https://github.com/openthread/openthread/pull/5664

PR #5673, opened 2020-10-20
[notifier] define 'kEventStrings[]' as const pointer

abtink:


simonlingoogle (approved):
LGTM :+1: 

I tried the same change and compared the size, but didn't see any chance in code or memory size. Why does it has no impact on memory size?

Irving-cl (approved):
LGTM

abtink:
> I tried the same change and compared the size, but didn't see any chance in code or memory size. Why does it has no impact on memory size?

@simonlingoogle , my guess is that complier may be already treating the array as a `const`.

simonlingoogle:
> > I tried the same change and compared the size, but didn't see any chance in code or memory size. Why does it has no impact on memory size?
> 
> @simonlingoogle , my guess is that complier may be already treating the array as a `const`.

But I **did** see memory increase for `sThreadErrorStrings` if `*const` is not used. I think it's safer to always use `*const`.

jwhui (approved):

PR #5680, opened 2020-10-22
[thread-cert] refactor case 5.1.1 using pktverify

FiveDimensions:


tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -168,6 +179,317 @@ def test(self):
...
+            filter(lambda p: {\
+                              CHALLENGE_TLV,\
+                              CONNECTIVITY_TLV,\
+                              LEADER_DATA_TLV,\
+                              LINK_LAYER_FRAME_COUNTER_TLV,\
+                              LINK_MARGIN_TLV,\
+                              RESPONSE_TLV,\
+                              SOURCE_ADDRESS_TLV,\
+                              VERSION_TLV,\
+                              MLE_FRAME_COUNTER_TLV\
```
simonlingoogle, line 262:
```suggestion
            filter(lambda p: {
                              CHALLENGE_TLV,
                              CONNECTIVITY_TLV,
                              LEADER_DATA_TLV,
                              LINK_LAYER_FRAME_COUNTER_TLV,
                              LINK_MARGIN_TLV,
                              RESPONSE_TLV,
                              SOURCE_ADDRESS_TLV,
                              VERSION_TLV,
                              MLE_FRAME_COUNTER_TLV
```
not necessary to use `\` ?

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -168,6 +179,317 @@ def test(self):
...
+        # Step 7: Leader sends an Address Solicit Response.
+        #         Ensure the Address Solicit Response is properly formatted:
+        #         CoAP Response Code
+        #             2.04 Changed
+        #         CoAP Payload
+        #             - Status TLV (value = Success)
+        #             - RLOC16 TLV
+        #             - Router Mask TLV
+
+        pkts.filter_wpan_src64(LEADER).\
```
librasungirl, line 393: Generally TMF in Thread Core should use RLOC address as source, would it be better to use a filter like `filter_wpan_src16()` here?

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -61,6 +70,8 @@ def test(self):
         self.simulator.go(7)
         self.assertEqual(self.nodes[ROUTER].get_state(), 'router')
 
+        self.collect_rloc16s()
+
```
librasungirl, line 74: should the packet verification code with deprecated method below be removed?

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
simonlingoogle: I recalled one issue regarding using rloc16 as the source filter: the rloc16 of device can change when it upgraded to Router or downgraded. Unlike RLOC16, the src64 is usually fixed. I think this is the reason why Test Harness also chose to use src64 for filtering. 
Currently, `pktverify` collects the RLOC16s of devices using `collect_rloc16s()`. So, the RLOC values could vary according to when this method is called during test procedure. 

Using the correct RLOC16 of each device at each step needs extra work, especially for the cases when the device could reset and reattach during the test. 

So, maybe using  `filter_wpan_src64` is easier and good enough for Test Plan?

Thoughts?

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -168,6 +179,317 @@ def test(self):
         for addr in self.nodes[LEADER].get_addrs():
             self.assertTrue(self.nodes[ROUTER].ping(addr))
```
simonlingoogle, line 180: The test plan actually requires pining DUT's link local address. 
```
Verify connectivity by
sending an ICMPv6 Echo
Request to the DUT link
local address. 
```

So, we might only need to ping the link local address rather than pinging each address?
Also, I think we should also ping ROUTER from LEADER because both ROUTER and LEADER can be the DUT device.

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -168,6 +179,317 @@ def test(self):
...
+                              NETWORK_DATA_TLV,\
+                              SOURCE_ADDRESS_TLV,\
+                              ROUTE64_TLV\
+                              } <= set(p.mle.tlv.type) or\
+                             {\
+                              ADDRESS16_TLV,\
+                              LEADER_DATA_TLV,\
+                              NETWORK_DATA_TLV,\
+                              SOURCE_ADDRESS_TLV\
+                               } <= set(p.mle.tlv.type)).\
```
simonlingoogle, line 363:
```suggestion
            filter(lambda p: {\
                              ADDRESS16_TLV,\
                              LEADER_DATA_TLV,\
                              NETWORK_DATA_TLV,\
                              SOURCE_ADDRESS_TLV\
                               } <= set(p.mle.tlv.type)).\
```
I think these two are same.

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -168,6 +179,317 @@ def test(self):
...
+        #             - Link-layer Frame Counter TLV
+        #             - Mode TLV
+        #             - Response TLV
+        #             - Timeout TLV
+        #             - TLV Request TLV
+        #             - Address16 TLV
+        #             - Network Data TLV
+        #             - Route64 TLV (optional)
+        #             - Version TLV
+        #             - MLE Frame Counter TLV (optional)
```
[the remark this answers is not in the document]
simonlingoogle: If Test Plan says that one TLV is optional, maybe we can just skip checking for it since whether this TLV is included does not impact the verification result. 
But if Test Plan requires extra checking when the TLV field exists, we should add corresponding code like `mle.tlv.xxx is null or mle.tlv.xxx == yyy`.

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -168,6 +179,317 @@ def test(self):
...
+            filter_mle_cmd(MLE_CHILD_ID_REQUEST).\
+            filter(lambda p: {\
+                              LINK_LAYER_FRAME_COUNTER_TLV,\
+                              MODE_TLV,\
+                              RESPONSE_TLV,\
+                              TIMEOUT_TLV,\
+                              TLV_REQUEST_TLV,\
+                              ADDRESS16_TLV,\
+                              NETWORK_DATA_TLV,\
+                              ROUTE64_TLV,\
```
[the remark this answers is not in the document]
simonlingoogle: We can use extra checking to achieve this purpose, e.g. `mle.tlv.addr16 is not null` should be True if it's a real `Address16 TLV`.

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -168,6 +179,317 @@ def test(self):
...
+                              MLE_FRAME_COUNTER_TLV\
+                              } <= set(p.mle.tlv.type) or\
+                             {\
+                              LEADER_DATA_TLV,\
+                              LINK_LAYER_FRAME_COUNTER_TLV,\
+                              LINK_MARGIN_TLV,\
+                              RESPONSE_TLV,\
+                              SOURCE_ADDRESS_TLV,\
+                              VERSION_TLV\
+                               } <= set(p.mle.tlv.type)).\
```
simonlingoogle, line 461:
```suggestion
            filter(lambda p: {\
                              LEADER_DATA_TLV,\
                              LINK_LAYER_FRAME_COUNTER_TLV,\
                              LINK_MARGIN_TLV,\
                              RESPONSE_TLV,\
                              SOURCE_ADDRESS_TLV,\
                              VERSION_TLV\
                               } <= set(p.mle.tlv.type)).\
```

These two code is same.

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
FiveDimensions: I thinks these duplicate verification should be removed

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -168,6 +179,317 @@ def test(self):
...
+            filter_mle_cmd(MLE_ADVERTISEMENT).\
+            filter(lambda p: {\
+                              LEADER_DATA_TLV,\
+                              ROUTE64_TLV,\
+                              SOURCE_ADDRESS_TLV\
+                              } == set(p.mle.tlv.type) and\
+                   p.ipv6.hlim == 255).\
+            must_next()
+
+        # Step 11: DUT responds with ICMPv6 Echo Reply
```
simonlingoogle, line 483: Both Leader and Router can be the DUT.

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -168,6 +179,317 @@ def test(self):
...
+                              SCAN_MASK_TLV,\
+                              VERSION_TLV\
+                              } <= set(p.mle.tlv.type) and\
+                   p.ipv6.hlim == 255)\
+            .next()
+
+        if _pkt is not None:
+            _pkt.must_verify(lambda p:
+                    p.mle.tlv.scan_mask.r == 1 and\
+                    p.mle.tlv.scan_mask.e == 1)
```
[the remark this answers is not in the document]
FiveDimensions: Yes, for the real devices could send the second parent request to all routers & reeds. In current test setup, there is no second parent request. Do we need to check the second one in this case? @librasungirl @simonlingoogle

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
simonlingoogle: FYI: if we remove the old verification code, we can set `USE_MESSAGE_FACTORY = False` for the test case to be a little bit more efficient. @FiveDimensions

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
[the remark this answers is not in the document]
FiveDimensions: use "is nullField" to check the these requested TLVS

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
FiveDimensions: added another ping for both DUT roles

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
FiveDimensions: added

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -168,6 +179,317 @@ def test(self):
...
+            filter(lambda p: {\
+                              CHALLENGE_TLV,\
+                              MODE_TLV,\
+                              SCAN_MASK_TLV,\
+                              VERSION_TLV\
+                              } <= set(p.mle.tlv.type) and\
+                   p.ipv6.hlim == 255 and\
+                   p.mle.tlv.scan_mask.r == 1 and\
+                   p.mle.tlv.scan_mask.e == 0).\
+            must_next()
```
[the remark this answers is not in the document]
FiveDimensions: this the first parent request from the router only to all routers, otherwise test fails

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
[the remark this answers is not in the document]
FiveDimensions: For the current nodes setup circumstance, there should be no second parent request, delete the check

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
FiveDimensions: Agree

simonlingoogle (approved):
LGTM :+1:

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -72,112 +73,14 @@ def test(self):
...
-        msg.assertSentWithHopLimit(255)
-        msg.assertSentToDestinationAddress("ff02::1")
-        msg.assertMleMessageContainsTlv(mle.SourceAddress)
-        msg.assertMleMessageContainsTlv(mle.LeaderData)
-        msg.assertMleMessageContainsTlv(mle.Route64)
-
-        # 11 - Leader, Router1
         for addr in self.nodes[LEADER].get_addrs():
-            self.assertTrue(self.nodes[ROUTER].ping(addr))
+            if addr[0:4] == 'fe80':
```
librasungirl, line 90: reuse `get_linklocal()` in node.py?

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
```diff
@@ -72,112 +73,14 @@ def test(self):
...
+                self.assertTrue(self.nodes[ROUTER].ping(addr))
+        self.simulator.go(5)
+
+        for addr in self.nodes[ROUTER].get_addrs():
+            if addr[0:4] == 'fe80':
```
librasungirl, line 95: ditto

tests/scripts/thread-cert/Cert_5_1_01_RouterAttach.py
FiveDimensions: use get_ip6_address instead to avoid spinel-cli unsupported ipaddr sub-cmd for linklocal

librasungirl (approved):

jwhui (approved):

Issue #5678, opened 2020-10-21
[Daemon]  Error decoding hdcl frame: nobuffs


tlifschitz:
**Describe the bug**
--------------
While running ot-daemon, hdlc frames are discarded in `HandleHdlcFrame` with error `OT_ERROR_NO_BUFS` when hdlc decoder tries to write decoded bytes from the RCP. This happens when multiple SEDs (10) attempt to attach at the same time or during intense data upload from SEDs to the FTD, making the network very unstable, and eventually making ot-daemon crash.

**To Reproduce**
-------------
1. Git commit id **9e3b450218582e721b102683de9786e7be5ef0b2**
2. IEEE 802.15.4 hardware platform **EFR32MG13/EFR32MG12**
3. Build steps
Linux recipe for ot-daemon
```
set(OT_BORDER_AGENT ON CACHE BOOL "enable border agent" FORCE)
set(OT_BORDER_ROUTER ON CACHE BOOL "enable border router feature" FORCE)
set(OT_BUILTIN_MBEDTLS_MANAGEMENT OFF CACHE BOOL "diable mbedTLS management" FORCE)
set(OT_COMMISSIONER ON CACHE BOOL "enable commissioner" FORCE)
set(OT_DAEMON ON CACHE BOOL "enable daemon mode" FORCE)
set(OT_JOINER ON CACHE BOOL "enable joiner" FORCE)
set(OT_LEGACY ON CACHE STRING "enable legacy network support" FORCE)
set(OT_MAC_FILTER ON CACHE STRING "enable MAC filter" FORCE)
set(OT_MBEDTLS_CONFIG_FILE "\"${PROJECT_SOURCE_DIR}/mbedtls-config.h\"" CACHE STRING "use mbedtls-config.h of this project" FORCE)
set(OT_PLATFORM "posix" CACHE STRING "use posix platform" FORCE)
set(OT_PLATFORM_NETIF ON CACHE BOOL "enable platform netif" FORCE)
set(OT_PLATFORM_UDP ON CACHE BOOL "enable platform UDP" FORCE)
set(OT_UDP_FORWARD OFF CACHE BOOL "diable udp forward" FORCE)
set(OT_SERVICE ON CACHE BOOL "enable services" FORCE)
set(OT_FULL_LOGS ON CACHE BOOL "enable full logs" FORCE)
set(OT_LOG_LEVEL_DYNAMIC ON CACHE BOOL "enable dynamic log level" FORCE)
[...]

PR #5687, opened 2020-10-22
[spinel-interface] add config option to change the rx frame buffer size

abtink:
Adds `OPENTHREAD_CONFIG_PLATFORM_RADIO_SPINEL_RX_FRAME_BUFFER_SIZE`
config option which specifies the rx frame buffer size used by
`SpinelInterface` in RCP host (posix) code. This is applicable/used when
`RadioSpinel` platform is used.

-----

This is related to openthread/openthread#5678.

### DIFF
--- a/src/core/config/platform.h
+++ b/src/core/config/platform.h
@@ -107,4 +107,15 @@
 #define OPENTHREAD_CONFIG_PLATFORM_RADIO_COEX_ENABLE 0
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_PLATFORM_RADIO_SPINEL_RX_FRAME_BUFFER_SIZE
+ *
+ * Specifies the rx frame buffer size used by `SpinelInterface` in RCP host (posix) code. This is applicable/used when
+ * `RadioSpinel` platform is used.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_PLATFORM_RADIO_SPINEL_RX_FRAME_BUFFER_SIZE
+#define OPENTHREAD_CONFIG_PLATFORM_RADIO_SPINEL_RX_FRAME_BUFFER_SIZE 8192
+#endif
+
 #endif // CONFIG_PLATFORM_H_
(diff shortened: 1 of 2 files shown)


src/core/config/platform.h
```diff
@@ -107,4 +107,15 @@
...
 
+/**
+ * @def OPENTHREAD_CONFIG_PLATFORM_RADIO_SPINEL_RX_FRAME_BUFFER_SIZE
+ *
+ * Specifies the rx frame buffer size used by `SpinelInterface` in RCP host (posix) code. This is applicable/used when
+ * `RadioSpinel` platform is used.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_PLATFORM_RADIO_SPINEL_RX_FRAME_BUFFER_SIZE
+#define OPENTHREAD_CONFIG_PLATFORM_RADIO_SPINEL_RX_FRAME_BUFFER_SIZE 2048
```
jwhui, line 118: Should we increase the default size given that we are already seeing issues with 2048?

abtink: I was considering it as well. I agree, it may be good to change the default to something larger (specially since this is intended for RCP host posix which would not be as memory constrained)

What would be good default number? `4096`?

abtink: Went for an 8K default buffer in new push.

Irving-cl (approved):
LGTM

simonlingoogle (approved):
LGTM :+1: 

Agree that we should increase the buffer size. 

Furthermore, should we increase the default baudrate to 460800 on RCP devices? Baudrate 115200 is simply not fast enough for transmitting 802.15.4 packets.

jwhui:
> Furthermore, should we increase the default baudrate to 460800 on RCP devices? Baudrate 115200 is simply not fast enough for transmitting 802.15.4 packets.

Yes, we should increase the default baud rate.

jwhui (approved):

PR #5694, opened 2020-10-23
[spinel] RCP restoration on failures

Moandor-y:
This PR implements the RCP restoration on failure feature. It keeps the host process running while resetting and initializing the RCP.


src/core/common/array_list.hpp
```diff
@@ -0,0 +1,153 @@
...
+};
+
+/**
+ * This class implements a list using array.
+ *
+ * @tparam Element   The type to element stored in the list.
+ * @tparam kMaxSize  The max size of the list.
+ *
+ */
+template <typename Element, ptrdiff_t kMaxSize> class ArrayList
```
simonlingoogle, line 56:
```suggestion
template <typename Element, uint16_t kMaxSize> class ArrayList
```
Can we use uint16_t ?

src/core/common/array_list.hpp
```diff
@@ -0,0 +1,153 @@
...
+     *
+     */
+    otError Append(const Element &aElement)
+    {
+        otError error = OT_ERROR_NONE;
+
+        if (mSize >= kMaxSize)
+        {
+            ExitNow(error = OT_ERROR_NO_BUFS);
+        }
```
simonlingoogle, line 81:
```suggestion
        VerifyOrExit (mSize < kMaxSize, error = OT_ERROR_NO_BUFS);
```

src/core/common/array_list.hpp
```diff
@@ -0,0 +1,153 @@
...
+        return error;
+    }
+
+    /**
+     * This method removes the element at the given index.
+     *
+     * @param[in] aIndex  The index to be removed.
+     *
+     */
+    void Remove(SizeType aIndex)
```
simonlingoogle, line 96: Define `Pop()` to remove the last element in the array?

Then we can remove the ith element efficiently using array[i] = array.Pop()

src/core/common/array_list.hpp
```diff
@@ -0,0 +1,153 @@
...
+    {
+        for (SizeType i = aIndex; i < mSize - 1; ++i)
+        {
+            mElements[i] = mElements[i + 1];
```
simonlingoogle, line 100: Should we use std::move ?

src/core/common/array_list.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * This method removes the element at the given index.
+     *
+     * @param[in] aIndex  The index to be removed.
+     *
+     */
+    void Remove(SizeType aIndex)
+    {
+        for (SizeType i = aIndex; i < mSize - 1; ++i)
```
simonlingoogle, line 98: Verify `aIndex < mSize` first?

include/openthread/error.h
```diff
@@ -234,6 +234,11 @@ typedef enum OT_MUST_USE_RESULT otError
      */
     OT_ERROR_PENDING = 36,
 
+    /**
+     * Operation is cancelled becuase RCP is recovering from failures.
+     */
+    OT_ERROR_RCP_RECOVERING = 37,
```
abtink, line 240: In general `otError` is for errors used by OT core modules. Core should not be aware of RCP so having error specific for this and the recovering process of RCP does not sound good to me.

Some Ideas/suggestions:
- Can you use existing error codes, like `BUSY`, `FAILED`, (maybe `PENDING`) for this purpose
- Looking at the use of this, seems like it's used to track whether the RCP recovery is happening to decide to save/discard frames. Can a boolean flag variable or state variable used to track this instead for this purpose instead of a new error value?

--------

(BTW,  I have not had the chance to read the rest of code yet, just noticed this `otError` addition so added this suggestion).

src/core/common/array_list.hpp
Moandor-y: Actually now I think it may not be a good idea to add this `ArrayList` to the codebase. It's more complicated than I thought. It should handle object destruction and copy control. Thoughts?

simonlingoogle: Can we limit the object type to be POD types to simplify the implementation?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -770,8 +779,11 @@ void RadioSpinel<InterfaceType, ProcessContextType>::HandleValueIs(spinel_prop_k
 
         if (status >= SPINEL_STATUS_RESET__BEGIN && status <= SPINEL_STATUS_RESET__END)
         {
-            // If RCP crashes/resets while radio was enabled, posix app exits.
-            VerifyOrDie(!IsEnabled(), OT_EXIT_RADIO_SPINEL_RESET);
+            if (IsEnabled())
+            {
+                HandleUnexpectedReset();
+                ExitNow(error = OT_ERROR_RCP_RECOVERING);
```
simonlingoogle, line 785:
```suggestion
                ExitNow(error = OT_ERROR_NONE);
```

Can we hide the error ?

src/core/common/array_list.hpp
abtink: I agree and also lean towards not adding `ArrayList` as its own data type (and use the model directly in posix RCP host code). 

I was considering if there would be other places within OT core where we could use this data type and could not immediately think of one. In general, I like the ["Three use to refactor" rule](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1329,9 +1443,22 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Set(spinel_prop_key_t aK
     assert(mWaitingTid == 0);
 
     mExpectedCommand = SPINEL_CMD_PROP_VALUE_IS;
+    mRcpRecoveredInOperation = false;
+
     va_start(mPropertyArgs, aFormat);
     error = RequestV(true, SPINEL_CMD_PROP_VALUE_SET, aKey, aFormat, mPropertyArgs);
     va_end(mPropertyArgs);
```
Moandor-y, line 1450: If RCP failure happens during calling `RequestV`, then in the recovery process `Set` is called before `RequestV` returns, `mPropertyArgs` will be overwritten and cause a crash. `mExpectedCommand` will also be overwritten.

Maybe we need to make these functions reentrant.

include/openthread/error.h
Moandor-y: Removed this error type.

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: Updated.

src/core/common/array_list.hpp
simonlingoogle: Removing this data structure seems good to me.

Moandor-y: Removed `ArrayList`.

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: This issue was solved by avoiding reentrancy. Moved recovery process out of `RequestV` calls.

src/lib/spinel/radio_spinel.hpp
```diff
@@ -884,22 +888,33 @@ template <typename InterfaceType, typename ProcessContextType> class RadioSpinel
...
+    otError                                                     mTxError;
+    char                                                        mVersion[kVersionStringSize];
+    otExtAddress                                                mIeeeEui64;
+    uint8_t                                                     mKeyIdMode;
+    uint8_t                                                     mKeyId;
+    otMacKey                                                    mPrevKey;
+    otMacKey                                                    mCurrKey;
+    otMacKey                                                    mNextKey;
+    ArrayList<uint16_t, OPENTHREAD_CONFIG_MLE_MAX_CHILDREN>     mSrcMatchShortEntries;
+    ArrayList<otExtAddress, OPENTHREAD_CONFIG_MLE_MAX_CHILDREN> mSrcMatchExtEntries;
```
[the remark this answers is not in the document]
Moandor-y: Done.

Moandor-y:
Functionalities completed. Now adding CI tests.

src/posix/platform/spi_interface.hpp
```diff
@@ -146,6 +146,11 @@ class SpiInterface
      */
     uint32_t GetBusSpeed(void) const { return ((mSpiDevFd >= 0) ? mSpiSpeedHz : 0); }
 
+    /**
+     * This method is called when RCP failure detected and restes internal states of the interface.
```
simonlingoogle, line 150:
```suggestion
     * This method is called when RCP failure detected and resets internal states of the interface.
```

src/posix/platform/hdlc_interface.hpp
```diff
@@ -156,6 +156,11 @@ class HdlcInterface
      */
     uint32_t GetBusSpeed(void) const { return mBaudRate; }
 
+    /**
+     * This method is called when RCP failure detected and restes internal states of the interface.
+     */
```
simonlingoogle, line 161:
```suggestion
     * This method is called when RCP failure detected and resets internal states of the interface.
     *
     */
```

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1297,6 +1429,21 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Get(spinel_prop_key_t aK
...
     mPropertyFormat = nullptr;
 
+#if OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_ENABLE
+    if (mRcpFailed)
```
simonlingoogle, line 1433: Do we need to check `mRcpFailed` **before** the first `RequestV` ?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1334,6 +1496,21 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Set(spinel_prop_key_t aK
...
+    {
+        RecoverFromRcpFailure();
+
+        mExpectedCommand = SPINEL_CMD_PROP_VALUE_IS;
+        va_start(mPropertyArgs, aFormat);
+        error = RequestV(true, SPINEL_CMD_PROP_VALUE_SET, aKey, aFormat, mPropertyArgs);
+        va_end(mPropertyArgs);
+        mExpectedCommand = SPINEL_CMD_NOOP;
+
+        VerifyOrDie(!mRcpFailed, OT_EXIT_FAILURE);
```
simonlingoogle, line 1510:
```suggestion
```

If we set properties in Spinel properly before calling `Set`, we actually don't need to retry the `Set` operation?
The device properties should be properly restored in `RecoverFromRcpFailure`.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1838,5 +2055,138 @@ uint32_t RadioSpinel<InterfaceType, ProcessContextType>::GetBusSpeed(void) const
...
+
+    mState = kStateDisabled;
+    mRxFrameBuffer.Clear();
+    mSpinelInterface.OnRcpReset();
+    mCmdTidsInUse = 0;
+    mCmdNextTid   = 1;
+    mTxRadioTid   = 0;
+    mWaitingTid   = 0;
+    mWaitingKey   = SPINEL_PROP_LAST_STATUS;
+    mIsReady      = false;
```
simonlingoogle, line 2106: When do we set `mIsReady = true` after reset restoration?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1297,6 +1429,21 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Get(spinel_prop_key_t aK
...
+#if OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_ENABLE
+    if (mRcpFailed)
+    {
+        RecoverFromRcpFailure();
+
+        mPropertyFormat = aFormat;
+        va_start(mPropertyArgs, aFormat);
+        error = RequestV(true, SPINEL_CMD_PROP_VALUE_GET, aKey, nullptr, mPropertyArgs);
+        va_end(mPropertyArgs);
+        mPropertyFormat = nullptr;
```
simonlingoogle, line 1441:
```suggestion
        mPropertyFormat = aFormat;
        va_start(mPropertyArgs, aFormat);
        error = RequestV(true, SPINEL_CMD_PROP_VALUE_GET, aKey, nullptr, mPropertyArgs);
        va_end(mPropertyArgs);
        mPropertyFormat = nullptr;
```
Reuse the code before `if (mRcpFailed)` somehow to reduce code redudancy?

src/lib/spinel/radio_spinel.hpp
```diff
@@ -900,6 +909,27 @@ template <typename InterfaceType, typename ProcessContextType> class RadioSpinel
     bool  mSupportsLogStream : 1; ///< RCP supports `LOG_STREAM` property with OpenThread log meta-data format.
     bool  mIsTimeSynced : 1;      ///< Host has calculated the time difference between host and RCP.
 
+    bool mResetRadioOnStartup : 1;
```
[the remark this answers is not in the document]
Moandor-y: Updated.

src/lib/spinel/radio_spinel.hpp
```diff
@@ -900,6 +909,27 @@ template <typename InterfaceType, typename ProcessContextType> class RadioSpinel
...
+    int16_t      mSrcMatchShortEntryCount;
+    otExtAddress mSrcMatchExtEntries[OPENTHREAD_CONFIG_MLE_MAX_CHILDREN];
+    int16_t      mSrcMatchExtEntryCount;
+    uint8_t      mScanChannel;
+    uint16_t     mScanDuration;
+    State        mRcpRecoveringState;
+    bool         mMacKeySet : 1;
+    bool         mRcpFailed : 1;
+    bool         mIsRcpRecovering : 1;
+    bool         mEnergyScanning : 1;
```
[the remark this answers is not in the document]
Moandor-y: Done.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1838,5 +2055,138 @@ uint32_t RadioSpinel<InterfaceType, ProcessContextType>::GetBusSpeed(void) const
...
+
+    if (mIsRcpRecovering)
+    {
+        otLogCritPlat("RCP failure detected during RCP recovering, exiting.");
+        DieNow(OT_EXIT_FAILURE);
+    }
+    otLogWarnPlat("RCP failure detected, trying to recover.");
+    mIsRcpRecovering = true;
+
+    mRcpRecoveringState = mState;
```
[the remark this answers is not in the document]
Moandor-y: Done.

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: It still works fine without checking before requesting, the first attempt will fail. Starting recovery earlier is good but also requires another block of code added to every function of get/set/insert/remove. Considering https://github.com/openthread/openthread/pull/5694#discussion_r517845367, is it better?

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: The stored properties are updated only after `Set` returns, so the values restored are old ones.

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: In `HandleValueIs`: <https://github.com/openthread/openthread/blob/7bdc83f8b3afe782dc01a231a51b79579a9aab11/src/lib/spinel/radio_spinel_impl.hpp#L793>

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: There are some subtle differences in these codes, e.g. `RequestV` at <https://github.com/openthread/openthread/blob/7bdc83f8b3afe782dc01a231a51b79579a9aab11/src/lib/spinel/radio_spinel_impl.hpp#L1428> and `Request` at <https://github.com/openthread/openthread/blob/7bdc83f8b3afe782dc01a231a51b79579a9aab11/src/lib/spinel/radio_spinel_impl.hpp#L1464>, `mPropertyFormat` at <https://github.com/openthread/openthread/blob/7bdc83f8b3afe782dc01a231a51b79579a9aab11/src/lib/spinel/radio_spinel_impl.hpp#L1426> and `mExpectedCommand` at <https://github.com/openthread/openthread/blob/7bdc83f8b3afe782dc01a231a51b79579a9aab11/src/lib/spinel/radio_spinel_impl.hpp#L1493>. Also, it may lead to long function names, e.g. `RequestAndWaitWithPropertyFormatV`. Thoughts?

src/lib/spinel/radio_spinel_impl.hpp
simonlingoogle: > The stored properties are updated only after `Set` returns, so the values restored are old ones.

Can we update stored properties before `Set` ? This should produce simpler code.

src/lib/spinel/radio_spinel_impl.hpp
simonlingoogle: This statement is skipped if `HandleRcpUnexpectedReset()` is called.  Do we have another chance to set `mIsReady` when restoring?

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: Then we have to revert the value back if the actual operation fails. Seems more complex.

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: `HandleRcpUnexpectedReset` only marks the failure state. Then when the actual recovery happens `mIsReady` will be set to false first, then set to true when RCP's first message arrives.

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: 2 instances of similar code don't need refactoring. But I can try to merge 4 instances from get/getwithparam into one function, and 6 instances from set/insert/remove into another function.

Moandor-y: Did the refactoring in <https://github.com/openthread/openthread/pull/5694/commits/95880327ca7f92595227079cac7caaed531c738a>. Looks slightly less redundant. I'm OK with both versions.

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: After a second thought I think it's unnecessary. Because the only places could set `mRcpFailed` are in `Process` and get/set/insert/remove operations, `RecoverFromRcpFailure` is called in each of them, so failures will be immediately recovered.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1316,11 +1498,24 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::GetWithParam(spinel_prop
...
+    if (mRcpFailed)
+    {
+        RecoverFromRcpFailure();
+
+        va_start(mPropertyArgs, aFormat);
+        error = RequestWithPropertyFormat(aFormat, SPINEL_CMD_PROP_VALUE_GET, aKey, SPINEL_DATATYPE_DATA_S, aParam,
+                                          aParamSize);
+        va_end(mPropertyArgs);
+
+        VerifyOrDie(!mRcpFailed, OT_EXIT_FAILURE);
```
abtink, line 1516: With current model, seems like we allow one RCP failure (if it happens again while recovering) we fail hard and `DieNow`). 

- Can we relax this to allow `x` number of back-to-back failure before stop (say 3 or 4), instead of 2.
- I see the same pattern of code (i.e. check if RCP failed, recover and the try the same command again) in all different methods (`Get()`, `Set()`, `Insert()`, ...). It would be good can we somehow re-use this pattern/code (we are calling `Insert()` and if it fails, we want to recover and then try `Insert()` again with same params). I don't have a clear suggestion on how we may do it (it's fine for me as is), just raising it so we can think about it.

abtink (approved):
Looks great overall. Thanks.
I have one suggestion/question below (it LGTM as is though)

src/lib/spinel/radio_spinel_impl.hpp
simonlingoogle: Agree.

Moandor-y: Agreed with retrying more than once.

> Another way to set the retry time might be changing the configuration

This looks good to me.

For the reusing of code, I did some refactoring <https://github.com/openthread/openthread/pull/5694#discussion_r518504037>. This seems the best I can do.

Moandor-y: Updated. Now using `OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT` to control max failure count allowed.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1295,11 +1466,22 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Get(spinel_prop_key_t aK
...
+#if OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT > 0
+    if (mRcpFailed)
+    {
+        RecoverFromRcpFailure();
+
+        va_start(mPropertyArgs, aFormat);
+        error = RequestWithPropertyFormatV(aFormat, SPINEL_CMD_PROP_VALUE_GET, aKey, nullptr, mPropertyArgs);
+        va_end(mPropertyArgs);
+
+        VerifyOrDie(!mRcpFailed, OT_EXIT_FAILURE);
```
simonlingoogle, line 1482: What if OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT >= 2 and RCP failed here?

We should restore from this RCP failure too since it's the 2nd fail.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1316,11 +1498,24 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::GetWithParam(spinel_prop
...
+        RecoverFromRcpFailure();
+
+        va_start(mPropertyArgs, aFormat);
+        error = RequestWithPropertyFormat(aFormat, SPINEL_CMD_PROP_VALUE_GET, aKey, SPINEL_DATATYPE_DATA_S, aParam,
+                                          aParamSize);
+        va_end(mPropertyArgs);
+
+        VerifyOrDie(!mRcpFailed, OT_EXIT_FAILURE);
+    }
+#endif
```
simonlingoogle, line 1518: Can we use code like below?

```c++

#if OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT > 0
    for (uint8_t i = 0; i <= OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT; i++)
    {
#endif

        va_start(mPropertyArgs, aFormat);
        error = RequestWithPropertyFormat(aFormat, SPINEL_CMD_PROP_VALUE_GET, aKey, SPINEL_DATATYPE_DATA_S, aParam,
                                          aParamSize);
        va_end(mPropertyArgs);

#if OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT > 0
        if (mRcpFailed)
        {
            RecoverFromRcpFailure();
        } else {
            break;
        }
    }

    VerifyOrDie(!mRcpFailed, OT_EXIT_FAILURE);
#endif
```
The code above is just for demonstration. But it shows that it's possible to somehow reuse the code between `va_start` and `va_end`. 

furthermore, I think it's incorrect to just `VerifyOrDie(!mRcpFailed, OT_EXIT_FAILURE);` after second `Get` or `Set`.

Moandor-y: Agreed. But for `OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT=n` it means we allow at most `n` consecutive failures in a single call to `RecoverFromRcpFailure()`, if we allow `n` calls to `RecoverFromRcpFailure()` then it's at most `n^2` failures in a set/get operation.

I think it's better to change the loop to a `while` loop so that it:

Keeps retrying get/set until one of the following is satisfied:
1. Set/get succeeds without failure.
1. More than `n` failures happen in a single recovery process.

Thoughts?

Moandor-y: Did the change in <https://github.com/openthread/openthread/pull/5694/commits/3b8c25b0a42d6720a91be1e8267ff5019e0327f9>. You can see if it's OK.

src/lib/spinel/radio_spinel_impl.hpp
simonlingoogle: Why do we need to revert? If the operation failed, shouldn't we restore RCP failure then?
If the restoration is successful, we are good. If not, we die. Setting value before actual `Set` removes the necessity to retry `Set`. Thought?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1295,11 +1466,20 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Get(spinel_prop_key_t aK
...
+#if OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT > 0
+        RecoverFromRcpFailure();
+#endif
+        va_start(mPropertyArgs, aFormat);
+        error = RequestWithPropertyFormatV(aFormat, SPINEL_CMD_PROP_VALUE_GET, aKey, nullptr, mPropertyArgs);
+        va_end(mPropertyArgs);
+    }
+#if OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT > 0
+    while (mRcpFailed);
+#endif
```
simonlingoogle, line 1482: nit:
```suggestion
#if OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT > 0
    do
    {
        RecoverFromRcpFailure();
#endif
        va_start(mPropertyArgs, aFormat);
        error = RequestWithPropertyFormatV(aFormat, SPINEL_CMD_PROP_VALUE_GET, aKey, nullptr, mPropertyArgs);
        va_end(mPropertyArgs);
#if OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT > 0
    }
    while (mRcpFailed);
#endif
```

src/lib/spinel/radio_spinel_impl.hpp
simonlingoogle: I am a little bit worried since the new code seems to introduce a potential dead loop.
Nevertheless, I don't see a obvious way to handle this.

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: The operation may fail because of other reasons, for example, the RCP reports an error.

src/lib/spinel/radio_spinel_impl.hpp
simonlingoogle: Maybe we had some misunderstanding. I was actually talking about the same issue as discuessed in https://github.com/openthread/openthread/pull/5694#discussion_r528461928 that there are code redundancies in `Get*` methods. 

I am resolving this since that issue is resolved.

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: If wrong logic causes a dead loop, I think it will be very obvious, and caught by CI tests.

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: Done.

src/lib/spinel/radio_spinel_impl.hpp
simonlingoogle: SGTM

src/lib/spinel/radio_spinel_impl.hpp
simonlingoogle: I can see your point. SGTM. 

But another thought is that since the RCP device is under total control of Host OpenThread, we should expect no failure if RCP is working well. Any error from RCP should indicate a device failure and thus should be fixed by RCP Restoration.

simonlingoogle (approved):
LGTM :+1: 
Thanks

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #5697, opened 2020-10-23
[efr32] avoid asserts during data requests

### LINKED ISSUE #5679: [efr32] radio: asserts during data requests
With very heavy traffic and very heavy MCU load, the following asserts trigger (lines 911, 919, 926):

https://github.com/openthread/openthread/blob/831406ce9055d622a3258a82a1a6ee556530fc49/examples/platforms/efr32mg12/radio.c#L902-L929

1. Git commit id: 831406ce9055d622a3258a82a1a6ee556530fc49
2. IEEE 802.15.4 hardware platform: efr32mg12
3. Build steps: `make -f examples/Makefile-efr32mg12 BOARD=BRD4161A`
4. Network topology: 1 NCP + 10 SEDs. The bug occurs on the NCP/RCP.

For this to happen, the MCU has to be *very heavily loaded*, so that the RAIL interrupt fires late. These asserts are due to the `RAIL_IEEE802154_GetAddress` and `RAIL_IEEE802154_SetFramePending` returning `RAIL_STATUS_INVALID_STATE` because it is too late to modify the outgoing Immediate ACK.

I could submit a PR addressing this, but it's not clear to me if we should inform the upper layers that the frame was acked with FP or not, even if these calls failed.

**Expected behavior** A clear and concise description of what you expected to happen.
No asserts.

**Console/log output** If applicable, add console/log output to help explain your problem.
None.

**Additional context** Add any other context about the problem here.
None.

dismirlian:
This PR fixes #5679.

If the RAIL interrupt fires late (for example, when the MCU is heavily loaded), then `RAIL_IEEE802154_GetAddress` and `RAIL_IEEE802154_SetFramePending` return `RAIL_STATUS_INVALID_STATE` because it is too late to modify the outgoing Immediate ACK. The proposed solution is to avoid asserting and print a warning to make the user aware of such problems.

### DIFF
--- a/examples/platforms/efr32/src/radio.c
+++ b/examples/platforms/efr32/src/radio.c
@@ -889,14 +889,14 @@ exit:
 
 static void ieee802154DataRequestCommand(RAIL_Handle_t aRailHandle)
 {
-    RAIL_Status_t status;
+    RAIL_Status_t status = RAIL_STATUS_NO_ERROR;
 
     if (sIsSrcMatchEnabled)
     {
         RAIL_IEEE802154_Address_t sourceAddress;
 
         status = RAIL_IEEE802154_GetAddress(aRailHandle, &sourceAddress);
-        assert(status == RAIL_STATUS_NO_ERROR);
+        otEXPECT(status == RAIL_STATUS_NO_ERROR);
 
         if ((sourceAddress.length == RAIL_IEEE802154_LongAddress &&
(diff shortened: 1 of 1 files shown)


romacdon:
@schinchani - can you also review this change?

romacdon (approved):

schinchani:
@romacdon  I agree, this change should be okay. Asserting out may not be the best way to handle this scenario. Although, I do think, we should capture this use case and see if can handle the situation better.

schinchani (approved):

jwhui (approved):

PR #5699, opened 2020-10-23
[efr32] use OT_PLAT_RESET_REASON_OTHER for brown-out/EM4 reset

### LINKED ISSUE #5685: [efr32] wrong reset reason
In `otPlatGetResetReason`, the return value implies that the last reset was caussed by an `assert`, when actually it was a brownout or an EM4 reset. I think a better fit would be `OT_PLAT_RESET_REASON_OTHER`.

https://github.com/openthread/openthread/blob/9aa2402753fab14d378d5035748c9c0bfc17173a/examples/platforms/efr32/src/misc.c#L83-L88

https://github.com/openthread/openthread/blob/9aa2402753fab14d378d5035748c9c0bfc17173a/examples/platforms/efr32/src/misc.c#L111-L115

lmnotran:
Closes #5685

CC: @dismirlian @schinchani @wbober


romacdon (approved):

dismirlian:
Looks good! Thanks.

jwhui (approved):

PR #5702, opened 2020-10-24
[cmdline-parser] avoid casting -INT32_MIN directly to int32_t

jwhui:
Fixes the following compiler error:

utils/parse_cmdline.cpp:230:28: runtime error: negation of -2147483648
cannot be represented in type 'int32_t' (aka 'int'); cast to an
unsigned type to negate this value to itself


abtink (approved):
LGTM. Thanks.

PR #5704, opened 2020-10-26
[logging] add optional `otPlatLogLine()` & use it in NCP/CLI

abtink:
This commit adds a new platform function `otPlatLogLine()`. This
function is optional and if not implemented by platform layer, a
default weak implementation is provided and used by the OpenThread
core using `otPlatLog()`.The new function is used by OpenThread core
when the feature `OPENTHREAD_CONFIG_LOG_DEFINE_AS_MACRO_ONLY` is not
enabled (which is the default behavior). In this case, the OT core
itself will prepare a full log line.

This commit also adds implementations of the new platform function for
NCP and CLI modules.

### DIFF
--- a/examples/apps/cli/main.c
+++ b/examples/apps/cli/main.c
@@ -133,13 +133,15 @@ pseudo_reset:
 #if OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_APP
 void otPlatLog(otLogLevel aLogLevel, otLogRegion aLogRegion, const char *aFormat, ...)
 {
-    OT_UNUSED_VARIABLE(aLogLevel);
-    OT_UNUSED_VARIABLE(aLogRegion);
-    OT_UNUSED_VARIABLE(aFormat);
-
     va_list ap;
     va_start(ap, aFormat);
     otCliPlatLogv(aLogLevel, aLogRegion, aFormat, ap);
     va_end(ap);
 }
+
+void otPlatLogLine(otLogLevel aLogLevel, otLogRegion aLogRegion, const char *aLogLine)
+{
+    otCliPlatLogLine(aLogLevel, aLogRegion, aLogLine);
+}
+
 #endif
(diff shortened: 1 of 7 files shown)


examples/apps/cli/main.c
```diff
@@ -133,13 +133,15 @@ int main(int argc, char *argv[])
...
-    OT_UNUSED_VARIABLE(aLogRegion);
-    OT_UNUSED_VARIABLE(aFormat);
-
     va_list ap;
     va_start(ap, aFormat);
     otCliPlatLogv(aLogLevel, aLogRegion, aFormat, ap);
     va_end(ap);
 }
+
+void otPlatLogLine(otLogLevel aLogLevel, otLogRegion aLogRegion, const char *aLogLine)
```
simonlingoogle, line 142: I think the best option might be renaming `otPlatLog` to `otPlatLogFormat` or `otPlatLogf` and let the new function be named `otPlatLog`. 
But it seems to be a huge change to the APIs and OT core.

simonlingoogle (approved):
LGTM :+1:

examples/apps/cli/main.c
abtink: I agree it may sound better (more clear) but as you said it would be a big platform API change.

In general, I think/agree that we should aim to keep existing platform and public OT APIs (name and behavior) as they are as much as possible (permit a change only when we have a very good justification for it). Within OT core we can be a lot more flexible.

include/openthread/platform/logging.h
```diff
@@ -150,6 +150,22 @@ typedef enum otLogRegion
  */
 void otPlatLog(otLogLevel aLogLevel, otLogRegion aLogRegion, const char *aFormat, ...);
 
+/**
+ * This (optional) platform function outputs a prepared log line.
+ *
+ * This platform function is used by OpenThread core when `OPENTHREAD_CONFIG_LOG_DEFINE_AS_MACRO_ONLY` is not enabled
+ * (in this case, the OT core itself will prepare a full log line).
+ *
+ * Note that this function is optional and if not provided by platform layer , a default (weak) implementation is
```
jwhui, line 159:
```suggestion
 * Note that this function is optional and if not provided by platform layer, a default (weak) implementation is
```

jwhui (approved):

PR #5705, opened 2020-10-26
[cmake] use init flags in toolchain files

bukepo:
From https://cmake.org/cmake/help/v3.10/variable/CMAKE_LANG_FLAGS_INIT.html,
`CMAKE_<LANG>_FLAGS_INIT` should be used instead of `CMAKE_<LANG>_FLAGS' in
toolchain files so that CMake can prepend or append content based on the
environment or target platform.


FiveDimensions (approved):
LGTM

jwhui (approved):

Issue #5712, opened 2020-10-26
[ot-cli] no-reset option causes app to crash


latheepanm:
Description:
When launching the ot-cli posix app with the no-reset option as a command line argument, the app crashes immediately.

Command:
`./output/posix/x86_64-unknown-linux-gnu/bin/ot-cli 'spinel+hdlc+uart:///dev/ttyACM2?no-reset'`

Logs:
```
Oct 26 13:39:13 latheepan-Latitude-7400 ./output/posix/x86_64-unknown-linux-gnu/bin/ot-cli[8830]: Running OPENTHREAD/20191113-01338-g12b967cce; POSIX; Oct 26 2020 12:53:47
Oct 26 13:39:13 latheepan-Latitude-7400 ./output/posix/x86_64-unknown-linux-gnu/bin/ot-cli[8830]: Thread version: 2
Oct 26 13:39:15 latheepan-Latitude-7400 ./output/posix/x86_64-unknown-linux-gnu/bin/ot-cli[8830]: [[CRIT]]-PLAT----: WaitResponse() at /home/latheepan/githubs/openthread/src/posix/../../src/lib/spinel/radio_spinel_impl.hpp:1388: RadioSpinelNoRespo
```

bukepo:
It is necessary to reset RCP on start so that RCP and host are in sync.

latheepanm:
@bukepo thank you for that! After reading your comments here and on the PR, it appears that we have misinterpreted/misunderstood the purpose of the no-reset feature (maybe the documentation could highlight that a reset of the RCP is still required through other means?). For our application, it would be desirable to not send a reset to the RCP when an app launches. Is there any other way to ensure that they are in sync? Is there a recommended set/order of commands that should be sent to the RCP?

bukepo:
@latheepanm I don't think it's easy to make host and RCP work properly without sync at start. RCP holds quite a few states, such as short address, extended address, source match table, transmit frame, frame counters and so on. Besides, the spinel protocol also keeps track of next expected transaction id, which is unknown if RCP is not reset. May I know the reason you don't want the reset of RCP?

latheepanm:
@bukepo We are working on a project to add multi-PAN support to the RCP, so that multiple host applications can use a single RCP for different networks concurrently.  (The networks have to be on the same channel.)  We're planning on opening an issue soon to contribute our work on this to OpenThread. 

[...]

PR #5713, opened 2020-10-26
Staging/bugfix/fix posix no reset

latheepanm:
Proposed fix for #5712

### DIFF
--- a/src/lib/spinel/radio_spinel_impl.hpp
+++ b/src/lib/spinel/radio_spinel_impl.hpp
@@ -212,10 +212,13 @@ void RadioSpinel<InterfaceType, ProcessContextType>::Init(bool aResetRadio, bool
     if (aResetRadio)
     {
         SuccessOrExit(error = SendReset());
+        SuccessOrExit(error = WaitResponse());
+        VerifyOrExit(mIsReady, error = OT_ERROR_FAILED);
     }
-
-    SuccessOrExit(error = WaitResponse());
-    VerifyOrExit(mIsReady, error = OT_ERROR_FAILED);
+    else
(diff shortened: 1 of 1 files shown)


latheepanm:
@googlebot I signed it!

romacdon:
@googlebot I signed it!

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -212,10 +212,13 @@ void RadioSpinel<InterfaceType, ProcessContextType>::Init(bool aResetRadio, bool
     if (aResetRadio)
     {
         SuccessOrExit(error = SendReset());
+        SuccessOrExit(error = WaitResponse());
+        VerifyOrExit(mIsReady, error = OT_ERROR_FAILED);
+    }
+    else
```
bukepo, line 218: When `aResetRadio` is false, the host assumes the RCP is reset through other approaches(e.g. reset PIN) instead of spinel, so the reset event is always expected to keep host and RCP in sync. The `no-reset` option is used to avoid double resets.

romacdon: @latheepanm - any thoughts on @bukepo's comment?

latheepanm: @romacdon I actually put my thoughts in #5712 and maybe I should have put it here as well. 

@bukepo For our application, it would be desirable to not send a reset to the RCP when an app launches. Is there any other way to ensure that they are in sync? Is there a recommended set/order of commands that should be sent to the RCP, that is not part of the initialization?

PR #5732, opened 2020-10-28
[clang-tidy] misc-unused-using-decls

jwhui:


abtink (approved):
LGTM. 👍

Irving-cl (approved):
LGTM

librasungirl (approved):

simonlingoogle (approved):
LGTM :+1:

simonlingoogle:
Not directly related by this PR, but I think the `script/make-pretty clang-tidy` is currently fixing `.h` headers in third_party/ directory (specifically mbedtls/cipher.h) by changing NULL to nullptr.

I investigated on this issue for a little while but couldn't find a suitable solution. Maybe we can fix this using the `-header-filter` argument or .clang-tidy configurations. However there seems no obvious solution to it. 

Thoughts? @jwhui

jwhui:
> Not directly related by this PR, but I think the `script/make-pretty clang-tidy` is currently fixing `.h` headers in third_party/ directory (specifically mbedtls/cipher.h) by changing NULL to nullptr.
> 
> I investigated on this issue for a little while but couldn't find a suitable solution. Maybe we can fix this using the `-header-filter` argument or .clang-tidy configurations. However there seems no obvious solution to it.
> 
> Thoughts? @jwhui

@simonlingoogle , I noticed this as well. However, I'm not yet able to determine how to ignore files in `third_party/`. For the `check` option, I had to explicitly grep the output file and exclude paths including `third_party/`.

https://github.com/openthread/openthread/blob/fc159a706f0f1fbdffb526cad8e5847209dd6b14/script/make-pretty#L174

PR #5735, opened 2020-10-28
[type-traits] add 'IsPointer<Type>' and use it in message/dataset

abtink:
This commit adds a new header `common/type_traits.hpp` which provides
`IsPointer<Type>` to check whether a template `Type` is a pointer type
or not. This is then used in `Message` and `Dataset` methods where we
append/set a generic object to check (as an `static_assert`) that the
object type is not a pointer. This adds protection (causing a compile
time error) against potential incorrect use of these methods (i.e.,
passing a pointer instead of a reference to an object by mistake).

### DIFF
--- a/src/core/common/type_traits.hpp
+++ b/src/core/common/type_traits.hpp
@@ -1,0 +1,92 @@
+/*
+ *  Copyright (c) 2020, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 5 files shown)


src/core/common/type_traits.hpp
```diff
@@ -0,0 +1,92 @@
...
+ * This type represents a true value (contains a `true` static `kValue` member variable).
+ *
+ */
+struct TrueValue
+{
+    constexpr static bool kValue = true; ///< true value.
+};
+
+/**
+ * This type represents a true value (contains a `false` static `kValue` member variable).
```
Irving-cl, line 50:
```suggestion
 * This type represents a false value (contains a `false` static `kValue` member variable).
```

Irving-cl (approved):
LGTM.

jwhui (approved):

PR #5741, opened 2020-10-29
[tlvs] add template 'AppendTlv/FindTlv/ReadTlv<ValueType>()'

abtink:
This commit adds new template versions of `AppendTlv()`, `FindTlv()`,
and `ReadTlv()` methods in the `Tlv` class allowing the `ValueType` to
be provided as a template type parameter. This, in turn, helps simplify
their use in the core modules.

----

This PR includes the commit from PR https://github.com/openthread/openthread/pull/5735.

### DIFF
--- a/src/core/backbone_router/bbr_manager.cpp
+++ b/src/core/backbone_router/bbr_manager.cpp
@@ -350,8 +350,8 @@ void Manager::HandleDuaRegistration(const Coap::Message &aMessage, const Ip6::Me
     VerifyOrExit(aMessageInfo.GetPeerAddr().GetIid().IsRoutingLocator(), error = OT_ERROR_DROP);
     VerifyOrExit(aMessage.IsConfirmablePostRequest(), error = OT_ERROR_PARSE);
 
-    SuccessOrExit(error = Tlv::FindTlv(aMessage, ThreadTlv::kTarget, &target, sizeof(target)));
-    SuccessOrExit(error = Tlv::FindTlv(aMessage, ThreadTlv::kMeshLocalEid, &meshLocalIid, sizeof(meshLocalIid)));
+    SuccessOrExit(error = Tlv::FindTlv(aMessage, ThreadTlv::kTarget, target));
+    SuccessOrExit(error = Tlv::FindTlv(aMessage, ThreadTlv::kMeshLocalEid, meshLocalIid));
 
 #if OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE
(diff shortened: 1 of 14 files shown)


src/core/backbone_router/bbr_manager.cpp
```diff
@@ -428,7 +428,7 @@ void Manager::SendDuaRegistrationResponse(const Coap::Message &      aMessage,
     SuccessOrExit(message->SetPayloadMarker());
 
     SuccessOrExit(Tlv::AppendUint8Tlv(*message, ThreadTlv::kStatus, aStatus));
-    SuccessOrExit(Tlv::AppendTlv(*message, ThreadTlv::kTarget, &aTarget, sizeof(aTarget)));
+    SuccessOrExit(Tlv::AppendTlv(*message, ThreadTlv::kTarget, aTarget));
```
simonlingoogle, line 431: Instead of using template <typename ValueType> methods, is it better (and possible) to define methods like 
Tlv::AppendTlv\<ThreadTlv::kTarget\>(*message, aTarget) ? I think this can achieve better type checking. 

Thoughts?

abtink: It is an interesting idea. I suggest we consider it as a potential future enhancement.

I am on the fence for this. On one hand, I like the type checking safety that this can add. On the other hand, I feel it may add a lot of complexity.

- Currently we have a common version of these method in `Tlv` base class (which is then shared/used by all)
- I think for this we would need a version like`AppendTlv<uint8_t TlvType, typename ValueType>` and then we need to add a template specialization for every simple TLV type (with its related `ValueType`)
- Also since the TLV integr type value are re-used by different TLV categories (e.g. TLV type value `0` is used as `Mle::Tlv::kSourceAddress = 0` and `MeshCop::Tlv::kChannel = 0` and `ThreadTlv::kTarget = 0`, etc.),  I don't think we can have a common template methods in `Tlv` base class and would need to have the template methods (re)defined by each TLV category (one set of methods in `Mle::Tlv`, one in `MeshCop::Tlv`, etc).

simonlingoogle: I prefer defining template methods on each TLV category (e.g. Mle::TLv, ThreadTlv) so as to differentiate equal TLV types. 
But we can certainly leave it for further enhancement. 

Furthermore, I think this should be implemented based on `AppendTlv<uint8_t TlvType, typename ValueType>`.

simonlingoogle (approved):
LGTM :+1: 

I think these methods would be more convenient to use.

abtink:
Closing this PR in favor of new model from PR https://github.com/openthread/openthread/pull/5755.

Issue #5689, opened 2020-10-22
[efr32] radio: assert during data request, unrelated to #5679


dismirlian:
**Note**: this `assert` is unrelated to #5679.

**Describe the bug** A clear and concise description of what the bug is.
With very heavy traffic, the following assert triggers (line 747):

https://github.com/openthread/openthread/blob/831406ce9055d622a3258a82a1a6ee556530fc49/examples/platforms/efr32mg12/radio.c#L745-L750

**To Reproduce** Information to reproduce the behavior, including:

1. Git commit id: 831406ce9055d622a3258a82a1a6ee556530fc49
2. IEEE 802.15.4 hardware platform: efr32mg12
3. Build steps: `make -f examples/Makefile-efr32mg12 BOARD=BRD4161A`
4. Network topology: 1 NCP + 10 SEDs. The bug occurs on the NCP/RCP.

**Expected behavior** A clear and concise description of what you expected to happen.
No asserts. The queue used to cache the packets sent with FP set is of the same size of the internal metadata FIFO. Quoting RAIL documentation:
> Supplementing the receive FIFO used for packet data, the RAIL library also allocates an internal packet metadata FIFO capable of holding up to 16 packets' worth of metadata. An entry in this FIFO is made on every packet completion (successful or not) signaled to the application. If the application calls RAIL_HoldRxPacket() that entry is held along with any of its packet data in the receive FIFO.

**Console/log output** If applicable, add console/log output to help explain your problem.
None.

**Additional context** Add any other context about the problem here.
I increased `ACKED_WITH_FP_SLOTS` to 32, and I haven't seen the problem again, even under heavy traffic. However, I would like to be *certain* that this is a safe value. Maybe Silabs can shed a light on this?

romacdon:
@dismirlian - thanks for reporting this issue.  We are looking into it.

suveshpratapa:
@dismirlian Given your testing with 10 sleepy end devices under heavy traffic, I'm inclined to say this value is safe.  You can go ahead and submit a PR.  We will also revisit our implementation of this `sAckedWithFPFifo` queue in our PAL soon.

As an aside, I'm curious -- what are the data rates / polling rates in your "heavy traffic" testing?

dismirlian:
[...]

PR #5742, opened 2020-10-29
[efr32] radio: increase FIFO size to allow handling of heavy traffic

dismirlian:
This PR increases the FIFO size used for storing the handles of the data request packets that were ACKed with the frame pending bit set. This is a temporary "fix" for #5689; the actual reason for the assert is being investigated by Silabs.

The logic behind the original implementation is here: #4484.


suveshpratapa (approved):
LGTM

romacdon (approved):

jwhui (approved):

PR #5744, opened 2020-10-29
[efr32] call mbedtls_hardware_poll directly in otPlatEntropyGet()

### LINKED ISSUE #4649: [efr32] Entropy questions
Hi all. I have some questions regarding the efr32 platforms' entropy implementation.

1. Why is the ADC used instead of the TRNG in efr32mg12 and efr32mg13? I know that there is an erratum associated with the xg13 chip revisions A~C, but not from D onwards. The mg21 platform abstraction uses the TRNG. I'd like to propose to use the TRNG for all platforms, with the ADC only as fallback. This should free the ADC for the application to use.

2. The entropy collection code for the mg21 is as follows:
https://github.com/openthread/openthread/blob/fdcb8553b12216d53ed3c5249a97ebe7ff2e10b7/examples/platforms/efr32mg21/entropy.c#L40-L63
The `otPlatEntropyGet` function discards 3 out of 4 entropy bytes. Why? Couldn't we call directly `mbedtls_hardware_poll`?

3. In the mg21 code, shouldn't we check `ouputLen`? Specifically, seeing the mbedtls/sl_crypto/trng.c file, the `mbedtls_trng_poll` call could fail, setting `outputLen` to 0.

Thanks!

lmnotran:
Closes #4649

### DIFF
--- a/examples/platforms/efr32/src/entropy.c
+++ b/examples/platforms/efr32/src/entropy.c
@@ -33,30 +33,27 @@
  */
 
 #include <openthread/platform/entropy.h>
-
 #include "mbedtls/entropy_poll.h"
 #include "utils/code_utils.h"
-
-static uint32_t randomUint32Get(void)
-{
-    uint32_t random   = 0;
-    size_t   ouputLen = 0;
-
-    mbedtls_hardware_poll(NULL, (unsigned char *)&random, sizeof(random), &ouputLen);
-
-    return random;
-}
 
 otError otPlatEntropyGet(uint8_t *aOutput, uint16_t aOutputLength)
 {
     otError error = OT_ERROR_NONE;
-
     otEXPECT_ACTION(aOutput, error = OT_ERROR_INVALID_ARGS);
 
-    for (uint16_t length = 0; length < aOutputLength; length++)
+    size_t outputLen = 0;
+
+    for (size_t partialLen = 0; outputLen < aOutputLength; outputLen += partialLen)
     {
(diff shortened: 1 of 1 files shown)


jwhui:
/cc/ @dismirlian

romacdon (approved):

examples/platforms/efr32/src/entropy.c
```diff
@@ -33,31 +33,29 @@
...
+        const uint16_t remaining = aOutputLength - outputLen;
+        partialLen               = 0;
+
+        // Non-zero return values for mbedtls_hardware_poll() signify an error has occurred
+        otEXPECT_ACTION(0 == mbedtls_hardware_poll(NULL, &aOutput[outputLen], remaining, &partialLen),
+                        error = OT_ERROR_FAILED);
+        otEXPECT_ACTION(partialLen > 0, error = OT_ERROR_FAILED);
     }
 
+    otEXPECT_ACTION(outputLen == aOutputLength, error = OT_ERROR_FAILED);
```
dismirlian, line 57: Isn't this redundant? The above loop breaks when `outputLen == aOutputLength`. Also, it exits directly on error, right?

dismirlian:
Thanks @lmnotran.

examples/platforms/efr32/src/entropy.c
lmnotran: You're definitely correct!

jwhui (approved):

PR #5745, opened 2020-10-30
[nrf528xx] allow users to implement power setting functions

zhanglongxia:
When users add a FEM after the radio chip, the actual output power is
determined by both the radio chip and FEM. The radio driver only provide
a function to set the radio output power. This PR adds Macros and functions
to allow users to implement the power setting functions by themselves.

### DIFF
--- a/examples/platforms/nrf528xx/src/platform-nrf5.h
+++ b/examples/platforms/nrf528xx/src/platform-nrf5.h
@@ -197,4 +197,35 @@ otError nrf5SdErrorToOtError(uint32_t aSdError);
 void nrf5SdSocFlashProcess(uint32_t aEvtId);
 #endif // SOFTDEVICE_PRESENT
 
+#if OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
+/**
+ * Initialization of custom FEM.
+ *
+ */
+void nrf5RadioCustomFemInit(void);
+
+/**
+ * Deinitialization of custom FEM.
+ *
+ */
+void nrf5RadioCustomFemDeinit(void);
+
+/**
+ * Enable custom FEM.
+ *
+ */
+void nrf5RadioCustomFemEnable(void);
+
+/**
+ * Disable custom FEM.
+ *
+ */
+void nrf5RadioCustomFemDisable(void);
+
+/**
+ * Adjust the FEM output power based on the given channel.
+ *
+ */
+void nrf5RadioCustomFemAdjustPower(uint8_t aChannel, bool aIsTransmit);
+#endif // OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
 #endif // PLATFORM_NRF5_H_
--- a/examples/platforms/nrf528xx/src/platform-nrf5.h
+++ b/examples/platforms/nrf528xx/src/platform-nrf5.h
@@ -225,6 +225,9 @@ void nrf5RadioCustomFemDisable(void);
 /**
  * Adjust the FEM output power based on the given channel.
  *
+ * @param[in]    aChannel      Radio channel.
+ * @param[in]    aIsTransmit   TRUE to begin transmit sequence or FALSE to begin receive sequence.
+ *
  */
 void nrf5RadioCustomFemAdjustPower(uint8_t aChannel, bool aIsTransmit);
 #endif // OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
--- a/examples/platforms/nrf528xx/src/platform-nrf5.h
+++ b/examples/platforms/nrf528xx/src/platform-nrf5.h
@@ -197,39 +197,4 @@ otError nrf5SdErrorToOtError(uint32_t aSdError);
 void nrf5SdSocFlashProcess(uint32_t aEvtId);
 #endif // SOFTDEVICE_PRESENT
 
-#if OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
-/**
- * Initialization of custom FEM.
- *
- */
-void nrf5RadioCustomFemInit(void);
-
-/**
- * Deinitialization of custom FEM.
- *
- */
-void nrf5RadioCustomFemDeinit(void);
-
-/**
- * Enable custom FEM.
- *
- */
-void nrf5RadioCustomFemEnable(void);
-
-/**
- * Disable custom FEM.
- *
- */
-void nrf5RadioCustomFemDisable(void);
-
-/**
- * Adjust the FEM output power based on the given channel.
- *
- * @param[in]    aChannel      Radio channel.
- * @param[in]    aIsTransmit   TRUE to begin transmit sequence or FALSE to begin receive sequence.
- *
- */
-void nrf5RadioCustomFemAdjustPower(uint8_t aChannel, bool aIsTransmit);
-#endif // OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
-
 #endif // PLATFORM_NRF5_H_
--- a/examples/platforms/nrf528xx/src/platform-nrf5.h
+++ b/examples/platforms/nrf528xx/src/platform-nrf5.h
@@ -228,4 +228,5 @@ void nrf5RadioCustomFemDisable(void);
  */
 void nrf5RadioCustomFemAdjustPower(uint8_t aChannel, bool aIsTransmit);
 #endif // OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
+
 #endif // PLATFORM_NRF5_H_
--- a/examples/platforms/nrf528xx/src/radio.c
+++ b/examples/platforms/nrf528xx/src/radio.c
@@ -722,7 +722,7 @@ otError otPlatRadioSetTransmitPower(otInstance *aInstance, int8_t aPower)
 
     return OT_ERROR_NONE;
 }
-#endif
+#endif // !OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
 
 otError otPlatRadioGetCcaEnergyDetectThreshold(otInstance *aInstance, int8_t *aThreshold)
 {
--- a/examples/platforms/nrf528xx/src/radio.c
+++ b/examples/platforms/nrf528xx/src/radio.c
@@ -106,7 +106,7 @@ static otInstance *  sInstance = NULL;
 static otRadioFrame sAckFrame;
 static bool         sAckedWithFramePending;
 
-#if !OPENTHREAD_CONFIG_ENABLE_PLATFORM_POWER_CUSTOM_SOURCE
+#if !OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
 static int8_t sDefaultTxPower;
 #endif
 static int8_t sLnaGain = 0;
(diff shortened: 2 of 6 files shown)


examples/platforms/nrf528xx/src/platform-nrf5.h
```diff
@@ -197,4 +197,35 @@ otError nrf5SdErrorToOtError(uint32_t aSdError);
...
+ *
+ */
+void nrf5RadioCustomFemDisable(void);
+
+/**
+ * Adjust the FEM output power based on the given channel.
+ *
+ */
+void nrf5RadioCustomFemAdjustPower(uint8_t aChannel, bool aIsTransmit);
+#endif // OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
```
edmont, line 230: Nit: new line here.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -106,7 +106,9 @@ static otInstance *  sInstance = NULL;
 static otRadioFrame sAckFrame;
 static bool         sAckedWithFramePending;
 
+#if !OPENTHREAD_CONFIG_ENABLE_PLATFORM_POWER_CUSTOM_SOURCE
```
edmont, line 109:
```suggestion
#if !OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
```

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -660,6 +694,7 @@ otError otPlatRadioEnergyScan(otInstance *aInstance, uint8_t aScanChannel, uint1
     return OT_ERROR_NONE;
 }
 
+#if !OPENTHREAD_CONFIG_ENABLE_PLATFORM_POWER_CUSTOM_SOURCE
```
edmont, line 697:
```suggestion
#if !OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
```

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -660,6 +694,7 @@ otError otPlatRadioEnergyScan(otInstance *aInstance, uint8_t aScanChannel, uint1
...
 otError otPlatRadioGetTransmitPower(otInstance *aInstance, int8_t *aPower)
```
edmont, line 698: These APIs are required by OpenThread and shouldn't be hidden behind the macro.

edmont (changes requested):
Thanks @zhanglongxia. Apart from my comments in the code I think that `OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE` should be documented in the platforms' READMEs.

examples/platforms/nrf528xx/src/radio.c
zhanglongxia: Updated.

examples/platforms/nrf528xx/src/radio.c
zhanglongxia: Updated.

examples/platforms/nrf528xx/src/radio.c
zhanglongxia: If the user defines the macro `OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE `,  he should implement the function `otPlatRadioGetTransmitPower ()` by himself, just like the function `otPlatRadioGetIeeeEui64()`. Thoughts?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -457,6 +480,17 @@ otError otPlatRadioTransmit(otInstance *aInstance, otRadioFrame *aFrame)
...
 
+#if OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
+    if (nrf_802154_state_get() == NRF_802154_STATE_SLEEP)
+    {
+        // Enable FEM before RADIO leaving SLEEP state.
+        nrf5RadioCustomFemEnable();
+    }
+
+    // Adjust the output power based on the given channel.
+    nrf5RadioCustomFemAdjustPower(aFrame->mChannel, true);
```
LuDuda, line 491: I might misunderstand the reason for this function but Radio Driver implements FAL layer. Wouldn't be better to replace `nrf_802154_fal_tx_power_get` function instead in your build? In the current implementation, I think that race conditions are possible when switching between TX/TX-ACK.

https://github.com/openthread/openthread/blob/master/third_party/NordicSemiconductor/drivers/radio/fal/nrf_802154_fal.c#L41

Possibly we can make the WEAK implementation of default function.

LuDuda (commented):
Thanks @zhanglongxia 

Radio Driver implements FEM/FAL abstraction layer. Wouldn't be better to add the third implementation for FEM next to Skyworks and nRF21540 that we have now?

examples/platforms/nrf528xx/src/platform-nrf5.h
```diff
@@ -197,4 +197,36 @@ otError nrf5SdErrorToOtError(uint32_t aSdError);
...
+ * Disable custom FEM.
+ *
+ */
+void nrf5RadioCustomFemDisable(void);
+
+/**
+ * Adjust the FEM output power based on the given channel.
+ *
+ */
+void nrf5RadioCustomFemAdjustPower(uint8_t aChannel, bool aIsTransmit);
```
Irving-cl, line 229: Add doc for the parameter?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -687,6 +722,7 @@ otError otPlatRadioSetTransmitPower(otInstance *aInstance, int8_t aPower)
 
     return OT_ERROR_NONE;
 }
+#endif
```
Irving-cl, line 725:
```suggestion
#endif // !OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
```
More than 20 lines between the `#if`.

zhanglongxia:
@LuDuda  The Nordic radio driver has implemented a FEM abstraction layer and the FEM abstraction layer only support  FEM pin PA, LNA and PDN. The FEM is using the pin CPS, CHL in our product. We have  asked Nordic to add CPS and CHL pin control logic to the Nordic’s FEM abstraction layer, but Nordic seems doesn’t want to do it ([JIRA](https://nrfsupport.atlassian.net/browse/NRF40-79?focusedCommentId=11847)).  So we added these pin control logic in our internal source code.  Besides,  the FEM abstraction layer only control the FEM pins, it doesn’t set the FEM gain. We add these FEM  functions to update the radio power and FEM gain in the radio driver.

examples/platforms/nrf528xx/src/radio.c
zhanglongxia: The gain of the FEM is not a fixed value. The actual FEM gain is different on different channel. In order to ensure that the antenna  output power meets the FCC requirements, we need to adjust the radio output power and FEM gain on each channel.The function `nrf_802154_fal_tx_power_get ` only gets the radio output power and it can’t update the FEM gain. The function `nrf5RadioCustomFemAdjustPower ` is used to set the radio output power and FEM gain. The FEM gain remains the same if we don’t change it. The radio output power is maintained by the radio driver library after we set it. Could you provide more detailed info about the race conditions? I pinged a SED from a router with 1000 bytes length for 15000 times, I didn’t see the radio driver crash.

examples/platforms/nrf528xx/src/platform-nrf5.h
zhanglongxia: Updated

examples/platforms/nrf528xx/src/radio.c
zhanglongxia: Updated

examples/platforms/nrf528xx/src/radio.c
LuDuda: Okay. When it comes to race conditions, I was more referring to the situation when we set output power for TX on channel X, but in the meantime, Radio Driver automatically replies with an ACK on channel Y with configured power for channel X. Is this possible?

zhanglongxia: Thanks for your detailed info about the race condition. I understand your concern now. The race condition does exist. Your suggestion about using `nrf_802154_fal_tx_power_get` is great. After I verified your suggestion, I will update the code.

examples/platforms/nrf528xx/src/platform-nrf5.h
```diff
@@ -197,4 +197,39 @@ otError nrf5SdErrorToOtError(uint32_t aSdError);
 void nrf5SdSocFlashProcess(uint32_t aEvtId);
 #endif // SOFTDEVICE_PRESENT
 
+#if OPENTHREAD_CONFIG_ENABLE_PLATFORM_FEM_CUSTOM_SOURCE
+/**
+ * Initialization of custom FEM.
+ *
+ */
+void nrf5RadioCustomFemInit(void);
```
bukepo, line 205: can this be put into platform-fem.h? And maybe we can make the prefix shorter as `nrf5Fem`. Then refactor the built-in fem code with the new APIs. Thoughts?

zhanglongxia: Yes. We can update the code likes that.

LuDuda:
> @LuDuda The Nordic radio driver has implemented a FEM abstraction layer and the FEM abstraction layer only support FEM pin PA, LNA and PDN. The FEM is using the pin CPS, CHL in our product. We have asked Nordic to add CPS and CHL pin control logic to the Nordic’s FEM abstraction layer, but Nordic seems doesn’t want to do it ([JIRA](https://nrfsupport.atlassian.net/browse/NRF40-79?focusedCommentId=11847)). So we added these pin control logic in our internal source code. Besides, the FEM abstraction layer only control the FEM pins, it doesn’t set the FEM gain. We add these FEM functions to update the radio power and FEM gain in the radio driver.

Thanks @zhanglongxia. @hubertmis will comment on Radio Driver part as soon as possible.

examples/platforms/nrf528xx/src/radio.c
zhanglongxia: I have moved all FEM APIs to the header file `platform-fem.h`, so that we don't need to add an extra macro to control the FEM API implementation. User can just implement all the FEM APIs in their own `fem.c`.  I also set the function `nrf_802154_fal_tx_power_get` to a weak function according to your suggestion.

examples/platforms/nrf528xx/src/platform-nrf5.h
zhanglongxia: Updated

hubertmis (approved):
Looks good.
I confirm what @LuDuda mentioned. We are working on implementation of PA/LNA power setting in PA/LNA module used by 802.15.4 driver. Our goal is to expose API where OT would use functions like `set_tx_power(requested_antenna_power)` and 802.15.4 itself would manage chip TX power and external PA to get antenna power as close to requested as possible. However it is still work in progress.

I think that solution proposed in this PR is valid. It could be replaced later with more portable one inside 802.15.4 driver.

LuDuda (approved):
LGTM 👍 Thanks!

bukepo (approved):
LGTM:+1:

brianljt (approved):
LGTM

edmont (approved):
Wouldn't be desirable to document the usage of `OPENTHREAD_CONFIG_ENABLE_PLATFORM_POWER_CUSTOM_SOURCE` in the platforms' README?

zhanglongxia:
@edmont  The function `nrf_802154_tx_power_set` only update the internal power variable, it's not really setting the output power. We are using the function `nrf_802154_fal_tx_power_get` to update the output power. So I removed the macro `OPENTHREAD_CONFIG_ENABLE_PLATFORM_POWER_CUSTOM_SOURCE ` from the radio driver.

abtink (approved):
LGTM.

Irving-cl (approved):
LGTM :+1:

jwhui (approved):

PR #5749, opened 2020-10-30
[logging] correct the log level and region settings

zhanglongxia:


bukepo (approved):
LGTM:+1:

simonlingoogle (approved):
LGTM 👍 

I noticed that logs are using multiple [], e.g. `[[DEBG]]-CLI-------`, but didn't take a deep look.

src/core/common/logging.hpp
```diff
@@ -53,12 +53,12 @@ extern "C" {
...
-#define _OT_LEVEL_WARN_PREFIX "[WARN]"
-#define _OT_LEVEL_NOTE_PREFIX "[NOTE]"
-#define _OT_LEVEL_INFO_PREFIX "[INFO]"
-#define _OT_LEVEL_DEBG_PREFIX "[DEBG]"
+#define _OT_LEVEL_NONE_PREFIX "NONE"
+#define _OT_LEVEL_CRIT_PREFIX "CRIT"
+#define _OT_LEVEL_WARN_PREFIX "WARN"
+#define _OT_LEVEL_NOTE_PREFIX "NOTE"
+#define _OT_LEVEL_INFO_PREFIX "INFO"
+#define _OT_LEVEL_DEBG_PREFIX "DEBG"
```
abtink, line 61: Thanks 👍

src/core/common/logging.cpp
```diff
@@ -96,13 +98,13 @@ static void Log(otLogLevel  aLogLevel,
...
 
-#if OPENTHREAD_CONFIG_LOG_PREPEND_REGION
         IgnoreError(logString.Append("[%s]", levelStr));
-#endif
     }
 #endif // OPENTHREAD_CONFIG_LOG_PREPEND_LEVEL
 
+#if OPENTHREAD_CONFIG_LOG_PREPEND_REGION
     IgnoreError(logString.Append("%s", aRegionPrefix));
+#endif
```
abtink, line 107: I think we may need this in general case (without `#if OPENTHREAD_CONFIG_LOG_PREPEND_REGION` check).

When `PREPEND_REGION` is not enabled but `PREPREND_LEVEL` is enabled, the prefix string becomes just `_OT_REGION_SUFFIX ` which would be `":"`. This is similar to how the macors work.

abtink (approved):
Overall looks good. Thanks 👍 

I just mentioned below, I think we may not need the check `#if OPENTHREAD_CONFIG_LOG_PREPEND_REGION`

src/core/common/logging.hpp
```diff
@@ -53,12 +53,12 @@ extern "C" {
  *
  */
 #if OPENTHREAD_CONFIG_LOG_PREPEND_LEVEL
-#define _OT_LEVEL_NONE_PREFIX "[NONE]"
-#define _OT_LEVEL_CRIT_PREFIX "[CRIT]"
-#define _OT_LEVEL_WARN_PREFIX "[WARN]"
-#define _OT_LEVEL_NOTE_PREFIX "[NOTE]"
-#define _OT_LEVEL_INFO_PREFIX "[INFO]"
-#define _OT_LEVEL_DEBG_PREFIX "[DEBG]"
```
abtink, line 61: Looking/thinking more about it, I think it may be better to keep the `[]` here and remove it from the function where we append the log string

These constants are used by both macro implementation and function one, so if we remove the `[ ]` here we need to update the macor implementation as well to add `[ ]`

src/core/common/logging.cpp
```diff
@@ -96,13 +98,13 @@ static void Log(otLogLevel  aLogLevel,
             break;
         }
 
-#if OPENTHREAD_CONFIG_LOG_PREPEND_REGION
         IgnoreError(logString.Append("[%s]", levelStr));
```
abtink, line 101: Let's remove the `[ ]` here.

src/core/common/logging.cpp
zhanglongxia: updated

src/core/common/logging.hpp
zhanglongxia: updated

src/core/common/logging.cpp
zhanglongxia: updated

librasungirl (approved):

abtink (approved):
Thanks. LGTM. 👍

jwhui (approved):

PR #5752, opened 2020-11-01
[github-actions] add bootstrap test

simonlingoogle:
This commit adds the bootstrap job in Build workflow to make sure that `script/bootstrap` works on both Ubuntu and macOS.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,6 +40,17 @@ jobs:
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
       if: "github.ref != 'refs/heads/master'"
 
+  bootstrap:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-10.15]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Bootstrap
+        run: |
+          script/bootstrap
+
   pretty:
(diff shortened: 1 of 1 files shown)


simonlingoogle:
The bootstrap jobs are now passed. @jwhui

.github/workflows/build.yml
```diff
@@ -40,6 +40,17 @@ jobs:
...
+  bootstrap:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-10.15]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Bootstrap
+        run: |
+          script/bootstrap
```
jwhui, line 52: What about leveraging `script/bootstrap` to bootstrap some of the existing tests so that we can validate includes everything that is needed?

simonlingoogle: In general, I think it's a good idea to use script/bootstrap rather than installing packages in each github action job. 

I think we didn't use it because `script/bootstrap` installs more tools than what each github action job needs.
Running `script/bootstrap` also slows down each test (it takes about 3 minutes in ubuntu github action).

Do you think it worth the cost?

jwhui: I was thinking of applying `script/bootstrap` only to a small subset of tests, not necessarily to all tests.

PR #5762, opened 2020-11-03
[mle] add helper 'ReadFrameCounters()' to read Link/MLE Frame Counters

abtink:
This commits adds a common helper method to read Link and/or MLE Frame
Counters from a message. The Link Frame Counter TLV is expected to be
present in the message. The MLE Frame Counter TLV, however, is optional
and when not present the Link Frame Counter is also used as MLE Frame
Counter value.


Irving-cl (approved):
LGTM :+1:

jwhui (approved):

PR #5763, opened 2020-11-03
[style] fix build with NDEBUG

bukepo:
`assert()` is not defined when `NDEBUG` is defined.

### DIFF
--- a/script/check-arm-build-cmake
+++ b/script/check-arm-build-cmake
@@ -98,6 +98,8 @@ build_samr21()
 
 main()
 {
+    export CPPFLAGS="${CPPFLAGS:-} -DNDEBUG"
+
     if [[ $# == 0 ]]; then
         build_cc1352
         build_cc2538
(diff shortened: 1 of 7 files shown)


script/check-arm-build-cmake
```diff
@@ -98,6 +98,8 @@ build_samr21()
 
 main()
 {
+    export CPPFLAGS="${CPPFLAGS:-} -DNDEBUG"
```
LuDuda, line 101: Do we want to check on CI build only with `NDEBUG` set? From our experience, the majority of customers keep asserts in the deployed product anyway.

LuDuda (approved):
Thanks :+1: One comment below.

script/check-arm-build-cmake
bukepo: This script is only used for CI check, and `script/cmake-build` doesn't have this flag. Adding test for the case without `NDEBUG` will double the check time. Since build without `NDEBUG` can easily be catched during daily work. I think it's fine to just check build with `NDEBUG`. Thoughts?

LuDuda: Okay let's go in this path. Thanks.

jwhui (approved):

PR #5769, opened 2020-11-04
[linked-list] add an overload for LinkedList::FindMatching for convenience

Irving-cl:
This PR adds an overload version for `LinkedList::FindMatching` which doesn't require `prev` as parameter. In many cases we don't need to get the previous item's pointer when finding in a list. This version would help in these cases.


simonlingoogle (approved):
LGTM :+1:

abtink (approved):
LGTM. 👍

jwhui (approved):

Issue #5773, opened 2020-11-05
[github-actions] expect/tun-realm-local-multicast.exp fails intermittently


jwhui:
**Describe the bug**

tests/scripts/expect/tun-realm-local-multicast.exp fails intermittently

**Console/log output**

Log archive: [logs_36395.zip](https://github.com/openthread/openthread/files/5492203/logs_36395.zip)

```
# 1 rcp
spawn /home/runner/work/openthread/openthread/build/cmake/openthread-posix-1.1/src/posix/ot-cli spinel+hdlc+uart:///home/runner/work/openthread/openthread/build/cmake/openthread-simulation-1.1/examples/apps/ncp/ot-rcp?forkpty-arg=1
routerselectionjitter 1
Done

> panid 0xface
Done

> ifconfig up
Done

> thread start
Done

> state
detached

Done

> state
detached

Done

> state
detached

Done

> state
leader

Done

> extaddr
9284c96b9a2aa25e

Done

> 
# 2 cli
spawn /home/runner/work/openthread/openthread/build/cmake/openthread-simulation-1.1/examples/apps/cli/ot-cli-ftd 2
routerselectionjitter 1
routerselectionjitter 1
Done
> panid 0xface
Done
> ifconfig up
Done
> thread start
Done
> state
detached
Done
> state
child
Done
> state
router
Done
> 
# 3 cli
spawn /home/runner/work/openthread/openthread/build/cmake/openthread-simulation-1.1/examples/apps/cli/ot-cli-ftd 3
routerselectionjitter 1
Done
> panid 0xface
Done
> macfilter addr add 9284c96b9a2aa25e
Done
> macfilter addr denylist
Done
> ifconfig up
Done
> thread start
Done
> state
detached
Done
> state
detached
Done
> state
router
Done
> extaddr
8acae7de11e9c473
Done
> 
# 4 cli
spawn /home/runner/work/openthread/openthread/build/cmake/openthread-simulation-1.1/examples/apps/cli/ot-cli-ftd 4
routerselectionjitter 1
Done
> panid 0xface
Done
> mode rn
Done
> macfilter addr add 8acae7de11e9c473
Done
> macfilter addr allowlist
Done
> ifconfig up
Done
> thread start
Done
> state
detached
Done
> state
detached
Done
> state
child
Done
> ipaddr mleid
fdde:ad00:beef:0:f55a:fcc0:f244:81fe
Done
> 
# 1
[...]

PR #5776, opened 2020-11-06
[tests] fix tun-realm-local-multicast.exp

librasungirl:
Given the linear topology in this test, this commit
introduces wait time so that the Thread Links are
synced well before ping happens.

### DIFF
--- a/tests/scripts/expect/tun-realm-local-multicast.exp
+++ b/tests/scripts/expect/tun-realm-local-multicast.exp
@@ -113,6 +113,7 @@ wait_for "state" "child"
 set mleid4 [get_mleid]
 
 switch_node 1
+sleep 10
 
 send "ping ${mleid4}\n"
 expect "Done"


librasungirl:
To fix #5773

tests/scripts/expect/tun-realm-local-multicast.exp
```diff
@@ -113,6 +113,7 @@ wait_for "state" "child"
 set mleid4 [get_mleid]
 
 switch_node 1
+sleep 10
```
librasungirl, line 116: Always pass during local ~ 60 runs.

bukepo (approved):
LGTM:+1: Thanks!

tests/scripts/expect/tun-realm-local-multicast.exp
simonlingoogle: This is fine as a quick fix. But should we improve the performance of OT because 10s without reachability seems to be an issue?

librasungirl: See the discussion in https://github.com/openthread/openthread/issues/5773#issuecomment-722865993
Feel free to share your thoughts there.

jwhui (approved):

PR #5778, opened 2020-11-06
[github-actions] remove coverage artifacts after uploading

Moandor-y:
This removes artifacts when they are no longer needed, saves storage space.


.github/workflows/posix.yml
```diff
@@ -303,3 +303,6 @@ jobs:
     - name: Upload Coverage
       run: |
         script/test upload_codecov
+    - name: Remove Artifacts
+      run: |
+        GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }} script/remove_coverage_artifacts.py
```
simonlingoogle, line 308: Why not remove artifacts in `upload_codeov`?

script/remove_coverage_artifacts.py
```diff
@@ -0,0 +1,80 @@
...
+        url = f'https://api.github.com/repos/openthread/openthread/actions/runs/{_GITHUB_RUN_ID}/artifacts'
+        logging.info('GET %s', url)
+        with urllib.request.urlopen(urllib.request.Request(
+                url=url,
+                headers=_HEADERS,
+        )) as response:
+            result = json.loads(response.read().decode())
+            logging.debug("Result: %s", result)
+
+            artifacts = result['artifacts']
```
simonlingoogle, line 57: move the code below out of the `with` since we have already got the `result`.

script/remove_coverage_artifacts.py
```diff
@@ -0,0 +1,80 @@
...
+            if len(artifacts) < artifact_count:
+                logging.info('Waiting for artifacts...')
+                time.sleep(10)
+                continue
```
simonlingoogle, line 61: what's the intention here?

script/remove_coverage_artifacts.py
simonlingoogle, line 57:
```suggestion
            artifacts = [artifact for artifact in result['artifacts'] if artifact['name'].startswith('cov-')]
```
should only count `cov-` artifacts?

script/remove_coverage_artifacts.py
Moandor-y: I thought the reason it didn't show any artifacts is latency, but I was wrong. No artifacts can be found this way before all jobs in the workflow finishes. We must take a totally different approach.

Moandor-y:
Because GitHub API can't tell if the request and the target artifact are from the same PR, it doesn't allow workflows to use `GITHUB_TOKEN` to delete artifacts. So it seems infeasible to do this. The only approach I can find is <https://github.com/marketplace/actions/delete-run-artifacts> which involves taking a personal access token, and I'm not sure how secure it is.

Moandor-y:
Found this [Delete Artifact](https://github.com/marketplace/actions/delete-artifact). It seems perfectly suited for this purpose. I'm trying it out.

Moandor-y:
Looks like it's working well. Marking this PR as ready for review.

simonlingoogle (approved):
LGTM :+1:

bukepo (approved):
LGTM:+1:

.github/workflows/posix.yml
```diff
@@ -303,3 +303,13 @@ jobs:
     - name: Upload Coverage
       run: |
         script/test upload_codecov
+
+  delete-coverage-artifacts:
+    needs: upload-coverage
+    if: always()
+    runs-on: ubuntu-18.04
```
jwhui, line 310: Can we use 20.04?

Moandor-y: Updated.

jwhui (approved):

Issue #5746, opened 2020-10-30
[nrf] Trying to do something after a frame received in radio but before ack sent


Irving-cl:
I'm now implementing Enhanced-ACK Based Probing. For this feature we need to aggregate Link Metrics data from frames sent from a specific Initiator, and put the aggregated data into the Enhanced-ACK sent to that Initiator.

My plan is to put the process in platform radio (I.E. implement the aggregation and fetch of data in `examples/platforms/xxx/radio.c`). The main reason for doing this is to ensure that the Link Metrics data in the latest Enh-ACK would contain the information of the latest received frame. 

For example, after Enhanced-ACK Based Probing is configured, the Initiator sends a 2015 frame **A** to the Subject. And the Subject sends an Enh-ACK containing Vendor-Specific IE. We want the Link Metrics data of frame **A** aggregated in the IE. If we aggregate it in OpenThread core, I think we won't be able to make this work. (The ack would have already been sent out when the code reaches higher layer).

So if I do the aggregation and fetch in platform radio, maybe do it in `nrf_802154_received_timestamp_raw`, can we ensure the process to be completed before ack being sent out? Or it's better to do this in `nrf_802154_tx_ack_started`? But as in `nrf_802154_tx_ack_started` we only have the generated ack, instead of the original received frame, I would prefer to do this process with the received frame. 

In short, I want to do something immediately after a frame is received in radio but before the ack sent out.

What would be your suggestions on this? @hubertmis @edmont

hubertmis:
> The main reason for doing this is to ensure that the Link Metrics data in the latest Enh-ACK would contain the information of the latest received frame.

I think that this requirement is the reason why the feature should be implemented in the 802.15.4 driver. The API of 802.15.4 driver is designed in the way that it does not guarantee strict timings between calls and real radio operations. It is possible to call 802.15.4 driver API on one CPU core, while radio operations are executed on other one with nondeterministic timings of IPC between. When OpenThread receives notification that a frame was received, the notification can be delayed by IPC.

We have a proof of concept injecting CSL data to CSL IEs in the 802.15.4 driver. I think similar feature should be implemented to inject Link Metrics data to IEs in the 802.15.4 driver. I can share this PoC with you if you are interested.

Irving-cl:
@hubertmis 

[...]

PR #5780, opened 2020-11-06
[low-power] implement enhanced-ack based probing

Irving-cl:
This PR implements Enhanced-ACK Probing (on simulation platform).

An overview of the PR:
* Add a new public API `otLinkMetricsConfigEnhAckProbing` which sends Link Metrics Management Request to configure the probing. This is called on **Probing Initiator** side.
* Add a new radio platform API `otPlatRadioConfigureEnhAckProbing`. This API should be called on **Probing Subject** side when handling the requests from Initiators. This API notifies the radio to start/stop aggregating link metrics info and include the data into Vendor IE in enhanced-ACK for the specific neighbor. As discussed in #5746, the code for doing such thing should be put in radio driver.
* Add a util module `util/link_metrics`, which provides a group of APIs to implement **Probing Subject** side logic for Enh-ACK Probing. Any platform could use this module to implement the feature easily in radio driver.
* Add new util APIs in `util/mac_frame` to generate Enh-ACK Probing IE (Vendor IE with Thread OUI and SubType = 0) and set value for this IE.
* Update the implementation in `simulation/radio.c` to support **Probing Subject** side logic for Enh-ACK Probing, using APIs in `util/link_metrics` and `util/mac_frame`.
* Add a test `v1_2_LowPower_7_1_01_SingleProbeLinkMetricsWithEnhancedAcks.py` for testing.
* Misc:
    * Add check for all the public Link Metrics APIs (Initiator side) to ensure that the Subject is a neighbor of the Initiator. If the address is not link-local or the neighbor is not found, an error `UNKNOWN_NEIGHBOR` would be returned.
    * Update `PrepareEmptyFrame` so that when it's called for reference device, its frame version would be set to `2015`.


examples/platforms/simulation/radio.c
```diff
@@ -843,13 +850,47 @@ void radioSendAck(void)
...
+
         otEXPECT(otMacFrameGenerateEnhAck(&sReceiveFrame, sReceiveFrame.mInfo.mRxInfo.mAckedWithFramePending,
                                           sAckIeData, sAckIeDataLength, &sAckFrame) == OT_ERROR_NONE);
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
         if (sCslPeriod > 0)
         {
             otMacFrameSetCslIe(&sAckFrame, (uint16_t)sCslPeriod, getCslPhase());
         }
+#endif
+#if OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE
```
edmont, line 889: Why not moving all the previous link metrics code here? Less variables would be needed.

examples/platforms/simulation/radio.c
```diff
@@ -843,13 +850,47 @@ void radioSendAck(void)
...
+        uint8_t  linkMetricsDataLen = 0;
+        uint8_t *dataPtr            = NULL;
+
+#if OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE
+        uint8_t      linkMetricsData[2];
+        otMacAddress macAddress;
+
+        otEXPECT(otMacFrameGetSrcAddr(&sReceiveFrame, &macAddress) == OT_ERROR_NONE);
+
+        if (macAddress.mType == OT_MAC_ADDRESS_TYPE_SHORT)
```
edmont, line 862: Can't this logic be put inside a single `otLinkMetricsEnhAckGetData` function with an `otMacAddress` argument?

examples/platforms/simulation/radio.c
```diff
@@ -888,6 +931,21 @@ void radioProcessFrame(otInstance *aInstance)
     otEXPECT_ACTION(otMacFrameDoesAddrMatch(&sReceiveFrame, sPanid, sShortAddress, &sExtAddress),
                     error = OT_ERROR_ABORT);
 
+#if OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE
+    otEXPECT_ACTION(otMacFrameGetSrcAddr(&sReceiveFrame, &macAddress) == OT_ERROR_NONE, error = OT_ERROR_PARSE);
+
+    if (macAddress.mType == OT_MAC_ADDRESS_TYPE_EXTENDED)
+    {
+        otLinkMetricsAggregateDataByExtAddress(&macAddress.mAddress.mExtAddress, sReceiveFrame.mInfo.mRxInfo.mLqi,
```
edmont, line 939: Ditto about `otLinkMetricsEnhAckGetDataBy*`.

examples/platforms/utils/mac_frame.cpp
```diff
@@ -200,3 +200,39 @@ void otMacFrameSetFrameCounter(otRadioFrame *aFrame, uint32_t aFrameCounter)
 {
     static_cast<Mac::Frame *>(aFrame)->SetFrameCounter(aFrameCounter);
 }
+
+uint8_t otMacFrameGenerateCslIeTemplate(uint8_t *aDest)
```
edmont, line 204: Missing CSL macro.

src/core/thread/link_metrics.cpp
```diff
@@ -119,6 +120,11 @@ otError LinkMetrics::LinkMetricsQuery(const Ip6::Address & aDestination,
     otError                error;
     LinkMetricsTypeIdFlags typeIdFlags[kMaxTypeIdFlags];
     uint8_t                typeIdFlagsCount = 0;
+    Mac::Address           macAddress;
+
+    VerifyOrExit(aDestination.IsLinkLocal(), error = OT_ERROR_UNKNOWN_NEIGHBOR);
```
edmont, line 125: Create a function with this logic which is repeated several times?

src/core/thread/link_metrics.cpp
```diff
@@ -537,6 +607,42 @@ void LinkMetrics::SetLinkMetricsMgmtResponseCallback(otLinkMetricsMgmtResponseCa
...
+
+    linkMetricsValues.mMetrics = aNeighbor.GetEnhAckProbingMetrics();
+
+    if (linkMetricsValues.mMetrics.mLqi && idx < aLen)
+    {
+        linkMetricsValues.mLqiValue = aData[idx++];
+    }
+    if (linkMetricsValues.mMetrics.mLinkMargin && idx < aLen)
+    {
+        linkMetricsValues.mLqiValue = aData[idx++];
```
edmont, line 632:
```suggestion
        linkMetricsValues.mLinkMargin = aData[idx++];
```

include/openthread/link_metrics.h
```diff
@@ -92,11 +81,23 @@ typedef struct otLinkMetricsSeriesFlags
...
+    OT_LINK_METRICS_ENH_ACK_CLEAR    = 0, ///< Clear.
+    OT_LINK_METRICS_ENH_ACK_REGISTER = 1, ///< Register.
+} otLinkMetricsEnhAckFlags;
+
 /**
  * Link Metrics Status values.
  *
  */
-typedef enum otLinkMetricsStatus : uint8_t
+typedef enum otLinkMetricsStatus
```
librasungirl, line 100: what's the reason for this change?

Just noticing that `aStatus` is of type `uint8_t` in below two callback declarations, better to update to be of type `otLinkMetricsStatus` (fine to me if you prefer a separate PR for this).

https://github.com/openthread/openthread/blob/6fda0786edab8e821e1647cc4be08b4a7059cf4a/include/openthread/link_metrics.h#L118-L130

include/openthread/link_metrics.h
```diff
@@ -129,6 +130,11 @@ typedef void (*otLinkMetricsReportCallback)(const otIp6Address *       aSource,
  */
 typedef void (*otLinkMetricsMgmtResponseCallback)(const otIp6Address *aSource, uint8_t aStatus, void *aContext);
 
+typedef void (*otLinkMetricsEnhAckProbingIeReportCallback)(const otShortAddress       aShortAddress,
```
librasungirl, line 133: add document

include/openthread/link_metrics.h
```diff
@@ -141,8 +147,9 @@ typedef void (*otLinkMetricsMgmtResponseCallback)(const otIp6Address *aSource, u
  * @param[in]  aCallback            A pointer to a function that is called when Link Metrics report is received.
  * @param[in]  aCallbackContext     A pointer to application-specific context.
  *
- * @retval OT_ERROR_NONE          Successfully sent a Link Metrics query message.
- * @retval OT_ERROR_NO_BUFS       Insufficient buffers to generate the MLE Data Request message.
+ * @retval OT_ERROR_NONE              Successfully sent a Link Metrics query message.
+ * @retval OT_ERROR_NO_BUFS           Insufficient buffers to generate the MLE Data Request message.
+ * @retval OT_ERROR_UNKNOWN_NEIGHBOR  @p aDestination is not link-local or the neighbor is not found.
```
librasungirl, line 152: Here `OT_ERROR_INVALID_ARGS` might be more appropriate than  `OT_ERROR_UNKNOWN_NEIGHBOR`?
https://github.com/openthread/openthread/blob/6fda0786edab8e821e1647cc4be08b4a7059cf4a/include/openthread/error.h#L156-L159

Better also check that the neighbor is of thread version >= 3 as Thread 1.1 neighbor won't support this feature for sure.

include/openthread/link_metrics.h
```diff
@@ -178,6 +186,34 @@ otError otLinkMetricsConfigForwardTrackingSeries(otInstance *
                                                  otLinkMetricsMgmtResponseCallback aCallback,
                                                  void *                            aCallbackContext);
 
+/**
+ * This function sends an MLE Link Metrics Management Request to configure/clear a Enhanced-ACK Based Probing.
```
librasungirl, line 190:
```suggestion
 * This function sends an MLE Link Metrics Management Request to configure/clear an Enhanced-ACK Based Probing.
```

include/openthread/link_metrics.h
```diff
@@ -178,6 +186,34 @@ otError otLinkMetricsConfigForwardTrackingSeries(otInstance *
...
+ *
+ * @param[in] aInstance          A pointer to an OpenThread instance.
+ * @param[in] aDestination       A pointer to the destination address.
+ * @param[in] aEnhAckFlags       Enh-ACK Flags to indicate whether to register or clear the probing. `0` to clear and
+ *                               `1` to register. Other values are reserved.
+ * @param[in] aLinkMetricsFlags  A pointer to flags specifying what metrics to query. Should be `NULL` when
+ *                               `aEnhAckFlags` is `0`.
+ * @param[in]  aCallback         A pointer to a function that is called when Link Metrics Management Response is
```
librasungirl, line 198:
```suggestion
 * @param[in]  aCallback         A pointer to a function that is called when an Enhanced Ack with Link Metrics is received.
```

include/openthread/platform/radio.h
```diff
@@ -946,6 +959,31 @@ otError otPlatRadioEnableCsl(otInstance *aInstance, uint32_t aCslPeriod, const o
...
+/**
+ * Enable/disable or update Enhanced-ACK Based Probing in radio for a specific Initiator.
+ *
+ * After Enhanced-ACK Based Probing is configured by a specific Probing Initiator, the Enhanced-ACK sent to that
+ * node should include Vendor-Specific IE containing Link Metrics data. This method informs the radio to start/stop to
+ * collect and aggregate Link Metrics data and include Vendor-Specific IE that containing the data in Enhanced-ACK sent
+ * to that Probing Initiator.
+ *
+ * @param[in]  aInstance    The OpenThread instance structure.
+ * @param[in]  aDataLength  Length of Link Metrics data in the Vendor-Specific IE. Per spec 4.11.3.4.4.6,
```
librasungirl: nit: no such parameter.

src/core/mac/mac.cpp
```diff
@@ -1301,6 +1301,7 @@ void Mac::RecordFrameTransmitStatus(const TxFrame &aFrame,
 #if OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE
```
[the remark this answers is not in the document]
librasungirl: Why Subject and Initiator should be differentiated? It should be allowed for one Thread device be both Probing Subject and Probing Initiator, right?

src/core/mac/mac.cpp
```diff
@@ -2372,6 +2373,28 @@ void Mac::ProcessCsl(const RxFrame &aFrame, const Address &aSrcAddr)
 }
 #endif // OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
 
+#if OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE
+void Mac::ProcessEnhAckProbing(const RxFrame &aFrame, const Neighbor &aNeighbor)
+{
+    const HeaderIe *enhAckProbingIe =
+        reinterpret_cast<const HeaderIe *>(aFrame.GetThreadIe(ThreadIe::kEnhAckProbingIe));
+    uint8_t data[2];
```
librasungirl, line 2381: nit: avoid magic number 2

src/core/mac/mac.cpp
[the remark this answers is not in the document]
edmont: @librasungirl, according to the _Table 4-1b. Version value indication of mandatory and optional feature support_:
- Devices that **may** support being Link Metrics Probing **Initiator**: Router, REED, FED, MED, SED, SSED.
- Devices that **must** support being Link Metrics Probing **Subject**: Router, REED.

So I believe it would be beneficial to allow building those features separately for the different device types.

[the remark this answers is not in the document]
edmont: Anyway that could be something to address in a different PR.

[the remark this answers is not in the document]
librasungirl: @edmont , thanks for the reminder, and agree with your summary 👍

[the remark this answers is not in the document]
Irving-cl: I would address this in another PR.

examples/platforms/utils/link_metrics.h
```diff
@@ -0,0 +1,135 @@
...
+#include <openthread/link_metrics.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * This method initializes the Link Metrics util module.
+ *
+ * @param[in]  aNoiseFloor    The noise floor that would be used in Link Metrics. It should be set to the platform
```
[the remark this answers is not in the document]
Irving-cl: @edmont 

I'd like to know who actually decides what to be used as the noise floor? Thread allows to use measured noise floor, a constant or receiver sensitivity. So does that mean, the vendor decides which one to be used?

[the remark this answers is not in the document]
jwhui: It is up to the platform how the noise floor is implemented.

examples/platforms/simulation/radio.c
Irving-cl: I move this part into the previous code block instead. Because we need to determine the length of all IE data before generating ACK data with the IE data.

Irving-cl: When generating the ACK:
1. first generate header IE patterns, determine what IEs are to be included and their lengths
2. generate the ACK with the IE just generated
3. update the IE's value

Currently we use previous link metrics code to determine the header IE pattern. So we have to do it before ACK is generated. And we have to set the value of the IE after the ACK is generated.

src/cli/cli.cpp
```diff
@@ -2169,6 +2205,40 @@ otError Interpreter::ProcessLinkMetricsMgmt(uint8_t aArgsLength, char *aArgs[])
...
+
+        if (strcmp(aArgs[2], "clear") == 0)
+        {
+            enhAckFlags  = OT_LINK_METRICS_ENH_ACK_CLEAR;
+            pLinkMetrics = nullptr;
+        }
+        else if (strcmp(aArgs[2], "register") == 0)
+        {
+            enhAckFlags = OT_LINK_METRICS_ENH_ACK_REGISTER;
+            VerifyOrExit(aArgsLength >= 4);
```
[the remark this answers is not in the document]
Irving-cl: When compiling the code, this would cause an error:

> error: embedding a directive within macro arguments is not portable

include/openthread/link_metrics.h
Irving-cl: The reason is that C compiler doesn't support this. The platform code including this header are compiled with C compiler. If we add it, we got such error:
```
In file included from ../../../examples/platforms/utils/link_metrics.h:42:0,
                 from ../../../examples/platforms/simulation/radio.c:43:
../../../include/openthread/link_metrics.h:100:14: error: ISO C forbids forward references to 'enum' types [-Wpedantic]
 typedef enum otLinkMetricsStatus : uint8_t
              ^~~~~~~~~~~~~~~~~~~
../../../include/openthread/link_metrics.h:100:34: error: expected identifier or '(' before ':' token
 typedef enum otLinkMetricsStatus : uint8_t
                                  ^
../../../include/openthread/link_metrics.h:108:3: error: data definition has no type or storage class
 } otLinkMetricsStatus;
```

This error is not found in last PR because at that time the radio doesn't use the link_metrics module which includes the public link metrics header.

include/openthread/link_metrics.h
Irving-cl: `OT_ERROR_INVALID_ARGS` has already been used to represent other argument errors in the following 3 APIs. I suggest that we add a new error that represents neighbor doesn't support 1.2 features. And in the case `aDestination` is not link-local or neighbor is not found, we use `OT_ERROR_UNKNOWN_NEIGHBOR`. Thoughts?

edmont (commented):
Thanks for the changes @Irving-cl.

Overall it looks good (apart from the nits in the comments), but I believe there is a key missing piece: how is the Probing Initiator supposed to trigger the retrieve of the data aggregated by the Probing Subject? At the moment it's only possible by sending an empty data frame as reference device, or by using CSL.

There should be a way for non CSL Receiver devices to make use of this feature. Please correct me if I missed something.

Irving-cl:
> Thanks for the changes @Irving-cl.
> 
> Overall it looks good (apart from the nits in the comments), but I believe there is a key missing piece: how is the Probing Initiator supposed to trigger the retrieve of the data aggregated by the Probing Subject? At the moment it's only possible by sending an empty data frame as reference device, or by using CSL.
> 
> There should be a way for non CSL Receiver devices to make use of this feature. Please correct me if I missed something.

@edmont 

I think the idea of Enhanced-ACK Based Probing is that the Initiator could avail of the regular messages (for example, in real world, the application would send some messages) to get Link Metrics data instead of sending a specific message to retrieve the data and thus reduce power consumption. After the Initiator successfully configured it at a Subject, any other frames (with version 2015) sent by it would get an Enhanced Ack with Vendor Specific IE.

edmont:
@Irving-cl yes, I agree there is no need for extra frames, but at the moment the application in the Probing Initiator has no way for telling the MAC to set `kFcfFrameVersion2015` other than using CSL (apart from optional Time Sync and reference device empty frames).

In order for this functionality to be useful there should be a mechanism to select when to trigger an 802.15.4-2015 frame. Probably we don't want all sent frames to be 2015 version, so at least we could establish that 1 every N frames it's. Thoughts?

Irving-cl:
@edmont 

That makes sense. From the spec, it feels like, when the application wants to know the Link Metrics information, then it sends a 2015 frame. Imagine that we are application users and using OpenThread with this feature, we may want to get the Link Metrics information after the Initiator sends a specific number of frames, say, 10. So maybe we can provide a new API that enables users to set the number **N**. And then every N frame, one frame's version would be set to 2015. If the API was not called, there would be a default **N**. Thoughts?

edmont:
For simplicity we could even start with just a build time configuration for N and add the API in the future if needed. Or even assume N=1 and force 802.15.4-2015 data and command frames whenever Enhance Ack probing is configured. Maybe the overhead is not so big and at the end the important thing is what the application does with `otLinkMetricsEnhAckProbingIeReportCallback`.

However, not sure if this would break Thread Certification somehow.

examples/platforms/simulation/radio.c
```diff
@@ -888,6 +922,13 @@ void radioProcessFrame(otInstance *aInstance)
     otEXPECT_ACTION(otMacFrameDoesAddrMatch(&sReceiveFrame, sPanid, sShortAddress, &sExtAddress),
                     error = OT_ERROR_ABORT);
 
+#if OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE
+    otEXPECT_ACTION(otMacFrameGetSrcAddr(&sReceiveFrame, &macAddress) == OT_ERROR_NONE, error = OT_ERROR_PARSE);
+
+    otLinkMetricsAggregateDataByMacAddress(&macAddress, sReceiveFrame.mInfo.mRxInfo.mLqi,
```
edmont, line 928: @Irving-cl I believe there has been some confusion interpreting the specification and there should not be any aggregation of metrics when Enhanced ACK probing is configured. The spec states, for the contents of the IE in the ACK:

> The remaining 1 or 2 bytes contain the requested Link Metrics data assessed **for the acknowledged frame** in order of request.

This means there is no need to store aggregated information for every neighbor, just appending the metrics for every triggering 802.15.14-2015 frame should be enough.

edmont (changes requested):

examples/platforms/simulation/radio.c
Irving-cl: @edmont 

Thanks for noticing that. I think your interpreting is correct. I'll update it. Then I think it's difficult to implement it on nrf platform for now. Because the ack are auto generated and we can only use `tx_ack_started` to update the value. However we cannot access the received frame's lqi or rssi in `tx_ack_started`. So maybe we implement it on nrf platform in next version (with the PoC you shared with me)?

edmont: Probably we'll need implementation for both approaches:
- For radios supporting automatic data injection to enhanced acks: a radio driver API to configure/deconfigure the requested metrics per Probing Initiator (short and extended addressing). Also an API to query the number of remaining slots for enhanced ack injection.
- For radios not supporting it: a way for the MAC implementation to process/store metrics for the last frame and to inject them in the enhanced ack (similar to what it's implemented now).

Makes this sense for you?

edmont: We could have a new `OT_RADIO_CAPS_AUTO_ENH_ACK`, which assumes CSL IE injection as well.

Irving-cl: @edmont 

> Probably we'll need implementation for both approaches:
> 
> *  For radios supporting automatic data injection to enhanced acks: a radio driver API to configure/deconfigure the requested metrics per Probing Initiator (short and extended addressing). Also an API to query the number of remaining slots for enhanced ack injection.
> *  For radios not supporting it: a way for the MAC implementation to process/store metrics for the last frame and to inject them in the enhanced ack (similar to what it's implemented now).

That makes seons to me.

> We could have a new OT_RADIO_CAPS_AUTO_ENH_ACK, which assumes CSL IE injection as well.

Let me clarify on this. Do we regard current nrf platform has the capability `OT_RADIO_CAPS_AUTO_ENH_ACK`? For now we update the CSL IE value in `tx_ack_started`. But we can only have the Link Metrics data for the last frame now.

edmont: Currently there is no support of OT_RADIO_CAPS_AUTO_ENH_ACK for nRF platforms, but it will come in the future, and will be the only option for dual core SoCs. So I believe it's good to take it into account when designing this feature.

Irving-cl: It seems that this capability doesn't affect OpenThread implementation. For example, for CSL, `OT_RADIO_CAPS_TRANSMIT_TIMING` would decide if we use the software timer in `SubMac`. But for this Enhanced-ACK Based Probing feature, we just need to implement it differently on different platforms (OpenThread doesn't need to know if the platform has this capability). 

The only usage I can think of is that the Probing Initiator needs to know if the Probing Subject has this capability so that it knows if the data in IE is for the last frame or for the previous frame. But this information is on Probing Subject, so more work is required so that Probing Initiator can have this information. Thoughts?

src/core/thread/mle.cpp
```diff
@@ -2444,7 +2444,7 @@ otError Mle::SendOrphanAnnounce(void)
 }
 
 #if OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE
```
edmont, line 2446: Add missing comment below:
`#endif // OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE`

edmont, line 2446: Add missing comment below:
`#endif // OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE`

And similar other cases for this macro.

examples/platforms/utils/link_metrics.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+        if (dataInfo == nullptr)
+        {
+            dataInfo = GetLinkMetricsDataInfoPool().Allocate();
+            VerifyOrExit(dataInfo != nullptr, error = OT_ERROR_NO_BUFS);
+            dataInfo->Clear();
+            GetLinkMetricsDataInfoActiveList().Push(*dataInfo);
+        }
+
+        dataInfo->Set(aLinkMetrics, aShortAddress, *aExtAddress);
```
edmont, line 235: Are we overwriting a previous configuration by design? Maybe a comment about that here would be useful.

examples/platforms/simulation/radio.c
edmont: Probably all the logic can be moved to the platform's radio implementation but I guess it would be beneficial to keep as much as possible within OT itself. But it depends on how would you approach the new requirement of not aggregating data from previous frames.

Maybe you are right and platforms can decide whether to reuse `examples/platforms/utils/` code or do it their own way.

About the your second paragraph, I don't think there should be other way than injecting the IE data for the last frame, as requested by the spec. If the Probing Subject can't achieve it, then it can't support this feature.

Irving-cl:
@edmont 

> About the your second paragraph, I don't think there should be other way than injecting the IE data for the last frame, as requested by the spec. If the Probing Subject can't achieve it, then it can't support this feature.

Yeah, I mean it's unclear to me how we should use `OT_RADIO_CAPS_AUTO_ENH_ACK`. If it indicates the support of IE injection, then I think the current nRF platform actually satisfies this requirement (it does use this for CSL IE). However it doesn't support to include the link metrics data for the latest frame.

I think CSL IE injection is a MUST requirement for platforms. If a platform doesn't support it, then it doesn't support OpenThread 1.2. For latest link metrics data, I think we can make it optional for a platform. If a platform doesn't support it, then it always include the data for previous frame.
* So if we make it compulsory to support 'latest link metrics data', then a platform must support all these things. Otherwise it doesn't support OpenThread 1.2. Then `OT_RADIO_CAPS_AUTO_ENH_ACK` doesn't indicate anything meaningful. (If it's not `true` then the platform doesn't suppport 1.2 at all, clearly we have enough information about if a platform supports 1.2 and we don't need this flag.)
* If we make it optional to support 'latest link metrics data', then we need a specific flag to indicate this information. It's only for 'latest link metrics data', not related to CSL IE and other IEs. However, even we add a new flag for this information, this information is on Probing Subject. And if the Probing Initiator want to know this, it may need to send some messages to get this information. I.E. the Probing Initiator then knows if the link metrics data it gets is for the latest frame or the previous frame.

So from my point of view, I think `OT_RADIO_CAPS_AUTO_ENH_ACK` is something compulsory for OpenThread 1.2 and we don't need to add a flag to indicate that. Simply if a platform doesn't support it, then it doesn't support OpenThread 1.2. And since we have enough information about if a device supports 1.2, we don't need to use this flag.
And for the capability of 'latest link metrics data', I think there's some workaround. We'll allow a platform to include the data for previous frame. And if that is the case, then it should send some special value in the first enh-ack to the Initiator and the n the Initiator knows that. (First enh-ack don't have any meaningful data if 'latest link metrics data' is not supported).

Thoughts?

examples/platforms/simulation/radio.c
Irving-cl: I updated the ack generating process. The aggregating process is removed. Instead, the raw information is transformed into IE data value and put into the IE.

edmont (commented):
@Irving-cl I don't think we should focus on the option about providing link metrics information from the last frame. That's against the spec, so the only option to support Thread 1.2 would be to provide link metrics information for the acknowledged frame.

As for supporting this in the current nRF platform, I believe we would only need a way to temporary store the LQI and RSSI when processing the received frame (`nrf_802154_received_timestamp_raw`) and retrieve that information when filling the IE data (`nrf_802154_tx_ack_started`).

We can probably forget about `OT_RADIO_CAPS_AUTO_ENH_ACK` as you suggested. When multicore platform support is added, all this can be handled by its `radio.c`, with no new changes in OpenThread implementation itself.

Do you think this is achievable?

examples/platforms/utils/link_metrics.cpp
```diff
@@ -0,0 +1,231 @@
...
+     *
+     * @param[in]   aLqi     LQI value of the acknowledeged frame.
+     * @param[in]   aRssi    RSSI value of the acknowledged frame.
+     * @param[out]  aData    A pointer to the output buffer. @p aData MUST NOT be `nullptr`. The buffer should be larger
+     * than 2 bytes at least. Otherwise the behavior would be undefined.
+     *
+     * @returns  The number of bytes written. If the writing fails, `0` would be returned.
+     *
+     */
+    uint8_t GetEnhAckData(uint8_t aLqi, int8_t aRssi, uint8_t *aData) const
```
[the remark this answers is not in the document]
Irving-cl: For this problem, I think we can always specify a fixed order. For example, `Pdu count`, `LQI`, `Link Margin`, `Rssi`. Whatever metrics are specified, they are in such order. Say, `LQI` and `Rssi` are specified, then the order is `LQI` and `Rssi`. And users then know the first byte is `LQI` and the second one is `Rssi`. I think the users won't care about the order and it's meaningless to specify a different order. Storing the order also brings extra overhead.

The spec indeed specifies the order. I'd like to open a JIRA to discuss the problem. If it is rejected, then we can update the implementation here. I think for now we can keep it as it is. Thoughts?

Irving-cl:
@edmont 

> As for supporting this in the current nRF platform, I believe we would only need a way to temporary store the LQI and RSSI when processing the received frame (nrf_802154_received_timestamp_raw) and retrieve that information when filling the IE data (nrf_802154_tx_ack_started).

I had the same idea in issue #5746. But as hubertmis [commented](https://github.com/openthread/openthread/issues/5746#issuecomment-719257961), this seems not work because the timing is undeterministic. So we cannot ensure `received` is called before `tx_ack_started`.

examples/platforms/utils/link_metrics.cpp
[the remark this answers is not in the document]
edmont: I don't have a strong opinion, both options look good to me. However probably there isn't a strong reason to modify the specification about this at the moment.

edmont:
> I had the same idea in issue #5746. But as hubertmis [commented](https://github.com/openthread/openthread/issues/5746#issuecomment-719257961), this seems not work because the timing is undeterministic. So we cannot ensure `received` is called before `tx_ack_started`.

You are right, I realized about that later on. However we could easily overcome this issue by gathering RSSI and LQI within `nrf_802154_tx_ack_started`:
- RSSI: `nrf_802154_rssi_last_get()`
- LQI: making `lqi_get()` (https://github.com/openthread/openthread/blob/master/third_party/NordicSemiconductor/drivers/radio/nrf_802154_core.c#L297) public

Irving-cl:
@edmont 

That sounds good to me! Thanks.

examples/platforms/utils/link_metrics.cpp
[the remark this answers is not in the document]
Irving-cl: @edmont 

Make sense to me. I think we can just disallow the users to specify the order in public API. Thus when OpenThread works as the Initiator, the order of metrics is fixed. However we need to allow OpenThread to handle a different order of metrics when OpenThread works as the Subject and works with other Initiators. And that fully conforms to the spec.

Currently the forwarding tracking series feature also ignores the order. I'd like to submit a separate PR later to handle the order for both forwarding tracking series and Enhanced-ACK probing. Thoughts?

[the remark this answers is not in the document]
edmont: Can you please add a TODO comment in this PR about that?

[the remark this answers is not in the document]
Irving-cl: Updated, thanks.

edmont (changes requested):
We are still missing a way to retrieve the metrics. Can we go the easy path and just force all frames to be 802.15.4-2015 whenever enh-ack probing has been configured with that neighbor?

Irving-cl:
> We are still missing a way to retrieve the metrics. Can we go the easy path and just force all frames to be 802.15.4-2015 whenever enh-ack probing has been configured with that neighbor?

@edmont 

SGTM. I'll update it soon.

Irving-cl:
@edmont 

Just updated to enforce the Initiator to send 2015 frames to Subject after configuring Enh-ACK Probing.

I've verified it locally:
Screen Shot 2020-11-25 at 7 01 49 PM

No.56 is a Link Management Request (Enh-ACK Probing). No.42 is a ping before the request, its frame version is 2006. And No.64 is a ping after the request, its frame version is 2015. And the IE is contained in the ACK No.65.

src/core/thread/topology.hpp
```diff
@@ -659,8 +659,23 @@ class Neighbor : public InstanceLocatorInit
...
+    /**
+     * This method indicates if Enh-ACK Probing is configured and active for this `Neighbor` object.
+     *
+     * @retval TRUE   Enh-ACK Probing is configured and active for this `Neighbor`.
+     * @retval FALSE  Otherwise.
+     *
+     */
+    bool IsEnhAckProbingActive() const
+    {
+        return (mEnhAckProbingMetrics.mPduCount != 0) || (mEnhAckProbingMetrics.mLqi != 0) ||
```
edmont, line 673: I guess `mPduCount` check can be removed since it's not allowed for enh-ack probing.
`mReserved` check could be conditioned for reference device only.

edmont (approved):
Thanks for all your work here @Irving-cl

src/core/thread/topology.hpp
Irving-cl: Actually `mReserved` is also unnecessary since the Subject only accepts metrics that are supported. This field is used for public API to let Initiator to configure some metrics with reserved field for reference devices.

src/core/thread/topology.hpp
```diff
@@ -639,12 +639,42 @@ class Neighbor : public InstanceLocatorInit
...
      *
      */
     void RemoveAllForwardTrackingSeriesInfo(void);
 
-#endif
+    /**
+     * This method gets the Enh-ACK Probing metrics (this `Neighbor` object is the Probing Subject).
+     *
+     */
+    otLinkMetrics GetEnhAckProbingMetrics(void) const { return mEnhAckProbingMetrics; }
```
abtink, line 651: Can we consider returning `const otLinkMetrics &` here? (This can help avoid creating a copy of the `otLinkMetrics` when we return it.

src/core/mac/mac.cpp
```diff
@@ -2364,6 +2365,33 @@ void Mac::ProcessCsl(const RxFrame &aFrame, const Address &aSrcAddr)
...
+
+    VerifyOrExit(enhAckProbingIe != nullptr);
+
+    dataLen = enhAckProbingIe->GetLength() - sizeof(VendorIeHeader);
+    VerifyOrExit(dataLen <= kEnhAckProbingIeMaxLen);
+
+    memcpy(data, reinterpret_cast<const uint8_t *>(enhAckProbingIe) + sizeof(HeaderIe) + sizeof(VendorIeHeader),
+           dataLen);
+
+    Get<LinkMetrics>().ProcessEnhAckIeData(data, dataLen, aNeighbor);
```
abtink, line 2389: I think we are copying the IE data from the `RxFrame` into `data[]` array and then pass it to `ProcessEnhAckIeData()`. And the `ProcessEnhAckIeData()` is treating it as `const uint8_t *` (so cannt change it).

A suggestion for an optimization here: 
Can we instead of copying it, find a pointer to to start of data (in the `RxFrame` buffer) and pass the pointer to `ProcessEnhAckIeData()`.

examples/platforms/utils/link_metrics.h
```diff
@@ -0,0 +1,104 @@
...
+ * write the values to @p aData.
+ *
+ * @param[in]  aMacAddress    The Mac address of the Initiator.
+ * @param[in]  aLqi           LQI value of the acknowledged frame.
+ * @param[in]  aRssi          RSSI value of the acknowledged frame.
+ * @param[out] aData          A pointer to the buffer where the data would be written to. The caller should make
+ *                            sure that the size of the buffer is not less than the size of Link Metrics data
+ *                            configured before.
+ *
+ * @retval  The size of data read. Would be `0` if the Initiator is not found or @p aData is invalid.
```
abtink, line 95:
```suggestion
 * @returns  The size of data read. Would be `0` if the Initiator is not found or @p aData is invalid.
```

include/openthread/link_metrics.h
abtink: Yes. (If I recall correctly) specifying `enum` size is supported from C++11 onwards.

abtink (approved):
Thanks @Irving-cl. Looks great. 👍 
Few small suggestions below.

src/core/thread/link_metrics.cpp
```diff
@@ -622,6 +719,51 @@ otLinkMetricsStatus LinkMetrics::ConfigureForwardTrackingSeries(uint8_t
...
+{
+    LinkMetricsStatus status = kLinkMetricsStatusSuccess;
+    otError           error  = OT_ERROR_NONE;
+
+    VerifyOrExit(!aLinkMetrics.mReserved, status = kLinkMetricsStatusCannotSupportNewSeries);
+
+    if (aEnhAckFlags == kEnhAckRegister)
+    {
+        VerifyOrExit(!aLinkMetrics.mPduCount, status = kLinkMetricsStatusOtherError);
+        VerifyOrExit(!(aLinkMetrics.mLqi && aLinkMetrics.mLinkMargin && aLinkMetrics.mRssi),
```
librasungirl, line 734: should here verify at least one is true?

src/core/thread/link_metrics.cpp
```diff
@@ -622,6 +719,51 @@ otLinkMetricsStatus LinkMetrics::ConfigureForwardTrackingSeries(uint8_t
...
+                     status = kLinkMetricsStatusOtherError);
+
+        error = Get<Radio>().ConfigureEnhAckProbing(aLinkMetrics, aNeighbor.GetRloc16(), aNeighbor.GetExtAddress());
+    }
+    else if (aEnhAckFlags == kEnhAckClear)
+    {
+        error = Get<Radio>().ConfigureEnhAckProbing(aLinkMetrics, aNeighbor.GetRloc16(), aNeighbor.GetExtAddress());
```
librasungirl, line 741: should here verify all (mLqi, mPduCount, mLinkMargin, mRssi) are unset?

examples/platforms/utils/mac_frame.h
```diff
@@ -283,6 +283,43 @@ uint32_t otMacFrameGetFrameCounter(otRadioFrame *aFrame);
...
+ * @returns  The total count of bytes (total length of CSL IE) written to the buffer.
+ *
+ */
+uint8_t otMacFrameGenerateCslIeTemplate(uint8_t *aDest);
+
+/**
+ * Write Enh-ACK Probing IE (Vendor IE with THREAD OUI) to a buffer.
+ *
+ * @p aIeData could be `NULL`. If @p aIeData is `NULL`, this method generates the IE with the data unset. This allows
+ * users to generate the pattern first and update value later. (For example, using `otMacFrameSetEnhAckProbingIe`)
```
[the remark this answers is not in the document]
Irving-cl: This is for different vendors to have different implementations: Some vendors may generate the Probing IE at once and then append it to ACK. Some vendors may generate the "pattern" for the IE first (without setting the IE value) and later set the IE value in ISR like `tx_ack_started` (currently on nRF). So when `@p aIeData` is `NULL`, the "pattern" was first generated. Thoughts?

src/core/radio/radio.hpp
```diff
@@ -594,6 +594,36 @@ class Radio : public InstanceLocator, private NonCopyable
...
+     * node should include Vendor-Specific IE containing Link Metrics data. This method informs the radio to
+     * starts/stops to collect and aggregate Link Metrics data and include Vendor-Specific IE that containing the data
+     * in Enhanced-ACK sent to that Probing Initiator.
+     *
+     * @param[in]  aInstance    The OpenThread instance structure.
+     * @param[in]  aDataLength  Length of Link Metrics data in the Vendor-Specific IE. Per spec 4.11.3.4.4.6,
+     *                          @p aDataLength should only be 1 or 2. The probing would be disabled if `aDataLength` is
+     *                          `0`.
+     * @param[in]  aShortAddr   The short address of the the probing Initiator.
+     * @param[in]  aExtAddr     The extended source address of the probing Initiator. @p aExtAddr MUST NOT be `NULL`.
```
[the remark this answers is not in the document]
Irving-cl: `aExtAddr` is reference here actually. I removed the sentence.

src/core/thread/link_metrics.cpp
```diff
@@ -164,30 +169,68 @@ otError LinkMetrics::SendMgmtRequestForwardTrackingSeries(const Ip6::Address &
...
+    uint8_t      subTlvs[sizeof(Tlv) + sizeof(enhAckFlags) + sizeof(LinkMetricsTypeIdFlags) * kMaxTypeIdFlags];
+    Tlv *        enhancedAckLinkMetricsConfigurationSubTlv = reinterpret_cast<Tlv *>(subTlvs);
+    Mac::Address macAddress;
+    Neighbor *   neighbor = nullptr;
+
+    VerifyOrExit(aDestination.IsLinkLocal(), error = OT_ERROR_UNKNOWN_NEIGHBOR);
+    aDestination.GetIid().ConvertToMacAddress(macAddress);
+    neighbor = Get<NeighborTable>().FindNeighbor(macAddress);
+
+    VerifyOrExit(neighbor != nullptr, error = OT_ERROR_UNKNOWN_NEIGHBOR);
```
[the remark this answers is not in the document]
Irving-cl: But `neighbor` is used later. We cannot get `neighbor` from `IsDestAddrValid`. Thoughts?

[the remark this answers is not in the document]
Irving-cl: I think I can add another method `GetNeighborFromDestAddr` and implement `IsDestAddrValid` with it.

examples/platforms/utils/mac_frame.h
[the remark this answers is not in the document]
librasungirl: Now it make sense to me since it is prepared for further support on nRF. Thanks.

librasungirl (approved):

examples/platforms/utils/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@
 add_library(openthread-platform-utils OBJECT
     debug_uart.c
     logging_rtt.c
+    link_metrics.cpp
```
jwhui, line 32:
```suggestion
    link_metrics.cpp
    logging_rtt.c
```

examples/platforms/utils/Makefile.am
```diff
@@ -42,6 +42,8 @@ libopenthread_platform_utils_a_SOURCES  = \
     debug_uart.c                          \
     logging_rtt.c                         \
     logging_rtt.h                         \
+    link_metrics.cpp                      \
+    link_metrics.h                        \
```
jwhui, line 46:
```suggestion
    link_metrics.cpp                      \
    link_metrics.h                        \
    logging_rtt.c                         \
    logging_rtt.h                         \
```

examples/platforms/utils/link_metrics.cpp
```diff
@@ -0,0 +1,238 @@
...
+#include "common/clearable.hpp"
+#include "common/linked_list.hpp"
+#include "common/pool.hpp"
+#include "thread/link_quality.hpp"
+
+#if OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE
+
+using namespace ot;
+
+static int8_t sNoiseFloor; ///< The noise floor that would be used in Link Metrics. It should be set to the platform's
```
jwhui, line 42:
```suggestion
static int8_t sNoiseFloor; ///< The noise floor used by Link Metrics. It should be set to the platform's
```

examples/platforms/utils/link_metrics.cpp
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * Set the information for this object.
+     *
+     * @param[in]  aLinkMetrics     Flags specifying what metrics to query.
+     * @param[in]  aShortAddress    Short Address of the Probing Initiator tracked by this object.
+     * @param[in]  aExtAddress      A reference to the Extended Address of the Probing Initiator tracked by this
+     *                              object.
+     *
+     */
+    void Set(const otLinkMetrics aLinkMetrics, const otShortAddress aShortAddress, const otExtAddress &aExtAddress)
```
jwhui, line 66:
```suggestion
    void Set(otLinkMetrics aLinkMetrics, otShortAddress aShortAddress, const otExtAddress &aExtAddress)
```

examples/platforms/utils/link_metrics.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * This method gets Link Metrics data stored in this object.
+     *
+     * TODO: Currently the order of Link Metircs data is fixed. Will update it to follow the order specified in TLV.
+     *
+     * @param[in]   aLqi     LQI value of the acknowledeged frame.
+     * @param[in]   aRssi    RSSI value of the acknowledged frame.
+     * @param[out]  aData    A pointer to the output buffer. @p aData MUST NOT be `nullptr`. The buffer should have 2
+     *                       bytes at least (Per spec 4.11.3.4.4.6). Otherwise the behavior would be undefined.
```
jwhui, line 81:
```suggestion
     * @param[out]  aData    A pointer to the output buffer. @p aData MUST NOT be `nullptr`. The buffer must have
     *                      at least 2 bytes (per spec 4.11.3.4.4.6). Otherwise the behavior would be undefined.
```

src/cli/cli.cpp
```diff
@@ -2003,6 +2008,29 @@ void Interpreter::HandleLinkMetricsMgmtResponse(const otIp6Address *aAddress, ui
     OutputLine("Status: %s", LinkMetricsStatusToStr(aStatus));
 }
 
+void Interpreter::HandleLinkMetricsEnhAckProbingIe(const otShortAddress       aShortAddress,
```
jwhui, line 2011:
```suggestion
void Interpreter::HandleLinkMetricsEnhAckProbingIe(otShortAddress       aShortAddress,
```

src/cli/cli.cpp
```diff
@@ -2003,6 +2008,29 @@ void Interpreter::HandleLinkMetricsMgmtResponse(const otIp6Address *aAddress, ui
...
+                                                   const otExtAddress *       aExtAddress,
+                                                   const otLinkMetricsValues *aMetricsValues,
+                                                   void *                     aContext)
+{
+    static_cast<Interpreter *>(aContext)->HandleLinkMetricsEnhAckProbingIe(aShortAddress, aExtAddress, aMetricsValues);
+}
+
+void Interpreter::HandleLinkMetricsEnhAckProbingIe(const otShortAddress       aShortAddress,
```
jwhui, line 2019:
```suggestion
void Interpreter::HandleLinkMetricsEnhAckProbingIe(otShortAddress       aShortAddress,
```

src/core/thread/link_metrics.hpp
```diff
@@ -202,26 +213,49 @@ class LinkMetrics : public InstanceLocator, private NonCopyable
      *                               accounted.
      * @param[in] aLinkMetricsFlags  A pointer to flags specifying what metrics to query.
      *
-     * @retval OT_ERROR_NONE          Successfully sent a Link Metrics Management Request message.
-     * @retval OT_ERROR_NO_BUFS       Insufficient buffers to generate the MLE Link Metrics Management Request message.
-     * @retval OT_ERROR_INVALID_ARGS  @p aSeriesId is not within the valid range.
+     * @retval OT_ERROR_NONE              Successfully sent a Link Metrics Management Request message.
+     * @retval OT_ERROR_NO_BUFS           Insufficient buffers to generate the MLE Link Metrics Management Request
+     * message.
```
jwhui, line 218: nit: alignment

src/core/thread/link_metrics.hpp
```diff
@@ -202,26 +213,49 @@ class LinkMetrics : public InstanceLocator, private NonCopyable
...
                                                  uint8_t                         aSeriesId,
                                                  const otLinkMetricsSeriesFlags &aSeriesFlags,
                                                  const otLinkMetrics *           aLinkMetricsFlags);
 
+    /**
+     * This method sends an MLE Link Metrics Management Request to configure/clear a Enhanced-ACK Based Probing.
+     *
+     * @param[in] aDestination       A reference to the IPv6 address of the destination.
+     * @param[in] aEnhAckFlags       Enh-ACK Flags to indicate whether to register or clear the probing. `0` to clear
+     * and `1` to register. Other values are reserved.
```
jwhui, line 233: nit: alignment

src/core/thread/link_metrics.hpp
```diff
@@ -202,26 +213,49 @@ class LinkMetrics : public InstanceLocator, private NonCopyable
...
+     * @param[in] aLinkMetricsFlags  A pointer to flags specifying what metrics to query. Should be `NULL` when
+     *                               `aEnhAckFlags` is `0`.
+     *
+     * @retval OT_ERROR_NONE              Successfully sent a Link Metrics Management Request message.
+     * @retval OT_ERROR_NO_BUFS           Insufficient buffers to generate the MLE Link Metrics Management Request
+     * message.
```
jwhui, line 239: nit: alignment

src/core/thread/link_metrics_tlvs.hpp
```diff
@@ -355,6 +360,33 @@ class SeriesFlags
      */
     SeriesFlags(const SeriesFlags &aSeriesFlags) { mSeriesFlags = aSeriesFlags.mSeriesFlags; }
 
+    /**
+     * This method sets the values of this object from a `otLinkMetricsSeriesFlags` object.
+     *
+     * @param[in]  aSeriesFlags  The `otLinkMetricsSeriesFlags` object.
+     *
+     */
+    void SetFromPublicSeriesFlags(const otLinkMetricsSeriesFlags aSeriesFlags)
```
jwhui, line 369:
```suggestion
    void SetFromOtSeriesFlags(otLinkMetricsSeriesFlags aSeriesFlags)
```

src/core/thread/topology.hpp
```diff
@@ -639,12 +647,42 @@ class Neighbor : public InstanceLocatorInit
...
      *
      */
     void RemoveAllForwardTrackingSeriesInfo(void);
 
-#endif
+    /**
+     * This method gets the Enh-ACK Probing metrics (this `Neighbor` object is the Probing Subject).
+     *
+     */
+    const otLinkMetrics &GetEnhAckProbingMetrics(void) const { return mEnhAckProbingMetrics; }
```
jwhui, line 659: nit: doc return value

src/core/thread/link_metrics.cpp
```diff
@@ -164,37 +173,76 @@ otError LinkMetrics::SendMgmtRequestForwardTrackingSeries(const Ip6::Address &
...
+otError LinkMetrics::SendMgmtRequestEnhAckProbing(const Ip6::Address &           aDestination,
+                                                  const otLinkMetricsEnhAckFlags aEnhAckFlags,
+                                                  const otLinkMetrics *          aLinkMetricsFlags)
+{
+    otError                error             = OT_ERROR_NONE;
+    LinkMetricsEnhAckFlags enhAckFlags       = static_cast<LinkMetricsEnhAckFlags>(aEnhAckFlags);
+    uint8_t                typeIdFlagsOffset = sizeof(Tlv) + sizeof(enhAckFlags);
+    uint8_t                typeIdFlagsCount  = 0;
+    uint8_t      subTlvs[sizeof(Tlv) + sizeof(enhAckFlags) + sizeof(LinkMetricsTypeIdFlags) * kMaxTypeIdFlags];
+    Tlv *        enhancedAckLinkMetricsConfigurationSubTlv = reinterpret_cast<Tlv *>(subTlvs);
```
jwhui, line 196: Not necessary for this PR, but I think this could be cleaned up by having a class that generates/parses Enhanced ACK Link Metrics Configuration Sub-TLVs.

Irving-cl: Added a class for generating Enhanced ACK Link Metrics Configuration Sub-TLVs.

Irving-cl:
Skipped LGTM.com here because it has some problems now: https://github.community/t/lgtm-cannot-spot-issue-in-checkout/148935

jwhui (approved):

PR #5784, opened 2020-11-07
[spinel] print error on RCP reset before dying

dismirlian:
This PR makes ot-daemon print the RCP reset cause before exit, in case the RCP resets unexpectedly.

### DIFF
--- a/src/lib/spinel/radio_spinel_impl.hpp
+++ b/src/lib/spinel/radio_spinel_impl.hpp
@@ -770,8 +770,12 @@ void RadioSpinel<InterfaceType, ProcessContextType>::HandleValueIs(spinel_prop_k
 
         if (status >= SPINEL_STATUS_RESET__BEGIN && status <= SPINEL_STATUS_RESET__END)
         {
-            // If RCP crashes/resets while radio was enabled, posix app exits.
-            VerifyOrDie(!IsEnabled(), OT_EXIT_RADIO_SPINEL_RESET);
+            if (IsEnabled())
+            {
+                // If RCP crashes/resets while radio was enabled, posix app exits.
+                otLogCritPlat("Unexpected RCP reset: %s", spinel_status_to_cstr(status));
+                DieNow(OT_EXIT_RADIO_SPINEL_RESET);
+            }
 
             otLogInfoPlat("RCP reset: %s", spinel_status_to_cstr(status));
             mIsReady = true;


bukepo (approved):
LGTM:+1:

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -770,8 +770,12 @@ void RadioSpinel<InterfaceType, ProcessContextType>::HandleValueIs(spinel_prop_k
 
         if (status >= SPINEL_STATUS_RESET__BEGIN && status <= SPINEL_STATUS_RESET__END)
         {
-            // If RCP crashes/resets while radio was enabled, posix app exits.
-            VerifyOrDie(!IsEnabled(), OT_EXIT_RADIO_SPINEL_RESET);
+            if (IsEnabled())
+            {
+                // If RCP crashes/resets while radio was enabled, posix app exits.
+                otLogCritPlat("Unexpected RCP reset: %s", spinel_status_to_cstr(status));
```
bukepo, line 776: VerifyOrDie() already prints the dying message.

dismirlian: Yes, but you don't get to know *why* the RCP was reset.

simonlingoogle (approved):
LGTM :+1:

src/lib/spinel/radio_spinel_impl.hpp
bukepo: I think `VerifyOrDie()` prints something like `HandleValueIs() at src/lib/spinel/radio_spinel_impl.hpp:774: RadioSpinelReset`. It already points out detailed information. `RadioSpinelReset` indicates unexpected reset encountered. How about renaming `OT_EXIT_RADIO_SPINEL_RESET` to `OT_EXIT_RADIO_SPINEL_UNEXPECTED_RESET` and update its string message to `RadioSpinelUnexpectedReset`?

abtink: At spinel status definition, we allow  bunch of (optional) different reset reasons (status values) to be provided by the platform layer, i.e. ASSERT, vs WATCHDOG, vs. POWER_ON, etc. (they would all be unexpected reset).
I agree that it would be helpful to log it separately to be able to log the reason as well.

abtink (approved):

abtink (approved):
LGTM. Thanks.

src/lib/spinel/radio_spinel_impl.hpp
dismirlian: @bukepo Yes, that is exactly what `VerifyOrDie()` prints now. We are missing the detail on why the RCP has reset unexpectedly (eg. RESET_EXTERNAL vs RESET_ASSERT), as this helps developers debug their code. This is exactly what I was missing in my particular case to diagnose a problem with unexpected resets; we were able to rule out software causes, and focus on a hardware issue.

dismirlian: Also, I agree that `RadioSpinelUnexpectedReset` is more explicit than plain `RadioSpinelReset`.

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #5785, opened 2020-11-08
[low-power] fix csl simulation

### LINKED ISSUE #5783: [thread-cert] LowPower_5_3_01 failed when VIRTUAL_TIME=0
**Describe the bug**
When running the case `v1_2_LowPower_5_3_01_SSEDAttachment.py` with `VIRTUAL_TIME=0` it fails. Currently the cases are all run with `VIRTUAL_TIME=0`. Currently our tests are all run with `VIRTUAL_TIME=1`.
 
```
THREAD_VERSION=1.2 VIRTUAL_TIME=0 CSL_RECEIVER=1 ./script/test build cert tests/scripts/thread-cert/v1_2_LowPower_5_3_01_SSEDAttachment.py
```

1. Git commit id: f886f34d0
2. IEEE 802.15.4 hardware platform: simulation
3. Build steps: `THREAD_VERSION=1.2 VIRTUAL_TIME=0 CSL_RECEIVER=1 ./script/test build`

Irving-cl:
This PR fixes #5783.

The cause is that, on nRF platform, `otPlatRadioGetNow` always gets the LSB part of an `uint64_t` time. However, on simulation platform, this is not the case. So in
```
            if (otPlatRadioGetNow(&GetInstance()) <
                mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime + mTransmitFrame.mInfo.mTxInfo.mTxDelay)
```
we don't get the result expected. `otPlatRadioGetNow(&GetInstance())` is a very big number on simulation platform, and `mTxDelayBaseTime` is the LSB part of the rx time.

### DIFF
--- a/src/core/mac/sub_mac.cpp
+++ b/src/core/mac/sub_mac.cpp
@@ -356,8 +356,8 @@ void SubMac::StartCsmaBackoff(void)
 
         if (ShouldHandleTransmitTargetTime())
         {
-            if (otPlatRadioGetNow(&GetInstance()) <
-                mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime + mTransmitFrame.mInfo.mTxInfo.mTxDelay)
+            if (Time(static_cast<uint32_t>(otPlatRadioGetNow(&GetInstance()))) <
+                Time(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime) + mTransmitFrame.mInfo.mTxInfo.mTxDelay)
             {
                 mTimer.StartAt(Time(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime),
                                mTransmitFrame.mInfo.mTxInfo.mTxDelay);


src/core/mac/sub_mac.cpp
```diff
@@ -356,7 +356,7 @@ void SubMac::StartCsmaBackoff(void)
 
         if (ShouldHandleTransmitTargetTime())
         {
-            if (otPlatRadioGetNow(&GetInstance()) <
+            if (static_cast<uint32_t>(otPlatRadioGetNow(&GetInstance())) <
```
bukepo, line 359:
```suggestion
            if (static_cast<int32_t>(static_cast<uint32_t>(otPlatRadioGetNow(&GetInstance())) - static_cast<uint32_t>(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime + mTransmitFrame.mInfo.mTxInfo.mTxDelay)) < 0)
```

brianljt (approved):
LGTM.

src/core/mac/sub_mac.cpp
Irving-cl: What about:
```
if (static_cast<uint32_t>(otPlatRadioGetNow(&GetInstance()) < 
    static_cast<uint32_t>(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime + mTransmitFrame.mInfo.mTxInfo.mTxDelay))
```
Is there a difference between this and your suggection?

bukepo: Will `mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime + mTransmitFrame.mInfo.mTxInfo.mTxDelay` overflow and wrap?

Irving-cl: Use `Time` operator to compare now.

src/core/mac/sub_mac.cpp
```diff
@@ -356,8 +356,8 @@ void SubMac::StartCsmaBackoff(void)
 
         if (ShouldHandleTransmitTargetTime())
         {
-            if (otPlatRadioGetNow(&GetInstance()) <
-                mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime + mTransmitFrame.mInfo.mTxInfo.mTxDelay)
+            if (Time(static_cast<uint32_t>(otPlatRadioGetNow(&GetInstance()))) <
+                Time(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime + mTransmitFrame.mInfo.mTxInfo.mTxDelay))
```
bukepo, line 360:
```suggestion
                Time(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime) + Time(mTransmitFrame.mInfo.mTxInfo.mTxDelay))
```

Irving-cl: Note that the `+` operator of `Time` is:
```
Time operator+(uint32_t aDuration) const { return Time(mValue + aDuration); } 
```
There isn't a version that accepts a `Time` type as parameter.

We can change it to:
```
Time(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime) + mTransmitFrame.mInfo.mTxInfo.mTxDelay
```

bukepo (approved):
LGTM:+1:

librasungirl (approved):

simonlingoogle (approved):
LGTM :+1:

src/core/mac/sub_mac.cpp
```diff
@@ -356,8 +356,8 @@ void SubMac::StartCsmaBackoff(void)
 
         if (ShouldHandleTransmitTargetTime())
         {
-            if (otPlatRadioGetNow(&GetInstance()) <
-                mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime + mTransmitFrame.mInfo.mTxInfo.mTxDelay)
+            if (Time(static_cast<uint32_t>(otPlatRadioGetNow(&GetInstance()))) <
+                Time(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime) + mTransmitFrame.mInfo.mTxInfo.mTxDelay)
```
abtink, line 360: 👍  on the conversion to `Time()` here to handle roll-over of `uint32_t` (be able to use `<` safely).

abtink (approved):
LGTM. 👍

jwhui (approved):

abtink:
I think this fix/change is good. 

@Irving-cl , a suggestion for longer-term: Can we consider changing the `mTxDelayBaseTime` to be `uint64` and same size as `otPlatRadioGetNow()`?  

I don't know how big of a change (how disruptive) it may be but if we can change it I think it would be good. The main advantage is that it would make it easier to use (less chance of mistake). With it being `uint32_`t the radio platform needs to also be careful of handling the roll-over correctly. 

Thoughts?

Irving-cl:
@abtink 

I think it would be a good idea. I added the field `mTxDelayBaseTime` in #5545. I chose `uint32_t` because as discussed(https://github.com/openthread/openthread/pull/5545#issuecomment-712138316), the platform only uses the LSB-part of the base time. Our timers (TimerMilli, TimerMicro) are also like that. 

I agree that changing these things to `uint64_t` makes it less possible for mistakes. Just note that when some platform (nrf does) or our timers uses the base time, the base time is converted to `uint32_t` (using the LSB part). I think it's fine. Thoughts?

PR #5791, opened 2020-11-10
[posix] reduce RCP time sync log

brianljt:


src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1805,6 +1803,9 @@ void RadioSpinel<InterfaceType, ProcessContextType>::CalcRcpTimeOffset(void)
      */
 
     VerifyOrExit(!mIsTimeSynced || (otPlatTimeGet() >= GetNextRadioTimeRecalcStart()));
+
+    otLogInfoPlat("Trying to get RCP time offset");
```
bukepo, line 1807:
```suggestion
    otLogDebgPlat("Trying to get RCP time offset");
```

Irving-cl (approved):
LGTM.

bukepo (approved):
LGTM:+1:

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1805,6 +1803,9 @@ void RadioSpinel<InterfaceType, ProcessContextType>::CalcRcpTimeOffset(void)
      */
 
     VerifyOrExit(!mIsTimeSynced || (otPlatTimeGet() >= GetNextRadioTimeRecalcStart()));
```
simonlingoogle, line 1805: Not directly related to this PR. But why do we need both `mIsTimeSynced` and `mRadioTimeRecalcStart`? 
Can we just set `mRadioTimeRecalcStart` to 0 initially and remove `mIsTimeSynced`?

simonlingoogle (approved):
LGTM :+1: 
with one question

src/lib/spinel/radio_spinel_impl.hpp
brianljt: If 1.2 is not enabled, `GetNow()` should return an error number as timer is not synced.

simonlingoogle:
```c++
    mRadioTimeRecalcStart = localRxTimestamp + RCP_TIME_OFFSET_CHECK_INTERVAL;
```

Seems we only store local time in mRadioTimeRecalcStart?

brianljt: I don't get your question here. `mIsTimeSynced` is used to indicate whether the value in `mRadioTimeOffset` is valid. `mRadioTimeRecalcStart` is only used to set next resync time.

simonlingoogle: I mean if we initialize mRadioTimeRecalcStart=0 , we can simplify the condition to:
```c++
VerifyOrExit(otPlatTimeGet() >= GetNextRadioTimeRecalcStart());
```

brianljt: Then again, in `GetNow()`, there is no variable to check if `mRadioTimeOffset` is a valid value. `mRadioTimeOffset` might not be valid because,
1. OT1.2 is not enabled.
2. Some error happens during `CalcRcpTimeOffset`

simonlingoogle: Indeed `mIsTimeSynced` is useful in GetNow(). SGTM.

jwhui (approved):

PR #5792, opened 2020-11-10
[memory] use heap for messages

bukepo:
This commit adds an option to allow share heap memory with message
buffer pool.

For MTD, there is usually no border agent or commissioner service. This
change allows the heap for meshcop to be used by application or network
management data after the device is attached. Thus the memory footprint
can be reduced.


src/core/common/message.cpp
```diff
@@ -133,7 +142,9 @@ void MessagePool::FreeBuffers(Buffer *aBuffer)
     while (aBuffer != nullptr)
     {
         Buffer *next = aBuffer->GetNextBuffer();
-#if OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT
+#if OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE
+        GetInstance().HeapFree(aBuffer);
+#elif OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT
```
simonlingoogle, line 147: OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT should be of highest priority?

src/core/common/message.cpp
```diff
@@ -143,30 +154,34 @@ void MessagePool::FreeBuffers(Buffer *aBuffer)
...
 }
+#endif
 
 uint16_t MessagePool::GetFreeBufferCount(void) const
 {
     uint16_t rval;
 
-#if OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT
+#if OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE
+    rval = static_cast<uint16_t>(GetInstance().GetHeap().GetFreeSize() / sizeof(Buffer));
```
simonlingoogle, line 183: This is not necessarily correct when using Heap because there might be fragmentation? Even when the free size is larger than sizeof(Buffer), we might still unable to allocate one. 

This also makes the method `MessagePool::ReclaimBuffers(int aNumBuffers, Message::Priority aPriority)` unable to guarantee that there are enough message buffers available.

src/core/common/message.cpp
```diff
@@ -254,23 +269,9 @@ otError Message::SetLength(uint16_t aLength)
...
-    {
-        bufs = (((totalLengthRequest - kHeadBufferDataSize) - 1) / kBufferDataSize) + 1;
-    }
-
-    if (totalLengthCurrent > kHeadBufferDataSize)
-    {
-        bufs -= (((totalLengthCurrent - kHeadBufferDataSize) - 1) / kBufferDataSize) + 1;
-    }
-
-    SuccessOrExit(error = GetMessagePool()->ReclaimBuffers(bufs, GetPriority()));
```
simonlingoogle, line 272: why is it not necessary anymore?

simonlingoogle (commented):
Do we need to also adjust OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE_NO_DTLS and OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE ?

examples/common-switches.mk
```diff
@@ -61,6 +61,7 @@ LOG_OUTPUT          ?= APP
 endif
 LINK_RAW            ?= 0
 MAC_FILTER          ?= 0
+MESSAGE_USE_HEAP    ?= 0
```
simonlingoogle, line 64: nit: what about `ALLOC_MESSAGES` ?

src/core/common/message.cpp
librasungirl: Pre-reclaim here does not make much difference from per buffer reclaim when required in `ResizeMessage()`. @simonlingoogle 

Just for discussion:
- One possible optimization is to make this piece of code and `MessagePool::ReclaimBuffers(int aNumBuffers, Message::Priority aPriority)` more meaningful is to add one method to get the available buffers evict-able, and abort earlier if (the available evict-able buffers + existing free buffers) could not meet the buffer requirements, thus avoid meaningless eviction.
- Is it necessary to keep `MessagePool::ReclaimBuffers(int aNumBuffers, Message::Priority aPriority)` if given existing per buffer reclaim/allocation implementation?

Thoughts? @bukepo @simonlingoogle

librasungirl: When reclaimation is required, pre-reclaimation here doesn't make much difference from per buffer reclaimation in `ResizeMessage()`

bukepo: I don't think the current heap allocator can give a precise number of buffers it can allocate.

librasungirl: To make the pre-check here more useful is to introduce some method to check whether or not evict-able buffers could meet the requirement before eviction if the existing available free buffers could not meet the requirement I think, thus avoiding meaningless eviction if it is predictable it could not meet the buffer requirement.

src/core/common/message.cpp
```diff
@@ -103,22 +104,25 @@ Buffer *MessagePool::NewBuffer(Message::Priority aPriority)
...
+#endif
+                   ) == nullptr)
+    {
+        SuccessOrExit(ReclaimBuffers(aPriority));
+    }
 
-    buffer = mBufferPool.Allocate();
-    VerifyOrExit(buffer != nullptr);
-
+#if !OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE
```
librasungirl, line 120:
```suggestion
#if !OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT && !OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE
```

bukepo:
> Do we need to also adjust OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE_NO_DTLS and OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE ?

It seems to be less efficient to use heap for message buffers when DTLS is not needed.

src/core/common/message.cpp
simonlingoogle: > I don't think the current heap allocator can give a precise number of buffers it can allocate.

It should be possible to enhance the allocator to do that?

@librasungirl Predicting if if eviction should be possible. If we can get the number of message buffers that the current heap can allocate (N), then the max number of message buffers we can get after eviction is `N + Number of message buffers of lower priority`.

src/core/common/message.cpp
bukepo: Yes, that's one of the reason removing the use of in SetLength().

bukepo: Yes, that's one of the reasons removing the use of in SetLength().

simonlingoogle:
> > Do we need to also adjust OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE_NO_DTLS and OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE ?
> 
> It seems to be less efficient to use heap for message buffers when DTLS is not needed.

Yes. But if DTLS is used, do we need to increase `OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE` if we allocate message buffers on heap?

src/core/common/message.cpp
bukepo: updated to detect configuration errors.

bukepo:
> > > Do we need to also adjust OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE_NO_DTLS and OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE ?
> > 
> > 
> > It seems to be less efficient to use heap for message buffers when DTLS is not needed.
> 
> Yes. But if DTLS is used, do we need to increase `OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE` if we allocate message buffers on heap?

It depends on the application and platform. The default value at least ensures it passes all CI tests, and that's how the original value for `OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE` was chosen.

src/core/common/message.cpp
```diff
@@ -143,30 +149,18 @@ void MessagePool::FreeBuffers(Buffer *aBuffer)
...
-exit:
-#else
-    OT_UNUSED_VARIABLE(aPriority);
-#endif
-
-    // First comparison is to get around issues with comparing
-    // signed and unsigned numbers, if aNumBuffers is negative then
-    // the second comparison wont be attempted.
-    return (aNumBuffers < 0 || aNumBuffers <= GetFreeBufferCount()) ? OT_ERROR_NONE : OT_ERROR_NO_BUFS;
+    return Get<MeshForwarder>().EvictMessage(aPriority);
```
zhanglongxia, line 154: The function `EvictMessage` only removes a message. The space left after releasing a message may not be enough for the new message. Especially when the traffic is heavy, it may cause insufficient space to create a MLE message. Which may cause the link to be broken

src/core/common/message.cpp
```diff
@@ -103,22 +104,25 @@ Buffer *MessagePool::NewBuffer(Message::Priority aPriority)
 {
     Buffer *buffer = nullptr;
 
-    SuccessOrExit(ReclaimBuffers(1, aPriority));
-
-#if OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT
-
-    buffer = static_cast<Buffer *>(otPlatMessagePoolNew(&GetInstance()));
-
+    while ((
```
Irving-cl, line 107: Why using `while` here instead `if`? It seems that the line `SuccessOrExit(ReclaimBuffers(aPriority));` won't execute for a second time. And even it does, the result won't change, then an infinite loop occurs.

Irving-cl (commented):
Do we need to add the option for gn build?

src/core/common/message.cpp
Irving-cl: it's wrapped in a `ReclaimBuffer` method and the method is used in a while loop until enough buffer is got.

examples/common-switches.mk
bukepo: It seems alloc is too general. Thoughts?

bukepo:
> Do we need to add the option for gn build?

Sure!

bukepo:
Something went wrong for cli-mtd reporting status, so re-triggered the CI.

src/core/common/message.cpp
abtink: IMO, I think this is fine as is, but I like @librasungirl  suggestion above (possibly as a future enhancement).

src/core/common/message.cpp
```diff
@@ -41,11 +41,12 @@
 #include "net/checksum.hpp"
 #include "net/ip6.hpp"
 
+#if OPENTHREAD_MTD || OPENTHREAD_FTD
```
abtink, line 44: Do we need this check? Is it required for heap use?

src/core/config/openthread-core-config-check.h
```diff
@@ -546,4 +546,13 @@
        "(and OPENTHREAD_CONFIG_LOG_DEFINE_AS_MACRO_ONLY)"
 #endif
 
+#if OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE
+#if !OPENTHREAD_CONFIG_DTLS_ENABLE
+#warning "OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE is strongly discouraged when OPENTHREAD_CONFIG_DTLS_ENABLE is off."
```
abtink, line 551: 👍

src/core/common/message.hpp
```diff
@@ -1471,9 +1471,9 @@ class MessagePool : public InstanceLocator, private NonCopyable
 private:
     Buffer *NewBuffer(Message::Priority aPriority);
     void    FreeBuffers(Buffer *aBuffer);
-    otError ReclaimBuffers(int aNumBuffers, Message::Priority aPriority);
+    otError ReclaimBuffers(Message::Priority aPriority);
 
-#if OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT == 0
+#if OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT == 0 && OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE == 0
```
abtink, line 1476: I know this is existing code, but since we are changing this line here, how about changing it to use `!` instead of `== 0` ?

We more often use the `!OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT` pattern. I pesonally prefer it and feel it is more readable compared to `== 0` check.

abtink (approved):
Looks good. Thanks 👍

src/core/common/message.cpp
bukepo: No, just moved the check from `otError MessagePool::ReclaimBuffers(int aNumBuffers, Message::Priority aPriority)` here. Because Message is not used at all on RCP.

src/core/common/message.hpp
simonlingoogle: agree

src/core/common/message.cpp
simonlingoogle: But should we keep the original code? @abtink @bukepo @librasungirl

bukepo: ResizeMessage() should be simplified if we make sure memory is enough before allocating. The code is well kept in history, so I tend not to keep it just for future reference.

zhanglongxia (approved):
LGTM

src/core/common/message.cpp
simonlingoogle: Even with the heap allocator, if `ReclaimBuffers` failed, we **can be sure** that we are not going to be able to `ResizeMessage`. 

Furthermore, evicting messages of lower priority can still increase the chance of having enough buffer for a higher-priority message. The the code is still helpful even for `MESSAGE_USE_HEAP=ON`

Not to mention, `MESSAGE_USE_HEAP=ON` is not the only implementation. When `MESSAGE_USE_HEAP=OFF`, the original code works just fine. Why remove the code if it's helpful in both cases? 

Thoughts?

librasungirl: >  if ReclaimBuffers failed, we can be sure that we are not going to be able to ResizeMessage.

Why requiring knowing `ReclaimBuffers failed` in advance before `ResizeMessage`? 

There is `ReclaimBuffer` when requesting new buffer in ResizeMessage() if required, and NO_BUFS error would be returned if the requirement could not be met.

librasungirl (approved):

simonlingoogle (approved):
LGTM :+1: 

This is a great memory optimization!

Irving-cl (approved):
LGTM :+1:

bukepo:
Looks like there is something wrong with the GitHub Actions. All tests passed, but an old `cli-mtd` job was not correctly removed. I tried re-opening and pushing but in vain.

simonlingoogle:
> Looks like there is something wrong with the GitHub Actions. All tests passed, but an old `cli-mtd` job was not correctly removed. I tried re-opening and pushing but in vain.

It happens when we remove or add Github Action tests.  It's normal.

src/core/config/openthread-core-config-check.h
simonlingoogle: I am afraid that the warning might not be noticed by users. 
Since we haven't set proper heap size for each platform when DTLS is OFF, maybe it's safer to prohibit using `MESSAGE_USE_HEAP` in such case?

Not to mention, using `MESSAGE_USE_HEAP` when DTLS is OFF is basically not helpful.

abtink: I agree. `#error` here makes more sense to me as well.

jwhui: @bukepo , thoughts?

bukepo:
```suggestion
#error "OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE is strongly discouraged when OPENTHREAD_CONFIG_DTLS_ENABLE is off."
```

jwhui (approved):

PR #3130, opened 2018-10-08
[core] remove dependency on libstdc++

bukepo:
Some platforms don't have libstdc++

* removed pure virtual function `__cxa_pure_virtual`
* added option use `-fno-exceptions -fno-rtti`

PR #5793, opened 2020-11-10
[mlr] implement native multicast routing for backbone router

simonlingoogle:
This commit implements native multicast routing using kernel Ip6 multicast forwarding. 

Backbone tests use latest commit of https://github.com/openthread/ot-br-posix/pull/604 to pass MLR checks.


src/posix/main.c
```diff
@@ -101,15 +101,13 @@ void __gcov_flush();
...
-#if OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
     OT_POSIX_OPT_BACKBONE_INTERFACE_NAME = 'B',
-#endif
-    OT_POSIX_OPT_DEBUG_LEVEL    = 'd',
-    OT_POSIX_OPT_DRY_RUN        = 'n',
-    OT_POSIX_OPT_HELP           = 'h',
-    OT_POSIX_OPT_INTERFACE_NAME = 'I',
-    OT_POSIX_OPT_TIME_SPEED     = 's',
-    OT_POSIX_OPT_VERBOSE        = 'v',
+    OT_POSIX_OPT_DEBUG_LEVEL             = 'd',
```
simonlingoogle, line 105: Explain:
The change of this file is not mandatory for this PR. However, the original code is incorrect because `OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE` is never defined in main.c. This change allows us to test Backbone features using `ot-cli` program.

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,563 @@
...
+    FinalizeMulticastRouterSock();
+
+exit:
+    otLogResultPlat(error, "MulticastRoutingManager: %s", __FUNCTION__);
+}
+
+void MulticastRoutingManager::Add(const Ip6::Address &aAddress)
+{
+    assert(mListenerSet.find(aAddress) == mListenerSet.end());
+    mListenerSet.insert(aAddress);
```
librasungirl, line 110: why update `mListenerSet` regardless of `IsEnabled()`?

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,563 @@
...
+    struct mf6cctl mf6cctl;
+    MifIndex       forwardMif = kMifIndexNone;
+
+    VerifyOrExit(aIif == kMifIndexThread || aIif == kMifIndexBackbone, error = OT_ERROR_INVALID_ARGS);
+
+    ExpireMulticastForwardingCache();
+
+    if (aIif == kMifIndexBackbone)
+    {
+        // Forward multicast traffic form Backbone to Thread if the group address is subscribed by MLR.
```
librasungirl, line 260:
```suggestion
        // Forward multicast traffic from Backbone to Thread if the group address is subscribed by any Thread device via MLR.
```

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,563 @@
...
+        if (mListenerSet.find(aGroupAddr) != mListenerSet.end())
+        {
+            forwardMif = kMifIndexThread;
+        }
+    }
+    else
+    {
+        // Forward multicast traffic from Thread to Backbone if multicast scope > kRealmLocalScope
```
librasungirl, line 268: What's the plan for Outbound-forwarding flag and Admin-local forwarding flag defined in 9.4.7.1 Multicast Forwarding Configuration?

src/posix/platform/multicast_routing.hpp
```diff
@@ -0,0 +1,194 @@
...
+    void    Disable(void);
+    void    Add(const Ip6::Address &aAddress);
+    void    Remove(const Ip6::Address &aAddress);
+    bool    IsEnabled(void) const { return mMulticastRouterSock >= 0; }
+    otError InitMulticastRouterSock(void);
+    void    FinalizeMulticastRouterSock(void);
+    void    ProcessMulticastRouterMessages(void);
+    otError AddMulticastForwardingCache(const Ip6::Address &aSrcAddr, const Ip6::Address &aGroupAddr, MifIndex aIif);
+    void    RemoveInboundMulticastForwardingCache(const Ip6::Address &aGroupAddr);
+    void    UnblockInboundMulticastForwardingCache(const Ip6::Address &aGroupAddr);
```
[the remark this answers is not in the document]
librasungirl: nit: is it necessary to use `Inbound` in the name?
It seems to me `bi-direction` forwarding here.

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,563 @@
...
+    memset(&mf6cctl, 0, sizeof(mf6cctl));
+
+    memset(&sioc_sg_req6, 0, sizeof(sioc_sg_req6));
+
+    for (std::map<MulticastRoute, MulticastRouteInfo>::const_iterator it = mMulticastForwardingCache.begin();
+         it != mMulticastForwardingCache.end();)
+    {
+        otError               error;
+        const MulticastRoute &route     = it->first;
+        MulticastRouteInfo    routeInfo = it->second;
```
librasungirl, line 404:
```suggestion
        MulticastRouteInfo   & routeInfo = it->second;
```

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,563 @@
...
+                                                             const Ip6::Address &aGroupAddr,
+                                                             MifIndex            aIif)
+{
+    otError        error = OT_ERROR_NONE;
+    struct mf6cctl mf6cctl;
+    MifIndex       forwardMif = kMifIndexNone;
+
+    VerifyOrExit(aIif == kMifIndexThread || aIif == kMifIndexBackbone, error = OT_ERROR_INVALID_ARGS);
+
+    ExpireMulticastForwardingCache();
```
librasungirl, line 256: It seems to me that now stale multicast route entries is only removed when there is new multicast traffic?
It might be more reasonable to flush stale entries periodically?

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,563 @@
...
+    for (const std::pair<const MulticastRoute, MulticastRouteInfo> &entry : mMulticastForwardingCache)
+    {
+        otError                   error;
+        const MulticastRoute &    route     = entry.first;
+        const MulticastRouteInfo &routeInfo = entry.second;
+
+        if (routeInfo.mIif != kMifIndexBackbone || routeInfo.mOif == kMifIndexThread || route.mGroupAddr != aGroupAddr)
+        {
+            continue;
+        }
```
librasungirl, line 319: what's the purpose of this `for` loop?
When adding new listener, it seems to me not possible that such entry exist already.

src/posix/main.c
simonlingoogle: I guess `posix/main.c` is actually a application code thus can not get the core configuration?
Any comment on this? @bukepo

src/posix/platform/multicast_routing.cpp
simonlingoogle: I am assuming that it's possible to for SBBR to add multicast listener (at least in the future). So the current implementation does not assume that Multicast Listener is always added when it's Primary. But the `MulticastRoutingManager` is only enabled when it's Primary.

```
Spec 9.4.8.3.1:

Upon a successful registration of a Multicast Listener, the BBR sends out a BMLR.ntf
multicast on the Backbone Link interface to propagate all successfully processed multicast
registration entries to Secondary BBRs on the same Thread Network.
```

Note that the logic is same as before when using smcroute.

src/posix/platform/multicast_routing.hpp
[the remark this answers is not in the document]
simonlingoogle: The current `Block` and `Remove` is actually accurate. The `Block` method does not add any MFC but unblock existing MFCs. Thus I tend to keep the current names.

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,564 @@
...
+bool MulticastRoutingManager::UpdateMulticastRouteInfo(const MulticastRoute &route)
+{
+    bool                updated = false;
+    struct sioc_sg_req6 sioc_sg_req6;
+    memset(&sioc_sg_req6, 0, sizeof(sioc_sg_req6));
+
+    memcpy(sioc_sg_req6.src.sin6_addr.s6_addr, route.mSrcAddr.GetBytes(), sizeof(sioc_sg_req6.src.sin6_addr.s6_addr));
+    memcpy(sioc_sg_req6.grp.sin6_addr.s6_addr, route.mGroupAddr.GetBytes(), sizeof(sioc_sg_req6.grp.sin6_addr.s6_addr));
+
+    MulticastRouteInfo &routeInfo = mMulticastForwardingCache[route];
```
librasungirl, line 462: nit: move forward

https://github.com/openthread/openthread/blob/master/STYLE_GUIDE.md
> All scoped (i.e. stack) variable declarations should be placed together at the top of the enclosing scope in which they are used.
There shall be an empty line after all such variable declarations.

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,564 @@
...
+        unsigned long validPktCnt;
+
+        otLogDebgPlat("MulticastRoutingManager: %s: SIOCGETSGCNT_IN6 %s => %s: bytecnt=%lu, pktcnt=%lu, wrong_if=%lu",
+                      __FUNCTION__, route.mSrcAddr.ToString().AsCString(), route.mGroupAddr.ToString().AsCString(),
+                      sioc_sg_req6.bytecnt, sioc_sg_req6.pktcnt, sioc_sg_req6.wrong_if);
+
+        validPktCnt = sioc_sg_req6.pktcnt - sioc_sg_req6.wrong_if;
+        if (validPktCnt != routeInfo.mValidPktCnt)
+        {
+            routeInfo.mValidPktCnt = sioc_sg_req6.pktcnt;
```
librasungirl, line 475: typo?
```suggestion
            routeInfo.mValidPktCnt = validPktCnt;
```

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,564 @@
...
+        const MulticastRoute &    route     = entry.first;
+        const MulticastRouteInfo &routeInfo = entry.second;
+
+        otLogDebgPlat("MulticastRoutingManager: %s %s => %s %s", MifIndexToString(routeInfo.mIif),
+                      route.mSrcAddr.ToString().AsCString(), route.mGroupAddr.ToString().AsCString(),
+                      MifIndexToString(routeInfo.mOif));
+    }
+
+    otLogDebgPlat("MulticastRoutingManager: ========================================================",
+                  mMulticastForwardingCache.size());
```
librasungirl, line 527: missing `"%d"`?

src/posix/platform/multicast_routing.hpp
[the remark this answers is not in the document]
simonlingoogle: I think `Unblock`, which is also used by smcroute, is better because people are usually unaware of `mrt->mfc_unres_queue` of the kernel.

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,564 @@
...
+    {
+        FinalizeMulticastRouterSock();
+    }
+
+    return error;
+}
+
+void MulticastRoutingManager::FinalizeMulticastRouterSock(void)
+{
+    VerifyOrExit(mMulticastRouterSock != -1);
```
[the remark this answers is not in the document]
simonlingoogle: The current implementation is `{ return mMulticastRouterSock >= 0; }`, but it could change in the future. 

`FinalizeMulticastRouterSock` focus on the closing the socket if it's opened. I prefer `VerifyOrExit(mMulticastRouterSock != -1);` since it clearly states the condition of this method.

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,564 @@
...
+        {
+            mMulticastForwardingCache.erase(it++);
+        }
+        else
+        {
+            it++;
+        }
+    }
+
+    mMulticastForwardingCache.clear();
```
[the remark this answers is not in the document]
simonlingoogle: Good catch!

The old implementation of this method is removing all MFCs. I failed to remove this line of code. :-1:

src/posix/platform/multicast_routing.cpp
simonlingoogle: Good catch!

src/posix/platform/multicast_routing.cpp
simonlingoogle: > It may require additional efforts for this optional feature far beyond update the listener set here comparatively.
Yes. But I expect SBBR will also add Multicast Listener just like PBBR does. The difference is that MulticastRoutingManager is not enabled for SBBR. So, whether or not `MulticastRoutingManager` is enabled, it tries to maintain `mListenerSet` so that it's exactly same as the Multicast Listeners Table.

> Besides, one concern is there might be some inconsistence between mListenerSet here and PBBR's multicast listener table if there is some primary/secondary state switch, might bring assert in Line 109.

Yes, this was my concern too. Actually the purpose of adding the address to `mListenerSet` even if `MulticastRoutingManager` is not enabled is to **make sure mListenerSet and Multicast Listeners Table are exactly same**.

src/posix/platform/multicast_routing.cpp
simonlingoogle: `expire per 60s` added

librasungirl (approved):
LGTM overall with one minor comments.

src/core/net/ip6_address.hpp
```diff
@@ -533,6 +533,15 @@ class Address : public otIp6Address, public Equatable<Address>, public Clearable
...
     const uint8_t *GetBytes(void) const { return mFields.m8; }
 
+    /**
+     * This method sets the IPv6 address from a given byte array.
+     *
+     * @param[in] aBuffer    Pointer to an array containing the IPv6 address. `kSize` bytes from the buffer
+     *                       are copied to form the IPv6 address.
+     *
+     */
+    void SetBytes(const uint8_t *aBuffer) { memcpy(mFields.m8, aBuffer, kSize); }
```
bukepo, line 543:
```suggestion
    void SetBytes(const uint8_t *aBytes) { memcpy(mFields.m8, aBytes, kSize); }
```

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,570 @@
...
+void MulticastRoutingManager::Enable(void)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(!IsEnabled());
+
+    SuccessOrExit(error = InitMulticastRouterSock());
+
+exit:
+    otLogResultPlat(error, "MulticastRoutingManager: %s", __FUNCTION__);
```
bukepo:
```suggestion
    otLogResultPlat(error, "MulticastRoutingManager: %s", __FUNCTION__);
exit:
    return;
```

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,570 @@
...
+    case OT_BACKBONE_ROUTER_MULTICAST_LISTENER_ADDED:
+        Add(aAddress);
+        break;
+    case OT_BACKBONE_ROUTER_MULTICAST_LISTENER_REMOVED:
+        Remove(aAddress);
+        break;
+    }
+}
+
+void MulticastRoutingManager::Enable(void)
```
bukepo, line 83: Bubble up the error?

src/posix/platform/multicast_routing.cpp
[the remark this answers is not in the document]
simonlingoogle: revised

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,560 @@
...
+exit:
+    otLogResultPlat(error, "MulticastRoutingManager: %s: %s", __FUNCTION__, aAddress.ToString().AsCString());
+}
+
+void MulticastRoutingManager::UpdateFdSet(fd_set &aReadFdSet, int &aMaxFd) const
+{
+    VerifyOrExit(IsEnabled());
+
+    FD_SET(mMulticastRouterSock, &aReadFdSet);
+    aMaxFd = std::max(aMaxFd, mMulticastRouterSock);
```
bukepo, line 130: avoid std::max since we intentionally removed dependency on stdc++ in #3130.

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,560 @@
...
+    struct mif6ctl      mif6ctl;
+
+    VerifyOrExit(!IsEnabled());
+
+    // Create a Multicast Routing socket
+    mMulticastRouterSock = socket(AF_INET6, SOCK_RAW, IPPROTO_ICMPV6);
+    VerifyOrExit(mMulticastRouterSock != -1, error = OT_ERROR_FAILED);
+
+    // Enable Multicast Forwarding in Kernel
+    VerifyOrExit(0 == setsockopt(mMulticastRouterSock, IPPROTO_IPV6, MRT6_INIT, &one, sizeof(one)),
```
bukepo, line 165: VerifyOrDie? It seems to be a critical error.

src/posix/platform/multicast_routing.hpp
```diff
@@ -0,0 +1,191 @@
...
+private:
+    enum
+    {
+        kMulticastForwardingCacheExpireTimeout    = 300, //< Expire timeout of Multicast Forwarding Cache (in seconds)
+        kMulticastForwardingCacheExpiringInterval = 60,  //< Expire interval of Multicast Forwarding Cache (in seconds)
+    };
+
+    enum MifIndex : uint8_t
+    {
+        kMifIndexNone     = 0xff,
```
bukepo, line 112: Is it possible to reuse otNetifIdentifier?

src/posix/platform/multicast_routing.hpp
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        bool operator<(const MulticastRoute &aOther) const;
+
+    private:
+        Ip6::Address mSrcAddr;
+        Ip6::Address mGroupAddr;
+    };
+
+    using Clock = std::chrono::steady_clock;
```
bukepo, line 135: avoid use stdc++?

src/posix/platform/platform-posix.h
```diff
@@ -428,7 +462,21 @@ extern char gBackboneNetifName[IFNAMSIZ];
...
+    {                                                                                  \
+        otError _err = (aError);                                                       \
+                                                                                       \
+        if (_err == OT_ERROR_NONE)                                                     \
+        {                                                                              \
+            otLogInfoPlat(aFormat ": %s", ##__VA_ARGS__, otThreadErrorToString(_err)); \
+        }                                                                              \
+        else                                                                           \
+        {                                                                              \
+            otLogCritPlat(aFormat ": %s", ##__VA_ARGS__, otThreadErrorToString(_err)); \
```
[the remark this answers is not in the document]
simonlingoogle: I think `WARN` usually means the functionality is still working but with some restrictions. 
The log level should be `ERROR` because if it failed, the multicast routing is not working. 
But we don't have `ERROR`.

src/posix/platform/multicast_routing.cpp
simonlingoogle: outdated

src/posix/platform/multicast_routing.hpp
simonlingoogle: It does not make absolute sense to use otNetifIdentifier. The MifIndex is a index to the MIF in the kernel. We used 0 for Thread and 1 for Backbone. These values are different from `otNetifIdentifier` values. 
Since they are different values, I don't feel the necessity to reuse otNetifIdentifier.

BTW, these values are private to `MulticastRoutingManager`.

src/posix/platform/multicast_routing.hpp
simonlingoogle: I don't have the strong intention to use stdc++. But I also not sure why using stdc++ is not good for POSIX.

src/posix/platform/multicast_routing.cpp
simonlingoogle: It's arguable to some extent. One may argue that Backbone Router without Multicast routing is also useful, thus should keep running. 
But agree that VerifyOrDie should be better for disclosing issues. Will change.

src/posix/platform/multicast_routing.cpp
simonlingoogle: Is it necessary for POSIX?

src/posix/platform/platform-posix.h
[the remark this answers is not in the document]
simonlingoogle: I will fix to use WARN.

src/posix/platform/multicast_routing.cpp
simonlingoogle: Now that we use VerifyOrDie for `InitMulticastRouterSock`, Enable never fail.

src/posix/platform/multicast_routing.cpp
simonlingoogle: changed to use OT_MAX

src/posix/platform/backbone.cpp
```diff
@@ -55,8 +57,22 @@ void platformBackboneInit(otInstance *aInstance, const char *aInterfaceName)
...
     otLogInfoPlat("Backbone interface is configured to %s (%d)", gBackboneNetifName, gBackboneNetifIndex);
 
-exit:
-    return;
+    sMulticastRoutingManager.Init(aInstance);
+}
+
+void platformBackboneUpdateFdSet(fd_set &aReadFdSet, int &aMaxFd)
+{
+    sMulticastRoutingManager.UpdateFdSet(aReadFdSet, aMaxFd);
```
bukepo, line 65: As discussed, It's possible to simplify the design according to a good kernel version, and I think we should update the current implementation which is designed according to work with a buggy kernel.

simonlingoogle: @jwhui @librasungirl @bukepo 

According to our investigation, latest 5.x Linux kernel supports (*, G) multicasting, meantime 4.x Linux and macOS(BSD) does not support (*, G) multicasting, but only support (S, G) multicasting. 

What MLR feature needs for the PBBR is (*, G) multicast routing, thus we need to implement (*, G) multicast routing using (S, G) multicast routing in the POSIX platform code.

Even though it's possible to use a simpler implementation using (*, G) multicast routing provided by latest Linux kernel, I think **the current implementation is valuable because it's more portable**. It works on a wide range of Linux kernel versions and can be ported to support macOS.

simonlingoogle: > I think we should update the current implementation which is designed according to work with a buggy kernel.

I am not sure if it's fair to say that the current implementation is designed to work with a buggy kernel.

The current implementation uses a set of APIs that are provided by many POSIX platforms, including Linux, macOS, BSD. 
It does not use [MRT6_ADD_MFC_PROXY](
https://github.com/torvalds/linux/blob/3d5e28bff7ad55aea081c1af516cc1c94a5eca7d/include/uapi/linux/mroute6.h#L32) which is only provided by latest Linux to add a (\*, G) MFC entry, which is not well documented and not proven to work (I had a version using MRT6_ADD_MFC_PROXY but it didn't work as intended).

That being said, I am not against a future enhancement that utilizes MRT6_ADD_MFC_PROXY for Linux kernel versions that has working  (\*, G) MFC. But we should keep this implementation for portability. 

Thoughts?

librasungirl: It is a good suggestion to do investigation on (*,G) which has some obvious advantages.
I may suggest do it later as future optimization without blocking this PR review.

Thoughts?
@bukepo

bukepo (approved):

src/posix/platform/multicast_routing.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+#include "openthread-posix-config.h"
+
+#if OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <map>
+#include <set>
```
jwhui, line 40: Are these used?

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,569 @@
...
+    mMulticastRouterSock = -1;
+
+exit:
+    return;
+}
+
+void MulticastRoutingManager::ProcessMulticastRouterMessages(void)
+{
+    otError         error = OT_ERROR_NONE;
+    char            buf[128];
```
jwhui, line 219: Is it possible to use sizeof something here?

src/posix/platform/multicast_routing.hpp
```diff
@@ -0,0 +1,182 @@
...
+                                                            otBackboneRouterMulticastListenerEvent aEvent,
+                                                            const otIp6Address *                   aAddress);
+    void               HandleBackboneMulticastListenerEvent(otBackboneRouterMulticastListenerEvent aEvent,
+                                                            const Ip6::Address &                   aAddress);
+
+    MulticastForwardingCache mMulticastForwardingCacheTable[kMulitcastForwardingCacheTableSize];
+    uint64_t                 mLastExpireTime;
+    int                      mMulticastRouterSock;
+    otInstance *             mInstance;
+    void                     RemoveMulticastForwardingCache(MulticastForwardingCache &aMfc) const;
```
jwhui, line 174: nit: colocate with other methods above?

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,569 @@
...
+        break;
+    case kMifIndexBackbone:
+        string = "Backbone";
+        break;
+    }
+
+    return string;
+}
+
+void MulticastRoutingManager::DumpMulticastForwardingCache(void)
```
jwhui, line 434: Declare as `const`?

src/posix/platform/multicast_routing.cpp
```diff
@@ -0,0 +1,569 @@
...
+    VerifyOrExit(IsEnabled());
+
+    RemoveInboundMulticastForwardingCache(aAddress);
+    otLogResultPlat(error, "MulticastRoutingManager: %s: %s", __FUNCTION__, aAddress.ToString().AsCString());
+
+exit:
+    return;
+}
+
+bool MulticastRoutingManager::HasMulticastListener(const Ip6::Address &aAddress)
```
jwhui, line 126: Declare as `const`?

src/posix/platform/multicast_routing.cpp
simonlingoogle: Changed to use `sizeof(struct mrt6msg)`.

jwhui (approved):

PR #5804, opened 2020-11-12
[build] include project Android.mk in OpenThread Android.mk

zhanglongxia:


Android.mk
```diff
@@ -26,7 +26,15 @@
...
-LOCAL_PATH := $(call my-dir)
+# Save LOCAL_PATH of the current Android.mk file
+SAVED_LOCAL_PATH := $(call my-dir)
+
+ifneq ($(OPENTHREAD_PROJECT_ANDROID_MK),)
+include $(OPENTHREAD_PROJECT_ANDROID_MK)
+endif
+
+# Restore LOCAL_PATH variable
+LOCAL_PATH:= $(SAVED_LOCAL_PATH)
```
bukepo, line 37: Does appending the following line at the end of this file works for your case?

```bash
include $(call all-subdir-makefiles) 
```

zhanglongxia: The code `include $(call all-subdir-makefiles)` can only include the Android.mk files in the next level directory. It can't get the Android.mk files in the deeper directories.

bukepo: Thanks for the update. Let's avoid `SAVED_LOCAL_PATH` by dragging the change to the tail.

zhanglongxia: Updated

bukepo (approved):
LGTM:+1:

gjc13 (approved):

jwhui (approved):

PR #5807, opened 2020-11-12
[utils] adding 'DatasetUpdater'

abtink:
This commit adds a new module `DatasetUpdater` which provides
easy-to-use APIs for user to request an update to any component (a
subset of components) in Operational Dataset of the Thread network.
The change can be requested from any node in the network. The
implementation uses `MGMT_PENDING_SET` command and takes care of
preparing the Pending Dataset and updating timestamps and monitoring
for the Dataset to be updated and possibly retrying. It reports the
outcome (success or failure status) back to the user in a callback.

This commit also updates the `ChannelManager` module is to use the
`DatasetUpdater` to change the channel.

------

The `DatasetUpdater` API can bse used to quickly update PSKc.


bukepo (commented):
Is it possible to add a few tests to cover this new feature?

abtink:
> Is it possible to add a few tests to cover this new feature?

@bukepo the feature is currently covered through multiple tests we have for `ChannelManager` (note that `ChannelManager` is now using the `DatasetUpdater`).  In particular the channel manager test verifies behaviors and corner cases like requesting change from a single node, same request from multiple nodes (in parallel), conflicting requests (on same node and from different nodes). So the behavior is already well-covered.

For future, we could add NCP (and wpantund)  and/or CLI APIs and then can add more specific tests. I would consider it in future PRs and as a later enhancement.

src/core/meshcop/dataset.hpp
```diff
@@ -550,6 +550,19 @@ class Dataset
             mComponents.mIsSecurityPolicyPresent = true;
         }
 
+        /**
+         * This method sets the Security Policy in the Dataset.
+         *
+         * @param[in] aRotationTime  A value for Key Rotation (in units of hours).
+         * @param[in] aFlags         Security policy flags
```
jwhui, line 557: Inconsistent doxygen.

src/core/utils/dataset_updater.hpp
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @param[in]  aDataset                Dataset info containing fields to change.
+     * @param[in]  aCallback               A callback to indicate when Dataset update request finishes.
+     * @param[in]  aContext                An arbitrary context passed to callback.
+     * @param[in]  aRetryWaitInterval      The wait time after sending Pending dataset before retrying (interval in ms).
+     *
+     * @retval OT_ERROR_NONE           Dataset update started successfully (@p aCallback will be invoked on completion).
+     * @retval OT_ERROR_INVALID_STATE  Device is disabled (MLE is disabled).
+     * @retval OT_ERROR_INVALID_ARGS   The @p aDataset is not valid (contains Active or Pending Timestamp).
+     * @retval OT_ERROR_BUSY           Cannot start update, a previous one is ongoing
```
jwhui, line 101:
```suggestion
     * @retval OT_ERROR_BUSY           Cannot start update, a previous one is ongoing.
```

include/openthread/dataset_updater.h
```diff
@@ -0,0 +1,126 @@
...
+ * @param[in]  aInstance               A pointer to an OpenThread instance.
+ * @param[in]  aDataset                A pointer to the Dataset containing the fields to change.
+ * @param[in]  aCallback               A callback to indicate when Dataset update request finishes.
+ * @param[in]  aContext                An arbitrary context passed to callback.
+ * @param[in]  aRetryWaitInterval      The wait time after sending Pending dataset before retrying (interval in ms).
+ *
+ * @retval OT_ERROR_NONE           Dataset update started successfully (@p aCallback will be invoked on completion).
+ * @retval OT_ERROR_INVALID_STATE  Device is disabled (MLE is disabled).
+ * @retval OT_ERROR_INVALID_ARGS   The @p aDataset is not valid (contains Active or Pending Timestamp).
+ * @retval OT_ERROR_BUSY           Cannot start update, a previous one is ongoing
```
jwhui, line 88:
```suggestion
 * @retval OT_ERROR_BUSY           Cannot start update, a previous one is ongoing.
```

jwhui (approved):

Issue #5801, opened 2020-11-12
Validate network name as UTF-8 string


bukepo:
**Is your feature request related to a problem? Please describe.** A clear and concise description of what the problem is.
Missing validation of Thread network name as a UTF-8 string.

**Describe the solution you'd like** A clear and concise description of what you want to happen.
Validate it.

**Describe alternatives you've considered** A clear and concise description of any alternative solutions or features you've considered.
None

**Additional context** Add any other context or screenshots about the feature request here.

simonlingoogle:
Why is it necessary?

bukepo:
> Why is it necessary?

Thread Network Name is defined as UTF-8 string.

abtink:
We added and used a `spinel_validate_utf8()` implementation (since spinel the gatekeeper to the OT API on NCP build it checks all inputs parsed as `SPIENL_DATATYPE_UTF8`).

I think we can re-use the same implementation within OT core (and perform common checks in OT).

PR #5810, opened 2020-11-13
[string] add UTF-8 validation

Moandor-y:
Fixes <https://github.com/openthread/openthread/issues/5801>.


src/core/common/utf8.hpp
```diff
@@ -0,0 +1,54 @@
...
+#ifndef UTF8_H_
+#define UTF8_H_
+
+#include <stdint.h>
+
+namespace ot {
+
+/**
+ * This function validates whether a given byte sequence (string) follows UTF-8 encoding.
+ *
```
Irving-cl, line 43: Illustrate that `0` marks the end of the string since length is not passed as parameter?

src/core/common/utf8.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#include "common/code_utils.hpp"
+
+namespace ot {
+
+bool ValidateUtf8(const uint8_t *aString)
+{
+    bool    ret = true;
+    uint8_t byte;
+    uint8_t continuation_bytes = 0;
```
Irving-cl, line 39: Use camel style to keep consistent?

src/core/common/utf8.hpp
Moandor-y: Updated the argument documentation.

src/core/common/utf8.cpp
Moandor-y: Done.

src/core/common/utf8.hpp
Moandor-y: After a second thought, I think we must add length as a parameter here.

Because we are accepting user inputs here, there will be the possibility of buffer overflow attack if we don't check bounds.

simonlingoogle:
What about network name setting via other means:

- dataset network xxx -> dataset active/pending commit
- MGMT_ACTIVE.SET
- MGMT_PENDING.SET

src/core/common/utf8.hpp
simonlingoogle: > Because we are accepting user inputs here, there will be the possibility of buffer overflow attack if we don't check bounds.

If we are talking about CLI commands, I think it's the responsibility of CLI implementation to make sure that command arguments are correctly validated.

src/core/common/utf8.hpp
```diff
@@ -0,0 +1,54 @@
...
+ * @param[in]  aString  A null-terminated byte sequence.
+ *
+ * @retval TRUE   The sequence is a valid UTF-8 string.
+ * @retval FALSE  The sequence is not a valid UTF-8 string.
+ *
+ */
+bool ValidateUtf8(const uint8_t *aString);
```
wgtdkp, line 50: `const char *` fits better for a null-terminated string. `const uint8_t *` gives me the sense that there will be no special character/byte in the string/byte-sequence.

src/core/common/utf8.hpp
```diff
@@ -0,0 +1,54 @@
...
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file contains UTF-8 related functions.
+ */
+
+#ifndef UTF8_H_
```
bukepo, line 34:
```suggestion
#ifndef OT_CORE_UTILS_UTF8_H_
```

src/core/api/thread_api.cpp
```diff
@@ -39,6 +39,7 @@
 #include "common/instance.hpp"
 #include "common/locator-getters.hpp"
 #include "common/settings.hpp"
+#include "common/utf8.hpp"
```
bukepo, line 42: move to core/utils?

abtink: How about adding the `ValidateUtf8` to the existing `string.hpp` (It contains a bunch other string related helper methods/functions).

src/core/common/utf8.cpp
```diff
@@ -0,0 +1,85 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "utf8.hpp"
+
+#include "common/code_utils.hpp"
+
+namespace ot {
+
+bool ValidateUtf8(const uint8_t *aString)
```
abtink, line 35: How about defining it as `const char *` instead?

(in spinel we used `const uint8_t *` as the type in  `spinel_validate_utf8()` since for the place where it is being used/called, during parsing of UTF8 formati input, the `data_in` is still being treated it as `const uint8_t *`).

abtink:
Thanks @Moandor-y for adding this.

There are bunch of other OT APIs that use string input parameters (in addition to  "network name"). 
Should we check if they need to follow UTF8 format and if they do, add code/check to verify this?

- `otThreadSetDomainName(otInstance *aInstance, const char *aDomainName)`  (I think this would need to be UTF8)
- `otCommissionerSetProvisioningUrl(otInstance *aInstance, const char *aProvisioningUrl);` 
- We have provisioning URL, vendor name, model, etc in `joiner.h` and `otJoinerStart()`.
- We have `aUriPath` in `coap.h` (not sure if it needs to be UTF8 or not).
- We have `aPskd` in `commissioner.h` (again not sure if it needs to be UTF8 or not - I expect  it to be though, would be good to double-check)
- Would be good to check if commissioner PSK (`aPskc`) needs to be UTF8 or not.
- And `otError otDatasetGeneratePskc(const char *aPassPhrase,...)` we have `aPassPhrase`.

bukepo:
Should a node reject a operational dataset with invalid network name?

simonlingoogle:
> Should a node reject a operational dataset with invalid network name?

I think a node should accept an operational dataset with invalid network name if the Leader is using the invalid network name already. **Connectivity matters.** 

**I think it's the responsibility of the Leader to prohibit using invalid network name in the first place.**

Thus the Leader should:

- Reject network configuration through public OT API if network name is invalid
- Reject operational dataset set (i.e. MGMT_ACTIVE_SET and MGMT_PENDING_SET) if network name is invalid (not sure if it's necessary because we can also prohibit this way in the public OT API)

abtink:
> Reject operational dataset set (i.e. MGMT_ACTIVE_SET and MGMT_PENDING_SET) if network name is invalid (not sure if it's necessary because we can also prohibit this way in the public OT API)

Fully agree with @simonlingoogle comment above.

On  `MGMT_ACTIVE_SET` and `MGMT_PENDING_SET`, we can consider adding a check on leader to protect against devices that may not be doing the check themselves (e.g. devices running older version of OT with no check at OT API level or possibly different stack). Thoughts?

jwhui:
> On `MGMT_ACTIVE_SET` and `MGMT_PENDING_SET`, we can consider adding a check on leader to protect against devices that may not be doing the check themselves (e.g. devices running older version of OT with no check at OT API level or possibly different stack). Thoughts?

`MGMT_ACTIVE_SET` and `MGMT_PENDING_SET` may also come from a device that has migrated from a different partition that allowed a non-UTF-8 network name. Thoughts?

src/core/common/utf8.hpp
Moandor-y: Done.

src/core/common/utf8.hpp
Moandor-y: Moved to `string.hpp`.

src/core/api/thread_api.cpp
Moandor-y: Moved to `string.hpp`.

src/core/common/utf8.cpp
Moandor-y: Done.

Moandor-y:
> Thanks @Moandor-y for adding this.
> 
> There are bunch of other OT APIs that use string input parameters (in addition to "network name").
> Should we check if they need to follow UTF8 format and if they do, add code/check to verify this?
> 
> * `otThreadSetDomainName(otInstance *aInstance, const char *aDomainName)`  (I think this would need to be UTF8)
> * `otCommissionerSetProvisioningUrl(otInstance *aInstance, const char *aProvisioningUrl);`
> * We have provisioning URL, vendor name, model, etc in `joiner.h` and `otJoinerStart()`.
> * We have `aUriPath` in `coap.h` (not sure if it needs to be UTF8 or not).
> * We have `aPskd` in `commissioner.h` (again not sure if it needs to be UTF8 or not - I expect  it to be though, would be good to double-check)
> * Would be good to check if commissioner PSK (`aPskc`) needs to be UTF8 or not.
> * And `otError otDatasetGeneratePskc(const char *aPassPhrase,...)` we have `aPassPhrase`.
> 
> Edit:
> 
> * Another place that we may want to check the network name string to be valid is the scan result

Added checks to verify provisioning URL, pass phrase, vendor name, vendor sw version, domain name, which are explicitly stated to be UTF-8 in the spec.

For vendor model, coap URI, and pskd, I don't see encoding requirements in the spec.

simonlingoogle:
> `MGMT_ACTIVE_SET` and `MGMT_PENDING_SET` may also come from a device that has migrated from a different partition that allowed a non-UTF-8 network name. Thoughts?

Or come from device that is running an old OpenThread version before this PR. This would cause that device to retry sending for ever? 

I think it's a good idea to prohibit invalid names only in the public API implementation, so that the OT core code does not need to worry about this issue at all. 

Thoughts?

abtink:
> MGMT_ACTIVE_SET and MGMT_PENDING_SET may also come from a device that has migrated from a different partition that allowed a non-UTF-8 network name. Thoughts?
> Or come from device that is running an old OpenThread version before this PR. This would cause that device to retry sending for ever?
> 
> I think it's a good idea to prohibit invalid names only in the public API implementation, so that the OT core code does not need to worry about this issue at all.
> 
> Thoughts?

Sounds reasonable to me.

jwhui:
> > `MGMT_ACTIVE_SET` and `MGMT_PENDING_SET` may also come from a device that has migrated from a different partition that allowed a non-UTF-8 network name. Thoughts?
> 
> Or come from device that is running an old OpenThread version before this PR. This would cause that device to retry sending for ever?
> 
> I think it's a good idea to prohibit invalid names only in the public API implementation, so that the OT core code does not need to worry about this issue at all.
> 
> Thoughts?

I agree.

abtink:
Thanks @Moandor-y for updating the code and PR.

Now that we have the UTF8 check within OT core, I'd like us to consider/discuss the "network name" in scan result (MLE discovery scan or MAC active scan).

The situation to consider is that we can have an incorrectly configured (or possibly rouge) node in vicinity of the network with bad network name (not following UTF8) and when we scan we can find it in the result which is then passed to user in a callback.

If the user treats the network name in scan result as UTF8 (assumes it would be), it can potentially cause issues (e.g., if it is encoded in dbus message, we can get dbus assert/crash). 

I see two options:
1. We can check the network name of scan result within OT core and not report the ones that don't follow UTF8 (drop them).
1. We can pass all results in the callback (no check in OT) and leave it to users to check it to be UTF8 (if they want to treat it as such).

I personally prefer/recommend first option. 

Thoughts?

bukepo:
Thread network name TLV is defined as a UTF-8 string in spec. I think we should treat an operational dataset containing an invalid network name TLV invalid. This conforms to the spec and will prevent a common Wi-Fi SSID issue, i.e. corrupted SSID name because of different encoding. I think that's the motivation of explicitly defining the TLV as UTF-8 encoded. If we allow non-UTF-8 encoding, OpenThread users would have to take care of non-UTF-8 cases. And if allowing invalid network name in consideration of connectivity, is an Extended PAN ID not 8 bytes in length also acceptable?

Moandor-y:
I think we don't need to worry about existing bad network names. Because users should follow the spec and use valid network names, even if our implementation didn't enforce it. The incompatibility caused by invalid network names is none of our business.

And for scan results, I prefer to drop them.

src/core/common/string.hpp
```diff
@@ -258,6 +258,17 @@ template <uint16_t SIZE> class String : private StringBase
...
+/**
+ * This function validates whether a given byte sequence (string) follows UTF-8 encoding.
+ *
+ * @param[in]  aString  A null-terminated byte sequence.
+ *
+ * @retval TRUE   The sequence is a valid UTF-8 string.
+ * @retval FALSE  The sequence is not a valid UTF-8 string.
+ *
+ */
+bool ValidateUtf8(const char *aString);
```
simonlingoogle, line 270: nit: 
```suggestion
bool IsValidUtf8String(const char *aString);
```

since it returns `bool`.

simonlingoogle (commented):
> I think we don't need to worry about existing bad network names. Because users should follow the spec and use valid network names, even if our implementation didn't enforce it. The incompatibility caused by invalid network names is none of our business.
> 
> And for scan results, I prefer to drop them.

No strong opinion. But it seems fine to "pass all results in the callback (no check in OT) and leave it to users to check it to be UTF8 (if they want to treat it as such)". Client-side UTF-8 decoding libraries usually provides decoding options to ignore encoding errors and only return the correct UTF-8 characters. Moreover, this should "never" happen after we merge this PR.

tests/scripts/expect/cli-networkname.exp
```diff
@@ -0,0 +1,80 @@
...
+send "networkname スレッド\n"
+expect "Done"
+send "thread start\n"
+expect "Done"
+wait_for "state" "leader"
+expect "Done"
+
+switch_node 2
+send "scan\n"
+expect "スレッド"
```
simonlingoogle, line 77: Add at least one case using invalid UTF8 string?

tests/scripts/expect/cli-networkname.exp
```diff
@@ -0,0 +1,80 @@
...
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
+source "tests/scripts/expect/_common.exp"
+
+spawn_node 1
+spawn_node 2
+
+switch_node 1
+send "networkname Thread\\ Network\n"
```
simonlingoogle, line 36: Add test case of `dataset networkname` ?

src/cli/cli_dataset.cpp
```diff
@@ -485,6 +486,7 @@ otError Dataset::ProcessNetworkName(uint8_t aArgsLength, char *aArgs[])
         size_t length;
 
         VerifyOrExit((length = strlen(aArgs[0])) <= OT_NETWORK_NAME_MAX_SIZE, error = OT_ERROR_INVALID_ARGS);
+        VerifyOrExit(ValidateUtf8(aArgs[0]), error = OT_ERROR_INVALID_ARGS);
```
simonlingoogle, line 489: Does this cover the case where dataset is set using hex format?

https://github.com/openthread/openthread/blob/master/src/cli/README_DATASET.md#set

tests/scripts/expect/cli-networkname.exp
Moandor-y: I didn't find a way to send binary data from `expect`.

When `send '\xbf'`, it considers `bf` as a code point, then encodes it using UTF-8.

simonlingoogle: try "\udfff" ?

Any character among '\ud800' ~ '\udfff' seems to be invalid UTF8.

src/cli/cli_dataset.cpp
Moandor-y: No. The current implementation of hex set doesn't do any check. It succeeds even if the data is just some random binary data. I think it's good to add the parsing and verification, but that also means some runtime overhead.

src/core/common/string.hpp
Moandor-y: Done.

tests/scripts/expect/cli-networkname.exp
Moandor-y: Added to `cli-dataset.exp`.

tests/scripts/expect/cli-networkname.exp
Moandor-y: It translates `\udfff` to `ed bf bf`, looks still valid UTF-8.

src/core/common/utf8.hpp
Moandor-y: Eventually added the length parameter as the network name TLV doesn't have the string terminator.

Moandor-y:
Added the check to network name TLV, so MGMT SET commands with invalid network names will be considered invalid.

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -511,7 +511,7 @@ class NetworkNameTlv : public Tlv, public TlvInfo<Tlv::kNetworkName>
      * @retval FALSE  If the TLV does not appear to be well-formed.
      *
      */
-    bool IsValid(void) const { return true; }
+    bool IsValid(void) const;
```
bukepo, line 514: Is this called everywhere the NetworkNameTlv is constructed?

simonlingoogle (approved):
LGTM :+1:

src/core/meshcop/meshcop_tlvs.hpp
Moandor-y: I added the check to scan results. This is already being called in most places. There are some cases where this isn't needed, e.g., initializing a `NetworkNameTlv` directly from `Mac::GetNetworkName`.

src/cli/cli_dataset.cpp
Moandor-y: Added checks for `dataset set`.

abtink (approved):
Looks great. Thank you. 👍

src/core/meshcop/meshcop_tlvs.hpp
bukepo: Thank you!

bukepo (approved):
LGTM:+1:

wgtdkp (approved):
LGTM 👍

Irving-cl (approved):
LGTM 👍

jwhui (approved):

PR #5812, opened 2020-11-14
[github-actions] upload core dump and executable (Don't merge)

Irving-cl:
This is a test PR to verify that executables and dump files could be uploaded successfully. Don't merge.

### DIFF
--- a/.github/workflows/simulation-1.2.yml
+++ b/.github/workflows/simulation-1.2.yml
@@ -70,13 +70,27 @@ jobs:
         ./script/test build
     - name: Run
       run: |
+        ulimit -c unlimited
+        ./script/test prepare_coredump_upload
         ./script/test unit
         ./script/test cert_suite tests/scripts/thread-cert/v1_2_*
+    - name: Copy Shared Libraries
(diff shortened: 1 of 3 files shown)


Irving-cl:
I have verified that we can use gdb to debug core dump files generated in github actions. We need to upload:
* executables
* core dump files
* `.so` libs (so that we can gdb core files generated on another machine)

Steps:
* unzip the packet, `cd` to the path
* `gdb build/cmake/openthread-simulation-1.2/examples/apps/cli/ot-cli-ftd ./ot-core-dump/corefile-XXX`
* Set so lib(Very important, use **absolute path**), in gdb: `set solib-absolute-prefix /ABSOLUTE/PATH/TO/lib/`, then `set solib-search-path /ABSOLUTE/PATH/TO/lib/`.
* in gdb: `backtrace` or `bt`.

Screen Shot 2020-11-24 at 11 37 19 AM

I'm now testing uploading `*.so` libs. I did it manually before using tmate session.

Irving-cl:
#5681 merged.

PR #5818, opened 2020-11-16
[efr32] radio: fill in rx info for received ACKs

dismirlian:
Currently, `processNextRxPacket` fills in the received frame's `mRxInfo` only for non-ack packets. When a SED polls its parent, the ACK frame is passed up to the OT core in `efr32RadioProcess`, but the rx info is incorrect (it is actually the rx info for the last received non-ack packet).

OT core uses the RSSI and LQI internally, and for applications that don't transmit/receive data frequenty, the rx info is stale. This PR changes this behavior, so the radio driver populates the rx info also for ACK packets.

### DIFF
--- a/examples/platforms/efr32/src/radio.c
+++ b/examples/platforms/efr32/src/radio.c
@@ -815,10 +815,23 @@ static void processNextRxPacket(otInstance *aInstance)
 
     sReceiveFrame.mLength = length;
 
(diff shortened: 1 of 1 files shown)


schinchani (commented):
@dismirlian  / @romacdon  - Changes look good to me.

schinchani (approved):

jwhui (approved):

PR #5833, opened 2020-11-19
[low-power] fix CSL scheduler issues to avoid packets stuck in queue for a long time

brianljt:
This PR
1. Fixes the issue in CSL scheduler dealing with multiple SSEDs.
2. Fixes the issue when CSL tx attempts reach max, CSL transmission to this child cannot resume.
3. Reset CSL sync status after child mode change to RxOnWhenIdle.
4. Adds MLE version, CSL synchronization status and queued message count to child table.


src/cli/cli.cpp
```diff
@@ -806,6 +806,9 @@ otError Interpreter::ProcessChild(uint8_t aArgsLength, char *aArgs[])
                 OutputFormat("|%1d", childInfo.mRxOnWhenIdle);
                 OutputFormat("|%1d", childInfo.mFullThreadDevice);
                 OutputFormat("|%1d", childInfo.mFullNetworkData);
+                OutputFormat("| %1d ", childInfo.mVersion);
+                OutputFormat("| %1d ", childInfo.mIsCslSynced);
+                OutputFormat("| %5d ", childInfo.mQueuedMessageCnt);
```
simonlingoogle, line 811:
```suggestion
                OutputFormat("|%1d ", childInfo.mVersion);
                OutputFormat("|%1d ", childInfo.mIsCslSynced);
                OutputFormat("|%5d ", childInfo.mQueuedMessageCnt);
```

simonlingoogle (approved):
LGTM 👍

src/cli/cli.cpp
brianljt: The whitespace is used to fill up the unused width in the table like RLOC16 above.

Irving-cl (approved):
LGTM :+1:

edmont (approved):

src/core/thread/csl_tx_scheduler.cpp
```diff
@@ -97,7 +98,7 @@ void CslTxScheduler::Clear(void)
 {
     for (Child &child : Get<ChildTable>().Iterate(Child::kInStateAnyExceptInvalid))
     {
-        child.SetCslTxAttempts(0);
+        child.ResetIndirectTxAttempts();
```
Irving-cl, line 101: The reason we separate TX attempts for CSL and indirect transmission is that, when CSL Tx fails for max attempts, we hope the message could still be successfully delivered by indirect transmission. If we use the same counter now, will the frame be discarded and cannot fetched by data poll any more after CSL Tx fails for max attempts?

brianljt: Yes. I think for a SSED, it is reasonable for it to use a very big data poll period(so as the save power) or even turn off data poll until it decides no packets are received during CSL timeout period and consider CSL out of sync. So I think it would be more practical to count CSL tx attempts as indirect sending attempts and just drop the packet if all the attempts fails.

For default setting, this attempt is 4. But in some products, this value could be something bigger like 16 which makes the waiting time way too long.

Irving-cl: Ok. Make sense to me.

Irving-cl (approved):
LGTM

src/core/thread/csl_tx_scheduler.cpp
jwhui: Thread 1.2 states the following:
```
Retries are handled using the mechanism defined for CSL in [IEEE 802.15.4-2015]. After
the maximum number of failed retransmissions, defined by macMaxFrameRetries, the Child
MUST be considered CSL de-synchronized.
```
and:
```
When a SSED Child is considered de-synchronized by the Parent, all packets to be
forwarded to the Child SHALL be sent by indirect transmission.
```
If we hit the max retries, then we will have to deal with incoming messages that may be queued for a long time waiting for a data poll.

include/openthread/thread_ftd.h
```diff
@@ -67,10 +67,13 @@ typedef struct
     int8_t       mLastRssi;             ///< Last observed RSSI
     uint16_t     mFrameErrorRate;       ///< Frame error rate (0xffff->100%). Requires error tracking feature.
     uint16_t     mMessageErrorRate;     ///< (IPv6) msg error rate (0xffff->100%). Requires error tracking feature.
+    uint16_t     mQueuedMessageCnt;     ///< Number of queued indirect messages for the child.
```
jwhui, line 70: 1) Suggest putting "indirect" in the field name.
```suggestion
    uint16_t     mIndirectQueueCnt;     ///< Number of queued indirect messages for the child.
```
2) IEEE 802.15.4-2015 does not count queued CSL transmissions as being in the indirect queue.

src/core/thread/csl_tx_scheduler.cpp
brianljt: Thank you Jonathan for pointing this out. I think the rational behind the spec is, if CSL transmission reaches `macMaxFrameRetries`(set to 4 by default), it is big enough to consider this SSED out of sync. Thus there are no point of dropping this frame and trying sending another packet to this child. In this case, the parent can do nothing but just queue the packet and wait for data poll from this child.

This change violates the spec, I will change this part back to the original implementation.

include/openthread/thread_ftd.h
brianljt: What I intended to show in this field is how many packets are queued. This could be packets not sent due to CSL sample not reached or data poll not received yet.

src/core/thread/csl_tx_scheduler.cpp
brianljt: I have changed back to use separate counter for CSL tx attempts and indirect sending attempts. Packet will only be dropped when indirect transmission counter reaches max.

Also fix one issue that CSL tx attempts counter will never be reset once it reaches max(only got reset with a successful CSL transmission). This will stuck CSL transmission even this child regain sync later with this router. PR description also updated(point 2). Please help take a look. Thank you!

include/openthread/thread_ftd.h
jwhui: Can we update the comment to match what was intended?

brianljt: Updated

abtink (approved):
Looks great. Thanks. 👍

jwhui (approved):

PR #5835, opened 2020-11-20
[thci] add low power thci

Irving-cl:
This PR adds THCI for low power features. I refered to the current 1_2 THCI and update its implementation according to our latest code in OpenThread. I did some local testing.

### DIFF
--- a/tools/harness-thci/OpenThread.py
+++ b/tools/harness-thci/OpenThread.py
@@ -2992,22 +2992,23 @@ class OpenThreadTHCI(object):
     def sendMACcmd(self, enh=False):
         self.log('call sendMACcmd')
         cmd = 'mac send datarequest'
+        print(cmd)
         return self.__executeCommand(cmd)[-1] == 'Done'
 
     @API
     def sendMACdata(self, enh=False):
         self.log('call sendMACdata')
         cmd = 'mac send emptydata'
+        print(cmd)
         return self.__executeCommand(cmd)[-1] == 'Done'
 
     @API
     def setCSLsuspension(self, suspend):
         self.log('call setCSLsuspension')
         if suspend:
-            cmd = 'pollperiod 1000000'
+            self.__setPollPeriod(1000000)
         else:
-            cmd = 'pollperiod 5000'
-        return self.__executeCommand(cmd)[-1] == 'Done'
+            self.__setPollPeriod(self.__sedPollPeriod)
 
     @staticmethod
     def __lstrip0x(s):
--- a/tools/harness-thci/OpenThread.py
+++ b/tools/harness-thci/OpenThread.py
@@ -2843,14 +2843,14 @@ class OpenThreadTHCI(object):
     # Low power THCI
     @API
     def setCSLtout(self, tout=30):
-        print('%s call setCSLtout' % self)
+        self.log('call setCSLtout')
         cmd = 'csl timeout %u' % tout
         print(cmd)
         return self.__executeCommand(cmd)[-1] == 'Done'
 
     @API
     def setCSLchannel(self, ch=11):
-        print('%s call setCSLchannel' % self)
+        self.log('call setCSLchannel')
         cmd = 'csl channel %u' % ch
         print(cmd)
         return self.__executeCommand(cmd)[-1] == 'Done'
@@ -2865,7 +2865,7 @@ class OpenThreadTHCI(object):
         period is converted from unit ms to ten symbols (160us per 10 symbols).
 
         """
-        print('%s call setCSLperiod' % self)
+        self.log('call setCSLperiod')
         cmd = 'csl period %u' % (period * 6.25)
         print(cmd)
         return self.__executeCommand(cmd)[-1] == 'Done'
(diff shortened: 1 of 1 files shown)


tools/harness-thci/OpenThread.py
```diff
@@ -2840,6 +2840,178 @@ def ValidateDeviceFirmware(self):
...
+        metricsFlagMap = {
+            0x40: 'p',
+            0x09: 'q',
+            0x0a: 'm',
+            0x0b: 'r',
+        }
+        if metrics in metricsFlagMap:
+            return metricsFlagMap[metrics]
+        else:
+            return '?'
```
simonlingoogle, line 2900:
```suggestion
        return metricsFlagMap.get(metrics, '?')
```

tools/harness-thci/OpenThread.py
```diff
@@ -2840,6 +2840,178 @@ def ValidateDeviceFirmware(self):
...
+        print('%s call sendUdp' % self)
+        assert payload is not None, 'payload should not be none'
+        cmd1 = 'udp open'
+        print(cmd1)
+        cmd2 = 'udp send %s %d %s' % (destination, port, payload)
+        print(cmd2)
+        return (self.__executeCommand(cmd1)[-1] == 'Done' and self.__executeCommand(cmd2)[-1] == 'Done')
+
+    @API
+    def send_udp(self, interface, destination, port, payload='12ABcd'):
```
simonlingoogle, line 2983: do we need both send_udp and sendUdp ?

Irving-cl: I found them from the current 1_2 THCI. The difference is payload format: one uses string while the other uses hex bytes.

tools/harness-thci/OpenThread.py
```diff
@@ -2840,6 +2840,175 @@ def ValidateDeviceFirmware(self):
...
+        cmd = 'mac send emptydata'
+        return self.__executeCommand(cmd)[-1] == 'Done'
+
+    @API
+    def setCSLsuspension(self, suspend):
+        self.log('call setCSLsuspension')
+        if suspend:
+            cmd = 'pollperiod 1000000'
+        else:
+            cmd = 'pollperiod 5000'
```
simonlingoogle, line 3009: why 5000?

tools/harness-thci/OpenThread.py
```diff
@@ -2840,6 +2840,175 @@ def ValidateDeviceFirmware(self):
...
+        return (self.__executeCommand(cmd1)[-1] == 'Done' and self.__executeCommand(cmd2)[-1] == 'Done')
+
+    @API
+    def sendMACcmd(self, enh=False):
+        self.log('call sendMACcmd')
+        cmd = 'mac send datarequest'
+        return self.__executeCommand(cmd)[-1] == 'Done'
+
+    @API
+    def sendMACdata(self, enh=False):
```
simonlingoogle, line 2998: why `enh` not used?

tools/harness-thci/OpenThread.py
```diff
@@ -2840,6 +2840,175 @@ def ValidateDeviceFirmware(self):
...
+        self.log('call send_udp')
+        assert payload is not None, 'payload should not be none'
+        cmd1 = 'udp open'
+        print(cmd1)
+        cmd2 = 'udp send %s %s -x %s' % (destination, port, payload)
+        print(cmd2)
+        return (self.__executeCommand(cmd1)[-1] == 'Done' and self.__executeCommand(cmd2)[-1] == 'Done')
+
+    @API
+    def sendMACcmd(self, enh=False):
```
simonlingoogle, line 2992: why `enh` not used?

simonlingoogle (approved):
LGTM :+1: 

with nits

tools/harness-thci/OpenThread.py
Irving-cl: Currently in test plan, when empty MAC data is used, it's used to send a 2015 frame. And currently our implementation doesn't support specify the frame version. I'd like to enhance it when we actually need it.

tools/harness-thci/OpenThread.py
Irving-cl: This value may be adjusted. There's no specific reason. So long as the value won't break the test cases.

simonlingoogle: The default pollperiod of a OT SED in certification is 3000. Maybe use 3000 instead?

librasungirl: agree that it might be better to restore the default poll period `self.__sedPollPeriod`

tools/harness-thci/OpenThread.py
```diff
@@ -2992,21 +2992,23 @@ def send_udp(self, interface, destination, port, payload='12ABcd'):
...
     @API
     def setCSLsuspension(self, suspend):
         self.log('call setCSLsuspension')
         if suspend:
-            cmd = 'pollperiod 1000000'
+            self.__setPollPeriod(1000000)
         else:
-            cmd = 'pollperiod 5000'
+            self.__setPollPeriod(self.__sedPollPeriod)
         return self.__executeCommand(cmd)[-1] == 'Done'
```
librasungirl, line 3012: this line should be removed ?

Irving-cl: Updated, thanks.

librasungirl (approved):

jwhui (approved):

Issue #5767, opened 2020-11-03
[nrf52840] first packet after reboot has wrong security


brianljt:
**Describe the bug**
Enable Thread 1.2 leader using RCP platform. After leader rebooted(ctrl-c and restart), first packet sent out will have wrong security. Second and the following packets are correct. Issue is gone when using builtin mbedtls.

**To Reproduce**
1. Git commit id
    tip of master. I'm using f886f34d0062a6e89639324c9d9f8b7bb0d67430
2. IEEE 802.15.4 hardware platform
    nrf52840 DK
3. Build steps
    make -f examples/Makefile-nrf52840 THREAD_VERSION=1.2
4. Network topology
    4.1 Start nrf52840DK as RCP leader: ./output/posix/x86_64-unknown-linux-gnu/bin/ot-cli -v "spinel+hdlc+uart:///dev/ttyACM1"
    4.2 Start another nrf52840DK as child. Child attach to the leader. Ping can go through from leader to child
    4.3 Kill posix app(ctrl - c) and restart. Start Thread.
    4.4 Ping from this leader to child.

**Expected behavior**
Expect ping response from child.

**Console/log output**
```bash
# initial setup, first ping can get ping response
> factoryreset
./output/posix/x86_64-unknown-linux-gnu/bin/ot-cli[475264]: Running OPENTHREAD/20191113-01381-g0b0d6eef1-dirty; POSIX; Nov  3 2020 14:52:31
./output/posix/x86_64-unknown-linux-gnu/bin/ot-cli[475264]: Thread version: 3
./output/posix/x86_64-unknown-linux-gnu/bin/ot-cli[475264]: RCP version: OPENTHREAD/20191113-01381-g0b0d6eef1-dirty; NRF52840; Nov  3 2020 14:53:25
> panid 0xbeef
Done
> ifconfig up
Done
> thread start
Done
> ping fe80:0:0:0:9cc5:a92c:227e:cd02
Done
> 16 bytes from fe80:0:0:0:9cc5:a92c:227e:cd02: icmp_seq=1 hlim=64 time=26ms
>
# Kill posix app and restart
./output/posix/x86_64-unknown-linux-gnu/bin/ot-cli -v "spinel+hdlc+uart:///dev/ttyACM1"
./output/posix/x86_64-unknown-linux-gnu/bin/ot-cli[488654]: Running OPENTHREAD/20191113-01381-g0b0d6eef1-dirty; POSIX; Nov  3 2020 14:52:31
[...]

PR #5837, opened 2020-11-20
[nrf52840] Improved nrf security - fix for first packet coming unencrypted

MarekPorwisz:
addresses https://github.com/openthread/openthread/issues/5767

Had to switch to latest nrf_security as the old one does not support ROM AES tables. But since the new one adds support for oberon backend, I made use of it as it is a lot faster than mbedtls.


MarekPorwisz:
@piotrkoziar please also have a look at it. I might have missed something that I was not aware of.

LuDuda (approved):
LGTM :+1:

jwhui:
@MarekPorwisz , the arm-gcc-4 build test is failing, can you take a look?

MarekPorwisz:
Squashed all commits, as the prior commits would have some issues working without the last one (in a specific conditions only). All tests passed before the squash.

jwhui:
@piotrkoziar , please review when you have a chance. Thanks!

third_party/NordicSemiconductor/libraries/nrf_security/config/nrf-config.h
```diff
@@ -653,15 +743,15 @@
  *
  * Comment macros to disable the curve and functions for it
  */
-#define MBEDTLS_ECP_DP_SECP192R1_ENABLED
-#define MBEDTLS_ECP_DP_SECP224R1_ENABLED
+/* #undef MBEDTLS_ECP_DP_SECP192R1_ENABLED */
```
piotrkoziar, line 746: Please take a look at https://github.com/openthread/openthread/pull/5265. Do you know if the additional curves are present in DTLS client hello? If still, we should apply changes as in https://github.com/openthread/openthread/pull/5265.

piotrkoziar (approved):
LGTM

third_party/NordicSemiconductor/libraries/nrf_security/config/nrf-config.h
MarekPorwisz: Discussed with @piotrkoziar and this is something that can be addressed at a later time. If this happens to be still applicable it is only a small inconsistency in the libs provided and configuration file.

jwhui (approved):
Thanks! 👍

PR #5841, opened 2020-11-22
[cli] add helper 'Output()' methods which prepend indentation space

abtink:
This commit simplifies the CLI module by adding new flavors of helper
methods `OutputLine/Format(aIndentSize, ...)` which prepend a given
number of indentation space characters before the format string.

It also harmonizes the other `Output{Item}()` methods which output an
item with a given indentation to use `aIndentSize` as the parameter
name and as the first parameter of the method.

It also simplifies the implementation of `OutputSpaces()`.

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -1057,24 +1057,24 @@ otError Interpreter::ProcessCoexMetrics(uint8_t aArgsLength, char *aArgs[])
         OutputLine("Stopped: %s", metrics.mStopped ? "true" : "false");
         OutputLine("Grant Glitch: %u", metrics.mNumGrantGlitch);
         OutputLine("Transmit metrics");
-        OutputLine("    Request: %u", metrics.mNumTxRequest);
-        OutputLine("    Grant Immediate: %u", metrics.mNumTxGrantImmediate);
-        OutputLine("    Grant Wait: %u", metrics.mNumTxGrantWait);
-        OutputLine("    Grant Wait Activated: %u", metrics.mNumTxGrantWaitActivated);
-        OutputLine("    Grant Wait Timeout: %u", metrics.mNumTxGrantWaitTimeout);
-        OutputLine("    Grant Deactivated During Request: %u", metrics.mNumTxGrantDeactivatedDuringRequest);
-        OutputLine("    Delayed Grant: %u", metrics.mNumTxDelayedGrant);
-        OutputLine("    Average Request To Grant Time: %u", metrics.mAvgTxRequestToGrantTime);
+        OutputLine(kIndentSize, "Request: %u", metrics.mNumTxRequest);
+        OutputLine(kIndentSize, "Grant Immediate: %u", metrics.mNumTxGrantImmediate);
+        OutputLine(kIndentSize, "Grant Wait: %u", metrics.mNumTxGrantWait);
+        OutputLine(kIndentSize, "Grant Wait Activated: %u", metrics.mNumTxGrantWaitActivated);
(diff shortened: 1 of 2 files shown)


Irving-cl (approved):
LGTM 👍

src/cli/cli.hpp
```diff
@@ -494,14 +524,14 @@ class Interpreter
...
-    void        OutputSpaces(uint16_t aCount);
-    void        OutputMode(const otLinkModeConfig &aMode, uint16_t aColumn);
-    void        OutputConnectivity(const otNetworkDiagConnectivity &aConnectivity, uint16_t aColumn);
-    void        OutputRoute(const otNetworkDiagRoute &aRoute, uint16_t aColumn);
-    void        OutputRouteData(const otNetworkDiagRouteData &aRouteData, uint16_t aColumn);
-    void        OutputLeaderData(const otLeaderData &aLeaderData, uint16_t aColumn);
-    void        OutputNetworkDiagMacCounters(const otNetworkDiagMacCounters &aMacCounters, uint16_t aColumn);
-    void        OutputChildTableEntry(const otNetworkDiagChildEntry &aChildEntry, uint16_t aColumn);
+
+    void OutputMode(uint16_t aIndentSize, const otLinkModeConfig &aMode);
```
simonlingoogle, line 528:
```suggestion
    void OutputMode(uint8_t aIndentSize, const otLinkModeConfig &aMode);
```
nit: unit8_t is enough?

src/cli/cli.cpp
```diff
@@ -4793,6 +4723,32 @@ void Interpreter::OutputLine(const char *aFormat, ...)
...
+}
+
+void Interpreter::OutputSpaces(uint16_t aCount)
+{
+    static const char kSpaces[] = {' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ', ' '};
+    uint16_t          len;
+
+    while (aCount > 0)
+    {
+        len = OT_MIN(aCount, sizeof(kSpaces));
```
simonlingoogle, line 4746:
```suggestion
        uint16_t len = OT_MIN(aCount, sizeof(kSpaces));

```
nit: put `uint16_t len` in smaller scope?

simonlingoogle (approved):
LGTM :+1: 

with nits

src/cli/cli.hpp
```diff
@@ -194,6 +205,25 @@ class Interpreter
...
+     */
+    void OutputLine(uint16_t aIndentSize, const char *aFormat, ...);
+
+    /**
+     * This method writes a given number of space chars to the CLI console.
+     *
+     * @param[in] aCount  Number of space chars to output.
+     *
+     */
+    void OutputSpaces(uint16_t aCount);
```
bukepo, line 225: Given this is not code that not necessary to consider tab width difference, is it possible to use tab for indentation?

src/cli/cli.hpp
abtink: changed to` uint8_t

src/cli/cli.hpp
abtink: If you are suggesting to use `'\t'` instead of `' '` (space) for indentation, I won't recommend it as the actual tab width would be platform dependent and I think it would be better to have a consistent output for CLI commands on all platforms - Also we used to use `\t` in some of the logs and it caused problems on some internal products/platforms.

Existing code already seems to uses different indentation sizes of 4, 8, **10**, 12 in different places (e.g. see `OutputRoute()`).

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #5848, opened 2020-11-23
[ip6] enable multicast loop

bukepo:
This commit enables the feature IPV6_MULTICAST_LOOP into OpenThread IPv6
stack. The default-enabled feature allows a node receives a multicast
message sent by its own.


bukepo:
solves #openthread/ot-br-posix#610.

include/openthread/ip6.h
```diff
@@ -223,6 +223,7 @@ typedef struct otMessageInfo
                                  ///< Otherwise, specifies the IPv6 Hop Limit.
     bool mIsHostInterface : 1;   ///< TRUE if packets sent/received via host interface, FALSE otherwise.
     bool mAllowZeroHopLimit : 1; ///< TRUE to allow IPv6 Hop Limit 0 in `mHopLimit`, FALSE otherwise.
+    bool mMulticastLoop : 1;     ///< TRUE to allow looping back mutlicast, FALSE otherwise.
```
abtink, line 226: With this we allow users to select which msgs would allow loop back and which won't.
Can we consider allowing it for all msgs ? Any side-effect/reason not to? Thanks.
Or can we somehow deduce this in another way? Thanks.

src/core/common/message.hpp
```diff
@@ -529,6 +530,23 @@ class Message : public Buffer
...
     bool IsSubTypeMle(void) const;
 
+    /**
+     * This method checks whether this multicast message may be looped back.
+     *
+     * @retval TRUE   If message may be looped back.
+     * @retval FALSE  If message must not be looped back.
+     *
+     */
+    bool GetMulticastLoop(void) const { return GetMetadata().mMulticastLoop; }
```
abtink, line 540: How about `GetAllowMulticastLoop()` and `SetAllowMulticastLoop()`?

src/core/net/ip6.cpp
```diff
@@ -934,29 +936,58 @@ otError Ip6::HandleExtensionHeaders(Message &    aMessage,
...
-otError Ip6::HandlePayload(Message &aMessage, MessageInfo &aMessageInfo, uint8_t aIpProto)
+void Ip6::HandlePayload(Message &          aMessage,
+                        MessageInfo &      aMessageInfo,
+                        uint8_t            aIpProto,
+                        Message::Ownership aMessageOwnership)
 {
-    otError error = OT_ERROR_NONE;
+    otError  error = OT_ERROR_NONE;
+    Message *message;
+
```
abtink, line 946: We use the `message` only if the `aIpProto` is UDP or ICMP6. 
Can we check this at the top of this method, before potentially creating a `Clone()` of message (which may not then be used)?

src/core/net/ip6.cpp
```diff
@@ -934,29 +936,58 @@ otError Ip6::HandleExtensionHeaders(Message &    aMessage,
...
+    {
+    case Message::kTakeCustody:
+        message = &aMessage;
+        break;
+
+    case Message::kCopyToUse:
+        VerifyOrExit((message = aMessage.Clone()) != nullptr, error = OT_ERROR_NO_BUFS);
+        break;
+
+    default:
```
abtink, line 957: We can consider removing the `default` case here  (mainly  since the `Ownership` is an `enum` type with two enumerator values and we get a warning/error if a enumerator case value is not covered in `switch()`.

src/core/net/ip6.cpp
```diff
@@ -934,29 +936,58 @@ otError Ip6::HandleExtensionHeaders(Message &    aMessage,
...
+exit:
+    if (error != OT_ERROR_NONE)
+    {
+        otLogNoteIp6("Failed to handle payload: %s", otThreadErrorToString(error));
+    }
+
+    if (message != nullptr)
+    {
+        message->Free();
+    }
```
abtink, line 990: We can use `FreeMessage(message)` here.

src/core/net/ip6.cpp
```diff
@@ -1199,9 +1236,16 @@ otError Ip6::HandleDatagram(Message &aMessage, Netif *aNetif, const void *aLinkM
...
 #endif
 
-        IgnoreError(ProcessReceiveCallback(aMessage, messageInfo, nextHeader, aFromNcpHost, Message::kCopyToUse));
+        error = ProcessReceiveCallback(aMessage, messageInfo, nextHeader, aFromNcpHost, Message::kCopyToUse);
+
+        if ((error == OT_ERROR_NONE || error == OT_ERROR_NO_ROUTE) && forward &&
+            !ShouldForwardToThread(messageInfo, aFromNcpHost))
+        {
+            forward = false;
+        }
```
abtink, line 1245: The `if()` block above seems to be new logic/code.  
@bukepo can you please help explain what it does (why we need it)? Thanks.

abtink (commented):
Overall looks fine. Thanks Few questions and suggestions below.

include/openthread/ip6.h
bukepo: If we allow all, this would increase overhead to all nodes sending multicast messages, including Thread protocol messages (MLE advertisements, link requests, address queries). POSIX also provides a flag controlling this behavior but per socket instead of per message.

src/core/common/message.hpp
bukepo: The flag name is from the most common used flag IPV6_MULTICAST_LOOP. Is it better to keep it as is?

src/core/net/ip6.cpp
bukepo: Good point! I was considering refactoring this part in a separate PR. Mainly it will carry the message lifecycle in its own, thus each layer can better handle it.

src/core/net/ip6.cpp
bukepo: Yes, but the compiler doesn't buy there are only two cases. It complains message may not be initialized. I am fine with initialize it.

src/cli/cli.cpp
```diff
@@ -4459,7 +4461,11 @@ void Interpreter::HandleDiagnosticGetResponse(otError aError, const otMessage *a
 
     SuccessOrExit(aError);
 
-    OutputFormat("DIAG_GET.rsp/ans: ");
+    OutputFormat("DIAG_GET.rsp/ans from ");
```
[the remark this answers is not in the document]
bukepo: Yes, required in the test case.

src/core/net/ip6_mpl.hpp
```diff
@@ -216,12 +216,14 @@ class Mpl : public InstanceLocator, private NonCopyable
      * @param[in]  aMessage    A reference to the message.
      * @param[in]  aAddress    A reference to the IPv6 Source Address.
      * @param[in]  aIsOutbound TRUE if this message was locally generated, FALSE otherwise.
+     * @param[out] aReceive    Set to FALSE if this message is already processed before,
+     *                         therefore must not be process receive again, untouched otherwise.
```
Irving-cl, line 220:
```suggestion
     *                         therefore must not be processed with receiving again, untouched otherwise.
```

src/posix/platform/udp.cpp
```diff
@@ -425,8 +425,22 @@ otError otPlatUdpSend(otUdpSocket *aUdpSocket, otMessage *aMessage, const otMess
...
+
         VerifyOrExit(len == otMessageRead(aMessage, 0, payload, len), error = OT_ERROR_INVALID_ARGS);
         SuccessOrExit(error = transmitPacket(fd, payload, len, *aMessageInfo));
+
+        if (aMessageInfo->mMulticastLoop)
+        {
+            int value = 0;
+            VerifyOrExit(setsockopt(fd, IPPROTO_IPV6, IPV6_MULTICAST_LOOP, &value, sizeof(value)) == 0,
+                         error = OT_ERROR_FAILED);
+        }
```
Irving-cl, line 443: What if the code exits before? Then then socket doesn't change back. Shall we add it after `exit`?

src/core/net/ip6.cpp
```diff
@@ -934,29 +936,58 @@ otError Ip6::HandleExtensionHeaders(Message &    aMessage,
...
 
     default:
         break;
     }
 
-    return error;
+exit:
+    if (error != OT_ERROR_NONE)
+    {
+        otLogNoteIp6("Failed to handle payload: %s", otThreadErrorToString(error));
```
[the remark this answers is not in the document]
bukepo: I don't think failing to handle the payload is a big issue, so no need to use warn level. This also keeps consistent with line 969. Thoughts?

src/core/net/ip6.cpp
```diff
@@ -1126,28 +1157,34 @@ otError Ip6::HandleDatagram(Message &aMessage, Netif *aNetif, const void *aLinkM
...
+            netif = aNetif;
         }
         else
         {
             forward = true;
+
+            netif = &Get<ThreadNetif>();
+        }
+
+        if ((aNetif != nullptr || aMessage.GetMulticastLoop()) && netif->IsMulticastSubscribed(header.GetDestination()))
```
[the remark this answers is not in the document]
bukepo: `netif` is always not `nullptr`.

src/core/net/ip6.cpp
bukepo: This prevents `ProcessReceiveCallback` be called multiple times. When line 1239 succeeds, I don't think the host needs to receive the message again. And if it fails for no route, another trial doesn't help. However this doesn't prevent trying again for failures caused by insufficient memory.

src/posix/platform/udp.cpp
bukepo: good point.

src/core/common/message.hpp
Moandor-y: How about `IsMulticastLoopEnabled()` since it's `bool`?

include/openthread/ip6.h
simonlingoogle: Allowing all would mean a significant change to the current OpenThread behavior. Might have some unexpected issues.

src/posix/platform/udp.cpp
```diff
@@ -415,18 +415,33 @@ otError otPlatUdpConnect(otUdpSocket *aUdpSocket)
...
+    if (aMessageInfo->mMulticastLoop)
     {
-        uint8_t  payload[kMaxUdpSize];
-        uint16_t len = otMessageGetLength(aMessage);
+        int value = 0;
 
-        VerifyOrExit(len == otMessageRead(aMessage, 0, payload, len), error = OT_ERROR_INVALID_ARGS);
-        SuccessOrExit(error = transmitPacket(fd, payload, len, *aMessageInfo));
+        VerifyOrDie(setsockopt(fd, IPPROTO_IPV6, IPV6_MULTICAST_LOOP, &value, sizeof(value)) == 0,
+                    error = OT_ERROR_FAILED);
```
simonlingoogle, line 444: specify error exit code?

src/core/net/ip6.cpp
```diff
@@ -1199,9 +1235,17 @@ otError Ip6::HandleDatagram(Message &aMessage, Netif *aNetif, const void *aLinkM
...
+
+        if ((error == OT_ERROR_NONE || error == OT_ERROR_NO_ROUTE) && forward &&
+            !ShouldForwardToThread(messageInfo, aFromNcpHost))
+        {
+            forward = false;
+        }
 
-        SuccessOrExit(error = HandlePayload(aMessage, messageInfo, nextHeader));
+        error =
+            HandlePayload(aMessage, messageInfo, nextHeader, (forward ? Message::kCopyToUse : Message::kTakeCustody));
```
simonlingoogle, line 1247: SuccessOrExit is not used anymore?

src/core/net/ip6.cpp
```diff
@@ -1199,9 +1235,17 @@ otError Ip6::HandleDatagram(Message &aMessage, Netif *aNetif, const void *aLinkM
...
 #endif
 
-        IgnoreError(ProcessReceiveCallback(aMessage, messageInfo, nextHeader, aFromNcpHost, Message::kCopyToUse));
+        error = ProcessReceiveCallback(aMessage, messageInfo, nextHeader, aFromNcpHost, Message::kCopyToUse);
+
+        if ((error == OT_ERROR_NONE || error == OT_ERROR_NO_ROUTE) && forward &&
+            !ShouldForwardToThread(messageInfo, aFromNcpHost))
+        {
+            forward = false;
+        }
```
simonlingoogle, line 1244:
```suggestion
        forward = forward && ShouldForwardToThread(messageInfo, aFromNcpHost);
```
Let's skip checking errors? I think retrying won't help in any case.

src/core/net/ip6.cpp
```diff
@@ -1199,9 +1235,17 @@ otError Ip6::HandleDatagram(Message &aMessage, Netif *aNetif, const void *aLinkM
         }
 #endif
 
-        IgnoreError(ProcessReceiveCallback(aMessage, messageInfo, nextHeader, aFromNcpHost, Message::kCopyToUse));
+        error = ProcessReceiveCallback(aMessage, messageInfo, nextHeader, aFromNcpHost, Message::kCopyToUse);
+
+        if ((error == OT_ERROR_NONE || error == OT_ERROR_NO_ROUTE) && forward &&
+            !ShouldForwardToThread(messageInfo, aFromNcpHost))
```
simonlingoogle, line 1241: We are potentially calling `ShouldForwardToThread` twice, however calling `ShouldForwardToThread` might be costly. 
Can we optimize that `ShouldForwardToThread` is called only once?
Maybe we can calculate `bool forwardToThread = ShouldForwardToThread(messageInfo, aFromNcpHost)`; just after the `forward` value is determined. Then, we can just use `forwardToThread` instead of calling `ShouldForwardToThread`. 

Thoughts?

Irving-cl (approved):
LGTM 👍

src/core/net/ip6.cpp
bukepo: It helps when failed for insufficient memory.

src/core/net/ip6.cpp
bukepo: To make sure turning off loop even transmit failed.

src/posix/platform/udp.cpp
bukepo:
```suggestion
                    OT_EXIT_ERROR_ERRNO);
```

src/core/common/message.hpp
abtink: I am fine with it as is.
Also like the @Moandor-y suggestion.

abtink (approved):
LGTM. Thanks

Moandor-y (approved):
LGTM

simonlingoogle (approved):
LGTM 👍

src/core/common/message.hpp
jwhui: @bukepo , thoughts on `IsMulticastLoopEnabled()`?

src/core/net/ip6_mpl.hpp
```diff
@@ -216,12 +216,14 @@ class Mpl : public InstanceLocator, private NonCopyable
      * @param[in]  aMessage    A reference to the message.
      * @param[in]  aAddress    A reference to the IPv6 Source Address.
      * @param[in]  aIsOutbound TRUE if this message was locally generated, FALSE otherwise.
+     * @param[out] aReceive    Set to FALSE if the MPL message is a duplicate therefore must not
+     *                         go receiving process again, untouched otherwise.
```
jwhui, line 220:
```suggestion
     * @param[out] aReceive    Set to FALSE if the MPL message is a duplicate must not
     *                         go through the receiving process again, untouched otherwise.
```

src/core/common/message.hpp
bukepo: I'm fine with changing the name, but prefer doing it later so that the PR can be merged without re-running all the CI tests.

src/posix/platform/udp.cpp
```diff
@@ -415,18 +415,32 @@ otError otPlatUdpConnect(otUdpSocket *aUdpSocket)
...
     fd = FdFromHandle(aUdpSocket->mHandle);
 
+    len = otMessageGetLength(aMessage);
+    VerifyOrExit(len == otMessageRead(aMessage, 0, payload, len), error = OT_ERROR_INVALID_ARGS);
+
+    if (aMessageInfo->mMulticastLoop)
+    {
+        int value = 1;
+
+        VerifyOrExit(setsockopt(fd, IPPROTO_IPV6, IPV6_MULTICAST_LOOP, &value, sizeof(value)) == 0,
```
jwhui, line 433: Should we make failures here more visible? Log? Die?

src/core/common/message.hpp
librasungirl: Another proposal `IsMulticastLoop()` 
- `Loop` itself is descriptive
- seems more match with below `SetMulticastLoop()`

bukepo: Just take another look, if rename the getter to `IsMulticastLoopEnabled()`, what's the proper name for the coresponding setter? If the current flag name is already common used, I prefer to keep it as is. Thoughts? @abtink @Moandor-y @jwhui @librasungirl

abtink: I am still fine with it as is :)

If you want to change it, for setter name, we can follow existing pattern from `IsLinkSecurityEnabled()` / `SetLinkSecurityEnabled()` or `IsTimeSync()` / `SetTimeSync()`, etc.  So we can go for `IsMulticastLoopEnabled()` and `SetMulticastLoopEnabled()`.

librasungirl: No strong preference, fine to me to keep it as is.

Moandor-y: I prefer `IsMulticastLoopEnabled()` and `SetMulticastLoopEnabled()`. I'm fine with it as is too.

librasungirl (approved):

jwhui (approved):

PR #5852, opened 2020-11-24
[scripts] test multicast routing across Thread and Backbone

simonlingoogle:
- [x] This commit fixes some issue in multicast forwarding from Thread to Backbone. 
- [x] It also adds a test to verify that multicast forwarding works across Thread and Backbone. 
  - [x] Test script `tests/scripts/thread-cert/mcast6.py` is added for Host to subscribe to a multicast address

**The ping reply can not reach ROUTER1 since DUA feature is not complete. But verified locally that ping works with https://github.com/openthread/ot-br-posix/pull/586 merged**

### DIFF
--- a/src/core/net/ip6.cpp
+++ b/src/core/net/ip6.cpp
@@ -533,7 +533,7 @@ void Ip6::HandleSendQueue(void)
     }
 }
 
-otError Ip6::HandleOptions(Message &aMessage, Header &aHeader, bool aForward, bool &aReceive)
+otError Ip6::HandleOptions(Message &aMessage, Header &aHeader, bool aIsOutbound, bool &aReceive)
 {
     otError        error = OT_ERROR_NONE;
     HopByHopHeader hbhHeader;
@@ -563,7 +563,7 @@ otError Ip6::HandleOptions(Message &aMessage, Header &aHeader, bool aForward, bo
         switch (optionHeader.GetType())
         {
         case OptionMpl::kType:
-            SuccessOrExit(error = mMpl.ProcessOption(aMessage, aHeader.GetSource(), aForward, aReceive));
+            SuccessOrExit(error = mMpl.ProcessOption(aMessage, aHeader.GetSource(), aIsOutbound, aReceive));
             break;
 
         default:
@@ -890,7 +890,7 @@ otError Ip6::HandleExtensionHeaders(Message &    aMessage,
                                     MessageInfo &aMessageInfo,
                                     Header &     aHeader,
                                     uint8_t &    aNextHeader,
-                                    bool         aForward,
+                                    bool         aIsOutbound,
                                     bool         aFromNcpHost,
                                     bool &       aReceive)
 {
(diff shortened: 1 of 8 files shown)


tests/scripts/thread-cert/backbone/test_mlr_multicast_routing_across_thread_pans.py
```diff
@@ -0,0 +1,184 @@
...
+    TOPOLOGY = {
+        PBBR1: {
+            'name': 'PBBR1',
+            'allowlist': [ROUTER1],
+            'is_otbr': True,
+            'version': '1.2',
+            'router_selection_jitter': 1,
+            'bbr_registration_jitter': BBR_REGISTRATION_JITTER,
+            'channel': CH1,
+            'prefer_router_id': PBBR1,  # Use prefer_router_id to avoid Router ID conflicts in the two channels.
```
librasungirl, line 71: Is it necessary? Any issues if not?

librasungirl: In real world, same short addresses in different partitions should be allowed. 

Not sure if the failure you met is specific for simulation, fine to me to keep it as is if yes.
But I would suggest to remove the prefer_router_id configuration and figure out the root cause for the failure when it occurs, just in case there is some real issue which should be fixed.

Thoughts?

simonlingoogle: removed.

tests/scripts/thread-cert/backbone/test_mlr_multicast_routing_across_thread_pans.py
```diff
@@ -0,0 +1,230 @@
...
+        # PBBR2 should forward ping reply to the Backbone link
+        ping_reply_pkts.filter_eth_src(PBBR2_ETH).must_next()
+
+        #
+        # Verify R1 pings MA2 to Host and R2
+        #
+
+        # ROUTER1 should send the multicast ping request
+        ping_ma2 = pkts.filter_wpan_src64(ROUTER1).filter_AMPLFMA().filter_ping_request().filter(
+            lambda p: p.ipv6.hlim == 64).must_next()
```
librasungirl, line 209: One suggest here is may use seed id to find the packet that originated from ROUTER1

librasungirl (commented):
LGTM overall with one minor suggestion

tests/scripts/thread-cert/backbone/test_mlr_multicast_routing_across_thread_pans.py
simonlingoogle: revised to check MPL seed_id

librasungirl (approved):

src/core/net/ip6.cpp
```diff
@@ -1186,29 +1189,34 @@ otError Ip6::HandleDatagram(Message &aMessage, Netif *aNetif, const void *aLinkM
     }
     else
     {
+        // unicast
```
bukepo, line 1192:
```suggestion
```

bukepo (approved):

src/core/net/ip6.cpp
simonlingoogle: the comment tells that this else block is for unicast. Doesn't it make the code more readable?

jwhui (approved):

PR #5853, opened 2020-11-24
[low-power] frame pending bit handling for enh-ack in simulation

brianljt:
MAC command frame has its command ID encrypted in frame version 802.15.4-2015. Radio driver cannot decrypt this frame payload without extended address. So the frame pending bit is not correctly set in the ACK in response to data request command with version 2015. This PR adds a special handling for frame version 2015 and only checks the frame type when deciding if it needs to set the frame pending bit.

### DIFF
--- a/examples/platforms/simulation/radio.c
+++ b/examples/platforms/simulation/radio.c
@@ -827,11 +827,15 @@ void radioTransmit(struct RadioMessage *aMessage, const struct otRadioFrame *aFr
 
 void radioSendAck(void)
 {
-    if (hasFramePending(&sReceiveFrame)
-#if OPENTHREAD_CONFIG_THREAD_VERSION < OT_THREAD_VERSION_1_2
-        && otMacFrameIsDataRequest(&sReceiveFrame)
+    if (
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+        // Determine if frame pending should be set
+        ((otMacFrameIsVersion2015(&sReceiveFrame) && otMacFrameIsCommand(&sReceiveFrame)) ||
+         otMacFrameIsData(&sReceiveFrame) || otMacFrameIsDataRequest(&sReceiveFrame))
+#else
+        otMacFrameIsDataRequest(&sReceiveFrame)
 #endif
-    )
+        && hasFramePending(&sReceiveFrame))
     {
         sReceiveFrame.mInfo.mRxInfo.mAckedWithFramePending = true;
     }
--- a/examples/platforms/simulation/radio.c
+++ b/examples/platforms/simulation/radio.c
@@ -827,14 +827,11 @@ void radioTransmit(struct RadioMessage *aMessage, const struct otRadioFrame *aFr
 
 void radioSendAck(void)
 {
-    if (
-#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
-        // Determine if frame pending should be set
-        (otMacFrameIsData(&sReceiveFrame) || otMacFrameIsDataRequest(&sReceiveFrame))
-#else
-        otMacFrameIsDataRequest(&sReceiveFrame)
+    if (hasFramePending(&sReceiveFrame)
+#if OPENTHREAD_CONFIG_THREAD_VERSION < OT_THREAD_VERSION_1_2
+        && otMacFrameIsDataRequest(&sReceiveFrame)
 #endif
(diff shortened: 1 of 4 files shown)


brianljt:
According to IEEE 802.15.4 - 2015, the command ID field will be encrypted.

image

As radio driver does not have child table and does not know the incoming frame's extended source address, it cannot decrypt the frame so it cannot decide if the received frame is a data request or not.

According to Thread 1.2 section 3.2.3, modification is needed for 802.15.4 section 7.2,
> Modified: When supporting the Enhanced Frame Pending (see Section 3.2.6.2 Enhanced Frame Pending), Section 7.2.2.3.1 SHALL be modified as follows to indicate the Frame Pending subfield in Acknowledgments for MAC data frames: The MHR for an acknowledgment frame shall contain only the Frame Control field and the Sequence Number field.In the Frame Control field, the Frame Type subfield shall contain the value that indicates an acknowledgment frame, as shown in Table 79. The device sending an acknowledgment frame shall determine whether it has data pending for the recipient. If the device can determine this before sending the acknowledgment frame (see 7.5.6.4.2), it shall set the Frame Pending subfield according to whether there is pending data.

So it is OK to set frame pending bit regardless of the incoming frame types.

For SED with version 1.1, it will just ignore the frame pending bit in response to none-data request command frame. So this change can still maintain backward compatible with pre-Thread 1.2 devices. @bukepo thoughts?

bukepo (commented):
Good catch! Is it possible to check frame types when the frame is 2015 format? We can only ack when the frame types is Data or MAC Command.

examples/platforms/simulation/radio.c
```diff
@@ -827,14 +827,11 @@ void radioTransmit(struct RadioMessage *aMessage, const struct otRadioFrame *aFr
...
 {
-    if (
-#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
-        // Determine if frame pending should be set
-        (otMacFrameIsData(&sReceiveFrame) || otMacFrameIsDataRequest(&sReceiveFrame))
-#else
-        otMacFrameIsDataRequest(&sReceiveFrame)
+    if (hasFramePending(&sReceiveFrame)
+#if OPENTHREAD_CONFIG_THREAD_VERSION < OT_THREAD_VERSION_1_2
+        && otMacFrameIsDataRequest(&sReceiveFrame)
```
Irving-cl, line 832: In Thread spec 1.2 3.2.6.2:

> When both the polling rx-off-when-idle Thread Interface and the MAC Coordinator Thread Interface are implementing the Enhanced Frame Pending feature and the Coordinator has determined that indirect buffered data is available for  the rx-off-when-idle interface, the Coordinator SHALL set the Frame Pending subfield value to  '1' in MAC Acknowledgement frames sent in response to **MAC Data frames** from the rx-off-when-idle interface.

I think for 1.2, we still need to ensure it's a Mac data or Mac Cmd frame. We don't need to do this for Beacon right?

Irving-cl:
> Good catch! Is it possible to check frame types when the frame is 2015 format? We can only ack when the frame types is Data or MAC Command.

@brianljt 

I think we can always check the frame type. The problem is that in 2015 frames we cannot get the Command id if the frames is a MAC Command.

examples/platforms/simulation/radio.c
brianljt: Thanks for the review. Updated.

Since radio cannot do decryption without ext address to short address mapping table, I think for now, we can only make the check as close to spec as possible. I've changed the check to 2015 frame && frame is of type command.

@Irving-cl @bukepo

src/core/mac/mac_frame.hpp
```diff
@@ -733,7 +733,8 @@ class Frame : public otRadioFrame
     otError SetCommandId(uint8_t aCommandId);
 
     /**
-     * This method indicates whether the frame is a MAC Data Request command (data poll)
+     * This method indicates whether the frame is a MAC Data Request command (data poll).
+     * For 802.15.4-2015 and above frame, the frame should be already decrypted.
```
Irving-cl, line 737:
```suggestion
     * This method indicates whether the frame is a MAC Data Request command (data poll).
     
     * For 802.15.4-2015 and above frame, the frame should be already decrypted.
```
nit: As an [example](https://github.com/openthread/openthread/blob/master/src/core/mac/mac.hpp#L234)

Irving-cl (approved):
LGTM 👍  Thanks.

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #5857, opened 2020-11-24
[cmake] avoid unnecessary rebuilds

bukepo:
This commit avoid unnecessary rebuilds. The issue is caused by the
openthread-mbedtls-config target added by add_custom_target will always
be run, so the generated openthread-mbedtls-config.h is always
re-generated. This commit changes to use add_custom_command to generate
the header file.

Test:
```bash
./script/test clean build
# the next command should cause no files rebuilt
./script/test build
touch third_party/mbedtls/mbedtls-config.h
# the next command should cause all targets depends on mbedtls rebuilt
./script/test build
```

### DIFF
--- a/third_party/mbedtls/CMakeLists.txt
+++ b/third_party/mbedtls/CMakeLists.txt
@@ -26,7 +26,7 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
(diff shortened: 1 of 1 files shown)


Moandor-y (approved):
LGTM

jwhui (approved):

PR #5862, opened 2020-11-25
[low-power] enable CCA sampling before CSL transmission

brianljt:
According to spec, CCA should be performed before CSL transmission. This commit enables CCA in CSL transmission in NRF platform.


brianljt:
In Thread 1.2 specification section 3.2.6.3.4 CSL Phase Definition for Thread Interfaces

> The CSL Transmitter SHALL start transmitting the first symbol of the preamble of the frame to transmit at the time specified in the CSL Phase (that is, CCA must be performed before).

src/core/mac/sub_mac.cpp
```diff
@@ -357,9 +357,10 @@ void SubMac::StartCsmaBackoff(void)
         if (ShouldHandleTransmitTargetTime())
         {
             if (Time(static_cast<uint32_t>(otPlatRadioGetNow(&GetInstance()))) <
-                Time(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime) + mTransmitFrame.mInfo.mTxInfo.mTxDelay)
+                Time(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime) + mTransmitFrame.mInfo.mTxInfo.mTxDelay -
+                    kCcaSampleInterval)
```
hubertmis, line 361: I agree that a margin should be added here to avoid too late requesting of precise 802.15.4 transmissions. However, I think the duration of this delays should depend on two factors:

1. If CCA is expected before the frame being transmitted (to be on a safe side, we can assume that CCA is always expected, like here).
2. What radio is being used. For sure this margin should be a little longer for nRF5340 than for nRF52840 (because of serialization of this call between CPUs). I expect other vendors would require different timings, especially when the turnaround time differs.

How about adding a function to radio platform that would report how long delay is required?

src/core/mac/sub_mac.cpp
```diff
@@ -357,9 +357,10 @@ void SubMac::StartCsmaBackoff(void)
...
             {
-                mTimer.StartAt(Time(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime),
+                mTimer.StartAt(Time(mTransmitFrame.mInfo.mTxInfo.mTxDelayBaseTime) - kCcaSampleInterval,
                                mTransmitFrame.mInfo.mTxInfo.mTxDelay);
             }
             else // Transmit without delay
```
hubertmis, line 366: BTW, I think that when this `else` is executed, the transmission should be skipped, and error logged, and a new attempt scheduled. Transmitting without delay would most probably be out of RX window of the peer, what is unnecessary channel bandwidth and energy waste.

hubertmis (commented):
The proposed change partially fixes a timing problem related to scheduled TX operations. I've added a proposal to fully fix it.

Irving-cl (approved):
LGTM :+1:

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -477,9 +477,8 @@ otError otPlatRadioTransmit(otInstance *aInstance, otRadioFrame *aFrame)
 #if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
     if (aFrame->mInfo.mTxInfo.mTxDelay != 0)
     {
-        if (!nrf_802154_transmit_raw_at(&aFrame->mPsdu[-1], aFrame->mInfo.mTxInfo.mCsmaCaEnabled,
-                                        aFrame->mInfo.mTxInfo.mTxDelayBaseTime, aFrame->mInfo.mTxInfo.mTxDelay,
-                                        aFrame->mChannel))
+        if (!nrf_802154_transmit_raw_at(&aFrame->mPsdu[-1], true, aFrame->mInfo.mTxInfo.mTxDelayBaseTime,
```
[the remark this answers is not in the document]
brianljt: Yes, this is the intended behavior for a CSL transmission. It should skip random backoff but still do 128µs of CCA before transmission. Currently there are no transmissions other than CSL in OT MAC requires delayed transmission. So I just put `true` here.

For a platform which does not support delayed transmission(`OT_RADIO_CAPS_TRANSMIT_TIMING`) but supports random backoff(`OT_RADIO_CAPS_CSMA_BACKOFF`), this field `aFrame->mInfo.mTxInfo.mCsmaCaEnabled` is used to indicate the platform to stop doing random backoff. The transmit timing is handled in OT SubMac.

src/core/mac/sub_mac.cpp
brianljt: For platform which supports delay transmission(`OT_RADIO_CAPS_TRANSMIT_TIMING`), `ShouldHandleTransmitTargetTime()` will return false and the frame will be delivered to Radio driver directly with `mTxDelayBaseTime` and `mTxDelay` fields set. The frame will arrives at Radio driver about `OPENTHREAD_CONFIG_MAC_CSL_REQUEST_AHEAD_US`(2000ms) ahead of the target transmission time. Radio driver can adjust the delay according to CCA, Tx ramp up, Tx setup time and software delay by itself. Just like what is done in Nordic platform `nrf_802154_delayed_trx_transmit()`.

The handling here is for the platform which does not support delay transmission(`OT_RADIO_CAPS_TRANSMIT_TIMING`). It will rely on `SubMac` to do the job. I agree for this case, we should define some platform specific delay to compensate the time lag between `otPlatRadioTransmit` and when the first symbol of the frame is finally sent.

src/core/mac/sub_mac.cpp
brianljt: I agree.

src/core/mac/sub_mac.cpp
hubertmis: > For platform which supports delay transmission(`OT_RADIO_CAPS_TRANSMIT_TIMING`), `ShouldHandleTransmitTargetTime()` will return false

I wasn't aware of that. I thought that this code is for platforms supporting delayed transmissions. Thank you for explanation.

hubertmis (approved):

src/core/mac/sub_mac.cpp
brianljt: Thanks @hubertmis for the review. I will mark this comment and the next one resolved for this PR. For enhancements to support the platforms which does not have delayed transmission, I will do it as a separate PR.

For the timing differences between nRF5340 and nRF52840, if they both supports `OT_RADIO_CAPS_TRANSMIT_TIMING`, then I think you can define different `TX_SETUP_TIME` or `TX_RAMP_UP_TIME` in your drivers for different chips.

hubertmis: > For the timing differences between nRF5340 and nRF52840, if they both supports `OT_RADIO_CAPS_TRANSMIT_TIMING`, then I think you can define different `TX_SETUP_TIME` or `TX_RAMP_UP_TIME` in your drivers for different chips.

On the 802.15.4 radio driver side, we'll of course trim constants to make sure that everything works correctly. However, I think that there might be required some changes in OT, like selecting `OPENTHREAD_CONFIG_MAC_CSL_REQUEST_AHEAD_US` value depending on a SoC. nRF5340 may require a little longer time than nRF52.

But we can handle it later, when we introduce nRF53 support.

brianljt: Yes, if default 2ms is still not enough for nRF53, then this value should be set larger.

BTW, we found one note [here](https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.thread_zigbee.v2.0.0%2Fgroup__nrf__802154__delayed__trx.html) which is different from the function description in nrf_802154.h. Is this note still valid?

> Delayed transmission does not timeout waiting for ACK automatically. Waiting for ACK shall be timed out by the next higher layer or the ACK timeout module. The ACK timeout timer shall start when the nrf_802154_tx_started function is called.

As Nordic driver claims the capability of `OT_RADIO_CAPS_ACK_TIMEOUT`, so OT stack does not start a timer to time out a transmission and will wait for callback from Radio for a transmission.

hubertmis: The limitation you found is not present anymore. The radio driver times out waiting for ACKs by itself for both delayed and typical TX operations.

brianljt: Thanks for the confirmation.

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #5863, opened 2020-11-25
[spinel] add RCP API Version and compatibility check

abtink:

**[spinel] add RCP API Version and compatibility check**

This commit adds a new compatibility check model for RCP and host. It
updates the compatibility guideline documentation stating how the
compatibility policy for NCP versus RCP are handled differently. For
NCP, we continue to require new NCP firmware to remain backward
compatible and work with an older host driver.

For RCP and host, this commit introduces the RCP API Version model.
The API version numbers are used to check the compatibility between
host implementation and RCP firmware. Generally, a newer host side
implementation would work with a range of previous or older RCP
firmware versions.

This commit adds `SPINEL_RCP_API_VERSION` definition which  specifies
the current spinel RCP API version. This number MUST be incremented
anytime there is a change in any of RCP specific spinel definitions.
`SPINEL_MIN_HOST_SUPPORTED_RCP_API_VERSION` specifies the minimum
spinel RCP API version which is supported by the host-side
implementation.

This commit also reserves a block of spinel properties value ranges
from `0xB0 - 0xFF` for RCP related property definitions. The smaller
property values help decrease the encoded byte stream length,
resulting in shorter spinel frames exchanged between RCP and host.

---------

**[radio-spinel] check RCP API Version to be in the supported range**

This commit adds a new method `CheckRcpApiVersion()` which is used
during `Init()` to get the API version from RCP and verify that it
is within supported range.


bukepo (approved):
LGTM:+1:

brianljt (approved):
Looks great. Thank you! 👍

src/lib/spinel/spinel.h
```diff
@@ -305,6 +308,24 @@
...
+ *   compatibility between host implementation and RCP firmware. Generally,
+ *   a newer host side implementation would work with a range of previous
+ *   or older RCP firmware versions.
+ *
+ *   - SPINEL_RCP_API_VERSION specifies the current spinel RCP API version.
+ *     This number MUST be incremented anytime there is a change in any of RCP
+ *     specific spinel definitions.
+ *
+ *   - SPINEL_MIN_HOST_SUPPORTED_RCP_API_VERSION specifies the minimum spinel
+ *      RCP API Version which is supported by the host-side implementation.
```
jwhui, line 323:
```suggestion
 *   - SPINEL_MIN_HOST_SUPPORTED_RCP_API_VERSION specifies the minimum spinel
 *     RCP API Version which is supported by the host-side implementation.
```

src/lib/spinel/spinel.h
```diff
@@ -305,6 +308,24 @@
...
+ *
+ *   - On start host implementation would query the RCP API version and accept
```
jwhui, line 325:
```suggestion
 *   - On start, host implementation queries the RCP API version and accepts
```

src/lib/spinel/spinel.h
```diff
@@ -305,6 +308,24 @@
...
+ *     any version number from SPINEL_MIN_HOST_SUPPORTED_RCP_API_VERSION up to
+ *     SPINEL_RCP_API_VERSION.
```
jwhui, line 327:
```suggestion
 *     and including SPINEL_RCP_API_VERSION.
```

jwhui (approved):

PR #5867, opened 2020-11-26
[radio] send max power table to radio

gjc13:
This change sends the max power table to the radio so that the radio
chip can manage the transmit power by self.


brianljt (commented):
Also delete max_power_table.hpp and update corresponding make files?

brianljt (approved):
LGTM 👍

src/posix/platform/radio.cpp
```diff
@@ -495,21 +488,13 @@ void otPlatDiagAlarmCallback(otInstance *aInstance)
 uint32_t otPlatRadioGetSupportedChannelMask(otInstance *aInstance)
 {
     OT_UNUSED_VARIABLE(aInstance);
-    return
-#if OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE
-        sMaxPowerTable.GetAllowedChannelMask() &
-#endif
-        sRadioSpinel.GetRadioChannelMask(false);
+    return sRadioSpinel.GetRadioChannelMask(false);
```
bukepo, line 491: This changes the behavior, that is the supported channel can be dynamically changed as well. How about moving the max power table into radio spinel itself, so that the RCP restoration can use it as well?

Irving-cl (approved):
LGTM 👍

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1835,5 +1836,16 @@ uint32_t RadioSpinel<InterfaceType, ProcessContextType>::GetBusSpeed(void) const
     return mSpinelInterface.GetBusSpeed();
 }
 
+template <typename InterfaceType, typename ProcessContextType>
+otError RadioSpinel<InterfaceType, ProcessContextType>::SetChannelMaxTransmitPower(uint8_t aChannel, int8_t aMaxPower)
+{
+    otError error = OT_ERROR_NONE;
+    VerifyOrExit(aChannel >= Radio::kChannelMin && aChannel <= Radio::kChannelMax, error = OT_ERROR_INVALID_ARGS);
+    error = Set(SPINEL_PROP_PHY_CHAN_MAX_POWER, SPINEL_DATATYPE_UINT8_S SPINEL_DATATYPE_INT8_S, aChannel, aMaxPower);
```
Moandor-y, line 1844: Does this table need to be saved and restored across RCP failure reboots?

examples/platforms/simulation/radio.c
```diff
@@ -1155,3 +1155,12 @@ void otPlatRadioSetMacFrameCounter(otInstance *aInstance, uint32_t aMacFrameCoun
 
     sMacFrameCounter = aMacFrameCounter;
 }
+
+otError otPlatRadioSetChannelMaxTransmitPower(otInstance *aInstance, uint8_t aChannel, int8_t aMaxPower)
```
Moandor-y, line 1159: Implement the table on simulation so that we can test this API?

src/posix/platform/radio.cpp
gjc13: Done

Moandor-y (approved):
LGTM :+1:

include/openthread/platform/radio.h
```diff
@@ -946,6 +946,36 @@ otError otPlatRadioEnableCsl(otInstance *aInstance, uint32_t aCslPeriod, const o
...
 
+/**
+ * Set the max transmit power for a specific channel.
+ *
+ * @param[in]  aInstance    The OpenThread instance structure.
+ * @param[in]  aChannel     The radio channel.
+ * @param[in]  aMaxPower    The max power in dBm.
+ *
+ */
+otError otPlatRadioSetChannelMaxTransmitPower(otInstance *aInstance, uint8_t aChannel, int8_t aMaxPower);
```
bukepo, line 957: `aMaxPower == 0x7f` already indicates the channel is not allowed. can we eliminate `otPlatRadioSetChannelSupported()`?

gjc13: I'd like to treat `aMaxPower == 0x7f` as we don't specify the max power and the underlying driver is free to decide the power by self.

bukepo: It seems we should take full control of tx power for certification. Any motivation to give up the tx power limit?

gjc13: When max power table is not passed, we'll use the `otPlatRadioSetTransmitPower` instead for controlling the power. And the max power table will be used for any additional restriction because of regional radio regulations.

include/openthread/platform/radio.h
```diff
@@ -946,6 +946,36 @@ otError otPlatRadioEnableCsl(otInstance *aInstance, uint32_t aCslPeriod, const o
...
+ *
+ */
+otError otPlatRadioSetChannelSupported(otInstance *aInstance, uint8_t aChannel, bool aSupported);
+
+/**
+ * Set whether a channel is preferred.
+ *
+ * @param[in]  aInstance    The OpenThread instance structure.
+ * @param[in]  aChannel     The radio channel.
+ * @param[in]  aEnabled     Whether the channel is preferred.
```
abtink, line 974:
```suggestion
 * @param[in]  aPreferred   Whether the channel is preferred.
```

src/lib/spinel/max_power_table.hpp
```diff
@@ -26,53 +26,86 @@
...
-#ifndef OT_POSIX_PLATFORM_MAX_POWER_TABLE_HPP_
-#define OT_POSIX_PLATFORM_MAX_POWER_TABLE_HPP_
+#ifndef OT_LIB_SPINEL_MAX_POWER_TABLE_HPP_
+#define OT_LIB_SPINEL_MAX_POWER_TABLE_HPP_
 
 #include "core/radio/radio.hpp"
 
 namespace ot {
-namespace Posix {
+namespace Spinel {
```
abtink, line 35: This is moving this file under `spinel` lib and `Spinel` namespace.
The `MaxPowerTable` seems to be fully unrelated to spinel itself and more of radio driver thing.

Can we consider putting this in another lib/location? (maybe same place as before under posix?)

include/openthread/platform/radio.h
```diff
@@ -946,6 +946,36 @@ otError otPlatRadioEnableCsl(otInstance *aInstance, uint32_t aCslPeriod, const o
  */
 void otPlatRadioUpdateCslSampleTime(otInstance *aInstance, uint32_t aCslSampleTime);
 
+/**
```
abtink, line 949: Here we are adding a bunch of APIs for channel related info. To have a more flexible way and reduce the number of `otPlatRadio` APIs for this purpose I have a suggestion:
- We can define a `struct` to specify all the data/info for a channel.
- Add a single API  that sets the info for a channel.

This also allows us to extend this in future (add more fields if needed).

Example:
```
typedef struct otRadioChannelInfo
{
    bool   mSupported;
    bool   mPreferred;
    int8_t mMaxTxPower;
} otRadioChannelInfo;

void otPlatRadioSetChannelInfo(uint8_t aChannel, const otRadioChannelInfo *aChannelInfo);
```

src/lib/spinel/spinel.h
```diff
@@ -1517,6 +1517,7 @@ enum
     SPINEL_PROP_PHY_PCAP_ENABLED   = SPINEL_PROP_PHY__BEGIN + 8,  ///< [b]
     SPINEL_PROP_PHY_CHAN_PREFERRED = SPINEL_PROP_PHY__BEGIN + 9,  ///< [A(C)]
     SPINEL_PROP_PHY_FEM_LNA_GAIN   = SPINEL_PROP_PHY__BEGIN + 10, ///< dBm [c]
+    SPINEL_PROP_PHY_CHAN_MAX_POWER = SPINEL_PROP_PHY__BEGIN + 11, ///< [cC]
```
abtink, line 1520: Suggest adding more documentation/detail for this.

Mainly adding  that this is a "write only" property and describing the format `[cC]` and that it  represents "channe"l and then "max power".

BTW, I think it should be `"[Cc]"` instead. `C` is `UINT8` for channel.

abtink (commented):
Thanks @gjc13. Looks fine overall.

I have one main suggestion below on combining radio APIs into one.

include/openthread/platform/radio.h
gjc13: After discussing with @bukepo, we decided to remove the channel mask in platform api.

examples/platforms/simulation/radio.c
```diff
@@ -359,6 +368,11 @@ void platformRadioInit(void)
 #else
     sTransmitFrame.mInfo.mTxInfo.mIeInfo = NULL;
 #endif
+
+    for (size_t i = 0; i <= kMaxChannel - kMinChannel; i++)
+    {
+        sChannelMaxTransmitPower[i] = INT8_MAX;
```
bukepo, line 374: How about defining a 'OT_RADIO_POWER_FORBIDDEN == 0x7f` and use it here?

src/core/radio/max_power_table.hpp
```diff
@@ -26,53 +26,51 @@
...
      *
-     * @returns The max allowed transmit power in dBm.
+     * @returns The max supported transmit power in dBm.
      *
      */
     int8_t GetTransmitPower(uint8_t aChannel) const { return mPowerTable[aChannel - Radio::kChannelMin]; }
 
     /**
-     * This method sets the max allowed transmit power of channel @p aChannel.
+     * This method sets the max supported transmit power of channel @p aChannel.
```
bukepo, line 55: Is `allowed` more descriptive?
```suggestion
     * This method sets the max allowed transmit power of channel @p aChannel.
```

src/lib/spinel/spinel.h
```diff
@@ -1568,7 +1568,15 @@ enum
     SPINEL_PROP_PHY_PCAP_ENABLED   = SPINEL_PROP_PHY__BEGIN + 8,  ///< [b]
     SPINEL_PROP_PHY_CHAN_PREFERRED = SPINEL_PROP_PHY__BEGIN + 9,  ///< [A(C)]
     SPINEL_PROP_PHY_FEM_LNA_GAIN   = SPINEL_PROP_PHY__BEGIN + 10, ///< dBm [c]
-    SPINEL_PROP_PHY__END           = 0x30,
+
+    /// Signal the max power for a channel
```
abtink, line 1572: Can you please also update/increment the recently added `SPINEL_RCP_API_VERSION`?

This is a good case of adding new RCP APIs (property) and thinking about whether the host MIN_API_VERSION need to be updated or not. From code above looks like the host would handle the case where there is no such property on RCP.

src/lib/spinel/spinel.h
```diff
@@ -1568,7 +1568,15 @@ enum
...
+    /** Format: `Cc`
+     *
+     * First byte is the channel then the max transmit power, write-only.
+     */
+    SPINEL_PROP_PHY_CHAN_MAX_POWER = SPINEL_PROP_PHY__BEGIN + 11,
```
abtink, line 1577: Can you please add the new property in `spinel.c` in `spinel_prop_key_to_cstr()` switch statement? Thanks.

src/posix/platform/radio.cpp
```diff
@@ -147,6 +118,34 @@ void platformRadioInit(otUrl *aRadioUrl)
...
+        constexpr int8_t kPowerDefault = 30; // Default power 1 watt (30 dBm).
+        const char *     str           = nullptr;
+        uint8_t          channel       = ot::Radio::kChannelMin;
+        int8_t           power         = kPowerDefault;
+
+        for (str = strtok(const_cast<char *>(maxPowerTable), ","); str != nullptr && channel <= ot::Radio::kChannelMax;
+             str = strtok(nullptr, ","))
+        {
+            power = static_cast<int8_t>(strtol(str, nullptr, 0));
+            SuccessOrDie(sRadioSpinel.SetChannelMaxTransmitPower(channel, power));
```
abtink, line 135: I notcied in `RestoreProperties()`, there is check for if the result is `OT_ERROR_NOT_FOUND` (i.e. to protect when RCP is older version and does not support the newly added spinel property for MAX_POWER).

But here seems like here we use `SuccessOrDie` and if RCP does not support the property we would exit.

- Either we need to update this code and allow host side code to work with older RCP version, 
- Or we should update the `SPINEL_MIN_HOST_SUPPORTED_RCP_API_VERSION` along with `SPINEL_RCP_API_VERSION` to ensure/verify correct compatibility.

gjc13: I didn't add the check since it's protected in a feature macro but it looks reasonable to tolerate older RCP versions.

bukepo (approved):
LGTM:+1:

src/posix/platform/radio.cpp
```diff
@@ -147,6 +118,43 @@ void platformRadioInit(otUrl *aRadioUrl)
...
+                DieNow(OT_ERROR_FAILED);
+            }
+            ++channel;
+        }
+
+        // Use the last power if omitted.
+        while (channel <= ot::Radio::kChannelMax)
+        {
+            error = sRadioSpinel.SetChannelMaxTransmitPower(channel, power);
+            if (error != OT_ERROR_NONE && error != OT_ERROR_FAILED)
```
abtink, line 148: I think this needs to be `OT_ERROR_NOT_FOUND` instead of `OT_ERROR_FAILED`?

src/lib/spinel/spinel.h
abtink: @gjc13, I think the `SPINEL_RCP_API_VERSION` is same as before. Can you please change it? 

We should increment it since there is an RCP related change in spinel (even if it does not break compatibility).  This is similar to the OT API version (which is incremented on any change in public/platform OT APIs, even a comment update/fix).

The policy:
https://github.com/openthread/openthread/blob/d218dfcb354309600c379e307ea4742236f0bacc/src/lib/spinel/spinel.h#L311-L329

abtink (approved):
I noticed two things that I noted below which I suggest to be fixed/changed.

After those two, it all looks good to me. Thanks 👍

gjc13:
@abtink Done.

include/openthread/instance.h
```diff
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (51)
+#define OPENTHREAD_API_VERSION (52)
```
jwhui, line 56: Need to increment this value again.

include/openthread/platform/radio.h
```diff
@@ -946,6 +947,16 @@ otError otPlatRadioEnableCsl(otInstance *aInstance, uint32_t aCslPeriod, const o
...
+ * Set the max transmit power for a specific channel.
+ *
+ * @param[in]  aInstance    The OpenThread instance structure.
+ * @param[in]  aChannel     The radio channel.
+ * @param[in]  aMaxPower    The max power in dBm, passing OT_RADIO_RSSI_INVALID will disable this channel.
+ *
```
jwhui, line 956: nit: document return values

jwhui (approved):

PR #5892, opened 2020-12-01
[thread-cert] fix 5.3.3 random fail by small chance

simonlingoogle:
5.3.3 pktverify fail by small chance when the ping reply is forwarded from `Router3 -> Leader -> Router2` instead of  `Router3 -> Router2`. In such case, the direct `wpan.dst16` is not Router2's RLOC16, but Leader's RLOC16. 

This commit change 5.3.3 to check for both `6lowpan mesh dst16` and `wpan.dst16` against Router2's RLOC16.

### DIFF
--- a/tests/scripts/thread-cert/Cert_5_3_03_AddressQuery.py
+++ b/tests/scripts/thread-cert/Cert_5_3_03_AddressQuery.py
@@ -231,7 +231,7 @@ class Cert_5_3_3_AddressQuery(thread_cert.TestCase):
             must_next()
         pkts.filter_ping_reply(identifier=_pkt.icmpv6.echo.identifier).\
             filter_wpan_src64(ROUTER_3).\
-            filter_wpan_dst16(ROUTER_2_RLOC16).\
+            filter_dst16(ROUTER_2_RLOC16).\
             must_next()
         pkts.filter_ping_reply(identifier=_pkt.icmpv6.echo.identifier).\
             filter_wpan_src64(ROUTER_2).\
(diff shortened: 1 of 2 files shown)


librasungirl:
There is no link quality configuration in this test, generally in a stable topology, it should always choose the lowest cost route, that is, generally the response route should be ROUTER_3->ROUTER_2.

My guess is 10s time might not be enough for the routers to sync link/routes completely, specifically the link between ROUTER_3 and ROUTER_2 is not established.
https://github.com/openthread/openthread/blob/b4120128433571741fcd5da73255e2a138f7b8cb/tests/scripts/thread-cert/Cert_5_3_03_AddressQuery.py#L110-L114

One alternative option might be to extend the wait time to fix the random failing.

simonlingoogle:
> My guess is 10s time might not be enough for the routers to sync link/routes completely, specifically the link between ROUTER_3 and ROUTER_2 is not established.

Agree.

> One alternative option might be to extend the wait time to fix the random failing.

Yes. My consideration is that the behavior (i.e. ping Router3 -> Leader -> Router2) itself is in line with the Test Spec, thus should be allowed technically.

librasungirl:
> Yes. My consideration is that the behavior (i.e. ping Router3 -> Leader -> Router2) itself is in line with the Test Spec, thus should be allowed technically.

I would agree it doesn't violate the test plan which only validate the DUT's forwarding functionality.
Either option looks fine to me.

librasungirl (approved):

jwhui (approved):

PR #5899, opened 2020-12-01
[github-actions] add dependent PR support

### LINKED ISSUE #5896: [github-actions] add PR dependency for CI
**Is your feature request related to a problem? Please describe.** A clear and concise description of what the problem is.
There are mutual dependency relationship between PRs of OpenThread and and several derived projects. I think we can add a dependency tag in PR's description so that it can automatically apply dependent changes when running CI tests.

**Describe the solution you'd like** A clear and concise description of what you want to happen.
provide a script to unify downloading OpenThread project source code, which can detect the label `Depends-On:` and apply the PRs. Examples:

`Depends-On: openthread/ot-ns#123`
`Depends-On: openthread/ot-br-posix#345`

On dependency per line.

**Describe alternatives you've considered** A clear and concise description of any alternative solutions or features you've considered.
Hard code seems not scalable.

**Additional context** Add any other context or screenshots about the feature request here.
https://github.com/openthread/openthread/pull/5888

bukepo:
fixes #5896

Depends-On: openthread/ot-ns#112

### DIFF
--- a/.github/workflows/simulation-1.2.yml
+++ b/.github/workflows/simulation-1.2.yml
@@ -283,6 +283,8 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Build OTBR Docker
+      env:
+        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
       run: |
         ./script/test build_otbr_docker
     - name: Bootstrap
--- a/.github/workflows/simulation.yml
+++ b/.github/workflows/simulation.yml
@@ -89,16 +89,13 @@ jobs:
         ./bootstrap
     - name: Run OTNS Tests
       env:
-        OTNS_COMMIT: "2d57478d9e748d09cad65cf5374942e9d96719ed" # openthread/ot-ns#112
+        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
       run: |
         export OT_DIR=$PWD
         mkdir -p /tmp/otns
         (
+          ./script/git-tool clone --depth 1 https://github.com/openthread/ot-ns.git /tmp/otns
           cd /tmp/otns
-          git init
-          git remote add origin https://github.com/openthread/ot-ns.git
-          git fetch origin "${OTNS_COMMIT:-master}" --depth 1
-          git reset --hard FETCH_HEAD
           ./script/test py-unittests py-examples
         )
     - uses: actions/upload-artifact@v2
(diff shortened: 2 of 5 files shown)


script/git-clone
```diff
@@ -0,0 +1,102 @@
...
+}
+
+try_git_clone()
+{
+    local dest_dir
+
+    dest_dir="$(git clone "$@" 2>&1 | tee | cut -d\' -f2)"
+
+    [[ ${GITHUB_ACTIONS+x} && ${GITHUB_REF-} =~ ^refs/pull/[0-9]+/merge ]] || {
+        echo 'Not a pull request.'
```
simonlingoogle, line 71: We don't need this in master (merged) ?

script/git-clone
```diff
@@ -0,0 +1,102 @@
...
+set -euxo pipefail
+
+do_check()
+{
+    local project_name
+
+    project_name=$(git remote get-url origin | grep -oE '[^/:]+/[^/:]+\.git$' | cut -d. -f1)
+    echo "project name: ${project_name}"
+    depends_on_pr="$(grep -E "^Depends-On: *${project_name}" | cut -d# -f2)" || {
+        echo "No dependency on ${project_name} found."
```
simonlingoogle, line 42: We should checkout master branch if no `Depends-On` found?

simonlingoogle (commented):
This is a amazing feature.

script/git-clone
bukepo: The goal of this feature is to allow a project's CI apply dependency PRs, so the process would be merge a PR and its dependencies all together.

simonlingoogle: We are going to fail when merged to master. 

Meanwhile the current way of using a specific commit ID, even though a little bit cumbersome, works for both PR and branch, without any check fails.

bukepo: I don't think it a good idea to use a commit id which doesn't exist in master branch and having to change the workflow file from time to time. For PRs in OpenThread, it should be merged together with its dependency PRs, so I don't think the CI would fail in OpenThread project. The dependency project will fail only if it tracks OpenThread with submodule in its CI system, which I think we can change to use master branch to prevent such failures. Thoughts?

simonlingoogle: Agree that it's a better solution than using fixed commit IDs. 

Again, we should checkout master branch when merged to master, right?

simonlingoogle (approved):
LGTM :+1:

Irving-cl (approved):
LGTM :+1:

abtink (approved):
Looks great. Thanks. This will be very useful 👍

jwhui (approved):

PR #5900, opened 2020-12-01
[message] Fix Message::Settings::GetDefault()

lmnotran:
`kDefault` was not being initialized properly and remains zero-initialized upon first-use.

image

This change initializes kDefault as a singleton at first-use, which fixes the issue


lmnotran:
Closing this as it's a workaround for https://github.com/openthread/ot-br-posix/issues/559. The true fix is in openthread#5905

PR #5901, opened 2020-12-02
[build] fix warning when building simulation with autotools

Irving-cl:
When building simulation with autotools, we got warnings like:
```
Making all in utils                                                                                                                                                                                                                                                                   
  CC       libopenthread_platform_utils_a-debug_uart.o                                                                                                                                                                                                                                
  CC       libopenthread_platform_utils_a-logging_rtt.o                                                                                                                                                                                                                               
  CXX      libopenthread_platform_utils_a-mac_frame.o                                                                                                                                                                                                                                 
  CC       libopenthread_platform_utils_a-soft_source_match_table.o                                                                                                                                                                                                                   
  CC       libopenthread_platform_utils_a-settings_ram.o                                                                                                                                                                                                                              
  CXX      libopenthread_platform_utils_a-otns_utils.o                                                                                                                                                                                                                                
In file included from /home/ot/openthread/examples/../src/core/openthread-core-config.h:72:0,                                                                                                                                                                                         
                 from /home/ot/openthread/examples/../examples/platforms/utils/debug_uart.c:28:                                                                                                                                                                                       
/home/ot/openthread/examples/../src/core/config/mac.h:335:69: warning: "OPENTHREAD_FTD" is not defined, evaluates to 0 [-Wundef]                                                                                                                                                      
     (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2) && (OPENTHREAD_FTD || OPENTHREAD_RADIO)                                                                                                                                                                              
                                                                     ^                                                                                                                                                                                                                
/home/ot/openthread/examples/../src/core/config/openthread-core-config-check.h:505:5: note: in expansion of macro 'OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE'                                                                                                                      
     ^                                                                                                                                                                                                                                                                                
In file included from /home/ot/openthread/examples/../src/core/openthread-core-config.h:72:0,                                                                                                                                                                                         
                 from /home/ot/openthread/examples/../examples/platforms/utils/settings.h:38,                                                                                                                                                                                         
                 from /home/ot/openthread/examples/../examples/platforms/utils/settings_ram.c:35:                                                                                                                                                                                     
/home/ot/openthread/examples/../src/core/config/mac.h:335:69: warning: "OPENTHREAD_FTD" is not defined, evaluates to 0 [-Wundef]                                                                                                                                                      
     (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2) && (OPENTHREAD_FTD || OPENTHREAD_RADIO)                                                                                                                                                                              
                                                                     ^                                                                     
/home/ot/openthread/examples/../src/core/config/openthread-core-config-check.h:505:5: note: in expansion of macro 'OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE'
     ^                                                               
In file included from /home/ot/openthread/examples/../src/core/openthread-core-config.h:72:0,                                                                                                                                                                                         
                 from /home/ot/openthread/examples/../examples/platforms/utils/soft_source_match_table.h:39,                                                                                                                                                                          
                 from /home/ot/openthread/examples/../examples/platforms/utils/soft_source_match_table.c:37:                                                                                                                                                                          
/home/ot/openthread/examples/../src/core/config/mac.h:335:69: warning: "OPENTHREAD_FTD" is not defined, evaluates to 0 [-Wundef]                                                                                                                                                      
     (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2) && (OPENTHREAD_FTD || OPENTHREAD_RADIO)                                             
```
The cause is that most components (like `utils`, `hdlc`, `spinel`, `cli`, etc) depend on `openthread-core-config` but they are not built with the `FTD/MTD/RADIO` flags. However many options in `openthread-core-config` depend on `FTD/MTD/RADIO`. I think the simplest solution is to add default value of `FTD/MTD/RADIO` in `openthread-core-config`. They would be evaluated to `0` anyway if not defined.


simonlingoogle:
It's a potential risk that we are using wrong definition for FTD, MTD or RADIO. 

Just curious if we can just remove `#include <openthread-core-config.h>` from these files.

Irving-cl:
> It's a potential risk that we are using wrong definition for FTD, MTD or RADIO.

I'm not clear about the risk you mentioned. These lines won't change the result of `FTD`, `MTD` or `RADIO`. Without these lines, they would still be evaluated to `0` anyway if not defined. Could you provide more details?

simonlingoogle:
> > It's a potential risk that we are using wrong definition for FTD, MTD or RADIO.
> 
> I'm not clear about the risk you mentioned. These lines won't change the result of `FTD`, `MTD` or `RADIO`. Without these lines, they would still be evaluated to `0` anyway if not defined. Could you provide more details?

The risk is not introduced by this PR, but existed originally. 
Take this `openthread-platform-utils` library for example, the same binary is linked with both `openthread-ftd` and `openthread-mtd`. So, the code of `openthread-platform-utils` should never depend on any configuration that further depend on `OPENTHREAD_FTD` or `OPENTHREAD_MTD`. From such perspective, these source files should not include these configurations in the first place. 
But I am not sure if that's totally doable.

Moandor-y:
> It's a potential risk that we are using wrong definition for FTD, MTD or RADIO.

In the development of <https://github.com/openthread/openthread/pull/5694> I found that some codes are compiled without correct definitions of these. That's why I added these lines: <https://github.com/openthread/openthread/pull/5694/files#diff-99156a76a546b36b1c24728a473b05a0032e90ced9579e795db35d9ad1746403R103-R104>

There could be more of such errors. Though I don't have an idea of how to find them.

librasungirl (approved):

Irving-cl:
I will find a better solution.

simonlingoogle:
> > It's a potential risk that we are using wrong definition for FTD, MTD or RADIO.
> 
> In the development of #5694 I found that some codes are compiled without correct definitions of these. That's why I added these lines: https://github.com/openthread/openthread/pull/5694/files#diff-99156a76a546b36b1c24728a473b05a0032e90ced9579e795db35d9ad1746403R103-R104
> 
> There could be more of such errors. Though I don't have an idea of how to find them.

I think we need to resolve these compile warnings to find these issues. That's why I think it's a potential risk to hide all these warnings. 

Your fix at https://github.com/openthread/openthread/pull/5694/files#diff-99156a76a546b36b1c24728a473b05a0032e90ced9579e795db35d9ad1746403R103-R104 is suitable for posix since posix platform is always FTD. But providing default definitions may not be proper for other platforms.

jwhui:
I agree that we should not try to mask these issues.

Irving-cl:
After re-thinking on this, I think we are actually facing two problems:

### 1. Warnings
Warnings for undefined `OPENTHREAD_FTD`, `OPENTHREAD_MTD`, `OPENTHREAD_RADIO`. The cause is that `openthread-core-config.h` is referenced directly or indirectly by all components however not all components are built with `FTD/MTD/RADIO` flags. For example, `openthread-core` are built with such flags while things like `mbedtls`, `spinel`, platform-lib are not.

> Just curious if we can just remove #include <openthread-core-config.h> from these files

I think we cannot and shouldn't do this. Some components reference the config indirectly (unintentionally) although they don't depend on core options:
```
In file included from /home/openthread/examples/../src/core/openthread-core-config.h:73:0,
                 from /home/openthread/examples/../src/core/common/debug.hpp:37,       
                 from /home/openthread/examples/../src/lib/hdlc/hdlc.hpp:43,                                                 
                 from /home/openthread/examples/../src/lib/hdlc/hdlc.cpp:33:                            
/home/openthread/examples/../src/core/config/mac.h:335:69: warning: "OPENTHREAD_FTD" is not defined, evaluates to 0 [-Wundef]               
     (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2) && (OPENTHREAD_FTD || OPENTHREAD_RADIO) 
```
And some components depend on the options to compile their own code. For example, in `mbedtls`:
```
#if OPENTHREAD_CONFIG_BORDER_AGENT_ENABLE || OPENTHREAD_CONFIG_COMMISSIONER_ENABLE || OPENTHREAD_CONFIG_COAP_SECURE_API_ENABLE
#define MBEDTLS_SSL_COOKIE_C 
#define MBEDTLS_SSL_SRV_C
#endif   
```
If some core features are not enabled, then we may reduce the corresponding code that are only used to support that feature in the lib. I think it would be a waste if we always build the libs with full features.

So I think it's reasonable that other components reference `openthread-core-config.h`. The problem is that we didn't specify `FTD/MTD/RADIO` for all components. `FTD/MTD/RADIO` options are kind of special. Unlike other feature options which would be at least set to a default value in `openthread-core-config.h`, `FTD/MTD/RADIO` are only set in build system. So some components don't have a default value for it because `FTD/MTD/RADIO` are not set in build system.
#### Solution
The simplest way is just as the PR currently do. We add default value for `FTD/MTD/RADIO` so that all components have at least a default value for `FTD/MTD/RADIO`.

> In the development of #5694 I found that some codes are compiled without correct definitions of these. That's why I added these lines: https://github.com/openthread/openthread/pull/5694/files#diff-99156a76a546b36b1c24728a473b05a0032e90ced9579e795db35d9ad1746403R103-R104

For issues like this, we need to specify `FTD/MTD/RADIO` flags in build system for such components and that's the only solution. Note that the default values are just to eliminate the warnings. If a component need `FTD/MTD/RADIO` flags, we have to specify explicitly. As a conclusion, we need to specify `FTD/MTD/RADIO` flags in build system for those components and add default value definition for `FTD/MTD/RADIO` to ensure all components have a default value for the flags.

### 2. FTD/MTD/RADIO version of libraries

> Take this openthread-platform-utils library for example, the same binary is linked with both openthread-ftd and openthread-mtd. So, the code of openthread-platform-utils should never depend on any configuration that further depend on OPENTHREAD_FTD or OPENTHREAD_MTD. From such perspective, these source files should not include these configurations in the first place.

That's true. And that's the most tricky part. As I said, I think it's reasonable for other libs to reference `openthread-core-config.h`. But it's inevitable that some options depend on `FTD/MTD/RADIO`. And some libs do depend on `FTD/MTD/RADIO` related options. For example, `CSL_RECEIVER` is used in platform lib (simulation). We tried to only enable it for `MTD`. However we cannot do that because `FTD` and `MTD` uses the same lib. This is not a big problem because both `FTD` and `MTD` can have this feature. We just keep it as it is for now. But if we have a feature that MUST be enabled on `FTD` and MUST NOT be enabled on `MTD` (like `CSL_TRANSMITTER`, but `CSL_TRANSMITTER` is not used outside `openthread-core`), then we are helpless. The only solution is to build different version of libs (ftd,mtd,radio) and link them correspondingly.

As a summary, we can reluctantly work to build same libs from those components for now. We might need to build different libs in the future.

### Possible enhancement
Maybe we can divide `openthread-core-config` into two parts: **device type related** and **non-related**. And we try to include the `non-related` part from other components as possible. This may help for most components to get rid of `FTD/MTD/RADIO`. For those components who just depend on `FTD/MTD/RADIO` related options, we use workaround or add different version libs for them if it's necessary.

Thoughts? @simonlingoogle @jwhui @Moandor-y

simonlingoogle:
@Irving-cl Thanks for the detailed analysis. 

My preference would be a combination of these means:
1. removing unnessary `#include <openthread-core-config.h>`
2. Change device type related in a way shown below:
```c++
#if defined(OPENTHREAD_FTD) && defined(OPENTHREAD_RADIO)
#define OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE \
    (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2) && (OPENTHREAD_FTD || OPENTHREAD_RADIO)
#endif  // defined(OPENTHREAD_FTD) && defined(OPENTHREAD_RADIO)
```
This can elide the warning. And if the library does not depend on `OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE`, we are good. If the library does depend on `OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE`, we will have another warning which indicate that we really need to build different versions of this library. 

3. Use version of libraries when the two means above does not work. 

Thoughts? @Irving-cl

Irving-cl:
@simonlingoogle Sounds good to me, thanks!

simonlingoogle:
> @simonlingoogle Sounds good to me, thanks!

BTW, I think it's better if we can make sure there are no more such warnings using CI checks (after we fix them all).

src/core/config/openthread-core-config-check.h
```diff
@@ -555,4 +557,16 @@
 #error "OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE conflicts with OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT."
 #endif
 
+#if defined(OPENTHREAD_FTD)
```
simonlingoogle, line 560:
```suggestion
#if defined(OPENTHREAD_CONFIG_TMF_PROXY_DUA_ENABLE)
```

src/core/config/openthread-core-config-check.h
```diff
@@ -555,4 +557,16 @@
...
+#if OPENTHREAD_CONFIG_TMF_PROXY_DUA_ENABLE && OPENTHREAD_CONFIG_THREAD_VERSION < OT_THREAD_VERSION_1_2
+#error "Thread 1.2 or higher version is required for OPENTHREAD_CONFIG_TMF_PROXY_DUA_ENABLE"
+#endif
+#endif
+
+#if defined(OPENTHREAD_FTD)
```
simonlingoogle, line 566:
```suggestion
#if defined(OPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE)
```

src/core/config/device_type_util.h
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * @file
+ *   This file includes OpenThread device type definition for utilization.
+ */
+
+#ifndef CONFIG_DEVICE_TYPE_UTIL_H_
+#define CONFIG_DEVICE_TYPE_UTIL_H_
+
+#if defined(OPENTHREAD_FTD) || defined(OPENTRHEAD_MTD) || defined(OPENTHREAD_RADIO)
+#define OPENTHREAD_DEVICE_TYPE 1
```
simonlingoogle, line 38:
```suggestion
#define _OPENTHREAD_DEVICE_TYPE_DEFINED 1
```

Start with `_` since this is not for users to use?

src/core/config/device_type_util.h
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * @file
+ *   This file includes OpenThread device type definition for utilization.
+ */
+
+#ifndef CONFIG_DEVICE_TYPE_UTIL_H_
+#define CONFIG_DEVICE_TYPE_UTIL_H_
+
+#if defined(OPENTHREAD_FTD) || defined(OPENTRHEAD_MTD) || defined(OPENTHREAD_RADIO)
```
simonlingoogle, line 37: Add check to make sure that only one is defined as 1, and others are defined as 0 or undefined ?

After this check pass, we can define any undefined macro to be 0 to make sure all three macros are defined properly whenever `OPENTHREAD_DEVICE_TYPE=1` so that we don't need extra macros like \_OPENTHREAD_FTD\_ ?

src/core/config/device_type_util.h
```diff
@@ -0,0 +1,61 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes OpenThread device type definition for utilization.
+ */
+
+#ifndef CONFIG_DEVICE_TYPE_UTIL_H_
+#define CONFIG_DEVICE_TYPE_UTIL_H_
```
simonlingoogle, line 35: rename to `device_type_check.h` ?

src/core/config/device_type_util.h
Irving-cl: I replaced  occurrences of `_OPENTHREAD_FTD_` in code. I kept it here in this file to do the check (only 1 is defined and is set to true) with more convenience.

src/core/config/device_type_check.h
```diff
@@ -0,0 +1,78 @@
...
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes OpenThread device type definition for utilization.
+ */
+
+#ifndef CONFIG_DEVICE_TYPE_UTIL_H_
```
simonlingoogle, line 34:
```suggestion
#ifndef CONFIG_DEVICE_TYPE_CHECK_H_
```

src/core/config/device_type_check.h
```diff
@@ -0,0 +1,78 @@
...
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes OpenThread device type definition for utilization.
```
simonlingoogle, line 31:
```suggestion
 *   This file includes OpenThread device type checking utilities.
```
nit

src/core/openthread-core-config.h
```diff
@@ -61,6 +61,7 @@
 #include "config/coap.h"
 #include "config/commissioner.h"
 #include "config/dataset_updater.h"
+#include "config/device_type_check.h"
```
simonlingoogle, line 64: Should move this file before other config headers?

src/core/config/openthread-core-config-check.h
```diff
@@ -34,6 +34,8 @@
 #ifndef OPENTHREAD_CORE_CONFIG_CHECK_H_
 #define OPENTHREAD_CORE_CONFIG_CHECK_H_
 
+#include "config/device_type_check.h"
```
simonlingoogle, line 37: is it necessary?

src/core/openthread-core-config.h
simonlingoogle: move it even before `#include OPENTHREAD_PROJECT_CORE_CONFIG_FILE` ?
I think we should detect device as early as possible.

src/core/config/openthread-core-config-check.h
Irving-cl: Yes, I used `_OPENTHREAD_DEVICE_TYPE_DEFINED` in this file.

src/core/config/tmf.h
```diff
@@ -35,6 +35,8 @@
 #ifndef CONFIG_TMF_H_
 #define CONFIG_TMF_H_
 
+#include "config/device_type_check.h"
```
simonlingoogle, line 38: Is it still necessary after we have included "config/device_type_check.h" at the beginning of `openthread-core-config.h`?

src/core/config/openthread-core-config-check.h
simonlingoogle, line 37: Is it still necessary after we have included "config/device_type_check.h" at the beginning of `openthread-core-config.h`

src/core/config/mac.h
```diff
@@ -35,6 +35,7 @@
 #ifndef CONFIG_MAC_H_
 #define CONFIG_MAC_H_
 
+#include "config/device_type_check.h"
```
simonlingoogle, line 38: Is it still necessary after we have included "config/device_type_check.h" at the beginning of `openthread-core-config.h`

src/core/config/openthread-core-config-check.h
simonlingoogle: Is it still necessary after we have included "config/device_type_check.h" at the beginning of `openthread-core-config.h`

Irving-cl:
A summary for the new approach to handle `FTD/MTD/RADIO` in this PR:

Basically it's trying to establish a guideline for using `FTD/MTD/RADIO` flags in our code. (As a bonus, it addresses the compile warning as well).
1. We allow some components to be compiled with `FTD/MTD/RADIO` flags and some components to be built without `FTD/MTD/RADIO` (Though they include the core config file).
2. `FTD/MTD/RADIO` flags specified for a component MUST be valid. That is, if and only if one of them is set to `1`. For example, `OPENTHREAD_FTD=1 OPENTHREAD_MTD=0 OPENTHREAD_RADIO=0` is valid. `OPENTHREAD_FTD=1` is valid. `OPENTHREAD_FTD=1 OPENTHREAD_MTD=1 OPENTHREAD_RADIO=0` is invalid. `OPENTHREAD_FTD=0 OPENTHREAD_MTD=0 OPENTHREAD_RADIO=0` is invalid (If a component want this, it can just leave this flag unset).
3. If `FTD/MTD/RADIO` flags are specified for a component, then `OPENTHREAD_FTD`, `OPENTHREAD_MTD`, `OPENTHREAD_RADIO` will all be defined. One has value `1` and the other two has value `0`. This would always be true because we would do the check in (2.). 
4. If `FTD/MTD/RADIO` flags are not specified for a component, then `OPENTHREAD_FTD`, `OPENTHREAD_MTD`, `OPENTHREAD_RADIO` would be undefined in code.
5. Some feature flags depend on `FTD/MTD/RADIO` flags and some components depend on those feature flags.  We wrap the default definitions and checks of these feature flags in our core config file by a new macro `_OPENTHREAD_DEVICE_TYPE_DEFINED`. This macro would be `true` if valid `FTD/MTD/RADIO` flags are specified for the component. In this way, those components (`src/lib`, `mbedtls`, etc) which doesn't depend on feature flags depending on `FTD/MTD/RADIO` wouldn't be bothered by `FTD/MTD/RADIO` anymore. (Those components don't have `FTD/MTD/RADIO` but will include core config file).
6. If a component does depend on feature flags depending on `FTD/MTD/RADIO`, then it should specify `FTD/MTD/RADIO` in the build system. If it doesn't, since we wrap that feature with `_OPENTHREAD_DEVICE_TYPE_DEFINED` (mentioned in 5.), that feature will be undefined and we would get a warning. And for these component, we might build different versions of libs in the future.

What do you think of this? @abtink @bukepo

simonlingoogle (approved):
LGTM :+1: 

Thanks. Great to see these warnings being resolved.
I would suggest to avoid such warnings in the future using CI checks.

bukepo:
Can this issue be solved by removing OPENTHREAD_FTD/MTD/RADIO from core config? The platform code currently doesn't distinguish device types, and device type is not a core configuration, so it should not appear in the core configuration files.

Irving-cl:
> Can this issue be solved by removing OPENTHREAD_FTD/MTD/RADIO from core config? The platform code currently doesn't distinguish device types, and device type is not a core configuration, so it should not appear in the core configuration files.

I think there are some difficulties. For example, we have `CSL_TRANSMITTER` option which should only be enabled on `FTD` and shouldn't be enabled on `MTD`. Currently we did this by defining `CSL_TRANSMITTER` in core config file with `FTD/MTD`:
```
/**
 *
 * This setting configures the CSL transmitter feature in Thread 1.2.
 * This is compulsory for 1.2 FTD.
 *
 */
#define OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE \
    (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2) && (OPENTHREAD_FTD || OPENTHREAD_RADIO)
```
Without this, we have to specify these device-type-dependent options in build system. (Otherwise, FTD and MTD will always have the same option.)

Thoughts? @bukepo

bukepo:
On one hand, turning feature flag on doesn't mean the feature will be available on all device types. For example, most upper MAC layer features don't take effect on RCP build(commissioner, border agent, border router and so on).
On the other hand, CSL transmitter can be enabled on non-FTD device as well, which is the base of the SSED-SSED link mentioned in 1.2 spec, right?

Irving-cl:
> On one hand, turning feature flag on doesn't mean the feature will be available on all device types. For example, most upper MAC layer features don't take effect on RCP build(commissioner, border agent, border router and so on).
> On the other hand, CSL transmitter can be enabled on non-FTD device as well, which is the base of the SSED-SSED link mentioned in 1.2 spec, right?

Sounds good to me. Then let's:
1. Remove all `FTD/MTD/RADIO` from core config file.
2. When wrapping code with feature macro (those depending on `FTD/MTD/RADIO`), we use `OPENTHREAD_FTD && XXX`. For example, `#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE`.

Irving-cl:
I would start another PR.

PR #5902, opened 2020-12-02
[udp6] initialize Socket variables in its constructor

chshu:
It failed to start CoAP socket due to the `mSocket.IsBound()` check ([link](https://github.com/openthread/openthread/blob/master/src/core/coap/coap.cpp#L1003)), the root cause is that the Socket variables were not initialized properly.

### DIFF
--- a/src/core/net/udp6.cpp
+++ b/src/core/net/udp6.cpp
@@ -74,7 +74,7 @@ exit:
 Udp::Socket::Socket(Instance &aInstance)
     : InstanceLocator(aInstance)
 {
-    mHandle = nullptr;
+    Clear();
 }
 
 Message *Udp::Socket::NewMessage(uint16_t aReserved, const Message::Settings &aSettings)
--- a/src/core/net/udp6.hpp
+++ b/src/core/net/udp6.hpp
@@ -39,6 +39,7 @@
 #include <openthread/udp.h>
 #include <openthread/platform/udp.h>
 
+#include "common/clearable.hpp"
 #include "common/linked_list.hpp"
 #include "common/locator.hpp"
 #include "common/non_copyable.hpp"
@@ -70,7 +71,7 @@ public:
      * This class implements a UDP/IPv6 socket.
      *
      */
-    class SocketHandle : public otUdpSocket, public LinkedListEntry<SocketHandle>
+    class SocketHandle : public otUdpSocket, public LinkedListEntry<SocketHandle>, public Clearable<SocketHandle>
     {
         friend class Udp;
         friend class LinkedList<SocketHandle>;


chshu:
@googlebot I signed it!

librasungirl (approved):

wgtdkp (approved):
LGTM :+1:

src/core/net/udp6.cpp
```diff
@@ -74,7 +74,12 @@ bool Udp::SocketHandle::Matches(const MessageInfo &aMessageInfo) const
 Udp::Socket::Socket(Instance &aInstance)
     : InstanceLocator(aInstance)
 {
-    mHandle = nullptr;
+    memset(&mSockName, 0, sizeof(mSockName));
+    memset(&mPeerName, 0, sizeof(mPeerName));
+    mHandler = nullptr;
+    mContext = nullptr;
+    mHandle  = nullptr;
+    mNext    = nullptr;
```
simonlingoogle, line 82:
```suggestion
    Clear();
```
Let `Socket` inherit from `Clearable` ?

abtink: I don't think we can have `Socket` be `Clearable` sicne it is `InstanceLocator`.
`Clear()` would `memset` all the fields to zero which we should not in this case (should not write over the instance ref/pointer).

src/core/net/udp6.cpp
```diff
@@ -74,7 +74,12 @@ bool Udp::SocketHandle::Matches(const MessageInfo &aMessageInfo) const
 Udp::Socket::Socket(Instance &aInstance)
     : InstanceLocator(aInstance)
 {
-    mHandle = nullptr;
+    memset(&mSockName, 0, sizeof(mSockName));
+    memset(&mPeerName, 0, sizeof(mPeerName));
```
abtink, line 78:
```suggestion
    GetSockName().Clear();
    GetPeerName().Clear();
```

src/core/net/udp6.cpp
abtink: Another approach here: 
- We can have `SocketHandle` be a `Clearable` (handle just contains the fields and mirrors the public OT `otUdpSocket` struct). 
- Then `Socket` would inherit the `Clear()` method from `SocketHandle` and can use it in its own constructor (which won't touch any fields potentially added by `InstanceLocator`).

abtink (approved):
LGTM. Thanks. 
Some suggestions below

src/core/net/udp6.cpp
chshu: Done, thanks.

abtink (approved):
LGTM. Thanks. 👍

simonlingoogle (approved):
LGTM :+1:

bukepo (approved):
LGTM:+1:

jwhui (approved):
Thanks, @chshu !

PR #5904, opened 2020-12-02
[thread-cert] add case 8.3.1 CommissionerPetition using pktverify

FiveDimensions:
correct wrong COMMISSIONING_DATA_TLV const in case 5.6.6 and 5.6.9 in this PR


tests/scripts/thread-cert/Cert_8_3_01_CommissionerPetition.py
```diff
@@ -0,0 +1,366 @@
...
+                              NM_COMMISSIONER_SESSION_ID_TLV,
+                              NM_STEERING_DATA_TLV
+                             } == set(p.thread_meshcop.tlv.type) and\
+                   p.mle.tlv.leader_data.data_version ==
+                   (_dr_pkt.mle.tlv.leader_data.data_version + 1) % 256 and\
+                   p.thread_nwd.tlv.stable == [0]
+                   ).\
+            must_next()
+
+        # Step 8: Commissioner sends a sends a resign request via commissioner keep-alive request
```
librasungirl, line 271:
```suggestion
        # Step 8: Commissioner sends a resign request via commissioner keep-alive request
```

tests/scripts/thread-cert/Cert_8_3_01_CommissionerPetition.py
```diff
@@ -0,0 +1,366 @@
...
+           must_next()
+
+        # Step 11: Leader accepts Commissioner to the network, sends a Leader Petition
+        #          Response (LEAD_PET.rsp) to Commissioner:
+        #          CoAP Response Code
+        #              2.04 Changed
+        #          CoAP Payload
+        #              State TLV (value = Accept)
+        #              Commissioner ID TLV (optional)
+        #              Commissioner Session ID TLV (contains higher Session ID number
```
librasungirl, line 352: Here the description in Test Plan might not be appropriate -these two are marked as optional in Thread Specification and only appear when accept, and skip when reject.
May file a ticket to Thread Group?

Additionally, session id number seems not checked

FiveDimensions: sure

tests/scripts/thread-cert/Cert_8_3_01_CommissionerPetition.py
```diff
@@ -357,7 +357,9 @@ def verify(self, pv):
                               NM_STATE_TLV,
                               NM_COMMISSIONER_SESSION_ID_TLV
                               } <= set(p.coap.tlv.type) and\
-                   p.thread_meshcop.tlv.state == MESHCOP_ACCEPT
+                   p.thread_meshcop.tlv.state == MESHCOP_ACCEPT and\
+                   (p.thread_meshcop.tlv.commissioner_sess_id -
+                   _dr_pkt3.thread_meshcop.tlv.commissioner_sess_id) % 256 <= 127
```
librasungirl, line 364: One note is that Session ID is of type unit16_t with random initialization, and increment (+1) along per new active commissioner.
https://github.com/openthread/openthread/blob/9380ffd07c39d7a2cfe9bca8cf2c7649873e6da2/src/core/meshcop/meshcop_leader.cpp#L99

librasungirl (approved):

PR #5908, opened 2020-12-03
[diag] check state before entering diagnostic mode

bukepo:
This commit restrict diagnostic mode only when network stack is not up,
which prevents some issues caused by switching to diag mode after the
device was attached. For example, timers may not be triggered properly
after returnning from diagnostic mode.

### DIFF
--- a/tests/scripts/expect/cli-misc.exp
+++ b/tests/scripts/expect/cli-misc.exp
@@ -164,4 +164,7 @@ expect "0:0:0:0::/64 s low"
 send "route remove ::/64\n"
 expect "Done"
 
+send "diag start\n"
+expect ": InvalidState"
+
 dispose_all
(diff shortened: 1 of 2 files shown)


tests/scripts/expect/cli-misc.exp
```diff
@@ -164,4 +164,7 @@ expect "0:0:0:0::/64 s low"
 send "route remove ::/64\n"
 expect "Done"
 
+send "diag start\n"
+expect ": InvalidState"
```
simonlingoogle, line 168:
```suggestion
expect "Error 13: InvalidState"
```
nit

simonlingoogle (approved):
LGTM :+1: 
with a nit

tests/scripts/expect/cli-misc.exp
bukepo: I intended to omit the error number in case it changed.

jwhui (approved):

PR #5910, opened 2020-12-03
[tmf] TMF not to use platform udp

simonlingoogle:
This commit fixes TMF issues related to the Host tun interface.

- [x] Do not pass TMF messages to Host tun interface
  - [x] Backbone TMF (also using port 61631) keeps passing TMF messages to Host tun interface
- [x] Re-enable packet filters for ALOC/RLOC destined and TMF packets  for PLAT_NETIF and PLAT_UDP

### DIFF
--- a/script/check-posix-pty
+++ b/script/check-posix-pty
@@ -129,6 +129,12 @@ send "ipaddr\r\n"
 expect "Done"
 send "coex\r\n"
 expect "Done"
+send "coap start\r\n"
+expect "Done"
+send "coap resource TestResource\r\n"
+expect "Done"
+send "coap set TestContent\r\n"
+expect "Done"
 wait
 EOF
 
@@ -142,7 +148,7 @@ EOF
         fi
     done
 
-    netstat -an | grep -q 61631 || die 'TMF port is not available!'
+    netstat -an | grep -q 5683 || die 'Application CoAP port is not available!'
 
     extaddr=$(grep -aoE '[0-9a-z]{16}' $OT_OUTPUT)
     echo "Extended address is: ${extaddr}"
@@ -158,14 +164,14 @@ EOF
 
     LEADER_ALOC=fdde:ad00:beef::ff:fe00:fc00
     # Retrievie extended address through network diagnostic get
-    coap_response=$(echo -n '120100' | xxd -r -p | coap-client -m POST coap://[${LEADER_ALOC}]:61631/d/dg -f- | xxd -p | grep 0008)
+    coap_response=$(coap-client -B 5 -m GET coap://[${LEADER_ALOC}]:5683/TestResource)
     echo "CoAP response is: ${coap_response}"
 
     # Verify CoAP response contains the extended address
-    if [[ ${coap_response} == *${extaddr}* ]]; then
+    if [[ ${coap_response} == *TestContent* ]]; then
         echo 'Success'
     else
-        die 'Failed to get extended address'
+        die 'Failed to access application CoAP'
     fi
 }
 
--- a/script/check-posix-pty
+++ b/script/check-posix-pty
@@ -163,11 +163,11 @@ EOF
     fi
 
     LEADER_ALOC=fdde:ad00:beef::ff:fe00:fc00
-    # Retrievie extended address through network diagnostic get
+    # Retrievie test resource through application CoAP
     coap_response=$(coap-client -B 5 -m GET coap://[${LEADER_ALOC}]:5683/TestResource)
     echo "CoAP response is: ${coap_response}"
 
-    # Verify CoAP response contains the extended address
+    # Verify CoAP response contains the test content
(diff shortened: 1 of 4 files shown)


bukepo (approved):
LGTM:+1:

librasungirl (approved):

simonlingoogle:
@bukepo @librasungirl 

Some new enhancements to this PR:
* Fixed the CoAP test in check-posix-pty 
* Fixed an issue that Backbone TMF was sent in the Thread network rather than on the Backbone link
* Restored some packet filters (which was disabled to so that TMF messages can be passed to Host)

Please take another look, thanks.

gjc13 (approved):

src/core/net/udp6.cpp
```diff
@@ -293,7 +309,7 @@ otError Udp::Connect(SocketHandle &aSocket, const SockAddr &aSockAddr)
     }
 
 #if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
-    if (!IsMlePort(aSocket.mSockName.mPort))
+    if (ShouldUsePlatformUdp(aSocket))
```
bukepo, line 312:
```suggestion
    if (aSocket.mHandle != nullptr)
```

simonlingoogle: Actually we always create platform socket in `Udp::Open` no matter it's MLE or TMF or other ports.

This behavior was so before this PR.

bukepo: Can we change it in this pr?

simonlingoogle: I suspect that we always creates OS socket because we don't know if platform udp is required in `Udp::Open` yet. 

If we only want to create OS socket when it's really necessary, we need to create OS socket later in `Udp::Bind`, in which we can determine whether or not to use platform UDP.

Thoughts?

bukepo: agree to move it to bind.

simonlingoogle: updated:
- creates OS socket in `Udp::Bind`
- Use `aSocket.mHandle` when appropriate

@bukepo

simonlingoogle: update:

**reverted the change**

the recent commit failed in multiple tests. Turnout it's more complex than we thought. `BindToNetif` might be called before `Bind` which would require the platform socket to be ready. 

I would prefer to refactor it (along with Backbone Router related code) in another PR since it's out of the scope of this PR.

bukepo (approved):

src/core/net/udp6.cpp
```diff
@@ -533,16 +548,31 @@ void Udp::HandlePayload(Message &aMessage, MessageInfo &aMessageInfo)
...
+
+bool Udp::ShouldUsePlatformUdp(const Udp::SocketHandle &aSocket) const
+{
+    uint16_t port       = aSocket.mSockName.mPort;
+    bool     usePlatUdp = false;
+
+    VerifyOrExit(!ShouldUsePlatformUdp(port), usePlatUdp = true);
+
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+    usePlatUdp = IsBackboneSocket(aSocket);
```
jwhui, line 569: nit: what about something like (similar to `ShouldUsePlatformUdp(uint16_t aPort)` above):
```c++
    return (ShouldUsePlatformUdp(port)
#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
            || IsBackboneSocket(aSocket)
#endif
    );
```

simonlingoogle: Yes. 
I changed to the current code to be more readable. But it seems actually less readable.

simonlingoogle: updated

script/check-posix-pty
```diff
@@ -158,14 +164,14 @@ EOF
 
     LEADER_ALOC=fdde:ad00:beef::ff:fe00:fc00
     # Retrievie extended address through network diagnostic get
-    coap_response=$(echo -n '120100' | xxd -r -p | coap-client -m POST coap://[${LEADER_ALOC}]:61631/d/dg -f- | xxd -p | grep 0008)
+    coap_response=$(coap-client -B 5 -m GET coap://[${LEADER_ALOC}]:5683/TestResource)
```
librasungirl, line 167: nit: not a diagnostic get test now, update the comment accordingly

script/check-posix-pty
```diff
@@ -158,14 +164,14 @@ EOF
...
     echo "CoAP response is: ${coap_response}"
 
     # Verify CoAP response contains the extended address
```
librasungirl, line 170: nit: update comment

librasungirl (approved):

jwhui (approved):

Issue #5909, opened 2020-12-03
[github-actions] Cert_5_3_06_RouterIdMask.py failed intermittently 


simonlingoogle:
**Describe the bug**
Cert_5_3_06_RouterIdMask.py failed intermittently 

**Console/log output**

[logs_42915.zip](https://github.com/openthread/openthread/files/5634268/logs_42915.zip)

[packet-verification-1.1-on-1.2-pcaps (2).zip](https://github.com/openthread/openthread/files/5634270/packet-verification-1.1-on-1.2-pcaps.2.zip)

FiveDimensions:
From the log, there is another advertisement with router costs set {1,0,1} in Step 5 which causes the sniff time difference value larger than 3 seconds, delete the second trickle time check in PR #5912

PR #5912, opened 2020-12-03
[thread-cert] fix case 5.3.6 issue #5909

FiveDimensions:


FiveDimensions:
delete the second trickle time check to fix issue #5909  since there may be another advertisement with router costs set {1,0,1} which will cause the sniff time difference value larger than 3 seconds

tests/scripts/thread-cert/Cert_5_3_06_RouterIdMask.py
```diff
@@ -197,7 +197,6 @@ def verify(self, pv):
             filter_LLANMA().\
             filter_mle_cmd(MLE_ADVERTISEMENT).\
             filter(lambda p: {1,2,1} == set(p.mle.tlv.route64.cost) and\
-                   p.sniff_timestamp - _pkt.sniff_timestamp <= 3 and\
```
librasungirl, line 200: One alternative option to ensure the timestamp check is to find two successive MLE advertisements with len(router id set) = 3, other than the strict cost check here.

Thoughts?

simonlingoogle: > delete the second trickle time check to fix issue #5909 since there may be another advertisement with router costs set {1,0,1} which will cause the sniff time difference value larger than 3 seconds

Not quite clear about the context. But shouldn't we find the one with least sniff_timestamp first?

FiveDimensions: the successive MLE advertisements already checking in line 182-190

librasungirl (approved):

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #5888, opened 2020-11-30
[style] simplify build paths

bukepo:
This commit simplifies the build paths:

* remove cmake prefix from `script/test`
* use fixed name for simulation and posix platforms just like others

Depends-On: openthread/ot-ns#112

PR #5916, opened 2020-12-04
[github-actions] update binaries path in upload action

Irving-cl:
#5888 updates the build path for binaries. I just found there were no binaries uploaded in artifacts of action where crash happened. This PR updates the path correspondingly in upload actions.

I will work on something which doesn't use hard-coding. Currently let's just keep it as it is.


simonlingoogle (approved):
LGTM :+1:

bukepo (approved):

jwhui (approved):

PR #5919, opened 2020-12-05
[github-actions] fix backbone test random fails

simonlingoogle:
WIP


.github/workflows/simulation-1.2.yml
```diff
@@ -277,7 +49,7 @@ jobs:
       THREAD_VERSION: 1.2
       INTER_OP: 1
       COVERAGE: 1
-      MULTIPLY: 1
+      MULTIPLY: 20
```
simonlingoogle, line 52:
```suggestion
      MULTIPLY: 50
```

simonlingoogle:
The random fails are caused by #5910 . Will fix in that PR.

PR #5920, opened 2020-12-05
[test] separate code coverage data files

bukepo:
This commit ensures each node create its own profile data, so that they
can be merged after tests to produce more accurate profile data.

This commit also prints the coverage data of message.cpp after all
profile data are merged.

### DIFF
--- a/tests/scripts/expect/_common.exp
+++ b/tests/scripts/expect/_common.exp
@@ -58,6 +58,7 @@ proc wait_for {command success {failure {[\r\n]FAILURE_NOT_EXPECTED[\r\n]}}} {
 proc spawn_node {id {type ""} {radio_url ""}} {
     global spawn_id
     global spawn_ids
+    global argv0
 
     if {${type} == ""} {
         set type $::env(OT_NODE_TYPE)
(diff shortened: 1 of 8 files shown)


tests/scripts/expect/_common.exp
```diff
@@ -69,25 +70,31 @@ proc spawn_node {id {type ""} {radio_url ""}} {
 
     send_user "\n# ${id} ${type}\n"
 
+    if {[info exists ::env(CC)] && $::env(CC) == "clang"} {
```
Moandor-y, line 73: It seems `llvm-cov` supports `GCOV_PREFIX` too. <https://llvm.org/docs/CommandGuide/llvm-cov.html>

Any chance to make this also applied to LLVM?

Moandor-y (approved):
LGTM overall :+1: 

With a question/suggestion below

tests/scripts/expect/_common.exp
bukepo: But there's no `gcov-tool merge` for `llvm-cov`

Moandor-y: OK, I see. Look good.

jwhui (approved):

Issue #5802, opened 2020-11-12
[github-actions] v1_2_LowPower_5_3_01_SSEDAttachment.py fails intermittently


jwhui:
**Describe the bug**

tests/scripts/thread-cert/v1_2_LowPower_5_3_01_SSEDAttachment.py fails intermittently

**Console/log output**

- [logs_37428.zip](https://github.com/openthread/openthread/files/5528157/logs_37428.zip)
- [packet-verification-low-power-pcaps.zip](https://github.com/openthread/openthread/files/5528159/packet-verification-low-power-pcaps.zip)
- https://github.com/openthread/openthread/runs/1388552080

```
Using virtual time
Using Thread 1.2 stack
File "tests/scripts/thread-cert/run_cert_suite.py", line 114, in parse_args
2020-11-12 03:47:32,938 - INFO - Max jobs: 4
File "tests/scripts/thread-cert/run_cert_suite.py", line 115, in parse_args
2020-11-12 03:47:32,939 - INFO - Multiply: 1
File "tests/scripts/thread-cert/run_cert_suite.py", line 116, in parse_args
2020-11-12 03:47:32,939 - INFO - Test scripts: 4
File "tests/scripts/thread-cert/run_cert_suite.py", line 186, in main
2020-11-12 03:47:32,939 - INFO - Has Backbone tests: False
2020-11-12 03:47:33,327 - INFO - Creating node 1: {'is_mtd': False, 'is_bbr': False, 'is_otbr': False, 'is_host': False, 'mode': 'rdn', 'panid': 64206, 'allowlist': [2, 3], 'version': '1.2', 'name': 'LEADER'}
/home/runner/work/openthread/openthread/build/cmake/openthread-simulation-1.2/examples/apps/cli/ot-cli-ftd 1
1: extpanid 000db80000000000
1: extaddr 166e0a0000000001
1: panid 64206
1: mode rdn
2020-11-12 03:47:33,340 - INFO - Creating node 2: {'is_mtd': False, 'is_bbr': False, 'is_otbr': False, 'is_host': False, 'mode': 'rdn', 'panid': 64206, 'allowlist': [1], 'version': '1.2', 'name': 'ROUTER', 'router_selection_jitter': 1}
/home/runner/work/openthread/openthread/build/cmake/openthread-simulation-1.2/examples/apps/cli/ot-cli-ftd 2
2: extpanid 000db80000000000
2: extaddr 166e0a0000000002
2: panid 64206
2: mode rdn
[...]

Issue #5774, opened 2020-11-05
[github-actions] v1_2_LowPower_7_2_01_ForwardTrackingSeries.py fails intermittently


jwhui:
**Describe the bug**

tests/scripts/thread-cert/v1_2_LowPower_7_2_01_ForwardTrackingSeries.py fails intermittently.

**Console/log output**

- [logs_36424.zip](https://github.com/openthread/openthread/files/5492306/logs_36424.zip)
- [packet-verification-low-power-pcaps.zip](https://github.com/openthread/openthread/files/5492315/packet-verification-low-power-pcaps.zip)

```
==============================
!!! FAIL:./tests/scripts/thread-cert/v1_2_LowPower_7_2_01_ForwardTrackingSeries.py
==============================
2020-11-05 05:08:23,869 - INFO - Creating node 1: {'is_mtd': False, 'is_bbr': False, 'is_otbr': False, 'is_host': False, 'mode': 'rdn', 'panid': 64206, 'allowlist': [2, 3], 'version': '1.2', 'name': 'LEADER'}
/home/runner/work/openthread/openthread/build/cmake/openthread-simulation-1.2/examples/apps/cli/ot-cli-ftd 1
1: extpanid 000db80000000000
1: extaddr 166e0a0000000001
1: panid 64206
1: mode rdn
2020-11-05 05:08:23,876 - INFO - Creating node 2: {'is_mtd': False, 'is_bbr': False, 'is_otbr': False, 'is_host': False, 'mode': '-', 'panid': 64206, 'allowlist': [1], 'version': '1.2', 'name': 'SED_1'}
/home/runner/work/openthread/openthread/build/cmake/openthread-simulation-1.2/examples/apps/cli/ot-cli-ftd 2
2: extpanid 000db80000000000
2: extaddr 166e0a0000000002
2: panid 64206
2: mode -
2020-11-05 05:08:23,881 - INFO - Creating node 3: {'is_mtd': False, 'is_bbr': False, 'is_otbr': False, 'is_host': False, 'mode': '-', 'panid': 64206, 'allowlist': [1], 'version': '1.2', 'name': 'SSED_1'}
/home/runner/work/openthread/openthread/build/cmake/openthread-simulation-1.2/examples/apps/cli/ot-cli-ftd 3
3: extpanid 000db80000000000
3: extaddr 166e0a0000000003
3: panid 64206
3: mode -
2: extaddr
1: macfilter addr add 166e0a0000000002
3: extaddr
[...]

PR #5921, opened 2020-12-06
[mac] stop timer before starting non-idle operation

Irving-cl:
This PR fixes issue #5802 (and probably #5774).

Luckily, I caught a failed run of the Heisenbug after the core dump uploading feature is enabled. So I analyzed the core dump:
```
GNU gdb (Ubuntu 9.2-0ubuntu1~20.04) 9.2
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from build/openthread-simulation-1.2/examples/apps/cli/ot-cli-ftd...
[New LWP 11076]
Core was generated by `/home/runner/work/openthread/openthread/build/openthread-simulation-1.2/example'.
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) 
(gdb) '
Undefined command: "".  Try "help".
(gdb) up
#1  0x00007ff86146a859 in __GI_abort () at abort.c:79
79      abort.c: No such file or directory.
(gdb) up
#2  0x00007ff86146a729 in __assert_fail_base (fmt=0x7ff861600588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x561fb6767671 "false", file=0x561fb6767656 "../../src/core/mac/mac.cpp", line=1549, function=<optimized out>) at assert.c:92
92      assert.c: No such file or directory.
(gdb) up
#3  0x00007ff86147bf36 in __GI___assert_fail (assertion=0x561fb6767671 "false", file=0x561fb6767656 "../../src/core/mac/mac.cpp", 
    line=1549, function=0x561fb6767778 "void ot::Mac::Mac::HandleTimer()") at assert.c:101
101     in assert.c
(gdb) up
#4  0x0000561fb664138b in ot::Mac::Mac::HandleTimer (this=0x561fb686d478 <ot::gInstanceRaw+26968>) at ../../src/core/mac/mac.cpp:1549
1549            OT_ASSERT(false);
(gdb) p mOperation
$1 = ot::Mac::Mac::kOperationTransmitDataDirect
(gdb) p mPendingTransmitDataCsl
$2 = true
(gdb) p mRxOnWhenIdle
$3 = true
(gdb) p mCslTxFireTime
$4 = {static kMaxDuration = 4294967295, mValue = 20497}
(gdb) p mTimer
$5 = {<ot::Timer> = {<ot::InstanceLocator> = {<No data fields>}, <ot::OwnerLocator> = {<No data fields>}, <ot::LinkedListEntry<ot::Timer>> = {<No data fields>}, static kMaxDelay = 2147483647, mHandler = @0x561fb6641188, mFireTime = {static kMaxDuration = 4294967295, 
      mValue = 20497}, mNext = 0x561fb686d5a0 <ot::gInstanceRaw+27264>}, <No data fields>}
(gdb) p mPendingTransmitDataDirect
$6 = false
(gdb) 
```

So the crash is caused by the assetion in `Mac::HandleTimer`:
```
void Mac::HandleTimer(void)                                                                                                                                                                                                                                                              
{                                                                                                                                                                                                                                                                                        
    switch (mOperation)                                                                                                                                                                                                                                                                  
    {                                                                                                                                                                                                                                                                                    
    case kOperationActiveScan:                                                                                                                                                                                                                                                           
        PerformActiveScan();                                                                                                                                                                                                                                                             
        break;                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                         
    case kOperationWaitingForData:                                                                                                                                                                                                                                                       
        otLogDebgMac("Data poll timeout");                                                                                                                                                                                                                                               
        FinishOperation();
        Get<DataPollSender>().HandlePollTimeout();
        PerformNextOperation();
        break;

    case kOperationIdle:
        if (!mRxOnWhenIdle)
        {
#if OPENTHREAD_CONFIG_MAC_STAY_AWAKE_BETWEEN_FRAGMENTS
            if (mDelayingSleep)
            {
                otLogDebgMac("Sleep delay timeout expired");
                mDelayingSleep = false;
                UpdateIdleMode();
            }
#endif
        }
        else if (mPendingTransmitDataCsl)
        {
            PerformNextOperation();
        }
#endif
        break;

    default:
        OT_ASSERT(false); // !!!!!!! Crashed here
        OT_UNREACHABLE_CODE(break);
    }
}
```
As shown in the gdb copying text, currently `mOperation` is `kOperationTransmitDataDirect`. And we can see that the time  for `mTimer` is the same with `mCslTxFireTime`. So the timer is started by CSL scheduling in `UpdateIdleMode`. This problem seems to be exactly the one we discussed in the initial CSL PR: https://github.com/openthread/openthread/pull/4557#issuecomment-657742939
So to prevent such cases, we need to stop the timer before staring any non-idle operations as suggested in the comment. As I see, currently it's not ensured in the code. I think it's my mistake that I missed it when submitting the initial CSL PR. 

Please have a review. @abtink @brianljt

### DIFF
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -854,10 +854,12 @@ void Mac::PerformNextOperation(void)
         break;
 
     case kOperationActiveScan:
+        mTimer.Stop();
         PerformActiveScan();
         break;
 
     case kOperationEnergyScan:
+        mTimer.Stop();
         PerformEnergyScan();
         break;
 
@@ -871,6 +873,7 @@ void Mac::PerformNextOperation(void)
 #endif
     case kOperationTransmitPoll:
     case kOperationTransmitOutOfBandFrame:
+        mTimer.Stop();
         BeginTransmit();
         break;
 
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -845,6 +845,8 @@ void Mac::PerformNextOperation(void)
     if (mOperation != kOperationIdle)
     {
         otLogDebgMac("Starting operation \"%s\"", OperationToString(mOperation));
+        mTimer.Stop(); // Stop the timer before any non-idle operation, have the operation itself be responsible to
+                       // start the timer (if it wants to).
     }
 
     switch (mOperation)
@@ -854,12 +856,10 @@ void Mac::PerformNextOperation(void)
         break;
 
     case kOperationActiveScan:
-        mTimer.Stop();
         PerformActiveScan();
         break;
 
     case kOperationEnergyScan:
-        mTimer.Stop();
         PerformEnergyScan();
         break;
 
@@ -873,12 +873,12 @@ void Mac::PerformNextOperation(void)
 #endif
     case kOperationTransmitPoll:
     case kOperationTransmitOutOfBandFrame:
-        mTimer.Stop();
         BeginTransmit();
         break;
 
     case kOperationWaitingForData:
         IgnoreError(mSubMac.Receive(mRadioChannel));
+        mTimer.Start(kDataPollTimeout);
         break;
     }
 
@@ -1419,7 +1419,6 @@ void Mac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, otError aError
 
             if (IsEnabled() && framePending)
             {
-                mTimer.Start(kDataPollTimeout);
                 StartOperation(kOperationWaitingForData);
             }
 


abtink (approved):
Nice find! LGTM. Thanks 👍

abtink:
Thank @Irving-cl. The fix you have in the PR works.

I was considering a different one (may be slightly smaller and I think may be better for future if we add more operations):

- In this PR, you have the `mTimer.Stop()` under all case values for scan and all tx related operation, but not `kOperationIdle` (which is what we want) and `kOperationWaitingForData` (which is agin correct).
- For `kOperationWaitingForData`, we currently start the timer when we are done with poll tx (assuming ack is indicating frame-pending) in `HandleTransmitDone()` and issue a `StartOperation(kOperationWaitingForData)` (which schedules the "waiting-for-data" operation).
- Since the `kOperationWaitingForData` is the highest priority in `PerformNextOperation()` we would go to it right after current poll tx operation is done (so after txing a poll (with frame pending ack) we should always go to "waiting-for-data" operation/state).

Here is what I have in mind:
- We move the timer start for the poll timeout from `HandleTransmitDone()` to `PerformNextOperation()` under
```
 case kOperationWaitingForData:
        IgnoreError(mSubMac.Receive(mRadioChannel));
+       mTimer.Start(kDataPollTimeout);
        break;
```
- With this we are still practically starting the timer on tx done of the poll (may be even better since it's tied to when we actually start the "wait-for-data" operation).

- Then we add a single `mTimer.Stop()` in `PerformNextOperation()` right before the `switch()` and in this `if()` block:
```
    if (mOperation != kOperationIdle)
    {
        otLogDebgMac("Starting operation \"%s\"", OperationToString(mOperation));
+       mTimer.Stop()`
    }
```
- Basically stopping `mTimer` when the we are about to perform any operation (any other than idle) and have the operation itself be responsible to start the timer (if it wants to use it).

Thouhts?

Irving-cl:
@abtink 

The new approach is better and more extensible. Thanks! I've updated.

abtink (approved):
Thanks. LGTM. 👍

brianljt (approved):
Nice catch! LGTM. Thanks 👍

jwhui (approved):

PR #5922, opened 2020-12-06
[dns] add helper methods to append/parse DNS names

abtink:
This commit adds `Dns::Name` type which provides a set of helper
methods to encode/decode DNS names. `AppendName()` method encodes and
appends a full name (e.g., "test.example.com") to a message. Other
helper methods enable appending labels (or groups of labels) and/or
constructing a compressed name (using pointer labels). `ParseName()`
method parses and skips over a full name in a message. `ReadLabel()`
method reads labels one by one and works independently of whether the
encoded name is compressed or not. `ReadName()` method read an entire
name from message. The new methods are used in `Dns::Client`.

This commit also add a unit test `test-dns` covering the behavior of
the newly added helper methods.

Finally, this commit adds few helper methods in `ResourceRecords` to
get the size and check the type of a DNS record.


wgtdkp (approved):
Thanks @abtink  ! This could be very useful for DNS message processing. :+1:

src/core/net/dns_headers.hpp
```diff
@@ -374,6 +374,182 @@ class Header : public Clearable<Header>
...
+     * This static method encodes and appends a single domain name label to a message.
+     *
+     * The @p aLabel is assumed to contain a single domain name label as a C string (null-terminated). Unlike
+     * `AppendMultipleLabels()` which parses the label string and treats it as sequence of multiple (dot-separated)
+     * labels, this method always appends @p aLabel as a single whole label. This allows the label string to even
+     * contain dot '.' character, which, for example, is useful for "Service Instance Names" where <Instance> portion
+     * is a user-friendly name and can contain dot character.
+     *
+     * @param[in] aLabel              The label string to append. MUST not be nullptr.
+     * @param[in] aMessage            The message to append to.
```
Irving-cl, line 400:
```suggestion
     * @param[out] aMessage            The message to append to.
```

Irving-cl: Or maybe we should call it `inout`? And other occurrences of `aMessage` in API docs.

src/core/net/dns_headers.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+        // Zero label length indicates end of a name.
+        VerifyOrExit(labelLength != 0, error = OT_ERROR_NOT_FOUND);
+
+        switch (labelLength & kLabelTypeMask)
+        {
+        case kTextLabelType:
+            VerifyOrExit(labelLength < aLabelBufferSize, error = OT_ERROR_NO_BUFS);
+            SuccessOrExit(error = aMessage.Read(aOffset + sizeof(uint8_t), aLabelBuffer, labelLength));
+            aLabelBuffer[labelLength] = '\0';
```
Irving-cl, line 197:
```suggestion
            aLabelBuffer[labelLength] = kNullChar;
```

Irving-cl (approved):
LGTM :+1:

src/core/net/dns_headers.hpp
```diff
@@ -374,6 +374,182 @@ class Header : public Clearable<Header>
...
+     * returned. This method handles compressed domain names which use pointer labels. So as the labels in a name are
+     * read, the @p aOffset may jump back in the message and at the end the @p aOffset does not necessarily point to
+     * the end the original name field.
+     *
+     * @param[in]    aMessage         The message to read the label from.
+     * @param[inout] aOffset          On input, the offset in @p aMessage pointing to the start of the label to read.
+     *                                On exit, when successfully read, @p aOffset is updated to point to the start of
+     *                                the next label.
+     * @param[in]    aHeaderOffset    The offset in @p aMessage to the start of the DNS header.
+     * @param[out]   aLabelBuffer     A pointer to a char array to output the read label as a null-terminated C string.
```
wgtdkp, line 507: What about also outputting the label length? It could be useful if we want to join multiple labels in a loop.

src/core/net/dns_headers.hpp
abtink: We seem to use `@param[in]` in a lot of other similar situations where things in a `aMessage` (or in general an object) can change by the method/function.  Examples in the code:
- `Mle::AppendHeader(Message &aMessage)`, `AppendMode()`, and a bunch more in `Mle`,
- In `Tlv` class we have a lot of `Tlv::Append(Message &aMessage)` flavors,
- `Dataset::AppendMleDatasetTlv(Message &aMessage)`,
- `MessageQueue::Enqueue()/Dequeue()` also refer to `aMessage` as `@param[in]` (though they technically change some state in the `aMessage` object).

For consistency, I'd prefer to keep it the same in this PR.

----

We can discuss whether it makes sense to change all other occurrences (from a different PR probably):

Here is my preference on this:
- I would say `@param[out]` would be good when variable/object is changed fully (e.g., a new instance of object is created/used or object is modified completely).
- For `Append()` like of methods I would prefer to continue to use `@param[in]` over `[inout]` (if we go for `[inout]` the situations for methods like `Enqueue()/Dequeue()` is harder to consider).

Thoughst? @jwhui, others.

src/core/net/dns_headers.hpp
abtink: We can return the label length in `aLabelBufferSize` (probably rename the variable). I was considering it but was not sure if would be immediately useful. Will add it later in the next push (I think it would not add much complexity/overhead and it is an added functionality).

----

Later, we can consider adding method `ReadName()` to read and create a dot separated domain name string. We should be careful of cases where a (first) label can have dot `'.'` char in it (e.g., Service Instance Name). The approach seems to be to escape the dot char with `\` (and then if there are `\` chars in the label they become `\\`).

abtink: Added the update on `aLabelLength` from `ReadLabel()` in the new pushed commit.

src/core/net/dns_headers.cpp
```diff
@@ -0,0 +1,238 @@
...
+otError DomainName::AppendPointerLabel(uint16_t aOffset, Message &aMessage)
+{
+    // A pointer label takes the form of a two byte sequence as a
+    // `uint16_t` value. The first two bits are ones. This allows a
+    // pointer to be distinguished from a text label, since the text
+    // label must begin with two zero bits (note that labels are
+    // restricted to 63 octets or less). The next 14-bits specify
+    // an offset value relative to start of DNS header.
+
+    uint16_t value = HostSwap16(aOffset | kPointerLabelTypeUint16);
```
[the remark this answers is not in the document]
abtink: I added an `OT_ASSERT()` check for this.

src/core/net/dns_headers.hpp
Irving-cl: SGTM.

src/core/net/dns_headers.hpp
wgtdkp: > Later, we can consider adding method ReadName() to read and create a dot separated domain name string.

This is what I am planning to add as this is immediately useful on the server side, good to have it if you can add it in this PR.

> We should be careful of cases where a (first) label can have dot '.' char in it (e.g., Service Instance Name). 

Sorry I didn't get the point. Since the name is already encoded as a sequences of labels each with a preceding length field, it looks to me that we only need to append a `.` at end of each label to create the full name.

Not sure if the dots in the instance or host name worth a special handling. Refusing such names at the client side will work for me, although other DNS implementations may allow this. I believe `allowing dot in service instance name` is a mistake :(

abtink: I can add `ReadName()` from this PR (will add from next next push). 😃 

---- 

From my understanding/reading, the "Instance Service Name" label seems to be treated differently from other labels. It is intended to be a human readable name so can have more characters than allowed by other labels (details in [RFC-6763-4.1.1](https://tools.ietf.org/html/rfc6763#section-4.1.1) and more in appendix). There is also a section suggesting the escaping model for the Instance Name [RFC6764-4.3](https://tools.ietf.org/html/rfc6763#section-4.3). I just mentioned this for us to be aware of it. 

For now we can have `ReadName()` assume there is no dot in the any of labels. We will see if we need to handle the Instance Service Name differently later. We can always read the Instance Service Name label separately using `ReadLabel()` which should handle it correctly.

abtink:
In the new push, added `ReadName()` which read a full domain name from a message (returning it as dot `"."` seperated sequence of labels).

To share the common parsing code, the new commit also adds a private sub-type `LabelIterator` definint some helper methods which are then used by `ReadLabel()`, `ReadName()` and `ParseName()`. Unit test also covers the `ReadName()`

src/core/net/dns_headers.hpp
abtink: Added `RaadName()` in new pushed commit.

wgtdkp (approved):
LGTM :+1:

src/core/net/dns_headers.cpp
```diff
@@ -0,0 +1,351 @@
...
+        {
+            mNameEndOffset = mNextLabelOffset + sizeof(uint16_t);
+        }
+
+        mNextLabelOffset = mHeaderOffset + (HostSwap16(pointerValue) & kPointerLabelOffsetMask);
+
+        // `OT_ERROR_PENDING` is used to indicate that we reached
+        // a pointer label.
+
+        error = OT_ERROR_PENDING;
```
simonlingoogle, line 303: Can we follow pointer labels until we find a text label (in this method) so that we can avoid returning `OT_ERROR_PENDING`? 
I think this can simplify the callers of `GetNextLabel`.

simonlingoogle (approved):
LGTM :+1: 

with a nit

src/core/net/dns_headers.cpp
abtink: This is currently used in `ParseName()` where we want to get the offset to the end of the encoded name field. For this, we want to stop if we reach a pointer label (we don't care about any next label in this case).

We can continue reading the next label from `GetNextLabel()` and check in `ParseName()` if `mEndOffset` is set on `LabelIterator` (which indicate whethere we have reached the end of the name field and passed over a pointer label).
The downside I see is that we would do a bit extra work if we do this during `ParseName()` but I agree it may potentially simplify the code in other cases. Let me try it out and see how it looks. Good suggestion. Thanks. 👍 

For return type, I think we need to return `otError` since we can have `PARSE_ERROR` as well (e.g., invalid label type or message ends abruptly and does not contain the label field and/or pointer value at the given offset).

src/core/net/dns_headers.hpp
```diff
@@ -374,6 +374,243 @@ class Header : public Clearable<Header>
...
+     */
+    static otError AppendLabel(const char *aLabel, Message &aMessage);
+
+    /**
+     * This static method encodes and appends a sequence of domain name labels to a given message.
+     *
+     * The @p aLabels must follow  "<label1>.<label2>.<label3>", i.e., a sequence of labels separated by dot '.' char.
+     * E.g., ""_http._tcp", "_http._tcp." (same as previous one), "host-1.test".
+     *
+     * This method validates that the @p aLabels is a valid domain name format, i.e., no empty label, and labels are
```
jwhui, line 415: Are we validating everything in RFC 1035 Section 2.3?

src/core/net/dns_headers.hpp
```diff
@@ -374,6 +374,243 @@ class Header : public Clearable<Header>
...
+     */
+    static otError AppendPointerLabel(uint16_t aOffset, Message &aMessage);
+
+    /**
+     * This static method encodes and appends a full domain name to a message.
+     *
+     * The @p aName must follow  "<label1>.<label2>.<label3>", i.e., a sequence of labels separated by dot '.' char.
+     * E.g., "example.com", "example.com." (same as previous one), local.", "default.service.arpa", "." or "" (root).
+     *
+     * This method validates that the @p aName is a valid domain name format, i.e. no empty labels, and labels are
```
jwhui, line 466: Are we validating everything in RFC 1035 Section 2.3?

src/core/net/dns_headers.hpp
```diff
@@ -374,6 +374,243 @@ class Header : public Clearable<Header>
...
+    static otError ReadLabel(const Message &aMessage,
+                             uint16_t &     aOffset,
+                             uint16_t       aHeaderOffset,
+                             char *         aLabelBuffer,
+                             uint8_t &      aLabelLength);
+
+    /**
+     * This static method reads a full domain name from a message.
+     *
+     * On successful read, the read name follow  "<label1>.<label2>.<label3>.", i.e., a sequence of labels separated by
```
jwhui, line 531:
```suggestion
     * On successful read, the read name follows  "<label1>.<label2>.<label3>.", i.e., a sequence of labels separated by
```

src/core/net/dns_headers.hpp
```diff
@@ -382,6 +619,54 @@ OT_TOOL_PACKED_BEGIN
...
+
+    /**
+     * This method initializes the resource record.
+     *
+     * @param[in] aType   The type of the resource record.
+     * @param[in] aClass  The class of the resource record (default is `kClassInternet`).
+     * @param[in] aTtl    The time to live field of the resource record (default is zero).
+     *
+     */
+    void Init(uint16_t aType, uint16_t aClass = kClassInternet, uint32_t aTtl = 0);
```
jwhui, line 654: Do we have a use case for other class values?

src/core/net/dns_headers.hpp
```diff
@@ -382,6 +619,54 @@ OT_TOOL_PACKED_BEGIN
...
+
+    /**
+     * This method indicates whether the resources records matches a given type and class code.
+     *
+     * @param[in] aType   The resource record type to compare with.
+     * @param[in] aClass  The resource record class code to compare with (default is `kClassInternet`).
+     *
+     * @returns TRUE if the resources records matches @p aType and @p aClass, FAlSE otherwise.
```
jwhui, line 662:
```suggestion
     * @returns TRUE if the resources records matches @p aType and @p aClass, FALSE otherwise.
```

src/core/net/dns_headers.hpp
```diff
@@ -374,6 +374,243 @@ class Header : public Clearable<Header>
...
+        kMaxLength      = 255, ///< Max number of characters in a domain name.
+    };
+
+    /**
+     * This static method encodes and appends a single domain name label to a message.
+     *
+     * The @p aLabel is assumed to contain a single domain name label as a C string (null-terminated). Unlike
+     * `AppendMultipleLabels()` which parses the label string and treats it as sequence of multiple (dot-separated)
+     * labels, this method always appends @p aLabel as a single whole label. This allows the label string to even
+     * contain dot '.' character, which, for example, is useful for "Service Instance Names" where <Instance> portion
```
jwhui, line 396: nit: I agree it would be desirable to reuse these methods for other related things. Note, however, that Service Instances Names include Domain Names, but Service Instances Names are not Domain Names in themselves. Should we move more general methods to a class like `DnsName`?

abtink: Good point. Thanks. I was actually considering renaming the type from `DomainName` to be just `Name` (which is already under `Dns` namespace, so it's a `Dns::Name`). I think your point here makes a good argument why we should not call it `DomainName`. I will change it in the next push.

src/core/net/dns_headers.hpp
abtink: No. Just the 3 things mentioned, are really checked, i.e., no empty label, labels limited to 63 chars or less, overall name 255 bytes (the first two are useful to ensure the encoded name can be parsed correctly).

We can add more checks/validations but I am not sure if it would be useful. Main downside is that it would add code-size. I think we can rely on caller/provider of the names to provide correct/valid ones.

src/core/net/dns_headers.hpp
wgtdkp: There is use cases in [DNS UPDATE](https://tools.ietf.org/html/rfc2136) that the class is used to decide the type of a DNS update (the `class` is compared against `ANY CLASS` and `NONE CLASS`).

src/core/net/dns_headers.cpp
simonlingoogle: > The downside I see is that we would do a bit extra work if we do this during ParseName()

The extra work would very likely to be just one more loop to find the real text label, which shouldn't be significant. 

I am fine to keep `otError` as the return value.

src/core/net/dns_headers.hpp
abtink: I suggest we keep the method as is for now. We can optimize it later.

src/core/net/dns_headers.cpp
abtink: Changed in new pushed commit.

src/core/net/dns_headers.hpp
abtink: Changed in new pushed commit.

jwhui: Thanks!

src/core/net/dns_headers.hpp
```diff
@@ -382,6 +618,54 @@ OT_TOOL_PACKED_BEGIN
...
+    /**
+     * Resource Record Types.
+     *
+     */
+    enum : uint16_t
+    {
+        kTypeA    = 1,  ///< Address record (IPv4).
+        kTypePtr  = 12, ///< PTR record.
+        kTypeTxt  = 16, ///< TXT record.
+        kTypeSrv  = 33, ///< Service locator record.
```
jwhui, line 630: nit: I don't think "service locator" is the right term. What about being consistent with the ones above?
```suggestion
        kTypeSrv  = 33, ///< SRV locator record.
```

jwhui (approved):

PR #5925, opened 2020-12-07
[dua] delay registration for DUA newly added

simonlingoogle:
**Background:**

```
     PBBR
   /      \
ROUTER---LEADER
```

In the tests of #5910 , we found that it's possible that a Router, in a topology like above, could send DUA.req so quickly after receiving `Data Response` from Leader that the PBBR is not able to process the same `Data Response` message and update the network data accordingly. Consequently, PBBR sends DUA.rsp with `kDuaInvalid` status because it can't find a valid Domain Prefix in the Network Data. 

This commit help resolve this issue with two fixes:
- Node to delay for up to 1000ms before sending DUA.req when domain prefix addresses is generated
- PBBR to respond with `kDuaGeneralFailure` to allow the node to re-register the same DUA again.

### DIFF
--- a/src/core/thread/dua_manager.cpp
+++ b/src/core/thread/dua_manager.cpp
@@ -329,7 +329,7 @@ void DuaManager::HandleNotifierEvents(Events aEvents)
 #if OPENTHREAD_CONFIG_DUA_ENABLE
     if (aEvents.ContainsAny(kEventIp6AddressAdded))
     {
-        UpdateRegistrationDelay(1);
+        UpdateRegistrationDelay(kNewDuaRegistrationDelay);
     }
 #endif
 }
--- a/src/core/thread/dua_manager.cpp
+++ b/src/core/thread/dua_manager.cpp
@@ -326,10 +326,12 @@ void DuaManager::HandleNotifierEvents(Events aEvents)
 #endif
     }
 
+#if OPENTHREAD_CONFIG_DUA_ENABLE
     if (aEvents.ContainsAny(kEventIp6AddressAdded))
     {
-        mRegistrationTask.Post();
+        UpdateRegistrationDelay(1);
     }
+#endif
 }
 
 void DuaManager::HandleBackboneRouterPrimaryUpdate(BackboneRouter::Leader::State               aState,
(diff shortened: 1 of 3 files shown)


simonlingoogle:
@librasungirl @jwhui 

I came to realize that this issue is more serious in simulation because we haven't implemented `CSMA-CA` in radio simulation. In real network, the `ROUTER` would send `DUA.req` after a random CSMA delay. That being said, I still think that a random delay can be helpful to avoid potential channel conflicts when the domain prefix is configured in the Network Data.

jwhui:
> I came to realize that this issue is more serious in simulation because we haven't implemented `CSMA-CA` in radio simulation. In real network, the `ROUTER` would send `DUA.req` after a random CSMA delay. That being said, I still think that a random 0~1000ms delay can be helpful to avoid potential channel conflicts when the domain prefix is configured in the Network Data.

@simonlingoogle , I believe we do implement the random backoff in simulation at the sub-MAC layer. However, we do not implement the clear channel assessment (CCA).

simonlingoogle:
> > I came to realize that this issue is more serious in simulation because we haven't implemented `CSMA-CA` in radio simulation. In real network, the `ROUTER` would send `DUA.req` after a random CSMA delay. That being said, I still think that a random 0~1000ms delay can be helpful to avoid potential channel conflicts when the domain prefix is configured in the Network Data.
> 
> @simonlingoogle , I believe we do implement the random backoff in simulation at the sub-MAC layer. However, we do not implement the clear channel assessment (CCA).

Yes. Maybe that's why this is a random fail by the a chance of about 30%. 
If so, this is also an issue for real devices because two nodes will conflict if they happened to use the same CSMA backoff time.

src/core/thread/dua_manager.cpp
```diff
@@ -326,10 +326,12 @@ void DuaManager::HandleNotifierEvents(Events aEvents)
 #endif
     }
 
+#if OPENTHREAD_CONFIG_DUA_ENABLE
     if (aEvents.ContainsAny(kEventIp6AddressAdded))
     {
-        mRegistrationTask.Post();
+        UpdateRegistrationDelay(1);
```
librasungirl, line 332: nit: use a constant, other than an inline magic number?

librasungirl (approved):

jwhui (approved):

PR #5939, opened 2020-12-11
[low-power] small improvements in CSL code

edmont:
Small code simplification in CSL part.

### DIFF
--- a/src/core/thread/csl_tx_scheduler.cpp
+++ b/src/core/thread/csl_tx_scheduler.cpp
@@ -119,7 +119,6 @@ void CslTxScheduler::Clear(void)
  */
 void CslTxScheduler::RescheduleCslTx(void)
 {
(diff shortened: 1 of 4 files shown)


src/core/thread/csl_tx_scheduler.cpp
```diff
@@ -133,7 +132,7 @@ void CslTxScheduler::RescheduleCslTx(void)
             continue;
         }
 
-        delay = GetNextCslTransmissionDelay(child, radioNow, cslTxDelay);
```
Irving-cl, line 136: Just to mention that I used same `radioNow` for each child so that we can compare each child's next csl tx delay based on same `Now` value and save the cost for calling radio API. Not sure how much time would pass after processing one child (I believe it's short enough to ignore). Anyway, I agree with the new approach.

Irving-cl (approved):
LGTM :+1: Thanks!

PR #5951, opened 2020-12-14
[thread-cert] add case 9.2.19 PendingDatasetGet using pktverify

FiveDimensions:


tests/scripts/thread-cert/Cert_9_2_19_PendingDatasetGet.py
```diff
@@ -0,0 +1,300 @@
...
+        self.nodes[LEADER].start()
+        self.simulator.go(5)
+        self.assertEqual(self.nodes[LEADER].get_state(), 'leader')
+
+        self.nodes[COMMISSIONER].start()
+        self.simulator.go(5)
+        self.assertEqual(self.nodes[COMMISSIONER].get_state(), 'router')
+        self.simulator.get_messages_sent_by(LEADER)
+
+        self.collect_leader_aloc(LEADER)
```
simonlingoogle, line 88: why do we need to collect leader ALOC? I believe leader ALOC is fixed in test.

simonlingoogle (approved):
LGTM :+1: 

with a nit

tests/scripts/thread-cert/Cert_9_2_19_PendingDatasetGet.py
FiveDimensions: added const

jwhui (approved):

PR #5957, opened 2020-12-14
[multi-radio] copy all TX buffers before encryption

jwhui:
Performing MAC frame encryption immediately after each copy results in
copying the already-encrypted frame, rather than the original frame
before encryption.


abtink (approved):
LGTM. Thanks for finding and fixing it. 👍

PR #5958, opened 2020-12-15
[harness-automation] fix the result parsed uncorrectly using Chrome 80

zwx1995esp:
When using "status = self._browser.find_element_by_class_name('title-test').text", the variable "status" will not contain "Pass" or "Failed", so, the result always fails.
After using "get_attribute('innerText')",  the variable "status" will contain "Pass" or "Failed" as expected


zwx1995esp:
@googlebot I signed it!

FiveDimensions (approved):
LGTM

jwhui (approved):

PR #5961, opened 2020-12-16
[nrf528xx] add max power table support

gjc13:


examples/platforms/nrf528xx/src/radio.c
```diff
@@ -444,13 +453,17 @@ otError otPlatRadioReceive(otInstance *aInstance, uint8_t aChannel)
...
     nrf_802154_channel_set(aChannel);
     if (nrf_802154_state_get() == NRF_802154_STATE_SLEEP)
     {
         // Enable FEM before RADIO leaving SLEEP state.
         nrf5FemEnable();
     }
 
+    sDefaultTxPower = sDefaultTxPower < sMaxTxPowerTable[sCurrentChannel - OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN]
+                          ? sDefaultTxPower
+                          : sMaxTxPowerTable[sCurrentChannel - OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN];
```
zhanglongxia, line 466: This code may cause the same channel uses different output powers.
For example:  We assume that the max power is 20dBm for the channel 11. The max power is 18dBm for the channel 25. The current channel `sCurrentChannel ` is channel 11, the current `sDefaultTxPower ` is 20dBm.

<1> Set the channel to 25, then `sDefaultTxPower ` will be set to 18 dBm.
<2> Set the channel to 11, then `sDefaultTxPower ` will still be set to 18 dBm.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -444,13 +453,17 @@ otError otPlatRadioReceive(otInstance *aInstance, uint8_t aChannel)
 
     bool result;
 
+    sCurrentChannel = aChannel;
```
zhanglongxia, line 456: Can we use the function `nrf_802154_channel_get` to replace the variable `sCurrentChannel `?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -1299,3 +1315,40 @@ otError otPlatRadioConfigureEnhAckProbing(otInstance *         aInstance,
...
+            sDefaultTxPower = sDefaultTxPower < aMaxPower ? sDefaultTxPower : aMaxPower;
+        }
+        nrf_802154_tx_power_set(sDefaultTxPower);
+    }
+
+exit:
+    return error;
+}
+
+int8_t nrf5GetChannelMaxTransmitPower(uint8_t aChannel)
```
zhanglongxia, line 1346: Which function calls this function?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -160,6 +163,12 @@ static void dataInit(void)
         sReceivedFrames[i].mPsdu = NULL;
     }
 
+    for (size_t i = 0; i < OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MAX - OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN + 1; i++)
```
brianljt, line 166: Since `sMaxTxPowerTable` is a fixed size array, how about
```suggestion
    for (size_t i = 0; i < sizeof(sMaxTxPowerTable); i++)
```

examples/platforms/nrf528xx/src/radio.c
brianljt: Echo. I don't quite get what is the purpose here. Why not get rid of `sDefaultTxPower` and use local variable?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -1299,3 +1315,40 @@ otError otPlatRadioConfigureEnhAckProbing(otInstance *         aInstance,
...
+}
+
+int8_t nrf5GetChannelMaxTransmitPower(uint8_t aChannel)
+{
+    if (aChannel < OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN || aChannel > OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MAX)
+    {
+        return OT_RADIO_POWER_INVALID;
+    }
+
+    return sMaxTxPowerTable[aChannel - OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN];
```
brianljt, line 1353: Need to check status of `sPowerInitialized` here?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -107,8 +107,11 @@ static otInstance *  sInstance = NULL;
 static otRadioFrame sAckFrame;
 static bool         sAckedWithFramePending;
 
-static int8_t sDefaultTxPower;
-static int8_t sLnaGain = 0;
+static int8_t  sMaxTxPowerTable[OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MAX - OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN + 1];
+static uint8_t sCurrentChannel;
```
bukepo, line 111: can we use `nrf_802154_channel_set()`?

examples/platforms/nrf528xx/src/radio.c
gjc13: Internal radio drivers.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -695,9 +708,15 @@ otError otPlatRadioGetTransmitPower(otInstance *aInstance, int8_t *aPower)
...
 {
     OT_UNUSED_VARIABLE(aInstance);
+    int8_t  power;
+    uint8_t channel = nrf_802154_channel_get();
 
-    sDefaultTxPower = aPower;
-    nrf_802154_tx_power_set(aPower);
+    sDefaultTxPower   = aPower;
+    sPowerInitialized = true;
+    power             = aPower < sMaxTxPowerTable[channel - OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN]
```
brianljt, line 716: Sanity check for `channel` before use or use `nrf5GetChannelMaxTransmitPower` instead?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -451,7 +461,10 @@ otError otPlatRadioReceive(otInstance *aInstance, uint8_t aChannel)
         nrf5FemEnable();
     }
 
-    nrf_802154_tx_power_set(sDefaultTxPower);
+    transmitPower = sDefaultTxPower < sMaxTxPowerTable[channel - OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN]
+                        ? sDefaultTxPower
+                        : sMaxTxPowerTable[channel - OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN];
```
brianljt, line 466: 1. Sanity check for `channel` before use or use `nrf5GetChannelMaxTransmitPower` instead?
2. Check `sPowerInitialized`?
```suggestion
    int8_t channelMaxTxPower = nrf5GetChannelMaxTransmitPower(channel);
    
    transmitPower = sPowerInitialized && sDefaultTxPower < channelMaxTxPower
                        ? sDefaultTxPower
                        : channelMaxTxPower;
```

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -451,7 +461,8 @@ otError otPlatRadioReceive(otInstance *aInstance, uint8_t aChannel)
         nrf5FemEnable();
     }
 
-    nrf_802154_tx_power_set(sDefaultTxPower);
+    transmitPower = sDefaultTxPower < channelMaxPower ? sDefaultTxPower : channelMaxPower;
```
brianljt, line 464: Any need to check `sPowerInitialized` before using `sDefaultTxPower`?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -107,8 +108,10 @@ static otInstance *  sInstance = NULL;
 static otRadioFrame sAckFrame;
 static bool         sAckedWithFramePending;
 
+static int8_t sMaxTxPowerTable[OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MAX - OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN + 1];
 static int8_t sDefaultTxPower;
```
zhanglongxia, line 112: Should we set initialize value to the variable `sDefaultTxPower`?  The function `otPlatRadioReceive ()` may be called before the variable `sDefaultTxPower` is initialized.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -1299,3 +1313,41 @@ otError otPlatRadioConfigureEnhAckProbing(otInstance *         aInstance,
...
+    otEXPECT_ACTION(aChannel >= OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN && aChannel <= OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MAX,
+                    error = OT_ERROR_INVALID_ARGS);
+
+    sMaxTxPowerTable[aChannel - OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN] = aMaxPower;
+    if (aChannel == nrf_802154_channel_get())
+    {
+        int8_t power;
+        if (!sPowerInitialized)
+        {
+            sDefaultTxPower = aMaxPower;
```
zhanglongxia, line 1331: The function `otPlatRadioSetChannelMaxTransmitPower ` is used to set the max transmit power, why we need to initialize the variable `sDefaultTxPower` at here?

examples/platforms/nrf528xx/src/radio.c
Irving-cl, line 464: We can use `OT_MIN` defined in `code_utils.hpp`.

Irving-cl (approved):
LGTM 👍

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -695,9 +706,12 @@ otError otPlatRadioGetTransmitPower(otInstance *aInstance, int8_t *aPower)
...
 {
     OT_UNUSED_VARIABLE(aInstance);
+    uint8_t channel         = nrf_802154_channel_get();
+    int8_t  channelMaxPower = nrf5GetChannelMaxTransmitPower(channel);
 
-    sDefaultTxPower = aPower;
-    nrf_802154_tx_power_set(aPower);
+    sDefaultTxPower   = aPower;
+    sPowerInitialized = true;
+    nrf_802154_tx_power_set(aPower < channelMaxPower ? aPower : channelMaxPower);
```
zhanglongxia, line 714: The function `otPlatRadioGetTransmitPower` calls `nrf_802154_tx_power_get` to get the transmit power, which may cause that when we are calling the function `otPlatRadioSetTransmitPower `to set the output power to A, but the fcuntion `otPlatRadioGetTransmitPower` returns the output power as B. Should we consider this case?

examples/platforms/nrf528xx/src/radio.c
gjc13: The behavior before this change is passed is to use 0 as the implicit default value (all static variables will be zero initialized). I prefer not to change this behavior when the max power table is not passed.

examples/platforms/nrf528xx/src/radio.c
gjc13: As mentioned before, the 0-value `sDefaultPower` doesn't mean we really want to output at 0dBm but just mean we haven't assigned a value to it yet. I add an initialization check to prefer use the max power for better radio quality.

examples/platforms/nrf528xx/src/radio.c
gjc13: This is currently the intended behavior. All the power will be guarded with the max power table.

examples/platforms/nrf528xx/src/radio.c
gjc13: The behavior before this change is to use 0 as the implicit initialization and I prefer not to change it. As long as the max power table is passed, we have a guarantee to use the value in the power table.

brianljt (approved):
LGTM, Thanks!

examples/platforms/nrf528xx/src/radio.c
gjc13: Ack, updated the logic and remove the line of setting `sDefaultTxPower` here. See the new function `GetTransmitPowerForChannel`

examples/platforms/nrf528xx/src/radio.c
gjc13: Done

zhanglongxia (approved):
LGTM

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -107,8 +108,10 @@ static otInstance *  sInstance = NULL;
 static otRadioFrame sAckFrame;
 static bool         sAckedWithFramePending;
 
+static int8_t sMaxTxPowerTable[OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MAX - OT_RADIO_2P4GHZ_OQPSK_CHANNEL_MIN + 1];
 static int8_t sDefaultTxPower;
-static int8_t sLnaGain = 0;
+static int8_t sLnaGain          = 0;
+static bool   sPowerInitialized = false;
```
bukepo, line 114: `sDefaultTxPower` is already used when initialize this module, does it mean it's always treat initialized?

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -144,6 +147,23 @@ static uint32_t        sAckFrameCounter;
...
 #endif
 
+static int8_t GetTransmitPowerForChannel(uint8_t aChannel)
+{
+    int8_t channelMaxPower = nrf5GetChannelMaxTransmitPower(aChannel);
+    int8_t power           = 0; // 0 dbm as default value
+
+    if (sPowerInitialized)
+    {
+        power = OT_MIN(channelMaxPower, sDefaultTxPower);
```
bukepo, line 157: Does `sDefaultTxPower` mean the TX power at the antenna or just the output of the radio chip?

examples/platforms/nrf528xx/src/radio.c
gjc13: This is used to mark whether we've set the power using `otPlatRadioSetTransmitPower`. Otherwise we will use 0dbM as the transmit power though we don't mean to.

examples/platforms/nrf528xx/src/radio.c
gjc13: All the power is meant to be the power at the antenna. If you have FEM/antenna you should modify the output in the function `nrf_802154_fal_tx_power_get`

examples/platforms/nrf528xx/src/radio.c
bukepo: what if `sDefaultTxPower` is set to `OT_RADIO_POWER_INVALID`? How about initializing `sDefaultTxPower` to `OT_RADIO_POWER_INVALID` and making sure `aPower` of  otPlatRadioSetTransmitPower() is not `OT_RADIO_POWER_INVALID`?

gjc13: Done

bukepo (approved):
LGTM:+1:

jwhui (approved):

Issue #5928, opened 2020-12-09
[github-actions] expects RCP mode test fail intermittently


simonlingoogle:
**Describe the bug** 

expects RCP mode test fail intermittently.

It happens quite often recently.

**Console/log output**

[logs_44259.zip](https://github.com/openthread/openthread/files/5663306/logs_44259.zip)

Moandor-y:
Looks like the node crashed. Do we have core dumps for this?

Irving-cl:
@Moandor-y 
I haven't added dumps for that yet. We can add dump uploading as: https://github.com/openthread/openthread/pull/5861.

Irving-cl:
I'll add dumps uploading for expect tests.

Moandor-y:
Happened again after core dump uploading added: <https://github.com/openthread/openthread/pull/5839/checks?check_run_id=1547881295>, but there are no core dumps in the uploaded artifacts. Maybe this is not a crash?

jwhui:
Here is one recent failure:
- [logs_45459.zip](https://github.com/openthread/openthread/files/5690858/logs_45459.zip)
- [core-expect-rcp.zip](https://github.com/openthread/openthread/files/5690860/core-expect-rcp.zip)
- [core-expect-tun.zip](https://github.com/openthread/openthread/files/5690861/core-expect-tun.zip)

jwhui:
Here is another recent failure:
- [logs_45435.zip](https://github.com/openthread/openthread/files/5690867/logs_45435.zip)
- [core-expect-rcp.zip](https://github.com/openthread/openthread/files/5690868/core-expect-rcp.zip)
- [core-expect-tun.zip](https://github.com/openthread/openthread/files/5690869/core-expect-tun.zip)

PR #5965, opened 2020-12-17
[spinel] fix errors in RCP recovery

Moandor-y:
If `mError` is set to error state before entering the RCP recovery, this error will be recognized as an error in the recovery process, and cause it to die, while there is actually no error.

This fixes it by clearing `mError` when starting the recovery process. Fixes <https://github.com/openthread/openthread/issues/5928>.


bukepo (approved):
LGTM:+1:

simonlingoogle (approved):
LGTM 👍

jwhui (approved):

PR #5971, opened 2020-12-18
[posix] disable trel platform when empty/no trel interface name is given

abtink:
This commit changes the trel platform implementation under posix such
that if the interface name used for trel is empty, the platform layer
would disable trel. This is done by having all `otPlatTrel` functions
returning immediately without performing any action. This commit also
changes the `OPENTHREAD_CONFIG_POSIX_APP_TREL_INTERFACE_NAME` to be
empty string. This is the posix config option specifying the interface
name to be used by the trel platform layer. Note that the trel
interface name can be given using this build-time config, or through
the `-trel-interface` (or `-t`) input arg option. When the input arg
is not used, the interface name is determined from the build-time
config `POSIX_APP_TREL_INTERFACE_NAME`. This commit therefore makes
the default behavior when `-t` is not used to disable trel platform.


src/posix/platform/trel_udp6.cpp
```diff
@@ -416,6 +417,10 @@ void otPlatTrelUdp6Init(otInstance *aInstance, const otIp6Address *aUnicastAddre
     int                 val;
     struct sockaddr_in6 sockAddr;
 
+    OT_UNUSED_VARIABLE(aInstance);
```
jwhui, line 420:
```suggestion
    OT_UNUSED_VARIABLE(aInstance);

    int                 val;
    struct sockaddr_in6 sockAddr;
```

jwhui (approved):

PR #5981, opened 2020-12-19
[dns-client] randomize ID of outgoing query

LuDuda:


abtink (approved):
LGTM. Thanks 👍

src/core/net/dns_client.cpp
```diff
@@ -199,6 +198,23 @@ void Client::SendCopy(const Message &aMessage, const Ip6::MessageInfo &aMessageI
     }
 }
 
+uint16_t Client::GenerateUniqueRandomId(void)
+{
+    uint16_t randomId;
+    otError  error;
+
+    OT_UNUSED_VARIABLE(error);
```
konradderda, line 206: 'error' is always used

src/core/net/dns_client.cpp
```diff
@@ -199,6 +198,23 @@ void Client::SendCopy(const Message &aMessage, const Ip6::MessageInfo &aMessageI
...
+
+    do
+    {
+        error = Random::Crypto::FillBuffer(reinterpret_cast<uint8_t *>(&randomId), sizeof(randomId));
+        OT_ASSERT(error == OT_ERROR_NONE);
```
konradderda, line 211: Is generating a DNS transaction id so critical that we should assert here? Crypto sources (library or other core) may need to gather more entropy or be blocked by another request. This could lead to random failures. Maybe retry mechanism (could be added at crypto layer) or reporting an error during and dropping the query would be better?

src/core/net/dns_client.cpp
```diff
@@ -346,7 +361,8 @@ void Client::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessag
     aMessage.MoveOffset(sizeof(responseHeader));
     offset = aMessage.GetOffset();
 
-    VerifyOrExit((message = FindRelatedQuery(responseHeader, queryMetadata)) != nullptr);
+    VerifyOrExit((message = FindQueryById(responseHeader.GetMessageId())) != nullptr);
+    queryMetadata.ReadFrom(*message);
```
konradderda, line 365: 'queryMetadata.ReadFrom(*message)' can be moved inside for loop few lines later, just before the following call (where it is needed): `FinalizeDnsTransaction(*message, queryMetadata, &record.GetAddress(), record.GetTtl(), OT_ERROR_NONE);`

src/core/net/dns_client.cpp
LuDuda: This is needed, because of `OT_ASSERT` macro, which is empty in some of GitHub Actions check.

konradderda: Then maybe it would be better to create a separate PR and change empty OT_ASSERT implementation for such cases to:
`#define OT_ASSERT(cond) (void)(cond);`

This way such OT_UNUSED_VARIABLE would not be needed. Thoughts?

src/core/net/dns_client.cpp
LuDuda: Current DNS Client implementation facilitates DNS cache poisoning attacks by implementing deterministic way of assigning IDs (just increment ID by one).

One option to remove assert would be to use PRNG, however, the attacker may send a series of fake replies with several ID numbers trying to guess the proper one. If ID is generated with poor Pseudo-Random Number Generator, attacker has higher probability of success. While i personally didn't do any calculation with use of our PRNG, there are few papers in the internet suggesting to use TRNG mechanism.

`Crypto` RNG module is used without error checking in few places in OpenThread, in other places assertion like here is implemented. I would say, that we can improve it in subsequent PR, for example by implementing a backup mechanism to use `NonCrypto` module in case `Crypto` fails. I believe simple retry mechanism for `Crypto` may not work always, and in some situation may lead to infinite loop. If that does not work, I can re-write the function to return error code. @abtink any suggestion?

src/core/net/dns_client.cpp
LuDuda: The `queryMetadata` is used also within `exit` label. Of course we could add `queryMetadata.ReadFrom(*message);` in two places before `FinalizeDnsTransaction`, but that would make sense only, if `queryMetadata.ReadFrom(*message);` was not needed in some of the flow, however i don't see such case i.e. after the Query message is found (line 364), `queryMetadata` has to be always read from the message.

src/core/net/dns_client.cpp
LuDuda: I don't have a strong opinion here, it's quite common to handle such cases in a place where assert is used. I think the `(void)(cond)` won't work, we could use `OT_UNUSED_VARIABLE` macro instead (to comply with all compilers). Let me resolve this conversation, feel free to file a PR with such improvement. Thanks!

src/core/net/dns_client.cpp
konradderda: Right. My comment was invalid.

src/core/net/dns_client.cpp
konradderda: > implementing a backup mechanism to use NonCrypto module in case Crypto fails
We should not substitute the results of `Crypto` functions with `NonCrypto` ones since this may lead to security vulnerabilities. 

Additionally, OT_ASSERT may be disabled. In such a case the current implementation of this particular function (and maybe in some other places as you pointed out) will just ignore an error and return some bytes but not necessarily random.

konradderda: It looks like a topic worth discussing in a separate ticket.

LuDuda: @konradderda in my last commit i removed assert and return error code now instead. Please have a look.

Irving-cl (approved):
LGTM :+1:

src/core/net/dns_client.cpp
```diff
@@ -199,6 +201,20 @@ void Client::SendCopy(const Message &aMessage, const Ip6::MessageInfo &aMessageI
...
 }
 
+otError Client::GenerateUniqueRandomId(uint16_t &aRandomId)
+{
+    otError error;
+
+    do
+    {
+        SuccessOrExit(error = Random::Crypto::FillBuffer(reinterpret_cast<uint8_t *>(&aRandomId), sizeof(aRandomId)));
+
```
konradderda, line 211: super nit: unnecessary empty line

konradderda (approved):
LGTM, thanks.

jwhui (approved):
Thanks! 👍🏼

src/core/net/dns_client.cpp
abtink: As @LuDuda mentioned the `(void)(cond)` would not work on all tool-chains. Note that we have different toolchain specific implementation for `OT_UNUSED_VARIABLE()` macro). Not if the `OT_UNUSED_VARIBLE()` can be used in `OT_ASSERT()` since the `cond` may not be a variable itself (some of the implementations of `OT_UNUSED_VARIBLE()` rely on the input param to be a variable and they use its address).

PR #5986, opened 2020-12-21
[net] initial implementation of the SRP server

wgtdkp:
This PR includes the initial implementation of the SRP server defined in https://tools.ietf.org/html/draft-ietf-dnssd-srp-07.

- [X] SRP server Service TLV propagation
  - Includes only the 2-bytes dynamic listening port in the Server TLV.
- [X] SRP request processing (parsing & validation)
- [X] LEASE & KEY-LEASE management
- [X] SIG(0) verification
- [X] Interface for advertising proxy
- [X] CLI commands for testing
- [X] Documentation
- [X] Support removing indivitual SRP service
  - Not verified, we need the OT SRP client to help verify this feature since apple currently doesn't support this.
- [X] CI tests: https://github.com/openthread/openthread/pull/6074

Manually tested with Apple's implementation.


src/core/border_router/srp_server.cpp
```diff
@@ -0,0 +1,1451 @@
...
+#include "common/new.hpp"
+#include "net/dns_headers.hpp"
+#include "thread/network_data_local.hpp"
+#include "thread/network_data_notifier.hpp"
+#include "thread/thread_netif.hpp"
+#include "utils/heap.hpp"
+
+namespace ot {
+
+namespace BorderRouter {
```
jwhui, line 50: SRP server does not have to be tied to a Thread Border Router. Shall we move this into `src/core/net`?

include/openthread/srp.h
```diff
@@ -0,0 +1,212 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ * @brief
+ *  This file defines the API for Service Registration Protocol (SRP).
+ */
+
+#ifndef OPENTHREAD_SRP_H_
```
jwhui, line 35: Do we envision APIs for both client and server? If so, should we rename to `srp-server.h`?

src/core/border_router/srp_server.cpp
wgtdkp: SGTM, will do.

include/openthread/srp.h
wgtdkp: Yes, looks like SRP client will have separate, independent API (no `struct` definitions are shared between client & server). Will rename it.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1515 @@
...
+
+otError SrpServer::PublishService()
+{
+    otError       error;
+    const uint8_t serviceData[] = {kThreadServiceTypeSrpServer};
+    uint8_t       serverData[sizeof(uint16_t)];
+
+    OT_ASSERT(mSocket.IsBound());
+
+    *reinterpret_cast<uint16_t *>(serverData) = Encoding::BigEndian::HostSwap16(mSocket.GetSockName().mPort);
```
abtink, line 436: Use `Encoding:BigEndian::WriteUint16()` instead. 

In general, casting a pointer to a buffer as `uint16_t *` and then reading/writing to it is not safe (will not behave correctly on all toolchains). Note that the `serverData` being a byte array has no alignment requirement and can start from any memory address, but `uint16_t` may have alignment requirement.

src/core/common/instance.hpp
```diff
@@ -862,6 +862,13 @@ template <> inline BorderRouter::RoutingManager &Instance::Get(void)
 }
 #endif
 
+#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
+template <> inline Srp::SrpServer &Instance::Get(void)
```
abtink, line 866: Suggest renaming to `Srp::Server` instead of `Srp::SrpServer`.

This harmonizes it with pattern used by other net protocol implementations in OT: `Dns::Clinet`, `Dhcp6::Client`, `Dhcp6::Server`, and `Sntp::Client`.

abtink (commented):
Thanks @wgtdkp for the PR. Overall looks great. 

While testing it with the client implementation, I found 3 main things that I added below.

(have not had the chance to read the whole code in this PR. Will read it later :))..

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1500 @@
...
+            VerifyOrExit(record.GetType() == Dns::ResourceRecord::kTypeAny && record.GetTtl() == 0 &&
+                             record.GetLength() == 0,
+                         error = OT_ERROR_FAILED);
+            SuccessOrExit(error = HandleDeleteAllResources(aHost, name));
+
+            aOffset += record.GetSize();
+            continue;
+        }
+
+        VerifyOrExit(record.GetClass() == aZone.GetClass(), error = OT_ERROR_FAILED);
```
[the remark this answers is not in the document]
wgtdkp: thanks @abtink , I think I forgot to update the check because I was not able to test the new feature of removing individual service.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1500 @@
...
+    Crypto::Sha256                 sha256;
+    Crypto::Sha256::Hash           hash;
+    Crypto::Ecdsa::P256::Signature signature;
+
+    VerifyOrExit(aSigRdataLength >= Crypto::Ecdsa::P256::Signature::kSize, error = OT_ERROR_INVALID_ARGS);
+
+    sha256.Start();
+
+    // SIG RDATA less signature.
+    sha256.Update(aMessage, aSigRdataOffset, aSigRdataLength - Crypto::Ecdsa::P256::Signature::kSize);
```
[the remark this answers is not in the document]
wgtdkp: I was struggling with it and tried to verify with uncompressed signer name but put it down to see if you have different interpreting. So it seems we should sign & verify with uncompressed form :)

I will make the change soon.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1500 @@
...
+            Dns::SrvRecord srvRecord;
+            Service *      service;
+            char           hostName[Dns::Name::kMaxLength + 1];
+            uint16_t       hostNameLength = sizeof(hostName);
+
+            SuccessOrExit(error = aMessage.Read(aOffset, srvRecord));
+            aOffset += sizeof(srvRecord);
+
+            SuccessOrExit(error = Dns::Name::ReadName(aMessage, aOffset, aHeaderOffset, hostName, hostNameLength));
+            VerifyOrExit(aHost.Matches(hostName), error = OT_ERROR_FAILED);
```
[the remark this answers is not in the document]
wgtdkp: You are right, I think to properly handle this case, we need to process the Host Description before Service Descriptions.

src/core/net/srp_server.cpp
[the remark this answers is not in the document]
abtink: To be honest, I would like/prefer if the wire format from message would be directly used for sig calculation (whether compressed or not). In my view, it would be simpler and easier to implement (in both server and client sides).

I think RFC-2335 requirs uncompressed name for sig calculation. I was hoping SIG(0) RFC-2931 would override that but from my reading/understanding of it, it seems not to.  I noticed that the implementation here [Abhayakara/mdnsresponder](https://github.com/Abhayakara/mdnsresponder/blob/main/ServiceRegistration/towire.c#L507) also seems to use uncompressed version for sig calculation but then append it as compressed.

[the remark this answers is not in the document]
wgtdkp: > To be honest, I would like/prefer if the wire format from message would be directly used for sig calculation (whether compressed or not). In my view, it would be simpler and easier to implement (in both server and client sides).
>
> I think RFC-2335 requirs uncompressed name for sig calculation. I was hoping SIG(0) RFC-2931 would override that but from my reading/understanding of it, it seems not to. 

I share the same idea and this is what I was struggling with. I guess the reason of using uncompressed form of the signer name is to include the actual signer name to produce the signature. For messages that are signed with selected RRs (covered-type), the actual signer name may not be included for creating the signature if uncompressed form is used. But this is not a problem for request signature, because it uses the request message starting from DNS header and the actual signer name will always be included even if uncompressed form is used. I think RFC-2931 is trying to be compatible with RFC 2535 on this portion.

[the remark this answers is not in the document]
jwhui: I think references to "RFC 2335" above should be "RFC 2535".

Yes, RFC 2535 specifies the canonical form to ensure that signature calculation is well-defined, especially in the presence of caching DNS servers. RFC 2931 leverages existing mechanisms and hence maintains usage of the canonical form.

include/openthread/srp_server.h
```diff
@@ -0,0 +1,212 @@
...
+ *                            service when it is not deleted (indicated by `otSrpServerService::mIsDeleted`)
+ *                            and un-publish it when deleted.
+ * @param[in]  aTimeout  The maximum time in milliseconds for the handler to process the service event.
+ * @param[in]  aContext  A pointer to application-specific context.
+ *
+ * @sa otSrpServerSetAdvertisingHandler
+ * @sa otSrpServerHandleAdvertisingResult
+ *
+ */
+typedef void (*otSrpServerAdvertisingHandler)(const otSrpServerHost *aHost, uint32_t aTimeout, void *aContext);
```
abtink, line 128: I noticed that the API callback definition is mentioning "Advertising Proxy" as the intended receiver of this callback and that what the user should do in callback (e.g. "should publish update with mDNS"). I understand that this would be the common way the SRP server would be used, but thinking more abstractly, should we try to describe SRP server callback reports. Basically SRP server should be more abstract and not aware of who/what entity is using its API (i.e. whether its an Advertising Proxy or not). 

I would suggest describing what it reports and what we expect:
- This callback is notify the user that there is some update requests for a given host (`aHost`).
- The user should then process the info (check for name conflicts, etc). 
- We expect the user then to call another server API to indicate the result of its processing of the SRP update request.
- We can then give more details as example if the user is an advertising proxy (e.g., it would publish it, etc).

Does this make sense? Thoughts?

include/openthread/srp_server.h
```diff
@@ -0,0 +1,212 @@
...
+ *
+ * @brief
+ *  This module includes functions to control service registrations.
+ *
+ * @{
+ *
+ */
+
+/**
+ * This structure represents a SRP service host.
```
abtink, line 58:
```suggestion
 * This opaque type represents an SRP service host.
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+        HandleAdvertisingResult(mOutstandingUpdates.GetTail(), OT_ERROR_RESPONSE_TIMEOUT);
+    }
+}
+
+Server::Service *Server::Service::New(const char *aFullName)
+{
+    void *   buf;
+    Service *service = nullptr;
+
+    buf = Instance::Get().HeapCAlloc(1, sizeof(Service));
```
abtink, line 1190: The `Instance::Get()` is only available in single instance case.

FWIW, we still support/have the multi-instance case within OT, so would be good to have SRP server support it as well.
Here is my suggestion:
- You can pass `instance` as paramter to `New()`
- Make the `Servcie` and `Host` be `InstanceLocator`. 
- Pass the `instance` in `new (buf) Service(instance);`

src/core/config/openthread-core-config-check.h
```diff
@@ -518,6 +518,24 @@
...
+#if !OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE
+#error "OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE is required for OPENTHREAD_CONFIG_SRP_SERVER_ENABLE"
+#endif
+
+#if !OPENTHREAD_CONFIG_ECDSA_ENABLE
+#error "OPENTHREAD_CONFIG_ECDSA_ENABLE is required for OPENTHREAD_CONFIG_SRP_SERVER_ENABLE"
+#endif
+
+#if !OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
+#error "OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE is required for OPENTHREAD_CONFIG_SRP_SERVER_ENABLE"
```
abtink, line 535: I know we are using heap for items in SRP server, but do we need external heap for it? 
I think we can support it with internal heap as well. Maybe we need  a check on heap size?

src/core/config/openthread-core-config-check.h
```diff
@@ -518,6 +518,24 @@
 
 #endif // OPENTHREAD_CONFIG_BORDER_ROUTING_ENABLE
 
+#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
```
abtink, line 521: It's good to have thee checks for config.

My suggestion is to add them in `srp_server.hpp` header itself. I feel it may be better place for them (close to the definition and implementation). This header file started as a place to track configs that are removed or renamed.

Thoughts?

src/core/net/dns_headers.hpp
```diff
@@ -341,6 +343,15 @@ class Header : public Clearable<Header>
...
     void SetAdditionalRecordsCount(uint16_t aCount) { mArCount = HostSwap16(aCount); }
 
+#define GetZoneCount GetQuestionCount
+#define SetZoneCount SetQuestionCount
+#define GetPrerequisiteCount GetAnswerCount
+#define SetPrerequisiteCount SetAnswerCount
+#define GetUpdateCount GetAuthorityRecordsCount
+#define SetUpdateCount SetAuthorityRecordsCount
+#define GetAdditionalDataCount GetAdditionalRecordsCount
+#define SetAdditionalDataCount SetAdditionalRecordsCount
```
abtink, line 353: Interesting defines 😃 

In client implementation I defined `UpdateHeader` as a sub-class of `Header` and defined these methods in it. 

I noticed that you included and squashed the DNS header related changes from the early branch https://github.com/abtink/openthread/commits/dns/new-rrs (the DNS header  RR changes) into the main server commit.

I included it in the client implementation as a seperate commit. There is also one for adding the log under SRP region.

I suggest rebasing the server code on top of those two commits, so we share the same common definitions and commit.  Thanks.

abtink (commented):
Thanks @wgtdkp.
I am slowing reading through the code. Overall looks great and reads easily.
A bunch of smaller things/suggestions below. Thanks.

src/core/net/dns_headers.hpp
wgtdkp: @abtink  Thanks for the DNS records commits :+1: , will update the server to use the latest implementation.

> In client implementation I defined UpdateHeader as a sub-class of Header and defined these methods in it.

Will it increases code size? I was trying to avoid creating another class which has no custom logic.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+    else
+    {
+        otLogInfoSrp("server: sent success response");
+    }
+
+exit:
+    if (error != OT_ERROR_NONE)
+    {
+        otLogWarnSrp("server: failed to send response: %s", otThreadErrorToString(error));
+        FreeMessageOnError(response, error);
```
simonlingoogle, line 970:
```suggestion
        FreeMessage(response);
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+
+    SuccessOrExit(error = mSocket.SendTo(*response, aMessageInfo));
+
+    otLogInfoSrp("server: sent response with granted lease: %u and key lease: %u", aLease, aKeyLease);
+
+exit:
+    if (error != OT_ERROR_NONE)
+    {
+        otLogWarnSrp("server: failed to send response: %s", otThreadErrorToString(error));
+        FreeMessageOnError(response, error);
```
simonlingoogle, line 1009:
```suggestion
        FreeMessage(response);
```

src/core/net/srp_server.hpp
```diff
@@ -0,0 +1,800 @@
...
+         *
+         */
+        bool Matches(const char *aName) const;
+
+    private:
+        Host(void);
+        ~Host(void);
+
+        char *       mFullName;                    // The full hostname.
+        Ip6::Address mAddresses[kMaxAddressesNum]; // The address list to be advertised.
```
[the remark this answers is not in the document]
wgtdkp: It is more complicate to manage a list of dynamically allocated objects.

src/core/common/string.cpp
```diff
@@ -65,6 +65,17 @@ const char *StringFind(const char *aString, char aChar)
     return ret;
 }
 
+bool StringStartsWith(const char *aString, const char *aPrefix)
+{
+    while (aString[0] != '\0' && aPrefix[0] != '\0' && aString[0] == aPrefix[0])
```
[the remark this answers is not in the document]
wgtdkp: The method is not used after refactoring, removed it.

src/posix/platform/system.cpp
```diff
@@ -70,6 +72,10 @@ otInstance *otSysInit(otPlatformConfig *aPlatformConfig)
     otInstance *        instance = nullptr;
     ot::Posix::RadioUrl radioUrl(aPlatformConfig->mRadioUrl);
 
+#if OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
+    otHeapSetCAllocFree(calloc, free);
```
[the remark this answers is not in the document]
wgtdkp: Sure.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+    mFullName       = nullptr;
+    mIsDeleted      = false;
+    mPriority       = 0;
+    mWeight         = 0;
+    mPort           = 0;
+    mTxtLength      = 0;
+    mTxtData        = nullptr;
+    mHost           = nullptr;
+    mNext           = nullptr;
+    mTimeLastUpdate = TimerMilli::GetNow();
```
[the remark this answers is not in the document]
wgtdkp: They are fields of `otSrpService` but not `Srp::Service`.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+}
+
+Server::Service::~Service(void)
+{
+    Instance::Get().HeapFree(mFullName);
+    Instance::Get().HeapFree(mTxtData);
```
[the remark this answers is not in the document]
wgtdkp: I don't think it is meaningful to reset the member fields in the destructor since the object is being destroyed.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+exit:
+    return error;
+}
+
+TimeMilli Server::Service::GetExpireTime(void) const
+{
+    OT_ASSERT(!IsDeleted());
+    OT_ASSERT(!GetHost().IsDeleted());
+
+    return mTimeLastUpdate + GetHost().GetLease() * 1000;
```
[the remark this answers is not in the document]
wgtdkp: I think we need change the `Time` to `uint64_t`. @abtink Thoughts?

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+{
+    OT_ASSERT(aFullName != nullptr);
+
+    otError error    = OT_ERROR_NONE;
+    char *  nameCopy = static_cast<char *>(Instance::Get().HeapCAlloc(1, strlen(aFullName) + 1));
+
+    VerifyOrExit(nameCopy != nullptr, error = OT_ERROR_NO_BUFS);
+    strcpy(nameCopy, aFullName);
+
+    if (mFullName != nullptr)
```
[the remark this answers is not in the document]
wgtdkp: We assert on `aFullName` but not `mFullName`.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+Server::Service *Server::Host::FindService(const char *aFullName)
+{
+    return mServices.FindMatching(aFullName);
+}
+
+otError Server::Host::AddIp6Address(const Ip6::Address &aIp6Address)
+{
+    otError error = OT_ERROR_NONE;
+
+    if (aIp6Address.IsMulticast() || aIp6Address.IsUnspecified() || aIp6Address.IsLoopback() ||
```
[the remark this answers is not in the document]
wgtdkp: For now, we don't enforce the address to be an OMR prefix to not bind the SRP server to the Border Routing module. We can reject a non-OMR address in the Advertising Proxy if it thinks the address is not suitable to be advertised on the infra link.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+        aIp6Address.IsLinkLocal())
+    {
+        // We don't like those address because they cannot be used for communication
+        // with exterior devices.
+        ExitNow(error = OT_ERROR_NONE);
```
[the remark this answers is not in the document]
wgtdkp: Good catch, it is a typo.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+                                   uint16_t &         aOffset,
+                                   Dns::Zone &        aZone)
+{
+    otError   error = OT_ERROR_NONE;
+    Dns::Zone zone;
+
+    VerifyOrExit(aDnsHeader.GetZoneCount() == 1, error = OT_ERROR_FAILED);
+
+    SuccessOrExit(Dns::Name::ParseName(aMessage, aOffset));
+    SuccessOrExit(aMessage.Read(aOffset, zone));
```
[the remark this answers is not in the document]
wgtdkp: The idea is that we should not touch the output unless we are done all jobs and sure that it will return `OT_ERROR_NONE`.

src/core/config/openthread-core-config-check.h
wgtdkp: Agree that requiring the external heap is too strict, will remove it.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+}
+
+void Server::HandleOutstandingUpdatesTimer(Timer &aTimer)
+{
+    aTimer.GetOwner<Server>().HandleOutstandingUpdatesTimer();
+}
+
+void Server::HandleOutstandingUpdatesTimer(void)
+{
+    while (!mOutstandingUpdates.IsEmpty() && mOutstandingUpdates.GetTail()->GetExpireTime() <= TimerMilli::GetNow())
```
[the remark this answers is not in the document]
wgtdkp: We add new updates at the head, so the tail expires first (since all updates shares the same timeout value).

src/core/net/srp_server.cpp
wgtdkp: SGTM, updated.

src/core/config/openthread-core-config-check.h
wgtdkp: I see 1.2 feature checks there, so followed it. But having individual modules checks by themselves sounds good to me. @jwhui @simonlingoogle Thoughts?

src/core/net/dns_headers.hpp
wgtdkp: updated.

src/cli/README_SRP_SERVER.md
```diff
@@ -0,0 +1,91 @@
...
+Enable the SRP server.
+
+```bash
+> srpserver enable
+Done
+```
+
+### host
+
+Usage: `srpserver host`
```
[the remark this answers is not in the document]
wgtdkp: Not feeling necessary as I don't see practice of such commands in OT CLI.

src/cli/README_SRP_SERVER.md
```diff
@@ -0,0 +1,91 @@
...
+srp-api-test-1.default.service.arpa.
+    addresses: fdde:ad00:beef:0:0:ff:fe00:fc10
+srp-api-test-0.default.service.arpa.
+    addresses: fdde:ad00:beef:0:0:ff:fe00:fc10
+Done
+```
+
+### srpserver service
+
+Usage: `srpserver service`
```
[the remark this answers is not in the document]
wgtdkp: ditto.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+    SuccessOrExit(error = aMessage.Read(offset, dnsHeader));
+    offset += sizeof(dnsHeader);
+
+    // Handles only queries.
+    VerifyOrExit(dnsHeader.GetType() == Dns::Header::Type::kTypeQuery, error = OT_ERROR_DROP);
+
+    switch (dnsHeader.GetQueryType())
+    {
+    case Dns::Header::kQueryTypeUpdate:
+        HandleDnsUpdate(aMessage, aMessageInfo, dnsHeader, offset);
```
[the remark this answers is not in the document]
wgtdkp: This adds implicit assumption on the methods arguments `aMessage` that its offset has been correctly set to desired value. I would prefer explicitly passing the offset to make it clear for the reader.

src/core/net/dns_headers.hpp
abtink: I don't expect any code increase with `UpdateHeader`. It adds no member variables and all its methods are inlined.

src/cli/cli.hpp
```diff
@@ -748,6 +753,9 @@ class Interpreter
         {"singleton", &Interpreter::ProcessSingleton},
 #if OPENTHREAD_CONFIG_SNTP_CLIENT_ENABLE
         {"sntp", &Interpreter::ProcessSntp},
+#endif
+#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
+        {"srpserver", &Interpreter::ProcessSrpServer},
```
abtink, line 758: In client code I went with `srp client` as the command (`srp` as the main command and then `clinet` as sub-command and was thinking `srp server` for server). Here we have `srpserver` as a full command name.

We should converge/decide on this (make them similar). I like the `srp clinet` and `srp server` model a little bit more, but it is not a strong preference.

Thoughts? Suggestion? @wgtdkp @jwhui others?

src/core/net/srp_server.cpp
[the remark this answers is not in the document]
simonlingoogle: Oh, I see. LinkededList.Push adds the element to the Head of the list (which I think is strange).

src/cli/cli.hpp
wgtdkp: I was thinking we should use independent CLI commands if we have separate public headers `srp_server.h` & `srp_client.h`, but having client & server as sub-commands of `srp` also makes sense to me 😃 .

src/core/net/srp_server.hpp
```diff
@@ -0,0 +1,800 @@
...
+ * This class implements the SRP server.
+ *
+ */
+class Server : public InstanceLocator, private NonCopyable
+{
+    friend class ot::Notifier;
+
+public:
+    class Host;
+    class Service;
```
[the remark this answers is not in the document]
wgtdkp: done.

src/core/net/srp_server.hpp
```diff
@@ -0,0 +1,800 @@
...
+        /**
+         * This method sets the full service name.
+         *
+         * @param[in]  aFullName  The full service name.
+         *
+         * @retval  OT_ERROR_NONE     Successfully set the full service name.
+         * @retval  OT_ERROR_NO_BUFS  Failed to allocate memory for the name.
+         *
+         */
+        otError SetFullName(const char *aFullName);
```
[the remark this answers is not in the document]
wgtdkp: SGTM, updated.

include/openthread/srp_server.h
wgtdkp: I just renamed it from `otSrpServerUpdateHandler`(or `otSrpServerEventHandler`), but it still makes sense to me to have a more generic name.

wgtdkp: renamed to `otSrpServerServiceUpdateHandler`.

include/openthread/srp_server.h
```diff
@@ -0,0 +1,212 @@
...
+ */
+typedef void otSrpServerHost;
+
+/**
+ * This structure represents a SRP service.
+ *
+ */
+struct otSrpServerService
+{
+    char *              mFullName;  ///< The full service name.
```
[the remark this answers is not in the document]
wgtdkp: I think the user is stopped from changing the service content by returning `const otSrpService *`.

I did some refactoring and this structure removed by providing getters (e.g. `otSrpServeverServiceGetFullName()`) just as we do for `otSrpServerHost`.

src/cli/cli_srp_server.cpp
```diff
@@ -0,0 +1,222 @@
+/*
+ *  Copyright (c) 2019, The OpenThread Authors.
```
simonlingoogle, line 2:
```suggestion
 *  Copyright (c) 2020, The OpenThread Authors.
```

src/core/net/srp_server.cpp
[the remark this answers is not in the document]
wgtdkp: limited the lease value to `TimerMilli::kMaxDelay / 1000`

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+{
+    mKeyLease       = aKeyLease;
+    mTimeLastUpdate = TimerMilli::GetNow();
+}
+
+TimeMilli Server::Host::GetExpireTime(void) const
+{
+    OT_ASSERT(!IsDeleted());
+
+    return mTimeLastUpdate + mLease * 1000;
```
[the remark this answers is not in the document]
wgtdkp: limited the lease value to `TimerMilli::kMaxDelay / 1000`.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+}
+
+TimeMilli Server::Host::GetKeyExpireTime(void) const
+{
+    return mTimeLastUpdate + mKeyLease * 1000;
```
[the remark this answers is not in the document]
wgtdkp: limited the lease value to `TimerMilli::kMaxDelay / 1000`.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+        otLogInfoSrp("server: drops duplicated SRP update request: messageId=%hu", aDnsHeader.GetMessageId());
+
+        // Silently drop duplicate requests.
+        // This could rarely happen, because the outstanding SRP update timer should
+        // be much shorter than the SRP update retransmission timer.
+        ExitNow(error = OT_ERROR_NONE);
+    }
+
+    // Per 2.3.2 of SRP draft 6, no prerequisites should be included in a SRP update.
+    VerifyOrExit(aDnsHeader.GetPrerequisiteCount() == 0, error = OT_ERROR_FAILED);
```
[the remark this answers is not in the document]
wgtdkp: https://tools.ietf.org/html/draft-ietf-dnssd-srp-08#section-2.3.2

> A DNS Update that contains any other adds,
   any other deletes, or any prerequisites, is not an SRP update.  Such
   messages should either be processed as regular RFC2136 updates,
   including access control checks and constraint checks, if supported,
   or else rejected with RCODE=REFUSED.

 We map `OT_ERROR_FAILED` to `REFUSED`.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+}
+
+void Server::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessageInfo)
+{
+    otError     error;
+    Dns::Header dnsHeader;
+    uint16_t    offset = aMessage.GetOffset();
+
+    SuccessOrExit(error = aMessage.Read(offset, dnsHeader));
+    offset += sizeof(dnsHeader);
```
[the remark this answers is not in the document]
wgtdkp: it adds the implicit assumption on the `aMessage` parameter that its offset has already been set to the proper value. Instead, I will prefer using an offset parameter to make this clear.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1577 @@
...
+    Dns::Zone zone;
+    Host *    host = nullptr;
+
+    uint16_t headerOffset = aOffset - sizeof(aDnsHeader);
+
+    otLogInfoSrp("server: received DNS update from %s", aMessageInfo.GetPeerAddr().ToString().AsCString());
+
+    SuccessOrExit(error = ProcessZoneSection(aMessage, aDnsHeader, aOffset, zone));
+
+    if (mOutstandingUpdates.FindMatching(aDnsHeader.GetMessageId()) != nullptr)
```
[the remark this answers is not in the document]
wgtdkp: updated to match outstanding updates with MessageInfo and DNS message ID.

src/core/net/srp_server.hpp
```diff
@@ -0,0 +1,601 @@
...
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes definitions for SRP server.
+ */
+
+#ifndef BORDER_ROUTER_SRP_SERVER_HPP_
```
simonlingoogle, line 34: Note that `border_router` is not the parent directory of this header. 

I would suggest that we create new directory: `src/core/service_discovery` move put srp_server.c/hpp there. 

```suggestion
#ifndef SERVICE_DISCOVERY_SRP_SERVER_HPP_
```

include/openthread/srp_server.h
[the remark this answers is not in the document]
abtink: A `const otSrpService *` would not protect against access to the pointed char buffer.

In ` const` structure, the member variables would be treated as `const`, so no one can change or assign new values to them. The `mFullName` would be treated as `char * const mFullName` (which means the `mFullName` pointer value itself cannot change but the buffer to which it points can be changed) and not `const char *const mFullName`.

Here is small test for this:
```
    struct Test
    {
        char *mName;
    };

    char name[2];
    char name2[2];
    Test test;
    const Test *tp = &test;

    test.mName = name;
    name[0] = 'A';
    name[1] = 0;

    printf("name:\"%s\"\n", tp->mName);

    tp->mName[0] = 'B';    // This works even though `tp` is `const Test *`
    printf("name:\"%s\"\n", tp->mName);

    // tp->mName = name2; // This errors "cannot assign to variable 'tp' with const-qualified type 'const Test *'"
```

src/core/common/logging.hpp
```diff
@@ -95,6 +95,7 @@ extern "C" {
 #define _OT_REGION_MLR_PREFIX "-MLR-----: "
 #define _OT_REGION_DUA_PREFIX "-DUA-----: "
 #define _OT_REGION_BR_PREFIX "-BR------: "
+#define _OT_REGION_SRP_PREFIX "-SRP-----: "
```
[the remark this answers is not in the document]
abtink: We have it as a seperate commit in  #6038. @wgtdkp has also rebased the server code on of top of that.

include/openthread/srp_server.h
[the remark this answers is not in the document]
abtink: The new model with opaque `Service` and getter APIs look good. Thank. I think it may add a bit to code size but for server code not sure if it matters much.

simonlingoogle:
@wgtdkp 
Would you add SRP_SERVER=1 to check-size so that we can see the ROM and RAM cost of this PR?

wgtdkp:
> @wgtdkp
> Would you add SRP_SERVER=1 to check-size so that we can see the ROM and RAM cost of this PR?

I don't feel that information will be useful as we allocate significant internal heap for SRP server.

simonlingoogle:
> > @wgtdkp
> > Would you add SRP_SERVER=1 to check-size so that we can see the ROM and RAM cost of this PR?
> 
> I don't feel that information will be useful as we allocate significant internal heap for SRP server.

The size report not only gives us the information the ROM and RAM size of the code, it also make sure that the PR does not accidentally  add unexpected sizes.
So I think we should use `check-size` as long as it's possible. 
Thoughts?

include/openthread/srp_server.h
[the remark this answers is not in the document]
wgtdkp: Thanks for verifying this, I was not aware of this behavior.

src/core/config/openthread-core-config-check.h
wgtdkp: Moved the check to `srp_server.hpp`.

src/core/common/logging.hpp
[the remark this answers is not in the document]
wgtdkp: rebased the implementation.

src/cli/cli.hpp
wgtdkp: @jwhui Do you have a preference?

src/core/net/srp_server.hpp
wgtdkp: Thanks for catching this. I don't think it is necessary to create a `service_discovery` directory for SRP server. For now I will rename it to `NET_SRP_SERVER_HPP_`. A `service_discovery` directory may be necessary when Discovery Proxy and more DNS module are introduced to OT.

wgtdkp:
> > > @wgtdkp
> > > Would you add SRP_SERVER=1 to check-size so that we can see the ROM and RAM cost of this PR?
> > 
> > 
> > I don't feel that information will be useful as we allocate significant internal heap for SRP server.
> 
> The size report not only gives us the information the ROM and RAM size of the code, it also make sure that the PR does not accidentally add unexpected sizes.
> So I think we should use `check-size` as long as it's possible.
> Thoughts?

Added `SRP_SERVER=1` it to `size-check`

src/cli/cli.hpp
wgtdkp: Updated the command to `srp server`.

script/test
```diff
@@ -53,6 +53,9 @@ build_simulation()
     local version="$1"
     local options=("-DOT_MESSAGE_USE_HEAP=ON" "-DOT_THREAD_VERSION=${version}" "-DBUILD_TESTING=ON" "-DOT_REFERENCE_DEVICE=ON")
 
+    # Always enbale SRP server & client.
+    options+=("-DOT_SRP_SERVER=ON -DOT_SRP_CLIENT=ON -DOT_SERVICE=ON -DOT_ECDSA=ON")
```
simonlingoogle, line 57:
```suggestion
    local options=("-DOT_MESSAGE_USE_HEAP=ON" "-DOT_THREAD_VERSION=${version}" "-DBUILD_TESTING=ON" "-DOT_REFERENCE_DEVICE=ON" "-DOT_SRP_SERVER=ON -DOT_SRP_CLIENT=ON -DOT_SERVICE=ON -DOT_ECDSA=ON")
```

src/core/net/srp_server.hpp
```diff
@@ -0,0 +1,613 @@
...
+{
+    friend class ot::Notifier;
+
+public:
+    class Host;
+    class Service;
+
+    enum : uint8_t
+    {
+        kMaxAddressesNum = 8u, ///< The maximum number of addresses for a host.
```
simonlingoogle, line 81: move to private?

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1604 @@
...
+
+bool Server::IsRunning(void) const
+{
+    return mSocket.IsBound();
+}
+
+void Server::SetEnabled(bool aEnabled)
+{
+    VerifyOrExit(mEnabled != aEnabled);
+    mEnabled = aEnabled;
```
simonlingoogle, line 106:
```suggestion
    VerifyOrExit(mEnabled != aEnabled);

    mEnabled = aEnabled;
```
nit

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1604 @@
...
+
+otError Server::ProcessZoneSection(const Message &          aMessage,
+                                   const Dns::UpdateHeader &aDnsHeader,
+                                   uint16_t &               aOffset,
+                                   Dns::Zone &              aZone)
+{
+    otError   error = OT_ERROR_NONE;
+    Dns::Zone zone;
+
+    VerifyOrExit(aDnsHeader.GetZoneRecordCount() == 1, error = OT_ERROR_FAILED);
```
simonlingoogle, line 531:
```suggestion
    VerifyOrExit(aDnsHeader.GetZoneRecordCount() == 1, error = OT_ERROR_PARSE);
```
which DNS rcode is better? FORMAT_ERROR or REFUSED ?

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1604 @@
...
+    // Handles only queries.
+    VerifyOrExit(dnsHeader.GetType() == Dns::UpdateHeader::Type::kTypeQuery, error = OT_ERROR_DROP);
+
+    switch (dnsHeader.GetQueryType())
+    {
+    case Dns::UpdateHeader::kQueryTypeUpdate:
+        HandleDnsUpdate(aMessage, aMessageInfo, dnsHeader, offset);
+        break;
+    default:
+        error = OT_ERROR_DROP;
```
simonlingoogle, line 1086: should respond with DNS RCODE NOT_IMPLEMENTED ?
```suggestion
        error = OT_ERROR_NOT_IMPLEMENTED;
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1604 @@
...
+{
+    otError           error;
+    Dns::UpdateHeader dnsHeader;
+    uint16_t          offset = aMessage.GetOffset();
+
+    SuccessOrExit(error = aMessage.Read(offset, dnsHeader));
+    offset += sizeof(dnsHeader);
+
+    // Handles only queries.
+    VerifyOrExit(dnsHeader.GetType() == Dns::UpdateHeader::Type::kTypeQuery, error = OT_ERROR_DROP);
```
simonlingoogle, line 1078:
```suggestion
    VerifyOrExit(dnsHeader.GetType() == Dns::UpdateHeader::Type::kTypeQuery, error = OT_ERROR_NOT_IMPLEMENTED);
```

should respond with DNS RCODE NOT_IMPLEMENTED ?

simonlingoogle (approved):
LGTM :+1: 

looks great overall. Only some nits.

src/core/net/srp_server.cpp
wgtdkp: I think the SRP server doesn't need to handle standard queries (which should be handled by a DNS-SD server), so `OT_ERROR_DROP` I think is better here.

src/core/net/srp_server.cpp
wgtdkp: I think the SRP server doesn't need to handle standard queries (which should be handled by a DNS-SD server), so OT_ERROR_DROP I think is better here.

src/core/net/srp_server.cpp
wgtdkp: I think we should use FORMAT_ERROR. https://tools.ietf.org/html/rfc2136#section-3. Will update.

src/core/config/openthread-core-default-config.h
```diff
@@ -265,7 +275,10 @@
  *
  */
 #ifndef OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE
-#if OPENTHREAD_CONFIG_COAP_SECURE_API_ENABLE
+#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
+// Internal heap doesn't support size larger than 64K bytes.
+#define OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE (63 * 1024)
```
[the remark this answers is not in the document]
wgtdkp: This change is for openthread core, not only posix platform.

abtink (approved):
LGTM. Thanks 👍

include/openthread/srp_server.h
```diff
@@ -0,0 +1,339 @@
...
+ * This function is called by the SRP server to notify that a SRP host and possibly SRP services
+ * are being updated. It is important that the SRP updates are not commited until the handler
+ * returns the result by calling otSrpServerHandleServiceUpdateResult or times out after @p aTimeout.
+ *
+ * A SRP service observer should always call otSrpServerHandleServiceUpdateResult with errorcode
+ * OT_ERROR_NONE immediately after receiving the update events.
+ *
+ * A more generic handler may perform validations on the SRP host/services and rejects the SRP updates
+ * if any validation fails. For example, an Advertising Proxy should advertise (or remove) the host and
+ * services on a multicast-capable link and returns specific errorcode if any failure occurs.
```
librasungirl, line 144:
```suggestion
 * services on a multicast-capable link and returns specific error code if any failure occurs.
```

src/cli/cli_srp_server.cpp
```diff
@@ -0,0 +1,246 @@
...
+        mInterpreter.OutputFormat("addresses: [");
+
+        addresses = otSrpServerHostGetAddresses(host, &addressesNum);
+
+        for (uint8_t i = 0; i < addressesNum; ++i)
+        {
+            mInterpreter.OutputIp6Address(addresses[i]);
+            if (i < addressesNum - 1)
+            {
+                mInterpreter.OutputSpaces(2);
```
librasungirl, line 155: nit: be consistent with below Line#219
```suggestion
                mInterpreter.OutputFormat(", ");
```

src/core/net/dns_headers.hpp
```diff
@@ -1189,6 +1206,56 @@ class KeyRecord : public ResourceRecord
...
+    /**
+     * This method returns the ECDSA P-256 public kek.
+     *
+     * @retval  A reference to the public key.
+     *
+     */
+    const Crypto::Ecdsa::P256::PublicKey &GetKey(void) const { return mKey; }
+
+    /**
+     * This comparator tells whether two Ecdsa256KeyRecord object equal to each other.
```
librasungirl, line 1239:
```suggestion
     * This comparator tells whether two Ecdsa256KeyRecord objects are equal.
```

src/core/crypto/ecdsa.hpp
```diff
@@ -246,7 +246,7 @@ class P256
         /**
          * This method gets the pointer to the buffer containing the public key (as an uncompressed curve point).
          *
-         * @return The pointer to the buffer containing the public key (with `kSize` bytes).
+         * @return The pointer to the buffer containing the public key (with `1 + kSize` bytes).
```
librasungirl, line 249: why add `1 + `?

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+    VerifyOrExit(response != nullptr, error = OT_ERROR_NO_BUFS);
+
+    header.SetMessageId(aHeader.GetMessageId());
+    header.SetType(Dns::UpdateHeader::kTypeResponse);
+    header.SetQueryType(aHeader.GetQueryType());
+    header.SetResponseCode(Dns::UpdateHeader::kResponseSuccess);
+    header.SetAdditionalRecordCount(1);
+    SuccessOrExit(error = response->Append(header));
+
+    SuccessOrExit(error = Dns::Name::AppendName(kRootName, *response));
```
librasungirl, line 1087: may use `Dns::Name::AppendTerminator()` instead

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+exit:
+    return error;
+}
+
+TimeMilli Server::Service::GetExpireTime(void) const
+{
+    OT_ASSERT(!mIsDeleted);
+    OT_ASSERT(!GetHost().IsDeleted());
+
+    return mTimeLastUpdate + GetHost().GetLease() * 1000;
```
librasungirl, line 1342: use existing method to convert s to ms here and the other occurrences.

```suggestion
    return mTimeLastUpdate + Time::SecToMsec(GetHost().GetLease());
```

librasungirl (approved):
LGTM with minor nits comments

bukepo (approved):
LGTM:+1:

src/cli/README_SRP_SERVER.md
```diff
@@ -0,0 +1,120 @@
+# OpenThread CLI - SRP Server
+
+## Quick Start
+
+See [README_SRP_CLIENT.md](README_SRP_CLIENT.md).
```
jwhui, line 5: Should this point to README_SRP.md?

src/cli/README_SRP_SERVER.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Command List
+
+- [help](#help)
+- [domain](#domain)
+- [disable](#disable)
```
jwhui, line 11: Alphabetical ordering?

src/cli/cli_srp_server.hpp
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * This method interprets a list of CLI arguments.
+     *
+     * @param[in]  aArgsLength  The number of elements in @p aArgs.
+     * @param[in]  aArgs        A pointer to an array of command line arguments.
+     *
+     */
+    otError Process(uint8_t aArgsLength, char *aArgs[]);
```
jwhui, line 75: Document return values?

src/core/api/srp_server_api.cpp
```diff
@@ -0,0 +1,160 @@
...
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *  This file defines the OpenThread SRP API.
```
jwhui, line 31:
```suggestion
 *  This file defines the OpenThread SRP server API.
```

src/core/config/openthread-core-default-config.h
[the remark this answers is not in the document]
jwhui: We are basically setting the heap size to max here. Any thoughts on what the actual bounds are?

src/core/net/dns_headers.hpp
```diff
@@ -1189,6 +1206,60 @@ class KeyRecord : public ResourceRecord
...
+     *
+     * Other record fields (TTL, length, flags, protocol) remain unchanged/uninitialized.
+     *
+     */
+    void Init(void);
+
+    /**
+     * This method tells whether this is a valid ECDSA DNSKEY with curve P-256.
+     *
+     * @retval  A boolean that indicates whether this is a valid ECDSA DNSKEY RR with curve P-256.
```
jwhui, line 1225:
```suggestion
     * @returns  A boolean that indicates whether this is a valid ECDSA DNSKEY RR with curve P-256.
```

src/core/net/dns_headers.hpp
```diff
@@ -1189,6 +1206,60 @@ class KeyRecord : public ResourceRecord
...
+     *
+     */
+    bool IsValid(void) const;
+
+    /**
+     * This method returns the ECDSA P-256 public kek.
+     *
+     * @retval  A reference to the public key.
```
jwhui, line 1233:
```suggestion
     * @returns  A reference to the public key.
```

src/core/net/dns_headers.hpp
```diff
@@ -1189,6 +1206,60 @@ class KeyRecord : public ResourceRecord
...
+     *
+     */
+    const Crypto::Ecdsa::P256::PublicKey &GetKey(void) const { return mKey; }
+
+    /**
+     * This comparator tells whether two Ecdsa256KeyRecord objects are equal.
+     *
+     * @param[in]  aOther  The other Ecdsa256KeyRecord object.
+     *
+     * @retval  TRUE if they are equal, FALSE if not.
```
jwhui, line 1243:
```suggestion
     * @returns  TRUE if they are equal, FALSE if not.
```

src/core/net/dns_headers.hpp
```diff
@@ -1189,6 +1206,60 @@ class KeyRecord : public ResourceRecord
...
+     *
+     */
+    bool operator==(const Ecdsa256KeyRecord &aOther) const { return memcmp(this, &aOther, sizeof(*this)) == 0; }
+
+    /**
+     * This comparator tells whether two Ecdsa256KeyRecord objects are not equal.
+     *
+     * @param[in]  aOther  The other Ecdsa256KeyRecord object.
+     *
+     * @retval  TRUE if they are not equal, FALSE if equal.
```
jwhui, line 1253:
```suggestion
     * @returns  TRUE if they are not equal, FALSE if equal.
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+
+exit:
+    return error;
+}
+
+uint32_t Server::GrantLease(uint32_t aLease) const
+{
+    OT_ASSERT(mMinLease <= mMaxLease);
+
+    return (aLease == 0) ? 0 : OT_MAX(mMinLease, OT_MIN(mMaxLease, aLease));
```
jwhui, line 156: What if `SetLeaseRange()` is called after `GrantLease()`?

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+}
+
+uint32_t Server::GrantKeyLease(uint32_t aKeyLease) const
+{
+    OT_ASSERT(mMinKeyLease <= mMaxKeyLease);
+
+    return (aKeyLease == 0) ? 0 : OT_MAX(mMinKeyLease, OT_MIN(mMaxKeyLease, aKeyLease));
```
jwhui, line 163: What if `SetLeaseRange()` is called after `GrantKeyLease()`?

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+// The caller MUST make sure that there is no existing host with the same hostname.
+void Server::AddHost(Host *aHost)
+{
+    OT_ASSERT(mHosts.FindMatching(aHost->GetFullName()) == nullptr);
+    IgnoreError(mHosts.Add(*aHost));
+}
+
+void Server::RemoveAndFreeHost(Host *aHost)
+{
+    otLogInfoSrp("[server] permanently removes host %s", aHost->GetFullName());
```
jwhui, line 221:
```suggestion
    otLogInfoSrp("[server] permanently remove host %s", aHost->GetFullName());
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+    aHost.SetLease(grantedLease);
+    aHost.SetKeyLease(grantedKeyLease);
+
+    existingHost = mHosts.FindMatching(aHost.GetFullName());
+
+    if (aHost.GetLease() == 0)
+    {
+        if (aHost.GetKeyLease() == 0)
+        {
+            otLogInfoSrp("[server] removes key of host %s", aHost.GetFullName());
```
jwhui, line 325:
```suggestion
            otLogInfoSrp("[server] remove key of host %s", aHost.GetFullName());
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+        else if (existingHost != nullptr)
+        {
+            Service *service = nullptr;
+
+            existingHost->SetLease(aHost.GetLease());
+            existingHost->SetKeyLease(aHost.GetKeyLease());
+
+            // Clear all resources associated to this host and its services.
+            existingHost->ClearResources();
+            otLogInfoSrp("[server] removes host '%s' (but retains its name)", existingHost->GetFullName());
```
jwhui, line 341:
```suggestion
            otLogInfoSrp("[server] remove host '%s' (but retains its name)", existingHost->GetFullName());
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+
+        aHost.Free();
+    }
+    else if (existingHost != nullptr)
+    {
+        const Service *service = nullptr;
+
+        // Merge current updates into existing host.
+
+        otLogInfoSrp("[server] updates host %s", existingHost->GetFullName());
```
jwhui, line 356:
```suggestion
        otLogInfoSrp("[server] update host %s", existingHost->GetFullName());
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+                otLogInfoSrp("[server] %s service %s", (existingService != nullptr) ? "updates existing" : "adds new",
+                             service->mFullName);
+            }
+        }
+
+        aHost.Free();
+    }
+    else
+    {
+        otLogInfoSrp("[server] adds new host %s", aHost.GetFullName());
```
jwhui, line 385:
```suggestion
        otLogInfoSrp("[server] add new host %s", aHost.GetFullName());
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(!IsRunning());
+
+    SuccessOrExit(error = mSocket.Open(HandleUdpReceive, this));
+    SuccessOrExit(error = mSocket.Bind(0));
+
+    SuccessOrExit(error = PublishServerData());
+
+    otLogInfoSrp("[server] starts listening on port %hu", mSocket.GetSockName().mPort);
```
jwhui, line 414:
```suggestion
    otLogInfoSrp("[server] start listening on port %hu", mSocket.GetSockName().mPort);
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+
+    while (!mOutstandingUpdates.IsEmpty())
+    {
+        mOutstandingUpdates.Pop()->Free();
+    }
+
+    mLeaseTimer.Stop();
+    mOutstandingUpdatesTimer.Stop();
+
+    otLogInfoSrp("[server] stops listening on %hu", mSocket.GetSockName().mPort);
```
jwhui, line 444:
```suggestion
    otLogInfoSrp("[server] stop listening on port %hu", mSocket.GetSockName().mPort);
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+
+exit:
+    return;
+}
+
+void Server::HandleNotifierEvents(Events aEvents)
+{
+    VerifyOrExit(mEnabled);
+
+    if (aEvents.Contains(kEventThreadRoleChanged))
```
jwhui, line 455: Use `VerifyOrExit()` to avoid additional nesting.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+
+    uint16_t headerOffset = aOffset - sizeof(aDnsHeader);
+
+    otLogInfoSrp("[server] receives DNS update from %s", aMessageInfo.GetPeerAddr().ToString().AsCString());
+
+    SuccessOrExit(error = ProcessZoneSection(aMessage, aDnsHeader, aOffset, zone));
+
+    if (FindOutstandingUpdate(aMessageInfo, aDnsHeader.GetMessageId()) != nullptr)
+    {
+        otLogInfoSrp("[server] drops duplicated SRP update request: messageId=%hu", aDnsHeader.GetMessageId());
```
jwhui, line 542:
```suggestion
        otLogInfoSrp("[server] drop duplicated SRP update request: messageId=%hu", aDnsHeader.GetMessageId());
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+                             const Dns::UpdateHeader &aDnsHeader,
+                             uint16_t                 aOffset)
+{
+    otError   error = OT_ERROR_NONE;
+    Dns::Zone zone;
+    Host *    host = nullptr;
+
+    uint16_t headerOffset = aOffset - sizeof(aDnsHeader);
+
+    otLogInfoSrp("[server] receives DNS update from %s", aMessageInfo.GetPeerAddr().ToString().AsCString());
```
jwhui, line 536:
```suggestion
    otLogInfoSrp("[server] receive DNS update from %s", aMessageInfo.GetPeerAddr().ToString().AsCString());
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+    header.SetType(Dns::UpdateHeader::kTypeResponse);
+    header.SetQueryType(aHeader.GetQueryType());
+    header.SetResponseCode(aResponseCode);
+    SuccessOrExit(error = response->Append(header));
+
+    SuccessOrExit(error = mSocket.SendTo(*response, aMessageInfo));
+
+    if (aResponseCode != Dns::UpdateHeader::kResponseSuccess)
+    {
+        otLogInfoSrp("[server] sends fail response: %d", aResponseCode);
```
jwhui, line 1050:
```suggestion
        otLogInfoSrp("[server] send fail response: %d", aResponseCode);
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+    }
+    else
+    {
+        otLogInfoSrp("[server] sends success response");
```
jwhui, line 1054:
```suggestion
        otLogInfoSrp("[server] send success response");
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+    SuccessOrExit(error = response->Append(optRecord));
+
+    leaseOption.Init();
+    leaseOption.SetLeaseInterval(aLease);
+    leaseOption.SetKeyLeaseInterval(aKeyLease);
+    SuccessOrExit(error = response->Append(leaseOption));
+
+    SuccessOrExit(error = mSocket.SendTo(*response, aMessageInfo));
+
+    otLogInfoSrp("[server] sent response with granted lease: %u and key lease: %u", aLease, aKeyLease);
```
jwhui, line 1102:
```suggestion
    otLogInfoSrp("[server] send response with granted lease: %u and key lease: %u", aLease, aKeyLease);
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+    TimeMilli earliestExpireTime = now.GetDistantFuture();
+    Host *    host               = mHosts.GetHead();
+
+    while (host != nullptr)
+    {
+        Host *nextHost = host->GetNext();
+
+        if (host->GetKeyExpireTime() <= now)
+        {
+            otLogInfoSrp("[server] KEY LEASE of host %s expires", host->GetFullName());
```
jwhui, line 1164:
```suggestion
            otLogInfoSrp("[server] KEY LEASE of host %s expired", host->GetFullName());
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+            service = host->GetNextService(nullptr);
+            while (service != nullptr)
+            {
+                OT_ASSERT(service->mIsDeleted);
+
+                Service *nextService = service->GetNext();
+
+                if (service->GetKeyExpireTime() <= now)
+                {
+                    otLogInfoSrp("[server] KEY LEASE of service %s expires", service->mFullName);
```
jwhui, line 1187:
```suggestion
                    otLogInfoSrp("[server] KEY LEASE of service %s expired", service->mFullName);
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+                }
+
+                service = nextService;
+            }
+        }
+        else if (host->GetExpireTime() <= now)
+        {
+            Service *service = nullptr;
+
+            otLogInfoSrp("[server] LEASE of host %s expires", host->GetFullName());
```
jwhui, line 1202:
```suggestion
            otLogInfoSrp("[server] LEASE of host %s expired", host->GetFullName());
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+                Service *nextService = service->GetNext();
+
+                if (service->mIsDeleted)
+                {
+                    // The service has been deleted but the name retains.
+                    earliestExpireTime = OT_MIN(earliestExpireTime, service->GetKeyExpireTime());
+                }
+                else if (service->GetExpireTime() <= now)
+                {
+                    otLogInfoSrp("[server] LEASE of service %s expires", service->mFullName);
```
jwhui, line 1234:
```suggestion
                    otLogInfoSrp("[server] LEASE of service %s expired", service->mFullName);
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+        }
+
+        host = nextHost;
+    }
+
+    if (earliestExpireTime != now.GetDistantFuture())
+    {
+        if (!mLeaseTimer.IsRunning() || earliestExpireTime <= mLeaseTimer.GetFireTime())
+        {
+            otLogInfoSrp("[server] lease timer is scheduled in %u seconds", (earliestExpireTime - now) / 1000);
```
jwhui, line 1256:
```suggestion
            otLogInfoSrp("[server] lease timer is scheduled for %u seconds", (earliestExpireTime - now) / 1000);
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+    GetInstance().HeapFree(mTxtData);
+    mTxtData   = nullptr;
+    mTxtLength = 0;
+}
+
+void Server::Service::DeleteResourcesButRetainName(void)
+{
+    mIsDeleted = true;
+    ClearResources();
+    otLogInfoSrp("[server] removes service '%s' (but retains its name)", mFullName);
```
jwhui, line 1407:
```suggestion
    otLogInfoSrp("[server] remove service '%s' (but retain its name)", mFullName);
```

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+}
+
+// Add a new service entry to the host, do nothing if there is already
+// such services with the same name.
+Server::Service *Server::Host::AddService(const char *aFullName)
+{
+    Service *service;
+
+    service = FindService(aFullName);
+    if (service == nullptr)
```
jwhui, line 1520: nit: use `VerifyOrExit()` to avoid additional nesting.

src/core/net/srp_server.cpp
```diff
@@ -0,0 +1,1654 @@
...
+{
+    mAddressesNum = 0;
+}
+
+void Server::Host::DeleteResourcesButRetainName(void)
+{
+    // Mark the host as deleted.
+    mLease = 0;
+    ClearResources();
+    otLogInfoSrp("[server] removes host '%s' (but retains its name)", mFullName);
```
jwhui, line 1560:
```suggestion
    otLogInfoSrp("[server] remove host '%s' (but retain its name)", mFullName);
```

src/core/net/srp_server.hpp
```diff
@@ -0,0 +1,643 @@
...
+        void Free(void);
+
+        /**
+         * This method tells if the SRP service has been deleted.
+         *
+         * A SRP service can be deleted but retains its name for future uses.
+         * In this case, the service instance is not removed from the SRP server/registry.
+         * It is guaranteed that all services are deleted if the host is deleted.
+         *
+         * @retval  TRUE if the service has been deleted, FALSE if not.
```
jwhui, line 114:
```suggestion
         * @returns  TRUE if the service has been deleted, FALSE if not.
```

jwhui (commented):
Looks good overall. Commented on a number of minor nits.

src/core/net/srp_server.cpp
wgtdkp: I think this is legal. Granted key lease will not be affected, and new host will be granted with new key lease range.

src/core/net/srp_server.cpp
wgtdkp: Granted lease will not be affected, and new service will be granted with new lease range.

src/core/config/openthread-core-default-config.h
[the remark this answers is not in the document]
wgtdkp: Memory cost of a typical Service Instance Entry:
- **fullname**: around 40 bytes for `my-service._ipps._tcp.default.service.arpa`, and 255 bytes for maximum.
- **metadata**: around 40 bytes (`weight`, `priority`, `port` and pointers).
- **TXT data**: this value can vary from 1 byte to 65535 bytes. For typical real use cases, I think it should be around 30 ~ 200 bytes.
- Total size: 40 + 40 + 30 = 110 bytes.

Memory cost of a typical Host Instance Entry:
- **fullname**: around 30 bytes for `my-host.default.service.arpa`, and 255 bytes for maximum.
- **metadata**: around 30 bytes (`lease`, `key lease` and pointers).
- **address**: we use static array (`kMaxAddressesNum = 8`) for address list and it costs `8 * 16` bytes. I think for most cases, there will be a single address. I will change this value to `2` to allow possible 1.2 addresses (e.g. DUA).
- **ECDSA KEY**: 64 bytes.
- Total size: 30 + 30 + 32 + 64 156 bytes.


I remember that the duckhorn BR requires the SRP server to serve around 500 services from 250 clients (2 services for each client). This will definitely exceeds the maximum size of the internal heap. We can do some optimizations to the SRP server implementation to reduce the memory consumption for each service and host entry. For example, we can stores only the service instance name and service name but not the full name for each service entry. But I don't think the maximum size can be limited to the internal heap capability. I think for posix and other resourceful platforms, we should just use external heap for the Border Router.

jwhui:
Merged with #6074

PR #5988, opened 2020-12-21
[heap] expand internal heap size for ECDSA

wgtdkp:
The ECDSA feature needs more heap to run because buffers for signing and signature verification are dynamically allocated.

The number `2600` is the minimum (not precise) size to pass the `test_ecdsa` unittest case. This is not revealed in CI tests because `DTLS` is enabled (and thus we have much larger heap) for unittest check.


abtink (approved):
LGTM. Thanks 👍

jwhui (approved):

PR #5989, opened 2020-12-21
[ecdsa] fix ECDSA public key size for DNSKEY

wgtdkp:
This PR fixes the ECDSA public key size for DNSKEY.

Per [Sec 4 of RFC 6605](https://tools.ietf.org/html/rfc6605#section-4):
```txt
ECDSA public keys consist of a single value, called "Q" in FIPS
186-3.  In DNSSEC keys, Q is a simple bit string that represents the
uncompressed form of a curve point, "x | y".
```
The bit string of the two curve points (X, Y) should have a fixed length of 64 bytes and the additional `1` byte prefix should not be included, because only uncompressed form is allowed.

The example in [Sec 6.1](https://tools.ietf.org/html/rfc6605#section-6.1)  also gives 64 bytes public key for the KEY RR.
```txt
example.net. 3600 IN DNSKEY 257 3 13 (
           GojIhhXUN/u4v54ZQqGSnyhWJwaubCvTmeexv7bR6edb
           krSqQpF64cYbcB7wNcP+e+MAnLr+Wi9xMWyQLc8NAA== )
```
Decoding the base64 encoded rdata of this KEY RR, you will get a hex string with 128 chars
```txt
1a88c88615d437fbb8bf9e1942a1929f28562706ae6c2bd399e7b1bfb6d1e9e75b92b4aa42917ae1c61b701ef035c3fe7be3009cbafe5a2f71316c902dcf0d00
```


abtink (approved):
LGTM. Thanks 👍

jwhui (approved):

PR #5997, opened 2020-12-22
[cli] router table command to output `Link`

simonlingoogle:
This commit enhance OT CLI command `router table` to output whether Link Established  in `Link` field.

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -3468,8 +3468,8 @@ otError Interpreter::ProcessRouter(uint8_t aArgsLength, char *aArgs[])
 
         if (isTable)
         {
-            OutputLine("| ID | RLOC16 | Next Hop | Path Cost | LQ In | LQ Out | Age | Extended MAC     | Link Est |");
-            OutputLine("+----+--------+----------+-----------+-------+--------+-----+------------------+----------+");
+            OutputLine("| ID | RLOC16 | Next Hop | Path Cost | LQ In | LQ Out | Age | Extended MAC     | Link |");
+            OutputLine("+----+--------+----------+-----------+-------+--------+-----+------------------+------+");
         }
 
         maxRouterId = otThreadGetMaxRouterId(mInstance);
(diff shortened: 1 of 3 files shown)


abtink (approved):
LGTM. 👍

src/cli/cli.cpp
```diff
@@ -3468,8 +3468,8 @@ otError Interpreter::ProcessRouter(uint8_t aArgsLength, char *aArgs[])
 
         if (isTable)
         {
-            OutputLine("| ID | RLOC16 | Next Hop | Path Cost | LQ In | LQ Out | Age | Extended MAC     |");
-            OutputLine("+----+--------+----------+-----------+-------+--------+-----+------------------+");
+            OutputLine("| ID | RLOC16 | Next Hop | Path Cost | LQ In | LQ Out | Age | Extended MAC     | Link |");
```
bukepo, line 3471: Does this mean bidirectional link or unidirectional link?
```suggestion
            OutputLine("| ID | RLOC16 | Next Hop | Path Cost | LQ In | LQ Out | Age | Extended MAC     | Linked |");
```

tests/scripts/thread-cert/node.py
```diff
@@ -1936,6 +1937,7 @@ def router_table(self):
                 'lqout': lqout,
                 'age': age,
                 'emac': emac,
+                'link': link,
```
bukepo, line 1940:
```suggestion
                'linked': linked,
```

src/cli/cli.cpp
simonlingoogle: Agree that `Link` is not a perfect header name. 
But I used `Link` because `router <id>` command is already using `Link: `

See https://github.com/openthread/openthread/blob/master/src/cli/README.md#router-id for reference.

bukepo (approved):

src/cli/cli.cpp
simonlingoogle: I will keep it as is to be inline with `router <id>` command.

tests/scripts/thread-cert/node.py
simonlingoogle: I will keep it as is to be inline with `router <id>` command.

jwhui (approved):

PR #5998, opened 2020-12-22
[key-manager] remove default master key

Moandor-y:
When initializing, we should use randomly generated network parameters instead of default ones, to eliminate the possibility of unintentional use of default credentials.


simonlingoogle:
What's the background for this change? 

I think OpenThread developers should be responsible for changing master key.

Moandor-y:
> What's the background for this change?
> 
> I think OpenThread developers should be responsible for changing master key. Meanwhile, the default master key is convenient for testing purposes.

The default master key did go into some of our products' legacy pairing process, and caused some vulnerabilities.

Yes the responsibility does fall upon the caller, but it's better to make such issues less likely to happen if the caller doesn't notice it. For testing purposes you can write expect scripts to automate the setup process.

Moandor-y:
> What's the background for this change?
> 
> I think OpenThread developers should be responsible for changing master key. Meanwhile, the default master key is convenient for testing purposes.

Yes the responsibility does fall upon the caller, but it's better to make such issues less likely to happen if the caller doesn't notice it. For testing purposes you can write expect scripts to automate the setup process.

simonlingoogle:
> > What's the background for this change?
> > I think OpenThread developers should be responsible for changing master key. Meanwhile, the default master key is convenient for testing purposes.
> 
> Yes the responsibility does fall upon the caller, but it's better to make such issues less likely to happen if the caller doesn't notice it. For testing purposes you can write expect scripts to automate the setup process.

Agree that it's more secure this way.

src/core/thread/key_manager.cpp
```diff
@@ -48,25 +49,6 @@ const uint8_t KeyManager::kThreadString[] = {
...
-    0x77,
-    0x88,
-    0x99,
-    0xaa,
-    0xbb,
-    0xcc,
-    0xdd,
-    0xee,
-    0xff,
-}};
```
abtink, line 68: Some of the tests may rely on the default values and/or verify them.

For example, I know, `test-002-form` under troanj does verify the default values  and checks once we use `wpantund` `form` command, the default values are not used by newly formed Thread network (i.e. new random values are used for key, pan-id, xpanid, (note that this is done by `wpantund` during `form`)).  

 With the change from this PR, we probably can remove some of the checks  in the test-002.

simonlingoogle: Let's make @qilinkang aware of this change too.

Moandor-y: It seems the default key verification in `test-002-form` is the only thing relying on the default value in toranj tests. Removed it and toranj tests passed.

tests/toranj/test-002-form.py
```diff
@@ -59,7 +59,6 @@
 DEFAULT_XPANID = '0xDEAD00BEEF00CAFE'
 
 verify(node.get(wpan.WPAN_STATE) == wpan.STATE_OFFLINE)
-verify(node.get(wpan.WPAN_KEY) == DEFAULT_KEY)
```
bukepo, line 62: Set master key when form?

Moandor-y:
We can't randomize other parameters' initial values (prefix, ext pan id, etc.) for now. Because if we set up 2 nodes with the same channel, pan id, master key, then start them at the same time, both of them will become leader and generate different active dataset, then merge into one network, but keep their different values forever (because none of their active timestamps are greater than the other).

I'll revert back to randomizing the master key only, and send an issue to discuss this further.

tests/toranj/test-002-form.py
Moandor-y: Done.

bukepo (approved):
LGTM:+1:

simonlingoogle (approved):
LGTM

src/core/thread/mle.cpp
```diff
@@ -201,6 +201,14 @@ otError Mle::Start(bool aAnnounceAttach)
     VerifyOrExit(!Get<Radio>().GetPromiscuous(), error = OT_ERROR_INVALID_STATE);
     VerifyOrExit(Get<ThreadNetif>().IsUp(), error = OT_ERROR_INVALID_STATE);
 
+    if (Get<KeyManager>().GetMasterKey() == MasterKey())
```
abtink, line 204: How about generating a random key in the constructor of `KeyManager` instead of the checking and setting here?

tests/toranj/test-002-form.py
```diff
@@ -53,7 +53,7 @@
 # Test implementation
 
 # default values after reset
-DEFAULT_KEY = '[00112233445566778899AABBCCDDEEFF]'
+DEFAULT_KEY = '[00000000000000000000000000000000]'
```
abtink, line 56: This is fine, but since we may no longer have the notion of a default key in OT, how about we simplify the test script here as well. We can consider  removing the `DEFAULT_KEY` and all checks against it in the test script.

abtink (commented):
Looks good overall. Thanks. 👍 
Two suggestions below.

src/core/thread/mle.cpp
Moandor-y: Could there be a chance that when booting up the system doesn't have enough entropy to generate a truly random master key?

tests/toranj/test-002-form.py
Moandor-y: Done.

src/core/thread/mle.cpp
abtink: I  think we generate a random ext address already in `Mac` constructor which requires crypto rng to be ready.

Moandor-y: Updated.

src/core/thread/key_manager.cpp
```diff
@@ -91,7 +73,7 @@ KeyManager::KeyManager(Instance &aInstance)
     , mIsPskcSet(false)
 {
     mMacFrameCounters.Reset();
-    mMasterKey = static_cast<const MasterKey &>(kDefaultMasterKey);
+    OT_ASSERT(mMasterKey.GenerateRandom() == OT_ERROR_NONE);
```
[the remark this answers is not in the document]
simonlingoogle: why is it resolved?
Should never put real logic code in `OT_ASSERT`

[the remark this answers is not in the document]
Moandor-y: But we can't return error codes from a constructor.

It isn't real logic, there shouldn't be any error here. A similar case is <https://github.com/openthread/openthread/blob/9ca79ddd9af3d4e3f78cb6e611a3117a71b2198c/src/core/mac/mac.cpp#L129> in which the error is ignored.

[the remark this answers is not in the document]
Moandor-y: OK, I misunderstood. Fixed.

abtink (approved):
Looks great. Thanks 👍

jwhui (approved):

PR #6002, opened 2020-12-23
[crypto] add new flavors of 'Update()' for SHA-256 and HMAC computation

abtink:
(edited)

This commit updates `Sha256` and `HmacSha256` classes adding new
flavors of `Update()` method which inputs bytes from a given buffer,
or from an object, or read from a `Message` for hash calculation.

This commit also updates unit test to add more test-cases (from RFC
4231) for HMAC and adds new set of test-cases for SHA-256. The test
also covers calculating hash over data read directly from a `Message`.

----

Quick note: The `Update()` flavor reading from `Message would be 
useful for SIG(0)  calculation in SRP.


src/core/crypto/hmac_sha256.hpp
```diff
@@ -87,13 +90,37 @@ class HmacSha256
...
     /**
-     * This method inputs bytes into the HMAC computation.
+     * This method feeds bytes into the HMAC computation.
      *
      * @param[in]  aBuf        A pointer to the input buffer.
      * @param[in]  aBufLength  The length of @p aBuf in bytes.
      *
      */
-    void Update(const uint8_t *aBuf, uint16_t aBufLength);
+    void FeedBytes(const void *aBuf, uint16_t aBufLength);
```
jwhui, line 99: nit: most hash APIs I've seen use "update" or "input". Is there precedent for using "feed"?

wgtdkp: Agree with @jwhui. If there is no semantic changes between `update` and `feed`, I would prefer keeping `update`.

wgtdkp (approved):
LGTM :+1: 
Thanks @abtink, This is useful for DNS TSIG verification, especially we can update SHA256 computation from data in messages.

src/core/crypto/hmac_sha256.hpp
abtink: I wanted  to somehow include the `Message` in the name of the method which reads the data from a `Message`. Initially I had it called `UpdateMessage()` which felt weird to me (it is as if we are updating the message and not reading from it)... Also considered `UpdateFromMessage()` but I feel it's too long.  So I changed them all to `Feed()`  and used `FeedMessage()`...  😄 

I don't have any strong preference here. If `Update` is commonly used I am ok using  `Update` for this. I will then overload the method name `Update()` for all flavors . We can tell them apart from the number and types  of the passed arguments - which I think is fine. I will go with this for now.

Thoughts/suggestions?

wgtdkp: Overloading the method works for me.

jwhui (approved):

PR #6007, opened 2020-12-24
[ip6] use ALOC source address

simonlingoogle:
#5632 convert ALOC source addresses to RLOCs.

This commit revert this feature.

**Background:**
[CoAP](https://tools.ietf.org/html/rfc7252#section-5.3.2) requires that the source address of a response must match the destination addresses of the request. So, if a device sends a TMF request using ALOC as destination address, the TMF response **SHOULD** use the ALOC as source address. Otherwise, CoAP implementations (beside OpenThread) may fail to match the request and response.


bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #6008, opened 2020-12-24
[border-router] add external route for on-link prefix and external OMR prefixes

wgtdkp:
This PR includes the enhancement that
1. adds non-default external route for on-link prefixes.
2. add external route for OMR prefixes advertised by BRs in other Thread Networks.
3. a new API to disable/enable the Border Routing manager at run time. CLI commands are added for testing.
4. a new platform API to get the link-local address of the infra interface so that we can filter out the ICMPv6 packets from myself by the Border Routing Manager rather than the platform implementation.

enhancements that adds random delay before changing the prefixes/routes, multiple prefixes for fault tolerance and duplicate OMR prefixes detection will be added in future PRs.

### DIFF
--- a/src/cli/README.md
+++ b/src/cli/README.md
@@ -22,6 +22,7 @@ Done
 ## OpenThread Command List
 
 - [bbr](#bbr)
+- [br](#br)
 - [bufferinfo](#bufferinfo)
 - [ccathreshold](#ccathreshold)
 - [channel](#channel)
@@ -309,6 +310,20 @@ Set jitter (in seconds) for Backbone Router registration for Thread 1.2 FTD.
 
 ```bash
 > bbr jitter 10
+Done
+```
+
+### br
+
+Enbale/disable the Border Routing functionality.
+
+```bash
+> br enable
+Done
+```
+
+```bash
+> br disable
 Done
 ```
 
--- a/src/core/border_router/router_advertisement.cpp
+++ b/src/core/border_router/router_advertisement.cpp
@@ -110,9 +110,12 @@ void PrefixInfoOption::SetPrefix(const Ip6::Prefix &aPrefix)
     mPrefix       = static_cast<const Ip6::Address &>(aPrefix.mPrefix);
 }
 
-void PrefixInfoOption::GetPrefix(Ip6::Prefix &aPrefix) const
+Ip6::Prefix PrefixInfoOption::GetPrefix(void) const
 {
-    aPrefix.Set(mPrefix.GetBytes(), mPrefixLength);
+    Ip6::Prefix prefix;
+
+    prefix.Set(mPrefix.GetBytes(), mPrefixLength);
+    return prefix;
 }
 
 RouteInfoOption::RouteInfoOption(void)
@@ -126,6 +129,23 @@ RouteInfoOption::RouteInfoOption(void)
     mPrefix.Clear();
 }
 
+void RouteInfoOption::SetPreference(otRoutePreference aPreference)
+{
+    mReserved &= ~kPreferenceMask;
+    mReserved |= (static_cast<uint8_t>(aPreference) << kPreferenceOffset) & kPreferenceMask;
+}
+
+otRoutePreference RouteInfoOption::GetPreference(void) const
+{
+    uint8_t preference = (mReserved & kPreferenceMask) >> kPreferenceOffset;
+
+    if (preference == ((OT_ROUTE_PREFERENCE_LOW & kPreferenceMask) >> kPreferenceOffset))
+    {
+        preference = OT_ROUTE_PREFERENCE_LOW;
+    }
+    return static_cast<otRoutePreference>(preference);
+}
+
 void RouteInfoOption::SetPrefix(const Ip6::Prefix &aPrefix)
 {
     // The total length (in bytes) of a Router Information Option
@@ -136,6 +156,23 @@ void RouteInfoOption::SetPrefix(const Ip6::Prefix &aPrefix)
 
     mPrefixLength = aPrefix.mLength;
     mPrefix       = static_cast<const Ip6::Address &>(aPrefix.mPrefix);
+}
+
+Ip6::Prefix RouteInfoOption::GetPrefix(void) const
+{
+    Ip6::Prefix prefix;
+
+    prefix.Set(mPrefix.GetBytes(), mPrefixLength);
+    return prefix;
+}
+
+bool RouteInfoOption::IsValid(void) const
+{
+    otRoutePreference pref = GetPreference();
+
+    return (GetLength() == kLengthUnit || GetLength() == 2 * kLengthUnit || GetLength() == 3 * kLengthUnit) &&
+           (mPrefixLength <= OT_IP6_ADDRESS_SIZE * CHAR_BIT) &&
+           (pref == OT_ROUTE_PREFERENCE_LOW || pref == OT_ROUTE_PREFERENCE_MED || pref == OT_ROUTE_PREFERENCE_HIGH);
 }
 
 RouterAdvMessage::RouterAdvMessage(void)
--- a/src/core/border_router/router_advertisement.cpp
+++ b/src/core/border_router/router_advertisement.cpp
@@ -55,12 +55,12 @@ const Option *Option::GetNextOption(const Option *aCurOption, const uint8_t *aBu
     }
     else
     {
-        nextOption = reinterpret_cast<const uint8_t *>(aCurOption) + aCurOption->GetLength();
+        nextOption = reinterpret_cast<const uint8_t *>(aCurOption) + aCurOption->GetSize();
     }
 
     VerifyOrExit(nextOption + sizeof(Option) <= bufferEnd, nextOption = nullptr);
-    VerifyOrExit(reinterpret_cast<const Option *>(nextOption)->GetLength() > 0, nextOption = nullptr);
-    VerifyOrExit(nextOption + reinterpret_cast<const Option *>(nextOption)->GetLength() <= bufferEnd,
+    VerifyOrExit(reinterpret_cast<const Option *>(nextOption)->GetSize() > 0, nextOption = nullptr);
+    VerifyOrExit(nextOption + reinterpret_cast<const Option *>(nextOption)->GetSize() <= bufferEnd,
                  nextOption = nullptr);
 
 exit:
@@ -137,13 +137,25 @@ void RouteInfoOption::SetPreference(otRoutePreference aPreference)
 
 otRoutePreference RouteInfoOption::GetPreference(void) const
 {
-    uint8_t preference = (mReserved & kPreferenceMask) >> kPreferenceOffset;
+    otRoutePreference preference;
 
-    if (preference == ((OT_ROUTE_PREFERENCE_LOW & kPreferenceMask) >> kPreferenceOffset))
+    switch ((mReserved & kPreferenceMask) >> kPreferenceOffset)
     {
+    case kPreferenceLow:
         preference = OT_ROUTE_PREFERENCE_LOW;
+        break;
+    case kPreferenceMed:
+        preference = OT_ROUTE_PREFERENCE_MED;
+        break;
+    case kPreferenceHigh:
+        preference = OT_ROUTE_PREFERENCE_HIGH;
+        break;
+    default:
+        preference = OT_ROUTE_PREFERENCE_LOW;
+        break;
     }
-    return static_cast<otRoutePreference>(preference);
+
+    return preference;
 }
 
 void RouteInfoOption::SetPrefix(const Ip6::Prefix &aPrefix)
@@ -152,7 +164,7 @@ void RouteInfoOption::SetPrefix(const Ip6::Prefix &aPrefix)
     // is: (8 bytes fixed option header) + (0, 8, or 16 bytes prefix).
     // Because the length of the option must be padded with 8 bytes,
     // the length of the prefix (in bits) must be padded with 64 bits.
-    SetLength(((aPrefix.mLength + kLengthUnit * CHAR_BIT - 1) / (kLengthUnit * CHAR_BIT) + 1) * kLengthUnit);
+    SetLength((aPrefix.mLength + kLengthUnit * CHAR_BIT - 1) / (kLengthUnit * CHAR_BIT) + 1);
 
     mPrefixLength = aPrefix.mLength;
     mPrefix       = static_cast<const Ip6::Address &>(aPrefix.mPrefix);
@@ -170,7 +182,7 @@ bool RouteInfoOption::IsValid(void) const
 {
     otRoutePreference pref = GetPreference();
 
-    return (GetLength() == kLengthUnit || GetLength() == 2 * kLengthUnit || GetLength() == 3 * kLengthUnit) &&
+    return (GetLength() == 1 || GetLength() == 2 || GetLength() == 3) &&
            (mPrefixLength <= OT_IP6_ADDRESS_SIZE * CHAR_BIT) &&
            (pref == OT_ROUTE_PREFERENCE_LOW || pref == OT_ROUTE_PREFERENCE_MED || pref == OT_ROUTE_PREFERENCE_HIGH);
 }
--- a/src/core/border_router/routing_manager.cpp
+++ b/src/core/border_router/routing_manager.cpp
@@ -521,6 +521,12 @@ const Ip6::Prefix *RoutingManager::EvaluateOnLinkPrefix(void)
             otLogInfoBr("EvaluateOnLinkPrefix: there is already smaller on-link prefix %s on interface %u",
                         smallestOnLinkPrefix->ToString().AsCString(), mInfraIfIndex);
 
+            // TODO: we removes the advertised on-link prefix by setting valid lifetime for PIO.
+            // But SLAAC addresses configured by PIO prefix will not be removed immediately (
+            // https://tools.ietf.org/html/rfc4862#section-5.5.3). This leads to a situation that
+            // a WiFi device keeps using the old SLAAC address but a Thread device cannot reach to
+            // it. One solution is to delay removing external route until the SLAAC addresses are
+            // actually expired/deprecated.
             RemoveExternalRoute(*mAdvertisedOnLinkPrefix);
         }
     }
--- a/src/core/border_router/routing_manager.cpp
+++ b/src/core/border_router/routing_manager.cpp
@@ -61,32 +61,58 @@ RoutingManager::RoutingManager(Instance &aInstance)
     : InstanceLocator(aInstance)
     , mIsRunning(false)
     , mInfraIfIndex(0)
+    , mEnabled(true) // The routing manager is by default enabled.
     , mAdvertisedOmrPrefixNum(0)
     , mAdvertisedOnLinkPrefix(nullptr)
+    , mDiscoveredPrefixNum(0)
+    , mDiscoveredPrefixInvalidTimer(aInstance, HandleDiscoveredPrefixInvalidTimer, this)
     , mRouterAdvertisementTimer(aInstance, HandleRouterAdvertisementTimer, this)
     , mRouterAdvertisementCount(0)
     , mRouterSolicitTimer(aInstance, HandleRouterSolicitTimer, this)
     , mRouterSolicitCount(0)
-    , mDiscoveredOnLinkPrefixInvalidTimer(aInstance, HandleDiscoveredOnLinkPrefixInvalidTimer, this)
 {
     mLocalOmrPrefix.Clear();
     memset(mAdvertisedOmrPrefixes, 0, sizeof(mAdvertisedOmrPrefixes));
 
     mLocalOnLinkPrefix.Clear();
-    mDiscoveredOnLinkPrefix.Clear();
+
+    memset(mDiscoveredPrefixes, 0, sizeof(mDiscoveredPrefixes));
 }
 
 otError RoutingManager::Init(uint32_t aInfraIfIndex)
 {
     otError error;
 
-    OT_ASSERT(!IsInitialized() && !Get<Mle::MleRouter>().IsAttached());
+    OT_ASSERT(!IsInitialized());
     VerifyOrExit(aInfraIfIndex > 0, error = OT_ERROR_INVALID_ARGS);
 
     SuccessOrExit(error = LoadOrGenerateRandomOmrPrefix());
     SuccessOrExit(error = LoadOrGenerateRandomOnLinkPrefix());
 
     mInfraIfIndex = aInfraIfIndex;
+
+exit:
+    return error;
+}
+
+otError RoutingManager::SetEnabled(bool aEnabled)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(IsInitialized(), error = OT_ERROR_INVALID_STATE);
+
+    VerifyOrExit(aEnabled != mEnabled);
+
+    mEnabled = aEnabled;
+
+    if (!mEnabled)
+    {
+        Stop();
+    }
+    else if (Get<Mle::MleRouter>().IsAttached())
+    {
+        Start();
+    }
 
 exit:
     return error;
@@ -150,6 +176,8 @@ void RoutingManager::Start(void)
 {
     if (!mIsRunning)
     {
+        otLogInfoBr("Border Routing manager is started");
+
         mIsRunning = true;
         StartRouterSolicitation();
     }
@@ -160,6 +188,11 @@ void RoutingManager::Stop(void)
     VerifyOrExit(mIsRunning);
 
     UnpublishLocalOmrPrefix();
+    if (mAdvertisedOnLinkPrefix != nullptr)
+    {
+        RemoveExternalRoute(*mAdvertisedOnLinkPrefix);
+    }
+    InvalidateAllDiscoveredPrefixes();
 
     // Use empty OMR & on-link prefixes to invalidate possible advertised prefixes.
     SendRouterAdvertisement(nullptr, 0, nullptr);
@@ -169,14 +202,17 @@ void RoutingManager::Stop(void)
 
     mAdvertisedOnLinkPrefix = nullptr;
 
-    mDiscoveredOnLinkPrefix.Clear();
-    mDiscoveredOnLinkPrefixInvalidTimer.Stop();
+    memset(mDiscoveredPrefixes, 0, sizeof(mDiscoveredPrefixes));
+    mDiscoveredPrefixNum = 0;
+    mDiscoveredPrefixInvalidTimer.Stop();
 
     mRouterAdvertisementTimer.Stop();
     mRouterAdvertisementCount = 0;
 
     mRouterSolicitTimer.Stop();
     mRouterSolicitCount = 0;
+
+    otLogInfoBr("Border Routing manager is stopped");
 
     mIsRunning = false;
 
@@ -189,12 +225,19 @@ void RoutingManager::RecvIcmp6Message(uint32_t            aInfraIfIndex,
                                       const uint8_t *     aBuffer,
                                       uint16_t            aBufferLength)
 {
+    otError                  error = OT_ERROR_NONE;
     const Ip6::Icmp::Header *icmp6Header;
+    const Ip6::Address *     infraLinkLocalAddr;
 
-    VerifyOrExit(mIsRunning);
+    VerifyOrExit(IsInitialized() && mIsRunning, error = OT_ERROR_DROP);
 
-    VerifyOrExit(aInfraIfIndex == mInfraIfIndex);
-    VerifyOrExit(aBuffer != nullptr && aBufferLength >= sizeof(*icmp6Header));
+    VerifyOrExit(aInfraIfIndex == mInfraIfIndex, error = OT_ERROR_DROP);
+    infraLinkLocalAddr = static_cast<const Ip6::Address *>(otPlatInfraIfGetLinkLocalAddress(mInfraIfIndex));
+
+    // Drop any ICMPv6 messages sent from myself.
+    VerifyOrExit(infraLinkLocalAddr != nullptr && aSrcAddress != *infraLinkLocalAddr, error = OT_ERROR_DROP);
+
+    VerifyOrExit(aBuffer != nullptr && aBufferLength >= sizeof(*icmp6Header), error = OT_ERROR_PARSE);
 
     icmp6Header = reinterpret_cast<const Ip6::Icmp::Header *>(aBuffer);
 
@@ -211,12 +254,15 @@ void RoutingManager::RecvIcmp6Message(uint32_t            aInfraIfIndex,
     }
 
 exit:
-    return;
+    if (error != OT_ERROR_NONE)
+    {
+        otLogDebgBr("drop ICMPv6 message: %s", otThreadErrorToString(error));
+    }
 }
 
 void RoutingManager::HandleNotifierEvents(Events aEvents)
 {
-    VerifyOrExit(IsInitialized());
+    VerifyOrExit(IsInitialized() && IsEnabled());
 
     if (aEvents.Contains(kEventThreadRoleChanged))
     {
@@ -253,8 +299,7 @@ uint8_t RoutingManager::EvaluateOmrPrefix(Ip6::Prefix *aNewOmrPrefixes, uint8_t 
     {
         uint8_t newPrefixIndex;
 
-        if (!IsValidOmrPrefix(onMeshPrefixConfig.GetPrefix()) || !onMeshPrefixConfig.mDefaultRoute ||
-            !onMeshPrefixConfig.mSlaac || onMeshPrefixConfig.mDp)
+        if (!IsValidOmrPrefix(onMeshPrefixConfig.GetPrefix()) || !onMeshPrefixConfig.mSlaac || onMeshPrefixConfig.mDp)
         {
             continue;
         }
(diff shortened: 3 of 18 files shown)


tests/scripts/thread-cert/node.py
```diff
@@ -1278,21 +1278,39 @@ def remove_prefix(self, prefix):
         self.send_command(cmd)
         self._expect('Done')
 
-    def get_prefixes(self):
-        netdata = self.netdata_show()
-        prefixes = []
+    def enable_br(self):
+        self.send_command("br enable")
```
bukepo, line 1282:
```suggestion
        self.send_command('br enable')
```

tests/scripts/thread-cert/node.py
```diff
@@ -2154,6 +2172,17 @@ def get_addrs(self) -> List[str]:
     def __repr__(self):
         return f'Host<{self.nodeid}>'
 
+    def get_matched_ula_addresses(self, prefix):
+        """ Get the IPv6 addresses that matches given prefix.
```
bukepo, line 2176:
```suggestion
        """Get the IPv6 addresses that matches given prefix.
```

src/core/border_router/routing_manager.hpp
```diff
@@ -85,6 +85,14 @@ class RoutingManager : public InstanceLocator
      */
     otError Init(uint32_t aInfraIfIndex);
 
+    /**
+     * This method enables/disables the border routing manager.
+     *
+     * @param[in]  aEnabled  A boolean to enable/disable the routing manager.
+     *
+     */
+    void SetEnabled(bool aEnabled);
```
bukepo, line 94: is dynamically enable/disable necessary?

bukepo (approved):

src/core/border_router/routing_manager.hpp
wgtdkp: The duckhorn use case doc requires the border router being configuration. I think for most cases, applications will not need this. But there needs a way to cease being a Border Route gracefully (remove prefixes before stop Thread), this is at least useful for testing. Without dynamical enable/disable, we need to wait minutes for the prefixes to be removed when a BR is stopped (we need to test with real time because there are non-thread devices, e.g. WiFi/Ethernet devices).

src/core/border_router/routing_manager.cpp
```diff
@@ -730,62 +844,214 @@ void RoutingManager::HandleRouterAdvertisement(const Ip6::Address &aSrcAddress,
...
+    return needReevaluate;
+}
+
+bool RoutingManager::UpdateDiscoveredPrefixes(const RouterAdv::RouteInfoOption &aRio)
+{
+    Ip6::Prefix prefix;
+    bool        needReevaluate = false;
+
+    aRio.GetPrefix(prefix);
+    if (!IsValidOmrPrefix(prefix))
```
librasungirl, line 915: Should we auto-install the external route for any RA(RIO) learned from infrastructure link? like a prefix that might be of 48 bit length

src/core/border_router/routing_manager.hpp
```diff
@@ -243,18 +275,68 @@ class RoutingManager : public InstanceLocator
...
+     */
+    void InvalidateAllDiscoveredPrefixes(void);
+
+    /**
+     * This methods adds a discovered on-link prefix found in RA message.
+     *
+     * If the same on-link prefix already exists, only the lifetime will be updated.
+     *
+     * @param[in]  aPrefix    The on-link prefix to be added.
+     * @param[in]  aLifetime  The lifetime of the prefix. In seconds.
```
librasungirl, line 326: doc `param[in] aIsOnLinkPrefix`

src/core/border_router/routing_manager.hpp
```diff
@@ -243,18 +275,68 @@ class RoutingManager : public InstanceLocator
...
+    uint8_t InvalidateDiscoveredPrefixes(const Ip6::Prefix *aPrefix = nullptr, bool aIsOnLinkPrefix = true);
+
+    /**
+     * This method invalidates all discovered on-link prefixes.
+     *
+     */
+    void InvalidateAllDiscoveredPrefixes(void);
+
+    /**
+     * This methods adds a discovered on-link prefix found in RA message.
```
librasungirl, line 321: `on-link` here seems not accurate?

tests/scripts/thread-cert/border_routing/test_multi_border_routers.py
```diff
@@ -102,35 +110,111 @@ def test(self):
...
+        # Router1 and Router2 can ping each other inside the Thread network.
+        self.assertTrue(self.nodes[ROUTER1].ping(self.nodes[ROUTER2].get_ip6_address(config.ADDRESS_TYPE.OMR)[0]))
+        self.assertTrue(self.nodes[ROUTER2].ping(self.nodes[ROUTER1].get_ip6_address(config.ADDRESS_TYPE.OMR)[0]))
+
+        # Both Router1 and Router2 can ping to/from the Host on infra link.
+        self.assertTrue(self.nodes[ROUTER1].ping(self.nodes[HOST].get_matched_ula_addresses(br1_on_link_prefix)[0]))
+        self.assertTrue(self.nodes[HOST].ping(self.nodes[ROUTER1].get_ip6_address(config.ADDRESS_TYPE.OMR)[0],
+                                              backbone=True))
+        self.assertTrue(self.nodes[ROUTER2].ping(self.nodes[HOST].get_matched_ula_addresses(br1_on_link_prefix)[0]))
+        self.assertTrue(self.nodes[HOST].ping(self.nodes[ROUTER2].get_ip6_address(config.ADDRESS_TYPE.OMR)[0],
```
librasungirl, line 161: may use a `for i in [ROUTER1, ROUTER2]` loop for repeat procedure

tests/scripts/thread-cert/border_routing/test_multi_border_routers.py
```diff
@@ -102,35 +110,111 @@ def test(self):
...
+                                              backbone=True))
+
+        #
+        # Case 2. Another BR continue provide border routing when current one is disabled.
```
librasungirl, line 165:
```suggestion
        # Case 2. Another BR continues providing border routing when current one is disabled.
```

tests/scripts/thread-cert/border_routing/test_multi_thread_networks.py
```diff
@@ -119,6 +119,25 @@ def test(self):
...
 
+        br1_omr_prefix = self.nodes[BR1].get_prefixes()[0]
+        br2_omr_prefix = self.nodes[BR2].get_prefixes()[0]
+
+        self.assertNotEqual(br1_omr_prefix, br2_omr_prefix)
+
+        # Each BR should independently register an external route for the on-link prefix
+        # and OMR prefix in another Thread Network.
+        self.assertTrue(len(self.nodes[BR1].get_routes()) == 2)
+        self.assertTrue(len(self.nodes[ROUTER1].get_routes()) == 2)
```
librasungirl, line 130: nit: update Line#43 in topology `ED` -> `ROUTER` to match the device type in use

tests/scripts/thread-cert/border_routing/test_single_border_router.py
```diff
@@ -143,6 +163,110 @@ def test(self):
...
+        logging.info("Host addresses: %r", self.nodes[HOST].get_addrs())
+
+        # The host should have at least an ULA address and a link-local address.
+        self.assertGreaterEqual(len(self.nodes[HOST].get_addrs()), 2)
+
+        self.collect_ipaddrs()
+
+        logging.info("BR1     addrs: %r", self.nodes[BR1].get_addrs())
+        logging.info("ROUTER1 addrs: %r", self.nodes[ROUTER1].get_addrs())
+        logging.info("HOST    addrs: %r", self.nodes[HOST].get_addrs())
```
librasungirl, line 181: nit: dup with Line#172

tests/scripts/thread-cert/border_routing/test_single_border_router.py
```diff
@@ -143,6 +163,110 @@ def test(self):
...
```
librasungirl, line 244: nit: dup with Line#235

src/core/border_router/routing_manager.cpp
```diff
@@ -730,62 +844,214 @@ void RoutingManager::HandleRouterAdvertisement(const Ip6::Address &aSrcAddress,
...
+        case Option::Type::kPrefixInfo:
+            pio = static_cast<const PrefixInfoOption *>(option);
+            if (pio->IsValid())
+            {
+                needReevaluate |= UpdateDiscoveredPrefixes(*pio);
+            }
+            break;
+
+        case Option::Type::kRouteInfo:
+            rio = static_cast<const RouteInfoOption *>(option);
```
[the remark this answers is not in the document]
wgtdkp: I think we don't need to track the router associated with this prefix on Thread interface. It is the platform (posix in our case) that needs to properly handle Route Preference and add routes for those prefix on the infra interface. Right?

src/core/border_router/routing_manager.cpp
```diff
@@ -730,62 +844,214 @@ void RoutingManager::HandleRouterAdvertisement(const Ip6::Address &aSrcAddress,
...
+    otLogInfoBr("discovered on-link prefix (%s, %u seconds) from interface %u", prefix.ToString().AsCString(),
+                aPio.GetValidLifetime(), mInfraIfIndex);
+
+    if (aPio.GetValidLifetime() == 0)
+    {
+        needReevaluate |= (InvalidateDiscoveredPrefixes(&prefix, /* aIsOnLinkPrefix */ true) > 0);
+    }
+    else
+    {
+        needReevaluate |= AddDiscoveredPrefix(prefix, /* aIsOnLinkPrefix */ true, aPio.GetValidLifetime());
```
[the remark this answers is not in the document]
wgtdkp: You are right, I copy & paste the code here and forget to update it.

src/core/border_router/routing_manager.cpp
wgtdkp: add external route for RIO is for connectivity to another Thread network which is connected to the same infra link. I don't feel it necessary to add explicit support for non-duckhorn BR prefixes in another Thread network. It there is no duckhorn BR in that Thread network, it would be an expected result that there is no connectivity to the Thread network. Thoughts?

src/core/border_router/routing_manager.hpp
```diff
@@ -264,7 +346,10 @@ class RoutingManager : public InstanceLocator
     Ip6::Prefix mLocalOmrPrefix;
 
     /**
-     * The advertised OMR prefixes.
+     * The advertised OMR prefixes. For a stable Thread network without
+     * non-standard BRs, there should be a single OMR prefix that being
```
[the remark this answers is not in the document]
wgtdkp: Changed to `non-standard Duckhorn BR`, sounds good to you?

src/core/border_router/routing_manager.cpp
```diff
@@ -357,6 +392,45 @@ void RoutingManager::UnpublishLocalOmrPrefix(void)
...
+{
+    otError                          error;
+    NetworkData::ExternalRouteConfig routeConfig;
+
+    OT_ASSERT(mIsRunning);
+
+    routeConfig.Clear();
+    routeConfig.GetPrefix() = aPrefix;
+    routeConfig.mStable     = true;
+    routeConfig.mPreference = 0;
```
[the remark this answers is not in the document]
wgtdkp: included it in this PR.

src/core/border_router/routing_manager.hpp
```diff
@@ -243,18 +276,74 @@ class RoutingManager : public InstanceLocator
...
+     */
+    bool UpdateDiscoveredPrefixes(const RouterAdv::RouteInfoOption &aRio);
+
+    /**
+     * This method removes expired or specific discovered prefixes.
+     *
+     * @param[in]  aPrefix          A pointer to a specific on-link prefix to be removed.
+     * @param[in]  aIsOnLinkPrefix  A boolean which indicates whether the prefix is an on-link prefix.
+     *
+     * @returns  The number of removed on-link prefixes.
```
librasungirl, line 311: nit: ~~`on-link`~~

src/core/border_router/routing_manager.hpp
```diff
@@ -243,18 +276,74 @@ class RoutingManager : public InstanceLocator
...
+     *
+     */
+    uint8_t InvalidateDiscoveredPrefixes(const Ip6::Prefix *aPrefix = nullptr, bool aIsOnLinkPrefix = true);
+
+    /**
+     * This method invalidates all discovered on-link prefixes.
```
librasungirl, line 317: nit: ~~`on-link`~~

src/core/border_router/routing_manager.hpp
```diff
@@ -243,18 +276,74 @@ class RoutingManager : public InstanceLocator
...
+     * This methods adds a prefix discovered in a RA message.
+     *
+     * If the same prefix already exists, only the lifetime will be updated.
+     *
+     * @param[in]  aPrefix           The on-link prefix to be added.
+     * @param[in]  aIsOnLinkPrefix   A boolean which indicates whether the prefix is an on-link prefix.
+     * @param[in]  aLifetime         The lifetime of the prefix. In seconds.
+     * @param[in]  aRoutePreference  The preference of the route which is associated with the prefix.
+     *
+     * @returns  A boolean indicates whether a new on-link prefix is added.
```
librasungirl, line 332: nit: ~~`on-link`~~

src/core/border_router/routing_manager.hpp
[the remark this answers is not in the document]
librasungirl: A standard Duckhorn BR should also allow advanced user configures more than one on-mesh prefixes.
Guess your intention here indicates that no prefix manually configured?

src/core/border_router/routing_manager.cpp
librasungirl: Fine to me to keep it as is.

librasungirl (approved):
LGTM 👍 , only a few nits

src/core/border_router/routing_manager.hpp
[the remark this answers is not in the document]
wgtdkp: Updated to use `manually configured OMR prefixes`.

src/posix/platform/infra_if.cpp
```diff
@@ -154,11 +160,16 @@ static void InitLinkLocalAddress(void)
...
-            break;
+            ExitNow(error = OT_ERROR_NONE);
         }
     }
 
+    otLogCritPlat("cannot find IPv6 link-local address for interface %s", sInfraIfName);
+    error = OT_ERROR_NOT_FOUND;
+
+exit:
     freeifaddrs(ifAddrs);
```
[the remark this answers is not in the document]
wgtdkp: I think the `freeifaddrs` will handle the case and the example in the linux man page shares the same pattern : https://linux.die.net/man/3/freeifaddrs

src/cli/README.md
```diff
@@ -312,6 +313,20 @@ Set jitter (in seconds) for Backbone Router registration for Thread 1.2 FTD.
 Done
 ```
 
+### br
+
+Enbale/disable the Border Routing functionality.
+
```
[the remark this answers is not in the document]
wgtdkp: We currently don't have a use of getting this info, may add it when there are needs.

src/core/border_router/routing_manager.cpp
```diff
@@ -730,62 +866,216 @@ void RoutingManager::HandleRouterAdvertisement(const Ip6::Address &aSrcAddress,
...
+    }
+
+exit:
+    return;
+}
+
+bool RoutingManager::UpdateDiscoveredPrefixes(const RouterAdv::PrefixInfoOption &aPio)
+{
+    Ip6::Prefix prefix = aPio.GetPrefix();
+    bool        needReevaluate;
```
simonlingoogle: Variable 'needReevaluate' is used uninitialized whenever 'if' condition is true

src/core/border_router/routing_manager.cpp
```diff
@@ -373,22 +467,49 @@ bool RoutingManager::ContainsPrefix(const Ip6::Prefix &aPrefix, const Ip6::Prefi
...
+    // We may not be able to remove our local on-link prefix when a bigger prefix is received.
+    // In this case, there will be two on-link prefixes on the infra link.
+    else if (mAdvertisedOnLinkPrefix != nullptr && *smallestOnLinkPrefix != *mAdvertisedOnLinkPrefix)
+    {
+        otLogInfoBr("EvaluateOnLinkPrefix: there is already smaller on-link prefix %s on interface %u",
+                    smallestOnLinkPrefix->ToString().AsCString(), mInfraIfIndex);
 
-    newOnLinkPrefix = &mLocalOnLinkPrefix;
+        RemoveExternalRoute(*mAdvertisedOnLinkPrefix);
+        newOnLinkPrefix = nullptr;
```
simonlingoogle, line 511:
```suggestion
```
I think `newOnLinkPrefix` must be nullptr here.

src/core/border_router/routing_manager.cpp
```diff
@@ -445,8 +566,14 @@ void RoutingManager::EvaluateRoutingPolicy(void)
 
 void RoutingManager::StartRouterSolicitation(void)
 {
+    uint32_t randomDelay;
+
     mRouterSolicitCount = 0;
-    mRouterSolicitTimer.Start(Random::NonCrypto::GetUint32InRange(0, kMaxRtrSolicitationDelay * 1000));
+
+    randomDelay = Random::NonCrypto::GetUint32InRange(0, kMaxRtrSolicitationDelay * 1000);
```
simonlingoogle, line 573: This call requires that `kMaxRtrSolicitationDelay` > 0

```suggestion
    static_assert(kMaxRtrSolicitationDelay > 0);
    randomDelay = Random::NonCrypto::GetUint32InRange(0, kMaxRtrSolicitationDelay * 1000);
```

src/core/border_router/routing_manager.cpp
```diff
@@ -730,62 +866,216 @@ void RoutingManager::HandleRouterAdvertisement(const Ip6::Address &aSrcAddress,
...
             mRouterSolicitTimer.Stop();
-            needReevaluate = true;
         }
-    }
 
-    if (needReevaluate)
+        newPrefix.mPrefix         = aPrefix;
+        newPrefix.mIsOnLinkPrefix = aIsOnLinkPrefix;
+        newPrefix.mExpireTime     = TimerMilli::GetNow() + GetPrefixExpireDelay(aLifetime);
+        if (newPrefix.mExpireTime < mDiscoveredPrefixInvalidTimer.GetFireTime())
```
simonlingoogle, line 1064: Can we be sure that `mDiscoveredPrefixInvalidTimer` is started? I think it's meaningless to compare the fire time if `mDiscoveredPrefixInvalidTimer` is not started yet.

src/core/border_router/routing_manager.cpp
```diff
@@ -211,12 +250,15 @@ void RoutingManager::RecvIcmp6Message(uint32_t            aInfraIfIndex,
...
+    if (error != OT_ERROR_NONE)
+    {
+        otLogDebgBr("drop ICMPv6 message: %s", otThreadErrorToString(error));
+    }
 }
 
 void RoutingManager::HandleNotifierEvents(Events aEvents)
 {
-    VerifyOrExit(IsInitialized());
+    VerifyOrExit(IsInitialized() && IsEnabled());
```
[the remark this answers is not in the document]
wgtdkp: The Border Routing Manager is by default enabled. I think the case is that we may receive the events when we are not initialized.

simonlingoogle (approved):
LGTM :+1:

src/core/border_router/router_advertisement.cpp
```diff
@@ -110,9 +110,12 @@ void PrefixInfoOption::SetPrefix(const Ip6::Prefix &aPrefix)
...
 }
 
-void PrefixInfoOption::GetPrefix(Ip6::Prefix &aPrefix) const
+Ip6::Prefix PrefixInfoOption::GetPrefix(void) const
 {
-    aPrefix.Set(mPrefix.GetBytes(), mPrefixLength);
+    Ip6::Prefix prefix;
+
+    prefix.Set(mPrefix.GetBytes(), mPrefixLength);
+    return prefix;
```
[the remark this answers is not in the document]
wgtdkp: Yes it is just for readability. Most compilers will do [NRVO](https://en.cppreference.com/w/cpp/language/copy_elision) to avoid unnecessary copies but it is not guaranteed. I am ok to revert it if we prefer performance against readability.

src/core/border_router/routing_manager.cpp
```diff
@@ -373,22 +471,48 @@ bool RoutingManager::ContainsPrefix(const Ip6::Prefix &aPrefix, const Ip6::Prefi
...
-    newOnLinkPrefix = &mLocalOnLinkPrefix;
+    // We start advertising our local on-link prefix if there is no existing one.
+    if (onLinkPrefixNum == 0)
+    {
+        if (AddExternalRoute(mLocalOnLinkPrefix, OT_ROUTE_PREFERENCE_MED) == OT_ERROR_NONE)
+        {
+            newOnLinkPrefix = &mLocalOnLinkPrefix;
+        }
+    }
+    // We may not be able to remove our local on-link prefix when a bigger prefix is received.
```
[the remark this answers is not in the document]
wgtdkp: In case an application-specific on-link prefix is received and it is bigger than our advertised prefix, both prefixes will stay on the infra link. Updated the documentation, please review.

src/core/border_router/router_advertisement.cpp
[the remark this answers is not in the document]
jwhui: I am OK with this as is. Thanks.

src/core/border_router/routing_manager.cpp
```diff
@@ -373,22 +471,50 @@ bool RoutingManager::ContainsPrefix(const Ip6::Prefix &aPrefix, const Ip6::Prefi
...
 
-    // We don't evaluate on-link prefix if we are doing
-    // Router Discovery or we have already discovered some
-    // on-link prefixes.
+    // We don't evaluate on-link prefix if we are doing Router Solicitation.
     VerifyOrExit(!mRouterSolicitTimer.IsRunning());
-    if (IsValidOnLinkPrefix(mDiscoveredOnLinkPrefix))
+
+    for (uint8_t i = 0; i < mDiscoveredPrefixNum; ++i)
     {
```
abtink, line 484: Just a suggestion: how about adding `ExternalPrefix &prefix = mDiscoveredPrefixes[i];` in `for` block (I like that we have this in `InvalidateDiscoveredPrefixes` and `AddDiscoveredPrefix` and makes it nicer and easier to read)

src/core/border_router/routing_manager.cpp
```diff
@@ -373,22 +471,50 @@ bool RoutingManager::ContainsPrefix(const Ip6::Prefix &aPrefix, const Ip6::Prefi
...
+    {
+        otLogInfoBr("EvaluateOnLinkPrefix: there is already smaller on-link prefix %s on interface %u",
+                    smallestOnLinkPrefix->ToString().AsCString(), mInfraIfIndex);
 
-    newOnLinkPrefix = &mLocalOnLinkPrefix;
+        RemoveExternalRoute(*mAdvertisedOnLinkPrefix);
+    }
 
 exit:
     return newOnLinkPrefix;
```
abtink, line 520: The `newOnLinkPrefix` would be null except when there is no existing on-link prefix in the list.

I see that this is then used to update the `mAdvertisedOnLinkPrefix` from `EvaluateRoutingPolicy()`. 

If we see that there are other prefixes but ours is smaller, should we not continue to to keep  `mAdvertisedOnLinkPrefix` as before (and send it in our RA?)

src/core/border_router/routing_manager.cpp
```diff
@@ -373,22 +471,50 @@ bool RoutingManager::ContainsPrefix(const Ip6::Prefix &aPrefix, const Ip6::Prefi
...
+    // advertised prefix, we will not remove the advertised prefix. In this case, there
+    // will be two on-link prefixes on the infra link. But all BRs will still converge to
+    // the same smallest on-link prefix and the application-specific prefix is not used.
+    else if (mAdvertisedOnLinkPrefix != nullptr && *smallestOnLinkPrefix != *mAdvertisedOnLinkPrefix)
+    {
+        otLogInfoBr("EvaluateOnLinkPrefix: there is already smaller on-link prefix %s on interface %u",
+                    smallestOnLinkPrefix->ToString().AsCString(), mInfraIfIndex);
 
-    newOnLinkPrefix = &mLocalOnLinkPrefix;
+        RemoveExternalRoute(*mAdvertisedOnLinkPrefix);
```
abtink, line 516: Question about this case, 
- we were advertising `mAdvertisedOnLinkPrefix` before in our RA.
- We now see there is a smaller prefix, so we try to remove ours
   - I don't think we send an RA with lifetime zero to remove ours  and basically expect the timeout to invalidate it
   - But we immediately remove the external route in thread net-data.
- So we can have a trasnistent time where the on-link prefix is still being used in infra net side (before timing out) and so devices on infra side can still use IP addresses based on this prefix to interact with Thread nodes, but this BR will immediately stop routing for this prefix (stop acting as external route for this prefix).
- Can this be a problem?

src/core/border_router/routing_manager.cpp
```diff
@@ -373,22 +471,50 @@ bool RoutingManager::ContainsPrefix(const Ip6::Prefix &aPrefix, const Ip6::Prefi
...
+        {
+            continue;
+        }
+
+        ++onLinkPrefixNum;
+
+        if (smallestOnLinkPrefix == nullptr ||
+            IsPrefixSmallerThan(mDiscoveredPrefixes[i].mPrefix, *smallestOnLinkPrefix))
+        {
+            smallestOnLinkPrefix = &mDiscoveredPrefixes[i].mPrefix;
```
abtink, line 495: One more question about the on-link prefix management
- Is it the case that when things converge to steady state, only on BR (the one with smallest prefix) would be advertising its on-link prefix in its RAs?
- If so,  if that BR is disconnected/gobe, the prefix will be gone (times out) and then all infra links need to allocate new address (based on new prefix from remaining BRs).
- This sound similar to me like the OMR prefix situation and that for fault tolerance, we want to have multiple BRs add the same prefix in net-data.
- Is this an issue? Do we need to consider this?

Thoughts?

src/core/border_router/routing_manager.hpp
```diff
@@ -243,18 +289,74 @@ class RoutingManager : public InstanceLocator
...
 
+    /**
+     * This method updates the discovered prefix list with given Prefix Information Option.
+     *
+     * @param[in]  aPio  A reference to the Prefix Information Option.
+     *
+     * @returns  TRUE if the prefix list was updated, FALSE otherwise.
+     *
+     */
+    bool UpdateDiscoveredPrefixes(const RouterAdv::PrefixInfoOption &aPio);
```
abtink, line 306: Just a suggestion (please feel free to leave for maybe later PRs): 
I noticed that you have nice documentation for all these methods (and many more from earlier PR), but these methods are all `private`. I think it may be better not to use the doxygen format for the private method so that they are not included in the doxygen API documentations). My suggestion like before is to move them in `cpp` file. Again, it's fine for me as is (just a suggestion for later maybe).

abtink (commented):
Thanks @wgtdkp. Overall looks very nicely written and clean.

I have some smaller suggestions below and couple of questions on how/when we should remove our advertised on-link prefix. Thanks.

src/core/border_router/routing_manager.cpp
jwhui: I agree this can be an issue. We should continue to advertise external route for any prefixes that may still be in use.

src/core/border_router/routing_manager.cpp
```diff
@@ -729,63 +871,221 @@ void RoutingManager::HandleRouterAdvertisement(const Ip6::Address &aSrcAddress,
...
-        pio->GetPrefix(prefix);
-        if (!IsValidOnLinkPrefix(prefix))
+        else
         {
-            otLogInfoBr("ignore invalid prefix in PIO: %s", prefix.ToString().AsCString());
-            continue;
+            earliestExpireTime = OT_MIN(earliestExpireTime, prefix.mExpireTime);
+            *keptPrefix        = prefix;
+            ++keptPrefix;
+            keptOnLinkPrefixNum += prefix.mIsOnLinkPrefix;
```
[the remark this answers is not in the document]
jwhui:
```suggestion
            if (prefix.mIsOnLinkPrefix)
            {
                keptOnLinkPrefixNum++;
            }
```

src/core/border_router/routing_manager.cpp
jwhui: I think it is generally a good goal to keep IPv6 addresses as stable as possible. Any change to IPv6 address will require devices to perform an SRP update to register their new address and any peers to determine that the IPv6 address has changed and perform DNS-SD again.

src/core/border_router/routing_manager.cpp
```diff
@@ -729,63 +871,221 @@ void RoutingManager::HandleRouterAdvertisement(const Ip6::Address &aSrcAddress,
...
+    for (uint8_t i = 0; i < mDiscoveredPrefixNum; ++i)
+    {
+        ExternalPrefix &prefix = mDiscoveredPrefixes[i];
+
+        if (aPrefix == prefix.mPrefix && aIsOnLinkPrefix == prefix.mIsOnLinkPrefix)
         {
-            if (mDiscoveredOnLinkPrefix == prefix)
+            prefix.mExpireTime = TimerMilli::GetNow() + GetPrefixExpireDelay(aLifetime);
+            if (mDiscoveredPrefixInvalidTimer.IsRunning() ||
+                prefix.mExpireTime < mDiscoveredPrefixInvalidTimer.GetFireTime())
```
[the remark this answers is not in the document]
wgtdkp: Thanks, it's a typo.

src/core/border_router/routing_manager.cpp
```diff
@@ -729,63 +871,221 @@ void RoutingManager::HandleRouterAdvertisement(const Ip6::Address &aSrcAddress,
...
-        if (pio->GetValidLifetime() == 0)
+    OT_ASSERT(mDiscoveredPrefixNum == 0);
+}
+
+bool RoutingManager::AddDiscoveredPrefix(const Ip6::Prefix &aPrefix,
+                                         bool               aIsOnLinkPrefix,
+                                         uint32_t           aLifetime,
+                                         otRoutePreference  aRoutePreference)
+{
+    OT_ASSERT(IsValidOnLinkPrefix(aPrefix));
```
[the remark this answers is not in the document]
wgtdkp: I forgot to update the check after some refactoring, thanks.

src/core/border_router/routing_manager.cpp
wgtdkp: Sure! I share the same preference but maybe forgot to update the loop here.

src/core/border_router/routing_manager.cpp
wgtdkp: Good Catch!

src/core/border_router/routing_manager.cpp
wgtdkp: > I don't think we send an RA with lifetime zero to remove ours and basically expect the timeout to invalidate it

The `newOnLinkPrefix` is default to `nullptr` and it will be used for `SendRouterAdvertisement()` to send RA with zero lifetime. I may be more readable to have `SendRouterAdvertisement` together with `RemoveExternalRoute` but we didn't do it so that we can send all RA options in a single RA message.

But the issue still exists, per https://tools.ietf.org/html/rfc4862#section-5.5.3:
>1.  If the received Valid Lifetime is greater than 2 hours or
       greater than RemainingLifetime, set the valid lifetime of the
       corresponding address to the advertised Valid Lifetime.
>2.  If RemainingLifetime is less than or equal to 2 hours, ignore
       the Prefix Information option with regards to the valid
       lifetime, unless the Router Advertisement from which this
       option was obtained has been authenticated (e.g., via Secure
       Neighbor Discovery [RFC3971]).  If the Router Advertisement
       was authenticated, the valid lifetime of the corresponding
       address should be set to the Valid Lifetime in the received
       option.
>
>3.  Otherwise, reset the valid lifetime of the corresponding
          address to 2 hours.

SLAAC address will be retained even when RA with zero valid lifetime is received (zero valid lifetime is useful to tell other BRs to start advertising their own prefixes). My idea is to apply the same rule of SLAAC to external routes and delay removing external routes before the SLAAC addresses are actually expired. Thoughts?

src/core/border_router/routing_manager.cpp
wgtdkp: yes, I think a similar strategy for OMR prefixes can be applied to on-link prefixes.

src/core/border_router/routing_manager.hpp
wgtdkp: Thanks, updated to the suggested style in new push, please review.

src/core/border_router/routing_manager.cpp
wgtdkp: Please notice that even a SLAAC address is expired/deprecated, it doesn't stops the application layer to use it:
>Note that the feasibility of initiating new communication using a
   non-deprecated address may be an application-specific decision, as
   only the application may have knowledge about whether the (now)
   deprecated address was (or still is) in use by the application.  For
example, if an application explicitly specifies that the protocol
   stack use a deprecated address as a source address, the protocol
   stack must accept that; the application might request it because that
   IP address is used in higher-level communication and there might be a
   requirement that the multiple connections in such a grouping use the
   same pair of IP addresses.

https://tools.ietf.org/html/rfc4862#section-5.5.4

If we remove the external route for the on-link prefix when those SLAAC addresses on infra link are deprecated, then we will not be able to support the cases mentioned above (though I don't think it is a usual case).

Properly handling the cases requires non-trivial changes to current implementation. I am planning to leave this to future enhancement PR, will it work for you? @abtink @jwhui

wgtdkp: Added `TODO` for this issue.

abtink (approved):
LGTM. Thanks.

src/core/border_router/routing_manager.cpp
abtink: SGTM. Thanks.

jwhui (approved):

PR #6020, opened 2020-12-29
[nrf528xx] enable unused-parameter warning for platform code

Irving-cl:
This PR enables `unused-parameter` warning for NRF driver code except for soft-device.

Currently this warning is disabled for all NRF driver code, which makes `OT_UNUSED_VARIABLE` in driver code useless. I found that we only need to suppress the warning for soft-device code because only that part includes some headers in `third_party/NordicSemiconductor` that have unused parameters. We'd better to make the rest part of driver code consistent with OpenThread code.


LuDuda (approved):
Thanks :+1:

jwhui (approved):

PR #6021, opened 2020-12-29
[nrf52840] implement enh-ack probing

Irving-cl:
This PR implements Enhanced-Ack Based Probing on nrf platform.

### DIFF
--- a/third_party/NordicSemiconductor/drivers/radio/nrf_802154_core.c
+++ b/third_party/NordicSemiconductor/drivers/radio/nrf_802154_core.c
@@ -169,6 +169,7 @@ static uint8_t *       mp_ack;         ///< Pointer to Ack frame buffer.
 static const uint8_t * mp_tx_data;     ///< Pointer to the data to transmit.
 static uint32_t        m_ed_time_left; ///< Remaining time of the current energy detection procedure [us].
 static uint8_t         m_ed_result;    ///< Result of the current energy detection procedure.
+static uint8_t         m_last_lqi;     ///< LQI value of last Rx frame.
 
 static volatile radio_state_t m_state; ///< State of the radio driver.
 
@@ -2144,6 +2145,8 @@ static void irq_crcok_state_rx(void)
             nrf_802154_pib_auto_ack_get())
         {
             mp_ack = nrf_802154_ack_generator_create(mp_current_rx_buffer->data);
+            m_last_lqi = lqi_get(mp_current_rx_buffer->data);
+
             if (NULL != mp_ack)
             {
                 send_ack = true;
@@ -3260,6 +3263,11 @@ bool nrf_802154_core_last_rssi_measurement_get(int8_t * p_rssi)
     }
 
     return result;
+}
+
+uint8_t nrf_802154_core_last_lqi_get(void)
+{
+    return m_last_lqi;
 }
 
 #if NRF_802154_INTERNAL_RADIO_IRQ_HANDLING
--- a/third_party/NordicSemiconductor/drivers/radio/nrf_802154_core.c
+++ b/third_party/NordicSemiconductor/drivers/radio/nrf_802154_core.c
@@ -3265,11 +3265,6 @@ bool nrf_802154_core_last_rssi_measurement_get(int8_t * p_rssi)
     return result;
 }
 
-uint8_t nrf_802154_core_last_lqi_get(void)
-{
-    return m_last_lqi;
-}
-
 #if NRF_802154_INTERNAL_RADIO_IRQ_HANDLING
 void RADIO_IRQHandler(void)
 #else // NRF_802154_INTERNAL_RADIO_IRQ_HANDLING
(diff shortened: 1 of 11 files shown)


Irving-cl:
I implemented it as discussed in https://github.com/openthread/openthread/pull/5780#issuecomment-732797975

I have verified Enh-ACK Based Probing on 2 nrf52840 DKs. LQI value seems correct. However, `nrf_802154_rssi_last_get()` fails and return invalid value (`NRF_802154_RSSI_INVALID`, namely `INT8_MAX`, 127). What could be the reason for invalid rssi? @edmont

edmont:
@Irving-cl the problem is it's not allowed to make requests to the 802.15.4 driver inside `nrf_802154_tx_ack_started`. Please try out the attached patch for your PR, this approach should work.

[fix_rssi.txt](https://github.com/openthread/openthread/files/5765934/fix_rssi.txt)

third_party/NordicSemiconductor/drivers/radio/nrf_802154_core.c
```diff
@@ -3262,6 +3265,11 @@ bool nrf_802154_core_last_rssi_measurement_get(int8_t * p_rssi)
     return result;
 }
 
+uint8_t nrf_802154_core_last_lqi_get(void)
```
edmont, line 3268: It seems we don't need this anymore.

edmont (approved):
Thanks, this works for me now.

konradderda:
Hello, I can see a potential issue with the current approach to use `nrf_802154_tx_ack_started` for such operations. As well as for the changes introduced in this function with https://github.com/openthread/openthread/pull/5184.

This function is called from an interrupt when the ACK is already being sent. Even if you set a breakpoint in this function the RADIO peripheral will continue sending the ACK. This can lead to a race condition with a root cause similar to the one described here: https://github.com/openthread/openthread/pull/5837.

To my mind, it would be better to modify ACKs when right after they are generated and before transmission is set up. I have an idea of how to deal with it, but I think it should be introduced in a separate PR. If this suits you, I can submit it soon.

Irving-cl:
@konradderda 

Thanks for reminding the issue. I think we do need to update something(CSL IE) in ack right after it's being sent because the spec defines the value as the time when first symbol starts to send. And then since the frame changed, if the frame's security is enabled, security needs to be processed.

For other things like Vendor IE in this PR, it would be good to set them right after the ack was generated.

So do you mean that, if we do the update of the ack in `nrf_802154_tx_ack_started`, it's possible that the frame has been sent out before the update was made. Am I right about this? I remember that our idea is, we only update some IEs and MIC(integrity code for security). And they're in back position of an ack. So in most cases the update would work before the data was sent out. But I'm not 100% sure. What do you think about this? @hubertmis @edmont @brianljt 

Anyway I would agree to update other information (like Vendor IE in this PR) for Ack right after it's generated. And we should try to do less thing in `nrf_802154_tx_ack_started` as possible. So it would be great if you submit a separate PR for it.

brianljt:
@konradderda 

I think in our test, nrf528xx software mbedtls library is fast enough so we did not see corrupted frame sent due to racing condition except #5837. In #5837, firmware needs to prepare AES tables for the first time which took a long time. I think this could be resolved by initializing AES table right after boot up or use ROM AES table instead.

But I agree, it is still possible to hit racing condition especially when MAC header opening payload is too short. It would be ideal if hardware could do encryption and MIC calculation on the fly when doing transmission like what you have for bluetooth in nrf528xx (nrf52840 spec 6.4.5 Encrypting packets on-the-fly in radio transmit mode). However, this hardware block does not support AES-CCM* used in Thread. I remember someone mentioned about using ECB block to do this. If you have an idea/PR for this, it would be great. I can help take a look at it and provide inputs from application/protocol perspective if need. Thank you!

hubertmis:
I agree with the comments above:

> This function is called from an interrupt when the ACK is already being sent. Even if you set a breakpoint in this function the RADIO peripheral will continue sending the ACK. This can lead to a race condition

> Anyway I would agree to update other information (like Vendor IE in this PR) for Ack right after it's generated. And we should try to do less thing in `nrf_802154_tx_ack_started` as possible

> However, this hardware block does not support AES-CCM* used in Thread. I remember someone mentioned about using ECB block to do this.

There is one more important problem: in nRF53 OpenThread is running in a separated core than the radio driver. Because of that `nrf_802154_tx_ack_started` does not have access to the buffer containing ACK being sent and latency of IPC prevents ACK buffer modifications by OpenThread.

The solution we propose is to add modules to 802.15.4 radio driver that update ACK data (CLS Phase, Link Metrics payload, MIC) on-the-fly using ECB peripheral. OpenThread would just use `nrf_802154_ack_data_set()` function to set IEs and if 802.15.4 driver detects CSL IE or Vendor IE including Link Metrics data, the driver would take care of its content.

We have a working PoC of CSL IEs injector and MIC generator, but we are not focused right now on this feature. We'll finalize it when we'll be preparing Thread 1.2 support for nRF53.

Irving-cl:
> The solution we propose is to add modules to 802.15.4 radio driver that update ACK data (CLS Phase, Link Metrics payload, MIC) on-the-fly using ECB peripheral. OpenThread would just use nrf_802154_ack_data_set() function to set IEs and if 802.15.4 driver detects CSL IE or Vendor IE including Link Metrics data, the driver would take care of its content.

Sounds good to me. Then I think we can leave the implementation for nRF52 as it is for now.

jwhui (approved):

PR #6025, opened 2020-12-31
[utils] implement a item-pointer iterator template for many iterators

Irving-cl:
Many Iterators have the same pattern in our code: they have `++`, deference, equal operators, `IsDone` method and hold a pointer to the an object. This PR extracts the common pattern into a template class `ItemPtrIterator` and use it as a base class for those similar Iterator classes.

### DIFF
--- a/src/core/backbone_router/ndproxy_table.hpp
+++ b/src/core/backbone_router/ndproxy_table.hpp
@@ -40,6 +40,7 @@
 #include <openthread/backbone_router_ftd.h>
 
 #include "backbone_router/bbr_leader.hpp"
+#include "common/iterator_utils.hpp"
 #include "common/locator.hpp"
 #include "common/non_copyable.hpp"
 #include "common/time.hpp"
@@ -246,8 +247,9 @@ private:
      * This class represents an iterator for iterating through the NdProxy Table.
      *
      */
-    class Iterator : public InstanceLocator
+    class Iterator : public InstanceLocator, public ItemPtrIterator<NdProxy, Iterator>
     {
+        friend class ItemPtrIterator<NdProxy, Iterator>;
         friend class NdProxyTable;
         friend class IteratorBuilder;
 
@@ -260,17 +262,9 @@ private:
         Iterator(Instance &aInstance, Filter aFilter);
         Iterator(Instance &aInstance, IteratorType);
 
-        bool           IsDone(void) const { return (mCurrent == nullptr); }
-        void           Advance(void);
-        void           operator++(void) { Advance(); }
-        void           operator++(int) { Advance(); }
-        const NdProxy &operator*(void)const { return *mCurrent; }
-        bool           operator==(const Iterator &aOther) const { return mCurrent == aOther.mCurrent; }
-        bool           operator!=(const Iterator &aOther) const { return !(*this == aOther); }
-        NdProxy *      operator->(void) { return mCurrent; }
-        NdProxy &      operator*(void) { return *mCurrent; }
+        void Advance(void);
 
-        NdProxy *mCurrent;
+        NdProxy *mItem;
         Filter   mFilter;
     };
 
--- a/src/core/backbone_router/ndproxy_table.hpp
+++ b/src/core/backbone_router/ndproxy_table.hpp
@@ -264,8 +264,7 @@ private:
 
         void Advance(void);
 
-        NdProxy *mItem;
-        Filter   mFilter;
+        Filter mFilter;
     };
 
     class IteratorBuilder : public InstanceLocator
--- a/src/core/common/iterator_utils.hpp
+++ b/src/core/common/iterator_utils.hpp
@@ -55,13 +55,30 @@ namespace ot {
  * Users of this class should follow CRTP-style inheritance, i.e., `IteratorType` class itself should publicly inherit
  * from `ItemPtrIterator<ItemType, IteratorType>`. `ItemType` is the type of the object that `IteratorType` points to.
  *
- * The template type `IteratorType` should contain a `mItem` member variable  which is a pointer of `ItemType`,
- * and have a method `Advance()`.
+ * The template type `IteratorType` MUST have a method `Advance()`.
  *
  */
 template <class ItemType, class IteratorType> class ItemPtrIterator
 {
 public:
+    /**
+     * Default constructor
+     *
+     */
+    ItemPtrIterator(void)
+        : mItem(nullptr)
+    {
+    }
+
+    /**
+     * Contructor with an Item pointer.
+     *
+     */
+    explicit ItemPtrIterator(ItemType *item)
+        : mItem(item)
+    {
+    }
+
     /**
      * This method indicates whether there are no more items to be accessed (iterator has reached the end).
      *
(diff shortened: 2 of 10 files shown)


src/core/backbone_router/ndproxy_table.hpp
```diff
@@ -260,17 +262,9 @@ class NdProxyTable : public InstanceLocator, private NonCopyable
...
-        const NdProxy &operator*(void)const { return *mCurrent; }
-        bool           operator==(const Iterator &aOther) const { return mCurrent == aOther.mCurrent; }
-        bool           operator!=(const Iterator &aOther) const { return !(*this == aOther); }
-        NdProxy *      operator->(void) { return mCurrent; }
-        NdProxy &      operator*(void) { return *mCurrent; }
-
-        NdProxy *mCurrent;
+        void Advance(void);
+
+        NdProxy *mItem;
```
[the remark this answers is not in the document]
Irving-cl: I'm afraid not. As I understand, this CRTP pattern has to be an 'interface' and cannot have any members. Note that:
```
bool IsDone(void) const { return (static_cast<const IteratorType *>(this)->mItem == nullptr); }
```
It `static_cast` itself to its derived class. It may go wrong if this base class has some members. Looking at `LinkedListEntry` and `Timer` in current code, the member `mNext` is also in `Timer` instead of `LinkedListEntry`. Thoughts?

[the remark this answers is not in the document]
abtink: I think we can use CRTP model can have members as well. 

`static_cast` between a base class and sub-class should always be safe and work correctly,  no matter if the base class has member variable or not (or if subclass has other base classes).  In `static_cast` the compiler will actually correctly handle the pointer conversion and may change the raw pointer address value when casting from base to sub class (e.g., if the base and sub-class are not aligned in memory, i.e. don't start from same address). The only requirement when casting from base to sub class is to be sure that the pointer to base class we have is indeed coming from a subclass instance, and I think this would be always the case in CRTP inheritance. `reinterpret_cast` on other hand won't be correct and/or safe.

For `LinkedListEntry` the reason we could not define the `mNext` in this it, is that in some of the cases the type itself is part of public OT APIs (which is a C based and is a `struct`) and therefore already defines the `mNext` field (e.g. `otNetifAddress).

I think with this type, we intend to use it only in the core and for C++ types, so we can consider having it define the `mItem` itself. We probably need it to be `protected` so that the inheriting type can also access it (it probably needs to implement `Advance()`).

src/core/common/iterator_utils.hpp
```diff
@@ -0,0 +1,152 @@
...
+ * object.
+ *
+ * Users of this class should follow CRTP-style inheritance, i.e., `IteratorType` class itself should publicly inherit
+ * from `ItemPtrIterator<ItemType, IteratorType>`. `ItemType` is the type of the object that `IteratorType` points to.
+ *
+ * The template type `IteratorType` should contain a `mItem` member variable  which is a pointer of `ItemType`,
+ * and have a method `Advance()`.
+ *
+ */
+template <class ItemType, class IteratorType> class ItemPtrIterator
```
abtink, line 62: One idea we can consider that may make this a bit simpler to just have one template parameter `IteratorType`.

Then we expect/require the `ItemType`  to be define by `IteratorType` itself (i.e., be a nested `typedef` in it), same way that we require it to have `Advance()`method.

In the implementation here, we can simply refer to it a `IteratorType::ItemType`. 

I also like the idea of having `mItem` be defined by this class itself.

Thoughts?

abtink (commented):
@Irving-cl  this is a great enhancement. Thank you 👍 👍 

I added some suggestion below on having `ItemType` be a nested type defined by `IteratorType`. I also like the suggestion to include `mItem` in the `ItemPtrIterator`. Thoughts?

Thanks.

src/core/common/iterator_utils.hpp
Irving-cl: Sounds good to me, thanks!

Irving-cl:
Well, I moved `mItem` to the base class. However, it seems difficult to use only one template parameter and define it in the derived class because of the **incomplete type** issue. As pointed in this [thread](https://stackoverflow.com/questions/36268400/crtp-refer-to-typedef-in-derived-class-from-base-class):
image

I also found a [thread](https://stackoverflow.com/questions/6006614/c-static-polymorphism-crtp-and-using-typedefs-from-derived-classes) using traits workaround for similar case. However I think it doesn't fit our case since we don't use template derived class (the iterators). We need different derived class so they have their own `Advance` method.

Thoughts? @abtink

abtink:
@Irving-cl thanks for looking into this. Yes it makes sense to need a complete type.

So let's have two template parameters and pass `ItemType` as template augment (thinking more about this, I feel it may even be better and the intention more clear with passing as template param).

src/core/common/iterator_utils.hpp
```diff
@@ -0,0 +1,166 @@
...
+ * This template class is used as a base class for those item-pointer iterators.
+ *
+ * These iterators have common methods and operators like `Advance()` and `++` and hold a pointer to the
+ * object.
+ *
+ * Users of this class should follow CRTP-style inheritance, i.e., `IteratorType` class itself should publicly inherit
+ * from `ItemPtrIterator<ItemType, IteratorType>`. `ItemType` is the type of the object that `IteratorType` points to.
+ *
+ * The template type `IteratorType` MUST have a method `Advance()`.
+ *
```
abtink, line 59: Add `@tparam` documentation?

src/core/common/iterator_utils.hpp
```diff
@@ -0,0 +1,166 @@
...
+ */
+template <class ItemType, class IteratorType> class ItemPtrIterator
+{
+public:
+    /**
+     * Default constructor
+     *
+     */
+    ItemPtrIterator(void)
```
abtink, line 68: Can consider define the constructor under `protected` section...
this way this can only be sub-classed.

abtink (approved):
Looks great. Thanks 👍

src/core/common/iterator_utils.hpp
```diff
@@ -0,0 +1,167 @@
...
+ * This template class is used as a base class for those item-pointer iterators.
+ *
+ * These iterators have common methods and operators like `Advance()` and `++` and hold a pointer to the
+ * object.
+ *
+ * Users of this class should follow CRTP-style inheritance, i.e., `IteratorType` class itself should publicly inherit
+ * from `ItemPtrIterator<ItemType, IteratorType>`.
+ *
+ * @tparam  ItemType      The type of the object that the iterator points to.
+ * @tparam  IteratorType  The Iterator class that inherits this class. The class MUST have a method `Advance()`.
```
simonlingoogle, line 59: I think we should be clear about the protocol of Advance(). It should move the `mItem` pointer, which will never be `nullptr` when Advance() is called, to the next element, and be set to nullptr when there are no more elements.

simonlingoogle (approved):
LGTM :+1: 

with a suggestion

jwhui (approved):

PR #6035, opened 2021-01-04
[scripts] optionally install pretty format for Raspbian in bootstrap

Irving-cl:
It seems to be unnecessary to install these format dependencies when setting up OpenThread on a Raspbian (Mostly people don't develop or submit PR from a Raspberry pi). So by default don't install these format dependencies on Raspbian to save time.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -33,6 +33,23 @@
 
 set -euxo pipefail
 
+install_packages_apt_pretty_format()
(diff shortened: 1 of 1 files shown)


simonlingoogle (approved):
LGTM 👍 

The change make sense to me.

jwhui (approved):

PR #6040, opened 2021-01-05
[build] remove unneeded-internal-declaration warning on Fuchsia

paragdixit-g:
By inlining static functions defined in the header file.

Since this increases the function size of functions which use the definition, an alternative is to define these functions as function templates with unused template parameters. This ensures that functions are created only in the files where they are used. Please let me know if I should do that instead.


abtink (approved):
LGTM.

jwhui (approved):

PR #6043, opened 2021-01-05
[ncp] increase size of NCP UART buffers to avoid dropped frames

jwhui:
Operational experience shows that having increased UART TX buffers on the RCP can help reduce the likelihood of dropping frames when sending from RCP to host - see https://github.com/openthread/ot-br-posix/issues/559#issuecomment-721982478.

Credit to @dougleaf and @gabekassel for identifying this issue.


bukepo (approved):
LGTM:+1:

PR #6044, opened 2021-01-06
[posix] tune core config for posix

simonlingoogle:
POSIX platform is generally less resource constrained.
This commit tune several OpenThread core configurations to better fit the POSIX platform.

### DIFF
--- a/src/posix/platform/openthread-core-posix-config.h
+++ b/src/posix/platform/openthread-core-posix-config.h
@@ -136,4 +136,84 @@
 #define OPENTHREAD_CONFIG_LOG_MAX_SIZE 1024
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_COMMISSIONER_MAX_JOINER_ENTRIES
+ *
+ * The maximum number of Joiner entries maintained by the Commissioner.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_COMMISSIONER_MAX_JOINER_ENTRIES
+#define OPENTHREAD_CONFIG_COMMISSIONER_MAX_JOINER_ENTRIES 4
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_ENTRIES
+ *
+ * The number of EID-to-RLOC cache entries.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_ENTRIES
+#define OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_ENTRIES 32
+#endif
+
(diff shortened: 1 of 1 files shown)


src/posix/platform/openthread-core-posix-config.h
```diff
@@ -136,4 +136,94 @@
...
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_MLE_MAX_CHILDREN
+ *
+ * The maximum number of children.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_MLE_MAX_CHILDREN
+#define OPENTHREAD_CONFIG_MLE_MAX_CHILDREN 32
```
bukepo, line 166:
```suggestion
#define OPENTHREAD_CONFIG_MLE_MAX_CHILDREN 64
```

src/posix/platform/openthread-core-posix-config.h
```diff
@@ -136,4 +136,94 @@
...
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE
+ *
+ * The size of heap buffer when DTLS is enabled.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE
+#define OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE (256 * 1024)
```
bukepo, line 216:
```suggestion
#define OPENTHREAD_CONFIG_HEAP_INTERNAL_SIZE (256 * 1024 * sizeof(void*))
```

simonlingoogle: This change would increase the heap size to 1M for 32bit and 2M for 64bit
Is that what you intended to have?

bukepo: Just to take pointers into consideration. You may choose a proper value based on the word length.

wgtdkp (approved):
LGTM :+1:

src/posix/platform/openthread-core-posix-config.h
simonlingoogle: Turnout the heap size is restricted to less than 64K for the current heap implementation because we are using `uint16_t` to track heap block sizes. 

So, I changed the heap size 63K for both 32bit and 64bit. We can remove the heap size limit in a future PR.

Irving-cl (approved):
LGTM

simonlingoogle:
Further comments for this PR ? @bukepo @librasungirl

bukepo (approved):

librasungirl (approved):

jwhui (approved):

PR #6045, opened 2021-01-06
[test] enhance parsing addresses

bukepo:
Expect sometimes matches the pattern before a line is fully printed to
TTY. As a result, the address returned may not be accurate. This commit
fixes this issue by matching line breaks.


tests/scripts/expect/_common.exp
```diff
@@ -149,17 +151,17 @@ proc dispose_all {} {
 proc get_ipaddr {type} {
     send "ipaddr $type\n"
     expect "ipaddr $type"
-    expect -re {(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4})}
-    set rval $expect_out(1,string)
+    expect -re {([\r\n ])(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4})([\r\n>])}
+    set rval $expect_out(2,string)
```
Moandor-y, line 155:
```suggestion
    expect -re {[\r\n ](([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4})[\r\n>]}
    set rval $expect_out(1,string)
```
These capture groups are unnecessary.

tests/scripts/expect/_common.exp
```diff
@@ -149,17 +151,17 @@ proc dispose_all {} {
...
 
     return $rval
 }
 
 proc get_extaddr {} {
     send "extaddr\n"
-    expect -re {([0-9a-fA-F]{16})}
-    set rval $expect_out(1,string)
+    expect -re {([\r\n ])([0-9a-fA-F]{16})([\r\n>])}
+    set rval $expect_out(2,string)
```
Moandor-y, line 164:
```suggestion
    expect -re {[\r\n ]([0-9a-fA-F]{16})[\r\n>]}
    set rval $expect_out(1,string)
```

tests/scripts/expect/_common.exp
```diff
@@ -168,8 +170,8 @@ proc get_extaddr {} {
 proc get_rloc16 {} {
     send "rloc16\n"
     expect "rloc16"
-    expect -re {([0-9a-fA-F]{4})}
-    set rval $expect_out(1,string)
+    expect -re {([\r\n ])([0-9a-fA-F]{4})([\r\n>])}
+    set rval $expect_out(2,string)
```
Moandor-y, line 174:
```suggestion
    expect -re {[\r\n ]([0-9a-fA-F]{4})[\r\n>]}
    set rval $expect_out(1,string)
```

tests/scripts/expect/_common.exp
```diff
@@ -168,7 +170,7 @@ proc get_extaddr {} {
 proc get_rloc16 {} {
     send "rloc16\n"
     expect "rloc16"
-    expect -re {([0-9a-fA-F]{4})}
+    expect -re {[\r\n ]([0-9a-fA-F]{4})[\r\n>]}
```
simonlingoogle, line 173: define a new routine like `expect_line` ?

tests/scripts/expect/_common.exp
```diff
@@ -168,7 +170,7 @@ proc get_extaddr {} {
...
     set rval $expect_out(1,string)
     expect "Done"
```
simonlingoogle, line 175:
```suggestion
    expect -re "{[\r\n ]Done[\r\n]}"
```
is it necessary to do so?

tests/scripts/expect/_common.exp
```diff
@@ -158,7 +160,7 @@ proc get_ipaddr {type} {
 
 proc get_extaddr {} {
     send "extaddr\n"
-    expect -re {([0-9a-fA-F]{16})}
+    expect -re {[\r\n ]([0-9a-fA-F]{16})[\r\n>]}
```
simonlingoogle, line 163:
```suggestion
    expect -re {[\r\n]([0-9a-fA-F]{16})[\r\n>]}
```
Why do we add space before the extaddr?

simonlingoogle (approved):
LGTM :+1: 

with nits

tests/scripts/expect/_common.exp
bukepo: sometimes `> ` comes before line break.

simonlingoogle: I see.

tests/scripts/expect/_common.exp
simonlingoogle: I realize that this is problematic because the  `\r\n` between extended address output and Done would have been consumed by `expect -re {[\r\n ]([0-9a-fA-F]{4})[\r\n>]}`, so `"{[\r\n ]Done[\r\n]}"` won't match the next line of `Done`.

I am fine to keep it as is.

Moandor-y: We can use a positive lookahead to assert the existence of newline without consuming it.

E.g.
```
expect -re {[\r\n ]([0-9a-fA-F]{4})(?=[\r\n>])}
```

`expect` doesn't seem to support lookbehind though. So only the newline after the pattern can be asserted this way, the newline before the pattern has to be comsumed.

bukepo: OpenThread prints `\r\n` for line break, `[\r\n>]` consumes at most a single `\r`. And lookbehind looks like a safer solution.

Moandor-y (approved):
LGTM 👍

jwhui (approved):

PR #6046, opened 2021-01-06
[radio-spinel] add option to skip RCP compatibility check during init

### LINKED ISSUE #6042: Add command line argument to ignore RCP API version checking
OT-Daemon can be used to manage RCP firmware on the coprocessor. However, in the case the RCP API version is incompatible, ot-daemon will fail to start (and thus block any diag commands used to put RCP into DFU). 

**Describe the solution you'd like** A clear and concise description of what you want to happen.
This request is to add a cli argument (e.g. `ignorecompatibility`) that ignores any version checks. 

**Describe alternatives you've considered** A clear and concise description of any alternative solutions or features you've considered.
It's plausible to implement vendor-specific Spinel commands and either allow ot-daemon to pass them in at invocation (different cli arg) or use a separate tool like PySpinel. 

**Additional context** Add any other context or screenshots about the feature request here.

abtink:
This commit adds new option to skip RCP API version and capability
compatibility check during initialization. The radio URL key
`skip-rcp-compatibility-check` can be used to enable this behavior.
This is only intended for situations where RCP won't be operating as a
radio (e.g., host trying to perform an RCP firmware update).

----

Resolves #6042

### DIFF
--- a/src/lib/spinel/radio_spinel_impl.hpp
+++ b/src/lib/spinel/radio_spinel_impl.hpp
@@ -220,7 +220,9 @@ RadioSpinel<InterfaceType, ProcessContextType>::RadioSpinel(void)
 }
 
 template <typename InterfaceType, typename ProcessContextType>
-void RadioSpinel<InterfaceType, ProcessContextType>::Init(bool aResetRadio, bool aRestoreDatasetFromNcp)
+void RadioSpinel<InterfaceType, ProcessContextType>::Init(bool aResetRadio,
+                                                          bool aRestoreDatasetFromNcp,
+                                                          bool aSkipRcpCompatibilityCheck)
 {
     otError error = OT_ERROR_NONE;
     bool    supportsRcpApiVersion;
@@ -253,8 +255,11 @@ void RadioSpinel<InterfaceType, ProcessContextType>::Init(bool aResetRadio, bool
         DieNow(exitCode);
     }
 
-    SuccessOrDie(CheckRcpApiVersion(supportsRcpApiVersion));
-    SuccessOrDie(CheckRadioCapabilities());
+    if (!aSkipRcpCompatibilityCheck)
+    {
+        SuccessOrDie(CheckRcpApiVersion(supportsRcpApiVersion));
+        SuccessOrDie(CheckRadioCapabilities());
+    }
 
     mRxRadioFrame.mPsdu  = mRxPsdu;
     mTxRadioFrame.mPsdu  = mTxPsdu;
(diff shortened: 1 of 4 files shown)


bukepo (approved):
LGTM:+1:

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -253,7 +255,11 @@ void RadioSpinel<InterfaceType, ProcessContextType>::Init(bool aResetRadio, bool
         DieNow(exitCode);
     }
 
-    SuccessOrDie(CheckRcpApiVersion(supportsRcpApiVersion));
+    if (!aSkipRcpApiVersionCheck)
+    {
+        SuccessOrDie(CheckRcpApiVersion(supportsRcpApiVersion));
+    }
+
     SuccessOrDie(CheckRadioCapabilities());
```
simonlingoogle, line 263: We also check RCP capabilities (via CheckRadioCapabilities). Should we also provide arguments to bypass it for cases like firmware update?

src/posix/platform/radio_url.cpp
```diff
@@ -92,7 +92,8 @@ const char *otSysGetRadioUrlHelpString(void)
            "    cca-threshold[=dbm]           Set the radio's CCA ED threshold in dBm measured at antenna connector.\n"
            "    fem-lnagain[=dbm]             Set the Rx LNA gain in dBm of the external FEM.\n"
            "    ncp-dataset                   Retrieve dataset from ncp.\n"
-           "    no-reset                      Do not send Spinel reset command to RCP on initialization.\n";
+           "    no-reset                      Do not send Spinel reset command to RCP on initialization.\n"
+           "    skip-api-version-check        Skip checking RCP API version during initialization.\n";
```
simonlingoogle, line 96:
```suggestion
           "    skip-compatibility-check        Skip checking RCP API version during initialization.\n";
```
make it more general and also skip `CheckRadioCapabilities` if specified ?

src/lib/spinel/radio_spinel_impl.hpp
abtink: I was initially considering to include the radio cap check, but decided not to do this, to keep the change simple. I was concerned that not doing `CheckRadioCapabilities` would keep `mRadioCaps` as empty or zero and if it may have side-effects (it would impact not just `RadioSpinel` but OT core and sub-mac). 

Looking now at all the places `mRadioCaps` is used I think it should be fine, so will change it.

abtink: Changed in new pushed commit.

simonlingoogle (approved):
LGTM :+1: 

thanks

jwhui (approved):

PR #6049, opened 2021-01-08
[github-actions] upload core dump only when crash happens

Irving-cl:
Currently when a test fails, the core dump uploading action would be executed. (Even when there's no core dump, the binaries and so libs would be uploaded, which is a waste.)

This PR uses env variables of github actions to only execute the uploading action when crash actually happens.

### DIFF
--- a/.github/workflows/simulation-1.2.yml
+++ b/.github/workflows/simulation-1.2.yml
@@ -90,12 +90,9 @@ jobs:
     - uses: actions/upload-artifact@v2
       if: ${{ failure() && env.CRASHED == '1' }}
       with:
-        name: core-packet-verification-low-power
+        name: core-packet-verification-thread-1-2
         path: |
-          ./build/openthread-simulation-1.2/examples/apps/cli/ot-cli-*
-          ./build/openthread-simulation-1.1/examples/apps/cli/ot-cli-*
           ./ot-core-dump/*
-          ./so-lib/*
     - name: Generate Coverage
       run: |
         ./script/test generate_coverage "${{ matrix.compiler.gcov }}"
@@ -155,10 +152,7 @@ jobs:
       with:
         name: core-packet-verification-low-power
         path: |
-          ./build/openthread-simulation-1.2/examples/apps/cli/ot-cli-*
-          ./build/openthread-simulation-1.1/examples/apps/cli/ot-cli-*
           ./ot-core-dump/*
-          ./so-lib/*
     - name: Generate Coverage
       run: |
         ./script/test generate_coverage gcc
@@ -239,11 +233,7 @@ jobs:
       with:
         name: core-expect-1-2
         path: |
-          ./build/openthread-simulation-1.2/examples/apps/cli/ot-cli-mtd
-          ./build/openthread-simulation-1.2/examples/apps/ncp/ot-rcp
-          ./build/openthread-posix-1.2/src/posix/ot-cli
           ./ot-core-dump/*
-          ./so-lib/*
     - name: Generate Coverage
       run: |
         ./script/test generate_coverage gcc
--- a/.github/workflows/simulation-1.2.yml
+++ b/.github/workflows/simulation-1.2.yml
@@ -76,15 +76,26 @@ jobs:
         ./script/test prepare_coredump_upload
         ./script/test unit
         ./script/test cert_suite tests/scripts/thread-cert/v1_2_*
-    - name: Copy Shared Libraries
+    - name: Check Crash
       if: ${{ failure() }}
       run: |
-          ./script/test copy_so_lib
+          CRASHED=$(./script/test check_crash | tail -1)
+          [[ $CRASHED -eq "1" ]] && echo "Crashed!" || echo "Not crashed."
+          echo "CRASHED=$CRASHED" >> $GITHUB_ENV
     - uses: actions/upload-artifact@v2
       if: ${{ failure() }}
       with:
         name: thread-1-2-${{ matrix.compiler.c }}-${{ matrix.arch }}-pcaps
         path: "*.pcap"
+    - uses: actions/upload-artifact@v2
+      if: ${{ failure() && env.CRASHED == '1' }}
+      with:
+        name: core-packet-verification-low-power
+        path: |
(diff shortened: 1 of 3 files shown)


Irving-cl:
I've verified that the new approach works.

Moandor-y (commented):
Also change this for other workflows?

Irving-cl:
> Also change this for other workflows?

Updated. Missed it before. Thanks!

@Moandor-y

.github/workflows/simulation.yml
```diff
@@ -320,6 +322,22 @@ jobs:
         ulimit -c unlimited
         ./script/test prepare_coredump_upload
         OT_OPTIONS='-DOT_READLINE=OFF -DOT_FULL_LOGS=ON -DOT_LOG_OUTPUT=PLATFORM_DEFINED' VIRTUAL_TIME=0 OT_NODE_TYPE=rcp ./script/test clean build expect
+    - name: Check Crash
+      if: ${{ failure() }}
+      run: |
+          CRASHED=$(./script/test check_crash | tail -1)
+          [[ $CRASHED -eq "1" ]] && echo "Crashed!" || echo "Not crashed."
+          echo "CRASHED=$CRASHED" >> $GITHUB_ENV
```
Moandor-y, line 330: Here we write this environment variable multiple times in one job. Did you verify the behavior in such situations?

.github/workflows/simulation.yml
```diff
@@ -320,6 +322,22 @@ jobs:
...
+    - uses: actions/upload-artifact@v2
+      if: ${{ failure() && env.CRASHED == '1' }}
+      with:
+        name: core-expect-rcp
+        path: |
+          ./build/openthread-simulation-1.2/examples/apps/cli/ot-cli-mtd
+          ./build/openthread-simulation-1.2/examples/apps/ncp/ot-rcp
+          ./build/openthread-posix-1.2/src/posix/ot-cli
```
Moandor-y, line 338:
```suggestion
          ./build/openthread-simulation-1.1/examples/apps/cli/ot-cli-mtd
          ./build/openthread-simulation-1.1/examples/apps/ncp/ot-rcp
          ./build/openthread-posix-1.1/src/posix/ot-cli
```

.github/workflows/simulation.yml
Irving-cl: Good catch! I'll do some investigation on this.

Irving-cl: I used a simple solution for this: add a postfix for the variable in different runs in the same job. Then there won't be any conflicts. Thoughts?

.github/workflows/simulation-1.2.yml
```diff
@@ -76,15 +76,26 @@ jobs:
...
         path: "*.pcap"
+    - uses: actions/upload-artifact@v2
+      if: ${{ failure() && env.CRASHED == '1' }}
+      with:
+        name: core-packet-verification-low-power
+        path: |
+          ./build/openthread-simulation-1.2/examples/apps/cli/ot-cli-*
+          ./build/openthread-simulation-1.1/examples/apps/cli/ot-cli-*
+          ./ot-core-dump/*
+          ./so-lib/*
```
simonlingoogle, line 98:
```suggestion
    - uses: actions/upload-artifact@v2
      if: ${{ failure() }}
      with:
        name: core-packet-verification-low-power
        path: |
          ./ot-core-dump/*
```
Can we copy everything we need to `ot-core-dump/` (including ot-cli-* executables and so-lib/*) when core dumped and clear ot-core-dump otherwise? So that we automatically uploads nothing if there was no core dump.

Moandor-y (approved):
LGTM :+1:

.github/workflows/simulation-1.2.yml
```diff
@@ -76,15 +76,26 @@ jobs:
...
         ./script/test unit
         ./script/test cert_suite tests/scripts/thread-cert/v1_2_*
-    - name: Copy Shared Libraries
+    - name: Check Crash
       if: ${{ failure() }}
       run: |
-          ./script/test copy_so_lib
+          CRASHED=$(./script/test check_crash | tail -1)
+          [[ $CRASHED -eq "1" ]] && echo "Crashed!" || echo "Not crashed."
+          echo "CRASHED=$CRASHED" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
Irving-cl: I'd prefer to have an output to indicate whether Crash happens (more obvious), thoughts?

.github/workflows/simulation-1.2.yml
Irving-cl: Put everything into one dir SGTM. While I still prefer to use the env variable. In this way, the upload action won't be started at all (Otherwise, we can still see the action processed, and it fails to find anything in the dir). I think that would be a stronger signal telling us that there's no crash. Thoughts?

simonlingoogle: I was hoping that this way we can make the github actions jobs simpler --- we only need to run `./script/test check_crash` and upload one artifact directory (i.e. ./ot-core-dump/) after that.
Thoughts?

Irving-cl: I mean I agree to put everything into one directory. And then in `script/test` we can use `file` command to scan the core dump files and collect what binaries crashed, putting them into the directory. In this way we can make these jobs in simplest way.

My point is just that we can skip the unnecessary upload action (which is apparently faster than running it though not uploading anything) by:
```
- if: ${{ failure() }}
+ if: ${{ failure() && env.CRASHED == '1'}}
```

So the line numbers don't change. Thoughts?

.github/workflows/simulation-1.2.yml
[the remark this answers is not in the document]
simonlingoogle: fine to me

.github/workflows/simulation-1.2.yml
simonlingoogle: This approach SGTM.

Irving-cl:
Testing.

Irving-cl:
I've verified.

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6052, opened 2021-01-09
[network-data] check serverTlv data length before parsing

abtink:
This commit adds a check in `GetBackboneRouterPrimary()` to verify
that the server sub-TLV data length is larger than an expected minimum
before casting it to `BackboneRouterServerData` and parsing/reading the
info.

### DIFF
--- a/src/core/thread/network_data_leader.cpp
+++ b/src/core/thread/network_data_leader.cpp
@@ -110,8 +110,14 @@ otError LeaderBase::GetBackboneRouterPrimary(BackboneRouter::BackboneRouterConfi
     for (const NetworkDataTlv *start                                                      = serviceTlv->GetSubTlvs();
          (serverTlv = FindTlv<ServerTlv>(start, serviceTlv->GetNext())) != nullptr; start = serverTlv->GetNext())
     {
-        const BackboneRouterServerData *serverData =
-            reinterpret_cast<const BackboneRouterServerData *>(serverTlv->GetServerData());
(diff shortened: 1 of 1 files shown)


wgtdkp (approved):
LGTM 👍

src/core/thread/network_data_leader.cpp
```diff
@@ -110,8 +110,14 @@ otError LeaderBase::GetBackboneRouterPrimary(BackboneRouter::BackboneRouterConfi
     for (const NetworkDataTlv *start                                                      = serviceTlv->GetSubTlvs();
          (serverTlv = FindTlv<ServerTlv>(start, serviceTlv->GetNext())) != nullptr; start = serverTlv->GetNext())
     {
-        const BackboneRouterServerData *serverData =
-            reinterpret_cast<const BackboneRouterServerData *>(serverTlv->GetServerData());
+        const BackboneRouterServerData *serverData;
+
+        if (serverTlv->GetServerDataLength() < sizeof(BackboneRouterServerData))
```
jwhui, line 115: Can the length be equal?
```suggestion
        if (serverTlv->GetServerDataLength() <= sizeof(BackboneRouterServerData))
```

abtink: I think if the server datalen is equal then if it fine to cast and parse it and we don't want to skip it.

jwhui (approved):

PR #6057, opened 2021-01-11
[udp] platform UDP sending depending on the 'IsHostInterface' flag

wgtdkp:
The problem with current implementation is that:
When `PLATFORM_UDP` is enabled and a custom UDP server is started with `socket.Bind(0)`, the server can receives datagrams from both host interface and Thread interface, but the responses are always sent via the platform UDP.

I think whether a UDP message should be sent via host interface or Thread interface should be determined by the `IsHostInterface` flag.

### DIFF
--- a/src/core/net/udp6.cpp
+++ b/src/core/net/udp6.cpp
@@ -372,7 +372,7 @@ otError Udp::SendTo(SocketHandle &aSocket, Message &aMessage, const MessageInfo 
     messageInfoLocal.SetSockPort(aSocket.GetSockName().mPort);
 
 #if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
-    if (ShouldUsePlatformUdp(aSocket))
+    if (ShouldUsePlatformUdp(aSocket) && messageInfoLocal.IsHostInterface())
     {
         SuccessOrExit(error = otPlatUdpSend(&aSocket, &aMessage, &messageInfoLocal));
     }


bukepo (approved):
LGTM:+1:

src/core/net/udp6.cpp
```diff
@@ -372,7 +372,7 @@ otError Udp::SendTo(SocketHandle &aSocket, Message &aMessage, const MessageInfo
     messageInfoLocal.SetSockPort(aSocket.GetSockName().mPort);
 
 #if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
-    if (ShouldUsePlatformUdp(aSocket))
+    if (ShouldUsePlatformUdp(aSocket) && messageInfoLocal.IsHostInterface())
```
simonlingoogle, line 375: Messages that are sent to Thread devices are also passed to Host tun interface in PLAT_UDP implementation.
So I think we should not check for `.IsHostInterface())` in this place.

We have two kinds of UDP sockets:
1. UDP sockets not using Platform UDP (e.g. MLE, TMF)
  - For these UDP sockets, IsHostInterface() is generally set to false for sending and received UDP messages. 
1. UDP sockets using Platform UDP
  - For these UDP sockets, IsHostInterface() tells where should the message be sent and from which host interface the received message is from. For example, `IsHostInterface` is set to false if the receiving host interface is `wpan0`, and is set to true otherwise.

simonlingoogle:
@wgtdkp 
Do you think this PR is deprecated by https://github.com/openthread/openthread/pull/6111?

wgtdkp:
> @wgtdkp
> Do you think this PR is deprecated by #6111?

Yes, I believe this PR is trying to fix the same issue. Will close it.

PR #6064, opened 2021-01-12
Thread 1.2 Harness Readiness

suveshpratapa:
Prepare the public openthread github repo for Thread 1.2 harness readiness, borrowing from the repo which has so far produced release candidates for the 1.2 harness.


simonlingoogle (commented):
@suveshpratapa 

You may have found that there are tons of duplicated code in thci_1_2/OpenThread and https://github.com/openthread/openthread/blob/master/tools/harness-thci/OpenThread.py, which is the OpenThread THCI for TestHarness 1.1.
I think it's better that we have only one uniform implementation of OpenThread THCI for both TestHarness 1.1 and TestHarness 1.2. Otherwise we would end up maintaining two versions of THCIs with significantly more long-term efforts.

1.2 THCI generally has more public APIs than 1.1 THCI. So we should add these missing methods to the existing OpenThread.py so that it works for both TestHarness 1.1 and TestHarness 1.2 .

Another major benefit of enhancing existing OpenThread.py is that it would automatically support OTBR (RCP+otbr-agent mode) since OTBR and SOC device share the same THCI implementation of OpenThread.py. 

Thoughts? @librasungirl @Irving-cl @jwhui

librasungirl:
@suveshpratapa Thanks for your efforts!

Agree with @simonlingoogle  that when open sourcing, we should integrate the additions for Thread 1.2 Test Harness into existing THCI, keep only one THCI to control both Thread 1.1 and 1.2 Devices which naturally work for both CLI device and OTBR now, and minimize future maintenance efforts.

-----
One note is that there are some cli command changes when moving to public repo, e.g. `linkprobing`(private) ->`linkmetrics`(public),  so the 1.2 THCI in private repo may not work for 1.2 on public master branch now.
Please refer to https://github.com/openthread/openthread/blob/master/src/cli/README.md for new or updated cli commands for 1.2 features, like `bbr`, `csl`, `linkmetrics`, `dua`, `mlr`, `fake`, `prefix` etc
Please help to supplement if any more @simonlingoogle @Irving-cl 

Fyi, #5835 has already added some THCI APIs for 1.2 low power part.

PR #6065, opened 2021-01-13
[doc] amend DUA and MLR fast poll comments

librasungirl:


simonlingoogle (approved):
LGTM :+1: 

thanks

src/core/thread/dua_manager.cpp
```diff
@@ -508,7 +508,8 @@ void DuaManager::PerformNextRegistration(void)
     mIsDuaPending   = true;
     mRegisteringDua = dua;
 
-    // TODO: (DUA) need update when CSL is enabled.
+    // Regardless of csl enabled or not, sleepy end device should always rely on
+    // fast data poll to fetch the response timely when the parent is of version 1.1.
```
jwhui, line 512: This comment doesn't seem clear to me. CSL cannot be used when parent "is of version 1.1". Can we make this more clear?

librasungirl: Updated. Please take another review.

jwhui: Much more clear. Thanks!

jwhui (approved):

PR #6068, opened 2021-01-14
[message] new helper methods to compare bytes in a message

abtink:
This commit adds new helper `Compare` methods in `Message` class to
compare the bytes from the message (at a given offset) with a byte
array or bytes from a (template) object or bytes read from another
message.

This commit also updates the message unit test to cover the behavior
of the newly added methods.

------

Background on this PR: The new methods would be helpful in parsing messages
(mainly DNS/SRP messages, e.g. matching names and/or resource records, etc).

### DIFF
--- a/src/core/common/message.cpp
+++ b/src/core/common/message.cpp
@@ -525,6 +525,48 @@ otError Message::Read(uint16_t aOffset, void *aBuf, uint16_t aLength) const
     return (ReadBytes(aOffset, aBuf, aLength) == aLength) ? OT_ERROR_NONE : OT_ERROR_PARSE;
 }
 
+bool Message::CompareBytes(uint16_t aOffset, const void *aBuf, uint16_t aLength) const
+{
+    uint16_t       bytesToCompare = aLength;
+    const uint8_t *bufPtr         = reinterpret_cast<const uint8_t *>(aBuf);
+    Chunk          chunk;
+
+    GetFirstChunk(aOffset, aLength, chunk);
+
+    while (chunk.GetLength() > 0)
+    {
+        VerifyOrExit(memcmp(bufPtr, chunk.GetData(), chunk.GetLength()) == 0);
+        bufPtr += chunk.GetLength();
+        bytesToCompare -= chunk.GetLength();
+        GetNextChunk(aLength, chunk);
+    }
+
+exit:
+    return (bytesToCompare == 0);
+}
+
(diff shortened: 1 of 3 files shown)


Irving-cl (approved):
LGTM :+1:

src/core/common/message.cpp
```diff
@@ -525,6 +525,48 @@ otError Message::Read(uint16_t aOffset, void *aBuf, uint16_t aLength) const
     return (ReadBytes(aOffset, aBuf, aLength) == aLength) ? OT_ERROR_NONE : OT_ERROR_PARSE;
 }
 
+bool Message::CompareBytes(uint16_t aOffset, const void *aBuf, uint16_t aLength) const
```
simonlingoogle, line 528:
```suggestion
bool Message::CompareBytes(uint16_t aOffset, const uint8_t *aBuf, uint16_t aLength) const
```

since the method name is `CompareBytes`.

simonlingoogle (approved):
LGTM :+1: 
with a nit

src/core/common/message.cpp
abtink: We intentionally use `void *` to make it easier for callers of these methods to be able to pass any pointer (instead of caller having to cast it, there is a common casting in the implementation). 

This is common pattern, we already use it in many places: for example, `ReadBytes()`, `AppendBytes()`, `WriteBytes()` (in `Message`) and `AppendTlv()`, `ReadTlv()` ... (in `Tlv`) or in `Sha256::Update()`, ...

src/core/common/message.hpp
```diff
@@ -710,6 +710,60 @@ class Message : public Buffer
...
+    /**
+     * This method compares the bytes in the message at a given offset with an object.
+     *
+     * The bytes in the message are compared with the bytes in @p aObject. If there are fewer bytes available in the
+     * message than the requested object size, it is treated as failed comparison (returns FALSE).
+     *
+     * @tparam     ObjectType   The object type to compare with the bytes in message.
+     *
+     * @param[in]  aOffset      Byte offset within the message to read from for the comparison.
+     * @param[out] aObject      A reference to the object to compare with the message bytes.
```
jwhui, line 754:
```suggestion
     * @param[in]  aObject      A reference to the object to compare with the message bytes.
```

jwhui (approved):

PR #6071, opened 2021-01-14
[random-manager] fail if unable to seed the CTR DRBG

### LINKED ISSUE #6014: [random manager] should we check the return value of mbedtls_ctr_drbg_seed?
Related to #6013, I experienced a hard-to-debug issue, which was masked by the fact that `RandomManager::CryptoCtrDrbg::Init` doesn't check the return value of `mbedtls_ctr_drbg_seed`, leaving the user unaware that the random hardware generator failed.

https://github.com/openthread/openthread/blob/7fe84c59cdf6ce1070ac2d877d826779a04df140/src/core/common/random_manager.cpp#L191-L195

Is this intended? Should we print an error message? Should we assert?

Thanks.

dismirlian:
Resolves #6014.

### DIFF
--- a/src/core/common/random_manager.cpp
+++ b/src/core/common/random_manager.cpp
@@ -200,7 +200,7 @@ void RandomManager::CryptoCtrDrbg::Init(void)
 
     if (rval != 0)
     {
-        otLogCritMbedTls("Can't seed the CTR DRBG");
+        otLogCritMbedTls("Failed to seed the CTR DRBG");
     }
 
     OT_ASSERT(rval == 0);
--- a/src/core/common/random_manager.cpp
+++ b/src/core/common/random_manager.cpp
@@ -41,6 +41,7 @@
 
 #include "common/code_utils.hpp"
 #include "common/debug.hpp"
+#include "common/logging.hpp"
 #include "common/random.hpp"
 #include "crypto/mbedtls.hpp"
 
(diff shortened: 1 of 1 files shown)


src/core/common/random_manager.cpp
```diff
@@ -190,8 +191,19 @@ int RandomManager::Entropy::HandleMbedtlsEntropyPoll(void *         aData,
...
+
     mbedtls_ctr_drbg_init(&mCtrDrbg);
-    mbedtls_ctr_drbg_seed(&mCtrDrbg, mbedtls_entropy_func, RandomManager::GetMbedTlsEntropyContext(), nullptr, 0);
+
+    rval =
+        mbedtls_ctr_drbg_seed(&mCtrDrbg, mbedtls_entropy_func, RandomManager::GetMbedTlsEntropyContext(), nullptr, 0);
+
+    if (rval != 0)
+    {
+        otLogCritMbedTls("Can't seed the CTR DRBG");
```
jwhui, line 203: nit:
```suggestion
        otLogCritMbedTls("Failed to seed the CTR DRBG");
```
I think "fail" is a more common search word when analyzing logs.

jwhui (commented):
Thanks! Just one minor nit.

src/core/common/random_manager.cpp
dismirlian: I agree.

Irving-cl (approved):
LGTM

jwhui (approved):

PR #6081, opened 2021-01-17
[bootstrap] fix gcc-arm-none-eabi install

### LINKED ISSUE #6080: Setting up Docker build environent, issue with ./script/bootstrap
When following the build guide at https://openthread.io/guides/build, which clones the current master branch to a Docker image. At the step './script/bootstrap' where the tools are installed there seems to be a problem with the compiler install.

In openthread/script/bootstrap starting at line 70 where the compiler is downloaded to /tmp. The archive is extracted to /tmp and then the archive is deleted. The script then tries to make a link `sudo ln -s /opt/gcc-arm-none-eabi-9-2020-q2-update/bin/* /usr/local/bin/` . But /opt is empty because the compiler is in /tmp. Should the compiler not be extracted to /opt or the link made from /tmp?

Git commit id  8625603

leonsteenkamp:
Added fix to extract archive to /opt rather than current directory which is /tmp

Fixes #6080

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -68,7 +68,7 @@ install_packages_apt()
         sudo apt-get --no-install-recommends install -y ca-certificates wget
         (cd /tmp \
             && wget -c https://developer.arm.com/-/media/Files/downloads/gnu-rm/9-2020q2/gcc-arm-none-eabi-9-2020-q2-update-x86_64-linux.tar.bz2 \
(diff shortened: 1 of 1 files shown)


leonsteenkamp:
@googlebot I signed it!

simonlingoogle (approved):
LGTM 👍

jwhui (approved):
Thanks! 👍🏼

PR #6084, opened 2021-01-18
[logging] refine FULL_LOGS option

simonlingoogle:
This commits refine the `FULL_LOGS` option (`OT_FULL_LOGS` for cmake):

- remove configuration of log regions since they should be 1 by default
- set `PLATFORM` log region to ON by default


src/core/config/logging.h
```diff
@@ -220,7 +235,7 @@
  *
  */
 #ifndef OPENTHREAD_CONFIG_LOG_PLATFORM
-#define OPENTHREAD_CONFIG_LOG_PLATFORM 0
+#define OPENTHREAD_CONFIG_LOG_PLATFORM OPENTHREAD_CONFIG_FULL_LOGS_ENABLE
```
Irving-cl, line 238: What's the reason to use `OPENTHREAD_CONFIG_FULL_LOGS_ENABLE` for `OPENTHREAD_CONFIG_LOG_PLATFORM` as default value while `1` is used as default value for all other LOG OPTIONS?

simonlingoogle: Because `OPENTHREAD_CONFIG_LOG_PLATFORM` is 0 by default and all other log options is 1 by default. 

This change is to keep this original behavior unchanged. But I am not aware of the reason why `PLATFORM` log is disabled by default.

jwhui: The LOG_PLATFORM config was added as part of #1057 and #1090 - likely because this log region typically involves code outside core. That said, I'm fine with changing LOG_PLATFORM default to 1 to be consistent with other log regions.

Irving-cl: I found that, in `src/core/config/logging.h`, most log options are by default set to `1`. It seems that these options wouldn't be affected by `OPENTRHEAD_CONFIG_FULL_LOGS_ENABLE`. (If we want to do this, we need to define every log option as `OPENTRHEAD_CONFIG_FULL_LOGS_ENABLE` just like `OPENTHREAD_CONFIG_LOG_PLATFORM`.

As I see, `OPENTRHEAD_CONFIG_FULL_LOGS_ENABLE` only sets the log level. So I think we can just set log level to debug in build system (looks like the new macro is not quite necessary) when `FULL_LOGS` option is set. Thoughts?

simonlingoogle: Turning on CORE logging by default make sense to me.

abtink (approved):
LGTM 👍

src/core/config/logging.h
simonlingoogle: Did following change:

- Set CORE logging to 1 by default (same as all other logging regions)
- When `OT_FULL_LOGS` options is ON, set UP multiple logging configurations:
  - OPENTHREAD_CONFIG_LOG_LEVEL=OT_LOG_LEVEL_DEBG
  - OPENTHREAD_CONFIG_LOG_PREPEND_LEVEL=1
  - OPENTHREAD_CONFIG_LOG_PREPEND_REGION=1

simonlingoogle: Please take another look. @Irving-cl @jwhui

Irving-cl (approved):
LGTM :+1:

wgtdkp (approved):
LGTM 👍

jwhui (approved):

Issue #5324, opened 2020-08-01
dua_manager.cpp bad ifdefs around prefix and exit


riedlse:
**Describe the bug** A clear and concise description of what the bug is.

Compile error with the following command line

**To Reproduce** Information to reproduce the behavior, including:

1. Git commit id

5276 (I think)

2. IEEE 802.15.4 hardware platform

nrf52840dk

3. Build steps

make -f examples/Makefile-nrf52840 clean
make -f examples/Makefile-nrf52840 COMMISSIONER=1 JOINER=1 COAP=1 DNS_CLIENT=1 MTD_NETDIAG=1 BORDER_ROUTER=1 MAC_FILTER=1 UDP_PROXY=1 BORDER_AGENT=1 SNTP_CLIENT=1 THREAD_VERSION=1.2 TIME_SYNC=1

4. Network topology

**Expected behavior** A clear and concise description of what you expected to happen.

Clean compile

**Console/log output** If applicable, add console/log output to help explain your problem.

  CXX      thread/libopenthread_ftd_a-network_data_leader_ftd.o
/Users/riedlse/Documents/GitHub/openthread/examples/../src/core/thread/dua_manager.cpp: In member function 'void ot::DuaManager::HandleDomainPrefixUpdate(ot::BackboneRouter::Leader::DomainPrefixState)':
/Users/riedlse/Documents/GitHub/openthread/examples/../src/core/thread/dua_manager.cpp:92:24: error: unused variable 'prefix' [-Werror=unused-variable]
   92 |     const Ip6::Prefix *prefix = nullptr;
      |                        ^~~~~~
/Users/riedlse/Documents/GitHub/openthread/examples/../src/core/thread/dua_manager.cpp:150:1: error: label 'exit' defined but not used [-Werror=unused-label]
  150 | exit:
      | ^~~~
  CXX      thread/libopenthread_ftd_a-network_data_local.o

**Additional context** Add any other context about the problem here.

I think the fix is to move #endif from 148 to 151 and add 
#if OPENTHREAD_CONFIG_DUA_ENABLE / #endif around line 92, but not sure what the code is doing, so I don't want to commit.

librasungirl:
Hi @riedlse, thanks for reporting this issue. Your analysis and suggestion are correct.
Submitted #5325 to fix the build issue, please have a try.

PR #5276, opened 2020-07-22
[dua] dua registration and re-registration

librasungirl:
This is one of the PRs for DUA feature, focusing on the registration and re-registration functionality (self and on behalf of children).

This PR implements
```
    [dua] dua registration and re-registration
    
    [Y] registration workflow framework
    [Y] registration and reregistration
    [Y] registration and reregistration on behalf of children
    [Y] add pseudo respone status for test/certification purpose
    [Y] add unit test
```
As discussed, 1.2 FTD should by default send DUA.req for its children even it doesn't enable DUA feature itself. 
Will supplement later or submit a separate PR to test this.

PR #6085, opened 2021-01-18
[scripts] add option to not build bbr when building 1.2

Irving-cl:
Currently when we build 1.2 binaries using `script/test`, we build it twice (one only with Low Power features, one only with Backbone Router feature).
As I remember,we separate it because when we implement it, we do the work individually. I think now it's not necessary to separate it anymore. We can always build one version and test all 1.2 features with it.

### DIFF
--- a/.github/workflows/simulation-1.2.yml
+++ b/.github/workflows/simulation-1.2.yml
@@ -68,7 +68,6 @@ jobs:
         python3 -m pip install -r tests/scripts/thread-cert/requirements.txt
     - name: Build
       run: |
-        ./bootstrap
         ./script/test build
     - name: Run
       run: |
@@ -111,6 +110,7 @@ jobs:
       THREAD_VERSION: 1.2
       MAC_FILTER: 1
       INTER_OP: 1
+      INTER_OP_BBR: 0
     steps:
     - uses: actions/checkout@v2
       with:
--- a/script/test
+++ b/script/test
@@ -45,6 +45,7 @@ readonly THREAD_VERSION="${THREAD_VERSION:-1.1}"
 readonly INTER_OP="${INTER_OP:-0}"
 readonly VERBOSE="${VERBOSE:-0}"
 readonly BORDER_ROUTING="${BORDER_ROUTING:-1}"
+readonly INTER_OP_BBR="${INTER_OP_BBR:-1}"
 
 readonly OT_COREDUMP_DIR="${PWD}/ot-core-dump"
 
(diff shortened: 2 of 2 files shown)


librasungirl:
> Currently when we build 1.2 binaries using `script/test`, we build it twice (one only with Low Power features, one only with Backbone Router feature).
> As I remember,we separate it because when we implement it, we do the work individually. I think now it's not necessary to separate it anymore. We can always build one version and test all 1.2 features with it.

@Irving-cl I think the intention of building 1.2 features twice is to get the 1.2 build with and without `BACKBONE_ROUTER` feature respectively - one build with MLR/DUA/LP feature, one build with extra `BACKBONE_ROUTER` feature (MLR/DUA/LP + BACKBONE_ROUTER), with the aims 1) verify build with and without `BACKBONE_ROUTER`; 2) verify functionality with and without `BACKBONE_ROUTER` feature (e.g. verify BBR Service registration, BBR state transition in core for `BACKBONE_ROUTER` enabled build, and verify device of build without `BACKBONE_ROUTER` could recognize the BBR service and send DUA.req/MLR.req to PBBR).

_Fyi, it was proposed (not done yet) to introduce one separate build for 1.2 FTD with MLR/DUA disabled and verify it could send DUA.req/MLR.req on behalf of its children(https://github.com/openthread/openthread/pull/5276#issue-455213582) which could help to avoid issues like build with different feature flags combination https://github.com/openthread/openthread/issues/5324_

I think it would be better to keep it as is. Thoughts?

Irving-cl:
@librasungirl 

Sounds good to me.

What about we add an option in `script/test` to cancel the build of `bbr` when it's not required? When we build 1.2 programs using `script/test`, bbr would always be built even when it unnecessary, which takes twice the time. We can add the option when we just want to build 1.2 without bbr. Thoughts?

librasungirl (approved):

.github/workflows/simulation-1.2.yml
```diff
@@ -111,6 +110,7 @@ jobs:
       THREAD_VERSION: 1.2
       MAC_FILTER: 1
       INTER_OP: 1
+      BBR: 0
```
simonlingoogle, line 113:
```suggestion
      INTER_OP_BBR: 0
```
Let's be clear is for turning on BBR inter op?

Irving-cl: updated.

simonlingoogle (approved):
LGTM :+1:

script/test
```diff
@@ -45,6 +45,7 @@ readonly THREAD_VERSION="${THREAD_VERSION:-1.1}"
 readonly INTER_OP="${INTER_OP:-0}"
 readonly VERBOSE="${VERBOSE:-0}"
 readonly BORDER_ROUTING="${BORDER_ROUTING:-1}"
+readonly INTER_OP_BBR="${INTER_OP_BBR:-1}"
```
jwhui, line 48: Add usage documentation for `INTER_OP_BBR`?

Irving-cl: Updated.

jwhui (approved):

PR #6087, opened 2021-01-18
[dns-header] fix 'ParseName()' when header offset is non-zero

abtink:
This commit fixes an issue with `Dns::Name::ParseName()`. This method
intentionally does not get the header offset (which is used for
parsing compressed DNS names) since its implementation does not need
to read the entire name and can stop as soon as we reach the end of
the current segment of the name. This commit changes `GetNextLabel()`
method adding a new input parameter `aStopOnEndOffsetSet` to stop as
soon as `mNameEndOffset` is found and set on the `Iterator`. This new
parameter is then used by `ParseName()` ensuring that we stop
immediately when we find the end offset and do not continue reading a
next label (jumping to an incorrect offset and potentially reading a
bad label length value and reporting it as a parse error).

The unit test `test_dns` is also updated to catch and cover this
situation by populating the guard region in the start of the message
with `0x80 + index` values which are invalid label length values
(length value's first two bits must be 00 or 11).

### DIFF
--- a/src/core/net/dns_headers.cpp
+++ b/src/core/net/dns_headers.cpp
@@ -205,7 +205,7 @@ otError Name::ParseName(const Message &aMessage, uint16_t &aOffset)
 
     while (true)
     {
-        error = iterator.GetNextLabel();
+        error = iterator.GetNextLabel(/* aStopOnEndOffsetSet */ true);
 
         VerifyOrExit((error == OT_ERROR_NONE) || (error == OT_ERROR_NOT_FOUND));
 
@@ -293,7 +293,7 @@ exit:
     return error;
 }
 
-otError Name::LabelIterator::GetNextLabel(void)
+otError Name::LabelIterator::GetNextLabel(bool aStopOnEndOffsetSet)
 {
     otError error;
 
@@ -338,6 +338,7 @@ otError Name::LabelIterator::GetNextLabel(void)
             if (!IsEndOffsetSet())
             {
                 mNameEndOffset = mNextLabelOffset + sizeof(uint16_t);
+                VerifyOrExit(!aStopOnEndOffsetSet);
             }
 
             mNextLabelOffset = mHeaderOffset + (HostSwap16(pointerValue) & kPointerLabelOffsetMask);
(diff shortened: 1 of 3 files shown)


src/core/net/dns_headers.cpp
```diff
@@ -293,7 +293,7 @@ otError Name::ReadName(const Message &aMessage,
     return error;
 }
 
-otError Name::LabelIterator::GetNextLabel(void)
+otError Name::LabelIterator::GetNextLabel(bool aStopOnEndOffsetSet)
```
simonlingoogle, line 296: This `aStopOnEndOffsetSet` looks a little bit strange to me. 
I think the better solution might be always specifying header offset when creating a `LabelIterator`. 

Another thought is that I think **always making sure that header offset is at zero** can be a good practice for DNS/SRP server/client logic. For example, we won't need to specify `aHeaderOffset` for some DNS utility methods, which would reduce code size. It's actually quite simple to guarantee that the DNS header is at offset 0:
- When we receive a UDP DNS message, we remove the DNS header.
- When we create a DNS message to send, we make sure DNS header is at the beginning.

Thoughts? @abtink @wgtdkp

abtink: My preference for now is to merge this fix to unblock the rest of the PRs. Changing this now with all pending and upcoming PRs would be rather messy.

For short-term (fix the behavior now):
- I don't favor having to provide `aHeaderOffset` to `ParseName()` (since we don't really need)
- Another approach: If you recall, in the initial version of `GetNextLabel()` we had a model to return `OT_ERROR_PENDING` when we got to "pointer label" to allow caller to decide to stop. We then removed this to optimize/simplify the code. We can use this instead of this PR
- I prefer the fix in this PR since it is simpler (and more clear).

For longer-term:
- I agree that requiring header offset to be at zero would simplify the use, but I don't like the limitation it does add.
- Another solution (I had this in mind for later enhancement) is to use the offset in `Message` itself to indicate start of DNS header (this way we keep it flexible while simplifying the APIs).

jwhui: > Another solution (I had this in mind for later enhancement) is to use the offset in Message itself to indicate start of DNS header (this way we keep it flexible while simplifying the APIs).

Having the message object carry its own offset value was primarily intended to serve use cases like this.

abtink:
Closing this in favor of https://github.com/openthread/openthread/pull/6101.

PR #6074, opened 2021-01-15
[srp] initial server and client implementation

wgtdkp:
This PR includes 5 commits:
- [net] SRP client implementation (#6038) 
- [cli] add CLI support for SRP client (#6038) 
- [net] SRP server implementation (#5986) 
- [tests] add CI tests for SRP server & client (https://github.com/openthread/openthread/pull/6074/commits/3365a94acd93e0810bff001ebee70480f698c8e5)
- [doc] add SRP server & client example (https://github.com/openthread/openthread/pull/6074/commits/fce3fed31dacebb0831db9d82585221826d734b7)

The first three commits are cherry-picked from #6038 and #5986, Review of those changes should go to their original PRs. Updates to those commits should be synced to here. **We should only review the last two commits which add CI tests & SRP examples** and when the three PRs are ready to be merged, we merges only this PR without squashing the commits. Thoughts?

PR #6088, opened 2021-01-19
[dns] fix dns failed to append names of length 255

simonlingoogle:


wgtdkp (approved):
LGTM :+1:

src/core/net/dns_headers.cpp
```diff
@@ -128,7 +128,7 @@ otError Name::AppendMultipleLabels(const char *aLabels, Message &aMessage)
 
     do
     {
-        VerifyOrExit(index < kMaxLength, error = OT_ERROR_INVALID_ARGS);
+        VerifyOrExit(index <= kMaxLength, error = OT_ERROR_INVALID_ARGS);
```
abtink, line 131: I think the condition here `index < kMaxLength` is correct (in the sense that we want `index < 255` to be checked here and if `index` gets to `255` then the name should be considered as invalid).

My understanding is that the entire encoded name should be `255` bytes of less. Note that the name should always end with a "terminator" or a zero length byte. Here in `AppendMultipleLabels()` we should not go to 255 to leave room for the "terminator" zero byte (which is added by `AppendName()`).

That said, it may be good to clarify the definition of `kMaxLength` itself and what it represents.
- If we consider a string representation of a DNS name (i.e. a string of dot `'.'` seperated sequence of labels), then the string representation's length can be at most `254` chars (or `253` char if the string omits the the ending `'.'` , i.e., using `example.com` instead o`example.com.` ).
   -  I am using "string length" definition here that does not include the `'\0'` null char at the end of string which is the commonly used notion.
   - So if the string representation has a length of `255` then it is invalid.
- Here is reason for `254` length:
  - The encoded name format is sequence of `"len (byte)"` followed by `len` chars ending with `"zero len (byte)"` and is limited to 255 bytes total.
  - When we represent it as string we use dot `'.'` for all `"len (byte)"` except for the very first one, so we end up with `254` chars as max length.
  - Note that this assumes there would no escaping of chars in the name (which is possible/recommended in case of "service instance name" (which can contain more characters than other labels).

Here is my suggestion:
- We define a private `enum` value of `kMaxEncodedSize = 255` to represent max encoded size and and use in the `AppendMultipleLabels()` to make this clear.
- We can also clarify/update the definition of `Name::kMaxLength` that it represent the maximum length of string representation of a DNS name 
   - I actually have such change  in the DNS-SD branch that I am working on. In that we define `kMaxSize = 255` (size instead of length) to indicate the "size" of char array to pass to read a name (which is again `255` to include the null char at the end of string).
  

(see related discussion here [stackoverflow.com/what-is-the-maximum-length-of-a-dns-name](https://stackoverflow.com/questions/32290167/what-is-the-maximum-length-of-a-dns-name)).

simonlingoogle: Yes, 255 is the max encoded size and the max length of a domain name string should be 254 if it ends with a dot, and 253 if it does not end with a dot. 

Yet, I think the original checking condition `index < kMaxLength` is not strictly correct. It might treat some invalid string as valid. 

I have refined the code accordingly so that it really check the encoded size against max length 255. 
Please have a look.

src/core/net/dns_headers.cpp
```diff
@@ -149,6 +147,7 @@ otError Name::AppendMultipleLabels(const char *aLabels, Message &aMessage)
                 ExitNow();
             }
 
+            VerifyOrExit(index + 1 < kMaxEncodedLength, error = OT_ERROR_INVALID_ARGS);
```
simonlingoogle, line 150: Explain: `index + 1` should always equals to the actual encoded name length after the next statement (i.e. AppendLabel) is executed. So this condition checks the exact condition we are looking for: `encoded length < 255`.

abtink (approved):
👍

src/core/net/dns_headers.hpp
```diff
@@ -486,8 +486,9 @@ class Name
 public:
     enum : uint8_t
     {
-        kMaxLabelLength = 63,  ///< Max number of characters in a label.
-        kMaxLength      = 255, ///< Max number of characters in a name.
+        kMaxLabelLength   = 63,  ///< Max number of characters in a label.
+        kMaxLength        = 254, ///< Max number of characters in a name.
+        kMaxEncodedLength = 255, ///< Max length of a encoded name.
```
librasungirl, line 491:
```suggestion
        kMaxEncodedLength = 255, ///< Max length of an encoded name.
```

librasungirl (approved):

simonlingoogle:
This PR is fine to merge now. However, I labeled it as **WIP** so that it's to be merged after #6074 to reduce the rebasing burden of #6074.  

@jwhui @wgtdkp @abtink @librasungirl

wgtdkp:
> This PR is fine to merge now. However, I labeled it as **WIP** so that it's to be merged after #6074 to reduce the rebasing burden of #6074.
> 
> @jwhui @wgtdkp @abtink @librasungirl

SGTM :+1:  Thanks!

jwhui (approved):

PR #6089, opened 2021-01-19
[posix] bind infra IF socket to specific interface

wgtdkp:
This is for ensuring sending ICMP6 packets on only the specified infrastructure interface.

### DIFF
--- a/src/posix/platform/infra_if.cpp
+++ b/src/posix/platform/infra_if.cpp
@@ -250,6 +250,13 @@ void platformInfraIfInit(otInstance *aInstance, const char *aIfName)
         DieNow(OT_EXIT_ERROR_ERRNO);
     }
 
+#ifdef __linux__
+    rval = setsockopt(sock, SOL_SOCKET, SO_BINDTODEVICE, sInfraIfName, strlen(sInfraIfName));
+#else  // __NetBSD__ || __FreeBSD__ || __APPLE__
+    rval = setsockopt(sock, IPPROTO_IP, IP_BOUND_IF, &sInfraIfIndex, sizeof(sInfraIfIndex));
+#endif // __linux__
+    VerifyOrDie(rval == 0, OT_EXIT_ERROR_ERRNO);
+
     sInfraIfIcmp6Socket = sock;
     InitLinkLocalAddress();
 }


src/posix/platform/infra_if.cpp
```diff
@@ -250,6 +250,9 @@ void platformInfraIfInit(otInstance *aInstance, const char *aIfName)
         DieNow(OT_EXIT_ERROR_ERRNO);
     }
 
+    rval = setsockopt(sock, SOL_SOCKET, SO_BINDTODEVICE, sInfraIfName, strlen(sInfraIfName));
+    verifyOrDie(rval == 0, OT_EXIT_ERROR_ERRNO);
```
simonlingoogle, line 254: follow the same code pattern as above?

```suggestion
    rval = setsockopt(sock, SOL_SOCKET, SO_BINDTODEVICE, sInfraIfName, strlen(sInfraIfName));
    if (rval < 0) 
    {
        otLogCritPlat("Can't set SO_BINDTODEVICE: %s", strerror(errno));
        DieNow(OT_EXIT_ERROR_ERRNO);
    }
```

src/posix/platform/infra_if.cpp
```diff
@@ -250,6 +250,9 @@ void platformInfraIfInit(otInstance *aInstance, const char *aIfName)
         DieNow(OT_EXIT_ERROR_ERRNO);
     }
 
+    rval = setsockopt(sock, SOL_SOCKET, SO_BINDTODEVICE, sInfraIfName, strlen(sInfraIfName));
```
simonlingoogle, line 253: also handle the case `__NetBSD__ || __FreeBSD__ || __APPLE__` ?

wgtdkp: Current infra IF implementation generally supports only linux. I think it is better to leave it in a future to add support for BSD and APPLE platforms.

src/posix/platform/infra_if.cpp
wgtdkp: We are simplifying the infra IF logging: https://github.com/openthread/openthread/pull/6090

simonlingoogle (approved):
LGTM :+1:

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #5935, opened 2020-12-11
[posix] add error logs in trel platform implementation under posix

abtink:
This commit adds explicit error logs messages for the more common
error situation in trel platform implementation under posix, e.g., if
the trel netif cannot be found, or a unicast address cannot be added
to the netif. The new logs provide is included to provide more details
in addition to log message by `VerifyOrDie()` and/or `DieNow()`

PR #6090, opened 2021-01-19
[posix] logging enhancements to infra if

wgtdkp:
This PR includes two small enhancements to the posix infra if module:
1. simplify logging of socket errors since the `VerifyOrDie` macro already prints the line number and `strerror()`.
2. suppress "drop ICMPv6 message" debug logs when there is no input data to be read on the socket.

### DIFF
--- a/src/posix/platform/infra_if.cpp
+++ b/src/posix/platform/infra_if.cpp
@@ -135,11 +135,7 @@ static void InitLinkLocalAddress(void)
 {
     struct ifaddrs *ifAddrs = nullptr;
 
-    if (getifaddrs(&ifAddrs) < 0)
-    {
-        otLogCritPlat("failed to get netif addresses: %s", strerror(errno));
-        DieNow(OT_EXIT_ERROR_ERRNO);
-    }
+    VerifyOrDie(getifaddrs(&ifAddrs) != -1, OT_EXIT_ERROR_ERRNO);
 
     for (struct ifaddrs *addr = ifAddrs; addr != nullptr; addr = addr->ifa_next)
     {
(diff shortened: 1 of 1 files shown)


simonlingoogle (approved):
LGTM 👍

src/posix/platform/infra_if.cpp
```diff
@@ -182,73 +178,41 @@ void platformInfraIfInit(otInstance *aInstance, const char *aIfName)
 
     // Initializes the infra interface.
     ifIndex = if_nametoindex(aIfName);
-    if (ifIndex == 0)
-    {
-        otLogCritPlat("failed to get the index for infra interface %s: %s", aIfName, strerror(errno));
-        DieNow(OT_EXIT_ERROR_ERRNO);
-    }
+    VerifyOrDie(ifIndex != 0, OT_EXIT_ERROR_ERRNO);
```
jwhui, line 181: I think it is valuable to have more friendly error output for common user-facing issues (e.g. when specifying an interface that cannot be found). As an example, we added some more user-friendly error output in #5935.

simonlingoogle: Agree with @jwhui . 

I think it's better to allow specifying an error message in `VerifyOrDie`. 

e.x.  
```cpp
VerifyOrDie(ifIndex != 0, OT_EXIT_ERROR_ERRNO, "failed to get the index for infra interface %s: %s", aIfName, strerror(errno));
```

wgtdkp: Makes sense. reverted the change to the interface name check.

src/posix/platform/infra_if.cpp
```diff
@@ -278,7 +246,7 @@ void platformInfraIfUpdateFdSet(fd_set &aReadFdSet, int &aMaxFd)
 
 void platformInfraIfProcess(otInstance *aInstance, const fd_set &aReadFdSet)
 {
-    otError  error = OT_ERROR_DROP;
+    otError  error = OT_ERROR_NONE;
```
librasungirl, line 249: If update the initial value as in this PR,  below line# 318 (`error = OT_ERROR_NONE;` ) could be removed?
Just curious, any specific reason to change the initial value?

wgtdkp: yes, line#318 can be removed, updated.
It is for suppress "drop ICMPv6 message" debug logs when there is no input data to be read on the socket.

librasungirl (approved):

jwhui (approved):

Issue #6011, opened 2020-12-25
Possible different datasets in the same network


Moandor-y:
**Describe the bug**

If we form a network with some nodes, and change some fields in the dataset (e.g. mesh local prefix) of some nodes, while keeping master key, channel, pan id the same values, they will still form a single network, but can't talk to others because of the different network parameters.

**To Reproduce**

Setup 2 nodes. Initialize the first one.
```
> dataset init new
Done
> dataset commit active
Done
> ifconfig up
Done
> thread start
Done
> state
leader
Done
> dataset active
Active Timestamp: 1
Channel: 17
Channel Mask: 0x07fff800
Ext PAN ID: 735c87b49993c682
Mesh Local Prefix: fd2d:5f03:f5d8:c6::/64
Master Key: ea14fcfb39bda2a41e5ac21e3dcd6ac3
Network Name: OpenThread-1f3a
PAN ID: 0x1f3a
PSKc: de6bbe039c144259560c54e6c76288f5
Security Policy: 672, onrcb
Done
> dataset active -x
0e080000000000010000000300001135060004001fffe00208735c87b49993c6820708fd2d5f03f5d800c60510ea14fcfb39bda2a41e5ac21e3dcd6ac3030f4f70656e5468726561642d3166336101021f3a0410de6bbe039c144259560c54e6c76288f50c0302a0ff
Done
```

Then set the dataset in the second node to the same except mesh local prefix.
```
> dataset set active 0e080000000000010000000300001135060004001fffe00208735c87b49993c6820708fd2d5f03f5d800c60510ea14fcfb39bda2a41e5ac21e3dcd6ac3030f4f70656e5468726561642d3166336101021f3a0410de6bbe039c144259560c54e6c76288f50c0302a0ff
Done
> dataset init active
Done
> dataset meshlocalprefix fd1d:5f03:f5d8:c6::
Done
> dataset commit active
Done
> ifconfig up
Done
> thread start
Done
> state
child
Done
```

We can see they are attached, but they can't talk to each other. And they will keep their differences forever because none of their active timestamps are greater than the other.

Should we consider this an intended behavior? Or should we make changes to the dataset mechanics so that they can reach consensus eventually? @jwhui @abtink @librasungirl @bukepo @simonlingoogle 

**Additional context**

[...]

PR #6091, opened 2021-01-19
[dataset] remove 'GenerateLocal'

ColdDivinity:
This PR removes `ActiveDataset::GenerateLocal` and enforces the recommended method of forming Thread networks, as discussed in <https://github.com/openthread/openthread/issues/6011>.

This is on top of <https://github.com/openthread/openthread/pull/6119>, please just review the last commit.

Test fixing progress:
- [ ] OTNS: https://github.com/openthread/ot-ns/pull/113
- [X] expect
- [ ] toranj - Need to update `wpantund`
- [x] NCP
- [X] Cert
- [x] External commissioner
- [X] OTCI

Depends-On: openthread/ot-ns#113
Depends-On: openthread/pyspinel#110
Depends-On: openthread/ot-commissioner#173


tests/scripts/thread-cert/thread_cert.py
```diff
@@ -235,6 +219,33 @@ def _setUp(self):
         self._inspector = debug.Inspector(self)
         self._collect_test_info_after_setup()
 
+    def _set_up_active_dataset(self, node, params, is_leader):
+        dataset = {
+            'master_key': binascii.hexlify(config.DEFAULT_MASTER_KEY).decode(),
+            'panid': params['panid'],
+        }
+
+        if is_leader:
```
simonlingoogle, line 228: why do we need to know if the node is leader or not?

Can we setup the exact same active dataset on every node no matter it's a Leader node or not?

abtink:
This is a rather big change in the behavior and impacts a lot of use-cases/tests.

I understand it's desired to move in this direction but I feel it may be too big of change to remove `GenerateLocal()` fully. 
My suggestion is to keep the behavior and add a build feature config around it (`OT_CONFIG`) first.

simonlingoogle:
> This is a rather big change in the behavior and impacts a lot of use-cases/tests.
> 
> I understand it's desired to move in this direction but I feel it may be too big of change to remove `GenerateLocal()` fully.
> My suggestion is to keep the behavior and add a build feature config around it (`OT_CONFIG`) first.

I am for the idea of removing `GenerateLocal()`. However, I think we need to find a working way without adding `is_leader` to all test topologies.

tests/scripts/thread-cert/thread_cert.py
Moandor-y: Looks like it works, and is better. Updated.

Moandor-y:
> This is a rather big change in the behavior and impacts a lot of use-cases/tests.
> 
> I understand it's desired to move in this direction but I feel it may be too big of change to remove `GenerateLocal()` fully.
> My suggestion is to keep the behavior and add a build feature config around it (`OT_CONFIG`) first.

I'm OK with adding the config flag if it's too big to fully change all tests.

BTW, is `wpantund` deprecated? Can we remove toranj tests if so?

simonlingoogle:
> I'm OK with adding the config flag if it's too big to fully change all tests.

Actually I don't think it's a good idea to introduce another configuration just for this behavior change. Connectivity is so important 
 to OpenThread that we should make it work in a unified way. 

> BTW, is `wpantund` deprecated? Can we remove toranj tests if so?

I think we should continue to support wpantund to some extent since it's required by community.

abtink:
>  BTW, is wpantund deprecated? Can we remove toranj tests if so?

We **absolutely should not** remove `toranj` from OpenThread CI.
- `toranj` is actually a test-framework for OpenThread, NCP interface, and `wpantund`  (not just `wpantund`).
-  It is actually covering a lot of functionality and behaviors in OpenThread itself (a lot of edge-cases and special behaviors are only covered by `toranj` test-cases).
- It's mostly using `wpantund` as the driver/interface to be able to run/implement the tests.

So unless all the same set of tests are added/replicated in another test frame-work and added as part of CI,  it should not be removed.


Regarding `wpantund` we (from Google side) probably won't be actively adding new features/functionality in `wpantund`, but there is no reason to stop the testing of existing, functioning well-tested (shipping) code.

-------


I think we need to consider what is the problem are we trying to solve here with this PR and what is its impact and what's the impact of removing this behavior? 

- In my view, this is rather big change and therefore risky.
- So I suggest that rather than removing this fully, we first add a build config feature around it and disable it in certain cases.
- This way, it's not completely removed and can be enabled back quickly.
- Also there is no  need to change and update all the tests and code at the same time.

Moandor-y:
> I think we need to consider what is the problem are we trying to solve here with this PR and what is its impact and what's the impact of removing this behavior?
> 
> * In my view, this is rather big change and therefore risky.
> * So I suggest that rather than removing this fully, we first add a build config feature around it and disable it in certain cases.
> * This way, it's not completely removed and can be enabled back quickly.
> * Also there is no  need to change and update all the tests and code at the same time.

IMO most codes in this change are for test cases. These tests should work with both GenerateLocal/non-GenerateLocal versions and are not risky. The actual behavior change is only 2 lines (in `ActiveDataset::StartLeader` and `MleRouter::BecomeLeader`).

I think one way is to split them into 2 PRs, one to make tests compatible, the other to do the behavior change. If anything goes wrong we can revert easily. Thoughts?

simonlingoogle:
> I think one way is to split them into 2 PRs, one to make tests compatible, the other to do the behavior change. If anything goes wrong we can revert easily. Thoughts?

I think it's a good idea to do the tests change first.

Then, we can remove the GenerateLocal() and add corresponding tests to verify that it works:
- Verify that devices without active dataset can attach to an existing Thread network
- Verify that devices without active dataset can not become Leader by itself

jwhui:
Superseded by #9215

PR #5622, opened 2020-10-09
[cmake] add build options for FTD, MTD, and RCP

jwhui:

PR #6092, opened 2021-01-19
[build] only build unit tests when FTD is enabled

Irving-cl:
Following #5622, build unit tests only when FTD is enabled.

With #5622, we can build MTD only. However the unit tests would still be built in this case and it links the `ftd` lib, which causes a link error. This PR solves this problem.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -26,4 +26,6 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
+if(OT_FTD)
 add_subdirectory(unit)
+endif()


tests/CMakeLists.txt
```diff
@@ -26,4 +26,6 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
+if(OT_FTD)
```
simonlingoogle, line 29: Just curious do we only run unit tests for FTD ?

jwhui: Ideally, we should run the subset of unit tests that apply to the current build.

simonlingoogle: It seems to be a big change and maybe unnecessary to allow building unit for MTD. Only running unit tests for FTD is fine to me.

simonlingoogle (approved):
LGTM 👍

tests/CMakeLists.txt
Irving-cl: After rethinking on this, I think we can do it in this way:
* So long as `OT_FTD` is ON, we link `openthread-ftd` for the unit tests.
* Otherwise, we link `openthread-mtd` for the unit tests.

I haven't tried yet. But I think it would work. Since currently we hard code linking  'openthread-ftd' for unit tests, there shouldn't be any MTD-only dependencies in unit tests.

simonlingoogle: Maybe you are going to need to add `#if OPENTHREAD_FTD ... #endif` for test code that tests FTD only features.

Be aware that OT_FTD and OT_MTD can be both ON. In such case, which library are you going to link?
Are we going to build separate FTD and MTD binaries  for each test?

Irving-cl: > Maybe you are going to need to add #if OPENTHREAD_FTD ... #endif for test code that tests FTD only features.

Yeah, many unit tests (`test_multicast_listeners_table`, `test_ndproxy_table`...) is already doing this.

> Be aware that OT_FTD and OT_MTD can be both ON. In such case, which library are you going to link?
> Are we going to build separate FTD and MTD binaries for each test?

No, in this case we link the ftd library. And this is currently how we handle it. Currently we build FTD and MTD together and we link unit tests with ftd lib. This means that, for the unit test code, MTD is always a subset of FTD. The purpose of doing this is just to make it possible to build unit tests (though only a subset) when only MTD is enabled. 

Perhaps there would be some test code for MTD-only features in the future. At that time, we would whatever need to link two different libs for unit tests. For now we only need to link ftd lib for unit tests.

simonlingoogle: > This means that, for the unit test code, MTD is always a subset of FTD.

Agree. That being the case, just building unit tests for FTD should be sufficient for testing purpose. Thoughts?

Irving-cl: SGTM. Then let's keep it as it is.

jwhui (approved):

Issue #6093, opened 2021-01-19
bootstrap script fails in docker environment container


deimi:
**Describe the bug**
`script/bootstrap` does not fully work in a docker build environment.

**To Reproduce** 

1. `docker run -it openthread/environment bash`
2. `cd openthread`
3. `script/bootstrap`

**Console/log output**
```
...
[...]

PR #6094, opened 2021-01-19
[bootstrap] overwrite existing links

deimi:
Create links with forced overwrite flag to prevent a failing script if a link already exist.

Fixes https://github.com/openthread/openthread/issues/6093


jwhui (approved):
Thanks! 👍🏼

PR #6095, opened 2021-01-20
[sed-to-sed] support indirect transmission on mtd

Irving-cl:
This PR makes indirect transmission code also available on MTD. An abstraction of the changes:
* Change the inheritance relationship of topology classes. Add a new class `SedCapableNeighbor` and move part of data fields (for indirect transmission) from `Child` to `SedCapableNeighbor`. And let `Child` inherits `SedCapableNeighbor`.
* Add a new table `SedCapableNeighborTable` which is used to manage `SedCapableNeighbor` objects. It has different implementation on `FTD` and `MTD`. For `FTD`, this class is similar to `NeighborTable` which accesses objects actually from `ChildTable`. For `MTD`, this class maintains an array of `SedCapableNeighbor` and the `Find` methods directly accesses these objects.
* In indirect transmission modules (`IndirectSender`, `DataPollHandler`, `CslTxScheduler`, `SrcMatchController), we abstract the concept of `Child` to `SedCapableNeighbor`. All occurrences of `Child` and `ChildTable` are substituted by `SedCapableNeighbor` and `SedCapableNeighborTable`.
* Implement indirect transmission in `MeshForwarder(MTD)::SendMessage`.
* Update those macros (`OPENTHREAD_FTD`) wrapping indirect transmission related code (since they are also needed on MTD now).

Indirect transmission with MTD hasn't been tested since S2S's implementation is not included in this PR. I would post this PR first to get some general feedback for this change.


abtink:
Thanks @Irving-cl , I looked through the code and also checking the comment here:

> Add a new table `SedCapableNeighborTable` which is used to manage `SedCapableNeighbor` objects. It has different implementation on FTD and MTD. For FTD, this class is similar to NeighborTable which accesses objects actually from ChildTable. For MTD, this class maintains an array of SedCapableNeighbor and the Find methods directly accesses these objects.

Here are some thoughts/suggestion.
- I like the concept behind the new `SedCapableNeighbor` type and how it sits in the class hierarchy, `Neighbor` -> `SedCapableNeighbor` -> `Child`
  - In case of FTD `SedCapableNeighborTable` seems to be same as the `ChildTable` (uses `ChildTable` actually).
   - In case of MTD, we add something that looks like `ChildTable` again (subset of APIs that `ChildTable` provides).
- The downside that I see with adding new `SedCapableNeighbor` type  is that in a lot of places we now need to use this new type instead of `Child` type (which is big code `diff` and also makes it harder to read).
- Here is my suggestion on this:
   - How about if we continue to use `Child` type everywhere. 
   - Of course, we would need to make the distinction/definition that on MTD, a `Child` is special type of SED capable neighbor (though in the code we still use `Child` type for it and refer to it as a child).
   - Basically in MTD build, we add a new definition for `Child` type which is similar to what we now define as a `SedCapableNeighbor`.
   - And in MTD build we also include/use `ChildTable` (maybe with a subset of its methods) which then avoids the need for us to define a separate `SedCapableNeighborTable`.
   - What do you think of this approach?
  
- Another thing I noticed that this PR is adding about 5K to code size and 4K on MTD builds (NCP/FTD). I don't think it would be a good idea to add this overhead on all MTD builds (specially when the SED to SED solution is not used).
   - I suggest we add have `OT_CONFIG` flags to enable support for this on MTD build when needed.

Thoughts?

Irving-cl:
@abtink 

Thanks! Also using `Child` sounds good to me and it's simpler. My thought for using `SedCapableNeighbor` is just to clarify the 'concept' (In MTD, the neighbor is not a child technically). I think it would be ok that we still use `Child` class add document to explain these things.

I would add a flag for controlling the MTD build to enable/disable this feature and disable by default.

Irving-cl:
I opened a new PR#6120 for the new implementation.

PR #6097, opened 2021-01-20
[dns-header] helper methods to parse/find/read record(s) in a message

abtink:
This commit adds new helper methods in `ResourceRecord` class.
`ParseRecords()` parses and skips over a given number of resource
records in a message. `FindRecord()` searches in a given message to
find the first resource record matching a given record name. And
`ReadRecord<RecordType>()` tries to read a matching resource record
(of a given type) from a message. If the record type does not matches
the type, it skips over the record. This commit also adds methods in
`PtrRecord`, `SrvRecord`, `SigRecord` and `TxtRecord` to parse and
read the data fields such as PTR name, host name, signer name and
TXT data.

This commit updates the DNS unit test to add a detailed new test
constructing and parsing a DNS message (covering the behavior of
the newly added and some of the existing methods).

------

This PR contain two other commits from #6101 and #6083. 
Please review the last commit only. Thanks.

### DIFF
--- a/src/core/net/dns_headers.cpp
+++ b/src/core/net/dns_headers.cpp
@@ -531,6 +531,169 @@ bool Name::LabelIterator::CompareLabel(const LabelIterator &aOtherIterator) cons
                                  mLabelLength);
 }
 
+otError ResourceRecord::ParseRecords(const Message &aMessage, uint16_t &aOffset, uint16_t aNumRecords)
+{
+    otError error = OT_ERROR_NONE;
+
+    while (aNumRecords > 0)
+    {
+        ResourceRecord record;
+
+        SuccessOrExit(error = Name::ParseName(aMessage, aOffset));
+        SuccessOrExit(error = record.ReadFrom(aMessage, aOffset));
+        aOffset += static_cast<uint16_t>(record.GetSize());
+        aNumRecords--;
+    }
+
+exit:
+    return error;
+}
+
+otError ResourceRecord::FindRecord(const Message &aMessage, uint16_t &aOffset, uint16_t &aNumRecords, const char *aName)
+{
+    otError error;
+
+    while (aNumRecords > 0)
+    {
+        bool           matches = true;
+        ResourceRecord record;
+
+        error = Name::CompareName(aMessage, aOffset, aName);
+
+        switch (error)
+        {
+        case OT_ERROR_NONE:
+            break;
+        case OT_ERROR_NOT_FOUND:
+            matches = false;
+            break;
+        default:
+            ExitNow();
+        }
+
+        SuccessOrExit(error = record.ReadFrom(aMessage, aOffset));
+        aNumRecords--;
+        VerifyOrExit(!matches);
+        aOffset += static_cast<uint16_t>(record.GetSize());
+    }
+
+    error = OT_ERROR_NOT_FOUND;
+
+exit:
+    return error;
+}
+
+otError ResourceRecord::ReadRecord(const Message & aMessage,
+                                   uint16_t &      aOffset,
+                                   uint16_t        aType,
+                                   ResourceRecord &aRecord,
+                                   uint16_t        aMinRecordSize)
+{
+    // This static method tries to read a matching resource record of a
+    // given type and a minimum record size from a message. The `aType`
+    // value of `kTypeAny` matches any type.  If the record in the
+    // message does not match, it skips over the record. Please see
+    // `ReadRecord<RecordType>()` for more details.
+
+    otError        error;
+    ResourceRecord record;
+
+    SuccessOrExit(error = record.ReadFrom(aMessage, aOffset));
+
+    if (((aType == kTypeAny) || (record.GetType() == aType)) && (record.GetSize() >= aMinRecordSize))
+    {
+        IgnoreError(aMessage.Read(aOffset, &aRecord, aMinRecordSize));
+        aOffset += aMinRecordSize;
+    }
+    else
+    {
+        // Skip over the entire record.
+        aOffset += static_cast<uint16_t>(record.GetSize());
+        error = OT_ERROR_NOT_FOUND;
+    }
+
+exit:
+    return error;
+}
+
+otError ResourceRecord::ReadName(const Message &aMessage,
+                                 uint16_t &     aOffset,
+                                 uint16_t       aStartOffset,
+                                 char *         aNameBuffer,
+                                 uint16_t       aNameBufferSize,
+                                 bool           aSkipRecord) const
+{
+    // This protected method parses and reads a name field in a record
+    // from a message. It is intended only for sub-classes of
+    // `ResourceRecord`.
+    //
+    // On input `aOffset` gives the offset in `aMessage` to the start of
+    // name field. `aStartOffset` gives the offset to the start of the
+    // `ResourceRecord`. `aSkipRecord` indicates whether to skip over
+    // the entire resource record or just the read name. On exit, when
+    // successfully read, `aOffset` is updated to either point after the
+    // end of record or after the the name field.
+    //
+    // When read successfully, this method returns `OT_ERROR_NONE`. On a
+    // parse error (invalid format) returns `OT_ERROR_PARSE`. If the
+    // name does not fit in the given name buffer it returns
+    // `OT_ERROR_NO_BUFS`
+
+    otError error = OT_ERROR_NONE;
+
+    SuccessOrExit(error = Name::ReadName(aMessage, aOffset, aNameBuffer, aNameBufferSize));
+    VerifyOrExit(aOffset <= aStartOffset + GetSize(), error = OT_ERROR_PARSE);
+
+    VerifyOrExit(aSkipRecord);
+    aOffset = aStartOffset;
+    error   = SkipRecord(aMessage, aOffset);
+
+exit:
+    return error;
+}
+
+otError ResourceRecord::SkipRecord(const Message &aMessage, uint16_t &aOffset) const
+{
+    // This protected method parses and skips over a resource record
+    // in a message.
+    //
+    // On input `aOffset` gives the offset in `aMessage` to the start of
+    // the `ResourceRecord`. On exit, when successfully parsed, `aOffset`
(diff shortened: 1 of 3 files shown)


src/core/net/dns_headers.hpp
```diff
@@ -855,7 +933,108 @@ class ResourceRecord
...
+     *
+     * @param[in]    aMessage        The message in which to search for a matching resource record.
+     *                               `aMessage.GetOffset()` MUST point to the start of DNS header.
+     * @param[inout] aOffset         On input, the offset in @p aMessage pointing to the start of the first record.
+     *                               On exit, if a matching record is found, @p aOffset is updated to point to the byte
+     *                               after the record name.
+     *                               If a matching record could not be found, @p aOffset is updated to point to the byte
+     *                               after the last record that was checked.
+     * @param[inout] aNumRecords     On input, the maximum number of records to check (starting from @p aOffset).
+     *                               On exit and if a matching records is found, @p aNumRecords is updated to give the
```
Irving-cl, line 963:
```suggestion
     *                               On exit and if a matching record is found, @p aNumRecords is updated to give the
```

src/core/net/dns_headers.cpp
```diff
@@ -373,5 +487,257 @@ otError Name::LabelIterator::ReadLabel(char *aLabelBuffer, uint8_t &aLabelLength
...
+otError ResourceRecord::ReadName(const Message &aMessage,
+                                 uint16_t &     aOffset,
+                                 uint16_t       aStartOffset,
+                                 char *         aNameBuffer,
+                                 uint16_t       aNameBufferSize,
+                                 bool           aSkipRecord) const
+{
+    // This protected method parses and reads a name field in a record
+    // from a message. It is intended only for sub-classes of
+    // `ResourceRecod`.
```
Irving-cl, line 628:
```suggestion
    // `ResourceRecord`.
```

tests/unit/test_dns.cpp
```diff
@@ -167,12 +172,60 @@ void TestDnsName(void)
...
+        SuccessOrQuit(Dns::Name::CompareName(*message, offset, test.mExpectedReadName), "Name::CompareName() failed");
+        VerifyOrQuit(offset == len, "Name::CompareName() returned incorrect offset");
+
+        offset = 0;
+        VerifyOrQuit(Dns::Name::CompareName(*message, offset, kBadName) == OT_ERROR_NOT_FOUND,
+                     "Name::CompareName() did not fail with incorrect name");
+        VerifyOrQuit(offset == len, "Name::CompareName() returned incorrect offset");
+
+        // Remove the terminating '.' in expected name and verify
+        // that it is can still be used by `CompareName()`
```
Irving-cl, line 208:
```suggestion
        // that it can still be used by `CompareName()`
```

Irving-cl (approved):
LGTM :+1:

src/core/net/dns_headers.hpp
```diff
@@ -636,24 +634,104 @@ class Name
...
+     *
+     * @param[in]    aMessage         The message to read the label from to compare. `aMessage.GetOffset()` MUST point
+     *                                to the start of DNS header (this is used to handle compressed names).
+     * @param[inout] aOffset          On input, the offset in @p aMessage pointing to the start of the label to read.
+     *                                On exit and only when label is successfully read and does match @p aLabel,
+     *                                @p aOffset is updated to point to the start of the next label.
+     * @param[in]    aLabel           A pointer to a null terminated string containing the label to compare with.
+
+     * @retval OT_ERROR_NONE          The label from @p aMessage matches @p aLabel. @p aOffset is updated.
+     * @retval OT_ERROR_NOT_FOUND     The label from @p aMessage does not match @p aLabel (note that @p aOffset is not
```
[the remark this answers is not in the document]
abtink: This method allows the labels in an encoded name in `message` to be read and compared one by one.

One thing to have in mind (and is different with labels vs full name) is that as we read and compare the labels, the offset can jump back in the message  in the case of compressed name.

I considered different models on how to update/manage the `aOffset` here:
1. If the label does not match, we skip over the label and update the offset to end of current label. But this did not make sense to me:
   - Why would the user want to skip over just a label if it does not match. It makes more sense for user to want try to compare label with maybe something else (in this case offset should remian the same)
   - `aOffset` would not be tracking end of original name field sicne we may have jumped back in message as we read labels one by one.
 2. If the label does not match, we skip over the entire name field. This makes more sense, but implementing it complicates the method and its use:
    - Now we need the user to provide multiple offsets as paramters to this method. The offset to start of the original name field in addition to offset to current label to read (just so that we skip over the original name filed if the label does not match).
    - This makes this method more complicated to use for a situation which may not that common/useful
       - If users want this behavior they can keep track of original offset of name themselve and then use `ParseName()` to parse and skip over it.
3. I decided on this model which made most sense to me. It's simple and practical (i.e, we do not update the offset if the label does not match).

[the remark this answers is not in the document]
Irving-cl: Make sense to me. Thanks!

src/core/net/dns_headers.hpp
```diff
@@ -881,6 +1065,70 @@ class PtrRecord : public ResourceRecord
...
+     * The @p aNameBuffer can be set to `nullptr` if the caller is only interested in the first label. This method is
+     * intended for "Service Instance Name" where first label (`<Instance>` portion) can be a user-friendly string and
+     * can contain dot character.
+     *
+     * @param[in]     aMessage          The message to read from. `aMessage.GetOffset()` MUST point to the start of
+     *                                  DNS header.
+     * @param[inout]  aOffset           On input, the offset in @p aMessage to the start of PTR name field.
+     *                                  On exit, when successfully read, @p aOffset  is updated to point to the byte
+     *                                  after the entire PTR record (skipping over the record).
+     * @param[out]    aLabelBuffer      A pointer to a char array to output the first label as a null-terminated C
```
simonlingoogle, line 1112: Is the first label supposed to be the instance name?

However, I think the instance name can contain `.` character. Should we correctly handle such case?

simonlingoogle (approved):
LGTM :+1: 

with a nit

src/core/net/dns_headers.hpp
simonlingoogle: FYI, in #6103, I introduced a method to split the full names into \<Instance\>.\<Service\>.\<Protocol\>.\<Domain\>, which would allow us to better handle a full service/instance name as a null-terminated string.

abtink: > Is the first label supposed to be the instance name?
> 
> However, I think the instance name can contain . character. Should we correctly handle such case?
> 

That's exactly the reason we have this method which reads the first label separately ... 😃 
I mention this in the comment/description just few lines above this (at line 1101).

> FYI, in #6103, I introduced a method to split the full names into <Instance>.<Service>.<Protocol>.<Domain>, which would allow us to better handle a full service/instance name as a null-terminated string.

My suggestion is to read and treat the "instance name" label seperately (that's what we do in the SPR client code and what I will do in DNS-SD client code).

If we have a full name as a C string with `'.'` separating the labesl, then we need to be able to tell if a dot `.` char is part of the instance name or it is a separator. We would need use some escaping mechanism. I think reading and keeping instance name label seperately as its own C string may be simpler.

simonlingoogle: Sorry I got it wrong. 

Let's discuss the it further in DNS-SD Server PR if necessary.

abtink: No worries at all. 😃

src/core/net/dns_headers.hpp
```diff
@@ -1402,6 +1736,33 @@ class SigRecord : public ResourceRecord, public Clearable<SigRecord>
...
+     *
+     */
+    otError ReadSignerName(const Message &aMessage,
+                           uint16_t &     aOffset,
+                           char *         aNameBuffer,
+                           uint16_t       aNameBufferSize) const
+    {
+        return ResourceRecord::ReadName(aMessage, aOffset, /* aStartOffset */ aOffset - sizeof(SigRecord), aNameBuffer,
+                                        aNameBufferSize,
+                                        /* aSkipRecord */ false);
```
jwhui, line 1763: It is well-documented above, but at first glance, this seems inconsistent with the other `Read*Name()` methods.

src/core/net/dns_headers.hpp
```diff
@@ -1009,6 +1303,36 @@ class SrvRecord : public ResourceRecord
...
+     * This method also verifies that the SRV record is well-formed (e.g., the record data length `GetLength()` matches
+     * the SRV encoded name).
+     *
+     * @param[in]     aMessage          The message to read from. `aMessage.GetOffset()` MUST point to the start of
+     *                                  DNS header.
+     * @param[inout]  aOffset           On input, the offset in @p aMessage to start of target host name field.
+     *                                  On exit when successfully read, @p aOffset is updated to point to the byte
+     *                                  after the entire SRV record (skipping over the record).
+     * @param[out]    aNameBuffer       A pointer to a char array to output the read name as a null-terminated C string
+     *                                  (MUTS NOT be nullptr).
```
jwhui, line 1318:
```suggestion
     *                                  (MUST NOT be nullptr).
```

src/core/net/dns_headers.hpp
abtink: The main difference is that in the case of SIG record `ReadSignerName()`, we still have the signature field after the name, so it updates the offset to end of the name to allow caller to read the signature field.

But in the case of PTR and SRV record the name is the last field in the record (last field we know and care about), so after reading the name we update offset to skip over the entire record.

Also I am treating the DNS records, similar to how we parse TLVs, in the sense that if there are any remaining fields/bytes in the record after the fields that we know and read, we just ignore them (this is to allow future extendibility of the record definitions). So in case of PTR/SRV, we do not assume that the offset after the name would be the same as the end of the record (we check and ensure that the name is in within the record data but when we want to skip over the record, we use the record data length directly).

librasungirl (approved):

jwhui (approved):

PR #6101, opened 2021-01-20
[dns-name] require message offset to point to DNS header

abtink:
This commit contains two changes in `Dns::Name` class. When reading or
parsing names/labels from a `Message`, we require that the message
offset (i.e., `aMessage.GetOffset()`) to be set by the caller to point
to the start of the DNS header in the message. This is then used to
decode compressed DNS names. This change helps simplify the code and
removes the need to pass `aHeaderOffset` as a separate parameter. The
second change is to the `ParseName()` implementation which now always
goes through and parses the entire compressed name before returning.
This ensures that we correctly detect invalid compressed names when
parsing the name in a message.

------

This is replacing PR https://github.com/openthread/openthread/pull/6087. 
I will inlcude this commit in other DNS-SD related PRs. My suggestion is to 
merge this after SRP related PRs are merged (but before DNS-SD feature).

### DIFF
--- a/src/core/net/dns_headers.cpp
+++ b/src/core/net/dns_headers.cpp
@@ -207,12 +207,20 @@ otError Name::ParseName(const Message &aMessage, uint16_t &aOffset)
     {
         error = iterator.GetNextLabel();
 
-        VerifyOrExit((error == OT_ERROR_NONE) || (error == OT_ERROR_NOT_FOUND));
-
-        if (iterator.IsEndOffsetSet())
+        switch (error)
         {
+        case OT_ERROR_NONE:
+            break;
+
+        case OT_ERROR_NOT_FOUND:
+            // We reached the end of name successfully.
             aOffset = iterator.mNameEndOffset;
-            ExitNow(error = OT_ERROR_NONE);
+            error   = OT_ERROR_NONE;
+
+            OT_FALL_THROUGH;
+
+        default:
+            ExitNow();
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/core/net/dns_headers.cpp
```diff
@@ -207,27 +207,31 @@ otError Name::ParseName(const Message &aMessage, uint16_t &aOffset)
     {
         error = iterator.GetNextLabel();
 
-        VerifyOrExit((error == OT_ERROR_NONE) || (error == OT_ERROR_NOT_FOUND));
-
-        if (iterator.IsEndOffsetSet())
+        switch (error)
         {
+        case OT_ERROR_NONE:
+            break;
```
simonlingoogle, line 213: should we exit in the case `error == OT_ERROR_NONE` same as the original code did?

simonlingoogle: BTW, I think the method name should be renamed to `SkipName` to reflect its actual behavior.

simonlingoogle (approved):
LGTM :+1: 

Making sure the DNS header is at `GetOffset()` looks good to me, even though I think it's even better to make sure DNS header is at offset zero.

src/core/net/dns_headers.cpp
Irving-cl: I think calling `ExitNow` causes extra instruction (`jmp`) and brings one more line of code.

Irving-cl (approved):
LGTM :+1:

src/core/net/dns_headers.cpp
abtink: > should we exit in the case error == OT_ERROR_NONE same as the original code did?

The original code did not exit on `error == OT_ERROR_NONE`. And it would not be not correct behavior and not what we want.

The behavior we want is to keep reading labels till we can get to the end of name (or in original code, till we find the end offset).

The original code was doing `VerifyOrExit((error == OT_ERROR_NONE) || (error == OT_ERROR_NOT_FOUND));` which exits if error is anything other than these two values. This is equivlaent to `default` in the new `switch(aError)` statement.

> BTW, I think the method name should be renamed to `SkipName` to reflect its actual behavior.

Skipping over the read info is somthing that all similar API already do (e.g. `ReadName()` also skips over the name). I prefer `ParseName` to be clear that we are parsing the name (i.e. verifying it is correctly formatted/encoded in the message).

simonlingoogle: thanks for explanation.

librasungirl (approved):

abtink:
Marked as `"work-in-progress"` so to be rebased after SPR related PRs are merged.

abtink:
Just rebased on master (with SRP related PRs merged) and updated `srp-server` uses of `ReadName()` (now that the offset is tracked by `message` itself).

jwhui (approved):

PR #6102, opened 2021-01-21
[thread-cert] ignore CLI log lines

simonlingoogle:
This commit allows node.py to ignore log lines from node CLI output. 

Only if `_expect_command_output` is used.


wgtdkp (approved):
Thanks, it is useful when manually run the tests with FULL LOGS.

jwhui (approved):

PR #6103, opened 2021-01-21
[dns-sd] DNS-SD server implementation

simonlingoogle:
This commit implements the DNS-SD server:

- [x] Handle Standard DNS Query from clients
  - [x] Supported resource records: PTR, SRV, TXT, AAAA
- [x] Query services from SRP server
- [x] Add tests
  - [x] simulation test using DNS client (only test AAAA query)
  - [x] OTBR test using [`dig`](https://en.wikipedia.org/wiki/Dig_(command)) command to verify all resource records can be queried successfully

### DIFF
--- a/examples/common-switches.mk
+++ b/examples/common-switches.mk
@@ -51,6 +51,7 @@ DIAGNOSTIC                ?= 0
 DISABLE_DOC               ?= 0
 DISABLE_TOOLS             ?= 0
 DNS_CLIENT                ?= 0
+DNSSD_SERVER              ?= 0
 DUA                       ?= 0
 DYNAMIC_LOG_LEVEL         ?= 0
 ECDSA                     ?= 0
@@ -183,6 +184,10 @@ endif
 
 ifeq ($(DNS_CLIENT),1)
 COMMONCFLAGS                   += -DOPENTHREAD_CONFIG_DNS_CLIENT_ENABLE=1
+endif
+
+ifeq ($(DNSSD_SERVER),1)
+COMMONCFLAGS                   += -DOPENTHREAD_CONFIG_DNSSD_SERVER_ENABLE=1
 endif
 
 ifeq ($(DUA),1)
--- a/src/core/config/dnssd_server.h
+++ b/src/core/config/dnssd_server.h
@@ -1,0 +1,58 @@
+/*
+ *  Copyright (c) 2021, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes compile-time configurations for the DNS-SD Server.
+ *
+ */
+
+#ifndef CONFIG_DNSSD_SERVER_H_
+#define CONFIG_DNSSD_SERVER_H_
+
+/**
+ * @def OPENTHREAD_CONFIG_DNSSD_SERVER_ENABLE
+ *
+ * Define to 1 to enable DNS-SD Server support.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_DNSSD_SERVER_ENABLE
+#define OPENTHREAD_CONFIG_DNSSD_SERVER_ENABLE 0
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_DNSSD_SERVER_PORT
+ *
+ * Define the the DNS-SD Server port.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_DNSSD_SERVER_PORT
+#define OPENTHREAD_CONFIG_DNSSD_SERVER_PORT 53
+#endif
+
+#endif // CONFIG_DNSSD_SERVER_H_
--- a/src/core/net/dns_headers.cpp
+++ b/src/core/net/dns_headers.cpp
@@ -35,6 +35,7 @@
 
 #include "common/code_utils.hpp"
 #include "common/debug.hpp"
+#include "common/logging.hpp"
 #include "common/random.hpp"
 #include "common/string.hpp"
 
@@ -104,20 +105,25 @@ otError Name::AppendLabel(const char *aLabel, Message &aMessage)
     return AppendLabel(aLabel, static_cast<uint8_t>(StringLength(aLabel, kMaxLabelLength + 1)), aMessage);
 }
 
-otError Name::AppendLabel(const char *aLabel, uint8_t aLabelLength, Message &aMessage)
+otError Name::AppendLabel(const char *aLabel, uint8_t aLength, Message &aMessage)
 {
     otError error = OT_ERROR_NONE;
 
-    VerifyOrExit((0 < aLabelLength) && (aLabelLength <= kMaxLabelLength), error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit((0 < aLength) && (aLength <= kMaxLabelLength), error = OT_ERROR_INVALID_ARGS);
 
-    SuccessOrExit(error = aMessage.Append(aLabelLength));
-    error = aMessage.AppendBytes(aLabel, aLabelLength);
+    SuccessOrExit(error = aMessage.Append(aLength));
+    error = aMessage.AppendBytes(aLabel, aLength);
 
 exit:
     return error;
 }
 
 otError Name::AppendMultipleLabels(const char *aLabels, Message &aMessage)
+{
+    return AppendMultipleLabels(aLabels, kMaxLength, aMessage);
+}
+
+otError Name::AppendMultipleLabels(const char *aLabels, uint8_t aLength, Message &aMessage)
 {
     otError  error           = OT_ERROR_NONE;
     uint16_t index           = 0;
@@ -128,7 +134,7 @@ otError Name::AppendMultipleLabels(const char *aLabels, Message &aMessage)
 
     do
     {
-        ch = aLabels[index];
+        ch = index < aLength ? aLabels[index] : static_cast<char>(kNullChar);
 
         if ((ch == kNullChar) || (ch == kLabelSeperatorChar))
         {
--- a/src/core/net/dns_headers.hpp
+++ b/src/core/net/dns_headers.hpp
@@ -495,6 +495,11 @@ public:
         kMaxEncodedLength = 255, ///< Max length of an encoded name.
     };
 
+    enum : char
+    {
+        kLabelSeperatorChar = '.',
+    };
+
     /**
      * This enumeration represents the name type.
      *
@@ -648,6 +653,26 @@ public:
     static otError AppendLabel(const char *aLabel, Message &aMessage);
 
     /**
+     * This static method encodes and appends a single name label of specified length to a message.
+     *
+     * The @p aLabel is assumed to contain a single name label of given @p aLength.  @p aLabel must not contain
+     * '\0' characters within the length @p aLength. Unlike `AppendMultipleLabels()` which parses the label string
+     * and treats it as sequence of multiple (dot-separated) labels, this method always appends @p aLabel as a single
+     * whole label. This allows the label string to even contain dot '.' character, which, for example, is useful for
+     * "Service Instance Names" where <Instance> portion is a user-friendly name and can contain dot characters.
+     *
+     * @param[in] aLabel         The label string to append. MUST NOT be nullptr.
+     * @param[in] aLength        The length of the label to append.
+     * @param[in] aMessage       The message to append to.
+     *
+     * @retval OT_ERROR_NONE          Successfully encoded and appended the name label to @p aMessage.
+     * @retval OT_ERROR_INVALID_ARGS  @p aLabel is not valid (e.g., label length is not within valid range).
+     * @retval OT_ERROR_NO_BUFS       Insufficient available buffers to grow the message.
+     *
+     */
+    static otError AppendLabel(const char *aLabel, uint8_t aLength, Message &aMessage);
+
+    /**
      * This static method encodes and appends a sequence of name labels to a given message.
      *
      * The @p aLabels must follow  "<label1>.<label2>.<label3>", i.e., a sequence of labels separated by dot '.' char.
@@ -668,6 +693,33 @@ public:
      *
      */
     static otError AppendMultipleLabels(const char *aLabels, Message &aMessage);
+
+    /**
+     * This static method encodes and appends a sequence of name labels within the specified length to a given message.
+     * This method stops appending labels if @p aLength characters are read or '\0' is found before @p aLength
+     * characters.
+     *
+     * This method is useful for appending a number of labels of the name instead of appending all labels.
+     *
+     * The @p aLabels must follow  "<label1>.<label2>.<label3>", i.e., a sequence of labels separated by dot '.' char.
+     * E.g., "_http._tcp", "_http._tcp." (same as previous one), "host-1.test".
+     *
+     * This method validates that the @p aLabels is a valid name format, i.e., no empty label, and labels are
+     * `kMaxLabelLength` (63) characters or less.
+     *
+     * @note This method NEVER adds a label terminator (empty label) to the message, even in the case where @p aLabels
+     * ends with a dot character, e.g., "host-1.test." is treated same as "host-1.test".
+     *
+     * @param[in]  aLabels            A name label string. Can be nullptr (then treated as "").
+     * @param[in]  aLength            The max length of the name labels to encode.
+     * @param[in]  aMessage           The message to which to append the encoded name.
+     *
+     * @retval OT_ERROR_NONE          Successfully encoded and appended the name label(s) to @p aMessage.
+     * @retval OT_ERROR_INVALID_ARGS  Name label @p aLabels is not valid.
+     * @retval OT_ERROR_NO_BUFS       Insufficient available buffers to grow the message.
+     *
+     */
+    static otError AppendMultipleLabels(const char *aLabels, uint8_t aLength, Message &aMessage);
 
     /**
      * This static method appends a name label terminator to a message.
@@ -906,8 +958,7 @@ public:
 private:
     enum : char
     {
-        kNullChar           = '\0',
-        kLabelSeperatorChar = '.',
+        kNullChar = '\0',
     };
 
     enum : uint8_t
@@ -955,8 +1006,6 @@ private:
         uint16_t       mNextLabelOffset;  // Offset in `mMessage` to the start of the next label.
         uint16_t       mNameEndOffset;    // Offset in `mMessage` to the byte after the end of domain name field.
     };
-
-    static otError AppendLabel(const char *aLabel, uint8_t aLabelLength, Message &aMessage);
 
     Name(const char *aString, const Message *aMessage, uint16_t aOffset)
         : mString(aString)
@@ -2374,6 +2423,12 @@ OT_TOOL_PACKED_BEGIN
 class Question
 {
 public:
+    /**
+     * Default constructor for Question.
+     *
+     */
+    Question() = default;
+
     /**
      * Constructor for Question.
      *
--- a/src/core/net/dnssd_server.cpp
+++ b/src/core/net/dnssd_server.cpp
@@ -1,0 +1,651 @@
+/*
+ *  Copyright (c) 2021, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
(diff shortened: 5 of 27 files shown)


src/core/net/srp_server.hpp
```diff
@@ -198,6 +198,17 @@ class Server : public InstanceLocator, private NonCopyable
          */
         bool Matches(const char *aFullName) const;
 
+        /**
+         * This method tells whether this service matches a given service name <Service>.<Domain>.
+         *
+         * @param[in] aServiceName  The full service name.
```
wgtdkp, line 204:
```suggestion
         * @param[in] aServiceName  The full service name less the instance name.
```

src/core/net/srp_server.hpp
```diff
@@ -198,6 +198,17 @@ class Server : public InstanceLocator, private NonCopyable
...
+         *
+         * @returns TRUE   If the service matches the full service name.
```
wgtdkp, line 206:
```suggestion
         * @retval TRUE   If the service matches the full service name.
```

tests/scripts/thread-cert/node.py
```diff
@@ -217,6 +217,58 @@ def bash(self, cmd: str) -> List[str]:
             else:
                 return lines
 
+    def dns_dig(self, server: str, name: str, qtype: str):
```
wgtdkp, line 220: It is helpful to give some examples of the return value of this method.

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,554 @@
...
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(!IsRunning());
+
+    SuccessOrExit(error = mSocket.Open(&Server::HandleUdpReceive, this));
+    SuccessOrExit(error = mSocket.Bind(kPort));
+
+exit:
+    otLogInfoDns("DNS-SD Server: started: %s", otThreadErrorToString(error));
```
wgtdkp, line 69: nit: since the logging methods already output the region label `-dns----`, we may just output `[server] ` for the server side and `[client] ` for the client side. We have this consistent logging format for both `Srp::Server` and `Srp::Client`, it would be good the follow the practice.
```suggestion
    otLogInfoDns("[server] started: %s", otThreadErrorToString(error));
```

wgtdkp: Another nit is, may better provide more readable log messages. `[server] started: kNone` is not very reader-friendly :)

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,554 @@
...
+    SuccessOrExit(error = responseMessage->SetLength(sizeof(Header)));
+
+    SuccessOrExit(error = ProcessQuery(aMessage, *responseMessage, requestHeader));
+
+exit:
+    if (error == OT_ERROR_NONE)
+    {
+        OT_ASSERT(responseMessage != nullptr);
+        error = SendMessage(*responseMessage, aMessageInfo);
+    }
```
wgtdkp, line 125:
```suggestion
    SuccessOrExit(error = SendMessage(*responseMessage, aMessageInfo));

exit:
```

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,554 @@
...
+    VerifyOrExit(question.GetClass() == ResourceRecord::kClassInternet ||
+                     question.GetClass() == ResourceRecord::kClassAny,
+                 responseHeader.SetResponseCode(Header::kResponseNotImplemented));
+
+    // prepare the information for name compression
+    VerifyOrExit(OT_ERROR_NONE == PrepareCompressInfo(qtype, name, nameSerializeOffset, compressInfo),
+                 responseHeader.SetResponseCode(Header::kResponseNameError));
+
+    // resolve the question
+    SuccessOrExit(error = ResolveQuestion(name, question, responseHeader, aResponse, compressInfo));
```
wgtdkp, line 178: For current implementation, we will not send a response when failed to resolve a question (we send the response in `Server::HandleUdpReceive` only when `error` is `OT_ERROR_NONE`). But I think we should send a response to the client in any fail cases as soon as we verified that the request message is a DNS query (we can just ignore the request if it is not a DNS query).

src/core/config/dnssd_server.h
```diff
@@ -0,0 +1,58 @@
...
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_DNSSD_SERVER_PORT
+ *
+ * Define the the DNS-SD Server port.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_DNSSD_SERVER_PORT
+#define OPENTHREAD_CONFIG_DNSSD_SERVER_PORT 53
```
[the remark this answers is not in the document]
simonlingoogle: Even if we are going to have a DNS over TCP implementation, we are going to use the same port. Isn't it?

src/core/net/dns_headers.cpp
```diff
@@ -293,6 +301,76 @@ otError Name::ReadName(const Message &aMessage, uint16_t &aOffset, char *aNameBu
     return error;
 }
 
+otError Name::SplitName(const char *aName, const char *aDomain, SplitInfo &aInfo)
```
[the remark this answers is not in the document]
simonlingoogle: It actually applies. 
The `aName` can be a instance name, service name, host name or even domain name itself, as I have described in the Doxygen.

src/core/net/dns_headers.hpp
```diff
@@ -491,6 +491,36 @@ class Name
         kMaxLength      = 255, ///< Max number of characters in a name.
     };
 
+    /**
+     * This structure represents the splitting information of a full name.
+     *
+     * @sa SplitName
+     *
+     */
+    struct SplitInfo
```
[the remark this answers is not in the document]
simonlingoogle: I would prefer the original name because the name can be an all kinds of names:

- instance name
- service name
- host name
- domain name

src/core/net/dns_headers.hpp
```diff
@@ -649,6 +724,25 @@ class Name
      */
     static otError ReadName(const Message &aMessage, uint16_t &aOffset, char *aNameBuffer, uint16_t aNameBufferSize);
 
+    /**
+     * This static method split a full name into different components.
```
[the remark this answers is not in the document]
simonlingoogle: as I have said, it can be all kinds of names.

src/core/net/srp_server.hpp
simonlingoogle: what do you mean?

src/core/net/dnssd_server.cpp
simonlingoogle: Note that DNS-SD is not exactly `DNS`. So, I think it's better to clearly states that it's `DNS-SD Server`.

src/core/net/dnssd_server.cpp
simonlingoogle: Actually we only return error when the server failure happend (e..g NoBufs). Thus, I think we can handle the errors by sending a ServerFailure RCODE.

src/core/net/dns_headers.hpp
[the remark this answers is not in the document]
simonlingoogle: rename to `NameComponentOffsetInfo`

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,632 @@
...
+                bool        serviceNameMatched  = service->MatchesServiceName(aName);
+                bool        instanceNameMatched = service->Matches(aName);
+                bool        ptrQueryMatched     = qtype == ResourceRecord::kTypePtr && serviceNameMatched;
+                bool        srvQueryMatched     = qtype == ResourceRecord::kTypeSrv && instanceNameMatched;
+                bool        txtQueryMatched     = qtype == ResourceRecord::kTypeTxt && instanceNameMatched;
+
+                if (service->IsDeleted())
+                {
+                    continue;
+                }
```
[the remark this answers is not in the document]
simonlingoogle: We already defined `Name::kLabelSeperatorChar`. I am going to use it.

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,632 @@
...
+}
+
+otError Server::FindPreviousLabel(const char *aName, uint8_t &aStart, uint8_t &aStop)
+{
+    otError error = OT_ERROR_NONE;
+    uint8_t i     = aStart;
+    uint8_t j;
+
+    VerifyOrExit(i > 0, error = OT_ERROR_NOT_FOUND);
+    VerifyOrExit(aName[--i] == '.', error = OT_ERROR_INVALID_ARGS);
```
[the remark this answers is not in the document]
simonlingoogle: We already defined `Name::kLabelSeperatorChar`. I am going to use it.

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,632 @@
...
+    : InstanceLocator(aInstance)
+    , mSocket(aInstance)
+{
+}
+
+otError Server::Start(void)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(!IsRunning());
```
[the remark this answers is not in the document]
simonlingoogle: I am going to remove `OT_ERROR_ALREADY` from doxygen.

tests/scripts/thread-cert/test_dnssd.py
```diff
@@ -0,0 +1,113 @@
...
+        ip, ttl = self.nodes[CLIENT1].dns_resolve(f"host2.{DOMAIN}", self.nodes[SERVER].get_mleid(), 53)
+        self.assertIn(ipaddress.IPv6Address(ip), map(ipaddress.IPv6Address, client2_addrs))
+
+        # TODO: test other query types using DNS-SD client
+
+    def _config_srp_client_services(self, client, instancename, hostname, port, priority, weight, addrs):
+        self.nodes[SERVER].netdata_show()
+        srp_server_port = self.nodes[SERVER].get_srp_server_port()
+
+        self.nodes[client].srp_client_start(self.nodes[SERVER].get_mleid(), srp_server_port)
```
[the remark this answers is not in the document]
simonlingoogle: I think it's our next enhancement for SRP and DNS-SD client to discover server automatically. Before that is implemented, we have to configure server address and port manually in test script.

@abtink  @librasungirl

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,632 @@
...
+
+    SuccessOrExit(error = mSocket.Open(&Server::HandleUdpReceive, this));
+    SuccessOrExit(error = mSocket.Bind(kPort));
+
+exit:
+    otLogInfoDns("[dns-sd server] started: %s", otThreadErrorToString(error));
```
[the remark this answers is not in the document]
simonlingoogle: Currently there is no other server in the DNS log region. I am fine to remove `dns-sd `.

src/core/net/dnssd_server.cpp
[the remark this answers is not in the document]
simonlingoogle: Can the compiler detect this and run `service->IsDeleted()` first before other assignments?

tests/scripts/thread-cert/test_dnssd.py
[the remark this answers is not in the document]
librasungirl: Fine to me to keep as is for now till more thorough discussion of discovery of SRP server.

src/core/thread/thread_netif.hpp
```diff
@@ -205,6 +206,9 @@ class ThreadNetif : public Ip6::Netif
 #if OPENTHREAD_CONFIG_SRP_CLIENT_ENABLE
     Srp::Client mSrpClient;
 #endif
+#if OPENTHREAD_CONFIG_DNSSD_SERVER_ENABLE
+    Dns::ServiceDiscovery::Server mDnsServer;
```
librasungirl, line 210: Do we want to name it as `mDnsServer` or `mDnsSdServer (/mDnssdServer)`?

librasungirl (approved):

src/core/config/dnssd_server.h
```diff
@@ -0,0 +1,58 @@
...
+#define CONFIG_DNSSD_SERVER_H_
+
+/**
+ * @def OPENTHREAD_CONFIG_DNSSD_SERVER_ENABLE
+ *
+ * Define to 1 to enable DNS-SD Server support.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_DNSSD_SERVER_ENABLE
+#define OPENTHREAD_CONFIG_DNSSD_SERVER_ENABLE 0
```
wgtdkp, line 45: nit: Just noticed that @abtink is using `OPENTHREAD_CONFIG_DNS_CLIENT_SERVICE_DISCOVERY_ENABLE` for the client, I think we should have aligned names.

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,646 @@
...
+                                                      *static_cast<const Ip6::MessageInfo *>(aMessageInfo));
+}
+
+void Server::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessageInfo)
+{
+    otError  error = OT_ERROR_NONE;
+    Header   requestHeader;
+    Message *responseMessage = nullptr;
+
+    // make sure DNS header is at offset zero to simplify server logic
```
wgtdkp, line 109:
```suggestion
    // Make sure DNS header is at offset zero to simplify server logic
```

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,646 @@
...
+    FreeMessageOnError(responseMessage, error);
+}
+
+void Server::ProcessQuery(Message &aMessage, Message &aResponse, const Header &aRequestHeader)
+{
+    Header           responseHeader;
+    uint16_t         readOffset, nameSerializeOffset;
+    Question         question;
+    uint16_t         qtype;
+    char             name[Dns::Name::kMaxLength + 1] = "";
```
wgtdkp, line 135:
```suggestion
    char             name[Dns::Name::kMaxLength + 1];
```

src/core/thread/thread_netif.hpp
simonlingoogle: I would prefer `mDnssdServer` . Thoughts?

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,646 @@
...
+otError Server::ResolveQuestion(const char *      aName,
+                                const Question &  aQuestion,
+                                Header &          aResponseHeader,
+                                Message &         aResponseMessage,
+                                NameCompressInfo &aCompressInfo)
+{
+    otError error = OT_ERROR_NONE;
+
+#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
+    {
```
wgtdkp, line 271:
```suggestion

```

tests/scripts/thread-cert/test_dnssd.py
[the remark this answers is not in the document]
simonlingoogle: Resolve this since for future enhancements

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,646 @@
...
+        if (aResponseHeader.GetAnswerCount() > 0)
+        {
+            SuccessOrExit(
+                error = ResolveQuestionBySrp(aName, aQuestion, aResponseHeader, aResponseMessage, true, aCompressInfo));
+        }
+        else
+        {
+            aResponseHeader.SetResponseCode(Header::kResponseNameError);
+        }
+    }
```
wgtdkp, line 284:
```suggestion
```

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,646 @@
...
+    aInfo.mDomainOffset = labelStart;
+
+    while (true)
+    {
+        error = FindPreviousLabel(aName, labelStart, labelStop);
+
+        VerifyOrExit(error == OT_ERROR_NONE, error = (error == OT_ERROR_NOT_FOUND ? OT_ERROR_NONE : error));
+
+        if (labelStop == labelStart + 4 && (memcmp(&aName[labelStart], kDnssdProtocolUdp, 4) == 0 ||
+                                            memcmp(&aName[labelStart], kDnssdProtocolTcp, 4) == 0))
```
wgtdkp, line 483:
```suggestion
                                            memcmp(&aName[labelStart], kDnssdProtocolTcp, sizeof(kDnssdProtocolTcp)) == 0))
```

wgtdkp (approved):
LGTM :+1:  Thanks @simonlingoogle

src/core/net/dnssd_server.cpp
simonlingoogle: Defined a new constant `kProtocolLabelLength = 4`. 

I intended not to use `sizeof(kDnssdProtocolTcp)` because if we changed the implementation to include '\0' in `kDnssdProtocolTcp`, the size would then be 5. I tend to keep the code more robust against potential future changes.

src/core/config/dnssd_server.h
simonlingoogle: I prefer to keep it as is since currently the server implementation is using a different philosophy from the client. 

There is no DNS server yet, so the DNS-SD server is only supposed to handle DNS-SD related requests. It is not an **Service Discovery** feature enhancement to some existing DNS server. However the DNS-SD client is actually extended features of the DNS client. So, I think it make sense for client code to have `OPENTHREAD_CONFIG_DNS_CLIENT_SERVICE_DISCOVERY_ENABLE` configuration.

simonlingoogle: I did thought about such strategy:
- Use `OPENTHREAD_CONFIG_DNS_SERVER_ENABLE` configuration to enable a DNS server feature (which just handle the UDP message and does nothing)
- Use `OPENTHREAD_CONFIG_DNS_SERVER_SERVICE_DISCOVERY_ENABLE` to enable the service discovery feature which resolves questions using SRP.

I didn't use this strategy because it seems unnecessary to me. Using just one `OPENTHREAD_CONFIG_DNSSD_SERVER_ENABLE` is simple and what we need at this moment. I think we can refactor it when we need a general DNS server on Thread devices. 

Thoughts?

librasungirl: fine to me to keep it as is

examples/common-switches.mk
```diff
@@ -50,6 +50,7 @@ DIAGNOSTIC                ?= 0
 DISABLE_DOC               ?= 0
 DISABLE_TOOLS             ?= 0
 DNS_CLIENT                ?= 0
+DNSSD_SERVER              ?= 0
```
librasungirl, line 53: add introduction on this feature switch in https://github.com/openthread/openthread/blob/master/examples/README.md

_@wgtdkp `BORDER_ROUTING `should be added as well._

src/core/thread/thread_netif.hpp
simonlingoogle: Renamed to mDnssdServer

tests/scripts/thread-cert/border_routing/test_dnssd_server.py
```diff
@@ -0,0 +1,281 @@
+#!/usr/bin/env python3
```
wgtdkp, line 1: Lets rename this directory to `border_router` and place all border router related tests in this directory, includes:
- border routing
- DNS-SD query
- advertising proxy
- discovery proxy

Thoughts?

simonlingoogle: We'd better resolve this in a separate PR?

simonlingoogle:
@abtink Please review this PR when you have time. We would prefer to merge this PR first and enhance the DNS-SD server with missing features (e.g. handling more than one questions). 
BTW, I think testing with this PR is the easiest way if you are to add test case for DNS-SD client #6116 .

@wgtdkp @librasungirl

abtink (approved):
Thanks @simonlingoogle . Looks great overall.
Some smaller suggestions below.
One thing we may want to add is to handle the "service resolution" query (where we question section contains two questions for SRV and TXT records for same name). I think we can check for this case and allow this.

src/core/net/dnssd_server.hpp
```diff
@@ -0,0 +1,296 @@
...
+
+    private:
+        const char *mDomainName;         // The serialized domain name (must NOT be nullptr)
+        const char *mServiceName;        // The serialized service name (must NOT be nullptr for PTR/SRV/TXT queries)
+        const char *mInstanceName;       // The serialized instance name or nullptr (only support one instance name)
+        const char *mHostName;           // The serialized host name or nullptr (only support one host name)
+        uint16_t    mDomainNameOffset;   // Offset of domain name serialization
+        uint16_t    mServiceNameOffset;  // Offset of service name serialization
+        uint16_t    mInstanceNameOffset; // Offset of instance name serialization
+        uint16_t    mHostNameOffset;     // Offset of host name serialization.
```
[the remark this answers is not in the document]
simonlingoogle: Actually this is the offset into the response `Message`. 
Will add comment.

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,659 @@
...
+    responseHeader.Clear();
+    responseHeader.SetResponseCode(Header::kResponseSuccess);
+    responseHeader.SetType(Header::kTypeResponse);
+    responseHeader.SetMessageId(aRequestHeader.GetMessageId());
+
+    // Validate the query
+    VerifyOrExit(aRequestHeader.GetQueryType() == Header::kQueryTypeStandard,
+                 responseHeader.SetResponseCode(Header::kResponseNotImplemented));
+    VerifyOrExit(!aRequestHeader.IsTruncationFlagSet(), responseHeader.SetResponseCode(Header::kResponseFormatError));
+    VerifyOrExit(aRequestHeader.GetQuestionCount() == 1, responseHeader.SetResponseCode(Header::kResponseFormatError));
```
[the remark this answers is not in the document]
simonlingoogle: Indeed we should. I will enhance it ASAP in another PR.

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,659 @@
...
+
+    // Add the question to the response, and save the serialize offset of name
+    nameSerializeOffset = aResponse.GetLength();
+    SuccessOrExit(error = AddQuestionToResponse(name, question, aResponse, responseHeader));
+
+    // Further validate the query
+    qtype = question.GetType();
+    VerifyOrExit(qtype == ResourceRecord::kTypePtr || qtype == ResourceRecord::kTypeSrv ||
+                     qtype == ResourceRecord::kTypeTxt || qtype == ResourceRecord::kTypeAaaa,
+                 responseHeader.SetResponseCode(Header::kResponseNotImplemented));
```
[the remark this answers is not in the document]
simonlingoogle: I was thinking about it but didn't do this way because of the calling to `ResolveQuestion` :
```c++
    // Resolve the question
    SuccessOrExit(error = ResolveQuestion(name, question, responseHeader, aResponse, compressInfo));
```

`ResolveQuestion` could set the RCODE and currently it just modify the `responseHeader`. If we define a new variable, we would need to add a out parameter `aResponseCode` to the `ResolveQuestion`. 

I am not sure which is simpler. I will revise it later with further experiments.

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,659 @@
...
+    SuccessOrExit(error = aMessage.Read(0, requestHeader));
+    VerifyOrExit(requestHeader.GetType() == Header::kTypeQuery, error = OT_ERROR_DROP);
+
+    responseMessage = NewMessage();
+    VerifyOrExit(responseMessage != nullptr, error = OT_ERROR_NO_BUFS);
+
+    // Allocate space for DNS header and move offset after DNS header
+    SuccessOrExit(error = responseMessage->SetLength(sizeof(Header)));
+
+    ProcessQuery(aMessage, *responseMessage, requestHeader);
```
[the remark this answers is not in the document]
simonlingoogle: Yes it is. It never fails to prepare the message. 

I will comment.

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,659 @@
...
+}
+
+void Server::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessageInfo)
+{
+    otError  error = OT_ERROR_NONE;
+    Header   requestHeader;
+    Message *responseMessage = nullptr;
+
+    // Make sure DNS header is at offset zero to simplify server logic
+    aMessage.RemoveHeader(aMessage.GetOffset());
```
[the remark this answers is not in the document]
simonlingoogle: Yes we can. I used to make sure that DNS header is at offset 0 to skimpily callings to a number of DNS utility methods. However this should be no longer necessary.

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,659 @@
...
+otError Server::ResolveQuestion(const char *      aName,
+                                const Question &  aQuestion,
+                                Header &          aResponseHeader,
+                                Message &         aResponseMessage,
+                                NameCompressInfo &aCompressInfo)
+{
+    otError error = OT_ERROR_NONE;
+
+#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
+    SuccessOrExit(error =
```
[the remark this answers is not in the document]
simonlingoogle: I will add `/* aAdditionl */` for now.

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,659 @@
...
+    // Make sure name ends with domain
+    VerifyOrExit(domainLen > 0 && nameLen >= domainLen, error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(memcmp(&aName[nameLen - domainLen], aDomain, domainLen) == 0, error = OT_ERROR_INVALID_ARGS);
+
+    labelStart          = nameLen - domainLen;
+    aInfo.mDomainOffset = labelStart;
+
+    while (true)
+    {
+        error = FindPreviousLabel(aName, labelStart, labelStop);
```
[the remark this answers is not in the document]
simonlingoogle: Yes. The code is trying to find the `_tcp` or `_udp` subdomain from back to front in the DNS name. It handles `extra` labels by skipping them.

src/core/net/dnssd_server.cpp
```diff
@@ -0,0 +1,651 @@
...
+    Header   requestHeader;
+    Message *responseMessage = nullptr;
+
+    SuccessOrExit(error = aMessage.Read(aMessage.GetOffset(), requestHeader));
+    VerifyOrExit(requestHeader.GetType() == Header::kTypeQuery, error = OT_ERROR_DROP);
+
+    responseMessage = mSocket.NewMessage(0);
+    VerifyOrExit(responseMessage != nullptr, error = OT_ERROR_NO_BUFS);
+
+    // Allocate space for DNS header and move offset after DNS header
```
jwhui, line 100: nit: where does the "move offset after DNS header" occur?

simonlingoogle: I used to move the offset, but the behavior was removed later.


Comment updated.

jwhui (approved):

PR #5560, opened 2020-09-21
[mle] remove secure data request mode bit

jwhui:
- Always set to 1 on transmission.
- Always ignore on reception.

Resolves #5559

PR #6107, opened 2021-01-21
[mle] ensure reserved bit in Mode TLV is set

LuDuda:
The #5560 removed a need of setting the Secure Request Bit in the Mode TLV and supposed to set this bit to 1 always on transmission.

The reserved bit is set when `DeviceMode::Set()` method is invoked. If application uses MTD library, and does not call `otThreadSetLinkMode()`, stack does not set the reserve bit at all - since it doesn't call above method (even from [constructor](https://github.com/openthread/openthread/blob/master/src/core/thread/mle_types.hpp#L348)).

This PR ensures that reserved bit is set.

Note that i found this problem when End Device continously was deattaching from the parent after Child ID Timeout, and retried attachment procedure. The issue was that End Device [dropped](https://github.com/openthread/openthread/blob/master/src/core/thread/mle.cpp#L3679) `Child Update Response` since the `Mode TLV` sent from the Child was not the same as one sent by the Parent. Basically `Child Update Request` didn't have reserved bit set, while `Child Update Response` had. 

Let me know if you think about different solution, i just aligned the implementation to one in `mle_router.cpp`.

### DIFF
--- a/src/core/thread/mle_types.hpp
+++ b/src/core/thread/mle_types.hpp
@@ -345,10 +345,7 @@ public:
      * @param[in] aMode   A mode TLV bitmask to initialize the `DeviceMode` object.
      *
      */
-    explicit DeviceMode(uint8_t aMode)
-        : mMode(aMode)
-    {
-    }
+    explicit DeviceMode(uint8_t aMode) { Set(aMode); }
 
     /**
      * This constructor initializes a `DeviceMode` object from a given mode configuration structure.


src/core/thread/mle.cpp
```diff
@@ -112,6 +111,8 @@ Mle::Mle(Instance &aInstance)
 {
     MeshLocalPrefix meshLocalPrefix;
 
+    mDeviceMode.Set(DeviceMode::kModeRxOnWhenIdle);
```
jwhui, line 114: Thanks for identifying this issue!

What about moving the call to `Set()` within the constructor?

https://github.com/openthread/openthread/blob/099333375f4b087ac6063140dc2c4f76db892637/src/core/thread/mle_types.hpp#L348

LuDuda: I initially thought about it, but my concern was [here](https://github.com/openthread/openthread/blob/36b16be3a4aebdd9e5d1a18dde984ccff0da8455/src/core/thread/mle.cpp#L3679) - that we won't check exactly what is in the Mode TLV. However, if the bit should be ignored on reception it actually might be better solution. Do you think we should move the `Set` into constructor then?

jwhui: Good points. As you mentioned, the reserved bit should be ignored on reception. We should fix this so that we conform to the spec.

Irving-cl (approved):
LGTM :+1:

librasungirl (approved):

MarekPorwisz (approved):

abtink (approved):
LGTM. Thanks.

edmont (approved):

jwhui (approved):

PR #6111, opened 2021-01-25
[udp] fix UDP not accessible via RLOC/ALOC with PLAT_UDP

simonlingoogle:
This commit makes sure that UDP servers are accessible via all IPv6 addresses when PLAT_UDP is on.

We still need to pass the messages with ALOC/RLOC destinations to the tun interface when PLAT_UDP is used. Otherwise, UDP servers can not receive any message when PLAT_UDP is enabled.

- [x] Fix the bug
- [x] Add test to verify a UDP server is accessible via MLEID / LINK-LOCAL / OMR / RLOC / ALOC (should fail due to the bug)

This PR deprecates #6057

### DIFF
--- a/src/core/net/ip6.cpp
+++ b/src/core/net/ip6.cpp
@@ -1009,12 +1009,14 @@ otError Ip6::ProcessReceiveCallback(Message &          aMessage,
 
     if (mIsReceiveIp6FilterEnabled)
     {
+#if !OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE
         // do not pass messages sent to an RLOC/ALOC, except Service Locator
         bool isLocator = Get<Mle::Mle>().IsMeshLocalAddress(aMessageInfo.GetSockAddr()) &&
                          aMessageInfo.GetSockAddr().GetIid().IsLocator();
 
         VerifyOrExit(!isLocator || aMessageInfo.GetSockAddr().GetIid().IsAnycastServiceLocator(),
                      error = OT_ERROR_NO_ROUTE);
+#endif
 
         switch (aIpProto)
         {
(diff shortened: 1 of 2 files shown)


src/core/net/ip6.cpp
```diff
@@ -1010,11 +1010,13 @@ otError Ip6::ProcessReceiveCallback(Message &          aMessage,
     if (mIsReceiveIp6FilterEnabled)
     {
         // do not pass messages sent to an RLOC/ALOC, except Service Locator
+#if !OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE
```
simonlingoogle, line 1013: Explain: we still need to pass the messages with ALOC/RLOC destinations to the tun interface when PLAT_UDP is used. Otherwise, UDP servers can not receive any message when PLAT_UDP is enabled.

src/core/net/ip6.cpp
```diff
@@ -1033,30 +1035,10 @@ otError Ip6::ProcessReceiveCallback(Message &          aMessage,
         case kProtoUdp:
         {
             Udp::Header udp;
-            uint16_t    destPort;
 
             IgnoreError(aMessage.Read(aMessage.GetOffset(), udp));
+            VerifyOrExit(Get<Udp>().ShouldUsePlatformUdp(udp.GetDestinationPort()), error = OT_ERROR_NO_ROUTE);
```
simonlingoogle, line 1040: This change is not part of the fix. However, I think the new code is simpler.

bukepo (approved):
LGTM:+1:

wgtdkp (approved):
LGTM :+1:

src/core/net/ip6.cpp
librasungirl, line 1013: nit: move it before the comment

librasungirl (approved):

jwhui (approved):

PR #6113, opened 2021-01-26
[srp] get parsed TXT entry from SRP server

wgtdkp:
This PR returns parsed TXT entries for service from the SRP server, it enables prettier CLI output and
easier Advertising Proxy processing.  We moved the `otSrpTxtEntry` to `include/openthread/dns.h`
because this structure can ( and will) also be used by DNS-SD client. We also added one more `mKeyLength`
field in `otDnsTxtEntry` for saving length of the TXT key so that we can iterate over the encoded TXT
data without dynamically allocating memory for each entry.

### DIFF
--- a/include/openthread/dns.h
+++ b/include/openthread/dns.h
@@ -58,6 +58,45 @@ extern "C" {
 #define OT_DNS_DEFAULT_SERVER_PORT 53                   ///< Defines default DNS Server port.
 
 /**
+ * Initializer for otDnsTxtIterator.
+ */
+#define OT_DNS_TXT_ITERATOR_INIT 0
+
+typedef uint16_t otDnsTxtIterator; ///< Used to iterate through the TXT entries.
+
+/**
+ * This structure represents a TXT record entry representing a key/value pair (RFC 6763 - section 6.3).
+ *
+ * The string buffers pointed to by `mKey` and `mValue` MUST persist and remain unchanged after an instance of such
+ * structure is passed to OpenThread (as part of `otSrpClientService` instance).
+ *
+ * An array of `otDnsTxtEntry` entries are used in `otSrpClientService` to specify the full TXT record (a list of
+ * entries).
+ *
+ */
+typedef struct otDnsTxtEntry
+{
+    /**
+     * The TXT record key string. It doesn't need to be a null-terminated string and `mKeyLength` gives its length.
+     *
+     * If `mKey` is not NULL, then the entry is treated as key/value pair with `mValue` buffer providing the value.
+     *   - The entry is encoded as follows:
+     *        - A single string length byte followed by "key=value" format (without the quotation marks).
+              - In this case, the overall encoded length must be 255 bytes or less.
+     *   - If `mValue` is NULL, then key is treated as a boolean attribute and encoded as "key" (with no `=`).
+     *   - If `mValue` is not NULL but `mValueLength` is zero, then it is treated as empty value and encoded as "key=".
+     *
+     * If `mKey` is NULL, then `mValue` buffer is treated as an already encoded TXT-DATA and is appended as is in the
+     * DNS message.
+     *
+     */
+    const char *   mKey;
+    const uint8_t *mValue;       ///< The TXT record value or already encoded TXT-DATA (depending on `mKey`).
+    uint16_t       mValueLength; ///< Number of bytes in `mValue` buffer.
+    uint8_t mKeyLength; ///< Number of bytes in `mKey` buffer. MUST be set even if `mKey` is a null-terminated string.
+} otDnsTxtEntry;
+
+/**
  * This structure implements DNS Query parameters.
  *
  */
--- a/src/core/net/dns_headers.hpp
+++ b/src/core/net/dns_headers.hpp
@@ -36,6 +36,8 @@
 
 #include "openthread-core-config.h"
 
+#include <openthread/dns.h>
+
 #include "common/clearable.hpp"
 #include "common/encoding.hpp"
 #include "common/message.hpp"
@@ -794,6 +796,56 @@ private:
 };
 
 /**
+ * This type represents a TXT record entry representing a key/value pair (RFC 6763 - section 6.3).
+ *
+ */
+class TxtEntry : public otDnsTxtEntry
+{
+    friend class TxtRecord;
+
+public:
+    /**
+     * This method encodes and appends the `TxtEntry` to a message.
+     *
+     * @param[in] aMessage  The message to append to.
+     *
+     * @retval OT_ERROR_NONE           Entry was appended successfully to @p aMessage.
+     * @retval OT_ERROR_INVALID_ARGS   The `TxTEntry` info is not valid.
+     * @retval OT_ERROR_NO_BUFS        Insufficient available buffers to grow the message.
+     *
+     */
+    otError AppendTo(Message &aMessage) const;
+
+    /**
+     * This static method appends an array of `TxtEntry` items to a message.
+     *
+     * @param[in] aEntries     A pointer to array of `TxtEntry` items.
+     * @param[in] aNumEntries  The number of entries in @p aEntries array.
+     * @param[in] aMessage     The message to append to.
+     *
+     *
+     * @retval OT_ERROR_NONE           Entries appended successfully to @p aMessage.
+     * @retval OT_ERROR_INVALID_ARGS   The `TxTEntry` info is not valid.
+     * @retval OT_ERROR_NO_BUFS        Insufficient available buffers to grow the message.
+     *
+     */
+    static otError AppendEntries(const TxtEntry *aEntries, uint8_t aNumEntries, Message &aMessage);
+
+private:
+    enum : char
+    {
+        kKeyValueSeparator = '=',
+    };
+
+    enum : uint8_t
+    {
+        kMinKeyLength           = 1,
+        kMaxKeyLength           = 9,
+        kMaxKeyValueEncodedSize = 255,
+    };
+};
+
+/**
  * This class implements Resource Record (RR) body format.
  *
  */
@@ -1146,6 +1198,8 @@ public:
         kType = kTypeTxt, ///< The TXT record type.
     };
 
+    typedef otDnsTxtIterator TxtIterator;
+
     /**
      * This method initializes the TXT Resource Record by setting its type and class.
      *
@@ -1158,6 +1212,8 @@ public:
 
     /**
      * This method parses and reads the TXT record data from a message.
+     *
+     * This method also checks if the TXT data is well-formed by calling `VerifyTxtData()`.
      *
      * @param[in]     aMessage          The message to read from.
      * @param[inout]  aOffset           On input, the offset in @p aMessage to start of TXT record data.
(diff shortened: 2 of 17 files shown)


include/openthread/dns.h
```diff
@@ -57,6 +57,45 @@ extern "C" {
...
+
+typedef uint16_t otDnsTxtIterator; ///< Used to iterate through the TXT entries.
+
+/**
+ * This structure represents a TXT record entry representing a key/value pair (RFC 6763 - section 6.3).
+ *
+ * The strings buffers pointed to by `mKey` and `mValue` MUST persist and remain unchanged after an instance of such a
+ * structure is passed to OpenThread (as part of `otSrpClientService` instance).
+ *
+ * An array of `otDnsTxtEntry` entries is used in `otSrpClientService` to specify the full TXT record (a list of
```
Irving-cl, line 73:
```suggestion
 * An array of `otDnsTxtEntry` entries are used in `otSrpClientService` to specify the full TXT record (a list of
```

include/openthread/dns.h
```diff
@@ -57,6 +57,45 @@ extern "C" {
...
+ * Initializer for otDnsTxtIterator.
+ */
+#define OT_DNS_TXT_ITERATOR_INIT 0
+
+typedef uint16_t otDnsTxtIterator; ///< Used to iterate through the TXT entries.
+
+/**
+ * This structure represents a TXT record entry representing a key/value pair (RFC 6763 - section 6.3).
+ *
+ * The strings buffers pointed to by `mKey` and `mValue` MUST persist and remain unchanged after an instance of such a
```
Irving-cl, line 70:
```suggestion
 * The string buffers pointed to by `mKey` and `mValue` MUST persist and remain unchanged after an instance of such
```

include/openthread/dns.h
```diff
@@ -57,6 +57,45 @@ extern "C" {
...
+     *   - If `mValue` is not NULL but `mValueLength` is zero, then it is treated as empty value and encoded as "key=".
+     *
+     * If `mKey` is NULL, then `mValue` buffer is treated as an already encoded TXT-DATA and is appended as is in the
+     * DNS message.
+     *
+     */
+    const char *   mKey;         ///< The TXT record key. Doesn't need to be NULL-terminated string.
+    const uint8_t *mValue;       ///< The TXT record value or already encoded TXT-DATA (depending on `mKey`).
+    uint16_t       mValueLength; ///< Number of bytes in `mValue` buffer.
+    uint8_t mKeyLength; ///< Number of bytes in `mKey` buffer. MUST be set even if `mKey` is a NULL-terminated string.
```
Irving-cl, line 96: align?

Irving-cl (approved):
LGTM 👍

src/core/net/srp_server.cpp
```diff
@@ -1324,6 +1324,50 @@ otError Server::Service::SetFullName(const char *aFullName)
...
+        uint8_t       length = mTxtData[i++];
+
+        entry.mKey         = reinterpret_cast<const char *>(mTxtData + i);
+        entry.mKeyLength   = length;
+        entry.mValue       = nullptr;
+        entry.mValueLength = 0;
+
+        for (uint8_t j = 0; j < length; ++j)
+        {
+            if (mTxtData[i + j] == kKeyValueSeparator)
```
simonlingoogle, line 1343:
```suggestion
            if (mTxtData[i + j] == kKeyValueSeparator)
```

Can we remove `mKeyLength` field but replace each `kKeyValueSeparator` to `\0` when the mTxtData is stored?

This would allow us to keep the `mKey` as a null-terminated string, which I think is better.

wgtdkp: That is the original idea in my mind, but there could be no `kKeyValueSeparator` for a TXT entry:

> If there is no '=' in a DNS-SD TXT record string, then it is a
   boolean attribute, simply identified as being present, with no value.

https://tools.ietf.org/html/rfc6763#section-6.4

include/openthread/dns.h
wgtdkp: `clang-format` did this :), I think it is because the comment is too long...

src/core/net/srp_client.hpp
```diff
@@ -96,7 +96,7 @@ class Client : public InstanceLocator, private NonCopyable
      * This type represents a TXT record entry representing a key/value pair (RFC 6763 - section 6.3).
      *
      */
-    class TxtEntry : public otSrpTxtEntry
+    class TxtEntry : public otDnsTxtEntry
```
abtink, line 99: Since the `oDnsTxtEntry` is now a common public DNS related type, shall we also move the `TxtEntry` definition from under `Srp::Clinet` here into a common internal header, say `dns_header.hpp` (or a new common header for use by core). Since we have this c++ type in the core, I think would better to use in SRP server as well (instead of directly using the public C `otDnsTxtEntry` definition)

Another idea/suggestion is to see if we can somehow define the method that does the iteration over TXT entries (and the definition of iterator) into the `TxtEntry` class as well.

For example, the method can be like:
```
otError TxtEntry::GetNextEntry(uint8_t *aTxtData, uint16_t aTxtLength, Iterator &aIterator, TxEntry &aEntry);
```
i.e., we pass in the encoded TXT data and its length and the current iterator value and it then populates the next entry.

src/core/net/srp_server.cpp
```diff
@@ -1324,6 +1324,50 @@ otError Server::Service::SetFullName(const char *aFullName)
     return error;
 }
 
+otError Server::Service::GetNextTxtEntry(otDnsTxtIterator &aIterator, otDnsTxtEntry &aTxtEntry) const
+{
+    otError error = OT_ERROR_NONE;
+
+    for (uint16_t i = aIterator; i < mTxtLength;)
+    {
+        otDnsTxtEntry entry;
```
abtink, line 1333: Was thinking if we can directly populating the `aTxtEntry` fields? Any reason we may not want to?

src/core/net/srp_server.cpp
abtink: I am fine with us specifcing the keylength as well.

If we want to make `mKey` be null-terminted, one idea can be to use  a more compelx `iterator` type (e.g., a struct containing a `index` and a `char`). 
Basically when we replace a char in `mTxtData` with `\0` (to make it null terminted), we save the replaced char in the `Iterator` itself and then on next call we replace it back. I think this may work but I see that it can get complex though.

src/core/net/srp_server.cpp
```diff
@@ -1324,6 +1324,50 @@ otError Server::Service::SetFullName(const char *aFullName)
...
+        uint8_t       length = mTxtData[i++];
+
```
abtink, line 1335: Should we do a check hereto make sure that the entire txt entry is within `mTxtData` and it` mTxtLength`?
Maybe something like this:
``` 
  VerifyOrExit(i + length <= mTxtLength, error = OT_ERROR_PARSE);
```
This is to ensure that when we return the `mKey` and `mValue` pointers (and their length) the buffers are all within the `mTxtData` buffer and that we won't read beyond end of buffer.

abtink (commented):
Looks good overall. Thanks @wgtdkp.
Some suggestions below.

src/core/net/srp_client.hpp
wgtdkp: Thanks, I forgot about the `TxtEntry` class in `Srp::Client`. You proposal sounds good to me. I will move the iterator method to the `TxtEntry` class and move the `TxtEntry` into `dns_headers.hpp`.

wgtdkp: Actually, I added the `GetNextEntry` function to `TxtRecord` as it looks like a more appropriate place :)

src/core/net/srp_server.cpp
wgtdkp: updated.

src/core/net/srp_server.cpp
wgtdkp: There is another method `VerifyTxtData` that makes sure the TXT data is well-formed, and this method accepts only valid TXT data. The idea is to reject invalid TXT data as a DNS format error before each entry is queried. This makes sure we don't accept a portion of the TXT data. I updated the documentation to make this clear.

include/openthread/dns.h
```diff
@@ -57,6 +57,45 @@ extern "C" {
 #define OT_DNS_DEFAULT_SERVER_IP "2001:4860:4860::8888" ///< Defines default DNS Server address - Google DNS.
 #define OT_DNS_DEFAULT_SERVER_PORT 53                   ///< Defines default DNS Server port.
 
+/**
```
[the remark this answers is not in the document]
wgtdkp: > I think we should keep a openthread/dns.h for all common definitions between srp/dns server and client (such as this one) and then have different headers for each client/server.

Makes sense to me.

src/core/net/srp_server.cpp
wgtdkp: I also thought about it but gave up because it seems too dirty :)

src/core/net/srp_server.cpp
abtink: Nice! Did not notice `VerifyTxtData`. Sounds good to me.

include/openthread/dns.h
```diff
@@ -57,6 +57,45 @@ extern "C" {
...
+ * structure is passed to OpenThread (as part of `otSrpClientService` instance).
+ *
+ * An array of `otDnsTxtEntry` entries are used in `otSrpClientService` to specify the full TXT record (a list of
+ * entries).
+ *
+ */
+typedef struct otDnsTxtEntry
+{
+    /**
+     * The TXT record key string. It doesn't need to be a NULL-terminated string and `mKeyLength` gives its length.
```
abtink, line 80:
```suggestion
     * The TXT record key string. It doesn't need to be a null-terminated string and `mKeyLength` gives its length.
```

include/openthread/dns.h
```diff
@@ -57,6 +57,45 @@ extern "C" {
...
+     *   - If `mValue` is not NULL but `mValueLength` is zero, then it is treated as empty value and encoded as "key=".
+     *
+     * If `mKey` is NULL, then `mValue` buffer is treated as an already encoded TXT-DATA and is appended as is in the
+     * DNS message.
+     *
+     */
+    const char *   mKey;
+    const uint8_t *mValue;       ///< The TXT record value or already encoded TXT-DATA (depending on `mKey`).
+    uint16_t       mValueLength; ///< Number of bytes in `mValue` buffer.
+    uint8_t mKeyLength; ///< Number of bytes in `mKey` buffer. MUST be set even if `mKey` is a NULL-terminated string.
```
abtink, line 96:
```suggestion
    uint8_t mKeyLength; ///< Number of bytes in `mKey` buffer. MUST be set even if `mKey` is a null-terminated string.
```

src/core/net/dns_headers.hpp
```diff
@@ -1178,6 +1232,37 @@ class TxtRecord : public ResourceRecord
                         uint8_t *      aTxtBuffer,
                         uint16_t &     aTxtBufferSize) const;
 
+    /**
```
abtink, line 1235: How about we define a typedef for `TxtIterator` here?
```
typedef otDnsTxtIterator TxtIterator;
```
I like (and persoanlly always try) to re-define `ot{Type}` if/when we can and not use them in core module. 😃

src/core/net/dns_headers.hpp
```diff
@@ -1178,6 +1232,37 @@ class TxtRecord : public ResourceRecord
...
+     *
+     * @returns  TRUE if @p aTxtData contains valid encoded TXT data, FALSE if not.
+     *
+     */
+    static bool VerifyTxtData(const uint8_t *aTxtData, uint16_t aTxtLength);
+
+    /**
+     * This static method returns the next TXT entry in the encoded TXT data buffer.
+     *
+     * This method should be called when @p aTXTData has been verified by VerifyTxtData.
```
abtink, line 1249:
```suggestion
     * This method should be called when @p aTxtData has been verified by VerifyTxtData.
```

src/core/net/srp_server.hpp
```diff
@@ -199,6 +198,11 @@ class Server : public InstanceLocator, private NonCopyable
         bool Matches(const char *aFullName) const;
 
     private:
+        enum : char
```
abtink, line 201: I think we can remove this now?

abtink (approved):
LGTM. Thanks. 👍 
Few small nit suggestions below.

src/core/net/dns_headers.hpp
wgtdkp: SGTM, updated.

src/core/net/srp_server.hpp
wgtdkp: yes, thanks!

simonlingoogle (approved):
LGTM 👍

src/core/net/dns_headers.hpp
```diff
@@ -1178,6 +1234,37 @@ class TxtRecord : public ResourceRecord
...
+     *
+     * @returns  TRUE if @p aTxtData contains valid encoded TXT data, FALSE if not.
+     *
+     */
+    static bool VerifyTxtData(const uint8_t *aTxtData, uint16_t aTxtLength);
+
+    /**
+     * This static method returns the next TXT entry in the encoded TXT data buffer.
+     *
+     * This method should be called when @p aTxtData has been verified by VerifyTxtData.
```
jwhui, line 1251:
```suggestion
     * This method assumes that @p aTxtData has already been verified by `VerifyTxtData()`.
```

src/core/net/dns_headers.hpp
```diff
@@ -1159,6 +1213,8 @@ class TxtRecord : public ResourceRecord
     /**
      * This method parses and reads the TXT record data from a message.
      *
+     * This method also checks if the TXT data is well-formed by calling VerifyTxtData.
```
jwhui, line 1216:
```suggestion
     * This method also checks if the TXT data is well-formed by calling `VerifyTxtData()`.
```

jwhui (approved):

PR #6114, opened 2021-01-26
[dns] match DNS name with specific domain

wgtdkp:
This PR includes two small commits:

The first commit adds a new method `Dns::Name::IsSubDomainOf` that matches a DNS name with given domain.
This is useful for SRP and DNS-SD server to verify if a RR is in our authorized domain.

The second commit adds domain checks for names in a SRP update to the SRP server.

### DIFF
--- a/src/core/net/dns_headers.cpp
+++ b/src/core/net/dns_headers.cpp
@@ -545,15 +545,24 @@ bool Name::IsSubDomainOf(const char *aName, const char *aDomain)
     uint16_t nameLength   = StringLength(aName, kMaxLength);
     uint16_t domainLength = StringLength(aDomain, kMaxLength);
 
-    VerifyOrExit(nameLength >= domainLength && domainLength >= 1);
+    if (nameLength > 0 && aName[nameLength - 1] == kLabelSeperatorChar)
+    {
+        --nameLength;
+    }
 
+    if (domainLength > 0 && aDomain[domainLength - 1] == kLabelSeperatorChar)
+    {
+        --domainLength;
+    }
+
+    VerifyOrExit(nameLength >= domainLength);
     aName += nameLength - domainLength;
-    VerifyOrExit(memcmp(aName, aDomain, domainLength) == 0);
 
     if (nameLength > domainLength)
     {
         VerifyOrExit(aName[-1] == kLabelSeperatorChar);
     }
+    VerifyOrExit(memcmp(aName, aDomain, domainLength) == 0);
 
     match = true;
 
--- a/src/core/net/dns_headers.cpp
+++ b/src/core/net/dns_headers.cpp
@@ -539,6 +539,28 @@ bool Name::LabelIterator::CompareLabel(const LabelIterator &aOtherIterator) cons
                                  mLabelLength);
 }
 
+bool Name::IsSubDomainOf(const char *aName, const char *aDomain)
+{
+    bool     match        = false;
+    uint16_t nameLength   = StringLength(aName, kMaxLength);
+    uint16_t domainLength = StringLength(aDomain, kMaxLength);
+
+    VerifyOrExit(nameLength >= domainLength && domainLength >= 1);
+
+    aName += nameLength - domainLength;
+    VerifyOrExit(memcmp(aName, aDomain, domainLength) == 0);
+
+    if (nameLength > domainLength)
+    {
+        VerifyOrExit(aName[-1] == kLabelSeperatorChar);
+    }
+
+    match = true;
+
+exit:
+    return match;
+}
+
 otError ResourceRecord::ParseRecords(const Message &aMessage, uint16_t &aOffset, uint16_t aNumRecords)
 {
     otError error = OT_ERROR_NONE;
--- a/src/core/net/dns_headers.hpp
+++ b/src/core/net/dns_headers.hpp
@@ -888,6 +888,17 @@ public:
      */
     static otError CompareName(const Message &aMessage, uint16_t &aOffset, const Name &aName);
 
+    /**
+     * This static method tests if a DNS name is a sub-domain of a given domain.
+     *
+     * @param[in]  aName    The dot-separated name.
+     * @param[in]  aDomain  THe dot-separated domain.
+     *
+     * @returns  TRUE if the name is a sub-domain of @p aDomain, FALSE if is not.
+     *
+     */
+    static bool IsSubDomainOf(const char *aName, const char *aDomain);
+
 private:
     enum : char
     {
--- a/src/core/net/dns_headers.hpp
+++ b/src/core/net/dns_headers.hpp
@@ -891,8 +891,10 @@ public:
     /**
      * This static method tests if a DNS name is a sub-domain of a given domain.
      *
+     * Both @p aName and @p aDomain can end without dot ('.').
+     *
      * @param[in]  aName    The dot-separated name.
-     * @param[in]  aDomain  THe dot-separated domain.
+     * @param[in]  aDomain  The dot-separated domain.
      *
      * @returns  TRUE if the name is a sub-domain of @p aDomain, FALSE if is not.
      *
--- a/src/core/net/srp_server.cpp
+++ b/src/core/net/srp_server.cpp
@@ -223,11 +223,11 @@ void Server::RemoveAndFreeHost(Host *aHost)
     aHost->Free();
 }
 
-Server::Service *Server::FindService(const char *aFullName)
+const Server::Service *Server::FindService(const char *aFullName) const
 {
-    Service *service = nullptr;
+    const Service *service = nullptr;
 
-    for (Host *host = mHosts.GetHead(); host != nullptr; host = host->GetNext())
+    for (const Host *host = mHosts.GetHead(); host != nullptr; host = host->GetNext())
     {
         service = host->FindService(aFullName);
         if (service != nullptr)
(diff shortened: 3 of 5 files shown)


src/core/net/dns_headers.hpp
```diff
@@ -735,6 +735,17 @@ class Name
      */
     static otError CompareName(const Message &aMessage, uint16_t &aOffset, const Message &aMessage2, uint16_t aOffset2);
 
+    /**
+     * This static method tests if a DNS name ends with given domain.
+     *
+     * @param[in]  aName    The dot-separated name.
```
simonlingoogle, line 741: I think both `aName` and `aDomain` should end with dot. 
Make it clear in doxygen?

src/core/net/dns_headers.hpp
```diff
@@ -735,6 +735,17 @@ class Name
...
+     * @param[in]  aDomain  THe dot-separated domain.
+     *
+     * @returns  TRUE if the name matches the domain, FALSE if doesn't match.
+     *
+     */
+    static bool MatchDomain(const char *aName, const char *aDomain);
```
simonlingoogle, line 747: I am hoping to use this method for match service instance name with service name. Can we rename it to be more general? e.g. `IsSubNameOf`.

src/core/net/dns_headers.hpp
wgtdkp: done.

src/core/net/dns_headers.hpp
wgtdkp: I think the DNS name system is built atop of the `domain` concept. For example, `_ipp._tcp.example.com.` is a subdomain but not a subname:

> Alternatively, a printer manager device could implement enough of the
   DNS protocol that it is able to answer DNS queries directly, and
   Example Co.'s main DNS server could delegate the
   "_ipp._tcp.example.com." subdomain to the printer manager device.

https://tools.ietf.org/html/rfc6763#section-10

So I think `IsSubDomainOf` is a more generic and appropriate name, if you prefer.

simonlingoogle: I think `IsSubDomain` can be a good choice.

tests/unit/test_dns.cpp
```diff
@@ -38,6 +38,34 @@
...
+    }
+
+    // Verify it doesn't match a portion of a label.
+    {
+        name   = "my-service._ipps._tcp.default.service.arpa.";
+        domain = "vice.arpa.";
+        VerifyOrQuit(!Dns::Name::MatchDomain(name, domain), "Name::MatchDomain() succeed");
+    }
+}
+
```
abtink, line 68: Should call the newly added `TestDnsNameMatchDomain` from `main()` below?

abtink: Actually I think it may be better if we add this code directly in `TestDnsName()` function itself? 
It's the function that tests all other methods in `Dns::Name` so makes sense for it to cover the `MatchDomain` as well.

Thoughts?

abtink (approved):
Overall looks great. 👍 
One comment about the test below.

src/core/net/dns_headers.hpp
```diff
@@ -735,6 +735,19 @@ class Name
      */
     static otError CompareName(const Message &aMessage, uint16_t &aOffset, const Message &aMessage2, uint16_t aOffset2);
 
+    /**
+     * This static method tests if a DNS name ends with given domain.
+     *
+     * @p aName and @p aDomain MUST (or not) end with dot ('.') at the same time.
```
simonlingoogle, line 741: I think we can disallow strings that do not end with dot unless it's actually useful.

simonlingoogle (approved):
LGTM :+1: 

with a nit

tests/unit/test_dns.cpp
wgtdkp: Thanks, I moved it to `TestDnsName()`.

src/core/net/dns_headers.hpp
abtink: I think when we read name from `message` we always append an ending  `.` to it (which if I recal correctly is the recommend way to represent the DNS names).

We can also consider here to handle it in the implementation of this method and allow either name to end or not end with `.` (basically making this method handle all cases, and making it more flexible to use).

Thoughts?

src/core/net/dns_headers.cpp
```diff
@@ -531,6 +531,28 @@ bool Name::LabelIterator::CompareLabel(const LabelIterator &aOtherIterator) cons
...
+
+    VerifyOrExit(nameLength >= domainLength && domainLength >= 1);
+
+    aName += nameLength - domainLength;
+    VerifyOrExit(memcmp(aName, aDomain, domainLength) == 0);
+
+    if (nameLength > domainLength)
+    {
+        VerifyOrExit(aName[-1] == kLabelSeperatorChar);
+    }
```
librasungirl, line 548: nit: move it before memcpy for efficiency -- exit earlier when possible?

librasungirl (approved):

src/core/net/dns_headers.hpp
simonlingoogle: I would prefer to make it more strict in the OT core logic so as to make the code simpler (handling less conditions). 

For example, if we allow domain names not to end with dot, then one question is that should we allow a empty domain (i.e. "") ?

Also, if we allow the domain names not to end with dot, then this method should work no matter `aName` or `aDomain` ends with dot. For example, it should work when `aName` ends with dot but `aDomain` does not. 

Thoughts? @wgtdkp @abtink

wgtdkp: There are no server side cases that the names will end without `.`, but I think there could such cases on the client side (e.g. the name or domain given by the user). I will change the implementation to allow it.

tests/unit/test_dns.cpp
```diff
@@ -145,6 +145,31 @@ void TestDnsName(void)
...
+    printf("----------------------------------------------------------------\n");
+    printf("Verify domain name match:\n");
+
+    {
+        const char *subDomain;
+        const char *domain;
+
+        subDomain = "my-service._ipps._tcp.local.";
+        domain    = "local.";
+        VerifyOrQuit(Dns::Name::IsSubDomainOf(subDomain, domain), "Name::IsSubDomainOf() failed");
```
simonlingoogle, line 157:
```suggestion
        subDomain = "my-service._ipps._tcp.local.";
        domain    = "local.";
        VerifyOrQuit(Dns::Name::IsSubDomainOf(subDomain, domain), "Name::IsSubDomainOf() failed");
        
        subDomain = "my-service._ipps._tcp.local";
        domain    = "local.";
        VerifyOrQuit(Dns::Name::IsSubDomainOf(subDomain, domain), "Name::IsSubDomainOf() failed");
        
        subDomain = "my-service._ipps._tcp.local.";
        domain    = "local";
        VerifyOrQuit(Dns::Name::IsSubDomainOf(subDomain, domain), "Name::IsSubDomainOf() failed");

        subDomain = "my-service._ipps._tcp.local";
        domain    = "local";
        VerifyOrQuit(Dns::Name::IsSubDomainOf(subDomain, domain), "Name::IsSubDomainOf() failed");
```

Test different combination of names that end with dot or not?

simonlingoogle (approved):
LGTM  :+1: 

with one suggestion on test cases.

tests/unit/test_dns.cpp
wgtdkp: updated.

src/core/net/srp_server.cpp
```diff
@@ -635,6 +637,7 @@ otError Server::ProcessHostDescriptionInstruction(Host &                   aHost
         Dns::ResourceRecord record;
 
         SuccessOrExit(error = Dns::Name::ReadName(aMessage, aOffset, name, sizeof(name)));
+        VerifyOrExit(Dns::Name::IsSubDomainOf(name, GetDomain()), error = OT_ERROR_PARSE);
```
jwhui, line 640: Is `OT_ERROR_PARSE` what we should indicate here? Typically `OT_ERROR_PARSE` is for malformed messages.

wgtdkp: Agree that `OT_ERROR_PARSE` is not a proper error code. The DNS response code for domain not matching is `kResponseNotZone` and `kResponseNotAuth`:
https://github.com/openthread/openthread/blob/c03458433ec188be21027d6fa0eab01d3dbf6c03/src/core/net/dns_headers.hpp#L267-L268

But we currently don't have an OT error code that is mapping to `kResponseNotZone`. For now, I think `OT_ERROR_SECURITY` may be a better error code as it will be converted to `kResponseRefused`:
https://github.com/openthread/openthread/blob/c03458433ec188be21027d6fa0eab01d3dbf6c03/src/core/net/srp_server.cpp#L71-L74
https://github.com/openthread/openthread/blob/c03458433ec188be21027d6fa0eab01d3dbf6c03/src/core/net/dns_headers.hpp#L263

One thing to notice is that the SRP client converts `kResponseNotZone` to `OT_ERROR_PARSE`, which I think is not perfect either.
https://github.com/openthread/openthread/blob/c03458433ec188be21027d6fa0eab01d3dbf6c03/include/openthread/srp_client.h#L173

For the server, one solution is that we can have those DNS processing method return `Dns::Response` but not OT error code. For the client, we can have it return both OT error code (for request and response processing errors, e.g. `OT_ERROR_TIMEOUT` and DNS response code (clearly indicates the error happens on the server side). thoughts? @jwhui @abtink

simonlingoogle: That is the reason I think converting errors for RCODRs is not a scalable idea. 
In DNSSD server, I choose to set the RCODE directly without returning any error. The server only return error in a server failure case, e.g. no more buffers.

jwhui (approved):

PR #6116, opened 2021-01-26
[dns-client] service discovery (DNS-SD) support and enhancements

abtink:
This commit re-designs the `Dns::Client` module enhancing the address
resolution implementation and also adding support for DNS-Based
Service Discovery (DNS-SD).

With regards to address resolution query the new model relaxes the
requirements of the public OT APIs such that caller does not need to
persist the query info (e.g., the host name string buffer can be a
temporary variable and does not need to persist during the query) and
it is all managed by the `Dns::Client` core implementation itself. The
new model also supports the case where the response contains multiple
IPv6 addresses providing new APIs to allow the user to iterate through
the list of addresses and retrieve them one by one.

The implementation also handles the case where the DNS query response
contains CNAME record mapping the queried host name to a canonical
name for which a list of addresses are then provided.

The core implementation is also simplified, instead of cloning a query
message and saving it for possible retx, the new code saves the query
related info from which it can re-construct the query message for retx
if/when needed.

This commit also adds support for DNS-SD in `Dns::Client`. The config
`OPENTHREAD_CONFIG_DNS_CLIENT_SERVICE_DISCOVERY_ENABLE` can be used to
disable service discovery feature. The implementation supports
"service instance enumeration" which is referred to as "browsing" and
"service instance resolution". Callbacks are used to notify the user
when a response is received. In the callback a pointer to an opaque
response object is given to the user which can then be used with the
new set of APIs to get more info about the response, such as the list
of discovered service instances or more details about a specific
service instance (e.g., port number, host name and its address) or
iterate through address lists. This provides a flexible and scalable
solution to handle larger lists in the response without adding memory
overhead in the implementation.


abtink:
Reviewing/Reading the `dns_clinet.hpp/cpp`, I suggest not checking the `diff` and reading the new implementation directly. Thanks.

simonlingoogle:
@abtink 
Add tests for Service Discovery features to increase code coverage?

abtink:
In the latest push, the changes are rebased over PR #6124  which is used to simplify the `Dns::Client` implantation 
Please read and review the last commit only.

abtink:
Latest push adds a new commit adding support for DNS-SD in CLI.

Regarding testing:
- The `Clinet` code itself and CLI implementation are verified in a custom test build branch (where the test code generates a response to query for browse). The test branch is available here: [dns/new-model-test on abtink/openthrad]( https://github.com/abtink/openthread/tree/dns/new-model-test).
- Future enehancmenet PR should later add CI tests covering both server from #6103 and client side behaviors

src/cli/README.md
```diff
@@ -758,6 +758,30 @@ Send DNS Query to obtain IPv6 address for given hostname. The latter two paramet
...
+Send a browse (service instance enumeration) DNS query to get the list of services for given service-name.
+
+```bash
+> dns browse _service._udp.example.com
+DNS browse response for _service._udp.example.com.
+inst1
+    Port:1234, Priority:1, Weight:2, TTL:7200
+    Host:host.example.com.
+    HostAddress:fd00:0:0:0:0:0:0:abcd TTL:7200
+    TXT-Data:(len:11) [096b65793d76616c756500] TTL:7300
```
wgtdkp, line 772: We added methods to parse the TXT records into `otDnsTxtEntry` in https://github.com/openthread/openthread/pull/6113, it would be better to have parsed TXT entries in the CLI.

src/cli/cli.cpp
```diff
@@ -1337,48 +1330,70 @@ otError Interpreter::ProcessDiscover(uint8_t aArgsLength, char *aArgs[])
...
-    Ip6::MessageInfo messageInfo;
-    otDnsQuery       query;
 
-    VerifyOrExit(aArgsLength > 0, error = OT_ERROR_INVALID_ARGS);
+otError Interpreter::GetDnsServerAddress(uint8_t     aArgsLength,
+                                         char *      aArgs[],
+                                         otSockAddr &aAddress,
+                                         uint8_t     aStartArgsIndex)
+{
+    // This method gets the optional the server address from given
```
wgtdkp, line 1339:
```suggestion
    // This method gets the optional server address from given
```

src/core/net/dns_client.cpp
```diff
@@ -43,357 +43,778 @@
...
+    SuccessOrExit(error = mSocket.Bind());
 
 exit:
-    FreeAndNullMessageOnError(messageCopy, error);
-    return messageCopy;
+    return error;
 }
 
-void Client::DequeueMessage(Message &aMessage)
+otError Client::Stop(void)
```
wgtdkp, line 408: nit: we should always be able to stop the client?
```suggestion
void Client::Stop(void)
```

src/core/net/dns_client.cpp
```diff
@@ -43,357 +43,778 @@
...
+    }
+
+    header.SetQuestionCount(kQuestionCount[aInfo.mQueryType]);
+
+    message = mSocket.NewMessage(0);
+    VerifyOrExit(message != nullptr, error = OT_ERROR_NO_BUFS);
+
+    SuccessOrExit(error = message->Append(header));
+
+    // Prepare the question section
```
wgtdkp, line 579:
```suggestion
    // Prepare the question section.
```

src/core/net/dns_client.hpp
```diff
@@ -45,64 +46,344 @@
  *   This file includes definitions for the DNS client.
  */
 
+/**
+ * This struct represents an opaque (and empty) type for a response to a address resolution DNS query.
```
wgtdkp, line 50:
```suggestion
 * This struct represents an opaque (and empty) type for a response to an address resolution DNS query.
```

include/openthread/dns_client.h
```diff
@@ -0,0 +1,470 @@
...
+ */
+typedef struct otDnsServiceInfo
+{
+    uint32_t     mTtl;                ///< Service record TTL (in seconds).
+    uint16_t     mPort;               ///< Service port number.
+    uint16_t     mPriority;           ///< Service priority.
+    uint16_t     mWeight;             ///< Service weight.
+    char *       mHostNameBuffer;     ///< Buffer to output the service host name (can be NULL if not needed).
+    uint16_t     mHostNameBufferSize; ///< Size of `mHostNameBuffer`.
+    otIp6Address mHostAddress;        ///< The host IPv6 address. Set to all zero if not available.
```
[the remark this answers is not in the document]
abtink: Added this to make it easier for user's of the API, so that they can get all service info in one call.

In particular, this allows the case where the user may not be interested in getting the host name and just want the IPv6 address of service instance. In this case, use can set `mHostNameBuffer` to `NULL` (no need for caller to allocate and a have a buffer to retrieve the host name).

`otDnsBrowseResponseGetHostAddress()` is there so we can handle the rare case where we may get multiple address (AAAA records) for the same host name.

include/openthread/dns_client.h
```diff
@@ -0,0 +1,470 @@
...
+ * @param[out] aLabelBuffer       A buffer to char array to output the service instance label (MUST NOT be NULL).
+ * @param[in]  aLabelBufferSize   The size of @p aLabelBuffer.
+ *
+ * @retval OT_ERROR_NONE          The service instance was read successfully.
+ * @retval OT_ERROR_NO_BUFS       The name does not fit in @p aNameBuffer.
+ * @retval OT_ERROR_NOT_FOUND     No service instance record in @p aResponse at @p aIndex.
+ * @retval OT_ERROR_PARSE         Could not parse the records in the @p aResponse.
+ *
+ */
+otError otDnsBrowseResponseGetServiceInstance(const otDnsBrowseResponse *aResponse,
```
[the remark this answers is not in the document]
abtink: I originally had it `GetInstanceLabel()` (I did not want to use a `Name` since we are getting the label only and not full name), later I changed it to `GetServiceInstance()`.

Still prefer not using `Name` here.

Thoughts?

src/cli/README.md
abtink: I actually wanted to use it, but noticed that the main public API is `otSrpServerServiceGetNextTxtEntry()` and is under `SRP_SERVER` config. The DNS client may be enabled/used without SRP server feature, so we cannot use this API.

I think later (in a new PR) we should move the TXT record parsing API under common `dns.h` and the implementation in `dns_header.cpp` so we can use it from all modules (independent of which features are enabled/disabled).

include/openthread/dns_client.h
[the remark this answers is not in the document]
wgtdkp: Makes sense to me to avoid confusing with full names. Thanks for sharing the thoughts!

src/cli/README.md
wgtdkp: `otSrpServerServiceGetNextTxtEntry` is a simple wrapper of `Dns::TxtRecord::GetNextTxtEntry`. I think we can also have an DNS client API like `otDnsBrowseResponseGetNextTxtNetry` which is also simply a wrapper of `Dns::TxtRecord::GetNextTxtEntry`. Thoughts?

abtink: A while after I added the comment, I actually also thought of the same model as you suggested 😃 (i.e., if we can add a `otDnsBrowseResponseGetNextTxtEntry` for `BrowseResponse` itself using `TxtRecord::GetNextEntry()`)

I don't think that it will be straight-forward:
- The `GetNextEntry()` requires `aTxtData` array to be given to it so it can parse through the entries. 
- In the case of `otDnsBrowseResponse`, we won't have the txt data in an array but we will have it in the response `Message`. 
- I don't see any easy way to extend `GetNextEntry()` to work with `Message` directly (since we use pointers to `aTxtData` array to get the "key" and "value" segment - which is a good model in my opinion).

The way I was/am considering is that we add a new common `ot` public function (in `dns.h`) like `otDnsGetNextTxtEntry(otDnsTxtEntry *aIterator, const char *aTxtData, uint16_t aTxtDataLength)` to parse any given txt data. This then makes it the caller's job to provide the txt data buffer and its content. The other upside I see is that the same `ot` API can be used for different cases (e.g., browse response or service response) and even maybe for SRP server services.

Thoughts? Other ideas?

wgtdkp: oh, yes we are retrieving data directly from the response message, agree that there is no easy way to have `otDnsBrowseResponseGetNextTxtEntry` and a public API `otDnsGetNextTxtEntry` should be what we need :)

abtink: BTW, @wgtdkp just letting you know so we don't both do the same thing. 😄 

I already started adding this (i.e. a common txt record parse api) - will also add a bunch of unit tests for the whole txt entry encoding/decoding in the same PR.

wgtdkp: @abtink Thanks! :+1: 
Looking forward to your PR!

src/core/net/dns_client.hpp
```diff
@@ -45,64 +46,344 @@
...
+         * This method gets info for a service instance from a DNS browse (service instance enumeration) response.
+         *
+         * This method MUST only be used from `BrowseCallback`.
+         *
+         * A browse DNS response should include the SRV, TXT, and AAAA records for the service instances that are
+         * enumerated (note that it is a SHOULD and not a MUST requirement). This method tries to retrieve this info
+         * for a given service instance.
+         *
+         * If no matching SRV record is found, `OT_ERROR_NOT_FOUND` is returned.
+         * If a matching SRV record is found, @p aServiceInfo is updated returning  `OT_ERROR_NONE`.
```
jwhui, line 264:
```suggestion
         * If a matching SRV record is found, @p aServiceInfo is updated returning `OT_ERROR_NONE`.
```

src/core/net/dns_client.cpp
```diff
@@ -43,357 +43,778 @@
...
+{
+    static_assert(kAddressQuery == 0, "kAddressQuery value is not correct");
+#if OPENTHREAD_CONFIG_DNS_CLIENT_SERVICE_DISCOVERY_ENABLE
+    static_assert(kBrowseQuery == 1, "kBrowseQuery value is not correct");
+    static_assert(kServiceQuery == 2, "kServiceQuery value is not correct");
+#endif
+}
 
-    mRetransmissionTimer.FireAtIfEarlier(aQueryMetadata.mTransmissionTime);
+otError Client::Start(void)
```
wgtdkp, line 397: Instead of having `aServerAddr` for each public query method, it should be better to specify the server address with the `Start()` method. And it will ease adding auto-selection of DNS-SD servers which are found in Network Data.

We can leave it to a separate PR which will add auto-selection of the DNS-SD server. Thoughts?

wgtdkp (approved):
Thanks @abtink  overall looks good to me :+1: 

Just two small nits.

src/core/net/dns_client.cpp
abtink: Let's consider this for a future/separate PR. 

- On one hand, I like the idea of being able to specify the server address for each query (e.g., address resolution query may want to use a different server than DNS-SD?). 
- On other hand, I see the benefit of being able to set the server address once and then use it. 

Maybe we can have a model where we have an API to set a "default server sock address" and if the caller from query API does not pass an address, we fallback and use the default one.

abtink:
Marking this as "work-in-progress".  Noticed after rebase on DNS-SD server code, the new `test_dnssd.py` seems to be failing. I think it may be due to the change to report multiple addresses in response to an address resolution in the CLI `dns resolve` command and how the CLI response is being parsed.

src/core/net/dns_client.cpp
```diff
@@ -43,357 +43,777 @@
...
 
-        if (record.Matches(ResourceRecord::kTypeAaaa))
-        {
-            // Return the first found IPv6 address.
-            FinalizeDnsTransaction(*message, queryMetadata, &record.GetAddress(), record.GetTtl(), OT_ERROR_NONE);
-            ExitNow(error = OT_ERROR_NONE);
-        }
+    // Check the Question Section
+
+    VerifyOrExit(header.GetQuestionCount() == kQuestionCount[aType], error = OT_ERROR_PARSE);
```
simonlingoogle, line 744: Is this check too strict? We have seen DNS servers that do not echo back any question. 

Besides that, I think a good benefit of not echoing back questions in Thread networks is reducing packet size.

simonlingoogle (approved):
LGTM 👍 

with suggestions (even accepted, can be resolved in future PRs)

src/core/net/dns_client.cpp
```diff
@@ -43,357 +43,777 @@
...
+    /* (2) kServiceQuery -> */ kServiceQueryRecordTypes,
+#endif
+};
 
-    SuccessOrExit(error = aQueryMetadata.AppendTo(*messageCopy));
-    mPendingQueries.Enqueue(*messageCopy);
+Client::Client(Instance &aInstance)
+    : InstanceLocator(aInstance)
+    , mSocket(aInstance)
+    , mTimer(aInstance, Client::HandleTimer)
```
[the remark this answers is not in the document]
abtink: Using a `Timer` makes more sense and is easier here. The code is simpler with `Timer` and it allow for any retx delay value (not just factors of second).

src/core/net/dns_client.cpp
```diff
@@ -43,357 +43,777 @@
...
+    }
+    else
+    {
+        SuccessOrExit(error = Name::ParseName(*mMessage, offset));
+    }
+
+    // Search in additional section for AAAA record for the host name.
 
-    VerifyOrExit(aQuery.IsValid(), error = OT_ERROR_INVALID_ARGS);
+    error = FindHostAddress(kAdditionalDataSection, hostName, /* aIndex */ 0,
```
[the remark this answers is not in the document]
abtink: I think if the server does this, it is not following the spec.

Here is the text from [RFC6763 section-12](https://tools.ietf.org/html/rfc6763#section-12) related to this:
> 
>    DNS has an efficiency feature whereby a DNS server may place
>    additional records in the additional section of the DNS message.
>    These additional records are records that the client did not
>    explicitly request, but the server has reasonable grounds to expect
>    that the client might request them shortly, so including them can
>    save the client from having to issue additional queries.
> 
>    This section recommends which additional records SHOULD be generated
>    to improve network efficiency, for both Unicast and Multicast DNS-SD
>    responses.

----

> DNS client can treat RRs in answer section and additional data section equally? This might even simplify the parsing logic (e.g. no need to select sections).

It would be possible and relatively easy if we allow search to happen in **any** of sections "answer", "authority" and "additional data"(basically over all RRs in the response).

But if we want to just treat "answer" and "additional data" equally (and exclude the "authority"), I don't think it would easy or simple. Note that we have helper methods that search within a section for `index`th occurrence of a record (matching a name and/or record type). If we need to do such iterations and search over two sections (which may not be consecutive in terms of how RRs are encoded in the response message), we need to update the helper methods to enable such behavior and add more complex logic/code.

----

For now, I prefer/suggest to keep the logic as is and follow the RFC spec.

src/cli/README.md
abtink: We have this now in PR #6157.

src/cli/README.md
```diff
@@ -758,6 +758,30 @@ Send DNS Query to obtain IPv6 address for given hostname. The latter two paramet
 > DNS response for ipv6.google.com - 2a00:1450:401b:801:0:0:0:200e TTL: 300
 ```
 
+### dns browse \<service-name\> \[DNS server IP\] \[DNS server port\]
```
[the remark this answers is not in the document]
abtink: @wgtdkp suggested the same in above comment.

We can take care of this in a next PR. 

I mentioned the idea of adding default server from https://github.com/openthread/openthread/pull/6116#discussion_r573462074. There are some other things I am thinkging to add along with this later (e.g., allow caller/user to set the number of retries and/or retx delay, recursive query flag, etc.)

tests/scripts/thread-cert/node.py
```diff
@@ -2411,10 +2411,11 @@ def dns_resolve(self, hostname, server=None, port=53):
...
         dns_resp = output[0]
-        # example output: DNS response for host1.default.service.arpa. - fd00:db8:0:0:ae43:4938:4c42:e6af TTL: 7190
-        ip, ttl = dns_resp.split(' - ')[1].split(' TTL: ')
-        ip = ip.strip()
-        ttl = int(ttl)
+        # example output: "DNS response for host1.default.service.arpa. - fd00:db8:0:0:fd3d:d471:1e8c:b60 TTL:7190 "
+        #                 " fd00:db8:0:0:0:ff:fe00:9000 TTL:7190"
+        addrs = dns_resp.strip().split(' - ')[1].split(' ')
+        ip = [item.strip() for item in addrs[::2]]
+        ttl = [int(item.split('TTL:')[1]) for item in addrs[1::2]]
```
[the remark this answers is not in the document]
abtink: Good suggestion. Thanks

src/core/net/dns_client.cpp
abtink: I am fine relaxing this check if it is valid/permitted by spec for server to not include question the response.

I tried doing a quick search for this in specs/rfcs, but could not find anything explicit about this. @simonlingoogle, do you happen to know if this is mentioned in any rfc.

I will relax the check in the code for now.

jwhui: RFC 1035 Section 7.3 has the following statement:
```
The next step is to match the response to a current resolver request.
The recommended strategy is to do a preliminary matching using the ID
field in the domain header, and then to verify that the question section
corresponds to the information currently desired.
```
This would imply to me that the resolver expects the question section to include the question that the name server is answering. However, I admit that I'm not sure why the question section is necessary in a response, given that there is an ID field.

@simonlingoogle , can you reference existing implementations that do not include the question in the response? If there is sufficient data that deployed implementations exclude the question, then I think it is something we can explore.

simonlingoogle: I think what I saw is mDNS responses which "MUST NOT contain any questions in the Question Section". So, unicast responses should still contain questions.

abtink: I did actually relax this check in current PR. Let me change that back.

jwhui (approved):

PR #6117, opened 2021-01-27
[dns-name] allow 'Dns::Name' to be given as string or from message

abtink:
This commit allows `Dns::Name` instances to be created which can be
empty or specified from a string (a dot '.' separated sequence of labels) or
from a message at a given offset (i.e., name is already encoded in a
message). `CompareName` and `FindRecord()` helper methods are updated
to allow the new `Dns::Name` type to be used when comparing/searching
for a name.

--------

_Background for this_: During DNS-SD client implementation I noticed this model of having 
a common representation of a `Dns::Name` that "may be a string" or maybe "already encoded
in a message" can be very useful. I expect it to help simplify support for handling of `CNAME`
records in a received response message.

### DIFF
--- a/src/core/net/dns_headers.hpp
+++ b/src/core/net/dns_headers.hpp
@@ -480,10 +480,10 @@ public:
 } OT_TOOL_PACKED_END;
 
 /**
- * This class implement helper methods for encoding/decoding of DNS Names.
+ * This class represents a DNS name and implements helper methods for encoding/decoding of DNS Names.
  *
  */
-class Name
+class Name : public Clearable<Name>
 {
 public:
     enum : uint8_t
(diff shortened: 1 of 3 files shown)


src/core/net/dns_headers.hpp
```diff
@@ -493,6 +493,138 @@ class Name
...
+    {
+        mString  = aString;
+        mMessage = nullptr;
+    }
+
+    /**
+     * This method sets the name from a message at a given offset.
+     *
+     * @param[in] aMessage   The message containing the encoded name. `aMessage.GetOffset()` MUST point to the start of
+     *                       the DNS header in the message (used to parse compressed name)..
```
librasungirl, line 589:
```suggestion
     *                       the DNS header in the message (used to parse compressed name).
```

src/core/net/dns_headers.hpp
```diff
@@ -493,6 +493,138 @@ class Name
...
+    {
+        mString  = nullptr;
+        mMessage = &aMessage;
+        mOffset  = aOffset;
+    }
+
+    /**
+     * This method gets the name as string.
+     *
+     * This method MUST be used only when the type if `kTypeString` and `IsFromString()` returns TRUE. Otherwise its
```
librasungirl, line 603:
```suggestion
     * This method MUST be used only when the type is `kTypeString` and `IsFromString()` returns TRUE. Otherwise its
```

__the type is `kTypeString`__  and __`IsFromString()` returns TRUE__ seem equivalent, is it necessary to repeat here?

src/core/net/dns_headers.hpp
```diff
@@ -493,6 +493,138 @@ class Name
...
+     *
+     * @returns A pointer to string.
+     *
+     */
+    const char *GetAsString(void) const { return mString; }
+
+    /**
+     * This method gets the name message and offset.
+     *
+     * This method MUST be used only when the type if `kTypeMessage` and `IsFromMessage()` returns TRUE. Otherwise its
```
librasungirl, line 614:
```suggestion
     * This method MUST be used only when the type is `kTypeMessage` and `IsFromMessage()` returns TRUE. Otherwise its
```

src/core/net/dns_headers.hpp
```diff
@@ -493,6 +493,138 @@ class Name
...
+     * behavior is undefined.
+     *
+     * @param[out]  aOffset    A reference to variable to out the offset of start of the name in the message.
```
librasungirl, line 617:
```suggestion
     * @param[out]  aOffset    A reference to a variable to out the offset of start of the name in the message.
```

librasungirl (approved):

src/core/net/dns_headers.hpp
```diff
@@ -493,6 +493,136 @@ class Name
...
     };
 
+    /**
+     * This enumeration represent the name type.
+     *
+     */
+    enum Type : uint8_t
+    {
+        kTypeEmpty,   ///< The name is empty (not specified).
+        kTypeString,  ///< The name is given as a string (dot '.' separated sequence of labels').
```
simonlingoogle, line 503:
```suggestion
        kTypeCString,  ///< The name is given as a string (dot '.' separated sequence of labels').
```

src/core/net/dns_headers.hpp
```diff
@@ -493,6 +493,136 @@ class Name
...
+        kTypeMessage, ///< The name is specified from a message at a given offset (encoded in the message).
```
simonlingoogle, line 504:
```suggestion
        kTypeEncoded, ///< The name is specified from a message at a given offset (encoded in the message).
```

src/core/net/dns_headers.hpp
```diff
@@ -788,9 +938,18 @@ class Name
...
+    Name(const char *aString, const Message *aMessage, uint16_t aOffset)
+        : mString(aString)
+        , mMessage(aMessage)
+        , mOffset(aOffset)
+    {
+    }
+
+    const char *   mString;  // String containing the name or `nullptr` if name is not from string.
+    const Message *mMessage; // Message containing the encoded name, or `nullptr` if `Name` is not from message.
+    uint16_t       mOffset;  // Offset in `mMessage` to the start of name (used when name is from `mMessage`).
```
simonlingoogle, line 952:
```suggestion
   union {
      const char *   mString;  // String containing the name or `nullptr` if name is not from string.     
      struct {
        const Message *mMessage; // Message containing the encoded name, or `nullptr` if `Name` is not from message.
        uint16_t       mOffset;  // Offset in `mMessage` to the start of name (used when name is from `mMessage`).
      } 
   }
```

Use union to reduce memory? We can use `mOffset` to determine whether it's a string (mOffset == 0) or not (mOffset > 0) because `mOffset` should never be zero when it's encoded.

simonlingoogle (approved):
LGTM :+1: 

with one small suggestion.

src/core/net/dns_headers.hpp
abtink: I prefer `kTypeMessage` over `kTypeEncoded` sicne it indicating that name is present in a message. `Encoded` may mean that it's is a buffer/array (encoded).

src/core/net/dns_headers.hpp
abtink: The idea of using `mOffset` of zero to indicate the type is interesting and smart, but I am not sure if we should go for it since I think it may be limiting. It's correct that in a DNS message the offset should never be zero, but we may use a `Message` as a data blob storage, for example to store metadata about query (we do this actually in the new DNS client) and in that case it can happen that a name is encoded at any offset (maybe at offset zero).

-------

On using `union`, I actually started with that. Basically I had a `mType` and then a `union` but decided against it.

- On the upside, I agree it can reduce the `sizeof(Name)` by few bytes. But I think in all cases, the instances of `Name` would be temporary/local vairable and the size won't matter much.
- On the downside, it adds a code size and computation overhead (since now we need to check `mType` against set of values). 

Not using `union` has the advantage that complier/toolchain can optimize/inline the code better:
- Compliers typically generate very efficient code fo comparing a pointer/value against zero (this can end up being combined into the instruction that loads the value from memory to a register (same instruction does zero comparison as well).
- Most advanced toolchain can optimize simple struct instance when passing them to methods/function and use  registers to pass the fields rather than allocating stack space to generate the object. `union` won't be optimized like this.

src/core/net/dns_headers.hpp
```diff
@@ -493,6 +493,136 @@ class Name
         kMaxEncodedLength = 255, ///< Max length of an encoded name.
     };
 
+    /**
+     * This enumeration represent the name type.
```
jwhui, line 497:
```suggestion
     * This enumeration represents the name type.
```

src/core/net/dns_headers.hpp
```diff
@@ -493,6 +493,136 @@ class Name
...
+     */
+    Name(void)
+        : Name(nullptr, nullptr, 0)
+    {
+    }
+
+    /**
+     * This constructor initializes the `Name` object with a given string.
+     *
+     * @param[in] aString   A string specifying the name (dot '.' separated sequence of labels').
```
jwhui, line 519:
```suggestion
     * @param[in] aString   A string specifying the name (dot '.' separated sequence of labels).
```

src/core/net/dns_headers.hpp
```diff
@@ -493,6 +493,136 @@ class Name
...
+     *
+     * @returns The name type.
+     *
+     */
+    Type GetType(void) const { return IsFromCString() ? kTypeCString : (IsFromMessage() ? kTypeMessage : kTypeEmpty); }
+
+    /**
+     * This method sets the name from a given string.
+     *
+     * @param[in] aString   A string specifying the name (dot '.' separated sequence of labels').
```
jwhui, line 576:
```suggestion
     * @param[in] aString   A string specifying the name (dot '.' separated sequence of labels).
```

src/core/net/dns_headers.hpp
jwhui: This appears unresolved :)

src/core/net/dns_headers.hpp
```diff
@@ -493,6 +493,136 @@ class Name
...
+     */
+    void SetFromMessage(const Message &aMessage, uint16_t aOffset)
+    {
+        mString  = nullptr;
+        mMessage = &aMessage;
+        mOffset  = aOffset;
+    }
+
+    /**
+     * This method gets the name as string.
```
jwhui, line 601:
```suggestion
     * This method gets the name as a C string.
```

src/core/net/dns_headers.hpp
```diff
@@ -493,6 +493,136 @@ class Name
...
+     *
+     * This method MUST be used only when the type is `kTypeString`. Otherwise its behavior is undefined.
+     *
+     * @returns A pointer to string.
```
jwhui, line 605:
```suggestion
     * @returns A pointer to a C string.
```

src/core/net/dns_headers.hpp
```diff
@@ -493,6 +493,136 @@ class Name
...
+    bool IsFromMessage(void) const { return mMessage != nullptr; }
+
+    /**
+     * This method gets the type of `Name` object indicating whether it is empty, specified by a C string or from a
+     * message
+     *
+     * @returns The name type.
+     *
+     */
+    Type GetType(void) const { return IsFromCString() ? kTypeCString : (IsFromMessage() ? kTypeMessage : kTypeEmpty); }
```
jwhui, line 571: nit: should we be more specific about what kind of type?
```suggestion
    Type GetFromType(void) const { return IsFromCString() ? kTypeCString : (IsFromMessage() ? kTypeMessage : kTypeEmpty); }
```

jwhui (approved):

src/core/net/dns_headers.hpp
simonlingoogle: I am fine to keep it as is since Name is mostly allocated on Heap.

PR #6118, opened 2021-01-27
[child-table] enhance address iterator method

Irving-cl:
It seems that moving the if statement out of the loop always brings the same result. So I think we can move it outside the loop.


simonlingoogle (approved):
LGTM :+1: 

this is slightly better (-16B code)

abtink (approved):
LGTM. Thanks.

jwhui (approved):

PR #6119, opened 2021-01-27
[tests] use recommended network forming method in tests

ColdDivinity:
This makes tests use the recommended method to form networks.

Depends-On: openthread/pyspinel#110


jwhui:
Superseded by #9215

PR #6120, opened 2021-01-27
[low-power] support indirect TX for MTD

Irving-cl:
This PR is the updated version of #6095.

As discussed in the thread, instead of adding a new class `SedCapableNeighbor`, we directly reuse `Child` class for MTD to do indirect transmission. An abstraction of this PR:
* Update `Child` class so that:
    * On FTD, nothing changes.
    * On MTD, it has the data members and methods for indirect transmission and disposes other stuffs.
* Update the macro wrappers (`OPENTHREAD_FTD`) for indirect transmission modules so that these modules would be built not only for FTD, but also for MTD with S2S feature.
    * Add an alias (`OPENTHREAD_MTD_S2S`) as the macro wrappper. This equals to (`OPENTHREAD_MTD && OPENTHREAD_CONFIG_MAC_SED_TO_SED_ENABLE`). The code would be messy if we put full expansion of this alias everywhere.
 * Add logic for indirect transmission in `MeshForwarder(MTD)::SendMessage`.
 * Add a config option `OPENTHREAD_CONFIG_MAC_SED_TO_SED_ENABLE`. On MTD, when this is enabled, the reduced `Child` and indirect transmission modules would be built. By default, this option is disabled.
 * Add a new version of `FindChild` which accepts an Ip6 address in `ChildTable` and use this method in `MeshForwarder::SendMessage`.


src/core/thread/child_table.hpp
```diff
@@ -36,15 +36,16 @@
...
-
 #include "common/iterator_utils.hpp"
 #include "common/locator.hpp"
 #include "common/non_copyable.hpp"
+#include "thread/s2s.hpp"
 #include "thread/topology.hpp"
 
 namespace ot {
 
+#if OPENTHREAD_FTD || OPENTHREAD_MTD_S2S
```
brianljt, line 47: Same here, put this line to line 39? We don't want to include these header files if it is not neccessary.

src/core/thread/csl_tx_scheduler.cpp
```diff
@@ -28,7 +28,7 @@
 
 #include "csl_tx_scheduler.hpp"
 
-#if !OPENTHREAD_MTD && OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
+#if (OPENTHREAD_FTD || OPENTHREAD_MTD_S2S) && OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
```
brianljt, line 31:
```suggestion
#if OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
```

src/core/thread/csl_tx_scheduler.hpp
```diff
@@ -50,7 +51,7 @@ namespace ot {
  * @{
  */
 
-#if !OPENTHREAD_MTD && OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
+#if (OPENTHREAD_FTD || OPENTHREAD_MTD_S2S) && OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
```
brianljt, line 54:
```suggestion
#if OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
```

src/core/thread/indirect_sender.cpp
```diff
@@ -163,14 +163,15 @@ void IndirectSender::ClearAllMessagesForSleepyChild(Child &aChild)
...
     mCslTxScheduler.Update();
 #endif
 
 exit:
     return;
 }
 
+#if OPENTHREAD_FTD
 void IndirectSender::SetChildUseShortAddress(Child &aChild, bool aUseShortAddress)
 {
```
brianljt, line 176:
```suggestion
void IndirectSender::SetChildUseShortAddress(Child &aChild, bool aUseShortAddress)
{
#if OPENTHREAD_FTD
```

How about making `SetChildUseShortAddress` an empty function when `OPENTHREAD_FTD` is not defined and does not use macro where `SetChildUseShortAddress` is called?

src/core/thread/indirect_sender.cpp
```diff
@@ -180,13 +181,16 @@ void IndirectSender::SetChildUseShortAddress(Child &aChild, bool aUseShortAddres
 exit:
     return;
 }
+#endif
 
 void IndirectSender::HandleChildModeChange(Child &aChild, Mle::DeviceMode aOldMode)
 {
+#if OPENTHREAD_FTD
```
brianljt, line 188:
```suggestion
```
Remove this macro where it is called

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -48,7 +48,7 @@ otError MeshForwarder::SendMessage(Message &aMessage)
 {
     Mle::MleRouter &mle   = Get<Mle::MleRouter>();
     otError         error = OT_ERROR_NONE;
-    Neighbor *      neighbor;
+    Child *         rx_off_child;
```
brianljt, line 51:
```suggestion
    Child *         rxOffChild;
```

src/core/thread/mesh_forwarder_mtd.cpp
```diff
@@ -35,15 +35,92 @@
...
 
+#include "common/locator-getters.hpp"
+
 namespace ot {
 
 otError MeshForwarder::SendMessage(Message &aMessage)
 {
+#if OPENTHREAD_MTD_S2S
+    Mle::MleRouter &mle = Get<Mle::MleRouter>();
+    Child *         rx_off_child;
```
brianljt, line 46:
```suggestion
    Child *         rxOffChild;
```

src/core/thread/mesh_forwarder_mtd.cpp
```diff
@@ -35,15 +35,92 @@
 
 #include "mesh_forwarder.hpp"
 
+#include "common/locator-getters.hpp"
+
 namespace ot {
 
 otError MeshForwarder::SendMessage(Message &aMessage)
```
brianljt, line 42: Almost same function with the one in mesh_forwarder_ftd.cpp. Is it doable to put this function to mesh_forwarder.cpp to avoid duplication?

src/core/thread/s2s.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include "openthread-core-config.h"
+
+/**
+ * @def OPENTHREAD_MTD_S2S
+ *
+ * This is a alias representing if `OPENTHREAD_MTD` and `OPENTHREAD_CONFIG_MAC_SED_TO_SED_ENABLE` are both enabled.
+ *
+ */
+#define OPENTHREAD_MTD_S2S OPENTHREAD_MTD &&OPENTHREAD_CONFIG_MAC_SED_TO_SED_ENABLE
```
brianljt, line 48:
```suggestion
#define OPENTHREAD_MTD_S2S OPENTHREAD_MTD && OPENTHREAD_CONFIG_MAC_SED_TO_SED_ENABLE
```

brianljt (commented):
Nice job. Look good to me in general. Some nits.

src/core/mac/data_poll_handler.cpp
```diff
@@ -31,17 +31,18 @@
...
 
 #include "common/code_utils.hpp"
 #include "common/instance.hpp"
 #include "common/locator-getters.hpp"
 #include "common/logging.hpp"
+#include "thread/s2s.hpp"
 
 namespace ot {
 
+#if OPENTHREAD_FTD || OPENTHREAD_MTD_S2S
```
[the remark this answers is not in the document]
Irving-cl: Note that `OPENTHREAD_MTD_S2S` is defined in `thread/s2s.hpp` as a alias. (Otherwise we need to write `OPENTHREAD_MTD && OPENTHREAD_CONFIG_MAC_SED_TO_SED_ENABLE` everywhere). Unlike other core config options, `OPENTHREAD_MTD_S2S` must be used after including `thread/s2s.hpp`. Thoughts?

src/core/mac/data_poll_handler.hpp
```diff
@@ -90,7 +91,7 @@ class DataPollHandler : public InstanceLocator, private NonCopyable
     class ChildInfo
     {
         friend class DataPollHandler;
-#if !OPENTHREAD_MTD && OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
+#if (OPENTHREAD_FTD || OPENTHREAD_MTD_S2S) && OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
```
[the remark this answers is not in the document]
Irving-cl: I'm considersing such a case: building MTD with S2S disabled, however CSL_TRASNMITTER is enabled accidentally (for example, ftd built together with mtd). In this case we should also disable this feature. So I think we need check like this. Thoughts?

src/core/thread/indirect_sender.cpp
Irving-cl: Well, I would prefer to not define it at all in MTD case for: 1. smaller code size 2. more efficient And I think this helps us to differentiate Child for normal case and for s2s case.

src/core/thread/indirect_sender.cpp
Irving-cl: Some explanation here: I directly reuse the code from `IndirectSender` and enable these code when `OPENTHREAD_FTD` **or** `OPENTHREAD_MTD_S2S` is enabled. However, unfortunately, there's still a little part of code that only makes sense for `FTD` in `IndirectSender`. In this `HandleChildModeChange` method, the short address logic is only for Child and doesn't make sense for the 'Peer' in S2S case.

src/core/mac/data_poll_handler.cpp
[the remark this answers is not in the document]
brianljt: Got it. Thanks!

src/core/mac/data_poll_handler.hpp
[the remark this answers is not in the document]
brianljt: OK. Then it is fine for me.

brianljt (approved):
LGTM 👍 Thanks!

src/core/config/mac.h
```diff
@@ -417,4 +417,18 @@
 #define OPENTHREAD_CONFIG_MAC_SCAN_DURATION 300
 #endif
 
+/*
+ * @def OPENTHREAD_CONFIG_MAC_SSED_TO_SSED_LINK_ENABLE
+ *
+ * SED-to-SED option.
+ *
+ * This is an experimental feature. This feature enables an SED to do CSL transmission (as CSL Transmitter) and
+ * csl reception, thus enabling SED to SED communication. When an SED is connected to another SED, they are
```
abtink, line 426:
```suggestion
 * CSL reception, thus enabling SED to SED communication. When an SED is connected to another SED, they are
```

src/core/thread/mesh_forwarder.cpp
```diff
@@ -152,6 +153,102 @@ void MeshForwarder::Stop(void)
...
+        aMessage.SetDirectTransmission();
+        break;
+    }
+#else
+    aMessage.SetDirectTransmission();
+#endif // OPENTHREAD_FTD || OPENTHREAD_MTD_S2S
+
+    mScheduleTransmissionTask.Post();
+
+    return OT_ERROR_NONE;
```
abtink, line 249: I think for future, we can consider making this method `void`. We did not return error (other than `NONE` from FTD version either).

src/core/thread/sed_to_sed.hpp
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * @def OPENTHREAD_MTD_S2S
+ *
+ * This is a alias representing if `OPENTHREAD_MTD` and `OPENTHREAD_CONFIG_MAC_SSED_TO_SSED_LINK_ENABLE` are both
+ * enabled.
+ *
+ */
+#define OPENTHREAD_MTD_S2S \
+    OPENTHREAD_MTD         \
+    &&OPENTHREAD_CONFIG_MAC_SSED_TO_SSED_LINK_ENABLE
```
abtink, line 51: Any reason to use multiple lines? Can we do this?

```
#define OPENTHREAD_MTD_S2S  (OPENTHREAD_MTD && OPENTHREAD_CONFIG_MAC_SSED_TO_SSED_LINK_ENABLE)
```

src/core/thread/sed_to_sed.hpp
```diff
@@ -0,0 +1,53 @@
...
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes definitions for SED-to-SED feature.
+ *
+ *   Note that this is not a config file. DO NOT add it to core config file.
```
abtink, line 33: 👍

src/core/thread/sed_to_sed.hpp
```diff
@@ -0,0 +1,53 @@
...
+ *
+ */
+
+#ifndef S2S_HPP_
+#define S2S_HPP_
+
+#include "openthread-core-config.h"
+
```
abtink, line 41: Should we add a check to ensure `OPENTHREAD_MTD_S2S` is not already defined and give error if it is (to ensure other config files or user don't define it directly). Something like:

```
#ifdef OPENTHREAD_MTD_S2S
#error "OPENTHREAD_MTD_S2S should not be directly defined. Use `OPENTHREAD_CONFIG_MAC_SSED_TO_SSED_LINK_ENABLE` to enable SSED to SSED link instead."
#endif 
```

abtink (approved):
Thanks @Irving-cl . Looks really great. 👍.
I have few small nit suggestions (looks great and clean already).

Irving-cl:
@jwhui

The PR is ready to merge. It's blocked by the cov issue.

PR #6121, opened 2021-01-27
[srp] fix setting srp domain when external heap is enabled

wgtdkp:
This Problem is that the SRP server is setting the domain in constructor, but for single OT instance with external heap, the `calloc` and `free` function can not be set by `otHeapSetCAllocFree` before creating the OT instance itself.

This commit delays setting domain to when starting the server.

### DIFF
--- a/src/core/net/srp_server.cpp
+++ b/src/core/net/srp_server.cpp
@@ -90,7 +90,6 @@ Server::Server(Instance &aInstance)
     , mOutstandingUpdatesTimer(aInstance, HandleOutstandingUpdatesTimer, this)
     , mEnabled(false)
 {
-    IgnoreError(SetDomain(kDefaultDomain));
 }
 
 Server::~Server(void)
@@ -405,6 +404,12 @@ void Server::Start(void)
     otError error = OT_ERROR_NONE;
 
     VerifyOrExit(!IsRunning());
+
+    if (GetDomain() == nullptr)
+    {
+        SuccessOrExit(error = SetDomain(kDefaultDomain));
+        OT_ASSERT(GetDomain() != nullptr);
+    }
 
     SuccessOrExit(error = mSocket.Open(HandleUdpReceive, this));
     SuccessOrExit(error = mSocket.Bind(0));
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (66)
+#define OPENTHREAD_API_VERSION (67)
 
 /**
  * @addtogroup api-instance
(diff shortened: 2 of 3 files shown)


src/core/net/srp_server.cpp
```diff
@@ -90,7 +90,6 @@ Server::Server(Instance &aInstance)
     , mOutstandingUpdatesTimer(aInstance, HandleOutstandingUpdatesTimer, this)
     , mEnabled(false)
 {
-    IgnoreError(SetDomain(kDefaultDomain));
```
simonlingoogle, line 93: Maybe we can eliminate the necessity of `otHeapSetCAllocFree`, i.e. use platform method for external heap alloc and free, e.g. otPlatExternalHeapAlloc, otPlatExternalHeapFree

abtink: I suggest in core module we continue to use `Instance::HeapAlloc()` (and not use platform layer `otPlat` directly). 

The intention behind having `Instanc::HeapAlloc()` is to provide a single API handling all different configurations (internal vs external, multiple/sinsgle instance, etc). Otherwise we need all `#if` check in the modules here.

simonlingoogle: I am not suggesting to use `otPlatExternalHeapAlloc` method directly in core. 

Currently, the user (posix) code has to set up function pointers using `otHeapSetCAllocFree`. 
```cpp
#if OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
    otHeapSetCAllocFree(calloc, free);
#endif
```

I am suggesting that we should define two more platform methods (i.e. otPlatExternalHeapAlloc and otPlatExternalHeapAFree) and platforms supporting external heap (e.g. POSIX) should define these two methods. Then, OT core is can freely call `Instance::HeapCAlloc` in constructors no matter it's using internal heap or external.

simonlingoogle:
> the `calloc` and `free` function can not be set by `otHeapSetCAllocFree` before creating the OT instance itself.


I think actually we can. `otHeapSetCAllocFree` don't need to pass a OT instances. It can be called before any OT instance is created. We can call `otHeapSetCAllocFree` before creating any OT instance to fix this issue. 

That being said, I still prefer to define platform methods to remove callings to `otHeapSetCAllocFree` totally.

Thoughts? @wgtdkp

src/core/net/srp_server.cpp
abtink: I see. Got it. This makes sense to me and sounds good.

If we want to use heap APIs from constructors, one thing to be careful is to ensure `Heap` constructor itself is called before other object constructors. It seems to be high on the list of member variables in `instance.hpp` so we should be fine.

wgtdkp: > I am suggesting that we should define two more platform methods (i.e. otPlatExternalHeapAlloc and otPlatExternalHeapAFree) and platforms supporting external heap (e.g. POSIX) should define these two methods. Then, OT core is can freely call Instance::HeapCAlloc in constructors no matter it's using internal heap or external.

sounds good to me as well 👍 I think we can close this PR and create another for adding the external heap platform APIs.

wgtdkp:
closing this PR as this issue is addresses by https://github.com/openthread/openthread/pull/6123.

PR #4041, opened 2019-07-31
[utils] add option to use external heap

e-rk:
This PR introduces a new API call:
```
otHeapSetCAllocFree
```
The API is available to the application from `heap.h` header and is enabled when `EXTERNAL_HEAP` common switch is set. 
The call must be invoked before the OpenThread instance is initialized.

This allows the OpenThread to use an external heap for allocation, for example from FreeRTOS.

PR #6123, opened 2021-01-28
[heap] fix external heap configuration

wgtdkp:
The `Instance::HeapCAlloc()` method can not be used in constructors when `EXTERNAL_HEAP`
is enabled but works well when internal heap is used. This is because the external heap methods
are set by public `otHeapSetCAllocFree` API after the OT instance is constructed.

This PR fixes this issue by having `Instance::HeapCAlloc()` using `otPlatCAlloc` when `EXTERNAL_HEAP`
is enabled. It also fixes build issues when `MULTIPLE_INSTANCE` and `EXTERNAL_HEAP` are used together.

Replaces #6121.

### DIFF
--- a/examples/apps/cli/main.c
+++ b/examples/apps/cli/main.c
@@ -51,7 +51,7 @@ void __gcov_flush();
 #define OPENTHREAD_ENABLE_COVERAGE 0
 #endif
 
-#if OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
+#if OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
 void *otPlatCAlloc(size_t aNum, size_t aSize)
 {
     return calloc(aNum, aSize);
--- a/src/core/api/heap_api.cpp
+++ b/src/core/api/heap_api.cpp
@@ -37,8 +37,6 @@
 
 #include "common/instance.hpp"
 
-#if !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE || OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
-
 #if OPENTHREAD_RADIO
 
 void *otHeapCAlloc(size_t aCount, size_t aSize)
@@ -61,25 +59,14 @@ void otHeapFree(void *aPointer)
     OT_ASSERT(false);
 }
 
-#else // OPENTHREAD_RADIO
-
+#else  // OPENTHREAD_RADIO
 void *otHeapCAlloc(size_t aCount, size_t aSize)
 {
-    return ot::Instance::Get().HeapCAlloc(aCount, aSize);
+    return ot::Instance::HeapCAlloc(aCount, aSize);
 }
 
 void otHeapFree(void *aPointer)
 {
-    ot::Instance::Get().HeapFree(aPointer);
+    ot::Instance::HeapFree(aPointer);
 }
-
 #endif // OPENTHREAD_RADIO
-
-#endif // !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE || OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
-
-#if OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE && !OPENTHREAD_RADIO
-void otHeapSetCAllocFree(otHeapCAllocFn aCAlloc, otHeapFreeFn aFree)
-{
-    ot::Instance::HeapSetCAllocFree(aCAlloc, aFree);
-}
-#endif // OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
--- a/src/core/common/instance.hpp
+++ b/src/core/common/instance.hpp
@@ -42,7 +42,7 @@
 #include <openthread/error.h>
 #include <openthread/heap.h>
 #include <openthread/platform/logging.h>
-#if OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
+#if OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
 #include <openthread/platform/memory.h>
 #endif
 
@@ -60,18 +60,16 @@
 #endif
 #if OPENTHREAD_FTD || OPENTHREAD_MTD
 #include "common/code_utils.hpp"
-#include "crypto/mbedtls.hpp"
-#include "thread/tmf.hpp"
-#if !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
-#include "utils/heap.hpp"
-#endif
 #include "common/notifier.hpp"
 #include "common/settings.hpp"
+#include "crypto/mbedtls.hpp"
 #include "meshcop/border_agent.hpp"
 #include "net/ip6.hpp"
 #include "thread/announce_sender.hpp"
 #include "thread/link_quality.hpp"
 #include "thread/thread_netif.hpp"
+#include "thread/tmf.hpp"
+#include "utils/heap.hpp"
 #if OPENTHREAD_CONFIG_CHANNEL_MANAGER_ENABLE && OPENTHREAD_FTD
 #include "utils/channel_manager.hpp"
 #endif
@@ -247,28 +245,11 @@ public:
     otError ErasePersistentInfo(void);
 
 #if OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
-    void HeapFree(void *aPointer)
-    {
-        OT_ASSERT(mFree != nullptr);
-
-        mFree(aPointer);
-    }
-
-    void *HeapCAlloc(size_t aCount, size_t aSize)
-    {
-        OT_ASSERT(mCAlloc != nullptr);
-
-        return mCAlloc(aCount, aSize);
-    }
-
-    static void HeapSetCAllocFree(otHeapCAllocFn aCAlloc, otHeapFreeFn aFree)
-    {
-        mFree   = aFree;
-        mCAlloc = aCAlloc;
-    }
-#elif !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
-    void  HeapFree(void *aPointer) { mHeap.Free(aPointer); }
-    void *HeapCAlloc(size_t aCount, size_t aSize) { return mHeap.CAlloc(aCount, aSize); }
+    static void  HeapFree(void *aPointer) { otPlatFree(aPointer); }
+    static void *HeapCAlloc(size_t aCount, size_t aSize) { return otPlatCAlloc(aCount, aSize); }
+#else
+    static void  HeapFree(void *aPointer) { sHeap.Free(aPointer); }
+    static void *HeapCAlloc(size_t aCount, size_t aSize) { return sHeap.CAlloc(aCount, aSize); }
 
     /**
      * This method returns a reference to the Heap object.
@@ -276,10 +257,7 @@ public:
      * @returns A reference to the Heap object.
      *
      */
-    Utils::Heap &GetHeap(void) { return mHeap; }
-#else
-    void  HeapFree(void *aPointer) { otPlatFree(aPointer); }
-    void *HeapCAlloc(size_t aCount, size_t aSize) { return otPlatCAlloc(aCount, aSize); }
+    Utils::Heap &GetHeap(void) { return sHeap; }
 #endif // OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
 
 #if OPENTHREAD_CONFIG_COAP_API_ENABLE
@@ -338,11 +316,8 @@ private:
 #if OPENTHREAD_MTD || OPENTHREAD_FTD
     // RandomManager is initialized before other objects. Note that it
     // requires MbedTls which itself may use Heap.
-#if OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
-    static otHeapFreeFn   mFree;
-    static otHeapCAllocFn mCAlloc;
-#elif !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
-    Utils::Heap  mHeap;
+#if !OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
+    static Utils::Heap sHeap;
 #endif
     Crypto::MbedTls mMbedTls;
 #endif // OPENTHREAD_MTD || OPENTHREAD_FTD
--- a/src/core/common/message.cpp
+++ b/src/core/common/message.cpp
@@ -111,7 +111,7 @@ Buffer *MessagePool::NewBuffer(Message::Priority aPriority)
 
     while ((
 #if OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE
-               buffer = static_cast<Buffer *>(GetInstance().HeapCAlloc(sizeof(Buffer), 1))
+               buffer = static_cast<Buffer *>(Instance::HeapCAlloc(1, sizeof(Buffer)))
 #elif OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT
                buffer = static_cast<Buffer *>(otPlatMessagePoolNew(&GetInstance()))
 #else
@@ -143,7 +143,7 @@ void MessagePool::FreeBuffers(Buffer *aBuffer)
     {
         Buffer *next = aBuffer->GetNextBuffer();
 #if OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE
-        GetInstance().HeapFree(aBuffer);
+        Instance::HeapFree(aBuffer);
 #elif OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT
         otPlatMessagePoolFree(&GetInstance(), aBuffer);
 #else
--- a/src/core/crypto/mbedtls.cpp
+++ b/src/core/crypto/mbedtls.cpp
@@ -49,29 +49,15 @@
 namespace ot {
 namespace Crypto {
 
-#if !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE && OPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS_MANAGEMENT
-
-static void *CAlloc(size_t aCount, size_t aSize)
-{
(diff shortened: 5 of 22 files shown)


src/core/config/openthread-core-default-config.h
```diff
@@ -294,13 +294,13 @@
...
 
 /**
- * @def OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
+ * @def OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE
  *
  * Enable the external heap.
  *
  */
-#ifndef OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
-#define OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE 0
```
simonlingoogle, line 303: Should we add a checking in openthread-core-config-check.h to tell users that `OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE` is now replaced by `OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE`?

examples/apps/cli/main.c
```diff
@@ -51,7 +51,7 @@ void __gcov_flush();
 #define OPENTHREAD_ENABLE_COVERAGE 0
 #endif
 
-#if OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
+#if OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE
```
simonlingoogle, line 54:
```suggestion
#if OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE || OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
```

src/core/utils/heap.hpp
```diff
@@ -37,6 +37,8 @@
 
 #include "openthread-core-config.h"
 
+#if !OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE
```
simonlingoogle, line 40: also need to add `#if !OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE ... #endif` to `heap.cpp`?

src/core/api/heap_api.cpp
```diff
@@ -63,6 +63,17 @@ void otHeapFree(void *aPointer)
...
+void *otHeapCAlloc(size_t aCount, size_t aSize)
+{
+    return ot::Instance::HeapCAlloc(aCount, aSize);
+}
+
+void otHeapFree(void *aPointer)
+{
+    ot::Instance::HeapFree(aPointer);
+}
+#else
```
simonlingoogle, line 76:
```suggestion
```

can we use the same implementation for both case?

src/core/common/instance.hpp
```diff
@@ -277,10 +258,7 @@ class Instance : public otInstance, private NonCopyable
      *
      */
     Utils::Heap &GetHeap(void) { return mHeap; }
```
simonlingoogle, line 260: No other module should be using this method. If they does, the code is not working when external heap is enabled. So, I would suggest to remove `GetHeap()` method in the first place to avoid potential issues.

src/posix/platform/memory.cpp
```diff
@@ -0,0 +1,44 @@
...
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "openthread-posix-config.h"
+#include "platform-posix.h"
```
simonlingoogle, line 30:
```suggestion
#include "platform-posix.h"
```

src/posix/platform/memory.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include <openthread/platform/memory.h>
```
simonlingoogle, line 32:
```suggestion
#include <stdlib.h>

#include <openthread/platform/memory.h>
```

src/core/config/openthread-core-default-config.h
wgtdkp: Sure, thanks!

src/core/api/heap_api.cpp
```diff
@@ -63,6 +63,17 @@ void otHeapFree(void *aPointer)
 
 #else // OPENTHREAD_RADIO
 
+#if OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE
+void *otHeapCAlloc(size_t aCount, size_t aSize)
+{
+    return ot::Instance::HeapCAlloc(aCount, aSize);
+}
+
+void otHeapFree(void *aPointer)
```
abtink, line 72: I think at some later point we may want to remove these public APIs. I see the `TODO` in the header files for these and they are marked to be deprecated.

abtink (approved):
Thanks @wgtdkp for taking care of this.

This change (as is) is already a great improvement. 👍 

One thing that I noticed (which we may address later in future PR), is that in `MULTI_INSTANCE` case we seem to always end up using external heap (basically the `OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE` becomes "don't-care"). The question for me is why do we have this restriction and do we need it?
- When using internal heap, each `Instance` object will have its own `Utils::Heap` so technically we can have internal heap in multi-instance case.
- The challenge/reason may have been due to way the public OT API `otHeapCAlloc()` and `otHeapFree()` are defined. They don't have `otInstacne *aInstance` as parameter, so may not know which instance to use in the multi-instance case (but I guess we can either change them or even better remove them since they are marked to be deprecated).

Thoughts?

src/core/api/heap_api.cpp
wgtdkp: good suggestion.

examples/apps/cli/main.c
wgtdkp: Actually, the platform APIs are required for only `EXTERNAL_HEAP_ENABLE` and we don't care `MULTIPLE_INSTANCE_ENABLE`. That says, we allow both internal and external heap for multiple OT instances.

src/core/common/instance.hpp
wgtdkp: This makes sense to me, but unfortunately, there is actually `OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE` that will use `GetHeap()` to get the capacity of the internal heap for message pool. I don't have a good idea about the meaningful capacity of external heap as as there is no platform APIs for this. For now, lets keep this for `OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE`.

src/posix/platform/memory.cpp
wgtdkp: Other posix platform source files have this header pair, let's keep it.

jwhui:
> One thing that I noticed (which we may address later in future PR), is that in `MULTI_INSTANCE` case we seem to always end up using external heap (basically the `OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE` becomes "don't-care"). The question for me is why do we have this restriction and do we need it?

@abtink, because mbedTLS has no way of passing the instance.
https://github.com/openthread/openthread/blob/46f796d4f041c80c716b1ede81eba13c9b9e4f06/third_party/mbedtls/repo/include/mbedtls/platform.h#L130-L140

src/core/api/heap_api.cpp
wgtdkp: > The challenge/reason may have been due to way the public OT API otHeapCAlloc() and otHeapFree() are defined. They don't have otInstacne *aInstance as parameter, so may not know which instance to use in the multi-instance case (but I guess we can either change them or even better remove them since they are marked to be deprecated).

Those APIs are currently used by nordic boards to setup mbedtls and I think the function signature can not be changed.

jwhui: As mentioned in https://github.com/openthread/openthread/pull/6123#issuecomment-769581199, the function signature is defined by mbedTLS.

https://github.com/openthread/openthread/blob/46f796d4f041c80c716b1ede81eba13c9b9e4f06/third_party/mbedtls/repo/include/mbedtls/platform.h#L130-L140

src/core/common/instance.hpp
simonlingoogle: It is used by `MessagePool::GetTotalBufferCount and MessagePool::GetFreeBufferCount`. I think a simple solution is to let these functions always return very large values (e.x. total=4096, free=4096) when external heap is enabled.

simonlingoogle:
> > One thing that I noticed (which we may address later in future PR), is that in `MULTI_INSTANCE` case we seem to always end up using external heap (basically the `OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE` becomes "don't-care"). The question for me is why do we have this restriction and do we need it?
> 
> @abtink, because mbedTLS has no way of passing the instance.
> https://github.com/openthread/openthread/blob/46f796d4f041c80c716b1ede81eba13c9b9e4f06/third_party/mbedtls/repo/include/mbedtls/platform.h#L130-L140

I think the calls to `mbedtls_platform_set_calloc_free` can be updated be like `mbedtls_platform_set_calloc_free(otPlatCAlloc, otPlatFree) instead of using OT public APIs. 
Other than this, I would agree with @abtink that these two APIs can be removed since they are not used anywhere else.

jwhui:
> > > One thing that I noticed (which we may address later in future PR), is that in `MULTI_INSTANCE` case we seem to always end up using external heap (basically the `OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE` becomes "don't-care"). The question for me is why do we have this restriction and do we need it?
> > 
> > 
> > @abtink, because mbedTLS has no way of passing the instance.
> > https://github.com/openthread/openthread/blob/46f796d4f041c80c716b1ede81eba13c9b9e4f06/third_party/mbedtls/repo/include/mbedtls/platform.h#L130-L140
> 
> I think the calls to `mbedtls_platform_set_calloc_free` can be updated be like `mbedtls_platform_set_calloc_free(otPlatCAlloc, otPlatFree)` instead of using OT public APIs.
> Other than this, I would agree with @abtink that these two APIs can be removed since they are not used anywhere else.

@simonlingoogle , but that doesn't address @abtink's question of why multiple instance requires external heap.

src/core/api/heap_api.cpp
wgtdkp: I think the the `otHeapCAlloc` and `otHeapFree` APIs cannot support `MULTIPLE_INSTANCE` unless we enforce `EXTERNAL_HEAP` for `MULTIPLE_INSTANCE`.

I think the intention for those two API is for the application to use OT's internal heap to setup mbedtls. If external heap is enabled, the application should just use `calloc` directly.

We should make it clear that the `otHeapCAlloc` and `otHeapFree` APIs are not available for `MULTIPLE_INSTANCE`. thoughts?

jwhui: I think multiple instance requires external heap due to mbedTLS constraints.

abtink:
> Those APIs are currently used by nordic boards to setup mbedtls and I think the function signature can not be changed.

> @abtink, because mbedTLS has no way of passing the instance.

Thanks guys. I see! It makes sense. 

----

> I think the calls to mbedtls_platform_set_calloc_free can be updated be like mbedtls_platform_set_calloc_free(otPlatCAlloc, otPlatFree) instead of using OT public APIs.
Other than this, I would agree with @abtink that these two APIs can be removed since they are not used anywhere else.

Not sure if we can do this with current model. If we are using internal heap with OT core, then I think there may not even be an `otPlatCAlloc` implementation. Right?

----

I would say let's keep it as is for now. 😃 

If I remember correctly from earlier changes,  the plan is at some point to move the `Util::Heap` into platform layer (as a platform library util) . Then from OT core perspective we always use platform APIs `otPlatCAlloc` and `otPlatFree()`. Platform layer can then choose to use the `Util:Heap` or another way to provide these APIs. I think if we do this, then we can have `mbedtls` use the `otPlatCAlloc()/otPlatFree()` APIs directly and remove the public `otHeap` APIs. Does this make sense?

wgtdkp:
> If I remember correctly from earlier changes/discussions, the plan is/was at some point to move the Util::Heap into platform layer (as a platform library util) . Then from OT core perspective we always use platform APIs otPlatCAlloc and otPlatFree(). Platform layer can then choose to use the Util:Heap or another way to provide these APIs. I think if we do this, then we can have mbedtls use the otPlatCAlloc()/otPlatFree() APIs directly and remove the public otHeap APIs. Does this make sense?

If we want to have multiple OT instances use the same internal heap, I think we can use a static `Utils::Heap` member in class `Instance`:
```c++
#if !OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENBALE
    static Utils::Heap mHeap;
#endif
```

simonlingoogle:
> Not sure if we can do this with current model. If we are using internal heap with OT core, then I think there may not even be an `otPlatCAlloc` implementation. Right?

I think the platform can always define these two platform methods if it need, no matter it's using internal heap or external heap.

simonlingoogle:
> That is a good idea to have the internal heap as independent library. If we want to have multiple OT instances use the same internal heap, I think we can use a static `Utils::Heap` member in class `Instance`:
> 
> ```c++
> #if !OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENBALE
>     static Utils::Heap mHeap;
> #endif
> ```

Agree with @wgtdkp that it's a good idea to make the Heap static so that it's usable without having reference to any OT instance. 

Actually I think it's a preferable solution because user can easily set the heap size to the available RAM size on the device, no matter how many OT instance it runs. The heap memory space is also automatically shared between OT instances, making it more memory efficient.

wgtdkp:
@jwhui @abtink @simonlingoogle We changed the `mHeap` of `Instance` to a static member in the latest push, please review.

src/core/common/message.cpp
```diff
@@ -111,7 +111,7 @@ Buffer *MessagePool::NewBuffer(Message::Priority aPriority)
 
     while ((
 #if OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE
-               buffer = static_cast<Buffer *>(GetInstance().HeapCAlloc(sizeof(Buffer), 1))
+               buffer = static_cast<Buffer *>(Instance::HeapCAlloc(sizeof(Buffer), 1))
```
simonlingoogle, line 114:
```suggestion
               buffer = static_cast<Buffer *>(Instance::HeapCAlloc(1, sizeof(Buffer)))
```

src/core/crypto/mbedtls.cpp
```diff
@@ -49,29 +49,29 @@
...
 
 MbedTls::MbedTls(void)
 {
-#if !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE && OPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS_MANAGEMENT
+#if OPENTHREAD_CONFIG_ENABLE_BUILTIN_MBEDTLS_MANAGEMENT
 #ifdef MBEDTLS_DEBUG_C
     // mbedTLS's debug level is almost the same as OpenThread's
     mbedtls_debug_set_threshold(OPENTHREAD_CONFIG_LOG_LEVEL);
 #endif
     mbedtls_platform_set_calloc_free(CAlloc, Free);
```
simonlingoogle, line 73:
```suggestion
    mbedtls_platform_set_calloc_free(Instance::HeapCAlloc, Instance::HeapFree);
```

simonlingoogle (approved):
LGTM 👍 

It looks much better than before!

src/core/common/instance.hpp
```diff
@@ -338,11 +316,8 @@ class Instance : public otInstance, private NonCopyable
 #if OPENTHREAD_MTD || OPENTHREAD_FTD
     // RandomManager is initialized before other objects. Note that it
     // requires MbedTls which itself may use Heap.
-#if OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
-    static otHeapFreeFn   mFree;
-    static otHeapCAllocFn mCAlloc;
-#elif !OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE
-    Utils::Heap  mHeap;
+#if !OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE
+    static Utils::Heap mHeap;
```
abtink, line 320: Shall we call it `sHeap` (since it is `static` member vairable)?

abtink (approved):
Looks really great. Thanks @wgtdkp. 👍 

One small nit suggestion.

src/core/config/openthread-core-config-check.h
```diff
@@ -93,6 +93,10 @@
 
 //---------------------------------------------------------------------------------------------------------------------
 
+#ifdef OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
+#error "OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE was replaced by OPENTHREAD_CONFIG_EXTERNAL_HEAP_ENABLE"
```
jwhui, line 97: Why are we renaming this? We have other configs that share the same `OPENTHREAD_CONFIG_HEAP_` prefix.

jwhui:
@LuDuda , given that #4041 and #4143 contributed to the existing external heap implementation, can you help review this PR?

src/core/config/openthread-core-config-check.h
wgtdkp: Not really necessary, reverted the change.

etc/gn/openthread.gni
```diff
@@ -139,7 +139,7 @@ if (openthread_enable_core_config_args) {
     openthread_config_mlr_enable = false
 
     # Enable external heap support
-    openthread_config_heap_external_enable = false
+    openthread_config_external_heap_enable = false
```
jwhui, line 142: Avoid this rename to be consistent with the macro?

LuDuda:
@MarekPorwisz is going to check PR. Thanks!

MarekPorwisz (approved):
I'm not that familiar with the code, but from my point of view changes look ok.

jwhui (approved):

PR #6125, opened 2021-01-29
[settings] add APIs to securely store critical settings

zhanglongxia:
All the settings are written into a local file on the posix platform. It is easy for
others to get the sensitive infomation, such as master key, from the local file.

This commit includes the following enhancement:
<1> Add API `otPlatSettingsSetCriticalKeys()` to tell the platform which keys
        should be stored in the secure area.
<2> Expose all settings keys to the header file `settings.h`, which allows
        the platform determines which settings should be processed specially.
<3> Add new posix APIs `otPlatSecureSettingsXxx` to posix platform to allow
        the vendor to implement the functions to store the critical settings to
        the secure area.

### DIFF
--- a/include/openthread/platform/settings.h
+++ b/include/openthread/platform/settings.h
@@ -52,6 +52,29 @@ extern "C" {
  */
 
 /**
+ * This enumeration defines the keys of settings.
+ *
+ * Note: When adding a new setings key, if the settings corresponding to the key contains security sensitive
+ *       information, the developer MUST add the key to the array `kCriticalKeys`.
+ *
+ */
+enum Key
+{
+    OT_SETTINGS_KEY_ACTIVE_DATASET  = 0x0001, ///< Active Operational Dataset
+    OT_SETTINGS_KEY_PENDING_DATASET = 0x0002, ///< Pending Operational Dataset
+    OT_SETTINGS_KEY_NETWORK_INFO    = 0x0003, ///< Thread network information
+    OT_SETTINGS_KEY_PARENT_INFO     = 0x0004, ///< Parent information
+    OT_SETTINGS_KEY_CHILD_INFO      = 0x0005, ///< Child information
+    OT_SETTINGS_KEY_RESERVED        = 0x0006, ///< Reserved (previously auto-start)
+    OT_SETTINGS_KEY_SLAAC_IID_SECRET_KEY =
+        0x0007,                              ///< Secret key used by SLAAC module for generating semantically opaque IID
+    OT_SETTINGS_KEY_DAD_INFO       = 0x0008, ///< Duplicate Address Detection (DAD) information.
+    OT_SETTINGS_KEY_OMR_PREFIX     = 0x0009, ///< Off-mesh routable (OMR) prefix.
+    OT_SETTINGS_KEY_ON_LINK_PREFIX = 0x000a, ///< On-link prefix for infrastructure link.
+    OT_SETTINGS_KEY_SRP_ECDSA_KEY  = 0x000b, ///< SRP client ECDSA public/private key pair.
+};
+
+/**
  * Performs any initialization for the settings subsystem, if necessary.
  *
  * @param[in]  aInstance The OpenThread instance structure.
@@ -68,13 +91,13 @@ void otPlatSettingsInit(otInstance *aInstance);
 void otPlatSettingsDeinit(otInstance *aInstance);
 
 /**
- * This function sets the critical keys that should be stored in a secure area.
+ * This function sets the critical keys that should be stored in the secure area.
  *
- * Note that the memory pointed by @p aKeys will never be released by the caller.
+ * Note that the memory pointed by @p aKeys MUST not be released before @p aInstance is destroyed.
  *
  * @param[in]  aInstance    The OpenThread instance structure.
- * @param[in]  aKeys        A pointer to the value of the critical keys.
- * @param[in]  aKeysLength  The length of the keys pointed to by @p akeys.
+ * @param[in]  aKeys        A pointer to an array containing the list of critical keys.
+ * @param[in]  aKeysLength  The number of entries in the @p aKeys array.
  *
  */
 void otPlatSettingsSetCriticalKeys(otInstance *aInstance, const uint16_t *aKeys, uint16_t aKeysLength);
--- a/include/openthread/platform/settings.h
+++ b/include/openthread/platform/settings.h
@@ -67,6 +67,18 @@ void otPlatSettingsInit(otInstance *aInstance);
  */
 void otPlatSettingsDeinit(otInstance *aInstance);
 
+/**
+ * This function sets the critical keys that should be stored in a secure area.
+ *
+ * Note that the memory pointed by @p aKeys will never be released by the caller.
+ *
+ * @param[in]  aInstance    The OpenThread instance structure.
+ * @param[in]  aKeys        A pointer to the value of the critical keys.
+ * @param[in]  aKeysLength  The length of the keys pointed to by @p akeys.
+ *
+ */
+void otPlatSettingsSetCriticalKeys(otInstance *aInstance, const uint16_t *aKeys, uint16_t aKeysLength);
+
 /// Fetches the value of a setting
 /** This function fetches the value of the setting identified
  *  by aKey and write it to the memory pointed to by aValue.
--- a/include/openthread/platform/settings.h
+++ b/include/openthread/platform/settings.h
@@ -60,14 +60,14 @@ extern "C" {
  */
 enum
 {
-    OT_SETTINGS_KEY_ACTIVE_DATASET  = 0x0001, ///< Active Operational Dataset
-    OT_SETTINGS_KEY_PENDING_DATASET = 0x0002, ///< Pending Operational Dataset
-    OT_SETTINGS_KEY_NETWORK_INFO    = 0x0003, ///< Thread network information
-    OT_SETTINGS_KEY_PARENT_INFO     = 0x0004, ///< Parent information
-    OT_SETTINGS_KEY_CHILD_INFO      = 0x0005, ///< Child information
-    OT_SETTINGS_KEY_RESERVED        = 0x0006, ///< Reserved (previously auto-start)
+    OT_SETTINGS_KEY_ACTIVE_DATASET  = 0x0001, ///< Active Operational Dataset.
+    OT_SETTINGS_KEY_PENDING_DATASET = 0x0002, ///< Pending Operational Dataset.
+    OT_SETTINGS_KEY_NETWORK_INFO    = 0x0003, ///< Thread network information.
+    OT_SETTINGS_KEY_PARENT_INFO     = 0x0004, ///< Parent information.
+    OT_SETTINGS_KEY_CHILD_INFO      = 0x0005, ///< Child information.
+    OT_SETTINGS_KEY_RESERVED        = 0x0006, ///< Reserved (previously auto-start).
     OT_SETTINGS_KEY_SLAAC_IID_SECRET_KEY =
-        0x0007,                              ///< Secret key used by SLAAC module for generating semantically opaque IID
+        0x0007, ///< Secret key used by SLAAC module for generating semantically opaque IID.
     OT_SETTINGS_KEY_DAD_INFO       = 0x0008, ///< Duplicate Address Detection (DAD) information.
     OT_SETTINGS_KEY_OMR_PREFIX     = 0x0009, ///< Off-mesh routable (OMR) prefix.
     OT_SETTINGS_KEY_ON_LINK_PREFIX = 0x000a, ///< On-link prefix for infrastructure link.
--- a/include/openthread/platform/settings.h
+++ b/include/openthread/platform/settings.h
@@ -60,18 +60,17 @@ extern "C" {
  */
 enum
 {
-    OT_SETTINGS_KEY_ACTIVE_DATASET  = 0x0001, ///< Active Operational Dataset.
-    OT_SETTINGS_KEY_PENDING_DATASET = 0x0002, ///< Pending Operational Dataset.
-    OT_SETTINGS_KEY_NETWORK_INFO    = 0x0003, ///< Thread network information.
-    OT_SETTINGS_KEY_PARENT_INFO     = 0x0004, ///< Parent information.
-    OT_SETTINGS_KEY_CHILD_INFO      = 0x0005, ///< Child information.
-    OT_SETTINGS_KEY_RESERVED        = 0x0006, ///< Reserved (previously auto-start).
-    OT_SETTINGS_KEY_SLAAC_IID_SECRET_KEY =
-        0x0007, ///< Secret key used by SLAAC module for generating semantically opaque IID.
-    OT_SETTINGS_KEY_DAD_INFO       = 0x0008, ///< Duplicate Address Detection (DAD) information.
-    OT_SETTINGS_KEY_OMR_PREFIX     = 0x0009, ///< Off-mesh routable (OMR) prefix.
-    OT_SETTINGS_KEY_ON_LINK_PREFIX = 0x000a, ///< On-link prefix for infrastructure link.
-    OT_SETTINGS_KEY_SRP_ECDSA_KEY  = 0x000b, ///< SRP client ECDSA public/private key pair.
+    OT_SETTINGS_KEY_ACTIVE_DATASET       = 0x0001, ///< Active Operational Dataset.
+    OT_SETTINGS_KEY_PENDING_DATASET      = 0x0002, ///< Pending Operational Dataset.
+    OT_SETTINGS_KEY_NETWORK_INFO         = 0x0003, ///< Thread network information.
+    OT_SETTINGS_KEY_PARENT_INFO          = 0x0004, ///< Parent information.
+    OT_SETTINGS_KEY_CHILD_INFO           = 0x0005, ///< Child information.
+    OT_SETTINGS_KEY_RESERVED             = 0x0006, ///< Reserved (previously auto-start).
+    OT_SETTINGS_KEY_SLAAC_IID_SECRET_KEY = 0x0007, ///< SLAAC key to generate semantically opaque IID.
+    OT_SETTINGS_KEY_DAD_INFO             = 0x0008, ///< Duplicate Address Detection (DAD) information.
+    OT_SETTINGS_KEY_OMR_PREFIX           = 0x0009, ///< Off-mesh routable (OMR) prefix.
+    OT_SETTINGS_KEY_ON_LINK_PREFIX       = 0x000a, ///< On-link prefix for infrastructure link.
+    OT_SETTINGS_KEY_SRP_ECDSA_KEY        = 0x000b, ///< SRP client ECDSA public/private key pair.
 };
 
 /**
--- a/include/openthread/platform/settings.h
+++ b/include/openthread/platform/settings.h
@@ -58,7 +58,7 @@ extern "C" {
  *       information, the developer MUST add the key to the array `kCriticalKeys`.
  *
  */
-enum Key
+enum
 {
     OT_SETTINGS_KEY_ACTIVE_DATASET  = 0x0001, ///< Active Operational Dataset
     OT_SETTINGS_KEY_PENDING_DATASET = 0x0002, ///< Pending Operational Dataset
--- a/src/core/common/settings.cpp
+++ b/src/core/common/settings.cpp
@@ -117,6 +117,11 @@ void SettingsDriver::Init(void)
     otPlatSettingsInit(&GetInstance());
 }
 
+void SettingsDriver::SetCriticalKeys(const uint16_t *aKeys, uint16_t aKeysLength)
+{
+    otPlatSettingsSetCriticalKeys(&GetInstance(), aKeys, aKeysLength);
+}
+
 void SettingsDriver::Deinit(void)
 {
     otPlatSettingsDeinit(&GetInstance());
@@ -160,6 +165,12 @@ void SettingsDriver::Init(void)
     mFlash.Init();
 }
 
+void SettingsDriver::SetCriticalKeys(const uint16_t *aKeys, uint16_t aKeysLength)
+{
+    OT_UNUSED_VARIABLE(aKeys);
+    OT_UNUSED_VARIABLE(aKeysLength);
+}
+
 void SettingsDriver::Deinit(void)
 {
 }
@@ -194,6 +205,7 @@ void SettingsDriver::Wipe(void)
 void Settings::Init(void)
 {
     Get<SettingsDriver>().Init();
+    Get<SettingsDriver>().SetCriticalKeys(mCriticalKeys, OT_ARRAY_LENGTH(mCriticalKeys));
 }
 
 void Settings::Deinit(void)
@@ -579,3 +591,13 @@ otError Settings::Delete(Key aKey)
 }
 
 } // namespace ot
+
+//---------------------------------------------------------------------------------------------------------------------
+// Default/weak implementation of settings platform APIs
+
+OT_TOOL_WEAK void otPlatSettingsSetCriticalKeys(otInstance *aInstance, const uint16_t *aKeys, uint16_t aKeysLength)
+{
+    OT_UNUSED_VARIABLE(aInstance);
+    OT_UNUSED_VARIABLE(aKeys);
+    OT_UNUSED_VARIABLE(aKeysLength);
+}
--- a/src/core/common/settings.cpp
+++ b/src/core/common/settings.cpp
@@ -43,6 +43,9 @@
 #include "thread/mle.hpp"
 
 namespace ot {
+// This array contains critical keys that should be stored in the secure area.
+static const uint16_t kCriticalKeys[] = {OT_SETTINGS_KEY_ACTIVE_DATASET, OT_SETTINGS_KEY_PENDING_DATASET,
+                                         OT_SETTINGS_KEY_SRP_ECDSA_KEY};
 
 // LCOV_EXCL_START
 
@@ -117,14 +120,14 @@ void SettingsDriver::Init(void)
     otPlatSettingsInit(&GetInstance());
 }
 
-void SettingsDriver::SetCriticalKeys(const uint16_t *aKeys, uint16_t aKeysLength)
-{
-    otPlatSettingsSetCriticalKeys(&GetInstance(), aKeys, aKeysLength);
-}
-
 void SettingsDriver::Deinit(void)
 {
     otPlatSettingsDeinit(&GetInstance());
+}
+
+void SettingsDriver::SetCriticalKeys(const uint16_t *aKeys, uint16_t aKeysLength)
+{
+    otPlatSettingsSetCriticalKeys(&GetInstance(), aKeys, aKeysLength);
 }
 
 otError SettingsDriver::Add(uint16_t aKey, const uint8_t *aValue, uint16_t aValueLength)
@@ -165,14 +168,14 @@ void SettingsDriver::Init(void)
     mFlash.Init();
 }
 
+void SettingsDriver::Deinit(void)
+{
+}
+
 void SettingsDriver::SetCriticalKeys(const uint16_t *aKeys, uint16_t aKeysLength)
 {
     OT_UNUSED_VARIABLE(aKeys);
     OT_UNUSED_VARIABLE(aKeysLength);
-}
-
-void SettingsDriver::Deinit(void)
-{
 }
 
 otError SettingsDriver::Add(uint16_t aKey, const uint8_t *aValue, uint16_t aValueLength)
@@ -205,7 +208,7 @@ void SettingsDriver::Wipe(void)
 void Settings::Init(void)
 {
     Get<SettingsDriver>().Init();
-    Get<SettingsDriver>().SetCriticalKeys(mCriticalKeys, OT_ARRAY_LENGTH(mCriticalKeys));
+    Get<SettingsDriver>().SetCriticalKeys(kCriticalKeys, OT_ARRAY_LENGTH(kCriticalKeys));
 }
 
 void Settings::Deinit(void)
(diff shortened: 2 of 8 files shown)


zhanglongxia:
This CL adds a function `otPlatSettingsSetCriticalKeys()` to the `include/openthread/platform/settings.h` to tell the platform which setting keys are critical setting keys. Another method  is to move all the setting key definitions from  `src/core/common/settings.hpp` to the file `include/openthread/platform/settings.h`. Platform defines which settings will be securely stored. Which method do you think is better? @abtink @bukepo

bukepo:
It seems more flexible to me if the platform can decide where to store each setting.

wgtdkp:
@zhanglongxia do you think if it is feasible for the platform to save all the settings values in secure storage, but not for specific items? I think if we want to stop other applications from touching `masterkey`, we should also want to stop them from touching other dataset parameters. For such platforms that requires secure settings, they can just implement the `settings`/`flash` platform APIs using platform-specific approaches, for example, write settings into databases. Thoughts?

zhanglongxia:
@wgtdkp We can store  all settings to the secure storage. But there is no need to store all settings to the secure storage,  there may be not enough secure area for some embeded platforms. As I said in the comment above, we can expose all the setting keys to the `include/openthread/platform/settings.h`. The platform can determine which settings should be securely stored.

simonlingoogle:
> As I said in the comment above, we can expose all the setting keys to the `include/openthread/platform/settings.h`. The platform can determine which settings should be securely stored.

Prefer this strategy too because it brings less APIs and less change to OpenThread. What OpenThread would do is just exposing the setting keys to public API so that platform code has a chance to decide by itself how to store each key.

src/posix/platform/include/openthread/platform/secure_settings.h
```diff
@@ -0,0 +1,178 @@
...
+ *
+ */
+
+/**
+ * This function performs any initialization for the secure settings subsystem, if necessary.
+ *
+ * @param[in]  aInstance The OpenThread instance structure.
+ *
+ */
+void otPlatSecureSettingsInit(otInstance *aInstance);
```
abtink, line 60: If I understand the code, these functions are intended to be used by the platform layer only, and not directly called by OpenThread core itself (i.e. OT code will only call/use `otPlatSettingsSetCriticalKeys()` to indicate which keys are critical and should use be saved differently). 
- Is this correct? 
- If so, I think it's better not to use `otPlat` prefix in the name of the functions here.  
- Generally, `otPlat{}` is used for platform layer functions/definitions that are intended for direct use by OT core code.

include/openthread/platform/settings.h
```diff
@@ -67,6 +67,18 @@ void otPlatSettingsInit(otInstance *aInstance);
  */
 void otPlatSettingsDeinit(otInstance *aInstance);
 
+/**
+ * This function sets the critical keys that should be stored in a secure area.
+ *
+ * Note that the memory pointed by @p aKeys will never be released by the caller.
+ *
+ * @param[in]  aInstance    The OpenThread instance structure.
+ * @param[in]  aKeys        A pointer to the value of the critical keys.
```
abtink, line 76:
```suggestion
 * @param[in]  aKeys        A pointer to an array containing the list of critical keys.
```

include/openthread/platform/settings.h
```diff
@@ -67,6 +67,18 @@ void otPlatSettingsInit(otInstance *aInstance);
...
+ * @param[in]  aKeysLength  The length of the keys pointed to by @p akeys.
```
abtink, line 77:
```suggestion
 * @param[in]  aKeysLength  The number of entries in the @p aKeys array.
```

abtink:
> This CL adds a function otPlatSettingsSetCriticalKeys() to the include/openthread/platform/settings.h to tell the platform which setting keys are critical setting keys. Another method is to move all the setting key definitions from src/core/common/settings.hpp to the file include/openthread/platform/settings.h. Platform defines which settings will be securely stored. Which method do you think is better? @abtink @bukepo

I prefer that we keep settings key definitions in OpenThread core. Here is how I see this:
- It's the OT core that is defining what needs to be saved and assigning setting key numbers and knows what is the data that ie being saved.  
- The platform layer should be simply saving the key/values as a dictionary/map and treat the value as blobs of data and not be aware of what the keys and/or values are. 
- So if we need to decide that certain data need to be saved differently (e.g in secure storange), I think it would be better for it to be decided by OT core. 
- I like the current model that OT core specifies which keys are critical through a new `otPlat` API (platform layer can then choose to do something different for those keys but does not need to know what the keys are).
- I don't think delegating the decision directly to platform would make it easier:
  - If we delegate this to platform layer, when a new settings key is added in OT core, we need to check all platform implementations and ensure they are doing the right thing for new key being added.

src/posix/platform/include/openthread/platform/secure_settings.h
zhanglongxia: Yes, the APIs `otPlatSecureSettingsXxx` will only be used by the posix platform layer. I will change them from 'otPlatSecureSettingsXxx' to `otPosixSecureSettingsXxx`.

include/openthread/platform/settings.h
zhanglongxia: Will update it.

include/openthread/platform/settings.h
zhanglongxia: Will update it.

zhanglongxia:
What do you think about abtink's  comments? @bukepo @simonlingoogle

include/openthread/platform/settings.h
```diff
@@ -67,6 +67,18 @@ void otPlatSettingsInit(otInstance *aInstance);
  */
 void otPlatSettingsDeinit(otInstance *aInstance);
 
+/**
+ * This function sets the critical keys that should be stored in a secure area.
+ *
+ * Note that the memory pointed by @p aKeys will never be released by the caller.
```
wgtdkp, line 73:
```suggestion
 * Note that the memory pointed by @p aKeys MUST not be released before @p aInstance is destroyed.
```

src/core/common/settings.hpp
```diff
@@ -1111,6 +1120,8 @@ class Settings : public SettingsBase, private NonCopyable
     otError Save(Key aKey, const void *aValue, uint16_t aSize);
     otError Add(Key aKey, const void *aValue, uint16_t aSize);
     otError Delete(Key aKey);
+
+    const uint16_t mCriticalKeys[3] = {kKeyActiveDataset, kKeyPendingDataset, kKeySrpEcdsaKey};
```
wgtdkp, line 1124: @jwhui @abtink Do we encourage in-class member initialization or always initialize in the constructor?

src/posix/platform/settings.cpp
```diff
@@ -47,6 +47,9 @@
 #include <openthread/platform/misc.h>
 #include <openthread/platform/radio.h>
 #include <openthread/platform/settings.h>
+#if OPENTHREAD_POSIX_CONFIG_SECURE_SETTINGS_ENABLE
+#include <openthread/platform/secure_settings.h>
+#endif
```
wgtdkp, line 52:
```suggestion
#include <openthread/platform/secure_settings.h>
```

src/posix/platform/settings.cpp
```diff
@@ -245,7 +291,15 @@ otError otPlatSettingsGet(otInstance *aInstance, uint16_t aKey, int aIndex, uint
...
 otError otPlatSettingsSet(otInstance *aInstance, uint16_t aKey, const uint8_t *aValue, uint16_t aValueLength)
 {
-    int swapFd = -1;
+    int     swapFd = -1;
+    otError error  = OT_ERROR_NONE;
+
+#if OPENTHREAD_POSIX_CONFIG_SECURE_SETTINGS_ENABLE
+    if (isCriticalKey(aKey))
+    {
+        ExitNow(error = otPlatSecureSettingsSet(aInstance, aKey, aValue, aValueLength));
```
wgtdkp, line 300: We have been using `VerifyOrDie` in this function, should we also use `VerifyOrDie` here or we change all `VerifyOrDie` to `VerifyOrExit`?

src/posix/platform/include/openthread/platform/secure_settings.h
simonlingoogle, line 60: where is the implementation of `otPlatSecureSettingsInit`?

simonlingoogle:
> > This CL adds a function otPlatSettingsSetCriticalKeys() to the include/openthread/platform/settings.h to tell the platform which setting keys are critical setting keys. Another method is to move all the setting key definitions from src/core/common/settings.hpp to the file include/openthread/platform/settings.h. Platform defines which settings will be securely stored. Which method do you think is better? @abtink @bukepo
> 
> I prefer that we keep settings key definitions in OpenThread core. Here is how I see this:
> 
> * It's the OT core that is defining what needs to be saved and assigning setting key numbers and knows what is the data that ie being saved.
> * The platform layer should be simply saving the key/values as a dictionary/map and treat the value as blobs of data and not be aware of what the keys and/or values are.
> * So if we need to decide that certain data need to be saved differently (e.g in secure storange), I think it would be better for it to be decided by OT core.
> * I like the current model in the PR that OT core specifies which keys are critical through a new `otPlat` API (platform layer can then choose to do something different for those keys but does not need to know what the keys are).
> * I don't think delegating the decision directly to platform would make it easier:
>   
>   * If we delegate this to platform layer, when a new settings key is added in OT core, we need to check all platform implementations and ensure they are doing the right thing for new key being added.

Agree that there are advantages to let OT core decide which keys are secure/critical.

However, I think there maybe other solutions that is simpler than the current PR:
1. Add a new OT API: otSettingsIsCriticalKey(uint16_t key) so that platform code can call this API to tell if each key is critical or not. This achieve the exact same purpose of this PR, but without any code change in posix platform code (if it's not implementing secure settings yet). 
2. Add a new set of platform functions (listed below) and let OT core call these functions for critical keys when `SECURE_SETTINGS=ON`. The advantage is that when users enable `SECURE_SETTINGS`, they automatically knows what function they need to implement to have this feature. 
  - otPlatSettingsSecureGet
  - otPlatSettingsSecureSet
  - otPlatSettingsAdd
  - otPlatSettingsDelete
  - otPlatSettingsWipe

I would prefer method 1 since it's very easy to implement. It's actually quite similar to exposing all key definitions to public header, so is this PR.

Thoughts? @zhanglongxia @abtink @wgtdkp

src/posix/platform/include/openthread/platform/secure_settings.h
simonlingoogle: BTW, I think the method name should be `platformSecureSettingXxx`

zhanglongxia: The APIs `otPosixSecureSettingsXxx` will be implemented by the Vendor. They may locate in another repo. So I think it is better to use the name `otPosixSecureSettingsXxx`. Comments?

include/openthread/platform/settings.h
zhanglongxia: Will update it.

src/posix/platform/settings.cpp
zhanglongxia: If we do define the macro `OPENTHREAD_POSIX_CONFIG_SECURE_SETTINGS_ENABLE `, the source code will not call the secure settings related APIs. Why do we still need to include the header?

src/posix/platform/settings.cpp
zhanglongxia: The APIs defined in `include/openthread/platform/settings.h` allows us to return error codes. The API `otPlatSecureSettingsXxx` will be implemented by the Vendor. There may be various implementations of the APIs `otPlatSecureSettingsXxx`. So I think it is better to let the Vendor to process the error code. Comments?

src/posix/platform/include/openthread/platform/secure_settings.h
zhanglongxia: These APIs should be implemented by the Vendor in their repo not in OpenThread.

src/posix/platform/settings.cpp
```diff
@@ -194,6 +233,13 @@ otError otPlatSettingsGet(otInstance *aInstance, uint16_t aKey, int aIndex, uint
     const off_t size   = lseek(sSettingsFd, 0, SEEK_END);
     off_t       offset = lseek(sSettingsFd, 0, SEEK_SET);
 
+#if OPENTHREAD_POSIX_CONFIG_SECURE_SETTINGS_ENABLE
```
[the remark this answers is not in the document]
zhanglongxia: This feature is designed for the new products. At the beginning of designing new products, the value `SECURE_SETTINGS ` should be determined. It should not be changed during the OTA. If you really need to change the `SECURE_SETTINGS` during OTA, you need to add extra code to process this case.

zhanglongxia:
@simonlingoogle 
I considered the Method 1 you mentioned before. The problem is that the API `otSettingsIsCriticalKey(uint16_t key)` should be provided by the OT Core rather than the platform. We need to add a new header file `include/openthread/settings.h` to declare this  function. Which distributes the settings related APIs into two header files. In terms of code structure, it is more complex than current CL. 

For the method 2 you mentioned, we need to add lots of `otPlatSettingsXxx` APIs. For this CL, the reason I added the `otPlatSecureSettingsXxx` APIs is that  I want to reuse the current posix settings implementation and implement the secure settings in our product. These APIs are restricted in posix platform layer. Assuming I don’t need to reuse the current settings' implementation in other platforms, there is no need to define `otPlatSecureSettingsXxx` APIs. I can just implement the `settings.cpp` by myself and link the `settings.cpp` to OT. So I prefer not to add new APIs "otPlatSettingsSecureXxx" to `include/openthread/platform/settings.h`.

simonlingoogle:
> @simonlingoogle
> I considered the Method 1 you mentioned before. The problem is that the API `otSettingsIsCriticalKey(uint16_t key)` should be provided by the OT Core rather than the platform. We need to add a new header file `include/openthread/settings.h` to declare this function. Which distributes the settings related APIs into two header files. In terms of code structure, it is more complex than current CL.

In the current PR, platforms that want to use secure settings need to implement `otPlatSettingsSetCriticalKeys`, which seems a little bit bothersome because platforms need to copy the keys to local static memory and the max number is not known beforehand. However, such memory footprint and copying is not necessary since the keys are const and already defined in OT core. So, it is straightforward to me to just add another OT API to determine if each key is secure. Yet, I don't think adding another header file is too complex to have. 

> For the method 2 you mentioned, we need to add lots of `otPlatSettingsXxx` APIs. For this CL, the reason I added the `otPlatSecureSettingsXxx` APIs is that I want to reuse the current posix settings implementation and implement the secure settings in our product. These APIs are restricted in posix platform layer. Assuming I don’t need to reuse the current settings' implementation in other platforms, there is no need to define `otPlatSecureSettingsXxx` APIs. I can just implement the `settings.cpp` by myself and link the `settings.cpp` to OT. So I prefer not to add new APIs "otPlatSettingsSecureXxx" to `include/openthread/platform/settings.h`.

Current PR adds a number of `otPosixSecureSettingXxx` methods to the POSIX platform header. I think it's not that different from adding a number of `otPlatSecureSettingXxx` methods to the OT platform header. But defining `otPlatSecureSettingXxx` methods in OT platform header would also help other (non-POSIX) platform.

Thoughts?

zhanglongxia:
> > @simonlingoogle
> > I considered the Method 1 you mentioned before. The problem is that the API `otSettingsIsCriticalKey(uint16_t key)` should be provided by the OT Core rather than the platform. We need to add a new header file `include/openthread/settings.h` to declare this function. Which distributes the settings related APIs into two header files. In terms of code structure, it is more complex than current CL.
> 
> In the current PR, platforms that want to use secure settings need to implement `otPlatSettingsSetCriticalKeys`, which seems a little bit bothersome because platforms need to copy the keys to local static memory and the max number is not known beforehand. However, such memory footprint and copying is not necessary since the keys are const and already defined in OT core. So, it is straightforward to me to just add another OT API to determine if each key is secure. Yet, I don't think adding another header file is too complex to have.

The comment of the function `otPlatSettingsSetCriticalKeys` indicates that the OT will not release the memory pointed by the @aKeys, which means the platform doesn't need to copy the keys to local memory. It can use a  pointer to point to the @aKeys to record the memory,  you can refer to the implementation of the function `otPlatSettingsSetCriticalKeys`. Compared to adding a file and adding a function, it may be easier to add a function. @abtink  @bukepo Thoughts?

> 
> > For the method 2 you mentioned, we need to add lots of `otPlatSettingsXxx` APIs. For this CL, the reason I added the `otPlatSecureSettingsXxx` APIs is that I want to reuse the current posix settings implementation and implement the secure settings in our product. These APIs are restricted in posix platform layer. Assuming I don’t need to reuse the current settings' implementation in other platforms, there is no need to define `otPlatSecureSettingsXxx` APIs. I can just implement the `settings.cpp` by myself and link the `settings.cpp` to OT. So I prefer not to add new APIs "otPlatSettingsSecureXxx" to `include/openthread/platform/settings.h`.
> 
> Current PR adds a number of `otPosixSecureSettingXxx` methods to the POSIX platform header. I think it's not that different from adding a number of `otPlatSecureSettingXxx` methods to the OT platform header. But defining `otPlatSecureSettingXxx` methods in OT platform header would also help other (non-POSIX) platform.
> 
This may be a benefit. @abtink  @bukepo Thoughts?

> Thoughts?

abtink:
> The comment of the function otPlatSettingsSetCriticalKeys indicates that the OT will not release the memory pointed by the @akeys, which means the platform doesn't need to copy the keys to local memory. It can use a pointer to point to the @akeys to record the memory, you can refer to the implementation of the function otPlatSettingsSetCriticalKeys. Compared to adding a file and adding a function, it may be easier to add a function. @abtink @bukepo Thoughts?

I read the comments above. I tend to agree with @zhanglongxia  on this and favor the model of having a  `otPlatSettingSetCriticalKeys` that we call from core to inform platform about critical keys compared to providing a function for the platform call to get info (e.g., `otSettingsIsCriticalKey()`).
- In general, the current platform layer abstraction, 
  - We ususally have functions that we call from core to inform platform layer (or ask it to do something).
  -  In cases where the platform calls a function defined by OT core, it's usually a callback (like `otPlatRadioTxDone()` rather a function for the platform to get info/state.
- The main reason that I prefer the model where we call platform rather the platform calling us (by "we" I usually mean OT core code 😄 ...) is to avoid possible incorrect use by platform developers. 
   - We have a rule that we expect all the `ot` public/platform APIs to be called from same platform OS context (same thread or same task in RTOS, etc).
   - If we provide a function for platform layer to call `otSettingsIsCriticalKey()`, we enable the possible incorrect use (i.e., it being called from a different OS context than the one OT is running on). This can lead to hard-to-find and tricky racing bugs.
   - On other hand, if we have a function that we call, we avoid such potential incorrect use.

> 
> Current PR adds a number of otPosixSecureSettingXxx methods to the POSIX platform header. I think it's not that different from adding a number of otPlatSecureSettingXxx methods to the OT platform header. But defining otPlatSecureSettingXxx methods in OT platform header would also help other (non-POSIX) platform.

This sounds reasonable to me well. It's a more involved approach compared to `otPlatSettingSetCriticalKeys()` (would be a bigger change).
One idea to consider related to this: How about instead of adding new `otPlatSecureSettingXxx` APIs, we change the existing `otPlatSettings` functions to have an extra parameter to indicate whether the key is secure (i.e. use secure secttings for it) or not.  
- The upside I see is that we won't need to add a bunch of new `otPlatSecureSettings` APIs (which look similar to `otPlatSetting` ones).
- The downside is that all platform implementations need to be updated (at minimum to ignore the new parameter)

Thoughts?

bukepo:
I tend to expose the settings keys because they are something persisted in platform layer. They are not some private numbers that OpenThread core can change, otherwise, the data persisted will not be able to parse. By exposing the key values, we stabilize them as part of the OpenThread API.

With the settings keys exposed, it's not necessary to add another parameter indicating the content should be secured or not. While the platform can determine what to secure based on its own restrictions. Different platforms may have different security strategy for OpenThread data. Some may just secure the masterkey, some with more resources and security/privacy considerations may choose to secure storing other data such as extend addresses, topology information and so on.

What do you think?

abtink:
> I prefer exposing the settings keys because they are something persisted in platform layer. They are not some private numbers that OpenThread core can change, otherwise, the data persisted will not be able to parse. By exposing the key values, we stabilize them as part of the OpenThread API.
> 
> With the settings keys exposed, it's not necessary to add another parameter indicating the content should be secured or not. While the platform can determine what to secure based on its own restrictions. Different platforms may have different security strategy for OpenThread data. Some may just secure the masterkey, some with more resources and security/privacy considerations may choose to secure storing other data such as extend addresses, topology information and so on.
> 
> What do you think?

The main downside I see with this model is that when/if we add a new setting key in OT core, we need to go and update/check all platform implementations to ensure they are doing the right thing for the newly added key (e.g. whether they would save it in secure setting or not).

I agree that delegating it to platform sounds simple, but I think as a longer term solution it would be better to have  such decisions in OT core to make it common for all platforms.

src/core/common/settings.hpp
abtink: How about declaring this as static constant variable, i.e. `static const uint16_t kCritialKeys[]` here? and then define the array values in `cpp` file? This way, we know for sure, the array would be in flash and there would be single instance of it.

The `const` member variable model should work fine but (most) toolchains may not optimize it to use flash (i.e. they would use a variable in RAM and then copy the constant array values to it from constructor). Technically a `const` member variable is saved in RAM and can be initialized from constructor and cannot be changed afterwards.

zhanglongxia:
As discuss with @bukepo and @abtink in the meeting, I will expose all the setting keys to the `openthread/platform/settings.h` and add a function `otPlatSettingSetCriticalKeys` to `openthread/platform/settings.h`. For the keys specified by the `otPlatSettingSetCriticalKeys`, the platform should securely store them. For other keys, the platform decides whether to store them securely.

include/openthread/platform/settings.h
```diff
@@ -51,6 +51,29 @@ extern "C" {
...
  */
 
+/**
+ * This enumeration defines the keys of settings.
+ *
+ * Note: When adding a new setings key, if the settings corresponding to the key contains security sensitive
+ *       information, the developer MUST add the key to the array `kCriticalKeys`.
+ *
+ */
+enum Key
```
simonlingoogle, line 61:
```suggestion
enum Key : uint16_t
```

include/openthread/platform/settings.h
```diff
@@ -67,6 +90,18 @@ void otPlatSettingsInit(otInstance *aInstance);
...
+ * This function sets the critical keys that should be stored in the secure area.
+ *
+ * Note that the memory pointed by @p aKeys MUST not be released before @p aInstance is destroyed.
+ *
+ * @param[in]  aInstance    The OpenThread instance structure.
+ * @param[in]  aKeys        A pointer to an array containing the list of critical keys.
+ * @param[in]  aKeysLength  The number of entries in the @p aKeys array.
+ *
+ */
+void otPlatSettingsSetCriticalKeys(otInstance *aInstance, const uint16_t *aKeys, uint16_t aKeysLength);
```
simonlingoogle, line 103: nit: 
Why do we still need this method if we have exposed all keys? @bukepo @zhanglongxia 
The platform can freely decide which keys are critical?

src/core/common/settings.cpp
```diff
@@ -209,7 +224,8 @@ void Settings::Wipe(void)
 
 otError Settings::SaveOperationalDataset(bool aIsActive, const MeshCoP::Dataset &aDataset)
 {
-    otError error = Save(aIsActive ? kKeyActiveDataset : kKeyPendingDataset, aDataset.GetBytes(), aDataset.GetSize());
```
simonlingoogle, line 212: nit:
define enum in OT core like ?
```
enum {
  kKeyActiveDataset = OT_SETTINGS_KEY_ACTIVE_DATASET, 
  ...
}
```

include/openthread/platform/settings.h
zhanglongxia: You can refer to the @abtink's comments above https://github.com/openthread/openthread/pull/6125#issuecomment-771775435.

simonlingoogle:
> > Current PR adds a number of otPosixSecureSettingXxx methods to the POSIX platform header. I think it's not that different from adding a number of otPlatSecureSettingXxx methods to the OT platform header. But defining otPlatSecureSettingXxx methods in OT platform header would also help other (non-POSIX) platform.
> 
> This sounds reasonable to me well. It's a more involved approach compared to `otPlatSettingSetCriticalKeys()` (would be a bigger change).
> One idea to consider related to this: How about instead of adding new `otPlatSecureSettingXxx` APIs, we change the existing `otPlatSettings` functions to have an extra parameter to indicate whether the key is secure (i.e. use secure secttings for it) or not.
> 
> * The upside I see is that we won't need to add a bunch of new `otPlatSecureSettings` APIs (which look similar to `otPlatSetting` ones).


Currently, the PR uses `otPlatSettingSetCriticalKeys` to pass secure keys to posix platform and defines a number of `otPosixSecureSettingsXxxx` methods that are to be defined by POSIX implementations. 
If so, why don't we just define a number of `otPlatSecureSettingXxxx` methods and call them directly in OT core? I think we are doing a lot of extra work here just to have the lesser effects of the `otPlatSecureSettingXxx` methods because we didn't cover platforms other than POSIX.


> * The downside is that all platform implementations need to be updated (at minimum to ignore the new parameter)

Platform implementations only need to be updated if they choose to support `SECURE_SETTINGS`. Actually, this gives platform clear guidance on what they need to implement. 

---

That being said, I think we should either **simply expose the keys** as @bukepo has proposed or being more involved by defining `otPlatSecureSettingXxx` methods and calling them in OT core. 

Thoughts? @bukepo @zhanglongxia @abtink

zhanglongxia:
> > > Current PR adds a number of otPosixSecureSettingXxx methods to the POSIX platform header. I think it's not that different from adding a number of otPlatSecureSettingXxx methods to the OT platform header. But defining otPlatSecureSettingXxx methods in OT platform header would also help other (non-POSIX) platform.
> > 
> > 
> > This sounds reasonable to me well. It's a more involved approach compared to `otPlatSettingSetCriticalKeys()` (would be a bigger change).
> > One idea to consider related to this: How about instead of adding new `otPlatSecureSettingXxx` APIs, we change the existing `otPlatSettings` functions to have an extra parameter to indicate whether the key is secure (i.e. use secure secttings for it) or not.
> > 
> > * The upside I see is that we won't need to add a bunch of new `otPlatSecureSettings` APIs (which look similar to `otPlatSetting` ones).
> 
> Currently, the PR uses `otPlatSettingSetCriticalKeys` to pass secure keys to posix platform and defines a number of `otPosixSecureSettingsXxxx` methods that are to be defined by POSIX implementations.
> If so, why don't we just define a number of `otPlatSecureSettingXxxx` methods and call them directly in OT core? I think we are doing a lot of extra work here just to have the lesser effects of the `otPlatSecureSettingXxx` methods because we didn't cover platforms other than POSIX.
> 
> > * The downside is that all platform implementations need to be updated (at minimum to ignore the new parameter)
> 
> Platform implementations only need to be updated if they choose to support `SECURE_SETTINGS`. Actually, this gives platform clear guidance on what they need to implement.
> 
> That being said, I think we should either **simply expose the keys** as @bukepo has proposed or being more involved by defining `otPlatSecureSettingXxx` methods and calling them in OT core.
> 
> Thoughts? @bukepo @zhanglongxia @abtink

For the security reason, we expose the critical keys through the function `otPlatSettingSetCriticalKeys()`. Suppose we add a new critical key to OT, the platform layer may be implemented in other repo by other developers. The developer may not know we have added a new critical key, they will not update their code to store the new critical settings, which may cause security issues. All the keys specified by the function `otPlatSettingSetCriticalKeys()` should be securely stored. The platform layer can use a common method to process them.

For the privacy reason, the developer may want to store the network info to a secure area, so we expose all the keys to let the platform determines which settings should be processed specially. 

These are the reasons why this CL includes these two methods.

include/openthread/platform/settings.h
zhanglongxia: I tried this method, it seems it is not a common method. Refer to the [link](https://stackoverflow.com/questions/9853633/how-to-change-the-integer-type-used-by-an-enum-c). I have removed the key word `Key`to follow other `enum` code style used in other header files.

simonlingoogle: SGTM. Didn't see it's a C header.

simonlingoogle (approved):
LGTM 👍 

It's a workable solution.

include/openthread/platform/settings.h
```diff
@@ -51,6 +51,29 @@ extern "C" {
...
+/**
+ * This enumeration defines the keys of settings.
+ *
+ * Note: When adding a new setings key, if the settings corresponding to the key contains security sensitive
+ *       information, the developer MUST add the key to the array `kCriticalKeys`.
+ *
+ */
+enum
+{
+    OT_SETTINGS_KEY_ACTIVE_DATASET  = 0x0001, ///< Active Operational Dataset
```
wgtdkp, line 63:
```suggestion
    OT_SETTINGS_KEY_ACTIVE_DATASET  = 0x0001, ///< Active Operational Dataset.
```
And other lines at below.

wgtdkp (approved):
LGTM 👍

bukepo (approved):

include/openthread/platform/settings.h
```diff
@@ -51,6 +51,29 @@ extern "C" {
...
+enum
+{
+    OT_SETTINGS_KEY_ACTIVE_DATASET  = 0x0001, ///< Active Operational Dataset.
+    OT_SETTINGS_KEY_PENDING_DATASET = 0x0002, ///< Pending Operational Dataset.
+    OT_SETTINGS_KEY_NETWORK_INFO    = 0x0003, ///< Thread network information.
+    OT_SETTINGS_KEY_PARENT_INFO     = 0x0004, ///< Parent information.
+    OT_SETTINGS_KEY_CHILD_INFO      = 0x0005, ///< Child information.
+    OT_SETTINGS_KEY_RESERVED        = 0x0006, ///< Reserved (previously auto-start).
+    OT_SETTINGS_KEY_SLAAC_IID_SECRET_KEY =
+        0x0007, ///< Secret key used by SLAAC module for generating semantically opaque IID.
```
abtink, line 70: Trying to make comment shorter to hopefully fit in same line:
```suggestion
        0x0007, ///< SLAAC key to generate semantically opaque IID.
```

abtink (approved):
Looks great. Thanks. 👍

include/openthread/platform/settings.h
zhanglongxia: Updated.

librasungirl (approved):

jwhui (approved):

PR #6126, opened 2021-01-29
[github-actions] use cmake to build nrf in size action

Irving-cl:
This PR updates nrf building in size check to use cmake with the latest cmake support. The PR also refines the shell implementation.

PR #6127, opened 2021-01-29
[build] cmake nrf add output directory for bin and lib

Irving-cl:
This PR makes binaries and libraries generated at specified directory in nrf cmake build. I think putting all the binaries and libraries in one directory is better. (Currently the `build` dir is chaotic.)

Note that currently it only affects cmake build for nrf. Other platforms like simulation are not affected. They are not updated because there are many hard-coded path in our test scripts. Updating them now would certainly break our github actions. I would update them as well later.

This PR is a prerequisite for #6126.


simonlingoogle (approved):
LGTM :+1: 
thanks

jwhui:
@Irving-cl , we are in the process of migrating example platforms into separate repos. How about we apply this enhancement to https://github.com/openthread/ot-nrf528xx instead?

Irving-cl:
@jwhui 

Ok, got it.

Irving-cl:
[ot-nrf528xx/#3](https://github.com/openthread/ot-nrf528xx/pull/3)

PR #6130, opened 2021-01-30
[net-data] add 'NetworkData::Service' module

abtink:

**[net-data] add 'NetworkData::Service' module**

This commit adds a new module `network_data_service` which includes
the "Thread Network Data Service" entry definitions (e.g., service
number and the associated server data format for each service type).
It also adds `NetworkData::Service::Manager` which manages the Network
Data Service entries and provides helper methods to other core modules
to easily add/remove and/or parse/find info about specific services.
For example, this commit adds a method to get info about the
list of SRP servers (their IPv6 address and port number) from the
"SRP Server Service" entries in the Thread Network Data.

**[bbr-local] simplify 'Add/RemoveService()'**

This commit updates `AddServcie()` and `RemoveService()` in
`BackboneRouter::Local` class such that the methods themselves
after adding/removing the service entry in local network data
inform `NetworkData::Notifier` to send the update to leader.
This helps simplify their use.


src/core/thread/network_data_service.hpp
```diff
@@ -0,0 +1,362 @@
...
+     * the first entry @p aIterator should be cleared (e.g., a new instance of `Iterator` or calling `Clear()` method).
+     *
+     * @param[out]   aInfo         A reference to `SrpServer::Info` to return the next SRP server info.
+     * @param[inout] aIterator     A reference to an iterator.
+     *
+     * @retval OT_ERROR_NONE       Successfully got the next SRP server info. @p aInfo and @p aIterator are updated.
+     * @retval OT_ERROR_NOT_FOUND  No more SRP server entries in Network Data.
+     *
+     */
+    otError GetNextSrpServerInfo(SrpServer::Info &aInfo, Iterator &aIterator) const;
```
wgtdkp, line 346: nit: put `inout` parameters before `out`
```suggestion
    otError GetNextSrpServerInfo( Iterator &aIterator, SrpServer::Info &aInfo) const;
```

wgtdkp (approved):
This looks really great! thanks @abtink  :+1:

src/core/thread/network_data_service.cpp
```diff
@@ -0,0 +1,187 @@
...
+    aConfig.mMlrTimeout          = rvalServerData->GetMlrTimeout();
+
+    error = OT_ERROR_NONE;
+
+exit:
+    return error;
+}
+
+#endif // (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
+
```
librasungirl, line 135: add macro `#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE || OPENTHREAD_CONFIG_SRP_CLIENT_ENABLE` for `GetNextSrpServerInfo` method?

abtink: I actually had such an `#if` check but removed it.

The reason is I think it is good to have these APIs/feature to discover list of server from network data independent of SRP client/server features (even if the OT SRP client and/or server are not being used).

For example,  a system/product may have its own SRP client implementation (part of its own sw stack) and not want to use the included SRP client from OT but it can be helpful for it to be able to discover the SRP servers (from network data).

librasungirl (approved):

simonlingoogle (approved):
LGTM 👍 
thanks

jwhui (approved):

PR #6132, opened 2021-02-02
[cli] ensure to init 'mInstance' before other member variables

abtink:
This commit changes the order of member variables in `Cli` class
ensuring `mInstance` is initialized before other member variables.
This then allows the `mInstance` to be safely used from the
constructor of other `Cli` sub-types.


abtink:
The code size reduction is nice side bonus here 😄 
My guess/theory is that since we use `mInstance` in a lot of places  it helps that it is now the first member variable in the list.

simonlingoogle (approved):
LGTM :+1: 

make sense to me.

jwhui (approved):

PR #6138, opened 2021-02-04
[locator] remove OwnerLocator class

simonlingoogle:
This commit removes the `OwnerLocator` class since it's actually useless.

`Timer` and `Tasklet` inherited from `OwnerLocator` to so that they can use `GetOwner<Type>()` to find the owner object. However, these two classes already inherited from `InstanceLocator`, thus can use `Get<Type>()` to get any component of the OT instance.


abtink (approved):
LGTM. Thank. 👍

jwhui:
@simonlingoogle , is this still "work-in-progress"?

simonlingoogle:
> @simonlingoogle , is this still "work-in-progress"?

No. Wip removed. Thanks .?

jwhui (approved):

PR #6134, opened 2021-02-03
[srp-client] new methods to get server address, set callback

abtink:
This commit adds new methods/APIs in `Srp::Client` class:
`GetServerAddress()` to get the SRP server's socket address that is
being used by the client, `IsRunning()` to check whether the SRP
client is started and running or not, and `SetCallback()` to allow
user to set the callback at any point (before/after call to
`Start()`). It also adds a new `srp client` CLI sub-command to get 
the server address and/or its port number.

PR #6140, opened 2021-02-05
[srp-client] adding auto-start feature

abtink:
This commit adds auto-start SRP client feature. Config option
`OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` adds support for
this feature in SRP client allowing auto-start mode to be enabled or
disabled. When enabled, the SRP client will monitor the Thread Network
Data for SRP Server Service entries and automatically start and stop
the client when an SRP server is detected. If multiple SRP servers are
found, a random one will be selected. If the selected SRP server is no
longer detected (not longer present in the Thread Network Data), the
SRP client will be stopped and then it may switch to another SRP
server (if available).

This commit also adds support for the new feature in CLI and adds a
new test `test_srp_auto_start_mode.py` covering the basic behavior of
this feature.

-----

This PR uses the commit from PR #6134. Please only review and check the 
last commit. Thanks.

### DIFF
--- a/include/openthread/srp_client.h
+++ b/include/openthread/srp_client.h
@@ -172,6 +172,21 @@ typedef void (*otSrpClientCallback)(otError                    aError,
                                     void *                     aContext);
 
 /**
+ * This function pointer type defines the callback used by SRP client to notify user when it is auto-started or stopped.
+ *
+ * This is only used when auto-start feature `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` is enabled.
+ *
+ * This callback is invoked when auto-start mode is enabled and the SRP client is either automatically started or
+ * stopped.
+ *
+ * @param[in] aSeverSockAddress    A non-NULL pointer indicates SRP sever was started and pointer will give the
+ *                                 selected server socket address. A NULL pointer indicates SRP sever was stopped.
+ * @param[in] aContext             A pointer to an arbitrary context (provided when callback was registered).
+ *
+ */
+typedef void (*otSrpClientAutoStartCallback)(const otSockAddr *aServerSockAddr, void *aContext);
+
+/**
  * This function starts the SRP client operation.
  *
  * SRP client will prepare and send "SRP Update" message to the SRP server once all the following conditions are met:
@@ -244,6 +259,60 @@ const otSockAddr *otSrpClientGetServerAddress(otInstance *aInstance);
  *
  */
 void otSrpClientSetCallback(otInstance *aInstance, otSrpClientCallback aCallback, void *aContext);
+
+/**
+ * This function enables the auto-start mode.
+ *
+ * This is only available when auto-start feature `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` is enabled.
+ *
+ * Config option `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_DEFAULT_MODE` specifies the default auto-start mode (whether
+ * it is enabled or disabled at the start of OT stack).
+ *
+ * When auto-start is enabled, the SRP client will monitor the Thread Network Data for SRP Server Service entries
+ * and automatically start and stop the client when an SRP server is detected.
+ *
+ * If multiple SRP servers are found, a random one will be selected. If the selected SRP server is no longer
+ * detected (not longer present in the Thread Network Data), the SRP client will be stopped and then it may switch
+ * to another SRP server (if available).
+ *
+ * When the SRP client is explicitly started through a successful call to `otSrpClientStart()`, the given SRP server
+ * address in `otSrpClientStart()` will continue to be used regardless of the state of auto-start mode and whether the
+ * same SRP server address is discovered or not in the Thread Network Data. In this case, only an explicit
+ * `otSrpClientStop()` call will stop the client.
+ *
+ * @param[in] aInstance   A pointer to the OpenThread instance.
+ * @param[in] aCallback   A callback to notify when client is auto-started/stopped. Can be NULL if not needed.
+ * @param[in] aContext    A context to be passed when invoking @p aCallback.
+ *
+ */
+void otSrpClientEnableAutoStartMode(otInstance *aInstance, otSrpClientAutoStartCallback aCallback, void *aContext);
+
+/**
+ * This function disables the auto-start mode.
+ *
+ * This is only available when auto-start feature `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` is enabled.
+ *
+ * Disabling the auto-start mode will not stop the client if it is already running but the client stops monitoring
+ * the Thread Network Data to verify that the selected SRP server is still present in it.
+ *
+ * Note that a call to `otSrpClientStop()` will also disable the auto-start mode.
+ *
+ * @param[in] aInstance   A pointer to the OpenThread instance.
+ *
+ */
+void otSrpClientDisableAutoStartMode(otInstance *aInstance);
+
+/**
+ * This function indicates the current state of auto-start mode (enabled or disabled).
+ *
+ * This is only available when auto-start feature `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` is enabled.
+ *
+ * @param[in] aInstance   A pointer to the OpenThread instance.
+ *
+ * @returns TRUE if the auto-start mode is enabled, FALSE otherwise.
+ *
+ */
+bool otSrpClientIsAutoStartModeEnabled(otInstance *aInstance);
 
 /**
  * This function gets the lease interval used in SRP update requests.
--- a/src/cli/README_SRP.md
+++ b/src/cli/README_SRP.md
@@ -57,15 +57,7 @@ fe80:0:0:0:a8cd:6e23:df3d:4193
 Done
 > srp server enable
 Done
-> netdata show
-Prefixes:
-Routes:
-Services:
-44970 5d c002 s 8400
-Done
 ```
-
-The SRP Server listening UDP port which is `c002`(`49154`) is included in the Server Data (listed by the `netdata show` command).
 
 ### Start SRP Client
 
@@ -100,11 +92,28 @@ Done
 Done
 > srp client service add my-service _ipps._tcp 12345
 Done
-srp client start fded:5114:8263:1fe1:68bc:ec03:c1ad:9325 49154
+> srp client autostart enable
 Done
 ```
 
+The last command enables the auto-start mode on the client which then monitors the network data to discover available SRP servers within the Thread network and automatically starts the client.
+
+Alternatively, the client can be started manually using the `srp client start`.
+
+The SRP Server listening UDP port (which is `c002`(`49154`) in the example below) can be found from the Server Data (listed by the `netdata show` command).
+
 Make sure the SRP Server address & port are used for the `srp client start` command.
+
+```bash
+> netdata show
+Prefixes:
+Routes:
+Services:
+44970 5d c002 s 8400
+Done
+srp client start fded:5114:8263:1fe1:68bc:ec03:c1ad:9325 49154
+Done
+```
 
 ### Verify the service status
 
--- a/src/cli/README_SRP_CLIENT.md
+++ b/src/cli/README_SRP_CLIENT.md
@@ -5,6 +5,7 @@
 Usage : `srp client [command] ...`
 
 - [help](#help)
+- [autostart](#autostart)
 - [callback](#callback)
 - [host](#host)
 - [keyleaseinterval](#keyleaseinterval)
@@ -12,6 +13,7 @@ Usage : `srp client [command] ...`
 - [server](#server)
 - [service](#service)
 - [start](#start)
+- [state](#state)
 - [stop](#stop)
 
 ## Command Details
@@ -24,6 +26,7 @@ Print SRP client help menu.
 
 ```bash
 > srp client help
+autostart
 callback
 help
 host
@@ -31,7 +34,33 @@ keyleaseinterval
 leaseinterval
 service
 start
+state
 stop
+Done
+```
+
+### autostart
+
+Usage `srp client autostart [enable|disable]`
+
+Enable/Disable auto start mode in SRP client. This command requires `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` feature to be enabled.
+
+Get the current autostart mode.
+
+```bash
+> srp client autostart
+Disabled
+Done
+```
+
+Set the autostart mode.
+
+```bash
+> srp client autostart enable
+Done
+
+> srp client autostart
+Enabled
 Done
 ```
 
@@ -303,10 +332,22 @@ Done
 
 Usage: `srp client start <serveraddr> <serverport>`
 
-Start the SRP client with a given server IPv6 address and port number
+Start the SRP client with a given server IPv6 address and port number.
 
 ```bash
 > srp client start fd00::d88a:618b:384d:e760 4724
+Done
+```
+
+### state
+
+Usage: `srp client state`
+
+Indicates the state of SRP client, i.e., whether it is enabled or disabled.
+
+```bash
+> srp client state
+Enabled
 Done
 ```
 
--- a/src/core/config/srp_client.h
+++ b/src/core/config/srp_client.h
@@ -46,6 +46,31 @@
 #endif
 
 /**
+ * @def OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE
+ *
+ * Define to 1 to enable SRP Client auto-start feature and its APIs.
+ *
+ * When enabled, the SRP client can be configured to automatically start when it detects the presence of an SRP server
+ *  (by monitoring the Thread Network Data for SRP Server Service entries).
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE
+#define OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE 1
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_DEFAULT_MODE
+ *
+ * Define the default mode (enabled or disabled) of auto-start mode.
+ *
+ * This config is applicable/used only when `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` is enabled.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_DEFAULT_MODE
+#define OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_DEFAULT_MODE 0
+#endif
+
+/**
  * @def OPENTHREAD_CONFIG_SRP_CLIENT_DOMAIN_NAME_API_ENABLE
  *
  * Define to 1 for the SRP client implementation to provide APIs that get/set the domain name.
--- a/src/core/net/srp_client.cpp
+++ b/src/core/net/srp_client.cpp
@@ -33,8 +33,10 @@
 #include "common/instance.hpp"
 #include "common/locator-getters.hpp"
 #include "common/logging.hpp"
+#include "common/random.hpp"
 #include "common/settings.hpp"
 #include "common/string.hpp"
+#include "thread/network_data_service.hpp"
 
 #if OPENTHREAD_CONFIG_SRP_CLIENT_ENABLE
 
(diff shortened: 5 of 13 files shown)


tests/scripts/thread-cert/node.py
```diff
@@ -835,6 +835,34 @@ def srp_client_stop(self):
...
 
+    def srp_client_get_state(self):
+        cmd = 'srp client state'
+        self.send_command(cmd)
+        return self._expect_command_output(cmd)[0]
+
+    def srp_client_get_auto_start_mode(self):
+        cmd = 'srp client autostart'
+        self.send_command(cmd)
+        return self._expect_command_output(cmd)[0]
```
simonlingoogle, line 846:
```suggestion
        return self._expect_command_output(cmd)[0] == 'Enabled'
```

nit: better return True or False

tests/scripts/thread-cert/test_srp_auto_start_mode.py
```diff
@@ -0,0 +1,104 @@
...
+        # 1. Enable auto start mode on client and check that selected sever
+        #
+
+        self.assertEqual(client.srp_client_get_state(), "Disabled")
+        client.srp_client_enable_auto_start_mode()
+        self.assertEqual(client.srp_client_get_auto_start_mode(), "Enabled")
+        self.simulator.go(2)
+
+        self.assertEqual(client.srp_client_get_state(), "Enabled")
+        self.assertEqual(client.srp_client_get_server_port(), client.get_srp_server_port())
```
simonlingoogle, line 100: Test the behavior of client when the `Service TLV` is removed or updated in Network Data?

src/core/net/srp_client.cpp
```diff
@@ -1428,6 +1460,75 @@ void Client::HandleTimer(void)
...
+        // current selected server with the new entry. This approach
+        // results in a uniform/same probability of selection among
+        // all server entries.
+
+        if ((numServers == 1) || (Random::NonCrypto::GetUint16InRange(0, numServers) == 0))
+        {
+            selectedServer = server;
+        }
+
+        if (IsRunning() && mAutoStartDidSelectServer && (GetServerAddress() == server.mSockAddr))
```
simonlingoogle, line 1511: Do we prefer the current server if it's working here? This behavior is not stated in the comments above.

src/core/net/srp_client.cpp
```diff
@@ -1428,6 +1460,75 @@ void Client::HandleTimer(void)
...
+
+    VerifyOrExit(mAutoStartModeEnabled);
+
+    // If the client is not running we check if there is any SRP sever
+    // info in Network Data and select one randomly and then start the
+    // client. If the client is already running with a server that was
+    // selected by the auto-start feature, we verify that the selected
+    // server is still present in the Network Data.
+
+    VerifyOrExit(!IsRunning() || mAutoStartDidSelectServer);
```
simonlingoogle, line 1493: Just curious when could it happen that `IsRunning() && !mAutoStartDidSelectServer` can be satisfied. 
I think it should be that the user started the SRP client manually and then enables `auto start` mode when the SRP client is already running. 
This seems to be a mixed use of `auto start` and `manual` modes, which could make things more complex than necessary. 

I think it would be slightly better if there is a distinct boundary between the two modes:
- `auto start` mode: always use auto selected server from network data
- `manual` mode: always use the manually specified server 

That would being following restrictions to the user APIs:
- User can not enable `auto start` mode when the SRP client is already running (manually started). User must stop the manually started SRP first. 
- User can not manually start the SRP server when `auto start` mode is enabled, even when there is no SRP server TLV in the network data (i.e. SRP client is not running). If I understand correctly, this is allowed by this PR which I think is quite a confusing behavior. 

Thoughts?

wgtdkp:
@abtink better if we can update the SRP example in [README_SRP.md](https://github.com/openthread/openthread/blob/master/src/cli/README_SRP.md).

src/core/net/srp_client.cpp
abtink: > when could it happen that `IsRunning() && !mAutoStartDidSelectServer` can be satisfied.

You mention one way, but the more common way is that auto-start is enabled by default (which is the feature we support through an OT config) and then explicit call to `Start()` client with a specific server. 

---

The API policy/behavior we have is as follows:
- `Start()` can be called to start the client if it is not already running.
- Auto-start mode can be enabled/disabled at any time.
- When client is started by an explicit call to `Start()` it continues to use SRP server specified by `Start()`: 
   - (i.e., we won't check the SRP server to be present in list of servers discovered by auto-start feature).

I feel this is a good and simple set of API policies.

We can add more conditions/restrictions that for example require user to disable auto-start mode before calling `Start()` and/or not allow user to change auto-start mode after `Start()` but there are some downsides to this:
- We need to add more code to check for these combinations and return error to user (increase code size for not so common behaviors).
- It makes the documentation/description of the APIs (the API use policy) more complex.
- It's true that user can use the APIs in uncommon ways (like enable `auto-start` after calling `Start()` which is pointless) but the way I see it is that as long as we behave consistently and correctly, we don't need to explicitly protect user from using uncommon API combinations. 

-------

For longer-term, we may add use-cases where we expand the auto-start feature and want to use/enable it even after the client is started by `Start()` call. 
   - A possible future enehancment is to use auto-start mode to (also) help switch the SRP server if we fail to get a response from the currently selected SRP server after many attempts/retries. 
      - In the current implementation when we select a server, we stick with it even if we never receive a response from it. We switch the server when the SRP server entry/info is not longer seen in Thread network-data.
      - Basically the idea of this enhancement is that if we don't get a response from the current SRP server (after many retries), we try to see if there is any other SRP server options (from network-data) and switch to it.
   - We may want to use/enable this behavior for both cases where the SRP server is selected by auto-start or through `Start()` call.

tests/scripts/thread-cert/node.py
abtink: I noticed the common pattern in `node.py` seems to be to use `get_{something}` over `is_{something}_enabled` and return the value from the corresponding CLI command. 

So to be consistent I would prefer to continue to use `get_auto_start_mode()` over `is_auto_start_mode_enabled()` and using `get_mode()` I think better to keep it string over bool return type?

src/core/net/srp_client.cpp
abtink: I think it's already mentioned in the comment at the top before the `while()` loop in this method.
```
    // If the client is not running we check if there is any SRP sever
    // info in Network Data and select one randomly and then start the
    // client. If the client is already running with a server that was
    // selected by the auto-start feature, we verify that the selected
    // server is still present in the Network Data.
```

tests/scripts/thread-cert/test_srp_auto_start_mode.py
abtink: In new pushed commit the test is expanded to cover the behavior when server is disabled/gone and when we have two servers and one goes away and client switches.

abtink:
> @abtink better if we can update the SRP example in README_SRP.md.

Good idea. Thanks. I changed the `README_SRP.md` example to use "auto-start" as the main way.

include/openthread/srp_client.h
```diff
@@ -245,6 +260,60 @@ const otSockAddr *otSrpClientGetServerAddress(otInstance *aInstance);
...
+ *
+ * Config option `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_DEFAULT_MODE` specifies the default auto-start mode (whether
+ * it is enabled or disabled at the start of OT stack).
+ *
+ * When auto-start is enabled, the SRP client will monitor the Thread Network Data for SRP Server Service entries
+ * and automatically start and stop the client when an SRP server is detected.
+ *
+ * If multiple SRP servers are found, a random one will be selected. If the selected SRP server is no longer
+ * detected (not longer present in the Thread Network Data), the SRP client will be stopped and then it may switch
+ * to another SRP server (if available).
```
wgtdkp, line 276: One thing we need to consider is that the DNS-SD client (I believe we want to add `auto-server-discovery` to the DNS-SD client as well) and SRP client should choose the same SRP server when multiple servers are present at the same time. Otherwise, the DNS-SD client may not be able to discover the services registered by the SRP client.

I think we need a srp-server-selecting module/method that can be shared by at least the SRP client and DNS-SD client.

src/core/net/srp_client.cpp
```diff
@@ -1428,6 +1460,75 @@ void Client::HandleTimer(void)
...
 }
 
+#if OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE
+
+void Client::EnableAutoStartMode(AutoStartCallback aCallback, void *aContext)
+{
+    mAutoStartCallback = aCallback;
+    mAutoStartContext  = aContext;
+
+    VerifyOrExit(!mAutoStartModeEnabled);
```
wgtdkp, line 1470:
```suggestion
    VerifyOrExit(!mAutoStartModeEnabled);

    mAutoStartCallback = aCallback;
    mAutoStartContext  = aContext;
```

src/core/net/srp_client.cpp
```diff
@@ -1428,6 +1460,75 @@ void Client::HandleTimer(void)
...
+    mAutoStartModeEnabled = true;
+    CheckAutoStart();
+
+exit:
+    return;
+}
+
+void Client::CheckAutoStart(void)
```
wgtdkp, line 1478: nit: the method name sounds like it won't change any states. May rename to names like `UpdateServerForAutoStart`?

wgtdkp (approved):
Thanks @abtink , LGTM :+1: 
Just a few nits.

include/openthread/srp_client.h
abtink: In my view, SRP and DND-SD while related are two different protocols and functions. SRP client is used by device to publish its services. DSN-SD client is used to discover services within the home/network. It can be true that we may end up running SRP server and DNS server on same device (on a BR), but we should not assume or rely that they are always on same device (specially from client perspectives). These are two different network entities. 

I think it is the job of a DNS name server to be able to answer questions related to all domains/services it is managing (how it knows and collects the service info, e.g. through interacting with SRP servers or mDNS or ... should not matter to clients that are using it).

That said, we already have API to get the selected server currently being used by SRP client (when auto-start is being used).  So if a use-case wants to assume SRP server is also DNS name server it can be done by next layer code but I don't think it's good idea to add  code/logic within OT itself that assumes this.

src/core/net/srp_client.cpp
abtink: The order here is intentional.
We want to allow user  to change the callback even when auto-strat is already enabled/running. 
Note that this method is `void` return so we allow it to be called anytime (no error returned if already running).

src/core/net/srp_client.cpp
abtink: This method is checking whether we need to auto-start or stop  and is called whenever there is a event/change.  I will rename to `ProessAutoStart()`.

include/openthread/srp_client.h
simonlingoogle: Agree with both of you. I think the details are yet to be defined by Thread BR Spec.

src/core/net/srp_client.cpp
simonlingoogle: > > when could it happen that `IsRunning() && !mAutoStartDidSelectServer` can be satisfied.
> 
> You mention one way, but the more common way is that auto-start is enabled by default (which is the feature we support through an OT config) and then explicit call to `Start()` client with a specific server.

Do you mean a explicit call to `Start()` with one server chosen from the Network Data, meaning the user want to implement custom rules for choosing a server?

> We can add more conditions/restrictions that for example require user to disable auto-start mode before calling `Start()` and/or not allow user to change auto-start mode after `Start()` but there are some downsides to this:
> 
> * We need to add more code to check for these combinations and return error to user (increase code size for not so common behaviors).

Not necessarily to increase code size? Currently we have code logic checking whether the server is actually chosen by `auto-start` feature even when `auto-start` feature is ON. These logic won't be necessary in the proposed strategy. 

> * It makes the documentation/description of the APIs (the API use policy) more complex.
> * It's true that user can use the APIs in uncommon ways (like enable `auto-start` after calling `Start()` which is pointless) but the way I see it is that as long as we behave consistently and correctly, we don't need to explicitly protect user from using uncommon API combinations.

In this current strategy, enabling `auto-start` after calling `Start()` would return OK. I think this might be a little bit confusing to users. It seems that `auto-start` is successfully, but the client is still **NOT** using servers in Network Data.

> For longer-term, we may add use-cases where we expand the auto-start feature and want to use/enable it even after the client is started by `Start()` call.
> 
> * A possible future enehancment is to use auto-start mode to (also) help switch the SRP server if we fail to get a response from the currently selected SRP server after many attempts/retries.
>   
>   * In the current implementation when we select a server, we stick with it even if we never receive a response from it. We switch the server when the SRP server entry/info is no longer seen in Thread network-data.
>   * Basically the idea of this enhancement is that if we don't get a response from the current SRP server (after many retries), we try to see if there is any other SRP server options (from network-data) and switch to it.
> * We may want to use/enable this behavior for both cases where the SRP server is selected by auto-start or through `Start()` call.

I think this is actually a good point. I think the `auto-start` feature can simply be a automatic way of adding servers, but the strategy of selecting server is independent on whether the servers are input from Network Data or user configuration. User would also be able to add multiple servers for the SRP client to chose the best server. 

That being said, I am fine that we work out a better solution in the future.

simonlingoogle (approved):
LGTM :+1:

include/openthread/srp_client.h
```diff
@@ -171,6 +171,21 @@ typedef void (*otSrpClientCallback)(otError                    aError,
...
+ * This function pointer type defines the callback used by SRP client to notify user when it is auto-started or stopped.
+ *
+ * This is only used when auto-start feature `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` is enabled.
+ *
+ * This callback is invoked when auto-start mode is enabled and the SRP client is either automatically started or
+ * stopped.
+ *
+ * @param[in] aSeverSockAddress    A non-NULL pointer indicates SRP sever was started and pointer will give the
+ *                                 selected server socket address. A NULL pointer indicates SRP sever was stopped.
+ * @param[in] aContex              A pointer to an arbitrary context (provided when callback was registered).
```
jwhui, line 184:
```suggestion
 * @param[in] aContext             A pointer to an arbitrary context (provided when callback was registered).
```

include/openthread/srp_client.h
```diff
@@ -245,6 +260,60 @@ const otSockAddr *otSrpClientGetServerAddress(otInstance *aInstance);
...
+ *
+ * This is only available when auto-start feature `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` is enabled.
+ *
+ * Config option `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_DEFAULT_MODE` specifies the default auto-start mode (whether
+ * it is enabled or disabled at the start of OT stack).
+ *
+ * When auto-start is enabled, the SRP client will monitor the Thread Network Data for SRP Server Service entries
+ * and automatically start and stop the client when an SRP server is detected.
+ *
+ * If multiple SRP servers are found, a random one will be selected. If the selected SRP server is no longer
```
jwhui, line 274: This is still TBD, but if there are multiple SRP servers, it may be necessary to register with all - for example, in multi-homed environments.

include/openthread/srp_client.h
```diff
@@ -245,6 +260,60 @@ const otSockAddr *otSrpClientGetServerAddress(otInstance *aInstance);
...
+
+/**
+ * This function disables the auto-start mode.
+ *
+ * This is only available when auto-start feature `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` is enabled.
+ *
+ * Disabling the auto-start mode will not stop the client if it is already running but the client stops monitoring
+ * the Thread Network Data to verify that the selected SRP server is still present in it.
+ *
+ * Note that a call to `otSrpClinetStop()` will also disable the auto-start mode.
```
jwhui, line 298:
```suggestion
 * Note that a call to `otSrpClientStop()` will also disable the auto-start mode.
```

src/cli/README_SRP.md
```diff
@@ -100,10 +92,27 @@ Done
...
+
+Alternatively, the client can be started manually using the `srp client start`.
+
+```bash
+> netdata show
+Prefixes:
+Routes:
+Services:
+44970 5d c002 s 8400
+Done
```
jwhui, line 109: Move this and the network data description line below above the "Alternatively, ..." line?

src/cli/README_SRP_CLIENT.md
```diff
@@ -24,17 +26,44 @@ Print SRP client help menu.
...
+state
 stop
 Done
 ```
 
+### autostart
+
+Usage `srp client autostart [enable|disable]`
+
+Enable/Disable auto start mode in SRP client. This command requires `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` feature to enabled.
```
jwhui, line 46:
```suggestion
Enable/Disable auto start mode in SRP client. This command requires `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` feature to be enabled.
```

src/cli/README_SRP_CLIENT.md
```diff
@@ -24,17 +26,44 @@ Print SRP client help menu.
...
+
+Get current autostart mode
```
jwhui, line 48:
```suggestion
Get current autostart mode.
```

src/cli/README_SRP_CLIENT.md
```diff
@@ -24,17 +26,44 @@ Print SRP client help menu.
...
+
+```bash
+> srp client autostart
+Disabled
+Done
+```
+
+Set the autostart mode
```
jwhui, line 56:
```suggestion
Set the autostart mode.
```

src/cli/README_SRP_CLIENT.md
```diff
@@ -303,13 +332,25 @@ Done
...
 ```bash
 > srp client start fd00::d88a:618b:384d:e760 4724
 Done
 ```
 
+### state
+
+Usage: `srp client state`
+
+Indicates the state of SRP client, i.e., whether it is enabled/running or disabled.
```
jwhui, line 346:
```suggestion
Indicates the state of SRP client, i.e., whether it is enabled or disabled.
```

src/core/config/srp_client.h
```diff
@@ -45,6 +45,31 @@
...
+#ifndef OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE
+#define OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE 1
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_DEFAULT_MODE
+ *
+ * Define the default mode (enabled or disabled) of auto-start mode.
+ *
+ * This conifg is applicable/used only when `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` is enabled.
```
jwhui, line 66:
```suggestion
 * This config is applicable/used only when `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` is enabled.
```

src/core/net/srp_client.hpp
```diff
@@ -276,16 +276,71 @@ class Client : public InstanceLocator, private NonCopyable
...
+    otError Start(const Ip6::SockAddr &aServerSockAddr) { return Start(aServerSockAddr, kRequesterUser); }
 
     /**
      * This method stops the SRP client operation.
      *
      * This method stops any further interactions with the SRP server. Note that it does not remove or clear host info
      * and/or list of services. It marks all services to be added/removed again once the client is started again.
      *
+     * If `OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE` (auto-start feature) is enabled, a call to this method
+     * also disabled the auto-start mode.
```
jwhui, line 288:
```suggestion
     * also disables the auto-start mode.
```

tests/scripts/thread-cert/test_srp_auto_start_mode.py
```diff
@@ -0,0 +1,151 @@
...
+import command
+import thread_cert
+
+# Test description:
+#   This test verifies SRP client auto-start functionality that SRP client can
+#   correctly discover and connect to SRP server.
+#
+# Topology:
+#
+#   ClIENT (leader) -- SERVER1 (router)
```
jwhui, line 42:
```suggestion
#   CLIENT (leader) -- SERVER1 (router)
```

include/openthread/srp_client.h
abtink: The current SRP client implementation supports registering items (service/host info) with a single server.

We can consider extending it to register with multiple servers in parallel but we may then need to re-think some aspect of its design (and its APIs). For example, we probably need to maintain a separate state per service/host item for each SRP server.

jwhui (approved):

PR #6144, opened 2021-02-05
[spinel] add domain name spinel property

XPAWAN:
This commit adds new domain name spinel property for NCP.
    
Such NCP will have also a new capability
SPINEL_CAP_NET_THREAD_1_2.


src/ncp/ncp_base_ftd.cpp
```diff
@@ -331,6 +331,24 @@ template <> otError NcpBase::HandlePropertySet<SPINEL_PROP_MAC_MAX_RETRY_NUMBER_
     return error;
 }
 
+template <> otError NcpBase::HandlePropertyGet<SPINEL_PROP_THREAD_DOMAIN_NAME>(void)
```
jwhui, line 334: Wrap with `#if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)`

abtink (approved):
Just the same comment/fix as @jwhui.
Otherwise, LGTM. Thanks 👍

src/ncp/ncp_base_ftd.cpp
XPAWAN: Done

jwhui (approved):

PR #6154, opened 2021-02-08
[efr32] update for compatibility with Gecko SDK 3.1.1

lmnotran:
This PR updates the EFR32 platform-abstraction layer to work with GSDK 3.1.1. It also has changes needed to handle #6123 

# Change Log
- memory.c
  - Implemented `otPlatCAlloc()` and `otPlatFree()` to handle https://github.com/openthread/openthread/pull/6123
- GSDK mbedtls was updated and moved to util/third_party/crypto/mbedtls
- mbedtls_config_autogen.h was regenerated
- system.c
  - NVIC initialization is now done via a simple call to sl_device_init_nvic()
- alarm.c
  - Stubs for 1.2 related platform timing functions in alarm.c
- ieee802154mac.h
  - Standardization of 802.15.4 header values and constants
- uart.c
  - Reduce the number of vulnerable windows where interrupts disabled by RAIL will cause dropped characters. The OpenThread task now polls the UART buffer for arrived data and reads it out. The USART0 IRQ handler takes care of waking up the task when data has arrived.
- radio.c
  - Implementation of Wi-Fi coexistence and antenna diversity features, and corresponding updates to run-time phy select configurations.
  - Support for enhanced frame pending and early frame pending for Thread 1.2 low power features (temporarily disabled until 1.2 support is available)
  - Implementation of enhanced ACKs with IEs for CSL.
  - Remove FIFO queue that was maintaining unique identifiers (contents of received packet: PHR+MAC header) This was being used to tell the upper MAC whether the frame pending bit was set in an ACK for a received packet. Changed the implementation so we save the state to a reserved bit in the held packet (in the MAC header) within the receive completion ISR, and retrieve the value in processNextRxPacket() to pass to where the upper MAC expects it.


romacdon (approved):
LGTM

romacdon:
@jwhui is there something holding up this PR from being merged?  I see there is 1 failure in the checks that are run.  I don't think thats related to the changes included in the PR though.

jwhui:
> @jwhui is there something holding up this PR from being merged? I see there is 1 failure in the checks that are run. I don't think thats related to the changes included in the PR though.

@romacdon , nothing blocking - apologies for the delay.

jwhui (approved):

PR #6155, opened 2021-02-08
[dns-server] handle multiple questions

simonlingoogle:
This commit enhances the DNS-SD server to handle multiple questions. 

Specifically for DNS service resolving query with SRV+TXT questions, this commit would reply with SRV+TXT+AAAA RRs, so that the service instance can be fully resolved with one query. 

- [x] Add tests to verify that DNS-SD client service browsing and resolving works. @abtink

#### More implementation details 
- This commit  enhances `NameCompressInfo` so that question names can be compressed together with RR names.
  - **Full compression** is limited to only one instance / service / host name.
- For any RR type in the question, this commit will exclude it in additional RRs. For example, if the query contains a AAAA question, the response won't include  any additional AAAA RR, but will only have AAAA answer(s). The purpose is to avoid duplicate RRs in answer section and additional section. 

Depends on #6183 to fix a POSIX console bug.

### DIFF
--- a/src/core/net/dnssd_server.cpp
+++ b/src/core/net/dnssd_server.cpp
@@ -113,141 +113,134 @@ exit:
 void Server::ProcessQuery(Message &aMessage, Message &aResponse, const Header &aRequestHeader)
 {
     Header           responseHeader;
-    uint16_t         readOffset, nameSerializeOffset;
+    uint16_t         readOffset;
     Question         question;
-    uint16_t         qtype;
     char             name[Dns::Name::kMaxNameSize];
-    otError          error = OT_ERROR_NONE;
-    NameCompressInfo compressInfo;
+    NameCompressInfo compressInfo(kDefaultDomainName);
+    Header::Response response          = Header::Response::kResponseSuccess;
+    otError          error             = OT_ERROR_NONE;
+    uint8_t          resolveAdditional = kResolveAdditionalAll;
 
     // Setup initial DNS response header
     responseHeader.Clear();
-    responseHeader.SetResponseCode(Header::kResponseSuccess);
     responseHeader.SetType(Header::kTypeResponse);
     responseHeader.SetMessageId(aRequestHeader.GetMessageId());
 
     // Validate the query
     VerifyOrExit(aRequestHeader.GetQueryType() == Header::kQueryTypeStandard,
-                 responseHeader.SetResponseCode(Header::kResponseNotImplemented));
-    VerifyOrExit(!aRequestHeader.IsTruncationFlagSet(), responseHeader.SetResponseCode(Header::kResponseFormatError));
-    VerifyOrExit(aRequestHeader.GetQuestionCount() == 1,
-                 responseHeader.SetResponseCode(Header::kResponseNotImplemented));
+                 response = Header::kResponseNotImplemented);
+    VerifyOrExit(!aRequestHeader.IsTruncationFlagSet(), response = Header::kResponseFormatError);
+    VerifyOrExit(aRequestHeader.GetQuestionCount() > 0, response = Header::kResponseFormatError);
 
-    // Read the query name and question
     readOffset = sizeof(Header);
-    VerifyOrExit(OT_ERROR_NONE == Dns::Name::ReadName(aMessage, readOffset, name, sizeof(name)),
-                 responseHeader.SetResponseCode(Header::kResponseFormatError));
-    VerifyOrExit(OT_ERROR_NONE == aMessage.Read(readOffset, question),
-                 responseHeader.SetResponseCode(Header::kResponseFormatError));
 
-    // Add the question to the response, and save the serialize offset of name
-    nameSerializeOffset = aResponse.GetLength();
-    SuccessOrExit(error = AddQuestionToResponse(name, question, aResponse, responseHeader));
-
-    // Further validate the query
-    qtype = question.GetType();
-    VerifyOrExit(qtype == ResourceRecord::kTypePtr || qtype == ResourceRecord::kTypeSrv ||
-                     qtype == ResourceRecord::kTypeTxt || qtype == ResourceRecord::kTypeAaaa,
-                 responseHeader.SetResponseCode(Header::kResponseNotImplemented));
-
-    VerifyOrExit(question.GetClass() == ResourceRecord::kClassInternet ||
-                     question.GetClass() == ResourceRecord::kClassAny,
-                 responseHeader.SetResponseCode(Header::kResponseNotImplemented));
-
-    // Prepare the information for name compression
-    VerifyOrExit(OT_ERROR_NONE == PrepareCompressInfo(qtype, name, nameSerializeOffset, compressInfo),
-                 responseHeader.SetResponseCode(Header::kResponseNameError));
-
-    // Resolve the question
-    SuccessOrExit(error = ResolveQuestion(name, question, responseHeader, aResponse, compressInfo));
-
-    otLogInfoDns("[server] TRANSACTION=0x%04x, QUESTION=[%s %d %d], RCODE=%d, ANSWER=%d, ADDITIONAL=%d",
-                 aRequestHeader.GetMessageId(), name, question.GetClass(), question.GetType(),
-                 responseHeader.GetResponseCode(), responseHeader.GetQuestionCount(),
-                 responseHeader.GetAdditionalRecordCount());
-exit:
-    if (error != OT_ERROR_NONE)
+    // Check and append the questions
+    for (uint16_t i = 0; i < aRequestHeader.GetQuestionCount(); i++)
     {
-        otLogWarnDns("[server] failed to handle DNS query: %s", otThreadErrorToString(error));
+        NameComponentsOffsetInfo nameComponentsOffsetInfo;
 
+        VerifyOrExit(OT_ERROR_NONE == Dns::Name::ReadName(aMessage, readOffset, name, sizeof(name)),
+                     response = Header::kResponseFormatError);
+        VerifyOrExit(OT_ERROR_NONE == aMessage.Read(readOffset, question), response = Header::kResponseFormatError);
+        readOffset += sizeof(question);
+
+        uint16_t qtype = question.GetType();
+
+        VerifyOrExit(qtype == ResourceRecord::kTypePtr || qtype == ResourceRecord::kTypeSrv ||
+                         qtype == ResourceRecord::kTypeTxt || qtype == ResourceRecord::kTypeAaaa,
+                     response = Header::kResponseNotImplemented);
+
+        VerifyOrExit(question.GetClass() == ResourceRecord::kClassInternet ||
+                         question.GetClass() == ResourceRecord::kClassAny,
+                     response = Header::kResponseNotImplemented);
+
+        VerifyOrExit(OT_ERROR_NONE == FindNameComponents(name, compressInfo.GetDomainName(), nameComponentsOffsetInfo),
+                     response = Header::kResponseNameError);
+
+        switch (question.GetType())
+        {
+        case ResourceRecord::kTypePtr:
+            VerifyOrExit(nameComponentsOffsetInfo.IsServiceName(), response = Header::kResponseNameError);
+            break;
+        case ResourceRecord::kTypeSrv:
+            VerifyOrExit(nameComponentsOffsetInfo.IsServiceInstanceName(), response = Header::kResponseNameError);
+            resolveAdditional &= ~kResolveAdditionalSrv;
+            break;
+        case ResourceRecord::kTypeTxt:
+            VerifyOrExit(nameComponentsOffsetInfo.IsServiceInstanceName(), response = Header::kResponseNameError);
+            resolveAdditional &= ~kResolveAdditionalTxt;
+            break;
+        case ResourceRecord::kTypeAaaa:
+            VerifyOrExit(nameComponentsOffsetInfo.IsHostName(), response = Header::kResponseNameError);
+            resolveAdditional &= ~kResolveAdditionalAaaa;
+            break;
+        default:
+            ExitNow(response = Header::kResponseNameError);
+        }
+
+        SuccessOrExit(error = AppendQuestion(name, question, aResponse, compressInfo));
+        responseHeader.SetQuestionCount(i + 1);
+    }
+
+    // Answer the questions
+    readOffset = sizeof(Header);
+    for (uint16_t i = 0; i < aRequestHeader.GetQuestionCount(); i++)
+    {
+        uint8_t resolveKind = kResolveAnswer;
+
+        VerifyOrExit(OT_ERROR_NONE == Dns::Name::ReadName(aMessage, readOffset, name, sizeof(name)),
+                     response = Header::kResponseFormatError);
+        VerifyOrExit(OT_ERROR_NONE == aMessage.Read(readOffset, question), response = Header::kResponseFormatError);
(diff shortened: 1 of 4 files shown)


simonlingoogle:
This PR failed in `posix-cli` checks. 

```
> 2: dns browse _ipps._tcp.default.service.arpa. fd00:db8:0:0:553a:d5e3:bcb2:f71f 53

<t.service.arpa. fd00:db8:0:0:553a:d5e3:bcb2:f71f 53
> DNS browse response for _ipps._tcp.default.service.arpa.
ins2
    Port:22222, Priority:2, Weight:2, TTL:7175
    Host:host2.default.service.arpa.
    HostAddress:fd00:db8:0:0:9982:902c:fc44:10cd TTL:7175
    TXT-Data:(len:1) [00] TTL:7175
ins1
    Port:11111, Priority:1, Weight:1, TTL:7170
    Host:host1.default.service.arpa.
    HostAddress:fd00:db8:0:0:bd9a:6033:459f:704 TTL:7170
    TXT-Data:(len:1) [00] TTL:7170
Done
```

Seems the POSIX CLI was unable to echo the command (i.e. `dns browse _ipps._tcp.default.service.arpa. fd00:db8:0:0:553a:d5e3:bcb2:f71f 53`) correctly because the command was executed successfully with correct output. I tend to believe this is not an issue of this PR, but an issue of the POSIX CONSOLE CLI.

Do you have any clue on this? @bukepo

src/core/net/dnssd_server.cpp
```diff
@@ -113,166 +113,176 @@ void Server::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessag
...
+            VerifyOrExit(nameComponentsOffsetInfo.IsServiceInstanceName(), response = Header::kResponseNameError);
+            resolveAdditional &= ~kResolveAdditionalSrv;
+            break;
+        case ResourceRecord::kTypeTxt:
+            VerifyOrExit(nameComponentsOffsetInfo.IsServiceInstanceName(), response = Header::kResponseNameError);
+            resolveAdditional &= ~kResolveAdditionalTxt;
+            break;
+        case ResourceRecord::kTypeAaaa:
+            VerifyOrExit(nameComponentsOffsetInfo.IsHostName(), response = Header::kResponseNameError);
+            resolveAdditional &= ~kResolveAdditionalAaaa;
```
abtink, line 175: If I understand the code correctly the `resolveAdditional` is used to track which RR types need to be added in additional section vs the ones that are already part of the answer section.

Here we go through the questions, and if see there is question for TXT RR, we clear the flag for TXT in `resolveAdditional`, and then later we won't allow adding of TXT record in additional section.

But how can we know if the questions are all for the same/related resource names?

For example, if there is question for SRV for one name  and AAA for an unrelated host name, I think we still want to include an AAAA record in additional section corresponding to host for the SRV record.

Or if the question contains a PTR question and single SRV question, I think we still need to list all services in response to PTR and add a corresponding SRV record for each in the  additional section except the one that may match the SRV question (and is already included in answer section)?

-----

Maybe before adding any RR in the additional section we add code to check in the previously prepared answer section to see if we already have a matching record (same name and type) there and if so we skip it (i.e. not duplicate it in additional section).  We can use `ResourceRecord::FindRecord()` to find it in answer section. Thoughts?

abtink (commented):
Thanks @simonlingoogle . Looks good overall. 
Some comments/questions below.

src/core/net/dnssd_server.cpp
```diff
@@ -113,166 +113,176 @@ void Server::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessag
...
+    // Answer the questions
+    readOffset = sizeof(Header);
+    for (uint16_t i = 0; i < aRequestHeader.GetQuestionCount(); i++)
+    {
+        uint8_t resolveKind = kResolveAnswer;
 
-    SuccessOrExit(error = FindNameComponents(aName, domain, nameComponentsInfo));
+        VerifyOrExit(OT_ERROR_NONE == Dns::Name::ReadName(aMessage, readOffset, name, sizeof(name)),
+                     response = Header::kResponseFormatError);
+        VerifyOrExit(OT_ERROR_NONE == aMessage.Read(readOffset, question), response = Header::kResponseFormatError);
```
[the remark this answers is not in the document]
simonlingoogle: IgnoreError SGTM.

src/core/net/dnssd_server.cpp
simonlingoogle: > If I understand the code correctly the `resolveAdditional` is used to track which RR types need to be added in additional section vs the ones that are already part of the answer section.
>
> Here we go through the questions, and if see there is question for TXT RR, we clear the flag for TXT in `resolveAdditional`, and then later we won't allow adding of TXT record in additional section.

You are totally correct. I used this implementation for two reasons:

- It works just fine for almost all DNS-SD client use cases. The DNS-SD client should never add unrelated names in normal use cases. 
- It is very simple to implement, without having to check which RRs have been appended to avoid duplicates. 
- It is in conformance with DNS-SD RFC because addtional RRs are a `SHOULD` requirement but not a `MUST`. It is fine that DNS-SD server do not add additoinal RRs.
- Another thought is that if the DNS-SD client do use unrelated questions, maybe it's exactly its intention **NOT** to have additional RRs. For example, if a client queries a unrated `AAAA` question, it would show that the client have some special intention for the `AAAA` resolution. This (sophisticated) client can certainly also add the related `AAAA` question itself if it wants to. 
 
> But how can we know if the questions are all for the same/related resource names?

We don't know. But they should be in most cases. 

> For example, if there is question for SRV for one name and AAA for an unrelated host name, I think we still want to include an AAAA record in additional section corresponding to host for the SRV record.
>
> Or if the question contains a PTR question and single SRV question, I think we still need to list all services in response to PTR and add a corresponding SRV record for each in the additional section except the one that may match the SRV question (and is already included in answer section)?

Yes, we should better add additional RRs for these two cases. However, this is a `SHOULD` but not `MUST` according to DNS-SD RFC. 

> Maybe before adding any RR in the additional section we add code to check in the previously prepared answer section to see if we already have a matching record (same name and type) there and if so we skip it (i.e. not duplicate it in additional section). We can use `ResourceRecord::FindRecord()` to find it in answer section. Thoughts?

This is a good implementation idea. However, I prefer the current implementation for the  reasons above. 

Thoughts? @abtink

src/core/net/dnssd_server.hpp
```diff
@@ -176,14 +191,14 @@ class Server : public InstanceLocator, private NonCopyable
...
-        const char *mDomainName;         // The serialized domain name (must NOT be nullptr)
-        const char *mServiceName;        // The serialized service name (must NOT be nullptr for PTR/SRV/TXT queries)
-        const char *mInstanceName;       // The serialized instance name or nullptr (only support one instance name)
-        const char *mHostName;           // The serialized host name or nullptr (only support one host name)
-        uint16_t    mDomainNameOffset;   // Offset of domain name serialization into the response message.
-        uint16_t    mServiceNameOffset;  // Offset of service name serialization into the response message.
-        uint16_t    mInstanceNameOffset; // Offset of instance name serialization into the response message.
-        uint16_t    mHostNameOffset;     // Offset of host name serialization into the response message.
+        const char *const mDomainName;   // The serialized domain name (must NOT be nullptr)
+        const char *      mServiceName;  // The serialized service name (must NOT be nullptr for PTR/SRV/TXT queries)
```
wgtdkp, line 195: nit:
```suggestion
        const char *      mServiceName;  // The serialized service name (MUST NOT be nullptr for PTR/SRV/TXT queries)
```

src/posix/console_cli.cpp
```diff
@@ -95,7 +95,8 @@ void otxConsoleInit(otInstance *aInstance)
     rl_instream           = stdin;
     rl_outstream          = stdout;
     rl_inhibit_completion = true;
-    sReadFd               = fileno(rl_instream);
+    rl_set_screen_size(1024, 1024);
```
wgtdkp, line 98: Add comments for why this is necessary.

wgtdkp (approved):
Thanks @simonlingoogle LGTM :+1:

tests/scripts/thread-cert/node.py
```diff
@@ -2419,6 +2419,87 @@ def dns_resolve(self, hostname, server=None, port=53):
...
+        # HostAddress:0:0:0:0:0:0:0:0 TTL:0
+        # TXT-Data:(len:1) [00] TTL:7155
+        # Done
+
+        m = re.match(
+            r'.*Port:(\d+), Priority:(\d+), Weight:(\d+), TTL:(\d+)\s+Host:(.*?)\s+HostAddress:(\S+) TTL:(\d+)\s+TXT-Data:\(len:\d+\) \[(.*?)\] TTL:(\d+)',
+            '\r'.join(output))
+        if m:
+            port, priority, weight, srv_ttl, hostname, address, aaaa_ttl, txt_data, txt_ttl = m.groups()
+            return {
```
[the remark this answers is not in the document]
simonlingoogle: Note that the caller has already specified the `instance` and `service`. So, it's already known to the caller.

src/core/net/dnssd_server.cpp
```diff
@@ -113,166 +113,176 @@ void Server::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessag
...
-    aCompressInfo.SetDomainNameOffset(aNameSerializeOffset + nameComponentsInfo.mDomainOffset,
-                                      aName + nameComponentsInfo.mDomainOffset);
-
 exit:
-    return error;
+    response = (error == OT_ERROR_NONE) ? response : Header::Response::kResponseServerFailure;
+
+    if (response == Header::Response::kResponseServerFailure)
+    {
+        otLogWarnDns("[server] failed to handle DNS query due to server failure");
```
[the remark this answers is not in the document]
simonlingoogle: Note that `error` might not be properly set  here even if `response == Header::Response::kResponseServerFailure`.

And the `error` would most certainly be `OT_ERROR_NO_BUFS`, so printing it is not that necessary after all.

src/core/net/dnssd_server.cpp
abtink: I am fine going for the current model now.

I think we can consider enhancing this later.  I feel checking the existing RRs in answer section may not be very complicated to add and it would make the implementation more capable. We can consider it later.

abtink (approved):
LGTM 👍

src/core/net/dnssd_server.cpp
```diff
@@ -113,166 +113,176 @@ void Server::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessag
...
 
-otError Server::AddQuestionToResponse(const char *    aName,
-                                      const Question &aQuestion,
-                                      Message &       aResponse,
-                                      Header &        aResponseHeader)
-{
-    otError error = OT_ERROR_NONE;
+        VerifyOrExit(qtype == ResourceRecord::kTypePtr || qtype == ResourceRecord::kTypeSrv ||
+                         qtype == ResourceRecord::kTypeTxt || qtype == ResourceRecord::kTypeAaaa,
+                     response = Header::kResponseNotImplemented);
```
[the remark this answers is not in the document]
jwhui: @simonlingoogle , are we intentionally keeping this check?

src/core/net/dnssd_server.hpp
```diff
@@ -176,14 +191,14 @@ class Server : public InstanceLocator, private NonCopyable
...
-        const char *mInstanceName;       // The serialized instance name or nullptr (only support one instance name)
-        const char *mHostName;           // The serialized host name or nullptr (only support one host name)
-        uint16_t    mDomainNameOffset;   // Offset of domain name serialization into the response message.
-        uint16_t    mServiceNameOffset;  // Offset of service name serialization into the response message.
-        uint16_t    mInstanceNameOffset; // Offset of instance name serialization into the response message.
-        uint16_t    mHostNameOffset;     // Offset of host name serialization into the response message.
+        const char *const mDomainName;       // The serialized domain name
+        const char *      mServiceName;      // The serialized service name (only support one service name)
+        const char *      mInstanceName;     // The serialized instance name or nullptr (only support one instance name)
+        const char *      mHostName;         // The serialized host name or nullptr (only support one host name)
```
jwhui, line 197: nit:
```suggestion
        const char *const mDomainName;       // The serialized domain name.
        const char *      mServiceName;      // The serialized service name (only support one service name).
        const char *      mInstanceName;     // The serialized instance name or nullptr (only support one instance name).
        const char *      mHostName;         // The serialized host name or nullptr (only support one host name).
```

src/core/net/dnssd_server.cpp
```diff
@@ -113,166 +113,170 @@ void Server::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessag
...
     OT_UNUSED_VARIABLE(aResponseMessage);
     OT_UNUSED_VARIABLE(aCompressInfo);
 
-    otError error = OT_ERROR_NONE;
+    Header::Response response = Header::kResponseNameError;
 
 #if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
-    SuccessOrExit(error = ResolveQuestionBySrp(aName, aQuestion, aResponseHeader, aResponseMessage,
-                                               /* aAdditional */ false, aCompressInfo));
+    response = ResolveQuestionBySrp(aName, aQuestion, aResponseHeader, aResponseMessage, aResolveKind, aCompressInfo);
```
jwhui, line 248: nit: the order of `aResolveKind` and `aCompressInfo` are swapped relative to `Server::ResolveQuestion()`. Is this intentional?

src/core/net/dnssd_server.cpp
```diff
@@ -358,31 +362,74 @@ otError Server::AppendAaaaRecord(Message &           aMessage,
...
+    }
+    else
+    {
+        uint8_t  domainStart          = static_cast<uint8_t>(StringLength(aName, Name::kMaxNameSize - 1) -
+                                                   StringLength(aCompressInfo.GetDomainName(), Name::kMaxNameSize - 1));
+        uint16_t domainCompressOffset = aCompressInfo.GetDomainNameOffset();
 
-    return Dns::Name::AppendPointerLabel(aCompressInfo.GetServiceNameOffset(), aMessage);
+        serviceCompressOffset = aMessage.GetLength();
+        aCompressInfo.SetHostNameOffset(serviceCompressOffset, aName);
```
jwhui, line 379: Should this be:
```suggestion
        aCompressInfo.SetServiceNameOffset(serviceCompressOffset, aName);
```

src/core/net/dnssd_server.cpp
simonlingoogle: > I think we can consider enhancing this later. I feel checking the existing RRs in answer section may not be very complicated to add and it would make the implementation more capable. We can consider it later.

Sure.

src/core/net/dnssd_server.cpp
simonlingoogle: No, it's not intentional. I will make them uniform.

src/core/net/dnssd_server.cpp
simonlingoogle: Good catch.

simonlingoogle:
@jwhui 
All checks are passed expect codecov. 
I think it is impossible that this PR decreased overall codecov by 2.26% because the code addition is small and the diff coverage should be very high (> 90%).

jwhui (approved):

PR #6156, opened 2021-02-09
[k32w] Update flash tools and instructions

doru91:
Add instructions for flashing a device using DK6Programmer. This option can be used when no debugging functionality is needed (.e.g: in which case the used has to install MCUXpresso IDE).


doru91:
The 'Simulation' test fails. However, this doesn't seem to be related to the changes introduced by this PR. Could you please advise?

jwhui:
NXP examples have moved to https://github.com/openthread/ot-nxp

PR #6131, opened 2021-01-30
[dns-sd] add CI tests for SRP server & Advertising Proxy

wgtdkp:
This commit includes basic service registration & advertising function verification. More tests including name conflicts and multiple devices cases will be in future PR.

We changed the `border_routing` test scripts directory to `border_router` as there will be more tests about Thread border router but not only the routing feature.

Depends-On: openthread/ot-br-posix#666

PR #6157, opened 2021-02-09
[dns-header] add common TXT record entry parsing API

abtink:
This commit adds a new public `ot` API for parsing an encoded DNS TXT
record data into `TxtEntry` key/value pairs. It also updates the TXT
data iterator model used for going through all TXT entries. The new
iterator definition provides a small buffer to store the key string.
This helps simplify the definition of `otDnsTxtEntry` to use
null-terminated C string for the key. The new parsing APIs are used in
CLI module.

This commit also fixes an issue with `TxtEntry::AppendEntries()` using
incorrect value for the key length.

Finally, this commit adds a test covering the behavior of all
`TxtEntry` related methods (encoding and parsing TXT data) in the unit
test `test_dns`.

Depends-On: openthread/ot-br-posix#700


include/openthread/dns.h
```diff
@@ -91,9 +91,51 @@ typedef struct otDnsTxtEntry
     const char *   mKey;
     const uint8_t *mValue;       ///< The TXT record value or already encoded TXT-DATA (depending on `mKey`).
     uint16_t       mValueLength; ///< Number of bytes in `mValue` buffer.
-    uint8_t mKeyLength; ///< Number of bytes in `mKey` buffer. MUST be set even if `mKey` is a null-terminated string.
 } otDnsTxtEntry;
 
+/**
+ * This structure represents an iterator for TXT record entires (key/value pairs).
+ *
+ * The data fields in this structure are intended for use by OpenThread core and caller should not read or change them.
```
wgtdkp, line 99: Instead of having the documentation tell users don't touch this structure, I think it is better to hide the structure definition as we do for many other types (e.g. `otInstance`, `otSrpServerHost`...).

```c++
typedef struct otDnsTxtEntryIterator otDnsTxtEntryIterator;
// or
typedef void otDnsTxtEntryIterator;
```
Both should work well. Thoughts?

include/openthread/srp_server.h
```diff
@@ -304,20 +304,24 @@ uint16_t otSrpServerServiceGetWeight(const otSrpServerService *aService);
...
  *
- * @retval OT_ERROR_NONE       Successfully found the next TXT entry.
- * @retval OT_ERROR_NOT_FOUND  No subsequent TXT entry exists in the service.
+ * @return The TXT record data length (number of bytes in buffer returned from `otSrpServerServiceGetTxtData()`).
  *
  */
-otError otSrpServerServiceGetNextTxtEntry(const otSrpServerService *aService,
-                                          otDnsTxtIterator *        aIterator,
-                                          otDnsTxtEntry *           aTxtEntry);
+uint16_t otSrpServerServiceGetTxtDataLength(const otSrpServerService *aService);
```
wgtdkp, line 324: We originally have a single API that returns the TXT data pointer and TXT data length together
```c++
const uint8_t *otSrpServerServiceGetTxtData(const otSrpServerService *aService, uint16_t *aTxtLength);
```
I think this is better because:
1. There are no cases that we can/should use the TXT data without knowing its length.
2. It reduces one public API.

Thoughts?

include/openthread/dns.h
abtink: I don't think it would be possible in this case to hide the definition of the `struct otDnsTxtEntryIterator`.

- In this case, we expect the user to define a variable (allocate an instance) of this iterator type and then pass a pointer to it to us (OT core), so we can populate and use it. 
- In other cases where we use opaque types, the user does not need to allocate an instance of the type, so hiding the definition works perfectly.
  - In the other cases, users only deal with pointers to a type that OT core passes to the user. 
- We can always have a pointer to a type which is not fully defined/known by the complier, but as soon as we need to define a variable/instance of the type, the compiler needs to know the full definition (mainly to determine its size).

include/openthread/srp_server.h
abtink: Using one API SGTM. Will change it to use this model later in next push.

include/openthread/dns.h
wgtdkp: yes, you are right.
I didn't realize that the iterator must be passed in from the caller.

src/core/api/srp_server_api.cpp
```diff
@@ -147,12 +147,16 @@ uint16_t otSrpServerServiceGetPriority(const otSrpServerService *aService)
...
+const uint8_t *otSrpServerServiceGetTxtData(const otSrpServerService *aService, uint16_t *aDataLength)
 {
-    return static_cast<const Srp::Server::Service *>(aService)->GetNextTxtEntry(
-        *aIterator, static_cast<Dns::TxtEntry &>(*aTxtEntry));
+    const Srp::Server::Service &service = *static_cast<const Srp::Server::Service *>(aService);
+
+    if (aDataLength != nullptr)
+    {
+        *aDataLength = service.GetTxtDataLength();
+    }
```
wgtdkp, line 157: nit: I think there are no cases that the caller should not get the TXT length, so may just crash if `NULL` is provided. We can document this non-NULL requirement for the `ot` API.
```suggestion
    *aDataLength = service.GetTxtDataLength();
```

wgtdkp (approved):
Thanks @abtink LGTM 👍 

I noticed that the border router testing is failing because the advertising proxy is not updated with the new TXT entry API. (see https://github.com/openthread/openthread/pull/6157/checks?check_run_id=1902556482)

We are using the TXT entry API in the `Advertising Proxy` (https://github.com/openthread/ot-br-posix/blob/master/src/agent/advertising_proxy.cpp#L307).

To fix the CI failure, you can create another OTBR PR that updates to use new TXT entry API and have this PR depend on the OTBR PR. You can take a look at https://github.com/openthread/openthread/pull/6131 if you haven't tried `Depends-On` yet 😄

src/core/api/srp_server_api.cpp
abtink: SGTM.

simonlingoogle (approved):
LGTM :+1:

include/openthread/dns.h
```diff
@@ -53,14 +55,11 @@ extern "C" {
...
+#define OT_DNS_MAX_LABEL_SIZE 64 ///< Maximum label string size (include null char at the end of string).
 
-/**
- * Initializer for otDnsTxtIterator.
- */
-#define OT_DNS_TXT_ITERATOR_INIT 0
+#define OT_DNS_TXT_KEY_MIN_LENGTH 1 ///< Minimum length of TXT record key string (RFC 6763 - section 6.4).
 
-typedef uint16_t otDnsTxtIterator; ///< Used to iterate through the TXT entries.
+#define OT_DNS_TXT_KEY_MAX_LENGTH 9 ///< Maximum length of TXT record key string (RFC 6763 - section 6.4).
```
jwhui, line 62: Note that RFC 6763 specifies this as a SHOULD and not a MUST.

include/openthread/dns.h
```diff
@@ -91,9 +91,51 @@ typedef struct otDnsTxtEntry
...
+ *
+ * The @p aIterator MUST be initialized using `otDnsInitTxtEntryIterator()` before calling this function and the TXT
+ * data buffer used to initialize the iterator MUST persist and remain unchanged. Otherwise the behavior of this
+ * function is undefined.
+ *
+ * @param[in]  aIterator   A pointer to the iterator (MUST NOT be NULL).
+ * @param[out] aEntry      A pointer to a `otDnsTxtEntry` structure to output the parsed/read entry (MUST NOT be NULL).
+ *
+ * @retval OT_ERROR_NONE       The next entry was parsed successfully. @p aEntry is updated.
+ * @retval OT_ERROR_NOT_FOUND  No more entry in the TXT data.
```
jwhui, line 133:
```suggestion
 * @retval OT_ERROR_NOT_FOUND  No more entries in the TXT data.
```

src/core/net/dns_headers.cpp
```diff
@@ -798,38 +798,103 @@ otError ResourceRecord::ReadFrom(const Message &aMessage, uint16_t aOffset)
...
+    while ((cur = GetTxtData() + GetTxtDataPosition()) < GetTxtDataEnd())
+    {
+        length = static_cast<uint8_t>(*cur);
+
+        cur++;
+        VerifyOrExit(cur + length <= GetTxtDataEnd(), error = OT_ERROR_PARSE);
+        IncreaseTxtDataPosition(sizeof(uint8_t) + length);
+
+        // Silently skip over an empty string or if the string starts with
+        // a `=` character (i.e., missing key) - RFC 6762 - section 6.4.
```
jwhui, line 831:
```suggestion
        // a `=` character (i.e., missing key) - RFC 6763 - section 6.4.
```

src/core/net/dns_headers.cpp
```diff
@@ -798,38 +798,103 @@ otError ResourceRecord::ReadFrom(const Message &aMessage, uint16_t aOffset)
...
+        {
+            if (cur[index] == kKeyValueSeparator)
+            {
+                keyBuffer[index++]  = kNullChar; // Increment index to skip over `=`.
+                aEntry.mValue       = reinterpret_cast<const uint8_t *>(&cur[index]);
+                aEntry.mValueLength = length - index;
+                ExitNow();
+            }
+
+            VerifyOrExit(index < kMaxKeyLength, error = OT_ERROR_PARSE);
```
jwhui, line 848: Should we use a different error here? RFC 6763 doesn't strictly forbid longer key lengths. This error should really indicate that this is a limitation of this implementation, not a problem with the TXT record itself.

src/core/net/dns_headers.hpp
```diff
@@ -1051,6 +1051,81 @@ class TxtEntry : public otDnsTxtEntry
...
+        /**
+         * This method parses the TXT data from the `Iterator` and gets the next TXT record entry (key/value pair).
+         *
+         * The `Iterator` instance MUST be initialized using `Init()` before calling this method and the TXT data
+         * buffer used to initialize the iterator MUST persist and remain unchanged.
+         *
+         * @param[out] aEntry          A reference to a `TxtEntry` to output the parsed/read entry.
+         *
+         * @retval OT_ERROR_NONE       The next entry was parsed successfully. @p aEntry is updated.
+         * @retval OT_ERROR_NOT_FOUND  No more entry in TXT data.
```
jwhui, line 1084:
```suggestion
         * @retval OT_ERROR_NOT_FOUND  No more entries in TXT data.
```

src/core/net/dns_headers.hpp
```diff
@@ -1051,6 +1051,81 @@ class TxtEntry : public otDnsTxtEntry
...
+            kIndexTxtLength   = 0,
+            kIndexTxtPosition = 1,
+        };
+
+        const char *GetTxtData(void) const { return reinterpret_cast<const char *>(mPtr); }
+        void        SetTxtData(const uint8_t *aTxtData) { mPtr = aTxtData; }
+        uint16_t    GetTxtDataLength(void) const { return mData[kIndexTxtLength]; }
+        void        SetTxtDataLength(uint16_t aLength) { mData[kIndexTxtLength] = aLength; }
+        uint16_t    GetTxtDataPosition(void) const { return mData[kIndexTxtPosition]; }
+        void        SetTxtDataPoistion(uint16_t aValue) { mData[kIndexTxtPosition] = aValue; }
```
jwhui, line 1102:
```suggestion
        void        SetTxtDataPosition(uint16_t aValue) { mData[kIndexTxtPosition] = aValue; }
```

src/core/net/srp_server.hpp
```diff
@@ -150,17 +150,20 @@ class Server : public InstanceLocator, private NonCopyable
         uint16_t GetPriority(void) const { return mPriority; }
 
         /**
-         * This method returns the next TXT entry of the service instance.
+         * This method returns the TXT record data of the service instance
```
jwhui, line 153:
```suggestion
         * This method returns the TXT record data of the service instance.
```

include/openthread/dns.h
abtink: Good point. Will change this to say: "Recommended maximum length ..."

src/core/net/dns_headers.cpp
abtink: Good point.
Changed it to return `OT_ERROR_NO_BUFS` (updated the `@retval` docs to mention what this error represents).

jwhui: Thinking about this more, I wonder if it might be useful to allow the user to skip over this "too long" key? Maybe they are looking for a key that is not too long. Thoughts?

abtink: I agree. It makes sense to let user skip over long keys.

I had this idea originally of returning things that we can not parse successfully in the `otDnsTxtEntry` with `mKey` set to `nullptr` and then return the encoded value in `mValue` and `mValueLength`. This fits well with how user is expected to provide the encoded txt entry to us.  I did not add it because I thought we cannot assume in general case that all an encoded TXT entry starts with a `length` byte (I think it can be anything in general case).

But in the case of long keys (assuming  DNS-SD format), I think using this approach fits well. I will change the code to use this model for the long key case (i.e. return success with `mKey` set to null and entire value given back to user).

abtink: Added this behavior in the latest pushed commit (also updated unit test to cover the long key case).

abtink:
Rebasing on master, noticed that we need to address the `test_dnssd` and how it parses the TXT info from CLI output due to the change to how the TXT record are now outputted from CLI.

jwhui (approved):

PR #6165, opened 2021-02-11
[bbr] handle case where no PBBR exists in network data

jwhui:


abtink (approved):
LGTM. 👍

src/core/backbone_router/bbr_leader.cpp
```diff
@@ -180,7 +180,10 @@ void Leader::UpdateBackboneRouterPrimary(void)
     State                state;
     uint32_t             origMlrTimeout;
 
-    IgnoreError(Get<NetworkData::Service::Manager>().GetBackboneRouterPrimary(config));
+    if (Get<NetworkData::Service::Manager>().GetBackboneRouterPrimary(config) != OT_ERROR_NONE)
```
simonlingoogle, line 183: I think `Get<NetworkData::Service::Manager>().GetBackboneRouterPrimary` should be responsible of setting `config.mServer16 = Mac::kShortAddrInvalid`?

librasungirl: Yes.  
When there is no BBR Service in Network Data, `mServer` field is set appropriately already.
https://github.com/openthread/openthread/blob/ab7a25faebb9704877dae70befd7233773997328/src/core/thread/network_data_service.cpp#L96

It is assumed that the server data is always valid, is there scenario of invalid bbr server data now?  Maybe for robustness, better to set `mServer16` appropriately in below line
https://github.com/openthread/openthread/blob/ab7a25faebb9704877dae70befd7233773997328/src/core/thread/network_data_service.cpp#L121

After second thought, here is the only calling of `GetBackboneRouterPrimary()`, 
I think we may either removing the return value of `GetBackboneRouterPrimary()` and keep `mServer` is always set appropriately inside, or keep the return value and usage like in this PR and simply `VerifyOrExit(iterator.mServiceTlv != nullptr)` in `GetBackboneRouterPrimary()`

Thoughts?

jwhui: Yes, this is to handle the scenario where the server data is not formatted properly (which can easily happen when using CLI to enter server data).

I agree that we should either:
1. Return void and always set `aConfig` appropriately or,
2. Keep this PR as is.

Thoughts?

simonlingoogle: > 1. Return void and always set `aConfig` appropriately or,
> 2. Keep this PR as is.

I would prefer 1 because it's the original idea and simpler of the two.

simonlingoogle: > Yes, this is to handle the scenario where the server data is not formatted properly (which can easily happen when using CLI to enter server data).

Another thought is that this is actually an issue of dataset CLI commands, not an issue of the BBR dataset. I think it's better to check the dataset raw bytes before saving it (if it's possible to do so).

librasungirl: No strong preference from me. If option2 is chosen, it may not be necessary to assign `mServer16` field when exit here
https://github.com/openthread/openthread/blob/ab7a25faebb9704877dae70befd7233773997328/src/core/thread/network_data_service.cpp#L96

jwhui: > Another thought is that this is actually an issue of dataset CLI commands, not an issue of the BBR dataset. I think it's better to check the dataset raw bytes before saving it (if it's possible to do so).

@simonlingoogle , Network Data can come from any device and implementations.

librasungirl (approved):
👍

simonlingoogle (approved):
LGTM :+1:

PR #6166, opened 2021-02-12
[mle] set router's Thread version in Link Accept handler

konradderda:
Issue:
Information regarding a neighbor's Thread version is lost.

Context:
This problem can be observed during playing with Link Metrics - few methods check whenever chosen neighbor supports Thread 1.2. This information is preserved when a child is promoted to the router role, but after the reset and Link Request - Link Accept sequence it is not stored in the neighbor table.


jwhui (approved):
Thanks! 👍🏼

PR #6167, opened 2021-02-12
[efr32] implement otPlatUartFlush()

lmnotran:
This PR implements `otPlatUartFlush()` for EFR32 with a configurable timeout, defaulting to 500ms. It also allows multiple buffers to be sent to the UARTDRV transmit FIFO queue


romacdon (approved):
LGTM

examples/platforms/efr32/src/openthread-core-efr32-config.h
```diff
@@ -167,4 +167,14 @@
...
 
+/**
+ * @def OPENTHREAD_CONFIG_UART_TX_FLUSH_TIMEOUT_MS
+ *
+ * Maximum time to wait for a flush to complete in otPlatUartFlush().
+ *
+ * Value is in milliseconds
+ *
+ */
+#define OPENTHREAD_CONFIG_UART_TX_FLUSH_TIMEOUT_MS 500
```
jwhui, line 178: Just curious - in what cases do we need to rely on the timeout?

What about renaming to `OPENTHREAD_CONFIG_EFR32_UART_TX_FLUSH_TIMEOUT_MS` to indicate that this is an EFR32-specific config?

romacdon: @lmnotran please make this change.

lmnotran: @jwhui The timeout should realistically never be hit. We just added it so that, just in case something bad happened to the UART, we wouldn't lock up the whole node in an infinite loop

jwhui (approved):

PR #6172, opened 2021-02-15
[dns-client] add 'QueryConfig' and default config

abtink:
This commit adds a new structure `otDnsQueryConfig` which represents a
DNS query configuration defining a set of parameters:

- The DNS server socket address (IPv6 address and port number).
- Response timeout interval (in msec).
- Maximum number of transmit attempts before reporting failure.
- "Recursion desired" flag (if server can resolve recursively).

There is a default query config used by DNS client. When OpenThread
stack starts, the default DNS query config is determined from a set of
OpenThread config options defined in `config/dns_clinet.h` such as
`OPENTHREAD_CONFIG_DNS_CLIENT_DEFAULT_SERVER_IP6_ADDRESS`, etc. New
APIs `otDnsClient{Get}/{Set}DefaultConfig()` are added to allow user
to get/set the current default config during operation.

When issuing a browse, address, or service resolution query, the user
can choose not to provide a query config in which case the current
default config will be used. The user can also provide a full or
partially specified `otDnsQueryConfig` instance, e.g., some of some of
the fields may be left unspecified (value zero). The corresponding
value from the current default config is then used for any unspecified
field in the given config.

This commit also updates the CLI to add `dns config` sub-command and
update other `dns` commands to follow the new API model.


include/openthread/dns_client.h
```diff
@@ -55,9 +55,64 @@ extern "C" {
...
+ * When OpenThread stack starts, the default DNS query config is determined from a set of OT config options such as
+ * `OPENTHREAD_CONFIG_DNS_CLIENT_DEFAULT_SERVER_IP6_ADDRESS`, `_DEFAULT_SERVER_PORT`, `_DEFAULT_RESPONSE_TIMEOUT`, etc.
+ * (please see `config/dns_clinet.h` for all related config options).
+ *
+ * @param[in]  aInstance        A pointer to an OpenThread instance.
+ *
+ * @returns A pointer to the current default config being used by DNS client.
+ *
+ */
+const otDnsQueryConfig *otDnsClientGetDefaultConfig(otInstance *aInstance);
```
simonlingoogle, line 95:
```suggestion
const otDnsQueryConfig *otDnsClientGetConfig(otInstance *aInstance);
```

There seems to be 3 configs:
- The starting default config when OT starts.
- The default config that is configurable 
- The config for each query

Would it be more clear if we change it to only 2 configs ? 

- The starting default config when OT starts.
- The current config that is configurable

Then each browse or resolve query would just use the current config without specifying a config.

include/openthread/dns_client.h
```diff
@@ -220,21 +277,25 @@ typedef struct otDnsServiceInfo
...
-otError otDnsClientBrowse(otInstance *        aInstance,
-                          const otSockAddr *  aServerSockAddr,
-                          const char *        aServiceName,
-                          otDnsBrowseCallback aCallback,
-                          void *              aContext);
+otError otDnsClientBrowse(otInstance *            aInstance,
+                          const char *            aServiceName,
+                          otDnsBrowseCallback     aCallback,
+                          void *                  aContext,
+                          const otDnsQueryConfig *aConfig);
```
simonlingoogle, line 298:
```suggestion
);
```

Remove this argument since the caller can already do everything it want with `otDnsClientSetDefaultConfig`.
I think we should make the API as simple as possible, unless the benefits are clear?

src/cli/README.md
```diff
@@ -746,22 +746,69 @@ Perform an MLE Discovery operation.
...
+Get the default query config used by DNS client.
 
-- DNS server IP: 2001:4860:4860::8888 (Google DNS Server)
-- DNS server port: 53
+The config includes the server IPv6 address and port, response timeout in msec (wait time to rx response), maximum tx attempts before reporting failure, boolean flag to indicate whether the server can resolve the query recursively or not.
+
+```bash
+> dns config
+ServerAddress:[fd00:0:0:0:0:0:0:1]:1234
+ResponseTimeout:5000
```
simonlingoogle, line 758:
```suggestion
ResponseTimeout:5000ms
```

src/cli/README.md
```diff
@@ -746,22 +746,69 @@ Perform an MLE Discovery operation.
...
-Send DNS Query to obtain IPv6 address for given hostname. The latter two parameters have following default values:
+Get the default query config used by DNS client.
 
-- DNS server IP: 2001:4860:4860::8888 (Google DNS Server)
-- DNS server port: 53
+The config includes the server IPv6 address and port, response timeout in msec (wait time to rx response), maximum tx attempts before reporting failure, boolean flag to indicate whether the server can resolve the query recursively or not.
+
+```bash
+> dns config
+ServerAddress:[fd00:0:0:0:0:0:0:1]:1234
```
simonlingoogle, line 757:
```suggestion
Server:[fd00:0:0:0:0:0:0:1]:1234
```
nit: since it's actually address:port

src/core/config/dns_client.h
```diff
@@ -46,33 +46,68 @@
...
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_DNS_CLIENT_DEFAULT_RESPONSE_TIMEOUT
+ *
+ * Specifies the default wait time that DNS client waits for a response from server (in milliseconds).
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_DNS_CLIENT_DEFAULT_RESPONSE_TIMEOUT
+#define OPENTHREAD_CONFIG_DNS_CLIENT_DEFAULT_RESPONSE_TIMEOUT 3000
```
simonlingoogle, line 89:
```suggestion
#define OPENTHREAD_CONFIG_DNS_CLIENT_DEFAULT_RESPONSE_TIMEOUT 6000
```

The 3s timeout seems too small. For example, `RFC8766 Discovery Proxy` sets the timeout to six seconds. So, the DNS client should use a longer timeout to better with DNS-SD server with Discovery Proxy. The default timeout of `dig` command is 5 seconds. 

Thoughts?

include/openthread/dns_client.h
```diff
@@ -55,9 +55,64 @@ extern "C" {
...
+/**
+ * This structure represents a DNS query configuration.
+ *
+ * Any of the fields in this structure can be set to zero to indicate that it is not specified.
+ *
+ */
+typedef struct otDnsQueryConfig
+{
+    otSockAddr         mServerSockAddr;  ///< Server address (IPv6 address/port). All zero or zero port for unspecified.
+    uint32_t           mResponseTimeout; ///< Wait time (in msec) to rx response. Zero indicates unspecified value.
```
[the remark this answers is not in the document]
abtink: Note that zero is **not** the default value for the timeout. Otherwise default behavior with zero timeout would be that we won't wait at all 😄 

Value zero indicates it is unspecified (i.e. user does not care to give and leave it to stack to choose/use a default)

include/openthread/dns_client.h
```diff
@@ -55,9 +55,64 @@ extern "C" {
...
+    uint8_t            mMaxTxAttempts;   ///< Maximum tx attempts before reporting failure. Zero for unspecified value.
```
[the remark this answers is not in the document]
abtink: Same.

include/openthread/dns_client.h
abtink: Note that the config parameter in each query is optional and caller can choose not to use/provide one.

This model gives the user flexibility to set a default server/config  (e.g. the one used for all DNS-SD queries) so service discovery can be done easier, but still be able to send query to other servers (specify config per query) without changing the default config.

> Remove this argument since the caller can already do everything it want with otDnsClientSetDefaultConfig (which I suggested to rename to otDnsClientSetConfig) ?
> I think we should make the API as simple as possible, unless the benefits are clear.

There is a reason and benefit for this model:
- User (next layer code) may not want to use same server for all query types:
   - e.g., address query may use a differernt name server, compared to the server used for DNS-SD queries (another use-case is nat64 which may require its own address queries).
- If we require all queries to use the default config, then the user need to call multiple APIs whenever she wants to send a query with a different config:
  -  First change the default config, start the query, change it back to original 
  - or user needs to always call two APIs: set the default, then call the query (which defeats the purpose of having a default config).

include/openthread/dns_client.h
abtink: Answered above.

src/core/config/dns_client.h
abtink: I am fine changing it to 6 seconds. 

But note that there is also by default 3 tx attempts before reporting failure, so with 6 seconds wait per attempt, total wait time is 6x3 = 18 seconds.

include/openthread/dns_client.h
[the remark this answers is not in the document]
simonlingoogle: By **default** I meant the starting default values.

It seems unclear what `unspecified value` means just by reading the comment here, even though it's explained later in `otDnsClientSetDefaultConfig`. 

Maybe add a `@sa otDnsClientSetDefaultConfig` ?

include/openthread/dns_client.h
simonlingoogle: Agree that the current model has better flexibility and so, no strong preference.

> e.g., address query may use a differernt name server, compared to the server used for DNS-SD queries (another use-case is nat64 which may require its own address queries).

The DNS-SD server could recursively query other name servers so that the DNS client don't need to use different DNS server for different purpuses.

> If we require all queries to use the default config, then the user need to call multiple APIs whenever she wants to send a query with a different config:

To reduce code size and complexity, I tend to prefer the strategy of not implementing features that are unclear if it's going to be used or how it's going to be used, until real use cases appear. 

Again, no strong preference. So, I am resolving this conversation.

src/core/config/dns_client.h
simonlingoogle: > But note that there is also by default 3 tx attempts before reporting failure, so with 6 seconds wait per attempt, total wait time is 6x3 = 18 seconds. I think should be fine.

I can see that. Actually that's the reason why `dig`'s 5s timeout is also fine. Thanks.

simonlingoogle (approved):
LGTM :+1: 

with small suggestions

include/openthread/dns_client.h
```diff
@@ -55,9 +55,64 @@ extern "C" {
...
+ * @returns A pointer to the current default config being used by DNS client.
+ *
+ */
+const otDnsQueryConfig *otDnsClientGetDefaultConfig(otInstance *aInstance);
 
-#define OT_DNS_DEFAULT_SERVER_IP "2001:4860:4860::8888" ///< Defines default DNS Server address - Google DNS.
+/**
+ * This function sets the default query config on DNS client.
+ *
+ * @note Any ongoing query will continue to use config from when it was started. The new default config will be
```
jwhui, line 100:
```suggestion
 * @note Any ongoing query will continue to use the config from when it was started. The new default config will be
```

include/openthread/dns_client.h
```diff
@@ -55,9 +55,64 @@ extern "C" {
...
+ * considered and used for any future DNS queries.
```
jwhui, line 101:
```suggestion
 * used for any future DNS queries.
```

include/openthread/dns_client.h
```diff
@@ -55,9 +55,64 @@ extern "C" {
...
+ *
+ * The @p aConfig can be NULL. In this case the default config will be set to the defaults from OT config options
+ * `OPENTHREAD_CONFIG_DNS_CLIENT_DEFAULT_{}`. This practically resets the default query config back to to the config
```
jwhui, line 104:
```suggestion
 * `OPENTHREAD_CONFIG_DNS_CLIENT_DEFAULT_{}`. This resets the default query config back to to the config
```

src/cli/README.md
```diff
@@ -746,22 +746,69 @@ Perform an MLE Discovery operation.
...
-Send DNS Query to obtain IPv6 address for given hostname. The latter two parameters have following default values:
+Get the default query config used by DNS client.
 
-- DNS server IP: 2001:4860:4860::8888 (Google DNS Server)
-- DNS server port: 53
+The config includes the server IPv6 address and port, response timeout in msec (wait time to rx response), maximum tx attempts before reporting failure, boolean flag to indicate whether the server can resolve the query recursively or not.
+
+```bash
+> dns config
+Server:[fd00:0:0:0:0:0:0:1]:1234
```
jwhui, line 757: We typically have space after the `:` in other CLI output. We should be consistent. Is there a preference to remove the space?

src/core/config/dns_client.h
```diff
@@ -46,33 +46,68 @@
...
 /**
- * @def OPENTHREAD_CONFIG_DNS_MAX_RETRANSMIT
+ * @def OPENTHREAD_CONFIG_DNS_CLIENT_DEFAULT_SERVER_IP6_ADDRESS
  *
- * Maximum number of retransmissions for DNS client.
+ * Specifies the default DNS server IPv6 address.
+ *
+ * It MUST be a C string representation of the server IPv6 address.
+ *
+ * Default value is set to "2001:4860:4860::8888" which is the Google Public DNS Ipv6 address.
```
jwhui, line 65:
```suggestion
 * Default value is set to "2001:4860:4860::8888" which is the Google Public DNS IPv6 address.
```

src/core/net/dns_client.hpp
```diff
@@ -408,74 +508,106 @@ class Client : public InstanceLocator, private NonCopyable
...
+    /**
+     * This method sets the default query config.
+     *
+     * @param[in] aQueryConfig   The new default query config.
+     *
+     */
+    void SetDefaultConfig(const QueryConfig &aQueryConfig);
+
+    /**
+     * This method resets the default config to config when OpenThread stack starts.
```
jwhui, line 528:
```suggestion
     * This method resets the default config to the config used when the OpenThread stack starts.
```

src/core/net/dns_client.hpp
```diff
@@ -408,74 +508,106 @@ class Client : public InstanceLocator, private NonCopyable
...
+     *
+     * When OpenThread stack starts, the default DNS query config is determined from a set of OT config options such as
+     * `OPENTHREAD_CONFIG_DNS_CLIENT_DEFAULT_SERVER_IP6_ADDRESS`, `_DEFAULT_SERVER_PORT`, or `_DEFAULT_RESPONSE_TIMEOUT`
+     * etc. (please see `config/dns_clinet.h` for all related config options).
```
jwhui, line 532:
```suggestion
     * etc. (see `config/dns_clinet.h` for all related config options).
```

include/openthread/dns_client.h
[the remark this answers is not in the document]
abtink: The goal here in the definition of `otDnsQueryConfig` is to mention what value to use to indicate that the field is unspecified.

Note that what replaces the unspecified field values depends on which API it is being used in (which is then explained in the related method documentation)

src/cli/README.md
abtink: I will change this, I also like the space after `:` when we list items per line (looks cleaner).

jwhui (approved):

PR #6174, opened 2021-02-16
[spinel] update domain prefix spinel property

XPAWAN:
This commit update the spinel domain prefix property with new Nd Dns and Domain Prefix flags.

### DIFF
--- a/src/lib/spinel/spinel.h
+++ b/src/lib/spinel/spinel.h
@@ -618,6 +618,12 @@ enum
 
 enum
 {
+    SPINEL_NET_FLAG_EXT_DP  = (1 << 6),
+    SPINEL_NET_FLAG_EXT_DNS = (1 << 7),
+};
+
+enum
+{
     SPINEL_ROUTE_PREFERENCE_HIGH   = (1 << SPINEL_NET_FLAG_PREFERENCE_OFFSET),
     SPINEL_ROUTE_PREFERENCE_MEDIUM = (0 << SPINEL_NET_FLAG_PREFERENCE_OFFSET),
     SPINEL_ROUTE_PREFERENCE_LOW    = (3 << SPINEL_NET_FLAG_PREFERENCE_OFFSET),
@@ -2294,7 +2300,7 @@ enum
     SPINEL_PROP_THREAD_STABLE_NETWORK_DATA_VERSION = SPINEL_PROP_THREAD__BEGIN + 9,
 
     /// On-Mesh Prefixes
-    /** Format: `A(t(6CbCbS))`
+    /** Format: `A(t(6CbCbSC))`
      *
      * Data per item is:
      *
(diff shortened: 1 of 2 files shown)


src/lib/spinel/spinel.h
```diff
@@ -2305,8 +2309,11 @@ enum
      *  `b`: "Is defined locally" flag. Set if this network was locally
      *       defined. Assumed to be true for set, insert and replace. Clear if
      *       the on mesh network was defined by another node.
+     *       This field is ignored for INSERT and REMOVE commands.
      *  `S`: The RLOC16 of the device that registered this on-mesh prefix entry.
      *       This value is not used and ignored when adding an on-mesh prefix.
+     *       This field is ignored for INSERT and REMOVE commands.
+     *  `C`: TLV flags extended (additional field for Thread 1.2 features).
```
abtink, line 2316: This works, but we can consider using `C: TLV flags` fields we already have here (which is used for other `SPINEL_NET_FLAGS`). I feel it may be easier (specially since new flags are `(1<<6)` and `(1<<7)` and fit in the byte. 

Thoughts?

XPAWAN: Hi Abtink
The 'C: TLV flags fields' is full. The 6th and 7th bits are used for storing default router preference data
('high', 'med', or 'low' status). I have added a new 'C: TLV flags fields extended' to keep additional flags separatelly.
The flags field has been added to the end of spinel structure to keep backward compatibility to previous versions.
In the INSERT command I am reading and parsing all the data until first flag field(this are mandatory data that must be present).
Additionally I am checking if "isLocal", 'rloc16' and 'extended flags' fields are present. 
These data is considered as optional.
When the fields "isLocal" and 'rloc16' are present, the variables content is ignored by INSERT command.
If the 'extended flags' data is present, it will be read and processed accordingly.


I have selected mdns flag in the extended flags bajt to be located under 7 bit and Dp flag at 6bit. 
This bits are selected to be compatible with OT kNdDnsFlag and kDpFlag constants provided in the network_data_tlvs.hpp file.

abtink: Makes sense. Thanks @XPAWAN .

abtink (approved):
Looks great. Thanks 👍

src/ncp/ncp_base_mtd.cpp
```diff
@@ -824,6 +847,16 @@ template <> otError NcpBase::HandlePropertyInsert<SPINEL_PROP_THREAD_ON_MESH_NET
     borderRouterConfig.mDefaultRoute = ((flags & SPINEL_NET_FLAG_DEFAULT_ROUTE) != 0);
     borderRouterConfig.mOnMesh       = ((flags & SPINEL_NET_FLAG_ON_MESH) != 0);
 
+    // A new filed 'TLV flags extended' has been added to the SPINEL_PROP_THREAD_ON_MESH_NETS property.
+    // To correctly handle a new filed for INSERT command the additinal fields 'is Local' and 'RLOC16' are read and ignored.
```
jwhui, line 851: typos:
```suggestion
    // A new field 'TLV flags extended' has been added to the SPINEL_PROP_THREAD_ON_MESH_NETS property.
    // To correctly handle a new field for INSERT command, the additional fields 'isLocal' and 'rloc16' are read and ignored.
```

jwhui:
@XPAWAN , thanks for the update! Can you help address the pretty-check failure?

XPAWAN:
> @XPAWAN , thanks for the update! Can you help address the pretty-check failure?

Thanks for checking my PR.
pretty-check has been performed and PR is rebased to the current OT master.

jwhui (approved):

PR #6175, opened 2021-02-16
[cmake] using cmake -E echo instead of shell echo command

tejlmand:
The echo command of the host system may have slight variations between
shell built-in echo and the echo command provided by the OS.

Especially the Windows OS echo command is different from practical any
other echo command.

This means that the currently used `COMMAND echo -e "<text>"` does not
work universally.

This commit introduces `etc/cmake/print.cmake` which allow to print a
CMake list where each item is printed on a new line, to get the desired
behavior on the `print-ot-config` target.

Using `cmake -E echo <text>` ensures identical behaviour across
different systems.

Signed-off-by: Torsten Rasmussen <Torsten.Rasmussen@nordicsemi.no>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,4 +194,8 @@ endif()
 
 add_subdirectory(tests)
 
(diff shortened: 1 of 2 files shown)


tejlmand:
@googlebot I signed it!

LuDuda (approved):

jwhui (approved):
Thanks! 👍🏼

PR #6176, opened 2021-02-16
[efr32] change sleepy-demo-mtd to start awake

lmnotran:
This changes the sleepy-demo-mtd app's startup behavior. Prior to the change, the MTD would go to sleep immediately. Instead, the MTD will not sleep until the button is pressed


romacdon (approved):
LGTM

jwhui (approved):

PR #6177, opened 2021-02-16
[ip6] enqueue reassembly buffer immediately

jwhui:
So that `mReassemblyList.Dequeue()` is always valid on any error.


abtink (approved):
Nice catch! 👍

wgtdkp (approved):
LGTM 👍

PR #6180, opened 2021-02-17
[qorvo] update Makefiles to reflect a new structure in third_party submodule

thcu-gp:
Qorvo has restructured its `third_party/Qorvo/repo submodule`
The linkerscripts for the embedded platforms _qpg6095_ and _qpg6100_ have also been moved to this submodule


thcu-gp:
@jwhui the only problem with the above checks seems to be the _codecov_, which seem to report untouched files

jwhui (approved):

PR #6186, opened 2021-02-19
[posix] handle dynamic infra interface status

wgtdkp:
This commit includes below changes:
1. handles dynamic infrastructure interface status changes.
The Routing Manager will be started if the infra interface status turns
to be `RUNNING` and a valid link-local address is present. Otherwise,
the Routing Manager will be stopped.
2.  changes the default  status of the Routing Manager from `enabled`
to disabled and it is initially enabled for the posix platform.
3. the posix implementation now doesn't require the presence of the
link-local address at initialization stage. In this case, the Routing Manager
will not be started before a link-local address is added to the infra interface.

TODO:
- [x] add CI tests for the new behavior.

### DIFF
--- a/include/openthread/border_router.h
+++ b/include/openthread/border_router.h
@@ -57,20 +57,30 @@ extern "C" {
  *
  * @note  This method MUST be called before any other otBorderRouting* APIs.
  *
- * @param[in]  aInstance      A pointer to an OpenThread instance.
- * @param[in]  aInfraIfIndex  The infrastructure interface index.
+ * @param[in]  aInstance                 A pointer to an OpenThread instance.
+ * @param[in]  aInfraIfIndex             The infrastructure interface index.
+ * @param[in]  aInfraIfIsRunning         A boolean that indicates whether the infrastructure
+ *                                       interface is running.
+ * @param[in]  aInfraIfLinkLocalAddress  A pointer to the IPv6 link-local address of the infrastructure
+ *                                       interface. NULL if the IPv6 link-local address is missing.
  *
- * @retval  OT_ERROR_NONE          Successfully started the Border Routing manager on given infrastructure.
- * @retval  OT_ERROR_INVALID_ARGS  The index of the infra interface is not valid.
- * @retval  OT_ERROR_FAILED        Internal failure. This is usually failed to generate random prefixes.
+ * @retval  OT_ERROR_NONE           Successfully started the Border Routing Manager on given infrastructure.
+ * @retval  OT_ERROR_INVALID_STATE  The Border Routing Manager has already been initialized.
+ * @retval  OT_ERROR_INVALID_ARGS   The index or the IPv6 link-local address of the infra interface is not valid.
+ * @retval  OT_ERROR_FAILED         Internal failure. Usually due to failure in generating random prefixes.
+ *
+ * @sa otPlatInfraIfStateChanged.
  *
  */
-otError otBorderRoutingInit(otInstance *aInstance, uint32_t aInfraIfIndex);
+otError otBorderRoutingInit(otInstance *        aInstance,
+                            uint32_t            aInfraIfIndex,
+                            bool                aInfraIfIsRunning,
+                            const otIp6Address *aInfraIfLinkLocalAddress);
 
 /**
  * This method enables/disables the Border Routing Manager.
  *
- * @note  The Border Routing Manager is enabled by default.
+ * @note  The Border Routing Manager is disabled by default.
  *
  * @param[in]  aInstance  A pointer to an OpenThread instance.
  * @param[in]  aEnabled   A boolean to enable/disable the routing manager.
--- a/include/openthread/platform/infra_if.h
+++ b/include/openthread/platform/infra_if.h
@@ -47,16 +47,6 @@ extern "C" {
 #endif
 
 /**
- * This method returns the IPv6 link-local address of given infrastructure interface.
- *
- * @param[in]   aInfraIfIndex  The index of the infrastructure interface.
- *
- * @returns  A pointer to the IPv6 link-local address. NULL if no valid IPv6 link-local address found.
- *
- */
-const otIp6Address *otPlatInfraIfGetLinkLocalAddress(uint32_t aInfraIfIndex);
-
-/**
  * This method sends an ICMPv6 Neighbor Discovery message on given infrastructure interface.
  *
  * See RFC 4861: https://tools.ietf.org/html/rfc4861.
@@ -99,6 +89,29 @@ extern void otPlatInfraIfRecvIcmp6Nd(otInstance *        aInstance,
                                      const otIp6Address *aSrcAddress,
                                      const uint8_t *     aBuffer,
                                      uint16_t            aBufferLength);
+
+/**
+ * The infra interface driver calls this method to notify OpenThread
+ * of the interface state changes.
+ *
+ * @param[in]  aInstance          The OpenThread instance structure.
+ * @param[in]  aInfraIfIndex      The index of the infrastructure interface.
+ * @param[in]  aIsRunning         A boolean that indicates whether the infrastructure
+ *                                interface is running.
+ * @param[in]  aLinkLocalAddress  A pointer to the IPv6 link-local address of the infrastructure
+ *                                interface. NULL if the IPv6 link-local address is lost.
+ *
+ * @retval  OT_ERROR_NONE           Successfully updated the infra interface status.
+ * @retval  OT_ERROR_INVALID_STATE  The Routing Manager is not initialized.
+ * @retval  OT_ERROR_INVALID_ARGS   The @p aInfraIfIndex doesn't match the infra interface the
+ *                                  Routing Manager are initialized with, or the @p aLinkLocalAddress
+ *                                  is not a valid IPv6 link-local address.
+ *
+ */
+extern otError otPlatInfraIfStateChanged(otInstance *        aInstance,
+                                         uint32_t            aInfraIfIndex,
+                                         bool                aIsRunning,
+                                         const otIp6Address *aLinkLocalAddress);
 
 #ifdef __cplusplus
 } // extern "C"
--- a/src/core/border_router/infra_if_platform.cpp
+++ b/src/core/border_router/infra_if_platform.cpp
@@ -52,4 +52,16 @@ extern "C" void otPlatInfraIfRecvIcmp6Nd(otInstance *        aInstance,
     instance.Get<BorderRouter::RoutingManager>().RecvIcmp6Message(
         aInfraIfIndex, static_cast<const Ip6::Address &>(*aSrcAddress), aBuffer, aBufferLength);
 }
+
+extern "C" otError otPlatInfraIfStateChanged(otInstance *        aInstance,
+                                             uint32_t            aInfraIfIndex,
+                                             bool                aIsRunning,
+                                             const otIp6Address *aLinkLocalAddress)
+{
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
+    return instance.Get<BorderRouter::RoutingManager>().HandleInfraIfStateChanged(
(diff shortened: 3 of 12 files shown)


wgtdkp:
We may want to auto start/stop the routing manager when the infra interface status changes.

src/posix/platform/system.cpp
```diff
@@ -100,8 +100,18 @@ otInstance *otSysInit(otPlatformConfig *aPlatformConfig)
...
 
 #if OPENTHREAD_CONFIG_BORDER_ROUTING_ENABLE
-    // Reuse the backbone interface name.
-    platformInfraIfInit(instance, aPlatformConfig->mBackboneInterfaceName);
+    {
+        otError error;
+
+        // Reuse the backbone interface name.
+        error = platformInfraIfInit(instance, aPlatformConfig->mBackboneInterfaceName);
+        otBorderRoutingSetEnabled(instance, /* aEnabled */ (error == OT_ERROR_NONE));
```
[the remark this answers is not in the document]
wgtdkp: I think it is more convenient to enable the routing manager in ot-posix.

src/core/border_router/infra_if_platform.cpp
```diff
@@ -52,4 +52,16 @@ extern "C" void otPlatInfraIfRecvIcmp6Nd(otInstance *        aInstance,
     instance.Get<BorderRouter::RoutingManager>().RecvIcmp6Message(
         aInfraIfIndex, static_cast<const Ip6::Address &>(*aSrcAddress), aBuffer, aBufferLength);
 }
+
+extern "C" void otPlatInfraIfStateChanged(otInstance *        aInstance,
```
simonlingoogle, line 56: Why is this method a `otPlatXxxx`? I think this is not a **platform** method that is to be defined by each platform, but a public OT API method that is defined in OT core and used by each platform. 

```suggestion
void otInfraIfNotifyStateChanged(otInstance *        aInstance,
```

src/core/border_router/routing_manager.cpp
```diff
@@ -260,20 +268,41 @@ void RoutingManager::RecvIcmp6Message(uint32_t            aInfraIfIndex,
...
-        if (Get<Mle::MleRouter>().IsAttached())
-        {
-            Start();
-        }
-        else
-        {
-            Stop();
-        }
+        VerifyOrExit(Get<Mle::MleRouter>().IsAttached() != mIsAttached);
+        mIsAttached = Get<Mle::MleRouter>().IsAttached();
```
simonlingoogle, line 304:
```suggestion
```
why do we need to save `Get<Mle::MleRouter>().IsAttached()` to `mIsAttached` if we can just call `Get<Mle::MleRouter>().IsAttached()` whenever we want?

tests/scripts/thread-cert/node.py
```diff
@@ -2592,6 +2592,55 @@ class LinuxHost():
...
+        : param addr: the IPv6 address including the trailing prefix
+                      length separated with '/'.
+
+        Example usage of adding a link-local address:
+            self.add_ether_address('link', 'fe80::42:c0ff:fea8:903/64')
+        """
+
+        self.bash(f'ip addr add dev {self.ETH_DEV} scope {scope} {addr}')
+
+    def remove_ether_address(self, scope, addr):
```
simonlingoogle, line 2620: If this method is useless, remove it?

src/posix/platform/system.cpp
```diff
@@ -100,8 +101,15 @@ otInstance *otSysInit(otPlatformConfig *aPlatformConfig)
...
-    // Reuse the backbone interface name.
-    platformInfraIfInit(instance, aPlatformConfig->mBackboneInterfaceName);
+    {
+        // Reuse the backbone interface name.
+        uint32_t infraIfIndex = platformInfraIfInit(instance, aPlatformConfig->mBackboneInterfaceName);
+
+        SuccessOrDie(otBorderRoutingInit(instance, infraIfIndex));
+        otPlatInfraIfStateChanged(instance, infraIfIndex, platformInfraIfIsRunning(),
+                                  platformInfraIfGetLinkLocalAddress());
+        IgnoreError(otBorderRoutingSetEnabled(instance, /* aEnabled */ true));
```
simonlingoogle, line 111: why ignore error?

src/posix/platform/infra_if.cpp
```diff
@@ -250,21 +289,97 @@ void platformInfraIfDeinit(void)
...
+}
+
+static void ReceiveNetLinkMessage(otInstance *aInstance)
+{
+    const size_t kMaxNetLinkBufSize = 8192;
+    ssize_t      len;
+    union
+    {
+        nlmsghdr mHeader;
+        uint8_t  mBuffer[kMaxNetLinkBufSize];
```
simonlingoogle, line 344: This buffer is huge. Do we really need such large buffer?

src/core/border_router/infra_if_platform.cpp
wgtdkp: Methods definitions in `platform/` are meant to be used/called by platform modules but not to be implemented by them. Examples can be found for `radio` and `uart`:
https://github.com/openthread/openthread/blob/14ef4e4082ca06cdd05eb916cf8414b40c88c230/include/openthread/platform/radio.h#L723-L751

https://github.com/openthread/openthread/blob/14ef4e4082ca06cdd05eb916cf8414b40c88c230/include/openthread/platform/uart.h#L99-L112

src/core/border_router/routing_manager.cpp
wgtdkp: Not really necessary, just to be symmetrical with the `mInfraIfisRunning` state. I am ok to remove it.

src/posix/platform/system.cpp
wgtdkp: We are sure this will success. `OT_ASSERT` will be verbose.

src/posix/platform/infra_if.cpp
wgtdkp: Yes.

src/posix/platform/system.cpp
wgtdkp: replace with `SuccessOrDie`.

tests/scripts/thread-cert/node.py
wgtdkp: Ethernet address related method are all removed now.

Moandor-y (approved):
LGTM :+1:

src/core/border_router/routing_manager.cpp
```diff
@@ -260,20 +267,39 @@ void RoutingManager::RecvIcmp6Message(uint32_t            aInfraIfIndex,
     }
 }
 
+void RoutingManager::HandleInfraIfStateChanged(uint32_t            aInfraIfIndex,
+                                               bool                aIsRunning,
+                                               const Ip6::Address *aLinkLocalAddress)
+{
+    VerifyOrExit(IsInitialized() && aInfraIfIndex == mInfraIfIndex);
+    VerifyOrExit(aLinkLocalAddress == nullptr || aLinkLocalAddress->IsLinkLocal());
```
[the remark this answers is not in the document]
wgtdkp: I don't think we want to accept partial argument list.I changed this function to return `OT_ERROR_INVALID_ARGS` for this case.

simonlingoogle (approved):
LGTM :+1:

include/openthread/platform/infra_if.h
```diff
@@ -100,6 +90,23 @@ extern void otPlatInfraIfRecvIcmp6Nd(otInstance *        aInstance,
...
+/**
+ * The infra interface driver calls this method to notify OpenThread
+ * of the interface state changes.
+ *
+ * @param[in]  aInstance          The OpenThread instance structure.
+ * @param[in]  aInfraIfIndex      The index of the infrastructure interface.
+ * @param[in]  aIsRunning         A boolean that indicates whether the infrastructure
+ *                                interface is running.
+ * @param[in]  aLinkLocalAddress  A pointer to the IPv6 link local address of the infrastructure
+ *                                interface. NULL if the IPv6 link local address is lost.
```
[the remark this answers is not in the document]
wgtdkp: I changed to return `otError`.

include/openthread/border_router.h
```diff
@@ -57,20 +57,29 @@ extern "C" {
...
  * @retval  OT_ERROR_NONE          Successfully started the Border Routing manager on given infrastructure.
- * @retval  OT_ERROR_INVALID_ARGS  The index of the infra interface is not valid.
+ * @retval  OT_ERROR_INVALID_ARGS  The index or the IPv6 link-local address of the infra interface is not valid.
  * @retval  OT_ERROR_FAILED        Internal failure. This is usually failed to generate random prefixes.
  *
+ * @sa otPlatInfraIfStateChanged.
+ *
  */
-otError otBorderRoutingInit(otInstance *aInstance, uint32_t aInfraIfIndex);
+otError otBorderRoutingInit(otInstance *        aInstance,
```
bukepo, line 74: It seems this API is not necessary, and all the information can be delivered by the otPlatInfraIfStateChanged() API. I'm fine to deal with this issue in separate PRs.

bukepo (approved):

include/openthread/platform/infra_if.h
```diff
@@ -100,6 +90,29 @@ extern void otPlatInfraIfRecvIcmp6Nd(otInstance *        aInstance,
...
+ *                                interface is running.
+ * @param[in]  aLinkLocalAddress  A pointer to the IPv6 link-local address of the infrastructure
+ *                                interface. NULL if the IPv6 link-local address is lost.
+ *
+ * @retval  OT_ERROR_NONE           Successfully updated the infra interface status.
+ * @retval  OT_ERROR_INVALID_STATE  The Routing Manager is not initialized.
+ * @retval  OT_ERROR_INVALID_ARGS   The @p aInfraIfIndex doesn't match the infra interface the
+ *                                  Routing Manager are initialized with, or the @p aLinkLocalAddress
+ *                                  is not a valid IPv6 link-local address.
+ *
```
abtink, line 109: This is a callback from the perspective of platform layer so that it can inform OT core about state change. Is there any behavior that we expect the platform layer to do in case of an error?

Can we consider not returning an `error`?

abtink (approved):
Looks great overall. 👍 
One suggestion below on considering if we can make the callback return `void`?

include/openthread/platform/infra_if.h
simonlingoogle: The minor issue I can see without returning an `error` is that the user code might failed to call this method with correct network interface and link-local address, and didn't realize that. The `error` can help users to find such incorrect behavior easier.

wgtdkp: I share the similar thoughts with @simonlingoogle. The platform implementation may invoke this method in invalid state or with invalid arguments and because this function is public API, I think it is better to return the error but not assert or simply die.

We simply die on this error for posix implementation because we are sure that we are providing the correct arguments.
```c++
SuccessOrDie(otPlatInfraIfStateChanged(aInstance, sInfraIfIndex, platformInfraIfIsRunning(),
                                                   platformInfraIfGetLinkLocalAddress()));
```

src/posix/platform/infra_if.cpp
```diff
@@ -231,15 +236,49 @@ void platformInfraIfInit(otInstance *aInstance, const char *aIfName)
...
+    // Initializes the infra interface.
+    ifIndex = if_nametoindex(aIfName);
+    if (ifIndex == 0)
+    {
+        otLogCritPlat("failed to get the index for infra interface %s", aIfName);
+        DieNow(OT_EXIT_INVALID_ARGUMENTS);
+    }
+    sInfraIfIndex = ifIndex;
+
+    // TODO: We want to verify if there are IPv6 link-local address at starting.
```
jwhui, line 265: Is this difficult to address in this PR?

wgtdkp: Thanks for catching this, this is a stale TODO which is addressed by after refactoring.

jwhui (approved):

PR #6187, opened 2021-02-19
[dns] fix and simplify OPT record reuse of TTL field

abtink:
This commit fixes and simplifies the `OptRecord` implementation. The
OPT RR re-purposes the existing TTL field (which is a `uint32_t`) and
uses it for Extended Response Code (`uint8_t`), Version (`uint8_t`),
and flags. The new implementation treats the `mTtl` as a byte array
and directly reads/writes bytes at different indexes corresponding to
different re-purposed data fields. This ensures that the fields are
encoded correctly independent of the endianness of the platform.

### DIFF
--- a/src/core/net/dns_types.hpp
+++ b/src/core/net/dns_types.hpp
@@ -2279,7 +2279,7 @@ private:
 } OT_TOOL_PACKED_END;
 
 /**
- * This class implements DNS OPT Pseudo Resource Record header for EDNS(0) (RFC 6981 - Section 6.1).
+ * This class implements DNS OPT Pseudo Resource Record header for EDNS(0) (RFC 6891 - Section 6.1).
  *
  */
 OT_TOOL_PACKED_BEGIN
@@ -2332,10 +2332,7 @@ public:
      * @return The upper 8-bit of the extended 12-bit Response Code.
      *
      */
-    uint8_t GetExtendedResponseCode(void) const
-    {
-        return static_cast<uint8_t>((mTtl & kExtRCodeMask) >> kExtRCodeffset);
-    }
+    uint8_t GetExtendedResponseCode(void) const { return GetTtlByteAt(kExtRCodeByteIndex); }
 
     /**
      * This method sets the upper 8-bit of the extended 12-bit Response Code.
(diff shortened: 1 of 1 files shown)


src/core/net/dns_headers.hpp
```diff
@@ -2340,20 +2329,22 @@ class OptRecord : public ResourceRecord
...
-        kVersionMask   = 0xf << kVersionOffset, // Version field mask.
-        kDnsSecFlag    = 1 << 15,               // DnsSec bit flag.
+        kExtRCodeByteIndex = 0,      // Byte index of Extended RCODE within the TTL field.
+        kVersionByteIndex  = 1,      // Byte index of Version within the TTL field.
+        kFlagByteIndex     = 2,      // Byte index of flag byte within the TTL field.
+        kDnsSecFlag        = 1 << 7, // DNSSec OK bit flag.
     };
 
+    uint8_t  GetTtlByteAt(uint8_t aIndex) const { return reinterpret_cast<const uint8_t *>(&mTtl)[aIndex]; }
+    uint8_t &GetTtlByteAt(uint8_t aIndex) { return reinterpret_cast<uint8_t *>(&mTtl)[aIndex]; }
```
wgtdkp, line 2346: It is a little bit weird to assign to the return value of a method, although it works. Instead of having a `GetTtl*` method returns the reference, what about using a  symmetrical `SetTtlByteAt` that sets the value?

wgtdkp (approved):
LGTM :+1: 
Just a nit.

src/core/net/dns_headers.hpp
abtink: I actually had a `SetTtlByteAt()` method initially in the code but switched it to this model (of getting a referencee to the `uint8_t`). The main reason is the new model allows code like this:
```
GetTtlByteAt(kFlagByteIndex) &= ~kDnsSecFlag;
```
instead of
```
SetTtlByteAt(kFlagByteIndex, GetTtlByteAt(kFlagByteIndex) & ~kDnsSecFlag);
```
Since it's private I felt making it easier to read would be better.

------

In general, we do use the concept of returning a reference from a getter in OT core in different modules (instead of or in addition to a setter). 
- For example, in `SockAddr` we have a `GetAddress()` which returns a reference.
- The main reason for this is to allow caller to use methods of `Ip6::Address` directly (e.g. `sockAddr.GetAddress().Clear()` or `sockAddr.GetAddress().SetToRoutingLocator(...)`, etc).

jwhui (approved):

Issue #6133, opened 2021-02-02
[codecov] pull request status checks is broken


jwhui:
**Describe the bug**

codecov status checks are either not being reported or updated properly.

- #6130 
- #6131 
- #6132

bukepo:
also for #5384

wgtdkp:
And https://github.com/openthread/openthread/pull/6123 is reporting wrong coverage.

brianljt:
also for #6070

simonlingoogle:
Technically speaking, this issue of wrong codecov reporting is not resolved yet. 
Should we keep it open until the actual codecov dropping issue is resolved? @jinran-google @jwhui

jwhui:
Closing stale issue.

PR #6189, opened 2021-02-21
[mac] refactor code to add frame version and header IE in TX frames

Irving-cl:
This PR refactors the code to set frame version, IE present field in frame control field and to append Header IE. This PR does these things:
* Split `UpdateFrameControlField` into two methods: `CalcIePresent` and `CalcFrameVersion`. Currently UpdateFrameControlField` is not easy-to-read and hard to extend the method to handle more cases.
* Move these code from `Mac` to `MeshForwarder`. I think it should be `MeshForwarder`'s responsibility (`PrepareDataFrame`) to prepare the frame (though `Mac` sometimes prepare data frames itself). Since when preparing the frame, most information comes from the `Message` which is only in `MeshForwarder`, we can put these code for preparing frame in `MeshForwarder`.
* Enhance the way to append Header IEs to the frame. Currently in `Mac::AppendHeaderIe`, it generates a list of `HeaderIe` first and then call `Frame::AppendHeaderIe`. There are two problems here: 1. it uses another copy (copy `HeaderIe` to the frame's buffer). 2. it has to initialize Vendor IE (e.g. TimeSync IE) after `Vendor IE` header has been copied to the buffer. Currently it directly finds the first Vendor IE in the buffer and initialize it as TimeSync IE, which is not extensible. This PR shifts the responsibility of 'how to append the IE' to `Frame`, by providing `AppendHeaderIeAt` method. And `MeshForwarder` can decide which IEs to append and directly call the method to append IEs to the frame.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,7 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "(Apple)?[Cc]lang")
 
     set(OT_CFLAGS
         $<$<COMPILE_LANGUAGE:C>:${OT_CFLAGS} -Wall -Wextra -Wshadow>
-        $<$<COMPILE_LANGUAGE:CXX>:${OT_CFLAGS} -Wall -Wextra -Wshadow -Wno-c++14-compat -Wno-instantiation-after-specialization -fno-exceptions>
+        $<$<COMPILE_LANGUAGE:CXX>:${OT_CFLAGS} -Wall -Wextra -Wshadow -Wno-c++14-compat -fno-exceptions>
     )
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,7 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "(Apple)?[Cc]lang")
 
     set(OT_CFLAGS
         $<$<COMPILE_LANGUAGE:C>:${OT_CFLAGS} -Wall -Wextra -Wshadow>
-        $<$<COMPILE_LANGUAGE:CXX>:${OT_CFLAGS} -Wall -Wextra -Wshadow -Wno-c++14-compat -fno-exceptions>
+        $<$<COMPILE_LANGUAGE:CXX>:${OT_CFLAGS} -Wall -Wextra -Wshadow -Wno-c++14-compat -Wno-instantiation-after-specialization -fno-exceptions>
     )
 endif()
 
--- a/src/core/mac/mac_frame.cpp
+++ b/src/core/mac/mac_frame.cpp
@@ -894,7 +894,19 @@ exit:
 }
 
 #if OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT
-template <typename T> otError Frame::AppendHeaderIeAt(uint8_t &aIndex)
+template <typename IeType> otError Frame::AppendHeaderIeAt(uint8_t &aIndex)
+{
+    otError error = OT_ERROR_NONE;
+
+    SuccessOrExit(error = InitIeHeaderAt(aIndex, IeType::kHeaderIeId, sizeof(IeType)));
+
+    InitIeContentAt<IeType>(aIndex);
+
+exit:
+    return error;
+}
+
+otError Frame::InitIeHeaderAt(uint8_t &aIndex, uint8_t ieId, uint8_t ieContentSize)
 {
     otError error = OT_ERROR_NONE;
 
@@ -905,13 +917,10 @@ template <typename T> otError Frame::AppendHeaderIeAt(uint8_t &aIndex)
 
     VerifyOrExit(aIndex != kInvalidIndex, error = OT_ERROR_NOT_FOUND);
 
-    reinterpret_cast<HeaderIe *>(mPsdu + aIndex)->Init(T::kHeaderIeId, sizeof(T));
+    reinterpret_cast<HeaderIe *>(mPsdu + aIndex)->Init(ieId, ieContentSize);
     aIndex += sizeof(HeaderIe);
 
-    InitIeContentAt<T>(aIndex);
-
-    mLength += sizeof(HeaderIe) + sizeof(T);
-
+    mLength += sizeof(HeaderIe) + ieContentSize;
 exit:
     return error;
 }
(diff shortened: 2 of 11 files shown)


brianljt:
In general. I think the splitting of `UpdateFrameControlField` is OK. But I don't quite like the way HeaderIE is prepared.

For most of the cases (CSL IE, Time sync IE), the method `AppendHeaderIe` will just reserve some payload length so that the content of it could be updated later. Both CSL IE and Time Sync IE are set in `nrf_802154_tx_started`. I would suggest to keep `AppendHeaderIe` as is.

For multiple vendor IE cases. We can still fill the OUI fields of the vendor header IE this way and change `Mac::GetTimeIeOffset` method to find the real TimeIE. If other vendor IEs are added, we can fill the OUI and type in `AppendHeaderIe` and get the offset with new method. Thoughts?

src/core/mac/mac_frame.cpp
```diff
@@ -886,25 +893,50 @@ uint8_t Frame::FindHeaderIeIndex(void) const
...
 
+    VerifyOrExit(aIndex != kInvalidIndex, error = OT_ERROR_NOT_FOUND);
+
+    reinterpret_cast<HeaderIe *>(mPsdu + aIndex)->Init(T::kHeaderIeId, sizeof(T));
+    aIndex += sizeof(HeaderIe);
+
+    InitIeContentAt<T>(aIndex);
+
+    mLength += sizeof(HeaderIe) + sizeof(T);
+
```
abtink, line 914: A suggestion here: We can cosnider refactoring to add a common shared method (everything up to `InitIeContentAt<T>()`) as a separate helper (non-template) method.

The main reason is that with the current code we can end with different instantiation of `AppendHeaderIeAt<Type>()` method (each as its own method) repeating similar looking code (the first part up to `InitIeContentAt`), and adding to the overall code size. 

Thoughts?

src/core/mac/mac_frame.hpp
```diff
@@ -1679,6 +1711,21 @@ class CslIe
...
+/**
+ * This class implements Termination2 IE.
+ *
+ * This class is empty for template Specialization.
+ *
+ */
+class Termination2Ie
+{
+public:
+    enum
```
abtink, line 1723: Same?

src/core/thread/mesh_forwarder.cpp
```diff
@@ -1491,6 +1510,82 @@ otError MeshForwarder::SendEmptyMessage(void)
...
+{
+    uint16_t version = Mac::Frame::kFcfFrameVersion2006;
+    OT_UNUSED_VARIABLE(aNeighbor);
+
+    if (aIePresent)
+    {
+        version = Mac::Frame::kFcfFrameVersion2015;
+    }
+    else
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_MAC_CSL_TRANSMITTER_ENABLE
```
abtink, line 1569: I think we can put the `else` inside the `#if` block and it would work. (this way I think we won't need the final empty `{}`).

abtink (approved):
Looks good overall. Few suggestions below. Thanks.

brianljt (approved):
LGTM

Irving-cl:
Seems like this PR causes some tests to hang. I'm debugging it.

Irving-cl:
> Seems like this PR causes some tests to hang. I'm debugging it.

Fixed.

Irving-cl:
@jwhui 

It's ready to merge now, blocked the cov issue #6133 .

src/core/mac/mac_frame.hpp
```diff
@@ -923,16 +946,22 @@ class Frame : public otRadioFrame
...
+     * @param[in/out]   aIndex  The index to append IE. If `aIndex` is `0` on input, this method would find the index
+     *                          for the first IE and append the IE at that position. If the position is not found
+     *                          successfully, `aIndex` will be set to `kInvalidIndex`. Otherwise the IE will be
+     *                          appended at `aIndex` on input. And on output, `aIndex` will be set to the end of the
+     *                          IE just appended.
      *
-     * @param[in]   aIeList  The pointer to the Header IEs array.
-     * @param[in]   aIeCount The number of Header IEs in the array.
+     * @tparam  IeType  The type of a Header IE, it MUST contain an enum `kHeaderIeId` which equals to the Id of that
+     *                  IE and an enum `kIeContentSize` which indicates the IE body's size.
```
jwhui, line 958:
```suggestion
     * @tparam  IeType  The Header IE type. It MUST contain an enum `kHeaderIeId` equal to the IE's Id
     *                  and an enum `kIeContentSize` indicating the IE body's size.
```

src/core/mac/mac_frame.hpp
```diff
@@ -71,6 +71,15 @@ class HeaderIe
      */
     void Init(void) { mFields.m16 = 0; }
 
+    /**
+     * This method initializes the Header IE with Id and Length.
+     *
+     * @param[in]  aID      The IE Element Id.
```
jwhui, line 77:
```suggestion
     * @param[in]  aId      The IE Element Id.
```

src/core/mac/mac_frame.hpp
```diff
@@ -1679,6 +1717,22 @@ class CslIe
     uint16_t mPeriod;
 } OT_TOOL_PACKED_END;
 
+/**
+ * This class implements Termination2 IE.
+ *
+ * This class is empty for template Specialization.
```
jwhui, line 1723:
```suggestion
 * This class is empty for template specialization.
```

src/core/mac/mac_frame.hpp
```diff
@@ -71,6 +71,15 @@ class HeaderIe
      */
     void Init(void) { mFields.m16 = 0; }
 
+    /**
+     * This method initializes the Header IE with Id and Length.
+     *
+     * @param[in]  aID      The IE Element Id.
+     * @param[in]  aLength  The IE content length.
```
jwhui, line 78:
```suggestion
     * @param[in]  aLen     The IE content length.
```

CMakeLists.txt
```diff
@@ -76,7 +76,7 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "(Apple)?[Cc]lang")
 
     set(OT_CFLAGS
         $<$<COMPILE_LANGUAGE:C>:${OT_CFLAGS} -Wall -Wextra -Wshadow>
-        $<$<COMPILE_LANGUAGE:CXX>:${OT_CFLAGS} -Wall -Wextra -Wshadow -Wno-c++14-compat -fno-exceptions>
+        $<$<COMPILE_LANGUAGE:CXX>:${OT_CFLAGS} -Wall -Wextra -Wshadow -Wno-c++14-compat -Wno-instantiation-after-specialization -fno-exceptions>
```
jwhui, line 79: Can we avoid suppressing `-Winstantiation-after-specialization`?

jwhui (approved):

PR #6190, opened 2021-02-22
[spinel] replace otPlatSettings with SettingsDriver

gjc13:
The otPlatSettings api isn't necessarily supported on every platform.
Replace with ot::SettingsDriver which supports both the settings and the
flash platform api.

### DIFF
--- a/src/lib/spinel/radio_spinel_impl.hpp
+++ b/src/lib/spinel/radio_spinel_impl.hpp
@@ -38,7 +38,6 @@
 
 #include <openthread/dataset.h>
 #include <openthread/platform/diag.h>
-#include <openthread/platform/settings.h>
 #include <openthread/platform/time.h>
 
 #include "common/code_utils.hpp"
@@ -418,7 +417,7 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::RestoreDatasetFromNcp(vo
 {
     otError error = OT_ERROR_NONE;
 
-    otPlatSettingsInit(mInstance);
+    Instance::Get().template Get<SettingsDriver>().Init();
 
     otLogInfoPlat("Trying to get saved dataset from NCP");
     SuccessOrExit(
@@ -427,7 +426,7 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::RestoreDatasetFromNcp(vo
         error = Get(SPINEL_PROP_THREAD_PENDING_DATASET, SPINEL_DATATYPE_VOID_S, &RadioSpinel::ThreadDatasetHandler));
 
 exit:
-    otPlatSettingsDeinit(mInstance);
+    Instance::Get().template Get<SettingsDriver>().Deinit();
     return error;
 }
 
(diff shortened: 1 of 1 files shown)


gjc13:
@googlebot I signed it!

gjc13:
@bukepo Please kindly review.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -47,6 +47,7 @@
 #include "common/logging.hpp"
 #include "common/new.hpp"
 #include "common/settings.hpp"
+#include "core/common/instance.hpp"
```
Moandor-y, line 50:
```suggestion
```
This is already included by `#include "common/instance.hpp"`.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -418,7 +419,7 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::RestoreDatasetFromNcp(vo
 {
     otError error = OT_ERROR_NONE;
 
-    otPlatSettingsInit(mInstance);
+    Instance::Get().template Get<SettingsDriver>().Init();
```
Moandor-y, line 422:
```suggestion
    Instance::Get().Get<SettingsDriver>().Init();
```
This keyword is not required here.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -427,7 +428,7 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::RestoreDatasetFromNcp(vo
         error = Get(SPINEL_PROP_THREAD_PENDING_DATASET, SPINEL_DATATYPE_VOID_S, &RadioSpinel::ThreadDatasetHandler));
 
 exit:
-    otPlatSettingsDeinit(mInstance);
+    Instance::Get().template Get<SettingsDriver>().Deinit();
```
Moandor-y, line 431:
```suggestion
    Instance::Get().Get<SettingsDriver>().Deinit();
```

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -732,9 +733,9 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::ThreadDatasetHandler(con
     opDataset.mComponents.mIsActiveTimestampPresent = true;
 
     SuccessOrExit(error = dataset.SetFrom(static_cast<MeshCoP::Dataset::Info &>(opDataset)));
-    SuccessOrExit(error = otPlatSettingsSet(
-                      mInstance, isActive ? SettingsBase::kKeyActiveDataset : SettingsBase::kKeyPendingDataset,
-                      dataset.GetBytes(), dataset.GetSize()));
+    SuccessOrExit(error = Instance::Get().template Get<SettingsDriver>().Set(
```
Moandor-y, line 736:
```suggestion
    SuccessOrExit(error = Instance::Get().Get<SettingsDriver>().Set(
```

Moandor-y (commented):
Also, this line should not be needed anymore.
https://github.com/openthread/openthread/blob/d47169fd426070c7d54b6a9d0bfb3dac70f9488f/src/lib/spinel/radio_spinel_impl.hpp#L41

src/lib/spinel/radio_spinel_impl.hpp
Moandor-y: It turns out that clang still requires this keyword here. GCC doesn't. Please ignore this suggestion and revert. :sweat_smile:

gjc13:
@Moandor-y Looks like whether compliers treat `Instance::Get().Get<T>` as dependant type varies from case to case.
For compatiblity we may still need the `template` keyword.

Moandor-y (approved):
LGTM :+1:

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -427,7 +426,7 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::RestoreDatasetFromNcp(vo
         error = Get(SPINEL_PROP_THREAD_PENDING_DATASET, SPINEL_DATATYPE_VOID_S, &RadioSpinel::ThreadDatasetHandler));
 
 exit:
-    otPlatSettingsDeinit(mInstance);
+    Instance::Get().template Get<SettingsDriver>().Deinit();
```
abtink, line 429: Can we possibly get away with not requiring `template` keyword if we include the `locator-getters.hpp` header in this file?

abtink: Reading more on this, I don't think it would help since we are already inclining `instance.hpp` header which defines  the template `Get<>` definition.

abtink (approved):
LGTM. 👍

jwhui (approved):

PR #6191, opened 2021-02-22
[dnssd] implement DNS-SD Discovery Proxy

simonlingoogle:
This commit implements DNS-SD Discovery Proxy.

- [x] Implemented DNS-SD Discovery Proxy functionalities - OT part
- [x] Fixed an memory issue of `NameCompressInfo` (introduced in #6155 , but somehow revealed by this commit)

Depends-On: openthread/ot-br-posix#706

---

**Some implementation deitals:**
* Discovery Proxy subscribes to services/service instances/hosts through callbacks set by a public OT API. It is up to the platform mDNS implementation to collect service instance/host information and notify OT via a public OT API.
* Discovery Proxy can handle DNS browsing of one service or DNS resolving of one service instance/host. We leave browsing multiple services or resolving multiple service instances/hosts for future enhancements if necessary.

### DIFF
--- a/include/openthread/dnssd_server.h
+++ b/include/openthread/dnssd_server.h
@@ -1,0 +1,180 @@
+/*
+ *  Copyright (c) 2021, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ * @brief
+ *  This file defines the DNS-SD server APIs.
+ */
+
+#ifndef OPENTHREAD_DNSSD_SERVER_H_
+#define OPENTHREAD_DNSSD_SERVER_H_
+
+#include <stdint.h>
+
+#include <openthread/error.h>
+#include <openthread/ip6.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @addtogroup api-dnssd-server
+ *
+ * @brief
+ *   This module includes APIs for DNS-SD server.
+ *
+ * @{
+ *
+ */
+
+/**
+ * This function is called when a DNS-SD query subscribes either of:
+ *      1. a service name.
+ *      2. a service instance name.
+ *      3. a host name.
+ *
+ * The DNS-SD query implementation is responsible for identifying what @p aFullName is.
+ * If @p aFullName is a service name or service instance name, the DNS-SD query implementation should discover
+ * corresponding service instance information and notify the DNS-SD server using
+ * `otDnssdQueryHandleDiscoveredServiceInstance`.
+ * If @p aFullName is a host name, the DNS-SD query implementation should
+ * discover the host information and notify the DNS-SD server using `otDnssdQueryHandleDiscoveredHost`.
+ *
+ * @note There can be multiple subscription to the same name. DNS-SD query implementation should  record the number of
+ * active subscriptions and stop notifying when there is no active subscription for @p aFullName.
+ *
+ * @param[in] aContext      A pointer to the application-specific context.
+ * @param[in] aFullName     The null-terminated full service name (e.g. "_ipps._tcp.default.service.arpa."),
+ *                          or full service instance name (e.g. "OpenThread._ipps._tcp.default.service.arpa."),
+ *                          or full host name (e.g. "ot-host.default.service.arpa.").
+ *
+ * @sa otDnssdQueryHandleDiscoveredServiceInstance
+ * @sa otDnssdQueryHandleDiscoveredHost
+ *
+ */
+typedef void (*otDnssdQuerySubscribeCallback)(void *aContext, const char *aFullName);
+
+/**
+ * This function is called when a DNS-SD query unsubscribes either of:
+ *      1. a service name.
+ *      2. a service instance name.
+ *      3. a host name.
+ *
+ * The DNS-SD query implementation is responsible for identifying what @p aFullName is.
+ *
+ * @note There can be multiple subscription to the same name. DNS-SD query implementation should  record the number of
+ * active subscriptions and stop notifying when there is no active subscription for @p aFullName.
+ *
+ * @param[in] aContext      A pointer to the application-specific context.
+ * @param[in] aFullName     The null-terminated full service name (e.g. "_ipps._tcp.default.service.arpa."), or
+ *                          full service instance name (e.g. "OpenThread._ipps._tcp.default.service.arpa.").
+ *
+ */
+typedef void (*otDnssdQueryUnsubscribeCallback)(void *aContext, const char *aFullName);
+
+/**
+ * This structure represents information of a discovered service instance for a DNS-SD query.
+ *
+ */
+typedef struct otDnssdServiceInstanceInfo
+{
+    const char *        mFullName;   ///< Full instance name (e.g. "OpenThread._ipps._tcp.default.service.arpa.").
+    const char *        mHostName;   ///< Host name (e.g. "ot-host.default.service.arpa.").
+    uint8_t             mAddressNum; ///< Number of host IPv6 addresses.
+    const otIp6Address *mAddresses;  ///< Host IPv6 addresses.
+    uint16_t            mPort;       ///< Service port.
+    uint16_t            mPriority;   ///< Service priority.
+    uint16_t            mWeight;     ///< Service weight.
+    uint16_t            mTxtLength;  ///< Service TXT RDATA length.
+    const uint8_t *     mTxtData;    ///< Service TXT RDATA.
+    uint32_t            mTtl;        ///< Service TTL (in seconds).
+} otDnssdServiceInstanceInfo;
+
+/**
+ * This structure represents information of a discovered host for a DNS-SD query.
+ *
+ */
+typedef struct otDnssdHostInfo
+{
+    uint8_t             mAddressNum; ///< Number of host IPv6 addresses.
+    const otIp6Address *mAddresses;  ///< Host IPv6 address.
+    uint32_t            mTtl;        ///< Service TTL (in seconds).
+} otDnssdHostInfo;
+
+/**
+ * This function sets DNS-SD server query callbacks. The DNS-SD server calls @p aSubscribe to subscribe to a service or
+ * service instance to resolve a DNS-SD query and @p aUnsubscribe to unsubscribe when the query is resolved or timeout.
+ *
+ * @note @p aSubscribe and @p aUnsubscribe should be both set or unset.
+ *
+ * @param[in] aInstance     The OpenThread instance structure.
+ * @param[in] aContext      A pointer to the application-specific context.
+ * @param[in] aSubscribe    A pointer to the callback function to subscribe a service or service instance.
+ * @param[in] aUnsubscribe  A pointer to the callback function to unsubscribe a service or service instance.
+ *
+ */
+void otDnssdQuerySetCallbacks(otInstance *                    aInstance,
+                              otDnssdQuerySubscribeCallback   aSubscribe,
+                              otDnssdQueryUnsubscribeCallback aUnsubscribe,
+                              void *                          aContext);
+
+/**
+ * This function notifies a discovered service instance. The external query resolver (e.g. Discovery Proxy) should call
+ * this function to notify OpenThread core of the subscribed services or service instances.
+ *
+ * @param[in] aInstance         The OpenThread instance structure.
+ * @param[in] aServiceFullName  The null-terminated full service name.
+ * @param[in] aInstanceInfo     A pointer to the discovered service instance information.
+ *
+ */
+void otDnssdQueryHandleDiscoveredServiceInstance(otInstance *                aInstance,
+                                                 const char *                aServiceFullName,
+                                                 otDnssdServiceInstanceInfo *aInstanceInfo);
+/**
+ * This function notifies a discovered host. The external query resolver (e.g. Discovery Proxy) should call
+ * this function to notify OpenThread core of the subscribed hosts.
+ *
+ * @param[in] aInstance         The OpenThread instance structure.
+ * @param[in] aHostFullName     The null-terminated full host name.
+ * @param[in] aHostInfo         A pointer to the discovered service instance information.
+ *
+ */
+void otDnssdQueryHandleDiscoveredHost(otInstance *aInstance, const char *aHostFullName, otDnssdHostInfo *aHostInfo);
+
+/**
+ * @}
+ *
+ */
+
+#ifdef __cplusplus
+} // extern "C"
+#endif
+
+#endif // OPENTHREAD_DNSSD_SERVER_H_
--- a/include/openthread/dnssd_server.h
+++ b/include/openthread/dnssd_server.h
@@ -55,7 +55,7 @@ extern "C" {
  */
 
 /**
- * This function is called when a DNS-SD query subscribes either of:
+ * This function is called when a DNS-SD query subscribes one of:
  *      1. a service name.
  *      2. a service instance name.
  *      3. a host name.
@@ -67,7 +67,7 @@ extern "C" {
  * If @p aFullName is a host name, the DNS-SD query implementation should
  * discover the host information and notify the DNS-SD server using `otDnssdQueryHandleDiscoveredHost`.
  *
- * @note There can be multiple subscription to the same name. DNS-SD query implementation should  record the number of
+ * @note There can be multiple subscription to the same name. DNS-SD query implementation should record the number of
  * active subscriptions and stop notifying when there is no active subscription for @p aFullName.
  *
  * @param[in] aContext      A pointer to the application-specific context.
@@ -82,14 +82,14 @@ extern "C" {
 typedef void (*otDnssdQuerySubscribeCallback)(void *aContext, const char *aFullName);
 
 /**
- * This function is called when a DNS-SD query unsubscribes either of:
+ * This function is called when a DNS-SD query unsubscribes one of:
  *      1. a service name.
  *      2. a service instance name.
  *      3. a host name.
  *
  * The DNS-SD query implementation is responsible for identifying what @p aFullName is.
  *
- * @note There can be multiple subscription to the same name. DNS-SD query implementation should  record the number of
+ * @note There can be multiple subscription to the same name. DNS-SD query implementation should record the number of
  * active subscriptions and stop notifying when there is no active subscription for @p aFullName.
  *
  * @param[in] aContext      A pointer to the application-specific context.
@@ -124,20 +124,22 @@ typedef struct otDnssdServiceInstanceInfo
 typedef struct otDnssdHostInfo
 {
     uint8_t             mAddressNum; ///< Number of host IPv6 addresses.
-    const otIp6Address *mAddresses;  ///< Host IPv6 address.
+    const otIp6Address *mAddresses;  ///< Host IPv6 addresses.
     uint32_t            mTtl;        ///< Service TTL (in seconds).
 } otDnssdHostInfo;
 
 /**
- * This function sets DNS-SD server query callbacks. The DNS-SD server calls @p aSubscribe to subscribe to a service or
- * service instance to resolve a DNS-SD query and @p aUnsubscribe to unsubscribe when the query is resolved or timeout.
+ * This function sets DNS-SD server query callbacks.
  *
- * @note @p aSubscribe and @p aUnsubscribe should be both set or unset.
+ * The DNS-SD server calls @p aSubscribe to subscribe to a service or service instance to resolve a DNS-SD query and @p
+ * aUnsubscribe to unsubscribe when the query is resolved or timeout.
+ *
+ * @note @p aSubscribe and @p aUnsubscribe must be both set or unset.
  *
  * @param[in] aInstance     The OpenThread instance structure.
- * @param[in] aContext      A pointer to the application-specific context.
  * @param[in] aSubscribe    A pointer to the callback function to subscribe a service or service instance.
  * @param[in] aUnsubscribe  A pointer to the callback function to unsubscribe a service or service instance.
+ * @param[in] aContext      A pointer to the application-specific context.
  *
  */
 void otDnssdQuerySetCallbacks(otInstance *                    aInstance,
@@ -146,8 +148,12 @@ void otDnssdQuerySetCallbacks(otInstance *                    aInstance,
                               void *                          aContext);
 
 /**
- * This function notifies a discovered service instance. The external query resolver (e.g. Discovery Proxy) should call
- * this function to notify OpenThread core of the subscribed services or service instances.
+ * This function notifies a discovered service instance.
+ *
+ * The external query resolver (e.g. Discovery Proxy) should call this function to notify OpenThread core of the
+ * subscribed services or service instances.
+ *
+ * @note @p aInstanceInfo must not contain unspecified or link-local or loop-back or multicast IP addresses.
  *
  * @param[in] aInstance         The OpenThread instance structure.
  * @param[in] aServiceFullName  The null-terminated full service name.
@@ -158,8 +164,12 @@ void otDnssdQueryHandleDiscoveredServiceInstance(otInstance *                aIn
                                                  const char *                aServiceFullName,
                                                  otDnssdServiceInstanceInfo *aInstanceInfo);
 /**
- * This function notifies a discovered host. The external query resolver (e.g. Discovery Proxy) should call
- * this function to notify OpenThread core of the subscribed hosts.
+ * This function notifies a discovered host.
+ *
+ * The external query resolver (e.g. Discovery Proxy) should call this function to notify OpenThread core of the
+ * subscribed hosts.
+ *
+ * @note @p aHostInfo must not contain unspecified or link-local or loop-back or multicast IP addresses.
  *
  * @param[in] aInstance         The OpenThread instance structure.
  * @param[in] aHostFullName     The null-terminated full host name.
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -1538,10 +1538,9 @@ void Interpreter::HandleDnsAddressResponse(otError aError, const otDnsAddressRes
             OutputFormat(" TTL:%u ", ttl);
             index++;
         }
-
-        OutputLine("");
     }
 
+    OutputLine("");
     OutputResult(aError);
 }
 
--- a/src/core/net/dnssd_server.cpp
+++ b/src/core/net/dnssd_server.cpp
@@ -56,6 +56,10 @@ const char Server::kDefaultDomainName[] = "default.service.arpa.";
 Server::Server(Instance &aInstance)
     : InstanceLocator(aInstance)
     , mSocket(aInstance)
+    , mQueryCallbackContext(nullptr)
+    , mQuerySubscribe(nullptr)
+    , mQueryUnsubscribe(nullptr)
+    , mTimer(aInstance, Server::HandleTimer)
 {
 }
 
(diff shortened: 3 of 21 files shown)


src/core/net/dnssd_server.hpp
```diff
@@ -107,11 +132,10 @@ class Server : public InstanceLocator, private NonCopyable
             kUnknownOffset = 0, // Unknown offset value (used when offset is not yet set).
         };
 
+        explicit NameCompressInfo(void) = default;
+
         explicit NameCompressInfo(const char *aDomainName)
```
wgtdkp, line 137:
```suggestion
        explicit NameCompressInfo(const char *aDomainName=nullptr)
```

The members are not initialized in the default constructor.

simonlingoogle: It's actually fine not to initialize it.

tests/scripts/thread-cert/border_router/test_dnssd_server.py
```diff
@@ -221,6 +223,42 @@ def test(self):
                 'status': 'NXDOMAIN',
             })
 
+        # verify Discovery Proxy works for _meshcop._udp
+        dp_service_name = '_meshcop._udp.default.service.arpa.'
```
wgtdkp, line 227: To avoid confusion with the mDNS service published by Border Agent, can we consider using `_ipps._tcp` which is already used for existing DNS-related tests?

src/core/net/dnssd_server.hpp
```diff
@@ -232,46 +236,111 @@ class Server : public InstanceLocator, private NonCopyable
...
-    static void      IncResourceRecordCount(Header &aHeader, bool aAdditional);
-    static otError   FindNameComponents(const char *aName, const char *aDomain, NameComponentsOffsetInfo &aInfo);
-    static otError   FindPreviousLabel(const char *aName, uint8_t &aStart, uint8_t &aStop);
+#if OPENTHREAD_CONFIG_DNSSD_DISCOVERY_PROXY_ENABLE
+    enum : uint32_t
+    {
+        kDiscoveryProxyTimeout = OT_DNSSD_DISCOVERY_PROXY_TIMEOUT,
+    };
+
+    class DiscoveryProxyRequest
```
wgtdkp, line 245:
```suggestion
    class DnsQueryTransaction
```

This sturct doesn't seem to be a Discovery Proxy request.

src/core/net/dnssd_server.hpp
```diff
@@ -232,46 +236,111 @@ class Server : public InstanceLocator, private NonCopyable
...
+        Ip6::MessageInfo mMessageInfo;
+        TimeMilli        mStartTime;
+    };
+
+    enum DiscoveryProxyAction
+    {
+        kDiscoveryProxyActionNone,
+        kDiscoveryProxyActionBrowse,
+        kDiscoveryProxyActionResolve,
+    };
```
wgtdkp, line 280:
```suggestion
    enum DnsQueryType
    {
        kDnsQueryTypeNone,
        kDnsQueryTypeBrowse,
        kDnsQueryTypeResolve,
    };
```

src/core/net/dnssd_server.hpp
```diff
@@ -232,46 +236,111 @@ class Server : public InstanceLocator, private NonCopyable
...
+                                            const char *      aInstanceName,
+                                            const void *      aTxtData,
+                                            uint16_t          aTxtLength,
+                                            uint32_t          aTtl,
+                                            NameCompressInfo &aCompressInfo);
+    static otError          AppendAaaaRecord(Message &           aMessage,
+                                             const char *        aHostName,
+                                             const Ip6::Address &aAddress,
+                                             uint32_t            aTtl,
+                                             NameCompressInfo &  aCompressInfo);
```
wgtdkp, line 324: It seems better to have those static methods as member methods of the `DnsQueryTransaction` class. Thoughts?

src/core/net/dnssd_server.hpp
```diff
@@ -232,46 +236,111 @@ class Server : public InstanceLocator, private NonCopyable
...
+                                         const Ip6::MessageInfo &aMessageInfo,
+                                         Ip6::Udp::Socket &      aSocket);
 
 #if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
+    Header::Response                   ResolveBySrp(const Header &            aRequestHeader,
+                                                    const Message &           aRequestMessage,
+                                                    Header &                  aResponseHeader,
+                                                    Message &                 aResponse,
+                                                    uint8_t                   aResolveAdditional,
+                                                    Server::NameCompressInfo &aCompressInfo);
```
wgtdkp, line 343:
```suggestion
    Header::Response                   ResolveBySrp(DnsQueryTransaction &aDnsQuery);
```

wgtdkp (commented):
Didn't go through all the details, some initial comments.

src/core/api/dns_api.cpp
```diff
@@ -209,3 +210,27 @@ otError otDnsServiceResponseGetHostAddress(const otDnsServiceResponse *aResponse
...
+void otDnssdDiscoveryProxyNotifyServiceInstance(otInstance *                aInstance,
+                                                const char *                aServiceFullName,
+                                                otDnssdServiceInstanceInfo *aInstanceInfo)
+{
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
+    OT_ASSERT(aServiceFullName != nullptr);
+    OT_ASSERT(aInstanceInfo != nullptr);
+
+    instance.Get<Dns::ServiceDiscovery::Server>().DiscoveryProxyNotifyServiceInstance(aServiceFullName, *aInstanceInfo);
```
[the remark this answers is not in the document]
simonlingoogle: Why do we want to only pass the first character of the service full name (null-terminated string) ?

src/core/net/dnssd_server.hpp
simonlingoogle: These methods are also used when resolving by SRP. They are not specific to Discovery Proxy.

src/core/net/dnssd_server.hpp
simonlingoogle: Adding `DnsQueryTransaction` SGTM. But it would cause significant refactoring. 

I already have future plan to refactor `resolving by SRP` part of code. So, I would prefer to do refactoring later after we deliver the working features first. 

Thoughts?

src/core/net/dnssd_server.hpp
simonlingoogle: Let's postpone this major refactoring and deliver the working feature first.

src/core/net/dnssd_server.hpp
simonlingoogle: Let's postpone this major refactoring and deliver the working feature first.

src/core/net/dnssd_server.hpp
simonlingoogle: Renamed to `DnsServiceDiscoveryType`.

src/core/net/dnssd_server.hpp
simonlingoogle: Let's postpone this major refactoring and deliver the working feature first.

include/openthread/dns.h
```diff
@@ -140,6 +141,71 @@ void otDnsInitTxtEntryIterator(otDnsTxtEntryIterator *aIterator, const uint8_t *
  */
 otError otDnsGetNextTxtEntry(otDnsTxtEntryIterator *aIterator, otDnsTxtEntry *aEntry);
 
+#define OT_DNSSD_DISCOVERY_PROXY_TIMEOUT 6000 ///< DNS-SD Discovery Proxy timeout (in milliseconds).
+
+/**
+ * This function pointer is called when the DNS-SD Discovery Proxy subscribes a service or service instance.
+ *
+ * @param[in] aContext      A pointer to the application-specific context.
+ * @param[in] aFullName     The null-terminated full service name (e.g. "_meshcop._udp.default.service.arpa."), or
```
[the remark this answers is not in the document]
simonlingoogle: renamed to `_ipps._tcp`

src/core/net/dnssd_server.hpp
```diff
@@ -82,11 +82,40 @@ class Server : public InstanceLocator, private NonCopyable
...
+    void SetQueryCallbacks(void *aContext, otDnssdQuerySubscribe aSubscribe, otDnssdQueryUnsubscribe aUnsubscribe);
+
+    /**
+     * This method notifies a discovered service instance.
+     *
+     * @param[in] aServiceFullName  The null-terminated full service name.
+     * @param[in] aInstanceInfo     A pointer to the discovered service instance information.
+     *
+     */
+    void QueryNotifyServiceInstance(const char *aServiceFullName, const otDnssdServiceInstanceInfo &aInstanceInfo);
```
wgtdkp, line 102:
```suggestion
    void HandleServiceQueryResult(const char *aServiceFullName, const otDnssdServiceInstanceInfo &aInstanceInfo);
```

src/core/net/dnssd_server.hpp
```diff
@@ -82,11 +82,40 @@ class Server : public InstanceLocator, private NonCopyable
...
+
+    /**
+     * This method notifies a discovered host.
+     *
+     * @param[in] aHostFullName     The null-terminated full host name.
+     * @param[in] aHostInfo         A pointer to the discovered host information.
+     *
+     */
+    void QueryNotifyHost(const char *aHostFullName, const otDnssdHostInfo &aHostInfo);
```
wgtdkp, line 111:
```suggestion
    void HandleHostQueryResult(const char *aHostFullName, const otDnssdHostInfo &aHostInfo);
```

src/core/net/dnssd_server.hpp
```diff
@@ -281,18 +352,42 @@ class Server : public InstanceLocator, private NonCopyable
...
+    Error             ResolveByQueryCallbacks(Header &                aResponseHeader,
+                                              Message &               aResponseMessage,
+                                              uint8_t                 aResolveAdditional,
+                                              NameCompressInfo &      aCompressInfo,
+                                              const Ip6::MessageInfo &aMessageInfo);
+    QueryTransaction *NewQueryTransaction(const Header &          aResponseHeader,
+                                          Message &               aResponseMessage,
+                                          const NameCompressInfo &aCompressInfo,
+                                          const Ip6::MessageInfo &aMessageInfo);
+    static bool       CanAnswerQueryTransaction(const QueryTransaction &          aQuery,
```
wgtdkp, line 363:
```suggestion
    static bool       CanAnswerQuery(const QueryTransaction &          aQuery,
```

src/core/net/dnssd_server.hpp
```diff
@@ -281,18 +352,42 @@ class Server : public InstanceLocator, private NonCopyable
...
+                                                const char *                      aServiceFullName,
+                                                const otDnssdServiceInstanceInfo &aInstanceInfo);
+    void              AnswerQueryTransaction(QueryTransaction &                aQuery,
```
wgtdkp, line 366:
```suggestion
    void              AnswerQuery(QueryTransaction &                aQuery,
```

src/core/net/dnssd_server.hpp
```diff
@@ -281,18 +352,42 @@ class Server : public InstanceLocator, private NonCopyable
...
+                                             const char *                      aServiceFullName,
+                                             const otDnssdServiceInstanceInfo &aInstanceInfo);
+    static bool       CanAnswerQueryTransaction(const Server::QueryTransaction &aQuery, const char *aHostFullName);
```
wgtdkp, line 369:
```suggestion
    static bool       CanAnswerQuery(const Server::QueryTransaction &aQuery, const char *aHostFullName);
```

src/core/net/dnssd_server.hpp
```diff
@@ -281,18 +352,42 @@ class Server : public InstanceLocator, private NonCopyable
...
+    void AnswerQueryTransaction(QueryTransaction &aQuery, const char *aHostFullName, const otDnssdHostInfo &aHostInfo);
```
wgtdkp, line 370:
```suggestion
    void AnswerQuery(QueryTransaction &aQuery, const char *aHostFullName, const otDnssdHostInfo &aHostInfo);
```

src/core/net/dnssd_server.cpp
```diff
@@ -102,24 +111,23 @@ void Server::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessag
 
     // ProcessQuery is assumed to always prepare the response DNS message and header properly even in the case of system
     // failures (e.g. no more buffers).
-    ProcessQuery(aMessage, *responseMessage, requestHeader);
-
-    error = mSocket.SendTo(*responseMessage, aMessageInfo);
-
+    ProcessQuery(requestHeader, aMessage, *responseMessage, aMessageInfo);
 exit:
-    FreeMessageOnError(responseMessage, error);
```
wgtdkp, line 110: We need to keep this line because `SuccessOrExit(error = responseMessage->SetLength(sizeof(Header)));` can fail.

wgtdkp: And, looks like we don't need to alloc the response message here. Instead, it should be better to create the response message when there are really response from SRP server or Discovery Proxy.

src/core/net/dnssd_server.cpp
```diff
@@ -132,16 +140,90 @@ void Server::ProcessQuery(Message &aMessage, Message &aResponse, const Header &a
...
+    {
+        SendResponse(responseHeader, response, aResponseMessage, aMessageInfo, mSocket);
+    }
+}
+
+void Server::SendResponse(Header                  aHeader,
+                          Header::Response        aResponse,
+                          Message &               aMessage,
+                          const Ip6::MessageInfo &aMessageInfo,
+                          Ip6::Udp::Socket &      aSocket)
```
wgtdkp, line 177: nit:
```suggestion
```

src/core/net/dnssd_server.cpp
```diff
@@ -132,16 +140,90 @@ void Server::ProcessQuery(Message &aMessage, Message &aResponse, const Header &a
...
+    error = aSocket.SendTo(aMessage, aMessageInfo);
+
+    if (error != kErrorNone)
+    {
+        otLogWarnDns("[server] failed to send DNS-SD reply: %s", otThreadErrorToString(error));
+        aMessage.Free();
+    }
+    else
+    {
+        otLogInfoDns("[server] send DNS-SD reply: %s, RCODE=%d", otThreadErrorToString(error), aResponse);
```
wgtdkp, line 202: We need to free `aMessage` as well.

src/core/net/dnssd_server.cpp
```diff
@@ -411,39 +426,66 @@ Error Server::AppendInstanceName(Message &aMessage, const char *aName, NameCompr
...
+    recordOffset = aMessage.GetLength();
+    SuccessOrExit(error = aMessage.SetLength(recordOffset + sizeof(txtRecord)));
+
+    SuccessOrExit(error = aMessage.AppendBytes(aTxtData, aTxtLength));
+
+    txtRecord.Init();
+    txtRecord.SetTtl(aTtl);
+    txtRecord.SetLength(aMessage.GetLength() - (recordOffset + sizeof(ResourceRecord)));
+
+    aMessage.Write(recordOffset, txtRecord);
```
wgtdkp, line 478:
```suggestion
    txtRecord.Init();
    txtRecord.SetTtl(aTtl);
    txtRecord.SetLength(aTxtLength);
    SuccessOrExit(error = aMessage.Append(txtRecord);
    SuccessOrExit(error = aMessage.AppendBytes(aTxtData, aTxtLength));
```

src/core/net/dnssd_server.cpp
```diff
@@ -668,34 +759,381 @@ const Srp::Server::Service *Server::GetNextSrpService(const Srp::Server::Host &
...
+exit:
+    FinalizeQueryTransaction(aQuery, error == kErrorNone ? Header::kResponseSuccess : Header::kResponseServerFailure);
+    ResetTimer();
+}
+
+void Server::SetQueryCallbacks(void *aContext, otDnssdQuerySubscribe aSubscribe, otDnssdQueryUnsubscribe aUnsubscribe)
+{
+    OT_ASSERT((aSubscribe == nullptr) == (aUnsubscribe == nullptr));
+
+    mQueryBrowseContext = aContext;
```
wgtdkp, line 941:
```suggestion
    mQueryCallbackContext = aContext;
```

src/core/net/dnssd_server.cpp
```diff
@@ -668,34 +759,381 @@ const Srp::Server::Service *Server::GetNextSrpService(const Srp::Server::Host &
...
+    return canAnswer;
+}
+
+bool Server::CanAnswerQueryTransaction(const Server::QueryTransaction &aQuery, const char *aHostFullName)
+{
+    char         name[Name::kMaxNameSize];
+    DnsQueryType sdType;
+
+    sdType = GetQueryType(aQuery.GetResponseHeader(), aQuery.GetResponseMessage(), name);
+    return sdType == kDnsQueryResolveHost && (strcmp(name, aHostFullName) == 0);
```
wgtdkp, line 854: nit:
```suggestion
    return (sdType == kDnsQueryResolveHost) && (strcmp(name, aHostFullName) == 0);
```

src/core/net/dnssd_server.cpp
```diff
@@ -668,34 +759,381 @@ const Srp::Server::Service *Server::GetNextSrpService(const Srp::Server::Host &
...
+    {
+        mTimer.Stop();
+    }
+}
+
+void Server::FinalizeQueryTransaction(QueryTransaction &aQuery, Header::Response aResponseCode)
+{
+    char name[Name::kMaxNameSize];
+
+    OT_ASSERT(mQueryUnsubscribe != nullptr);
```
wgtdkp, line 1108:
```suggestion
    OT_ASSERT(mQueryUnsubscribe != nullptr);

    char name[Name::kMaxNameSize];
```

simonlingoogle: I think we always put variable definition at the top.

src/core/net/dnssd_server.cpp
simonlingoogle: Note that it's a static method.

wgtdkp (approved):
LGTM :+1:  Looks good overall! A few nits:

simonlingoogle:
All checks are resolved except API version conflict. 
Any comments on the implementation? @abtink @librasungirl

include/openthread/dns.h
```diff
@@ -140,6 +141,93 @@ void otDnsInitTxtEntryIterator(otDnsTxtEntryIterator *aIterator, const uint8_t *
...
 
+/**
+ * This function is called when a DNS-SD query subscribes a service or service instance.
+ *
+ * @param[in] aContext      A pointer to the application-specific context.
+ * @param[in] aFullName     The null-terminated full service name (e.g. "_ipps._tcp.default.service.arpa."), or
+ *                          full service instance name (e.g. "OpenThread._ipps._tcp.default.service.arpa.").
+ *
+ */
+typedef void (*otDnssdQuerySubscribe)(void *aContext, const char *aFullName);
```
abtink, line 152: How about moving these definitions in a new header file specific for `dnssd_server.h`?
I think the idea is that we use `dns.h` for all the common definitions (shared between SRP/DNS client and server) and then each module can have its own header file to define its public API.

include/openthread/dns.h
```diff
@@ -140,6 +141,93 @@ void otDnsInitTxtEntryIterator(otDnsTxtEntryIterator *aIterator, const uint8_t *
  */
 otError otDnsGetNextTxtEntry(otDnsTxtEntryIterator *aIterator, otDnsTxtEntry *aEntry);
 
+/**
+ * This function is called when a DNS-SD query subscribes a service or service instance.
+ *
+ * @param[in] aContext      A pointer to the application-specific context.
+ * @param[in] aFullName     The null-terminated full service name (e.g. "_ipps._tcp.default.service.arpa."), or
+ *                          full service instance name (e.g. "OpenThread._ipps._tcp.default.service.arpa.").
```
abtink, line 149: How would we represent the name if the `instance name` label has a `.` in it (which is allowed)?
We can consider providing the instance label as a separate string to make it easier for the user.

Thought?

include/openthread/dns.h
abtink, line 152: How about renaming to `otDnssdQuerySubscribeCallback`? (this make it more clear that it is a callback type and not a function)?

include/openthread/dns.h
```diff
@@ -140,6 +141,93 @@ void otDnsInitTxtEntryIterator(otDnsTxtEntryIterator *aIterator, const uint8_t *
...
+
+/**
+ * This function is called when a DNS-SD query unsubscribes a service or service instance.
+ *
+ * @param[in] aContext      A pointer to the application-specific context.
+ * @param[in] aFullName     The null-terminated full service name (e.g. "_ipps._tcp.default.service.arpa."), or
+ *                          full service instance name (e.g. "OpenThread._ipps._tcp.default.service.arpa.").
+ *
+ */
+typedef void (*otDnssdQueryUnsubscribe)(void *aContext, const char *aFullName);
```
abtink, line 162: Same suggestion here? `otDnssdQueryUnsubscribeCallback`?

include/openthread/dns.h
```diff
@@ -140,6 +141,93 @@ void otDnsInitTxtEntryIterator(otDnsTxtEntryIterator *aIterator, const uint8_t *
...
+
+/**
+ * This structure represents information of a discovered service instance for a DNS-SD query.
+ *
+ */
+typedef struct otDnssdServiceInstanceInfo
+{
+    const char *   mFullName;  ///< Full instance name (e.g. "OpenThread._ipps._tcp.default.service.arpa.").
+    const char *   mHostName;  ///< Host name.
+    otIp6Address   mAddress;   ///< Host IPv6 address.
```
abtink, line 172: If there are more than one address associated with host, do we give the first one? if so may be good to mention it here?

include/openthread/dns.h
```diff
@@ -140,6 +141,93 @@ void otDnsInitTxtEntryIterator(otDnsTxtEntryIterator *aIterator, const uint8_t *
...
+ * @note @p aSubscribe and @p aUnsubscribe should be both set or unset.
+ *
+ * @param[in] aInstance     The OpenThread instance structure.
+ * @param[in] aContext      A pointer to the application-specific context.
+ * @param[in] aSubscribe    A pointer to the callback function to subscribe a service or service instance.
+ * @param[in] aUnsubscribe  A pointer to the callback function to unsubscribe a service or service instance.
+ *
+ */
+void otDnssdQuerySetCallbacks(otInstance *            aInstance,
+                              void *                  aContext,
```
abtink, line 204: How about moving `aContext` as the last parameter?

A side note here:
In general, unfortunately we don't follow same pattern for callbacks definition. In some callbacks the `aContext` is the first parameter (e.g. in `coap.h` callbacks),  in some others it is last (in `link.h` or `thread.h`). But when registering/setting callback it seems like we put the `aContext` to be always last param.
I have been aiming to harmonize this for a while, but it would be a bigger change and impacts public OT APIs.

src/cli/cli.cpp
```diff
@@ -1491,10 +1491,9 @@ void Interpreter::HandleDnsAddressResponse(otError aError, const otDnsAddressRes
             OutputFormat(" TTL:%u ", ttl);
             index++;
         }
-
-        OutputLine("");
     }
 
+    OutputLine("");
```
abtink, line 1496: Will this impact tests and how they expect the output?

 I think the original DNS CLI output included error on same line as `"DNS response for {name} - "`

src/core/net/dnssd_server.hpp
```diff
@@ -82,21 +82,40 @@ class Server : public InstanceLocator, private NonCopyable
      */
     void Stop(void);
 
+    /**
+     * This method sets DNS-SD query callbacks.
+     *
+     * @param[in] aContext      A pointer to the application-specific context.
```
abtink, line 88: Same suggestion here: Should we move `aContext` as the last parameter?

src/core/net/dnssd_server.hpp
```diff
@@ -232,67 +230,146 @@ class Server : public InstanceLocator, private NonCopyable
...
-    static Error     FindPreviousLabel(const char *aName, uint8_t &aStart, uint8_t &aStop);
+    enum : uint32_t
+    {
+        kQueryTimeout = OPENTHREAD_CONFIG_DNSSD_QUERY_TIMEOUT,
+    };
+
+    class QueryTransaction
+    {
+    public:
+        explicit QueryTransaction()
```
abtink, line 241:
```suggestion
        explicit QueryTransaction(void)
```

src/core/net/dnssd_server.cpp
```diff
@@ -132,16 +133,90 @@ void Server::ProcessQuery(Message &aMessage, Message &aResponse, const Header &a
...
+        if (kErrorNone == ResolveByQueryCallbacks(responseHeader, *responseMessage, compressInfo, aMessageInfo))
+        {
+            resolveByQueryCallbacks = true;
+        }
+    }
+
+exit:
+    if (error == kErrorNone)
+    {
+        if (!resolveByQueryCallbacks)
```
abtink, line 156: Combine the two `if` into one?

abtink (approved):
Thanks @simonlingoogle. Looks really great. Very clean and nice implementation. 👍 
I have some small nit suggestions below.

src/cli/cli.cpp
simonlingoogle: > I think the original DNS CLI output included error on same line as `"DNS response for {name} - "`

Yes, which is actually incorrect.

This PR fixes this minor issue.

include/openthread/dns.h
simonlingoogle: I don't have strong opinion on it. Putting `aContext` before callback arguments has the advantage that the position of `aContext` won't change when we add more callback functions.

include/openthread/dns.h
simonlingoogle: In the current interface design, the implementation is responsible for identifying what `@p aFullName` is. It can be either of:
1. service name
2. service instance name
3. host name

I think it's possible to correctly identify the `instance name` even when it contains a `.` in it. So, the interface should have included enough information. 

The benefit of the current interface is that the DNS-SD server in OT core don't need to further split the full name into `instance name` and `full service name`, which needs 256B buffers. 

The alternative is to provide different interfaces for different names:
1. otDnssdQuerySubscribeService(aFullServiceName)
2. otDnssdQuerySubscribeServiceInstance(aFullServiceName, aInstanceName)
3. otDnssdQuerySubscribeHost(aFullHostName)

Together with the 3 corresponding "Unsubscribe" APIs, there will be totally 6 callbacks to setup. I prefer the current interfaces (with only 2 callbacks) because it's simpler for the OT core. 

Thoughts?

include/openthread/dns.h
simonlingoogle: Should be better to allow multiple IP6 addresses. Will enhance.

simonlingoogle:
@abtink 
Thanks for your comments. Please also take a look at https://github.com/openthread/ot-br-posix/pull/706 which implements `Discovery Proxy` on mDNSResponder.

include/openthread/dns.h
abtink: The current model SGTM.

include/openthread/dnssd_server.h
```diff
@@ -0,0 +1,180 @@
...
+ */
+typedef struct otDnssdHostInfo
+{
+    uint8_t             mAddressNum; ///< Number of host IPv6 addresses.
+    const otIp6Address *mAddresses;  ///< Host IPv6 address.
+    uint32_t            mTtl;        ///< Service TTL (in seconds).
+} otDnssdHostInfo;
+
+/**
+ * This function sets DNS-SD server query callbacks. The DNS-SD server calls @p aSubscribe to subscribe to a service or
```
jwhui, line 132: nit: first sentence is the brief - keep in its own block. Add an empty line after the first sentence. Other comment blocks should be updated as well.

include/openthread/dnssd_server.h
```diff
@@ -0,0 +1,180 @@
...
+ * service instance to resolve a DNS-SD query and @p aUnsubscribe to unsubscribe when the query is resolved or timeout.
+ *
+ * @note @p aSubscribe and @p aUnsubscribe should be both set or unset.
```
jwhui, line 135:
```suggestion
 * @note @p aSubscribe and @p aUnsubscribe must be both set or unset.
```

src/core/net/dnssd_server.cpp
```diff
@@ -132,16 +142,87 @@ void Server::ProcessQuery(Message &aMessage, Message &aResponse, const Header &a
...
+
+#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
+    // Answer the questions
+    response = ResolveBySrp(responseHeader, *responseMessage, compressInfo);
+#endif
+
+    // Resolve the question using query callbacks if SRP server failed to resolve the questions.
+    if (responseHeader.GetAnswerCount() == 0)
+    {
+        if (kErrorNone == ResolveByQueryCallbacks(responseHeader, *responseMessage, compressInfo, aMessageInfo))
```
jwhui, line 156: nit: combine with above if-statement using `&&`?

src/core/net/dnssd_server.cpp
```diff
@@ -668,34 +749,386 @@ const Srp::Server::Service *Server::GetNextSrpService(const Srp::Server::Host &
...
     return error;
 }
-#endif // OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
+
+Server::QueryTransaction *Server::NewQuery(const Header &          aResponseHeader,
+                                           Message &               aResponseMessage,
+                                           const NameCompressInfo &aCompressInfo,
+                                           const Ip6::MessageInfo &aMessageInfo)
+{
+    QueryTransaction *newQeury = nullptr;
```
jwhui, line 784:
```suggestion
    QueryTransaction *newQuery = nullptr;
```

src/core/net/dnssd_server.hpp
```diff
@@ -82,21 +82,42 @@ class Server : public InstanceLocator, private NonCopyable
      */
     void Stop(void);
 
+    /**
+     * This method sets DNS-SD query callbacks.
+     *
+     * @param[in] aContext      A pointer to the application-specific context.
```
jwhui, line 88: nit: update doxygen order to match arg order?

include/openthread/dnssd_server.h
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * This function is called when a DNS-SD query unsubscribes either of:
+ *      1. a service name.
+ *      2. a service instance name.
+ *      3. a host name.
+ *
+ * The DNS-SD query implementation is responsible for identifying what @p aFullName is.
+ *
+ * @note There can be multiple subscription to the same name. DNS-SD query implementation should  record the number of
```
jwhui, line 92:
```suggestion
 * @note There can be multiple subscription to the same name. DNS-SD query implementation should record the number of
```

src/core/net/dnssd_server.cpp
```diff
@@ -668,34 +749,386 @@ const Srp::Server::Service *Server::GetNextSrpService(const Srp::Server::Host &
...
+        }
+
+        if (HasQuestion(aQuery.GetResponseHeader(), aQuery.GetResponseMessage(), aInstanceInfo.mHostName,
+                        ResourceRecord::kTypeAaaa) == !additional)
+        {
+            for (uint8_t i = 0; i < aInstanceInfo.mAddressNum; i++)
+            {
+                const Ip6::Address &address = static_cast<const Ip6::Address &>(aInstanceInfo.mAddresses[i]);
+
+                OT_ASSERT(!address.IsUnspecified() && !address.IsLinkLocal() && !address.IsMulticast() &&
```
jwhui, line 883: Are we assuming that the querier will ensure that such addresses are never provided?

include/openthread/dnssd_server.h
```diff
@@ -0,0 +1,180 @@
...
+
+#include <openthread/error.h>
+#include <openthread/ip6.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @addtogroup api-dnssd-server
```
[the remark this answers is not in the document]
simonlingoogle: it was added.

src/core/net/dnssd_server.cpp
simonlingoogle: Yes we are. We are trying to be strict in DNS-SD server when input from Discovery Proxy implementations so as to catch bugs earlier.

jwhui: Should we call out these constraints in the API docs?

jwhui (approved):

PR #6194, opened 2021-02-22
[dns-client] add arg validation on ResolveService()

jwhui:


src/cli/cli.cpp
```diff
@@ -1535,34 +1535,33 @@ void Interpreter::HandleDnsBrowseResponse(otError aError, const otDnsBrowseRespo
     char             label[OT_DNS_MAX_LABEL_SIZE];
     uint8_t          txtBuffer[255];
     otDnsServiceInfo serviceInfo;
+    uint16_t         index = 0;
 
-    IgnoreError(otDnsBrowseResponseGetServiceName(aResponse, name, sizeof(name)));
+    SuccessOrExit(aError);
+
+    SuccessOrExit(aError = otDnsBrowseResponseGetServiceName(aResponse, name, sizeof(name)));
```
abtink, line 1542: @jwhui if I am not mistaken, I think we should never get a parse error for the service name here in the callback. It should always work.

Here is the flow:
- When we call `otDnsClientBrowse()` we pass in the `aServiceName`.
- I expect if the name is not well-formed, we get `PARSE_ERROR` from `otDnsClientBrowse()` itself.
   - I noticed it is not documented but I think we should get this error: `Client::Browse() -> StartQuery() -> AllocateQuery()` which then tries to append the name to `Query` message and would fail if not well-formed.
- Later on from `Browse` callback, we would be reading the "service name" from same `Query` message instance in which we wrote it earlier, so it should be well-formed (no parse error).

Maybe I am missing something?

jwhui: Yes, you are correct. Digging deeper, when `NULL` is passed as `aInstanceLabel` to `otDnsClientResolveService()`, then a subsequent call to `ReadLabel()` returns `OT_ERROR_NOT_FOUND`. I've updated this PR to add some additional parameter checking to `otDnsClientResolveService()`. Thoughts?

abtink: SGTM.

abtink (approved):
LGTM. 👍

PR #6195, opened 2021-02-23
[github-actions] move POSIX expects to POSIX workflow

bukepo:


.github/workflows/posix.yml
```diff
@@ -243,6 +315,8 @@ jobs:
 
   upload-coverage:
     needs:
+    - expects-macOS
```
Moandor-y, line 318: We need to add uploading steps (like [line 93](https://github.com/openthread/openthread/pull/6195/files#diff-258c8e6cdf61e87e9d94e06c76fb5e23f602423313565dcb8a46afe648587aceR93)) in `expects-macOS` if we are going to upload its coverage data.

bukepo: I failed to enable coverage on macOS, and we can try it later.

Moandor-y: Then I think we should remove this dependency first, and add it when coverage on macOS is ready.

Moandor-y (approved):
LGTM :+1:

jwhui (approved):

PR #6197, opened 2021-02-23
[dnssd] set DNS-SD server port number

simonlingoogle:


etc/cmake/options.cmake
```diff
@@ -172,6 +172,11 @@ if(OT_DNSSD_SERVER)
     target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_DNSSD_SERVER_ENABLE=1")
 endif()
 
+option(OT_DNSSD_SERVER_PORT "specify DNS-SD server port" "")
+if(NOT OT_DNSSD_SERVER_PORT STREQUAL "OFF")
+    target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_DNSSD_SERVER_PORT=${OT_DNSSD_SERVER_PORT}")
+endif()
```
abtink, line 178: Can we in this case rely on `OPENTHREAD_CONFIG_DNSSD_SERVER_PORT` definition directly?

 I suggest we consider carefully which parameters we do expose through the cmake/build configs.

My view on this is that would be better to have cmake option to enable/disable bigger features, but for configuration parameters related to a specific feature I think would be better to rely on the config headers (e.g. project specific config header).

The challenge I see is that if we start exposing config parameters through cmake (or other build systems), we end up with a lot of cmake/build configs (basically mirroring the same set as defined in `config/` headers).

simonlingoogle: You suggestion sgtm. I will close this PR for now.

Issue #6201, opened 2021-02-24
SED battery drained quickly because of frequent re-attach sequences


dismirlian:
Under the following circumstances a SED's battery is drained very quickly:
- device with very marginal connectivity (de-attaches frequently)
- default `OPENTHREAD_CONFIG_MLE_ATTACH_BACKOFF_MINIMUM_INTERVAL` (251ms)
- "fast" poll period (in our case, 11s)

This happens because the SED de-attaches frequently, and tries to attach very fast (the default `OPENTHREAD_CONFIG_MLE_ATTACH_BACKOFF_MINIMUM_INTERVAL` causes 5 attach attempts in less than ~8s). A workaround may be to increase `OPENTHREAD_CONFIG_MLE_ATTACH_BACKOFF_MINIMUM_INTERVAL`, but this would result in slow re-attaches for devices with robust links.

Another idea would be to implement some kind of "backoff": OT could guarantee that there is some minimum (configurable) interval _between_ re-attach sequences. What do you think?

Thank you.

jwhui:
@dismirlian , thanks for raising this issue.

As you probably know, OpenThread actually implements a backoff mechanism using the following configurations:

- [OPENTHREAD_CONFIG_MLE_ATTACH_BACKOFF_ENABLE](https://github.com/openthread/openthread/blob/d24327fd71a96bda0619f6b6331184e313b73909/src/core/config/mle.h#L121)
- [OPENTHREAD_CONFIG_MLE_ATTACH_BACKOFF_ENABLE](https://github.com/openthread/openthread/blob/d24327fd71a96bda0619f6b6331184e313b73909/src/core/config/mle.h#L135)
- [OPENTHREAD_CONFIG_MLE_ATTACH_BACKOFF_MAXIMUM_INTERVAL](https://github.com/openthread/openthread/blob/d24327fd71a96bda0619f6b6331184e313b73909/src/core/config/mle.h#L149)
- [OPENTHREAD_CONFIG_MLE_ATTACH_BACKOFF_JITTER_INTERVAL](https://github.com/openthread/openthread/blob/d24327fd71a96bda0619f6b6331184e313b73909/src/core/config/mle.h#L163)

Currently, the backoff interval should only be reset to the minimum after successfully attaching. Is this what you are observing? Or are you proposing to increase the backoff interval even after the device is successfully attaching?

dismirlian:
[...]

PR #6202, opened 2021-02-25
[mle] add wait time for resetting attach backoff interval after attach

abtink:
This commit adds a new behavior related to MLE attach backoff feature.
It adds a configurable delay wait interval before resetting the attach
backoff interval to its minimum value upon a successful attach.
`OPENTHREAD_CONFIG_MLE_ATTACH_BACKOFF_DELAY_TO_RESET_BACKOFF_INTERVAL`
specifies the delay interval.

If it is set to zero then the device resets its backoff attach
interval immediately after a successful attach. With a non-zero value,
if after a successful attach, the device happens to detach within the
delay interval, the reattach process resumes with the previous backoff
interval (as if the attach did not happen).

This behavior is helpful in the situation where a battery-powered
device has poor link quality to its parent and therefore attaches and
detaches frequently from its parent.  Using a non-zero wait interval
ensures that the attach backoff does not reset on each attach and that
the device does not drain its battery quickly trying to re-attach too
frequently.

------

Should help address https://github.com/openthread/openthread/issues/6201

### DIFF
--- a/src/core/config/mle.h
+++ b/src/core/config/mle.h
@@ -164,6 +164,26 @@
 #endif
 
 /**
+ * @def OPENTHREAD_CONFIG_MLE_ATTACH_BACKOFF_DELAY_TO_RESET_BACKOFF_INTERVAL
+ *
+ * Specifies the delay wait interval (in milliseconds) used by attach backoff feature after a successful attach before
+ * it resets the current backoff interval back to the minimum value.
+ *
+ * If it is set to zero then the device resets its backoff attach interval immediately after a successful attach. With
+ * a non-zero value, if after a successful attach, the device happens to detach within the delay interval, the reattach
+ *  process resumes with the previous backoff interval (as if the attach did not happen).
+ *
+ * This behavior is helpful in the situation where a battery-powered device has poor link quality to its parent and
+ * therefore attaches and detaches frequently from the parent.  Using a non-zero wait interval ensures that the attach
+ * backoff interval does not reset on each attach and that the device does not drain its battery quickly trying to
+ * re-attach too frequently.
+ *
+ */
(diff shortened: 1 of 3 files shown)


abtink:
Adding some details on this change (while the `diff` is relatively small, there is a lot paths/flows in the code to take into account, so will try to describe it to help with review):
- When device becomes a child (`SetStateChild()`) instead of immediatley clearing `mAttachCounter` and stopping the timer, we start the timer for the `DelayToReset` value.
- If we stay attached till the timer fires, `mAttachCounter` will be set to zero from `HandleAttachTimer()` under the `case kAttachStateIdle`.

Now the thing to consider is how else the `mAttachTimer` may be used/started and if we are ok in all cases. `mAttachTimer` is used for two purposes:
1. It is used when we are trying to find a parent (used to switch between states, like sending parent request to routers, then to REEDs, etc).     
   - This can be used when we are already attached, (e.g., from the "periodic parent search" feature or if merging between partitions, etc) or if/when device is detached.
   - In all cases the process starts by calling `BecomeChild()`.
   - In new code from this PR, we check in `BecomeChild()` if we are already attached and if so we reset the `mAttachCounter` to zero. 
      - This should cover all cases like "periodic parent search" where the `mAttachTimer` may be re-schedueld before we get the `DelayToReset` time
   - `SetStateDetached()` will stop the `mAttachTimer`
       - This covers the main behavior, i.e. if we detach before the `DelayToReset` time expires, `mAttachCounter` would remian the same as before and we start with same attach backoff interval as before.

2. The other use of `mAttachTimer` is to delay processing of a received Announce message from `HandleAnnounce()`
   - In this case, we add new code from this PR to explicitly reset the `mAttachCounter`.

jwhui (approved):

PR #6205, opened 2021-02-25
[nrf528xx] enlarge internal heap size

wgtdkp:
The ECDSA signing of the SRP client requires additional internal heap.
This commit enlarges the internal heap size to `4000` bytes. The last value
I tried but failed is `3800`.


abtink (approved):
👍

LuDuda (approved):

jwhui (approved):

PR #6206, opened 2021-02-25
[mle] update 'ToString()' methods in MLE to use string array

abtink:
This commit changes `ToString()` methods in MLE module which convert
from a given `enum` value to a string to use string array tables. It
also adds `statci_assert` checks to verify the values of `enum`
maps correctly to the array indexes.

### DIFF
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -4116,132 +4116,116 @@ void Mle::LogError(MessageAction aAction, MessageType aType, otError aError)
 
 const char *Mle::MessageActionToString(MessageAction aAction)
 {
-    const char *str = "Unknown";
+    static const char *const kMessageActionStrings[] = {
+        "Send",           // (0) kMessageSend
+        "Receive",        // (1) kMessageReceive
+        "Delay",          // (2) kMessageDelay
+        "Remove Delayed", // (3) kMessageRemoveDelayed
+    };
 
-    switch (aAction)
-    {
-    case kMessageSend:
-        str = "Send";
-        break;
-    case kMessageReceive:
-        str = "Receive";
-        break;
-    case kMessageDelay:
-        str = "Delay";
-        break;
-    case kMessageRemoveDelayed:
-        str = "Remove Delayed";
-        break;
-    }
+    static_assert(kMessageSend == 0, "kMessageSend value is incorrect");
+    static_assert(kMessageReceive == 1, "kMessageReceive value is incorrect");
+    static_assert(kMessageDelay == 2, "kMessageDelay value is incorrect");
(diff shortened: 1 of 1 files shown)


jwhui (approved):

src/core/thread/mle.cpp
```diff
@@ -4116,132 +4116,116 @@ void Mle::LogError(MessageAction aAction, MessageType aType, otError aError)
...
-        str = "Receive";
-        break;
-    case kMessageDelay:
-        str = "Delay";
-        break;
-    case kMessageRemoveDelayed:
-        str = "Remove Delayed";
-        break;
-    }
+    static_assert(kMessageSend == 0, "kMessageSend value is incorrect");
```
simonlingoogle, line 4126: I guess the purpose of this assert is to make sure that `kMessageActionStrings[kMessageSend]` is "Send". However we don't require that `kMessageSend` must be 0. I think it's possible to verify the actual purpose. 

```

constexpr bool strings_equal(char const *a, char const *b)
{
    return *a == *b && (*a == '\0' || strings_equal(a + 1, b + 1));
}

const char *Mle::MessageActionToString(MessageAction aAction)
{
    static constexpr const char *const kMessageActionStrings[] = {
        "Send",           // (0) kMessageSend
        "Receive",        // (1) kMessageReceive
        "Delay",          // (2) kMessageDelay
        "Remove Delayed", // (3) kMessageRemoveDelayed
    };

    static_assert(strings_equal("Send", kMessageActionStrings[kMessageSend]), "kMessageSend value is incorrect");
    static_assert(strings_equal("Receive", kMessageActionStrings[kMessageReceive]),
                  "kMessageReceive value is incorrect");
    static_assert(strings_equal("Delay", kMessageActionStrings[kMessageDelay]), "kMessageDelay value is incorrect");
    static_assert(strings_equal("Remove Delayed", kMessageActionStrings[kMessageRemoveDelayed]),
                  "kMessageRemoveDelayed value is incorrect");

    return kMessageActionStrings[aAction];
}
```

Thoughts?

simonlingoogle (commented):
LGTM :+1:

src/core/thread/mle.cpp
abtink: Nice suggestion. 👍  I like it! Specially I think this helps for cases where we have `#if` checks in enum definition (the enum list of numbers is not always same). For cases where we have simple enum lists I think current model would be ok/fine. 

There is one thing I think we need to be careful here and double-check, i.e. if all toolchains (we care about) can correctly handle a `static constexpr const` array definition (we do want the array to reside in code/flash space and don't want the complier to store it in RAM). I recall some weird behaviors related to `static consexpr` local definition on some older toolchains (using RAM instead of flash which is technically allowed).

PR #6209, opened 2021-02-26
[script] adding 'update-makefiles.py' script

abtink:
This commit adds `update-makefiles.py` script which helps update
different make/build files supported by OpenThread project including
`CMakeList.txt`, `BUILD.gn`, `Android.mk`, `Andriod.bp`, and
`Makefile.am` based on the source (`cpp`, `hpp`, and `h`) files in
`./src/core` and `./include/openthread` folders. This script is
intended to make it easier for developer when adding/removing/renaming
source files within OT core to help update all the make/build files.

This commit also updates `Makefile.am` ensuring the lists are sorted
alphabetically and adding some missing file (this is generated by 
running the script).

### DIFF
--- a/script/update-makefiles.py
+++ b/script/update-makefiles.py
@@ -1,0 +1,204 @@
+#!/usr/bin/env python3
+#
+#  Copyright (c) 2021, The OpenThread Authors.
+#  All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions are met:
+#  1. Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#  2. Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in the
+#     documentation and/or other materials provided with the distribution.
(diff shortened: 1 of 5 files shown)


script/update-makefiles.py
```diff
@@ -0,0 +1,200 @@
...
+                new_lines.extend(new_list)
+                state = STATE_MATCH_END
+        elif state == STATE_MATCH_END:
+            if line.startswith(end_string):
+                new_lines.append(line)
+                state = STATE_DONE
+        else:
+            new_lines.append(line)
+
+    write_txt_file(file_name, new_lines)
```
simonlingoogle, line 96:
```suggestion
    if (state != STATE_DONE):
        raise RuntimeError('failed to update build file: %s' % file_name)

    if new_lines != lines: write_txt_file(file_name, new_lines)
```
Make sure we really updated the file?

simonlingoogle (approved):
LGTM :+1: 

Thanks for this great tool!

script/update-makefiles.py
abtink: Good suggestions. Thanks.

jwhui (approved):

PR #6212, opened 2021-02-26
[efr32] revert removal of support for multiple efr32BandConfig's

lmnotran:
In https://github.com/openthread/openthread/pull/6154, the `radio.c` changes included the removal of a lot of code wrapped in `#if RADIO_CONFIG_915MHZ_OQPSK_SUPPORT` macros. The problem is there was still one section of code which was not removed. 

https://github.com/openthread/openthread/blob/b45f287629c4c54e855262edd474f0658804feaa/examples/platforms/efr32/src/radio.c#L552-L556

This caused the 2.4GHz `sBandConfig` to be overwritten with sub-GHz config, ultimately causing the efr32mg12 BRD4170A to assert somewhere in radio initialization. 

This PR reverts the removal of support for multiple `efr32BandConfig`s.

An alternative fix would be to remove the final `#if RADIO_CONFIG_915MHZ_OQPSK_SUPPORT` section in `efr32ConfigInit()`

### DIFF
--- a/examples/platforms/efr32/src/radio.c
+++ b/examples/platforms/efr32/src/radio.c
@@ -114,6 +114,15 @@
 #define QUARTER_DBM_IN_DBM 4
 #define US_IN_MS 1000
 
+enum
+{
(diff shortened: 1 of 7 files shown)


romacdon (approved):
LGTM

jwhui (approved):

PR #6214, opened 2021-02-27
[mesh-forwarder] add helper 'RemoveMessageIfNoPendingTx()'

abtink:
This commit adds a helper method `RemoveMessageIfNoPendingTx()` in
`MeshForwarder` which removes (dequeues from `mSendQueue` and frees
 it) if there is no pending direct or indirect transmission.

 This commit also changes `UpdateSendMessage()` to exit early when
 there are remaining fragments of the message to be sent.


simonlingoogle (approved):
LGTM  :+1:

jwhui (approved):

PR #6216, opened 2021-02-28
[border-router] test host leaves and rejoins

wgtdkp:
This commit tests if a host is still reachable after leaving and rejoining the infrastructure network. This makes sure that the routing manager handles router solicitations.

### DIFF
--- a/tests/scripts/thread-cert/border_router/test_single_border_router.py
+++ b/tests/scripts/thread-cert/border_router/test_single_border_router.py
@@ -309,6 +309,19 @@ class SingleBorderRouter(thread_cert.TestCase):
         self.assertTrue(self.nodes[HOST].ping(self.nodes[ROUTER1].get_ip6_address(config.ADDRESS_TYPE.OMR)[0],
                                               backbone=True))
 
+        #
+        # Case 5. Test if the linux host is still reachable if rejoin the network.
+        #
+
+        self.nodes[HOST].disable_ether()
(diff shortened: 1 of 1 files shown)


tests/scripts/thread-cert/border_router/test_single_border_router.py
```diff
@@ -309,6 +309,19 @@ def test(self):
...
                                               backbone=True))
 
+        #
+        # Case 5. Test if the linux host is still reachable if rejoin the network.
+        #
+
+        self.nodes[HOST].disable_ether()
+        self.simulator.go(10)
+        self.nodes[HOST].enable_ether()
+        self.simulator.go(10)
```
simonlingoogle, line 319:
```suggestion
        self.simulator.go(3)
```

use smaller delay so as to make sure that RS is handled quickly ?

simonlingoogle (approved):
LGTM :+1: 

with a small nit

tests/scripts/thread-cert/border_router/test_single_border_router.py
wgtdkp: I tried with `2` but failed intermediately. I think the delay comes from the way how Linux resets a network interface and it is not determined.

simonlingoogle: Fine to me to keep it 10s.

jwhui (approved):

PR #6217, opened 2021-03-01
[config] perform OT config checks in the related module

abtink:
This commit adds the checking of the OT config options directly in the
header files of the related module, e.g., the check for DUA feature
being used along with Thread 1.2 is now done in `dua_manager.hpp`. The
`openthread-core-config-check.h` is now used to track the list of
removed, renamed, or replaced configs. This change makes it easier to
find and maintain the dependency/relationship between OT configs
(easier to find them in the related header file than within the list
in `openthread-core-config-check.h`).


simonlingoogle (approved):
LGTM 👍

jwhui (approved):

PR #6221, opened 2021-03-01
[message] use opaque 'otMessage' and add 'otMessageBuffer'

abtink:
This commit defines `otMessage` to be an opaque public `struct`. This
change ensures users of OpenThread APIs can only get `otMessage`
instances through OpenThread public APIs and callbacks. It also hides
the `mNext` field which is not intended for use/access by OT API users
(OT users may assume `mNext` is pointing to the next message when the
message is placed in an `otMessageQueue` which is not correct since
`mNext` is used to chain the fixed-size `Buffer` instances of the same
`Message` instance).

This commit adds `otMessageBuffer` in `platform/messagepool.h` which
is then used when the platform is configured to manage the buffer pool.
The `otMessageBuffer` struct defines `mNext` to ensure the returned
`otMessageBuffer` instances from platform layer are aligned correctly
in memory (i.e., they can be safely cast to `ot::Buffer`).

### DIFF
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (80)
+#define OPENTHREAD_API_VERSION (81)
 
 /**
  * @addtogroup api-instance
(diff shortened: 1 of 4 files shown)


simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6223, opened 2021-03-02
[mac] only append header termination IE2 when data payload is present

brianljt:


brianljt:
According to IEEE802.15.4, header termination IE2 is present only when header IE and payload data is present.

image

Irving-cl (approved):
LGTM :+1:

brianljt:
codecov not updated correctly. Issue #6133

jwhui (approved):

PR #6227, opened 2021-03-03
[dataset] change updater to use DatasetManager

jwhui:
This PR changes the dataset updater implementation to leverage
`MeshCoP::DatasetManager`, which already provides a mechanism to
communicate Pending Dataset updates with the Leader. The
DatasetManager implementation ensures that only one CoAP transaction
is outstanding at a time.
    
This commit also ensures that any unspecified values are copied from
the Active Dataset.

This commit also add:
- CLI implementation of Dataset Updater APIs
- CI test for Dataset Updater


tests/scripts/thread-cert/test_dataset_updater.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python3
+#
+#  Copyright (c) 2020, The OpenThread Authors.
```
abtink, line 3: small nit: 2021?

abtink (approved):
Looks great! Thanks. 👍

PR #6228, opened 2021-03-03
[spinel] handle link metrics

piotrkoziar:
This PR allows the spinel interface to deal with link metrics.

There are 7 new properties introduced:
* `THREAD_LINK_METRICS_QUERY`
* `THREAD_LINK_METRICS_QUERY_RESULT`
* `THREAD_LINK_METRICS_PROBE`
* `THREAD_LINK_METRICS_MGMT_ENH_ACK`
* `THREAD_LINK_METRICS_MGMT_ENH_ACK_IE`
* `THREAD_LINK_METRICS_MGMT_FORWARD`
* `THREAD_LINK_METRICS_MGMT_RESPONSE`

PR #6231, opened 2021-03-04
[spinel] add multicast listener registration

piotrkoziar:
This PR adds Multicast Listeners Register Request and Response spinel properties:

* `SPINEL_PROP_THREAD_MLR_REQUEST`
* `SPINEL_PROP_THREAD_MLR_RESPONSE`

The numbers of the properties have been adjusted to match the ones from https://github.com/openthread/openthread/pull/6228/.


src/ncp/ncp_base.hpp
```diff
@@ -337,6 +337,16 @@ class NcpBase
     void        HandleJoinerCallback(otError aError);
 #endif
 
+    static void HandleMlrRegResult_Jump(void *              aContext,
```
simonlingoogle, line 340:
```suggestion
    static void HandleMlrRegResult(void *              aContext,
```
Usually we use same method name for the static and non-static methods.

src/ncp/ncp_base_mtd.cpp
```diff
@@ -259,6 +259,87 @@ template <> otError NcpBase::HandlePropertyGet<SPINEL_PROP_THREAD_CSL_CHANNEL>(v
...
+    SuccessOrExit(mEncoder.BeginFrame(SPINEL_HEADER_FLAG | SPINEL_HEADER_IID_0, SPINEL_CMD_PROP_VALUE_IS,
+                                      SPINEL_PROP_THREAD_MLR_RESPONSE));
+
+    SuccessOrExit(mEncoder.WriteUint8(ThreadErrorToSpinelStatus(aError)));
+    SuccessOrExit(mEncoder.WriteUint8(aMlrStatus));
+
+    for (size_t i = 0U; i < aFailedAddressNum; ++i)
+    {
+        SuccessOrExit(mEncoder.WriteIp6Address(aFailedAddresses[i]));
+    }
```
simonlingoogle, line 334:
```suggestion
    if (aError == OT_ERROR_NONE) {
        for (size_t i = 0U; i < aFailedAddressNum; ++i)
        {
            SuccessOrExit(mEncoder.WriteIp6Address(aFailedAddresses[i]));
        }
    }
```
Nit: safer
The doc of `otIp6RegisterMulticastListenersCallback` states that `aFailedAddresses` and `aFailedAddressNum` are only defined when `aError` is `OT_ERROR_NONE`.

simonlingoogle (approved):
LGTM :+1: with nits

Thanks.

ankuns:
@googlebot I consent.

src/ncp/ncp_base.hpp
piotrkoziar: I think we should leave the `_Jump` postfix due to other functions using it in this file (`HandleJamStateChange_Jump`, `HandleCommissionerPanIdConflict_Jump`, and others). I believe using the different rules for `HandleMlrRegResult` will be an unnecessary complication.

src/lib/spinel/spinel.h
```diff
@@ -2945,6 +2953,46 @@ enum
...
+     */
+    SPINEL_PROP_THREAD_MLR_REQUEST = SPINEL_PROP_THREAD_EXT__BEGIN + 52,
+
+    /// Multicast Listeners Register Response
+    /** Format `CCA(6)` - Unsolicited notifications only
+     * Required capability: `SPINEL_CAP_NET_THREAD_1_2`
+     *
+     * `C`: Status
+     * `C`: MlrStatus (The Multicast Listener Registration Status)
+     * `A(6)`: Array of IPv6 addresses that failed to be updated on the primary BBR
```
abtink, line 2989: I suggest that we include this `A()` in a struct` as `t(A(6))`.

The reason is to allow other fields to be added to this property in futrue (not rely on end of message to determine the array end)?

abtink (commented):
Thanks @piotrkoziar. Looks great.
One suggestion below.

src/ncp/ncp_base_mtd.cpp
```diff
@@ -259,6 +259,90 @@ template <> otError NcpBase::HandlePropertyGet<SPINEL_PROP_THREAD_CSL_CHANNEL>(v
...
+    otIp6Address addresses[kIPv6AddressesNumMax];
+    uint8_t      addressesCount = 0U;
+    bool         timeoutPresent = false;
+    uint32_t     timeout;
+
+    SuccessOrExit(error = mDecoder.OpenStruct());
+
+    while (mDecoder.GetRemainingLengthInStruct())
+    {
+        VerifyOrExit(addressesCount < kIPv6AddressesNumMax, error = OT_ERROR_PARSE);
```
LuDuda, line 275:
```suggestion
        VerifyOrExit(addressesCount < kIPv6AddressesNumMax, error = OT_ERROR_NO_BUFS);
```

LuDuda (approved):
LGTM :+1:

piotrkoziar:
Applied suggestions and rebased.

src/ncp/ncp_base_mtd.cpp
```diff
@@ -262,6 +262,93 @@ template <> otError NcpBase::HandlePropertyGet<SPINEL_PROP_THREAD_CSL_CHANNEL>(v
...
+        ++addressesCount;
+    }
+
+    SuccessOrExit(error = mDecoder.CloseStruct());
+
+    while (mDecoder.GetRemainingLengthInStruct())
+    {
+        SuccessOrExit(error = mDecoder.OpenStruct());
+
+        uint8_t paramId;
```
abtink, line 289: Suggest moving the `uint8_t paramId` definition  to the beginning of `{` block.

Some toolchains complain if variables are defined after potential jump to `exit`.

I think it is also required/mentioned in our [style guide](https://github.com/openthread/openthread/blob/main/STYLE_GUIDE.md#white-space).

src/lib/spinel/spinel.h
```diff
@@ -2955,6 +2963,46 @@ enum
      */
     SPINEL_PROP_THREAD_DOMAIN_NAME = SPINEL_PROP_THREAD_EXT__BEGIN + 44,
 
+    /// Multicast Listeners Register Request
+    /** Format `t(A(6))A(t(CD))` - Write-only
+     * Required capability: `SPINEL_CAP_NET_THREAD_1_2`
+     *
+     * `t(A(6))`: Array of IPv6 multicast addresses
+     * `A(t(CD))`: Array of structs holding optional parameters as follows
```
abtink, line 2971: This model of having an array of mapping id to value works fine (sort of like a dictionary) and I think is good here.

The other model I can suggest for optional params (I think we use in some spinel definitions) is to place each param in its own `struct`. When parsing, if struct is empty, we assume it is not provided, otherwise parse the value from the struct.

abtink (approved):
LGTM. Thanks 👍 

Just one suggestion below about moving the variable definition to top of `{}` block. Thanks.

src/lib/spinel/spinel.h
ankuns: @abtink What you propose has some drawbacks:
- if optional item is not sent the struct is empty, but it's length still needs to be sent, 
- the order of structs means which optional parameter is which (non-empty struct means passev value to some parameter), when there are many such parameters and only last one is provided value all previous need to be sent as zero-length what takes again some bytes to encode the struct length. This is not the case here but it could turn into such as the property evolves. 
However it has advantages:
- it seems simple to parse
- there is possibly such convention already as you say

From your words "... I can suggest ... " I'm not sure if you accept proposed solution or should we change the model for optional parameter as you suggest (Could you please share some sample property where it's used?).

abtink: @ankuns I think the current model is good here and no need to change. :)

I just suggested it as another way to have in mind for other/new properties.

I agree each model has its pros and cons:
- The dictionary model is good if we have a bunch of optional params. 
   - The downside is we need to define an `id` for each param and then include it in the message (so if we have a lot of optional parameters and most often use many of them, the dictionary model can add the overhead of including the id bytes).
   - We use this model  for dataset (where we have a lot of optional param and a subset can be provided).
- The empty struct `t()` model has the advantage that the spinel format can mirror the ot API call format (when we have pointers to params that can be `null` if not provided). It is more useful when the params are often used. 
  - We use this in few  cases (not that many), e.g. `STREAM` props (for example for PHY data or vendor data), or `ADDRESS_CACHE_TABLE`  or somewhat similar in `MESHCOP_COMMISSIONER_JOINERS` (empty struct to indicate any joiner).

piotrkoziar:
@abtink @ankuns thanks for the comments. I pushed the commit that moves the variable definition.

abtink (approved):
LGTM. Thanks. 👍

jwhui (approved):

PR #6233, opened 2021-03-04
[dtls] make the max dtls buffer length configurable

gjc13:
On systems where OpenThread share mbedtls configuration with HTTPS and
other TLS based protocols, the TLS max content length can be way larger
than what OpenThread requires. Add an OpenThread config macro to reduce stack
consumption under such circumstance.

### DIFF
--- a/src/core/config/dtls.h
+++ b/src/core/config/dtls.h
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2016, The OpenThread Authors.
+ *  Copyright (c) 2021, The OpenThread Authors.
  *  All rights reserved.
  *
  *  Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


src/core/config/dtls.h
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * @def OPENTHREAD_DTLS_MAX_CONTENT_LEN
+ *
+ * The max length of the OpenThread dtls content buffer.
+ *
+ */
+#ifndef OPENTHREAD_DTLS_MAX_CONTENT_LEN
+#define OPENTHREAD_DTLS_MAX_CONTENT_LEN MBEDTLS_SSL_MAX_CONTENT_LEN
+#endif
```
bukepo, line 51:
```suggestion
#ifndef OPENTHREAD_CONFIG_DTLS_MAX_CONTENT_LEN
#define OPENTHREAD_CONFIG_DTLS_MAX_CONTENT_LEN MBEDTLS_SSL_MAX_CONTENT_LEN
#endif
```

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #6234, opened 2021-03-04
[git] pull dependency pull requests

### LINKED ISSUE #6184: Enhance PR dependency "Depends-On:" feature
The dependent PR  feature is useful when having changes that impact other repos.  While using it from https://github.com/openthread/openthread/pull/6157 noticed that the current code seem to require that the `"Depends-On:"` to be last line in the PR decsription text body. 

- The reason seems to be that we get `\r\n` in the text body we retrive from github API for the PR.
- In `depends_on_pr="$(grep -E "^Depends-On: *${project_name}" | cut -d# -f2)"`  the output of  `grep` will still include the `\r` character and then parsed PR number from text string would not be correct.
- When the ``"Depends-On:"`` is the last line it works since there is no `\r\n` at the end.

Adding this issue so we can later enahnce this and allow it to be on any line (would be particularly useful if we have more than one `"Depends-On:"` for same PR).

bukepo:
This commit fixes #6184

Depends-On: openthread/ot-nrf528xx#5

### DIFF
--- a/.github/workflows/size.yml
+++ b/.github/workflows/size.yml
@@ -44,8 +44,6 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v2
-      with:
-        submodules: true
     - name: Bootstrap
       if: "github.event_name == 'push'"
       run: |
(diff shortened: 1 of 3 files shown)


simonlingoogle (approved):
LGTM :+1:

wgtdkp (approved):
LGTM :+1

jwhui (approved):

PR #6236, opened 2021-03-04
[mac] fix default Thread Domain name per v1.2.1-draft4 spec

EskoDijk:
Name is updated here to specified, latest name "DefaultDomain".

Some additional info:
Note that for non-CCM Thread Devices the domain name is not even available, so the code to "get" a domain name should in fact only be available for CCM Thread Devices. (They take it out of their LDevID certificate.) The "set" Domain Name should not be possible for that CCM Thread Device.

In Border Routers that are non-CCM, there should be code to both get and set the Domain Name but I would expect that to be in the ot-br-posix repo instead! 

Removing this code from 1.2 devices could save some bytes; I could create an issue for this. The "get" function could be kept in preparation for CCM devices.

### DIFF
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -70,7 +70,7 @@ const otExtendedPanId Mac::sExtendedPanidInit = {
 const char Mac::sNetworkNameInit[] = "OpenThread";
 
 #if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
-const char Mac::sDomainNameInit[] = "Thread";
+const char Mac::sDomainNameInit[] = "DefaultDomain";
 #endif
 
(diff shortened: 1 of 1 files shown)


wgtdkp (approved):
> In Border Routers that are non-CCM, there should be code to both get and set the Domain Name but I would expect that to be in the ot-br-posix repo instead!

Openthread includes all platform independent features while ot-br-posix includes only posix platform implementation that are required for a border router devices.

EskoDijk:
Ah, ok, in that case it could be worthwhile to make the "get" call to return an empty string in case of a non-BR device? (As it does not know the real name.)  And for the "set" call to be not available i.e. a link error if used.
For a BR then, both "get" and "set" could be available.

jwhui (approved):
Thanks! 👍🏼

PR #6241, opened 2021-03-06
[mle] use designated initializers to initialize MLE message type to string

brianljt:
Instead of checking all the enum value which is not scalable when more macros are added to the MLE message types and creates more permutations, this change uses designated initializer for the string array to create 1:1 mapping of MLE message enum type to its name string.

Adding/deleting enum value or reorder enum value which is not reflected in `Mle::MessageTypeToString` will be caught at compile time.

### DIFF
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -4138,98 +4138,52 @@ const char *Mle::MessageActionToString(MessageAction aAction)
 
 const char *Mle::MessageTypeToString(MessageType aType)
 {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpedantic"
     static const char *const kMessageTypeStrings[] = {
-        "Advertisement",         // (0)  kTypeAdvertisement
-        "Announce",              // (1)  kTypeAnnounce
-        "Child ID Request",      // (2)  kTypeChildIdRequest
-        "Child ID Request",      // (3)  kTypeChildIdRequestShort
-        "Child ID Response",     // (4)  kTypeChildIdResponse
-        "Child Update Request",  // (5)  kTypeChildUpdateRequestOfParent
-        "Child Update Response", // (6)  kTypeChildUpdateResponseOfParent
-        "Data Request",          // (7)  kTypeDataRequest
-        "Data Response",         // (8)  kTypeDataResponse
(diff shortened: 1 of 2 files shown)


brianljt:
If we don't support this c99 extension exception in our coding style, we may consider some other approaches,

1. Use a string-enum pair which cost extra memory to save the enum value.
```
struct mleTypeToName
{
    const char *mName;
    enum MessageType mType;
}
```
2. Use X-macro which needs a separate header file and include this file in the middle of enum and string array definition. Ref: https://www.drdobbs.com/the-new-c-x-macros/184401387

abtink:
Thanks @brianljt. 

I think you point out the challenge with this, i.e., the support for this style of array initialization is toolchain dependent and I think we may want to avoid it due to this (to keep OT toolchain independent as much as possible, specially considering users may use custom (chip specific) toolchain to build OT stack). 

Related to using `struct kMleTypeToName` I think it will increase code size (and we also need a look up search compared to the fast array access). 
- I think (and favor) using the simple C string array model for this. I think this works fine as the conversion model (we can include `#if` checks within array definition itself and it would be fine). 
- The challenge is with the `static_assert` checking the numerical value of each enumerator and that we would need to cover all set of combinations of `#if` checks.

Here are the options I would suggest:
- We can continue to use the `static_assert` and cover all combinations (as done in current master).
   - This may be fine for now as we don't have too many combinations and I don't think we will be adding too many new cases.
   - Since this `enum` is mainly used for logging, I think it would very unlikely for a developer to change the `enum` definitions without noticing or needing to change to `MessageTypeToString` method.

- @simonlingoogle  suggested a great approach in this  https://github.com/openthread/openthread/pull/6206#discussion_r584452913 where we use `static_assert` to verify that the string from array directly matches what we want. I definitely favor and like this.
  - My concern with this was if all toolchains (we care about) can correctly handle a `static constexpr const` array definition within a method/function (we do want the array to reside in code/flash space and don't want the complier to store it in RAM). 
  - One way around this, is to define the `static constexpr const` array as a class member variable and not a method specific variable.

brianljt:
Close this PR as it might cause toolchain dependency issue.
> I think you point out the challenge with this, i.e., the support for this style of array initialization is toolchain dependent and I think we may want to avoid it due to this (to keep OT toolchain independent as much as possible, specially considering users may use custom (chip specific) toolchain to build OT stack).

PR #6243, opened 2021-03-07
[api] remove platform UART API

bukepo:
This commit removes application library CLI/NCP dependency on platform
layer APIs. Instead, application developer provides callbacks sending
CLI/NCP data.

With this change, platforms have native support for formatted output can
simply implement the CLI output callback with something like `vprintf()`.

Depends-On: openthread/ot-br-posix#722
Depends-On: openthread/ot-nrf528xx#6


include/openthread/cli.h
```diff
@@ -77,33 +77,25 @@ typedef struct otCliCommand
  * @returns                Number of bytes processed by the callback.
  *
```
abtink, line 78: Update the documentation to mirror the change in parameters?

script/check-size
```diff
@@ -65,6 +65,13 @@ markdown_init()
     echo '| :----: | :------: | -----: | ----: | ---: | ----: |'
 }
 
+nm_size()
+{
+    arm-none-eabi-nm --print-size --size-sort -C "$1" > nmsize_old
+    arm-none-eabi-nm --print-size --size-sort -C "$2" > nmsize_new
+    diff -Nuar nmsize_old nmsize_new
+}
+
```
abtink, line 74: This is not directly related to rest of changes, I am just curious. (I am fine with it being added in this PR).

src/cli/cli.cpp
```diff
@@ -2498,7 +2500,7 @@ otError Interpreter::ProcessPskc(uint8_t aArgsLength, char *aArgs[])
         const otPskc *pskc = otThreadGetPskc(mInstance);
 
         OutputBytes(pskc->m8);
-        OutputLine("");
+        OutputFormat("\r\n");
```
abtink, line 2503: What is the reason for replacing the `OutputLine("")` with `OutputFormat("\r\n")`?
`OutputLine()` already should add the `\r\n`.

src/cli/cli.cpp
```diff
@@ -5161,24 +5163,33 @@ void Interpreter::OutputLine(uint8_t aIndentSize, const char *aFormat, ...)
...
 
-    while (aCount > 0)
-    {
-        uint8_t len = OT_MIN(aCount, sizeof(kSpaces));
+    sprintf(format, "%%%us", aCount);
 
-        Output(kSpaces, len);
-        aCount -= len;
-    }
+    OutputFormat(format, "");
```
abtink, line 5170: Nice! 👍

abtink (approved):
Thanks @bukepo . Overall looks really clean and good. 👍 

I have couple of smaller suggestions and questions. But LGTM already.

script/check-size
bukepo: I added this to check what caused the code size increase. Still working on it.

bukepo: The size increase is mainly caused by the newly added `CliUartOutput`(in examples/apps/cli/cli_uart.cpp) which includes an enhancement for output long string. It eliminate a big buffer in the stack for formatting output string by reusing the whole Uart tx buffer for this purpose. The cost is text size increase. I think it acceptable.

examples/platforms/cc2650/uart.c
```diff
@@ -34,7 +34,7 @@
 #include <driverlib/uart.h>
 
 #include <utils/code_utils.h>
-#include <openthread/platform/uart.h>
+#include <utils/uart.h>
```
zhanglongxia, line 37: nit: `#include "utils/uart.h"`

script/check-size
bukepo: removed from this PR.

zhanglongxia (approved):
LGTM

jwhui (approved):

romacdon:
@bukepo this change broke the Silabs builds.  We are now getting the following error.  Can you help me understand the recommended fix?  I will make the change.

/home/buildengineer/jenkins-slave/workspace/openthread_wrapper/openthread/examples/../examples/platforms/efr32/sleepy-demo/sleepy-demo-mtd/main.c: In function "main":
/home/buildengineer/jenkins-slave/workspace/openthread_wrapper/openthread/examples/../examples/platforms/efr32/sleepy-demo/sleepy-demo-mtd/main.c:100:5: error: implicit declaration of function 'otCliUartInit'; did you mean "otCliInit"? [-Werror=implicit-function-declaration]
     otCliUartInit(instance);
     ^~~~~~~~~~~~~
     otCliInit

jwhui:
@romacdon , I think you can mirror the same changes as in:
- `examples/apps/cli/Makefile.am`
- `examples/apps/cli/ftd.cmake`
- `examples/apps/clii/main.c`

Note that  `examples/apps/cli/cli_uart.cpp` source needs to be pulled in as well.

/cc/ @bukepo

romacdon:
Thanks @jwhui -  @suveshpratapa fyi

doru91:
> @romacdon , I think you can mirror the same changes as in:
> 
> * `examples/apps/cli/Makefile.am`
> * `examples/apps/cli/ftd.cmake`
> * `examples/apps/clii/main.c`
> 
> Note that `examples/apps/cli/cli_uart.cpp` source needs to be pulled in as well.
> 
> /cc/ @bukepo

Sorry for posting in a closed issue. 

It's still unclear to me how to proceed in case in case we need cli functionality in an application. Basically, we are building demo applications above OpenThread using OT libs and interface files. 

Now, for CLI functionality, my understanding is that the functions from examples/apps/cli/cli_uart.cpp needs to be exposed to the demo applications. Is it expected that we need to also compile cli_uart.cpp in our demo applications? If yes, then we also need to include some include files which are not exposed by OT interface and I assume this is not desired.

@jwhui @marius-preda

bukepo:
After this commit, the CLI becomes an application module which is not tied to a specific type of underlying transport. So it's not no longer necessary to implement APIs defined in examples/platforms/utils/uart.h.

examples/apps/cli/cli_uart.cpp is just for backward compatibility. You may implement your own CLI handler based on the transport your platform has. To work with the new API, a vprintf-like callback is provided to `otCliInit()` for writing CLI outputs(see a simple implementation https://github.com/openthread/openthread/blob/main/src/posix/cli.cpp#L91). And whenever there is a command line received, the application should feed it to CLI module with `otCliInputLine()`(see https://github.com/openthread/openthread/blob/main/src/posix/cli.cpp#L146).

Existing implementations for reference:

* src/posix/cli.cpp
* src/posix/platform/daemon.cpp

I hope it helps.

doru91:
Thanks a lot, @bukepo . It does help.

PR #6246, opened 2021-03-08
[tests] test updating SRP host address

wgtdkp:
Test if we can update the host address.


abtink (approved):
Thanks. LGTM. 👍

simonlingoogle (approved):
LGTM :+1:

librasungirl (commented):
For completeness, should we also verify that advertising proxy could respond the right/updated address when the host on the infrastructure link want to query and talk with the service provider?

wgtdkp:
> For completeness, should we also verify that advertising proxy could respond the right/updated address when the host on the infrastructure link want to query and talk with the service provider?

@librasungirl  Thanks! Makes sense to me, will add the same case for Advertising Proxy soon.

librasungirl (approved):

jwhui (approved):

PR #6248, opened 2021-03-08
[srp-server] notify advertising proxy of host/service expiration

wgtdkp:
There is an issue that the SRP host or service are not deregistered by the `Advertising Proxy`
when the host or service is expired. This is because we didn't notifies the `Advertising Proxy` of
 the host/service expiration. This PR fixes this issue and adds tests for it.

Depends-On: openthread/ot-br-posix#731

### DIFF
--- a/src/core/net/srp_server.cpp
+++ b/src/core/net/srp_server.cpp
@@ -132,7 +132,7 @@ otError Server::SetLeaseRange(uint32_t aMinLease, uint32_t aMaxLease, uint32_t a
 
     // TODO: Support longer LEASE.
     // We use milliseconds timer for LEASE & KEY-LEASE, this is to avoid overflow.
-    VerifyOrExit(aMaxKeyLease <= TimerMilli::kMaxDelay / 1000, error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(aMaxKeyLease <= Time::MsecToSec(TimerMilli::kMaxDelay), error = OT_ERROR_INVALID_ARGS);
     VerifyOrExit(aMinLease <= aMaxLease, error = OT_ERROR_INVALID_ARGS);
     VerifyOrExit(aMinKeyLease <= aMaxKeyLease, error = OT_ERROR_INVALID_ARGS);
     VerifyOrExit(aMinLease <= aMinKeyLease, error = OT_ERROR_INVALID_ARGS);
@@ -214,11 +214,41 @@ void Server::AddHost(Host *aHost)
     IgnoreError(mHosts.Add(*aHost));
 }
 
-void Server::RemoveAndFreeHost(Host *aHost)
+void Server::RemoveHost(Host *aHost, bool aRetainName, bool aNotifyServiceHandler)
 {
-    otLogInfoSrp("[server] fully remove host %s", aHost->GetFullName());
-    IgnoreError(mHosts.Remove(*aHost));
-    aHost->Free();
+    VerifyOrExit(aHost != nullptr);
+
+    aHost->mLease = 0;
+    aHost->ClearResources();
+
+    if (aRetainName)
+    {
+        otLogInfoSrp("[server] remove host '%s' (but retain its name)", aHost->mFullName);
+    }
+    else
+    {
+        aHost->mKeyLease = 0;
+        IgnoreError(mHosts.Remove(*aHost));
+        otLogInfoSrp("[server] fully remove host '%s'", aHost->mFullName);
+    }
+
+    if (aNotifyServiceHandler && mServiceUpdateHandler != nullptr)
+    {
+        mServiceUpdateHandler(aHost, kDefaultEventsHandlerTimeout, mServiceUpdateHandlerContext);
+        // We don't wait for the reply from the service update handler,
+        // but always remove the host (and its services) regardless of
+        // host/service update result. Because removing a host should fail
+        // only when there is system failure of the platform mDNS implementation
+        // and in which case the host is not expected to be still registered.
+    }
+
+    if (!aRetainName)
+    {
+        aHost->Free();
+    }
+
+exit:
+    return;
 }
 
 const Server::Service *Server::FindService(const char *aFullName) const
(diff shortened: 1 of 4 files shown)


simonlingoogle (approved):
LGTM :+1: 

with nits

src/core/net/srp_server.cpp
```diff
@@ -1526,20 +1543,54 @@ Server::Service *Server::Host::AddService(const char *aFullName)
     return service;
 }
 
-void Server::Host::RemoveAndFreeService(Service *aService)
+void Server::Host::RemoveService(Service *aService, bool aRetainName, bool aNotifyServiceHandler)
 {
-    if (aService != nullptr)
+    VerifyOrExit(aService != nullptr);
```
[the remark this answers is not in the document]
wgtdkp: There are cases the `aService` could be nullptr. `VerifyOrExit` is to avoid checking `aService` everytime before calling this method.

src/core/net/srp_server.cpp
```diff
@@ -1526,20 +1545,56 @@ Server::Service *Server::Host::AddService(const char *aFullName)
...
+        otLogInfoSrp("[server] remove service '%s' (but retain its name)", aService->mFullName);
+    }
+    else
+    {
+        otLogInfoSrp("[server] fully remove service '%s'", aService->mFullName);
+    }
+
+    if (aNotifyServiceHandler && server.mServiceUpdateHandler != nullptr)
+    {
+        LinkedList<Service> leftServices = mServices;
```
jwhui, line 1568: nit: rename `leftServices` to `remainingServices`?

src/core/net/srp_server.cpp
```diff
@@ -214,11 +214,41 @@ void Server::AddHost(Host *aHost)
...
+    else
+    {
+        aHost->mKeyLease = 0;
+        IgnoreError(mHosts.Remove(*aHost));
+        otLogInfoSrp("[server] fully remove host '%s'", aHost->mFullName);
+    }
+
+    if (aNotifyServiceHandler && mServiceUpdateHandler != nullptr)
+    {
+        mServiceUpdateHandler(aHost, kDefaultEventsHandlerTimeout, mServiceUpdateHandlerContext);
```
jwhui, line 237: Is it safe for the SRP Server observer to call `otSrpServerHandleServiceUpdateResult()` in the handler? Is possible to run into a "double free" issue on `aHost` below?

src/core/net/srp_server.cpp
```diff
@@ -214,11 +214,41 @@ void Server::AddHost(Host *aHost)
     IgnoreError(mHosts.Add(*aHost));
 }
 
-void Server::RemoveAndFreeHost(Host *aHost)
+void Server::RemoveHost(Host *aHost, bool aRetainName, bool aNotifyServiceHandler)
 {
-    otLogInfoSrp("[server] fully remove host %s", aHost->GetFullName());
-    IgnoreError(mHosts.Remove(*aHost));
-    aHost->Free();
+    VerifyOrExit(aHost != nullptr);
```
jwhui, line 219: Under what scenarios is `aHost` equal to `nullptr`?

src/core/net/srp_server.cpp
abtink: I think @wgtdkp knows this better and can confim this
But I guess/think it should be safe since in `UpdateResult ` we seem to check in `mOutstandingUpdates` list and here we are not adding anything to `mOutstandingUpdates` list. So the `UpdatResult` (if called) should do nothing.

abtink (approved):
LGTM. Thanks @wgtdkp. 👍 

Just couple of nit suggestions.

src/core/net/srp_server.cpp
wgtdkp: Thanks @abtink, yes it is safe for the observer to call `otSrpServerHandleServiceUpdateResult` and it will be ignored.
We have the invariant that a host is either in the `OutstandingUpdates` or in the `mHosts` list.

But I am feeling that using the `Host` pointer to identify a SRP update transaction is not safe. For, example, a new Host object may be allocated with the same address of an already freed host object and a callback for the freed host will match the newly allocated host object. 

I have been planning to use an `uint32_t` SRP update/transaction ID that can uniquely identify a SRP update but did't get a chance yet. I will have the change ASAP.

src/core/net/srp_server.cpp
wgtdkp: The`RemoveService` allows `aService` to be nullptr and I think it would be expected that `RemoveHost` should also accept nullptr. I updated the usage of `RemoveHost` to avoid checking `aHost` before calling it.

wgtdkp:
@jwhui  I am marking this WIP because the test script need to be updated.

wgtdkp:
@jwhui  I am marking this WIP because the test script need to be updated.

src/core/net/srp_server.cpp
```diff
@@ -1526,20 +1541,56 @@ Server::Service *Server::Host::AddService(const char *aFullName)
...
+    {
+        otLogInfoSrp("[server] fully remove service '%s'", aService->mFullName);
+    }
+
+    if (aNotifyServiceHandler && server.mServiceUpdateHandler != nullptr)
+    {
+        LinkedList<Service> remainingServices = mServices;
+
+        mServices.Clear();
+        IgnoreError(mServices.Add(*aService));
```
abtink, line 1567: @wgtdkp, I just noticed this. I think this may cause some issue since `aService`  (which is already in the `remainingServices` linked list) is being placed into a new list and this can mess up the original `remainingServcies` list. 

One solution can be that we remove it from `remainingServices` before adding it here to `mServcies` and then re-add it to `reminingServices`.

src/core/net/srp_server.cpp
abtink: > I have been planning to use an uint32_t SRP update/transaction ID that can uniquely identify a SRP update. Thoughts?

This sounds like a very good idea to me. 👍

src/core/net/srp_server.cpp
wgtdkp: Good catch! I revised accordingly, please re-review.

wgtdkp:
@jwhui this case is now fixed together with https://github.com/openthread/ot-br-posix/pull/731.

jwhui (approved):

PR #6255, opened 2021-03-09
[fuzz] disable exit command on fuzz target

bukepo:
fuzz target must not exit() or abort() on any input.

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -1662,7 +1662,7 @@ exit:
     return error;
 }
 
-#if OPENTHREAD_POSIX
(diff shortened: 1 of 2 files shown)


bukepo:
I found the fuzz test failed once for the input 'exit'

jwhui:
> I found the fuzz test failed once for the input 'exit'

I'm not opposed to this change, but we typically do not enable `OPENTHREAD_POSIX` in fuzz builds - see [tests/fuzz/oss-fuzz-build](https://github.com/openthread/openthread/blob/main/tests/fuzz/oss-fuzz-build). Did something change?

bukepo:
superseded by #6259

PR #6257, opened 2021-03-09
[srp-client] adding buffer and service pool module

abtink:
This commit adds a new module `Utils::SrpClientBuffers` which provides
string/data buffers and a pool of service entries for use for SRP
client. OpenThread SRP client implementation requires that the
service, string/data buffers passed as parameters to its API to
persist. This allows flexibility in how the SRP client can be used
(users of API can decide on how to manage the memory). The new module
provides an option to OT API user to use the buffer and service pool
to allocate/free items (e.g., service entries). The new module is
enabled by `OPENTHREAD_CONFIG_SRP_CLIENT_BUFFERS_ENABLE`. A set of new
configs specifies the number of service entries in the pool, number of
host address entries, and the size of different string buffers (e.g,
size of service instance string, etc).

This commit also updates CLI and NCP implementations to use the new
common service pool for SRP client commands.

### DIFF
--- a/include/openthread/srp_client_buffers.h
+++ b/include/openthread/srp_client_buffers.h
@@ -1,0 +1,179 @@
+/*
+ *  Copyright (c) 2021, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ * @brief
+ *  This file defines the OpenThread SRP (Service Registration Protocol) client buffers and service pool.
+ */
+
+#ifndef OPENTHREAD_SRP_CLIENT_BUFFERS_H_
+#define OPENTHREAD_SRP_CLIENT_BUFFERS_H_
+
+#include <openthread/dns.h>
+#include <openthread/srp_client.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @addtogroup api-srp
+ *
+ * @brief
+ *   This module includes functions for SRP client buffers and service pool.
+ *
+ * @{
+ *
+ * Functions in this module are only available when feature OPENTHREAD_CONFIG_SRP_CLIENT_BUFFERS_ENABLE is enabled.
+ *
+ */
+
+/**
+ * This struct represents a SRP client service pool entry.
+ *
+ */
+typedef struct otSrpClientBuffersServiceEntry
+{
+    otSrpClientService mService;  ///< The SRP client service structure.
+    otDnsTxtEntry      mTxtEntry; ///< The SRP client TXT entry.
+} otSrpClientBuffersServiceEntry;
+
+/**
+ * This function gets the string buffer to use for SRP client host name.
+ *
+ * @param[in]  aInstance  A pointer to the OpenThread instance.
+ * @param[out] aSize      Pointer to a variable to return the size (number of bytes) of the string buffer (MUST NOT be
+ *                        NULL).
+ *
+ * @returns A pointer to char buffer to use for SRP client host name.
+ *
+ */
+char *otSrpClientBuffersGetHostNameString(otInstance *aInstance, uint16_t *aSize);
+
+/**
+ * This function gets the array of IPv6 address entries to use as SRP client host address list.
+ *
+ * @param[in]  aInstance     A pointer to the OpenThread instance.
+ * @param[out] aArrayLength  Pointer to a variable to return the array length i.e., number of IPv6 address entries in
+ *                           the array (MUST NOT be NULL).
+ *
+ * @returns A pointer to an array of `otIp6Address` entries (number of entries is returned in @p aArrayLength).
+ *
+ */
+otIp6Address *otSrpClientBuffersGetHostAddressesArray(otInstance *aInstance, uint8_t *aArrayLength);
+
+/**
+ * This function allocates a new service entry from the pool.
+ *
+ * The returned service entry instance will be initialized as follows:
+ *
+ *  - `mService.mName` will point to an allocated string buffer which can be retrieved using the function
+ *    `otSrpClientBuffersGetServiceEntryServiceNameString()`.
+ *  - `mService.mInstanceName` will point to an allocated string buffer which can be retrieved using the function
+ *    `otSrpClientBuffersGetServiceEntryInstanceNameString()`.
+ *  - `mService.mTxtEntries` will point to `mTxtEntry`.
+ *  - `mService.mNumTxtEntries` will be set to one.
+ *  - Other `mService` fields (port, priority, weight) are set to zero.
+ *  - `mTxtEntry.mKey` is set to NULL (value is treated as already encoded).
+ *  - `mTxtEntry.mValue` will point to an allocated buffer which can be retrieved using the function
+ *    `otSrpClientBuffersGetServiceEntryTxtBuffer()`.
+ *  - `mTxtEntry.mValueLength` is set to zero.
+ *
+ * @param[in] aInstance   A pointer to the OpenThread instance.
+ *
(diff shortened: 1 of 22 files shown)


src/cli/cli_srp_client.cpp
```diff
@@ -190,9 +183,26 @@ otError SrpClient::ProcessHost(uint8_t aArgsLength, char *aArgs[])
...
+            hostName = otSrpClientPoolGetHostNameString(mInterpreter.mInstance, &size);
+
+            len = strlen(aArgs[1]);
+            VerifyOrExit(len + 1 <= size, error = OT_ERROR_INVALID_ARGS);
+
+            // We first make sure we can set the name, and if so
+            // we copy it to the persisted string buffer and set
+            // the host name again now with the persisted buffer.
+            // This ensures that we do not overwrite a previous
+            // buffer when a host name that cannot be set.
```
jwhui, line 200:
```suggestion
            // buffer with a host name that cannot be set.
```

src/cli/cli_srp_client.cpp
```diff
@@ -215,24 +225,30 @@ otError SrpClient::ProcessHost(uint8_t aArgsLength, char *aArgs[])
...
-                SuccessOrExit(error = ParseAsIp6Address(aArgs[0], addresses[numAddresses++]));
+                SuccessOrExit(error = ParseAsIp6Address(aArgs[index], addresses[index - 1]));
             }
 
-            VerifyOrExit(numAddresses > 0, error = OT_ERROR_INVALID_ARGS);
-            memcpy(mHostAddresses, addresses, sizeof(addresses));
+            SuccessOrExit(error = otSrpClientSetHostAddresses(mInterpreter.mInstance, addresses, numAddresses));
 
-            error = otSrpClientSetHostAddresses(mInterpreter.mInstance, mHostAddresses, numAddresses);
+            memcpy(hostAddressArray, addresses, numAddresses * sizeof(otIp6Address));
```
jwhui, line 250:
```suggestion
            memcpy(hostAddressArray, addresses, numAddresses * sizeof(hostAddressArray[0]));
```

src/cli/cli_srp_client.cpp
```diff
@@ -349,89 +361,94 @@ otError SrpClient::ProcessService(uint8_t aArgsLength, char *aArgs[])
...
-        {
-            SuccessOrExit(error = ParseAsUint16(aArgs[4], entry->mService.mPriority));
-        }
+    VerifyOrExit(4 <= aArgsLength && aArgsLength <= 7, error = OT_ERROR_INVALID_ARGS);
 
-        if (aArgsLength >= 6)
-        {
-            SuccessOrExit(error = ParseAsUint16(aArgs[5], entry->mService.mWeight));
-        }
+    entry = otSrpClientPoolAllocateService(mInterpreter.mInstance);
```
jwhui, line 406: Do we allow duplicate entries? Checking would add additional code, so I'm not sure duplicate checking is worth the tradeoff.

src/core/api/srp_client_service_pool_api.cpp
```diff
@@ -0,0 +1,97 @@
...
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file implements the OpenThread SRP client APIs.
```
jwhui, line 31:
```suggestion
 *   This file implements the OpenThread SRP client service pool APIs.
```

src/core/config/srp_client.h
```diff
@@ -218,4 +218,72 @@
...
+ *
+ */
+#define OPENTHREAD_CONFIG_SRP_CLIENT_SERVICE_POOL_ENABLE OPENTHREAD_CONFIG_SRP_CLIENT_ENABLE
+
+/**
+ * @def OPENTHREAD_CONFIG_SRP_CLIENT_SERVICE_POOL_MAX_SERVICES
+ *
+ * Specifies number of service entries in the SRP client service pool.
+ *
+ * This config is applicable/used only when `OPENTHREAD_CONFIG_SRP_CLIENT_SERVICE_POOL_ENABLE` is enabled.
```
jwhui, line 234: nit: reduce redundant words (here and below)
```suggestion
 * This config is applicable only when `OPENTHREAD_CONFIG_SRP_CLIENT_SERVICE_POOL_ENABLE` is enabled.
```

include/openthread/srp_client_service_pool.h
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * This function gets the string buffer for service name from a service entry.
+ *
+ * @param[in]  aEntry   A pointer to a previously allocated service entry (MUST NOT be NULL).
+ * @param[out] aSize    If not NULL, returns the size (number of bytes) of the string buffer.
+ *
+ * @returns A pointer to the string buffer.
+ *
+ */
+char *otSrpClientPoolGetServiceEntryServiceNameString(otSrpClientPoolServiceEntry *aEntry, uint16_t *aSize);
```
jwhui, line 142: Why do we need an API to return the string buffer? Can't the client access the pointer directly in the struct? I understand the need to query the size.

src/core/utils/srp_client_service_pool.hpp
```diff
@@ -0,0 +1,228 @@
...
+         *
+         * @returns A pointer to the string buffer.
+         *
+         */
+        char *GetInstanceNameString(uint16_t *aSize);
+
+        /**
+         * This method gets the  buffer for the TXT value from the service entry.
+         *
+         * @param[out] aSize    If not NULL, returns the size (number of bytes) of the  buffer.
```
jwhui, line 135:
```suggestion
         * @param[out] aSize    If not NULL, returns the size (number of bytes) of the buffer.
```

include/openthread/srp_client_service_pool.h
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * This function gets the string buffer to use for SRP client host name.
+ *
+ * @param[in] aInstance   A pointer to the OpenThread instance.
+ * @param[out] aSize      If not NULL, returns the size (number of bytes) of the string buffer.
+ *
+ * @returns A pointer to char buffer to use for SRP client host name.
+ *
+ */
+char *otSrpClientPoolGetHostNameString(otInstance *aInstance, uint16_t *aSize);
```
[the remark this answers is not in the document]
abtink: The pool is really only for the service entries since we can have multiple services and add and remove them dynamically.

The host name and host address array mirror the SRP client APIs, where they are really single instances (e.g. if we us the API to set the host name, it is actually replacing any previous string).

include/openthread/srp_client_service_pool.h
abtink: Yes, the caller can access the pointer to the `mName` and `mInstanceName` in `otSrpClientService` struct, but the challenge is that the they are defined as `const char *`, so the caller technically should not change the string buffer content. 

We can document it that caller needs to do `const_cast` but since we need a function to query the size of the buffer, I opted to return the pointer to the string buffer (now as `char *`) as well.

src/cli/cli_srp_client.cpp
abtink: I agree that checking for duplicate entries may not be worth the code overhead. 

I think the SRP server would reject duplicate service entries (if we try to register same service again) so I expect that  we get an error from server (in the callback) if duplicate entries are added.

include/openthread/srp_client_service_pool.h
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * This function gets the array of IPv6 address entries to use as SRP client host address list.
+ *
+ * @param[in] aInstance      A pointer to the OpenThread instance.
+ * @param[out] aArrayLength  If not NULL, returns the array length (number of IPv6 address entries in the array).
```
wgtdkp, line 82: I think `aArrayLength` MUST not be `NULL`. Are there chances that the user can/should use the returned address buffer without knowing its length?

src/core/utils/srp_client_service_pool.cpp
```diff
@@ -0,0 +1,132 @@
...
+    entry->mService.mNumTxtEntries = 1;
+    entry->mTxtEntry.mValue        = entry->mTxtBuffer;
+
+exit:
+    return entry;
+}
+
+void SrpClientServicePool::FreeService(ServiceEntry &aService)
+{
+    mServicePool.Free(static_cast<ServiceEntry &>(aService));
```
wgtdkp, line 121:
```suggestion
    mServicePool.Free(aService);
```

src/ncp/ncp_base_mtd.cpp
```diff
@@ -63,6 +63,9 @@
 #if (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
 #include "openthread/backbone_router.h"
 #endif
+#if OPENTHREAD_CONFIG_SRP_CLIENT_ENABLE
```
wgtdkp, line 66: nit:
```suggestion
#if OPENTHREAD_CONFIG_SRP_CLIENT_SERVICE_POOL_ENABLE
```

wgtdkp (approved):
LGTM :+1: , just a few nits :)

include/openthread/srp_client_service_pool.h
abtink: We can change this and require a non-`NULL` length variable in all the similar getter functions, but I think it may be useful in some cases to let the caller use `NULL`.

The caller may aleady know the length of the array and/or string buffers from having access to the project specific config that it defined when building openthread library (i.e. basically from knowing the value of the `OPENTHREAD_CONFIG_SRP_CLIENT_SERVICE_POOL_MAX_HOST_ADDRSSES` config, etc.).  This situation can happen when OT is integrated into a product where the modules using OT (and its API) have access to the config header file that was provided to OT (e.g., this can happen in SoC style product integration).

wgtdkp: For cases that the user has access to `OPENTHREAD_CONFIG_SRP_CLIENT_SERVICE_POOL_MAX_HOST_ADDRSSES`, I feel it still not a good practice to use the configuration directly but not get the length along with the returned pointer.

Besides, disallow `NULL` should be able to simplifies the implementation a bit :)

abtink: I agree it makes the code simpler. I changed this in the latest pushed commit.

include/openthread/srp_client_service_pool.h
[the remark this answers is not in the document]
jwhui: Should we rename to something like: `otSrpClientBufs*`?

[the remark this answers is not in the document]
abtink: Renamed to use `otSrpClientBuffers*` in the last pushed commit.

include/openthread/srp_client_buffers.h
```diff
@@ -0,0 +1,179 @@
...
+{
+    otSrpClientService mService;  ///< The SRP client service structure.
+    otDnsTxtEntry      mTxtEntry; ///< The SRP client TXT entry.
+} otSrpClientBuffersServiceEntry;
+
+/**
+ * This function gets the string buffer to use for SRP client host name.
+ *
+ * @param[in] aInstance   A pointer to the OpenThread instance.
+ * @param[out] aSize      Pointer to a variable to return the size (number of bytes) of the string buffer (MUST NOT be
```
jwhui, line 71: nit: alignment (here and below)

jwhui (commented):
Also need to increment `OPENTHREAD_API_VERSION`.

jwhui (approved):

PR #6258, opened 2021-03-09
[spinel] add support for spinel interface id

mnp222:
This feature allows the RCP to support multiple host stacks
on different PANs by making use of the spinel Interface ID.


src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1835,21 +1854,21 @@ void RadioSpinel<InterfaceType, ProcessContextType>::HandleTransmitDone(uint32_t
...
 
-    unpacked = spinel_datatype_unpack(aBuffer, aLength, SPINEL_DATATYPE_BOOL_S, &framePending);
-    VerifyOrExit(unpacked > 0, error = OT_ERROR_PARSE);
-
-    aBuffer += unpacked;
-    aLength -= static_cast<uint16_t>(unpacked);
-
     if (status == SPINEL_STATUS_OK)
     {
+        unpacked = spinel_datatype_unpack(aBuffer, aLength, SPINEL_DATATYPE_BOOL_S, &framePending);
```
mnp222, line 1859: This change fixes a bug where a spinel error message gets masked by OT_ERROR_PARSE message because the error message can't be further parsed.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -2208,9 +2227,12 @@ void RadioSpinel<InterfaceType, ProcessContextType>::RecoverFromRcpFailure(void)
     if (mResetRadioOnStartup)
     {
         SuccessOrDie(SendReset());
+        SuccessOrDie(WaitResponse());
+    }
+    else
+    {
+        mIsReady = true;
     }
```
mnp222, line 2235: The above change allows recovery without resetting the RCP, which is necessary in the case of multiple host processes.

src/ncp/ncp_base.hpp
```diff
@@ -559,7 +568,11 @@ class NcpBase
 
     uint8_t mTxBuffer[kTxBufferSize];
 
+#if OPENTHREAD_CONFIG_MULTIPAN_RCP_ENABLE
+    spinel_tid_t mNextExpectedTid[4];
+#else
     spinel_tid_t mNextExpectedTid;
+#endif
```
mnp222, line 576: Should we get rid of the switch here and just always use the array?

The value 4 should be replaced by (SPINEL_HEADER_IID_MAX + 1) defined in spinel.h.  I will fix that.

src/ncp/ncp_base_radio.cpp
```diff
@@ -451,11 +449,16 @@ otError NcpBase::HandlePropertySet_SPINEL_PROP_STREAM_RAW(uint8_t aHeader)
...
     }
     else
     {
-        error = WriteLastStatusFrame(aHeader, ThreadErrorToSpinelStatus(error));
+        // If we fail to report the error now, it will be reported later in HandleReceive() of ncp_base.cpp.
+        if (WriteLastStatusFrame(aHeader, ThreadErrorToSpinelStatus(error)) == OT_ERROR_NONE)
+        {
+            return OT_ERROR_NONE;
+        }
     }
```
mnp222, line 462: The above changes starting at line 443 are a bug we came across with how errors were being reported up.

src/posix/main.c
```diff
@@ -261,10 +261,11 @@ static otInstance *InitInstance(PosixConfig *aConfig)
 
     syslog(LOG_INFO, "Running %s", otGetVersionString());
     syslog(LOG_INFO, "Thread version: %hu", otThreadGetVersion());
-    IgnoreError(otLoggingSetLevel(aConfig->mLogLevel));
 
     instance = otSysInit(&aConfig->mPlatformConfig);
 
+    IgnoreError(otLoggingSetLevel(aConfig->mLogLevel));
+
     atexit(otSysDeinit);
```
mnp222, line 269: This fixes a bug where otSysInit() sets the log level to the compile time config, rendering the command line argument useless.

src/core/mac/sub_mac.cpp
```diff
@@ -710,7 +718,12 @@ bool SubMac::ShouldHandleCsmaBackOff(void) const
 {
     bool swCsma = true;
 
+#if OPENTHREAD_CONFIG_MULTIPAN_RCP_ENABLE
+    // Use software csma to perform retries in case of a busy RCP
+    goto exit;
+#else
     VerifyOrExit(!RadioSupportsCsmaBackoff(), swCsma = false);
+#endif
```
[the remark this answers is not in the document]
mnp222: Hi Abtin, I agree, I did not like those changes to the OT core modules. This change is for the case where one process wants to tx while the other one is already transmitting.  I can place logic to handle it (basically performing retries) into radio_spinel_impl.hpp.  Does that seem like an ok place for it?

Thanks for the additional discussion, it is very helpful.  In your case were your stack entities running in a single process so that the arbiter module sat under both of them, or were they in separate processes?  If the latter, how did the IPC work?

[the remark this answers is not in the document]
abtink: Thanks @mspang.

I agree that handling it in `radio_spinel_imp.hpp` sounds reasonable. I would say if we can move it to RCP side itself, it may be even better. I think RCP  is the place where we need to be aware of multiple PANs/entities and it can best handle any conflicting requests.

I think from the host side of `radio_spinel_impl`, we may only need to know to use certain spinel IID to talk to the RCP (would it matter to know if there are other entities that are controlling the RCP or not on the host side?).

----------

Let me try add more details (from what I recall) on our internal project and how it worked:

- It was all running on an RTOS platform. All modules were part of the firmware running on NPC (OT stack, radio arbiter and driver, and other (legacy) stacks) all part of NCP firmware. 
- OT stack had its own RTOS task:
   - OT stack requires that all the OT public and platform APIs and callbacks to be invoked from same OS context, e.g. same RTOS task.
- Arbiter and radio driver had their own shared RTOS task (and I think other/legacy stack ran on its own RTOS task as well).
- I think the different tasks synced using ROTS locks/semaphore.

Arbiter module provided APIs to all entities which wanted to control the radio:
- For example, `otPlatRadio` APIs to OT stack and very similar set of APIs to the other (legacy) stack.
- Arbiter module managed state variables per entity to remember the last request/command (API call). 
   - For example, if `otPlatRadioSleep()` was called to put radio to sleep, or `otPlatRadioReceive()` to rx, or a request to tx frame or perform energy scan, etc.
- It then handled the requests/commands from different entities and decided what to do actually with the radio. Here are some of the rules I recall:
   - We put radio to sleep mode only when all entities requested it to sleep.
   - When different entities requested radio to be in rx mode: 
        - If the channels matched, it was straight-forward and we put the radio in rx mode on that channel.
        - But if the requests were for different channels, I think we had a simple rule to prioritize one entity (I think OT stack request won on the rx channel).
        - Just a note on this: 
            - In our use-case, all stacks (OT and legacy) were actually operating on the same channel, but we still could have the situation that they put radio on different channels, for example to support a discover/mac scan operation (during which we can switch to different channels).
            - We also had the idea to use a windowed rx model when we have conflicting rx channels (rx on first channel for some time interval, then switch to other channel, an so on) but I think we did not implement/use this model.
   - If we get a tx request by an entity while radio was busy with an ongoing tx from another entity we just waited for the current tx to complete (success or failure) and then proceeded with the next tx request.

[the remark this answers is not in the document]
mnp222: Hi @abtink, thanks for this discussion.  I would like to move the arbitration of transmissions and energy scans to the RCP side as you suggest.  I can add a queue to NcpBase which queues up transmission and energy scan commands if the radio is busy, then proceeds with them when the radio is free. That code would only be enabled in MULTIPAN_RCP mode.  Let me know if that sounds ok.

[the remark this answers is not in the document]
abtink: Thanks @mnp222 . Yes, I think it makes sense and sounds good to me. 👍 

I think we can also consider adding this in the `LinkRaw` class (under `OPENTHREAD_RADIO`) since `LinkRaw` is the point of interaction with radio in an RCP build. `LinkRaw` practically provides a mirror of `radio.hpp` APIs.  

I think placing the multi-pan (radio busy) checks in either `Ncp` or `LinkRaw` would be fine. Thanks.

mnp222:
HI @abtink.  I added a new api to link_raw.h, otLinkRawIsTransmittingOrScanning(), so that ncp_base_radio.cpp could check if the radio is transmitting or scanning before submitting a transmission or energy scan.  However, one of the checks (API Version) seems to be unhappy about that.  How do I fix this?
https://github.com/openthread/openthread/pull/6258/checks?check_run_id=2222515229

abtink:
> HI @abtink. I added a new api to link_raw.h, otLinkRawIsTransmittingOrScanning(), so that ncp_base_radio.cpp could check if the radio is transmitting or scanning before submitting a transmission or energy scan. However, one of the checks (API Version) seems to be unhappy about that. How do I fix this?
> https://github.com/openthread/openthread/pull/6258/checks?check_run_id=2222515229

You just need to increment `OPENTHREAD_API_VERSION ` in the `include/openthread/instance.h` in your commit/PR:
https://github.com/openthread/openthread/blob/4fb42e245bfc6f818d3c88918acbcb0d2e5ae15b/include/openthread/instance.h#L48-L56

src/ncp/ncp_base_radio.cpp
```diff
@@ -434,28 +445,38 @@ otError NcpBase::HandlePropertySet_SPINEL_PROP_STREAM_RAW(uint8_t aHeader)
...
     {
-        // Don't do anything here yet. We will complete the transaction when we get a transmit done callback
-    }
-    else
-    {
-        error = WriteLastStatusFrame(aHeader, ThreadErrorToSpinelStatus(error));
+        // If we fail to report the error now, it will be reported later in HandleReceive() of ncp_base.cpp.
+        if (WriteLastStatusFrame(aHeader, ThreadErrorToSpinelStatus(error)) == OT_ERROR_NONE)
+        {
+            return OT_ERROR_NONE;
```
abtink, line 478: This help ensure we still have single `return` in each function/method (which I think is required by our style guide).
```suggestion
            error = OT_ERROR_NONE;
```

abtink (approved):
Thanks @mnp222. Looks great to me. 👍 

I have one question on the change to `aResetRadio`. From my perspective and the code itself it looks fine, but would like @bukepo (and maybe others) to also check that to make sure if it may impact any current use-cases or not.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -2214,12 +2232,16 @@ void RadioSpinel<InterfaceType, ProcessContextType>::RecoverFromRcpFailure(void)
     mIsReady      = false;
     mIsTimeSynced = false;
 
+#if OPENTHREAD_CONFIG_MULTIPAN_RCP_ENABLE
```
jinran-google, line 2235: In the case of failures on multi-PAN RCP, I don't think we can recover transparently. `OPENTHREAD_CONFIG_MULTIPAN_RCP_ENABLE` and `OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT > 0` should be mutually exclusive.

src/ncp/ncp_base.cpp
```diff
@@ -825,6 +839,144 @@ otError NcpBase::SendQueuedResponses(void)
...
+    mCurScanChannel = static_cast<int8_t>(entry->mScanChannel);
+    mCurTransmitIID = entry->mIid;
+
+exit:
+    if (error != OT_ERROR_NONE)
+    {
+        if (mEncoder.BeginFrame(SPINEL_HEADER_FLAG | static_cast<uint8_t>(entry->mIid << SPINEL_HEADER_IID_SHIFT),
+                                SPINEL_CMD_PROP_VALUE_IS, SPINEL_PROP_MAC_SCAN_STATE) == OT_ERROR_NONE)
+        {
+            if (mEncoder.WriteUint8(SPINEL_SCAN_STATE_IDLE) == OT_ERROR_NONE)
```
jinran-google, line 937: If some error occurs in the middle of the construction of this frame, the encoder will be left at an intermediate state and the error will be ignored. Will that cause some problems?

jinran-google (commented):
Besides the comments, I think there should be some simulation test cases covering these scenarios:
- A host transmits while another is scanning, and vice versa
- Multiple hosts listen on different channels and talk to different nodes
- Multiple hosts forming the same network talk to each other

src/ncp/ncp_base.cpp
```diff
@@ -825,6 +839,144 @@ otError NcpBase::SendQueuedResponses(void)
...
+#if OPENTHREAD_RADIO || OPENTHREAD_CONFIG_LINK_RAW_ENABLE
+
+uint8_t NcpBase::GetWrappedPendingCommandQueueIndex(uint8_t aPosition)
+{
+    while (aPosition >= kPendingCommandQueueSize)
+    {
+        aPosition -= kPendingCommandQueueSize;
+    }
+
+    return aPosition;
```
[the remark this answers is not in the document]
mnp222: I like your suggestion.  I was following the precedent set in GetWrappedResponseQueueIndex().  I will change it there too.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -230,13 +232,18 @@ void RadioSpinel<InterfaceType, ProcessContextType>::Init(bool aResetRadio,
...
     {
         SuccessOrExit(error = SendReset());
+        SuccessOrExit(error = WaitResponse());
+        VerifyOrExit(mIsReady, error = OT_ERROR_FAILED);
+    }
+    else
+    {
+        mIsReady = true;
     }
-
```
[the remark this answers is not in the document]
mnp222: This works as long as the recovery mechanism is enabled, and does not send a new reset when it receives an unexpected reset.  I made a change that prevents this, but I see that it is overly general.  I will clean that up.

src/lib/spinel/radio_spinel_impl.hpp
mnp222: I think it's necessary to enable the RCP restoration mechanism because the RCP may be reset by another host process.  The key point is that upon receipt of an unexpected reset, the host should not reset the RCP again during recovery.  I see that the current change I made here is too general because it prevents reset in the case of an RCP timeout as well as an unexpected reset.  I can fix this by changing mRcpFailed from a boolean to an enum. How does that sound?

src/ncp/ncp_base.cpp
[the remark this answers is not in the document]
mnp222: Perhaps OT has a coding standard against using the modulus operator in embedded code, which would explain why the while loop was used instead. @abtink it looks like you wrote this code in the commit below.  Do you recall your thinking?
https://github.com/openthread/openthread/commit/9ce57b239e7460f1ac9ce68b6ab9e1c759817e63

src/ncp/ncp_base.cpp
jinran-google: I think that makes sense, the error is still ignored but it's OK as `LinkRawEnergyScanDone()` is already ignoring it.

BTW, active scans are async too. Do they need to be handled the same way?

src/lib/spinel/radio_spinel_impl.hpp
mnp222: The host applications need to coordinate with each about long-term global properties of the radio such as channel and power and enable/disable.  In addition, a policy system can be established in the platform layer on the RCP, for example, only one of the processes has permission to change these.  In any case, it is out of scope of the core stack.

For per-IID properties such as pan id, long id, short id, and the src match lists, storing these separately for each IID is also accomplished at the platform layer.  When handling a set or get command, the PAL looks up the current IID via otNcpPlatGetCurCommandIid().

jinran-google: Will the property handling be implemented in this PR or in a separate PR? Looks like the multi PAN feature cannot work properly without coordination between hosts. I think it's good to add some integration test cases so we can better understand the use cases.

jinran-google: Will the property handling be implemented in this PR or in a separate PR? Looks like the multi PAN feature cannot work properly without coordination between hosts. I think it's good to add some integration test cases so we can better understand the use cases.

src/ncp/ncp_base.cpp
mnp222: For the RCP, active scans are handled up on the host using the transmit primitives.  I don't think we can handle it the same way on the RCP.

src/lib/spinel/radio_spinel_impl.hpp
mnp222: The property handling is in the platform layer so it will be in a separate PR.  Testing requires the platform changes and a mux to combine the spinel host streams.  We perform this testing, but I will also look into how this could be added to the openthread testing.  Can you point me to some existing test examples for reference?

jinran-google: You can see <https://github.com/openthread/openthread/tree/main/tests/scripts/thread-cert> for virtual time tests and see <https://github.com/openthread/openthread/tree/main/tests/scripts/expect> for real time tests, specifically <https://github.com/openthread/openthread/blob/main/tests/scripts/expect/posix-rcp-restoration.exp>, which contains manual control of RCP and host processes.

src/ncp/ncp_base.cpp
```diff
@@ -708,14 +709,14 @@ void NcpBase::RegisterPeekPokeDelagates(otNcpDelegateAllowPeekPoke aAllowPeekDel
...
 
-uint8_t NcpBase::GetWrappedResponseQueueIndex(uint8_t aPosition)
+uint8_t NcpBase::GetWrappedQueueIndex(uint8_t aPosition, uint8_t aQueueSize)
 {
-    while (aPosition >= kResponseQueueSize)
+    while (aPosition >= aQueueSize)
     {
-        aPosition -= kResponseQueueSize;
+        aPosition -= aQueueSize;
     }
```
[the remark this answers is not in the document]
jinran-google: This function returns `aPosition % aQueueSize`, so removing this loop won't change any behavior. An if isn't needed either.

jwhui:
@mnp222 @romacdon , are you planning to move this PR forward?

romacdon:
> @mnp222 @romacdon , are you planning to move this PR forward?

@jwhui - yes we definitely want to get this pushed in.  There was a request to add some integration tests and we didn't have time to do it when requested.  Would it be possible to push the changes in without those tests?  Also, we know we need to refresh the PR with the latest code since its been almost a year. We can get to that in the next couple of weeks?  Is someone waiting on these changes or are you just trying to close up old PRs?

jwhui:
> @jwhui - yes we definitely want to get this pushed in. There was a request to add some integration tests and we didn't have time to do it when requested. Would it be possible to push the changes in without those tests? Also, we know we need to refresh the PR with the latest code since its been almost a year. We can get to that in the next couple of weeks? Is someone waiting on these changes or are you just trying to close up old PRs?

@romacdon , thanks for the update, I was just trying to clean up old PRs. We prefer to have relevant CI test with along with new functionality.

Hedda:
Any news on this being merged?

romacdon:
> Any news on this being merged?

We do still plan on merging this.  We just need to create some unit tests.  Do you have a specific need for this?

Hedda:
> > Any news on this being merged?
> 
> @Hedda We do still plan on merging this. We just need to create some unit tests. Do you have a specific need for this?

Yes I believe so, and here is the long back-story behind that in which I try to explain the specific current demand for this feature:

[agners](https://github.com/agners) (who is [Nabu Casa](https://www.nabucasa.com) employee Stefan Agner) is a developer of the very popular home automation software [Home Assistant](https://www.home-assistant.io) is currently developing an “**_OpenThread Border Router Add-on_**” and that OTBR add-on should, in theory, hopefully in the future be able to be compatible with all OpenThread Border Router firmware with Spinel interface in RPC mode no matter which underlying host stack and hardware it is running on. As I understand his OTBR host application add-on will be used as a dependency to enable Matter/CHIP over Thread support in [Home Assistant](https://www.home-assistant.io) which he is also currently developing:

https://github.com/home-assistant/addons-development/tree/master/openthread_border_router

I believe this feature will be needed to for an host application with Spinel RCP support to be compatible with multiple host stacks regardless of underlying hardware and version of stack for OpenThread Border Router, right?  

If so then think a lot of us in open-source home automation communities are specifically interested in having one and the same host application being support DIY-users that have bought different OpenThread Border Router hardware adapters and are using firmware versions on them, and initially for the mentioned/linked host application add-on for Home Assistant to support OpenThread Border Router firmware with Spinel interface in RPC mode running on both Silicon Labs EFR32 based adapter hardware and Texas Instruments CC2652/CC1352 based adapter hardware.

https://github.com/home-assistant/addons-development/issues/38

That is, there is a specific interest in first supporting different OTBR Spinel RPC firmware running on Silicon Labs EFR32 based adapters and Texas Instruments CC2652/CC1352 based adapters. The reason for that is that both those are already readily available as USB radio adapters sold to consumers as Zigbee Coordinator dongles (pre-loaded with Zigbee NCP firmware). There TI CC26x2/CC13x2 based USB adapters are currently by far the most popular hardware within the DIY home automation communities because they are less expensive and there are available globally from different manufacturers, (and stock of Silicon Labs EFR32 based USB radio adapters have also suffered much more than Texas Instruments in the current worldwise chip shortage).

https://www.zigbee2mqtt.io/guide/adapters/#based-on-texas-instruments-cc2652-cc1352-chip

So far [agners](https://github.com/agners) has however only worked with Silicon Labs EFR32MG1x/EFR32MG2x based adapters (primarily Silabs EFR32MG21 chip) with OpenThread Spinel RPC firmware for Matter/CHIP over Thread so his host application add-on but I believe that is only because Nabu Casa choose to use a Silabs EFR32MG21 chip based adapter that will ship on the board as part on the official Home Assistant Yellow (formerly Home Assistant Amber) hardware, which [agners](https://github.com/agners) is the lead developer as well:

https://github.com/zigpy/zigpy/discussions/894

https://www.crowdsupply.com/nabu-casa/home-assistant-yellow

https://www.home-assistant.io/blog/2021/09/13/home-assistant-yellow/

Hopefully, this new feature will lead to all Thread USB radio adapter with OTBR Spinel RPC firmware regardless of stack version and chip hardware manufacturers or models will as such in the future be compatible with the upcoming Thread based “Matter” (Project CHIP / Connected Home over IP) devices if used in Home Assistant with their other add-ons for the that is also in development.

https://github.com/home-assistant/addons-development/tree/master/chip_tool

https://github.com/home-assistant/addons-development/tree/master/chip_controller_repl

jwhui:
@Hedda , thanks for providing the additional context.

Spinel is already OpenThread's standard serial protocol for communicating between a host and 802.15.4 SoC (see our [Co-Processor Designs](https://openthread.io/platforms/co-processor) page on openthread.io). Practically all of the [vendors supporting OpenThread](https://openthread.io/platforms) support RCP with spinel.

This PR is proposing to extend spinel to multiplex multiple 802.15.4-based networks simultaneously across a single serial bus. This is for RCPs that can support multiple 802.15.4 PANs simultaneously.

src/ncp/ncp_base.cpp
```diff
@@ -696,19 +709,20 @@ void NcpBase::RegisterPeekPokeDelagates(otNcpDelegateAllowPeekPoke aAllowPeekDel
...
 
-    return aPosition;
+    return aPosition % aQueueSize;
```
abtink, line 719: I guess we don't need the `%`?

src/ncp/ncp_base.cpp
```diff
@@ -2561,6 +2709,11 @@ otError otNcpStreamWrite(int aStreamId, const uint8_t *aDataPtr, int aDataLen)
     return error;
 }
 
+extern "C" spinel_iid_t otNcpPlatGetCurCommandIid()
```
abtink, line 2712:
```suggestion
extern "C" spinel_iid_t otNcpPlatGetCurCommandIid(void)
```

mnp222:
We will open a new PR that is more up to date.

PR #6259, opened 2021-03-10
[posix] detect POSIX in source

bukepo:


simonlingoogle (approved):
LGTM :+1:

bukepo:
retrigger GitHub actions.

bukepo:
it seems something went wrong with github actions. it fails to cancel previous checks.

include/openthread/config.h
```diff
@@ -35,6 +35,14 @@
 #ifndef OPENTHREAD_CONFIG_H_
 #define OPENTHREAD_CONFIG_H_
 
+#ifndef OPENTHREAD_POSIX
+#if defined(__unix__) || defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) || defined(__ANDROID__)
```
rob-the-dude, line 39: I don't know if this is a good idea.  When we run FreeRTOS simulation on Linux or mac OS, will this automatically define OPENTHREAD_POSIX because the compiler recognizes those platforms?  Is the workaround to define OPENTHREAD_POSIX=0 on the command line in that case?

rob-the-dude, line 39: Please add defined(__NetBSD__) as well :-).

bukepo: Good point! Yes, this change intentionally only defines OPENTHREAD_POSIX when its not defined. I'm also considering implement the `exit` command as a platform specified command.

bukepo: @rob-the-dude On FreeRTOS simulation, does it introduce troubles to have the `exit` command which ends the simulation?

rob-the-dude: I'm not sure, but I think having the "exit" command should be fine (I have an "exit" command in my own fork of it, but it's quite diverged).

Also, as it happens, I have been using OPENTHREAD_POSIX for the FreeRTOS simulator, so this won't break that as I feared.

(On the other hand, I've recently started using the "official" FreeRTOS simulator for POSIX, and unfortunately, that breaks on mac OS when using POSIX APIs inside FreeRTOS tasks -- i.e., using "select" to monitor stdio.  But I think I'll be able to fix that on OPENTHREAD_POSIX by running the UART TX/RX as pthreads, and treat their completions as interrupts as far as FreeRTOS is concerned.  It actually ends up making the POSIX code "more like" the MCU versions anyway.)

include/openthread/config.h
```diff
@@ -35,6 +35,15 @@
 #ifndef OPENTHREAD_CONFIG_H_
 #define OPENTHREAD_CONFIG_H_
 
+#ifndef OPENTHREAD_POSIX
```
jwhui, line 38: Do we use this anywhere outside of CLI? Should we constrain this to a CLI-specific header?

jwhui (approved):

PR #6261, opened 2021-03-10
[dtls] fix ECDSA configuration for newer mbed TLS versions

edmont:
When using newer versions of mbed TLS, two issues appear when using `OPENTHREAD_ECDSA`:
- An empty list of configuration hashes is treated as a configuration error: DTLS fails.
https://github.com/ARMmbed/mbedtls/commit/e131bfec29565946cf1354449cc75b0fbe87378e
- Identifiers with double underscore are not used anymore: signature hashes configuration is not included.
https://github.com/ARMmbed/mbedtls/commit/eccd88871767e2fba5f3a079cfdfcb77c376cf20

This commit addresses both issues.


wgtdkp (approved):
LGTM 👍

frkv (approved):
LGTM

jwhui (approved):
Thanks! 👍🏼

PR #6263, opened 2021-03-10
[mac] increase macMaxFrameRetries to 15

jwhui:
As specified by Thread 1.2.1-draft4.

### DIFF
--- a/src/core/config/mac.h
+++ b/src/core/config/mac.h
@@ -66,11 +66,11 @@
  *
  * The default maximum number of retries allowed after a transmission failure for direct transmissions.
  *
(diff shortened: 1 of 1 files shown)


abtink (approved):
LGTM. Thanks. 👍 

Just curios: Is there any change in draft for indirect retries? (can we consider changing that to 2?)

jwhui:
> Just curios: Is there any change in draft for indirect retries? (can we consider changing that to 2?)

No change yet for indirect retries. This change was briefly discussed, but it also assumes that the SED is awake long enough to receive a retransmission.

PR #6268, opened 2021-03-11
[posix] fix socket issues

### LINKED ISSUE #6264: [github-actions] test_plat_udp_accessiblity.py failing
**Describe the bug**

./tests/scripts/thread-cert/border_router/test_plat_udp_accessiblity.py fails intermittently.

**Console/log output**

- [logs_59853.zip](https://github.com/openthread/openthread/files/6118887/logs_59853.zip)
- [thread-border-router-results.zip](https://github.com/openthread/openthread/files/6118888/thread-border-router-results.zip)

simonlingoogle:
This commit fixes #6264 
- [x] Fix Multicast Routing failed to initialize MRT6 after reset
- [x] Fix DHCP6 client failed to stop the trickle timer

I think there sockets in other modules that need to add `CLOSE_EXEC` flag too:
- TREL over UDP6
- ND Proxy

These can be addressed in future PRs.

### DIFF
--- a/src/core/net/dhcp6_client.cpp
+++ b/src/core/net/dhcp6_client.cpp
@@ -178,6 +178,7 @@ exit:
 
 void Client::Stop(void)
 {
+    mTrickleTimer.Stop();
     IgnoreError(mSocket.Close());
 }
 
@@ -223,6 +224,8 @@ bool Client::HandleTrickleTimer(void)
 {
     bool rval = true;
 
+    OT_ASSERT(mSocket.IsBound());
+
     VerifyOrExit(mIdentityAssociationCurrent != nullptr, rval = false);
 
     switch (mIdentityAssociationCurrent->mStatus)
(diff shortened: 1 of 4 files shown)


wgtdkp (approved):
LGTM :+1:

bukepo (approved):

src/core/net/dhcp6_client.cpp
```diff
@@ -178,6 +178,7 @@ void Client::Start(void)
 
 void Client::Stop(void)
 {
+    mTrickleTimer.Stop();
```
simonlingoogle, line 181: It's possible that the DHCP6 client is stopped while it's still soliciting. That is the DHCP6 prefix is removed right after the client solicited.
Currently, the issue is revealed because the BR (which is DHCP6 server & client & owner of the DHCP6 prefix) learns the Network Data containing the DHCP6 prefix from another Thread device after reset, and removes the DHCP6 prefix immediately because of Network Data inconsistency. 
Such issue can also occur by a small chance when a DHCP6 client starts soliciting and DHCP6 server removes the prefix at the same time.

Stopping the trickle timer when stopping DHCP6 client resolves this issue.

jwhui (approved):

PR #6274, opened 2021-03-12
[ip6] add operator `<` overload for 'Ip6::Prefix' comparison

abtink:
This commit adds an overload of operator `<` in `Ip6::Prefix` class. A
prefix with shorter length is considered smaller than the one with
longer length. If the prefix lengths are equal, then the prefix bytes
are compared directly to determine the smaller prefix. This commit
then changes `RoutingManager` to use the newly added operator `<`
overload.

Unit test `test_ip6_address.cpp` is updated to cover the behavior of
the new operator.


simonlingoogle (approved):
LGTM :+1:

src/core/net/ip6_address.hpp
```diff
@@ -233,6 +233,20 @@ class Prefix : public otIp6Prefix, public Clearable<Prefix>
...
+     * A prefix with shorter length is considered smaller than the one with longer length. If the prefix lengths are
+     * equal, then the prefix bytes are compared directly.
+     *
+     * @param[in] aOther  The other prefix to compare against.
+     *
+     * @retval TRUE   If the prefix is smaller than @p aOther.
+     * @retval FALSE  If the prefix is not smaller than @p aOther.
+     *
+     */
+    bool operator<(const Prefix &aOther) const;
```
wgtdkp, line 248: Just some thoughts:

I think it is a good practice to always have one relational operator and dedrive the rest from it. For example, we may consider implement `operator==` with:
```c++
bool operator==(const Prefix &aOther) const
{
    return !( *this < aOther || aOther < *this);
}
```

But it requires two `<` operations and current implementation should faster :)

wgtdkp (approved):
Thanks @abtink.
We originally have the comparator as `operator<` which is exactly the same with your PR. But we decided not to put it in the `Ip6::prefix` class because we didn't find the standard definition of how two prefixes should be compared and the meaning of one prefix being smaller than another. So we defined the method in the Routing Manager to restrict the scope of the prefix comparator. @librasungirl 

But rethink about this, I think the `operator<` is just for making the `Ip6::Prefix` sortable and any strategy should be okay. So this looks good to me as well :)

src/core/net/ip6_address.hpp
abtink: Yes. I think having simpler `operator==()` would be better here. As you mention, definitley faster.

I did not add it for now, but we can later consider adding a CRTP style template class that adds support for other operators `>`, `<=`, `>=` implemented from `<` and `==` provided by the main class.  
We can also add/use similar model for adding `!=` from an existing `==` from the main class (it is practically a syntactic sugar optimization).

jwhui (approved):

tests/unit/test_ip6_address.cpp
```diff
@@ -266,14 +267,37 @@ void TestIp6Prefix(void)
...
                     VerifyOrQuit(!prefix.IsEqual(subPrefix.GetBytes(), subPrefix.GetLength()),
                                  "Prefix::IsEqual() failed");
+                    VerifyOrQuit(subPrefix < prefix, "Prefix::operator<() failed");
                 }
             }
+
+            for (uint8_t bitNumber = 0; bitNumber < prefixLength; bitNumber++)
+            {
+                ot::Ip6::Prefix prefix2;
+                uint8_t         mask  = static_cast<uint16_t>(1U << (7 - (bitNumber & 7)));
```
jwhui, line 284:
```suggestion
                uint8_t         mask  = static_cast<uint8_t>(1U << (7 - (bitNumber & 7)));
```

src/core/border_router/routing_manager.hpp
```diff
@@ -240,9 +240,6 @@ class RoutingManager : public InstanceLocator
                              otRoutePreference  aRoutePreference = OT_ROUTE_PREFERENCE_MED);
     bool NetworkDataContainsOmrPrefix(const Ip6::Prefix &aPrefix) const;
 
-    // Decides the first prefix is numerically smaller than the second one.
-    static bool     IsPrefixSmallerThan(const Ip6::Prefix &aFirstPrefix, const Ip6::Prefix &aSecondPrefix);
-    static bool     IsValidOmrPrefix(const NetworkData::OnMeshPrefixConfig &aOnMeshPrefixConfig);
```
wgtdkp, line 245: @abtink This should be a mistake :)

The CI is failing because of this.

abtink: Yes. Thanks.   :)
I rebased on `main` and I think this was then removed by mistake.

abtink: It's now passing all tests.

PR #6275, opened 2021-03-12
[simulation] do not update CSL IE in retransmission

brianljt:


brianljt:
A retransmission frame should not have its CSL IE updated which will leads to incorrect MIC.

For the CI test case failure, it is caused by racing condition in test.
1. SSED sets CSL channel to 12 which is different from the Thread operating channel(channel 11).
2. Leader sends ping request at channel 12 during sample time.
3. Leader receive Enh-ACK from SSED on channel 12.
4. Leaded change back to channel 11 when this Enh-ACK is set to posix host and `MAC` state is back to idle.
5. SSED sends ping response on channel 11 to leader.

A racing condition is step 5 comes before step 4. So simulation radio will reject the ping response as it is currently on channel 12 while the receiving frame is on channel 11.

SSED will do a retransmission again with CSL IE updated. This breaks the integrity of MIC. Leader will drop the retransmitted frame. Thus ping fail.

bukepo (approved):

examples/platforms/simulation/radio.c
```diff
@@ -679,7 +679,7 @@ void radioSendMessage(otInstance *aInstance)
 #endif // OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT && OPENTHREAD_CONFIG_TIME_SYNC_ENABLE
 
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
-    if (sCslPeriod > 0)
+    if (sCslPeriod > 0 && !sTransmitFrame.mInfo.mTxInfo.mIsARetx)
```
bukepo, line 682: Just realized this may cause the parent lost synchronization with the child. I think there are some confusions about when the "CSL IE" should be included.

From 802.15.4-2015 and Thread spec, it seems the header IE is only required for synchronization messages:

> 2 The SSED Child MAY send periodic traffic to the Router in any form including application
3 data messages or MLE messages to maintain synchronized communication with its Parent.
4 These messages MUST include the CSL IE. An SSED Child that changes its CSLPeriod after
5 initial synchronization SHOULD send an immediate IEEE 802.15.4 MAC Data or Command
6 frame containing a CSL IE to the Router.

So not all messages sent out from SSED need to include "CSL IE". From my understanding, if a message expects retransmissions, it should not include "CSL IE", and a retransmission frame can safely keep identical to its original frame.
Thoughts? @brianljt @jwhui

brianljt: Thank you @bukepo . This is just the first step towards fixing the retransmission issue with CSL IE.

This works for simulation CI tests as the following steps(after out-of-sync) are sending packet from SSED to parent which will make them resync again.

I'm preparing a follow up fix to address the retransmission issue. The idea is to set the retransmission bit in `TxInfo` in radio if it is a retransmission(which breaks the synchronization). Then Mac layer will check this bit in `TxInfo` and schedule a data poll transmission right after the retransmission to fix the wrong CSL IE in previous retransmission. In this way, the parent will just be temporary out of sync with its SSED.

For the approach you mentioned, I think it can be used to mitigate the issue. If the last packet from SSED to its parent is within a certain threshold, then it does not need to include CSL IE in data frame to avoid retransmission. Comments?

Ref: https://threadgroup.atlassian.net/browse/SPEC-887

bukepo (approved):

brianljt:
A follow up PR is #6297

Irving-cl (approved):
LGTM :+1:

jwhui (approved):

Issue #6271, opened 2021-03-11
[github-actions] test_advertising_proxy.py fails intermittently


jwhui:
**Describe the bug**

tests/scripts/thread-cert/border_router/test_advertising_proxy.py fails intermittently.

**Console/log output**

- [logs_60121.zip](https://github.com/openthread/openthread/files/6124867/logs_60121.zip)
- [thread-border-router-results.zip](https://github.com/openthread/openthread/files/6124868/thread-border-router-results.zip)

brianljt:
Another failed run artifacts. FYI.

[thread-border-router-results.zip](https://github.com/openthread/openthread/files/6128236/thread-border-router-results.zip)
[logs_60331.zip](https://github.com/openthread/openthread/files/6128237/logs_60331.zip)

PR #6276, opened 2021-03-12
[tests] fix advertising proxy tests

wgtdkp:
This PR addresses https://github.com/openthread/openthread/issues/6271. The cause of https://github.com/openthread/openthread/issues/6271 is that [zeroconf](https://pypi.org/project/zeroconf/) is not able ot discover the services after updating the host adress. zeroconf was working but maybe there are updates to the GitHub Action containers and it fails now.

We fixed this issue by replacing `zeroconf` with `dns-sd`.

### DIFF
--- a/tests/scripts/thread-cert/node.py
+++ b/tests/scripts/thread-cert/node.py
@@ -39,7 +39,6 @@ import time
 import traceback
 import unittest
 from typing import Union, Dict, Optional, List
-from zeroconf import Zeroconf, ServiceInfo
 
 import pexpect
 import pexpect.popen_spawn
@@ -302,9 +301,6 @@ class OtbrDocker:
     def _setup_sysctl(self):
         self.bash(f'sysctl net.ipv6.conf.{self.ETH_DEV}.accept_ra=2')
         self.bash(f'sysctl net.ipv6.conf.{self.ETH_DEV}.accept_ra_rt_info_max_plen=64')
-
-        # Zeroconf complains that there is not enough BUFS to subscribe multicast groups.
-        self.bash('sysctl net.ipv4.igmp_max_memberships=1024')
 
 
 class OtCli:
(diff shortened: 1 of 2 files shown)


tests/scripts/thread-cert/node.py
```diff
@@ -2727,27 +2723,52 @@ def discover_mdns_service(self, instance, name, host_name, timeout=2):
...
-                    'addresses': info.parsed_addresses()
-                }
-        finally:
-            zeroconf.close()
+
+        self.bash(f'dns-sd -Z {name} local. > /tmp/{name} 2>&1 &')
+        self.bash(f'dns-sd -G v6 {host_name}.local. > /tmp/{host_name} 2>&1 &')
+        time.sleep(timeout)
+
+        self.bash(f'pkill dns-sd')
```
simonlingoogle, line 2731:
```suggestion
        self.bash('pkill dns-sd')
```

tests/scripts/thread-cert/node.py
```diff
@@ -2727,27 +2723,52 @@ def discover_mdns_service(self, instance, name, host_name, timeout=2):
...
+        # Timestamp     A/R Flags if Hostname                               Address                                     TTL
+        # 9:38:09.274  Add     23 48 my-host.local.                         2001:0000:0000:0000:0000:0000:0000:0002%<0>  120
+        #
+        for line in self.bash(f'cat /tmp/{host_name}'):
+            elements = line.split()
+            if not elements or len(elements) < 6 or not elements[4].startswith(host_name):
+                continue
+            addresses.append(elements[5].split('%')[0])
+
+        print(addresses)
```
simonlingoogle, line 2748:
```suggestion
```
remove debugging prints ?

simonlingoogle (approved):
LGTM :+1: with nits

I am actually glad that we removed `zeroconf`.

tests/scripts/thread-cert/node.py
wgtdkp: It will be helpful if the test case fails.

jwhui (approved):
Thanks! 👍🏼

Issue #6224, opened 2021-03-02
Compilation issues inside radio_spinel_impl.hpp


doru91:
The compilation issues seems to be caused by commit id 33f15fac2faad5aa31920a24cbbc8b2fc49dbe08

Compilation issues:

```
../openthread/src/lib/spinel/radio_spinel_impl.hpp: In member function 'otError ot::Spinel::RadioSpinel<InterfaceType, ProcessContextType>::RestoreDatasetFromNcp()':
../openthread/src/lib/spinel/radio_spinel_impl.hpp:420:19: error: no matching function for call to 'ot::Instance::Get()'
  420 |     Instance::Get().template Get<SettingsDriver>().Init();
      |                   ^
In file included from ../openthread/src/lib/spinel/radio_spinel_impl.hpp:45,
                 from ../openthread/src/lib/spinel/radio_spinel.hpp:1015,
                 from ../src/imx_rt/rt1060/radio.cpp:39:
../openthread/src/core/common/instance.hpp:298:43: note: candidate: 'template<class Type> Type& ot::Instance::Get()'
  298 |     template <typename Type> inline Type &Get(void);
      |                                           ^~~
../openthread/src/core/common/instance.hpp:298:43: note:   template argument deduction/substitution failed:
In file included from ../openthread/src/lib/spinel/radio_spinel.hpp:1015,
                 from ../src/imx_rt/rt1060/radio.cpp:39:
../openthread/src/lib/spinel/radio_spinel_impl.hpp:420:19: note:   couldn't deduce template parameter 'Type'
  420 |     Instance::Get().template Get<SettingsDriver>().Init();
      |                   ^
../openthread/src/lib/spinel/radio_spinel_impl.hpp:420:48: error: expected primary-expression before '>' token
  420 |     Instance::Get().template Get<SettingsDriver>().Init();
      |                                                ^
../openthread/src/lib/spinel/radio_spinel_impl.hpp:420:50: error: expected primary-expression before ')' token
  420 |     Instance::Get().template Get<SettingsDriver>().Init();
      |                                                  ^
../openthread/src/lib/spinel/radio_spinel_impl.hpp:429:19: error: no matching function for call to 'ot::Instance::Get()'
  429 |     Instance::Get().template Get<SettingsDriver>().Deinit();
      |                   ^
In file included from ../openthread/src/lib/spinel/radio_spinel_impl.hpp:45,
                 from ../openthread/src/lib/spinel/radio_spinel.hpp:1015,
[...]

PR #6277, opened 2021-03-13
[radio-spinel] perform "RestoreDatasetFromNcp()" with single OT instance config

abtink:
This commit changes the `RestoreDatasetFromNcp()` to be performed on
single OT instance config and to be skipped when multiple OT instance
config is enabled. This change ensures that the `RadioSpinel` module
can be compiled with `MULTIPLE_INSTANCE_ENABLE` config. Note that the
"restore data set from NCP" mechanism is performed during platform
initialization from `RadioSpinel::Init()` which happens before
any OT instance is allocated or initialized (all platform layer
components/drivers should be initialized before OT instance). So the
"restore dataset from NCP" feature only makes sense under since OT
instance case.

----

Should help address issue https://github.com/openthread/openthread/issues/6224.

### DIFF
--- a/src/lib/spinel/radio_spinel_impl.hpp
+++ b/src/lib/spinel/radio_spinel_impl.hpp
@@ -248,7 +248,9 @@ void RadioSpinel<InterfaceType, ProcessContextType>::Init(bool aResetRadio,
 
         if (aRestoreDatasetFromNcp)
         {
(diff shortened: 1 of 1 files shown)


Irving-cl (approved):
LGTM 👍

brianljt (approved):
LGTM, Thanks! 👍

jwhui (approved):

PR #6280, opened 2021-03-15
[border-router] send at least one router solicitation

wgtdkp:
Per section 6.3 of RFC 4861:

> a host SHOULD send at least one solicitation in the case where an
advertisement is received prior to having sent a solicitation. Responses
to solicited advertisements may contain more information than unsolicited
advertisements.

In the Thread BR case, there seems no difference between solicited and unsolicited
RA. But consider that there are multiple BRs on the infra link, we may not learn
RIOs of other BRs if the router discovery process is stopped immediately
before sending any RS messages because of receiving an unsolicited RA.

This change ensures that at least one RS message will sent once started
the router discovery process.

### DIFF
--- a/src/core/border_router/routing_manager.cpp
+++ b/src/core/border_router/routing_manager.cpp
@@ -1122,12 +1122,18 @@ bool RoutingManager::AddDiscoveredPrefix(const Ip6::Prefix &aPrefix,
 
         SuccessOrExit(AddExternalRoute(aPrefix, aRoutePreference));
 
-        if (aIsOnLinkPrefix)
+        if (aIsOnLinkPrefix && mRouterSolicitCount > 0)
         {
             // Stop Router Solicitation if we discovered a valid on-link prefix.
             // Otherwise, we wait till the Router Solicitation process times out.
             // So the maximum delay before the Border Router starts advertising
-            // its own on-link prefix is 9 (4 + 4 + 1) seconds.
+            // its own on-link prefix is 10 seconds = RS_DELAY (1) + RS_INTERNAL (4)
+            // + RS_INTERVAL (4) + RS_DELAY (1).
+            //
+            // Always send at least one RS message so that all BRs will respond.
+            // Consider that there are multiple BRs on the infra link, we may not learn
+            // RIOs of other BRs if the router discovery process is stopped immediately
+            // before sending any RS messages because of receiving an unsolicited RA.
             mRouterSolicitTimer.Stop();
         }
 


simonlingoogle (approved):
LGTM :+1:

src/core/border_router/routing_manager.cpp
```diff
@@ -1129,12 +1129,17 @@ bool RoutingManager::AddDiscoveredPrefix(const Ip6::Prefix &aPrefix,
...
             // Stop Router Solicitation if we discovered a valid on-link prefix.
             // Otherwise, we wait till the Router Solicitation process times out.
             // So the maximum delay before the Border Router starts advertising
-            // its own on-link prefix is 9 (4 + 4 + 1) seconds.
+            // its own on-link prefix is 10 seconds = RS_DELAY (1) + RS_INTERNAL (4)
+            // + RS_INTERVAL (4) + RS_DELAY (1).
+            //
+            // Per section 6.3 of RFC 4861, a host SHOULD send at least one
+            // solicitation in the case where an advertisement is received
+            // prior to having sent a solicitation.
```
jwhui, line 1142: The purpose of this is because: 
```
   Responses to solicited advertisements may contain more information
   than unsolicited advertisements.
```
Are there situations where we might be missing information if we do not send the Router Solicitation?

wgtdkp: Consider that there are multiple border routers on the infra link, we may miss RIOs of other BRs if an unsolicited RA is received and we stop sending RS immediately. Always sending RS so that every RSs will response. Thoughts?

jwhui: Yes, that makes sense. However, I think this is a different reason than what the comment suggests. How about updating the comment to reflect the true purpose of requiring an RS?

src/core/border_router/routing_manager.cpp
```diff
@@ -1122,12 +1122,22 @@ bool RoutingManager::AddDiscoveredPrefix(const Ip6::Prefix &aPrefix,
...
+            // + RS_INTERVAL (4) + RS_DELAY (1).
+            //
+            // Per section 6.3 of RFC 4861, a host SHOULD send at least one
+            // solicitation in the case where an advertisement is received prior
+            // to having sent a solicitation. Because responses to solicited
+            // advertisements may contain more information than unsolicited
+            // advertisements. For example, if there are multiple BRs on the same
+            // infra link, we will miss RIOs of other BRs if an unsolicited RA is
+            // received and the router discovery process is stopped immediately.
+            // Always sending at least one RS message so that all BRs will respond.
```
jwhui, line 1140:
```suggestion
            // advertisements. In addition, if there are multiple BRs on the same
            // infra link, we may not learn RIOs of other BRs if an unsolicited RA is
            // received and the router discovery process is stopped immediately.
            // Always sending at least one RS message so that all BRs will respond.
```

Note that the RS has two purposes:
1. To ask the router to provide all information in the RA.
2. To ask other routers to send RAs.

I tried to make a minor tweak to the comment to ensure that both are captured.

src/core/border_router/routing_manager.cpp
wgtdkp: @jwhui Thanks for the suggestion! I updated the comments accordingly, PTAL.

src/core/border_router/routing_manager.cpp
wgtdkp: Sorry I made a new force push that removed the mention of RFC 4861 requirements, do you think we should keep it?

jwhui: I think what you have in the PR now (without the mention of RFC 4861) is good. Thanks!

jwhui (approved):

PR #6281, opened 2021-03-15
[build] set Thread 1.2 as default version

jwhui:
This PR includes a number of minor fixes necessary for GitHub Actions
checks to pass.


.github/workflows/posix.yml
```diff
@@ -47,6 +47,7 @@ jobs:
       CXXFLAGS: -DCLI_COAP_SECURE_USE_COAP_DEFAULT_HANDLER=1 -DOPENTHREAD_CONFIG_MLE_MAX_CHILDREN=15 -DOPENTHREAD_CONFIG_MAC_SCAN_DURATION=500
       CC: clang
       CXX: clang++
+      THREAD_VERSION: 1.1
```
jwhui, line 50: Thread 1.2 multicast feature does not build on macOS.

.github/workflows/posix.yml
```diff
@@ -142,6 +143,7 @@ jobs:
       PYTHONUNBUFFERED: 1
       READLINE: readline
       REFERENCE_DEVICE: 1
+      THREAD_VERSION: 1.1
```
jwhui, line 146: Tests do not pass with 1.2 enabled.

.github/workflows/posix.yml
```diff
@@ -182,6 +184,7 @@ jobs:
...
```
jwhui, line 187: Tests do not pass with 1.2 enabled.

simonlingoogle (approved):
LGTM :+1: 

We can add 1.2 tests later

.github/workflows/simulation-1.1.yml
```diff
@@ -316,6 +323,8 @@ jobs:
 
   external-commissioner:
     runs-on: ubuntu-20.04
+    env:
+      THREAD_VERSION: 1.1
```
wgtdkp, line 327: But the CI tests on `ot-commissioner` master will be broken.

jwhui: Do you have a suggested path forward?

abtink (approved):
👍

.github/workflows/simulation-1.1.yml
wgtdkp: I think I misunderstood this change. I thought there are some external commissioner failures that force you setting the `THREAD_VERSION` to 1.1 for this test case, but seems we are just using 1.1 for all test cases in this workflow and the workflow itself is also renamed to `simulation-1.1.yml`.

Actually, the external-commissioner test script doesn't respect the `THREAD_VERSION` env variable and it actually tests with latest openthread master commit and the CI on `ot-commissioner` master  will conitnue working.

wgtdkp (approved):
LGTM :+1:

bukepo (approved):
LGTM:+1:

bukepo:
Do we still need check size diff for 1.1?

jwhui:
> Do we still need check size diff for 1.1?

Good question. How about we address this in a separate PR?

bukepo:
> > Do we still need check size diff for 1.1?
> 
> Good question. How about we address this in a separate PR?

Sure!

PR #6284, opened 2021-03-15
[utils] add 'Utils::PingSender'

abtink:
This commit adds a new module `Utils::PingSender`. This module
is then used by CLI to support ping command.

----

**_Background_**: This change helps remove the need for CLI to use core 
types and definition (like `Timer`).


src/core/utils/ping_sender.hpp
```diff
@@ -0,0 +1,154 @@
...
+#include "common/timer.hpp"
+#include "net/icmp6.hpp"
+#include "net/ip6_address.hpp"
+
+namespace ot {
+namespace Utils {
+
+/**
+ * This class implements sending ping (ICMPv6 echo request) messages and processing the ping (ICMP echo response)
+ * message.
```
jwhui, line 57:
```suggestion
 * This class implements sending ICMPv6 Echo Request messages and processing ICMPv6 Echo Reply
 * messages.
```

src/core/utils/ping_sender.hpp
```diff
@@ -0,0 +1,154 @@
...
+{
+public:
+    /**
+     * This type represents a ping response.
+     *
+     */
+    typedef otPingSenderResponse Response;
+
+    /**
+     * This type represents a ping request configuration.
```
jwhui, line 70:
```suggestion
     * This class represents a ping request configuration.
```

src/core/config/ping_sender.h
```diff
@@ -0,0 +1,82 @@
...
+#ifndef CONFIG_PING_SENDER_H_
+#define CONFIG_PING_SENDER_H_
+
+/**
+ * @def OPENTHREAD_CONFIG_PING_SENDER_ENABLE
+ *
+ * Define to 1 to enable ping sender module.
+ *
+ * Ping sender module implements sending ping (ICMPv6 Echo Request) messages and processing the ping (ICMP Echo
+ * Response) message.
```
jwhui, line 44:
```suggestion
 * Ping sender module implements sending ICMPv6 Echo Request messages and processing ICMPv6 Echo
 * Reply messages.
```

include/openthread/ping_sender.h
```diff
@@ -0,0 +1,110 @@
...
+#include <openthread/error.h>
+#include <openthread/instance.h>
+#include <openthread/ip6.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * This structure represents a ping response.
```
jwhui, line 49: nit: RFC 4443 uses the term "reply".

include/openthread/ping_sender.h
```diff
@@ -0,0 +1,110 @@
...
+ *
+ */
+typedef struct otPingSenderResponse
+{
+    otIp6Address mSenderAddress;  ///< Response sender IPv6 address
+    uint32_t     mRoundTripTime;  ///< Round trip time in msec
+    uint16_t     mSize;           ///< Data size (number of bytes) in response (excluding IP and ICMP6 header).
```
jwhui, line 56:
```suggestion
    uint16_t     mSize;           ///< Data size (number of bytes) in response (excluding IPv6 and ICMPv6 headers).
```

include/openthread/ping_sender.h
abtink: Thanks. Renamed the types and docs to use "reply".

src/core/utils/ping_sender.cpp
```diff
@@ -0,0 +1,196 @@
...
+    SuccessOrExit(Get<Ip6::Icmp>().SendEchoRequest(*message, messageInfo, mIdentifier));
+
+#if OPENTHREAD_CONFIG_OTNS_ENABLE
+    Get<Utils::Otns>().EmitPingRequest(mConfig.GetDestination(), mConfig.mSize, now.GetValue(), mConfig.mHopLimit);
+#endif
+
+    message = nullptr;
+
+exit:
+    FreeMessage(message);
```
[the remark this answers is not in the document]
abtink: We set the `message = nullptr` at the end of this method before the `exit` label (basically if everything is successful and `message` is passed to `Ip6::Icmp` we set `message` back to `nullptr`). So with this in the `exit` label, we can use `FreeMessage(message)` which frees it if it is not already `nullptr`.

We don't track or return any `error` here. We can consider tracking error, but the main use would be to log. I initially considered adding logs in this module but did not add any to keep the code size smaller. Also I am not sure if logs in this module would add much value since `Ip6::Icmp` class already logs tx of echo request and then we have logs in `MeshForwarder` about TX/RX of IPv6 message with the type and src/dst addresses.

src/core/config/ping_sender.h
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * @def OPENTHREAD_CONFIG_PING_SENDER_ENABLE
+ *
+ * Define to 1 to enable ping sender module.
+ *
+ * Ping sender module implements sending ICMPv6 Echo Request messages and processing ICMPv6 Echo Reply messages.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_PING_SENDER_ENABLE
+#define OPENTHREAD_CONFIG_PING_SENDER_ENABLE 1
```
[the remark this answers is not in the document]
abtink: I agree it make sense to not have it enabled by default. I kept it enabled since it is used by CLI and I think ping operation may be used by some of the tests. I will add `CMake` and autoconf build option for `PING_SENDER` and then have it enabled only under certain CI builds. 

One interesting observation here:
- I expected it to have little code increase on NCP build, but we see it increased its code size (I think in practical case where an NCP build is being used, it makes sense for project integrator to disable this feature).
- I assumed since NCP code would not use the new APIs most of the code would be removed during link time.
- The complier can indeed remove unused functions/methods, but I think it may be able to detect that some callbacks (e.g. `Timer` or `Icmp` callbacks) are not going to be used. And those add overhead to the code size.

simonlingoogle (approved):
LGTM :+1: 

Thanks

abtink:
In the last push, by default the feature is disabled, but added clang `OT_PING_SENDER` and common-switches `PING_SENDER` configs and they are then used in a bunch of related make/build files.

jwhui:
> In the last push, by default the feature is disabled, but added clang `OT_PING_SENDER` and common-switches `PING_SENDER` configs and they are then used in a bunch of related make/build files.

@abtink , thinking about this more, it seems that most people would want the ping sender feature enabled when CLI is enabled. Can we default the ping sender to enabled if CLI is enabled?

abtink:
> @abtink , thinking about this more, it seems that most people would want the ping sender feature enabled when CLI is enabled. Can we default the ping sender to enabled if CLI is enabled?

I considered the same thing (to have it enabled by default under CLI). But I am not sure if there is an easy to tell when building OT core (as library) whether it is going to be used with CLI or NCP or other build. 
- We can try to do a check in the CLI source code that requires the `PING_SENDER` to be enabled (give an error), but I am not sure if it is a good way. 
- Do you have any suggestions/ideas on this?

jwhui:
> I considered the same thing (to have it enabled by default under CLI). But I am not sure if there is an easy way to tell when building OT core (as a library) whether it is going to be used in a CLI or NCP or other build.
> 
> * We can try to do a check in the CLI source code that requires the `PING_SENDER` to be enabled (give an error), but I am not sure if it is a good way.
> * Do you have any suggestions/ideas on this?

I was thinking in the Makefile and CMake files to enable PING_SENDER if building the CLI example is enabled. Thoughts?

abtink:
| I was thinking in the Makefile and CMake files to enable PING_SENDER if building the CLI example is enabled. Thoughts?

Sounds good to me. I added it under `CMake`
```
option(OT_PING_SENDER "enable ping sender support")
if(OT_PING_SENDER OR OT_APP_CLI)
    target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_PING_SENDER_ENABLE=1")
endif()
```
For auto-conf, I cannot find a good way to add it. So maybe we can require explicit enable for auto-conf?

etc/cmake/options.cmake
```diff
@@ -257,6 +257,11 @@ if(OT_MULTIPLE_INSTANCE)
     target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE=1")
 endif()
 
+option(OT_PING_SENDER "enable ping sender support")
+if(OT_PING_SENDER OR OT_APP_CLI)
+    target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_PING_SENDER_ENABLE=1")
+endif()
```
wgtdkp, line 263: nit:
```suggestion
option(OT_PING_SENDER "enable ping sender support" ${OT_APP_CLI})
if(OT_PING_SENDER)
    target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_PING_SENDER_ENABLE=1")
endif()
```

Per CMake documentation for [option](https://cmake.org/cmake/help/latest/command/option.html):
> option(<variable> "<help_text>" [value])
Provides an option for the user to select as ON or OFF. If no initial <value> is provided, OFF is used. If <variable> is already set as a normal or cache variable, then the command does nothing (see policy CMP0077).

The suggested change gives a chance to explicitly disable `OT_PING_SENDER` when `OT_APP_CLI` is enabled.

wgtdkp (approved):
LGTM :+1:

etc/cmake/options.cmake
abtink: Good suggestion. Thanks @wgtdkp. Changed it in the last push.

PR #6285, opened 2021-03-16
[dnssd-server] bind to Thread network interface

simonlingoogle:
This commit binds the DNS-SD server to the Thread network interface to avoid potential port conflicts.


simonlingoogle:
Issue resolved by #6318

PR #6287, opened 2021-03-16
[nrf528xx] improve CSL phase calculation

edmont:
Fill a CSL Phase IE with a value for the CSL transmitter to aim at the center of the receive window, instead of aiming at the beginning.

Find attached two test captures performed with two nRF52840 DKs. CSL is configured with 500 ms period and 60 s timeout. Pings are sent by the parent every 50 s. We can observe how the phase correction provides much better stability to the CSL communication.

[pcaps.zip](https://github.com/openthread/openthread/files/6148186/pcaps.zip)

### DIFF
--- a/examples/platforms/nrf528xx/src/radio.c
+++ b/examples/platforms/nrf528xx/src/radio.c
@@ -51,7 +51,6 @@
 #include <platform-config.h>
 #include <openthread/platform/alarm-micro.h>
 #include <openthread/platform/diag.h>
-#include <openthread/platform/logging.h>
 #include <openthread/platform/radio.h>
 #include <openthread/platform/time.h>
 
@@ -119,9 +118,10 @@ static uint8_t  sEnergyDetectionChannel;
 static int8_t   sEnergyDetected;
 
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
-static uint32_t      sCslPeriod;
-static uint32_t      sCslSampleTime;
-static const uint8_t sCslIeHeader[OT_IE_HEADER_SIZE] = {CSL_IE_HEADER_BYTES_LO, CSL_IE_HEADER_BYTES_HI};
+static uint32_t       sCslPeriod;
+static uint32_t       sCslSampleTime;
+static const uint32_t sCslSampleDur                   = OPENTHREAD_CONFIG_CSL_SAMPLE_WINDOW * OT_US_PER_TEN_SYMBOLS;
+static const uint8_t  sCslIeHeader[OT_IE_HEADER_SIZE] = {CSL_IE_HEADER_BYTES_LO, CSL_IE_HEADER_BYTES_HI};
 #endif // OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
 
 typedef enum
@@ -1068,8 +1068,9 @@ static uint16_t getCslPhase()
 {
     uint32_t curTime       = otPlatAlarmMicroGetNow();
     uint32_t cslPeriodInUs = sCslPeriod * OT_US_PER_TEN_SYMBOLS;
-    uint32_t diff = (cslPeriodInUs - (curTime % cslPeriodInUs) + (sCslSampleTime % cslPeriodInUs)) % cslPeriodInUs;
-
+    uint32_t diff =
+        (cslPeriodInUs - (curTime % cslPeriodInUs) + (sCslSampleTime % cslPeriodInUs) + (sCslSampleDur / 2)) %
+        cslPeriodInUs;
     return (uint16_t)(diff / OT_US_PER_TEN_SYMBOLS + 1);
 }
 #endif


hubertmis (approved):
Thanks for addressing this!

hubertmis:
@ankuns, FYI

Irving-cl (approved):
LGTM 👍

jwhui (approved):

brianljt:
There is a parameter called `kCslReceiveTimeAhead` in `SubMac::HandleCslTimer`, which means in current implementation, CSL receiver already wakes up 30 symbols ahead. So the actual sample window is 80 symbols and the phase conveyed to the CSL transmitter is T<sub>phase</sub> = T<sub>wake</sub> + 30 and T<sub>sleep</sub> = T<sub>phase</sub> + 50 = T<sub>wake</sub> + 80.

The change here is just adding more time to wake. I guess there are some delay when `SubMac` calls `otPlatRadioReceive`  until radio is really pull up from sleep state. So I think it would be better to fix this using some API like `otPlatRadioReceiveAt` if radio driver supports this accurate wake or adjust `kCslReceiveTimeAhead` according to platform differences. How do you think? @Irving-cl @bukepo

jwhui:
I think T<sub>phase</sub> should communicate the ideal time at when the CSL transmitter should start transmitting.

Then the CSL receiver should ensure that sample window is centered on T<sub>phase</sub>, because it assumes that the drift and error can either be going forward or backward. So if the window size is 10ms, the receiver should be ready to receive at T<sub>phase</sub>-5 and turn off at T<sub>phase</sub>+5.

bukepo:
Yes, I think the value of CSL Phase should be a fixed value in normal case until the receiver intentionally want to change it, because there's no ack to for enh-ack. If the receiver recalculates this value every time it fills in the CSL IE, the chance of the receiver and transmitter having different CSL Phase values increases.

brianljt:
I agree to use `kCslReceiveTimeAhead` and put T<sub>phase</sub> in the middle of receiver's sample window. But this PR also points out another problem, on nrf528xx platform, the delay between `otPlatRadioReceive` is called and the time when radio is ready to receive is big and it is something we should try to mitigate. If the delay is big, then CSL receiver should use a bigger sample window.

hubertmis:
> If the receiver recalculates this value every time it fills in the CSL IE, the chance of the receiver and transmitter having different CSL Phase time increases.

But SSED must recalculate CSL Phase each time it transmits something. Scheduled RX window has constant position and SSED's MAC has little control deciding when to transmit to parent. Data frames should be sent ASAP, while Acks have strict timing requirements. One value is constant, the other varies. Delta must be calculated each time and most probably the result is different for each transmitted frame.

> the delay between `otPlatRadioReceive` is called and the time when radio is ready to receive is big and it is something we should try to mitigate

Unfortunately, it's hard to mitigate in OT.  There are two reasons:

1. OT alarm is not deterministic enough to use it for precise timings of radio operations. The callback may be called sooner or later depending on alarm platform implementation, other tasks executed in the main program loop, ISR latency. These delays are specific for given application and are not deterministic. The worst case value could be passed from application to OT during initialization. 
2. OT does not know how much time given radio driver needs to enable receiver. This time may also be undeterministic and depend on current crystal state, coexistence interface, multiprotocol scheduling, ISR latency. The worst case may be again passed from the radio driver to OT.

OT could take into account both worst cases and sum them to `mCslReceiveTimeAhead`, however it would not be very power efficient, because the device would enable RX much too early, preparing for the worst case that happens rarely. 

The correct and power efficient solution is to schedule precise radio operations (like opening RX window) to the radio driver. The driver should not be affected by alarm.c problems or main loop latency (it depends on particular implementation). And the driver may take into account current circumstances known only to this module to enable receiver precisely at the time requested by OT.

Following these thoughts, we created receive_at() service in Nordic radio driver that should be used by OT to schedule RX windows precisely. 

Please note that above explanation is only about how to start RX window and how to take into account undeterministic timings to make sure it is not started too late. It does not change that CSL Phase should point in the middle of planned RX window.

If CSL Phase pointed to the beginning of the window, and window was started too early to mitigate the worst case timings, usually it would work correctly, because window would be open before time stated in CSL Phase. However, sometimes when the program hits the worst case timings or when receive_at() would be used, it would fail. The CSL Phase would point to the moment when RX starts, and because of uncertainty and drift, transmitter may send earlier. 

To sum up, we have two problems to address:

1. Uncertainty of receiver that should be addressed by using receive_at(), alternatively by `mCslReceiveTimeAhead` taken from application and radio driver.
2. Uncertainty of transceiver and relative clock drifts, mitigated by scheduling long enough RX Window and pointing CSL Phase in the middle of this window.

brianljt:
@hubertmis Maybe also a `sleep_at` API is needed for precise sample window control or use a pre-config number if radio does not support this API? As you pointed out, the time OT stack,

1. Does not have deterministic alarm.
2. It does not know how long does radio driver needs to finally put its tranceiver to sleep.

So it would be not precise enough for OT stack to schedule the sleep time according to its timer in alarm.

hubertmis:
> Maybe also a `sleep_at` API is needed for precise sample window control

receive_at() function we designed as timing arguments take:
1. Window start timestamps
2. Window duration

This way you don't need separated sleep_at() to handle closing window precisely. 

For `mCslReceiveTimeAhead` solution it is not needed to worry about timings while closing window, because processing time and other latencies will cause closing it too late. It won't be energy efficient but will work.

bukepo:
> >If the receiver recalculates this value every time it fills in the CSL IE, the chance of the receiver and transmitter having different CSL Phase time increases.

> But SSED must recalculate CSL Phase each time it transmits something. Scheduled RX window has constant position and SSED's MAC has little control deciding when to transmit to parent. Data frames should be sent ASAP, while Acks have strict timing requirements. One value is constant, the other varies. Delta must be calculated each time and most probably the result is different for each transmitted frame.

Sorry for the inaccurate description. I mean T<sub>phase</sub> should mean the same offset in each CSL period, which should not depend on the sample window size. And it's the receiver's actually sample window and sample time that need to be adjusted according to the estimation of time drift distribution of the transmitter.

hubertmis:
@bukepo, I think we are describing the same using different words. To clarify I've sketched a diagram. Do you agree with timings in this diagram?

image

PR #6288, opened 2021-03-16
[ncp] remove unused code

lmaciejonczyk:
For the case when otInstanceReset returns during pseudo reset action
setting here SPINEL_STATUS_RESET_SOFTWARE is useless as line below
the current sNcpInstance is cleared. The new status never gets
a chance to be transmitted to the host.


abtink (approved):
LGTM. 👍

jwhui (approved):

PR #6293, opened 2021-03-17
[indirect-sender] skip over an empty frame in 'HandleSentFrameToChild()'

### LINKED ISSUE #6298: [github-actions] toranj-rcp-trel fails intermittently
toranj-rcp-trel fails intermittently

[logs-61324.txt](https://github.com/openthread/openthread/files/6154782/logs-61324.txt)

abtink:
This commit updates `IndirectSender` to check and exit quickly from the
`HandleSentFrameToChild()` callback if the sent frame is an empty
frame. `PrepareFrameForChild()` prepares an empty frame when there is
no indirect message in the send queue for the child. This can happen
in the (not common) situation where the radio platform does not
support the "source address match" feature and always include "frame
pending" flag in acks to data poll frames. The change in this commit
address a potential issue in this situation and ensures that the "send
done" is not incorrectly processed for any newly added indirect
message after preparing the empty frame (which would cause the message
to be assumed sent and removed without actual tx).

------

**_Background:_**
We discovered this issue investigating the occasional failures of the toranj `test-007-traffic-router-sleepy.py` when used under TREL radio platform. The situation is that a message may be occasionally dropped when the radio platform layer does not support "source address match", i.e. it always includes "frame pending" in acks to data polls from its children.

Note that TREL is not really intended for use by sleepy devices and uses deferred ack model, so from perspective of data poll acks, it behaves as if it always responded to acks with "frame pending".  This behavior is mainly added to be able to run the same set of tests under TREL radio platform.

Fixes #6298


src/core/thread/indirect_sender.cpp
```diff
@@ -422,10 +423,28 @@ void IndirectSender::HandleSentFrameToChild(const Mac::TxFrame &aFrame,
...
+    // frame generated by `PrepareFrameForChild()` when there was no
+    // indirect message in the send queue for the child. This can happen
+    // in the (not common) case where the radio platform does not
+    // support the "source address match" feature and always include
+    // "frame pending" flag in acks to data poll frames. In such a case,
+    // `IndirectSender` prepares and sends an empty frame to the child
+    // after it sends a data poll. Here in `HandleSentFrameToChild()` we
+    // exit quickly if we detect the "send done" is for the empty frame
+    // to ensure we do not update any newly added indirect message after
+    // preparing the empty frame.
```
jwhui, line 441:
```suggestion
    // A zero `nextOffset` indicates that the sent frame is an empty
    // frame generated by `PrepareFrameForChild()` when there was no
    // indirect message in the send queue for the child. This can happen
    // in the (not common) case where the radio platform does not
    // support the "source address match" feature and always includes
    // "frame pending" flag in acks to data poll frames. In such a case,
    // `IndirectSender` prepares and sends an empty frame to the child
    // after it sends a data poll. Here in `HandleSentFrameToChild()` we
    // exit quickly if we detect the "send done" is for the empty frame
    // to ensure we do not update any newly added indirect message after
    // preparing the empty frame.
```

jwhui (approved):

PR #6295, opened 2021-03-17
[tests] wait longer after updating SRP host address

wgtdkp:


tests/scripts/thread-cert/border_router/test_advertising_proxy.py
```diff
@@ -140,7 +140,7 @@ def test(self):
         #
 
         client.srp_client_set_host_address('2001::2')
-        self.simulator.go(2)
+        self.simulator.go(4)
```
simonlingoogle, line 143:
```suggestion
        self.simulator.go(10)
```

nit: We can use longer time for stability.

simonlingoogle (approved):
LGTM 👍 

with a nit

abtink (approved):
LGTM. 👍

PR #6297, opened 2021-03-17
[low-power] resync SSED with its parent after retransmission

brianljt:
Per 15.4 spec, transmitter cannot modify frame content in retransmission which will leads to out-of-date CSL IE being sent to CSL transmitter. This PR fixes this issue by sending a follow up data poll with CSL IE to resync the CSL transmitter.

IEEE Std 802.15.4 - 6.7.4.3 Retransmission,

> When not using TSCH mode and a frame with the Security Enabled field set to one is retransmitted, the frame shall be retransmitted without changes and without passing through the outgoing frame security procedure, as defined in 9.2.1.


brianljt:
As discussed with @bukepo , we will remove the CSL IE in normal data frame and data poll and only include CSL IE in the frame which will be sent out once without retransmission(like Enh-ACK). SSED will keep a sync timer and use data poll without `Mac` retry to resync its parent when there are no packet sent before sync timeout.

Move this PR to WIP state.

PR #6306, opened 2021-03-18
[timer] simplify 'TrickleTimer'

abtink:
This commit simplifies and enhances the `TrickleTimer` class. It adds
support for suppression behavior and removes unused `Mode` and unused
end of interval callback.

### DIFF
--- a/src/core/common/trickle_timer.hpp
+++ b/src/core/common/trickle_timer.hpp
@@ -36,6 +36,7 @@
 
 #include "openthread-core-config.h"
 
+#include "common/numeric_limits.hpp"
 #include "common/timer.hpp"
 
 namespace ot {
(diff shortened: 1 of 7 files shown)


src/core/common/trickle_timer.hpp
```diff
@@ -102,60 +102,85 @@ class TrickleTimer : public TimerMilli
...
+    void Stop(void) { TimerMilli::Stop(); }
 
-#ifdef ENABLE_TRICKLE_TIMER_SUPPRESSION_SUPPORT
     /**
-     * This method indicates to the trickle timer a 'consistent' state.
+     * This method indicates to the trickle timer a 'consistent' event.
+     *
+     * The 'consistent' events are used to control suppression behavior. The trickle timer keeps track of number of
+     * 'consistent' event in each interval. The timer handler is invoked only if the number of `consistent` events
+     * received in the interval is less than the redundancy constant.
```
jwhui, line 142:
```suggestion
     * The 'consistent' events are used to control suppression behavior. The trickle timer keeps track of the number of
     * 'consistent' events in each interval. The timer handler is invoked only if the number of `consistent` events
     * received in the interval is less than the redundancy constant.
```

src/core/common/trickle_timer.hpp
```diff
@@ -102,60 +102,85 @@ class TrickleTimer : public TimerMilli
...
-    uint32_t mIntervalMax;            // Maximum interval (aka `Imax`).
-    uint32_t mInterval;               // Current interval (aka `I`).
-    uint32_t mTimeInInterval;         // Time in interval (aka `t`).
-    Handler  mTransmitHandler;        // Transmit handler callback.
-    Handler  mIntervalExpiredHandler; // Interval expired handler callback.
-    Mode     mMode;                   // Trickle timer mode.
-    bool     mIsRunning : 1;          // Indicates if the trickle timer is running.
-    bool     mInTransmitPhase : 1;    // Indicates if in transmit phase (before time `t` in current interval `I`).
+
+    // Shadow base class `TimerMilli` methods to ensure they are hidden.
```
jwhui, line 169: Can we make `TimerMilli` private?

abtink: I was also thinking of the same and wanted to see if we can improve this 😃 
But I could not find a good way for it:

-  First approach I considered was instead of inheriting from `TimerMilli` we define it as `mTimer` member variable
   - The problem with this is that from the callback of `mTimer`, we cannot back to the owning the `TrickleTimer` object (we can use `offsetof()` with pointer tricks but not safe/recommended for all tool-chains).

- We cannot define the inheritance from `TimerMilli` as `private` because we need to publicly inherit the `InstanceLocator` methods (which are used in the callback to get back to the object owning the `TrickleTimer`).
   - We can try to use `virtual` inheritance to allow multiple base class inheritance from `InstacneLocator` but that itself adds its own overheads.

- Another option I considered was to use a `mTimer` member variable but define it as `TimerMilliContext` (which explicily saves some context).
   - This works fine but increases the RAM size of `TrickleTimer`.

- Finally I opted to continue to use the existing model (inherit publicly form `TimerMill` but try to hide the methods that we don't want by shadow private definitions).

jwhui (approved):

Issue #6304, opened 2021-03-17
[github-actions] test_reset.py fails intermittently


jwhui:
**Describe the bug**

tests/scripts/thread-cert/test_reset.py fails intermittently.

**Console/log output**

- [logs_61590.zip](https://github.com/openthread/openthread/files/6160181/logs_61590.zip)
- [posix-cli-thread-cert.zip](https://github.com/openthread/openthread/files/6160182/posix-cli-thread-cert.zip)

PR #6309, opened 2021-03-18
[tests] add allowlist after reset to restore same network topology

brianljt:
This PR closes issue #6304


brianljt:
Reset command clears the allowlist. So ED might attach to the leader rather than the router after reset and cannot ping router's link local address. Need to add allowlist to ED after reset to make sure the desired network topology is resumed.

bukepo (approved):

Irving-cl (approved):
LGTM 👍

jwhui (approved):
Thanks! 👍🏼

PR #5504, opened 2020-09-09
[nrf528xx] check nrf_802154_sleep_if_idle() return status to set correct pending event

brianljt:

PR #6311, opened 2021-03-18
[low-power] adaptative CSL window duration

edmont:
This PR puts the CSL Receiver on the safe side for assuring frames reception over long synchronization periods, by means of:

- Make CSL receive window dependent on the accuracy of the crystals used for scheduling and on the elapsed time since last synchronization.
- Center the CSL receive window in the middle of the sample time.

Prior to this, CSL receive window duration had a fixed value and clock drifts would most likely make it happen outside of CSL transmission slot from the parent, specially after long periods without synchronization (as with the default 100 seconds CSL Timeout).

There is room for low power improvement by adding some intelligent drift corrections and vendor specific optimizations.


src/core/config/mac.h
```diff
@@ -399,14 +399,27 @@
...
+ *
+ * The total duration, in units of microseconds, required for the CSL receiver to fully receive and acknowledge a frame
+ * of maximum possible length.
+ * - Maximum frame size with preamble: 6*2+127*2 symbols
+ * - AIFS: 12 symbols
+ * - Maximum ACK size with preamble: 6*2+33*2 symbols
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_CSL_MIN_RECEIVE_ON
+#define OPENTHREAD_CONFIG_CSL_MIN_RECEIVE_ON 356 * 16
```
brianljt, line 420: I don't think we need to assume the worse case. We can keep the sample window just big enough to receive the preamble of packet. `otPlatRadioSleep` just let radio driver knows it is OK to sleep. But when the preamble of frame is received, and radio is in the middle of receiving a frame, it will wait until the end of the frame then goes to sleep.

We have made changes in PR #5449 and #5504 to make sure radio only sleep when it is truly idle and the sleep API will not break any ongoing transmission or reception(`nrf_802154_sleep` --> `nrf_802154_sleep_if_idle`).

src/core/mac/sub_mac.cpp
```diff
@@ -261,14 +262,21 @@ void SubMac::HandleReceiveDone(RxFrame *aFrame, Error aError)
         UpdateFrameCounter(aFrame->mInfo.mRxInfo.mAckFrameCounter);
     }
 
-#if OPENTHREAD_CONFIG_MAC_CSL_DEBUG_ENABLE
+#if OPENTHREAD_CONFIG_MAC_CSL_DEBUG_ENABLE || OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
     if (aFrame != nullptr && aError == kErrorNone)
     {
+        if ((mState == kStateCslSample) && aFrame->mInfo.mRxInfo.mAckedWithSecEnhAck)
+        {
+            mCslLastSync = TimerMicro::GetNow();
```
brianljt, line 270: How can we assume an Enh-ACK is really received by peer, thus making the assumption that CSL transmitter has been synchronized?

src/core/mac/sub_mac.cpp
```diff
@@ -487,7 +495,12 @@ void SubMac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, Error aErro
         {
             mCallbacks.RecordCcaStatus(ccaSuccess, aFrame.GetChannel());
         }
-
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+        if (mState == kStateCslSample)
+        {
+            mCslLastSync = TimerMicro::GetNow();
```
brianljt, line 501: Invalid state check. CSL receiver does not need to wait for `kStateCslSample` state to send packet to its parent. The router keeps its radio on and SSED can send data packet to it whenever it wants.

src/core/mac/sub_mac.cpp
```diff
@@ -934,6 +947,8 @@ void SubMac::SetCslPeriod(uint16_t aPeriod)
         mCslSampleTime = TimerMicro::GetNow();
         Get<Radio>().UpdateCslSampleTime(mCslSampleTime.GetValue());
 
+        mCslLastSync = mCslSampleTime;
```
brianljt, line 950: `mCslLastSync` should be set when the first packet with CSL IE include is sent(in `SubMac::HandleTransmitDone`). In this case, a child update request.

Ref: in `Mac::SetCslPeriod` --> `Get<Mle::Mle>().ScheduleChildUpdateRequest()`

src/core/mac/sub_mac.cpp
```diff
@@ -1004,6 +1020,27 @@ void SubMac::HandleCslTimer(void)
...
+    {
+        elapsedUs = UINT64_MAX - mCslLastSync.GetValue() + curTime;
+    }
+    else
+    {
+        elapsedUs = curTime - mCslLastSync.GetValue();
+    }
+
+    // TODO: Use CSL Accuracy TLV values from the parent
+    duration += elapsedUs * 2 * (kCslCrystalPpm + kCslWorstCrystalPpm) / 1000000;
```
brianljt, line 1040: Cap with some maximum value?

examples/platforms/nrf528xx/nrf52840/openthread-core-nrf52840-config.h
```diff
@@ -279,13 +279,13 @@
 #endif
 
 /**
- * @def OPENTHREAD_CONFIG_CSL_SAMPLE_WINDOW
+ * @def OPENTHREAD_CONFIG_CSL_CRYSTAL_PPM
```
brianljt, line 282: This macro is not needed as the value could be got from radio API `otPlatTimeGetXtalAccuracy`

brianljt (commented):
Thank you for helping with this enhancement. Here are some comments from me.

src/core/mac/sub_mac.hpp
```diff
@@ -553,14 +554,20 @@ class SubMac : public InstanceLocator, private NonCopyable
...
+        kMinCslWindow = OPENTHREAD_CONFIG_CSL_MIN_RECEIVE_ON, ///< CSL receive window for the longest possible
+                                                              ///< frame and ack duration.
         kCslReceiveTimeAhead =
             OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD, ///< CSL receivers would wake up `kCslReceiveTimeAhead` earlier
                                                       ///< than expected sample window. The time is in unit of 10
                                                       ///< symbols.
     };
 
+    enum : uint8_t{
+        kCslCrystalPpm = OPENTHREAD_CONFIG_CSL_CRYSTAL_PPM, ///< Accuracy of the crystal used to schedule CSL
```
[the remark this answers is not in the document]
edmont: Makes sense, however _The device's XTAL accuracy, in ppm_ is not clear for me in this situation. It might happen that for CSL operations a different precision is achieved. In any case I will use it now and add improvements when this changes in the future.

src/core/config/mac.h
hubertmis: > But when the preamble of frame is received, and radio is in the middle of receiving a frame, it will wait until the end of the frame then goes to sleep

`radio.h` does not define `otPlatRadioSleep()` like this. I think it is up to given platform port to abort receiving and enter sleep mode, or continue receiving until the frame is received. Until `otPlatRadioSleep()` clearly states that it shall not abort ongoing reception, I think it is needed to assume worst case.

brianljt: I agree. Looks like we have to update the API description in `radio.h` or create a new API called `otPlatRadioSleepIfIdle` for this purpose.

edmont: Yes, I was commenting the same, your assumptions @brianljt are only for nRF platform. This configuration option allows for every platform to adapt for it's requirements. Optimized values can be configured for different platforms.

brianljt: One more thing. With this worst case assumption, should `kMinCslWindow` be `OPENTHREAD_CONFIG_CSL_MIN_RECEIVE_ON * 2` as CSL phase is in the center of the sample window? This creates a minimum sample window bigger than 11ms which could make CSL less appealing than data poll regarding power consumption.

edmont: You are right about adding the _max_frame_ to the second half of the sample window.

Also about the low performance for the worst case, but in real products it can be greatly improved, specially if the device's temperature is taken into account for the crystal's PPM value.

src/core/mac/sub_mac.cpp
edmont: You are right, however I believe the risk of losing the Enh-ACK is pretty low compared to the power consumption advantage of keeping the receive window small. Specially taking into account that an SSED is likely to not send any frame to the parent apart from ACKs.

brianljt: Then I think it would be better to add some comments here.

Another thing is, I don't think this check is needed `mState == kStateCslSample` as OT stack might have already change to sleep state when the TxDone even is called. As long as the transmitted frame is acked with CSL IE included, we can consider CSL transmitter being synchronized. How do you think?

edmont: Maybe we should just check that `mCslPeriod > 0.`

src/core/mac/sub_mac.cpp
edmont: Now that I capped the maximum duration removing this initialization should be okay.

edmont:
I performed some testing after the changes, with two nRF52840 DKs and the following values:
```
kCslWorstCrystalPpm = 5
kCslWorstUncertainty = 2
XTAL_ACCURACY = 2
```
I'm sending 20 pings from the parent with a random delay of:
- Test A: 1-99 seconds
- Test B: 1-2 seconds
No ping failed or required retransmission.

[review_changes_pcaps.zip](https://github.com/openthread/openthread/files/6170716/review_changes_pcaps.zip)

edmont:
How to cope with the lack of `otPlatTimeGetXtalAccuracy` for simulation platform?

src/core/mac/sub_mac.cpp
```diff
@@ -1004,6 +1022,44 @@ void SubMac::HandleCslTimer(void)
...
+}
+
+uint32_t SubMac::GetCslSampleDuration(void)
+{
+    uint32_t periodDuration = mCslPeriod * kUsPerTenSymbols;
+    uint32_t duration = kMinCslWindow;
+
+    // TODO: Use CSL Accuracy TLV values from the parent
+    duration += kCslWorstUncertainty * 100;
+    duration += GetCslElapsedTime() * (otPlatTimeGetXtalAccuracy() / 2 +  kCslWorstCrystalPpm) / 1000000;
```
edmont, line 1059: Actually these two values added to the `duration` should be doubled.

edmont:
With the last change I added:
```
    uint8_t   mCslLocalDrift;   ///< Drift of timer used for scheduling CSL reception, in ± ppm.
    uint8_t   mCslParentDrift;  ///< Drift of timer used for scheduling CSL transmission by the parent, in ± ppm.
    uint8_t   mCslParentUncert; ///< Uncertainty of the scheduling CSL of transmission by the parent, in ±100 us units.
```
This allows for:
- Accounting for the real parent accuracy, to be updated from Parent Response or Child Update Response once https://threadgroup.atlassian.net/browse/SPEC-961 is in.
- Optimizing the local drift by applying temperature compensation, for instance. A new OT API will be required for this.

edmont:
Test result with:
```
        kCslLocalCrystalPpm = 2,
        kCslWorstCrystalPpm  = 2,
        kCslWorstUncertainty = 1,
```
[short-and-long-pings.pcap.zip](https://github.com/openthread/openthread/files/6172239/short-and-long-pings.pcap.zip)

src/core/mac/sub_mac.hpp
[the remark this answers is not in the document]
edmont: I replaced this value with `otPlatRadioGetCslAccuracy` that would allow for platforms to implement real time corrections on the accuracy.

hubertmis (approved):
Looks good!
Two potential improvements marked

examples/platforms/nrf528xx/nrf52840/openthread-core-nrf52840-config.h
```diff
@@ -279,24 +279,25 @@
 #endif
 
 /**
- * @def OPENTHREAD_CONFIG_CSL_SAMPLE_WINDOW
+ * @def OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD
  *
- * CSL sample window in units of 10 symbols.
+ * Reception scheduling and ramp up time needed for the receiver to be ready, in units of 10 symbols.
```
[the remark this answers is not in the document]
edmont: This was just keeping previous code base, which I guess it was based on the IEEE 802.15.4 CSL Period definition, also in units of 10 symbols. But I agree there is no reason not to have better granularity here. It would also save some multiplications in the code.

src/core/mac/sub_mac.cpp
```diff
@@ -261,14 +264,22 @@ void SubMac::HandleReceiveDone(RxFrame *aFrame, Error aError)
...
     }
 
-#if OPENTHREAD_CONFIG_MAC_CSL_DEBUG_ENABLE
+#if OPENTHREAD_CONFIG_MAC_CSL_DEBUG_ENABLE || OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
     if (aFrame != nullptr && aError == kErrorNone)
     {
+        // Assuming the risk of the parent missing the Enh-ACK in favor of smaller CSL receive window
+        if ((mCslPeriod > 0) && aFrame->mInfo.mRxInfo.mAckedWithSecEnhAck)
+        {
+            mCslLastSync = TimeMicro(static_cast<uint32_t>(otPlatRadioGetNow(&GetInstance())));
```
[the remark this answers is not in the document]
edmont: With ±20 ppm crystals in both parent and child, a 100 us error here has an impact of 8 us in the receive window duration. Do you think it's worth adding more code here to address it?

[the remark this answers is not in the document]
hubertmis: Not necessarily in this PR. However the problem we should address is undeterministic latency of execution from the main thread. This latency causes that timestamps or transmitted packets (captured here) may be incorrect. We address such undeterministic latency in other places with kCslTxTimeAhead, RxTimeAhead, etc.

We should think of a way to address this problem here as well. I think the cleanest solution would be to get precise TX timstamps from the 802.15.4 driver, but it does not expose such feature yet. Temporarily we could use another safety constant like kCslTxTimestampLatency and its value could be selected for given application depending on expected latencies in the main thread execution.

src/core/mac/sub_mac.cpp
```diff
@@ -487,7 +498,12 @@ void SubMac::HandleTransmitDone(TxFrame &aFrame, RxFrame *aAckFrame, Error aErro
         {
             mCallbacks.RecordCcaStatus(ccaSuccess, aFrame.GetChannel());
         }
-
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+        if (mCslPeriod > 0)
+        {
+            mCslLastSync = TimeMicro(static_cast<uint32_t>(aAckFrame->mInfo.mRxInfo.mTimestamp));
```
hubertmis, line 504: I think this timestamp is the timestamp of the end of received ACK, not the timestamp of the beginning of the frame being transmitted. To get correct number, you should subtract ACK duration, AIFS, and the transmitted frame duration.

src/core/mac/sub_mac.cpp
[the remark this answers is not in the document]
edmont: I changed it to use `mTimestamp`. For the RX path we should be on the safe side since the timestamp is prior to the sync Enh-ACK. For the TX path we get the timestamp of the received ACK, which is later than the sync TX frame, but always bounded.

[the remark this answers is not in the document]
hubertmis: I agree. I've already commented: https://github.com/openthread/openthread/pull/6311#discussion_r599713984

src/core/mac/sub_mac.cpp
edmont: Yes. Worst case should be: 2*(127+41)*16+192 = 5568 us. This imprecision will only add 1 us to the receive window after more than 200 s from last synchronization, assuming ±20 ppm crystals in both parent and child. Do we need extra arithmetic to cope with this small error?

hubertmis: It actually 2 * (127+41)***32**+192 = 10944 us. For worst case: 200s assuming RC +/- 500 PPM, we get 5.5 us error. I think we don't need to bother with such small error, because it is much smaller than scheduling granularity. However, I would add a comment that we intentionally omit this error.

edmont: Timestamp from `aAckFrame` can't be used since not all platforms provide it (including simulation). Getting back to current MAC time and adding a comment assuming the small error.

hubertmis: This assumption may be incorrect on some platforms. I'm fine to keep it right now to move forward.
But I would prefer to use radio timestamps on platforms where this feature is available, to avoid potential problems caused by software latency.

edmont:
I squashed the review comments and added a descriptive timing plot with the relevant points.

src/core/mac/sub_mac.cpp
```diff
@@ -1004,6 +1024,46 @@ void SubMac::HandleCslTimer(void)
         break;
     }
 }
+
+uint32_t SubMac::GetCslSafeSemiWindow(void)
```
brianljt, line 1028: Nit
```suggestion
uint32_t SubMac::GetCslDriftSemiWindow(void)
```

src/core/mac/sub_mac.cpp
```diff
@@ -985,14 +1005,14 @@ void SubMac::HandleCslTimer(void)
         mCslState = kCslSample;
 
         mCslSampleTime += mCslPeriod * kUsPerTenSymbols;
-        Get<Radio>().UpdateCslSampleTime(mCslSampleTime.GetValue());
-
-        mCslTimer.StartAt(mCslSampleTime, kCslSampleWindow);
+        duration = GetCslSampleDuration();
+        mCslTimer.StartAt(mCslSampleTime, duration);
```
brianljt, line 1009: As `mCslSampleTime` is at the center of CSL sample window, the end of sample window should be sampleTime + drift + uncertainty + `kMinCslWindow`. Not need to X2 in `GetCslSampleDuration` for drift and `mCslParentUncert`?

src/core/mac/sub_mac.cpp
```diff
@@ -1004,6 +1024,46 @@ void SubMac::HandleCslTimer(void)
...
+    }
+    else
+    {
+        semiWindow = static_cast<uint32_t>(curTime - mCslLastSync.GetValue());
+    }
+
+    semiWindow = semiWindow * (Get<Radio>().GetCslAccuracy() + mCslParentDrift) / 1000000;
+
+    // Cap the CSL receive window duration to three quarters of the CSL Period
+    return (semiWindow > threeQuartersSemiPeriod) ? threeQuartersSemiPeriod : semiWindow;
```
brianljt, line 1046: Move the cap to `GetCslSampleStart` and `GetCslSampleEnd`? Thoughts?

src/core/mac/sub_mac.cpp
```diff
@@ -1004,6 +1024,46 @@ void SubMac::HandleCslTimer(void)
...
+}
+
+uint32_t SubMac::GetCslSampleStart(void)
+{
+    uint32_t cslPeriodUs = mCslPeriod * kUsPerTenSymbols;
+    uint32_t timeAhead   = kCslReceiveTimeAhead;
+
+    timeAhead += mCslParentUncert * 100;
+    timeAhead += GetCslSafeSemiWindow();
+    return (timeAhead > cslPeriodUs) ? 0 : (cslPeriodUs - timeAhead);
```
brianljt, line 1056: Something like,
```suggestion
    return (timeAhead > cslPeriodUs / 2) ? (cslPeriodUs / 2) : (cslPeriodUs - timeAhead);
```

src/core/mac/sub_mac.cpp
```diff
@@ -1004,6 +1024,46 @@ void SubMac::HandleCslTimer(void)
...
+}
+
+uint32_t SubMac::GetCslSampleDuration(void)
+{
+    uint32_t duration = kMinCslWindow;
+
+    duration += 2 * mCslParentUncert * 100;
+    duration += 2 * GetCslSafeSemiWindow();
+    return duration;
+}
```
brianljt, line 1066:
```suggestion
uint32_t SubMac::GetCslSampleEnd(void)
{
    uint32_t timeEnd = kMinCslWindow;

    timeEnd += mCslParentUncert * 100;
    timeEnd += GetCslSafeSemiWindow();
    return (timeEnd > cslPeriodUs / 2) ? (cslPeriodUs / 2) : (cslPeriodUs - timeEnd);
}
```

brianljt (commented):
Overall looks really good. Thanks!

Some comments on the end time of the sample window.

src/core/mac/sub_mac.cpp
```diff
@@ -966,12 +984,14 @@ void SubMac::HandleCslTimer(Timer &aTimer)
 
 void SubMac::HandleCslTimer(void)
 {
+    uint32_t duration;
+
```
brianljt, line 988: Maybe we can put a diagram here for better understanding

```suggestion
    /*
     * CSL sample timing illustration
     * Maybe we can put a diagram here for better understanding

```suggestion
    /*
     * CSL sample timing illustration
     *      |<-------------------------------------Wake----------------------------------------->|<------------Sleep------------->|
     *      |                                                                                    |                                |
     *      |<--Ahead-->|<--UnCert-->|<--ClockDrift-->|<--ClockDrift-->|<--UnCert-->|<--MinWin-->|                                |
     *      |           |            |                |                |            |            |                                |
     * -----|-----------|------------|----------------|----------------|------------|------------|---------------//---------------|------
     * SampleStart                                 CslPhase                                  SampleEnd                        SampleStart
     */
    uint32_t duration;

```
     */
    uint32_t duration;

```

**Review by brianljt — APPROVED** [review]
LGTM. Thanks! 👍

File: src/core/mac/sub_mac.cpp [hunk 18]
```diff
@@ -261,14 +264,22 @@ void SubMac::HandleReceiveDone(RxFrame *aFrame, Error aError)
         UpdateFrameCounter(aFrame->mInfo.mRxInfo.mAckFrameCounter);
     }
 
-#if OPENTHREAD_CONFIG_MAC_CSL_DEBUG_ENABLE
+#if OPENTHREAD_CONFIG_MAC_CSL_DEBUG_ENABLE || OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
```
Irving-cl, line 267:
```suggestion
#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
```
Note that `OPENTHREAD_CONFIG_MAC_CSL_DEBUG_ENABLE` can only be enabled when `OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE`. We have a compile-time check for ensuring this. So we can just use `OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE` here.

src/core/mac/sub_mac.cpp
```diff
@@ -1004,6 +1034,28 @@ void SubMac::HandleCslTimer(void)
...
+    {
+        elapsed = static_cast<uint32_t>(UINT64_MAX - mCslLastSync.GetValue() + curTime);
+    }
+    else
+    {
+        elapsed = static_cast<uint32_t>(curTime - mCslLastSync.GetValue());
+    }
+
+    semiWindow = elapsed * (Get<Radio>().GetCslAccuracy() + mCslParentDrift) / 1000000;
+    semiWindow += mCslParentUncert * 100;
```
Irving-cl, line 1054: Why is `mCslParentUncert`'s unit 100us? Can we add a const enum for it? For example, `kUsPerUncertUnit`.

src/core/mac/sub_mac.cpp
```diff
@@ -1004,6 +1034,28 @@ void SubMac::HandleCslTimer(void)
         break;
     }
 }
+
+void SubMac::GetCslWindowEdges(uint32_t *ahead, uint32_t *after)
```
Irving-cl, line 1038: Use reference for the parameters instead?

Irving-cl (approved):
LGTM :+1: 

Some suggestions.

src/core/mac/sub_mac.cpp
edmont: Currently 100 us is agreed, but I'm proposing to use units of 10 us: https://threadgroup.atlassian.net/browse/SPEC-961?focusedCommentId=49637

edmont: BTW, it would be a good exercise to identify the required uncertainty value in OT when transmit_at is not used.

hubertmis: > it would be a good exercise to identify the required uncertainty value in OT

Uncertainty depends on application. You cannot identify constant for a network stack. The easiest way would be to add public API to OT so that application enabling Thread 1.2 router would provide this value. Alternatively, OT could try to measure uncertainty, and select probable "worst case" using statistical analysis.

jwhui (approved):
Thanks! 👍🏼

PR #6313, opened 2021-03-19
[border-router] add validation of updated on mesh prefixes

XPAWAN:
This commit adds validation of the on mesh prefixes.

    With this change, the domain on mesh prefix can be directly configured after successfull verification.

### DIFF
--- a/src/core/api/border_router_api.cpp
+++ b/src/core/api/border_router_api.cpp
@@ -80,18 +80,18 @@ otError otBorderRouterAddOnMeshPrefix(otInstance *aInstance, const otBorderRoute
 
     OT_ASSERT(aConfig != nullptr);
 
-    SuccessOrExit(error = instance.Get<NetworkData::Local>().AddOnMeshPrefix(*config));
+    SuccessOrExit(error = instance.Get<NetworkData::Local>().ValidateOnMeshPrefix(*config));
 
 #if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
-    // Only try to configure Domain Prefix after the parameter is validated via above `AddOnMeshPrefix()`.
     if (aConfig->mDp)
     {
-        // Restore local server data
-        IgnoreError(instance.Get<NetworkData::Local>().RemoveOnMeshPrefix(config->GetPrefix()));
-
         instance.Get<BackboneRouter::Local>().SetDomainPrefix(*config);
     }
+    else
 #endif
+    {
+        SuccessOrExit(error = instance.Get<NetworkData::Local>().AddOnMeshPrefix(*config));
+    }
 
 exit:
     return error;
--- a/src/core/api/border_router_api.cpp
+++ b/src/core/api/border_router_api.cpp
@@ -80,11 +80,10 @@ otError otBorderRouterAddOnMeshPrefix(otInstance *aInstance, const otBorderRoute
 
     OT_ASSERT(aConfig != nullptr);
 
-    SuccessOrExit(error = instance.Get<NetworkData::Local>().ValidateOnMeshPrefix(*config));
-
 #if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
     if (aConfig->mDp)
     {
+        SuccessOrExit(error = instance.Get<NetworkData::Local>().ValidateOnMeshPrefix(*config));
         instance.Get<BackboneRouter::Local>().SetDomainPrefix(*config);
     }
     else
--- a/src/core/thread/network_data_local.cpp
+++ b/src/core/thread/network_data_local.cpp
@@ -54,28 +54,6 @@ Local::Local(Instance &aInstance)
 }
 
 #if OPENTHREAD_CONFIG_BORDER_ROUTER_ENABLE
-Error Local::ValidatePrefix(const Ip6::Prefix &aPrefix)
-{
-    return (aPrefix.GetLength() > 0) && aPrefix.IsValid() ? kErrorNone : kErrorInvalidArgs;
-}
-
-Error Local::ValidatePreference(int8_t aPrf)
-{
-    otError error = kErrorNone;
-
-    switch (aPrf)
-    {
-    case OT_ROUTE_PREFERENCE_LOW:
-    case OT_ROUTE_PREFERENCE_MED:
-    case OT_ROUTE_PREFERENCE_HIGH:
-        break;
-    default:
-        error = kErrorInvalidArgs;
-    }
-
-    return error;
-}
-
 Error Local::ValidateOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
 {
     Error error = kErrorNone;
@@ -90,12 +68,7 @@ Error Local::ValidateOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
     // of an IEEE 802.15.4 interface MUST have a length of 64 bits.
     VerifyOrExit(!aConfig.mSlaac || aConfig.mPrefix.mLength == OT_IP6_PREFIX_BITSIZE, error = kErrorInvalidArgs);
 
-    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
-    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
-
-    // Check if the prefix overlaps mesh-local prefix.
-    VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()),
-                 error = kErrorInvalidArgs);
+    SuccessOrExit(error = ValidatePrefixAndPreference(aConfig.GetPrefix(), aConfig.mPreference));
 
 exit:
     return error;
@@ -104,6 +77,9 @@ exit:
 Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
 {
     uint16_t flags = 0;
+    Error    error = kErrorNone;
+
+    SuccessOrExit(error = ValidateOnMeshPrefix(aConfig));
 
     if (aConfig.mPreferred)
     {
@@ -147,8 +123,11 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
     }
 #endif
 
-    return AddPrefix(aConfig.GetPrefix(), NetworkDataTlv::kTypeBorderRouter, aConfig.mPreference, flags,
-                     aConfig.mStable);
+    error =
+        AddPrefix(aConfig.GetPrefix(), NetworkDataTlv::kTypeBorderRouter, aConfig.mPreference, flags, aConfig.mStable);
+
+exit:
+    return error;
 }
 
 Error Local::RemoveOnMeshPrefix(const Ip6::Prefix &aPrefix)
@@ -156,32 +135,31 @@ Error Local::RemoveOnMeshPrefix(const Ip6::Prefix &aPrefix)
     return RemovePrefix(aPrefix, NetworkDataTlv::kTypeBorderRouter);
 }
 
-Error Local::ValidateHasRoutePrefix(const ExternalRouteConfig &aConfig)
+Error Local::ValidatePrefixAndPreference(const Ip6::Prefix &aPrefix, int8_t aPrf)
 {
     Error error = kErrorNone;
 
-    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
-    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
+    VerifyOrExit((aPrefix.GetLength() > 0) && aPrefix.IsValid(), error = kErrorInvalidArgs);
+
+    switch (aPrf)
+    {
+    case OT_ROUTE_PREFERENCE_LOW:
+    case OT_ROUTE_PREFERENCE_MED:
+    case OT_ROUTE_PREFERENCE_HIGH:
+        break;
+    default:
+        ExitNow(error = kErrorInvalidArgs);
+    }
 
     // Check if the prefix overlaps mesh-local prefix.
-    VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()),
-                 error = kErrorInvalidArgs);
-
+    VerifyOrExit(!aPrefix.ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()), error = kErrorInvalidArgs);
 exit:
     return error;
 }
 
 Error Local::AddHasRoutePrefix(const ExternalRouteConfig &aConfig)
 {
-    Error error = kErrorNone;
-
-    SuccessOrExit(error = ValidateHasRoutePrefix(aConfig));
-
-    SuccessOrExit(
-        error = AddPrefix(aConfig.GetPrefix(), NetworkDataTlv::kTypeHasRoute, aConfig.mPreference, 0, aConfig.mStable));
-
-exit:
-    return error;
+    return AddPrefix(aConfig.GetPrefix(), NetworkDataTlv::kTypeHasRoute, aConfig.mPreference, 0, aConfig.mStable);
 }
 
 Error Local::RemoveHasRoutePrefix(const Ip6::Prefix &aPrefix)
@@ -198,6 +176,8 @@ Error Local::AddPrefix(const Ip6::Prefix &  aPrefix,
     Error      error = kErrorNone;
     uint8_t    subTlvLength;
     PrefixTlv *prefixTlv;
+
+    SuccessOrExit(error = ValidatePrefixAndPreference(aPrefix, aPrf));
 
     IgnoreError(RemovePrefix(aPrefix, aSubTlvType));
 
--- a/src/core/thread/network_data_local.cpp
+++ b/src/core/thread/network_data_local.cpp
@@ -159,7 +159,8 @@ exit:
 
 Error Local::AddHasRoutePrefix(const ExternalRouteConfig &aConfig)
 {
-    return AddPrefix(aConfig.GetPrefix(), NetworkDataTlv::kTypeHasRoute, aConfig.mPreference, 0, aConfig.mStable);
+    return AddPrefix(aConfig.GetPrefix(), NetworkDataTlv::kTypeHasRoute, aConfig.mPreference, /* aFlags */ 0,
+                     aConfig.mStable);
 }
 
 Error Local::RemoveHasRoutePrefix(const Ip6::Prefix &aPrefix)
--- a/src/core/thread/network_data_local.cpp
+++ b/src/core/thread/network_data_local.cpp
@@ -54,10 +54,31 @@ Local::Local(Instance &aInstance)
 }
 
 #if OPENTHREAD_CONFIG_BORDER_ROUTER_ENABLE
-Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
+Error Local::ValidatePrefix(const Ip6::Prefix &aPrefix)
 {
-    Error    error;
-    uint16_t flags = 0;
+    return (aPrefix.GetLength() > 0) && aPrefix.IsValid() ? kErrorNone : kErrorInvalidArgs;
+}
+
+Error Local::ValidatePreference(int8_t aPrf)
+{
+    otError error = kErrorNone;
+
+    switch (aPrf)
+    {
+    case OT_ROUTE_PREFERENCE_LOW:
+    case OT_ROUTE_PREFERENCE_MED:
+    case OT_ROUTE_PREFERENCE_HIGH:
+        break;
+    default:
+        error = kErrorInvalidArgs;
+    }
+
+    return error;
+}
+
+Error Local::ValidateOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
+{
+    Error error = kErrorNone;
 
     // Add Prefix validation check:
     // Thread 1.1 Specification 5.13.2 says
@@ -68,6 +89,21 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
     // An IPv6 address prefix used for stateless autoconfiguration [RFC4862]
     // of an IEEE 802.15.4 interface MUST have a length of 64 bits.
     VerifyOrExit(!aConfig.mSlaac || aConfig.mPrefix.mLength == OT_IP6_PREFIX_BITSIZE, error = kErrorInvalidArgs);
+
+    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
+    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
+
+    // Check if the prefix overlaps mesh-local prefix.
+    VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()),
+                 error = kErrorInvalidArgs);
+
+exit:
+    return error;
+}
+
+Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
+{
+    uint16_t flags = 0;
 
     if (aConfig.mPreferred)
     {
@@ -111,11 +147,8 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
     }
 #endif
 
-    error =
-        AddPrefix(aConfig.GetPrefix(), NetworkDataTlv::kTypeBorderRouter, aConfig.mPreference, flags, aConfig.mStable);
-
-exit:
-    return error;
+    return AddPrefix(aConfig.GetPrefix(), NetworkDataTlv::kTypeBorderRouter, aConfig.mPreference, flags,
+                     aConfig.mStable);
 }
 
 Error Local::RemoveOnMeshPrefix(const Ip6::Prefix &aPrefix)
@@ -123,10 +156,32 @@ Error Local::RemoveOnMeshPrefix(const Ip6::Prefix &aPrefix)
     return RemovePrefix(aPrefix, NetworkDataTlv::kTypeBorderRouter);
 }
 
+Error Local::ValidateHasRoutePrefix(const ExternalRouteConfig &aConfig)
+{
+    Error error = kErrorNone;
+
+    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
+    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
+
+    // Check if the prefix overlaps mesh-local prefix.
+    VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()),
+                 error = kErrorInvalidArgs);
+
+exit:
+    return error;
+}
+
 Error Local::AddHasRoutePrefix(const ExternalRouteConfig &aConfig)
 {
-    return AddPrefix(aConfig.GetPrefix(), NetworkDataTlv::kTypeHasRoute, aConfig.mPreference, /* aFlags */ 0,
-                     aConfig.mStable);
+    Error error = kErrorNone;
+
+    SuccessOrExit(error = ValidateHasRoutePrefix(aConfig));
+
+    SuccessOrExit(
+        error = AddPrefix(aConfig.GetPrefix(), NetworkDataTlv::kTypeHasRoute, aConfig.mPreference, 0, aConfig.mStable));
+
+exit:
+    return error;
 }
 
 Error Local::RemoveHasRoutePrefix(const Ip6::Prefix &aPrefix)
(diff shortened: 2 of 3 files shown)


src/core/thread/network_data_local.hpp
```diff
@@ -71,6 +71,17 @@ class Local : public NetworkData, private NonCopyable
     explicit Local(Instance &aInstance);
 
 #if OPENTHREAD_CONFIG_BORDER_ROUTER_ENABLE
+    /**
+     * This method validate on mesh prefix.
```
wgtdkp, line 75:
```suggestion
     * This method validates an on-mesh prefix.
```

src/core/thread/network_data_local.hpp
```diff
@@ -71,6 +71,17 @@ class Local : public NetworkData, private NonCopyable
...
+     *
+     * @param[in]  aConfig  A reference to the on mesh perfix configuration.
+     *
+     * @retval kErrorNone         Successfully verified the Border Router entry.
```
wgtdkp, line 79:
```suggestion
     * @retval kErrorNone         Successfully verified the on-mesh prefix.
```

src/core/thread/network_data_local.hpp
```diff
@@ -71,6 +71,17 @@ class Local : public NetworkData, private NonCopyable
     explicit Local(Instance &aInstance);
 
 #if OPENTHREAD_CONFIG_BORDER_ROUTER_ENABLE
+    /**
+     * This method validate on mesh prefix.
+     *
+     * @param[in]  aConfig  A reference to the on mesh perfix configuration.
```
wgtdkp, line 77:
```suggestion
     * @param[in]  aConfig  A reference to the on-mesh perfix configuration.
```

src/core/thread/network_data_local.cpp
```diff
@@ -69,6 +68,14 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
     // of an IEEE 802.15.4 interface MUST have a length of 64 bits.
     VerifyOrExit(!aConfig.mSlaac || aConfig.mPrefix.mLength == OT_IP6_PREFIX_BITSIZE, error = kErrorInvalidArgs);
 
+exit:
```
simonlingoogle, line 71: Should we validate more:

- VerifyOrExit((aPrefix.GetLength() > 0) && aPrefix.IsValid(), error = kErrorInvalidArgs);
- Verify `Preference` is valid
- Verify Prefix contains Mesh-local Prefix
- Verify Network Data contains enough space

I think the original code intended to make sure that the prefix can be added (enough space in Network Data). 

@librasungirl

src/core/thread/network_data_local.hpp
```diff
@@ -71,6 +71,17 @@ class Local : public NetworkData, private NonCopyable
...
+     *
+     * @retval kErrorNone         Successfully verified the Border Router entry.
+     * @retval kErrorInvalidArgs  The prefix is not valid on mesh prefix.
```
wgtdkp, line 80:
```suggestion
     * @retval kErrorInvalidArgs  The prefix is not a valid on-mesh prefix.
```

wgtdkp (approved):
Thanks @XPAWAN LGTM :+1:

XPAWAN:
@simonlingoogle ,  @wgtdkp all remarks has been applied. Can You please review PR?

wgtdkp:
> @simonlingoogle , @wgtdkp all remarks has been applied. Can You please review PR?

thanks @XPAWAN! I have approved this PR!

src/core/thread/network_data_local.cpp
```diff
@@ -69,6 +68,44 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+        default:
+            ExitNow(error = kErrorInvalidArgs);
+    }
+
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+    if (aConfig.mDp)
+    {
+        // Check if domain prefix is already configured
+        OnMeshPrefixConfig prefixConfig;
+        VerifyOrExit(Get<BackboneRouter::Local>().GetDomainPrefix(prefixConfig) == kErrorNotFound, error = kErrorAlready);
```
simonlingoogle, line 90: I think we should allow adding domain prefix when there is already one.

This is allowed in the original code:
```
node 1> prefix
2001:0:0:0::/64 parosD med 8c00
Done
node 1> prefix add 2002::/64 parosD
Done
node 1> prefix
2002:0:0:0::/64 parosD med fffe
Done
```

This would make buffer size checking for domain prefix more complex than a common prefix because the old domain prefix will be removed, thus leaving more space for the new domain prefix.

src/core/thread/network_data_local.cpp
```diff
@@ -69,6 +68,44 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+        VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()), error = kErrorAlready);
+        // Verify network data contains enough space
+        VerifyOrExit(CheckPrefixBufferSize(aConfig.GetPrefix()) == kErrorNone, error = kErrorInvalidArgs);
+    }
+
+exit:
+    return error;
+}
+
+Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
```
simonlingoogle, line 105: Since `AddOnMeshPrefix` is also called by other modules besides `otBorderRouterAddOnMeshPrefix`, we should call `ValidateOnMeshPrefix` in `AddOnMeshPrefix` again to make sure the prefix is valid?

src/core/thread/network_data_local.hpp
```diff
@@ -179,6 +191,7 @@ class Local : public NetworkData, private NonCopyable
     void  UpdateRloc(PrefixTlv &aPrefixTlv);
     bool  IsOnMeshPrefixConsistent(void) const;
     bool  IsExternalRouteConsistent(void) const;
+    Error CheckPrefixBufferSize(const Ip6::Prefix &  aPrefix);
```
simonlingoogle, line 194:
```suggestion
    Error CheckBufferSizeForPrefix(const Ip6::Prefix &  aPrefix);
```
nit

src/core/thread/network_data_local.cpp
XPAWAN: @simonlingoogle There can be configured only one domain on-mesh prefix at a time. I have added a similar functionality to wpantund (https://github.com/openthread/wpantund/pull/489/files) to prohibit setting a new on-mesh domain prefix if there is already one. To add a new on-mesh domain prefix the old one must be explicitly removed. Which approach should I follow ?

src/core/thread/network_data_local.cpp
XPAWAN: @simonlingoogle right. I have added external validation.

src/core/thread/network_data_local.cpp
```diff
@@ -69,6 +68,47 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+    default:
+        ExitNow(error = kErrorInvalidArgs);
+    }
+
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+    if (aConfig.mDp)
+    {
+        // Check if domain prefix is already configured
+        OnMeshPrefixConfig prefixConfig;
+        VerifyOrExit(Get<BackboneRouter::Local>().GetDomainPrefix(prefixConfig) == kErrorNotFound,
```
simonlingoogle, line 90: Do we need to `CheckBufferSizeForPrefix` for the domain prefix too?

src/core/thread/network_data_local.cpp
```diff
@@ -111,29 +151,74 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+    case OT_ROUTE_PREFERENCE_MED:
+    case OT_ROUTE_PREFERENCE_HIGH:
+        break;
+    default:
+        ExitNow(error = kErrorInvalidArgs);
+    }
+
+    // Check if address already configured
+    VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()),
+                 error = kErrorAlready);
```
simonlingoogle, line 183: This part of check is common for both `ValidateOnMeshPrefix` and  `ValidateHasRoutePrefix`. (used to be in the `AddPrefix` method). 

Should use a same method (e.g. `ValidatePrefix`) for these logic to reduce code size?

src/core/thread/network_data_local.cpp
```diff
@@ -111,29 +151,74 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
 
+Error Local::CheckBufferSizeForPrefix(const Ip6::Prefix &aPrefix, NetworkDataTlv::Type aSubTlvType)
+{
+    uint8_t subTlvLength;
+
+    subTlvLength = (aSubTlvType == NetworkDataTlv::kTypeBorderRouter)
+                       ? sizeof(BorderRouterTlv) + sizeof(BorderRouterEntry)
+                       : sizeof(HasRouteTlv) + sizeof(HasRouteEntry);
+
+    return CanInsert(sizeof(PrefixTlv) + aPrefix.GetBytesSize() + subTlvLength) ? kErrorNone : kErrorNoBufs;
```
simonlingoogle, line 219: `Local::AddPrefix` actually always remove the same prefix (if exist) before adding. So, we should consider that for calculating the space required?

src/core/thread/network_data_local.cpp
simonlingoogle, line 183:
```suggestion
                 error = kErrorInvalidArgs);
```

I think this check is to make sure that the prefix does not contains the mesh local prefix. It does not mean that the prefix already exists.

src/core/thread/network_data_local.cpp
simonlingoogle: There are calls to `AddOnMeshPrefix` that bypasses `ValidateOnMeshPrefix` (e.g. in routing_manager.cpp).

Should we just call `ValidateOnMeshPrefix` again in `AddOnMeshPrefix`? `otBorderRouterAddOnMeshPrefix` would then call `ValidateOnMeshPrefix` twice but that does not seems to be a real issue anyway.

src/core/thread/network_data_local.cpp
XPAWAN: There can be only one domain prefix set to mDomainPrefixConfig in bbr_local. I think we need to check only if there was a previous one configured. The question is if we need to check if domain prefix is already configured to prohibit overwritting or allow to set domain prefix without removing the previous entry ?

XPAWAN:
@simonlingoogle , @wgtdkp . All remarks has been applied and the PR has been rebased. Can You verify this PR again?

src/core/thread/network_data_local.cpp
```diff
@@ -111,29 +166,94 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+    subTlvLength = (aSubTlvType == NetworkDataTlv::kTypeBorderRouter)
+                       ? sizeof(BorderRouterTlv) + sizeof(BorderRouterEntry)
+                       : sizeof(HasRouteTlv) + sizeof(HasRouteEntry);
+
+    subTlvLength += sizeof(PrefixTlv) + aPrefix.GetBytesSize();
+
+    // Check if prefix already exist.
+    // If entry exist add additional space after prefix removal
+    // to buffer size calculation.
+    if ((tlv = FindPrefix(aPrefix))) != nullptr)
```
wgtdkp, line 234:
```suggestion
    if ((tlv = FindPrefix(aPrefix)) != nullptr)
```

src/core/thread/network_data_local.cpp
```diff
@@ -111,29 +166,94 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
 
+Error Local::ValidateHasRoutePrefix(const ExternalRouteConfig &aConfig)
+{
+    Error error = kErrorNone;
+
+    // Validate Prefix
+    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
+
+    // Validate Preference
+    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
```
wgtdkp, line 186: nit:
```suggestion
    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
```

src/core/thread/network_data_local.cpp
```diff
@@ -111,29 +166,94 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+
+    // Check if address already configured
+    VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()),
+                 error = kErrorInvalidArgs);
```
wgtdkp, line 190:
```suggestion
    // Check if the prefix overlaps mesh-local prefix.
    VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()),
                 error = kErrorInvalidArgs);
```

src/core/thread/network_data_local.cpp
```diff
@@ -111,29 +166,94 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+
+    // Verify network data contains enough space
```
wgtdkp, line 192: nit:
```suggestion
    // Verify network data contains enough space.
```

src/core/thread/network_data_local.cpp
```diff
@@ -69,6 +90,40 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
     // of an IEEE 802.15.4 interface MUST have a length of 64 bits.
     VerifyOrExit(!aConfig.mSlaac || aConfig.mPrefix.mLength == OT_IP6_PREFIX_BITSIZE, error = kErrorInvalidArgs);
 
+    // Validate Prefix
+    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
+
+    // Validate Preference
+    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
```
wgtdkp, line 97:
```suggestion
    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
```

src/core/thread/network_data_local.cpp
```diff
@@ -69,6 +90,40 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+    {
+        // Check if domain prefix is already configured
+        OnMeshPrefixConfig prefixConfig;
+        VerifyOrExit(Get<BackboneRouter::Local>().GetDomainPrefix(prefixConfig) == kErrorNotFound,
+                     error = kErrorAlready);
+    }
+    else
+#endif
+    {
+        // Check if address already configured
```
wgtdkp, line 110: nit:
```suggestion
        // Check if the prefix overlaps mesh-local prefix.
```

src/core/thread/network_data_local.cpp
```diff
@@ -69,6 +90,40 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+        VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()),
+                     error = kErrorInvalidArgs);
+
+        // Verify network data contains enough space
```
wgtdkp, line 114: nit:
```suggestion
        // Verify network data contains enough space.
```

src/core/thread/network_data_local.cpp
```diff
@@ -111,29 +166,94 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
 {
     return RemovePrefix(aPrefix, NetworkDataTlv::kTypeHasRoute);
 }
 
+Error Local::CheckBufferSizeForPrefix(const Ip6::Prefix &aPrefix, NetworkDataTlv::Type aSubTlvType)
+{
+    uint8_t subTlvLength;
+    PrefixTlv *tlv;
+    uint8_t removedTlvLength = 0;
+    Error error = kErrorNone;
```
wgtdkp, line 223: Seems like you forgot to make the code pretty, try running:
```shell
./script/make-pretty clang-format
```

XPAWAN:
All remarks has been applied.

abtink (commented):
Thanks @XPAWAN for helping with this.
Some suggestions/questions below. Thanks.

src/core/api/border_router_api.cpp
```diff
@@ -80,18 +80,18 @@ otError otBorderRouterAddOnMeshPrefix(otInstance *aInstance, const otBorderRoute
 
     OT_ASSERT(aConfig != nullptr);
 
-    SuccessOrExit(error = instance.Get<NetworkData::Local>().AddOnMeshPrefix(*config));
+    SuccessOrExit(error = instance.Get<NetworkData::Local>().ValidateOnMeshPrefix(*config));
```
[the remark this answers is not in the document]
simonlingoogle: I would suggest exactly the same and the method is precisely `AddOnMeshPrefix`. I think we should move `Domain Prefix` specific logic to `AddOnMeshPrefix`, instead of the current strategy of calling `AddOnMeshPrefix` in `SetDomainPrefix. 

That being said, I am fine to enhance it in a separate PR since this is not an issue that introduced by this PR.

src/core/thread/network_data_local.cpp
simonlingoogle: `NetworkData::Local::AddPrefix` always calls `RemovePrefix` to remove the same prefix from Network Data before adding it. So, I think we should keep this behavior for Domain Prefix unless there is a good reason not to. 

Thoughts? @XPAWAN @abtink @librasungirl

src/core/thread/network_data_local.cpp
```diff
@@ -111,29 +163,91 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
+    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
+
+    // Check if the prefix overlaps mesh-local prefix.
+    VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()),
+                 error = kErrorInvalidArgs);
+
+    // Verify network data contains enough space.
+    VerifyOrExit(CheckBufferSizeForPrefix(aConfig.GetPrefix(), NetworkDataTlv::kTypeHasRoute) == kErrorNone,
+                 error = kErrorInvalidArgs);
```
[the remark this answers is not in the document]
simonlingoogle: When BBR is not enabled, the domain prefix is not actually added to the `Network Data`. That is why we called `AddOnMeshPrefix` and `RemoveOnMeshPrefix` for `Domain Prefix` just to check if the prefix is valid. 

So, it seems unnecessary to check for buffer size for because:
1. If the BBR is not enabled, setting domain prefix consumes no space in `Network Data`
2. If the BBR is enabled, `AppendTlv` already checked the `Network Data` space

So, I think we should remove calling `CheckBufferSizeForPrefix` in both `ValidateOnMeshPrefix` and `ValidateOnMeshPrefix`.

I am sorry for the confusion here. It has been quite a while since this code was written and the considerations back then is  unclear by now.

Thoughts? @abtink @XPAWAN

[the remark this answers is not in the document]
XPAWAN: `CheckBufferSizeForPrefix()` has been removed.

src/core/thread/network_data_local.hpp
```diff
@@ -176,6 +200,8 @@ class Local : public NetworkData, private NonCopyable
                     uint16_t             aFlags,
                     bool                 aStable);
     Error RemovePrefix(const Ip6::Prefix &aPrefix, NetworkDataTlv::Type aSubTlvType);
+    Error ValidatePrefix(const Ip6::Prefix &aPrefix);
+    Error ValidatePreference(int8_t aPrf);
```
simonlingoogle, line 204: nit: these methods can be static

src/core/thread/network_data_local.hpp
```diff
@@ -105,6 +117,18 @@ class Local : public NetworkData, private NonCopyable
      */
     Error AddHasRoutePrefix(const ExternalRouteConfig &aConfig);
 
+    /**
+     * This method validates Has Route prefix for Thread Network data.
+     *
+     * @param[in]  aConfig        A reference to the external route configuration.
+     *
+     * @retval kErrorNone         Successfully verified the Has Route prefix.
+     * @retval kErrorAlready      The prefix is already configured.
```
simonlingoogle, line 126:
```suggestion
```
`ValidateHasRoutePrefix` doesn't return `kErrAlready`

src/core/thread/network_data_local.cpp
```diff
@@ -69,6 +90,30 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
+
+#if OPENTHREAD_FTD && OPENTHREAD_CONFIG_BACKBONE_ROUTER_ENABLE
+    if (aConfig.mDp)
+    {
+        // Check if domain prefix is already configured
+        OnMeshPrefixConfig prefixConfig;
+        ExitNow(error = ((Get<BackboneRouter::Local>().GetDomainPrefix(prefixConfig) == kErrorNotFound) ? kErrorNone : kErrorAlready);
+    }
+#endif
```
simonlingoogle, line 103: I think it's fine to add a domain prefix when there is already one configured (same as before).

XPAWAN: @simonlingoogle , @librasungirl, @wgtdkp @abtink  I have removed checking if domian prefix is already configured. The PR has been rebased. I think the wpantund should also allow to add domain prefix if the one is already configured. Currently the old one must be removed to add a new one.

src/core/thread/network_data_local.cpp
librasungirl: A little bit different from the other on-mesh prefixes, domain prefix is unique and designed to allow in-place change directly, meanwhile allow removing and then adding like the other prefixes.

One note is the domain prefix may only exist in bbr local, and not in local network data until bbr is enabled (currently bbr is enabled by default). If somehow bbr is disabled, you may not be able to check the existence of domain prefix configured via local network data.

src/core/thread/network_data_local.cpp
librasungirl: > There can be only one domain prefix set to mDomainPrefixConfig in bbr_local. I think we need to check only if there was a previous one configured. The question is if we need to check if domain prefix is already configured to prohibit overwritting or allow to set domain prefix without removing the previous entry ?

See above comment, domain prefix is designed to allow in-place change, without requiring clearing previous config.

librasungirl (commented):
> With this change, the domain on mesh prefix can be directly configured after successfull verification.

Currently bbr is enabled by default, and domain prefix would always take effect when configured, the shortcoming before this PR is the process would be `add into local network data->remove from local network data->set bbr local->add into local network data`. 

I agree that the first two operations (addition and removal) seem a bit weird, however comparatively speaking, one benefit I can see is that previous implementation could always ensure the prefix is valid and fit in the local network data, at least not overflow when configuring the domain prefix, regardless bbr is enabled or not, given that bbr enable state is configurable by design.

Btw, one note is this PR increases code size a bit (250+ B).

Thoughts?

abtink (approved):
Thanks @XPAWAN . LGTM. 👍

XPAWAN:
The PR has been squashed and rebased. @abtink , @wgtdkp, @librasungirl, @simonlingoogle   can we merge this PR as of it is now ? Other Improvememnts and remarks can be handled in separate PRs.

src/core/thread/network_data_local.cpp
```diff
@@ -111,22 +147,41 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
+    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
+
+    // Check if the prefix overlaps mesh-local prefix.
+    VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()),
+                 error = kErrorInvalidArgs);
+
+exit:
+    return error;
+}
```
simonlingoogle, line 172:
```suggestion
Error Local::ValidatePrefix(const ExternalRouteConfig &aConfig)
{
    Error error = kErrorNone;

    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));

    // Check if the prefix overlaps mesh-local prefix.
    VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()),
                 error = kErrorInvalidArgs);

exit:
    return error;
}
```

This logic is actually common for both `OnMesh` and `HasRoute`

src/core/thread/network_data_local.cpp
```diff
@@ -111,22 +147,41 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+
 Error Local::AddHasRoutePrefix(const ExternalRouteConfig &aConfig)
 {
-    return AddPrefix(aConfig.GetPrefix(), NetworkDataTlv::kTypeHasRoute, aConfig.mPreference, /* aFlags */ 0,
-                     aConfig.mStable);
+    Error error = kErrorNone;
+
+    SuccessOrExit(error = ValidateHasRoutePrefix(aConfig));
+
```
simonlingoogle, line 179:
```suggestion
```
move this check to AddPrefix since the validation is common to both OnMesh and HasRoute?

src/core/thread/network_data_local.cpp
```diff
@@ -144,20 +199,6 @@ Error Local::AddPrefix(const Ip6::Prefix &  aPrefix,
...
-    case OT_ROUTE_PREFERENCE_MED:
-    case OT_ROUTE_PREFERENCE_HIGH:
-        break;
-    default:
-        ExitNow(error = kErrorInvalidArgs);
-    }
-
-    VerifyOrExit(!aPrefix.ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()), error = kErrorInvalidArgs);
-
     IgnoreError(RemovePrefix(aPrefix, aSubTlvType));
```
simonlingoogle, line 202: Validate the prefix using `ValidatePrefix` here?

src/core/thread/network_data_local.cpp
```diff
@@ -69,6 +90,21 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
     // of an IEEE 802.15.4 interface MUST have a length of 64 bits.
     VerifyOrExit(!aConfig.mSlaac || aConfig.mPrefix.mLength == OT_IP6_PREFIX_BITSIZE, error = kErrorInvalidArgs);
 
+    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
+    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
+
+    // Check if the prefix overlaps mesh-local prefix.
+    VerifyOrExit(!aConfig.GetPrefix().ContainsPrefix(Get<Mle::MleRouter>().GetMeshLocalPrefix()),
+                 error = kErrorInvalidArgs);
+
```
simonlingoogle, line 99: The logic here is same as `ValidateHasRoutePrefix` . Use a common method for this validation?

src/core/thread/network_data_local.cpp
```diff
@@ -69,6 +90,21 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
+exit:
+    return error;
+}
+
+Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
+{
+    uint16_t flags = 0;
```
simonlingoogle, line 106:
```suggestion
    uint16_t flags = 0;
    
    SuccessOrExit(error = ValidateOnMeshPrefix(aConfig));
```

src/core/thread/network_data_local.cpp
```diff
@@ -111,22 +147,41 @@ Error Local::AddOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
...
 {
     return RemovePrefix(aPrefix, NetworkDataTlv::kTypeBorderRouter);
 }
 
+Error Local::ValidateHasRoutePrefix(const ExternalRouteConfig &aConfig)
+{
+    Error error = kErrorNone;
+
+    SuccessOrExit(error = ValidatePrefix(aConfig.GetPrefix()));
+    SuccessOrExit(error = ValidatePreference(aConfig.mPreference));
```
simonlingoogle, line 164: If these two methods are only used here, should we just inline the code without calling methods?

XPAWAN:
@simonlingoogle All Remarks has been applied.

librasungirl (approved):

simonlingoogle (approved):
LGTM with nits

Thanks.
Glad to see the code size cost is reduced by recent changes.

jwhui (approved):

PR #6318, opened 2021-03-20
[udp] bind to Thread interface by default

bukepo:
Bind to Thread network by default so that UDP sockets in CLI can send multicast packets through Thread network interface.

### DIFF
--- a/src/core/net/srp_server.cpp
+++ b/src/core/net/srp_server.cpp
@@ -424,7 +424,7 @@ void Server::Start(void)
     VerifyOrExit(!IsRunning());
 
     SuccessOrExit(error = mSocket.Open(HandleUdpReceive, this));
-    SuccessOrExit(error = mSocket.Bind(0));
+    SuccessOrExit(error = mSocket.Bind(0, OT_NETIF_UNSPECIFIED));
 
     SuccessOrExit(error = PublishServerData());
 
--- a/src/core/coap/coap.cpp
+++ b/src/core/coap/coap.cpp
@@ -1666,8 +1666,7 @@ Error Coap::Start(uint16_t aPort, otNetifIdentifier aNetifIdentifier)
     SuccessOrExit(error = mSocket.Open(&Coap::HandleUdpReceive, this));
     socketOpened = true;
 
-    SuccessOrExit(error = mSocket.BindToNetif(aNetifIdentifier));
-    SuccessOrExit(error = mSocket.Bind(aPort));
+    SuccessOrExit(error = mSocket.Bind(aPort, aNetifIdentifier));
 
 exit:
     if (error != kErrorNone && socketOpened)
--- a/src/core/meshcop/dtls.cpp
+++ b/src/core/meshcop/dtls.cpp
@@ -215,7 +215,7 @@ Error Dtls::Bind(uint16_t aPort)
     VerifyOrExit(mState == kStateOpen, error = kErrorInvalidState);
     VerifyOrExit(mTransportCallback == nullptr, error = kErrorAlready);
 
-    SuccessOrExit(error = mSocket.Bind(aPort));
+    SuccessOrExit(error = mSocket.Bind(aPort, OT_NETIF_UNSPECIFIED));
 
 exit:
     return error;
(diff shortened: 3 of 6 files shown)


bukepo:
fixes the issue reported in https://groups.google.com/g/openthread-users/c/gfB6e1W-opM/m/YfYEQhqQFQAJ

src/core/net/srp_server.cpp
```diff
@@ -424,7 +424,7 @@ void Server::Start(void)
     VerifyOrExit(!IsRunning());
 
     SuccessOrExit(error = mSocket.Open(HandleUdpReceive, this));
-    SuccessOrExit(error = mSocket.Bind(0));
+    SuccessOrExit(error = mSocket.Bind(0, OT_NETIF_UNSPECIFIED));
```
wgtdkp, line 427: I think the SRP server should bind to Thread interface.

It doesn't make much sense for the SRP server to accept registrations
from both Thread network and Wi-Fi/Ethernet networks. At least, there
is currently no such use case I think.

Besides, I think the SRP server (and the DNSSD server) should not bind a
platform UDP socket at all because they don't work on both networks. The
only case that I see platform UDP is explicitly required is the Border Agent
that connects external Commissioner and the Leader (and Joiner). For other
general UDP services, I think the expected default behavior is that they should
only work in Thread network unless explicitly required (we may have a flag in
`Udp::Socket` that indicates whether we want to bind to a platform UDP). I think
it is safer and simpler in this way (e.g. we don't need to explicitly filter out TMF
and MLE ports) ,  Thoughts?
@bukepo @jwhui @simonlingoogle @librasungirl

bukepo: This PR focuses on dealing with the multicast issue reported from openthread-users. This change just keeps SRP server's current behavior. I don't have strong opinion on whether the SRP server should be accessible from non-Thread interfaces, but I remember Thread end device may choose any SRP server on the hub network in early discussions. Anyway, that's beyond the purpose of this PR. Should deal with that problem in separate PRs(such as https://github.com/openthread/openthread/pull/6285)?

wgtdkp: Sure, just came up with some thoughts around this.

wgtdkp (approved):
LGTM 👍

jwhui (approved):

src/core/net/srp_server.cpp
simonlingoogle: Let's just use this place for discussion. 

> Besides, I think the SRP server (and the DNSSD server) should not bind a platform UDP socket at all because they don't work on both networks.

Agree that most sockets don't even need to bind to the Thread TUN interface.

> (we may have a flag in Udp::Socket that indicates whether we want to bind to a platform UDP). I think
it is safer and simpler in this way (e.g. we don't need to explicitly filter out TMF and MLE ports). 

Agree that the currently method of listing all UDP ports that do not use PLAT_UDP in `ShouldUsePlatformUdp` is a little bit hardcoding. One solution I thought of is to check Thread UDP sockets first and only forward to Host if the UDP packet is not processed by Thread UDP sockets. But I think this would require significant (and difficult) refactoring to `Ip6::HandleDatagram`.

bukepo: I did some work with the similar idea in https://github.com/openthread/openthread/compare/main...bukepo:refactor/message/read-only-flag

krbvroc1:
Any idea if there are still issues when running under Zephyr? Posted here and then I saw these multiple PR - none of which are in latest Nordic Connect :(

https://groups.google.com/g/openthread-users/c/iS_4MRlc9zo/m/ptD3MnutAAAJ

jwhui:
> Any idea if there are still issues when running under Zephyr? Posted on openthread-users and then I saw these multiple PR - none of which are in latest Nordic Connect :(
> 
> @bukepo @jwhui @wgtdkp
> 
> https://groups.google.com/g/openthread-users/c/iS_4MRlc9zo/m/ptD3MnutAAAJ

This PR addresses issues that only appear when using the the posix platform - specifically when OpenThread is managing a network interface on the host OS as well as Thread. I don't believe this should be an issue in nRF Connect SDK.

PR #6320, opened 2021-03-20
[meshcop] support Thread 1.2 Security Policy flags

wgtdkp:


include/openthread/dataset.h
```diff
@@ -125,14 +125,37 @@ struct otPskc
...
+     */
+    uint16_t mRotationTime;
+
+    /*
+     * The Security Policy Flags as defined in Thread 1.2 Section 8.10.1.15.
+     * For Thread 1.1 or lower, only the first byte is available. For Thread
+     * 1.2 or higher, the available `mFlags` bytes are indicated by `mFlagsLength`.
+     *
+     */
+    uint8_t mFlags[OT_THREAD_1_2_SECURITY_POLICY_FLAGS_LENGTH];
```
abtink, line 149: What do you think of  simplifying the public API/definition for user/caller of OT?
- We can define it here as `uint16_t` `mFlag`
- Update the `OT_SECURITY_POLICY_TOBLE_LINK` and `OT_SECURITY_POLICY_NON_CCM_ROUTERS` to use the higher bits in `uint16_t`
- When dealing with a 1.1 build/device we can leave the extra flags as zero (and ignore them if set by user in a OT API call).
- We then deal with the different flag sizes and 1.1 vs 1.2 inside the core implementation.

Thoughts?

src/core/thread/key_manager.hpp
```diff
@@ -61,6 +62,124 @@ namespace ot {
  * @{
  */
 
+/**
+ * This class represents Security Policy Rotation and Flags.
+ *
+ */
+OT_TOOL_PACKED_BEGIN
+class SecurityPolicy : public otSecurityPolicy
```
abtink, line 70: `otSecurityPolicy` does not seem to be defined as `PACKED` but we are defining its subclass `SecurityPolicy` as packed. So we need it to be packed (I don't think we include it directly in a TLV or in message?).

src/core/thread/key_manager.hpp
```diff
@@ -61,6 +62,124 @@ namespace ot {
...
+        kDefaultKeyRotationTime = 672, ///< Default Key Rotation Time (in unit of hours).
+    };
+
+    enum : uint8_t
+    {
+        kThread11FlagsLength =
+            OT_THREAD_1_1_SECURITY_POLICY_FLAGS_LENGTH, ///< The Thread 1.1 Security Policy Flags length.
+        kThread12FlagsLength =
+            OT_THREAD_1_2_SECURITY_POLICY_FLAGS_LENGTH, ///< The Thread 1.2 Security Policy Flags length.
+    };
```
abtink, line 85: Can this be `private`?

src/core/thread/key_manager.hpp
```diff
@@ -61,6 +62,124 @@ namespace ot {
...
+    {
+        mRotationTime = kDefaultKeyRotationTime;
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+        Encoding::BigEndian::WriteUint16(kDefaultFlags, mFlags);
+        mFlagsLength = kThread12FlagsLength;
+#else
+        mFlags[0]    = kDefaultFlags;
+        mFlagsLength = kThread11FlagsLength;
+#endif
+    }
```
abtink, line 115: Suggestion: How about adding a method `SetToDefault()` and then use it in constructor?
The reason is to help `mSecurityPolicy  = SecurityPolicy();` use the same method in the `dataset.cpp`.

src/core/meshcop/dataset.cpp
```diff
@@ -67,12 +67,11 @@ Error Dataset::Info::GenerateRandom(Instance &aInstance)
...
-    mActiveTimestamp              = 1;
-    mChannel                      = preferredChannels.ChooseRandomChannel();
-    mChannelMask                  = supportedChannels.GetMask();
-    mSecurityPolicy.mRotationTime = KeyManager::kDefaultKeyRotationTime;
-    mSecurityPolicy.mFlags        = KeyManager::kDefaultSecurityPolicyFlags;
-    mPanId                        = Mac::GenerateRandomPanId();
+    mActiveTimestamp = 1;
+    mChannel         = preferredChannels.ChooseRandomChannel();
+    mChannelMask     = supportedChannels.GetMask();
+    mSecurityPolicy  = SecurityPolicy();
```
abtink, line 73: Continue from previous comment:

Suggest  having a method `SecurityPolicy::SetToDefault()` and use it here instead of `mSecurityPolicy  = SecurityPolicy();`?
- Makes the code more readable
- Also a bit more efficient on most toolchains. 
   - With `mSecurityPolicy  = SecurityPolicy();` some toolchains may do an extra object allocation/constructor call and then do a copy. 
   -  With `mSecurityPolicy.SetToDefault()` we know the fields of `mSecurityPolicy` are set directly (on all toolchains).

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -971,52 +971,31 @@ class SecurityPolicyTlv : public Tlv, public TlvInfo<Tlv::kSecurityPolicy>
      * @retval FALSE  If the TLV does not appear to be well-formed.
      *
      */
```
abtink, line 973: I think with the flag extension the  `SecurityPolicyTlv` can have different overall TLV length?
- When appending it to messages, do we always use the extended model (largest TLV length)? Otherwise, we need to be be careful when appending/writing this TLV to a message (some of the `Tlv` methods use the overall size of `Tlv` type).

src/lib/spinel/spinel.h
```diff
@@ -2755,7 +2755,7 @@ enum
      *
      * Content is
      *   `S` : Key Rotation Time (in units of hour)
-     *   `C` : Security Policy Flags (as specified in Thread 1.1 Section 8.10.1.15)
+     *   `D` : Security Policy Flags (as specified in Thread 1.1 Section 8.10.1.15)
```
abtink, line 2758: Suggest to be more specific here in documenting this. `D` really means "blob of data" without `len` being encoded. 

I think would be good to define as this (which keeps it backward compatible):
- `C` Security flag (lower bits)
- `C` Flag extension (higher bits) - This field is optional and may not be present [Then we can describe what is done on `SET` if it is not present and what default values for flags are added].

include/openthread/dataset.h
abtink: We can actually consider defining flags as `bool` bitfields in `otSecurityPolicy` struct and hide/remove the actual bit-flag `OT_SECURITY_POLICY_{}` definitions. Thoughts?

```
typedef struct otSecurityPolicy
{
     uint16_t mRotationTime;  ///< The value for thrKeyRotation in units of hours
     bool mObtainMasterKey : 1;
     bool mNativeCommissioning: 1;
    ...
};
```

I feel this would be even simpler to use by OT API users. Thoughts?

abtink (commented):
Thanks @wgtdkp for taking care of this. 
Some suggestions below

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -971,52 +971,31 @@ class SecurityPolicyTlv : public Tlv, public TlvInfo<Tlv::kSecurityPolicy>
...
+    void SetSecurityPolicy(const SecurityPolicy &aSecurityPolicy);
 
 private:
+    void     SetRotationTime(uint16_t aRotationTime) { mRotationTime = HostSwap16(aRotationTime); }
+    uint16_t GetRotationTime(void) const { return HostSwap16(mRotationTime); }
+    uint8_t  GetFlagsLength(void) const { return GetLength() - sizeof(mRotationTime); }
+
     uint16_t mRotationTime;
-    uint8_t  mFlags;
+    uint8_t  mFlags[SecurityPolicy::kThread12FlagsLength];
```
simonlingoogle, line 998: Should this TLV size kept unchanged for 1.1 builds?

include/openthread/dataset.h
wgtdkp: Defines as bit-fields sounds good to me! I will try to refactor it.

src/core/thread/key_manager.hpp
wgtdkp: Good catch! I was trying to use this in the `SecurityPolicyTlv` but gave it up later. I will remove `OT_TOOL_PACKED_BEGIN` here.

src/core/thread/key_manager.hpp
wgtdkp: `kThread11FlagsLength` is used by `SecurityPolicyTlv` to validation itself.

src/core/meshcop/meshcop_tlvs.hpp
simonlingoogle: Agree. I think a Thread 1.1 device should always append `SecurityPolicyTlv` with 1B flags, and 1.2 device 2B flags.
The best way to guarantee that is to define different structure for different Thread version.

include/openthread/dataset.h
```diff
@@ -125,14 +125,37 @@ struct otPskc
...
+
+    /**
+     * The length of the Security Policy Flags (`mFlags`).
+     *
+     * @sa OT_THREAD_1_1_SECURITY_POLICY_FLAGS_LENGTH
+     * @sa OT_THREAD_1_2_SECURITY_POLICY_FLAGS_LENGTH
+     *
+     */
+    uint8_t mFlagsLength;
```
simonlingoogle, line 158: Can we remove this field so that we always store 2B flags in the Dataset, even for 1.1 devices. We can set the 2nd byte to a default value for 1.1 SecurityPolicyTlv.

wgtdkp: If remove this field, a 1.2 device will always store 2-bytes security policy flags when receiving an operational dataset from a Thead 1.1 device. I don't think this is what we want.

wgtdkp: @librasungirl @abtink Thoughts?

src/core/thread/key_manager.hpp
wgtdkp: I moved the definition to `SecurityPolicyTlv` and declared it as private.

src/core/meshcop/dataset.cpp
wgtdkp: `mSecurityPolicy` is acutally `otSecurityPolicy`, but we can define `SetToDefault` as a static method of `SecurityPolicy` and set `mSecurityPolicy` with `SecurityPolicy::SetToDefault(mSecurityPolicy);`. Updated, PTAL.

include/openthread/dataset.h
wgtdkp: Actually, having those flags as two `uint8_t` can simplifies the NCP implementation. This PR is intended to extend the Security Policy flags to two bytes, is it acceptable to keep current flags representation?

abtink: We can consider including the flag bit mask in NCP/spinel code and convert the `otSecurityPolicy` flags to one or two byte flag that are appended in spinel message. I know it makes the NCP code a bit messy, but I think it may be worth it to be able hide the complexity from the public OT API and make it easier/simpler for user of OT API.

Dealing with two flag bytes and different `OT_SECURITY_POLICY_{}` values (some of which have the same value (same bit)) may make it a bit confusing for some new OT API user (without knowing the details).

Thoughts?

src/core/meshcop/dataset.cpp
abtink: How about doing a `static_cast` on the `mSecurityPolicy`? I thin would work and would be simpler.

We do this with all other types, like from `otIp6Address` to `Ip6::Address`, etc.

wgtdkp: SGTM, updated!

include/openthread/dataset.h
wgtdkp: I thought it is not good to expose internal headers/definitions to CLI/NCP but looks like we are already doing so :)

I made the change accordingly but not finish yet. Will re-request your review when it is ready! Thanks @abtink

wgtdkp: @abtink I made the change, PTAL.

abtink (approved):
Looks great. Thanks @wgtdkp. 👍 👍

simonlingoogle (approved):
LGTM :+1:

src/core/thread/key_manager.hpp
```diff
@@ -61,6 +62,72 @@ namespace ot {
...
+
+    /**
+     * This constructor initializes the object with default Key Rotation Time
+     * and Security Policy Flags.
+     *
+     */
+    SecurityPolicy(void) { SetToDefault(); }
+
+    /**
+     * This static method sets the Security Policy to default values.
```
jwhui, line 86:
```suggestion
     * This method sets the Security Policy to default values.
```

src/cli/cli_dataset.cpp
```diff
@@ -620,6 +595,14 @@ otError Dataset::ProcessMgmtSetCommand(uint8_t aArgsLength, char *aArgs[])
             dataset.mComponents.mIsChannelMaskPresent = true;
             SuccessOrExit(error = ParseAsUint32(aArgs[index], dataset.mChannelMask));
         }
+        else if (strcmp(aArgs[index], "securitypolicy") == 0)
```
jwhui, line 598: Update [src/cli/README_DATASET.md](https://github.com/openthread/openthread/blob/main/src/cli/README_DATASET.md).

src/lib/spinel/spinel.h
```diff
@@ -2797,6 +2797,8 @@ enum
      * Content is
      *   `S` : Key Rotation Time (in units of hour)
      *   `C` : Security Policy Flags (as specified in Thread 1.1 Section 8.10.1.15)
+     *   `C` : Optional Security Policy Flags extension (as specified in Thread 1.2 Section 8.10.1.15).
+     *         0xf8 is used if this field is missining.
```
jwhui, line 2801:
```suggestion
     *         0xf8 is used if this field is missing.
```

src/cli/cli_dataset.cpp
wgtdkp: Updated.

src/cli/README_DATASET.md
```diff
@@ -109,26 +109,36 @@ After the device successfully attaches to a Thread network, the device will retr
...
-- [securitypolicy](#securitypolicy)
+- [OpenThread CLI - Operational Datasets](#openthread-cli---operational-datasets)
+  - [Overview](#overview)
+    - [Active Operational Dataset](#active-operational-dataset)
+    - [Pending Operational Dataset](#pending-operational-dataset)
+  - [Quick Start](#quick-start)
+    - [Form Network](#form-network)
+    - [Attach to Existing Network](#attach-to-existing-network)
+  - [Command List](#command-list)
+  - [Command Details](#command-details)
```
jwhui, line 120: Is this update intentional? These sections are above and not a part of the command list.

src/cli/README_DATASET.md
```diff
@@ -109,26 +109,36 @@ After the device successfully attaches to a Thread network, the device will retr
...
+    - [masterkey](#masterkey)
+    - [meshlocalprefix](#meshlocalprefix)
+    - [mgmtgetcommand](#mgmtgetcommand)
+    - [mgmtsetcommand](#mgmtsetcommand)
+    - [networkname](#networkname)
+    - [panid](#panid)
+    - [pending](#pending)
+    - [pendingtimestamp](#pendingtimestamp)
+    - [pskc](#pskc)
+    - [securitypolicy](#securitypolicy)
```
jwhui, line 140: How about updating the [Security Policy](https://github.com/openthread/openthread/blob/main/src/cli/README_DATASET.md#securitypolicy) section?

src/cli/README_DATASET.md
wgtdkp: Sorry I changed it accidently, reverted.

src/cli/README_DATASET.md
```diff
@@ -513,6 +513,10 @@ Set security policy.
 - r: Thread 1.x Routers are enabled.
 - c: External Commissioner authentication is allowed using PSKc.
 - b: Thread 1.x Beacons are enabled.
+- C: Thread 1.2 Commercial Commissioning is enabled.
+- e: Thread 1.2 Autonomous Enrollment is enabled.
+- p: Thread 1.2 Network Master Key Provisioning is enabled.
+- R: Non-CCM routers are allowed in Thread 1.2 CCM networks.
 
 ```bash
 > dataset securitypolicy 672 onrcb
```
wgtdkp, line 522: @jwhui I didn't update the example command to include the new flags, because those new flags are about CCM and we currently don't have CCM implementation here.

jwhui (approved):

PR #6322, opened 2021-03-22
[srp-server] bind SRP Server to Thread netif

wgtdkp:
This PR binds SRP Server to Thread netif.

We also add a configuration option for the SRP server UDP port to allow using a specific/static port.

See also discussions at https://github.com/openthread/openthread/pull/6318#discussion_r598065071.

### DIFF
--- a/src/core/config/srp_server.h
+++ b/src/core/config/srp_server.h
@@ -46,6 +46,16 @@
 #endif
 
 /**
+ * @def OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT
+ *
+ * Specifies the SRP Server UDP port.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT
+#define OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT 0
+#endif
+
+/**
  * @def OPENTHREAD_CONFIG_SRP_SERVER_SERVICE_NUMBER
  *
  * Specifies the Thread Network Data Service number for SRP Server.
--- a/src/core/net/srp_server.cpp
+++ b/src/core/net/srp_server.cpp
@@ -424,7 +424,7 @@ void Server::Start(void)
     VerifyOrExit(!IsRunning());
 
     SuccessOrExit(error = mSocket.Open(HandleUdpReceive, this));
-    SuccessOrExit(error = mSocket.Bind(0, OT_NETIF_UNSPECIFIED));
+    SuccessOrExit(error = mSocket.Bind(kUdpPort, OT_NETIF_THREAD));
 
     SuccessOrExit(error = PublishServerData());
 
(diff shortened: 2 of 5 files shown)


src/core/config/srp_server.h
```diff
@@ -45,6 +45,16 @@
...
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT
+ *
+ * Specifies the SRP Server UDP port.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT
+#define OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT 53535
```
simonlingoogle, line 55: Dynamic UDP ports range from 49152 to 65535. If this port is larger than 49152, we need to skip 53535 in `Udp::GetEphemeralPort`. 
That being said, I would suggest to use a UDP port < 49152.

Thoughts?

wgtdkp: oh, that is a mistake, I think we should use `35353`.

simonlingoogle (approved):
LGTM :+1:

abtink (approved):
LGTM. 👍

src/core/config/srp_server.h
```diff
@@ -45,6 +45,16 @@
...
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT
+ *
+ * Specifies the SRP Server UDP port.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT
+#define OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT 35353
```
bukepo, line 55: Is this port allocated for SRP?

wgtdkp: No, but since we are binding to only Thread netif, I think it should be okay (just like `49191` and `61631`).

bukepo: Should we make it default to 0? And choose a fixed value in particular project?

bukepo (approved):

src/core/config/srp_server.h
wgtdkp: SGTM!

src/core/config/srp_server.h
wgtdkp: The Border Agent port (`49191`) and Backbone Router port (`61631`) are allocated in this range as well.

src/core/config/srp_server.h
bukepo: as discussed offline, we should consider using `0` for border agent service as well, which makes it easier for us to support multiple ot-daemons on a single host.

wgtdkp: Let's have dynamic port by default.

jwhui (approved):

PR #6328, opened 2021-03-22
[efr32] switch NAME_WLE to NAME_WE to be CMake 3.10 compatible

lmnotran:
NAME_WLE was [added](https://cliutils.gitlab.io/modern-cmake/chapters/intro/newcmake.html#cmake-314--file-utilities-aka-cmake-%CF%80) in CMake 3.14 and isn't available in 3.10, which is the default requirement for openthread

This switches `imported_libs.cmake` to NAME_WL, which is in 3.10


romacdon (approved):
LGTM

jwhui (approved):

PR #6332, opened 2021-03-22
[udp] bind to an interface when sending/connecting

bukepo:
A follow up of #6318 to ensure a socket is bound to an interface when no Udp::Bind is called.


simonlingoogle (approved):
LGTM :+1: 

The minor issue is that `Connect` always use `OT_NETIF_THREAD` if the socket was not previously bound. But I think it's enough for now.

jwhui (approved):

PR #6333, opened 2021-03-23
[ot-ctl] send a line at once

bukepo:
Calling write multiple times may casing different arguments of a command line reach to the daemon at different times. This commit composes arguments into a full command line and sends it at once.

### DIFF
--- a/src/posix/client.cpp
+++ b/src/posix/client.cpp
@@ -177,8 +177,9 @@ int main(int argc, char *argv[])
         for (int i = 1; i < argc; i++)
         {
             int rval = snprintf(&buffer[count], (sizeof(buffer) - count), "%s ", argv[i]);
+
             VerifyOrExit(rval > 0 && static_cast<size_t>(rval) < (sizeof(buffer) - count), ret = OT_EXIT_FAILURE);
-            count += rval;
+            count += static_cast<size_t>(rval);
         }
 
         // replace the trailing space with newline
--- a/src/posix/client.cpp
+++ b/src/posix/client.cpp
@@ -28,6 +28,8 @@
 
 #include "platform/openthread-posix-config.h"
 
+#include "cli/cli_config.h"
+
 #include <openthread/platform/toolchain.h>
 
 #ifndef HAVE_LIBEDIT
@@ -62,7 +64,7 @@
 
 enum
 {
-    kLineBufferSize = 256,
+    kLineBufferSize = OPENTHREAD_CONFIG_CLI_MAX_LINE_LENGTH,
 };
 
 static_assert(kLineBufferSize >= sizeof("> "), "kLineBufferSize is too small");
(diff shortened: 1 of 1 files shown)


bukepo:
Should be helpful to openthread/ot-br-posix#748

src/posix/client.cpp
```diff
@@ -161,23 +163,27 @@ static bool ReconnectSession(void)
...
     {
+        char   buffer[kLineBufferSize];
+        size_t count = 0;
+
         for (int i = 1; i < argc; i++)
         {
-            VerifyOrExit(DoWrite(sSessionFd, argv[i], strlen(argv[i])), ret = OT_EXIT_FAILURE);
-            VerifyOrExit(DoWrite(sSessionFd, " ", 1), ret = OT_EXIT_FAILURE);
+            int rval = snprintf(&buffer[count], (sizeof(buffer) - count), "%s ", argv[i]);
+            VerifyOrExit(rval > 0 && static_cast<size_t>(rval) < (sizeof(buffer) - count), ret = OT_EXIT_FAILURE);
```
wgtdkp, line 180: nit:
```suggestion
            int rval = snprintf(&buffer[count], (sizeof(buffer) - count), "%s ", argv[i]);

            VerifyOrExit(rval > 0 && static_cast<size_t>(rval) < (sizeof(buffer) - count), ret = OT_EXIT_FAILURE);
```

wgtdkp (approved):
LGTM :+1:

simonlingoogle (commented):
Agree that writing a line once is better. 

But why did it have something to do with openthread/ot-br-posix#748 ?
I think the behavior should have no real difference.

@bukepo

bukepo:
> Agree that writing a line once is better.
> 
> But why did it have something to do with [openthread/ot-br-posix#748](https://github.com/openthread/ot-br-posix/issues/748) ?
> I think the behavior should have no real difference.
> 
> @bukepo

The daemon server expects a full line in the current implementation.

zhanglongxia (approved):
LGTM

src/posix/client.cpp
```diff
@@ -28,6 +28,8 @@
 
 #include "platform/openthread-posix-config.h"
 
+#include "cli/cli_config.h"
```
simonlingoogle, line 31: nit: I think `cli` is an app, and posix/client is another app. Is it proper for one app to include headers of another app?

simonlingoogle (approved):
LGTM :+1:

src/posix/client.cpp
bukepo: I take `cli` as a text based control protocol of OpenThread, which can be embedded in other apps, like otbr-agent.

simonlingoogle: Do you imply that we should move `cli` to core?

bukepo: No, `cli` is already a standalone library which can be optionally linked to other apps.

jwhui (approved):

src/posix/client.cpp
```diff
@@ -195,8 +202,11 @@ int main(int argc, char *argv[])
     while (!isFinished)
     {
         fd_set readFdSet;
-        char   buffer[OPENTHREAD_CONFIG_DIAG_CMD_LINE_BUFFER_SIZE];
-        int    maxFd = sSessionFd;
+        char   lineBuffer[kLineBufferSize];
+        char   buffer[kLineBufferSize];
+        size_t lineBufferWritePos = 0;
+        bool   isBeginOfLine      = true;
```
simonlingoogle, line 208: lineBuffer, lineBufferWritePos, isBeginOfLine should be moved out of this scope. We should not re-initialize these values in each loop. 
@bukepo

PR #6334, opened 2021-03-23
[net] add `Ip4::Address` class

abtink:
This commit adds IPv4 address definitions in OpenThread:
- Adds `Ip4::Address` class.
- Adds methods to convert IPv4 address to/from string.
- Updates the `Ip6::Address::FromString()` to use the IPv4
  `FromString()` for the case where the IPv6 address contains
  an embedded IPv4 address.
- Renames unit test to `test_ip_address` and updates it to
  cover the behavior of the newly added IPv4 definitions in 
  addition to IPv6 related tests.

### DIFF
--- a/src/core/net/ip4_address.hpp
+++ b/src/core/net/ip4_address.hpp
@@ -1,0 +1,115 @@
+/*
+ *  Copyright (c) 2021, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes definitions for IPv4 addresses.
+ */
+
+#ifndef IP4_ADDRESS_HPP_
+#define IP4_ADDRESS_HPP_
+
+#include "openthread-core-config.h"
+
+#include "common/clearable.hpp"
+#include "common/equatable.hpp"
+#include "common/error.hpp"
+#include "common/string.hpp"
+
+namespace ot {
+namespace Ip4 {
+
+/**
+ * This class represents an IPv4 address.
+ *
+ */
+OT_TOOL_PACKED_BEGIN
+class Address : public Equatable<Address>, public Clearable<Address>
+{
+public:
+    enum : uint8_t
+    {
+        kSize              = 4,  ///< Size of an IPv4 Address (in bytes).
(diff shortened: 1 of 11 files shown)


src/core/net/ip6_address.hpp
```diff
@@ -917,10 +919,7 @@ class Address : public otIp6Address, public Equatable<Address>, public Clearable
      * @retval false  The IPv6 address is larger than or equal to @p aOther.
      *
      */
-    bool operator<(const Ip6::Address &aOther) const
-    {
-        return memcmp(mFields.m8, aOther.mFields.m8, sizeof(Ip6::Address)) < 0;
-    }
+    bool operator<(const Address &aOther) const { return memcmp(mFields.m8, aOther.mFields.m8, sizeof(Address)) < 0; }
```
simonlingoogle, line 922: Can we compare `m32` directly ?

```suggestion
    bool operator<(const Address &aOther) const { return BigEndian::HostSwap32(mFields.m32) < BigEndian::HostSwap32(aOther.mFields.m32) }
```

src/core/api/ip4_api.cpp
```diff
@@ -0,0 +1,45 @@
...
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file implements the OpenThread IPv4 related API.
+ */
```
simonlingoogle, line 32:
```suggestion
 *   This file implements the OpenThread IPv4 related API.
 *
 */
```

simonlingoogle (approved):
LGTM :+1:

src/core/net/ip6_address.hpp
abtink: @simonlingoogle , this change is actually in `Ip6::Address` not `Ip4::Address`. The IPv6 has more than one `m32[]` field, so `memcmp` is a better way to compare all bytes.

The main change here in this commit/PR is just replacing `Ip6::Address` with `Address` as the type used in the method declaration (since it is all already under the namespace `Ip6`).
```
bool operator<(const Ip6::Address &aOther) const
``` 
to  
```
bool operator<(const Address &aOther) const
```

simonlingoogle: Thanks for the tip.

zhanglongxia (approved):
LGTM

jwhui:
@abtink , I'm wondering if IPv4 addresses should even be exposed via the API. Thread is fundamentally an IPv6 network. Even if the DNS Client is synthesizing IPv6 addresses from IPv4 addresses, the application should just see the synthesized IPv6 address.

Do you have a use case in mind?

abtink:
> @abtink , I'm wondering if IPv4 addresses should even be exposed via the API. Thread is fundamentally an IPv6 network. Even if the DNS Client is synthesizing IPv6 addresses from IPv4 addresses, the application should just see the synthesized IPv6 address.
> 
> Do you have a use case in mind?

I had some ideas about how/where to use this: 
- In the case where we handle the translation in DNS client itself, we can consider to report the original IPv4 address along with the synthesized IPv6 address to the user (considering this for logging/testing). 
- We may consider exposing the v4 to v6 synthesis logic as a public OT API to user.
   - This can be used for cases where the user may already know the IPv4 address (somehow) and just wants to translate it.
- Was also considering adding support and exposing API in DNS client to query for A records directly.

jwhui:
> * In the case where we handle the translation in DNS client itself, we can consider to report the original IPv4 address along with the synthesized IPv6 address to the user (considering this for logging/testing).

I think logging the synthesized IPv6 address is sufficient, since it directly embeds the IPv4 address.

> * We may consider exposing the v4 to v6 synthesis logic as a public OT API to user.
>   
>   * This can be used for cases where the user may already know the IPv4 address (somehow) and just wants to translate it.

I think we should discourage IP address referral where possible.

> * Was also considering adding support and exposing API in DNS client to query for A records directly.

I think it would be useful to have an API that can enable/disable DNS64/NAT64. For example, a client may want to disable NAT64 to ensure that it does not unintentionally communicate with cloud/off-network services. I'm not sure I see the need for allowing the client to query A records directly yet. I think we should discourage IPv4 wherever possible.

abtink:
@jwhui sounds good to me. Will remove the public `otIp4Address` definition and just keep the interval.

src/core/net/ip4_address.hpp
```diff
@@ -0,0 +1,115 @@
...
+ *
+ */
+OT_TOOL_PACKED_BEGIN
+class Address : public Equatable<Address>, public Clearable<Address>
+{
+public:
+    enum : uint8_t
+    {
+        kSize              = 4,  ///< Size of an IPv4 Address (in bytes).
+        kAddressStringSize = 17, ///< Max string size to represent an IPv4 address (`ToString()`).
```
jwhui, line 58: Is the worst case 16 chars (including null char) - e.g. 255.255.255.255?

abtink: Yes. The worst-case string length we would need indeed `16`. 
But I went for `17` here intentionally as a small code-size optimization 😃 
- Every time we use `String<size>` with some size value, a new instance of the template is created (along with its method). Though they are small, they still do increase code size.
- If I see that the info string size for a new type is just few bytes different from an existing size that we already use for some other type, I use that instead to share the `String<size>` template instantiation (and reduce code size).
- We do have some of other types that use 17 as their `kInfoStringSize`.
- I have used this trick in some other cases, we have a bunch of types that use `45` as their string size.

jwhui: Nice! Maybe update the comment to clarify that this is not really the "max string size to represent an IPv4 address"?

tests/unit/test_ip6_address.cpp
```diff
@@ -29,39 +29,47 @@
 #include <limits.h>
 
 #include "common/encoding.hpp"
+#include "net/ip4_address.hpp"
```
jwhui, line 32: Rename `test_ip6_address.cpp` to `test_ip_address.cpp`?

abtink: Good idea. Renamed it in the latest pushed commit.

jwhui (approved):

PR #6340, opened 2021-03-23
[equatable] 'Unequatable' class adding overload of operator `!=`

abtink:
This commit adds `Unequatable<Type>` class which adds an overload of
operator `!=` using an existing `==` overload provided by the `Type`
class. The `Unequatable` should follow CRTP-style inheritance, i.e.,
the `Type` class should publicly inherit from `Unequatable<Type>`.
This provides a solution to avoid repeating the simple implementation
of `operator!=()` overload in different classes.

### DIFF
--- a/src/core/mac/channel_mask.hpp
+++ b/src/core/mac/channel_mask.hpp
@@ -39,6 +39,7 @@
 #include <limits.h>
 #include <openthread/platform/radio.h>
 
+#include "common/equatable.hpp"
 #include "common/string.hpp"
 #include "radio/radio.hpp"
 
@@ -61,7 +62,7 @@ namespace Mac {
  * It is a wrapper class around a `uint32_t` bit vector representing a set of channels.
  *
  */
-class ChannelMask
+class ChannelMask : public Unequatable<ChannelMask>
 {
 public:
     enum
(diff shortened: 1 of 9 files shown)


src/core/mac/channel_mask.hpp
```diff
@@ -61,7 +62,7 @@ namespace Mac {
  * It is a wrapper class around a `uint32_t` bit vector representing a set of channels.
  *
  */
-class ChannelMask
+class ChannelMask : public NotEquatable<ChannelMask>
```
simonlingoogle, line 65: nit: the name `NotEquatable` is a little bit odd because `ChannelMask` objects are equitable but it inherited from `NotEquatable<ChannelMask>`. It sounds like that we can't use `==` and `!=` on `ChannelMask` objects. 

Maybe rename to `Unequatable` ?

```suggestion
class ChannelMask : public Unequatable<ChannelMask>
```

simonlingoogle (approved):
LGTM :+1: 

with a nit

src/core/mac/channel_mask.hpp
abtink: Thanks @simonlingoogle.  I definitely do like the `Unequatable` more (was not very fond of `NotEquatable` name).

jwhui (approved):

PR #6341, opened 2021-03-24
[ping-sender] show ping reply statistics

superwhd:


include/openthread/ping_sender.h
```diff
@@ -58,29 +58,57 @@ typedef struct otPingSenderReply
...
  *
  * @param[in] aReply      A pointer to a `otPingSenderReply` containing info about the received ping reply.
  * @param[in] aContext    A pointer to application-specific context.
  *
  */
-typedef void (*otPingSenderCallback)(const otPingSenderReply *aReply, void *aContext);
+typedef void (*otPingSenderReplyCallback)(const otPingSenderReply *aReply, void *aContext);
+
+/**
+ * This function pointer type specifies the callback to notify receipt of a ping statistics.
```
abtink, line 85:
```suggestion
 * This function pointer type specifies the callback to report the ping statistics.
```

include/openthread/ping_sender.h
```diff
@@ -58,29 +58,57 @@ typedef struct otPingSenderReply
...
+ *
+ * @param[in] aStatistics      A pointer to a `otPingSenderStatistics` containing info about the received ping
+ * statistics.
```
abtink, line 88:
```suggestion
 * @param[in] aStatistics      A pointer to a `otPingSenderStatistics` containing info about the received ping
 *                             statistics.
```

src/core/net/icmp6.hpp
```diff
@@ -325,6 +325,12 @@ class Icmp : public InstanceLocator, private NonCopyable
      */
     bool ShouldHandleEchoRequest(const MessageInfo &aMessageInfo);
 
+    /**
+     * This method returns the ICMPv6 Echo sequence number.
+     * @retval Sequence number of the next ICMPv6 Echo request.
+     */
```
abtink, line 331:
```suggestion
    /**
     * This method returns the ICMPv6 Echo sequence number.
     *
     * @returns The sequence number of the next ICMPv6 Echo request.
     *
     */
```

src/core/net/icmp6.hpp
```diff
@@ -325,6 +325,12 @@ class Icmp : public InstanceLocator, private NonCopyable
...
+    uint16_t GetEchoSequence() const { return mEchoSequence; }
```
abtink, line 332:
```suggestion
    uint16_t GetEchoSequence(void) const { return mEchoSequence; }
```

src/core/utils/ping_sender.cpp
```diff
@@ -60,12 +60,26 @@ void PingSender::Config::SetUnspecifiedToDefault(void)
...
+exit:
+    return;
 }
 
-void PingSender::Config::InvokeCallback(const Reply &aReply) const
+void PingSender::Config::InvokeStatisticsCallback(const Statistics &aStatistics) const
 {
-    VerifyOrExit(mCallback != nullptr);
-    mCallback(&aReply, mCallbackContext);
+    VerifyOrExit(mReplyCallback != nullptr);
```
abtink, line 81:
```suggestion
    VerifyOrExit(mStatisticsCallback != nullptr);
```

src/core/utils/ping_sender.hpp
```diff
@@ -65,6 +66,25 @@ class PingSender : public InstanceLocator, private NonCopyable
...
+        Statistics() { this->Clear(); }
+
+        void Clear()
+        {
+            mTotalPingCount     = 0;
+            mSentCount          = 0;
+            mReceivedCount      = 0;
+            mTotalRoundTripTime = 0;
+            mMinRoundTripTime   = ot::NumericLimits<uint32_t>::Max();
+            mMaxRoundTripTime   = ot::NumericLimits<uint32_t>::Min();
```
abtink, line 84: We can skip the `ot::` since the `PingSender` itself is already under namespace `ot`.
```suggestion
            mMinRoundTripTime   = NumericLimits<uint32_t>::Max();
            mMaxRoundTripTime   = NumericLimits<uint32_t>::Min();
```

src/cli/cli.cpp
```diff
@@ -3155,6 +3155,25 @@ void Interpreter::HandlePingReply(const otPingSenderReply *aReply)
...
 }
 
+void Interpreter::HandlePingStatistics(const otPingSenderStatistics *aStatistics, void *aContext)
+{
+    static_cast<Interpreter *>(aContext)->HandlePingStatistics(aStatistics);
+}
+
+void Interpreter::HandlePingStatistics(const otPingSenderStatistics *aStatistics)
+{
+    double packetLossRate = 1 - static_cast<double>(aStatistics->mReceivedCount) / aStatistics->mTotalPingCount;
```
abtink, line 3165: Noticed that CLI builds (ftd, mtd) has about 2.4KB code size increase from this change. My guess/theory is that it may be due to performing `double` division here to calculate the loss rate.

My suggestion is to use the int (`uint32_t`) division instead and avoid using `double` calculation.
   - We can multiply the `mReceivedCount` by `1000` and then divide using `uint32_t` division.
    - Result `/ 10` gives the percentage and the result `% 10` give the one digit after decimal point `.`

src/cli/cli.cpp
```diff
@@ -3155,6 +3155,25 @@ void Interpreter::HandlePingReply(const otPingSenderReply *aReply)
...
+    OutputFormat("%u packets transmitted, %u packets received, %.1f%% packet loss. ", aStatistics->mTotalPingCount,
+               aStatistics->mReceivedCount, packetLossRate * 100);
+    if (aStatistics->mReceivedCount != 0)
+    {
+        double avgRoundTripTime = static_cast<double>(aStatistics->mTotalRoundTripTime) / aStatistics->mReceivedCount;
```
abtink, line 3170: Same suggestion here to avoid the `double` arithmetic.

abtink (commented):
Thanks @superwhd. Overall looks good. 
Some comments and suggestions below.

tests/scripts/thread-cert/node.py
```diff
@@ -1751,7 +1751,7 @@ def scan(self, result=1):
             return self._expect_results(
                 r'\|\s(\S+)\s+\|\s(\S+)\s+\|\s([0-9a-fA-F]{4})\s\|\s([0-9a-fA-F]{16})\s\|\s(\d+)')
 
-    def ping(self, ipaddr, num_responses=1, size=None, timeout=5):
+    def ping(self, ipaddr, num_responses=1, size=None, timeout=10):
```
simonlingoogle, line 1754: why do we change timeout?

tests/scripts/thread-cert/node.py
```diff
@@ -1762,13 +1762,17 @@ def ping(self, ipaddr, num_responses=1, size=None, timeout=5):
 
         responders = {}
 
+        received_statistics = False
+
         result = True
         # ncp-sim doesn't print Done
         done = (self.node_type == 'ncp-sim')
-        while len(responders) < num_responses or not done:
+        while len(responders) < num_responses or not done or not received_statistics:
```
simonlingoogle, line 1770:
```suggestion
        while not done or not received_statistics:
```
We can exit as long as `received_statistics` is True because the statistics should be the last line? But we need to set result to False when `len(responders) < num_responses`. 

Thoughts?

include/openthread/ping_sender.h
```diff
@@ -58,29 +58,57 @@ typedef struct otPingSenderReply
...
-    bool                 mAllowZeroHopLimit; ///< Indicates whether hop limit is zero.
+    otIp6Address              mDestination;   ///< Destination address to ping.
+    otPingSenderReplyCallback mReplyCallback; ///< Callback function to report replies (can be NULL if not needed).
+    otPingSenderStatisticsCallback
+             mStatisticsCallback; ///< Callback function to report statistics (can be NULL if not needed).
+    void *   mCallbackContext;    ///< A pointer to the callback application-specific context.
+    uint16_t mSize;               ///< Data size (# of bytes) excludes IPv6/ICMPv6 header. Zero for default.
+    uint16_t mCount;              ///< Number of ping messages to send. Zero to use default.
+    uint32_t mInterval;           ///< Ping tx interval in milliseconds. Zero to use default.
+    uint32_t mTimeout; ///< Time in milliseconds to wait for each ping reply after sending out the ping request .
```
simonlingoogle, line 108:
```suggestion
    uint16_t mTimeout; ///< Time in milliseconds to wait for each ping reply after sending out the ping request .
```

uint16_t should be enough

src/core/utils/ping_sender.cpp
```diff
@@ -90,7 +104,11 @@ Error PingSender::Ping(const Config &aConfig)
     mConfig.SetUnspecifiedToDefault();
     VerifyOrExit(mConfig.mInterval <= Timer::kMaxDelay, error = kErrorInvalidArgs);
 
+    mStatistics.Clear();
+    mStatistics.mTotalPingCount = mConfig.mCount;
```
simonlingoogle, line 108:
```suggestion
```

Remove `mStatistics.mTotalPingCount` field ?

include/openthread/ping_sender.h
```diff
@@ -58,29 +58,57 @@ typedef struct otPingSenderReply
...
+ *
+ */
+typedef struct otPingSenderStatistics
+{
+    uint32_t mTotalPingCount;     ///< The number of total ping requests to send.
+    uint32_t mSentCount;          ///< The number of ping requests already sent.
+    uint32_t mReceivedCount;      ///< The number of ping replies received.
+    uint32_t mTotalRoundTripTime; ///< The total round trip time of ping requests.
+    uint32_t mMinRoundTripTime;   ///< The min round trip time among ping requests.
+    uint32_t mMaxRoundTripTime;   ///< The max round trip time among ping requests.
```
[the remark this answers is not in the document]
superwhd: I used `uint32_t` because `otPingSenderReply` is using this type.

Also, I think `mTotalRoundTripTime` could be large (more than 100 seconds) when we use echo a lot of times in a ping request.

src/core/utils/ping_sender.cpp
```diff
@@ -180,12 +210,25 @@ void PingSender::HandleIcmpReceive(const Message &          aMessage,
...
+    VerifyOrExit(reply.mRoundTripTime <= mConfig.mTimeout);
+
+    mStatistics.mReceivedCount++;
+    mStatistics.mTotalRoundTripTime += reply.mRoundTripTime;
+    mStatistics.mMaxRoundTripTime = OT_MAX(mStatistics.mMaxRoundTripTime, reply.mRoundTripTime);
+    mStatistics.mMinRoundTripTime = OT_MIN(mStatistics.mMinRoundTripTime, reply.mRoundTripTime);
+
+    mConfig.InvokeReplyCallback(reply);
+    // Received all ping replies, no need to wait longer.
+    if (aIcmpHeader.GetSequence() == mTargetEchoSequence)
```
simonlingoogle, line 222: This is not strictly correct. 

Besides `PingSender`, user can also use `otIcmp6SendEchoRequest` to send `Ping Requets` even when the `Ping Sender` is busy. `otIcmp6SendEchoRequest` will increase the `Echo Sequence` so that `mTargetEchoSequence` won't match the last echo sequence of `Ping Requests` sent by `Ping Sender`.

I think we should set `mTargetEchoSequence` to the next Echo Sequence in `SendPing` instead of setting it once in `Ping()`, so that `mTargetEchoSequence` effectively tracks the next Echo Sequence to expect, as below:

```suggestion
    if (mCount == 0 && aIcmpHeader.GetSequence() == mTargetEchoSequence)
```

Thoughts?

include/openthread/ping_sender.h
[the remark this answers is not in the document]
simonlingoogle: > Also, I think `mTotalRoundTripTime` could be large (more than 100 seconds) when we use echo a lot of times in a ping request.

We can save the average RTT and update average RTT for each new reply?
```
NewAvgRtt = ((OldAvgRtt * Count) + Reply.RTT) / (Count+1)
```

tests/scripts/thread-cert/node.py
superwhd: Our default ping command timeout is 5 seconds. Using the same timeout here may not be sufficient for the test to wait for ping command to complete.

include/openthread/ping_sender.h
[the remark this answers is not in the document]
simonlingoogle: > I used `uint32_t` because `otPingSenderReply` is using this type.

Should we change `mRoundTripTime` of `otPingSenderReply` to `uint16_t` ?
I think 65535 ms is enough for storing RoundTripTime.

I realized that `otPingSenderReply` is allocated in stack so its memory size is less restricted. 
But should make `otPingSenderStatistics` use less memory and consistent with `otPingSenderReply`. 

Thoughts? 
@abtink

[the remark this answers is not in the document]
superwhd: I'm inclined to keep the variable mTotalRoundTripTime, and make it a uint32_t. 
I tried calculating the average RTT using the equation above, and found the precision poor.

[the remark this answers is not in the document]
abtink: One thing to consider related to RTT is that if we may deal with SED devices which poll very infrequently and then we may have rather long ping delay. `uint16_t` would give us ~65sec. I think it may be possible that we have SEDs that data poll in order of minutes.  So if we can keep it as `uint32_t` it may be better.

> We can save the average RTT and update average RTT for each new reply?
> I tried calculating the average RTT using the equation above, and found the precision poor.

I think the sum model is fine here. But if we want to use running average model, we may need to cauclate the average of "values times some factor" to increase precision. As an example `RssAverager` in `link_quality.hpp` keeps track of average of "RSS times 8".

tests/scripts/thread-cert/node.py
simonlingoogle: Fine to me as long as TC passes.

include/openthread/ping_sender.h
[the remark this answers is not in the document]
simonlingoogle: I don't have strong opinion against `uint32_t mTotalRoundTripTime`, even though I think Thread users usually do no care about ping delay precision under 1ms. 

> I think it may be possible that we have SEDs that data poll in order of minutes. So if we can keep it as uint32_t it may be better.

Agree SED ping delay can be larger than 65s. 

> But if we want to use running average model, we may need to cauclate the average of "values times some factor" to increase precision.

That would make reduce the range of uint16_t from 65s to less than 10s, which would not be enough. I would prefer using uint32_t to this proposal.

tests/scripts/thread-cert/node.py
```diff
@@ -1765,10 +1765,14 @@ def ping(self, ipaddr, num_responses=1, size=None, timeout=5):
         result = True
         # ncp-sim doesn't print Done
         done = (self.node_type == 'ncp-sim')
-        while len(responders) < num_responses or not done:
+
+        # ncp-sim doesn't print statistics
+        received_statistics = (self.node_type == 'ncp-sim')
+        is_multicast = ipaddr.startswith('ff')
```
simonlingoogle, line 1771:
```suggestion
        is_multicast = ipaddress.IPv6Address(ipaddr).is_multicast
```
nit: use [ipaddress library](https://docs.python.org/3/library/ipaddress.html)

abtink (approved):
LGTM. Thanks @superwhd . 👍

src/core/config/ping_sender.h
```diff
@@ -57,6 +57,17 @@
...
 
+/**
+ * @def OPENTHREAD_CONFIG_PING_SENDER_DEFAULT_DEADLINE
+ *
+ * Specifies the default ping timeout in milliseconds. The timeout is the max acceptable time gap between each ping
+ * request and its reply.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_PING_SENDER_DEFAULT_TIMEOUT
+#define OPENTHREAD_CONFIG_PING_SENDER_DEFAULT_TIMEOUT 2000
```
simonlingoogle, line 68:
```suggestion
#define OPENTHREAD_CONFIG_PING_SENDER_DEFAULT_TIMEOUT 3000
```
nit: 2000 is quite small.  Can 3000 work?

simonlingoogle (approved):
LGTM :+1: 

with a nit

include/openthread/ping_sender.h
```diff
@@ -62,35 +62,63 @@ extern "C" {
...
-    bool                 mAllowZeroHopLimit; ///< Indicates whether hop limit is zero.
+    otIp6Address              mDestination;   ///< Destination address to ping.
+    otPingSenderReplyCallback mReplyCallback; ///< Callback function to report replies (can be NULL if not needed).
+    otPingSenderStatisticsCallback
+             mStatisticsCallback; ///< Callback function to report statistics (can be NULL if not needed).
+    void *   mCallbackContext;    ///< A pointer to the callback application-specific context.
+    uint16_t mSize;               ///< Data size (# of bytes) excludes IPv6/ICMPv6 header. Zero for default.
+    uint16_t mCount;              ///< Number of ping messages to send. Zero to use default.
+    uint32_t mInterval;           ///< Ping tx interval in milliseconds. Zero to use default.
+    uint16_t mTimeout; ///< Time in milliseconds to wait for each ping reply after sending out the ping request .
```
jwhui, line 118: nit: trying to make this shorter
```suggestion
    uint16_t mTimeout; ///< Time in milliseconds to wait for a reply after sending out the request.
```

src/core/utils/ping_sender.cpp
```diff
@@ -175,16 +207,30 @@ void PingSender::HandleIcmpReceive(const Message &          aMessage,
     timestamp = HostSwap32(timestamp);
 
     reply.mSenderAddress  = aMessageInfo.GetPeerAddr();
-    reply.mRoundTripTime  = TimerMilli::GetNow() - TimeMilli(timestamp);
+    reply.mRoundTripTime  = static_cast<uint16_t>(TimerMilli::GetNow() - TimeMilli(timestamp));
```
jwhui, line 210: If difference is larger than 65535, should we set to 65535?

jwhui (commented):
Looks great overall. Just a couple minor comments.

jwhui (approved):

PR #6342, opened 2021-03-24
[low-power] resync SSED with its parent after retransmission

brianljt:
Per 15.4 spec, transmitter cannot modify frame content in retransmission which will leads to out-of-date CSL IE being sent to CSL transmitter. This PR fixes this issue by,

1. Send a follow up data poll with CSL IE to resync the CSL transmitter.
2. Send periodic data poll with CSL IE to keep SSED's parent in sync.
3. Move `PrepareDataRequest` to `data_poll_sender`.
4. Add test cases to cover.

IEEE Std 802.15.4 - 6.7.4.3 Retransmission,
>When not using TSCH mode and a frame with the Security Enabled field set to one is retransmitted, the frame shall be retransmitted without changes and without passing through the outgoing frame security procedure, as defined in 9.2.1.

Thread Spec V1.2 - 4.6.5.1.1 Communication between Parent and SSED Child
>A SSED Device MUST include IEEE 802.15.4 CSL Information Elements (CSL Phase, CSL Period) in all the IEEE 802.15.4 Commands, ACKs, and Data frames unicast to its Parent router, as defined in [IEEE 802.15.4-2015] and Table 3-7.


bukepo (approved):

src/core/mac/data_poll_sender.cpp
```diff
@@ -504,8 +544,88 @@ void DataPollSender::HandlePollTimer(Timer &aTimer)
...
 uint32_t DataPollSender::GetDefaultPollPeriod(void) const
 {
-    return Time::SecToMsec(Get<Mle::MleRouter>().GetTimeout()) -
-           static_cast<uint32_t>(kRetxPollPeriod) * kMaxPollRetxAttempts;
+    uint32_t period = Time::SecToMsec(Get<Mle::MleRouter>().GetTimeout());
+
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+    if (Get<Mac::Mac>().IsCslEnabled())
+    {
+        UpdateIfLarger(period, Time::SecToMsec(Get<Mac::Mac>().GetCslTimeout()));
```
Irving-cl, line 552: Csl Timeout should always be smaller than or equal to Router Timeout. If it somehow Csl Timeout is larger than Router Timeout, we shoud use the smaller value?

src/core/mac/data_poll_sender.hpp
```diff
@@ -178,16 +153,28 @@ class DataPollSender : public InstanceLocator, private NonCopyable
...
 
     /**
-     * This method informs the data poll sender to process a MAC frame.
+     * This method informs the data poll sender to process a received MAC frame.
      *
-     *   1. Data Frame: send an immediate data poll if "frame pending" is set.
-     *   2. Ack Frame for a secured data frame in version 1.2 or newer: send an immediate data poll if
-     *      "frame pending" is set, otherwise reset the keep-alive timer for sending next poll.
+     * @param[in] aFrame     The received frame to process.
+     * @param[in] aError     Error status of the transmission.
```
Irving-cl, line 159: There is no such parameter.

src/core/mac/data_poll_sender.hpp
```diff
@@ -178,16 +153,28 @@ class DataPollSender : public InstanceLocator, private NonCopyable
     void HandlePollTimeout(void);
 
     /**
-     * This method informs the data poll sender to process a MAC frame.
+     * This method informs the data poll sender to process a received MAC frame.
      *
-     *   1. Data Frame: send an immediate data poll if "frame pending" is set.
-     *   2. Ack Frame for a secured data frame in version 1.2 or newer: send an immediate data poll if
-     *      "frame pending" is set, otherwise reset the keep-alive timer for sending next poll.
+     * @param[in] aFrame     The received frame to process.
```
Irving-cl, line 158:
```suggestion
     * @param[in] aFrame     A reference to the received frame to process.
```

src/core/mac/data_poll_sender.hpp
```diff
@@ -178,16 +153,28 @@ class DataPollSender : public InstanceLocator, private NonCopyable
...
      *
-     * @param[in] aFrame     The frame to process.
+     */
+    void ProcessRxFrame(const Mac::RxFrame &aFrame);
+
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+    /**
+     * This method informs the data poll sender to process a transmitted MAC frame.
+     *
+     * @param[in]  aFrame      The frame that was transmitted.
```
Irving-cl, line 168:
```suggestion
     * @param[in]  aFrame      A reference to the frame that was transmitted.
```

Irving-cl (commented):
Overall LGTM. Some small issues.

src/core/mac/data_poll_sender.cpp
Irving-cl: My mistake here. Please ignore.

src/core/mac/data_poll_sender.cpp
```diff
@@ -262,7 +262,12 @@ void DataPollSender::HandlePollSent(Mac::TxFrame &aFrame, Error aError)
         otLogInfoMac("Failed to send data poll, error:%s, retx:%d/%d", ErrorToString(aError), mPollTxFailureCounter,
                      kMaxPollRetxAttempts);
 
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+        if (mPollTxFailureCounter <
+            ((aFrame.GetHeaderIe(Mac::CslIe::kHeaderIeId) != nullptr) ? kMaxCslPollRetxAttempts : kMaxPollRetxAttempts))
```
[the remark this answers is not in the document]
brianljt: When CSL IE is included in data poll frame, frame retransmission at Submac and radio layer will be turned off(`data_poll_sender.cpp` line 616). So all the retry of sending data poll will be handled by Mac. The original retry of this is 4 which is too small. We should increase this value when Submac/radio retransmission is disabled.

edmont (approved):

Irving-cl (approved):
LGTM :+1:

src/core/mac/data_poll_sender.cpp
```diff
@@ -496,8 +536,88 @@ void DataPollSender::HandlePollTimer(Timer &aTimer)
...
+    uint32_t period = Time::SecToMsec(Get<Mle::MleRouter>().GetTimeout());
+
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+    if (Get<Mac::Mac>().IsCslEnabled())
+    {
+        period = OT_MIN(period, Time::SecToMsec(Get<Mac::Mac>().GetCslTimeout()));
+    }
+#endif
+
+    period -= (static_cast<uint32_t>(kRetxPollPeriod) * kMaxPollRetxAttempts);
```
abtink, line 548: How about if we protect the calculation here for the roll-over? (if what we subtract is larger than the `period`)?

src/core/mac/data_poll_sender.cpp
```diff
@@ -496,8 +536,88 @@ void DataPollSender::HandlePollTimer(Timer &aTimer)
...
+    {
+        Get<MeshForwarder>().AppendHeaderIe(nullptr, *frame);
+    }
+
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+    if (frame->GetHeaderIe(Mac::CslIe::kHeaderIeId) != nullptr)
+    {
+        // Disable frame retransmission when the data poll has CSL IE included
+        aTxFrames.SetMaxFrameRetries(0);
+    }
```
abtink, line 613: It makes sense to move this method here in `DataPollSender` from `Mac`. 

Just double-chekcing that the `if` block above is the only addition to this method (compared to what it was in `Mac`). Is this correct? Thanks.

src/core/mac/data_poll_sender.hpp
```diff
@@ -130,31 +131,6 @@ class DataPollSender : public InstanceLocator, private NonCopyable
...
-     * This method gets the destination MAC address for a data poll frame.
-     *
-     * @param[out] aDest       Reference to a `MAC::Address` to output the poll destination address (on success).
-     *
-     * @retval kErrorNone   @p aDest was updated successfully.
-     * @retval kErrorAbort  Abort the data poll transmission (not currently attached to any parent).
-     *
-     */
-    Error GetPollDestinationAddress(Mac::Address &aDest) const;
-#endif // #if OPENTHREAD_CONFIG_MULTI_RADIO
```
abtink, line 156: Nice! 👍  on moving them to be private now

abtink (approved):
Looks great overall. Thanks. 👍 

One suggestion below to check and protect if the calculation may roll-over on default period (just to be safe with all config values).

src/core/mac/data_poll_sender.cpp
brianljt: Sounds good!

src/core/mac/data_poll_sender.cpp
brianljt: That's right.

abtink (approved):

src/core/thread/mle.cpp
```diff
@@ -2305,6 +2305,7 @@ Error Mle::SendChildUpdateRequest(void)
     if (!IsRxOnWhenIdle())
     {
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+        Get<DataPollSender>().RecalculatePollPeriod();
```
jwhui, line 2308: What scenario requires adding this here?

brianljt: When user set a new CSL timeout period to SSED, SSED will schedule a Child Update Request. Ref: https://github.com/openthread/openthread/blob/main/src/core/mac/mac.cpp#L2533

jwhui: For Child Timeout, we make this call early in `Mle::SetTimeout()` - see https://github.com/openthread/openthread/blob/21417407dd4b446b4775d889948e9eb06852da82/src/core/thread/mle.cpp#L727

Should we be more consistent in when we call `RecalculatePollPeriod()`? For example, should we move the call in `Mle::SetTimeout()` to here? Or should we move this call into `Mle::SetCslTimeout()`?

brianljt: SGTM, updated.

abtink (approved):
LGTM. 👍

jwhui (approved):

PR #6343, opened 2021-03-24
[low-power] add receive timing capability to CSL receiver

edmont:
This PR uses the receive timing capability from the radio to schedule CSL receive operations with a higher precision, allowing for shorter receive windows.

`mCslTimer` is set up to be triggered `OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD` before the start of the receive window.

Depends on:
- [x]  https://github.com/openthread/openthread/pull/6311 [low-power] adaptative CSL window duration

### DIFF
--- a/examples/platforms/nrf528xx/nrf52811/openthread-core-nrf52811-config.h
+++ b/examples/platforms/nrf528xx/nrf52811/openthread-core-nrf52811-config.h
@@ -161,6 +161,16 @@
 #endif
 
 /**
+ * @def OPENTHREAD_CONFIG_MAC_SOFTWARE_RX_TIMING_ENABLE
+ *
+ * Define to 1 to enable software reception target time logic.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_MAC_SOFTWARE_RX_TIMING_ENABLE
+#define OPENTHREAD_CONFIG_MAC_SOFTWARE_RX_TIMING_ENABLE (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
+#endif
+
+/**
  * @def OPENTHREAD_CONFIG_PLATFORM_USEC_TIMER_ENABLE
  *
  * Define to 1 if you want to support microsecond timer in platform.
@@ -224,6 +234,27 @@
  */
 #ifndef RADIO_CONFIG_SRC_MATCH_EXT_ENTRY_NUM
 #define RADIO_CONFIG_SRC_MATCH_EXT_ENTRY_NUM 0
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD
+ *
+ * Reception scheduling and ramp up time needed for the CSL receiver to be ready, in units of microseconds.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD
+#define OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD 2000
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_CSL_MIN_RECEIVE_ON
+ *
+ * The minimum CSL receive window (in microseconds) required to receive an IEEE 802.15.4 frame.
+ * - Frame preamble: 6*2 symbols + margin
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_CSL_MIN_RECEIVE_ON
+#define OPENTHREAD_CONFIG_CSL_MIN_RECEIVE_ON 12 * 16
 #endif
 
 /*
--- a/examples/platforms/nrf528xx/src/radio.c
+++ b/examples/platforms/nrf528xx/src/radio.c
@@ -79,6 +79,7 @@
 #define SECURITY_ENABLED_BIT     (1 << 3)  ///< Security enabled bit.
 
 #define RSSI_SETTLE_TIME_US   40           ///< RSSI settle time in microseconds.
+#define SAFE_DELTA            1000         ///< A safe value for the `dt` parameter of delayed operations.
 
 #if defined(__ICCARM__)
 _Pragma("diag_suppress=Pe167")
@@ -488,12 +489,26 @@ otError otPlatRadioReceive(otInstance *aInstance, uint8_t aChannel)
     }
 
     nrf_802154_tx_power_set(GetTransmitPowerForChannel(aChannel));
-
     result = nrf_802154_receive();
     clearPendingEvents();
 
     return result ? OT_ERROR_NONE : OT_ERROR_INVALID_STATE;
 }
+
+#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
+otError otPlatRadioReceiveAt(otInstance *aInstance, uint8_t aChannel, uint32_t aStart, uint32_t aDuration)
+{
+    OT_UNUSED_VARIABLE(aInstance);
+
+    bool result;
+
+    nrf_802154_tx_power_set(GetTransmitPowerForChannel(aChannel));
+    result = nrf_802154_receive_at(aStart - SAFE_DELTA, SAFE_DELTA, aDuration, aChannel);
+    clearPendingEvents();
+
+    return result ? OT_ERROR_NONE : OT_ERROR_FAILED;
+}
+#endif
 
 otError otPlatRadioTransmit(otInstance *aInstance, otRadioFrame *aFrame)
 {
@@ -575,7 +590,7 @@ otRadioCaps otPlatRadioGetCaps(otInstance *aInstance)
 
     return (otRadioCaps)(OT_RADIO_CAPS_ENERGY_SCAN | OT_RADIO_CAPS_ACK_TIMEOUT | OT_RADIO_CAPS_CSMA_BACKOFF |
 #if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
-                         OT_RADIO_CAPS_TRANSMIT_SEC | OT_RADIO_CAPS_TRANSMIT_TIMING |
+                         OT_RADIO_CAPS_TRANSMIT_SEC | OT_RADIO_CAPS_TRANSMIT_TIMING | OT_RADIO_CAPS_RECEIVE_TIMING |
 #endif
                          OT_RADIO_CAPS_SLEEP_TO_TX);
 }
@@ -1053,6 +1068,7 @@ void nrf_802154_receive_failed(nrf_802154_rx_error_t error)
     case NRF_802154_RX_ERROR_ABORTED:
     case NRF_802154_RX_ERROR_DELAYED_TIMESLOT_DENIED:
     case NRF_802154_RX_ERROR_INVALID_LENGTH:
+    case NRF_802154_RX_ERROR_DELAYED_ABORTED:
         sReceiveError = OT_ERROR_FAILED;
         break;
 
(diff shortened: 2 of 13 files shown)


edmont:
In the attached capture CSL transmissions are being triggered 300 s after last synchronization. Build from this PR is configured with ±1 ppm crystals in both ends and ±100 us parent uncertainty. CSL receive window duration goes from 492 us at the sync time to 1692 us after 300 s.

[csl_tx_300s.pcap.zip](https://github.com/openthread/openthread/files/6213125/csl_tx_300s.pcap.zip)

Irving-cl:
I'm thinking that, for SSED, not all 'Receive' operations are for CSL Sampling. For example, an SSED can also send a data poll and wait for packets. If we implement `otPlatRadioReceive` as:
```
    if (sCslSampleDur > 0)
    {
        result = nrf_802154_receive_at(sCslSampleTime, sCslSampleStart, sCslSampleDur, aChannel);
    }
    else
#endif
```
In the data poll case, the SSED need to start receiving immediately after it sends the poll. It should't wailt until next sample window.

Thoughts? @edmont @brianljt @abtink

src/core/radio/radio.hpp
```diff
@@ -408,11 +408,22 @@ class Radio : public InstanceLocator, private NonCopyable
...
 
+#if OPENTHREAD_CONFIG_MAC_SOFTWARE_RX_TIMING_ENABLE
+    /**
+     * This method updates the CSL start time and duration in radio.
+     *
+     * @param[in]  aCslSampleStart  The CSL sample start time.
+     * @param[in]  aCslSampleDur    The CSL sample duration.
+     *
+     */
+    void UpdateCslSampleStart(uint32_t aCslSampleStart, uint32_t aCslSampleDur);
```
brianljt, line 424: Looks more like updating a sample window(start and end) than just start to me
```suggestion
    void UpdateCslSampleWindow(uint32_t aCslSampleStart, uint32_t aCslSampleDur);
```

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -489,7 +491,16 @@ otError otPlatRadioReceive(otInstance *aInstance, uint8_t aChannel)
 
     nrf_802154_tx_power_set(GetTransmitPowerForChannel(aChannel));
 
-    result = nrf_802154_receive();
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+    if (sCslSampleDur > 0)
```
brianljt, line 495: I would suggest to use a separate API like `otPlatRadioReceiveAt` for this purpose rather than reuse `otPlatRadioReceive` with some global parameters. There are some wakes not related to CSL managed by OT stack. If we change this implementation like this, it would mess up the OT managed wake.

edmont: Makes sense, I'll update.

src/core/mac/sub_mac.cpp
```diff
@@ -1013,16 +1012,25 @@ void SubMac::HandleCslTimer(void)
         break;
 
     case kCslSleep:
-        mCslState = kCslSample;
+        mCslSampleTime += periodUs;
 
+#if OPENTHREAD_CONFIG_MAC_SOFTWARE_RX_TIMING_ENABLE
```
brianljt, line 1017: Define something like `OT_RADIO_CAPS_RECEIVE_TIMING` like `OT_RADIO_CAPS_TRANSMIT_TIMING` as radio capability and replace this macro check with runtime check `SubMac::ShouldHandleReceiveTargetTime` like `SubMac::ShouldHandleTransmitTargetTime`?

edmont: This was my original idea for the first commit (https://github.com/openthread/openthread/pull/6343/commits/0d92ffb0cd52937527776e681b7fb92a46b18236), but I see two problems:
- Adds unneeded code size.
- Forces all platforms to implement a dummy `otPlatRadioReceiveAt`.

brianljt: I think it is for supporting both NCP mode and RCP mode(posix host + RCP firmware). In RCP mode, `SubMac.cpp` will be included in **both** posix host and RCP firmware. In host side, it will get the radio capability from `SubMac::GetCaps` which combines the real radio capability and the SW capability as the `SubMac` can add SW capability which is missed in radio. If `posix host` finds RCP supports `OT_RADIO_CAPS_RECEIVE_TIMING`, it should not start the receive timer and leave it to RCP. On RCP side, this could be done in `SubMac` layer or radio layer, depends on whether radio has the capability or not in `Get<Radio>().GetCaps()`. If radio supports it, it will leave it to radio and should not start receive timer, if not, it will handle it in `SubMac` layer.

In NCP mode, there will be only one `SubMac` layer and it decides if `SubMac` SW stack should handle the receive timer or not base on Radio's capability.

The slight increase of code size is unavoidable as we share the same `SubMac` code for both posix host build and NCP, RCP firmware build. This is also inline with other radio capabilities like `OT_RADIO_CAPS_ACK_TIMEOUT`, `OT_RADIO_CAPS_TRANSMIT_RETRIES` and `OT_RADIO_CAPS_CSMA_BACKOFF`.

For dummy `otPlatRadioReceiveAt`, I think you can define a weak function in general in `src/core/radio/radio_platform.cpp` to avoid implement it for all platforms like what we have for `otPlatRadioSetMacFrameCounter` or `otPlatRadioGetCslAccuracy`. How do you think?

edmont: Thanks for the clarification. It makes sense, I'll add the changes.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -488,13 +489,27 @@ otError otPlatRadioReceive(otInstance *aInstance, uint8_t aChannel)
...
 
     nrf_802154_tx_power_set(GetTransmitPowerForChannel(aChannel));
-
     result = nrf_802154_receive();
     clearPendingEvents();
 
     return result ? OT_ERROR_NONE : OT_ERROR_INVALID_STATE;
 }
 
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE && OPENTHREAD_CONFIG_MAC_SOFTWARE_RX_TIMING_ENABLE
```
brianljt, line 498:
```suggestion
#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
```
`otPlatRadioGetCaps` indicates `OT_RADIO_CAPS_RECEIVE_TIMING` for OT1.2, then this method should be provided for OT1.2.

examples/platforms/nrf528xx/src/radio.c
```diff
@@ -1065,7 +1081,17 @@ void nrf_802154_receive_failed(nrf_802154_rx_error_t error)
     sAckedWithSecEnhAck = false;
 #endif
 
-    setPendingEvent(kPendingEventReceiveFailed);
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE && OPENTHREAD_CONFIG_MAC_SOFTWARE_RX_TIMING_ENABLE
```
brianljt, line 1084:
```suggestion
#if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
```

brianljt (approved):
LGTM 👍, Thanks!

Irving-cl (approved):
LGTM

edmont:
Added last review changes and squashed.

include/openthread/platform/radio.h
```diff
@@ -680,6 +681,18 @@ otError otPlatRadioSleep(otInstance *aInstance);
  */
 otError otPlatRadioReceive(otInstance *aInstance, uint8_t aChannel);
 
+/**
+ * Schedule a radio reception window at a specific time and duration.
+ *
+ * @param[in]  aChannel  The radio channel on which the frame is to be received.
+ * @param[in]  aStart     The receive window start time, in microseconds.
```
jwhui, line 688: nit: alignment
```suggestion
 * @param[in]  aChannel   The radio channel on which the frame is to be received.
 * @param[in]  aStart     The receive window start time, in microseconds.
```

examples/platforms/nrf528xx/nrf52811/openthread-core-nrf52811-config.h
```diff
@@ -160,6 +160,16 @@
...
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_MAC_SOFTWARE_RX_TIMING_ENABLE
+ *
+ * Define to 1 to enable software reception target time logic.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_MAC_SOFTWARE_RX_TIMING_ENABLE
+#define OPENTHREAD_CONFIG_MAC_SOFTWARE_RX_TIMING_ENABLE (OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2)
```
jwhui, line 170: Is it safe to use `OPENTHREAD_CONFIG_THREAD_VERSION` in the project config? The defaults are included after the project config.

https://github.com/openthread/openthread/blob/8467a3d5223c274b65a71080f6aad94dc182a9bb/src/core/openthread-core-config.h#L45-L51

src/core/radio/radio.hpp
```diff
@@ -408,11 +408,24 @@ class Radio : public InstanceLocator, private NonCopyable
     /**
      * This method updates the CSL sample time in radio.
      *
-     * @param[in]  aCslSampleTime  The CSL sample time.
+     * @param[in]  aCslSampleTime   The CSL sample time.
```
jwhui, line 411: nit: unnecessary change?
```suggestion
     * @param[in]  aCslSampleTime  The CSL sample time.
```

src/core/radio/radio.hpp
```diff
@@ -408,11 +408,24 @@ class Radio : public InstanceLocator, private NonCopyable
...
      *
      */
     void UpdateCslSampleTime(uint32_t aCslSampleTime);
 
+    /**
+     * This method schedules a radio reception window at a specific time and duration.
+     *
+     * @param[in]  aChannel   The radio channel on which the frame is to be received.
```
jwhui, line 419: nit: we don't know that there will be a frame to receive
```suggestion
     * @param[in]  aChannel   The radio channel on which to receive.
```

examples/platforms/nrf528xx/nrf52811/openthread-core-nrf52811-config.h
```diff
@@ -226,6 +236,28 @@
...
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD
+#define OPENTHREAD_CONFIG_CSL_RECEIVE_TIME_AHEAD 2000
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_CSL_MIN_RECEIVE_ON
+ *
+ * The total duration, in units of microseconds, required for the CSL receiver to fully receive and acknowledge a frame
```
jwhui, line 252: I think this is something more like: "The minimum CSL receive window (in microseconds) required to receive an IEEE 802.15.4 frame."

Total duration indicates the total time it takes to receive and transmit the ack, which should be much longer than the 12 * 16 = 196us value used below.

examples/platforms/nrf528xx/nrf52811/openthread-core-nrf52811-config.h
edmont: It seems not. This was just inheriting convention from previous TX capabilities.

I would say the right way should be to just confirm or deny the capabilities in the platforms configs (`0` or `1`) and make sure they are used correctly within OT code depending on the configured Thread version.

But could we address this in a different PR?

jwhui: I am OK to resolve in a separate PR.

jwhui (approved):

Issue #6344, opened 2021-03-24
[posix] ot-br-posix dry run will be affected by failed initialization of irrelavant modules


Irving-cl:
**Describe the bug** A clear and concise description of what the bug is.
When I run ot-br-posix on a device, I found that I cannot correctly get RCP version by using `--radio-version`. The reason is that `platformBackboneInit` in `otSysInit` failed and the program directly exited. The code to handle `radio-version` is after `otSysInit` and doesn't have a chance to run. I think it's possible that some devices don't have backbone interface or temporarily the interface doesn't work. But it should break other dry-runs of ot-br-posix. For example, users just want to output the RCP version. Then it should be correctly output even backbone is missing.

I cannot think of a 'perfect and elegant' implementation. My idea is that we keep the declaration of `otSysInit` for posix, and we devide it into two functions: one before `Instance` created, one after `Instance` created. And we put those initialization that are dependent on `Instance` (e.g. `platformBackboneInit`) in the second function. And `InitInstance` looks like:
```
instance = otSysInit(&aConfig->mPlatformConfig);

atexit(otSysDeinit);

if (aConfig->mPrintRadioVersion)
{
    ......
}
else
{
    ......
}

if (aConfig->mIsDryRun)                                                                                                                                                                                                                                                                                     
{                                                                                                                                                                                                                                                                                                           
    ......                                                                                                                                                                                                                                                  
}

otSysInitWithInstance(&aConfig->mPlatformConfig, instance); // do things like `platformBackboneInit`
```

I think when we do dry-run, we don't actually need to initialize everything. Thoughts? @bukepo @librasungirl @wgtdkp @abtink

bukepo:
dry run can be used to test the stack can be initialized successfully.

simonlingoogle:
Is this feature absolute required in some scenario ? I think we can just specify `-B ....` even for dry runs. It seems a overkill for me to implement such feature just to type less letters.

PR #6347, opened 2021-03-25
[posix] add posix system minimal init

Irving-cl:
This PR solves #6344.

This PR adds a method `otSysInitMinimal` for initialization when `dry-run` option is passed. When doing dry-run, many components are not necessary to initialize. This also avoids dry-run action (like print radio version) from being broken by failure in initialization of other unrelated components.

### DIFF
--- a/src/posix/main.c
+++ b/src/posix/main.c
@@ -314,9 +314,16 @@ static otInstance *InitInstance(PosixConfig *aConfig)
     syslog(LOG_INFO, "Thread version: %hu", otThreadGetVersion());
     IgnoreError(otLoggingSetLevel(aConfig->mLogLevel));
 
-    instance = otSysInit(&aConfig->mPlatformConfig);
-
     atexit(otSysDeinit);
+
+    if (aConfig->mIsDryRun)
+    {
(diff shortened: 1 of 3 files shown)


wgtdkp (approved):
LGTM :+1:

src/posix/main.c
```diff
@@ -314,10 +314,17 @@ static otInstance *InitInstance(PosixConfig *aConfig)
     syslog(LOG_INFO, "Thread version: %hu", otThreadGetVersion());
     IgnoreError(otLoggingSetLevel(aConfig->mLogLevel));
 
-    instance = otSysInit(&aConfig->mPlatformConfig);
-
     atexit(otSysDeinit);
 
+    if (aConfig->mIsDryRun)
+    {
+        instance = otSysInitMinimal(&aConfig->mPlatformConfig);
```
bukepo, line 321: What if we add another option to get information only available when another component is initialized?

Irving-cl: For extension in the future, I think we can calculate which components should be initialized in this function from the config. Thoughts? @bukepo

Irving-cl:
As discussed, we will keep it as it is and add backbone parameters when starting the daemon.

PR #6348, opened 2021-03-25
[border-router] accept RA initiated from the infra interface

wgtdkp:
Accept RA initiated from the infra interface so that we can detect the existence of another RA daemon
which is working on the same infra interface.
(We also did an enhancement to the `test_single_border_router.py` test script to make it more readable.

TODO:
- [x] add CI tests.

### DIFF
--- a/src/core/border_router/routing_manager.cpp
+++ b/src/core/border_router/routing_manager.cpp
@@ -249,9 +249,6 @@ void RoutingManager::RecvIcmp6Message(uint32_t            aInfraIfIndex,
     VerifyOrExit(aInfraIfIndex == mInfraIfIndex, error = kErrorDrop);
     infraLinkLocalAddr = static_cast<const Ip6::Address *>(&mInfraIfLinkLocalAddress);
 
-    // Drop any ICMPv6 messages sent from myself.
-    VerifyOrExit(infraLinkLocalAddr != nullptr && aSrcAddress != *infraLinkLocalAddr, error = kErrorDrop);
-
     VerifyOrExit(aBuffer != nullptr && aBufferLength >= sizeof(*icmp6Header), error = kErrorParse);
 
     icmp6Header = reinterpret_cast<const Ip6::Icmp::Header *>(aBuffer);
@@ -262,6 +259,8 @@ void RoutingManager::RecvIcmp6Message(uint32_t            aInfraIfIndex,
         HandleRouterAdvertisement(aSrcAddress, aBuffer, aBufferLength);
         break;
     case Ip6::Icmp::Header::kTypeRouterSolicit:
+        // Drop Router Solicitations initiated from infra interface.
+        VerifyOrExit(aSrcAddress != *infraLinkLocalAddr, error = kErrorDrop);
         HandleRouterSolicit(aSrcAddress, aBuffer, aBufferLength);
         break;
     default:
@@ -637,6 +636,8 @@ void RoutingManager::StartRoutingPolicyEvaluationDelay(void)
 // between 0 and kMaxRtrSolicitationDelay.
 void RoutingManager::StartRouterSolicitationDelay(void)
 {
+    OT_ASSERT(mAdvertisedOnLinkPrefix == nullptr);
+
     uint32_t randomDelay;
 
     mRouterSolicitCount = 0;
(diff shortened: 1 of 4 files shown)


tests/scripts/thread-cert/border_router/test_single_border_router.py
```diff
@@ -132,195 +130,205 @@ def test(self):
...
+        br.start_radvd_service(prefix=config.ONLINK_GUA_PREFIX, slaac=True)
+        self.simulator.go(5)
+
+        self.assertEqual(len(br.get_routes()), 1)
+        self.assertTrue(br.get_routes()[0].startswith(config.ONLINK_GUA_PREFIX.split('::/')[0]))
+        self.assertEqual(len(router.get_routes()), 1)
+        self.assertTrue(router.get_routes()[0].startswith(config.ONLINK_GUA_PREFIX.split('::/')[0]))
+
+        self.assertTrue(router.ping(host.get_ip6_address(config.ADDRESS_TYPE.ONLINK_GUA)[0]))
+        self.assertTrue(host.ping(router.get_ip6_address(config.ADDRESS_TYPE.OMR)[0], backbone=True))
```
wgtdkp, line 331: Here is the test case for the new behavior.

tests/scripts/thread-cert/border_router/test_single_border_router.py
```diff
@@ -132,195 +130,205 @@ def test(self):
...
```
wgtdkp, line 331: Here is the test case for the new behavior.

abtink (approved):
LGTM. 👍

simonlingoogle (approved):
LGTM :+1:

src/core/border_router/routing_manager.cpp
```diff
@@ -262,6 +259,8 @@ void RoutingManager::RecvIcmp6Message(uint32_t            aInfraIfIndex,
         HandleRouterAdvertisement(aSrcAddress, aBuffer, aBufferLength);
         break;
     case Ip6::Icmp::Header::kTypeRouterSolicit:
+        // Drop Router Solicitations initiated from infra interface.
+        VerifyOrExit(aSrcAddress != *infraLinkLocalAddr, error = kErrorDrop);
```
bukepo, line 263: Does IPv6 allows multiple link local addresses on a single network interface? Is it safer to use network interface index instead of the link local address here?

bukepo (approved):

src/core/border_router/routing_manager.cpp
```diff
@@ -249,9 +249,6 @@ void RoutingManager::RecvIcmp6Message(uint32_t            aInfraIfIndex,
     VerifyOrExit(aInfraIfIndex == mInfraIfIndex, error = kErrorDrop);
```
wgtdkp, line 249: Filter out packets with interface index.

src/core/border_router/routing_manager.cpp
wgtdkp: 1. I don't think it is possible to get the network interface index of the sender.
2. Even if we can, the sender network interface index can not be used for filtering the packet, because network interface on host A and host B can have the same index.
3. It is possible for an interface to have multiple LLA, but this should be rare. I didn't observe multiple LLAs on a single interface.

bukepo: I meant can we have a function like `FindNetifIndex(aSrcAddress)`, and compare it with InfraLink's index.

wgtdkp: Thanks, I think it works in this way. May address it in separate PR I think.

jwhui (approved):

src/core/border_router/routing_manager.cpp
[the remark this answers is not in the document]
wgtdkp: @jwhui Sorry I am not sure if got you. The code here is dropping RS initiated from infra interface. The RS is typically sent by ourselves or some RA daemon which is working on the same interface. The RS messages sent from neighboring routers will be processed.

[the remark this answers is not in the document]
jwhui: Should we handle the case where an RS is sent, but no RA is received from a router we are maintaining prefix state for?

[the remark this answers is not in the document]
wgtdkp: Yes that is an enhancement we'd like to add. I am assuming there will not be two Border Routers on the same host and work on the same infra interface. Are you considering this case? For cases that two BRs are running on different hosts, the RS message will be respected.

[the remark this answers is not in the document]
jwhui: Maybe. There are users of OpenThread that have BRs with multiple Thread interfaces. But I agree that processing a locally-generated RS is a low priority optimization.

PR #6357, opened 2021-03-26
[dns-client] simplify `SendQuery()` and add comment

abtink:
This commit contains two small changes in `Dns::Client`: simplifying
the `SendQuery()` call and adding a comment about reason for parsing
the response in a separate method.

### DIFF
--- a/src/core/net/dns_client.cpp
+++ b/src/core/net/dns_client.cpp
@@ -564,7 +564,7 @@ Error Client::StartQuery(QueryInfo &        aInfo,
     SuccessOrExit(error = AllocateQuery(aInfo, aLabel, aName, query));
     mQueries.Enqueue(*query);
 
-    SendQuery(*query);
+    SendQuery(*query, aInfo, /* aUpdateTimer */ true);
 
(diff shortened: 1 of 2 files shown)


src/core/net/dns_client.cpp
```diff
@@ -788,6 +779,11 @@ void Client::ProcessResponse(const Message &aMessage)
 
     response.mMessage = &aMessage;
 
+    // We intentionally parse the response in a separate method
+    // `ParseResponse()` to free all the stack allocated variables
+    // (e.g., `QueryInfo`) used during parsing of the message before
+    // finalizing the query and invoking the user's callback.
```
simonlingoogle, line 785: Yes. I noticed same requirement when implementing DNS server code. I tried to make sure that `char name[256]` for storing DNS names are only used in leaf functions. 

Wonder if we can have some tool that can analyze the stack behavior of the OT program.

simonlingoogle (approved):
LGTM 👍

jwhui (approved):

PR #6368, opened 2021-03-29
[dns] allow disabling of DNS name compression for testing

abtink:
This commit adds support for "DNS name compression" mode. By default
the DNS name compression is enabled. When disabled, DNS names are
appended as full and never compressed. This is applicable to
OpenThread's DNS and SRP client/server modules. This is intended for
testing only and available when config option
`OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE` is enabled.

This commit also add a CLI command for this feature as 
`dns compression [enable|disable]` and updates the CLI's `README.md`.

### DIFF
--- a/include/openthread/dns.h
+++ b/include/openthread/dns.h
@@ -38,6 +38,7 @@
 #include <stdint.h>
 
 #include <openthread/error.h>
+#include <openthread/instance.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -139,6 +140,32 @@ void otDnsInitTxtEntryIterator(otDnsTxtEntryIterator *aIterator, const uint8_t *
  *
  */
 otError otDnsGetNextTxtEntry(otDnsTxtEntryIterator *aIterator, otDnsTxtEntry *aEntry);
+
+/**
+ * This function enables/disables the "DNS name compression" mode.
+ *
+ * By default DNS name compression is enabled. When disabled, DNS names are appended as full and never compressed. This
+ * is applicable to OpenThread's DNS and SRP client/server modules.
+ *
+ * This is intended for testing only and available when `OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE` config is enabled.
+ *
+ * Note that in the case `OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE` is used, this mode applies to all OpenThread
+ * instances (i.e., calling this function enables/disables the compression mode on all OpenThread instances).
+ *
+ * @param[in] aEnabled   TRUE to enable the "DNS name compression" mode, FALSE to disable.
+ *
+ */
+void otDnsSetNameCompressionEnabled(bool aEnabled);
+
+/**
+ * This function indicates whether the "DNS name compression" mode is enabled or not.
+ *
+ * This is intended for testing only and available when `OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE` config is enabled.
+ *
+ * @returns TRUE if the "DNS name compression" mode is enabled, FALSE otherwise.
+ *
+ */
+bool otDnsIsNameCompressionEnabled(void);
 
 /**
  * @}
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (92)
+#define OPENTHREAD_API_VERSION (93)
 
 /**
  * @addtogroup api-instance
--- a/src/cli/README.md
+++ b/src/cli/README.md
@@ -991,6 +991,33 @@ Send a service instance resolution DNS query for a given service instance. Servi
 
 The parameters after `service-name` are optional. Any unspecified (or zero) value for these optional parameters is replaced by the value from the current default config (`dns config`).
 
+### dns compression \[enable|disable\]
+
+Enable/Disable the "DNS name compression" mode.
+
+By default DNS name compression is enabled. When disabled, DNS names are appended as full and never compressed. This is applicable to OpenThread's DNS and SRP client/server modules.
+
+This is intended for testing only and available under `REFERENCE_DEVICE` config.
+
+Get the current "DNS name compression" mode.
+
+```
+> dns compression
+Enabled
+```
+
+Set the "DNS name compression" mode.
+
+```
+> dns compression disable
+Done
+>
+>
+> dns compression
+Disabled
+Done
+```
+
 ### domainname
 
 Get the Thread Domain Name for Thread 1.2 device.
--- a/src/core/common/instance.hpp
+++ b/src/core/common/instance.hpp
@@ -283,6 +283,29 @@ public:
     Coap::CoapSecure &GetApplicationCoapSecure(void) { return mApplicationCoapSecure; }
 #endif
 
+#if OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE
+    /**
+     * This method enables/disables the "DNS name compressions" mode.
+     *
+     * By default DNS name compression is enabled. When disabled, DNS names are appended as full and never compressed.
+     * This is applicable to OpenThread's DNS and SRP client/server modules.
+     *
+     * This is intended for testing only and available under a `REFERENCE_DEVICE` config.
+     *
+     * @param[in] aEnabled   TRUE to enable the "DNS name compression" mode, FALSE to disable.
+     *
+     */
+    static void SetDnsNameCompressionEnabled(bool aEnabled) { sDnsNameCompressionEnabled = aEnabled; }
+
+    /**
+     * This method indicates whether the "DNS name compression" mode is enabled or not.
+     *
+     * @returns TRUE if the "DNS name compressions" mode is enabled, FALSE otherwise.
+     *
+     */
+    static bool IsDnsNameCompressionEnabled(void) { return sDnsNameCompressionEnabled; }
+#endif
+
 #endif // OPENTHREAD_MTD || OPENTHREAD_FTD
 
     /**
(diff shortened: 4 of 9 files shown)


src/core/common/instance.hpp
```diff
@@ -395,7 +418,10 @@ class Instance : public otInstance, private NonCopyable
 #if OPENTHREAD_CONFIG_DIAG_ENABLE
     FactoryDiags::Diags mDiags;
 #endif
-    bool mIsInitialized;
+    bool mIsInitialized : 1;
+#if OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE && (OPENTHREAD_FTD || OPENTHREAD_MTD)
+    bool mUncompressedDnsName : 1;
```
simonlingoogle, line 423: I guess this does not work in `MULTIPLE_INSTANCE` because we can't call `Instance::Get().IsUncompressedDnsNameEnabled()` in `MULTIPLE_INSTANCE`. 

I am fine with this feature not working in MULTIPLE_INSTANCE, but can we just make this field static so that we don't need to explicitly specify that this feature does not work in `MULTIPLE_INSTANCE`.  

```suggestion
    static bool mUncompressedDnsName : 1;
```

simonlingoogle (approved):
LGTM 👍 

Implementing this feature in `AppendPointerLabel` looks wonderful.

src/core/common/instance.hpp
abtink: I was also considering the same (since defining `mUncompressedDnsName` as `static` member variable fits well here) but decided not to go for it.
 
I think unless we have a very good reason it would be good to avoid using `static` member variable in any of OT classes (and mainly instance) so that `sizeof(ot::Instance)` can give us  the OT RAM/memory would use.

simonlingoogle: > `sizeof(ot::Instance)` can give us the OT RAM/memory would use.

Any elaboration on why this is useful?

abtink: In general, in terms of sw architecture, it is good practise to have all variables (memory use) for OT library to follow the same pattern.

One use-case is to be able to easily tell how much RAM OT lib is using when OT is integrated in a production sw (with a selected set of production/project configs). The `static` sort of hides the memory use by the OT library.

We have this related point in our style guide:

> - Global and Scoped Static Construction
>   - There shall be no use of global, static or otherwise, object construction. The use of scoped static object construction should be avoided.
>

include/openthread/dns.h
```diff
@@ -140,6 +141,34 @@ void otDnsInitTxtEntryIterator(otDnsTxtEntryIterator *aIterator, const uint8_t *
...
+ * DNS and SRP client/server modules.
+ *
+ * This is intended for testing only and available when `OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE` config is enabled
+ * and `OPENTHREAD_CONFIG_MULTIPLE_INSTANCE_ENABLE` is disabled.
+ *
+ * @param[in] aInstance    A pointer to an OpenThread instance.
+ * @param[in] aEnabled     TRUE to enable the "uncompressed DNS name" mode, FALSE to disable.
+ *
+ */
+void otDnsSetUncompressedDnsNameEnabled(otInstance *aInstance, bool aEnabled);
```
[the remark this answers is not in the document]
abtink: Thanks. Good suggestion. Makes sense. 👍

src/core/net/dns_types.cpp
```diff
@@ -221,20 +222,38 @@ Error Name::AppendTerminator(Message &aMessage)
...
     // an offset value relative to start of DNS header.
 
-    uint16_t value;
-
     OT_ASSERT(aOffset < kPointerLabelTypeUint16);
 
     value = HostSwap16(aOffset | kPointerLabelTypeUint16);
 
-    return aMessage.Append(value);
+    ExitNow(error = aMessage.Append(value));
```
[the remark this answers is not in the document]
abtink: From code flow, no it is not necessary. Added it so that when `REFERENCE_DEVICE` config is not enabled, we don't get the error that the `exit` label is not used. 
- Alternative is to add `#if` around the `exit:` label (which looks kind of ugly).  
- Or another way is to not use `exit:` at all and have the whole block in an  `else` for the `if (Instance::Get().IsUncompressedDnsNameEnabled())`.

I felt adding `ExitNow()` at the end may be the cleanest way, but I don't have any strong preferences here. 

@jwhui, what do you think?

src/core/common/instance.hpp
simonlingoogle: I would agree on the principle. However, we don't turn on `REFERENCE_DEVICE` for production sw. So, defining static variables for `REFERENCE_DEVICE` is actually fine?

src/core/net/dns_types.cpp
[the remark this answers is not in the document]
jwhui: Ah, yes, that makes sense. LGTM, thanks!

src/core/common/instance.hpp
abtink: It sets a precedence for use of `static` and I think we should have a good justification/advantage when we do so.

In the case here the main advantage is that we we want to limit the definition and more importantly the  use of the related API under `MULTIPLE_INSTANCE`. We can already do so by adding proper `#if` checks (which we do in the `dns_api.cpp` in this PR).

simonlingoogle: > It sets a precedence for use of static and I think we should have a good justification/advantage when we do so.

We already did that:
```c++
#if !OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE
    static Utils::Heap sHeap;
#endif
```

> In the case here the main advantage is that we we want to limit the definition and more importantly the use of the related API under `MULTIPLE_INSTANCE`. 

Limiting the definition and use of related APIs under `MULTIPLE_INSTANCE` seems unnecessary to me. If less code can just work for `MULTIPLE_INSTANCE` and non `MULTIPLE_INSTANCE`, why should we add limitations with more code?

---

That being said. It's a minor issue for me. But just prefer to make it also usable in `MULTIPLE_INSTANCE` when there is no obvious reason why not.

abtink: I changed this in last push.
Somehow I did not notice that with static we can get it to work under `MULTIPLE_INSTANCE` too. I think that may be good enough reason to allow it here.

include/openthread/dns.h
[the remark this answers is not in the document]
abtink: Changed in the latest pushed commit.

include/openthread/dns.h
```diff
@@ -140,6 +141,32 @@ void otDnsInitTxtEntryIterator(otDnsTxtEntryIterator *aIterator, const uint8_t *
...
+ *
+ */
+void otDnsSetNameCompressionEnabled(bool aEnabled);
+
+/**
+ * This function indicates whether the "DNS name compression" mode is enabled or not.
+ *
+ * This is intended for testing only and available when `OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE` config is enabled.
+ *
+ * @returns TRUE if the "DNS name compression" mode is enabled, FALSE otherwise.
```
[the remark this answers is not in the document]
abtink: I personally prefer `@returns` for simple boolean return value (and using `FALSE otherwise`). It avoids having two lines and then repeating almost the same statement again in a separate line.

include/openthread/instance.h
```diff
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (92)
+#define OPENTHREAD_API_VERSION (93)
```
abtink, line 56: @jwhui I think you updated the version, right? Thank you 😃

include/openthread/dns.h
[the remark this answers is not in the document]
simonlingoogle: SGTM

src/core/net/dns_types.cpp
```diff
@@ -221,20 +222,38 @@ Error Name::AppendTerminator(Message &aMessage)
 
 Error Name::AppendPointerLabel(uint16_t aOffset, Message &aMessage)
 {
+    Error    error;
+    uint16_t value;
+
+#if OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE
+    if (!Instance::IsDnsNameCompressionEnabled())
```
wgtdkp, line 229: Should the `IsDnsNameCompressionEnabled` be defined in the `Dns::Name` class? And the method name can be shorter I think :)

wgtdkp (approved):
LGTM :+1:

src/core/net/dns_types.cpp
abtink: I was thinking of the same (I agree the method name would be nicer/shorter), but I thought that it would be better to keep the `static` member definitions in `Instance` so we can more easily track them. 

My view is that we should avoid `static` member definitions in core module unless there is a good reason and if we use them we should aim to keep them in one place/module (`Instance` makes sense) to make easy to track them. Otherwise the `static` member definition can be added/hidden is some some type and harder to track.

Thoughts?

wgtdkp: Thanks, makes sense to me!

src/cli/README.md
```diff
@@ -991,6 +991,33 @@ Send a service instance resolution DNS query for a given service instance. Servi
 
 The parameters after `service-name` are optional. Any unspecified (or zero) value for these optional parameters is replaced by the value from the current default config (`dns config`).
 
+### dns compression \[enable|disable\]
+
+Enable/Disable the "DNS name compression" mode.
+
+By default DNS name compression is enabled. When enabled, DNS names are appended as full and never compressed. This is applicable to OpenThread's DNS and SRP client/server modules.
```
jwhui, line 998:
```suggestion
By default DNS name compression is enabled. When disabled, DNS names are appended as full and never compressed. This is applicable to OpenThread's DNS and SRP client/server modules.
```

jwhui (approved):

PR #6369, opened 2021-03-30
[samr21] move to openthread/ot-samr21

jwhui:


simonlingoogle (approved):
LGTM :+1: 

Great to see OpenThread getting slim.

PR #6370, opened 2021-03-30
[ping-sender] let ping return statistics in OTCI

superwhd:
Ping enhancements
- If it's not multicast, `ping` command in CLI will print `Done` in the end, after printing replies and statistics. The old behavior is, `ping` command prints `Done` before any replies.
- In OTCI, let `ping` function return statistics.
- Support timeout parameter in `ping` command.
- Fix a bug that the arguments are not properly passed to `ping` in `node.py`.
- Adjust timeouts in tests.

### DIFF
--- a/include/openthread/ping_sender.h
+++ b/include/openthread/ping_sender.h
@@ -115,10 +115,10 @@ typedef struct otPingSenderConfig
     uint16_t mSize;               ///< Data size (# of bytes) excludes IPv6/ICMPv6 header. Zero for default.
     uint16_t mCount;              ///< Number of ping messages to send. Zero to use default.
     uint32_t mInterval;           ///< Ping tx interval in milliseconds. Zero to use default.
-    uint16_t mTimeout; ///< Time in milliseconds to wait for final reply after sending final request.
-                       ///< Zero to use default.
-    uint8_t mHopLimit; ///< Hop limit (used if `mAllowZeroHopLimit` is false). Zero for default.
-    bool    mAllowZeroHopLimit; ///< Indicates whether hop limit is zero.
+    uint16_t mTimeout;            ///< Time in milliseconds to wait for final reply after sending final request.
+                                  ///< Zero to use default.
+    uint8_t mHopLimit;            ///< Hop limit (used if `mAllowZeroHopLimit` is false). Zero for default.
+    bool    mAllowZeroHopLimit;   ///< Indicates whether hop limit is zero.
 } otPingSenderConfig;
 
 /**
--- a/include/openthread/ping_sender.h
+++ b/include/openthread/ping_sender.h
@@ -115,10 +115,10 @@ typedef struct otPingSenderConfig
     uint16_t mSize;               ///< Data size (# of bytes) excludes IPv6/ICMPv6 header. Zero for default.
     uint16_t mCount;              ///< Number of ping messages to send. Zero to use default.
     uint32_t mInterval;           ///< Ping tx interval in milliseconds. Zero to use default.
-    uint16_t mTimeout;            ///< Time in milliseconds to wait for a reply after sending out the request.
-                                  ///< Zero to use default.
-    uint8_t mHopLimit;            ///< Hop limit (used if `mAllowZeroHopLimit` is false). Zero for default.
-    bool    mAllowZeroHopLimit;   ///< Indicates whether hop limit is zero.
+    uint16_t mTimeout; ///< Time in milliseconds to wait for the final echo reply after sending out the request. Zero to
+                       ///< use default.
+    uint8_t mHopLimit; ///< Hop limit (used if `mAllowZeroHopLimit` is false). Zero for default.
+    bool    mAllowZeroHopLimit; ///< Indicates whether hop limit is zero.
 } otPingSenderConfig;
 
 /**
--- a/include/openthread/ping_sender.h
+++ b/include/openthread/ping_sender.h
@@ -115,8 +115,8 @@ typedef struct otPingSenderConfig
     uint16_t mSize;               ///< Data size (# of bytes) excludes IPv6/ICMPv6 header. Zero for default.
     uint16_t mCount;              ///< Number of ping messages to send. Zero to use default.
     uint32_t mInterval;           ///< Ping tx interval in milliseconds. Zero to use default.
-    uint16_t mTimeout; ///< Time in milliseconds to wait for the final echo reply after sending out the request. Zero to
-                       ///< use default.
+    uint16_t mTimeout; ///< Time in milliseconds to wait for final reply after sending final request.
+                       ///< Zero to use default.
     uint8_t mHopLimit; ///< Hop limit (used if `mAllowZeroHopLimit` is false). Zero for default.
     bool    mAllowZeroHopLimit; ///< Indicates whether hop limit is zero.
 } otPingSenderConfig;
--- a/src/cli/README.md
+++ b/src/cli/README.md
@@ -1882,7 +1882,7 @@ Set the preferred Thread Leader Partition ID.
 Done
 ```
 
-### ping \<ipaddr\> [size][count] [interval][hoplimit]
+### ping \<ipaddr\> [size][count] [interval][hoplimit] [timeout]
 
 Send an ICMPv6 Echo Request.
 
@@ -1890,6 +1890,7 @@ Send an ICMPv6 Echo Request.
 - count: The number of ICMPv6 Echo Requests to be sent.
 - interval: The interval between two consecutive ICMPv6 Echo Requests in seconds. The value may have fractional form, for example `0.5`.
 - hoplimit: The hoplimit of ICMPv6 Echo Request to be sent.
+- timeout: Time in seconds to wait for the final ICMPv6 Echo Reply after sending out the request. The value may have fractional form, for example `3.5`.
 
 ```bash
 > ping fdde:ad00:beef:0:558:f56b:d688:799
--- a/src/cli/README.md
+++ b/src/cli/README.md
@@ -1882,7 +1882,7 @@ Set the preferred Thread Leader Partition ID.
 Done
 ```
 
-### ping \<ipaddr\> [size] [count] [interval] [hoplimit] [timeout]
+### ping \<ipaddr\> \[size\] \[count\] \[interval\] \[hoplimit\] \[timeout\]
 
 Send an ICMPv6 Echo Request.
 
--- a/src/cli/README.md
+++ b/src/cli/README.md
@@ -1882,7 +1882,7 @@ Set the preferred Thread Leader Partition ID.
 Done
 ```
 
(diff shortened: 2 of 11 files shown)


tools/otci/otci/otci.py
```diff
@@ -175,30 +175,41 @@ def factory_reset(self):
...
-            cmd += f' {hoplimit}'
-
-        self.execute_command(cmd)
+        cmd = f'ping {ip} {size} {count} {interval} {hoplimit} {timeout}'
+
+        timeout_allowance = 3
+        lines = self.execute_command(cmd, timeout=(count - 1) * interval + timeout + timeout_allowance)
+        packet_count_pattern = re.compile(r'(\d+) packets transmitted, (\d+) packets received.')
+        packet_loss_pattern = re.compile(r'Packet loss = (\d+\.\d+)%')
+        round_trip_time = re.compile(r'Round-trip min/avg/max = (\d+)/(\d+\.\d+)/(\d+) ms')
```
simonlingoogle, line 194: better compile regex just once, e.g. as a class member of OTCI

e.g. 
```python
_PING_ROUND_TRIP_TIME_PATTERN = re.compile(r'Round-trip min/avg/max = (\d+)/(\d+\.\d+)/(\d+) ms')
```

simonlingoogle, line 194: Are these statistics separated in multiple lines?

tools/otci/otci/otci.py
```diff
@@ -175,30 +175,41 @@ def factory_reset(self):
     # Network Operations
     #
 
-    def ping(self, ip: str, size: int = None, count: int = None, interval: int = None, hoplimit: int = None):
-        """Send an ICMPv6 Echo Request.
+    def ping(self, ip: str, size: int = 8, count: int = 1, interval: float = 1, hoplimit: int = 64, timeout: float = 3):
```
simonlingoogle, line 178:
```suggestion
    def ping(self, ip: str, size: int = 8, count: int = 1, interval: float = 1, hoplimit: int = 64, timeout: float = 3) -> Dict:
```

tools/otci/tests/test_otci.py
```diff
@@ -402,7 +402,14 @@ def _test_otci_multi_nodes(self, leader, commissioner, child1, child2):
...
         for dst_ip in leader.get_ipaddrs():
-            commissioner.ping(dst_ip, size=10, count=1, interval=2, hoplimit=3)
+            statistics = commissioner.ping(dst_ip, size=10, count=10, interval=2, hoplimit=3)
+            self.assertTrue('transmitted_packets' in statistics)
+            self.assertTrue('received_packets' in statistics)
+            self.assertTrue('packet_loss' in statistics)
+            self.assertTrue('round_trip_time' in statistics)
+            self.assertTrue('min' in statistics['round_trip_time'])
+            self.assertTrue('avg' in statistics['round_trip_time'])
+            self.assertTrue('max' in statistics['round_trip_time'])
```
simonlingoogle, line 412: Can we verify whether the statistics are correct?

tools/otci/otci/otci.py
superwhd: No. They are in the same line.

simonlingoogle: If so we may use one line of regex.

superwhd: My concern is, the regex then may be too complex. I can try merging them though.

superwhd: Updated.

tools/otci/tests/test_otci.py
superwhd: I don't have much idea about how to do this verification, especially for the RTTs.

superwhd: Per offline discussion, added some basic validations.

tools/otci/tests/test_otci.py
```diff
@@ -402,7 +402,19 @@ def _test_otci_multi_nodes(self, leader, commissioner, child1, child2):
...
+            self.assertEqual(statistics['transmitted_packets'], 10)
+            self.assertTrue('received_packets' in statistics)
+            self.assertEqual(statistics['received_packets'], 10)
+            self.assertTrue('packet_loss' in statistics)
+            self.assertAlmostEqual(statistics['packet_loss'], 0.0, delta=1e-9)
+            self.assertTrue('round_trip_time' in statistics)
+            rtt = statistics['round_trip_time']
+            self.assertTrue('min' in rtt)
+            self.assertTrue('avg' in rtt)
+            self.assertTrue('max' in rtt)
```
simonlingoogle, line 416:
```suggestion
```
not necessary because the next statement already guarantee that these three fields must exist in rtt.

simonlingoogle (approved):
LGTM :+1:

src/cli/README.md
```diff
@@ -1855,14 +1855,15 @@ Set the preferred Thread Leader Partition ID.
 Done
 ```
 
-### ping \<ipaddr\> [size][count] [interval][hoplimit]
+### ping \<ipaddr\> [size][count] [interval][hoplimit] [timeout]
```
jwhui, line 1858:
```suggestion
### ping \<ipaddr\> [size] [count] [interval] [hoplimit] [timeout]
```

include/openthread/ping_sender.h
```diff
@@ -115,10 +115,10 @@ typedef struct otPingSenderConfig
     uint16_t mSize;               ///< Data size (# of bytes) excludes IPv6/ICMPv6 header. Zero for default.
     uint16_t mCount;              ///< Number of ping messages to send. Zero to use default.
     uint32_t mInterval;           ///< Ping tx interval in milliseconds. Zero to use default.
-    uint16_t mTimeout;            ///< Time in milliseconds to wait for a reply after sending out the request.
-                                  ///< Zero to use default.
-    uint8_t mHopLimit;            ///< Hop limit (used if `mAllowZeroHopLimit` is false). Zero for default.
-    bool    mAllowZeroHopLimit;   ///< Indicates whether hop limit is zero.
+    uint16_t mTimeout; ///< Time in milliseconds to wait for the final echo reply after sending out the request. Zero to
+                       ///< use default.
```
jwhui, line 119: nit: try to shorten so that the comments remain aligned?
```suggestion
    uint16_t mTimeout; ///< Time in milliseconds to wait for final reply after sending final request.
                       ///< Zero to use default.
```

src/core/utils/ping_sender.cpp
```diff
@@ -108,6 +108,10 @@ Error PingSender::Ping(const Config &aConfig)
 
     mStatistics.Clear();
     mStatistics.mIsMulticast = static_cast<Ip6::Address *>(&mConfig.mDestination)->IsMulticast();
+    if (mStatistics.mIsMulticast)
```
jwhui, line 111: Should we avoid special-casing multicast?

superwhd: Updated. Note that multicast ping will wait for a full timeout for the final echo request, regardless of how many replies it has received.

src/cli/cli.cpp
```diff
@@ -3202,7 +3202,8 @@ void Interpreter::HandlePingStatistics(const otPingSenderStatistics *aStatistics
 {
     OutputFormat("%u packets transmitted, %u packets received.", aStatistics->mSentCount, aStatistics->mReceivedCount);
 
-    if ((aStatistics->mSentCount != 0) && !aStatistics->mIsMulticast)
+    if ((aStatistics->mSentCount != 0) && !aStatistics->mIsMulticast &&
+        aStatistics->mReceivedCount <= aStatistics->mSentCount)
```
simonlingoogle, line 3206:
```suggestion
    if (aStatistics->mSentCount != 0)
```

Should we treat `mReceivedCount` > `mSentCount` as if it's `mSentCount`? (i.e. 0% loss)
and avoid special-casing mutlicast ?

src/core/utils/ping_sender.cpp
```diff
@@ -199,6 +199,7 @@ void PingSender::HandleIcmpReceive(const Message &          aMessage,
 {
     Reply    reply;
     uint32_t timestamp;
+    bool timerWasRunning;
```
simonlingoogle, line 202:
```suggestion

        VerifyOrExit(mTimer.IsRunning())
```

`PingSender` is not expecting any ping reply if `mTimer` is not running ?

src/cli/cli.cpp
superwhd: I think `mReceivedCount > mSentCount` doesn't imply 0% packet loss. So far I can think of two cases:
- In unicast case, when something's not working properly, the destination sends 2 replies for every request it receives. Then the `mReceivedCount` will be twice as much as `mSentCount`, assuming there's no packet loss. However, even if there are some packet lost, the `mReceivedCount` could still be greater than `mSentCount`.

src/core/utils/ping_sender.cpp
superwhd: After timer stops, PingSender can still handle incoming echo replies and print its RTT, but it won't print statistics.

simonlingoogle: > After timer stops, PingSender can still handle incoming echo replies and print its RTT, but it won't print statistics.

But these echo replies are unlikely for echo requests sent by `PingSender` since timeout has expired.

src/cli/cli.cpp
simonlingoogle: I think we should try to print the `Packet loss` as long as it's possible. 

> In unicast case, when something's not working properly, the destination sends 2 replies for every request it receives.

I think it's an abnormal case that we need to fix. `mReceivedCount` should not be greater than `mSendCount` for unicast ping.

> In multicast case, when there are two nodes receiving echo requests, the mReceivedCount could become greater than mSentCount easily when the packet loss rate is < 50%.

It would be better if we can handle the multicast case correctly by showing duplicates., e.g. like below:
```
$ ping 224.0.0.1
PING 224.0.0.1 (224.0.0.1) 56(84) bytes of data.
64 bytes from 100.84.187.252: icmp_seq=1 ttl=64 time=0.578 ms
64 bytes from 100.84.187.253: icmp_seq=1 ttl=64 time=0.627 ms
64 bytes from 100.84.187.253: icmp_seq=2 ttl=64 time=0.560 ms
64 bytes from 100.84.187.252: icmp_seq=2 ttl=64 time=0.560 ms
^C
--- 224.0.0.1 ping statistics ---
2 packets transmitted, 2 received, +2 duplicates, 0% packet loss, time 1019ms
rtt min/avg/max/mdev = 0.560/0.581/0.627/0.027 ms
```

But I am fine to leave for future enhancement.

src/core/utils/ping_sender.cpp
superwhd: Done

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6372, opened 2021-03-30
[ip6] improve frame filtering

MarekPorwisz:
Some packets used by openthread internally were passed outside with
filters on. This fixes the issue with coap frames being passed
outside when OpenThread CoAP API is used.
Fixes "ICMPv6 Destination Unreachable (Port unreachable)" in zephyr.

### DIFF
--- a/include/openthread/udp.h
+++ b/include/openthread/udp.h
@@ -290,6 +290,18 @@ void otUdpForwardReceive(otInstance *        aInstance,
                          uint16_t            aSockPort);
 
 /**
+ * Determines if the given UDP port is used by OpenThread, or belongs entirely to the platform
+ *
+ * @param[in]  aInstance            A pointer to an OpenThread instance.
+ * @param[in]  port                 UDP port number to verify.
+ *
+ * @retval true    The port is being used by OpenThread
+ * @retval false   The port is not used by OpenThread i.e.it belongs to the platform or is incorrect.
+ *
+ */
+bool otUdpIsPortInUse(otInstance *aInstance, uint16_t port);
+
+/**
  * @}
  *
  */
--- a/src/core/net/ip6.cpp
+++ b/src/core/net/ip6.cpp
@@ -1035,7 +1035,7 @@ Error Ip6::ProcessReceiveCallback(Message &          aMessage,
             Udp::Header udp;
 
             IgnoreError(aMessage.Read(aMessage.GetOffset(), udp));
-            VerifyOrExit(Get<Udp>().ShouldUsePlatformUdp(udp.GetDestinationPort()), error = kErrorNoRoute);
+            VerifyOrExit(Get<Udp>().IsPortInUse(udp.GetDestinationPort()) == false, error = OT_ERROR_NO_ROUTE);
 
             break;
         }
--- a/src/core/net/udp6.cpp
+++ b/src/core/net/udp6.cpp
@@ -524,6 +524,18 @@ exit:
     return;
 }
 
+bool Udp::IsPortInUse(uint16_t aPort) const
+{
+    for (const SocketHandle * socket = mSockets.GetHead(); socket != nullptr; socket=socket->GetNext())
+    {
+        if(socket->GetSockName().GetPort() == aPort)
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
 bool Udp::ShouldUsePlatformUdp(uint16_t aPort) const
 {
     return (aPort != Mle::kUdpPort && aPort != Tmf::kUdpPort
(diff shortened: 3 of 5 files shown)


src/core/net/udp6.cpp
```diff
@@ -524,6 +524,18 @@ void Udp::HandlePayload(Message &aMessage, MessageInfo &aMessageInfo)
...
+bool Udp::IsPortInUse(uint16_t aPort) const
+{
+    for (const SocketHandle * socket = mSockets.GetHead(); socket != nullptr; socket=socket->GetNext())
+    {
+        if(socket->GetSockName().GetPort() == aPort)
+        {
+            return true;
+        }
+    }
+    return false;
```
simonlingoogle, line 536: style: OpenThread always use one return in each method.

src/core/net/udp6.hpp
```diff
@@ -573,6 +573,18 @@ class Udp : public InstanceLocator, private NonCopyable
...
 #endif
 
+    /**
+     * This method returns whether a udp port is being used by OpenThread or any of it's optional
+     * features, e.g. CoAP API.
+     *
+     * @param[in]   aPort       The udp port
+     *
+     * @retval True when port is used by the OpenThread
+     * @retval False when the port is not used by OpenThrad.
```
simonlingoogle, line 583:
```suggestion
     * @retval False when the port is not used by OpenThread.
```

src/core/net/udp6.hpp
```diff
@@ -573,6 +573,18 @@ class Udp : public InstanceLocator, private NonCopyable
     }
 #endif
 
+    /**
+     * This method returns whether a udp port is being used by OpenThread or any of it's optional
+     * features, e.g. CoAP API.
+     *
+     * @param[in]   aPort       The udp port
```
simonlingoogle, line 580:
```suggestion
     * @param[in]   aPort       The UDP port
```

simonlingoogle (commented):
I am afraid this PR is not a correct fix of your problem. @MarekPorwisz 

Could you explain more on the issue?

include/openthread/udp.h
```diff
@@ -289,6 +289,18 @@ void otUdpForwardReceive(otInstance *        aInstance,
...
+ * Determines if the given UDP port is used by OpenThread, or belongs entirely to the platform
+ *
+ * @param[in]  aInstance            A pointer to an OpenThread instance.
+ * @param[in]  port                 UDP port number to verify.
+ *
+ * @retval true    The port is being used by OpenThread
+ * @retval false   The port is not used by OpenThread i.e.it belongs to the platform or is incorrect.
+ *
+ */
+bool otUdpIsPortInUse(otInstance *aInstance, uint16_t port);
```
wgtdkp, line 302: Are there usages of this public API?

src/core/net/ip6.cpp
```diff
@@ -1035,7 +1035,7 @@ Error Ip6::ProcessReceiveCallback(Message &          aMessage,
             Udp::Header udp;
 
             IgnoreError(aMessage.Read(aMessage.GetOffset(), udp));
-            VerifyOrExit(Get<Udp>().ShouldUsePlatformUdp(udp.GetDestinationPort()), error = kErrorNoRoute);
+            VerifyOrExit(Get<Udp>().IsPortInUse(udp.GetDestinationPort()) == false, error = OT_ERROR_NO_ROUTE);
```
wgtdkp, line 1038: I think this will not work with the Border Agent. It listens on `49191` for both Thread interface and platform interfaces and it expect packets from both Thread network and external network.

MarekPorwisz:
HAHA, That's a Draft PR :) I didn't expect anyone to look into it. I'm still looking for a proper solution to our issue.
I had problems running tests locally so I created draft PR to run tests on server.

The issue comes from Zephyr and NCS:
The otIp6SetReceiveCallback is used to pass all the frames to Zephyr's IP stack. The filtering filters Thread control frames nicely, but it does not filter anything else.
When OpenThread CoAP API is used, a socket is allocated in OpenThread, but the frames are passed to Zephyr as well and since the socket is not open on zephyr's side we get "ICMPv6 Destination Unreachable (Port unreachable)".
I was thinking about calling IsPortInUse from zephyr after frame is received, but I the place I could call this zephyr did not yet parse the packet, and OT does not pass the protocol and port to the callback function.
Now I am trying the approach with additional filter option, so the API will allow to enable filtering only control msgs, or everything that is used by OpenThread. I'm still not sure this will fix all the problems but at least should not break existing functionalities.

MarekPorwisz:
closing for now

PR #6373, opened 2021-03-30
[cmake] allow custom values for package name and package version

suveshpratapa:
@jwhui Let us know if you're okay with this small change.


CMakeLists.txt
```diff
@@ -94,10 +94,13 @@ execute_process(
 )
```
bukepo, line 94: It seems not necessary to compute `OT_VERSION` when `OT_PACKAGE_VERSION` is defined? Should we remove guard this expression?

etc/cmake/options.cmake
```diff
@@ -317,6 +317,16 @@ if(OT_UDP_FORWARD)
     target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_UDP_FORWARD_ENABLE=1")
 endif()
 
+option(OT_PACKAGE_NAME "enable vendor package name")
```
wgtdkp, line 320: Should we use `variable` but not `option`? `option` is designed for boolean switches https://cmake.org/cmake/help/v3.10/command/option.html#option.

And better to have `set(OT_PACKAGE_NAME ...)` at beginning of `CMakeListstxt` but not this `options.cmake` file.

CMakeLists.txt
suveshpratapa: Fixed! Thanks.

etc/cmake/options.cmake
suveshpratapa: Should be fixed now, thanks.

wgtdkp (commented):
LGTM :+1: 
Thanks!

wgtdkp (approved):

CMakeLists.txt
```diff
@@ -87,17 +87,19 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "(Apple)?[Cc]lang")
...
+set(OT_PACKAGE_VERSION "" CACHE STRING "OpenThread Package Version")
+if(OT_PACKAGE_VERSION STREQUAL "")
+    execute_process(
+        COMMAND bash "-c" "third_party/nlbuild-autotools/repo/scripts/mkversion -b ${OT_DEFAULT_VERSION}"
+        WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+        OUTPUT_VARIABLE OT_PACKAGE_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    message(STATUS "Setting default package version: ${OT_PACKAGE_VERSION}")
+endif()
+target_compile_definitions(ot-config INTERFACE "PACKAGE_VERSION=\"${OT_PACKAGE_VERSION}\"")
```
bukepo, line 102: print the version?
```
message(STATUS "Version: ${OT_PACKAGE_VERSION}")
```

bukepo (approved):

CMakeLists.txt
suveshpratapa: Fixed now, thanks.

jwhui (approved):

PR #6378, opened 2021-03-31
[ot-ctl] fix ot-ctl hang

bukepo:
Fixes #6377 
Fixes #6376 
Fixes #6355

PR #6379, opened 2021-03-31
[ot-ctl] fix wrong exit in interactive mode 

simonlingoogle:
This commit fixes ot-ctl exiting in interactive mode, which was introduced by #6333, which changed the scope of variables wrongly.


jwhui:
@simonlingoogle , is this a dup of #6378?

simonlingoogle:
Duplicated by #6378

PR #6390, opened 2021-04-01
[spinel] use full names in string lookup table

abtink:
This commit changes the string lookup tables used in `spinel.c` for to 
use the full enum names in the table.

This makes the names "search/grep"-able which is helpful during
renaming/refactoring or when adding a new property (by searching and
copying code for an existing one).


bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #6395, opened 2021-04-02
[data-poll-sender] use OT_MIN and replace enum with constexpr

Irving-cl:
This PR replaces some enum definition with constexpr since these enum
are used for comparing. The compiler would always give warnings when 
comparing enum with other non-enumeral values. With this change, we 
can directly compare the consts with other values. And this PR replaces
`UpdateIfLarger` with `OT_MIN`.

### DIFF
--- a/src/core/mac/data_poll_sender.cpp
+++ b/src/core/mac/data_poll_sender.cpp
@@ -200,7 +200,7 @@ uint32_t DataPollSender::GetKeepAlivePollPeriod(void) const
 
     if (mExternalPollPeriod != 0)
     {
-        UpdateIfLarger(period, mExternalPollPeriod);
+        period = OT_MIN(period, mExternalPollPeriod);
     }
 
     return period;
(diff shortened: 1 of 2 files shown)


src/core/mac/data_poll_sender.cpp
```diff
@@ -200,7 +200,7 @@ uint32_t DataPollSender::GetKeepAlivePollPeriod(void) const
 
     if (mExternalPollPeriod != 0)
     {
-        UpdateIfLarger(period, mExternalPollPeriod);
+        period = OT_MAX(period, mExternalPollPeriod);
```
brianljt, line 203:
```suggestion
        period = OT_MIN(period, mExternalPollPeriod);
```

src/core/mac/data_poll_sender.hpp
```diff
@@ -285,11 +275,21 @@ class DataPollSender : public InstanceLocator, private NonCopyable
...
+    // Poll period under different conditions (in milliseconds).
+    static constexpr uint32_t kAttachDataPollPeriod =
+        OPENTHREAD_CONFIG_MAC_ATTACH_DATA_POLL_PERIOD; ///< Poll period during attach.
+    static constexpr uint32_t kRetxPollPeriod =
+        OPENTHREAD_CONFIG_MAC_RETX_POLL_PERIOD;      ///< Poll retx period due to tx failure.
+    static constexpr uint32_t kFastPollPeriod = 188; ///< Period used for fast polls.
+    static constexpr uint32_t kMinPollPeriod =
+        OPENTHREAD_CONFIG_MAC_MINIMUM_POLL_PERIOD;                  ///< Minimum allowed poll period.
+    static constexpr uint32_t kMaxExternalPeriod = ((1 << 26) - 1); ///< Maximum allowed user-specified period.
+                                                                    ///< i.e. (0x3FFFFF)ms, about 18.64 hours.
```
abtink, line 287: How about removing/shortening the comments to make it align and look better. 

Here is my suggestion (which is removing most of the comments since they are repeating what can be inferred from the constant variable name itself).
```suggestion
    static constexpr uint32_t kAttachDataPollPeriod = OPENTHREAD_CONFIG_MAC_ATTACH_DATA_POLL_PERIOD;
    static constexpr uint32_t kRetxPollPeriod       = OPENTHREAD_CONFIG_MAC_RETX_POLL_PERIOD;
    static constexpr uint32_t kFastPollPeriod       = 188;
    static constexpr uint32_t kMinPollPeriod        = OPENTHREAD_CONFIG_MAC_MINIMUM_POLL_PERIOD;
    static constexpr uint32_t kMaxExternalPeriod    = ((1 << 26) - 1);  // ~18.6 hours.
```

abtink (approved):
LGTM. Thanks
- Need the change which @brianljt commented on use of `OT_MIN` instead of `OT_MAX`.
- Also have a suggestion belwo on removing some comments to make code align and look nicer

brianljt (approved):
LGTM. Thanks!

jwhui (approved):

PR #6396, opened 2021-04-02
[meshcop] fix UDP proxy

wgtdkp:
The MeshCoP UDP Proxy packets should not be sent/received with platform UDP.

Depends-On: openthread/ot-commissioner#185


bukepo:
From https://github.com/openthread/openthread/blob/main/src/core/net/udp6.hpp#L64, these two options will not be enabled together.

wgtdkp:
> From https://github.com/openthread/openthread/blob/main/src/core/net/udp6.hpp#L64, these two options will not be enabled together.

This commit is for `UDP proxy (UDP_TX.ntf and UDP_RX.ntf)` which encapsulates packets between arbitrary Thread end device and the external commisioner, but not `UDP forward`.

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #6401, opened 2021-04-03
[announce-sender] use trickle to limit number of announce tx

abtink:
This commit updates and enhances the `AnnounceSender` feature. When
this feature is enabled, routers and REEDs periodically send MLE
Announcements (on all channels from the channel mask in the current
Active Operational Dataset). This commit updates the implementation to
use `TrickleTimer` such that a device monitors the number of received
Announce messages within a cycle and if the number is above a
redundancy threshold, the device skips sending Announcement in that
cycle itself. This helps limit the number of Announcement tx in cycle
in a dense network. The same mechanism is used by both routers and
REEDs.


src/core/thread/announce_sender.hpp
```diff
@@ -95,23 +111,76 @@ class AnnounceSenderBase : public InstanceLocator, private NonCopyable
...
+    void SetPeriod(uint32_t aPeriod) { mPeriod = aPeriod; }
+
+    /**
+     * This method gets the current jitter interval.
      *
-     * @returns A constant reference to channel mask
+     * @returns The jitter interval (in milliseconds).
+     *
+     */
+    uint16_t GetJitter(void) const { return mJitter; }
```
brianljt, line 138: Not used. Remove this?

brianljt (approved):
LGTM. Any CI test cases to verify MLE Announce frame?

src/core/thread/announce_sender.cpp
```diff
@@ -150,35 +236,44 @@ void AnnounceSender::CheckState(void)
...
-    {
-        period = kMinTxPeriod;
-    }
-
-    VerifyOrExit(!IsRunning() || (period != GetPeriod()) || (GetChannelMask() != channelMask));
+    // Start the trickle timer with same min and max interval as the
+    // desired Announce Tx cycle interval.
 
-    SendAnnounce(channelMask, 0, period, kMaxJitter);
+    mTrickleTimer.Start(TrickleTimer::kModeTrickle, kInterval, kInterval, kRedudancyConstant);
```
Irving-cl, line 242:
```suggestion
    mTrickleTimer.Start(TrickleTimer::kModeTrickle, kInterval, kInterval, kRedundancyConstant);
```

src/core/thread/mle.cpp
```diff
@@ -3813,10 +3813,14 @@ void Mle::HandleAnnounce(const Message &aMessage, const Ip6::MessageInfo &aMessa
...
     {
-        // do nothing
-        // timestamps are equal: no behaviour specified by the Thread spec.
-        // If SendAnnounce is executed at this point, there exists a scenario where
-        // multiple devices keep sending MLE Announce messages to one another indefinitely.
+        // Timestamps are equal.
+
+#if OPENTHREAD_CONFIG_ANNOUNCE_SENDER_ENABLE
+        // Notify `AnnounceSender` of the received Announce
+        // message so it can update it state to determine
```
Irving-cl, line 3820:
```suggestion
        // message so it can update its state to determine
```

Irving-cl (approved):
LGMT 👍

abtink:
> Any CI test cases to verify MLE Announce frame?

@brianljt  Yes. I think multiple existing tests already cover it. One main example is `test-603-channel-manager-announce-recovery.py` which verifies that an orphaned node (on a different channel) is recovered through MLE Announcement.

src/core/thread/announce_sender.hpp
abtink: I think it would be good to keep getter when we have a setter for the property. It's small and inline and if not used will be linked out.

src/core/thread/announce_sender.hpp
```diff
@@ -95,23 +111,76 @@ class AnnounceSenderBase : public InstanceLocator, private NonCopyable
...
 
     /**
-     * This method gets the channel mask for the latest scheduled MLE Announce transmission (the one in progress or the
-     * last finished one).
+     * This method sets the period interval.
+     *
+     * The period along with jitter value from (`Get/SetJitter()`) determines the interval between two successive MLE
+     * Announcement transmissions (possibly) on different channels from the specified channel mask.
+
+     * @param[in] aPeriod   The period interval (in milliseconds).
```
CiaranWoodward, line 127: nit:
```suggestion
     * Announcement transmissions (possibly) on different channels from the specified channel mask.
     * @param[in] aPeriod   The period interval (in milliseconds).
```

CiaranWoodward (commented):
Just took a read through, please correct my understanding if I'm mistaken!
- So basically this is running a trickle timer and sending announces (by starting an 'announce cycle') when it expires.
- Fixed trickle 'interval' (I) (Currently 11 min 28 seconds) (so we don't use the adaptive interval of trickle)
- Redundancy constant of 8 (k) ( so if we receive 8 announces in the 'window', we don't bother starting an 'announce cycle')
- The actual 'window' for trickle is randomly between interval/2 and interval
- So we would expect an absolute maximum of 16 announces per channel in a 5 min 44 second interval. (of devices all in radio range)

Concern: When a trickle timer event is 'triggered', this triggers an 'announce cycle' over a period of time that is very likely to exceed the trickle timer window length. It seems like the mechanisms within trickle work best if the 'transmission time' as RFC6206 would describe it is much less than the interval. Otherwise the delay between action and feedback are going to cause oscillations in the number of announces sent on the network, and the announces will likely exceed the above limit.

It looks like RFC6202 has some recommendations on how to set the values, such as sections 5 and 6.6. I think just cycling through every channel on a channel page (with say half-second delay and half-second jitter) would be more suitable than staggering it across an entire interval period. This would allow trickle to function properly, and still shouldn't cause any disruption on the node.

abtink:
Thanks @CiaranWoodward. We talked about this in the Thread Low-power WG meeting today. Just adding a quick overview here:

- We have a cycle interval (~11.5 min). 
- We use a trickle timer and use the same cycle interval as both min and max trickle timer interval. 
- Trickle timer practically is running as a periodic timer with this interval.
- In each cycle, the trickle timer picks a random point of time within the second half of the interval.
- Up the randomly selected point, we listen and count the number of received Announce message from neighbors. 
-  If the number is higher than the redundancy threshold, device skips sending Announce message in that cycle, and then wait till end of the interval and repeats the same process.
- If the number is lower than the threshold, then device decides to send MLE annoucnemnt in this cycle
    - The first annoucne message is send within a short jitter delay after the decision and it is sent on the current PAN channel (being used by nodes).
    - This is then followed by sending Announce on all channels in the current channel mask (till we get back to PAN channel). 
    - In the new pushed commit, I changed the logic to use half the interval for sending all channels (instead of full interval). So we divide half interval by the number of channels in mask and use it as period between two successive Announce Tx on different channels.
    - This way, we should always finish the tx from previous cycle, before we need to make a decision on the next cycle.

CiaranWoodward (approved):
Thanks for summarizing our discussion here @abtink - I'm satisfied my earlier concerns will not be a problem (because trickle waits entire interval before restarting & current channel is announced first).  I also think halving the send-all-channels announce period is a good change anyway.

LGTM!

src/core/thread/announce_sender.hpp
```diff
@@ -95,23 +111,76 @@ class AnnounceSenderBase : public InstanceLocator, private NonCopyable
...
+     *
+     * @returns The current starting channel.
+     *
+     */
+    uint8_t GetStartingChannel(void) const { return mStartingChannel; }
+
+    /**
+     * This method sets the starting channel, i.e., the first channel in a TX cycle to send MLE Announcement on.
+     *
+     * @p aStartinChannel MUST be present in the current channel mask (from `GetChannelMask()`), otherwise it is
```
jwhui, line 175:
```suggestion
     * @p aStartingChannel MUST be present in the current channel mask (from `GetChannelMask()`), otherwise it is
```

jwhui (commented):
Looks great overall. One minor nit.

jwhui (approved):

PR #6406, opened 2021-04-06
If the target TUN already exists and is persist, OTBR will make it non-persist and take over it.

superwhd:


simonlingoogle:
Expecting further information from https://github.com/openthread/ot-br-posix/issues/741#issuecomment-818079800

PR #6407, opened 2021-04-06
[srp-server] replace SRP Host pointer with SRP service update ID

wgtdkp:
We are using the SRP Host pointer for indexing an outstanding SRP update when
the Advertising Proxy result is passed back with `otSrpServerHandleServiceUpdateResult`.
There are chances that an old, freed Host object may be matched again with a newly allocated
Host object which coincidentally has the same address with the old Host object.

This PR solves this issue by using a random generated `uint32_t` SRP service update ID for indexing
outstanding updates.

Depends-On: openthread/ot-br-posix#774

### DIFF
--- a/include/openthread/srp_server.h
+++ b/include/openthread/srp_server.h
@@ -68,6 +68,12 @@ typedef void otSrpServerHost;
 typedef void otSrpServerService;
 
 /**
+ * The ID of a SRP service update transaction on the SRP Server.
+ *
+ */
+typedef uint32_t otSrpServerServiceUpdateId;
+
+/**
  * This method returns the domain authorized to the SRP server.
  *
  * If the domain if not set by SetDomain, "default.service.arpa." will be returned.
@@ -144,11 +150,11 @@ otError otSrpServerSetLeaseRange(otInstance *aInstance,
  * if any validation fails. For example, an Advertising Proxy should advertise (or remove) the host and
  * services on a multicast-capable link and returns specific error code if any failure occurs.
  *
- * @param[in]  aHost     A pointer to the otSrpServerHost object which contains the SRP updates.
- *                       The pointer should be passed back to otSrpServerHandleServiceUpdateResult, but
- *                       the content MUST not be accessed after this method returns. The handler
- *                       should publish/un-publish the host and each service points to this host
- *                       with below rules:
+ * @param[in]  aId       The service update transaction ID. This ID must be passed back with
+ *                       `otSrpServerHandleServiceUpdateResult`.
+ * @param[in]  aHost     A pointer to the otSrpServerHost object which contains the SRP updates. The
+ *                       handler should publish/un-publish the host and each service points to this
+ *                       host with below rules:
  *                         1. If the host is not deleted (indicated by `otSrpServerHostIsDeleted`),
  *                            then it should be published or updated with mDNS. Otherwise, the host
  *                            should be un-published (remove AAAA RRs).
@@ -163,7 +169,10 @@ otError otSrpServerSetLeaseRange(otInstance *aInstance,
  * @sa otSrpServerHandleServiceUpdateResult
  *
  */
-typedef void (*otSrpServerServiceUpdateHandler)(const otSrpServerHost *aHost, uint32_t aTimeout, void *aContext);
+typedef void (*otSrpServerServiceUpdateHandler)(otSrpServerServiceUpdateId aId,
+                                                const otSrpServerHost *    aHost,
+                                                uint32_t                   aTimeout,
+                                                void *                     aContext);
 
 /**
  * This method sets the SRP service updates handler on SRP server.
@@ -185,12 +194,13 @@ void otSrpServerSetServiceUpdateHandler(otInstance *                    aInstanc
  * processing of a SRP update.
  *
  * @param[in]  aInstance  A pointer to an OpenThread instance.
- * @param[in]  aHost      A pointer to the Host object which represents a SRP update.
+ * @param[in]  aId        The service update transaction ID. This should be the same ID
+ *                        provided via `otSrpServerServiceUpdateHandler`.
  * @param[in]  aError     An error to be returned to the SRP server. Use OT_ERROR_DUPLICATED
  *                        to represent DNS name conflicts.
  *
  */
-void otSrpServerHandleServiceUpdateResult(otInstance *aInstance, const otSrpServerHost *aHost, otError aError);
+void otSrpServerHandleServiceUpdateResult(otInstance *aInstance, otSrpServerServiceUpdateId aId, otError aError);
 
 /**
  * This method returns the next registered host on the SRP server.
--- a/src/core/net/srp_server.cpp
+++ b/src/core/net/srp_server.cpp
@@ -39,6 +39,7 @@
 #include "common/locator-getters.hpp"
 #include "common/logging.hpp"
 #include "common/new.hpp"
+#include "common/random.hpp"
 #include "net/dns_types.hpp"
 #include "thread/network_data_service.hpp"
 #include "thread/thread_netif.hpp"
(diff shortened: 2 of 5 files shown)


src/core/net/srp_server.hpp
```diff
@@ -521,28 +522,30 @@ class Server : public InstanceLocator, private NonCopyable
...
                        const Ip6::MessageInfo & aMessageInfo);
 
-        TimeMilli         mExpireTime;
-        Dns::UpdateHeader mDnsHeader;
-        Host *            mHost;        // The host will be updated. The UpdateMetadata has no ownership of this host.
-        Ip6::MessageInfo  mMessageInfo; // The message info of the DNS update request.
-        UpdateMetadata *  mNext;
+        TimeMilli                  mExpireTime;
+        Dns::UpdateHeader          mDnsHeader;
+        otSrpServerServiceUpdateId mId;   // The ID of this service update transaction.
```
abtink, line 545: Suggestion: we can add `typedef otSrpServerServiceUpdateId ServiceUpdateId` at the top of the `Server` class to mirror the type and then use it inside (helps make the type name shorter).

src/core/net/srp_server.cpp
```diff
@@ -1691,6 +1694,7 @@ Server::UpdateMetadata::UpdateMetadata(Instance &               aInstance,
     : InstanceLocator(aInstance)
     , mExpireTime(TimerMilli::GetNow() + kDefaultEventsHandlerTimeout)
     , mDnsHeader(aHeader)
+    , mId(Random::NonCrypto::GetUint32())
```
abtink, line 1697: Do we need to worry about possible duplicate (already in use) IDs (or is it very unlikely)?

abtink (approved):
@wgtdkp Looks good to me. Thanks. 👍 

One question and a suggestion belwo.

src/core/net/srp_server.cpp
abtink: Do we need the ID to be random? If I am not mistaken, we may only care that it is different per host item
Does this sound correct? If so, can we then consider using an increasing counter for ID? (this way it will take a long time for it to wrap around and reuse the same value).

wgtdkp: Yeah, I was assuming that there are rare chance for two outstanding updates to have conflict IDs. I chose random ID just because it needs less change :)
But I would agree incremental ID is more suitable and faster though it needs to maintain the `ID` in `Srp::Server`.

src/core/net/srp_server.cpp
```diff
@@ -86,6 +87,7 @@ Server::Server(Instance &aInstance)
     , mMaxKeyLease(kDefaultMaxKeyLease)
     , mLeaseTimer(aInstance, HandleLeaseTimer)
     , mOutstandingUpdatesTimer(aInstance, HandleOutstandingUpdatesTimer)
+    , mServiceUpdateId(Random::NonCrypto::GetUint32())
```
wgtdkp, line 90: Use random initial value so that there will be small chances for IDs between resets of openthread stack to conflict with each other. This is for cases that the Advertising Proxy implementation may not be reset when openthread stack is reset.

include/openthread/srp_server.h
```diff
@@ -185,12 +194,13 @@ void otSrpServerSetServiceUpdateHandler(otInstance *                    aInstanc
  * processing of a SRP update.
  *
  * @param[in]  aInstance  A pointer to an OpenThread instance.
- * @param[in]  aHost      A pointer to the Host object which represents a SRP update.
+ * @param[in]  aId        The ID of a service update transaction. This should be the same ID
```
simonlingoogle, line 197:
```suggestion
 * @param[in]  aId        The service update transaction ID. This should be the same ID
```

include/openthread/srp_server.h
```diff
@@ -144,11 +150,11 @@ otError otSrpServerSetLeaseRange(otInstance *aInstance,
  * if any validation fails. For example, an Advertising Proxy should advertise (or remove) the host and
  * services on a multicast-capable link and returns specific error code if any failure occurs.
  *
- * @param[in]  aHost     A pointer to the otSrpServerHost object which contains the SRP updates.
- *                       The pointer should be passed back to otSrpServerHandleServiceUpdateResult, but
- *                       the content MUST not be accessed after this method returns. The handler
- *                       should publish/un-publish the host and each service points to this host
- *                       with below rules:
+ * @param[in]  aId       The ID of this service update transaction. This ID must be passed back with
```
simonlingoogle, line 153:
```suggestion
 * @param[in]  aId       The service update transaction ID. This ID must be passed back with
```

src/core/net/srp_server.hpp
```diff
@@ -71,13 +71,22 @@ namespace Srp {
 class Server : public InstanceLocator, private NonCopyable
 {
     friend class ot::Notifier;
+    friend class UpdateMetadata;
+    friend class Service;
+    friend class Host;
```
simonlingoogle, line 76: Make Server::AllocateId public?

simonlingoogle (approved):
LGTM :+1:

include/openthread/srp_server.h
```diff
@@ -163,7 +169,10 @@ otError otSrpServerSetLeaseRange(otInstance *aInstance,
  * @sa otSrpServerHandleServiceUpdateResult
  *
  */
-typedef void (*otSrpServerServiceUpdateHandler)(const otSrpServerHost *aHost, uint32_t aTimeout, void *aContext);
+typedef void (*otSrpServerServiceUpdateHandler)(otSrpServerServiceUpdateId aId,
```
simonlingoogle, line 172:
```suggestion
typedef otSrpServerServiceUpdateId (*otSrpServerServiceUpdateHandler)(
```
nit: an alternative solution is let `otSrpServerServiceUpdateHandler` return a `Transaction ID`

wgtdkp: The SRP Server needs to create the ID and pass it to the Advertising Proxy. The return value will not work.

src/core/net/srp_server.hpp
wgtdkp: the ID is implementation-specific stuff and I think there are no other modules that will need access to it.

Irving-cl (approved):
LGTM 👍

include/openthread/srp_server.h
```diff
@@ -144,11 +150,11 @@ otError otSrpServerSetLeaseRange(otInstance *aInstance,
...
  * services on a multicast-capable link and returns specific error code if any failure occurs.
  *
- * @param[in]  aHost     A pointer to the otSrpServerHost object which contains the SRP updates.
- *                       The pointer should be passed back to otSrpServerHandleServiceUpdateResult, but
- *                       the content MUST not be accessed after this method returns. The handler
- *                       should publish/un-publish the host and each service points to this host
- *                       with below rules:
+ * @param[in]  aId       The service update transaction ID. This ID must be passed back with
+ *                       `otSrpServerHandleServiceUpdateResult`.
+ * @param[in]  aHost     A pointer to the otSrpServerHost object which contains the SRP updates.The
```
jwhui, line 155:
```suggestion
 * @param[in]  aHost     A pointer to the otSrpServerHost object which contains the SRP updates. The
```

include/openthread/srp_server.h
```diff
@@ -185,12 +194,13 @@ void otSrpServerSetServiceUpdateHandler(otInstance *                    aInstanc
  * processing of a SRP update.
  *
  * @param[in]  aInstance  A pointer to an OpenThread instance.
- * @param[in]  aHost      A pointer to the Host object which represents a SRP update.
+ * @param[in]  aId        The service update transaction ID. This should be the same ID
+ *                        got in `otSrpServerServiceUpdateHandler` arguments.
```
jwhui, line 198:
```suggestion
 *                        provided via `otSrpServerServiceUpdateHandler`.
```

jwhui (approved):

PR #6408, opened 2021-04-06
[cli] add more output to `eidcache` command

jwhui:


src/cli/cli.cpp
```diff
@@ -1630,6 +1630,46 @@ void Interpreter::HandleDnsServiceResponse(otError aError, const otDnsServiceRes
...
+const char *EidCacheStateToString(otCacheEntryState aState)
+{
+    static const char *const kStateStrings[4] = {
+        "cache",
+        "snoop",
+        "query",
+        "retry",
+    };
+
+    return aState <= OT_CACHE_ENTRY_STATE_RETRY_QUERY ? kStateStrings[aState] : "unkno";
```
simonlingoogle, line 1642:
```suggestion
    return aState <= OT_CACHE_ENTRY_STATE_RETRY_QUERY ? kStateStrings[aState] : "unknown";
```

Is it intentional ot use "unkno" ?

src/cli/cli.cpp
```diff
@@ -1643,9 +1683,7 @@ otError Interpreter::ProcessEidCache(uint8_t aArgsLength, char *aArgs[])
     for (uint8_t i = 0;; i++)
     {
         SuccessOrExit(otThreadGetNextCacheEntry(mInstance, &entry, &iterator));
-
-        OutputIp6Address(entry.mTarget);
-        OutputLine(" %04x", entry.mRloc16);
+        PrintEidCacheEntry(entry);
```
simonlingoogle, line 1686:
```suggestion
        OutputEidCacheEntry(entry);
```
nit

src/cli/cli.cpp
```diff
@@ -1630,6 +1630,46 @@ void Interpreter::HandleDnsServiceResponse(otError aError, const otDnsServiceRes
...
+    OutputIp6Address(aEntry.mTarget);
+    OutputFormat(" %04x", aEntry.mRloc16);
+    OutputFormat(" %s", EidCacheStateToString(aEntry.mState));
+    OutputFormat(" canEvict=%d", aEntry.mCanEvict);
+
+    if (aEntry.mState == OT_CACHE_ENTRY_STATE_CACHED)
+    {
+        if (aEntry.mValidLastTrans)
+        {
+            OutputFormat(" transTime=%d eid=", aEntry.mLastTransTime);
```
simonlingoogle, line 1656:
```suggestion
            OutputFormat(" transTime=%u eid=", aEntry.mLastTransTime);
```

simonlingoogle (approved):
LGTM :+1: 

with nits

src/cli/cli.cpp
simonlingoogle, line 1642:
```suggestion
    return aState < OT_ARRAY_LENGTH(kStateStrings) ? kStateStrings[aState] : "unkno";
```

in case we change the state values.

src/cli/cli.cpp
```diff
@@ -1630,6 +1630,46 @@ void Interpreter::HandleDnsServiceResponse(otError aError, const otDnsServiceRes
...
+        "cache",
+        "snoop",
+        "query",
+        "retry",
+    };
+
+    return aState <= OT_CACHE_ENTRY_STATE_RETRY_QUERY ? kStateStrings[aState] : "unkno";
+}
+
+void Interpreter::PrintEidCacheEntry(const otCacheEntryInfo &aEntry)
```
abtink, line 1645: How about using `OutputEidCacheEntry()` instead of `PrintEidCacheEntry()` (we seem to use `Output{Something}` pattern in CLI).

abtink (approved):
LGTM. 👍 
One nit suggestion about rename.

src/cli/cli.cpp
abtink: I suggested the same below (noticed this afterwards 😃)

src/cli/cli.cpp
jwhui: It was intentional to keep the table aligned, but I don't think alignment is that important.

jwhui: I had to add `static_cast` to avoid the following:
```
../../src/cli/cli.cpp:1642:19: error: result of comparison of constant 4 with expression of type 'otCacheEntryState' is always true [-Werror,-Wtautological-constant-out-of-range-compare]
    return aState < OT_ARRAY_LENGTH(kStateStrings) ? kStateStrings[aState] : "unknown";
           ~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

Irving-cl (approved):
LGTM 👍

PR #6413, opened 2021-04-07
[config] update HEADER_IE_SUPPORT to a derived config

Irving-cl:
This PR makes `OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT` to be derived 
from other configurations and make it unconfigurable. I'm thinking 
that `OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT` is  technically not a 
'feature'. This macro wraps some implementation that serves for 
some features (TIME_SYNC, CSL, etc). But it doesn't make sense to 
enable `OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT` without enabling 
any other features that are dependent on it.

I call config like `OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT` as 
'derived config'. This PR adds a dedicated config file `derived.h` 
at the end of core config file for putting these 'derived config'.


src/core/mac/mac_frame.cpp
```diff
@@ -975,6 +974,8 @@ const uint8_t *Frame::GetHeaderIe(uint8_t aIeId) const
     return header;
 }
 
+#endif // OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT
```
edmont, line 977: Notice this `#endif` can be removed together with the `#if` around line 897, which is redundant.

edmont (commented):
Thanks, I was about to post a related PR and saw this.

Please just add some `#endif // OPENTHREAD_CONFIG_MAC_HEADER_IE_SUPPORT` at the end of `nrf_802154_tx_started()` in `examples/platforms/nrf528xx/src/radio.c`.

Irving-cl:
obselete

PR #6414, opened 2021-04-07
[dns-client] add support for NAT64 during address resolution

abtink:
This commit adds support for NAT64 address translation (from IPv4 to
IPv6) during address resolution in `Dns::Client`. This feature is
enabled using `OPENTHREAD_CONFIG_DNS_CLIENT_NAT64_ENABLE` config
option. When enabled, the caller can specify in the address query
configuration whether to allow or disallow NAT64 address translation
for a specific query. The default behavior can also be set using the
default DNS client config.

When the NAT64 is allowed, if the response from server does not
provide an IPv6 address for the host name, we first check if the
server has provided IPv4 address information in the response, and if
not, start a new query for the IPv4 address. Once the IPv4 address(es)
of the host name are resolved, they are translated to IPv6 address(es)
and reported to the user in the query callback function.

--------------

**_Note:_**
In current PR, the NAT64 prefix is determined using `GetNat64Prefix()` method
which for now simply returns the well-known prefix. This is a temporary solution 
and will be updated in a future PR.

### DIFF
--- a/include/openthread/dns_client.h
+++ b/include/openthread/dns_client.h
@@ -67,6 +67,20 @@ typedef enum
 } otDnsRecursionFlag;
 
 /**
+ * This enumeration type represents the NAT64 mode in an `otDnsQueryConfig`.
+ *
+ * The NAT64 mode indicates whether to allow or disallow NAT64 address translation during DNS client address resolution.
+ * This mode is only used when `OPENTHREAD_CONFIG_DNS_CLIENT_NAT64_ENABLE` is enabled.
+ *
+ */
+typedef enum
+{
+    OT_DNS_NAT64_UNSPECIFIED = 0, ///< NAT64 mode is not specified. Use default NAT64 mode.
+    OT_DNS_NAT64_ALLOW       = 1, ///< Allow NAT64 address translation during DNS client address resolution.
+    OT_DNS_NAT64_DISALLOW    = 2, ///< Do not allow NAT64 address translation during DNS client address resolution.
+} otDnsNat64Mode;
+
+/**
  * This structure represents a DNS query configuration.
  *
  * Any of the fields in this structure can be set to zero to indicate that it is not specified. How the unspecified
@@ -79,6 +93,7 @@ typedef struct otDnsQueryConfig
     uint32_t           mResponseTimeout; ///< Wait time (in msec) to rx response. Zero indicates unspecified value.
     uint8_t            mMaxTxAttempts;   ///< Maximum tx attempts before reporting failure. Zero for unspecified value.
     otDnsRecursionFlag mRecursionFlag;   ///< Indicates whether the server can resolve the query recursively or not.
+    otDnsNat64Mode     mNat64Mode;       ///< Allow/Disallow NAT64 address translation during address resolution.
 } otDnsQueryConfig;
 
 /**
--- a/src/core/net/dns_client.cpp
+++ b/src/core/net/dns_client.cpp
@@ -60,6 +60,9 @@ Client::QueryConfig::QueryConfig(InitMode aMode)
     SetResponseTimeout(kDefaultResponseTimeout);
     SetMaxTxAttempts(kDefaultMaxTxAttempts);
     SetRecursionFlag(kDefaultRecursionDesired ? kFlagRecursionDesired : kFlagNoRecursion);
+#if OPENTHREAD_CONFIG_DNS_CLIENT_NAT64_ENABLE
+    SetNat64Mode(kDefaultNat64Allowed ? kNat64Allow : kNat64Disallow);
+#endif
 }
 
 void Client::QueryConfig::SetFrom(const QueryConfig &aConfig, const QueryConfig &aDefaultConfig)
@@ -94,6 +97,13 @@ void Client::QueryConfig::SetFrom(const QueryConfig &aConfig, const QueryConfig 
     {
         SetRecursionFlag(aDefaultConfig.GetRecursionFlag());
     }
+
+#if OPENTHREAD_CONFIG_DNS_CLIENT_NAT64_ENABLE
+    if (GetNat64Mode() == kNat64Unspecified)
+    {
+        SetNat64Mode(aDefaultConfig.GetNat64Mode());
+    }
+#endif
 }
 
 //---------------------------------------------------------------------------------------------------------------------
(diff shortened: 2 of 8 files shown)


src/core/net/dns_client.hpp
```diff
@@ -108,6 +108,19 @@ class Client : public InstanceLocator, private NonCopyable
...
         };
 
+#if OPENTHREAD_CONFIG_DNS_CLIENT_NAT64_ENABLE
+        /**
+         * This enumeration type represents the NAT64 mode.
+         *
+         */
+        enum Nat64Mode
+        {
+            kNat64Unspecified = OT_DNS_NAT64_UNSPECIFIED, ///< NAT64 mode is not specified.
```
simonlingoogle, line 118:
```suggestion
            kNat64Unspecified = OT_DNS_NAT64_UNSPECIFIED, ///< NAT64 mode is not specified. Use default NAT64 mode.
```

simonlingoogle (commented):
I guess the DNS-SD server needs to implement additional `A` RR for `AAAA` query.

src/core/net/dns_client.cpp
```diff
@@ -694,7 +810,8 @@ void Client::FinalizeQuery(Query &aQuery, Error aError)
     Response  response;
     QueryInfo info;
 
-    response.mQuery = &aQuery;
+    response.mInstance = &Get<Instance>();
```
Irving-cl, line 813: What is `mInstance` used for? It looks like in this PR it's only set but not used.

Irving-cl (approved):
LGTM 👍

src/core/net/dns_client.cpp
abtink: Correct. It is not used in this PR. It will be used in a follow-up PR to find the NAT64 prefix from network-data. I considered whether to add it in this one or leave it for the next one. At the end I felt, it may be better to add it in this PR along with the rest of changes in `Dns::Client`.

src/core/net/dns_client.hpp
abtink: > I guess the DNS-SD server needs to implement additional A RR for AAAA query.

Client will query for `A` in a new request/message if it is not included in the original `AAAA` query response.
So we don't require the server to include it. It is optional but it will be a good optimization.

src/core/net/dns_client.cpp
```diff
@@ -257,9 +307,60 @@ Error Client::Response::FindServiceInfo(Section aSection, const Name &aName, Ser
...
+        SuccessOrExit(error = GetNat64Prefix(nat64Prefix));
+
+        section = (info.mQueryType == kIp4AddressQuery) ? kAnswerSection : kAdditionalDataSection;
+        SuccessOrExit(error = FindARecord(section, name, aIndex, aRecord));
+
+        aAddress.SetFromTranslatedIp4Address(nat64Prefix, aRecord.GetAddress());
+        aTtl = aRecord.GetTtl();
+
+        ExitNow();
+    }
```
wgtdkp, line 339: In case this is an IPv6 address query and the response includes `AAAA` RR in the `Answer Section` and `A` RR in the `Additional Section`, current implementation with prefer the IPv4 address. Should we prefer IPv6 address for IPv6 query?

simonlingoogle (approved):
LGTM :+1:

src/core/net/dns_client.cpp
abtink: I don't think that is the case in current code. If there is an `AAAA` record in the response, it will be preferred and used and NAT64 won't be used.

Note that the `if` block here is tied to `mHasEmbeddedIp4` being `true`. This boolean variable is only set when there is an error code from server (which indicates it could not provide `AAAA` record in response). Only in such a case, we check if there happens to be an `A` record in the received response and if so, we set `mHasEmbeddedIp4`.  This is all done in `ParseResponse()` method (new code in this PR).

include/openthread/dns_client.h
```diff
@@ -66,6 +66,20 @@ typedef enum
...
+/**
+ * This enumeration type represents the NAT64 mode in an `otDnsQueryConfig`.
+ *
+ * The NAT64 mode indicates whether to allow or disallow NAT64 address translation during DNS client address resolution.
+ * This mode is only used when `OPENTHREAD_CONFIG_DNS_CLIENT_NAT64_ENABLE` is enabled.
+ *
+ */
+typedef enum
+{
+    OT_DNS_NAT64_UNSPECIFIED = 0, ///< NAT64 mode is not specified. Use default mode.
```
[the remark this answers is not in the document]
abtink: I prefer `UNSPECIFIED` since we use it in other definitions here as well.

src/core/net/dns_client.cpp
simonlingoogle: > This boolean variable is only set when there is an error code from server (which indicates it could not provide `AAAA` record in response).

I wonder what RCODE should the DNS server use when there is no AAAA RR available but A RR is available. This PR implies that RCODE should be NXDOMAIN. But I think it's also logical to use RCODE=NO_ERROR. I think it's better to search for A RR as long as there is no AAAA RR available, no matter what RCODE is.

wgtdkp: @abtink Thanks for clarification, I got the `mHasEmbeddedIp4` wrong! Maybe we can rename it to `mHasOnlyEmbeddedIp4` :)

wgtdkp (approved):
LGTM 👍  Thanks @abtink !

src/core/net/dns_client.cpp
abtink: > But I think it's also logical to use RCODE=NO_ERROR. I think it's better to search for A RR as long as there is no AAAA RR available, no matter what RCODE is.

If I am not mistaken, I think if the server cannot provide any response for the main query question (which is `AAAA` records in this case) then it may not be allowed to use `NO_ERROR` response code.

We can add such a check in client to cover if server happens behave as this.

Let me see how complex it will be to add. If it does not add much overhead (code size, ect). I am ok adding it.

abtink: I changed this in the new pushed commit, i.e. we don't check the response error code from server and instead check if the response does not contain an AAAA record for the query name.

simonlingoogle: RCODE is defined to be NO_ERROR for a similar case:

https://tools.ietf.org/html/draft-vavrusa-dnsop-aaaa-for-free-00#section-5.3

But this draft also suggests to add to answer instead of additional section, which conflicts with this PR.

Maybe we should treat ANSWER and ADDITIONAL RRs as equal  so as to work with all server variations.

jwhui: Maybe @Abhayakara can chime in what RCODE is expected when no AAAA record is returned, but an A record is included in the ADDITIONAL section.

Abhayakara: If there is no AAAA record, but there are other answers on the name, the correct RCODE to return is "no error." The answer count of zero tells us that there's no data. The additional section is then available for synthesis.

Abhayakara: The draft is wrong, by the way. That would never get past the DNSOP working group's review.

abtink: Perfect. Thanks guys!

src/core/net/dns_client.cpp
```diff
@@ -257,9 +307,60 @@ Error Client::Response::FindServiceInfo(Section aSection, const Name &aName, Ser
...
+        Section     section;
+        ARecord     aRecord;
+        Ip6::Prefix nat64Prefix;
+
+        SuccessOrExit(error = GetNat64Prefix(nat64Prefix));
+
+        section = (info.mQueryType == kIp4AddressQuery) ? kAnswerSection : kAdditionalDataSection;
+        SuccessOrExit(error = FindARecord(section, name, aIndex, aRecord));
+
+        aAddress.SetFromTranslatedIp4Address(nat64Prefix, aRecord.GetAddress());
```
jwhui, line 335: Looking at this now, the method name seems somewhat strange. `aRecord.GetAddress()` is a straight-up IPv4 address, but the method name indicates this is a "translated IPv4 address".

src/core/net/dns_client.cpp
```diff
@@ -257,9 +307,60 @@ Error Client::Response::FindServiceInfo(Section aSection, const Name &aName, Ser
...
 {
-    return FindHostAddress(kAnswerSection, Name(*mQuery, kNameOffsetInQuery), aIndex, aAddress, aTtl);
+    Error error = kErrorNone;
+    Name  name(*mQuery, kNameOffsetInQuery);
+
+#if OPENTHREAD_CONFIG_DNS_CLIENT_NAT64_ENABLE
+
+    // If the response is for an IPv4 address query or if it is an
+    // IPv6 address query response with no IPv6 address but with an
+    // embedded IPv4 in its additional section, we read the IPv4
```
jwhui, line 317: What is an "embedded IPv4"?

src/core/net/dns_client.hpp
jwhui: Should we include "Use default NAT64 mode." in the comment, as @simonlingoogle  suggested aove?

abtink: I think I may have missed adding it. Thanks.

abtink: Looks like it was already in latest commit (but somehow it does not show here that its "outdated").

abtink: I see! I added it in `dns_clinet.h` but still missing in `.hpp` one.

src/core/net/dns_client.cpp
abtink: How about  `SetByTranslatingIp4Address()`? (open to better suggestions as well).

abtink: I will rename it to `SynthesizeFromIp4Address()`.

src/core/net/dns_client.cpp
abtink: Removed/replaced this.

src/core/net/dns_client.cpp
jwhui: SGTM.

jwhui (approved):

PR #6419, opened 2021-04-08
[thread-cert] specify MAX_JOBS via env

simonlingoogle:
This commit allow specify `MAX_JOBS` via env to set the max number of tests running concurrently.

### DIFF
--- a/tests/scripts/thread-cert/run_cert_suite.py
+++ b/tests/scripts/thread-cert/run_cert_suite.py
@@ -39,7 +39,7 @@ import config
 
 THREAD_VERSION = os.getenv('THREAD_VERSION')
 VIRTUAL_TIME = int(os.getenv('VIRTUAL_TIME', '1'))
-MAX_JOBS = multiprocessing.cpu_count() * 2 if VIRTUAL_TIME else 10
+MAX_JOBS = int(os.getenv('MAX_JOBS', (multiprocessing.cpu_count() * 2 if VIRTUAL_TIME else 10)))
 
 _BACKBONE_TESTS_DIR = 'tests/scripts/thread-cert/backbone'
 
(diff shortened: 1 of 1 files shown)


tests/scripts/thread-cert/run_cert_suite.py
```diff
@@ -39,7 +39,7 @@
 
 THREAD_VERSION = os.getenv('THREAD_VERSION')
 VIRTUAL_TIME = int(os.getenv('VIRTUAL_TIME', '1'))
-MAX_JOBS = multiprocessing.cpu_count() * 2 if VIRTUAL_TIME else 10
+MAX_JOBS = int(os.getenv('MAX_JOBS', (multiprocessing.cpu_count() * 2 if VIRTUAL_TIME else 10)))
```
bukepo, line 42: is the environment variable just for local use?

simonlingoogle: yes.

bukepo: Prefer adding a `OT_` prefix for new such kind of environments, thoughts?
```suggestion
MAX_JOBS = int(os.getenv('OT_MAX_JOBS', (multiprocessing.cpu_count() * 2 if VIRTUAL_TIME else 10)))
```

bukepo (approved):

tests/scripts/thread-cert/run_cert_suite.py
simonlingoogle: Note that we don't have `OT_` prefix for `VIRTUAL_TIME` and `THREAD_VERSION`.
I prefer `MAX_JOBS` without `OT_` because it's just for local debugging (not to be used by general OT users).

simonlingoogle: Maybe we can have `OT_TEST_` prefix for test variables, i.e. `OT_TEST_MAX_JOBS`

simonlingoogle, line 42:
```suggestion
MAX_JOBS = int(os.getenv('OT_TEST_MAX_JOBS', (multiprocessing.cpu_count() * 2 if VIRTUAL_TIME else 10)))
```

Irving-cl (approved):
LGTM 👍

jwhui (approved):

PR #6423, opened 2021-04-08
[mesh-forwarder] add missing anycast forwarding features

jwhui:
- [dhcp6-server] fix removal of anycast address
- [dhcp6-server] add anycast address for stateless config
- [mesh-forwarder] use path cost in DHCPv6 ALOC lookup
- [tests] add test for anycast forwarding
- [mesh-forwarder] add support for Neighbor Discovery Agent Anycast


src/core/net/nd_agent.cpp
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+exit:
+    return;
+}
+
+void Agent::ApplyMeshLocalPrefix(void)
+{
+    VerifyOrExit(mAloc.mNext != &mAloc);
```
simonlingoogle, line 120:
```suggestion
    VerifyOrExit(Get<ThreadNetif>().HasUnicastAddress(mAloc));
```

Implement `HasUnicastAddress` in ThreadNetif by searching `mUnicastAddresses` list?

The current code depends on the how `LinkedList`  manages the `mNext` pointer.

abtink: I think this model may be fine here since `Agent` itself is owner of the `mAloc` variable.

We know that `LinkedList` class chains the entries together and always end with `nullptr` (since that's how we iterate through the list), so having an entry point back to itself is a good way to know that the entry is not in a linked-list.

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -507,8 +523,8 @@ Error MeshForwarder::UpdateIp6RouteFtd(Ip6::Header &ip6Header, Message &aMessage
         }
         else if (aloc16 <= Mle::kAloc16DhcpAgentEnd)
         {
-            uint8_t contextId = static_cast<uint8_t>(aloc16 & Mle::kAloc16DhcpAgentMask);
-            SuccessOrExit(error = ContextIdRouteLookup(contextId, mMeshDest));
+            uint8_t contextId = static_cast<uint8_t>(aloc16 - Mle::kAloc16DhcpAgentStart + 1);
```
abtink, line 526: I don't think `kAloc16DhcpAgentMask` is used anymore with this change, we can consider removing it?

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -402,6 +402,85 @@ Error MeshForwarder::UpdateMeshRoute(Message &aMessage)
...
+            else
+            {
+                cost = Get<Mle::MleRouter>().GetLinkCost(Mle::Mle::RouterIdFromRloc16(server16));
+            }
+
+            // Choose the minimum cost
+            if (cost < curCost)
+            {
+                curCost = cost;
+            }
```
abtink, line 474: How about `curCost = OT_MIN(curCost, cost);`?

abtink (commented):
Thanks @jwhui . Looks really great. 👍 
Couple of small nit suggestions.

abtink (approved):

src/core/net/nd_agent.cpp
simonlingoogle: > We know that `LinkedList` class chains the entries together and always end with `nullptr` (since that's how we iterate through the list), so having an entry point back to itself is a good way to know that the entry is not in a linked-list.

An alternative implementation of `LinkedList` is let the laster element's `mNext` pointer points to the first element. In such case, every element of a linked list must have `mNext != nullptr`, and we should use `mAloc.mNext != nullptr` to check if `mAloc` is in use. 
That's why I think the current code depends on how `LinkedList` is implemented.
But this seems to be a minor issue. I will resolve.

src/core/thread/mesh_forwarder_ftd.cpp
```diff
@@ -402,6 +402,101 @@ Error MeshForwarder::UpdateMeshRoute(Message &aMessage)
...
+        }
+
+        if ((bestDest == Mac::kShortAddrInvalid) || (curCost < bestCost))
+        {
+            bestDest = server16;
+            bestCost = curCost;
+        }
```
simonlingoogle, line 481: This code is extremely similar to the route cost calculation code in `MeshForwarder::GetDestinationRlocByServiceAloc`. 
Should we add a method for this common logic?

simonlingoogle (approved):
LGTM :+1: 

with a nit

src/core/thread/mesh_forwarder_ftd.cpp
jwhui: Updated.

abtink (approved):
👍  LGTM

simonlingoogle (approved):
LGTM :+1:

PR #6427, opened 2021-04-10
[nrf528xx] set key id and frame counter for retransmissions

konradderda:
For nRF52 platforms in order to achieve the best CSL accuracy frame security processing was moved to the nrf_802154_tx_started() radio driver callback which is called from the interrupt context when frame transmission is started by the hardware peripheral. 

There is a possibility that a requested transmission will be aborted before nrf_802154_tx_started() is called. The frame can be then retransmitted and if nrf_802154_tx_started() is called it will skip setting key index and frame counter (so they will remain unchanged with a value equal to whatever is present in RAM). 

It seems that we should always set these fields. Moreover, security is processed only once which is controlled by mTxInfo.mIsSecurityProcessed member so checking for retransmission is redundant. Thoughts?

### DIFF
--- a/examples/platforms/nrf528xx/src/radio.c
+++ b/examples/platforms/nrf528xx/src/radio.c
@@ -1222,11 +1222,8 @@ void nrf_802154_tx_started(const uint8_t *aFrame)
 
     sTransmitFrame.mInfo.mTxInfo.mAesKey = &sCurrKey;
 
-    if (!sTransmitFrame.mInfo.mTxInfo.mIsARetx)
-    {
-        otMacFrameSetKeyId(&sTransmitFrame, sKeyId);
-        otMacFrameSetFrameCounter(&sTransmitFrame, sMacFrameCounter++);
-    }
+    otMacFrameSetKeyId(&sTransmitFrame, sKeyId);
+    otMacFrameSetFrameCounter(&sTransmitFrame, sMacFrameCounter++);
 
     processSecurity = true;
 #endif // OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2


edmont (approved):
LGTM, it seems `mIsSecurityProcessed` should be enough.

In any case, could you apply the same change to the simulation radio in order to verify CI is not broken?

brianljt (approved):
LGTM, Thanks! Could you please apply the similar change to function `radioProcessTransmitSecurity` in `example/platforms/simulation/radio.c`?

Irving-cl (approved):
LGTM

LuDuda (approved):
Thanks :+1:

jwhui (approved):

abtink:
I think the existing code was correct and the behavior we wanted. 

I noticed this from https://github.com/openthread/openthread/pull/6446 and added more details in this comment https://github.com/openthread/openthread/pull/6446#discussion_r613479046.

Here is an overview of the overall behavior:
- The handling of tx security (and therefore frame counter assignment) can be either done by OT core or by platform layer (determine by `OT_RADIO_CAPS_TRANSMIT_SEC` capability falg from platform)
- Independent of this, OT core will always set the frame counter and key id when it marks a frame as retx `.mInfo.mTxInfo.mIsARetx`
- So platform radio layer is expected to set the frame counter, key id only when it is managing the tx security and the frame is not marked as a retx.

------------

> There is a possibility that a requested transmission will be aborted before nrf_802154_tx_started() is called. The frame can be then retransmitted and if nrf_802154_tx_started() is called it will skip setting key index and frame counter (so they will remain unchanged with a value equal to whatever is present in RAM).

Regarding this, I think we need a different fix. I think if platform layer is handling tx security and therefore the frame counter assignment, then we want to assign frame counter on first frame tx before we abort it.

konradderda:
@abtink Great points in both PRs, what do you think moving this mIsARetransmission check and setting frame counter and key index in otPlatRadioTransmit? Then we would avoid both problem with aborted transmissions and messing up indirect transmissions.

abtink:
Thanks @konradderda. Yes that makes sense.

PR #6428, opened 2021-04-11
[posix] remove NCP app

bukepo:
The POSIX NCP app(ot-ncp) is simulating of NCP with an RCP. With the latest otbr-agent, this is not needed any more. It is only used in CI tests. This commit removes the app to save maintenance efforts.


tests/toranj/build.sh
```diff
@@ -174,86 +168,6 @@ case ${build_config} in
         cp -p ${top_builddir}/examples/apps/ncp/ot-ncp-ftd ${top_builddir}/examples/apps/ncp/ot-ncp-ftd-15.4-trel
         ;;
 
-    rcp)
```
abtink, line 177: Suggest keeping the rcp build option here.   use it a lot during development and testing (quick way to conifigure a build with rcp only).

tests/toranj/build.sh
```diff
@@ -174,86 +168,6 @@ case ${build_config} in
...
-            --enable-coverage=${coverage} \
-            --enable-ncp \
-            --enable-radio-only \
-            --with-examples=simulation \
-            --disable-docs \
-            --enable-tests=$tests || die
-        make -j 8 || die
-        ;;
-
-    posix | posix-)
```
abtink, line 195: Actually I suggest keeping all the rest of configurations here in the `build.sh`. There is no harm/overhead in keeping them (even if they may not be by any CI).

It's useful tool to quickly configure a build flavor and test builds during development.

abtink:
@bukepo , is there a big win by removing the POSIX NCP build? I don't think we have a lot of efforts/overhead in maintaining it.

My concern is that this PR seem to be removing a lot of RCP/posix test in our CI (`toranj` test and others). I prefer to keep running them on all PRs.

simonlingoogle (approved):
LGTM :+1: 

Less maintenance overhead sounds better.

librasungirl (approved):

librasungirl:
`wpantund+ot-ncp+rcp` seems an interim architecture when migrating from `wpantund + ncp` to `daemon + rcp` solution, unless there is outstanding use case or advantage, it may be not necessary to commit any further maintenance efforts IMHO.

tests/toranj/build.sh
bukepo: I'm fine with keeping this part.

tests/toranj/build.sh
bukepo: After removing the NCP app, it seems these options cannot build valid binaries for toranj tests. Are you sure want to keep these options?

bukepo:
> @bukepo , is there a big win by removing the POSIX NCP build? I don't think we have a lot of efforts/overhead in maintaining it.
> 
> My concern is that this PR seem to be removing a lot of RCP/posix test in our CI (`toranj` test and others). I prefer to keep running them on all PRs.

I couldn't find good use cases for POSIX ncp app. It does take time to ensure CI tests passing with posix ncp simulation. I think removing it can make life of OpenThread development easier. We can add it back when one day there's a valid use case. Thoughts?

tests/toranj/build.sh
abtink: Yes. Let's keep them. We can change them to build CLI instead of NCP.

bukepo: Reverted changes in build.sh.

bukepo:
Tests passed, but GitHub Actions will keep waiting responses of deleted tests in PR. @abtink @jwhui

tests/toranj/build.sh
abtink: I think we may need to change the script for this build flavors to actually build CLI (and not NCP anymore).
It's ok I can take care of later in separate PR.

abtink (approved):
LGTM. 👍

jwhui (approved):

PR #6431, opened 2021-04-12
[border-router] learn RA header and parameters from infra interface

wgtdkp:
The PR enhance the Routing Manager to learn RA header and parameters from
RA messages initiated from infra interface. This addresses the issue that the BR
may be using different RA header against another RA daemon which is working
on the same infra interface (Have BR and another RA daemon working on the
same interface is common for `Wi-Fi + Thread` routers).

We also changed to accept RS messages initiated from the infra interface:
this is for case that multiple Border Routers run on the same host (as we are
trying to support multiple ot-daemons on the same host).

----
TODO:
- [X] CI tests.

### DIFF
--- a/src/core/border_router/router_advertisement.cpp
+++ b/src/core/border_router/router_advertisement.cpp
@@ -187,15 +187,25 @@ bool RouteInfoOption::IsValid(void) const
            (pref == OT_ROUTE_PREFERENCE_LOW || pref == OT_ROUTE_PREFERENCE_MED || pref == OT_ROUTE_PREFERENCE_HIGH);
 }
 
-RouterAdvMessage::RouterAdvMessage(void)
-    : mReachableTime(0)
-    , mRetransTimer(0)
+void RouterAdvMessage::SetToDefault(void)
 {
-    OT_UNUSED_VARIABLE(mReachableTime);
-    OT_UNUSED_VARIABLE(mRetransTimer);
-
     mHeader.Clear();
     mHeader.SetType(Ip6::Icmp::Header::kTypeRouterAdvert);
+    mReachableTime = 0;
+    mRetransTimer  = 0;
+}
+
+const RouterAdvMessage &RouterAdvMessage::operator=(const RouterAdvMessage &aOther)
+{
+    mHeader = aOther.mHeader;
+
+    // Set zero value and let platform do the calculation.
+    mHeader.SetChecksum(0);
+
+    mReachableTime = aOther.mReachableTime;
+    mRetransTimer  = aOther.mRetransTimer;
+
+    return *this;
 }
 
 RouterSolicitMessage::RouterSolicitMessage(void)
--- a/src/core/border_router/routing_manager.cpp
+++ b/src/core/border_router/routing_manager.cpp
@@ -253,8 +253,6 @@ void RoutingManager::RecvIcmp6Message(uint32_t            aInfraIfIndex,
         HandleRouterAdvertisement(aSrcAddress, aBuffer, aBufferLength);
         break;
     case Ip6::Icmp::Header::kTypeRouterSolicit:
-        // Drop Router Solicitations initiated from infra interface.
-        VerifyOrExit(!otPlatInfraIfHasAddress(mInfraIfIndex, &aSrcAddress), error = kErrorDrop);
         HandleRouterSolicit(aSrcAddress, aBuffer, aBufferLength);
         break;
     default:
@@ -623,6 +621,8 @@ void RoutingManager::StartRouterSolicitationDelay(void)
 
     uint32_t randomDelay;
 
+    mRouterAdvMessage.SetToDefault();
+
     mRouterSolicitCount = 0;
 
     static_assert(kMaxRtrSolicitationDelay > 0, "invalid maximum Router Solicitation delay");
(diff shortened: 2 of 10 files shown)


src/core/border_router/routing_manager.cpp
```diff
@@ -253,8 +253,6 @@ void RoutingManager::RecvIcmp6Message(uint32_t            aInfraIfIndex,
         HandleRouterAdvertisement(aSrcAddress, aBuffer, aBufferLength);
         break;
     case Ip6::Icmp::Header::kTypeRouterSolicit:
-        // Drop Router Solicitations initiated from infra interface.
-        VerifyOrExit(!otPlatInfraIfHasAddress(mInfraIfIndex, &aSrcAddress), error = kErrorDrop);
         HandleRouterSolicit(aSrcAddress, aBuffer, aBufferLength);
```
wgtdkp, line 256: Accepts RS messages initiated from the infra interface: this is for case that multiple Border Routers run on the same host (as we are trying to support multiple `ot-daemon`s on the same host).

Notice that we are already dropping the RS messages in `HandleRouterSolicit` when we are discovering routers:
https://github.com/openthread/openthread/blob/2796615cf7e64bc0fc633b6f8c6430dade1b466d/src/core/border_router/routing_manager.cpp#L892

So we will not respond to RS messages sent by ourselves.

src/core/border_router/router_advertisement.cpp
```diff
@@ -187,15 +187,27 @@ bool RouteInfoOption::IsValid(void) const
...
+}
+
+const RouterAdvMessage &RouterAdvMessage::operator=(const RouterAdvMessage &aOther)
+{
+    uint16_t checksum = mHeader.GetChecksum();
+
+    mHeader = aOther.mHeader;
+
+    // There is no necessary to learn the checksum.
+    mHeader.SetChecksum(checksum);
```
jwhui, line 205: Is there value to saving and restoring the checksum?

src/core/border_router/routing_manager.cpp
```diff
@@ -655,18 +655,12 @@ void RoutingManager::SendRouterAdvertisement(const Ip6::Prefix *aNewOmrPrefixes,
...
-    // router for infra link so that hosts on infra link will not create default route
-    // to the Border Router when received RA.
-    routerAdv.SetRouterLifetime(0);
+    uint8_t  buffer[kMaxRouterAdvMessageLength];
+    uint16_t bufferLength = 0;
 
-    OT_ASSERT(bufferLength + sizeof(routerAdv) <= sizeof(buffer));
-    memcpy(buffer, &routerAdv, sizeof(routerAdv));
-    bufferLength += sizeof(routerAdv);
+    OT_ASSERT(bufferLength + sizeof(mInfraIfRouterAdvMessage) <= sizeof(buffer));
```
jwhui, line 661: `bufferLength` is always 0 here. Can we use `static_assert` instead?

src/core/border_router/routing_manager.hpp
```diff
@@ -282,6 +282,11 @@ class RoutingManager : public InstanceLocator
     ExternalPrefix mDiscoveredPrefixes[kMaxDiscoveredPrefixNum];
     uint8_t        mDiscoveredPrefixNum;
 
+    // The RA header and parameters for the infra interface.
+    // This value is initialized with `RouterAdvMessage::SetToDefault`
+    // and updated with RA messages initiated from infra interface.
```
jwhui, line 287: What happens when the infrastructure interface stops advertising?

wgtdkp: Per https://tools.ietf.org/html/rfc2461#section-6.2.5 a Router SHOULD send RAs whith zero Router Lifetime when it ceases advertising. We can reset `mInfraIfRouterAdvMessage` to the default value in case a RA with zero Router Lifetime is received.  I included this in the latest push.

wgtdkp: But there are still cases that the RA daemon may crash before sending RAs with zero Router Lifetime. I am working on the `STALE_RA_TIME` and `Vicarious Router Discovery` described in https://tools.ietf.org/html/draft-lemon-stub-networks-02#section-3.1.1 and with this we can find out that the RA daemon is no longer alive after `STALE_RA_TIME` which is typically smaller than the Router Lifetime.

src/core/border_router/router_advertisement.cpp
wgtdkp: We are currently using zero checksum value to let the platform fill the ICMPv6 checksum. I updated to set zero directly in the latest push.

wgtdkp: Also updated `platform/infra_if.h` to make this clearer.

src/core/border_router/routing_manager.cpp
```diff
@@ -623,6 +621,8 @@ void RoutingManager::StartRouterSolicitationDelay(void)
 
     uint32_t randomDelay;
 
+    mInfraIfRouterAdvMessage.SetToDefault();
```
wgtdkp, line 624: @jwhui Current implementation will start Router Discovery only when all discovered on-link prefixes get expired. That means we will not be able to forget the RA header if:
1. The RA daemon crashed before sending RAs with zero Router Lifetime.
2. There are multiple RA daemons that are both advertising.
3. The crashed RA daemon never come back.

We need to rely on the `STALE_RA_TIME` approach to actively detect the liveness of RA daemons to address this issue.

src/core/border_router/routing_manager.cpp
```diff
@@ -961,6 +955,22 @@ void RoutingManager::HandleRouterAdvertisement(const Ip6::Address &aSrcAddress,
...
+    {
+        if (routerAdvMessage->GetRouterLifetime() == 0)
+        {
+            mRouterAdvMessage.SetToDefault();
+        }
+        else
+        {
+            mRouterAdvMessage = *routerAdvMessage;
+            // TODO: add a timer for invalidating the learned RA parameters
+            // for cases that the other RA daemon crashed or is force killed.
```
wgtdkp, line 970: Will be addressed in https://github.com/openthread/openthread/pull/6514.

abtink (approved):
Looks great! Thanks. 👍

tests/scripts/thread-cert/border_router/test_radvd_coexist.py
```diff
@@ -0,0 +1,184 @@
...
+        pv.summary.show()
+        logging.info(f'vars = {vars}')
+
+        # Ensure the topology is formed correctly
+        pv.verify_attached('Router', 'BR')
+
+        # verify that radvd sends RA messages with PIO.
+        radvd_ra = pkts.filter_eth_src(vars['BR_ETH']) \
+            .filter_icmpv6_nd_ra() \
+            .filter(lambda p: 3 in p.icmpv6.opt.type) \
```
simonlingoogle, line 150: define  constant (in this script) for 3?

tests/scripts/thread-cert/border_router/test_radvd_coexist.py
```diff
@@ -0,0 +1,184 @@
...
+        # Verify that the BR forgets radvd's RA parameters.
+        pkts.filter_eth_src(vars['BR_ETH']) \
+            .filter_icmpv6_nd_ra() \
+            .filter(lambda p: 24 in p.icmpv6.opt.type) \
+            .filter(lambda p: p.icmpv6.nd.ra.router_lifetime == 0) \
+            .filter(lambda p: p.icmpv6.nd.ra.retrans_timer == 0) \
+            .filter(lambda p: p.icmpv6.nd.ra.reachable_time == 0) \
+            .must_next()
+
+        # TODO: verify that the BR forgets the radvd's RA parameters in
```
simonlingoogle, line 179: What's blocking us from implementing this test step.

simonlingoogle (approved):
LGTM :+1: 

with nits

jwhui (approved):

PR #6445, opened 2021-04-14
[mle] clear tag only on a "Child Update Req" Rx from a detached child

abtink:
This commit changes the `MleRouter::HandleChildUpdateRequest()` to
clear the "last rx fragment tag" only if the "Child Update Request" is
from a detached child trying to restore its link with its parent. This
is indicated by the presence of Challenge TLV in the received message.
The "last rx fragment tag" is used for duplicate frame detection when
`MULTI_RADIO` is enabled.

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -2507,7 +2507,14 @@ void MleRouter::HandleChildUpdateRequest(const Message &         aMessage,
     }
 
 #if OPENTHREAD_CONFIG_MULTI_RADIO
-    child->ClearLastRxFragmentTag();
+    // We clear the fragment tag only if the "Child Update Request" is
+    // from a detached child trying to restore its link with its
+    // parent which is indicated by the presence of Challenge TLV in
+    // the message.
+    if (challenge.mLength != 0)
+    {
+        child->ClearLastRxFragmentTag();
+    }
 #endif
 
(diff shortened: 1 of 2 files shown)


Irving-cl (approved):
LGTM 👍

src/core/thread/mle_router.cpp
```diff
@@ -2507,7 +2507,14 @@ void MleRouter::HandleChildUpdateRequest(const Message &         aMessage,
...
 
 #if OPENTHREAD_CONFIG_MULTI_RADIO
-    child->ClearLastRxFragmentTag();
+    // We clear the fragment tag only if the "Child Update Request" is
+    // from a detached child trying to restore its link with its
+    // parent which is indicated by the presence of Challenge TLV in
+    // the message.
+    if (challenge.mLength != 0)
+    {
+        child->ClearLastRxFragmentTag();
```
simonlingoogle, line 2516: I wonder if the parent lose it's `fragment tag` during a quick reboot, how can the Child be aware and clear the last fragment tag.

abtink: It is the same process again. The parent would send the "Child Update Request" to recover the child (child would be marked as in "restoring" state on parent node). 

In new pushed commit I added the same check in `mle.cpp` method (though we can also assume parent would often include "challenge" in Child Update Request" as I guess it mainly used for child.

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6446, opened 2021-04-14
[simulation] set key id and frame counter for retransmissions

konradderda:
Frame counter and key index should be set for every frame that is not yet secured.

### DIFF
--- a/examples/platforms/simulation/radio.c
+++ b/examples/platforms/simulation/radio.c
@@ -644,11 +644,8 @@ static otError radioProcessTransmitSecurity(otRadioFrame *aFrame)
 
     aFrame->mInfo.mTxInfo.mAesKey = key;
 
-    if (!aFrame->mInfo.mTxInfo.mIsARetx)
-    {
-        otMacFrameSetKeyId(aFrame, keyId);
-        otMacFrameSetFrameCounter(aFrame, sMacFrameCounter++);
-    }
+    otMacFrameSetKeyId(aFrame, keyId);
+    otMacFrameSetFrameCounter(aFrame, sMacFrameCounter++);
 #else
     otEXPECT(!aFrame->mInfo.mTxInfo.mIsSecurityProcessed);
 #endif // OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2


konradderda:
Change requested by @edmont and @brianljt comments in https://github.com/openthread/openthread/pull/6427. Thanks for pointing this out! :+1:

edmont (approved):
Thanks.

examples/platforms/simulation/radio.c
```diff
@@ -644,11 +644,8 @@ static otError radioProcessTransmitSecurity(otRadioFrame *aFrame)
 
     aFrame->mInfo.mTxInfo.mAesKey = key;
 
-    if (!aFrame->mInfo.mTxInfo.mIsARetx)
-    {
-        otMacFrameSetKeyId(aFrame, keyId);
-        otMacFrameSetFrameCounter(aFrame, sMacFrameCounter++);
-    }
+    otMacFrameSetKeyId(aFrame, keyId);
+    otMacFrameSetFrameCounter(aFrame, sMacFrameCounter++);
```
abtink, line 648: @konradderda I think the existing code and behavior is correct and what we want.

In the case of retx, the higher layer (`Mac` or higher) should already set the frame counter and key id on the frame.

Also I think spec requires us to reuse the same frame counter for all retx of the same frame.

konradderda: The frame structure will remain the same for retransmissions, there is an 'expect' statement that checks if aFrame->mInfo.mTxInfo.mIsSecurityProcessed is false. This change should not have an impact from the functional perspective, it's more the alignment to https://github.com/openthread/openthread/pull/6427 since on the hardware platform there is a possibility that a transmission is aborted due to external conditions and it will be retransmitted with wrong fc and key index. Here there is no callback from interrupt context but it may confuse someone who compares the code in the future why we check for retransmissions here and not in the nrf528xx's radio platform.

abtink: I noticed that a similar change is already merged in https://github.com/openthread/openthread/pull/6427 for Nordic platform driver. I think we want to change/revert that.

> It seems that we should always set these fields. Moreover, security is processed only once which is controlled by mTxInfo.mIsSecurityProcessed member so checking for retransmission is redundant. Thoughts?

Note that a retx frame may not necessarily be the same `otRadioFrame` instance as the first attempt. So we cannot rely on `mTxInfo.mIsSecurityProcessed` to determine if it is retx.

For example, in an indirect frame tx, the retx can happen much later (since for indirect tx, parent need to wait for a new data poll from child to retx) and the frame is reconstructed again (it won't the same `otRadioFrame` instance as the first attempt) and we alreadt ensure to use the same seq number, frame counter and key id for the retx of indirect frame in `DataPollHandler`:
https://github.com/openthread/openthread/blob/a9d43fac5da39cc108cb3b57eb7df835db611c80/src/core/mac/data_poll_handler.cpp#L189-L207

brianljt: Thank you @abtink for pointing this out. I overlooked the indirect transmission case in the review. I think @konradderda raised a valid fix by moving this `mIsARetransmission` check and setting frame counter and key index in `otPlatRadioTransmit` in PR #6427.

jwhui:
Superseded by #6458

PR #6454, opened 2021-04-15
[posix] RCP(USB CDC ACM) hard reset support

lmaciejonczyk:
This commit handles the situation when RCP(USB CDC) performs hard reset. This commit brings an alternative to pseudo reset by handling the connection with RCP(USB CDC ACM) by reset the socket connection after the device is reset.

When RCP is connected by USB CDC and we reset it, the RCP device is numerated in host operating system. 
So far for RCP(USB CDC) pseudo reset has been performing meaning only specific parts of firmware were reset thanks to  results the connection hasn't been broken. For hard resetting RCP(USB CDC) device the host needs to handle the connection properly.

It uses udev library to check if device enumeration has ended if libudev-dev package is present in the host system otherwise it waits few seconds.

It works based on symlink to the RCP device created by udev in /dev/serial/by-id/ directory.


src/posix/platform/CMakeLists.txt
```diff
@@ -53,6 +53,18 @@ if(NOT OT_CONFIG)
...
+    find_library(LIBUDEV "udev")
+    if (LIBUDEV)
+        set(UDEV_LINK_LIBRARIES "${LIBUDEV}")
+        list(APPEND OT_PLATFORM_DEFINES "OPENTHREAD_USE_UDEV=1")
+    else()
+        message(STATUS "test for udev failed, using sleep instead")
+    endif()
+endif()
+
+list(APPEND OT_PLATFORM_DEFINES "OPENTHREAD_POSIX=1")
```
bukepo, line 66: Is this necessary?

bukepo:
As a network protocol, Thread should just reset software states, which doesn't necessarily need to reset the chip. Resetting the chip also makes it not friendly when BLE coexists on the the chip. Should we change the implementation of `reset` to just re-initialize Thread instance? @jwhui @abtink @simonlingoogle

simonlingoogle:
> As a network protocol, Thread should just reset software states, which doesn't necessarily need to reset the chip. 

Thread is a network protocol, but OpenThread is not merely Thread.
OpenThread is a Thread solution that runs on bare metal / RTOS / POSIX, etc. When running on bare metal, OpenThread provides interfaces (e.g. CLI) to manage the chip because there is no other thing running on the chip. 

> Resetting the chip also makes it not friendly when BLE coexists on the the chip. 

If OpenThread is running as a RTOS `Task`, I would agree that OpenThread should not reset the whole system. RTOS OpenThread solution can freely remove `reset` command. 

> Should we change the implementation of `reset` to just re-initialize Thread instance?

I think we shouldn't.

I would strongly prefer `reset` to reset as much state as it can:
- On bare metal CHIP, `reset` should better use system-specific software reset provided by the chip.
- On POSIX, `reset` should restart the whole process (e.g. by `exec`).

The potential issues of just `reinitializing` the Thread instance:
- `reinitializing` does not give us a clean start that we can trust. It's very difficult to make everything right after a `reinitializing`. We can never say that "Now, OpenThread is now free of `reinitializing` bugs." 
- It's very easy to introduce new `reinitializing` bugs. There are many such bug fix OpenThread PRs. Debugging these bugs has taken us significant time and effort.
---
That being said, I think the current `reset` command is suitable for bare metal and POSIX, but we should also make reinitializing Thread instance work for RTOS (which I think is difficult). 

Thoughts?

simonlingoogle:
@lmaciejonczyk Thanks for this enhancement. I have some questions regarding this PR:

- Can we always reset the socket connection no matter how RCP is reset? This can save us the `RESET_SOCKET` configuration.
- Is it possible to wait for the RCP device reset without using `libudev` , for example by trying to interact with RCP device using via spinel protocol until it's working?
- Could hard reset cause the device filename to change: e.g. from `/dev/ttyUSB0` to `/dev/ttyUSB1`?

lmaciejonczyk:
>     * Can we always reset the socket connection no matter how RCP is reset? This can save us the `RESET_SOCKET` configuration.
We can but it makes sense only for devices connected by USB CDC ACM. For hardware UART the connection is not broken after hard reset.

>     * Is it possible to wait for the RCP device reset without using `libudev` , for example by trying to interact with RCP device via spinel protocol until it's working?
It's not possible. The socket is invalid because the device has changed filename. The connection is broken until we reset socket.

>     * Could hard reset cause the device filename to change: e.g. from `/dev/ttyUSB0` to `/dev/ttyUSB1`?
Yes this is what actually happens when device is connected by USB CDC ACM. Because of that we need to use symlink and wait until device is numerated after the device is reset.

src/posix/platform/CMakeLists.txt
lmaciejonczyk: That's a good catch, thanks.

bukepo:
> > As a network protocol, Thread should just reset software states, which doesn't necessarily need to reset the chip.
> 
> Thread is a network protocol, but OpenThread is not merely Thread.

OpenThread should focus on Thread protocol. OpenThread is not a platform.

> OpenThread is a Thread solution that runs on bare metal / RTOS / POSIX, etc.
> When running on bare metal, OpenThread provides interfaces (e.g. CLI) to manage the chip because there is no other thing running on the chip.

If a platform needs a `reset` command, they can be provided by platform specific commands, like `diag reset`. I don't think it reasonable for a network protocol to request resetting the whole chip.

> When running on POSIX, OpenThread has the legal right to control the whole process.
> 
> > Resetting the chip also makes it not friendly when BLE coexists on the the chip.
> 
> If OpenThread is running as a RTOS `Task`, I would agree that OpenThread should not reset the whole system. RTOS OpenThread solution can freely remove `reset` command.
> 
> > Should we change the implementation of `reset` to just re-initialize Thread instance?
> 
> I think we shouldn't.
> 
> Note that `reset` is supposed to do (chip) `software reset`:

This can be removed.

> 
> ```c++
> /**
>  * This function performs a software reset on the platform, if supported.
>  *
>  * @param[in] aInstance  The OpenThread instance structure.
>  *
>  */
> void otPlatReset(otInstance *aInstance);
> ```
> 
> * On bare metal CHIP, `software reset` should better use system-specific software reset provided by the chip.
> * On POSIX, `software reset` should restart the whole process (e.g. by `exec`).
> 
> The potential issues of just `reinitializing` the Thread instance:
> 
> * `reinitializing` does not give us a clean start that we can trust. It's very difficult to make everything right after a `reinitializing`. We can never say that "Now, OpenThread is now free of `reinitializing` bugs."
> * It's very easy to introduce new `reinitializing` bugs. There are many such bug fix OpenThread PRs. Debugging these bugs has taken us significant time and effort.
> 

I don't think a software's quality should be guaranteed by restart, just like we don't like the operating system don't fixes its bugs, but simply suggests users reboot the computer.

> The current `reset` command is `software reset`. We can certainly have another `thread reset` command for `reinitializing`, but `software reset` is exactly what we want for tests.
> 
> That being said, I think the current `reset` command is suitable for bare metal and POSIX, but we should also make reinitializing Thread instance work for RTOS (which I think is difficult).
> 
> Thoughts?

The `reset` command has been causing issues for a long time. That's why pseudo-reset was introduced. We should improve the quality by fixing bugs instead of ignoring and adding workarounds. Thoughts?

lmaciejonczyk:
> > > As a network protocol, Thread should just reset software states, which doesn't necessarily need to reset the chip.
> > 
> > 
> > Thread is a network protocol, but OpenThread is not merely Thread.
> 
> OpenThread should focus on Thread protocol. OpenThread is not a platform.

Changes from this PR have been done in POSIX platform code of OpenThread repository. Therefore I would say it's platform specific not protocol specific feature. If POSIX platfrom had separated repository I would commit there and here we could focus only on Thread protocol.  

> > OpenThread is a Thread solution that runs on bare metal / RTOS / POSIX, etc.
> > When running on bare metal, OpenThread provides interfaces (e.g. CLI) to manage the chip because there is no other thing running on the chip.
> 
> If a platform needs a `reset` command, they can be provided by platform specific commands, like `diag reset`. I don't think it reasonable for a network protocol to request resetting the whole chip.

OpenThread requests reset platform. How this is performed it's platform specific and we shouldn't discuss here if it's reasonable or not to reset the whole chip. It's the matter of platform coders. What this PR is about is to provide option to make it work if someone chooses to reset the whole chip.

> > ```c++
> > /**
> >  * This function performs a software reset on the platform, if supported.
> >  *
> >  * @param[in] aInstance  The OpenThread instance structure.
> >  *
> >  */
> > void otPlatReset(otInstance *aInstance);
> > ```
> > 
> > 
> > 
> > * On bare metal CHIP, `software reset` should better use system-specific software reset provided by the chip.
> > * On POSIX, `software reset` should restart the whole process (e.g. by `exec`).
> > 
> > The potential issues of just `reinitializing` the Thread instance:
> > 
> > * `reinitializing` does not give us a clean start that we can trust. It's very difficult to make everything right after a `reinitializing`. We can never say that "Now, OpenThread is now free of `reinitializing` bugs."
> > * It's very easy to introduce new `reinitializing` bugs. There are many such bug fix OpenThread PRs. Debugging these bugs has taken us significant time and effort.
> 
> I don't think a software's quality should be guaranteed by restart, just like we don't like the operating system don't fixes its bugs, but simply suggests users reboot the computer.

The main reason which stands behind `void otPlatReset(otInstance *aInstance);` is to provide clean start of the platform including CHIP.  It's far easier here (in OpenThread POSIX platform) to make sure it's working with resetting CHIP than trying provide CHIP `reinitializing` state espacially if it's working under RTOS.

> > The current `reset` command is `software reset`. We can certainly have another `thread reset` command for `reinitializing`, but `software reset` is exactly what we want for tests.
> > That being said, I think the current `reset` command is suitable for bare metal and POSIX, but we should also make reinitializing Thread instance work for RTOS (which I think is difficult).
> > Thoughts?
> 
> The `reset` command has been causing issues for a long time. That's why pseudo-reset was introduced. We should improve the quality by fixing bugs instead of ignoring and adding workarounds. Thoughts?

Using `void otPlatReset(otInstance *aInstance);` as way of "fixing bugs" almost never works out. It simply introduces unstable solution. To provide production quality bugs need to be fixed anyway. But `void otPlatReset(otInstance *aInstance);` is used also for providing clean start state which is the main concern here for CHIPS working under RTOS and communicate by USB CDC ACM. This is not about adding workarounds this is about choosing better solution.

lmaciejonczyk:
@jwhui @abtink Any thoughts here?

abtink:
Sorry for lat reply. Somehow I missed this PR fully. 

While I agree with some of the points from @bukepo  on that we need to be very selective on new APIs in OT (specially if they are OS related functionality and are not related/used by OT core), on the reset API I think the model we currently have makes sense and agree  @lmaciejonczyk and @simonlingoogle  on this.

> OpenThread requests reset platform. How this is performed it's platform specific and we shouldn't discuss here if it's reasonable or not to reset the whole chip. It's the matter of platform coders. What this PR is about is to provide option to make it work if someone chooses to reset the whole chip.

I agree with the above statement. I think it is fine and valid for OT core to request a reset of platform layer. How it is done and implemented is up to platform layer implementation. For example, on NCP the platform code may opt to trigger a full chip reset, but on a posix platform it can behave differently (e.g. `Finalize()` the `otInstance` and then re-initialize it, or choose to restart the process, etc).

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -234,6 +234,9 @@ void RadioSpinel<InterfaceType, ProcessContextType>::Init(bool aResetRadio,
     if (aResetRadio)
     {
         SuccessOrExit(error = SendReset());
+#ifdef OPENTHREAD_POSIX_RESET_SOCKET
```
abtink, line 237: Suggest using `OPENTHREAD_SPINEL_CONFG_RESET_SOCKET` for name to be similar to other config - we can  also define it in  `openthread-spinel-config.h` header file.

Also suggest using `#if` check instead of `#ifdef`.

src/posix/platform/hdlc_interface.cpp
```diff
@@ -656,6 +667,94 @@ void HdlcInterface::HandleHdlcFrame(otError aError)
...
+otError HdlcInterface::WaitForUsbDeviceRenumeration(const char *aRadioUrlPath)
+{
+    otError error = OT_ERROR_NONE;
+
+#ifdef OPENTHREAD_USE_UDEV
+    struct udev *udev = udev_new();
+    if (!udev)
+    {
+        otLogCritPlat("Can't create udev");
+        return OT_ERROR_FAILED;
```
abtink, line 694: From OT style guide we aim to have a single `return` from any method/function (at the end of it). 
Suggest suing `VerifyOrExit()` or other macros for exiting early.

src/posix/platform/hdlc_interface.cpp
```diff
@@ -656,6 +667,94 @@ void HdlcInterface::HandleHdlcFrame(otError aError)
...
+    }
+
+    struct udev_monitor *mon = udev_monitor_new_from_netlink(udev, "udev");
```
abtink, line 697: Again per style guide, 

> - All scoped (i.e. stack) variable declarations should be placed together at the top of the enclosing scope in which they are used.


(some toolchains and analyzers would not be happy if a new variable declaration happens after a jump/goto exit).

src/posix/platform/hdlc_interface.cpp
```diff
@@ -656,6 +667,94 @@ void HdlcInterface::HandleHdlcFrame(otError aError)
...
+    udev_monitor_filter_add_match_subsystem_devtype(mon, "tty", NULL);
+    udev_monitor_enable_receiving(mon);
+    int fd = udev_monitor_get_fd(mon);
```
abtink, line 700: Same for `fd`, `end` and `now`

src/posix/platform/hdlc_interface.cpp
```diff
@@ -656,6 +667,94 @@ void HdlcInterface::HandleHdlcFrame(otError aError)
...
+        now = otPlatTimeGet();
+    } while (end > now);
+
+    error = OT_ERROR_FAILED;
+
+exit:
+    udev_unref(udev);
+
+#else
+    (void)aRadioUrlPath;
```
abtink, line 750: Use `OT_UNUSED_VARIABLE()` instead.

The `(void)aVar` does not work on all tool-chains so we have `OT_UNUSED_VARIABLE()` for this purpose.

src/posix/platform/radio_url.cpp
```diff
@@ -101,6 +101,11 @@ namespace Posix {
 
 RadioUrl::RadioUrl(const char *aUrl)
 {
+    if (aUrl == nullptr)
```
abtink, line 104: Use `VerifyOrExit()`?

abtink (commented):
Thanks @lmaciejonczyk.
Some suggestions below.

src/posix/platform/hdlc_interface.cpp
```diff
@@ -656,6 +667,88 @@ void HdlcInterface::HandleHdlcFrame(otError aError)
...
+
+                if (strncmp(action, "add", strlen("add")) == 0)
+                {
+                    udev_list_entry_foreach(entry, udev_device_get_devlinks_list_entry(dev))
+                    {
+                        const char *name = udev_list_entry_get_name(entry);
+                        if (strncmp(name, aRadioUrlPath, strlen(aRadioUrlPath)) == 0)
+                        {
+                            udev_device_unref(dev);
+                            goto exit;
```
abtink, line 733: `ExitNow()` instead of `goto exit`?

src/posix/platform/hdlc_interface.cpp
```diff
@@ -656,6 +667,88 @@ void HdlcInterface::HandleHdlcFrame(otError aError)
...
+                        }
+                    }
+                }
+                udev_device_unref(dev);
+            }
+        }
+        now = otPlatTimeGet();
+    } while (end > now);
```
abtink, line 741: Suggestion: `now` variable seems to be only used in `while()` check, we can consider removing it and directly using `otPalatTimeGet()`?

src/posix/platform/hdlc_interface.cpp
```diff
@@ -656,6 +667,88 @@ void HdlcInterface::HandleHdlcFrame(otError aError)
...
+
+    error = OT_ERROR_FAILED;
+
+exit:
+    udev_unref(udev);
```
abtink, line 746: Wondering if we need to check against `NULL` here or is calling `udev_unref()` on a possibly `NULL`  pointer fine/ok? Thanks.

abtink (approved):
Looks good to me. Thanks.
Just couple of smaller suggestions below.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -234,6 +234,9 @@ void RadioSpinel<InterfaceType, ProcessContextType>::Init(bool aResetRadio,
     if (aResetRadio)
     {
         SuccessOrExit(error = SendReset());
+#if OPENTHREAD_SPINEL_CONFIG_RESET_CONNECTION
+        SuccessOrExit(error = mSpinelInterface.ResetConnection());
```
jinran-google, line 238: There is another call to `SendReset()` at <https://github.com/openthread/openthread/blob/60c15cea75556db4772a14762fcc6b57138376d9/src/lib/spinel/radio_spinel_impl.hpp#L2232>

There should also be a call to `ResetConnection()` there.

jinran-google (approved):
LGTM. A small suggestion:

src/lib/spinel/openthread-spinel-config.h
```diff
@@ -55,4 +55,14 @@
...
 #endif
 
+/**
+ * @def OPENTHREAD_SPINEL_CONFIG_RESET_CONNECTION
+ *
+ * Define 1 to reset connection after hard resetting RCP(USB CDC ACM) device.
+ *
+ */
+#ifndef OPENTHREAD_SPINEL_CONFIG_RESET_CONNECTION
+#define OPENTHREAD_SPINEL_CONFIG_RESET_CONNECTION 0
```
simonlingoogle, line 65: Can we remove this configuration and just use `OPENTHREAD_PLATFORM_USE_PSEUDO_RESET` to determine if we need to reset connection?

src/posix/platform/hdlc_interface.cpp
```diff
@@ -167,6 +171,8 @@ otError HdlcInterface::Init(const RadioUrl &aRadioUrl)
         ExitNow(error = OT_ERROR_INVALID_ARGS);
     }
 
+    mRadioUrl = aRadioUrl;
```
jinran-google, line 174: Assigning a `RadioUrl` isn't safe. The pointers in `otUrl` will point to the buffer in the old one, and become dangling pointers.

We need to reinitialize the `RadioUrl` here, and maybe reconsider the design of `RadioUrl` to apply RAII.

src/posix/platform/CMakeLists.txt
```diff
@@ -53,6 +53,17 @@ if(NOT OT_CONFIG)
     set(OT_CONFIG ${OT_CONFIG} PARENT_SCOPE)
 endif()
 
+option(OT_RESET_CONNECTION "reset connection after hard resetting RCP(USB CDC ACM) device" OFF)
```
simonlingoogle, line 56:
```suggestion
option(OT_SPINEL_RESET_CONNECTION "reset connection after hard resetting RCP(USB CDC ACM) device" OFF)
```

`OT_RESET_CONNECTION` seems too general.

src/posix/platform/hdlc_interface.cpp
```diff
@@ -656,6 +667,88 @@ void HdlcInterface::HandleHdlcFrame(otError aError)
...
+        ret = select(fd + 1, &fds, NULL, NULL, &tv);
+        if (ret > 0 && FD_ISSET(fd, &fds))
+        {
+            struct udev_device *dev = udev_monitor_receive_device(mon);
+            if (dev)
+            {
+                struct udev_list_entry *entry;
+                const char *            action = udev_device_get_action(dev);
+
+                if (strncmp(action, "add", strlen("add")) == 0)
```
simonlingoogle, line 725: How do we make sure that the added device is the original RCP device? I mean there could be multiple devices plugged on the Host.

simonlingoogle (commented):
> It's not possible. The socket can be invalid because the device could have changed filename. Then the connection is broken and we don't have spinel communication until we reset socket. Alternatively we can just wait blindly short slice of time and proceed with no guarantee that host ends the device numeration.

Can we retry opening the socket until the RCP device is working again? That could save us from adding a dependency on `udev` library.

src/posix/platform/hdlc_interface.hpp
```diff
@@ -230,6 +269,7 @@ class HdlcInterface
     int           mSockFd;
     uint32_t      mBaudRate;
     Hdlc::Decoder mHdlcDecoder;
+    RadioUrl      mRadioUrl;
```
simonlingoogle, line 272:
```suggestion
    const RadioUrl      *mRadioUrl;
```
no need to copy?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -2227,6 +2230,9 @@ void RadioSpinel<InterfaceType, ProcessContextType>::RecoverFromRcpFailure(void)
     if (mResetRadioOnStartup)
     {
         SuccessOrDie(SendReset());
+#if OPENTHREAD_SPINEL_CONFIG_RESET_CONNECTION
+        SuccessOrExit(error = mSpinelInterface.ResetConnection());
```
simonlingoogle, line 2234:
```suggestion
        SuccessOrDie(error = mSpinelInterface.ResetConnection());
```

src/posix/platform/hdlc_interface.cpp
```diff
@@ -656,6 +667,88 @@ void HdlcInterface::HandleHdlcFrame(otError aError)
...
+                {
+                    udev_list_entry_foreach(entry, udev_device_get_devlinks_list_entry(dev))
+                    {
+                        const char *name = udev_list_entry_get_name(entry);
```
simonlingoogle, line 729: handle `name == nullptr` ?

src/posix/platform/hdlc_interface.cpp
```diff
@@ -656,6 +667,88 @@ void HdlcInterface::HandleHdlcFrame(otError aError)
...
+                        if (strncmp(name, aRadioUrlPath, strlen(aRadioUrlPath)) == 0)
```
simonlingoogle, line 730:
```suggestion
                        if (strcmp(name, aRadioUrlPath) == 0)
```

The original code would think that `/dev/ttyACM11`  equals to `/dev/ttyACM1` ?

src/posix/platform/hdlc_interface.cpp
simonlingoogle, line 725:
```suggestion
                if (strcmp(action, "add") == 0)
```
What's the benefit of using `strncmp`?

src/posix/platform/hdlc_interface.cpp
```diff
@@ -656,6 +667,88 @@ void HdlcInterface::HandleHdlcFrame(otError aError)
...
+        FD_SET(fd, &fds);
+
+        ret = select(fd + 1, &fds, NULL, NULL, &tv);
+        if (ret > 0 && FD_ISSET(fd, &fds))
+        {
+            struct udev_device *dev = udev_monitor_receive_device(mon);
+            if (dev)
+            {
+                struct udev_list_entry *entry;
+                const char *            action = udev_device_get_action(dev);
```
simonlingoogle, line 723: check for `action == nullptr`?

lmaciejonczyk (commented):
> > It's not possible. The socket can be invalid because the device could have changed filename. Then the connection is broken and we don't have spinel communication until we reset socket. Alternatively we can just wait blindly short slice of time and proceed with no guarantee that host ends the device numeration.
> 
> Can we retry opening the socket until the RCP device is working again? That could save us from adding a dependency on `udev` library.

First things first. The host communicates with RCP by the "tty" file not a socket. The other thing is that by calling `open` the host OS indicates that's ready for communication by setting UART line control: `DTR`. This helps the RCP device in synchronization and prevent from spinel communication errors, which eventually whould have impact on the host. This makes that polling `open` function simply does not work.
Waiting for kernel event indicating that the USB CDC ACM device is added to host OS guarantees us that after reconnection the communication will work. This is what udev brings in here.
Take notice that we don't have any additional dependency until this functionality is enabled.

src/lib/spinel/openthread-spinel-config.h
lmaciejonczyk: > Can we remove this configuration and just use `OPENTHREAD_PLATFORM_USE_PSEUDO_RESET` to determine if we need to reset connection?


We cannot do that. We don't want to change behaviour for "true" UART communication for which `OPENTHREAD_PLATFORM_USE_PSEUDO_RESET` is disabled as well.

src/posix/platform/hdlc_interface.cpp
lmaciejonczyk: Fixed by overriding assignment operator.

src/posix/platform/hdlc_interface.cpp
lmaciejonczyk: Current `udev_unref()` implementation internally checks against null argument. Nevertheless I've added the check to be safe in any case.

src/posix/platform/hdlc_interface.cpp
lmaciejonczyk: I needed to changed my previous solution because of last changes in: #6543 
Currently changing the reference object RadioUrl to static we prevent from having dangling pointers.

src/lib/spinel/openthread-spinel-config.h
simonlingoogle: Sgtm

src/posix/platform/system.cpp
```diff
@@ -69,8 +69,8 @@ static void processStateChange(otChangedFlags aFlags, void *aContext)
 
 otInstance *otSysInit(otPlatformConfig *aPlatformConfig)
 {
-    otInstance *        instance = nullptr;
-    ot::Posix::RadioUrl radioUrl(aPlatformConfig->mRadioUrl);
+    otInstance *               instance = nullptr;
+    static ot::Posix::RadioUrl radioUrl(aPlatformConfig->mRadioUrl);
```
simonlingoogle, line 73: Add a comment on why this needs to be static?

simonlingoogle (approved):
LGTM :+1: 

Thanks

src/posix/platform/hdlc_interface.cpp
simonlingoogle: You can run `script/make-pretty clang-format` to fix the pretty issue. @lmaciejonczyk

lmaciejonczyk: Thanks

lmaciejonczyk:
@jwhui The PR is approved and CI is green, can it be merged now?

src/posix/platform/hdlc_interface.hpp
```diff
@@ -212,7 +221,37 @@ class HdlcInterface
...
+     *
+     * @retval The file descriptor of newly opened file.
+     */
     int OpenFile(const Url::Url &aRadioUrl);
+
+    /**
+     * This method closes file associated with the file descriptor.
+     *
+     */
+    void CloseFile();
```
jwhui, line 238: nit:
```suggestion
    void CloseFile(void);
```

src/posix/Makefile-posix
```diff
@@ -70,6 +70,7 @@ SERVICE                              ?= 1
 SNTP_CLIENT                          ?= 1
 SRP_CLIENT                           ?= 1
 SRP_SERVER                           ?= 1
+RESET_CONNECTION                     ?= 0
```
jwhui, line 73: nit: maintain alphabetical ordering

src/posix/platform/CMakeLists.txt
```diff
@@ -53,6 +53,17 @@ if(NOT OT_CONFIG)
     set(OT_CONFIG ${OT_CONFIG} PARENT_SCOPE)
 endif()
 
+option(OT_SPINEL_RESET_CONNECTION "reset connection after hard resetting RCP(USB CDC ACM) device" OFF)
```
jwhui, line 56: Update [examples/README.md](https://github.com/openthread/openthread/blob/main/examples/README.md).

jwhui (commented):
Looks good overall. Just a few minor nits.

jwhui (approved):
Thanks! 👍🏼

Issue #6337, opened 2021-03-23
Throughput degradation with Thread 1.2 features enabled


KaSroka:
**Describe the bug**
During testing we have found out that enabling Thread 1.2 features results in decreased throughput in child to child communication.

**To Reproduce**
1. Flash the devkits with image build from latest main branch with Thread 1.2 features enabled. The devkitss are up and running.
2. Form topology: end-device <> router <> end-device (end devices are FTD, using NCP wpantund interface,
router uses CLI interface.
3. On the first devkit host, run the Iperf server, which will receive throughput from client.
Example UDP server command:
$ iperf -V -u -s -p 5000 -i 1
5. On the second devkit host, run the Iperf client, which will be sending throughput to server.
Example UDP client command:
$ iperf -V -u -c <IPv6 of DK1> -p 5000 -b 200k -l 1128 -i 1 -t 60
NOTE: use IPv6 mesh local
6. Verify that the server successfully receives data from the client.
7. Verify that the value of average throughput with Thread 1.2 features is ~3kbit/s.

**Expected behavior**
Throughput in point 7. with image build from same commit but with Thread 1.2 features disabled is ~45kbit/s

**Additional context**
- We are using nRF52840 devkits in this test but it looks like the problem is platform independent - I assume problem lays on MAC layer.
- We are using NCP + wpandund for end devices as this allows us to run iperf easily.
- We are not observing this problem in router <> router communication.

Can someone help with further investigation?

simonlingoogle:
Hi @KaSroka, 

Thanks for catching this issue. 

Could you sniffer the traffic and upload the PCAPs (both 1.1 and 1.2). This would be very helpful for identifying the cause.

luga-ns:
Thread 1.1:
build with:
make -f examples/Makefile-nrf52840 BORDER_AGENT=1 BORDER_ROUTER=1 COAP=1 COAPS=1 COMMISSIONER=1 DHCP6_CLIENT=1 DHCP6_SERVER=1 DNS_CLIENT=1 ECDSA=1 IP6_FRAGM=1 JOINER=1 LINK_RAW=1 MAC_FILTER=1 SERVICE=1 SLAAC=1 SNTP_CLIENT=1 UDP_FORWARD=1 THREAD_VERSION=1.1
results:
test_app_layer_thput_performance_short_child_router_child result comment: {'min': 36.1, 'max': 54.1, 'avg': 50.62, 'ninety_perc': 54.1, 'unit': 'k'}
[...]

PR #6458, opened 2021-04-15
[nrf528xx] set frame counter and key idx in otPlatRadioTransmit()

konradderda:
This change tackles both problems resolved and introduced by https://github.com/openthread/openthread/pull/6427.

For direct transmissions, if the first frame transmission fails, e.g. due to a CCA failure then retransmissions will contain an invalid frame counter. It causes problems especially for long IP packets - the receiver drops a fragment but the transmitter will flood it with the next fragments since it is not aware that the other party dropped a fragment. We have identified that it's a root cause of the problem reported in https://github.com/openthread/openthread/issues/6337.

Previous solution (remove the check if a frame is retransmitted) may cause problems with indirect transmissions where radio frames are created from scratch and then their frame counter would be overridden. That is why the best place for setting frame counter and key index in otPlatRadioTransmit().

### DIFF
--- a/examples/platforms/nrf528xx/src/radio.c
+++ b/examples/platforms/nrf528xx/src/radio.c
@@ -509,6 +509,12 @@ otError otPlatRadioTransmit(otInstance *aInstance, otRadioFrame *aFrame)
     }
 
 #if OPENTHREAD_CONFIG_THREAD_VERSION >= OT_THREAD_VERSION_1_2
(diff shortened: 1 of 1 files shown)


konradderda:
@abtink It's a follow-up on our discussion regarding the previous proposal (thanks a lot for pointing out the problem with indirect transmissions :+1: ).

@edmont, @brianljt I think that radio.c in simulation platform handles both these cases properly and need no changes.

abtink (approved):
Looks great. Thanks. 👍 

Also thanks for always adding nice and detailed PR descriptions 😃

brianljt (approved):
LGTM. Thanks. 👍

LuDuda (approved):

jwhui (approved):

PR #6463, opened 2021-04-16
[scripts] MATN-TC-09: Failure of Primary BBR – Outbound Multicast

superwhd:


tests/scripts/thread-cert/pktverify/layer_fields.py
```diff
@@ -42,7 +42,7 @@
 
 def _auto(v: Union[LayerFieldsContainer, LayerField]):
     """parse the layer field automatically according to its format"""
-    assert not isinstance(v, LayerFieldsContainer) or len(v.fields) == 1, v.fields
+    assert not isinstance(v, LayerFieldsContainer) or len(v.fields) == 1 or v.get_default_value() is not None, v.fields
```
simonlingoogle, line 45: What's the purpose of this change?

tests/scripts/thread-cert/border_router/MATN_09_DefaultBRMulticastForwarding.py
```diff
@@ -0,0 +1,186 @@
...
+        # 4a. Switch off BR_1
+        br1.disable_backbone_router()
+        br1.thread_stop()
+        br1.interface_down()
+        self.simulator.go(180)
+
+        # 4b. BR_2 Detects the missing Primary BBR and becomes the the Leader of
+        # the Thread Network.
+        self.assertEqual('disabled', br1.get_state())
+        self.assertEqual('leader', br2.get_state())
```
simonlingoogle, line 118: I wonder why BR_2 must become leader in this case (even though the Test Spec. clearly states so). 
Can Router_1 become Leader even with lowest Partition ID? @librasungirl

tests/scripts/thread-cert/pktverify/layer_fields.py
superwhd: The assertion failed without this condition. I met a case that v doesn't have field fields.

simonlingoogle: Which field has this problem?

simonlingoogle (approved):
LGTM :+1:

tests/scripts/thread-cert/border_router/MATN_09_DefaultBRMulticastForwarding.py
librasungirl: Configure a larger NETWORK_ID_TIMEOUT value (maybe 160s, larger than default 120s) for ROUTER_1, thus it takes longer for ROUTER_1 to detect the lost of Leader, and when it tries to reattach after NETWORK_ID_TIMEOUT, BR_2 has already formed a new partition.

tests/scripts/thread-cert/pktverify/layer_fields.py
superwhd: `fields`

superwhd: Here's what I got without the added condition: 
```
  File "/usr/local/google/home/handaw/src/openthread/tests/scripts/thread-cert/pktverify/layer_fields.py", line 694, in get_layer_field
    v = _LAYER_FIELDS[field_uri](v)
  File "/usr/local/google/home/handaw/src/openthread/tests/scripts/thread-cert/pktverify/layer_fields.py", line 45, in _auto
    assert not isinstance(v, LayerFieldsContainer) or len(v.fields) == 1, v.fields
AssertionError: [<LayerField thread_nwd.tlv.server.16: 0x00001c00>, <LayerField thread_nwd.tlv.server.16: 0x00001c00>]

The above exception was the direct cause of the following exception:

```

superwhd: Here's what I got without the added condition: 
```
assert not isinstance(v, LayerFieldsContainer) or len(v.fields) == 1, v.fields
AssertionError: [<LayerField thread_nwd.tlv.server.16: 0x00001c00>, <LayerField thread_nwd.tlv.server.16: 0x00001c00>]

The above exception was the direct cause of the following exception:

```

tests/scripts/thread-cert/border_router/MATN_09_DefaultBRMulticastForwarding.py
simonlingoogle: I am not sure if TestHarness did the same `NETWORK_ID_TIMEOUT` configuration. 
BTW, if we configure `NETWORK_ID_TIMEOUT`, there is no need to configure lowest partition ID on `Router_1`.

tests/scripts/thread-cert/pktverify/layer_fields.py
simonlingoogle: Seems there is multiple `thread_nwd.tlv.server.16` field in this layer. 

Maybe we should change the layer field from `_auto` to `_list(_auto)` ? @superwhd

tests/scripts/thread-cert/border_router/MATN_09_DefaultBRMulticastForwarding.py
librasungirl: > I am not sure if TestHarness did the same NETWORK_ID_TIMEOUT configuration.

May check test harness script for confirmation.

> BTW, if we configure NETWORK_ID_TIMEOUT, there is no need to configure lowest partition ID on Router_1.

Guess such joint configurations are to ensure the expecting topology - though Leader periodically send MLE advertisement with new router id sequence to link local multicast address, however whether or not neighboring devices hear it depends on the environment in real on-device test, that is, there is some uncertainty which device assumes the lost of the Leader first. Though the communication is more stable for simulation scenarios, there should be no harm to keep it.

simonlingoogle (approved):
LGTM :+1:

librasungirl (approved):

jwhui (approved):

PR #6465, opened 2021-04-16
[trel] initialize TREL interface in `AfterInit`

simonlingoogle:
to avoid checking `mInitialized` repeatedly in TX tasklet.

### DIFF
--- a/src/core/radio/trel_link.cpp
+++ b/src/core/radio/trel_link.cpp
@@ -65,15 +65,13 @@ Link::Link(Instance &aInstance)
     mRxFrame.SetRadioType(Mac::kRadioTypeTrel);
 #endif
 
-    // `mTxTasklet` is used for initializing the interface (in addition
-    // to handling `Send()` requests). Invoking `Interface::Init()` from
-    // a tasklet ensures that it happens after the constructors for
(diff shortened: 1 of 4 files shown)


src/core/radio/trel_link.cpp
```diff
@@ -65,17 +65,15 @@ Link::Link(Instance &aInstance)
...
 #endif
 
-    // `mTxTasklet` is used for initializing the interface (in addition
-    // to handling `Send()` requests). Invoking `Interface::Init()` from
-    // a tasklet ensures that it happens after the constructors for
-    // `Instance` and all its objects are done, allowing the interface
-    // to safely use any of the core object methods (e.g., get the
-    // randomly generated MAC address).
-
     mTxTasklet.Post();
```
abtink, line 68: I think with `AfterInit()` we don't need to post the tasklet from here anymore.

abtink (approved):
Looks good. Thanks 👍 
One comment about removing the `Post()` from `Link` constructor.

src/core/radio/trel_link.cpp
simonlingoogle, line 68:
```suggestion
```

jwhui (approved):

PR #6469, opened 2021-04-16
[posix] allow using `TREL` without `802.15.4`

simonlingoogle:
This commit allows to run posix app with `TREL` interface only without 802.15.4 radio.

```
$ sudo ot-cli -d 7 -v --trel-interface eth0
```

### DIFF
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -105,7 +105,11 @@ Mac::Mac(Instance &aInstance)
     , mPanId(kPanIdBroadcast)
     , mPanChannel(OPENTHREAD_CONFIG_DEFAULT_CHANNEL)
     , mRadioChannel(OPENTHREAD_CONFIG_DEFAULT_CHANNEL)
+#if OPENTHREAD_CONFIG_RADIO_LINK_IEEE_802_15_4_ENABLE
     , mSupportedChannelMask(Get<Radio>().GetSupportedChannelMask())
+#else
+    , mSupportedChannelMask(Radio::kSupportedChannels)
+#endif
     , mScanChannel(Radio::kChannelMin)
     , mScanDuration(0)
     , mMaxFrameRetriesDirect(kDefaultMaxFrameRetriesDirect)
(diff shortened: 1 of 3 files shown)


src/core/mac/mac.cpp
```diff
@@ -105,7 +105,11 @@ Mac::Mac(Instance &aInstance)
     , mPanId(kPanIdBroadcast)
     , mPanChannel(OPENTHREAD_CONFIG_DEFAULT_CHANNEL)
     , mRadioChannel(OPENTHREAD_CONFIG_DEFAULT_CHANNEL)
+#if OPENTHREAD_CONFIG_RADIO_LINK_IEEE_802_15_4_ENABLE
     , mSupportedChannelMask(Get<Radio>().GetSupportedChannelMask())
+#else
```
abtink, line 110: Do we need this? 
I think we alredy have a default weak implementation for this function which should then be used.

abtink (approved):
Looks good overall. 
One thing on the channel mask.

src/core/mac/mac.cpp
simonlingoogle: If we want to use the default weak implementation, we would also need to put the platform implementation within a `#if OPENTHREAD_CONFIG_RADIO_LINK_IEEE_802_15_4_ENABLE ... #endif` block.

Which one do you think `OPENTHREAD_CONFIG_RADIO_LINK_IEEE_802_15_4_ENABLE == 0` should mean:
- The platform must not provide radio implementations (functions)
- The platform is free to provide radio implementations, but core must not use these functions.

simonlingoogle: > Do we need this?

BTW, it crashes without this because spinel radio was not initialized but we are trying to access it here.

abtink: In general, in a TREL only build (or basically a build in which `15.4` radio link is disabled/not used) we should not need the platform layer to provide an implementation for any of `otPlatRadio` APIs except for the following ones:
https://github.com/openthread/openthread/blob/main/src/core/radio/radio.hpp#L625-L647

These are the APIs that are not directly tied to 15.4 radio and are more tied to device itself.

I think we can consider adding `#if` check in platform later and implement these directly under `TREL` only build and exclude `radio_spinel` module fully. Thoughts?

simonlingoogle:
Close for now.

Issue #6462, opened 2021-04-16
Failed to build POSIX `ot-ctl` with TREL link only (without 802.15.4)


simonlingoogle:
**Describe the bug**

Failed to build on POSIX `ot-ctl` with TREL link only.

**To Reproduce** 
```bash
$ script/cmake-build posix -DOT_TREL=ON -DCMAKE_CXX_FLAGS="-DOPENTHREAD_CONFIG_RADIO_LINK_IEEE_802_15_4_ENABLE=0"
```

**Expected behavior**
Build ot-ctl successfully.

**Console/log output** 
```
[11/483] Building CXX object src/posix/CMakeFiles/ot-ncp.dir/ncp.cpp.o
FAILED: src/posix/CMakeFiles/ot-ncp.dir/ncp.cpp.o 
/usr/bin/c++ -DMBEDTLS_CONFIG_FILE=\"openthread-mbedtls-config.h\" -DOPENTHREAD_POSIX_APP_TYPE=OT_POSIX_APP_TYPE_NCP -DOPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE=1 -Ietc/cmake -I../../etc/cmake -I../../include -I../../src -I../../src/core -I../../src/posix/platform -I../../src/posix/platform/include -I../../third_party/mbedtls/repo/include -DOPENTHREAD_CONFIG_RADIO_LINK_IEEE_802_15_4_ENABLE=0 -g -Werror -pedantic-errors -Wall -Wextra -Wshadow -Wno-c++14-compat -fno-exceptions -std=gnu++11 -MD -MT src/posix/CMakeFiles/ot-ncp.dir/ncp.cpp.o -MF src/posix/CMakeFiles/ot-ncp.dir/ncp.cpp.o.d -o src/posix/CMakeFiles/ot-ncp.dir/ncp.cpp.o -c ../../src/posix/ncp.cpp
In file included from ../../src/core/openthread-core-config.h:80,
                 from ../../src/posix/platform/openthread-posix-config.h:32,
                 from ../../src/posix/ncp.cpp:29:
../../src/core/config/radio_link.h:61:2: error: #error "No radio link type is enabled - at least one radio link type should be enabled"
   61 | #error "No radio link type is enabled - at least one radio link type should be enabled"
      |  ^~~~~
[14/483] Building CXX object src/posix/CMakeFiles/ot-cli.dir/cli.cpp.o
[...]

PR #6474, opened 2021-04-16
[posix] update `ot-cli` to include `ot-config` definitions

abtink:
It also updates `ot-daemon`.

### DIFF
--- a/src/posix/cli.cmake
+++ b/src/posix/cli.cmake
@@ -59,6 +59,7 @@ target_link_libraries(ot-cli
     openthread-spinel-rcp
     ${OT_MBEDTLS}
     ${READLINE_LINK_LIBRARIES}
+    ot-config
 )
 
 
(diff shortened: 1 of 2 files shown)


abtink:
This is intended as a fix for issue https://github.com/openthread/openthread/issues/6462.
However as I mentioned in the above issue with this change, one of the tests keep failing. `POSIX` workflow test `pty-macos OT_DAEMON=on` ([link to a run](https://github.com/openthread/openthread/pull/6474/checks?check_run_id=2366988420)). I leave the more detailed investigation to the folks who know more about `posix` cmake file and what the the related failing test does.

simonlingoogle:
I heard that `include/openthread` headers should not contain any OT core configurations so that applications code never depends on OT core configurations.
Should we reconsider this restriction for `include/openthread`?

@abtink @bukepo @jwhui

bukepo (approved):
LGTM:+1:

We intended not to expose core config headers to apps. I think we can merge this as a quick fix and figure out how to deal with it later.

jwhui (approved):

PR #6476, opened 2021-04-17
[virtual-time-simulator] calculate frame transmission delay

simonlingoogle:


simonlingoogle:
Some test is failing due to this change.
No plan to move forward for now.

PR #6477, opened 2021-04-18
[network-data] check and skip over incomplete TLVs during TLV iteration

abtink:
This commit updates `NetworkData::Iterate()` to check that an entire
TLV (or sub-TLV) is contained within the Network Data (up to its
length) and skips over TLVs that are incomplete. It also updates
`FindTlv()` methods to directly use the stricter check in the
`for()` loop used to iterate over all TLVs.

### DIFF
--- a/src/core/thread/network_data.cpp
+++ b/src/core/thread/network_data.cpp
@@ -131,10 +131,8 @@ const NetworkDataTlv *NetworkData::FindTlv(const NetworkDataTlv *aStart,
 {
     const NetworkDataTlv *tlv;
 
-    for (tlv = aStart; tlv < aEnd; tlv = tlv->GetNext())
+    for (tlv = aStart; (tlv + 1 <= aEnd) && (tlv->GetNext() <= aEnd); tlv = tlv->GetNext())
     {
-        VerifyOrExit((tlv + 1) <= aEnd && tlv->GetNext() <= aEnd, tlv = nullptr);
-
         if (tlv->GetType() == aType)
         {
             ExitNow();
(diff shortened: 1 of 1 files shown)


jwhui:
@abtink , thanks for cleaning this up. Should we also apply the same to instances in [src/core/thread/network_data_leader_ftd.cpp](https://github.com/openthread/openthread/blob/main/src/core/thread/network_data_leader_ftd.cpp)?

abtink:
> @abtink , thanks for cleaning this up. Should we also apply the same to instances in src/core/thread/network_data_leader_ftd.cpp?

@jwhui, I checked the uses of `NetworkDataTlv` in `network_data_leader_ftd`. I think we may already cover all the cases:
- In `Validate()` (and `ValidPrefix()` and `ValidateService()`) we already do similar check and unlike in the `FindTlv()` case where we could use the stricter check directly in `for()` loop, I think in these method, we need to check this seperately. Mainly because, the return (error) value is different here (i.e. if `cur` reaches `end` then we validated every TLV and can return `kErrorNone`, but if we see a TLV going beyond the `end`, we need to return `kErrorParse`). 
- Other methods where we iterate over TLVs (like `Add{Prefix}/Service}()` or `RemoveRloc` and its other flavors) are private and only used after we call `Validate()` on the network data, so we know it is already well-formed. 
- In some other methods we use `FindTlv<>` which would do the same check for us.

BTW, I tried adding a helper method `IsContainedBy(aEnd)` in `NetworkDataTlv` itself (to do the same check and simplify its use), but weirdly it does impact code size. Defining it as inline (which ideally should be exactly the same code), adds [about ~110 extra bytes](https://github.com/abtink/openthread/runs/2371519614?check_suite_focus=true#step:4:5692). I tried also forcing it not be to inlined, it then adds [about ~220 bytes to code size](https://github.com/abtink/openthread/runs/2371926050?check_suite_focus=true#step:4:6388). I recall similar situation when we tried adding similar method in the main `Tlv` class. It seems that when we use `this` pointer (in the `this + 1 <= aEnd` check) in a method, the toolchains cannot inline (and optimize it) as efficiently as the direct `cur + 1 <= aEnd` statement.

jwhui (approved):

PR #6493, opened 2021-04-20
[radio] update mac-layer channel mask when region changes

jinran-google:


src/posix/platform/radio.cpp
```diff
@@ -570,8 +570,15 @@ otError otPlatRadioSetChannelMaxTransmitPower(otInstance *aInstance, uint8_t aCh
 
 otError otPlatRadioSetRegion(otInstance *aInstance, uint16_t aRegionCode)
 {
-    OT_UNUSED_VARIABLE(aInstance);
-    return sRadioSpinel.SetRadioRegion(aRegionCode);
+    otError error;
+
+    SuccessOrExit(error = sRadioSpinel.SetRadioRegion(aRegionCode));
+
+    static_cast<ot::Instance *>(aInstance)->Get<ot::Mac::Mac>().SetSupportedChannelMask(
```
simonlingoogle, line 577: Why is the platform API calling an OT core method?
I think normally we call platform API from OT core.

jinran-google: Should it call `otLinkSetSupportedChannelMask()` here? Or is there a better way to do it?

abtink: @jinran-google I fully agree with Simon, the platform layer should not access internal to core definition/methods (in few cases this rule is broken which is unfortunate and we need to address). I would also not recommend using `otLinkSetSupportedChannelMask()` for this as that is a public OT API and is intended for use by next layer above Thread stack rather than the platform layer. 

Here's what I suggest:

- We already have `otPlatRadioGetSupportedChannelMask()` as a radio platform API in `radio.h` which should return the supported channel mask to OT core (from platform layer).
- You can update of `otPlatRadioGetSupportedChannelMask()` in `posix/platfomr/radio.cpp` to update what it returns based on the region (get the mask from RCP and then filter it more based the region).
- The challenge with this is how to indicate to OT core that channel mask is now changed. 
  - We can add a callback in `radio.h` like `extern void otPlatRadioChannelMaskChanegd(otInstance *aInstance);
` to indicate to OT that channel mask is changed, so that OT can get it back from radio (that would be useful if we expect there may be other ways that the mask may change on radio).
  - The simpler way can be to have this happen in core itself, when `Radio::SetRegionCode()` method is called, it calls a method in `Mac` to notify it that channel mask may have changed and then `Mac` would query radio to get it.

However, in general the addition of `otPlatRadioSetRegion()` makes it rather complicated. 
- If device is already commissioned and operating on a channel (with a set channel mask in , how should we behave if the region is changed causing the channel we are operating on now invalid).

jinran-google: Thanks @abtink 

I agree with the callback approach. For the simpler way, `otPlatRadioSetRegion()` can be called directly from other components (currently CLI and OTBR are calling this function, none of them are from core), so I don't think we can guarantee that all calls to `otPlatRadioSetRegion()` will update `Mac` as well.

simonlingoogle: I think both the callback approach and let caller of `otPlatRadioSetRegion()` update `Mac` as well would work. 

But I wonder why is `otPlatRadioSetRegion` a platform function? In which region is the device operating seems unrelated to what radio platform it is using.

I think we are saying that in some region, the some channels are restrained. Should we implement this in a higher layer, and maybe it's OT user's responsibility to configure channel mask properly?

abtink: @simonlingoogle there may be platform (radio) specific RF calibrations (other than just the channel masks) that are tied to the regulatory requirement of operating in certain region, so the radio platform may need to be aware of the region if it can get changed.

simonlingoogle: @abtink 
Configuring region on radio make sense to me, even though I couldn't find any platform region-specific code in OpenThread, except for simulation.

I think we already proposed 2 approaches:
- a) Platform trigger `otPlatRadioChannelMaskChanegd` to notify OT that radio channel mask is changed
- b) Caller of `otPlatRadioSetRegion` should update OT core channel mask too.

Another approach is that c) we just don't do anything special in OT, but just require the user **MUST** setup channel mask and region correctly.

I actually prefer c) a little bit because it's not hard for user to do that.

abtink: @simonlingoogle , I fully agree and like on option c) if it is possible.

My understanding is there were/are internal discussions on use-cases where region may be allowed to change on an already attached device. I personally think such a use-case should not be allowed (we should require a device to be factory-reset and paired again upon changing its operating region). I think @bukepo or other folks may be able to provide more info (may be in an internal chat if it cannot be shared on GitHub).

PR #6495, opened 2021-04-21
[trel] fix potential compile issues

simonlingoogle:


src/core/mac/mac_frame.cpp
```diff
@@ -1061,49 +1061,37 @@ const TimeIe *Frame::GetTimeIe(void) const
 #if OPENTHREAD_CONFIG_MULTI_RADIO
 uint16_t Frame::GetMtu(void) const
 {
-    uint16_t mtu;
+    uint16_t mtu = OT_RADIO_FRAME_MAX_SIZE;
```
simonlingoogle, line 1064: Fix: `error: 'mtu' may be used uninitialized in this function [-Werror=maybe-uninitialized]`

src/core/mac/mac_frame.cpp
```diff
@@ -1061,49 +1061,37 @@ const TimeIe *Frame::GetTimeIe(void) const
...
+        break;
     }
 
     return mtu;
 }
 
 uint8_t Frame::GetFcsSize(void) const
 {
-    uint8_t fcsSize;
+    uint8_t fcsSize = k154FcsSize;
```
simonlingoogle, line 1083: Fix: `error: ‘fcsSize’ may be used uninitialized in this function [-Werror=maybe-uninitialized]`

src/core/mac/mac_frame.cpp
```diff
@@ -1061,49 +1061,37 @@ const TimeIe *Frame::GetTimeIe(void) const
...
-        break;
-#endif
-
 #if OPENTHREAD_CONFIG_RADIO_LINK_TREL_ENABLE
     case kRadioTypeTrel:
         mtu = Trel::Link::kMtuSize;
         break;
 #endif
 
     default:
```
abtink, line 1074: Since we are removing the `OT_ASSERT`, I suggest we also remove the `default` case fully. 

This way, if a valid `enum` case is not included in `switch`, we will get an warning (on some toolchain).

abtink (approved):
Looks good to me.

One suggestion on removing the `default` case fully along with `OT_ASSERT()`.

src/core/mac/mac_frame.cpp
abtink: I am wondering which toolchain gives this warning/error.

simonlingoogle: See https://github.com/openthread/ot-br-posix/pull/803/checks?check_run_id=2396712279#step:4:2930

src/core/mac/mac_frame.cpp
jwhui: @simonlingoogle , thoughts on removing the default case?

simonlingoogle: To remove default, we need to add `case kRadioTypeIeee802154` back.

simonlingoogle: Verified by https://github.com/openthread/ot-br-posix/pull/803/checks that the current change compiles without issue.

abtink (approved):

jwhui (approved):

PR #6503, opened 2021-04-22
[mesh-forwarder] allow 15.4 broadcast of unsecure link-local multicast

jwhui:


abtink (approved):
LGTM. 👍

krbvroc1:
@jwhui I have been exploring the openthread code and I came across (and temporarily) commented out this same line of code while testing. 

However, shouldn't your line of code be:

```
 if (mle.IsChild() && aMessage.IsLinkSecurityEnabled())
```
rather than negated?

I am still testing, but on the receiver side I cannot receive the UDP I am sending. However, if I use the  otLinkSetPcapCallback() I can get the packet. I do not understand what is blocking the delivery using the normal UDP receive. I did add it as an unsecured UDP port.

krbvroc1:
> However, shouldn't your line of code be:
> 
> ```
>  if (mle.IsChild() && aMessage.IsLinkSecurityEnabled())
> ```
> 
> rather than negated?
> 
> I am still testing, but on the receiver side I cannot receive the UDP I am sending. However, if I use the otLinkSetPcapCallback() I can get the packet. I do not understand what is blocking the delivery using the normal UDP receive. I did add it as an unsecured UDP port.

Just to update you, the receiver side is working ok now, I think it had to do with the order of calling otIp6AddUnsecurePort() in relation to Open/Bind, but I cannot reproduce.

Unfortunately your one line fix isn't working for me unless I change the negation as I have above and then it works.

jwhui:
> Unfortunately your one line fix isn't working for me unless I change the negation as I have above and then it works.

This PR has been updated.

PR #6509, opened 2021-04-23
[dua-manager] avoid null dereference in CoAP response handler

abtink:
This commit changes `DuaManager::HandleDuaResponse()` to use a pointer
to `aMessage` and `aMessageInfo` (instead of reference). Note that in
case of a failure, the `Coap` module may invoke the response handler
with `nullptr` for these parameters.


Irving-cl (approved):
LGTM 👍

simonlingoogle (approved):
LGTM :+1:  Good catch!

simonlingoogle:
The original code is no doubly incorrect.  But I guess it won't cause any issue in actual execution.

librasungirl (approved):
👍

jwhui (approved):

PR #6510, opened 2021-04-25
[joiner] avoid null dereference in CoAP response handler

simonlingoogle:


abtink (approved):
Nice! Thanks. 👍

src/core/meshcop/joiner.cpp
```diff
@@ -525,27 +525,27 @@ void Joiner::HandleJoinerFinalizeResponse(void *               aContext,
...
-void Joiner::HandleJoinerFinalizeResponse(Coap::Message &aMessage, const Ip6::MessageInfo *aMessageInfo, Error aResult)
+void Joiner::HandleJoinerFinalizeResponse(Coap::Message *aMessage, const Ip6::MessageInfo *aMessageInfo, Error aResult)
 {
     OT_UNUSED_VARIABLE(aMessageInfo);
 
     uint8_t state;
 
-    VerifyOrExit(mState == kStateConnected && aResult == kErrorNone && aMessage.IsAck() &&
-                 aMessage.GetCode() == Coap::kCodeChanged);
+    VerifyOrExit(mState == kStateConnected && aResult == kErrorNone && aMessage->IsAck() &&
```
jwhui, line 537: Should we check `aMessage` for null pointer?

jwhui: We have a similar check here:
https://github.com/openthread/openthread/blob/7b07dc503bf7228238451b919d2a0728bcda77e9/src/core/thread/dua_manager.cpp#L555

I think we should be consistent in these two cases. Should we include or exclude the check?

/cc/ @abtink

simonlingoogle: I was trying to minimize the change. 

revised accordingly.

abtink: I think if the `aResult` is `kErrorNone` we should assume that `aMessage` won't be `nullptr`. We can add this in the callback documentation.

I think I added the `OT_ASSERT()` in DUA module to be safe, but I am fine removing it and just relying on `aResult == kErrorNone` check.

jwhui: Looks like @simonlingoogle already updated this PR. If we want to make the doc updates and remove `OT_ASSERT`, we can do that in a separate PR.

simonlingoogle: > We can add this in the callback documentation.

Actually we implied that by saying `NULL if no response was received.` in the current doc. 
Apparently `aResult` can't be `OT_ERROR_NONE` if no response was received.

jwhui (approved):

PR #6513, opened 2021-04-25
[tmf] add `AnycastLocator` module

abtink:
This commit adds a new class `AnycastLocator` which can be used to
locate the closest destination of an anycast IPv6 address (i.e., find
the destination's mesh local EID and RLOC16). The closest destination
is determined based on the the current routing table and path costs
within the Thread mesh.

The implementation uses a CoAP confirmable post request to a newly
added URI path ("a/yl"). The destination IPv6 address of such as
request message is set to the anycast address to be located. The
receiver of the request message sends a CoAP response which includes
the "Mesh Local EID"  and "Thread RLOC16" TLVs.

This commit also adds support this new feature in CLI (adding a new
`locate <anycastaddr>` command).


abtink:
Code size when feature is enabled:
```
  libopenthread-ftd.a_1.2      -524da6d  249001  24    47032  296057
                               +ba595b6  249520  24    47056  296600
                               +/-       +519    0     +24    +543
  libopenthread-mtd.a_1.2      -524da6d  173280  24    28608  201912
                               +ba595b6  173827  24    28632  202483
                               +/-       +547    0     +24    +571


  ot-cli-ftd_1.2               -524da6d  451836  784   65076  517696
                               +ba595b6  452476  784   65100  518360
                               +/-       +640    0     +24    +664
  ot-cli-mtd_1.2               -524da6d  362020  784   46644  409448
                               +ba595b6  362716  784   46668  410168
                               +/-       +696    0     +24    +720
  ot-ncp-ftd_1.2               -524da6d  439296  784   67916  507996
                               +ba595b6  439480  784   67940  508204
                               +/-       +184    0     +24    +208
  ot-ncp-mtd_1.2               -524da6d  352232  784   49492  402508
                               +ba595b6  352432  784   49516  402732
                               +/-       +200    0     +24    +224

```

simonlingoogle:
Curious what's the use case of this feature?

Is the COAP uri `a/yl` going to be added to Thread Specification?

abtink:
> Curious what's the use case of this feature?

The main/first use-case for this is to be able to determine the SRP/DNS server's address (when using the anycast address model). Some related ideas were discussed in Thread TC meetings (in members meeting last week). This PR is prototype of a Thread based solution for this (using simple TMF/CoAP request/response exchange). Another idea discussed is using DNS query for a specific registered name for this purpose. The Thread based solution seems to be preferred due to it being simpler and quicker (to implement and add to spec) and also more generic (i.e, can be use for any anycast address).
 
> Is the COAP uri a/yl going to be added to Thread Specification?

Yes. I expect so.

jwhui:
> Curious what's the use case of this feature?

When using DNS Push, it is desirable to use a stable IPv6 address (rather than anycast).

simonlingoogle:
> > Curious what's the use case of this feature?
> 
> If/when using DNS Push, it is desirable to use a stable IPv6 address (rather than anycast).



> Code size when feature is enabled:
> 
> ```
>   libopenthread-ftd.a_1.2      -524da6d  249001  24    47032  296057
>                                +ba595b6  249520  24    47056  296600
>                                +/-       +519    0     +24    +543
>   libopenthread-mtd.a_1.2      -524da6d  173280  24    28608  201912
>                                +ba595b6  173827  24    28632  202483
>                                +/-       +547    0     +24    +571
> 
> 
>   ot-cli-ftd_1.2               -524da6d  451836  784   65076  517696
>                                +ba595b6  452476  784   65100  518360
>                                +/-       +640    0     +24    +664
>   ot-cli-mtd_1.2               -524da6d  362020  784   46644  409448
>                                +ba595b6  362716  784   46668  410168
>                                +/-       +696    0     +24    +720
>   ot-ncp-ftd_1.2               -524da6d  439296  784   67916  507996
>                                +ba595b6  439480  784   67940  508204
>                                +/-       +184    0     +24    +208
>   ot-ncp-mtd_1.2               -524da6d  352232  784   49492  402508
>                                +ba595b6  352432  784   49516  402732
>                                +/-       +200    0     +24    +224
> ```

Add this feature to size-report?

src/core/thread/anycast_locator.hpp
```diff
@@ -0,0 +1,114 @@
...
+     */
+    Error Locate(const Ip6::Address &aAnycastAddress, Callback aCallback, void *aContext);
+
+    /**
+     * This method indicates whether an earlier request is ongoing.
+     *
+     * @returns TRUE if a previous request is ongoing, FALSE otherwise.
+     *
+     */
+    bool IsRunning(void) const { return (mCallback != nullptr); }
```
simonlingoogle, line 95:
```suggestion
    bool IsBusy(void) const { return (mCallback != nullptr); }
```

src/core/thread/anycast_locator.cpp
```diff
@@ -0,0 +1,163 @@
...
+                 error = kErrorInvalidArgs);
+
+    if (mCallback != nullptr)
+    {
+        // We set the `mCallback` to `nullptr` first and then abort the
+        // ongoing CoAP transaction (if any). Note that the abort method
+        // call can invoke the response callback `HandleResponse()` from
+        // which we invoke the `mCallback`.
+
+        mCallback = nullptr;
```
simonlingoogle, line 72:
```suggestion
```
I think we should call `mCallback` on abort. 

More over, is it better to return `kErrorBusy` in such case (allow only one outstanding query)?

src/core/thread/anycast_locator.cpp
```diff
@@ -0,0 +1,163 @@
...
+    SuccessOrExit(Tlv::Find<ThreadMeshLocalEidTlv>(*aMessage, meshLocalAddress.GetIid()));
+    SuccessOrExit(Tlv::Find<ThreadRloc16Tlv>(*aMessage, rloc16));
+
+    address = &meshLocalAddress;
+
+exit:
+    if (mCallback != nullptr)
+    {
+        mCallback(mContext, address, rloc16);
+        mCallback = nullptr;
```
simonlingoogle, line 129:
```suggestion
        mCallback = nullptr;
        mCallback(mContext, address, rloc16);
```

Set `mCallback` to null first? User may query again in `mCallback`.

src/core/thread/anycast_locator.hpp
simonlingoogle: BTW, this method not used at all.

include/openthread/thread.h
```diff
@@ -869,6 +869,41 @@ void otThreadSetDiscoveryRequestCallback(otInstance *                     aInsta
...
+ * request.
+ *
+ * @param[in] aContext            A pointer to an arbitrary context (provided when callback is registered).
+ * @param[in] aMeshLocalAddress   A pointer to the mesh-local EID of the closest destination of the anycast address.
+ *                                When no destination can be found (or in case of any failure), it will be NULL.
+ * @param[in] aRloc16             The RLOC16 of the destination if found, otherwise invalid RLOC16 (0xfffe).
+ *
+ *
+ */
+typedef void (*otThreadAnycastLocatorCallback)(void *aContext, const otIp6Address *aMeshLocalAddress, uint16_t aRloc16);
```
simonlingoogle, line 883: Is it meaningful to allow returning other addresses (e.g. DUA address is more reliable than ML-EID) ?

src/core/thread/anycast_locator.cpp
```diff
@@ -0,0 +1,163 @@
...
+    uint16_t            rloc16  = Mac::kShortAddrInvalid;
+    const Ip6::Address *address = nullptr;
+    Ip6::Address        meshLocalAddress;
+
+    SuccessOrExit(aResult);
+    OT_ASSERT(aMessage != nullptr);
+
+    meshLocalAddress.SetPrefix(Get<Mle::Mle>().GetMeshLocalPrefix());
+    SuccessOrExit(Tlv::Find<ThreadMeshLocalEidTlv>(*aMessage, meshLocalAddress.GetIid()));
+    SuccessOrExit(Tlv::Find<ThreadRloc16Tlv>(*aMessage, rloc16));
```
simonlingoogle, line 121: Should we save them to EID-RLOC cache as an optimization?
@jwhui @abtink @librasungirl

src/core/thread/anycast_locator.cpp
abtink: I intentionally went for this model over returning  `kErrorBusy` to keep the implementation simple (and small for MTD).

First implementation I had it return `kErrorBusy`, and it adds a lot of additional code and overhead, we need a new `Cancel()` method and then a related public API for this and then a CLI command to cancel an ongoing `locate` command (note that if we try to locate  an anycast address that is not present within mesh,  the existing CoAP tx parameters/mechansim (which I opted to go for to keep it simple) will try hard to retx the message (CoAP has long random wait time and up to 4 retx attempts before giving up which can take up to a minute or more).

I am actually considering to remove the public OT APIs for this (and its CLI support) and keep it as internal feature only.

include/openthread/thread.h
abtink: I think mesh-local unicast address is good for this puporse. Anycast address itself is also mesh-local scope.

src/core/thread/anycast_locator.cpp
abtink: Good idea to add it in `AddressResolver` cache (as a snooped entry). I think it makes sense as a implementation optimization like other snooped cache entries (not something to require in spec).  One thing to note, `AddressResolver` itself is used under FTD only, while `AnycastLocator` can be used under FTD or MTD.

src/core/thread/anycast_locator.cpp
simonlingoogle: I think we need to guarantee that if `otThreadLocateAnycastDestination` returned `OT_ERROR_NONE`, then `mCallback` **MUST** be called in all cases, including aborted.

Thoughts?

abtink: For most cases I agree. But for general case, I would say it depends on the method and how different methods are defined and intended to be used.

For example, if a set of  methods/APIs are intended for use by single module and there is a `Cancel()` method to cancel a previous request (which had its own async callback), then it can be fine for the callback not be invoked on a explicit `Cancel()` call (it's basically like notifying the same module that it called the `Cancel()`). 

BTW, I did not mention it in comment above, but I was/am going to change it in the next push to report error in callback.

abtink:
> Add this feature to size-report?

I thought of this, but I felt it may be better to add to code-size report only when we also add code that uses it (e.g. when adding DNS push support).

simonlingoogle:
> > Add this feature to size-report?
> 
> I thought of this, but I felt it may be better to add to code-size report only when we also add code that uses it (e.g. when adding DNS push support).

SGTM

src/core/thread/anycast_locator.hpp
abtink: I prefer `IsRunning()` (already used the same name in a bunch of other places).
yes, it is not used currently (it's a simple inline getter).

abtink: I actually renamed it to `IsInProgress()`

src/core/thread/anycast_locator.cpp
abtink: Added in new pushed commit.

src/cli/README.md
```diff
@@ -1581,6 +1582,41 @@ Set the link quality on the link to a given extended address.
...
+> netdata show
+Prefixes:
+Routes:
+Services:
+44970 5d c002 s c800
+44970 5d c002 s cc00
+Done
+
+> locate fdde:ad00:beef:0:0:ff:fe00:fc10
+Locate result fdde:ad00:beef:0:a477:dc98:a4e4:71ea RLOC16: 0xcc00
```
simonlingoogle, line 1617:
```suggestion
fdde:ad00:beef:0:a477:dc98:a4e4:71ea RLOC16: 0xcc00
```
nit: remove redundant output to make it easier to parse.

simonlingoogle (approved):
LGTM :+1: 

with nits

src/cli/README.md
abtink: The `locate` command is async and the response can come back later (specially in cases of failure it can be much later). So added the extra `Locate result` to make it clear this is the outcome of a previous `locate` command.

src/core/thread/anycast_locator.cpp
```diff
@@ -0,0 +1,163 @@
...
+                 error = kErrorInvalidArgs);
+
+    message = Get<Tmf::Agent>().NewMessage();
+    VerifyOrExit(message != nullptr, error = kErrorNoBufs);
+
+    SuccessOrExit(error = message->InitAsConfirmablePost(UriPath::kAnycastLocate));
+    SuccessOrExit(error = message->SetPayloadMarker());
+
+    // No payload, so we need to remove CoAP payload marker.
+    IgnoreError(message->SetLength(message->GetLength() - 1));
```
abtink, line 72: Adding a reminder note here that once we have the change from PR https://github.com/openthread/openthread/pull/6516 merged, this line here should be removed

src/cli/README.md
simonlingoogle: SGTM.

src/core/thread/anycast_locator.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    meshLocalAddress.SetPrefix(Get<Mle::Mle>().GetMeshLocalPrefix());
+    SuccessOrExit(Tlv::Find<ThreadMeshLocalEidTlv>(*aMessage, meshLocalAddress.GetIid()));
+    SuccessOrExit(Tlv::Find<ThreadRloc16Tlv>(*aMessage, rloc16));
+
+#if OPENTHREAD_FTD
+    if (Get<AddressResolver>().UpdateCacheEntry(meshLocalAddress, rloc16) != kErrorNone)
+    {
+        Get<AddressResolver>().AddSnoopedCacheEntry(meshLocalAddress, rloc16);
+    }
```
simonlingoogle, line 118: We used more complex conditions for anoter calling to `AddSnoopedCacheEntry`:
```
        if (Get<AddressResolver>().UpdateCacheEntry(ip6Header.GetSource(), aMeshSource.GetShort()) == kErrorNotFound)
        {
            // Thread 1.1 Specification 5.5.2.2: FTDs MAY add/update
            // EID-to-RLOC Map Cache entries by inspecting packets
            // being received. We exclude frames from an MTD child
            // source and verify that the destination is the device
            // itself or an MTD child of the device.

            if (Get<Mle::MleRouter>().IsFullThreadDevice() &&
                !Get<Mle::MleRouter>().IsMinimalChild(aMeshSource.GetShort()) &&
                (aMeshDest.GetShort() == Get<Mac::Mac>().GetShortAddress() ||
                 Get<Mle::MleRouter>().IsMinimalChild(aMeshDest.GetShort())))
            {
                Get<AddressResolver>().AddSnoopedCacheEntry(ip6Header.GetSource(), aMeshSource.GetShort());
            }
        }
```

Is it necessary to check for `Get<Mle::MleRouter>().IsFullThreadDevice() && !Get<Mle::MleRouter>().IsMinimalChild(aMeshSource.GetShort())` here?

abtink: I noticed the same earlier. I actually wanted to simplify and move this common check into `AddressResolver` itself.  Just submitted PR https://github.com/openthread/openthread/pull/6604.

simonlingoogle: > I noticed the same earlier. I actually wanted to simplify and move this common check into `AddressResolver` itself. Just submitted PR #6604.

Same thought.

src/core/thread/anycast_locator.hpp
```diff
@@ -0,0 +1,114 @@
...
+     * If a previous `Locate()` request is ongoing, a subsequent call to this method will cancel and replace the
+     * earlier request.
+     *
+     * @param[in] aAnycastAddress   The anycast address to locate the closest destination of.
+     * @param[in] aCallback         The callback handler to report the result.
+     * @param[in] aContext          An arbitrary context used with @p aCallback.
+     *
+     * @retval kErrorNone         The request started successfully. @p aCallback will be invoked to report the result.
+     * @retval kErrorNoBufs       Out of buffer to prepare and send the request message.
+     * @retval kErrorInvalidArgs  The @p aAnycastAddress is not a valid anycast address or @p aCallback is `nullptr`.
```
librasungirl, line 84: Taking the same example as in the README, before srp server is enabled, it would take 60+ s to return timeout for the CON CoAP transaction if locating a valid ALOC which does not belong to any node yet.

```
> locate fd11:269a:9cbf:892d:0:ff:fe00:fc10
> Locate result
Error 28: ResponseTimeout
```

In fact we can learn whether the valid ALOC is available from network data in the mesh, similar as the mesh forwarder component 
https://github.com/openthread/openthread/blob/e0a1987336de203cdae30f06ff16cc2309817387/src/core/thread/mesh_forwarder_ftd.cpp#L546-L586

Maybe we can do some optimization here to only initiate `a/yl` CON CoAP msg when we know there must be some device has this valid ALOC?

src/cli/cli.cpp
```diff
@@ -2589,6 +2589,48 @@ otError Interpreter::ProcessLinkMetricsProbe(uint8_t aArgsLength, Arg aArgs[])
...
 #endif // OPENTHREAD_CONFIG_MLE_LINK_METRICS_ENABLE
 
+#if OPENTHREAD_CONFIG_TMF_ANYCAST_LOCATOR_ENABLE
+
+otError Interpreter::ProcessLocate(uint8_t aArgsLength, Arg aArgs[])
+{
+    otError      error = OT_ERROR_INVALID_ARGS;
+    otIp6Address anycastAddress;
+
+    VerifyOrExit(aArgsLength == 1);
```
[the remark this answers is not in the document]
abtink: Good suggestion. Updated in the new pushed commit.

src/core/thread/anycast_locator.hpp
abtink: I noticed the longer timeout when there is no response. `AnycastLocator` uses the default COAP tx paramters which does up to 4 retries with some long randomr etry wait time, so the overall retries can take longer (up to even minutes). We can consider using a different `Coap::TxParameters` to make the timeout shorter but I am not sure if there is any advantages for this.

Related to determining if the anycast address is available before sending the CoAP message, we can consider it but again I am not sure of its benefit and whether the the complexity and code size impact it can add is worth such an optimization:
- `AnycastLocator` module needs to run on all types of builds/devices (MTD and FTD), so it's good to keep it lean (smaller code size).
- The code in `mesh_forwarder_ftd` is under FTD only and I think it requires that device to know the full network data (which may not be the case on an MTD build).

src/core/thread/anycast_locator.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    SuccessOrExit(aError);
+    OT_ASSERT(aMessage != nullptr);
+
+    meshLocalAddress.SetPrefix(Get<Mle::Mle>().GetMeshLocalPrefix());
+    SuccessOrExit(Tlv::Find<ThreadMeshLocalEidTlv>(*aMessage, meshLocalAddress.GetIid()));
+    SuccessOrExit(Tlv::Find<ThreadRloc16Tlv>(*aMessage, rloc16));
+
+#if OPENTHREAD_FTD
+    if (Get<AddressResolver>().UpdateCacheEntry(meshLocalAddress, rloc16) != kErrorNone)
```
abtink, line 115: Adding a note here to remember to update this when/if PR https://github.com/openthread/openthread/pull/6604 is merged.

jwhui: Merged :)

abtink: Thanks. Just rebased and updated it

src/core/thread/anycast_locator.hpp
simonlingoogle: > I noticed the longer timeout when there is no response. `AnycastLocator` uses the default COAP tx paramters which does up to 4 retries with some long randomr etry wait time, so the overall retries can take longer (up to even minutes). We can consider using a different `Coap::TxParameters` to make the timeout shorter but I am not sure if there is any advantages for this.

I think we should use the default coap to parameters.

> - `AnycastLocator` module needs to run on all types of builds/devices (MTD and FTD), so it's good to keep it lean (smaller code size).
> 
> - The code in `mesh_forwarder_ftd` is under FTD only and I think it requires that device to know the full network data (which may not be the case on an MTD build).

Checking network data for FTF only seems to be a good optimization. But I am wondering how often a FTD needs to use AnycastLocator.

librasungirl: Just think of some questions - how does the caller know the ALOC it want to locate?
Should the general use case is the caller monitors the network data event and constructs the ALOC address when there are services (service id, context id, or commissioner session id etc) it cares and then `locate`? 

If we have such assumption, I think we should leave it as is; otherwise maybe we can consider this check optimization for Thread Devices (FTD or MTD) that requests full network data, avoiding unnecessary CoAP transaction and retx (which occupies limited message buffer resources for a while) and return result sooner, with the cost of iterating the network data once (?) (since we can know what services the ALOC is according to the address construction rule).

librasungirl (approved):

src/core/thread/anycast_locator.hpp
simonlingoogle: If an FTD with full Network Data sends `a/yl` for an invalid ALOC, will it fail immediately because no route can be found?
If so, there won't be a long CoAP transaction with multiple retx, anyway. 

Another thought is that when a `a/yl` message is sent from an MTD child to its parent for an invalid ALOC, then the parent can't find any route. Will it send an `ICMPv6 Destination Unreachable` back to MTD? And will the MTD stop CoAP retx receiving `ICMPv6 Destination Unreachable` ?

abtink: > Should the general use case is the caller monitors the network data event and constructs the ALOC address when there are services (service id, context id, or commissioner session id etc) it cares and then locate? 
>
> If we have such assumption, I think we should leave it as is; 

I agree. I think this is the intended use-case and assumption for this. The main use-case for now for this feature is DNS push related where we discover the anycast address itself from the network-data  (from the "DNS/SRP Service" entry in in network data) and then want to locate the destination of the anycast address.

> Will it send an ICMPv6 Destination Unreachable back to MTD? And will the MTD stop CoAP retx receiving ICMPv6 Destination Unreachable ?

This is an interesting idea, but I think we may want to consider it as a behavior to add in `Coap` itself (i.e., if we get an ICMPv6 "des unreachable" error we stop a CoAP transaction (not perform retx anymore). I think we can leave it for a future enhancement and PR.

src/core/thread/anycast_locator.cpp
```diff
@@ -0,0 +1,160 @@
...
+    Get<Tmf::Agent>().AddResource(mAnycastLocate);
+}
+
+Error AnycastLocator::Locate(const Ip6::Address &aAnycastAddress, Callback aCallback, void *aContext)
+{
+    Error            error   = kErrorNone;
+    Coap::Message *  message = nullptr;
+    Ip6::MessageInfo messageInfo;
+
+    VerifyOrExit((aCallback != nullptr) && Get<Mle::Mle>().IsAnycastLocator(aAnycastAddress),
```
simonlingoogle, line 62: Small question:
Is it useful to use this mechanism to convert RLOC to ML-EID?

As a purpose of this functionality, we want to convert ALOC to ML-EID because ML-EID is more stable. 
The same should apply to RLOC. 

Should we allow converting either ALOC or RLOC to ML-EID ? This enhancement should not add any code size, but may even reduce code size.

Thoughts?

@abtink @librasungirl @jwhui

librasungirl: I think RLOC is generally used by the Thread core stack, not upper layer application. 

@simonlingoogle is there use case to associate RLOC with ML-EID in your mind?

src/core/thread/anycast_locator.hpp
librasungirl: > If an FTD with full Network Data sends `a/yl` for an invalid ALOC, will it fail immediately because no route can be found?
> If so, there won't be a long CoAP transaction with multiple retx, anyway.

The upper layer CoAP retx is independent from no route decision at mesh forwarder layer, that is,  CoAP layer only knows there is no response (either due to not sent, or not arrived etc), then retx.




> > Will it send an ICMPv6 Destination Unreachable back to MTD? And will the MTD stop CoAP retx receiving ICMPv6 Destination Unreachable ?
> 
> This is an interesting idea, but I think we may want to consider it as a behavior to add in `Coap` itself (i.e., if we get an ICMPv6 "des unreachable" error we stop a CoAP transaction (not perform retx anymore). I think we can leave it for a future enhancement and PR.

Agree 👍

src/core/thread/anycast_locator.cpp
abtink: We can consider using it for RLOC as well but I am not sure if we have use-case for it. I prefer limiting it to `Anycast` for now (unless we have a good reason/use-case to extend it to RLOC).

> As a purpose of this functionality, we want to convert ALOC to ML-EID because ML-EID is more stable.
The same should apply to RLOC.

The main goal is to determine the closest destination of ALOC address (since we can have multiple destinations for the same anycasy address). I don't think it is related to ML-EID being more stable.

When using `ML-EID` as destination of message, device needs to address query to discover the RLOC. So if we have RLOC of destination, it sounds more reasonable to use it directly than to try to get the related ML-EID and then use it instead.

simonlingoogle: > When using `ML-EID` as destination of message, device needs to address query to discover the RLOC. So if we have RLOC of destination, it sounds more reasonable to use it directly than to try to get the related ML-EID and then use it instead.

Considering connection protocols like TCP, it would be better to use more stable addresses like ML-EID or OMR address than to use RLOC, which changes as the topology changes. 

That being said, I am fine to consider RLOC when we have a use-case for it.

src/core/thread/anycast_locator.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include "coap/coap.hpp"
+#include "common/locator.hpp"
+#include "common/non_copyable.hpp"
+#include "net/ip6_address.hpp"
+
+namespace ot {
+
+/**
+ * This class implement Anycast Locator functionality which allows caller to determine the mesh local EID and RLOC16 of
```
jwhui, line 49:
```suggestion
 * This class implements Anycast Locator functionality which allows caller to determine the mesh local EID and RLOC16 of
```

src/core/thread/anycast_locator.hpp
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * If a previous `Locate()` request is ongoing, a subsequent call to this method will cancel and replace the
+     * earlier request.
+     *
+     * @param[in] aAnycastAddress   The anycast address to locate the closest destination of.
+     * @param[in] aCallback         The callback handler to report the result.
+     * @param[in] aContext          An arbitrary context used with @p aCallback.
+     *
+     * @retval kErrorNone         The request started successfully. @p aCallback will be invoked to report the result.
+     * @retval kErrorNoBufs       Out of buffer to prepare and send the request message.
```
jwhui, line 83:
```suggestion
     * @retval kErrorNoBufs       Out of buffers to prepare and send the request message.
```

jwhui (commented):
Looks great overall. Can we add a CI test?

abtink:
> Looks great overall. Can we add a CI test?

In the latest push, added `test_anycast_locator.py` to cover the behavior of `AnycastLocator`, locating leader ALOC and also service ALOC (when there is a signle node adding the service and when there are two nodes adding the same service and verifying that the closest destination is indeed located).

abtink:
Rebased on `main` and fixed the conflicts.

src/cli/README.md
```diff
@@ -1612,6 +1613,59 @@ Set the link quality on the link to a given extended address.
 Done
 ```
 
+### locate
+
+Gets the current state (in-progress or idle) of anycast locator.
```
librasungirl, line 1618:
```suggestion
Gets the current state (`In Progress` or `Idle`) of anycast locator.
```

tests/scripts/thread-cert/node.py
```diff
@@ -1058,6 +1058,25 @@ def _parse_srp_client_service(self, line: str):
...
         return dict(zip(keys, values))
 
+    def locate(self, anycast_addr):
+        cmd = 'locate ' + anycast_addr
+        self.send_command(cmd)
+        self.simulator.go(5)
+        return self._parse_locate_result(self._expect_command_output(cmd)[0])
+
+    def _parse_locate_result(self, line: str):
+        """Parse anycast locate result as list of ml-edi and rloc16.
```
librasungirl, line 1068:
```suggestion
        """Parse anycast locate result as list of ml-eid and rloc16.
```

tests/scripts/thread-cert/test_anycast_locator.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class AnycastLocator(thread_cert.TestCase):
+    USE_MESSAGE_FACTORY = False
+    SUPPORT_NCP = False
+
+    TOPOLOGY = {
+        LEADER: {
+            'name': 'LEADER',
+            'masterkey': '00112233445566778899aabbccddeeff',
```
librasungirl, line 63: nit: `masterkey` is renamed as `networkkey`, and could be removed as it is default in config https://github.com/openthread/openthread/blob/deb2a4c97a95242ed858ddef5d79e00271f7d798/tests/scripts/thread-cert/config.py#L86-L103

tests/scripts/thread-cert/test_anycast_locator.py
```diff
@@ -0,0 +1,171 @@
...
+            self.assertEqual(int(result[1], 0), leader.get_addr16())
+
+        #
+        # 2. Add a service on router4 and locate its ALOC
+        #
+
+        router4.add_service(SRV_ENT_NUMBER, SRV_SERVICE_DATA, SRV_SERVER_DATA)
+        router4.register_netdata()
+        self.simulator.go(5)
+        servies = leader.get_services()
```
librasungirl, line 132: typo: `servies`->`services`

src/core/thread/anycast_locator.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include "thread/thread_tlvs.hpp"
+#include "thread/uri_paths.hpp"
+
+namespace ot {
+
+AnycastLocator::AnycastLocator(Instance &aInstance)
+    : InstanceLocator(aInstance)
+    , mCallback(nullptr)
+    , mContext(nullptr)
+    , mAnycastLocate(UriPath::kAnycastLocate, HandleAnycastLocate, this)
```
librasungirl, line 50: Should we consider distinguishing the requester and responder for this feature?  I am thinking only server-capable device (FTD or prefix/route/service provider etc) has the responsibility to answer. Mostly MTD only need to request locate and not necessary to respond.

Thoughts?

abtink: Sorry! somehow I forgot about this PR. It's rebased and updated now. PTAL.

@librasungirl good suggestion. I added a config to allow "sending response" to be enabled/disabled separately. I set the default for it to be enabled (i.e,. when `ANYCAST_LOCATOR` feature is enabled then by default we can request and respond). The code size impact of "response" part is rather small. It's about ~170 additional bytes to have support for "sending response":
```
  text    data     bss     dec     hex filename
 359492     416   49764  409672   64048 build/bin/ot-cli-mtd (full feature ANYCAST_LOCATOR)
 359324     416   49748  409488   63f90 build/bin/ot-cli-mtd (ANYCAST_LCOATOR request only)
```

src/cli/cli.cpp
```diff
@@ -2554,6 +2554,51 @@ otError Interpreter::ProcessLinkMetricsProbe(Arg aArgs[])
...
+
+    if (aArgs[0].IsEmpty())
+    {
+        OutputLine(otThreadIsAnycastLocateInProgress(mInstance) ? "In Progress" : "Idle");
+        ExitNow(error = OT_ERROR_NONE);
+    }
+
+    SuccessOrExit(error = aArgs[0].ParseAsIp6Address(anycastAddress));
+    SuccessOrExit(error = otThreadLocateAnycastDestination(mInstance, &anycastAddress, HandleLocateResult, this));
+
```
simonlingoogle, line 2572: error = kErrorPending;

src/cli/cli.cpp
```diff
@@ -2554,6 +2554,51 @@ otError Interpreter::ProcessLinkMetricsProbe(Arg aArgs[])
...
+    if (aError == OT_ERROR_NONE)
+    {
+        OutputIp6Address(*aMeshLocalAddress);
+        OutputLine(" RLOC16: 0x%04x", aRloc16);
+    }
+    else
+    {
+        OutputLine("- Error:%s", otThreadErrorToString(aError));
+    }
+}
```
simonlingoogle, line 2598:
```suggestion

        OutputResult(aError);
}
```

simonlingoogle (approved):
Now, we can make the command blocking.

src/cli/cli.cpp
abtink: We had it as blocking earlier (returning `kErrorPending` and then doing `OutputResult()` from callback). But in the case there is no response, the CoAP timeout is rather longe (up to minutes), and CLI will be blocked during all of this. So I changed it to be non-blocking.

simonlingoogle: Can we use a shorter timeout?

I think the current implementation won't work with `ot-ctl`:
```
# ot-ctl locate fdde:ad00:beef:0:0:ff:fe00:fc00
Done
#
```

abtink: We can use the newly added CLI `mTimer` for this (to wait for a shorter time ~2second in the CLI) and then make the command be blocking again.  I will change this in next push (with this we can use also change the output format (make it shorter and just output the RLOC16)).

The core implementation uses the default `Coap` with its setting (to keep the code simpler and also has smaller code size). So would be good to keep the core implementation as is.

src/cli/cli.cpp
```diff
@@ -2554,6 +2557,55 @@ otError Interpreter::ProcessLinkMetricsProbe(Arg aArgs[])
...
+                                     otError             aError,
+                                     const otIp6Address *aMeshLocalAddress,
+                                     uint16_t            aRloc16)
+{
+    static_cast<Interpreter *>(aContext)->HandleLocateResult(aError, aMeshLocalAddress, aRloc16);
+}
+
+void Interpreter::HandleLocateResult(otError aError, const otIp6Address *aMeshLocalAddress, uint16_t aRloc16)
+{
+    VerifyOrExit(mLocateInProgress);
```
simonlingoogle, line 2593:
```suggestion
    VerifyOrExit(mLocateInProgress);
    
    mLocateInProgress = false;
```
IIUC

simonlingoogle (approved):
LGTM :+1:

src/cli/cli.cpp
abtink: Yes. We do need this. Thanks for catching it.

abtink:
Rebased on `main` (so to use `AsCoreType()` in the newly added API functions).

jwhui (approved):

PR #6514, opened 2021-04-26
[border-router] add vicarious router solicitation

wgtdkp:
This commit adds vicarious router solicitation which is described in
https://www.ietf.org/archive/id/draft-lemon-stub-networks-02.html#section-3.1.1

### DIFF
--- a/src/core/border_router/routing_manager.cpp
+++ b/src/core/border_router/routing_manager.cpp
@@ -65,9 +65,12 @@ RoutingManager::RoutingManager(Instance &aInstance)
     , mAdvertisedOmrPrefixNum(0)
     , mAdvertisedOnLinkPrefix(nullptr)
     , mDiscoveredPrefixNum(0)
+    , mTimeRouterAdvMessageLastUpdate(TimerMilli::GetNow())
     , mDiscoveredPrefixInvalidTimer(aInstance, HandleDiscoveredPrefixInvalidTimer)
+    , mDiscoveredPrefixStaleTimer(aInstance, HandleDiscoveredPrefixStaleTimer)
     , mRouterAdvertisementTimer(aInstance, HandleRouterAdvertisementTimer)
     , mRouterAdvertisementCount(0)
+    , mVicariousRouterSolicitTimer(aInstance, HandleVicariousRouterSolicitTimer)
     , mRouterSolicitTimer(aInstance, HandleRouterSolicitTimer)
     , mRouterSolicitCount(0)
     , mRoutingPolicyTimer(aInstance, HandleRoutingPolicyTimer)
@@ -216,10 +219,12 @@ void RoutingManager::Stop(void)
     memset(mDiscoveredPrefixes, 0, sizeof(mDiscoveredPrefixes));
     mDiscoveredPrefixNum = 0;
     mDiscoveredPrefixInvalidTimer.Stop();
+    mDiscoveredPrefixStaleTimer.Stop();
 
     mRouterAdvertisementTimer.Stop();
     mRouterAdvertisementCount = 0;
 
+    mVicariousRouterSolicitTimer.Stop();
     mRouterSolicitTimer.Stop();
     mRouterSolicitCount = 0;
 
@@ -492,7 +497,7 @@ const Ip6::Prefix *RoutingManager::EvaluateOnLinkPrefix(void)
     const Ip6::Prefix *smallestOnLinkPrefix = nullptr;
 
     // We don't evaluate on-link prefix if we are doing Router Solicitation.
-    VerifyOrExit(!mRouterSolicitTimer.IsRunning());
+    VerifyOrExit(!mRouterSolicitTimer.IsRunning(), newOnLinkPrefix = mAdvertisedOnLinkPrefix);
 
     for (uint8_t i = 0; i < mDiscoveredPrefixNum; ++i)
     {
@@ -536,7 +541,7 @@ const Ip6::Prefix *RoutingManager::EvaluateOnLinkPrefix(void)
             otLogInfoBr("EvaluateOnLinkPrefix: there is already smaller on-link prefix %s on interface %u",
                         smallestOnLinkPrefix->ToString().AsCString(), mInfraIfIndex);
 
-            // TODO: we removes the advertised on-link prefix by setting valid lifetime for PIO.
+            // TODO: we remove the advertised on-link prefix by setting zero valid lifetime for PIO,
             // But SLAAC addresses configured by PIO prefix will not be removed immediately (
             // https://tools.ietf.org/html/rfc4862#section-5.5.3). This leads to a situation that
             // a WiFi device keeps using the old SLAAC address but a Thread device cannot reach to
@@ -617,19 +622,21 @@ void RoutingManager::StartRoutingPolicyEvaluationDelay(void)
 // between 0 and kMaxRtrSolicitationDelay.
 void RoutingManager::StartRouterSolicitationDelay(void)
 {
-    OT_ASSERT(mAdvertisedOnLinkPrefix == nullptr);
-
     uint32_t randomDelay;
 
-    mRouterAdvMessage.SetToDefault();
+    VerifyOrExit(!mRouterSolicitTimer.IsRunning() && mRouterSolicitCount == 0);
 
-    mRouterSolicitCount = 0;
+    mVicariousRouterSolicitTimer.Stop();
 
     static_assert(kMaxRtrSolicitationDelay > 0, "invalid maximum Router Solicitation delay");
     randomDelay = Random::NonCrypto::GetUint32InRange(0, Time::SecToMsec(kMaxRtrSolicitationDelay));
 
     otLogInfoBr("start Router Solicitation, scheduled in %u milliseconds", randomDelay);
+    mTimeRouterSolicitStart = TimerMilli::GetNow();
     mRouterSolicitTimer.Start(randomDelay);
+
+exit:
+    return;
 }
 
 Error RoutingManager::SendRouterSolicitation(void)
@@ -802,6 +809,27 @@ void RoutingManager::HandleRouterAdvertisementTimer(void)
     EvaluateRoutingPolicy();
 }
 
+void RoutingManager::HandleVicariousRouterSolicitTimer(Timer &aTimer)
+{
+    aTimer.Get<RoutingManager>().HandleVicariousRouterSolicitTimer();
+}
+
+void RoutingManager::HandleVicariousRouterSolicitTimer(void)
+{
+    otLogInfoBr("vicarious router solicitation time out");
+
+    for (uint8_t i = 0; i < mDiscoveredPrefixNum; ++i)
+    {
+        ExternalPrefix &prefix = mDiscoveredPrefixes[i];
+
+        if (prefix.mTimeLastUpdate <= mTimeVicariousRouterSolicitStart)
+        {
+            StartRouterSolicitationDelay();
+            break;
+        }
+    }
+}
+
 void RoutingManager::HandleRouterSolicitTimer(Timer &aTimer)
 {
     aTimer.Get<RoutingManager>().HandleRouterSolicitTimer();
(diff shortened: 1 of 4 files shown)


wgtdkp:
Marking WIP as the test case is failing.

abtink (approved):
Looks great! Thanks. 👍

src/core/border_router/routing_manager.hpp
```diff
@@ -154,36 +154,52 @@ class RoutingManager : public InstanceLocator
...
-        kMaxRaDelayTime              = 500, // The maximum delay of sending RA after receiving RS. In milliseconds.
-        kRtrSolicitationInterval     = 4,   // The interval between Router Solicitations. In seconds.
-        kMaxRtrSolicitationDelay     = 1,   // The maximum delay for initial solicitation. In seconds.
-        kMaxRoutingPolicyDelay       = 1,   // The maximum delay for routing policy evaluation. In seconds.
+        kMaxInitRtrAdvInterval       = 16,   // Maximum Initial Router Advertisement Interval. In seconds.
+        kMaxRaDelayTime              = 500,  // The maximum delay of sending RA after receiving RS. In milliseconds.
+        kRtrSolicitationInterval     = 4,    // The interval between Router Solicitations. In seconds.
+        kMaxRtrSolicitationDelay     = 1,    // The maximum delay for initial solicitation. In seconds.
+        kMaxRoutingPolicyDelay       = 1,    // The maximum delay for routing policy evaluation. In seconds.
+        kRtrAdvStaleTime             = 1500, // The period of Router Solicitations. In seconds.
```
jwhui, line 174: This is not really the period of RS messages. In particular, when there are multiple on-link prefixes, we should enter router discover after the last non-stale on-link prefix becomes stale.

src/core/border_router/routing_manager.cpp
```diff
@@ -617,19 +620,21 @@ void RoutingManager::StartRoutingPolicyEvaluationDelay(void)
...
 {
-    OT_ASSERT(mAdvertisedOnLinkPrefix == nullptr);
-
     uint32_t randomDelay;
 
-    mRouterAdvMessage.SetToDefault();
+    VerifyOrExit(!mRouterSolicitTimer.IsRunning() && mRouterSolicitCount == 0);
 
-    mRouterSolicitCount = 0;
+    mVicariousRouterSolicitTimer.Stop();
```
simonlingoogle, line 627: Need to stop `mVicariousRouterSolicitTimer` in `RoutingManager::Stop`?

simonlingoogle (approved):
LGTM :+1:

src/core/border_router/routing_manager.cpp
```diff
@@ -1116,59 +1187,51 @@ bool RoutingManager::AddDiscoveredPrefix(const Ip6::Prefix &aPrefix,
...
-        mDiscoveredPrefixInvalidTimer.FireAtIfEarlier(newPrefix.mExpireTime);
-
-        ++mDiscoveredPrefixNum;
-        added = true;
+        discoveredPrefix = &mDiscoveredPrefixes[mDiscoveredPrefixNum++];
+        addedNewPrefix   = true;
     }
-    else
+
+    if (discoveredPrefix == nullptr)
```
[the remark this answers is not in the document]
wgtdkp: The `discoveredPrefix` may be updated in above if-statement.

src/core/border_router/routing_manager.cpp
```diff
@@ -1042,69 +1102,80 @@ bool RoutingManager::UpdateDiscoveredPrefixes(const RouterAdv::RouteInfoOption &
...
+        if (!aStaleTimeInitialized || aPrefixStaleTime >= aStaleTime)
+        {
+            aStaleTime = aPrefixStaleTime;
+        }
+    }
+    else
+    {
+        if (!aStaleTimeInitialized || aPrefixStaleTime <= aStaleTime)
+        {
+            aStaleTime = aPrefixStaleTime;
```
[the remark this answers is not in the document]
wgtdkp: We add an external route for each OMR prefix in RIO and this consumes significant network data. This is for removing any OMR prefixes sooner to free network data space.

src/core/border_router/routing_manager.cpp
```diff
@@ -1116,59 +1187,51 @@ bool RoutingManager::AddDiscoveredPrefix(const Ip6::Prefix &aPrefix,
...
-            mDiscoveredPrefixInvalidTimer.FireAtIfEarlier(prefix.mExpireTime);
-
-            otLogInfoBr("discovered prefix %s refreshed lifetime: %u seconds", aPrefix.ToString().AsCString(),
-                        aLifetime);
-            ExitNow();
+            discoveredPrefix = &prefix;
+            break;
         }
+
+        UpdateStaleTime(staleTime, staleTimeInitialized, prefix.mIsOnLinkPrefix, prefix.GetStaleTime());
```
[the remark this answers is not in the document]
wgtdkp: We choose the furthest on-link prefix stale time to start RS when the last on-link prefix becomes stale.
We choose the nearest OMR prefix stale time to start RS when any OMR prefix becomes stale.

src/core/border_router/routing_manager.cpp
```diff
@@ -1191,20 +1258,22 @@ bool RoutingManager::NetworkDataContainsOmrPrefix(const Ip6::Prefix &aPrefix) co
...
         mRouterAdvMessage.SetToDefault();
     }
     else
     {
-        mRouterAdvMessage = aRouterAdvMessage;
-        // TODO: add a timer for invalidating the learned RA parameters
-        // for cases that the other RA daemon crashed or is force killed.
+        mRouterAdvMessage = *aRouterAdvMessage;
+        mDiscoveredPrefixStaleTimer.FireAtIfEarlier(mTimeRouterAdvMessageLastUpdate +
+                                                    Time::SecToMsec(kRtrAdvStaleTime));
```
wgtdkp, line 1276: @jwhui We also update the STALE timer for the learned RA parameters so that we can always forget it even when the Wi-Fi daemon crashes without sending RA messages with zero router lifetime.

src/core/border_router/routing_manager.cpp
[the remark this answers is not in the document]
jwhui: But we may break out of the loop above. Is it OK if we do not evaluate every on-link and OMR prefix?

src/core/border_router/routing_manager.cpp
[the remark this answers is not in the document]
jwhui: Above if-statement will always set `discoveredPrefix` to a non-null value, right?

src/core/border_router/routing_manager.cpp
```diff
@@ -1042,69 +1104,82 @@ bool RoutingManager::UpdateDiscoveredPrefixes(const RouterAdv::RouteInfoOption &
...
+        mDiscoveredPrefixes[i].mValidLifetime = 0;
     }
 
     InvalidateDiscoveredPrefixes();
 
     OT_ASSERT(mDiscoveredPrefixNum == 0);
 }
 
+static void UpdateStaleTime(TimeMilli &aStaleTime,
+                            bool       aStaleTimeInitialized,
```
jwhui, line 1161: Is this supposed to be pass-by-reference?

wgtdkp: You are right, thanks!

src/core/border_router/routing_manager.cpp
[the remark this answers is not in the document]
wgtdkp: The implementation has problem when there are both OMR prefixes and on-link prefixes, I updated to evaluate on-link prefix stale time and OMR prefix stale time separately.

At the same time, I think it may be beneficial to remove any on-link prefix as soon as we detected that the on-link prefix is no longer being advertised, for the same reason we remove the OMR prefixes. Thoughts?

src/core/border_router/routing_manager.cpp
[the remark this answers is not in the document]
wgtdkp: > Above if-statement will always set discoveredPrefix to a non-null value, right?

I think it is only when there is enough buffer for the new prefix and `AddExternalRoute` returns no error.

I updated to use nested if-statements, please review.

src/core/border_router/routing_manager.cpp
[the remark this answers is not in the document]
wgtdkp: no, we need to evaluate all prefixes, updated.

src/core/border_router/routing_manager.cpp
```diff
@@ -1042,69 +1104,74 @@ bool RoutingManager::UpdateDiscoveredPrefixes(const RouterAdv::RouteInfoOption &
...
     }
 
     InvalidateDiscoveredPrefixes();
 
     OT_ASSERT(mDiscoveredPrefixNum == 0);
 }
 
+static void UpdateStaleTime(TimeMilli &aOnLinkPrefixstaleTime,
+                            TimeMilli &aOmrPrefixstaleTime,
+                            bool       aIsOnLink,
```
jwhui, line 1162:
```suggestion
static void UpdateStaleTime(TimeMilli &aOnLinkPrefixStaleTime,
                            TimeMilli &aOmrPrefixStaleTime,
                            bool       aIsOnLink,
```

src/core/border_router/routing_manager.cpp
```diff
@@ -1116,59 +1183,51 @@ bool RoutingManager::AddDiscoveredPrefix(const Ip6::Prefix &aPrefix,
     OT_ASSERT(aIsOnLinkPrefix ? IsValidOmrPrefix(aPrefix) : IsValidOnLinkPrefix(aPrefix));
     OT_ASSERT(aLifetime > 0);
 
-    bool added = false;
+    bool            addedNewPrefix        = false;
+    ExternalPrefix *discoveredPrefix      = nullptr;
+    TimeMilli       now                   = TimerMilli::GetNow();
+    TimeMilli       onLinkPrefixstaleTime = now;
+    TimeMilli       omrPrefixstaleTime    = now.GetDistantFuture();
```
jwhui, line 1190:
```suggestion
    TimeMilli       onLinkPrefixStaleTime = now;
    TimeMilli       omrPrefixStaleTime    = now.GetDistantFuture();
```

src/core/border_router/routing_manager.hpp
```diff
@@ -154,36 +154,52 @@ class RoutingManager : public InstanceLocator
...
-        kMaxRaDelayTime              = 500, // The maximum delay of sending RA after receiving RS. In milliseconds.
-        kRtrSolicitationInterval     = 4,   // The interval between Router Solicitations. In seconds.
-        kMaxRtrSolicitationDelay     = 1,   // The maximum delay for initial solicitation. In seconds.
-        kMaxRoutingPolicyDelay       = 1,   // The maximum delay for routing policy evaluation. In seconds.
+        kMaxInitRtrAdvInterval       = 16,   // Maximum Initial Router Advertisement Interval. In seconds.
+        kMaxRaDelayTime              = 500,  // The maximum delay of sending RA after receiving RS. In milliseconds.
+        kRtrSolicitationInterval     = 4,    // The interval between Router Solicitations. In seconds.
+        kMaxRtrSolicitationDelay     = 1,    // The maximum delay for initial solicitation. In seconds.
+        kMaxRoutingPolicyDelay       = 1,    // The maximum delay for routing policy evaluation. In seconds.
+        kRtrAdvStaleTime             = 1500, // The STALE time of Router Advertisement message and prefixes. In seconds.
```
jwhui, line 174: How did you pick 1500?

src/core/border_router/routing_manager.cpp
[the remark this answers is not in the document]
jwhui: I think it depends on what we mean by "stale". Remember that just because we are no longer receiving RAs with the on-link prefix, does not mean that some host on the link still considers the on-link prefix as valid. We should be careful about removing an external route too quickly.

src/core/border_router/routing_manager.hpp
wgtdkp: - It is shorter than the `kDefaultOnLinkPrefixLifetime` and `kDefaultOmrPrefixLifetime` which is 1800 seconds.
- It is larger than twice of the `kMaxRtrAdvInterval` so that a BR get two chances of receiving non-solicited RAs before considering a prefix is `STALE`. I think the `STALE` time should be at least larger than `kMaxRtrAdvInterval`.

wgtdkp: added comments and s static assertion for it.

jwhui: RFC 4861 allows MaxRtrAdvInterval to be up to 1800 seconds.

wgtdkp: Thanks, what about below configuration?
```C++
kDefaultOnLinkPrefixLifetime = 2400,
kMaxRtrAdvInterval = 600,
kRtrAdvStaleTime = 2000,
```

jwhui: Having both `kDefaultOnLinkPrefixLifetime` and `kRtrAdvStaleTime` set to 1800 may work. RAs are sent with a random delay that has max value of MaxRtrAdvInterval. Thoughts?

wgtdkp: also works for me, updated.

jwhui (approved):

Issue #6472, opened 2021-04-16
Lack of TREL posix implementation test coverage with removal of `ot-ncp` posix app


abtink:
TREL poisx platform implementation (in general multi-radio build) were covered by running full set of `toranj` tests under the previous `toranj-rcp-trel` in GitHub actions.

The `ot-ncp` posix app was recently removed from https://github.com/openthread/openthread/pull/6428 which basically also removed these tests from the GH action.

Note that we have a TREL platform implementation under simulation which is intended to verify behavior of TREL/multi-radio in OT core (which is covered by the tests). But more importantly we also have TREL platform implementation under `posix` which is intended for use on products/projects. I think it is very important we cover the posix TREL platform implementation (not only make sure it builds ok, but run it through a set of tests).

I think there was a misunderstanding (on my part) that the PR above would add a CLI based test to cover posix platform under TREL to replace the existing test that were being removed, but I don't think we have this.

We need to either bring back the `ot-ncp` and the related `toranj` test that cover TREL posix platform, or add new test under CLI.

abtink:
Submitted https://github.com/openthread/openthread/pull/6518 which is adding some build checks under posix with TREL enabled in the CI.

abtink:
We can close this as TREL is covered by tests in `ot-br-posix`

PR #6515, opened 2021-04-27
[posix-trel] support TREL URL

simonlingoogle:
This commit supports TREL URL (e.g. `trel://eth0`) for Posix.

Depends-On: openthread/ot-br-posix#803

PR #6518, opened 2021-04-27
[build] update build scripts and add multi-radio (TREL) CI build checks

abtink:
This commit updates `toranj/build.sh` script for posix based builds to
generate CLI executable. It also updates the posix platform to provide
log functionality when `LOG_OUTPUT` is set to `LOG_OUTPUT_APP`.

This commit also adds CI build checks for multi-radio and TREL build
configurations under posix platform.

-----

This is related to https://github.com/openthread/openthread/issues/6472.


src/posix/platform/logging.cpp
```diff
@@ -35,7 +35,8 @@
 
 #include <openthread/platform/logging.h>
 
-#if OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED
+#if (OPENTHREAD_CONFIG_LOG_OUTPUT == OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED) || \
```
simonlingoogle, line 38: Why define `otPlatLog` in platform when `LOG_OUTPUT == APP` ?

simonlingoogle (approved):

src/posix/platform/logging.cpp
abtink: Under `posix` build the posix platform itself is acting as the app. If we don't add it the under `LOG_OUTPUT_APP` we would get an error (missing definition of `otPlatLog`). I guess such a config may have not ben covered by any CI build before.

abtink:
@jwhui just a quick note on this. This have now two commits (which I guess we can squash when merging). Thanks.

src/posix/platform/logging.cpp
simonlingoogle: > Under `posix` build the posix platform itself is acting as the app

I think this is not totally true. Should we keep the boundary of posix platform and posix app more clear?

> If we don't add it the under LOG_OUTPUT_APP we would get an error (missing definition of otPlatLog). 

We should fix in a more suitable way?

Thoughts? @bukepo

src/posix/platform/trel_udp6.cpp
```diff
@@ -109,20 +109,20 @@ static const char *BufferToString(const uint8_t *aBuffer, uint16_t aLength)
     char *   cur = &string[0];
     char *   end = &string[sizeof(string) - 1];
 
-    cur += snprintf(cur, end - cur, "[(len:%d) ", aLength);
+    cur += snprintf(cur, (uint16_t)(end - cur), "[(len:%d) ", aLength);
```
bukepo, line 112:
```suggestion
    cur += snprintf(cur, static_cast<uint16_t>(end - cur), "[(len:%d) ", aLength);
```

src/posix/platform/logging.cpp
bukepo: The platform should only define otPlatLog when `LOG_OUTPUT == PLATFORM_DEFINED`. @abtink may I know how to reproduce the compile error you encountered?

src/posix/platform/trel_udp6.cpp
abtink: This is really a c implementation not cpp file and therefore all other casts follow C style.
(we need to change the fule name to make it clear later in new PR).

src/posix/platform/logging.cpp
abtink: This PR is intended to add back to our github action CI some TREL posix builds (somewhat addressing the issue https://github.com/openthread/openthread/issues/6472 caused due to removal of `ot-ncp` and previous TREL posix build checks and tests). 

The use of `LOG_OUTPUT_APP` with CLI under posix seems to have been broken (I guess we did not detect it because  this configuration was probably not being built/checked before). This config used to work correctly with NCP posix build (but as you know the `ot-ncp` was removed, so now we want to use CLI based builds).

I want this PR to be merged to somewhat address the lack any of build check for posix TREL build config. It does not make sense to deal with issues related to posix (I think they should be addressed in other PRs). The goal of this PR is to unblock other PRs adding changes in TREL like https://github.com/openthread/openthread/pull/6515 (so that at minimum we have a CI build check for TREL posix).

So in the new pushed commit in this PR I changed the `toranj` config to use `PLATFORM_DEFINED` for the posix builds, so we don't cover or use this config.

Note that use of `LOG_OUTPUT_APP` config is still broken on posix build. @bukepo , to reproduce simply set the `LOG_OUTPUT` to `LOG_OUTPUT_APP` and build posix CLI executable. 
- The reason for this seems to be that in the CLI, the `otPlatLog` is defined `examples/apps/cli/main.c` but in `posix` build does not seem to include this file (`posix` has its own `main.c` and the only place under posix where `otPlatLog` is defined is in `src/posix/platform/logging.cpp`).
- I leave it to you guys to address this (feel free to later change back the `toranj` config to use `OUTPUT_APP` under posix to add build coverage for this config).

bukepo: Thanks for the explanation. The `LOG_OUTPUT_APP` was intentionally not implemented in posix ot-cli because we can get the same effect with `-v` option(see https://github.com/openthread/openthread/blob/main/src/posix/main.c#L367). The simulation platform is for testing purpose, so we should support as many configurations as possible on that platform. And in my opinion, we may just support what we need on POSIX platform. Thoughts? @abtink

bukepo (approved):
LGTM:+1:

jwhui (approved):

src/posix/platform/logging.cpp
simonlingoogle: Can we define a weak `otPlatLog` in posix platform when `LOG_OUTPUT=LOG_OUTPUT_APP` ?

src/posix/platform/trel_udp6.cpp
simonlingoogle: > This is really a c implementation not cpp file and therefore all other casts follow C style.

Is all Posix platform sources CPP? Is it beneficial to keep the code pure C?

abtink: It is matter of style in each file. As I said, all other casts in this file are c format. l I don't have any preference here, but I don't think we want to change something like this from this PR.

PR #6522, opened 2021-04-27
[multi-radio] add mechanism to timeout last received frag tag

abtink:
This commit adds a new mechanism in `Neighbor` to timeout the last
received fragment tag being tracked per neighbor. The fragment tag is
used when `MULTI_RADIO` is enabled to suppress duplicate received
frames (over different radios).

### DIFF
--- a/src/core/config/radio_link.h
+++ b/src/core/config/radio_link.h
@@ -72,4 +72,14 @@
 #define OPENTHREAD_CONFIG_MULTI_RADIO 0
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_MULTI_RADIO_FRAG_TAG_TIMEOUT
+ *
+ * Specifies the fragment tag timeout interval in milliseconds.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_MULTI_RADIO_FRAG_TAG_TIMEOUT
+#define OPENTHREAD_CONFIG_MULTI_RADIO_FRAG_TAG_TIMEOUT (20 * 1000)
+#endif
+
 #endif // CONFIG_RADIO_LINK_H_
--- a/src/core/thread/topology.cpp
+++ b/src/core/thread/topology.cpp
@@ -154,6 +154,19 @@ bool Neighbor::MatchesFilter(StateFilter aFilter) const
     return matches;
 }
 
+#if OPENTHREAD_CONFIG_MULTI_RADIO
+void Neighbor::SetLastRxFragmentTag(uint16_t aTag)
+{
+    mLastRxFragmentTag     = (aTag == 0) ? 0xffff : aTag;
+    mLastRxFragmentTagTime = TimerMilli::GetNow();
+}
+
+bool Neighbor::IsLastRxFragmentTagSet(void) const
+{
+    return (mLastRxFragmentTag != 0) && (TimerMilli::GetNow() <= mLastRxFragmentTagTime + kLastRxFragmentTagTimeout);
+}
+#endif
+
 void Neighbor::GenerateChallenge(void)
 {
     IgnoreError(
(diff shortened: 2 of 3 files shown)


src/core/config/radio_link.h
```diff
@@ -72,4 +72,19 @@
...
+ * Specifies the fragment tag timeout interval in milliseconds.
+ *
+ * The default value is calculated as follows: The fragment tag is 16-bit wide, so 2^15 values are considered ahead of
+ * current tag value (using mod 2^16 arithmetic). It is incremented for every tx message that includes the fragment
+ * header. Considering a pessimistic average of 8 messages tx per second (containing fragment header), gives us
+ * `2^15/8 = 4096` seconds (or ~68 minutes).
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_MULTI_RADIO_FRAG_TAG_TIMEOUT
+#define OPENTHREAD_CONFIG_MULTI_RADIO_FRAG_TAG_TIMEOUT (4096 * 1000)
```
simonlingoogle, line 87: I thought I have commented about this, but it seems not.

Should we define `OPENTHREAD_CONFIG_MULTI_RADIO_FRAG_TAG_TIMEOUT` based on the max possible delay of frame transmission on the TREL link?

abtink: @simonlingoogle can you please help explain more? What do you mean by the max possible delay?

The frag tag is incremented for all tx out of device (independent of the radio link(s) being used and/or which neighbor it is for). I explained the assumptions for selecting this default value in the comment above the definition.

simonlingoogle: I mean if frame transmission delay on the TREL link never exceed 10 seconds, then as long as the timeout since last fragment exceed 10 seconds, it's safe to say that the fragment **CAN NOT** be a duplicated transmission of 802.15.4 frame even when the fragment tag is smaller or equal to the current fragment tag. 

Thoughts?

abtink: I see. You can thinking about in a different way.

- If we get a frag tag value of `20` and suppose we assume max of 10 second delay on radio link.
- We then remember and use the `20` tag value for the next 10 seconds to duplicate subpression.
  -  For example, if we receive get a msg with tag `20` or `19` or lower we reject it as as duplicate within the next 10s
  - After the 10 seconds we don't do any check anymore (accept msg with any farg tag value) since should not receive messages that late anyways.

The question for me is that is there any benefit to limit the window to this shorter max delay rather than using a longer wait time.

I think ideally we want to use the tag value for as long we can ( we would not have used any timeout/aging model if we could - similar to how for MLE and link frame counter we don't use any aging model, but since the tag is `uint16_t` it is possible for it to roll-over on the sender side and we need to handle this situation).

- Suppose 5 minutes from setting the frag to `20` we get a msg with tag `19`.
- If we use 10 as timeout, then we would accept such a message.
- But we know it would be very unlikely that sender  have rolled-over its tag within 5 mintes and therefore it must be a duplicate.

The other aspect that I am not sure if we can easily define a max delay in a good way. The tag duplicate detection logic is applicable to all types of nodes and radios which can include sleepy devices with varying data poll rates.

simonlingoogle: > But we know it would be very unlikely that sender  have rolled-over its tag within 5 mintes and therefore it must be a duplicate.

I was wondering if the BRs with TREL links are communicating between each other at high throughput, e.g. transmitting OTA images, then would it be possible that datagram tags are consumed so quickly that rolling over may happen in 5 minutes?

abtink:
I rebased this PR. I think would be good to merge this in.
I changed the default tag timeout to be shorter, 20 seconds.

simonlingoogle (approved):
LGTM :+1: 

Thanks

jwhui (approved):

PR #6524, opened 2021-04-28
[cli] simplify arg checking/parsing with new `Arg` class

abtink:
This commit adds a new `Arg` class in `parse_cmdline` which represents
a single argument (wrapper over a C string). The `Arg` class is then
used in CLI modules. This helps simplify the CLI code mainly using
simpler syntax, e.g., using an overload of `==` operator to compare an
argument with a C string, or calling `ParseAs{Type}()` methods
directly on an `Arg` instance.

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -86,19 +86,8 @@
 #include "common/new.hpp"
 #include "common/string.hpp"
 #include "mac/channel_mask.hpp"
-#include "utils/parse_cmdline.hpp"
 
 using ot::Encoding::BigEndian::HostSwap16;
-
-using ot::Utils::CmdLineParser::ParseAsBool;
-using ot::Utils::CmdLineParser::ParseAsHexString;
-using ot::Utils::CmdLineParser::ParseAsInt8;
-using ot::Utils::CmdLineParser::ParseAsIp6Address;
-using ot::Utils::CmdLineParser::ParseAsIp6Prefix;
-using ot::Utils::CmdLineParser::ParseAsUint16;
-using ot::Utils::CmdLineParser::ParseAsUint32;
-using ot::Utils::CmdLineParser::ParseAsUint64;
-using ot::Utils::CmdLineParser::ParseAsUint8;
 
 namespace ot {
 namespace Cli {
@@ -181,15 +170,15 @@ int Interpreter::OutputIp6Address(const otIp6Address &aAddress)
         HostSwap16(aAddress.mFields.m16[5]), HostSwap16(aAddress.mFields.m16[6]), HostSwap16(aAddress.mFields.m16[7]));
 }
 
-otError Interpreter::ParseEnableOrDisable(const char *aString, bool &aEnable)
+otError Interpreter::ParseEnableOrDisable(const Arg &aArg, bool &aEnable)
 {
     otError error = OT_ERROR_NONE;
 
-    if (strcmp(aString, "enable") == 0)
+    if (aArg == "enable")
     {
         aEnable = true;
     }
-    else if (strcmp(aString, "disable") == 0)
+    else if (aArg == "disable")
     {
         aEnable = false;
     }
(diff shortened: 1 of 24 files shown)


src/core/utils/parse_cmdline.hpp
```diff
@@ -284,6 +270,297 @@ otError ParseAsHexString(const char *       aString,
...
+     */
+    uint16_t GetLength(void) const { return static_cast<uint16_t>(strlen(mString)); }
+
+    /**
+     * This method gets the argument as C string.
+     *
+     * @returns A pointer to the argument as a C string.
+     *
+     */
+    const char *GetString(void) const { return mString; }
```
simonlingoogle, line 294:
```suggestion
    const char *GetCString(void) const { return mString; }
```

src/cli/cli.cpp
```diff
@@ -4891,32 +4884,43 @@ void Interpreter::ProcessLine(char *aBuf)
...
 
-    // Check user defined commands if built-in command has not been found
+    SuccessOrExit(ProcessUserCommands(argsLength, args));
+    OutputResult(OT_ERROR_INVALID_COMMAND);
+
+exit:
+    return;
+}
+
+otError Interpreter::ProcessUserCommands(uint8_t aArgsLength, Arg *aArgs)
```
simonlingoogle, line 4906:
```suggestion
otError Interpreter::ProcessUserCommands(uint8_t aArgsLength, Arg aArgs[])
```
nit: prefer a little bit.

bukepo:
Can we provide a `string_view`-like class and let the `Arg` inherit it so that the `string_view` part can be reused in core?

bukepo (approved):
LGTM:+1:

src/cli/cli.cpp
abtink: I also agree. Makes it clear it's an array. Changed it in new pushed commit.

abtink:
> Can we provide a string_view-like class and let the Arg inherit it so that the string_view part can be reused in core?

We can consider it for later, but may need to think of how to use it in `parse_cmd` and CLI. The `parse_cmd` module is sort of unique in the sense that it is a utility module that is used by core (in `factory_diags.hpp`) but also by CLI modules. We want the CLI modules not to use any of the core definitions. For now the `parse_cmd` is somewhat standalone module (the header file does not use any of the core headers).

simonlingoogle (approved):
LGTM :+1:

src/core/utils/parse_cmdline.hpp
```diff
@@ -284,6 +270,297 @@ otError ParseAsHexString(const char *       aString,
...
+
+    /**
+     * This method parses the argument as an IPv6 prefix.
+     *
+     * The string is parsed as `{IPv6Address}/{PrefixLength}`.
+     *
+     * @param[out] aPrefix   A reference to an `otIp6Prefix` to output the parsed IPv6 prefix.
+     *
+     * @retval kErrorNone         The argument was parsed successfully.
+     * @retval kErrorInvalidArgs  The argument does not contain valid IPv6 prefix
```
jwhui, line 464:
```suggestion
     * @retval kErrorInvalidArgs  The argument does not contain valid IPv6 prefix.
```

jwhui (approved):

src/cli/cli.cpp
```diff
@@ -4891,32 +4884,43 @@ void Interpreter::ProcessLine(char *aBuf)
...
 
     if (command != nullptr)
     {
         OutputResult((this->*command->mHandler)(argsLength - 1, &args[1]));
         ExitNow();
     }
 
-    // Check user defined commands if built-in command has not been found
+    SuccessOrExit(ProcessUserCommands(argsLength, args));
+    OutputResult(OT_ERROR_INVALID_COMMAND);
```
lmnotran, line 4900: @abtink In the case where an application has UserCommands registered, it looks like this line would be executed even after a matching command has been successfully executed. This looks like a bug to me as we were able to reproduce it in one of our sample apps. Could you double check?

abtink: Thanks @lmnotran. Yes. That looks incorrect. Will submit a fix for this soon.

abtink: Just submitted https://github.com/openthread/openthread/pull/6562.

PR #6532, opened 2021-04-29
[ip] make mesh-local addresses non-preferred

bukepo:
This commit marks anycast addresses and mesh-local addresses as IPv6 non-preferred addresses.

The background is host IPv6 stack(such as lwIP, Linux kernel and so on) cannot tell these addresses to OMR addresses from the scope id. By marking these addresses non-preferred, the host IPv6 stack will not select these addresses as the source address of application layer traffic unless explicitly assigned.


include/openthread/ip6.h
```diff
@@ -427,7 +427,7 @@ typedef struct otIp6AddressInfo
     const otIp6Address *mAddress;       ///< A pointer to the IPv6 address.
     uint8_t             mPrefixLength;  ///< The prefix length of mAddress if it is a unicast address.
     uint8_t             mScope : 4;     ///< The scope of this address.
-    bool                mIsAnycast : 1; ///< Whether this is an anycast address.
+    bool                mPreferred : 1; ///< Whether this is a preferred address.
```
abtink, line 430: Missing `API_VERSION` increment.

abtink (approved):
LGTM. Thanks. 👍

src/core/thread/mle.cpp
```diff
@@ -141,10 +142,12 @@ Mle::Mle(Instance &aInstance)
 
     mMeshLocal64.InitAsThreadOriginRealmLocalScope();
     mMeshLocal64.GetAddress().GetIid().GenerateRandom();
+    mMeshLocal64.mPreferred = false;
```
simonlingoogle, line 145: Is it conceptually incorect to set ML64 as `not preferred`? ML64 is preferred within the Thread network, but may not be preferred on host interface. I think we need to differentiate the two cases.

src/core/thread/mle.cpp
```diff
@@ -128,6 +128,7 @@ Mle::Mle(Instance &aInstance)
     mLinkLocal64.GetAddress().SetToLinkLocalAddress(Get<Mac::Mac>().GetExtAddress());
 
     mLeaderAloc.InitAsThreadOriginRealmLocalScope();
+    mLeaderAloc.mPreferred = false;
```
simonlingoogle, line 131: We should set mPreferred = false in `InitAsThreadOriginRealmLocalScope` ?

src/core/net/netif.cpp
```diff
@@ -45,20 +45,19 @@ namespace Ip6 {
...
-        mIsAnycast    = aInstance.Get<Mle::MleRouter>().IsAnycastLocator(aAddress.GetAddress());
+        mPreferred    = aAddress.mPreferred;
     }
 
     explicit AddressInfo(const NetifMulticastAddress &aAddress)
     {
         mAddress      = &aAddress.GetAddress();
         mPrefixLength = kMulticastPrefixLength;
         mScope        = aAddress.GetAddress().GetScope();
-        mIsAnycast    = false;
```
simonlingoogle, line 61: Should keep `mPreferred = false` since the structure is not zero-initialized?

src/posix/platform/netif.cpp
```diff
@@ -334,7 +334,7 @@ static void UpdateUnicastLinux(const otIp6AddressInfo &aAddressInfo, bool aIsAdd
 
     req.nh.nlmsg_len = NLMSG_ALIGN(req.nh.nlmsg_len) + rta->rta_len;
 
-    if (aAddressInfo.mIsAnycast)
+    if (!aAddressInfo.mPreferred)
```
wgtdkp, line 337: I cannot directly comment that line, but we need to set `ifa_cacheinfo.ifa_prefered` to zero to mark it not preferred. (we are setting only `ifa_valid`).

wgtdkp: And, update for `BSD` platforms as well?

bukepo: > I cannot directly comment that line, but we need to set `ifa_cacheinfo.ifa_prefered` to zero to mark it not preferred. (we are setting only `ifa_valid`).

It's  memset to `0`.

wgtdkp: > > I cannot directly comment that line, but we need to set `ifa_cacheinfo.ifa_prefered` to zero to mark it not preferred. (we are setting only `ifa_valid`).
> 
> It's memset to `0`.

1. It is not readable/maintainable to rely on default zero-initialization.
2. We need to set `ifa_cacheinfo.ifa_prefered` to non-zero value (e.g. `UINT32_MAX`) when the address is preferred.
3. Besides, I think `ifa_cacheinfo. ifa_valid ` should always be set to `UINT32_MAX` no matter this address is preferred or not (as it is always valid address).

bukepo: > > > I cannot directly comment that line, but we need to set `ifa_cacheinfo.ifa_prefered` to zero to mark it not preferred. (we are setting only `ifa_valid`).
> > 
> > 
> > It's memset to `0`.
> 
> 1. It is not readable/maintainable to rely on default zero-initialization.

It's not default zero-initialization, but explicitly memset to 0. If this pattern is not readable/maintainable, I guess we need to change a massive amount of code in OpenThread. I don't think it necessary to double set a field. Thoughts?

> 2. We need to set `ifa_cacheinfo.ifa_prefered` to non-zero value (e.g. `UINT32_MAX`) when the address is preferred.
> 3. Besides, I think `ifa_cacheinfo. ifa_valid ` should always be set to `UINT32_MAX` no matter this address is preferred or not (as it is always valid address).

Please note this attribute is only attached when this address is not preferred.

src/core/thread/mle.cpp
```diff
@@ -124,7 +124,7 @@ Mle::Mle(Instance &aInstance)
     mParentCandidate.Clear();
     ResetCounters();
 
-    mLinkLocal64.InitAsThreadOrigin();
+    mLinkLocal64.InitAsThreadOrigin(true);
```
wgtdkp, line 127:
```suggestion
    mLinkLocal64.InitAsThreadOrigin(/* aPreferred */ true);
```

wgtdkp (approved):
LGTM 👍

simonlingoogle (approved):
LGTM :+1: 

with a small suggestion

abtink (approved):
LGTM. 👍

jwhui (approved):

PR #6533, opened 2021-04-29
[dnssd] move reference counting to core

gjc13:
This commit adds reference counting for the pending dns queries for the
same name. The subscribe/unsubscribe callback will be only be called on
the first/last query after this change.

### DIFF
--- a/src/core/net/dnssd_server.cpp
+++ b/src/core/net/dnssd_server.cpp
@@ -758,6 +758,7 @@ Error Server::ResolveByQueryCallbacks(Header &                aResponseHeader,
     QueryTransaction *query = nullptr;
     DnsQueryType      queryType;
     char              name[Name::kMaxNameSize];
+    bool              firstQuery;
 
     Error error = kErrorNone;
 
@@ -766,10 +767,14 @@ Error Server::ResolveByQueryCallbacks(Header &                aResponseHeader,
     queryType = GetQueryType(aResponseHeader, aResponseMessage, name);
     VerifyOrExit(queryType != kDnsQueryNone, error = kErrorNotImplemented);
 
-    query = NewQuery(aResponseHeader, aResponseMessage, aCompressInfo, aMessageInfo);
+    firstQuery = !HasSameQuery(name, queryType);
+    query      = NewQuery(aResponseHeader, aResponseMessage, aCompressInfo, aMessageInfo);
     VerifyOrExit(query != nullptr, error = kErrorNoBufs);
 
-    mQuerySubscribe(mQueryCallbackContext, name);
+    if (firstQuery)
+    {
+        mQuerySubscribe(mQueryCallbackContext, name);
+    }
 
 exit:
     return error;
@@ -1103,8 +1108,32 @@ void Server::FinalizeQuery(QueryTransaction &aQuery, Header::Response aResponseC
     OT_ASSERT(sdType != kDnsQueryNone);
     OT_UNUSED_VARIABLE(sdType);
 
-    mQueryUnsubscribe(mQueryCallbackContext, name);
     aQuery.Finalize(aResponseCode, mSocket);
+    // We are the last subscribing this name
+    if (!HasSameQuery(name, sdType))
+    {
+        mQueryUnsubscribe(mQueryCallbackContext, name);
+    }
+}
+
+bool Server::HasSameQuery(const char *aName, DnsQueryType aQueryType)
+{
+    bool found = false;
+
+    for (QueryTransaction &query : mQueryTransactions)
+    {
+        if (!query.IsValid())
+        {
+            continue;
+        }
+
+        if (query.HasQuestionName(aName, aQueryType))
+        {
+            found = true;
+            break;
+        }
+    }
+    return found;
 }
 
 void Server::QueryTransaction::Init(const Header &          aResponseHeader,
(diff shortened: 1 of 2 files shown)


src/core/net/dnssd_server.cpp
```diff
@@ -836,6 +841,38 @@ bool Server::CanAnswerQuery(const Server::QueryTransaction &aQuery, const char *
...
+    uint8_t     skipDot = 0;
+    const char *split   = aName;
+
+    switch (aQueryType)
+    {
+    case kDnsQueryBrowse:
+        skipDot = 2;
+        break;
+    case kDnsQueryResolve:
+        skipDot = 3;
```
simonlingoogle, line 855: Instance name itself may contain dots. So, the number of dots to skip is not fixed for instance resolving.

src/core/net/dnssd_server.cpp
```diff
@@ -1103,8 +1140,39 @@ void Server::FinalizeQuery(QueryTransaction &aQuery, Header::Response aResponseC
...
+        char         queryName[Name::kMaxNameSize];
+        DnsQueryType queryType;
+        const char * querySplit;
+
+        if (!query.IsValid())
+        {
+            continue;
+        }
+        queryType  = GetQueryType(query.GetResponseHeader(), query.GetResponseMessage(), queryName);
+        querySplit = GetDomainSplit(queryName, queryType);
```
simonlingoogle, line 1167: Why do we compare without the domains? 

The external query callbacks do not assume that it's going to be resolved by a Discovery Proxy. So, DNS names with different domains are different in DNS-SD server's viewpoint.

src/core/net/dnssd_server.cpp
```diff
@@ -1103,8 +1140,39 @@ void Server::FinalizeQuery(QueryTransaction &aQuery, Header::Response aResponseC
...
+}
+
+bool Server::HasSameQuery(const char *name, DnsQueryType aQueryType)
+{
+    const char *split = GetDomainSplit(name, aQueryType);
+    bool        found = false;
+
+    for (QueryTransaction &query : mQueryTransactions)
+    {
+        char         queryName[Name::kMaxNameSize];
```
simonlingoogle, line 1158:
```suggestion
```
I think it's possible to implement query transaction searching without using another buffer. 
`Dns::Name::CompareName` allows comparing C strings with DNS names in messages without reading to buffers. 

BTW, We have been very careful on allocating buffers of `char [Name::kMaxNameSize]` in the callchains of DNS-SD server to avoid increasing stack usage.

src/core/net/dnssd_server.hpp
```diff
@@ -258,6 +258,7 @@ class Server : public InstanceLocator, private NonCopyable
         TimeMilli               GetStartTime(void) const { return mStartTime; }
         NameCompressInfo &      GetNameCompressInfo(void) { return mCompressInfo; };
         void                    Finalize(Header::Response aResponseMessage, Ip6::Udp::Socket &aSocket);
+        bool                    HasQuestionName(const char *aName);
```
simonlingoogle, line 261: const?

src/core/net/dnssd_server.hpp
```diff
@@ -356,6 +357,7 @@ class Server : public InstanceLocator, private NonCopyable
     static bool       CanAnswerQuery(const Server::QueryTransaction &aQuery, const char *aHostFullName);
     void AnswerQuery(QueryTransaction &aQuery, const char *aHostFullName, const otDnssdHostInfo &aHostInfo);
     void FinalizeQuery(QueryTransaction &aQuery, Header::Response aResponseCode);
+    bool HasSameQuery(const char *aName);
```
simonlingoogle, line 360: const?

simonlingoogle (approved):
LGTM :+1:

src/core/net/dnssd_server.cpp
```diff
@@ -1103,8 +1108,32 @@ void Server::FinalizeQuery(QueryTransaction &aQuery, Header::Response aResponseC
...
-    mQueryUnsubscribe(mQueryCallbackContext, name);
     aQuery.Finalize(aResponseCode, mSocket);
+    // We are the last subscribing this name
+    if (!HasSameQuery(name))
+    {
+        mQueryUnsubscribe(mQueryCallbackContext, name);
+    }
+}
+
+bool Server::HasSameQuery(const char *aName)
```
simonlingoogle, line 1119: We should follow the name searching rules implemented in `GetQueryType`?

Currently `GetQueryType` searches for PTR/TXT/SRV questions first, and then AAAA/A questions.
Otherwise the logic is not strictly correct.

Thoughts?

gjc13: Updated code to check the query type as well.

src/core/net/dnssd_server.hpp
```diff
@@ -356,6 +357,7 @@ class Server : public InstanceLocator, private NonCopyable
     static bool       CanAnswerQuery(const Server::QueryTransaction &aQuery, const char *aHostFullName);
     void AnswerQuery(QueryTransaction &aQuery, const char *aHostFullName, const otDnssdHostInfo &aHostInfo);
     void FinalizeQuery(QueryTransaction &aQuery, Header::Response aResponseCode);
+    bool HasSameQuery(const char *aName, DnsQueryType aQueryType);
```
simonlingoogle, line 360: const?

src/core/net/dnssd_server.hpp
```diff
@@ -258,6 +266,7 @@ class Server : public InstanceLocator, private NonCopyable
         TimeMilli               GetStartTime(void) const { return mStartTime; }
         NameCompressInfo &      GetNameCompressInfo(void) { return mCompressInfo; };
         void                    Finalize(Header::Response aResponseMessage, Ip6::Udp::Socket &aSocket);
+        bool                    HasQuestionName(const char *aName, DnsQueryType aQueryType) const;
```
simonlingoogle, line 269: const ?

simonlingoogle (approved):
I am fine with this PR.

However, I think openthread/ot-br-posix#821 should not be merged because reference counting is still necessary in Discovery Proxy. 

Thoughts?

gjc13:
Replaced by https://github.com/openthread/openthread/pull/6611

PR #6534, opened 2021-04-29
[tmf] avoid binding TMF port for user UDP

wgtdkp:


src/core/net/udp6.cpp
```diff
@@ -229,10 +229,11 @@ Error Udp::Bind(SocketHandle &aSocket, const SockAddr &aSockAddr, otNetifIdentif
         do
         {
             aSocket.mSockName.mPort = GetEphemeralPort();
+        } while (aSocket.mSockName.mPort == Tmf::kUdpPort
```
simonlingoogle, line 232: nit: Should we move this logic to `GetEphemeralPort` so that it never returns `Tmf::kUdpPort` ?

wgtdkp: SGTM!

bukepo (approved):
LGTM:+1:

simonlingoogle (approved):
lgtm 👍

jwhui (approved):

PR #6539, opened 2021-04-30
[mle] ignore "Child Update Request" from a child not yet in valid state

abtink:
This commit updates `MleRouter::HandleChildUpdateRequest()` related to
how a "Child Update Request" message from a child is processed on a
parent. If the child is present in the parent's child table but it is
not yet in valid state, the parent ignores the received "Child Update
Request". This applies to a child being restored due to parent's reset
(child is in `kStateRestored`) or a child which is in the middle of
attach process (e.g., child is in `kStateParentRequest`). If a
matching child entry cannot be found in the parent's child table, the
behavior remains as before (i.e., parent sends a response with an
error Status TLV to notify the child to re-attach).

This change ensures that after a parent reset to restore a child, the
parent node is the one initiating and sending a "Child Update Request"
to the child and then receives a "Child Update Response" from it. Such
an exchange between the parent and the child will then include a
random challenge and response (authenticating the child and providing
replay protection). Also in this exchange the "Child Update Response"
from the child will include the the child's current MAC and MLE frame
counters, therefore ensuring that the parent correctly learns and
tracks the frame counters being used by the child.

### DIFF
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -2796,7 +2796,7 @@ void Mle::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessageIn
 #if OPENTHREAD_FTD
         if (IsRouterOrLeader())
         {
-            Get<MleRouter>().HandleChildUpdateRequest(aMessage, aMessageInfo, keySequence);
+            Get<MleRouter>().HandleChildUpdateRequest(aMessage, aMessageInfo);
         }
         else
 #endif
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -2314,9 +2314,7 @@ exit:
     LogProcessError(kTypeChildIdRequest, error);
 }
 
-void MleRouter::HandleChildUpdateRequest(const Message &         aMessage,
-                                         const Ip6::MessageInfo &aMessageInfo,
-                                         uint32_t                aKeySequence)
+void MleRouter::HandleChildUpdateRequest(const Message &aMessage, const Ip6::MessageInfo &aMessageInfo)
 {
     static const uint8_t kMaxResponseTlvs = 10;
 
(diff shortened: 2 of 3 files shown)


abtink:
@jwhui and I noticed this while discussing and looking into some other topic related to how to reset the frag tag under `MULTI_RADIO` config on child recovery. 

I tried looking into the history of the code and found a related PR #1853 (from 2017). Adding the info here so @jwhui  and others can also help check and review this (to make sure the new behavior does also  handle the situation from PR #1853).

- The PR #1853 addressed a potential issue with recovery of a non-sleepy child after parent reset (detailed scenario described in #1853 description). 
- In short, the issue was we could get to a situation where the child would assume it is connected to the parent (since it continued to receive "Child Update Response" to its "Child Update Requests") but parent would not change the child's state to "valid". 
- PR #1853 fixed this by setting the child to `kStateValid` upon receiving a "Child Update Request" from it (when trying to restore the child). 
- I think the behavior change from the current PR should also address the original situation from #1853.
   - By having the parent no longer respond to the child's "Child Update Request" messages (when the child entry in parent's child table is not marked as valid), the child then detects this and tries to re-attach (which is what we want to happen).

Irving-cl:
> By having the parent no longer respond to the child's "Child Update Request" messages (when the child entry in parent's child table is not marked as valid), the child then detects this and tries to re-attach (which is what we want to happen).

Does this mean, after a parent reset, if a child happened to send "Child Update Request" before the parent sent "Child Update Request" and set its internal state (about that child) to 'valid', then we ignore the "Child Update Request" and let the child to re-attach?

abtink:
> Does this mean, after a parent reset, if a child happened to send "Child Update Request" before the parent sent "Child Update Request" and set its internal state (about that child) to 'valid', then we ignore the "Child Update Request" and let the child to re-attach?

Yes. If for example both child and parent are reset and both are trying to do quick re-attach (sending "Child Update Request" to each other), we do want them to do a full re-attach (to ensure the MLE/MAC counters get updated properly between the two).

If parent is reset, and trying to recover the child using quick re-attach (sending "Child Update Request" to the child), but at the same time, child happens to be sending its own "Child Update Request" to the parent, we ignore it until the child responds to the parent's request (which would help validate the child to the parent (due to presence of challenge/response TLV in the exchange and also sync the MLE/MAC counters of the child on the parent). This ensures we don't accept a child without validating it and synching its counter. The child in this case typically retries its "Child Update Request" (for some number of attempts before detaching and doing full attach) and can get a response in a retry attempt (after it itself responded to the parent's request).

jwhui (approved):

PR #6541, opened 2021-04-30
[meshcop] support native commissioner

bukepo:
This commit adds native commissioner support with the following changes:

* Accept unsecure traffic when Thread is not enabled.
* Accept traffic to border agent service when native commissioner is
  allowed.

Depends-On: openthread/ot-commissioner#201

### DIFF
--- a/src/posix/platform/netif.cpp
+++ b/src/posix/platform/netif.cpp
@@ -576,8 +576,14 @@ static void processTransmit(otInstance *aInstance)
     rval = read(sTunFd, packet, sizeof(packet));
     VerifyOrExit(rval > 0, error = OT_ERROR_FAILED);
 
-    message = otIp6NewMessage(aInstance, nullptr);
-    VerifyOrExit(message != nullptr, error = OT_ERROR_NO_BUFS);
+    {
+        otMessageSettings settings;
+
+        settings.mLinkSecurityEnabled = (otThreadGetDeviceRole(aInstance) != OT_DEVICE_ROLE_DISABLED);
+        settings.mPriority            = OT_MESSAGE_PRIORITY_LOW;
+        message                       = otIp6NewMessage(aInstance, &settings);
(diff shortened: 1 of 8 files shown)


bukepo:
The check external-commissioner is rename to ot-commissioner because it also covers as native commissioner now. The checks already completed. @jwhui

wgtdkp (approved):
LGTM 👍

src/posix/platform/netif.cpp
```diff
@@ -565,18 +565,19 @@ static void processReceive(otMessage *aMessage, void *aContext)
...
-    ssize_t    rval;
-    char       packet[kMaxIp6Size];
-    otError    error  = OT_ERROR_NONE;
-    size_t     offset = 0;
+    otMessage *       message = nullptr;
+    ssize_t           rval;
+    char              packet[kMaxIp6Size];
+    otError           error  = OT_ERROR_NONE;
+    size_t            offset = 0;
+    otMessageSettings settings{otThreadGetDeviceRole(aInstance) != OT_DEVICE_ROLE_DISABLED, OT_MESSAGE_PRIORITY_LOW};
```
abtink, line 573: Just a suggestion here which I think help with readability of the code (I am fine with it as is).

Instead of initializing the `otMessageSetting` struct when we declare the variable, how about if we delare the variable and then init its fields later when we use it:
```
    otMessageSettings settings;
```
And before `message = otIp6NewMessage(aInstance, &settings)` statement:
```
    setting.mLinkSecurityEnabled = (otThreadGetDeviceRole(aInstance) != OT_DEVICE_ROLE_DISABLED);
    setting.mPriority            = OT_MESSAGE_PRIORITY_LOW;
```

abtink (approved):
Looks great. Thanks. 👍 
I have one suggestion belwo (I am ok with the code as is as well).

abtink (approved):

jwhui (approved):

PR #6543, opened 2021-04-30
[posix] use Url instead of RadioUrl in headers

bukepo:


src/posix/platform/hdlc_interface.cpp
```diff
@@ -140,7 +140,7 @@ void HdlcInterface::OnRcpReset(void)
     mHdlcDecoder.Reset();
 }
 
-otError HdlcInterface::Init(const RadioUrl &aRadioUrl)
+otError HdlcInterface::Init(const Url::Url &aRadioUrl)
```
zhanglongxia, line 143: The class `RadioUrl`  inherits from `ot::Url::Url` and add a variable `char mUrl[kRadioUrlMaxSize];`. Can we delete the class `RadioURl` and move the buffer `char mUrl[kRadioUrlMaxSize];` to the class `ot::Url::Url`?

bukepo: This intentionally keeps flexibility of the URL buffer.

zhanglongxia: Can we use a template to allow the user to specify the length of the buffer like https://github.com/openthread/openthread/blob/main/src/core/common/string.hpp#L120 ?

simonlingoogle: Maybe we can reduce the buffer size to 128?. 512B seems too much for most cases, and takes significant stack size.

simonlingoogle (approved):
LGTM 👍

src/posix/platform/hdlc_interface.cpp
zhanglongxia: The length of the SPI Radio Url is about 200 bytes.

bukepo: > > This intentionally keeps flexibility of the URL buffer.
> 
> Since url is specified by arguments, the memory is always accessible. Maybe we don't need to copy the url to buffers in the first place. We only need to keep the pointers to components of the url. And this also makes copying Url legal.

The parser modifies the URL content.

bukepo: > Can we use a template to allow the user to specify the length of the buffer like https://github.com/openthread/openthread/blob/main/src/core/common/string.hpp#L120 ?

Do we have multiple radio url types with different size limits like the string template? Are you suggesting adding an OT_CONFIG to allow a different size limit?

simonlingoogle: > > > This intentionally keeps flexibility of the URL buffer.
> 
> > 
> 
> > Since url is specified by arguments, the memory is always accessible. Maybe we don't need to copy the url to buffers in the first place. We only need to keep the pointers to components of the url. And this also makes copying Url legal.
> 
> 
> 
> The parser modifies the URL content.

Yes, but it's safe to do it in place. And after that we can use copy Url around without any problem.

bukepo: I don't modified buffer would get correctly parsed after resetting(the current implementation).

zhanglongxia: @bukepo You said you want to keep the flexibility of the URL buffer.  I mean we can use a template to allow the user to set the length of the URL buffer to keep the flexibility. Or we can add an OT_CONFIG to configure the length of the URL buffer as you mentioned.

zhanglongxia (commented):
LGTM

zhanglongxia (approved):

src/posix/platform/hdlc_interface.cpp
simonlingoogle: > @bukepo You said you want to keep the flexibility of the URL buffer.  I mean we can use a template to allow the user to set the length of the URL buffer to keep the flexibility. Or we can add an OT_CONFIG to configure the length of the URL buffer as you mentioned.

We can implement it when we really need it? I would suggest against having too much flexibility when we don't need it yet.

jwhui (approved):

PR #6545, opened 2021-04-30
[ip6] improve frame filtering

MarekPorwisz:
If OpenThread is part of an OS with it's own IP stack filtering only
control messages is insufficien.
Example could be that if application is using OpenThread API to listen
for CoAP messages received messages are passed also to the IP callback.
The message then is delivered to the OS IP stack and generates error
response: "ICMPv6 Destination Unreachable (Port unreachable)" as the
port is not open on the OS side. OS IP stack is unaware of the
OpenThread existence so it can not be filtered on that level and the
shim layer does not receive port, or protocol information from OT so it
can not filter out a frame without parsing it first.
The only reasonable solution seems to be extending filtering
capabilities to filter all "OpenThread initiated" comunication.
Added extended filtering option to filter all of the udp packets destiend
to the port that is handled by OpenThread.

### DIFF
--- a/include/openthread/ip6.h
+++ b/include/openthread/ip6.h
@@ -171,6 +171,17 @@ enum
 };
 
 /**
+ * This enumeration defines possible rx packets filtering modes
+ *
+ */
+typedef enum otIp6ReceiveFilterMode
+{
+    OT_IP6_RX_FILTER_MODE_NO_FILTER                     = 0, ///< Filter is disabled.
+    OT_IP6_RX_FILTER_MODE_FILTER_THREAD_CONTROL_TRAFFIC = 1, ///< Filter all Thread control traffic
+    OT_IP6_RX_FILTER_MODE_FILTER_ALL_THREAD_USED_PORTS  = 2, ///< Filter all ports being used by OpenThread.
+} otIp6ReceiveFilterMode;
+
+/**
  * This structure represents an IPv6 network interface unicast address.
  *
  */
@@ -402,16 +413,16 @@ typedef void (*otIp6ReceiveCallback)(otMessage *aMessage, void *aContext);
 /**
  * This function registers a callback to provide received IPv6 datagrams.
  *
- * By default, this callback does not pass Thread control traffic.  See otIp6SetReceiveFilterEnabled() to
- * change the Thread control traffic filter setting.
+ * By default, this callback does not pass Thread control traffic.  See otIp6SetReceiveFilterMode() to
+ * change the Thread filter setting.
  *
  * @param[in]  aInstance         A pointer to an OpenThread instance.
  * @param[in]  aCallback         A pointer to a function that is called when an IPv6 datagram is received or
  *                               NULL to disable the callback.
  * @param[in]  aCallbackContext  A pointer to application-specific context.
  *
- * @sa otIp6IsReceiveFilterEnabled
- * @sa otIp6SetReceiveFilterEnabled
+ * @sa otIp6GetReceiveFilterMode
+ * @sa otIp6SetReceiveFilterMode
  *
  */
 void otIp6SetReceiveCallback(otInstance *aInstance, otIp6ReceiveCallback aCallback, void *aCallbackContext);
@@ -452,7 +463,7 @@ typedef void (*otIp6AddressCallback)(const otIp6AddressInfo *aAddressInfo, bool 
 void otIp6SetAddressCallback(otInstance *aInstance, otIp6AddressCallback aCallback, void *aCallbackContext);
 
 /**
- * This function indicates whether or not Thread control traffic is filtered out when delivering IPv6 datagrams
+ * This function indicates how inbound traffic is filtered out when delivering IPv6 datagrams
  * via the callback specified in otIp6SetReceiveCallback().
  *
  * @param[in]  aInstance A pointer to an OpenThread instance.
@@ -460,23 +471,23 @@ void otIp6SetAddressCallback(otInstance *aInstance, otIp6AddressCallback aCallba
  * @returns  TRUE if Thread control traffic is filtered out, FALSE otherwise.
  *
  * @sa otIp6SetReceiveCallback
- * @sa otIp6SetReceiveFilterEnabled
+ * @sa otIp6SetReceiveFilterMode
  *
  */
-bool otIp6IsReceiveFilterEnabled(otInstance *aInstance);
+otIp6ReceiveFilterMode otIp6GetReceiveFilterMode(otInstance *aInstance);
 
 /**
- * This function sets whether or not Thread control traffic is filtered out when delivering IPv6 datagrams
+ * This function sets sets how inbound traffic should be filtered out when delivering IPv6 datagrams
  * via the callback specified in otIp6SetReceiveCallback().
  *
- * @param[in]  aInstance A pointer to an OpenThread instance.
- * @param[in]  aEnabled  TRUE if Thread control traffic is filtered out, FALSE otherwise.
+ * @param[in]  aInstance   A pointer to an OpenThread instance.
+ * @param[in]  aFilterType TRUE if Thread control traffic is filtered out, FALSE otherwise.
  *
  * @sa otIp6SetReceiveCallback
- * @sa otIsReceiveIp6FilterEnabled
+ * @sa otIp6GetReceiveFilterMode
  *
  */
-void otIp6SetReceiveFilterEnabled(otInstance *aInstance, bool aEnabled);
+void otIp6SetReceiveFilterMode(otInstance *aInstance, otIp6ReceiveFilterMode aFilterMode);
 
 /**
  * This function sends an IPv6 datagram via the Thread interface.
--- a/include/openthread/udp.h
+++ b/include/openthread/udp.h
@@ -290,6 +290,18 @@ void otUdpForwardReceive(otInstance *        aInstance,
                          uint16_t            aSockPort);
 
 /**
+ * Determines if the given UDP port is exclusively opened by OpenThread API.
+ *
+ * @param[in]  aInstance            A pointer to an OpenThread instance.
+ * @param[in]  port                 UDP port number to verify.
+ *
+ * @retval true    The port is being used exclusively by OpenThread
+ * @retval false   The port is not used by any of the OpenThread API or is shared (e.g. is Backbone socket ).
+ *
+ */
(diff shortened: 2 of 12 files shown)


src/core/net/udp6.cpp
```diff
@@ -527,6 +527,18 @@ void Udp::HandlePayload(Message &aMessage, MessageInfo &aMessageInfo)
...
+{
+    for (const SocketHandle * socket = mSockets.GetHead(); socket != nullptr; socket=socket->GetNext())
+    {
+        if(socket->GetSockName().GetPort() == aPort)
+        {
+            return true;
+        }
+    }
+    return false;
+}
```
abtink, line 540: Here we are going through all the sockets in the list. The list here includes sockets added by OT core itself (e.g., internal Thread use), but I think it also contains sockets that may be added by user of OT APIs, from `otUdpOpen()`. So I think we would be then filtering all port used by any UDP socket (internal by core of external through `otUdpOpen()`).

abtink (commented):
Thanks @MarekPorwisz for submitting this. 
In general, I think such a filtering mode would be useful.
I have couple of suggestions below. Thanks.

include/openthread/ip6.h
```diff
@@ -473,11 +474,43 @@ bool otIp6IsReceiveFilterEnabled(otInstance *aInstance);
...
+ *
+ * @param[in]  aInstance A pointer to an OpenThread instance.
+ * @param[in]  aEnabled  TRUE if Thread initiated traffic is filtered out, FALSE otherwise.
+ *
+ * @sa otIp6SetReceiveCallback
+ * @sa otIp6IsReceiveFullFilterEnabled
+ * @sa otIp6SetReceiveFilterEnabled
+ *
+ */
+void otIp6SetReceiveFullFilterEnabled(otInstance *aInstance, bool aEnabled);
```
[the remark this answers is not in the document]
MarekPorwisz: Thanks, I was considering this, however decided not to do that. This would make API incompatible with previous one, and every project using it would need to be modified a bit. If you think that's ok, I can change it.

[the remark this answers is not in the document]
jwhui: While I agree we should generally try to keep the existing API stable, in this particular case, I am fine to change the API. I don't think the `otIp6GetReceiveFilterMode()` and `otIp6SetReceiveFilterMode()` APIs are used in many places and it should be easy to resolve any conflicts that exist.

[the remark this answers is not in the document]
abtink: I agree. In general we want to keep exiting API but in this  case, it may be fine to allow the API change to keep the number of APIs smaller and easier to use (I also expect that this API is not being used in many places/projects).

src/core/net/udp6.cpp
MarekPorwisz: My intention is to filter any traffic initiated by OpenThread and all of its API. Unless I am missing some bigger picture here, when port is opened using otUdpOpen() the communication is handled using otUdpAddReceiver, otUdpSent, etc. and does not need to be passed to mReceiveIp6DatagramCallback.
In our case we noticed an issue with zephyr:
When OpenThread API is used to handle CoAP it opens an udp port. Any frame received is first handled by OpenThread, but after that it is passed to Zephyr and since the port is not opened there it responds with "ICMPv6 Destination Unreachable (Port unreachable)".

include/openthread/udp.h
```diff
@@ -289,6 +289,18 @@ void otUdpForwardReceive(otInstance *        aInstance,
                          const otIp6Address *aPeerAddr,
                          uint16_t            aSockPort);
 
+/**
+ * Determines if the given UDP port is used by OpenThread, or belongs entirely to the platform
```
simonlingoogle, line 293: What does it mean for a UDP port to be `used by OpenThread`. Define it clearly in the Doxygen?

src/core/net/ip6.cpp
```diff
@@ -1035,8 +1036,15 @@ Error Ip6::ProcessReceiveCallback(Message &          aMessage,
             Udp::Header udp;
 
             IgnoreError(aMessage.Read(aMessage.GetOffset(), udp));
-            VerifyOrExit(Get<Udp>().ShouldUsePlatformUdp(udp.GetDestinationPort()), error = kErrorNoRoute);
 
+            if(mIsReceiveIp6FullFilterEnabled)
+            {
+                VerifyOrExit(Get<Udp>().IsPortInUse(udp.GetDestinationPort()) == false, error = kErrorNoRoute);
```
simonlingoogle, line 1042: In the case of `PLAT_UDP + full filter`, the packet won't be passed to the Thread netif (e.g. wpan0) if the port is bound in OT. 
However, currently OT (in Udp::HandleMessage) only pass the message to `HandlePayload` if it's not using Platform UDP (e.g. MLE, TMF). 
We might need to update `Udp::HandleMessage` too accordingly.

I think we need to add test for that. For example, we need to run `test_plat_udp_accessibility.py` using `full filtering`.

src/core/net/udp6.cpp
```diff
@@ -527,6 +527,18 @@ void Udp::HandlePayload(Message &aMessage, MessageInfo &aMessageInfo)
     return;
 }
 
+bool Udp::IsPortInUse(uint16_t aPort) const
+{
+    for (const SocketHandle * socket = mSockets.GetHead(); socket != nullptr; socket=socket->GetNext())
+    {
+        if(socket->GetSockName().GetPort() == aPort)
+        {
+            return true;
```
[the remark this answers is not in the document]
simonlingoogle: Currently there are 4 kinds of UDP sockets:
1. Pure Thread stack (non-Platform_UDP) sockets (e.g. MLE, TMF)
2. Sockets bound to All Netifs (i.e. OT_NETIF_UNSPECIFIED)
3. Sockets bound Thread Netif (i.e. OT_NETIF_THREAD)
4. Sockets bound to Bacbkone Netif (i.e. OT_NETIF_BACKBONE)

I think we actually needs to pass the UDP messages to Host for case 4 because in case 4 the UDP port is not actually used by OT. OT bound the UDP socket to the Backbone netif, however there might be an user APP that bound to the same port but on Thread netif only. That APP should be able to receive the UDP messages using POSIX socket APIs.

src/core/net/udp6.cpp
```diff
@@ -527,6 +527,18 @@ void Udp::HandlePayload(Message &aMessage, MessageInfo &aMessageInfo)
     return;
 }
 
+bool Udp::IsPortInUse(uint16_t aPort) const
+{
+    for (const SocketHandle * socket = mSockets.GetHead(); socket != nullptr; socket=socket->GetNext())
```
simonlingoogle, line 532: Currently there are 4 kinds of UDP sockets:
1. Pure Thread stack (non-Platform_UDP) sockets (e.g. MLE, TMF)
2. Sockets bound to All Netifs (i.e. OT_NETIF_UNSPECIFIED)
3. Sockets bound Thread Netif (i.e. OT_NETIF_THREAD)
4. Sockets bound to Bacbkone Netif (i.e. OT_NETIF_BACKBONE)

I think we actually needs to pass the UDP messages to Host for case 4 because in case 4 the UDP port is not actually used by OT. OT bound the UDP socket to the Backbone netif, however there might be an user APP that bound to the same port but on Thread netif only. That APP should be able to receive the UDP messages using POSIX socket APIs.

MarekPorwisz: Any idea how to solve all the possible cases?
I was wondering about instead of automatic filtering all ports in use allow application to pass list of ports to filter out - if app is using OpenThread API for communication it usually is aware of it ;)

simonlingoogle: > Any idea how to solve all the possible cases?

As I have described, we need to skip case 4. BTW, we can use`Udp::IsBackboneSocket` to check if a socket is of case 4.

> I was wondering about instead of automatic filtering all ports in use allow application to pass list of ports to filter out - if app is using OpenThread API for communication it usually is aware of it ;)

I think automatic filtering should work.

MarekPorwisz:
Addressed some of the comments. Still need to look into some (Testing possibilities, Udp::HandleMessage, ...)

src/core/net/ip6.cpp
MarekPorwisz: Could you explain this some more please? How is HandleMessage related to my changes? Where is this test_plat_udp_accessibility.py test located?

src/core/net/ip6.cpp
```diff
@@ -1035,8 +1036,21 @@ Error Ip6::ProcessReceiveCallback(Message &          aMessage,
...
+                break;
+            }
+            case OT_IP6_RX_FILTER_MODE_FILTER_ALL_THREAD_USED_PORTS:
+            {
+                VerifyOrExit(Get<Udp>().IsPortInUse(udp.GetDestinationPort()) == false, error = kErrorNoRoute);
+                break;
+            }
+            default:
+                OT_ASSERT(false);
+            }
```
simonlingoogle, line 1053:
```suggestion
            VerifyOrExit(Get<Udp>().ShouldUsePlatformUdp(udp.GetDestinationPort()) && !Get<Udp>().IsPortInUse(udp.GetDestinationPort()), error = kErrorNoRoute);
```
I think it's unnecessary to define `mIp6ReceiveFilterType` after all. 
We should be able to handle UDP messages correctly without defining an extra `filter type`.
Thoughts? @MarekPorwisz @abtink @bukepo

MarekPorwisz: I had it originally that way (but without backbone sockets included in the IsPortInUse) and some tests were failing. I do not remember if all were backbone router related. I also wonder if someone has already based their design  on limited filtering and may need to filter only control traffic.

MarekPorwisz: created a separate draft PR with your proposal to see if it passes tests:
https://github.com/openthread/openthread/pull/6641

MarekPorwisz: The other PR fails on test_plat_udp_accessiblity.py: _test_srp_server. TBH I am not sure if this is related to the filters change, but that is possible.

simonlingoogle: Very likely. Failing on `test_plat_udp_accessiblity.py` constantly would mean that the filter change does not work with platform UDP yet (need fix).

MarekPorwisz:
I had to take care of some other tasks. I will come back to this issue some time later.

jwhui:
Superseded by #6711

PR #6548, opened 2021-04-30
[cli] fix args length checking in `dataset securitypolicy`

jwhui:


src/cli/cli_dataset.cpp
```diff
@@ -822,7 +822,7 @@ Error Dataset::ParseSecurityPolicy(otSecurityPolicy &aSecurityPolicy, uint8_t aA
     memset(&policy, 0, sizeof(policy));
     SuccessOrExit(error = aArgs[0].ParseAsUint16(policy.mRotationTime));
 
-    VerifyOrExit(aArgsLength >= 1);
+    VerifyOrExit(aArgsLength >= 2);
```
wgtdkp, line 825: It seems we should put the check to the beginning of this method?

wgtdkp (approved):
LGTM 👍

PR #6552, opened 2021-05-02
[posix] support interface name in ot-ctl

bukepo:
This commit adds the Thread interface name to unix socket of OpenThread daemon, so that `ot-ctl` can specify which daemon to connect to.

### DIFF
--- a/src/posix/client.cpp
+++ b/src/posix/client.cpp
@@ -43,6 +43,7 @@
 #define OPENTHREAD_USE_READLINE (HAVE_LIBEDIT || HAVE_LIBREADLINE)
 
 #include <assert.h>
+#include <getopt.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -62,6 +63,14 @@
 
 #include "platform-posix.h"
 
+namespace ot {
+namespace Client {
+
+struct Config
+{
+    const char *mNetifName;
+};
+
 enum
 {
     kLineBufferSize = OPENTHREAD_CONFIG_CLI_MAX_LINE_LENGTH,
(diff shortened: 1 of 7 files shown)


bukepo:
will update after #6557 is merged.

src/posix/platform/system.cpp
```diff
@@ -162,6 +162,8 @@ otInstance *otSysInit(otPlatformConfig *aPlatformConfig)
     platformNetifInit(instance, aPlatformConfig->mInterfaceName);
 #elif OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
     platformUdpInit(aPlatformConfig->mInterfaceName);
+#else
+    gNetifName[0] = '\0';
```
simonlingoogle, line 166: Why not initialize gNetifName with "" ?

src/posix/client.cpp
```diff
@@ -161,28 +175,82 @@ static bool ReconnectSession(void)
...
+        }
+    }
+
+    aArgCount -= optind;
+    aArgVector += optind;
+
+    return config;
+}
+
+extern "C" int main(int argc, char *argv[])
```
simonlingoogle, line 234: why do we need `extern "C"` ?

src/posix/client.cpp
```diff
@@ -111,7 +120,7 @@ static bool DoWrite(int aFile, const void *aBuffer, size_t aSize)
     return ret;
 }
 
-static int ConnectSession(void)
```
simonlingoogle, line 114: why do we remove `static`?

src/posix/platform/system.cpp
bukepo: Ensure it's initialized.

src/posix/client.cpp
bukepo: We use `static` to avoid global function conflicts. Now they are all inside of `ot::Client` namespace.

simonlingoogle (approved):
LGTM :+1: 
with questions

src/posix/client.cpp
bukepo: save `ot::Client::` for function calls.

jwhui (approved):

PR #6555, opened 2021-05-02
[mbedtls] update to 2.25

LuDuda:
This PR updates mbedTLS version to 2.25 (used in the TF-M 1.3 project) and contains two commits:

**[mbedtls] update to 2.25**

Which simply replaces the mbedtls files in the `third_party/mbedtls/repo`

**[mbedtls] adapt OpenThread to new mbedtls version**

Which:
- modifies Autotools, CMake and Android build configuration
- increases heap size as new version requires additional 64B
- replaces deprecated API (`mbedtls_ecdsa_sign_det`)
- removes workaround with defining `mbedtls_ecp_restart_ctx`

Apart from making CI green, I've done some manual commissioning tests on nRF52833/nRF52840 with various combinations.
The heap size has been increased by 48B and 64B experimentally (the issue with old values appeared only on CI with `-m32` build).

Probably it might be easier just to review the last commit. Finally, the squash would be needed when merging.


script/test
```diff
@@ -283,6 +283,7 @@ do_build_otbr_docker()
     (
         ./script/git-tool clone https://github.com/openthread/ot-br-posix.git --depth 1 "${otbrdir}"
         cd "${otbrdir}"
+        sed -i '/^#define MBEDTLS_SSL_EXPORT_KEYS/a #define MBEDTLS_SSL_KEEP_PEER_CERTIFICATE' third_party/openthread/mbedtls-config.h
```
LuDuda, line 286: This needs to be removed after https://github.com/openthread/ot-br-posix/pull/827 is merged.

romacdon (approved):
LGTM but I will defer final to @hemanth-silabs who is much closer to this than I.  The one question I have is this.  Is 2.25 the latest public version?  I went here: https://github.com/ARMmbed/mbedtls/releases and I see a 2.26 version.

LuDuda:
@romacdon the only reason is that this exact version is a part of the latest ARM's official TF-M project (released 3 weeks ago). As we try simultaneously introduce PSA Architecture this might help to reduce potential issues and maintain just one mbedTLS version. I haven't done any experiments to check if there might be any compatibility issues with the 2.26 version, to be honest - so this argument might not be strong enough :smile:.

hemanth-silabs (approved):
LGTM

LuDuda:
Rebased to master, and removed setting the `MBEDTLS_SSL_KEEP_PEER_CERTIFICATE` define by default to save RAM (though still keeping `otCoapSecureGetPeerCertificateBase64` API for backward-compatibility - which I can remove as well).

script/test
LuDuda: Reverted.

jwhui (approved):

PR #6556, opened 2021-05-03
[posix] escape CLI arguments

bukepo:


jwhui (approved):

src/posix/client.cpp
```diff
@@ -179,10 +184,15 @@ int main(int argc, char *argv[])
...
-            count += static_cast<size_t>(rval);
+            for (const char *c = argv[i]; *c; ++c)
+            {
+                if (IsSeparator(*c))
+                {
+                    buffer[count++] = '\\';
+                }
+                buffer[count++] = *c;
+            }
+            buffer[count++] = ' ';
```
simonlingoogle, line 195: I know it's merged. 

But does it mean that we need to use a larger buffer size? e.g. `char buffer[kLineBufferSize * 2];`
Because in the worse case where the whole argument is like "        " (kLineBufferSize spaces), then it's escaped to be a string of `kLineBufferSize * 2` size like `\ \ \ \ \ \ \ \ \ \ ....`

bukepo: No, `OPENTHREAD_CONFIG_CLI_MAX_LINE_LENGTH` already covers the escape characters. But we do need to avoid buffer overflow here.

simonlingoogle: Wrong comment.

PR #6557, opened 2021-05-03
[cli] move netif command to posix platform

bukepo:


src/cli/cli.cpp
```diff
@@ -2947,25 +2947,6 @@ otError Interpreter::ProcessNeighbor(uint8_t aArgsLength, Arg aArgs[])
 }
 #endif // OPENTHREAD_FTD
 
-#if OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE
-otError Interpreter::ProcessNetif(uint8_t aArgsLength, Arg aArgs[])
```
wgtdkp, line 2951: RTOS platforms may also want this command and they will need to duplicate the code for posix platforms. Thoughts?

bukepo: I don't think this is a common command. And platforms should be able to have their flexibility to writer their own.

wgtdkp (approved):

src/posix/platform/netif.cpp
```diff
@@ -150,6 +150,15 @@ extern int
...
 
+const char *otSysGetThreadNetifName(void)
+{
+    return gNetifName;
+}
+
+unsigned int otSysGetThreadNetifIndex(void)
+{
+    return gNetifIndex;
+}
```
jwhui, line 161: 1. nit: move to bottom to colocate with `otPlatGetNetif()`?
2. Is `otPlatGetNetif()` something we should keep?

bukepo: I forgot to delete the function. And I intentionally put it here so that it always available(even OPENTHREAD_CONFIG_PLATFORM_NETIF_ENABLE is not enabled).

include/openthread/platform/misc.h
```diff
@@ -193,26 +193,6 @@ otError otPlatSetMcuPowerState(otInstance *aInstance, otPlatMcuPowerState aState
  */
 otPlatMcuPowerState otPlatGetMcuPowerState(otInstance *aInstance);
 
-/**
- * @}
- *
- */
```
jwhui, line 199: Do not remove?

jwhui (approved):

PR #6559, opened 2021-05-04
[router-advertisement] fix implicit conversion loss

jwhui:
Thanks to @gabekassel for pointing this out.


wgtdkp (approved):
Thanks @gabekassel !

LGTM 👍

PR #6562, opened 2021-05-04
[cli] fix extra error output when handling user CLI commands

abtink:
It also adds a check in `cli-misc.exp` to verify error on an invalid
CLI command.

------

Thanks to @lmnotran for noticing and reporting this (in https://github.com/openthread/openthread/pull/6524#discussion_r626058987).


lmnotran (approved):
LGTM. Thanks for fixing this so quickly!

jwhui (approved):

PR #6564, opened 2021-05-05
[string] introduce StringWriter class

bukepo:
This commit moves methods of String out of the header file to save code
size.

Since most use cases of String writers ignores errors, this commit
changes the return of writers to the Writer object itself for chained
writes.

### DIFF
--- a/src/core/common/string.cpp
+++ b/src/core/common/string.cpp
@@ -32,6 +32,7 @@
  */
 
 #include "string.hpp"
+#include "debug.hpp"
 
 #include <string.h>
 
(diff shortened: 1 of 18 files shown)


abtink (approved):
This looks great. Thanks @bukepo. 👍

src/core/common/string.hpp
```diff
@@ -90,76 +90,64 @@ const char *StringFind(const char *aString, char aChar);
...
+private:
+    char mBuffer[kSize];
 };
 
-/**
- * This class defines a fixed-size string.
- *
- */
-template <uint16_t SIZE> class String : private StringBase
+class StringWriter
```
jwhui, line 118: nit: add doxygen?

src/core/common/string.cpp
```diff
@@ -72,30 +73,56 @@ bool StringEndsWith(const char *aString, char aChar)
...
 {
-    Error error = kErrorNone;
-    int   len;
+    mBuffer[0] = 0;
+}
 
-    len = vsnprintf(aBuffer + aLength, aSize - aLength, aFormat, aArgs);
+StringWriter &StringWriter::Clear(void)
+{
+    mBuffer[0] = 0;
```
jwhui, line 86: nit: to be more explicit
```suggestion
    mBuffer[0] = '\0';
```

src/core/common/string.hpp
```diff
@@ -90,76 +90,64 @@ const char *StringFind(const char *aString, char aChar);
...
+     * @retval  true    Buffer is sufficient.
+     * @retval  false   Buffer is insufficient.
      *
      */
-    void Clear(void)
-    {
-        mBuffer[0] = 0;
-        mLength    = 0;
-    }
+    bool IsSufficient(void) const { return mLength < mSize; }
```
jwhui, line 150: nit: a user of this method may find something like `IsTruncated()` more intuitive?

jwhui (approved):

PR #6578, opened 2021-05-06
[posix] avoid CLI buffer overflow

bukepo:


src/posix/client.cpp
```diff
@@ -182,22 +182,56 @@ int main(int argc, char *argv[])
...
+                        state = kStateArgument;
+                    }
+                    else if (IsSeparator(*c))
+                    {
+                        state           = kStateEscaping;
+                        buffer[count++] = '\\';
+                        continue;
+                    }
+
+                    buffer[count++] = *c++;
```
simonlingoogle, line 214:
```suggestion
                    if (state == kStateEscaping)
                    {
                        state = kStateArgument;
                        buffer[count++] = *c++;
                    }
                    else if (IsSeparator(*c))
                    {
                        state           = kStateEscaping;
                        buffer[count++] = '\\';
                    }
```

Is this more clear?

src/posix/client.cpp
```diff
@@ -182,22 +182,56 @@ int main(int argc, char *argv[])
...
+                for (const char *c = *arg; *c && count < sizeof(buffer);)
                 {
-                    buffer[count++] = '\\';
+                    if (state == kStateEscaping)
+                    {
+                        state = kStateArgument;
+                    }
+                    else if (IsSeparator(*c))
+                    {
+                        state           = kStateEscaping;
```
simonlingoogle, line 209: If the last character is `\`, the current behavior is incorrect because it will escape the next separator. 

I think we should treat this `\` as `\\`.

src/posix/platform/daemon.cpp
```diff
@@ -270,7 +270,7 @@ void platformDaemonProcess(const otSysMainloopContext *aContext)
     {
         uint8_t buffer[OPENTHREAD_CONFIG_CLI_MAX_LINE_LENGTH];
```
simonlingoogle, line 271: use `uint8_t buffer[OPENTHREAD_CONFIG_CLI_MAX_LINE_LENGTH+1];`?

src/posix/client.cpp
```diff
@@ -181,23 +181,61 @@ int main(int argc, char *argv[])
...
+                state = kStateEnd;
+                // fall through
             }
-            buffer[count++] = ' ';
+        default:
+            break;
         }
 
-        // replace the trailing space with newline
-        buffer[count - 1] = '\n';
```
simonlingoogle, line 199: Why we don't append `\n` any more in this change?

src/posix/client.cpp
```diff
@@ -181,23 +181,61 @@ int main(int argc, char *argv[])
...
-        VerifyOrExit(DoWrite(sSessionFd, buffer, count), ret = OT_EXIT_FAILURE);
+        // leave at least 1 byte for the null-terminator.
```
simonlingoogle, line 229: We do we append the traling '\0' and why do we need it?

src/posix/client.cpp
```diff
@@ -161,9 +161,9 @@ static bool ReconnectSession(void)
     return ok;
 }
 
-static bool IsSeparator(char aChar)
+static bool NeedEscape(char aChar)
```
simonlingoogle, line 164: We defined `IsEscapable` in openthread: `utils/parse_cmdline.cpp`. 
nit: Can we use `IsEscapable` directly here?

src/posix/client.cpp
```diff
@@ -181,23 +181,61 @@ int main(int argc, char *argv[])
...
+            } while (--argc && count < sizeof(buffer));
+
+        default:
+            break;
         }
 
-        // replace the trailing space with newline
-        buffer[count - 1] = '\n';
-        VerifyOrExit(DoWrite(sSessionFd, buffer, count), ret = OT_EXIT_FAILURE);
+        // leave at least 1 byte for the null-terminator.
```
Irving-cl, line 229: Where is the null-terminator set?

bukepo: not necessary at client side.

src/posix/client.cpp
```diff
@@ -181,23 +181,61 @@ int main(int argc, char *argv[])
...
+        } state = kStateArgument;
+
+        --argc;
+
+        switch (state)
+        {
+            do
             {
-                if (IsSeparator(*c))
+            case kStateEnd:
```
simonlingoogle, line 199: +1 suggest against `switch { do { case ...: ... } while(...); }` pattern.

@zhanglongxia

Irving-cl (approved):
LGTM 👍

src/posix/client.cpp
zhanglongxia: I think we may need to improve the code read readability here. This code pattern is not friendly to the code readers.

simonlingoogle: Readability matters.

Despite a small program, src/posix/client.cpp (i.e. ot-ctl) has a history of subtle bugs:
- #5087
- #5134
- #6378
- This PR

I think one reason is that we are trying too hard to make it as efficient as possible, meanwhile it actually does not matter at all for a command like `ot-ctl`.

zhanglongxia (approved):
LGTM

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6582, opened 2021-05-06
[string] add simple constructors to `String<kSize>`

abtink:
This commit adds two simple constructors to `String<kSize>` class,
one initializing the string as empty, and another one initializing it with
a given C string.


abtink:
Closing this due to impact on code-size

Issue #6583, opened 2021-05-06
[log] bad string/output in logs


abtink:
I have noticed recently some bad/weird formats in some of the log lines (e.g., the `Send Advertisement` below).
```
[INFO]-MLE-----: Receive Link Accept and Request (fe80:0:0:0:30e0:f941:5b85:5122,0x9000)
"?Oմ*)-MLE-----: Send Link Accept (fe80:0:0:0:30e0:f941:5b85:5122oZ~??.]{P
[INFO]-MAC-----: Sent IPv6 UDP msg, len:107, chksum:5906, to:32e0f9415b855122, sec:no, prio:net
[INFO]-MAC-----:     src:[fe80:0:0:0:f0ef:3bc:197e:3434]:19788
[INFO]-MAC-----:     dst:[fe80:0:0:0:30e0:f941:5b85:5122]:19788
[INFO]-MLE-----: Send Advertisement (ff02:0:0:0:0:0:0:1b?
                                                         j?QZP?????*)
[INFO]-MAC-----: Sent IPv6 UDP msg, len:91, chksum:b113, to:0xffff, sec:no, prio:net
[INFO]-MAC-----:     src:[fe80:0:0:0:f0ef:3bc:197e:3434]:19788
[INFO]-MAC-----:     dst:[ff02:0:0:0:0:0:0:1]:19788
[INFO]-MAC-----: Received IPv6 UDP msg, len:91, chksum:55c5, from:32e0f9415b855122, sec:no, prio:net, rss:-20.0
```

Adding this issue to investigate this further.

abtink:
I think I found the cause of this issue. It seems that it was introduced from PR https://github.com/openthread/openthread/pull/6564 (adding `StringWriter`).

Here is, for example, the method that logs MLE message (e.g., `Send Advertisement`).
```c++
void Mle::Log(MessageAction aAction, MessageType aType, const Ip6::Address &aAddress, uint16_t aRloc)
{
    enum : uint8_t
    {
        kRlocStringSize = 17,
    };

    String<kRlocStringSize> rlocString;

    if (aRloc != Mac::kShortAddrInvalid)
    {
        StringWriter(rlocString).Append(",0x%04x", aRloc);
    }

    otLogInfoMle("%s %s%s (%s%s)", MessageActionToString(aAction), MessageTypeToString(aType),
                 MessageTypeActionToSuffixString(aType, aAction), aAddress.ToString().AsCString(),
                 rlocString.AsCString());
}
```

[...]

PR #6584, opened 2021-05-06
[log] fix uninitialized string use

abtink:
This commit fixes an issue with some of the log related methods where
`String` object may be used without being initialized (this was added
from commit b39f5b6dd and PR #6564).

This commit ensures that `String` variable definitions are always
immediately  followed by a `StringWriter()` definition which then
initializes the string buffer (sets it as empty).

This commit also updates `test_string` unit test.

-----

This should help address https://github.com/openthread/openthread/issues/6583.

### DIFF
--- a/src/core/mac/mac_frame.cpp
+++ b/src/core/mac/mac_frame.cpp
@@ -1447,14 +1447,15 @@ Frame::InfoString Frame::ToInfoString(void) const
 
 BeaconPayload::InfoString BeaconPayload::ToInfoString(void) const
 {
-    NetworkName name;
-    InfoString  string;
+    NetworkName  name;
+    InfoString   string;
+    StringWriter writer(string);
 
     IgnoreError(name.Set(GetNetworkName()));
 
-    StringWriter(string).Append("name:%s, xpanid:%s, id:%d, ver:%d, joinable:%s, native:%s", name.GetAsCString(),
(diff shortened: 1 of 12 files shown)


bukepo (approved):
LGTM:+1: Thanks!

src/core/mac/mac_frame.cpp
```diff
@@ -1447,14 +1447,15 @@ Frame::InfoString Frame::ToInfoString(void) const
 
 BeaconPayload::InfoString BeaconPayload::ToInfoString(void) const
 {
-    NetworkName name;
-    InfoString  string;
+    NetworkName  name;
+    InfoString   string;
```
simonlingoogle, line 1451: Why don't we initialize `string` in its constructor and make `StringWriter` just append characters not caring the string is empty or not?
You mentioned that this was your preference. @abtink

abtink: Copying from the issue:

> I do like this as well sine it is a safer model, but did not go for it for now due to its code size impact. Adding a constructor to clear the String adds up to ~110 bytes (please see #6582). It somewhat defeats one of the main benefits of StringWriter (which is/was to reduce code size). For now, I think would be good to fix this quickly and then later we can consider different options.

simonlingoogle: Thanks for clarafication.

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6587, opened 2021-05-07
[netdata] adding `NetworkData::Publisher`

abtink:
This commit implements a new feature "Network Data Publisher" which
provides mechanisms to limit the number of similar DNS/SRP service
entries in the Thread Network Data by monitoring the Network Data and
managing if or when to add or remove entries. This feature is
enabled using `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_ENABLE` config, or
`NETDATA_PUBLISHER` in autoconf, or `OT_NETDATA_PUBLISHER` cmake
option.

When there is a request to publish an entry, the `Publisher` monitors
the Network Data and counts the number of similar entries. If there
are fewer entries than a desired target number, the entry is added
after a short random wait time.

If there are too many similar entries, `Publisher` starts the process
of removing its own entry (again after some random wait time). When
removing entries, certain entries are preferred over others (e.g., an
entry from a router over one from an end-device or if they are from
the same type of node, the one with smaller RLOC16). If `Publisher`
determines that its own entry is a preferred one, it adds an extra
wait time before removing its entry. This gives higher chance for a
non-preferred entry from another device to be removed before removing
a preferred entry which helps towards quicker convergence of the
process to the desired number of entries.

This commit also adds `test_netdata_publisher.py` to verify the
behavior of the `Publisher`.

---------

~This PR currently contains the commit from PR #6501. So please only
review the second commit.~ 

~The implementation in this PR supports "DNS/SRP service" entries, but
it is written in a way that we can extended it to support publishing
"off-mesh route" and/or "on-mesh prefix" entries as well. Will add this
in a set of follow-up PRs.~
Added support for on-mesh and external route in new pushed commit

~The OT SRP server currently directly adds/removes service entry
in/from network data. We should change it to use the `Publisher`
instead - we may add this in a new PR or may add it to this PR (after
testing).~ [included now in the PR]. 

~I have verified the behavior of the publisher doing manual testing, in
a later PR we should add CI tests for this.~
The latest pushed commit adds CI test `test_netdata_publisher.py` 
which  covers the behavior of the publisher 

----

Upon further discussion, we realized related the prefix entries
(on-mesh  and/or external route) on a device it may be desirable to
treat them as a set (i.e., have them all be added together and/or
removed together). The earlier implementation treats every prefix entry
individually. Such a change  requires some additional thoughts and
solutions (e.g., on how to decide in a combined  way (how to count
multiple existing entries) when/if to add/remove all prefix entries  as a set).
So in the latest push in this PR/branch removed prefix support and
only contains support for publishing DNS/SRP service entries.


src/cli/README_NETDATA.md
```diff
@@ -157,9 +159,36 @@ Print netdata help menu.
...
+
+This command requires `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_ENABLE`.
+
+The following formats are used to publish DNS/SRP Service entry (available when `OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE` is enabled):
+
+- `netdata publish dnssrp anycast <serial-num>` to publish "DNS/SRP Service Anycast Address" with a given serial number to be published in the Thread Network Data.
+- `netdata publish dnssrp unicast <address> <port>` to publish "DNS/SRP Service Unicast Address" with given address and port number info. The address/port info is included in Service TLV data.
+- `netdata publish dnssrp unicast <port>` to publish "DNS/SRP Service Unicast Address" with given port number and the device's mesh-local EID for the address. The address and port info is included in Server TLV data.
+
+A `netdata publish dnssrp` command will remove and replace any previous "DNS/SRP Service" entry that was being published (from earlier `netdata publish dnssrp` commands).
```
jwhui, line 182: Is this command implemented?

src/core/config/netdata_publisher.h
```diff
@@ -0,0 +1,110 @@
...
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_NETDATA_PUBLISHER_MAX_WAIT_TIME_TO_REMOVE
+#define OPENTHREAD_CONFIG_NETDATA_PUBLISHER_MAX_WAIT_TIME_TO_REMOVE 20000
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_NETDATA_PUBLISHER_EXTRA_WAIT_TIME_TO_REMOVE_PREFERRED
+ *
+ * Specifies the extra wait time (in milliseconds) used when removing a preferred entry.
```
jwhui, line 79: Document what is a "preferred" entry?

src/core/net/srp_server.cpp
```diff
@@ -445,7 +445,9 @@ void Server::Start(void)
     SuccessOrExit(error = mSocket.Open(HandleUdpReceive, this));
     SuccessOrExit(error = mSocket.Bind(kUdpPort, OT_NETIF_THREAD));
 
-    SuccessOrExit(error = PublishServerData());
+    // Publish "DNS/SRP Unicast Address Service" in Thread Network
+    // Data. using the device's mesh-local EID as the address
```
jwhui, line 449:
```suggestion
    // Data using the device's mesh-local EID as the address
```

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,601 @@
...
+    bool registerWithLeader = false;
+
+#if OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE
+    if (mDnsSrpServiceEntry.HandleNotifierEvents(aEvents))
+    {
+        registerWithLeader = true;
+    }
+#endif
+
+    if (registerWithLeader)
```
jwhui, line 74: Are we planning future enhancements that do not require `OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE`? If so, maybe add a comment here?

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,601 @@
...
+    bool registerWithLeader = false;
+
+#if OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE
+    if (mDnsSrpServiceEntry.HandleTimer())
+    {
+        registerWithLeader = true;
+    }
+#endif
+
+    if (registerWithLeader)
```
jwhui, line 96: Are we planning future enhancements that do not require `OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE`? If so, maybe add a comment here?

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,601 @@
...
+
+        Remove(/* aRegisterWithLeader */ true);
+    }
+
+    mType                   = kTypeUincastMeshLocalEid;
+    mInfo.mUnicast.mAddress = Get<Mle::Mle>().GetMeshLocal64();
+    mInfo.mUnicast.mPort    = aPort;
+    SetState(kToAdd);
+
+    Process();
```
jwhui, line 328: This code seems to be repeated at least three times - is it beneficial to break into a separate private method?

include/openthread/netdata_publisher.h
```diff
@@ -0,0 +1,127 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @addtogroup api-thread-general
+ *
+ * @{
+ *
+ * The Network Data Publisher provides mechanisms to limit the number of similar Service and/or Prefix (on-mesh prefix
```
jwhui, line 49: Indicate that the "Prefix (on-mesh prefix or external route)" portion is not yet implemented? Or exclude its mention?

src/core/config/netdata_publisher.h
jwhui, line 79: Be more clear on what "extra wait time" means? For example, document what the resulting [min, max] range is?

src/core/config/netdata_publisher.h
```diff
@@ -0,0 +1,110 @@
...
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_NETDATA_PUBLISHER_EXTRA_WAIT_TIME_TO_REMOVE_PREFERRED
+#define OPENTHREAD_CONFIG_NETDATA_PUBLISHER_EXTRA_WAIT_TIME_TO_REMOVE_PREFERRED 20000
```
jwhui, line 83: Should `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_EXTRA_WAIT_TIME_TO_REMOVE_PREFERRED` always be >= `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_MAX_WAIT_TIME_TO_REMOVE`?

src/core/thread/network_data_service.cpp
```diff
@@ -57,26 +57,30 @@ Error Manager::AddService(const void *aServiceData,
...
+                          bool        aRegisterWithLeader)
 {
     Error error;
 
     SuccessOrExit(error = Get<Local>().AddService(
                       kThreadEnterpriseNumber, reinterpret_cast<const uint8_t *>(aServiceData), aServiceDataLength,
                       aServerStable, reinterpret_cast<const uint8_t *>(aServerData), aServerDataLength));
 
+    VerifyOrExit(aRegisterWithLeader);
     Get<Notifier>().HandleServerDataUpdated();
```
jwhui, line 70: Other network data attributes such as on-mesh prefixes and routes do not call `HandleServerDataUpdate()` and require the user to call it separately. I'm wondering if we should remove the call here to be consistent with other network data methods.

src/cli/README_NETDATA.md
abtink: I think you mean the reference to `netdata publish dnssrp` in the above statement. 
I mean any flavor of this command (with any args). Will update the doc to make it clear.

src/core/thread/network_data_publisher.cpp
abtink: Yes. For on-mesh prefix and external route entries. 

I have the implementation to add support for those and was originally planning to add it as a separate PR but I think would be better to update this PR itself (include everything in one commit). Will add this in next push.

src/core/thread/network_data_publisher.cpp
abtink: Good suggestion. Thanks. I need to think if/how we can refactor this (since the code here is somewhat tied to each `Type` and its associated parameters).

include/openthread/netdata_publisher.h
abtink: Will squash support for on-mesh and external route in next push (everything in the same PR/commit).

src/core/config/netdata_publisher.h
abtink: Ideally yes (if we want to ensure the preferred entry is removed after giving chance to all non-preferred one) but I don't think we want to check and require this (can recommend it in description of this config).

src/core/thread/network_data_service.cpp
abtink: I agree. I was initially thinking that this may make it easier for `NetworkData::Service` by taking care of registering with leader from same method, but with `Publisher` taking care of this, it may be easier to go back to the model of caller needing to take care of `HandleServerDataUpdate()` call.

abtink: Updated in new pushed commit.

abtink:
The new pushed commit now includes support in `Publisher` for on-mesh prefix and external route entries in addition to DNS/SRP service entry. Updated the PR description to match the new commit message.

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1014 @@
...
+    const ServiceTlv *                  serviceTlv = nullptr;
+    const ServerTlv *                   serverSubTlv;
+
+    while ((serviceTlv = Get<Leader>().FindNextMatchingService(serviceTlv, Service::kThreadEnterpriseNumber,
+                                                               reinterpret_cast<const uint8_t *>(&serviceData),
+                                                               serviceData.GetLength())) != nullptr)
+    {
+        for (const NetworkDataTlv *start = serviceTlv->GetSubTlvs();
+             (serverSubTlv = NetworkData::FindTlv<ServerTlv>(start, serviceTlv->GetNext())) != nullptr;
+             start = serverSubTlv->GetNext())
```
abtink, line 665: Adding a note  to update the loops in this module to use `TlvIterator` with https://github.com/openthread/openthread/pull/6621 merged.

abtink:
The new pushed commit contains the following:

- It adds `test_netdata_publisher.py` which implements a CI test covering the behavior of `Publisher` in detail (for service, prefix and route entries).
- It also updates how the entries from different nodes are preferred (when trying to remove). We now prefer an entry from a router over similar entry from an end-device.

tests/scripts/thread-cert/node.py
```diff
@@ -1708,6 +1708,34 @@ def register_netdata(self):
         self.send_command('netdata register')
         self._expect_done()
 
+    def publish_netdata_dnssrp_anycast(self, seqnum):
```
simonlingoogle, line 1711:
```suggestion
    def publish_netdata_dnssrp_anycast(self, seqnum):
```
```suggestion
    def netdata_publish_dnssrp_anycast(self, seqnum):
```
nit: Use comm prefix `netdata` ?

src/core/thread/network_data_service.cpp
```diff
@@ -59,28 +59,15 @@ Error Manager::AddService(const void *aServiceData,
...
-
-    SuccessOrExit(error = Get<Local>().AddService(
-                      kThreadEnterpriseNumber, reinterpret_cast<const uint8_t *>(aServiceData), aServiceDataLength,
-                      aServerStable, reinterpret_cast<const uint8_t *>(aServerData), aServerDataLength));
-
-    Get<Notifier>().HandleServerDataUpdated();
-
-exit:
-    return error;
+    return Get<Local>().AddService(kThreadEnterpriseNumber, reinterpret_cast<const uint8_t *>(aServiceData),
```
simonlingoogle, line 62: I think this code change is to avoid multiple sending multiple `NETDATA.ntf` for each small change. 

But this adds extra complexity to callers of `AddService` because now callers have to call `HandleServerDataUpdated` subsequently. 

Can we keep this method unchaged, but change `HandleServerDataUpdated` to **NOT** call `SynchronizeServerData` directly, but post a tasklet to call `SynchronizeServerData` ? 
This can merge multiple `AddService` or `RemoveService` to one `NETDATA.ntf`, and also simplify the logic of callers of `AddService`.

Thoughts?

simonlingoogle (commented):
This is a great feature, thanks. 

Some initial comments.

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1016 @@
...
+    // Do not make any changes if device is not attached, and wait
+    // for role change event.
+    VerifyOrExit(Get<Mle::Mle>().IsAttached());
+
+    VerifyOrExit(GetState() != kNoEntry);
+
+    switch (mType)
+    {
+    case kTypeOnMeshPrefix:
+        CountOnMeshPrefixEntries(numEntries, numPreferredEntries);
```
simonlingoogle, line 904: What if this prefix entry is already added to the `Network Data`  but the state is not `kAdded` ? This may forbid this Prefix entry to change to its correct state `kAdded`. I think such case is possible when the device received a network data after reset.

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1016 @@
...
+        break;
+
+    case kTypeUnicast:
+    case kTypeUincastMeshLocalEid:
+        CountUnicastEntries(numEntries, numPreferredEntries);
+        desiredNumEntries = kDesiredNumUnicast;
+        break;
+    }
+
+    UpdateState(numEntries, numPreferredEntries, desiredNumEntries);
```
simonlingoogle, line 654: What if the service TLV is actually already contained in the `Network Data`, but the entry state is not `kAdded`. The entry may not be able to change to the correct state `kAdded`, and consequently unable to remove it when there are too many service TLVs.

src/core/thread/network_data_publisher.cpp
simonlingoogle: I realized that `Publisher` only need to be consistent with Local network data, so no more issue.

src/core/thread/network_data_publisher.cpp
simonlingoogle: I realized that `Publisher` only need to be consistent with Local network data, so no more issue.

src/core/api/netdata_publisher_api.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <openthread/netdata_publisher.h>
+
+#include "common/instance.hpp"
+#include "common/locator_getters.hpp"
+
+using namespace ot;
+
+#if OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE
+
+void otNetDataPublishDnsSrpServiceAnycast(otInstance *aInstance, uint8_t aSequenceNumber)
```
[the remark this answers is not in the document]
abtink: This API is useful and needed in a project where SRP server from OT is not being used and SRP server is implemented by next/app layer. 
Also useful to have this for testing.

src/core/backbone_router/bbr_local.cpp
```diff
@@ -187,6 +187,8 @@ Error Local::AddService(bool aForce)
     serverData.SetMlrTimeout(mMlrTimeout);
 
     SuccessOrExit(error = Get<NetworkData::Service::Manager>().Add<NetworkData::Service::BackboneRouter>(serverData));
+    Get<NetworkData::Notifier>().HandleServerDataUpdated();
```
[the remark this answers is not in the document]
abtink: `Service::Manager:Add<ServiceType>()` is changed, not longer registers with elader.

src/core/config/netdata_publisher.h
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * @def OPENTHREAD_CONFIG_NETDATA_PUBLISHER_EXTRA_WAIT_TIME_TO_REMOVE_PREFERRED
+ *
+ * Specifies the extra wait time (in milliseconds) used when removing a preferred entry.
+ *
+ * When removing entries, certain entries are preferred over others. How the preference is determined depends on the
+ * entry type (e.g., (for DNS/SRP service entries associated with smaller RLOC16 or smaller IPv6 address are preferred).
+ * If the `Publisher` determines that its own entry is a preferred one, it adds the above extra wait time in addition
+ * to the random wait selected based on `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_MAX_WAIT_TIME_TO_REMOVE` before  removing
+ * its entry. This gives higher chance for a non-preferred entry from another device to be removed before removing a
```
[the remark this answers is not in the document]
abtink: > Can a preferred one never remove itself when the number of entries with same or higher preference is not larger than the threshold? 

This would require/assume that all nodes are running `Publisher` (or similar logic to remove their entry is they see too many similar ones in net data) which may not be the case.

src/core/config/netdata_publisher.h
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * @def OPENTHREAD_CONFIG_NETDATA_PUBLISHER_DESIRED_NUM_EXTERNAL_ROUTE_ENTRIES
+ *
+ * Specifies the desired number of External Route entries in the Thread Network Data.
+ *
+ * Publisher attempts to limit the number of such entries to this value.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_NETDATA_PUBLISHER_DESIRED_NUM_EXTERNAL_ROUTE_ENTRIES
+#define OPENTHREAD_CONFIG_NETDATA_PUBLISHER_DESIRED_NUM_EXTERNAL_ROUTE_ENTRIES 10
```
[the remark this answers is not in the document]
abtink: Yes, it is per external route prefix.

src/core/net/srp_server.hpp
```diff
@@ -42,6 +42,10 @@
 #error "OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE is required for OPENTHREAD_CONFIG_SRP_SERVER_ENABLE"
 #endif
 
+#if !OPENTHREAD_CONFIG_NETDATA_PUBLISHER_ENABLE
+#error "OPENTHREAD_CONFIG_NETDATA_PUBLISHER_ENABLE is required for OPENTHREAD_CONFIG_SRP_SERVER_ENABLE"
```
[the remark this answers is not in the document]
abtink: It is already this way in the code. 😃 . Here is the definition in  `src/core/config/netdata_publisher.h`:

```c++
#ifndef OPENTHREAD_CONFIG_NETDATA_PUBLISHER_ENABLE
#define OPENTHREAD_CONFIG_NETDATA_PUBLISHER_ENABLE OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
#endif
```

src/core/thread/network_data_service.cpp
abtink: > Can we keep this method unchaged, but change HandleServerDataUpdated to NOT call SynchronizeServerData directly, but post a tasklet to call SynchronizeServerData ?

This is a good idea. I think we considered something similar this a long time ago but from I somewhat recall there were some corner cases that needed some care/thought to handle.

I think this is good thing to consider for a future enhancement/PR (it would be a bigger behavior change of `NetworkData::Notifier`).

----

In the first version of this change I had a paramter `aRegisterWithLeader` in `AddService()`, ... which indicated whether the `AddService()` itself should do the registeration with leader or leave it for the caller to do. This made it somewhat more complex to use. I changed and simplified the `AddService()` per @jwhui  suggestion from https://github.com/openthread/openthread/pull/6587#discussion_r633219552 which make it consistent with other  network data entries such as on-mesh prefixes and routes.

src/core/thread/network_data_publisher.hpp
```diff
@@ -0,0 +1,374 @@
...
+     * (from earlier call to any of `PublishDnsSrpService{Type}()` methods).
+     *
+     * This method publishes the "DNS/SRP Service Unicast Address" by including the address and port info in the
+     * Service TLV data.
+     *
+     * @param[in] aAddress   The DNS/SRP server address to publish.
+     * @param[in] aPort      The SRP server port number to publish.
+     *
+     */
+    void PublishDnsSrpServiceUnicast(const Ip6::Address &aAddress, uint16_t aPort)
```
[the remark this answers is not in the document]
abtink: I agree it should be easy to add some a check. I suggest to consider this when/if we enable such a use-case.

src/core/thread/network_data_publisher.hpp
```diff
@@ -0,0 +1,374 @@
...
+
+        bool Add(void);
+        bool Remove(State aNextState = kNoEntry);
+        void Process(void);
+        void CountOnMeshPrefixEntries(uint8_t &aNumEntries, uint8_t &aNumPreferredEntries) const;
+        void CountExternalRouteEntries(uint8_t &aNumEntries, uint8_t &aNumPreferredEntries) const;
+
+        Type        mType;
+        Ip6::Prefix mPrefix;
+        bool        mStable;
```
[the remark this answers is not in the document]
abtink: Yes, we treat them separately since they do behave differently and can be used by nodes in different ways.

Practically, I expect publisher to be used with stable entries.

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1016 @@
...
+#if OPENTHREAD_CONFIG_BORDER_ROUTER_ENABLE
+
+Error Publisher::PublishOnMeshPrefix(const OnMeshPrefixConfig &aConfig)
+{
+    Error        error;
+    PrefixEntry *entry;
+
+    VerifyOrExit(aConfig.IsValid(GetInstance()), error = kErrorInvalidArgs);
+
+    SuccessOrExit(error = AllocatePrefixEntry(aConfig.GetPrefix(), entry));
```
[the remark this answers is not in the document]
abtink: The new call can be for same prefix but with different flags. Retuning `kErrorAlready` is useful to indicate this.

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1016 @@
...
+}
+
+Error Publisher::PublishExternalRoute(const ExternalRouteConfig &aConfig)
+{
+    Error        error;
+    PrefixEntry *entry;
+
+    VerifyOrExit(aConfig.IsValid(GetInstance()), error = kErrorInvalidArgs);
+
+    SuccessOrExit(error = AllocatePrefixEntry(aConfig.GetPrefix(), entry));
```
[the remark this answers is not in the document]
abtink: Same.

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1016 @@
...
+    return error;
+}
+
+Error Publisher::UnpublishPrefix(const Ip6::Prefix &aPrefix)
+{
+    Error        error = kErrorNone;
+    PrefixEntry *entry;
+
+    entry = FindMatchingPrefixEntry(aPrefix);
+    VerifyOrExit(entry != nullptr, error = kErrorNotFound);
```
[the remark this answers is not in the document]
abtink: Other netdata APIs/method seem to return `kErrorNotFound` on "remove" (`Local::RemoveOnMeshPrefix()`, ...). I think would be good to keep it the same/conisistent here (in this PR). We can consider changing them all in a later enhancement/PR.

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1016 @@
...
+}
+
+bool Publisher::Entry::IsPreferred(uint16_t aRloc16) const
+{
+    // Indicates whether or not an entry from `aRloc16` is preferred
+    // over our entry (based on our RLOC). We prefer an entry from a
+    // router over an entry from an end-device (e.g., a aREED). If both
+    // are the same type, then the one with smaller RLOC16 is preferred.
+
+    bool isOtherRouter = Mle::Mle::IsActiveRouter(aRloc16);
```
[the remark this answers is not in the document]
abtink: Not a typo. It's indicating whether the other node (that we are comparing against) is a router or not?

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1016 @@
...
+    return registerWithLeader;
+}
+
+bool Publisher::DnsSrpServiceEntry::Add(void)
+{
+    // Adds the service entry to the network data.
+
+    bool registerWithLeader = false;
+
+    VerifyOrExit(GetState() == kAdding);
```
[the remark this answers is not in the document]
abtink: Good idea. I actually removed it fully (it's checked in other code paths).

src/core/thread/network_data_publisher.hpp
```diff
@@ -0,0 +1,374 @@
...
+namespace NetworkData {
+
+/**
+ * This class implements the Network Data Publisher.
+ *
+ * It provides mechanisms to limit the number of similar Service and/or Prefix (on-mesh prefix or external route)
+ * entries in the Thread Network Data by monitoring the Network Data and managing if or when to add or remove entries.
+ *
+ */
+class Publisher : public InstanceLocator, private NonCopyable
```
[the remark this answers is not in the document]
abtink: I like the shorter name `Publisher`, since it's going to be added/used as prefix in all other definitions.

tests/scripts/thread-cert/node.py
abtink: Good suggestion. Will change it in next push.

src/core/api/netdata_publisher_api.cpp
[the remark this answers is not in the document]
simonlingoogle: Thanks for the explanation. 
I think this API is also useful if the SRP server is not the BR, but another device on the infrastructure link. In such case, the BR will need to publish the SRP service for the actual SRP server.

src/core/config/netdata_publisher.h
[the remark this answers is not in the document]
simonlingoogle: Thanks

src/core/net/srp_server.hpp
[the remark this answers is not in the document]
simonlingoogle: Thanks

tests/scripts/thread-cert/test_netdata_publisher.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+class NetDataPublisher(thread_cert.TestCase):
+    USE_MESSAGE_FACTORY = False
+    SUPPORT_NCP = False
+
+    TOPOLOGY = {
+        LEADER: {
+            'name': 'LEADER',
+            'masterkey': '00112233445566778899aabbccddeeff',
```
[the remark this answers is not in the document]
abtink: Looks like we can skip it and rely on the default. Thanks for the hint. I remove it in this PR.

I am getting to learn more on the `thred-cert` test code. I noticed some of other existing tests specify the key in `TOPOLOGY` definition so there is an opportunity to harmonize this.

src/core/config/netdata_publisher.h
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * @def OPENTHREAD_CONFIG_NETDATA_PUBLISHER_DESIRED_NUM_ANYCAST_DNS_SRP_SERVICE_ENTRIES
+ *
+ * Specifies the desired number of "DNS/SRP Service Anycast Address" entries in the Thread Network Data.
+ *
+ * Publisher attempts to limit the number of such entries to this value.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_NETDATA_PUBLISHER_DESIRED_NUM_ANYCAST_DNS_SRP_SERVICE_ENTRIES
+#define OPENTHREAD_CONFIG_NETDATA_PUBLISHER_DESIRED_NUM_ANYCAST_DNS_SRP_SERVICE_ENTRIES 8
```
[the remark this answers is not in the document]
abtink: I derived this trying to limit the overall TLV size to ~40 bytes.

For this type, we have 6 bytes for "Service TLV" and its data and then each "Server TLV" is 4 bytes. 
So total is `6 + 4x` (where `x` is number of nodes adding such entry) and with `8` we have `6+4x8 = 38`.

src/cli/README_NETDATA.md
```diff
@@ -157,9 +159,36 @@ Print netdata help menu.
...
+
+### publish
+
+The Network Data Publisher provides mechanisms to limit the number of similar DNS/SRP service entries in the Thread Network Data by monitoring the Network Data and managing if or when to add or remove entries.
+
+The Publisher requires `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_ENABLE` and `OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE`.
+
+The following formats are available: :
+
+- `netdata publish dnssrp anycast <seq-num>` to publish "DNS/SRP Service Anycast Address" with a given sequence number to be published in the Thread Network Data.
```
jwhui, line 178:
```suggestion
- `netdata publish dnssrp anycast <seq-num>` to publish "DNS/SRP Service Anycast Address" with a given sequence number in the Thread Network Data.
```

src/cli/cli_network_data.cpp
```diff
@@ -216,6 +217,78 @@ otError NetworkData::ProcessHelp(uint8_t aArgsLength, Arg aArgs[])
...
+    VerifyOrExit(aArgsLength >= 3, error = OT_ERROR_INVALID_ARGS);
+
+    if (aArgs[0] == "dnssrp")
+    {
+        if (aArgs[1] == "anycast")
+        {
+            uint8_t sequenceNumber;
+
+            VerifyOrExit(aArgsLength == 3, error = OT_ERROR_INVALID_ARGS);
+            SuccessOrExit(error = aArgs[2].ParseAsUint8(sequenceNumber));
```
jwhui, line 234: Leverage #6700 here?

src/core/thread/network_data_service.cpp
jwhui: Posting a tasklet works well when the APIs are being called from the same tasklet. However, posting a tasklet does not handle the case where APIs are being invoked as a result of CLI (or spinel in the case of NCP) interaction, even when the CLI is being driven by a script.

src/cli/cli_network_data.cpp
abtink: Yes. 😄 Changed it in new push.

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,469 @@
...
+            // If the entry is already added, we need to update it
+            // so we remove it and add it back immediately with
+            // the new mesh-local address.
+
+            Remove(/* aNextState */ kAdding);
+            Add();
+        }
+    }
+
+    if (aEvents.ContainsAny(kEventThreadNetdataChanged | kEventThreadRoleChanged))
```
librasungirl, line 434: is it necessary to process for `kEventThreadRoleChanged` here? 
srp server itself processes role change event to publish or unpublish accordingly.
https://github.com/openthread/openthread/blob/ac21e00fa160988619c6a8a55967a92f436fb54b/src/core/net/srp_server.cpp#L492

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,469 @@
...
+    // router over an entry from an end-device (e.g., a REED). If both
+    // are the same type, then the one with smaller RLOC16 is preferred.
+
+    bool isOtherRouter = Mle::Mle::IsActiveRouter(aRloc16);
+
+    return (Get<Mle::Mle>().IsRouterOrLeader() == isOtherRouter) ? (aRloc16 < Get<Mle::Mle>().GetRloc16())
+                                                                 : isOtherRouter;
+}
+
+void Publisher::UpdateState(uint8_t aNumEntries, uint8_t aNumPreferredEntries, uint8_t aDesiredNumEntries)
```
librasungirl, line 258: Some rough thoughts for discussion
- `aDesiredNumEntries` (i.e. `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_DESIRED_NUM_UNICAST_DNS_SRP_SERVICE_ENTRIES` or `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_DESIRED_NUM_ANYCAST_DNS_SRP_SERVICE_ENTRIES`) is configurable, if BRs from different vendors have different configuration, the BR configured with larger value seems always win.  
e.g. if `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_DESIRED_NUM_UNICAST_DNS_SRP_SERVICE_ENTRIES` in BR1 is 2, and 3 for BR2 and BR3, BR1 would deregister its service though it might have smaller RLOC16.  Is it expected?
- Should we keep SRP server preferred if it has service records registered already? 
- Should SRP server be notified so that it could remove services registered (and unpublish in advertising proxy) if it re-sign its server role?
- Should SRP client on hold service registration for a while to avoid registering to some server which will be removed?

src/core/thread/network_data_publisher.cpp
abtink: Yes. It is necessary.

`Publisher` can be used independent of whether OT SRP server feature is being used or not (e.g. in cases where app/next layer is providing SRP sever functionality).

src/core/thread/network_data_publisher.cpp
abtink: Thanks @librasungirl .

On the publisher configs, we (ideally) want to make sure all BRs behave the same, use the `Publisher` logic with the same set of configs. 

In the last Thread TC call, I brought this up and there was agreement to add this (and the specific configs) as part of Duckhorn BR spec (and may be even cover this as part of cert tests).

----

> Should SRP server be notified so that it could remove services registered (and unpublish in advertising proxy) if it re-sign its server role?

Yes. I think so. I mentioned in one of our internal meeting that as next step we want to add callback to notify next layer and/or SRP server when entry is actually added/removed (to take any action they want). I was hopping/planning to add this as follow up PR but since this PR has been open and under review for rather long time I may end up adding it to the same PR. 😄 

In general, I prefer it if we can all help review PRs more quickly and try to add feature in smaller PRs which build on top of the previous ones. This would be easier to review and also develop and in general I think more agile.

----

> Should we keep SRP server preferred if it has service records registered already?

Good idea. We can add a new API to mark the entry s preferred on device. Again we can add this later in a new PR.

---

> Should SRP client on hold service registration for a while to avoid registering to some server which will be removed?

I would not go for such enhancement unless we have actual data to show there is an issue.

abtink:
I added two additional commit on top of the previous one (kept it seperate for easier review)
- Adding callback support to `Publisher`
- Updates the `Srp::Server` to use the callback and start/stop when related entry is actually added/removed in Network Data.

src/core/thread/network_data_publisher.cpp
simonlingoogle: > Yes. I think so. I mentioned in one of our internal meeting that as next step we want to add callback to notify next layer and/or SRP server when entry is actually added/removed (to take any action they want).

Why do we need to notify SRP server? I mean are there any actual use case for this feature?

simonlingoogle: > * `aDesiredNumEntries` (i.e. `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_DESIRED_NUM_UNICAST_DNS_SRP_SERVICE_ENTRIES` or `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_DESIRED_NUM_ANYCAST_DNS_SRP_SERVICE_ENTRIES`) is configurable, if BRs from different vendors have different configuration, the BR configured with larger value seems always win.
>   e.g. if `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_DESIRED_NUM_UNICAST_DNS_SRP_SERVICE_ENTRIES` in BR1 is 2, and 3 for BR2 and BR3, BR1 would deregister its service though it might have smaller RLOC16.  Is it expected?

Great point! Seems we should make this feature into the specification?

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,486 @@
...
+    SetState(kToAdd);
+    Process();
+
+exit:
+    return;
+}
+
+void Publisher::SetState(State aState)
+{
+    State oldState = aState;
```
librasungirl, line 121:
```suggestion
    State oldState = mState;
```

src/core/thread/network_data_publisher.hpp
```diff
@@ -0,0 +1,225 @@
...
+{
+    friend class ot::Notifier;
+
+public:
+    /**
+     * This type represents the callback function pointer used to notify when a "DNS/SRP Service" entry is added to or
+     * removed from the Thread Network Data.
+     *
+     * On remove the callback is invoked independent of whether the entry is removed by `Publisher` (e.g., when there
+     * too many similar entries already present in the Network Data) or through an explicit call to unpublish the entry
```
librasungirl, line 78:
```suggestion
     * are too many similar entries already present in the Network Data) or through an explicit call to unpublish the entry
```

src/core/net/srp_server.cpp
```diff
@@ -101,24 +103,29 @@ void Server::SetServiceHandler(otSrpServerServiceUpdateHandler aServiceHandler,
...
     mEnabled = aEnabled;
 
-    if (!mEnabled)
+    if (mEnabled)
     {
-        Stop();
+        // Select a port and then publish "DNS/SRP Unicast Address
+        // Service" in Thread Network Data using the device's
+        // mesh-local EID as the address. Then wait for callback to
+        // start server operation when entry is is published (i.e.,
```
librasungirl, line 117:
```suggestion
        // start server operation when entry is published (i.e.,
```

src/core/thread/network_data_publisher.cpp
abtink: > Great point! Seems we should make this feature into the specification?

@simonlingoogle I did mention that I brought this up ion last Thread Tech committee discussion to be included in the Duckhorn spec (please see https://github.com/openthread/openthread/pull/6587#discussion_r646836099). 


------------

> Why do we need to notify SRP server? I mean are there any actual use case for this feature?

We want server to stop acting as advertising proxy for any previously registered service. Otherwise, the clients that previously registered with it, would notice that its entry is removed fronm Network Data and then try to register with another server which would reject the entries (since old server still proxy advertising the services).

src/core/net/srp_server.cpp
```diff
@@ -447,55 +454,61 @@ void Server::CommitSrpUpdate(Error                    aError,
...
     }
 }
 
 void Server::Stop(void)
 {
-    VerifyOrExit(IsRunning());
+    VerifyOrExit(mRunning);
 
-    UnpublishServerData();
+    mRunning = false;
```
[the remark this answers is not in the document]
abtink: We do want to remove all host/service info. See https://github.com/openthread/openthread/pull/6587#discussion_r647152930.

tests/scripts/thread-cert/test_netdata_publisher.py
```diff
@@ -0,0 +1,270 @@
...
+import command
+import thread_cert
+
+# Test description:
+#   This test verifies network data publisher behavior with DNS/SRP service entries and on-mesh prefix and external
+#   route entries.
+#
+# Topology:
+#
+#   1 leader, 5 routers and 4 end-devices all connected
```
librasungirl, line 42:
```suggestion
#   1 leader, 5 routers and 5 end-devices all connected
```

tests/scripts/thread-cert/test_netdata_publisher.py
```diff
@@ -0,0 +1,270 @@
...
+            self.simulator.go(5)
+            self.assertEqual(router.get_state(), 'router')
+
+        for end_dev in end_devs:
+            end_dev.start()
+            self.simulator.go(5)
+            self.assertEqual(end_dev.get_state(), 'child')
+
+        #---------------------------------------------------------------------------------
+        # DNS/RSP anycast entries
```
librasungirl, line 195:
```suggestion
        # DNS/SRP anycast entries
```

src/core/thread/network_data_publisher.cpp
simonlingoogle: Thanks for explanation. @abtink

tests/scripts/thread-cert/test_netdata_publisher.py
```diff
@@ -0,0 +1,270 @@
...
+        services = leader.get_services()
+        self.assertEqual(len(services), min(1 + len(routers), DESIRED_NUM_ANYCAST))
+        self.verify_anycast_services(services)
+
+        # Publish same entry on all end-devices (5 nodes).
+
+        for node in end_devs:
+            node.netdata_publish_dnssrp_anycast(ANYCAST_SEQ_NUM)
+            print(node.name)
+        self.simulator.go(WAIT_TIME)
```
librasungirl, line 215: may move publish on all end-devices before on all leader and routers, and add verification that leader&routers are preferred over end devices when exceeding the limitation?

tests/scripts/thread-cert/test_netdata_publisher.py
```diff
@@ -0,0 +1,270 @@
...
+            num -= 1
+            services = leader.get_services()
+            self.assertEqual(len(services), min(num, DESIRED_NUM_ANYCAST))
+            self.verify_anycast_services(services)
+
+        #---------------------------------------------------------------------------------
+        # DNS/SRP unicast entries
+
+        # Publish DNS/SRP unicast address on all routers, first using
+        # MLE-EID address, then change to use specific address. Verify
```
librasungirl, line 240:
```suggestion
        # ML-EID address, then change to use specific address. Verify
```

src/core/thread/network_data_publisher.cpp
abtink: Thanks!

tests/scripts/thread-cert/test_netdata_publisher.py
abtink: The preferred entry logic comes into picture when we exceed the desired number and want to remove some entries. There is no easy and reliable way in the test to create a situation where the number of entries exceed the limit. It is due to  how nice the algo works :)... each node picks random wait time and then as soon they see we have enough entries, they stop adding their own entry.

I had this situation covered in the test when I had the "on-mesh/external route" prefix support included. But it is removed. There I could use the "preference" (high/med/low) of prefix entries to create a situation where total number entries would exceed the limit. We cannot use the same with service entries (there is no "preference" in service entries).

abtink: There may be a way around this by manually adding similar "service" entries directly on some node. It's a bit hacky though.

abtink: Added a new test case (in a new commit) where check the when removing entries, publisher prefers and keeps the entries from routers over ones from end-devices.

tests/scripts/thread-cert/test_netdata_publisher.py
```diff
@@ -0,0 +1,307 @@
...
+            self.simulator.go(WAIT_TIME)
+
+            services = leader.get_services()
+            self.assertEqual(len(services), num)
+            self.verify_anycast_services(services)
+
+            service_rlocs = [int(service[4], 16) for service in services]
+
+            for node in [leader, router1, router2]:
+                self.assertIn(node.get_addr16(), service_rlocs)
```
librasungirl, line 303: Check that the router which has newly registered its service is in the Thread Network Data.

```suggestion
            for server in [leader, router1, router2,  node]:
                self.assertIn(server.get_addr16(), service_rlocs)
```

librasungirl (approved):
LGTM with one minor comment 👍

src/core/common/instance.hpp
```diff
@@ -640,6 +640,13 @@ template <> inline NetworkData::Notifier &Instance::Get(void)
 }
 #endif
 
+#if OPENTHREAD_CONFIG_NETDATA_PUBLISHER_ENABLE
+template <> inline NetworkData::Publisher &Instance::Get(void)
+{
+    return mThreadNetif.mNetworkDataPublisher;
```
simonlingoogle, line 646: There seems to be only one `Publisher`? How can we publish multiple items (both SRP unicast and external routes)?

tests/scripts/thread-cert/test_netdata_publisher.py
librasungirl: > There may be a way around this by manually adding similar "service" entries directly on some node. It's a bit hacky though.
The workaround in this PR does sound good to me. the other possible option might be add API to allow configure different publisher entries on child/router, just like what Thread 1.1 Test Harness do for TC 5.2.3, which test maximal active router capacity, by configuring routerupgradethreshold/routerdowngradethreshold.
Such API and CLI command might be required if adding similar active SRP Server capacity test.

tests/scripts/thread-cert/test_netdata_publisher.py
abtink: Changed in new push.

src/core/net/srp_server.cpp
```diff
@@ -101,24 +103,29 @@ void Server::SetServiceHandler(otSrpServerServiceUpdateHandler aServiceHandler,
...
     {
-        Stop();
+        // Select a port and then publish "DNS/SRP Unicast Address
+        // Service" in Thread Network Data using the device's
+        // mesh-local EID as the address. Then wait for callback to
+        // start server operation when entry is is published (i.e.,
+        // added to the Network Data).
+
+        SelectPort();
+        Get<NetworkData::Publisher>().SetDnsSrpServiceCallback(HandleNetDataPublisherEntryChange, this);
```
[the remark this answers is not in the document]
abtink: Changed in new push.

src/core/common/instance.hpp
abtink: The `Publisher` class can provide different methods/APIs for publishing different things. Same way that we have one `NetworkData::Local` object but can use it to add service and/or on-mesh/external-route prefix.

tests/scripts/thread-cert/test_netdata_publisher.py
abtink: That can also work, but in genera, I would prefer and suggest to avoid adding new APIs for testing purposes only if/when we can. The main reason is I think it would be better that the firmware we use for testing to be closest to what we intend to use/ship in products.

src/core/common/instance.hpp
simonlingoogle: But it seems to only be able to publish one thing at a time? 
Should we move Publisher instance to be a member of Srp::Server?

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,492 @@
...
+    mState = aState;
+
+    // Invoke callback to notify the change to the Network Data, when
+    // the entry is added (on transition to new state `kAdded`), or
+    // when it is removed (on transition from state `kAdded`).
+
+    VerifyOrExit((mState == kAdded) || (oldState == kAdded));
+
+#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
+    Get<Srp::Server>().HandleNetDataPublisherEntryChange(mState == kAdded);
```
simonlingoogle, line 135: Seems we are turning off SRP server when `mState` is not `kAdded`. 

However, if `mState` is changing from `kAdded` to `kRemoving`, it's not actually removed from `Network Data`, yet. 
And it's also possible for Publisher to revert the state from `kRemoving` to `kAdded` to cancel the removal. 

Should we keep the SRP server running in such case since the Service TLV is still in `Network Data` ? Clients may still send registrations.

src/core/common/instance.hpp
abtink: > But it seems to only be able to publish one thing at a time?

In current code, yes! We can extend it later. Check this branch [abtink/netdata/publisher-full](https://github.com/abtink/openthread/tree/netdata/publisher-full) if you are interested to see how (this is from earlier with on-mesh/off mesh support treated as individual entries)

----

> Should we move Publisher instance to be a member of Srp::Server?

`Publisher` can be used independent of whether OT SRP server feature is being used or not (e.g. in cases where app/next layer is providing SRP sever functionality itself).

src/core/thread/network_data_publisher.cpp
abtink: Good point! We can stop the server too early.

In new push, I changed how and when `Publisher` notifies (emits events  and invokes callback). I kept it as a separate commit.

abtink:
Submitted https://github.com/openthread/openthread/pull/6768 which replaces this PR and contains the full implementation of `Publisher` (for both SRP/DNS service and on-mesh/external-route prefix entries).


> Upon further discussion, we realized related the prefix entries
> (on-mesh and/or external route) on a device it may be desirable to
> treat them as a set (i.e., have them all be added together and/or
> removed together). The earlier implementation treats every prefix entry
> individually. Such a change requires some additional thoughts and
> solutions (e.g., on how to decide in a combined way (how to count
> multiple existing entries) when/if to add/remove all prefix entries as a set).
> So in the latest push in this PR/branch removed prefix support and
> only contains support for publishing DNS/SRP service entries.


The  [SPEC-985](https://threadgroup.atlassian.net/browse/SPEC-985) discussion in Thread TC workgroup changes how domain ID is managed (requires BRs to choose the domain ID when registering the entries directly). This change makes it easier to manage the prefix entries on a BR and in `Publisher` we no longer need to treat on-mesh and external-route prefixes all together as a set and can treat each prefix entry individually. This not only helps simplify the `Publisher` implementation (now in https://github.com/openthread/openthread/pull/6768), it also uses the Thread Network Data (which has limited size) more efficiently.

abtink:
Closing this PR in favor of 
- #6768

src/core/thread/network_data_service.cpp
simonlingoogle: > Posting a tasklet works well when the APIs are being called from the same tasklet. However, posting a tasklet does not handle the case where APIs are being invoked as a result of CLI (or spinel in the case of NCP) interaction, even when the CLI is being driven by a script.

I don't understand why. Could you give more explanation?

Calling `SynchronizeServerData` in a tasklet allows multiple changes using `AddService` or `RemoveService` to be merged in one `NETDATA.ntf`, right after this loop.

jwhui: The tasklet model only consolidates multiple updates when `AddService` and `RemoveService` are called from the same task context. When issuing commands via CLI or spinel that invoke calls to `AddService` and `RemoveService` via CLI or spinel, there is no way to guarantee that such calls will happen in the same task context.

simonlingoogle: > The tasklet model only consolidates multiple updates when AddService and RemoveService are called from the same task context.

Yes. But isn't this solve our problem? Currently we are calling multiple AddService/RemoveService in the same call stack (and hense same Tasklet) and use `registerWithLeader` to track if we need to call `HandleServerDataUpdated`. But if we just post a tasklet for running `HandleServerDataUpdated` whenever we call AddService/RemoveService, we don't need to use an additional boolean (i.e. registerWithLeader) to track if we need to call `HandleServerDataUpdated`, which will simplify our code. It also makes calling AddService/RemoveService simpler, not having to call  `HandleServerDataUpdated` afterwards.

jwhui: I don't think we necessarily achieve the same benefit if someone is invoking `AddPrefix` and `RemovePrefix` using the CLI. How do we guarantee that the `prefix add` and `prefix remove` commands will be sent and processed within the same task context?

PR #6596, opened 2021-05-09
[netdata] simplify config definitions and implementations

abtink:
This commit adds a new header file `networks_data_type.hpp` and its
related `cpp` file which now contain all the Network Data related
types and config definitions (for on-mesh prefix, external route,
service, and server entries) and their implementation (moved from
`network_data.hpp`).

It adds `IsValid()`, `ConvertToTlvFlags()`, and `SetFromTlvFlags()` in
`OnMeshPrefixConfig` and `ExternalRouteConfig` to validate the config
and convert it to/from the corresponding TLV bitmask flags. These
changes help simplify the use of the configs by other classes.

--------

_Notes on this PR:_

Currently, there are two commits in this PR. I kept the commit which defines 
the new `network_data_type.hpp/cpp` and moves the existing code form 
`network_data.hpp/cpp` to the new source files as separate first commit. 
The second commit contains all the code/implementation changes. This is
to help with review of the changes. **So please review the second commit**.
Before merge we can squash the two commits into one.

The new methods in PR converting the configs to TLV flags should help
with adding support for on-mesh prefix and off-mesh route entries in 
`NetworkData::Publisher` from https://github.com/openthread/openthread/pull/6587.


src/core/thread/network_data_types.hpp
```diff
@@ -0,0 +1,250 @@
...
+class Instance;
+
+namespace NetworkData {
+
+/**
+ * @addtogroup core-netdata-core
+ *
+ */
+
+// Forward decelerations
```
jwhui, line 56:
```suggestion
// Forward declarations
```

jwhui (approved):

PR #6597, opened 2021-05-09
[srp-client] disallow adding of duplicate services (same service/instc name)

abtink:
This commit updates `Srp::Client::AddService()` to reject and return
`kErrorAlready` if the caller tries to add duplicate service entries
(services with matching service and instance names).

### DIFF
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (112)
+#define OPENTHREAD_API_VERSION (113)
 
 /**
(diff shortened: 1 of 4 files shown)


simonlingoogle (approved):
LGTM :+1:

wgtdkp (approved):
LGTM :+1:

jwhui (approved):

abtink:
> @abtink, we discussed this before in #6257 (comment) :)

Yes I remember. 😃 
The `+48` code size is below what I expected. So I think it's worth the protection it adds. 

Also,  I assumed that server would reject a duplicate service registration but in the case it is from same device (with same key), server just simply treats it as a refresh for the same service, and so remembers the last one (which I think is actually correct/intended behavior).

Issue #6586, opened 2021-05-07
Potential vulnerability in remove - re-add service pattern.


kkasperczyk-no:
**Problem**
Some time ago we discussed with @abtink @LuDuda @Damian-Nordic and @jwhui the problem of missing service update feature and we concluded it that this is how SRP protocol works like, so in case of necessity to update specific service record it should be removed and re-added once again. 

I introduced that solution in the Project CHIP, but @msandstedt noticed (https://github.com/project-chip/connectedhomeip/pull/6539#issuecomment-833593097) that there is potential security problem with this pattern and bad actor may try to steal removed service name and add it with own key, before the SRP client will manage to do that again. I'm not able to judge if it is a real threat, but seems to be worth considering. @abtink @jwhui could you comment on that?

**Alternatives**
I think that risk could be reduced by sending remove and following add requests in one SRP Update message thanks to the  OPENTHREAD_CONFIG_SRP_CLIENT_UPDATE_TX_DELAY feature (https://github.com/openthread/openthread/blob/main/src/core/config/srp_client.h#L162), however it doesn't seem to solve problem in 100%, as for some reason OpenThread thread could be blocked by some other application thread and that would result in TX_DELAY expire and sending two separate SRP Updates.

**Proposed solution**
I wonder if it would be reasonable to implement some API allowing to force sending few requests in one SRP Update message. I imagine some `StartTransaction` method that would start gathering requests and `StopTransaction` that would stop it and send one SRP Update filled with whole gathered data at once.

wgtdkp:
> Some time ago we discussed with @abtink @LuDuda @Damian-Nordic and @jwhui the problem of missing service update feature and we concluded it that this is how SRP protocol works like, so in case of necessity to update specific service record it should be removed and re-added once again.

Not sure if I understand the problem correctly, but if your requirement is to update service RRs like TXT entries or port number in SRV RR, I think SRP supports this well. The client can always re-new the SRP service with different RRs.

I see you PR in the CHIP repo https://github.com/project-chip/connectedhomeip/pull/6539. The problem is that you are trying to add two different services with the same instance and service name and I believe this is not CHIP want to do: the device want to update the RRs associated with the same service (identified by the union of `service name` and `instance name`), right?

We should reuse the SRP service entry with the same name on the CHIP end device but not always adding a new one.

kkasperczyk-no:
[...]

PR #6598, opened 2021-05-09
[srp-client] add `Srp::Client::ClearService()`

abtink:
This method clears a service from SRP client, immediately removing it
from the list. Unlike `RemoveService()` which sends an update message
to server to remove/unregister the service, this method clears the
service from client list without any interaction with server. On a
successful call to this method, the `Callback` will NOT be called and
service entry can be reclaimed and re-used by the caller immediately.

This method can be used along with a subsequent call to `AddService()`
(potentially reusing the same service entry with the same service and
instance names) to update some of the parameters in an existing
service.

This commit adds `otSrpClientClearService()` as a related public OT
API and support for this in CLI (command `service delete`). It also
adds support for the deleting service in `NcpBase` and spinel.

### DIFF
--- a/include/openthread/srp_client.h
+++ b/include/openthread/srp_client.h
@@ -474,6 +474,26 @@ otError otSrpClientAddService(otInstance *aInstance, otSrpClientService *aServic
 otError otSrpClientRemoveService(otInstance *aInstance, otSrpClientService *aService);
 
 /**
+ * This function clears a service, immediately removing it from the client service list.
+ *
+ * Unlike `otSrpClientRemoveService()` which sends an update message to the server to remove the service, this function
+ * clears the service from the client's service list without any interaction with the server. On a successful call to
+ * this function, the `otSrpClientCallback` will NOT be called and the @p aService entry can be reclaimed and re-used
+ * by the caller immediately.
+ *
+ * This function can be used along with a subsequent call to `otSrpClientAddService()` (potentially reusing the same @p
+ * aService entry with the same service and instance names) to update some of the parameters in an existing service.
+ *
+ * @param[in] aInstance        A pointer to the OpenThread instance.
+ * @param[in] aService         A pointer to a `otSrpClientService` instance to delete.
+ *
+ * @retval OT_ERROR_NONE       The @p aService is deleted successfully. It can be reclaimed and re-used immediately.
+ * @retval OT_ERROR_NOT_FOUND  The service could not be found in the list.
+ *
+ */
+otError otSrpClientClearService(otInstance *aInstance, otSrpClientService *aService);
+
+/**
  * This function gets the list of services being managed by client.
  *
  * @param[in] aInstance        A pointer to the OpenThread instance.
@@ -507,8 +527,8 @@ otError otSrpClientRemoveHostAndServices(otInstance *aInstance, bool aRemoveKeyL
 /**
  * This function clears all host info and all the services.
  *
- * Unlike `otSrpClientRemoveHostAndServices()` which sends an update message to server to remove/unregister all the
- * info, this function clears all the info immediately without any interaction with server.
+ * Unlike `otSrpClientRemoveHostAndServices()` which sends an update message to the server to remove all the info, this
+ * function clears all the info immediately without any interaction with the server.
  *
  * @param[in] aInstance        A pointer to the OpenThread instance.
  *
--- a/src/core/net/srp_client.hpp
+++ b/src/core/net/srp_client.hpp
@@ -500,6 +500,22 @@ public:
     Error RemoveService(Service &aService);
 
     /**
+     * This method clears a service, immediately removing it from the client service list.
+     *
+     * Unlike `RemoveService()` which sends an update message to the server to remove the service, this method clears
+     * the service from the client's service list without any interaction with the server. On a successful call
+     * to this method, the `Callback` will NOT be called and the @p aService entry can be reclaimed and re-used by the
+     * caller immediately.
+     *
+     * @param[in] aService     A service to delete from the list.
+     *
+     * @retval kErrorNone      The @p aService is cleared successfully. It can be reclaimed and re-used immediately.
+     * @retval kErrorNotFound  The service could not be found in the list.
+     *
+     */
+    Error ClearService(Service &aService);
+
+    /**
      * This method gets the list of services being managed by client.
      *
      * @returns The list of services.
@@ -530,8 +546,8 @@ public:
     /**
      * This method clears all host info and all the services.
      *
-     * Unlike `RemoveHostAndServices()` which sends an update message to server to remove/unregister all the info, this
-     * method clears all the info immediately without any interaction with server.
+     * Unlike `RemoveHostAndServices()` which sends an update message to the server to remove all the info, this method
+     * clears all the info immediately without any interaction with the server.
      *
      */
     void ClearHostAndServices(void);
@@ -751,7 +767,6 @@ private:
     void         ChangeHostAndServiceStates(const ItemState *aNewStates);
     void         InvokeCallback(Error aError) const;
     void         InvokeCallback(Error aError, const HostInfo &aHostInfo, const Service *aRemovedServices) const;
-    void         ClearHostInfoAndServices(void);
     void         HandleHostInfoOrServiceChange(void);
     void         SendUpdate(void);
     Error        PrepareUpdateMessage(Message &aMessage);
--- a/src/lib/spinel/spinel.h
+++ b/src/lib/spinel/spinel.h
@@ -4112,7 +4112,9 @@ enum
     /** Format: `A(t(UUSSSd))` - Read/Insert/Remove
      * Required capability: `SPINEL_CAP_SRP_CLIENT`.
      *
-     * This property provide a list/array of services. Data per item is
+     * This property provides a list/array of services.
+     *
+     * Data per item for `SPINEL_CMD_PROP_VALUE_GET` and/or `SPINEL_CMD_PROP_VALUE_INSERT` operation is as follows:
      *
      *   `U` : The service name labels (e.g., "_chip._udp", not the full domain name.
      *   `U` : The service instance name label (not the full name).
@@ -4120,7 +4122,15 @@ enum
      *   `S` : The service priority.
      *   `S` : The service weight.
      *
-     * During remove operation, only service name and service instance name would be used.
+     * For `SPINEL_CMD_PROP_VALUE_REMOVE` command, the following format is used:
+     *
+     *   `U` : The service name labels (e.g., "_chip._udp", not the full domain name.
+     *   `U` : The service instance name label (not the full name).
+     *   `b` : Indicates whether to clear the service entry (optional).
+     *
+     * The last boolean (`b`) field is optional. When included it indicates on `true` to clear the service (clear it
+     * on client immediately with no interaction to server) and on `false` to remove the service (inform server and
+     * wait for the service entry to be removed on server). If it is not included, the value is `false`.
      *
      */
     SPINEL_PROP_SRP_CLIENT_SERVICES = SPINEL_PROP_OPENTHREAD__BEGIN + 23,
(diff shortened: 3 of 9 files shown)


include/openthread/srp_client.h
```diff
@@ -473,6 +473,26 @@ otError otSrpClientAddService(otInstance *aInstance, otSrpClientService *aServic
...
+ * aService entry with the same service and instance names) to update some of the parameters in an existing service.
+ *
+ * @param[in] aInstance        A pointer to the OpenThread instance.
+ * @param[in] aService         A pointer to a `otSrpClientService` instance to delete.
+ *
+ * @retval OT_ERROR_NONE       The @p aService is deleted successfully. It can be reclaimed and re-used immediately.
+ * @retval OT_ERROR_NOT_FOUND  The service could not be found in the list.
+ *
+ */
+otError otSrpClientDeleteService(otInstance *aInstance, otSrpClientService *aService);
```
simonlingoogle, line 494: I think this API is quite confusing and error-prone:
- User may call this method to delete a service and expect the service to be removed on server (without reading the Doxygen carefully).
- Users generally don't need an API to delete a service without deregistering from the server. So, this API should not exist.

If we want to support updating existing service, maybe we can allow user to modify the `otSrpClientService` buffer in place and call an API (e.g. otSrpClientCommitService) to commit the change.

Thoughts?

abtink: > If we want to support updating existing service, maybe we can allow user to modify the otSrpClientService buffer in place and call an API (e.g. otSrpClientCommitService) to commit the change.'

I think allowing caller to change an `otSrpClientService` which it is does not own anymore (has passed it to OT and is now owned by OT core and SRP client) is more error-prone (easier to use incorrectly and can lead to tricky racing bugs). I describe the reason why we do not want to allow this in https://github.com/openthread/openthread/issues/6586#issuecomment-834774771 (copying the relevant part below): 

> Back to the point about allowing an existing/registered `otSrpClientService` instances to be changed by caller (after it is given to the OT):
> - I agree this can help allow reusing the same entry and we can add a new OT API to "update" an existing service (not need to remove and add with two different entry instances).
> - My concern is that if we relax the requirement for `otSrpClientService` to not be changed by caller, it may lead to potential incorrect use and/or tricky bugs.
>    -  For example, we can have a race situation if the next layer code tries to change some fields/pointers in an already registered `otSrpClientService` struct from a different OS context (e.g., different RTOS task) while the OT core is also accessing the same content.


And more in https://github.com/openthread/openthread/issues/6586#issuecomment-835925083:

> Thinking more on this, I came up with an approach to allow caller (next layer code) to re-use an existing service entry to update some service paramter without relaxing our requirement for the `otSrpClientService` not to be changed while it is given to and owned by OT core (`Srp::Client` module).
> 
> The idea is we add a new `DeleteService()` (or `otSrpClientDeleteService()` as public OT API) method
> - This method deletes a service from SRP client, immediately removing it from the list. 
> - Unlike `RemoveService()` which sends an SRP update message to server to remove/unregister the service on server , this method deletes the service from the client list without any interaction with server. 
> - On a successful call to this method, the `Callback` will NOT be called and service entry can be reclaimed and re-used by the caller immediately.
> - Basically after `DeleteService()` the service is removed on the client side, but may still be present on the server till it expires its current lease time.

------

I don't think it will be hard for caller/user to distinguish between `RemoveServcie` and `DeleteService`. We already have `RemoveHostAndService` and `ClearHostAndService` which mirror similar behavior (one informs server, the other one clears everything immediately with no interaction with server).

src/lib/spinel/spinel.h
```diff
@@ -4112,15 +4112,26 @@ enum
...
      *   `S` : The service weight.
      *
-     * During remove operation, only service name and service instance name would be used.
+     * For `SPINEL_CMD_PROP_VALUE_REMOVE` command, the following format is used:
+     *
+     *   `U` : The service name labels (e.g., "_chip._udp", not the full domain name.
+     *   `U` : The service instance name label (not the full name).
+     *   `b` : Indicates whether to delete the service entry (optional).
+     *
+     * The last boolean (`b`) field is optional. When included it indicates on `true` to delete the service (deleter it
```
Irving-cl, line 4131:
```suggestion
     * The last boolean (`b`) field is optional. When included it indicates on `true` to delete the service (delete it
```

Irving-cl (approved):
LGTM 👍

include/openthread/srp_client.h
simonlingoogle: > For example, we can have a race situation if the next layer code tries to change some fields/pointers in an already registered otSrpClientService struct from a different OS context (e.g., different RTOS task) while the OT core is also accessing the same content.

User should already be aware that OT APIs should be called in the OT thread. Furthermore, we can make it clear in the doxygen of `otSrpServiceCommitService` interface.

> Unlike RemoveService() which sends an SRP update message to server to remove/unregister the service on server , this method deletes the service from the client list without any interaction with server. 

The problem is that we are introducing an API `DeleteService`, which however is **NOT** meant to be used to **delete** a service. 

> We already have RemoveHostAndService and ClearHostAndService which mirror similar behavior (one informs server, the other one clears everything immediately with no interaction with server).

Even though we have `otSrpClientClearHostAndServices`, we actually never use it. I think it more logical for user to call `otSrpClientClearHostAndServices` to clear the host and all services on it in the case when the SRP client knows somehow that the SRP server is down. But for `otSrpClientDeleteService`, I can't think of any use case that user may want to delete one service without notifying the SRP server.

---

If we want to provide an API for updating an existing service, I think we should provide an API just for that, and make it clear how user can update the service data. We have already decided in our original design that it's callers' responsibility to manage the `otSrpClientService` buffers, so it should be fine if we continue to follow this design?

Thoughts?

abtink: Thanks @simonlingoogle  for your input, but I don't agree with you on this  and feel strongly (and know from my experience) that _"relaxing the requirement for the `otSrpClientService` to be immutable after ownership is passed to OT"_ is not a good idea and it is error-prone (can lead to tricky bugs and incorrect API use).

> User should already be aware that OT APIs should be called in the OT thread. 

The `otSrpClientServcie` is not an API. It is an object which we expect the caller to allocate it, populate it, and then pass it to OT API. It can be created/populated in another OS thread/task (any OS context). There is no reason to restrict how it is being created by caller in the first place (in general sense) and limit it to same OS thread/task as OT. Once it is given to OT, it is owned by OT and becomes immutable.

> Furthermore, we can make it clear in the doxygen of otSrpServiceCommitService interface.

It is much simpler and easier for developers to understand and follow a rule that once ownership is given to OT the `otSrpClientServcie` become immutable (must not be changed by caller anymore). The ownership model is commonly used in multi-threaded (multi-task) designs.

Compared to something like this:
- `otSrpClientServcie` can be allocated and updated anyway the user want (in any thread/task OS context) before it is passed to OT APIs. 
- But after it is passed to OT APIs, it can be changed/updated only from same OS context where the OT API are being called.

This is complex to explain and hard to follow for users (easy to forget and use incorrectly).

> The problem is that we are introducing an API DeleteService, which however is NOT meant to be used to delete a service.

The `DeleteServcie()` is only meant for deleting a service (without notifying the server) and claiming back ownership of `otSrpClientServcie` immediately.

The caller can choose to use it to update a service by deleting it, claiming back the entry and then adding it back sameway that `RemoveServcie()` can be used today for this.

> If we want to provide an API for updating an existing service, I think we should provide an API just for that, and make it clear how user can update the service data. We have already decided in our original design that it's callers' responsibility to manage the otSrpClientService buffers, so it should be fine if we continue to follow this design?

We don't want to provide an API for updating an existing servcie, because it relaxes the notion of ownership of service entries.

simonlingoogle: > The `DeleteServcie()` is only meant for deleting a service (without notifying the server) and claiming back ownership of `otSrpClientServcie` immediately.
> 
> The caller can choose to use it to update a service by deleting it, claiming back the entry and then adding it back sameway that `RemoveServcie()` can be used today for this.

As I have said, `deleting a service without notifying the server` is almost never what users want to do. And `Delete` is very similar to `Remove`, causing confusion. 

Should we use another name like `otSrpClientDetachService` to make it clear that this API just detaches the `otSrpClientServcie` object from the SRP client, without doing anything else?

abtink: > As I have said, deleting a service without notifying the server is almost never what users want to do

We actually mention one possible use-case of "delete" API in the documentation of the API itself, i.e. if the API users do not want to allocate a new entry to update an existing service, they can use "delete" to reclaim an existing service entry immediately and then to use it again to add. There may be other use-cases in future.

> And Delete is very similar to Remove, causing confusion.

I think we can expect API user to read documentation and understand the behavior. 

Besides we already have other APIs with similar behavior, `otSrpClientRemoveHostAndServices()` and 
`otSrpClientClearHostAndServices()`.

I initially actually used `ClearService()` instead of `DeleteService()` for the new method (it may be evident from name of the branch `srp-clinet/clear-service-method` 😄) but I felt `DeleteService()` was better. I am not sure if `DetachServcie()` name would be more clear.  I don't mind going back to `ClearService()` if other agree.

simonlingoogle (approved):
LGTM :+1: 

Even though the method is a little confusing, it solves the problem we have at hand.

include/openthread/srp_client.h
wgtdkp: @abtink I think it make sense to use the similar name with `otSrpClientClearHostAndServices` to help user use the API correctly. That said, I lean to `ClearService` even that `clear` usually means remove a bunch of things. But considering that a service includes a few RRs (PTR, SRV, TXT) and buffers, I think it should not be very strange to call it `clear` 😄 .

wgtdkp (approved):
LGTM 👍

include/openthread/srp_client.h
abtink: In new pushed commit, renamed it back to` ClearService()`.

include/openthread/srp_client.h
```diff
@@ -473,6 +473,26 @@ otError otSrpClientAddService(otInstance *aInstance, otSrpClientService *aServic
  */
 otError otSrpClientRemoveService(otInstance *aInstance, otSrpClientService *aService);
 
+/**
+ * This function clears a service, immediately removing it from the client service list.
+ *
+ * Unlike `otSrpClientRemoveService()` which sends an update message to server to remove/unregister the service, this
```
jwhui, line 479:
```suggestion
 * Unlike `otSrpClientRemoveService()` which sends an update message to the server to remove the service, this
```

src/core/net/srp_client.hpp
```diff
@@ -499,6 +499,22 @@ class Client : public InstanceLocator, private NonCopyable
 
     Error RemoveService(Service &aService);
 
+    /**
+     * This method clears a service, immediately removing it from the client service list.
+     *
+     * Unlike `RemoveService()` which sends an update message to server to remove/unregister the service, this
```
jwhui, line 505:
```suggestion
     * Unlike `RemoveService()` which sends an update message to the server to remove the service, this
```

src/lib/spinel/spinel.h
```diff
@@ -4112,15 +4112,26 @@ enum
...
+     * For `SPINEL_CMD_PROP_VALUE_REMOVE` command, the following format is used:
+     *
+     *   `U` : The service name labels (e.g., "_chip._udp", not the full domain name.
+     *   `U` : The service instance name label (not the full name).
+     *   `b` : Indicates whether to clear the service entry (optional).
+     *
+     * The last boolean (`b`) field is optional. When included it indicates on `true` to clear the service (clear it
+     * on client immediately with no interaction to server) and on `false` to remove the service (inform server and
+     * wait for the service entry to be removed on server). If it is not specified service will be removed (informing
+     * the server).
```
jwhui, line 4134:
```suggestion
     * wait for the service entry to be removed on server). If it is not included, the value is `false`.
```

jwhui (approved):

PR #6600, opened 2021-05-10
[spinel] add spinel prop get/set log

Irving-cl:
This PR adds logging for spinel property when `Get/Set` method is 
called. For example, we have a property PROP_NAME whose format is
`t(bS)` and two arguments: a `BOOL(true)` and a `UINT16(3)`, we will 
log it as **`PROP_NAME: {(BOOL: 1), (UINT16: 3)}`**

The logging is implemented by a new method `spinel_prop_display`. 
`spinel_prop_display` accepts the spinel pack format and an argument
list, and then output the property to a buffer specified.

Note: The PR isn't completed and I haven't fully tested it (I need to
 refine the doc and add more tests). I'd like to have suggestions at 
an early stage so that I can avoid detours. Please help me review 
from a high-level perspective.

### DIFF
--- a/src/lib/spinel/radio_spinel.hpp
+++ b/src/lib/spinel/radio_spinel.hpp
@@ -856,6 +856,12 @@ private:
      */
     otError Remove(spinel_prop_key_t aKey, const char *aFormat, ...);
 
+    /**
+     * This method logs a property operated by get/set method.
+     *
+     */
+    void LogSpinelProperty(spinel_prop_key_t aKey, const char *aFormat, bool aGet, va_list aArgs);
+
     spinel_tid_t GetNextTid(void);
     void         FreeTid(spinel_tid_t tid) { mCmdTidsInUse &= ~(1 << tid); }
 
@@ -1007,6 +1013,11 @@ private:
     int64_t  mRadioTimeOffset;      ///< Time difference with estimated RCP time minus host time.
 
     MaxPowerTable mMaxPowerTable;
+
+#if (OPENTHREAD_CONFIG_LOG_LEVEL >= OT_LOG_LEVEL_DEBG)
+    static constexpr uint16_t kPropLogBufferSize = 256;
+    char                      mPropLogBuffer[kPropLogBufferSize];
+#endif
 };
 
 } // namespace Spinel
--- a/src/lib/spinel/radio_spinel_impl.hpp
+++ b/src/lib/spinel/radio_spinel_impl.hpp
@@ -49,6 +49,7 @@
 #include "lib/platform/exit_code.h"
 #include "lib/spinel/radio_spinel.hpp"
 #include "lib/spinel/spinel_decoder.hpp"
+#include "lib/spinel/spinel_prop_display.hpp"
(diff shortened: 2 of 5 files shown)


src/lib/spinel/spinel.h
```diff
@@ -4951,6 +4951,60 @@ SPINEL_API_EXTERN const char *spinel_link_metrics_status_to_cstr(uint8_t status)
...
+ * @retval  SPINEL_PROP_DISPLAY_ERROR_BUFFER_OVERFLOW        @p `buf_size` is not enough to write the content.
+ * @retval  SPINEL_PROP_DISPLAY_ERROR_INVALID_ARGUMENTS      @p `buf` is `NULL`.
+ * @retval  SPINEL_PROP_DISPLAY_ERROR_UNSUPPORTED_DATA_TYPE  @p `data_type` is not supported.
+ *
+ */
+SPINEL_API_EXTERN int32_t spinel_prop_display_simple_data_type(spinel_datatype_t data_type,
+                                                               bool              ptr_args,
+                                                               va_list           args,
+                                                               char *            buf,
+                                                               uint16_t          buf_size);
```
abtink, line 4982: Will different error be useful here. We can consider maing it `int` and return `-1` for all errors (which would be similar to `spinel_datatype_unpack()`.

src/lib/spinel/spinel.c
```diff
@@ -1634,6 +1634,188 @@ const char *spinel_capability_to_cstr(spinel_capability_t capability)
...
+
+    case SPINEL_DATATYPE_EUI48_C:
+    {
+        spinel_eui48_t *arg_ptr = va_arg(args, spinel_eui48_t *);
+        bytes = snprintf(buf, buf_size, "(EUI48: %x:%x:%x:%x:%x:%x)", arg_ptr->bytes[0], arg_ptr->bytes[1],
+                         arg_ptr->bytes[2], arg_ptr->bytes[3], arg_ptr->bytes[4], arg_ptr->bytes[5]);
+        require_action(bytes > 0 && bytes < buf_size, bail, ret = SPINEL_PROP_DISPLAY_ERROR_BUFFER_OVERFLOW);
+        break;
+    }
+
```
abtink, line 1752: I guess we can handle more complex formats (structs and array, etc) as well.
For example, for `struct` we can probably use recursive calls to the same function to handle the struct fields.

abtink (commented):
Thank @Irving-cl . Conceptually this sounds good to me. We had something similar to this in mind for `wpantund` (we ended up logging subset of commands directly).

I think this kind of logging may be useful and make sense on the host side but not on RCP/NCP side (it will add a lot of overhead and also the log itself is a spinel frame send to host).

Irving-cl:
Thanks! @abtink 

> I think this kind of logging may be useful and make sense on the host side but not on RCP/NCP side (it will add a lot of overhead and also the log itself is a spinel frame send to host).

Yeah, this is only intended to be used on host side.

Irving-cl:
Close for now

PR #6605, opened 2021-05-11
[cli] process exit in ot-ctl

bukepo:
The exit command should only exit ot-ctl itself in daemon mode. This
commit adds processing exit command in the ot-ctl.

This commit also moves the exit command implementation out from core
as a user command on simulation and posix cli mode.

### DIFF
--- a/src/posix/client.cpp
+++ b/src/posix/client.cpp
@@ -81,11 +81,45 @@ static_assert(kLineBufferSize >= sizeof("Error "), "kLineBufferSize is too small
 
 int sSessionFd = -1;
 
+void QuitOnExit(const char *aBuffer)
+{
+    constexpr char kExit[] = "exit";
+
+    while (*aBuffer == ' ' || *aBuffer == '\t')
+    {
+        ++aBuffer;
+    }
+
+    VerifyOrExit(strstr(aBuffer, kExit) == aBuffer);
+
+    aBuffer += sizeof(kExit) - 1;
+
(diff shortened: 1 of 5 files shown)


src/posix/client.cpp
```diff
@@ -81,11 +81,45 @@ static_assert(kLineBufferSize >= sizeof("Error "), "kLineBufferSize is too small
...
 int sSessionFd = -1;
 
+void QuitOnExit(const char *aBuffer)
+{
+    constexpr char kExit[] = "exit";
+
+    while (*aBuffer == ' ' || *aBuffer == '\t')
+    {
+        ++aBuffer;
+    }
```
simonlingoogle, line 91:
```suggestion
```
nit: not necessary to skip spaces? (more strict)

src/posix/client.cpp
```diff
@@ -81,11 +81,45 @@ static_assert(kLineBufferSize >= sizeof("Error "), "kLineBufferSize is too small
...
+
+    VerifyOrExit(strstr(aBuffer, kExit) == aBuffer);
+
+    aBuffer += sizeof(kExit) - 1;
+
+    while (*aBuffer == ' ' || *aBuffer == '\t')
+    {
+        ++aBuffer;
+    }
```
simonlingoogle, line 100: ditto

simonlingoogle (approved):
LGTM :+1: 
with nits

Issue #6602, opened 2021-05-10
[github-actions] MATN_15_ChangeOfPrimaryBBRTriggersRegistration.py fails intermittently


jwhui:
[logs_75543.zip](https://github.com/openthread/openthread/files/6453904/logs_75543.zip)
[thread-border-router-results.zip](https://github.com/openthread/openthread/files/6453905/thread-border-router-results.zip)

librasungirl:
I think It could still be considered as some kind of optimization if the device is going to become router in short time (like in 5s), to avoid invalid MLR transaction and the MLR.rsp retransmission in MAC layer as in this failing run.

Agree with your founding on the difference between DUA and MLR when RLOC changes. However for DUA, we defer the DUA.req if the device `IsExpectedToBecomeRouter()`, and for MLR, we defer MLR.req if the device is going to become router soon (in a short window).

PR #6607, opened 2021-05-11
[github-actions] Fix MATN_15_ChangeOfPrimaryBBRTriggersRegistration.py

superwhd:
The test case failed because pktverify failed to find an ACK packet sent to `TD`. However, `TD_RLOC` has changed during the test while we were using the latter RLOC to verify the packet. Therefore it wasn't able to find the packet. 

As said by @simonlingoogle in: https://github.com/openthread/openthread/issues/6602

> One finding is that the TD registers it's MA just at the same moment as it sends ADDR_SOL.req.
This caused the TD to change its RLOC, so that it can not receive the MLR.rsp.

For this test case, we can verify the existence of that ACK packet using the source address of its according CON packet.


simonlingoogle (approved):
LGTM 👍

Issue #1619, opened 2017-04-19
MTD vrs FTD builds, and Makefile.am If/Else/Endif issues


DuaneEllis-TI:
As discussed in #1467 (Changes to Support IDEs)

These problems are intertwined and share a number of common traits and are being addressed together in a single commit, sadly I don't see a way to break this up.

**PROBLEM 1**   
In src/core/Makefile.am - lists numerous IF/ELSE/ENDIF which choose which files are added or not added to a specific library. While IDEs can provide some level of exclusion, they are not that flexible.

**PROBLEM 2**  
Need both --enable-cli-ftd --enable-cli-mtd, and likewise --enable-mtd

Need granularity such that CLI can be built with MTD only, And NCP can be built using MTD or FTD
Current ./configure solution is: FTD for both, or FTD for Neither

(there are no details for this item below)

**PROBLEM 3**
Linking the CLI application with the CLI library, and NCP with the NCP library

Currently when the CLI (or NCP) library is built - it is built using only *one* of the two internal header configurations, ie: FTD - with all of the structure/class elements, or  MTD with various structure/class elements removed.

This causes problems with link time optimization and static code analysis.

**PROBLEM 4**
Appvayor.

I do not have access to the version of Visual Studio used here, I have Visual Studio 2015, but the project appears to use an older version of Visual Studio, my experience has been:  VS2015 will upgrade the projects... this is a one-way process  

if someone can advise, it would be helpful.

DETAILS FOLLOW

================================================

**PROBLEM 1 - details**

a)  Makefile.am files should not longer use If/Else/Endif type constructs to add/remove files from a build.
Instead, the makefile should generally compile everything.

For example some IDEs - such as Eclipse CDT based IDEs - compile all files "*.c" and "*.cpp" and are of the view - if the file does not belong, or should not be compiled then it should not be present in the directory.  While other IDEs have other schemes, in IDE land - there does not exist a single common way to exclude/include files based upon the build configuration.

**Solution:** Where required, appropriate #if/#else/#endif type statements need to be added to various source files so that the entire quote "translation unit" - effectively becomes a pseudo-empty translation unit.

A number of files did not have, or do not #include "openthread-config.h" - which is required to support the #if/#else/#endif described above, this was added to a number of files.

Result: A number of places additional tests must be applied to the conditional.

For example in several places, the #if SOMETHING becomes: #if SOMETHING && OPENTHREAD_FTD

**PROBLEM 2** Details

This is described above, not expanded here.

[...]

PR #6612, opened 2021-05-11
[build] enhance mtd source files

Irving-cl:
This PR remove some FTD-only source files from mtd build. I think 
it may help to differentiate source files that are only used by FTD.
This PR renames some source files by adding the suffix 'ftd' to 
indicate that they are only compiled for FTD. And we should follow 
this convention in the future. The `update-makefile.py` script is 
also updated so that common sources and ftd-only sources are updated 
separately for cmake build.


abtink:
I recall we had a similar sort of config related checks to include/exlude the src file in makefile (in OT earlier data 2017), but decided to move away from that model. I found the issue https://github.com/openthread/openthread/issues/1619 from back them. The main reason was to make it easier to port OT code to other build systems. A comment related to this was added in OT `Makefile.am` 😃:
https://github.com/openthread/openthread/blob/621769411a99febbe7f6187fd8298ca1e33d7b0e/src/core/Makefile.am#L71-L110


We need to weigh in the benefit of such a change versus the potential complexity it can add (for code maintenance and adding policy to include `ftd` in the file name). 

- What is the main goal for this? Is it to speed up the MTD build? If so, do know by how much?
- Is there any other advantage/benefits to separating the FTD files?

Irving-cl:
@abtink 

Make sense to me. I was thinking that: 1. We can avoid some code that are not intended for MTD built in MTD unexpected 2. Make it more clear to us about what source files are FTD only.

I agree with your point. This introduces extra complexity. Let's follow the current rule.

PR #6621, opened 2021-05-13
[netdata] add `TlvIterator` and `FindSubTlv<TlvType>()`

abtink:
This commit contains the following changes:

- It adds `TlvIterator` to help iterate over TLVs of a given type
  in a given sequence of TLVs or sub-TLVs of another TLV.
- It adds helper method `PrefixTlv::FindSubTlv<TlvType>()` to
  search within the sub-TLVs for a specific one.
- It moves the `static` helper methods which find a TLV (with a
  given type) to `NetworkDataTlv` (from `NetworkData`).


src/core/thread/network_data_tlvs.cpp
```diff
@@ -0,0 +1,120 @@
...
+namespace NetworkData {
+
+//---------------------------------------------------------------------------------------------------------------------
+// NetworkDataTlv
+
+const NetworkDataTlv *NetworkDataTlv::Find(const NetworkDataTlv *aStart, const NetworkDataTlv *aEnd, Type aType)
+{
+    const NetworkDataTlv *tlv;
+
+    for (tlv = aStart; (tlv + 1 <= aEnd) && (tlv->GetNext() <= aEnd); tlv = tlv->GetNext())
```
Irving-cl, line 46:
```suggestion
    for (tlv = aStart; (tlv < aEnd) && (tlv->GetNext() <= aEnd); tlv = tlv->GetNext())
```

Here `tlv + 1` moves the address of `tlv` forward by 4 bytes (pointer size). What's the point of comparing that address with `aEnd`?

Irving-cl (commented):
Looks great 👍 

I have some small questions.

src/core/thread/network_data_tlvs.hpp
```diff
@@ -209,6 +209,138 @@ class NetworkDataTlv
...
+     * This static method searches in a given sequence of TLVs to find the first TLV with a given type.
+     *
+     * @param[in]  aStart  A pointer to the start of the sequence of TLVs to search within.
+     * @param[in]  aEnd    A pointer to the end of the sequence of TLVs.
+     * @param[in]  aType   The TLV type to find.
+     *
+     * @returns A pointer to the TLV if found, or nullptr if not found.
+     *
+     */
+    static NetworkDataTlv *Find(NetworkDataTlv *aStart, NetworkDataTlv *aEnd, Type aType)
```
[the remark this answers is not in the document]
abtink: I think you mean using like this
```
static NetworkDataTlv *Find(const NetworkDataTlv *aStart, const NetworkDataTlv *aEnd, ...)
```
If so, I don't think we can use this. Note that `const NetworkDataTlv *aStart` means that the pointed object is `const` and not the `aStart` pointer itself (i.e., `NetworkDataTlv` object being pointed to cannot be changed).

This flavor of this method wants to return a non-const `NetworkDataTlv *` (which is found from  the sequence of TLVs we get from `aStart` up to `aEnd`). If the original `aStart` and `aEnd` pointers are specifying the sequence of TLVs as `const NetworkData *` then we cannot safely get and return a non-const pointer from the sequence.

src/core/thread/network_data_tlvs.cpp
abtink: Note that this is just an existing implementation (just moved from `NetworkData` class to `NetworkDataTlv` class)

Also the `(tlv + 1 <= aEnd)` is very much intentional. We use this pattern in a lot of places when dealing with `Tlv`s:

- `tlv + 1` increases the pointer by the size of its type, which `NetworkDataTlv`. 
- `NetworkDataTlv` stuct contains two `uint8_t` member vairables: `mType` and `mLength`.
- So `tlv + 1` points to end of `NetworkDataTlv` struct (or two bytes after the `tlv` address).
- The  `(tlv + 1 <= aEnd)` checks that the type and length fields are before `aEnd` pointer.
- This check is important before the `tlv->GetNext()` call which itself will access `mLength` field (otherwise we may read beyond `aEnd` pointer from `tlv->GetNext()` call which is not generally safe and also results in warnings/errors by fuzzer and/or other analyzer tools).

Irving-cl: Oh I made a mistake here. Thanks.

Irving-cl (approved):
LGTM 👍

src/core/thread/network_data_tlvs.hpp
```diff
@@ -1298,6 +1513,97 @@ class ServerTlv : public NetworkDataTlv
...
+     *
+     */
+    TlvIterator(const NetworkDataTlv *aStart, const NetworkDataTlv *aEnd)
+        : mStart(aStart)
+        , mEnd(aEnd)
+    {
+    }
+
+    /**
+     * This constructor initializes the `TlvIterator` iterate over TLVs from a given buffer.
```
jwhui, line 1537:
```suggestion
     * This constructor initializes the `TlvIterator` to iterate over TLVs from a given buffer.
```

jwhui (commented):
Looks great overall. One minor nit.

jwhui (approved):

PR #6626, opened 2021-05-14
[udp] forbid udp duplicate open

simonlingoogle:
This commit forbid opening a UDP socket twice.

### DIFF
--- a/include/openthread/udp.h
+++ b/include/openthread/udp.h
@@ -161,6 +161,17 @@ otMessage *otUdpNewMessage(otInstance *aInstance, const otMessageSettings *aSett
 otError otUdpOpen(otInstance *aInstance, otUdpSocket *aSocket, otUdpReceive aCallback, void *aContext);
 
 /**
+ * Check if a UDP socket is open.
+ *
+ * @param[in]  aInstance  A pointer to an OpenThread instance.
+ * @param[in]  aSocket    A pointer to a UDP socket structure.
+ *
+ * @returns Whether the UDP socket is open.
+ *
+ */
+bool otUdpIsOpen(otInstance *aInstance, const otUdpSocket *aSocket);
+
+/**
  * Close a UDP/IPv6 socket.
  *
  * @param[in]  aInstance  A pointer to an OpenThread instance.
--- a/src/cli/cli_udp.cpp
+++ b/src/cli/cli_udp.cpp
@@ -109,7 +109,13 @@ otError UdpExample::ProcessOpen(uint8_t aArgsLength, Arg aArgs[])
     OT_UNUSED_VARIABLE(aArgsLength);
     OT_UNUSED_VARIABLE(aArgs);
 
-    return otUdpOpen(mInterpreter.mInstance, &mSocket, HandleUdpReceive, this);
+    otError error;
+
+    VerifyOrExit(!otUdpIsOpen(mInterpreter.mInstance, &mSocket), error = OT_ERROR_ALREADY);
+    error = otUdpOpen(mInterpreter.mInstance, &mSocket, HandleUdpReceive, this);
+
+exit:
+    return error;
 }
 
 otError UdpExample::ProcessSend(uint8_t aArgsLength, Arg aArgs[])
--- a/src/core/net/udp6.cpp
+++ b/src/core/net/udp6.cpp
@@ -87,6 +87,11 @@ Error Udp::Socket::Open(otUdpReceive aHandler, void *aContext)
     return Get<Udp>().Open(*this, aHandler, aContext);
 }
 
+bool Udp::Socket::IsOpen(void) const
+{
+    return Get<Udp>().IsOpen(*this);
+}
+
 Error Udp::Socket::Bind(const SockAddr &aSockAddr, otNetifIdentifier aNetifIdentifier)
 {
     return Get<Udp>().Bind(*this, aSockAddr, aNetifIdentifier);
@@ -185,6 +190,8 @@ exit:
 Error Udp::Open(SocketHandle &aSocket, otUdpReceive aHandler, void *aContext)
 {
     Error error = kErrorNone;
+
+    OT_ASSERT(!IsOpen(aSocket));
 
     aSocket.GetSockName().Clear();
     aSocket.GetPeerName().Clear();
@@ -307,6 +314,8 @@ exit:
 Error Udp::Close(SocketHandle &aSocket)
 {
     Error error = kErrorNone;
+
+    VerifyOrExit(IsOpen(aSocket));
 
 #if OPENTHREAD_CONFIG_PLATFORM_UDP_ENABLE
     error = otPlatUdpClose(&aSocket);
--- a/src/core/net/udp6.hpp
+++ b/src/core/net/udp6.hpp
@@ -172,6 +172,14 @@ public:
         Error Open(otUdpReceive aHandler, void *aContext);
 
         /**
+         * This method returns if the UDP socket is open.
+         *
+         * @returns If the UDP socket is open.
+         *
+         */
+        bool IsOpen(void) const;
+
+        /**
          * This method binds the UDP socket.
          *
          * @param[in]  aSockAddr            A reference to the socket address.
@@ -445,6 +453,16 @@ public:
      *
      */
     Error Open(SocketHandle &aSocket, otUdpReceive aHandler, void *aContext);
+
+    /**
+     * This method returns if a UDP socket is open.
+     *
+     * @param[in]  aSocket   A reference to the socket.
+     *
+     * @returns If the UDP socket is open.
+     *
+     */
+    bool IsOpen(const SocketHandle &aSocket) const { return mSockets.Contains(aSocket); }
 
     /**
      * This method binds a UDP socket.
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (128)
+#define OPENTHREAD_API_VERSION (129)
 
 /**
  * @addtogroup api-instance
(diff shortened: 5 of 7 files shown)


Irving-cl (approved):
LGTM 👍

abtink (approved):
LGTM. 👍

src/core/net/udp6.cpp
```diff
@@ -186,6 +186,8 @@ Error Udp::Open(SocketHandle &aSocket, otUdpReceive aHandler, void *aContext)
 {
     Error error = kErrorNone;
 
+    VerifyOrExit(!mSockets.Contains(aSocket), error = kErrorAlready);
```
bukepo, line 189: Is crashing a better way to tell users there's a logic error? I think open a socket twice is logic error.
```suggestion
    OT_ASSERT(aHandler != nullptr);
    OT_ASSERT(aSocket.mHandler == nullptr);
```

simonlingoogle: I am not very sure. Should we crash on duplicate UDP open? 
@abtink @jwhui

wgtdkp: I think we should not use assert here. Opening a socket twice is a mistake of the user but not logic error. Logic errors are those that should never happen.

1. assertion should never be used on user interface because we can not have assumption on how user will use the API.
2. assertion is not enough to capture user errors because it is removed in release build by default.
3. If the user are very sure that the socket can never be opened already (e.g. the port is reserved), he can add assertion in user code.

simonlingoogle: I prefer using assert slightly. Opening a socket twice is a logic error in user app. Using assert to catch user logic error helps users to reveal bugs in their program as early as possible.

abtink: I think all are fine (no strong preference either way).

I lean a bit towards @wgtdkp comment that assert may not be useful for public APIs. We can choose to ignore it as well. 

Opening same socket is similar to adding same address again, or  brining interface up again or starting MLE `otThreadSetEnabled(instance, true)` again. We return error in first case, check for it and then ignore it in second case, and just ignore it (start again) in third case.

src/core/net/udp6.hpp
```diff
@@ -442,6 +442,7 @@ class Udp : public InstanceLocator, private NonCopyable
      *
      * @retval kErrorNone     Successfully opened the socket.
      * @retval kErrorFailed   Failed to open the socket.
+     * @retval kErrorAlready  The socket is already opened.
```
simonlingoogle, line 445:
```suggestion
```

include/openthread/udp.h
```diff
@@ -154,8 +154,9 @@ otMessage *otUdpNewMessage(otInstance *aInstance, const otMessageSettings *aSett
  * @param[in]  aCallback  A pointer to the application callback function.
  * @param[in]  aContext   A pointer to application-specific context.
  *
- * @retval OT_ERROR_NONE    Successfully opened the socket.
- * @retval OT_ERROR_FAILED  Failed to open the socket.
+ * @retval OT_ERROR_NONE        Successfully opened the socket.
+ * @retval OT_ERROR_FAILED      Failed to open the socket.
+ * @retval OT_ERROR_ALREADY     The socket is already opened.
```
simonlingoogle, line 159:
```suggestion
 * @retval OT_ERROR_NONE    Successfully opened the socket.
* @retval OT_ERROR_FAILED Failed to open the socket.
```

src/core/net/udp6.cpp
bukepo: I see there is a difference between `otUdpOpen` and `otThreadSetEnabled`. The former would result in unexpected behavior if user ignores the already error, because `otUdpOpen` takes a callback which may be different from the last call. While `otThreadSetEnabled` has consistent behavior. I still prefer treat opening twice as a logic error in user's application. Thoughts?

bukepo (approved):
LGTM:+1:

src/core/net/udp6.cpp
simonlingoogle: > I lean a bit towards @wgtdkp comment that assert may not be useful for public APIs. We can choose to ignore it as well.

We actually used `OT_ASSERT` a lot in API implementations to validate arguments. Using `OT_ASSERT` to make sure users are calling APIs in a valid way has benefits:
- Avoids returning `otError` for APIs that does not actually need. This also help reduce OT code size, and save users from checking `otError` return values. 
- Help users catch bugs earlier.

> Opening same socket is similar to adding same address again, or brining interface up again or starting MLE `otThreadSetEnabled(instance, true)` again. We return error in first case, check for it and then ignore it in second case, and just ignore it (start again) in third case.

Agree with @bukepo that `otThreadSetEnabled` should be reentrant while `otUdpOpen` takes a callback that may be different from previous call. 

For `brining interface up again`, do you mean `otIp6SetEnabled` ?

simonlingoogle: One good news is that using `assert` in Udp::Open has already revealed a bug in DHCP6 server that opens the UDP socket for multiple times, which will cause DHCP6 server to crash if PLATFORM_UDP is used.

@bukepo @wgtdkp @abtink

abtink: In general, I am ok with `OT_ASSERT()` for this.
I mentioned the 3 ways I noticed we already handle similar situation. I agree that doing nothing in UDP open may not be the best option.

> We actually used OT_ASSERT a lot in API implementations to validate arguments. Using OT_ASSERT to make sure users are calling APIs in a valid way has benefits:

A side note here: I am aware that we use `OT_ASSERT()` in some of `api` implementations to check for pointer input parameters not being `nullptr. It is done in some of the APIs and not everywhere.  I actually favor removing such asserts in the code and instead adding documentation in the API where a pointer can be null or must not be null. Main reason is to reduce code size and also being consistent everywhere.

@simonlingoogle do you happen to know examples of us using `OT_ASSERT()` to check an API paramters beyond pointer check (something that is logic check). I am just curious.

simonlingoogle: > I actually favor removing such asserts in the code and instead adding documentation in the API where a pointer can be null or must not be null.

Uisng asserts and adding documentation are not mutually exclusive? We can have both. But I thinking documentation may not be that necessary since it's common for users to assume a poiner argument must not be `nullptr` unless it's documented to allow `nullptr`. 

>  Main reason is to reduce code size and also being consistent everywhere.

Users can always turn asserts off if code size is a major issue for them. The asserts can still help catch bugs in debug/eng builds, so I would prefer to keep asserts.

abtink: > Users can always turn asserts off if code size is a major issue for them. The asserts can still help catch bugs in debug/eng builds, so I would prefer to keep asserts.

This can be done on release build. However, one of the times where code size becomes important is during product development (integration of OT into projects/products) where user may want to have a development/debug build where  asserts are enabled (and maybe even detailed logging) and still want to fit such a build on a resource-limited device (e.g., to use for in-field testing). 

Asserts are useful but it is important to be selective about where/how we use them. For example, if we add an assert on every public OT API to check if the parameter `aInstance` is not null, it increases code size with little practical benefit.

simonlingoogle: There seems to be some different opinions on using asserts in OT APIs. 
I think this deserve another comprehensive conversation. 
For now, this PR continues to use asserts to validates UDP open.
Conversation closed for now.

wgtdkp (approved):
LGTM :+1:

src/core/net/udp6.cpp
jwhui: Using `OT_ASSERT` here will also mean that it is possible to trigger the assert via CLI commands. It might be frustrating for a user that typed `udp open` twice. At the same time, the fuzzer will trip on the `OT_ASSERT`. Thoughts?

simonlingoogle: > Using `OT_ASSERT` here will also mean that it is possible to trigger the assert via CLI commands. It might be frustrating for a user that typed `udp open` twice. At the same time, the fuzzer will trip on the `OT_ASSERT`. Thoughts?

I think this is a bug of CLI implementation, i.e. CLI should not call `otUdpOpen` twice even when user enters `udp open` twice. This crashes in the original code too because of asserts in platform code. 

I think we need to fix this bug in this PR(labeled as WIP).

simonlingoogle: Noted that code size increased by 96B due to this change. I planned to submit another PR that allows `otUdpOpen` to return `OT_ERROR_ALREADY` to compare the code size, but chose not to due to many code like below in OT core:
```
    SuccessOrExit(error = mSocket.Open(&Mle::HandleUdpReceive, this));
    SuccessOrExit(error = mSocket.Bind(kUdpPort));
```

Such code seems to assume that `Udp::Open` never returns `OT_ERROR_ALREADY`, otherwise these code need to handle it. 

Also, same as how this PR revealed a bug in dhcp6 server, asserts in `Udp::Open` can also help other core code to catch duplicate UDP open.

src/cli/cli_udp.cpp
```diff
@@ -109,7 +109,13 @@ otError UdpExample::ProcessOpen(uint8_t aArgsLength, Arg aArgs[])
     OT_UNUSED_VARIABLE(aArgsLength);
     OT_UNUSED_VARIABLE(aArgs);
 
-    return otUdpOpen(mInterpreter.mInstance, &mSocket, HandleUdpReceive, this);
+    otError error;
+
+    VerifyOrExit(mSocket.mHandler == nullptr, error = OT_ERROR_ALREADY);
```
jwhui, line 114: Should we provide a public API like `otUdpIsOpen()`?

src/core/net/udp6.cpp
```diff
@@ -87,6 +87,11 @@ Error Udp::Socket::Open(otUdpReceive aHandler, void *aContext)
     return Get<Udp>().Open(*this, aHandler, aContext);
 }
 
+bool Udp::Socket::IsOpened(void) const
```
jwhui, line 90: nit: `IsOpen()` seems more correct from an English language perspective.
```suggestion
bool Udp::Socket::IsOpen(void) const
```

simonlingoogle:
@jwhui Comments are resolved now. PTAL again.

jwhui (approved):

PR #6628, opened 2021-05-14
[child-supervision] remove dummy implementation of child supervision

Irving-cl:
This PR removes dummy implementation of `ChildSupervisor` and
`SupervisionListener`. And add macro to ensure that these data
members only exist in `ThreadNetif` when child supervisor option is
enabled. In particular, `ChildSupervisor` only exists when child
supervisor option is enabled and it is FTD.


abtink (approved):
LGTM. 👍 

The original idea behind the empty code was to avoid adding `#if` checks in the code but it's not that many places.

jwhui (approved):

PR #6629, opened 2021-05-14
[border-router] border routing optional at runtime

bukepo:
This commit allows launching an OpenThread daemon which was built with
backbone-router and border-routing support but disabling them at runtime
by not specifying the `-B` option.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -145,6 +145,7 @@ LOCAL_EXPORT_C_INCLUDE_DIRS     := \
 
 LOCAL_CPPFLAGS                                                              := \
     -std=c++11                                                                 \
+    -Wno-error=non-virtual-dtor                                                \
     -pedantic-errors                                                           \
     $(NULL)
 
@@ -355,6 +356,7 @@ LOCAL_SRC_FILES                                                  := \
     src/posix/platform/hdlc_interface.cpp                           \
     src/posix/platform/infra_if.cpp                                 \
     src/posix/platform/logging.cpp                                  \
+    src/posix/platform/mainloop.cpp                                 \
     src/posix/platform/memory.cpp                                   \
     src/posix/platform/misc.cpp                                     \
     src/posix/platform/multicast_routing.cpp                        \
--- a/src/posix/platform/Makefile.am
+++ b/src/posix/platform/Makefile.am
@@ -51,6 +51,7 @@ libopenthread_posix_a_SOURCES             = \
     hdlc_interface.cpp                      \
     infra_if.cpp                            \
     logging.cpp                             \
+    mainloop.cpp                            \
     memory.cpp                              \
     misc.cpp                                \
     multicast_routing.cpp                   \
@@ -72,6 +73,7 @@ libopenthread_posix_a_LIBADD                                                    
 
 noinst_HEADERS                            = \
     hdlc_interface.hpp                      \
+    mainloop.hpp                            \
     multicast_routing.hpp                   \
     openthread-posix-config.h               \
     platform-posix.h                        \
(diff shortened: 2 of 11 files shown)


wgtdkp:
The `Routing Manager` is also using the `-B` option, please disable the `Routing Manager` as well in this case.

src/posix/platform/Makefile.am
```diff
@@ -51,6 +51,7 @@ libopenthread_posix_a_SOURCES             = \
     hdlc_interface.cpp                      \
     infra_if.cpp                            \
     logging.cpp                             \
+    mainloop.cpp                            \
```
Irving-cl, line 54: Run `script/update-makefiles.py` to add it for gn build file?

Android.mk
```diff
@@ -145,6 +145,7 @@ LOCAL_EXPORT_C_INCLUDE_DIRS     := \
 
 LOCAL_CPPFLAGS                                                              := \
     -std=c++11                                                                 \
+    -Wno-error=non-virtual-dtor                                                \
```
Irving-cl, line 148: We'd better not suppress the warnings. Is there a way to get rid of the warning? Not very sure why such warning would occur in this PR, can we avoid it by declaring a virtual destructor?

src/posix/platform/Makefile.am
bukepo: gn doesn't use posix platform.

Android.mk
bukepo: This is only required when `delete` is used on a class. And there's another flag `-Werror=delete-non-virtual-dtor` enables warning on this, which is already included in `-Wall`.

Irving-cl (approved):
LGTM :+1:

src/posix/platform/mainloop.hpp
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * This method processes the mainloop events.
+     *
+     * @param[in]   aContext    A reference to the mainloop context.
+     *
+     */
+    virtual void Process(const otSysMainloopContext &aContext) = 0;
+
+private:
+    Source *mNext;
```
simonlingoogle, line 69:
```suggestion
    Source *mNext = nullptr;
```

src/posix/platform/mainloop.hpp
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * This function returns the Mainloop singleton.
+     *
+     * @returns A refernce to the Mainloop singleton.
+     *
+     */
+    static Manager &Get(void);
+
+private:
+    Source *mSources = nullptr;
```
simonlingoogle, line 120: nit: Use std::list?

simonlingoogle (approved):
LGTM :+1: 

with nits

src/posix/platform/mainloop.cpp
```diff
@@ -0,0 +1,100 @@
...
+    mSources      = &aSource;
+}
+
+void Manager::Remove(Source &aSource)
+{
+    assert(mSources != nullptr);
+
+    if (&aSource == mSources)
+    {
+        mSources = mSources->mNext;
```
simonlingoogle, line 53:
```suggestion
        mSources = mSources->mNext;
        aSource.mNext = nullptr;
```
nit

src/posix/platform/mainloop.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    {
+        Source *source = mSources;
+
+        do
+        {
+            if (source->mNext == &aSource)
+            {
+                source->mNext = aSource.mNext;
+                aSource.mNext = nullptr;
```
simonlingoogle, line 65:
```suggestion
```
nit: will set to nullptr at exit.

src/posix/platform/mainloop.hpp
bukepo: using `std::list` will involves dynamic memory, virtual destructor, and overhead of the template. I think it's not necessary for now because the possible sources are known at compile time.

wgtdkp (approved):
LGTM :+1:

jwhui (approved):

PR #6636, opened 2021-05-16
[ip6] add public APIs to convert an IPv6 address or prefix to string

abtink:
This commit adds new public OpenThread APIs to convert an IPv6 address
or an IPv6 prefix to a human-readable string representation. The
implementation is shared with the `Address/Prefix::ToString()` used by
core modules (through adding a private `ToString(StringWriter &)`
method which can then be used with a `String<kSize>` or a char
buffer).

### DIFF
--- a/include/openthread/ip6.h
+++ b/include/openthread/ip6.h
@@ -562,7 +562,7 @@ const uint16_t *otIp6GetUnsecurePorts(otInstance *aInstance, uint8_t *aNumEntrie
 bool otIp6IsAddressEqual(const otIp6Address *aFirst, const otIp6Address *aSecond);
 
 /**
- * Convert a human-readable IPv6 address string into a binary representation.
+ * This function converts a human-readable IPv6 address string into a binary representation.
  *
  * @param[in]   aString   A pointer to a NULL-terminated string.
  * @param[out]  aAddress  A pointer to an IPv6 address.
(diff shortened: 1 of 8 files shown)


Irving-cl (approved):
LGTM :+1:

wgtdkp (approved):
LGTM :+1: 

@abtink what about replacing the implementation in https://github.com/openthread/openthread/blob/main/src/posix/platform/ip6_utils.hpp with your change?

abtink:
> @abtink what about replacing the implementation in https://github.com/openthread/openthread/blob/main/src/posix/platform/ip6_utils.hpp with your change?

Thanks. Good suggestion. I was not aware of this util file. Will change it in the next push. 👍

src/posix/platform/ip6_utils.hpp
```diff
@@ -45,13 +45,10 @@ class Ip6AddressString
...
      *
-     * @param[in]   aAddress    A pointer to a buffer holding an IPv6 address.
+     * @param[in]   aAddress    A pointer to an IPv6 address (MUST NOT be nullptr).
      *
      */
-    Ip6AddressString(const void *aAddress)
-    {
-        VerifyOrDie(inet_ntop(AF_INET6, aAddress, mBuffer, sizeof(mBuffer)) != nullptr, OT_EXIT_ERROR_ERRNO);
-    }
+    Ip6AddressString(const otIp6Address *aAddress) { otIp6AddressToString(aAddress, mBuffer, sizeof(mBuffer)); }
```
bukepo, line 51: intended to use void* to accept both otIp6Address and POSIX IPv6 address structures. Note inet_ntop's prototype is void*. I prefer keeping it as is. Thoughts?

abtink: We don't seem to use `void *` currently from anywhere, but I don't think it matters.  Will change it back to `void *`

abtink: Actually thinking more on this, I think it's better to keep the posix as it was,
- The `inet_ntop()` is available on posix and also it can give shorter (optimized/compressed format) with `::` .
- We don't really care about code size on posix platform, so little reason to use OT API. 

I will change the `ip6_utils.hpp` back to what it was.

bukepo (approved):
LGTM:+1:

jwhui (approved):

include/openthread/ip6.h
```diff
@@ -573,6 +573,39 @@ bool otIp6IsAddressEqual(const otIp6Address *aFirst, const otIp6Address *aSecond
...
+ *
+ * @param[in]  aAddress  A pointer to an IPv6 address (MUST NOT be NULL).
+ * @param[out] aBuffer   A pointer to a char array to output the string (MUST NOT be NULL).
+ * @param[in]  aSize     The size of @p aBuffer (in bytes). Recommended to use `OT_IP6_ADDRESS_STRING_SIZE`.
+ *
+ */
+void otIp6AddressToString(const otIp6Address *aAddress, char *aBuffer, uint16_t aSize);
+
+#define OT_IP6_PREFIX_STRING_SIZE 45 ///< Recommended size for string representation of an IPv6 prefix.
+/**
```
jwhui, line 594:
```suggestion
#define OT_IP6_PREFIX_STRING_SIZE 45 ///< Recommended size for string representation of an IPv6 prefix.

/**
```

src/core/common/string.hpp
```diff
@@ -131,12 +131,19 @@ class StringWriter
...
+     *
      */
     StringWriter(char *aBuffer, uint16_t aSize);
 
     /**
      * This constructor initializes the object as cleared on a fixed-size string.
      *
+     * @tparam kSize  The size of the string (number of chars).
+     *
+     * @param[in] aStringBuffer   A reference to a `String<kSize` to write into.
```
jwhui, line 145:
```suggestion
     * @param[in] aStringBuffer   A reference to a `String<kSize>` to write into.
```

abtink:
Thanks @jwhui. Just noticed you already added the changes and merged it. 😃

PR #6638, opened 2021-05-17
[dataset] add check for security policy rotation time

Irving-cl:


bukepo:
Update tests/scripts/expect/cli-dataset.exp to cover the change?

Irving-cl:
@bukepo Updated.

src/cli/cli_dataset.cpp
```diff
@@ -821,6 +821,7 @@ Error Dataset::ParseSecurityPolicy(otSecurityPolicy &aSecurityPolicy, uint8_t aA
 
     memset(&policy, 0, sizeof(policy));
     SuccessOrExit(error = aArgs[0].ParseAsUint16(policy.mRotationTime));
+    VerifyOrExit(policy.mRotationTime != 0, error = OT_ERROR_INVALID_ARGS);
```
jwhui, line 824: We already have methods to check the validity of an Operational Dataset.
https://github.com/openthread/openthread/blob/d3b1a0dac73f644b8c983c8d25bd675a6d80d534/src/core/meshcop/meshcop_tlvs.cpp#L43
We should be consistent about where and when we validate these values.

Irving-cl: Shall we convert the policy to a TLV and then call `IsValid`?

jwhui: We have a number of other parameters that also require validation at some point. Currently, we only validate the dataset before we apply the configuration. We could probably validate the dataset whenever we set the active or pending operational dataset. I don't know if we should also validate each individual parameter as it is being set within the dataset itself. While that it would be ideal, it seems to further add code size overhead and duplicate checks that would be done later.

Thoughts? Curious to know what others think as well.

abtink: I agree. I think it would be better to do the validation in the related core methods (and not in CLI/NCP modules). 

This way the code is shared/common. I think it also fits better in the core in the sense that the CLI or NCP should ideally just act as a mirror of OT APIs and should not know/decide what param values are valid for each API.

Irving-cl: I totally agree that the CLI/NCP should not know if a value is valid. The dataset case is a little special: the `sDataset` is stored in `cli` instead of the core. In most cases, for example, `SetNetworkName`, the code goes to OT core and tell if the name is valid and return to the API. However, when setting some field of the dataset through `cli`, the code stays in `cli`.

I think the current defect is that the errors in dataset parameters may be hidden from the users. For example, if I set an invalid channel as follow:
```
> dataset init new
[INFO]-CLI-----: execute command: dataset init new
Done
> dataset commit active
[INFO]-CLI-----: execute command: dataset commit active
[INFO]-MESH-CP-: Active dataset set
[DEBG]-MAC-----: RadioPanId: 0x9e3a
Done
> [INFO]-CORE----: Notifier: StateChanged (0x101fc100) [KeySeqCntr Channel PanId NetName ExtPanId MstrKey PSKc SecPolicy ActDset]

> dataset channel 1
[INFO]-CLI-----: execute command: dataset channel 1
Done
> dataset commit active
[INFO]-CLI-----: execute command: dataset commit active
[INFO]-MESH-CP-: Active dataset set
Done
> [INFO]-CORE----: Notifier: StateChanged (0x10000000) [ActDset]
```
Users are not notified that the setting actually fails.

I think:
1. We can add error handling for `otDatasetSetActive/Pending`. Currently the error returned by `ApplyConfiguration` is ignored in `DatasetManager::Restore` (https://github.com/openthread/openthread/blob/main/src/core/meshcop/dataset_manager.cpp#L103).
2. With 1. , it's a little difficult for us to know which parameter is invalid since we only got one error. Shall we put the `sDataset` into core? And we add value checks in core functions that are called by OT API.

Thoughts?

abtink: What do you think of the following model?

- We can add a public API to validate a given `otOperationalDataset` instance, i.e. check all params are within good range, etc.
- We keep `sDataset` in CLI and whenever we populate a new field (say channel), we call the new OT API to validate it and if there is an error we report it to user.
- In CLI code, we can remember the previous value of the field, say the channel, (if there is any) and if the validate API call with the new value fails we set it back to what it was. This way the `sDataset` should remain valid (i.e. if we try to set channel to something invalid, we get an error and channel in `sDataset` stays as what is was before).


Thoughts?

jwhui: Sounds good to me - I was actually thinking the same.

PR #6641, opened 2021-05-18
[ip6][udp]improve frame filtering

MarekPorwisz:
Second proposal for the solution  originating from PR #6545 - one to be selected and merged

If OpenThread is part of an OS with it's own IP stack filtering only
control messages is insufficien.
Example could be that if application is using OpenThread API to listen
for CoAP messages received messages are passed also to the IP callback.
The message then is delivered to the OS IP stack and generates error
response: "ICMPv6 Destination Unreachable (Port unreachable)" as the
port is not open on the OS side. OS IP stack is unaware of the
OpenThread existence so it can not be filtered on that level and the
shim layer does not receive port, or protocol information from OT so it
can not filter out a frame without parsing it first.
Extended filtering also to ports originating from OpenThread.

### DIFF
--- a/src/core/net/ip6.cpp
+++ b/src/core/net/ip6.cpp
@@ -49,6 +49,7 @@
 #include "net/ip6_filter.hpp"
 #include "net/netif.hpp"
 #include "net/udp6.hpp"
+#include "openthread/ip6.h"
 #include "thread/mle.hpp"
 
 using IcmpType = ot::Ip6::Icmp::Header::Type;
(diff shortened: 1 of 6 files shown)


src/core/net/ip6.cpp
```diff
@@ -1035,8 +1036,8 @@ Error Ip6::ProcessReceiveCallback(Message &          aMessage,
             Udp::Header udp;
 
             IgnoreError(aMessage.Read(aMessage.GetOffset(), udp));
-            VerifyOrExit(Get<Udp>().ShouldUsePlatformUdp(udp.GetDestinationPort()), error = kErrorNoRoute);
-
+            VerifyOrExit(Get<Udp>().ShouldUsePlatformUdp(udp.GetDestinationPort())
+                         && !Get<Udp>().IsPortInUse(udp.GetDestinationPort()), error = kErrorNoRoute);
```
simonlingoogle, line 1040: Here the new code blocks the message to be passed to the TUN interface (i.e. wpan0), so we need to pass it to the socket `Udp::HandlePayload` in `Udp::HandleMessage`.

Unfortunately, when `PLATFORM_UDP` is used, the following code will block the message:

https://github.com/openthread/openthread/blob/713231e144c9ccd46126323fb14d2ab06cdb6361/src/core/net/udp6.cpp#L479-L481

I think we need to allow UDP messages for OpenThread in-use ports be passed to `Udp::HandlePayload`.

@bukepo Do you have any other thoughts on this issue?

PR #6644, opened 2021-05-19
[posix] split readline and stdio handler

bukepo:
This commit splits readline and stdio CLI handlers in POSIX app to
better demonstrate how to work with the CLI module.

### DIFF
--- a/src/posix/cli_stdio.cpp
+++ b/src/posix/cli_stdio.cpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2019, The OpenThread Authors.
+ *  Copyright (c) 2021, The OpenThread Authors.
  *  All rights reserved.
  *
  *  Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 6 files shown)


src/posix/cli_stdio.cpp
```diff
@@ -0,0 +1,113 @@
...
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "platform/openthread-posix-config.h"
+
+#if !OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE
+
+#ifndef HAVE_LIBEDIT
+#define HAVE_LIBEDIT 0
```
jwhui, line 34: Don't these evaluate to 0 when not defined?

jwhui (approved):
Thanks! 👍🏼

PR #6645, opened 2021-05-19
[radio] use stack reset in RCP

jinran-google:
This commit makes RCP use a stack reset instead of doing a hard reset when a reset command is received.

### DIFF
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -258,6 +258,8 @@ void otInstanceFactoryReset(otInstance *aInstance);
 /**
  * This method resets the internal states of the OpenThread radio stack. Callbacks and configurations are preserved.
  *
+ * This API is only available under radio builds (`OPENTHREAD_RADIO = 1`).
+ *
  * @param[in]  aInstance A pointer to an OpenThread instance.
  */
 void otInstanceResetRadioStack(otInstance *aInstance);
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (166)
+#define OPENTHREAD_API_VERSION (167)
 
 /**
  * @addtogroup api-instance
@@ -254,6 +254,13 @@ void otInstanceReset(otInstance *aInstance);
  * @param[in]  aInstance A pointer to an OpenThread instance.
  */
 void otInstanceFactoryReset(otInstance *aInstance);
+
+/**
+ * This method resets the internal states of the OpenThread radio stack. Callbacks and configurations are preserved.
+ *
+ * @param[in]  aInstance A pointer to an OpenThread instance.
+ */
+void otInstanceResetRadioStack(otInstance *aInstance);
 
 /**
  * This function erases all the OpenThread persistent info (network settings) stored on non-volatile memory.
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -244,23 +244,28 @@ void otRemoveStateChangeCallback(otInstance *aInstance, otStateChangedCallback a
  * The reset process ensures that all the OpenThread state/info (stored in volatile memory) is erased. Note that the
  * `otPlatformReset` does not erase any persistent state/info saved in non-volatile memory.
  *
- * @param[in]  aInstance A pointer to an OpenThread instance.
+ * @param[in]  aInstance  A pointer to an OpenThread instance.
+ *
  */
 void otInstanceReset(otInstance *aInstance);
 
 /**
  * This method deletes all the settings stored on non-volatile memory, and then triggers platform reset.
  *
- * @param[in]  aInstance A pointer to an OpenThread instance.
+ * @param[in]  aInstance  A pointer to an OpenThread instance.
+ *
  */
 void otInstanceFactoryReset(otInstance *aInstance);
 
 /**
- * This method resets the internal states of the OpenThread radio stack. Callbacks and configurations are preserved.
+ * This method resets the internal states of the OpenThread radio stack.
+ *
+ * Callbacks and configurations are preserved.
  *
  * This API is only available under radio builds (`OPENTHREAD_RADIO = 1`).
  *
- * @param[in]  aInstance A pointer to an OpenThread instance.
+ * @param[in]  aInstance  A pointer to an OpenThread instance.
+ *
  */
 void otInstanceResetRadioStack(otInstance *aInstance);
 
--- a/src/core/common/instance.cpp
+++ b/src/core/common/instance.cpp
@@ -173,6 +173,14 @@ void Instance::Reset(void)
     otPlatReset(this);
 }
 
+#if OPENTHREAD_RADIO
+void Instance::ResetRadioStack(void)
+{
+    mRadio.Init();
+    mLinkRaw.Init();
+}
+#endif
+
 void Instance::AfterInit(void)
 {
     mIsInitialized = true;
--- a/src/core/common/instance.hpp
+++ b/src/core/common/instance.hpp
@@ -194,6 +194,14 @@ public:
      */
     void Reset(void);
 
+#if OPENTHREAD_RADIO
+    /**
+     * This method resets the internal states of the radio.
+     *
+     */
+    void ResetRadioStack(void);
+#endif
+
     /**
      * This method returns the active log level.
      *
--- a/src/core/mac/link_raw.cpp
+++ b/src/core/mac/link_raw.cpp
@@ -51,17 +51,27 @@ namespace Mac {
 
 LinkRaw::LinkRaw(Instance &aInstance)
     : InstanceLocator(aInstance)
-    , mReceiveChannel(OPENTHREAD_CONFIG_DEFAULT_CHANNEL)
-    , mPanId(kPanIdBroadcast)
-    , mReceiveDoneCallback(nullptr)
-    , mTransmitDoneCallback(nullptr)
-    , mEnergyScanDoneCallback(nullptr)
 #if OPENTHREAD_RADIO
     , mSubMac(aInstance)
 #elif OPENTHREAD_CONFIG_LINK_RAW_ENABLE
     , mSubMac(aInstance.Get<SubMac>())
 #endif
 {
+    Init();
+}
+
+void LinkRaw::Init(void)
+{
+    mEnergyScanDoneCallback = nullptr;
+    mTransmitDoneCallback   = nullptr;
+    mReceiveDoneCallback    = nullptr;
+
+    mReceiveChannel      = OPENTHREAD_CONFIG_DEFAULT_CHANNEL;
+    mPanId               = kPanIdBroadcast;
+    mReceiveDoneCallback = nullptr;
+#if OPENTHREAD_RADIO
+    mSubMac.Init();
+#endif
 }
 
 Error LinkRaw::SetReceiveDone(otLinkRawReceiveDone aCallback)
--- a/src/core/mac/sub_mac.cpp
+++ b/src/core/mac/sub_mac.cpp
@@ -52,32 +52,48 @@ namespace Mac {
 SubMac::SubMac(Instance &aInstance)
     : InstanceLocator(aInstance)
     , mRadioCaps(Get<Radio>().GetCaps())
-    , mState(kStateDisabled)
-    , mCsmaBackoffs(0)
-    , mTransmitRetries(0)
-    , mShortAddress(kShortAddrInvalid)
-    , mRxOnWhenBackoff(true)
-    , mEnergyScanMaxRssi(kInvalidRssiValue)
-    , mEnergyScanEndTime(0)
     , mTransmitFrame(Get<Radio>().GetTransmitBuffer())
     , mCallbacks(aInstance)
     , mPcapCallback(nullptr)
     , mPcapCallbackContext(nullptr)
-    , mFrameCounter(0)
-    , mKeyId(0)
     , mTimer(aInstance, SubMac::HandleTimer)
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
-    , mCslPeriod(0)
-    , mCslChannel(0)
-    , mIsCslChannelSpecified(false)
-    , mCslLastSync(0)
     , mCslParentAccuracy(kCslWorstCrystalPpm)
     , mCslParentUncert(kCslWorstUncertainty)
-    , mCslState(kCslIdle)
     , mCslTimer(aInstance, SubMac::HandleCslTimer)
 #endif
 {
+    Init();
+}
+
+void SubMac::Init(void)
+{
+    mState           = kStateDisabled;
+    mCsmaBackoffs    = 0;
+    mTransmitRetries = 0;
+    mShortAddress    = kShortAddrInvalid;
     mExtAddress.Clear();
+    mRxOnWhenBackoff   = true;
+    mEnergyScanMaxRssi = kInvalidRssiValue;
+    mEnergyScanEndTime = Time{0};
+
+    mPrevKey.Clear();
+    mCurrKey.Clear();
+    mNextKey.Clear();
+
+    mFrameCounter = 0;
+    mKeyId        = 0;
+    mTimer.Stop();
+
+#if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
+    mCslPeriod             = 0;
+    mCslChannel            = 0;
+    mIsCslChannelSpecified = false;
+    mCslSampleTime         = TimeMicro{0};
+    mCslLastSync           = TimeMicro{0};
+    mCslState              = kCslIdle;
+    mCslTimer.Stop();
+#endif
 }
 
 otRadioCaps SubMac::GetCaps(void) const
--- a/src/core/mac/sub_mac.hpp
+++ b/src/core/mac/sub_mac.hpp
@@ -81,6 +81,10 @@ namespace Mac {
 #error "OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE is required for OPENTHREAD_CONFIG_MAC_CSL_DEBUG_ENABLE."
 #endif
 
+#if OPENTHREAD_RADIO || OPENTHREAD_CONFIG_LINK_RAW_ENABLE
+class LinkRaw;
+#endif
+
 /**
  * This class implements the IEEE 802.15.4 MAC (sub-MAC).
  *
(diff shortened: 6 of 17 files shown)


src/core/mac/link_raw.cpp
```diff
@@ -64,6 +64,21 @@ LinkRaw::LinkRaw(Instance &aInstance)
 {
 }
 
+void LinkRaw::Reset(void)
```
zhanglongxia, line 67: Could we call the constructor to reset the class to avoid manually initialize all variables . Otherwise, when we add a new variable to this class, we have to initialize the variable in two functions.
```
void LinkRaw::Reset(Instance &aInstance)
{
    new (this) LinkRaw(aInstance);
}
```

src/core/common/instance.hpp
```diff
@@ -190,6 +190,14 @@ class Instance : public otInstance, private NonCopyable
      */
     void Reset(void);
 
+#if OPENTHREAD_RADIO
+    /**
+     * This method resets the internal states of the instance.
+     *
+     */
+    void StackReset(void);
```
zhanglongxia, line 198: This function only reset the RCP stack rather than the all OpenThread stack. The function name confuses code readers. How about changing the name to `RcpReset` or something.

src/core/mac/link_raw.cpp
jinran-google: Some variables will be intentionally preserved.

In fact `mReceiveDoneCallback` should be preserved, but current codes use it as an indicator of whether the raw link layer is enabled, so I reset it as a workaround.

src/core/common/instance.hpp
jinran-google: We are planning to change it to a full stack reset in the future. But I'm OK with changing the name for now.

src/core/mac/link_raw.cpp
jinran-google: Updated. We can explicitly preserve variables when needed.

src/core/common/instance.hpp
jinran-google: Updated.

src/core/common/instance.cpp
```diff
@@ -167,6 +167,14 @@ void Instance::Reset(void)
     otPlatReset(this);
 }
 
+#if OPENTHREAD_RADIO
+void Instance::RadioStackReset(void)
+{
+    mRadio.Reset();
+    mLinkRaw.Reset();
```
brianljt, line 174: need to call SubMac reset here?

jinran-google: Updated. `SubMac::Init` is called from `LinkRaw::Init`.

src/ncp/ncp_base.cpp
```diff
@@ -1212,6 +1212,19 @@ otError NcpBase::CommandHandler_RESET(uint8_t aHeader)
 
     otError error = OT_ERROR_NONE;
 
+#if OPENTHREAD_RADIO
+    mInstance->RadioStackReset();
```
simonlingoogle, line 1216: One advantage of original RESET is that it is able to recover from software errors. 
What if a platform still want to use `otInstanceReset` for `RESET` command?

Should we add a optional argument to `RESET` command to specify the reset kind?

@bukepo

jinran-google: > Can we add an optional argument to RESET command to specify the reset kind?

That will break the compatibility between RCP and host. I think we have 2 approaches:
- Add a build flag to choose reset behavior
- Leave the current `RESET` command untouched and add a new `STACK_RESET` command

simonlingoogle: > > Can we add an optional argument to RESET command to specify the reset kind?
> 
> That will break the compatibility between RCP and host. I think we have 2 approaches:

Not, it won't as long as its an optional argument. I think it's fine to add more fields to a Spinel message structure. 

> * Add a build flag to choose reset behavior
> * Leave the current `RESET` command untouched and add a new `STACK_RESET` command

That works for me also.

jinran-google: Optional argument looks good. Updated.

src/core/mac/link_raw.cpp
```diff
@@ -51,17 +51,27 @@ namespace Mac {
...
+
+void LinkRaw::Init(void)
+{
+    mEnergyScanDoneCallback = nullptr;
+    mTransmitDoneCallback   = nullptr;
+    mReceiveDoneCallback    = nullptr;
+
+    mReceiveChannel      = OPENTHREAD_CONFIG_DEFAULT_CHANNEL;
+    mPanId               = kPanIdBroadcast;
+    mReceiveDoneCallback = nullptr;
```
zhanglongxia, line 71: When the `Instance::RadioStackReset()` calls the `mLinkRaw.Init()` , why we need to reset the callback pointers `mReceiveDoneCallback`? Do we expect the up layer to set the callback pointer again after the the `RadioStackReset()` is called?

src/core/radio/radio.cpp
```diff
@@ -33,6 +33,58 @@
...
+    SetPanId(Mac::kPanIdBroadcast);
+    SetExtendedAddress(extAddr);
+    SetShortAddress(kDefaultShortAddress);
+    SetMacKey(0, 0, prevKey, currKey, nextKey);
+    SetMacFrameCounter(0);
+
+    error = SetTransmitPower(0);
+    OT_ASSERT(error == OT_ERROR_NONE);
+
+    error = SetCcaEnergyDetectThreshold(0);
```
zhanglongxia, line 81: Generally, the platform layer defines a default `CcaEnergyDetectThreshold` value. The OpenThread host may not set the  `CcaEnergyDetectThreshold`, then the platform can directly use the platform defined default value. If we set the `CcaEnergyDetectThreshold` value here,  the value CcaEnergyDetectThreshold `0` will be used by the platform.  This may cause issues. The line 78 also has this problem.

src/core/radio/radio.cpp
```diff
@@ -33,6 +33,58 @@
 
 namespace ot {
 
+namespace {
+#if OPENTHREAD_CONFIG_RADIO_LINK_IEEE_802_15_4_ENABLE
+constexpr Mac::ShortAddress kDefaultShortAddress = 0xfffe;
```
zhanglongxia, line 38: Can we directly use the constant `Mac::kShortAddrInvalid `？

src/core/mac/link_raw.cpp
jinran-google: Because `LinkRaw` is using the callback to indicate if it's enabled. I think it's better to use a separate bool to store the enabled state but we can do it in a separate PR.

Currently the upper layer will set the callback regardless of what the lower layer does.

src/core/radio/radio.cpp
```diff
@@ -33,6 +33,58 @@
...
+    {
+        error = Sleep();
+        OT_ASSERT(error == OT_ERROR_NONE);
+
+        error = Disable();
+        OT_ASSERT(error == OT_ERROR_NONE);
+    }
+
+    SetPanId(Mac::kPanIdBroadcast);
+    SetExtendedAddress(extAddr);
```
bukepo, line 73:
```suggestion
    SetExtendedAddress(Mac::ExtAddress{});
```

src/core/radio/radio.cpp
```diff
@@ -33,6 +33,58 @@
...
+    SetShortAddress(kDefaultShortAddress);
+    SetMacKey(0, 0, prevKey, currKey, nextKey);
```
bukepo, line 75:
```suggestion
    SetMacKey(0, 0, Mac::Key{}, Mac::Key{}, Mac::Key{});
```

src/core/mac/link_raw.cpp
```diff
@@ -51,17 +51,27 @@ namespace Mac {
...
 #endif
 {
+    Init();
+}
+
+void LinkRaw::Init(void)
+{
+    mEnergyScanDoneCallback = nullptr;
+    mTransmitDoneCallback   = nullptr;
+    mReceiveDoneCallback    = nullptr;
```
bukepo, line 67: Is it necessary to reset these pointers?

brianljt (approved):
LGTM. Thanks!

src/core/radio/radio.cpp
jinran-google: Removed.

src/core/radio/radio.cpp
jinran-google: Done.

src/core/radio/radio.cpp
jinran-google: Done.

src/core/radio/radio.cpp
jinran-google: Done.

src/core/mac/link_raw.cpp
jinran-google: `mEnergyScanDoneCallback` and `mTransmitDoneCallback` are temporary callbacks, they are set every time when scans or transmits.

`mReceiveDoneCallback` is used as the indicator of whether the raw link layer is enabled, it's necessary to reset to disabled state.

bukepo (approved):
LGTM:+1:

zhanglongxia (approved):
LGTM

simonlingoogle (approved):
LGTM :+1:

src/ncp/ncp_base.cpp
```diff
@@ -1210,22 +1210,47 @@ otError NcpBase::CommandHandler_RESET(uint8_t aHeader)
...
+    }
 
-#if OPENTHREAD_MTD || OPENTHREAD_FTD
-    // We only get to this point if the
-    // platform doesn't support resetting.
-    // In such a case we fake it.
+#if OPENTHREAD_RADIO
+    if (reset_type == SPINEL_RESET_STACK)
+    {
+        mInstance->RadioStackReset();
```
abtink, line 1225: I suggest adding a public API for this (probably in `instance.h`), so that we don't need to call method on `Instance` from `NcpBase`

src/core/common/instance.cpp
```diff
@@ -167,6 +167,14 @@ void Instance::Reset(void)
     otPlatReset(this);
 }
 
+#if OPENTHREAD_RADIO
+void Instance::RadioStackReset(void)
```
abtink, line 171: How about `ResetRadioStack()` insetad? 
(though not required by our style guide, we often start a method name with the "verb").

src/core/mac/sub_mac.cpp
```diff
@@ -52,35 +52,48 @@ namespace Mac {
...
-    , mState(kStateDisabled)
-    , mCsmaBackoffs(0)
-    , mTransmitRetries(0)
-    , mShortAddress(kShortAddrInvalid)
-    , mRxOnWhenBackoff(true)
-    , mEnergyScanMaxRssi(kInvalidRssiValue)
-    , mEnergyScanEndTime(0)
     , mTransmitFrame(Get<Radio>().GetTransmitBuffer())
     , mCallbacks(aInstance)
     , mPcapCallback(nullptr)
```
abtink, line 57: Why exlcuding the `mPcapCallback` and its context from `Init()`?

src/core/mac/sub_mac.cpp
```diff
@@ -52,35 +52,48 @@ namespace Mac {
...
-    , mFrameCounter(0)
-    , mKeyId(0)
     , mTimer(aInstance, SubMac::HandleTimer)
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
-    , mCslPeriod(0)
-    , mCslChannel(0)
-    , mIsCslChannelSpecified(false)
-    , mCslLastSync(0)
     , mCslParentDrift(kCslWorstCrystalPpm)
     , mCslParentUncert(kCslWorstUncertainty)
```
abtink, line 62: Same question for these two?

abtink (commented):
Thanks @jinran-google.

Sorry for my later review of this. Overall looks good, some questions/suggestios below.

One general question: What is the background/reason for this? 
- In which situation do we want to just reset the radio and not the RCP as whole?

jinran-google:
> Thanks @jinran-google.
> 
> Sorry for my later review of this. Overall looks good, some questions/suggestios below.
> 
> One general question: What is the background/reason for this?
> 
> * In which situation do we want to just reset the radio and not the RCP as whole?

Thanks for your comments, @abtink .

The main reason is that if multiple radios are running on the same chip (e.g. bluetooth), we probably don't want the whole chip to be reset when the Thread host starts, resetting only the needed part by software is more graceful. This increases complexity though.

src/core/mac/sub_mac.cpp
jinran-google: Callbacks are used by higher level software to receive events, and don't change the internal behavior of the stack (except for `mReceiveDoneCallback`). So I think resetting the stack shouldn't require the caller to set callbacks again.

src/core/mac/sub_mac.cpp
jinran-google: These are configuration values. Resetting the stack doesn't include resetting configurations.

src/ncp/ncp_base.cpp
jinran-google: Done.

src/core/common/instance.cpp
jinran-google: Done.

abtink (commented):
> The main reason is that if multiple radios are running on the same chip (e.g. bluetooth), we probably don't want the whole chip to be reset when the Thread host starts, resetting only the needed part by software is more graceful. This increases complexity though.

Thanks @jinran-google. It makes sense. Overall the code LGTM. I just added few smaller suggestions below.

But thinking more about this, I have an idea/suggestion for a rather different overall approach which I'd like to bring up.

- I understand the main goal/reason that we don't want to reset the entire chip if there are multiple radio (BLE, etc) on it when Thread starts.
- The solution in this PR is to add mechnasim to just reset the radio.
- But we are leaving all other platform modules that OT may use and the `otInstance` itself to continue to run as before (e.g. `Alarm/Timer` platform, or UART, etc.).  I guess this is fine and works for RCP (for now).

What we really want is that when we get a reset command (from host) to not reset the whole chip. We really want to have all the platform modules reset and then reinit the `otInstance` (clear all RAM variable) and start running `otInstance` fresh.

I think we can realize this differently:
- From `Instance::Reset()` we call `otPlatReset()` which we currently assume would reset the whole chip.
- We can change the implementation of `otPlatReset()` in platform layer to not trigger chip reset, but instead it can reset all the plaform modules (Radio, alaram, etc). 
- In OT code, we change the `Instance::Reset()` such that after calling `otPlatReset()` we call `Finalize()` immediately followed by `InitSingle()` call again (to reinit the `otInstance`). 
  - If the `otPlatReset()` does actually reset the chip, then we won't get to the `Finalize()` and `InitSingle()` callls (which is fine).
  - If the `otPlatReset()` resets all the platform components/modules, then we are reinitializing the `otInstance` to start fresh (which clears all RAM variables). 

I think this can work. The main advantages I see is that it resets all components without resetting the chip. It delegates the reseting of platform modules to the platform implementation itself which can reset all platform modules (radio, timer, any other platform module we may use in future)... It avoids the need for us to add mechanism in OT To handle reset of one platform components (like reseting radio only - which can be tricky to do).

Thoughts?

jinran-google:
> But thinking more about this, I have an idea/suggestion for a rather different overall approach which I'd like to bring up.

Because there were many bugs introduced by this kind of pseudo reset, we didn't take the finalize and re-initialize approach. There are many variables outside of `otInstance` and defined as static, they often cause issues.

So instead of resetting the instance, we define a set of attributes that should be reset (e.g. channel, src match), and explicitly check them in tests. When developing other components we can keep in mind that only these things will be reset. This should help us avoid issues in the future. Thoughts?

src/core/mac/sub_mac.hpp
```diff
@@ -206,6 +206,12 @@ class SubMac : public InstanceLocator, private NonCopyable
      */
     explicit SubMac(Instance &aInstance);
 
+    /**
+     * This method initializes the states of the sub-MAC layer.
+     *
+     */
+    void Init(void);
```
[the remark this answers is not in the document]
jinran-google: Done.

include/openthread/instance.h
```diff
@@ -255,6 +255,13 @@ void otInstanceReset(otInstance *aInstance);
  */
 void otInstanceFactoryReset(otInstance *aInstance);
 
+/**
```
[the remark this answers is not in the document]
jinran-google: Done.

src/core/radio/radio.cpp
```diff
@@ -33,6 +33,41 @@
 
 namespace ot {
 
+void Radio::Init(void)
```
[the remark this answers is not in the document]
jinran-google: Done. Added to both the declaration and the definition.

abtink:
> Because there were many bugs introduced by this kind of pseudo reset, we didn't take the finalize and re-initialize approach. There are many variables outside of otInstance and defined as static, they often cause issues.
> 
> So instead of resetting the instance, we define a set of attributes that should be reset (e.g. channel, src match), and explicitly check them in tests. When developing other components we can keep in mind that only these things will be reset. This should help us avoid issues in the future. Thoughts?

I think "pesdue-reset" as we have in the code was added quickly (more of a workaorund), so it may not handle all situation (which led to the bugs). 

I would argue that a proper platform reset implementation would ensure to clear all necessary platform specific static variables during reset. It would make sense for the platform to actually decide and handle this since it knows what parameters (static variables) it is saving for each platform module implementation and how it uses them. Having a clear platform reset and then re-initing the `otInstance` object (for it to also start fresh with no previous saved info) sounds like much safer and cleaner long term model.

jinran-google:
> I think "pesdue-reset" as we have in the code was added quickly (more of a workaorund), so it may not handle all situation (which led to the bugs).
> 
> I would argue that a proper platform reset implementation would ensure to clear all necessary platform specific static variables during reset. It would make sense for the platform to actually decide and handle this since it knows what parameters (static variables) it is saving for each platform module implementation and how it uses them. Having a clear platform reset and then re-initing the otInstance object (for it to also start fresh with no previous saved info) sounds like much safer and cleaner long term model.

I agree that a proper implementation should handle all situations. That's why we want a clear definition of what needs to be reset and what doesn't. I think there are some advantages to doing a stack reset in the core:
1. It makes the behavior consistent and independent of hardware. This makes it possible to do CI tests.
2. We can have a standardized procedure of reset, and make it into the spec in the long term.
3. We think that the hard reset (including pseudo-reset) should only be used for testing purposes. The RCP reset is part of the host boot process, so it shouldn't depend on a hard reset.

abtink (approved):
LGTM. Thanks. 👍 

Just quick update: @jinran-google and I chatted offline about different reset model. I think the radio reset model (the code in this PR) looks good. We can later indepedenly consider adding different platform reset model (only resetting the platform modules and re-initing the `otInstance`) if there is a need/use-case for it.

jwhui (approved):

PR #6646, opened 2021-05-19
[radio] remove `OT_RADIO_CHANNEL_PAGE_MAX` and address gcc warnings

schinchani:
In this PR, I have updated the OpenThread stack to support a Max Channel Page value of 255 when working with proprietary radio configurations. With OPENTHREAD_CONFIG_PLATFORM_RADIO_PROPRIETARY_CHANNEL_PAGE configuration option available for the platform to modify, restricting its max value to 2, does not seem correct.

Secondly, I have also modified the (<=) usage in two ‘if’ conditions, as these operations as written today (with constants), are resulting in the compile time warnings (and cmake errors) specifically when working with proprietary radio configurations and setting Channel Min to 0.
Following was an error observed when building the OT BR app with proprietary channel min set to 0.
```
../../third_party/openthread/repo/src/core/radio/radio.hpp:638:52: error: comparison is always true due to limited range of data type [-Werror=type-limits]
         return (aCslChannel == 0) || ((kChannelMin <= aCslChannel) && (aCslChannel <= kChannelMax));
                                        ~~~~~~~~~~~~^~~~~~~~~~~~~~
compilation terminated due to -Wfatal-errors
```

These errors do not show up for the non-proprietary cases is because the minimum channel values used in these cases are 1 (915MHz) and 11 (2.4GHz).

It appears like there may be more than one way to address this ( for example, using a local variable, typecasting to an int). Tweaking the condition, seemed like a better pick.


include/openthread/platform/radio.h
```diff
@@ -88,7 +88,11 @@ enum
     OT_RADIO_CHANNEL_PAGE_0_MASK = (1U << OT_RADIO_CHANNEL_PAGE_0), ///< 2.4 GHz IEEE 802.15.4-2006
     OT_RADIO_CHANNEL_PAGE_2      = 2,                               ///< 915 MHz IEEE 802.15.4-2006
     OT_RADIO_CHANNEL_PAGE_2_MASK = (1U << OT_RADIO_CHANNEL_PAGE_2), ///< 915 MHz IEEE 802.15.4-2006
-    OT_RADIO_CHANNEL_PAGE_MAX    = OT_RADIO_CHANNEL_PAGE_2,         ///< Maximum supported channel page value
+#if OPENTHREAD_CONFIG_PLATFORM_RADIO_PROPRIETARY_SUPPORT
+    OT_RADIO_CHANNEL_PAGE_MAX = 0xFF,
```
jwhui, line 92: Is setting to 255 problematic for this code:
https://github.com/openthread/openthread/blob/8b63efe2aed519647c8173b5317b5331b8ba59ff/src/core/radio/radio.hpp#L105
?

I wonder if we should just remove the `OT_RADIO_CHANNEL_PAGE_MAX`. I'm not sure what value it provides.

schinchani: Hi @jwhui , I used 255 as the spec currently provides an entire byte for this field.  However, I agree that setting the channel page to any value greater than 31 is going to be problematic for `kSupportedChannelPages`. While one option could be to remove OT_RADIO_CHANNEL_PAGE_MAX, I am wondering if we should instead keep it set to 31 (instead of 255) for the proprietary case. Thoughts?

schinchani: A third option could be - to remove OT_RADIO_CHANNEL_PAGE_MAX, and add a compile-time check to make sure that the proprietary channel page value (set by platform) is not greater than 31.

jwhui: > A third option could be - to remove OT_RADIO_CHANNEL_PAGE_MAX, and add a compile-time check to make sure that the proprietary channel page value (set by platform) is not greater than 31.

Let's go with this option, if that's OK with you.

schinchani: Done. Thanks.

jwhui (approved):

PR #6649, opened 2021-05-20
[srp-server] fix removing expired services

wgtdkp:
The SRP server tries to notify the advertising proxy when a service is expired.
The expired service is firstly removed from the service list and later added back
to the head of the service list if the service name should be retailed. This will result
in deadloop when there are two or more services on the same host.

This commit fixes this issue by keeping the expired service in the service list by simply
mark the service `deleted` and pass all services of the host to the advertising proxy. The
advertising proxy will republish those services that are not expired but this should be fine
since service expiration is not a common case.

Also update the advertising proxy test case to cover multiple services.


simonlingoogle (approved):
LGTM :+1:

abtink (approved):
LGTM. 👍 
Thanks for quick fix.

jwhui (approved):

PR #6491, opened 2021-04-20
Proposal for a TCP API for OpenThread

samkumar:
At the TCP API developer meeting (see Issue #6411), there was broad support for designing a new TCP API in the spirit of OpenThread's existing APIs that is geared toward the RAM-constrained devices that OpenThread targets. This pull request constitutes a proposal for such a TCP API.

I met with @simonlingoogle, @jwhui, @abtink, and @StuartCheshire today to learn about their ideas for a TCP API and receive their feedback on this proposal. The TODO list below describes feedback items I intend to incorporate into this API but haven't yet. Here are my slides from that meeting, which show the intended usage of this API and explain the rationale behind this API: [openthread_tcp_api_design.pdf](https://github.com/openthread/openthread/files/6340616/openthread_tcp_api_design.pdf). That said, I do provide a brief summary of the key design points below.

## Brief Summary of Design
The first important aspect of the design is to avoid dynamic memory allocation in the system. Rather, the application allocates memory and hands it to the system for TCP.

We have two kinds of sockets. *Active sockets* correspond to connection endpoints; they are used to send or receive data over a TCP connections. *Passive sockets* are used to listen for incoming TCP connections. If you've used the Go programming language, then an active socket corresponds to `net.Conn` and a passive socket corresponds to `net.Listener`. Passive sockets use significantly less memory than active sockets, so using separate structs for active sockets and passive sockets allow applications to allocate less memory when they need a passive socket.

The APIs strive to support zero-copy I/O. Zero-copy I/O is useful in this setting because it reduces RAM usage; one doesn't have to allocate memory into which to copy data. That said, it's unclear whether we want to use zero-copy I/O in all cases, especially given the livelock concerns I write about at the end of the slides. Nevertheless, the API should be *compatible* with zero-copy I/O, even if the implementation isn't zero-copy to start with, as this will give us the most flexibility in the implementation (e.g., to reduce copying of data in the future). It's the need for a zero-copy API that motivates the `otLinkedBuffer` type.

## TODO List
Here are some TODOs from today's meeting that I intend to implement over the next few days:
* Add an API function to "extend" the final `otLinkedBuffer` in the send buffer by a specified number of bytes. This will make it easier for the application to do many repeated small writes (e.g., adding many separate writes one byte at a time).
[...]

Issue #6456, opened 2021-04-15
Developer Meeting: TCP Implementation


jwhui:
The OpenThread project will be hosting its next developer meeting on Thursday April 29th.

**Topic:** Transport Control Protocol (TCP) in OpenThread

**Time:** Thursday April 29th 08:00 PDT, 11:00 EDT, 15:00 UTC, 17:00 CEST, 23:00 CST.

**Agenda**

TCP Implementation

**Meeting Info**

If you would like to participate in this meeting, please send me your name and email address and I will add you to the calendar invite.

backkem:
Hi @jwhui, I'm wondering if this meeting is open to the public. I'm trying to broaden LAN support in browsers, see [Local devices API](https://discourse.wicg.io/t/idea-local-devices-api-lan-services). Enabling secure access to Thread devices from the browser would be a perfect use-case.

jwhui:
> Hi @jwhui, I'm wondering if this meeting is open to the public. I'm trying to broaden LAN support in browsers, see [Local devices API](https://discourse.wicg.io/t/idea-local-devices-api-lan-services). Enabling secure access to Thread devices from the browser would be a perfect use-case.

@backkem , yes, this meeting is open to anyone in the OpenThread community that would like to join. Please send me your name and email address and I will add you to the calendar invite.

backkem:
@jwhui Michiel De Backker; ![image](https://user-images.githubusercontent.com/38858977/116272807-7e81fd00-a781-11eb-837b-a2e6bcfd28b8.png)

samkumar:
Here are the slides that I presented at today's meeting: [openthread_tcp_design_implementation.pdf](https://github.com/openthread/openthread/files/6400253/openthread_tcp_design_implementation.pdf)

There was broad support for using TCPlp. As we discussed, I'll make a pull request with the entire TCP stack so that people can try out the code, but I'll also submit "bite-sized" pull requests so that the code can be reviewed and merged incrementally.

romacdon:
> Here are the slides that I presented at today's meeting: [openthread_tcp_design_implementation.pdf](https://github.com/openthread/openthread/files/6400253/openthread_tcp_design_implementation.pdf)
> 
> There was broad support for using TCPlp. As we discussed, I'll make a pull request with the entire TCP stack so that people can try out the code, but I'll also submit "bite-sized" pull requests so that the code can be reviewed and merged incrementally.

[...]

PR #6650, opened 2021-05-20
TCPlp for OpenThread

samkumar:
This pull request implements TCP for OpenThread using TCPlp. This follows up on the second OpenThread developer meeting (#6456), where there was support for implementing TCP in OpenThread using TCPlp.

Currently, support for the proposed API (Pull Request #6491) is far enough along to try out TCPlp. But it isn't complete. Some of the main missing features at the moment are:
* Support for SendByExtension (functionality is there, but not yet exposed in the API).
* Support for auto-bind. For now, always bind your socket before connecting to another host.
* Support for TCP Fast Open (might come in another pull request).
* Support for the Contiguify function (will come in another pull request).
That said, you can create TCP connections and exchange data, and there's CLI support for doing that. That's why I felt it was an appropriate time to create this pull request.

TCPlp is derived from the TCP stack in the FreeBSD Operating System. Files taken from FreeBSD retain their original copyright notices and licenses (which are all BSD-like licenses). New code written by me is provided under the BSD 3-Clause License. I'm happy to relicense the code I wrote if it would make it easier to merge in to OpenThread.


samkumar:
I'd like guidance from the OpenThread community on how they'd like to see TCPlp integrated into OpenThread. In the current version of this pull request, I have the FreeBSD-derived code, along with my re-implementations of buffering, as a `third_party` library in C, while the main TCP driver (derived from TCPlp code that I wrote) is in C++ in `tcp6.cpp`.

Additionally, in this version of TCPlp, I deleted any extraneous declarations in the FreeBSD header files that I didn't need. But I did keep the old, unused code (often commented out) in the source files, in case the reviewers wanted to see the changes I made to the FreeBSD code itself. But this makes the source files quite cluttered. I personally think I should clean the source files up by deleting the unused and commented out portions of the old FreeBSD code, but before I went ahead and did that, I wanted to confirm with the OpenThread community that they would like me to do that.

There's also the question of how to integrate TCPlp into OpenThread long term. Here are some possibilities I could think of:
1. Leave the TCPlp code in a third party module, but clean it up as described above.
2. Move the TCPlp code into OpenThread's C++ core, and wrap it in a namespace, but leave the variable and function naming conventions as is.
3. Move the TCPlp code into OpenThread's C++ core, and rename all functions and variables to conform with OpenThread's style (e.g., use camel case, all struct members start with `m`, all function arguments start with `a`, etc.).

I'm currently leaning toward the second option. But I want to get the OpenThread community's feedback to understand what they want. I'm happy to do the implementation work to support whichever option that OpenThread community wants.

abtink:
Thanks @samkumar for submitting this PR and bringing this in. This is great to see :)

> There's also the question of how to integrate TCPlp into OpenThread long term. Here are some possibilities I could think of:
> 1. Leave the TCPlp code in a third party module, but clean it up as described above.
> 2. Move the TCPlp code into OpenThread's C++ core, and wrap it in a namespace, but leave the variable and function naming conventions as is.
> 3. Move the TCPlp code into OpenThread's C++ core, and rename all functions and variables to conform with OpenThread's style (e.g., use camel case, all struct members start with `m`, all function arguments start with `a`, etc.).

I lean towards option 1  (which is similar to how `mebdtls` is currently integrated and used). I think would be good to keep core for new/original code for OT (under same license and copyright). Also I don't think we need to consider/change the style in third party source code. The style guide is mainly for original/new code in OT itself.

jwhui:
> There's also the question of how to integrate TCPlp into OpenThread long term. Here are some possibilities I could think of:
> 
> 1. Leave the TCPlp code in a third party module, but clean it up as described above.
> 2. Move the TCPlp code into OpenThread's C++ core, and wrap it in a namespace, but leave the variable and function naming conventions as is.
> 3. Move the TCPlp code into OpenThread's C++ core, and rename all functions and variables to conform with OpenThread's style (e.g., use camel case, all struct members start with `m`, all function arguments start with `a`, etc.).

Any code that carries a copyright/license other than the standard OpenThread copyright/license header needs to be placed in `/third_party`. This includes derivative work. Note that this is primarily an issue of source code placement.

samkumar:
Thanks @abtink and @jwhui for your thoughts. I should clarify that the primary intent behind Options 2 and 3 wasn't whether the code is in `third_party` or `src/core` --- it's about the extent to which we move the code toward using OpenThread's structures, conventions, and style instead of FreeBSD's. We can keep the code in `third_party`, but wrap it in a namespace, make the files C++ source instead of C source, and then use OpenThread's C++ structures and utilities, if we want.

I'm happy to go with the first option --- incidentally, that one probably requires the least effort for me to implement. That said, I want to explain why I'm leaning toward the second options, so that you can consider those reasons as well:
* **Less code duplication.** The BSD functions have their own implementations of TCP structures (e.g., `tcphdr`, `in6_addr` and `sockaddr_in6` structures), and utility functions (e.g., `htons`), which are repetitive in light of the structures provided by OpenThread. Some of the code to handle these structures is repeated, increasing the code size. Additionally, the interface between the C++ driver code in `tcp6.cpp` and the C code for TCPlp is a set of functions that add additional overhead for code space (there are additional layers of indirection for things like sending out packets, setting timers, etc.). This could probably be eliminated if we integrated TCPlp more fully into OpenThread. Integrating TCPlp more fully into OpenThread would help eliminate some of this duplication.
* **Less complexity.** Currently, there's a bit of complexity and pointer hacking involved in moving between the C++ state (e.g., the contents of `otTcpActiveSocket`) and C state (e.g., the contents of `struct tcpcb`). If we integrated TCPlp more fully into OpenThread, we could just put all of the TCP state into a single structure, eliminating this complexity.
* **More natural.** This is a subjective argument, but TCPlp doesn't stand alone without an IP stack. Unlike `mbedtls`, which as far as I can tell, is just cloned from an external repository, I made modifications to TCPlp for it to work with OpenThread. That's why it took me a couple of weeks to make this pull request. :) Additionally, TCPlp isn't just a library OpenThread is using; TCPlp makes calls back into OpenThread, for example. That might be an argument for doing a fuller integration. Finally, some of the C utilities that TCPlp uses (circular buffer, bitmaps, linked buffer manipulation) would benefit from being written in OpenThread C++ so that other parts of OpenThread could benefit from those data structures.
* **More portable.** OpenThread currently has built-in support for cross-toolchain compatibility, to robustly detect things like endianness and to have macros for packed structs, etc. that work well across different toolchains. In contrast, FreeBSD (as far as I can tell) was designed for a single toolchain, and therefore might be less robust than OpenThread. For example, I already had to change the code to remove nonstandard types (e.g., `u_char`, `u_int`, etc.) for the code to build properly for the NRF52840.

To be clear, I'm totally on board with using the first option if that's what the OpenThread community wants, but I want to put forth these reasons for your consideration. From a purely practical standpoint, it might make the most sense to stick with Option 1 for now (to avoid delaying the pull request), and then move toward Option 2 or Option 3 in a future pull request if that's what the community wants. I can also try to integrate TCPlp more fully into OpenThread (e.g., modify parts of TCPlp to use OpenThread structures) while keeping it separate in C for now.

To clarify with respect to Jonathan's comments: I also don't think licensing will be an issue, since OpenThread is licensed with BSD 3-clause. (If I had chosen to use Linux's TCP stack, it might be a different story --- but thankfully I chose FreeBSD.) Some of the copyright notices in the FreeBSD code include FreeBSD contributors --- see the top of [tcp_input.c](https://github.com/openthread/openthread/blob/7c511aa953ca35c461fccb930ba94736430fd209/third_party/tcplp/bsdtcp/tcp_input.c)  --- and the BSD license seems to say that we have to keep those copyright notices.

I know @Abhayakara had some thoughts on this as well --- maybe he can chime in with his thoughts.

jwhui:
> Thanks @abtink and @jwhui for your thoughts. I should clarify that the primary intent behind Options 2 and 3 wasn't whether the code is in `third_party` or `src/core` --- it's about the extent to which we move the code toward using OpenThread's structures, conventions, and style instead of FreeBSD's. We can keep the code in `third_party`, but wrap it in a namespace, make the files C++ source instead of C source, and then use OpenThread's C++ structures and utilities, if we want.

@samkumar , thanks for clarifying. I had misunderstood your initial proposal. I agree that making the files C++ source has the benefits that you summarized.

Part of the tradeoff is how far we deviate from the original FreeBSD source. With mbedTLS, we try to use a pristine version to make integrating newer versions of mbedTLS and select bug fixes simple. If we view TCPlp as effectively maintained by the OpenThread project, then that may be a lesser concern.

> To be clear, I'm totally on board with using the first option if that's what the OpenThread community wants, but I want to put forth these reasons for your consideration. From a purely practical standpoint, it might make the most sense to stick with Option 1 for now (to avoid delaying the pull request), and then move toward Option 2 or Option 3 in a future pull request if that's what the community wants. I can also try to integrate TCPlp more fully into OpenThread (e.g., modify parts of TCPlp to use OpenThread structures) while keeping it separate in C for now.

It all depends on how long you think it will take to implement Option 2 relative to Option 1 and whether that fits within the desired timelines.

> To clarify with respect to Jonathan's comments: I also don't think licensing will be an issue, since OpenThread is licensed with BSD 3-clause. (If I had chosen to use Linux's TCP stack, it might be a different story --- but thankfully I chose FreeBSD.) Some of the copyright notices in the FreeBSD code include FreeBSD contributors --- see the top of [tcp_input.c](https://github.com/openthread/openthread/blob/7c511aa953ca35c461fccb930ba94736430fd209/third_party/tcplp/bsdtcp/tcp_input.c) --- and the BSD license seems to say that we have to keep those copyright notices.

I agree. I do not have any concern of including code derived from FreeBSD in the OpenThread project - we just need to locate that source in `/third_party`.

samkumar:
Just pushed an update implementing `SendByExtension` and rebasing onto the latest TCP API from #6491 (including changing variable names and internal type names to be consistent with that).

samkumar:
Just rebased this on top of the latest TCP API from #6491.

samkumar:
I just rebased this on top of `main`, including the change made in reviews of earlier pull requests.

samkumar:
For people who would like to try out this code, I've posted a video (https://youtu.be/ppZ784YUKlI) showing how to use TCP via the functionality I added to the CLI tool. It doesn't go over all of the functionality supported in the CLI, but hopefully it will help you get started.

jwhui:
> For people who would like to try out this code, I've posted a video (https://youtu.be/ppZ784YUKlI) showing how to use TCP via the functionality I added to the CLI tool. It doesn't go over all of the functionality supported in the CLI, but hopefully it will help you get started.

Great video, @samkumar ! Thanks for sharing.

samkumar:
The purpose of this pull request was to provide a way to try out TCPlp and discuss its integration. TCPlp was added into OpenThread over a series of nine pull requests (#6744, #6770, #6790, #6857, #6885, #6925, #6926, #7074, and #7190). Now that those pull requests have been merged, I am closing this pull request.

PR #6651, opened 2021-05-20
[static-analysis] address static code analysis issues

MarekPorwisz:
We have run the static code analyzer on the OpenThread code and some
issues were discovered.
This commit addresses those that were marked as High impact and were not
marked false positive after the inspection.

### DIFF
--- a/src/core/mac/data_poll_handler.cpp
+++ b/src/core/mac/data_poll_handler.cpp
@@ -272,12 +272,17 @@ void DataPollHandler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError, 
             {
                 uint32_t frameCounter;
                 uint8_t  keyId;
+                Error   error;
 
-                IgnoreError(aFrame.GetFrameCounter(frameCounter));
+                error = aFrame.GetFrameCounter(frameCounter);
+                OT_ASSERT(error == kErrorNone);
                 aChild.SetIndirectFrameCounter(frameCounter);
 
-                IgnoreError(aFrame.GetKeyId(keyId));
+                error = aFrame.GetKeyId(keyId);
+                OT_ASSERT(error == kErrorNone);
                 aChild.SetIndirectKeyId(keyId);
+
+                OT_UNUSED_VARIABLE(error);
             }
 
             ExitNow();
--- a/src/core/mac/data_poll_handler.cpp
+++ b/src/core/mac/data_poll_handler.cpp
@@ -272,7 +272,7 @@ void DataPollHandler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError, 
             {
                 uint32_t frameCounter;
                 uint8_t  keyId;
-                Error   error;
+                Error    error;
 
                 error = aFrame.GetFrameCounter(frameCounter);
                 OT_ASSERT(error == kErrorNone);
--- a/src/ncp/ncp_base_mtd.cpp
+++ b/src/ncp/ncp_base_mtd.cpp
@@ -911,7 +911,7 @@ template <> otError NcpBase::HandlePropertySet<SPINEL_PROP_THREAD_ASSISTING_PORT
     {
         uint16_t port;
 
-        IgnoreError(mDecoder.ReadUint16(port));
+        SuccessOrExit(error = mDecoder.ReadUint16(port));
         SuccessOrExit(error = otIp6AddUnsecurePort(mInstance, port));
     }
 


src/core/mac/data_poll_handler.cpp
```diff
@@ -273,10 +273,10 @@ void DataPollHandler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError,
                 uint32_t frameCounter;
                 uint8_t  keyId;
 
-                IgnoreError(aFrame.GetFrameCounter(frameCounter));
+                OT_ASSERT(aFrame.GetFrameCounter(frameCounter) == kErrorNone);
```
abtink, line 276: Use of `OT_ASSERT()` like this is not correct and won't work properly on all builds.
Note that `OT_ASSERT()` may be configured to compile out as NOOP (empty). Then we won't get the frame counter value.

src/core/mac/data_poll_handler.cpp
```diff
@@ -273,10 +273,10 @@ void DataPollHandler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError,
...
                 aChild.SetIndirectFrameCounter(frameCounter);
 
-                IgnoreError(aFrame.GetKeyId(keyId));
+                OT_ASSERT(aFrame.GetKeyId(keyId) == kErrorNone);
```
abtink, line 279: I wonder what problem was detected with the use of `IgnoreError()` here?

abtink (commented):
Thanks @MarekPorwisz.

I commented below. The use of `OT_ASSERT()` in the PR won't be correct in all config.

src/core/mac/data_poll_handler.cpp
MarekPorwisz: Both this and GetFrameCounter can fail on FindSecurityHeaderIndex(); leading to output parameter not being initialized.
If the frame would be created somewhere around this calls I would verify if it is properly set and mark as False Positive in our system if so. However if I am not mistaken we are in the callback for transmission being finished and we should not assume it was not malformed on the way e.g while being handled by the radio driver.

src/core/mac/data_poll_handler.cpp
MarekPorwisz: ok, I will update it to VerifyOrExit .
I realize that there is not much we can do in such case and originally I was going to mark it as "Intended", but I realized that leaving it as is may lead to sending some not-so-random data from the stack that may happen to be sensitive.

src/core/mac/data_poll_handler.cpp
MarekPorwisz: I reverted this change as VerifyOrExit around GetFrameCounter also covers the fail scenario for this call i.e. they will always both fail or both pass.

src/core/mac/data_poll_handler.cpp
jwhui: I think `OT_ASSERT` is appropriate because we want to ensure that transmitted frames are always formed properly. However, to @abtink 's point, we have to separate the logic from verification into separate lines:
```
error = aFrame.GetFrameCounter(frameCounter);
OT_ASSERT(error == kErrorNone);
```

src/core/mac/data_poll_handler.cpp
jwhui: Same as above.
```
error = aFrame.GetKeyId(keyId);
OT_ASSERT(error == kErrorNone);
```

src/core/mac/data_poll_handler.cpp
abtink: I agree with @jwhui and I think assert would be better here (helps detect any issue with platform implementation since we expect the tx done callback to give back the same frame as the one given by OT stack which should be well-formed). 

We just need to seperate the logic. Another thing to note, since `OT_ASSERT()` can become NOOP, and we are adding a new local `Error error` variable that is only used in the assert check, we need `OT_UNUSED_VARIABLE(error);` to silence the warning that `error` is unused in the configs where assert is disabled. You can check `RandomManager::RandomManager(void)` or `CoapSecure::SetPsk()` which use very similar pattern.

MarekPorwisz: Sorry. My bad, I did not realize that assert evaluates to nothing and didn't get you the first time. Will fix that on Monday.

abtink (approved):
LGTM. Thanks. 👍

src/core/mac/data_poll_handler.cpp
```diff
@@ -272,12 +272,17 @@ void DataPollHandler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError,
             {
                 uint32_t frameCounter;
                 uint8_t  keyId;
+                Error   error;
```
jwhui, line 275: To fix pretty check.
```suggestion
                Error    error;
```

jwhui (approved):
Thanks! 👍🏼

PR #6652, opened 2021-05-20
[github-actions] pin to clang-10 for coverage

jwhui:
To avoid llvm gcov version mismatch.


abtink (commented):
Thanks for quick fix 👍

PR #6657, opened 2021-05-21
[posix] refactor netif and netlink management

bukepo:
This commit merges duplicate code about network interface on POSIX
platform. This also enables infra_if building on macOS.


wgtdkp (commented):
LGTM :+1: Thanks!

wgtdkp (approved):

src/posix/platform/trel_udp6.cpp
```diff
@@ -141,60 +137,20 @@ static const char *BufferToString(const uint8_t *aBuffer, uint16_t aLength)
...
+        otIp6AddressInfo{aUnicastAddress, TREL_UNICAST_ADDRESS_PREFIX_LEN, TREL_UNICAST_ADDRESS_SCOPE, true});
 }
 
 static void RemoveUnicastAddress(const otIp6Address *aUnicastAddress)
 {
     otLogDebgPlat("[trel] RemoveUnicastAddress(%s)", Ip6AddrToString(aUnicastAddress));
-    UpdateUnicastAddress(aUnicastAddress, /* aToAdd */ false);
+    ot::Posix::NetifManager::Get().RemoveUnicast(
+        sInterfaceIndex,
+        otIp6AddressInfo{aUnicastAddress, TREL_UNICAST_ADDRESS_PREFIX_LEN, TREL_UNICAST_ADDRESS_SCOPE, true});
```
abtink, line 153: Thanks @bukepo  for this PR. I think refactoring the netlink and netif manager into their own module that is then used by different platform modules is indeed very good idea.

But there is one thing that makes be feel a bit uneasy about  change to the code in the trel posix platform code. Since https://github.com/openthread/openthread/issues/6472 we are no longer covering the trel posix behavior in any CI test and if anything gets broken we won't be detecting it.  I feel much better if we leave the `trel_udp6.cpp` as is for now, and change the code here later after some test coverage is added back. Thoughts?

PR #6658, opened 2021-05-21
[string] change `String<kSize>` to inherit from `StringWriter`

abtink:
This commit changes the `String<kSize>` class to inherit from
`StringWriter` class. It helps simplify their use allowing
`StringWriter` methods to be directly called on an `String` instance
while ensuring the implementation is still shared among all template
variants of `String<kSize>` class. This change also ensures that
`String<kSize>` is always initialized and set to an empty string from
its constructor.


bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #6672, opened 2021-05-23
[srp-client] support saving selected server info (by auto-start)

abtink:
This commit adds support for a new feature in SRP client which allows
it to save the selected server info (by the auto-start feature) in
non-volatile settings. On SRP client restart (e.g., due to a device
reset) the client will select the same server when searching to
discover and pick one from the Thread Network Data service entries.
Config `OPENTHREAD_CONFIG_SRP_CLIENT_SAVE_SELECTED_SERVER_ENABLE` can
be used to enable/disable this feature.

The server info is saved only after the host info is successfully
registered with the server and if it is selected by auto-start from a
network data SRP *unicast* service entry.

This commit also adds `test_srp_client_save_server_info.py` test-case
which verifies the behavior of the new feature.

### DIFF
--- a/tests/scripts/thread-cert/test_srp_client_save_server_info.py
+++ b/tests/scripts/thread-cert/test_srp_client_save_server_info.py
@@ -1,0 +1,162 @@
+#!/usr/bin/env python3
+#
+#  Copyright (c) 2021, The OpenThread Authors.
+#  All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions are met:
+#  1. Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#  2. Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in the
+#     documentation and/or other materials provided with the distribution.
+#  3. Neither the name of the copyright holder nor the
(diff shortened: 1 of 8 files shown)


simonlingoogle (approved):
LGTM 👍

tests/scripts/thread-cert/test_srp_client_save_server_info.py
```diff
@@ -0,0 +1,162 @@
...
+            'mode': 'rn',
+        },
+        SERVER3: {
+            'name': 'SRP_SERVER3',
+            'masterkey': '00112233445566778899aabbccddeeff',
+            'mode': 'rn',
+        },
+    }
+
+    def test(self):
```
jwhui, line 81: How about having a step that resets the node so it can test SRP registration after reattach?

abtink: I was actually considering this at first. I played a bit with `cli reset` to see if/how it works under simulation and in posix/RCP case. I am not sure if it works reliably on all configs, so decided to avoid it and not use `reset` (side-note: I know there are efforts to make `reset` function better and I think would be useful for testing as well).

For testing this feature itself, the stop/start of the SRP client behaves the same way as device reset and can cover the new code and the feature, so I went for this model in the current test.

jwhui (approved):

PR #6678, opened 2021-05-25
[mac] always set Key Identifier for Transmit Security

lmaciejonczyk:
For multi-layer application, the radio requires from the upper layer
to provide information for security processing like Key Identifier,
in the Auxiliary Security Header according to IEEE802.15.4-2015.

This change matters for handling security processing by radio only.

### DIFF
--- a/src/core/mac/sub_mac.cpp
+++ b/src/core/mac/sub_mac.cpp
@@ -328,9 +328,15 @@ void SubMac::ProcessTransmitSecurity(void)
     const ExtAddress *extAddress = nullptr;
     uint8_t           keyIdMode;
 
-    VerifyOrExit(ShouldHandleTransmitSecurity());
     VerifyOrExit(mTransmitFrame.GetSecurityEnabled());
     VerifyOrExit(!mTransmitFrame.IsSecurityProcessed());
+
+    if (!mTransmitFrame.IsARetransmission())
+    {
+        mTransmitFrame.SetKeyId(mKeyId);
+    }
+
+    VerifyOrExit(ShouldHandleTransmitSecurity());
 
     SuccessOrExit(mTransmitFrame.GetKeyIdMode(keyIdMode));
     VerifyOrExit(keyIdMode == Frame::kKeyIdMode1);
@@ -341,7 +347,6 @@ void SubMac::ProcessTransmitSecurity(void)
     {
         uint32_t frameCounter = GetFrameCounter();
 
-        mTransmitFrame.SetKeyId(mKeyId);
         mTransmitFrame.SetFrameCounter(frameCounter);
         UpdateFrameCounter(frameCounter + 1);
     }


src/core/mac/sub_mac.cpp
```diff
@@ -328,10 +328,16 @@ void SubMac::ProcessTransmitSecurity(void)
     const ExtAddress *extAddress = nullptr;
     uint8_t           keyIdMode;
 
-    VerifyOrExit(ShouldHandleTransmitSecurity());
     VerifyOrExit(mTransmitFrame.GetSecurityEnabled());
     VerifyOrExit(!mTransmitFrame.IsSecurityProcessed());
 
+    if (!mTransmitFrame.IsARetransmission())
```
edmont, line 334: Probably this check is not needed since Key Id will remain the same across retransmissions anyway.

lmaciejonczyk: On the other hand why to set it again if it's already set? In any case I'm ok with both options.

edmont: The only reason not to check is code size. I'm also okay with both options.

src/core/mac/sub_mac.cpp
```diff
@@ -328,10 +328,13 @@ void SubMac::ProcessTransmitSecurity(void)
     const ExtAddress *extAddress = nullptr;
     uint8_t           keyIdMode;
 
-    VerifyOrExit(ShouldHandleTransmitSecurity());
     VerifyOrExit(mTransmitFrame.GetSecurityEnabled());
     VerifyOrExit(!mTransmitFrame.IsSecurityProcessed());
 
+    mTransmitFrame.SetKeyId(mKeyId);
```
abtink, line 334: I think this would be incorrect.

If the frame is an indirect retx, then the key id (along with frame counter) is already set by the next layers above `Mac/SubMac` on the frame.   Adding this line here will overwrite the value.

abtink (commented):
Thanks @lmaciejonczyk.

I think setting the `KeyId` on frame from `ProcessTransmitSecurity()` would not be correct and overwrite the value for indirect retx.

abtink:
> For multi-layer application, the radio requires from the upper layer
> to provide information for security processing like Key Identifier,
> in the Auxiliary Security Header according to IEEE802.15.4-2015.
>
> This change matters for handling security processing by radio only.

I think the interaction is as follows:
- If OT (subMac/Mac) handles tx security then they manage the frame counter and set the key id and frame counter on the frame before passing it to radio platform layer.
- But if the radio platform handles tx security itself, then it is expected to manage frame counters and set the counter and key id on the frame (based on `mIsRetx` flag in the frame).

https://github.com/openthread/openthread/blob/abcf3ba152cd226e73e3d60217687b62613de98b/include/openthread/platform/radio.h#L240-L258

The radio platform should know the counter and key id from https://github.com/openthread/openthread/blob/abcf3ba152cd226e73e3d60217687b62613de98b/include/openthread/platform/radio.h#L546-L575

src/core/mac/sub_mac.cpp
abtink: > Probably this check is not needed since Key Id will remain the same across retransmissions anyway.

Indirect frame retransmission can happen much later and on newly prepared frame (when/if the next data poll is received from the child).

src/core/mac/sub_mac.cpp
abtink: Note that indirect frame retransmission can happen much later and on newly prepared frame (when/if the next data poll is received from the child). We need to ensure to use the same frame counter, key id, and seq number for such re-transmissions. 
This is all managed by `DataPollHandler`:
- If a an indirect tx fails, it saves the frame counter, key id and seq number in the `Child` entry (in child table) (`DataPollHandler::HandleSentFrame()`).
- For retx, it sets the same values on the prepared frame `DataPollHandler::HandleFrameRequest()` and mark the frame as a retx (`mIsRetx` set to `true`).

hubertmis:
> * But if the radio platform handles tx security itself, then it is expected to manage frame counters and set the counter and key id on the frame (based on `mIsRetx` flag in the frame).

According to IEEE 802.15.4, whole Auxiliary Security Header is constructed by the module processing TX security. Not only Frame Counter and Key Id fields. To do that the module processing security requires set of data per each frame being transmitted. This set is defined in MCPS-DATA.req (SecurityLevel, KeyIdMode, KeySource, KeyIndex). This set of data can be passed from OT MAC to radio platform in multiple ways:

1. We extend `otPlatRadioTransmit()` function with new parameters including these values.
2. We extend `otRadioFrame` structure with new fields including these values.
3. We embed these values in MAC Header of the frame already being passed to Radio module in `aFrame` parameter.

In OpenThread interaction between OpenThread MAC and Radio platform is _not_ based on the set of parameters from MCPS-DATA.req primitive, but instead it is based on PHY payload (including MAC Header) prepared by OpenThread. One example of such interaction is AR field in FCF. Based on value of this field set by OpenThread, radio platform behaves differently. To continue the same way of interaction between OT MAC and Radio platform, the best way of passing security details to Radio platform module is 3 (embed data directly in MAC Header). Moreover, this way of interaction is easier to serialize, because you embed data in frame PHY payload being serialized anyway in RCP configuration. You don't need to add more parameters to serialization format.

The goal of this pull request is to pass missing data from OT MAC to radio platform when radio platform is performing encryption.

src/core/mac/sub_mac.cpp
hubertmis: > If the frame is an indirect retx, then the key id (along with frame counter) is already set by the next layers

So perhaps higher layers shall set key id regardless security being handled by Radio Platform. This way Radio Platform would know which key to use for this particular frame. What do you think?

abtink:
@hubertmis , some background on this:
- The capbility `OT_RADIO_CAPS_TRANSMIT_SEC` which allows radio driver to handle security was added from https://github.com/openthread/openthread/pull/4919. 
- The main reason is to support enhanced ack (allow radio platform to generate and send enhanced ack).
- For this radio platform needs to manage the frame counter and perform AES-CCM* encryption.
- This is limited to Key ID mode 1 only (other modes will be handled by `Mac`, that is the tx frame will be encrypted by `Mac` and will be marked with its `mIsSecurityProcessed` being `true`.
- So the radio platform is only expected to handle Key ID mode 1.
- I think to process tx security, the radio platform should already have all the info (the counter is passed to radio platform from `otPlatRadioSetMacFrameCounter()` and the keys and key id value is given in`void otPlatRadioSetMacKey()`).

I think radio platform can implement similar logic as what we have in `SubMac` to perform this:
https://github.com/openthread/openthread/blob/abcf3ba152cd226e73e3d60217687b62613de98b/src/core/mac/sub_mac.cpp#L326-L360
(basically if radio platform does not provide certain capability `SubMac` will provide it, so `SubMac` code can be used as  an example of how radio platform should behave).

src/core/mac/sub_mac.cpp
abtink: I think it may be fine to set the key id field in the frame from `SubMac` but we definitely need the `if (!IsARetransmission())` check to ensure to not overwrite the value for indirect retx.

@brianljt, @jwhui  what do you think?

hubertmis:
@abtink , I understand OpenThread perspective. In OpenThread there is a "default key" being used to secure any frame being transmitted. However, IEEE 802.15.4 does not have a concept of "default key". In 802.15.4 you provide key value alongside payload for each frame.

It makes integration of OpenThread radio platform with generic 802.15.4 drivers problematic. Such glue layer would need to remember "default key index" passed in `otPlatRadioSetMacKey()` and parse frame being transmitted only to copy this key index value in appropriate place in Auxiliary Security Header.

I think if OpenThread filled keyIndex value regardless which layer pastes frame counter to the frame, it would cause no harm. If radio platform implementation copies key index based on the default value, it would just override value passed by OpenThread. But it would make glue between OpenThread Radio Platform and generic 802.15.4 drivers expecting key index inside MAC Header much easier to implement.

abtink:
> I think if OpenThread filled keyIndex value regardless which layer pastes frame counter to the frame, it would cause no harm.

@hubertmis I agree. I don't see any harm in setting the `KeyId` field on the frame from `SubMac` (if not set already by `DataPollHandler`). 

The frame counter though (if not already set) needs to be handled/updated on the frame by the radio platform since radio platform would be one managing the counter and knowing the latest value to use (as it is shared by for tx frames and enhanced ack tx frames).

abtink (approved):
LGTM. 👍

jwhui (approved):

PR #6679, opened 2021-05-25
[ip6] fix message leak when using platform TCP

gjc13:
When using platform TCP, the `Ip6::HandlePayload` function will be called with `aIpProto = kProtoTcp` and `aMessageOwnerShip = kTakeCustody`.

 The `FreeMessage(message);` will be called with `message = nullptr` currently, causing the message to leak. This PR fixes this issue.

### DIFF
--- a/src/core/net/ip6.cpp
+++ b/src/core/net/ip6.cpp
@@ -946,19 +946,13 @@ Error Ip6::HandlePayload(Message &          aMessage,
                          Message::Ownership aMessageOwnership)
 {
     Error    error   = kErrorNone;
-    Message *message = nullptr;
+    Message *message = (aMessageOwnership == Message::kTakeCustody) ? &aMessage : nullptr;
 
     VerifyOrExit(aIpProto == kProtoUdp || aIpProto == kProtoIcmp6);
 
-    switch (aMessageOwnership)
+    if (aMessageOwnership == Message::kCopyToUse)
     {
-    case Message::kTakeCustody:
-        message = &aMessage;
-        break;
-
-    case Message::kCopyToUse:
         VerifyOrExit((message = aMessage.Clone()) != nullptr, error = kErrorNoBufs);
-        break;
     }
 
     switch (aIpProto)


bukepo (approved):
LGTM:+1:

src/core/net/ip6.cpp
```diff
@@ -948,7 +948,7 @@ Error Ip6::HandlePayload(Message &          aMessage,
     Error    error   = kErrorNone;
     Message *message = nullptr;
 
-    VerifyOrExit(aIpProto == kProtoUdp || aIpProto == kProtoIcmp6);
+    OT_ASSERT(aIpProto != kProtoTcp || aMessageOwnership != Message::kCopyToUse);
```
simonlingoogle, line 951:
```suggestion
    VerifyOrExit(aIpProto == kProtoUdp || aIpProto == kProtoIcmp6, {
      if (aMessageOwnership == Message::kTakeCustody) {
          message = &aMessage;
      }
    });
```

I prefer slightly not to use `OT_ASSERT` here. I think this method is tend to be general to handle all kinds of situations. 

Thoughts? @abtink

simonlingoogle (commented):
Great catch! Thanks.

src/core/net/ip6.cpp
abtink: @gjc13 , very good catch!

I fully agree with @simonlingoogle . `OT_ASSERT()` is not good here and the code here should be general.

I think would be even simpler to set the `message` at the start of this method instead of in `VerifyOrExit()` 
- In general, I feel using braces `{ }` in `VerifyOrExit()` make it less readable, and better to use `if() { ....;  ExitNow(); }` for this. 

Here is my suggestion for the change:
```c++
Error Ip6::HandlePayload(Message &          aMessage,
                         MessageInfo &      aMessageInfo,
                         uint8_t            aIpProto,
                         Message::Ownership aMessageOwnership)
{
    Error    error   = kErrorNone;
    Message *message = (aMessageOwnership == Message::kTakeCustody) ? &aMessage : nullptr;

    VerifyOrExit(aIpProto == kProtoUdp || aIpProto == kProtoIcmp6);

    if (message == nullptr) // `aMessageOwnership` is `kCopyToUse`
    {
        VerifyOrExit((message = aMessage.Clone()) != nullptr, error = kErrorNoBufs);
    }
```

Basically we set the `message` at the top of the method based on the `aMessageOwnership` and then do the `VerifyOrExit()` check. If we see `message` is `nullptr` it indicates we need to copy.

bukepo: I guess @gjc13 just wants to assert that TCP will not be handled and forwarded at the same time. Because TCP doesn't work on multicast. If we assert here, attackers may turn a node down by sending TCP traffic with a multicast destination.

simonlingoogle: Fully agree with @abtink. Actually, had almost the same code in mind.

jinran-google: Good catch! Thank you, @gjc13 !

I agree with @simonlingoogle and @abtink . Though currently calls with `aIpProto == kProtoTcp && aMessageOwnership == kCopyToUse` only happen with multicast TCP messages (which shouldn't exist), there's no guarantee that the node would never receive such messages.

gjc13: Updated.

src/core/net/ip6.cpp
```diff
@@ -946,19 +946,13 @@ Error Ip6::HandlePayload(Message &          aMessage,
                          Message::Ownership aMessageOwnership)
 {
     Error    error   = kErrorNone;
-    Message *message = nullptr;
+    Message *message = (aMessageOwnership == Message::kTakeCustody) ? &aMessage : nullptr;
 
     VerifyOrExit(aIpProto == kProtoUdp || aIpProto == kProtoIcmp6);
 
-    switch (aMessageOwnership)
+    if (aMessageOwnership == Message::kTakeCustody)
```
jwhui, line 953:
```suggestion
    if (message == nullptr)
```
or
```suggestion
    if (aMessageOwnership == Message::kCopyToUse)
```

simonlingoogle (approved):
LGTM :+1: after resolving Jonathan's comment.

abtink (approved):
Looks great! Thanks. 👍

jwhui (approved):

PR #6681, opened 2021-05-25
[cli] make OPENTHREAD_FTD and OPENTHREAD_MTD PUBLIC in cli library

### LINKED ISSUE #6673: Macro `OPENTHREAD_FTD` is not available in app code
**Describe the bug** 

I was doing the OpenThread API [codelab](https://openthread.io/codelabs/openthread-apis#0). At step [#9](https://openthread.io/codelabs/openthread-apis#8) it used the `OPENTHREAD_FTD` macro in the cli app to determine whether it's built for an FTD device. However, I found this macro seemed to always be 0/undefined. When I connected to the device using  `screen`, I found the `routerselectionjitter` to be 120 and the dataset was not the same as expected. I moved   `otDatasetSetActive(aInstance, &aDataset);` out of the `#ifdef` scope and the dataset became the same as expected. Therefore I think the cause was that `OPENTHREAD_FTD` is not available in app code.

superwhd:
The purpose is to make `OPENTHREAD_FTD` and `OPENTHREAD_MTD` available in `examples/apps/cli/main.c`

Fixes #6673

### DIFF
--- a/src/cli/ftd.cmake
+++ b/src/cli/ftd.cmake
@@ -29,7 +29,7 @@
 add_library(openthread-cli-ftd)
 
 target_compile_definitions(openthread-cli-ftd
-    PRIVATE
+    PUBLIC
         OPENTHREAD_FTD=1
 )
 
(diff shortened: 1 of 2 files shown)


superwhd:
According to an offline discussion with @bukepo , the `OPENTHREAD_FTD` macro is exclusively for internal use. Therefore I'll close this PR without exposing that macro.

PR #6683, opened 2021-05-25
[settings] simplify and enhance `Settings` (reuse similar patterns)

abtink:
This commit updates the `Settings` implementation. It adds template
versions of methods `Read<EntryType>()`, `Save<EntryType>()`, and
`Delete<EntryType>()` which make it easier for other modules to use,
while ensuing to share a common underlying implementation for all
entry types.

The settings entry types (e.g., `NetworkInfo`, `ParentInfo`, etc.) are
changed in this commit to provide a constant `kKey` specifying the
associated setting key with the entry. For simple entries, a related
class is added to define the `kKey` constant and the associated value
type (e.g., `OmrPrefix` defines `Ip6::Prefix` as its `ValueType`).
This model makes it easier to add new setting keys and entries in
future.

This commit also simplifies the logging in `Settings`. The entry types
now provide `Log()` method to log their data and a common
`Settings::Log()` method is added which handles all logging (on
success or in case of an error).

Finally, this commit moves the `SettingsDriver` definition to its own
`common/settings_driver.hpp` header file and and ensures that its
methods are defined as inline.

------

Notes:
- This PR is currently contains the commit from https://github.com/openthread/openthread/pull/6672 (it is actually written over this commit so to make rebase easier). Please only review and check the second commit. Thanks. 
   - The code size report also includes this commit. The new commit will actually reduce code size (from ~200 to 500 for different build configs). 
- A lot of the methods are actually removed (replaced by the template version) in `settings.cpp` so the diff looks complex and I suggest just looking at the new implementation 😃

### DIFF
--- a/src/core/BUILD.gn
+++ b/src/core/BUILD.gn
@@ -393,6 +393,7 @@ openthread_core_files = [
   "common/random_manager.hpp",
   "common/settings.cpp",
   "common/settings.hpp",
+  "common/settings_driver.hpp",
   "common/string.cpp",
   "common/string.hpp",
   "common/tasklet.cpp",
--- a/src/core/Makefile.am
+++ b/src/core/Makefile.am
@@ -399,6 +399,7 @@ HEADERS_COMMON                                  = \
     common/random.hpp                             \
     common/random_manager.hpp                     \
     common/settings.hpp                           \
+    common/settings_driver.hpp                    \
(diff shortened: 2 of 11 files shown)


jwhui (approved):
Great to see reduced code size (in terms of both LoCs and compiled code)! 👍🏼

PR #6687, opened 2021-05-27
[netdata] assign domain IDs during registration of new entries

abtink:
This commit adds support for assigning and updating domain ID values
to Prefix TLVs on leader when registering new entries into the Network
Data (per Thread spec).

A border router adding an external route entry can specify which
source prefix(es) should be used with the external route by assigning
the same domain ID to the related prefix TLVs. Leader changes the
domain ID values assigning new IDs to new prefixes and the same IDs to
any existing prefixes in the Network Data, while ensuring to maintain
the same relationships between prefix TLVs.

### DIFF
--- a/src/core/thread/network_data_leader_ftd.cpp
+++ b/src/core/thread/network_data_leader_ftd.cpp
@@ -692,10 +692,10 @@ exit:
     return status;
 }
 
-void Leader::RegisterNetworkData(uint16_t aRloc16, const uint8_t *aTlvs, uint8_t aTlvsLength)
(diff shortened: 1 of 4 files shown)


abtink:
With this change, it seems like the  `border_router/test_single_border_router.py` is failing. Will investigate this.

abtink:
Closing this PR in light of [SPEC-985](https://threadgroup.atlassian.net/browse/SPEC-985) discussion (which simplifies the model and requires BRs to choose the domain ID when registering the entries). 

> Ensuring that BRs are responsible for selecting the Domain ID values makes it possible for BRs to avoid having to register a Border Router TLV in order to register a Has Route TLV.

PR #6693, opened 2021-05-29
[ip6] add public API to convert `otSockAddr` to string

abtink:
This commit adds public API `otIp6SockAddrToString()` which converts
an IPv6 socket address (`otSockAddr`) to human-readable string. The
CLI `ProcessNetStat()` is then updated to use the new API.


include/openthread/ip6.h
```diff
@@ -590,6 +590,24 @@ otError otIp6AddressFromString(const char *aString, otIp6Address *aAddress);
...
+ *
+ * If the resulting string does not fit in @p aBuffer (within its @p aSize characters), the string will be truncated
+ * but the outputted string is always null-terminated.
+ *
+ * @param[in]  aSockAddr A pointer to an IPv6 socket address (MUST NOT be NULL).
+ * @param[out] aBuffer   A pointer to a char array to output the string (MUST NOT be NULL).
+ * @param[in]  aSize     The size of @p aBuffer (in bytes). Recommended to use `OT_IP6_SOCK_ADDR_STRING_SIZE`.
+ *
+ */
+void otIp6SockAddrToString(const otSockAddr *aSockAddr, char *aBuffer, uint16_t aSize);
```
bukepo, line 609: returning the destination buffer might be more user-friendly.

```suggestion
const char* otIp6SockAddrToString(const otSockAddr *aSockAddr, char *aBuffer, uint16_t aSize);
```

abtink: I agree it can be a bit more convenient but I recommend keeping it `void`.
- The other similar APIs for IPv6 address and prefix are also `void`.
- The caller should already have the `aBuffer` as a variable, so there is no need to return it back (it is more of convenience for the caller if we return it).
- Having these APIs return `(const) char *` does impact code size.  It is small but the small changes can add up.
```
   text	   data	    bss	    dec	    hex	filename
 432572	    420	  56244	 489236	  77714	build/bin/ot-cli-ftd  - All `ToString()`s are `void`.
 432604	    420	  56244	 489268	  77734	build/bin/ot-cli-ftd  - All `ToString()` return `const char *`.
 ```

File: include/openthread/ip6.h [hunk 2]
```diff
@@ -590,6 +590,24 @@ otError otIp6AddressFromString(const char *aString, otIp6Address *aAddress);
  */
 void otIp6AddressToString(const otIp6Address *aAddress, char *aBuffer, uint16_t aSize);
 
+#define OT_IP6_SOCK_ADDR_STRING_SIZE 48 ///< Recommended size for string representation of an IPv6 socket address.
```
bukepo, line 593: Does this include the null-terminator?

bukepo (approved):
LGTM:+1:

include/openthread/ip6.h
simonlingoogle: Is it actually the maxmal size?

simonlingoogle (approved):
LGTM 👍

include/openthread/ip6.h
abtink: Yes. `48` should include the null terminator as well. Here is the calculation:
- IP address is max 40 chars (including null).
- We have `[` and `]:`  which adds 3 chars.
- Then the port number as decimal value with max `65536` or 5 chars.
- So total should be `40 + 3 + 5 = 48`.
- Or for `"[1111:2222:3333:4444:5555:6666:7777:8888]:65536"` (which has len 47 and counting null would be 48).

jwhui (approved):

Issue #6690, opened 2021-05-27
[github-actions] test_srp_lease.py fails intermittently


jwhui:
[logs_79964.zip](https://github.com/openthread/openthread/files/6555877/logs_79964.zip)
[packet-verification-pcaps.zip](https://github.com/openthread/openthread/files/6555880/packet-verification-pcaps.zip)

PR #6694, opened 2021-05-29
[tests] give more time for SRP client to upgrade to router

wgtdkp:
https://github.com/openthread/openthread/issues/6690 indicates that there are chances that a SRP client may not be able
to upgrade to a router in 5 seconds. This commit gives it 10 seconds to upgrade.

```txt
2021-05-27T17:51:04.6018976Z ======================================================================
2021-05-27T17:51:04.6021398Z FAIL: test (__main__.SrpRegisterSingleService)
2021-05-27T17:51:04.6022746Z ----------------------------------------------------------------------
2021-05-27T17:51:04.6024043Z Traceback (most recent call last):
2021-05-27T17:51:04.6025888Z   File "./tests/scripts/thread-cert/test_srp_lease.py", line 88, in test
2021-05-27T17:51:04.6027733Z     self.assertEqual(client.get_state(), 'router')
2021-05-27T17:51:04.6029001Z AssertionError: 'child' != 'router'
2021-05-27T17:51:04.6030288Z - child
2021-05-27T17:51:04.6030875Z + router
```

### DIFF
--- a/tests/scripts/thread-cert/test_srp_lease.py
+++ b/tests/scripts/thread-cert/test_srp_lease.py
@@ -84,7 +84,7 @@ class SrpRegisterSingleService(thread_cert.TestCase):
 
         client.srp_server_set_enabled(False)
         client.start()
-        self.simulator.go(5)
+        self.simulator.go(10)
         self.assertEqual(client.get_state(), 'router')
 
         #


simonlingoogle (approved):
LGTM :+1: 

Normally I think it should not happen that the client haven't upgrade to be a Router within 5 seconds since the default `routerselectionjitter` has been set to 1s. 
But considering the uncertain runtime of real time simulations in GitHub Runners, we can certainly use a longer delay.

jwhui (approved):

PR #6695, opened 2021-05-30
[cli] disallow concurrent commands

simonlingoogle:
This commit disallows concurrent commands. 
With this commit, the CLI only execute a new command after it complete the previous command. CLI can also prompt properly after the command execution is done. 

Other fixes and enhancements:
- Fixes premature command prompt:
  **With this enhancement:**
  ```
  > ping fdde:ad00:beef:0:0:ff:fe00:fc00 10 3 1
  18 bytes from fdde:ad00:beef:0:0:ff:fe00:fc00: icmp_seq=1 hlim=64 time=0ms
  18 bytes from fdde:ad00:beef:0:0:ff:fe00:fc00: icmp_seq=2 hlim=64 time=0ms
  18 bytes from fdde:ad00:beef:0:0:ff:fe00:fc00: icmp_seq=3 hlim=64 time=0ms
  3 packets transmitted, 3 packets received. Packet loss = 0.0%. Round-trip min/avg/max = 0/0.0/0 ms.
  Done
  > scan
  | J | Network Name     | Extended PAN     | PAN  | MAC Address      | Ch | dBm | LQI |
  +---+------------------+------------------+------+------------------+----+-----+-----+
  | 0 | OpenThread       | 000db80000000000 | face | 166e0a0000000001 | 11 | -20 |   0 |
  Done
  > networkdiagnostic get ff02::1 0 1
  DIAG_GET.rsp/ans from fe80:0:0:0:2437:8587:62c4:b2cc: 00082637858762c4b2cc01027400
  Ext Address: '2637858762c4b2cc'
  Rloc16: 0x7400
  DIAG_GET.rsp/ans from fe80:0:0:0:146e:a00:0:1: 0008166e0a000000000101028800
  Ext Address: '166e0a0000000001'
  Rloc16: 0x8800
  Done
  > <prompt>
  ```
  **Orignal:**
  ```
  > ping fdde:ad00:beef:0:0:ff:fe00:fc00 10 3 1
  > 18 bytes from fdde:ad00:beef:0:0:ff:fe00:fc00: icmp_seq=1 hlim=64 time=0ms
  18 bytes from fdde:ad00:beef:0:0:ff:fe00:fc00: icmp_seq=2 hlim=64 time=0ms
  18 bytes from fdde:ad00:beef:0:0:ff:fe00:fc00: icmp_seq=3 hlim=64 time=0ms
  3 packets transmitted, 3 packets received. Packet loss = 0.0%. Round-trip min/avg/max = 0/0.0/0 ms.
  Done
  scan
  | J | Network Name     | Extended PAN     | PAN  | MAC Address      | Ch | dBm | LQI |
  +---+------------------+------------------+------+------------------+----+-----+-----+
  > | 0 | OpenThread       | 000db80000000000 | face | 2637858762c4b2cc | 11 | -20 |   0 |
  Done
  networkdiagnostic get ff02::1 0 1
  > DIAG_GET.rsp/ans from fe80:0:0:0:146e:a00:0:1: 0008166e0a000000000101028800
  Ext Address: '166e0a0000000001'
  Rloc16: 0x8800
  Done
  DIAG_GET.rsp/ans from fe80:0:0:0:2437:8587:62c4:b2cc: 00082637858762c4b2cc01027400
  Ext Address: '2637858762c4b2cc'
  Rloc16: 0x7400
  Done
  <prompt>
  ```
- Add `ping async` command for ping in async mode: output `Done` immediately but print ping responses later on. 
- Fixes `networkdiagnostic get` outputs multiple `Done` by always waiting for 5 seconds. 

Depends-On: openthread/ot-ns#130

### DIFF
--- a/examples/apps/cli/cli_uart.cpp
+++ b/examples/apps/cli/cli_uart.cpp
@@ -134,9 +134,9 @@ static otError ProcessCommand(void);
 
 static void ReceiveTask(const uint8_t *aBuf, uint16_t aBufLength)
 {
-    static const char sEraseString[]   = {'\b', ' ', '\b'};
-    static const char CRNL[]           = {'\r', '\n'};
-    static const char sCommandPrompt[] = {'>', ' '};
+    static const char sEraseString[] = {'\b', ' ', '\b'};
+    static const char CRNL[]         = {'\r', '\n'};
+    static uint8_t    sLastChar      = '\0';
     const uint8_t *   end;
 
     end = aBuf + aBufLength;
@@ -145,17 +145,18 @@ static void ReceiveTask(const uint8_t *aBuf, uint16_t aBufLength)
     {
         switch (*aBuf)
         {
-        case '\r':
         case '\n':
-            Output(CRNL, sizeof(CRNL));
-            if (sRxLength > 0)
+            if (sLastChar == '\r')
             {
-                sRxBuffer[sRxLength] = '\0';
-                IgnoreError(ProcessCommand());
+                break;
             }
 
-            Output(sCommandPrompt, sizeof(sCommandPrompt));
+            OT_FALL_THROUGH;
 
+        case '\r':
+            Output(CRNL, sizeof(CRNL));
+            sRxBuffer[sRxLength] = '\0';
+            IgnoreError(ProcessCommand());
             break;
 
 #if OPENTHREAD_POSIX && !defined(FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION)
@@ -187,6 +188,8 @@ static void ReceiveTask(const uint8_t *aBuf, uint16_t aBufLength)
 
             break;
         }
+
+        sLastChar = *aBuf;
     }
 }
 
@@ -194,16 +197,12 @@ static otError ProcessCommand(void)
 {
     otError error = OT_ERROR_NONE;
 
-    while (sRxBuffer[sRxLength - 1] == '\n' || sRxBuffer[sRxLength - 1] == '\r')
+    while (sRxLength > 0 && (sRxBuffer[sRxLength - 1] == '\n' || sRxBuffer[sRxLength - 1] == '\r'))
     {
         sRxBuffer[--sRxLength] = '\0';
     }
 
-    if (sRxLength > 0)
-    {
-        otCliInputLine(sRxBuffer);
-    }
-
+    otCliInputLine(sRxBuffer);
     sRxLength = 0;
 
     return error;
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -101,6 +101,7 @@ Interpreter::Interpreter(Instance *aInstance, otCliOutputCallback aCallback, voi
     , mOutputContext(aContext)
     , mUserCommands(nullptr)
     , mUserCommandsLength(0)
+    , mCommandIsPending(false)
 #if OPENTHREAD_CONFIG_SNTP_CLIENT_ENABLE
     , mSntpQueryingInProgress(false)
 #endif
@@ -110,6 +111,7 @@ Interpreter::Interpreter(Instance *aInstance, otCliOutputCallback aCallback, voi
 #if OPENTHREAD_CONFIG_TCP_ENABLE && OPENTHREAD_CONFIG_CLI_TCP_ENABLE
     , mTcp(*this)
 #endif
+    , mTimer(*aInstance, HandleTimer, this)
 #if OPENTHREAD_CONFIG_COAP_API_ENABLE
     , mCoap(*this)
 #endif
@@ -139,22 +141,31 @@ Interpreter::Interpreter(Instance *aInstance, otCliOutputCallback aCallback, voi
 #if OPENTHREAD_FTD
     otThreadSetDiscoveryRequestCallback(mInstance, &Interpreter::HandleDiscoveryRequest, this);
 #endif
+
+    OutputPrompt();
 }
 
 void Interpreter::OutputResult(otError aError)
 {
-    switch (aError)
+    OT_ASSERT(mCommandIsPending);
+
+    VerifyOrExit(aError != OT_ERROR_PENDING);
+
+    if (aError == OT_ERROR_NONE)
     {
-    case OT_ERROR_NONE:
         OutputLine("Done");
-        break;
-
-    case OT_ERROR_PENDING:
-        break;
-
-    default:
+    }
+    else
+    {
         OutputLine("Error %d: %s", aError, otThreadErrorToString(aError));
     }
+
+    mCommandIsPending = false;
+    mTimer.Stop();
+    OutputPrompt();
+
+exit:
+    return;
 }
 
 void Interpreter::OutputBytes(const uint8_t *aBytes, uint16_t aLength)
@@ -3127,18 +3138,28 @@ void Interpreter::HandlePingStatistics(const otPingSenderStatistics *aStatistics
     }
 
     OutputLine("");
-    OutputResult(OT_ERROR_NONE);
+
+    if (!mPingIsAsync)
+    {
+        OutputResult(OT_ERROR_NONE);
+    }
 }
 
 otError Interpreter::ProcessPing(Arg aArgs[])
 {
     otError            error = OT_ERROR_NONE;
     otPingSenderConfig config;
+    bool               async = false;
 
     if (aArgs[0] == "stop")
     {
         otPingSenderStop(mInstance);
         ExitNow();
+    }
+    else if (aArgs[0] == "async")
+    {
+        async = true;
+        aArgs++;
     }
 
     memset(&config, 0, sizeof(config));
@@ -3206,7 +3227,14 @@ otError Interpreter::ProcessPing(Arg aArgs[])
     config.mStatisticsCallback = Interpreter::HandlePingStatistics;
     config.mCallbackContext    = this;
 
-    error = otPingSenderPing(mInstance, &config);
+    SuccessOrExit(error = otPingSenderPing(mInstance, &config));
+
+    mPingIsAsync = async;
+
+    if (!async)
+    {
+        error = kErrorPending;
+    }
 
 exit:
     return error;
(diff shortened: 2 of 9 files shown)


abtink:
@simonlingoogle , is there any use-case for this change? Is it just a enhenamcnet for user?

There are certain commands/operation that are async by nature and can take some time to respond, so may be good to allow user to issue other commands in between
- Energy or active/discover scan command
- dns resolver or browse commands, 
- getting net diags
- sntp query
- mlr registeration
- etc.

simonlingoogle:
> @simonlingoogle , is there any use-case for this change? Is it just a enhenamcnet for user?
> 
> There are certain commands/operation that are async by nature and can take some time to respond, so may be good to allow user to issue other commands in between
> 
> * Energy or active/discover scan command
> * dns resolver or browse commands,
> * getting net diags
> * sntp query
> * mlr registeration
> * etc.

It's an enhancement for user. And also I am seeing it as a bug fix:
- Outputting prompt `> ` before the command finishes (i.e. output Done) is actually an incorrect CLI behavior. 

For a long time we have been suffering from handling `'> '` that could appear unexpectedly in the CLI output when interacting with CLI in scripts. This PR should make the CLI output consistent to what we have been experiencing on a conventional shell (e.g. bash). 

For these "async" commands, I would say they are not actually async by nature. For example, `dig` or `ping` command on Linux blocks the terminal from responding until success or timeout. I believe that should be the default model we use for these commands. They should block the CLI for a period until success or timeout. 

Moreover, we don't have any need for async commands in testing. It seems also reasonable for human user to wait for a command to finish before issuing another command. 

Thoughts?

abtink:
@simonlingoogle , in general I like this.

To be frank, I am also not fan of how the `>` is shown in middle when we `scan` or other similar commands with current model.
```
> scan
| J | Network Name     | Extended PAN     | PAN  | MAC Address      | Ch | dBm | LQI |
+---+------------------+------------------+------+------------------+----+-----+-----+
> | 0 | OpenThread       | 000db80000000000 | 0001 | 166e0a0000000002 | 11 | -20 |   0 |
| 1 | OpenThread       | 000db80000000000 | 0002 | 166e0a0000000003 | 11 | -20 |   0 |
Done
```

And I noticed in `node.py` code how it is handling the extar `>` (which I agree is not good).

As an option, we can conisder adding  mechansim for user to stop a running command (if it happens to take too long) and get the command prompt back (for example, can we monitor and handle somethin like `ctrl+c`?)

simonlingoogle:
> As an option, we can conisder adding mechansim for user to stop a running command (if it happens to take too long) and get the command prompt back (for example, can we monitor and handle somethin like `ctrl+c`?)

That's a good idea, which we can enhance in future PRs. However, I can see some difficulties in implementing this though. We are going to need a command lifetime model for each "async" command to invoke timeout or cancel callbacks.

wgtdkp:
Instead of adding a new `async` sub-command for each command that may expect async results, what about have a new dedicated command to switch between blocking and non-blocking CLI mode:
```shell
> climode blocking
Done
>
> climode nonblocking
Done
>
```

I think this should be more scalable.

simonlingoogle:
After resolving several issues discovered by GitHub Actions, this feature is now working. 

Feel free to check CLI changes, especially for **async** commands like `ping`, `scan` and `networkdiagnostic get`. 

@bukepo @abtink @wgtdkp

simonlingoogle:
> Instead of adding a new `async` sub-command for each command that may expect async results, what about having a new dedicated command to switch between blocking and non-blocking CLI mode:
> 
> ```shell
> > climode blocking
> Done
> >
> > climode nonblocking
> Done
> >
> ```
> 
> I think this should be more scalable.

I think it's getting more complex than necessary to allow toggling CLI non-blocking mode for now.

`ping async` is currently useful only in OTNS. 

But we can definitely explore more scalable solution if there are multiple commands that need `async` mode.

bukepo (commented):
Great enhancement! I noticed one issue: press enter doesn't solicit a prompt, i.e. I expect multiple empty command should result in multiple prompts as follows:
```bash
>
>
>
```
This is common behavior of command lines and useful to check if the shell is stuck. Can we keep the original behavior?

simonlingoogle:
> Great enhancement! I noticed one issue: press enter doesn't solicit a prompt, i.e. I expect multiple empty command should result in multiple prompts as follows:
> 
> ```bash
> 
> >
> 
> >
> 
> >
> 
> ```
> 
> This is common behavior of command lines and useful to check if the shell is stuck. Can we keep the original behavior?

Agree. This is my expected behavior too, and it should be so. How did you run CLI (simulation or posix or daemon)?

tests/scripts/expect/_common.exp
```diff
@@ -58,6 +58,7 @@ proc wait_for {command success {failure {[\r\n]FAILURE_NOT_EXPECTED[\r\n]}}} {
 }
 
 proc expect_line {line} {
+    set timeout 10
```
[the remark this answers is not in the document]
simonlingoogle: Since most calls to `expect_line` should return very quickly, the 10s timeout won't be causing any problem?

IIUC, we actually set the default timeout to 10s at the end of `_common.exp`

bukepo:
> > Great enhancement! I noticed one issue: press enter doesn't solicit a prompt, i.e. I expect multiple empty command should result in multiple prompts as follows:
> > ```shell
> > >
> > 
> > >
> > 
> > >
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is common behavior of command lines and useful to check if the shell is stuck. Can we keep the original behavior?
> 
> Agree. This is my expected behavior too, and it should be so. How did you run CLI (simulation or posix or daemon)?

I played with this PR on posix platform. And readline is enabled.

simonlingoogle:
> > > Great enhancement! I noticed one issue: press enter doesn't solicit a prompt, i.e. I expect multiple empty command should result in multiple prompts as follows:
> > > ```shell
> > > >
> > > 
> > > >
> > > 
> > > >
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This is common behavior of command lines and useful to check if the shell is stuck. Can we keep the original behavior?
> > 
> > 
> > Agree. This is my expected behavior too, and it should be so. How did you run CLI (simulation or posix or daemon)?
> 
> I played with this PR on posix platform. And readline is enabled.

I think this is an optimization done by `readline`. We actually did send the empty line to CLI and CLI has printed another prompt `> `.

wgtdkp:
> For example, we may allow using ping <ip> ... & to ping in the background in the future, similar to what we do in shell.

SGTM!

wgtdkp (approved):

examples/apps/cli/cli_uart.cpp
```diff
@@ -147,15 +147,12 @@ static void ReceiveTask(const uint8_t *aBuf, uint16_t aBufLength)
         {
         case '\r':
         case '\n':
-            Output(CRNL, sizeof(CRNL));
-            if (sRxLength > 0)
+            if (!(sLastChar == '\r' && *aBuf == '\n'))
```
abtink, line 150: How about moving this check under `case '\n'`?  I guess the goal is to skip if last is `\r` and new char is `\n`?

- Under `case \n`, we can do ` if (sLastChar == '\r')` and if it matches we `break`, otherwise fall through to `case \r`?

src/cli/cli.cpp
```diff
@@ -4570,21 +4600,18 @@ void Interpreter::ProcessLine(char *aBuf)
 
     OT_ASSERT(aBuf != nullptr);
 
+    // Ignore the command if another command is pending.
+    VerifyOrExit(!mCommandIsPending, args[0].Clear());
```
abtink, line 4604: Can we instead of doing `args[0].Clear()` set the error to `OT_ERROR_PENDING`?
I think we handle it the same way in `exit` but may be more clear to set the error. Thoughts?

src/cli/cli.cpp
```diff
@@ -139,10 +141,14 @@ Interpreter::Interpreter(Instance *aInstance, otCliOutputCallback aCallback, voi
...
     otThreadSetDiscoveryRequestCallback(mInstance, &Interpreter::HandleDiscoveryRequest, this);
 #endif
+
+    OutputPrompt();
 }
 
 void Interpreter::OutputResult(otError aError)
 {
+    OT_ASSERT(mCommandIsPending);
+
```
abtink, line 151: Current code is correct but just a suggestion to maybe make it simpler: 
How about if we do a `VerifyOrExit(aError != OT_ERROR_PENDING)` at the top here?
Then we can remove both `case OT_ERROR_PENDING` and also the `if (aError != OT_ERROR_PENDING)` check.

src/cli/cli.cpp
```diff
@@ -4667,6 +4698,7 @@ otError Interpreter::ProcessNetworkDiagnostic(Arg aArgs[])
     {
         SuccessOrExit(error = otThreadSendDiagnosticGet(mInstance, &address, tlvTypes, count,
                                                         &Interpreter::HandleDiagnosticGetResponse, this));
+        SetCommandTimeout(5000);
```
abtink, line 4701: Avoid magic number?

abtink, line 4701: Is the `SetCommandTimeout()` and `mTimer` only used for the `otThreadSendDiagnosticGet()`? Do we intend to use it for other command later?

I see that we don't call `OutputResult()` in `HandleDiagnosticGetResponse()` anymore. So does it mean that we always wait for 5seconds for this command (even if we receive a response, we still wait for 5 seconds, is this correct?).

If it will only be used for this command,  can we consider adding the timeout in the `otThreadSendDiagnosticGet()` (and `NetworkDiagnostic` module) rather than in CLI?

I am ok if we leave such a change for a later PR. Just wanted to suggest it and get peoples ideas/thoughts on this.

abtink (approved):
Looks great! Thanks @simonlingoogle 👍 

Few smaller suggestions below. It LGTM as is.

src/cli/cli.cpp
simonlingoogle: > Is the `SetCommandTimeout()` and `mTimer` only used for the `otThreadSendDiagnosticGet()`? Do we intend to use it for other command later?

Yes. I intend to use this single Timer for all commands that need timeouts, saving RAM size.
Another possible use case for command timeout is `joiner start`. It's also useful for `wait` command in https://github.com/openthread/openthread/pull/6979.

> I see that we don't call `OutputResult()` in `HandleDiagnosticGetResponse()` anymore. So does it mean that we always wait for 5seconds for this command (even if we receive a response, we still wait for 5 seconds, is this correct?).

Yes, that's correct. 

Currently `networkdiagnostic get` command always wait for 5s. We can add a `timeout` argument to `networkdiagnostic get` command when when need more controls over timeouts.

src/cli/cli.cpp
simonlingoogle: If we set the error to `OT_ERROR_PENDING` here, it will `OutputResult(OT_ERROR_PENDING) at exit, which is not what we want. 

The logic here is a little tricky. There are 4 cases we want to handle:
1. If the command executed sucessfully, we should `OutputResult(error)` with error being `OT_ERROR_NONE`.
2. If the command execution failed, we should `OutputResult(error)` with error not being `OT_ERROR_NONE`. 
3. If there is another command executing, we should output nothing by just ignoreing the input
4. If the input is empty, and there is no other command executing, we should only output a new prompt

So, we are implementing case 3 by ignore the input if another command is executing.

simonlingoogle: I looked at `OutputResult(OT_ERROR_PENDING)` again and found that it actually does nothing. 

So, you are correct that setting error to `OT_ERROR_PENDING` can be effectively same. But by doing so, we don't know at exit if CLI is executing the current command input or there is another command executing and the current command input is ignored. I am worried the unclearance may cause other issues in the future. 

Currently, the conditions are clear:
- If `error != OT_ERROR_NONE` , the current command was executed and error was returned
- If `error == OT_ERROR_NONE && !args[0].IsEmpty()`, the current command was executed successfully
- If `error == OT_ERROR_NONE && args[0].IsEmpty()`, the current command was ignored either because it's empty or because there is already an outstanding command executing.

abtink: Thanks. SGTM.

src/cli/cli.cpp
abtink: SGTM. Thanks.

abtink (approved):

bukepo (approved):

jwhui (approved):

PR #6697, opened 2021-05-31
[udp] fix potential infinite loop in `Udp::GetEphemeralPort`

superwhd:
The previous code will go into an infinite loop when `mEphemeralPort == Tmf::kUdpPort`, since `rval` is never updated in the loop.
```
uint16_t Udp::GetEphemeralPort(void)
{
    uint16_t rval = mEphemeralPort;

    do
    {
        if (mEphemeralPort < kDynamicPortMax)
        {
            mEphemeralPort++;
        }
        else
        {
            mEphemeralPort = kDynamicPortMin;
        }
    } while (rval == Tmf::kUdpPort);

    return rval;
}
```

### DIFF
--- a/src/core/net/udp6.cpp
+++ b/src/core/net/udp6.cpp
@@ -407,8 +407,6 @@ exit:
 
 uint16_t Udp::GetEphemeralPort(void)
 {
-    uint16_t &rval = mEphemeralPort;
-
     do
     {
         if (mEphemeralPort < kDynamicPortMax)
@@ -419,9 +417,9 @@ uint16_t Udp::GetEphemeralPort(void)
         {
             mEphemeralPort = kDynamicPortMin;
         }
-    } while (rval == Tmf::kUdpPort);
+    } while (mEphemeralPort == Tmf::kUdpPort);
 
-    return rval;
+    return mEphemeralPort;
 }
 
 Message *Udp::NewMessage(uint16_t aReserved, const Message::Settings &aSettings)
--- a/src/core/net/udp6.cpp
+++ b/src/core/net/udp6.cpp
@@ -407,7 +407,7 @@ exit:
 
 uint16_t Udp::GetEphemeralPort(void)
 {
-    uint16_t rval = mEphemeralPort;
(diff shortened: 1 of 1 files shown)


src/core/net/udp6.cpp
```diff
@@ -407,7 +407,7 @@ void Udp::RemoveSocket(SocketHandle &aSocket)
 
 uint16_t Udp::GetEphemeralPort(void)
 {
-    uint16_t rval = mEphemeralPort;
+    uint16_t &rval = mEphemeralPort;
```
simonlingoogle, line 410:
```suggestion
```
Do we really need an alias `&rval = mEphemeralPort`?

simonlingoogle (approved):
LGTM :+1: 

Good catch!

wgtdkp (approved):
LGTM :+1:

src/core/net/udp6.cpp
```diff
@@ -419,9 +417,9 @@ uint16_t Udp::GetEphemeralPort(void)
         {
             mEphemeralPort = kDynamicPortMin;
         }
-    } while (rval == Tmf::kUdpPort);
+    } while (mEphemeralPort == Tmf::kUdpPort);
```
bukepo, line 420:
```suggestion
    } while (IsPortInUse(mEphemeralPort));
```

superwhd: What about `IsPortReserved`? I guess it's not easy to implement a function returning ports being currently used.

simonlingoogle: I think both ports needs to be skipped:
1) Ports reserved: TMF port only
2) Ports in use

But this is out of the scope of this PR. I think this PR should focus on fixing this apparent bug. 
We can create another PR to enhance 2) later.

Thoughts? @wgtdkp @bukepo @superwhd

superwhd: I agree to focus on fixing the bug in this PR.

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #6700, opened 2021-06-02
[cli] adding helper `ProcessGet/Set()` to handle simpler commands

abtink:
This commit adds a set of helper methods in `Cli` to process simple
commands where we are getting and/or setting a single `uint` value.
This replaces common pattern in processing of commands where if there
are no arguments, we output the value we retrieve by calling a related
getter OT API, otherwise we parse the argument (to get the proper
`uint` value) and call a related setter OT API. For example, for
"channel" command, we use the `otLinkGetChannel()` as the getter and
`otLinkSetChannel()` as setter.

### DIFF
--- a/src/cli/cli.hpp
+++ b/src/cli/cli.hpp
@@ -68,6 +68,7 @@
 #include "common/code_utils.hpp"
 #include "common/debug.hpp"
 #include "common/instance.hpp"
+#include "common/type_traits.hpp"
 #include "utils/lookup_table.hpp"
 #include "utils/parse_cmdline.hpp"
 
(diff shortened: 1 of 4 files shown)


src/core/utils/parse_cmdline.hpp
```diff
@@ -468,6 +468,19 @@ class Arg
 
 #endif // OPENTHREAD_FTD || OPENTHREAD_MTD
 
+    /**
+     * This template method parses the argument as a specified value type.
+     *
+     * @tparam Type   The value type
```
jwhui, line 474:
```suggestion
     * @tparam Type   The value type.
```

src/cli/cli.hpp
```diff
@@ -313,6 +314,93 @@ class Interpreter
...
+    template <typename ValueType> using GetHandler         = ValueType (&)(otInstance *);
+    template <typename ValueType> using SetHandler         = void (&)(otInstance *, ValueType);
+    template <typename ValueType> using SetHandlerFailable = otError (&)(otInstance *, ValueType);
+
+    template <typename ValueType> otError ProcessGet(uint8_t aArgsLength, GetHandler<ValueType> aGetHandler)
+    {
+        static_assert(TypeTraits::IsSame<ValueType, uint8_t>::kValue ||
+                          TypeTraits::IsSame<ValueType, uint16_t>::kValue ||
+                          TypeTraits::IsSame<ValueType, uint32_t>::kValue,
+                      "ValueType must be uint8_t, uint16_t, or uint32_t");
```
jwhui, line 326: What about signed versions to support things like `otThreadGetParentPriority()` and `otThreadSetParentPriority()`?

abtink: I added support for `int` value types as well but we don't have currently many cases where we can use them.

jwhui (approved):

PR #6701, opened 2021-06-02
[srp-server] Choose a UDP port different from the last one on a reboot

superwhd:
This PR implements a mechanism to let an SRP server listen to a different UDP port on a reboot. 

- When an SRP server starts, it reads the UDP port used in the last time from `Settings`. Then the SRP server listens to the next port in the reserved range. 
- When SRP server successfully starts, store its UDP port in `Settings`.


superwhd:
Pushed a wrong branch. Closing this PR.

PR #6702, opened 2021-06-02
[srp-server] choose different UDP port from previous on reboot

superwhd:
This PR implements a mechanism to let an SRP server listen to a different UDP port on a reboot. 

- When an SRP server starts, it reads the UDP port used in the last time from `Settings`. Then the SRP server listens to the next port in the reserved range. 
- When SRP server successfully starts, store its UDP port in `Settings`.


src/core/net/srp_server.cpp
```diff
@@ -438,15 +438,32 @@ void Server::CommitSrpUpdate(Error                    aError,
...
+    {
+        port = info.GetPort() + 1;
+        if (port > kUdpPortMax)
+        {
+            port = kUdpPortMin;
+        }
+    }
+#endif
 
+    VerifyOrExit(!IsRunning());
```
abtink, line 456: How about doing `VerifyOfExit(!IsRunning())` at the top of the method and before reading the info from `Settings`?

src/core/config/srp_server.h
```diff
@@ -46,13 +46,37 @@
...
+ *
+ * Define to 1 to enable SRP server feature to save its own port in non-volatile settings.
+ *
+ * When enabled, the SRP server will save the its port in the non-volatile settings. On a server
+ * restart (e.g., due to a device reset) it will choose a UDP port different from the saved one. The info is written to
+ * Settings after an SRP server successfully starts.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_SRP_SERVER_SAVE_INFO
+#define OPENTHREAD_CONFIG_SRP_SERVER_SAVE_INFO 1
```
simonlingoogle, line 79:
```suggestion
```
always save info?

abtink (approved):
Looks great. Thanks @superwhd. 👍 
Few smaller suggestions below.

src/core/net/srp_server.cpp
```diff
@@ -438,15 +438,32 @@ void Server::CommitSrpUpdate(Error                    aError,
 
 void Server::Start(void)
 {
-    Error error = kErrorNone;
+    Error    error = kErrorNone;
+    uint16_t port  = kUdpPortMin;
 
-    VerifyOrExit(!IsRunning());
+#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE && OPENTHREAD_CONFIG_SRP_SERVER_SAVE_INFO
```
simonlingoogle, line 444:
```suggestion
```

OPENTHREAD_CONFIG_SRP_SERVER_ENABLE must be 1 in this file?

src/core/net/srp_server.cpp
```diff
@@ -438,15 +438,32 @@ void Server::CommitSrpUpdate(Error                    aError,
...
+    Settings::SrpServerInfo info;
+    if (Get<Settings>().Read(info) == kErrorNone)
+    {
+        port = info.GetPort() + 1;
+        if (port > kUdpPortMax)
```
simonlingoogle, line 449: What if port < kUdpPortMin?

src/core/net/srp_server.cpp
simonlingoogle, line 456: Choose port after this line of code?

src/core/net/srp_server.cpp
```diff
@@ -438,15 +438,32 @@ void Server::CommitSrpUpdate(Error                    aError,
...
+    VerifyOrExit(!IsRunning());
     SuccessOrExit(error = mSocket.Open(HandleUdpReceive, this));
-    SuccessOrExit(error = mSocket.Bind(kUdpPort, OT_NETIF_THREAD));
+    SuccessOrExit(error = mSocket.Bind(port, OT_NETIF_THREAD));
 
     SuccessOrExit(error = PublishServerData());
 
+#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE && OPENTHREAD_CONFIG_SRP_SERVER_SAVE_INFO
+    info.SetPort(port);
+    IgnoreError(Get<Settings>().Save(info));
```
simonlingoogle, line 464: Save the port when the first SRP service is registered?

@librasungirl

src/core/net/srp_server.hpp
```diff
@@ -78,7 +78,8 @@ class Server : public InstanceLocator, private NonCopyable
 public:
     enum : uint16_t
     {
-        kUdpPort = OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT, ///< The SRP Server UDP listening port.
+        kUdpPortMin = OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT_MIN, ///< The reserved min SRP Server UDP listening port.
+        kUdpPortMax = OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT_MAX, ///< The reserved max SRP Server UDP listening port.
```
simonlingoogle, line 82: `static_assert` that `kUdpPortMin <= kUdpPortMax` ?

tests/scripts/thread-cert/test_srp_server_reboot_port.py
```diff
@@ -0,0 +1,125 @@
...
+import command
+import thread_cert
+
+# Test description:
+#   This test verifies when an SRP server reboots, it will listen to a UDP port
+#   that wasn't used in the last time.
+#
+# Topology:
+#
+#   CLIENT (leader) -- SERVER (router)
```
simonlingoogle, line 43:
```suggestion
#   CLIENT (Leader) -- SERVER (Router)
```

tests/scripts/thread-cert/test_srp_server_reboot_port.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class SrpAutoStartMode(thread_cert.TestCase):
+    USE_MESSAGE_FACTORY = False
+    SUPPORT_NCP = False
+
+    TOPOLOGY = {
+        CLIENT: {
+            'name': 'SRP_CLIENT',
+            'masterkey': '00112233445566778899aabbccddeeff',
```
simonlingoogle, line 57:
```suggestion
```

tests/scripts/thread-cert/test_srp_server_reboot_port.py
```diff
@@ -0,0 +1,125 @@
...
+            'mode': 'rdn',
+        },
+        SERVER: {
+            'name': 'SRP_SERVER',
+            'masterkey': '00112233445566778899aabbccddeeff',
```
simonlingoogle, line 62:
```suggestion
```

tests/scripts/thread-cert/test_srp_server_reboot_port.py
```diff
@@ -0,0 +1,125 @@
...
+        self.assertEqual(client.srp_client_get_auto_start_mode(), 'Enabled')
+        self.simulator.go(2)
+
+        self.assertEqual(client.srp_client_get_state(), 'Enabled')
+        self.assertTrue(server.has_ipaddr(client.srp_client_get_server_address()))
+
+        ports = [server.get_srp_server_port()]
+
+        # Repeat the following steps for 20 times
+        for i in range(25):
```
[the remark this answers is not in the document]
simonlingoogle: We'd better define a constant in local file.

tests/scripts/thread-cert/test_srp_server_reboot_port.py
```diff
@@ -0,0 +1,125 @@
...
+
+            #
+            # 3. Enable server and check client starts again. Verify that the
+            # server is using a different port.
+
+            server.srp_server_set_enabled(True)
+            self.simulator.go(5)
+            self.assertEqual(client.srp_client_get_state(), 'Enabled')
+            self.assertEqual(client.srp_client_get_server_address(), server.get_mleid())
+            self.assertNotEqual(old_port, server.get_srp_server_port())
```
simonlingoogle, line 119: I think we also need to verify that the SRP client will re-register its SRP service to the SRP server after SRP server is reset, which is the real purpose of this change.

src/core/net/srp_server.cpp
abtink: 👍  very good point @simonlingoogle. We need to handle such a case (i.e., the config min/max port range may be different between version that saved the value vs the one reading it later).

src/core/config/srp_server.h
superwhd: According to offline discussion, we decided to keep this macro.

tests/scripts/thread-cert/test_srp_server_reboot_port.py
superwhd: I switched to a style that most `test_srp*` files used.

src/core/config/srp_server.h
simonlingoogle: SGTM.

simonlingoogle, line 79:
```suggestion
#define OPENTHREAD_CONFIG_SRP_SERVER_PORT_SWITCH_ENABLE 1
```
I think the configuration name should be what the feature is, rather than how we implemented it?

src/core/net/srp_server.cpp
```diff
@@ -438,15 +438,32 @@ void Server::CommitSrpUpdate(Error                    aError,
...
+#if OPENTHREAD_CONFIG_SRP_SERVER_SAVE_INFO
+    Settings::SrpServerInfo info;
+    if (Get<Settings>().Read(info) == kErrorNone)
+    {
+        port = info.GetPort() + 1;
+        if (port < kUdpPortMin || port > kUdpPortMax)
+        {
+            port = kUdpPortMin;
+        }
+    }
```
simonlingoogle, line 455:
```suggestion
{
    Settings::SrpServerInfo info;
    
    if (Get<Settings>().Read(info) == kErrorNone)
    {
        port = info.GetPort() + 1;
        if (port < kUdpPortMin || port > kUdpPortMax)
        {
            port = kUdpPortMin;
        }
    }
}
```
Variable must be at the front, so...

tests/scripts/thread-cert/test_srp_server_reboot_port.py
```diff
@@ -0,0 +1,180 @@
...
+#
+# Topology:
+#
+#   LEADER (SRP client) -- ROUTER (SRP server)
+#
+
+CLIENT = 1
+SERVER = 2
+
+REBOOT_TIMES = 25
```
simonlingoogle, line 49:
```suggestion
REBOOT_TIMES = 3
```
maybe we don't need to reboot so many times?

simonlingoogle (commented):
SGTM overall. 

But we need to decide when is the best time to save the port. 

@librasungirl has proposed to save the port when the first SRP client successfully registered a service, which I agree is the best strategy. 

Thoughts?

@superwhd @librasungirl @jwhui @wgtdkp @abtink

tests/scripts/thread-cert/test_srp_server_reboot_port.py
superwhd: I meant to let it go over all reserved ports and begin again from `kUdpPortMin`.

include/openthread/platform/settings.h
```diff
@@ -72,6 +72,7 @@ enum
     OT_SETTINGS_KEY_ON_LINK_PREFIX       = 0x000a, ///< On-link prefix for infrastructure link.
     OT_SETTINGS_KEY_SRP_ECDSA_KEY        = 0x000b, ///< SRP client ECDSA public/private key pair.
     OT_SETTINGS_KEY_SRP_CLIENT_INFO      = 0x000c, ///< The SRP client info (selected SRP server address).
+    OT_SETTINGS_KEY_SRP_SERVER_INFO      = 0x000d, ///< The SRP server info (address and port).
```
wgtdkp, line 75:
```suggestion
    OT_SETTINGS_KEY_SRP_SERVER_INFO      = 0x000d, ///< The SRP server info (UDP port).
```

wgtdkp (approved):
LGTM :+1: Thanks!

include/openthread/platform/settings.h
superwhd: Good catch! :+1:

superwhd:
> SGTM overall.
> 
> But we need to decide when is the best time to save the port.
> 
> @librasungirl has proposed to save the port when the first SRP client successfully registered a service, which I agree is the best strategy.
> 
> Thoughts?
> 
> @superwhd @librasungirl @jwhui @wgtdkp @abtink

So far I'm still inclined to save the port when the SRP server starts. Reasons: 1) We have reserved 20 ports for this purpose so it's rare that all ports are used up in a short period of time. 2) It takes a bit more efforts to implement 'save the port when the first service registered' and the logic has to be added at two places, which makes it a bit harder to maintain.

wgtdkp:
> SGTM overall.
> 
> But we need to decide when is the best time to save the port.
> 
> @librasungirl has proposed to save the port when the first SRP client successfully registered a service, which I agree is the best strategy.
> 
> Thoughts?
> 
> @superwhd @librasungirl @jwhui @wgtdkp @abtink

I agree with @superwhd that there are not many benefits for saving only after the first service registration. Changing the port when there is no service doesn't introduce burdens since there is no SRP client. But saving port at start makes the code simpler and more reasonable.

simonlingoogle:
> > SGTM overall.
> > But we need to decide when is the best time to save the port.
> > @librasungirl has proposed to save the port when the first SRP client successfully registered a service, which I agree is the best strategy.
> > Thoughts?
> > @superwhd @librasungirl @jwhui @wgtdkp @abtink
> 
> I agree with @superwhd that there are not many benefits for saving only after the first service registration. Changing the port when there is no service doesn't introduce burdens since there is no SRP client. But saving port at start makes the code simpler and more reasonable.

We are comparing 2 strategies:
1. Current: Save port when the `SRP service TLV` is added to the local network data
2. Alternative: Save the port when a SRP service is registered by the first SRP client. 

Basically the `Alternative` strategy delays saving port until it has made sure that the `SRP service TLV` has been successfully added to the Leader's network data because some Thread device has been aware of the change. 

I think the `Alternative` solution has advantages:
1. When the reserved port range is small (e.g. 2 ports), the BR is less possible to use the same port after multiple fast reboots (SRP server will keep using the same port if there is no SRP registration). 
2. When there are SED/SSED children, the BR will not save the port until a SED child has registered a SRP service, which reduces the chance that multiple reboots may cause BR to switch back to the original port. 
3. After #6587 is merged, the SRP Server can no longer guarantee that its SRP service TLV will eventually be added the Network Data. For example, if there are 3 SRP servers, only 2 SRP servers will adds their SRP service TLV to the network data. The one that didn't add its SRP service TLV to the Network Data should better not change port.

Again, I don't see a significant difference between the two strategies. But the `Alternative` solution seems to be more robust in some corner cases. So, I believe we should implement it if it's not too complex.

Thoughts? @superwhd @wgtdkp @librasungirl

src/core/config/srp_server.h
```diff
@@ -46,13 +46,37 @@
...
+ *
+ * Specifies the max port number in the port range reserved for SRP server.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT_MAX
+#define OPENTHREAD_CONFIG_SRP_SERVER_UDP_PORT_MAX 53554
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_SRP_SERVER_SAVE_INFO
```
librasungirl, line 69:
```suggestion
 * @def OPENTHREAD_CONFIG_SRP_SERVER_PORT_SWITCH_ENABLE
```

src/core/config/srp_server.h
```diff
@@ -46,13 +46,37 @@
...
+ *
+ * Define to 1 to enable SRP server feature to save its own port in non-volatile settings.
+ *
+ * When enabled, the SRP server will save the its port in the non-volatile settings. On a server
+ * restart (e.g., due to a device reset) it will choose a UDP port different from the saved one. The info is written to
+ * Settings after an SRP server successfully starts.
```
librasungirl, line 75:
```suggestion
 * When enabled, the SRP server will save its port in the non-volatile settings. On a server
 * restart (e.g., due to a device reset) it will restore the port and change to a different one. The info is written to
 * Settings after the SRP server successfully starts.
```

librasungirl:
Agree with @simonlingoogle 's arguments. The optimized `Alternative` option may also be good to the lifetime of the flash as it only stores when necessary, reducing the write/erase operation.

Fine to me to have the optimization in separate PR.

src/core/common/settings.hpp
```diff
@@ -677,6 +678,48 @@ class SettingsBase : public InstanceLocator
...
+         *
+         */
+        void SetPort(uint16_t aPort) { mPort = Encoding::LittleEndian::HostSwap16(aPort); }
+
+    private:
+        void Log(Action aAction) const;
+
+        uint16_t mPort; // (in little-endian encoding)
+    } OT_TOOL_PACKED_END;
+#endif // OPENTHREAD_CONFIG_SRP_SERVER_ENABLE && OPENTHREAD_CONFIG_SRP_SERVER_SAVE_INFO
```
librasungirl, line 721:
```suggestion
#endif // OPENTHREAD_CONFIG_SRP_SERVER_ENABLE && OPENTHREAD_CONFIG_SRP_SERVER_PORT_SWITCH_ENABLE
```

superwhd:
Changed the logic to save the port number when the first port is registered. Also added a test case to verify that the port number won't change if there's no service registered in the last run.

src/core/net/srp_server.hpp
```diff
@@ -672,6 +674,7 @@ class Server : public InstanceLocator, private NonCopyable
 
     ServiceUpdateId mServiceUpdateId;
     bool            mEnabled;
+    bool            mHasRegisteredAnyService;
```
simonlingoogle, line 677:
```suggestion
    bool            mEnabled: 1;
    bool            mHasRegisteredAnyService: 1;
```
nit

src/core/net/srp_server.cpp
```diff
@@ -420,6 +421,15 @@ void Server::CommitSrpUpdate(Error                    aError,
...
         otLogInfoSrp("[server] add new host %s", aHost.GetFullName());
         AddHost(&aHost);
+#if OPENTHREAD_CONFIG_SRP_SERVER_PORT_SWITCH_ENABLE
+        if (!mHasRegisteredAnyService)
+        {
+            mHasRegisteredAnyService = true;
+            Settings::SrpServerInfo info;
+            info.SetPort(mSocket.mSockName.mPort);
+            IgnoreError(Get<Settings>().Save(info));
+        }
```
simonlingoogle, line 431:
```suggestion
        {
            Settings::SrpServerInfo info;

            mHasRegisteredAnyService = true;
            info.SetPort(mSocket.mSockName.mPort);
            IgnoreError(Get<Settings>().Save(info));
        }
```
style

simonlingoogle (approved):
LGTM :+1: 

with nits

librasungirl (approved):
👍

src/core/config/srp_server.h
```diff
@@ -46,13 +46,37 @@
...
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_SRP_SERVER_PORT_SWITCH_ENABLE
+ *
+ * Define to 1 to enable SRP server feature to save its own port in non-volatile settings.
+ *
+ * When enabled, the SRP server will save its port in the non-volatile settings. On a server
+ * restart (e.g., due to a device reset) it will restore the port and change to a different one. The info is written to
+ * Settings after the SRP server successfully starts.
```
jwhui, line 75:
```suggestion
 * When enabled, the SRP server will save its port in the non-volatile settings. On a server
 * restart (e.g., due to a device reset) it will restore the port and change to a different one. The info is written to
 * Settings after the first service has been registered due to receiving an SRP Update.
```

jwhui (commented):
Looks great overall. One minor suggestion to update the doc to reflect the latest implementation.

src/core/net/srp_server.cpp
```diff
@@ -438,13 +449,27 @@ void Server::CommitSrpUpdate(Error                    aError,
...
-    Error error = kErrorNone;
+    Error    error = kErrorNone;
+    uint16_t port  = kUdpPortMin;
 
     VerifyOrExit(!IsRunning());
 
-    SuccessOrExit(error = mSocket.Open(HandleUdpReceive, this));
-    SuccessOrExit(error = mSocket.Bind(kUdpPort, OT_NETIF_THREAD));
+    {
+#if OPENTHREAD_CONFIG_SRP_SERVER_PORT_SWITCH_ENABLE
```
abtink, line 458: Move the braces inside the `#if` block?

abtink (approved):
LGTM. Thanks.

jwhui (approved):

PR #6703, opened 2021-06-02
[mac] new public API to validate and set `otNetworkName`

abtink:
This commit adds a new public API `otNetworkNameSet()` to set an
`otNetworkName` instance from a given null terminated C string. The new
API will also validate that the given name string follows UTF-8
encoding and its length is not longer than max network name size.

This commit also simplifies the UTF-8 validation of `NetworkName` by
moving the check from `Mac::SetNetworkName()` to a newly added method
in `Mac::NetworkName` to set it from a C string.

It also defines `DomainName` as a `typedef` of `NetworkName` which
helps simplify its use and remove repeated code.

Finally, it updates the `Cli::Dataset` to use the new public API for
validating `otNetworkName`.

### DIFF
--- a/include/openthread/dataset.h
+++ b/include/openthread/dataset.h
@@ -75,10 +75,12 @@ typedef struct otMasterKey otMasterKey;
 /**
  * This structure represents a Network Name.
  *
+ * The `otNetworkName` is a null terminated C string (i.e., `m8` char array MUST end with null char `\0`).
+ *
  */
 typedef struct otNetworkName
 {
-    char m8[OT_NETWORK_NAME_MAX_SIZE + 1]; ///< Byte values
+    char m8[OT_NETWORK_NAME_MAX_SIZE + 1]; ///< Byte values. The `+ 1` is for null char.
 } otNetworkName;
 
 #define OT_EXT_PAN_ID_SIZE 8 ///< Size of a Thread PAN ID (bytes)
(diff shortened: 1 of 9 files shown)


include/openthread/dataset.h
```diff
@@ -515,6 +517,21 @@ otError otDatasetGeneratePskc(const char *           aPassPhrase,
...
+ * `OT_NETWORK_NAME_MAX_SIZE`.
+ *
+ * @param[out] aNetworkName        A pointer to the `otNetworkName` to set.
+ * @param[in]  aNameString         A name C string.
+ *
+ * @retval OT_ERROR_NONE           Successfully set @p aNetworkName from @p aNameString.
+ * @retval OT_ERROR_INVALID_ARGS   @p aNameStrng is invalid (too long or does not follow UTF-8 encoding).
+ *
+ */
+otError otNetworkNameSet(otNetworkName *aNetworkName, const char *aNameString);
```
jwhui, line 533: We have other similar APIs that use `FromString()` rather than `Set()`. Thoughts?
```suggestion
otError otNetworkNameFromString(otNetworkName *aNetworkName, const char *aNameString);
```

abtink: Thanks for suggestion. I also like `FromString()` name better. Changed it in latest push.

jwhui (approved):

abtink:
@jwhui , just noticed in my branch that I forgot to update the commit message to use  the new `otNetworkNameFromString()` and saw that you already fixed it when merging it. Thank you 😃 😃

PR #6704, opened 2021-06-04
[mac] avoid potential assert with SetKeyId() on an invalid frame

jwhui:
https://github.com/openthread/openthread/blob/f3671d5a758a3655fb9533986c297ffc190ce080/src/core/mac/mac_frame.cpp#L582


abtink (approved):
Nice find. 👍

PR #6706, opened 2021-06-04
Disabling OPENTHREAD_CONFIG_LOG_MAC makes otbr-agent processing faster

dougleaf:
We noticed an issue with OPENTHREAD_CONFIG_LOG_MAC when running OTBR on slower host processors (when compared to a RPi 3B+). These logs would add up to 7ms on every HDLC-lite frame processing time, heavily affecting BR throughput. It seems the issue is masked on RPis.

Thanks @jwhui for pointing us to the right direction when looking at this problem.

My question is whether all the low level logs listed in this configuration file shouldn't be disabled by default.

@AliakbarEski @gabekassel @ndyck14

### DIFF
--- a/src/core/config/logging.h
+++ b/src/core/config/logging.h
@@ -180,7 +180,7 @@
  *
  */
 #ifndef OPENTHREAD_CONFIG_LOG_MAC
-#define OPENTHREAD_CONFIG_LOG_MAC 1
+#define OPENTHREAD_CONFIG_LOG_MAC 0
 #endif
 
 /**


dougleaf:
@googlebot I signed it!

simonlingoogle:
What log level can cause this issue? @dougleaf 

If there is some logging that are extremely slow, should we remove these log instead?

src/core/config/logging.h
```diff
@@ -180,7 +180,7 @@
  *
  */
 #ifndef OPENTHREAD_CONFIG_LOG_MAC
-#define OPENTHREAD_CONFIG_LOG_MAC 1
+#define OPENTHREAD_CONFIG_LOG_MAC 0
```
abtink, line 183: @dougleaf , thanks for this PR.

I don't favor disabling the default for a specific log reagion (e.g. `LOG_MAC` region)  in the oepnthread repo.  I think the default should be the same for all different log regions.

 It can be of course controlled by application/product specifc OT config files (if there is a need).

In general, I suggest using "log level" insetad. Here is high-level use (in some of our internal project/products)
- `LOG_LEVEL_DEBG` is mainly used in special build by developers/engineers for debugging only (it's really verbose)
   - I recommend using `DEBG` level only in special case/build  (e.g. debugging radio driver issues, etc).
- `LOG_LEVEL_INFO` or `LOG_LEVEL_NOTE` are commonly used during development/intergation of OT and for in-field testing:
   - `NOTE` level commonly logs recoverable failure cases (e.g. dropping a message, failing to recv ack after retx, etc).
   -  `INFO` provides more info and logs even successful interactions (e.g. successful msg send or receive) - which are useful during integration/debugging (this is in addition to `NOTE` level logs).
   
- On release firmware, the log level is decided based on capability of device. On more capable devices `INFO` or `NOTE` may be used but often the logs are fully disabled, or `WARN` level is used.
- We also have dynamic log level feature which when used, allows log level to be changed during run-time.

dougleaf: @simonlingoogle @abtink you are right. When building for release, CMake sets it to `LOG_LEVEL_INFO`, which prevents the low level logs. We were previously on CMake build type debug, which sets LOG_LEVEL to `LOG_LEVEL_DEBG`.

I'll close the PR as invalid. Interestingly we never saw these logs and thus took longer to understand the cause. Possibly because `OT_LOG_OUTPUT` was set to `NONE`

Issue #6665, opened 2021-05-21
Logging is printed across multiple lines


gabekassel:
**Describe the bug** 
Logging is unusually printed broken up across many lines.
```
May 21 17:29:46.443 user.info otbr-agent[2080]: [INFO]-PLAT----: ae
May 21 17:29:46.443 user.info otbr-agent[2080]: [INFO]-PLAT----: 45
May 21 17:29:46.443 user.info otbr-agent[2080]: [INFO]-PLAT----: c0
May 21 17:29:46.443 user.info otbr-agent[2080]: [INFO]-PLAT----: 01
May 21 17:29:46.443 user.info otbr-agent[2080]: [INFO]-PLAT----:  s
May 21 17:29:46.443 user.info otbr-agent[2080]: [INFO]-PLAT----:  1c00
May 21 17:29:46.443 user.info otbr-agent[2080]: [INFO]-PLAT----: ^M
May 21 17:29:46.443 user.info otbr-agent[2080]: [INFO]-PLAT----: Done
May 21 17:29:46.443 user.info otbr-agent[2080]: [INFO]-PLAT----: ^M
May 21 17:29:46.443 user.info otbr-agent[2080]: [INFO]-PLAT----: >
```

**To Reproduce** Information to reproduce the behavior, including:

1. Git commit id https://github.com/openthread/ot-br-posix/commit/f43ccf60be2bb65723c9216fed17874ca2cf79f9
2. IEEE 802.15.4 hardware platform
3. Build steps
4. Network topology

jwhui:
This appears to be due to #6639.

abtink:
@bukepo looking at the code in `daemon.cpp` I noticed a pattern that I want to bring up:
- I think we can have a conifguration to have logging provided by the CLI app itself
- In this case, it will use `OutputFormatV` to implement base logging platform API  from `otCliPlatLogv()` and 
`otCliPlatLogLine()`
- But then in `OutputFormatV()` implementation in `daemon.cpp` we can use `otLog{Level}Plat()` to log.
- Can this then create a recursive call loop (for such a config)?
- Should we check and not use `otLog()` in `OutputFormatV()` fo such a config?

bukepo:
[...]

PR #6709, opened 2021-06-06
[posix] fix issue with logging CLI output

abtink:
This commit reverts the change from PR #6639 which added logging
of the CLI output. Note that the CLI output can be generated and
outputted in smaller pieces (not as a whole line) so the model
from PR #6639 would emit each piece as a different log line which
is not desirable.

----

This is to address Issue https://github.com/openthread/openthread/issues/6665.


bukepo (approved):

Irving-cl (approved):
LGTM

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6712, opened 2021-06-08
[posix] increase default address cache size from 32 to 128

ndyck14:
Having this value sufficiently large is important to reduce latency when multiple simultaneous requests come in to a BR for many if not all the nodes on the network. This could happen when a client reconnects on Wifi and needs to query and/or update the state for many nodes at the same time.


gabekassel:
@jwhui @abtink @simonlingoogle do you know how this change impacts host memory consumption? curious what rationale went into the original default. seems previously chosen for posix platform in #6044

jwhui:
> @jwhui @abtink @simonlingoogle do you know how this change impacts host memory consumption? curious what rationale went into the original default. seems previously chosen for posix platform in #6044

@gabekassel , each cache entry consumes 32 bytes. The current default of 32 cache entries is a carryover from our legacy Nest products that generally assumed a modest maximum network size. I agree that increasing the cache size on platforms that have the available resources makes sense.

abtink (approved):
LGTM.

@gabekassel, @ndyck14, there are couple of other configs that I highly recommend to be enabled on a BR  (or router) style device/products:
- For cache table
  -  Suggest increasing `OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_MAX_SNOOP_ENTRIES`  as well (helps allow more of the 128 to be used for snoop optimization)
  - Also the default value for `OPENTHREAD_CONFIG_TMF_ADDRESS_QUERY_MAX_RETRY_DELAY`  is rather long (8 hours which I think is given in spec). We typically use much shorter values in internal projects/products.
 

Other feature to enable on routers/BRs:
- `OPENTHREAD_CONFIG_MLE_SEND_LINK_REQUEST_ON_ADV_TIMEOUT` 
  - This helps make the links between router a lot more reliable).
- I also highly recommend `OPENTHREAD_CONFIG_CHILD_SUPERVISION_ENABLE` on router (and on SEDs). 
- And increasing number of message buffers (or heap size if heap is being used for msg buffers).

On SED type devices, I recommend:
- `OPENTHREAD_CONFIG_MLE_INFORM_PREVIOUS_PARENT_ON_REATTACH`
- `OPENTHREAD_CONFIG_MLE_ATTACH_BACKOFF_ENABLE` (which is enabled by default).

ndyck14:
@googlebot I signed it!

jwhui:
> LGTM.
> 
> @gabekassel, @ndyck14, there are couple of other configs that I highly recommend to be enabled on a BR (or router) style device/products:
> 
> * For cache table
>   
>   * Suggest increasing `OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_MAX_SNOOP_ENTRIES`  as well (helps allow more of the 128 to be used for snoop optimization)
>   * Also the default value for `OPENTHREAD_CONFIG_TMF_ADDRESS_QUERY_MAX_RETRY_DELAY`  is rather long (8 hours which I think is given in spec). We typically use much shorter values in internal projects/products.
> 
> Other feature to enable on routers/BRs:
> 
> * `OPENTHREAD_CONFIG_MLE_SEND_LINK_REQUEST_ON_ADV_TIMEOUT`
>   
>   * This helps make the links between router a lot more reliable).
> * I also highly recommend `OPENTHREAD_CONFIG_CHILD_SUPERVISION_ENABLE` on router (and on SEDs).
> * And increasing number of message buffers (or heap size if heap is being used for msg buffers).
> 
> On SED type devices, I recommend:
> 
> * `OPENTHREAD_CONFIG_MLE_INFORM_PREVIOUS_PARENT_ON_REATTACH`
> * `OPENTHREAD_CONFIG_MLE_ATTACH_BACKOFF_ENABLE` (which is enabled by default).

@abtink , I agree that we should encourage the recommended configuration that we have found to work well in practice. However, I think we can introduce those in separate PRs.

The `OPENTHREAD_CONFIG_MLE_SEND_LINK_REQUEST_ON_ADV_TIMEOUT` feature is now incorporated into the latest Thread Specification. I just submitted #6713 to address that one.

ndyck14:
@jwhui i believe our CLA should now be in place - i did an organization level. not sure if google bot isnt picking up for some reason.

jwhui:
> @jwhui i believe our CLA should now be in place - i did an organization level. not sure if google bot isnt picking up for some reason.

@ndyck14 , it can take a few days to process the Google CLA form submission. Apologies for the delay.

jwhui:
@googlebot kick

jwhui:
> @ndyck14 , it can take a few days to process the Google CLA form submission. Apologies for the delay.

@ndrake , Google CLA check is now passing. Thanks!

jwhui:
> @ndyck14 , it can take a few days to process the Google CLA form submission. Apologies for the delay.

@ndyck14 , Google CLA check is now passing. Thanks!

simonlingoogle (commented):
Should we also adjust `OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_MAX_SNOOP_ENTRIES` ?

src/posix/platform/openthread-core-posix-config.h
```diff
@@ -153,7 +153,7 @@
  *
  */
 #ifndef OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_ENTRIES
-#define OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_ENTRIES 32
+#define OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_ENTRIES 128
```
abtink, line 156: How about changing `OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_MAX_SNOOP_ENTRIES` here as well?

Suggest using some value in ~25 range.
```
#define OPENTHREAD_CONFIG_TMF_ADDRESS_CACHE_MAX_SNOOP_ENTRIES  25
```

Basically in default we had 10 total entries, and allowed 2 of them to be used for snoop optimization. If we scale up the total number of entry to 128, we can allow ~20 for snoop (keeping ratio roughly same).

Snoop optimization can help reduce the number of address queries in general.

simonlingoogle (approved):
LGTM :+1:

src/posix/platform/openthread-core-posix-config.h
abtink: The traffic pattern that matter is when the node sending a (request) message to some entity outside of Thread mesh and then expecting to receive a response from it. This all goes through a BR.

With snoop optimization, the BR will create a snoop cache entry when it sees the the request message and then does not need to perform address query when it gets the response to forward to the node. 

In the design, we limit the number snoop entries (which is the `ADDRESS_CACHE_MAX_SNOOP_ENTRIES`) to ensure snooped entries do not overwrite the other cache entries.  If a snooped entry is used to resolve, the entry become a normal cache entry.  We keep an snooped entry around (don't allow it to reclaimed) for a short time specified by `SNOOP_CACHE_ENTRY_TIMEOUT` to give a chance for it to be used.

Keeping the snoop entries to default of `2` will still work fine, but if we increase the total number of cache from 10 (or 32) to 128, it make sense to me to also scale up the number of snoop entries (to gain the benefit from it).

Here is how I would choose: If we think we want `x` cache entries in stead state (based on the use-case), we can use `1.2 * x` as the total and allow `0.2 * x` to be used as snoop entries.

------

This feature and other such optimizations (which are not really part of Thread spec but enhancement added in OpenThread implementation) typically originate from performance observations on the deployed Thread networks in our internal products/projects.

ndyck14: seems this then become a question of how much un-acked traffic is travelling off the mesh. That's the only case that would create a potentially negative scenario from having the cache_snoop entries higher (i don't believe any of this matters if devices on mesh < max_cache).

I'd personally vote for that value to be based around updated testing on real world scenarios (particularly, with devices >128). Our testing at the moment does not touch this case.

ndyck14: or at least the change should originate from someone else who has done real world testing.

abtink: @ndyck14 If you don't feel comfortable changing snoop config, it's ok. No worries. Our team members can update this later. 😃 

My intention is to help share our experience from integrating and deploying Thread in different Nest/Google products 
and point out which configs can be useful in different scenarios.

---------

> seems this then become a question of how much un-acked traffic is travelling off the mesh. That's the only case that would create a potentially negative scenario from having the cache_snoop entries higher (i don't believe any of this matters if devices on mesh < max_cache).

Note that frames are acked at every hop at Mac layer. If a frame is sent across multiple hops over Thread mesh to a BR, the BR will not sent an ack to the original sender.

As I said, Thread network should behave fine even if snoop optimization is not used (i.e., setting the `MAX_SNOOP_ENTRIES` to zero). Using the snoop optimization will help reduce the delay in the (first) interaction (request response exchange) originating from a Thread device to an off-mesh entity (server). With snoop optimization when BR receives the response from server to forward within mesh, it does not need to perform address query for the mesh destination and can use the snooped cache entry.

ndyck14: thanks @abtink , I definitely appreciate the suggestions. If you guys can add a separate PR, that would be preferrable. We'll definitely circle back to this once we have feature sets that reach out off-mesh proactively, which should be around Duckhorn timeframe.

jwhui (approved):

PR #6715, opened 2021-06-10
[tests] remove unnecessary host SLAAC assertions

wgtdkp:
The assertion is verifying the (linux) host SLAAC behavior which is not guaranteed on different
platform and linux distributions. Moreover, it is required for testing the BR behavior.


jwhui (approved):
Thanks! 👍🏼

PR #6717, opened 2021-06-10
[cmake] always use git hash for default package version

jwhui:


doru91:
Tested it on NXP K32W061 and indeed, the version command contains the commit id of the openthread core stack. However, I think it would be useful to also include the commit id of the vendor repo that was used for compilation. What would be the easiest way for achieving this?

jwhui:
> Tested it on NXP K32W061 and indeed, the version command contains the commit id of the openthread core stack. However, I think it would be useful to also include the commit id of the vendor repo that was used for compilation. What would be the easiest way for achieving this?

@doru91 , I just abstracted it into a separate function so that the parent project can also make the same call. After merging, I can update the platform-specific repos to extract the git hash of the parent repo.

bukepo (approved):

PR #6719, opened 2021-06-11
[srp-client] add feature for switching server on failure

abtink:
This commit adds new feature in SRP client to switch server on failure
(when auto-start feature is used). When enabled, the client will try
to select the next server from entries in Thread Network Data when
the current server rejects an SRP update or there are timeouts
(no response) after multiple attempts.

### DIFF
--- a/src/core/net/srp_client.cpp
+++ b/src/core/net/srp_client.cpp
@@ -35,6 +35,7 @@
 #include "common/instance.hpp"
 #include "common/locator_getters.hpp"
 #include "common/logging.hpp"
+#include "common/numeric_limits.hpp"
 #include "common/random.hpp"
 #include "common/settings.hpp"
 #include "common/string.hpp"
@@ -167,6 +168,9 @@ Client::Client(Instance &aInstance)
     , mAutoStartCallback(nullptr)
     , mAutoStartContext(nullptr)
     , mServerSequenceNumber(0)
+#if OPENTHREAD_CONFIG_SRP_CLIENT_SWITCH_SERVER_ON_FAILURE
+    , mTimoutFailureCount(0)
+#endif
 #endif
     , mDomainName(kDefaultDomainName)
     , mTimer(aInstance, Client::HandleTimer)
@@ -215,7 +219,7 @@ exit:
     return error;
 }
 
-void Client::Stop(Requester aRequester)
+void Client::Stop(Requester aRequester, StopMode aMode)
 {
     // Change the state of host info and services so that they are
     // added/removed again once the client is started back. In the
@@ -247,14 +251,28 @@ void Client::Stop(Requester aRequester)
     ChangeHostAndServiceStates(kNewStateOnStop);
 
     IgnoreError(mSocket.Close());
+
     mShouldRemoveKeyLease = false;
     mTxFailureRetryCount  = 0;
-    ResetRetryWaitInterval();
+
+    if (aMode == kResetRetryInterval)
+    {
+        ResetRetryWaitInterval();
+    }
+
     SetState(kStateStopped);
 
 #if OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE
-    VerifyOrExit((aRequester == kRequesterAuto) && (mAutoStartCallback != nullptr));
-    mAutoStartCallback(nullptr, mAutoStartContext);
+#if OPENTHREAD_CONFIG_SRP_CLIENT_SWITCH_SERVER_ON_FAILURE
+    mTimoutFailureCount = 0;
+#endif
+
+    mAutoStartDidSelectServer = false;
+
+    if ((aRequester == kRequesterAuto) && (mAutoStartCallback != nullptr))
+    {
+        mAutoStartCallback(nullptr, mAutoStartContext);
+    }
 #endif
 
 exit:
@@ -1137,6 +1155,10 @@ void Client::ProcessResponse(Message &aMessage)
 
     otLogInfoSrp("[client] Received response");
 
+#if OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE && OPENTHREAD_CONFIG_SRP_CLIENT_SWITCH_SERVER_ON_FAILURE
+    mTimoutFailureCount = 0;
+#endif
+
     error = Dns::Header::ResponseCodeToError(header.GetResponseCode());
 
     if (error != kErrorNone)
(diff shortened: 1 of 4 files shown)


simonlingoogle (commented):
I have concerns regarding this feature:
- SRP clients will detect `network data` change and switch to another server if the BR is off for a few minutes. So, we already have some form of auto-switching feature.  
- If TREL is enabled on BRs, reachability to one BR basically means reachability to all BRs. Which means SRP clients won't have many reachability issues with BRs.
- If a SRP client switches to another server, the origianl server, if running, still has the SRP services. The SRP client may fail to register on the new server due to probing failures.

It seems unclear to me what's the problem that switching is trying to solve.

abtink:
@simonlingoogle , please see below.

> It seems unclear to me what's the problem that switching is trying to solve.

This is adding recovery mechanism on SRP client for a misbehaving server. We don't want to stay stuck with a server that never responds to any update message.

> If a SRP client switches to another server, the origianl server, if running, still has the SRP services. The SRP client may fail to register on the new server due to probing failures.

The "switch server logic" added in this PR is only used when nothing is yet registered with any server. There is a guard check for this in `SelectNextServer()` method.

> Occasional bust of channel failures may cause registration failures. In such case, it's not useful to switch to another server. SRP clients should keep using the current SRP server in such case.

The switch happens only after certain (configurable) number of back-to-back attempts and timeout failures.

src/core/net/srp_client.cpp
```diff
@@ -1656,6 +1714,90 @@ void Client::ProcessAutoStart(void)
...
+    // Data happens to be changed and the entry removed but
+    // the "changed" event from `Notifier` may have not yet been
+    // processed (note that events are emitted from their own
+    // tasklet). In such a case we keep `serverSockAddr` as empty.
+
+exit:
+    if (!serverSockAddr.GetAddress().IsUnspecified())
+    {
+        mHostInfo.SetState(kToAdd);
+        Stop(kRequesterAuto, kKeepRetryInterval);
```
simonlingoogle, line 1795: Why is the logic ` mHostInfo.SetState(kToAdd);` not part of `Stop` ?

I noticed that Stop will change `kAdding` to `kToRefreshing`. Why doesn't `Stop` change `kAdding` to `kToAdd` directly?

simonlingoogle (approved):
lgtm :+1: 
with a small question.

src/core/net/srp_client.cpp
abtink: Thanks @simonlingoogle .

Yes, In `Stop()`, we do go from `kAdding` to `kToRefresh`.  I added it in first implementation of SRP client with the idea to be safe in the case where client is stopped after we sent an update to server and while waiting for a response from it (there is a a comment in the code which tries to explain it):

```
void Client::Stop(Requester aRequester)
{
    // Change the state of host info and services so that they are
    // added/removed again once the client is started back. In the
    // case of `kAdding`, we intentionally move to `kToRefresh`
    // instead of `kToAdd` since the server may receive our add
    // request and the item may be registered on the server. This
    // ensures that if we are later asked to remove the item, we do
    // notify server.
```

In the new code handling the switch to a new server, we know that we have tried multiple times and waited long enough and received no response from server, so in this case, we manually switch to `kToAdd` (so that we remember we have not yet registered the host info and allow switch to server to happen if things continue to fail.

It's not the best (most clear) approach but it works. I will try to add a comment around `mHostInfo.SetState(kToAdd)` statement to explain this in the code. If you have any suggestions  or ideas around this, please let me know.

simonlingoogle: Thanks for the explanation. That works for me.

src/core/net/srp_client.cpp
```diff
@@ -1160,6 +1181,27 @@ void Client::ProcessResponse(Message &aMessage)
         GrowRetryWaitInterval();
         SetState(kStateToRetry);
         InvokeCallback(error);
+
+#if OPENTHREAD_CONFIG_SRP_CLIENT_AUTO_START_API_ENABLE && OPENTHREAD_CONFIG_SRP_CLIENT_SWITCH_SERVER_ON_FAILURE
+        if ((error == kErrorDuplicated) || (error == kErrorSecurity))
```
wgtdkp, line 1186: `kErrorDuplicated` this usually indicates we should use a new name. Why switching to a new server will resolve the error?

src/core/net/srp_client.cpp
```diff
@@ -1524,6 +1566,22 @@ void Client::HandleTimer(void)
...
+        // After certain number of back-to-back timeout failures, we try
+        // to switch the server. This is again done after invoking the
+        // callback. It works correctly due to the guard check at the
+        // top of `SelectNextServer()`.
+
+        mTimoutFailureCount++;
+
+        if (mTimoutFailureCount >= kMaxTimeoutFailuresToSwitchServer)
+        {
+            mTimoutFailureCount = 0;
```
wgtdkp, line 1581: Seems we forget to reset `mTimoutFailureCount` in `Client::ProcessResponse`. Should `mTimoutFailureCount` be reset in `SelectNextServer` only when we really switched to a new server?

src/core/net/srp_client.cpp
abtink: The `kErrorDuplicated` can be  helpful in situation where another server is still remembering the service (which client registered with it earlier) and not letting it be registered with a new server (particularly when there is no replication between servers). In such a case,  on the client we try to switch the servers till we get back to same server that is keeping/remembering the name.

src/core/net/srp_client.cpp
abtink: > Seems we forget to reset mTimoutFailureCount in Client::ProcessResponse

I think we reset it already. It's around  line ~1158 in new code from this PR in `ProcessResponse()` method.

> Should `mTimoutFailureCount` be reset in `SelectNextServer()` only when we really switched to a new server?

Good suggestion. Thanks. I will change it in next push.

src/core/net/srp_client.cpp
jwhui: What if there is truly a duplicate? [draft-ietf-dnssd-srp-09 Section 2.2.5.2](https://datatracker.ietf.org/doc/html/draft-ietf-dnssd-srp-09#section-2.2.5.2) suggests changing the existing name to resolve the conflict. When would we/the client decide to apply this logic?

abtink: We have a callback from `Client` to reprot the failures to the caller and the `SelectNextServer()` is called after invoking the callback.

In the case of an actual duplicate, I think we can expect the caller to take action and change the name (mainly since the names are really selected by the caller).

jwhui: But how would the client know whether to wait to retry a new server or change the name?

simonlingoogle: Some instance names are unique by nature and can't be changed, e.g. CHIP ID. 
Other instances names can be changed to a different name, e.g. `Printer` -> `Printer (2)`. 
Seems it's better to have different strategy for different kind of instances?

abtink: > But how would the client know whether to wait to retry a new server or change the name?

I think the user of `Srp::Client` can be fully aware of what's happening (from different callback and API calls) and apply  different strategies based on its use-case (as @simonlingoogle  mentioned different names may need different strategies):
- Caller can register a callback to get outcome when registration fails (e.g. server rejects).
- Caller should know whether it has enabled auto-start mode or not (and whether switch server feature is enabled or not)
- It can also track the server changes through the auto-start callback (which notifies caller whenever auto-start selects a new server).

As a simple strategy example, caller can implement a rule of trying to change the name after some `x` back-to-back server rejections (as duplicate).
Or a more complex case, caller can track the server changes and try to change the name when it sees that we get back to the first server again (exhaust all server options and get rejected by all).

I think later on, if we see there is a common useful strategy, we can consider adding it in `Srp::Clinet` directly (i.e. have a separate callback to notify caller that `Srp::Clinet` recommends a name change).

wgtdkp (commented):
LGTM! thanks @abtink :+1:

wgtdkp (approved):

jwhui (approved):

Issue #6714, opened 2021-06-09
[srp] support for service sub-types


abtink:
It can be helpful to add  (native) support for service sub-types in SRP/DNS modules in OpenThread. This issue is to track the efforts to enable this

FYI. @LuDuda, @Damian-Nordic

Damian-Nordic:
FYI @kkasperczyk-no

@abtink mDNS responder allows you to specify subtypes after the main service type, e.g. "_chip._udp,S1,D2" etc. We wondered if OpenThread couldn't supported a similar, compact format.

abtink:
That may be useful for CLI but I think in `otSrpClientService` it would be easier to provide sub-type labels as an (optional) array of strings.

```
typedef struct otSrpClientService
{
    const char *         mName;          ///< The service name labels (e.g., "_chip._udp", not the full domain name).
    const char *         mInstanceName;  ///< The service instance name label (not the full name).
    const char **        mSubTypeNames;  ///< Array of service sub-type labels (must end with `NULL` or can be `NULL`).  
    ...
```

kkasperczyk-no:
@abtink I made some early implementation in order to verify if the assumptions that we made with @Damian-Nordic are correct and it seems that this way we can properly register sub types for both MDNS and Avahi Publishers. That is WIP branch that still needs some work to do (naming refactor, docs, tests alignment) but I hope it can give you a highlight what the general idea is: https://github.com/kkasperczyk-no/ot-br-posix/commit/645c3e582e8e4c170844dbfed47e5acab8e621d9.

I think that presented approach could be beneficial, because it is very compact solution from the SRP client perspective, as we can just add following subtypes separated with comma to the service type. No additional data are necessary to store, no static buffers allocated for it and there is no increasing memory consumption that is already pretty big problem in Matter.

As far as I understand subtype support could be achieved just by modifying OTBR posix code that is responsible for parsing instance name, service type and domain name. Additionally I added vector of subtypes that could be filled with parsed data and passed to the MDNS/Avahi Publisher. I would be grateful for your statement/comments on that proposition.

abtink:
Thanks @KaSroka .

If I understand your approach correctly, the idea is to include the sub-types as comma separated list at end of service name which is then used in all PTR and SRV records in the message. I am not sure if this model is aligned and follows the RFCs on how the names should be formatted and how subtype to be supported.

[...]

PR #6724, opened 2021-06-12
[srp-client] add support for service subtypes

abtink:
This commit adds support for service subtype in `Srp::Client`. The
`otSrpClientService` now includes a pointer to an array of service
subtype labels (which can be set to `NULL` if there is no subtype).
Each subtype is appended as a PTR record (i.e., a Service Discovery
Instruction) in the SRP update message.

This commit also updates The `Utils::SrpClientBuffers` to provide an
array for the subtype labels (pointers). The default size of the
service name buffer is also incremented so that it can be used for
saving subtype label strings.

The CLI `srp client` implementation is also updated to support service
subtypes where the `<servicename>` can optionally include a list of
subtype labels separated by comma (e.g., `_ser._udp,_s1,_s2`).

----

Relates to  https://github.com/openthread/openthread/issues/6714.

### DIFF
--- a/include/openthread/srp_client.h
+++ b/include/openthread/srp_client.h
@@ -92,6 +92,7 @@ typedef struct otSrpClientService
 {
     const char *         mName;          ///< The service name labels (e.g., "_chip._udp", not the full domain name).
     const char *         mInstanceName;  ///< The service instance name label (not the full name).
+    const char *const *  mSubTypeLabels; ///< Array of service sub-type labels (must end with `NULL` or can be `NULL`).
     const otDnsTxtEntry *mTxtEntries;    ///< Array of TXT entries (number of entries is given by `mNumTxtEntries`).
     uint16_t             mPort;          ///< The service port number.
     uint16_t             mPriority;      ///< The service priority.
--- a/src/cli/cli_srp_client.cpp
+++ b/src/cli/cli_srp_client.cpp
@@ -388,6 +388,7 @@ otError SrpClient::ProcessServiceAdd(Arg aArgs[])
     uint16_t                        size;
     char *                          string;
     otError                         error;
+    char *                          label;
 
     entry = otSrpClientBuffersAllocateService(mInterpreter.mInstance);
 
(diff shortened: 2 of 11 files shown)


kkasperczyk-no (approved):

LuDuda (approved):

include/openthread/srp_client.h
```diff
@@ -92,6 +92,7 @@ typedef struct otSrpClientService
 {
     const char *         mName;          ///< The service name labels (e.g., "_chip._udp", not the full domain name).
     const char *         mInstanceName;  ///< The service instance name label (not the full name).
+    const char *const *  mSubTypeLabels; ///< Array of service sub-type labels (must end with `NULL` or can be `NULL`).
```
wgtdkp, line 95: I am thinking if it is more convenient to encode the subtypes in `mName` or `mSubTypeLabels` by separating subtypes with `,`?

This reduces the number of buffers the user needs to pre-allocate and maintain.

abtink: I considered this and I agree it may be easier for allocation on caller if all strings are dynamic. But I went for this model to keep `mName` and `mSubTypeLabeles` as separate strings and allow caller to use `const` strings for them.

I think service names (`mName`) are more often fixed/constant, so allowing them to use `const` string would be helpful, but the subtype labels may or may not be (e.g., I think in `matter`/CHIP most of the subtypes include some dynamic value). Thoughts?

wgtdkp: SGTM!

simonlingoogle (approved):
LGTM :+1: 

with nits

src/core/utils/srp_client_buffers.hpp
```diff
@@ -86,6 +86,12 @@ class SrpClientBuffers : public InstanceLocator, private NonCopyable
          */
         kServiceNameSize = OPENTHREAD_CONFIG_SRP_CLIENT_BUFFERS_SERVICE_NAME_SIZE,
 
+        /**
+         * Array length for service subtype label.
+         *
+         */
+        kServiceSubTypeArrayLength = OPENTHREAD_CONFIG_SRP_CLIENT_BUFFERS_SERVICE_SUB_TYPE_ARRAY_LENGTH,
```
simonlingoogle, line 93: With `OPENTHREAD_CONFIG_SRP_CLIENT_BUFFERS_SERVICE_SUB_TYPE_ARRAY_LENGTH` being 7, we can register at most 6 subtypes. 

Why not change it as below so that users know exactly how many subtypes they can use?
```
#define OPENTHREAD_CONFIG_SRP_CLIENT_BUFFERS_SERVICE_MAX_SUB_TYPES 6
```

src/cli/cli_srp_client.cpp
```diff
@@ -397,6 +398,33 @@ otError SrpClient::ProcessServiceAdd(uint8_t aArgsLength, Arg aArgs[])
...
+
+            label = strchr(label, ',');
+
+            if (label == nullptr)
+            {
+                break;
+            }
+        }
+
+        VerifyOrExit(label == nullptr, error = OT_ERROR_INVALID_ARGS);
```
simonlingoogle, line 425:
```suggestion
        VerifyOrExit(label == nullptr, error = OT_ERROR_NO_BUFS);
```

src/core/utils/srp_client_buffers.hpp
abtink: Thanks. Good suggestion. Changed it in new push.

wgtdkp (approved):
LGTM 👍

src/cli/cli_srp_client.cpp
```diff
@@ -404,6 +405,35 @@ otError SrpClient::ProcessServiceAdd(Arg aArgs[])
...
+        for (uint16_t index = 0; index + 1 < arrayLength; index++)
+        {
+            *label++             = '\0';
+            subTypeLabels[index] = label;
+
+            label = strchr(label, ',');
+
+            if (label == nullptr)
+            {
+                break;
```
jwhui, line 428: Do we need to explicitly set `subTypeLabels[index + 1]` to `nullptr`?

abtink: We can consider adding it but it is not technically necessary since `otSrpClientBuffersAllocateService()` will ensure all buffers and strings are cleared on the newly allocated entry. I will add a comment in `otSrpClientBuffersAllocateService()` to mention this.

jwhui (approved):

PR #6728, opened 2021-06-14
[settings] fix OT_LOG_LEVEL macros

schinchani:
In this PR, I have updated the ActionToString and KeyToString functions to use the correct OT_LOG_LEVEL macro guards as specified in the settings header.


abtink (approved):
LGTM. Thanks. 👍 

Yes. The `#if` checks were incorrect and switched. We need `ActionToSTring()` under `INFO` and `KeyToString` under `WARN`. 

Not in this PR, but later in a new PR, it would be good to add CI build test covering every log level.

jwhui (approved):

PR #6730, opened 2021-06-15
[mle] ensure MLE messages are sent via broadcast

jwhui:
MLE Announce messages are sent with link security enabled. This commit
ensures that REEDs always send MLE Announce messages to the broadcast
destination address.


abtink (approved):
LGTM. Nice find!. 👍

PR #6732, opened 2021-06-15
[common] adding `HeapString` class (heap allocated string) + use in `Srp::Server`

abtink:

**[common] adding `HeapString` class (heap allocated string)**

This commit adds a new class `HeapString` (a heap allocated string).
The buffer to store the string is allocated from heap and is managed
by the `HeapString` class itself, e.g., it may be reused and/or freed
and reallocated when the string is set. The `HeapString` destructor
will always free the allocated buffer.

This commit also adds a unit test `test_heap_string` for the new
class.

**[srp-server] update to use `HeapString` for names**

### DIFF
--- a/src/core/common/heap_string.cpp
+++ b/src/core/common/heap_string.cpp
@@ -1,0 +1,98 @@
+/*
+ *  Copyright (c) 2021, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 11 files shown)


src/core/common/heap_string.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    memcpy(mStringBuffer, aCString, newSize);
+
+exit:
+    return error;
+}
+
+Error HeapString::Set(HeapString &&aString)
+{
+    VerifyOrExit(mStringBuffer != aString.mStringBuffer);
```
wgtdkp, line 71: I think by the API of class `HeapString`, this should never happen. What about using `OT_ASSERT`?

wgtdkp (approved):
Thanks @abtink :+1:

src/core/common/heap_string.cpp
abtink: Thanks @wgtdkp.

I think calling `Set()` with rvalue reference being itself is not technically incorrect use. So I am not sure if we want to restrict its use and `OT_ASSERT()` on it.

It can happen in a statement equivalent to ```name.Set(static_cast<HeapString &&>(name))```.

Though I agree that it is very unlikely to see such a statement (directly as is) used in the code, I think a very similar/equivalent statement can happen indirectly through using a return value from a function/method, compiler inlining the function/method along with RVO (return value optimization), and use of `std::move` (which behaves practically as a `static_cast<HeapString &&>()`). 

Here is a somewhat contrived example to create such a situation but I think similar pattern may happen in the code as well:
```
struct SomeType 
{
   HeapString mName;
   HeapString mOtherName;

   HeapString GetName(void) { return std::move(mName); }
   HeapString GetOtherName(void) { return std::move(mOtherName); }

   void SomeMethod(void)
   {
        bool useOtherName = false;
 
        // Later in code somewhere 
        mName.Set(useOtherName ? GetOtherName() : GetName()));

```

wgtdkp: Thanks @abtink LGTM 👍

jwhui (approved):

PR #6733, opened 2021-06-15
[string] new `StringFind()/StringEndsWith()` for a sub-string

abtink:
This commit adds a new flavor of function `StringFind()` which finds
the first occurrence of a given sub-string within a string. It also
adds `StringEndsWith()` function that works with a sub-string. The
unit test `test_string` is also updated to cover the new behavior of
the newly added function.

----

Background on this PR: The new helper methods will be useful for 
handling of sub-type service name in SRP code.

### DIFF
--- a/src/core/common/string.cpp
+++ b/src/core/common/string.cpp
@@ -66,11 +66,39 @@ const char *StringFind(const char *aString, char aChar)
     return ret;
 }
 
+const char *StringFind(const char *aString, const char *aSubString)
+{
+    const char *ret    = nullptr;
+    size_t      len    = strlen(aString);
+    size_t      subLen = strlen(aSubString);
+
+    VerifyOrExit(subLen <= len);
+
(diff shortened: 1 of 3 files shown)


src/core/common/string.cpp
```diff
@@ -66,13 +66,49 @@ const char *StringFind(const char *aString, char aChar)
...
+        if (*aString == '\0')
+        {
+            start = nullptr;
+            break;
+        }
+
+        if (*aString++ != *sub++)
+        {
+            sub   = aSubString;
+            start = aString;
```
jwhui, line 90: Does this properly handle cases where the substring itself has repetition? For example:
* aString = "abcabcabcd"
* aSubString = "abcabcd"

abtink: You are absolutely right! It does not. Thanks for catching this. 
Will fix it in next push.

jwhui (approved):

PR #6734, opened 2021-06-16
[csl] fix overflow in calculating semiWindow

zwx1995esp:
As for:
```
semiWindow = elapsed * (Get<Radio>().GetCslAccuracy() + mCslParentDrift) / 1000000;
```
Consider that the worst `Get<Radio>().GetCslAccuracy()` and `mCslParentDrift` are both `255`, the max of `uint32` is `4294967295`, so when `elapsed` is  about `8500000`(just 8.5s elapsed since the last CSL sync), the product of `elapsed` and `(Get<Radio>().GetCslAccuracy() + mCslParentDrift)` will be overflowed.

In this PR, use the type `uint64_t` instead of `uint32_t` to avoid the overflow.


edmont (approved):
Good catch, thanks!

jwhui (approved):
Thanks! 👍🏼

PR #6739, opened 2021-06-16
[mesh-forwarder] calculate PAN ID compression according to 802.15.4-2015

suveshpratapa:
Noticed incorrect calculation of PAN ID compression.

Reference: https://mentor.ieee.org/802.15/dcn/15/15-15-0911-01-0mag-proper-pan-id-field-settings-for-802-15-4-2015.docx


jwhui (approved):
Thanks! 👍🏼

PR #6742, opened 2021-06-16
[github-actions] add CI build check for different log levels

abtink:
This commit updates `check-simulation-build-cmake` scripts
- Adds a local `option` variable for Thread 1.2 build configs
- Adds `build_for_all_log_levels()` which builds OT with full
  feature at all different log levels.

### DIFF
--- a/script/check-simulation-build-cmake
+++ b/script/check-simulation-build-cmake
@@ -42,17 +42,27 @@ build_all_features()
     reset_source
     "$(dirname "$0")"/cmake-build simulation -DOT_THREAD_VERSION=1.1
 
+    local options=(
+        "-DOT_OTNS=ON"
+        "-DOT_SIMULATION_VIRTUAL_TIME=ON"
+        "-DOT_THREAD_VERSION=1.2"
+        "-DOT_DUA=ON"
+        "-DOT_MLR=ON"
+        "-DOT_BACKBONE_ROUTER=ON"
+        "-DOT_CSL_RECEIVER=ON"
+    )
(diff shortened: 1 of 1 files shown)


abtink:
One thing to consider is adding it in `check-simulation-build-cmake` will add OT build for every log level (6 options) per each gcc version, clang version, etc in `build` github action workflow. It can add total time it takes for the `build` workflow.

I am wondering whether to keep it as is (build different configs on all toolchain versions/flavors), or limit it to just one (or subset of) toolchain options.

Thoughts?

jwhui:
> One thing to consider is adding it in `check-simulation-build-cmake` will add OT build for every log level (6 options) per each gcc version, clang version, etc in `build` github action workflow. It can add total time it takes for the `build` workflow.
> 
> I am wondering whether to keep it as is (build different configs on all toolchain versions/flavors), or limit it to just one (or subset of) toolchain options.
> 
> Thoughts?
> 
> Checking the run-times, I think building for all 6 log levels roughly adds about ~5-8 min to the total build time (under each toolchain version/flavor).

I am OK to keep across all toolchains for now. We can prune if it turns out to be too long in practice.

simonlingoogle (approved):
LGTM 👍

abtink:
Some of the builds (unrelated to the change) were failing earlier with this PR. Rebased now on `main` to see if they are still failing.

abtink:
Closing this for now to avoid adding more build time to CI. We can consider it again later if needed.

PR #6744, opened 2021-06-17
[tcp] add TCP class and stub implementations of the TCP API

samkumar:
This is a first in a series of pull requests bringing TCPlp to OpenThread. I'm making a series of small pull requests so that it's easier for the OpenThread community to review the code. If you'd like to preview a working version of TCPlp in OpenThread, see #6650.

This pull request implements the boilerplate code for implementing the TCP API in #6491. It creates a TCP class in src/core/net, and implements the functions needed to transfer control there when a TCP API function is called. The TCP class itself will be implemented in future pull requests.

This pull request achieves the following:
* It allows future pull requests can focus on implementing the actual API functions.
* Code written against the TCP API will now link against OpenThread (even though the functions themselves are stub implementations for now).


samkumar:
@abtink and @simionlingoogle Do you mind reviewing this pull request? I would add you on GitHub as reviewers, but it seems that I lack permissions to do that.

samkumar:
@abtink and @simonlingoogle Do you mind reviewing this pull request? I would add you on GitHub as reviewers, but it seems that I lack permissions to do that.

samkumar:
Thanks Jonathan for adding reviewers.

simonlingoogle (approved):
LGTM :+1:

src/core/net/tcp6.hpp
```diff
@@ -0,0 +1,302 @@
...
+     *
+     */
+    class Endpoint : public otTcpEndpoint, public LinkedListEntry<Endpoint>
+    {
+        friend class Tcp;
+        friend class LinkedList<Endpoint>;
+    };
+
+    /**
+     * This class represents a TCP/IPv6 listener.
```
simonlingoogle, line 76:
```suggestion
     * This class represents a TCP/IPv6 listener.
     *
```

src/core/net/tcp6.hpp
```diff
@@ -0,0 +1,302 @@
...
+    /**
+     * This constructor initializes the object.
+     *
+     * @param[in] aInstance  A reference to the OpenThread instance.
+     *
+     */
+    explicit Tcp(Instance &aInstance);
+
+    /**
+     * See documentation for otTcpEndpointInitialize in include/openthread/tcp.h.
```
simonlingoogle, line 182:
```suggestion
     * @sa otTcpEndpointInitialize
```
nit I am not sure what's the best doxygen here.

simonlingoogle (approved):

src/core/net/ip6.hpp
```diff
@@ -373,6 +374,7 @@ class Ip6 : public InstanceLocator, private NonCopyable
     Tasklet       mSendQueueTask;
 
     Icmp mIcmp;
+    Tcp  mTcp;
```
abtink, line 377: I think it may be useful to add the `Tcp` under an OT config (e.g. `OPENTREAD_CONFIG_TCP6_ENABLE`). This way users have control whether to enable and use the OT TCP stack or not.

src/core/net/tcp6.hpp
```diff
@@ -0,0 +1,302 @@
...
+
+    /**
+     * See documentation for otTcpEndpointGetContext in include/openthread/tcp.h.
+     */
+    static void *GetContext(Endpoint &aEndpoint);
+
+    /**
+     * See documentation for otTcpGetLocalAddress in include/openthread/tcp.h.
+     */
+    static const SockAddr &GetLocalAddress(Endpoint &aEndpoint);
```
abtink, line 199: How about we change these to be methods of the `Endpoint` class?

src/core/net/tcp6.hpp
```diff
@@ -0,0 +1,302 @@
...
+
+    /**
+     * See documentation for otTcpListenerGetContext in include/openthread/tcp.h.
+     */
+    static void *GetContext(Listener &aListener);
+
+    /**
+     * See documentation for otTcpListen in include/openthread/tcp.h.
+     */
+    Error Listen(Listener &aListener, const SockAddr &aSockName);
```
abtink, line 274: How about defining these as methods of `Listener` class?

src/core/net/tcp6.hpp
```diff
@@ -0,0 +1,302 @@
...
+    explicit Tcp(Instance &aInstance);
+
+    /**
+     * See documentation for otTcpEndpointInitialize in include/openthread/tcp.h.
+     */
+    Error Initialize(Endpoint &aEndpoint, otTcpEndpointInitializeArgs &aArgs);
+
+    /**
+     * See documentation for otTcpEndpointGetInstance in include/openthread/tcp.h.
+     */
```
abtink, line 188: I suggest we copy the documentation of methods/functions again here, instead of just pointing to `tcp.h`. 

Having it in the OT internal types again can be useful (someone reading the code and starts from internal definition. Also some IDEs (editors) can parse and show documentation when writing code which can be helpful).

In other modules, we usually repeat the documentation (we have in public OT header and then again in internal types/definitions).

abtink (commented):
Thanks @samkumar. Overall looks good.
I have couple of suggestions below.

src/core/net/ip6.hpp
simonlingoogle: I woud suggest `OPENTREAD_CONFIG_TCP_ENABLE` without `6`.

src/core/net/tcp6.hpp
samkumar: Thanks.

samkumar: I'll apply this convention uniformly throughout in the next update.

src/core/net/tcp6.hpp
samkumar: @abtink's suggestion is to replace this entirely with a copy of the documentation in `include/openthread/tcp.h`, so I'll probably do that instead.

src/core/net/tcp6.hpp
samkumar: OK, I am happy to do this. My original reasoning was to avoid having duplicate comments (since then we would have to remember to change the other copy if we decide to change things). But if there's precedent for copying the comments elsewhere in OpenThread, then I agree that we should do that.

src/core/net/tcp6.hpp
samkumar: I agree that moving these member functions into the `Endpoint` class is natural, and I'm happy to make this change.

The reason I had originally put them into the `Tcp` class with `Endpoint &` as the first argument is that `Udp` did it this way. If you look at `src/core/net/udp6.cpp`, the functions are implemented in the `Udp` class, with methods in the `Listener` and `SocketHandle` classes calling the functions in the `Udp` class.

samkumar: FYI, this would change the flow a little bit, as some of these functions will need to access members of the `Tcp` class, so looking up the instance would move from `tcp_api.cpp` here.

src/core/net/tcp6.hpp
samkumar: I'm happy to do this.

I was originally following the precedent set by the `Udp` class (see my response to your other comment about defining some methods in the `Endpoint` class).

src/core/net/ip6.hpp
samkumar: I'm happy to add the flag. If the flag is off (set to `0`), then what should the API functions do?

samkumar: Is there a preferred error to return? And should that only be implemented in the external API, or in the internal API as well (i.e., if something in OpenThread uses `Tcp::Endpoint` and `Tcp::Listener` directly)?

jwhui: The convention so far has been to not implement the related functions when a given build feature is disabled, which would cause link failures with anything that tries to use them.

Whether we need to separate public from internal APIs depends on if there is a use case for it. For many features, we do not have such separation. However, for coaps we do - Thread Commissioning relies on coaps and applications may also want to leverage the same coaps implementation.

samkumar: Thanks Jonathan for clarifying. Since it's OK for applications that try to use TCP when it's disabled to fail to link, it sounds like we should also wrap most of `tcp6.cpp` in a `#ifdef` so that those functions aren't defined at all. Does that sound reasonable?

jwhui: Yes, that's what we do for other components. A couple examples:
https://github.com/openthread/openthread/blob/3a7e836dd7c96a932cbd8ab7733b1042bfd86588/src/core/api/coap_api.cpp#L36
https://github.com/openthread/openthread/blob/3a7e836dd7c96a932cbd8ab7733b1042bfd86588/src/core/api/jam_detection_api.cpp#L36

src/core/net/tcp6.hpp
samkumar: I'll also include the @sa line in the doxygen though.

src/core/net/tcp6.hpp
abtink: I think having them defined in `Endpoint`/`Listener` class would be simpler 

Since we already assume/require that `Endpoint` and `Listener` track the `ot::Instance` they belong to, they can then get to any other object in OT from `Instance` itself (using the `Instance::Get<>()` specialization for different types, e.g. `Get<Tcp>()` to get the `Tcp` class). 

I won't recommend the `Udp` socket model as a good example to follow. The model we have used there is more tied to some issues in the first API design and implementation and us wanting to keep the public APIs unchanged later on. You can see more on this from Issue https://github.com/openthread/openthread/issues/5162 and this PR https://github.com/openthread/openthread/pull/5231.

One thing I suggest to avoid the same issue that came up in the `Udp::Socket` case is that we don't declare `Endpoint` and `Listener` as `InstanceLocator` and instead have `otInstance *mInstance` as a member variable directly included in the `/* Other implementation-defined fields go here. */` section of the original `struct otTcpEndpoint` or `struct otTcpLisner`.

samkumar:
I just pushed an update that incorporates the latest comments. @jwhui, @abtink, and @simonlingoogle, feel free to take another look when you get a chance.

src/core/config/ip6.h
```diff
@@ -177,4 +177,14 @@
...
 #endif
 
+/**
+ * @def OPENTHREAD_CONFIG_TCP_ENABLE
+ *
+ * Define as 1 to enable TCP.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_TCP_ENABLE
+#define OPENTHREAD_CONFIG_TCP_ENABLE 1
```
abtink, line 187: I think this is good here now, in future I expect we probably will have more configs for TCP module, we can then add them all in a new header `config/tcp.h`.

abtink (approved):
Thanks @samkumar. Looks great. 👍 

Just added some smaller style suggestions.

src/core/net/tcp6.hpp
```diff
@@ -0,0 +1,518 @@
...
+         *
+         * @sa otTcpReceiveByReference
+         *
+         * @param[out]  aBuffer    A pointer to the linked buffer chain referencing data currently in the receive
+         * buffer.
+         *
+         * @retval kErrorNone    Successfully completed the operation.
+         * @retval kErrorFailed  Failed to complete the operation.
+         */
+        Error ReceiveByReference(const otLinkedBuffer *&aBuffer);
```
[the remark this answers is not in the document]
samkumar: Noted.

src/core/net/tcp6.hpp
```diff
@@ -0,0 +1,518 @@
...
+
+        /**
+         * Obtains the Instance that was associated with this Endpoint upon
+         * initialization.
+         *
+         * @sa otTcpEndpointGetInstance
+         *
+         * @returns  The Instance pointer associated with this Endpoint.
+         */
+        Instance *GetInstance();
```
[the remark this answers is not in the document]
samkumar: OK, I will go through the files (both `tcp6.hpp` and `tcp6.cpp` and change all empty argument lists from `()` to `(void)`.

src/core/net/tcp6.hpp
```diff
@@ -0,0 +1,518 @@
...
+
+        /**
+         * Obtains the context pointer that was associated this Endpoint upon
+         * initialization.
+         *
+         * @sa otTcpEndpointGetContext
+         *
+         * @returns  The context pointer associated with this Endpoint.
+         */
+        void *GetContext();
```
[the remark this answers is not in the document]
samkumar: OK, I will go through the files (both `tcp6.hpp` and `tcp6.cpp` and change all empty argument lists from `()` to `(void)`.

src/core/net/tcp6.hpp
```diff
@@ -0,0 +1,518 @@
...
+         * Obtains a pointer to a TCP endpoint's local host and port.
+         *
+         * The contents of the host and port may be stale if this socket is not in a
+         * connected state and has not been bound after it was last disconnected.
+         *
+         * @sa otTcpGetLocalAddress
+         *
+         * @returns  The local host and port of this Endpoint.
+         */
+        const SockAddr &GetLocalAddress();
```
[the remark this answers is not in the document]
samkumar: OK. Should we also go back to the API in `include/openthread/tcp.h` and change those functions to declare the appropriate arguments as `const`? For example, `const otSockAddr *otTcpGetLocalAddress(otTcpEndpoint *aEndpoint);` could become `const otSockAddr *otTcpGetLocalAddress(const otTcpEndpoint *aEndpoint);`.

src/core/net/tcp6.hpp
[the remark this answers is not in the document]
samkumar: Yes, I believe that this can be `const` as well.

src/core/config/ip6.h
samkumar: Noted.

samkumar:
@abtink Thanks for your approval and for your feedback. I just pushed an update addressing your feedback, including updating the TCP API to declare certain parameters `const`. I also squashed and rebased the commits so that it can be merged.

src/core/net/tcp6.hpp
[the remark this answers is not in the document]
jwhui: We generally try to apply `const` wherever we can and is appropriate.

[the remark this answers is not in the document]
samkumar: Yes, I included the appropriate changes to the API in the latest update.

abtink (approved):
Looks great. Thanks. 👍

jwhui (approved):
Thanks! 👍🏼

PR #6746, opened 2021-06-17
[diag] don't let the client of diag reserve the '\0' byte

Irving-cl:
This PR initializes the `output` buffer to all zero.

The output buffer isn't initialized to all zero currently. This makes it possible that `OutputFormat` prints some random bytes at the end of the original output generated by processing the command.

I checked the output buffer by dumping the buffer:
**Before processing command:**
diag_1
**After processing command:**
diag-2

So it is possible that non `\0` characters come after the output and cause incorret output at the end. (In the above case, it happens to be `\0` after the output.)

### DIFF
--- a/include/openthread/diag.h
+++ b/include/openthread/diag.h
@@ -54,6 +54,9 @@ extern "C" {
 /**
  * This function processes a factory diagnostics command line.
  *
+ * The output of this function (the content written to @p aOutput) MUST terminate with `\0` and the `\0` is within the
+ * output buffer.
+ *
  * @param[in]   aInstance       A pointer to an OpenThread instance.
  * @param[in]   aArgsLength     The number of elements in @p aArgs.
  * @param[in]   aArgs           An array of arguments.
@@ -73,6 +76,9 @@ otError otDiagProcessCmd(otInstance *aInstance,
 
 /**
  * This function processes a factory diagnostics command line.
+ *
+ * The output of this function (the content written to @p aOutput) MUST terminate with `\0` and the `\0` is within the
+ * output buffer.
  *
  * @param[in]   aInstance       A pointer to an OpenThread instance.
  * @param[in]   aString         A NULL-terminated input string.
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -4521,12 +4521,9 @@ otError Interpreter::ProcessDiag(Arg aArgs[])
     char    output[OPENTHREAD_CONFIG_DIAG_OUTPUT_BUFFER_SIZE];
 
     // all diagnostics related features are processed within diagnostics module
-    output[0]                  = '\0';
-    output[sizeof(output) - 1] = '\0';
-
     Arg::CopyArgsToStringArray(aArgs, args);
 
-    error = otDiagProcessCmd(mInstance, Arg::GetArgsLength(aArgs), args, output, sizeof(output) - 1);
+    error = otDiagProcessCmd(mInstance, Arg::GetArgsLength(aArgs), args, output, sizeof(output));
 
     OutputFormat("%s", output);
 
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (137)
+#define OPENTHREAD_API_VERSION (138)
 
 /**
  * @addtogroup api-instance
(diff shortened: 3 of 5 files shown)


src/cli/cli.cpp
```diff
@@ -4612,12 +4612,9 @@ otError Interpreter::ProcessDiag(uint8_t aArgsLength, Arg aArgs[])
 {
     otError error;
     char *  args[kMaxArgs];
-    char    output[OPENTHREAD_CONFIG_DIAG_OUTPUT_BUFFER_SIZE];
+    char    output[OPENTHREAD_CONFIG_DIAG_OUTPUT_BUFFER_SIZE] = {'\0'};
```
abtink, line 4615: I think you intend to have all chars set to zero, is this right?
If so, how about `memset()` the `output` to zero instead.

Irving-cl: SGTM, Thanks.

zhanglongxia (approved):
LGTM

abtink (approved):
LGTM. Thanks. 👍

src/cli/cli.cpp
```diff
@@ -4615,8 +4615,7 @@ otError Interpreter::ProcessDiag(uint8_t aArgsLength, Arg aArgs[])
     char    output[OPENTHREAD_CONFIG_DIAG_OUTPUT_BUFFER_SIZE];
 
     // all diagnostics related features are processed within diagnostics module
-    output[0]                  = '\0';
-    output[sizeof(output) - 1] = '\0';
+    memset(output, 0, sizeof(output));
```
jwhui, line 4618: Shouldn't the client ensure that what they write to the output buffer is null-terminated? In what case do we expect the client to not null terminte?

That said, I am not opposed to initializing the output buffer to all zeros.

simonlingoogle: Besides, we actually don't define how `output` should be initialized in API `otDiagProcessCmd`. Technically speaking, we should not initialize `output` at all. 
It's `Diags::ProcessCmd`'s responsiblity to make sure the output is null-terminated?

Irving-cl: It looks like currently the implementation of `otDiagProcessCmd` doesn't ensure the output will be null-terminated. For example, in most cases `AppendErrorResult` is called at the end and it doesn't append a `\0` to the output. 

I think the problem is: do we want to let `otDiagProcessCmd` to ensure the output is null-terminated (then we also need to require `otPlatDiagProcess` to make sure that)? And we should add document and update the implementation for those APIs. Or we require the caller of the API to pass an all-zero buffer to `otDiagProcessCmd`. Both approaches make sense to me. Thoughts? @jwhui @simonlingoogle

simonlingoogle: I think `otDiagProcessCmd` should make sure the output is null-terminated as long as it returns `kErrorNone`.

abtink: > I think `otDiagProcessCmd` should make sure the output is null-terminated as long as it returns kErrorNone.

I fully agree. +1 on this.

LuDuda (approved):

Irving-cl:
Update:

Sorry I made a mistake here. Our diag commands are correctly implemented so that the output of them always terminate with `\0`. This is actually ensured by `snprintf` and `vsnprintf`. So the client don't need to worry about this.

I made this clear by adding comments in the Diag APIs. Also I removed the code in the client which sets `\0` in the end of the output buffer and reserves one byte.

Thoughts? @jwhui @simonlingoogle @abtink

src/cli/cli.cpp
```diff
@@ -4615,12 +4615,11 @@ otError Interpreter::ProcessDiag(uint8_t aArgsLength, Arg aArgs[])
     char    output[OPENTHREAD_CONFIG_DIAG_OUTPUT_BUFFER_SIZE];
 
     // all diagnostics related features are processed within diagnostics module
-    output[0]                  = '\0';
-    output[sizeof(output) - 1] = '\0';
+    output[0] = '\0';
```
simonlingoogle, line 4618: This seems unnecessary according to `otDiagProcessCmd`'s implementation?

include/openthread/diag.h
```diff
@@ -54,6 +54,9 @@ extern "C" {
 /**
  * This function processes a factory diagnostics command line.
  *
+ * The output of this function (the content written to @p aOutput) MUST terminate with `\0` and the `\0` is within the
```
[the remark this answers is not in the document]
Irving-cl: Yes. Note the implementation of `Diag::ProcessLine` and `Diag::ProcessCmd`, when the error is not `kErrorNone`, it will write an error message into the output buffer.

src/cli/cli.cpp
Irving-cl: Make sense to me.

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6747, opened 2021-06-17
[srp-server] take ownership of resources on updating existing `Service`

abtink:
This commit changes the `Srp::Server` to take ownership of resources
(e.g., heap allocated buffer for TXT data) when updating an existing
`Service` entry (instead of copying the data). With this change, the
method `Service::SetTxtData()` is no longer needed and is removed.

---

_Background on this PR_:
- Taking ownership (move semantic) would be helpful for adding service subtype
  support (https://github.com/openthread/openthread/issues/6714) - We will have a list of subtype `HeapString` labels in the `Service` 
and the move model would allow us to move/merge them (instead of copy).
- I added it as a separate PR (small change) to make it easier to review. Thanks.

### DIFF
--- a/src/core/net/srp_server.cpp
+++ b/src/core/net/srp_server.cpp
@@ -366,7 +366,7 @@ void Server::CommitSrpUpdate(Error                    aError,
     }
     else if (existingHost != nullptr)
     {
-        const Service *service = nullptr;
+        Service *service = nullptr;
 
         // Merge current updates into existing host.
 
@@ -386,9 +386,9 @@ void Server::CommitSrpUpdate(Error                    aError,
                 Service *newService = existingHost->AddService(service->GetFullName());
 
                 VerifyOrExit(newService != nullptr, aError = kErrorNoBufs);
-                SuccessOrExit(aError = newService->CopyResourcesFrom(*service));
+                newService->TakeResourcesFrom(*service);
                 otLogInfoSrp("[server] %s service %s", (existingService != nullptr) ? "update existing" : "add new",
-                             service->GetFullName());
+                             newService->GetFullName());
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/core/net/srp_server.cpp
```diff
@@ -386,7 +386,7 @@ void Server::CommitSrpUpdate(Error                    aError,
                 Service *newService = existingHost->AddService(service->GetFullName());
 
                 VerifyOrExit(newService != nullptr, aError = kErrorNoBufs);
-                SuccessOrExit(aError = newService->CopyResourcesFrom(*service));
+                newService->TakeResourcesFrom(*service);
```
wgtdkp, line 389: @abtink  Will it be easier for us to support sub-types if we simply adds new services to the existing host? For example:
```c++
// pseudo code for moving service from aHost to existingHost
existingHost->Remove(existingService);
aHost.Remove(service);
existingHost->Add(service);
service->mTimeLastUpdate = TimerMilli::GetNow();
```
(We will need to refactor the while-loop as we are deleting elements during iteration

abtink: @wgtdkp I am not sure if we can use the `Remove` and `Add` model you are suggesting (basically replacing the existing service with the new `Service` being committed).

I think it is possible and valid for subtypes to be added or removed in seperate (follow up) SRP update messages for an existing service. In such a case, the `service` we parse from the new `message` and want to commit may not have the full list of previously registered sub-types (just the new ones being added/removed), so we cannot replace the existing one with new one being committed and need to combine/merge them.

------

_Some more info on the data model I am considering for the sub-types_:

The `Service` is mainly a representation of a `SRV` and `TXT` records we get for a "service instance name". We can have multiple `PTR` records all pointing to the same "service instance name" (basically different sub-types names all associated with same `SRV`/`TXT` info or the same `Service` object). 

So the data model I am considering for subtypes is to have a list of sub-type label `HeapString`s in the `Service` object (I am leaning towards using a linked-list for this though also considered a fixed-size array which may be easier to implement but adds a hard limit on number of sub-types). We also need to track the "update time" for each sub-type seperately since they be registered in different messages and at different times.  If you have any suggestions/thoughts on this, please let me know. Thanks. 😃

wgtdkp: okay, it makes sense if we want to support more flexible sub-types updating :+1:

src/core/net/srp_server.cpp
```diff
@@ -386,7 +386,7 @@ void Server::CommitSrpUpdate(Error                    aError,
...
                 otLogInfoSrp("[server] %s service %s", (existingService != nullptr) ? "update existing" : "add new",
                              service->GetFullName());
```
wgtdkp, line 391: it may be better for us to stop accessing the service whose resources has been removed, though we don't actually remove its name.
```suggestion
                             newService->GetFullName());
```

wgtdkp (approved):
Thanks @abtink :+1:

src/core/net/srp_server.cpp
abtink: Good point! Will change it. 

We don't currently move  the `service->mFullname` (and I think copy it in `AddService()` which will call `Service::New(const char *aFullName)`) but we can consider changing it later to do this (would be a bit more efficient). I won't change it from this PR though (to keep the change small).

jwhui (approved):

Issue #6447, opened 2021-04-14
SRP Client on the accessory used incorrect srp server IP address  


Abhayakara:
The SRP client uses an IP address that isn't configured on the SRP server.

**To Reproduce**

Git commit id: d8a2cd9a70cfda7c2657645b4a57411eadaa4e61
Border router: Apple's implementation of Duckhorn router running on Raspi pi 4 Model B
Accessory: running on nordic nrf52840-DK

**Expected behavior** We expect the SRP client to use the address advertised in the network data

**Console/log output**
```
I: 48707 [DIS]Found admin paring for admin 0, node BC5C01
D: 48712 [DIS]Using Thread MAC for hostname.
I: 48717 [SVR]Device completed Rendezvous process
E: 48911 [DL]Long dispatch time: 264 ms
D: 48915 [DL]In DriveBLEState
D: 48918 [DL]Indication for CHIPoBLE TX characteristic done (ConnId 0x00, result 0x00)
D: 48926 [DL]OpenThread State Changed (Flags: 0x1104d11d)
D: 48931 [DL]   Device Role: DETACHED
D: 48935 [DL]   Network Name: OpenThread
D: 48938 [DL]   PAN Id: 0xABCD
D: 48941 [DL]   Extended PAN Id: 0xDEAD00BEEF00CAFE
D: 48946 [DL]   Channel: 11
D: 48948 [DL]   Mesh Prefix: fdde:ad00:beef::/64
D: 48953 [DL]   Thread Unicast Addresses:
D: 48956 [DL]        fdde:ad00:beef:0:c41c:d4a5:3f62:b55e/64 valid preferred
D: 48963 [DL]        fe80::14fd:46fd:1be9:34db/64 valid preferred
D: 49486 [DL]OpenThread State Changed (Flags: 0x00000100)
I: 50507 [DL]SRP Client was started, as detected server addressed: fdf2:7a68:5f3e:108c:0:ff:fe00:fc11
D: 50517 [DL]OpenThread State Changed (Flags: 0x301332b7)
D: 50522 [DL]   Device Role: CHILD
D: 50525 [DL]   Network Name: MyHomeBR
D: 50528 [DL]   PAN Id: 0xABCD
D: 50531 [DL]   Extended PAN Id: 0x949CD1F473FF7008
D: 50536 [DL]   Channel: 11
D: 50538 [DL]   Mesh Prefix: fdf2:7a68:5f3e:108c::/64
[...]

PR #6752, opened 2021-06-21
[backbone-router] relax matching BBR dataset

simonlingoogle:
This commit relaxes BBR dataset matching to only compare the first byte of Service Data (i.e. `0x01`). 

This commit also requires all Service TLV searching methods to explicitly specify `aExactServiceDataMatch` argument, helping to reduce unconscious mistakes.

### DIFF
--- a/src/core/thread/network_data.cpp
+++ b/src/core/thread/network_data.cpp
@@ -565,39 +565,27 @@ const PrefixTlv *NetworkData::FindPrefix(const uint8_t *aPrefix,
     return prefixTlv;
 }
 
-const ServiceTlv *NetworkData::FindService(uint32_t       aEnterpriseNumber,
-                                           const uint8_t *aServiceData,
-                                           uint8_t        aServiceDataLength) const
+const ServiceTlv *NetworkData::FindService(uint32_t         aEnterpriseNumber,
+                                           const uint8_t *  aServiceData,
+                                           uint8_t          aServiceDataLength,
+                                           ServiceMatchMode aServiceMatchMode) const
 {
-    return FindService(aEnterpriseNumber, aServiceData, aServiceDataLength, mTlvs, mLength);
+    return FindService(aEnterpriseNumber, aServiceData, aServiceDataLength, aServiceMatchMode, mTlvs, mLength);
 }
 
-const ServiceTlv *NetworkData::FindService(uint32_t       aEnterpriseNumber,
-                                           const uint8_t *aServiceData,
-                                           uint8_t        aServiceDataLength,
-                                           const uint8_t *aTlvs,
-                                           uint8_t        aTlvsLength)
-{
-    return FindService(aEnterpriseNumber, aServiceData, aServiceDataLength, /* aExactServiceDataMatch */ true, aTlvs,
-                       aTlvsLength);
-}
-
-const ServiceTlv *NetworkData::FindService(uint32_t       aEnterpriseNumber,
-                                           const uint8_t *aServiceData,
-                                           uint8_t        aServiceDataLength,
-                                           bool           aExactServiceDataMatch,
-                                           const uint8_t *aTlvs,
-                                           uint8_t        aTlvsLength)
+const ServiceTlv *NetworkData::FindService(uint32_t         aEnterpriseNumber,
+                                           const uint8_t *  aServiceData,
+                                           uint8_t          aServiceDataLength,
+                                           ServiceMatchMode aServiceMatchMode,
+                                           const uint8_t *  aTlvs,
+                                           uint8_t          aTlvsLength)
 {
     TlvIterator       tlvIterator(aTlvs, aTlvsLength);
     const ServiceTlv *serviceTlv;
 
     while ((serviceTlv = tlvIterator.Iterate<ServiceTlv>()) != nullptr)
     {
-        if ((serviceTlv->GetEnterpriseNumber() == aEnterpriseNumber) &&
-            (serviceTlv->GetServiceDataLength() >= aServiceDataLength) &&
-            (!aExactServiceDataMatch || (serviceTlv->GetServiceDataLength() == aServiceDataLength)) &&
-            (memcmp(serviceTlv->GetServiceData(), aServiceData, aServiceDataLength) == 0))
+        if (MatchService(*serviceTlv, aEnterpriseNumber, aServiceData, aServiceDataLength, aServiceMatchMode))
         {
             break;
         }
@@ -606,10 +594,11 @@ const ServiceTlv *NetworkData::FindService(uint32_t       aEnterpriseNumber,
     return serviceTlv;
 }
 
-const ServiceTlv *NetworkData::FindNextMatchingService(const ServiceTlv *aPrevServiceTlv,
-                                                       uint32_t          aEnterpriseNumber,
-                                                       const uint8_t *   aServiceData,
-                                                       uint8_t           aServiceDataLength) const
+const ServiceTlv *NetworkData::FindNextService(const ServiceTlv *aPrevServiceTlv,
+                                               uint32_t          aEnterpriseNumber,
+                                               const uint8_t *   aServiceData,
+                                               uint8_t           aServiceDataLength,
+                                               ServiceMatchMode  aServiceMatchMode) const
 {
     const uint8_t *tlvs;
     uint8_t        length;
@@ -625,8 +614,35 @@ const ServiceTlv *NetworkData::FindNextMatchingService(const ServiceTlv *aPrevSe
         length = static_cast<uint8_t>((mTlvs + mLength) - tlvs);
     }
 
-    return FindService(aEnterpriseNumber, aServiceData, aServiceDataLength, /* aExactServiceDataMatch */ false, tlvs,
-                       length);
+    return FindService(aEnterpriseNumber, aServiceData, aServiceDataLength, aServiceMatchMode, tlvs, length);
+}
+
+bool NetworkData::MatchService(const ServiceTlv &aServiceTlv,
+                               uint32_t          aEnterpriseNumber,
+                               const uint8_t *   aServiceData,
+                               uint8_t           aServiceDataLength,
+                               ServiceMatchMode  aServiceMatchMode)
+{
+    bool match = false;
+
+    VerifyOrExit(aServiceTlv.GetEnterpriseNumber() == aEnterpriseNumber &&
+                 aServiceTlv.GetServiceDataLength() >= aServiceDataLength);
+
+    switch (aServiceMatchMode)
+    {
+    case kServiceExactMatch:
+        VerifyOrExit(aServiceTlv.GetServiceDataLength() == aServiceDataLength);
+        OT_FALL_THROUGH;
+
+    case kServicePrefixMatch:
+        VerifyOrExit(memcmp(aServiceTlv.GetServiceData(), aServiceData, aServiceDataLength) == 0);
+        break;
+    }
+
+    match = true;
+
+exit:
+    return match;
 }
 
 NetworkDataTlv *NetworkData::AppendTlv(uint16_t aTlvSize)
--- a/src/core/thread/network_data.hpp
+++ b/src/core/thread/network_data.hpp
@@ -457,7 +457,7 @@ protected:
      * @param[in]  aEnterpriseNumber  Enterprise Number.
      * @param[in]  aServiceData       A pointer to a Service Data.
      * @param[in]  aServiceDataLength The Service Data length pointed to by @p aServiceData.
-     * @param[in]  aServiceMatchMode  The Service Data match mode. @sa ServiceDataMatchMode
+     * @param[in]  aServiceMatchMode  The Service Data match mode.
      *
      * @returns A pointer to the Service TLV if one is found or nullptr if no matching Service TLV exists.
      *
@@ -473,7 +473,7 @@ protected:
      * @param[in]  aEnterpriseNumber  Enterprise Number.
      * @param[in]  aServiceData       A pointer to a Service Data.
      * @param[in]  aServiceDataLength The Service Data length pointed to by @p aServiceData.
-     * @param[in]  aServiceMatchMode  The Service Data match mode. @sa ServiceDataMatchMode
+     * @param[in]  aServiceMatchMode  The Service Data match mode.
      * @param[in]  aTlvs              A pointer to a specified tlvs buffer.
      * @param[in]  aTlvsLength        The specified tlvs buffer length pointed to by @p aTlvs.
      *
@@ -498,7 +498,7 @@ protected:
      * @param[in]  aEnterpriseNumber  Enterprise Number.
      * @param[in]  aServiceData       A pointer to a Service Data.
      * @param[in]  aServiceDataLength The Service Data length pointed to by @p aServiceData.
-     * @param[in]  aServiceMatchMode  The Service Data match mode. @sa ServiceDataMatchMode
+     * @param[in]  aServiceMatchMode  The Service Data match mode.
      * @param[in]  aTlvs              A pointer to a specified tlvs buffer.
      * @param[in]  aTlvsLength        The specified tlvs buffer length pointed to by @p aTlvs.
      *
@@ -523,7 +523,7 @@ protected:
      * @param[in]  aEnterpriseNumber  Enterprise Number.
      * @param[in]  aServiceData       A pointer to a Service Data to match with Service TLVs.
      * @param[in]  aServiceDataLength The Service Data length pointed to by @p aServiceData.
-     * @param[in]  aServiceMatchMode  The Service Data match mode. @sa ServiceDataMatchMode
+     * @param[in]  aServiceMatchMode  The Service Data match mode.
      *
      * @returns A pointer to the next matching Service TLV if one is found or nullptr if it cannot be found.
      *
--- a/src/core/thread/network_data.hpp
+++ b/src/core/thread/network_data.hpp
@@ -309,6 +309,16 @@ public:
 
 protected:
     /**
+     * This enumeration defines Service Data match mode.
+     *
+     */
+    enum ServiceMatchMode : uint8_t
+    {
+        kServicePrefixMatch, ///< Match the Service Data by prefix.
+        kServiceExactMatch,  ///< Match the full Service Data exactly.
+    };
+
+    /**
      * This method returns a pointer to the start of Network Data TLV sequence.
      *
      * @returns A pointer to the start of Network Data TLV sequence.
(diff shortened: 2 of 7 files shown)


src/core/thread/network_data.hpp
```diff
@@ -424,70 +424,81 @@ class NetworkData : public InstanceLocator
     /**
      * This method returns a pointer to a matching Service TLV.
      *
-     * @param[in]  aEnterpriseNumber  Enterprise Number.
-     * @param[in]  aServiceData       A pointer to a Service Data.
-     * @param[in]  aServiceDataLength The Service Data length pointed to by @p aServiceData.
+     * @param[in]  aEnterpriseNumber        Enterprise Number.
+     * @param[in]  aServiceData             A pointer to a Service Data.
+     * @param[in]  aServiceDataLength       The Service Data length pointed to by @p aServiceData.
+     * @param[in]  aExactServiceDataMatch   Whether to use exact match.
```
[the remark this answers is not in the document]
abtink: +1 on @librasungirl suggestion.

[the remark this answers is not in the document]
simonlingoogle: Sounds good to me. 

Can we define two enum values:
```c++
enum ServiceDataMatchMode: uint8_t {
  kExactMatch, 
  kPrefixMatch,
}
```

@abtink @librasungirl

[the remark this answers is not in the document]
jwhui: Works for me.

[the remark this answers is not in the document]
simonlingoogle: Updated. Please review again.

src/core/thread/network_data.cpp
```diff
@@ -565,39 +565,27 @@ const PrefixTlv *NetworkData::FindPrefix(const uint8_t *aPrefix,
...
-const ServiceTlv *NetworkData::FindService(uint32_t       aEnterpriseNumber,
-                                           const uint8_t *aServiceData,
-                                           uint8_t        aServiceDataLength,
-                                           bool           aExactServiceDataMatch,
-                                           const uint8_t *aTlvs,
-                                           uint8_t        aTlvsLength)
+const ServiceTlv *NetworkData::FindService(uint32_t             aEnterpriseNumber,
+                                           const uint8_t *      aServiceData,
+                                           uint8_t              aServiceDataLength,
+                                           ServiceDataMatchMode aServiceMatchMode,
```
librasungirl, line 579:
```suggestion
                                           ServiceMatchMode aServiceMatchMode,
```

Or
`ServiceDataMatchMode aServiceDataMatchMode,`

src/core/thread/network_data.hpp
```diff
@@ -308,6 +308,16 @@ class NetworkData : public InstanceLocator
     Error GetNextServer(Iterator &aIterator, uint16_t &aRloc16) const;
 
 protected:
+    /**
+     * Defines Service Data match mode.
```
librasungirl, line 312:
```suggestion
     * Define Service Data match mode.
```

src/core/thread/network_data.hpp
```diff
@@ -427,14 +437,18 @@ class NetworkData : public InstanceLocator
      * @param[in]  aEnterpriseNumber  Enterprise Number.
      * @param[in]  aServiceData       A pointer to a Service Data.
      * @param[in]  aServiceDataLength The Service Data length pointed to by @p aServiceData.
+     * @param[in]  aServiceMatchMode  The Service Data match mode.
```
librasungirl, line 440:
```suggestion
     * @param[in]  aServiceMatchMode  The Service Data match mode (@sa ServiceDataMatchMode)
```

src/core/thread/network_data_service.cpp
```diff
@@ -100,29 +100,36 @@ void Manager::GetBackboneRouterPrimary(ot::BackboneRouter::BackboneRouterConfig
...
-    Iterator                          iterator;
 
     aConfig.mServer16 = Mac::kShortAddrInvalid;
 
-    iterator.mServiceTlv = Get<Leader>().FindService(kThreadEnterpriseNumber, &BackboneRouter::kServiceData,
-                                                     sizeof(BackboneRouter::kServiceData));
-
-    while (IterateToNextServer(iterator) == kErrorNone)
+    for (const ServiceTlv *serviceTlv = Get<Leader>().FindService(
+             kThreadEnterpriseNumber, &BackboneRouter::kServiceData, 1, NetworkData::kServiceDataPrefixMatch);
```
librasungirl, line 107: define a constant, like `THREAD_SERVICE_DATA_BBR_SIZE` as 1, in the header and avoid the magic value 1 here?

src/core/thread/network_data_service.cpp
```diff
@@ -165,7 +172,8 @@ Error Manager::GetNextDnsSrpAnycastInfo(Iterator &aIterator, DnsSrpAnycast::Info
 
     do
     {
-        tlv = Get<Leader>().FindNextMatchingService(tlv, kThreadEnterpriseNumber, &serviceData, sizeof(serviceData));
+        tlv = Get<Leader>().FindNextMatchingService(tlv, kThreadEnterpriseNumber, &serviceData, sizeof(serviceData),
```
librasungirl, line 175: Should we also relax the matching here as well?
Currently service data for `DNS-SD Anycast Dataset` is of length 2, but may also subject to have additional data in the future likewise.

src/core/thread/network_data_service.cpp
```diff
@@ -237,9 +245,9 @@ Error Manager::GetNextDnsSrpUnicastInfo(Iterator &aIterator, DnsSrpUnicast::Info
 
         // Find the next matching Service TLV.
 
-        aIterator.mServiceTlv =
-            Get<Leader>().FindNextMatchingService(aIterator.mServiceTlv, kThreadEnterpriseNumber,
-                                                  &DnsSrpUnicast::kServiceData, sizeof(DnsSrpUnicast::kServiceData));
+        aIterator.mServiceTlv = Get<Leader>().FindNextMatchingService(
+            aIterator.mServiceTlv, kThreadEnterpriseNumber, &DnsSrpUnicast::kServiceData,
+            sizeof(DnsSrpUnicast::kServiceData), NetworkData::kServiceDataPrefixMatch);
```
librasungirl, line 250: ditto

src/core/thread/network_data.cpp
EskoDijk: The type name 'ServiceMatchMode' would fit best, given the names of functions (FindService() , MatchService() )

src/core/thread/network_data_service.cpp
EskoDijk: FYI Below in the code, instead of '1' ,  the expression 'sizeof(serviceData)' was used. Whether this or constant or numeral 1, it's best to keep a consistent approach.

src/core/thread/network_data_service.cpp
EskoDijk: @librasungirl  Agree; that looks good now (using 'kServiceDataPrefixMatch') and that is per the specification (length may grow in the future).

EskoDijk (approved):
Overall ok, see comments!

src/core/thread/network_data.cpp
simonlingoogle: Thanks for the suggestion. Will fix.

src/core/thread/network_data_service.cpp
abtink: @librasungirl the existing code already did a partial match pf service data. `FindNextMatchingService()` was added to do partial service data check  from this PR and for purpose of using it for DNS/SRP service info:
- #6473

> This commit adds a new method in `NetworkData` class to find and
> iterate over all Service TLVs that start with a given Service Data.
> Unlike the existing `FindService()` method which searches for a
> Service TLV with an exact match with the given Service Data,  this
> method performs a relaxed check allowing a matching Service TLV to
> contain additional bytes after given data bytes to search for.
> 
> This commit also updates `test_network_data` unit test to add new
> test to cover the behavior of newly added method.
> 
> ----
> 
> This will be useful for parsing Service TLVs for SRP/DNS server info as
> discussed/described in https://github.com/openthread/openthread/issues/6447.  
>

abtink (approved):
Thanks @simonlingoogle for submitting this. Looks good overall. 👍 

I think adding `MatchMode` to all `FindServcie()` methods will be helpful (avoids unintentional bad use of them in future). 

I have two suggestions below in the code.

One more suggestion on the title the the commit and PR. It uses `[backbone-router] relax ...`. But the code change in this PR is more in `NetworkData` and enhancing all the `FindService()` methods (to now have `MatchMode` parameter) I feel may be use if we rename it to be under `[netdata] ...`. Thoughts?

src/core/thread/network_data.hpp
abtink: How about `* This enumeration defines Service Data match mode`.

src/core/thread/network_data.hpp
```diff
@@ -443,78 +457,82 @@ class NetworkData : public InstanceLocator
...
      *
      * @returns A pointer to the next matching Service TLV if one is found or nullptr if it cannot be found.
      *
      */
-    const ServiceTlv *FindNextMatchingService(const ServiceTlv *aPrevServiceTlv,
-                                              uint32_t          aEnterpriseNumber,
-                                              const uint8_t *   aServiceData,
-                                              uint8_t           aServiceDataLength) const;
+    const ServiceTlv *FindNextMatchingService(const ServiceTlv *   aPrevServiceTlv,
```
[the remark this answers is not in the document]
librasungirl: +1 on `FindNextService()`

src/core/thread/network_data_service.cpp
librasungirl: Agree that there is partial match already and Thread device could find the service of `new` format (with additional field).  
My proposal of using a constant of `2` is out of backward compatibility concern, future specification may add new field, but won't remove existing field. If someday a new field is introduced, and the thread end device of new version joins an existing Thread network with `old` service, we may still want it be able to recognize the `old` service and perform SRP.  However with `sizeof(serviceData)` approach, new device could only recognize `new` service.

Thoughts?

src/core/thread/network_data_service.cpp
librasungirl: Refer to https://github.com/openthread/openthread/pull/6752#discussion_r662985288 for concern on `sizeof` approach

src/core/thread/network_data_service.cpp
EskoDijk: @librasungirl Agree, an implementation should treat any size >=2 as valid in order to correctly recognize current-version service data and also future-version service data.   That means using sizeof() is risky for future implementations in case the default (new) service data size grows, because then that newer implementation won't recognize the 'old' size=2 data anymore which it should do.   It may be useful to check if this issue also occurs in other places.

abtink: @librasungirl I think the existing code already does what you are proposing (i.e. it accepts `service data len >= size=2`) and should work fine if additional fields are added in future to service data (i.e., device running the current code would accept them as valid and ignore the additional data). 

Maybe I am missing something here. Let me describe the code:

```c++
    uint8_t           serviceData = DnsSrpAnycast::kServiceNumber;
    const ServiceTlv *tlv         = aIterator.mServiceTlv;

    do
    {
        tlv = Get<Leader>().FindNextMatchingService(tlv, kThreadEnterpriseNumber, &serviceData, sizeof(serviceData));
        VerifyOrExit(tlv != nullptr, error = kErrorNotFound);

    } while (tlv->GetServiceDataLength() < sizeof(DnsSrpAnycast::ServiceData));
```

- The `serviceData` we are using in `FindNextMatchingService()` is just a single byte, so `sizeof(serviceData)` is 1.
  - In the call to `FindNextMatchingService()` we are basically accepting any service where the first byte matches the `DnsSrpAnycast::kServiceNumber (0x5c)`.
  - We then later check if there are any other fields after the service number.
- The check in `while()` loop `tlv->GetServiceDataLength() < sizeof(DnsSrpAnycast::ServiceData)` ensures that the service data is at least the minimum length we expect (or `2`) - we skip over any matching service that does not have the minimum expected length (we currently know)
   - Of course, if in future we add more fields to service data,  the code needs to change to handle the new fields and handle the backward compatibility, e.g., future/new code adding one more byte to service data may accept both size 2 or 3 or more and in case of an entry we size of 2 we need to define how to set the missing info (init with some default value, etc,). This is up to new code/spec (when/if there is a change to service data format). 
   - But any device running the current code (i.e., old code) would continue to work correctly.
- I think we even cover the extra bytes in service data in the unit test `TestNetworkDataDsnSrpServices()` (i.e. we have an entry in Network Data example in this test with some additional bytes in service data after the 2 and we ensure that it is accepted and parsed correctly).
- Do you see any concern with this? Thanks.

librasungirl: Thanks for the details. I overlooked `uint8_t           serviceData = DnsSrpAnycast::kServiceNumber;`  in previous diff review somehow. 
Agree that no issue on `FindNextMatchingService()`.  For `while()` loop, may consider using a constant 2 in comparison as it is known minimal length other than a variable `sizeof(class)` for best compatibility now. But as you pointed, when there is new fields, corresponding process should be updated as well, sgtm to leave it as is.

src/core/thread/network_data_service.cpp
librasungirl: may use `sizeof(BackboneRouter::kServiceData)` to avoid magic number 1 for consistency.

EskoDijk: Using sizeof() for a constant that should remain 1 also in the future, seems risky (even when it is 'protected' by unit tests), because size of BackboneRouter::kServiceData may increase one day while the constant 1 here is not meant to go along with that increase. In case a constant is used it should be like "THREAD_SERVICE_DATA_BBR_MIN_SIZE" or so because it's not about the current size of the data structure but about its minimum size.  In any case, I'm okay with any resolution of this including keeping the magic number :)

src/core/thread/network_data_service.cpp
EskoDijk: Just noting for code below (line 171) it uses the expression sizeof(DnsSrpAnycast::ServiceData)  , which is then only valid as long as the service data length is not increased by the specifications. When the specification changes and so the size of the DnsSrpAnycast::ServiceData becomes longer, the sizeof() construct can no longer be used. That property seems to make it less useful to use a sizeof() construct.  I would assume sizeof() is handy for code that automatically keeps up with increases or decreases in the size of the given object, which is probably not the intended behaviour right here.

src/core/thread/network_data_service.cpp
simonlingoogle: Agree with @EskoDijk

simonlingoogle:
@EskoDijk @librasungirl @abtink 

Thanks for the comments. I have updated the PR accordingly. PTAL.

EskoDijk (approved):

src/core/thread/network_data_service.cpp
abtink: Thanks @librasungirl , Thanks @EskoDijk.

My view on this is that if/when the definition of service data changes (new fields are added), it is the responsibility of new code to handle this. If the new code continues to use `sizeof()` then it is a bug in new code. 

We can define `kMinLength` today and use it in the code to try to protect for future changes but it does not add much protection since the new code needs to anyways handle the additional fields as optional (check against different lengths as parsing the data). 

The main reason I would say we, in general, prefer and use `sizeof()` over defining a constant value directly is that complier is better at getting the size over people (compliers hopefully makes less mistake 😄).

abtink (approved):
LGTM. Thanks. 👍

abtink (approved):
👍

librasungirl (approved):
👍

src/core/thread/network_data.hpp
```diff
@@ -427,14 +437,18 @@ class NetworkData : public InstanceLocator
      * @param[in]  aEnterpriseNumber  Enterprise Number.
      * @param[in]  aServiceData       A pointer to a Service Data.
      * @param[in]  aServiceDataLength The Service Data length pointed to by @p aServiceData.
+     * @param[in]  aServiceMatchMode  The Service Data match mode. @sa ServiceDataMatchMode
```
jwhui, line 440: nit: (other instances below)
```suggestion
     * @param[in]  aServiceMatchMode  The Service Data match mode.
```

src/core/thread/network_data.hpp
```diff
@@ -443,78 +457,82 @@ class NetworkData : public InstanceLocator
...
-     * @param[in]  aPrevServiceTlv    Set to nullptr to start from the beginning of the TLVs (finding the first
-     *                                matching Service TLV), or a pointer to the previous Service TLV returned from
-     *                                this method to iterate to the next matching Service TLV.
+     * @param[in]  aPrevServiceTlv    Set to nullptr to start from the beginning of the TLVs (finding the first matching
+     *                                Service TLV), or a pointer to the previous Service TLV returned from this method
+     *                                to iterate to the next matching Service TLV.
      * @param[in]  aEnterpriseNumber  Enterprise Number.
      * @param[in]  aServiceData       A pointer to a Service Data to match with Service TLVs.
      * @param[in]  aServiceDataLength The Service Data length pointed to by @p aServiceData.
+     * @param[in]  aServiceMatchMode  The Service Data match mode. @sa ServiceDataMatchMode
```
simonlingoogle, line 526:
```suggestion
     * @param[in]  aServiceMatchMode  The Service Data match mode.
```

src/core/thread/network_data.hpp
```diff
@@ -443,78 +457,82 @@ class NetworkData : public InstanceLocator
...
     }
 
     /**
      * This method returns a pointer to a Service TLV in a specified tlvs buffer.
      *
      * @param[in]  aEnterpriseNumber  Enterprise Number.
-     * @param[in]  aServiceData       A pointer to an Service Data.
+     * @param[in]  aServiceData       A pointer to a Service Data.
      * @param[in]  aServiceDataLength The Service Data length pointed to by @p aServiceData.
+     * @param[in]  aServiceMatchMode  The Service Data match mode. @sa ServiceDataMatchMode
```
simonlingoogle, line 501:
```suggestion
     * @param[in]  aServiceMatchMode  The Service Data match mode.
```

src/core/thread/network_data.hpp
```diff
@@ -443,78 +457,82 @@ class NetworkData : public InstanceLocator
      * @param[in]  aEnterpriseNumber  Enterprise Number.
      * @param[in]  aServiceData       A pointer to a Service Data.
      * @param[in]  aServiceDataLength The Service Data length pointed to by @p aServiceData.
+     * @param[in]  aServiceMatchMode  The Service Data match mode. @sa ServiceDataMatchMode
```
simonlingoogle, line 460:
```suggestion
     * @param[in]  aServiceMatchMode  The Service Data match mode.
```

src/core/thread/network_data.hpp
```diff
@@ -443,78 +457,82 @@ class NetworkData : public InstanceLocator
...
+                                  ServiceMatchMode aServiceMatchMode) const;
 
     /**
      * This method returns a pointer to a Service TLV in a specified tlvs buffer.
      *
      * @param[in]  aEnterpriseNumber  Enterprise Number.
-     * @param[in]  aServiceData       A pointer to an Service Data.
+     * @param[in]  aServiceData       A pointer to a Service Data.
      * @param[in]  aServiceDataLength The Service Data length pointed to by @p aServiceData.
+     * @param[in]  aServiceMatchMode  The Service Data match mode. @sa ServiceDataMatchMode
```
simonlingoogle, line 476:
```suggestion
     * @param[in]  aServiceMatchMode  The Service Data match mode.
```

jwhui (approved):

PR #6754, opened 2021-06-21
[srp-server] free `aHost` in `CommitSrpUpdate()` in case of error

abtink:
This commit ensures that in `Srp::Server::CommitSrpUpdate()` the
`aHost` parameter is correctly freed in case of an error and an early
jump to exit (e.g., from `SuccessOrExit(aError)` at the start of this
method). The `aHost` is now freed at the `exit` label and a boolean
`shouldFreeHost` tracks whether or not it needs to be freed. The
boolean variable is set to `false` when  the ownership of `aHost` is
taken over (e.g., from the call to `AddHost(aHost)`).


wgtdkp (approved):
thanks! 👍

jwhui (approved):

PR #6763, opened 2021-06-23
[tests] restart SRP server (instead of device reset) to validate port change

### LINKED ISSUE #6743: [github-actions] test_srp_server_reboot_port.py fails intermittently
[logs_82299.zip](https://github.com/openthread/openthread/files/6667852/logs_82299.zip)
[packet-verification-1.1-on-1.2-pcaps.zip](https://github.com/openthread/openthread/files/6667853/packet-verification-1.1-on-1.2-pcaps.zip)

abtink:
Added a second commit to this PR. First commit changes the test to do restart:

**[test] restart SRP server (instead of device reset) to validate port change (#6763)**

This commit updates `test_srp_server_reboot_port` to just restart the
SRP sever instead of performing a full device reset. The SRP restart
will trigger the port number change by the server which is the
behavior this test-case is intended to verify. This change should
help address intermittent failures of this test.

Second commit:

**[srp-server] ensure to clear `mHasRegisteredAnyService` on server stop (#6763)**

This commit updates `Srp::Server` to clear `mHasRegisteredAnyService`
when server is stopped and all the hosts and services are deleted.
This boolean flag is used to save the port number in non-volatile
settings so that we can ensure to pick a different port number the
next time the server starts.

This change addresses an issue where the port number may not be saved
and therefore not changed if the server is stopped and restarted
two times in a row.

---

Closes #6743.

### DIFF
--- a/src/core/net/srp_server.cpp
+++ b/src/core/net/srp_server.cpp
@@ -487,6 +487,7 @@ void Server::Stop(void)
 
     otLogInfoSrp("[server] stop listening on %hu", mSocket.GetSockName().mPort);
     IgnoreError(mSocket.Close());
+    mHasRegisteredAnyService = false;
 
 exit:
     return;
--- a/tests/scripts/thread-cert/test_srp_server_reboot_port.py
+++ b/tests/scripts/thread-cert/test_srp_server_reboot_port.py
@@ -35,7 +35,7 @@ import command
 import thread_cert
 
 # Test description:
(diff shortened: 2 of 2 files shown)


superwhd (approved):
LGTM :+1: 
Thanks!

simonlingoogle (approved):
LGTM :+1: 

Thanks for fixing the bug and the test. 
Even though the original issue after `reset` is still a mystery.

jwhui (approved):

PR #6764, opened 2021-06-24
[test] enhance `{Verify/Success}OrQuit()` and their use in unit test

abtink:
This commit updates `VerifyOrQuit()` and `SuccessOrQuit()` macros to
include the failed condition in the error message that is printed on
a failure (in addition to function name and line number where the
error happened). This commit also changes the second parameter
(`aMessage`) to in these macros to be optional.

This commit also updates unit tests to remove the second `aMessage`
string in cases where the failure can be inferred from the condition
itself.

----

Here are some examples of error messages:
```
FAILED TestDnsName:198 - VerifyOrQuit(len != test.mEncodedLength) Encoded length does not match expected value
FAILED TestStringEndsWith:190 - VerifyOrQuit(!StringEndsWith("foobar", 'r')) 
```

### DIFF
--- a/tests/unit/test_aes.cpp
+++ b/tests/unit/test_aes.cpp
@@ -65,22 +65,22 @@ void TestMacBeaconFrame(void)
         0xAC, 0xDE, 0x48, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x05, 0x02,
     };
 
-    VerifyOrQuit(instance != nullptr, "Null OpenThread instance");
+    VerifyOrQuit(instance != nullptr);
 
     aesCcm.SetKey(key, sizeof(key));
     aesCcm.Init(headerLength, payloadLength, tagLength, nonce, sizeof(nonce));
     aesCcm.Header(test, headerLength);
-    VerifyOrQuit(aesCcm.GetTagLength() == tagLength, "AesCcm::GetTagLength() failed");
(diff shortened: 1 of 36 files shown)


simonlingoogle (approved):
LGTM :+1: 

Can we also add messages for non-testing VerifyOrExit and SuccessOrExit ?
It can be enabled by default on POSIX and disabled by default on SoC.

jwhui:
> Can we also add messages for non-testing VerifyOrExit and SuccessOrExit ?
> It can be enabled by default on POSIX and disabled by default on SoC.

@simonlingoogle , did you mean `VerifyOrQuit()` and `SuccessOrQuit()`?

abtink:
> Can we also add messages for non-testing VerifyOrExit and SuccessOrExit ?
> It can be enabled by default on POSIX and disabled by default on SoC.

@simonlingoogle , if you mean emitting a  log from `VerifyOrExit()` when the condition fails (e.g., under posix), I don't think it would be useful and can add a lot of additional log lines which may not be that helpful (and make reading logs and finding the useful lines harder). 

In particular, we use `VerifyOrExit()` as a syntax sugar to check for certain condition and avoid  nested `if` statements. It is not used like an `assert()` where we expect the condition to be (almost) always true. So a failure of `VerifyOfExit()` check does not indicate an error and is just a fancier `if()` statement.

jwhui (approved):

PR #6766, opened 2021-06-25
[simulation] remove use of internal definitions/headers in radio platform

abtink:
This commit updates the `simulation/radio.c` and removes the uses of
core internal macro `VerifyOfExit()` and replaces them with
`otEXPECT_ACTION()`. This ensures that radio platform layer does not
include the internal OT header like "common/code_utils.hpp".

----

Related PRs:
- https://github.com/openthread/ot-nrf528xx/pull/90

### DIFF
--- a/examples/platforms/simulation/radio.c
+++ b/examples/platforms/simulation/radio.c
@@ -39,7 +39,6 @@
 #include <openthread/platform/radio.h>
 #include <openthread/platform/time.h>
 
-#include "common/code_utils.hpp"
 #include "utils/code_utils.h"
 #include "utils/link_metrics.h"
 #include "utils/mac_frame.h"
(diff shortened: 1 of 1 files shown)


simonlingoogle (approved):
LGTM :+1: 

This looks good to me. But I am also wondering if it's good to have some utilities accessible by both Core and platform.

abtink:
> This looks good to me. But I am also wondering if it's good to have some utilities accessible by both Core and platform.

I agree it would be useful. I think `platform/util` modules can be used for this. It is a group of common utility module that can be used by all platform implementation (and we can use and share core definitions in the implementation of the utility functions (e.g., we do this with `util/mac_frame.hpp`)). We can also consider exposing some common header files and definitions.

One thing to have in mind is that for now we should assume platform can be in C (same as our OT public/private APIs which are in C), so we cannot expose any C++ related definitions from any of these utility headers.

Irving-cl (approved):
LGTM 👍

jwhui (approved):

PR #6768, opened 2021-06-27
[netdata] adding `NetworkData::Publisher`

abtink:
This commit implements a new feature "Network Data Publisher" which
provides mechanisms to limit the number of similar entries (service
and/or prefix) in the Thread Network Data by monitoring the Network
Data and managing if or when to add or remove entries. This feature is
enabled using `OPENTHREAD_CONFIG_NETDATA_PUBLISHER_ENABLE` config, or
`NETDATA_PUBLISHER` in autoconf, or `OT_NETDATA_PUBLISHER` cmake
option.

This commit adds support for publishing DNS/SRP anycast/unicast
service, on-mesh prefix, and external route prefix entries.

When there is a request to publish an entry, the `Publisher` monitors
the Network Data and counts the number of similar entries. If there
are fewer entries than a desired target number, the entry is added
after a short random wait time.

If there are too many similar entries, `Publisher` starts the process
of removing its own entry (again after some random wait time). When
removing entries, certain entries are preferred over others (e.g., an
entry from a router over one from an end-device or if they are from
the same type of node, the one with smaller RLOC16). If `Publisher`
determines that its own entry is a preferred one, it adds an extra
wait time before removing its entry. This gives higher chance for a
non-preferred entry from another device to be removed before removing
a preferred entry which helps towards quicker convergence of the
process to the desired number of entries.

On-mesh prefix and external route entries have a "preference" field.
When publishing such an entry, a matching entry in the network data is
counted only if its preference is same or higher than the entry's
preference. This ensures that a device with a higher preference entry
publishes its entry even when there are many lower preference similar
entries in the network data (potentially causing a lower preference
entry to be removed).

This commit also adds `test_netdata_publisher.py` to verify the
behavior of the `Publisher`.

----------

This PR replaces (and contains) PR https://github.com/openthread/openthread/pull/6587. It contains the 
full publisher implementation (for both DNS/SRP service and 
on-mesh/external-route prefix entries (please see https://github.com/openthread/openthread/pull/6587#issuecomment-869232331 
for more details).


src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1079 @@
...
+    SetState(kAdded);
+    Get<Publisher>().NotifyPrefixEntryChange(kEventEntryAdded, mPrefix);
+
+exit:
+    return registerWithLeader;
+}
+
+bool Publisher::PrefixEntry::Remove(State aNextState)
+{
+    // Remove the prefix entry to the network data.
```
librasungirl, line 923:
```suggestion
    // Remove the prefix entry from the network data.
```

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1079 @@
...
+{
+    const PrefixTlv *      prefixTlv;
+    const BorderRouterTlv *brSubTlv;
+    int8_t                 preference             = BorderRouterEntry::PreferenceFromFlags(mFlags);
+    uint16_t               flagsWithoutPreference = BorderRouterEntry::FlagsWithoutPreference(mFlags);
+
+    prefixTlv = Get<Leader>().FindPrefix(mPrefix);
+    VerifyOrExit(prefixTlv != nullptr);
+
+    brSubTlv = prefixTlv->FindSubTlv<BorderRouterTlv>(mStable);
```
librasungirl, line 992:
```suggestion
    brSubTlv = prefixTlv->FindSubTlv<BorderRouterTlv>();
```

Should iterate all border router subtlvs and prefer stable over unstable in below `for()` loop?

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1079 @@
...
+{
+    const PrefixTlv *  prefixTlv;
+    const HasRouteTlv *hrSubTlv;
+    int8_t             preference             = HasRouteEntry::PreferenceFromFlags(static_cast<uint8_t>(mFlags));
+    uint8_t            flagsWithoutPreference = HasRouteEntry::FlagsWithoutPreference(static_cast<uint8_t>(mFlags));
+
+    prefixTlv = Get<Leader>().FindPrefix(mPrefix);
+    VerifyOrExit(prefixTlv != nullptr);
+
+    hrSubTlv = prefixTlv->FindSubTlv<HasRouteTlv>(mStable);
```
librasungirl, line 1039: ditto

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1079 @@
...
+    // Count the number of matching "DNS/SRP Anycast" service entries
+    // in the Network Data (the match requires the entry to use same
+    // "sequence number" value). We prefer the entries associated with
+    // smaller RLCO16.
+
+    Service::DnsSrpAnycast::ServiceData serviceData(mInfo.GetSequenceNumber());
+    const ServiceTlv *                  serviceTlv = nullptr;
+
+    while ((serviceTlv = Get<Leader>().FindNextMatchingService(serviceTlv, Service::kThreadEnterpriseNumber,
+                                                               reinterpret_cast<const uint8_t *>(&serviceData),
```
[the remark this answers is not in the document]
abtink: Here we internationally want to to include the "seq number" when matching the service data (i.e., we want both "service number" and "seq number" to be present and match). If we see an existing DNS/SRP anycast service entry in network data with a different "seq number", then it is not a match with our entry and we should not count it as such (it may be an older seq number).

Regarding forward compatibility, this is already handled since we are using `FindNextMatchingService()` which does a partial match of service data (it accepts the service entry if service data is 2 or more).

src/core/thread/network_data_publisher.cpp
abtink: It is a good point @librasungirl.

I actually did consider this when writing the code but decided to go with the current (and simpler) model where we treat prefix entries marked as stable completely separate from entries not marked as stable. 

- If our entry itself is stable, then this is what we want as the correct behavior,
    - i.e., if we see the same prefix in network data but not marked as stable, we should not count it as a match (and should still add ours if there are not enough matching stable ones).
- If our entry is itself not marked as stable then we can consider an existing stable prefix entry as a match (I did not go for it since it adds complexity to the code we need iterate over all matching BR subTLvs).

I think practically the `Publisher` will be used to publish stable entry. We can cosndier limiting the API to check and allow only stable prefix entries.

------

A related note:  I also considered earlier on how to determine a match related to other flags:
- If we see an existing prefix entry that has all the same flags as our entry but it also has some additional flags that we don't have in our entry, should we count it as a match or not? 
- After some thoughts on this, I decided not to go for this and require an exact match of flags.
   - The challenge with allowing and checking additional flags is that we need to take into account the meaning of  each flag and their combinations (whether in all cases a prefix which has more flags than ours is equivalent). 
   - It adds complexity in the code for a not-so-practical situation.

librasungirl: If prefixes are the same, stable Prefix and unstable Prefix from different Servers would share the same outer stable Prefix TLV in the network data with two  border router sub-tlvs inside - the one for the `unstable` server would have `stable` flag as false as below picture

Screen Shot 2021-07-08 at 5 23 09 PM

Adding stable comparison with border router subtlvs in the network data seems not too complex.

---
Agree with your thoughts on the other flags.

src/core/thread/network_data_publisher.cpp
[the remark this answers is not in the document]
librasungirl: IIUC, `sequence number` aims to trigger re-registration in some situations, it may change over time for some reason,  we may always prefer larger `sequence number`.

src/core/thread/network_data_publisher.cpp
abtink: Yes. We will have a single Prefix TLV but it can possibly have two BR sub-TLVs one stable and one not stable. And each sub-TLV can have a bunch of entries (which we want to iterate through and count matching ones).

- If our entry is stable, then we want to find the stable sub-TLV, so the current code `prefixTlv->FindSubTlv<BorderRouterTlv>(mStable);` is behavior we want.

- If our entry itself is not stable, the current code/model counts the matching entries which are also not stable. If we want to change this to count both stable and not stable matches, we need to add another enclosing loop to go through both BR sub-TLVs (stable and not stable sub-TLVs). 
   - This is not terribly complex to add but it is not a single line change. 
   - Note that we cannot just use the `FindSubTlv(void)` flavor instead of `FindSubTlv(mStable)`. This won't really iterate over both sub-TLVs.  `FindSubTlv(void)` will find the "first" matching TLV (ignoring the stable flag).

My goal here has been to avoid adding additional logic and behavior in the code to optimize the case that is not really intended as the main use (i.e., using `Publisher` with prefix entries that are not marked as stable). And if we change it since it would be a new behavior we probably need to also add test-case for it (again for something that is not intended to be used).

Considering more on this, I think  it may be better to change the `Publisher` public methods to only allow stable prefix entries. This help keep the code simple and make the intended use clear. I will change this in the next push.

simonlingoogle: Not considering the added code complexity, which I believe is small, do we have good reason to not prefer stable to unstable?

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1079 @@
...
+    {
+        mInfo.SetAddress(Get<Mle::Mle>().GetMeshLocal64());
+
+        if (GetState() == kAdded)
+        {
+            // If the entry is already added, we need to update it
+            // so we remove it and add it back immediately with
+            // the new mesh-local address.
+
+            Remove(/* aNextState */ kAdding);
```
librasungirl, line 596: Combine SRV_DATA.ntf with below Add() which happens soon as agreed on https://github.com/openthread/openthread/pull/6587#discussion_r646270190
```suggestion
            Remove(/* aNextState */ kAdding, /* aRegisterWithLeader*/ false);
```

librasungirl (commented):
I know we have talked similar topic as in https://github.com/openthread/openthread/pull/6587#discussion_r646836099, however different from DNS/SRP service which is new feature, On-mesh prefix and external route are features introduced since Thread 1.0/1.1, would it become a concern? To be fair, is it possible to propose to TG a consistent `publisher config` for prefixes among different vendors likewise?

jwhui:
> I know we have talked similar topic as in [#6587 (comment)](https://github.com/openthread/openthread/pull/6587#discussion_r646836099), however different from DNS/SRP service which is new feature, On-mesh prefix and external route are features introduced since Thread 1.0/1.1, would it become a concern? To be fair, is it possible to propose to TG a consistent `publisher config` for prefixes among different vendors likewise?

Yes, we should propose spec text that BRs should follow when publishing prefixes and routes.

src/core/thread/network_data_publisher.cpp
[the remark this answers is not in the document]
abtink: That's correct. That's how the current SRP client code uses the entries from network data (the one with higher seq number is used). How the seq number is determined would be up to SRP server (e.g. to discover and sync with other servers  - which I guess would be part of the replication mechanism (to be determined)).

However, I think the `Publisher` should not care or be aware of any of this (the same way that it is not determining/managing the prefix (e.g., OMR prefix) and/or external route entries). The `Publisher` is instructed to publish certain entries and its responsibility is to determine when/if to add/remove the published entry in network data (by monitoring it for matching entries)

src/core/thread/network_data_publisher.cpp
abtink: @simonlingoogle the existing code (from the initial implementation) already did prefer the stable over unstable matches (I explained in above comments under the case "if our entry is stable" and that's the behavior we do want).

The question was whether we want to have `Publisher` support publishing entries which are not stable and add complexity to the code to optimize how we handle them. This is not really a practical or intended use for `Publisher` so we now simply don't allow `Publisher` to be used with non-stable entries.

src/core/thread/network_data_publisher.cpp
abtink: > Combine SRV_DATA.ntf with below Add() which happens soon as agreed on #6587 (comment)

@librasungirl  I think this is already the case with code  (i.e. there will be a single `SRV_DATA.ntf` after both `Remove()` and `Add()` calls - which is done from `Publisher::HandleNotifierEvents()`).

Note that `Remove()` and `Add()` no longer have `aRegisterWithLeader` parameter and always expect the caller to take care of it (they have a `bool` return value  `registerWithLeader` to indicate if something was actually changed in the local netdata). This is different and changed from the code in https://github.com/openthread/openthread/pull/6587#discussion_r646270190

abtink:
> I know we have talked similar topic as in #6587 (comment), however different from DNS/SRP service which is new feature, On-mesh prefix and external route are features introduced since Thread 1.0/1.1, would it become a concern? To be fair, is it possible to propose to TG a consistent publisher config for prefixes among different vendors likewise?

I agree that it would be good to define the config to use on BR specially since we are defining new type of prefix entries like  OMR prefix or external route entries (associated with on-link prefix on infra net side of a BR) which are now managed directly by a Thread BR (rather than managed by user of Thread and/or OpenThread).
-  This would ensure harmonized behavior across all BRs for such prefixes and we should definitely do so 
- I mentioned that I already brought it up in one of the tech committee calls and there was general agreement on this. 

That said, I don't see any immediate concerns with adding support and/or using publisher for prefix entries (@librasungirl  please share if you see some). 
- The `Publisher` itself is an added feature and the users can choose to not use it and instead of use the APIs to directly add/remove entries in network data. 
- In some cases the decision to use `Publisher` can be done independently of how other devices/BRs behave, e.g., if a BR decides to use publisher for an on-mesh prefix, it is indicating that from its perspective it is ok as long as the same on-mesh prefix is present in the network data, and it may not matter how the other BRs added the entry (whether it was added manually or through use of `Publisher` on other nodes).

src/core/thread/network_data_publisher.hpp
```diff
@@ -0,0 +1,466 @@
...
+            kAdded,    // Entry is added in network data, monitoring to determine if/when to remove.
+            kRemoving, // Entry is being removed from network data (random wait interval before remove).
+        };
+
+        enum : uint32_t
+        {
+            // All intervals are in milliseconds.
+            kMaxWaitTimeToAdd          = OPENTHREAD_CONFIG_NETDATA_PUBLISHER_MAX_WAIT_TIME_TO_ADD,
+            kMaxWaitTimeToRemove       = OPENTHREAD_CONFIG_NETDATA_PUBLISHER_MAX_WAIT_TIME_TO_REMOVE,
+            kExtraWaitToRemovePeferred = OPENTHREAD_CONFIG_NETDATA_PUBLISHER_EXTRA_WAIT_TIME_TO_REMOVE_PREFERRED,
```
wgtdkp, line 280: nits:
```suggestion
            kMaxDelayToAdd          = OPENTHREAD_CONFIG_NETDATA_PUBLISHER_MAX_WAIT_TIME_TO_ADD,
            kMaxDelayToRemove       = OPENTHREAD_CONFIG_NETDATA_PUBLISHER_MAX_WAIT_TIME_TO_REMOVE,
            kExtraDelayToRemovePeferred = OPENTHREAD_CONFIG_NETDATA_PUBLISHER_EXTRA_WAIT_TIME_TO_REMOVE_PREFERRED,
```

tests/scripts/thread-cert/test_netdata_publisher.py
```diff
@@ -0,0 +1,456 @@
...
+DESIRED_NUM_ON_MESH_PREFIX = 3
+DESIRED_NUM_EXTERNAL_ROUTE = 10
+
+THREAD_ENTERPRISE_NUMBER = 44970
+ANYCAST_SERVICE_NUM = 0x5c
+UNICAST_SERVICE_NUM = 0x5d
+
+
+def parse_hex_string(hexstr: str) -> bytes:
+    return bytes(int(hexstr[i:i + 2], 16) for i in range(0, len(hexstr), 2))
```
wgtdkp, line 82: We can use [bytes.from_string()](https://docs.python.org/3/library/stdtypes.html#bytes.fromhex).

tests/scripts/thread-cert/node.py
```diff
@@ -1716,6 +1716,34 @@ def register_netdata(self):
         self.send_command('netdata register')
         self._expect_done()
 
+    def netdata_publish_dnssrp_anycast(self, seqnum):
+        self.send_command('netdata publish dnssrp anycast %s' % seqnum)
```
wgtdkp, line 1720: nit: use f-string:
```suggestion
        self.send_command(f'netdata publish dnssrp anycast {seqnum}')
```

tests/scripts/thread-cert/test_netdata_publisher.py
```diff
@@ -0,0 +1,456 @@
...
+
+WAIT_TIME = 55
+
+ON_MESH_PREFIX = 'fd00:1234::/64'
+ON_MESH_FLAGS = 'paso'
+
+EXTERNAL_ROUTE = 'fd00:abce::/64'
+EXTERNAL_FLAGS = 's'
+
+ANYCAST_SEQ_NUM = '4'
```
wgtdkp, line 65:
```suggestion
ANYCAST_SEQ_NUM = 4
```

tests/scripts/thread-cert/test_netdata_publisher.py
```diff
@@ -0,0 +1,456 @@
...
+
+DNSSRP_ADDRESS = 'fd00::cdef'
+DNSSRP_PORT = '8765'
```
wgtdkp, line 68:
```suggestion
DNSSRP_PORT = 49152
```

include/openthread/netdata_publisher.h
```diff
@@ -0,0 +1,279 @@
...
+ * This function requires the feature `OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE` to be enabled.
+ *
+ * A call to this function will remove and replace any previous "DNS/SRP Service" entry that was being published (from
+ * earlier call to any of `otNetDataPublishDnsSrpService{Type}()` functions).
+ *
+ * This function publishes the "DNS/SRP Service Unicast Address" by including the address and port info in the Service
+ * TLV data.
+ *
+ * @param[in] aInstance  A pointer to an OpenThread instance.
+ * @param[in] aAddress   The DNS/SRP server address to publish (MUST NOT be NULL).
```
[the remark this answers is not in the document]
abtink: I agree. I think expecting pointers to b non-NULL and document only when they can be us would be a better model. I think we should address this in a future PR and update all current API documentation to follow this. 

For now, and in this PR I think I will keep these(we can change them in future PR with other APIs).

src/core/net/srp_server.cpp
```diff
@@ -407,48 +416,54 @@ void Server::CommitSrpUpdate(Error                    aError,
...
+
+    mRunning = true;
+
     SuccessOrExit(error = mSocket.Open(HandleUdpReceive, this));
-    SuccessOrExit(error = mSocket.Bind(port, OT_NETIF_THREAD));
-    SuccessOrExit(error = PublishServerData());
+    SuccessOrExit(error = mSocket.Bind(mPort, OT_NETIF_THREAD));
 
-    otLogInfoSrp("[server] start listening on port %hu", mSocket.GetSockName().mPort);
+    otLogInfoSrp("[server] start listening on port %u", mPort);
```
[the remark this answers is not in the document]
abtink: In general, I recommend not using using `%hu` (I avoid using it myself in the code).
- Some (older) toolchain and lib implementations don't support it or handle it correctly (e.g. [link](https://www.microchip.com/forums/m581592.aspx))
- In a var arg function/method, the ints are anyways promoted to larger ints (more bits) when passed as arguments.

src/core/net/srp_server.hpp
```diff
@@ -42,10 +42,18 @@
...
+#endif
+
 #if !OPENTHREAD_CONFIG_ECDSA_ENABLE
 #error "OPENTHREAD_CONFIG_ECDSA_ENABLE is required for OPENTHREAD_CONFIG_SRP_SERVER_ENABLE"
 #endif
 
+#if !OPENTHREAD_CONFIG_NETDATA_PUBLISHER_ENABLE
+#error "OPENTHREAD_CONFIG_NETDATA_PUBLISHER_ENABLE is required for OPENTHREAD_CONFIG_SRP_SERVER_ENABLE"
+#endif
+
```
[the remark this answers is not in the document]
abtink: Thanks. I don't know how it was added twice  we really want to double-check to make sure 😄 ...

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1079 @@
...
+                        aNumPreferredEntries++;
+                    }
+                }
+                else
+                {
+                    aNumPreferredEntries++;
+                }
+            }
+
+            if (serverSubTlv->GetServerDataLength() >= sizeof(Service::DnsSrpUnicast::ServerData))
```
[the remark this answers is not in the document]
abtink: We want to keep it `if` and not change it to `else if`. It's intentional here.

From the definition of the DNS/SRP service TLV, it is valid to have a service TLV that has address info in the "service data" portion and then also in its "server data" (for some sub-tlvs and entries). In OT we currently don't use this model but it is still valid and in `Publisher` we want to count both cases.

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1079 @@
...
+{
+    // Adds the prefix entry to the network data.
+
+    bool registerWithLeader = false;
+
+    union
+    {
+        OnMeshPrefixConfig  mOnMeshPrefix;
+        ExternalRouteConfig mExternalRoute;
+    } config;
```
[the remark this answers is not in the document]
abtink: We use `union` here to reduce the use of stack in this method (the two configs are large objects and we only need one). 

This matters because from `Add()` we can call user's callback to notify that entry was added (`mPrefixCallback` from `NotifyPrefixEntryChange()`). Whenever we invoke a user callback it's good to optimize our stack use if we can (leave more stack for user to use).

include/openthread/netdata_publisher.h
```diff
@@ -0,0 +1,279 @@
...
+/**
+ * This function requests an on-mesh prefix to be published in the Thread Network Data.
+ *
+ * This function requires the feature `OPENTHREAD_CONFIG_BORDER_ROUTER_ENABLE` to be enabled.
+ *
+ * @param[in] aInstance           A pointer to an OpenThread instance.
+ * @param[in] aConfig             The on-mesh prefix config to publish (MUST NOT be NULL).
+ *
+ * @retval OT_ERROR_NONE          The on-mesh prefix is published successfully.
+ * @retval OT_ERROR_INVALID_ARGS  The @p aConfig is not valid (bad prefix, invalid flag combinations, or not stable).
```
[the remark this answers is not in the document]
abtink: We mention that that config must be stable. I will add another line in the main method/function descriptionmaking it even more explicit.

Regarding other flags (or other cases where `aConfig` can be considered invalid), I am not sure if we want to document it here in `Publisher` API.
- The valid flags combinations are mainly specified by the Thread spec (based on the meaning of each flag) - It's not a `Publisher` specific behavior.
-  The other APIs like `otBorderRouterAddOnMeshPrefix()` or `otBorderRouterAddRoute()` don't document this
- We can consider describing the valid combination where `otBorderRouterConfig` and `otExternalRouteConfig` structs are defined (I would leave it for a future PR)

src/core/net/srp_server.hpp
```diff
@@ -774,7 +782,9 @@ class Server : public InstanceLocator, private NonCopyable
     LinkedList<UpdateMetadata> mOutstandingUpdates;
 
     ServiceUpdateId mServiceUpdateId;
+    uint16_t        mPort;
     bool            mEnabled : 1;
+    bool            mRunning : 1;
```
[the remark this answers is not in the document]
abtink: Changed it in the new push. PTAL.

src/core/thread/network_data_publisher.hpp
abtink: Updated them in the new push.

tests/scripts/thread-cert/test_netdata_publisher.py
abtink: Changed in the new push (actually removed the function and directly used the `bytes.fromhex()`.

tests/scripts/thread-cert/node.py
abtink: Changed in new push.

include/openthread/netdata_publisher.h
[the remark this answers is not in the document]
wgtdkp: Thanks, SGTM!

src/core/thread/network_data_publisher.cpp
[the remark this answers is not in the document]
wgtdkp: aha that makes sense to me!
Maybe leave a comment to describe the intention?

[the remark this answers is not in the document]
abtink: I actually changed this in new push, added two new small methods `AddOnMeshPrefix()` and `AddExternalRoute()` which we then call from `Add()`. This helps moves the `config` definitions to the new methods and their stack use should be freed before we invoke user callback.

include/openthread/netdata_publisher.h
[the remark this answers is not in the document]
wgtdkp: SGTM, Thanks!

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1089 @@
...
+                        aNumPreferredEntries++;
+                    }
+                }
+                else
+                {
+                    aNumPreferredEntries++;
+                }
+            }
+
+            if (serverSubTlv->GetServerDataLength() >= sizeof(Service::DnsSrpUnicast::ServerData))
```
wgtdkp, line 793: What about Server TLVs that don't include the IP address? Do we support such format?

src/core/thread/network_data_publisher.cpp
simonlingoogle: Fine to me if we simply don't allow publishing non-stable entries.

src/core/thread/network_data_publisher.cpp
abtink: Yes. Definitely (we use this when we publish using "ML-EID". 

In such a case, we only count the entry if the enclosing service TLV data contains the IP address. That's the first `if` here in this loop.

Basically here in the code we are iterating over all Server sub-TLVs:
- We first check if the enclosing Service TLV does contain IPv6 address, if so, we count it as one entry (each server TLV represents one RLOC (one BR) supporting this address). 
- Then we also check if the "server sub-TLV" has an address embedded in its data and if it does, we count it as an entry as well.

src/core/net/srp_server.hpp
```diff
@@ -651,6 +656,13 @@ class Server : public InstanceLocator, private NonCopyable
         kDefaultEventsHandlerTimeout = OPENTHREAD_CONFIG_SRP_SERVER_SERVICE_UPDATE_TIMEOUT,
     };
 
+    enum State : uint8_t
+    {
+        kStateDisabled,
+        kStateRunning,
+        kStateStopped,
```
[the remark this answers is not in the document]
abtink: I like and prefer the `kStateStopped` name (that @wgtdkp  suggested as well). It mirrors the `Stop()` API name.
`Idle` or `standby` can mean device is running/active but has nothing to do. This is not the case here, we are stopped and not running when in this state.

src/core/thread/network_data_publisher.cpp
wgtdkp: So we are not counting the SRP servers that include an IP address in neither `Service TLV` nor `Server TLV`? I think such format is okay and am wondering why we are not counting it as an unicast entry.

abtink: That's correct. The "Unicast Service Entry" definition allows the address info (address & port) to be either included as part of "service data" or "server data". If both are empty, then it is not a valid entry (other nodes cannot use such an entry to determine the address and port info of the server). And we don't count them.

src/core/net/srp_server.hpp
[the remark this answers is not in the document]
simonlingoogle: SGTM

simonlingoogle (approved):
LGTM :+1:

src/core/thread/network_data_publisher.cpp
wgtdkp: I thought we can construct the server address with RLOC16 in the Server TLV.Is it because the RLOC address is not stable that we don't support such format?

abtink: Using RLOC16 from Server sub-TLV to determine the server address is not supported since it won't work in all cases:
- MTD devices that don't get the full network data won't see or know the RLOC16 of entries:
   -  Parent will replace  the RLOC16 in all TLV entries with `0xfffe` (or ALOCs) before sending it to such a child
   - This is done from `NetworkData::RemoveTemporaryData()` methods (and is what Thread spec requires).
- We also need the port number (we could consider assuming a fixed/known port number but the draft spec requires port number to be included for unicast address model).

wgtdkp: Thanks! that's clear.

wgtdkp (approved):
LGTM :+1: Thanks!

abtink:
In the last push, rebased on `main` and addressed conflicts (also changed the constants to use `constexpr` instead of `enum`)

src/core/thread/network_data_publisher.cpp
```diff
@@ -0,0 +1,1089 @@
...
+    VerifyOrExit(prefixTlv != nullptr);
+
+    hrSubTlv = prefixTlv->FindSubTlv<HasRouteTlv>(/* aStable */ true);
+    VerifyOrExit(hrSubTlv != nullptr);
+
+    for (const HasRouteEntry *entry = hrSubTlv->GetFirstEntry(); entry <= hrSubTlv->GetLastEntry();
+         entry                      = entry->GetNext())
+    {
+        uint8_t entryFlags      = entry->GetFlags();
+        int8_t  entryPreference = BorderRouterEntry::PreferenceFromFlags(entryFlags);
```
librasungirl, line 1056:
```suggestion
        int8_t  entryPreference = HasRouteEntry::PreferenceFromFlags(entryFlags);
```

librasungirl (approved):
LGTM overall with some minor inline comments

src/core/common/notifier.cpp
```diff
@@ -183,12 +183,15 @@ void Notifier::EmitEvents(void)
...
 #endif
-#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
-    Get<Srp::Server>().HandleNotifierEvents(events);
-#endif
 #if OPENTHREAD_CONFIG_SRP_CLIENT_ENABLE
     Get<Srp::Client>().HandleNotifierEvents(events);
 #endif
+#if OPENTHREAD_CONFIG_NETDATA_PUBLISHER_ENABLE
+    // The `NetworkData::Publisher` is notified after SRP/DNS
+    // server/client to allow them to request changes to what is
```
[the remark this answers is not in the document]
abtink: Thanks. Updated the comment in new push.

src/core/thread/network_data_publisher.cpp
abtink: Thanks for noticing. Fixed this in new push.

I was wondering why the test did not catch this, turns out this caused the `entryPreference` to become zero (which is same "med" level which we used in the test. I changed the test to use "low" instead.

src/core/thread/network_data_publisher.hpp
```diff
@@ -0,0 +1,464 @@
...
+
+    /**
+     * This constructor initializes `Publisher` object.
+     *
+     * @param[in]  aInstance     A reference to the OpenThread instance.
+     *
+     */
+    explicit Publisher(Instance &aInstance);
+
+#if OPENTHREAD_CONFIG_TMF_NETDATA_SERVICE_ENABLE
```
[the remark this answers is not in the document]
abtink: We cannot not use `OPENTHREAD_CONFIG_SRP_SERVER_ENABLE` here, since it is possible for user to use `Publisher` without actually using the OT SRP server implementation. For example, in the case where SRP server is provided by infrastructure.

src/core/thread/network_data_tlvs.hpp
```diff
@@ -890,10 +914,7 @@ class BorderRouterEntry : public Equatable<BorderRouterEntry>
      * @returns the Preference value.
      *
      */
-    int8_t GetPreference(void) const
-    {
-        return static_cast<int8_t>(static_cast<int16_t>(HostSwap16(mFlags)) >> kPreferenceOffset);
```
abtink, line 895: I copied this into a new function `PreferenceFromFlags()` in this PR.

But I just noticed that we are using right shift of a signed int. I think his is not explicitly specfiied by C++ standard and its behavior may be implementation/toolchain dependent. Better to avoid it if we can.

I will address it in a separate PR (into `main`) and then we can rebase and fix here as well.

abtink: This is what I found: https://en.cppreference.com/w/cpp/language/operator_arithmetic

> For negative a, the behavior of a << b is undefined.
> 
> For unsigned a and for signed and non-negative a, the value of a >> b is the integer part of a/2b
> 
> For negative a, the value of a >> b is implementation-defined (in most implementations, this performs arithmetic right shift, so that the result remains negative).
>

src/core/thread/network_data_publisher.hpp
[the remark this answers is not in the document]
librasungirl: Ok, I see. Make sense

jwhui (approved):

PR #6769, opened 2021-06-28
[cli] add cli-radio build

Irving-cl:
Background:
In an early stage of developing a new product, developers may want to verify
the Thread related hardware functionality, for example, send, receive, etc.
However if the product uses the posix <-> RCP mode, it's hard to verify it
before the posix daemon can run on the host. cli-ftd and cli-mtd could be an
alternative. However, if the product uses the posix <-> RCP mode, the space of
the 802.15.4 radio chip will be very small. For example, nRF52811 doesn't have
enough space to flash cli-ftd or cli-mtd firmware. So this PR creates a new
firmware `ot-cli-radio` which has very limited functionality for verifying the
hardware.

Basically, the `ot-cli-radio` consists of `openthread-radio` and
`libopenthread-cli-radio.a`. It supports very few cli commands. The most
important command is `diag`. Currently, there are: `diag`, `help`, `reset`,
`version`.

Currently, the diag commands have different implementation on RCP:
* For commands like `diag start`, the process function (`Diags::ProcessStart`)
is first called on host. Then host sends an SPI frame to the RCP and
`Diags::ProcessStart` (different implementation) is called again on RCP.
* For commands like `diag send`, `Diags::ProcessSend` is first called on host.
Then host sends an SPI frame to the RCP, directly calling send API. And there 
is no `Diags::ProcessSend` implemented on RCP.

Let's call the implementation of `Diags::Process*` currently on host as native
diag commands. When we run `ot-cli-radio`, we should use the native diag 
commands because it won't interact with a posix daemon and it processes the
diag commands through the whole process. So this PR adds a new option
`OPENTHREAD_CONFIG_DIAG_NATIVE_CMDS_ON_RCP` to control whether to use the
native diag implementation on RCP. When we use a normal RCP, the option should
be disabled. While when use `ot-cli-radio`, the option should be enabled.


examples/apps/cli/Makefile.am
```diff
@@ -119,6 +119,36 @@ ot_cli_mtd_SOURCES                                                     = \
     $(SOURCES_COMMON)                                                    \
     $(NULL)
 
+if OPENTHREAD_ENABLE_RADIO_ONLY
+bin_PROGRAMS                                                          += \
+    ot-cli-radio                                                         \
+    $(NULL)
```
abtink, line 125: I am not sure, but it kind of looks as if the `\` at the end here is not aligned with the rest of them below it?

src/core/config/diag.h
```diff
@@ -75,4 +75,17 @@
...
 
+/**
+ * @def OPENTHREAD_CONFIG_DIAG_NATIVE_CMDS_ON_RCP
+ *
+ * Define to 1 to support native diag commands on RCP.
+ *
+ * This option is used when RCP is used standalone (using `ot-cli-radio`, without a posix daemon). In this case, users
+ * directly control the rcp through CLI. When this option is enabled, users can directly use native diag commands
+ * (`send`, `transmit`, `receive`...) on RCP.
+ */
```
abtink, line 86:
```suggestion
 *
 */
```

abtink (approved):
Thanks @Irving-cl. Looks really great. 👍 

I agree and I think that the `cli-radio` would be very useful.

One question/suggestion: I am not sure if the new build (`cli-radio`) is already covered by any of github action workflows. If not, I think would be good to add it so we make sure future changes don't break it.

Irving-cl:
@abtink Thanks!

I added a build check for cli-radio.

etc/cmake/options.cmake
```diff
@@ -162,6 +162,11 @@ if(OT_DIAGNOSTIC)
     target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_DIAG_ENABLE=1")
 endif()
 
+option(OT_DIAG_NATIVE_RCP "enable native diag commands on rcp" OFF)
```
bukepo, line 165: since it's more than just diag.
```suggestion
option(OT_APP_RCP_CLI "enable RCP cli app" OFF)
```

bukepo (approved):

simonlingoogle (approved):
LGTM :+1: 

Nice app to have. 
Maybe one day we can build otbr-agent upon cli-radio.

script/check-simulation-build-cmake
```diff
@@ -42,6 +42,13 @@ build_all_features()
     reset_source
     "$(dirname "$0")"/cmake-build simulation -DOT_THREAD_VERSION=1.1
 
+    # Build Thread 1.1 cli-radio
+    # cli-radio should be built alone because in a full build, `libopenthread-radio` should be built with
+    # `OPENTHREAD_CONFIG_DIAG_NATIVE_CMDS_ON_RCP=0`. While cli-radio needs `libopenthread-radio` built with
+    # `OPENTHREAD_CONFIG_DIAG_NATIVE_CMDS_ON_RCP=1`
```
jwhui, line 48: Is there a way to remove this config? For example, can we always build the native diag code but rely on linking and/or initialization to actually enable it?

Irving-cl: Sounds good to me. I will try to do that.

Irving-cl: After some trying and re-thinking, I think:
* It may not be good to always build both diag code and choose which one to use at runtime
    * For a usual RCP, the native diag code is totally useless. On some SoCs, our RCP application FW just fits in the size limit. So we should avoid adding code to normal RCP as possible.
    * This makes the code logic more complex. For example, we may need to add a flag in the Diags module to tell if it should use the native diag code or RCP diag code. And this applys to both Diags modules (Host and RCP). That says, we need to ensure the flag is correct on both host and RCP. We can use configuration as a default value for the flag. But this goes back, we still add one more config. 
* I tried a linking solution: move different diag code to two new source files and build two different `openthread-radio` library (most things are common except for some diag and radio_callback code). This should work. One flaw is that those class members (only for native diag code) will be there anyway. (I cannot think of a way to avoid this, is there any trick to do that without adding any configuration?)
* I think the best way is to add a new build type for openthread core (For example, `OPENTHREAD_RADIO_ALONE`, which is a sibling of `OPENTHREAD_FTD`, `OPENTHREAD_MTD` and `OPENTHREAD_RADIO`). And we will also build a different lib `openthread-radio-alone` and let ot-cli-rcp link this lib. This is much simpler.

Thoughts? @jwhui

jwhui: Having another build type/lib makes sense. However, I'm open to others opinions on the preferred approach. Thoughts from others?

Irving-cl:
@jwhui 

I've updated the implementation. I added a new lib `openthread-radio-alone` which is slightly different from `openthread-radio`. I think we can add more features to `openthread-radio-alone` for building a firmware to debug the hardware. Thoughts?

etc/cmake/options.cmake
```diff
@@ -1,5 +1,5 @@
 #
-#  Copyright (c) 2019, The OpenThread Authors.
+#  Copyright (c) 2019, The OpenThread Authors.
```
jwhui, line 2: nit: unintentional change?

examples/apps/cli/radio.cmake
```diff
@@ -0,0 +1,47 @@
...
+    ${OT_PLATFORM_LIB}
+    openthread-radio-alone
+    ${OT_PLATFORM_LIB}
+    openthread-cli-radio
+    ${OT_MBEDTLS}
+    ot-config
+)
+
+install(TARGETS ot-cli-radio
+    DESTINATION bin)
```
jwhui, line 47: nit:
```suggestion
    DESTINATION bin
)
```

src/cli/radio.cmake
```diff
@@ -0,0 +1,54 @@
...
+)
+
+target_link_libraries(openthread-cli-radio
+    PUBLIC
+        openthread-radio
+    PRIVATE
+        ${OT_MBEDTLS}
+        ot-config
+)
+
```
jwhui, line 54: nit: extra line

etc/cmake/options.cmake
jwhui, line 2: nit: unintentional change?

jwhui:
> I've updated the implementation. I added a new lib `openthread-radio-alone` which is slightly different from `openthread-radio`. I think we can add more features to `openthread-radio-alone` for building a firmware to debug the hardware. Thoughts?

@Irving-cl , can we think of a better name than `openthread-radio-alone`? I don't think `openthread-radio-alone` provides much insight into how it differs from `openthread-radio`. How about something like `openthread-radio-cli`?

Irving-cl:
> > I've updated the implementation. I added a new lib `openthread-radio-alone` which is slightly different from `openthread-radio`. I think we can add more features to `openthread-radio-alone` for building a firmware to debug the hardware. Thoughts?
> 
> @Irving-cl , can we think of a better name than `openthread-radio-alone`? I don't think `openthread-radio-alone` provides much insight into how it differs from `openthread-radio`. How about something like `openthread-radio-cli`?

@jwhui
SGTM. I'll update it.

jwhui:
> SGTM. I'll update it.

@Irving-cl , thanks. Looking at this again, is `OPENTHREAD_RADIO_CLI` a struct superset of `OPENTHREAD_RADIO`? In other words, are there situations where code is included with `OPENTHREAD_RADIO_CLI` but not `OPENTHREAD_RADIO`? I'm wondering if we can reduce the number of `OPENTHREAD_RADIO || OPENTHREAD_RADIO_CLI` occurrences.

Irving-cl:
> @Irving-cl , thanks. Looking at this again, is OPENTHREAD_RADIO_CLI a struct superset of OPENTHREAD_RADIO? In other words, are there situations where code is included with OPENTHREAD_RADIO_CLI but not OPENTHREAD_RADIO? I'm wondering if we can reduce the number of OPENTHREAD_RADIO || OPENTHREAD_RADIO_CLI occurrences.

@jwhui `OPENTHREAD_RADIO_CLI` is almost a superset of `OPENTHREAD_RADIO`. However, `OPENTHREAD_RADIO` and `OPENTHREAD_RADIO_CLI` are different in some functions:
* In [factory_diags.cpp](https://github.com/openthread/openthread/pull/6769/files#diff-1778f6a8a1054137b0f4f1f2592c0b61db1a5fa830c29ed418abfef3bf859109R145), they have different implementations of factory diag functions
* In [radio_callbacks.cpp](https://github.com/openthread/openthread/pull/6769/files#diff-6375658cad81330dc9f7ed6651bb13cf2a799bbadedffcd4276684e1385291bfL64), they have different implementation of `HandleDiagsReceiveDone` and `HandleDiagsTransmitDone`.

Do you mean we still use `OPENTHREAD_RADIO` for the common cases and add another macro for the cases where they are different?

jwhui:
> Do you mean we still use `OPENTHREAD_RADIO` for the common cases and add another macro for the cases where they are different?

@Irving-cl , yes, exactly. What do you think?

Irving-cl:
> @Irving-cl , yes, exactly. What do you think?

Make sense to me. I will update it.

Irving-cl:
@jwhui 

I've updated it. Now I made `OPENTHREAD_RADIO_CLI` as a sub option when using `OPENTHREAD_RADIO`. `OPENTHREAD_RADIO_CLI` will only have effect when `OPENTHREAD_RADIO` is enabled.

examples/apps/cli/Makefile.am
```diff
@@ -120,9 +120,42 @@ ot_cli_mtd_SOURCES                                                     = \
     $(SOURCES_COMMON)                                                    \
     $(NULL)
 
+if OPENTHREAD_ENABLE_RADIO_ONLY
```
jwhui, line 123: Rename to `OPENTHREAD_ENABLE_RADIO_CLI`?

configure.ac
```diff
@@ -542,6 +542,30 @@ AC_ARG_ENABLE(radio-only,
 AC_MSG_RESULT(${enable_radio_only})
 AM_CONDITIONAL([OPENTHREAD_ENABLE_RADIO_ONLY], [test "${enable_radio_only}" = "yes"])
 
+#
+# Radio Cli Library
+#
+
+AC_MSG_CHECKING([whether to build radio-cli library])
+AC_ARG_ENABLE(radio-only,
```
jwhui, line 550:
```suggestion
AC_ARG_ENABLE(radio-cli,
```

jwhui (approved):

PR #6770, opened 2021-06-28
[cli] add support for TCP in the CLI tool

samkumar:
This is the second in a series of pull requests (after #6744) bringing TCPlp to OpenThread. I'm making a series of small pull requests so that it's easier for the OpenThread community to review the code. If you'd like to preview a working version of TCPlp in OpenThread, see #6650.

This pull request adds support for TCP in OpenThread's CLI tool. As TCP isn't implemented, the TCP functions don't work yet, but the code complies and links due to the stubs added in the previous pull request (#6744). It's separated into three commits: one that restores functionality in the `Interpreter` class removed in #6691, one that implements a basic CLI tool, and one that adds a benchmark utility to the CLI tool.

The next pull request in the sequence, which implements TCP functionality, doesn't depend on this one being merged in. So, I might not wait for this pull request to be merged before making the third pull request in the sequence.

### DIFF
--- a/src/cli/cli_tcp.cpp
+++ b/src/cli/cli_tcp.cpp
@@ -1,0 +1,520 @@
+/*
+ *  Copyright (c) 2021, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file implements a TCP CLI tool.
+ */
+
+#include "openthread-core-config.h"
+
+#include "cli_config.h"
+
+#if OPENTHREAD_CONFIG_TCP_ENABLE && OPENTHREAD_CONFIG_CLI_TCP_ENABLE
+
+#include "cli_tcp.hpp"
+
+#include <openthread/tcp.h>
+
+#include "cli/cli.hpp"
+#include "common/encoding.hpp"
+#include "common/timer.hpp"
+
+namespace ot {
+namespace Cli {
+
+constexpr TcpExample::Command TcpExample::sCommands[];
+
+TcpExample::TcpExample(Interpreter &aInterpreter)
+    : mInterpreter(aInterpreter)
+    , mInitialized(false)
+    , mEndpointConnected(false)
+    , mSendBusy(false)
+    , mBenchmarkBytesTotal(0)
+    , mBenchmarkLinksLeft(0)
+{
+}
+
+otError TcpExample::ProcessHelp(Arg aArgs[])
+{
+    OT_UNUSED_VARIABLE(aArgs);
+
+    for (const Command &command : sCommands)
+    {
+        mInterpreter.OutputLine(command.mName);
+    }
+
+    return OT_ERROR_NONE;
+}
+
+otError TcpExample::ProcessInit(Arg aArgs[])
+{
+    otError error = OT_ERROR_NONE;
+    size_t  receiveBufferSize;
+
+    VerifyOrExit(!mInitialized, error = OT_ERROR_ALREADY);
+
+    if (aArgs[0].IsEmpty())
+    {
+        receiveBufferSize = sizeof(mReceiveBuffer);
+    }
+    else
+    {
+        uint32_t windowSize;
+
+        SuccessOrExit(error = aArgs[0].ParseAsUint32(windowSize));
+        VerifyOrExit(aArgs[1].IsEmpty(), error = OT_ERROR_INVALID_ARGS);
+
+        receiveBufferSize = windowSize + ((windowSize + 7) >> 3);
+        VerifyOrExit(receiveBufferSize <= sizeof(mReceiveBuffer) && receiveBufferSize != 0,
+                     error = OT_ERROR_INVALID_ARGS);
+    }
+
+    {
+        otTcpEndpointInitializeArgs endpointArgs;
+
+        memset(&endpointArgs, 0x00, sizeof(endpointArgs));
+        endpointArgs.mEstablishedCallback      = HandleTcpEstablishedCallback;
+        endpointArgs.mSendDoneCallback         = HandleTcpSendDoneCallback;
+        endpointArgs.mReceiveAvailableCallback = HandleTcpReceiveAvailableCallback;
+        endpointArgs.mDisconnectedCallback     = HandleTcpDisconnectedCallback;
+        endpointArgs.mContext                  = this;
+        endpointArgs.mReceiveBuffer            = mReceiveBuffer;
+        endpointArgs.mReceiveBufferSize        = receiveBufferSize;
+
+        SuccessOrExit(error = otTcpEndpointInitialize(mInterpreter.mInstance, &mEndpoint, &endpointArgs));
+    }
+
+    {
+        otTcpListenerInitializeArgs listenerArgs;
+
+        memset(&listenerArgs, 0x00, sizeof(listenerArgs));
+        listenerArgs.mAcceptReadyCallback = HandleTcpAcceptReadyCallback;
+        listenerArgs.mAcceptDoneCallback  = HandleTcpAcceptDoneCallback;
+        listenerArgs.mContext             = this;
+
+        error = otTcpListenerInitialize(mInterpreter.mInstance, &mListener, &listenerArgs);
+        if (error != OT_ERROR_NONE)
+        {
+            IgnoreReturnValue(otTcpEndpointDeinitialize(&mEndpoint));
+            ExitNow();
+        }
+    }
+
+    mInitialized = true;
+
+exit:
+    return error;
+}
+
+otError TcpExample::ProcessDeinit(Arg aArgs[])
+{
+    otError error = OT_ERROR_NONE;
+    otError endpointError;
+    otError listenerError;
+
+    VerifyOrExit(aArgs[0].IsEmpty(), error = OT_ERROR_INVALID_ARGS);
+    VerifyOrExit(mInitialized, error = OT_ERROR_INVALID_STATE);
+
+    endpointError = otTcpEndpointDeinitialize(&mEndpoint);
+    mSendBusy     = false;
+
+    listenerError = otTcpListenerDeinitialize(&mListener);
+    mInitialized  = false;
+
+    SuccessOrExit(error = endpointError);
+    SuccessOrExit(error = listenerError);
+
+exit:
+    return error;
+}
+
+otError TcpExample::ProcessBind(Arg aArgs[])
+{
+    otError    error;
+    otSockAddr sockaddr;
+
+    VerifyOrExit(mInitialized, error = OT_ERROR_INVALID_STATE);
+
+    SuccessOrExit(error = aArgs[0].ParseAsIp6Address(sockaddr.mAddress));
+    SuccessOrExit(error = aArgs[1].ParseAsUint16(sockaddr.mPort));
+    VerifyOrExit(aArgs[2].IsEmpty(), error = OT_ERROR_INVALID_ARGS);
+
+    error = otTcpBind(&mEndpoint, &sockaddr);
+
+exit:
+    return error;
+}
+
+otError TcpExample::ProcessConnect(Arg aArgs[])
+{
+    otError    error;
+    otSockAddr sockaddr;
+
+    VerifyOrExit(mInitialized, error = OT_ERROR_INVALID_STATE);
+
+    SuccessOrExit(error = aArgs[0].ParseAsIp6Address(sockaddr.mAddress));
+    SuccessOrExit(error = aArgs[1].ParseAsUint16(sockaddr.mPort));
+    VerifyOrExit(aArgs[2].IsEmpty(), error = OT_ERROR_INVALID_ARGS);
+
+    SuccessOrExit(error = otTcpConnect(&mEndpoint, &sockaddr, OT_TCP_CONNECT_NO_FAST_OPEN));
+    mEndpointConnected = false;
+
+exit:
+    return error;
+}
+
+otError TcpExample::ProcessSend(Arg aArgs[])
+{
+    otError error;
+
+    VerifyOrExit(mInitialized, error = OT_ERROR_INVALID_STATE);
+    VerifyOrExit(!mSendBusy, error = OT_ERROR_BUSY);
+    VerifyOrExit(mBenchmarkBytesTotal == 0, error = OT_ERROR_BUSY);
+
+    mSendLink.mNext = nullptr;
+    mSendLink.mData = mSendBuffer;
+    VerifyOrExit(!aArgs[0].IsEmpty(), error = OT_ERROR_INVALID_ARGS);
+    mSendLink.mLength = OT_MIN(aArgs[0].GetLength(), sizeof(mSendBuffer));
+    memcpy(mSendBuffer, aArgs[0].GetCString(), mSendLink.mLength);
+    VerifyOrExit(aArgs[1].IsEmpty(), error = OT_ERROR_INVALID_ARGS);
+
+    SuccessOrExit(error = otTcpSendByReference(&mEndpoint, &mSendLink, 0));
+    mSendBusy = true;
+
+exit:
+    return error;
+}
+
+otError TcpExample::ProcessBenchmark(Arg aArgs[])
+{
+    otError  error;
+    uint32_t toSendOut;
+
+    VerifyOrExit(!mSendBusy, error = OT_ERROR_BUSY);
+    VerifyOrExit(mBenchmarkBytesTotal == 0, error = OT_ERROR_BUSY);
+
+    if (aArgs[0].IsEmpty())
+    {
+        mBenchmarkBytesTotal = OPENTHREAD_CONFIG_CLI_TCP_DEFAULT_BENCHMARK_SIZE;
+    }
+    else
+    {
+        SuccessOrExit(error = aArgs[0].ParseAsUint32(mBenchmarkBytesTotal));
+        VerifyOrExit(mBenchmarkBytesTotal != 0, error = OT_ERROR_INVALID_ARGS);
+    }
+    VerifyOrExit(aArgs[1].IsEmpty(), error = OT_ERROR_INVALID_ARGS);
+
+    memset(mSendBuffer, 'a', sizeof(mSendBuffer));
+
+    mBenchmarkLinksLeft = (mBenchmarkBytesTotal + sizeof(mSendBuffer) - 1) / sizeof(mSendBuffer);
+    toSendOut           = OT_MIN(OT_ARRAY_LENGTH(mBenchmarkLinks), mBenchmarkLinksLeft);
+    mBenchmarkStart     = TimerMilli::GetNow();
+    for (uint32_t i = 0; i != toSendOut; i++)
+    {
+        mBenchmarkLinks[i].mNext   = nullptr;
+        mBenchmarkLinks[i].mData   = mSendBuffer;
+        mBenchmarkLinks[i].mLength = sizeof(mSendBuffer);
+        if (i == 0 && mBenchmarkBytesTotal % sizeof(mSendBuffer) != 0)
+        {
+            mBenchmarkLinks[i].mLength = mBenchmarkBytesTotal % sizeof(mSendBuffer);
+        }
+        SuccessOrExit(error = otTcpSendByReference(&mEndpoint, &mBenchmarkLinks[i],
+                                                   i == toSendOut - 1 ? 0 : OT_TCP_SEND_MORE_TO_COME));
+    }
+
+exit:
+    if (error != OT_ERROR_NONE)
+    {
+        mBenchmarkBytesTotal = 0;
+        mBenchmarkLinksLeft  = 0;
+    }
+    return error;
+}
+
+otError TcpExample::ProcessSendEnd(Arg aArgs[])
+{
+    otError error;
+
+    VerifyOrExit(mInitialized, error = OT_ERROR_INVALID_STATE);
+    VerifyOrExit(aArgs[0].IsEmpty(), error = OT_ERROR_INVALID_ARGS);
+
+    error = otTcpSendEndOfStream(&mEndpoint);
+
+exit:
+    return error;
+}
+
+otError TcpExample::ProcessAbort(Arg aArgs[])
+{
+    otError error;
+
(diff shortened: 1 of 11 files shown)


samkumar:
I noticed that some of the CI builds are failing because we exceeded an SRAM limit. I'm wondering if there's a target memory footprint that OpenThread is targeting? If needed, I can shrink `mReceiveBuffer` in the CLI tool from 4 KiB to 3 KiB to fit within the limit. That won't affect performance over a single hop, but it potentially could over a large number (more than about 5) wireless hops.

Before I went ahead and implemented this, I wanted to get some guidance on the best way to solve those issues.

samkumar:
I noticed that one of the CI builds is failing because the code exceeds an SRAM limit. If needed, I can shrink `mReceiveBuffer` in the CLI tool from 4 KiB to 3 KiB to fit within the limit. That won't affect performance over a single hop, but it potentially could over a large number (more than about 5) wireless hops.

Before I went ahead and implemented this, I wanted to get some guidance on the best way to solve this issue.

jwhui:
> I noticed that one of the CI builds is failing because the code exceeds an SRAM limit. If needed, I can shrink `mReceiveBuffer` in the CLI tool from 4 KiB to 3 KiB to fit within the limit. That won't affect performance over a single hop, but it potentially could over a large number (more than about 5) wireless hops.
> 
> Before I went ahead and implemented this, I wanted to get some guidance on the best way to solve this issue.

@samkumar , another option might be to have platform-specific recommended configurations. For example:

1. Platforms that are resource rich, recommend the configuration that maximizes performance.
2. Platforms that are resource constrained, configure the minimum that would allow TCP to function (possibly with reduced performance in various settings).

samkumar:
Thanks @jwhui for the suggestion. I want to clarify that this only applies to the CLI tool at the moment, but applications in general need to allocate memory for the receive buffer. Are you suggesting to change the receive buffer size in the CLI tool depending on the target platform, or that we should additionally make recommendations to application developers using TCP?

jwhui:
> Thanks @jwhui for the suggestion. I want to clarify that this only applies to the CLI tool at the moment, but applications in general need to allocate memory for the receive buffer. Are you suggesting to change the receive buffer size in the CLI tool depending on the target platform, or that we should additionally make recommendations to application developers using TCP?

@samkumar , ideally both. We should have recommended configurations for a given platform. At the same time, some documentation that helps a developer understand the various tradeoffs is also useful.

samkumar:
Thanks @jwhui. Are there any existing examples of platform-specific constants and their usage that I can follow? I'm not sure how to go about defining such a constant.

samkumar:
Thanks @jwhui. I went ahead and added suggested receive buffer sizes in `tcp.h`. I've also added a default value in `cli_config.h` that can be overridden for specific platforms.

Can you please add reviewers like you did last time? I don't have permission to add reviewers.

abtink:
> It's separated into three commits: one that restores functionality in the Interpreter class removed in #6691,

@samkumar I see you are adding `Interpretter::Output()`. 
- The PR #6691 did not actually remove any functionality. It just removed method declarations in `Interpreter` that had no implementation or definition (without any implementation the declaration were not usable anyway).
- I went through the code history. It seems like the `Output()` method was declared without providing an implementation in PR https://github.com/openthread/openthread/pull/5387 (probably as a mistake).
   - Before https://github.com/openthread/openthread/pull/5387 we had multiple (sub-classes of) `Cli::Server` which provided an `Output()` method as the low-level (driver) function emitting a seq of chars to the console (which was then used by the `Cli::Interpreter`).
   - The equivalent of this in `main` today is `otCliOutputCallback` which is callback registered with `Cli::Interpretter` when it is initialized and used to output to console.


-----------

I see you added `Output` in this PR as:
```c++
int Interpreter::Output(const char *aBuf, uint16_t aBufLength)
 {
     int written = 0;
     for (uint16_t i = 0; i != aBufLength; i++)
     {
         written += OutputFormat("%c", aBuf[i]);
     }
     return written;
 }
```

We can use the following printf formatter `"%.*s"` to print a given number of chars from an string instead.
```c++
   OutputFormat("%.*s", aBufLength, aBuf);
```

------

I see that you are using this to print the received payload from `HandleTcpReceiveAvailable()`
```
mInterpreter.Output(reinterpret_cast<const char *>(data->mData), data->mLength);
```

In general printing as a sequence of char may not be good, since the received bytes may not be printable chars (will mess up the console). We can use hex bytes `OutputBytes()` for this instead.

Or a better suggestion is to use hex dump style format.  I know that  in `Interpretter::HandleLinkPcapReceive()` we already have an implementation of hex dump style. This can be refactored as an `OutputHexDump()` method and then used by `Tcp` as well.

samkumar:
@abtink, thank you for looking at the pull request and for your feedback.

I was envisioning that the CLI tool would be used for connecting to another device running the OpenThread CLI. Thus, any data received over TCP would have originated from another instance of the OpenThread CLI tool and therefore is guaranteed to consist of printable characters. That's why I felt it was appropriate to print out the characters directly.

The UDP CLI tool also appears to print the data directly: see https://github.com/openthread/openthread/blob/main/src/cli/cli_udp.cpp#L284-L294.

I am against printing out hex-encoded data. If the user sends `hello` over TCP, they expect to receive `hello` on the other node, not `68656c6c6f`.

I think the hex-dump idea is interesting. I can understand that it may be useful if someone uses the OpenThread CLI to connect to a TCP service that is not an instance of the OpenThread CLI. In light of what I said above, do you still think it's necessary? If so, I'm happy to implement it.

src/cli/cli_tcp.hpp
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * This mehtod interprets a list of CLI arguments.
+     *
+     * @param[in]  aArgs   An array of command line arguments.
+     *
+     */
+    otError Process(Arg aArgs[]);
+
+private:
+    class Callbacks;
```
abtink, line 77: Do we need this forward declaration of `Callbacks`? I cannot find if it being used anywhere.

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,522 @@
...
+    : mInterpreter(aInterpreter)
+    , mEndpointInitialized(false)
+    , mListenerInitialized(false)
+    , mEndpointConnected(false)
+    , mSendBusy(false)
+    , mBenchmarkBytesTotal(0)
+    , mBenchmarkLinksLeft(0)
+{
+    memset(&mEndpoint, 0x00, sizeof(mEndpoint));
+    memset(&mListener, 0x00, sizeof(mListener));
```
abtink, line 61: Do we nee to `memset()` these? Should the `otTcpEndpointInitialize()` already take care of it (if needed).

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,522 @@
...
+        if (receiveBufferSize > sizeof(mReceiveBuffer) || receiveBufferSize == 0)
+        {
+            ExitNow(error = OT_ERROR_INVALID_ARGS);
+        }
+    }
+
+    if (!mEndpointInitialized)
+    {
+        otTcpEndpointInitializeArgs endpointArgs;
+        memset(&endpointArgs, 0x00, sizeof(endpointArgs));
```
abtink, line 103:
```suggestion
        otTcpEndpointInitializeArgs endpointArgs;
        
        memset(&endpointArgs, 0, sizeof(endpointArgs));
```

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,522 @@
...
+        endpointArgs.mReceiveBufferSize        = receiveBufferSize;
+
+        SuccessOrExit(error = otTcpEndpointInitialize(mInterpreter.mInstance, &mEndpoint, &endpointArgs));
+        mEndpointInitialized = true;
+    }
+
+    if (!mListenerInitialized)
+    {
+        otTcpListenerInitializeArgs listenerArgs;
+        memset(&listenerArgs, 0x00, sizeof(listenerArgs));
```
abtink, line 119:
```suggestion
        otTcpListenerInitializeArgs listenerArgs;
        
        memset(&listenerArgs, 0, sizeof(listenerArgs));
```

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,522 @@
...
+
+    return OT_ERROR_NONE;
+}
+
+otError TcpExample::ProcessInit(Arg aArgs[])
+{
+    otError error = OT_ERROR_NONE;
+    size_t  receiveBufferSize;
+
+    VerifyOrExit(aArgs[0].IsEmpty() || aArgs[1].IsEmpty(), error = OT_ERROR_INVALID_ARGS);
```
abtink, line 81: We can remove this line. Will be covered below.

Can possibly add `VerifyOrExit(aArgs[1].IsEmpty(), error = OT_ERROR_INVALID_ARGS)` after the `SuccessOrExit(error = aArgs[0].ParseAsUint32(windowSize));` (to make sure we have only one agr)

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,522 @@
...
+    VerifyOrExit(!mEndpointInitialized || !mListenerInitialized, error = OT_ERROR_INVALID_STATE);
+
+    if (aArgs[0].IsEmpty())
+    {
+        receiveBufferSize = sizeof(mReceiveBuffer);
+    }
+    else
+    {
+        uint32_t windowSize;
+        SuccessOrExit(error = aArgs[0].ParseAsUint32(windowSize));
```
abtink, line 91:
```suggestion
        uint32_t windowSize;
        
        SuccessOrExit(error = aArgs[0].ParseAsUint32(windowSize));
```

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,522 @@
...
+
+        receiveBufferSize = windowSize + ((windowSize + 7) >> 3);
+        if (receiveBufferSize > sizeof(mReceiveBuffer) || receiveBufferSize == 0)
+        {
+            ExitNow(error = OT_ERROR_INVALID_ARGS);
+        }
```
abtink, line 97:
```suggestion
        VerifyOrExit(receiveBufferSize <= sizeof(mReceiveBuffer) && receiveBufferSize != 0, error = OT_ERROR_INVALID_ARGS);
```

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,522 @@
...
+            {
+                mInterpreter.OutputLine("TCP Benchmark Failed");
+                mBenchmarkBytesTotal = 0;
+            }
+        }
+        else if (mBenchmarkLinksLeft == 0)
+        {
+            uint32_t milliseconds         = TimerMilli::GetNow() - mBenchmarkStart;
+            uint32_t thousandTimesGoodput = (1000 * (mBenchmarkBytesTotal << 3) + (milliseconds >> 1)) / milliseconds;
+            mInterpreter.OutputLine("TCP Benchmark Complete: Transferred %u bytes in %u milliseconds",
```
abtink, line 409:
```suggestion
            uint32_t thousandTimesGoodput = (1000 * (mBenchmarkBytesTotal << 3) + (milliseconds >> 1)) / milliseconds;
            
            mInterpreter.OutputLine("TCP Benchmark Complete: Transferred %u bytes in %u milliseconds",
```

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,522 @@
...
+                                           size_t         aBytesRemaining)
+{
+    OT_UNUSED_VARIABLE(aBytesRemaining);
+    OT_ASSERT(aEndpoint == &mEndpoint);
+
+    if (aBytesAvailable > 0)
+    {
+        const otLinkedBuffer *data;
+        IgnoreError(otTcpReceiveByReference(aEndpoint, &data));
+        size_t totalReceived = 0;
```
abtink, line 431: All variable should be toghther at the top of scope.

```suggestion
        const otLinkedBuffer *data;
        size_t totalReceived = 0;

        IgnoreError(otTcpReceiveByReference(aEndpoint, &data));
```

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,522 @@
...
+    case OT_TCP_DISCONNECTED_REASON_RESET:
+        mInterpreter.OutputLine("TCP: Connection reset");
+        break;
+    }
+
+    /*
+     * We set this to false even for the TIME-WAIT state, so that we can reuse
+     * the active socket if an incoming connection comes in instead of waiting
+     * for the 2MSL timeout.
+     */
```
abtink, line 476: Use `// ` comment style to avoid extra lines for `/*` and `*/`?

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,522 @@
...
+{
+    otError  error;
+    uint32_t toSendOut;
+
+    VerifyOrExit(!mSendBusy, error = OT_ERROR_INVALID_STATE);
+    VerifyOrExit(mBenchmarkBytesTotal == 0, error = OT_ERROR_INVALID_STATE);
+
+    if (aArgs[0].IsEmpty())
+    {
+        mBenchmarkBytesTotal = 72 << 10;
```
abtink, line 223: Just curious why this number? Should we defined as a constant?

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,522 @@
...
+            mSendBusy = false;
+        }
+    }
+    else
+    {
+        OT_ASSERT(aData != &mSendLink);
+        mBenchmarkLinksLeft--;
+        if (mBenchmarkLinksLeft >= OT_ARRAY_LENGTH(mBenchmarkLinks))
+        {
+            if (otTcpSendByReference(&mEndpoint, aData, 0) != OT_ERROR_NONE)
```
abtink, line 399: What if the `SendDone` is for the first entry, i.e. `mBenchmarkLinks[0]`which can have a length smaller than other ones?

Should we change the `aData.mLength`?

abtink (approved):
Thanks @samkumar.
I added some questions and some style suggestions below, but overall looks great. 👍

abtink:
> I think the hex-dump idea is interesting. I can understand that it may be useful if someone uses the OpenThread CLI to connect to a TCP service that is not an instance of the OpenThread CLI. In light of what I said above, do you still think it's necessary? If so, I'm happy to implement it.

I think printing char string is fine since we expect the other end is also running OT CLI (and/or always sending printable chars to us). I would suggest using `OutputFormat("%.*s", aBufLength, aBuf)` for it though (to avoid adding a new method when we can do the same with an existing one).

We can consider adding hex dump model later on if there is a use-case for it.

src/cli/cli_tcp.hpp
samkumar: Thanks for noticing this. I'll remove it.

src/cli/cli_tcp.cpp
samkumar: The `memset()` call isn't necessary. I put it in because I saw `cli_udp.cpp` do it.  I'm happy to remove it.

src/cli/cli_tcp.cpp
samkumar: I chose it to be large enough to get a consistent measurement, but not so large that the benchmark would take more than a few seconds. I will define this as a constant.

src/cli/cli_tcp.cpp
samkumar: Good point, I will update. I suppose this bug didn't show up in my testing because, `mBenchmarkBytesTotal` was a multiple of `sizeof(mSendBuffer)`, so I will do a bit more testing there as well.

samkumar:
Thanks @abtink for your feedback. I just updated the pull request to fix all of the issues that you pointed out.

samkumar:
I recognize that it may be tricky to review this code, since you can't test it until the TCP functionality is also included. For people who would like to try out this code, I've rebased Pull Request #6650 on top of this so that you can try out the CLI with the TCP implementation included. I've also posted a video (https://youtu.be/ppZ784YUKlI) that might be helpful when getting started with the TCP functionality in the CLI.

simonlingoogle (approved):
LGTM :+1: 

with nits

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,517 @@
...
+    otError    error;
+    otSockAddr sockaddr;
+
+    VerifyOrExit(mEndpointInitialized, error = OT_ERROR_INVALID_STATE);
+
+    SuccessOrExit(error = aArgs[0].ParseAsIp6Address(sockaddr.mAddress));
+    SuccessOrExit(error = aArgs[1].ParseAsUint16(sockaddr.mPort));
+    VerifyOrExit(aArgs[2].IsEmpty(), error = OT_ERROR_INVALID_ARGS);
+
+    error = otTcpBind(&mEndpoint, &sockaddr);
```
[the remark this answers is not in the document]
samkumar: We don't have a separate API call for binding the listenener; the `otTcpListen` call does that for us. Thus, that functionality is included within the `tcp listen` command in the CLI tool.

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,517 @@
...
+        receiveBufferSize = sizeof(mReceiveBuffer);
+    }
+    else
+    {
+        uint32_t windowSize;
+
+        SuccessOrExit(error = aArgs[0].ParseAsUint32(windowSize));
+        VerifyOrExit(aArgs[1].IsEmpty(), error = OT_ERROR_INVALID_ARGS);
+
+        receiveBufferSize = windowSize + ((windowSize + 7) >> 3);
```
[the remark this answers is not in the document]
samkumar: This is checking for buffer overflow. The argument here is meant to be the size of the receive buffer. The amount of memory passed in to `otTcpEndpointInitialize` has to have space for the requested receive buffer and for the bitmap to check for out-of-order segments, hence this formula.

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,517 @@
...
+        endpointArgs.mDisconnectedCallback     = HandleTcpDisconnectedCallback;
+        endpointArgs.mContext                  = this;
+        endpointArgs.mReceiveBuffer            = mReceiveBuffer;
+        endpointArgs.mReceiveBufferSize        = receiveBufferSize;
+
+        SuccessOrExit(error = otTcpEndpointInitialize(mInterpreter.mInstance, &mEndpoint, &endpointArgs));
+        mEndpointInitialized = true;
+    }
+
+    if (!mListenerInitialized)
```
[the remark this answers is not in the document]
samkumar: Back when I was writing this, I think that I was thinking of the case where initialization succeeds for one but not the other. But in retrospect, that is better handled by deinitializing one if initializing the other fails. I'll go ahead and implement that, and also switch to using one bool instead of two.

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,517 @@
...
+            mBenchmarkLinks[i].mLength = mBenchmarkBytesTotal % sizeof(mSendBuffer);
+        }
+        SuccessOrExit(error = otTcpSendByReference(&mEndpoint, &mBenchmarkLinks[i],
+                                                   i == toSendOut - 1 ? 0 : OT_TCP_SEND_MORE_TO_COME));
+    }
+
+exit:
+    if (error != OT_ERROR_NONE)
+    {
+        mBenchmarkBytesTotal = 0;
```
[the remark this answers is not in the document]
samkumar: I added that line in as well. (I don't think it's absolutely needed, but it's safer and more readable to have it in I think.)

src/cli/cli_tcp.hpp
```diff
@@ -0,0 +1,160 @@
...
+    bool mEndpointInitialized;
+    bool mListenerInitialized;
+    bool mEndpointConnected;
+    bool mSendBusy;
+
+    otLinkedBuffer mSendLink;
+    uint8_t        mSendBuffer[OPENTHREAD_CONFIG_CLI_MAX_LINE_LENGTH];
+    uint8_t        mReceiveBuffer[OPENTHREAD_CONFIG_CLI_TCP_RECEIVE_BUFFER_SIZE];
+
+    otLinkedBuffer mBenchmarkLinks[(sizeof(mReceiveBuffer) + sizeof(mSendBuffer) - 1) / sizeof(mSendBuffer)];
```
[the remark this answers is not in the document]
samkumar: I'm happy to add a separate configuration flag for the TCP CLI. But, I guess I'm not 100% clear on why it's needed? I had assumed that the CLI tool wouldn't actually be used in production, so it's memory footprint doesn't matter for real applications. Am I mistaken?

src/cli/cli_tcp.cpp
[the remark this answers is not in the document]
simonlingoogle: Thanks for clarification.

src/cli/cli_tcp.cpp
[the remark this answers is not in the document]
simonlingoogle: Deinitializing sounds good to me.

But I am wonder what might possibility go wrong in `otTcpListenerInitialize` or `otTcpEndpointInitialize`. They should always succeed?

samkumar:
@simonlingoogle I just pushed an update that addresses almost all of your suggestions. I still have to add the config flag for TCP CLI, but I saw you were making another pass now so I pushed so you can see the latest code.

src/cli/cli_tcp.hpp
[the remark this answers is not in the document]
simonlingoogle: I think it's logical for a user program on a SoC device to use TCP and CLI, but don't need TCP CLI ?

src/cli/cli_tcp.cpp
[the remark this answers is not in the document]
samkumar: They *should* always succeed, But that's based on looking at the implementation. The API still has a return code that the CLI tool should handle as well (since someone may later change the implementation so that there's a chance of failure...)

src/cli/cli_tcp.hpp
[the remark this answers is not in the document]
samkumar: I had thought of the CLI as being a standalone application. I didn't realize it could be part of another user program. Can you please elaborate?

I'll add the flag, but I'd like to understand this use case more. :)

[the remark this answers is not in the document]
samkumar: Shall I add the flag to cli_config.h?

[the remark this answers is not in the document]
simonlingoogle: For example, otbr-agent is a Border Router program, but has all CLI commands integrated. 

I think CLI commands are so convenient that users may just add CLI library to whatever application they are building.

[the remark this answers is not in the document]
simonlingoogle: `cli_config.h` works for me.

samkumar:
@simonlingoogle I just pushed a version that addresses all of the points you brought up. Can you confirm that it's OK?

src/cli/cli_tcp.cpp
```diff
@@ -0,0 +1,520 @@
...
+    VerifyOrExit(mInitialized, error = OT_ERROR_INVALID_STATE);
+
+    endpointError = otTcpEndpointDeinitialize(&mEndpoint);
+    mSendBusy     = false;
+
+    listenerError = otTcpListenerDeinitialize(&mListener);
+    mInitialized  = false;
+
+    SuccessOrExit(error = endpointError);
+    SuccessOrExit(error = listenerError);
```
jwhui, line 151: What are the nature of these errors? I'm curious whether we should update `mSendBusy` and/or `mInitialized` when something other than `OT_ERROR_NONE` is returned. Thoughts?

src/cli/cli_tcp.hpp
```diff
@@ -0,0 +1,159 @@
...
+                                                            size_t         aBytesAvailable,
+                                                            bool           aEndOfStream,
+                                                            size_t         aBytesRemaining);
+    void                          HandleTcpDisconnected(otTcpEndpoint *aEndpoint, otTcpDisconnectedReason aReason);
+    otTcpIncomingConnectionAction HandleTcpAcceptReady(otTcpListener *   aListener,
+                                                       const otSockAddr *aPeer,
+                                                       otTcpEndpoint **  aAcceptInto);
+    void HandleTcpAcceptDone(otTcpListener *aListener, otTcpEndpoint *aEndpoint, const otSockAddr *aPeer);
+
+    static constexpr Command sCommands[] = {
```
jwhui, line 121: Can we add a `README_TCP.md` that documents these commands and perhaps a quick start section? See [src/cli/README_UDP.md](https://github.com/openthread/openthread/blob/main/src/cli/README_UDP.md) for reference.

jwhui (commented):
Looks great overall. Couple minor comments.

src/cli/cli_tcp.cpp
samkumar: In the actual implementation, deinitialization will never fail unless the endpoint/listener wasn't initialized (which would be undefined behavior anyway, since you called a function on uninitialized memory). I just thought it was good practice to check for the error in the CLI tool. Because there isn't a concrete case where this would fail unexpectedly, I wasn't sure how best to handle an error here.

src/cli/cli_tcp.hpp
samkumar: Sounds good, I will work on this. I'll try to write this up tomorrow. Do you want to include it in this pull request or should I make it a separate pull request?

src/cli/cli_tcp.cpp
jwhui: Do we envision a use case for the errors? If not, should the API return `void` instead?

src/cli/cli_tcp.hpp
jwhui: I don't have a preference whether to include in this PR or a separate PR. Thanks.

src/cli/cli_tcp.cpp
samkumar: The reason I left the return value as `otError` was to future-proof the API: perhaps, the implementation will change in some unforeseen way that it can return a meaningful error. Additionally, even in the current implementation, it can return an error in some cases, but only due to programmer error in how the function is used (e.g., if you remove an endpoint that isn't part of the global linked list of endpoints).

I'm happy to change the return value to `void` if you think that that's better.

src/cli/cli_tcp.hpp
samkumar: OK, I'll work on it. If, by the time I finish, this pull request is merged, I'll create a new pull request. Otherwise, I'll add it to this one.

src/cli/cli_tcp.cpp
jwhui: I'm fine with returning error here. Thanks.

jwhui (approved):

PR #6773, opened 2021-06-29
[dnssd] add support for service subtypes in DNS-SD server and client

abtink:
This commit updates `Dns::ServiceDiscovery::Server` to add support for
browsing a service subtype. In particular, it updates the code which
breaks a DNS name into its components such that it can correctly
parse and accept a service subtype name. It also updates how the
service name compression is done to handle the subtype names.

This commit also updates `Dns::Client` to remove the extra check of
service name (which is already verified) when trying to find a PTR
record for a given service instance name. This change enables browse
for service subtype (parsing the browse response).

This commit also updates and enhances the `test_dnssd.py` script
adding test-cases for browsing for service subtype.

----

This related to  issue 
- #6714

This PR includes the commits from 
-  #6760
-  #6724

**Please check and review the last commit only. Thanks.**

### DIFF
--- a/src/core/net/dnssd_server.cpp
+++ b/src/core/net/dnssd_server.cpp
@@ -49,6 +49,7 @@ namespace ServiceDiscovery {
 
 const char Server::kDnssdProtocolUdp[4] = {'_', 'u', 'd', 'p'};
 const char Server::kDnssdProtocolTcp[4] = {'_', 't', 'c', 'p'};
+const char Server::kDnssdSubTypeLabel[] = "._sub.";
 const char Server::kDefaultDomainName[] = "default.service.arpa.";
 
 Server::Server(Instance &aInstance)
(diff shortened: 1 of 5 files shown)


src/core/net/dnssd_server.cpp
```diff
@@ -547,10 +566,20 @@ Error Server::FindNameComponents(const char *aName, const char *aDomain, NameCom
...
+    // Check for service subtype
     error = FindPreviousLabel(aName, labelBegin, labelEnd);
     VerifyOrExit(error == kErrorNone, error = (error == kErrorNotFound ? kErrorNone : error));
 
+    if ((labelEnd == labelBegin + kSubTypeLabelLength) &&
+        (memcmp(&aName[labelBegin], kDnssdSubTypeLabel, kSubTypeLabelLength) == 0))
+    {
+        SuccessOrExit(error = FindPreviousLabel(aName, labelBegin, labelEnd));
+        VerifyOrExit(labelBegin == 0, error = kErrorInvalidArgs);
+        aInfo.mSubTypeOffset = labelBegin;
```
wgtdkp, line 578: Is this new field used anywhere?

wgtdkp (approved):
LGTM 👍

src/core/net/dnssd_server.cpp
abtink: You are correct. It is not used currently, but since we already detect and parse the subtype label I added it (it may be useful if we later move the code that breaks the name into components and define it a common DNS name utility helper).

src/core/net/dnssd_server.cpp
```diff
@@ -49,6 +49,7 @@ namespace ServiceDiscovery {
 
 const char Server::kDnssdProtocolUdp[4] = {'_', 'u', 'd', 'p'};
 const char Server::kDnssdProtocolTcp[4] = {'_', 't', 'c', 'p'};
+const char Server::kDnssdSubTypeLabel[] = "_sub.";
```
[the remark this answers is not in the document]
abtink: Good idea. Will change in next push.

jwhui (approved):

PR #6775, opened 2021-06-30
[thci] add support for SSED DUT

edmont:
Enable joining a network as SSED, which is required for several Thread 1.2 certification tests.

### DIFF
--- a/tools/harness-thci/OpenThread.py
+++ b/tools/harness-thci/OpenThread.py
@@ -1204,6 +1204,10 @@ class OpenThreadTHCI(object):
                 print('join as sleepy end device')
                 mode = '-'
                 self.__setPollPeriod(self.__sedPollPeriod)
+            elif eRoleId == Thread_Device_Role.SSED:
+                print('join as SSED')
+                mode = '-'
+                self.setCSLperiod(self.cslPeriod)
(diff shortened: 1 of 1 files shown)


tools/harness-thci/OpenThread.py
```diff
@@ -1204,6 +1204,10 @@ def joinNetwork(self, eRoleId):
                 print('join as sleepy end device')
                 mode = '-'
                 self.__setPollPeriod(self.__sedPollPeriod)
+            elif eRoleId == Thread_Device_Role.SSED:
+                print('join as SSED')
+                mode = '-'
```
simonlingoogle, line 1209: Question: Why didn't we use `n` mode for SED/SSED?
@librasungirl

simonlingoogle (approved):
LGTM :+1: 

Thanks

tools/harness-thci/OpenThread.py
edmont: I guess requiring stable network data only is a requirement for reference devices.

Irving-cl (approved):
LGTM 👍

jwhui (approved):
Thanks! 👍🏼

PR #6777, opened 2021-06-30
[otci] fix OTCI.wait arg order

gabekassel:
In wait, [match_line is called with the arguments flipped](https://github.com/openthread/openthread/blob/main/tools/otci/otci/otci.py#L78). The first argument is supposed to the be the line being checked, and the second is supposed to be the pattern to match. The reason this still works for strings is because output is treated as a list of lines to expect, and each is just compared with == to the actual expected line

### DIFF
--- a/tools/otci/otci/otci.py
+++ b/tools/otci/otci/otci.py
@@ -75,7 +75,7 @@ class OTCI(object):
 
             while duration > 0:
                 output = self.__otcmd.wait(1)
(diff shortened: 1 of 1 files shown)


simonlingoogle (approved):
LGTM :+1: 
Thanks for catching the bug.

jwhui (approved):
Thanks! 👍🏼

PR #6778, opened 2021-07-01
[mac] remove the out-of-band frame tx API

abtink:
This commit removes the API to enqueue an out of band frame for
transmission by MAC layer. This API was added as a work-around to
enable legacy (non-Thread based) frames to be sent along with Thread
frames. This model is no longer used and generally not recommended.
The preferred way to handle multi-radio/multi-protocol is to add
support for it in radio platform layer.


bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #6779, opened 2021-07-01
[cmd-parser] add `nullptr` check to `ParseHexString()`

jwhui:


abtink (approved):
LGTM. Thank you. 👍

src/core/utils/parse_cmdline.cpp
```diff
@@ -302,12 +302,17 @@ enum HexStringParseMode
...
-    size_t stringLength = strlen(aString);
-    size_t expectedSize = (stringLength + 1) / 2;
-    size_t parsedSize   = 0;
+    Error  error      = kErrorNone;
+    size_t parsedSize = 0;
+    size_t stringLength;
+    size_t expectedSize;
     bool   skipFirstDigit;
 
+    VerifyOrExit(aString != nullptr, error = kErrorInvalidArgs);
```
simonlingoogle, line 311: When can this be nullptr?

jwhui: With #6767, the model changed to now rely on `ParseAs()` to return error when no argument exists. One example:
https://github.com/openthread/openthread/blob/8967d62ead1679c5cdd0e94e194826fee8f8f7b8/src/cli/cli.cpp#L2861

PR #6781, opened 2021-07-01
[script] fix Thread version for check-simulation-build

Irving-cl:
The default Thread version has been set to 1.2 when building. Our test
script which checks building hasn't been updated accordingly. Currently it
specifies the version when it 1.2 is wanted and doesn't specify it when 1.1
is wanted. However, if THREAD_VERSIOn is not specified, the version would be
1.2. This PR fixes this error.

### DIFF
--- a/script/check-simulation-build-cmake
+++ b/script/check-simulation-build-cmake
@@ -40,7 +40,7 @@ build_all_features()
 {
     # Build Thread 1.1 with full features
     reset_source
-    "$(dirname "$0")"/cmake-build simulation
+    "$(dirname "$0")"/cmake-build simulation -DOT_THREAD_VERSION=1.1
 
     # Build Thread 1.2 with full features
     reset_source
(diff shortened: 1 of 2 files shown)


script/check-simulation-build-cmake
```diff
@@ -40,31 +40,31 @@ build_all_features()
 {
     # Build Thread 1.1 with full features
     reset_source
-    "$(dirname "$0")"/cmake-build simulation
+    "$(dirname "$0")"/cmake-build simulation -DOT_THREAD_VERSION=1.1
 
     # Build Thread 1.2 with full features
     reset_source
-    "$(dirname "$0")"/cmake-build simulation -DOT_OTNS=ON -DOT_SIMULATION_VIRTUAL_TIME=ON -DOT_THREAD_VERSION=1.2 -DOT_DUA=ON -DOT_MLR=ON -DOT_BACKBONE_ROUTER=ON -DOT_CSL_RECEIVER=ON
+    "$(dirname "$0")"/cmake-build simulation -DOT_OTNS=ON -DOT_SIMULATION_VIRTUAL_TIME=ON -DOT_DUA=ON -DOT_MLR=ON -DOT_BACKBONE_ROUTER=ON -DOT_CSL_RECEIVER=ON
```
jwhui, line 47: How about keeping the `OT_THREAD_VERSION=1.2` option?

Irving-cl: Sounds good :+1:  Updated.

PR #6782, opened 2021-07-01
[posix] add external routes to kernel on POSIX

superwhd:
This is helpful when POSIX kernel is handling a packet at the thread TUN device.


src/posix/platform/netif.cpp
```diff
@@ -505,6 +510,182 @@ static void UpdateLink(otInstance *aInstance)
...
+        bool                  found = false;
+        while (otNetDataGetNextRoute(aInstance, &iterator, &config) == OT_ERROR_NONE)
+        {
+            if (otIp6IsAddressEqual(&config.mPrefix.mPrefix, &sAddedExternalRoutes[i].mPrefix) &&
+                config.mPrefix.mLength == sAddedExternalRoutes[i].mLength)
+            {
+                found = true;
+                break;
+            }
+        }
```
wgtdkp, line 643: This function nested too much. What about having a short function for matching the added external route? Like:
```c++
otExternalRouteConfig *FindExternalRoute(otInstance *aInstance, const otPrefix &aExternalRoute);
```

src/posix/platform/netif.cpp
```diff
@@ -505,6 +510,182 @@ static void UpdateLink(otInstance *aInstance)
...
+        if (!found)
+        {
+            if (DeleteExternalRoute(sAddedExternalRoutes[i]) == OT_ERROR_NONE)
+            {
+                otIp6Prefix &x = sAddedExternalRoutes[i], &y = sAddedExternalRoutes[sAddedExternalRoutesNum - 1];
+                otIp6Prefix  tmp = x;
+                x                = y;
+                y                = tmp;
+                --sAddedExternalRoutesNum;
+                --i;
```
wgtdkp, line 653: We don't need to swap the ith element with the last one - simply overwrite the ith element with the last one and decrease `sAddedExternalRoutesNum` should work.

src/posix/platform/netif.cpp
```diff
@@ -505,6 +510,182 @@ static void UpdateLink(otInstance *aInstance)
...
+                    }
+                }
+                if (!found)
+                {
+                    AddExternalRoute(config.mPrefix);
+                    sAddedExternalRoutes[sAddedExternalRoutesNum] = config.mPrefix;
+                }
+            }
+        }
+    }
```
wgtdkp, line 683: Ditto, write a short function for the matching logic.

tests/scripts/thread-cert/border_router/test_external_route.py
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env python3
+#
+#  Copyright (c) 2020, The OpenThread Authors.
```
wgtdkp, line 3:
```suggestion
#  Copyright (c) 2021, The OpenThread Authors.
```

src/posix/platform/netif.cpp
```diff
@@ -186,7 +187,11 @@ using namespace ot::Posix::Ip6Utils;
 #endif // OPENTHREAD_TUN_DEVICE
 
 #if defined(__linux__)
-static uint32_t sNetlinkSequence = 0; ///< Netlink message sequence.
+#define kDefaultExternalRoutePriority 512
```
wgtdkp, line 190:
```suggestion
static const uint32_t kDefaultExternalRoutePriority = 512;
```

src/posix/platform/netif.cpp
```diff
@@ -505,6 +510,182 @@ static void UpdateLink(otInstance *aInstance)
...
+    req.msg.rtm_tos      = 0;
+    req.msg.rtm_scope    = RT_SCOPE_UNIVERSE;
+    req.msg.rtm_type     = RTN_UNICAST;
+    req.msg.rtm_table    = RT_TABLE_MAIN;
+    req.msg.rtm_protocol = RTPROT_BOOT;
+    req.msg.rtm_flags    = 0;
+
+    otIp6AddressToString(&aPrefix.mPrefix, addrBuf, OT_IP6_ADDRESS_STRING_SIZE);
+    inet_pton(AF_INET6, addrBuf, data);
+    SuccessOrExit(AddRtAttr(&req.header, sizeof(req), RTA_DST, data, 16 /* IP6 address size in bytes */));
```
wgtdkp, line 573:
```suggestion
    SuccessOrExit(AddRtAttr(&req.header, sizeof(req), RTA_DST, data, sizeof(data));
```

src/posix/platform/netif.cpp
```diff
@@ -505,6 +510,182 @@ static void UpdateLink(otInstance *aInstance)
...
+    if (aLen)
+    {
+        memcpy(RTA_DATA(rta), aData, aLen);
+    }
+    aHeader->nlmsg_len = NLMSG_ALIGN(aHeader->nlmsg_len) + RTA_ALIGN(len);
+exit:
+    return error;
+}
+
+int AddRtAttr32(struct nlmsghdr *aHeader, uint32_t aMaxLen, uint8_t aType, uint32_t aData)
```
wgtdkp, line 534:
```suggestion
otError AddRtAttr32(struct nlmsghdr *aHeader, uint32_t aMaxLen, uint8_t aType, uint32_t aData)
```

src/posix/platform/netif.cpp
```diff
@@ -505,6 +510,182 @@ static void UpdateLink(otInstance *aInstance)
...
+        struct nlmsghdr header;
+        struct rtmsg    msg;
+        char            buf[128];
+    } req{};
+    unsigned char data[sizeof(in6_addr)];
+    char          addrBuf[OT_IP6_ADDRESS_STRING_SIZE];
+    unsigned int  ifIdx = otSysGetThreadNetifIndex();
+
+    VerifyOrExit(sNetlinkFd >= 0);
+    VerifyOrExit(sAddedExternalRoutesNum < kMaxExternalRoutesNum);
```
wgtdkp, line 552: In those `AddExternalRoute` and `DeleteExternalRoute` functions, we are silently ignore any failures, I believe we want to catch or log any errors.

src/posix/platform/netif.cpp
superwhd: The reason I defined this as a macro was I saw this block in this file:
```
#define kAddAddress true
#define kRemoveAddress false
#define kUnicastAddress true
#define kMulticastAddress false
```

I thought the advantage of using a macro is, it will be easier to specify a value during compile time. However, I'm inclined to defining it as a constant, if there's no such demand.

src/posix/platform/netif.cpp
superwhd: Thanks! Good catch.

src/posix/platform/netif.cpp
simonlingoogle: > it will be easier to specify a value during compile time

Do you mean specifying build configuration?

simonlingoogle: Use a POSIX configuration (ON by default) for this feature? I won't be surprised if someone way to disable this feature.

src/posix/platform/netif.cpp
```diff
@@ -523,6 +704,12 @@ void platformNetifStateChange(otInstance *aInstance, otChangedFlags aFlags)
     {
         UpdateLink(aInstance);
     }
+#if defined(__linux__)
+    if (OT_CHANGED_THREAD_NETDATA & aFlags)
+    {
+        UpdateExternalRoutes(aInstance);
```
simonlingoogle, line 731: Can we apply the change after a small delay? (e.g. 1s). This can help us merge multiple Network Data changes to one update.

src/posix/platform/netif.cpp
```diff
@@ -505,6 +510,182 @@ static void UpdateLink(otInstance *aInstance)
...
+        struct nlmsghdr header;
+        struct rtmsg    msg;
+        char            buf[512];
+    } req{};
+    unsigned char data[sizeof(in6_addr)];
+    char          addrBuf[OT_IP6_ADDRESS_STRING_SIZE];
+    unsigned int  ifIdx = otSysGetThreadNetifIndex();
+    otError       error = OT_ERROR_NONE;
+
+    VerifyOrExit(sNetlinkFd >= 0);
```
simonlingoogle: set error ?

src/posix/platform/netif.cpp
```diff
@@ -505,6 +510,182 @@ static void UpdateLink(otInstance *aInstance)
...
+    req.msg.rtm_table    = RT_TABLE_MAIN;
+    req.msg.rtm_protocol = RTPROT_BOOT;
+    req.msg.rtm_flags    = 0;
+
+    otIp6AddressToString(&aPrefix.mPrefix, addrBuf, OT_IP6_ADDRESS_STRING_SIZE);
+    inet_pton(AF_INET6, addrBuf, data);
+    SuccessOrExit(AddRtAttr(&req.header, sizeof(req), RTA_DST, data, 16 /* IP6 address size in bytes */));
+    SuccessOrExit(AddRtAttr32(&req.header, sizeof(req), RTA_OIF, ifIdx));
+
+    VerifyOrExit(send(sNetlinkFd, &req, sizeof(req), 0) > 0);
```
simonlingoogle:
```suggestion
    SuccessOrExit(error = AddRtAttr(&req.header, sizeof(req), RTA_DST, data, 16 /* IP6 address size in bytes */));
    SuccessOrExit(error = AddRtAttr32(&req.header, sizeof(req), RTA_OIF, ifIdx));

    VerifyOrExit(send(sNetlinkFd, &req, sizeof(req), 0) > 0, error = ...);
```

src/posix/platform/netif.cpp
superwhd: > Do you mean specifying build configuration?
Yes. It seems we don't have a strong demand for including this value in the configuration. Therefore, I made it a constant.

src/posix/platform/netif.cpp
```diff
@@ -505,6 +510,182 @@ static void UpdateLink(otInstance *aInstance)
...
+}
+
+static void UpdateExternalRoutes(otInstance *aInstance)
+{
+    for (int i = 0; i < static_cast<int>(sAddedExternalRoutesNum); ++i)
+    {
+        otNetworkDataIterator iterator = OT_NETWORK_DATA_ITERATOR_INIT;
+        otExternalRouteConfig config;
+        bool                  found = false;
+        while (otNetDataGetNextRoute(aInstance, &iterator, &config) == OT_ERROR_NONE)
```
simonlingoogle:
```suggestion
        bool                  found = false;

        while (otNetDataGetNextRoute(aInstance, &iterator, &config) == OT_ERROR_NONE)
```

src/posix/platform/netif.cpp
```diff
@@ -186,7 +187,11 @@ using namespace ot::Posix::Ip6Utils;
 #endif // OPENTHREAD_TUN_DEVICE
 
 #if defined(__linux__)
-static uint32_t sNetlinkSequence = 0; ///< Netlink message sequence.
+#define kDefaultExternalRoutePriority 512
+static uint32_t      sNetlinkSequence        = 0; ///< Netlink message sequence.
+static const uint8_t kMaxExternalRoutesNum   = 8;
```
simonlingoogle: Use a configuration?

src/posix/platform/netif.cpp
```diff
@@ -505,6 +510,182 @@ static void UpdateLink(otInstance *aInstance)
...
+{
+    struct
+    {
+        struct nlmsghdr header;
+        struct rtmsg    msg;
+        char            buf[128];
+    } req{};
+    unsigned char data[sizeof(in6_addr)];
+    char          addrBuf[OT_IP6_ADDRESS_STRING_SIZE];
+    unsigned int  ifIdx = otSysGetThreadNetifIndex();
```
simonlingoogle, line 549: check `ifIdx` > 0?

src/posix/platform/netif.cpp
superwhd: According to an offline discussion, will leave it as is. Maybe we'll add OT timer API in the future so we can implement this mechanism.

wgtdkp (approved):
LGTM :+1:

src/posix/platform/netif.cpp
```diff
@@ -523,6 +727,12 @@ void platformNetifStateChange(otInstance *aInstance, otChangedFlags aFlags)
     {
         UpdateLink(aInstance);
     }
+#if defined(__linux__) && OPENTHREAD_POSIX_CONFIG_UPDATE_EXTERNAL_ROUTE_ENABLE
+    if (OT_CHANGED_THREAD_NETDATA & aFlags)
+    {
+        UpdateExternalRoutes(aInstance);
+    }
+#endif // defined(__linux__) && OPENTHREAD_POSIX_CONFIG_UPDATE_EXTERNAL_ROUTE_ENABLE
```
simonlingoogle, line 735:
```suggestion
#endif
```
We usually don't add comment if the block is small.

src/posix/platform/netif.cpp
```diff
@@ -505,6 +513,202 @@ static void UpdateLink(otInstance *aInstance)
...
+}
+
+bool HasAddedExternalRoute(const otIp6Prefix &aExternalRoute)
+{
+    bool found = false;
+
+    for (int i = 0; i < static_cast<int>(sAddedExternalRoutesNum); ++i)
+    {
+        if (otIp6IsAddressEqual(&sAddedExternalRoutes[i].mPrefix, &aExternalRoute.mPrefix) &&
+            sAddedExternalRoutes[i].mLength == aExternalRoute.mLength)
```
simonlingoogle, line 657:
```suggestion
        if (otIp6IsPrefixEqual(&sAddedExternalRoutes[i], &aExternalRoute))
```

Should we add `otIp6IsPrefixEqual`? It does not make perfect sense to check IPv6 Prefixes using `otIp6IsAddressEqual`.

src/posix/platform/netif.cpp
```diff
@@ -505,6 +513,202 @@ static void UpdateLink(otInstance *aInstance)
...
+        if ((error = DeleteExternalRoute(sAddedExternalRoutes[i])) != OT_ERROR_NONE)
+        {
+            otLogWarnPlat("failed to delete an external route in kernel: %s", otThreadErrorToString(error));
+        }
+        else
+        {
+            sAddedExternalRoutes[i] = sAddedExternalRoutes[sAddedExternalRoutesNum - 1];
+            --sAddedExternalRoutesNum;
+            --i;
+        }
```
simonlingoogle, line 687:
```suggestion
        if ((error = DeleteExternalRoute(sAddedExternalRoutes[i])) != OT_ERROR_NONE)
        {
            otLogWarnPlat("failed to delete an external route in kernel: %s", otThreadErrorToString(error));
        }

        sAddedExternalRoutes[i] = sAddedExternalRoutes[sAddedExternalRoutesNum - 1];
        --sAddedExternalRoutesNum;
        --i;
```
I think we can continue to remove the route from `sAddedExternalRoutes` even when we failed to remove it from kernel, since it's unlikely that retrying will be successful.

src/posix/platform/netif.cpp
```diff
@@ -505,6 +513,202 @@ static void UpdateLink(otInstance *aInstance)
...
+    }
+
+    while (otNetDataGetNextRoute(aInstance, &iterator, &config) == OT_ERROR_NONE)
+    {
+        if (config.mRloc16 == otThreadGetRloc16(aInstance) || HasAddedExternalRoute(config.mPrefix))
+        {
+            continue;
+        }
+        VerifyOrExit(sAddedExternalRoutesNum < kMaxExternalRoutesNum,
+                     otLogWarnPlat("no buffer to add more external routes in kernel"));
```
simonlingoogle, line 697: Better also print the prefix ?

simonlingoogle (commented):
SGTM overall, with some comments.

src/posix/platform/openthread-posix-config.h
```diff
@@ -155,6 +155,36 @@
...
+#endif
+
+/**
+ * @def OPENTHREAD_POSIX_CONFIG_EXTERNAL_ROUTE_PRIORITY
+ *
+ * This macro defines the priority of external routes added to kernel.
+ *
+ */
+#ifndef OPENTHREAD_POSIX_CONFIG_EXTERNAL_ROUTE_PRIORITY
+#define OPENTHREAD_POSIX_CONFIG_EXTERNAL_ROUTE_PRIORITY 512
```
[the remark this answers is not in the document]
superwhd: No strong reason. I just saw most other routes have 256 as the priority.

src/posix/platform/netif.cpp
superwhd: If we removes the route from `sAddedExternalRoutes` then we won't be able to track and remove it from kernel before rebooting. By the way, why is retrying unlike to succeed?

src/posix/platform/netif.cpp
```diff
@@ -505,6 +513,202 @@ static void UpdateLink(otInstance *aInstance)
...
+            sAddedExternalRoutes[i].mLength == aExternalRoute.mLength)
+        {
+            found = true;
+            break;
+        }
+    }
+    return found;
+}
+
+static void UpdateExternalRoutes(otInstance *aInstance)
```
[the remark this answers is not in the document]
superwhd: I wrote this functionality in this file since this file has other functionalities based on netlink. Also, I'd better share the sNetlinkFd with the rest of code. If we want to create a class solely for router management, we may also need to extract shared netlink components and put them in a new class/file.

src/posix/platform/openthread-posix-config.h
```diff
@@ -155,6 +155,36 @@
 #define OPENTHREAD_POSIX_CONFIG_SECURE_SETTINGS_ENABLE 0
 #endif
 
+/**
+ * @def OPENTHREAD_POSIX_CONFIG_UPDATE_EXTERNAL_ROUTE_ENABLE
```
simonlingoogle, line 159:
```suggestion
 * @def OPENTHREAD_POSIX_CONFIG_INSTALL_EXTERNAL_ROUTES_ENABLE
```

src/posix/platform/netif.cpp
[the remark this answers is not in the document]
wgtdkp: I think it is okay to add the new functions in this file. It sounds good to me to refactor this file to define a class, but it will be a huge change. I think it is not necessary to do it in this PR.

src/posix/platform/openthread-posix-config.h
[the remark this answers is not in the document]
simonlingoogle: We need to reduce the priority of external routes. 512 looks fine to me.

src/posix/platform/netif.cpp
simonlingoogle: Fine to keep it as is.

src/posix/platform/netif.cpp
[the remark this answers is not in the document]
simonlingoogle: Fine to me to leave it as is.

src/posix/platform/netif.cpp
```diff
@@ -505,6 +517,207 @@ static void UpdateLink(otInstance *aInstance)
     }
 }
 
+#if OPENTHREAD_POSIX_CONFIG_INSTALL_EXTERNAL_ROUTES_ENABLE
+otError AddRtAttr(struct nlmsghdr *aHeader, uint32_t aMaxLen, uint8_t aType, const void *aData, uint8_t aLen)
```
simonlingoogle, line 521:
```suggestion
void AddRtAttr(struct nlmsghdr *aHeader, uint32_t aMaxLen, uint8_t aType, const void *aData, uint8_t aLen)
```
I think it should never fail

src/posix/platform/netif.cpp
```diff
@@ -505,6 +517,207 @@ static void UpdateLink(otInstance *aInstance)
...
+{
+    uint8_t        len = RTA_LENGTH(aLen);
+    struct rtattr *rta;
+    otError        error = OT_ERROR_NONE;
+
+    VerifyOrExit(NLMSG_ALIGN(aHeader->nlmsg_len) + RTA_ALIGN(len) <= aMaxLen, error = OT_ERROR_NO_BUFS);
```
simonlingoogle, line 527:
```suggestion
    OT_ASSERT(NLMSG_ALIGN(aHeader->nlmsg_len) + RTA_ALIGN(len) <= aMaxLen);
```
I think it should never fail.

src/posix/platform/netif.cpp
```diff
@@ -505,6 +517,207 @@ static void UpdateLink(otInstance *aInstance)
...
+    req.msg.rtm_protocol = RTPROT_BOOT;
+    req.msg.rtm_flags    = 0;
+
+    otIp6AddressToString(&aPrefix.mPrefix, addrBuf, OT_IP6_ADDRESS_STRING_SIZE);
+    inet_pton(AF_INET6, addrBuf, data);
+    SuccessOrExit(error = AddRtAttr(&req.header, sizeof(req), RTA_DST, data, sizeof(data)));
+    SuccessOrExit(error = AddRtAttr32(&req.header, sizeof(req), RTA_PRIORITY, kExternalRoutePriority));
+    SuccessOrExit(error = AddRtAttr32(&req.header, sizeof(req), RTA_OIF, ifIdx));
+
+    VerifyOrExit(send(sNetlinkFd, &req, sizeof(req), 0) > 0, error = OT_ERROR_FAILED);
```
simonlingoogle, line 587:
```suggestion
    VerifyOrDie(send(sNetlinkFd, &req, sizeof(req), 0) > 0);
```
What might be causing error here? If the error is fatal, maybe we can just crash the OT program?
@wgtdkp @bukepo

src/posix/platform/netif.cpp
```diff
@@ -505,6 +517,205 @@ static void UpdateLink(otInstance *aInstance)
...
+    rta->rta_type = aType;
+    rta->rta_len  = len;
+    if (aLen)
+    {
+        memcpy(RTA_DATA(rta), aData, aLen);
+    }
+    aHeader->nlmsg_len = NLMSG_ALIGN(aHeader->nlmsg_len) + RTA_ALIGN(len);
+}
+
+void AddRtAttr32(struct nlmsghdr *aHeader, uint32_t aMaxLen, uint8_t aType, uint32_t aData)
```
simonlingoogle, line 539:
```suggestion
void AddRtAttrUint32(struct nlmsghdr *aHeader, uint32_t aMaxLen, uint8_t aType, uint32_t aData)
```
nit: make it clear

bukepo (approved):

src/posix/platform/netif.cpp
bukepo: maybe we need analyze the errors first.

simonlingoogle: > maybe we need analyze the errors first.

Agree. We can check the `errno`. Errnos like `EAGAIN`, `EINTR`, `EWOULDBLOCK` seems recoverable, while others are fatal.

simonlingoogle (approved):
LGTM 👍

include/openthread/ip6.h
```diff
@@ -561,6 +561,18 @@ const uint16_t *otIp6GetUnsecurePorts(otInstance *aInstance, uint8_t *aNumEntrie
...
+ * Test if two IPv6 prefixes are the same.
+ *
+ * @param[in]  aFirst   A pointer to the first IPv6 prefix to compare.
+ * @param[in]  aSecond  A pointer to the second IPv6 prefix to compare.
+ *
+ * @retval TRUE   The two IPv6 prefixes are the same.
+ * @retval FALSE  The two IPv6 prefixes are not the same.
+ *
+ */
+bool otIp6IsPrefixEqual(const otIp6Prefix *aFirst, const otIp6Prefix *aSecond);
```
jwhui, line 574: nit:
```suggestion
bool otIp6ArePrefixesEqual(const otIp6Prefix *aFirst, const otIp6Prefix *aSecond);
```

src/posix/platform/netif.cpp
jwhui: nit: to avoid confusion with the word `if`?
```suggestion
    unsigned int  netifIndex = otSysGetThreadNetifIndex();
```

src/posix/platform/netif.cpp
```diff
@@ -505,6 +517,212 @@ static void UpdateLink(otInstance *aInstance)
...
+    otExternalRouteConfig config;
+    char                  prefixString[OT_IP6_PREFIX_STRING_SIZE];
+
+    for (int i = 0; i < static_cast<int>(sAddedExternalRoutesNum); ++i)
+    {
+        if (HasExternalRouteInNetData(aInstance, sAddedExternalRoutes[i]))
+        {
+            continue;
+        }
+        if ((error = DeleteExternalRoute(sAddedExternalRoutes[i])) != OT_ERROR_NONE)
```
jwhui, line 688: Is it normal to fail to delete an external route in some situations? What if route was removed by some external process? Should we from `sAddedExternalRoutes` even on failure?

simonlingoogle: What will `DeleteExternalRoute` actually return if the external route has already been removed externally? @superwhd

superwhd: `DeleteExternalRoute` doesn't actually return the result of the delete operation. It returns the result whether the delete request is sent. If it returns an error, that means the delete operation couldn't succeed. On the other hand, if a route has been deleted by someone else, `DeleteExternalRoute` will still return no error as long as the delete request is successfully sent.

jwhui: Sounds good. Thanks.

src/posix/platform/netif.cpp
jwhui: @superwhd , thoughts on this nit?

superwhd: Sorry I missed this one. Updated.

jwhui (approved):

PR #6783, opened 2021-07-01
[dns-client] use NAT64 prefix from network data

abtink:
This commit updates `AddressResponse::GetNat64Prefix()` to search in
Network Data to find NAT64 prefixes added by BRs. Prefixes with
higher preference are selected first.

### DIFF
--- a/src/core/net/dns_client.cpp
+++ b/src/core/net/dns_client.cpp
@@ -350,13 +350,29 @@ exit:
 
 Error Client::AddressResponse::GetNat64Prefix(Ip6::Prefix &aPrefix) const
 {
-    // Use well-known prefix "64:ff9b::/96" (temporary solution).
-    static const uint8_t kWellknownPrefix[] = {0x00, 0x64, 0xff, 0x9b, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+    Error                            error      = kErrorNotFound;
+    NetworkData::Iterator            iterator   = NetworkData::kIteratorInit;
+    signed int                       preference = OT_ROUTE_PREFERENCE_LOW;
+    NetworkData::ExternalRouteConfig config;
 
     aPrefix.Clear();
-    aPrefix.Set(kWellknownPrefix, 96);
 
-    return kErrorNone;
+    while (mInstance->Get<NetworkData::Leader>().GetNextExternalRoute(iterator, config) == kErrorNone)
+    {
+        if (!config.mNat64 || !config.GetPrefix().IsValidNat64())
+        {
+            continue;
+        }
+
+        if ((aPrefix.GetLength() == 0) || (config.mPreference > preference))
+        {
+            aPrefix    = config.GetPrefix();
+            preference = config.mPreference;
+            error      = kErrorNone;
+        }
+    }
+
+    return error;
 }
 
 #endif // OPENTHREAD_CONFIG_DNS_CLIENT_NAT64_ENABLE


src/core/net/dns_client.cpp
```diff
@@ -350,13 +350,29 @@ Error Client::AddressResponse::GetAddress(uint16_t aIndex, Ip6::Address &aAddres
...
+
+        if ((aPrefix.GetLength() == 0) || (config.mPreference > preference))
+        {
+            aPrefix    = config.GetPrefix();
+            preference = config.mPreference;
+            error      = kErrorNone;
+        }
+    }
+
+    return error;
```
simonlingoogle, line 375: Should we fallback to `kWellknownPrefix` when not found in Network Data?

src/core/net/dns_client.cpp
```diff
@@ -350,13 +350,29 @@ Error Client::AddressResponse::GetAddress(uint16_t aIndex, Ip6::Address &aAddres
...
 
-    return kErrorNone;
+    while (mInstance->Get<NetworkData::Leader>().GetNextExternalRoute(iterator, config) == kErrorNone)
+    {
+        if (!config.mNat64)
+        {
+            continue;
+        }
+
+        if ((aPrefix.GetLength() == 0) || (config.mPreference > preference))
```
simonlingoogle, line 367: Should we ignore prefixes with length != 96 (according to https://datatracker.ietf.org/doc/html/rfc6052#section-2.2) ?

wgtdkp: aoocrding to https://datatracker.ietf.org/doc/html/rfc6052#section-2.2,  the NAT64 prefix length can be 32, 40, 48, 56, 64, or 96.

wgtdkp (approved):
LGTM :+1:

src/core/net/dns_client.cpp
jwhui: What benefit does defaulting to `kWellknownPrefix` provide? A BR either needs to discover that a NAT64 prefix exists on the infrastructure or locally generate its own. I think we have to assume that the network data can hold at least one NAT64 prefix.

src/core/net/dns_client.cpp
abtink: We already check the prefixes to be a valid NAT64 prefix (`Prefix::IsValidNat64()`) before letting it be added to network data (on leader).

With this, though unlikely we can still possibility have a leader that does not perform such a check (e.g., if leader is not running OT stack) and let an invalid NAT64 prefix to be added in Network Data. So may be good to do such a check again here and skip over any invalid prefix. Added such a check in  the new push.

src/core/net/dns_client.cpp
abtink: The use of the hard-coded (well-known) prefix was just a temporary solution (mainly as a placeholder) till we agreed and finalized the mechanism for device to discover the NAT64 prefix (which we now have using Network Data external route entries).

If there are no NAT64 prefixes in network data, we return back `kErrorNotFound` (informing caller that address resolution failed and could not find any address).

librasungirl (approved):

src/core/net/dns_client.cpp
simonlingoogle: The obvious benefit of defaulting to `64:ff9b::/96` is that if the NAT64 prefix is indeed `64:ff9b::/96`, which is very likely, we can save the Network Data space by not adding it. 

But this will require wide support at the end-device side.

jwhui: I think we should require NAT64 prefix in the network data for clients to make use of NAT64 service. Lack of NAT64 prefix is an explicit statement that NAT64 service is not available.

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6784, opened 2021-07-02
[netdata] allow zero-len external route prefix

abtink:
This commit allows a zero-len prefix `::/0` to be added as an external
route in Network Data. Such a prefix can act as a default route.

### DIFF
--- a/tests/scripts/expect/cli-misc.exp
+++ b/tests/scripts/expect/cli-misc.exp
@@ -164,6 +164,13 @@ expect "0:0:0:0::/64 s low"
 send "route remove ::/64\n"
 expect_line "Done"
 
+send "route add ::/0 s low\n"
+expect_line "Done"
+send "route\n"
+expect "0:0:0:0::/0 s low"
+send "route remove ::/0\n"
+expect_line "Done"
+
 send "diag start\n"
 expect ": InvalidState"
 
--- a/tests/scripts/thread-cert/test_zero_len_external_route.py
+++ b/tests/scripts/thread-cert/test_zero_len_external_route.py
@@ -33,7 +33,7 @@ import thread_cert
 
 # Test description:
 #
-#   This test verifies the behavior of network when a zero length external router prefix "::/0" is added.
+#   This test verifies the behavior of the network when a zero length external route prefix "::/0" is added.
 #
 # Topology:
 #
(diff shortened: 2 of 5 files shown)


tests/scripts/expect/cli-misc.exp
```diff
@@ -164,6 +164,13 @@ expect "0:0:0:0::/64 s low"
 send "route remove ::/64\n"
 expect_line "Done"
 
+send "route add ::/0 s low\n"
```
simonlingoogle, line 167: Should we also test if this external route is working?

simonlingoogle (commented):
LGTM :+1: 

Better to test if the default route really works.

abtink:
In new pushed commit added a testcase for zero len external route.

simonlingoogle (approved):
LGTM

tests/scripts/thread-cert/test_zero_len_external_route.py
```diff
@@ -0,0 +1,145 @@
...
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
+import unittest
+
+import thread_cert
+
+# Test description:
+#
+#   This test verifies the behavior of network when a zero length external router prefix "::/0: is added.
```
wgtdkp, line 36:
```suggestion
#   This test verifies the behavior of network when a zero length external router prefix "::/0" is added.
```

wgtdkp (approved):
LGTM 👍

tests/scripts/thread-cert/test_zero_len_external_route.py
```diff
@@ -0,0 +1,145 @@
...
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
+import unittest
+
+import thread_cert
+
+# Test description:
+#
+#   This test verifies the behavior of network when a zero length external router prefix "::/0" is added.
```
jwhui, line 36:
```suggestion
#   This test verifies the behavior of the network when a zero length external route prefix "::/0" is added.
```

tests/scripts/thread-cert/test_zero_len_external_route.py
abtink: Thanks @jwhui I think you changed this already. 😃

jwhui (approved):

PR #6785, opened 2021-07-02
[srp-server] avoid re-accessing committed SRP updates

wgtdkp:
We should remove the `UpdateMetadata`  before committing the SRP update to ensure that
the same `UpdateMetadata` (and the `Host` object) can never be touched again.

### DIFF
--- a/src/core/net/srp_server.cpp
+++ b/src/core/net/srp_server.cpp
@@ -305,9 +305,8 @@ void Server::HandleServiceUpdateResult(ServiceUpdateId aId, Error aError)
 
 void Server::HandleServiceUpdateResult(UpdateMetadata *aUpdate, Error aError)
 {
-    CommitSrpUpdate(aError, aUpdate->GetDnsHeader(), aUpdate->GetHost(), aUpdate->GetMessageInfo());
-
     IgnoreError(mOutstandingUpdates.Remove(*aUpdate));
+    CommitSrpUpdate(aError, aUpdate->GetDnsHeader(), aUpdate->GetHost(), aUpdate->GetMessageInfo());
     aUpdate->Free();
 
     if (mOutstandingUpdates.IsEmpty())


abtink (approved):
LGTM. 

Just wondering if this a bug/issue (I think it's not). Is there any situation where current code can cause an issue?

wgtdkp:
> LGTM.
> 
> Just wondering if this a bug/issue (I think it's not). Is there any situation where current code can cause an issue?

Yes the same `UpdateMetadata` should not be matched again as we are using increasing `UpdateId` for indexing. But there isn't a way for us to associate the `UpdateId` to each RPC call to platform mDNS (e.g. mDNSResponder and avahi) in the Advertising Proxy in ot-br-posix. Actually, we are matching the SRP updates by service and host names: https://github.com/openthread/ot-br-posix/blob/8751bde1140b1c1ce93827273cee362d9a69d131/src/agent/advertising_proxy.cpp#L168. That said, in current implementation, there are possible chances that the same `UpdateMetadata` can be matched twice, even if we erase it immediately after calling `CommitSrpUpdate`.

The same `UpdateMetatdata`/`Host` may be accessed-after-free with below flow in ot-br-posix:
1. Set the lease time to 20 seconds;
3. Add a break point at `CommitSrpUpdate` and register a service `my-service._ipps._tcp` and host `my-host`;
4. The program will hit the break point at `CommitSrpUpdate` and let's wait for 25 seconds before continuing the program;
5. The newly added service and host will expire when we call `HandleLeaseTimer` at the end of `CommitSrpUpdate`. And this is propagated to Advertising Proxy causing the method `Server::HandleServiceUpdateResult` being called again in the chain.

It may not be realistic for `CommitSrpUpdate` to take 20 seconds to complete but I think it is always right for us to remove all references to an object before it is released (i.g. remove all references of the`mHost` in `UpdateMetadata` before calling `CommitSrpUpdate` which will free the `mHost`).

librasungirl (approved):

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6786, opened 2021-07-02
[srp-server] track lease time per service instance

wgtdkp:
The SRP draft 09 spec suggests that the SRP Servers SHOULD also track a lease time per service instance. 
>SRP Servers SHOULD also track a lease time per service instance.  The
   reason for doing this is that a client may re-register a host with a
   different set of services, and not remember that some different
   service instance had previously been registered.  In this case, when
   that service instance lease expires, the SRP server SHOULD remove the
   service instance (although the KEY record for the service instance
   SHOULD be retained until the key lease on that service expires).
   This is beneficial because if the SRP client continues to renew the
   host, but never mentions the stale service again, the stale service
   will continue to be advertised.

### DIFF
--- a/src/core/net/srp_server.cpp
+++ b/src/core/net/srp_server.cpp
@@ -333,6 +333,11 @@ void Server::CommitSrpUpdate(Error                    aError,
 
     aHost.SetLease(grantedLease);
     aHost.SetKeyLease(grantedKeyLease);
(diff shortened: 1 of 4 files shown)


librasungirl (approved):

simonlingoogle (approved):
LGTM :+1:

abtink (approved):
Looks great. Thanks. 👍 👍

jwhui (approved):

PR #6787, opened 2021-07-02
[dns-client] feature to auto set the default server address

abtink:
This commit adds a new feature in DNS client to automatically set and
update the server's IPv6 address in the default config. This is done
only when user does not explicitly set or specify the server address.
This behavior requires SRP client and its auto-start feature to be
enabled. SRP client will then monitor the Thread Network Data for
DNS/SRP Service entries to select an SRP server. The selected SRP
server address is also set as the DNS server address in the default
config.

This commit also adds `test_dns_client_config_auto_start.py` which
verifies the behavior of the new feature. In particular, that the DNS
client's default config gets changed and mirrors the SRP client
selected server address (when not set by user) and that it remains
unchanged if user explicitly sets the address in the default config.

### DIFF
--- a/src/core/net/dns_client.cpp
+++ b/src/core/net/dns_client.cpp
@@ -557,6 +557,9 @@ Client::Client(Instance &aInstance)
     , mSocket(aInstance)
     , mTimer(aInstance, Client::HandleTimer)
     , mDefaultConfig(QueryConfig::kInitFromDefaults)
+#if OPENTHREAD_CONFIG_DNS_CLIENT_DEFAULT_SERVER_ADDRESS_AUTO_SET_ENABLE
+    , mUserDidSetDefaultAddress(false)
+#endif
 {
     static_assert(kIp6AddressQuery == 0, "kIp6AddressQuery value is not correct");
 #if OPENTHREAD_CONFIG_DNS_CLIENT_NAT64_ENABLE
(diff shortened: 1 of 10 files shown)


tests/scripts/thread-cert/node.py
```diff
@@ -2479,6 +2479,21 @@ def send_proactive_backbone_notification(self, target: str, mliid: str, ltt: int
         self.send_command(cmd)
         self._expect_done()
 
+    def dns_get_config(self):
+        cmd = f'dns config'
```
wgtdkp, line 2483: What about adding a documentation for the return value, or give a return value example? So that, we can know what fields we have the next time we want to use this method.

For example:
https://github.com/openthread/openthread/blob/48b129e74b19bef719cf8388906ef2e7903eae87/tests/scripts/thread-cert/node.py#L854-L874
and
https://github.com/openthread/openthread/blob/48b129e74b19bef719cf8388906ef2e7903eae87/tests/scripts/thread-cert/node.py#L222-L249

wgtdkp (approved):
LGTM :+1:

simonlingoogle (commented):
> This behavior requires SRP client and its auto-start feature to be enabled. 

I wonder if we can use this behavior without `auto-start feature`.

src/core/net/dns_client.cpp
```diff
@@ -589,12 +592,34 @@ void Client::SetDefaultConfig(const QueryConfig &aQueryConfig)
...
+#endif
 }
 
 void Client::ResetDefaultConfig(void)
 {
     mDefaultConfig = QueryConfig(QueryConfig::kInitFromDefaults);
+
+#if OPENTHREAD_CONFIG_DNS_CLIENT_DEFAULT_SERVER_ADDRESS_AUTO_SET_ENABLE
+    mUserDidSetDefaultAddress = false;
+    UpdateDefaultConfigAddress();
```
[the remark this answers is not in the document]
abtink: Good point! Thanks. Changed it in new push.

tests/scripts/thread-cert/node.py
abtink: Changed in new push.

abtink:
> I wonder if we can use this behavior without auto-start feature.

The current code requires the auto-start feature. This PR is just providing some convenience for the user by automatically changing the default DNS config to mirror what SRP client is using (user can choose to not use this, and parse network data and select address themselves).

In general SRP and DNS-SD are enabled and used together. If there is a need/use-case to support them separately we can consider in future to refactor some of the code from auto-start into a new module that can then be used by user and/or SRP/DNS clients.

simonlingoogle:
> > I wonder if we can use this behavior without auto-start feature.
> 
> The current code requires the auto-start feature. This PR is just providing some convenience for the user by automatically changing the default DNS config to mirror what SRP client is using (user can choose to not use this, and parse network data and select address themselves).
> 
> In general SRP and DNS-SD are enabled and used together. If there is a need/use-case to support them separately we can consider in future to refactor some of the code from auto-start into a new module that can then be used by user and/or SRP/DNS clients.

Fine to me. 
From another standpoint, users who will configure the SRP server manually should also be capable of configuring DNS server manually.

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6788, opened 2021-07-02
[mle] security policy router upgrading rules

edmont:
Implement rules mandated by the network Security Policy for REED upgrading to router.

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -111,7 +111,24 @@ void MleRouter::HandlePartitionChange(void)
 
 bool MleRouter::IsRouterEligible(void) const
 {
-    return mRouterEligible && IsFullThreadDevice();
+    bool                 rval      = true;
+    const SecurityPolicy secPolicy = Get<KeyManager>().GetSecurityPolicy();
+
+    VerifyOrExit(mRouterEligible && IsFullThreadDevice(), rval = false);
+
+#if OPENTHREAD_CONFIG_THREAD_VERSION == OT_THREAD_VERSION_1_1
+    VerifyOrExit(secPolicy.mRoutersEnabled, rval = false);
+#else
+    if (secPolicy.mCommercialCommissioningEnabled)
+    {
+        VerifyOrExit(secPolicy.mNonCcmRoutersEnabled, rval = false);
+    }
+    VerifyOrExit(secPolicy.mVersionThresholdForRouting + kVersionThresholdOffsetVersion <= kThreadVersion,
+                 rval = false);
+#endif
+
+exit:
+    return rval;
 }
 
 Error MleRouter::SetRouterEligible(bool aEligible)
@@ -153,16 +170,6 @@ Error MleRouter::BecomeRouter(ThreadStatusTlv::Status aStatus)
     VerifyOrExit(!IsDisabled(), error = kErrorInvalidState);
     VerifyOrExit(!IsRouter(), error = kErrorNone);
     VerifyOrExit(IsRouterEligible(), error = kErrorNotCapable);
-
-#if OPENTHREAD_CONFIG_THREAD_VERSION == OT_THREAD_VERSION_1_1
-    VerifyOrExit(Get<KeyManager>().GetSecurityPolicy().mRoutersEnabled, error = kErrorNotCapable);
-#else
-    if (Get<KeyManager>().GetSecurityPolicy().mCommercialCommissioningEnabled)
-    {
-        VerifyOrExit(Get<KeyManager>().GetSecurityPolicy().mNonCcmRoutersEnabled, error = kErrorNotCapable);
-    }
-    VerifyOrExit(Get<KeyManager>().GetSecurityPolicy().mVersionThresholdForRouting < 1, error = kErrorNotCapable);
-#endif
 
     otLogInfoMle("Attempt to become router");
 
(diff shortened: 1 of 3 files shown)


src/core/thread/mle_router.cpp
```diff
@@ -154,6 +154,16 @@ Error MleRouter::BecomeRouter(ThreadStatusTlv::Status aStatus)
...
     VerifyOrExit(IsRouterEligible(), error = kErrorNotCapable);
 
+#if OPENTHREAD_CONFIG_THREAD_VERSION == OT_THREAD_VERSION_1_1
+    VerifyOrExit(Get<KeyManager>().GetSecurityPolicy().mRoutersEnabled, error = kErrorNotCapable);
+#else
+    if (Get<KeyManager>().GetSecurityPolicy().mCommercialCommissioningEnabled)
+    {
+        VerifyOrExit(Get<KeyManager>().GetSecurityPolicy().mNonCcmRoutersEnabled, error = kErrorNotCapable);
+    }
+    VerifyOrExit(Get<KeyManager>().GetSecurityPolicy().mVersionThresholdForRouting < 1, error = kErrorNotCapable);
```
simonlingoogle, line 164: Define a constant for `1` ?

simonlingoogle (approved):
LGTM 👍 

with a nit

src/core/thread/mle_router.cpp
librasungirl: According to Table 8-10, may update as below to be more generic?

`!(Get<KeyManager>().GetSecurityPolicy().mRoutersEnable)
&& (Get<KeyManager>().GetSecurityPolicy().mVersionThresholdForRouting + 3 <=
 kThreadVersion)`

src/core/thread/mle_router.cpp
```diff
@@ -154,6 +154,16 @@ Error MleRouter::BecomeRouter(ThreadStatusTlv::Status aStatus)
...
+#endif
```
librasungirl, line 165: How about moving the new code into `IsRouterEligible()` so that Router as well as Leader could follow the security policy rules?

src/core/thread/mle_router.cpp
```diff
@@ -111,7 +111,24 @@ void MleRouter::HandlePartitionChange(void)
...
+    VerifyOrExit(mRouterEligible && IsFullThreadDevice(), rval = false);
+
+#if OPENTHREAD_CONFIG_THREAD_VERSION == OT_THREAD_VERSION_1_1
+    VerifyOrExit(secPolicy.mRoutersEnabled, rval = false);
+#else
+    if (secPolicy.mCommercialCommissioningEnabled)
+    {
+        VerifyOrExit(secPolicy.mNonCcmRoutersEnabled, rval = false);
+    }
+    VerifyOrExit(secPolicy.mVersionThresholdForRouting + kVersionThresholdOffsetVersion <= kThreadVersion,
```
librasungirl, line 126: If `R` bit is 1,  Protocol version 1 or 2, or higher could become router, no need to check `VR` bits. That is,  only need check `VR` bits when `R` is 0.

Thoughts?

edmont: You are right. Guard added.

librasungirl (approved):

edmont:
It seems the CI errors are unrelated.

src/core/thread/mle_router.cpp
```diff
@@ -111,7 +111,27 @@ void MleRouter::HandlePartitionChange(void)
 
 bool MleRouter::IsRouterEligible(void) const
 {
-    return mRouterEligible && IsFullThreadDevice();
+    bool                 rval      = true;
+    const SecurityPolicy secPolicy = Get<KeyManager>().GetSecurityPolicy();
```
wgtdkp, line 115:
```suggestion
    const SecurityPolicy &secPolicy = Get<KeyManager>().GetSecurityPolicy();
```

src/core/thread/mle_types.hpp
```diff
@@ -302,6 +302,16 @@ enum class ChildDuaState : uint8_t
...
 
+/**
+ * Security Policy
+ *
+ */
+enum
+{
+    kVersionThresholdOffsetVersion = 3, ///< Offset between the Thread Version and the
+                                        ///< Version-thresold valid for Routing.
+};
```
wgtdkp, line 313: It is better to define this enum in class `SecurityPolicy`.

wgtdkp (approved):
LGTM 👍  Thanks!

src/core/thread/mle_types.hpp
edmont: Agree. Done.

src/core/thread/mle_router.cpp
edmont: Done.

src/core/thread/mle_router.cpp
```diff
@@ -111,7 +112,28 @@ void MleRouter::HandlePartitionChange(void)
 
 bool MleRouter::IsRouterEligible(void) const
 {
-    return mRouterEligible && IsFullThreadDevice();
+    bool                  rval      = true;
```
jwhui, line 115: nit: I think you could slightly simplify / reduce code size by:

1. Initializing to `false` here.
2. Removing `rval = false` from each `VerifyOrExit()` below.
3. Adding `rval = true` just before the `exit:` label.

Thoughts?

edmont: Done, thanks for the suggestion.

jwhui (approved):
Thanks! 👍🏼

PR #6799, opened 2021-07-07
[link-metrics] fix Link Metrics Status Sub-TLV value

edmont:
When the Initiator requests a reserved Metric Type ID Flag the status value should be "Value 254: Failure – Other failure".


Irving-cl (approved):
LGTM 👍

jwhui (approved):
Thanks! 👍🏼

PR #6800, opened 2021-07-07
[mle] include the CCM flag in the Discovery Response

edmont:
Commercial Commissioning Mode flag from the Security Policy has to be set in the Discovery Response TLV of the MLE Discovery Response messages.

### DIFF
--- a/src/core/meshcop/meshcop_tlvs.hpp
+++ b/src/core/meshcop/meshcop_tlvs.hpp
@@ -2248,6 +2248,33 @@ public:
         }
     }
 
+    /**
+     * This method indicates whether or not the Commercial Commissioning Mode flag is set.
+     *
+     * @retval TRUE   If the Commercial Commissioning Mode flag is set.
+     * @retval FALSE  If the Commercial Commissioning Mode flag is not set.
+     *
+     */
+    bool IsCommercialCommissioningMode(void) const { return (mFlags & kCCMMask) != 0; }
+
+    /**
+     * This method sets the Native Commissioner flag.
+     *
+     * @param[in]  aNativeCommissioner  TRUE if set, FALSE otherwise.
+     *
+     */
+    void SetCommercialCommissioningMode(bool aCCM)
+    {
+        if (aCCM)
+        {
+            mFlags |= kCCMMask;
+        }
+        else
+        {
+            mFlags &= ~kCCMMask;
+        }
+    }
+
 private:
     enum
     {
@@ -2255,6 +2282,8 @@ private:
         kVersionMask   = 0xf << kVersionOffset,
         kNativeOffset  = 3,
         kNativeMask    = 1 << kNativeOffset,
+        kCCMOffset     = 2,
+        kCCMMask       = 1 << kCCMOffset,
     };
     uint8_t mFlags;
     uint8_t mReserved;
--- a/src/core/meshcop/meshcop_tlvs.hpp
+++ b/src/core/meshcop/meshcop_tlvs.hpp
@@ -2258,9 +2258,9 @@ public:
     bool IsCommercialCommissioningMode(void) const { return (mFlags & kCCMMask) != 0; }
 
     /**
-     * This method sets the Native Commissioner flag.
+     * This method sets the Commercial Commissioning Mode flag.
      *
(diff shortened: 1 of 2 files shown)


simonlingoogle (approved):
LGTM 👍 

Thanks

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -2248,13 +2248,42 @@ class DiscoveryResponseTlv : public Tlv, public TlvInfo<Tlv::kDiscoveryResponse>
...
+     * @retval TRUE   If the Commercial Commissioning Mode flag is set.
+     * @retval FALSE  If the Commercial Commissioning Mode flag is not set.
+     *
+     */
+    bool IsCommercialCommissioningMode(void) const { return (mFlags & kCCMMask) != 0; }
+
+    /**
+     * This method sets the Native Commissioner flag.
+     *
+     * @param[in]  aNativeCommissioner  TRUE if set, FALSE otherwise.
```
librasungirl, line 2263:
```suggestion
     * This method sets the Commercial Commissioning Mode flag.
     *
     * @param[in]  aCCM  TRUE if set, FALSE otherwise.
```

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -2248,13 +2248,42 @@ class DiscoveryResponseTlv : public Tlv, public TlvInfo<Tlv::kDiscoveryResponse>
...
     }
 
+    /**
+     * This method indicates whether or not the Commercial Commissioning Mode flag is set.
+     *
+     * @retval TRUE   If the Commercial Commissioning Mode flag is set.
+     * @retval FALSE  If the Commercial Commissioning Mode flag is not set.
+     *
+     */
+    bool IsCommercialCommissioningMode(void) const { return (mFlags & kCCMMask) != 0; }
```
wgtdkp, line 2258:
```suggestion
    bool IsCommercialCommissioningMode(void) const { return !(mFlags & kCCMMask); }
```
The CCM flag has reverse meaning - `1` indicates that CCM mode is disabled.

src/core/meshcop/meshcop_tlvs.hpp
```diff
@@ -2248,13 +2248,42 @@ class DiscoveryResponseTlv : public Tlv, public TlvInfo<Tlv::kDiscoveryResponse>
...
+
+    /**
+     * This method sets the Native Commissioner flag.
+     *
+     * @param[in]  aNativeCommissioner  TRUE if set, FALSE otherwise.
+     *
+     */
+    void SetCommercialCommissioningMode(bool aCCM)
+    {
+        if (aCCM)
```
wgtdkp, line 2268:
```suggestion
        if (!aCCM)
```

src/core/meshcop/meshcop_tlvs.hpp
edmont: That's the case for Security Policy TLV only. For Discovery Request TLV:

> Its value is taken as the negated CCM bit value from the Security Policy TLV.

simonlingoogle: The three negated bits are so confusing:
- kCommercialCommissioningMask
- kAutonomousEnrollmentMask
- kNetworkKeyProvisioningMask

I think we should rename them to be like `kNoCommercialCommissioningMask` or `kNegatedCommercialCommissioningMask`

Thoughts?

librasungirl: @simonlingoogle fyi, kangping should have hidden the negated bits for security policy flags already in #6320, like
https://github.com/openthread/openthread/blob/4484042ced953b9d24a47cc8f3a04ea0e531e18c/src/core/thread/key_manager.cpp#L91

librasungirl (approved):

src/core/meshcop/meshcop_tlvs.hpp
wgtdkp: @edmont Thanks for clarification! I missed the fact that the code is for the `Discovery Response TLV`.

wgtdkp (approved):
LGTM 👍 
Thanks!

jwhui (approved):
Thanks! 👍🏼

PR #6806, opened 2021-07-09
[network-diagnostic] fix initialization of max child timeout

jinran-google:
`MleRouter::GetMaxChildTimeout` relies on the initial value of `maxTimeout`. This adds the correct initialization.

### DIFF
--- a/src/core/thread/mle_router.cpp
+++ b/src/core/thread/mle_router.cpp
@@ -4337,9 +4337,9 @@ Error MleRouter::GetMaxChildTimeout(uint32_t &aTimeout) const
 {
     Error error = kErrorNotFound;
 
-    VerifyOrExit(IsRouterOrLeader(), error = kErrorInvalidState);
-
     aTimeout = 0;
+
+    VerifyOrExit(IsRouterOrLeader(), error = kErrorInvalidState);
 
     for (Child &child : Get<ChildTable>().Iterate(Child::kInStateValid))
     {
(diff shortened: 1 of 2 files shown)


jwhui (approved):

src/core/thread/network_diagnostic.cpp
```diff
@@ -430,7 +430,7 @@ Error NetworkDiagnostic::FillRequestedTlvs(const Message &       aRequest,
 #if OPENTHREAD_FTD
         case NetworkDiagnosticTlv::kMaxChildTimeout:
         {
-            uint32_t maxTimeout;
+            uint32_t maxTimeout = 0;
```
jwhui, line 433: Should we update `GetMaxChildTimeout()` so that it does not require the client to initialize?

simonlingoogle: +1

jinran-google: Updated.

src/core/thread/mle_router.cpp
```diff
@@ -4339,6 +4339,8 @@ Error MleRouter::GetMaxChildTimeout(uint32_t &aTimeout) const
 
     VerifyOrExit(IsRouterOrLeader(), error = kErrorInvalidState);
 
+    aTimeout = 0;
```
simonlingoogle, line 4342: nit: move before `VerifyOrExit` ?

simonlingoogle (approved):
LGTM :+1: 

with a nit

src/core/thread/mle_router.cpp
jinran-google: I agree. Actually I prefer to leave the value untouched when exiting with errors. But current solution is worse because it's not consistent, and fixing that will add significant complexity. I'll move it before `VerifyOrExit` to make it consistent.

abtink (approved):
👍 Thanks.

jwhui (approved):

PR #6807, opened 2021-07-09
[utils] adding `HistoryTracker` module

abtink:
This commit adds History Tracker feature and its CLI support. This
feature records history of different events as the Thread network
operates (e.g., history of RX and TX IPv6 messages or network info
changes).

Recorded entries are timestamped. When the history list is read, the
timestamps are given as the entry age relative to the time the list
is being read. For example in CLI a timestamp can be shown as
`02:31:50.628 ago` indicating the entry was recorded 2 hours, 31 min,
50 sec, and 628 msec ago. Number of days is added for events that are
older than 24 hours, e.g., `31 days 03:00:23.931 ago`. Timestamps use
millisecond accuracy and are tacked up to 49 days. If an event is
older than 49 days, the entry is still tracked in the list but the
timestamp is shown as old or `more than 49 days ago`.

The `HistoryTracker` currently maintains 3 lists. The Network Info
history tracks changes to Device Role, Mode, RLOC16 and Partition ID.
The RX/TX history list records information about the received/sent
IPv6 messages:
- Message type (UDP, TCP, ICMP6 (and its subtype), etc.)
- Source and destination IPv6 addresses and port numbers
- IPv6 payload length
- Whether or not the link-layer security was used
- Message priority: low, norm, high, net (for control messages)
- Short address (RLOC16) of neighbor who send/received the msg
- Received Signal Strength (in dBm) for RX only
- Radio link info (15.4/TREL) on which msg was sent/received
  (useful when `OPENTHREAD_CONFIG_MULTI_RADIO` is enabled)

Config `HISTORY_TRACKER_EXCLUDE_THREAD_CONTROL_MESSAGES` can be used
to configure `HistoryTracker` to exclude Thread Control message
(e.g., MLE, TMF) from TX and RX history.

The number of entries recorded for each history list is configurable
through a set of OpenThread config options, e.g., number of entries
in Network Info history list is specified by OpenThread config option
`OPENTHREAD_CONFIG_HISTORY_TRACKER_NET_INFO_LIST_SIZE`. The
`HistoryTracker` will keep the most recent entries overwriting oldest
ones when the list gets full.

This commit also adds support for `HistoryTracker` in CLI. The CLI
commands provide two style for printing the history information: A
table format (more human-readable) and list style (better suited for
parsing by machine/code). `README_HISTORY.md` is added to document
the commands and the info provided by each history list entry.

This commit also adds `test_history_tracker.py` test-case which
covers the behavior of `HistoryTracker`.

--------

This feature is similar to `wpantund`'s `StatCollector` and are both
mainly intended for testing and debugging.


abtink:
In the latest push I add code to skip the newly added `test_history_tracker` test under `distcheck` build only (in `simulation 1.1` workflow).

The new testcase fails under this build. It seems to be related to the long simulation wait times in this test (waiting for days up to 51 days).  Other build flavors seem to be working fine and the test is passing successfully on them. I also ran a stress test of 500 iterations of this test (locally) with no issues.  I tried to do a bit of investigation on the distcheck build itself but could not find any quick fixes for this. Since this test is far from critical and is already covered by other build flavors, I opted to skip it on distcheck build.

src/cli/README_HISTORY.md
```diff
@@ -0,0 +1,290 @@
...
+- Whether or not link-layer security was used.
+- Message priority: low, norm, high, net (for Thread Control traffic).
+- RSS: Received Signal Strength (in dBm) - averaged over all received fragment frames that formed the message.
+- Short address (RLOC16) of neighbor from which message was received (`0xfffe` if not known).
+- Radio link on which the message was received (useful when `OPENTHREAD_CONFIG_MULTI_RADIO` is enabled).
+
+Print the IPv6 message RX history as a table:
+
+```bash
+> history rx
```
wgtdkp, line 123: What about filtering the messages by its type? For example, if I care only about UDP messages, I can use `history rx udp`.

tests/scripts/thread-cert/test_history_tracker.py
```diff
@@ -0,0 +1,206 @@
...
+            self.assertEqual(child_tx[index]['type'], 'ICMP6(EchoReply)')
+
+            self.assertEqual(leader_tx[index]['len'], child_rx[index]['len'])
+            self.assertEqual(leader_rx[index]['len'], child_tx[index]['len'])
+
+
+if __name__ == '__main__':
+    # Skip the test under distcheck build. Error code
+    # 77 means skip this test case in automake tests
+    if os.getenv('DISTCHECK_BUILD') == '1':
```
wgtdkp, line 203: It may be better to have a `TODO` or `FIXME` here to explain why the test case cannot work with distcheck build.

simonlingoogle (commented):
What's the background for this feature?

jwhui:
> What's the background for this feature? @abtink 

@simonlingoogle , users of `wpantund` transitioning to RCP requested that the stats collector module functionality be supported when using RCP mode.

simonlingoogle:
Add to size-report ? @abtink

wgtdkp:
> > What's the background for this feature? @abtink
> 
> @simonlingoogle , users of `wpantund` transitioning to RCP requested that the stats collector module functionality be supported when using RCP mode.

Can the logs in RCP mode already serve this purpose?

jwhui:
> Can the logs in RCP mode already serve this purpose?

@wgtdkp , it depends on the log level and processor capabilities - see #6706.

src/cli/README_HISTORY.md
abtink: These are good suggestions.  Thanks.

This is the initial implementation of the `HistoryTracker` feature (adding its foundations). My intention is that we build on top of this in future PRs.

There are a lot of other ideas/functionality we can later add onto this. I was considering some of the same things that you suggested, my idea is to later add a general filter model: filter based on type, or msg size, or src or dst address, whether it used link security, priority, etc. and/or allow user to define a combination of conditions to filter.

There are also some ideas on new items/entries to have `HistoryTracker` track:
- Child table and neighbor table (keep history when a child/neighbor attaches/detaches, etc).
- History of changes to network data (when prefix/route/service entries are added/removed, etc).
- History of unicast/multicast IPv6 addresses 
- History of route table changes (next hops, router table, router ids, track network topology).
- ...

I suggest leaving more functions/feature for future PRs.

include/openthread/history_tracker.h
```diff
@@ -0,0 +1,211 @@
...
+ *
+ */
+
+/**
+ * This constant specifies the maximum age of entries which is 49 days (in msec).
+ *
+ * Entries older than the max age will give this value as their age.
+ *
+ */
+#define OT_HISTORY_TRACKER_MAX_AGE (49 * 24 * 60 * 60 * 1000u)
```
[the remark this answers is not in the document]
abtink: I don't think it will be useful to make it configurable.

Users can not increase it beyond 49 days, and if they want a shorter interval, then they can simply ignore entries that have longer age.

This value is coming from our design and the fact that we use `uint32_t` value (as `TimeMilli`). `2^32` msec is about 49 days and 17 hours. So we track entries up to 49 days and every 16 hours check for entries that are about to reach the max (mark then as old).

include/openthread/history_tracker.h
```diff
@@ -0,0 +1,211 @@
...
+};
+
+/**
+ * This structure represents a RX/TX IPv6 message info.
+ *
+ * Some of the fields in this struct are applicable to a RX message or a TX message only, e.g., `mAveRxRss` is the
+ * average RSS of all fragment frames that form a received message and is only applicable for a RX message.
+ *
+ */
+typedef struct otHistoryTrackerMessageInfo
```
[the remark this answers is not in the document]
abtink: We usually only used `PACKED` for types that may be added in in message. 
The fields in this `struct` are already defined in an order to try to avoid alignment gaps.

In general, the `HistoryTracker` is not really intended for use on devices that have resource limitation. So memory use should not really considered critical here.

include/openthread/history_tracker.h
```diff
@@ -0,0 +1,211 @@
...
+ * @param[inout] aIterator   A pointer to an iterator. MUST be initialized or the behavior is undefined.
+ * @param[out]   aEntryAge   A pointer to a variable to output the entry's age. MUST NOT be NULL.
+ *                           Age is provided as the duration (in milliseconds) from when entry was recorded to
+ *                           @p aIterator initialization time. It is set to `OT_HISTORY_TRACKER_MAX_AGE` for entries
+ *                           older than max age.
+ *
+ * @returns A pointer to `otHistoryTrackerNetworkInfo` entry or `NULL` if no more entries in the list.
+ *
+ */
+const otHistoryTrackerNetworkInfo *otHistoryTrackerIterateNetInfoHistory(otInstance *              aInstance,
```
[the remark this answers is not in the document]
abtink: I do like and prefer the `Iterate` method name a lot better for these APIs (specially since we are are using an `Iterator`).

Funny thing is I used `GetNext()` name model initially when I was writing the code as a placeholder so that later I can find a better name 😄  (which I then changed to `Iterate{SomeList}()`).  `GetNext()` in my view is tied to linked-list.

----

I also found some interesting discussions (they point that it is  grammatically correct to say "iterate the list" compared to "iterate over the list" or "iterate through a list" (though the later two are used often) -> https://stackoverflow.com/questions/2715388/is-it-iterate-through-or-iterate-over-something

include/openthread/history_tracker.h
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * If the resulting string does not fit in @p aBuffer (within its @p aSize characters), the string will be truncated
+ * but the outputted string is always null-terminated.
+ *
+ * @param[in]  aEntryAge The entry age (duration in msec).
+ * @param[out] aBuffer   A pointer to a char array to output the string (MUST NOT be NULL).
+ * @param[in]  aSize     The size of @p aBuffer (in bytes). Recommended to use `OT_IP6_ADDRESS_STRING_SIZE`.
+ *
+ */
+void otHistoryTrackerEntryAgeToString(uint32_t aEntryAge, char *aBuffer, uint16_t aSize);
```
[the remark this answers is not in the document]
abtink: I am not sure if this will be useful. This will restrict the API use (caller need to give an exact array size or do pointer array casting which is tricky and can be error-prone).
- We want to allow user to give us any buffer with whatever size they want (e.g. user may have a large static buffer which they use for all string conversions).
- Other public APIs that convert to string accept any buffer.

src/core/config/history_tracker.h
```diff
@@ -0,0 +1,94 @@
...
+#define CONFIG_HISTORY_TRACKER_H_
+
+/**
+ * @def OPENTHREAD_CONFIG_HISTORY_TRACKER_ENABLE
+ *
+ * Define as 1 to enable History Tracker module.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_HISTORY_TRACKER_ENABLE
+#define OPENTHREAD_CONFIG_HISTORY_TRACKER_ENABLE 0
```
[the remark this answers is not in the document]
abtink: Good idea. Will change this in next push.

src/cli/cli.hpp
```diff
@@ -296,6 +298,22 @@ class Interpreter
      */
     void SetUserCommands(const otCliCommand *aCommands, uint8_t aLength, void *aContext);
 
+    static constexpr uint8_t kLinkModeStringSize = 4; ///< Size of string buffer to represent a MLE Link Mode.
```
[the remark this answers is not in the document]
abtink: Good suggestion. Will change in next push.

src/core/utils/history_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+    case kTxMessage:
+        entry = mTxHistory.AddNewEntry();
+        break;
+    }
+
+    VerifyOrExit(entry != nullptr);
+
+    entry->mPayloadLength        = ip6Header.GetPayloadLength();
+    entry->mNeighborRloc16       = aMacAddresss.IsShort() ? aMacAddresss.GetShort() : kInvalidRloc16;
```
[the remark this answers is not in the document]
abtink: This fiels is used for both tx and rx msg. "next hop" is applicable for tx, but not necessarily in the rx case.

src/core/utils/history_tracker.cpp
```diff
@@ -0,0 +1,362 @@
...
+exit:
+    return;
+}
+
+void HistoryTracker::HandleNotifierEvents(Events aEvents)
+{
+    if (!mTimer.IsRunning())
+    {
+        mTimer.Start(kAgeCheckPeriod);
+    }
```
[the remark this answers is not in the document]
abtink: It is due to a restriction of how the timer unit test `test_timer` works. If we start a timer in any constructor, unit test will fail.

I am working on a PR to change the timer unit test to remove this restriction.

src/core/utils/history_tracker.cpp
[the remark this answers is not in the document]
simonlingoogle: Thanks for clarification.

include/openthread/history_tracker.h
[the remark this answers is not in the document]
simonlingoogle: > We want to allow user to give us any buffer with whatever size they want (e.g. user may have a large static buffer which they use for all string conversions).

A larger static buffer can be converted to a `char (*aBuffer)[OT_HISTORY_TRACKER_ENTRY_AGE_STRING_SIZE])` safely, even though it requires some skill of pointer manipulation.

The suggested code can help reduce code size. But fine to me to keep it as is.

include/openthread/history_tracker.h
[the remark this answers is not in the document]
simonlingoogle: We are using `otModuleGetNextSomething` a lot in OT APIs, but never used `otModuleIterateSomething` once. 

Even if `Iterate` may sounds better, we should keep wording in OT APIs consistent?

tests/scripts/thread-cert/test_history_tracker.py
abtink: Extended the commit in the new push.

abtink:
> Add to size-report ? @abtink

I did not add the `OT_HISTORY_TRACKER` in `chec-size` script intentionally since this feature is not really intended for resource-limited devices. It is intended for posix platform (or posix-like platforms) where memory/code size is not critical. For NCP model, it should be host driver (e.g. wpantund) to collect such history.

abtink:
>> What's the background for this feature? @abtink
>
>@simonlingoogle , users of wpantund transitioning to RCP requested that the stats collector module functionality be supported when using RCP mode.

As @jwhui  mentioned, this is replacing `StatCollector` feature in `wpantund`
- `StatCollector` was very useful, helped us a lot in finding issues and bugs. 
- Product team (testing) also relied on it and used it a lot during OT integration.

> Can the logs in RCP mode already serve this purpose?

Logs are always helpful but as Jonathan mentioned, they may not be enabled or available. 

During lab or in-field testings, it was common that logs were missing or not collected during critical times (specially around the times something was not working and needed debugging). Having device collect history of different events and then being able to get it later was very useful. We also have a feature "auto-log" in `StatCollector` (which we may later add in `HistoryTracker` as well), which dumped a subset of collected history in the logs periodically (e.g., every day).

include/openthread/history_tracker.h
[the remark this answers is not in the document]
abtink: I think using `Iterate` name here is better. It would be clear to OT API users what it means. We use `Iterate` in internal core in many places. It may be better to mirror the API names with the core internal names.

etc/cmake/options.cmake
```diff
@@ -204,6 +204,11 @@ if(OT_EXTERNAL_HEAP)
     target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_HEAP_EXTERNAL_ENABLE=1")
 endif()
 
+option(OT_HISTORY_TRACKER "enable history tracker support")
+if(OT_HISTORY_TRACKER)
+    target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_HISTORY_TRACKER_ENABLE=1")
+endif()
```
simonlingoogle, line 210:
```suggestion
if(OT_HISTORY_TRACKER)
    target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_HISTORY_TRACKER_ENABLE=1")
else()
    target_compile_definitions(ot-config INTERFACE "OPENTHREAD_CONFIG_HISTORY_TRACKER_ENABLE=0")
endif()
```

include/openthread/history_tracker.h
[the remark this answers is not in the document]
simonlingoogle: I think `Iterate` makes more sense if we are using a C++ iterator, which is true for internal core code. 

But for C APIs without Iterator, `GetNext` seems more suitable?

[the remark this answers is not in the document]
abtink: We are using an iterator in this API which we pass as a parameter.

I think we may be spending too much time discussing names and/or smaller nit/style topics which may not be very productive. Let's adopt the policy in PR reviews that we can suggest name changes, but then leave the decision to the person writing the code. I want us to avoid having a lot of back-and-forth on things like this where all options are fine and/or make little difference practically.

simonlingoogle (approved):
LGTM :+1:

tests/scripts/thread-cert/test_history_tracker.py
```diff
@@ -0,0 +1,209 @@
...
+
+            self.assertEqual(leader_rx[index]['type'], 'ICMP6(EchoReply)')
+            self.assertEqual(child_tx[index]['type'], 'ICMP6(EchoReply)')
+
+            self.assertEqual(leader_tx[index]['len'], child_rx[index]['len'])
+            self.assertEqual(leader_rx[index]['len'], child_tx[index]['len'])
+
+
+if __name__ == '__main__':
+    #  We skip the test under distcheck build (the simulation under
```
wgtdkp, line 201: nit:
```suggestion
    #  FIXME(abtink): We skip the test under distcheck build (the simulation under
```

Add `FIXME` because:
1. It is easier for us to search `FIXMEs` in the future to address pending issues;
2. Many tools supports intellisense with `TODO` and `FIXME`;

wgtdkp (approved):
LGTM :+1:

tests/scripts/thread-cert/test_history_tracker.py
abtink: Done!

include/openthread/history_tracker.h
```diff
@@ -0,0 +1,211 @@
...
+    otSockAddr mDestination;         ///< Destination IPv6 address and port (if UDP/TCP).
+    uint8_t    mIpProto;             ///< IP Protocol number (`OT_IP6_PROTO_*` enumeration).
+    uint8_t    mIcmp6Type;           ///< ICMP6 type if msg is ICMP6, zero otherwise (`OT_ICMP6_TYPE_*` enumeration).
+    int8_t     mAveRxRss;            ///< RSS of received message or OT_RADIO_INVALI_RSSI if not known.
+    bool       mLinkSecurity : 1;    ///< Indicates whether msg used link security.
+    bool       mTxSuccess : 1;       ///< Indicates TX success (e.g., ack received). Applicable for TX msg only.
+    uint8_t    mPriority : 2;        ///< Message priority (`OT_HISTORY_TRACKER_MSG_PRIORITY_*` enumeration).
+    bool       mRadioIeee802154 : 1; ///< Indicates whether msg was sent/received over a 15.4 radio link.
+    bool       mRadioTrelUdp6 : 1;   ///< Indicates whether msg was sent/received over a TREL radio link.
+} otHistoryTrackerMessageInfo;
```
jwhui, line 120: What about adding the UDP/TCP/ICMPv6 checksum?

src/cli/README_HISTORY.md
```diff
@@ -0,0 +1,290 @@
...
+- [rx](#rx)
+- [tx](#tx)
+
+## Timestamp Format
+
+Recorded entries are timestamped. When the history list is printed, the timestamps are shown relative the time the command was issues (i.e., when the list was printed) indicating how long ago the entry was recorded.
+
+```bash
+> history netinfo
+| Time                     | Role     | Mode | RLOC16 | Partition ID |
```
jwhui, line 24: How about renaming column to "Age" so we do not have to print "ago" in each entry?

src/cli/README_HISTORY.md
```diff
@@ -0,0 +1,290 @@
...
+Print the Network Info history as a table.
+
+```bash
+> history netinfo
+| Time                     | Role     | Mode | RLOC16 | Partition ID |
++--------------------------+----------+------+--------+--------------+
+|         00:00:10.069 ago | router   | rdn  | 0x6000 |    151029327 |
+|         00:02:09.337 ago | child    | rdn  | 0x2001 |    151029327 |
+|         00:02:09.338 ago | child    | rdn  | 0x2001 |    151029327 |
+|         00:07:40.806 ago | child    | -    | 0x2001 |    151029327 |
```
jwhui, line 74: In what scenario do we get three `child` entries in a row?

src/cli/README_HISTORY.md
```diff
@@ -0,0 +1,290 @@
...
+|         00:03:34.186 ago | src: [fe80:0:0:0:146e:a00:0:1]:19788                          |
+|                          | dst: [ff02:0:0:0:0:0:0:1]:19788                               |
++--------------------------+------------------+-------+-----+------+------+--------+-------+
+|                          | UDP              |    51 |  no |  net |  yes | 0xffff |  15.4 |
+|         00:04:06.502 ago | src: [fe80:0:0:0:146e:a00:0:1]:19788                          |
+|                          | dst: [ff02:0:0:0:0:0:0:1]:19788                               |
+Done
+```
+
+Print the IPv6 message RX history as a list (7 latest entries):
```
jwhui, line 257:
```suggestion
Print the IPv6 message TX history as a list (7 latest entries):
```

include/openthread/history_tracker.h
```diff
@@ -0,0 +1,211 @@
...
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *
+ * @brief
+ *   This file includes the OpenThread API for History Tracker module.
+ */
```
jwhui, line 28:
```suggestion
 */

/**
 * @brief
 *   This file includes the OpenThread API for History Tracker module.
 */
```

src/cli/README_HISTORY.md
abtink: I did notice the same at the time. 😄 

On this node, I did the following, `mode -` , then attached to the network and then later issued `mode rdn`.

This line `00:07:40.806 ago | child    | -    | 0x2001 |    151029327` is for when role change to child (device attached first time)

The other two are for when "mode" changed. Mode change actually can be recorded as if two changes happened.
```
|         00:02:09.337 ago | child    | rdn  | 0x2001 |    151029327 |
|         00:02:09.338 ago | child    | rdn  | 0x2001 |    151029327 |
```

 The reason for this is tied to fact that mode change triggers once from `SetDeviceMode()` itself calling `HistoryTracker::RecordNetworkInfo()` (we don't have a `Notifier` event for mode change) and then we may also get another `Notifier` event (later) for RLOC added/removed events. 

Here is the call chain: `Mle::SetDeviceMode()` to set the new mode which can call `SetStateChild(GetRloc16())` which itself calls `SetRloc16(aRloc16)` which alwys removes and re-adds `mMeshLocal16` (even if the rloc16 is same), which trigger notififer event` kEventThreadRlocRemoved` and `kEventThreadRlocAdded`

I opted to let it behave like this for now, but was considering to address it later (in a seperate PR):
- One  solution is to add a `Notifier` event for mode change then all events are emitted together and we process it once. I slightly prefer this one. The only downside is consuming one of the bits in `OT_CHANGED_*` - we are already using 30 bits and it's 32 bit value so 2 bit left)
- We can also consider changing the `SetRlco16()` code to skip remove and re-add (if the `mMeshLocal16` happens to be same, but I am not sure if it may have other side-effects).

Thoughts/Ideas?

jwhui: Good points. I'm not sure if there's a good reason for always executing the code in `SetRloc16()` even if the short address does not change. I would prefer avoiding repeated execution if there is no good reason for it. I'm find with addressing this in a separate PR.

abtink:
The new pushed commit contains the following changes:
- Record checksum in rx/tx list 
- Added a new CLI command `history rxtx` which shows combined list of all RX and TX msg history
- Unified the CLI table output format for RX/TX history.
- The `README_HISTORY.md` is also updated
- Removed "ago" from timestamp string and use "Age" as title.

-----

For some reason the CI "API version check" is failing even though the commit does change `OPENTHREAD_API_VERSION`

jwhui:
> For some reason the CI "API version check" is failing even though the commit does change `OPENTHREAD_API_VERSION`

The current `OPENTHREAD_API_VERSION` is already 150. https://github.com/openthread/openthread/blob/cd6d57d1affa94ac72d9762a82270e9e8d7347e2/include/openthread/instance.h#L56

abtink:
Thanks @jwhui.

jwhui (approved):

src/cli/README_HISTORY.md
abtink: I tried checking different code-paths. I think changing `SetRloc16()` to avoid removing and re-adding the `mMeshLocag16` address if remains unchanged should work fine. I will submit a PR to do this.

PR #6810, opened 2021-07-12
[network-key] Change thrNetworkKey for docs

gavinmcgovern:


jwhui:
@gavinmcgovern , need to also increment:
https://github.com/openthread/openthread/blob/05fd1deb497451464ff050102d8f2994e53e6820/include/openthread/instance.h#L56

jwhui (approved):

PR #6811, opened 2021-07-12
Add API to obtain Receiver Poll state at run-time.

adabreuti:
It is not currently possible to obtain the Polling state of the receiver from the radio layer via OTLink API's. This API is being added to add support for this use case.

### DIFF
--- a/include/openthread/platform/radio.h
+++ b/include/openthread/platform/radio.h
@@ -833,6 +833,16 @@ otError otPlatRadioEnergyScan(otInstance *aInstance, uint8_t aScanChannel, uint1
 extern void otPlatRadioEnergyScanDone(otInstance *aInstance, int8_t aEnergyScanMaxRssi);
 
 /**
+ * This function indicates whether or not an IEEE 802.15.4 MAC is currently waiting for data.
+ * MAC module is in the receive state during a polling procedure pending external frame transmission.
+ *
+ * @param[in] aInstance A pointer to an OpenThread instance.
+ * @returns true if an IEEE 802.15.4 MAC is in the polling receive state, false otherwise.
+ *
+ */
+extern bool otPlatRadioIsRxStatePolling(otInstance *aInstance);
+
+/**
  * Enable/Disable source address match feature.
  *
  * The source address match feature controls how the radio layer decides the "frame pending" bit for acks sent in
--- a/src/core/mac/mac.hpp
+++ b/src/core/mac/mac.hpp
@@ -551,6 +551,13 @@ public:
     bool IsInTransmitState(void) const;
 
     /**
+     * This method returns if the MAC layer is in polling receive state.
+     *
+     * The MAC layer is in the receive state during a polling procedure pending external frame transmission.
+     */
+    bool IsRxStatePolling(void) const;
+
+    /**
      * This method registers a callback to provide received raw IEEE 802.15.4 frames.
      *
      * @param[in]  aPcapCallback     A pointer to a function that is called when receiving an IEEE 802.15.4 link frame
--- a/src/core/radio/radio_platform.cpp
+++ b/src/core/radio/radio_platform.cpp
@@ -115,6 +115,18 @@ exit:
     return;
 }
 
+bool otPlatRadioIsRxStatePolling(otInstance *aInstance)
+{
+    Instance *instance = static_cast<Instance *>(aInstance);
+    bool ret = false;
+
+    if (instance->IsInitialized())
+    {
+        ret = instance->Get<Mac::Mac>().IsRxStatePolling();
+    }
+    return (ret);
+}
+
 #if OPENTHREAD_CONFIG_DIAG_ENABLE
 extern "C" void otPlatDiagRadioReceiveDone(otInstance *aInstance, otRadioFrame *aFrame, Error aError)
 {
--- a/src/core/radio/radio_platform.cpp
+++ b/src/core/radio/radio_platform.cpp
@@ -117,14 +117,9 @@ exit:
 
 bool otPlatRadioIsRxStatePolling(otInstance *aInstance)
 {
-    Instance *instance = static_cast<Instance *>(aInstance);
-    bool ret = false;
+    Instance &instance = *static_cast<Instance *>(aInstance);
 
-    if (instance->IsInitialized())
-    {
-        ret = instance->Get<Mac::Mac>().IsRxStatePolling();
-    }
-    return (ret);
+    return instance.Get<Mac::Mac>().IsRxStatePolling();
 }
 
 #if OPENTHREAD_CONFIG_DIAG_ENABLE
--- a/src/core/radio/radio_platform.cpp
+++ b/src/core/radio/radio_platform.cpp
@@ -36,6 +36,7 @@
 #include "common/code_utils.hpp"
 #include "common/instance.hpp"
 #include "radio/radio.hpp"
+#include "mac/mac.hpp"
 
 using namespace ot;
 
--- a/src/core/radio/radio_platform.cpp
+++ b/src/core/radio/radio_platform.cpp
@@ -34,9 +34,9 @@
 #include <openthread/platform/time.h>
 
 #include "common/code_utils.hpp"
+#include "mac/mac.hpp"
 #include "common/instance.hpp"
 #include "radio/radio.hpp"
-#include "mac/mac.hpp"
 
 using namespace ot;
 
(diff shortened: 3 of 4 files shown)


adabreuti:
@googlebot I signed it!

abtink:
@adabreuti thanks for submitting.

I am just wondering what is a use-case for this API?

adabreuti:
> @adabreuti thanks for submitting.
> 
> I am just wondering what is a use-case for this API?

@abtink This API is used in conjunction with the existing MAC state based OTLink API's to make run-time decisions at the radio layer for RF command prioritization.

abtink:
>> @adabreuti thanks for submitting.
>> 
>> I am just wondering what is a use-case for this API?
>
> @abtink This API is used in conjunction with the existing MAC state based OTLink API's to make run-time decisions at the radio layer for RF command prioritization.

If the goal is to use this at radio platform layer can we consider adding such mechansim/API as part of platform layer APIs e.g., in `otPlatRadio` APIs? This way we avoid cross-layer interactions.
- We have `otPlat` APIs which defines the interactions between platform layer (e.g., radio driver) and OT stack
- And then we have public OT APIs (like `otLink{}` or `otThread{}`) which are intended for user of OT stack.
- I think it'd be good to avoid platform layer using any public OT APIs (limit the platform interaction with OT to the platform APIs)

Can you please share what are some run-time decisions at radio layer that we want to enable with this API?
- Radio platform should already be aware that is being told to enter "rx" mode (through the call to `otPlatRadioReceive()`)
- Does it need to know the reason why it is being asked to "rx"? (e.g., is it CSL sample or rx after data poll with ack from parent indicating pending frame)? Is this the idea? 

Thanks.

adabreuti:
> > > @adabreuti thanks for submitting.
> > > I am just wondering what is a use-case for this API?
> > 
> > 
> > @abtink This API is used in conjunction with the existing MAC state based OTLink API's to make run-time decisions at the radio layer for RF command prioritization.
> 
> If the goal is to use this at radio platform layer, can we consider adding such mechansim/API as part of platform layer APIs e.g., in `otPlatRadio` APIs? This way we avoid cross-layer interactions.
> 
> * We have `otPlat` APIs which define the interactions between platform layer (e.g., radio driver) and OT stack
> * And then we have public OT APIs (like `otLink{}` or `otThread{}`) which are intended for user of OT stack.
> * I think it'd be good to avoid platform layer using any public OT APIs (limit the platform interaction with OT to the platform APIs)
> 
> Can you please share what are some run-time decisions at radio layer that we want to enable with this API?
> 
> * Radio platform should already be aware that is being told to enter "rx" mode (through the call to `otPlatRadioReceive()`)
> * Does it need to know the reason why it is being asked to "rx"? (e.g., is it CSL sample or rx after data poll with ack from parent indicating pending frame)? Is this the idea?
> 
> Thanks.

Thanks for the feedback, some more information regarding the use case:
 
We're tracking all of the Thread radio states, transmit polling, receive polling, direct vs indirect transmission etc. and using this information to prioritize radio commands in our multi-stack applications.
As such, we require the use of more specific data besides just being in the receive state as in the case of an FTD sitting in the Receive idle state would be less of a priority than actively receiving an ack following a transmission.

In terms of adding an otPlatRadio API, is your suggestion to update radio_platform.cpp/radio.h with a new otPlatRadioIsRxStatePolling() call which would call otLinkIsRxStatePolling()?

abtink:
> In terms of adding an otPlatRadio API, is your suggestion to update radio_platform.cpp/radio.h with a new otPlatRadioIsRxStatePolling() call which would call otLinkIsRxStatePolling()?

adabreuti:
> > In terms of adding an otPlatRadio API, is your suggestion to update radio_platform.cpp/radio.h with a new otPlatRadioIsRxStatePolling() call which would call otLinkIsRxStatePolling()?
> 
> Yes. Something similar. I would say we don't even add `otLinkIsRxStatePolling()` at all. We add a new platform API for this purpose in `radio.h` which gets the info from `Mac` class. This makes it clear that this function is intended for use by platform layer.
> 
> In `radio.h` we have APIs (function declerations) where we expect the platform layer to implement and we use from OT stack. There are also functions (marked as `extern` in `radio.h`) which OT stack provides for use by platform layer (they are typically callbacks but they can be functions like `otPlatRadioIsRxStatePolling()` allowing radio driver to ask for additional info).
> 
> Thoughts?

Understood, the PR has been updated to reflect your feedback. Let me know if you see any issues with the changes. Thanks!

src/core/radio/radio_platform.cpp
```diff
@@ -115,6 +116,13 @@ extern "C" void otPlatRadioEnergyScanDone(otInstance *aInstance, int8_t aEnergyS
     return;
 }
 
+bool otPlatRadioIsRxStatePolling(otInstance *aInstance)
```
abtink, line 119:
```suggestion
extern "C" bool otPlatRadioIsRxStatePolling(otInstance *aInstance)
```

src/core/mac/mac.hpp
```diff
@@ -550,6 +550,13 @@ class Mac : public InstanceLocator, private NonCopyable
      */
     bool IsInTransmitState(void) const;
 
+    /**
+     * This method returns if the MAC layer is in polling receive state.
+     *
+     * The MAC layer is in the receive state during a polling procedure pending external frame transmission.
+     */
+    bool IsRxStatePolling(void) const;
```
abtink, line 558: Inline?
```suggestion
    bool IsRxStatePolling(void) const { return (mOperation == kOperationWaitingForData); }
```

include/openthread/platform/radio.h
```diff
@@ -832,6 +832,16 @@ otError otPlatRadioEnergyScan(otInstance *aInstance, uint8_t aScanChannel, uint1
  */
 extern void otPlatRadioEnergyScanDone(otInstance *aInstance, int8_t aEnergyScanMaxRssi);
 
+/**
+ * This function indicates whether or not an IEEE 802.15.4 MAC is currently waiting for data.
+ * MAC module is in the receive state during a polling procedure pending external frame transmission.
+ *
+ * @param[in] aInstance A pointer to an OpenThread instance.
+ * @returns true if an IEEE 802.15.4 MAC is in the polling receive state, false otherwise.
```
abtink, line 840:
```suggestion
 * @param[in] aInstance A pointer to an OpenThread instance.
 
 * @returns true if an IEEE 802.15.4 MAC is in the polling receive state, false otherwise.
```

src/core/radio/radio_platform.cpp
```diff
@@ -115,6 +116,13 @@ extern "C" void otPlatRadioEnergyScanDone(otInstance *aInstance, int8_t aEnergyS
...
+{
+    Instance &instance = *static_cast<Instance *>(aInstance);
+
+    return instance.Get<Mac::Mac>().IsRxStatePolling();
+}
```
abtink, line 124: I am not sure, but I think we may need to exclude this under `OPENTHREAD_RADIO` or RCP build (since `Mac::Mac` won't be available under RCP).

abtink (approved):
LGTM. Thanks. 

Some smaller style suggestions below.

src/core/radio/radio_platform.cpp
adabreuti: Ah that's why there were build failures, the polling state should be available to all builds though. If implementing it within the radio_platform layer is a limitation for RCP builds, could we keep the API in the OTLink layer instead?

jwhui:
> We're tracking all of the Thread radio states, transmit polling, receive polling, direct vs indirect transmission etc. and using this information to prioritize radio commands in our multi-stack applications.
> As such, we require the use of more specific data besides just being in the receive state as in the case of an FTD sitting in the Receive idle state would be less of a priority than actively receiving an ack following a transmission.

@adabreuti , I believe the radio driver can determine whether OpenThread is waiting for data by observing that it just sent an IEEE 802.15.4 Data Request frame and received an ACK with the Frame Pending bit set. If so, can we avoid introducing a new API?

src/core/radio/radio_platform.cpp
abtink: Won't help to move it to `otLink`. RCP cannot see that either. `Mac::Mac` itself is not present on `RCP` build. It is running on host side.

adabreuti:
Thanks for the feedback, I'm closing PR as the implementation for determining Poll state is now in the radio layer. Thanks!

PR #6813, opened 2021-07-13
[meshcop] fix issues in sending updated datasets to leader

### LINKED ISSUE #6804: [GRL Test 9.2.7] Operational Dataset synchronization issue
Certification Test 9.2.7 fails if the Router node in the topology is running the OpenThread stack.

As per Thread 1.1.1 spec section 8.4.3.5.1  and 8.4.3.5.2,  when attaching to a Thread Network Partition, if the newly received Active/Pending Timestamp is older than what the Thread Device had before attaching, the Thread Device MUST attempt to update the Leader with its newer Active/Pending Operational Dataset. This behavior is not observed on the Router node. The Router (testbed) node fails to synchronize its (newer) active and pending operational datasets, in turn, causing the test case to fail.

It appears that there may also be a potential test plan bug, as the test plan (version R68) steps 5 and 11 instruct the harness to initiate this synchronization for the Router, which should not be the case, as per spec section 8.7.4.2 and  8.7.5.2 which states that:
A Thread Device that has a newer Active/Pending Operational Dataset, as determined by the Active/Pending Timestamp included in Thread Network Data, MUST provide its newer Active/Pending Operational Dataset to the Leader by sending a MGMT_ACTIVE/PENDING_SET.req to the Leader

Although the attached traces are for an older version of OpenThread, I have also observed this issue on a relatively newer version of the OpenThread stack

jwhui:
This commit fixes a couple issues with sending MGMT_SET.req to update
the leader with a new operational dataset.

- When handling an MLE Child ID Response message,
  DatasetManager::Save() is called before the role is updated to
  child. As a result, a device with a newer dataset prior to attaching
  the network will not send a MGMT_SET.req message. This commit adds a
  delay to allow the role to change to child.

- When receiving a MGMT_SET.rsp, the device will immediately check if
  it should send a MGMT_SET.req message again. However, the
  operational dataset may not have had a chance to propagate from the
  leader. This commit adds additional delay to allow the new
  operational dataset to propagate.

Fixes #6804

### DIFF
--- a/src/core/meshcop/dataset_manager.cpp
+++ b/src/core/meshcop/dataset_manager.cpp
@@ -167,8 +167,7 @@ Error DatasetManager::Save(const Dataset &aDataset)
     }
     else if (compare < 0)
     {
-        VerifyOrExit(!Get<Mle::MleRouter>().IsLeader(), error = kErrorInvalidState);
-        SendSet();
+        mTimer.Start(kSendSetDelay);
     }
 
     SignalDatasetChange();
@@ -279,7 +278,7 @@ void DatasetManager::SendSet(void)
     Dataset          dataset(GetType());
 
     VerifyOrExit(!mCoapPending, error = kErrorBusy);
-    VerifyOrExit(Get<Mle::MleRouter>().IsAttached(), error = kErrorInvalidState);
+    VerifyOrExit(Get<Mle::MleRouter>().IsChild() || Get<Mle::MleRouter>().IsRouter(), error = kErrorInvalidState);
     VerifyOrExit(mLocal.Compare(GetTimestamp()) < 0, error = kErrorInvalidState);
 
     if (IsActiveDataset())
@@ -323,7 +322,7 @@ exit:
         break;
 
     case kErrorNoBufs:
-        mTimer.Start(kDelayNoBufs);
+        mTimer.Start(kSendSetDelay);
         OT_FALL_THROUGH;
 
     default:
@@ -348,7 +347,7 @@ void DatasetManager::HandleCoapResponse(void *               aContext,
 void DatasetManager::HandleCoapResponse(void)
 {
     mCoapPending = false;
-    SendSet();
+    mTimer.Start(kSendSetDelay);
 }
 
 void DatasetManager::HandleGet(const Coap::Message &aMessage, const Ip6::MessageInfo &aMessageInfo) const
--- a/src/core/meshcop/dataset_manager.hpp
+++ b/src/core/meshcop/dataset_manager.hpp
@@ -363,7 +363,7 @@ private:
     enum
     {
         kMaxDatasetTlvs = 16,   // Maximum number of TLVs in a Dataset.
-        kDelayNoBufs    = 1000, // Milliseconds
+        kSendSetDelay   = 5000, // Milliseconds
     };
 
     bool       mCoapPending : 1;


abtink (approved):
LGTM. Thanks. 👍

wgtdkp (approved):
LGTM :+1:

simonlingoogle (approved):
LGTM :+1:

schinchani:
Hi @jwhui 

I tried running your changes locally, however, they do not seem to fix the issue. The operational datasets remain out of sync as reported earlier. To reproduce the issue, I tried the following test:

**Node 1:** 
Blacklist Node 2
Form network
Start commissioner
Set Active timestamp to 10
Stop commissioner

**Node 2:**
Blacklist Node 1
Configure a lower partition ID (0x1)
Form network with same network parameters as Node 1. 
Start commissioner
Set Active timestamp to 20, 
Set Pending Operational Dataset with Pending Timestamp to 30, Active Timestamp to 25, DelayTimer to 360000
Stop commissioner

**Now Clear filters on both Node 1 and Node 2**
Node 2 attaches to Node 1. However, both nodes continue to hold on to their earlier configured operational datasets (remaining out of sync). MGMT_SET.req/rsp messages are not observed.

jwhui:
@schinchani , can you provide the exact CLI commands you used in your test?

jwhui:
@schinchani , these are the steps I used to verify correct operation:
On node 1:
```
> dataset init new
Done
> dataset commit active
Done
> ifconfig up
Done
> thread start
Done
> dataset active -x
0e080000000000010000000300001335060004001fffe002087650f3b4ae25e0790708fd35ab3dbfca91d50510d4c397b5fcf766b86c2320d0e1117147030f4f70656e5468726561642d663865630102f8ec04102a959a5a930f4dbc9a8a57370e8ff5980c0402a0fff8
Done
```

On node 2:
```
> dataset set active 0e080000000000010000000300001335060004001fffe002087650f3b4ae25e0790708fd35ab3dbfca91d50510d4c397b5fcf766b86c2320d0e1117147030f4f70656e5468726561642d663865630102f8ec04102a959a5a930f4dbc9a8a57370e8ff5980c0402a0fff8
Done
> dataset init active
Done
> dataset activetimstamp 22
Done
> dataset commit active
Done
> ifconfig up
Done
> thread start
Done
```
Then on node 1, validate that the active timestamp has changed to 22:
```
> dataset active
Active Timestamp: 22
Channel: 19
Channel Mask: 0x07fff800
Ext PAN ID: 7650f3b4ae25e079
Mesh Local Prefix: fd35:ab3d:bfca:91d5::/64
Network Key: d4c397b5fcf766b86c2320d0e1117147
Network Name: OpenThread-f8ec
PAN ID: 0xf8ec
PSKc: 2a959a5a930f4dbc9a8a57370e8ff598
Security Policy: 672 onrcb
Done
```

schinchani:
> @schinchani , can you provide the exact CLI commands you used in your test?

Please see the attach console outputs:

[Node1.txt](https://github.com/openthread/openthread/files/6811466/Node1.txt)
[Node2.txt](https://github.com/openthread/openthread/files/6811467/Node2.txt)

jwhui:
> Please see the attach console outputs:
> 
> [Node1.txt](https://github.com/openthread/openthread/files/6811466/Node1.txt)
> [Node2.txt](https://github.com/openthread/openthread/files/6811467/Node2.txt)
> 
> These steps are similar to the test commands used by the Test Harness for 9.2.7

@schinchani , thanks for providing the console logs. I found the issue. If the device is a leader and migrating to a different partition, it would not schedule transmission of the MGMT_SET.req message. I have updated this PR to address it. Please review again.

schinchani:
The updated fix resolves the issue. I haven't tested it for 9.2.7 specifically, but I will keep an eye and report back if I observe any issues. Thanks

jwhui:
> The updated fix resolves the issue. I haven't tested it for 9.2.7 specifically, but I will keep an eye and report back if I observe any issues. Thanks

@schinchani , thanks for validating! If you are good with this PR, would you like to approve it?

schinchani (approved):
Sorry, missed approving it.

src/core/meshcop/dataset_manager.cpp
```diff
@@ -279,7 +278,7 @@ void DatasetManager::SendSet(void)
     Dataset          dataset(GetType());
 
     VerifyOrExit(!mCoapPending, error = kErrorBusy);
-    VerifyOrExit(Get<Mle::MleRouter>().IsAttached(), error = kErrorInvalidState);
+    VerifyOrExit(Get<Mle::MleRouter>().IsChild() || Get<Mle::MleRouter>().IsRouter(), error = kErrorInvalidState);
```
wgtdkp, line 281: @jwhui  Will this reject if the leader sends MGMT_SET.req to itself?

jwhui: Yeah, the idea is that the Leader is authoritative and calls to `SendSet()` should be ignored on the Leader. In what case do we need to call `SendSet()` on the Leader to update itself?

wgtdkp: Yeah, but it should be legal, right? It sounds more reasonable that the leader should be able to send requests to itself.

Thoughts?
@librasungirl @abtink

librasungirl: Possible use case - when the administrator want to commission the network to migrate to another network, there might be chance that the device that receives such command is a Leader.

Currently ot-br-posix will always sending pending dataset regardless of the device is leader in [thread_helper](https://github.com/openthread/ot-br-posix/blob/main/src/utils/thread_helper.cpp#L500)

```
void ThreadHelper::AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, ResultHandler aHandler)
{
....
    SuccessOrExit(error = otDatasetSendMgmtPendingSet(mInstance, &emptyDataset, datasetTlvs.mTlvs, datasetTlvs.mLength,
                                                      MgmtSetResponseHandler, this));
...
}
```

To solve it, either allowing leader sends mgmt to itself, or may updating `thread_helper` to set pending dataset directly on leader local.

//cc// @jinran-google @bukepo for vis.

jwhui: @wgtdkp @librasungirl , how does calling `otDatasetSendMgmtPendingSet()` lead to calling `SendSet()`?

librasungirl: @wgtdkp 's on device test log `Failed to send Dataset set to leader: InvalidState` seems it would somehow enter `SentSet()`,  I didn't carefully look into the calling path, and yes, you are right, it's `SendSetRequest()` that is called in `otDatasetSendMgmtPendingSet()`. 

 Will follow up with @wgtdkp on his setup and test procedure to see how the log occurs offline. Temporarily close this discussion now.

PR #6816, opened 2021-07-13
[dua] fix potential issue in DUA.rsp & DUA.ntf processing

simonlingoogle:


simonlingoogle:
@librasungirl @edmont any comments on this fix?

edmont:
@simonlingoogle any background about the fixed issue?

simonlingoogle:
> @simonlingoogle any background about the fixed issue?

There is a potential bug in `DuaManager::ProcessDuaResponse(Coap::Message &aMessage)`. 

The original code finds the Child by `Child *child = Get<ChildTable>().GetChildAtIndex(mChildIndexDuaRegistering);`, which assumes that the `DUA.rsp` must be targeting the current registering Child. This is true for `DUA.rsp`, but not correct for `DUA.ntf` because BBR may send `DUA.ntf` at any time to any device.

So, this comment changes `DuaManager::ProcessDuaResponse` code to search for the correct Child with the target DUA instead of using `mChildIndexDuaRegistering`. 

This commit also contains other refactoring on `mChildIndexDuaRegistering` usage.

edmont (approved):
Thanks, looks like a proper fix to me.

src/core/thread/dua_manager.cpp
```diff
@@ -645,27 +649,41 @@ Error DuaManager::ProcessDuaResponse(Coap::Message &aMessage)
 #endif
 #if OPENTHREAD_FTD && OPENTHREAD_CONFIG_TMF_PROXY_DUA_ENABLE
     {
-        Child *child = Get<ChildTable>().GetChildAtIndex(mChildIndexDuaRegistering);
+        Child *  child = nullptr;
+        uint16_t childIndex;
+
+        for (Child &iter : Get<ChildTable>().Iterate(Child::kInStateValidOrRestoring))
```
librasungirl, line 655: Only iterate valid child here? Note that the parent only register for valid child.
https://github.com/openthread/openthread/blob/3a953086730eaad12f4319f5608e42bbc333727d/src/core/thread/dua_manager.cpp#L479-L488

src/core/thread/dua_manager.cpp
```diff
@@ -645,27 +649,41 @@ Error DuaManager::ProcessDuaResponse(Coap::Message &aMessage)
...
-        VerifyOrExit(child->HasIp6Address(target), error = kErrorNotFound);
+
+        childIndex = Get<ChildTable>().GetChildIndex(*child);
 
         switch (status)
         {
         case ThreadStatusTlv::kDuaSuccess:
             // Mark as Registered
-            mChildDuaRegisteredMask.Set(mChildIndexDuaRegistering, true);
+            if (mChildDuaMask.Get(childIndex))
```
librasungirl, line 672: Is this ` if` condition necessary?

src/core/thread/dua_manager.cpp
simonlingoogle: resolved.

librasungirl (approved):

jwhui (approved):

PR #6823, opened 2021-07-16
[dns-server] add DNS64 recursive query server support

gjc13:
The DNS64 server allows devices in the Thread network to visit IPv4
hostnames transparently via the 6-to-4 border router.

Depends-On: https://github.com/openthread/ot-br-posix/pull/932

### DIFF
--- a/include/openthread/dns.h
+++ b/include/openthread/dns.h
@@ -39,6 +39,7 @@
 
 #include <openthread/error.h>
 #include <openthread/instance.h>
+#include <openthread/ip6.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -166,6 +167,19 @@ void otDnsSetNameCompressionEnabled(bool aEnabled);
  *
  */
 bool otDnsIsNameCompressionEnabled(void);
+
+/**
+ * This function gets the NAT64 prefix.
+ *
+ * The resolved addresses for IPv4 hosts will begin with this prefix.
+ *
+ * @param[in]  aInstance    The OpenThread instance.
+ * @param[out] aPrefix      The NAT64 prefix.
+ *
+ * @retval OT_ERROR_NONE        The address is successfully fetched.
+ * @retval OT_ERROR_NOT_FOUND   There is no NAT64 prefix in the network.
+ */
+otError otDnsGetNat64Prefix(otInstance *aInstance, otIp6Prefix *aPrefix);
 
 /**
  * @}
--- a/include/openthread/instance.h
+++ b/include/openthread/instance.h
@@ -53,7 +53,7 @@ extern "C" {
  * @note This number versions both OpenThread platform and user APIs.
  *
  */
-#define OPENTHREAD_API_VERSION (138)
+#define OPENTHREAD_API_VERSION (139)
 
 /**
  * @addtogroup api-instance
--- a/src/core/api/dns_api.cpp
+++ b/src/core/api/dns_api.cpp
@@ -39,6 +39,7 @@
 #include "common/locator_getters.hpp"
 #include "net/dns_client.hpp"
 #include "net/dns_types.hpp"
+#include "openthread/instance.h"
 
 using namespace ot;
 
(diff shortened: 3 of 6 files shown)


abtink:
@gjc, can you please share some background on this PR?  

My understanding is that we won't be using DNS64 in Thread. 

The preferred way (from Thread TC discussion) has been for DNS client (i.e. device itself) to perform the v4 to v6 address conversion instead of doing this on server/resolver. The mechanism is for DNS client to query for A records if it gets no response for AAAA records, and then perform address synthesize on the client. This is what we have already implemented in OT, e.g., from PRs:
- https://github.com/openthread/openthread/pull/6414
- https://github.com/openthread/openthread/pull/6429
- https://github.com/openthread/openthread/pull/6783

gjc13:
> @gjc, can you please share some background on this PR?
> 
> My understanding is that we won't be using DNS64 in Thread.
> 
> The preferred way (from Thread TC discussion) has been for DNS client (i.e. device itself) to perform the v4 to v6 address conversion instead of doing this on server/resolver. The mechanism is for DNS client to query for A records if it gets no response for AAAA records, and then perform address synthesize on the client. This is what we have already implemented in OT, e.g., from PRs:
> 
> * [[dns-client] add support for NAT64 during address resolution #6414](https://github.com/openthread/openthread/pull/6414)
> * [[network-data] add support for NAT64 flag in HasRoute TLV #6429](https://github.com/openthread/openthread/pull/6429)
> * [[dns-client] use NAT64 prefix from network data #6783](https://github.com/openthread/openthread/pull/6783)

We've encountered several difficluties when using DNS client on the end device to translate IPv4 addresses.

### Compatibility with discovery delegate

OpenThread is now providing the discovery delegate feature on the 53 port. Using another DNS server for internet name resolution requires the device to shift between different DNS servers. In the multi-task context this will hard to manage.

### Avaiblity of global IPv6 connectivity

There can be three options for the DNS server used:
- A global IPv6 DNS server on the Internet such as Google DNS
- The NAT64 address of a global IPv4 DNS server
- The border router itself as a recursive server

#### Using global IPv6 DNS

This requires global IPv6 connectivity. The global IPv6 connectivity is only available by using a DHCPv6-PD allocated address as the routable prefix. The ISP network and the user's WiFI router doesn't necessarily support prefix delegation and it's not yet implemented in OpenThread. As a result, global IPv6 DNS is always available.

#### Using NAT64 address of global IPv4 DNS

This solution provides a stronger guarantee for the connectivity to the DNS server. In practice we can always connect to the DNS server as long as the border router is connected to the internet. However, since the client prefers AAAA records to  A records, it will always connect to the IPv6 address when possible. As we've discussed before, the global IPv6 connectivity is still hard to guarantee.

#### Using border router as the recursive server

Common recursive server packages such as `bind9` will occupy the same port as the discovery delegate. As a result, we need to implement the recursive server inside OpenThread. This will lead to a solution quite similar to this PR except that it will directly forward the AAAA records. I'll explain why doing the DNS64 translation is the desired behavior in the next section.

### Difficulty in application layer porting

Application layer code tends to use `gethostbyname` and `getaddrinfo` for DNS queries. These functions will use the system DNS libraries (the kernel or lwip) rather than the OpenThread DNS client. Thus they cannot get the NAR64 translated addresses unless we modify the application layer code. Doing so is an extra burden both on code effort and complexity. We'd like OpenThread to be transparent to application layer and have them use the common POSIX apis.

Providing the translated AAAA record from the border router, on the other hand, eliminates these difficulties. The DNS server is always reachable since it's on the Thread network. The provided records can be used by `gethostbyname` directly and the address is guaranteed to be connectable.

Maybe we should reconsider how we'd like to provide DNS records for the global names. Thoughts? @abtink @simonlingoogle @librasungirl

jwhui:
@gjc13 , these design considerations were discussed at length in the Thread Group. Making these changes would require changes to the draft specification.

> We've encountered several difficluties when using DNS client on the end device to translate IPv4 addresses.
> 
> ### Compatibility with discovery delegate
> OpenThread is now providing the discovery delegate feature on the 53 port. Using another DNS server for internet name resolution requires the device to shift between different DNS servers. In the multi-task context this will hard to manage.

This issue was discussed and there was consensus that forwarding DNS messages between SRP server and a DNS server was feasible. dnsmasq and bind9 support DNS message forwarding.

> ### Avaiblity of global IPv6 connectivity
> There can be three options for the DNS server used:
> 
> * A global IPv6 DNS server on the Internet such as Google DNS
> * The NAT64 address of a global IPv4 DNS server
> * The border router itself as a recursive server

Current draft specification requires the border router to implement DNS recursive resolver.

> ### Difficulty in application layer porting
> Application layer code tends to use `gethostbyname` and `getaddrinfo` for DNS queries. These functions will use the system DNS libraries (the kernel or lwip) rather than the OpenThread DNS client. Thus they cannot get the NAT64 translated addresses unless we modify the application layer code. Doing so is an extra burden both on code effort and complexity. We'd like OpenThread to be transparent to application layer and have them use the common POSIX apis.
> 
> Providing the translated AAAA record from the border router, on the other hand, eliminates these difficulties. The DNS server is always reachable since it's on the Thread network. The provided records can be used by `gethostbyname` directly and the address is guaranteed to be connectable.
> 
> Maybe we should reconsider how we'd like to provide DNS records for the global names. Thoughts?

There were a number of reasons why we chose to require IPv6 address synthesis on the client. Some of those reasons are stated in [RFC 6147 Section 5.5](https://datatracker.ietf.org/doc/html/rfc6147#section-5.5) and [RFC 7050 Section 1](https://datatracker.ietf.org/doc/html/rfc7050).

While we acknowledged that there may be existing implementations that would require updating, there was consensus that the benefits of doing so outweighs the costs.

PR #6824, opened 2021-07-16
[mac] replace `enum` constants with `constexpr`

abtink:
This commit replaces the `enum` constants with `constexpr` definitions
in all the modules under `core/mac`. The `constexpr` constants have
the advantage that thet are explicitly typed while still being
compile-time constants with no code size overhead.

This commit also contains some smaller changes: typo fixes in the
comments, removal of unused constants, and adding type to the named
`enum` definitions.

---

We originally used `enum` for constants due to OT complying to 
`C++03`  which we transitioned to `C++11` from PR https://github.com/openthread/openthread/pull/5077.

### DIFF
--- a/src/core/mac/data_poll_handler.hpp
+++ b/src/core/mac/data_poll_handler.hpp
@@ -68,16 +68,13 @@ class DataPollHandler : public InstanceLocator, private NonCopyable
     friend class Mac::Mac;
 
 public:
-    enum
-    {
-        kMaxPollTriggeredTxAttempts = OPENTHREAD_CONFIG_MAC_MAX_TX_ATTEMPTS_INDIRECT_POLLS,
-    };
+    static constexpr uint8_t kMaxPollTriggeredTxAttempts = OPENTHREAD_CONFIG_MAC_MAX_TX_ATTEMPTS_INDIRECT_POLLS;
 
     /**
      * This enumeration defines frame change request types used as input to `RequestFrameChange()`.
      *
      */
-    enum FrameChange
+    enum FrameChange : uint8_t
     {
         kPurgeFrame,   ///< Indicates that previous frame should be purged. Any ongoing indirect tx should be aborted.
         kReplaceFrame, ///< Indicates that previous frame needs to be replaced with a new higher priority one.
--- a/src/core/mac/data_poll_sender.hpp
+++ b/src/core/mac/data_poll_sender.hpp
@@ -63,12 +63,9 @@ namespace ot {
 class DataPollSender : public InstanceLocator, private NonCopyable
 {
 public:
-    enum
-    {
-        kDefaultFastPolls  = 8,  ///< Default number of fast poll transmissions (@sa StartFastPolls).
-        kMaxFastPolls      = 15, ///< Maximum number of fast poll transmissions allowed.
-        kMaxFastPollsUsers = 63, ///< Maximum number of the users of fast poll transmissions allowed.
-    };
+    static constexpr uint8_t kDefaultFastPolls  = 8;  ///< Default number of fast poll tx (@sa StartFastPolls).
+    static constexpr uint8_t kMaxFastPolls      = 15; ///< Maximum number of fast poll tx allowed.
+    static constexpr uint8_t kMaxFastPollsUsers = 63; ///< Maximum number of the users of fast poll tx allowed.
 
     /**
      * This constructor initializes the data poll sender object.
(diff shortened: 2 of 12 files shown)


src/core/mac/sub_mac.hpp
```diff
@@ -623,7 +609,7 @@ class SubMac : public InstanceLocator, private NonCopyable
 #if OPENTHREAD_CONFIG_PLATFORM_USEC_TIMER_ENABLE
     TimerMicro mTimer;
 #else
-    TimerMilli mTimer;
+    TimerMilli                mTimer;
```
abtink, line 612: `clang-format` is changing this. I have no idea why. I guess it is trying to align it with something?

src/core/mac/data_poll_handler.hpp
```diff
@@ -68,16 +68,13 @@ class DataPollHandler : public InstanceLocator, private NonCopyable
...
-        kMaxPollTriggeredTxAttempts = OPENTHREAD_CONFIG_MAC_MAX_TX_ATTEMPTS_INDIRECT_POLLS,
-    };
+    static constexpr uint8_t kMaxPollTriggeredTxAttempts = OPENTHREAD_CONFIG_MAC_MAX_TX_ATTEMPTS_INDIRECT_POLLS;
 
     /**
      * This enumeration defines frame change request types used as input to `RequestFrameChange()`.
      *
      */
-    enum FrameChange
+    enum FrameChange : uint8_t
```
wgtdkp, line 77:
```suggestion
    enum class FrameChange : uint8_t
```

src/core/mac/data_poll_sender.hpp
```diff
@@ -259,17 +256,11 @@ class DataPollSender : public InstanceLocator, private NonCopyable
...
-        kMaxCslPollRetxAttempts =
-            OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_DIRECT, ///< Maximum number of retransmit attempts of data
-                                                                    ///< poll with CSL IE (mac data request).
-    };
+    static constexpr uint8_t kQuickPollsAfterTimeout = 5; // Quick data poll tx in case of back-to-back poll timeouts.
+    static constexpr uint8_t kMaxPollRetxAttempts    = OPENTHREAD_CONFIG_FAILED_CHILD_TRANSMISSIONS;
+    static constexpr uint8_t kMaxCslPollRetxAttempts = OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_DIRECT;
 
-    enum PollPeriodSelector
+    enum PollPeriodSelector : uint8_t
```
wgtdkp, line 263:
```suggestion
    enum class PollPeriodSelector : uint8_t
```

wgtdkp (approved):
LGTM :+1: 
Should we prefer [Scoped Enumerations](https://en.cppreference.com/w/cpp/language/enum) (`enum class`)?

abtink:
> Should we prefer Scoped Enumerations (enum class)?

I think this would be separate thing to consider (a different PR where we change all such definitions if we agree to). 

I myself prefer normal `enum` use over `enum class`:

- `enum` support by toolchains today provides (almost) the same level of safety and build-time checks compared to `enum class`.
- Style-wise with `enum class` to use the value we need the name of the `enum`
   - e.g., `DataPollHandler::FrameChange::kPurgeFrame` over `DataPollHandler::kPurgeFram`. I personally prefer the shorter version (less verbose).
- From what I recall, some (older) embedded toolchains did not optimize `enum class` as well as `enum` definitions (e.g., passed them to methods/functions as objects (pointers/references) instead of simple `uint` value which can lead to unnecessary run-time and code size overheads).

wgtdkp:
> * `enum` support by toolchains today provides (almost) the same level of safety and build-time checks compared to `enum class`.

`enum class` disallows implicit conversion to integers. This avoids subtle bug/typo like:
```c++
if (frameChange = kPurgeFrame) { // There is a typo, we actually want "frameChange == kPurgeFrame"
...
}
```
The code can compile with `enum` but will not with `enum class`.


> * Style-wise with `enum class` to use the value we need the name of the `enum`
>   
>   * e.g., `DataPollHandler::FrameChange::kPurgeFrame` over `DataPollHandler::kPurgeFram`. I personally prefer the shorter version (less verbose).

Actually, I will remove the `Frame` suffix from all enum constants if it is defined with `enum class`, and they should be accessed as `DataPollHandler::FrameChange::kPurge`. The downside of `DataPollHandler::kPurgeFrame` is that, the reader cannot get the information that whether this constant is a Frame Change type enumeration.

> * From what I recall, some (older) embedded toolchains did not optimize `enum class` as well as `enum` definitions (e.g., passed them to methods/functions as objects (pointers/references) instead of simple `uint` value which can lead to unnecessary run-time and code size overheads).

AFAIK, `enum class` by definition is implemented with integers specified with `: <integer-type>` and I will not expect a compile should create an different intermediate/underlying type for an `enum class` - it should be just be an integer after being compiled to assembly code. So I don't think such toolchain implementation is valid and conforms to the standard, and thus we may just ignore them. Especially when it will not cause a compiling error or runtime bug.

Thoughts?

abtink:
@wgtdkp. this is a good discussion, but I will say we consider this as a separate thing from this PR. 

This PR is to replace unnamed enum constants with `constexpr` (I will have other PRs to do the sam in other modules). 

On the `enum class` the main benefit is that they don't auto-convert to/from ints which is good but not critical (does not often used).

The case of something like `if (frameChange = kPurgeFrame)` is typically caught by LGTM check and/or toolchain warnings (they do such a check independent of whether the underlying variable is enum class` or not).

On naming, yes we can rename the constants, I would say it is matter of style and personal taste. I persoanlly like `kOperationIdle` over `Operation::kIdle`.

src/core/mac/mac_types.hpp
```diff
@@ -424,10 +416,7 @@ OT_TOOL_PACKED_BEGIN
 class Key : public otMacKey, public Equatable<Key>, public Clearable<Key>
 {
 public:
-    enum
-    {
-        kSize = OT_MAC_KEY_SIZE, // Key size in bytes.
-    };
+    static constexpr uint16_t kSize = OT_MAC_KEY_SIZE; // Key size in bytes.
```
bukepo, line 419: Is it intentional not using `//<`? What practice are we going to take for `//` and `//<`?

bukepo (approved):
Thanks!

Just a question about using `//<` and `//`.

src/core/mac/mac_types.hpp
abtink: No. Not intentional. I probably just did not notice it and left it as `//` which was before. I will change it in next push. Thanks.

In general, here is what I would recommend about this:
- We use `///<` only for `public` definitions (we can also skip it if the constant name is descriptive enough and comment does not add much value.). 
- We use `///<` only when the entire description fits in same line, otherwise it's better to have a comment before the definition using `/** */` style.
- For `private` definitions, adding comment/docs is fully optional (up to developer), but we should not use `///<` so it won't be added in doxygen (or to other tools that collect such docs) since the definitions is private.

jwhui (approved):

PR #6826, opened 2021-07-18
[srp-server] support address modes & sharing of socket with DNS-SD

abtink:
This commit updates `Srp::Server` to add support for selecting its
address mode (unicast or anycast). The address mode specifies how the
SRP server determines its address and port number and how this info
is published in Thread Network Data.

In anycast address mode, the SRP server will use port number 53 which
is also used by DNS resolver/server. So if both `Srp::Server` and
`Dns::ServiceDiscovery::Server` are enabled on a device, they both
need to listen on the same port number and share the same UDP socket
instance. This commit adds the mechanism in the code to realize this.
The `Srp::Server` will monitor when/if DNS-SD is active and checks
whether the port numbers match and if so it will close its own socket
and re-use the one from DNS-SD. In this case, an incoming UDP message
is first given to `Srp::Server` to process and if it fails to process
it then the message is given to DNS-SD.

This commit also adds CLI commands under `srp server` to get/set the
address mode (and update the `README_SRP_SERVER.md` documentation).
It also adds a new test `test_srp_server_anycast_mode` which covers
the SRP server address mode and the newly added behaviors.


include/openthread/srp_server.h
```diff
@@ -194,6 +207,52 @@ otError otSrpServerSetDomain(otInstance *aInstance, const char *aDomain);
...
+uint8_t otSrpServerGetAnycastModeSequenceNumber(otInstance *aInstance);
+
+/**
+ * This function sets the sequence number used with anycast address mode.
+ *
+ * @param[in] aInstance        A pointer to an OpenThread instance.
+ * @param[in] aSequenceNumber  The sequence number to use.
+ *
+ * @retval OT_ERROR_NONE            Successfully set the address mode.
+ * @retval OT_ERROR_INVALID_STATE   The SRP server is enabled and the sequence number cannot be changed.
```
wgtdkp, line 251: Will this error code be returned if the SRP server is in mode `OT_SRP_SREVER_ADDRESS_MODE_UNICAST`?

include/openthread/srp_server.h
```diff
@@ -194,6 +207,52 @@ otError otSrpServerSetDomain(otInstance *aInstance, const char *aDomain);
...
+otError otSrpServerSetAddressMode(otInstance *aInstance, otSrpServerAddressMode aMode);
+
+/**
+ * This function returns the sequence number used with anycast address mode.
+ *
+ * The sequence number is included in "DNS/SRP Service Anycast Address" entry published in the Network Data.
+ *
+ * @param[in] aInstance  A pointer to an OpenThread instance.
+ *
+ * @returns The anycast sequence number.
```
wgtdkp, line 239: What's the return value if the SRP server is in mode `OT_SRP_SREVER_ADDRESS_MODE_UNICAST`?

src/core/net/srp_server.cpp
```diff
@@ -441,25 +477,96 @@ void Server::SelectPort(void)
 
 void Server::Start(void)
 {
-    Error error = kErrorNone;
-
     VerifyOrExit(mState == kStateStopped);
 
     mState = kStateRunning;
+    PrepareSocket();
+    otLogInfoSrp("[server] start listening on port %u", mPort);
```
wgtdkp, line 484: `Srp::Server::Stop`  needs to be updated accordingly.

src/core/net/dnssd_server.cpp
```diff
@@ -107,13 +115,21 @@ void Server::HandleUdpReceive(void *aContext, otMessage *aMessage, const otMessa
...
     Header requestHeader;
 
-    SuccessOrExit(error = aMessage.Read(aMessage.GetOffset(), requestHeader));
-    VerifyOrExit(requestHeader.GetType() == Header::kTypeQuery, error = kErrorDrop);
+#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
+    // We first let the `Srp::Server` process the received message.
+    // It returns `kErrorNone` to indicate that it successfully
+    // processed the message.
+
+    VerifyOrExit(Get<Srp::Server>().HandleDnssdServerUdpReceive(aMessage, aMessageInfo) != kErrorNone);
```
wgtdkp, line 125: This may result in sending multiple responses for a single DNS query: If this message is a SRP message but the key and signature doesn't match, the SRP server will send a Security error response and here the DNS-SD server will also send a `NotImplemented` response as it is not a standard DNS query.

The `Srp::Server` returns `kErrorDrop` and silently drop this message if it is not a DNS UPDATE message, I think we should try DNS-SD server message handler only if `Get<Srp::Server>().HandleDnssdServerUdpReceive(...)` returns `kErrorDrop`, thoughts?

tests/scripts/thread-cert/test_srp_server_anycast_mode.py
```diff
@@ -0,0 +1,200 @@
...
+#   published in the Thread Network Data. When using anycast address
+#   mode, the SRP server and DNS server/resolver will both listen on port
+#   53 and they re-use the same socket. This test verifies the behavior of
+#   both modules in such a situation.
+#
+#
+# Topology:
+#
+#   One leader and two router nodes, all connected. The leader acts as SRP
+#   server and DNS resolver. One router acts at SRP client, and the
```
wgtdkp, line 49:
```suggestion
#   server and DNS resolver. One router acts as SRP client, and the
```

include/openthread/srp_server.h
abtink: No. It will be accepted fine. The idea is as long as the server is not enabled, the caller can call the APIs to change address mode or its related parameters (like changing the anycast mode seq number) as many times and/or in whatever order.

include/openthread/srp_server.h
abtink: See above.

src/core/net/srp_server.cpp
abtink: @wgtdkp I am not sure if I follow your suggestion here. Is it related to the log?

We already have this log line in the `Stop()`:
```c++
void Server::Stop(void)
{
    VerifyOrExit(mState == kStateRunning);

    [...]

    otLogInfoSrp("[server] stop listening on %u", mPort);
    IgnoreError(mSocket.Close());
    mHasRegisteredAnyService = false;

exit:
    return;
}
```

src/core/net/dnssd_server.cpp
abtink: > If this message is a SRP message but the key and signature doesn't match, the SRP server will send a Security error response and here the DNS-SD server will also send a NotImplemented response as it is not a standard DNS query.

@wgtdkp I don't think this can happen in the current code. Here is the `Srp::Server::ProcessMessage()`

```c++
Error Server::ProcessMessage(Message &aMessage, const Ip6::MessageInfo &aMessageInfo)
{
    Error             error;
    Dns::UpdateHeader dnsHeader;
    uint16_t          offset = aMessage.GetOffset();

    SuccessOrExit(error = aMessage.Read(offset, dnsHeader));
    offset += sizeof(dnsHeader);

    // Handles only queries.
    VerifyOrExit(dnsHeader.GetType() == Dns::UpdateHeader::Type::kTypeQuery, error = kErrorDrop);

    switch (dnsHeader.GetQueryType())
    {
    case Dns::UpdateHeader::kQueryTypeUpdate:
        ProcessDnsUpdate(aMessage, aMessageInfo, dnsHeader, offset);
        break;
    default:
        error = kErrorDrop;
        break;
    }

exit:
    return error;
}
```
If the DNS message is an query "update" then only SRP server will process it (note that no error from `ProcessDnsUpdate()` method call).

bukepo (approved):

src/core/net/dnssd_server.cpp
wgtdkp: You are right, I missed that `ProcessDnsUpdate` doesn't propagate the error back.

src/core/net/srp_server.cpp
wgtdkp: I meant we may want to close the socket depending on whether it is sharing the DNS-SD socket. But I think it is always okay to just close `mSocket` in `::Srp::Server::Stop`.

wgtdkp (approved):
LGTM 👍 
Thanks @abtink !

src/core/net/dnssd_server.cpp
```diff
@@ -107,13 +115,21 @@ void Server::HandleUdpReceive(void *aContext, otMessage *aMessage, const otMessa
...
 void Server::HandleUdpReceive(Message &aMessage, const Ip6::MessageInfo &aMessageInfo)
 {
-    Error  error = kErrorNone;
     Header requestHeader;
 
-    SuccessOrExit(error = aMessage.Read(aMessage.GetOffset(), requestHeader));
-    VerifyOrExit(requestHeader.GetType() == Header::kTypeQuery, error = kErrorDrop);
+#if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
+    // We first let the `Srp::Server` process the received message.
+    // It returns `kErrorNone` to indicate that it successfully
```
simonlingoogle, line 122: I would prefer to dispatch UDP messages according to the DNS Header:
- Dispatch to SRP server if DNS type is kTypeQuery and Query Type is kQueryTypeUpdate
- DNS-SD server process other messages

I think it's not safe to rely on errors.
Thoughts? @abtink @wgtdkp

wgtdkp: If you look into the implementation of `HandleDnssdServerUdpReceive`, it's exactly dispatching the messages with DNS type.

src/core/net/srp_server.cpp
abtink: Yes. We can close an already closed socket again which then does nothing.

src/core/net/dnssd_server.cpp
abtink: We need the dispatching/filtering in `Srp::Server` independent of whether we share the socket or use our own socket. Even if we use our own socket we still may receive other type of messages on the socket and need to filter (accept DNS type query `kQueryTypeUpdate` only).

We can add the dispatching/filtering code in `HandleDnssdServerUdpReceive()` but then we would need to repeat the same code when we use our own socket.

So it make sense to have a common place where we do this (i.e., in `ProcessMessage()`) and use it from both callbacks (whether we receive from our own socket `HandleUdpReceive()` or from shared socket `HandleDnssdServerUdpReceive()`).

src/core/net/srp_server.cpp
```diff
@@ -441,25 +477,96 @@ void Server::SelectPort(void)
...
+
+    VerifyOrExit(!mSocket.IsOpen());
+    SuccessOrExit(error = mSocket.Open(HandleUdpReceive, this));
+    error = mSocket.Bind(mPort, OT_NETIF_THREAD);
 
 exit:
     if (error != kErrorNone)
     {
-        otLogCritSrp("[server] failed to start: %s", ErrorToString(error));
+        otLogCritSrp("[server] failed to prepare socket: %s", ErrorToString(error));
```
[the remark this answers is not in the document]
abtink: We already call `Stop()` here on `error` which would take care of closing the socket.

src/core/net/dnssd_server.cpp
simonlingoogle: SGTM

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6827, opened 2021-07-18
[cli] new config to log input command and the resulting output

abtink:
This commit adds `OPENTHREAD_CONFIG_CLI_LOG_INPUT_OUTPUT_ENABLE`
option which configures the CLI module to log its input command
string and the resulting output. This is done in addition to emitting
the output string to the console. By default this config is enabled
on all POSIX based platforms and when `LOG_OUTPUT` is not set to use
CLI console itself.

--------

Notes:
- This should hopefully help with debugging test failures (e.g., on github 
workflows).
- Suggested this model from https://github.com/openthread/openthread/issues/6665#issuecomment-855443633
(related to https://github.com/openthread/openthread/pull/6639 which caused
the the CLI output to be logged in pieces).
- The implementation in this PR emits the CLI output line by line to the log 
(each output line is emitted as a separate log line, so the logs mirror what 
is seen on the console output).
```
> netdata show
Prefixes:
2001:dead:beef:cafe::/64 paros med 2800
Routes:
Services:
44970 112233 aabbcc s 2800
Done

The related logs:

[NOTE]-CLI-----: Input: netdata show
[NOTE]-CLI-----: Output: Prefixes:
[NOTE]-CLI-----: Output: 2001:dead:beef:cafe::/64 paros med 2800
[NOTE]-CLI-----: Output: Routes:
[NOTE]-CLI-----: Output: Services:
[NOTE]-CLI-----: Output: 44970 112233 aabbcc s 2800
[NOTE]-CLI-----: Output: Done
```
- By default this config is not enabled when `LOG_OUTPUT` is using CLI 
itself (since such a log would be redundant). However, the implementation
works correctly under such a config. Here is an example when this is 
enabled:
```
> prefix add 2001:dead:beef:cafe::/64 paros med
[NOTE]-CLI-----: Input: prefix add 2001:dead:beef:cafe::/64 paros med
Done
[NOTE]-CLI-----: Output: Done
```

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -82,6 +82,7 @@
 #include <openthread/platform/debug_uart.h>
 #endif
 
+#include "common/logging.hpp"
 #include "common/new.hpp"
 #include "common/string.hpp"
 #include "mac/channel_mask.hpp"
@@ -126,6 +127,10 @@ Interpreter::Interpreter(Instance *aInstance, otCliOutputCallback aCallback, voi
 #endif
 #if OPENTHREAD_CONFIG_SRP_SERVER_ENABLE
     , mSrpServer(*this)
+#endif
+#if OPENTHREAD_CONFIG_CLI_LOG_INPUT_OUTPUT_ENABLE
+    , mOutputLength(0)
+    , mIsLogging(false)
 #endif
 {
 #if OPENTHREAD_FTD
@@ -4551,6 +4556,10 @@ void Interpreter::ProcessLine(char *aBuf)
 
     VerifyOrExit(StringLength(aBuf, kMaxLineLength) <= kMaxLineLength - 1, error = OT_ERROR_PARSE);
 
+#if OPENTHREAD_CONFIG_CLI_LOG_INPUT_OUTPUT_ENABLE
+    otLogNoteCli("Input: %s", aBuf);
+#endif
+
     error = Utils::CmdLineParser::ParseCmd(aBuf, args);
 
     if (error != OT_ERROR_NONE)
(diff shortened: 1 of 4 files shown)


src/cli/cli.cpp
```diff
@@ -4958,8 +5058,20 @@ extern "C" void otCliPlatLogv(otLogLevel aLogLevel, otLogRegion aLogRegion, cons
 
     VerifyOrExit(Interpreter::IsInitialized());
 
+#if OPENTHREAD_CONFIG_CLI_LOG_INPUT_OUTPUT_ENABLE
+    // CLI output can be used for logging. The `mIsLogging` flag is
+    // used to indicate whether it is being used for a CLI command
+    // output or for logging.
+    Interpreter::GetInterpreter().mIsLogging = true;
```
simonlingoogle, line 5065: nit: Better define a method than assigning `mIsLogging` direclty?

src/cli/cli_config.h
```diff
@@ -56,4 +56,30 @@
...
+/**
+ * @def OPENTHREAD_CONFIG_CLI_LOG_INPUT_OUTPUT_LOG_STRING_SIZE
+ *
+ * The log string buffer size (in bytes).
+ *
+ * This is only used when `OPENTHREAD_CONFIG_CLI_LOG_INPUT_OUTPUT_ENABLE` is enabled.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_CLI_LOG_INPUT_OUTPUT_LOG_STRING_SIZE
+#define OPENTHREAD_CONFIG_CLI_LOG_INPUT_OUTPUT_LOG_STRING_SIZE 200
```
simonlingoogle, line 82: Why 200 ? Why not use `OPENTHREAD_CONFIG_CLI_MAX_LINE_LENGTH` ?

simonlingoogle (approved):
LGTM  :+1: 

with nits

src/cli/cli_config.h
abtink: Changed in new push.

src/cli/cli.cpp
```diff
@@ -4918,7 +4927,98 @@ void Interpreter::OutputSpaces(uint8_t aCount)
...
+        //
+        // We move (newLen + 1 = 5) chars from `lineEnd` to start of
+        // `mOutputString` which will include the `\0` char.
+        //
+        // If `lineEnd` and `mOutputString[mOutputLength]` are the same
+        // the code works correctly as well  (new length set to zero and
+        // the `\0` is copied).
+
+        mOutputLength = static_cast<uint16_t>(&mOutputString[mOutputLength] - lineEnd);
+        memmove(mOutputString, lineEnd, mOutputLength + 1);
```
jwhui, line 5007: Is it better to avoid the `memmove` and try to leave the log strings in place?

abtink: It's possible to avoid `memmove()` but not sure if it is worth it. 
- We need to track an `startIndex` (how much of the `mOutputString` was previously used and emitted to logs). 
- Such a model would also assume that the CLI outputs will always end at some point with `\n\r` with nothing after it (which may be the case in way we generate output today in the code but I think may be better if we don't rely on it).
- The `memmove()` model allow us to use the entire `mOutputString` buffer for every CLI output line. 
-  Also practically very frequently, `memove()` copies just one byte the `\0` (since most often the `\n\r` is added on its own or at end of single string from an `Output()` method call.

simonlingoogle: Noticed `memmove()`, but I think it's harmless as @abtink described.

jwhui (approved):

PR #6829, opened 2021-07-19
[config] include the default user configuration header

zhanglongxia:
OpenThread uses the compile option
`-DOPENTHREAD_PROJECT_CORE_CONFIG_FILE=openthread-core-posix-config.h`
to specify the OpenThread configuration file. When the Blaze compiles
the OpenThread, the include scanner in Blaze does not understand the
cases where the header file to be scanned comes from a preprocessor
macro like this. Then the compilation fails.

This CL includes the default user implemented configuration header file
to avoid specifying the configuration file.

### DIFF
--- a/src/core/openthread-core-config.h
+++ b/src/core/openthread-core-config.h
@@ -44,6 +44,10 @@
 
 #ifdef OPENTHREAD_PROJECT_CORE_CONFIG_FILE
 #include OPENTHREAD_PROJECT_CORE_CONFIG_FILE
+#elif defined(OPENTHREAD_CONFIG_CORE_USER_CONFIG_HEADER_ENABLE)
+// This configuration header file should be implemented by the user when the
+// OPENTHREAD_CONFIG_CORE_USER_CONFIG_HEADER_ENABLE is defined to 1.
+#include "openthread-core-user-config.h"
 #endif
 
 #ifndef OPENTHREAD_CONFIG_THREAD_VERSION


wgtdkp (approved):
LGTM :+1:

bukepo (approved):

src/core/openthread-core-config.h
```diff
@@ -44,6 +44,10 @@
 
 #ifdef OPENTHREAD_PROJECT_CORE_CONFIG_FILE
 #include OPENTHREAD_PROJECT_CORE_CONFIG_FILE
+#elif defined(OPENTHREAD_CONFIG_CORE_USER_CONFIG_HEADER_ENABLE)
+// This configuration header file should be implemented by the user when the
```
abtink, line 48:
```suggestion
// This configuration header file should be provided by the user when the
```

abtink (approved):
LGTM. Thanks. 

One suggestion below and one on the commit message. 

> This CL includes the default user implemented configuration header file
to avoid specifying the configuration file.

Let's not use `CL`. Github uses the git  "pull request" (PR) term.
I would say,  for commit message log it'd be even better to say "This commit includes ..." instead of "This PR" (or "This CL").

src/core/openthread-core-config.h
jwhui:
```suggestion
// This configuration header file should be provided by the user when
```

zhanglongxia:
> LGTM. Thanks.
> 
> One suggestion below and one on the commit message.
> 
> > This CL includes the default user implemented configuration header file
> > to avoid specifying the configuration file.
> 
> Let's not use `CL`. Github uses the git "pull request" (PR) term.
> I would say, for commit message log it'd be even better to say "This commit includes ..." instead of "This PR" (or "This CL").

I have updated the commit message.

jwhui (approved):

PR #6830, opened 2021-07-19
[border-routing] dump sent/received RA messages on infra if

wgtdkp:
Dump the RA messages sent/received on infra if for testing scripts.
This commit removes the requirement of `OT_REFERENCE_DEVICE`
but always dump the RA messages at log level `DEBG`.

### DIFF
--- a/src/core/border_router/routing_manager.cpp
+++ b/src/core/border_router/routing_manager.cpp
@@ -765,9 +765,7 @@ void RoutingManager::SendRouterAdvertisement(const Ip6::Prefix *aNewOmrPrefixes,
         if (error == kErrorNone)
         {
             otLogInfoBr("Sent Router Advertisement on interface %u", mInfraIfIndex);
-#if OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE
-            otDumpCertBr("[BR-CERT] direction=send | type=RA |", buffer, bufferLength);
-#endif
+            otDumpDebgBr("[BR-CERT] direction=send | type=RA |", buffer, bufferLength);
         }
         else
         {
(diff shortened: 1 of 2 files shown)


simonlingoogle (approved):
LGTM :+1:

abtink:
@wgtdkp , what's the reason for this change?

With this the code will dump the entire content of received/sent RA messages at `INFO` level (which is often used). Would it not add too much logs?

simonlingoogle:
> @wgtdkp , what's the reason for this change?
> 
> With this the code will dump the entire content of received/sent RA messages at `INFO` level (which is often used). Would it not add too much logs?

We need this RA dump for testing. 

One possible solution is that we can use `DEBG` level for RA dump and also allow using `log level` to switch the dynamic log level to `DEBG`.

wgtdkp:
@abtink The problem is that the BR is the `Device Under Test (DUT)` which should not enable the `OT_REFERENCE_DEVICE` feature. One way is that we downgrade the log level to `DEBG` as suggested by @simonlingoogle, or we introduce additional configuration flag for dumping the RA message. Thoughts? @librasungirl

abtink:
> @abtink The problem is that the BR is the Device Under Test (DUT) which should not enable the OT_REFERENCE_DEVICE feature. One way is that we downgrade the log level to DEBG as suggested by @simonlingoogle, or we introduce additional configuration flag for dumping the RA message. Thoughts? @librasungirl

Make sense. Thanks. All options are ok in my view. I would lean towards adding a new config for this maybe. 
- Another idea is if we don't need the format `"[BR-CERT] direction=recv | type=RA |"` and/or the full message payload, to consider logging a shorter message saying we received RA with this size with so and so info (i.e. parse the important info from RA to log).

wgtdkp:
> Make sense. Thanks. All options are ok in my view. I would lean towards adding a new config for this maybe.
> 
> * Another idea is if we don't need the format `"[BR-CERT] direction=recv | type=RA |"` and/or the full message payload, to consider logging a shorter message saying we received RA with this size with so and so info (i.e. parse the important info from RA to log).

Thanks for the suggestions! We decided to use `DEBG` logging level.
Regarding the log format, I think we need to dump the original RA messages blob for the testing script to verify that the RA messages sent by the BR are in good format. Actually, the testing script should capture this traffic with a Wi-Fi sniffer but that is not ready currently, so we help dump the RA messages here.

librasungirl (approved):

abtink (approved):
LGTM. Thanks.

jwhui (approved):

PR #2690, opened 2018-04-28
[posix] radio only host side driver

bukepo:
This PR mainly implements a new POSIX platform in `src/posix` based on OpenThread transceiver.

It's based on the existing POSIX simulation platform. Main changes happens in the implementation of radio part. Instead of manipulating UDP socket, this platform transmits and receives real radio frames through a radio transceiver. Currently this platform only speaks spinel with transceiver which is added into OpenThread in #2085.

*NOTE* To reduce the difficulty of reviewing, this PR keeps most code of POSIX simulation untouched. Enhancements is going to be done in future PRs.

* added new POSIX platform and two apps(*ot-ncp* and *ot-cli*) based on this platform.
* added two Travis tests for this platform.
* modified test scripts to support this platform.

PR #6837, opened 2021-07-20
[radio-spinel] fix incorrect RX channel after frame TX on another channel

abtink:
This commit fixes a bug in `RadioSpinel` platform implementation where
the request from OT core stack to enter RX mode may be ignored
causing the RCP/radio to stay in RX mode on an incorrect channel.

This situation can happen after a frame TX request on a different
channel (note that the `otRadioFrame` struct specifies the channel on
which the frame should be transmitted). After frame TX, radio driver
is expected to enter RX mode on the same channel on which the frame
TX happened. The OpenThread MAC layer always explicitly instructs the
radio to enter RX mode on the PAN channel (from `Mac::UpdateIdleMode()`
which in turn calls the radio platform API `otPlatRadioRecive()` with
the channel passed in as a parameter). The earlier `RadioSpinel`
platform implementation cached the channel in a local member variable
and skipped setting the channel on RCP if the channel was same as
before (setting the `SPINEL_PROP_PHY_CHAN` instructs the RCP to enter
RX mode on the correct channel).

-------

This may help address https://github.com/openthread/ot-br-posix/issues/935.

------

I looked at history of the `RadioSpinel` code. Seems like this behavior/bug 
may have present since the very first implementation:  
- https://github.com/openthread/openthread/pull/2690 

The question is how it was not detected earlier. I think a bunch of  behaviors
helped mask this issue
- A follow-up frame tx (on PAN channel) can cause RCP to get back to 
same channel (router would send "Link Adv" every ~30sec).
-  "channel monitor" feature which periodically issues energy scan can 
force the  RCP to get back to the right channel
- I would like us to figure out a test-case that covers this  situation 
specifically.

---------

Adding one more note on this:
- I recalled that in `SubMac `implementation, the OT core would put the radio in 
RX mode before any TX request (the radio is put in RX on same channel as TX),
i.e. we have a call to `otPlatRadioReceive()` before the `otPlatRadioTransmit()`
- This would basically ensure that existing `RadioSpinel` code to function correctly
(since there would be call to `Receive()` and change the cached `mChannel` before
any tx request).
- However, the behavior in `SubMac` was changed from 
https://github.com/openthread/openthread/pull/4697
- With this change the call to `Receive()` would be skipped if the radio supports 
the newly added capability `OT_RADIO_CAPS_SLEEP_TO_TX` falg.
- I think Nordic radio driver includes this flag and `RadioSpinel` would simply
forward the same set of capability flags it gets from its RCP.
- So looks like this error situation (with `RadioSpinel` driver) is a regression after 
the above PR.

### DIFF
--- a/src/lib/spinel/radio_spinel_impl.hpp
+++ b/src/lib/spinel/radio_spinel_impl.hpp
@@ -1514,6 +1514,8 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::EnergyScan(uint8_t aScan
     SuccessOrExit(error = Set(SPINEL_PROP_MAC_SCAN_PERIOD, SPINEL_DATATYPE_UINT16_S, aScanDuration));
     SuccessOrExit(error = Set(SPINEL_PROP_MAC_SCAN_STATE, SPINEL_DATATYPE_UINT8_S, SPINEL_SCAN_STATE_ENERGY));
 
+    mChannel = aScanChannel;
+
 exit:
     return error;
 }
@@ -1923,6 +1925,7 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Transmit(otRadioFrame &a
         // Waiting for `TransmitDone` event.
         mState        = kStateTransmitting;
         mTxRadioEndUs = otPlatTimeGet() + TX_WAIT_US;
+        mChannel      = mTransmitFrame->mChannel;
     }
 
 exit:


jwhui:
@mspang , can you see if this addresses your issue?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1936,12 +1936,9 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Receive(uint8_t aChannel
...
 
-    if (mChannel != aChannel)
-    {
-        error = Set(SPINEL_PROP_PHY_CHAN, SPINEL_DATATYPE_UINT8_S, aChannel);
-        SuccessOrExit(error);
-        mChannel = aChannel;
-    }
+    error = Set(SPINEL_PROP_PHY_CHAN, SPINEL_DATATYPE_UINT8_S, aChannel);
+    SuccessOrExit(error);
+    mChannel = aChannel;
```
simonlingoogle, line 1941: Is it workable to cache the the transmitting channel in mChannel after a frame TX request?

mspang:
> @mspang , can you see if this addresses your issue?

So far I have not been able to reproduce the issue with this change.

src/lib/spinel/radio_spinel_impl.hpp
abtink: This can also work. Was considering it as well. We may need to do the same on energy scan.

simonlingoogle: I think we should avoid unnecessary calls to Set(SPINEL_PROP_PHY_CHAN, ...) since it's a blocking operation, thus having significant cost.

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1514,6 +1514,8 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::EnergyScan(uint8_t aScan
     SuccessOrExit(error = Set(SPINEL_PROP_MAC_SCAN_PERIOD, SPINEL_DATATYPE_UINT16_S, aScanDuration));
     SuccessOrExit(error = Set(SPINEL_PROP_MAC_SCAN_STATE, SPINEL_DATATYPE_UINT8_S, SPINEL_SCAN_STATE_ENERGY));
 
+    mChannel = aScanChannel;
```
bukepo, line 1517: Remove mScanChannel?

bukepo (approved):

src/lib/spinel/radio_spinel_impl.hpp
abtink: That's exactly the reason I was thinking of the same. It is already changed in new push.

src/lib/spinel/radio_spinel_impl.hpp
abtink: This may be an optimization to be considered for future (a separate PR).
I don't think we want to add optimization from this PR (focus should be on the fix).

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1923,6 +1925,7 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Transmit(otRadioFrame &a
         // Waiting for `TransmitDone` event.
         mState        = kStateTransmitting;
         mTxRadioEndUs = otPlatTimeGet() + TX_WAIT_US;
+        mChannel      = mTransmitFrame->mChannel;
```
simonlingoogle, line 1928: After `SPINEL_PROP_STREAM_RAW` request is sent, the RCP haven't sent the frame because it's doing CMSA backoff. So, the actual channel won't be changed to `mTransmitFrame->mChannel` until CSMA backoff is finished. This will cause a temporary mismatch between `mChannel` and actual RCP channel.
I am not sure if this is going to cause any problem.

abtink: The safer solution is to avoid caching `mChannel` and always set it on RCP (which has the overhead of extra calls `Set(SPINEL_PROP_PHY_CHAN, ...)` we discussed in https://github.com/openthread/openthread/pull/6837#discussion_r673658912.

I don't think the above situation is an issue.  
- After any backoff delay the radio should perform CA check on the same channel. 
- It is also the same behavior with any radio platform (RCP or not).
- OT core will always issue a `otPlatRadioReceive()` after any frame tx (success or failed) from `UpdateIdleMode()`. 

The `RadioSpinel`/RCP corner case I can think of (which should still work fine) is if RCP hangs/resets in middle of a frame tx and we have the "RCP reset recovery" feature enabled (i.e. `OPENTHREAD_SPINEL_CONFIG_RCP_RESTORATION_MAX_COUNT > 0`). 
- In this case the "RCP recovery logic" will set the channel to the frame tx channel.
- But it should be ok, since from OT stack perspective we are in middle of tx (on that channel) and expect a `txdone` callback from which we will call `UpdateIdleMode()` and set the channel to whatever PAN channel is (or basically what OT stack wants it to be).

simonlingoogle (approved):
LGTM :+1:

src/lib/spinel/radio_spinel_impl.hpp
simonlingoogle: SGTM.

jwhui (approved):
Thanks! 👍🏼

PR #6840, opened 2021-07-21
[posix] fix memcpy compiler warning false positive

superwhd:
Fixes https://github.com/openthread/ot-br-posix/issues/931.

In the issue, compiler complained we were copying 16 bytes to a location of 12 bytes. This was because in the check, the compiler thought we are going to copy data to a `nlmsghdr` which is 12 bytes. The reality is that we're copying data to a place behind `nlmsghdr` in `req` struct.


abtink (approved):
LGTM. 👍

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6845, opened 2021-07-22
[core] replace `enum` constants with `constexpr`

abtink:
This PR contains a set of separate commits changing the constants 
to use `constexpr` instead of `enume` under under different 
sub-folders: `coap`, `common`, `crypto` and `diags`.


simonlingoogle (approved):
LGTM :+1:

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #6849, opened 2021-07-23
[mac] add `mIsHeaderUpdated` frame flag

edmont:
When a platform provides `OT_RADIO_CAPS_TRANSMIT_SEC` capability,
it updates the frame header with the dynamic data (frame counter,
key id...) and returns it back to MAC together with the transmission
result.

So far the `mIsARetx` flag was being used to instruct the radio
whether to update the header or not, but there was no mean to know
if the radio was able to do so.

This commit introduces the `mIsHeaderUpdated` flag with is used in
both directions, so the MAC layer can react to cases in which the
frame header was not properly updated by the radio plaform.

The usage of `mIsARetx` remains restricted now to two cases:
- When a SSED decides to re-synchronize CSL parameters after a
retransmission.
- To check if sequence number was set for CSL/indirect retransmissions.

### DIFF
--- a/include/openthread/platform/radio.h
+++ b/include/openthread/platform/radio.h
@@ -239,24 +239,28 @@ typedef struct otRadioFrame
             uint8_t         mMaxFrameRetries; ///< Maximum number of retries allowed after a transmission failure.
 
             /**
-             * Indicates whether the frame is a retransmission or not.
+             * Indicates whether frame counter and CSL IEs are properly updated in the header.
              *
              * If the platform layer does not provide `OT_RADIO_CAPS_TRANSMIT_SEC` capability, it can ignore this flag.
              *
              * If the platform provides `OT_RADIO_CAPS_TRANSMIT_SEC` capability, then platform is expected to handle tx
              * security processing and assignment of frame counter. In this case the following behavior is expected:
              *
-             * When `mIsARetx` is set, it indicates that OpenThread core has already set the frame counter and key id
-             * (if security is enabled) in the prepared frame. The counter is ensured to match the counter value from
-             * the previous attempts of the same frame. The platform should not assign or change the frame counter (but
-             * may still need to perform security processing depending on `mIsSecurityProcessed` flag).
+             * When `mIsHeaderUpdated` is set, it indicates that OpenThread core has already set the frame counter and
+             * CSL IEs (if security is enabled) in the prepared frame. The counter is ensured to match the counter value
+             * from the previous attempts of the same frame. The platform should not assign or change the frame counter
+             * (but may still need to perform security processing depending on `mIsSecurityProcessed` flag).
              *
-             * If `mIsARetx` is not set, then the frame counter and key id are not set in the frame by OpenThread core
-             * and it is the responsibility of the radio platform to assign them. The platform should update the frame
-             * (assign counter and key id) even if the transmission gets aborted or fails (e.g., channel access error).
+             * If `mIsHeaderUpdated` is not set, then the frame counter and key CSL IE not set in the frame by
+             * OpenThread core and it is the responsibility of the radio platform to assign them. The platform
+             * must update the frame header (assign counter and CSL IE values) before sending the frame over the air,
+             * however if the the transmission gets aborted and the frame is never sent over the air (e.g., channel
+             * access error) the platform may choose to not update the header. If the platform updates the header,
+             * it must also set this flag before passing the frame back from the `otPlatRadioTxDone()` callback.
              *
              */
-            bool mIsARetx : 1;
+            bool mIsHeaderUpdated : 1;
+            bool mIsARetx : 1;             ///< Indicates whether the frame is a retransmission or not.
             bool mCsmaCaEnabled : 1;       ///< Set to true to enable CSMA-CA for this packet, false otherwise.
             bool mCslPresent : 1;          ///< Set to true if CSL header IE is present.
             bool mIsSecurityProcessed : 1; ///< True if SubMac should skip the AES processing of this frame.
--- a/src/core/mac/data_poll_handler.cpp
+++ b/src/core/mac/data_poll_handler.cpp
@@ -242,8 +242,9 @@ void DataPollHandler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError, 
         break;
 
     case kErrorNoAck:
-        aChild.IncrementIndirectTxAttempts();
+        OT_ASSERT(!aFrame.GetSecurityEnabled() || aFrame.IsHeaderUpdated());
 
+        aChild.IncrementIndirectTxAttempts();
         otLogInfoMac("Indirect tx to child %04x failed, attempt %d/%d", aChild.GetRloc16(),
                      aChild.GetIndirectTxAttempts(), kMaxPollTriggeredTxAttempts);
 
@@ -268,7 +269,7 @@ void DataPollHandler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError, 
 
             aChild.SetIndirectDataSequenceNumber(aFrame.GetSequence());
 
-            if (aFrame.GetSecurityEnabled())
+            if (aFrame.GetSecurityEnabled() && aFrame.IsHeaderUpdated())
             {
                 uint32_t frameCounter;
                 uint8_t  keyId;
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -950,7 +950,7 @@ void Mac::ProcessTransmitSecurity(TxFrame &aFrame)
         aFrame.SetAesKey(keyManager.GetKek());
         extAddress = &GetExtAddress();
 
-        if (!aFrame.IsARetransmission())
+        if (!aFrame.IsHeaderUpdated())
         {
             aFrame.SetFrameCounter(keyManager.GetKekFrameCounter());
             keyManager.IncrementKekFrameCounter();
@@ -974,13 +974,13 @@ void Mac::ProcessTransmitSecurity(TxFrame &aFrame)
         aFrame.SetAesKey(*mLinks.GetCurrentMacKey(aFrame));
         extAddress = &GetExtAddress();
 
-        // If the frame is marked as a retransmission, `MeshForwarder` which
+        // If the frame header is marked as updated, `MeshForwarder` which
         // prepared the frame should set the frame counter and key id to the
-        // same values used in the earlier transmit attempt. For a new frame (not
-        // a retransmission), we get a new frame counter and key id from the key
+        // same values used in the earlier transmit attempt. For a new frame (header
+        // not updated), we get a new frame counter and key id from the key
         // manager.
 
-        if (!aFrame.IsARetransmission())
+        if (!aFrame.IsHeaderUpdated())
         {
             mLinks.SetMacFrameCounter(aFrame);
             aFrame.SetKeyId((keyManager.GetCurrentKeySequence() & 0x7f) + 1);
@@ -1086,7 +1086,7 @@ void Mac::BeginTransmit(void)
         frame = Get<DataPollHandler>().HandleFrameRequest(txFrames);
         VerifyOrExit(frame != nullptr);
 
-        // If the frame is marked as a retransmission, then data sequence number is already set.
+        // If the frame is marked as retransmission, then data sequence number is already set.
         if (!frame->IsARetransmission())
         {
             frame->SetSequence(mDataSequence++);
@@ -1100,7 +1100,7 @@ void Mac::BeginTransmit(void)
         frame = Get<CslTxScheduler>().HandleFrameRequest(txFrames);
         VerifyOrExit(frame != nullptr);
 
-        // If the frame is marked as a retransmission, then data sequence number is already set.
+        // If the frame is marked as retransmission, then data sequence number is already set.
         if (!frame->IsARetransmission())
         {
             frame->SetSequence(mDataSequence++);
--- a/src/core/thread/csl_tx_scheduler.cpp
+++ b/src/core/thread/csl_tx_scheduler.cpp
@@ -249,8 +249,9 @@ void CslTxScheduler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError, C
         break;
 
     case kErrorNoAck:
-        aChild.IncrementCslTxAttempts();
+        OT_ASSERT(!aFrame.GetSecurityEnabled() || aFrame.IsHeaderUpdated());
 
+        aChild.IncrementCslTxAttempts();
         otLogInfoMac("CSL tx to child %04x failed, attempt %d/%d", aChild.GetRloc16(), aChild.GetCslTxAttempts(),
                      kMaxCslTriggeredTxAttempts);
 
@@ -274,7 +275,7 @@ void CslTxScheduler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError, C
         {
             aChild.SetIndirectDataSequenceNumber(aFrame.GetSequence());
 
-            if (aFrame.GetSecurityEnabled())
+            if (aFrame.GetSecurityEnabled() && aFrame.IsHeaderUpdated())
             {
                 uint32_t frameCounter;
                 uint8_t  keyId;
--- a/src/lib/spinel/radio_spinel_impl.hpp
+++ b/src/lib/spinel/radio_spinel_impl.hpp
@@ -1843,9 +1843,10 @@ void RadioSpinel<InterfaceType, ProcessContextType>::HandleTransmitDone(uint32_t
                                                                         const uint8_t *   aBuffer,
                                                                         uint16_t          aLength)
 {
-    otError         error        = OT_ERROR_NONE;
-    spinel_status_t status       = SPINEL_STATUS_OK;
-    bool            framePending = false;
+    otError         error         = OT_ERROR_NONE;
+    spinel_status_t status        = SPINEL_STATUS_OK;
+    bool            framePending  = false;
+    bool            headerUpdated = false;
     spinel_ssize_t  unpacked;
 
     VerifyOrExit(aCommand == SPINEL_CMD_PROP_VALUE_IS && aKey == SPINEL_PROP_LAST_STATUS, error = OT_ERROR_FAILED);
@@ -1862,6 +1863,12 @@ void RadioSpinel<InterfaceType, ProcessContextType>::HandleTransmitDone(uint32_t
     aBuffer += unpacked;
     aLength -= static_cast<uint16_t>(unpacked);
 
+    unpacked = spinel_datatype_unpack(aBuffer, aLength, SPINEL_DATATYPE_BOOL_S, &headerUpdated);
+    VerifyOrExit(unpacked > 0, error = OT_ERROR_PARSE);
+
+    aBuffer += unpacked;
+    aLength -= static_cast<uint16_t>(unpacked);
+
     if (status == SPINEL_STATUS_OK)
     {
         SuccessOrExit(error = ParseRadioFrame(mAckRadioFrame, aBuffer, aLength, unpacked));
@@ -1873,7 +1880,10 @@ void RadioSpinel<InterfaceType, ProcessContextType>::HandleTransmitDone(uint32_t
         error = SpinelStatusToOtError(status);
     }
 
-    if ((mRadioCaps & OT_RADIO_CAPS_TRANSMIT_SEC) && static_cast<Mac::TxFrame *>(mTransmitFrame)->GetSecurityEnabled())
+    static_cast<Mac::TxFrame *>(mTransmitFrame)->SetIsHeaderUpdated(headerUpdated);
+
+    if ((mRadioCaps & OT_RADIO_CAPS_TRANSMIT_SEC) && headerUpdated &&
+        static_cast<Mac::TxFrame *>(mTransmitFrame)->GetSecurityEnabled())
     {
         uint8_t  keyId;
         uint32_t frameCounter;
@@ -1905,20 +1915,21 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Transmit(otRadioFrame &a
     otPlatRadioTxStarted(mInstance, mTransmitFrame);
 
     error = Request(SPINEL_CMD_PROP_VALUE_SET, SPINEL_PROP_STREAM_RAW,
-                    SPINEL_DATATYPE_DATA_WLEN_S                               // Frame data
-                        SPINEL_DATATYPE_UINT8_S                               // Channel
-                            SPINEL_DATATYPE_UINT8_S                           // MaxCsmaBackoffs
-                                SPINEL_DATATYPE_UINT8_S                       // MaxFrameRetries
-                                    SPINEL_DATATYPE_BOOL_S                    // CsmaCaEnabled
-                                        SPINEL_DATATYPE_BOOL_S                // IsARetx
-                                            SPINEL_DATATYPE_BOOL_S            // SkipAes
-                                                SPINEL_DATATYPE_UINT32_S      // TxDelay
-                                                    SPINEL_DATATYPE_UINT32_S, // TxDelayBaseTime
+                    SPINEL_DATATYPE_DATA_WLEN_S                                   // Frame data
+                        SPINEL_DATATYPE_UINT8_S                                   // Channel
+                            SPINEL_DATATYPE_UINT8_S                               // MaxCsmaBackoffs
+                                SPINEL_DATATYPE_UINT8_S                           // MaxFrameRetries
+                                    SPINEL_DATATYPE_BOOL_S                        // CsmaCaEnabled
+                                        SPINEL_DATATYPE_BOOL_S                    // IsHeaderUpdated
+                                            SPINEL_DATATYPE_BOOL_S                // IsARetx
+                                                SPINEL_DATATYPE_BOOL_S            // SkipAes
+                                                    SPINEL_DATATYPE_UINT32_S      // TxDelay
+                                                        SPINEL_DATATYPE_UINT32_S, // TxDelayBaseTime
                     mTransmitFrame->mPsdu, mTransmitFrame->mLength, mTransmitFrame->mChannel,
                     mTransmitFrame->mInfo.mTxInfo.mMaxCsmaBackoffs, mTransmitFrame->mInfo.mTxInfo.mMaxFrameRetries,
-                    mTransmitFrame->mInfo.mTxInfo.mCsmaCaEnabled, mTransmitFrame->mInfo.mTxInfo.mIsARetx,
-                    mTransmitFrame->mInfo.mTxInfo.mIsSecurityProcessed, mTransmitFrame->mInfo.mTxInfo.mTxDelay,
-                    mTransmitFrame->mInfo.mTxInfo.mTxDelayBaseTime);
+                    mTransmitFrame->mInfo.mTxInfo.mCsmaCaEnabled, mTransmitFrame->mInfo.mTxInfo.mIsHeaderUpdated,
+                    mTransmitFrame->mInfo.mTxInfo.mIsARetx, mTransmitFrame->mInfo.mTxInfo.mIsSecurityProcessed,
+                    mTransmitFrame->mInfo.mTxInfo.mTxDelay, mTransmitFrame->mInfo.mTxInfo.mTxDelayBaseTime);
 
     if (error == OT_ERROR_NONE)
     {
--- a/src/ncp/ncp_base_radio.cpp
+++ b/src/ncp/ncp_base_radio.cpp
@@ -148,8 +148,9 @@ void NcpBase::LinkRawTransmitDone(otRadioFrame *aFrame, otRadioFrame *aAckFrame,
 
     if (mCurTransmitTID)
     {
-        uint8_t header       = SPINEL_HEADER_FLAG | SPINEL_HEADER_IID_0 | mCurTransmitTID;
-        bool    framePending = (aAckFrame != nullptr && static_cast<Mac::RxFrame *>(aAckFrame)->GetFramePending());
+        uint8_t header        = SPINEL_HEADER_FLAG | SPINEL_HEADER_IID_0 | mCurTransmitTID;
+        bool    framePending  = (aAckFrame != nullptr && static_cast<Mac::RxFrame *>(aAckFrame)->GetFramePending());
+        bool    headerUpdated = static_cast<Mac::TxFrame *>(aFrame)->IsHeaderUpdated();
 
         // Clear cached transmit TID
         mCurTransmitTID = 0;
@@ -157,13 +158,14 @@ void NcpBase::LinkRawTransmitDone(otRadioFrame *aFrame, otRadioFrame *aAckFrame,
         SuccessOrExit(mEncoder.BeginFrame(header, SPINEL_CMD_PROP_VALUE_IS, SPINEL_PROP_LAST_STATUS));
         SuccessOrExit(mEncoder.WriteUintPacked(ThreadErrorToSpinelStatus(aError)));
         SuccessOrExit(mEncoder.WriteBool(framePending));
+        SuccessOrExit(mEncoder.WriteBool(headerUpdated));
 
         if (aError == OT_ERROR_NONE)
         {
             SuccessOrExit(PackRadioFrame(aAckFrame, aError));
         }
 
-        if (static_cast<Mac::TxFrame *>(aFrame)->GetSecurityEnabled())
+        if (static_cast<Mac::TxFrame *>(aFrame)->GetSecurityEnabled() && headerUpdated)
         {
             uint8_t  keyId;
             uint32_t frameCounter;
@@ -386,6 +388,7 @@ otError NcpBase::DecodeStreamRawTxRequest(otRadioFrame &aFrame)
     uint16_t       payloadLen;
     bool           csmaEnable;
     bool           isARetx;
+    bool           isHeaderUpdated;
     bool           isSecurityProcessed;
 
     SuccessOrExit(error = mDecoder.ReadDataWithLen(payloadPtr, payloadLen));
@@ -403,6 +406,7 @@ otError NcpBase::DecodeStreamRawTxRequest(otRadioFrame &aFrame)
     aFrame.mInfo.mTxInfo.mMaxCsmaBackoffs     = OPENTHREAD_CONFIG_MAC_MAX_CSMA_BACKOFFS_DIRECT;
     aFrame.mInfo.mTxInfo.mMaxFrameRetries     = OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_DIRECT;
     aFrame.mInfo.mTxInfo.mCsmaCaEnabled       = true;
+    aFrame.mInfo.mTxInfo.mIsHeaderUpdated     = false;
     aFrame.mInfo.mTxInfo.mIsARetx             = false;
     aFrame.mInfo.mTxInfo.mIsSecurityProcessed = false;
     aFrame.mInfo.mTxInfo.mTxDelay             = 0;
@@ -416,11 +420,13 @@ otError NcpBase::DecodeStreamRawTxRequest(otRadioFrame &aFrame)
     SuccessOrExit(mDecoder.ReadUint8(aFrame.mInfo.mTxInfo.mMaxCsmaBackoffs));
     SuccessOrExit(mDecoder.ReadUint8(aFrame.mInfo.mTxInfo.mMaxFrameRetries));
     SuccessOrExit(mDecoder.ReadBool(csmaEnable));
+    SuccessOrExit(mDecoder.ReadBool(isHeaderUpdated));
     SuccessOrExit(mDecoder.ReadBool(isARetx));
     SuccessOrExit(mDecoder.ReadBool(isSecurityProcessed));
     SuccessOrExit(mDecoder.ReadUint32(aFrame.mInfo.mTxInfo.mTxDelay));
     SuccessOrExit(mDecoder.ReadUint32(aFrame.mInfo.mTxInfo.mTxDelayBaseTime));
     aFrame.mInfo.mTxInfo.mCsmaCaEnabled       = csmaEnable;
+    aFrame.mInfo.mTxInfo.mIsHeaderUpdated     = isHeaderUpdated;
     aFrame.mInfo.mTxInfo.mIsARetx             = isARetx;
     aFrame.mInfo.mTxInfo.mIsSecurityProcessed = isSecurityProcessed;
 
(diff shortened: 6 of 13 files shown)


include/openthread/platform/radio.h
```diff
@@ -239,24 +239,26 @@ typedef struct otRadioFrame
...
              *
              * If the platform provides `OT_RADIO_CAPS_TRANSMIT_SEC` capability, then platform is expected to handle tx
              * security processing and assignment of frame counter. In this case the following behavior is expected:
              *
-             * When `mIsARetx` is set, it indicates that OpenThread core has already set the frame counter and key id
-             * (if security is enabled) in the prepared frame. The counter is ensured to match the counter value from
-             * the previous attempts of the same frame. The platform should not assign or change the frame counter (but
-             * may still need to perform security processing depending on `mIsSecurityProcessed` flag).
+             * When `mIsHeaderUpdated` is set, it indicates that OpenThread core has already set the frame counter and
+             * key id (if security is enabled) in the prepared frame. The counter is ensured to match the counter value
```
hubertmis, line 250: Mentioning `key id` here is misleading. Radio platform with `OT_RADIO_CAPS_TRANSMIT_SEC` expect `key id` to be set regardless `mIsHeaderUpdated` value. `mIsHeaderUpdated = false` indicates that Frame Counter and CSL IE are not set.

src/core/mac/mac.cpp
```diff
@@ -1086,8 +1086,8 @@ void Mac::BeginTransmit(void)
         frame = Get<DataPollHandler>().HandleFrameRequest(txFrames);
         VerifyOrExit(frame != nullptr);
 
-        // If the frame is marked as a retransmission, then data sequence number is already set.
-        if (!frame->IsARetransmission())
+        // If the frame header is marked as not updated, then data sequence number is already set.
```
hubertmis, line 1089:
```suggestion
        // If the frame header is marked as updated, then data sequence number is already set.
```

src/core/mac/mac.cpp
```diff
@@ -1100,8 +1100,8 @@ void Mac::BeginTransmit(void)
         frame = Get<CslTxScheduler>().HandleFrameRequest(txFrames);
         VerifyOrExit(frame != nullptr);
 
-        // If the frame is marked as a retransmission, then data sequence number is already set.
-        if (!frame->IsARetransmission())
+        // If the frame header is marked as not updated, then data sequence number is already set.
```
hubertmis, line 1103:
```suggestion
        // If the frame header is marked as updated, then data sequence number is already set.
```

abtink:
> So far the mIsARetx flag was being used to instruct the radio
> whether to update the header or not, but there was no mean to know
> if the radio was able to do so.
> 
> This commit introduces the mIsHeaderUpdated flag with is used in
> both directions, so the MAC layer can react to cases in which the
> frame header was not properly updated by the radio plaform.

Thanks @edmont . I have couple of questions/suggestions on this:

- In what situation the radio platform may not being able to update the header? 
- Can we try to address this in radio platform layer (i.e., ensuring even if tx is being aborted, the frame header is updated before invoking callback)?

I think the change from this PR would require all platform implementations to be updated. 

In general, I think it'd be good to minimize/avoid changes to radio platform layer definitions and behaviors as much as we can. And try to only add new behaviors (in order to keep things backward compatible if/when we can).

jciupis:
> In what situation the radio platform may not be able to update the header?

As an example, let's suppose that a transmission of a frame with CSL IE is requested while the channel is busy. The CSL Phase field requires accuracy of 10 symbols, which is quite tight. As a result, the radio platform must wait with updating that field at least until it's certain that the transmission can be carried out, or possibly even when the transmission is already ongoing. If the channel is busy and all CSMA/CA backoffs end with a failed CCA, the CSL Phase might never be updated by the radio platform. That information can be used when a retransmission is requested.

abtink:
Thanks @jciupis 

> As an example, let's suppose that a transmission of a frame with CSL IE is requested while the channel is busy. The CSL Phase field requires accuracy of 10 symbols, which is quite tight. As a result, the radio platform must wait with updating that field at least until it's certain that the transmission can be carried out, or possibly even when the transmission is already ongoing. If the channel is busy and all CSMA/CA backoffs end with a failed CCA, the CSL Phase might never be updated by the radio platform. That information can be used when a retransmission is requested.

Two questions: 
- Would it matter to OT core whether the phase is set on a failed frame tx or not?
- If it does, can we address it in radio platform layer, i.e. in case of such failure, we still set the fields in frame header to some reasonable values before invoking the `txdone` callback.

I think it'd be good to avid the need for a behavior change in radio platform layer (if/when we can) which can require all platform implementations to be updated.

jciupis:
> Would it matter to OT core whether the phase is set on a failed frame tx or not? If it does, can we [...] still set the fields in frame header to some reasonable values before invoking the txdone callback

Suppose that instead of providing the `mIsHeaderUpdated` flag, the radio platform always updates the header with "best effort" values available before passing the notification to the upper layer. To avoid security breaches caused by using the same nonce to encrypt a different plain text, CSL phase cannot be updated once Frame Counter was set by the radio platform. Even with this reasonable, "best effort" CSL phase value written before invoking the callback, CSL synchronization will be lost for all scenarios that could otherwise not modify the header at all and set `mIsHeaderUpdated` to false. An example of such scenario is the busy channel case mentioned earlier.

I can't tell if it's worth breaking the backwards compatibility though.

hubertmis:
> I think it'd be good to avid the need for a behavior change in radio platform layer

I don't think this patch requires changes in any other platform layer than Nordic. It seems Nordic radio is the only one available in OpenThread that supports encryption in radio platform layer (`OT_RADIO_CAPS_TRANSMIT_SEC`). If given platform layer does not support frame encryption, `mIsHeaderUpdated` is set by OpenThread MAC layer during processing outgoing frame security procedure. Only if OpenThread MAC skips security processing (because given radio platform runs encryption instead), the radio platform is responsible to mark if frame counter was set on given frame or not.

Moreover, now looking at this PR I'm thinking if platforms without encryption support in radio platforms support frame encryption correctly with CSL.
Each transmission operation starts with `otPlatRadioTransmit()` and ends with `otPlatRadioTxDone()`. Meantime there may or may not be an event that frame transmission actually started like `nrf_802154_tx_started()`. This event is omitted if transmission was aborted before complete preamble was transmitted. This event is used to update CSL IE and encrypt outgoing frame. It is possible that when `otPlatRadioTxDone()` is called, the frame was encrypted (if `nrf_802154_tx_started()` occurred, or not (if `nrf_802154_tx_started()`  was not called). However, `otPlatRadioTxDone()` assumes that frame is already encrypted (setting `IsARetransmission` flag). This patch should fix described issue, by calling `SetIsHeaderUpdated()` only after security is processed. Not unconditionally in `otPlatRadioTxDone()`.

edmont:
Hi @abtink, I do agree with @hubertmis observations. Are you aware of any other platform implementing `OT_RADIO_CAPS_TRANSMIT_SEC` at the moment?

abtink:
> Moreover, now looking at this PR I'm thinking if platforms without encryption support in radio platforms support frame encryption correctly with CSL.
> Each transmission operation starts with otPlatRadioTransmit() and ends with otPlatRadioTxDone(). Meantime there may or may not be an event that frame transmission actually started like nrf_802154_tx_started(). This event is omitted if transmission was aborted before complete preamble was transmitted. This event is used to update CSL IE and encrypt outgoing frame. It is possible that when otPlatRadioTxDone() is called, the frame was encrypted (if nrf_802154_tx_started() occurred, or not (if nrf_802154_tx_started() was not called). However, otPlatRadioTxDone() assumes that frame is already encrypted (setting IsARetransmission flag). This patch should fix described issue, by calling SetIsHeaderUpdated() only after security is processed. Not unconditionally in otPlatRadioTxDone().

We do have a flag `bool mIsSecurityProcessed : 1;` in the `otRadioFrame` which indicates whether the tx security is processed or not. It is set by OT stack when it processes the tx security. I think the same flag can be used by the radio platform to indicate whether it did perform sec processing or not (from `TxDone()` when frame is given back to OT stack).

The `mIsRetx` and `mIsSecurityProcessed` are not directly related. We can have frame marked as `mIsRetx`  independent of `mIsSecurityProcessed`.

------

I still think it'd be better to not add `mIsHeaderUpdated` flag and instead ensure that platform layer always update the header.
- This makes it behave similar to how OT `SubMac` behaves and keep the code simpler.
- OT stack only cares that frame counter and key id and seq number are properly set on the frame (from `TxDone()` indicating tx failure). Whether the frame security is actually processed does not matter.
- The counter, key id, seq number info is only used  for indirect frame retx:
   -  OT saves the the info for the failed indirect tx (stored in the related `Child` entry in `ChildTable`)
   - In the next retx attempt of this frame (which can be much later, i.e. after rx of next data poll or next CSL sample window of the child), the frame is reconstructed. 
   - In this case, the same counter, key id, seq number are used in the prepared frame and the frame `aIsRetx` is marked (which inform `SubMac` and/or radio platform to not update these fields).

If we add `mIsHeaderUpdated` and allow radio platform to not set the header fields on certain failures, then we need to add code in the OT stack to handle this case which I don't think we have currently in this PR.
- For example, we need to change the `DataPollHandler::HandleSentFrame()` where we save the info to check whether the fields are set in frame header.
- But we should still remember that we need to retx the frame and that we don't have frame counter, key id for it.
- Then in `DataPollHandler::HandleFrameRequest()` we need to handle all such cases, i.e. it is retx and we do have frame counter, key id saved for it, or it is retx and we don't have info.
- I think it adds unnecessary complexity in the code (and potential for bugs).

I think it would much simpler to have the radio platform layer set the frame counter, key id and seq number header fields when it invokes `TxDone()` (even when tx fails and if the radio platform did not perform security processing) and avoid changing the core implementation.

Regarding CSL Phase I don't think OT stack will use it from `TxDone()` callback so whether it is set of not should not matter.

Thoughts?

----

> Are you aware of any other platform implementing OT_RADIO_CAPS_TRANSMIT_SEC at the moment?

Probably Nordic is the main public one. But there may be private platform implementations. For example, we used to have some internal projects with private/custom platform implementations.

But independent of this I think we want to make sure we have a good/simple radio API model. I think the `radio.h` platform APIs has been slowly becoming more and more complex (many capabilities and APIs). It's a good idea to avoid adding more complexity in radio APIs. Even simple changes can have unexpected consequences. As an example we recently debugged and fixed a tricky and unfortunate bug  from https://github.com/openthread/openthread/pull/6837 which was triggered by a seemingly unrelated and innocent looking change from https://github.com/openthread/openthread/pull/4697 adding "Sleep to Tx" capability.

hubertmis:
> I still think it'd be better to not add `mIsHeaderUpdated` flag and instead ensure that platform layer always update the header.

I'm afraid we cannot force each radio platform to always update the header. If given radio does not update frame counter value if CCA resulted in busy channel, we cannot work around it in radio.c maintaining monotonically incrementing frame counter value.

> Regarding CSL Phase I don't think OT stack will use it from `TxDone()` callback so whether it is set of not should not matter.

If we maintain FrameCounter value for retransmission, we must maintain also CSL Phase value. OT should preserve it, if it can retransmit frames including CSL Phase.

> * For example, we need to change the `DataPollHandler::HandleSentFrame()` where we save the info to check whether the fields are set in frame header.

Is something missing in `DataPollHandler::HandleSentFrame()` patch in this PR?

> But independent of this I think we want to make sure we have a good/simple radio API model.

I agree. However, to support MAC layer features efficiently, we need to add more features to radio platform (like injecting CSL Phase, outgoing security procedure) and align OT implementation with new hardware and drivers.

I think `mIsHeaderUpdated` is simplifying code in OpenThread. Without this flag, one needed to guess if header included correct values or empty placeholders by looking at `mIsRetx` and FSM's states. This guessing could easily lead to errors. New flag introduced in this patch allows the code to rely on information if dynamic data in MAC header is correct. Regardless it is set by OT MAC, or by the radio platform.

abtink:
> Is something missing in DataPollHandler::HandleSentFrame() patch in this PR?

Yes. I think the PR as is may not work. Currently the only change I see in this PR in `DataPollHandler::HandleSentFrame()` is skipping over incrementing the counter and logging (and only in the case of `kErrorNoAck` and not other errors `kErrorChannelAccessFailure` and `kErrorAbort`). The existing code in this method (within the same `switch`) will still read the frame counter and key id from frame and save it in `Child` for retx (which in the case where we have `mIsHeaderUpdated == false` won't be set and will be random/incorrect values in the frame header). 

I think we need the following changes to be added in OT core (copying from comment above) - these are immediate things I could think of (there may be more changes which we need to think of).

> If we add `mIsHeaderUpdated` and allow radio platform to not set the header fields on certain failures, then we need to add code in the OT stack to handle this case which I don't think we have currently in this PR.
> - For example, we need to change the `DataPollHandler::HandleSentFrame()` where we save the info to check whether the fields are set in frame header.
> - But we should still remember that we need to retx the frame and that we don't have frame counter, key id for it.
> - Then in `DataPollHandler::HandleFrameRequest()` we need to handle all such cases, i.e. it is retx and we do have frame counter, key id saved for it, or it is retx and we don't have info.
> - I think it adds unnecessary complexity in the code (and potential for bugs).

-------

> I think mIsHeaderUpdated is simplifying code in OpenThread. Without this flag, one needed to guess if header included correct values or empty placeholders by looking at mIsRetx and FSM's states. This guessing could easily lead to errors. New flag introduced in this patch allows the code to rely on information if dynamic data in MAC header is correct. Regardless it is set by OT MAC, or by the radio platform.

I don't think it's helping make it simpler. It's actually pushing the handling of this situation up the stack rather than handling at lower layers (at `SubMac` or radio platform layer). It's simpler (and I would say safer assumption for future) to require that frames passed back from `Mac` in `HandeSentFrame()/TxDone()` callbacks to the next layers (e.g., to `MeshForwarder`, `DataPollHandler`, etc) are properly formed and have all the fields set in the header.

--------

> I'm afraid we cannot force each radio platform to always update the header. If given radio does not update frame counter value if CCA resulted in busy channel, we cannot work around it in radio.c maintaining monotonically incrementing frame counter value.

I think this can be implemented in the radio platform. 
- Radio platform can track with an internal boolean flag whether it has updated the header or not (e.g., set frame counter)
- Before calling `TxDone()` to pass the frame back to OT stack, it can check this flag and if it the header needs to be updated, then we allocate a new frame counter and populate the frame header.

I think we can require this behavior from radio platform (same way that we expect it from `SubMac`). Any reason/situation why this may not be possible?

src/lib/spinel/radio_spinel_impl.hpp
```diff
@@ -1903,20 +1903,21 @@ otError RadioSpinel<InterfaceType, ProcessContextType>::Transmit(otRadioFrame &a
...
+                                                    SPINEL_DATATYPE_UINT32_S      // TxDelay
+                                                        SPINEL_DATATYPE_UINT32_S, // TxDelayBaseTime
                     mTransmitFrame->mPsdu, mTransmitFrame->mLength, mTransmitFrame->mChannel,
                     mTransmitFrame->mInfo.mTxInfo.mMaxCsmaBackoffs, mTransmitFrame->mInfo.mTxInfo.mMaxFrameRetries,
-                    mTransmitFrame->mInfo.mTxInfo.mCsmaCaEnabled, mTransmitFrame->mInfo.mTxInfo.mIsARetx,
-                    mTransmitFrame->mInfo.mTxInfo.mIsSecurityProcessed, mTransmitFrame->mInfo.mTxInfo.mTxDelay,
-                    mTransmitFrame->mInfo.mTxInfo.mTxDelayBaseTime);
+                    mTransmitFrame->mInfo.mTxInfo.mCsmaCaEnabled, mTransmitFrame->mInfo.mTxInfo.mIsHeaderUpdated,
+                    mTransmitFrame->mInfo.mTxInfo.mIsARetx, mTransmitFrame->mInfo.mTxInfo.mIsSecurityProcessed,
+                    mTransmitFrame->mInfo.mTxInfo.mTxDelay, mTransmitFrame->mInfo.mTxInfo.mTxDelayBaseTime);
```
abtink, line 1920: This changes the spinel format of `SPINEL_PROP_STREAM_RAW ` (adding an extra boolean in middle of its encoding format) so we need to increment  `SPINEL_RCP_API_VERSION`.

I think this will also cause a backward incompatibility between host and RCP.
- Host that switches to this code won't work anymore with RCP using old format.
- Same situation with RCP's running new code but talking to host that is running older code.

In general, it'd be good to avoid this. But if we need to add backward incompatible change, then  we should ensure to not only incremenent `SPINEL_RCP_API_VERSION` but also `SPINEL_MIN_HOST_SUPPORTED_RCP_API_VERSION` (or need to add code on host to handle spinel frame from older version of RCP).

src/ncp/ncp_base_radio.cpp
```diff
@@ -403,6 +404,7 @@ otError NcpBase::DecodeStreamRawTxRequest(otRadioFrame &aFrame)
     aFrame.mInfo.mTxInfo.mMaxCsmaBackoffs     = OPENTHREAD_CONFIG_MAC_MAX_CSMA_BACKOFFS_DIRECT;
     aFrame.mInfo.mTxInfo.mMaxFrameRetries     = OPENTHREAD_CONFIG_MAC_DEFAULT_MAX_FRAME_RETRIES_DIRECT;
     aFrame.mInfo.mTxInfo.mCsmaCaEnabled       = true;
+    aFrame.mInfo.mTxInfo.mIsHeaderUpdated     = false;
```
abtink, line 407: Since we use the new `mIsHeaderUpdated` flag in `TxDone()` callbacks, we need to make sure that RCP report it back to host. I think this is missing in the current implementation. 

We have the change in `RadioSpinel` where we pass the flag down to the RCP for tx (from `SPINEL_CMD_PROP_VALUE_SET(SPINEL_PROP_STREAM_RAW)`) but we also need it to be passed as part of TX DONE event from RCP (in `NcpBase::LinkRawTransmitDone()` used on RCP) and then add code to parse it on host side in `RadioSpinel<InterfaceType, ProcessContextType>::HandleTransmitDone()`.

----

The more I read and think about it, the more I am convinced that adding `mIsHeaderUpdated` flag adds a lot of avoidable complexity with little benefit in return.

hubertmis:
> Any reason/situation why this may not be possible?

If given radio hardware sets frame counter in outgoing frames, radio.c implementation cannot set frame counter value in a way that makes reusing frame counter value impossible. If we implemented setting frame counter in TxDone in radio.c, during execution of such function hardware could transmit Enh-Ack with the same frame counter value.

We should have a single entity in the system setting frame counter to avoid races leading to reusing the same frame counter value. If hardware sets frame counter in outgoing frames, we should not implement setting frame counter value anywhere in the software.

The real device that works like I described above is nRF53. One core is running OpenThread (including radio.c), while the other one is responsible for radio operations, including sending Enh-Ack and setting frame counter in outgoing frames.

abtink:
> The real device that works like I described above is nRF53. One core is running OpenThread (including radio.c), while the other one is responsible for radio operations, including sending Enh-Ack and setting frame counter in outgoing frames.

I agree that one entity should be in charge of assigning frame counter. 

Is there no API or mechanism to request for a frame counter from the entity (whether it is hardware or software (or another core))?  Basically is the only way to get a frame counter is for the frame to be sent over the air?

-------

Here is my view on this:

I still prefer if we avoid the adding of `mIsHeaderUpdated` flags due to all reasons above (mainly the complexity it adds).

But at the end you guys know the hardware and radio driver and if you say, there is no solution that can be done at radio driver level to address this, then I guess we need `mIsHeaderUpdated` to notify the OT stack and handle it in next layer.

If we go for this, then we need more changes in this PR:
-  Changes in `DataPollHandler` and `CslHandler` to handle this flag (mentioned some in above comments).
- We need to address the changes related to this flag in RCP/host code (from comments above) 
   - Also since this would break backward compatibility between RCP and host, we need to decide how to handle it (i.e. require both host and RCP to be updated and run new model, or add code on host to handle both new and old formats).
-  I would also recommend testing this change on hw (may be stress testing in real deployment) since I don't think our CI tests will cover this.

hubertmis:
> Is there no API or mechanism to request for a frame counter from the entity (whether it is hardware or software (or another core))?

For nRF53 it would be possible to create such API, because there is software running in the radio core. However, such API would be prone to race conditions:

1. radio.c requests a frame counter value
2. radio core returns frame counter (value 100) and sends Enh-Ack just after (fc = 101)
3. radio.c requests transmission of a frame with fc=100
4. radio core sends a frame with fc=100 after Enh-Ack with fc=101

If radio.c did not set frame counter in the frame that was not yet transmitted, frame counter would keep incrementing monotonically.

We are aware there is an exception in Thread Spec to allow data or command frames with fc lower than previously transmitted Enh-Ack, but this race condition would also affect other scenarios like responding to Data Poll request frame received during CSMA-CA backoff, where frame counter is expected to keep monotonically incrementing.

Also, for other devices with hardware encryption it might not be possible to retrieve next frame counter value from the hardware.

The rationale I described above was the reason we tried to handle frames after TX attempt that did not set frame counter yet. First attempt with using `mIsARetx` ended with failure (hard to track race conditions). Creating `mIsHeaderUpdated` simplified the code with clear indication if frame counter (and CSL Phase) is set in given frame, or should it be set during next TX attempt.

edmont:
Thanks for your review @abtink, I added your proposed changes.

src/core/thread/csl_tx_scheduler.cpp
```diff
@@ -249,10 +249,16 @@ void CslTxScheduler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError, C
         break;
 
     case kErrorNoAck:
-        aChild.IncrementCslTxAttempts();
+        if (aFrame.IsHeaderUpdated())
+        {
+            // Only consider a transmission attempt the case when the radio platform returns back a
+            // frame with an updated header, i.e.: there was not an abrupt transmission operation
+            // cancelation.
```
abtink, line 256: The `kErrorNoAck` error indicates that the frame was sent but no ack was received (ack timeout).

I think it then make sense to assume/require that with this error, we would never have frame header not updated.
(if frame is sent over the air, then all fields should be set on it).

If the frame is not sent over the air at all, then we can have other error codes to indicate this:
- `kErrorChannelAccessFailure` to indicate we tried to access channel but channel was busy.
- `kErrorAbort` to indicate frame tx was abruptly canceled/aborted.

I think we just need to check and handle `IsHeaderUpdate()` under the these errors and undere `kErrorNoAcks` we can consider adding an assert ensuring `IsHeaderUpdated() == true`.

src/core/thread/csl_tx_scheduler.cpp
```diff
@@ -274,7 +280,7 @@ void CslTxScheduler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError, C
         {
             aChild.SetIndirectDataSequenceNumber(aFrame.GetSequence());
```
abtink, line 281: The sequence number may not be set as well if `IsHeaderUpdated()` is false.

abtink:
> For nRF53 it would be possible to create such API, because there is software running in the radio core. However, such API would be prone to race conditions:
> 
> - radio.c requests a frame counter value
> - radio core returns frame counter (value 100) and sends Enh-Ack just after (fc = 101)
> - radio.c requests transmission of a frame with fc=100
> - radio core sends a frame with fc=100 after Enh-Ack with fc=101
> 
> If radio.c did not set frame counter in the frame that was not yet transmitted, frame counter would keep incrementing monotonically.
> 

@hubertmis , the race (counter mismatch between data and enh-ack frames) may happen in different situations and independent of having or not having an API to allocate a frame counter on "TxDone". For example, during retx of a frame 
- Frame is about to be sent (first time), we assign `fc=100`, encrypt it, and then send it.
- We get no ack, so need to retx the frame so we start the next attempt.
- While performing CSMA-CA for the next tx attempt, we happen to receive a frame (from same neighbor) and then prepare and send an enh-ack using `fc=101`.
- Now we are sending a data frame with `fc=100` after and enh-ack frame of `fc=101` (and we cannot re-assign/change `fc` on a retx frame).

I think the above situation is what you meant by this. Is this correct?
> this race condition would also affect other scenarios like responding to Data Poll request frame received during CSMA-CA backoff, where frame counter is expected to keep monotonically incrementing.


As you mentioned, this is already addressed in a Thread spec change on how the enh-ack frame counter is managed on receiver side by allowing enh-ack frame counter on receiver to be ahead of the "data" frame counter (basically we try to track an `mLinkAckFrameCounter` separately from `mLinkFrameCounters` per `Neighbor`). 
- I think it was added in OT from https://github.com/openthread/openthread/pull/5583.

So we only require that the frame counters are monotonically increasing for "data" frame tx per neighbor (and separately for "enh-ack" frames).

Since the data frames are always prepared and given to radio platform by the OT stack, we should be fine to allocate and assign a frame counter to a frame even in the case where it was never sent (basically use such an API to request a frame counter to be allocated before `TxDone`).
- Then OT stack will either try to re-send the same frame again, e.g. if it is indirect frame tx (where we do want to use the same frame counter for retx)
- Or ignore/drop this data frame (the assigned frame counter is wasted but that's ok). Any future data frame will get a new counter.

src/lib/spinel/radio_spinel_impl.hpp
edmont: Implemented.

src/ncp/ncp_base_radio.cpp
edmont: Implemented.

src/core/thread/csl_tx_scheduler.cpp
edmont: Agree, added the assert.

src/core/thread/csl_tx_scheduler.cpp
edmont: I don't think that's the case since the radio platform never takes responsibility of updating the sequence numbers. Doc for `mIsHeaderUpdated` states:

> If `mIsHeaderUpdated` is not set, then the frame counter and key id are not set in the frame by OpenThread core and it is the responsibility of the radio platform to assign them. The platform should update the frame (assign counter and key id) even if the transmission gets aborted or fails (e.g., channel access error).

hubertmis:
@abtink I agree. I was thinking about more generic usage of 802.15.4. It seems that Thread specification, and especially OpenThread implementation prevents decrementing frame counter value per link (with the exception of Enh-Ack discussed above). It does not prevent decrementing frame counter between two links (I think it would be possible to send CSL transmission to peer A, followed by direct retransmission with lower frame counter value to peer B). But it's fine, because in Thread we always maintain single link per peer.

I'm investigating alternative solution to handle setting MAC header data in radio.c when radio driver or hardware skipped setting it. But I see a few problems:

I'm looking at 802.15.4 specification to find a way to abstract "atomically get and increment frame counter" operation in systems like Zephyr RTOS. I see an obstacle here, because according to spec we can use separated MLME-GET.request and MLME-SET.request, what is not atomic. If OpenThread required radio platform to always update frame counter value, even if given radio driver did not do that, given platform would need to be extended with proprietary `frame_counter_get_and_increment()` feature. Alternatively, we could disable transceiver while radio.c calls MLME-GET - MLME-SET sequence, but it would degrade the performance of the radio.

Also, if radio.c were to set frame counter in the frame, radio.c would need to set CSL IE as well (or we would need to handle cases where MAC header is partially updated, what would be even more complicated). I'm afraid radio.c does not have enough information to set correct CSL values. It would need to extract CSL Period from radio driver (or hardware) and estimate Phase. To estimate Phase value, radio.c would need to depend on a timer what would be one more complication.

I enjoy this discussion, looking at the problem and solutions from multiple angles. I hope the patch proposed in this PR is an acceptable solution.

edmont:
Failing CI tests are passing locally for me.

src/core/thread/csl_tx_scheduler.cpp
abtink: You are correct. 

But there is situation we may need to think of:
- What if the seq-num is set but the frame-counter/key-id is not (mac always sets the seq number, but radio platform may or may not update the frame-counter/key-id
- Do we need to track this situation and ensure on next retx we re-use the same seq-num but somehow mark that we don't yet have the frame-counter/key-id and need the sub-mac or radio platform to assign them.
- Need to think more on this.

abtink:
Thanks @edmont for update.

I went through the whole code in detail. I think there are some situations and code paths that may not covered and need to be updated to use the new `IsHeaderUpdated` flag properly.

As I said adding a new flag sounds simple, but it can be very tricky and we need to be super careful of all the different situations.  I will try my best to help with this  😃 ... But to be honest, I am still afraid that this change will break some functionality.

-------

Some general observations (which help with the analysis below):
- Before sending or preparing a new frame to tx, we always `Clear()` the frame which init the flags (it sets `mIsRetx` and `mIsHeaderUpdated` to `false`). This is happening from `Mac::BeginTransmit()` and call to `txFrames.Clear()`.
- We can have OT handle tx security (done at `SubMac`) or be delegated to radio platform
- We can have OT handle retx of frame (again done at `SubMac`) or handled by radio platform layer.
- Indirect frame retx preparation (data poll or CSL based) are handled by the `DataPollHandler` and `CslTxScheduler`.
  - They save the seq num, key id and frame counter on failed tx attempts, and ensure to set the same values when preparing the frame again for retx (which can happen much later on fresh new frame instance).

-----------


Here are some cases that I found so far that the current code in this PR may not function properly:

**1: Indirect retx of a secure frame**
- When we prepare an indirect tx frame for a second (or next) tx attempt (from `DataPollHandler::HandleFrameRequest()` we set the seq-num, frame counter and key id but do not mark `mIsHeaderUpdated`.
- Then the `SubMac` or radio platform will overwrite the values in the header which would be incorrect
- How does it work in `main`: We set `frame->SetIsARetransmission(true)` in the `DataPollHandler::HandleFrameRequest()`
- **Fix**: We need to do the same for `mIsHeaderUpdated` in `DataPollHandler::HandleFrameRequest()`

--------

**2: Indirect retx of a secure CSL frame**
- The same situation as above can happen with CSL retx
- **Fix**: We need to update `mIsHeaderUpdated` in `CslTxScheduler::HandleFrameRequest()` if frame is a retx.

--------

**3: Indirect re-tx (CSL or data poll) of secure frame due to "NoAck" error when `SubMac` handles the security**
- New frame is prepared and `mFrameIsUpdated` starts as `false`
-  `SubMac` handles security, it will actually set frame counter and key id, but I don't see any place in the code where we would set the `mFrameIsUpdated` on the frame.
- So `mFrameIsUpdated` remains `false`.
- We send the frame and happen to get no ack error.
- The error is propagated through all "TxDone()" callbacks till we get to `DataPollHandler::HandleSentFrame()`.
- Now under `case kErrorNoAck` we have `OT_ASSERT(aFrame.IsHeaderUpdated());` which would assert.
- Same situation can happen with CSL and from `CslTxScheduler::HandleSentFrame()`
- **Fix**: we need to make sure `SubMac`/`Mac` set the `mIsHeaderUpdated` when they do set the frame counter on frame.

---------

**4: Indirect re-tx (CSL or data poll based) of _non-secure_ frame with "NoAck" error**
- Frame does not use link layer security (example of this is an MLE message)
- So there is no need to add frame counter, key id.
- New frame is prepared and `mFrameIsUpdated` starts as `false`
   - If `SubMac` handles security, I don't see any place where we set it to `true` 
   - But even if radio platform handles security, it may not  mark the flag, since it won't need to add frame counter, key id (these fields are not even used or present in the frame).
- We send the frame, happen to get no ack error
- The error is propgated through all "TxDone()" callbacks, till we get to ``DataPollHandler::HandleFrameRequest()`
- Now again under `case kErrorNoAck` we have `OT_ASSERT(aFrame.IsHeaderUpdated());` which would assert.
- -Same situation under CSL from ``CslTxScheduler::HandleSentFrame()`
- **Fix**: We need to relax the assert check here and check this only when `aFrame.GetSecurityEnabled()`

---------

**5: Direct frame retx when `SubMac` takes care of re-tx and security**
- The frame starts `mIsHeaderUpdated` as `false`.
- The frame will be prepared and then `SubMac` will notice see the `mIsHeaderUpdated` is false and sets the frame counter and key id, but the `mIsHeaderUpdated` remains `false`
- If the frame tx fails, `SubMac` will retry the frame but the `mIsHeaderUpdated` remains `false`
- I don't see any immediate place this can cause issue, but not good idea to keep it like this.
- How does it work in `main`: In `SubMac::HandleTransmitDone()` when we decide to retx the frame we do `aFrame.SetIsARetransmission(true)`.
- **Fix**: Better set the `mIsHeaderUpdated` when `SubMac`/`Mac` set the frame counter and key id. Same fix as in `3`.

-------

**6: Indirect retx of _non-secure_ frame**
- In this case we still need to save seq-num from first attempt to reuse same seq-num when preparing same frame for retx
- But there is no frame counter or key id (since no link layer security)
- The `Mac::BeginTransmit()` in this PR is changed to use `frame->IsHeaderUpdated()` to determine whether to set the sequence number or not `frame->SetSequence(mDataSequence++);`
- **Fix**: So this means when we prepare the frame for retx from `DataPollHandler::HandleFrameRequest() or `CslTxScheuler::HandleFrameRequest()` we need to set the `mIsHeaderUpdated even when the frame does not use security.
- Now what if the frame is secure and the radio platform is handling the security processing.
- On the first tx attempt we can have the following situation: `Mac` will always set the seq number on frame, but now suppose the frame tx is aborted and radio platform does not update the header.
- In the `DataPollHandler::HandleSentFrame()` we will save the seq-number but since `mIsHeaderUpdated` remains `false` we don't save the frame counter and or key id (nothing to save)
- However, I think this situation works ok:
  - In this situation the `aChild.IncrementIndirectTxAttempts()` won't be called and we remain at attempt counter of zero, 
  - So the next time we want to re-send the frame, we don't treat as a retx and will keep `mIsHeaderUpdated` as `false`
  - This means `Mac` will re-assign a new seq number to the frame but that should be fine (we just waste one seq number)
  - This is fine for now, but we need to be aware of this behavior for any future changes (I know there are some discussions in Thread TC about using seq number for some situations).

src/core/thread/csl_tx_scheduler.cpp
abtink: Added more details about this in https://github.com/openthread/openthread/pull/6849#issuecomment-893858383 at item no 6.

edmont:
@abtink thanks for the thoughtful re-review :)

**1, 2, 3, 5**
I believe you missed that this PR updates `Frame::SetFrameCounter()` to also set the `mIsHeaderUpdated` to `true`. It's done like that for code optimization reasons, but let me know if you'd prefer to move this setting to all the places you described.

**4**
Agree, done.

**6**
Actually relying on `frame->IsHeaderUpdated()` for learning about the sequence number status doesn't seem consistent with the flags documentation. Changed back to check `frame->IsARetransmission()`.

abtink:
Thanks @edmont.

I did indeed miss the change in `SetFrameCounter()`. It does address 1, 2, 3, 5 above, and I think it is fine to set the flag from this method.

I actually did checkout this branch and did a `git grep SetIsHeaderUpdated` to find all places from which the flag is set (but did not think that the `mIsHeaderUpdated` may be being set directly so did not search for it 😃). May be better to update  the `SetFrameCounter()` method to use the `SetIsHeaderUpdated()` as well?

For **6**, I think going back to using `IsARetransmission()` should work fine and I agree is a better model.

include/openthread/platform/radio.h
```diff
@@ -239,24 +239,26 @@ typedef struct otRadioFrame
...
+             * from the previous attempts of the same frame. The platform should not assign or change the frame counter
+             * (but may still need to perform security processing depending on `mIsSecurityProcessed` flag).
              *
-             * If `mIsARetx` is not set, then the frame counter and key id are not set in the frame by OpenThread core
-             * and it is the responsibility of the radio platform to assign them. The platform should update the frame
-             * (assign counter and key id) even if the transmission gets aborted or fails (e.g., channel access error).
+             * If `mIsHeaderUpdated` is not set, then the frame counter and key id are not set in the frame by
+             * OpenThread core and it is the responsibility of the radio platform to assign them. The platform should
+             * update the frame (assign counter and key id) even if the transmission gets aborted or fails (e.g.,
+             * channel access error).
```
abtink, line 257: Should we update this commit since it is no longer required to set them
```suggestion
             * OpenThread core and it is the responsibility of the radio platform to assign them. 
```

Also how about adding this to indicate the optionality of it and that radio platform must updae the flag when/if it updates the header:
```
If `mIsHeaderUpdated` is not set, then the frame counter and key id are not set in the frame by
OpenThread core and it is the responsibility of the radio platform to assign them. The platform
must update the frame header(assign counter and key id) before sending the frame over the air,
however if the the transmission gets aborted and the frame is never sent over the air(e.g., channel
access error) the platform may choose to not update the header. However, if the platform updates
the header, it must also set this flag before passing the frame back from the `otPlatRadioTxDone()`
callback.
```

src/core/thread/csl_tx_scheduler.cpp
```diff
@@ -249,8 +249,9 @@ void CslTxScheduler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError, C
         break;
 
     case kErrorNoAck:
-        aChild.IncrementCslTxAttempts();
+        OT_ASSERT(aFrame.GetSecurityEnabled() == aFrame.IsHeaderUpdated());
```
abtink, line 252: How about relaxing the check as this?
```suggestion
        OT_ASSERT(!aFrame.GetSecurityEnabled() || aFrame.IsHeaderUpdated());
```

In the context here, if frame security is not enabled, we  don't care about `IsHeaderUpdate()`. 

The reason I am considering for relaxing this is we may end up using `IsHeaderUpdated` for CSL IE  with MLE frame in future (which would not then use link security).

src/core/mac/data_poll_handler.cpp
```diff
@@ -242,8 +242,9 @@ void DataPollHandler::HandleSentFrame(const Mac::TxFrame &aFrame, Error aError,
         break;
 
     case kErrorNoAck:
-        aChild.IncrementIndirectTxAttempts();
+        OT_ASSERT(aFrame.GetSecurityEnabled() == aFrame.IsHeaderUpdated());
```
abtink, line 245: Similar suggestion here.

abtink (approved):
Thanks @edmont. Overall LGTM. 👍 

Few smaller suggestions below.

include/openthread/platform/radio.h
edmont: Done.

src/core/mac/data_poll_handler.cpp
edmont: Done.

src/core/thread/csl_tx_scheduler.cpp
edmont: Done.

abtink (approved):
Looks great! Thanks @edmont. 👍

hubertmis (approved):
Implementation looks good! :+1: 
Found two nits in code comments.

hubertmis (approved):

jwhui (approved):
Thanks! 👍🏼

PR #6847, opened 2021-07-23
[border agent] [WIP] publish meshcop service when thread is disabled (part 1/3): Add otBorderAgentStart() API

superwhd:
This PR is part 1 of the fix of https://github.com/openthread/ot-br-posix/issues/736.

Part 1: https://github.com/openthread/openthread/pull/6847
Part 2: https://github.com/openthread/ot-br-posix/pull/942
Part 3: https://github.com/openthread/openthread/pull/6850 (locally tested, to be updated after https://github.com/openthread/openthread/pull/6865 is merged)

Depends-On: openthread/ot-br-posix#942

PR #6850, opened 2021-07-23
[border agent] [WIP] publish meshcop service when thread is disabled (part 3/3): 

superwhd:
This PR is part 3 of the fix of https://github.com/openthread/ot-br-posix/issues/736.

In this PR, we introduce a new test case to verify that OTBR publishes the meshcop service with proper TXT data.

Part 1: https://github.com/openthread/openthread/pull/6847
Part 2: https://github.com/openthread/ot-br-posix/pull/942
Part 3: https://github.com/openthread/openthread/pull/6850

### DIFF
--- a/src/core/api/border_agent_api.cpp
+++ b/src/core/api/border_agent_api.cpp
@@ -56,4 +56,11 @@ uint16_t otBorderAgentGetUdpPort(otInstance *aInstance)
     return instance.Get<MeshCoP::BorderAgent>().GetUdpPort();
 }
 
+void otBorderAgentStart(otInstance *aInstance)
+{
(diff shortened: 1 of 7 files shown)


src/core/api/border_agent_api.cpp
```diff
@@ -56,4 +56,11 @@ uint16_t otBorderAgentGetUdpPort(otInstance *aInstance)
     return instance.Get<MeshCoP::BorderAgent>().GetUdpPort();
 }
 
+void otBorderAgentStart(otInstance *aInstance)
```
simonlingoogle, line 59: Why do we need this API? Why not start BorderAgent right after OT Instance construction (maybe in `MeshCoP::BorderAgent` constructor).

PR #6853, opened 2021-07-25
[routing-manager] add `Array` & use it for OMR/on-link prefix list

abtink:

**[routing-manager] Use `Array` for list of OMR or on-link prefixes**

This commit simplifies the `RoutingManager` class by using `Array`
to store lists of new or advertised OMR/on-link prefixes.

**[common] adding `Array` class**

This commit adds a generic `Array<Type, kMaxSize>` class which
represents a fixed size array of elements of template `Type` which
tracks its current length. Entries can be appended to the array
(`PushBack()`) and removed from the end of array (`PopBack()`). This
simple data model is used in some other modules, so having a generic
version can help simplify the code. This commit also adds a unit test
to verify `Array` methods.

### DIFF
--- a/src/core/common/array.hpp
+++ b/src/core/common/array.hpp
@@ -1,0 +1,326 @@
+/*
+ *  Copyright (c) 2021, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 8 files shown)


src/core/common/array.hpp
```diff
@@ -0,0 +1,312 @@
...
+     * This method indicates whether or not the array is empty.
+     *
+     * @retval TRUE when array is empty.
+     * @retval FALSE when array is not empty.
+     *
+     */
+    bool IsEmpty(void) const { return (mLength == 0); }
+
+    /**
+     * This method indicates whether or not the array is full
```
wgtdkp, line 80: nit:
```suggestion
     * This method indicates whether or not the array is full.
```

wgtdkp (approved):
Thanks @abtink ! 👍 
I believe this Class will be useful to many other modules!

src/core/common/array.hpp
```diff
@@ -0,0 +1,325 @@
...
+
+    /**
+     * This method returns the maximum array size (max number of elements).
+     *
+     * @returns The maximum array size (max number of elements that can be added to the array).
+     */
+    uint16_t GetMaxSize(void) const { return kMaxSize; }
+
+    /**
+     * This method return the current length of array (number of elements).
```
jwhui, line 106:
```suggestion
     * This method returns the current length of array (number of elements).
```

src/core/common/array.hpp
```diff
@@ -0,0 +1,325 @@
...
+     * @retval FALSE when array is not full.
+     *
+     */
+    bool IsFull(void) const { return (mLength == kMaxSize); }
+
+    /**
+     * This method returns the maximum array size (max number of elements).
+     *
+     * @returns The maximum array size (max number of elements that can be added to the array).
+     */
```
jwhui, line 102:
```suggestion
     * @returns The maximum array size (max number of elements that can be added to the array).
     *
     */
```

src/core/common/array.hpp
```diff
@@ -0,0 +1,325 @@
...
+     *
+     * @returns The current array length.
+     *
+     */
+    uint16_t GetLength(void) const { return mLength; }
+
+    /**
+     * This method overloads the `[]` operator to get the element at a given index.
+     *
+     * This method does not perform index bounds checking. Behavior is undefined if the @p aIndex is not valid.
```
jwhui, line 116:
```suggestion
     * This method does not perform index bounds checking. Behavior is undefined if @p aIndex is not valid.
```

src/core/common/array.hpp
```diff
@@ -0,0 +1,325 @@
...
+     *
+     * @returns A reference to the element in array at @p aIndex.
+     *
+     */
+    Type &operator[](uint16_t aIndex) { return mElements[aIndex]; }
+
+    /**
+     * This method overloads the `[]` operator to get the element at a given index.
+     *
+     * This method does not perform index bounds checking. Behavior is undefined if the @p aIndex is not valid.
```
jwhui, line 128:
```suggestion
     * This method does not perform index bounds checking. Behavior is undefined if @p aIndex is not valid.
```

tests/unit/test_array.cpp
```diff
@@ -0,0 +1,255 @@
+/*
+ *  Copyright (c) 2019, The OpenThread Authors.
```
jwhui, line 2:
```suggestion
 *  Copyright (c) 2021, The OpenThread Authors.
```

jwhui (approved):

PR #6857, opened 2021-07-27
[cli] add README for TCP CLI tool

samkumar:
This is the fourth in a series of pull requests (after #6744, #6770, and #6790) bringing TCPlp to OpenThread. I'm making a series of small pull requests so that it's easier for the OpenThread community to review the code. If you'd like to preview a working version of TCPlp in OpenThread, see #6650.

This pull request adds a README_TCP.md file containing documentation for the TCP CLI functionality. This was requested in the review of Pull Request #6770.

### DIFF
--- a/src/cli/README_TCP.md
+++ b/src/cli/README_TCP.md
@@ -32,17 +32,20 @@ On node 2, initialize the TCP CLI module, connect to node 1, and send a simple m
 ### Result
 
 After running the `tcp connect` command on node 1, you should see a printout on node 1 similar to below:
+
 ```bash
 TCP: Connection established
 ```
 
 In addition, you should also see a printout on node 2 similar to below:
+
 ```bash
 Accepted connection from [fe80:0:0:0:8f3:f602:bf9b:52f2]:49152
 TCP: Connection established
 ```
 
 After running the `tcp send` command on node 1, you should see a printout on node 2 similar to below:
+
 ```bash
 TCP: Received 5 bytes: hello
 ```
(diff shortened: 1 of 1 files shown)


src/cli/README_TCP.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+### Result
+
+After running the `tcp connect` command on node 1, you should see a printout on node 1 similar to below:
+```bash
+TCP: Connection established
+```
+
+After running the `tcp connect` command on node 1, you should see a printout on node 2 similar to below:
```
jwhui, line 39:
```suggestion
After running the `tcp connect` command on node 2, you should see a printout on node 2 similar to below:
```

samkumar: I think the original text is correct. You only ever run `tcp connect` on node 1 in that example. On node 2, you run `tcp listen` but not `tcp connect`.

jwhui: Ah, yes, I read it too quickly. I think it would be slightly more natural to say something like:
```suggestion
In addition, you should also see a printout on node 2 similar to below:
```
Thoughts?

jwhui:
The pretty check is failing. You can use `script/make-pretty markdown` to automatically reformat.

src/cli/README_TCP.md
samkumar: I'm happy with this. I'll update the pull request.

samkumar:
I have incorporated the change you suggested and run `script/make-pretty markdown`.

jwhui (approved):
Thanks! 👍🏼

PR #6858, opened 2021-07-27
[tests] attempt to fix intermittent failures of MATN-12

superwhd:
The test failed because the script failed to read the output of `ipaddr rloc`. I suspect it can be fixed by adding some wait time at the end of the test.

### DIFF
--- a/tests/scripts/thread-cert/border_router/MATN_12_HopLimitProcessing.py
+++ b/tests/scripts/thread-cert/border_router/MATN_12_HopLimitProcessing.py
@@ -151,6 +151,7 @@ class MATN_12_HopLimitProcessing(thread_cert.TestCase):
         self.collect_rloc16s()
         self.collect_rlocs()
         self.collect_extra_vars()
+        self.simulator.go(5)
 
     def verify(self, pv: pktverify.packet_verifier.PacketVerifier):
         pkts = pv.pkts
--- a/tests/scripts/thread-cert/border_router/MATN_12_HopLimitProcessing.py
+++ b/tests/scripts/thread-cert/border_router/MATN_12_HopLimitProcessing.py
@@ -151,7 +151,6 @@ class MATN_12_HopLimitProcessing(thread_cert.TestCase):
         self.collect_rloc16s()
         self.collect_rlocs()
         self.collect_extra_vars()
-        self.simulator.go(5)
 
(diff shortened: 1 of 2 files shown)


tests/scripts/thread-cert/border_router/MATN_12_HopLimitProcessing.py
```diff
@@ -151,6 +151,7 @@ def test(self):
         self.collect_rloc16s()
         self.collect_rlocs()
         self.collect_extra_vars()
+        self.simulator.go(5)
```
simonlingoogle, line 154: It won't be useful adding delay after `collect_*` methods. 

If it's true that `ipaddr rloc` failed to get the command output within the timeout, should we increase the command timeout in simulations?

superwhd: I agree the delay in `collect_*` won't help. However, I don't have much idea how this test case is related to simulation since  `VIRTUAL_TIME` is set to 0 in border router test cases. Could you send a PR when you get a chance?

simonlingoogle: Without `VIRTUAL_TIME`, it's real-time simulation. The default tiemout is 10, which may not be enough in Docker Runners. 
```
self.pexpect = pexpect.popen_spawn.PopenSpawn(cmd, timeout=10)
```

superwhd: Thanks! PR updated.

tests/scripts/thread-cert/node.py
```diff
@@ -135,7 +135,7 @@ def _launch_docker(self):
         assert launch_ok
 
         cmd = f'docker exec -i {self._docker_name} ot-ctl'
-        self.pexpect = pexpect.popen_spawn.PopenSpawn(cmd, timeout=10)
+        self.pexpect = pexpect.popen_spawn.PopenSpawn(cmd, timeout=15)
```
simonlingoogle, line 138:
```suggestion
        self.pexpect = pexpect.popen_spawn.PopenSpawn(cmd, timeout=30)
```
why not even larger?

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #6868, opened 2021-07-29
[dataset] add API `otDatasetUpdaterRequestUpdateTlvs`

jinran-google:
This adds an API `otDatasetUpdaterRequestUpdateTlvs`, which is the same as `otDatasetUpdaterRequestUpdate` except that it accepts an `otOperationalDatasetTlvs` as a parameter.

### DIFF
--- a/include/openthread/dataset_updater.h
+++ b/include/openthread/dataset_updater.h
@@ -94,6 +94,16 @@ otError otDatasetUpdaterRequestUpdate(otInstance *                aInstance,
                                       void *                      aContext);
 
 /**
+ * See otDatasetUpdaterRequestUpdate.
+ *
+ * @param aDatasetTlvs A pointer to the Dataset TLVs containing the fields to change.
+ */
+otError otDatasetUpdaterRequestUpdateTlvs(otInstance *                    aInstance,
+                                          const otOperationalDatasetTlvs *aDatasetTlvs,
+                                          otDatasetUpdaterCallback        aCallback,
+                                          void *                          aContext);
+
+/**
  * This function cancels an ongoing (if any) Operational Dataset update request.
  *
  * @param[in]  aInstance         A pointer to an OpenThread instance.
--- a/include/openthread/dataset_updater.h
+++ b/include/openthread/dataset_updater.h
@@ -94,9 +94,22 @@ otError otDatasetUpdaterRequestUpdate(otInstance *                aInstance,
                                       void *                      aContext);
 
 /**
- * See otDatasetUpdaterRequestUpdate.
+ * This function requests an update to Operational Dataset.
  *
- * @param aDatasetTlvs A pointer to the Dataset TLVs containing the fields to change.
+ * @p aDataset should contain the fields to be updated and their new value. It must not contain Active or Pending
+ * Timestamp fields. The Delay field is optional, if not provided a default value (1000 ms) would be used.
+ *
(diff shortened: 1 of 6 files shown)


src/cli/cli_dataset.cpp
```diff
@@ -912,7 +912,22 @@ otError Dataset::ProcessUpdater(Arg aArgs[])
     }
     else if (aArgs[0] == "start")
     {
-        error = otDatasetUpdaterRequestUpdate(mInterpreter.mInstance, &sDataset, &Dataset::HandleDatasetUpdater, this);
+        if (aArgs[1] == "tlvs")
```
jwhui, line 915: Update [src/cli/README_DATASET.md](https://github.com/openthread/openthread/blob/main/src/cli/README_DATASET.md)?

include/openthread/dataset_updater.h
```diff
@@ -93,6 +93,29 @@ otError otDatasetUpdaterRequestUpdate(otInstance *                aInstance,
                                       otDatasetUpdaterCallback    aCallback,
                                       void *                      aContext);
 
+/**
+ * This function requests an update to Operational Dataset.
+ *
+ * @p aDataset should contain the fields to be updated and their new value. It must not contain Active or Pending
```
wgtdkp, line 99:
```suggestion
 * @p aDatasetTlvs should contain dataset fields as encoded TLVs. It must not contain Active or Pending
```

wgtdkp (approved):
LGTM :+1:

src/cli/cli_dataset.cpp
jinran-google: Done.

src/cli/cli_dataset.cpp
```diff
@@ -912,7 +912,22 @@ otError Dataset::ProcessUpdater(Arg aArgs[])
...
+        {
+            MeshCoP::Dataset         dataset;
+            otOperationalDatasetTlvs datasetTlvs;
+
+            SuccessOrExit(error = dataset.SetFrom(*static_cast<MeshCoP::Dataset::Info *>(&sDataset)));
+            dataset.ConvertTo(datasetTlvs);
+
```
abtink, line 922: When or why would the CLI user want to use the `tlvs` command? 

This sub-command is converting `sDataset` to TLV sequence which is then again converted back to struct format by the implementation. To the CLI user this behaves exactly as `update start`. 

Also, in gneral we want to avoid using core internal types and definitions in CLI (e.g., `MeshCoP::Dataset` or `SetFrom()` and `ConvertTo()`). The CLI should only see and use public OT APIs and types.

abtink (commented):
Overall looks good. 

Few suggestions and questions below. Thanks

abtink:
@jinran-google just noticed the PR is now closed. Did you intend to close it?

jinran-google:
> @jinran-google just noticed the PR is now closed. Did you intend to close it?

Yes. Because, when doing migrations, we think it might be better to use `MGMT_ACTIVE_SET` instead.

abtink:
> Yes. Because, when doing migrations, we think it might be better to use MGMT_ACTIVE_SET instead.

Sounds good. The `DatasetUpdater` is mainly intended for cases where we want to change one or a  subset of dataset fields (e.g., just change the channel or just change PSKc, etc). If we want to change whole dataset other APIs (like `ACTIVE_SET`) would be better.

PR #6870, opened 2021-07-29
[tests] simplify border router tests

wgtdkp:
This commit simplifies the border router tests by replacing
`self.nodes[<NODE_ID>]` with shorter names.

### DIFF
--- a/tests/scripts/thread-cert/border_router/test_dnssd_server.py
+++ b/tests/scripts/thread-cert/border_router/test_dnssd_server.py
@@ -48,8 +48,6 @@ SERVER = BR1 = 1
 CLIENT1, CLIENT2 = 2, 3
 HOST = 4
 
-DIGGER = HOST
-
 DOMAIN = 'default.service.arpa.'
 SERVICE = '_testsrv._udp'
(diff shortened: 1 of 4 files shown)


superwhd (approved):
LGTM :+1:

tests/scripts/thread-cert/node.py
```diff
@@ -939,6 +939,10 @@ def srp_client_start(self, server_address, server_port):
         self.send_command(f'srp client start {server_address} {server_port}')
         self._expect_done()
 
+    def srp_client_autostart(self):
+        self.send_command('srp client autostart enable')
+        self._expect_done()
+
```
abtink, line 945: I think we don't need this. We already have  `srp_client_enable_auto_start_mode()` which does the same.

abtink (approved):
Looks great. Thanks for changing this. 👍

tests/scripts/thread-cert/node.py
wgtdkp: Thanks! Didn't notice that.

jwhui (approved):

PR #6872, opened 2021-07-29
[array] select the index/length `uint` type based on array size

abtink:
This commit enhances the template `Array` class implementation to
allow the `uint` type used for array size (which is also used for
length and index) to be specified as a template type parameter. If
not explicitly specified, a default `uint` type is selected based on
the array size `kMaxSize`, i.e., if `kMaxSize <= 255` then `uint8_t`
will be used, otherwise `uint16_t` will be used.

------------

Added a second related commit in this PR:

**[array] add `FindMatching()` and `ContainsMatching()` (#6872)**

This commit adds template `Find/ContainsMatching<Indicator>()` methods
which search in the array to find the first element matching a given
indicator. Unit test `test_array` is also updated to check the
behavior of newly added methods.

### DIFF
--- a/src/core/common/type_traits.hpp
+++ b/src/core/common/type_traits.hpp
@@ -104,6 +104,27 @@ template <typename Type> struct IsSame<Type, Type> : public TrueValue
 {
 };
 
+/**
+ * This type selects between two given types based on a boolean condition at compile time.
+ *
+ * It provides member type named `Type` which is defined as `TypeOnTrue` if `kCondition` is `true` at compile time, or
+ * as `TypeOnFalse` if `kCondition` is `false`.
+ *
+ * @tparam kCondition   The boolean condition which is used to select between the two types.
+ * @tparam TypeOnTrue   The type to select when `kCondition` is `true`.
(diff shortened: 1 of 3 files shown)


wgtdkp (approved):
LGTM :+1:

src/core/common/type_traits.hpp
```diff
@@ -104,6 +104,27 @@ template <typename Type> struct IsSame<Type, Type> : public TrueValue
 {
 };
 
+/**
+ * This type selects between two given types based on a boolean condition at compile time.
+ *
+ * It provides member type named `Type` which is defined as `TypeOnTrue` if `kCondition` is `true` at compile time, or
+ * as `TypeOnFalse` if `kCondition` is `false`.
+ *
+ * @tparam kCondidtion  The boolean condition which is used to select between the two types.
```
jwhui, line 113:
```suggestion
 * @tparam kCondition   The boolean condition which is used to select between the two types.
```

simonlingoogle (approved):
LGTM 👍

jwhui (approved):

PR #6876, opened 2021-07-30
[mle] add delay before orphan announce

librasungirl:
`mAnnounceDelay` is about 80ms by default, seems too short to receive possible parent response for the third parent request before announce. 

Here adopts the same `kParentRequestReedTimeout` delay.

### DIFF
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -1780,7 +1780,7 @@ void Mle::HandleAttachTimer(void)
             SetAttachState(kAttachStateAnnounce);
             IgnoreError(SendParentRequest(kParentRequestTypeRoutersAndReeds));
             mAnnounceChannel = Mac::ChannelMask::kChannelIteratorFirst;
-            delay            = mAnnounceDelay;
+            delay = kParentRequestReedTimeout;
             break;
         }
 


src/core/thread/mle.cpp
```diff
@@ -1780,7 +1780,7 @@ void Mle::HandleAttachTimer(void)
             SetAttachState(kAttachStateAnnounce);
             IgnoreError(SendParentRequest(kParentRequestTypeRoutersAndReeds));
             mAnnounceChannel = Mac::ChannelMask::kChannelIteratorFirst;
-            delay            = mAnnounceDelay;
+            delay = kParentRequestReedTimeout;
```
abtink, line 1783: @librasungirl what is the background of this change?

I think the existing code is correct and this change will make extra long wait time between first announce send and the next one on the next channel.
- Note that we stay in `kAttachStateAnnounce` state for multiple iterations (of `AttachTimer` firing), each time sending announce on a different channel.
- Each iteration uses `mAnnounceDelay` wait before sending the next announce (on the next channel).
- `mAnnounceDelay` is calculated as `kAnnounceTimeout / (channelMask.GetNumberOfChannels() + 1)`
   - We also have a check where if the calculated `mAnnounceDelay` is too short (less than `kMinAnnounceDelay`) we then use the min value (which `80` msec) between two announce txs on different channels.
- This means overall we wait at least for `kAnnounceTimeout` which is `1400` msec and is longer than `kParentRequestReedTimeout` (which is `1250` msec).

jwhui: Is the `SendParentRequest()` call two lines up appropriate? It seems to me the existing code is sending an MLE Parent Request and then delaying `mAnnounceDelay` time before calling `SendOrphanceAnnounce()`.

abtink: Yes. That's intentional and I think is correct. This is idea behind the existing code. Each "attach attempt" iteration consists of 3 phases:
- We have the first phase where we send "Parent Request to routers only" and wait for a response.
- Followed by the second phase "Parent Request to routers and REEDs".
- If both fails (no response), then we enter the third phase and send Announce messages (which we send over ~1400 msec interval on all channels)
- At the start of this third phase, we also send a third "Parent Request" to add another chance to find a parent.
- During this phase we still process "Parent Response" from any parent and try to attach to it.

One thing to note is If we change the timing here, it can impact overall time a SED device stays in rx per attach attempt and can then impact battery lifetime (we then need to update the attach backoff parameters accordingly)

jwhui: OK, I believe the third MLE Parent Request is not a behavior stated in the Thread Specification. We should probably add a comment to that effect and describe its purpose.

abtink: Yes. It's extra one to add more chance of attaching while sending announce messages (and being in rx mode).

One thing I noticed is in the third phase, we may accept a parent more quickly and not wait long to collect more parent responses (though we can argue that if we don't any response in previous phases, then it is ok to accept one more quickly).

This is handled from the top of this method itself (i.e. `Mle::HandleAttachTimer()` where we check `mParentCandidate` and decide whether to send "Child ID reqeust" to it). We can add code here to wait longer (e.g. check if we are in `kAttachStateAnnounce` state, wait till we have no more channels to send announce (which indicates we are end of the phase).  I am not sure if it worth it to change this though. Thoughts?

librasungirl: @abtink Thanks for your explanation. I simply thought them as linear procedures - 3 parent requests then announce, and didn't get your intention of leveraging announce period and giving more chance to receive parent response by issuing the third parent request at the beginning of orphan announce. 

---

For the third phase, it seems to me not only `not wait long to collect more parent responses`, but also `not wait long enough to collect announce responses on all channels`. 

According to existing implementation, once the orphaned device receives an Announce Response from a different channel/pan, it will start a timer of 250ms `kAnnounceProcessTimeout` to collect announce responses (from devices working on the channels this orphan device has already announced on) or parent responses on the working channel, and **won't continue to send orphan announce on the remaining channels any more**. If there is no parent response during this period, the device will switch to the pan/channel for attach attempt. (side note - if the channel is different,  the unicast parent responses that are randomly delayed `< kParentResponseMaxDelayAll 1000ms` on original working channel will occupy the working channel for a while for retx due to NoAck)

I think it should wait orphan announce on all channels ends if there are no parent responses, then choose a channel/pan with **highest active timestamp** for attach attempts. Additionally,  it also makes sense to collect parent responses during whole announce period and accept **the best one available** after orphan announce ends.

abtink: Regarding the announce, I don't think spec requires us to wait to collect from all channels.

After receiving an announce we do wait for `kAnnounceProcessTimeout` before processing it and listen for other Announce messages (with potentially higher timestamp).

We can consider adding this an extra behavior in OT implementation related to handling of annouce. Though I feel it adds complexity in the code for a not-so-likely situation. Also note that on the detached device, we may receive announce in other states during attach (not just while we are in `kAttachStateAnnounce`, i.e., after our own announce tx ) so we need to keep the existing logic in `HandleAnnounce()` (i.e. we need to keep state `kAttachStateProcessAnnounce` and start timer to listen for more announce msg on the same channel).

Here is my view on this:
- The change for handling parent request (wait till transition from `kAttachStateAnnounce ` before processing to attach to selected parent candidate) should be relatively simple. I will take care of it in a future PR.
- But the optimization/change for announce is not simple and in my view does worth the additional code overhead and complexity it adds.
- If you feel strongly about the announce change, please feel free to submit a PR, but I highly suggest keeping the change small. As you probably agree the MLE attach code is already complex and uses multiple state variables and is used in different situations (initial attach, better parent search, etc), so  we should be very careful when adding new states and code paths (to avoid the likelihood of unforeseen bugs).

abtink: - Submitted https://github.com/openthread/openthread/pull/6897

librasungirl: Agree that there is no explicit requirement to wait orphan announce finish in Thread Specification, though in 8.7.8 when trying to merge channel and pan id partitions, it is expected to finally merge to the dataset with latest active timestamp - which easily gives the impression that the available networks with latest activestamp should always be preferred.

No strong preference either. May take it as some optimization and postpone it when needed.

Comments? @jwhui

librasungirl:
Closing this as the delay design is intentional.

PR #6878, opened 2021-07-30
[routing-manager] delay removing external route for on-link prefix

wgtdkp:
This change includes three separate commits:
### [routing-manager] delay removing external route for on-link prefixes
Per RFC 4862 section 5.5.3, the SLAAC address of a on-link prefix
will not immediately be invalidated even if the on-link prefix is
sent with zero PIO valid lifetime but we currently removes the
external route for the on-link prefix when we see zero valid lifetime.
This results in an error case that a Thread device can not reach
a valid SLAAC address of a Wi-Fi host on the same infra link because
there is no routes for this address.
This commit fixes this issue by delay removing the external route
for the discovered on-link prefix untill its valid lifetime expires.
Meanwhile, the Border Router now considers a deprecated (zero preferred
lifetime) on-link prefix not usable. Thus, the Border Router will
start advertising its own on-link prefix when current on-link prefix
is deprecated.
### [routing-manager] replace mAdvertisedOnLinkPrefix with a boolean flag.
The mAdvertisedOnLinkPrefix can only be nullptr or mLocalOnLinkPrefix.
Change to use a boolean flag mIsAdvertisingOnLinkPrefix to make this
 explicit.

------
Prefer being merged as separate commits @jwhui

### DIFF
--- a/include/openthread/border_router.h
+++ b/include/openthread/border_router.h
@@ -87,6 +87,37 @@ otError otBorderRoutingInit(otInstance *aInstance, uint32_t aInfraIfIndex, bool 
 otError otBorderRoutingSetEnabled(otInstance *aInstance, bool aEnabled);
 
 /**
+ * This method returns the off-mesh-routable (OMR) prefix.
+ *
+ * The randomly generated 64-bit prefix will be published
+ * in the Thread network if there isn't already an OMR prefix.
+ *
+ * @param[in]   aInstance  A pointer to an OpenThread instance.
+ * @param[out]  aPrefix    A pointer to where the prefix will be output to.
+ *
+ * @retval  OT_ERROR_INVALID_STATE  The Border Routing Manager is not initialized yet.
+ * @retval  OT_ERROR_NONE           Successfully retrieved the OMR prefix.
+ *
+ */
+otError otBorderRoutingGetOmrPrefix(otInstance *aInstance, otIp6Prefix *aPrefix);
+
+/**
+ * This method returns the on-link prefix for the adjacent infrastructure link.
+ *
+ * The randomly generated 64-bit prefix will be advertised
+ * on the infrastructure link if there isn't already a usable
+ * on-link prefix being advertised on the link.
+ *
+ * @param[in]   aInstance  A pointer to an OpenThread instance.
+ * @param[out]  aPrefix    A pointer to where the prefix will be output to.
+ *
+ * @retval  OT_ERROR_INVALID_STATE  The Border Routing Manager is not initialized yet.
+ * @retval  OT_ERROR_NONE           Successfully retrieved the on-link prefix.
+ *
+ */
+otError otBorderRoutingGetOnLinkPrefix(otInstance *aInstance, otIp6Prefix *aPrefix);
+
+/**
  * This method provides a full or stable copy of the local Thread Network Data.
  *
  * @param[in]     aInstance    A pointer to an OpenThread instance.
--- a/src/cli/cli_network_data.cpp
+++ b/src/cli/cli_network_data.cpp
@@ -63,7 +63,7 @@ void NetworkData::OutputPrefix(const otBorderRouterConfig &aConfig)
     char  flagsString[kMaxFlagStringSize];
     char *flagsPtr = &flagsString[0];
 
-    OutputIp6Prefix(aConfig.mPrefix);
+    mInterpreter.OutputIp6Prefix(aConfig.mPrefix);
 
     if (aConfig.mPreferred)
     {
@@ -135,7 +135,7 @@ void NetworkData::OutputRoute(const otExternalRouteConfig &aConfig)
     char  flagsString[kMaxFlagStringSize];
     char *flagsPtr = &flagsString[0];
 
-    OutputIp6Prefix(aConfig.mPrefix);
+    mInterpreter.OutputIp6Prefix(aConfig.mPrefix);
 
     if (aConfig.mStable)
     {
@@ -157,15 +157,6 @@ void NetworkData::OutputRoute(const otExternalRouteConfig &aConfig)
     OutputPreference(aConfig.mPreference);
 
     mInterpreter.OutputLine(" %04x", aConfig.mRloc16);
-}
-
-void NetworkData::OutputIp6Prefix(const otIp6Prefix &aPrefix)
-{
-    char string[OT_IP6_PREFIX_STRING_SIZE];
-
-    otIp6PrefixToString(&aPrefix, string, sizeof(string));
-
-    mInterpreter.OutputFormat("%s", string);
 }
 
 void NetworkData::OutputPreference(signed int aPreference)
--- a/src/core/border_router/router_advertisement.hpp
+++ b/src/core/border_router/router_advertisement.hpp
@@ -238,6 +238,14 @@ public:
     void SetPreferredLifetime(uint32_t aPreferredLifetime) { mPreferredLifetime = HostSwap32(aPreferredLifetime); }
 
     /**
+     * THis method returns the preferred lifetime of the prefix in seconds.
+     *
+     * @returns  The preferred lifetime in seconds.
+     *
+     */
+    uint32_t GetPreferredLifetime(void) const { return HostSwap32(mPreferredLifetime); }
+
+    /**
      * This method sets the prefix.
      *
      * @param[in]  aPrefix  The prefix contained in this option.
@@ -261,7 +269,8 @@ public:
      */
     bool IsValid(void) const
     {
-        return (GetSize() == sizeof(*this)) && (mPrefixLength <= OT_IP6_ADDRESS_SIZE * CHAR_BIT);
+        return (GetSize() == sizeof(*this)) && (mPrefixLength <= OT_IP6_ADDRESS_SIZE * CHAR_BIT) &&
+               (GetPreferredLifetime() <= GetValidLifetime());
     }
 
 private:
--- a/src/core/border_router/routing_manager.cpp
+++ b/src/core/border_router/routing_manager.cpp
@@ -62,7 +62,7 @@ RoutingManager::RoutingManager(Instance &aInstance)
     , mIsEnabled(false)
     , mInfraIfIsRunning(false)
     , mInfraIfIndex(0)
-    , mAdvertisedOnLinkPrefix(nullptr)
+    , mIsAdvertisingLocalOnLinkPrefix(false)
     , mOnLinkPrefixDeprecateTimer(aInstance, HandleOnLinkPrefixDeprecateTimer)
     , mTimeRouterAdvMessageLastUpdate(TimerMilli::GetNow())
     , mDiscoveredPrefixInvalidTimer(aInstance, HandleDiscoveredPrefixInvalidTimer)
@@ -222,9 +222,9 @@ void RoutingManager::Stop(void)
 
     UnpublishLocalOmrPrefix();
 
-    if (mAdvertisedOnLinkPrefix != nullptr)
+    if (mIsAdvertisingLocalOnLinkPrefix)
     {
-        RemoveExternalRoute(*mAdvertisedOnLinkPrefix);
+        RemoveExternalRoute(mLocalOnLinkPrefix);
 
         // Start deprecating the local on-link prefix to send a PIO
         // with zero preferred lifetime in `SendRouterAdvertisement`.
@@ -235,7 +235,7 @@ void RoutingManager::Stop(void)
     SendRouterAdvertisement(OmrPrefixArray(), nullptr);
 
     mAdvertisedOmrPrefixes.Clear();
-    mAdvertisedOnLinkPrefix = nullptr;
+    mIsAdvertisingLocalOnLinkPrefix = false;
     mOnLinkPrefixDeprecateTimer.Stop();
 
     InvalidateAllDiscoveredPrefixes();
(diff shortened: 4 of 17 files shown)


include/openthread/border_router.h
```diff
@@ -86,6 +86,37 @@ otError otBorderRoutingInit(otInstance *aInstance, uint32_t aInfraIfIndex, bool
...
 otError otBorderRoutingSetEnabled(otInstance *aInstance, bool aEnabled);
 
+/**
+ * This method returns the off-mesh-routable (OMR) prefix.
+ *
+ * The randomly generated 64-bit prefix will be published
+ * in the Thread network if there isn't already a OMR prefix.
+ *
+ * @param[in]  aInstance  A pointer to an OpenThread instance.
+ * @param[in]  aPrefix    A pointer to where the prefix will be output to.
```
abtink, line 96:
```suggestion
 * @param[out] aPrefix    A pointer to where the prefix will be output to.
```

include/openthread/border_router.h
```diff
@@ -86,6 +86,37 @@ otError otBorderRoutingInit(otInstance *aInstance, uint32_t aInfraIfIndex, bool
...
+
+/**
+ * This method returns the on-link prefix for infrastructure link.
+ *
+ * The randomly generated 64-bit prefix will be advertised
+ * on the infrastructure link if there isn't already usable
+ * on-link prefix.
+ *
+ * @param[in]  aInstance  A pointer to an OpenThread instance.
+ * @param[in]  aPrefix    A pointer to where the prefix will be output to.
```
abtink, line 112:
```suggestion
 * @param[out] aPrefix    A pointer to where the prefix will be output to.
```

src/core/border_router/routing_manager.hpp
```diff
@@ -109,6 +109,35 @@ class RoutingManager : public InstanceLocator
      */
     Error SetEnabled(bool aEnabled);
 
+    /**
+     * This method returns the off-mesh-routable (OMR) prefix.
+     *
+     * The randomly generated 64-bit prefix will be published
+     * in the Thread network if there isn't already a OMR prefix.
+     *
+     * @param[in]  aPrefix  A reference to where the prefix will be output to.
```
abtink, line 118:
```suggestion
     * @param[out]  aPrefix  A reference to where the prefix will be output to.
```

src/core/border_router/routing_manager.hpp
```diff
@@ -109,6 +109,35 @@ class RoutingManager : public InstanceLocator
...
+    Error GetOmrPrefix(Ip6::Prefix &aPrefix);
+
+    /**
+     * This method returns the on-link prefix for infrastructure link.
+     *
+     * The randomly generated 64-bit prefix will be advertised
+     * on the infrastructure link if there isn't already usable
+     * on-link prefix.
+     *
+     * @param[in]  aPrefix  A reference to where the prefix will be output to.
```
abtink, line 133:
```suggestion
     * @param[out]  aPrefix  A reference to where the prefix will be output to.
```

src/core/border_router/routing_manager.cpp
```diff
@@ -514,21 +542,34 @@ const Ip6::Prefix *RoutingManager::EvaluateOnLinkPrefix(void)
...
                         smallestOnLinkPrefix->ToString().AsCString(), mInfraIfIndex);
-
-            // TODO: we remove the advertised on-link prefix by setting zero valid lifetime for PIO,
-            // But SLAAC addresses configured by PIO prefix will not be removed immediately (
-            // https://tools.ietf.org/html/rfc4862#section-5.5.3). This leads to a situation that
-            // a WiFi device keeps using the old SLAAC address but a Thread device cannot reach to
-            // it. One solution is to delay removing external route until the SLAAC addresses are
-            // actually expired/deprecated.
-            RemoveExternalRoute(*mAdvertisedOnLinkPrefix);
+            DeprecateOnLinkPrefix();
```
abtink, line 545: Just a question to make sure I undertsand the code:
- Here insetad of removing `mAdvertisedOnLinkPrefix ` as an external route immediately, we call `DeprecateOnLinkPrefix()` which starts a timer and from timer callback later we remove `mLocalOnLinkPrefix`.
- Is it then correct that `mAdvertisedOnLinkPrefix` can only be `&mLocalOnLinkPrefix` or `nullptr`.
   - Reading the code, I think this is indeed the case but want to confirm that there is no use-case/situation where it can be some other prefix.
   - If that's the case, can we consider (in a future PR) to change this to be a boolean flag (something like `mIsAdvertisingLocalPrefix` instead of being a pointer)? Does this make sense?

wgtdkp: @abtink You are right about above points!
I changed to use `mIsAdvertisingLocalOnLinkPrefix` in the third commit https://github.com/openthread/openthread/pull/6878/commits/383f12f4fabbc1f8f08a3d8750f8d562967ca995, PTAL.

src/core/border_router/routing_manager.cpp
```diff
@@ -199,20 +222,23 @@ void RoutingManager::Stop(void)
...
+    if (mIsAdvertisingLocalOnLinkPrefix)
     {
-        RemoveExternalRoute(*mAdvertisedOnLinkPrefix);
-    }
+        RemoveExternalRoute(mLocalOnLinkPrefix);
 
-    InvalidateAllDiscoveredPrefixes();
+        // Start deprecating the local on-link prefix to send
+        // a PIO with zero preferred lifetime.
+        DeprecateOnLinkPrefix();
```
simonlingoogle, line 231: mOnLinkPrefixDeprecateTimer is stopped later in this function. How is this method effective?

src/core/border_router/routing_manager.cpp
```diff
@@ -574,8 +613,8 @@ void RoutingManager::EvaluateRoutingPolicy(void)
     }
 
     // 3. Update advertised on-link & OMR prefixes information.
-    mAdvertisedOnLinkPrefix = newOnLinkPrefix;
-    mAdvertisedOmrPrefixes  = newOmrPrefixes;
+    mIsAdvertisingLocalOnLinkPrefix = (newOnLinkPrefix == &mLocalOnLinkPrefix);
```
simonlingoogle, line 616: Need to deprecate local On-Link Prefix when mIsAdvertisingLocalOnLinkPrefix is changed from true to false?

src/core/border_router/router_advertisement.hpp
```diff
@@ -264,7 +272,8 @@ class PrefixInfoOption : public Option
      */
     bool IsValid(void) const
     {
-        return (GetSize() == sizeof(*this)) && (mPrefixLength <= OT_IP6_ADDRESS_SIZE * CHAR_BIT);
+        return (GetSize() == sizeof(*this)) && (mPrefixLength <= OT_IP6_ADDRESS_SIZE * CHAR_BIT) &&
+               (GetPreferredLifetime() <= GetValidLifetime());
```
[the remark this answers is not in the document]
wgtdkp: Per RFC 4862, the receiver should treat this as invalid PIO and drop it:
https://datatracker.ietf.org/doc/html/rfc4862#section-5.5.3
> If the preferred lifetime is greater than the valid lifetime,
      silently ignore the Prefix Information option.  A node MAY wish to
      log a system management error in this case.

src/core/border_router/routing_manager.cpp
wgtdkp: The `SendRouterAdvertisement` depends on the `mOnLinkPrefixDeprecateTimer` status to know whether it should send such PIO or not. Updated the comment to make this clearer.

src/core/border_router/routing_manager.cpp
wgtdkp: Yes, it is done in `EvaluateOnLinkPrefix`. We update `mIsAdvertisingLocalOnLinkPrefix` here only after we have sent the PIO in `SendRouterAdvertisement`.

src/core/border_router/routing_manager.cpp
abtink: LGTM. Thanks for taking care of it (also thank for keeping it a separate commit, easy to check and review).

abtink (approved):

simonlingoogle (approved):
LGTM 👍

librasungirl (approved):

include/openthread/border_router.h
```diff
@@ -86,6 +86,37 @@ otError otBorderRoutingInit(otInstance *aInstance, uint32_t aInfraIfIndex, bool
...
+ * This method returns the off-mesh-routable (OMR) prefix.
+ *
+ * The randomly generated 64-bit prefix will be published
+ * in the Thread network if there isn't already an OMR prefix.
+ *
+ * @param[in]   aInstance  A pointer to an OpenThread instance.
+ * @param[out]  aPrefix    A pointer to where the prefix will be output to.
+ *
+ * @retval  OT_ERROR_INVALID_STATE  The Border Routing Manager is not initialized yet.
+ * @retval  OT_ERROR_NONE           Successfully got the OMR prefix.
```
jwhui, line 99:
```suggestion
 * @retval  OT_ERROR_NONE           Successfully retrieved the OMR prefix.
```

include/openthread/border_router.h
```diff
@@ -86,6 +86,37 @@ otError otBorderRoutingInit(otInstance *aInstance, uint32_t aInfraIfIndex, bool
...
+ *
+ */
+otError otBorderRoutingGetOmrPrefix(otInstance *aInstance, otIp6Prefix *aPrefix);
+
+/**
+ * This method returns the on-link prefix for infrastructure link.
```
jwhui, line 105:
```suggestion
 * This method returns the on-link prefix for the adjacent infrastructure link.
```

include/openthread/border_router.h
```diff
@@ -86,6 +86,37 @@ otError otBorderRoutingInit(otInstance *aInstance, uint32_t aInfraIfIndex, bool
...
+ *
+ * The randomly generated 64-bit prefix will be advertised
+ * on the infrastructure link if there isn't already usable
+ * on-link prefix.
```
jwhui, line 109:
```suggestion
 * The randomly generated 64-bit prefix will be advertised
 * on the infrastructure link if there isn't already a usable
 * on-link prefix being advertised on the link.
```

include/openthread/border_router.h
```diff
@@ -86,6 +86,37 @@ otError otBorderRoutingInit(otInstance *aInstance, uint32_t aInfraIfIndex, bool
...
+ *
+ * @param[in]   aInstance  A pointer to an OpenThread instance.
+ * @param[out]  aPrefix    A pointer to where the prefix will be output to.
+ *
+ * @retval  OT_ERROR_INVALID_STATE  The Border Routing Manager is not initialized yet.
+ * @retval  OT_ERROR_NONE           Successfully got the on-link prefix.
```
jwhui, line 115:
```suggestion
 * @retval  OT_ERROR_NONE           Successfully retrieved the on-link prefix.
```

src/cli/cli_network_data.cpp
```diff
@@ -158,15 +158,6 @@ void NetworkData::OutputRoute(const otExternalRouteConfig &aConfig)
     mInterpreter.OutputLine(" %04x", aConfig.mRloc16);
 }
 
-void NetworkData::OutputIp6Prefix(const otIp6Prefix &aPrefix)
```
jwhui, line 161: 👍🏼

src/core/border_router/routing_manager.hpp
```diff
@@ -109,6 +109,35 @@ class RoutingManager : public InstanceLocator
...
+    /**
+     * This method returns the off-mesh-routable (OMR) prefix.
+     *
+     * The randomly generated 64-bit prefix will be published
+     * in the Thread network if there isn't already an OMR prefix.
+     *
+     * @param[out]  aPrefix  A reference to where the prefix will be output to.
+     *
+     * @retval  kErrorInvalidState  The Border Routing Manager is not initialized yet.
+     * @retval  kErrorNone          Successfully got the OMR prefix.
```
jwhui, line 121:
```suggestion
     * @retval  kErrorNone          Successfully retrieved the OMR prefix.
```

src/core/border_router/routing_manager.hpp
```diff
@@ -109,6 +109,35 @@ class RoutingManager : public InstanceLocator
...
+     * This method returns the on-link prefix for infrastructure link.
+     *
+     * The randomly generated 64-bit prefix will be advertised
+     * on the infrastructure link if there isn't already usable
+     * on-link prefix.
+     *
+     * @param[out]  aPrefix  A reference to where the prefix will be output to.
+     *
+     * @retval  kErrorInvalidState  The Border Routing Manager is not initialized yet.
+     * @retval  kErrorNone          Successfully got the on-link prefix.
```
jwhui, line 136:
```suggestion
     * @retval  kErrorNone          Successfully retrieved the on-link prefix.
```

src/core/border_router/routing_manager.hpp
```diff
@@ -109,6 +109,35 @@ class RoutingManager : public InstanceLocator
...
+     * @param[out]  aPrefix  A reference to where the prefix will be output to.
+     *
+     * @retval  kErrorInvalidState  The Border Routing Manager is not initialized yet.
+     * @retval  kErrorNone          Successfully got the OMR prefix.
+     *
+     */
+    Error GetOmrPrefix(Ip6::Prefix &aPrefix);
+
+    /**
+     * This method returns the on-link prefix for infrastructure link.
```
jwhui, line 127:
```suggestion
     * This method returns the on-link prefix for the adjacent infrastructure link.
```

src/core/border_router/routing_manager.hpp
```diff
@@ -109,6 +109,35 @@ class RoutingManager : public InstanceLocator
...
+     *
+     * The randomly generated 64-bit prefix will be advertised
+     * on the infrastructure link if there isn't already usable
+     * on-link prefix.
```
jwhui, line 131:
```suggestion
     * on the infrastructure link if there isn't already a usable
     * on-link prefix being advertised on the link.
```

PR #6879, opened 2021-07-31
[message] simplify `Metadata`

abtink:
This commit simplifies the `Metadata` definition in `Message` by
rearranging the member variables (to avoid alignment gaps). It moves
the definition to be sub-type of `Buffer`. It also removes the
doxygen documentation for `private` or `protected` definitions
(mainly for simple getters/setters).

### DIFF
--- a/src/core/common/message.cpp
+++ b/src/core/common/message.cpp
@@ -53,6 +53,9 @@
 
 namespace ot {
 
+//---------------------------------------------------------------------------------------------------------------------
+// MessagePool
+
 MessagePool::MessagePool(Instance &aInstance)
     : InstanceLocator(aInstance)
 #if !OPENTHREAD_CONFIG_PLATFORM_MESSAGE_MANAGEMENT && !OPENTHREAD_CONFIG_MESSAGE_USE_HEAP_ENABLE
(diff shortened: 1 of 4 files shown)


src/core/common/message.cpp
```diff
@@ -937,7 +957,7 @@ void PriorityQueue::Dequeue(Message &aMessage)
     aMessage.Next()         = nullptr;
     aMessage.Prev()         = nullptr;
 
-    aMessage.SetMessageQueue(nullptr);
+    aMessage.SetPriorityQueue(nullptr);
```
simonlingoogle, line 960: This will set `mInPriorityQ = true`. Is this wanted behavior?

abtink: Yes. When the queue is set to `nullptr` the `mInPriorityQ` flag does not matter.

We want to use `SetPriorityQueue(nullptr)` here since we are dequeuing the message from a `PriorityQueue`. The use of `SetMessageQueue()` here is not technically correct and relies on the knowledge of private/internal behavior of `Metdata` and how it re-using the same pointer for both queue type which `PriorityQueue` class should not really know.

abtink: @simonlingoogle are you ok with this? Thanks.

abtink: Ping @simonlingoogle .

simonlingoogle: I am fine with this.

jwhui (approved):

PR #6881, opened 2021-08-02
[cmake] avoid unnecessary rebuild when git commit id changes

bukepo:
This commit removes the PACKAGE_VERSION definition from the global
config target so that when the git commit id is changed, only a few
sources will be re-compiled.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -239,3 +239,7 @@ endif()
 if(OT_RCP)
     include(radio.cmake)
 endif()
+
+set_property(SOURCE api/instance_api.cpp
+    APPEND PROPERTY COMPILE_DEFINITIONS "PACKAGE_VERSION=\"${OT_PACKAGE_VERSION}\""
+)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,7 +101,6 @@ if(OT_PACKAGE_VERSION STREQUAL "")
     ot_git_version(OT_PACKAGE_VERSION)
     message(STATUS "Setting default package version: ${OT_PACKAGE_VERSION}")
 endif()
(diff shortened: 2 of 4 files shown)


src/core/CMakeLists.txt
```diff
@@ -239,3 +239,7 @@ endif()
 if(OT_RCP)
     include(radio.cmake)
 endif()
+
+set_property(SOURCE api/instance_api.cpp
+    APPEND PROPERTY COMPILE_DEFINITIONS "PACKAGE_VERSION=\"${OT_PACKAGE_VERSION}\""
+)
```
simonlingoogle, line 245: Even though this can reduce the build time, but it seems not scalable having to add `PACKAGE_VERSION` for each source file separately?

bukepo: Not each source file, but only those need this definition. I don't think this macro would be widely used in this project.

simonlingoogle: Not sure if it works, but can we only use PACKAGE_VERSION in one header file and let other sources include this header file?

bukepo: I think it more friendly to @zhanglongxia if generating a header file.

simonlingoogle: Header file may not work since header files are not compile units. 

I think it will work if we add a cpp source file that contains code like:
```
const char * const gPackageVersion = PACKAGE_VERSION;
```
and another header file that contains code like:
```
extern const char * const gPackageVersion; 
```

Then, we can just use `gPackageVersion` in other source files.

simonlingoogle: That being said, I am fine with this PR as is.

simonlingoogle (approved):
LGTM :+1:

zhanglongxia (approved):
LGTM

jwhui (approved):

PR #6885, opened 2021-08-02
[tcp] add third_party module for TCPlp

samkumar:
This is the fifth in a series of pull requests (after #6744, #6770, #6790, and #6857) bringing TCPlp to OpenThread. I'm making a series of small pull requests so that it's easier for the OpenThread community to review the code. If you'd like to preview a working version of TCPlp in OpenThread, see #6650.

This pull requests creates a module in `third_party` for TCPlp code. It is a placeholder for now; the actual code will be added in future pull requests. In reviewing this pull request, please focus on the changes I've made to Makefiles to add the new `third_party` module. For now, the `tcplp.a` archive doesn't actually contain any code.

### DIFF
--- a/src/posix/Makefile.am
+++ b/src/posix/Makefile.am
@@ -57,6 +57,7 @@ LIBTOOLFLAGS_COMMON = --preserve-dup-deps
 LDADD_COMMON                                                           = \
     $(top_builddir)/src/posix/platform/libopenthread-posix.a             \
     -lutil                                                               \
+    $(top_builddir)/third_party/tcplp/libtcplp.a                         \
     $(NULL)
 
 if OPENTHREAD_TARGET_LINUX
--- a/tests/fuzz/Makefile.am
+++ b/tests/fuzz/Makefile.am
@@ -43,6 +43,7 @@ AM_CPPFLAGS                                               = \
 COMMON_LDADD                                              = \
     $(top_builddir)/src/core/libopenthread-ftd.a            \
     $(top_builddir)/third_party/mbedtls/libmbedcrypto.a     \
+    $(top_builddir)/third_party/tcplp/libtcplp.a            \
     $(LIB_FUZZING_ENGINE)                                   \
     $(NULL)
 
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -29,3 +29,5 @@
 if(NOT OT_EXTERNAL_MBEDTLS)
     add_subdirectory(mbedtls)
 endif()
+
+add_subdirectory(tcplp)
--- a/configure.ac
+++ b/configure.ac
@@ -1007,6 +1007,7 @@ src/lib/url/Makefile
 third_party/Makefile
 third_party/jlink/Makefile
 third_party/mbedtls/Makefile
+third_party/tcplp/Makefile
 examples/Makefile
 examples/apps/Makefile
 examples/apps/cli/Makefile
--- a/examples/apps/cli/Makefile.am
+++ b/examples/apps/cli/Makefile.am
@@ -40,6 +40,7 @@ CPPFLAGS_COMMON                                                       += \
     $(NULL)
 
 LDADD_COMMON                                                          += \
+    $(top_builddir)/third_party/tcplp/libtcplp.a                         \
     $(NULL)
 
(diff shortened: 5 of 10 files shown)


simonlingoogle (approved):
LGTM :+1: 
thanks

abtink (approved):
LGTM. 👍

samkumar:
@jwhui Is this OK to merge now?

src/posix/Makefile.am
```diff
@@ -57,6 +57,7 @@ LIBTOOLFLAGS_COMMON = --preserve-dup-deps
 LDADD_COMMON                                                           = \
     $(top_builddir)/src/posix/platform/libopenthread-posix.a             \
     -lutil                                                               \
+	$(top_builddir)/third_party/tcplp/libtcplp.a                         \
```
jwhui, line 60: nit: use spaces instead of tab
```suggestion
    $(top_builddir)/third_party/tcplp/libtcplp.a                         \
```

tests/fuzz/Makefile.am
```diff
@@ -43,6 +43,7 @@ AM_CPPFLAGS                                               = \
 COMMON_LDADD                                              = \
     $(top_builddir)/src/core/libopenthread-ftd.a            \
     $(top_builddir)/third_party/mbedtls/libmbedcrypto.a     \
+	$(top_builddir)/third_party/tcplp/libtcplp.a            \
```
jwhui, line 46: nit: use spaces instead of tab
```suggestion
    $(top_builddir)/third_party/tcplp/libtcplp.a            \
```

third_party/CMakeLists.txt
```diff
@@ -29,3 +29,5 @@
 if(NOT OT_EXTERNAL_MBEDTLS)
     add_subdirectory(mbedtls)
 endif()
+
+add_subdirectory(tcplp)
```
jwhui, line 33: Do we always want to add the `tcplp` subdir? Or be conditional on a CMake option?

jwhui (commented):
Looks good overall. Just a few nits.

third_party/CMakeLists.txt
samkumar: I believe there is a config option for enabling TCP, but I’m not sure if it’s a CMake option that is understood by the Makefile. I’ll look into it.

Other than that flag, I can’t think of anything else we might check before adding this.

src/posix/Makefile.am
samkumar: I've addressed this now in the latest version.

tests/fuzz/Makefile.am
samkumar: Thanks for finding this. I've addressed this now in the latest version.

third_party/CMakeLists.txt
samkumar: Hi Jonathan,
I looked into this. I believe that the current OpenThread configuration flag for TCP (`OPENTHREAD_CONFIG_TCP_ENABLE`) can't be inspected in `CMakeLists.txt`; it's not in `options.cmake`.

There's the question if we should add a CMake option for this. I feel it would be confusing to add it, since it would have to be kept in sync with the `OPENTHREAD_CONFIG_TCP_ENABLE` configuration flag; it wouldn't make sense to enable one but not the other. Additionally, my understanding is that even without a CMake option, `OPENTHREAD_CONFIG_TCP_ENABLE` is `0`, then the TCPlp library won't be used, and the linker will optimize it out. I checked this by using `arm-none-eabi-objdump` to inspect the final executable files and verify that `OPENTHREAD_CONFIG_TCP_ENABLE` set to `0` implies that the TCPlp functions aren't included in the final executable.

So the only advantage I can think of for adding a CMake option is to speed up the build when it is disabled (since the TCPlp files won't have to be compiled when the CMake option is disabled. Personally, I don't see enough of an advantage to doing this to justify the complexity of an additional CMake option that needs to be kept in sync with the existing `OPENTHREAD_CONFIG_TCP_ENABLE` flag.

@jwhui What is your opinion on this?

jwhui: I'm fine with the current PR. Thanks.

jwhui (approved):

PR #6891, opened 2021-08-04
[csl] fix switch between SSED -> MED -> SSED modes

canisLupus1313:
This commit introduces two changes. First is adding capablility to turn
off CSL, by implementing missing 0 value handling.
Secondly adds proper transition between SSED (CSL on) and MED states
(CSL off).

### DIFF
--- a/src/core/mac/mac.cpp
+++ b/src/core/mac/mac.cpp
@@ -2387,10 +2387,14 @@ void Mac::SetCslPeriod(uint16_t aPeriod)
 
     Get<DataPollSender>().RecalculatePollPeriod();
 
-    if (IsCslEnabled())
+    if ((GetCslPeriod() == 0) || IsCslEnabled())
     {
         IgnoreError(Get<Radio>().EnableCsl(GetCslPeriod(), Get<Mle::Mle>().GetParent().GetRloc16(),
                                            &Get<Mle::Mle>().GetParent().GetExtAddress()));
+    }
+
+    if (IsCslEnabled())
+    {
         Get<Mle::Mle>().ScheduleChildUpdateRequest();
     }
 
@@ -2399,7 +2403,12 @@ void Mac::SetCslPeriod(uint16_t aPeriod)
 
 bool Mac::IsCslEnabled(void) const
 {
-    return (GetCslPeriod() > 0) && !GetRxOnWhenIdle() && Get<Mle::MleRouter>().IsChild() &&
+    return !GetRxOnWhenIdle() && IsCslCapable();
+}
+
+bool Mac::IsCslCapable(void) const
+{
+    return (GetCslPeriod() > 0) && Get<Mle::MleRouter>().IsChild() &&
            Get<Mle::Mle>().GetParent().IsEnhancedKeepAliveSupported();
 }
 
--- a/src/core/thread/mle.cpp
+++ b/src/core/thread/mle.cpp
@@ -692,10 +692,10 @@ void Mle::SetStateChild(uint16_t aRloc16)
 #endif
 
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
-    if (Get<Mac::Mac>().IsCslEnabled())
+    if (Get<Mac::Mac>().IsCslCapable())
     {
-        IgnoreError(Get<Radio>().EnableCsl(Get<Mac::Mac>().GetCslPeriod(), GetParent().GetRloc16(),
-                                           &GetParent().GetExtAddress()));
+        uint32_t period = IsRxOnWhenIdle() ? 0 : Get<Mac::Mac>().GetCslPeriod();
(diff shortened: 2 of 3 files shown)


src/core/mac/mac.cpp
```diff
@@ -2387,6 +2387,12 @@ void Mac::SetCslPeriod(uint16_t aPeriod)
 
     Get<DataPollSender>().RecalculatePollPeriod();
 
+    if (GetCslPeriod() == 0)
```
edmont, line 2390:
```suggestion
    if ((GetCslPeriod() == 0) || IsCslEnabled())
```
Small suggestion for code simplification, allowing to remove the second `EnableCsl()` call below.

src/core/thread/mle.cpp
```diff
@@ -692,10 +692,17 @@ void Mle::SetStateChild(uint16_t aRloc16)
 #endif
 
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
-    if (Get<Mac::Mac>().IsCslEnabled())
+    if (Get<Mac::Mac>().IsCslCapable())
     {
-        IgnoreError(Get<Radio>().EnableCsl(Get<Mac::Mac>().GetCslPeriod(), GetParent().GetRloc16(),
-                                           &GetParent().GetExtAddress()));
+        if (!IsRxOnWhenIdle())
```
edmont, line 697: Suggestion for code simplification:
```
uint32_t period = IsRxOnWhenIdle() ? 0 : Get<Mac::Mac>().GetCslPeriod();
IgnoreError(Get<Radio>().EnableCsl(period, GetParent().GetRloc16(), &GetParent().GetExtAddress()));
```

edmont (approved):

src/core/mac/mac.cpp
canisLupus1313: Originally it was this way.
However this is the think which causes 1.2 csl certification test to fail. Since we are sending extra Child update request. Which will push through [this](https://github.com/openthread/openthread/blob/main/tests/scripts/thread-cert/v1_2_test_csl_transmission.py#L94) ping request.

edmont: I mean this:
```
    if ((GetCslPeriod() == 0) || IsCslEnabled())
    {
        IgnoreError(Get<Radio>().EnableCsl(GetCslPeriod(), Get<Mle::Mle>().GetParent().GetRloc16(),
                                           &Get<Mle::Mle>().GetParent().GetExtAddress()));
    }

    if (IsCslEnabled())
    {
        Get<Mle::Mle>().ScheduleChildUpdateRequest();
    }
```

src/core/thread/mle.cpp
```diff
@@ -692,10 +692,17 @@ void Mle::SetStateChild(uint16_t aRloc16)
 #endif
 
 #if OPENTHREAD_CONFIG_MAC_CSL_RECEIVER_ENABLE
-    if (Get<Mac::Mac>().IsCslEnabled())
+    if (Get<Mac::Mac>().IsCslCapable())
```
Irving-cl, line 695: I think the name `IsCslCapable` is a little confusing since `RxOffWhenIdle` is apparently a necessity for CSL capability. What about we change the code like this, still using the `IsCslEnabled`:
```
if (Get<Mac::Mac>().IsCslEnabled())
{
    IgnoreError(Get<Radio>().EnableCsl(Get<Mac::Mac>().GetCslPeriod(), GetParent().GetRloc16(),
                                           &GetParent().GetExtAddress()));
}
else if (!IsRxOnWhenIdle())
{
    IgnoreError(Get<Radio>().EnableCsl(0, GetParent().GetRloc16(), &GetParent().GetExtAddress()));
}
```
I think the code is the same in terms of logic. Thoughts? @edmont @canisLupus1313

src/core/mac/mac.cpp
canisLupus1313: fixed

src/core/thread/mle.cpp
canisLupus1313: @Irving-cl Unfortunatelly the IsCslEnabled method is not returning real state of CSL. The flag "RxOnWhenIdle" is stored in two places (MLE)[https://github.com/openthread/openthread/blob/main/src/core/thread/mle.hpp#L1611] and (MAC)[https://github.com/openthread/openthread/blob/main/src/core/mac/mac.hpp#L855]. During transitions between SSED -> MED and MED -> SSED those two aren't synchronized. 
The synchronization happens in sending child update request (here)[https://github.com/openthread/openthread/blob/main/src/core/thread/mle.cpp#L2366] or on reception of child update request reply (here)[https://github.com/openthread/openthread/blob/main/src/core/thread/mle.cpp#L3840]. Ofcourse both of those places are already to late to EnableCsl.
That's why I added extra method which is independent from `IsRxOnWhenIdle` from MAC.

Irving-cl (approved):
LGTM

jwhui (approved):
Thanks! 👍🏼

